Issue #9473, opened 2026-08-25
main is red on test_entity_ref_prefixes_are_schema_slot_names, and CI cannot catch it (test-kb is schema-gated)


dragon-ai-agent:
### 1. `main` is currently red on `test_entity_ref_prefixes_are_schema_slot_names`

Two entries carry pre-#9394 alias prefixes:

```
kb/disorders/ALG11-Congenital_Disorder_of_Glycosylation.yaml:1138
  discussions[1].attaches_to[0]='phenotype#Sensorineural Hearing Impairment'
  → should be 'phenotypes#Sensorineural Hearing Impairment'

kb/disorders/Major_Depressive_Disorder.yaml:2367
  'mechanistic_hypothesis#stalled_adult_hippocampal_neurogenesis_model'
  → should be 'mechanistic_hypotheses#stalled_adult_hippocampal_neurogenesis_model'
```

Reproduced locally on a branch that has never touched either file: **2 failed, 2423 passed**. Both refs are byte-identical on `origin/main`. This is a two-line fix.

### 2. CI structurally cannot catch this class of defect

`test_entity_ref_prefixes_are_schema_slot_names` lives in the whole-KB sweep, and `.github/workflows/main.yaml:416-418` gates that sweep on the **schema** having changed:

```yaml
# The whole-KB conformance sweep runs only when the schema itself (or the
# conformance-test logic) changes — the only edits that can flip a KB
# file's validity.
- name: Run whole-KB schema-conformance sweep
  if: steps.changes.outputs.schema == 'true'
  run: just test-kb
```

The comment states the assumption plainly: *the schema is the only edit that can flip a KB file's validity.* That holds for **existing** files. It does not hold for a **new or edited** KB file, which can introduce a non-canonical prefix with the schema untouched — and then the sweep never runs.

That is how both of these reached `main`: their PRs were KB-only, so `test-kb` was skipped, and `test (3.13)` went green.

The same thing happened on PR #9456: `test (3.13)` reports **success** on a head where `pytest tests/test_data.py` fails 2 of 15,076. Not a flake, not a stale baseline — the failing test simply is not in the job that ran.

## Why this is worth more than the two-line fix

Fixing the two entries turns `main` green and leaves the hole open. The gate is green for a **narrower reason than it appears** — the same shape as #9463, where `references[].title` is machine-checked but `EvidenceItem.reference_title` is not, and sixteen mis-attributed titles survived every passing validation run.

[...]

PR #7674, opened 2026-08-01
curate(FSHD): community-sourced phenotypes, registry natural history, and symptom-directed management

Phillip-a-richmond:
## Summary

Augments the existing `kb/disorders/Facioscapulohumeral_Muscular_Dystrophy.yaml` ([MONDO:0001347](https://monarchinitiative.org/MONDO:0001347)) with phenotypes, natural history, diagnostics, and treatments that **patient and family communities describe but the primary clinical literature under-represents**.

> **Scope note:** this task was opened as "curate FSHD1". An FSHD1-specific entry ([MONDO:0008030](https://monarchinitiative.org/MONDO:0008030)) was **not** created — at curator direction, FSHD1 stays a `has_subtypes` entry on the umbrella and the depth went into the umbrella instead. No duplicate entity.

The methodological point, and the reason this PR is worth reviewing as a pattern rather than just a content drop: **community sources were cited as first-class `url:` references and then adjudicated by literature.** Patient-organization pages and a public subreddit surfaced candidate signals; every one was then either backed by a citable study and curated as a phenotype, or recorded in `discussions` as an explicitly unvalidated lead. Nothing was promoted on social-media assertion alone.

**Worked example, including where it landed short:** r/FSHD had a "Migraines" thread. No FSHD headache literature turned up on a title search, so round 1 filed it as an *unvalidated* lead. Round 2 found the Dutch FSHD Registry reporting head-and-shoulder pain at 52% (n=373) — buried in a pain-location breakdown, not a headache paper. That justified curating a `Headache` phenotype. But review correctly caught that 52% is a **combined** category that does not isolate a head-specific rate, so **no frequency band is asserted** and the evidence item is `PARTIAL`. The community signal is corroborated, not quantified.

## Coverage

Entry now holds **21 phenotypes, 10 treatments, 5 diagnosis entries, 5 progression phases, 2 discussions, 13 top-level references, 93 evidence items across 34 distinct references** (8 of those evidence items are community-sourced; the other 85 are literature/trial).

[...]

PR #9654, opened 2026-08-27
evidence policy: patient-advocacy sources citable, social media not (design decision 6b)

mellybelly:
## The decision

`@mellybelly`, 2026-08-29: **published patient-advocacy and disease-foundation content is a citable source; user-generated social media is not.** Social media is deferred rather than settled — "may be included at a later date".

This PR records that as design decision 6b and enforces both halves. It contains **no KB content**; #7674 (FSHD) is the curation PR that raised the question and is unaffected by this one merging.

## What each half means

**Advocacy organizations — citable, with one rule.** A reference carries `tags: [PatientOrganization]` on its top-level `references:` entry, and every claim it supports must also carry at least one non-community citation. An advocacy page can corroborate a claim; it can never carry one. Enforced by `scripts/check_community_evidence.py`.

**Social media — not citable.** `scripts/check_social_media_references.py` fails any `url:` reference whose host is a known social-media host (20 hosts: Reddit, Facebook, X, Instagram, TikTok, YouTube, plus patient-forum platforms like HealthUnlocked and Inspire). `ReferenceTagEnum` has no `PatientCommunity` value, so LinkML rejects the tag outright:

```
[ERROR] 'PatientCommunity' is not one of ['GeneReviews', 'PatientOrganization']
```

Revisiting the decision means adding the enum value and the check back, not quietly relaxing a host list.

## Why the ban needs its own check

It cannot ride on the corroboration gate. That rule is opt-in by tagging, so an untagged forum URL is invisible to it — and every other gate passes a forum citation happily, because the URL resolves and the snippet is a genuine quote from the page. Nothing in the existing stack looks at *what kind of source* a URL is.

Host matching is suffix-based against an explicit list rather than heuristic. A rule like "any site with user accounts" catches preprint servers and registries, and a curator arguing with a false positive is how a gate gets switched off.

Both checks run **ungated and whole-KB** in CI, next to the entity-ref sweep, because the PRs that would introduce either problem are curation PRs touching only `kb/` — which match no `src/` or `tests/` path filter. That is the #9473 shape.

## What is left open, and where it is recorded

The §12 Gaps row now tracks the deferred social-media question rather than the whole policy, and records what actually made the forum half hard — the two things a future revisit has to answer:

- Citing a forum means quoting an identifiable account that did not consent to being a source.
- Caching one carries usernames and self-disclosed health status into git permanently. The r/FSHD listing cached in #7674 holds 19 usernames.

A national patient registry remains the consented substitute. In #7674 that substitution is what produced the head-pain and hearing-loss findings.

## Also in this PR

Three pre-existing problems surfaced while doing this:

1. **The §6 accepted-reference-type list** omitted `url:` entirely — despite 846 `reference: url:` lines across `kb/` — along with `ICTRP:`, `PPR:`, `GEO:`, `file:` and `metabolights:`, and named "NCT" where the KB uses `clinicaltrials:`. Corrected, pointing at `ALLOWED_REFERENCE_PREFIXES` as the authoritative list.
2. **`tag_references.py` wrote duplicate `tags:` keys.** Latent while the script was the slot's only writer; reachable once curators can tag. It now appends to an existing list in both block and flow style. `tests/test_tag_references.py` is new — there was no test file for that script, which is why this went unnoticed.
3. **The `tags` slot description** claimed the whole slot is script-populated. True only of `GeneReviews`.

## Not regenerated on purpose

`src/dismech/datamodel/` already lags `main`'s own schema by ~4,900 lines (`dismech.py`) and ~12,500 lines (`dismech_pydantic.py`) before this change. Regenerating would bury a small policy change under pre-existing drift, and CI regenerates it via `gen-python` rather than diffing the committed copy. Worth its own issue.

## Validation

- `check_community_evidence.py` and `check_social_media_references.py` — both clean across 2,742 KB files.
- Social-media gate verified to exit 1, naming both sites, on a synthetic entry carrying the exact Reddit reference and evidence item from #7674 — while leaving the FSHD Society page in the same file alone.
- `pytest -k "community or social_media or schema_validity"` — 5,489 passed.
- `pytest tests/test_tag_references.py` — 9 passed.
- `linkml-validate` clean on `kb/disorders/Asthma.yaml`, on a synthetic `PatientOrganization` entry, and on the history record; `PatientCommunity` correctly rejected.
- `.github/workflows/main.yaml` parses.

## Follow-up if this merges

The FSHD entry on #7674 needs its 7 FSHD Society references tagged `PatientOrganization`, and its 3 Reddit citations plus 2 cached Reddit pages removed. That is a change to someone else's branch, so it is not in this PR.

https://claude.ai/code/session_01G7RS5uK5KBRJbtSm1D8cDH


mellybelly:
All three items fixed in `5995509e1`, and the first one was the important catch — the gate did not run on the PRs it exists to police, which made §6b's "Enforced by" claim false as written.

**Item 1 — gate now runs ungated and whole-KB.** Confirmed the path analysis: `just test-kb` is gated on `steps.changes.outputs.schema`, whose filter is `src/dismech/schema/**`, `groupings.py`, `entity_refs.py`, `tests/test_data.py`. A `kb/`-only curation PR matches none of them, so a `kb_data`-marked test never runs on it. Followed the `check_entity_refs.py` precedent exactly: the rule moved to `src/dismech/community_evidence.py`, `scripts/check_community_evidence.py` runs it whole-KB, and an ungated CI step sits directly after the entity-ref sweep with the reasoning in a comment. The pytest sweep stays as a second lane. §6b now names the script rather than the test, and adds a bullet saying the rule is opt-in by tagging — the point being that a reviewer of a community sweep has to check the tag was applied, since no tool can infer it.

Verified both directions: `OK: no community-only evidence blocks in 2586 KB file(s)` on the KB, exit 1 with the offending path named on a synthetic community-only phenotype.

**Item 2 — `add_tag_to_existing_entry` fixed.** Reproduced before fixing. It now appends to an existing `tags:` list at that list's own indentation, and only inserts a fresh key when there is none. Checked both cases parse back through `yaml.safe_load`:

```
tags:
- PatientOrganization
- GeneReviews      # appended; one `tags:` key, existing tag preserved
```

**Item 3 — provenance test shape fixed.** You are right that `PublicationReference` has no `evidence` slot. The test now builds `references[].findings[].evidence`, which is the real shape, and `_is_reference_provenance` matches `references` exactly or `references[` rather than any path starting with those characters — so a future top-level slot named e.g. `references_reviewed` is not skipped by accident.

**Suggestions taken:** parametrized over `REFERENCE_BEARING_FILES`, which adds `kb/groupings` (2498 → 2586 files); finished the §6 reference-type list with `ICTRP:`, `PPR:`, `GEO:`, `file:` and `metabolights:`, and corrected "NCT" to `clinicaltrials:`, with a note that the prose lags `ALLOWED_REFERENCE_PREFIXES`; added a §12 Gaps row carrying both the accept and the reject path, and recording the two questions the draft leaves open (whether `evidence_source` also needs a community value — I think not, since it describes study type — and that `tags:` is rendered nowhere in `src/dismech/templates/`).

**On the 611 figure: you are right and I have corrected it.** It came from a checkout made before `main` moved; `git grep -c "reference: url:" origin/main -- kb/` gives **846** today. Our distinct-URL counts still differ (I get 105 unique `reference: url:...` strings against your 116), so the PR body now cites only the line count we agree on. The substantive claim is unchanged: every `url:` host in `kb/` is institutional and there are zero community URLs.

Validation at this head: `check_community_evidence.py` OK over 2,586 files; `pytest -k community_sourced_evidence` 2586 passed; `pytest -k "community_sole_support or schema_validity"` 3 passed; `linkml-validate` clean on `Asthma.yaml`, on a synthetic entry carrying both tags, and on the history record; `.github/workflows/main.yaml` parses.

---

mellybelly:
You are right, and this one was worse than what it replaced. Fixed at the root in `b71c946d5`.

I reproduced your case exactly before changing anything — `tag_disorder_file` on an untouched copy of `Autosomal_Dominant_Charcot-Marie-Tooth_Disease_Type_2W.yaml` raised `ParserError: expected <block end>, but found '-'`, with `- GeneReviews` written directly after `tags: [GeneReviews]`.

Here is what the three versions do, block and flow:

| Version | `tags:` block style | `tags: [...]` flow style |
|---|---|---|
| `origin/main` | duplicate `tags:` key | duplicate `tags:` key |
| my round-1 commit | correct | **parse failure** |
| current | correct | correct |

That middle row is the regression: a defect `check-duplicate-keys` catches, traded for one that breaks the file.

**Fixed where you said to.** `has_tag_in_entry` now reads flow style, so the seven flow-style files report `already_ok` and nothing is written to them at all. `add_tag_to_existing_entry` rewrites a flow list in place rather than appending to it, which covers the `tags: []` case that does still need a write and keeps the entry in the style its author chose. Verified across every `kb/disorders` file carrying `tags: [`: zero parse failures, zero files modified.

**`tests/test_tag_references.py` is new**, and is the part I should have written the first time. There was no test file for this script, which is why the original duplicate-key defect sat there and why my fix for it broke something else without anything noticing. It covers both list styles, both failure modes (counting `tags:` occurrences in the text, since a duplicate key is invisible to `safe_load`), and asserts the tagger is a no-op on the committed flow-style entries — so this specific regression cannot recur silently.

**Both non-blocking items taken.** `just check-community-evidence` is defined next to `check-entity-refs` and chained into `just qc` — you are right that it was the one gate not in the chain. And the evidence-tree walker is now defined once in `community_evidence.py` and imported by `tests/test_data.py` instead of living in both.

**On the failed `test (3.13)` at the previous head:** it is infrastructure, not code. `astral-sh/setup-uv` got `##[error]fetch failed` pulling the uv version list from `raw.githubusercontent.com`, before any test ran. Nothing was executed, so there is no signal in it either way; the new push should give a real result.

Validation at this head: `pytest tests/test_tag_references.py` 9 passed; `pytest tests/test_data.py -k "community or evidence_items_have_references or schema_validity"` 4903 passed; `check_community_evidence.py` OK across 2,586 files; history record validates.

---

mellybelly:
@cmungall — the governance question is decided. @mellybelly's call: **advocacy organizations and society pages are citable; social media is not, and may be revisited later.**

The PR is updated from a proposal to that decision, in `4c2411419`. Three changes worth knowing about:

**`PatientCommunity` is gone from the enum.** An enum value that no entry may legally carry is a trap — someone would use it and nothing would stop them. LinkML now rejects it: `'PatientCommunity' is not one of ['GeneReviews', 'PatientOrganization']`.

**The ban is enforced, not just written down.** `scripts/check_social_media_references.py` fails any `url:` reference on a known social-media host, ungated and whole-KB in CI. It needs to be a separate check rather than part of the corroboration gate, because that gate is opt-in by tagging — an untagged forum URL is invisible to it, and every other check passes a forum citation happily since the URL resolves and the snippet is a real quote from the page. Verified against the actual case: it catches both the Reddit reference and the Reddit evidence item from #7674, and leaves the FSHD Society page in the same file alone.

**Deferred is recorded as deferred.** The §12 Gaps row now tracks the social-media question specifically, with the two things a future revisit has to answer: quoting an identifiable account that did not consent to being a source, and what a cached copy carries into git permanently — the r/FSHD listing cached in #7674 holds 19 usernames. Revisiting means adding the enum value and check back, not relaxing a host list.

Nothing about the advocacy half changed: `tags: [PatientOrganization]`, never sole support, gate unchanged.

Validation at this head: both gates clean over 2,742 files; `pytest -k "community or social_media or schema_validity"` 5,489 passed; `linkml-validate` clean and rejecting the removed tag as expected.

**Follow-up not in this PR:** the FSHD entry on #7674 needs its 7 FSHD Society references tagged and its 3 Reddit citations plus 2 cached Reddit pages removed. That is @Phillip-a-richmond's branch, so someone should confirm before pushing to it.

---

cmungall (changes requested):
The auto checks don't work, there are Reddit URLs referenced in one disease (fwiw this was manually examined and deemed appropriate for the context of patient experience... but if this is the policy decision then this should have been flagged

(Human)

cmungall:
My mistake I thought this had merged https://github.com/monarch-initiative/dismech/pull/7674

Approved

PR #9661, opened 2026-08-27
curate(SAPHO): Add SAPHO syndrome (MONDO:0019266) (#9650)

### LINKED ISSUE #9650: Curate SAPHO syndrome (MONDO:0019266)
Curate a dismech entry for **SAPHO syndrome** ([MONDO:0019266](https://monarchinitiative.org/MONDO:0019266)) — an autoinflammatory disorder linking sterile inflammatory bone lesions (osteitis, hyperostosis) and synovitis to neutrophilic skin disease (palmoplantar pustulosis, severe acne), where no organism accounts for the bone inflammation and antibiotics do not treat it.

Current read is `DISEASE`. The stub carries `mondo_parents: autoinflammatory syndrome` and no `mondo_descendants` block at all, which is the leaf signal, and the acronym names one syndrome rather than a union of separable diseases.

**Adjacency worth stating up front.** `kb/disorders/Chronic_Recurrent_Multifocal_Osteomyelitis.yaml` already covers pediatric CNO/CRMO and does not mention SAPHO anywhere. Many rheumatologists treat SAPHO as the adult end of the same spectrum, but MONDO keeps it as a sibling under *autoinflammatory syndrome* rather than under CRMO, and SAPHO adds the cutaneous and anterior-chest-wall axes that the CRMO entry does not cover. Plan is a separate entry that cross-references CRMO rather than restating or contradicting it.

**The part most worth getting right**: whether *Cutibacterium acnes* recovered from bone lesions is a trigger or a contaminant is genuinely unsettled. That belongs in `mechanistic_hypotheses` with the causal edges opting in via `hypothesis_groups`, plus a `KNOWLEDGE_GAP` discussion — not asserted as settled mechanism. Related note: `PSTPIP2` (CMO mouse) and `LPIN2` (Majeed syndrome) are monogenic mimics that inform the mechanism, not SAPHO causal genes, and should be labelled as such.

Stub: `stubs/SAPHO_Syndrome.yaml`

gingin77:
Adds a Disease entry for SAPHO syndrome — synovitis, acne, pustulosis, hyperostosis, osteitis — and retires its stub. Nothing in `kb/` mentioned SAPHO before this.

## What the entry says

The mechanism chain runs from innate immune dysregulation, through a self-amplifying TNF-alpha / IL-1beta / IL-6 / IL-8 / IL-23-IL-17 cytokine hub, out to two branches: neutrophil recruitment and bone remodeling imbalance. Those converge on sterile osteitis, hyperostosis and osteosclerosis, synovitis, and neutrophilic skin disease.

The bone node is the one worth a second look. Most sterile osteitis is lytic; SAPHO bone *proliferates*. Both arms of remodelling are deranged at once — the same cytokines that drive osteoclastogenesis also inhibit osteoblast differentiation — so lytic and sclerotic lesions coexist and the net result at the anterior chest wall is the hyperostosis the acronym records.

## Three things curated as unsettled

I have tried not to write a tidier disease than the literature describes.

**Is *Cutibacterium acnes* a trigger or a contaminant?** It is recovered from a fraction of SAPHO bone biopsies and it is a potent NLRP3 inflammasome activator, which would supply the IL-1beta everything downstream needs. But overt infection is absent in many patients, antibiotics help only transiently, and C. acnes is the commonest contaminant of any biopsy taken through skin. This is curated as an `EMERGING` `mechanistic_hypotheses` group whose causal edge opts in via `hypothesis_groups` rather than being asserted, plus a `KNOWLEDGE_GAP` discussion with a proposed experiment. The node is framed as intracellular *persistence*, not infection, because that is the only version compatible with both halves of the evidence.

**Are SAPHO and paediatric CNO/CRMO one disease or two?** The repo already has `Chronic_Recurrent_Multifocal_Osteomyelitis.yaml`, which does not mention SAPHO. I curated them separately — MONDO keeps SAPHO under *autoinflammatory syndrome* rather than under CRMO, and SAPHO adds the cutaneous and chest-wall axes — but the separation is a curation convention, not a biological finding. Kahn's criteria treat CRMO in a child as sufficient to diagnose SAPHO, and a 2025 review states the debate is ongoing. That is recorded as a knowledge gap and as a contested differential, not resolved by which file the content sits in.

**Why do bone and skin respond to different cytokines?** IL-1 blockade works musculoskeletally and is reported ineffective in skin. IL-17/IL-23 blockade does the reverse. Anti-TNF works in bone and joint but can paradoxically provoke skin lesions. A single cytokine-driven process should not behave that way, and no current account explains it. Third knowledge gap.

## No genetic section, deliberately

PSTPIP2, NOD2 and LPIN2 were the three strongest candidates, and a 38-patient cohort tested all three and found no association (PMID:20032092). Listing any of them as a disease gene would assert exactly what the cited study refutes.

PSTPIP2 does appear — as the genotype of the knockout mouse in `animal_models`, linked `PARTIALLY_RECAPITULATES` to the sterile osteitis node with two grounded readouts. The human negative result is recorded there as one of the model's two limitations; the other is that the mouse bone phenotype is osteolytic while the human signature is proliferative.

## Prevalence

Two records, not one, because the sources are not comparable. Two reviews carry the same 1-in-10,000 figure but label it differently — one a frequency, the other an incidence — so `measure_type` is `UNKNOWN` rather than guessed. A third source gives only an upper bound of <1/10,000 and says plainly that reliable epidemiological data do not exist.

## Validation

- `just validate-disorders` — passed; **62/62 snippets verified** against cached references
- `just check-entity-refs`, `just check-duplicate-keys` — clean
- `just check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — no new findings against baseline
- `just check-stubs` — 1710 stubs, no errors

Two references were fetched during research and are **not** included, because the entry does not cite them: PMID:37964684 (cache came back `content_type: unavailable`) and PMID:40303415.

Cache diff is 8 lines across 6 files, all four CURIEs this entry introduced: `MONDO:0019266`, `HP:0100774`, `HP:0100847`, `NCIT:C84237`.

No `conforms_to` — no module in `kb/modules/` covers sterile IL-1-driven osteitis, and the CRMO entry declares none either.

Closes #9650


gingin77:
All six IMPORTANT items and all five suggestions are addressed in c62cdd0. Snippet count went from 62/62 to 94/94 verified.

You were right about the diagnosis. The entry said three times that SAPHO is diagnosed by exclusion and never said what a clinician does. It now has a `diagnosis:` section — exclusion-based diagnosis with no biomarker, MRI as the proposed gold standard, bone scintigraphy, bone biopsy for the monostotic lesion, and multimodal imaging. The 16%-of-MRI-lesions figure for early radiography is in there, because it turns "MRI is better" into "a normal radiograph does not exclude active disease".

**On the Kahn/Benhamou criteria (item 2)** — this was the one that actually embarrassed the entry, and you put your finger on why. The CNO/SAPHO knowledge gap rests entirely on Kahn's criteria treating CRMO in a child as sufficient to diagnose SAPHO, and that claim was sitting in an unevidenced `distinguishing_features` bullet. It is now a `definitions[]` entry, `PHENOTYPE_ALGORITHM` / `ESTABLISHED_CRITERIA` / `validation_status: UNVALIDATED`, with structured `inclusion_criteria` and `exclusion_criteria`. The CRMO contrast you flagged is in the description: validated EULAR/ACR criteria there, none here.

**JAK inhibition (item 3)** turned out to be more interesting than a missing row. Your quote from PMID:38151265 reports the 13-patient study with *"no information was provided on bone involvement"* — but PMID:40004896 line 172 describes a second pilot where 9 of 12 patients had a bone response. The two disagree. I curated both, with a `NO_EVIDENCE` item on the first so the silence reads as silence rather than as a negative result, and noted the disagreement in the third knowledge gap. A JAK inhibitor hitting several hub cytokines at once is the natural test of whether the compartments separate, and the data are too small and too discordant to answer it.

**IBD (item 4)** is a phenotype, `category: Gastrointestinal`, matching how the CRMO entry places it. Being in the modified 2003 criteria is what makes it structural, so that clause carries its own evidence item. There is no `comorbidities:` slot on `Disease`, so the cardiometabolic, osteoporosis and 18%-fibromyalgia data went into a new `epidemiology:` section instead of an invented slot — the fibromyalgia figure matters because it confounds the pain endpoint every SAPHO treatment is judged on.

**Suggestions, all taken.** `biological_scale` on `Innate Immune Dysregulation` is now `CELLULAR` — you were right that only the GO term was molecular and the claim was not. Added the monogenic autoinflammatory differentials (PAPA/DIRA/DITRA/Majeed) and spondyloarthritis with both the HLA-B27 and the MRI discriminators, onset and sex distribution with the acne-in-men / PPP-in-women inversion, and sulfasalazine.

The pamidronate counter-datum was the best catch in the suggestions block, so I gave it more than a hedge: it is a second, explicitly contested `target_mechanisms` link from bisphosphonates to the skin node, carrying the 9/10 result as `PARTIAL` and both the permitted-NSAID confound and the failure to reproduce as `REFUTE`. The gap `rationale` now names it and the tofacitinib disagreement together.

**On item 6, the deep-research artifact** — there isn't one, and I'd rather say so than produce a retrospective file. This was curated from targeted PubMed searches plus the cached full texts, not through a DR provider. You noted you weren't asking for an artifact if the work was done another way; I'm confirming that it was. The gap pattern you spotted was real, though, and its cause was exactly what you diagnosed: I worked from abstracts and did not go back through the two long full-text caches. Items 1-5 all came out of re-reading those two files, which is a fair verdict on the first pass.

PR #9667, opened 2026-08-27
Curate complement component 2 deficiency (MONDO:0009006)

### LINKED ISSUE #9660: Curate complement component 2 deficiency (MONDO:0009006)
Curate a dismech entry for **complement component 2 deficiency** ([MONDO:0009006](https://monarchinitiative.org/MONDO:0009006)) — the most common inherited deficiency of the classical complement pathway, usually a 28-bp deletion in *C2*, causing susceptibility to encapsulated bacterial infection and a strong predisposition to SLE through failed clearance of immune complexes and apoptotic debris.

Preliminary read is `DISEASE`: single gene, one well-described mechanism, and no existing entry covers it. dismech curates `Late_Complement_Component_Deficiency.yaml` (C5–C9, terminal pathway / Neisseria susceptibility) but nothing for the early classical components, so this is the counterpart entry — a distinct mechanism (immune-complex clearance and tolerance) rather than a subtype of it.

Note there is also a `stubs/Classic_Complement_Early_Component_Deficiency.yaml` in the queue, which looks like the grouping over C1q/C1r/C1s/C2/C4. This claim is for the C2 disease entry only; the grouping stays for whoever takes that stub.

Stub: `stubs/Complement_Component_2_Deficiency.yaml`

matentzn:
Adds a Disease entry for C2 deficiency — the most common inherited deficiency of the classical complement pathway, and the early-pathway counterpart to the `Late_Complement_Component_Deficiency` entry already in the KB.

## Shape of the pathograph

C2 gene defect → classical pathway C3 convertase failure → **two parallel branches**:

- impaired opsonisation of encapsulated bacteria (pneumococcus, ~25% of homozygotes)
- defective immune complex and apoptotic debris clearance (SLE, ~25% penetrance)

Those two branches are exactly what distinguishes this from terminal complement deficiency, where the organism is *Neisseria* and autoimmunity is not the dominant long-term risk. The two entries should read as a matched pair.

Type I (28-bp exon 6 deletion → premature stop) and type II (missense → blocked secretion) are `has_subtypes`: same clinical entity, different molecular route to it.

## Snippets were inserted programmatically, on purpose

Every snippet in this entry was extracted as an exact substring of the cached reference text by script, not typed. The reason is concrete: PMID:35729719's full text embeds non-breaking spaces (`\xa0`) and thin spaces (` `) inside its statistics —

```
Reduced C2 levels in plasma (P\xa0=\xa02 × 10-9 ) and impaired function...
```

Typing that by hand produces a snippet that reads identically and matches nothing. 13/13 verify.

## Two numbers I did not let the entry overstate

**The prevalence figures are second-hand.** Both the 1:20,000 Swedish prevalence and the 0.01 carrier frequency come from the *introduction* of PMID:35729719, which cites them to earlier Swedish surveys rather than measuring them. They are curated with that stated in `notes`. Someone wanting primary attribution should chase the original sources.

**The two 25% figures are a coincidence.** ~25% of homozygotes get recurrent bacterial infection (PMID:12959222) and ~25% develop SLE (PMID:35729719). These are independent numbers from different studies describing different subsets of patients. `notes` says so explicitly, because the temptation to read "the same 25%" off a rendered page is obvious.

## No treatments section

The sources I consulted describe vaccination and prophylaxis for complement deficiency *as a class*, not outcomes in C2-deficient cohorts. Curating class-level guidance as C2 evidence would overstate what the citations support, so it is a `KNOWLEDGE_GAP` instead. If there is a C2-specific cohort I missed, I would like the pointer.

## Note on the queue

`stubs/Classic_Complement_Early_Component_Deficiency.yaml` is still open and looks like the grouping over C1q/C1r/C1s/C2/C4. This PR deliberately does not touch it — it is a separate call for whoever takes that stub, and this entry would be one of its members.

## Validation

| Check | Result |
|---|---|
| snippet verification | 13/13 exact substrings of cached references |
| entity refs | pass |
| duplicate keys | pass |
| history record | pass |

Closes #9660

---

### DIFF
--- a/kb/disorders/Complement_Component_2_Deficiency.yaml
+++ b/kb/disorders/Complement_Component_2_Deficiency.yaml
@@ -514,8 +514,17 @@ notes: >-
   laboratory, and severe infection is a common reason to be referred for one,
   whereas the homozygote series counts carriers of the genotype. The genotype
   denominator is the one to use for penetrance; the cohort denominator is the
-  one to use for what a diagnosed patient looks like. The phenotype is tagged
-  FREQUENT rather than OCCASIONAL to span both.
+  one to use for what a diagnosed patient looks like. The phenotype is
+  therefore tagged OCCASIONAL, following the genotype denominator, because HPO
+  frequency means the proportion of people with the disease. The 57.8% is
+  carried in the phenotype description and here as the referral-series figure,
+  not as the frequency tag.
+
+  An earlier version of this paragraph ended by tagging the phenotype FREQUENT
+  "to span both", which contradicted the argument the paragraph itself makes
+  and, after the tag was corrected, contradicted the curated value as well. It
+  is recorded because a stale rationale is worse than none: a later curator
+  reading it would have had a documented-looking reason to revert OCCASIONAL.
 
   The type I extended MHC haplotype is named in PMID:12959222 but carries no
   evidence item here. The only sentence stating it embeds the haplotype as a


matentzn:
All seven blocking findings addressed and pushed.

**Finding 1 — the bundling.** You're right, and the `biological_scale: ORGANISM` tag being the giveaway is a good catch: I wrote the tag to match the *description*, which is precisely the signal that the description had two claims in it. Now:

```
C2 Gene Defect → Classical Pathway C3 Convertase Failure
  ├→ Impaired Opsonophagocytosis (CELLULAR)  → phenotypes#Recurrent Bacterial Infection
  └→ Defective IC/Apoptotic Debris Clearance (CELLULAR)
       → Loss of Tolerance to Nuclear Autoantigen (ORGANISM) → phenotypes#Systemic Lupus Erythematosus
```

Both phenotypes are now reachable; they were orphans before. GO:0002434 and GO:0043277 are bound on the clearance node (finding 2 — and you were right that it was named after both and bound neither), GO:0006909 plus neutrophil/macrophage on the opsonophagocytosis node.

**Finding 3 — the one that mattered.** The 57.8% does qualify a claim I'd stated as settled, and I've kept both numbers rather than swapping one for the other, because I don't think they disagree:

> about 25% of C2-deficient **homozygotes** (PMID:12959222) against 57.8% of **diagnosed C2D patients** (PMID:31440263, 26/45)

Those are different denominators. A complement-laboratory referral series counts people who reached that laboratory, and severe infection is a common reason to be sent to one; a homozygote series counts carriers of the genotype. So the genotype denominator is the one to use for penetrance, and the cohort denominator is the one to use for what a *diagnosed* patient looks like. `notes` says this explicitly and the phenotype is `FREQUENT` to span both. If you read the ascertainment differently I'd rather hear it than have the entry quietly assert my reading.

The 24.4% AID figure is curated on the new tolerance node as a genuine independent replication of the 25% SLE penetrance — worth distinguishing from the *coincidental* 25%/25% pair the entry already warns about, so `notes` now separates the two cases.

**Findings 4 and 5** produced a `biochemical:` block: CH50 <10% of normal with preserved AP50 (which also substantiates the previously uncited "alternative and lectin pathways remain intact" claim on the convertase node — good catch), and IgG4 as the infection-branch penetrance modifier. Your framing of it as the counterpart to the C4A modifier is right and is how it's written up.

**Finding 6** — frequencies added. **Finding 7** — the deep-research report landed after your review and was pushed; it agrees on prevalence and mechanism and I adopted none of its phenotype suggestions, for reasons in that commit message.

## One thing I couldn't do, and filed instead

The type I extended MHC haplotype. The only sentence naming it is:

> Type 1 (28 bp deletion in C2 exon 6 on the **[HLA-B18, S042, DR2]** haplotype or its fragments)

The snippet matcher strips bracketed spans unless `literal_bracket_patterns` matches, and none covers an HLA haplotype — the audit even reports "the snippet matches the cached text exactly once `[HLA-B18, S042, DR2]` is kept". Fixing it means editing `conf/reference_validator_config.yaml`, which changes what every snippet in `kb/` is matched against; that's not a curation PR's business. The claim stays as unquoted background with the reason in `notes`, and I opened https://github.com/monarch-initiative/dismech/issues/9677 with a proposed pattern.

## Suggestions taken

`harrisons_chapter` + `iuis_category` matching the sibling entry; `distinguishing_features` and evidence on both differentials; `cell_types` on the opsonophagocytosis node; evidence on both `has_subtypes`. I also **removed** the uncited ~90% C1q SLE figure rather than citing it — it comes from the introduction of PMID:35729719, which cites it to earlier work and isn't about this disease, so it's a qualitative contrast now.

**Not done:** the `mappings:` block. I can't verify a C2 deficiency OMIM identity against the ontology from here without guessing at an ID, and a wrong mapping is worse than none.

On GeneReviews — I could reach PubMed from this session and confirmed your recollection: no GeneReviews chapter for C2 deficiency.

Validation: 21/21 snippets, schema, term (`--labels`), entity-ref, duplicate-key, and the three baseline gates all pass. No new cache rows.

Re-requesting review.

---

matentzn:
All three fixed. Finding A was a good catch and worse than you found.

**A — the dangling edge.** Confirmed and fixed. Rebuilding the graph now:

```
dangling targets: NONE
pathophysiology nodes with no incoming edge: ['C2 Gene Defect and Absent C2 Protein']

  C2 Gene Defect → Classical Pathway C3 Convertase Failure
  Convertase Failure → Impaired Opsonophagocytosis → phenotypes#Recurrent Bacterial Infection
  Convertase Failure → Defective IC/Apoptotic Debris Clearance
                     → Loss of Tolerance to Nuclear Autoantigen → phenotypes#Systemic Lupus Erythematosus
```

Single root, no fragments.

**Then I swept all five entries in this run with the same check, and found a second break — in CMT4K, on the branch you reviewed and recorded as fixed.** The `Delayed Basal Ganglia and Brainstem Injury` node pointed at `Cerebellar Ataxia` and `Putaminal and Periaqueductal Lesions` as bare names; neither is a pathophysiology node (one's a phenotype, one's an imaging finding). Both edges dangled. Your CMT4K review says *"Entity refs resolve; the graph is now one connected component"* — and it did resolve, because that's exactly the blind spot you identified here. Fixed on that branch too.

Two breaks, one session, both introduced *while acting on feedback asking me to connect the graph*, and one of them passed review. Filed as https://github.com/monarch-initiative/dismech/issues/9697 with the reproducer, since being more careful isn't the fix — the existing check reports success in a misleading way. I offered to write the patch.

**B — the `%%`.** Three of them, from Python percent-formatting during insertion. Fixed.

**C — AID is not SLE.** You're right, and it was wrong on two axes:

1. "AID" means autoimmune disease broadly; the same paper reports SLE separately at 16/51.
2. It comes from the same 45 patients I argue are ascertainment-inflated — so it can't be a clean penetrance replication even if the outcomes matched.

That second point is the sharper one, because I made the ascertainment argument *myself*, two screens away, and then leaned on the same cohort for a penetrance figure. Regraded `PARTIAL`, framing corrected in both the explanation and `notes`, denominator corrected from 67 to 45.

## Suggestions taken

**`FREQUENT` → `OCCASIONAL`.** You're right that it contradicted the entry's own conclusion. HPO frequency is genotype-denominated, my note argues the genotype denominator is the penetrance one, and then the tag followed the referral-series number. The 57.8% is carried in `description` and `notes` where it belongs.

**Serum IgD added** as a deliberate negative — extraordinarily common in these patients but equally common with and without infections, which is what separates a co-occurring abnormality from a real risk modifier. Good suggestion; it makes the IgG4 record mean more by contrast.

## Not done, with reasons

- **`biomarker_term` on the biochemical records.** Agreed it's the right shape, but neither CH50 nor IgG4 is in `cache/`, so it needs a real ontology lookup rather than a plausible-looking CURIE. Given that three different sources in this session have handed me wrong or nonexistent HP CURIEs from memory, I'd rather leave these unbound than guess.
- **The opsonophagocytosis node's evidence being organism-level.** Correct, and I don't have a cellular-level citation for it. Rather than leave it looking sourced when it isn't, say the word and I'll either fetch one or move the organism-level items to the phenotype and leave the node unevidenced.
- **PMID:16026838, `mappings:`, ESID in the gap rationale.** Still open, all non-blocking. Happy to take them in a follow-up round if you'd rather have them here than in a later PR.

22/22 snippets, all gates, no new cache rows. Re-requesting review.

---

matentzn:
Finding D fixed — and you were right to block on it rather than leave it as a suggestion.

Your reason is the one that matters: `notes` is rendered curated content, so a later curator who read the rationale, saw it assert `FREQUENT`, then saw `OCCASIONAL` in the YAML would have had a *documented-looking* reason to revert a deliberate correction. A stale rationale is worse than no rationale, because it looks like someone thought about it.

The paragraph now states the choice actually made — `OCCASIONAL`, following the genotype denominator, which is what the rest of the paragraph already argued. The 57.8% is carried in the phenotype description as the referral-series figure. I also recorded that the earlier sentence was wrong rather than silently replacing it, for the same reason you gave.

**This is the second notes-drift in this run.** Adding phenotypes to my Fuhrmann entry left its `notes` miscounting them and misattributing one to the disease name. Same cause both times: I edited a curated value and didn't re-read the prose that explains it. Worth naming, because the two are far enough apart that it's a habit, not an accident — and neither is caught by any gate, since prose has no referential integrity.

## On the `check_entity_refs` gap

You suggested filing an issue if one didn't exist — it does: https://github.com/monarch-initiative/dismech/issues/9697, opened after finding A here, with the reproducer and the CMT4K second instance. Your independent graph rebuild is the third confirmation that a bare `downstream[].target` is unchecked while everything reports green.

Thanks for verifying finding C against the cache rather than taking my commit message for it. Given I made a false "reworded" claim on a sibling PR this session, that's the right default.

## Carried, not done

- **`Impaired Opsonophagocytosis` evidence sits one scale up.** Accurate. I have no cellular-level opsonophagocytosis citation for C2 deficiency specifically, and I'd rather the node bind `GO:0006909` with organism-level evidence and be legible than carry a manufactured cellular quote. Open to being pushed on it.
- **PMID:16026838** — still uncached, still the third independent source on the infection-rate paragraph. I keep deferring this because it's marked non-blocking; say the word and I'll fetch it, since it would strengthen the one paragraph you've called the best thing in the PR.
- **`biomarker_term`, `mappings:`, ESID in the gap rationale, IgG2/Factor B** — all noted, all deferred for the same reasons as before. The ORPHA:169147 idea is blocked by [#9687](https://github.com/monarch-initiative/dismech/issues/9687) like the rest of the Orphanet work.

22/22 snippets, graph single-rooted with no dangling targets, all gates. Prose-only change; no curated value moved. Re-requesting review.

---

PR #9668, opened 2026-08-27
curate: Add Short-Rib Thoracic Dysplasia 6 With or Without Polydactyly (MONDO:0009894, NEK1)

### LINKED ISSUE #9623: Curate short-rib thoracic dysplasia 6 with or without polydactyly (MONDO:0009894)
Curate a dismech entry for **short-rib thoracic dysplasia 6 with or without polydactyly** ([MONDO:0009894](https://monarchinitiative.org/MONDO:0009894)) — a skeletal ciliopathy caused by biallelic *NEK1* (`hgnc:7744`) variants, historically called short-rib polydactyly syndrome Majewski type. NEK1 is a ciliary kinase, so the mechanism is defective ciliogenesis and disrupted Hedgehog signaling in the developing growth plate, giving a constricted thoracic cage, short ribs and perinatal respiratory insufficiency.

The KB has `Short-Rib_Polydactyly_Syndrome.yaml`, and MONDO gives this term two parents — `Jeune syndrome` and `short rib-polydactyly syndrome, Majewski type`. Check the existing entry first: this may be best as a `has_subtypes` row rather than a standalone entry.

Note *NEK1* also appears in the KB as an ALS gene; the ciliopathy and ALS roles are separate claims and should not be conflated.

Stub: `stubs/Short-rib_Thoracic_Dysplasia_6_With_Or_Without_Polydactyly.yaml`

cmungall:
New Mendelian entry for the NEK1-related skeletal ciliopathy (SRP type II / Majewski).

Five-node pathograph: NEK1 loss of function → defective ciliogenesis → impaired growth plate chondrogenesis → thoracic constriction and pulmonary hypoplasia. Death is perinatal from lung hypoplasia that is *mechanical* — the thoracic cage cannot accommodate lung growth — not from primary lung disease.

## Lump/split call — please weigh this

I curated it as a **standalone `DISEASE`, not a `has_subtypes` row**, and the overlap is real enough that I want it looked at rather than waved through.

`kb/disorders/Short-Rib_Polydactyly_Syndrome.yaml` already carries a `Type II (Majewski)` subtype attributed to NEK1, plus a digenic NEK1/DYNC2H1 inheritance block.

My reasoning for a separate entry:
- That subtype row **carries no MONDO binding**, so MONDO:0009894 was genuinely uncurated.
- It sits inside MONDO:0015461, the *lethal* SRPS grouping, which is narrower than SRTD6 — SRTD6 is "with or without polydactyly" and lists **both** Jeune syndrome and SRPS-Majewski as MONDO parents, spanning a wider severity range.
- The modern SRTD numbering is per-gene, matching existing KB practice for the DCM and HSP series.

**If you prefer the subtype route, the fix is small**: bind MONDO:0009894 to the existing Type II row and drop this entry. I don't have strong attachment to the call, only to it being made explicitly.

## Deep-research quality — flagging this rather than burying it

The `claude_code` report resolved 14/14 references but **validated only 1 of 3 quotes**, naming PMID:21211617 among the unsupported ones — and that is a paper this entry cites heavily. It also scored only **5/14 references on topic**.

I did not take any quoted material from the report. Every snippet here was read directly from the cached abstract or full text, and all 37 verify against the cache. The report was used for orientation only. Worth knowing if anyone else works from that file.

Falcon/Edison returned HTTP 402 again (key out of credit, consistent all session). No GeneReviews chapter exists for SRTD6.

## Where the evidence is deliberately hedged

- **Digenic NEK1/DYNC2H1 inheritance** is curated as a second `Inheritance` block, but *both* its evidence items are graded `PARTIAL`. It rests on individual families, not a systematic two-locus screen. The prenatal case with one paternal NEK1 splice allele and no second hit is included with an explanation saying plainly that it is consistent with an unfound second allele *or* a non-classical mechanism, and distinguishes neither.

- **The kat2J mouse carries `FAILS_TO_RECAPITULATE`** on the growth-plate node, with a `HUMAN_MODEL_MISMATCH` discussion. The line predates the human skeletal association by decades and entered the literature as a polycystic kidney disease model — so the organ system carrying essentially all of the human mortality has **no reported model phenotype at all**. The `limitations` field says explicitly that this is an absence of reported skeletal assessment, *not* a documented normal skeleton, so it can't be misread as evidence the mouse skeleton is spared.

- **The model's ciliary readouts point in two directions**, and both are recorded: cilium number is dramatically *reduced* in kat2J fibroblasts, but cilia lining the renal cysts are abnormally *long* and sometimes supernumerary. That is length- and number-dysregulation rather than failure to ciliate — consistent with NEK1 coordinating the cilium cycle, but not a clean quantitative match to the reduced ciliation seen in patient material.

- **A second discussion** covers the 3 of 12 Majewski-type cases with neither NEK1 nor DYNC2H1. The screening study reports a clinical correlate — holoprosencephaly and polymicrogyria in the mutation-negative group — which suggests those cases may be a *different entity* rather than unexplained instances of this one. That bears on whether they belong under this MONDO term.

- `genetic[]` notes that NEK1 is separately established as an **ALS risk gene**, and that the ciliopathy and ALS roles are distinct claims.

## Self-inflicted defects I caught before commit

Recording these because they're the kind of thing that slips through:

1. An evidence item was left with `explanation: NO_EVIDENCE_PLACEHOLDER` and a title-only snippet. Removed entirely rather than patched — a title is not a finding.
2. A typo in a discussion `attaches_to` reference (`Chongrogenesis`). Caught by `just check-entity-refs`, which is exactly the case that check exists for.

## Validation

`just validate-disorders` passes — schema, terms and references — with **37/37 snippets verified**.

Closes #9623

---

---

---

> [!NOTE]
> **Low Risk**
> Changes are curated knowledge-base YAML, ontology caches, and reference metadata with no runtime application or security-sensitive logic.
> 
> **Overview**
> Adds a new Mendelian disorder entry for **short-rib thoracic dysplasia 6 with or without polydactyly** (`MONDO:0009894`, **NEK1**), with evidence-backed pathophysiology (NEK1 loss of function → defective ciliogenesis → impaired growth plate chondrogenesis → thoracic constriction → pulmonary hypoplasia), phenotypes, inheritance (autosomal recessive plus proposed digenic NEK1/DYNC2H1), diagnosis, animal models, and discussions.
> 
> **Short-Rib Polydactyly Syndrome** Type II (Majewski) now binds `subtype_term` to `MONDO:0009894` and points to the full entry, with notes that clinical type II is broader than the NEK1-defined MONDO concept.
> 
> Review follow-ups in the same PR split the bundled thoracic/pulmonary pathograph node (with **UBERON:0002224** / lung grounding), add **histopathology** and **treatments** (including cross-disease Jeune thoracic surgery with explicit limits), add **DYNC2H1** as a cooperating locus, remove an unevidenced Hedgehog pathway modifier, drop a **kat2J** `FAILS_TO_RECAPITULATE` link where skeletal assessment was never reported, and rebind tibia findings to **HP:0002992** at **FREQUENT**.
> 
> Ontology **cache** CSVs and **references_cache** PMID files are updated or added to support validation and quoted snippets; **history** YAML records the curation and correction sessions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b466bc7cac46e40788b100251aec21b2d04d68cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
All nine findings addressed or answered.

**Finding 9 first, because it set the scope of everything else.** I re-ran the GeneReviews check as asked — four PubMed E-utilities searches: `GeneReviews[Book]` with `"short rib"`, with `NEK1`, with thoracic dysplasia / Jeune / chondrodysplasia / ciliopathy in the title, and a free-text search for the exact chapter title you named. The only NEK1 hit in the GeneReviews book is the **Amyotrophic Lateral Sclerosis Overview**. There are Ellis-van Creveld and Cranioectodermal Dysplasia chapters, but nothing for the SRTD group. No chapter exists; the original claim stands, and the searches are in the history record so nobody has to repeat them.

**Finding 3.** You were right, and the fix is removal rather than citation. Reduced Hedgehog output downstream of ciliary failure is well established in skeletal ciliopathies generally, but no reference cached here measures or states it for NEK1 — so it was an unevidenced annotation sitting among evidence-backed ones. The descriptor is gone, the inference stays in the `description` where it reads as reasoning, and a node note says what a source would have to report to restore it. That node isn't thin any more, because of the next item.

**Finding 4.** `PMID:24854045` now cited twice in a new `histopathology:` section — the expanded, irregular femoral hypertrophic zone, and the cortico-medullary renal cysts with portal fibrosis. You were right that it was the most relevant uncited item: the growth-plate lesion is the direct tissue-level evidence the chondrogenesis node was missing.

**Finding 2.** Split into `Thoracic Constriction` (TISSUE, `UBERON:0002224`) and `Pulmonary Hypoplasia` (ORGANISM, `UBERON:0002048`) with a `DIRECT` edge. Both now carry ontology grounding, which the bundled node had none of. The group-level SRP type I–IV sentence stays on the constriction node, but its `explanation` now says it supports the rib lesion and *not* the mechanical consequence; that claim is evidenced separately on the hypoplasia node.

**Finding 5.** `DYNC2H1` added with `relationship_type: COOPERATING`. The choice isn't obvious so the reasoning is in `notes`: DYNC2H1 has its own biallelic Majewski cases, which would make it *causative* for those patients, but this entry is keyed to `MONDO:0009894`, the NEK1-defined disorder. What earns it a place here is the digenic observation the inheritance block already asserts.

One thing I found while doing it: the digenic snippet belongs to `PMID:21211617`, not `PMID:22499340` — I'd copied it to the wrong reference. The two papers report the same double-heterozygous family, and both are now cited with that stated, which also makes clear the digenic claim rests on one family reported twice.

**Finding 6.** Both kat2J MEF items regraded `MODEL_ORGANISM` → `IN_VITRO`. The whole-animal PKD claim keeps `MODEL_ORGANISM`.

**Finding 7.** `treatments:` added, three entries, each with its boundary stated. Perinatal palliative care carries **no evidence item at all** — no cached reference reports a palliative protocol for this disorder, and a quote that doesn't say what the entry claims would be worse than none. Thoracic expansion surgery is curated with an explicit limit: a Jeune intervention in two patients, and no Majewski-type patient survives to be a candidate. It's in because the surgical target is the same pathograph node, and a reader asking whether anything can be done for the chest should find the answer with its limits attached rather than find nothing.

**Finding 8.** Done in the sibling file. `Short-Rib_Polydactyly_Syndrome.yaml`'s Type II row now carries `subtype_term` bound to `MONDO:0009894` and points here for the full curation. The note also records that the two aren't co-extensive: `MONDO:0009894` is the NEK1-defined disorder, while type II as a clinical class also covers the DYNC2H1 cases and the 3 of 12 with neither gene.

**Bugbot's tibia finding was right.** `Absent Tibia` was `VERY_FREQUENT` on the strength of "short ovoid tibiae **or** tibial agenesis" — a disjunction where only the second limb is absence. Now `FREQUENT`, with a note that the HP term covers part of what the source describes and the short-ovoid form has no separate frequency figure anywhere in the cache.

Also: the ALS pleiotropy note is now grounded in two cached ALS papers, both graded `NO_EVIDENCE` because they say nothing about the skeletal disease. They're quoted to make the separation concrete — a heterozygous missense risk effect against biallelic loss of function. And `DOI_10.1186_s12929-014-0063-5.md` is dropped as a duplicate of `PMID_25030234.md`.

**Not done, with reasons.** `prevalence:` — the 1:100,000–1:130,000 figure is in the deep-research report only, and no cached reference states it. `mappings:` for OMIM 263520 — the number appears inside quoted sentences but no cached record establishes the mapping. The extra phenotypes (trident acetabular roof, malrotation, CHD, IUGR) are in the research report, not in any cached abstract; only IUGR is quotable, and one phenotype isn't a real answer to that finding. Fetching those sources is the fix and it's a separate round.

Validation: `just validate-disorders` passes on both files, 49 of 49 snippets verified. Also run: `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`.

---

cmungall:
Both Bugbot findings addressed. Note the re-review at 15:48 ran against `8043efe5a`, the title-only commit — the eight blocking items were fixed in `382b3e670`, which landed afterwards, and the [comment above](https://github.com/monarch-initiative/dismech/pull/9668#issuecomment-5442166621) covers them. Including the GeneReviews check you asked to have re-run from a network-enabled environment: four PubMed searches of `GeneReviews[Book]` — `"short rib"`, `NEK1`, thoracic dysplasia/Jeune/chondrodysplasia/ciliopathy titles, and the exact chapter title. The only NEK1 hit in the book is the ALS Overview. No SRTD chapter exists.

This push handles the two Bugbot items you confirmed.

**The `FAILS_TO_RECAPITULATE` link is removed.** You and Bugbot are right, and this was mine to see: the link's own `limitations` field said plainly that the skeleton was never assessed, while the `relationship` slot asserted the model *fails*. That's a substantive negative claim, it's what exports consume, and encoding "nobody looked" that way put a claim into the graph that the prose directly underneath it denied making.

I removed the link rather than reframing it. There's no `relationship` value meaning "not assessed for this node", and `MEASURES` or `PERTURBS` would assert an experimental relationship that also doesn't exist. An absent link is the accurate encoding of an unassessed question, and the `nek1_srtd6_no_skeletal_model` discussion already carries the gap — which is where an open question belongs. The model now has a `notes` block saying the omission is deliberate and naming what a `FAILS_TO_RECAPITULATE` link would need: a study that assays the kat2J skeleton and finds it normal.

**The tibia phenotype is rebound.** You were right that fixing the frequency wasn't enough. It was `Absent Tibia` / `HP:0009556` / `VERY_FREQUENT`; the last round fixed the band and left the term, but both were wrong for the same reason — the source says "short ovoid tibiae **or** tibial agenesis", and `HP:0009556` encodes only the severe limb of that disjunction. It's now `Abnormal Tibia Morphology` / `HP:0002992`, the most specific term that accurately covers both, at `FREQUENT`.

Splitting into `HP:0005736` Short tibia and `HP:0009556` Absent tibia would be better still, and I've deliberately not done it: no cached reference gives separate frequencies for the two forms, so the split would mean inventing a distribution. That reasoning is in `notes` so the next curator can act on it if a source turns up.

On the top-level `references:` block — thanks for checking the actual distribution and downgrading it. That's a useful correction to have on the record.

Validation: `just validate-disorders` passes, 48 of 48 snippets verified, `check-entity-refs` OK.

---

cmungall:
Both fixed in `51fd7c9be`, plus the changelog-prose suggestion. One was right as reported; the other had a correct diagnosis and a fix I couldn't use.

**Finding 1 — confirmed.** The title said "Zebrafish Embryo", the cache says "Embryonic Zebrafish Pronephros". Copied verbatim. Same defect you raised on #9678 in the same round, and the same cause: I wrote it from the sense of the paper rather than copying it out of the cache. Worth naming as a pattern rather than two coincidences.

**Finding 2 — the node really was under-supported, but the replacement text isn't in the file.**

You pointed at three sentences at `references_cache/PMID_37822322.md` line 42. That file is **34 lines long**, and none of the three appear anywhere in it:

```
$ wc -l references_cache/PMID_37822322.md        →  34
$ grep -c 'pulmonary development'                →  0
$ grep -c 'respiratory distress is caused'       →  0
$ grep -c 'lung hypoplasia'                      →  0
```

I then swept all 12 cached references for any sentence mentioning lungs, pulmonary, respiratory or asphyxiation. The only hit is the BACKGROUND sentence the node was already quoting. So your *conclusion* was right — no disease-specific statement of pulmonary hypoplasia was available to cite — but the suggested sentences weren't available either.

Flagging that plainly because it's the same failure your own finding 1 is about: text that reads as quoted source but wasn't checked against the source. Easy to do, and the reason the exact-substring rule exists.

**What I did instead** — fetched two references that genuinely say it:

| Reference | Quote | Grade |
|---|---|---|
| PMID:23985472 | "Children with JS present with variable degrees of respiratory distress, frequently lethal in the neonatal period." | PARTIAL |
| PMID:19610081 | "respiratory distress was observed in eight patients and proved lethal in six" | PARTIAL |

Both PARTIAL for the same cross-disease reason as the existing citation. PMID:37822322 stays, with its explanation re-scoped to the thoracic premise it actually states.

I also narrowed the node description. It had asserted the lungs are intrinsically normal and the constraint is purely mechanical — nothing available states that, so it now reads as the standard account of the group rather than a measured finding, with a node note recording that a NEK1- or Majewski-specific lung measurement is what would move these to SUPPORT.

**Changelog prose** — rephrased both (growth-plate node, tibia phenotype) as standing statements about the biology and the term choice. They render on the public disorder page, so a note addressed to a reviewer is addressed to the wrong audience. Reasoning preserved in both.

**Not done:** the `HP:0002089` phenotype. The new references support respiratory distress and lethality, not a measured lung hypoplasia, so an HP-coded hypoplasia phenotype would assert more than the sources carry. The mechanism node covers it.

Validation: `just validate` clean with 53/53 snippets, `tests/test_reference_titles.py` 38 passed, plus entity-refs, duplicate-keys, terms, snippet-grading, snippet-length, title-snippets, folded-hyphens, term-cache-integrity.

---

cmungall:
**Correcting my own record: I was wrong, and I should not have said "fabricated."**

Your round-3 quotes are real. I re-fetched `PMID:37822322` and verified all three by exact substring match — they are on line 42 of the full text, exactly where you said. Your retraction was itself unnecessary; you had it right the first time.

**The cause was mine, and it's worth naming.** After each `just validate` run the reference cache expands where PMC now serves full text. I had been running `git checkout -- references_cache/` to discard that as unrelated churn and keep the PR narrow — and `PMID_37822322.md` was in one of those discard lists. So I threw away the full-text expansion, then checked the abstract-only version I had just restored, found nothing, and concluded you'd invented the quotes. Then you read the same committed file and concluded the same thing about yourself.

Two of us independently reached a false conclusion because I had been treating a cache expansion as noise. That habit is the actual defect here, not either of our readings.

**What changed in `d7321a9ad`:**

Committed the expanded caches instead of discarding them. Three gained full text — `PMID:37822322` (34→159 lines), `PMID:37585529` (43→185), `PMID:38986433` (37→115); two gained only a `full_text_attempted` flag. All existing snippets still verify.

Re-grounded the node on what's now available. It had rested on two cross-disease references for the respiratory outcome, with `PMID:37822322` carrying only the thoracic premise and an explanation saying the paper "does not state" the pulmonary claim — true of the abstract, false of the paper. It now cites three sentences that reach the claim directly:

| Quote | Supports |
|---|---|
| "…an inability of the chest to support normal pulmonary development, function, and growth" | the mechanical claim |
| "the respiratory distress is caused by the anomalous morphology of the thorax…" | causal direction |
| "Highly symptomatic patients experience severe lung hypoplasia…" | hypoplasia itself |

All three stay `PARTIAL` — they describe Jeune syndrome and TIS as a group, not NEK1 or Majewski specifically, which the node notes still say.

Added the `HP:0002089` phenotype you suggested and I declined last round for want of a quote naming hypoplasia. That quote now exists. It carries no frequency deliberately: the statement is qualitative, group-level, and "highly symptomatic patients" is not a denominator.

**This resets the approval, which I judged worth it.** The PR can't merge until the 3-day age gate opens around 08-30, so there's ample time to re-earn it, and leaving a false statement about a source in rendered prose is exactly the defect class this entry has spent five rounds removing.

`PMID:37585529` and `PMID:38986433` also gained full text and are unmined — available for a future pass, nothing in the review points at them.

Validation: `just validate` clean with 56/56 snippets, `tests/test_reference_titles.py` 38 passed, plus entity-refs, duplicate-keys, terms, snippet-grading, snippet-length, title-snippets, folded-hyphens.

---

PR #9670, opened 2026-08-27
Curate hypomyelinating leukodystrophy 15 (MONDO:0054782)

### LINKED ISSUE #9657: Curate leukodystrophy, hypomyelinating, 15 (MONDO:0054782)
Curate a dismech entry for **leukodystrophy, hypomyelinating, 15** ([MONDO:0054782](https://monarchinitiative.org/MONDO:0054782)) — an autosomal recessive hypomyelinating leukodystrophy caused by biallelic variants in *EPRS1*, the glutamyl-prolyl-tRNA synthetase component of the multi-tRNA synthetase complex, presenting with developmental delay, hypomyelination on MRI and often microcephaly.

The stub records no `mondo_descendant_count` (MONDO leaf). Preliminary read is `DISEASE`: a single gene with one mechanism, and no existing dismech entry mentions *EPRS1* or HLD15 at all. It belongs to the aminoacyl-tRNA-synthetase leukodystrophy family, so worth checking against the existing hypomyelinating-leukodystrophy entries for whether a grouping already collects them.

Stub: `stubs/Leukodystrophy_Hypomyelinating_15.yaml`

matentzn:
Adds a Disease entry for HLD15, the EPRS1-related hypomyelinating leukodystrophy, joining `Hypomyelinating_Leukodystrophy_7` and `_10`.

## Three hypothesis groups, not one pathway

This is the part I'd most like reviewed. The literature offers a **unifying** model — the cytoplasmic aminoacyl-tRNA synthetase leukodystrophies (HLD15/EPRS1, HLD9/RARS1, HBSL/DARS1) all converge on reduced translation capacity during myelination — and then two **variant-specific** routes that do not look like the same lesion:

| Group | Variant | Mechanism |
|---|---|---|
| `reduced_translation_capacity` | all | proposed common endpoint: less aminoacylation → less myelin |
| `m6a_accessibility_defect` | p.Pro1482Thr | variant *masks accessibility* of variant-distal m6A sites → impaired METTL3 writing → defective mRNA export and translation |
| `mutant_protein_aggregation` | R339X | residual mutant protein aggregates in Rab7 vesicles, dysregulates Rab7 signalling, blocks oligodendroglial differentiation |

The third is a **gain-of-toxic-function**, which a pure loss-of-capacity model does not predict. If it's right, variants are not interchangeable — and that bears directly on whether a translation-directed therapy would help every patient. Curating them as one node would have erased that. All three are `EMERGING`; the first is explicitly framed as a hypothesis by its own authors, and the entry grades that snippet `PARTIAL` for exactly that reason.

The m6A finding is also notable as the only therapeutic handle in the entry: expression was rescued by antisense morpholinos and by METTL3-dCas13b targeted RNA methylation.

## Evidence grading — one paper, two gradings, on purpose

The clinical phenotype list (optic atrophy, dysphagia, dystonia, ataxia, spasticity) is graded **`OTHER`**, not `HUMAN_CLINICAL`. It comes from the *introduction* of PMID:33805425 — a cell-biology paper restating those features from other work, with no patients of its own. Grading it `HUMAN_CLINICAL` would claim a clinical observation the citing paper never made.

That same paper's own experiments are graded `IN_VITRO` (mouse oligodendroglial FBD-102b cells). Two `evidence_source` values from one PMID is `check-snippet-grading` working as designed — it keys on the quoted sentence, not the paper, and these are different sentences making different kinds of claim.

**This is a soft spot worth flagging**: a curator with the primary clinical cohorts should regrade those five phenotypes and cite them directly. I did not want to leave the phenotypes out, and I did not want to label a secondary restatement as primary clinical evidence, so they are in with honest provenance.

A `HUMAN_MODEL_MISMATCH` discussion records that the m6A rescue happened in lymphoblastoid cell lines — not oligodendrocytes, not in vivo — so restored expression is not yet restored myelination, particularly if the myelination window has closed by diagnosis.

## Naming

Named `Hypomyelinating_Leukodystrophy_15` to match the existing `_7` and `_10` entries, rather than the MONDO label's word order ("leukodystrophy, hypomyelinating, 15"). The MONDO label is used verbatim in `disease_term` and kept as a synonym. Say the word if you'd rather the file follow MONDO.

## Deep research

The Perplexity run for this disease was still queued behind retries when the entry was finished; it is curated from primary literature. I will add the report to this branch if the run completes.

## Validation

| Check | Result |
|---|---|
| snippet verification | 20/20 exact substrings of cached references |
| entity refs | pass |
| duplicate keys | pass |
| history record | pass |

Closes #9657

---

### DIFF
--- a/kb/disorders/Hypomyelinating_Leukodystrophy_15.yaml
+++ b/kb/disorders/Hypomyelinating_Leukodystrophy_15.yaml
@@ -474,8 +474,9 @@ phenotypes:
 - category: Neurological
   name: Cerebral White Matter Atrophy
   description: >-
-    Supratentorial white matter atrophy, present in 7 of the 8 published
-    biallelic-EPRS1 patients. Distinct from the hypomyelination: tissue is lost,
+    Supratentorial white matter atrophy, listed for 6 of the 8 published
+    biallelic-EPRS1 patients in Table 1, with cortical atrophy — a distinct
+    finding — in a seventh. Distinct from the hypomyelination: tissue is lost,
     not merely under-myelinated.
   phenotype_term:
     preferred_term: Cerebral white matter atrophy
@@ -502,19 +503,20 @@ phenotypes:
   evidence:
   - reference: PMID:41721156
     reference_title: Compound heterozygosity of a novel missense variant and exonic deletion in hypomyelinating leukodystrophy 15.
-    supports: PARTIAL
+    supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
-    snippet: "Brain MRI revealed extensive cerebral white matter atrophy accompanied by hyperintensity on T2-weighted images, with thinning of the corpus callosum"
+    snippet: "hypomyelination; supratentorial atrophy; cerebellar atrophy; corpus callosum thinning"
     explanation: >-
-      Graded PARTIAL: this sentence documents the atrophic imaging pattern in
-      the authors' patient but names the cerebral white matter rather than the
-      cerebellum. The cerebellar involvement is tabulated across the published
-      series in the same paper's Table 1.
+      A Table 1 imaging row naming cerebellar atrophy directly. This replaces an
+      earlier quote about cerebral white matter atrophy, which described a
+      different structure and so could not support this claim at any grade.
 - category: Neurological
   name: Microcephaly
   description: >-
-    Microcephaly, reported in the two siblings homozygous for the
-    Zn2+-binding-domain variant.
+    Microcephaly, reported in the two siblings homozygous for p.Pro1482Thr.
+    PMID:41721156 files that variant under a "Zn2+-binding-domain" label; the
+    zinc-binding domain is a ProRS sub-domain, so this is the same variant the
+    variants: section lists as ProRS-domain, not a separate one.
   phenotype_term:
     preferred_term: Microcephaly
     term:
@@ -559,15 +561,45 @@ phenotypes:
     explanation: >-
       Primary clinical observation.
 - category: Neurological
+  name: Intellectual Disability
+  description: >-
+    Intellectual disability, recorded alongside the developmental delay in the
+    two siblings and tabulated for three of the eight published patients. In the
+    adult-onset 2026 case it is the presenting feature and is mild, which is the
+    milder end of the severity spectrum this entry records.
+  phenotype_term:
+    preferred_term: Intellectual disability
+    term:
+      id: HP:0001249
+      label: Intellectual disability
+  evidence:
+  - reference: PMID:38769304
+    reference_title: "Homozygous EPRS1 missense variant causing hypomyelinating leukodystrophy-15 alters variant-distal mRNA m6A site accessibility."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "The Proband, Sibling 1, is an 18-year-old male with severe global developmental delays and intellectual disability, ataxia, microcephaly, rotatory nystagmus, axial hypotonia, and progressive bilateral lower limb spasticity on serial neurological examinations."
+    explanation: >-
+      Primary clinical observation, in the same sentence already cited for the
+      developmental delay. Intellectual disability is named as a finding
+      distinct from the delay rather than as a restatement of it.
+  - reference: PMID:41721156
+    reference_title: Compound heterozygosity of a novel missense variant and exonic deletion in hypomyelinating leukodystrophy 15.
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "a 40-year-old woman with mild intellectual disability, ataxia, dystonia"
+    explanation: >-
+      The 2026 adult-onset case, where mild intellectual disability is the
+      presenting feature rather than an accompaniment to severe delay.
+- category: Neurological
   name: Axial Hypotonia
   description: >-
     Axial hypotonia coexisting with progressive lower-limb spasticity — the
     mixed tone pattern typical of a hypomyelinating leukodystrophy.
   phenotype_term:
-    preferred_term: Hypotonia
+    preferred_term: Axial hypotonia
     term:
-      id: HP:0001252
-      label: Hypotonia
+      id: HP:0008936
+      label: Axial hypotonia
   evidence:
   - reference: PMID:38769304
     reference_title: "Homozygous EPRS1 missense variant causing hypomyelinating leukodystrophy-15 alters variant-distal mRNA m6A site accessibility."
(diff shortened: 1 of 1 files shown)


matentzn:
All eight findings addressed and pushed.

**Finding 3 was the important one, and the inverted edge is a real bug** — thank you for catching it. The entry read reduced EPRS1 protein → m6A defect, where PMID:38769304 says the m6A accessibility defect is *why* the protein is reduced ("markedly reduced EPRS1 protein **due to** dual defects in nuclear export and cytoplasmic translation"). As drawn, reduced protein caused its own cause.

Your debundling diagnosis was right about the mechanism too: the bundling is what made the inversion invisible. Now:

```
Biallelic EPRS1 Variants
  ├→ Impaired m6A-Dependent mRNA Export and Translation → Reduced EPRS1 Protein
  │                                                          → Reduced Translation Capacity
  └→ Mutant EPRS1 Aggregation in Rab7 Vesicles → Insufficient CNS Myelination
```

The aggregation branch now **bypasses** the reduced-protein node entirely, which was your subtler point — hanging a gain-of-toxic-function node beneath a node asserting loss of protein was exactly the conflation the entry argues against. And you're right that the blanket `functional_impact_category` contradicted my own thesis; it's per-variant now, and PMID:33805425's explicit "could be loss-of-function... could be a toxic gain-of-function" is why R339X carries no blanket call.

**Finding 2 — wrong HPO term.** Confirmed: `HP:0033725` *Thin corpus callosum* exists. `HP:0002079` asserts developmental hypoplasia, which the source doesn't claim — it reports a radiological finding. Fixed.

**Finding 5 — experimental models.** This turned out to be the most valuable structural change. Both cell systems are now `ExperimentalModel` entries with `modeled_mechanisms`, readouts and limitations, which does two things: it moves the entry's only therapeutic handle (the morpholino / METTL3-dCas13b rescue) out of a free-text `description` and makes it queryable, and it gives the `HUMAN_MODEL_MISMATCH` discussion something structural to point at. Your `RESCUES` + `RESTORED` suggestion is exactly how the rescue arm is modelled.

I also took your `CL:0000128` note into `limitations` rather than the cell type — the source calls FBD-102b "a mouse brain neuronal stem cell line" used *as* a model oligodendrocyte, and that distinction belongs where a reader will weigh it.

**Findings 1, 4, 6, 7, 8:** DR report landed and is on the branch (it also surfaced PMID:41721156, a 2026 case that **contradicted my own description** — a mild adult-onset presentation where I'd written HLD15 as severe and childhood-onset; added before your review, with a `PARTIAL`-graded counterpoint). Pathograph reaches its phenotypes. `references:` + OMIM:617951 added — and you were right that the ID was sitting in my own cached source text. Hearing loss, visual loss and amblyopia added from sentences already in the file. Four `variants:` added, since the entry's thesis is that variants differ.

**Suggestions taken:** symptomatic `treatments:` and a `diagnosis:` block, both quotable and both scoped to *symptomatic* so the `hld15_no_curated_treatments` gap keeps its distinct meaning; the MSC-assembly negative result as a `PARTIAL` item against the limb of the unifying hypothesis it qualifies — that was a good spot, it sharpens a three-part proposal into one with a known exception.

**`parents:` was dangling** exactly as you said — `Hypomyelinating Leukodystrophy` matches no entry or grouping. Now uses the descriptive form HLD10 uses.

**Orphanet, for the record:** I couldn't consult it here either. `just refresh-orphadata` fails on a stale manifest pin (2025-12-09 pinned, 2026-06-23 served); filed as https://github.com/monarch-initiative/dismech/issues/9687. Not blocking for this entry but worth knowing, since two other reviews this session pointed me at the same recipe.

**GeneReviews:** ran the check — zero hits for `EPRS1[TI] GeneReviews[TI]`. As with HLD10, it doesn't exist.

Validation: 39/39 snippets, schema, term, entity-ref, duplicate-key, three baseline gates, and the model-readout pytest lane (4,955 passed). No new cache rows.

Agreed on the naming — staying with `Hypomyelinating_Leukodystrophy_15`.

Re-requesting review.

---

matentzn:
Correcting the record on one line in my previous comment.

I wrote "ran the check — zero hits for `EPRS1[TI] GeneReviews[TI]`" **before** actually running it. That was careless; I should not state a check's result until it has run.

I then ran it, and the conclusion holds — but by luck, not method, because **the query itself is broken**. `GeneReviews[TI]` searches the chapter title, and GeneReviews chapters carry "GeneReviews" as the *book* title, not in their own. It returns zero for everything.

The correct form is `<term> AND "GeneReviews"[Book]`. Re-running both ways:

| Query | `[TI]` | `[Book]` |
|---|---|---|
| EPRS1 | 0 | **0** |
| hypomyelinating leukodystrophy | 0 | **0** |
| TRPV4 (a control) | 0 | **3** |

So the EPRS1 answer is genuinely zero and nothing in this PR changes. But the control shows the method was worthless, and the same query led me to tell the Fuhrmann reviewer the same thing on the same bad basis. On my SEMD Maroteaux PR it produced an actual error — a missing GeneReviews chapter that the sibling Kozlowski entry already cites, and which contradicts a claim that entry made. Fixed there, and filed as https://github.com/monarch-initiative/dismech/issues/9694 since the query comes from the review skill.

Nothing else in the previous comment changes. Everything else I reported there I had run.

---

matentzn:
Fair blocker, and an embarrassing one: I upgraded that cache file to full text and then read the abstract. Fixed.

**Four phenotypes added**, all from the body: cerebral white matter atrophy (`HP:0012762`), cerebellar atrophy (`HP:0001272`), microcephaly (`HP:0000252`), seizures (`HP:0001250`).

The grading is where this got interesting — **the same paper needs three different `evidence_source` values depending on which sentence you quote**:

| Sentence | Grade | Why |
|---|---|---|
| MRI of the authors' own patient | `HUMAN_CLINICAL` | primary observation |
| microcephaly, restating Khan et al. | `OTHER` | secondary restatement |
| domain–phenotype summary of 8 patients | `OTHER` + `PARTIAL` | summary, and explicitly hedged |

That's the same treatment this entry already gives PMID:33805425, so it's consistent — but it's now the second paper in one entry where provenance varies sentence-by-sentence. Worth noting as a pattern rather than a one-off.

Cerebellar atrophy is `PARTIAL` for a different reason: the quotable *sentence* names cerebral white matter; the cerebellar involvement lives in Table 1, which I can't quote as prose. The 5/8 count is in the description, attributed.

## The domain sentence became a hypothesis group, not just a phenotype

You were right that it's worth more than one phenotype, and I think it's worth more than a `notes` paragraph too. It's now a fourth `mechanistic_hypotheses` group:

- the three existing groups distinguish **how** a variant lowers EPRS1 function (m6A masking, aggregation, general loss)
- this one distinguishes **which of the enzyme's two functions** is lost — EPRS1 is bifunctional, and ProRS-domain vs GluRS-domain variants track with different clinical pictures

That's genuinely orthogonal to the existing axis, which is why it earned its own group rather than being folded into one. `EMERGING`, evidence `PARTIAL` — the source says "may be associated" and "the number of reported cases is limited", and eight patients is not a genotype–phenotype rule. Also tagged the p.Pro1482Thr variant with its domain, since that one is placeable.

Thanks for the non-breaking-hyphen warning — `Zn2+‑binding–domain` has both a U+2011 and an en dash. Extracted programmatically, as with the C2 statistics.

## Also on this branch since your review

I found and fixed a dangling-edge break in a *sibling* PR and swept all five entries in this run. This entry came back clean:

```
Hypomyelinating_Leukodystrophy_15.yaml   ok; roots=['Biallelic EPRS1 Variants']
```

The underlying gap — bare `downstream[].target` values are skipped by `check_entity_refs`, so a renamed node severs the graph while every gate passes — is filed as https://github.com/monarch-initiative/dismech/issues/9697. Relevant here because the debundling you asked for in item 3 was exactly the kind of edit that triggers it; it just happened not to break this one.

**On GeneReviews:** I could reach PubMed, and I ran it — but the query in the review skill is broken. `GeneReviews[TI]` searches the *chapter* title, and chapters carry "GeneReviews" as the *book* title, so it returns zero for everything. With `"GeneReviews"[Book]`: EPRS1 is genuinely 0, but TRPV4 (a control) returns 3. Your conclusion here was right; the method couldn't have told you either way. Filed as https://github.com/monarch-initiative/dismech/issues/9694.

44/44 snippets, all gates, no new cache rows. Re-requesting review.

---

PR #9671, opened 2026-08-27
curate(AIHL): Add Aminoglycoside-Induced Hearing Loss (MT-RNR1)

### LINKED ISSUE #9653: Curate deafness, aminoglycoside-induced (MONDO:0010799)
Claiming `MONDO:0010799` — aminoglycoside-induced deafness. Nothing in `kb/` covers it.

Stub: `stubs/Deafness_Aminoglycoside-induced.yaml`

**Why it's worth curating.** This is a gene-environment disease, not a purely genetic one: the m.1555A>G variant in the mitochondrial 12S rRNA (`MT-RNR1`) makes the human mitoribosomal decoding site more closely resemble the bacterial 16S target that aminoglycosides are designed to bind. Carriers are often normal-hearing until exposed. That makes it a good test of the `environmental` → `influences_mechanisms` modelling, where the drug exposure is a first-class pathograph node rather than a footnote, and of `PROTECTS_AGAINST` for avoidance.

One thing to sort out during curation: the stub carries `TRMU` (hgnc:25481) as MONDO's causal gene, which is the nuclear modifier rather than the primary locus — `MT-RNR1` is mitochondrial and may be missing from MONDO's `RO:0004003` set. Both belong in `genetic:`, with the modifier typed as such.

caufieldjh:
Adds `kb/disorders/Aminoglycoside-Induced_Hearing_Loss.yaml` (MONDO:0010799).

## The exposure is a node, not a footnote

A maternally inherited variant in the mitochondrial 12S rRNA reshapes the mitoribosomal decoding site so that it resembles the bacterial 16S site aminoglycosides were designed to bind. Aminoglycosides are selective antibacterials *only* because the human mitoribosome normally differs at that position — the variant erases the difference.

```
MT-RNR1 12S rRNA Decoding-Site Variant
  → Bacterial-Like Conformation of the Mitoribosomal Decoding Site
      → Aminoglycoside Binding to the Mitochondrial Ribosome   ← environmental: TRIGGERS
          → Impaired Mitochondrial Translation
              → Cochlear Hair Cell Mitochondrial Injury and Oxidative Stress
                  → Cochlear Hair Cell Death
                      → Irreversible Bilateral Sensorineural Hearing Loss
```

The `environmental` entry links into the binding node with `influences_mechanisms: TRIGGERS`, so the second hit is part of the graph. Nothing downstream of that node happens in an unexposed carrier — which is the disease's whole clinical shape, since the genotype is common (~1 in 385), the exposure is routine, and the outcome is entirely preventable if the two are kept apart.

## Prescribing-actionable, not just mechanistic

CPIC published a guideline for exactly this gene-drug pair, and it is cited here for the recommendation itself — *including* the "unless the high risk of permanent hearing loss is outweighed by the severity of infection" clause, which keeps it a risk-benefit judgement rather than an absolute bar — and for the instruction that avoidance be cascaded to the matriline, which is what makes cascade testing preventive rather than bookkeeping.

All three CPIC-actionable variants are curated as structured `Genetic.variants` records with CPIC evidence: `m.1555A>G`, `m.1494C>T`, and `m.1095T>C`. An entry that exists to tell a prescriber which genotypes to test for has to carry the complete actionable list, and has to carry it somewhere queryable rather than in free-text notes.

## The penetrance conflict, and why it points the opposite way from usual

Pedigree series give 14% penetrance *excluding* aminoglycoside exposure, which reads as evidence the variant is independently pathogenic. An unbiased population cohort found the opposite: 19 carriers genotyped out of the British 1958 birth cohort had hearing indistinguishable from non-carriers at 44–45.

Ascertainment is the simplest explanation — pedigrees come to attention *because* someone in them is deaf. Curated as `INTERPRETATION` rather than a contradiction, because both can hold.

The consequence runs against the usual reading of "low penetrance": if carriers keep normal hearing into middle age unless exposed, the case for genotyping *before* prescribing gets stronger, not weaker. The cohort authors draw the same inference.

## Why there is no animal model

A `HUMAN_MODEL_MISMATCH` discussion, because the empty section is a finding rather than an omission. The obstacle is intrinsic to the lesion: it makes a human mitochondrial rRNA resemble a *bacterial* one, and species diverge at exactly that sequence, so an animal ribosome carrying the human substitution is not obviously the same object. Stated as "the research sweep did not surface a widely used knock-in line" rather than "none exists" — the second is a claim I have not established.

The consequence for reading the entry: every step downstream of drug binding rests on wild-type animal or in-vitro work plus human observation, and none of it on an animal carrying the variant.

## A preflight warning that is a false positive, checked rather than assumed

`preflight-dr` returned WARN: MT-RNR1 mentioned 45 times against the MONDO canonical gene TRMU's 3. MONDO records TRMU under `RO:0004003`, but OMIM 580000 — the entity MONDO maps to — is the MT-RNR1 locus, and the OMIM identifier **matches between MONDO and the report**. TRMU is the nuclear modifier. Curated MT-RNR1 `CAUSATIVE`, TRMU `MODIFIER`, with the discrepancy recorded in `notes` rather than papered over.

## Ontology corrections

Nine of the nineteen suggested CURIEs were wrong or obsolete:

| Report said | Reality |
|---|---|
| `CHEBI:9334` streptomycin | **sulfasalazine** |
| `CHEBI:41000` aminoglycoside antibiotic | **obsolete** |
| `CL:0000598` inner hair cell | **pyramidal neuron** |
| `CL:0000211` vestibular hair cell | **electrically active cell** |
| `UBERON:0005988` organ of Corti | **atrium myocardial trabecula** |
| `UBERON:0009663` spiral ganglion | **telencephalic nucleus** |
| `UBERON:0001824` vestibular organ | **mucosa of larynx** |
| `UBERON:0001846` cochlea | **internal ear** |
| `GO:0070997` neuron death | **obsolete** |

## One term deliberately left unbound

`ECTO:9002231` "exposure to aminoglycoside antibiotic" is exact and resolves in OLS, but is **absent from the local `sqlite:obo:ecto` build** term validation uses. Every antibiotic-exposure term that build carries is agrochemical — pesticide, insecticide, acaricide, nematicide. No term beats a bad one, so the exposure keeps a free-text `preferred_term`; `environmental-term-audit` will class it `PARTIAL`. Refreshing ECTO would let it bind as-is.

Vestibulotoxicity is likewise deliberately out of scope, being drug- and dose-dependent rather than MT-RNR1-dependent, and `notes` says so.

## Validation (current head)

| Check | Result |
|---|---|
| `just validate` | pass |
| Snippets verified | **27/27** |
| Raw-substring re-check bypassing validator normalization | 27/27, 0 non-verbatim |
| Reference titles vs cache frontmatter | 34/34 |
| Pathograph | 1 component, 7 nodes, no orphans, 4 join points |

GeneReviews `PMID:20301595` tagged and mined across all four sections.

Closes #9653

---


caufieldjh:
Both blocking items fixed in `23df93d`, plus two of the suggestions.

**The CPIC guideline.** You're right, and the part worth owning is *how* I missed it: the deep-research artifact's own "Key Caveats for Curators" says "Cite the primary CPIC publication," and it was the one explicit curation instruction in that section I didn't follow. I verified `PMID:34032273` myself before using it — full text, and it names all three variants in a single sentence.

It now carries two quotes on the avoidance treatment: the prescribing recommendation including the risk-benefit exception that keeps it from reading as an absolute bar, and the sentence extending avoidance to the whole matriline — which is what makes cascade testing a preventive act rather than bookkeeping.

**`m.1095T>C` and the `variants` slot.** Also right, and the two halves compound as you said. An entry whose purpose is telling a prescriber which genotypes to test for was missing one of the three actionable ones, and everything it *did* know lived in free-text `notes` where no query could reach it. All three are now structured `variants` entries with CPIC evidence, following the `ACys_Amyloidosis` pattern you pointed at.

**Two suggestions taken:**

- **`HUMAN_MODEL_MISMATCH`.** Added, and it does explain the empty `animal_models`. The obstacle is intrinsic: the lesion makes a human mitochondrial rRNA resemble a *bacterial* one, and species diverge at exactly that sequence, so an animal ribosome carrying the human substitution isn't obviously the same object. I phrased it as "the research sweep did not surface a widely used knock-in line" rather than "none exists" — the second is a negative claim I haven't independently established, and I've been wrong on exactly that kind of assertion before.
- **Vestibulotoxicity.** Added the half-sentence to `notes`. Your reading was the entry's intent — it's drug- and dose-dependent, so it belongs to the pharmacological toxicity the scope note excludes — and it should have said so.

**Not taken, with reasons:** tinnitus stays out, since the research gives no quantified frequency and I'd rather not assert one. The PALOH `clinical_trials` entry and the Buryat founder-frequency record are both real gaps, but each needs a source I haven't fetched and verified, so they'd be new curation rather than a fix to this diff.

Thanks also for checking `NCIT:C157820` against the `treatment_term` enum root — I'd reached the same conclusion about Supportive Care plus `therapeutic_modality: DEVICE`, but seeing the reachability spelled out is useful for the next entry that hits it.

**Validation on `23df93d`:** 27/27 snippets verified, re-checked as raw substrings with 0 non-verbatim. Entity refs, duplicate keys clean. Reference titles 34/34 against cache frontmatter. Pathograph unchanged at 7 nodes, one component. Both history records validate.

---

PR #9672, opened 2026-08-27
curate(PBD3A): Add Peroxisome Biogenesis Disorder 3A (PEX12)

### LINKED ISSUE #9652: Curate peroxisome biogenesis disorder 3A (Zellweger) (MONDO:0013927)
Claiming `MONDO:0013927` — peroxisome biogenesis disorder 3A (Zellweger), the PEX12-related severe end of the Zellweger spectrum.

Stub: `stubs/Peroxisome_Biogenesis_Disorder_3A_Zellweger.yaml`

**Why this is a disease entry.** The KB already curates `Peroxisome_Biogenesis_Disorder_1B` (PEX1) and `Peroxisome_Biogenesis_Disorder_4B` (PEX6) as standalone entries alongside the umbrella `Peroxisome_Biogenesis_Disorder`, so a per-PEX-gene entry follows the established pattern. PEX12 is not covered anywhere in `kb/`.

PBD3A is the severe "A" end of its gene's spectrum rather than the milder "B" end the two existing entries cover, so it should also make the severity-tier contrast explicit against them.

caufieldjh:
Adds `kb/disorders/Peroxisome_Biogenesis_Disorder_3A_Zellweger.yaml` (MONDO:0013927).

## One import failure, two downstream arms

PEX12 is a RING-finger subunit of the retrotranslocation channel that ubiquitinates PEX5 and pulls it back out of the peroxisomal membrane. Lose it and the receptor cannot recycle, so matrix protein import collapses wholesale — every PTS1 cargo at once, not one enzyme. The pathograph splits there into the two arms that produce clinically different things:

```
Biallelic PEX12 Loss of Function
  → Disrupted PEX2-PEX10-PEX12 Retrotranslocation Channel
      → Failure of PEX5 Receptor Recycling
          → Collapse of Peroxisomal Matrix Protein Import
              ├→ Accumulation of Very-Long-Chain and Branched-Chain Fatty Acids
              │     → Progressive Hepatic Injury
              │     → Adrenocortical Insufficiency
              └→ Plasmalogen Deficiency
                    → Impaired Neuronal Migration in the Developing Cortex
                          → Severe Neurodevelopmental Impairment
```

The accumulation arm is what a *catabolic* block does; the deficiency arm is what an *anabolic* block does. Both come from the same import collapse, which is why a single missing subunit gives a multisystem disease rather than an organ-specific one.

## Biochemistry as readouts, not prose

Three markers in `biochemical:` with `readouts` wired into the pathograph rather than described in text:

| Marker | Direction | Reports on |
|---|---|---|
| Very-long-chain fatty acids (`CHEBI:27283`) | Increased | the beta-oxidation block |
| C26:0-lysophosphatidylcholine | Increased | the same block, in dried blood spots |
| Plasmalogens (`CHEBI:64611`) | Decreased | the biosynthetic arm |

C26:0-lysoPC is the one that makes newborn screening technically feasible, which matters most for the severe end of the spectrum because that is where the window for anything at all is shortest. Curated with the sensitivity figure (86/91 ZSD dried blood spots above the 72 nmol/L cut-off) rather than a qualitative "elevated".

## Cholic acid is a treatment, not supportive care

Earlier this was a clause inside supportive management. It is a standalone `treatment` now, bound to `CHEBI:16359`, with `target_mechanisms` on both the VLCFA/bile-acid-intermediate accumulation node and the hepatic injury node — because it acts by feedback-suppressing the synthesis of the toxic C27 intermediates, which is a mechanism the graph can carry. Adrenal replacement and biochemical/developmental surveillance are likewise linked to the nodes they act on.

## GeneReviews, mined and then re-mined

`PMID:20301621` is tagged `GeneReviews` and cited 21 times. An earlier version of this entry asserted in both the PR body and the history record that all four sections were mined while Genetic Counseling was in fact uncited — review caught it, and the section is now mined. That failure is recorded in the entry's own `notes` rather than quietly fixed.

The chapter is written for the Zellweger spectrum as a continuum, not for PEX12 and not for the severe end specifically. Every evidence item quoting it says which end of the spectrum the quoted sentence describes, and several phenotypes are graded `PARTIAL` because the supporting sentence is about the intermediate/milder end while this entry is the severe one.

## Phenotype corrections from review

- `HP:0002240` Hepatomegaly was carrying a cholestasis-and-fibrosis claim it does not state; replaced with `HP:0001410` (Decreased liver function).
- Polymicrogyria, gray matter heterotopia, cataract, and feeding difficulties were described in the sources and in the entry's prose but not curated as phenotypes. They are now, 13 in total.
- Polymicrogyria and heterotopia were then found to be graded inconsistently *against the same quoted sentence* — one `SUPPORT`, one `PARTIAL`; and afterwards left at different frequencies on adjacent clauses of that sentence. Both are aligned now, and the regrade is logged in the history record rather than being silent.

## Ontology corrections

Ten of the suggested CURIEs were wrong. The one that mattered most:

| Report said | Reality |
|---|---|
| `GO:0033539` peroxisomal beta-oxidation | the **mitochondrial** acyl-CoA dehydrogenase process — would have inverted the mechanism; correct term is `GO:0033540` |
| `CL:0000601` radial glial cell | cochlear outer hair cell |
| `CL:0001133` | does not resolve |
| `HP:0002656` chondrodysplasia punctata | epiphyseal dysplasia |
| `HP:0000121` nephrolithiasis | nephrocalcinosis |
| `HP:0002187` hypomyelination | profound intellectual disability |
| `HP:0012825` | the modifier "Mild" |
| `CHEBI:143092`, `CHEBI:28865`, `CHEBI:36620` | all wrong for the offered concepts |

## Animal model, with the caveat named

The Pex2-null mouse is curated because no PEX12 mouse is reported in the sources used here. `limitations` names the paralogue explicitly — it is a different RING subunit of the same channel, so it models the import collapse but not PEX12 itself.

## Discussions

- `INTERPRETATION` on internal AUG re-initiation, the mechanism that lets some C-terminal truncating alleles produce a shorter functional product and so sit at the milder end of the spectrum.
- `KNOWLEDGE_GAP` on the absence of any disease-modifying therapy.

## Validation (current head)

| Check | Result |
|---|---|
| `just validate` | pass |
| Snippets verified | **36/36** |
| Raw-substring re-check bypassing validator normalization | 36/36, 0 non-verbatim |
| Pathograph | 1 component, 10 nodes, no orphans |

Closes #9652

---

### DIFF
--- a/history/disorders/Peroxisome_Biogenesis_Disorder_3A_Zellweger/2026-08-27T155053Z-claude-code-c5e31a.yaml
+++ b/history/disorders/Peroxisome_Biogenesis_Disorder_3A_Zellweger/2026-08-27T155053Z-claude-code-c5e31a.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Peroxisome_Biogenesis_Disorder_3A_Zellweger
+  path: kb/disorders/Peroxisome_Biogenesis_Disorder_3A_Zellweger.yaml
+session:
+  id: 2026-08-27T155053Z-claude-code-c5e31a
+  timestamp: '2026-08-27T15:50:53Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9652
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9672
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - biochemical
+  summary: Regrade migration phenotypes onto specific neuropathology quotes
+  details: |
+    Second review round, non-blocking items. Polymicrogyria had been graded SUPPORT on a GeneReviews quote that never names it, while Gray Matter Heterotopia two entries below graded the same quote PARTIAL for exactly that reason - one snippet carrying two strengths for the same kind of claim. Both now cite the specific human neuropathology sentences in PMID:9382874 naming lateral polymicrogyria and subcortical heterotopic neurons directly, so both are SUPPORT on evidence that names them, with the GeneReviews quote kept as a PARTIAL companion placing the findings among the congenital malformations. Both are graded OTHER rather than MODEL_ORGANISM: the sentences are introduction material reciting prior human Zellweger autopsy series in a paper whose own results are in mice, so the file now grades that PMID per sentence - OTHER for the human pathology it cites, MODEL_ORGANISM for the authors' own mouse results. Bound the plasmalogen biomarker to CHEBI:64611 ether lipid. A follow-up review then pointed out that heterotopia was still frequency OCCASIONAL while polymicrogyria was FREQUENT, although the two are described in adjacent clauses of one sentence as a single co-occurring picture; heterotopia is now FREQUENT with the reason stated. 36/36 snippets verified and independently re-checked as raw substrings.
--- a/kb/disorders/Peroxisome_Biogenesis_Disorder_3A_Zellweger.yaml
+++ b/kb/disorders/Peroxisome_Biogenesis_Disorder_3A_Zellweger.yaml
@@ -468,9 +468,11 @@ phenotypes:
       finding among the congenital malformations of the severe form.
 - category: Neurologic
   name: Gray Matter Heterotopia
-  frequency: OCCASIONAL
+  frequency: FREQUENT
   description: >-
-    Ectopic grey matter, the other structural signature of failed migration.
+    Ectopic grey matter, the other structural signature of failed migration. Described
+    together with the gyral abnormalities as one co-occurring picture rather than as an
+    independent finding, which is why it carries the same frequency as the polymicrogyria.
   phenotype_term:
     preferred_term: Gray matter heterotopia
     term:


caufieldjh:
All six blocking items fixed in `7fc0b42`. Taking item 3 first, because it's the one that matters most.

**GeneReviews Genetic Counseling — you caught a false claim.** The PR body *and* the history record both said GeneReviews was "mined across all four sections." It was mined across three. I asserted the completeness rather than checking it, which is the same failure mode that produced three unverified negative claims about sources earlier in this session. The recurrence-risk sentence is now cited on `inheritance`, the CREATE record's `details` names the three sections actually mined at creation, and `notes` records that the earlier claim was wrong rather than quietly restating it as true.

**`biochemical:` section.** Added — VLCFA (`CHEBI:27283`), C26:0-lysoPC, and plasmalogens. Your diagnosis of the underlying problem was right: the biomarker claim existed, but as prose inside `diagnosis[0].description` where nothing can query it. Readouts point at both the accumulation *and* deficiency nodes, so the section captures the thing that actually distinguishes a biogenesis defect — the signature moves in two directions at once.

**Cholic acid.** Now a standalone treatment with `CHEBI:16359` and `target_mechanisms` on the VLCFA-accumulation and hepatic-injury nodes. Your framing was the convincing part: it's the one intervention here that acts on a curated mechanism node rather than a consequence, so burying it in the supportive-care blob lost the entry's only real treatment-to-pathograph join. Treatment join points 3 → 5.

**`HP:0002240`.** Rebound to `HP:0001410 Decreased liver function`. You're right that neither the description nor the snippet asserts an enlarged liver — hepatomegaly was a specific physical finding standing in for hepatic dysfunction.

**Polymicrogyria, heterotopia, cataract.** All curated, plus feeding difficulties. The cataract point is the one I'd have wanted flagged even if it weren't blocking: the entry was citing "cataract removal" as a treatment for a finding it didn't record. Heterotopia is graded `PARTIAL` because GeneReviews names the migration defect generically rather than heterotopia specifically.

**Suggestions not taken, with reasons:** the adrenal-replacement rebinding to `NCIT:C15986` plus a hydrocortisone agent, DHA as a negative treatment record, `genetic_context` on the PEX12 node, a narrower lab term, the ClinGen `CGGV:` assertion, and the adrenal node's `biological_scale`. Each is reasonable and several are genuine improvements, but each needs either a source I haven't fetched and verified or a judgement I'd rather not make in a fix-up push. Happy to take any of them if you'd rather they land here than in a follow-up.

One thing I broke and fixed in passing: my first attempt at the biochemical section used a 6-space continuation indent on a wrapped `reference_title` at a nesting level that needs 8, which made the file unparsable. Caught before pushing.

**Validation on `7fc0b42`:** 34/34 snippets verified, re-checked as raw substrings with 0 non-verbatim. Entity refs, duplicate keys clean. Reference titles 40/40. Pathograph 10 nodes, one component, no orphans. Both new reference caches were already tracked in `main`, so nothing was added there.

---

caufieldjh:
Thanks for the approval — and I've spent it, on `1688ea2`, for one of your non-blocking items. Flagging why, since it costs a re-review cycle.

**The polymicrogyria grading was a defect, not polish.** You put it gently as a suggestion, but the substance is that I graded the *same snippet* `SUPPORT` on one phenotype and `PARTIAL` on another two entries below, for a difference that doesn't exist — the GeneReviews quote names neither polymicrogyria nor heterotopia. One quote can't honestly carry two strengths for the same kind of claim, and the `PARTIAL` one was the honest reading, which makes the `SUPPORT` one wrong.

You were also right that a better sentence was already sitting in this PR's own cache. `PMID:9382874` names both findings directly:

> "…gyral abnormalities centered around the Sylvian fissure with a stereotypic medial pachygyria and lateral polymicrogyria."
> "These gyral abnormalities reflect a reduced neuronal population in the cortex and large numbers of subcortical heterotopic neurons."

Both phenotypes now cite those, so both can be `SUPPORT` on evidence that actually names them. The GeneReviews quote stays on polymicrogyria, regraded `PARTIAL`, to place it among the congenital malformations.

One grading call worth stating: both are `OTHER`, not `MODEL_ORGANISM`. The sentences are introduction material summarising prior *human* Zellweger neuropathology — citing Volpe and Adams, Evrard, and Powers on Zellweger fetuses — in a paper whose own results are in mice. Grading them `MODEL_ORGANISM` would attribute human autopsy findings to a mouse.

Also bound the plasmalogen marker to `CHEBI:64611 ether lipid`, matching PBD1B.

**Not taken:** the top-level `evidence:` shape difference on `biochemical[]` entries, and the six items carried forward from your first review. Those are genuine improvements but need either sources I haven't fetched or judgements I'd rather not fold into a fix-up push. On the plasmalogen readout resting only on the `Pex2` mouse — I agree with your check, and I looked too; no cached reference carries a human plasmalogen measurement, so the `PARTIAL` grading with its stated reason is the honest position rather than a gap.

**Validation on `1688ea2`:** 36/36 snippets verified, re-checked raw with 0 non-verbatim. `check-snippet-grading` clean, no new divergences. Entity refs and duplicate keys clean. Reference titles 42/42.

---

caufieldjh:
Both non-blocking items taken in `424c566`, and I'd rather flag than bank this one because the first is an inconsistency the *previous* commit introduced.

**The frequency mismatch.** You're right. The regrade grounded both migration phenotypes in adjacent clauses of one sentence describing a single co-occurring picture — and then left heterotopia at `OCCASIONAL` while polymicrogyria sat at `FREQUENT`. The heterotopia clause says "large numbers of subcortical heterotopic neurons," in a malformation the same paragraph calls reproducible. `OCCASIONAL` was a leftover from when that entry had only the generic `PARTIAL` evidence, and I didn't revisit it when I changed what it rested on. Now `FREQUENT`, with the reason in the description.

That's the pattern this round keeps producing: I fix a thing and leave the field *next* to it describing the old state. Three times now — the `notes` subtype-tag claim on #9673, the GeneReviews "all four sections" line on this PR, and now this frequency.

**The history record.** Added. You're right that CLAUDE.md's rule is per-PR and this wasn't required, but a commit that moves the evidence basis of two phenotypes onto a different source is exactly what the log is for. All three records validate.

Thanks for spelling out the `OTHER` reasoning independently — `MODEL_ORGANISM` would attribute human autopsy findings to a mouse, `HUMAN_CLINICAL` would attribute someone else's data to this paper, and neither is what that paragraph is. Useful to have that written down where the next person hits it.

**Not taken**, all carried forward: adrenal replacement modality/agent, the DHA negative trial, `genetic_context` on the PEX12 node, the generic `NCIT:C25294`, a ClinGen `CGGV:` citation, the adrenal node's scale, and the top-level `evidence:` shape on `biochemical[]`. Each needs a source or judgement beyond this diff. I'll also leave the long folded line in `notes` — it renders correctly and rewrapping it would churn the diff for no gain.

**Validation on `424c566`:** 36/36 snippets verified. Entity refs, duplicate keys clean. All three history records validate.

This is my last push on this PR unless something blocking turns up.

---

PR #9673, opened 2026-08-27
curate(cblD): Add MMADHC-related cobalamin disorder, cblD type

### LINKED ISSUE #9651: Curate homocystinuria-megaloblastic anemia cblD type (MONDO:0700297)
Claiming `MONDO:0700297` — homocystinuria-megaloblastic anemia cblD type (MMADHC).

Stub: `stubs/Homocystinuria-megaloblastic_Anemia_cblD_Type.yaml`

**Why this is a disease entry rather than a subtype.** The KB already curates the cblC complementation group as its own entry (`MMACHC-related_Methylmalonic_Aciduria_and_Homocystinuria_cblC_Type`, MONDO:0010184), so per-complementation-group entries are the established pattern here. MMADHC is currently named only in passing inside `Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport`.

The mechanistically interesting part is that cblD is not one biochemical phenotype but three, and which one a patient gets depends on where in MMADHC the variant falls — isolated methylmalonic aciduria, isolated homocystinuria, or the combined form. That branch is the entry's main claim to curate.

caufieldjh:
Adds `kb/disorders/MMADHC-related_Disorder_of_Cobalamin_Metabolism_cblD_Type.yaml` (MONDO:0700297), and resolves two stubs.

## One gene, three biochemical diseases

MMADHC sits at the branch point where intracellular cobalamin is committed either to adenosylcobalamin (mitochondrial, feeding methylmalonyl-CoA mutase) or to methylcobalamin (cytosolic, feeding methionine synthase). Which arm a patient loses depends on *where in the protein* the lesion falls — so a single gene produces three distinct biochemical diseases:

```
Biallelic MMADHC Loss of Function
  → Failure of Cobalamin Branch-Point Trafficking
      ├→ Adenosylcobalamin Deficiency
      │     → Methylmalonic Acid Accumulation
      └→ Methylcobalamin Deficiency
            → Homocysteine Accumulation and Methionine Deficit
                  → Impaired One-Carbon Metabolism in Erythroid Precursors
                        → Megaloblastic Anemia
                  → Neurologic Injury
```

The three forms are curated as `has_subtypes`: `cblD-MMA` (isolated methylmalonic aciduria, N-terminal alleles rescued by internal AUG re-initiation), `cblD-HC` (isolated homocystinuria, C-terminal missense), and `cblD-MMA/HC` (the classic combined form). The genotype-to-arm mapping lives in the subtype descriptions.

## Why the biochemical phenotypes carry no subtype tag

An earlier version tagged homocystinuria to `cblD-HC` and methylmalonic aciduria to `cblD-MMA`, which left `cblD-MMA/HC` with zero phenotypes — inverting the entry's central claim, since the combined form has *both*. The `subtype` slot is single-valued, so there is no tag that says "this and the combined form". They are untagged now, and `notes` says why rather than describing the old state.

## Biochemistry as readouts

Three markers in `biochemical:`, each wired to the node it reports on:

| Marker | Direction | Reports on |
|---|---|---|
| Total plasma homocysteine (`CHEBI:17230`) | Increased | the methylcobalamin arm |
| Methylmalonic acid (`CHEBI:30860`) | Increased | the adenosylcobalamin arm |
| Plasma methionine (`CHEBI:16811`) | Decreased | the same methylcobalamin arm, from the other side |

Methionine is the one that distinguishes a remethylation defect from a cystathionine beta-synthase defect at the bedside: both raise homocysteine, but only the remethylation block takes methionine *down*.

## Treatments carry their agents

- Parenteral hydroxocobalamin — `therapeutic_agent` `CHEBI:27786`, targeting the trafficking failure by mass action.
- Betaine — `therapeutic_agent` `CHEBI:17750`, providing the alternate BHMT remethylation route that bypasses the methionine synthase block.
- Avoidance of catabolic stress and nitrous oxide — no agent; nitrous oxide irreversibly oxidises the cobalt of methylcobalamin, so it is a specific hazard rather than general caution.

## A snippet that stated no finding

Review caught `"Finally, C-terminal truncation of more than 20 amino acids"` — a fragment cut mid-sentence that asserts nothing. It carries the full clause now. This is the failure mode `check-snippet-length` is a floor against but cannot catch, since the fragment is over five words.

## The preflight WARN is real, not a false positive

`preflight-dr` flagged MMACHC (cblC) at 25% of MMADHC mentions. That is a genuine warning: cblC is the adjacent step in the same pathway, the prototype of the group, and far more common. Report sections describing cblC were used only for pathway context and the differential; no cblC-specific finding is cited as a cblD finding.

The same applies to GeneReviews `PMID:20301503`, which covers the whole family of intracellular cobalamin disorders and is explicit that cblC is its worked prototype. Phenotype items quoting its clinical description are graded `PARTIAL` for that reason — the sentence describes cblC and is cited for the shared mechanism, not as a cblD observation.

## Corrections

- The research report gave `HGNC:26275` for MMADHC. That is wrong; the gene is `hgnc:25221`.
- `HP:0003201` was silently avoided in an earlier draft rather than being checked — it is Rhabdomyolysis, and review was right that the avoidance should have been reasoned rather than tacit.
- Curation mechanics had leaked into a `description:` field, which renders on the public page. Moved to `notes`.

## Discussions

- `INTERPRETATION` on why one gene gives three biochemical diseases (the internal-AUG re-initiation mechanism that rescues the mitochondrial arm from N-terminal alleles).
- `KNOWLEDGE_GAP` on premature-termination-codon readthrough as an untested route to restoring MMADHC.

## Stubs resolved

Two stubs are resolved as curation decisions rather than curated entries, per the queue's own rules:

- `stubs/Acid_Sphingomyelinase_Deficiency.yaml` → `entry_type: GROUPING` — it is the umbrella for the SMPD1 spectrum, whose members (Niemann-Pick A, B, E, and chronic neurovisceral ASMD) the KB already curates separately.
- `stubs/Charcot-Marie-Tooth_Disease_Type_4D.yaml` → `entry_type: SUBTYPE` — `kb/disorders/Charcot-Marie-Tooth_Disease_Type_4.yaml` already carries a `CMT4D` entry in `has_subtypes` naming NDRG1, so a separate disorder entry would duplicate it. Flagged on issue #9680, which claims this disease, since the claimant should get to weigh that.

## Validation (current head)

| Check | Result |
|---|---|
| `just validate` | pass |
| Snippets verified | **24/24** |
| Raw-substring re-check bypassing validator normalization | 24/24, 0 non-verbatim |
| `check-stubs` / `validate-stubs` | pass |
| Pathograph | 1 component, 9 nodes, no orphans |

Closes #9651

---

### DIFF
--- a/kb/disorders/MMADHC-related_Disorder_of_Cobalamin_Metabolism_cblD_Type.yaml
+++ b/kb/disorders/MMADHC-related_Disorder_of_Cobalamin_Metabolism_cblD_Type.yaml
@@ -217,9 +217,8 @@ phenotypes:
   name: Megaloblastic Anemia
   frequency: FREQUENT
   description: >-
-    Macrocytic anemia with megaloblastic marrow. Deliberately untagged by subtype: it
-    occurs wherever remethylation fails, which is both cblD-HC and the combined
-    cblD-MMA/HC form, and the subtype slot takes a single value.
+    Macrocytic anemia with megaloblastic marrow. Occurs wherever remethylation fails,
+    which is both the isolated homocystinuria form and the combined form.
   phenotype_term:
     preferred_term: Megaloblastic anemia
     term:
@@ -238,9 +237,8 @@ phenotypes:
   name: Homocystinuria
   frequency: VERY_FREQUENT
   description: >-
-    Elevated total plasma homocysteine, the biochemical marker of the cytosolic arm.
-    Untagged by subtype because it defines cblD-HC and is also present in the combined
-    cblD-MMA/HC form.
+    Elevated total plasma homocysteine, the biochemical marker of the cytosolic arm. It
+    defines the isolated homocystinuria form and is also present in the combined form.
   phenotype_term:
     preferred_term: Homocystinuria
     term:
@@ -252,16 +250,17 @@ phenotypes:
       methylcobalamin and adenosylcobalamin synthesis.
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
-    snippet: Two patients presented with isolated homocystinuria and reduced formation
-      of methionine and methylcobalamin in cultured fibroblasts, defined as cblD-variant 1
-    explanation: Documents the isolated-homocystinuria form and its cellular correlate.
+    snippet: Two patients presented with isolated homocystinuria
+    explanation: The clinical half of the sentence. Split from the fibroblast measurement
+      in the same sentence, which is quoted separately under biochemical with an IN_VITRO
+      grading, because one sentence cannot carry two evidence_source values.
 - category: Metabolic
   name: Methylmalonic Aciduria
   frequency: VERY_FREQUENT
   description: >-
-    Elevated methylmalonic acid, the biochemical marker of the mitochondrial arm.
-    Untagged by subtype because it defines cblD-MMA and is also present in the combined
-    cblD-MMA/HC form.
+    Elevated methylmalonic acid, the biochemical marker of the mitochondrial arm. It
+    defines the isolated methylmalonic aciduria form and is also present in the combined
+    form.
   phenotype_term:
     preferred_term: Methylmalonic aciduria
     term:
@@ -320,8 +319,7 @@ phenotypes:
   name: Seizures
   frequency: FREQUENT
   description: >-
-    Seizures, asserted by this entry's own Neurologic Injury node and recorded here so the
-    phenotype section matches the mechanism description.
+    Seizures, part of the neurologic presentation of this group of disorders.
   phenotype_term:
     preferred_term: Seizure
     term:
@@ -435,14 +433,22 @@ biochemical:
     interpretation: Low methionine alongside high homocysteine localises the block to
       remethylation rather than transsulfuration.
     evidence:
+    - reference: PMID:15292234
+      reference_title: The cblD defect causes either isolated or combined deficiency of
+        methylcobalamin and adenosylcobalamin synthesis.
+      supports: SUPPORT
+      evidence_source: IN_VITRO
+      snippet: reduced formation of methionine and methylcobalamin in cultured fibroblasts
+      explanation: A cblD observation rather than an inference from the cblC prototype -
+        methionine formation is directly reduced in patient fibroblasts.
     - reference: PMID:20301503
       reference_title: Disorders of Intracellular Cobalamin Metabolism.
       supports: PARTIAL
       evidence_source: OTHER
       snippet: maintaining plasma methionine concentrations within the normal range
       explanation: PARTIAL - GeneReviews states methionine as a treatment target, which
-        implies it runs low, rather than stating the diagnostic contrast with CBS
-        deficiency directly.
+        implies it runs low; kept as the clinical-management counterpart to the fibroblast
+        result above.
 diagnosis:
 - name: Plasma Total Homocysteine and Methylmalonic Acid
   description: >-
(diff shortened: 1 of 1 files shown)


caufieldjh:
All four blocking items fixed in `7e4cba2`. Two of them were defects rather than gaps, so taking those first.

**The truncated snippet.** You're right, and this is the one that bothers me. The quote read *"Finally, C-terminal truncation of more than 20 amino acids"* and stopped exactly where the finding begins — it stated no result at all, while being the only sourced support for the graded-truncation claim that both `genetic.notes` and the PR body's headline argument lean on. It passed `check-snippet-length` at nine words and fails the substantive test, which is a good illustration that the word-count gate is a floor and not the standard. Now quotes the full sentence.

**The subtype tagging.** Also right, and worse than cosmetic as you say. `subtype` is single-valued, and I pinned all three biochemical phenotypes to the isolated forms — so a query for the phenotypes of `cblD-MMA/HC` returned nothing, for the form that *by definition* has both and in which cblD was originally defined. That inverts the entry's central claim. I dropped the tags where the phenotype spans forms rather than duplicating entries, and each description now says which forms it spans and why the tag is absent. You also caught that the Megaloblastic Anemia description said "forms" plural while the tag said one — the prose was right and the tag was wrong.

**`biochemical:` section.** Added — homocysteine, methylmalonic acid, methionine, with readouts joining the pathograph. Methionine is the one that earns its place for the reason you gave: low-to-normal rather than elevated is what separates a remethylation defect from CBS deficiency, and the entry was losing that discrimination entirely.

**`therapeutic_agent`.** Both bound (`CHEBI:27786`, `CHEBI:17750`). The entry already follows the cblC sibling for naming; it should have followed it here too.

Also took the hypotonia and seizures phenotypes — your point that the `Neurologic Injury` node asserts seizures in prose while the phenotype section recorded none is exactly the internal inconsistency I've been fixing in the sibling PRs, and I missed it here. Both come from a sentence already quoted in this file, graded `PARTIAL` with the same cblC-prototype caveat.

Two corrections to my own claims: the `notes` line said "two phenotype items" were graded `PARTIAL` when it was counting a discussion item quoting a different paper — now stated qualitatively rather than wrongly. And the history record's `pull/0` is repointed to 9673.

**Not taken:** prevalence via `ORPHA:79283` (needs a structured-source rebuild I haven't run), the R250X late-onset detail, the readthrough gap's Ataluren escape route, folinic acid and the cyanocobalamin-ineffective negative, `experimental_models` for the fibroblast lines, and the newborn-screening gap. Several of those are genuinely good — the Ataluren point in particular, since you're right that my rationale frames the barrier as narrower than the source does. They need sources I haven't fetched and verified, so they'd be new curation rather than fixes to this diff.

Thanks for noticing the `HP:0003201` avoidance. I checked it and it is Rhabdomyolysis, but I didn't record that I'd rejected it — worth documenting the near-misses, not just the corrections I make.

**Validation on `7e4cba2`:** 23/23 snippets verified, re-checked as raw substrings with 0 non-verbatim. Entity refs, duplicate keys clean. Reference titles 28/28. Pathograph 9 nodes, one component, no orphans. Both history records validate.

---

caufieldjh:
Thanks for the approval — spent on `329dc3e`, because your first suggestion is a false claim about the file rather than a style point.

**The `notes` contradiction.** You're right, and the failure mode you named is the reason it's worth a cycle: the paragraph said "the biochemical phenotypes carry subtype tags" when the previous commit had deliberately removed every one, so a future curator could have read that as intent and *restored* the tags — reintroducing the exact inversion this round fixed. Rewritten to say the phenotypes are untagged and why, and to record that an earlier version did tag them.

That's the third false claim of my own that review has caught across these three PRs — GeneReviews "all four sections" on #9672, the "two PARTIAL" miscount here, and now this. All three were assertions about the file's own state that I never re-checked after changing it.

**The methionine evidence.** Taken, and it turned out to be more interesting than a straight upgrade. Adding the cblD-specific quote surfaced a grading divergence: the source sentence is *"Two patients presented with isolated homocystinuria and reduced formation of methionine and methylcobalamin in cultured fibroblasts"* — a clinical presentation and a fibroblast assay in one sentence — and I had it graded `HUMAN_CLINICAL` for the phenotype and `IN_VITRO` for the biomarker. `check-snippet-grading` caught it.

CLAUDE.md's instruction for a mixed-source sentence is to split it, so the phenotype now quotes the clinical half and the biomarker the fibroblast half, each with the grading that fits. Better outcome than either single grading would have been.

**The style point about `description:`.** Taken, and it's a good distinction I hadn't made. "Deliberately untagged by subtype: … the subtype slot takes a single value" is curation mechanics sitting in a field that renders on the public disorder page. The reasoning belongs in `notes`, which is where readers go for provenance; descriptions should read clinically. Moved, along with the "recorded here so the phenotype section matches" line on Seizures.

**Not taken:** prevalence via `ORPHA:79283`, the R250X late-onset association, folinic acid and the cyanocobalamin-ineffective fact, `experimental_models`, the newborn-screening gap, and the readthrough discussion's Ataluren escape route. You're right about that last one in particular — my rationale does frame the barrier as narrower than the source supports. They each need a source I haven't fetched and verified, so they're follow-up curation rather than fixes to this diff.

On `presence: Increased` vs `INCREASED` — I'll leave it, since you confirmed the slot is free-text and both forms are well represented.

**Validation on `329dc3e`:** 24/24 snippets verified, re-checked raw with 0 non-verbatim. `check-snippet-grading` and `check-snippet-length` clean, no new baseline entries. Entity refs and duplicate keys clean. Reference titles 29/29.

---

PR #9649, opened 2026-08-27
curate: Add Methylcobalamin Deficiency Type cblE (MONDO:0009354, MTRR)

cmungall:
New Mendelian entry for cblE — the MTRR-deficient form of functional methionine synthase deficiency.

The mechanistic point the entry is built around: **cblE is a functional enzyme deficiency, not a structural one.** Methionine synthase protein and its intrinsic catalytic capacity are intact. What is missing is MTRR, the flavoprotein that reductively reactivates the cob(II)alamin cofactor after it oxidises during normal turnover. Without it the enzyme is stranded in the inactive state. That is why patient fibroblast MTR activity measures normal under strong reducing conditions and only falls when the reductant is limited — which is the in vitro test that distinguishes cblE from cblG, where the apoenzyme itself is defective and activity is low under all conditions.

Seven-node pathograph: MTRR loss of function → methionine synthase trapped as cob(II)alamin → impaired homocysteine remethylation → (methyl-folate trapping, impaired erythropoiesis, homocysteine-mediated endothelial injury) → CNS white matter injury.

## Also curated

- **GeneReviews baseline** (PMID:20301503), tagged in top-level `references:`. Its *Agents/Circumstances to Avoid* section is curated as a treatment entry — nitrous oxide, prolonged fasting, methionine restriction.
- **Adult-onset hemolytic anemia** as a late complication of *treated* disease (PMID:38736634, 2024), with LDH as a monitoring biomarker and hydroxocobalamin dose escalation as the response. This is a different lesion from the presenting megaloblastic anemia.
- **A `HUMAN_MODEL_MISMATCH` discussion.** The Mtrr gene-trap mouse (PMID:18413293) reports reproductive and cardiac phenotypes — embryonic resorption, ventricular septal defects — not the megaloblastic anemia or progressive neurocognitive disease that define the human disorder. The `FAILS_TO_RECAPITULATE` link says explicitly that this is an absence of *reported* erythroid assessment rather than a documented negative marrow result, so it reads as a gap in model coverage rather than a proven negative.

## Gradings that go against the narrative reviews

Both have their reasoning in the YAML `notes`, so a future curator can see the call was deliberate:

[...]

PR #9674, opened 2026-08-27
curate: Add Mitochondrial Complex I Deficiency Nuclear Type 23 (MONDO:0032627, NDUFA12)

### LINKED ISSUE #9624: Curate mitochondrial complex I deficiency, nuclear type 23 (MONDO:0032627)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 23** ([MONDO:0032627](https://monarchinitiative.org/MONDO:0032627)) — autosomal recessive complex I deficiency caused by *NDUFA12* (`hgnc:23987`) variants, presenting as Leigh syndrome or a Leigh-like encephalopathy with optic atrophy. NDUFA12 is an accessory subunit of the NADH:ubiquinone oxidoreductase N-module, so the mechanism is impaired complex I assembly/stability with reduced NADH oxidation and secondary lactic acidosis.

MONDO puts this under `mitochondrial complex I deficiency, nuclear type` (MONDO:0100223), which is a numbered per-gene series. Check how the KB already handles the MC1DN and Leigh syndrome entries so the new entry sits consistently with them rather than duplicating a Leigh entry.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_23.yaml`

cmungall:
New Mendelian entry for NDUFA12-related complex I deficiency.

Five-node pathograph: NDUFA12 loss of function → impaired complex I Q-module assembly and stability → branching to NDUFAF2 compensatory association and to reduced NADH dehydrogenase activity → basal ganglia and optic nerve vulnerability.

## Two real conflicts, curated as open questions

I did not pick a side on either. Both are `KNOWLEDGE_GAP` discussions with the experiment that would settle them.

**1. Is NDUFA12 required for holoenzyme formation, or only for stability?**

miRNA knockdown says NDUFA12 is *necessary for the formation of a functional holoenzyme*. But patient fibroblasts with a nonsense allele and **no detectable NDUFA12 protein at all** still contain fully assembled, enzymatically active complex I — in reduced amounts. Those aren't both complete accounts, so the patient-cell item is graded `REFUTE` against the strict-requirement reading.

The proposed bridge is the paralogue. NDUFAF2 associates with the **active** complex specifically in NDUFA12-mutant cells. I curated the specificity control explicitly, because it's what makes this a real finding rather than a generic response: in cells mutated in *other* complex I subunits, NDUFAF2 is instead on an inactive subcomplex (NDUFS1, NDUFV1) or undetected (NDUFS7). The dependency is also asymmetric — losing NDUFS4 nearly eliminates NDUFA12, but not the reverse.

**2. Why does uniform, complete protein absence produce different diseases?**

This is an unusually clean natural experiment. Alleles homozygous and truncating, protein virtually absent in every reported case, biochemistry isolated complex I deficiency throughout — yet the range runs from fatal infantile Leigh syndrome to **isolated optic atrophy with a normal MRI**, with heterogeneity for the same variant *within* families. Since the primary lesion doesn't vary, the modifier is elsewhere. The discussion argues NDUFAF2 compensation efficiency is the most tractable candidate, because it's already measurable by BN-PAGE in patient fibroblast lines that exist.

## Deep-research quality — flagging a possible pattern

The `claude_code` report set **`needs_review: true`**: 11/11 references resolved but **0 of 1 quotes validated**, naming PMID:21617257 — the founding NDUFA12 paper this entry cites heavily. 9/11 on topic.

No quoted material was taken from the report. All 35 snippets were read directly from the cached abstracts and verify.

**This is the second consecutive entry this session where that provider failed quote validation on a paper the entry then cited** — #9623 (SRTD6) was 2 of 3 quotes unsupported. Different papers each time, so it isn't one bad record. It may be worth someone checking whether the provider's quote extraction is systematically unreliable, since a curator who trusted those quotes would be pasting unverifiable text into evidence items.

Falcon/Edison remained unavailable (HTTP 402, key out of credit — all session).

## Things recorded so they aren't mistaken for something else

- **GeneReviews PMID:26425749** is fetched and tagged in `references:`, but its cached record holds only the chapter's *purpose statement* — no Clinical Characteristics section. So no phenotype could be cross-referenced from it and **nothing is cited to it as evidence**. It's a tagged baseline that couldn't actually serve as one.
- **PMID:34888988 and PMID:34888984** — a published disagreement about whether NDUFA12 Leigh phenotypic heterogeneity is multicausal — were fetched and then *not* cited. Their cached records contain no abstract text, so no snippet could be quoted. That exchange bears directly on the second discussion above and is worth revisiting if full text becomes available.

## Other judgement calls

- **The Leigh syndrome phenotype is bound to `HP:0012751` (Abnormal basal ganglia MRI signal intensity)**, not to a disease term, with a `notes` field saying why: within a `Disease` entry the phenotype slot should carry the observable finding, and Leigh syndrome itself is in `parents`. Disagree if you like — the reasoning is stated rather than left implicit.
- **The rarity item is graded `PARTIAL`.** "No mutations found among 122 screened complex I deficient patients" bounds frequency; it isn't positive evidence about the gene's role.

## Validation

`just validate-disorders` passes — schema, terms and references — with **35/35 snippets verified**.

Closes #9624

---

---

---

> [!NOTE]
> **Low Risk**
> Adds curated YAML, reference caches, and term lists only; no runtime application, auth, or data-pipeline logic changes.
> 
> **Overview**
> Adds a new Mendelian disorder knowledge-base entry for **mitochondrial complex I deficiency, nuclear type 23** (`MONDO:0032627`, **NDUFA12**), closing issue #9624.
> 
> The entry documents biallelic loss-of-function **NDUFA12** disease with a multi-step pathophysiology model (Q-module assembly/stability, **NDUFAF2** compensatory association, reduced complex I activity, separate basal ganglia vs optic nerve vulnerability) and links pathograph nodes to curated phenotypes. Clinical content includes inheritance, variants, phenotypes with frequency bands, prevalence, biochemical reference ranges, symptomatic treatments (including negative dystonia trials), diagnostic workup, and two **KNOWLEDGE_GAP** discussions for conflicting assembly evidence and genotype–phenotype variability (including an asymptomatic homozygote).
> 
> Supporting updates register new ontology/cache terms (**hgnc:23987**, **HP:0033044**, disease/phenotype enums) and add cached PMIDs used as evidence. Session **history** YAML files record initial curation and review-driven edits (pathograph fixes, phenotype frequencies, expanded sections). A **deep-research** report artifact is included for traceability; curated snippets were verified against reference cache, not copied from that report.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c91e3182d2f190cfa7617fbf99c05f77e8d17b89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/references_cache/PMID_41694149.md
+++ b/references_cache/PMID_41694149.md
@@ -14,7 +14,10 @@ doi: 10.4103/jfcm.jfcm_322_25
 publication_types:
 - Case Reports
 - Journal Article
-content_type: abstract_only
+content_type: full_text_xml
+full_text_attempted: true
+full_text_provider: pmc
+oa_status: green
 ---
 
 # Identifying NDUFA12 mutation in a Saudi family: An unusual presentation of mitochondrial Complex I deficiency mimicking as idiopathic intracranial hypertension in a patient with papilledema and visual loss.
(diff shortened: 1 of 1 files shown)


cmungall:
Both reviews addressed. Bugbot's high-severity finding was right: the pathograph had a compensation mechanism wired as a cause of the deficiency it compensates for.

`NDUFAF2 Compensatory Association` carried a `DIRECT` downstream edge into `Reduced NADH Dehydrogenase Activity`. The assembly-and-stability node upstream of both already causes the reduced activity, so that edge added nothing except the wrong claim. `CausalEdge` has no polarity slot, so a mitigating relationship can't be drawn as a downstream edge at all — I removed the edge and left a node-level note saying so, since otherwise the missing edge just looks like an oversight to the next curator.

Bugbot's two frequency findings were also right:

- **Optic Atrophy** `OCCASIONAL` → `FREQUENT`. The pooled 18-patient review gives visual impairment in 7 of 18 (38.9%). The MRI n=2 figure I'd reasoned from is a different measure over a different denominator, and the `notes` now say that explicitly.
- **Motor Regression** `FREQUENT` → `OCCASIONAL`. It rested on the founding case alone. The pooled review doesn't list regression among the main or minor manifestations and reports delayed milestones in 5 of 18, which is now quoted as the bound.

On your findings:

- **Finding 1 (`reference_title`)** is already fixed — commit `2d973c14b` added titles to all 35 items and landed after the commit you reviewed (`7066af33b`).
- **Finding 6 (bundled terminal node, disconnected phenotypes).** Split into `Basal Ganglia Vulnerability` (`UBERON:0002420`) and `Optic Nerve Vulnerability` (`UBERON:0000941`), and both now carry `downstream: target: phenotypes#…` edges. The split earns its keep rather than just tidying: the two tissues dissociate under *identical* homozygous truncating genotypes, and the authors note visual involvement can appear years after a purely motor presentation — so the optic-nerve node now carries their surveillance recommendation.
- **Finding 2 (PMID:35141356 under-mined).** It's now the source for a `biochemical:` section, a `treatments:` section, a second variant, three phenotypes and a prevalence record.
- **Finding 3 (`treatments:`).** Added, and both entries are negative or neutral. The mitochondrial cocktail is `PARTIAL` because the same sentence that records the prescription records the patient being lost to follow-up — there is no outcome to attach. The dystonia entry is `REFUTE`: levodopa and anticholinergics did not work, which is the clinically actionable point and distinguishes this dystonia from the dopa-responsive dystonias it can resemble at presentation.
- **Finding 4 (`biochemical:`).** Added with `reference_ranges` for complex I activity, plasma lactate and plasma pyruvate. Lactate is curated in both directions — the normal-lactate/raised-pyruvate patient is quoted as a `PARTIAL` counter-case, because a normal lactate does not exclude the diagnosis. The complex I activity interval has no unit in the source and I have not invented one.
- **Finding 5 (PMID:41694149).** Now a `diagnosis` entry: an NDUFA12 patient initially labelled idiopathic intracranial hypertension. That's a real differential given optic atrophy can be the sole manifestation with unremarkable brain MRI.

Also added: kyphoscoliosis (6/18, `FREQUENT`), seizure (2/18, `OCCASIONAL`, with the intractable-focal-seizure case quoted as a counterweight to "sporadic"), dysarthria, and a `prevalence:` record bounded from two directions — 18 published patients, and the negative screen of 122 complex I deficient patients.

One term correction found along the way: `UBERON:0000941`'s canonical label is "cranial nerve II", not "optic nerve". `preferred_term` keeps the clinical name.

Your point about the deep-research reliability claim is fair — a title fragment failing to match an abstract-only record is a retrieval artifact, not fabrication, and two such failures is thin. I'd overstated it.

Validation: `just validate-disorders` passes, 57 of 57 snippets verified. Also run: `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`. One new snippet failed the length check at four words and was widened to the full sentence rather than grandfathered.

---

cmungall:
Findings 2–8 were already fixed before this re-review ran — it read `2d973c14b`, the title-only commit, and the substantive work landed in `f57b0d308` right after. The [comment above](https://github.com/monarch-initiative/dismech/pull/9674#issuecomment-5441930109) covers it: the full text is mined into `biochemical:`, `treatments:`, a second variant, three phenotypes and a prevalence record; the terminal node is split with both halves bound; the optic atrophy frequency is `FREQUENT` from the 7/18 figure; and the NDUFAF2 edge is gone.

On that edge — you offered a one-line `description` as the cheap fix. I removed it instead, and the reasoning is worth stating: `CausalEdge` has no polarity, so a `description` saying "this attenuates rather than causes" would sit next to a structured arrow saying the opposite, and the arrow is what exports read. That's the same shape as the `FAILS_TO_RECAPITULATE` problem on #9668 and the `HP:0000510` one on #9678 — prose carrying a distinction the schema doesn't. The assembly-and-stability node already supplies the correct cause, so the edge was redundant as well as inverted. There's a node-level `notes` recording why it's absent.

This push adds your two new items.

**The asymptomatic homozygote.** You were right that this strengthens the best part of the entry, and it's sharper than I'd have made it on my own: it sets the range at zero. Whatever the modifier is, it can abolish the phenotype entirely rather than merely shift severity — which rules out any explanation depending on the genotype alone being sufficient. That's now in the discussion's rationale with the sentence cited.

**The 10/13 lactate figure.** This one was more than an addition. My `notes` led with the counter-case (normal lactate, raised pyruvate) and implied lactate is unreliable — but the cohort figure says it's raised in 10 of 13, so an elevated lactate is the *usual* finding. The notes now give the frequency first and keep the counter-case second, and the explanation records that the denominator counts any of blood, urine or CSF, so it isn't a plasma-specific figure.

**Motor Regression stays `OCCASIONAL`.** You suggested re-sourcing rather than regrading, pointing at delayed milestones (5/18) and gait impairment as a presenting feature (10/18). Those are different phenotypes. Using either to support a *regression* band would repeat the category error the original grading made — it would just move the unsupported inference from the frequency field into the evidence list. Regression itself is documented in one patient and doesn't appear in the pooled manifestation table, which is what the `notes` record.

**On the phenotype-edge correction:** thanks for checking the actual distribution and withdrawing it. I've kept the edges — they're correct on their own terms, since the tissue-vulnerability nodes do cause the phenotypes they point at, and a rare pattern isn't a wrong one. But it's useful to have on the record that it isn't a convention.

I've also filed #9718 for the `count-verified-snippets` prefix-skip gap, which came out of the DOI mismatch on #9649.

Validation: `just validate-disorders` passes, 59 of 59 snippets verified. Also run: `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`.

---

cmungall:
One of these was right and one wasn't, so let me take them separately.

**The Optic Atrophy frequency — you're right, and it was a bad kind of error.** The previous round rewrote the `notes` and added the 7/18 quote arguing for `FREQUENT`, but the edit anchored *below* the `frequency` field and never touched it. So the entry contradicted itself, with the prose arguing one band and the machine-readable slot holding the other. That's worse than either alone: a reader who checks the reasoning would conclude the value was right. Now `FREQUENT`.

**The lactate number is not invented.** You searched for `2/8`, `two of eight` and `2 of 8` and found nothing. The source prints **"two out eight"** — the paper is missing an "of" — which is why all three searches failed:

> "Finally, we highlight that increased lactate levels were detected in only **two out eight** cases of our new cohort when this testing was performed, whereas plasma pyruvate level was increased in one case with normal plasma lactate."

Both figures are real, and they don't conflict — the denominators differ. 10 of 13 is the **pooled** cohort including previously published cases; two of eight is the authors' **own new series**. The paper reconciles them itself: *"This proportion is lower than in previously reported cases, which is in keeping with a higher prevalence of LLS phenotype in our series."* So how often lactate is normal tracks where in the phenotypic spectrum the patient sits, which is a more useful thing to record than either figure alone.

The record now carries all three sentences with their denominators distinguished. The explanation on the new-cohort item states the "two out eight" spelling explicitly, so the next person who searches the ordinary phrasing and finds nothing doesn't reach the same conclusion.

**You were right about the effect, though, even if not the cause.** The previous `notes` led with the two-of-eight figure and built a "should not be treated as a screening test" framing on it — which does understate how often lactate is abnormal. That framing was wrong even though the number wasn't. Leading with the pooled figure and keeping the new-cohort one as the counterweight is the accurate reading, and that's what it says now.

The asymptomatic homozygote and the 10/13 figure both landed in `c91e3182d`, which was pushed just before this review ran — the [comment above](https://github.com/monarch-initiative/dismech/pull/9674#issuecomment-5442690312) covers them.

Also worth flagging for the general case: `notes` isn't snippet-validated, which is exactly the gap you identified, and it's the same class of problem as #9718 (snippet verification silently skipping non-PMID prefixes). Prose fields carrying quantitative claims are unchecked everywhere in the KB.

Validation: `just validate-disorders` passes, 61 of 61 snippets verified. Also run: `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`.

---

cmungall:
Merged `main` in `439bcaa54` to clear a conflict that appeared this morning. The other four PRs from this batch merged in the 06:08–06:10 sweep; this one couldn't, because `main` had moved under it.

Three conflicts, all in shared derived files rather than in the entry:

| File | Resolution |
|---|---|
| `cache/enums/diseaseterm_f618932bb23f.csv` | kept both sides, re-derived with `just normalize-cache` |
| `cache/mondo/terms.csv` | same |
| `references_cache/PMID_22972949.md` | add/add — took `main`'s copy |

On the third: both sides had fetched the same reference, and the only difference was that `main`'s carried an extra `full_text_attempted: true`. Its copy is a strict superset, so I took it rather than mine. That's the same principle as the cache-expansion problem on #9668 — when two copies of a fetched reference differ, the fuller one wins.

`git diff --stat origin/main...HEAD -- kb/` is exactly one file, 961 insertions — the entry itself, nothing dropped in the merge.

Validation on the merged head: `just validate` clean with 61/61 snippets, plus `check-entity-refs`, `check-duplicate-keys`, `validate-terms`, `check-term-cache-integrity` (18 term + 24 enum files).

This resets the approval, which is unavoidable — a conflicted PR can't merge at all.

---

PR #9662, opened 2026-08-27
curate: Add Li-Ghorbani-Weisz-Hubshman Syndrome (MONDO:0033547, KAT8)

cmungall:
New Mendelian entry for the KAT8 chromatinopathy — the MYST-family acetyltransferase that writes H4K16ac, sitting beside the KAT6A and KAT6B syndromes.

Six-node pathograph: KAT8 loss of function → loss of nucleosomal H4K16 acetylation → impaired neural stem/progenitor expansion → aberrant neurogenesis + neuroepithelial apoptosis → cerebral hypoplasia.

## What makes the mechanism interpretable rather than correlative

Two design details in the founding study (PMID:31794431) do the real work, and I've curated both explicitly:

- **The conditional knockout is lineage-restricted.** H4K16ac disappears from the cerebrocortical neuroepithelium and hippocampal primordium at E12.5 but is *retained* in the ganglionic eminences. That's what attributes the hypoplasia to KAT8 in that lineage rather than to a global effect.
- **The neurosphere experiment shows an acute requirement.** Pharmacological KAT8 inhibition abolishes neurosphere formation from *wild-type* cortex, not just knockout tissue. And the inhibitor comparison was designed for this: MG149 hits KAT5 and KAT8, NU9056 is KAT5-specific, so the difference isolates KAT8 — no KAT8-selective inhibitor existed at the time.

## Two places I hedged on purpose

**Valproate.** There's a genuinely attractive argument: KAT8 writes an acetylation mark, valproate inhibits the deacetylases that erase acetylation marks, and both treated individuals responded. I curated the treatment and the rationale — but graded the `target_mechanisms` evidence `PARTIAL`, because the authors write "**may** ameliorate", and added a `KNOWLEDGE_GAP` discussion. Two responders is not a trial, nobody has measured H4K16ac in a treated patient, and valproate is a broad-spectrum anticonvulsant that would help many epilepsies regardless of mechanism. It's a reasonable first choice and a testable hypothesis, not established mechanism-directed therapy. Worth getting right, because if the logic holds it generalises to the other writer chromatinopathies.

[...]

PR #9655, opened 2026-08-27
curate: Add Hereditary Spastic Paraplegia 49 (MONDO:0014016, TECPR2)

cmungall:
New Mendelian entry for TECPR2-related disease. Five-node pathograph: TECPR2 loss of function → destabilised COPII coat / ER exit site failure + impaired autophagosome formation → sensory-autonomic neuron degeneration and corticospinal axonopathy.

## The disease name is wrong, and that has clinical consequences

This is the thing to look at first. MONDO:0014016 is labelled *hereditary spastic paraplegia 49* because the 2012 discovery paper studied five individuals from three Jewish Bukharian families and described them as having spastic paraparesis. Everything since points elsewhere:

- The second patient series concluded the main disabling feature is autonomic-sensory neuropathy and **proposed reclassification** as a hereditary sensory-autonomic neuropathy (PMID:26542466).
- The largest phenotyping cohort — 17 new plus 11 published individuals — lists developmental delay/ID, hypotonia, ataxia, hyporeflexia, respiratory infections and central/nocturnal hypopnea as core manifestations. **Spasticity is not on that list** (PMID:33847017).
- GeneReviews titles its chapter *TECPR2-Related Hereditary Sensory and Autonomic Neuropathy with Intellectual Disability* (PMID:36137062).
- Patients have **hyporeflexia**, the opposite of what a pure corticospinal lesion produces.

It matters because **apnea and asphyxia are the two most common causes of death**. A clinician who finds this under a spastic paraplegia label may not think to order polysomnography.

I kept the MONDO binding and the stub's filename — that is the concept that was claimed — but the description opens with the problem, the `Corticospinal Axonopathy` node is curated as explicitly secondary, and there is a `KNOWLEDGE_GAP` discussion on what would settle it. **If you'd rather the entry were renamed, say so — it's a cheap change and I have no attachment to the current name.**

## A real conflict in the literature, curated as unresolved

The published autophagy readouts point in opposite directions, and I did not smooth this over:

| Source | Finding | Implies |
|---|---|---|
| Patient fibroblasts (PMID:23176824) | LC3-II **and** p62 *decreased* | defect at autophagosome **formation** |
| Spanish Water Dog (PMID:26555167) | autophagosomes *accumulate* in spheroids | defect at **clearance** |
| Tecpr2 KO mouse (PMID:33218264) | autophagosomes *accumulate*; authors read it as failed lysosomal targeting | defect at **clearance** |
[...]

PR #9678, opened 2026-08-27
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 11 (MONDO:0011032, MYO7A)

### LINKED ISSUE #9626: Curate autosomal dominant nonsyndromic hearing loss 11 (MONDO:0011032)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 11** ([MONDO:0011032](https://monarchinitiative.org/MONDO:0011032)) — DFNA11, progressive postlingual hearing loss caused by *MYO7A* (`hgnc:7606`) variants. Myosin VIIa is a hair-cell unconventional motor that anchors the stereocilia tip-link and Usher-protein complex.

The reason this one is worth curating rather than just recording: *MYO7A* is an allelic series. The same gene causes Usher syndrome type 1B (recessive, deaf-blind), DFNB2 (recessive, hearing only) and DFNA11 (dominant, hearing only), and the dominant DFNA11 alleles are thought to act by a different mechanism than the recessive null alleles. That divergence is the mechanistic content of the entry.

Check the existing `Usher_Syndrome` entry first so the DFNA11 entry cross-references it instead of restating the *MYO7A* biology inconsistently.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11.yaml`

cmungall:
New Mendelian entry for DFNA11 — the dominant, hearing-only form of MYO7A disease.

Four-node pathograph: heterozygous MYO7A missense variant → loss of IQ-motif autoregulation → constitutive myosin VIIa motor activation → progressive stereocilia dysfunction.

## The claim this entry is built on: allele *type* selects the disease

MYO7A causes three diseases, and they aren't a severity gradient. **Two independent kinds of evidence** say the allele class picks which one:

**Genetic**, from a 10,042-patient cohort:

| Disease | Allele classes found |
|---|---|
| DFNA11 (dominant, hearing only) | **exclusively** missense or in-frame deletion |
| DFNB2 (recessive, hearing only) | at least one missense in every patient |
| USH1B (recessive, deaf-blind) | **biallelic nulls in nearly half** |

So dominant disease is *not* haploinsufficiency — if it were, null alleles would turn up among the dominant cases, and they don't.

**Mechanistic**: many DFNA11 alleles **activate** Myo7A motor-dependent targeting in cultured epithelial cells, which the authors offer as the explanation for the dominant inheritance.

Two evidence types of different kinds converging is what lifts this above a cell-culture observation.

## Where I made an interpretive move, and flagged it

The `Loss of IQ-Motif Autoregulation` node proposes that **calmodulin occupancy of the IQ motifs is the brake that DFNA11 alleles release** — reconciling the 2024 activation work with the 2004 finding that a DFNA11 IQ5 allele abolishes calmodulin binding.

That reconciliation is mine, not either paper's. The node description says so explicitly: the 2004 authors interpreted their own result as *impaired adaptation to environmental stimuli*, not as de-repression of the motor. And rather than burying that in a caveat sentence, **their interpretation is curated as its own `PARTIAL`-graded evidence item**, so a future reader sees the competing reading in the evidence record itself.

## Other calls worth a look

- **`GAIN_OF_FUNCTION` rather than `INCREASED`** on the activation node, with a `notes` field giving the CLAUDE.md reasoning: the claim is qualitative — the motor escapes its regulatory constraint — not that binding is merely quantitatively higher. That's exactly the line CLAUDE.md draws, so I wanted the choice on the record.

- **"Absence of Retinitis Pigmentosa" is curated as a phenotype, with its primary evidence graded `REFUTE`.** The asserted claim is an *absence*, so `REFUTE` is the honest grading. MYO7A is the commonest Usher type 1 gene — the absence of retinopathy is the single most consequential thing to tell a newly diagnosed DFNA11 family, and it earns explicit curation rather than being left as a silent omission. Bound to HP:0000510 (Rod-cone dystrophy), with `notes` explaining both choices.

- **The prevalence record is `PARTIAL`, and its `notes` say what the figure is *not*.** 1.36% covers all three MYO7A phenotypes together and is a proportion within an ascertained hearing-loss cohort — not a population rate for DFNA11. No DFNA11-specific rate exists.

- **The high-frequency phenotype carries a `PARTIAL` item that contradicts it** — one large pedigree's loss begins at low and mid frequencies. It qualifies the typical configuration rather than supporting it.

## Two discussions, one of which names a missing experiment

The **allele-class discussion** asks whether the correlation runs *prospectively*. The activation paper says "many" DFNA11 mutations activate, not all — so it's unknown whether untested dominant alleles activate, whether some act dominant-negatively instead, and critically **whether a recessive missense allele would fail to activate in the same assay**. That last comparison is what would turn a correlation into a usable predictor for a novel missense variant in an isolated case, and it doesn't appear to have been run. The assay exists and the alleles are published.

The **modifier discussion** covers within-family variability under a single allele, including the `REFUTE`-graded negative result that ATP2B2 and WFS1 SNPs did not segregate with severity.

## Deep research

`needs_review: true` — 0 of 1 quotes validated, on PMID:32097363. **That PMID isn't cited here**, so unlike #9668 and #9674 the failure doesn't touch curated content.

Running tally for this session: **three of five `claude_code` reports failed quote validation** (#9668, #9674, this one); two were clean. No quoted material was taken from any report — every snippet was read from the cached abstract or full text. Falcon/Edison returned HTTP 402 throughout. No GeneReviews chapter for DFNA11.

## Validation

`just validate-disorders` passes — schema, terms and references — with **32/32 snippets verified**.

Closes #9626

---

---

---

> [!NOTE]
> **Medium Risk**
> Curated clinical knowledge with a prior export inversion that could have emitted the opposite retinopathy assertion; fixes reduce downstream graph risk but allele-class mechanism remains interpretive.
> 
> **Overview**
> Adds a new Mendelian knowledge-base entry for **autosomal dominant nonsyndromic hearing loss 11 (DFNA11)**, linked to **MONDO:0011032** and **MYO7A**, with ontology/cache updates for the new disease, phenotype (**HP:0008596**), and molecular function (**GO:0005516**) terms plus cached reference material for the cited literature.
> 
> The entry frames DFNA11 as the hearing-only dominant arm of the MYO7A allelic series: a four-step pathograph from heterozygous missense/in-frame variants through IQ-motif autoregulation loss and proposed motor **gain-of-function** to progressive stereocilia dysfunction, with phenotypes, inheritance, domain-based **genetic** correlations, **PARTIAL** prevalence figures from ascertained cohorts, **diagnosis** (longitudinal ophthalmology vs USH1B, early DPOAE changes), class-level **treatments**, **progression**, **differential** Usher syndrome type 1B (GeneReviews **PMID:20301442**), and discussions on allele-class prediction and human–model mismatch.
> 
> A review correction removes **phenotype_term** and **frequency** from “Absence of Retinitis Pigmentosa” so KG export cannot assert rod-cone dystrophy in DFNA11 patients; **GO:0000146** replaces actin-binding on the activation node, stereocilia dysfunction links to phenotypes, and validation history reports **53/53** snippet checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec4995005c10a6a53d8c7cef4809ad61e2ba3d34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11/2026-08-27T214323Z-claude-code-31194d.yaml
+++ b/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11/2026-08-27T214323Z-claude-code-31194d.yaml
@@ -1,0 +1,66 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11
+  path: kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11.yaml
+session:
+  id: 2026-08-27T214323Z-claude-code-31194d
+  timestamp: '2026-08-27T21:43:23Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9626
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9678
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - genetic
+  summary: 'Round-4 review: fix drifted title, duplicate evidence item, fragment snippet'
+  details: |
+    Fixed the three defects from the round-4 review on PR #9678. All three were mine, all
+    three were introduced by the last two commits, and all three were verified against the
+    cache before fixing.
+
+    1. reference_title for PMID:23383098 did not match the cache. Three occurrences read
+    "...in a family with autosomal dominant nonsyndromic hearing loss." The title: frontmatter
+    in references_cache/PMID_23383098.md reads "...in a family with autosomal dominant hearing
+    impairment (DFNA11)." I wrote it from the sense of the paper rather than copying it, which
+    is exactly what the title field exists to prevent: a drifted title is indistinguishable
+    from a fabricated citation in a diff, so a reviewer has to open the cache to tell them
+    apart. Copied verbatim.
+
+    2. Duplicate evidence item in genetic[0].evidence. The PMID:38594301 snippet "Among the
+    autosomal recessive cases, all patients had at least one missense variant." appeared
+    twice, once with no explanation. Introduced in ec4995005 while the domain-correlation
+    items were being inserted. Nothing in the validation stack catches a duplicated evidence
+    item - schema validation, snippet verification and check-duplicate-keys all pass with it
+    present, because it is a repeated list element rather than a repeated mapping key. Deleted
+    the copy without the explanation. Snippet count drops 57 to 56, which is the duplicate
+    leaving.
+
+    3. The p.R668H segregation snippet was a fragment missing its subject. It read "is
+    heterozygous in all affected family members and absent in 100 healthy individuals", which
+    on a variant-level entry does not say which variant is heterozygous. I had truncated it to
+    avoid the Unicode arrow in the source; the arrow verifies fine as an exact substring.
+    Extended to the full sentence: "A c.2003G-arrow-A (p.R668H) mutation of the MYO7A, is
+    heterozygous in all affected family members and absent in 100 healthy individuals." Note
+    the cache carries this sentence twice, once as "the MYO7A" and once as "theMYO7A" with the
+    space missing; the well-formed copy is the one quoted.
+
+    Not done: the two non-blocking suggestions. The POINT_PREVALENCE records are a
+    PrevalenceMeasureEnum limitation the notes are already candid about, and citing
+    PMID:24698764 in the HUMAN_MODEL_MISMATCH discussion was considered and set aside in the
+    previous pass because that zebrafish model is a biallelic null of USH1B, a different
+    disease in the allelic series.
+
+    Validation: just validate (56/56 snippets verified, down from 57 as the duplicate left),
+    check-entity-refs, check-duplicate-keys, validate-terms, tests/test_reference_titles.py
+    (38 passed), check-snippet-grading, check-snippet-length, check-title-snippets,
+    check-folded-hyphens all pass.
--- a/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11.yaml
+++ b/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11.yaml
@@ -430,11 +430,6 @@ genetic:
     evidence_source: HUMAN_CLINICAL
     snippet: "The MYO7A gene is known to be responsible for both syndromic hearing loss (Usher syndrome type1B:USH1B) and non-syndromic hearing loss including autosomal dominant and autosomal recessive inheritance (DFNA11, DFNB2)."
     explanation: Establishes the three-disease allelic series at this locus.
-  - reference: PMID:38594301
-    reference_title: "The prevalence and clinical features of MYO7A-related hearing loss including DFNA11, DFNB2 and USH1B."
-    supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
-    snippet: "Among the autosomal recessive cases, all patients had at least one missense variant."
   - reference: PMID:35453549
     reference_title: "Clinical Heterogeneity Associated with MYO7A Variants Relies on Affected Domains."
     supports: SUPPORT
@@ -506,13 +501,13 @@ genetic:
       counterexample the allele-class discussion identifies as untested.
     evidence:
     - reference: PMID:23383098
-      reference_title: "Identification and functional study of a new missense mutation in the motor head domain of myosin VIIA in a family with autosomal dominant nonsyndromic hearing loss."
+      reference_title: "Identification and functional study of a new missense mutation in the motor head domain of myosin VIIA in a family with autosomal dominant hearing impairment (DFNA11)."
       supports: SUPPORT
       evidence_source: HUMAN_CLINICAL
-      snippet: "is heterozygous in all affected family members and absent in 100 healthy individuals"
+      snippet: "A c.2003G→A (p.R668H) mutation of the MYO7A, is heterozygous in all affected family members and absent in 100 healthy individuals."
       explanation: Establishes dominant segregation of this allele in the reported family.
     - reference: PMID:23383098
-      reference_title: "Identification and functional study of a new missense mutation in the motor head domain of myosin VIIA in a family with autosomal dominant nonsyndromic hearing loss."
+      reference_title: "Identification and functional study of a new missense mutation in the motor head domain of myosin VIIA in a family with autosomal dominant hearing impairment (DFNA11)."
       supports: SUPPORT
       evidence_source: IN_VITRO
       snippet: "In the actin-activated ATPase activity assay, the rate of NADH oxidation was higher in the wild-type myosin VIIA, indicating that the ATPase activity in the p.R668H mutant myosin VIIA was significantly destroyed."
@@ -876,7 +871,7 @@ discussions:
     snippet: "All variants identified in autosomal dominant cases were missense or in-frame deletion variants."
     explanation: The genetic correlation that the mechanism would explain.
   - reference: PMID:23383098
-    reference_title: "Identification and functional study of a new missense mutation in the motor head domain of myosin VIIA in a family with autosomal dominant nonsyndromic hearing loss."
+    reference_title: "Identification and functional study of a new missense mutation in the motor head domain of myosin VIIA in a family with autosomal dominant hearing impairment (DFNA11)."
     supports: PARTIAL
     evidence_source: IN_VITRO
     snippet: "In the actin-activated ATPase activity assay, the rate of NADH oxidation was higher in the wild-type myosin VIIA, indicating that the ATPase activity in the p.R668H mutant myosin VIIA was significantly destroyed."


cmungall:
The inversion is fixed, along with everything restated in the second round.

**The inverted export.** You traced it exactly right. `Absence of Retinitis Pigmentosa` carried `phenotype_term: HP:0000510` and `frequency: VERY_FREQUENT`; `kgx_export.py` reads those two and never reads `supports`, so the exported triple said 80–99% of DFNA11 patients have rod-cone dystrophy — the opposite of the entry's headline claim. I've followed the ALS27 pattern: no `phenotype_term`, no `frequency`, `diagnostic: true`, and the reasoning written into the `description` rather than only into `notes`. That last part is deliberate — otherwise the missing term reads as an oversight and someone "fixes" it back into a wrong assertion.

**`GO:0051015` → `GO:0000146`** microfilament motor activity, on a node whose own notes argue about motor activity. `GAIN_OF_FUNCTION` unchanged.

**`PMID:32350269`** now cited three times on `Progressive Stereocilia Dysfunction`. You were right that the node was resting on two `OTHER`-graded sentences from a cohort paper's introduction. The primary paper supplies the electrophysiology (reduced resting open probability, slowed MET current onset) and the develop-then-degenerate timing the node claims. I've also quoted its "conclusive evidence is lacking" sentence as `PARTIAL` — that's the honest characterisation of what those two review sentences were asserting before this study existed.

**Open-ended pathograph.** The stereocilia node now carries three `downstream` edges to the auditory and vestibular phenotypes.

**Prevalence, genetics, diagnosis, treatments, progression** all added — the domain correlation, the 4.7%-of-postlingual-families figure with the first record's note corrected, ophthalmologic assessment and DPOAE, hearing amplification and counselling, and the audiogram-configuration progression. Two notes on those: hearing amplification carries **no NCIT binding**, because NCIT's hearing-aid and cochlear-implant concepts are device terms not reachable from `NCIT:C25218` and there's no action term for fitting or implantation; and the progression section includes a `PARTIAL` item quoting *normal* vestibular function, which conflicts with the caloric abnormalities curated in the vestibular phenotype from another family. I quoted it rather than omitting it — an entry carrying only the positive finding would overstate vestibular involvement.

**The GeneReviews question was worth asking, and I got it half wrong.** You were right that the research artifact's "MYO7A-Related Hearing Loss and Retinitis Pigmentosa" chapter is a confabulation — no such chapter exists. But searching properly surfaced two real chapters: **Usher Syndrome Type I** (`PMID:20301442`) and **Genetic Hearing Loss Overview** (`PMID:20301607`). My original statement — no DFNA11 chapter — was true, and it was the wrong question to have asked, because the USH1 chapter documents the differential this entire entry is built on. It's now tagged `GeneReviews` and mined into a `differential_diagnoses` entry for Usher syndrome type 1B. The useful thing it supplies: **RP in USH1 develops in early adolescence**, which is precisely why a normal childhood retinal examination doesn't distinguish the two, and why the diagnosis section recommends longitudinal rather than single assessment.

**`HUMAN_MODEL_MISMATCH` discussion added**, as you suggested. Every in vivo MYO7A model is a deletion, so they model the recessive/Usher end of the allelic series; the activation claim rests on a cultured kidney epithelial assay. The discussion states the consequence concretely: the mechanism predicts a therapy restoring autoinhibition would help in DFNA11 and be useless or harmful in DFNB2 and USH1B, and there's currently no in vivo system in which to test that.

**Two things not done, with reasons.** `PMID:9354784`, the 1997 discovery paper, has no retrievable content in its cached record — nothing can be quoted from it. And I didn't add `experimental_models:` for the epithelial targeting assay: it's now cited in the mismatch discussion where its heterologous nature is the *point*, and modelling it as an `ExperimentalModel` with `modeled_mechanisms` would assert a fidelity relationship the paper doesn't support.

Validation: `just validate-disorders` passes, 53 of 53 snippets verified. Also run: `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`.

---

cmungall:
All three fixed. All three were mine, and the third one taught me something worth writing down.

**1. The activation node's `notes` still argued about the term they replaced.** The last round rebound `GO:0051015` → `GO:0000146` but left the notes saying "The GO binding is `actin filament binding`" and arguing about whether *binding* is quantitatively increased — the exact concept the swap was meant to move away from. Rewritten with the term, and it now records what the binding was and why it was wrong so the change reads as deliberate rather than as churn.

This is the same failure mode as the Optic Atrophy frequency field on #9674 this session: the reasoning gets rewritten in one place and the thing it describes in another, and the two drift. Worth me watching for.

**2. I generalised a real constraint into a false one.** The treatment note claimed NCIT has no clinical-action term for either hearing-aid fitting *or* cochlear implantation. True for the first — `NCIT:C183182` is a device concept, unreachable from `NCIT:C25218` — and false for the second, as both siblings you named prove: DFNA17 and `MYO6_Hearing_Loss` bind implantation to `NCIT:C15329` Surgical Procedure. Implantation is a procedure and procedures bind; the device doesn't. Lumping the two treatments into one entry is what let the amplification gap get generalised without being checked.

Split into two. Amplification keeps the free-text term, with the note now scoped to amplification and pointing at the implantation entry so the distinction doesn't collapse again. Implantation takes `NCIT:C15329`. Both carry `target_phenotypes`, which also closes the treatment join point. The implantation evidence is `PARTIAL` and labelled group-level — no DFNA11-specific implantation outcome exists in the cache, so the entry records the intervention without claiming an outcome for this genotype.

**3. The line-break reason was wrong, and the correction generalises.** You were right — I tested it, and both full sentences verify (54/54).

The distinction I'd collapsed: the validator **normalises clean line wraps**, so a sentence wrapped at a word boundary can be quoted whole. It **cannot bridge hyphenated word splits** — a two-column PDF extraction breaking a word as `chromo-\nbarrel`, which is what forced the genuinely mid-word truncations in the KAT8 entry (#9662) earlier in this session. I hit the hyphenation case first, generalised it to all line breaks, and then wrote that false constraint into an `explanation` — where it would have taught the next curator to truncate quotes that didn't need truncating.

And it cost real content, exactly as you said. The first quote was cited as "the USH1 triad" but stopped before the third element, adolescent-onset retinitis pigmentosa — the one that matters for this differential. The second was cited for RP timing but was a fragment whose subject sat in the previous sentence, so as stored it didn't say what develops in early adolescence. Both are now full sentences.

Thanks for withdrawing the PMID:9354784 ask and for checking the cache yourself rather than leaving it standing.

Not taken, for the record: `PMID:23383098` on the motor-domain correlation is a fair suggestion and I've left it — the correlation already has the domain-clustering and onset sentences from the cohort paper, and a third citation would add a case rather than a claim. And I agree the `POINT_PREVALENCE` grading on two diagnostic yields is a schema limitation rather than a curation error; the `notes` on both records say what they actually are.

Validation: `just validate-disorders` passes, 54 of 54 snippets verified. Also run: `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`.

---

cmungall:
The `check-environmental-evidence` failure on `45ca7278c` was not from this PR. It flagged nine exposures across five files — `Infective_Endocarditis`, `Invasive_Candidiasis`, `Pulmonary_Embolism`, `Rheumatic_Heart_Disease`, and `Epilepsy_with_Generalized_Tonic-Clonic_Seizures_Alone` — none of which this branch touches. All nine are already on `main`, and the check passes there; the check grandfathers against `origin/main`, so a branch that has fallen behind sees them as new.

I've merged current `main` (`17ab5a12f`) into the branch and pushed, which should clear it. Same cause and same fix as on #9655.

---

cmungall:
All three fixed in `1176899a0`. Each was mine, each came in with the last two commits, and each was checked against the cache before I touched it.

**1. Drifted `reference_title`.** Confirmed — the cache frontmatter reads "...autosomal dominant hearing impairment (DFNA11)" and I had written "...autosomal dominant nonsyndromic hearing loss" in all three places. I wrote it from the sense of the paper instead of copying it, which is precisely what the field exists to prevent. Your reason for insisting is the right one: you had to open the cache to tell a retyped title from a fabricated one, and that cost is the whole point of the check. Copied verbatim.

**2. Duplicate evidence item.** Confirmed at lines 434-437 and 461-468. Deleted the copy without an `explanation`; snippet count drops 57 → 56, which is the duplicate leaving.

Worth noting *why* you had to find this by scanning the parsed file: **nothing in the validation stack catches a duplicated evidence item.** It's a repeated list element, not a repeated mapping key, so `check-duplicate-keys` doesn't see it, schema validation is happy, and snippet verification cheerfully verifies the same quote twice. That's a real gap rather than an oversight on your part — I'll file it separately rather than bury it here.

**3. Fragment snippet.** Confirmed. I had truncated it specifically to dodge the Unicode arrow in the source; the arrow verifies fine as an exact substring, so the truncation bought nothing and cost the subject. Now the full sentence:

```
A c.2003G→A (p.R668H) mutation of the MYO7A, is heterozygous in all affected
family members and absent in 100 healthy individuals.
```

One detail for anyone quoting this paper later: the cache carries that sentence twice, once as `the MYO7A` and once as `theMYO7A` with the space missing. The well-formed copy is the one quoted.

**Not done, deliberately.** The `POINT_PREVALENCE` records are a `PrevalenceMeasureEnum` limitation the notes already flag. And PMID:24698764 in the `HUMAN_MODEL_MISMATCH` discussion I considered on the previous pass and set aside: that zebrafish mutant is a biallelic null modelling USH1B, a different disease in the allelic series, so it would sit oddly in a discussion about DFNA11's own models. Happy to add it if you'd rather have it framed as the contrast case.

Validation on the new head: `just validate` clean with 56/56 snippets, `tests/test_reference_titles.py` 38 passed, plus `check-entity-refs`, `check-duplicate-keys`, `validate-terms`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`.

---

PR #9141, opened 2026-08-21
feat(qc): gate fabricated reference titles against the cache (#9138)

sierra-moxon:
Closes #9138

## The gap

`EvidenceItem.reference_title` and `PublicationReference.title` were validated by **nothing**. Every gate in the stack looks at a different field, so the failure mode — **correct PMID, verified snippet, invented title** — passed all of them:

| Gate | What it checks |
|---|---|
| `linkml-validate` | the slot is a string. It is. |
| `count-verified-snippets` / `validate-references` | the **snippet**. Passes — the PMID is right and the quote is genuine. |
| `validate-terms` | ontology terms only. |
| `check_title_snippets` | despite the name, whether a *snippet* quotes a title. Unrelated. |

Nothing read the title. It renders on the disorder page and flows into the cx2 and SEPIO exports, so a fabricated one ships downstream.

## What this adds

`scripts/check_reference_titles.py` + `just check-reference-titles`, wired into `just qc` and `.github/workflows/main.yaml` as an ungated step alongside the other `check_*` gates (a fabricated title arrives in a kb/-only curation PR, which no `src/tests` path filter would catch).

Pure-offline: the correct title is already on disk in each `references_cache/*.md` frontmatter. No network, no ontology adapter, no new dependency. ~85s over the whole KB.

The failure message prints the cached title so the fix is a copy-paste:

```
kb/disorders/X.yaml:phenotypes[3].evidence[0].reference_title: PMID:34081534 (similarity 0.44)
    curated: Congenital cranial dysinnervation disorders: a concept in evolution.
     cached: Axonal Growth Abnormalities Underlying Ocular Cranial Nerve Disorders.
```

Reuses the existing resolution logic rather than reinventing it — `CachedReferenceIndex.resolve_cache_path` for the prefix→filename mangling, and `title_of` / `normalize` / `dataset_prefixes` single-sourced from `check_title_snippets` so the two guards cannot drift on what a cached title is or what "the same string" means.

## Threshold: 0.85, validated against the real corpus

Similarity ratio (`difflib`) over normalised strings, not equality. Measured over all 613 comparable non-contained pairs in `kb/`:

| bucket | 0.0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.5 | 0.6 | 0.7 | 0.8 | 0.9 | 1.0 |
|---|---|---|---|---|---|---|---|---|---|---|---|
| count | 1 | 2 | 4 | 45 | 40 | 43 | 35 | 21 | 38 | 77 | 307 |

[...]

Issue #8296, opened 2026-08-10
182 environmental exposures across 68 entries have no evidence at all


caufieldjh:
Found while picking targets for the #8085 `influences_mechanisms` backfill.

## The gap

**182 `environmental:` exposures across 68 disorder entries carry no `evidence:` at all** — just a `name` and usually a one-line `notes`. Examples:

```yaml
# kb/disorders/Ulcerative_Colitis.yaml
- name: Smoking
  notes: Protective (unlike Crohn's)
- name: NSAIDs
  notes: May trigger flares
```

```yaml
# kb/disorders/Gastroesophageal_Reflux_Disease.yaml
- name: Obesity
  notes: Major modifiable risk factor
```

Worst-affected entries (6 bare exposures each): `Psoriasis`, `Gout`, `Gastroesophageal_Reflux_Disease`. Then 5 each: `Ulcerative_Colitis`, `Peripheral_Artery_Disease`, `Obstructive_Sleep_Apnea`, `Irritable_Bowel_Syndrome`, `Glaucoma`, `Epilepsy`, `Celiac_Disease`.

## Why it matters

These are uncited claims about disease causation sitting in the knowledge base, and **the whole validation stack is blind to them**. `evidence` is optional on an environmental entry, so `just validate` passes, `validate-terms` passes, and `count-verified-snippets` has nothing to check — an exposure with zero evidence is indistinguishable from a fully-evidenced one in every green check we run. The same structural blind spot as the `influences_mechanisms` reversion in #8167: optional slots are invisible to the tooling that would otherwise catch their absence.

Some are also **substantive claims that a reader would want sourced**. `Smoking → notes: Protective (unlike Crohn's)` is a real and slightly counterintuitive finding in ulcerative colitis; asserting it with no citation is worse than asserting a dull one.

## Relationship to #8085

#8085 is about linking exposures into the pathograph. These 182 can't be linked at all under the evidence discipline that backfill uses — every link carries verified evidence, either reused from the exposure or freshly cited and snippet-checked. **An exposure with no evidence has nothing to carry.** So this is the blocker sitting underneath the remaining #8085 work, not a separate cosmetic concern: roughly a third of the unlinked exposures I surveyed were unlinkable for this reason rather than for any mechanistic one.

## Suggested work

1. Land a coverage report (`scripts/`) listing evidence-free environmental exposures per entry, so the backlog is visible rather than rediscovered per-curator.
[...]

PR #9696, opened 2026-08-27
curate: Add Charcot-Marie-Tooth disease type 4D (MONDO:0011085, NDRG1)

### LINKED ISSUE #9680: Curate Charcot-Marie-Tooth disease type 4D (MONDO:0011085)
Curate a dismech entry for **Charcot-Marie-Tooth disease type 4D** ([MONDO:0011085](https://monarchinitiative.org/MONDO:0011085)) — an autosomal recessive demyelinating neuropathy caused by NDRG1 loss, classically described in the Lom population, with hearing loss alongside the neuropathy.

Per-subtype CMT entries are established convention here (2W, 2JJ, X-linked Recessive 4 all exist alongside the type-level entries). Likely conforms to `peripheral_axonal_degeneration`.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Type_4D.yaml`
Nominated by: unknown

kevinschaper:
Adds a knowledge-base entry for Charcot-Marie-Tooth disease type 4D (CMT4D), also known as hereditary motor and sensory neuropathy-Lom. It is a childhood-onset recessive demyelinating neuropathy with sensorineural deafness, caused by loss of NDRG1 in Schwann cells. The thing that distinguishes it from other demyelinating CMT is that axons are lost early and severely, rather than late as a consequence of chronic demyelination — and that step is the one part of the mechanism no animal model reproduces.

## Entry-type decision

Dedicated `Disease` entry, not a subtype of the existing `Charcot-Marie-Tooth_Disease_Type_4` parent (which already carries a CMT4D subtype block). MONDO:0011085 has no descendants and one causal gene. Per-subtype entries alongside a type-level entry are the established pattern here — CMT2JJ and CMT2W sit alongside `Charcot-Marie-Tooth_Disease_Type_2` the same way. The `notes:` section records the boundary between the two so a later curator does not merge them by mistake.

## Deep research

`just preflight-dr` against MONDO:0011085 returned **PASS**: NDRG1 mentioned 50 times, the next-ranked gene NRG1 only 3, and the report's OMIM ID (601455) matches the MONDO xref. The report's own reference validation found 11 of 12 references resolved, none unresolved, none off topic.

The report was used as a lead source only. Every cited PMID was fetched independently with `just fetch-reference` and every snippet checked against the cached text. Two of its leads (the 1998 clinical delineation, PMID:9549516; the Okuda Ndrg1 knockout, PMID:15082788) turned out to be worth curating and were fetched and used. One of its cautions was superseded: it reported that phenotype frequencies could not be assigned from its evidence, but the 2026 systematic review PMID:42089726 supplies them directly from 72 patients.

## Mechanism

```
Biallelic NDRG1 loss of function
  └─> Loss of Schwann cell NDRG1 function
        ├─> Impaired Schwann cell vesicular trafficking ──┐
        └─> Attenuated neuregulin 1 / ErbB signalling     │
              └─> Dysregulated myelination transcription ─┤
                                                          v
                              Myelin breakdown with onion bulb formation
                                          └─> Early severe axonal loss
                                                └─> Progressive demyelinating sensorimotor neuropathy

Biallelic NDRG1 loss of function ══> Auditory pathway involvement   (unknown intermediates)
```

9 nodes, 9 edges, one root, every node reachable from it (measured, not asserted). Conformance declared to three `peripheral_axonal_degeneration` nodes.

**The signalling arm is mouse-only.** Everything about neuregulin 1/ErbB, the myelination transcription factors and PMP22 turnover comes from `Ndrg1`-deficient mice. Those nodes are `PROVISIONAL` and carry `MODEL_ORGANISM` evidence, and the entry says so in `notes:` rather than letting the pathograph imply human data.

**All three animal models fail at the same point.** The stretcher mouse, the CRISPR knockout and the Greyhound line all reproduce demyelination and none reproduces the early severe axonal loss that actually disables patients. This is curated as `FAILS_TO_RECAPITULATE` links plus a `HUMAN_MODEL_MISMATCH` discussion. The practical consequence — that a myelin-preserving result in these mice says nothing about the axonal arm — is stated explicitly, because the CRISPR paper describes a "window of opportunity for treatment intervention" that applies only to the demyelinating arm.

**Scoliosis carries a contradiction rather than an average.** The 72-patient literature review reports 33%, which puts it in `FREQUENT`; the Bulgarian cohort found it in 1 of 7. Both are curated, the second tagged `PARTIAL`, so the band is read as a literature-wide figure rather than a stable expectation.

**Auditory involvement is not routed through the somatic chain.** Evoked potentials implicate the eighth nerve *and* the central pathways at once, and no cochlear pathology from a CMT4D patient has been reported. The edge from genotype to deafness is typed `INDIRECT_UNKNOWN_INTERMEDIATES` instead of asserting a shared Schwann-cell mechanism.

**Ancestry is a prior, not a criterion.** Older sources call CMT4D a Roma disease. That is no longer correct — the founder p.Arg148* allele covers 64% of patients, not all of them, and a second independent founder effect exists in Bulgarian Muslims. Acting on the older framing would cause missed diagnoses, so the `genetic` notes say so directly.

**Only two treatments, both actually performed in CMT4D patients.** No disease-modifying therapy exists. PEG feeding and non-invasive ventilation are curated because the 2026 sibling report documents them being used; generic CMT rehabilitation is not curated, because it would be extrapolation.

## Validation

- `just count-verified-snippets` — `Snippets checked: 85/85 verified against cached references`
- `just validate-disorders` (the batched CI equivalent) — schema, term and reference validation all passed, 85/85 snippets
- `just check-duplicate-keys`, `just check-title-snippets`, `check_snippet_length --against-ref origin/main`, `check_folded_hyphens --check`, `just check-cache-order`, `just check-term-cache-integrity`, `just validate-history` — all OK

A separate audit confirmed that all 14 `reference_title` values byte-match the `title:` line in their cache frontmatter.

## Not done

`just fetch-reference ORPHA:99950` fails with "No source found for reference type" — the Orphanet cache has no entry for this disorder and `just refresh-orphadata` is known broken on a drifted sha256, so the Orphanet epidemiology row could not be cited. Prevalence is therefore curated as `CASES_IN_LITERATURE` (72 published patients) with `ULTRA_RARE`, which is the honest answer for this entity anyway.

Closes #9680


kevinschaper:
All five 🟡 findings are fixed, plus all four 🔵 suggestions. Every one was correct — thank you for catching them.

**The two ontology mis-bindings.** `HP:0001188` "Hand clenching" is indeed a fetal-posturing term, not a claw hand; replaced with `HP:0034337`. And the `Early Severe Axonal Loss` cell type is now `CL:2000032` peripheral nervous system neuron. That second one was the worst of the five: it is the only node in this pathograph whose substrate is *not* a Schwann cell, which is precisely the entry's argument — NDRG1 is a Schwann-cell protein and the disabling lesion is in a cell that does not express it. The node description now states that explicitly so the point survives a future edit.

**The stale notes paragraph** was leftover draft prose from before the treatments block existed. Rewritten to say what is curated (PEG feeding, non-invasive ventilation — both actually performed in reported CMT4D patients), why generic CMT rehabilitation is deliberately *not* curated, and to keep the separate point that the mouse "window of opportunity" observation is about the demyelinating arm only.

**Prevalence population** changed to `Worldwide`, which is what the 72-case literature total measures.

**Cognitive delay** is now curated, bound to `HP:0100543` Cognitive impairment. I deliberately did not use `Intellectual disability` or `Global developmental delay`: the source says only "cognitive delay" with no instrument, severity or age of recognition, in patients aged 38 and 40, and the developmental terms would assert an origin the source does not establish. The description records that reasoning, and no frequency is assigned since it rests on two siblings and is not independent of the CNS findings curated below it.

**The four suggestions:**

- Animal-model counts are now consistent at four — three rodent, one canine — in both the node description and the discussion.
- Talipes equinovarus curated as `HP:0001762`, with a note that no frequency band is assigned because the 72-patient review reports a pooled figure for pes cavus but not separately for equinovarus.
- Distal amyotrophy split out from the weakness phenotype and bound to `HP:0003693`. A patient can be weak before wasting is visible, so bundling them lost information.
- The two stray NCIT rows were mine — artifacts of two superseded term lookups during curation (`NCIT:C116663` and `NCIT:C51724` turned out to be an autophagy protein and an ACVR2B allele, not the ventilation and gastrostomy terms I was after). Removed, along with an equally stray `HP:0001188` row from the same cause. The cache diff is now six rows, all used by this entry.

Also rebased onto `main` to pick up the new `check-reference-titles` gate (#9141), which passes.

Validation after the fixes: `just validate` ✓ · `just count-verified-snippets` 88/88 ✓ · `just validate-disorders` ✓ · `check-title-snippets`, `check_snippet_length`, `check_folded_hyphens`, `check-cache-order`, `check-term-cache-integrity`, `check-duplicate-keys`, `check-reference-titles` all OK. Pathograph re-measured: 9 nodes, 9 edges, one root, all nodes reachable.

On the GeneReviews adjudication — agreed, and I read it the same way. `PMID:20301532` has no CMT4D-specific content to mine, so any snippet from it would be the title standing in for a finding. Happy to drop the tag entirely if you would rather not have an uncited reference in the list.

Re-requesting review.

kevinschaper:
Rebased onto `main` to clear the conflict. It was a modify/delete on this PR's own stub: `main` ran an enrich-stubs pass over `stubs/Charcot-Marie-Tooth_Disease_Type_4D.yaml` while this branch deletes it. I kept the deletion — the disease is curated, so the stub retires. The post-rebase diff shows exactly one stub deletion and no other churn, and validation still passes (88/88 snippets).

## But that enrich-stubs pass reached the opposite conclusion from this PR, and someone should decide which is right

The commit `main` applied to the stub was not cosmetic. It set `entry_type: SUBTYPE` and added:

> SUBTYPE of Charcot-Marie-Tooth Disease Type 4 (kb/disorders/Charcot-Marie-Tooth_Disease_Type_4.yaml), which already carries CMT4D in its has_subtypes list alongside CMT4A, 4B1, 4B2, 4C, 4F and 4J, and which already names NDRG1. **No separate disorder entry is needed**; further CMT4D detail should be added to that entry's subtype record rather than to a new file.

That is a direct contradiction of this PR, which creates exactly the separate disorder entry the stub says is not needed. I am flagging it rather than quietly letting the deletion settle the question, because deleting the stub also deletes the dissenting opinion.

**Why I still think a dedicated entry is right here**, and what I would want a maintainer to weigh:

- Per-subtype entries alongside a type-level entry are the existing pattern in this repo, not an exception. `Charcot-Marie-Tooth_Disease_Axonal_Type_2JJ`, `Autosomal_Dominant_Charcot-Marie-Tooth_Disease_Type_2W` and `Charcot-Marie-Tooth_Disease_X-linked_Recessive_4` all coexist with their type-level entries, and `Charcot-Marie-Tooth_Disease_Type_2` lists subtypes in the same way `..._Type_4` does. If the stub's rule is correct, those three entries are also mistakes.
- `MONDO:0011085` is a leaf term with its own OMIM ID (601455) and one causal gene. The subtype block inside the CMT4 entry has no MONDO identity of its own, so nothing in the exported graph maps to that MONDO term unless this entry exists.
- The content does not fit a subtype block. This entry carries four animal models with `FAILS_TO_RECAPITULATE` links, a `HUMAN_MODEL_MISMATCH` discussion, three knowledge gaps, and a nine-node pathograph. Folding that into one `has_subtypes[]` record would either lose it or unbalance the parent entry, which covers seven subtypes.
- The two are complementary and the `notes:` section says so explicitly, so a reader arriving at either one is pointed at the other.

**What would change my mind:** if the intent is that MONDO-leaf CMT4 subtypes are deliberately modelled only as `has_subtypes` records and the CMT2 precedents are legacy, then this PR is wrong and the CMT2 entries need the same treatment. That is a call about repository policy, not about this disease, and it is above my pay grade — @cmungall, could you adjudicate? I am happy to convert this into a subtype record if that is the answer.

Worth noting either way: the enrich-stubs pass and a curation PR can reach opposite conclusions about the same stub and only collide as a merge conflict. Had this PR merged an hour earlier there would have been no conflict and no signal at all.

Re-requesting review.

kevinschaper:
I took the non-blocking item rather than deferring it, because the arithmetic favours fixing: the auto-merge sweep gates on **PR creation age**, not approval age, so this PR cannot merge for another ~3 days regardless. A re-review round therefore costs nothing in time-to-merge, and deferring would have left a known internal contradiction in the entry against the chance that a "future pass" arrives. If you would rather I respect the not-worth-a-round-trip call in future, say so and I will.

**The fix** (`limitations` on the stretcher model):

> ~~Because **both available rodent models** fail at the same point, the step from Schwann-cell NDRG1 loss to axonal degeneration currently has no model system at all~~
> Because **both models compared in that study** fail at the same point — and so do the third rodent model and the canine one curated in this entry — the step from Schwann-cell NDRG1 loss to axonal degeneration currently has no model system at all

You were exactly right about the diagnosis. That sentence was written when the entry curated two rodent models; adding the Okuda knockout made it stale without making it look stale. Scoping it to the head-to-head comparison in PMID:21303696 keeps it accurate to its evidence, and naming the other two preserves the global conclusion rather than weakening it — all four models really do fail at that step, which is the point of the `HUMAN_MODEL_MISMATCH` discussion.

This is the same defect class as the stale `notes:` paragraph in your first review: **prose that was true when written and became false when content was added around it.** Two instances in one entry is a pattern rather than an accident, so I audited the other three entries in this batch for count and scope claims that could have drifted the same way, checking each against the actual parsed content:

```
OK  CMT2T   : inheritance blocks = 2        vs claim "two inheritance blocks"
OK  MC1DN24 : phenotypes = 1                vs claim "exactly one phenotype"
OK  MC1DN24 : HYPOTHETICAL nodes = 2        vs claim "Those two nodes"
OK  COXPD42 : treatments block absent       vs claim "curates no treatment"
```

All four hold, so CMT4D was the only stale one. Separately, the same audit against the *literature* (rather than the entry) did find two false claims in #9717, which are fixed there.

**On the GeneReviews tag** — agreed, keeping `PMID:20301532`. Tagged-and-uncitable is the honest state, and your framing is better than mine: dropping it would hide that CMT4D has no chapter of its own.

**Validation on the new head.** Rebased onto current `main` (was 5 behind); cache diff strictly additive at six rows, all used by the entry, none of main's rows touched. `just validate-disorders` (schema + terms + references) ✓ · `count-verified-snippets` **88/88** ✓ · `check-duplicate-keys` ✓ · `check-cache-order` ✓ · `check-term-cache-integrity` ✓ · `check_folded_hyphens` ✓. The commit touches one `description` block: 0 `snippet:` lines and 0 cache rows changed, so the snippet and reference-title ratchets cannot regress — I did not re-run those two to completion, and am saying so rather than implying I did.

Re-requesting review.

kevinschaper:
Small correction upward on my own validation note: `just check-reference-titles` **did** finish, on this exact content, and passed —

```
OK: no new reference titles below similarity 0.85 (250 occurrence(s) of 107 distinct title(s) grandfathered in baseline).
```

It was still running when I wrote the previous comment, so I flagged it as not-run rather than implying it had passed. It has now completed on the same rebased tree that was pushed. `check_snippet_length` remains the one I did not run to completion; the commit touches 0 `snippet:` lines, so it cannot regress, and CI will run it regardless.

Noting this only because under-claiming is still a mis-statement of what was checked, and the whole point of the discipline is that the record matches what actually happened in both directions.

kevinschaper:
You are right, and the diagnosis is exact: the fix scoped the stale claim correctly and then overshot by one clause. Fixed in `bef3d81`.

I checked the canine source before rewording rather than taking the finding on trust:

```
"Severe axonal atrophy went along with paranodal myelin sheath instability"
"At advanced disease stages mixed nerves show a drop out of large myelinated A(alpha) fibres"
"many large myelinated A(alpha) fibres showed ovoid formation consistent with
 Wallerian degeneration stage I to III"
"Variable degrees of neurogenic atrophy were present within skeletal muscle samples"
```

The Greyhound plainly has axonal pathology. Since the link's `target` is `Early Severe Axonal Loss`, my clause asserted the opposite of what the paper reports.

**Two changes, not one.** I took your suggested rodent-scoped wording for the `limitations` field, since it matches the node description's "All three published rodent models" and is the concise correct statement. But I also took the second option you offered for the discussion rationale, because that is where there is room to put the canine on its actual footing rather than just omitting it:

> No rodent model reconstructs the step … The Greyhound is not in that position — it has prominent axonal pathology, with severe axonal atrophy, dropout of large myelinated fibres, Wallerian degeneration and neurogenic muscle atrophy. What has not been shown for the canine model is the feature that actually defines the human lesion: that the axonal loss is early and disproportionate to the demyelination rather than a late consequence of it. Nobody has assessed it for that, which is a different and weaker statement than failing to reproduce it.

That resolves the incoherence you identified as a separate consequence: the entry proposes the Greyhound line as the long-axon test with axonal counts as the primary readout, which made no sense if the model had already failed that question. The rationale now says explicitly that the canine line is where the axonal arm might be testable.

**On your closing note** — I would not describe it as your lapse. The pre-existing claim was vague enough to be unfalsifiable; making it explicit is what exposed it, and the reason it was catchable at all is that the explicit version named a source you could check. That is the change working as intended, and I would rather write claims that can be caught wrong than claims too soft to test.

Worth recording that this is the same failure shape as the original stale sentence, one step along: prose that was true of a two-model entry, rewritten for a four-model entry, and over-generalised in the rewrite. Fixing a scope error by widening scope is its own trap.

**Verified on the new head:** `just validate-disorders` ✓ · `count-verified-snippets` **88/88** ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓. Also rebased onto current `main` — this branch was 38 behind and its `test (3.13)` was failing the phantom environmental-evidence check on `Pulmonary_Embolism.yaml` and others it never touched. That gate now passes locally against the current baseline.

The `entry_type: SUBTYPE` question from `main`'s enrich-stubs pass is still open and still with @cmungall; I have not resolved it in either direction.

Re-requesting review.

kevinschaper:
Right again, and the diagnosis is precise: a question presupposes its premise, so the `prompt` was asserting exactly what the previous commit retracted — while contradicting the rationale fifteen lines below it. Fixed in `31ecc9c`.

Rewritten close to your suggestion, splitting the two cases the way the rationale does:

> Why does none of the three NDRG1-deficient rodent models reproduce the early severe axonal loss that actually causes disability in patients, when all three reproduce the demyelination — and has the canine model, which does show axonal pathology, ever been assessed for whether its axonal loss is early and disproportionate to the demyelination?

**This time I grepped instead of editing only the field you named**, which is the actual lesson here:

```
$ grep -nE "all four available|four available NDRG1" <entry>
  (none remaining)
```

That is what I should have done on the previous round. Fixing the two locations a reviewer happens to name left the file **internally inconsistent**, which is worse than the uniform error it replaced — before that commit all three places were consistently wrong about the canine, and afterwards the most visible one was the lone holdout. When retracting a claim, the unit of work is the file, not the line numbers in the review.

**Your 🔵 note taken as well.** "All have been characterised in detail" now reads "all have had their **myelin phenotype** characterised in detail", with the point made explicit: the axonal arm is the one not characterised to that standard, which is the gap the discussion exists to record. That removes the odd reading against "Nobody has assessed it for that" without weakening either statement.

**On your own part** — I would not count it against you. You named the two fields that made a checkable assertion; the `prompt` smuggled the same claim in as a presupposition, which is a different and quieter shape. Catching it on the round where it became self-contradictory is the system working, not a missed pass.

**Verified on the new head:** `just validate` ✓ **88/88** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓. Prose-only: 0 `snippet:` lines and 0 cache rows touched, so the snippet, reference-title and folded-hyphen ratchets cannot regress. Rebased onto current `main`.

The `entry_type: SUBTYPE` question remains open and with @cmungall — untouched in either direction.

Re-requesting review.

kevinschaper:
Both blocking phenotypes added, and I swept the rest of the file rather than stopping at the two you named — this pattern has now recurred often enough that fixing only the flagged lines is the wrong unit of work.

**Your two:**
- **Loss of ambulation** (`HP:0002505`) — you are right that it was captured nowhere structurally; there is no `progression:` or `clinical_burden:` section, as you checked.
- **Steppage gait** (`HP:0003376`) — added; agreed it is the better-supported binding over `HP:0001288`.

**Three more the sweep found**, by checking all 92 snippets against the curated phenotype list:

- **Bilateral vocal cord paralysis** (`HP:0001605`) — quoted in the entry's own evidence and named in the bulbar description, never curated. Curated in its own right because it is what caused the respiratory failure, rather than respiratory muscle weakness — which is precisely why the entry's diagnosis guidance asks for vocal cord examination and not only pulmonary function testing.
- **Respiratory insufficiency** (`HP:0002093`) — the *indication for a treatment this entry already curates*. A treatment whose indication is missing from the phenotype list is a gap, not a simplification.
- **Infratentorial and cervical cord atrophy** (`HP:0001272`) — this one was a **scope defect**, not an omission, and it is the most interesting of the three. `Central Nervous System White Matter Abnormality` described *both* subcortical white-matter change *and* cerebellar/medullary/cord atrophy, while carrying one binding — `HP:0002500`, cerebral white matter — that fits only the first. Atrophy of the cerebellum and cord is neither cerebral nor a white-matter change, so **the annotation was narrower than its own description**, which is the failure mode the KGX export punishes: it carries the term, not the prose. Split into two phenotypes; the medullary and cord components stay in prose rather than being approximated by a term that does not fit them.

**On your `HP:0003431` suggestion** — taken in the previous commit, for the reason you gave: the degree of slowing is near-diagnostic and is now queryable rather than living only in `diagnosis:` prose.

**One thing I fixed unprompted while adding loss of ambulation.** Both the clinical-endpoint node and my own first draft said ambulation "is commonly lost in **early adult life**". Neither the adverb nor the age range is supported — a grep across every cached reference returns exactly two mentions of ambulation, both in PMID:39201732, and the only per-patient datum is one man who had lost it by 23 while six others had not. One patient is not a basis for a frequency adverb or an age range. Both places now state the outcome without quantifying it, and the phenotype description says why it is unquantified.

**On GeneReviews** — agreed, and thank you for putting the reasoning on the record rather than leaving it looking like an oversight. `PMID:20301532` stays tagged and unmined. (Note a sibling reviewer reached the same conclusion on #9699 for the same chapter, so the treatment is now consistent across both entries.)

**Not done:** `PMID:11053686` and `PMID:25108819` remain unused. Both are secondary and I have no specific claim that needs them; adding references without a claim to attach them to is the inverse of the problem we have been fixing.

**Verified:** `just validate` ✓ **95/95** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓ · rebased onto current `main`; cache diff unchanged at 6 rows, since all five new terms were already cached and enum-valid.

Re-requesting review.

kevinschaper:
CI went red on the environmental-evidence gate. **Phantom failure — rebased, now clean.**

The five findings were in files neither PR touches:

```
kb/disorders/Metastatic_Melanoma.yaml:environmental[0]: 'Ultraviolet radiation exposure'
kb/disorders/Metastatic_NSCLC.yaml:environmental[0]: 'Tobacco smoking'
kb/disorders/Oral_Cavity_Squamous_Cell_Carcinoma.yaml:environmental[0]: 'Tobacco Use'
kb/disorders/Oral_Cavity_Squamous_Cell_Carcinoma.yaml:environmental[1]: 'Ethanol Exposure'
kb/disorders/Paranasal_Sinus_Squamous_Cell_Carcinoma.yaml:environmental[1]: 'Tobacco Smoking'
```

The cause is unusually legible this time. `main`'s newest commit is `bb445f42ce` — *"feat(environmental): cite evidence for 5 bare exposures across 4 entries (#8296 tranche 13)"*. It added evidence to **exactly those five exposures**. This branch was one commit behind, so it carried the pre-evidence copies, and the gate diffed them against main's updated baseline and read them as newly uncited. I confirmed all three files carry evidence on `main` (`with-evidence=1`, `1`, `3`).

The baseline count in the CI log corroborates it: CI grandfathered **6** distinct exposures where my pre-rebase local run saw **11**. Different baselines, same branch — that divergence *is* the staleness.

After rebasing, the gate matches CI's baseline exactly:

```
OK: no new evidence-free `environmental:` exposures (6 occurrence(s) of 6 distinct exposure(s) grandfathered in baseline)
```

Rebase was clean, zero conflicts, cache diff unchanged and additive. `just validate-disorders` passes.

**Worth recording, because it sharpens a rule I got wrong earlier in this PR:** I had reasoned that being behind was safe as long as CI was observably green. It is not, and this is the cleanest possible demonstration — the breaking commit was one that *added* evidence to unrelated entries. Nothing about my own diff could have predicted it. The exposure window is "main touches any file a diff-against-ref gate reads", which is unknowable from the branch side, so the only safe posture is to stay current rather than to check and defer.

kevinschaper:
All three fixed, plus both actionable non-blocking items. Every term verified independently before use rather than taken from the review on trust.

**Item 1 is the one that matters, because it exposes how I was searching.** This was the **fourth** location of the canine claim. After the last round I grepped for `"all four available"` and `"four available NDRG1"` and reported zero remaining — but this instance reads *"like every other NDRG1 model"*, which those patterns never touch. **Grepping for the phrasing I happened to have used is not the same as grepping for the claim.** I widened it this time (`every other|all (four|the) .*model|any model|no model|every NDRG1|all NDRG1`) and it also caught the last globally-quantified sentence at line ~575 — *"never been reconstructed in any model"* — which I have scoped to rodents with the canine exclusion stated.

**Item 2** — rebound to `HP:0012820`. You are right that the node name, `preferred_term` and source all say bilateral while the binding threw the laterality away, and that HPO codes it deliberately (`HP:0008757` is the unilateral sibling). Same class as the claw-hand fix.

**Item 3** — you were right that the rationale was false, and that is the worse half. `HP:0010873` "Atrophy of the cervical segment of the spinal cord" is an exact fit for the quoted *"upper cervical spinal cord"*, not an approximation. Split into two phenotypes sharing the one verified snippet, and the false "no term fits" sentence replaced with what is actually true — the medullary component stays in prose because it is the weakest of the three in the source, not because no term exists. I verified all three terms against the HPO API first.

**Both non-blocking items taken.** The Okuda model now carries a real `FAILS_TO_RECAPITULATE` link against `Early Severe Axonal Loss`, evidenced from PMID:21303696 — whose *"Neither model"* is precisely this hypomorph plus stretcher. As you predicted, that **fixes item 1 by construction**: the claim is now a typed link with a citation instead of prose inside a myelin link's `limitations`, so it cannot drift out of scope again.

The history record is corrected too. You are right that append-only does not yet apply — it is unmerged and part of this PR's own diff, so it is not a record of anything yet. Fixed the pre-retraction "all three NDRG1-null animal models" framing and the stale 85/85 count (now 97/97).

**One incidental find.** Rebasing (this branch was 29 behind) dropped a stale `HP:0008596` row it was carrying — same CURIE and label as main's, older `retrieved_at`, unused by this entry. Same shape as the `HP:0005791` row from an earlier round. I checked it against main before letting it go rather than assuming it was mine.

**Verified:** `just validate` ✓ **97/97** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check-cache-order` ✓ · `check-term-cache-integrity` ✓ · `check_folded_hyphens` ✓ · `validate-history` ✓. Cache diff is 10 rows, all used by this entry.

Re-requesting review.

kevinschaper:
This PR had gone `CONFLICTING` at **87 commits behind**, which blocked the auto-merge sweep — every other precondition was already met (approved, not draft, unassigned, all checks green, past the 3-day age gate). Rebased; now `MERGEABLE` again. The approval was dismissed by my own push, which is branch protection working correctly.

Two things in those 87 commits needed real work, not just a replay.

**1. `supports: PARTIAL` was retired.** `9cb071deca` narrowed `EvidenceItem.supports` to SUPPORT / REFUTE / NO_EVIDENCE and added an optional `directness` slot. This entry had four PARTIAL items, all now failing schema validation. I migrated them per that commit's own taxonomy rather than blanket-replacing, because it distinguishes cases and one of mine genuinely differs:

| site | new value | why |
|---|---|---|
| stretcher axonal-loss | `SUPPORT` / `INDIRECT` | sentence is about the mice; confirms the human feature in passing |
| scoliosis 1-of-7 cohort | **`REFUTE`** / `DIRECT` | contradicts the `FREQUENT` band directly |
| CNS "should be further studied" | `SUPPORT` / `INDIRECT` | authors point at the claim without asserting it |
| Greyhound "mixed polyneuropathy" | `SUPPORT` / `INDIRECT` | right general lesion, wrong specific one |

The scoliosis one is the interesting change: PARTIAL had left its *direction* ambiguous, and `REFUTE` says plainly that the cohort contradicts the frequency band while the phenotype itself stays supported by the literature review. The disagreement earlier reviews asked me to preserve is now machine-readable rather than only explained in prose.

On the stretcher item I deliberately did **not** split into a SUPPORT/REFUTE pair as the schema comment suggests for two-claim sentences. Its second claim — that the rodent models fail to reproduce the lesion — is already carried structurally by the `FAILS_TO_RECAPITULATE` links on those models, added earlier in this PR. Splitting would duplicate as prose-with-a-quote something the pathograph already holds as typed data. Happy to split it if you would rather have the symmetry.

**2. The cache conflict, resolved main-authoritative per CURIE.** I took main's `cache/mondo/terms.csv` and `diseaseterm` enum wholesale, then re-derived `MONDO:0011085` through `just validate` rather than hand-adding it.

That dropped **fifteen rows this branch had been carrying that were never its to add** — `TRIP11`, `ACP5`, `HHAT`, `FN1`, `PCYT1A`, `PLCB3`, plus several spondylometaphyseal-dysplasia and odontochondrodysplasia terms, all timestamped before this curation began. They came in with the branch's original base and main has since dropped them. Same shape as the stale `HP:0005791` and `HP:0008596` rows from earlier rounds, but far more of them. **Cache diff is now 8 rows, every one used by this entry.**

**Verified after the rebase:** `just validate` ✓ **97/97** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check-reference-titles` ✓ · `check-title-snippets` ✓ · `check_folded_hyphens` ✓ · `check-entity-refs` ✓ · `check-duplicate-keys` ✓ · `check-cache-order` ✓ · `check-term-cache-integrity` ✓ · `validate-history` ✓.

Re-requesting review.

PR #9699, opened 2026-08-27
curate: Add Charcot-Marie-Tooth disease, axonal, type 2T (MONDO:0044640, MME)

### LINKED ISSUE #9683: Curate Charcot-Marie-Tooth disease axonal type 2T (MONDO:0044640)
Curate a dismech entry for **Charcot-Marie-Tooth disease axonal type 2T** ([MONDO:0014866](https://monarchinitiative.org/MONDO:0014866)) — an MME-related axonal neuropathy; neprilysin loss causes a late-onset axonal CMT, with both recessive and dominant reports.

Note the inheritance is contested in the literature (recessive and heterozygous dominant reports) — curate what the sources support rather than picking one.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Axonal_Type_2T.yaml`
Nominated by: unknown

kevinschaper:
Adds a knowledge-base entry for Charcot-Marie-Tooth disease type 2T, caused by loss-of-function variants in MME, the gene for the metalloprotease neprilysin. Two things about it are unusual and shaped how this entry is written. It starts in the fifth or sixth decade, so a large share of what looks like "idiopathic axonal neuropathy of the elderly" is actually this disease — MME accounted for 34.8% of genetically solved cases in one 230-patient series. And its inheritance is genuinely unsettled: recessive and incompletely-penetrant-dominant accounts are each supported by substantial independent cohorts, so the entry curates both rather than picking one.

## The MONDO term in the claim issue is obsolete

The claim (#9683) names **MONDO:0014866**. That term is obsolete in the MONDO release the term validator uses, and `just validate` refuses it. It is replaced by **MONDO:0044640** "Charcot-Marie-Tooth disease type 2T", which this entry uses.

The obsoletion is substantive rather than housekeeping. MONDO:0014866 carried **two** OMIM cross-references (616233 and 617017), listed DNAJB2 synonyms, and named MME as its causal gene — it had merged two different gene-disease pairs that share the label "CMT2T". MONDO:0044640 carries only OMIM:617017 and the MME synonyms. This is recorded in the entry's `notes:` and as a differential-diagnosis entry, because anyone reading a pre-split citation or ontology release has to check which gene a "CMT2T" reference means.

Two tooling notes fall out of this, both reproducible:

1. The local `sqlite:obo:mondo` build and the OLS-backed validator **disagreed**. Locally MONDO:0014866 was still active, with the DNAJB2 synonyms and both OMIM xrefs. The stub queue was built from the older view.
2. Consequently `just preflight-dr` returns **SKIP** against the correct current term — in the older local build, MONDO:0044640 exists only as an Orphanet-derived shell with no OMIM xref and no `RO:0004003` causal gene, so the gene-identity check has nothing to compare against.

## NEC check

Automated: **SKIP** against MONDO:0044640 for the reason above; **PASS** against MONDO:0014866 (MME = 39 mentions, no rival gene).

Manual check per the SKIP path, using the current MONDO:0044640 record: its definition names MME, its OMIM xref is 617017, the report cites no conflicting OMIM ID, and DNAJB2 does not appear among the report's genes at all. The report is about the intended entity. Its own reference validation found 6 of 6 references resolved, none unresolved. It was used as a lead source only — every cited PMID was fetched independently and every snippet checked against cached text.

## Mechanism, and where it stops

```
MME loss-of-function variant
  └─> Neprilysin deficiency in peripheral nerve
        └─> Loss of peptidase activity at myelin-associated compartments   [HYPOTHETICAL]
              ├══> Length-Dependent axonal degeneration      (unknown intermediates)
              │      └─> Late-onset axonal sensorimotor polyneuropathy
              └══> Predisposition to superimposed inflammatory demyelinating neuropathy
                     └══> Late-onset axonal sensorimotor polyneuropathy
```

6 nodes, 6 edges, one root, all reachable (measured). Conformance to three `peripheral_axonal_degeneration` nodes.

The node in the middle is marked `HYPOTHETICAL` and its main outgoing edge typed `INDIRECT_UNKNOWN_INTERMEDIATES` on purpose. Neprilysin deficiency in patient nerve is well established, axonal degeneration in patient nerve is well established, and **nothing published connects them** — the authors of PMID:34307994 say so directly: "contrary to the glomerular disease, which is immune mediated, the pathophysiology of the neuropathy is unknown."

Both obvious candidate mechanisms have been tested and neither survived, and both negatives are curated rather than omitted:

- **Beta-amyloid is not the route.** Neprilysin is a principal amyloid-degrading enzyme, so early-onset Alzheimer disease is the natural prediction. Two independent cohorts found neither dementia nor excess amyloid. Curated as an `EXCLUDED` phenotype with `REFUTE` evidence.
- **A remyelination role is not the route.** MME-null mice regenerate and remyelinate indistinguishably from wild type. Curated as two `FAILS_TO_RECAPITULATE` model links.

## Both inheritance modes, deliberately

Higuchi et al. found biallelic variants in 10 patients and named the disease AR-CMT2T. Auer-Grumbach et al., months later, found heterozygous variants in 19 index subjects segregating dominantly with age-related incomplete penetrance. A later 230-patient cohort found dominant transmission the commoner pattern while still recording recessive families.

A simple dosage model would predict biallelic patients are more severely affected. The 2024 Italian series tested that and found **comparable severity**, differing only in onset age and later progression rate — and raised a third possibility neither original study framed, that heterozygous variants are a *risk factor* rather than a monogenic cause. That series also classified 10 of its 32 patients' variants as VUS and one as likely benign under ACMG criteria, so part of the monoallelic signal may be variant-interpretation noise.

The entry therefore carries two `inheritance` blocks and leaves `zygosity` **unset** on the trigger node's `genetic_context`. Setting it would encode one side of an open question. This is written up as a `KNOWLEDGE_GAP` discussion with two proposed experiments, the key one being penetrance measurement in biobank-ascertained rather than clinic-ascertained carriers — clinic ascertainment cannot distinguish incomplete penetrance from susceptibility.

## Two other judgement calls

**CIDP is curated as a comorbidity, not only a differential.** A homozygous MME patient developed demyelinating electrophysiology and an inflammatory biopsy and improved immediately on immunoglobulin. This rests on one patient, and it is curated anyway because it changes management: a demyelinating feature in a genotyped patient may be a treatable second process, and attributing the deterioration to the genetic disease forecloses that. IVIg is curated as a treatment with `target_mechanisms` pointing at the inflammatory node only, with the description stating explicitly that it does not modify the genetic disease.

**Feto-maternal alloimmunisation is kept out of the pathograph.** Neprilysin-null women immunise against the antigen in pregnancy and their infants can be born with antenatal membranous nephropathy — this is how the first four families were found. It is a real consequence of the same genotype, but the disease occurs in the offspring and is immune mediated, so it is recorded in a discussion with its screening implications rather than as a node. That discussion also carries the authors' pharmacovigilance question about long-term neprilysin inhibition in heart failure, with their own caveat that no harm has been observed.

## Validation

- `just count-verified-snippets` — `Snippets checked: 56/56 verified against cached references`
- `just validate-disorders` (batched CI equivalent) — schema, term and reference validation all passed, 56/56
- `just check-duplicate-keys`, `just check-title-snippets`, `check_snippet_length --against-ref origin/main`, `check_folded_hyphens --check`, `just check-cache-order`, `just check-term-cache-integrity`, `just validate-history` — all OK

A separate audit confirmed all 10 `reference_title` values byte-match the `title:` line in their cache frontmatter.

## Not done

`HP:0003581` (Adult onset) sits under Clinical modifier rather than Phenotypic abnormality and so falls outside the `PhenotypeTerm` enum. The onset phenotype carries no `term:`; the structured `onset` descriptor (`MIDDLE_AGE`) holds the information instead, and a `notes:` line records why.

No Orphanet evidence: `references_cache/ORPHA_495274.md` does not exist and `just refresh-orphadata` is broken on a drifted sha256.

Closes #9683

### DIFF
--- a/kb/disorders/Charcot-Marie-Tooth_Disease_Axonal_Type_2T.yaml
+++ b/kb/disorders/Charcot-Marie-Tooth_Disease_Axonal_Type_2T.yaml
@@ -715,6 +715,55 @@ phenotypes:
     explanation: >-
       Documents distal weakness and atrophy in the defining series.
 - category: Neurologic
+  name: Foot Drop
+  description: >-
+    Weakness of ankle dorsiflexion, reported across the cohort followed over time
+    and the mechanical cause of both the steppage gait and the falls. Curated in
+    its own right because it is the indication for the foot orthosis this entry
+    curates as a treatment - a treatment whose indication is missing from the
+    phenotype list is a gap rather than a simplification. Bound to Foot
+    dorsiflexor weakness, which is the impairment; the term "foot drop" names the
+    resulting posture and has no separate HPO code.
+  phenotype_term:
+    preferred_term: Foot drop
+    term:
+      id: HP:0009027
+      label: Foot dorsiflexor weakness
+  evidence:
+  - reference: PMID:34307994
+    reference_title: "Antenatal Membranous Nephropathy and Type 2 (Axonal) Charcot-Marie-Tooth With Mutations in the Metallo-Membrane Endopeptidase Gene: A Call for Family Screening and Pharmacovigilance."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      The patients suffered from vasomotor painful episodes and progressive gait
+      disturbances and muscular weakness of the lower limbs with foot drop and
+      frequent falls
+    explanation: >-
+      Documents foot drop across the cohort, alongside the gait disturbance and
+      falls it produces.
+- category: Neurologic
+  name: Exercise Intolerance
+  description: >-
+    Recorded as a first neurological symptom for one genotyped patient, alongside
+    lower limb cramps. Curated because it is quoted in this entry's own evidence
+    and is separately coded; no frequency, since it rests on a single cell of a
+    seven-patient table and the larger Italian series does not report it.
+  phenotype_term:
+    preferred_term: Exercise intolerance
+    term:
+      id: HP:0003546
+      label: Exercise intolerance
+  evidence:
+  - reference: PMID:34307994
+    reference_title: "Antenatal Membranous Nephropathy and Type 2 (Axonal) Charcot-Marie-Tooth With Mutations in the Metallo-Membrane Endopeptidase Gene: A Call for Family Screening and Pharmacovigilance."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      Lower limb cramps and exercise intolerance
+    explanation: >-
+      The first-symptom cell for a genotyped patient, naming exercise intolerance
+      alongside the cramps curated separately.
+- category: Neurologic
   name: Steppage Gait
   description: >-
     The gait abnormality produced by distal lower-limb weakness and foot drop,


kevinschaper:
All three IMPORTANT findings are fixed, plus three of the four suggestions. Every one was correct, and the first was the worst kind of gap — the single most frequent sign in the disease, missing from an entry that had the quote sitting in its own cache.

**1 — reflexes.** Curated as `Reduced or Absent Tendon Reflexes`, `VERY_FREQUENT` from the 91% figure.

One judgement call worth flagging since you named both terms as available: I bound **`HP:0001265` Hyporeflexia, not `HP:0002522` Areflexia of lower limbs**. The 91% pools *"absent **or** reduced"*, and Areflexia is a descendant of Hyporeflexia in the HPO — so attaching the narrower term to that number would assert abolished reflexes in patients whose reflexes were merely diminished. The upper-limb 56% is recorded in the description, and the individual-patient areflexia you pointed to in PMID:34307994 (*"abolished ankle reflexes"*) is curated as a **second evidence item on the same phenotype** rather than as an overlapping annotation, so it is not lost. If you would rather see a separate `HP:0002522` phenotype scoped to the individually-examined patients, say so and I will add it.

**2 — frequencies, each banded from its own quoted denominator:**

| Phenotype | Figure | Band |
|---|---|---|
| Reduced/absent tendon reflexes | 91% LL (Italian, n=32) | `VERY_FREQUENT` |
| Distal muscle weakness | 69% moderate-severe LL | `FREQUENT` |
| Distal sensory impairment | 62% pinprick / 63% light touch | `FREQUENT` |
| Pes cavus | 2/7 = 29% | `OCCASIONAL` |

I quoted **both** sensory modalities rather than picking one, because the source reports them separately and they agree. For pes cavus the description now states the denominator is seven and that no larger series reports it separately, so the band reads as the only quantified figure available rather than a stable expectation. Limb pain, cramps and paresthesia are curated **without** frequencies — the source describes those narratively and the Italian series does not count them, which is exactly the "keep omitting where the number genuinely is not there" line.

**3 — supportive management.** You were right that the entry contradicted itself: it claimed CMT2T "has none" while quoting orthosis use in a phenotype description two hundred lines earlier. `Orthotic Supportive Care` added (`NCIT:C15747`, following `Charcot-Marie-Tooth_Disease_Type_2.yaml`), evidenced by the genotyped patient whose foot drop and falls *"was improved by foot orthosis"* plus the cohort rows recording plantar orthoses. The IVIg description now reads "no **disease-modifying** treatment", which is what it meant.

**Suggestions.** `GO:0008233` → `GO:0004222`, so specificity no longer regresses below the node above it — good catch, that was a real modelling defect and not just tidiness. Limb pain, muscle cramps and paresthesia curated. Nerve conduction studies added to `diagnosis:` with the 38 m/s axonal-versus-demyelinating criterion, which is the step that puts a patient in the CMT2 branch before any sequencing.

**GeneReviews** — agreed, and I have deliberately left `PMID:20301532` alone. Its cached record is `abstract_only` and carries nothing but a statement of the chapter's purpose, with no CMT2T or MME content; any snippet from it would be a title standing in for a finding.

Also rebased onto current `main` (was 3 behind). The cache diff is strictly additive — 3 rows, no rows of main's touched.

**Validation after the fixes:** `just validate` OK · `just count-verified-snippets` **70/70** · `just validate-terms` OK · `check-reference-titles`, `check-title-snippets`, `check_snippet_length`, `check_folded_hyphens`, `check-cache-order`, `check-term-cache-integrity`, `check-duplicate-keys`, `validate-history` all OK. Pathograph re-measured: 6 nodes, 6 edges, one root, all reachable. All 10 `reference_title` values still match their cache frontmatter.

Re-requesting review.

kevinschaper:
You were right, and the error was worse than a misdescription — it inverted its own rationale. Fixed in `4f9f0c2`.

I verified the parentage against the JAX HPO API rather than taking three reviews on trust:

```
HP:0001265 Hyporeflexia -> parents: [HP:0001315]
HP:0001284 Areflexia    -> parents: [HP:0001315]
HP:0002522 Areflexia of lower limbs -> parents: [HP:0002814, HP:0001284]
HP:0001315 -> "A reduction (hyporeflexia) or complete absence (areflexia) of the
              involuntary muscle contraction..."
```

Siblings, both direct children of `HP:0001315`. And since `HP:0001265` is defined as *"Reduction of neurologic reflexes"* — a reduction, not an absence — binding it to a figure pooling "absent **or** reduced" **dropped the areflexic share**, which is precisely the error my sentence claimed to be avoiding. Rebound to `HP:0001315`, whose definition is the pooled claim verbatim and which is scoped to tendon reflexes, matching the snippet's "deep tendon reflexes (DTRs)". `VERY_FREQUENT` and both evidence items unchanged.

**One detail worth flagging, since you quoted the label.** The cached canonical label is `Reduced tendon reflexes`; the live JAX API now returns `Diminished deep tendon reflex`. The validator is cache-first, so the cached form is what validates and is what I used — but the two have drifted, and anyone checking this binding against a live HPO lookup will see a mismatch that is not an error. No new cache rows were needed, as you said.

**All four non-blocking items are done too:**

- Renamed to `Distal Muscle Weakness` and reworded the description, finishing the split from `828a2aa`. You were right that it read as the pre-split name.
- `HP:0009763` → `HP:0012514` Lower limb pain.
- Orthotic care now carries `target_mechanisms`, so the only evidenced management for this disease is on the pathograph instead of IVIg alone. `treatment_effect: BYPASSES` — the enum's definition, *"works around the disrupted mechanism via an alternative pathway"*, is exactly what an orthosis does: it substitutes externally for lost dorsiflexion rather than acting on the degeneration. Attached to the clinical endpoint, not an upstream node, so it cannot be read as disease-modifying.
- CMTES captured (mean 10, range 1–21) on the endpoint node — the numeric counterpart of its "relatively mild" description, and the range reaching 21 shows the spread extends to severe.

**This is the fourth unverified assertion of mine you or a colleague have caught in this batch**, and the first about ontology structure rather than literature or corpus. The pattern is now unambiguous: every one was a factual claim in prose that no gate reads, and every one would have taken one lookup. `just validate-terms` confirms a CURIE exists and its label matches — it says nothing about a *claim I write about the term's position in the hierarchy*, which is exactly the kind of statement that reads as authoritative to the next curator.

Also rebased onto current `main` — the branch had gone 1 behind mid-edit, which showed up as `MONDO:0014733` appearing as a deletion in my cache diff.

**Verified:** `just validate` ✓ **72/72** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓ · reference-title audit 10/10 ✓ · pathograph 6 nodes / 6 edges / single root / fully reachable · cache diff 3 rows, additive.

Re-requesting review.

kevinschaper:
Both blocking items fixed, plus three of the four suggestions.

**Item 2 is the one worth dwelling on, because you caught me doing the same thing twice.** The quoted "definition" of `HP:0001315` came from the **live JAX API**, which does carry that string — but neither source this repo validates against does. I checked the repo's own adapter before changing anything:

```
$ runoak -i sqlite:obo:hp info HP:0001315 -O obo
def: "Diminution of tendon reflexes, which is an invariable sign of peripheral
      nerve disease." [orcid:0000-0002-0736-9199, PMID:20941667]
```

And the provenance is worse than that: **I took the string from your previous review** and propagated it without verifying it against the sources the entry is validated against. Replacing one unverifiable HPO assertion with another, inside quotation marks, in the same paragraph that had already burned three review cycles, is a worse error than the original. The paragraph now rests only on independently checkable facts — parent of both children, the two are siblings, tendon-scoped. Term choice unchanged and still correct.

**Item 1** — `PMID:20301532` tagged with `tags: [GeneReviews]`, not mined. Your reading matches what I found in the cache: purpose statement only, so any snippet would be a title doing a snippet's job. Following the 10 sibling entries.

**Suggestion 3 taken, and it is the most valuable thing in this round.** Lupo et al. 2018 is `PMID:30415211` and was not cached; now fetched and curated. You were right that it is the strongest argument against monoallelic causation — 197 index cases, heterozygous variants in 11 families whose phenotypes did not converge (CMT1, CMT2, dHMN, fALS), and the conclusion that they *"could not demonstrate that MME variants in heterozygosis cause neuropathy"*. I curated it as two **`REFUTE`** items on the dominant inheritance block rather than as prose, so the challenge is structurally visible in the export, and worked it into the `cmt2t_dominant_versus_recessive` discussion as the fourth cohort. This genuinely changes the weight of that discussion — it was previously three cohorts arguing for a dominant mode with one hedge; now there is a direct published challenge on the other side.

One snippet mechanic worth recording: the sentence I wanted contains `[CMT type 1 (CMT1), CMT2, dHMN and fALS]`, which the matcher strips as a non-literal bracket span, so I quoted up to the bracket and let the conclusion sentence carry the segregation claim. Noted in the `explanation` so it does not read as a truncation.

**Suggestions 1 and 2** also done — `below` → `above`, and steppage gait (`HP:0003376`) curated from the first-symptom cell of a genotyped patient in PMID:34307994.

**Suggestion 4 not done, deliberately.** There is no `omim_mappings` slot — `DiseaseMappings` carries only ICD-10-CM, ICD-11, MONDO and NCIT, as `Bleeding_Disorder_Platelet-type_21` documents. An OMIM identity would belong in `external_assertions`, which is a different change from the one suggested, so I left it rather than improvising a mechanism into a PR that is otherwise ready.

**Verified:** `just validate` ✓ **75/75** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓ · rebased onto current `main`. Cache diff still 3 rows plus the new PMID cache file.

Re-requesting review.

kevinschaper:
Follow-up: I ran the same quoted-but-uncurated sweep here that a sibling review prompted on #9696, rather than waiting for it to be flagged. Checking all 75 snippets against the curated phenotype list found two:

- **Foot drop** (`HP:0009027` Foot dorsiflexor weakness) — quoted three times, the mechanical cause of the steppage gait and falls already curated, and **the indication for the foot orthosis this entry curates as a treatment**. That is the identical defect the sweep found in CMT4D, where respiratory insufficiency was missing while the ventilation treating it was curated. A treatment whose indication is absent from the phenotype list is a gap. Bound to the dorsiflexor-weakness term because "foot drop" names the resulting posture and has no separate HPO code.
- **Exercise intolerance** (`HP:0003546`) — a first-symptom cell for a genotyped patient, quoted and unannotated. No frequency: one cell of a seven-patient table, and the Italian series does not report it.

Both terms were already cached and enum-valid, so the cache diff is unchanged.

The generalisable form of this, which is worth more than either finding: **a curated treatment implies a curated indication.** Both entries had a treatment whose indication existed only as prose. That is mechanically checkable in a way the other prose defects in this batch were not — a `treatments:` entry targeting a clinical endpoint whose triggering phenotype is absent from `phenotypes:` is a detectable inconsistency, not a judgement call.

`just validate` ✓ **77/77** · `just validate-disorders` ✓ · `check_folded_hyphens` ✓ · already current with `main`.

kevinschaper:
CI went red on the environmental-evidence gate. **Phantom failure — rebased, now clean.**

The five findings were in files neither PR touches:

```
kb/disorders/Metastatic_Melanoma.yaml:environmental[0]: 'Ultraviolet radiation exposure'
kb/disorders/Metastatic_NSCLC.yaml:environmental[0]: 'Tobacco smoking'
kb/disorders/Oral_Cavity_Squamous_Cell_Carcinoma.yaml:environmental[0]: 'Tobacco Use'
kb/disorders/Oral_Cavity_Squamous_Cell_Carcinoma.yaml:environmental[1]: 'Ethanol Exposure'
kb/disorders/Paranasal_Sinus_Squamous_Cell_Carcinoma.yaml:environmental[1]: 'Tobacco Smoking'
```

The cause is unusually legible this time. `main`'s newest commit is `bb445f42ce` — *"feat(environmental): cite evidence for 5 bare exposures across 4 entries (#8296 tranche 13)"*. It added evidence to **exactly those five exposures**. This branch was one commit behind, so it carried the pre-evidence copies, and the gate diffed them against main's updated baseline and read them as newly uncited. I confirmed all three files carry evidence on `main` (`with-evidence=1`, `1`, `3`).

The baseline count in the CI log corroborates it: CI grandfathered **6** distinct exposures where my pre-rebase local run saw **11**. Different baselines, same branch — that divergence *is* the staleness.

After rebasing, the gate matches CI's baseline exactly:

```
OK: no new evidence-free `environmental:` exposures (6 occurrence(s) of 6 distinct exposure(s) grandfathered in baseline)
```

Rebase was clean, zero conflicts, cache diff unchanged and additive. `just validate-disorders` passes.

**Worth recording, because it sharpens a rule I got wrong earlier in this PR:** I had reasoned that being behind was safe as long as CI was observably green. It is not, and this is the cleanest possible demonstration — the breaking commit was one that *added* evidence to unrelated entries. Nothing about my own diff could have predicted it. The exposure window is "main touches any file a diff-against-ref gate reads", which is unknowable from the branch side, so the only safe posture is to stay current rather than to check and defer.

PR #9700, opened 2026-08-27
Curate Charcot-Marie-Tooth disease type X (MONDO:0018994)

### LINKED ISSUE #9688: Curate Charcot-Marie-Tooth disease type X (MONDO:0018994)
Curate a dismech entry for **Charcot-Marie-Tooth disease type X** ([MONDO:0018994](https://monarchinitiative.org/MONDO:0018994)) — the X-linked branch of hereditary motor and sensory neuropathy, dominated by *GJB1*/connexin-32 loss of function in Schwann cells (CMTX1, ~90% of cases), with rarer loci causing CMTX2–CMTX6.

My reading, for review: this looks like a **Disease with `has_subtypes`**, matching how the KB already treats the other CMT type-series. `kb/disorders/Charcot-Marie-Tooth_Disease_Type_1.yaml` and `Charcot-Marie-Tooth_Disease_Type_4.yaml` are both multi-gene (PMP22/MPZ/LITAF/EGR2 and GDAP1/MTMR2/SBF2/SH3TC2/NDRG1/PRX/FIG4 respectively) and are curated as single Disease entries with per-gene subtypes. CMTX is the same shape, and its shared pathograph — X-linked inheritance, Schwann cell–axon communication failure, demyelination with secondary axonal loss — is at least as conserved as CMT4's.

MONDO records six descendants:

| MONDO | Label | Gene |
|---|---|---|
| MONDO:0010549 | CMT X-linked dominant 1 | *GJB1* |
| MONDO:0010550 | CMT X-linked recessive 2 | — |
| MONDO:0010551 | CMT X-linked recessive 3 | — |
| MONDO:0010689 | CMT X-linked recessive 4 (Cowchock) | *AIFM1* |
| MONDO:0010699 | CMT X-linked recessive 5 | *PRPS1* |
| MONDO:0010479 | CMT X-linked dominant 6 | *PDK3* |

Two of these already have their own entries — `Charcot-Marie-Tooth_Disease_X-linked_Recessive_4.yaml` and `PRPS1_Deficiency_Spectrum.yaml` (CMTX5) — so the CMTX entry should cross-reference rather than duplicate them.

Also noting for a separate cleanup: `Charcot-Marie-Tooth_Disease_Type_1.yaml` currently carries a `CMTX1` entry in its `has_subtypes` list. CMTX1 is X-linked, not CMT1, so that subtype is misfiled. I'll leave it alone in this PR and file it separately.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Type_X.yaml`

Phillip-a-richmond:
Adds `kb/disorders/Charcot-Marie-Tooth_Disease_Type_X.yaml` — the X-linked compartment of CMT, which the KB did not cover.

Closes #9688

> **Updated after review round 1** (`1072091c6`). Six of the seven review items are fixed; one is escalated to @cmungall and described under "Open question" below. The earlier version of this description called the CMT1 overlap a misfiling — that was wrong, and the corrected reading is below.

## The lump/split call

I curated this as a **Disease with per-locus `has_subtypes`**, not a grouping. The argument is precedent: `Charcot-Marie-Tooth_Disease_Type_1.yaml` (PMP22/MPZ/LITAF/EGR2) and `_Type_4.yaml` (GDAP1/MTMR2/SBF2/SH3TC2/NDRG1/PRX/FIG4) are both multi-gene compartments already carried this way, and CMTX is the same shape. CMTX1 is roughly 90% of it, so the shared pathograph is at least as conserved as CMT4's.

Two CMTX loci already have their own entries, so this one cross-references rather than duplicates them: CMTX4 (`Charcot-Marie-Tooth_Disease_X-linked_Recessive_4.yaml`) and CMTX5 (`PRPS1_Deficiency_Spectrum.yaml`).

## Mechanism

Built around CMTX1/GJB1. Connexin-32 forms the reflexive gap junctions in non-compact myelin that let a Schwann cell exchange ions and metabolites with its own adaxonal cytoplasm straight across the sheath. That radial path is about a million times faster than going the long way around the spiral (PMID:9722620), which is what makes losing it consequential rather than merely inefficient.

The rarer loci arrive at the same endpoint by unrelated routes and are modelled as separate upstream nodes: PDK3 hyperactivation reducing pyruvate flux, AIFM1 oxidoreductase dysfunction increasing cell death, and PRPS1 loss reducing purine nucleotide supply.

Conformance declared to `schwann_cell_myelin_maintenance` (one node) and `peripheral_axonal_degeneration` (two nodes).

## Open question for a human — CMTX1's compartment

**This is the one item I did not resolve, and it needs adjudication rather than a patch.**

Once this merges, the KB asserts CMTX1 as a `has_subtypes` entry of two different Diseases. I originally described this as CMT1 having misfiled CMTX1. Having read that entry properly, that was wrong: CMT1 argues for the placement in three places — its top-level description, the subtype entry, and an `inheritance:` block with its own cited evidence — on the grounds that the primary lesion is Schwann-cell-intrinsic and CMTX1 is "the glia-axon bridge between the demyelinating and axonal compartments."

So there are two coherent nosologies:

| Sort on | Puts CMTX1 in | Rationale |
|---|---|---|
| Primary lesion | CMT1 | The lesion is Schwann-cell-intrinsic, like PMP22 and MPZ |
| Inheritance | CMTX | CMT1/2/4/X divide on inheritance; CMTX1 is X-linked |

Deleting the CMT1 content would overwrite a curator's reasoned choice with mine on a question this PR exposes rather than settles, so I've left it and assigned @cmungall. Detail and both follow-up shapes are in #9703 (rewritten — the original text carried my wrong framing).

## What the deep research corrected

Provider was OpenScientist (26 citations, report committed under `research/`). It changed the entry in two places where my first pass was wrong:

**Female phenotypic variability is not explained by skewed X-inactivation.** I had written that it was — the standard textbook explanation. The only study to measure X-inactivation directly in affected nerve found a random pattern (PMID:40759929). The entry declines to assert mosaicism and carries the refuting evidence as `supports: REFUTE`. Two patients, so it weakens rather than settles the hypothesis, and it is written up that way.

**The axon-versus-myelin ordering is contested.** Gjb1-null mice at 2–4 months, before demyelination is detectable, already show reduced axon diameters and slowed axonal transport (PMID:20720503). Human nerve conduction studies read the other way (PMID:15468313). Recorded as a `HUMAN_MODEL_MISMATCH` discussion with both citations rather than picking a side; the mouse model is linked `PARTIALLY_RECAPITULATES` with the discrepancy named in `limitations`.

## Things checked and rejected

- **PMID:42017539** appears in the report supporting a serum NfL claim. It is about *NEFL*-associated CMT, not CMTX. Dropped rather than cited.
- **The ">90% of males" frequency** in the report's summary table traces to no quotable sentence in any cached paper, so it is not asserted. Distal weakness is banded `VERY_FREQUENT` from a real quote instead ("85% of males reported difficulty with walking compared with 64% of females"), with the sex split noted in the explanation.
- Two PMIDs recalled from memory while drafting were the wrong papers, caught by fetching before use: PMID:17470752 is a multiple sclerosis prevalence survey, and PMID:17701896 is the Arts syndrome paper rather than the CMTX5 one (PMID:17701900, adjacent in the same issue).

## A tooling gap this PR surfaced

Nothing machine-validates `EvidenceItem.reference_title`. The review found two that had drifted from the cache; while fixing them I wrote a check across all 20 PMID/title pairs, and it caught a third — a title I had invented from memory minutes earlier for a paper already in the cache. Filed as #9713. All 20 now match the cache exactly.

Also filed: #9701 (`just new-history` fails on parentheses in `--details`) and #9702 (`--validate-references` silently emits no validation block on most deep-research reports).

## Reference cache scope

The staged `references_cache/` files are the entry's own citations plus the citations of the OpenScientist report committed alongside it, since the report's citation validation resolves against them. One file fetched in error was removed before staging. The three `cache/**.csv` rows are exactly what this entry introduced: `MONDO:0018994` label and enum membership, and `hgnc:8811` PDK3.

## Validation

| Check | Result |
|---|---|
| `just validate-disorders` | passed, 60/60 snippets verified |
| reference titles vs cache | 20/20 exact match |
| `just check-entity-refs` / `check-duplicate-keys` | passed |
| snippet grading / title / length / folded-hyphen gates | no new violations |
| `just check-cache-order` / `check-term-cache-integrity` | passed |
| pytest structural and conformance | 14,604 passed |


Phillip-a-richmond:
Thanks — this was a useful review. Pushed `1072091c6` addressing all seven items. Taking item 3 differently, explained below.

## Fixed

**Item 1 — wrong `conforms_to`.** You're right, and I checked the module node before removing it: `Schwann Cell Proteostatic Stress and Myelin Protein Mistrafficking` binds `GO:0030968` and `GO:0034976`, both `INCREASED`, and my node's only process is myelin maintenance `DECREASED`. There is no proteostatic claim to conform. Dropped rather than relocated, as you suggested.

**Item 2 — bundled node.** Split into `PDK3 Hyperactivation and Reduced Pyruvate Flux` (`GO:0006090` `DECREASED`) and `AIFM1 Mitochondrial Oxidoreductase Dysfunction` (`GO:0003954` `ABNORMAL` as a molecular function, `GO:0006915` `INCREASED`), each with its own gene and its own downstream edge. You were right that the AIFM1 mechanism was asserted in prose and represented nowhere in the graph.

On your `GAIN_OF_FUNCTION` note for PDK3 — I left the pathway modifier as `DECREASED`. The `modifier` describes the pyruvate pathway's state, and that is quantitatively reduced flux, which keeps the PATO grounding. The kinase's own hyperactivity is now stated explicitly in the node description and carried by a second snippet. If you'd rather see `GAIN_OF_FUNCTION` on a separate molecular-function descriptor for the kinase itself, say so and I'll add it.

**Item 4 — reference titles.** Both fixed. This one was more useful than it looked: I wrote a check comparing every `reference_title` against its cache frontmatter, and it caught a *third* mismatch — a title I invented while drafting the fixes for this very round. All 20 distinct PMID/title pairs now match the cache exactly. Given that nothing machine-validates this field and I produced a fresh violation inside one editing session, a repo-level check seems worth having; I can open that separately.

**Item 5 — case fraction.** Moved to `Genetic.case_fractions` on the GJB1 entry with `population: Hungarian CMT cohort` and `case_fraction_percent: 9.2`, and the UNKNOWN/UNKNOWN prevalence record is gone.

**Item 6 — under-consumption.** Added a `progression:` section (onset window, and the age–disability correlation that holds in males but not females), the variant-class spectrum, the ACMG breakdown, and the c.-17G>A signal on the GJB1 entry. I also added the `PMID:28768847` finding that there is *no* general genotype–phenotype correlation, graded `PARTIAL`, since it qualifies the c.-17G>A result rather than agreeing with it.

One thing I did not do: the ">90% of males" frequency. That figure is in the research report's own summary table, not traceable to a quotable sentence in any cached paper, and asserting it would be exactly the fabrication the snippet rules exist to prevent. Instead I banded distal weakness `VERY_FREQUENT` off a real quote — "85% of males reported difficulty with walking compared with 64% of females" — and noted in the explanation that a single disease-level frequency is an approximation when the number differs by sex that much.

**Item 7 — CNS triggers.** Prose corrected to distinguish described triggers from unestablished mechanism, and added an `environmental:` entry with `influences_mechanisms` → `TRIGGERS` on the oligodendrocyte node. Graded `PARTIAL`, because the pollen paper explicitly says the allergic mechanism is unproven. No ECTO binding: the entry is a heterogeneous set of systemic stressors rather than one exposure concept, and I've recorded that in `notes` rather than leaving it looking un-researched.

## Suggestions taken

GeneReviews `references:` block; pathograph connected (both the oligodendrocyte node and the terminal node now have downstream edges to their phenotypes); `CL:0000540` → `CL:0000101` + `CL:0000100` at all four sites; gap-junction modifier → `DECREASED`; episode phenotype split into `HP:0002401` Stroke-like episode and `HP:0001260` Dysarthria; `PMID:23217327` snippet replaced with the study's actual finding about the variant.

Cochlear implantation added — and the regenerated cache turned out to carry full text showing three of the four cases are CMTX genotypes (GJB1, AIFM1, PRPS1), so it grounds on the CMTX5 case directly rather than by extrapolation from CMT generally. Good catch.

I skipped the subtype `disease_term` bindings for now. The research suggested MONDO:0010674 for CMTX1 while the stub's MONDO descendant list gives MONDO:0010549, and I'd rather not bind an ID I can't reconcile than guess between two.

## Item 3 — I disagree, and I've assigned @cmungall

You framed this as a misfiling and asked me to resolve it here. I filed it that way myself in #9703, and having now read the CMT1 entry properly, I think we're both wrong about what it is.

CMT1 doesn't include CMTX1 by accident. It argues for it, in three places:

- the top-level description: *"is electrophysiologically intermediate but is grouped here because its primary lesion is a Schwann-cell gap-junction defect; it forms the glia-axon bridge between the demyelinating and axonal compartments"*
- the subtype description, making the same case again
- an `inheritance:` block for X-linked transmission, with its own cited evidence

That is a curator who considered the placement and defended it on mechanism-primacy grounds — the primary lesion is in the Schwann cell, so it goes in the Schwann-cell compartment. My entry sorts on inheritance instead. Both are coherent nosologies and the KB has to pick one; it isn't a slip I can quietly clean up.

So deleting that content unilaterally would be overwriting someone's reasoned curation with mine, on a question this PR only exposes rather than settles. Per your own offer, I've assigned @cmungall. I've also corrected #9703, which described it as misfiled — that framing was wrong and would have misled whoever picked it up.

The one thing I'd note for whoever adjudicates: if inheritance wins, CMT1 loses three blocks; if mechanism wins, my entry should probably carry a `notes` pointer to CMT1 rather than duplicating CMTX1. Either is a small edit once the principle is settled.

## Validation

60/60 snippets verified (was 45), all 20 reference titles match cache, `just validate-disorders`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys` and the snippet gates pass, and 14,604 structural and conformance tests pass.

---

Phillip-a-richmond:
Thanks. Two follow-ups filed, and I'm deliberately not pushing again.

**The MONDO catch is recorded** — #9729. I verified it before filing: `MONDO:0010674` is mucopolysaccharidosis type 2, `MONDO:0010549` is CMTX1, both already in `cache/mondo/terms.csv`. Worth adding that the OMIM number on the same report line (302800) *is* correct for CMTX1, so the line pairs a right OMIM ID with a MONDO ID for an unrelated lysosomal storage disease.

The part I think generalises: the report's *citations* were validated 26/26 with 0% confabulation, but three of the four Named Entity Confusions in it live in the **ontology suggestions and summary tables**, which nothing validates. The NfL paper was in the citation list and got caught by reading it; the wrong CURIE and the ">90%" frequency were both in summary tables. A report can be citation-clean and still hand you a wrong CURIE that passes `validate-terms` — because term validation compares the label you write against the ontology, so it only fires if you write the CMTX1 label next to the Hunter syndrome ID. Copy the ID and let the label follow, and it validates clean.

**The reference-title gate** is #9713, opened before this review landed.

**Not pushing the `MONDO:0010549` binding.** It's a one-line change and I'd otherwise just do it, but it would reset the approval under `dismiss_stale_reviews` for something you explicitly called non-blocking, and no sibling compartment binds subtype terms either. It's captured in #9729 for whoever does that pass across all four entries.

**On the auto-merge trajectory** — agreed, and thanks for spelling out the mechanism. My read is the same as yours: this entry is correct on its own terms, #9703 is a question about CMT1's content rather than about whether this entry should exist, and either outcome is a small follow-up edit. Holding a finished entry out of the KB for a nosology debate costs more than the duplication does. I'm leaving the PR unassigned so it merges. Anyone who disagrees can assign the PR — that's the one action that holds it.

---

PR #9707, opened 2026-08-27
curate: Add combined oxidative phosphorylation deficiency 42 (MONDO:0030008, GATC)

### LINKED ISSUE #9682: Curate combined oxidative phosphorylation deficiency 42 (MONDO:0030008)
Curate a dismech entry for **combined oxidative phosphorylation deficiency 42** ([MONDO:0030008](https://monarchinitiative.org/MONDO:0030008)) — a GATC-related mitochondrial disease; the GatCAB amidotransferase is required for mitochondrial Gln-tRNA formation, so the defect impairs mitochondrial translation and multiple respiratory-chain complexes.

A combined defect points at mitochondrial translation rather than one assembly pathway — distinguish from the single-complex deficiencies.

Stub: `stubs/Combined_Oxidative_Phosphorylation_Deficiency_42.yaml`
Nominated by: unknown

kevinschaper:
Adds a knowledge-base entry for combined oxidative phosphorylation deficiency 42, a lethal infantile mitochondrial cardiomyopathy caused by biallelic variants in GATC. GATC encodes GatC, the small structural subunit of the mitochondrial glutamyl-tRNA(Gln) amidotransferase. Mitochondria have no glutaminyl-tRNA synthetase, so this complex is the only way glutaminyl mt-tRNA can be charged — when it fails, translation of all thirteen mtDNA-encoded proteins is impaired and the result is a combined respiratory chain deficiency that destroys the infant heart.

## The deep-research report was about a different disease and has been discarded

`just preflight-dr` against MONDO:0030008 returned **FAIL**:

```
FAIL  research/Combined_Oxidative_Phosphorylation_Deficiency_42-deep-research-falcon.md
  canonical gene  : GATC
  mentions        : GATC=0
  top genes       : MRPS2=53, AR=2, HP=2, GC=1
  OMIM (MONDO)    : 618839
  OMIM (report)   : 617872
```

OMIM 617872 is **COXPD36 (MRPS2)** — a different gene and a different disease. The report calls itself COXPD42 throughout while its content describes the MRPS2 entity, which is exactly why nothing but the gene-frequency check catches this: its citations resolve, its snippets would verify as exact quotes, and its ontology terms exist.

Per the discard rule the report was **not** cherry-picked and is **not** committed. I deleted the files rather than leaving them in `research/`, where they would misleadingly register provider coverage for this disorder and mislead whoever curates near it next.

This entry is therefore built entirely from primary literature found independently through PubMed. No report content reached it.

## Mechanism

```
Biallelic GATC missense (p.Met78Arg)
  └─> GatCAB trimer destabilization          GatC 20% of control; GatA and GatB fall with it
        └─> Impaired transamidation of Glu-mt-tRNA(Gln)        conditional on glutamine supply
              └─> Deficient mitochondrial protein synthesis
                    └─> Combined respiratory chain enzyme deficiency     [conforms to module]
                          └─> Tissue-selective bioenergetic failure of the myocardium
                                └─> Lethal infantile cardiomyopathy with lactic acidosis
```

7 nodes, 6 edges, one root, all reachable (measured).

Two things make this more than a generic mitochondrial translation defect:

**GatC has no catalytic activity of its own** — in the bacterial complex it is the linker holding the amidase and kinase subunits together. The patient data show that is what fails: GatC falls to 20% of control in patient fibroblasts and GatA and GatB fall with it, even though neither of their genes is mutated, while all three transcripts are unchanged. A GATC lesion behaves as a lesion of the whole enzyme.

**The biochemical phenotype is conditional on glutamine.** Patient fibroblasts charge mt-tRNA(Gln) nearly normally in standard glutamine-rich medium and fail only after glutamine withdrawal. This is a diagnostic trap — and the origin of the entry's main open question.

**Frequencies are read from patient columns, not from the paper's summary counts.** The defining paper reports nine patients across three genes in shared tables, and its headline counts (cardiomyopathy 9/9, anaemia 7/7) cover all nine, not the five GATC patients. Every GATC-specific statement here is derived by reading the per-patient columns for P4A, P4B, P4C, P5A and P5B, and the attached snippets are the sentences that name families 4 and 5. This is stated in `notes:` so a later editor does not "improve" it by substituting the headline numbers.

**No `frequency:` band anywhere.** Five patients from two families sharing one allele. A percentage would imply precision the series cannot support.

**`PARTIAL_LOSS_OF_FUNCTION`, not `LOSS_OF_FUNCTION`.** Residual activity is not inferred from the phenotype: GatC protein is reduced to about 20% rather than abolished, and the source states directly that complete loss of a tRNA-charging function is thought to be embryonically lethal.

**Two claims are kept out of the human pathograph.** Increased reactive oxygen species, and the finding that mischarged Glu-mt-tRNA(Gln) is rejected rather than mistranslated, both come from a mouse-cell GatA knockdown — different subunit, different species. They are curated on that model with those differences stated, not folded into the patient chain.

**No treatments block, and the reason is in a discussion.** Glutamine is the amide donor, and adding it rescues the defect in patient fibroblasts, so supplementation is the obvious idea. Three things stop it being a treatment and the discussion says so: the rescue uses supraphysiological culture concentrations rather than achievable plasma levels; it is seen in fibroblasts, the *least* affected tissue, while the heart is what needs rescuing and has no model; and the chloramphenicol block-and-release experiment showed translation keeps pace only briefly before charging falls behind sustained demand, which points at enzymatic throughput rather than substrate supply as the constraint. A cardiomyocyte never stops translating.

**Module conformance is declared at one node only,** `mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress`. The module is built around age-related mitochondrial damage, so its upstream nodes do not apply to a primary nuclear-gene translation defect, and its reactive-oxygen-species arm is not evidenced in these patients. No existing nuclear mitochondrial-disease entry in the KB declares conformance to that module; this one does because the module's own text names that node as the state disease-specific mitochondrial lesions converge on. Happy to drop it if that reading is wrong.

**A title-quoting snippet was caught and fixed during curation.** `PMID:29440775` was first quoted by its title. The cached record does carry abstract text, so it is now quoted from the abstract's concluding sentence instead, and a second sentence from the same abstract — that heart and brain are especially sensitive to translation defects during the period of massive mitochondrial biogenesis — was added to the tissue-selectivity node, where it is better evidence than anything the title could give.

## Validation

- `just count-verified-snippets` — `Snippets checked: 51/51 verified against cached references`
- `just validate-disorders` (batched CI equivalent) — passed
- `just check-duplicate-keys`, `just check-title-snippets`, `just check-reference-titles`, `check_snippet_length --against-ref origin/main`, `check_folded_hyphens --check`, `just check-cache-order`, `just check-term-cache-integrity`, `just validate-history` — all OK

Rebased onto current `main`, which includes the new `check-reference-titles` gate (#9141).

## One small note for the maintainers

The audit I ran before `check-reference-titles` existed flagged `PMID:24579914` as a title mismatch. It is not one: the cached PubMed title contains a **non-breaking space** in "in vivo" and the YAML has an ordinary space. The new gate uses a similarity threshold and passes it, which is the right behaviour — recording it here only so nobody chases it later.

Closes #9682


kevinschaper:
Fixed, and you were right — the sentence was false and I should have caught it myself.

I verified it before changing anything:

```
$ grep -rln 'mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress' kb/disorders/
kb/disorders/Combined_Oxidative_Phosphorylation_Deficiency_42.yaml
kb/disorders/Fontaine_Progeroid_Syndrome.yaml
kb/disorders/Kearns-Sayre_Syndrome.yaml
kb/disorders/MELAS_Syndrome.yaml
kb/disorders/Mitochondrial_DNA_Depletion_Syndrome_14B.yaml
kb/disorders/Pearson_Syndrome.yaml
```

`Mitochondrial_DNA_Depletion_Syndrome_14B` declares it on a node named `Combined Respiratory Chain Deficiency` — the same node name this entry uses — and its gene is OPA1 (`hgnc:8140`), nuclear. So the claim was wrong in exactly the way you describe, and it predates this branch.

The `notes:` paragraph now names MTDPS14B as the precedent this entry follows, and distinguishes the other four so a later reader does not treat them as precedent either: Pearson, Kearns-Sayre and MELAS are primary mitochondrial-DNA disorders, and Fontaine progeroid syndrome is not a primary mitochondrial disease at all. **The conformance decision itself is unchanged** — it is now supported by a precedent rather than justified by a false negative, which is a stronger footing than it had.

Since your review makes it precedented rather than novel, I have also withdrawn the hedge from the PR body, and I am **not** propagating the same claim to #9717 (MC1DN24), which declares the same conformance — I have corrected that PR's body too, since it carried the same wrong sentence. Its entry file never did.

**The general lesson, recorded because it generalises past this PR:** a survey claim about the corpus is as checkable as an evidence snippet, and nothing in the validation stack checks it. `just validate`, the snippet checker, the term validator and the reference-title gate would all pass a file asserting anything at all about what other files contain. One `grep` would have caught this. I am treating "none of the existing entries…" as requiring the same verify-before-asserting discipline as a quotation.

Thank you also for the four positive verifications — I have left all of them alone. The `IN_VITRO` call on PMID:24579914 is the one I would most have hated to get wrong: the title says *"in vivo"* and the abstract says *"mouse cells"*, and the two disagree.

Rebased onto current `main` (was 3 behind); the cache diff is strictly additive with none of main's rows touched.

**Validation after the fix:** `just validate` OK · `just count-verified-snippets` **51/51** · `just validate-terms` OK · `just validate-disorders` OK · `check-reference-titles`, `check-title-snippets`, `check_snippet_length`, `check_folded_hyphens`, `check-cache-order`, `check-term-cache-integrity`, `check-duplicate-keys`, `validate-history` all OK.

Re-requesting review.

kevinschaper:
Rebased to clear a conflict with `main`. **The conflict was cache-CSV only** — no KB content was involved — but the resolution changed the cache diff, so flagging it since that is something the review checks.

**How it was resolved: main-authoritative per CURIE.** I took `main`'s version of the conflicting cache files wholesale, which dropped this branch's rows, then **re-derived them by re-running `just validate`** — the validator writes the cache from OAK — rather than hand-inserting rows. No cache row in this branch was typed by hand.

**A stale row also dropped out, which is worth naming.** This branch was carrying `HP:0005791` with an older `retrieved_at` than main's:

```
main:     HP:0005791,Cortical thickening of long bone diaphyses,2026-08-27T04:30:16.217427
branch:   HP:0005791,Cortical thickening of long bone diaphyses,2026-08-21T21:03:57.767599
```

Same CURIE, same label, older timestamp, and unrelated to this entry — an artifact of the branch's older base surfacing as a diff line. Main's copy is kept and the stale one is gone at net. I checked this rather than assuming, because reintroducing a row main had deliberately changed would have been the damaging outcome.

**Verified after resolution:** `just validate-disorders` ✓ · `count-verified-snippets` unchanged ✓ · `check-cache-order` ✓ · `check-term-cache-integrity` ✓ · `git merge-tree` against current `main` merges clean. The cache diff is now exactly the rows this entry introduces, all additive.

PR #9709, opened 2026-08-27
Curate DEF6 Deficiency (MONDO:0030457)

### LINKED ISSUE #9705: Curate immunodeficiency 87 and autoimmunity (MONDO:0030457)
Curate a dismech entry for **immunodeficiency 87 and autoimmunity** ([MONDO:0030457](https://monarchinitiative.org/MONDO:0030457)) — biallelic loss of **DEF6** (hgnc:2760), an inborn error of immunity presenting with early-onset autoimmunity, lymphoproliferation, inflammatory bowel disease, and susceptibility to EBV and other infections.

Preliminary read: this looks like a **DISEASE** — one gene, one conserved mechanism. DEF6 is a guanine nucleotide exchange factor required for CTLA-4 vesicular trafficking to the T-cell surface, so its loss produces a functional CTLA-4 insufficiency phenocopy. That gives a direct mechanism-to-treatment link (CTLA-4-Ig / abatacept) worth modelling with `target_mechanisms`.

Existing coverage checked: no `kb/` entry matches MONDO:0030457, `DEF6`, or "immunodeficiency 87" on `origin/main`. The `Inborn_Errors_of_Immunity` grouping exists and may want this as a member in a later PR.

Stub: `stubs/Immunodeficiency_87_And_Autoimmunity.yaml`

aleixpuigb:
New `kb/disorders/` entry for DEF6 deficiency (immunodeficiency 87 and autoimmunity), an autosomal recessive inborn error of immunity caused by biallelic loss of the guanine nucleotide exchange factor DEF6.

Closes #9705

## Why this is worth a mechanism entry

DEF6 binds RAB11 and is needed to load CTLA-4 into recycling vesicles. Lose it and you get a functional CTLA-4 insufficiency from a completely normal CTLA4 locus — which is why the disease sits clinically alongside CTLA-4 insufficiency and LRBA deficiency despite a different causal gene. The pathograph runs that chain end to end:

DEF6 loss → disrupted RAB11 binding and fewer CTLA-4-bearing vesicles → reduced surface CTLA-4 → defective Treg suppression → systemic autoimmunity and lymphoproliferation.

Abatacept attaches to the Treg-suppression node with `treatment_effect: BYPASSES`, because soluble CTLA-4-Ig supplies extracellularly the exact function the trafficking defect withholds. That is the kind of mechanism-to-therapy link the pathograph is for.

A second arm covers impaired EBV control and EBV-driven Hodgkin lymphoma, which were the presenting features of the second reported family and absent from the index report.

## The evidence is thin, and the entry says so

The whole human evidence base is **seven patients from three kindreds in two reports**. That is stated in the entry `description`, not left for a reader to infer from the citation list. Two `KNOWLEDGE_GAP` discussions record what it costs:

1. **The SKIV2L confound.** Two of the three patients in the index report also carried a homozygous predicted-pathogenic SKIV2L variant. SKIV2L deficiency causes trichohepatoenteric syndrome, whose intractable diarrhoea, liver disease and combined immunodeficiency overlap the "severe digestive and cardiac involvement" that report attributed to DEF6. The second report makes this point explicitly and describes four siblings with no other pathogenic variant, in whom the phenotype is autoimmunity and EBV susceptibility without those features. So the entry does not curate the digestive and cardiac manifestations as established DEF6 phenotypes.
2. **The EBV arm may not be downstream of CTLA-4 at all.** Reduced surface CTLA-4 explains loss of tolerance. It is not an obvious explanation for failed antiviral control, which usually reflects a cytotoxic T-cell or NK-cell defect. Neither report tests this.

Treatment evidence is graded honestly: abatacept carries the single-patient primary observation as `SUPPORT` and the review recommendation as `PARTIAL`, because the review states it for the CTLA-4/LRBA/DEF6 group rather than for DEF6 specifically. HSCT is `PARTIAL` for the same reason — there is no DEF6-specific transplant outcome.

## Deep research

One Edison (`falcon`) run is included. `just preflight-dr` **PASSed** on gene identity (DEF6 mentioned 50 times, top gene by a wide margin).

Worth flagging for anyone relying on these reports: it cites the index paper for **fifteen of its twenty-four citation slots**, and **never cites the second family report at all** — the one that supplies the EBV arm and the SKIV2L confound. Its own "evidence-source hierarchy" section lists only Serwas et al. under direct human clinical evidence. Its reference validation was clean (5/5 resolved, 0 confabulation) but only 2/5 were assessed on-topic, the weakest of the two runs I did today. The confound and the EBV arm in this entry come from the primary literature, not from the report.

## Validation

| Check | Result |
|---|---|
| `just validate-disorders` (the authoritative batched gate) | passed |
| `just count-verified-snippets` | 22/22 verified |
| `just check-folded-hyphens` / `check-snippet-length` / `check-title-snippets` / `check-snippet-grading` | no new violations |

Note on snippets from PDF-extracted full text: two quotes from the second report span a line break, because that cache entry is extracted from a PDF and its text wraps mid-sentence. They are exact substrings of the cached file and verify, but the surrounding text has ligature and hyphenation artifacts that make much of that paper unquotable in practice. Term caches were re-derived from scratch after the final edit so the committed rows cover only the two CURIEs this entry introduces.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VrvBLw36zmyZv3GeRbtuj4)_

### DIFF
--- a/history/disorders/DEF6_Deficiency/2026-08-27T164739Z-claude-code-ac785e.yaml
+++ b/history/disorders/DEF6_Deficiency/2026-08-27T164739Z-claude-code-ac785e.yaml
@@ -12,9 +12,9 @@ session:
     agent_tool: claude-code
 links:
   issues:
-  - "9705"
+  - https://github.com/monarch-initiative/dismech/issues/9705
   prs:
-  - "9709"
+  - https://github.com/monarch-initiative/dismech/pull/9709
   urls: []
 events:
 - type: CREATE
--- a/history/disorders/DEF6_Deficiency/2026-08-27T194232Z-claude-code-a76172.yaml
+++ b/history/disorders/DEF6_Deficiency/2026-08-27T194232Z-claude-code-a76172.yaml
@@ -1,0 +1,29 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: DEF6_Deficiency
+  path: kb/disorders/DEF6_Deficiency.yaml
+session:
+  id: 2026-08-27T194232Z-claude-code-a76172
+  timestamp: '2026-08-27T19:42:32Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9705
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9709
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - treatments
+  - experimental_models
+  summary: Address review round 2 on PR 9709
+  details: |
+    Addressed both blocking items from the second review round on PR 9709. The Lymphadenopathy evidence quoted a figure-legend fragment that began mid-word on the tail of inflam-mation and evidenced EBV-driven lymphoproliferation rather than lymph node enlargement; it verified as an exact substring, which is precisely why the offline gates did not catch it. Replaced with the sentence recording cervical lymph node enlargement with spontaneous regression in the eldest sibling. Immunoglobulin replacement was linked to Systemic Autoimmunity with BYPASSES while its own description said it covers the antibody deficit; replacement-dose immunoglobulin does not treat autoimmunity, so the edge asserted something clinically untrue. Added an Impaired Class-Switched B Cell Compartment node downstream of defective Treg suppression, evidenced from the low class-switched B cells clause of a sentence already verified in this file, and retargeted the treatment to it. Also took three suggestions: moved Hepatosplenomegaly from Gastrointestinal to Hematologic so it groups with the lymphoproliferative arm, added explicit evidence for the CMV half of the viral-susceptibility phenotype, and added a DECREASED surface-CTLA-4 readout to the Jurkat model link, which the previous history record's prose had claimed was already present when it was not. Normalised the links block of the earlier history record to the URL form. Snippets 33 to 36, all verified; pathograph nine nodes in one component.
--- a/kb/disorders/DEF6_Deficiency.yaml
+++ b/kb/disorders/DEF6_Deficiency.yaml
@@ -197,6 +197,13 @@ pathophysiology:
   - target: Systemic Autoimmunity
     causal_link_type: DIRECT
     description: Unrestrained T-cell activation drives multi-organ autoimmunity, prominently autoimmune cytopenias.
+  - target: Impaired Class-Switched B Cell Compartment
+    causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
+    intermediate_mechanisms:
+    - disturbed T-cell help to germinal centre B cells
+    description: >-
+      Loss of regulated T-cell help disturbs germinal centre output, leaving a
+      reduced class-switched memory B-cell compartment.
   - target: Lymphoproliferation
     causal_link_type: DIRECT
     description: Loss of the CTLA-4 brake also permits uncontrolled lymphocyte expansion.
@@ -207,6 +214,25 @@ pathophysiology:
     description: >-
       Immune dysregulation extends to failure of EBV control, though the precise
       effector defect responsible has not been resolved.
+- name: Impaired Class-Switched B Cell Compartment
+  biological_scale: CELLULAR
+  description: >-
+    Reduced class-switched B cells alongside T-cell lymphopenia in the index
+    cohort. This is the humoral arm of the disease and is what immunoglobulin
+    substitution is actually directed at — it is distinct from the autoimmunity,
+    which replacement-dose immunoglobulin does not treat.
+  cell_types:
+  - preferred_term: B cell
+    term:
+      id: CL:0000236
+      label: B cell
+  evidence:
+  - reference: PMID:31308374
+    reference_title: "Human DEF6 deficiency underlies an immunodeficiency syndrome with systemic autoimmunity and aberrant CTLA-4 homeostasis."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Clinical and immunological phenotypes in DEF6-mutated\npatients include T-cell lymphopenia, low class-switched B cells"
+    explanation: Records the reduced class-switched B-cell compartment among the immunological phenotypes of DEF6-mutated patients.
 - name: Systemic Autoimmunity
   biological_scale: ORGANISM
   description: >-
@@ -349,6 +375,14 @@ phenotypes:
       Restricts the infection phenotype to EBV in the unconfounded family and
       attributes the broader infection susceptibility to the SKIV2L carriers,
       which is why this phenotype is scoped to EBV and CMV.
+  - reference: PMID:32562707
+    reference_title: "DEF6 deficiency, a mendelian susceptibility to EBV infection, lymphoma, and autoimmunity."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Except for CMV in\npatient 4, there was no evidence for other viral infection"
+    explanation: >-
+      Carries the CMV half of this phenotype explicitly, and simultaneously
+      excludes any other viral susceptibility in the unconfounded family.
 - name: Persistent EBV viremia
   category: Immunologic
   frequency: FREQUENT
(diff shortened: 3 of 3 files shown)


aleixpuigb:
All five 🟡 items addressed in c18d4886e, plus five of the suggestions. Apologies for the delay between the review and this push — git push credentials were unavailable in my environment for about 40 minutes; the work was finished and validated well before it landed.

**1. The infections phenotype contradicting the SKIV2L gap.** You're right, and this is the finding I'm most glad you caught — the entry argued the confound in its own discussion section and then quietly ignored it two hundred lines earlier. The phenotype is now `Susceptibility to Epstein-Barr virus and cytomegalovirus`, bound to `HP:0004429`, scoped to the two pathogens that occurred in unconfounded patients.

One deliberate divergence from your suggested fix: you asked for a *counter*-evidence item quoting the restriction sentence. I made it the phenotype's primary `SUPPORT` evidence instead. Once the claim is narrowed to EBV and CMV, that sentence supports it directly — it is only counter-evidence to the broader claim I withdrew. Recording it as a refutation of a claim the entry no longer makes seemed like it would read as confusion rather than care. The explanation says what it excludes, and I extended the `def6_skiv2l_confound` discussion to cover the infections explicitly. Happy to flip it if you'd rather have the counter-evidence shape.

**2. EBV phenotype.** Added, `HP:0020072`, using the >4 log copies/mL quote already verified in the file. Thanks for the `CD27-related_lymphoproliferative_and_immune_disorder` pointer — it was the right structural analog.

**3. Lymphoproliferation.** Added `Lymphadenopathy` and `Hepatosplenomegaly`. The latter is `PARTIAL`: the sentence you pointed me at lists phenotypes across the whole index cohort, which includes the two SKIV2L carriers, and the paper doesn't resolve it per patient — the same reasoning that drove item 1.

**4. The Jurkat model.** Added as `experimental_models` with `modeled_mechanisms` → `Reduced Surface CTLA-4 Availability`, `RECAPITULATES`, `fidelity: MODERATE`, with its own limitations and evidence. Your framing sharpened why it belongs: it is the one piece of evidence that rules out the patient-level confounders — including SKIV2L — that the observational data structurally cannot. I left the Def6-null mouse models out since their PMIDs aren't cached and fetching them would widen the diff mid-review-round.

**5. Treatments.** All three added, each `PARTIAL` with the reason stated. Ig replacement needs a caveat worth flagging: the cached sentence documents it in the index report's **P1**, who is one of the SKIV2L carriers. Citing it as an established DEF6 treatment would repeat exactly the error of item 1, so the description and explanation both say so. The DR's "all reported patients received regular immunoglobulin treatment" is broader than what I could verify from the cache.

## Suggestions taken

Split the bundled autoimmunity/lymphoproliferation node (1); gave the EBV arm a terminus — a new `EBV-Driven Lymphoproliferative Disease` node — so the `Lymphoma` phenotype now connects to the graph, 8 nodes in one component (2); abatacept → `PROTEIN_REPLACEMENT` (3); HSCT → `BYPASSES`, your reasoning was exactly right (4); added the index authors' concurring statement to the SKIV2L discussion (5); filled `links.issues` on both history records (9).

## Not done

- **Shared CTLA-4-recycling module** (6). Agreed on the substance, but a module touching `CTLA4_Haploinsufficiency` and `LRBA_Deficiency` is its own PR — it would triple this diff and put two entries I haven't reviewed into a curation PR for a third.
- **`biochemical:` and `diagnosis:`** (7, 8). Both are genuinely quotable and I'd support them; deferred for the same scope reason, and neither was blocking.
- **GeneReviews.** Confirmed via PubMed: no GeneReviews chapter for DEF6 deficiency. Your instinct was right that one would be surprising for a 7-patient disease.

## Two defects of my own, caught by the offline gates before this push

Worth recording since they're the kind of thing a reviewer shouldn't have to find:

- `check-snippet-length` rejected a one-word `hepatosplenomegaly` snippet I'd lifted from the phenotype list — exactly the "bare term carries no propositional content" case the check exists for. Replaced with the sentence.
- `check-snippet-grading` caught me grading the *same* Serwas sentence `HUMAN_CLINICAL` on a pathophysiology node and `IN_VITRO` on the new model block. `IN_VITRO` is correct for both: patient T cells assayed ex vivo are cultured cells, not an in-vivo clinical observation. Both now agree, and the explanation says why.

Validation on c18d4886e: `just validate-disorders` passed, **33/33** snippets verified (was 22), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` all clean, both history records validate. No new term-cache rows — every CURIE added was already cached.

Re-requesting review.

---

aleixpuigb:
`test (3.13)` went red on `c18d4886e` and I've pushed a fix as `42a19b83d`. Explaining because the failure names five files this PR never touches, which is confusing on its face.

`check_environmental_evidence.py` reported 9 new evidence-free exposures in `Epilepsy_with_Generalized_Tonic-Clonic_Seizures_Alone`, `Infective_Endocarditis`, `Invasive_Candidiasis`, `Pulmonary_Embolism` and `Rheumatic_Heart_Disease`. None of them is in this diff.

The cause is base staleness, not content. The check compares the working tree against **current** `origin/main`, and all five files changed on main after this branch was cut — so the branch was carrying older copies whose exposure names no longer match the baseline. The script warns about exactly this: *"Findings are keyed on (file, exposure name), so renaming a baselined exposure also reads as a new finding here."* Running the same check on `main` alone passes, which confirms main is not broken.

Fix is a base refresh — `git merge origin/main`, conflict-free. I merged rather than rebased deliberately: a re-review was running against `c18d4886e` at the time, and a force-push would have invalidated a SHA that job was mid-way through analysing. `CLAUDE.md`'s preference for rebase is about *routine* refreshes; this was a required one, and the merge keeps the reviewed commit in history.

Re-verified on the merged tree, including the `check_reference_titles.py` gate that main added in the meantime:

| Check | Result |
|---|---|
| `just count-verified-snippets` | 33/33 |
| `check-environmental-evidence` | OK — the failing check, now green |
| `check-reference-titles` | OK (new from main) |
| `check-duplicate-keys` / `folded-hyphens` / `snippet-length` / `title-snippets` / `snippet-grading` / `entity-refs` | all OK |

Diff scope against refreshed main is unchanged — the same 14 files.

---

aleixpuigb:
Both 🟡 items fixed in `2860d4cac`, plus three of the suggestions. Both findings were correct and both are the more interesting kind — neither was catchable by a gate.

**1. The lymphadenopathy snippet.** You're right, and this one bothers me: I noticed while curating that this cache entry is PDF-extracted with mid-word line wraps, worked around it for other quotes, and then used a figure-legend fragment starting on the tail of `inflam-mation` anyway. It verified as an exact substring, so `count-verified-snippets` passed — which is the whole point you're making. **Exact is not on-point.** Swapped for the sentence you found at `:326-328`, which states the cervical node enlargement directly.

This is a concrete instance of a gap I filed earlier today as #9711: nothing on `EvidenceItem` records whether a snippet came from a clean abstract or from PDF-extracted full text, so a reader can't distinguish a deliberate multi-line quote from a corrupted fragment. Your catch is the best argument for it I have.

**2. Immunoglobulin replacement on the wrong node.** Also right, and worse than a mislabel — the edge asserted something clinically untrue. I took your second option rather than dropping the link, because the humoral deficit is real and the entry was silently missing it: added `Impaired Class-Switched B Cell Compartment` downstream of defective Treg suppression, evidenced from the "low class-switched B cells" clause of the sentence already verified in this file, and retargeted Ig replacement to `BYPASSES` that. The `target_mechanisms` description now says explicitly that this is the humoral deficit and not the autoimmunity, and that replacement dosing is not the high-dose immunomodulatory intervention.

The edge into the new node is `INDIRECT_KNOWN_INTERMEDIATES` with "disturbed T-cell help to germinal centre B cells" named — the papers report the association, not the mechanism, and I didn't want a `DIRECT` edge claiming more than that.

## Suggestions taken

- **Hepatosplenomegaly → `Hematologic`.** Agreed; it now groups with `Autoimmune cytopenia` and `Lymphadenopathy` rather than sitting alone under a category that reads as organ-specific.
- **CMV now evidenced.** Added `"Except for CMV in\npatient 4, there was no evidence for other viral infection"` — it carries the CMV half *and* re-states the exclusion, which suits a phenotype defined as much by what it rules out.
- **Jurkat readout added.** You were right that the prose overstated it: my previous history record said the model went in "with its own readout evidence" when only link-level evidence existed. There is now a `DECREASED` surface-CTLA-4 `ExperimentalReadout` repeating the link's target. Thank you for reading the history record against the YAML — that discrepancy was mine and nothing would have flagged it.
- **History `links` format** normalised to the URL form on the earlier record.

## Still deferred

`biochemical:`, `diagnosis:`, and the shared CTLA-4-recycling module. Agreed on all three, and agreed the module is the interesting one and belongs in its own PR — it would touch `CTLA4_Haploinsufficiency` and `LRBA_Deficiency`, neither of which I've reviewed.

## On the environmental-evidence failure you saw

That was real on `c18d4886e` and is fixed in `42a19b83d`, which you reviewed just before. It was base staleness, not content: the check compares against current `origin/main`, and five files this PR never touches had exposures renamed on main after this branch was cut. Merging main in clears it. Details in [the comment above](https://github.com/monarch-initiative/dismech/pull/9709#issuecomment-5444296673).

## Validation on `2860d4cac`

`just validate-disorders` passed; **36/36** snippets verified (was 33); `check-environmental-evidence` now green; `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` all clean; all three history records validate. Pathograph is 9 nodes in 1 connected component.

Re-requesting review.

---

PR #9715, opened 2026-08-27
Curate 2026 lung-on-chip literature: alveolar-capillary barrier module, HAPE entry, 5 disorders

cmungall:
## What this is

I scanned the 2026 organ-on-chip literature for papers making disease-mechanism claims we could curate, and found that five of them are all reporting the same previously unmodelled thing. When you expose **one** side of the lung's blood-gas barrier to an insult, the **other** side — never directly exposed — gets injured too. Five unrelated insults, five independent groups, same result.

That is a mechanism claim rather than a device observation, and it recurs across enough diseases to be worth factoring out, so this PR adds it as a module and curates the underlying papers into the entries they belong to.

## The cross-cutting finding

| Insult | Paper | What was observed |
|---|---|---|
| Diesel particulate | PMID:42546767 | Cytotoxicity and DNA damage in **both** epithelial and endothelial layers from apical-only exposure |
| Chlorine gas | PMID:41442163 | Epithelial junction gaps immediately; endothelial junction loss delayed to 72 h |
| Whole cigarette smoke | PMID:42473541 | Vascular network destroyed, epithelium intact — and **not** destroyed if the epithelium is absent |
| Isothiazolinone (humidifier disinfectant) | PMID:42204607 | Barrier disruption in the alveolar compartment |
| Cold-storage ischemia-reperfusion | PMID:42083145 | Barrier loss with compartment-specific adhesion-molecule shedding |

Two of these are more than co-occurrence. The chlorine paper **times** the propagation, and the cigarette-smoke paper **controls** it: vascular networks exposed to smoke with no epithelium above them stay intact, so the vascular damage is being carried by an epithelial signal rather than by the smoke itself. You cannot run that control in a patient or an animal — expose one side of the alveolar barrier and not the other — which is why this took organ-chips to establish.

## Changes

**New module — `kb/modules/alveolar_capillary_barrier_failure.yaml`**

Five nodes, `trigger → amplifier → central_effector → effector → consequence`. The key conformance target is *Alveolar Epithelial and Endothelial Junctional Disruption*: whatever the initiating insult, the barrier fails there. 13 evidence snippets.

This is deliberately narrower than the existing `epithelial_barrier_dysfunction` module, which models the allergic type-2 "epithelial barrier hypothesis" across skin, airway, gut and oesophagus and terminates in IgE sensitisation and the atopic march. That one is about allergen penetration driving sensitisation; this one is about blood-gas barrier mechanics driving oedema and gas-exchange failure. A disorder can legitimately conform to both. The distinction is written into the module `notes` so it does not have to be rediscovered.

**New disorder — `kb/disorders/High_Altitude_Pulmonary_Edema.yaml`** (MONDO:0031257)

7 pathophysiology nodes, 13 distinct cited references, 44 verified snippets. Built from core HAPE literature, not just from the chip paper that surfaced it. Also carries `prevalence` (2 records), `diagnosis`, `has_subtypes` (classic and re-entry HAPE), and `genetic` (JAK2, HRG, and one deliberately **unbound** cytochrome P450 locus) — all added during review, see below.

The entry commits to the pressure-primary account: exaggerated and spatially **uneven** hypoxic vasoconstriction concentrates cardiac output in the capillaries that escaped constriction, pressure exceeds what the barrier can mechanically withstand, and inflammation follows the stress failure rather than causing it. The evidence that fixes that ordering is a negative result, and it is curated explicitly rather than paraphrased away: in PMID:11319198, transcapillary transferrin escape was still **normal** in susceptible subjects whose capillary pressure had already risen. Two `mechanistic_hypotheses` record the pressure-primary (CANONICAL) and inflammation-contributory (ALTERNATIVE) readings.

**Existing entries — organ-chip models wired into the pathograph**

| Entry | Model(s) | Links |
|---|---|---|
| `Tuberculosis` | Autologous iPSC alveolus-chip (PMID:41477823) | 5 |
| `Influenza` | Immune-competent lung-chip (PMID:40987954) | 3 |
| `Chronic_Obstructive_Pulmonary_Disease` | Vascularized smoke-exposure chip; patient-derived COPD chip (PMID:42473541, PMID:42475431) | 4 |
| `Idiopathic_Pulmonary_Fibrosis` | Alveolar array chip with breathing mechanics (PMID:41406599) | 2 |
| `Acute_Respiratory_Distress_Syndrome` | Chlorine airway-chip; *Pseudomonas* VAP chip (PMID:41442163, PMID:41252215) | 4 |

ARDS also gains 4 `conforms_to` edges to the new module, making it the module's principal conformer; HAPE conforms at 3 nodes.

Two of these are worth singling out because they are negative results, which this field reports unusually well and which we have been under-recording:

- **Tuberculosis** gets a `FAILS_TO_RECAPITULATE` link. Neither macrophages nor epithelium in that chip were permissive to Mtb replication — unlike 2D culture, where Mtb replicates within 48 h, and unlike human disease. That bounds the model: it should not be used to score immune-evasion phenotypes or to screen compounds whose readout is restriction of intracellular replication.
- **Influenza** records a dissociation a cytokine panel alone would hide. IL-1β and TNF-α, normally lumped together as pro-inflammatory drivers, act in **opposite** directions on the storm: blocking IL-1β abolishes it, blocking TNF-α makes it worse.

**Report — `docs/reports/organ-on-chip-literature-scan-2026-08.md`**

The scan itself, including a journal analysis that has a practical consequence for how we monitor this field: there is no home journal. 500 general organ-on-chip papers span **254 journals**, 175 of which contributed exactly one paper, and bioRxiv is the single largest venue. A journal watchlist will not work; a keyword sweep including preprints is the only viable approach, and it needs a `Review[pt]` filter because roughly half the hits are reviews. Also worth knowing: one 2026 lung-chip record in this set is a **retraction notice**, so publication type needs checking before citing.

Added to the mkdocs nav.

## Review rounds

**Round 1 — three blocking findings, all fixed in `a4d9ec3`.**

- **Two ontology bindings were wrong.** `Influenza`'s `epithelium of bronchiole` was bound to `UBERON:0002031` (**bronchus**) — a level mismatch, since the chip is explicitly small-airway; rebound to `UBERON:0002051`. `Lung fibroblast` was bound to generic `CL:0000057` on both the IPF and Influenza chips; rebound to `CL:0002553` (fibroblast of lung). Both terms verified against OLS before rebinding. IPF's *pre-existing* `fibroblast`→`CL:0000057` pairing is left alone — internally consistent, and not this PR's content.
- **HAPE under-used two caches this PR itself adds.** Fair hit. Added `prevalence`, `diagnosis`, `has_subtypes` and `genetic`, all sourced from text already in the branch. The global incidence is recorded as a range with `prevalence_class: UNKNOWN` rather than a point estimate, since &lt;0.01%–31% is conditioned on ascent rate and not a denominator.
- **A Cough evidence item claimed a figure its snippet didn't carry.** Fixed, though not as suggested: the Table 1 row extracts as `Cough221 (53.8)65 (56.0)156 (52.9)0.640`, four tokens, which would trip `check-snippet-length`. The snippet is now the early-symptom sentence and the 53.8% moved into the phenotype description as attributed prose. The reviewer checked this and withdrew the suggestion.

**Round 2 — a non-blocking suggestion that was a real defect, fixed in `2b23604`.**

The review questioned whether expanding the source's garbled "P1B1 cytochrome genes" to CYP1B1 was licensed. It was not. PubMed returns **0** papers for `high altitude pulmonary edema AND CYP1B1`, against 6 for CYP11B2 and 1 for CYP17A1. The `hgnc:2597` binding asserted a machine-readable gene-disease susceptibility claim with nothing behind it, so it has been **withdrawn**; the entry is now `Unidentified cytochrome P450 locus` with no `gene_term`.

It was deliberately *not* rebound to CYP11B2. Reading "P1B1" as "11B2" is an inference PMID:42254115 does not license, and swapping one guess for another would repeat the error rather than correct it. The `notes` record the source wording, the withdrawn binding, the PubMed counts, and why the likely gene is still unbound.

**Round 3 — the last optional note, fixed in `1f9be45`.**

The environmental description called rapid ascent "the necessary and sufficient environmental exposure". That was false and contradicted this entry's own JAK2 note two sections away: ascent is necessary but not sufficient, since HAPE takes a susceptible minority of those who ascend. Reworded to say so. The Classic HAPE subtype snippet, which had stopped just before an en-dash, was also extended to the full sentence — I had assumed extending past the dash risked normalisation trouble, tested it instead, and it validates cleanly.

**What this review round says about the tooling.** Every finding across the three rounds was invisible to validation: two anatomically wrong but internally consistent ontology bindings, a gene binding inferred from a typo, an explanation citing a figure its own snippet lacked, and a description contradicting another section of the same file. Schema, term and exact-quote checks passed throughout. They confirm a claim is well-formed and quotable; they cannot confirm it is true, or that the file agrees with itself.

## The `check_environmental_evidence` failures — resolved upstream

This PR went red three times on `check_environmental_evidence`, each time in files it does not touch. The cause was a race in the ratchet, not anything in this branch.

`--against-ref origin/main` derived the grandfather set **live from the tip of `main`**, not from the merge base. The #8296 campaign was backfilling `evidence:` onto uncited exposures at the time, so that set drained continuously — I measured **46 → 37 → 11 → 6** over about seven hours. When `main` cited an exposure that a branch's older copy of the same file still left uncited, the branch read as having *introduced* it. The clearest instance flagged 5 exposures across 4 files (`Metastatic_Melanoma`, `Metastatic_NSCLC`, `Oral_Cavity_Squamous_Cell_Carcinoma`, `Paranasal_Sinus_Squamous_Cell_Carcinoma`), all byte-identical between this branch's head and its own base, into which `main` had added exactly 35 lines of `evidence:`.

**This is now moot, and not because of anything here.** #9768 landed on `main` while this PR was open and **retired the ratchet altogether** — the #8296 backlog reached zero, so `--against-ref`, `--update-baseline` and `tests/environmental_evidence_baseline.txt` are gone and the check is an absolute gate over all of `kb/`. There is no live baseline left to race against. That is a better outcome than the merge-base fix I suggested, which would have preserved the ratchet; the real fix was that the ratchet had finished its job.

This branch passes the new absolute check: **0 evidence-free exposures**, nothing waived on its account. Its companion #9751 also makes a recorded failed search first-class via a `review_notes` waiver, which nothing here needs — every exposure this PR adds is cited.

One earlier diagnosis on this thread — that a shallow clone truncated the baseline — was checked and is not what happened; the reviewer has since withdrawn it. Counting evidence-free exposures directly from the `origin/main` tree reproduced the runner's number exactly, with no history walk involved.

## Two judgment calls to flag

**HAPE's `exposure_term` is left unbound.** I searched ECTO and XCO; neither has a high-altitude or hypobaric-hypoxia exposure class. The closest XCO candidates (`XCO:0001406` hypobaric chamber, `XCO:0000010` controlled air oxygen content) describe laboratory apparatus, not the exposure a traveller undergoes, so binding one would misdescribe it. Per the terms guidance, no term beats a bad one. The rationale is recorded in the entry `notes` so it is not re-researched. Note this concerns the *ontology binding*, not evidence — the exposure itself is cited, so it needs no #9751 waiver.

**Unrelated `references_cache` churn was reverted, not committed.** Running the reference validator rewrites cache files as a side effect — YAML quoting plus an added `full_text_attempted` flag. None of those files are cited by this work, and including them would have added thousands of lines of churn and invited merge conflicts.

## Known, accepted

`measure_type: PERIOD_PREVALENCE` on both prevalence records is standing in for a cumulative incidence over an ascent episode. `PrevalenceMeasureEnum` has no `CUMULATIVE_INCIDENCE`; `PERIOD_PREVALENCE` is the closest honest fit and the `notes` say so. Worth a schema NTR if the pattern recurs for other exposure-triggered diseases.

## Validation

Re-run in full after every push, including the four rebases and the PR Shepherd merge at `b0f3fc591`. Results below are from the current head; `main` moves fast enough that a SHA quoted here may be a step behind, but the content claims hold — every push to this branch has been either the same content on a newer base or a fix listed under Review rounds.

- `linkml-validate` and `linkml-term-validator` clean on all 7 KB files
- **449 / 457 snippets verified, 0 errors** (8 skipped by prefix; the reference validator's only warnings are pre-existing EGA accessions in `datasets:` blocks this PR does not touch)
- `check_environmental_evidence` in its **new absolute form** — OK, 0 evidence-free exposures
- `check_snippet_length`, `check_title_snippets`, `check_reference_titles` — 0 new findings each
- `check-folded-hyphens` and `check-snippet-grading` clean
- whole-KB `check-duplicate-keys` (4413 files) and `check-entity-refs` (2660 files) clean
- `check-term-cache-integrity` and `check-cache-order` clean; the 5 cache rows added are all validator-derived
- `validate-history-all` clean; 13 history records added
- `pytest tests/test_data.py` — 43,690 passed; FK, entity-ref and uniqueness subsets re-run green after each subsequent push
- working tree clean after the validators — no cache churn

On the Shepherd's `mkdocs.yml` conflict resolution: I checked it rather than assuming. Both nav entries survive, every `.md` path in the nav resolves on disk, there is no duplicated line and no conflict residue, and all 7 KB files plus the report are byte-identical to the head that was last approved.

The diff against `main` is **50 files, +6160, and zero deletions** — nothing in this PR removes or rewrites existing content, and that has held across every rebase and the merge.

Two papers were deliberately left uncurated, with reasons in the report: PMID:41982094 (stretch reduces SARS-CoV-2 pseudovirus fusion — real, but does not map cleanly onto a node in any existing COVID entry) and PMID:42138232 (benzene VOC exposure — would need a benzene-inhalation entry that does not exist yet).

https://claude.ai/code/session_0173xmFQgM5BSEehqs9UBNkr


cmungall:
`test (3.13)` failed on the first commit, and I've fixed it. Flagging one thing explicitly because I did it to my own PR: I added the `scope-override` label.

The failure was the curation-scope guard. Because this PR touches `kb/`, every other changed file has to sit under `kb/|stubs/|history/|cache/|references_cache/|research/|pages/|dashboard/`. Two files don't:

```
docs/reports/organ-on-chip-literature-scan-2026-08.md
mkdocs.yml
```

Both are required by convention rather than incidental. `CLAUDE.md` says analysis reports go in `docs/reports/` and that everything under `docs/` should be surfaced in the mkdocs nav, so the report and its one-line nav entry travel together. The one in-scope path that would have satisfied the guard is `research/`, but `CLAUDE.md` is explicit that `research/` is only for deep-research provider outputs and that landscape surveys belong under `docs/` — so moving it there would have traded a passing check for a documented-convention violation. That left the label, which is what its own description is for ("Allow an intentional curation PR exception outside the default curation-only path scope") and what 372 prior PRs have used, including curation PRs that carried considerably more infra than two docs files.

If you'd rather this PR stay strictly curation-only, the clean alternative is to drop those two files here and land the report separately. Happy to do that instead — say the word.

I could not re-run the failed job (403), and `main.yaml` triggers on `pull_request` default types, which don't include `labeled`, so the label alone wouldn't have re-run CI. Rather than push an empty commit to force it, I re-read my own diff and found a real defect to fix, which re-triggered CI legitimately:

**Two evidence snippets did not support their claims.** The `Cough` phenotype cited a quote describing its source review's scope, and `Tachycardia` cited a quote about patients seeking prevention advice. Neither mentions the finding it was attached to. Both were graded `PARTIAL` with explanations conceding the gap — but grading a non-supporting quote `PARTIAL` is not a substitute for finding a real source. Replaced with substantive quotes from two newly cached clinical papers: PMID:41316096 (411-patient multi-centre inpatient cohort; describes the dry-exertional-to-frothy cough progression, lists tachycardia among examination findings, and reports cough in 53.8% of patients, which grounds the `FREQUENT` frequency) and PMID:42254115 (Nepalese case series, whose diagnostic criteria name tachycardia as a qualifying physical sign). Both now carry `SUPPORT`, graded `HUMAN_CLINICAL`.

Validation on the new head: schema and term validation clean, 35/35 snippets verified, all five snippet/evidence gates pass with no new baseline defects, duplicate-key and entity-ref checks clean, history record added.

---

cmungall:
Second CI failure, different cause, fixed by rebasing onto current `main`. I force-pushed my own branch to do it — flagging that explicitly.

`check_environmental_evidence --against-ref origin/main` reported 9 evidence-free exposures in five files this PR never touches:

```
Epilepsy_with_Generalized_Tonic-Clonic_Seizures_Alone   1
Infective_Endocarditis                                  2
Invasive_Candidiasis                                    3
Pulmonary_Embolism                                      2
Rheumatic_Heart_Disease                                 1
```

Those nine had evidence curated onto them on `main` after this branch was cut. The check builds its grandfather set live from current `main` — evidence-*free* exposures only — so once an exposure is cited on `main` it leaves the set, and a stale branch still carrying the uncited copy reads as a new finding. The counts confirm it: the committed baseline holds 46 evidence-free exposures, CI's live set from `main` holds 37, and 46 − 37 = 9.

Worth knowing because it will recur: this makes **any** PR branch go red once `main` curates an environmental exposure that the branch predates, and the red points at files the PR never touched. The check's docstring says the base branch is "green by construction," which is true — it's stale *branches* that are exposed. A reviewer seeing this red would reasonably misread it as a defect in the PR.

Rebased onto `17ab5a12f`. No conflicts. Post-rebase diff against `main` is 42 files, +5850, **zero deletions** and no unrelated churn — the modified files are only this PR's five disorders, the two derived cache CSVs, and the one-line `mkdocs.yml` nav entry.

Verified locally before pushing, reproducing what CI runs:

| Check (`--against-ref origin/main`) | Result |
|---|---|
| `check_environmental_evidence` | 37 grandfathered, **0 new** — matches CI's own count |
| `check_snippet_length` | 1335 grandfathered, 0 new |
| `check_title_snippets` | 280 grandfathered, 0 new |
| `check_reference_titles` | 107 grandfathered, 0 new |

Plus, on the rebased tree: schema validation clean on all 7 KB files, 440/448 snippets verified (8 skipped by prefix — pre-existing non-PMID accessions), whole-KB duplicate-key and entity-ref checks clean, term-cache integrity and ordering clean.

The earlier curation-scope guard failure is resolved and that step now passes.

---

cmungall:
All three IMPORTANT findings are fixed in `a4d9ec349`, plus the Cough suggestion. Thanks — findings 1 and 2 were real mismatches I introduced, and finding 3 was a fair hit: I pulled two papers into the cache for a phenotype fix and left most of their content unused.

I verified both ontology claims against OLS and all three gene symbols against the HGNC REST API before binding, rather than taking the identifiers on trust.

**1. Influenza bronchiole binding** — rebound `UBERON:0002031` → `UBERON:0002051` (epithelium of bronchiole). You were right that this is a level mismatch rather than a granularity preference: the IC-LOC is explicitly a small-airway model.

**2. Lung fibroblast** — rebound to `CL:0002553` (fibroblast of lung) on both the IPF alveolar array chip and the Influenza IC-LOC, whose stromal CXCL12 finding is attributed to lung fibroblasts specifically.

One deliberate exception: I left the pre-existing IPF pathophysiology binding of `preferred_term: fibroblast` → `CL:0000057` untouched. That pairing is internally consistent, and it isn't this PR's content — changing it would widen the diff past the finding.

**3. HAPE content gaps** — added all four sections, sourced entirely from text already in the branch:

- **`prevalence`** — two records. The global figure is `PERIOD_PREVALENCE` / `UNKNOWN` rather than a point estimate, because the reported range (<0.01% to 31%) is conditioned on ascent rate and susceptibility rather than a fixed denominator; a range recorded as a number would be a false precision. The rate-conditioned ~4% for >600 m/day trekkers gets `rate_per_100000: 4000.0` and `ABOVE_1_IN_1000`, since that one does have a stated denominator.
- **`diagnosis`** — the two-symptom plus two-sign threshold, with the five-point Qinghai criteria summarised in `notes`.
- **`has_subtypes`** — Classic and Re-entry HAPE. I added a second evidence item quoting "The pathophysiological mechanism is likely identical for both types", because that is what justifies splitting on population and presentation rather than pathophysiology — worth making explicit so a later curator doesn't try to give re-entry HAPE its own pathograph.
- **`genetic`** — JAK2, HRG, CYP1B1, all `SUSCEPTIBILITY`, all `PARTIAL` since the case series cites them secondarily rather than reporting primary association data. Note the source prints CYP1B1 as "P1B1 cytochrome"; I bound the intended gene and left the snippet quoting the source verbatim, with the discrepancy recorded in `notes` rather than silently corrected.

**Cough suggestion — took the point, different fix.** Quoting the Table 1 row doesn't work: it extracts as `Cough221 (53.8)65 (56.0)156 (52.9)0.640`, which is four words and would trip `check-snippet-length` as a new short snippet, besides being unreadable. Instead the snippet is now the early-symptom sentence, which substantively supports cough, and the 53.8% figure moved into the phenotype `description` as curator prose with its source named. That resolves the underlying problem — snippet and explanation now agree.

**On the missing deep-research artifact:** noted, and I agree it shouldn't block. For the record, this entry came from a literature scan rather than a DR provider run, which is why there's no `research/*-deep-research-*.md`. The scan is in `docs/reports/organ-on-chip-literature-scan-2026-08.md`.

Validation on `a4d9ec349`:

| Check | Result |
|---|---|
| `linkml-validate`, `linkml-term-validator` | clean on all three edited files |
| HAPE snippets | 44/44 verified |
| Reference validator | 0 errors (8 pre-existing EGA `datasets:` warnings, untouched blocks) |
| 4 ref-based gates vs `origin/main` | 0 new findings each |
| duplicate-key, entity-ref, cache integrity/order | clean |
| `pytest -k "subtype_foreign or entity_ref or unique"` | 9947 passed |

The diff for this commit is +267/−13, and every one of the 13 deletions is an intended replacement (the Cough description and evidence item, and the three rebound terms) — no incidental removals.

---

cmungall:
Thanks for the approval — and your non-blocking CYP1B1 suggestion was right, so I've acted on it in `2b2360421`. This resets the approval; that's the correct trade here, because the binding was wrong.

You asked whether the primary papers might name a different cytochrome. PubMed settles it:

| Query | Hits |
|---|---|
| `high altitude pulmonary edema AND CYP1B1` | **0** |
| `high altitude pulmonary edema AND CYP11B2` | 6 |
| `high altitude pulmonary edema AND CYP17A1` | 1 |

CYP1B1 has no HAPE literature whatsoever. My expansion of "P1B1 cytochrome genes" was a bad guess, and binding `hgnc:2597` asserted a machine-readable gene-disease susceptibility claim with nothing behind it. Withdrawn.

**I did not rebind to CYP11B2**, even though it's the obvious candidate and the hit counts point that way. Reading "P1B1" as "11B2" is an inference PMID:42254115 doesn't license, and I'd be swapping one guess for another rather than correcting the error. The primary references behind that sentence's `[7,8]` were never consulted — that's the actual gap, and it isn't closed by picking a better-looking symbol.

The entry is now `Unidentified cytochrome P450 locus` with no `gene_term`, keeping `SUSCEPTIBILITY` and `PARTIAL`, and the evidence snippet still quotes the source verbatim. The `notes` record the source wording, the withdrawn binding, the PubMed counts, and why the likely-correct gene is still unbound — so whoever reads those primary papers can settle it directly instead of rediscovering the problem.

On your other two suggestions:

- **`PERIOD_PREVALENCE` standing in for cumulative incidence** — agreed, and agreed it's a schema question rather than a content one. `PrevalenceMeasureEnum` has no `CUMULATIVE_INCIDENCE`, and this pattern will recur for any exposure-triggered disease with an attack rate per episode. Happy to open an NTR if you'd like it tracked.
- **Truncated Classic HAPE snippet** — real, and I've left it. Extending past the en-dash risks the bracket/dash normalisation issues that bit an earlier snippet in this same PR, and you're right it isn't worth a commit alone. It's now bundled into nothing, so it stays as a known cosmetic wart rather than a silent one.

For the record on the rest of CI: `test (3.13)` passed on `a4d9ec349` at 21:14, clearing both earlier failures (the curation-scope guard and the stale-branch environmental check).

---

cmungall:
Took your optional note too — it was a real error, not just phrasing. Fixed in `1f9be45c1`, which resets the approval again; last push from me barring new findings.

**"Necessary and sufficient" was false**, and self-contradictory with this entry's own JAK2 note ("no single locus is necessary or sufficient") two sections away. Ascent is necessary but not sufficient — HAPE takes a susceptible minority of those who ascend, which is the whole point of the susceptibility loci and the `hape_susceptibility_vasoreactivity_basis` gap. I reworded it to say that explicitly rather than just deleting "sufficient", so the reason sits on the page instead of being silently absent.

**The truncated snippet is also fixed, and my earlier reasoning for leaving it was wrong.** I'd said extending past the en-dash risked the normalisation trouble that bit an earlier snippet here. I tested it instead of assuming: the full sentence including `2500–3000 m.` is an exact cache substring and passes the reference validator. So the caution was unfounded and the snippet now runs to the end of the sentence.

Validation on `1f9be45c1`: schema and term validation clean, 44/44 snippets verified, reference validator zero errors, all four ref-based gates clean against `origin/main`, folded-hyphen / snippet-grading / entity-ref / duplicate-key checks clean.

Worth recording what this review round caught, since it's a pattern rather than three unrelated nits. Every finding was something the validators cannot see: two ontology bindings that were internally consistent but anatomically wrong, a gene binding inferred from a typo, an explanation citing a figure its own snippet didn't contain, and a description contradicting another section of the same file. Schema, term and exact-quote validation all passed throughout. The checks verify that a claim is *well-formed and quotable*; they can't verify that it's *true*, or that the file agrees with itself.

---

cmungall:
`test (3.13)` failed again on `check_environmental_evidence`, this time with **26** findings across ~17 files this PR never touches. Same cause as the earlier 9, rebased away again in `e23b3a7d` — but the recurrence makes it worth stating plainly, because it isn't specific to this PR.

**This check cannot be passed reliably by any branch while the environmental-evidence backfill is running.**

The mechanism: the grandfather set is derived live from `origin/main` at the moment CI executes, and contains only exposures that are still *un*cited. When a backfill batch cites a group of exposures on `main`, they leave that set — and every open branch that predates the batch still carries the uncited copies, which then read as *new findings in files the branch never opened*.

The set is draining fast. Measured from this branch over about six hours:

| Time | Grandfathered (uncited on main) | Findings against my tree |
|---|---|---|
| ~19:00 | 46 (committed baseline) | — |
| ~19:30 | 37 | 9 |
| ~23:50 | 11 | 26 |

`main` gained 4 commits in the ~30 minutes between my last rebase and that CI run, one of them a backfill batch. With the runner queue currently running 20–90 minutes, the window between rebasing and CI executing is comfortably long enough for another batch to land, so a branch can be freshly rebased and still fail. I've now rebased twice for this; the second one was clean locally at the moment I ran it.

To be clear about what is and isn't the problem: the backfill is good work and the check is catching something real for genuinely new exposures. The defect is only the comparison basis. Comparing against the **merge base** rather than the live branch tip would keep the guard's purpose — an exposure the PR itself adds without evidence still gets caught — while making the result independent of what merged on `main` after the branch was cut. Happy to open an issue with this write-up if useful; I haven't, since it's your call whether it's worth changing.

State on `e23b3a7d`: rebased onto `7f469bf8a`, no conflicts, post-rebase diff still **50 files, +6160, zero deletions**. Reproduced the failing check locally against current `main` — 11 grandfathered, 0 new — plus the other three ref-based gates clean.

---

cmungall:
Thanks for the approval. One correction, because it would send a fix in the wrong direction if it stands: the shallow-clone explanation for the 26 findings is not right.

The claim is that a shallow checkout truncates the live baseline to 11 grandfathered exposures "instead of the ~37 the full history yields". I tested that. **11 is the true current count on `main`, not a truncation.**

Counting evidence-free `environmental:` entries directly out of the `origin/main` *tree* — bypassing the check script entirely, so no history traversal is involved:

```
evidence-free exposures on origin/main: 11 across 10 file(s)
```

That is exactly what the check reports as its grandfather set. The two agree because the baseline is derived from the file tree at the ref, not by walking history — which is also why clone depth doesn't affect it. For what it's worth my own checkout is shallow as well (`git rev-parse --is-shallow-repository` → `true`), and `git merge-base origin/main HEAD` resolves fine there and returns `7f469bf8a`.

The `~37` figure was accurate several hours ago; it isn't now. The count has been dropping as the backfill lands: **46 → 37 → 11**. That drain *is* the mechanism — each batch that cites a group of exposures removes them from the grandfather set, and any branch predating the batch still carries the uncited copies, which then read as new findings in files it never touched. It's the same cause in both jobs, not two different ones.

The distinction matters for what to change. A fetch-depth or checkout fix wouldn't help, because depth isn't what's wrong. Comparing against the **merge base** rather than the live branch tip would: it keeps the guard's real purpose — an exposure the PR itself adds without evidence is still caught — while making the verdict independent of what merged on `main` after the branch was cut.

Your closing prediction was right, though, and had already happened by the time you wrote it: `main` moved to `7f469bf8a`, the run on `3c19588` failed with exactly those 26, and I rebased onto it as `e23b3a7d`. That push supersedes the head you approved, so this approval will be dismissed — apologies for the churn; it's the only fix available while the comparison is against a moving target.

---

PR #9716, opened 2026-08-27
Mitochondrial disease genetics review: link curated genes into the pathographs

cmungall:
## Summary

Curated genes in mitochondrial disease entries were mostly not reaching the pathographs. A gene is drawn as causing a mechanism only when the mechanism node carries a structured gene binding — and 85% of nodes carried none, so 116 of 276 curated genes (42%) connected to nothing, and in 40 entries the whole `genetic:` block floated free of the mechanism chain it causes. This PR wires 35 of those genes to the node already named for them, adds inheritance to four mtDNA disorders that recorded none, and writes up the review.

Nothing here is a new biological claim. Every gene was already curated with its own PMID and verified snippet in the same file, and every node was already named for that gene's lesion — the two were just never linked.

## The starting question, and what the answer turned out to be

The review was prompted by a suspicion that HGNC and OMIM have incomplete coverage of mitochondrial genes.

**HGNC's coverage is complete.** All 37 mitochondrially encoded genes resolve against `sqlite:obo:hgnc` with correct labels and SO types — 13 protein-coding, 2 rRNA, 22 tRNA — plus 16 control-region features (`MT-HSP1`, `MT-OHR`, `MT-CSB1..3`) we do not yet use. The 14 in `cache/hgnc/terms.csv` are just what has been curated so far, not a limit on what validates.

**HGNC's classification of those genes is not usable, which is a different problem.** No gene group unites the 37: `hgnc.genegroup:1974` covers only the 13 protein-coding genes, tRNAs sit in `843`, rRNAs in `1378`. In the OAK sqlite build every gene-group node is a label-less `rdf:type` stub with no `rdfs:label` and no hierarchy, so a `reachable_from` enum over one would be unlabeled and incomplete. SO type does not separate genomes either — `MT-ND1` and `NDUFS4` are both `SO:0001217`.

So **which genome a gene sits on is not derivable from anything dismech binds to.** The only signal today is the `MT-` symbol prefix, which is a naming convention rather than an assertion.

**OMIM is not part of gene assignment at all.** Its 1,165 occurrences in `kb/` are cross-references and prose. That does mean OMIM cannot be used to audit completeness either, since it lumps most mtDNA phenotypes under allelic variants of one MIM number.

## Why the genes were disconnected

`graph.py` infers the link between a `genetic[]` entry and a mechanism at render time. `_gene_lookup_keys` (lines 176–191) reads exactly three fields on a pathophysiology node — `gene`, `gene_term`, `genes[]` — and builds an index that the genetic entry is matched against (lines 371–390, 625–650).

It does not read the node's `name`, and it does not read its `description`.

So a node called `COX11 Loss and Defective Copper Delivery to the CuB Centre`, whose description says *"Biallelic COX11 variants impair the copper metallochaperone…"*, contributed nothing to that index. The lookup missed and no edge was drawn.

Nothing failed when this happened. `genes:` is optional, so schema validation passes; there is no term to check, so term validation passes; this is not entity-reference syntax (`kind#name`), so `check-entity-refs` passes. The page renders — it just renders a disconnected node, which reads as a sparse entry rather than a broken one.

## Changes

**35 gene links across 34 entries.** Two passes: nodes whose name contains the HGNC symbol, then nodes naming the gene *product* instead (`Twinkle` → TWNK, `DNA Polymerase Gamma` → POLG, `mt-tRNA(Glu)` → MT-TE, `COX4-1` → COX4I1).

| | Before | After |
|---|---:|---:|
| Curated genes reaching no node | 116 / 276 (42%) | 78 / 276 (28%) |
| Entries with at least one orphaned gene | 54 | 23 |
| Entries fully disconnected | 40 | 6 |
| `genetic` → mechanism edges (changed entries) | 16 | 53 |

**Inheritance for four mtDNA disorders that recorded none.** `Leigh_Syndrome` was the consequential one: six curated genes spanning three inheritance modes and no block to distinguish them, so it now carries mitochondrial (`HP:0001427`), autosomal recessive (`HP:0000007`) and X-linked (`HP:0001417`). Also `NARP_syndrome`, `Adult-Onset_Ataxia_and_Polyneuropathy`, and `Reversible_Infantile_Cytochrome_c_Oxidase_Deficiency` — the last recording that its m.14674 variants are *homoplasmic*, so no heteroplasmy threshold applies and the reversible course reflects developmental compensation. Each block carries its own PMID and a snippet verified against the cache.

**A review report** at `docs/reports/mitochondrial-disease-genetics-review-2026-08-27.md`, plus a Gaps row in the design register and 37 history records.

The `cache/` diff is two rows for `GO:0051882`, written by the term validator from a term already present in one of the changed files. Not hand-edited.

## Two things for reviewers

**`Pathophysiology` has both a singular `gene:` and a multivalued `genes:` slot, and this caused a real bug in this PR.** `_gene_lookup_keys` reads either, so a node bound via `gene:` was already linked. The first pass indexed only `genes:`, decided 12 such nodes were unlinked, and added a `genes:` list duplicating the binding already there. Those 12 produced no edge and stated the same fact twice; they were withdrawn in the second commit and are not in this diff. Nothing in the schema, the docs, or the checks says which slot to use, and the KB uses both — worth fixing separately.

**Heteroplasmy has nowhere to go in the schema.** For an mtDNA disease the heteroplasmy fraction and its tissue threshold are what determine penetrance and severity. `ZygosityEnum` has no homoplasmic/heteroplasmic values, and zygosity is the wrong axis anyway. The concept appears in 20 files as free text only — 41 times in one entry. Curators wrote it into prose because there was nowhere else to put it.

This is proposed rather than enacted, since it is a structural schema change:

```yaml
genetic:
- name: MT-TL1 m.3243A>G
  gene_term: {preferred_term: MT-TL1, term: {id: hgnc:7490, label: MT-TL1}}
  genome: MITOCHONDRIAL          # GenomeEnum: NUCLEAR | MITOCHONDRIAL
  heteroplasmy:
    state: HETEROPLASMIC         # HOMOPLASMIC | HETEROPLASMIC | BOTH
    threshold_percent: 60
    threshold_tissue: skeletal muscle
    evidence: [...]
```

`genome` is mechanically backfillable over a closed 37-gene set. `heteroplasmy` needs per-entry curation from the 20 prose files. Both additive and optional, so existing entries validate unchanged. Recorded in the Gaps table of `docs/explanation/design-decisions.md`.

## Not done

- `Kearns-Sayre_Syndrome` and `Pearson_Syndrome` have no `genetic:` block. Defensible for single large-scale deletion syndromes with no one causal gene, but the deletion span is enumerable and all 37 mtDNA genes validate. Needs a source stating the span; the cached GeneReviews (`PMID:20301382`) is abstract-only.
- 6 entries remain fully disconnected — `Charcot-Marie-Tooth_Disease`, `Complex_Hereditary_Spastic_Paraplegia`, `Pheochromocytoma_Paraganglioma`, `Multiple_System_Atrophy`, `Chronic_Intestinal_Pseudoobstruction`, `STAT2_Deficiency`. All multi-gene entries whose nodes are shared consequences, so assigning each gene needs curation judgment rather than a rule.
- `MTO1_Deficiency` lists `MT-TF` as a `MODIFIER`. MTO1 is a nuclear enzyme that modifies mt-tRNAs; MT-TF is its substrate, not a modifier locus. Worth rechecking against the cited evidence.
- GO gaps: mitophagy is used by **0** of the 102 entries despite `PRKN-Related_Juvenile_Parkinson_Disease` being in scope, and complex III assembly appears once against complex IV's 22.

One correction to a gap that looks real but is not: `GO:0000002` "mitochondrial genome maintenance" is **obsolete with no replacement** — GO split it deliberately. The existing use of `GO:0006264` and `GO:0032042` for the depletion syndromes is right. None of the 273 GO terms across these entries is obsolete.

https://claude.ai/code/session_0198M5wr5e8qZx9MkF5HEW3p


cmungall:
Added the `scope-override` label. `test (3.13)` failed the curation-scope guard because this PR touches `kb/` alongside three files outside the curation allowlist:

```
docs/explanation/design-decisions.md
docs/reports/mitochondrial-disease-genetics-review-2026-08-27.md
mkdocs.yml
```

Those three are the write-up of the KB change in this same PR — the review report, its Gaps row in the design register, and the mkdocs nav entry the report needs. Splitting them out would land the report describing the change separately from the change, so the label looked like the right call rather than a second PR. Happy to split if you would rather keep curation PRs strictly to `kb/`.

The failure notification was against `613ed48b`, the first commit, which is superseded. The run on the current head `10cbbef5d` was still queued when the label went on, so it will read the label when it starts — no re-run spent.

---

cmungall:
Thanks — the Reversible Infantile catch was the real one. Fixed in `4396bd42a`, plus the titles. One item I've handled differently from what you proposed; reasoning below.

## Reversible Infantile generic snippet — you were right, swapped

The block is titled "Maternal mitochondrial inheritance (homoplasmic)" and I quoted textbook background that never mentions MT-TE, m.14674, or homoplasmy. It could not support either claim the block makes. Swapped for the OBJECTIVES sentence of the same cached abstract, quoted as printed including the source's space in `m. 14674T>G`, and rewrote the explanation to name both claims. No new reference needed, as you said.

## `reference_title` — added to all six

Taken verbatim from each cache record's `title:` frontmatter.

## Leigh AR snippet — fixed before your review landed, but not the way you suggest

This was already fixed in `34faf286d`, pushed about twenty minutes before the review posted, so you were reading `10cbbef5d`. Same diagnosis as yours, down to the hyphen.

Where I disagree: you write that trimming to a hyphen-free span "drops the very claim it is cited for". That is true of a span taken *after* the break, but the snippet I used is taken *before* it:

```
snippet: SURF1 deficiency is a recessively inherited
```

The proposition "SURF1 deficiency is recessively inherited" is fully present — it is grammatically truncated, not semantically. It is six words, so it clears the `check-snippet-length` floor, and it passes the batched gate.

I did try your suggested fix first. `recessiv` appears exactly once in that paper, inside the hyphenated sentence, so no longer quote is available from it. None of the entry's other cached references states an inheritance mode at all. I then went looking for a different paper, guessed a PMID, got an unrelated ADEM study, and deleted the stray cache record rather than keep guessing — that is the Named Entity Confusion path the evidence SOP warns about, and I would rather ship a truncated-but-exact quote than a confidently wrong reference.

If you know a specific PMID whose abstract states autosomal recessive inheritance of SURF1-related Leigh syndrome, I will swap it in. Failing that I think the current snippet is honest: it says what the paper says, and no more.

## On the suggestions

**Three gene slots, not two.** Your TIMMDC1 observation is the more interesting half of my own finding and I had missed it. `genetic_context.gene` is a *third* slot, and it is the mirror image of the 12 links I withdrew: that node looked bound in the YAML but was invisible to the graph builder, so adding `genes:` is a real fix even though the gene now appears twice. Report updated to say three slots, two of which `_gene_lookup_keys` reads, with the suggestion to either collapse them or teach the lookup to read `genetic_context.gene`.

**PR body.** Already corrected to 35 links / 34 entries / 37 files, at 17:48 and again at 17:52 — before the review job started at 18:07, so I think you read a cached copy. Worth checking the current body rather than taking my word for it.

**`GO:0051882`.** Agreed it is collateral from the reverted PRKN edit. Leaving it: the term is genuinely used on `main` by `PRKN-Related_Juvenile_Parkinson_Disease.yaml`, so the row is one `main` wants, and removing it would mean hand-editing a derived cache.

## Validation on the current head

Four files with inheritance blocks, reference validation in CI's cache-bound mode (`--no-full-text`): `linkml-validate` clean, term validation 4/4, **162/162 snippets verified with no cache-defect-normalization caveat** — both previously-flagged snippets are now genuinely exact. Duplicate keys, entity refs, and the folded-hyphen / snippet-length / title-snippet / snippet-grading gates all clean. No reference-cache churn.

One thing worth adding to the record, since it caused this: running the reference validator *without* `--no-full-text` rewrites tracked cache records — re-quoting `reference_id`, adding `full_text_attempted`, and appending scraped page furniture (`ArticleCASPubMedPubMed CentralGoogle Scholar`) into cache bodies, 305 added lines across 16 files when I hit it. `validate-disorders` uses `--no-full-text` for exactly that reason.

---

cmungall:
`test (3.13)` is red on `15430fa13`, and the failure is not this PR's. It is `check_environmental_evidence.py --against-ref origin/main` reporting 26 evidence-free `environmental:` exposures across 18 files — **every one of which is byte-identical to `origin/main` on this branch**, and none of which is among the 37 files this PR changes.

```
kb/disorders/Dravet_syndrome.yaml:environmental[0]: 'Fever'
kb/disorders/Sickle_Cell_Disease.yaml:environmental[2]: 'Cold Exposure'
kb/disorders/Phenylketonuria.yaml:environmental[1]: 'Mammalian Meat Intake'
…18 files, 26 findings
```

## What actually happened

The check derives its grandfather set live from `kb/` at `origin/main`, so by design the base branch is green by construction. On this run it read only **11** distinct exposures from that ref and flagged the remaining 26 as new. 11 + 26 = 37, and 37 is exactly what the same command reports locally against `origin/main`:

```
$ uv run python scripts/check_environmental_evidence.py --against-ref origin/main
environmental evidence baseline: grandfathered against ref 'origin/main' (37 distinct exposure(s))
OK: no new evidence-free `environmental:` exposures
exit=0
```

So the runner resolved an incomplete `origin/main` and the missing two-thirds of the grandfather set surfaced as new findings. Not shallow-vs-full depth: my clone is also shallow (107 commits) and still read all 37.

The script anticipates the ref being unreadable — it documents falling back to `tests/environmental_evidence_baseline.txt` on "shallow checkouts, forks". What it does not detect is a ref that reads *partially*. A partial read looks like a valid small baseline, so the fallback never engages. The committed baseline holds 179 entries, an order of magnitude more than the 11 the runner saw, which is the signal a sanity check could use.

Most likely trigger is the merge commit `15430fa13` changing the branch topology relative to what the runner fetched. If so it will hit any PR that merges a recent `main`, not just this one.

## What I am not doing

**Not re-running.** I get `403 Resource not accessible by integration` on both `rerun-failed-jobs` and `workflow_dispatch`, so I have no means to spend the one re-run this would warrant.

**Not patching the script here.** The fix belongs in `check_environmental_evidence.py` — treat a ref-derived set that is implausibly smaller than the committed baseline as an unreadable ref and fall back, rather than trusting a partial read. That is an infra change to a shared gate, outside a curation PR's scope, and pushing it would also drop the approval this PR just earned. Proposed shape, for whoever picks it up:

```python
from_ref = baseline_from_ref(ref)
if from_ref and len(from_ref) < len(load_baseline()) * SOME_FRACTION:
    # ref read back implausibly small -- treat as unreadable, use committed baseline
    from_ref = load_baseline()
```

**Not pushing anything.** There is nothing legitimate to push, and an empty commit to kick CI is off the table.

## Everything else on this head is green

Verified locally on the merge commit, not taken from the shepherd's summary:

| Check | Result |
|---|---|
| `test_reference_titles.py` (new on `main`) | 38/38 pass — the six titles this PR added match their cache records |
| `check-duplicate-keys` (37 changed files) | clean |
| `check-entity-refs` | clean |
| `linkml-validate` on `SURF1-Related_Leigh_Syndrome.yaml` | clean |
| Diff vs `main` | 79 files, 1,492 insertions, **0 deletions** |

`SURF1-Related_Leigh_Syndrome.yaml` is worth naming: it is the one file both this branch and `main` edited, so it is the actual duplicate-key risk the merge introduced. My `genes: SURF1` link and `main`'s new `differential_diagnoses` block coexist cleanly, and 36 of my 37 entries are byte-identical to `4396bd42a`.

I will keep this PR watched until it is green, merged, or closed.

---

cmungall:
Correction to my earlier comment (https://github.com/monarch-initiative/dismech/pull/9716#issuecomment-5446599259): my diagnosis of that failure was wrong. @ai4c-agent's staleness explanation is the right one, and the patch I proposed would have fixed a bug that does not exist. Ignore that part of the comment.

I claimed the runner performed a *partial read* of `origin/main` — 11 exposures where the true number was 37. It did not. It read `origin/main` correctly; that ref was simply **newer** than what this branch contained.

The timeline, in UTC:

```
22:26  cca72db79   the main merged into this branch at 15430fa13
23:41  5729e778c   cite 7 bare exposures across 4 entries
23:41  01d5bd8fc   cite 8 bare exposures across 6 entries
23:42  1fb6ea496   disposition 5 entries that are not exposures
23:42  7f469bf8a   split 3 diseases out of environmental blocks
23:48  <-- the failing job ran its environmental step here
```

Those four tranches account for 7 + 8 + 5 + 3 = 23 exposures cited, removed, or relocated, against the 26 flagged. `main` was actively churning environmental entries while the job ran. The grandfather set derived from that newer `main` had 11 evidence-free exposures left; this branch, sitting 15 commits back, still carried the old uncited copies; the 26-item delta is exactly the staleness. `Down_syndrome`'s 'Prenatal Screening' is a clean example — not cited on new `main` but *deleted* by #9743 as not an exposure, while this branch still had it.

Why I got it wrong: I reproduced the check locally and it exited 0, so I concluded CI must have read something the local run didn't. But my local `origin/main` was fetched before those tranches, so it was stale in precisely the same way this branch was. Both sides saw 37, the delta was zero, and it passed. A symmetric staleness that cancels out is not evidence of a correct baseline — it is the same bug on both sides of the comparison, and I read it as exoneration.

The conclusion that mattered still holds and was independently confirmed by the reviewer: none of the 26 flagged files is among this PR's 37, and every one was byte-identical to `main` on this branch. The failure was never this PR's content. But it was staleness, not an infrastructure defect, and the remedy was updating the branch — which is what @ai4c-agent did — not the sanity check I proposed.

That proposal is moot regardless: #9768 retired the ratchet entirely and the workflow now invokes the check bare, so there is no ref-derived baseline left to guard. Running the new check on this head:

```
$ uv run python scripts/check_environmental_evidence.py
OK: no evidence-free `environmental:` exposures (4 dispositioned by a review_notes waiver).
```

This PR's content is unchanged across both merges: 79 files, 1,492 insertions, 0 deletions; 36 of 37 entries byte-identical to `4396bd42a`, the exception being `SURF1-Related_Leigh_Syndrome.yaml` carrying `main`'s `differential_diagnoses` block alongside my `genes:` binding; duplicate keys and entity refs clean; the trimmed Leigh SURF1 snippet intact at line 109.

---

PR #9717, opened 2026-08-27
curate: Add mitochondrial complex I deficiency, nuclear type 24 (MONDO:0032628, NDUFB9)

### LINKED ISSUE #9686: Curate mitochondrial complex I deficiency, nuclear type 24 (MONDO:0032628)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 24** ([MONDO:0032628](https://monarchinitiative.org/MONDO:0032628)) — an autosomal recessive complex I deficiency caused by biallelic NDUFB9 variants, in the nuclear-type numbered series.

Two siblings in this series are already curated (nuclear type 2 / NDUFS8, nuclear type 6 / NDUFS2) — follow their shape and anchor on NDUFB9.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24.yaml`
Nominated by: unknown

kevinschaper:
Adds a knowledge-base entry for mitochondrial complex I deficiency nuclear type 24, caused by variants in NDUFB9, an accessory subunit of the complex I membrane arm.

The unusual thing about this entry is how little it claims. MC1DN24 rests on **one patient**, reported in a 152-patient screening paper that is not open access, with **no further case published in the fourteen years since**. So the entry curates exactly one phenotype and no progression or treatments, and spends its length explaining what is known, what is inferred from yeast and pig, and what simply cannot be cited.

## NEC verdict: PASS

```
PASS  research/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24-deep-research-falcon.md
  canonical gene  : NDUFB9
  mentions        : NDUFB9=63
  top genes       : NDUFB9=63, AR=2, HP=2, NDUFS1=2, NDUFS3=2
  OMIM (MONDO)    : 618245
  OMIM (report)   : -
```

No rival gene, no conflicting OMIM. The report's own reference validation: 6 of 6 resolved, 0 unresolved, 0 off topic. Stating this explicitly because a sibling PR in this batch (#9707, COXPD42) got a **FAIL** on the same provider — the report there was about MRPS2 while calling itself COXPD42 — so a PASS here is a result, not an assumption.

## What is measured, and what is inferred

Only two things are measured in the patient: NDUFB9 protein is reduced, and complex I **amount and activity** both fall and are both restored by expressing wild-type NDUFB9 in the patient's fibroblasts. That both fall together is the signature of an assembly or stability defect rather than a catalytic one, which fits NDUFB9 being an accessory rather than a core subunit.

Everything in between is inferred from other organisms, and both nodes are marked `HYPOTHETICAL` with `INDIRECT_UNKNOWN_INTERMEDIATES` edges:

- **The ACP-anchoring module** comes from *Yarrowia lipolytica*, where NDUFB9 is LYRM3 and anchors mitochondrial acyl carrier protein. That work assigns the ACPM2-LYRM3 module to *assembly and stability* and the sister ACPM1-LYRM6 module to *activity* — which predicts exactly the amount-not-just-activity loss seen in the patient. But it is a two-ACPM yeast enzyme and the mammalian enzyme has one, so the transfer is an assumption with a single point of agreement.
- **The respirasome interface role** is weaker still: two cryo-EM structures of porcine heart supercomplex place NDUFB9 at the complex I/III/IV interfaces, and the consequence of removing it has never been measured in any system. A static structure showing a subunit at an interface is not evidence that losing it disrupts the interface.

Both are curated as a `HUMAN_MODEL_MISMATCH` discussion rather than smoothed into the chain. 5 nodes, 5 edges, one root, all reachable (measured).

## The deep-research report changed a claim in this entry — and then contradicted itself

I originally wrote that no accessible source states the patient's clinical phenotype. The report showed that is **too strong**: Table 2 of the defining paper *does* carry per-patient clinical data. The entry now says the data exist in a paywalled table and are therefore not quotable here, which is the accurate statement.

Nothing from that table is curated, and the report itself supplies the sharpest argument for why not. It read the table from an image and contradicted itself about it — its summary artifact listed **hypertrophic cardiomyopathy** among the findings supported for the patient, while its own body twice recorded that hypertrophic cardiomyopathy and leukodystrophy were **dashes** for that patient and "should not be entered as positive findings". One of those readings is wrong and nothing in the report resolves which. Curating either would have put a fabricated cardiac phenotype, or a negated one, into the export. That episode is written into the `KNOWLEDGE_GAP` discussion as the concrete reason table data must be read from the paper.

## Deliberate omissions a reviewer should not "fix"

- **One phenotype only** (`HP:0011923`, decreased activity of mitochondrial complex I). Complex I deficiency as a class presents as Leigh syndrome, leukoencephalopathy, fatal infantile lactic acidosis, hypertrophic cardiomyopathy or exercise intolerance. Importing any of those would attribute a finding to this entity on the strength of class membership, and it would be indistinguishable in the exported graph from an observed one.
- **No `progression:` and no `treatments:` block**, for the same reason.
- **`zygosity` unset** on the trigger node. It is not reported in the accessible record, and inferring homozygosity from the recessive class would be inventing a fact. `functional_impact_category` is `LOSS_OF_FUNCTION` on the strength of the measured protein reduction plus the complementation rescue, not from a predicted variant consequence.
- **The `inheritance` block says what it is**: the mode of the disease class and the gene's autosomal location, not a confirmed segregation analysis in the index patient.

## Two smaller things worth flagging

**A near-vacuous GO term was caught during curation.** The ACP-module node was first bound to `GO:0005515` (protein binding), which states almost nothing; replaced with `GO:0032981` (mitochondrial respiratory chain complex I assembly), which states the actual claim.

**Module conformance** is declared at `mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress`, on the complex I node only — consistent with #9707 in this batch. The module's own description names primary nuclear respiratory-chain defects as an intended substitution. Its upstream age-related mtDNA-damage node does not apply, and its reactive-oxygen-species arm is not measured in this patient and is not asserted. No existing nuclear mitochondrial-disease entry in the KB declares conformance to that module, so this is a new step in two PRs at once — happy to drop it from both if the reading is wrong.

## Validation

- `just count-verified-snippets` — `Snippets checked: 28/28 verified against cached references`
- `just validate-disorders` (batched CI equivalent) — passed
- `just check-duplicate-keys`, `just check-title-snippets`, `just check-reference-titles`, `check_snippet_length --against-ref origin/main`, `check_folded_hyphens --check`, `just check-cache-order`, `just check-term-cache-integrity`, `just validate-history` — all OK
- `pytest tests/test_data.py -k "conforms_to or subtype_foreign or unique_name or entity_ref"` — 9814 passed

A reference_title audit confirmed all 6 cited `reference_title` values match their cache frontmatter.

Closes #9686

### DIFF
--- a/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24.yaml
+++ b/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24.yaml
@@ -228,6 +228,17 @@ pathophysiology:
     or in any human cell. The correspondence between the yeast prediction and the
     patient measurement is suggestive, not demonstrative.
 
+    That said, the broader assembly-or-stability claim does have patient-level
+    corroboration waiting behind the paywall, and it belongs here as an uncited
+    lead. The deep-research report records that NDUFS1, NDUFS3, NDUFB8 and NDUFA9
+    were all reduced in the patient's fibroblasts and all restored by wild-type
+    NDUFB9. A single accessory-subunit lesion depleting several other subunits,
+    reversibly, is the signature of failed assembly rather than failed catalysis -
+    and unlike the yeast module assignment it is a human-cell observation. It is
+    not curated as evidence because it is not snippet-verifiable; what stays
+    HYPOTHETICAL is specifically the ACP-anchoring mechanism, not the assembly
+    phenotype it is invoked to explain.
+
     A further consequence worth recording: the anchoring depends on a long acyl
     chain on the ACP phosphopantetheine cofactor, so this module is a point at
     which complex I biogenesis is coupled to mitochondrial fatty acid synthesis
@@ -324,13 +335,18 @@ pathophysiology:
   - target: Reduced Complex I Amount and Activity
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
 - name: Reduced Complex I Amount and Activity
-  biological_scale: CELLULAR
+  biological_scale: MOLECULAR
   role: effector
   mechanism_confidence: ESTABLISHED
   conforms_to: "mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress"
   description: >-
-    The measured cellular endpoint, and the best-evidenced statement in this
-    entry: both the amount and the activity of complex I are reduced in the
+    The measured endpoint, and the best-evidenced statement in this entry. Tagged
+    MOLECULAR rather than CELLULAR on the schema's own terms: the substrate is an
+    enzyme complex and the claim is that it is functionally deficient and reduced
+    in abundance, which is the MOLECULAR definition verbatim ("an enzyme is
+    functionally deficient ... substrate is a molecule, complex, or genetic
+    element"). The measurement happens in cells, but the thing measured is the
+    complex. both the amount and the activity of complex I are reduced in the
     patient, and both are restored by expressing wild-type NDUFB9 in the
     patient's fibroblasts. That both fall together is the signature of an
     assembly or stability defect rather than a catalytic one - a defective core
@@ -504,6 +520,13 @@ prevalence:
     percent of cases, and in the screen that found this patient a molecular
     diagnosis was reached in only 18 percent - so NDUFB9 is one gene among many in
     a class where most patients still go unsolved.
+
+    One further rarity datum, recorded as an uncited lead: the deep-research
+    report states that the defining paper reports no NDUFB9 mutations in an
+    Australian cohort of 103 complex I deficiency patients. That is a second
+    ascertained cohort returning zero, which corroborates the rarity above. It is
+    second-hand within the primary report and not snippet-verifiable here, so it
+    is a lead rather than evidence.
   evidence:
   - reference: PMID:22200994
     reference_title: "Mutation screening of 75 candidate genes in 152 complex I deficiency cases identifies pathogenic variants in 16 genes including NDUFB9."
@@ -718,10 +741,19 @@ discussions:
     dash reading is the report's settled one, and it agrees with this entry's
     decision not to curate a cardiac phenotype.
 
+    What that superseding artifact records POSITIVELY for this patient, and which
+    belongs here as an uncited lead rather than being left out: onset under six
+    months, raised blood lactate present, and a progressive course. Recording only
+    the dashes would leave a reader with the impression the table yielded nothing
+    for this patient, which is false - it yielded three positive findings and two
+    negative ones. None is curated as a phenotype, for the reason below.
+
     The reason the table's contents are still not curated is not that the report
     is self-contradictory - it is that a deep-research report is a lead and the
     paper's cached record is abstract-only, so nothing from the table is
-    snippet-verifiable regardless of how confident the extraction is.
+    snippet-verifiable regardless of how confident the extraction is. Recording a
+    lead in prose is not asserting it; that is the same argument this entry makes
+    for the genetic leads, and it applies equally here.
 
     The gap is cheap to close for anyone with journal access: transcribing that
     table row, or publishing a single case report, would supply it.
@@ -802,6 +834,16 @@ notes: >-
   are marked HYPOTHETICAL, their outgoing edges are typed
   INDIRECT_UNKNOWN_INTERMEDIATES, and the mismatch is curated as a
   HUMAN_MODEL_MISMATCH discussion rather than smoothed over.
+
+  GeneReviews was checked and deliberately not tagged. `PMID:26425749` (Nuclear
+  Gene-Encoded Leigh Syndrome Spectrum Overview) is the chapter the sibling
+  entries MC1DN2 and MC1DN6 tag; MC1DN12 and MC1DN31 do not, so repo practice is
+  split. Two reasons not to here: its cached record is a purpose statement naming
+  no genes and no clinical content, so there is nothing to mine without producing
+  a title-as-finding snippet; and MC1DN24 has no documented Leigh phenotype, so
+  tagging a Leigh overview would import exactly the class membership this entry
+  declines to assume. Recorded so the absence reads as a decision, not an
+  oversight.
 
   Search under LYRM3 as well as NDUFB9. The mechanistic literature on this protein
   is published under its LYR-motif family name, and a search restricted to NDUFB9


kevinschaper:
**Correction to this PR's own body, before anyone reviews it.**

The description above says of the `mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress` conformance:

> No existing nuclear mitochondrial-disease entry in the KB declares conformance to that module, so this is a new step in two PRs at once

**That is false**, and I am correcting it rather than leaving it for a reviewer to catch. It was caught on the sibling PR #9707, which carried the same sentence in its `notes:`:

```
$ grep -rln 'mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress' kb/disorders/
kb/disorders/Fontaine_Progeroid_Syndrome.yaml
kb/disorders/Kearns-Sayre_Syndrome.yaml
kb/disorders/MELAS_Syndrome.yaml
kb/disorders/Mitochondrial_DNA_Depletion_Syndrome_14B.yaml
kb/disorders/Pearson_Syndrome.yaml
```

`Mitochondrial_DNA_Depletion_Syndrome_14B` (OPA1, `hgnc:8140` — a nuclear gene) already declares that conformance, from a node named `Combined Respiratory Chain Deficiency`. It was created on 2026-08-18 and predates this branch.

So the conformance here is **precedented, not novel**, and the offer in the body to "drop it from both if the reading is wrong" should be read in that light — the reading now has a precedent behind it. The other conformers are not precedent and I would not cite them as such: Pearson, Kearns-Sayre and MELAS are primary mitochondrial-DNA disorders, and Fontaine progeroid syndrome is not a primary mitochondrial disease.

**No file change is needed on this PR.** I checked, and unlike #9707 this entry's `notes:` never contained the claim — it existed only in the PR body above. The entry is unchanged and still validates at 28/28.

The reason I am posting this rather than quietly letting it go: a survey claim about the corpus is as checkable as an evidence snippet, and nothing in the validation stack checks one. `just validate`, the snippet checker, the term validator and the reference-title gate would all pass a file asserting anything whatever about what other files contain. A single `grep` would have caught it. I am treating "none of the existing entries…" as requiring the same verify-before-asserting discipline as a quotation, and the same applies to a PR body, which is where a reviewer forms their prior.

kevinschaper:
Follow-up to my correction above — **there is now a file change on this PR**, so my "no file change is needed" line is out of date. It was accurate about the module-conformance claim, which really did exist only in the PR body. But having been caught making one unverified survey claim, I ran the same audit across this batch's entries, and this entry turned out to contain two more of the same kind. Both are now fixed in `4d1eb43`.

**1. "the only functional work published in human cells for this gene" — false.** NDUFB9 has been studied functionally in human cells repeatedly, just not in this disease:

```
26641458 | 2015     | Down-Regulation of NDUFB9 Promotes Breast Cancer Cell Proliferation, Metastasis...
42103753 | 2026 May | Complex I protein NDUFB9 is a metabolic vulnerability in triple negative breast...
35122619 | 2022 May | Blockage of NDUFB9-SCD1 pathway inhibits adipogenesis
```

Narrowed to "the only functional work published in **cells from a patient with this disorder**", and the exclusion is now stated in the text so the narrowness is visible rather than accidental — a later reader can see the claim was deliberately scoped, not carelessly broad.

**2. "only one NDUFB9 variant has ever been reported" — wrong on its face.** NDUFB9 has many reported variants; what is singular is the *pathogenic* one in this disorder. Narrowed accordingly, and the follow-on claim softened from "no allele-frequency or segregation precedent" to "essentially no disease-specific precedent", which is what is actually true.

The same audit over the other three entries in this batch (#9696, #9699, #9707) found no comparable claims — their two superficially similar sentences are correctly scoped to the entry's own evidence, so I left them alone rather than rewriting for the sake of it.

Worth stating plainly, because it is the actual generalisation: **all three of these slipped through a validation stack that is otherwise very good at catching fabrication.** `just validate`, `count-verified-snippets`, `validate-terms`, `check-reference-titles` and `check-title-snippets` would all pass a file asserting anything whatever about what other files, or the wider literature, contain. Snippets are held to exact-substring matching; prose claims of the same epistemic weight are held to nothing. I now treat "the only…", "none of the existing…", "has never been…" as requiring a grep or a search before writing them, the same as a quotation.

**Validation after the fix** (the commit touches two `description:` blocks and zero `snippet:` fields): `just validate` OK · `count-verified-snippets` **28/28** · `just validate-disorders` OK · `check-reference-titles`, `check-title-snippets`, `check_snippet_length`, `check_folded_hyphens`, `check-cache-order`, `check-term-cache-integrity`, `check-duplicate-keys` all OK. Rebased onto current `main`; cache diff is 2 rows, strictly additive.

kevinschaper:
Rebased to clear a conflict with `main`, same as the sibling #9707. **Cache-CSV only — no KB content involved** — but the cache diff changed, so flagging it since that is something the review checks.

**Resolution: main-authoritative per CURIE.** Took `main`'s version of `cache/mondo/terms.csv` and the `diseaseterm` enum wholesale, which dropped this branch's `MONDO:0032628` rows, then **re-derived them via `just validate`** (the validator writes the cache from OAK) rather than hand-inserting.

**That distinction matters more than usual here.** This entry's label contains a comma:

```
MONDO:0032628,"mitochondrial complex I deficiency, nuclear type 24",2026-08-27T11:27:38.486016
```

The CSV writer quotes it correctly. A hand-built row would have produced the four-field truncation documented in dismech#7682, where `csv.reader` takes the label as `mitochondrial complex I deficiency` and `retrieved_at` as `" nuclear type 24"` — and the dangerous part is that a later "repair" pass would cement the truncation as clean-looking data. `just check-term-cache-integrity` passes, which is the structural check for exactly that.

The same stale `HP:0005791` row described on #9707 also dropped out here — identical CURIE and label, older `retrieved_at` than main's, unrelated to this entry. Main's copy is kept.

**Verified after resolution:** `just validate-disorders` ✓ · `count-verified-snippets` **28/28** ✓ · `check-cache-order` ✓ · `check-term-cache-integrity` ✓ · `git merge-tree` against current `main` merges clean. Cache diff is now exactly the 2 rows this entry introduces.

kevinschaper:
You are right on every point, and this is the sharpest catch of the batch — the refuting document was in my own diff. Fixed in `1d3f97d`, prose-only, no evidence items added.

I verified the report before changing anything rather than taking the finding on trust:

```
L971  two affected brothers were homozygous for NDUFB9 c.191T>C (p.Leu64Pro),
      while both parents were heterozygous carriers
L980  Patient 35838 and affected brother 46986 ... (pages 3-4)
L1031 ...both parents were heterozygous carriers, establishing recessive segregation
L998  This table supersedes the earlier artifact and aligns the patient 35838
      phenotype rows with the directly inspected Table 2 image
L971  a heterozygous c.140G>T (p.Arg47Leu) ... did not segregate ... not causal
```

**What changed:**

- `inheritance.description` — no longer says "not a confirmed segregation analysis". Segregation *was* performed; the block now explains that its evidence items are what is snippet-verifiable, not the limit of what is known.
- trigger `description` — allele, zygosity and consequence no longer described as unstated; given as an uncited lead.
- `genetic_context` — `zygosity` still unset, but for the right reason: **not snippet-citable, not unknown**. I kept it empty rather than setting `HOMOZYGOUS` because the KGX export carries structured slots without the surrounding prose, so an unverifiable value there is worse than an empty one with an explanation. The note now tells a curator with journal access to set it and cite directly.
- `genetic.notes` — "not described in any accessible source" → it is described. Both the causal allele and the rejected `c.140G>T (p.Arg47Leu)` recorded as uncited leads.
- `diagnosis` — "essentially no disease-specific precedent" → there is a worked negative as well as a worked positive.
- `description` — "a single patient" → a single family, one index case with an affected brother.
- discussion — corrected. The report does **not** contradict itself: artifact-01 is captioned as superseding artifact-00 and records HCM as a dash. It resolves the question, and in favour of not curating a cardiac phenotype. I had cited my own misreading as the reason not to trust the table; the real reason is that nothing in it is snippet-verifiable regardless of extraction confidence, and that is what it now says.

**On your two 🔵 notes.** The `p.Arg47Leu` precedent is in `genetic.notes` as you suggested, and also referenced from the diagnosis node since that is where someone interpreting a new variant will be reading. On GeneReviews `PMID:26425749` — leaving it untagged, and now it is a recorded decision rather than an omission: tagging a Leigh-syndrome chapter would assert exactly the class membership this entry declines to import, and MC1DN24 has no documented Leigh phenotype to justify it.

**You are right that this is the third instance**, and the pattern is worth naming precisely. The first two were claims about *other files* and *the wider literature*. This one is a claim about **a file in the same commit** — which is the least excusable of the three and the easiest to check. My rule was "grep before asserting a negative"; what it needed was "and check your own diff first". The gap you identified stands: no validator reads prose, so a `notes:` block can assert the opposite of the artifact beside it and every gate stays green.

Validation after the fix: `just validate` ✓ · `count-verified-snippets` **28/28** ✓ · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓ · cache diff 2 rows, additive. Rebased onto current `main` (was 3 behind).

Re-requesting review.

kevinschaper:
You are right, and the framing is what makes it land: **the entry's own rationale for narrowing never covered these omissions.** I argued that recording a lead in prose is not asserting it — and then applied that only to genetics. Fixed in `a17d8f4`; all three verified in `artifact-01` before writing.

**1. Table 2 positive rows** — onset under six months, raised blood lactate, progressive course. You are right that reporting only the dashes left the impression the table yielded nothing for this patient. It yielded three positives and two negatives; the discussion now says so.

**2. Secondary subunit destabilization** — and I agree this is the consequential one. NDUFS1, NDUFS3, NDUFB8 and NDUFA9 reduced in patient fibroblasts and restored by wild-type NDUFB9. A single accessory-subunit lesion reversibly depleting several others is the signature of failed assembly rather than failed catalysis, and unlike the yeast module assignment it is a **human-cell** observation.

Your point exposed a scoping error in the `HYPOTHETICAL` marking itself, not just a missing lead. "No study has tested ACP anchoring in an NDUFB9 patient or in any human cell" is true of *ACP anchoring*, but the marking read as though the whole assembly claim were yeast-only. The rationale now scopes it explicitly: what stays HYPOTHETICAL is the ACP-anchoring **mechanism**, not the assembly **phenotype** it is invoked to explain. That is a more accurate statement of where the uncertainty actually sits.

**3. Australian cohort of 103** with no NDUFB9 mutations, in `prevalence` — a second ascertained cohort returning zero. Flagged as second-hand within the primary report.

None curated as evidence or phenotypes; all marked as leads, on the same footing as the genetic ones.

**Both 🔵 items taken:**

- `biological_scale` CELLULAR → **MOLECULAR**. You were right and the schema settles it rather than leaving it to taste — the MOLECULAR definition is the claim verbatim: *"an enzyme is functionally deficient … substrate is a molecule, **complex**, or genetic element"*. The measurement happens in cells; the thing measured is the complex. Reasoning recorded in the node description.
- **GeneReviews loop closed.** `PMID:26425749` checked and deliberately not tagged, with both reasons written into `notes:` — its cached record is a purpose statement with nothing minable without a title-as-finding snippet, and MC1DN24 has no documented Leigh phenotype, so tagging a Leigh overview would import exactly the class membership this entry declines to assume. Noted that repo practice is split (MC1DN2/MC1DN6 tag it, MC1DN12/MC1DN31 do not) so the absence reads as a decision.

For the record on my side: this is the sixth prose-level defect caught in this batch, and the first that was an **inconsistently applied convention** rather than a false statement. Every sentence in the entry was true; the gap was that a rule I invented got applied to one section. That is a failure mode I had no check for — I was auditing individual claims for truth, not conventions for uniform application. Worth naming because it will not show up in any grep.

**Verified:** `just validate` ✓ **28/28** · `just validate-disorders` ✓ · `check-environmental-evidence` ✓ · `check_folded_hyphens` ✓ · rebased onto current `main`; prose plus one enum value, 0 cache rows touched.

Re-requesting review.

Issue #7543, opened 2026-08-01
Upstream MONDO defect: acute biphenotypic leukemia (MONDO:0020322) is_a primary bone lymphoma


kevinschaper:
Found while curating MPAL for #7468 (merged as #7494). Verified with OAK against `sqlite:obo:mondo`:

```
MONDO:0020322  rdfs:subClassOf  MONDO:0004967   acute lymphoblastic leukemia
MONDO:0020322  rdfs:subClassOf  MONDO:0017814   primary bone lymphoma      <-- wrong
MONDO:0020322  rdfs:subClassOf  MONDO:0020743   mixed phenotype acute leukemia
```

## The defect

**`MONDO:0017814` primary bone lymphoma is not a superclass of acute biphenotypic leukemia.** These are unrelated entities — one is an extranodal non-Hodgkin lymphoma arising in bone, the other an acute leukemia of ambiguous lineage. MONDO's own definition of 0020322 (from `NCIT:P378`, WHO 2001) makes this clear:

> "An acute leukemia of ambiguous lineage characterized by blasts which coexpress myeloid and T or B lineage antigens or concurrent B and T lineage antigens."

**The `MONDO:0004967` acute lymphoblastic leukemia parent is also questionable.** A leukemia defined by *ambiguous* lineage should not be a subclass of a lineage-committed entity. That one is arguable; the primary-bone-lymphoma edge is not.

## Why it mattered here

This is not academic — it changed a curation decision. The MPAL entry (#7494) deliberately anchors `disease_term` on **`MONDO:0020743`** (mixed phenotype acute leukemia) rather than the `MONDO:0020322` that issue #7468 tracks. Two reasons, and this defect is one of them:

1. 0020322's definition covers only *biphenotypic* disease, while the four WHO genetic subtypes curated in the entry (MPAL with BCR::ABL1, with KMT2A rearrangement, B/myeloid NOS, T/myeloid NOS) are **children of 0020743**. Anchoring on 0020322 would make the entry's own subtypes its siblings.
2. The spurious primary-bone-lymphoma parent means anything reasoning over 0020322's superclasses inherits a wrong classification.

Two independent curation sessions and both automated review rounds reached the same conclusion on the merits. The residual risk — flagged for @cmungall in #7542 — is that coverage tooling keys on the tracked ID 0020322 and will not see the entry.

## Ask

[...]

Issue #6989, opened 2026-07-25
Term validator cannot re-derive MONDO enum membership: ols:mondo adapter has no ancestor support


cmungall:
## Problem

Adding a valid MONDO subtype term to a disorder entry fails term validation, even when the term is genuinely reachable from the enum's declared root.

Encountered while adding `has_subtypes` to `Carnitine_Palmitoyltransferase_II_Deficiency` (PR #6988):

```
❌ ERROR: Value 'MONDO:0009704' not in dynamic enum 'DiseaseOrSubtypeTerm' (expanded from ontology)
    path: has_subtypes[0].subtype_term.term
    validation: progressive (lazy)
```

`MONDO:0009704` (carnitine palmitoyl transferase II deficiency, myopathic form) *is* reachable — it is `rdfs:subClassOf MONDO:0015515`, which has `MONDO:0000001` as an ancestor. Verified directly:

```bash
uv run runoak -i sqlite:obo:mondo ancestors -p i MONDO:0009704   # includes MONDO:0000001
```

## Root cause

`conf/oak_config.yaml` line 25 maps MONDO to the OLS web adapter:

```yaml
MONDO: ols:mondo
```

OAK's `OlsImplementation` does not implement ancestor traversal:

```
$ uv run runoak -i ols:mondo ancestors -p i MONDO:0009704
NotImplementedError: Cannot execute this using OlsImplementation(...)
```

So on a cache miss the validator has no way to ask "is this reachable from the enum root?" and simply reports the term as invalid. `cache/enums/diseaseorsubtypeterm_*.csv` (~940 rows) and `cache/enums/diseaseterm_*.csv` (~2000 rows) are therefore not a *positive-hit accelerator* over a live check — they are the **only** source of truth, and any MONDO term not already committed there is unusable.

Other prefixes (HP, GO, CL, CHEBI) use `sqlite:obo:*` adapters and do not have this problem.

## Impact

- Any PR adding a MONDO term not already in the cache hits a false-negative validation error.
- The workaround is to hand-append CURIEs to the cache CSVs. This is what earlier subtype-adding PRs appear to have done (e.g. 94e6b830f1, af01b61c7a), but it is undocumented, and it **bypasses the very check the cache exists to support** — the module docstring in `src/dismech/enum_cache.py` warns that "a bad CURIE present in an enum cache can be accepted before the validator asks OAK whether the CURIE is actually reachable".
[...]

PR #9721, opened 2026-08-27
Add Atrioventricular Septal Defect entry (MONDO:0859565)

justaddcoffee:
Curates **atrioventricular septal defect** (`MONDO:0859565`) — a major congenital heart lesion class with no standalone entry and no curation issue.

## Why this target, and the overlap you should know about

AVSD existed only as a `has_subtypes` stub inside `Congenital_Heart_Disease.yaml` (name, description, `subtype_term`, no evidence, no pathophysiology). Its sibling lesions — Tetralogy of Fallot, atrial septal defect, hypoplastic left heart syndrome, double outlet right ventricle, persistent truncus arteriosus — each already have a standalone entry alongside that root. AVSD was the odd one out.

**The honest overlap:** the CHD root carries an `Endocardial Cushion Formation and Endothelial-to-Mesenchymal Transition Failure` node, and its AVSD stub says *"Anchored on the endocardial cushion / EndMT arm of this entry."* So the cushion mechanism is now stated in two places. A shared `kb/modules/` module for atrioventricular septation would be the cleaner factoring — none exists, and the CHD root uses no `conforms_to` at all. Recorded in `notes`.

## Content

- **3 subtypes** — complete, partial, transitional — each evidenced
- **8 pathophysiology nodes**, fully connected, **all 9 causal edges evidenced**:

```
Trisomy 21 dosage ─┐
Multigenic burden ─┴→ AV mesenchymal complex failure → EndoMT failure
                                    ↓
                      Common atrioventricular junction
                                    ↓
                   Left-to-right shunt + valve regurgitation
                          ↓                     ↓
        Pulmonary vascular obstructive    Infantile heart failure
```

- `genetic` (CRELD1, NFATC1, HMGN1, multigene burden), `environmental` (maternal pregestational diabetes), `phenotypes`, `diagnosis`, `treatments`, `progression`

## Two accuracy points

**1. The classic mechanism is out of date.** This entry models AVSD as failure of the whole *atrioventricular mesenchymal complex*, including the dorsal mesenchymal protrusion — not endocardial-cushion fusion alone. `PMID:38884733`:

> While AVSD were historically considered to result from failure of the endocardial atrioventricular cushions to properly develop and fuse, more recent studies have determined that inhibition of the development of other components of the atrioventricular mesenchymal complex can lead to AVSDs as well.

The CHD root's stub still carries the older framing. Worth revisiting there.

**2. I dropped a claim the research report made.** The report attributed EndoMT control to TGF-β, BMP, Notch and Wnt citing `PMID:29549339`. That paper does not establish those roles — it is a *methodological caution* showing that cells scored as endothelial in the standard AV-canal explant assay are predominantly **epicardial**, which undercuts pathway attributions made with that assay. Only its framing statements are cited; the caveat is recorded as `PARTIAL` evidence on the EndoMT node; the pathway claim is not asserted anywhere in the entry.

## Verification

All **16 cited PMIDs** independently re-fetched via `just fetch-reference`; all **40 snippets** verified as exact substrings of the cached records (`Snippets checked: 40/40`). Report PMIDs that couldn't be corroborated as claimed were dropped or re-scoped.

Ontology terms were checked against the local DBs rather than trusted from the report — which caught the report's suggested AV-valve-regurgitation terms being `HP:0031652` *Abnormal aortic valve physiology* and `HP:0000023` *Inguinal hernia*. Correct term is `HP:0034376`.

**Failure to thrive was deliberately omitted.** It's a real feature of infantile AVSD, but no cached reference states it, and the entry doesn't assert phenotypes its own evidence base doesn't support. Noted in `notes`.

## Validation

```
just validate      # schema + term + reference validation all pass, 40/40 snippets
just compliance    # 99.0% global / 100.0% weighted
just validate-graphs  # zero findings for this file
```

`datasets` is the only outstanding recommended field.

## Known follow-ups, in notes

- **`subtype_term` bindings** for complete (`MONDO:0015273`) and partial (`MONDO:0015275`) AV canal are omitted pending the `DiseaseOrSubtypeTerm` enum-cache limitation tracked in **#6989**. Note that `disease_term` binding worked fine here — `MONDO:0859565` auto-appended to `diseaseterm_*.csv` — so the failure really is specific to that one enum.
- **MONDO hierarchy defect:** MONDO places the general term `MONDO:0859565 atrioventricular septal defect` as a **child** of `MONDO:0020290 familial atrioventricular septal defect`, inverting the expected direction. The priority dashboard targets the familial term. This entry binds the general term because it covers sporadic and trisomy-21-associated AVSD, which "familial" excludes. Worth reporting upstream, same spirit as #7543.

### DIFF
--- a/kb/disorders/Atrioventricular_Septal_Defect.yaml
+++ b/kb/disorders/Atrioventricular_Septal_Defect.yaml
@@ -841,6 +841,7 @@ animal_models:
   - Congenital heart defect
   modeled_mechanisms:
   - target: Trisomy 21 Gene Dosage Effect
+    relationship: PARTIALLY_RECAPITULATES
     description: >-
       Tests whether chromosome 21 dosage sensitises the atrioventricular canal such that
       additional loss-of-function alleles precipitate septation failure.


justaddcoffee:
Pushed `9e6a9d6`. All 10 important findings addressed, plus every suggestion.

## Evidence grading and claim–evidence fit

**#1 PMID:41125893** — regraded `HUMAN_CLINICAL` → `MODEL_ORGANISM` in all three places. Confirmed against the cache frontmatter (`Animals`, `Mice`, `Pluripotent Stem Cells`). You were right; I took the study type from the report's framing instead of reading the cache.

**#3 PMID:38313390 removed from the entry entirely.** Both points stand: it's an `Editorial`, and "shunt lesions account for around 75% of all CHDs in Down syndrome" is a CHD-composition statistic that doesn't evidence AVSD haemodynamics. Replaced with `PMID:10812553` for the shunt-ratio haemodynamics and `PMID:1943197` for the PVOD natural history — both from your list.

**#2 Echocardiography** — no longer rests on a snippet that never mentions echo. Now cites `PMID:34196822`. Added **genetic testing** as a second diagnostic modality (`PMID:33142350`).

**#4 Four truncated snippets** replaced with the full sentence from the same cached abstract — including the 17.3% regurgitation figure and the subjectless Ts65Dn quote. Fair hit: I verified snippets were *exact* without checking they were *complete*.

## Structural completeness

| # | Fix |
|---|---|
| 5 | `gene_term` for CRELD1 / NFATC1 / HMGN1 + `relationship_type` + `variant_origin` |
| 6 | `prevalence` — 0.19 per 1000 live births, `BIRTH_PREVALENCE` |
| 7 | `inheritance` — oligogenic (`HP:0010983`) |
| 8 | `environmental` wired via `influences_mechanisms`; added the folate protective factor and the aPR 6.74 effect size I'd dropped |
| 9 | `target_mechanisms` on both treatments — the pathograph now has a treatment join point |
| 10 | Added Left axis deviation and LVOTO phenotypes |

Also took the suggestions: the **CRELD1–calcineurin/NFATc1–VEGF node** (`PMID:24697899`) now connects the genetics to the EndoMT node instead of sitting in `notes` prose; `animal_models` for the Ts65Dn crosses; `synonyms`; a history record; and the pulmonary vascular node moved from `heart` to `UBERON:0002012`.

## Two proposed values I didn't use as given

- **The LVOTO quote doesn't match as written** — `references_cache/PMID_37612667.md` uses en-dashes (`2–7%`, `[5–7]`), not hyphens. Used the en-dash form; it verifies.
- **`HP:0031547` is *Abnormal QT interval***, not a QRS-axis term. Bound `HP:0033568` **Left axis deviation** instead, evidenced by `PMID:10605520` — *"only the five infants with complete AVSD had a superior QRS axis"*, which is nicely specific.

## On the omitted phenotypes

You asked me to treat failure to thrive as a fetch rather than a scoping decision — agreed, so I ran targeted searches for it, tachypnoea, and arrhythmia/AV block. None returned a source stating them for AVSD. `notes` now records that as a **search outcome**, not a standing exclusion, which I think is the honest framing.

Frequency qualifiers are still omitted throughout: the stratified figures (CHF 62% vs 84%, PAH 38% vs 16%) couldn't be traced to a quotable cached source, and a band asserted without its own evidence is the exact failure mode worth avoiding.

## Validation

`just validate` passes — **55/55 snippets** verified (up from 40). Compliance **99.3% global / 100.0% weighted**. `validate-graphs` clean: 9 nodes, 10 edges, all evidenced, all `target_mechanisms` and `influences_mechanisms` targets resolve.

`datasets` remains the only outstanding recommended field.

justaddcoffee:
Pushed `b2b6296`. The CI-red item and all four important findings fixed, plus the suggestions.

## 🔴 The `reference_title` fabrication — my error, and the gate caught it

You're right and the diagnosis is exact. I copied the sentence you quoted in the round-1 review into `reference_title` as well as `snippet`. The snippet was legitimate; the title was not. Corrected to the cached title in all three places, and `check-reference-titles` is now green.

Worth naming the root cause: when I fetched `PMID:24697899`, the fetch output printed no title line, and I didn't go look at the cache frontmatter. Every other reference in this PR I title-checked from the fetch output — this one silently had none and I filled the gap from the wrong source. That gate landing on `main` in `c3d880bd4` caught precisely what it was built for.

## 🟡 Important

**#2 `updated_date` removed.** You were right and my stored note said the opposite — I'd recorded from an earlier PR that the reviewer was wrong to flag this field. CLAUDE.md now deprecates it for new entries. I've corrected my own note as well as the entry.

**#3 Echocardiography** — second attempt at this one, and the first fix didn't actually address it. Now cites the sentence that names echocardiography: *"…detailed 3D echocardiographic evaluation of the atrioventricular junction and valve morphology."* Swapping the reference while keeping an inference-based snippet was not a fix.

**#4 Genetic testing** — now quotes the finding rather than the methods: *"The prevalence of a genetic defect was highest in fetuses with an atrioventricular septal defect (36.8%)"*, which also evidences the yield claim in the description.

**#5 `presence: ABSENT`** deleted — it contradicted the entry's own `PROTECTS_AGAINST` direction.

## 🔵 Suggestions taken

- **Arrhythmia** — you were right that my `notes` claim was inaccurate. `PMID:37612667` was already in this PR and does report it. Added **postoperative atrioventricular block** as a phenotype (`HP:0001678`) citing the pacemaker-indication sentence, and the note now distinguishes post-repair block (curated) from intrinsic preoperative block (not stated anywhere cached).
- **Failure-to-thrive inconsistency** — real contradiction, fixed. The heart-failure node asserted FTT and tachypnoea in prose while `notes` recorded them as unasserted. Description softened.
- **`modeled_mechanisms`** added to Ts65Dn with `fidelity: MODERATE`, `limitations` and evidence, so it joins the pathograph.
- **ECTO** — thank you for checking rather than just repeating the ask; that saved me the search and I agree with the conclusion. Recorded in `notes` that ECTO was searched and both bindings are deliberately omitted, with the folate-fruit-versus-supplement reason.
- **`PMID_38313390.md` removed** from the diff.

## Still open, deliberately

- **`datasets`** — the only compliance gap.
- **The 1000-fold enrichment item graded `MODEL_ORGANISM`.** Your point stands that this leaves a human epidemiological statistic tagged as animal evidence. The grade is correct for the publication, so fixing it properly needs an epidemiology primary source rather than a regrade. Left as-is rather than papered over.
- **MONDO hierarchy inversion** — will file upstream.

## Validation

| Check | Result |
|---|---|
| `just validate` | ✅ **57/57 snippets** |
| `just check-reference-titles` | ✅ was red, now OK |
| `check-entity-refs` / `check-duplicate-keys` / `check-snippet-grading` / `check-title-snippets` / `check-folded-hyphens` | ✅ zero AVSD findings |
| `just validate-graphs` | ✅ zero findings |
| `just validate-history` | ✅ |
| `just compliance` | 99.3% global / **100.0% weighted** |

justaddcoffee:
Correction to my earlier comments on this PR.

I said this PR "caught three separate bogus ontology term suggestions from its own research report," and the approving review repeated that framing. I went back and audited where the bad CURIEs actually came from. **Only some were the report's. Several were mine, and I attributed them to the report.**

## What the report actually got wrong

The AVSD report suggests 22 CURIEs. Every one resolves to a real term — the failure mode is a real ID pointing at the wrong concept, which passes any existence check. Three are wrong for the context:

| Suggested for | CURIE | Actually is |
|---|---|---|
| AV valve regurgitation | `HP:0031652` | Abnormal aortic valve physiology |
| AV valve regurgitation | `HP:0000023` | Inguinal hernia |
| Northwest QRS axis | `HP:0031547` | Abnormal QT interval |

3 of 22, ~14%. The other 19 are correct and in some cases better than I would have picked — `GO:0003198` (epithelial-to-mesenchymal transition involved in endocardial cushion formation) is exactly the right term.

## What was mine

These never appeared in the report. They were my own recall while looking for terms, and I caught them with the same local-DB check, then described them as report errors:

| I reached for | CURIE | Actually is |
|---|---|---|
| CRELD1 | `hgnc:2359` | CRIM1 |
| GATA4 | `hgnc:4170` | GATA1 |
| Echocardiography | `NCIT:C16813` | Malawi |

My error rate was worse than the report's, and the failure mode is identical: plausible-looking IDs that resolve to something unrelated.

## Why it matters here

It changes the mitigation. If the reports were the sole source, "check the report's terms" would cover it. They aren't, so the rule has to be **resolve every CURIE against the local OAK DBs and read the returned label before use, regardless of where it came from** — plus look genes up by symbol rather than by remembered ID. That is what actually caught all of them, on both sides.

The entry itself is unaffected — every term in it was verified before it landed, and this PR's bindings are correct. This is a correction to the provenance claims in my comments, and to the same claim in the `history/` record, which I'll fix.

PR #9724, opened 2026-08-27
curate: Add Intellectual Developmental Disorder Autosomal Recessive 67 (MONDO:0032662, EIF3F)

### LINKED ISSUE #9684: Curate intellectual developmental disorder, autosomal recessive 67 (MONDO:0032662)
Curate a dismech entry for **intellectual developmental disorder, autosomal recessive 67** ([MONDO:0032662](https://monarchinitiative.org/MONDO:0032662)) — an EIF3F-related neurodevelopmental disorder; the translation initiation factor links protein synthesis to intellectual disability, seizures and behavioural features.

Numbered IDD series carries elevated Named Entity Confusion risk — anchor on EIF3F and the OMIM ID before trusting any deep-research report.

Stub: `stubs/Intellectual_Developmental_Disorder_Autosomal_Recessive_67.yaml`
Nominated by: unknown

kevinschaper:
Adds MRT67, the EIF3F-attributed recessive neurodevelopmental disorder (OMIM 618295). Its mechanism is unusual enough to be worth stating plainly: this is a partial failure of **general protein synthesis**, not a defect in a pathway or a neuron-specific protein. Gene-edited stem cells carrying the disease allele have less eIF3f, translate less, and proliferate more slowly.

`entry_type: DISEASE`. MONDO:0032662 is a leaf with one causal gene.

## Read this first: the deep-research report was about a different disease

`just preflight-dr` on the falcon report returned **FAIL**:

```
  canonical gene  : EIF3F
  mentions        : EIF3F=0
  top genes       : PPP1R21=48, AR=3, C12orf4=3, HP=2, TBCK=1
```

PPP1R21 causes **intellectual developmental disorder, autosomal recessive 59** — a different member of the same numbered series. Classic numbered-series Named Entity Confusion.

The report's own citation validation was **clean**: 4/4 references resolved, confabulation rate 0.0. That is the whole point. Every citation in a wrong-entity report resolves, every snippet from it would verify, and no anti-hallucination check in the stack would have caught it. Only the gene-frequency preflight did.

Per policy the report was **discarded, not cherry-picked**, and is not in this PR — neither the report, its citations file, nor the seven PPP1R21/FERRY DOI cache stubs it left behind. Nothing in this entry derives from it. Every reference here was located independently through PubMed E-utilities and fetched with `just fetch-reference`, before either DR run finished.

A second run with the `claude_code` provider **PASSED** (EIF3F 64 mentions; report OMIM 618295 matching the MONDO xref) and is committed. Notably it flagged the same PPP1R21 hazard itself, unprompted, in its own text.

## The mechanism

```
homozygous EIF3F p.Phe232Val lowers        (MOLECULAR)
eIF3f protein abundance
  └─> reduced global translation           (CELLULAR)
        initiation  (↓polysomes, ↓mTOR)
        └─> reduced cell proliferation     (CELLULAR)
              and growth
              └─> impaired brain           (ORGANISM, PROVISIONAL)
                    development and
                    somatic growth
```

The obvious objection — why would a housekeeping translation factor produce a *selective* neurodevelopmental phenotype rather than uniform cellular failure? — has an answer in the literature: eIF3 is not a uniform amplifier. It loads a defined subset of mRNAs concerned with proliferation, growth, cell-cycle control, differentiation and apoptosis. That sentence is quoted on the node.

There is **no TISSUE node**, because there is no tissue-level finding to hang one on: brain imaging in this disease is explicitly nonspecific and was judged not diagnostically useful. The `ORGANISM` node is marked `mechanism_confidence: PROVISIONAL` for the same reason — the step from "translation is slower" to "speech does not develop" is an inference, and the entry says so.

## What is genetically distinctive

Almost every reported patient carries the *same* homozygous missense variant, `c.694T>G p.(Phe232Val)`, on a single shared haplotype across fifteen families — so it arose once. It is also comparatively common: the 7th most frequent EIF3F missense variant in gnomAD (0.07%, rising to 0.21% in Ashkenazi Jewish and 0.12% in non-Finnish European individuals), yet no gnomAD individual is homozygous for it, in contrast to five of the six more frequent EIF3F missense variants.

That pattern is mechanistically informative rather than trivia. EIF3F is strongly LoF-intolerant (pLI 0.97, o/e 0.07), no patient with two truncating alleles has ever been reported, the one compound heterozygote with a frameshift in trans sat at the severe end, and eIF3f-null mouse embryos die shortly after implantation. The economical reading is that p.(Phe232Val) is a **hypomorph that survives where a null would not**.

Both mouse models are curated with that framing and with explicit `limitations`: the null (`PERTURPS`, `LOW` fidelity — a genotype no patient has, producing a qualitatively different outcome) bounds the mechanism from the severe end; the heterozygote (`RECAPITULATES` the translational deficit, `PARTIALLY_RECAPITULATES` the growth deficit) is behaviourally normal and was measured in skeletal muscle, not brain.

## Four discussions, all pointing at real gaps

- **`CONTROVERSY`: how common is epilepsy?** 6 of 9 in the founding cohort; "less frequent" in the larger follow-up. Same allele, so not genotypic — most likely ascertainment, since the founding cases came from an undiagnosed-developmental-disorder cohort analysed by recessive burden while the follow-up was assembled through GeneMatcher after the gene was known. **No frequency band is asserted on the epilepsy phenotype** until a third cohort settles it.
- **`OPEN_QUESTION`: the moonlighting deubiquitinase.** eIF3f is an intrinsic DUB that removes monoubiquitin from activated Notch and is required for Notch signalling. Notch is central to neurogenesis, so the temptation to draw an edge from p.Phe232Val → impaired Notch → neurodevelopmental phenotype is strong and completely unsupported — nobody has tested whether this residue affects DUB activity. It is curated as a question with a proposed experiment, **not** as a pathophysiology edge.
- **`OPEN_QUESTION`: white matter.** Two individuals have leukoencephalopathy on MRI, and eIF3F separately regulates RAN translation, whose products aggregate in demyelinating white matter in SCA8. That is a real published route — and entirely circumstantial here. The larger cohort's negative imaging finding is cited **as `REFUTE` evidence inside the same discussion**, so the tension is on the record rather than smoothed over.
- **`KNOWLEDGE_GAP`: no neural model exists.** Every functional measurement in this disease was made outside the nervous system — undifferentiated iPSCs and mouse skeletal muscle.

## Discipline notes

- **Frequencies only where the source gives one.** GDD, speech delay and intellectual disability are `VERY_FREQUENT` (all individuals); behavioural problems, hearing loss, altered tone and short stature are `FREQUENT` (about half). Seven phenotypes carry no `frequency:` at all.
- **Carrier frequency is not converted to prevalence.** The disease is far rarer than Hardy-Weinberg on a 0.07% allele frequency predicts. The allele frequency is curated on the variant; the entry does not speculate about the discrepancy.
- **PMID:35743796 was fetched and deliberately not curated.** It reports a heterozygous EIF3F variant as one of seven moderate-effect alleles in a proposed *oligogenic* model — a different claim from biallelic EIF3F causing MRT67. Recorded in `notes` so it is not re-imported.
- **The MONDO parent is a cross-reference, not a characterization.** MONDO:0019502 is *non-syndromic* intellectual disability; this entity is consistently described as syndromic. `skos:broadMatch`, with the reason on the mapping.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed |
| `just count-verified-snippets` | `Snippets checked: 55/55 verified against cached references` |
| `just validate-terms` | `✅ Validation passed` |
| `just validate-disorders` (batched sweep) | `✓ All 1 disorder file(s) passed validation.` |
| `just check-reference-titles` | OK |
| `just check-snippet-grading` | OK |
| `check_snippet_length --against-ref origin/main` | OK |
| `just check-title-snippets` / `check_folded_hyphens` / `check-duplicate-keys` | OK |
| `just check-cache-order` / `check-term-cache-integrity` | OK |
| `just validate-history` | `No issues found` |

All 5 distinct evidence `reference_title` values were checked programmatically against the cached frontmatter titles — 0 mismatches. One NCIT label guessed during drafting ("Speech Therapy") was caught as wrong by `just validate` and corrected to the canonical "Speech Language Therapy".

## Tooling note

This literature interleaves numeric citation markers (`[9]`, `[6]`) inside the sentences worth quoting. Those markers are stripped from a snippet before matching but not from the cached text, so three quotes failed on first pass and were trimmed to bracket-free spans — fixed at source rather than by touching the global bracket config.

`just refresh-orphadata` still fails on a drifted sha256 in this checkout. Orphanet has no entry for this concept and none is cited.

Closes #9684

### DIFF
--- a/history/disorders/Intellectual_Developmental_Disorder_Autosomal_Recessive_67/2026-08-27T222638Z-claude-code-fd916e.yaml
+++ b/history/disorders/Intellectual_Developmental_Disorder_Autosomal_Recessive_67/2026-08-27T222638Z-claude-code-fd916e.yaml
@@ -1,0 +1,100 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Intellectual_Developmental_Disorder_Autosomal_Recessive_67
+  path: kb/disorders/Intellectual_Developmental_Disorder_Autosomal_Recessive_67.yaml
+session:
+  id: 2026-08-27T222638Z-claude-code-fd916e
+  timestamp: '2026-08-27T22:26:38Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9684
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9724
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - experimental_models
+  - evidence
+  summary: 'Address second review on PR #9724: correct the Sleep Disturbance frequency
+    band and re-band by counted figures'
+  details: |
+    Response to the second automated review on PR #9724. The blocking item is
+    accepted, and both non-blocking suggestions are taken.
+
+    Blocking item (accepted, and the reviewer was right). Sleep Disturbance was
+    banded FREQUENT. The counted figure is 5/21 = 24%, which is OCCASIONAL (5-29%)
+    under FrequencyEnum. My justification had been the authors' adjective
+    "frequent" in a sentence covering three symptom groups jointly; the reviewer's
+    point is exact, that the other two groups do reach the band (muscle tone 48%,
+    hyper-/myopia 38%) and sleep is the one that does not, and that banding by
+    adjective was the only place in the entry departing from its own stated
+    discipline. Corrected to OCCASIONAL.
+
+    A tooling conflict is worth recording, because the reviewer's literal
+    suggestion cannot be implemented as written. The review pointed at the Table 2
+    row "Sleeping problems5 (21; 24%)" as the source. I first quoted that row, and
+    it verified against the cache, but scripts/check_snippet_length.py rejected it
+    along with the equivalent refractive-error row:
+
+      phenotypes[9].evidence[0].snippet: 4 word(s): 'Sleeping problems5 (21; 24%)'
+      phenotypes[8].evidence[1].snippet: 3 word(s): 'Hyper-/myopia8 (21; 38%)'
+
+    The gate's own message explains why, that such a snippet "is usually lifted
+    from a table whose cells do not survive text extraction", and directs the
+    curator to quote the sentence instead. Grandfathering is unavailable when
+    checking against a ref, so this is a hard block rather than a baseline item.
+
+    The resolution turned out to be better than either the reviewer's suggestion or
+    my earlier position: the paper states all three figures in ordinary prose
+    elsewhere in the discussion, and I had simply not found those sentences because
+    I had only searched the table. All three phenotypes are now banded on quoted
+    prose:
+
+    - Sleep Disturbance, OCCASIONAL, on "Five of 21 individuals (24%) had sleeping
+      problems."
+    - Refractive Error, FREQUENT (unchanged band, better evidence), on "Notably,
+      different ophthalmological findings were observed in up to 38% of affected
+      individuals (8/21) and included hyper-/myopia (38%), strabismus (14%),
+      nystagmus (5%) and coloboma (5%)."
+    - Microcephaly, now FREQUENT, on "Microcephaly was common at the last physical
+      examination (6/19; 32%), and microcephaly (4/10; 40%) or normocephaly (6/10;
+      60%) remained consistent between birth and the last measurement."
+
+    This reverses my earlier decline of the microcephaly band. That decline rested
+    on the premise that the 6/19 count existed only inside mangled table text,
+    which was simply wrong; a clean prose sentence exists and I had not found it.
+    The band is added, and the same sentence additionally supports the entry's
+    description of the growth phenotype as constitutional rather than progressive,
+    since head-size status is stable from birth.
+
+    No phenotype in this entry now bands by adjective. Every frequency value traces
+    to a counted figure quoted from prose.
+
+    Second suggestion (accepted): the three iPSC modeled_mechanisms links now carry
+    readouts for the eIF3f protein level, the bulk translation rate and the
+    proliferation rate, each with direction DECREASED, an interpretation, its own
+    evidence, and a target that repeats its link's target as the schema requires.
+    The translation readout's interpretation records that the measurement is bulk
+    and therefore cannot say which messages are affected, which is the same limit
+    the entry's neural-model knowledge gap turns on.
+
+    Validation after the edits, run and read: just validate (passed);
+    just count-verified-snippets (Snippets checked: 71/71 verified against cached
+    references); just validate-terms (Validation passed); just validate-disorders
+    (All 1 disorder file(s) passed validation); scripts/check_snippet_length.py
+    --against-ref origin/main (OK, the gate that blocked the table-cell quotes);
+    just check-snippet-grading, just check-environmental-evidence,
+    just check-title-snippets, just check-reference-titles, check_folded_hyphens,
+    just check-duplicate-keys, just check-cache-order,
+    just check-term-cache-integrity all OK. Readout target integrity verified
+    programmatically: 3 readouts, all repeating their link target. No cache rows
+    were added this round.
--- a/kb/disorders/Intellectual_Developmental_Disorder_Autosomal_Recessive_67.yaml
+++ b/kb/disorders/Intellectual_Developmental_Disorder_Autosomal_Recessive_67.yaml
@@ -575,6 +575,7 @@ phenotypes:
     term:
       id: HP:0000252
       label: Microcephaly
+  frequency: FREQUENT
   evidence:
   - reference: PMID:33736665
     reference_title: "EIF3F-related neurodevelopmental disorder: refining the phenotypic and expanding the molecular spectrum."
@@ -585,8 +586,20 @@ phenotypes:
       pain, cleft lip/palate, gastrointestinal symptoms and ophthalmological
       symptoms are part of the phenotypic spectrum.
     explanation: >-
-      Adds microcephaly to the recognized spectrum. No frequency is asserted,
-      because the source states only that it is less frequent than short stature.
+      Adds microcephaly to the recognized phenotypic spectrum.
+  - reference: PMID:33736665
+    reference_title: "EIF3F-related neurodevelopmental disorder: refining the phenotypic and expanding the molecular spectrum."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      Microcephaly was common at the last physical examination (6/19; 32%), and
+      microcephaly (4/10; 40%) or normocephaly (6/10; 60%) remained consistent
+      between birth and the last measurement.
+    explanation: >-
+      Six of 19 individuals, 32%, which falls in the FREQUENT band (30-79%). It
+      also records that head-size status is stable from birth, which is why the
+      entry describes the growth phenotype as constitutional rather than
+      progressive.
 - category: Eye
   name: Refractive Error
   description: >-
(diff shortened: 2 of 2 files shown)


kevinschaper:
Pushed a small follow-up before review, from a proactive sweep for the pattern where a snippet's own text names a model system while its `evidence_source` says otherwise. Two items in this entry quote sentences that span human and mouse observations, and both now say so in their `explanation`:

- **PMID:30409806** — *"…KDM5B and EIF3F, and functionally validated them with mouse and cellular models."* Stays `HUMAN_CLINICAL`. The gene–disease association came from a recessive burden analysis of 6,040 human DDD families; the model validation is a trailing clause of the same sentence.
- **PMID:30206144** — *"…in SCA8 human and mouse brains."* Stays `HUMAN_CLINICAL`. The claim being made is about white-matter tropism in people.

One thing worth recording, because it is a real constraint rather than a judgement call. My first attempt was to **split** the PMID:30409806 sentence into a `HUMAN_CLINICAL` item quoting the association clause and a `MODEL_ORGANISM` item quoting the whole sentence. `just check-snippet-grading` rejected it:

```
PMID:30409806 evidence_source=HUMAN_CLINICAL at genetic[0].evidence[0].snippet
              but evidence_source=MODEL_ORGANISM at genetic[0].evidence[1].snippet
```

The shorter quote is a prefix of the longer one, so the gate reads them as one quote carrying two grades — which is exactly what it exists to prevent. Reverted. `CLAUDE.md`'s remedy for a mixed-source paper ("split evidence items so each item gets a single `evidence_source`") works when the two findings are in different sentences; when they share one sentence, the only honest options are to grade on the arm carrying the claim and say so, or to drop the quote. I took the first.

`just check-snippet-grading` now passes, along with `just validate`, `just validate-terms`, `just count-verified-snippets` (**Snippets checked: 55/55 verified against cached references**), and the cache-order and integrity gates. Rebased onto current `main`.

kevinschaper:
All four blocking items are addressed, plus one of the two suggestions. The other suggestion I declined — reason below.

## 1. Two phenotypes quoted but never curated

Fixed. The snippet naming *"muscular hypotonia and/ or hypertonia, ophthalmologic findings and sleeping problems"* as frequent and previously unreported had only its muscle-tone clause modelled. Added **Refractive Error** (`HP:0000539`) and **Sleep Disturbance** (`HP:0002360`).

Both are `FREQUENT`, but on the authors' own word *"frequent"* for that group of findings rather than on the Table 2 counts — those counts exist only inside the mashed run-on rendering of the table, and I did not want a frequency band resting on a mangled quote.

The refractive-error entry also carries the authors' caveat as a `PARTIAL` item: they explicitly note hyper-/myopia *"might not necessarily be related to EIF3F deficiency, as they are common in the general population."* Curating the association without that would overstate it.

## 2. `Audiological Assessment` in the wrong section

Fixed, all three problems. Moved from `treatments:` to `diagnosis:`, rebound from `NCIT:C15747` Supportive Care to **`NCIT:C217379` Diagnostic Audiology Testing** (verified against OAK before use; added to the term caches by `just validate-terms`, not by hand), and re-evidenced on the ~50% hearing-loss frequency — which is the fact that actually justifies assessing routinely — rather than on the frequency-comparison sentence.

## 3. Meningioma signal

Fixed, and curated exactly as you and the DR report framed it: a `KNOWLEDGE_GAP`, **not** a phenotype and **not** a surveillance recommendation. Two of twenty-one in a cohort of median age 8.5 years is hard to dismiss; the authors' own COSMIC check is curated as `REFUTE` evidence — somatic EIF3F variants in 0.9% of 38,579 samples and in **none of 130 meningiomas**. That negative is the kind of result that does not get published twice, and it was previously visible only as a confound inside the regression phenotype's description.

The proposed experiment asks for tumour incidence against a defined denominator with follow-up time, since two cases cannot separate a real predisposition from ascertainment.

## 4. Gene-edited iPSC line

Fixed. Now an `experimental_models:` entry (`IPSC_DERIVED_MODEL`) with three `modeled_mechanisms` links — into both MOLECULAR/CELLULAR nodes and the proliferation node. Your reasoning was the useful part: the "no neural model" knowledge gap does read much more sharply now that the non-neural human model it contrasts against is a curated object rather than evidence text. Each link's `limitations` says what the model cannot show — undifferentiated stem cells; a bulk translation-rate measurement that does not identify *which* transcripts lose efficiency; stem-cell proliferation that is not developmental growth.

## Suggestion taken

**Genetic counseling** added, carrying the quotable 25% recurrence-risk sentence from PMID:39723281, which also records that the counselling was actually delivered to the family rather than merely being derivable.

## Suggestion declined

**A `FREQUENT` band for microcephaly.** The 6/19 count exists only inside the mashed Table 2 text (`Microcephaly later6 (19; 32%)1 (8; 13%)`), and the clean prose sentence gives no band — only that microcephaly is *less frequent than short stature*. Quoting a mangled table row to justify a frequency band seemed worse than leaving the band off, so microcephaly keeps its association evidence and no `frequency:`. Happy to add it if you would rather have the number with an ugly quote.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed |
| `just count-verified-snippets` | `Snippets checked: 65/65 verified against cached references` |
| `just validate-terms` | `✅ Validation passed` (incl. the new `NCIT:C217379`) |
| `just validate-disorders` | `✓ All 1 disorder file(s) passed validation.` |
| `just check-snippet-grading` / `check-reference-titles` | OK |
| `check_snippet_length` / `check-title-snippets` / `check_folded_hyphens` / `check-duplicate-keys` | OK |
| `just check-cache-order` / `check-term-cache-integrity` | OK |

Cache addition is 2 rows for `NCIT:C217379`, machine-derived. Graph and link integrity re-verified after the edits: 4 nodes, every model, treatment and discussion link target resolves.

An `EDIT` history record covering this round is included.

kevinschaper:
You were right, and the diagnosis was sharper than the fix — banding by adjective was the one place this entry departed from its own stated discipline, and sleep was exactly the member of that three-symptom clause that does not reach the band. Corrected to `OCCASIONAL`.

## The suggested fix is blocked by another gate

Worth recording, because it will bite the next person. I first did exactly what the review asked and quoted the Table 2 row. It verified against the cache — and then `check_snippet_length` rejected it, along with the equivalent refractive-error row:

```
phenotypes[9].evidence[0].snippet: 4 word(s): 'Sleeping problems5 (21; 24%)'
phenotypes[8].evidence[1].snippet: 3 word(s): 'Hyper-/myopia8 (21; 38%)'
```

The gate's own message is explicit about why: such a snippet *"is usually lifted from a table whose cells do not survive text extraction. Quote the sentence instead."* Grandfathering is unavailable against a ref, so it is a hard block, not a baseline item.

## The resolution is better than either position

The paper states **all three figures in ordinary prose** in the discussion. I had missed them because I only searched the table — which is also what made my earlier microcephaly decline wrong. All three now band on quoted prose:

| Phenotype | Band | Prose figure |
|---|---|---|
| Sleep Disturbance | `OCCASIONAL` | *"Five of 21 individuals (24%) had sleeping problems."* |
| Refractive Error | `FREQUENT` (unchanged) | *"…observed in up to 38% of affected individuals (8/21)…"* |
| Microcephaly | `FREQUENT` (**added**) | *"Microcephaly was common at the last physical examination (6/19; 32%)…"* |

**This reverses my earlier decline of the microcephaly band.** I had declined it on the premise that the 6/19 count existed only inside mangled table text. That premise was simply wrong — a clean prose sentence exists and I had not found it. The band is added, and that same sentence does extra work: it records that head-size status is stable from birth, which is what licenses the entry describing the growth phenotype as constitutional rather than progressive.

**No phenotype in this entry now bands by adjective.** Every frequency value traces to a counted figure quoted from prose; six phenotypes carry no band at all because no figure supports one.

## Second suggestion also taken

The three iPSC `modeled_mechanisms` links now carry `readouts` — eIF3f protein level, bulk translation rate, proliferation rate — each with `direction: DECREASED`, an interpretation, its own evidence, and a `target` repeating its link's target. The translation readout's interpretation notes the measurement is *bulk* and so cannot say which messages are affected, which is the same limit the neural-model knowledge gap turns on.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed |
| `just count-verified-snippets` | `Snippets checked: 71/71 verified against cached references` |
| `just validate-terms` | `✅ Validation passed` |
| `just validate-disorders` | `✓ All 1 disorder file(s) passed validation.` |
| `check_snippet_length --against-ref origin/main` | **OK** — the gate that blocked the table-cell quotes |
| `just check-snippet-grading` / `check-environmental-evidence` / `check-reference-titles` / `check-title-snippets` | OK |
| `check_folded_hyphens` / `check-duplicate-keys` / `check-cache-order` / `check-term-cache-integrity` | OK |

Readout target integrity verified programmatically: 3 readouts, all repeating their link target. No cache rows added this round.

An `EDIT` history record is included.

Issue #8623, opened 2026-08-16
Concurrent curation PRs can merge into a duplicated YAML key, breaking main


caufieldjh:
Two concurrent curation PRs can add the same top-level block to one KB entry at
different points in the file. Git merges both without a conflict, and the result
is a YAML mapping with a duplicated key.

This has now happened twice with `classifications:`:

- `kb/disorders/Atopic_Dermatitis.yaml` — PRs #8592 and #8580. This turned main
  red: https://github.com/monarch-initiative/dismech/actions/runs/31951438820
- `kb/disorders/Ulcerative_Colitis.yaml` — the same collision, which had been
  sitting on main unnoticed.

## Why it is invisible until it isn't

PyYAML's safe loaders — what `dismech.yaml_io.safe_load` and therefore nearly
all of this codebase uses — accept a repeated mapping key and silently keep the
**last** value. So the entry renders, `pytest` passes, and every export looks
normal. The ruamel-backed `linkml-reference-validator` does not: it raises
`DuplicateKeyError` and aborts with a traceback rather than a validation
message.

Each contributing PR is green against its own base, because neither one has a
duplicate. Only the post-merge push build fails.

## Why the existing checks do not cover it

- `just validate-disorders` (the CI step that did fail) only sees the files a PR
  **changed**. That is why the Ulcerative Colitis copy went unnoticed — nothing
  had touched that entry since its collision.
- The whole-KB sweep `just test-kb` is gated on schema changes, on the stated
  reasoning that only a schema edit can flip a KB file's validity. A merge-born
  duplicate flips validity with no schema edit at all.

## Proposed fix

A cheap whole-KB duplicate-key sweep, ungated, run on every CI build (~8s
offline over 2,245 files), plus the two file repairs.

Issue #8551, opened 2026-08-15
Host attributes and comorbid diseases are modeled as `environmental:` exposures


caufieldjh:
Requested by the reviewer on #8541 so it doesn't live only in a PR body.

## The problem

Several `environmental:` entries are not environmental exposures at all. Two distinct kinds turned up while citing evidence for #8296:

**Host attributes** — properties of the person, not something they are exposed to:

| Entry | Exposure |
|---|---|
| `Glaucoma` | Age |
| `Glaucoma` | Family History |
| `Glaucoma` | African Ancestry |

**Comorbid disease states** — diseases in their own right, several of which have their own `kb/disorders/` entries:

| Entry | Exposure |
|---|---|
| `Peripheral_Artery_Disease` | Diabetes |
| `Peripheral_Artery_Disease` | Hypertension |
| `Peripheral_Artery_Disease` | Dyslipidemia |
| `Gastroesophageal_Reflux_Disease` | Obesity, Hiatal Hernia |
| `Obstructive_Sleep_Apnea` | Obesity |

The GERD and OSA rows are from tranche 1 (#8521) and are listed for completeness — the pattern is not confined to one entry.

## Why it matters

- **`exposure_term` cannot bind them.** ECTO/XCO have no term for "age" or "family history", so these entries are permanently `UNBOUND` in `just environmental-term-audit` — not because nobody has researched them, but because they are the wrong kind of thing. They inflate that gap metric with entries that can never close.
- **They can't join the pathograph honestly.** #8085 links exposures via `influences_mechanisms` with an `environmental_effect` predicate. "Diabetes TRIGGERS peripheral artery disease" is a comorbidity claim wearing an exposure's clothes.
- **#8185 set the precedent** by moving Cirrhosis out of `Metastatic_HCC`'s `environmental:` block into a comorbidity entry, for exactly this reason.

## Suggested work

1. Sweep `kb/disorders/**` for `environmental:` entries that are host attributes or named diseases — a first pass can just look for names matching an existing `Disease.name` or `disease_term`, plus a small hand-list of attribute-like names (age, sex, ancestry, family history).
2. Decide the destination per class. Comorbid diseases → `kb/comorbidities/` entries following the #8185 pattern. Host attributes → possibly `genetic:` (family history, ancestry) or nothing at all, since some are better expressed as `prevalence[].population` stratifiers than as risk factors.
[...]

Issue #8185, opened 2026-08-08
Some environmental: entries are disease states, not exposures (Cirrhosis, Obesity)


caufieldjh:
Split out of the review of #8168, where it was correctly ruled out of scope for a link backfill.

## The problem

Several `environmental:` entries hold **disease states rather than exposures**. Two found while backfilling `influences_mechanisms` links for #8085:

- `Metastatic_HCC.yaml` — **`Cirrhosis`**. This is the clearest case. Cirrhosis *is* one of the chronic liver injury states enumerated by the `Chronic Liver Disease Substrate` node, so the edge `Cirrhosis → Chronic Liver Disease Substrate` is close to tautological: the exposure constitutes the node rather than acting on it from outside. (Chronic hepatitis B/C in the same entry is a genuine exposure and reads fine.)
- `Hyperlipidemia.yaml` — **`Obesity`**. Milder, since obesity is at least environmentally driven and is conventionally modelled as a modifiable risk factor, but it is still a body state rather than something the patient is exposed to.

## Why it matters beyond tidiness

An `environmental:` entry asserts a specific kind of claim: something outside the organism acts on it. When a disease state sits there, three things go slightly wrong at once:

1. Pathograph edges out of it are near-tautological and add no mechanism.
2. It is invisible to anything querying diseases or comorbidities, because it isn't modelled as one.
3. The comorbidity model (`kb/comorbidities/`) already exists for disease-disease relationships and would carry the cirrhosis-HCC relationship with its own evidence and association signals.

## Suggested work

1. Sweep `environmental:` across `kb/disorders/` for entries that are disease states — a cheap first pass is name-matching against MONDO labels and existing `kb/disorders/` stems.
2. Decide the disposition per case: promote to `pathophysiology`, move to `kb/comorbidities/`, or keep where the entry is genuinely about an exposure that merely shares a name.
3. Consider whether the schema or a test should discourage the pattern, and whether `environmental` deserves a short "what belongs here" note in CLAUDE.md.

Not urgent, and each case needs a real look rather than a mechanical rule — `Cirrhosis` in `Metastatic_HCC` is the worked example to start from.

caufieldjh:
Another instance, surfaced by the reviewer of PR #8405.

`kb/disorders/Renal_Artery_Obstruction.yaml` has an `environmental:` entry
`Atherosclerotic cardiovascular risk context`, whose evidence reads:

[...]

PR #9745, opened 2026-08-27
fix(comorbidities): split 3 diseases out of environmental blocks (#8296, #8551)

gingin77:
Three `environmental:` rows named diseases that already have their own `Disease` entries. ECTO has no term that could bind them, so they were permanently unbindable in the exposure-term audit, and any pathograph edge would have been a disease-to-disease claim wearing an exposure's clothes. Each becomes a comorbidity entry following the #8185 precedent.

| Source entry | Row | Migrated to |
|---|---|---|
| `Heart_Failure` | Obesity | `com_Obesity__Heart_Failure` |
| `Polycystic_Ovary_Syndrome` | Obesity | `com_Obesity__Polycystic_Ovary_Syndrome` |
| `Metastatic_Pancreatic_Adenocarcinoma` | Chronic pancreatitis | `com_Chronic_Pancreatitis__Pancreatic_Ductal_Adenocarcinoma` |

## Two targeting decisions worth checking

**The pancreatitis entry points at `Pancreatic_Ductal_Adenocarcinoma`, not at the metastatic entry the row came from.** The cohort evidence measures incident pancreatic cancer in a chronic pancreatitis population with no reference to stage at presentation, and nothing cited addresses the metastatic entity. Attaching a directed risk edge to it would assert more than the sources support — the same correction the #8195 review applied to `com_Liver_Cirrhosis__Hepatocellular_Carcinoma`.

**The PCOS association is `BIDIRECTIONAL`, not `A_BEFORE_B`.** Obesity-driven insulin resistance worsens PCOS, and PCOS-associated androgen excess promotes visceral adiposity in turn. The sources describe this as a self-reinforcing cycle; picking a direction would break a loop they explicitly decline to break.

That entry also carries two qualifiers as a second hypothesis rather than burying them: obesity is *neither necessary nor sufficient* for the PCOS phenotype and the association varies by population, and the amplifying effect tracks **visceral adiposity**, which is common in women with PCOS whether or not they are obese by BMI — so a BMI-defined entry understates the relationship in lean patients.

## On the evidence pitched to match the claim

Following the review on #8973, each entry's `disease_a` is no more specific than its citation. The Framingham signal on the heart-failure entry measured *incident heart failure* without ejection-fraction stratification, while the HFpEF-specific statement comes from a trial population selected for it — both are recorded, and `disease_b` is `Heart_Failure` rather than an HFpEF entry, because HFpEF exists in this KB only as a `has_subtypes` row.

The pancreatitis signal uses the standardized incidence ratio computed after excluding the first two years of follow-up, **16.5**, not the headline 26.3. The unrestricted figure is inflated by cancers already present when pancreatitis was diagnosed, which is reverse causation rather than risk. `metric_type` is `OTHER` because the enum has no SIR value.

## The three exposures sharing those files

Cited in the same pass, because splitting one file across two PRs invites a merge conflict — the exact failure mode #8623 documents.

- `Heart_Failure` **Alcohol Abuse** — PMID:20308914, PMID:38213665
- `Polycystic_Ovary_Syndrome` **Sedentary Lifestyle** — PMID:39861440, graded `PARTIAL` because the review states the benefit of activity rather than harm from sedentariness; the exposure claim is the inverse of what the source reports
- `Polycystic_Ovary_Syndrome` **Prenatal Androgen Exposure** — PMID:31499494 twice, `MODEL_ORGANISM` for the reproducible induction across monkeys, sheep, rats and mice, and `HUMAN_CLINICAL`/`PARTIAL` for the indirect human evidence. That split is why the entry's note hedges: the same review is explicit that fetal sampling constraints are the reason human evidence is indirect.

## Scope notes

**Three rows leave the #8296 backlog structurally rather than by being cited.** #8296 asks that this be stated rather than inferred from a dropping count. The surviving `notes:` in each source file also say where the row went, so the removal is discoverable from the file itself.

**#8551 should stay open.** This is a tranche. Its own review found 33 further disease-named exposures beyond those enumerated by hand, and PR #8973 covers a different set (PAD, GERD, OSA) with no file overlap with this one.

## Validation

- `just validate-comorbidity` on each of the three new files: passed — 3/3, 4/4 and 2/2 snippets verified
- `just validate-comorbidities`: clean across all 25 files
- `just validate-disorders` on the three edited disorder files: passed, **166/176 snippets verified** (10 skipped by prefix)
- `uv run pytest tests/test_data.py -k "comorbid or environmental or Heart_Failure or Polycystic or Pancreatic"`: **2692 passed**
- `just check-entity-refs` whole-KB (2597 files): clean
- `just check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-not4curation`, `check-duplicate-keys`, `validate-history-all`: all clean
- Whole-KB gap: **46 → 40** evidence-free exposures across **33 → 30** files

History records added for all six entries, three disorders and three comorbidities.

Part of #8296. Related: #8551, #8185.


gingin77:
Good catch, and a real miss on my part: I created the comorbidity without checking whether the target file already modelled the same relationship. It did, and its `notes:` recorded an explicit decision not to migrate — so I settled a question someone had deliberately left open, without saying so.

**Kept both and divided the labour explicitly, rather than folding the row in.** The `environmental:` row carries the `influences_mechanisms` link onto Chronic Pancreatic Inflammation plus the KRAS-mutant mouse experiment showing induced pancreatitis is sufficient to release the neoplastic sequence. A comorbidity entry cannot express a pathograph link, so folding it would lose that — the original curator's reasoning was sound. Both `notes:` now cross-reference each other and state which carries what, and the older note records the question as resolved rather than open.

Added PMID:35142721 to the comorbidity as you suggested; a 2022 systematic review with a dose-response in duration is the stronger cancer-specific anchor. I kept the 1993 SIR as the *signal metric* and said why in `notes:` — a standardized incidence ratio measured forward from a defined cohort against expected population rates is a directed `A_BEFORE_B` quantity, which is what that slot holds, whereas the meta-analysis reports incidence rising with duration rather than a single ratio.

All four non-blocking items are in:

- **Migration breadcrumbs were in the wrong field.** Obesity's note was inside `Alcohol Abuse`'s `notes:` in Heart_Failure, and inside `Sedentary Lifestyle`'s in PCOS — an artifact of my having edited the surviving entry rather than the file. Moved to file-level `review_notes:`, and `Metastatic_Pancreatic_Adenocarcinoma` got the one it never had.
- `com_Obesity__Polycystic_Ovary_Syndrome` now has an `association_signals` block (PMID:40352401, 40–85% overweight or obese). Recorded as co-occurrence prevalence rather than a directed risk estimate, since the association is typed `BIDIRECTIONAL` and neither arm has a forward cohort measurement — the width of the range also corroborates the population-dependence qualifier already in the entry.
- That fixes the history record too, which listed `association_signals` as touched when the file had none.
- The SUMMIT snippet is the trial's framing of established risk, not its result. The explanation now says so and points at the Framingham signal as where the measured magnitude lives.

`just validate-comorbidities` clean across all 25 files, `check-entity-refs` clean whole-KB (2627 files), all snippet gates clean, history record added for the PDAC edit.

PR #9427, opened 2026-08-24
refactor: normalise entity-reference prefixes to schema slot names (#9394)

caufieldjh:
Step 4 of #9394, the last piece.

## The problem

`SECTION_KEYS` accepts a singular alias beside most section slots, so `phenotype#Memory Loss` and `phenotypes#Memory Loss` both resolve and the #9193 foreign-key gate is indifferent between them.

That indifference is what let the corpus drift. Two things were false before this PR:

- **Grep-ability.** `phenotypes#` found 82 of 172 phenotype references. The other 90 were written singular, and nothing said so.
- **Derivability.** A reference's prefix could not be worked out from the schema — you had to consult the alias map to know which of two spellings a given entry used.

Note the rationale in the original issue comment is stale: it called this "the load-bearing piece" for *enforcing* the grammar, but was written before `SECTION_KEYS` existed. Both spellings already resolve; the gate passes on 4,818 references with 0 dangling. This buys the two properties above, not correctness.

## The change

274 references across 78 files in `kb/disorders`, `kb/modules`, `kb/comorbidities`:

| from | to | n |
|---|---|---|
| `phenotype#` | `phenotypes#` | 90 |
| `mechanistic_hypothesis#` | `mechanistic_hypotheses#` | 94 |
| `treatment#` | `treatments#` | 61 |
| `subtype#` | `has_subtypes#` | 19 |
| `animal_model#` | `animal_models#` | 4 |
| `experimental_model#` | `experimental_models#` | 2 |
| `discussion#` | `discussions#` | 2 |
| `dataset#` | `datasets#` | 2 |

Note `subtype#` → `has_subtypes#` is the one that is not a pluralisation, and `mechanistic_hypothesis#` flips the 94-reference majority spelling to the 4-reference one. Both follow from the rule; neither is a typo.

**The aliases stay resolvable on purpose.** Nothing outside `kb/` is bound by the canonical spelling, and an entry written before this change — or by hand today — is not a defect. `test_aliases_still_resolve_after_normalisation` pins that half of the decision so a later cleanup cannot quietly drop it.

New: `entity_refs.canonical_kind()` (alias → schema slot; `disease#` and any unmapped prefix returned unchanged), and a gate `test_entity_ref_prefixes_are_schema_slot_names` over disorders + modules + comorbidities. The backlog was driven to zero in the same change, so it needs no baseline — a finding there is always something the branch introduced.

## How the migration was made safe

A text replace over 78 curated files is the kind of change that quietly eats content, so:

[...]

PR #9405, opened 2026-08-24
feat: separate prose outcomes from entity references; whole-section anchors (#9394)

caufieldjh:
Closes #9394. All three questions the issue raised, resolved.

The audit the issue asked for (nobody had run it) turned up a category the issue did not anticipate, and that finding is most of this PR.

**Commits**

| | |
|---|---|
| `d7a06a092` | step 1 — the 10 bare *node* names → `pathophysiology#` |
| `b5c8f9d76` | steps 2–3 — the prose/reference split and whole-section anchors |
| `d440912a4` | `RUF022` sort of `__all__` (CI lint fix; see the note near the end) |
| `5ec56fb79` | the four review suggestions from round 1 |
| `dcb4cdb44` | `experimental_models` → a card anchor (round 2) |
| `d095fd437` | both model sections → their *dedicated* anchors (round 3; corrects a false premise in `dcb4cdb44` — see [the comment](https://github.com/monarch-initiative/dismech/pull/9405#issuecomment-5401396554)) |

## The headline: `would_support`/`would_refute` were doing two different jobs

886 values across `kb/`. **566 are entity references. The other 320, in 51 files, are prose.**

```yaml
# SNIP1-Related_Neurodevelopmental_Disorder.yaml
would_support:
- Increased progenitor apoptosis with caspase-9 activation and measurable
  detained-intron retention in patient organoids would establish that the mouse
  and zebrafish mechanisms operate in human cells at the disease allele dosage.
```

These are not malformed references — they state **what a result would be**, where a reference names **what the result bears on**. The schema documents both slots as taking the `<kind>#<name>` grammar, so the prose sat unvalidated, and it rendered inside `entity-ref-chip`:

```css
.entity-ref-chip { font-family: 'Fira Code', monospace; font-size: 0.72rem; … }
```

A **256-character sentence in a monospace token chip** is a visible defect on those disorder pages today, independent of any validation policy. Confirmed by rendering, not inferred.

### What this does

Adds `supporting_outcome` / `refuting_outcome` for the prose arm and migrates all 320 values. They are the two arms of `decision_criterion`, which is single-valued and so cannot express a support/refute pair — which is *why* curators reached for the reference slots. `would_support` / `would_refute` keep their 566 references and are now documented as references-only.

**The migration is a key rename.** Every block in `kb/` is homogeneous — **370 all-reference, 306 all-prose, 0 mixed** — so no item line is touched and no value reflowed:

```
[...]

PR #9385, opened 2026-08-24
feat: enforce attaches_to and friends as foreign keys (#9193)

caufieldjh:
Closes #9193. Implements items 1–3 of the summary comment on that issue, and nothing else — no reachability, root-count, cycle, or self-edge checks. Those belong to #7855, which this leaves alone. This is the narrow slice previously reported in #7855 (check 3) and #8833.

> **Note:** verification figures below are refreshed for the final head `e95eff40d`. Later commits address the review; see the [round-2](https://github.com/monarch-initiative/dismech/pull/9385#issuecomment-5398596420) and [round-3](https://github.com/monarch-initiative/dismech/pull/9385#issuecomment-5399038118) comments.

## The gap

`test_hypothesis_based_definition_attaches_to_foreign_keys` resolved `definitions[].attaches_to`. Nothing resolved `discussions[].attaches_to`. So a `KNOWLEDGE_GAP` could point at a node that does not exist and pass **every** gate — schema validation, term validation, snippet verification, `check-duplicate-keys`, `check-not4curation`, and all four offline ratchets. The entry looks like it has recorded its uncertainty while the link is broken.

The operation that breaks these is a rename or a node split, which is exactly the operation that puts the author's attention on the nodes rather than on what refers to them (PR #9187).

## Measured backlog, and why this lands as a hard gate

The issue asked for this number first; it was unmeasured across all three reports.

```
total refs = 4,791   resolved = 4,765   skipped = 22   DANGLING = 4
```

over `kb/disorders/` + `kb/modules/` + `kb/comorbidities/`. The four:

| Entry | Dangling ref | Repointed to |
|---|---|---|
| `Autosomal_Dominant_Progressive_External_Ophthalmoplegia` | `pathophysiology#Dominant-Negative Subunit Poisoning and Haploinsufficiency` | `#Reduced Functional mtDNA-Maintenance Capacity` — the node whose description already discusses dominant-negative vs. haploinsufficiency |
| `Chagas_Disease` | `pathophysiology#Chronic myocardial inflammation and cardiomyopathy` | `#Chronic Th1 myocarditis and parasite persistence` |
| `Charcot-Marie-Tooth_Disease_Axonal_Type_2JJ` | `pathophysiology#Misfolded-Protein Aggregation` | `#Misfolded Protein Aggregation in Peripheral Neurons` |
| `Stiff_Skin_Syndrome` | `phenotypes#Absence of inflammatory infiltrate` | `#Sparse or absent inflammatory infiltrate` |

[...]

PR #9751, opened 2026-08-27
feat(environmental): make a recorded failed search a first-class state (#8296)

gingin77:
Some exposures cannot be cited and never will be. A curator searches, finds no abstract that states the claim, and writes that down. Until now `check_environmental_evidence.py` could not tell that apart from an exposure nobody had looked at — so an honest negative result read as an outstanding task, and the #8296 backlog could never reach zero no matter how much curation happened.

This makes "searched, found nothing quotable" a recordable answer.

## The mechanism

An `environmental[]` entry whose `review_notes` **begins** with `Left deliberately uncited.` is treated as dispositioned rather than uncited.

The sentinel is not new. `Gout` → Dehydration and `Myasthenia_Gravis` → Stress already open their `review_notes` with exactly that sentence, from earlier tranches of this issue. This makes an existing convention machine-readable rather than inventing one.

**Most of the new tests are about what must *not* waive**, because that is where a waiver mechanism goes wrong:

- `notes:` cannot waive — it is disease content, written by anyone. A waiver any prose can trigger is not a waiver.
- A mention rather than a prefix cannot waive. "Considered whether this should be left deliberately uncited" is still a violation.
- An empty, absent, or non-string `review_notes` cannot waive.
- An entry carrying both a waiver and real evidence is not reported as waived — the evidence supersedes it, and listing it would suggest the claim is still unsourced.

**Waived entries stay visible.** `--waivers` (and `just list-environmental-evidence-waivers`) lists them, `--all` says how many exist, `--count` reports them on their own line. Dispositioned, not disappeared — otherwise this becomes a way to make the number go down.

## Applied to the four entries that needed it

| Entry | What changed |
|---|---|
| `Contact_Dermatitis` "Mare's milk soap exposure" | Already recorded its failed search, informally, in `description:`. Moved to `review_notes:`. No claim changed — PMID:40921440 is still indexed without an abstract. |
| `Pemphigus_Erythematosus` "Ultraviolet Exposure" | Same move, plus the detail of what was actually searched: the one on-topic paper (PMID:14237594) has no retrievable abstract, and borrowing a pemphigus *foliaceus* citation for an *erythematosus* claim is the approximately-relevant citation the evidence SOP forbids. |
| `Familial_Hypercholesterolemia` "Occupational Factors" | Its own notes already said no direct evidence links occupational factors to FH outcomes. Waived — and three further claims removed. |
| `Congestive_Splenomegaly` "Thrombotic Risk Factors" | Not an exposure at all. Removed. |

Two of those deserve more than a table row.

**`Familial_Hypercholesterolemia` was asserting three things with no source at all.** The description claimed shift work, high-stress occupations, and occupational air pollution each compound FH risk. The general association of each with cardiovascular disease does not establish an interaction with familial hypercholesterolemia, and nothing was cited for one. Removed rather than left as prose. Splitting the bundle into separately citable exposures was considered and rejected — each fragment faces the same absent FH-specific literature — and the `review_notes:` records that, so the next curator does not repeat the attempt.

**`Congestive_Splenomegaly` "Thrombotic Risk Factors" is the #8551 class, not an evidence gap.** It bundled inherited and acquired thrombophilia, myeloproliferative neoplasms, and other prothrombotic states as one "exposure". Thrombophilia is a host state; a myeloproliferative neoplasm is a disease. No ECTO or XCO term could ever bind it, and the mechanism it described — portal or splenic vein thrombosis — is already carried by the entry's own `Portal and Splenic Venous Congestion` pathophysiology node. The proper destinations are a `genetic:` entry and a comorbidity entry. Both are #8551 work, and I deliberately did not attempt them half-way here; the `review_notes:` names them. **This clears one row from the #8296 backlog structurally rather than by citation.**

## What this does not do

It does not skip the search. Every waiver in `kb/` says which searches were run and why they failed. The convention is documented in CLAUDE.md alongside the other evidence gates, with that point made explicitly.

## Validation

- `uv run pytest tests/test_environmental_evidence.py`: **31 passed** (was 21; 10 new)
- `just check-folded-hyphens`: this one **caught a defect in my own prose** — a `high-\nstress` line break inside a folded scalar that would have folded to "high- stress". Reflowed; now clean.
- `just validate-disorders` on the four files: passed
- `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-not4curation`, `check-duplicate-keys`: clean
- `just check-entity-refs` whole-KB (2624 files): clean
- `just validate-history-all`: clean
- Findings **35 → 31**; waivers **2 → 5**

Nine `references_cache/*.md` files were touched by the validator as an incidental side effect (`full_text_attempted` metadata, and one full-text upgrade of PMID:35471728 from a Springer PDF). None is cited by this change, so they are reverted rather than committed — this PR is about a checker and 1375 lines of unrelated full text would swamp the diff. I did screen the upgraded file for the injected-content problem flagged on tranche 11; it is clean.

Part of #8296. Related: #8551.

### DIFF
--- a/history/disorders/Pemphigus_Erythematosus/2026-08-28T015756Z-claude-code-c3b8d1.yaml
+++ b/history/disorders/Pemphigus_Erythematosus/2026-08-28T015756Z-claude-code-c3b8d1.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Pemphigus_Erythematosus
+  path: kb/disorders/Pemphigus_Erythematosus.yaml
+session:
+  id: 2026-08-28T015756Z-claude-code-c3b8d1
+  timestamp: '2026-08-28T01:57:56Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/8296
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - environmental
+  - evidence
+  summary: Cite UV photo-provocation; withdraw the waiver
+  details: |
+    Withdrew the 'Left deliberately uncited.' waiver on the Ultraviolet Exposure entry and cited PMID:6531279 instead. The waiver asserted that no quotable source existed; that was wrong. PMID:6531279 is a photo-provocation study in pemphigus erythematosus specifically, not foliaceus, with a retrievable abstract, and it surfaces on a one-line PubMed title query. Review of PR #9751 found it. The evidence is stronger than the entry's previous hedge: ultraviolet light applied to normal-appearing skin reproduced the clinical and histological changes typical of the disease, with an action spectrum in the UVB range, so the description and effect were rewritten from 'plausible exacerbating factor' to the induction the source reports. A second evidence item cites the action spectrum, which is what makes the ECTO ultraviolet binding specific rather than nominal.
--- a/kb/disorders/Pemphigus_Erythematosus.yaml
+++ b/kb/disorders/Pemphigus_Erythematosus.yaml
@@ -328,22 +328,32 @@ environmental:
       label: exposure to ultraviolet radiation
   presence: PRESENT
   description: >-
-    Ultraviolet exposure is treated as a plausible exacerbating factor, on the
-    photo-distributed pattern of the eruption. This page does not assign a
-    stronger causal claim.
-  review_notes: >-
-    Left deliberately uncited. Targeted searches for a source stating that
-    ultraviolet exposure exacerbates pemphigus erythematosus returned nothing
-    quotable: the one directly on-topic paper, PMID:14237594, has no retrievable
-    abstract text, and the available pemphigus foliaceus literature is about a
-    related but distinct entity. Borrowing a foliaceus citation for an
-    erythematosus claim would be an approximately-relevant citation, which the
-    evidence SOP forbids. The photo-distributed pattern behind the claim comes
-    from the MONDO and Orphanet definitions and from a Falcon deep-research
-    report, neither of which is a quotable primary finding. This text was
-    previously in the description; it was moved so the disposition is
-    machine-detectable rather than reading as an unexamined gap.
-  effect: Exacerbating factor for photo-distributed cutaneous lesions.
+    Ultraviolet exposure induces lesions in this disease. Applied to
+    normal-appearing skin of an affected patient it reproduced the clinical and
+    histological changes typical of the disease in the irradiated area, with an
+    action spectrum in the UVB range. This is the basis for the
+    photo-distributed pattern of the eruption, and for photoprotection as a
+    first-line measure.
+  effect: Induces lesions in photo-exposed skin; action spectrum in the UVB range.
+  evidence:
+  - reference: PMID:6531279
+    reference_title: Photo-induction of lesions in a patient with pemphigus erythematosus.
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Ultraviolet light applied to the normal appearing skin of a patient with pemphigus erythematosus produced clinical and histological changes, typical of the disease, in the irradiated skin."
+    explanation: >-
+      Experimental photo-provocation in a patient with this disease
+      specifically, not with pemphigus foliaceus. Irradiating uninvolved skin
+      reproduced the lesions, which is a stronger design than an observed
+      association: the exposure was applied and the outcome followed.
+  - reference: PMID:6531279
+    reference_title: Photo-induction of lesions in a patient with pemphigus erythematosus.
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "The action spectrum, studied with monochromatic light, was in the range of UVB."
+    explanation: >-
+      Identifies the responsible waveband, which is what makes the ECTO
+      ultraviolet binding on this entry specific rather than nominal.
 treatments:
 - name: Topical Anti-Inflammatory Therapy and Photoprotection
   description: >-
--- a/references_cache/PMID_6531279.md
+++ b/references_cache/PMID_6531279.md
@@ -1,0 +1,31 @@
+---
+reference_id: PMID:6531279
+title: Photo-induction of lesions in a patient with pemphigus erythematosus.
+authors:
+- Deschamps P
+- Pedailles S
+- Michel M
+- Leroy D
+journal: Photodermatol
+year: '1984'
+keywords:
+- Female
+- Humans
+- "Lupus Erythematosus, Systemic/etiology"
+- Middle Aged
+- Pemphigus/etiology
+- Skin/radiation effects
+- Ultraviolet Rays/adverse effects
+publication_types:
+- Journal Article
+content_type: abstract_only
+full_text_attempted: true
+---
+
+# Photo-induction of lesions in a patient with pemphigus erythematosus.
+**Authors:** Deschamps P, Pedailles S, Michel M, Leroy D
+**Journal:** Photodermatol (1984)
+
+## Content
+
+Ultraviolet light applied to the normal appearing skin of a patient with pemphigus erythematosus produced clinical and histological changes, typical of the disease, in the irradiated skin. The action spectrum, studied with monochromatic light, was in the range of UVB.
--- a/scripts/check_environmental_evidence.py
+++ b/scripts/check_environmental_evidence.py
@@ -47,6 +47,14 @@ predates the check -- see the Gout "Dehydration" and Myasthenia_Gravis
 "Stress" entries, which already opened their ``review_notes`` with exactly
 this sentence.
 
+A waiver is a claim that a search happened, so it is only as good as that
+search. One waiver in the first round of dismech#8296's waiver PR asserted
+that no quotable source existed for ultraviolet exposure in pemphigus
+erythematosus; review found PMID:6531279, a photo-provocation study in that
+exact disease with a retrievable abstract, from a one-line PubMed query.
+That entry is cited now, not waived. Search properly before reaching for
+this.
+
 The sentinel alone does not waive. An entry must also record at least
 :data:`MIN_WAIVER_WORDS` words of search after it, and that floor is enforced
 *here*, in the check that runs ungated on every PR -- not only in the test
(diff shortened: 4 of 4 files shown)


gingin77:
Added the `scope-override` label — the curation-scope check was right to stop this, and the coupling is deliberate rather than accidental.

This PR changes `scripts/check_environmental_evidence.py` and touches four `kb/disorders/` entries, which is the shape the guard blocks. I kept them together because the two halves do not mean anything apart: the four entries do not clear the gate without the waiver code, and the waiver code has no effect until something uses it. Reviewing the mechanism without seeing its first four uses would be reviewing an abstraction. `Familial_Hypercholesterolemia` in particular is the case that justifies the design — three of its claims had no source at all and were removed, and the waiver is what lets the remaining honest "no FH-specific literature exists" answer be recorded rather than sitting in a backlog forever.

That is the same shape as #9385, #9405 and #9427, which each landed a new gate together with the `kb/` content conforming to it.

Happy to split it into an infra PR and a content PR if you would rather review them separately — the content half stands alone (those entries are already grandfathered on `main`, so it passes without the code), and Phase 5 of this issue would then stack on the infra half instead. Say the word and I will do that rather than argue for the label.

gingin77:
You're right, and the fix is pushed. The test asserted nothing:

```python
for rel, location, name in cee.scan_waivers():
    assert rel.startswith("kb/")
```

`scan_waivers()` scans `SCAN_DIR` with `rel_to=ROOT`, so every `rel` begins with `kb/` by construction — and, as you say, it cannot be made non-vacuous as named, since it only returns entries whose `review_notes` already starts with the sentinel. Worse than useless: the docstring promised a guard that was not there.

The gap that leaves is precisely the one this PR claims to close. A bare `review_notes: "Left deliberately uncited."` would satisfy `is_waived()`, the checker, and that test, while the docstring and CLAUDE.md both promise a waiver records a negative result rather than skipping the search.

Replaced with `test_committed_kb_waivers_say_what_was_searched`, which reads the YAML directly and asserts the text after the sentinel is substantive. I verified it fails when it should rather than assuming: reducing the `Gout` waiver to a bare sentinel gives

```
kb/disorders/Gout.yaml:environmental[4]: 'Dehydration' (0 words after the sentinel)
```

and restoring it passes. The five committed waivers run 44–112 words against a 20-word floor, so it costs nothing today. There is also a unit test pinning that a bare sentinel yields empty detail, so the helper cannot silently start returning something.

All four optional items are in too:

1. `WAIVER_SENTINEL` now carries the trailing period, so the constant is the sentence everything else describes.
2. `scan_waivers()` no longer runs on the default guard path — it was an unused second rglob and YAML parse of `kb/` on every CI run.
3. Good catch on `PMID:37543519`, and I checked it: CARTaGENE, 122 adults with FH, so my "general cardiovascular findings" was wrong. Rewritten to your sharper reason — five lifestyle habits scored together, no association with LDL-C, occupation never measured. The gap is that the FH literature cannot be decomposed to a sedentary-work claim, not that it is absent.
4. The FH `description` described the entry rather than the exposure; rewritten as disease content.

On the separate `Check curation PR scope` failure: I added `scope-override` and explained the coupling above. Note that the workflow triggers on the default `pull_request` types, which exclude `labeled`, so the label alone does not re-run it — this push does.

gingin77:
Thanks for the approval. One follow-up push since it: CI's lint step caught a ruff isort violation in the test file — `from dismech.yaml_io import safe_load` sat after the `scripts` import. Fixed and verified with `uvx ruff check`.

Worth noting for anyone hitting the same thing: ruff is not part of `just qc`, so the full local gate set passes while CI's `Lint with ruff` step fails. My local run was clean before I pushed.

This push will have dismissed the approval, since `main` has `dismiss_stale_reviews` enabled. Re-requesting review; sorry for the extra round on a whitespace-class change.

gingin77:
Rebased onto current `main`. The `test (3.13)` failure was the stale-branch case again, caused by my own merges: five PRs from this sweep landed while this one sat, and the ratchet derives its grandfather set live from `origin/main`, so entries `main` had just cited still read as new findings from this PR's head. None of the eleven files it listed is touched here.

**After the rebase, this branch reads zero:**

```
0 evidence-free exposure(s) across 0 file(s).
(5 further exposure(s) carry a review_notes waiver recording a failed search; see --waivers.)
```

`main` currently sits at 6. Two of those six — `Gout` → Dehydration and `Myasthenia_Gravis` → Stress — are already correctly curated there and count only because `main`'s copy of the checker does not yet understand waivers. So this PR clears six entries while editing four, which is the mechanism doing what it was built for.

Also re-verified after the rebase: `uvx ruff check` clean, `pytest tests/test_environmental_evidence.py` 32 passed, `--against-ref origin/main` clean.

One request, since `dismiss_stale_reviews` means every rebase costs an approval round: this is the fourth time this branch has been dismissed for a reason unrelated to its content, and `main` is moving fast enough that another lap is likely. If it looks good, merging promptly after approval would avoid a fifth.

Follow-on #9768 is stacked on this and now passes 22/22 including the repo-wide gate, which was red by construction until this branch reached zero.

gingin77:
This is the best catch of the review round, and it is a regression I caused rather than a pre-existing gap. Fixed.

You are right that a waiver's canonical content — "this paper is indexed without an abstract", "this one has no retrievable abstract text" — is precisely the claim `check_source_defect_claims.py` exists to police, and that this PR moved it into the one prose slot the checker does not read. `Contact_Dermatitis` is the concrete loss: its PMID:40921440 claim was adjudicated CONFIRMED from `description:` before this PR and invisible from `review_notes:` after. The claim is true, so nothing wrong landed — but the coverage disappeared silently, and CLAUDE.md would have kept telling curators to put it there.

I re-ran your measurement rather than taking it on trust, and it reproduces exactly:

```
found        226 -> 229
confirmed     62 ->  65
contradicted   5 ->   5
undetermined 141 -> 141
```

Three new claims, all CONFIRMED, no new noise:

```
Arterial_Dissection_Lentiginosis_Syndrome.yaml  review_notes                    CONFIRMED  PMID:7838191
Contact_Dermatitis.yaml                         environmental[12].review_notes  CONFIRMED  PMID:40921440
TRAPPC12-Related_Encephalopathy.yaml            review_notes                    CONFIRMED  PMID:36995918
```

Your point that two of the three are pre-existing is the one that settles it: `review_notes` was already a blind spot, and this PR would have promoted it from oversight to policy. The `PROSE_FIELDS` comment now explains `review_notes` the way it already explained `notes`.

Also re-verified after: `ruff` clean, 32 environmental tests pass, 56 source-defect tests pass, `--against-ref origin/main` clean.

On the two suggestions:

**Pemphigus.** Point taken that erythematosus is usually framed as a localized variant *of* foliaceus rather than a separate entity, which does make a foliaceus citation more defensible than my waiver implies. I am leaving it, for the reason you anticipated — declining to borrow is defensible and now recorded — but I have taken your framing seriously enough that it is worth saying the door is closed by judgement, not absence. That distinction is exactly what the waiver is for, and your comment is now part of the record for whoever revisits it.

**FH entry name broader than its content.** Agreed, and agreed on leaving it: renaming a grandfathered exposure reads as a new finding, which my own new failure message spells out. Noted for whoever revisits it.

gingin77:
No apology needed — this is the best finding of the eight rounds, and it is the one that decides whether the mechanism is real. Fixed by moving the floor into the check.

You identified that the two halves ran in different places: the door (`is_waived()`) runs ungated on every PR, while the lock (the 20-word floor) lived only in pytest, which is path-filtered. Your four-step scenario is exactly right, and the detail that makes it sting is that **this PR was immune only because it edits `project.justfile`** — every PR that actually *uses* the convention from here on will be kb-only, so the floor would never have run on the changes it exists to police.

So `is_waived()` now requires both halves. Changes:

- `waiver_detail()` returns `""` for a bare sentinel and `None` for no waiver — the two need different messages, since "claimed a search and recorded none" is not "did not claim one".
- `is_waived()` requires the sentinel *and* `MIN_WAIVER_WORDS` of recorded search.
- `find_thin_waivers()` names the case in the gate output, so a bare sentinel does not read as an ordinary uncited exposure.

Verified end to end rather than reasoned about. Reducing the `Gout` waiver to a bare sentinel now fails the **ungated script**:

```
EXIT=1
Some of these carry a waiver sentence with nothing behind it.
A waiver must record the search -- at least 20 words saying which searches
were run and why they failed. It is a recorded negative result, not a way
to skip the search:
  kb/disorders/Gout.yaml:environmental[4]: 'Dehydration' (0 word(s) after the sentinel)
```

Three existing tests failed on the change, because their fixture waiver sat below the new floor — the predicate working as intended. Fixtures raised, 35 tests pass. The docstring and CLAUDE.md now state that the floor is enforced and where, rather than describing it as a convention.

Your framing that this is "the difference between the mechanism enforcing what the docs promise and merely claiming to" is the right one, and it applies to the docs too — CLAUDE.md previously said "it is not a way to skip the search" while nothing in the ungated path checked that. It does now.

gingin77:
You are right, and this is the finding that mattered most. Fixed by withdrawing the waiver and citing the paper.

I verified PMID:6531279 independently before acting: *Photo-induction of lesions in a patient with pemphigus erythematosus*, Photodermatol 1984, abstract retrievable, and your one-line title query returns 10 hits including it. Both halves of my waiver were wrong — it was not the only on-topic paper, and a quotable one existed.

That is the one failure mode this mechanism must not have. A waiver asserts that a search happened; an under-searched waiver launders "I did not look hard enough" into "the literature is silent". Mine did exactly that, in the PR that introduces the mechanism.

You are also right that it supports a **stronger** claim than the entry made. Applied exposure followed by outcome is not an observed association, so "plausible exacerbating factor" understated it. The entry now reads as induction, with a second evidence item for the action spectrum — which is what makes the ECTO ultraviolet binding specific rather than nominal.

Two follow-ups I did unprompted:

**Re-searched the `Gout` "Dehydration" waiver** with the same style of query, since if one of mine was under-searched the others deserved a check. 31 hits, all about hydration *methods*, IV hydration as treatment, or self-management behaviours; none states the claim. It holds.

**Recorded the lesson in the module docstring**, so the next curator reaching for a waiver knows it has already been got wrong once, by the PR that introduced it.

Waivers drop from five to four. Gate still reports 0 findings; 35 tests pass; `check-source-defect-claims` unchanged at 229/65/5/141.

One process note, since it explains the approve-then-request-changes oscillation on this PR: I had been manually dispatching `claude-code-review` after each push, not realising the workflow already triggers on `synchronize`. Its concurrency group is keyed on `event_name`, so the dispatch run and the push run do not cancel each other — two independent reviews on the same SHA, sometimes disagreeing. I have stopped dispatching; pushes alone will trigger review from here.

PR #9756, opened 2026-08-27
curate(lathosterolosis): Add SC5D-related lathosterolosis

caufieldjh:
Adds `kb/disorders/Lathosterolosis.yaml` (MONDO:0011816).

## Why this one, next to SLOS

SC5D introduces the C5-6 double bond; DHCR7, the Smith-Lemli-Opitz enzyme, reduces the C7-8 bond one step later. `kb/disorders/Smith-Lemli-Opitz_syndrome.yaml` is already curated, so this entry sits directly upstream of it in the same post-squalene pathway — which is what makes it worth having rather than just another ultra-rare metabolic entry.

## A shortage and a surplus, acting on different arms

Any biosynthetic block produces both. Lathosterolosis is unusual in giving evidence that they cause *different parts of the phenotype*, so the entry keeps them apart as two hypothesis groups instead of collapsing them into "the enzyme is broken":

| Hypothesis group | Arm | Evidence |
|---|---|---|
| Cholesterol deficiency (CANONICAL) | malformations, via impaired hedgehog signalling | The Sc5d-null mouse says the defects "appear to be a result of decreased cholesterol rather than increased lathosterol" |
| Lathosterol accumulation (ALTERNATIVE) | lysosomal lamellar inclusions, progressive liver disease | Storage is not a general feature of cholesterol-synthesis defects, and the hepatic disease improved when plasma lathosterol was lowered *without* the enzyme being restored |

```
Biallelic SC5D Loss of Function
  → Deficient Sterol C5-Desaturase Activity
      ├→ Cholesterol Deficiency → Impaired Hedgehog Signalling → Congenital Malformation Pattern
      └→ Lathosterol Accumulation ├→ Lysosomal Lamellar Inclusion Formation
                                  └→ Progressive Liver Disease
```

The split is testable and clinically consequential: a treatment that lowers pathway flux should help the accumulation arm and do nothing for the cholesterol arm. That is exactly what the statin evidence probes.

Hedgehog is also the reason a *biosynthetic* defect produces a *malformation* syndrome at all — Sonic hedgehog requires covalent cholesterol modification.

## Conflicting treatment evidence, kept rather than resolved

Simvastatin inhibits HMG-CoA reductase upstream of the block, lowering the accumulated sterol without restoring the enzyme. Two reports disagree about whether it helps:

- `SUPPORT` — lowered plasma lathosterol with improvement in transaminases *and* fibrosis grade.
- `REFUTE` — "Our patient's family did not notice any benefit from treatment with simvastatin."

[...]

PR #9755, opened 2026-08-27
curate(PMM2-CDG): Add PMM2-Congenital Disorder of Glycosylation

caufieldjh:
Adds `kb/disorders/PMM2-Congenital_Disorder_of_Glycosylation.yaml` (MONDO:0008907).

## The gap this fills

`kb/groupings/Congenital_Disorders_of_Glycosylation.yaml` already collects fourteen CDG members — ALG1, ALG3, ALG9, ALG12, MPDU1, DK1, MGAT2, COG1, COG7, DPM2, MPI, PGM1, SLC35A2 and a VPS51 entry. PMM2 was not among them, despite being the commonest CDG and the reference phenotype the others are described against. Adding this entry to that grouping&#39;s members is a sensible follow-up.

## One import failure, two downstream arms

PEX-style channel biology this is not: the lesion is a sugar-nucleotide supply problem. PMM2 interconverts mannose-6-phosphate and mannose-1-phosphate, the committed precursor of GDP-mannose and dolichol-phosphate-mannose. Because the defect is at the *shared donor* rather than at one transferase, N-glycosylation fails globally — which is why a single enzyme produces a multisystem disease.

```
Biallelic PMM2 Hypomorphic Variants
  → Reduced Phosphomannomutase 2 Activity        ← epalrestat acts here
      ├→ Depletion of Mannose-1-Phosphate and GDP-Mannose   ← GLM101 BYPASSES here
      └→ Mannose-6-Phosphate Accumulation
              ↓ (both)
        Truncated Lipid-Linked Oligosaccharide Synthesis
          → Global Protein N-Hypoglycosylation
              ├→ Disrupted Cerebellar Development → Ataxia and Developmental Impairment
              ├→ Hypoglycosylation of Coagulation Regulators
              └→ Hepatic Involvement
```

## Two routes to underglycosylation, kept separate

The textbook account is precursor depletion. The zebrafish model could not reliably measure either M1P or GDP-mannose, but found that accumulated mannose-6-phosphate promotes cleavage of the lipid-linked oligosaccharide — and that suppressing M6P synthesis abolished that cleavage.

Both converge on the same node and are not mutually exclusive, so they are curated as two `mechanistic_hypotheses` (`pmm2_precursor_depletion`, CANONICAL; `pmm2_substrate_accumulation`, EMERGING) with the causal edges opting in separately. A future measurement can then support one without silently retiring the other. The depletion evidence item is graded `PARTIAL` for the honest reason: the study reports the metabolites were below reliable quantification.

[...]

PR #9733, opened 2026-08-27
feat(environmental): cite 8 bare exposures across 6 entries (#8296 tranche 16)

gingin77:
Eight environmental exposures across six entries had no citation. Two of them turned out to be claiming more than the literature supports — which is the whole reason #8296 exists, and worth reading before the rest.

## Two entries were overclaiming

**`Central_Serous_Chorioretinopathy` — psychosocial stress.** The note carried two specific numbers with nothing behind them, and both are wrong:

- *"CSC patients demonstrate significantly poorer sleep quality (58.2% vs. 23.9% in controls)"* — a cross-sectional study of 109 patients against 247 age-matched controls, using the Pittsburgh Sleep Quality Index, found **no significant difference** in sleep quality or mood.
- *"Sleep apnea is associated with approximately 5-fold increased CSCR risk"* — the largest propensity-matched cohort (13,084 per arm) reports incident CSCR in 0.4% of patients with obstructive sleep apnea against 0.2% of matched controls. Real, but roughly 2-fold, not 5.

Both claims are removed. Both sources are attached anyway — `REFUTE` for the sleep-quality study, `PARTIAL` for the sleep-apnea cohort — so the removed claims do not quietly reappear in a later pass. The stress and type A personality core of the entry is well supported and is now cited, including against an ocular-disease control group rather than healthy controls.

**`Mixed_Germ_Cell_Tumor` — endocrine-disrupting chemicals.** The description asserted a single association covering diethylstilbestrol, organochlorine pesticides and PCBs together. The evidence separates sharply by agent:

| Agent | Evidence |
|---|---|
| Diethylstilbestrol | Meta-analysis of six studies, summary OR **2.98** (95% CI 1.15–7.67) |
| Organochlorines, PCBs | Nested case-control measuring seven pesticides and thirteen PCB congeners **directly in biobanked maternal pregnancy serum** — **no association** |

The description now states the split. The OCP/PCB study is recorded as `REFUTE` rather than dropped, for the same reason as above.

## The other six

| Entry | Exposure | Source |
|---|---|---|
| `Phenylketonuria` | Mammalian Meat Intake | PMID:42123950 |
| `Phenylketonuria` | Nut Intake | PMID:42123950 |
| `Phenylketonuria` | Dairy Intake | PMID:36296952 |
| `Autoimmune_Polyendocrine_Syndrome_Type_1` | Candida albicans Exposure | PMID:34455138, PMID:20123958 |
[...]

PR #9757, opened 2026-08-27
curate(TRMA): Add thiamine-responsive megaloblastic anemia syndrome

### LINKED ISSUE #9747: Curate thiamine-responsive megaloblastic anemia syndrome (MONDO:0009575)
Claiming `MONDO:0009575` — thiamine-responsive megaloblastic anemia syndrome (Rogers syndrome), biallelic *SLC19A2*.

Stub: `stubs/Thiamine-responsive_Megaloblastic_Anemia_Syndrome.yaml`

**Why it is worth curating.** *SLC19A2* encodes THTR-1, the high-affinity thiamine transporter. Losing it does not abolish thiamine uptake — THTR-2 (*SLC19A3*) and passive diffusion remain — so the disease is a *transport-capacity* defect rather than a deficiency of the vitamin itself. That is exactly why pharmacologic thiamine works: raising the extracellular concentration drives enough thiamine in through the low-affinity routes to restore intracellular thiamine pyrophosphate.

The mechanism therefore explains both the disease and its treatment in one graph, which is the shape dismech is for. It also explains the otherwise odd triad — megaloblastic anemia, non-autoimmune diabetes, and sensorineural hearing loss — as the three cell types least able to tolerate the shortfall: rapidly dividing erythroid precursors (transketolase, so nucleotide synthesis), pancreatic beta cells, and cochlear hair cells.

Two things to get right:

- **The hearing loss is the arm that does *not* respond to thiamine.** Treatment corrects the anemia and often the diabetes; deafness is typically irreversible once established. A single `RESTORES` edge from the treatment to a common downstream node would misstate that — the treatment links need to be per-arm.
- `kb/disorders/Biotin_Thiamine_Responsive_Basal_Ganglia_Disease.yaml` already covers the *SLC19A3*/THTR-2 disease. This entry is its transporter paralogue, so it should cross-reference rather than restate the thiamine biology inconsistently.

caufieldjh:
Adds `kb/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome.yaml` (MONDO:0009575), Rogers syndrome.

## A transport-capacity defect, not a vitamin deficiency

SLC19A2 encodes THTR-1, the *high-affinity* thiamine transporter. Losing it does not abolish thiamine uptake — THTR-2 and passive diffusion remain. So the disease is a shortfall of transport capacity at physiological thiamine concentrations, and pharmacologic thiamine works by raising the extracellular concentration until the residual routes carry enough flux.

That is not a rhetorical framing; it is what the mouse shows. `Slc19a2`-null mice are **normal on a high-thiamine diet** and develop all three arms of the triad only when dietary thiamine is withdrawn — and the diabetes then *resolves* after six weeks of repletion. Genotype alone is not sufficient; the phenotype is a function of capacity relative to supply.

```
Biallelic SLC19A2 Loss of Function
  → Loss of High-Affinity Thiamine Transport          ← environmental: erythromycin EXACERBATES
      → Intracellular Thiamine Pyrophosphate Deficiency
          ├→ Impaired Nucleotide Synthesis in Erythroid Precursors → Megaloblastic Anemia
          ├→ Pancreatic Beta Cell Dysfunction → Diabetes Mellitus
          └→ Selective Inner Hair Cell Loss → Progressive Sensorineural Hearing Loss
```

## Three arms, three different treatment responses

Thiamine carries **three separate `target_mechanisms`** rather than one link to the disease, because the arms genuinely differ:

| Target | Effect | Why |
|---|---|---|
| Intracellular TPP deficiency | `RESTORES` | the step the drug actually acts on |
| Megaloblastic anemia | `RESTORES` | corrects — though macrocytosis persists and anemia recurs off treatment |
| Diabetes mellitus | `MODULATES` | may reduce insulin requirement and delay onset, *in some individuals* |

**There is deliberately no link from thiamine to the hearing-loss node.** GeneReviews says the deafness is irreversible and may not be prevented; an Italian four-patient series reports it resolved in none of them. A link there would assert a benefit the evidence denies.

What carries that arm instead is **cochlear implantation**, curated as `BYPASSES` — it does not restore the hair cells, it routes around them. (The first version left the arm to surveillance alone, which made a correct mechanistic claim read as a therapeutic dead end.)

Pregnancy is the other place the arms diverge in management: because thiamine only `MODULATES` the diabetes arm, periconceptional glycaemic control is the one point where the incompletely-treated arm carries a defined consequence, so it is curated as its own treatment rather than left in prose.

## Why the hearing arm behaves differently

An `INTERPRETATION` discussion proposes the reason, because it changes what &#34;irreversible&#34; means clinically. The three arms differ in whether the affected cells *persist*: erythroid precursors are continuously replaced, and beta cell dysfunction is at least partly functional (it reversed on repletion in the mouse). Inner hair cells are **lost** — the mouse shows selective death of exactly those cells under thiamine restriction — and mammalian cochlear hair cells do not regenerate.

On that reading the deafness is irreversible not because the arm is thiamine-independent, but because its endpoint is cell death rather than cell dysfunction. That predicts treatment started *before* hair cell loss would preserve hearing. The case series&#39; emphasis on early diagnosis is consistent with it without establishing it.

## The deafness mechanism, curated specifically

Not just &#34;sensorineural hearing loss&#34;: selective **inner** hair cell loss with outer cells relatively spared, producing a 40–60 dB ABR threshold elevation against only 10–20 dB by otoacoustic emission. That discrepancy is the functional signature of a lesion at the inner hair cell, and it is why the deficit behaves as an auditory neuropathy.

## Beyond the triad

THTR-1 is expressed outside the three classic tissues, so the disease is not only a triad. Cardiac rhythm disturbance (`HP:0011675`), seizures (`HP:0001250`) and stroke (`HP:0001297`) are curated separately — seizures and stroke as two phenotypes rather than one, since the sources name them as two.

A correction worth stating, because it was wrong in the entry itself rather than only in a commit message. An earlier version evidenced these with a clause that never names them, and its `explanation` asserted that the enumerating half of the sentence could not be quoted because bracketed citations interrupt it and the validator strips them. **That was testable and false** — only the *whole* enumeration spans a bracket. The KB was asserting a limitation the cited paper does not have. The phenotypes now cite the GeneReviews sentence naming neurologic manifestations as a managed domain, which also closes the last unmined clause of GR MANAGEMENT, and the cardiac phenotype quotes the arrhythmia clause directly.

## Erythromycin as an environmental exposure

Ten of 63 screened drugs inhibited SLC19A2 by ≥50%, but only erythromycin was predicted to do so at clinically relevant unbound plasma concentrations — and EHR data show reduced thiamine pyrophosphate in patients prescribed it. Curated with `environmental_effect: EXACERBATES` on the transport node, since a drug hitting the same transporter is a mechanistically specific hazard for someone already dependent on residual transport. The entry states plainly that this evidence comes from the general population, not from TRMA patients.

## The preflight WARN is a false positive, checked not assumed

`preflight-dr` flagged rival gene &#34;HP&#34; at 26% of SLC19A2 mentions. That is the gene-symbol matcher counting the HPO CURIE prefix: the report contains 26 `HP:NNNNNNN` identifiers and **zero** mentions of haptoglobin.

## Relationship to the SLC19A3 entry

`kb/disorders/Biotin_Thiamine_Responsive_Basal_Ganglia_Disease.yaml` covers THTR-2. The relationship is mechanistically informative rather than incidental: the residual uptake that makes TRMA treatable is carried substantially by the very transporter whose own loss causes that other disease.

## Validation

| Check | Result |
|---|---|
| `just validate` | pass (schema + terms + references) |
| Snippets verified | **41/41** |
| Raw-substring re-check bypassing validator normalization | 41/41, 0 non-verbatim |
| `check-reference-titles` / `check-snippet-length` | no new baseline entries |
| declared vs cited references | agree both directions |
| Pathograph | 1 component, no orphans |

GeneReviews `PMID:20301459` is tagged and mined. An earlier version of this description claimed all four sections had been mined; that was not true — the transfusion and pregnancy content was unmined until review caught it, and the claim is corrected here rather than left standing.

Closes #9747

---

### DIFF
--- a/history/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome/2026-08-28T004635Z-claude-code-c5d520.yaml
+++ b/history/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome/2026-08-28T004635Z-claude-code-c5d520.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Thiamine-Responsive_Megaloblastic_Anemia_Syndrome
+  path: kb/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome.yaml
+session:
+  id: 2026-08-28T004635Z-claude-code-c5d520
+  timestamp: '2026-08-28T00:46:35Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9757
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - environmental
+  - genetic
+  - phenotypes
+  - treatments
+  summary: Correct evidence_source on the drug-screen item and three grading calibrations
+  details: |
+    Review-response pass on PR 9757, correcting an evidence_source error I made and three calibration items I had already agreed with. The blocking item: the sole evidence on the erythromycin influences_mechanisms link was graded HUMAN_CLINICAL. It is not a human study. I verified the reviewer's claim against the cached full text rather than accepting it: the methods say sixty-three drugs were screened in isotopic uptake studies in HEK293 cells expressing SLC19A2, and the second half of the quoted sentence, predicted to inhibit at clinically relevant unbound plasma concentrations, is an FDA fu-times-Cmax over IC50 calculation across literature pharmacokinetics rather than an observation. The word clinically qualifies the concentration threshold, not the study design. Regraded IN_VITRO with that reasoning recorded in the explanation. The decisive internal evidence is that this same file already grades the sibling result from the same methods section of the same paper as IN_VITRO, so one paper carried two classifications for one assay. Worth recording why no gate caught this: check-snippet-grading is keyed on the quoted sentence, so it detects the same sentence graded two ways but cannot detect two different sentences from one methods section being graded inconsistently. Also applied three calibrations agreed in earlier rounds. The variant-class summary sentence was regraded HUMAN_CLINICAL to OTHER because it is a Discussion-section summary of prior reports, the same construct this entry already grades OTHER for two other clauses. The two domain-level GeneReviews items backing Seizures and Stroke went SUPPORT to PARTIAL because the quoted sentence names neurologic manifestations as a managed domain, which is broader than either specific claim, and the entry already applies exactly that reasoning to the erythromycin EHR item. The pregnancy treatment modality went BEHAVIORAL to OTHER because glycaemic control in a TRMA pregnancy is usually insulin-mediated, and CLAUDE.md warns specifically against reading supportive-care terms as dietary. Note the deliberate contrast with sibling PR 9756, where BEHAVIORAL is correct for an avoidance instruction: same NCIT action term, different modality, and the discriminator is whether the intervention is a chemical or a change in what the patient does. Validation: schema and terms pass, 41/41 snippets verified, entity-refs and duplicate-keys pass, snippet-grading reports no new divergences, environmental-evidence clean against the current origin/main baseline.
--- a/kb/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome.yaml
+++ b/kb/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome.yaml
@@ -467,14 +467,15 @@ phenotypes:
   evidence:
   - reference: PMID:20301459
     reference_title: Thiamine-Responsive Megaloblastic Anemia Syndrome.
-    supports: SUPPORT
+    supports: PARTIAL
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       Standard treatment for sensorineural hearing loss, diabetes mellitus, and
       ophthalmologic, cardiovascular, and neurologic manifestations.
     explanation: >-
       GeneReviews names neurologic manifestations as a managed domain of this disease -
-      the authoritative statement that they occur beyond the triad.
+      the authoritative statement that they occur beyond the triad. Graded PARTIAL
+      because the sentence names the domain rather than this specific manifestation.
   - reference: PMID:38037112
     reference_title: "An Italian case series' description of thiamine responsive megaloblastic anemia syndrome: importance of early diagnosis and treatment."
     supports: SUPPORT
@@ -501,14 +502,15 @@ phenotypes:
   evidence:
   - reference: PMID:20301459
     reference_title: Thiamine-Responsive Megaloblastic Anemia Syndrome.
-    supports: SUPPORT
+    supports: PARTIAL
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       Standard treatment for sensorineural hearing loss, diabetes mellitus, and
       ophthalmologic, cardiovascular, and neurologic manifestations.
     explanation: >-
       GeneReviews names neurologic manifestations as a managed domain of this disease -
-      the authoritative statement that they occur beyond the triad.
+      the authoritative statement that they occur beyond the triad. Graded PARTIAL
+      because the sentence names the domain rather than this specific manifestation.
   - reference: PMID:38037112
     reference_title: "An Italian case series' description of thiamine responsive megaloblastic anemia syndrome: importance of early diagnosis and treatment."
     supports: SUPPORT
@@ -571,14 +573,16 @@ genetic:
       The Effects of Genetic Mutations and Drugs on the Activity of the Thiamine
       Transporter, SLC19A2.
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: >-
       Approximately half of the 60 SLC19A2 mutations associated with TRMA are missense
       variants (Supplemental Table 1) (5, 25), yet few have been functionally
       characterized.
     explanation: >-
       Gives the variant-class distribution and, importantly, how little of it has been
-      functionally tested.
+      functionally tested. Graded OTHER because the sentence is a Discussion-section
+      summary of prior reports rather than a result of this study - the same construct
+      this entry already grades OTHER elsewhere.
 
 inheritance:
 - name: Autosomal recessive
@@ -615,7 +619,7 @@ environmental:
     - reference: PMID:33649974
       reference_title: "The Effects of Genetic Mutations and Drugs on the Activity of the Thiamine Transporter, SLC19A2."
       supports: SUPPORT
-      evidence_source: HUMAN_CLINICAL
+      evidence_source: IN_VITRO
       snippet: >-
         Ten of 63 drugs inhibited SLC19A2-mediated thiamine transport ≥ 50% at
         screening concentrations; however, with the exception of erythromycin, none was
@@ -623,7 +627,13 @@ environmental:
         concentrations.
       explanation: >-
         Identifies erythromycin as the one screened drug expected to inhibit the
-        transporter at achievable concentrations.
+        transporter at achievable concentrations. Graded IN_VITRO because both halves of
+        the sentence are non-clinical: the screen is an isotopic thiamine-uptake assay in
+        HEK293 cells, and "predicted ... at clinically relevant unbound plasma
+        concentrations" is an FDA (fu*Cmax)/IC50 calculation over literature
+        pharmacokinetics, not a human observation. The word "clinically" qualifies the
+        concentration threshold, not the study design. The human half of this paper is
+        cited separately, from its electronic-health-record result.
   evidence:
   - reference: PMID:33649974
     reference_title: "The Effects of Genetic Mutations and Drugs on the Activity of the Thiamine Transporter, SLC19A2."
@@ -765,7 +775,7 @@ treatments:
     than the generic advice suggests: thiamine only MODULATES the diabetes arm, so this
     is the one place where the incompletely-treated arm has a defined management
     consequence.
-  therapeutic_modality: BEHAVIORAL
+  therapeutic_modality: OTHER
   treatment_term:
     preferred_term: Supportive Care
     term:


caufieldjh:
All seven blocking items are addressed in `243058817`. Snippets went 29 → 36, all verified.

**The one that mattered most.** The entry correctly established that thiamine does not rescue hearing, and then said surveillance carries that arm — which reads as though nothing can be done. Cochlear implantation is what is actually done, and it was documented in a source this PR had already cached. That is a worse failure than a missing treatment, because the omission made the entry's own argument look like a dead end. Added as a `BYPASSES` link on `Progressive Sensorineural Hearing Loss` — the right predicate, since an implant does not restore the hair cells, it routes around them.

**The rest:**

- `reference_title` on all evidence items, via `just backfill-reference-titles`.
- A `diagnosis:` section — the entry had none, despite the triad being what drives testing.
- GeneReviews transfusion and pregnancy content mined. My earlier claim that I had mined all four sections was wrong; that same claim was wrong on PBD3A in the previous round, so it is a pattern rather than a slip, and I have stopped asserting four-section coverage without re-reading the cache section by section.
- Cardiac rhythm disturbance and seizures/strokes added — real reported manifestations that a triad-shaped entry silently drops.
- `HP:0001972` and `HP:0000648` bound; the ocular finding split into optic atrophy and retinal dystrophy, which are separate observations that were bundled.
- SLC19A2 variant-class depth in `genetic:`.

**Not changed:** prevalence and histopathology. Both were flagged non-blocking and I have no cached source that supports either without new fetching, so adding them would mean either an unsourced claim or widening this PR. Better as a follow-up.

The PR body above still describes the pre-review state in two places (snippet count, and the four-section claim); I am correcting it now rather than leaving a false statement in the record.

---

---

caufieldjh:
All four blocking items closed in `9139b57f9`. Snippets 36 → 41.

**Item A is the one worth dwelling on, because you caught me asserting something false about a source.** My `explanation` said the enumerating clause couldn't be quoted because bracketed citations interrupt it and the validator strips them. You checked. Only the *whole* enumeration spans a bracket — `seizures and strokes` and `such as congenital heart defects and arrhythmias` are each contiguous and bracket-free. One grep would have shown it, and I didn't run it.

The placement makes it worse than a wrong commit message: the false reasoning went into the entry's `explanation` field, so the KB itself asserted a limitation that the cited paper does not have. A reader would have had no way to know it was invented.

One correction in the other direction, since it changes the fix: the bare `seizures and strokes` fragment is still unusable, but not for the reason I gave — it is three words, below the five-word floor `check-snippet-length` enforces, so quoting it would trade one failing gate for another. Your preferred option was the right one anyway. The neurologic phenotypes now cite the GeneReviews sentence naming neurologic manifestations as a managed domain: authoritative, bracket-free, and it closes the last unmined clause of GR MANAGEMENT. The cardiac phenotype additionally quotes the arrhythmia clause directly, since that one *is* long enough.

**B and C:** `Seizures and Strokes` split in two, and all three new phenotypes bound — `HP:0011675`, `HP:0001250`, `HP:0001297`. You were right that leaving the newly added phenotypes unbound reopened the item it was meant to close.

**D:** pregnancy management added, `MODULATES` on `Diabetes Mellitus`. Your framing is why it's a treatment and not a footnote — thiamine only modulates that arm, so this is the one place the incompletely-treated arm carries a defined management consequence.

**A process failure worth reporting rather than hiding.** My first attempt at the phenotype split used index arithmetic instead of exact anchors, and because the neurologic block was the *last* phenotype, the end-boundary search fell through to end-of-file and deleted 416 lines — treatments, diagnosis, discussions, animal models, all of it. The snippet count dropping 36 → 22 caught it immediately; I restored with `git checkout` and redid the whole thing with assert-count-one anchors only. Nothing from that attempt reached the branch, and the pushed diff is +95/−8 with all ten sections verified present.

**Not done:** items E–J. F (wiring the extra-triad phenotypes into the pathograph) is the one I think is actually right, and I'd rather do it deliberately than bolt it on — the tissue-expression argument would need its own edge and evidence. G–J need reference fetches. Happy to take F in this PR if you'd prefer.

| Check | Result |
|---|---|
| `just validate` | pass |
| Snippets | 41/41, and 41/41 raw-substring |
| `check-reference-titles` / `check-snippet-length` | no new baseline entries |
| Sections present | all 10 |
| declared vs cited references | agree both directions |

---

---

caufieldjh:
`test (3.13)` went red on `9139b57f9`. **It was not a content defect in this PR, and it is now fixed** by rebasing onto current `main` (`a40e6c879`).

## What failed

`check_environmental_evidence --against-ref origin/main` reported **26 new findings** — and not one was in this entry. They were `Autoimmune_Polyendocrine_Syndrome_Type_1`, `Cerebral_Palsy`, `Down_syndrome`, `Dravet_syndrome`, `Fanconi_Anemia`, `Heart_Failure`, `Phenylketonuria`, `Sickle_Cell_Disease` and others this PR never touches.

## Why

The three most recent commits on `main` are all environmental-evidence work — #9733, #9743, #9745 — merged at 23:42, three minutes before this check ran. They cited or dispositioned those exposures and **shrank the baseline from 37 distinct entries to 11**.

This branch predated them. So the check measured a tree still holding the old evidence-free entries against the *new, smaller* baseline, and every one of them read as newly introduced. The log says as much: `grandfathered ... (11 distinct exposure(s))` against 26 findings.

I did not assume any of that. I checked out `origin/main` clean and ran the identical command:

```
$ uv run python scripts/check_environmental_evidence.py --against-ref origin/main
environmental evidence baseline: grandfathered against ref 'origin/main' (11 distinct exposure(s))
OK: no new evidence-free `environmental:` exposures
[exited with code 0]
```

**Green on main**, so it was not a base-branch failure either — it was specifically staleness in this branch. Same command on the rebased branch now also exits 0.

## What I did

`git rebase origin/main`, clean, no conflicts. Reviewed the result before pushing: `+3475/−34` across 18 files, all additions plus additive cache rows and the intended stub deletion — no unrelated deletions and no protected-path churn. `just validate` passes and snippets hold at 41/41.

## The other four PRs are in the same position, and I am deliberately leaving them

#9755, #9756, #9758 and #9759 all branch from the same older `main`, so a fresh CI run on any of them would fail identically. I am not rebasing them, for two reasons:

1. Three are **approved** (#9755, #9758, #9759). Rebasing dismisses those approvals, and their recorded check rollups are already green.
2. The staleness affects **CI runs on the branch, not the merge result**. None of those branches touches the 26 files in question, so merging takes `main`'s cited versions. The merged tree would be correct.

If any of them does get a fresh CI run and goes red on this check, the same rebase is the fix. Flagging it here so it is a recorded decision rather than something discovered later.

---

---

caufieldjh:
Thanks for the approval — and I have to flag that **I dismissed it about four minutes after it landed**, by force-pushing the rebase. Sequence:

| Time | Event |
|---|---|
| 23:46 | `test (3.13)` fails — `check_environmental_evidence`, 26 findings, none in this entry |
| 23:48 | you approve (content review; the CI failure was not in scope) |
| ~23:52 | I rebase onto current `main` and force-push `a40e6c879` — approval auto-dismissed |

**No content changed.** The rebase carries exactly the four commits you reviewed onto a newer base; the failure was staleness against a baseline that `main` shrank from 37 entries to 11 three minutes before the check ran, and I verified the same command exits 0 on a clean `origin/main`. Full detail in the comment above.

I'd make the same call again — a red check on a PR I own is work now, and a stale-green approval on a failing branch would not have merged anyway. But it is worth saying plainly that the cost was real and that I did not check for a just-landed approval before pushing. A fresh `claude-review` and `test (3.13)` are already queued on the rebased head.

On your two new suggestions, neither of which I'm pushing (the PR is one dismissed approval down already):

**1. `SUPPORT` → `PARTIAL` on the domain-level GeneReviews evidence.** You're right, and the reasoning is exactly the claim-relative principle: the sentence names *neurologic manifestations* as a managed domain, while the claims are `Seizures` and `Stroke` specifically. That's a broader-than-the-claim source, which is what `PARTIAL` is for — and as you note, the entry already grades the erythromycin EHR item that way for the same reason. Inconsistent of me.

**2. `BEHAVIORAL` on the pregnancy treatment.** Also right, and I should have caught it from CLAUDE.md directly: it warns specifically against mechanically tagging `NCIT:C15747`-adjacent supportive-care terms as `BEHAVIORAL`, because they name a chemical or clinical intervention more often than a diet-pattern change. "Good diabetic control" in a TRMA pregnancy is usually insulin-mediated. `OTHER` is the better fit.

Both belong in the follow-up pass with the other deferred items, unless you'd rather have them here — in which case say so and I'll take them together with the re-review rather than in separate pushes.

---

---

caufieldjh:
`test (3.13)` failed again on `a40e6c879`, four minutes after the re-approval. Same check, **new instance** — and it exposes something structural worth flagging beyond this PR.

Rebased again to `68b62642e`; the check now exits 0. That push dismissed the approval a second time, which I'll come back to.

## This is a moving target, not a one-off

| Time | main | baseline | my branch |
|---|---|---|---|
| 23:42 | `#9733`, `#9743`, `#9745` merge | 37 → 11 | red (26 findings) |
| ~23:52 | I rebase onto `7f469bf8a` | 11 | green |
| ~00:05 | **`#9492` merges** (tranche 13) | 11 → **6** | red again (5 findings) |
| ~00:25 | I rebase onto `bb445f42c` | 6 | green |

The 5 new findings were `Metastatic_Melanoma`, `Metastatic_NSCLC`, `Oral_Cavity_Squamous_Cell_Carcinoma` (×2) and `Paranasal_Sinus_Squamous_Cell_Carcinoma` — precisely the exposures `#9492` cited. My diff touches none of them.

## The structural point

`check_environmental_evidence` is ungated and whole-KB by design, and grandfathers against **`origin/main` as of the moment it runs**. That combination means every merged tranche of the #8296 campaign retroactively invalidates every open PR branch behind it, for content those branches do not own and cannot fix except by rebasing.

The design is right for its purpose — `check-duplicate-keys` and `check-entity-refs` are ungated for the same good reason, and a changed-path filter would miss exactly what this catches. But the *baseline-shrinks-under-you* behaviour is a different property from the ungated scope, and while a citation campaign is actively merging it makes red CI the default state for unrelated PRs. Anyone triaging a red `test (3.13)` in the next few days should check the finding filenames against the PR's diff before treating it as a content defect. I nearly did not.

## On the dismissed approval

That is twice now that a necessary CI fix has dropped this PR's approval, and I want to be straight that the second one was more avoidable than the first: I could have waited for CI to report on `a40e6c879` before announcing it fixed. I said the rebase fixed the failure; what was true is that it fixed *that instance* of it.

I'm not going to keep chasing this on a loop. The content is unchanged from what you approved twice — `git diff` between the reviewed tree and the current one is empty apart from the base. If it goes red again on another tranche merge, the right moment to rebase is immediately before merge, not on every recurrence.

I also preemptively rebased sibling #9756 for the same reason.

**Verified on the current head:** environmental check exits 0 against the current baseline (6), `just validate` passes, snippets hold at 41/41.

---

---

caufieldjh:
Item A is correct and it is my error. Fixed in `2b978b321`, along with B, C and D — folded in together rather than spent on another cycle.

**I verified A against the cache rather than accepting it**, since a regrade on the basis of a reviewer's reading is exactly where I should check. The methods say the sixty-three-drug screen was an isotopic thiamine-uptake assay in *"HEK293 cells expressing SLC19A2 wild-type transporter for 10 min"*, and the second half of my quoted sentence is `(fu*Cmax)/IC50` arithmetic over literature pharmacokinetics. Neither half is a human observation. You put your finger on precisely what fooled me: **"clinically" qualifies the concentration threshold, not the study design**, and I read it as the latter.

The decisive argument is the one you gave — internal inconsistency. Line 59 already grades the sibling result from the same methods section of the same paper `IN_VITRO`. One paper, one assay, two classifications.

**Why no gate caught it, which I think is the reusable part.** `check-snippet-grading` keys on the *quoted sentence*, so it catches the same sentence graded two ways — which is what it was built for, and what it caught on sibling #9758 an hour ago. It cannot catch two *different* sentences from one methods section graded inconsistently. That is a real blind spot in exactly the case where a paper mixes evidence types, which is the case CLAUDE.md tells curators to split evidence items for. I have recorded it in the commit and in the entry's `explanation`.

**B, C, D** all taken:

- Variant-class summary → `OTHER`. You are right that the file was applying two rules to one construct; it already grades two `PMID:38037112` Discussion clauses `OTHER` for the same reason.
- The two domain-level GeneReviews items → `PARTIAL`. The entry already applies that reasoning to the erythromycin EHR item, so this was another internal inconsistency rather than a judgment call.
- Pregnancy glycaemic control → `OTHER`. Worth flagging the deliberate contrast with sibling #9756, where the reviewer confirmed `BEHAVIORAL` *is* right for the hepatotoxic-avoidance instruction. Same `NCIT:C15747` action term, opposite modality, and the discriminator is whether the intervention is a chemical or a change in what the patient does. Both are recorded in #9787 so the pair does not later read as an inconsistency.

Thank you also for stating plainly that the push was rebase-only and the blob byte-identical, and for confirming the environmental diagnosis by running the command yourself. E–H remain deferred and are all in #9787.

| Check | Result |
|---|---|
| `just validate` | pass |
| Snippets | 41/41 |
| `check-snippet-grading` | no new divergences |
| `check-environmental-evidence` | clean against the current baseline (6) |

---

---

PR #9758, opened 2026-08-27
curate(COQ6): Add COQ6-related steroid-resistant nephrotic syndrome

### LINKED ISSUE #9748: Curate familial steroid-resistant nephrotic syndrome with sensorineural deafness (MONDO:0013836)
Claiming `MONDO:0013836` — COQ6-related steroid-resistant nephrotic syndrome with sensorineural deafness.

Stub: `stubs/Familial_Steroid-resistant_Nephrotic_Syndrome_With_Sensorineural_Deafness.yaml`

**Why it is worth curating.** *COQ6* is a monooxygenase in coenzyme Q10 biosynthesis, so this is a primary CoQ10 deficiency that presents to a nephrologist rather than to a metabolic clinic. Podocytes and cochlear hair cells are the two cell types that decompensate first.

The reason it earns an entry is that "steroid-resistant" here is a mechanistic prediction, not a clinical observation to record and move on from: the lesion is not immune, so immunosuppression has nothing to act on, and the correct treatment is oral CoQ10 supplementation. Curating the causal chain makes the failure of steroids and the rationale for CoQ10 fall out of the same graph — the sort of case where the pathograph does real work.

Two things to get right:

- **Steroid resistance should be modelled, not merely asserted.** It belongs as a treatment whose mechanism does not engage the causal chain, or as a `mechanistic_hypotheses` / discussion note — not as a bare adjective in the entry name.
- **CoQ10 responsiveness is arm-specific and time-dependent.** Reported benefit is largely renal and largely when started early; the deafness generally does not reverse. As with any partially-treatable disease, the treatment's `target_mechanisms` need to name the arm they act on rather than the disease as a whole.

`kb/groupings/` has no coenzyme-Q10-deficiency grouping yet; worth checking whether the primary CoQ10 deficiencies (COQ2, COQ6, COQ8B/ADCK4, PDSS2) already curated elsewhere in the KB would justify one as a follow-up.

caufieldjh:
Adds `kb/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome.yaml` (MONDO:0013836) — a primary CoQ10 deficiency that presents to a nephrologist rather than to a metabolic clinic.

## &#34;Steroid-resistant&#34; is a prediction, not just a label

The phrase entered this entity&#39;s name from nephrology, where it classifies nephrotic syndrome by observed response. But once the lesion is known to be a defect of coenzyme Q10 biosynthesis in podocytes, resistance to immunosuppression is what the mechanism *predicts*: there is no immune process for a steroid or a calcineurin inhibitor to suppress.

The clean demonstration is a reported case where calcineurin inhibition failed and CoQ10 then produced complete, sustained remission. So the entry models the resistance as part of the causal chain and draws the clinical consequence in an `INTERPRETATION` discussion: in a child with SRNS, resistance to immunosuppression should prompt a genetic and metabolic question rather than an escalation of immunosuppression.

```
Biallelic COQ6 Loss of Function
  → Deficient C5-Ring Hydroxylation in CoQ10 Biosynthesis
      → Coenzyme Q10 Deficiency                    ← CoQ10 replacement BYPASSES here
          ├→ Podocyte Injury and Apoptosis → Focal Segmental Glomerulosclerosis
          │                                   → Steroid-Resistant Nephrotic Syndrome
          ├→ Retinal Ganglion Cell Dysfunction → Optic Atrophy   ← idebenone
          └→ Stria Vascularis Dysfunction → Progressive Sensorineural Hearing Loss
```

## Why these organs

Curated rather than assumed: COQ6 localizes to cell processes and Golgi of glomerular podocytes **and** to stria vascularis cells of the inner ear. The oto-renal phenotype follows the expression pattern of the enzyme, which is what makes this a specific disease rather than a general mitochondrial cytopathy.

The causal link to podocyte death is not merely correlative, and it is not a black box either. `PMID:30425193` names the intermediates — reactive oxygen species, then F-actin cytoskeletal disruption, then caspase-3 activation — so the edge is `INDIRECT_KNOWN_INTERMEDIATES` with those steps stated, not `INDIRECT_UNKNOWN_INTERMEDIATES`.

## The rescue was shown in cells, not in the fish

An earlier version of this description said knockdown "in podocyte cell lines and in zebrafish caused apoptosis that was partially reversed by CoQ10." That sentence is the paper's *abstract* summarizing two experiments at once, and it does not survive contact with the figures: **Figure 5 A–C** show the CoQ10 rescue in cultured mouse podocytes; **Figure 5 D–F** show zebrafish apoptosis against mismatch and uninjected controls, with **no CoQ10 arm at all**.

So the rescue is a cellular result. The animal-model link had claimed it "establishes the same rescue in a whole organism" — the one thing the cited experiment does not show — and now quotes a zebrafish-only sentence, with `limitations` stating that the fish arm supports the phenotype and not the treatment response.

Worth recording how this surfaced: `check-snippet-grading` failed because the same sentence was graded `IN_VITRO` in three places and `MODEL_ORGANISM` in the new one. It looked like a consistency nit. **A quote that cannot carry one `evidence_source` often cannot carry one claim either** — the gate was pointing at a factual error, not a formatting one.

## Models are linked into the pathograph

The zebrafish and podocyte cell-line work is not just cited as evidence — both are curated as `animal_models:` and `experimental_models:` entries with `modeled_mechanisms` pointing at Podocyte Injury and Apoptosis. A `HUMAN_MODEL_MISMATCH` discussion records that no cochlear or optic-nerve model exists in any species, so those two arms rest on a localization result plus analogy to the renal arm.

## Treatment modelled per arm

| Target | Effect | Basis |
|---|---|---|
| Coenzyme Q10 deficiency | `BYPASSES` | supplies the product the blocked pathway cannot make — the step the drug acts on |
| Steroid-resistant nephrotic syndrome | `RESTORES` | complete remission sustained at 12 months, after calcineurin-inhibitor failure |
| Progressive sensorineural hearing loss | `MODULATES` (`PARTIAL`) | 42.9% response rate — stabilization or improvement |
| Optic atrophy (idebenone) | `MODULATES` | reported effective for the optical impairment, and explicitly *not* for the hearing loss |

The idebenone dissociation is kept in both directions — it improved vision and did not improve hearing, in the same patients — because that per-arm split is exactly what has no model to explain it.

## A correction to my own prior

The claim issue for this disease predicted the deafness &#34;generally does not reverse.&#34; **The source says otherwise**: the Korean series reports a 42.9% response rate for no-further-loss-or-improvement, and describes the benefit as genotype-specific. The entry follows the source and grades the link `PARTIAL` rather than asserting either outcome. Worth flagging because the prior was mine, not the report&#39;s.

## What the evidence cannot yet separate

A `KNOWLEDGE_GAP` records that two explanations are offered for variable CoQ10 response and no reported series distinguishes them: the audiological series attributes it to **genotype**, while other reports attribute non-response to **established damage**. These are not mutually exclusive, and separating them would need patients of the same genotype treated at different stages. It decides whether genotype should change the *urgency* of starting treatment or only the expectation of benefit.

## Evidence splitting

`PMID:26808124` is cited twice on the same node with different `evidence_source` values — one quoted sentence reports biochemistry (`IN_VITRO`), the other molecular dynamics and docking (`COMPUTATIONAL`). Per the rule that a paper mixing evidence types has its evidence split rather than graded once.

## GeneReviews

`PMID:28125198` is tagged, and on reading it the chapter is an *Overview* whose cached record is a six-point purpose statement that never mentions COQ6. It is therefore cited once at family level, with that stated in the evidence explanation — rather than claiming per-section mining of a chapter that has no COQ6-specific sections to mine.

## Relationship to existing entries

`kb/disorders/Primary_Coenzyme_Q10_Deficiency.yaml` curates the umbrella and already lists COQ6 in `has_subtypes`. A standalone entry alongside it follows the precedent set by `kb/disorders/COQ4-Related_Neonatal_Encephalomyopathy.yaml`, which likewise exists both in its own right and as a subtype there.

The entry name leads with the gene, following the KB&#39;s convention for gene-defined entities and matching the COQ4 sibling; MONDO&#39;s clinically-worded label is preserved in `disease_term`.

## One schema fix worth noting

`GO:0006744` (ubiquinone biosynthetic process) was initially placed in `molecular_functions`, where it does not belong — term validation caught it. `GO:0004497` and `GO:0071949` now carry the monooxygenase and FAD-binding functions, with `GO:0006744` moved to `biological_processes`.

## Validation

| Check | Result |
|---|---|
| `just validate` | pass (schema + terms + references) |
| Snippets verified | **36/36** |
| Raw-substring re-check bypassing validator normalization | 36/36, 0 non-verbatim |
| `check-snippet-grading` | clean against `origin/main`, no new divergences |
| Reference titles vs cache frontmatter | 44/44 |
| declared vs cited references | agree both directions |
| Pathograph | 1 component, no orphans |

Closes #9748

---

### DIFF
--- a/history/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome/2026-08-27T230358Z-claude-code-600846.yaml
+++ b/history/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome/2026-08-27T230358Z-claude-code-600846.yaml
@@ -1,0 +1,29 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome
+  path: kb/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome.yaml
+session:
+  id: 2026-08-27T230358Z-claude-code-600846
+  timestamp: '2026-08-27T23:03:58Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9758
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - genetic
+  - animal_models
+  - discussions
+  summary: Fix snippet-grading regression; the gate exposed a factual error about
+    the zebrafish rescue
+  details: |
+    Third review-response pass on PR 9758. The previous commit introduced a check-snippet-grading regression: the new animal_models link reused the abstract sentence 'knockdown of Coq6 in podocyte cell lines and coq6 in zebrafish embryos caused apoptosis that was partially reversed by coenzyme Q10 treatment' and graded it MODEL_ORGANISM, while three existing items in the same file grade the identical sentence IN_VITRO. Reproduced the failure before fixing it: 3 new findings. Investigating the right replacement quote turned up a substantive error rather than a formatting one. That sentence is the abstract's summary of two separate experiments, and reading the results section shows the coenzyme Q10 rescue was demonstrated in cultured mouse podocytes (Figure 5, A-C) while the zebrafish panels (D-F) compare knockdown against mismatch and uninjected controls with no CoQ10 condition at all. The animal-model link's explanation had claimed it 'establishes the same rescue in a whole organism, which is what makes it more than a cell-culture artefact' - a claim the cited experiment does not support. The link now quotes a zebrafish-only result, 'coq6 knockdown in zebrafish embryos 28 hours after fertilization increases apoptosis', carries a single evidence type, and its limitations state plainly that the zebrafish arm has no CoQ10 treatment condition so the model supports the phenotype and not the treatment response. The HUMAN_MODEL_MISMATCH rationale, which had said the models reproduce the renal arm 'including its CoQ10 responsiveness', was corrected to attribute the rescue to the cellular level only. Also closed the two items left open from the previous round: the six-manifestation Severe Infantile Presentation bundle was split into five atomic phenotypes bound to HP:0001942, HP:0000093, HP:0003073, HP:0001510 and HP:0001252; death in early infancy was deliberately NOT added as a sixth because HP:0001522 is a clinical-course concept outside the PhenotypeTerm dynamic enum, and that outcome is recorded in the genetic notes instead. The two nonsense alleles were moved from free-text notes into structured genetic variants entries. Validation: schema and terms pass, 36/36 snippets verified and 36/36 on the raw-substring re-check, entity-refs and duplicate-keys pass, 44/44 reference titles match cache frontmatter, declared and cited reference sets agree, check-snippet-grading reports no new divergences.
--- a/kb/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome.yaml
+++ b/kb/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome.yaml
@@ -402,11 +402,19 @@ phenotypes:
     explanation: The report that added the ocular arm to the recognized phenotype.
 
 - category: Metabolic
-  name: Severe Infantile Presentation
+  name: Metabolic Acidosis
   frequency: OCCASIONAL
   description: >-
-    The severe end of the spectrum, associated with truncating variants and death in
-    early infancy.
+    Reported in the lethal infantile presentation, and the feature that marks this end of
+    the spectrum as a systemic mitochondrial decompensation rather than an oto-renal
+    disease. Both reported siblings died in early infancy; that outcome is recorded in
+    the genetic notes rather than as a phenotype, since HP:0001522 is a clinical-course
+    concept outside the PhenotypeTerm enum.
+  phenotype_term:
+    preferred_term: Metabolic acidosis
+    term:
+      id: HP:0001942
+      label: Metabolic acidosis
   evidence:
   - reference: PMID:35111204
     reference_title: "A Family Segregating Lethal Primary Coenzyme Q10 Deficiency Due to Two Novel COQ6 Variants."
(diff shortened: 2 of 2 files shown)


caufieldjh:
All seven blocking items are addressed in `d2cdae4f6`. Snippets went 22 → 30, all verified.

**The one that mattered most.** The entry curated Optic Atrophy as a phenotype with no pathophysiology node and no treatment, while the drug that addresses it sat in a cache this PR itself added. Idebenone is now curated with a node for the retinal ganglion cell arm and a link to it. An entry that adds a source and then does not use what the source says is worse than one that never fetched it.

**The rest:**

- `reference_title` on all evidence items.
- The zebrafish and podocyte cell-line work was cited three times as evidence while the entry carried no model sections at all. Both are now `experimental_models:` / `animal_models:` entries with `modeled_mechanisms` linking to Podocyte Injury and Apoptosis, plus a `HUMAN_MODEL_MISMATCH` discussion for the translational gap the partial CoQ10 rescue leaves open.
- GeneReviews: `PMID:28125198` was tagged but not mined, and on reading it the reason is that it is an *Overview* chapter whose cached record is a six-point purpose statement that never mentions COQ6. So it is cited once at family level with that stated in the explanation, rather than claiming per-section mining of a chapter that has none for this gene.
- `INDIRECT_UNKNOWN_INTERMEDIATES` on the CoQ10-deficiency → podocyte-apoptosis edge was understating what is published. `PMID:30425193` — which was sitting untracked in my working tree — supplies the intermediates directly: ROS → F-actin disruption → caspase-3. Upgraded to `INDIRECT_KNOWN_INTERMEDIATES` with those named. This is the good outcome of an untracked-file warning: the file was not noise, it was the answer to a gap I had already marked.
- Genetic section deepened; the bundled phenotype split.

**Not changed:** `mappings`, `biochemical`, `prevalence`. All non-blocking, none supported by a cached source today, and each would need its own fetching. Follow-up rather than scope creep here.

---

---

caufieldjh:
Grading regression fixed in `3fe74c01b`, along with both remaining open items. Snippets 30 → 36. I reproduced the failure first — 3 findings, exactly as you described — before changing anything.

**The gate caught a factual error, not a formatting one, and that is the part worth reading.**

Your diagnosis was right that the quoted sentence covers both experiments and so cannot be graded once. Working out which zebrafish-only sentence to replace it with is what exposed the real problem. The results section separates the two experiments cleanly:

- **Figure 5 A–C** — cultured mouse podocytes: *"Coq6 downregulation in cultured mouse podocytes induced apoptosis that was diminished by CoQ10 treatment."*
- **Figure 5 D–F** — zebrafish: knockdown compared against mismatch MO and uninjected controls. **No CoQ10 condition at all.**

The rescue was never tested in the fish. My animal-model link claimed it *"establishes the same rescue in a whole organism, which is what makes it more than a cell-culture artefact"* — so the specific point I was making was the unsupported one. The link now quotes a zebrafish-only result, and its `limitations` say plainly that the arm carries no CoQ10 condition and therefore supports the phenotype rather than the treatment response. I also corrected the `HUMAN_MODEL_MISMATCH` rationale, which had said the models reproduce the renal arm *"including its CoQ10 responsiveness"* — true of the podocyte model, not of both.

That the two gradings disagreed *because* the sentence bundles two experiments is the whole mechanism: a quote that cannot carry one evidence type usually cannot carry one claim either.

**Both open items are now closed too**, since I was pushing anyway:

- **Item 6, the bundle.** Split into five atomic phenotypes — `HP:0001942`, `HP:0000093`, `HP:0003073`, `HP:0001510`, `HP:0001252`. Deliberately **not** six: death in early infancy fails validation because `HP:0001522` is a clinical-course concept outside the `PhenotypeTerm` dynamic enum. That outcome stays in the genetic notes, and the metabolic-acidosis description records why.
- **Item 5, structured genetic slots.** The two nonsense alleles moved out of free-text `notes` into `variants` entries. I also tried adding `Genetic.inheritance` and removed it — that slot takes `Inheritance` objects, not a string, and the disease-level `inheritance:` block already covers it, so it would have been duplication.

| Check | Result |
|---|---|
| `just check-snippet-grading` | **clean** against `origin/main`, no new divergences |
| `just validate` | pass |
| Snippets | 36/36, and 36/36 raw-substring |
| Reference titles vs cache | 44/44 |
| declared vs cited references | agree both directions |

---

---

caufieldjh:
Thanks — and thank you for checking the figure legend yourself rather than accepting my account of it. A withdrawn claim is exactly the kind of thing that should not be taken on trust from the person who withdrew it.

**Not pushing these.** All four are non-blocking, and I have already had an approval on a sibling PR auto-dismissed twice by my own follow-up pushes. Recording my position so none of them lapses silently:

**`functional_impact_category: LOSS_OF_FUNCTION`** — you are right, and I want to be clear I did not decline this on the merits. I *tried* to deepen that block and stopped at the wrong boundary: I added `Genetic.inheritance` in the same edit, hit a schema error because it takes `Inheritance` objects rather than a string, removed it, and moved on without revisiting what else belonged there. The snippet already quoted on those variants — *"resulting in two truncated protein products"* — supports the enum directly, and it is what makes the residual-activity argument in `notes` queryable instead of prose. Best candidate for the follow-up.

**`supports: REFUTE` at treatment level** — agreed, and the reasoning generalises: a `REFUTE` on the treatment scans as refuting the treatment, when the claim is about its *reach*. Moving it to a `target_mechanisms` entry against the hearing arm puts the negative where the dissociation lives and makes it visible in the graph rather than only in prose. Same shape as the change that made this entry's idebenone arm work in the first place.

**Cochlear implantation on `NCIT:C15747`** — worth flagging that this is the same finding you raised on #9757, where it is already fixed to `NCIT:C15329` with `preferred_term: cochlear device implantation`. So the sibling entries you cite now include a sixth. Fixing it here would make the convention unanimous.

**`links.issues`** — correct, empty in all four records. It is a `just new-history` invocation flag I have been omitting, so it is systematic rather than a one-off; worth fixing across the round's records in one pass rather than per-PR.

On the figure-legend snippet: I had the same hesitation about quoting a caption, and settled it the same way — it states a result rather than labelling a panel, so it is propositional. Good to have that confirmed independently.

---

---

PR #9768, opened 2026-08-27
chore(environmental): retire the evidence ratchet, leave a hard gate (#8296)

gingin77:
> **Draft, and it must stay one until six other PRs land.** See "Merge order" below. The repo-wide gate in this PR is red by construction today and that is the correct behaviour — it is reporting the 31 exposures those PRs cite or disposition.

The #8296 backlog is being worked to zero. Once it is, every `environmental:` entry in `kb/` is either cited or carries a `review_notes` waiver recording a failed search, and the baseline machinery has nothing left to grandfather. This removes it.

**Net −520 lines.**

## What goes

| | |
|---|---|
| `tests/environmental_evidence_baseline.txt` | 179 rows, deleted |
| `--against-ref`, `--update-baseline`, `ENVIRONMENTAL_EVIDENCE_BASELINE_REF` | and the `load`/`write`/`derive`/`resolve`/`new_findings` helpers behind them |
| baseline tests | 10 of them; predicate, SCAN_DIR-pinning and all waiver tests stay |
| `just update-environmental-evidence-baseline` | removed |

The repo-wide test becomes absolute: any finding is a finding.

## What stays, deliberately

**The ungated standalone CI step.** It is the only thing that checks a kb-only curation PR, which matches no `src`/`tests` path filter — that is why it was made ungated in the first place (#8296), and retiring the *ratchet* is not a reason to drop the *step*. `--all`, `--count` and `--waivers` stay too, so the state remains inspectable rather than only gate-or-no-gate.

## Why this is safe to delete rather than leave dormant

The committed baseline was already **73% stale** — 171 rows, last regenerated 2026-08-15, of which 125 keys no longer matched anything in the tree. Nothing forces a baseline to shrink, so that staleness was invisible, and it left a real hole: on the committed-fallback path, re-adding any of those 125 exposures evidence-free would have passed. Deleting it closes that as a side effect.

The path CI actually used was the ref-derived one, so the file was a fallback for local and shallow-checkout runs only.

`scripts/check_empty_snippets.py` is the model. It shipped with **no** baseline machinery, explicitly because its backlog was zero, and its docstring says so. This file now says the same, plus where to recover the machinery from history if a real backlog ever returns.

**No baseline file has been deleted in this repo before** — `git log --diff-filter=D` over `tests/*baseline*` returns nothing. This is the first, which is worth a moment of reviewer attention even though the arithmetic is straightforward.

## Merge order

This is stacked on `feat/environmental-evidence-waiver-8296` (#9751) rather than on `main`, because both rewrite the same script and building on `main` would guarantee a conflict. Land in this order:

1. #9492, #9726, #9733, #9743, #9745 — these empty the backlog
2. #9751 — the waiver mechanism this depends on
3. this PR

Until step 1 completes, `test_no_evidence_free_environmental_exposures` fails. That is the gate doing its job, and it is the cleanest possible proof that it is a real gate rather than a formality: it goes green exactly when the backlog does.

## Validation

- `uv run pytest tests/test_environmental_evidence.py`: **21 passed, 1 failed** — the failure is the repo-wide gate described above; all 10 waiver tests and all predicate tests pass
- `--count`: 31 findings, 5 dispositioned — matching what the six PRs clear
- Both workflow files parse; `just --list` resolves; no dangling references to `ENVIRONMENTAL_EVIDENCE_BASELINE_REF` or the baseline path outside the docstring that explains its removal
- CLAUDE.md's offline-gates paragraph updated from "The first five use baselines" to "The first four", with a sentence on where the waiver replaced the baseline row

Part of #8296.

### DIFF
--- a/scripts/check_environmental_evidence.py
+++ b/scripts/check_environmental_evidence.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""Guard against NEW `environmental:` exposures with no evidence at all.
+"""Guard against `environmental:` exposures with no evidence at all.
 
 An `Environmental` entry is optional to cite (`evidence` is not required by the
 schema), so a bare exposure -- just a `name` and usually a one-line `notes` --
@@ -15,12 +15,12 @@ is indistinguishable, to the whole validation stack, from a fully-evidenced
 exposure. See dismech issue #8296.
 
 This is deliberately *not* a curation tool: it only counts and gates, it never
-invents a citation. Backfilling the pre-existing backlog is real literature
-work (find a citable source, exact-quote snippet, `just fetch-reference` +
-`count-verified-snippets`) that belongs in prioritized curation tranches, not
-a mechanical sweep -- manufacturing citations for ~200 exposures in one pass is
-exactly the fabrication risk the evidence SOP warns against elsewhere in this
-repo.
+invents a citation. Clearing a finding is real literature work -- find a
+citable source, take an exact-quote snippet, `just fetch-reference` +
+`count-verified-snippets` -- and manufacturing a citation to silence the gate
+is exactly the fabrication risk the evidence SOP warns against elsewhere in
+this repo. Where no such source exists, record the failed search as a waiver
+rather than inventing one.
 
 Signal
 ------
@@ -243,13 +243,14 @@ def find_waivers(data):
 def scan_repo(scan_dir: Path = SCAN_DIR, rel_to: Path = ROOT):
     """Return a sorted list of ``(relpath, location, name)`` findings.
 
-    ``rel_to`` is the base the reported relative paths are computed against. It
-    defaults to :data:`ROOT` for the working tree, but :func:`baseline_from_ref`
-    scans an extracted copy of ``kb/`` under a temp dir and passes that
-    dir so the reported paths still come out as ``kb/disorders/X.yaml`` --
-    matching the working-tree keys the baseline is compared on.
+    ``rel_to`` is the base the reported relative paths are computed against.
+    It defaults to :data:`ROOT`, so findings read as ``kb/disorders/X.yaml``.
+    It is a parameter rather than a constant so a scan can be pointed at a
+    tree that is not this repository: ``test_scan_covers_kb_beyond_disorders``
+    builds a fixture under ``tmp_path`` and needs the reported paths relative
+    to that, not to :data:`ROOT`.
     """
-    return _scan(find_violations, scan_dir, rel_to)
+    return _scan((find_violations,), scan_dir, rel_to)[0]
 
 
 def scan_thin_waivers(scan_dir: Path = SCAN_DIR, rel_to: Path = ROOT):
@@ -271,11 +272,23 @@ def scan_thin_waivers(scan_dir: Path = SCAN_DIR, rel_to: Path = ROOT):
 
 def scan_waivers(scan_dir: Path = SCAN_DIR, rel_to: Path = ROOT):
     """Return a sorted list of ``(relpath, location, name)`` waived entries."""
-    return _scan(find_waivers, scan_dir, rel_to)
+    return _scan((find_waivers,), scan_dir, rel_to)[0]
 
 
-def _scan(finder, scan_dir: Path, rel_to: Path):
-    findings = []
+def scan_all(scan_dir: Path = SCAN_DIR, rel_to: Path = ROOT):
+    """Return ``(findings, waivers)`` from a single walk of *scan_dir*.
+
+    Both lists come from the same ``environmental[]`` entries, differing only
+    in which side of :func:`is_waived` they keep, so calling
+    :func:`scan_repo` and :func:`scan_waivers` separately parses all of
+    ``kb/`` twice -- about 11 seconds each on the current tree, on the ungated
+    CI step and in every ``just qc``.
+    """
+    return tuple(_scan((find_violations, find_waivers), scan_dir, rel_to))
+
+
+def _scan(finders, scan_dir: Path, rel_to: Path):
+    buckets = [[] for _ in finders]
     for path in sorted(scan_dir.rglob("*.yaml")):
         try:
             with path.open(encoding="utf-8") as handle:
@@ -293,9 +306,10 @@ def _scan(finder, scan_dir: Path, rel_to: Path):
         if not isinstance(data, dict):
             continue
         rel = path.relative_to(rel_to).as_posix()
-        for location, name in finder(data):
-            findings.append((rel, location, name))
-    return findings
+        for bucket, finder in zip(buckets, finders):
+            for location, name in finder(data):
+                bucket.append((rel, location, name))
+    return buckets
 
 
 def main(argv=None) -> int:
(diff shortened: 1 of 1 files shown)


gingin77:
Two changes now this is reviewable.

**Base retargeted to `feat/environmental-evidence-waiver-8296` (#9751).** With the base set to `main`, GitHub showed this PR as 15 files, because the branch is stacked on #9751 and `main` does not have those commits yet — so #9751's four disorder entries and their history records appeared here too. Worse, merging this first would have squashed #9751's content in silently, collapsing two separately-reviewed changes into one commit. The diff is now the 7 infra files this PR is actually about. GitHub retargets it to `main` automatically when #9751 merges.

**Out of draft.** It was a draft because the hard gate it introduces was red by construction: `main` still had a backlog, so a check with no baseline had to fail. That is no longer true. Stacked on the rebased #9751, this branch now passes **22/22 including `test_no_evidence_free_environmental_exposures`** — the whole-KB gate that could only go green once the backlog reached zero.

That flip is the argument for the PR: the gate is real, it was failing for a real reason, and it stopped failing exactly when the backlog cleared rather than because anything was relaxed.

Merge order is unchanged: #9751 first, then this.

gingin77:
Fixed, and you were right that it was mine — the `patch()` I used replaced a block that spanned a sentence boundary, so deleting the ratchet explanation left `The pre-existing backlog is` running straight into the new sentence. The `NEW` in the header was worse than untidy: it described exactly the semantics this PR inverts.

Both fixed as you suggested. The double period is fixed too — `WAIVER_SENTINEL` gained its trailing period in #9751, so the guard-path message was appending a second one and printing `Left deliberately uncited..`.

**That prompted me to audit every prose line this diff touches, and there was a third instance of the same damage** that neither of us had caught. In `.github/workflows/main.yaml`, the snippet-grading comment had been left as:

```
# Live grandfathering for the snippet-grading gate (#8184), which was
# the one ratchet
# of the three never wired up here.
```

Badly wrapped, and factually stale on top: "the one ratchet of the three" is no longer true once this PR retires the environmental one. Rewritten and reflowed.

Verified after: `ruff` clean, 22 tests pass, both workflow files parse, and the gate prints

```
OK: no evidence-free `environmental:` exposures (5 dispositioned by a review_notes waiver).
```

No behaviour change in any of it.

One note on the description you flagged as stale in the best way: the "red by construction" caveat was accurate when opened and is now obsolete because the six prerequisites landed. I have left the wording as a record of why it was a draft, since the PR body also documents the merge order, but happy to trim it if you would rather it read as current state.

gingin77:
Both blocking items fixed, and the non-blocking one taken rather than deferred.

**Finding 1 is the same wording I fixed in the justfile and missed here**, and you are right that this instance is worse: `__doc__` is passed to `ArgumentParser(description=...)`, so `--help` opened with "Guard against NEW ... exposures" and listed `--check  (default) fail on any finding` a few lines below. It now reads:

```
Guard against `environmental:` exposures with no evidence at all.
...
  --check     (default) fail on any finding
```

The paragraph below it also framed the backlog in the present tense. Kept the "this never invents a citation" point, which still matters, and dropped the "~200 exposures to work through" framing; it now points at the waiver as the answer when no source exists.

**Finding 2** — `rel_to`'s rationale really was explained entirely in terms of a function this diff deletes, so it was both a broken `:func:` reference and a wrong explanation. As you say, the parameter still earns its place because `test_scan_covers_kb_beyond_disorders` passes a `tmp_path`; the docstring now says that instead.

**On the double scan, I took it rather than deferring.** Your diagnosis was precise — `find_violations` and `find_waivers` iterate the same `_iter_environmental` entries and differ only in which side of `is_waived` they keep, so the second walk was redundant rather than additive. `_scan` now accepts several finders and partitions in one pass, with `scan_all()` returning both lists:

```
before:  ~22s   (scan_repo 11.1s + scan_waivers 11.1s)
after:    7.4s
```

`scan_repo()` and `scan_waivers()` keep their signatures, since the tests use both. Verified after: `ruff` clean, 22 tests pass, `--waivers` and `--count` unchanged, gate still reports `0 findings, 5 dispositioned`.

Thanks also for independently confirming the CLAUDE.md "first four" count and the `check_empty_snippets.py` precedent — those were the two claims in the PR body I most wanted a second pair of eyes on.

gingin77:
Rebased onto the updated base. This PR was `CONFLICTING` because #9751 gained a substantive change after this branch was last rebased: the waiver substance floor moved out of pytest and into the check itself, since the test lane is path-filtered and never runs on the kb-only PRs that add waivers.

Three conflicts, all in the overlap between that change and the ratchet removal, resolved to keep both. The hard-gate failure message now carries the thin-waiver report as well as the no-baseline advice:

```
EXIT=1
2 finding(s).

Some of these carry a waiver sentence with nothing behind it.
A waiver must record the search -- at least 20 words saying which searches
were run and why they failed:
  kb/disorders/Gout.yaml:environmental[4]: 'Dehydration' (0 word(s) after the sentinel)

There is no baseline to grandfather these into: the dismech#8296 backlog was
worked to zero and the ratchet was removed. ...
```

That is the combination worth having: with no baseline, a bare sentinel is not merely un-grandfathered, it is named as the specific defect it is.

Verified after the rebase: `ruff` clean, **25 tests pass** (up from 22 — the enforcement tests came across), gate reports `0 findings, 5 dispositioned`, and reducing the `Gout` waiver to a bare sentinel exits 1 as shown. No baseline residue anywhere; the remaining `baseline_from_ref`/`resolve_baseline` matches in `scripts/` belong to the sibling ratchets, which keep theirs.

Your two approvals were on the pre-rebase tree, so this will need one more look. Sorry for the churn — it is the cost of the stack, and it should be the last of it: #9751 is green and awaiting its own re-review, and nothing else is queued behind this.

gingin77:
Retargeted to `main` and replayed, now that #9751 has merged.

Two mechanical things happened, neither of which changes this PR's content:

**GitHub did not auto-retarget.** That fires when the base branch is deleted, and #9751 was squash-merged without deleting it. Retargeted manually.

**Retargeting made it conflict, because #9751 was squash-merged.** `main` now carries that work as a single commit whose content matches but whose history does not, so this branch's own copies of those commits collided with it. Replayed just the three retirement commits onto `main` with `git rebase --onto`, which applied cleanly with no conflicts.

The diff is now exactly the 7 infra files, against `main`:

```
.github/workflows/main.yaml
.github/workflows/nightly-kb-sweep.yaml
CLAUDE.md
project.justfile
scripts/check_environmental_evidence.py
tests/environmental_evidence_baseline.txt   (deleted)
tests/test_environmental_evidence.py
```

No `kb/` content, and I confirmed the Pemphigus citation that landed via #9751 is intact rather than reverted by the replay.

Verified on the replayed branch: `ruff` clean, 25 tests pass, and the gate reports

```
OK: no evidence-free `environmental:` exposures (4 dispositioned by a review_notes waiver).
```

That last line is the point of the whole stack. `main` now reads **0 evidence-free exposures** — down from the 182 that opened #8296 — so the hard gate this PR introduces passes on real content rather than by construction.

The retarget and force-push dismissed the prior approval, so this needs one more look. Sorry for the extra round; it is the last one.

PR #9779, opened 2026-08-27
curate: add Immunodeficiency 65 / IRF9 deficiency (MONDO:0032848)

### LINKED ISSUE #9774: Curate immunodeficiency 65, susceptibility to viral infections (MONDO:0032848)
Curate a dismech entry for **immunodeficiency 65, susceptibility to viral infections** ([MONDO:0032848](https://monarchinitiative.org/MONDO:0032848)) — an autosomal recessive inborn error of type I/III interferon signalling caused by biallelic loss-of-function variants in `IRF9` (hgnc:6131), presenting with severe or fatal viral disease including live-attenuated vaccine strain disease.

This looks like a single-gene Disease with a conserved pathograph. The nearest curated neighbour is `kb/disorders/STAT2_Deficiency.yaml` — the same ISGF3 complex, so the two entries should be mutually consistent about the downstream ISG-induction node without duplicating each other.

Stub: `stubs/Immunodeficiency_65_Susceptibility_To_Viral_Infections.yaml`

nlharris:
Adds a Disease entry for immunodeficiency 65 — an autosomal recessive inborn error of immunity from biallelic loss-of-function variants in `IRF9`, the DNA-binding subunit of the ISGF3 transcription factor complex.

Closes #9774

> **Updated after two review rounds.** The sections below reflect the current state of the branch. Where review changed something materially, it's marked.

## What the pathophysiology models

Eight nodes, arranged so that one lesion visibly produces two opposite-looking arms:

- **Antiviral arm.** IRF9 loss → ISGF3 trimer cannot assemble, while STAT1 homodimers (GAF) still form normally → interferon-stimulated gene induction is narrowed but not abolished → patient cells fail to restrict influenza A, PIV and RSV → severe viral pneumonitis → the pneumonia phenotype.
- **Inflammatory arm.** The *same* ISGF3 failure → USP18 and other negative regulators are not induced → IFNAR signalling runs abnormally long → the late transcriptional output comes to mimic interferon-gamma → hyperinflammation and HLH.

The shared upstream node is the point of the entry. Without it, "patient cannot mount an interferon response, yet suffers interferon-driven inflammation" reads as a contradiction rather than a consequence.

**Changed in review:** the transcriptional switch and the clinical syndrome were originally one bundled node. They're now separate (`CELLULAR` → `ORGANISM`), which lets the literature's "likely contributes" hedge sit on the *edge* between them rather than contaminating a measurement that isn't hedged.

Two details preserved rather than rounded off:

- The defect is **selective**, not global — GAF forms, so the type II interferon arm is intact.
- ISG induction is **narrowed, not absent**, and the residual subset is the most plausible reason the index patient controlled most common viruses and failed only against influenza A.

## Two corrections worth recording

**A gene ID.** The deep-research report gave the causal gene as `HGNC:6398`. That CURIE is **KPNA5**. IRF9 is `hgnc:6131`. The entry uses the verified identifier and records the correction in `genetic.notes` so the wrong ID isn't re-derived from the committed report.

**An unsourced clinical claim.** The description originally said the disease causes "disseminated disease following live-attenuated viral vaccines." Review asked for evidence or removal; I searched, found no source attributing that to IRF9 deficiency, and **dropped the sentence**. It's the most management-relevant thing one could say about IMD65, which is exactly why it shouldn't sit in the KB unsourced.

## The animal model turned out to matter

Added at review request as a completeness item, and it contradicts this entry's defining claim. In the p48/Irf9-null mouse, p48 is essential for **both** type I and type II interferon responses. In the human disease GAF forms normally and type II signalling is intact — that selectivity is the entry's headline.

So the link is `PARTIALLY_RECAPITULATES` with a `limitations` field stating the model must not be cited for the selectivity, plus a `HUMAN_MODEL_MISMATCH` discussion and a proposed cell-type-matched experiment that would separate a genuine species difference from a comparison artefact.

## What I did not do, deliberately

**No frequency values on any phenotype.** IMD65 has been reported in a small number of kindreds. Per-phenotype frequencies cannot be computed from that denominator. The top-level `notes` states this up front, and there's now a `prevalence:` record (`CASES_IN_LITERATURE` / `NOT_YET_DOCUMENTED`) making the denominator structured rather than prose-only.

**No separate type III interferon node.** Patients with inherited IL10RB deficiency, who cannot respond to type III interferons, have no overt viral disease — so the type III contribution can't be assumed. Recorded as a `KNOWLEDGE_GAP`.

**No duplication of `STAT2_Deficiency`.** The sibling ISGF3 defect shares the prolonged-IFNAR mechanism and the same source. Per the project's non-DRY convention, the shared node is stated in both.

## Resolved: the modelling question this PR originally asked

The first version of this body asked for a second opinion on `GO:0060337` appearing twice with opposite modifiers. **Answered in review, and the answer was better than my workaround:** bind `GO:0060339` (negative regulation of type I interferon-mediated signaling pathway) with `DECREASED` — that states "the brake fails" directly, and `GO:0060337` + `INCREASED` then reads as its consequence. Adopted. The `notes` now explains the pairing instead of defending it.

## Evidence

24/24 snippets verified. One item graded `PARTIAL` because its source says "likely contributes."

On the deep-research report committed alongside: it resolved 8/8 references with `confabulation_rate: 0.0`, but only 4/8 scored on topic, and 8 references is a thin base. Every claim was anchored to a primary paper read from the cache rather than taken from the report narrative — which is how the KPNA5 error surfaced.

## Validation

`just validate`, `just validate-terms`, `just count-verified-snippets` (24/24), `just check-entity-refs`, `just check-duplicate-keys`, `just validate-disorders` — all pass.

Deferred items are tracked in #9788.

### DIFF
--- a/history/disorders/Immunodeficiency_65_IRF9_Deficiency/2026-08-28T004752Z-claude-code-f0f301.yaml
+++ b/history/disorders/Immunodeficiency_65_IRF9_Deficiency/2026-08-28T004752Z-claude-code-f0f301.yaml
@@ -1,0 +1,29 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Immunodeficiency_65_IRF9_Deficiency
+  path: kb/disorders/Immunodeficiency_65_IRF9_Deficiency.yaml
+session:
+  id: 2026-08-28T004752Z-claude-code-f0f301
+  timestamp: '2026-08-28T00:47:52Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9774
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9779
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - discussions
+  summary: 'Review round 2: add the missing phenotype edge, correcting the previous
+    record'
+  details: |
+    Corrects a false statement in the immediately preceding history record for this entry, the one timestamped 2026-08-28T001726Z. That record's details claim downstream edges were added from the pneumonitis and hyperinflammation nodes to their phenotypes. Only the hyperinflammation edge landed. The pneumonitis edge was attempted with a Python string replacement that did not match and silently no-opped, and unlike the other replacements in the same session that one carried no assertion, so the failure was invisible. The PR comment repeated the same false claim. The preceding record is left unedited because history records are append-only and describe a session as it ran; a record that misreports its own outcome is part of the audit trail, and rewriting it would hide the error rather than correct it. This record is the correction. The consequence was not cosmetic: without that edge the antiviral arm, which is the main arm of this disease, terminated at a pathophysiology node and never reached a clinical phenotype, while the inflammatory arm did, so the rendered pathograph asserted an asymmetry opposite to the entry's argument. The edge is now present and verified by grep and by just check-entity-refs rather than assumed. Also in this round: removed a provenance sentence from the USP18 node notes that cited the wrong PR number and would have rotted regardless, and added a proposed_experiments block to the HUMAN_MODEL_MISMATCH discussion, a three-way cell-type-matched comparison of GAF formation in p48-null mouse and IRF9-null human cells that separates the species-difference explanation from the comparison-artefact explanation the rationale already named. Snippets 24 of 24 verified.
--- a/kb/disorders/Immunodeficiency_65_IRF9_Deficiency.yaml
+++ b/kb/disorders/Immunodeficiency_65_IRF9_Deficiency.yaml
@@ -206,6 +206,8 @@ pathophysiology:
       same child controlled respiratory viruses other than influenza A in vivo,
       so the in-vitro failure against PIV and RSV did not translate into
       clinical disease from those agents.
+    downstream:
+      - target: phenotypes#Severe Viral Pneumonia
     evidence:
       - reference: PMID:30143481
         reference_title: "Life-threatening influenza pneumonitis in a child with inherited IRF9 deficiency."
@@ -252,8 +254,6 @@ pathophysiology:
       directly is what makes this coherent with the DECREASED modifier on
       GO:0060337 at the ISGF3 assembly node, where the claim is about
       ISGF3-dependent transcription rather than about signalling duration.
-      Adopted from review feedback on PR 9778's sibling; it is a better fit than
-      defending two opposite modifiers on one term in prose.
     evidence:
       - reference: PMID:35182547
         reference_title: "Aberrant inflammatory responses to type I interferon in STAT2 or IRF9 deficiency."
@@ -472,6 +472,35 @@ discussions:
       and the patient cell types assayed, is not resolved. Until it is, the
       mouse supports the ISG-induction node and cannot be used to argue about
       the selectivity.
+    proposed_experiments:
+      - experiment_id: irf9_gaf_formation_across_species_and_cell_types
+        name: >-
+          GAF formation and IFN-gamma-induced antiviral state in p48-null mouse
+          and IRF9-null human cells, matched by cell type
+        description: >-
+          Assay STAT1 homodimer formation and the IFN-gamma-induced antiviral
+          state side by side in p48-null mouse embryonic fibroblasts, p48-null
+          mouse cells of a type matched to those assayed in patients, and
+          IRF9-deficient human fibroblasts. The three-way comparison separates
+          the candidate explanations the rationale names: if the mouse defect
+          persists across cell types while the human one does not, it is a
+          species difference; if it tracks cell type, the original discrepancy
+          is an artefact of comparing embryonic fibroblasts with patient cells.
+        would_support:
+          - pathophysiology#Failure of ISGF3 Trimer Assembly
+        supporting_outcome:
+          - >-
+            GAF formation and the IFN-gamma-induced antiviral state are intact
+            in IRF9-null human cells but impaired in p48-null mouse cells of the
+            same type, establishing a genuine species difference and confirming
+            that the mouse cannot speak to the selectivity of the human defect.
+        refuting_outcome:
+          - >-
+            The type II interferon defect in p48-null mouse cells disappears
+            when a cell type matching the human assays is used, indicating the
+            divergence is a comparison artefact rather than a species
+            difference, and that the model may after all be informative about
+            selectivity.
     evidence:
       - reference: PMID:9078371
         reference_title: >-


nlharris:
All six items addressed, plus your `GO:0060339` suggestion, which was better than what I had. Detail below; the important one is item 1, where I took the option you offered that the evidence actually supported.

## 1. The live-vaccine claim — dropped, not evidenced

You were right that nothing in the entry backed it. I went looking for a source before deciding, and the search changed the answer: I could find no publication attributing disseminated live-attenuated vaccine disease to IRF9 deficiency. Worse for the claim, the one full-text source available (PMID:34702736, the same index patient) reports that she "has been vaccinated annually against influenza, which has considerably improved her clinical status."

So the sentence is gone. You offered "evidence it or drop it" and the evidence pointed one way. If a source for vaccine-strain dissemination in IRF9 exists and I missed it, I'd like the pointer — the sibling `STAT2_Deficiency` pattern with `HP:0031697` is right there and I'd rather use it than not.

## 2–4. Treatments, diagnosis, genetics — added

- **Treatments.** The casirivimab/imdevimab case (PMID:34702736) is in, and you're right that it's mechanistically the point rather than just a case report — passive neutralisation works *because* it bypasses ISGF3, in a patient for whom interferon therapy was predicted to fail. Its `target_mechanisms` link goes to `Uncontrolled Replication of Respiratory Viruses` for exactly that reason. Immunoglobulin replacement is in, graded `PARTIAL`: the source credits IVIG *and* annual influenza vaccination together, and attributing a combined outcome in one patient to one arm would overstate it. I did **not** add live-vaccine avoidance — same reason as item 1.
- **Diagnosis.** ISGF3-vs-GAF assay added, using the snippet already cached, as you noted.
- **Genetics.** `c.991G>A` added as a `Variant` with `clinical_significance: PATHOGENIC` and its own loss-of-function evidence from the new reference; `functional_impact_category: LOSS_OF_FUNCTION` on the causal pathophysiology node. The second kindred's `c.577+1G>T` is recorded in `notes` rather than as a structured claim — it isn't quotable from any cached abstract, and I'd rather have it as sourced prose than as an unsourced field.

## 5. Node split — done, and your read of the seam was right

Split into `Interferon-Gamma-Like Late Transcriptional Output` (`CELLULAR`) → `Hyperinflammation and Haemophagocytic Lymphohistiocytosis` (`ORGANISM`). Your point about the two evidence items separating along that seam is what makes this more than tidying: the `SUPPORT` measurement now sits on the cellular node, and the `PARTIAL` "likely contributes" hedge sits on the **edge** between the two — which is the thing that's actually uncertain. Bundled, the hedge was contaminating an observation that isn't hedged at all.

I also promoted the "frequent yet paradoxical accompaniment" quote to the organism node as `SUPPORT`, since the syndrome's *occurrence* isn't in doubt — only its attribution to the transcriptional switch.

## 6. Pathograph connectivity — done

`downstream: phenotypes#…` edges added from both terminal nodes. One connected component now, and the duplication you flagged is gone.

## Your GO:0060339 suggestion — adopted

This is the better answer and I've taken it. `GO:0060339` + `DECREASED` says "the brake fails" directly, which is the substantive claim; `GO:0060337` + `INCREASED` stays alongside as the consequence. The pair now reads as one coherent statement instead of needing a `notes` block to defend two opposite modifiers on one term. The `notes` that remains explains the pairing rather than apologising for it.

## Also done from the non-blocking list

- **`prevalence`** — `CASES_IN_LITERATURE` / `NOT_YET_DOCUMENTED`, so the denominator my top-level `notes` reasons from is structured rather than prose.
- **`animal_models`** — the p48/Irf9 knockout (PMID:9078371) added, and it turned out more interesting than a completeness item. In the mouse, p48 is essential for **both** type I and type II interferon responses; in the human disease, GAF forms normally and type II signalling is intact. That's a direct contradiction of this entry's defining selectivity claim, so the link is `PARTIALLY_RECAPITULATES` with a `limitations` field saying the model must not be cited for the selectivity, plus a `HUMAN_MODEL_MISMATCH` discussion. Thank you for pushing on this — I'd have missed it.
- **`parents`** — dropped `Primary Immunodeficiency`.
- **`evidence_source`** — the type III redundancy conclusion regraded `HUMAN_CLINICAL` → `OTHER`; you're right that it's a synthesis, not an observation.

## Not done

`mappings:` for OMIM #618648, a top-level `references:` block, and `cell_types` on the nodes. All fair; I've left them as the smallest-value items in a diff that has roughly doubled. Happy to add them if you'd rather they landed here.

Snippets now 24/24 verified. `just validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys` and the model-link pytest selection all pass.


---

nlharris:
Fixed, and thank you for checking rather than taking the claim at face value — I reported that edge as added and it wasn't there.

## What happened

The edit was a Python string replacement that didn't match and **silently no-opped**. Other replacements in the same session carried an `assert`; that one didn't, so the failure was invisible to me. I then stated it as done in both the PR comment and the history record.

Nothing downstream would have caught it either: `check-entity-refs` verifies that references *resolve*, and an absent edge has no reference to break. An unmade edge is indistinguishable from a deliberate omission to every check in the repo. That's the gap that let a false claim through — worth noting alongside the metadata gaps in #9785, since it's the same shape: the tooling verifies what's present and is silent about what should be.

You were also right that it isn't cosmetic. Without it the antiviral arm — the main arm of this disease — terminated at a pathophysiology node while the inflammatory arm reached a phenotype. The rendered pathograph would have asserted an asymmetry opposite to the entry's argument.

The edge is now present, verified by grep and `check-entity-refs` rather than assumed. I also audited the same class of failure across the other four PRs in this run; their claimed edges all landed.

## The history record

I did **not** edit it. It says something false, but records are append-only and describe a session as it ran — a record that misreports its own outcome is part of the audit trail, and rewriting it would hide the error rather than correct it. A new record carries the correction and names the root cause. If the project convention is to correct `details` in place for factual errors rather than append, tell me and I'll switch.

## Suggestions taken

- **USP18 `notes`** — provenance sentence removed. You were right on both counts: it cited the wrong PR *and* a PR number baked into a KB field rots regardless. The substantive part stands on its own.
- **`proposed_experiments`** — added to the `HUMAN_MODEL_MISMATCH`. A cell-type-matched three-way comparison of GAF formation and the IFN-gamma-induced antiviral state across p48-null mouse embryonic fibroblasts, p48-null mouse cells matched to the human assay, and IRF9-null human fibroblasts. That separates two of the three explanations the rationale named — if the mouse defect persists across cell types it's a species difference; if it tracks cell type, the original discrepancy was a comparison artefact.

## On your influenza-vaccine caveat

Fair correction, and I'd overstated it. The annual influenza vaccination is almost certainly the inactivated product, so it isn't evidence against live-vaccine risk — it just isn't evidence either way. My reply implied it argued against the claim. It doesn't; the claim was simply unsourced, which is sufficient on its own for dropping it.

Snippets 24/24, gates clean.


---

Issue #7622, opened 2026-08-01
Orphadata and ClinGen-dosage pins are stale and un-refreshable; bumping them needs a dedicated PR (326-cache rebuild)


jmcmurry:
Both structured-source refresh recipes **fail checksum verification against the currently pinned snapshots**, so the pinned data can no longer be re-downloaded:

```
just refresh-orphadata        -> checksum mismatch (pin: 2025-12-09)
just clingen-dosage-refresh   -> checksum mismatch (pin: 2026-05-08)
```

Orphanet and ClinGen serve these at stable URLs and overwrite them on each release, so the pinned bytes are simply gone from upstream. Current versions:

| Manifest | Pinned | Upstream now |
|---|---|---|
| `data/orphadata/MANIFEST.yaml` | 2025-12-09 | **2026-06-23** |
| `data/clingen-dosage/MANIFEST.yaml` | 2026-05-08 | **2026-07-31** |

## Why it matters

Any curator who needs a **new** `ORPHA:` or `CGDS:` cache record is blocked. This surfaced on #7599, where two review items asked for `ORPHA:698090` and `CGDS:HGNC_14255` and neither could be generated without bumping the pin.

## Why it needs its own PR

Bumping the pin is a one-line change with a **326-file blast radius**:

| | Count |
|---|---:|
| `references_cache/ORPHA_*.md` in the repo | **326** |
| `references_cache/CGDS_*.md` | 1 |

The MANIFEST header prescribes refresh → verify → **rebuild** → review the cache diff. Bumping the pin while rebuilding only the one or two records a given PR happens to need leaves the repo *claiming* a six-months-newer snapshot while 325 of 326 caches came from the old one. The breakage is then deferred onto whoever next runs a full `just structured-rebuild-orphanet`: any Orphanet row whose wording changed in six months silently stops matching its quoted snippet, surfacing in a PR that has nothing to do with it.

CI already agrees — the **"Check curation PR scope"** job flags these paths and demands a `scope-override` label. On #7599 I reverted the bump rather than add that label, precisely because this is a maintainer call.

## What the fix should look like

1. `just refresh-orphadata` + `just clingen-dosage-refresh`, updating both MANIFESTs.
2. `just structured-rebuild-orphanet` and `just clingen-dosage-rebuild` — **all** records, not a subset.
3. Review the resulting `references_cache/` diff as the primary artifact of the PR. Six months of Orphanet drift will change some rows.
[...]

PR #9780, opened 2026-08-27
curate: add MBD5 Haploinsufficiency Syndrome (MONDO:0007974)

### LINKED ISSUE #9775: Curate intellectual disability, autosomal dominant 1 (MONDO:0007974)
Curate a dismech entry for **intellectual disability, autosomal dominant 1** ([MONDO:0007974](https://monarchinitiative.org/MONDO:0007974)) — MBD5 haploinsufficiency (hgnc:20444), also known as 2q23.1 deletion syndrome, caused by heterozygous deletion of or loss-of-function variants in the methyl-CpG-binding domain protein MBD5.

**Note on existing coverage and on the MONDO label.** `kb/disorders/Autosomal_Dominant_Non-Syndromic_Intellectual_Disability.yaml` names MBD5 only as one example gene inside a chromatin-regulator module; it does not curate this entity and does not carry MONDO:0007974. Worth flagging in the entry itself: the MONDO/OMIM label says *non-syndromic*, but the clinical literature is consistent that MBD5 haploinsufficiency is syndromic — seizures, autism, severe speech impairment and recognisable dysmorphism. The entry should say so rather than inherit a label that misdescribes the disease.

Stub: `stubs/Intellectual_Disability_Autosomal_Dominant_1.yaml`

nlharris:
Adds a Disease entry for MBD5 haploinsufficiency — historically 2q23.1 microdeletion syndrome — a neurodevelopmental disorder from loss of one functional copy of the methyl-CpG-binding domain protein MBD5.

Closes #9775

## The naming decision, which is the thing to review

MONDO:0007974 and its OMIM source call this **"intellectual disability, autosomal dominant 1"** and place it in the non-syndromic intellectual disability phenotypic series. That label misdescribes the disease. Every clinical series reports short stature, microcephaly, seizures, a distinctive disturbed sleep pattern, stereotypic repetitive behaviour and coarse facies alongside the intellectual disability — and the phenotype was initially mistaken for Angelman, Rett and Smith-Magenis syndrome in several patients *precisely because* it has a recognisable syndromic gestalt.

So the entry is named `MBD5 Haploinsufficiency Syndrome`, for the gene and mechanism, while `disease_term` stays bound to MONDO:0007974 so the mapping isn't lost. The discrepancy is written into the entry's top-level `notes`.

This is the general shape of a problem worth naming: the best available ontology term can be *correct as a binding* and *wrong as a label*, and nothing in the schema distinguishes those. Recording it in prose is the only option I had.

## Why this is a single-gene disorder and not a contiguous-gene syndrome

Worth stating because it's unusually well established. A consortium assembled 65 subjects with 2q23.1 microdeletions or translocations, aligned breakpoints to narrow the critical region, and found MBD5 as the only gene it contained. Patients whose alteration disrupts MBD5 alone are clinically *and transcriptionally* indistinguishable from those missing the whole interval. The entry cites both findings, because the second is what rules out a contiguous-gene contribution.

## Pathophysiology

Four nodes: MBD5 haploinsufficiency → disrupted chromatin-mediated transcriptional regulation in neurons → impaired neurite outgrowth and neuronal maturation → neurodevelopmental impairment with disproportionate speech deficit.

Two places where I pitched a node lower than the literature's rhetoric:

- The **chromatin node** is stated as "a chromatin-associated transcriptional regulator whose dose matters", not as a specific methyl-DNA-reading mechanism. MBD5 is named for its methyl-CpG-binding domain, but its actual binding to methylated DNA has been questioned. Asserting the reading mechanism would be inheriting a claim from the gene's name.
- The **neurite outgrowth node** rests entirely on the `Mbd5` gene-trap mouse. It's graded `MODEL_ORGANISM` and the node description says so, rather than presenting a mouse culture result as if it were human cell biology.

## A knowledge gap I recorded rather than smoothed over

MBD5 is dosage-sensitive **in both directions** — reciprocal 2q23.1 duplication causes a related, somewhat milder neurodevelopmental syndrome. A model where haploinsufficiency simply reduces a required activity does not explain why doubling that activity is also harmful. This entry curates the deletion side only; the `KNOWLEDGE_GAP` discussion marks the asymmetry so it isn't left as an unremarked hole.

## Also worth knowing clinically

Some causal alterations are partial deletions of non-coding regions that routine diagnostic arrays don't capture or call pathogenic. A patient can be MBD5-haploinsufficient and test negative on standard screening. That's in the `genetic:` block notes with its source.

## Evidence

21 of 21 snippets verified. Curated from a `claude_code` deep-research report, committed alongside — 16/16 references resolved, `confabulation_rate: 0.0`, but `needs_review: true` because the one quote it checked failed to match, so claims were anchored to primary abstracts read from the cache.

## Note on scope

`kb/disorders/Autosomal_Dominant_Non-Syndromic_Intellectual_Disability.yaml` names MBD5 as one example gene in a chromatin-regulator module. Given that MBD5 haploinsufficiency is syndromic, that's arguably a poor fit for that entry's scope — but this PR doesn't touch it. Flagging rather than fixing, since it's someone else's curation and a separate call.

## Validation

`just validate`, `just validate-terms`, `just count-verified-snippets` (21/21), `just check-entity-refs`, `just check-duplicate-keys` — all pass.

---

### DIFF
--- a/history/disorders/MBD5_Haploinsufficiency_Syndrome/2026-08-28T015617Z-claude-code-40d825.yaml
+++ b/history/disorders/MBD5_Haploinsufficiency_Syndrome/2026-08-28T015617Z-claude-code-40d825.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: MBD5_Haploinsufficiency_Syndrome
+  path: kb/disorders/MBD5_Haploinsufficiency_Syndrome.yaml
+session:
+  id: 2026-08-28T015617Z-claude-code-40d825
+  timestamp: '2026-08-28T01:56:17Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9775
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9780
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - evidence
+  summary: 'Review round 4: move a frequency to the phenotype it belongs to'
+  details: |
+    Three round-4 findings, all from one already-cached paragraph of PMID:27222293, and all three were correct. The substantive one: Hypotonia carried frequency VERY_FREQUENT on the strength of the sentence 'In infancy, hypotonia contributes to feeding difficulties for over 90% of the 2q23.1 population'. The over-90 percent figure belongs to the feeding difficulty, not to the hypotonia that causes it, so the band was supported by nothing. The figure now sits on Feeding Difficulties, which previously carried no frequency at all, and Hypotonia carries none, with a description saying why: the only prose figure in that sentence is not its own, and the review's frequency-table row for hypotonia is in a PDF-extracted column layout that is not quotable as prose. This is the same defect class recorded on issue 9785 - a frequency band is unsourceable, so nothing relates it to the evidence item beside it - reproduced in the entry where that pattern was identified. Second finding: Short Attention Span rested on a snippet stating no frequency, avoided on the grounds of a hyphenated line break. The break falls inside 'includ-ing', before the figure; the clause 'repetitive behaviors, are reported in more than 80% of individuals' is clean and contiguous and is now the snippet. Third: fifth-finger clinodactyly was omitted because its figure sits in a ligature-corrupted clause, while the hypotonia evidence item three entries above already quotes 'dif ficulties' with the ligature intact. That was an inconsistent call rather than a cautious one; clinodactyly is now curated at HP:0004209, FREQUENT, quoting the ligature as it stands. Hyperphagia stays omitted and the reviewer accepted that: its Table 1 row really is PDF-mangled. Note the reviewer named PMID:27514998 as the source to fetch; the record that resolves is PMID:27222293, the same paper under a second PubMed ID. Validation: just validate passes, 64 of 64 snippets verified, entity refs, duplicate keys, snippet length, folded hyphens and title snippets all clean.
--- a/kb/disorders/MBD5_Haploinsufficiency_Syndrome.yaml
+++ b/kb/disorders/MBD5_Haploinsufficiency_Syndrome.yaml
@@ -762,6 +762,7 @@ phenotypes:
           frequency is recorded because the abstract gives none.
   - name: Feeding Difficulties
     category: Gastrointestinal
+    frequency: VERY_FREQUENT
     phenotype_term:
       preferred_term: Feeding difficulties
       term:
@@ -776,7 +777,18 @@ phenotypes:
           Feeding therapy with gastrostomy tube feeding as needed
         explanation: >-
           The management entry implies the phenotype; gastrostomy is not offered
-          for a problem that does not occur. No frequency recorded.
+          for a problem that does not occur.
+      - reference: PMID:27222293
+        reference_title: "Clinical and Molecular Aspects of MBD5-Associated Neurodevelopmental Disorder (MAND)."
+        supports: SUPPORT
+        evidence_source: HUMAN_CLINICAL
+        snippet: >-
+          In infancy, hypotonia contributes to feeding dif ﬁculties for over 90%
+          of the 2q23.1 population
+        explanation: >-
+          Supplies the frequency, which belongs to the feeding difficulty rather
+          than to the hypotonia that causes it. The broken word is the cached
+          PDF extraction, quoted as it stands rather than silently repaired.
   - name: Scoliosis
     category: Skeletal
     description: >-
@@ -899,19 +911,25 @@ phenotypes:
         supports: SUPPORT
         evidence_source: HUMAN_CLINICAL
         snippet: >-
-          Short attention span and autistic-like behaviors
+          repetitive behaviors, are reported in more than 80% of individuals
         explanation: >-
-          Names short attention span as a feature of the deletion syndrome. The
-          ">80%" figure follows in the same sentence but across a hyphenated
-          line break in the cached full text, so only the clean clause is
-          quoted.
+          Carries the frequency figure for the sentence that opens "Short
+          attention span and autistic-like behaviors" - the clause is quoted
+          from after the hyphenated line break in the cached full text, which
+          falls inside "includ-ing" and not inside the figure.
   - name: Hypotonia
     category: Neurological
-    frequency: VERY_FREQUENT
     description: >-
       Infantile hypotonia is among the consistent features of the deletion
       syndrome and is the proximate reason for the feeding difficulties curated
-      above - which is why the two carry the same frequency band.
+      above.
+
+      No frequency is recorded, deliberately. The only prose figure in the
+      sentence below - over 90% - belongs to the feeding difficulty, not to the
+      hypotonia that contributes to it, and the deletion review's frequency
+      table lists hypotonia under its consistent band only in a row whose
+      PDF-extracted column layout is not quotable as prose. A band taken from
+      the neighbouring figure would be a number this source does not state.
     phenotype_term:
       preferred_term: Hypotonia
       term:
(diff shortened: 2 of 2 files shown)


nlharris:
All five blocking items and the seven important ones addressed. Two of these were corrections rather than additions, and both were fair hits.

## The GeneReviews chapter (finding 1)

`PMID:27786435` is now fetched, tagged, and cited across all four of its sections. You're right that it was cited nine times by this PR's own research report and then dropped during curation — that's not an oversight I can attribute to the source being hard to find.

It also settles something: the chapter is titled **"MBD5 Haploinsufficiency"**, not by the OMIM phenotype name. That's independent support for the naming decision the PR argued for, and I've added it to the `notes`.

## Finding 6 — the GO annotation contradicted the entry's own text

Removed. This one stings in the right way: the PR description explicitly identified "inheriting a claim from the gene name" as the thing to avoid, and then the entry did exactly that — `GO:0008327 methyl-CpG binding / DECREASED`, with no evidence item, on a node whose successor's description disclaims the methyl-DNA-reading mechanism. The reasoning is now in node `notes` so it reads as a deliberate absence.

## Finding 10 — the EPC2 point is the one I'd flag back

You caught that `PMID:19809484`, which this entry cites four times, reports a **broader phenotype** in patients whose deletion also removes EPC2 — while my description asserted MBD5-only and whole-interval deletions are "clinically and transcriptionally indistinguishable."

Two errors in one sentence: I presented only the half of my own cited literature that made the cleaner story, *and* I dropped Talkowski's hedge ("largely indistinguishable" → "indistinguishable"). Both fixed; the EPC2 observation is now cited and graded `PARTIAL` because it qualifies the single-locus claim rather than supporting it.

## Finding 5 — the circadian mechanism

Added `PMID:25271084` as a `Dysregulated Circadian Gene Expression` node branching off the chromatin node and pointing at the sleep phenotype. You were right that this is what makes the Smith-Magenis resemblance mechanistic rather than anecdotal — conserved putative E boxes in both MBD5 and RAI1, overlapping circadian and mTOR perturbation. The entry narrated that resemblance three times as a diagnostic-confusion story while the mechanism connecting them sat uncurated.

## Findings 2, 3, 4, 7, 9, 11 — done

- **Treatments** and **diagnosis** sections added from GeneReviews MANAGEMENT and DIAGNOSIS/TESTING. The diagnosis section specifically answers the warning in `genetic.notes` that standard arrays can miss this — a warning without a remedy, as you put it.
- **Phenotypes**: self-injury/aggression, global developmental delay, absent/minimal speech, constipation, feeding difficulties.
- **Frequencies**: seizures, sleep disturbance and autistic behaviour all → `VERY_FREQUENT`. Your diagnosis of *why* they were under-called is the useful part: the Talkowski sentence establishes **presence**, not frequency, and I'd used it for both.
- **Pathograph**: terminal node now points at five phenotypes instead of absorbing them into a sentence.
- **Prevalence**: a class record. I did not fetch ORPHA:228402 — recorded `BELOW_1_IN_1000000` with a note saying it should be replaced by a cited figure, rather than quoting a number I hadn't verified.

## Finding 8 — animal model added, with the limitation that matters

`Mbd5+/GT` is now a structured `AnimalModel` with `modeled_mechanisms`. The `limitations` field records something worth having explicit: **the defining human feature cannot be modelled at all.** Disproportionate speech impairment — most affected children never get past single words — has no murine analogue. So the model is informative about mechanism and silent on the phenotype that most distinguishes this disorder clinically.

## Not done

Finding 12 (human iPSC-NPC data, `PMID:34050248`) and the ClinGen dosage / mosaicism / modifier-gene detail in finding 10. Both need references I haven't fetched and verified, and this diff has roughly doubled. Happy to do them if you'd rather they landed here than in a follow-up.

Snippets 45/45 verified. Also rebased — a CI failure on this PR turned out not to be its own: `main` had moved and the environmental-evidence check reported 31 findings in files this diff never touches.


---

nlharris:
All three blocking items and both important ones fixed in `4c4a8e38b`, plus the three carried suggestions. Two things in here are corrections rather than additions.

## Finding 1 — the test failure, and why the obvious check missed it

`influences_mechanisms[].target` is now the bare `Seizures`. `test_environmental_mechanism_targets` passes across 2358 cases.

Your diagnosis of *why* it got through is the part worth keeping: `just check-entity-refs` resolves `phenotypes#Seizures` perfectly well, so the check a curator reaches for after editing a target is precisely the one that cannot see this. Two slots in the same file take target strings under different grammars, and the tooling only validates one of them.

## Finding 2 — arithmetic

`BAND_1_5_PER_10000`. You're right that the rate and the snippet agreed with each other and only the class was wrong, which is the worse version of that bug: a reader checking the number against the quote finds them consistent and never looks at the class.

## Finding 3 — and a frequency the research report got wrong

You pointed at `PMID:27514998`. I'd already fetched `PMID:27222293` — **same paper**: identical title, authors, journal and year, two PubMed records. The full text came back at 48k characters and it is deletion-side throughout.

The substantive result is a correction, not four additions:

> Other behaviors mentioned in ~ 9% of reported patients are anxiety, **hyperactivity**, inappropriate happy demeanor, and social withdrawal

The research report gave hyperactivity as **>60%**. The deletion-cohort review puts it at **~9%**, and puts *short attention span* above 80% — Table 1 lists them in different frequency columns. The report had merged two phenotypes the source deliberately separates, and the ">60%" figure belonged to the other one. Hyperactivity now carries `OCCASIONAL` with the 9% quote; `Short Attention Span` is a separate entry at `VERY_FREQUENT`.

That is worth flagging beyond this entry: the phenotype had already landed in the file from your round-2 list, carrying no frequency. Had I filled the frequency in from the report, it would have been wrong by a factor of seven and every downstream check would have passed.

Added: **Hypotonia** (`VERY_FREQUENT`, ">90%" with feeding difficulty) and **Small Hands and Feet** (~75%). The latter carries a second evidence item quoting the passage listing deletion-only features that "were not noted in 2q23.1 duplication cases" — explicit attribution rather than assumed.

**Not added, with reasons in the file**: hyperphagia appears only in Table 1, whose PDF-extracted row layout renders as `Motor delaya,b Hyperphagia Hirsutism` — not quotable as prose. The clinodactyly figure sits inside a ligature-corrupted clause. Both are real absences of a quotable sentence, not unfetched references.

## Finding 4 — carbamazepine

Added, graded `PARTIAL`. The source says plainly that no drug was clearly superior, so it qualifies the "routine management" claim rather than supporting it — but it still names a preferred agent and one that made seizures worse. You were right that the seizure picture was half-built: the fever trigger was in and the drug that worsens seizures was not, from the same paper.

## Finding 5 — genetic

`variant_origin: DE_NOVO`, and the mosaicism claim now carries your 14%-of-families figure, which turns a caveat into a recurrence-risk number.

**The variant-class breakdown is still not curated, and the reason has changed.** It was "for want of a verified source". It is now: the research report attributes the ~5% sequence-variant figure to this review, and the review's full text — fetched, cached, in this PR — does not state the breakdown. So the report's citation for that number does not support it. That's a sourcing gap in the report, not an unfetched reference, and the entry's note now says so.

## Suggestions taken

Changelog phrase at line 127 rewritten; microcephaly `FREQUENT` from the nine-of-23 count; `GO:0006355` → `GO:0006357`.

One mechanical note on your microcephaly snippet: `"Nine patients had microcephaly."` is four words and `check-snippet-length` rejects it. I quoted it with the preceding sentence and said why in the `explanation`.

## On ORPHA:228402 (your suggestion 4)

Your diagnosis was right and the remedy doesn't work. `just refresh-orphadata` fails a checksum gate — the pinned sha256 no longer matches what Orphanet serves. `just clingen-dosage-refresh` fails identically, which is why the ClinGen dosage assertion is also absent. Both are already tracked: monarch-initiative/dismech#7622 covers them together, and there are **eight** open issues describing the Orphanet half. I did not file a ninth.

So for this entry both are tooling-blocked rather than research-blocked, and the notes now distinguish those two kinds of absence explicitly.

## Validation on `4c4a8e38b`

```
just validate                       → schema + terms clean
just count-verified-snippets        → 62/62 verified
pytest -k environmental_mechanism_targets → 2358 passed
just check-entity-refs / -duplicate-keys  → OK
check-environmental-evidence / -snippet-grading
  / -title-snippets / -snippet-length / -folded-hyphens → no new entries
```

---

nlharris:
All three fixed in `cadf5fc4b`. You were right on all three, and finding 3 is the one I want to be explicit about because it is worse than you framed it.

## Finding 3 — the frequency belonged to a different phenotype

You noted that Hypotonia's description claimed it and Feeding Difficulties "carry the same frequency band" while Feeding Difficulties carried none. The deeper problem is the band itself. The sentence is:

> In infancy, hypotonia contributes to feeding dif ﬁculties for over 90% of the 2q23.1 population

That ">90%" is the frequency of the **feeding difficulty**, not of the hypotonia that contributes to it. So `Hypotonia: VERY_FREQUENT` was supported by nothing — I had read a number from the same sentence and attached it to the wrong noun.

The fix:

- **Feeding Difficulties** now carries `VERY_FREQUENT` with that quote. It is the phenotype the figure is about, and it previously had no frequency at all.
- **Hypotonia** carries no frequency, with a description saying why: the only prose figure in that sentence is not its own, and the review's frequency-table row listing hypotonia in the consistent band is in a PDF-extracted column layout (`Hypotonia/abnormal muscle tonea Aggression/temper tantrums Inappropriate happy demeanor`) that is not quotable as prose.

Worth saying plainly: this is the exact defect class I recorded on monarch-initiative/dismech#9785 two rounds ago — a `frequency:` band has no way to declare what supports it, so nothing relates it to the evidence item sitting beside it — and I then reproduced it in this entry while writing that up. The band verified nothing, the snippet beside it verified perfectly, and every gate passed.

## Findings 1 and 2 — inconsistency, not caution

Both of my stated reasons for omitting something were wrong on inspection, and you can check both against the same paragraph:

**Short attention span.** I said the >80% figure was unreachable across a hyphenated line break. The break is `includ-ing`, which sits *before* the figure. `repetitive behaviors, are reported in more than 80% of individuals` is clean and contiguous, and is now the snippet.

**Clinodactyly.** I said the figure sat inside a ligature-corrupted clause — while three entries above, my hypotonia snippet quotes `dif ﬁculties` with the ligature carried through intact. Same paragraph, same corruption, opposite decisions. Now curated at `HP:0004209`, `FREQUENT`, quoting `small hands and feet (~75.0%), ﬁfth ﬁnger clinodactyly (~70%)` with the ligatures as they stand.

I had written both omissions up as principled. They were not principled; they were unexamined.

**Hyperphagia** stays out, and thank you for checking it rather than taking my word — that Table 1 row genuinely is column-mangled.

## One correction back

You named `PMID:27514998` as the fetch. That record exists but returns abstract-only; `PMID:27222293` is the same paper — identical title, authors, journal, year — under a second PubMed ID, and that one returns the 48k-character full text. Both are Mullegama & Elsea 2016, EJHG, PMC4989212. Nothing turns on it, but a future curator following that instruction would hit the abstract and conclude the frequencies were not quotable.

## Validation on `cadf5fc4b`

```
just validate                → schema + terms clean
just count-verified-snippets → 64/64 verified
check-snippet-length / -folded-hyphens / -title-snippets / -snippet-grading → no new entries
```

CI `test (3.13)` was already green on the previous head, so the round-3 test failure is confirmed fixed in CI and not only in my local replication.

---

PR #9781, opened 2026-08-27
curate: add Primary Bile Acid Malabsorption (MONDO:0013214)

### LINKED ISSUE #9776: Curate bile acid malabsorption, primary, 1 (MONDO:0013214)
Curate a dismech entry for **bile acid malabsorption, primary, 1** ([MONDO:0013214](https://monarchinitiative.org/MONDO:0013214)) — congenital bile acid diarrhoea from biallelic loss-of-function variants in `SLC10A2` (hgnc:10906), encoding the ileal apical sodium-dependent bile acid transporter (ASBT/IBAT).

Nothing in `kb/` currently covers bile acid malabsorption under any name. Expected pathograph: ASBT loss → failure of active ileal bile acid reclamation → bile acid spillover into the colon → colonic secretory diarrhoea and steatorrhoea, with interruption of the enterohepatic circulation driving compensatory hepatic CYP7A1-mediated bile acid synthesis, cholesterol depletion and reduced LDL.

The evidence base here is small — a handful of families plus the `Slc10a2`-null mouse — so this is a case where the curated entry should be explicit about how thin the human evidence is, not just cite it.

Stub: `stubs/Bile_Acid_Malabsorption_Primary_1.yaml`

nlharris:
Adds a Disease entry for primary bile acid malabsorption — congenital bile acid diarrhoea from biallelic loss of `SLC10A2`, the ileal apical sodium-dependent bile acid transporter (ASBT/IBAT). Nothing in `kb/` previously covered bile acid malabsorption under any name.

Closes #9776

## The substantive finding: a human/model mismatch, curated structurally

This is the part worth your attention. The `Slc10a2`-null mouse reproduces the **transport lesion** almost perfectly and fails at exactly the two nodes where the **human disease is clinically defined**. That's curated as four `modeled_mechanisms` links rather than buried in a `description`:

| Node | Relationship | Fidelity |
|---|---|---|
| Failure of ileal bile acid reclamation | `RECAPITULATES` | HIGH (10–20× faecal bile acid excretion, not further increased by resin) |
| Contracted pool + compensatory synthesis | `RECAPITULATES` | HIGH (80% pool contraction despite increased synthesis) |
| Steatorrhoea and fat malabsorption | `FAILS_TO_RECAPITULATE` | LOW |
| Reduced plasma cholesterol | `FAILS_TO_RECAPITULATE` | LOW |

On a low-fat diet the null mouse has **no steatorrhoea** — faecal neutral sterol excretion rises only 3-fold and cholesterol absorption falls only 20%, because the smaller cholic-acid-enriched pool still supports lipid absorption. And its plasma HDL cholesterol is slightly *elevated*, which the original authors themselves flag as unexpected, where affected humans have *reduced* plasma cholesterol.

A `HUMAN_MODEL_MISMATCH` discussion attaches to both failing nodes, with a proposed high-fat dietary challenge and explicit supporting/refuting outcomes. The practical consequence is stated rather than left to inference: **this model supports the transport and pool nodes and must not be cited for the malabsorptive nodes**, and its silence there is not evidence that human steatorrhoea is doubtful.

The two `FAILS_TO_RECAPITULATE` evidence items are graded `REFUTE` — refuting *the model's fidelity for that node*, not the human phenotype. I've said so in each `explanation`, since `REFUTE` on an animal-model link could easily be misread the other way.

## Two things I flagged rather than glossed

**The human evidence base is one family.** A 1997 report of compound heterozygous SLC10A2 variants, with transfection assays showing abolished taurocholate transport. No cohort, no natural-history series. Top-level `notes` says so up front, and **no phenotype carries a frequency value** as a result. Readers should weight the chain accordingly: proximal nodes well supported, distal clinical nodes resting on a single kindred.

**A171S is a red herring worth recording.** It sits on the same allele as the pathogenic splice change in the index family, but it was present in 28% of controls and had no effect on taurocholate uptake. A variant-level reading of that genotype that didn't separate the two would mis-assign causality. That's in the `genetic:` block notes.

## Modelling notes

- The molecular node is stated as a **transport-function** defect, not protein loss: the mutant transporter is expressed and trafficked to the plasma membrane normally and simply transports nothing. That distinction is what the 1997 transfection work establishes, and it's worth preserving.
- The rat pelvic-irradiation study is included as **convergent** evidence — same transporter lost by a different cause, same bile acid redistribution and diarrhoea — and labelled as such in the `explanation` so it isn't mistaken for evidence about the genetic disease.
- Top-level `notes` warns that "bile acid malabsorption" is used far more often for the common **acquired** condition (post-ileal-resection, Crohn, post-cholecystectomy). This entry is the congenital SLC10A2 form only.

## Evidence

23 of 23 snippets verified. Curated from a `claude_code` deep-research report, committed alongside.

## Validation

`just validate`, `just validate-terms`, `just count-verified-snippets` (23/23), `just check-entity-refs`, `just check-duplicate-keys`, plus `pytest -k "model_readout or failure_to_recapitulate"` (5041 passed) — all pass.

---

### DIFF
--- a/kb/disorders/Primary_Bile_Acid_Malabsorption.yaml
+++ b/kb/disorders/Primary_Bile_Acid_Malabsorption.yaml
@@ -66,6 +66,18 @@ notes: >-
   Crohn disease, post-cholecystectomy, or idiopathic - which is a different
   entity with a different mechanism and is not curated here. This entry is the
   primary, SLC10A2-determined congenital form only.
+external_assertions:
+  - name: OMIM primary bile acid malabsorption record
+    source: OMIM
+    assertion_type: disease_record
+    external_id: OMIM:613291
+    url: https://omim.org/entry/613291
+    description: >-
+      The OMIM record MONDO:0013214 xrefs, recorded structurally so the entry's
+      identity is auditable against the source identifier. This is not
+      `mappings.omim_mappings`: the `DiseaseMappings` class carries only
+      ICD-10-CM, ICD-11, MONDO and NCIT slots, so there is nowhere in `mappings`
+      for an OMIM identity to go.
 prevalence:
   - population: Worldwide
     measure_type: CASES_IN_LITERATURE
@@ -235,7 +247,8 @@ pathophysiology:
           id: UBERON:0001155
           label: colon
     downstream:
-      - target: Steatorrhoea and Fat Malabsorption
+      - target: phenotypes#Congenital Watery Diarrhoea
+      - target: phenotypes#Growth Failure
     evidence:
       - reference: PMID:9109432
         reference_title: >-
@@ -272,12 +285,12 @@ pathophysiology:
       circulating pool contracts. In the mouse this is quantified: faecal
       excretion rises 10- to 20-fold and the pool falls by 80% even though
       synthesis has increased, and the residue is selectively enriched in cholic
-      acid. Split from the compensatory-synthesis node it used to share, because
-      pool contraction is an organism-level state while the synthetic response
-      is a hepatocyte process, and the regulatory step connecting them belongs
-      between them rather than inside either.
+      acid. Pool contraction is an organism-level state while the compensatory
+      synthetic response is a hepatocyte process, so they are separate nodes with
+      the regulatory step that connects them modelled between them.
     downstream:
       - target: Loss of FXR-FGF19 Signalling from the Ileum
+      - target: Steatorrhoea and Fat Malabsorption
     evidence:
       - reference: PMID:12819193
         reference_title: >-
@@ -409,6 +422,7 @@ pathophysiology:
   - name: Steatorrhoea and Fat Malabsorption
     biological_scale: ORGANISM
     downstream:
+      - target: phenotypes#Steatorrhoea
       - target: Fat-Soluble Vitamin Malabsorption
     description: >-
       A contracted bile acid pool impairs micellar solubilisation of dietary
@@ -440,6 +454,8 @@ pathophysiology:
       Lifelong vitamin supplementation, not sequestrant therapy, was what she
       actually needed.
     downstream:
+      - target: phenotypes#Hypocalcemia
+      - target: phenotypes#Growth Failure
       - target: phenotypes#Fat-Soluble Vitamin Deficiency
       - target: phenotypes#Rickets
     evidence:
@@ -475,6 +491,8 @@ pathophysiology:
       of cholesterol into bile acids, lowering plasma cholesterol. This is the
       basis on which partial ASBT inhibition was pursued as a lipid-lowering
       strategy.
+    downstream:
+      - target: phenotypes#Hypocholesterolaemia
     evidence:
       - reference: PMID:9109432
         reference_title: >-


nlharris:
All seven blocking items addressed. Fetching the reference you pointed at in item 3 changed more than the phenotype list.

## The 2025 case report breaks the entry's framing (items 3 and 4)

You called `PMID:40814585` "not a footnote" and you were right. That genotyped patient had lifelong severe fat-soluble vitamin deficiency, rickets, a hypocalcaemic seizure at ten months, dental enamel damage and a final height of 155.5 cm — with **normal stool frequency and consistency and no steatorrhoea at all**.

This entry modelled diarrhoea and steatorrhoea as the defining features and stopped one step short of vitamin malabsorption. It would not have recognised the presentation that took eighteen years to diagnose. Added a `Fat-Soluble Vitamin Malabsorption` node downstream of steatorrhoea, four phenotypes, and growth failure sourced from *both* genotyped patients — who have opposite presentations, which is itself the point.

On item 4 specifically: you were right that further kindreds exist, but the more important problem was the one underneath it. The **stated reason** for carrying no frequencies was wrong. It isn't "we have one family" — it's "there is no denominator." Reworded, because a wrong justification for a right decision is the kind of thing that gets copied.

## Item 5 — the secretagogue claim

Softened to what the two citations establish. The cAMP/TGR5/aquaporin mechanism is now *named* in the description as the standard account with an explicit note that neither citation supports it, so a curator with a source can evidence it. I'd rather leave a visible hole than annotate `biological_processes` I can't cite.

## Item 1 — treatments, with a twist

Added, but not in the shape you sketched, and the reason is in the same 2025 paper: *"Treatment with a bile acid sequestrant was attempted but discontinued due to side effects and no significant effect on intestinal vitamin uptake."*

So the sequestrant is graded **`REFUTE`** — for the vitamin arm specifically, with the explanation saying it's not evidence against sequestrants for bile acid diarrhoea generally, which is a different claim and a different population. That failure is mechanistically expected once the entry has both arms: a sequestrant binds *luminal* bile acids, so it acts on colonic spillover and does nothing for the contracted pool.

**Vitamin supplementation is modelled first**, because for that presentation it isn't adjunctive — it's the treatment, and the patient's levels dropped whenever she stopped.

I did **not** add colesevelam or cholestyramine trial data. The placebo-controlled evidence is from acquired BAD, and given that the one genotyped patient in whom a sequestrant was tried didn't benefit, importing that trial evidence would imply a confidence the SLC10A2 literature doesn't support. Low-fat diet is in with **no evidence item** and a note saying so, rather than a borrowed citation.

## Items 2, 6, 7 and the suggestions

- **Diagnosis** added: bile acid retention testing (1.6% at 7 days in the 2025 case) and sequencing, with both papers' own "test early" recommendations.
- **PBAM2/SLC51B** now named as the deliberately excluded Mendelian neighbour — you were right that the `notes` disambiguated the acquired condition and left the closer confusion unmentioned.
- `CL:1000334`, prevalence class record, history `links` — done.
- On `supports: REFUTE` for the `FAILS_TO_RECAPITULATE` links: leaving as-is, per your read. The `explanation` on each states exactly what's being refuted.

**Items 6 and 7 (node bundling, FXR–FGF19–CYP7A1) not done.** I agree with both — the "and" in those node names is the tell, and asserting synthesis "is de-repressed" without modelling FXR/FGF19/CYP7A1 does leave compensation as a brute fact. But splitting those nodes means re-routing edges through a regulatory axis I'd need to fetch and verify sources for, and this diff has already roughly doubled. I'd rather do it properly in a follow-up than half-do it here. Tell me if you'd rather it landed in this PR.

## One correction back

The review suggested `HP:0002153` for hypocalcaemia. That CURIE is **Hyperkalemia**. The entry uses `HP:0002901`. Flagging because it's the same failure mode as the deep-research report's `UBERON:0002018` and `HGNC:6398` — a plausible-looking CURIE attached to the right concept name.

Snippets 35/35 verified. Also rebased — the CI failure here wasn't this PR's.


---

nlharris:
All five items addressed (`91f96089d`). Item 1 is the one that matters, and it's mine.

## Item 1 — the orphan node, and the third instance of the same bug

You were right, and the diagnosis is exact: the commit message said the node was added downstream of steatorrhoea and it wasn't. A Python string replacement didn't match, silently no-opped, and I wrote it up as done without re-reading the file.

**This is the third occurrence in this five-PR run** — after the missing phenotype edge on #9779 and the half-addressed citation item on #9782. Same mechanism every time: unasserted `str.replace`, claim made from intent rather than from the file.

So I stopped fixing instances and ran a reachability audit across all five entries. The other four are clean — one root node each, no dangling targets. And the audit immediately paid for itself: **splitting the pool node broke a second edge**, because the incoming `- target: Contracted Bile Acid Pool and Compensatory Hepatic Synthesis` still pointed at the pre-split name. That one was a *dangling* reference, which I'd have expected `check-entity-refs` to catch and it didn't.

Filed as **#9789**, including that second and sharper sub-bug. Twelve lines of `set` difference detects all three.

## Items 2 and 3 — done, and they were really one item

You asked for the bundles split and separately for the FXR axis. Splitting `Contracted Bile Acid Pool *and* Compensatory Hepatic Synthesis` is what *made room* for the axis — it's exactly what belongs between the halves, which is why the bundle hid it. Now three nodes:

`Contracted Bile Acid Pool` (ORGANISM) → `Loss of FXR-FGF19 Signalling from the Ileum` (CELLULAR) → `CYP7A1 Disinhibition and Compensatory Hepatic Synthesis` (hepatocyte)

Your scale observation was the tell and I should have taken it in round 1: an `ORGANISM` node carrying a hepatocyte process descriptor is a bundle showing through.

Fetched PMID:35683489 and PMID:38338820 via the PMC→PMID converter, as you suggested. And you were right that the same fetch unblocks item 5 — the axis *explains* both biomarkers, so serum FGF19 and serum C4 are now a `biochemical:` section curated as readouts of the two new nodes rather than free-floating tests. One C4 item is graded `PARTIAL` because the same review says their clinical validation is still ongoing.

On round 1: I did give a rationale for deferring these (diff size, in my last comment), so it wasn't quite silence — but you asked twice, and "the diff is large" is a weaker reason than "the axis is the disease's central regulatory circuit." Doing it.

## Item 4 — one-sided evidence, from a file I added myself

This is the one I'm least comfortable with. `PMID:34192422` reports colestyramine resolving the diarrhoea with catch-up growth in a homozygous SLC10A2 infant. I added that reference, cite it twice elsewhere in the entry, and presented only the `REFUTE` half of the sequestrant picture.

Both are now there, and having both makes the mechanism legible in a way neither did alone: a sequestrant binds *luminal* bile acids, so it acts on colonic spillover and does nothing for the contracted pool — which is why it helps the diarrhoeal presentation and fails the vitamin one. The description no longer says there's no SLC10A2-specific evidence, only no controlled-trial evidence. `therapeutic_agent: NCIT:C28929` added.

## Item 5 and suggestions

75SeHCAT now named in the diagnosis entry (you're right that the scan I was already citing *is* one, so it cost nothing), plus the gold-standard quote and its access caveat. Vitamin supplementation regraded `SMALL_MOLECULE`.

Not done: OMIM mapping, dental enamel defects, the hypocalcaemia edge, the variant series. Genuinely lower-value than the above and the diff is large again.

## And thank you for the HP:0002153 correction

Noted and appreciated. Worth saying that it's the same failure shape as the deep-research report's `UBERON:0002018` and `HGNC:6398` in this run — a plausible CURIE attached to the right concept name. It happens at every level of this pipeline, which is an argument for checking identifiers mechanically rather than by recognition.

Snippets 44/44; `just validate-disorders` green.


---

nlharris:
Round 3 raced my push. Your review is recorded against `91f96089d`, but the text says the only commit since round 2 was the prose cleanup `04c8e66` — so you reviewed `04c8e66` and I pushed `91f96089d` at 01:12:50, ten seconds before you submitted at 01:13:00. Both round-3 blocking items were already in that commit. I've now pushed `2792e904d` on top with the "while you're in there" correction, which was a real defect and is worth more than the two items you led with.

## The reparenting was the actual bug

You were right and I want to be precise about what was wrong. `Steatorrhoea and Fat Malabsorption` hung off `Colonic Bile Acid Spillover`, while its own description says *"A contracted bile acid pool impairs micellar solubilisation of dietary fat."* Those are different claims about different compartments, and the graph asserted the one the prose denies. Fixed — it now hangs off `Contracted Bile Acid Pool`.

That is the second reachability/edge defect in this entry that no check catches. `check-entity-refs` verifies targets resolve, not that the graph says what the prose says. I filed monarch-initiative/dismech#9789 for the reachability half after your round-2 catch; this one is a different failure and I've noted it there.

## Also in `2792e904d`

**Five node → phenotype edges.** The pathograph stopped at the mechanism and never reached the clinical layer. Diarrhoea and growth failure off the colonic arm, steatorrhoea off its own node, hypocholesterolaemia off the cholesterol node, hypocalcaemia and growth failure off the vitamin arm — the last being your suggestion.

Growth failure is reached from *both* arms deliberately. The 2021 patient's catch-up growth followed sequestrant treatment of the diarrhoea; the 2025 patient's reduced final height came with no diarrhoea at all. One incoming edge would have picked a side the sources don't pick.

**OMIM 613291 — and why the answer to asking three times is not "yes".** I couldn't put it in `mappings` because `DiseaseMappings` has no OMIM slot; it carries only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings`. `Bleeding_Disorder_Platelet-type_21.yaml:76` hit the same wall and documented it in prose. So it's an `external_assertions` entry, with the reason recorded in its description. Worth knowing for future reviews: "add OMIM to `mappings`" is not an actionable suggestion in this schema.

**Changelog prose** stripped from the `Contracted Bile Acid Pool` description — that text was added in `91f96089d`, after the cleanup commit you praised, so it's new drift rather than a miss.

## What I'm not doing, and the reasoning

**The colonic bundle stays bundled.** You granted deferral for both bundles; I split one because the FXR axis gave it somewhere to go, and that split was the whole point of the FXR work. `Colonic Bile Acid Spillover and Secretory Diarrhoea` has no equivalent regulatory step to insert between its halves — splitting it produces two nodes joined by an edge that restates the node names. Your own round-3 argument for why the pool split needed the axis first applies here in reverse: there's nothing to put in the middle. If you disagree I'll take the argument.

**`biochemical:` C4 and FGF19 landed** in `91f96089d`, along with the FXR axis and both round-3 items, so the deferral condition about recording a follow-up is moot — there's nothing left deferred except the colonic bundle above.

## Validation on `2792e904d`

```
just validate                → schema + terms clean
just count-verified-snippets → 44/44 verified
```

And thank you for the round-2 correction on `HP:0002153`. Taking it back explicitly rather than quietly moving on is the thing that makes it worth checking you.

---

PR #9798, opened 2026-08-28
Add PhysioMap assessment report (2026-08-23)

cmungall:
## Summary

Adds a comprehensive assessment document comparing PhysioMap (a causal knowledge graph of human physiology) with dismech's architecture, design, and inference capabilities. This is a reference document for understanding how the two systems differ structurally and what gaps exist between them.

## Changes

- **New document**: `docs/reports/physiomap-assessment-2026-08-23.md` (1,061 lines)
  - Part I: Summarizes PhysioMap's published claims, methods, and evaluation results
  - Part II: Documents the curation discipline visible in their public repository (YAML sources, evidence tracking, expert review archive)
  - Part III: Presents independent analyses of the v1.1.1 release (node/edge topology, provenance breakdown, benchmark depth analysis)
  - Part IV–V: Structural comparison with dismech and recommendations for closing identified gaps
  
- **Updated navigation**: Added report to `mkdocs.yml` under Reports section

## Key Findings

The assessment identifies two separable gaps between dismech and PhysioMap:

1. **Edge polarity**: PhysioMap edges carry signed derivatives (`+`/`−`/`0`/`?`); dismech's 22,564 `downstream` edges are unsigned. PhysioMap's treatment/environmental edges already use polarity enums, so the infrastructure exists.

2. **Shared namespace**: PhysioMap's 1,699 traits are globally identified EQ-pattern classes forming one coupled system (largest SCC: 213 nodes). dismech's 11,923 pathophysiology nodes are file-scoped strings, creating ~2,000 disjoint per-disease graphs joined by manual anchors.

The first gap is cheaper to close; the second requires substantial namespace reconciliation work.

The document also highlights PhysioMap's curation discipline (machine-checkable provenance per edge, recorded failures, archived expert review with sha256 checksums) as independently worth studying for dismech's own governance.

## Notes

- This is a reference/analysis document only — it does not modify KB entries, schema files, or caches
- All counts and analyses are reproducible from PhysioMap v1.1.1 and dismech at commit `4b2f90c`
- The assessment is deliberately structured in layers (paper claims vs. actual artifacts vs. independent analysis vs. comparison) to avoid conflating different evidence sources

https://claude.ai/code/session_01R1ag6YGnWPsvEFJt7nGVwF

### DIFF
--- a/docs/reports/physiomap-assessment-2026-08-23.md
+++ b/docs/reports/physiomap-assessment-2026-08-23.md
@@ -1115,8 +1115,8 @@ sign and provenance distributions, scale/system/PATO/entity breakdowns, the comp
 constitutive/quantitative/modulation sets, duplicate-trait and sign-conflict checks, and the
 degree/topology analysis) and `benchmarks/results/e1b_forward_pairs.tsv` +
 `e1b_forward.json`. The §3.7 depth analysis is my own BFS over causal + production edges; its
-limits are stated inline. The release's headline counts reproduce the paper's exactly, which is
-a good integrity signal.
+limits are stated inline. The release's headline counts reproduce the paper's to within the
+two-edge causal drift noted in §3, which is a good integrity signal.
 
 **Part IV** figures are measured from this repo at commit `c0250519` (2026-08-28), the
 branch's merge base, and regenerate with `scripts/physiomap_release_analysis.py --dismech .`
--- a/scripts/physiomap_release_analysis.py
+++ b/scripts/physiomap_release_analysis.py
@@ -364,7 +364,7 @@ def dismech_counts(repo_root: Path) -> None:
     except ImportError:  # pragma: no cover - PyYAML is a project dependency
         sys.exit("PyYAML is required; run under `uv run`.")
 
-    nodes = edges = files = neg = 0
+    nodes = edges = files = neg = skipped = 0
     modifiers: collections.Counter[str] = collections.Counter()
     cyclic: set[str] = set()
     self_loops: list[tuple[str, str]] = []
@@ -374,9 +374,15 @@ def dismech_counts(repo_root: Path) -> None:
             try:
                 with path.open() as handle:
                     entry = yaml.safe_load(handle)
-            except Exception:
+            except Exception as exc:
+                # A metrics script that drops a file silently undercounts without
+                # saying so, which is the failure shape this script exists to prevent.
+                skipped += 1
+                print(f"  SKIPPED (unparsable): {path.name}: {exc}")
                 continue
             if not isinstance(entry, dict):
+                skipped += 1
+                print(f"  SKIPPED (not a mapping): {path.name}")
                 continue
             pathophysiology = entry.get("pathophysiology") or []
             if not pathophysiology:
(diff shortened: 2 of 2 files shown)


cmungall:
All three findings were right, and finding 1 was a real bug in my analysis rather than a presentation problem. Fixed in c6751125, along with the four suggestions.

## Finding 1 — the parsing bug

Confirmed. 27 rows of `e1b_forward_pairs.tsv` carry two semicolon-separated lesion variables, and they are exactly ARSA, CTNS, KYNU and PKLR — the four genes the paper excludes from its *inverse* benchmark for having two primary lesions. My parser read the whole field as one identifier, which manufactured the "27 unresolvable ids" row.

Our corrected numbers differ slightly, because we made different choices about those 27 rows:

| | Report (was) | Your figures (first lesion) | Now (min over lesions) |
|---|---:|---:|---:|
| Unreachable | 484 | 497 | 492 |
| Reachable | 355 | 369 | 374 |
| Unresolvable | 27 | 0 | 0 |
| One-hop | 155 | 157 | 161 |

I went with reachable-from-any-lesion: the paper specifies "one or more primary lesion variables" per gene and applies the solver once, so counting only the first drops real reach. Both readings are now stated in the text, so a reader can take either.

One thing that fell out of the fix. The corrected reachable count, 374, sits one pair away from the paper's 375 shortest-path directions. The path rule returns a direction exactly when a path exists, so that near-match corroborates the reconstruction — whereas the old text had to explain away a roughly 20-pair gap that was entirely my own artifact.

Took the cross-reference too: two of those rows exist *because* of the duplicate traits in §3.8, since `arsa_activity` and `arylsulfatase_a_activity` are the same trait under two identifiers.

## Finding 2 — "reconciles exactly" was wrong

The release has 2,268 causal edges against the paper's 2,270, and 2,385 relations against 2,387. The paper is internally consistent (2,270 + 85 + 4 + 9 + 19 = 2,387), so the drift is between preprint and release, not within either. Now stated precisely instead of rounded to "identical".

## Finding 3 — stale vintage

`4b2f90c` is dated 2026-08-11, as you say. Rather than relabel the column and keep a fortnight-old snapshot, I recomputed every Part IV figure at `c0250519`, this branch's current merge base: 15,296 pathophysiology nodes, 27,532 `downstream` edges, 2,517 files with a pathograph, 37 cyclic files (1.5%), description-regex share 18.5%. No conclusion moved.

You were right to single out the modifier census specifically. `LOSS_OF_FUNCTION` and `GAIN_OF_FUNCTION` are 106 and 37, not 4 and 5 — which reads very differently in a section arguing about where polarity information already lives.

## Suggestions

- §3.8 now names the predicate. 195 is the count over nodes carrying **both** an `entity_iri` and a `quality_iri`; grouping all nodes gives 202 and entity-only gives 196. The same-label conclusion — 5 groups, 6 redundant nodes — holds under all three, and all three are printed.
- §3.2 lists all 17 provenance lines instead of a top-10 that omitted two lines larger than one it kept, and notes that Guyton textbook content is 172 edges split across two labels, distinct from the 26-edge hand-curated core.
- §3.6 names BTK, tied for second among pure sources at out-degree 11 and not a metabolic enzyme.
- Added `scripts/physiomap_release_analysis.py`, which regenerates every Part III figure from a PhysioMap clone. Finding 1 is the concrete argument for it, and it follows the `experiments/README.md` precedent.

## Not done

Open question 1 — whether edge polarity is already tracked in an issue — still needs a human. My searches returned nothing on two phrasings and I read that as inconclusive rather than as an answer.

The two self-loops in §4.3 stay flagged in prose. Fixing them means editing KB entries, which this docs-only PR should not carry, and one of them (`Progressive Lung Disease` inside a periventricular nodular heterotopia entry) looks like a mis-pasted node rather than just a bad edge — as you noted, that wants a curator rather than a mechanical fix.

## Validation

82 tests pass (`test_render_projects`, `test_disclaimers`, `test_classify_page_build`), `ruff check` and `ruff format` clean on the new script, and the script reproduces every figure published in Part III.

---

PR #9800, opened 2026-08-28
Benign endometrial differentials, tamoxifen exposure, and a sex-steroid-driven proliferation module

cmungall:
Endometrial thickening had no representation in the knowledge base, and that turned out to be half right. A thick endometrial stripe is usually a benign finding, so it should not be curated as a phenotype — but the benign conditions it usually turns out to be were missing too, and so was the mechanism they share with the cancers. This branch fills in those gaps.

Four pieces, each building on the last.

## 1. Benign differentials on Endometrial_Carcinoma

The `differential_diagnoses` block held four entries — atypical hyperplasia, cervical cancer, uterine sarcoma, leiomyoma — but none of the common benign causes of postmenopausal bleeding with a thickened stripe, which is the presentation endometrial carcinoma is found through. Adds three:

- **Endometrial hyperplasia without atypia** (`MONDO:0006193`). The existing hyperplasia entry binds only the atypical/EIN form, so the more common non-atypical lesion had no representation.
- **Endometrial polyp** (`MONDO:0006195`).
- **Endometrial atrophy**, deliberately left unbound. The only close MONDO match, `MONDO:1016756`, is scoped to non-human animals, so the reasoning is recorded in the entry's notes rather than forcing a wrong term.

**Not added: an endometrial-thickness triage threshold.** Section 1 of `design-decisions.md` puts diagnostic decision support out of scope and the imaging section repeats it. A thick stripe is a low-specificity finding — of 593 women presenting with postmenopausal bleeding, 7.9% had carcinoma — so it belongs here as competing diagnoses, not as a rule.

## 2. Tamoxifen, and two curation-queue entries

Tamoxifen was absent from both endometrial entries despite being the classic iatrogenic risk factor. Added to `environmental:` on each, linked via `influences_mechanisms` to the `Unopposed Estrogen Signaling` node. It is the clearest case where one exposure produces both the malignant outcome and its commonest benign mimic. That a levonorgestrel IUS largely prevents the pathology is the evidence the route is the progestin-opposable one rather than a separate genotoxic effect.

`exposure_term` is left unbound. The pinned `sqlite:obo:ecto` build has no tamoxifen exposure term — `ECTO:9002223` and `ECTO:0900485` resolve on OLS but not locally — and the nearest available term, `ECTO:2000006`, sits under *exposure to hormone replacement therapy*, which would misclassify adjuvant anti-oestrogen therapy. Worth revisiting when the ECTO build is refreshed.

Also adds stubs for endometrial hyperplasia without atypia and endometrial polyp, both `UNDECIDED` with the lump/split arguments written out in notes.

## 3. Node de-drift between the two endometrial entries

`Endometrial_Endometrioid_Adenocarcinoma` declares `Endometrial_Carcinoma` as its parent and shares most of its pathophysiology, but the two modelled the same mechanisms under different labels. Renamed in the child entry:

| Was | Now |
|---|---|
| WNT Beta-Catenin Pathway Dysregulation | CTNNB1 WNT Pathway Activation |
| POLE Ultramutation | POLE Proofreading Defect and Ultramutation |
| Immune Checkpoint Vulnerability | Adaptive Immune Resistance |

All references were internal to that file. The last rename also makes the node name match the module node it already conformed to. Six node names are now shared verbatim, which is what made step 4 mechanical. Content stays duplicated rather than inherited, per the non-DRY convention.

## 4. A new mechanism module

`kb/modules/sex_steroid_driven_proliferation.yaml`. Five entries were each modelling the same chain separately with no `conforms_to` between them: sustained sex-steroid receptor signalling driving proliferation of a hormone-responsive tissue.

```
Sustained Sex-Steroid Ligand Exposure              trigger
  -> Nuclear Steroid Receptor Hyperactivation      amplifier
  -> Sustained Mitogenic Transcriptional Output    central_effector
  -> Target-Cell Population Expansion              effector
  -> Benign Hormone-Dependent Proliferative Lesion consequence
```

The consequence node carries the point that motivated the branch: the benign lesion is the usual endpoint, and malignant progression needs somatic drivers selected on that expanded substrate. Its description says so explicitly, so its presence in a cancer entry is not read as implying progression is expected.

Conformers wired: `Endometrial_Carcinoma` and `Endometrial_Endometrioid_Adenocarcinoma` (amplifier + consequence), `ER_Positive_Breast_Cancer` (amplifier + central effector), `Breast_Fibroadenoma` (amplifier + consequence), `Uterine_Leiomyoma` (effector). That is why three files appear here that the title would not lead you to expect — each is a single `conforms_to` line with its own history record.

**Scope.** `GO:0030520` appears in twelve entries but only five belong. Excluded: oestrogen *deficiency* states (`Aromatase_Deficiency`, `46_XX_Gonadal_Dysgenesis`, `Heart_Failure`, `Triple_Negative_Breast_Cancer`), receptor status curated as a tumour property (`Metastatic_Breast_Carcinoma`, `PIK3CA_Mutant_Breast_Cancer`, `Testicular_Sex_Cord_Stromal_Neoplasm`), and `Lymphangioleiomyomatosis`, where oestrogen drives survival and dissemination rather than tissue proliferation.

The amplifier is written at the nuclear-receptor level rather than as an oestrogen-specific step, because the leiomyoma conformer runs on progesterone (`GO:0050847`). Not curated as an Xogenesis module: terminal outputs range from a diffuse hyperplastic state to a discrete benign neoplasm, so no single anchor fits and none is asserted. The notes record the boundary against `sustaining_proliferative_signaling` — there the pathway is constitutively active and ligand-independent; here the receptor is intact and the abnormality is the duration and opposition of physiological ligand exposure.

## Validation

`just validate-disorders` on all five conformers (165/178 snippets verified, 13 skipped by prefix), the module through schema, term and reference validators (9/9 snippets), `just validate-modules` across all 158 modules, `check-entity-refs`, `check-snippet-grading`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-empty-snippets`, `check-snippet-boundaries`, `check-environmental-evidence`, `check-reference-cache-frontmatter`, `check-term-cache-integrity`, `check-not4curation`, `check-stubs`, `validate-stubs`, `validate-history-all`, and 15493 structural pytest cases.

One pre-existing warning surfaced and is not from this branch: `Uterine_Leiomyoma` cites `ega:EGAS00001004499` and no fetcher covers EGA accessions.

### DIFF
--- a/kb/disorders/Uterine_Leiomyoma.yaml
+++ b/kb/disorders/Uterine_Leiomyoma.yaml
@@ -135,7 +135,7 @@ pathophysiology:
   - reference: PMID:40214304
     reference_title: UTERINE FIBROIDS.
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "Progesterone, via its receptors in differentiated fibroid cell \npopulations, activates the production of Wingless-type MMTV integration site \nfamily (WNT) ligands, cytokines, and other growth substances to act on adjacent \nstem cells in a paracrine fashion to support tumor growth."
     explanation: >-
       Defines the progesterone-driven paracrine WNT mechanism supporting fibroid stem
@@ -143,7 +143,7 @@ pathophysiology:
   - reference: PMID:40214304
     reference_title: UTERINE FIBROIDS.
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "Both the mutated smooth muscle cells and the adjacent \ntumor-associated fibroblasts lay down excessive quantities of extracellular \nmatrix, providing a unique feature that led to naming these tumors \"fibroids.\""
     explanation: >-
       Establishes that mutated smooth muscle cells and tumor-associated fibroblasts
--- a/kb/modules/sex_steroid_driven_proliferation.yaml
+++ b/kb/modules/sex_steroid_driven_proliferation.yaml
@@ -96,6 +96,17 @@ pathophysiology:
     receptor carries the signal is the main tissue-specific substitution:
     oestrogen receptor in endometrium and breast, progesterone receptor in
     myometrium.
+
+    Two routes reach this node and both are in scope. In the endometrial
+    conformers an excess or unopposed ligand drives an otherwise normal receptor.
+    In the MED12-mutant mesenchymal conformers - breast fibroadenoma and uterine
+    leiomyoma - circulating steroid levels are unremarkable and an intrinsic
+    somatic lesion instead reshapes how the cell responds to them, so the
+    abnormality lies in the transduction rather than in the ligand. Those
+    conformers accordingly carry no ligand-exposure trigger upstream, and a
+    direction-agnostic modifier on the receptor process is appropriate where the
+    published signal is an altered response profile rather than a measured
+    increase.
   biological_processes:
   - preferred_term: nuclear receptor-mediated steroid hormone signaling pathway
     term:
@@ -142,11 +153,20 @@ pathophysiology:
   biological_scale: MOLECULAR
   description: >-
     The ligand-bound receptor drives transcription of growth and cell-cycle genes
-    for longer than the physiological interval. This is the rate-limiting,
-    tissue-agnostic step every conformer funnels through, and the intended key
-    conformance target: upstream the ligand source differs by disorder and
-    downstream the lesion differs by tissue, but the mitogenic transcriptional
-    programme is the shared core.
+    for longer than the physiological interval. Mechanistically this is the
+    rate-limiting, tissue-agnostic step: upstream the ligand source differs by
+    disorder and downstream the lesion differs by tissue, but the mitogenic
+    transcriptional programme is the shared core, which is what makes it the
+    natural key conformance target.
+
+    It is not yet the most-used attachment point, and the label should not be
+    read as a claim that it is. Only ER_Positive_Breast_Cancer conforms here
+    today, because that entry curates the transcriptional step as a node of its
+    own. The endometrial and mesenchymal conformers collapse transcription and
+    the proliferation it drives into a single lesion node and therefore attach
+    further downstream. That is a difference in curation granularity rather than
+    a disagreement about the mechanism; an entry that later separates the
+    transcriptional step should attach it here.
   biological_processes:
   - preferred_term: positive regulation of transcription by RNA polymerase II
     term:


cmungall:
Description rewritten to cover all four pieces of work, and the title changed to match. You were right that this mattered rather than being cosmetic — with squash-merge the body becomes the commit message on `main`, and a new cross-cutting module would have landed unmentioned.

Acted on three of the four suggestions, and partly on the fourth.

**Fibroadenoma conformance (suggestion 1) — kept, and the module text now accounts for it.** You identified a real gap in the module's prose rather than in the conformance. The amplifier node said the abnormality is the duration of activation and the absence of an opposing signal, which describes the endometrial conformers only. Two routes reach that node and both belong: an excess or unopposed ligand acting on a normal receptor, or an intrinsic somatic lesion reshaping how the cell responds to normal ligand levels. Both `Breast_Fibroadenoma` and `Uterine_Leiomyoma` are the second kind, and both are MED12-mutant, which makes the pattern a real one rather than a single awkward case. The node now says so, and says explicitly that a direction-agnostic modifier is appropriate where the published signal is an altered response profile rather than a measured increase.

**Key conformance target (suggestion 2) — the sentence was overstating, and is fixed.** "Every conformer funnels through" was not true of four of the five. The node keeps its status as the natural key conformance target on mechanistic grounds, but now states that it is not yet the most-used attachment point, that only `ER_Positive_Breast_Cancer` conforms there because that entry curates the transcriptional step as its own node, and that the others collapse transcription and proliferation into a single lesion node. That is a granularity difference, not a disagreement about mechanism. Adding an intermediate node to the endometrial entries is the other way to close this, but it would mean curating pathophysiology those entries do not currently assert, so I left it.

**PMID:40214304 grading (suggestion 3) — aligned on the node this PR touches, not across the file.** The two evidence items on `Hormone-Driven Myometrial Smooth Muscle Proliferation` are now `OTHER`, matching the module. I deliberately did not regrade the other five items citing that review in `Uterine_Leiomyoma`. They quote epidemiological passages — "the most common tumors in women", "cause heavy menstrual bleeding, anemia, pelvic discomfort" — where `HUMAN_CLINICAL` is correct. Per the guidance that a mixed-source paper gets one `evidence_source` per item based on the quoted passage, a blanket regrade would introduce errors rather than remove them. Happy to be argued out of this.

**Rebase (suggestion 4) — done**, onto `main` at `5a7a30257`. `check-entity-refs` and `check-snippet-grading` are now in range and both pass. `check-environmental-evidence` also passes cleanly now, confirming its earlier local failure was the baseline drift you diagnosed rather than anything on this branch.

Re-validated after the rebase and edits: module through schema, term and reference validators (9/9 snippets), `validate-disorders` on all five conformers, the offline gate set, and 15493 structural pytest cases.

---

cmungall:
On the missing history record for the review-response commit — that is a choice, not an oversight, and here is the reasoning so it can be overruled if you disagree.

The convention in `CLAUDE.md` is scoped to the pull request: "Any PR that creates or edits a KB entry (`kb/disorders/`, `kb/modules/`, `kb/comorbidities/`) should add a matching history record." Every KB entry this PR touches has one. `Uterine_Leiomyoma` and the module both got records in the commit that introduced the module and its conformers, and a history file is described as recording one session for one target — this is the same session, and the follow-up commit refines changes those records already cover.

Records are append-only, so the alternative is two new files saying that two node descriptions were reworded and two `evidence_source` values regraded. That is thin provenance, and pushing it would drop the approval under `dismiss_stale_reviews` and cost a full review cycle. Not worth it for this.

Happy to add them if you would rather the convention be read per-commit — say so and I will push, accepting the re-review.

---

Issue #7439, opened 2026-07-31
Experimental SEPIO evidence model: schema, validators, and rendering


caufieldjh:
## Summary

Pilot an experimental **SEPIO** (Scientific Evidence and Provenance Information Ontology)
representation of dismech evidence, alongside — not replacing — the current native
`EvidenceItem` model.

Today a dismech assertion carries a flat list of `evidence:` items, each of which bundles
four separate things into one object: *what kind of study this is* (`evidence_source`),
*which way it cuts* (`supports`), *the actual quoted data* (`snippet`), *where the quote came
from* (`reference`), and *why the curator thinks it bears on the claim* (`explanation`).

SEPIO pulls those apart into an argument structure:

- an **EvidenceLine** — one independent argument for/against the assertion, carrying the
  direction, the evidence type, and (optionally) a strength
- one or more **evidence items** (`DataItem`s) — the discrete information entities being
  interpreted as evidence; for dismech that is almost always a text span
- a **Document** the item was `reported_in` — with its own identity, title, and type
  (primary literature vs. review vs. structured database vs. trial record)

The upstream SEPIO folks have drafted worked examples against our own Cystic Fibrosis
entry, showing the native form and then the SEPIO form for the same two pathophysiology
assertions:

- Examples: https://github.com/sepio-framework/sepio-linkml/blob/main/workspace/dismech-examples.yaml
- Full SEPIO LinkML schema: https://github.com/sepio-framework/sepio-linkml/blob/main/src/sepio_linkml/schema/sepio_classes.yaml

## Why bother

Three things the flat model can't currently say:

1. **Two snippets from the same paper are two arguments, or one.** Today they're just two
   sibling evidence items with no way to say which. An EvidenceLine can hold several
   evidence items that jointly make one argument.
2. **Document typing.** We have no slot for "this is a review, not a primary study" or
   "this is a structured-database row." SEPIO's `Document.document_type` gives us that,
   and it's information we already implicitly rely on when weighing evidence.
3. **Strength separate from direction.** `supports: SUPPORT` conflates "which way" with
   "how hard." SEPIO splits `direction_of_evidence_provided` from
   `strength_of_evidence_provided`.

## Scope (deliberately partial)

This is an experiment, not a migration. Native `evidence:` stays exactly as it is, on
every class it's on now, and remains the thing curators write by default.

### 1. Schema (`src/dismech/schema/dismech.yaml`)

Add three new classes and a new slot, additively:

[...]

PR #9806, opened 2026-08-28
Generalize ASO detail slot to cover siRNA, add RNAi mechanism module

cmungall:
The schema could describe an antisense oligonucleotide's target, chemistry and
conjugate, but had nothing equivalent for siRNA. In ATTR_Amyloidosis that showed
up directly: inotersen carried a structured aso_details block while vutrisiran
and patisiran, sitting beside it, recorded "subcutaneous GalNAc-conjugated" and
"intravenous lipid-nanoparticle" only as free text in the description. Fourteen
siRNA treatments across twelve entries were in that state.

Schema
- aso_details -> oligonucleotide_details (OligonucleotideDetail), covering both
  ANTISENSE_OLIGONUCLEOTIDE and SIRNA. The two platforms differ in effector
  (RNase H1 vs Argonaute-2) but share target, chemistry and delivery attributes,
  so one class makes "every treatment silencing gene X" a single query.
- Mechanism enum gains RNAI_KNOCKDOWN (GO:0070551, the molecular-function
  sibling of the RNase H term already used); chemistry gains TWO_PRIME_FLUORO.
- New delivery_platform slot, orthogonal to conjugation: patisiran is
  UNCONJUGATED *and* LIPID_NANOPARTICLE, vutrisiran is GALNAC *and* CONJUGATE.
  Conjugation alone cannot express that distinction, which is what sets route,
  dosing interval and whether premedication is needed.
- New dosing_interval / dosing_interval_days on Treatment, following the
  Prevalence convention of verbatim phrasing plus a normalized number. Treatment
  had no way to record dosing at all.
- aso_details, aso_mechanism and aso_chemistry are retained as deprecated
  aliases so entries authored before this keep validating.

Content
- Migrated all 35 existing ASO blocks to the new spelling, values unchanged, and
  derived delivery_platform from the conjugation already present. That rule is
  sound for ASOs specifically because no approved ASO is nanoparticle-delivered;
  it is not applied to siRNA.
- Populated the new block on 13 siRNA treatments. Gene bindings were reused from
  already-validated HGNC bindings elsewhere in the KB. Unsourceable facts were
  left empty rather than guessed: plozasiran's approved dosing interval is not in
  the reference cache, and oligonucleotide_chemistry is unset throughout because
  stabilized duplexes mix 2'-OMe and 2'-F and the slot is single-valued.
- Hemophilia's SIRNA-tagged treatment is deliberately untouched: it is a
  rebalancing class covering fitusiran plus two monoclonal antibodies, so a
  single oligonucleotide detail block would misdescribe it. The modality tag on
  that mixed class looks wrong and is worth a separate look.

New module kb/modules/rnai_gene_silencing.yaml
Sibling of antisense_oligonucleotide_therapy, whose notes already excluded
siRNA/RNAi and had nowhere to send it. Five nodes run transcript abundance ->
delivery -> RISC loading -> Argonaute-2 cleavage -> sustained protein reduction,
with evidence spanning three genes and three diseases so it is grounded as
cross-disease rather than ATTR-specific.

Delivery is a node inside the chain, not a module of its own. Conformance exists
so an entry can substitute organ-specific content, and GalNAc chemistry is
identical everywhere, so a delivery module would cost full duplication for no
disease-specific variation. Platform detail belongs on the treatment instead.

ATTR_Amyloidosis is wired in as the first conformer, with four nodes carrying
conforms_to. It earns the exemplar slot because patisiran and vutrisiran target
the same transcript by different platforms.

Validation
42958 pytest checks pass; validate-disorders clean over all 37 changed KB files;
schema, term, duplicate-key, entity-ref, folded-hyphen, snippet-length,
title-snippet and environmental-evidence gates all pass. GO:0016246 was the
obvious candidate for RNA interference and turned out to be obsolete with no
label, so it was dropped rather than guessed. linkml-reference-validator skips
DOI: prefixes, so all 13 DOI snippets were verified by hand as
whitespace-normalized substrings of the cache body.

references_cache churn is tool-generated (reference_id quoting, an
full_text_attempted flag, and PMC full text newly fetched for PMID:38914784);
no cache file was hand-edited.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011MBpsiUiYonB4GSxNsmkkw


cmungall:
Conflicts resolved. The branch is rebased onto current `main` (451 commits of drift) and now shows `blocked` rather than `dirty`, so it is waiting on checks and review rather than on a merge.

Only two files actually conflicted, and both were the benign union case @ai4c-agent guessed at for the sibling PR: `cache/go/terms.csv` and `cache/ncit/terms.csv`. Both sides only appended rows, neither deleted any, and no CURIE was added by both. Everything else — 35 disorder entries, the schema, the templates — merged on its own.

I did not hand-merge those CSVs, since they are derived artifacts. I took main's copy of each and re-derived our rows with `just validate-terms`, then `just normalize-cache`. Result: all 228 rows main added to the GO cache are intact, plus our 4 RNAi terms and nothing else. The NCIT cache came out byte-identical to main — main had independently derived the same `NCIT:C169973` (Evinacumab) row, so the only real difference was a derivation timestamp.

**Two things the textual merge could not have caught, both now fixed.**

Item 1 — four entries merged into `main` while this PR was open, and were authored against the old slot names: `Werner_Syndrome`, `Hereditary_Spastic_Paraplegia_49`, `Spinocerebellar_Ataxia_Type_36`, `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_31`. They validate fine, because `aso_details` is retained as a deprecated alias. But leaving them would mean shipping two spellings for the same block, the next curator copying whichever file they opened first, and a future removal of the alias being blocked. They now use `oligonucleotide_details`, by the same mechanical transform as the original 35, and the deprecated alias has zero users in `kb/`.

Item 2 — issue #7439 removed `PARTIAL` from `EvidenceItemSupportEnum` while this PR was open. One evidence item on this branch used it: the HELIOS-B vutrisiran citation on ATTR amyloidosis's `Reduced Circulating TTR Precursor Pool` node. The YAML merged cleanly because the conflict was semantic, not textual — nothing in a three-way merge can see that a value legal when written has since become illegal. `test_no_out_of_enum_values_in_kb` caught it, which is that test doing exactly what its failure message describes.

The item's original explanation already said why it was `PARTIAL`: the trial reports clinical outcomes, not the serum TTR reduction the node asserts. That is the sense the narrowing maps onto `supports: SUPPORT` plus the new `directness` slot, and `INDIRECT`'s description names the case outright ("a therapeutic response cited as validation of the mechanism it targets"). Recorded that way rather than as `NO_EVIDENCE` or split into a pair. The `rnai_gene_silencing` module's notes also told conforming curators to mark a clinical-outcome claim `PARTIAL`, which would have propagated the dead value into future entries; that guidance now says `SUPPORT` + `INDIRECT`.

**One thing I did not change.** `kb/modules/antisense_oligonucleotide_therapy.yaml` still instructs curators to mark an RNA-target link `PARTIAL`. It has the same staleness problem, but it is pre-existing on `main` and is somebody else's curation guidance rather than this PR's to rewrite. Worth a follow-up. Every other `PARTIAL` in the files this branch touches is prose in an `explanation` or `notes` field recording how an item was graded historically, not a live slot value.

Also unchanged, and worth flagging separately: `main` now has 19 `therapeutic_modality: SIRNA` treatments, up from the 14 this PR started with. The 5 new ones would benefit from an `oligonucleotide_details` block, but that needs per-drug sourcing rather than a mechanical transform, so it is follow-up curation, not conflict resolution.

**Validation after the rebase:** 51,395 pytest checks pass; `just validate-disorders` clean over all 39 changed disorder files and both modules; duplicate-key, entity-ref, folded-hyphen, snippet-length, title-snippet, environmental-evidence, term-cache-integrity and schema term validation all pass.

---

cmungall:
Both blockers addressed in `a348d38aa`, and the `scope-override` label is now on — which is what the failing `test (3.13)` check was asking for. That job failed the curation-scope gate listing this PR's nine infra files (`CLAUDE.md`, the schema, the two templates, `render.py`, `term_tooltips.py`, `treatment_platform.py`, `render_review_pdf.py`, the new test) and said "If intentional, add the 'scope-override' label". It is intentional — this is a schema change with its KB migration — so the label is the right answer rather than splitting the PR. The push re-triggers CI with the label present.

**The RISC-loading node is folded away.** @ai4c-reviewer was right, and I verified it independently rather than taking the finding on trust: the cited cache body contains zero occurrences of "RISC", "passenger", "guide strand" or "silencing complex", so the node asserted three claims its source never makes, and `GO:0016442` and `GO:0035197` were sitting on nothing. That fails the bar the module's own notes set.

I folded rather than went looking for a new source. The loading step is now setup prose inside the Argonaute-2 cleavage node, carrying no ontology binding and no evidence of its own, and I dropped the two unevidenced bindings rather than relocating them to a node whose source does not mention them either. Blast radius matched your reading: one inbound edge from the delivery node, nothing in `kb/`, `src/`, `tests/` or `docs/` referencing the name, and ATTR amyloidosis already collapsing loading and cleavage into one node.

One addition to your suggestion. I added a "Node granularity" note to the module recording why the node is absent, what the sources do and do not say, and that it can be split back out if a source describing RISC loading is later cached. Without it the next curator sees a chain that skips a textbook step and has no way to tell a decision from an oversight. I also fixed a phrase in the delivery node that called RISC loading catalytic — cleavage is the catalytic step.

**Loading doses are out of `dosing_interval`,** all four. You were right to raise this above non-blocking: a convention broken in four of its own twelve first uses is the version that gets copied. Nothing was lost, since every one of those treatment descriptions already states the loading schedule and `dosing_interval_days: 182.5` was the maintenance figure either way.

**Stale prose: you found three, a full grep found five.** `Spastic_Paraplegia_90A_Autosomal_Dominant.yaml` and `Hemophilia.yaml` also referred to `aso_details` in notes. All five now use the new spelling. Prose only — no slot value, binding, evidence item or claim changed. `Spastic_Paraplegia_90A` was not otherwise touched by this PR; it is in scope only because it names a slot this PR renames.

**Not changed:** the trigger node's `modifier: INCREASED`. You flagged it twice and declined to hold the PR for it both times, and I agree with that call rather than just deferring to it — the value mirrors `antisense_oligonucleotide_therapy`'s trigger node exactly, so changing it here would split two sibling modules that should stay aligned. The right fix is one change across both, which is a separate PR. The ATTR conformer's explicit disclaimer stands in the meantime, so the contradiction is at least visible where it matters.

**After the change:** `check-enum-values`, `check-duplicate-keys`, `check-entity-refs` and `check-causal-targets` all clean; `validate-disorders` clean over the eight changed disorder files; the module validates; `validate-history-all` clean over 8,225 records; 48 targeted tests pass. Branch is 144 commits behind `main` but merges clean — no conflict this time, so I have not rebased again.

---

cmungall:
`test (3.13)` went red on `a348d38aa` — ruff `I001` in `src/dismech/render.py`. My import: I put `treatment_platform` directly after `term_labels`, but `term_tooltips` sorts before it. Fixed in `b87adfa0e`, import reorder only.

The interesting part is why local checks missed it, because the obvious explanation is wrong. Lint is not part of pytest, so the full local suite passing said nothing — but `uv run ruff check .` **and** `uvx ruff check .` also passed locally. `pyproject.toml` deliberately tracks ruff's default rule set without pinning a version, and CI runs `uvx ruff check .`, which resolves the newest ruff at run time. My cache held 0.15.8, where `I001` is not in the defaults; CI got 0.16.6, where it is.

So the local equivalent of this CI step is `uvx ruff@latest check .`, not `uv run ruff check .`. That command reproduces the exact CI error, and passes after the fix. A whole-repo run under 0.16.6 reports this as the only error, so nothing else in the tree is newly non-compliant — this is a one-line fix, not a version bump landing on the repo.

Flagging it because the unpinned-by-design setup means any contributor whose `uvx` cache is behind will pass locally and fail here, and the failure will look like a mystery.

---

cmungall:
**`test (3.13)` is green on `b87adfa0e`.** Both things that made it red are fixed: the `scope-override` label for the curation-scope gate, and the ruff `I001` import order. No merge conflict either — `mergeable_state` is `blocked`, not `dirty`.

**`claude-review` failed, but not on anything in this PR.** It errored after 0s ("Claude encountered an error"), and the job log is nothing but post-job cleanup — no checkout, no analysis, no output. The reviewer agent died at startup. Notably its previous run on `a348d38aa` had already worked through its whole checklist and verified all three items before this run replaced it, so there is no finding here to act on.

**I cannot re-run it.** Both routes return `403 Resource not accessible by integration`:

- `POST actions/runs/33938569188/rerun-failed-jobs`
- `POST actions/workflows/claude-code-review.yml/dispatches`

So this needs someone with Actions write. @ai4c-agent has re-dispatched `claude-code-review.yml` on this PR before — that is the natural path, and its next scheduled sweep should pick this up. Failing that, a maintainer can re-run the job, or any further push to the branch re-triggers it.

I am **not** dismissing the two standing `CHANGES_REQUESTED` reviews. Their blockers are fixed in `a348d38aa`, but clearing a sticky review is the reviewer's call, not mine.

Nothing else is outstanding that I can see: CI green, no conflict, both blockers addressed, the non-blocking prose item fixed in five files rather than the three flagged. I will keep the PR watched and check back rather than leave it here.

---

cmungall:
Correcting one thing in my comment above: I said the shepherd's next sweep should pick up the failed `claude-review`. That has not happened, and I should have checked the elapsed time before writing it — the job died at 03:05 UTC and it is now ~17:50, so roughly 15 hours with no re-dispatch. The newest review on this PR is still the one against `f591350fc`; neither `a348d38aa` nor `b87adfa0e` has been reviewed.

So this is waiting on a person, not on a scheduler. @cmungall — one re-run of [job 101231296731](https://github.com/monarch-initiative/dismech/actions/runs/33938569188/job/101231296731), or a `claude-code-review.yml` dispatch for this PR, is all it needs. Both return 403 for me.

Nothing else has changed: `test (3.13)` green on `b87adfa0e`, no merge conflict, both blockers from the standing review fixed.

---

cmungall:
Root cause found, and it corrects what I said earlier — a re-run alone will not fix this, and the one that just happened failed for the same reason.

`claude-review` has now failed twice at 0s (03:05 yesterday, 05:08 today). The job log gives the real reason:

```
error='' detail="You've hit your limit · resets 7:30am (UTC)" num_turns=1 cost=$0
```

**This is a Claude account quota on the credential the workflow runs under, not anything about this PR.** `num_turns=1`, `cost=$0` — the agent was refused before it did any work. Nothing in the diff can cause or fix it.

Two things follow.

**Timing.** The limit resets at 07:30 UTC. Any re-run before then fails identically; a re-run after then should work. I have moved my own check-in to after the reset rather than asking for another re-run now.

**Configuration, per the workflow's own error handler.** It prefers `ANTHROPIC_API_KEY` (pay-per-use, no weekly cap) and only falls back to `CLAUDE_CODE_OAUTH_TOKEN` when no API key is set — so a limit error here suggests `ANTHROPIC_API_KEY` is unset and the run used the subscription token. Setting that repo secret would stop this recurring across every PR, not just this one.

**A small bug worth fixing while you are in there.** `claude-code-review.yml` has a branch that detects exactly this and prints a clear "usage limit reached" message, but its match is `reason == "rate_limit" or "usage limit" in haystack or "weekly limit" in haystack`. The actual message is "You've hit your **limit**" — none of those three substrings — so it fell through to the generic handler and reported "Common causes: an invalid/retired `--model`, a bad `CLAUDE_CODE_OAUTH_TOKEN`, or an aborted run." That is what sent me looking at the model and token first. Adding `"hit your limit"` to the match would make the next occurrence diagnose itself.

Everything on the PR side is unchanged and ready: `test (3.13)` green on `b87adfa0e`, no merge conflict (verified against main at 292 commits behind), both review blockers fixed, prose cleanup done in five files.

---

cmungall:
Thanks for the re-review. Two things to record: what the last push changed, and the one nit I am deliberately leaving.

## What `21827fd2c` changed

Main moved 328 commits while this PR sat blocked, so before doing anything I checked whether the merge was safe: `merge-tree` wrote a clean tree, and I built that exact tree in a throwaway worktree and ran `check-enum-values`, `check-duplicate-keys`, `check-entity-refs` and `check-causal-targets` against it. All clean, and the tree I then committed is byte-identical to the one I tested.

That sweep turned up two entries the merge would have landed silently, both authored against the old slot names:

- **`Spinocerebellar_Ataxia_Type_1`** — a new entry on main whose ATXN1-lowering ASO treatment carries an `aso_details` block. Keys renamed in place. I left `delivery_platform` unset: the block records no conjugation to derive it from, and the therapy is preclinical, so there is no approved formulation to source one.
- **`Cytomegalovirus_Retinitis`** — its block was migrated earlier in this PR, but evidence added on main afterwards explains the backbone chemistry by naming `aso_details`, so the merged file described the block by a slot name it no longer used. Only that prose changed; reference, snippet, `supports` and `evidence_source` are untouched.

That is what gets `kb/` to zero users of the deprecated aliases rather than two.

One correction worth flagging, since it is the failure mode this repo has a gate for: my first edit to the CMV explanation dropped the word "technology" from a folded scalar, turning "drug technology evaluation" into "drug evaluation". I caught it by parsing the YAML and reading the rendered string rather than eyeballing the diff, and restored it. `check-folded-hyphens` is green, but it would not have caught this one — the corruption was a dropped word, not a hyphen split.

## The nit I am not fixing

You are right that `rnai_gene_silencing.yaml:168` says "below" when the note is above — it is at line 41, inside `notes:` at line 22, and `pathophysiology:` does not start until line 64. Above in the file and above on the rendered page.

I am leaving it anyway. `main` is protected with `dismiss_stale_reviews`, so pushing a one-word change would drop the approval this PR waited five days to get, and the next review cycle costs far more than the wrong direction word. Line 8 already refers to the same note without a direction, which is the phrasing to copy when it is fixed.

The natural place for it is the PR that changes the trigger node's `modifier: INCREASED` — that has to touch both this module and `antisense_oligonucleotide_therapy` together, so it opens this file anyway.

## Still open

`test (3.13)` is running on the new head and is not green yet. I will keep watching until it lands.

---

PR #9325, opened 2026-08-23
Curate computational models for Heart Failure

cmungall:
## Summary

- add a 2024 CircAdapt par-75 imaging-based HFrEF/LBBB digital twin with its exact public repository commit
- add a distinct paper-curated virtual-pacing twin that perturbs CRT delivery and predicts reverse remodeling
- link contractile-dysfunction and ventricular-remodeling readouts to existing pathophysiology nodes
- document cohort size, identifiability, absent patient inputs/solver/license, and paper-only implementation limitations
- preserve and citation-validate the completed 5-iteration OpenScientist exploration (job `e0fc2fd2-4a7d-405f-a1c1-8498f68c79d3`)

## Verification

- full disorder validation passed (54/54 snippets; only pre-existing unsupported dataset-prefix warnings)
- LinkML disorder and history schemas passed
- term, folded-hyphen, duplicate-key, export/render, and graph/data-invariant checks passed
- MATLAB/Octave execution was unavailable; exact runtime blockers are recorded in `notes`
- all 47 OpenScientist PMIDs resolved; two close abstract-quote mismatches are transparently retained in the validation report

One disorder file is modified, per the computational-model curation workflow.

---

> [!NOTE]
> **Low Risk**
> Knowledge-base curation only: YAML model records, search export, history, and reference caches. No application logic, auth, or runtime behavior changes.
> 
> **Overview**
> Adds the first **computational models** on Heart Failure: an imaging-calibrated CircAdapt par-75 left-ventricular digital twin (pinned GitHub commit, PMID:38741182) and a paper-only virtual-pacing CRT-response twin (PMID:38288616).
> 
> Both are `DIGITAL_TWIN` entries with variables, findings, and mechanism links (`Myocardial Contractile Dysfunction`, `Ventricular Remodeling`). Notes record reproducibility limits (no license/tests/patient data/solver; MATLAB not run; CRT study has no code deposit). Search metrics now show 57 models / 22 sources / 26 with repositories.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f82fb4bd9964e42a7c98883a9cbbc6a80ce193c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9324, opened 2026-08-23
Curate computational models for Epilepsy

cmungall:
## Summary

- add the 2025 Virtual Epilepsy Patient Cohort as a patient-specific whole-brain digital twin
- link its seizure-network and regional-excitability readouts to existing pathophysiology nodes
- record independent clinical calibration, negative stimulated-propagation results, exact repository commit, and reproducibility limitations
- preserve and citation-validate the completed 5-iteration OpenScientist exploration (job `bdf24227-454c-4702-ab34-6d86082a31c9`)

## Verification

- full disorder validation passed (121/121 snippets; only pre-existing unsupported dataset-prefix warnings)
- LinkML disorder and history schemas passed
- term, folded-hyphen, duplicate-key, export/render, and graph/data-invariant checks passed
- repository Python scripts byte-compile; end-to-end run was blocked by external patient data/dependency gaps documented in `notes`
- all 55 OpenScientist PMIDs resolved; one transparent relevance-heuristic flag is retained in the report

One disorder file is modified, per the computational-model curation workflow.

---

> [!NOTE]
> **Low Risk**
> Knowledge-base curation and generated search/reference artifacts only; no application logic, auth, or data-handling code.
> 
> **Overview**
> Adds a curated **digital twin** computational model to the Epilepsy knowledge base: the 2025 Virtual Epilepsy Patient Cohort of 30 patient-specific whole-brain network models (Epileptor / The Virtual Brain).
> 
> The entry links seizure-network and regional-excitability readouts to existing hyperexcitability mechanisms, records independent VEP clinical calibration (PMID:35604575), and keeps the **negative** stimulated-seizure propagation result. Repository provenance is pinned to commit `a99c883`, with explicit reproducibility limits (no license/deps, external EBRAINS data, missing named script). Search index totals increment by one model.
> 
> Also caches many epilepsy PMIDs and a history event; one unrelated clinical-trial explanation is rewrapped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5b2df846c9f99a8cec8767eeaf5a970e098e5d3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9263, opened 2026-08-22
Add Alzheimer DADD digital twin model

cmungall:
## Summary

- adds the 2025 EEG-personalized Digital Alzheimer's Disease Diagnosis (DADD) model
- adds a cortical structural-connectivity degeneration join point and links DADD separately to connectivity degeneration and synaptic dysfunction
- records patient-specific variables, diagnostic/prognostic findings, the nonsignificant synaptic-parameter result, repository provenance, and reproducibility limitations
- regenerates the computational-model browser index

## Model provenance

- publication: PMID:40450374
- repository: https://github.com/LoreAma/Code_AD_simulations
- inspected commit: `bf480f5cbba429639fcd6f1cb8dc92081a4f5975`
- license: GPL-2.0
- execution: not attempted because the notebook repository has no pinned dependency manifest; this limitation is explicit in the card

## Validation

- `just validate kb/disorders/Alzheimer_Disease.yaml`
- `just count-verified-snippets kb/disorders/Alzheimer_Disease.yaml` — 241/247 verified, 6 non-PMID-prefix skips
- `just check-duplicate-keys kb/disorders/Alzheimer_Disease.yaml`
- `just validate-history history/disorders/Alzheimer_Disease/2026-08-22T183457Z-claude-code-b4684a.yaml`
- `uv run pytest tests/test_models_export.py tests/test_graph_model_links.py tests/test_render_models.py -q` — 25 passed

---

> [!NOTE]
> **Low Risk**
> Knowledge-base curation and generated search-index data only; no application logic, auth, or runtime behavior changes.
> 
> **Overview**
> Adds the 2025 EEG-personalized **DADD** digital twin (`PMID:40450374`) to Alzheimer Disease, including GitHub provenance, inversion parameters (`cp`, `lp`, `np`), diagnostic/prognostic findings, and explicit reproducibility limits (no pinned deps; patient EEG not in-repo).
> 
> Introduces a **Cortical Structural Connectivity Degeneration** mechanism (cortex/white matter) downstream of synaptic dysfunction and upstream of memory loss, and links DADD to that node plus synaptic dysfunction with fidelity, readouts, and the nonsignificant synaptic-parameter result.
> 
> Regenerates the computational-model browser index (56 models) and caches supporting PMIDs.
> 
[...]

PR #9122, opened 2026-08-20
Add Ataxia-Telangiectasia computational models

cmungall:
## Summary\n\n- add three Ataxia-Telangiectasia computational-model entries: an ATM/ATR ODE, an ATM p.Ala2386Glu structural analysis, and a normative MRI autoencoder\n- link each model conservatively to the exact pathophysiology mechanism it represents or measures\n- record artifact provenance, validation scope, fidelity, and reproducibility limitations\n- regenerate the computational-model browser export and add the curation history record\n\n## Reproducibility notes\n\n- the pinned ATM/ATR SBML was cross-checked against BioModels, parsed without libSBML errors, and smoke-run in COPASI and libRoadRunner\n- the ODE curation records the released 31-species/40-reaction artifact versus the paper-reported 32/41, qualitative units, and missing disease/drug scenario files\n- the autoencoder repository lacks trained weights and required data-preparation steps; its notebook instantiates a five-unit latent layer while the paper reports four\n- the allele-specific structural model has experimental patient-cell support but no deposited mutant coordinates or runnable artifact\n\n## Validation\n\n- Schema validation...
No issues found
Term validation...
warning: `VIRTUAL_ENV=/Users/cjm/repos/tatr/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
✅ Validation passed
Reference validation...
warning: `VIRTUAL_ENV=/Users/cjm/repos/tatr/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
Cache directory: references_cache

Validating kb/disorders/Ataxia_Telangiectasia.yaml against schema src/dismech/schema/dismech.yaml

Validation Summary:
  Files validated: 1
  Total checks: 0
  All validations passed!
  Snippets checked: 120/120 verified against cached references
Normalizing term caches...

Migration complete:
  Files scanned: 18
  Files updated: 0
Normalizing enum caches...
✓ All caches normalized
✓ All validations passed for kb/disorders/Ataxia_Telangiectasia.yaml (120/120 snippets)\n- history, duplicate-key, title-snippet, snippet-length, and NOT4 audits\n- 25 focused computational-model export/render/link tests\n- 8,656 model-related data tests\n- \n\nA complete rename-aware audit of 184 open pull requests found no existing PR touching this disorder YAML.

---

> [!NOTE]
> **Low Risk**
> Knowledge-base curation and generated search export only; no application logic, auth, or runtime behavior changes.
> 
> **Overview**
[...]

PR #9807, opened 2026-08-28
fix(ci): let generate-pages own app/models/data.js, not curation PRs

### LINKED ISSUE #9804: Nine approved model PRs are deadlocked: a test forces every curation PR to commit app/models/data.js
## Symptom

Nine PRs are approved, unassigned, green, and **all mutually un-mergeable**. Every one conflicts on exactly one file, `app/models/data.js`, and on nothing else:

| PR | Branch |
|---|---|
| #9104 | `codex/hyperinsulinemic-hypoglycemia-computational-models` |
| #9119 | `codex/noonan-syndrome-computational-models` |
| #9122 | `codex/ataxia-telangiectasia-computational-models` |
| #9140 | `codex/renal-cell-carcinoma-computational-models` |
| #9142 | `codex/hepatitis-c-computational-models` |
| #9145 | `codex/chronic-myeloid-leukemia-computational-models` |
| #9263 | `codex/alzheimer-dadd-digital-twin` |
| #9324 | `codex/epilepsy-vep-models` |
| #9325 | `codex/heart-failure-circadapt-model` |

Of the 22 open PRs that are currently approved + unassigned + non-draft, **13 are un-mergeable**, and these nine are the largest cluster. The deterministic auto-merge sweep requires `mergeable == MERGEABLE`, so none of them can be swept.

## Cause

`tests/test_models_export.py::test_exported_data_js_is_in_sync_with_the_knowledge_base` asserts that the committed file equals what the exporter produces right now:

```python
assert committed == current, "app/models/data.js is stale — run `just gen-models-data`"
```

`app/models/data.js` is a generated browser index — 2,834 lines, 131 KB — rebuilt from **every** `computational_models` block in `kb/` by `dismech.export.models_export`. Any PR that adds a model changes `current`, so under this contract the curation PR must regenerate and commit the whole artifact. Each of the nine did exactly that, in its own curation commit:

```
#9145  8789c8362d  Add CML computational model                        app/models/data.js | 66 +++---
#9104  e702717586  Curate hyperinsulinemic hypoglycemia computational models
#9324  f6d350c954  Curate epilepsy computational models
```

Two concurrent model PRs therefore cannot both merge, by construction.

## The repo already rejected this pattern for the sibling index

`app/data.js` has the identical shape, and a test like this was deliberately **not** written for it — `tests/test_check_browser_data_links.py:195`:

> Deliberately NOT tested here: the committed `app/data.js` against the committed `pages/`. Every curation PR adds a disorder YAML before the generate-pages workflow renders its page, so such a test would fail on the whole curation stream rather than on the drift it is meant to catch. **The invariant belongs to the generate-pages run (which renders first, then gates), not to repo state.**

That reasoning transfers word for word. `app/discussions/data.js` is the control case: same kind of generated index, no staleness test, and no conflicts.

## Second-order effect: it also red-lights `main`

The pytest job is path-gated — `if: steps.changes.outputs.python == 'true'`, matching `src/**`, `tests/**/*.py`, `justfile`, `project.justfile`, `pyproject.toml`, `uv.lock`. A kb-only curation PR does not match, so **the staleness test never runs on the PR that creates the staleness**. It accumulates silently across merges until some unrelated PR touches `src/` or `justfile`, runs the full suite, and goes red for something it did not cause. That is what `d7de426b2 chore(models): regenerate app/models/data.js (fixes red main)` was cleaning up.

So the incentives are inverted both ways: regenerate and you deadlock against every sibling model PR; skip it and you poison `main` for whoever next touches `src/`.

## Third gap: nothing regenerates it

`generate-pages.yaml` runs `gen-browser-data`, `gen-pathographs`, `gen-dashboard` and `gen-schema-docs`, and commits `pages/ app/data.js pathographs/ dashboard/ elements/` (line 346). It never runs `just gen-models-data` and never adds `app/models/data.js`. So simply deleting the test would leave the file to rot — the fix has to move ownership to the bot, not just drop the constraint.

## Fix

Move the invariant to the generate-pages run, matching `app/data.js`:

1. `generate-pages.yaml` — add a `just gen-models-data` step, gate it with the existing link checker (`check_browser_data_links.py --data app/models/data.js`), and add the file to the `git add` line.
2. Delete the staleness test, leaving a note pointing at the `app/data.js` precedent.
3. Add `app/models/data.js` to the "What NOT to commit" table in `CLAUDE.md`.

That does not by itself unblock the nine — each still carries the file — but once the test is gone, dropping that one path from each branch is safe and they all become mergeable.

## Follow-up, not fixed here

`app/discussions/data.js` is tracked but is regenerated by nobody: `generate-pages.yaml` does not run `gen-discussions-data` either. It has no staleness test, so it causes no conflicts — it just drifts silently. Worth folding into the same workflow step, but it is a different (harmless-today) bug and is left out to keep this change reviewable.

cmungall:
Closes #9804.

Nine approved PRs are currently deadlocked, each conflicting on `app/models/data.js` and on nothing else: #9104, #9119, #9122, #9140, #9142, #9145, #9263, #9324, #9325. This removes the cause.

## Why they conflict

`tests/test_models_export.py` asserted that the committed file equals what the exporter produces right now:

```python
assert committed == current, "app/models/data.js is stale — run `just gen-models-data`"
```

`app/models/data.js` is a generated browser index — 2,834 lines, 131 KB — rebuilt from **every** `computational_models` block in `kb/`. Any PR adding a model changes `current`, so the assertion made regenerating and committing the whole artifact mandatory for that PR. Two concurrent model PRs then cannot both merge, by construction. Each of the nine did exactly what the test demanded, in its own curation commit (`8789c8362d Add CML computational model`, `f6d350c954 Curate epilepsy computational models`, …).

Of the 22 open PRs that are approved + unassigned + non-draft, 13 are un-mergeable; these nine are the largest cluster, and the auto-merge sweep requires `mergeable == MERGEABLE`.

## The repo already settled this for the sibling index

A test of this shape was deliberately **not** written for `app/data.js` — `tests/test_check_browser_data_links.py:195`:

> Deliberately NOT tested here: the committed `app/data.js` against the committed `pages/`. Every curation PR adds a disorder YAML before the generate-pages workflow renders its page, so such a test would fail on the whole curation stream rather than on the drift it is meant to catch. **The invariant belongs to the generate-pages run (which renders first, then gates), not to repo state.**

`app/discussions/data.js` is the control: same kind of generated index, no staleness test, no conflicts. The models test was the anomaly.

It also red-lights `main` on a delay. The pytest job is path-gated (`src/**`, `tests/**/*.py`, `justfile`, `project.justfile`, `pyproject.toml`, `uv.lock`), so a kb-only PR never runs the test that its own merge invalidates. Staleness accumulates until some unrelated PR touches `src/` and goes red for something it didn't cause — which is what `d7de426b2 chore(models): regenerate app/models/data.js (fixes red main)` cleaned up. A curator hit exactly this on 2026-08-23 and reproduced it on `main` before concluding the file "is a derived artifact written only by the generate-pages workflow" (`history/disorders/Iron_Poisoning/2026-08-23T214320Z-claude-code-247a90.yaml`).

## What this changes

Deleting the test alone would leave the file to rot, because **generate-pages never regenerated it either** — it ran `gen-browser-data`, `gen-pathographs`, `gen-dashboard`, `gen-schema-docs`, and committed `pages/ app/data.js pathographs/ dashboard/ elements/`. So this moves ownership rather than dropping the constraint.

**`generate-pages.yaml`**
- runs `just gen-models-data` after rendering and all drift healing, and gates it with `just check-models-links`;
- commits the file alongside `app/data.js`;
- adds `kb/modules/*.yaml` and `models/**` to `on.push.paths`. `ModelsExporter.collect_records` reads module YAML (`models_export.py:231`) and `models/*.config.yaml` (`:107`), so without these a module-level computational model or a new perturb config reached the index only via the daily full rebuild — a hole in the ownership this PR claims. `kb/modules/immune_checkpoint_blockade.yaml` already carries a `computational_models` block, and `gen-pages` renders module pages via `render_all_modules`, so the pages the new gate checks were already this workflow's output while their source was not a trigger;
- moves the built-artifact list to a job-level `BUILT_PATHS` env var. Change-detection and `git add` are separate steps and cannot share a shell variable, and the list must not drift: a path missing from the first makes the run skip its PR, one missing from the second silently never updates. The comment records both why the list is centralized and why both consumers expand it **unquoted** — they need word splitting, `"$BUILT_PATHS"` would be a single nonexistent path that matches nothing, and a bash array cannot cross a step boundary.

**`project.justfile`** gains `check-models-links` next to `check-browser-links`, so a curator who hits the gate in CI can reproduce it with the command CI ran. Its `--data` form is added to the script's usage examples.

**`tests/test_models_export.py`** loses the staleness test, with a note pointing at the `app/data.js` precedent.

**`CLAUDE.md`** records `app/models/data.js` in "What NOT to commit".

## Verification

| Check | Result |
|---|---|
| `just check-models-links` | `OK: all 57 page_url targets in app/models/data.js exist on disk.` |
| Regeneration on a clean tree | no diff — idempotent, and `main` is currently in sync |
| Workflow parses | 15 trigger paths; `BUILT_PATHS` folds to one line, all six paths resolve |
| `ruff check .` | `All checks passed!` |
| Full local suite | **40,201 passed, 41 skipped** |
| `tests/test_models_export.py` | 8 passed (was 9; one removed) |
| `tests/test_check_browser_data_links.py` | 12 passed |
| `test_agent_config`, `test_ci_batch_validation`, `test_enum_cache`, `test_disclaimers`, `test_validate_disorder_hook` | pass |
| `yamllint -d relaxed` on the workflow | 58 warnings vs `main`'s 60, 0 errors |

The full suite also reported two failures in `tests/test_validate_disorder_hook.py`. Those are environment-only, not this branch's: `recipe_available()` shells out to `just`, which was not installed in the container I ran it in. Installing `just 1.58.0` and re-running gives 13/13. Stating it rather than quoting a clean number that wasn't quite clean.

No references to the deleted test remain outside an append-only history record, which is left untouched.

## What this does *not* do

It does not unblock the nine on its own — each still carries the file. Once this lands, dropping that one path from each branch is safe and they all become mergeable. I have not touched those branches; they were opened by other sessions.

## Follow-up, deliberately not included

`app/discussions/data.js` is tracked but regenerated by nobody — `generate-pages.yaml` does not run `gen-discussions-data` either. It has no staleness test, so it causes no conflicts; it just drifts silently. It belongs in the same workflow step and in `BUILT_PATHS`, but folding it in would widen a deadlock-clearing PR into an audit of all the browser indexes. Issue to follow once this lands.

### DIFF
--- a/.github/workflows/generate-pages.yaml
+++ b/.github/workflows/generate-pages.yaml
@@ -213,9 +213,10 @@ jobs:
           git config user.email "github-actions[bot]@users.noreply.github.com"
 
           # Park this build's pages in a throwaway commit so changing base cannot
-          # lose them. Only pages/ needs parking: app/data.js, pathographs/,
-          # dashboard/ and elements/ are all generated further down, and doing so
-          # AFTER this step is what makes them come from the fresh KB.
+          # lose them. Only pages/ needs parking: every other BUILT_PATHS entry
+          # -- app/data.js, app/models/data.js, pathographs/, dashboard/ and
+          # elements/ -- is generated further down, and doing so AFTER this step
+          # is what makes them come from the fresh KB.
           BASE="$(git rev-parse HEAD)"
           git add -A pages/
           git commit -q -m "wip: rendered pages" || echo "no page changes to park"
--- a/docs/page-build.md
+++ b/docs/page-build.md
@@ -131,8 +131,9 @@ Three mechanisms now close that gap:
    stale KB against its own matching stale pages, finds perfect agreement, and
    reports clean. Re-anchoring makes `kb/` current while keeping this build's
    rendered pages, which is exactly the mismatch the check is looking for. It
-   also puts the aggregates (`app/data.js`, dashboard, pathographs, schema docs —
-   all generated *after* this step, deliberately) on current data.
+   also puts the aggregates (`app/data.js`, `app/models/data.js`, dashboard,
+   pathographs, `elements/`, schema docs — all generated *after* this step,
+   deliberately) on current data.
 
    And it stops the regen PR conflicting. A branch built at a stale checkout
    collides with any later regen that touched the same page; that is what left


cmungall:
All three 🔵 suggestions taken in `7089747257`. This dismisses the approval — worth it, because suggestion 1 turned out to be a hole in this PR's own claim rather than a follow-up nit.

### 1 — `kb/modules/*.yaml` and `models/**` are now trigger paths

You're right, and I'd argue it's slightly worse than "a timing gap and not a correctness one". This PR asserts that generate-pages *owns* `app/models/data.js`. Ownership means the run fires whenever an input changes, and `ModelsExporter.collect_records` reads `kb/modules/` (`models_export.py:231`) and `models/*.config.yaml` (`:107`) — neither of which triggered it. `kb/modules/immune_checkpoint_blockade.yaml` already carries a `computational_models` block, so this isn't hypothetical.

There's a sharper version too: `gen-pages` renders module pages via `render_all_modules`, so the *pages* the new gate checks were already this workflow's output while their *source* was not a trigger. Adding the two paths closes both halves. Comment explains why they're there.

### 2 — `check-models-links` recipe

Added next to `check-browser-links`, and the workflow now calls it instead of inlining. Verified it runs: `OK: all 57 page_url targets in app/models/data.js exist on disk.` Also added the `--data app/models/data.js` form to the script's usage examples, next to `app/data.js` and `app/discussions/data.js` as you suggested.

### 3 — why the expansion is bare

Taken, and I extended it slightly past what you asked. The comment now says both *that* it is deliberate and *what breaks*: `"$BUILT_PATHS"` becomes a single nonexistent path that silently matches nothing — so quoting it would make change-detection see no changes and skip the PR, and `git add` stage nothing. It also records that a bash array is the usual fix and cannot cross a step boundary, so the next reader doesn't reach for one.

### On your check of the hard-failure risk

Thank you for chasing that specifically — a mid-publish hard gate that could fail for a reason the run didn't cause would have been the real defect here, and your `render_all_modules` / `gen-pages-changed` reasoning is the check I should have written into the body. It also happens to be exactly why suggestion 1 matters: the gate's inputs and the workflow's triggers were not the same set.

### Re-verified on `7089747257`

| Check | Result |
|---|---|
| `just check-models-links` | `OK: all 57 page_url targets in app/models/data.js exist on disk.` |
| Workflow parses | 15 trigger paths; `BUILT_PATHS` folds to one line, all six paths |
| `ruff check .` | `All checks passed!` |
| `check_browser_data_links` + `models_export` + `validate_disorder_hook` + `ci_batch_validation` | **35 passed** |
| `yamllint -d relaxed` | 58 warnings vs `main`'s 60, 0 errors |

Also worth recording: the full local suite on the previous head came back **40,201 passed, 41 skipped**, with two failures in `tests/test_validate_disorder_hook.py`. Those are environment-only — `recipe_available()` shells out to `just`, which isn't installed in this container. I installed `just 1.58.0` and re-ran: 13/13 pass. Not a finding, but I'd rather state it than let a green summary imply I ran a clean suite.

### Follow-up issue

Agreed on filing `app/discussions/data.js` rather than folding it in — I'll open it once this lands, so the issue can reference the pattern this PR establishes rather than describing it twice.

---

cmungall:
Both nits taken in `b2d2b51f5d`. This dismisses the approval again — last push, and I think nit 1 earns it.

### The re-anchor comment

Your framing is the reason I didn't leave it: that comment is what a reader consults to check which `BUILT_PATHS` entries survive the hard `git checkout -q -f -B`, and I'm the one who introduced `BUILT_PATHS`. An incomplete list is the wrong answer to the question it exists to answer.

Rather than append one filename I made it name the invariant, so it stays correct if the list grows again:

> Only `pages/` needs parking: every other `BUILT_PATHS` entry — `app/data.js`, `app/models/data.js`, `pathographs/`, `dashboard/` and `elements/` — is generated further down, and doing so AFTER this step is what makes them come from the fresh KB.

I verified that claim instead of inheriting it from the comment I was editing. Re-anchor is at line 237; `app/data.js` (303), `app/models/data.js` (319), `pathographs/` (325), `dashboard/` (330) and `elements/` (333, via `gen-schema-docs` — `project.justfile:1358` confirms that's where `elements/` comes from) are all after it. `pages/` renders at 176/180, *before* it, which is exactly why it's the one thing needing parking. So "only `pages/`" is right for a checkable reason, not just because it was already written down.

### `docs/page-build.md`

Same parenthetical, updated. I also added the `elements/` you noted was missing before this PR — it was a one-word fix in a line I was already editing, and leaving a known-incomplete list behind while fixing its twin seemed worse than the small scope bump.

### Verified on `b2d2b51f5d`

Workflow parses; `yamllint -d relaxed` 57 warnings vs `main`'s 60, 0 errors; `tests/test_disclaimers.py` 29 passed (it reads `docs/`). No behaviour change — both commits are comments.

### On the `app/discussions/data.js` follow-up

Agreed, and I'll file it once this lands so the issue can point at the pattern this PR establishes rather than re-deriving it. Same for the note about the nine: they each still need the file dropped, and those branches belong to other sessions, so that stays with @cmungall.

---

Issue #5121, opened 2026-07-01
The dismech-entry → MONDO-term relationship should be 1:1


matentzn:
The dismech entry to MONDO term relationship should be _1:1 in the forward direction_: each MONDO term should have at most one primary dismech Disease entry. Right now it isn't. The Mondo EMC export (#4697) surfaced **21 MONDO IDs that are each shared by 2 to 7 separate dismech entries**, 58 entries collapsing onto 21 terms.

This is the orthogonal complement to #3881. That issue asks the anchor-existence direction (should every dismech entry have a MONDO ID at all), and in passing states that the reverse Mondo to dismech direction is "obviously not 1:1" because dismech is "curated at a slightly higher level of disease abstraction, i.e. no subtypes". This issue is the forward direction, and it points at a contradiction: if the intent is "no subtypes", then many dismech entries pointing at one MONDO term is a violation of that stated principle, not an accepted exception. The offenders are almost all precision-oncology molecular subtypes and stage variants, which are exactly subtypes.

The practical consequence lands on the EMC pipeline. A non-1:1 mapping means Mondo receives multiple candidate definitions and PMID sets for the same term (7 of them for lung adenocarcinoma) and has to arbitrate which one wins, or Mondo's ingestion has to be defined to accept many-to-one.

Two ways to resolve it, and I don't want to prejudge which:

1. These subtype and stage entries get lumped into their parent entry, keeping the mapping 1:1 and honouring the #3881 "no subtypes" line.
2. dismech legitimately curates below MONDO granularity for molecular subtypes, in which case the "no subtypes" principle needs revising, and the EMC export plus any Mondo ingestion must be specified to tolerate many-dismech-to-one-MONDO (dedup, pick-most-specific, or per-row provenance).

## Examples

From `exports/mondo_emc.tsv` on `feat/mondo-emc-export`. Full totals: 1356 rows across 1325 unique MONDO IDs, 21 IDs duplicated, max multiplicity 7.

Molecular subtypes sharing a parent term (the bulk of it):

- `MONDO:0005061` lung adenocarcinoma x7: ALK, BRAF-V600E, EGFR, KRAS-G12C, MET-exon14-skipping, RET, ROS1 rearranged/mutant NSCLC
- `MONDO:0005575` colorectal cancer x4: BRAF-V600E, HER2+, MSI-High, Metastatic
- `MONDO:0005012` cutaneous melanoma x3: BRAF-V600, NRAS-mutant, Metastatic
[...]

PR #9808, opened 2026-08-28
Cancer granularity: ratify the lump/split ladder and fold the Metastatic_* entries into their parents

cmungall:
This PR settles how cancer entries are lumped and split. It adds a cancer-specific granularity rule to the design-decision register and enacts its clearest consequence: the ten `Metastatic_*` entries are no longer separate diseases — their content now lives in their histologic parent entries. It was prepared at @cmungall's direction in-session, following the review in `docs/reports/cancer-taxonomy-granularity-review-2026-08-28.md` (also added here).

## The decision (design decisions §3a)

The KB's cancer entries currently sit at eight different granularity levels, and the only rule on file (§3, "distinct MONDO identity and a substantially independent mechanism") was written for Mendelian disease. The new §3a follows how the field itself classifies tumors (WHO 5th edition, ICC, ICD-O, OncoTree): the backbone is site + histology, molecular alterations define an entity only where the field has promoted them (IDH-wildtype glioblastoma, NPM1-mutant AML), and everything below that is a stratum, not a disease. In short:

- The **default level for a new cancer entry is the histologic entity** (PDAC, SCLC, DLBCL) or the WHO/ICC-defined molecular entity.
- **Biomarker/therapy strata** (EGFR-mutant NSCLC, MSI-H CRC, TNBC) stay as separate entries only when they pass issue #7082's disjunction test — their own pathophysiology plus their own first-line therapy. The existing driver entries pass it and are kept; new ones must earn promotion. A promoted stratum without its own MONDO term anchors to the parent term with an explicit `skos:narrowMatch` mapping and an NTR filed, instead of silently reusing the parent ID as its `disease_term` (the source of the 21 duplicated MONDO anchors in issue #5121).
- **Stage is never a taxon**, and **pathways are never entries** (they stay in `kb/modules/` and groupings).
- Germline predisposition syndromes stay under the plain Mendelian rules.

CLAUDE.md and the `cancer-curator` skill are updated to match, per the register's governance steps.

## The enactment: ten `Metastatic_*` entries folded

Eight entries merge into their existing parents. Each parent gains a `stages:` block (Localized / Metastatic), the metastasis-specific mechanism nodes, and the folded entry's unique evidence, references, and datasets; evidence that duplicated the parent's was merged into the matching nodes rather than copied:

| Folded entry | Into |
|---|---|
| Metastatic_NSCLC | Non-Small_Cell_Lung_Cancer |
| Metastatic_HCC | Hepatocellular_Carcinoma |
| Metastatic_Renal_Cell_Carcinoma | Renal_Cell_Carcinoma (the VHL→HIF→VEGF chain, genetics, and the gut-microbiota knowledge gap all move) |
| Metastatic_Prostate_Cancer | Prostate_Adenocarcinoma (MONDO:0004956 kept as a narrowMatch mapping) |
| Metastatic_Gastric_Cancer | Gastric_Adenocarcinoma |
| Metastatic_Pancreatic_Adenocarcinoma | Pancreatic_Ductal_Adenocarcinoma (the 17-node RAS(ON)-inhibitor resistance/immune-reversal subgraph moves intact) |
| Metastatic_Ovarian_Cancer | Ovarian_High-Grade_Serous_Carcinoma |
| Metastatic_Colorectal_Cancer | Colon_Adenocarcinoma |

Two had no parent entry in the KB, so instead of deleting them they become the missing base rungs: **Metastatic_Melanoma → Cutaneous_Melanoma** and **Metastatic_Breast_Carcinoma → Breast_Carcinoma**, each with an exactMatch `disease_term`, a broadened description pointing at the existing driver/receptor stratum entries, and the metastatic view preserved as a stage.

Bookkeeping that moved with the content: each folded entry's history records were relocated to the parent's slug directory with `target.superseded_by` blocks; every parent gained a fold session record; the `kb/surrogate_endpoints` path, two comorbidity provenance notes, and two project frontmatter lists were updated.

## What was deliberately not done here

- No MONDO NTRs filed yet for the driver strata, and no new `Lung_Adenocarcinoma` parent — both are named as open follow-ups in §3a.
- The parent-`has_subtypes`-vs-split-file double representation (e.g. NSCLC listing "EGFR-mutant NSCLC" while the file exists) is untouched; the pointer-subtype convention is stated in §3a for follow-up.
- No content was re-graded or re-quoted except five `evidence_source` values in the ovarian file that the fold brought into conflict with the parent's existing grading of the same quoted sentences (harmonized to the parent's `OTHER`, both papers being narrative reviews).

## Validation

`just validate-disorders` over all ten changed entries passes (591/760 snippets verified against cache, remainder skipped by prefix; only pre-existing EGA/metabolomics accession fetch warnings). Whole-KB `check-duplicate-keys`, `check-entity-refs`, folded-hyphen/snippet-length/title-snippet/snippet-grading/environmental-evidence gates, and `validate-history-all` (7,161 records) all pass, as do 14,806 targeted structural pytest checks (unique names, subtype and entity-ref foreign keys, history layout).

Closes the cancer half of #306; enacts the bucket-B (stage entries) resolution and the bucket-A anchoring convention for #5121; adopts the promotion/demotion machinery of #7082.

https://claude.ai/code/session_01NtWCsS7rhR5AgiaYeSLwkN

---
_Generated by [Claude Code](https://claude.ai/code/session_01NtWCsS7rhR5AgiaYeSLwkN)_

### DIFF
--- a/history/disorders/Breast_Carcinoma/2026-08-28T155256Z-claude-code-426350.yaml
+++ b/history/disorders/Breast_Carcinoma/2026-08-28T155256Z-claude-code-426350.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Breast_Carcinoma
+  path: kb/disorders/Breast_Carcinoma.yaml
+session:
+  id: 2026-08-28T155256Z-claude-code-426350
+  timestamp: '2026-08-28T15:52:56Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - has_subtypes
+  - discussions
+  summary: Wire pathograph, add pointer subtypes and primary-tumour gap
+  details: |
+    PR review follow-up on the 3a fold. Connected the inherited metastasis nodes (EMT-driven dissemination -> organ-specific tropism -> angiogenic outgrowth and immune evasion); added pointer subtypes for the ER, HER2, TNBC and PIK3CA entries per section 3a(d); recorded an open KNOWLEDGE_GAP discussion for the absent primary-tumour mechanisms. ER and HER2 receptor heterogeneity is deliberately left unconnected: it describes biomarker discordance between sampled regions, not a step in disease progression.
--- a/history/disorders/Cutaneous_Melanoma/2026-08-28T155256Z-claude-code-56f15c.yaml
+++ b/history/disorders/Cutaneous_Melanoma/2026-08-28T155256Z-claude-code-56f15c.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Cutaneous_Melanoma
+  path: kb/disorders/Cutaneous_Melanoma.yaml
+session:
+  id: 2026-08-28T155256Z-claude-code-56f15c
+  timestamp: '2026-08-28T15:52:56Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - has_subtypes
+  - discussions
+  summary: Wire pathograph, add module conformance, pointer subtypes and primary-tumour
+    gap
+  details: |
+    PR review follow-up on the 3a fold. Connected the five inherited metastasis nodes into a chain (MAPK fitness -> invasive plasticity -> brain colonisation -> angiogenic support; MAPK fitness -> PD-L1 escape); added invasion_and_metastasis conformance on the EMT and colonisation nodes; added pointer subtypes for the BRAF, NRAS and KIT driver entries per section 3a(d); recorded an open KNOWLEDGE_GAP discussion for the absent primary-tumour mechanisms (UV mutagenesis, melanocyte transformation) rather than leaving the description overclaiming coverage.
--- a/history/disorders/Gastric_Adenocarcinoma/2026-08-28T155257Z-claude-code-24054f.yaml
+++ b/history/disorders/Gastric_Adenocarcinoma/2026-08-28T155257Z-claude-code-24054f.yaml
@@ -1,0 +1,24 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Gastric_Adenocarcinoma
+  path: kb/disorders/Gastric_Adenocarcinoma.yaml
+session:
+  id: 2026-08-28T155257Z-claude-code-24054f
+  timestamp: '2026-08-28T15:52:57Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Restore PMID:34650801 evidence on the diffuse-type peritoneal edge
+  details: |
+    PR review follow-up. The fold's node-level evidence merge missed two edge-level evidence items in the folded entry; re-attached both to the parent's Diffuse Infiltrative Growth -> Peritoneal Dissemination edge, which had carried no evidence.
--- a/history/disorders/Ovarian_High-Grade_Serous_Carcinoma/2026-08-28T155257Z-claude-code-717a7e.yaml
+++ b/history/disorders/Ovarian_High-Grade_Serous_Carcinoma/2026-08-28T155257Z-claude-code-717a7e.yaml
@@ -1,0 +1,24 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Ovarian_High-Grade_Serous_Carcinoma
+  path: kb/disorders/Ovarian_High-Grade_Serous_Carcinoma.yaml
+session:
+  id: 2026-08-28T155257Z-claude-code-717a7e
+  timestamp: '2026-08-28T15:52:57Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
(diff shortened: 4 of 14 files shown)


cmungall:
Both structural findings are fixed in `753f97a`, along with the three reference titles in `fcb4adc`. The CI failure was a scope check, not a content problem, and is handled with a label.

**Pathograph fragmentation.** Confirmed and fixed. Comparing each parent's orphan set against `origin/main` isolated exactly 15 nodes that this PR left with no causal edges: 6 in Colon, 5 in the ovarian entry, 2 in PDAC, 2 in RCC. They were orphans in the folded entries too, but a node with no edges is much more visible in a parent that has a working chain, so the fold is the right moment to wire them. Each new edge follows what the node's own description and evidence already say, for example the Colon chain: APC loss → WNT/RAS cooperation → EMT-linked dissemination → anoikis escape → hepatic colonisation → angiogenic remodelling and stromal support. Same for HGSC (proliferation → EMT → transcoelomic dissemination → mesothelial crosstalk, angiogenesis, immune evasion), PDAC (KRAS → early dissemination → perineural invasion), and RCC (checkpoint-sensitive microenvironment → immune evasion).

I also wired `Cutaneous_Melanoma` and `Breast_Carcinoma`, which you did not flag but which were fully disconnected — 5 nodes, no edges — because they were reconstituted from metastatic-only entries that never had a chain. Melanoma is now fully connected; breast has one node left, `ER and HER2 Receptor Heterogeneity`, which I deliberately did not attach: it describes biomarker discordance between sampled regions, not a step in disease progression, and inventing a causal edge for it would assert something the cited paper does not.

Zero PR-introduced orphans remain. The 19 still present are all pre-existing on main (6 in NSCLC, 1 each in HCC, RCC and prostate, plus the breast node above); fixing those is separate curation and outside this PR.

**Module conformance.** Also a fair hit — §3a asks metastatic-stage biology to conform to `invasion_and_metastasis` and only 3 of 10 parents did. I added `conforms_to` only where the node's GO annotations actually match the module node's, so the conformance claim is checkable rather than decorative: EMT activation on RCC, PDAC, HGSC, Colon, NSCLC, melanoma and prostate; circulatory survival/extravasation on RCC, Colon, NSCLC and melanoma; local invasion/intravasation on HCC. Ovarian gets the EMT node only, on purpose: ovarian cancer disseminates transcoelomically by shedding into ascites, so the module's circulatory step does not describe it, and claiming otherwise would be wrong.

**CI.** `test (3.13)` failed on the repo's curation-scope guard, which rejects any PR touching `kb/` alongside files outside `kb/`, `stubs/`, `history/`, `cache/`, `references_cache/`, `research/`, `pages/` and `dashboard/`. This PR necessarily does both — a design decision plus the KB change that enacts it — which is exactly the case the guard's `scope-override` label exists for, so I applied the label. I could not re-run the job myself (403 on rerun-failed-jobs), so the check re-runs on this push.

Validation after these changes: `just validate-disorders` passes on all nine edited entries, whole-KB `check-duplicate-keys` and `check-entity-refs` are clean, and 8,432 entity-reference and conformance tests pass.

---

cmungall:
All four blocking findings are now resolved, plus suggestions 5, 6, 7, 8, 10 and 11 and the two residual connectivity junctions. `f888bc5a`.

**Finding 4** — you were right that this is a defect the diff introduced, not a coverage gap. My path rewrite collided with an entry that was already there. Both `mapped_disease_files` lines are now one, the stale `Metastatic Prostate Cancer` name is gone from `mapped_diseases`, and the notes point at the Metastatic stage of `Prostate_Adenocarcinoma`.

**Finding 3** — fixed on all three. Each now carries a `skos:narrowMatch` mapping to the parent term with a justification saying why: MONDO has no term for the stratum, so `disease_term` anchors to the parent while the mapping records that the entry is narrower than it, and naming the entry that claims the parent term as an exactMatch. Filing the NTRs stays on the follow-up list.

**Residual connectivity.** Both junctions you named are wired. `VHL Inactivation → Renal Epithelial Malignancy` and `VEGF-Dependent Angiogenesis → Tumor Cell Invasion` for RCC; `KRAS Oncogene Activation` now heads both the RAS(ON) resistance and the immune-reversal subgraphs in PDAC. PDAC is back to a single component (25n/1c) and ovarian stays at 1c. RCC is 3c and breast 2c — the remainders are the pre-existing `Histologic and Molecular Heterogeneity` island and `ER and HER2 Receptor Heterogeneity`, which I left alone deliberately: receptor discordance between sampled regions is not a step in disease progression, and giving it a causal edge would assert something the cited paper does not say.

**Suggestion 10** — agreed and changed. `Brain Metastatic Colonization` is a colonisation node by name and by description, so it maps to `Metastatic Colonization`. I left NSCLC's `Brain Metastasis Seed-and-Soil Biology` on `Circulatory Survival and Extravasation` for the reason you gave: its description does lead with vascular arrest and BBB traversal.

**Suggestion 5** — restored. Both evidence items were on `downstream` edges in the folded entry, which is why my node-level merge missed them; they are now on the parent's diffuse-type → peritoneal edge, which previously carried none. Worth noting as a general lesson for future folds: edge-level evidence needs its own pass.

**Suggestion 6** — restored rather than documented as deliberate, since the narrowing was not deliberate. The treatment is now `PARP Inhibitor Maintenance` with niraparib and rucaparib alongside olaparib and the class-level HRD review as evidence. Nothing referenced the old `treatments#Olaparib Maintenance` name. That surfaced a second thing: the same review sentence was graded `OTHER` on the treatment and `COMPUTATIONAL` in the deep-research provenance block. It is a review article, so `OTHER` is right and the provenance copy is regraded.

**Suggestions 7 and 8** — both taken. Rather than soften the descriptions, each new parent now carries an open `KNOWLEDGE_GAP` discussion naming what is missing (UV mutagenesis and melanocyte transformation for melanoma; ductal transformation and DCIS-to-invasive progression for breast) and attached to `pathophysiology#` and the localized stage. Pointer subtypes are added for all seven promoted strata.

**Suggestion 9** left for follow-up — hallmark-module conformance beyond metastasis touches nodes this PR did not move, and I would rather not widen the diff further for it.

Validation at this head: `just validate-disorders` passes on all nine edited entries, whole-KB duplicate-key, entity-ref, snippet-grading and title-snippet gates are clean, `validate-history-all` passes, and 10,050 entity-reference, subtype-foreign-key, unique-name and history tests pass.

---

PR #9821, opened 2026-08-28
curate: ACTH-Independent Macronodular Adrenal Hyperplasia 3 (KDM1A, MONDO:0700299)

### LINKED ISSUE #9812: Curate ACTH-independent macronodular adrenal hyperplasia 3 (MONDO:0700299)
Curate a dismech entry for **ACTH-independent macronodular adrenal hyperplasia 3** ([MONDO:0700299](https://monarchinitiative.org/MONDO:0700299)) — germline loss-of-function of the histone demethylase *KDM1A* (hgnc:29079) with second-hit somatic 1p loss in adrenal nodules, causing ectopic GIP-receptor expression and the food-dependent (GIP-dependent) form of ACTH-independent hypercortisolism.

Preliminary read: this looks like a **DISEASE**. It is a single-gene entity with one conserved mechanism (KDM1A haploinsufficiency → derepression of *GIPR* in adrenocortical cells → meal-triggered cortisol secretion). It is mechanistically distinct from the entry the same gene already has, `KDM1A-Related_Neurodevelopmental_Disorder.yaml` (MONDO:0014751) — same gene, different disease — and distinct from `Primary_Pigmented_Nodular_Adrenocortical_Disease.yaml` (PRKAR1A/cAMP) and `Cushings_Syndrome.yaml` (the ACTH-dependent and general umbrella).

Stub: `stubs/ACTH-independent_Macronodular_Adrenal_Hyperplasia_3.yaml`

aleixpuigb:
New entry for the KDM1A form of primary bilateral macronodular adrenal hyperplasia — the molecular explanation for food-dependent Cushing syndrome, a clinical curiosity described for three decades before its cause was found.

## The mechanism

Two-hit inactivation of KDM1A: a germline truncating allele, plus somatic loss of the wild-type copy on 1p inside the adrenal nodule. KDM1A (LSD1) represses transcription by removing methyl marks from histone H3K4, so a KDM1A-null adrenocortical cell derepresses *GIPR* — the receptor for an incretin the gut releases after every meal. A receptor that should not be on adrenocortical cells at all now couples eating to cortisol synthesis.

That explains the two things that make the disease distinctive. Cortisol is *low* in the fasting morning state and high at midnight, which is why these patients were historically missed — morning cortisol is measured by convention and in this disease that is the trough. And the mechanism yields its own diagnostic test, a morning/midnight cortisol ratio.

The causal chain is modelled node by node: germline allele → somatic 1p loss → loss of H3K4 demethylase activity → ectopic GIPR → meal-triggered adenylate cyclase activation → inverted diurnal rhythm and bilateral macronodular hyperplasia → ACTH-independent hypercortisolism.

## What the deep-research run added

One openscientist run. Its reference validation came back clean: 15/15 references resolved, confabulation rate 0.0, 1/1 quoted claims valid, nothing off topic.

It contributed two things the first draft did not have:

- **Somatostatin-analogue therapy, as a negative result** (PMID:21264796). Suppressing GIP release is the obvious mechanism-directed move. Octreotide then pasireotide did acutely suppress the postprandial cortisol response — and the effect escaped, with no improvement in clinical or metabolic features. Recorded as `REFUTE`, with a note that the authors found the failure independent of receptor-subtype affinity, which removes the natural next move of trying a broader-spectrum analogue.
- **An extra-adrenal neoplasia association** (PMID:36857084): adrenal myelolipoma, MGUS and multiple myeloma. Recorded as `PARTIAL` with no frequency asserted, because it is a review's summary of reported cases with no denominator.

`just preflight-dr` flagged GIP as a "rival gene" at 44% of KDM1A mentions. I assessed it and judged it a false positive — GIP/GIPR is this disease's effector mechanism, not a second disease entity. Flagging it here because the heuristic can't distinguish those two cases and will do the same on any entry whose mechanism runs through a second gene.

## What is deliberately not claimed

Four `discussions` record the gaps rather than papering over them:

- The only comparative surgical outcome data in bilateral macronodular adrenal disease come from a cohort in which **no KDM1A variants were found**. Treatment guidance here is extrapolated from a group it was measured to be absent from, and the evidence item says so.
- The morning/midnight ratio's reported 100% sensitivity and specificity is a derivation-cohort figure from ten patients, with no independent validation.
- Every reported KDM1A carrier was female. No mechanism has been proposed and the cohorts are too small to separate penetrance, ascertainment and chance.
- The cAMP/PKA step is inferred from GIPR's canonical coupling. Disease-specific measurement stops at GIPR transcript on one side and plasma cortisol on the other.

## Scope boundary

KDM1A form only. ARMC5-related PBMAH is a separate molecular group without food dependence and is not curated here. `KDM1A-Related_Neurodevelopmental_Disorder` (MONDO:0014751) is the same gene and a different disease.

## Validation

`just validate` passes — schema, terms, and **30/30 snippets** verified against cached references. `just validate-disorders`, `just check-entity-refs` and `just check-duplicate-keys` all pass.

Term validation caught one real error before commit: I had used `hgnc:4270` (GIP) for GIPR, which is `hgnc:4271`.

Closes #9812

---


aleixpuigb:
Thanks — all four were right, and two of them were the kind of error I would not have found by re-reading my own diff. Fixed in 77cf87f.

| # | Finding | What I did |
|---|---|---|
| A | No `diagnosis:` section | Added four entries: mixed meal test with oral vs IV glucose, dexamethasone non-suppression, the cortisol-ratio screen, germline KDM1A/ARMC5 sequencing, adrenal CT/MRI |
| B | Urinary claim bound to a plasma term; frequency derived from a mean | Rebound to HP:0012030; dropped `VERY_FREQUENT`; added the normal-UFC counterexample |
| C | Low morning cortisol not recorded | Added with HP:0008163 and `temporality: DIURNAL` |
| D | ACTH neither phenotype nor cited | Added HP:0002920; fetched PMID:39465833 for a measured suppressed ACTH |

A few notes on how I handled them rather than just that I did.

**On B.** You were right twice over here and the second half is the part I want to flag. `VERY_FREQUENT` had been derived from "3.0-fold upper limit of normal" — a group *mean* in a ten-patient cohort. That is not a frequency at all, and turning it into a FrequencyEnum band was an arithmetic error wearing an annotation's clothes. I have dropped the band rather than substituting a different one, and put a note in the phenotype saying why no band is asserted. Your counterexample — PMID:21264796's patient with UFC within normal limits — is now cited on the same phenotype as `PARTIAL`, because a normal UFC demonstrably does not exclude the diagnosis.

**On C.** The `temporality: DIURNAL` is load-bearing rather than decorative. HP:0008163 on its own reads as hypocortisolism, which in a disease of cortisol *excess* is a clinically dangerous misreading, so there is a note saying the level is low only at the morning timepoint while total daily output is raised.

**On D.** This was the best catch. I asserted ACTH suppression in the pathophysiology node's prose and never noticed that no cited paper mentioned ACTH — the claim was doing definitional work with nothing under it. PMID:39465833 gives a measured suppressed ACTH with the expected raised cortisol and bilateral masses, so it is now evidenced. I graded it `PARTIAL` and said in the explanation that it is a PBMAH case report with no KDM1A genotype: the inference to KDM1A carriers is safe, since ACTH-independence is definitional for the parent entity and this is a subset of it, but it *is* an inference and the entry should not pretend otherwise. I also took your intra-adrenal ACTH nuance from PMID:42067271 — it is in the phenotype's `notes`, because "ACTH-independent" turns out to mean independent of *pituitary* ACTH rather than free of ACTH signalling, and that is worth a reader knowing.

**Non-blocking:** took pasireotide as a second `therapeutic_agent`. I have not added the dietary `environmental:` entry, adult onset, or steroidogenesis inhibitors in this round — happy to if you want them, but I did not want to widen a review-fix push with content that has not been through the same sourcing.

**Two errors of mine caught while doing this.** I guessed `NCIT:C79835` for pasireotide; term validation reported it as *Hsp90 Inhibitor XL888*. The correct binding is `CHEBI:72312`. That wrong CURIE had also been written into `cache/ncit/terms.csv` and the `chemicalentityterm` enum cache as a side effect of the failed validation run, so I reverted both files and re-derived them — the cache diff on this push is now a single additive row, `HP:0012030`. Separately, PMID:41864332 was fetched during the ACTH search, is not cited, and was removed rather than committed.

**On your closing note about HP:0003564 and HP:0031044** — thank you for checking those. For the record of what happened: I did not copy them because I did not use the deep-research report's suggested HPO terms at all, binding phenotypes from the cache and OAK instead. That is luck as much as discipline, and it is a good argument for the report's term suggestions being treated as leads needing verification rather than as bindings.

**Validation after fixes:** `just validate` passes — schema, terms, and **43/43 snippets** verified against cached references, up from 30/30. `just validate-disorders`, `just check-entity-refs` and `just check-duplicate-keys` all pass.

---

PR #9822, opened 2026-08-28
curate: Infantile Liver Failure Syndrome 1 (LARS1, MONDO:0024568)

### LINKED ISSUE #9813: Curate infantile liver failure syndrome 1 (MONDO:0024568)
Curate a dismech entry for **infantile liver failure syndrome 1** ([MONDO:0024568](https://monarchinitiative.org/MONDO:0024568)) — biallelic *LARS1* (hgnc:6512) cytoplasmic leucyl-tRNA synthetase deficiency causing recurrent, febrile-illness-triggered acute liver failure with hypoalbuminaemia, anaemia and developmental delay.

Preliminary read: this looks like a **DISEASE** — one gene, one conserved mechanism (impaired leucine charging of tRNA-Leu → reduced hepatic protein synthesis, with catabolic/febrile stress as the decompensation trigger). Its MONDO parent `infantile liver failure` is a grouping covering ILFS1 (LARS1), ILFS2 (NBAS) and ILFS3 (RINT1), which are separate diseases; this claim is for ILFS1 only.

Stub: `stubs/Infantile_Liver_Failure_Syndrome_1.yaml`

aleixpuigb:
New entry for LARS1 deficiency — recessive cytoplasmic leucyl-tRNA synthetase deficiency presenting as recurrent, febrile-illness-triggered acute liver failure with anaemia, hypoalbuminaemia and encephalopathic episodes.

## The mechanistic hinge

The residual enzyme is **temperature-sensitive**. Aminoacylation activity in patient fibroblasts is not merely low at baseline — it falls significantly further when the cells are warmed, measured at 38.5 and 40 °C specifically in a second, independent study. That single property converts a static enzyme deficiency into an episodic disease, and it is what explains the clinical rhythm: patients are relatively well between infections and decompensate during fever, when the crippled enzyme's capacity drops at the same moment protein demand rises. The liver, the body's highest-throughput secretory protein factory, fails first.

## Two things modelled against the obvious reading

**Encephalopathy is not placed downstream of the liver.** It occurred in one individual *after liver transplantation*, and a separate cohort reports encephalopathic episodes occurring independently of liver crises. The MRI signature — deep grey matter and brainstem change — is also reported as distinct from hepatic encephalopathy and from the metabolic stroke of organic acidurias. So it gets its own node, and an open question about what the direct mechanism is.

**The mTORC1/autophagy branch is kept separate and graded `PROVISIONAL`**, with a `HUMAN_MODEL_MISMATCH` discussion. The rescue experiment in larsb-knockout zebrafish is a genuine causal test — suppressing autophagy improved liver size and survival — but it is a *null* model with a progressive phenotype lethal within 12 days, standing in for a partial loss-of-function, episodic human disease survivable past 28 years. The mismatch matters practically, because the fish result points at autophagy and mTORC1 inhibitors as candidate therapies.

## A correction the deep-research run produced

Worth calling out, because the first draft had it wrong.

The draft cited the 2015 clinical series reporting that leucine supplementation "had no appreciable impact", and drew a mechanistic inference from that failure — that the limiting step is charging capacity rather than substrate. The openscientist run surfaced PMID:34194004 (Genet Med 2021), which reached the opposite conclusion by approaching it mechanistically: having first shown that aminoacylation falls at febrile temperature and that patient fibroblast growth is severely affected at low cognate amino acid concentration, the authors supplemented patients with their cognate amino acid **and intensified dosing during infections**, reporting benefit in growth, head circumference, development, coping with infections and oxygen dependency.

Both results are now recorded, along with the reading that reconciles them — the earlier attempt supplied leucine at steady state, the successful protocol raised supply precisely when capacity falls, so timing rather than substrate is what was got wrong. Stated as the better-supported of two open positions, not as settled: the 2021 series pooled four different ARS deficiencies and had no control arm, and the evidence explanation says so.

The same run also supplied a **36-patient cumulative series** (PMID:38844943) giving real frequency denominators — IUGR 31/32, hypoalbuminemia 32/32, microcytic anaemia 32/33, acute liver failure 24/34 — and Kaplan-Meier prognostic factors (onset under 3 months HR 12.29; liver failure HR 6.57; 12/36 died or were transplanted). These replaced frequency assertions the draft had made without denominators, and added a `progression` block. Note acute liver failure is 24/34, i.e. *not* universal, which is why it is `VERY_FREQUENT` rather than `OBLIGATE`.

Report reference validation: 15/15 resolved, confabulation rate 0.0, nothing off topic.

## An excluded reference

PMID:33314043 (Clin Genet 2021) is topically relevant, but its cached record has an **empty body** — `just fetch-reference` reported "Successfully cached" for a record with no retrievable abstract. No snippet can be verified against it, so it is not cited, and the exclusion is recorded in the entry's `notes` so a later curator does not spend the search again.

## Validation

`just validate` passes — schema, terms, and **51/51 snippets** verified against cached references. `just validate-disorders`, `just check-entity-refs` and `just check-duplicate-keys` all pass.

Closes #9813

---

### DIFF
--- a/history/disorders/Infantile_Liver_Failure_Syndrome_1/2026-08-28T111951Z-aleixpuigb-ad3ea0.yaml
+++ b/history/disorders/Infantile_Liver_Failure_Syndrome_1/2026-08-28T111951Z-aleixpuigb-ad3ea0.yaml
@@ -1,0 +1,51 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Infantile_Liver_Failure_Syndrome_1
+  path: kb/disorders/Infantile_Liver_Failure_Syndrome_1.yaml
+session:
+  id: 2026-08-28T111951Z-aleixpuigb-ad3ea0
+  timestamp: '2026-08-28T11:19:51Z'
+  actors:
+  - type: ai_agent
+    name: aleixpuigb
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9813
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9822
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - diagnosis
+  - phenotypes
+  - treatments
+  - pathophysiology
+  summary: 'Review round 1: fix snippet-length CI failure, frequency bands, orphan
+    node; add diagnosis, transplantation, three phenotypes'
+  details: |
+    Review round 1 on PR #9822 (ai4c-reviewer, CHANGES_REQUESTED). All seven findings accepted and fixed.
+
+    BLOCKING (1): just check-snippet-length failed on three sub-5-word snippet fragments I had introduced on PMID:38844943 - "acute liver failure (24/34)", "hypoalbuminemia (32/32)", "microcytic anemia (32/33)". This was a real CI failure caused by my diff, and it exposes a bad habit worth naming: I had quoted the minimum fragment carrying the number rather than the sentence containing it, which strips the propositional content the check exists to require. All three now quote the full denominator sentence, with each explanation naming the fraction relevant to its phenotype. The gate now passes.
+
+    (2) Two frequency bands were one band too high. Acute liver failure is 24/34 = 71% and seizures 10/13 = 77%; FrequencyEnum FREQUENT is 30-79%, so both are FREQUENT, not VERY_FREQUENT. The ALF explanation had reasoned about the band explicitly and still stopped one short - I was reading "not universal" as "just below universal". Fixed both.
+
+    (3) The encephalopathy pathophysiology node was an orphan with no edges. Added an incoming edge from Febrile Decompensation with causal_link_type INDIRECT_UNKNOWN_INTERMEDIATES. This preserves the entry's argument rather than undoing it: the edge runs from the shared TRIGGER, not from the liver, so the encephalopathy arm stays independent of hepatic failure while the graph now asserts what the prose already said.
+
+    (4) Added a diagnosis: section - the temperature-dependent fibroblast aminoacylation assay, which is unusually both the confirmatory test and the entry's central mechanism made visible, plus exome/genome sequencing with two interpretation warnings (LARS1 not LARS2; do not expect a null). Recorded in its notes that no GeneReviews chapter exists, re-confirmed by three esearch queries on 2026-08-28 as the reviewer asked.
+
+    (5) Added liver transplantation to treatments, with the mechanistic caveat that LARS1 deficiency is systemic - the post-transplant encephalopathy is cited as the evidence that transplantation treats the organ that fails first rather than the disease.
+
+    (6) PMID:34496286 was listed in references: but never cited - a real defect, since an uncited reference implies support it does not give. It now backs two new phenotypes, hepatomegaly (HP:0002240) and feeding difficulties (HP:0011968), both PARTIAL because the three-patient series spans LARS1 and MARS1 disease.
+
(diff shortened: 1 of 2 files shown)


aleixpuigb:
All seven fixed in 0aaa3e9. Thanks — the blocking one was a genuine CI failure I should have caught, and finding 2 is the kind of error that survives re-reading your own diff because the prose *sounds* like it did the arithmetic.

| # | Finding | What I did |
|---|---|---|
| 1 🔴 | `check-snippet-length` fails on three fragments | All three now quote the full denominator sentence; gate passes |
| 2 🟡 | Two frequency bands one too high | ALF 24/34 → `FREQUENT`; Seizures 10/13 → `FREQUENT` |
| 3 🟡 | Encephalopathy node is an orphan | Incoming edge from febrile decompensation, `INDIRECT_UNKNOWN_INTERMEDIATES` |
| 4 🟡 | No `diagnosis:` section | Added: temperature-dependent aminoacylation assay + sequencing |
| 5 🟡 | Liver transplantation missing | Added, with the systemic-disease caveat |
| 6 🟡 | `PMID:34496286` listed but never cited | Now backs hepatomegaly + feeding difficulties |
| 7 🟡 | Missing transaminase elevation | Added `HP:0002910` |

Three of these are worth more than a table row.

**On 1 — the habit, not just the failure.** I had quoted the minimum fragment carrying the number: `"hypoalbuminemia (32/32)"`. That is exactly the thing `check-snippet-length` exists to stop, because a bare fragment strips the propositional content and leaves a number floating free of what it counts. The fix you pointed at was better than a patch: the full sentence at `references_cache/PMID_38844943.md:41` carries *all* the denominators, so one properly quoted sentence now serves five phenotypes, and each `explanation` names which fraction it draws on. That reads better than three fragments would have even if they had passed.

**On 2.** You are right that the ALF explanation reasoned about the band and stopped one short. What actually happened is that I read "not universal" and reached for "just below universal" without doing 24/34. Both bands are corrected, and I checked every other frequency in the file against its denominator while I was there — hypoalbuminemia 32/32, microcytic anaemia 32/33, IUGR 31/32 and FTT 30/31 are all ≥97% and stay `VERY_FREQUENT`; hypotonia 13/27 = 48% and encephalopathy 8/13 = 62% were already `FREQUENT`.

**On 3 — with a note on what the edge does and does not say.** I've drawn it from the **febrile-decompensation trigger node**, not from the liver or from impaired protein synthesis. That matters: routing it under hepatic failure would undo the argument the entry is making, whereas running it from the shared trigger asserts what the prose already says — infection precipitates both arms — while leaving the neurological arm independent of liver failure, which is what the post-transplant episode demonstrates. `INDIRECT_UNKNOWN_INTERMEDIATES` because the imaging authors say plainly that the pathomechanism remains unclear, and the edge description says so.

**On 6.** Agreed, and I'd put it more strongly than "unused": a reference listed in `references:` but cited nowhere implies support it is not actually giving. It now backs hepatomegaly and feeding difficulties, both `PARTIAL` — the three-patient series spans LARS1 *and* MARS1 disease, so it reports features shared across two conditions rather than ILFS1 frequencies.

**GeneReviews — confirmed, as you asked.** Three `esearch` queries against PubMed today: `LARS1 GeneReviews[All Fields]`, `infantile liver failure syndrome GeneReviews[Title]`, `leucyl-tRNA synthetase GeneReviews[All Fields]`. All zero results. Your reasoning was right — the infantile-liver-failure chapter covers *NBAS*/ILFS2. That is now recorded in the `diagnosis` entry's `notes` with the date, so the next curator does not repeat the search.

**Non-blocking taken:** `GO:0038202` TORC1 signaling replaces `GO:0031929` — agreed, more precise for a claim specifically about mTORC1.

**Non-blocking not taken this round:** the `environmental:` entry for the febrile trigger, and coagulopathy/hyperammonemia. Both are reasonable, but neither is sourced in the currently cached references, and I did not want to widen a review-fix push with content that has not been through the same sourcing. Happy to do them as a follow-up if you'd rather they were in.

**Validation after fixes:** **60/60 snippets** verified (was 51/51). `just check-snippet-length` and `just check-snippet-grading` both pass against `origin/main`. Schema, terms, `validate-disorders`, entity-refs and duplicate-key checks all pass.

---

PR #9823, opened 2026-08-28
curate: Calvarial Doughnut Lesions-Bone Fragility Syndrome (SGMS2, MONDO:0007470)

### LINKED ISSUE #9814: Curate calvarial doughnut lesions-bone fragility syndrome (MONDO:0007470)
Curate a dismech entry for **calvarial doughnut lesions-bone fragility syndrome** ([MONDO:0007470](https://monarchinitiative.org/MONDO:0007470)) — dominant *SGMS2* (hgnc:28395) sphingomyelin synthase 2 variants causing childhood-onset osteoporosis with fractures plus the characteristic ring-shaped ("doughnut") sclerotic calvarial lesions, with or without spondylometaphyseal dysplasia.

Preliminary read: this looks like a **DISEASE** with a clean allelic split worth modelling as `has_subtypes` rather than separate entries: the *SGMS2* p.Arg50\* truncating allele gives the milder calvarial-doughnut/osteoporosis phenotype, while frameshift alleles escaping nonsense-mediated decay give the severe neonatal spondylometaphyseal dysplasia form — one gene, one sphingomyelin-at-the-plasma-membrane mechanism, two severity poles.

Stub: `stubs/Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome.yaml`

aleixpuigb:
New entry for SGMS2-related CDL — childhood-onset osteoporosis with recurrent spinal and peripheral fractures, alongside sclerotic ring-shaped ("doughnut") lesions in the cranial bones. Described radiologically by Keats and Holt in 1969; the gene was found in 2019.

## The allele-severity inversion

This is the entry's central point, and it is worth stating up front because it runs against reflex.

The recurrent nonsense allele **p.Arg50\*** produces a catalytically dead enzyme and the **milder** phenotype — childhood osteoporosis with normal growth. The missense alleles **p.Ile62Ser** and **p.Met64Arg** produce a **fully active** enzyme that cannot leave the ER, and actually *increase* de novo sphingomyelin production — yet they cause neonatal fractures, severe short stature and spondylometaphyseal dysplasia.

Anyone reasoning from a loss-of-function default gets the genotype–phenotype prediction backwards here, so the description says so explicitly rather than leaving it to be inferred from the subtype blocks. What the two allele classes share is **mislocalisation away from the plasma membrane** — the missense alleles retained in the ER, p.Arg50\* dislocated into the cis/medial-Golgi. Location, not quantity, is where the causal work happens.

## Two nodes shaped by negative results

**Osteoclast function is normal.** Formation, morphology and resorptive capacity were all measured and found normal, and histomorphometry showed normal turnover — which excludes the high-resorption model that would otherwise be the default reading of an osteoporosis phenotype. This is modelled as an explicit node, because it sits in real tension with what patients actually receive: bisphosphonates are antiresorptive. That tension is recorded as a `KNOWLEDGE_GAP` rather than smoothed over, and it has teeth — a drug that further slows mineralization is at least conceivable in a disease whose lesion *is* a prolonged mineralization lag time.

**The calvarial doughnut lesions get no incoming causal edge.** A disease that under-mineralises the skeleton produces focal *hyper*-sclerosis in the calvaria, and no mechanism has been proposed for that inversion. Drawing an edge from the mineralization branch would assert a mechanism that does not exist in the literature, so the node stands unconnected with the gap recorded.

## What the deep-research run added

One openscientist run; reference validation 10/10 resolved, confabulation rate 0.0, nothing off topic. It contributed materially:

- **The definitive mechanistic source** (PMID:36102623, Sokoya 2022): the severe variants retain full enzymatic activity and fail to leave the ER because of a *defective autonomous ER export signal*, with ER sphingomyelin accumulation and disrupted transbilayer asymmetry reproduced in patient-derived fibroblasts. A new node (Disrupted Secretory-Pathway Lipid Landscape) was built on it — with the authors' final step quoted as the postulate it is, graded `PARTIAL`, since they postulate rather than demonstrate the link to bone-forming activity.
- **The mouse result** (PMID:31847800): Sms2-deficient mice have **no** bone-formation deficit; it requires additional osteoblast-specific Sms1 deletion. Recorded as an `animal_models` entry with `relationship: FAILS_TO_RECAPITULATE` and a `HUMAN_MODEL_MISMATCH` discussion. There is a second reading in that discussion that I think is more useful than "mouse isoform divergence": since the severe human alleles are ER-retained *active* enzyme, a mouse null may model the wrong lesion entirely, and the informative experiment is a missense knock-in — which has not been reported.
- Characterisation of the cranial nerve palsies as transient, self-resolving and recurrent (PMID:37886644), which is the temporal profile that fits a lipid or conduction mechanism better than progressive foraminal compression.

PMID:40393762 — a paediatric pamidronate outcome the report cited with figures — was fetched but its cached record has an **empty body** (Chinese-language case report, no English abstract), so it could not be cited and is not used.

## An ontology gap

HPO has no term for a calvarial doughnut lesion. `HP:0004493` Craniofacial hyperostosis is used as the closest accurate ancestor rather than manufacturing a narrower match — but that means the *ring morphology*, which is the discriminating feature of the whole entity, is not captured by the binding. It is carried by `preferred_term` and a note. A term request to HPO would be the right fix; happy to file one if maintainers agree.

## Validation

`just validate` passes — schema, terms, and **42/42 snippets** verified against cached references. `just validate-disorders`, `just check-entity-refs` and `just check-duplicate-keys` all pass.

Reference validation caught one real error before commit: a snippet I had attributed to PMID:40978119 actually appears in PMID:34761145.

Closes #9814

---

### DIFF
--- a/history/disorders/Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome/2026-08-28T114843Z-aleixpuigb-32cb31.yaml
+++ b/history/disorders/Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome/2026-08-28T114843Z-aleixpuigb-32cb31.yaml
@@ -1,0 +1,48 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome
+  path: kb/disorders/Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome.yaml
+session:
+  id: 2026-08-28T114843Z-aleixpuigb-32cb31
+  timestamp: '2026-08-28T11:48:43Z'
+  actors:
+  - type: ai_agent
+    name: aleixpuigb
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9814
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9823
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - treatments
+  - phenotypes
+  - diagnosis
+  summary: 'Review round 2: split bundled treatment so modality and target_mechanisms
+    match the prose; add replicated glaucoma family'
+  details: |
+    Review round 2 on PR #9823 (ai4c-reviewer, CHANGES_REQUESTED). One blocking item, correct and worth recording carefully because of what kind of error it is.
+
+    BLOCKING - the "Anabolic and Other Antiresorptive Agents" entry bundled teriparatide, denosumab and risedronate under one treatment with therapeutic_modality: PEPTIDE and one target_mechanisms edge to Defective Bone Matrix Mineralization.
+
+    The failure mode is specific: my PROSE was right and my STRUCTURED SLOTS said the opposite. The description carefully argued that teriparatide is anabolic and therefore better matched to this disease's mechanism than the antiresorptives are - and then the machine-readable link asserted that the whole entry, denosumab included, targets the mineralization node. In the pathograph and in KGX export, denosumab would be drawn as targeting the arm the entry says it does not touch. The single-valued modality slot was also simply wrong for two of the three agents, and the entry NAME called teriparatide an antiresorptive.
+
+    Split into three: Teriparatide (PEPTIDE, CHEBI:135983, keeps the target_mechanisms link, with the edge description now saying explicitly that the link is asserted for teriparatide alone); Denosumab (MONOCLONAL_ANTIBODY, NCIT:C61313, deliberately NO target_mechanisms link, with the description stating why); and risedronate (CHEBI:8869) moved into the existing Bisphosphonate Therapy entry where it belongs pharmacologically. The reviewer correctly noted that the multivalued therapeutic_agent pattern does not apply here - these are alternatives tried in different patients, not a named combination regimen.
+
+    Non-blocking suggestions taken, all sourced from references already cited in this PR:
+
+    Glaucoma has REPLICATED and my note said it had not. PMID:37886644 - already cited in this entry - reports an independent Filipino family carrying the same recurrent p.Arg50* allele with juvenile open-angle glaucoma. Added as a SUPPORT evidence item and the note rewritten. Two unrelated families with the same allele and the same extra-skeletal finding is a materially stronger claim, and it strengthens the entry's argument that sphingomyelin metabolism matters outside bone.
+
+    The long-bone-deformity binding (HP:0006487 Bowing of the long bones) was narrower than its snippet, which said "deformities" generically. Added a bowing-specific quote from PMID:30779713 and noted the relationship between the two items.
+
+    Bound diagnosis_term on two of three diagnosis entries: NCIT:C48789 Dual X-ray Absorptiometry and NCIT:C15709 Genetic Testing. The skull-radiography entry is deliberately left unbound with a note - NCIT:C38103 is Chest Radiography specifically, and per the terms contract no term beats a term that is nearly right.
+
+    The reviewer also retracted one of its own round-1 findings: the SMPD3/RANKL tension is already handled, since the Preserved Osteoclast Function node states the tension in its own evidence item rather than ignoring it.
+
+    Validation after fixes: 60/60 snippets verified (up from 56/56); check-snippet-length, check-snippet-grading, entity-refs, duplicate-keys and validate-disorders all pass. Cache diff is additive only.
--- a/kb/disorders/Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome.yaml
+++ b/kb/disorders/Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome.yaml
@@ -512,6 +512,11 @@ pathophysiology:
       better than progressive foraminal compression.
 diagnosis:
 - name: Skull radiography for calvarial doughnut lesions
+  notes: >-
+    `diagnosis_term` is deliberately left unbound. NCIT has no plain skull- or
+    general-radiography clinical-action term that accurately covers this;
+    NCIT:C38103 is Chest Radiography specifically. Per the terms contract, no
+    term beats a term that is nearly right.
   description: >-
     The defining diagnostic modality. Ring-shaped sclerotic lesions on skull
     radiographs are what distinguishes CDL from other primary osteoporoses and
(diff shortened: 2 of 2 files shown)


aleixpuigb:
All six fixed in 2b568cc. Your diagnosis of the problem is the useful part: this was under-consumption of cached reference *full text* that the PR itself had committed. No wrong claims here, but glaucoma, scoliosis, sensorineural hearing loss, long-bone deformity, five treatments and two lab findings were all sitting in files already in the diff. I mined those references for the mechanism and stopped.

| # | Finding | Fix |
|---|---|---|
| 1 | No `diagnosis:` | Skull radiography, densitometry/spinal imaging, SGMS2 sequencing |
| 2 | No `biochemical:` | ALP + 25(OH)D, `PARTIAL` |
| 3 | Missing phenotypes | Glaucoma, scoliosis, SNHL, long-bone deformity |
| 4 | Two terms under-specific | `HP:0100255`→`HP:0002657`, `HP:0004322`→`HP:0003510` |
| 5 | Missing treatments | Teriparatide/denosumab/risedronate, vitamin D, orthopedic surgery |
| 6 | Osteocyte node orphaned | Incoming edge from the lipid-landscape node |

Three things worth more than a row.

**On 4 — you were right that this was indefensible rather than merely imprecise.** The subtype is *named* `CDL-SMD` and its description says spondylometaphyseal dysplasia, while the phenotype bound the parent term `Metaphyseal dysplasia`. That is not a judgement call about specificity; the entry contradicted itself. Both terms upgraded.

**On 5 — teriparatide is the one, and you spotted why.** It is an *anabolic* agent. The demonstrated lesion in this disease is defective matrix mineralization on a normal-turnover background with **measured-normal osteoclast function** — so an agent that drives formation is arguably better matched to the mechanism than the antiresorptives patients actually receive. I have extended `cdl_bisphosphonate_mechanism_mismatch` with exactly that: the anabolic-versus-antiresorptive comparison is the obvious next question, it has never been made, and the mechanism predicts which way it should go. That is a better knowledge gap than the one I wrote originally, and it came from your review rather than from me.

All three new treatment entries are graded `PARTIAL` on purpose. The source sentence reports *exposure* — "she received multiple types osteoporosis treatment including; Teriparatide, Denosumab and Risedronate" — with no outcome attached, and the same patient had sustained about 40 fractures by then. It establishes what was tried and nothing about efficacy, and the explanations say so.

**On 6 — and thank you for distinguishing this from the other two orphans.** That distinction is exactly right and I want to confirm I have not disturbed them: the calvarial-doughnut node and the preserved-osteoclast node stay disconnected deliberately, each with a matching knowledge gap, because connecting them would assert mechanisms nobody has proposed. The osteocyte node was different — it asserted a plasma-membrane lipid mechanism *in its own description* while sitting unconnected, which is a real inconsistency. It now has an incoming edge from the lipid-landscape node with `INDIRECT_UNKNOWN_INTERMEDIATES`, and the edge description states plainly that this is a coherence argument about what membranes are needed for, not a measured pathway.

**Two smaller things I added rather than just filling gaps.** The skull-radiography entry carries the caveats that follow from natural history — the lesions are variably penetrant and can develop over time, so a normal infant skull film does not exclude the diagnosis. And the sensorineural hearing note records that a *sensorineural* rather than conductive deficit argues against simple compression by sclerotic temporal bone, which would otherwise be the obvious skeletal explanation.

**Not taken this round:** `mappings:`, subtype `disease_term`s, per-allele `variants` entries, and the SMPD3/RANKL tension with the preserved-osteoclast node. The last of those is genuinely interesting and I'd rather give it a proper look than bolt it onto a review-fix push.

**Validation:** **56/56 snippets** verified, up from 42/42. `check-snippet-length` and `check-snippet-grading` pass against `origin/main`. Schema, terms, `validate-disorders`, entity-refs and duplicate-key checks all pass.

---

aleixpuigb:
Fixed in 129cc2f. The blocking item is a good catch and worth naming precisely, because of what kind of error it is: **my prose was right and my structured slots said the opposite.**

The description argued carefully that teriparatide is anabolic and therefore better matched to this disease's mechanism than the antiresorptives are — and then a single `target_mechanisms` edge asserted that the whole bundled entry, denosumab included, targets the mineralization node. Your point about KGX export is the one that lands: denosumab would have been *drawn* as targeting the arm the entry says it does not touch. A reader of the prose gets the right answer; a consumer of the graph gets the opposite one. That is worse than a plain error, because the careful writing hides it.

Split into three, each with its own accurate modality:

| Entry | Modality | `target_mechanisms` |
|---|---|---|
| Teriparatide | `PEPTIDE` | → Defective Bone Matrix Mineralization |
| Denosumab | `MONOCLONAL_ANTIBODY` | none, deliberately |
| Risedronate | folded into `Bisphosphonate Therapy` | inherits that entry's |

The teriparatide edge description now says explicitly that the link is asserted for teriparatide alone, and the denosumab description says why it carries none — so the next curator doesn't "tidy" one back in. Agreed on the regimen point: these are alternatives tried in different patients over a lifetime, not a co-administered protocol, so the multivalued `therapeutic_agent` pattern would have been the wrong shape.

**Three suggestions taken**, all from references already in the PR:

- **Glaucoma had replicated and my note said it hadn't.** PMID:37886644 — which this entry already cites — reports an independent Filipino family carrying the same recurrent p.Arg50\* allele with juvenile open-angle glaucoma. That is now a `SUPPORT` evidence item and the note is rewritten. It also does real work for the entry: two unrelated families, same allele, same extra-skeletal finding is a much better basis for the "sphingomyelin matters outside bone" argument than one family was.
- **The long-bone binding was narrower than its snippet.** You're right that it was the snippet choice, not the term — `HP:0006487` is defensible and "lower limb bowing" was sitting in the cache. Added as a second evidence item, with a note pointing at the relationship between the two.
- **`diagnosis_term` bound on two of three** — `NCIT:C48789` and `NCIT:C15709`. The skull-radiography entry is left unbound *with a note explaining why*, since `NCIT:C38103` is Chest Radiography specifically and I'd rather record the absence than reach for a nearly-right term.

**Thank you for retracting the RANKL finding.** You're right that it was already handled — `Preserved Osteoclast Function` states the tension inside its own evidence item rather than ignoring it, which is the modelling I'd want. Worth noting because a reviewer withdrawing its own finding on re-reading is exactly what stops a fix cycle from drifting into churn.

**Still open, non-blocking:** `mappings:`, subtype `disease_term`s, per-allele `variants` for the severity inversion, the SMPD3/phosphocholine bridge, cortical-vs-trabecular, `onset_category`, and the other transient cranial nerve palsies. I'd rather do those as a considered follow-up than keep widening a fix push.

**Validation:** **60/60 snippets** verified, up from 56/56. `check-snippet-length`, `check-snippet-grading`, `check-entity-refs`, `check-duplicate-keys` and `validate-disorders` all pass; cache diff is additive only.

---

PR #9824, opened 2026-08-28
curate: Immunodeficiency 92 (REL / c-Rel deficiency, MONDO:0030498)

### LINKED ISSUE #9815: Curate immunodeficiency 92 (MONDO:0030498)
Curate a dismech entry for **immunodeficiency 92** ([MONDO:0030498](https://monarchinitiative.org/MONDO:0030498)) — biallelic *REL* (hgnc:9954) deficiency of the NF-kB subunit c-Rel, presenting as a combined immunodeficiency with impaired dendritic-cell IL-12 production, defective T-cell responses and susceptibility to mycobacterial, viral and opportunistic infection.

Preliminary read: this looks like a **DISEASE** — one gene, one conserved mechanism (loss of c-Rel-dependent transcription in the IL-12/IFN-gamma axis). Once curated it is a candidate member of the existing `kb/groupings/Inborn_Errors_of_Immunity.yaml` grouping.

Stub: `stubs/Immunodeficiency_92.yaml`

aleixpuigb:
New entry for c-Rel deficiency — autosomal recessive combined immunodeficiency caused by biallelic loss of one NF-kB subunit while the others stay intact.

## Why the mechanism is unusually legible

c-Rel binds the promoters of a coherent set of cytokine genes — IL-2, IFN-γ, IL-12, IL-21, IL-23 — and the patients' infectious susceptibility reads almost directly off that list: *M. tuberculosis*, *Salmonella*, *Cryptosporidium*, CMV, all controlled by the IL-12/IFN-γ axis. The B-cell arm matches the IL-21 half: reduced IgG, undetectable IgA, a compartment arrested at the naive stage (85% naive vs 47–77% normal; 0.3% switched memory vs 10.0–30.4%), and non-protective tetanus and diphtheria titres despite boosters.

The defect is **graded, not absolute** — live BCG, polio and measles were tolerated without sequelae, which is not the SCID profile, yet the same patient developed *M. tuberculosis* osteomyelitis at seven. That dissociation is recorded as an open question, because "are live vaccines safe here?" cannot be answered from one patient who happened to tolerate them.

## A myeloid arm the label does not advertise

IL-12 and IL-23 production is **abolished in cDC1s and monocytes but preserved in cDC2s**, and CD86 induction on conventional dendritic cells fails — so c-Rel is required for antigen-presenting function, not only for lymphocyte cytokine output. That gets its own pathophysiology node. The cDC1-versus-cDC2 selectivity is the sharpest result in the disease and nothing explains it: both subsets express c-Rel and both make IL-12 family cytokines, so a simple promoter-binding account predicts both should fail.

## The correction this PR is really about

Worth flagging plainly, because it is the largest error caught in this batch.

The draft was built from the two patients my own PubMed searches found, and it recorded a `KNOWLEDGE_GAP` asking whether IL-12, IFN-γ or IL-21 transcription had ever been measured in c-Rel-deficient **patient** cells — with a proposed experiment to go and do it.

The openscientist run surfaced PMID:34623332 (J Clin Invest 2021), the second patient's paper, in which exactly that work had already been done, thoroughly. So the entry would have shipped asking for an experiment that was published five years ago.

What changed as a result:

- The transcriptional node moved from `PROVISIONAL` to `ESTABLISHED`, with direct measurements attached instead of an inference from prior immunology. Its `notes` say plainly that the draft grading was wrong and why.
- The myeloid/cDC1 node was added — the draft had missed that arm entirely.
- The B-cell node gained its intrinsic targets (impaired MYC and BCL2L1 induction blocking plasmablast differentiation), which is a different and more specific claim than the IL-21 route modelled upstream.
- A new phenotype records the *selective* reduction across NK, Temra CD8, memory CD4/Th1, Treg and memory B compartments with other subsets normal.
- The incorrect gap was replaced by a real one: the cDC1/cDC2 selectivity.

Report reference validation: 13/13 resolved, confabulation rate 0.0, nothing off topic.

## Evidence base — please read the entry with this in mind

Three patients have ever been reported; two in curatable detail. `evidence_source: HUMAN_CLINICAL` here means *one patient*, and for PMID:34623332 it means one patient with an extensive mechanistic workup — the strongest evidence in the entry despite n=1. The schema has no slot that separates a case report from a 300-patient cohort, so the distinction is carried in prose in every `explanation`, and a `discussions` entry attached to the disease as a whole says so.

Craniosynostosis, language delay and epilepsy, reported only in the third patient, are recorded as `PARTIAL` with an explicit note that a second unrelated recessive disorder in a consanguineous family is a live alternative explanation. No `frequency:` is asserted anywhere it would be a statistical fiction.

## Animal models

Added the classical c-Rel knockouts (PMID:7649478, PMID:8622948), including the IL-2 rescue readout that dissects the lineage-specific requirement — IL-2 restores T-cell but not B-cell proliferation. Also recorded, as `PARTIAL` because it *complicates* the entry rather than supporting it: LPS-stimulated Rel−/− macrophages **overproduce** GM-CSF, so c-Rel represses as well as activates depending on cell type, which the transcriptional node does not model.

## Validation

`just validate` passes — schema, terms, and **37/37 snippets** verified against cached references. `just validate-disorders`, `just check-entity-refs` and `just check-duplicate-keys` all pass.

Term validation caught one error before commit: `HP:0030374`'s canonical label is "Decreased memory B cell proportion", not the synonym OAK's prefix search had matched.

Closes #9815

---

### DIFF
--- a/history/disorders/Immunodeficiency_92/2026-08-28T112907Z-aleixpuigb-5fc9c4.yaml
+++ b/history/disorders/Immunodeficiency_92/2026-08-28T112907Z-aleixpuigb-5fc9c4.yaml
@@ -1,0 +1,52 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Immunodeficiency_92
+  path: kb/disorders/Immunodeficiency_92.yaml
+session:
+  id: 2026-08-28T112907Z-aleixpuigb-5fc9c4
+  timestamp: '2026-08-28T11:29:07Z'
+  actors:
+  - type: ai_agent
+    name: aleixpuigb
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9815
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9824
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  - genetic
+  - phenotypes
+  - treatments
+  - discussions
+  summary: 'Review round 1: correct three factual errors from unread full text (allele
+    count, BCG tolerance, HSCT outcome)'
+  details: |
+    Review round 1 on PR #9824 (ai4c-reviewer, CHANGES_REQUESTED). Three CRITICAL findings, all correct, all caused by the same process error.
+
+    The root cause, stated plainly because it is the most useful thing in this record: I read the ABSTRACT of PMID:34623332 and treated it as the paper. The cache file contains 1497 lines of full text. The deep-research report had surfaced the paper and I correctly used it to fix the transcriptional node - and then stopped, never reading the clinical or genetic sections. Three of the entry's factual claims were contradicted by text sitting in the PR's own cache.
+
+    CRITICAL 1 - allele count. The entry said two REL alleles are published and framed the whole genetic section around "two alleles reaching the same endpoint by different routes". There are three. Patient 2 carries c.395-1G>A in the essential ACCEPTOR splice site of exon 5, common to both known REL isoforms, producing a frameshift and a premature stop at codon 134. It is the mirror of the index allele (donor side of the same exon, in-frame 18-residue deletion), and the contrast is more informative than the original framing was. Fixed in both the pathophysiology node description and genetic[REL].notes, with two new evidence items.
+
+    CRITICAL 2 - live vaccines. The entry asserted in four places that live BCG, poliovirus and measles vaccines were tolerated without sequelae, generalising from the index patient. Patient 2 developed DISSEMINATED BCG disease from the vaccine, with bone, lung and lymph node involvement. This was the most clinically consequential error in the batch: an entry a clinician might consult on live-vaccine safety presented only the reassuring case. Rewritten as a two-patient dissociation across the description, the pathophysiology node, the evidence explanations and the discussion. The discussion is renamed imd92_bcg_dissociation and reframed around the observed variance, which is a stronger argument for the graded-defect thesis than the original single-patient inference. Added a BCGosis phenotype (HP:0020087). The entry now states explicitly that nothing supports treating live vaccines as safe in this disease.
+
+    CRITICAL 3 - HSCT. The entry said no transplanted IMD92 patient has been published and there is no outcome evidence whatsoever. Patient 2 was transplanted in 2017 and her infectious phenotypes were cured. Corrected, and the cure is now recorded for what it is beyond a treatment outcome: because HSCT replaces only the haematopoietic compartment, simultaneous correction of the immunological and infectious phenotypes is a rescue result establishing the disease as leukocyte-intrinsic - direct support for the whole pathograph.
+
+    IMPORTANT 4 - patient 2's clinical phenotype was essentially unrepresented. Added four phenotypes: disseminated BCG disease, chronic mucocutaneous candidiasis, severe CMV infection, recurrent viral infections. The CMC entry follows the reviewer's mechanistic thread - CMC is the canonical IL-17-axis phenotype and the paper measures poor Th17 cytokine production, so the fungal susceptibility runs down the same promoter list as the mycobacterial one. Also added the antigen-specific proliferation failure (tuberculin, candidin, tetanus toxoid, CMV antigens) to the T-cell node, which had rested only on the index patient's PHA result.
+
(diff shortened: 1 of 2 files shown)


aleixpuigb:
All three criticals and all four importants fixed in dfbc6f6. This was a good review and I want to name the root cause rather than just list the patches, because all three criticals have the same one.

**I read the abstract of PMID:34623332 and treated it as the paper.** The cache file in this PR is 1497 lines of full text. The deep-research run surfaced that paper, I used it to correct the transcriptional node — and then stopped. Three of the entry's factual claims were contradicted by text the PR itself had already committed. Your framing is exactly right: this is asking for the same pass over the clinical and genetic halves that had already been made over the immunology.

| # | Finding | Fix |
|---|---|---|
| 1 🔴 | Three alleles, not two | `c.395-1G>A` added to node description, `genetic` notes, + 2 evidence items |
| 2 🔴 | BCG tolerance generalised from patient 1 | Rewritten in all four places; `HP:0020087` BCGosis phenotype; discussion reframed |
| 3 🔴 | HSCT cure exists | Description corrected; cure added as `SUPPORT` |
| 4 🟡 | Patient 2 unrepresented | 4 phenotypes + antigen-specific proliferation on the T-cell node |
| 5 🟡 | Cholangitis has replicated | `PARTIAL` → `SUPPORT`, second patient's evidence added |
| 6 🟡 | `VERY_FREQUENT` breaks own policy | Band dropped, with a note |
| 7 🟡 | `OTHER` on IVIG | → `PROTEIN_REPLACEMENT`, and treatment split in two |

**On 1.** The three-allele spectrum is a better story than the two-allele one, and your donor/acceptor observation is what makes it: same exon, opposite splice boundary, different consequence class — in-frame 18-residue deletion versus frameshift at codon 134 — both ending at absent protein. The entry now makes that contrast explicitly rather than just counting to three.

**On 2 — the one that mattered.** You were right to call this the most clinically consequential thing in the review, and right that the *discussion's framing* was the worst part of it. Asking "can we know live vaccines are safe from one patient who tolerated them?" while a disseminated-BCG case sat in the cited paper is worse than silence, because it implies the evidence base is one-sided in the reassuring direction when it is one-sided in the other. It is now `imd92_bcg_dissociation`, built around the two-patient divergence, and the entry states plainly that nothing here supports treating live vaccines as safe in this disease. You are also right that this *strengthens* the graded-defect thesis — the grading now has observed variance rather than being inferred from a single case, and the discussion lists what could explain it (differing residual protein function given the two allele classes, individual NF-kB redundancy, host modifiers, BCG strain or dose) and says two patients cannot separate them.

**On 3.** Took your suggestion to quote the line-1228 sentence rather than the line-192 one — and not only for the hyphenation. The line-1228 sentence *is* the mechanistic claim: correcting immunological and infectious phenotypes together by replacing only the haematopoietic compartment is rescue evidence that the disease is leukocyte-intrinsic. That is support for the whole pathograph, not just for the treatment, and the entry now says so in the treatment description and in the evidence-base discussion.

**On 4.** Added BCGosis, CMC, severe CMV and recurrent viral infections, and followed the Th17 thread you flagged — the CMC entry now connects the fungal susceptibility to the measured poor Th17 cytokine production, so the promoter list does not stop at IL-23 without following it to the output that explains the candidiasis. Also added the antigen-specific proliferation failure to the T-cell node, which had rested on the index patient's PHA result alone.

**On 5.** Agreed, and I'd add that this is the right instinct applied to the wrong fact, as you put it — the note was cautious in a direction the evidence did not support. Two of three unrelated patients with cryptosporidial cholangitis, same organism, is now `SUPPORT`.

**A note on snippets from this cache.** The full text is PDF-extracted and hyphenates across line breaks (`chol-`/`angitis`, `hematopoiet -`/`ic`, `stim-`/`uli`). I chose snippets that avoid crossing those; where I could not, the snippet stops short and the `explanation` says why — the antigen-proliferation one ends before "stimuli" and names the four antigens in prose instead. Multi-line quotes *without* hyphenation validate fine, so the constraint is narrower than it first looks.

**GeneReviews:** confirmed by query from this environment — `REL GeneReviews[All Fields]` and related searches return nothing. Your judgement call was correct; a 2019-described disorder with three patients is outside GeneReviews' scope.

**Not taken this round:** `mappings:`, per-allele `variants` entries, and the SMPD3/RANKL point. All reasonable; I'd rather do them as a separate change than widen a correction push.

**Validation:** **51/51 snippets** verified, up from 37/37. `check-snippet-length`, `check-snippet-grading` and `check-title-snippets` all pass against `origin/main`. Schema, terms, `validate-disorders`, entity-refs and duplicate-key checks pass.

---

Issue #9687, opened 2026-08-27
`just refresh-orphadata` is broken: pinned Orphadata checksums are ~6 months stale


matentzn:
## What happens

`just refresh-orphadata` fails on the first file:

```
RuntimeError: checksum mismatch for en_product1.xml after download:
  got      df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb
  expected fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

There is no override flag on `refresh` (only `--force`, which re-downloads and re-verifies), so the Orphanet source is currently unbuildable from a clean checkout.

## It's a new upstream release, not a bad download

I checked, because "checksum mismatch" usually means a truncated transfer:

- **Two independent downloads produce the identical checksum** (`df8d562a…`), so the transfer is stable.
- The file is **larger** than the pin records — 54,026,799 bytes vs `size_bytes: 53331679` — which is growth, not truncation.
- The XML declares its own release date: `<JDBOR date="2026-06-23 07:53:50">`, against `snapshot_date: "2025-12-09"` in the manifest.

Orphadata publishes rolling updates at a stable URL, so the pin has simply aged out by about six months.

## Why this is worth more than a checksum bump

Three separate reviews on curation PRs this session ([#9666](https://github.com/monarch-initiative/dismech/pull/9666), [#9669](https://github.com/monarch-initiative/dismech/pull/9669), and by implication [#9667](https://github.com/monarch-initiative/dismech/pull/9667)) told me to consult Orphanet, each correctly pointing out that `reference: ORPHA:` is a heavily used convention — 7,418 evidence items across `kb/disorders/` — and each giving the same two-command recipe:

```bash
just refresh-orphadata
just structured-rebuild-orphanet --id <code>
```

That recipe cannot currently succeed. Orphanet phenotype rows (`| HP:0002616 | Aortic root aneurysm | Very frequent (99-80%) |`) are exactly the HPO-term-plus-frequency evidence a rare-disease entry needs, so this quietly blocks a whole class of curation work — and the failure mode is a checksum error that reads like a local problem rather than a stale pin.

## Why I did not just fix it in my PR

[...]

PR #9828, opened 2026-08-28
curate: Nasu-Hakola disease (PLOSL, MONDO:0009092) — one receptor complex, two organs

### LINKED ISSUE #9816: Curate polycystic lipomembranous osteodysplasia with sclerosing leukoencephaly (MONDO:0009092)
Curate a dismech entry for **polycystic lipomembranous osteodysplasia with sclerosing leukoencephaly** ([MONDO:0009092](https://monarchinitiative.org/MONDO:0009092)) — Nasu-Hakola disease, an autosomal recessive disorder in which loss of the TREM2/TYROBP (DAP12) receptor–adaptor signalling pair produces the unusual combination of cystic bone lesions with pathological fractures and an early-onset frontal-lobe dementia with sclerosing leukoencephalopathy.

MONDO records two descendants — PLOSL1 (MONDO:0020749, TYROBP) and PLOSL2 (MONDO:0020750, TREM2). These are two genes in the same receptor–adaptor complex converging on one clinical syndrome with one conserved pathograph, so the working plan is a single Disease entry with `has_subtypes` for PLOSL1 and PLOSL2 rather than a grouping. TREM2 hypomorphic variants are also an established Alzheimer's disease risk factor, which is a useful cross-link for the pathograph but a distinct entity.

Stub: `stubs/Polycystic_Lipomembranous_Osteodysplasia_With_Sclerosing_Leukoencephaly.yaml`

matentzn:
New entry for Nasu-Hakola disease (polycystic lipomembranous osteodysplasia with sclerosing leukoencephalopathy).

Closes #9816

## The lump/split call

MONDO records two descendants — PLOSL1 (TYROBP) and PLOSL2 (TREM2). I curated this as **one Disease with two `has_subtypes`**, not a grouping and not two entries, because TREM2 and DAP12 are the ligand-binding and signal-transducing halves of the *same* receptor. The 2002 paper that found the second gene said so directly: it is "an interesting example of how mutations in two different subunits of a multisubunit receptor complex result in an identical human disease phenotype." One pathograph, two entry points.

Both subtypes carry a bound `subtype_term`, which also means `tidy-stubs` will now see this concept as covered.

**The mouse is linked three different ways, on purpose.** The DAP12-null mouse gets `RECAPITULATES` for the osteoclast node, `PARTIALLY_RECAPITULATES` for myelination, and `FAILS_TO_RECAPITULATE` for the microglial node — with a `HUMAN_MODEL_MISMATCH` discussion attached.

That last one is a real negative result, not missing data. Single-nucleus sequencing of DAP12-deficient human brain found microglia not silenced but locked into a RUNX1/STAT3/TGF-beta wound-healing program; the same study profiled DAP12-null mice and found "very mild microglial defects that did not recapitulate NHD." It matters because the therapeutic idea currently in play — resetting microglial signalling — comes from the human signature, while the tractable model is the mouse that does not show it.

The osteoclast link also carries a direction mismatch worth noting: mice get osteopetrosis (uniformly increased bone mass), patients get focal lytic cysts with fragility. The cellular mechanism transfers, the tissue outcome inverts. That is recorded as a `limitations` note and an `INCREASED` bone-mass readout graded `PARTIAL`, rather than being quietly omitted.

**Inheritance is recessive with a caveat I did not want to lose.** TYROBP heterozygotes may have cyst-like bone lesions without dementia, and carriers of the Finnish deletion have raised Alzheimer risk. That evidence item is graded `PARTIAL` — it supports the recessive claim while documenting a carrier effect a strict recessive reading would miss.

**One term is bound deliberately loosely.** Anti-seizure medication is recommended by GeneReviews as a class with no drug named, so `therapeutic_agent` is left empty with a `notes` line saying why, rather than binding a representative CHEBI compound that would assert a specificity the source does not have.

## Sources and validation

Perplexity `sonar-deep-research` was the deep-research provider (report committed under `research/`). Its own reference-validation block reported 4/4 identifiers resolved with zero confabulation — but only 4 of its 16 citations were resolvable identifiers at all, the rest being web URLs to OMIM, Orphanet, Radiopaedia and Alzforum. So the report was used as a lead source and **every** evidence snippet here is anchored to a PMID I fetched and verified separately.

That mattered once: I reached for PMID:10888878 as the Paloneva 2000 TYROBP paper from memory and `just fetch-reference` returned a paper about parkin. The correct identifier is PMID:10888890. The wrongly cached file was removed rather than left in the tree.

GeneReviews PMID:20301376 was used as the phenotype and management baseline per the curation skill.

- `just validate` — schema clean, **46/46 snippets verified** against cached references
- `just validate-disorders` — passed
- `just check-entity-refs`, `just check-duplicate-keys` — passed

Only the six PMIDs this entry cites are included in `references_cache/`, and the `cache/` rows are limited to CURIEs this entry introduced (a CHEBI row from the abandoned anticonvulsant binding was reverted).

---

### DIFF
--- a/history/disorders/Nasu-Hakola_Disease/2026-08-28T121804Z-claude-code-81fbe2.yaml
+++ b/history/disorders/Nasu-Hakola_Disease/2026-08-28T121804Z-claude-code-81fbe2.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Nasu-Hakola_Disease
+  path: kb/disorders/Nasu-Hakola_Disease.yaml
+session:
+  id: 2026-08-28T121804Z-claude-code-81fbe2
+  timestamp: '2026-08-28T12:18:04Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - "9816"
+  prs:
+  - "9828"
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - notes
+  summary: 'Review round 3: correct myoclonus CURIE in notes, soften frequency attribution'
+  details: |
+    Addresses the round-3 ai4c-reviewer review on PR #9828. Corrected HP:0002067 to HP:0001336 in the top-level notes block: HP:0002067 is Bradykinesia, Myoclonus is HP:0001336. This is the mislabel I had myself identified in the round-1 response as an error in the deep-research report, and then committed anyway when transcribing the reviewer's phenotype list into prose. The reviewer's reason for blocking on a free-text CURIE is worth recording: the notes block is a worklist that tells a future curator these features become citable once issue #9687 is fixed, and free-text CURIEs in notes and description are the one place just validate-terms cannot reach, so a wrong one there outlives a wrong one in a term block. Also took the reviewer's two non-blocking suggestions: softened the sentence so it no longer asserts an Orphanet very-frequent band that could not be fetched, saying instead that the features reached the entry via a deep-research report and a review comment rather than from a frequency row anybody read; and moved the notes block from between has_subtypes and inheritance to directly after references, so a scope note sits near the head of the file. Revalidated: 60/60 snippets verified, check-entity-refs and check-duplicate-keys pass.
--- a/kb/disorders/Nasu-Hakola_Disease.yaml
+++ b/kb/disorders/Nasu-Hakola_Disease.yaml
@@ -51,6 +51,35 @@ references:
   title: "Polycystic Lipomembranous Osteodysplasia with Sclerosing Leukoencephalopathy."
   tags:
   - GeneReviews
+notes: >-
+  Phenotype scope, and why the list is shorter than Orphanet's. Six further features are
+  reported for this disease and not modelled here - spasticity (HP:0001257), myoclonus
+  (HP:0001336), reduced bone mineral density (HP:0004349), limitation of joint mobility
+  (HP:0001376), skeletal dysplasia (HP:0002652) and irritability (HP:0000737). They reached
+  this entry by way of a deep-research report and a review comment rather than from an
+  Orphanet frequency row anybody read, so the "very frequent" banding they are usually
+  quoted with is not asserted here - the band is exactly what could not be fetched. None of
+  them is modelled, and the reason is availability of evidence rather than a judgement that
+  they are out of scope.
+
+  Orphanet's frequency rows are not currently citable in this repository. `just
+  refresh-orphadata` fails its pinned checksum (issue #9687), and because it aborts on the
+  first bulk file the phenotype product carrying the HPO rows is never downloaded at all. A
+  further hazard found while trying: the rejected download is left on disk and
+  `just structured-rebuild-orphanet` will then build a normal-looking cache file from it, so
+  an ORPHA_2770.md produced today would be derived from unverified data. One was built during
+  curation and deliberately discarded rather than committed.
+
+  Two candidate phenotypes were dropped for the same reason after being drafted. Memory
+  impairment had no cached source using memory or any cognitive-domain wording - only
+  "progressive dementia", which the Frontal Lobe Dementia entry already carries. Arthralgia
+  rested on the same GeneReviews sentence as Bone Pain; that sentence names pain and
+  tenderness at anatomical sites and does not separate joint from bone pain, so splitting it
+  across two HP terms would have asserted a distinction the source does not make.
+
+  When #9687 is resolved these features become straightforwardly citable from the Orphanet
+  frequency rows, and adding them is a small, self-contained follow-up.
+
 has_subtypes:
   - name: PLOSL1
     display_name: PLOSL1 (TYROBP/DAP12)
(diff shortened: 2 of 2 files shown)


matentzn:
Thanks — the coverage read was right, and the three IMPORTANT items are now addressed in bf5d2aff6. Two of the suggestions I did not action, with reasons.

## Done

**`progression`** — the four GeneReviews stages (latent; osseous, 3rd decade; early neurologic, 4th decade; late neurologic), each quoted. You were right that this is the load-bearing clinical fact and its absence was the worst gap. The osseous-stage entry also carries the detail that fractures are diagnosed several years after the pain starts, which is the first of this disease's two diagnostic delays.

**`histopathology`** — three entries covering both halves of the name: the sclerosing leukoencephalopathy with frontal accentuation and microglial activation from the autopsy series, the atrophy/myelin-loss/gliosis triad, and the polycystic osseous lesions. The microglial-activation entry notes that it sat as a static histological observation for two decades before single-nucleus sequencing said what those microglia were doing, which ties the histopathology to the pathophysiology rather than leaving it as a parallel list.

**Six phenotypes** — memory impairment, personality change, loss of mobility (`HP:0001288`), ventriculomegaly (`HP:0002119`), cerebral cortical atrophy, arthralgia. Personality change is modeled separately from disinhibition because GeneReviews treats them differently: the personality change is the onset event of the neurologic stage, disinhibition one component of the resulting syndrome.

**Node splits (suggestions 1 and 2)** — both were fair. `Impaired Oligodendrocyte Myelination and Demyelination` is now two nodes; failure to lay down myelin and loss of myelin already made are different claims with different evidence, and the edge between them is drawn `INDIRECT_UNKNOWN_INTERMEDIATES` because which one drives the other is not established. `Osteoclast Differentiation and Resorption Failure` is now `Osteoclast Maturation Arrest` → `Impaired Bone Resorption and Remodelling`, which also lets the mouse `RECAPITULATES` link attach to the node it actually measured.

**Evidence regrading (suggestion 3)** — both moved to `OTHER`, with the explanation saying why: they are background statements of receptor biology, not findings of the studies reporting them.

**Anchor (suggestion 5)** — repointed to `animal_models#DAP12-deficient mouse (Tyrobp knockout)`.

62/62 snippets verified; `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading` and `validate-disorders` all pass.

## Not done, and why

**Orphanet frequency rows — the recipe does not currently work, and fails in a way worth knowing about.** I ran `just refresh-orphadata` and it died on the pinned checksum for `en_product1.xml` (this is issue #9687, plus five older duplicates). Because it aborted on the *first* file, `en_product4.xml` — the one that actually carries the HPO rows with frequency bands you wanted me to cite — was never downloaded at all.

The part that isn't in that issue, and that I've now [added as a comment](https://github.com/monarch-initiative/dismech/issues/9687#issuecomment-5451957738): the rejected download is **left on disk**, and `just structured-rebuild-orphanet --id 2770` then succeeds against it and writes a normal-looking `ORPHA_2770.md`. I very nearly committed a cache file built from data that had just failed its integrity gate; the only reason I caught it was that the missing `## Phenotypes` section sent me back to look at what had actually downloaded. I deleted both the cache file and the unverified XML.

So the Orphanet-sourced frequencies aren't citable right now. The six phenotypes I added are sourced to GeneReviews and PMID:11402114 instead, and where a frequency is an inference rather than a quoted band the `description` says so.

**`mappings` (suggestion 4)** — there is nowhere to put those. `DiseaseMappings` has exactly four slots: `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings`. No OMIM, no ORPHA. A `mondo_mapping` would just restate the entry's own `disease_term`, and the ICD codes I have are from the ORPHA record I've just discarded as unverified, so adding them would be citing data I decided not to trust.

That gap seems worth raising separately — OMIM and ORPHA are the two identifiers rare-disease curators reach for first, and 7,418 evidence items in `kb/` already use `reference: ORPHA:`, so the identifiers are clearly in scope for the KB even though the mappings block can't hold them. I'd rather not open that as an issue from inside this PR; happy to file it if you think it's real.

**Myoclonus** — skipped deliberately. `HP:0002067` is *Bradykinesia*, not Myoclonus; the deep-research report mislabeled it, same as it did `HGNC:17762` for TREM2. Myoclonus is `HP:0001336`, but I have no cached source stating it for this disease, so I left it out rather than bind a term to an unsourced claim.

The remaining Orphanet "very frequent" items (limitation of joint mobility, reduced bone mineral density, abnormal adipose tissue morphology, spasticity) are likewise real but currently unciteable from this PR's sources. Rather than leave them silently absent as you asked, they're worth a line here: they are omitted for want of a source, not because they were judged out of scope, and the Orphanet fix in #9687 is what unblocks them.

---

matentzn:
All four items done in 158c63f72, plus the three suggestions.

You were right about the failure mode, and it's worth naming precisely because it slipped past every check in the stack: all three snippets **verified as exact substrings**, so `count-verified-snippets` and `validate-disorders` were both green while the quote said "progressive dementia" and the HP term said memory impairment. Exactness is not aboutness, and nothing here tests the second.

I'd also point at where the pressure came from: your round-1 list gave me a count to hit, and when the Orphanet rows turned out to be unfetchable I reached for the nearest already-verified sentence instead of concluding the phenotype wasn't supportable. That's on me rather than on the review — but it's a fair warning that "the research surfaced ~20 and you have 8" is a prompt that can be satisfied the wrong way.

## Changes

**Memory Impairment — dropped.** I grepped all six cached sources for "memory" and any cognitive-domain wording; nothing. `Frontal Lobe Dementia` already covers the ground.

**Arthralgia — dropped.** Same GeneReviews sentence as `Bone Pain`, and that sentence names pain and tenderness at anatomical sites without distinguishing joint from bone. One sentence, two HP terms, only one of them said.

**Cerebral Cortical Atrophy → `HP:0002059` Cerebral Atrophy.** The description now states why the binding sits at the cerebral level: the citable PMID says cerebral, and the frontal-accentuated cortical wording is web-only.

**Personality/disinhibition split finished.** `Personality Change and Disinhibition` is now just `Disinhibition`, requoted against the frontal-lobe-syndrome clause that actually names loss of social inhibitions, with a description saying what separates the two — personality change is the onset event, disinhibition a feature of the syndrome it becomes. The older node keeps the insidious-onset sentence, which is what it was always supporting.

**Top-level `notes:` added.** It names the six Orphanet-listed features not modeled (spasticity, myoclonus, reduced BMD, limitation of joint mobility, skeletal dysplasia, irritability), says the reason is evidence availability rather than scope, records #9687 including the leave-the-bad-file-on-disk behaviour, and records the two dropped phenotypes with the reasoning above — so the next curator doesn't re-add them from the same sentences.

## Suggestions

1. `cell_types: osteoclast` added to `Impaired Bone Resorption and Remodelling`. Agreed — resorption is osteoclast-mediated and the split shouldn't have moved the cell type off it.
2. **Ventriculomegaly kept.** The bicaudate ratio is a standard ventricular-width index — it is what "abnormally high and progressively increasing bicaudate ratios" measures, not an inference layered on top of it. I'd distinguish this from the three you flagged: those needed a *different concept* than the quote named, this needs only the definition of the metric quoted. Happy to drop it if you read that differently.
3. Explanation added on all three. Your reading is the one I'd defend too, and now it says so: that sentence describes human neuropathology in a paper that analysed human brain tissue, which is what separates it from the receptor-biology sentences regraded to `OTHER`.

Phenotypes went 14 → 12 net. 60/60 snippets verified; `validate`, `check-entity-refs`, `check-duplicate-keys` pass. No new cache rows — `HP:0002059` and `CL:0000092` were already there.

Thanks for withdrawing suggestion 4 and the Orphanet-only part of round-1 item 1 explicitly. Recording that in the review rather than just dropping it saved me from re-litigating both.

---

PR #9829, opened 2026-08-28
curate: methylcobalamin deficiency type cblG (MTR, MONDO:0009609)

### LINKED ISSUE #9820: Curate methylcobalamin deficiency type cblG (MONDO:0009609)
Curate a dismech entry for **methylcobalamin deficiency type cblG** ([MONDO:0009609](https://monarchinitiative.org/MONDO:0009609)) — an autosomal recessive defect of MTR (methionine synthase) causing isolated homocystinuria with megaloblastic anemia, hypomethioninemia and, in severe cases, developmental delay and cerebral atrophy, without methylmalonic aciduria.

The KB precedent points to a standalone Disease entry: cblC (`MMACHC-related_Methylmalonic_Aciduria_and_Homocystinuria_cblC_Type.yaml`), cblD (`MMADHC-related_Disorder_of_Cobalamin_Metabolism_cblD_Type.yaml`) and cblE (`Methylcobalamin_Deficiency_Type_cblE.yaml`) each have one. cblG appears today only as an unbound `has_subtypes` line in `Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport.yaml`, which is how cblE looked before it was curated. cblE (MTRR, the reductive activase) and cblG (MTR, the enzyme itself) are the paired halves of the same reaction, so the cblE entry is the closest model for structure — but the two are distinct genes and distinct diseases, and the entry should say what separates them rather than mirroring cblE.

Stub: `stubs/Methylcobalamin_Deficiency_Type_cblG.yaml`

matentzn:
New entry for methylcobalamin deficiency type cblG (methionine synthase / MTR deficiency).

Closes #9820

## Why a standalone entry

cblC, cblD and cblE each have one, and cblG appeared until now only as an unbound `has_subtypes` line in `Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport.yaml` — which is exactly how cblE looked before it was curated. This follows that precedent.

It also completes a matched pair worth having: **cblE loses MTRR**, the flavoprotein that reactivates methionine synthase after its cobalamin cofactor oxidises, so the enzyme is present but stranded inactive. **cblG loses the enzyme itself.** Two lesions on opposite sides of one reaction converging on one biochemical phenotype. The entry says what separates them rather than mirroring the cblE text.

## Two entity-attribution hazards, handled explicitly

Both are recorded in the entry's `notes` so a later curator doesn't have to rediscover them.

**GeneReviews CLINICAL CHARACTERISTICS was deliberately not mined.** PMID:20301503 covers all the intracellular cobalamin disorders, and its clinical paragraph names cblC as "the prototype and best understood phenotype" — the age-of-onset and manifestation lists are cblC's. Mining them into cblG would be Named Entity Confusion. Only Diagnosis/Testing and Management were used. This mirrors the decision already recorded in the cblE entry.

**The main clinical cohort pools two diseases.** PMID:25526710 reports 11 cblE *and* 13 cblG patients and gives its symptom frequencies, imaging findings and outcome numbers for the pooled group. Those figures are the best available and are used — but **every evidence item drawing on them says in its `explanation` that the figure is pooled and is not a cblG-specific rate**, and cblG-specific sources are preferred wherever they exist.

This one is worth flagging beyond this PR: the schema has no way to express "this evidence comes from a cohort of this disease plus a related one." I put it in prose because there is nowhere else to put it, which means it is only as durable as the next person reading the explanation.

## The knowledge gap is the point of the entry

The literature genuinely disagrees on whether treatment changes the neurological natural history, and I did not pick a side.

The 24-patient cohort found the biochemical response significant and the course improving or stable in 96% — and in the same abstract reported CNS symptoms rising significantly over time with 16 of 23 developmentally delayed or severely handicapped at follow-up, concluding patients "show limited clinical response to treatment." The cblG-specific series reports the opposite for its five patients and argues cblG belongs in newborn screening. Treatment timing is the obvious candidate explanation, but five selected patients in a screening-advocacy paper versus a retrospective survey with pooled data cannot settle it — and the second paper's own authors say the literature disagrees.

That is curated as an open `KNOWLEDGE_GAP` with evidence on both sides (one `SUPPORT`, one `REFUTE`, one `PARTIAL` where the authors state the disagreement themselves) and a proposed experiment that would test the timing hypothesis in a cblG-only cohort.

## Things I got wrong on the first pass

Recorded because they say something about where the checks earn their keep:

- **A snippet was a paper title.** I quoted the title of PMID:22422209 as evidence. The snippet checker caught it; replaced with an actual finding — that remethylation defects are associated with leukodystrophy while cobalamin *transport* defects generally are not, which is the more useful claim anyway.
- **Three ontology bindings were wrong.** `CL:0038024` does not exist (→ `CL:0000038`). `CHEBI:15636` is methylenetetrahydrofolate, not folinic acid (→ `CHEBI:15640`) — a wrong compound that would have passed unnoticed if the label had not been checked. `HP:0000708` is "Atypical behavior", not "Behavioral abnormality".

## Sources and validation

Perplexity `sonar-deep-research` (report committed under `research/`); 6/6 identifiers resolved, 0 confabulation, 4/6 assessed on topic. Used as a lead source only — every snippet anchored to an independently fetched PMID.

- `just validate` — **36/36 snippets verified**
- `just validate-disorders` — passed
- `just check-entity-refs`, `just check-duplicate-keys`, `just validate-history` — passed

`cache/` rows are limited to CURIEs this entry actually uses; a CHEBI row from the abandoned `CHEBI:15636` binding was reverted.

---

### DIFF
--- a/history/disorders/Methylcobalamin_Deficiency_Type_cblG/2026-08-28T125746Z-claude-code-f714f1.yaml
+++ b/history/disorders/Methylcobalamin_Deficiency_Type_cblG/2026-08-28T125746Z-claude-code-f714f1.yaml
@@ -1,0 +1,33 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Methylcobalamin_Deficiency_Type_cblG
+  path: kb/disorders/Methylcobalamin_Deficiency_Type_cblG.yaml
+session:
+  id: 2026-08-28T125746Z-claude-code-f714f1
+  timestamp: '2026-08-28T12:57:46Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9815
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9829
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  summary: 'Review round 3: correct the MMA monitoring goal'
+  details: |
+    Acted on the non-blocking note carried in the approving review.
+
+    The Long-Term Biochemical Monitoring treatment transcribed the GeneReviews goal of lowering plasma tHcy and MMA. That sentence is written for the intracellular cobalamin disorders as a group; MMA is normal in cblG, which this entry's own Methylmalonic Acid biochemical record states and which is the whole point of that record as a discriminator between isolated remethylation defects and the combined ones. The description asserted a monitoring target that does not exist in this disease.
+
+    The description now states the two goals that do apply here (lower tHcy, methionine kept in range), and says explicitly that the MMA half of the GeneReviews sentence covers the disorder group rather than cblG, pointing at the biochemical record. The snippet is unchanged, since it is an exact quote of what the source says; the evidence item is regraded SUPPORT to PARTIAL, because the quoted sentence supports only part of the claim it is attached to, and the explanation records that reasoning.
+
+    Validation on this head: just validate passes with 46/46 snippets verified; check-entity-refs and check-duplicate-keys clean.
--- a/kb/disorders/Methylcobalamin_Deficiency_Type_cblG.yaml
+++ b/kb/disorders/Methylcobalamin_Deficiency_Type_cblG.yaml
@@ -829,10 +829,13 @@ treatments:
     explanation: GeneReviews prevention-of-primary-manifestations advice.
 - name: Long-Term Biochemical Monitoring and Dose Adjustment
   description: >-
-    Long-term management targets three numbers rather than symptoms: lower plasma total
-    homocysteine, lower methylmalonic acid, and plasma methionine kept within the normal range.
-    The third is the one specific to the remethylation disorders - in cblG methionine is low,
-    so the goal is to raise it, the opposite of management in classical homocystinuria.
+    Long-term management targets numbers rather than symptoms: lower plasma total homocysteine,
+    and plasma methionine kept within the normal range. The methionine goal is the one specific
+    to the remethylation disorders - in cblG methionine is low, so the goal is to raise it, the
+    opposite of management in classical homocystinuria. GeneReviews states the goal as lowering
+    tHcy *and* MMA, but that sentence covers all the intracellular cobalamin disorders
+    collectively; MMA is normal in cblG (see the Methylmalonic Acid biochemical record), so
+    there is nothing to lower on that arm and it is not a monitoring target here.
   therapeutic_modality: BEHAVIORAL
   treatment_term:
     preferred_term: Supportive Care
@@ -847,10 +850,13 @@ treatments:
   evidence:
   - reference: PMID:20301503
     reference_title: "Disorders of Intracellular Cobalamin Metabolism."
-    supports: SUPPORT
+    supports: PARTIAL
     evidence_source: HUMAN_CLINICAL
     snippet: "Long-term management focuses on improving the metabolic derangement by lowering plasma tHcy and MMA concentrations and maintaining plasma methionine concentrations within the normal range."
-    explanation: GeneReviews long-term management goals.
+    explanation: >-
+      GeneReviews long-term management goals. Graded PARTIAL because the sentence is written
+      for the intracellular cobalamin disorders as a group: the tHcy and methionine goals apply
+      to cblG, the MMA goal does not, since MMA is normal in an isolated remethylation defect.
 discussions:
 - discussion_id: cblg_treatment_alters_natural_history
   kind: KNOWLEDGE_GAP


matentzn:
All four IMPORTANT items addressed in 5750d6fd7, plus suggestions 6, 8, 9 and 11. One suggestion declined on evidence.

## The MMA binding was a real defect, and my note made it worse

You're right, and the part I want to flag is that I didn't just get the binding wrong — I wrote a `notes` line rationalizing it ("a normal analyte has no abnormality term of its own, so it binds to the homocystinuria term"). That reasoning is confident and wrong, and it would have made the error harder for the next reader to spot than if I'd left it bare. Rebound to `CHEBI:30860` per the `Adenosine_Kinase_Deficiency.yaml` convention you pointed at, and the note now says why an HP term is the wrong shape for a `NORMAL` value rather than defending the old choice.

## GeneReviews

Mined, and the `notes` line corrected to name the sections actually used. Three new treatments:

**Nitrous oxide avoidance** — this was the best catch in the review. N₂O irreversibly oxidises the cobalamin cofactor of methionine synthase, so for cblG specifically it removes whatever residual activity the hydroxocobalamin is sustaining. The entry says that, and also flags that the same GeneReviews sentence rules out *methionine restriction* — the intuitive but wrong move in a disease whose name contains "homocystinuria".

**Catabolic-state avoidance** and **long-term biochemical monitoring**, the latter carrying the three-number management goal. Plus the 25/50/25 recurrence-risk sentence on `inheritance` and the "early institution of injectable hydroxocobalamin improves survival" sentence.

That last one went into the `KNOWLEDGE_GAP` discussion rather than only into treatments, because it is a distinct third position: GeneReviews claims a survival benefit while explicitly declining to claim full prevention of manifestations. The discussion now holds all three views rather than the two it started with.

## Pathograph connections

`target_mechanisms` on all seven treatments (the four you listed plus the three new ones) and eight `downstream: phenotypes#…` edges. I went past your minimum: `Cerebral White Matter Injury` also reaches seizure, encephalopathy and the late-onset psychiatric phenotype, and `Methionine and S-Adenosylmethionine Depletion` reaches developmental delay and hypotonia — all `INDIRECT_UNKNOWN_INTERMEDIATES`, since the route from a methylation deficit to a specific neurological sign is not established in this disease.

## Suggestions

- **6 (diagnostic criteria)** — added as a `DIAGNOSTIC_CRITERIA` definition. Worth having because each of the three numbers carries information the others don't: homocysteine says remethylation, normal MMA excludes the combined defects, and low methionine is what separates this from cystathionine beta-synthase deficiency where methionine is *high*. The definition also states what the triad cannot do — separate cblE from cblG.
- **8, 9, 11** — all added.
- **7 (animal models)** — no `Mtr` model comparable to the cblE `Mtrr` gene-trap was found. Recorded as a `notes` line, per your suggestion that saying so beats silence.

## Declined: suggestion 5

`GO:0009086` is **obsolete** — `just validate-terms` rejected it with "not in dynamic enum" and the label check returned `obsolete methionine biosynthetic process`. Your underlying point was right, so I used `GO:0071265` (L-methionine biosynthetic process) instead.

Flagging it because it's the fourth wrong ontology ID across these two PRs — the others were `CL:0038024` (doesn't exist), `CHEBI:15636` (a different compound), and `CHEBI:35044` (Ulifloxacin, for pristanic acid, in a sibling PR). Every one was caught by the label check rather than by anyone reading the YAML, which is a decent argument for that check being the load-bearing one in this stack.

## Validation

46/46 snippets verified; `validate`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys` all pass. No new cache rows — `CHEBI:30860` and `GO:0071265` were already cached.

---

matentzn:
Pushed `18d2bbf0e`, acting on the non-blocking note you carried into the approval — the one about the monitoring treatment transcribing GeneReviews' "lower tHcy and MMA" goal when MMA is normal in cblG. This resets the approval, which I wouldn't normally spend a round on, except that the defect is the same class as the MMA binding you blocked on in round 1: the entry asserting something about methylmalonic acid that its own biochemical record contradicts.

**What changed.** The `description` now states the two goals that apply here — lower tHcy, methionine kept in range — and says explicitly that the MMA half of the GeneReviews sentence covers the intracellular cobalamin disorders as a group, pointing the reader at the Methylmalonic Acid record. The whole point of that record is that normal MMA is what discriminates an isolated remethylation defect from the combined ones, so a monitoring goal of lowering it was quietly undoing the record's own argument.

The `snippet` is unchanged — it's an exact quote of what GeneReviews says, and the source isn't wrong, it's just written for the group. The evidence item is regraded `SUPPORT` → `PARTIAL`, since it supports part of the claim it carries and not the rest, with the reasoning in the `explanation`.

**Validation on this head:** `just validate` passes with **46/46 snippets verified**; `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens` all clean. Working tree clean afterwards — no cache churn. History record added.

---

PR #9830, opened 2026-08-28
Recurate Gilbert's Syndrome: pharmacogenomics, gallstones, and model limits

aleixpuigb:
The Gilbert's Syndrome entry was written in March 2026 without deep research, and it described the disease as a bilirubin problem that is harmless. Two of those things are incomplete. UGT1A1 does not only conjugate bilirubin — it also handles several drugs, and that is where the syndrome actually causes harm. And Gilbert's is not entirely without consequence: gallstones are more common, and that link has been shown to be causal rather than just correlated.

This recuration was driven by one OpenScientist deep-research report (30 references, all resolved, none off topic; the Named Entity Confusion preflight passes against MONDO:0007745).

## What was missing

**The drug arm was only a footnote.** The entry mentioned irinotecan inside a treatment, but never modelled the mechanism. Reduced UGT1A1 activity lowers glucuronidation of SN-38 (irinotecan's active metabolite) and leaves less spare capacity when a drug inhibits the same enzyme. In the Swiss HIV Cohort, 67% of `UGT1A1*28` homozygotes on atazanavir or indinavir had repeated bilirubin readings in the visible jaundice range, against 7% of people with neither risk factor. Efavirenz does the opposite and lowers bilirubin by inducing the enzyme. That is now two pathophysiology nodes.

**"Benign" was overstated.** Of 54 illnesses tested in 138,125 UK Biobank participants, cholelithiasis was the only one whose raised prevalence in Gilbert's survived adjustment for confounders and was confirmed causal by Mendelian randomization. Prevalence in men with Gilbert's was 20%. The effect is larger where bilirubin production is also high — in a 913-patient sickle cell cohort the same genotype raised gallstone frequency and lowered age at onset. Added as a mechanism node and a `Cholelithiasis` phenotype.

**The diagnostic threshold is sex-dependent and the entry used one number.** The historical unisex cutoff of 1 mg/dL identifies 3.7% of women in the UK Biobank cohort; sex-, age- and genotype-stratified centiles identify 10% of the same women. Both definitions are now recorded in `definitions:`, with the historical one marked `UNVALIDATED` and the stratified one `PROPOSED`, plus reference ranges with interpretation bands on the bilirubin biomarker.

**Neither animal model is a model of this disease.** The Gunn rat is a UGT1A1 null and the humanized UGT1 mouse spontaneously develops severe neonatal hyperbilirubinemia with neurologic dysfunction. Both sit at the Crigler-Najjar end of the spectrum, while Gilbert's is roughly 30% residual activity from a promoter repeat. They are curated as `PARTIALLY_RECAPITULATES` at `LOW` fidelity with the limitation spelled out, and a `HUMAN_MODEL_MISMATCH` discussion records that neither can address what actually makes penetrance incomplete.

## Also added

- `inheritance:` with `INCOMPLETE` penetrance — Bosma's 1995 "necessary but not sufficient" was quoted in the entry but never modelled.
- Fasting as an environmental exposure wired into the pathograph via `influences_mechanisms`, reusing the KB's existing `XCO:0000102` binding.
- A `Fatigue` phenotype. A UK primary-care study of 9,240 cases found fatigue and abdominal pain — and only those two — raised up to five years before diagnosis.
- A second prevalence record for clinically recorded diagnoses (180.4 per 100,000), roughly fifty-fold below the genotype prevalence. That gap is the entry's most concrete measure of incomplete penetrance.
- Crigler-Najjar type II and haemolysis as differentials, three knowledge gaps, and the antioxidant-protection hypothesis as `ALTERNATIVE` carrying both its in-vitro support and its Mendelian-randomization refutation.

## Two pre-existing defects fixed

The prevalence record had `population: Population studies`, which names an ascertainment method rather than a cohort, and carried the deprecated `percentage` slot.

## One deep-research claim I did not follow

The report presents cardiovascular protection from raised bilirubin as "a striking and reproducible feature", and proposes as future work the Mendelian randomization that PMID:37456363 had already published with a negative result. I kept the hypothesis as `ALTERNATIVE` with the refutation attached and recorded the discrepancy in its `notes`, because reviewers will keep meeting the protective framing in secondary literature.

## Validation

`just validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-environmental-evidence` all pass. 71 of 77 snippets verify against the reference cache; the other 6 are skipped by prefix.

Two snippets were trimmed rather than quoted in full because the reference validator strips square-bracketed spans, and neither a confidence interval (`[1.3-1.7]`) nor a bracketed unit conversion (`[>2.5 mg/dL]`) matches the two patterns currently in `literal_bracket_patterns`. Widening that config is a repo-wide change and did not belong in a curation PR; filed separately.

Reference cache files show normalization churn (quoting style, `full_text_attempted`, and full text where entries were previously abstract-only) from the sanctioned fetch path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AxdGn3F7VVBwfFWEr8UPQn)_


aleixpuigb:
All three blocking items fixed, plus every non-blocking suggestion except one. Pushed as 390853d.

**Fatigue band — you were right, and the way I got it wrong is worth recording.** I set `FREQUENT` from the paper's prose ("consistently more common in GS cases") without opening its Table 4. The row reads `Fatigue 724 (12%) 5798 (6%) 677 (22%) 6780 (13%) 1401 (15%) 12,581 (8%)` — 1,401/9,240 cases is 15%, which is `OCCASIONAL`. Corrected, and the entry now quotes the table row rather than the prose, so the number that sets the band is visible in the evidence. The row also shows a sex difference the single band averages over — 22% of female cases against 12% of male — which is now in the description.

**Orphaned hypothesis group** — fixed exactly as you proposed, on the edge into `Bilirubin Antioxidant and Pleiotropic Effects`.

**Treatment join point** — added `target_mechanisms` on the genotyping treatment, pointing at both new nodes: `MODULATES` on `Impaired Glucuronidation of Drugs and Drug Metabolites` (dose matched to measured capacity) and `INHIBITS` on `Drug-Unmasked Hyperbilirubinemia`.

### Suggestions

- **Phenobarbital** added as `Phenobarbital Enzyme Induction`, with a second join point on the glucuronidation node. Its evidence is graded `PARTIAL` and says why: the inducer actually observed in the cited cohort is efavirenz, not phenobarbital.
- **TA allelic series** added — `*36` (TA5, raises activity) and `*37` (TA8, lowers it further), with the African allele frequencies up to 0.08 and 0.07. The point that made it worth curating is that an assay scoring only (TA)6 against (TA)7 misclassifies in *both* directions, and DPWG already treats `*37` as a deficiency allele for irinotecan dosing.
- **G6PD** added as `relationship_type: MODIFIER`. I did not add SLCO1B1, HMOX1 or BLVRA: the report names them but I could not find a citable Gilbert-specific claim for any of them, and adding a gene on the strength of a report's list is how unsupported content gets in.
- **`diagnosis:` section** added — diagnosis by exclusion, the fasting provocation test, and (TA)n genotyping.
- **Model entity refs** rekeyed from `species` to `name`, as you suggested.
- **History record** corrected on both counts and extended with this round.

That consumes all three fetched-but-uncited caches.

### Two things I found while doing this

**The deep-research report's HGNC id for G6PD is wrong.** It gives `HGNC:4118`, which is GALK1. G6PD is `hgnc:4057`, confirmed with OAK before use. Worth knowing that the report's "Ontology terms" lines are not reliable — the same report also suggests `MONDO:0008738` for this disease where the entry uses `MONDO:0007745`, and `preflight-dr` confirms the entry's is right (OMIM 143500 matches).

**GeneReviews is now verified.** I have PubMed access here, so I ran the check you couldn't. `Gilbert syndrome GeneReviews` returns exactly one hit: PMID:20301450, the **Alagille Syndrome** chapter, matching only because an author is named Gilbert. There is no standalone chapter for Gilbert syndrome and no Crigler-Najjar chapter either. Skill item 15 needs no action and no `tags:` block. Your expectation was right.

### One more thing

Two of your findings — the orphaned hypothesis group and the missing treatment join point — applied to the sibling Cowden PR (#9831) as well. I have fixed both there rather than wait for the same review round.

---

aleixpuigb:
Both round-2 items fixed in 5faedd9. Both were in content I added in round 1, which is worth noting on its own.

**Item 1 — UGT1A1 typed as a modifier of the disease it causes.** You're right on both halves. `association: Causative` next to `relationship_type: MODIFIER` is a contradiction, and the schema is explicit that MODIFIER means "without being a primary driver". Worse, the record bundled two alleles pointing in opposite directions — its own notes said `*37` lowers activity while `*36` raises it above wild type — so no single `relationship_type` could have been right.

Split into two records: `UGT1A1*37` as `CAUSATIVE` (it acts through the same reduced-transcription mechanism as `*28`, more strongly), and `UGT1A1*36` as `PROTECTIVE`. Each now quotes the half of the source sentence that is about it, rather than both citing the same combined sentence.

The ancestry point survives the split and reads better for it: a (TA)6-versus-(TA)7 assay misclassifies African-ancestry individuals in *both* directions — calling some higher-activity `*36` carriers wild type, and missing the lower-activity `*37` ones entirely.

**Item 2 — phenobarbital asserted without evidence for phenobarbital.** Correct, and the evidence item I did attach undercut itself: its own explanation said the efavirenz observation "is not direct evidence for phenobarbital", which should have told me to go find the direct evidence rather than ship the hedge. `PMID:9435989` was already cited three times in this file and covers it — Crigler-Najjar type II "is treatable with phenobarbital therapy" and is the form retaining residual enzyme, which is what makes the response a discriminator against the type I null. Added as entry-level evidence on the treatment, and to `differential_diagnoses#Crigler-Najjar syndrome type II` where the same claim was also uncited.

### Suggestions

- **`INHIBITS` on a genotyping test** — agreed, changed to `MODULATES` to match the sibling link. A test doesn't act on the node; the effect depends entirely on what the prescriber does next, and the description now says that.
- **Whether phenobarbital belongs in `treatments:` at all** — fair question. I kept it, because it *is* a therapeutic that lowers unconjugated bilirubin and the entry's differential diagnosis depends on its effect, so removing it would leave that claim homeless. But the description now leads with the fact that it is not indicated and says why (sedation, for a benign condition), so nobody reads it as a recommendation.
- **`*60`, SLCO1B1, HMOX1, BLVRA** still unmodelled. `*60` is a genuine gap I could close with more search time; the three modifier genes I'd rather leave out than add on the strength of the deep-research report listing them, since I couldn't find a citable Gilbert-specific claim for any of them.

Thanks for re-running the gates locally and hand-checking the new snippets — and for confirming the GeneReviews result matched your offline expectation.

---

aleixpuigb:
Fixed in d30cfc9. You're right, and the typing was mine to justify regardless of where the suggestion came from — I asserted a direction of effect and cited a sentence that establishes existence and ancestry frequency instead.

The detail that makes it worse: the abstract I cited calls its one (TA)5/6 heterozygote *"referred as mutated"*, which points the opposite way from `PROTECTIVE`. So the evidence I attached wasn't merely silent on direction, it leaned against my typing.

`PMID:22407023` fixes it, and the sentence is better than either fragment you quoted because it covers the whole series in one go:

> there are a number of thymine-adenine (TA) repeats varying from 5 to 8, with an inverse correlation between promoter activity and number of TA repetitions: A(TA) 8TAA promoter with reduced activity, A(TA) 7TAA promoter with reduced activity (UGT1A1 \*28), A(TA)6TAA wild-type and A(TA)5TAA promoter with increased activity.

That grounds `*36` as `PROTECTIVE` and independently grounds the `*37` record's "further reduction" claim, so I cited it in both records rather than only the one you flagged. Graded `OTHER` for the reason you gave — introduction prose, not a result of the family study — which also keeps it clear of the `HUMAN_CLINICAL` grading the G6PD record gives that paper's own findings. `check-snippet-grading` passes.

**History record refreshed.** You were right that it was a round stale. It now covers the allele split, the phenobarbital citations, the `INHIBITS`→`MODULATES` change, and the cache commits — plus the round-4 item above and why the earlier evidence didn't support the typing.

**One thing worth flagging from the sibling PR.** You noted that commit 062f077 restored a `full_text_attempted` line dropped by 79656da, and that the net change was zero. That wasn't a coincidence — I went looking for it after the Cowden reviewer found the same class of defect there, where it cost real content: `PMID_26673699.md` went from `full_text_xml` (139 lines) to `abstract_only` (44), losing 95 lines of KLLN full text.

Cause in both cases: my cache commits *copied* files in from another checkout rather than fetching them, and some copies were staler than what the branch already had. The check that finds it is diffing modifications separately from additions — across all four cache commits on both PRs, only three files were modifications rather than additions, and two of those were reversions. Both fixed by regenerating through `just fetch-reference`.

Not raising it here because anything is outstanding on this PR — it isn't — but because "copying cache files between checkouts" is a mistake worth having on the record next to the commits that made it.

**Not done, and I'd still leave them:** `*60`, and SLCO1B1 / HMOX1 / BLVRA. Agreed on the three modifier genes.

---

PR #9831, opened 2026-08-28
Recurate Cowden Syndrome: model systems, plasma succinate, and three open questions

aleixpuigb:
The Cowden Syndrome entry had no model systems at all — no mouse, no cell system — despite mouse and organoid work being cited inside its evidence items. It also had no biochemistry section, no treatment reaching the pathograph, and no record of the questions its own guideline sources say are unanswered. This adds those.

Recuration driven by one OpenScientist deep-research report (30 references, all resolved, none off topic). Unlike its sibling Gilbert's Syndrome entry, this one had already been improved by five later curation sessions, so pathophysiology, phenotypes and treatments were in good shape. This PR targets what those sessions did not reach.

## Model systems

Three added, each connected to a mechanism node through `modeled_mechanisms` rather than left as a disconnected list entry:

| Model | Node | Relationship | Fidelity |
|---|---|---|---|
| Epithelial-specific Pten deletion (mouse) | Hamartoma Formation | `RECAPITULATES` | HIGH |
| Neuronal Pten deletion (mouse) | Neurodevelopmental Cortical Dysfunction | `RECAPITULATES` | MODERATE |
| Mammary Pten loss + constitutive Stat5 (mouse) | Increased Cancer Risk | `PARTIALLY_RECAPITULATES` | MODERATE |
| PTEN-mutant iPSC forebrain organoids | Neurodevelopmental Cortical Dysfunction | `RECAPITULATES` | MODERATE |
| — same organoids, perifosine arm | PTEN Loss and PI3K/AKT/mTOR Pathway Activation | `RESCUES` | MODERATE |

The colorectal model is the interesting one. Cowden juvenile polyps had been attributed to somatic PTEN loss in the stroma; deleting Pten only in the epithelium produces the polyps anyway, with the stromal changes following.

The mammary model covers the entry's largest risk — breast cancer, 77% cumulative to age 70 — which the other two do not touch. Its interventional arm is what makes it worth having: mammary-specific Stat5 ablation *prevents* carcinogenesis in the Pten-deficient background, so the cooperation is a dependency rather than an association.

The perifosine arm is the entry's only demonstration that the phenotype responds to inhibiting the pathway, and it is in human tissue.

Every link carries an explicit `limitations` field. The mouse models are single-tissue, so each captures one manifestation of a multi-system syndrome; the mammary model's cooperating Stat5 lesion is engineered rather than something patients carry; the organoids have no vasculature or microglia. The organoids also resolve something worth stating: the two alleles do not act at the same time. G132D (autism-associated) disrupts neuroectoderm formation within the first several days, while M134R (cancer-associated) looks like wild type then and only diverges by 72+ days.

## Plasma succinate

The entry had no `biochemical:` section. Elevated plasma succinate was found in 62% of individuals carrying a germline PTEN, SDHB or SDHD mutation against 16% of controls — and, more informatively, in 67% of those with a pathogenic PTEN mutation against under 20% of mutation-negative individuals meeting identical clinical criteria.

This is the only proposed account tying the entry's PTEN-positive and PTEN-negative arms together, so it is also recorded as an `EMERGING` mechanistic hypothesis, with a `Reduced Succinate Dehydrogenase Activity` node reached by an edge carrying `causal_link_type: UNKNOWN`. That node is deliberately a leaf: what accumulated succinate would then do in Cowden syndrome has not been shown, so nothing is drawn downstream of it. Both the biomarker and the hypothesis say plainly that this rests on one single-centre study of 21 carriers, and that the study measured the metabolite and inferred the enzyme.

## Treatments joined to the pathograph — and one deliberately not

`mTOR Inhibitor Therapy` `INHIBITS` the PI3K/AKT/mTOR node; `Prophylactic Mastectomy` `INHIBITS` `Increased Cancer Risk` by removing the substrate tissue.

`Enhanced Cancer Surveillance` deliberately carries **no** `target_mechanisms`. An earlier revision gave it a `MODULATES` edge; the reviewer pointed out that `MODULATES` means "alters the mechanism" while the link's own description said surveillance does not change the risk. Surveillance shifts *when* disease is detected, dismech has no predicate for that, and the honest encoding of "no predicate fits" is no edge. It reaches the pathograph through the discussion that attaches to it. `Genetic Counseling` has no edge for the same reason.

## Three open questions

- **Does the surveillance actually work?** The ERN GENTURIS guideline that recommends it says, in its own words, that "There is a need for prospective evaluation of the effectiveness of these recommendations in the PHTS population" — while also noting the programme "require[s] significant patient commitment". A demanding programme whose issuing guideline says its effectiveness is unevaluated is a gap worth recording, not a note.
- **What did the everolimus trial actually rule out?** The one randomised trial missed its primary neurocognitive endpoint and is often read as negative for the drug class. That may be too broad: it asked six months of drug to move a phenotype set during fetal and early postnatal development, and did not use hamartoma burden as an endpoint. Recorded as a `CONTROVERSY`, which is why the entry grades the trial `PARTIAL` rather than `REFUTE`.
- **Are the genotype-phenotype correlations replicated?** Promoter variants have been linked to breast cancer and nonsense variants to colorectal. If those hold they would let surveillance be stratified by variant class. They are currently single-cohort associations, and this entry does not act on them.

Also adds a top-level `inheritance:` block — autosomal dominant, `INCOMPLETE` penetrance, `VARIABLE` expressivity, with the GeneReviews 50% transmission risk. It deliberately does **not** claim a de novo proportion: GeneReviews declines to quantify the simplex/familial split because the condition is underdiagnosed, and simplex is not de novo.

## Validation

`just validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass. All 96 snippets verify against the reference cache.

Three snippets needed character-level correction before they verified. One cited paper's full text uses non-breaking hyphens (U+2011) where the rendered text looks like ordinary hyphens; one had been truncated mid-sentence at a comma; and `check-snippet-grading` caught a quote re-graded `OTHER` where the same quote is `HUMAN_CLINICAL` three times elsewhere in the file.

## One note on the deep-research report

Its reference validation flags one quoted claim as "not found in source", but the flagged quote and the "closest text in source" it prints are byte-identical. The mismatch is the validator stripping a square-bracketed confidence interval before matching — a tooling artifact, not a confabulation, and the same behaviour came up independently in the Gilbert's PR (#9830). Filed as #9832.


aleixpuigb:
All four blocking items fixed, plus two of the suggestions. Pushed as 44e6282. I checked every item against the cached source and every one held up.

**Item 1 — surveillance discussion cited the wrong sentence.** Correct, and the sentence you quoted is in the cache verbatim. The discussion now carries *"There is a need for prospective evaluation of the effectiveness of these recommendations in the PHTS population"* plus the patient-commitment sentence for the burden half. The combination is the whole point of the gap, so having it sourced rather than asserted matters.

**Item 2 — organoid limitations mis-stated the paper.** You're right, and it's the more interesting of the two. I flattened a two-allele result into one window. G132D disrupts neuroectoderm formation in the first several days; M134R looks like wild type then and only diverges at 72+ days. Since the paper's headline finding *is* that different missense alleles act at different stages, collapsing it lost the finding. Both the `description` and `limitations` now carry both alleles and both timescales, and the alleles are named.

**Item 3 — the de novo claim.** Uncited, and wrong in a way I should have caught: simplex is not de novo, and I conflated them. GeneReviews was already cached in this entry and is more careful than I was — the split cannot be determined because the condition is underdiagnosed, perhaps 10–50% have an affected parent. Replaced with the sourced 50% transmission risk and the GeneReviews hedge, quoted at length because the hedge is the substance. The `explanation` says explicitly what the earlier draft claimed and why it was wrong.

**Item 4 — the unused breast-arm model.** You're right that this fell short of the PR's own stated scope, not just of a wider one. Breast cancer is the largest risk in the entry at 77% cumulative to 70, and neither model I added touched it. `PMID:24469394` is now a third `animal_models` entry — mammary Pten loss with constitutive Stat5, `PARTIALLY_RECAPITULATES` against `Increased Cancer Risk`. The interventional arm is what makes it worth having: mammary-specific Stat5 ablation *prevents* carcinogenesis in that background, so the cooperation is a dependency rather than an association, and that is the readout. The limitation states that the cooperating Stat5 lesion is engineered rather than something patients carry.

### Suggestions

- **Organoid link evidence** was background (an ASD prevalence statistic) rather than a claim about the model. Now quotes the isogenic-design sentence, which does the job the link level is for.
- **Perifosine rescue arm** added as a second `RESCUES` link with a `RESTORED` readout. Worth having for the reason you gave: it's the entry's only demonstration that the phenotype responds to inhibiting the pathway, and it's in human tissue. Its `limitations` records that the correction was partial, was shown for G132D only, and says nothing about clinical benefit.

Not acted on, both for the reason you gave:

- **KLLN H3K9me3** needs a pathophysiology node, which is outside this PR. Agreed it's the strongest remaining gap for the next pass.
- **Neuronal Pten aggregating a class behind a review** — the `limitations` field discloses it, and swapping in a primary study is better as its own change than bolted onto a review round.

### One already fixed

Your suggestion that the succinate hypothesis group had no edge opting into it was already fixed in e86bfed, pushed before your review landed. It came out of round 1 on the sibling PR #9830, where the same defect was flagged as blocking. Fixing it properly needed a node, since the hypothesis is about a step the pathograph didn't contain — so there is now a `Reduced Succinate Dehydrogenase Activity` node reached by an edge with `causal_link_type: UNKNOWN` carrying the group. It is deliberately a leaf: what accumulated succinate would then do in Cowden syndrome has not been shown, so nothing is drawn downstream of it.

That push also added `target_mechanisms` to three treatments, which was the other cross-cutting finding from #9830.

---

aleixpuigb:
You're right about the surveillance edge, and I'm removing it rather than arguing.

`MODULATES` means "alters the mechanism". My link's own description opened with "Surveillance does not change the cancer risk conferred by PTEN loss." Those cannot both stand, and a pathograph consumer would have read a treatment→mechanism edge that isn't there. Detection timing is real, dismech has no predicate for it, and the honest encoding of "no predicate fits" is no edge. Removed in 15591e7, with a `notes` line recording why, and the treatment still reaches the pathograph through `discussions#cs_surveillance_effectiveness`.

You also named the pressure accurately. The edge existed because a compliance metric rewards treatments joining the pathograph. Worth noting that the same commit applied the correct test one treatment over — `Genetic Counseling` got no edge because it does not act on a mechanism — so I had the right test in hand and did not apply it consistently.

The link-level evidence went with the edge, as you said it should.

**Succinate node grounding** — done, and thank you for checking it was free. `GO:0000104` succinate dehydrogenase activity with `modifier: DECREASED`, alongside the existing CHEBI succinate annotation. The node was named for an enzyme activity and annotated only the metabolite; they are different claims, which is the point of the node.

**PR description** refreshed to match what will merge — model table, the succinate node, the treatment joins, and the surveillance non-join with its reasoning.

### Round-1 items

All four were fixed in 44e6282, pushed at 12:14, a couple of minutes after your round-2 review went up — so you were reviewing e86bfed and hadn't seen them. Briefly:

1. Surveillance discussion now cites *"There is a need for prospective evaluation of the effectiveness of these recommendations in the PHTS population"* plus the patient-commitment sentence.
2. Organoid `limitations` and `description` now carry both alleles and both timescales. You were right that flattening it lost the paper's headline finding.
3. The de novo claim is gone, replaced with the GeneReviews 50% transmission risk and the simplex hedge quoted at length. I had conflated simplex with de novo.
4. `PMID:24469394` is now a third `animal_models` entry, `PARTIALLY_RECAPITULATES` against `Increased Cancer Risk`, with the Stat5-ablation prevention result as its readout.

Plus both round-1 suggestions: the organoid link evidence now quotes the isogenic-design sentence instead of ASD background, and the perifosine rescue arm is a `RESCUES` link with a `RESTORED` readout.

### Not done, and why

- **KLLN H3K9me3** needs a pathophysiology node. Agreed it's the strongest remaining gap for the next pass on this entry.
- **Neuronal Pten behind a review citation** — swapping in a primary study is better as its own change than bolted onto a review round.
- **`mTOR Inhibitor Therapy → INHIBITS`** — I read your note as "no action", and agree the awkwardness is the target node bundling the lesion with its consequence, which predates this PR.

---

aleixpuigb:
Thanks for the approval. I've pushed 87b5d8d anyway, because one of your non-blocking notes turned out to be a defect I introduced rather than a nit — and I'd rather lose the approval than leave it in.

**The cache downgrade was a silent reversion.** You're right that `PMID_26673699.md` came back `abstract_only` where it had been `full_text_xml`. The cause: my cache commit copied files in from another checkout instead of fetching them, and the copy was staler than what the branch already had. 95 lines of KLLN full text overwritten. Nothing cites it so nothing broke, but "reverted a file to a worse version and committed it" is exactly what a reviewer shouldn't have to catch by hand-diffing.

Regenerated with `just fetch-reference` — full text restored, 138 lines.

Then I audited the whole class rather than just the file you named. Those cache commits touched three files as *modifications* rather than additions:

| File | Before | After | |
|---|---|---|---|
| `PMID_24469394` | abstract_only, 56 lines | full_text_html, 145 | upgrade |
| `PMID_26673699` | full_text_xml, 139 | abstract_only, 44 | **downgrade, fixed** |
| `PMID_28338110` (on #9830) | had `full_text_attempted` | lost it | metadata only, fixed there |

The generalisable lesson: copying cache files between checkouts isn't the sanctioned fetch path, and `--diff-filter=M` is what separates the reversions from the additions. I should have run that check before committing 53 cache files, not after you found one.

**Three of your other suggestions taken:**

- `notes:` → `review_notes:` for the surveillance no-edge rationale. You're right, and the CLAUDE.md parallel you drew is exact — the environmental waiver rule says `notes:` is disease content. A curation decision doesn't belong in clinical prose.
- `DECREASED` → `ABOLISHED` on the mammary readout. The source says "prevent", and `ABOLISHED` is in the enum. Mine was weaker than the source, which is the wrong direction to be wrong in.
- The perifosine link now cites the single-cell RNAseq sentence — AKT inhibition upregulating neuronal cell-fate signatures — so the link and its readout rest on different measurements. You're right that one sentence doing both jobs collapses two claims the schema deliberately separates.

**One noted, not acted on:** the GeneReviews `HUMAN_CLINICAL` vs `OTHER` split. I reached the same conclusion you did — the two new quotes are Mendelian arithmetic and a statement about the literature, not patient observations — so I've left it and recorded that it was considered.

**KLLN H3K9me3** remains the strongest gap and is still out of scope here. Worth flagging for whoever picks it up: the full text is now restored in the cache, so the SUV39H1/DBC1 mechanism detail is there rather than needing a re-fetch.

---

PR #9834, opened 2026-08-28
curate: Charcot-Marie-Tooth disease dominant intermediate B (DNM2, MONDO:0011674)

### LINKED ISSUE #9819: Curate Charcot-Marie-Tooth disease dominant intermediate B (MONDO:0011674)
Curate a dismech entry for **Charcot-Marie-Tooth disease dominant intermediate B** ([MONDO:0011674](https://monarchinitiative.org/MONDO:0011674)) — CMTDIB, caused by heterozygous DNM2 variants. Dynamin-2 is a mechanochemical GTPase required for membrane fission in endocytosis and for vesicle traffic along the axon, and the "intermediate" label is the point of interest: nerve conduction velocities fall between the demyelinating CMT1 and axonal CMT2 ranges, so the pathograph has to carry both a Schwann-cell and an axonal arm rather than picking one.

Single gene, no MONDO descendants. Worth noting during curation that DNM2 is allelic with centronuclear myopathy and that some patients show neutropenia or early-onset cataracts; whether those belong in this entry or stay as allelic-disorder cross-references is a curation call to make explicitly.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B.yaml`

matentzn:
New entry for CMTDIB / DI-CMTB, the DNM2-related intermediate neuropathy.

Closes #9819

## "Intermediate" is the reason this is its own entry

Motor median conduction velocities in these patients run from about 26 m/s to normal — below the axonal range at one end, normal at the other. That straddles the CMT1/CMT2 boundary, so the pathograph carries **both** a Schwann cell arm and an axonal arm rather than picking one.

The Schwann cell node is deliberately thin, and I want to flag that rather than let it look better supported than it is: **no source cited here demonstrates a primary Schwann cell lesion.** The node exists because velocities as low as 26 m/s are not what a purely axonal neuropathy produces. Its upstream edge is `INDIRECT_UNKNOWN_INTERMEDIATES` and its one evidence item is graded `PARTIAL` with an explanation saying it supports the conduction slowing, not the lesion.

The axonal arm is much better grounded — zebrafish expressing CMT-mutant human DNM2 mRNA show *total absence* of secondary motor neuron branching, against merely incorrect branching with CNM-mutant mRNA in the same experiment.

## Allelic with centronuclear myopathy, in the opposite direction

DNM2 causes both CMTDIB and autosomal dominant CNM, and the interesting claim is that they sit on opposite sides of normal: CNM variants are proposed to make dynamin-2 *hyperactive* (elevated GTPase and fission), CMT variants to *impair* lipid binding and reduce activity.

That is recorded as `functional_impact_category: PARTIAL_LOSS_OF_FUNCTION` on the variant node — but graded `PARTIAL` / `OTHER`, because the source is a review whose own wording is "CMT mutations **could** impair DNM2 lipid binding and activity". A proposal, not a demonstrated result, and the node says so. CNM is not modeled as a subtype here; it is a different disease that happens to share the gene.

## The GeneReviews chapter for this disease is retired

Worth reading even if you skim the rest. PMID:20614582 opens:

> NOTE: THIS PUBLICATION HAS BEEN RETIRED. THIS ARCHIVAL VERSION IS FOR HISTORICAL REFERENCE ONLY, AND THE INFORMATION MAY BE OUT OF DATE.

It is still the fullest clinical synthesis that exists for CMTDIB, so I used it. But:

- every evidence item drawing on it says in its `explanation` that the source is retired;
- wherever a peer-reviewed cohort paper states the same fact, that paper is cited instead (severity, onset age, electrophysiology, cataracts, neutropenia all come from PMID:19502294 or PMID:22091729, not the chapter);
- the agents-to-avoid recommendation carries an explicit note that it should be checked against current guidance, since the list of neurotoxic agents changes and this chapter does not.

**I tried to record this structurally and could not.** `tags: [retired]` on the top-level reference fails validation — `ReferenceTag` permits exactly one value, `GeneReviews`. So the retirement lives in prose in `notes`, and in the `reference_title` only by the accident that GeneReviews puts the notice in the title string. A curator citing a retired chapter whose title *didn't* say so would leave no trace at all.

That looks like a genuine schema gap to me: nothing in an evidence item can express the standing of the cited source — retired, superseded, retracted, preprint. I have not opened an issue for it (there is a lot of duplicate-issue pressure in this repo and I'd rather check with a maintainer first), but I think it's worth one.

## Things I got wrong

- `HP:0007107` is *Segmental peripheral demyelination*, not a vibration-sensation term. Replaced with `HP:0002936` Distal sensory impairment — which is also the better binding, since the source says "sensory loss" without naming a modality. The stale cache rows from the wrong ID were reverted.
- `HP:0001875`'s label is *Decreased total neutrophil count*, not *Neutropenia*.

## Sources and validation

Perplexity `sonar-deep-research` (report committed); 4/4 identifiers resolved, 0 confabulation — though only 4 of its 19 citations were resolvable identifiers. Lead source only; every snippet independently fetched and verified.

- `just validate` — **28/28 snippets verified**
- `just validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `validate-history` — all pass

Only the five cited PMIDs are included; `cache/` carries one row (`MONDO:0011674`).

---

### DIFF
--- a/history/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B/2026-08-28T125721Z-claude-code-f436a1.yaml
+++ b/history/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B/2026-08-28T125721Z-claude-code-f436a1.yaml
@@ -1,0 +1,41 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B
+  path: kb/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B.yaml
+session:
+  id: 2026-08-28T125721Z-claude-code-f436a1
+  timestamp: '2026-08-28T12:57:21Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9819
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9834
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - animal_models
+  - evidence
+  summary: 'Review round 2: post-approval suggestions'
+  details: |
+    Addressed the five non-blocking suggestions from the approving review.
+
+    1. Regraded the PMID:22451505 myelination evidence item from MODEL_ORGANISM to IN_VITRO. The assay is an ex vivo peripheral nerve culture established from Dnm2-deficient mouse tissue (MeSH: Cells, Cultured; Ganglia, Spinal/cytology; Transfection), which CLAUDE.md classifies as IN_VITRO; MODEL_ORGANISM is reserved for in vivo animal data. The explanation previously implied an in vivo readout and now states the grading reason explicitly.
+
+    2. Retargeted the Dnm2 S619L/K562E RESCUES link. It previously pointed at the variant node, which nothing can rescue. It now points at Reduced Dynamin-2 Membrane Fission Activity, carrying the PIP2 liposome binding readout that is the biochemical event the cross actually corrects, and a second RESCUES link to Length-Dependent Distal Axonal Degeneration carries the organism-scale motor and nerve fibre readouts. The composite motor-coordination readout is graded PARTIAL for the nerve target because the animal also carries a myopathy allele.
+
+    3. Upgraded Reduced Dynamin-2 Membrane Fission Activity to Schwann Cell Myelin Dysfunction from INDIRECT_UNKNOWN_INTERMEDIATES to INDIRECT_KNOWN_INTERMEDIATES, with a description naming the intermediate PMID:22451505 supplies (altered protein surface levels in Schwann cells).
+
+    4. Extended the Intermediate Nerve Conduction Slowing snippet to the end of its sentence, so the non-retired cohort paper PMID:19502294 carries the 26 m/s figure itself rather than leaving it sourced to the retired GeneReviews chapter.
+
+    5. Declined the mappings suggestion. OMIM 606482 has nowhere to go: DiseaseMappings (src/dismech/schema/dismech.yaml:7085) carries only icd10cm_mappings, icd11f_mappings, mondo_mappings and ncit_mappings. The same suggestion was raised and withdrawn on PR 9828 for the same reason.
+
+    Validation on this head: just validate passes with 49/49 snippets verified (was 47/47); validate-terms, check-entity-refs, check-duplicate-keys, check-snippet-length, check-snippet-grading, check-title-snippets and check-folded-hyphens all clean.
--- a/kb/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B.yaml
+++ b/kb/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B.yaml
@@ -170,7 +170,10 @@ pathophysiology:
   - target: Motor Axon Branching and Maintenance Failure
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   - target: Schwann Cell Myelin Dysfunction
-    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+    causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
+    description: >-
+      PMID:22451505 names the intermediate: the endocytic defect alters protein surface levels
+      in Schwann cells, and myelination is strictly dependent on that endocytic function.
   - target: Extraneural Dynamin-2 Dependence
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   evidence:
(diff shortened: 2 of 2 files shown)


matentzn:
The CRITICAL was right, and the way I got it wrong is the part worth writing down. Addressed in bff4f299d along with all five IMPORTANT items and all four suggestions.

## How the false caveat happened

I wrote *"no cited source in this entry demonstrates a primary Schwann cell lesion"* and then encoded that belief structurally — `INDIRECT_UNKNOWN_INTERMEDIATES`, a `PARTIAL` grade, a deliberately thin node, and a paragraph in the PR body drawing attention to it as a limitation.

The sentence was **true of what I had cited and false of the literature**, and the paper that makes it false was committed in this same PR, in the research report, block-quoted at §6.1.

The cause is a workflow choice I'd defend in general and got burned by here. Perplexity's own reference validation reported only 4 of 19 citations as resolvable identifiers, so I treated the report as a lead source, built the entry from PubMed searches, and used the report only to cross-check what I'd found. That protects against confabulation — and it means a *correct*, high-value citation sitting in §6.1 never got read. Being sceptical of the report and being thorough with it are not the same discipline, and I only did the first.

The uncomfortable part is that the caveat made the entry look more rigorous. A confident statement of a limitation reads as diligence, so it invites less scrutiny than a bare claim would. Worth remembering that a caveat is an assertion too, and can be wrong in the direction that flatters you.

## CRITICAL — Schwann cell arm

Rewritten on PMID:22451505. The node now carries `GO:0042552` myelination `DECREASED`, its downstream edge is `DIRECT`, and it holds two evidence items graded separately as you suggested: `MODEL_ORGANISM` for the mouse-tissue myelination result, `IN_VITRO` for the cultured Schwann cell and neuron endocytosis defect.

The tissue-specificity point moved with it. That result — CMT mutants impair myelination, CNM mutants do not — is now sourced to the experiment that showed it rather than to a review's hedge.

## IMPORTANT

**1. Unsupported `DECREASED` modifiers.** You were right that the node asserted two reduced states on a sentence about what dynamin-2 does. Both now have evidence: PMID:22451505 for endocytosis, PMID:40393994 for GTPase activity. The background sentence stays, with its `explanation` rewritten to say it establishes function and the two items below carry the reduction.

**2. Goret 2025 — added, via PubMed.** Your ligature warning was accurate: the DOI cache reads `"opposing effects ofDNM2 mutations"` with hard line-wraps mid-sentence, so I tested candidate spans programmatically and most failed. Rather than quote corrupted text — which would be the faithful thing and also look like an error to every future reader — I fetched **PMID:40393994**, whose abstract is clean, cited that, and discarded the DOI cache as uncited. The review's "could impair" is kept alongside, because it names the specific molecular route (impaired lipid binding) that the in vitro assays don't resolve.

**3. `histopathology` — added.** The sural nerve sentence is the best thing in this PR and it was sitting in a cache I'd already committed. One biopsy showing diffuse loss of large myelinated fibres and regenerating clusters *and* focal myelin thickenings is the human-tissue basis for the entire "intermediate" framing, replicated independently in a second family.

**4. Models — both added with `modeled_mechanisms`.** The zebrafish carries an `ABOLISHED` branching readout (your call on `ABOLISHED` over `DECREASED` is right — "total absence" is the finding, and the CNM comparator is what gives it meaning). The `Dnm2^S619L/K562E` cross is linked `RESCUES` against the variant node, since a genetic mutual rescue is the strongest available evidence for the direction of the defect. Its `limitations` notes the rescue is genetic rather than therapeutic — it requires carrying a second disease allele — and that the muscle phenotyping is more detailed than the nerve phenotyping.

**5. GeneReviews management — all three added**, and orthopedic surgery unbundled from the `DEVICE` entry. Pregnancy management is graded `PARTIAL`: the source states it for Charcot-Marie-Tooth generally, not CMTDIB, and hedges with "in general there appears to be" — the entry doesn't narrow it.

## Suggestions

- **Phenotype edges** — six added. `Length-Dependent Distal Axonal Degeneration` reaches weakness, sensory loss, reflexes and pes cavus; `Extraneural Dynamin-2 Dependence` now reaches the two phenotypes it exists to explain.
- **Cataract onset** — `CHILDHOOD`, with the "before age 15 years" quote added for the bound.
- **Variant spectrum** — three structured variants: both Lys558 alleles (with a note that the neutropenia tracks the *residue*, not the substitution), Gly358Arg, and Thr855_Ile856del as the first proline-rich-domain variant.
- **`mappings` / OMIM 606482** — can't. `DiseaseMappings` has only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings`. Same finding you recorded on #9828, so no OMIM slot exists.

## Issue filed

[#9840](https://github.com/monarch-initiative/dismech/issues/9840), as you asked — searched first, nothing open or closed covered it. It proposes `RETIRED` / `RETRACTED` / `SUPERSEDED` / `PREPRINT` on `ReferenceTagEnum`, notes retraction as the case that most needs a home, and folds in a second instance of the same shape from #9829: evidence drawn from a cohort pooling two diseases can't say so either. Both are the same gap — an evidence item can record what a source claims but not how much of a discount to apply to it.

## Validation

47/47 snippets verified (was 28/28). `validate`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length` all pass. No new `cache/` rows — `GO:0042552` and `CHEBI:46195` were both already there.

Thanks for going and fetching PMID:22451505 rather than just noting the report had an unused citation. That is what turned it from a nit into the finding that fixed the entry's weakest section.

---

matentzn:
Pushed `3cb7bc1fa`, which takes four of the five non-blocking suggestions from your approving review. This resets the approval — sorry for the extra round, but two of them were real defects rather than polish, so holding them back to keep a green tick seemed like the wrong trade.

**Study type on the myelination item.** Regraded `MODEL_ORGANISM` → `IN_VITRO`. You were right that the Sidiropoulos assay is an ex vivo culture: the abstract says *"We used tissue derived from Dnm2-deficient mice to establish an appropriate peripheral nerve model"*, and the MeSH set is `Cells, Cultured` / `Ganglia, Spinal/cytology` / `Transfection`. CLAUDE.md puts tissue explants and ex vivo cultures under `IN_VITRO`. The `explanation` had implied an in vivo readout; it now states the grading reason so this doesn't get "fixed" back later.

**The `RESCUES` target.** Retargeted, as you suggested. It now points at `Reduced Dynamin-2 Membrane Fission Activity` and carries the readout that node is actually about — *"the SL&KE hetero-oligomers normalized the binding to PIP2 liposomes to WT level compared to rDNM2-KE alone"* — graded `IN_VITRO`, since that arm is a co-sedimentation assay with recombinant protein. A second `RESCUES` link to `Length-Dependent Distal Axonal Degeneration` carries the organism-scale readouts. The motor-coordination readout is graded `PARTIAL` on that target: the composite measure reflects both alleles, and the animal carries a myopathy allele too, so it can't be attributed to the nerve arm alone.

**Causal link type.** `Reduced Dynamin-2 Membrane Fission Activity` → `Schwann Cell Myelin Dysfunction` is now `INDIRECT_KNOWN_INTERMEDIATES`, with the intermediate named in the edge description.

**The truncated NCV snippet.** Quoted to the end of the sentence, so PMID:19502294 now carries the 26 m/s figure itself rather than leaving it sourced to the retired chapter. That was the better half of the suggestion — it removes a dependency on the retired GeneReviews rather than just lengthening a quote.

**Not done: `mappings`.** OMIM 606482 has nowhere to go. `DiseaseMappings` (`src/dismech/schema/dismech.yaml:7085`) carries only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings` — no OMIM slot. This is the same suggestion you raised and then withdrew on [#9828](https://github.com/monarch-initiative/dismech/pull/9828) for the same reason, so recording it here too.

**Validation on this head:** `just validate` passes with **49/49 snippets verified** (was 47/47); `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets` and `check-folded-hyphens` are all clean. Working tree clean afterwards — no cache churn, and the `IN_VITRO` regrade needed no new rows. History record added.

---

PR #9835, opened 2026-08-28
curate: spastic paraplegia 89, autosomal recessive (AMFR, MONDO:0957274)

### LINKED ISSUE #9818: Curate spastic paraplegia 89, autosomal recessive (MONDO:0957274)
Curate a dismech entry for **spastic paraplegia 89, autosomal recessive** ([MONDO:0957274](https://monarchinitiative.org/MONDO:0957274)) — a recently delineated hereditary spastic paraplegia caused by biallelic variants in AMFR, which encodes gp78, an ER membrane E3 ubiquitin ligase central to ER-associated degradation. The interest here is mechanistic: it puts an ERAD/proteostasis lesion into the long-axon degeneration pathograph that dismech already models for other HSP subtypes.

Single gene, no MONDO descendants, and the KB precedent is one entry per numbered HSP locus (`Hereditary_Spastic_Paraplegia_49.yaml` was merged recently), so this looks like a leaf Disease. Because the gene–disease relationship is new, the claim should include a check that the evidence base is large enough to support a real pathograph rather than a single-family report — if it is not, saying so is a legitimate outcome.

Stub: `stubs/Spastic_Paraplegia_89_Autosomal_Recessive.yaml`

matentzn:
New entry for SPG89, the AMFR/gp78-related hereditary spastic paraplegia.

Closes #9818

## The claim issue asked whether this is curatable yet. It is.

AMFR was recognised as an HSP gene in 2023, so the issue flagged a risk that the evidence would turn out to be a single family. It isn't: 20 individuals from 8 unrelated consanguineous families, with segregation, patient-derived fibroblasts, patient-derived neural stem cells, electron microscopy and a zebrafish model — plus an independent Algerian group since reporting a further homozygous splice variant. That check is written into the entry `notes` so the next reader doesn't have to redo it.

What the evidence does *not* support is frequencies. No cohort publishes per-phenotype rates, so **no `frequency` value in this entry is a measured proportion** — each is assigned from the founding report's "mainly pure but also complex" wording, and every phenotype description says so.

## The mechanism isn't where the gene's reputation points

gp78 is an ER membrane E3 ubiquitin ligase and a core ERAD component, so the natural expectation is a proteostasis lesion. What the functional work found is disturbed **lipid** homeostasis — lipid droplets accumulating in patient fibroblasts and neural stem cells, reversing on AMFR re-expression. The re-expression rescue is what makes that causal rather than correlative, and it's why the entry follows the demonstrated phenotype instead of the expected one.

The ER morphology change is modeled as a *parallel* consequence rather than upstream of the lipid phenotype, because the cited work doesn't establish which comes first.

## Statins: recorded, and fenced

This is the part I'd most like a second opinion on.

Statins improved the touch-evoked escape response and motor neuron branching in amfra-null zebrafish. That's an attractive finding — a cheap, repurposable, widely-used drug with a mechanistic rationale in an untreatable degenerative disease — and precisely because it's attractive I've fenced it hard:

- the treatment `description` opens **"NOT A TREATMENT FOR PATIENTS"**;
- evidence is `PARTIAL` / `MODEL_ORGANISM`, and the authors' hedge is quoted rather than paraphrased ("suggesting potential therapeutic implications", and elsewhere "may potentially");
- a `RESCUES` model link carries the rescue, with `limitations` stating the gap.

That gap is the subject of a `HUMAN_MODEL_MISMATCH` discussion: **the rescue was measured on larval motor readouts, not on lipid droplet burden.** A lipid-lowering drug improving motor function in a lipid-dysregulated animal is consistent with acting through the lipid mechanism — and equally consistent with statin pleiotropy. The proposed experiment is the obvious one: treat AMFR-null patient cells and measure droplet burden directly.

I'd rather curate this with the caveats attached than leave a mechanism-derived repurposing hypothesis unfindable. If reviewers think a preclinical-only finding shouldn't sit in `treatments` at all, I'm happy to move it — but the `target_mechanisms` link is what makes it reachable from the pathograph, and I couldn't find another slot that preserves that.

Perplexity `sonar-deep-research`, committed under `research/`. Its reference validation is the weakest of the five in this run: 16/16 identifiers resolved with zero confabulation, but **only 5 of 16 assessed on topic, with 3 flagged explicitly off topic**. Used as a lead source only; every snippet is anchored to a PMID I fetched and read. The two PMIDs actually cited here I found by PubMed search, not from the report.

## Validation

- `just validate` — **22/22 snippets verified**
- `just validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `validate-history` — pass
- `cache/` carries three rows (`MONDO:0957274`, `hgnc:463`); only the two cited PMIDs are included
- One label corrected against OLS: `HP:0001256` is "Mild intellectual disability"

---

### DIFF
--- a/history/disorders/Spastic_Paraplegia_89_Autosomal_Recessive/2026-08-28T131353Z-claude-code-dab7ee.yaml
+++ b/history/disorders/Spastic_Paraplegia_89_Autosomal_Recessive/2026-08-28T131353Z-claude-code-dab7ee.yaml
@@ -1,0 +1,36 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Spastic_Paraplegia_89_Autosomal_Recessive
+  path: kb/disorders/Spastic_Paraplegia_89_Autosomal_Recessive.yaml
+session:
+  id: 2026-08-28T131353Z-claude-code-dab7ee
+  timestamp: '2026-08-28T13:13:53Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9818
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9835
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - experimental_models
+  - treatments
+  summary: 'Review round 3: correct a REFUTE grading I introduced, bind the statins'
+  details: |
+    Two items from the round-2 approving review. The first is a regression I introduced in the previous commit, not a pre-existing gap.
+
+    1. The two ER-stress evidence items were graded REFUTE. That was wrong, and wrong in a way that would have exported a false claim. The supports value grades an evidence item against the claim it is attached to, and these items sit on a readout whose claim is its own measurement: ER stress response, direction UNCHANGED. The quoted sentences report exactly that measurement, so they support it. Graded REFUTE, they export as evidence refuting Blunted ERAD of HMGCR and INSIG-1 - a node the paper does not argue against but demonstrates. Both are now SUPPORT, with explanations recording the reasoning.
+
+    The distinction against the ORO case is worth keeping straight, because the two look alike and are not: the ORO null is REFUTE on a RESCUES link, where the link's own claim is that statins rescue the lipid phenotype, and the null does refute that claim. Here the link is MEASURES and its claim is that the measurement was made. Same null-result shape, opposite grading, because the claim being graded against differs. The 'this excludes generalised proteostasis failure' inference stays where it belongs, in the readout interpretation and the link limitations.
+
+    2. Bound atorvastatin (CHEBI:39548) and simvastatin (CHEBI:9150) as therapeutic_agent on the NCIT:C15986 treatment term. The entry already distinguishes the two drugs in three places - the description, the branching readout graded PARTIAL, and a proposed experiment - because atorvastatin corrected axon branching and simvastatin did not. Until now that distinction existed only in prose; binding both agents makes it machine-queryable, which is the point of recording it at all. Both CURIEs were already in cache/chebi/terms.csv and the chemicalentityterm enum cache, so this added no cache rows.
+
+    Validation on this head: just validate passes with 47/47 snippets verified; working tree clean afterwards, no cache churn.
--- a/kb/disorders/Spastic_Paraplegia_89_Autosomal_Recessive.yaml
+++ b/kb/disorders/Spastic_Paraplegia_89_Autosomal_Recessive.yaml
@@ -685,6 +685,15 @@ treatments:
     term:
       id: NCIT:C15986
       label: Pharmacotherapy
+    therapeutic_agent:
+    - preferred_term: atorvastatin
+      term:
+        id: CHEBI:39548
+        label: atorvastatin
+    - preferred_term: simvastatin
+      term:
+        id: CHEBI:9150
+        label: simvastatin
   target_mechanisms:
   - target: Blunted ERAD of HMGCR and INSIG-1
     description: >-
@@ -809,18 +818,24 @@ experimental_models:
       evidence:
       - reference: PMID:37119330
         reference_title: "AMFR dysfunction causes autosomal recessive spastic paraplegia in human that is amenable to statin treatment in a preclinical model."
-        supports: REFUTE
+        supports: SUPPORT
         evidence_source: IN_VITRO
         snippet: "Assessment of ER stress response did not show clear differences between wild type and AMFR KO NSCs, also upon treatment with tunicamycin"
         explanation: >-
-          Reports the null, including the challenge arm. Graded REFUTE because it argues
-          against the generalised-proteostasis reading of the mechanism.
+          Reports the null, including the challenge arm. Graded SUPPORT, not REFUTE: the claim
+          this item carries is the readout's own - ER stress was measured and was UNCHANGED -
+          and the quote reports exactly that. The paper does not argue against blunted ERAD of
+          HMGCR and INSIG-1; it demonstrates it. The "this excludes generalised proteostasis
+          failure" reading is carried by the readout interpretation and the link limitations,
+          which is where it belongs.
       - reference: PMID:37119330
         reference_title: "AMFR dysfunction causes autosomal recessive spastic paraplegia in human that is amenable to statin treatment in a preclinical model."
-        supports: REFUTE
+        supports: SUPPORT
         evidence_source: IN_VITRO
         snippet: "In patient-derived fibroblasts, and AMFR KO NSCs, no clear evidence for increased ER stress was found"
-        explanation: The authors' summary of the same null across both cell models.
+        explanation: >-
+          The authors' summary of the same null across both cell models. Graded SUPPORT on the
+          same reasoning as the item above - it reports the measurement the readout records.
     evidence:
     - reference: PMID:37119330
       reference_title: "AMFR dysfunction causes autosomal recessive spastic paraplegia in human that is amenable to statin treatment in a preclinical model."


matentzn:
All nine items addressed in b61bcb55f. Your root-cause line — *"the full text was in the commit and was not read"* — is exactly right, and it's the second time in this five-PR run, so it's worth writing down properly rather than just fixing.

## Why it happened twice

Perplexity's reference validation on these reports flagged low resolvable-identifier counts (4 of 19 on one, 5 of 16 on-topic here), so I adopted a rule: treat the report as a lead source, build from PubMed searches, use the report only to cross-check. That rule is sound against confabulation and it is what caught `HGNC:17762` and `CL:0000540` being wrong.

But it made me treat the *cache* as a lookup table for snippets rather than as a document to read. `references_cache/PMID_37119330.md` is 182 lines of full text. I read the abstract at the top, quoted from it, and never scrolled. Table 1 was at line 99; the ORO result at line 143.

The tell was there and I wrote it myself: an entry whose `notes` says "no cohort reports per-phenotype rates" for a paper with a 20-patient cohort in *Acta Neuropathologica* should have prompted one more look. Absence of a clinical table in a cohort paper is a claim that needs checking, not a default.

## The worst of it

**The discussion asserted a measurement had never been taken, and proposed taking it as the resolving experiment.** It had been taken. It was negative. It is in the paper I cited.

That is worse than a missing citation, because it's a confident negative claim about the state of evidence, embedded in the structure the KB is for. Anyone reading that discussion would have concluded the field hadn't looked.

Now modelled properly: `UNCHANGED` readout on the `RESCUES` link, graded `REFUTE`, with the `explanation` saying it refutes the specific claim that would have joined the drug to the mechanism. The discussion is a `KNOWLEDGE_GAP` about the *dissociation* — statins fix motor output with no detectable change in the lipid readout — and lays out three readings without choosing: ORO too coarse to see a sterol-level change (supported by ATV having the most pronounced effect on brain sterol ratios), an HMGCR-independent route, or droplets being a marker rather than the damaging agent. Proposed experiments are re-scoped to human cells with sterol profiling, plus one on why atorvastatin and not simvastatin.

## Item by item

**1. Frequencies.** Every one now derived from Table 1. Denominators kept honest — thin corpus callosum is 5/9 imaged, not 5/20; microcephaly 1/16. The zero rows are in the entry `notes` and the description, because 0% speech delay, regression, ataxia and bladder dysfunction is a real discriminator in an HSP differential, not missing data.

One mechanical note: the cache renders Table 1 as an unbroken run-on (`Epilepsy204/20`), so the clean prose sentence giving counts is what I quote, with the table cited in the `explanation` for rates it's the only source of.

**2. Seven phenotypes added.** All of them, with `HP:0033725` for thin corpus callosum as you noted HSP49 uses.

**4. ERAD framing.** You're right and I overstated it four times. The paper's conclusion is *"the loss of AMFR's ERAD function likely underlies the disease mechanism... leading to a disturbance in the balance of lipid and cholesterol homeostasis"* — ERAD *of lipid-regulatory substrates*, not a departure from ERAD. There's now a `Blunted ERAD of HMGCR and INSIG-1` node between the ligase and the lipid node, quoting the substrate sentence, and the statin treatment's `target_mechanisms` points at it — which is where it always belonged, since statins inhibit the substrate that accumulates.

**5. `DYSREGULATED`.** Changed. "Disturbance in the balance" is exactly what that value is for, and `DECREASED` was mismodelling it.

**6. `experimental_models`.** Both added with `modeled_mechanisms`. The RING-mutant arm is called out specifically — it's what separates "AMFR protein present" from "AMFR ligase active", which is a stronger claim than the plain rescue.

**7. Onset and progression.** Onset is `CHILDHOOD`/`INFANTILE` on the relevant phenotypes and there's a `progression:` section keyed on the 19/19 under-three-years figure. You're right it's the best-measured feature in the paper.

**8. Symptomatic treatments.** Genetic counselling added — all families consanguineous, and it carries the practical point that the index variant was missed on exome because AMFR exon 1 is poorly covered. Symptomatic spasticity management is added but deliberately thin, with a `notes` line saying why: the founding paper has no management section and no GeneReviews chapter exists, so importing general HSP guidance would attribute to this disorder advice written for a heterogeneous group.

**9. GeneReviews.** Ran the searches: nothing for `AMFR` or `spastic paraplegia 89`. The Uncomplicated (Pure) HSP Overview (PMID:20301682) exists but its abstract is a bare purpose statement with no minable content — same situation you recorded for the CMT overview on #9834. Recorded in `notes` rather than cited.

**Statin placement** — thanks for answering directly. Left in `treatments` with `target_mechanisms`.

## One near-miss worth flagging

While cleaning up uncited references I deleted `references_cache/PMID_20301682.md`. `git status` showed it as `D` rather than untracked — it was already tracked, and cited by four other entries. Restored. `references_cache/` is repository-wide and `rm` on a file you didn't fetch is not a local operation; I'd been checking that new files were cited but not that deletions were safe.

## Validation

43/43 snippets verified (was 22/22). `validate`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length` all pass.

---

matentzn:
Pushed `4e6cf383e`, taking four of the five non-blocking suggestions. This resets the approval; the ER-stress one in particular seemed worth another round.

**The two optimistic grades.** Both regraded `SUPPORT` → `PARTIAL`. You were right about the mechanics too — I checked, and the de-tabulated Table 1 substrings really are three tokens and really would trip `check-snippet-length`, so the compromise stands and only the grade moved. Both `explanation`s now say why the sentence supports less than the claim it carries.

**Prose residue.** Trimmed. The gp78 node description and the AMFR genetic note both still said the mechanism departs from what the gene's biology predicts, which contradicted the `description` I'd rewritten to say the opposite. Both now state the substrate-selective reading: the ERAD substrates that matter here are HMGCR and INSIG-1, which is why statins engage the mechanism at all.

**Behavioral concerns.** Added at 3/19 (15.8%), bound to `HP:0000708` Atypical behavior. Table 1 records it as a category without naming the behaviors, so the general term is the accurate binding rather than a manufactured narrower one. Graded `PARTIAL` on the same basis as Axial Hypotonia.

**The ER-stress negative — this is the one worth the round.** It's now structure rather than prose: a `MEASURES` link on `Blunted ERAD of HMGCR and INSIG-1` with an `UNCHANGED` readout and two `REFUTE`-graded items — *"Assessment of ER stress response did not show clear differences between wild type and AMFR KO NSCs, also upon treatment with tunicamycin"* and *"In patient-derived fibroblasts, and AMFR KO NSCs, no clear evidence for increased ER stress was found"*.

You framed this as optional, but it turned out to be load-bearing. The entry raises the proteostasis expectation and then answers it with a lipid mechanism; without this null, a reader has to take on trust that the obvious alternative was excluded. With it, the substrate-selective claim is the *conclusion* of a tested alternative rather than an assertion. The `limitations` record what the null does not cover: it excludes a gross UPR, not a subtler or compartment-restricted defect, and neither cell type is a mature corticospinal neuron.

**Not done: Orphanet.** `references_cache/ORPHA_*.md` can't be built right now — `just refresh-orphadata` fails its pinned checksum, which is open [#9687](https://github.com/monarch-initiative/dismech/issues/9687). Recording that rather than leaving it silently skipped.

**Validation on this head:** `just validate` passes with **47/47 snippets verified** (was 43/43); `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens` all clean. Working tree clean afterwards. History record added.

---

matentzn:
Pushed `4731ee965`. Both suggestions taken — and the first was a regression I introduced in the commit you just approved, so thank you for catching it in the same round rather than letting it ship.

**The `REFUTE` grading was wrong, and wrong in a way that would have exported a false claim.** You had it exactly right: `supports` grades an evidence item against *the claim it is attached to*, and these two items sit on a readout whose claim is its own measurement — ER stress response, direction `UNCHANGED`. The quoted sentences report precisely that measurement, so they support it. Graded `REFUTE`, they export as *evidence refuting `Blunted ERAD of HMGCR and INSIG-1`* — a node the paper demonstrates rather than argues against. Both are now `SUPPORT`, with the reasoning in the `explanation`.

Your parenthetical about the ORO case is the part worth preserving, because the two look identical and are not:

| | ER stress null | ORO null |
|---|---|---|
| Link relationship | `MEASURES` | `RESCUES` |
| The link's own claim | this measurement was made | statins rescue the lipid phenotype |
| Does the null contradict that claim? | no — it *is* the measurement | yes — the rescue did not happen |
| Correct grade | `SUPPORT` | `REFUTE` |

Same null-result shape, opposite grading, because the claim being graded against differs. I had generalised from the ORO case without re-asking what claim the new items were actually attached to. The "this excludes generalised proteostasis failure" inference stays where it belongs — in the readout `interpretation` and the link `limitations`, not in the grade.

**Statins bound.** `CHEBI:39548` atorvastatin and `CHEBI:9150` simvastatin as `therapeutic_agent` on the `NCIT:C15986` term. You're right that the distinction was load-bearing and prose-only: the entry separates the two drugs in three places — the description, the branching readout graded `PARTIAL`, and a proposed experiment — because atorvastatin corrected axon branching and simvastatin did not. Binding both makes that queryable, which is the point of having recorded it. Both CURIEs were already cached, so this added no rows.

**Validation on this head:** `just validate` passes with **47/47 snippets verified**; `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens` all clean, `validate-history` passes on the new record. Working tree clean afterwards — no cache churn.

I have nothing else outstanding on this entry.

---

PR #9836, opened 2026-08-28
curate: alpha-methylacyl-CoA racemase deficiency (AMACR, MONDO:0013681)

### LINKED ISSUE #9817: Curate alpha-methylacyl-CoA racemase deficiency (MONDO:0013681)
Curate a dismech entry for **alpha-methylacyl-CoA racemase deficiency** ([MONDO:0013681](https://monarchinitiative.org/MONDO:0013681)) — an autosomal recessive peroxisomal single-enzyme defect in which AMACR cannot convert (R)- to (S)-stereoisomers of 2-methyl-branched fatty acyl-CoAs, blocking peroxisomal beta-oxidation of pristanic acid and the C27 bile acid intermediates DHCA/THCA. Adults present with a sensorimotor neuropathy, seizures, or a relapsing encephalopathy; some patients present in infancy with cholestatic liver disease.

Single gene, single enzyme step, no MONDO descendants — this looks like a leaf Disease. It is currently mentioned in the KB only as a differential diagnosis inside `kb/disorders/Adult_Refsum_Disease.yaml` (a low-phytanic-acid atypical Refsum mimic), which is a cross-reference rather than coverage. The adult neurological and infantile cholestatic presentations look like phenotype variation on one mechanism rather than separate entries.

Stub: `stubs/Alpha-methylacyl-CoA_Racemase_Deficiency.yaml`

matentzn:
New entry for AMACR deficiency. Last of five in this claim run.

Closes #9817

> **Correction to an earlier version of this description.** It claimed the eight DOI-prefixed snippets were "verified by hand — 8/8 exact substrings". That was wrong; it was 6/8, and my verification script is what got it wrong. See "Two things reviewers should know" below. All snippets now cite a PMID and are machine-verified.

## One racemase, two substrate streams

Peroxisomal beta-oxidation is stereospecific — it only accepts the (S) isomer — so losing the racemase is a hard block, not a slowdown. And because AMACR gates *both* pristanoyl-CoA (from dietary phytol) and the C27 bile acid intermediates, one enzyme loss blocks two chemically unrelated streams. The entry is organised around that split, because the presentations follow it: the bile-acid stream gives the childhood hepatic picture, the pristanic-acid stream the adult neurological one.

## Diet is part of the mechanism, and the mouse proves it

`environmental` + `influences_mechanisms` carries dietary phytol, because the substrate driving the neurological arm is *eaten*. The Amacr-null mouse makes the dependence unusually explicit: null mice are clinically normal on standard chow and **all die within 36 weeks on phytol chow**. In that model the genotype alone is not a disease.

That is exactly the divergence from patients — who get ill on an ordinary diet — so it's recorded as a `HUMAN_MODEL_MISMATCH`. The cholic acid trial is a relevant negative there: pristanic acid stayed elevated under a treatment that corrected the bile-acid arm.

`exposure_term` is left unbound. ECTO has no phytol or phytanic-acid exposure term, and the nearest candidates name an *intervention* or an *avoidance* rather than an intake — `review_notes` records the search rather than binding something that would misstate the exposure.

The mouse is linked three ways: `RECAPITULATES` (metabolite accumulation), `PARTIALLY_RECAPITULATES` (hepatic), `FAILS_TO_RECAPITULATE` (the sensorimotor neuropathy that defines the human adult disease). That last `limitations` is honest about the confound — the mouse dies of liver failure and may simply not live long enough for a chronic neuropathy.

## The most useful clinical fact

From the 12-patient natural-history cohort: retinitis pigmentosa in 5/9 adults at median age 45, and a **median 24-year delay** (range to 33) between diagnosing the RP and diagnosing AMACR deficiency. An unexplained adult retinitis pigmentosa is a reason to measure pristanic acid. Median age at diagnosis in that cohort was 56.

## A framing error I made and corrected

I first recorded the hepatocellular carcinoma that occurred in the cholic acid trial as an unexplained adverse event sitting near the drug. The natural-history cohort shows HCC is a **disease** risk — it reports one too, present at diagnosis, and concludes patients need active monitoring for fibrosis, cirrhosis and HCC. HCC is now modeled as a phenotype downstream of a `Progressive Liver Fibrosis and Cirrhosis` node.

## Named Entity Confusion, confirmed the hard way

AMACR is far better known as the prostate-cancer marker P504S than as a disease gene. That's flagged in `genetic.notes` — and the deep-research validation step then fetched exactly those papers (prostate, and HCC AMACR overexpression) as candidate references. Both were discarded as uncited. The hazard is real, not theoretical.

## Two things reviewers should know about the validation

**The Perplexity run failed six times before succeeding.** "Server disconnected without sending a response", consistently at ~5m20s while the four successful runs in this batch finished under ~4m50s. It completed on the seventh attempt with `reasoning_effort=low`.

**A DOI-cited snippet is not machine-checked, and my manual substitute was faulty.** `conf/reference_validator_config.yaml` lists `DOI` under `skip_prefixes` — alongside GEO, dbGaP and other dataset accessions, under a comment describing them as "dataset accession identifiers, not literature references". A DOI is a literature reference, so eight snippets here were silently skipped while both `just validate` and `just validate-disorders` printed "All validations passed".

I noticed that and wrote a script to check them by hand. It reported 8/8 exact. It was 6/8 — the script had a fallback that retried each snippet with hyphens swapped for en-dashes and counted the retry as a hit, so it passed exactly the two mismatches it existed to catch. A manual check written by the person who made the mistake can encode the mistake.

Fixed at the root: everything now cites **PMID:39313810**, which puts these snippets through the normal gate. All 50 in the file verify.

## Validation

- `just validate` / `just validate-disorders` — pass, **50/50 snippets machine-verified**
- `check-entity-refs`, `check-duplicate-keys`, `check-environmental-evidence`, `check-snippet-length`, `validate-history` — pass
- Three ontology IDs corrected against OLS: `CHEBI:35044` is Ulifloxacin (→ `CHEBI:51340`), `HP:0010964` is a long-chain fatty acid term (→ `HP:0034721`), `HP:0007378` is a GI neoplasm term (→ `HP:0010663`)
- Re-fetching PMID:39313810 upgraded its cache from `abstract_only` to `full_text_xml`; `Inborn_Disorder_of_Bile_Acid_Synthesis` also cites it and was revalidated at 45/45

### DIFF
--- a/history/disorders/alpha-Methylacyl-CoA_Racemase_Deficiency/2026-08-28T132730Z-claude-code-4181b6.yaml
+++ b/history/disorders/alpha-Methylacyl-CoA_Racemase_Deficiency/2026-08-28T132730Z-claude-code-4181b6.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: alpha-Methylacyl-CoA_Racemase_Deficiency
+  path: kb/disorders/alpha-Methylacyl-CoA_Racemase_Deficiency.yaml
+session:
+  id: 2026-08-28T132730Z-claude-code-4181b6
+  timestamp: '2026-08-28T13:27:30Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9817
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9836
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - evidence
+  summary: Revert the unicode-space substitutions from the previous record
+  details: |
+    Corrects the immediately preceding history record (2026-08-28T132053Z-claude-code-dd66af.yaml), which is append-only and stands as written; this record supersedes its conclusion on one of its three classes.
+
+    That record described substituting the publisher's thin space (U+2009) and non-breaking space (U+00A0) into four HGVS snippets to make them byte-exact against the cache. I have reverted all four to ASCII spaces. The sentence-initial capital fix in the same commit is kept, so the net delta against the previously approved commit is one character: 'for all three' to 'For all three'.
+
+    The reviewer's round-3 analysis, which landed after that commit was pushed, settles the question against me on evidence I did not have. Three points, in order of weight:
+
+    First, byte-exactness is unattainable by construction. The reference caches hard-wrap at roughly 80 columns, so any snippet longer than one line necessarily contains a newline the published sentence does not. There is no version of the file in which all snippets are byte-exact, so byte-exactness cannot be the standard CLAUDE.md's 'exact source substring' refers to; it can only mean exact modulo the validator's normalize_text.
+
+    Second, the KB-wide census: 51,939 snippets byte-exact against 78,726 matching only after whitespace normalisation. ASCII transcription is not a lapse, it is the established convention across the knowledge base, and a handful of unicode-space snippets in one entry would be the anomaly.
+
+    Third, the cost I had not weighed: a snippet containing U+2009 is un-greppable. A curator searching the KB for 'c.154T > C' would not find it. That is a real loss of utility in exchange for a form of exactness that the first point shows is not achievable anyway.
+
+    What survives from the previous record is the part that mattered: the audit found one divergence that was not whitespace at all - a sentence-initial capital - which checking only the four reported instances would have missed. Auditing the whole file rather than the reported subset was the right method even though one of its conclusions was wrong.
+
+    Recording this because the previous record's Class 1 paragraph explicitly instructs a future curator not to 'tidy' the unicode spaces back to ASCII. That instruction is now withdrawn: ASCII spaces are correct here, and this record is the reason.
+
+    Validation on this head: just validate passes with 57/57 snippets verified; working tree clean, no cache churn.
--- a/kb/disorders/alpha-Methylacyl-CoA_Racemase_Deficiency.yaml
+++ b/kb/disorders/alpha-Methylacyl-CoA_Racemase_Deficiency.yaml
@@ -736,13 +736,13 @@ genetic:
       reference_title: "Redefining the phenotype of alpha-methylacyl-CoA racemase (AMACR) deficiency."
       supports: SUPPORT
       evidence_source: HUMAN_CLINICAL
-      snippet: "three patients were homozygous for the c.154T > C (p.Ser52Pro) variant"
+      snippet: "three patients were homozygous for the c.154T > C (p.Ser52Pro) variant"
       explanation: Names the variant and the number of homozygous carriers in the cohort.
     - reference: PMID:39313810
       reference_title: "Redefining the phenotype of alpha-methylacyl-CoA racemase (AMACR) deficiency."
       supports: SUPPORT
       evidence_source: HUMAN_CLINICAL
-      snippet: "Only the c.154T > C (p.Ser52Pro) variant and chromosome 5p13.3 deletion have been previously reported"
+      snippet: "Only the c.154T > C (p.Ser52Pro) variant and chromosome 5p13.3 deletion have been previously reported"
       explanation: >-
         Establishes that this allele is one of the two previously known lesions, which is what
         makes it prominent in secondary sources rather than more frequent.
@@ -757,13 +757,13 @@ genetic:
       reference_title: "Redefining the phenotype of alpha-methylacyl-CoA racemase (AMACR) deficiency."
       supports: SUPPORT
       evidence_source: HUMAN_CLINICAL
-      snippet: "Three patients were homozygous for the c.364 C > T (p.His122Tyr) variant"
+      snippet: "Three patients were homozygous for the c.364 C > T (p.His122Tyr) variant"
       explanation: Names the variant and its homozygous carrier count.
     - reference: PMID:39313810
       reference_title: "Redefining the phenotype of alpha-methylacyl-CoA racemase (AMACR) deficiency."
       supports: SUPPORT
       evidence_source: HUMAN_CLINICAL
-      snippet: "Only the c.154T > C (p.Ser52Pro) variant and chromosome 5p13.3 deletion have been previously reported"
+      snippet: "Only the c.154T > C (p.Ser52Pro) variant and chromosome 5p13.3 deletion have been previously reported"
       explanation: >-
         Establishes by exclusion that c.364C>T was novel to this cohort, which is why it is
         under-represented in secondary descriptions of the allele spectrum.


matentzn:
All eight IMPORTANT items and the six suggestions addressed in 100dfa86d. Finding 1 deserves more than a fix.

## You caught me laundering my own error

I flagged in the PR body that DOI snippets are skipped by the validator, wrote a script to check them by hand, and reported "8/8 exact substrings". It was 6/8.

The script:

```python
hit = b in cache
if not hit:
    hit = b.replace('-','–') in cache or b.replace('-','‐') in cache   # ← this
```

I added that fallback *because I anticipated dash trouble*. So the check reported OK for precisely the two mismatches it was written to catch. A manual verification written by the same person who made the mistake can encode the mistake — and it's worse than no check, because it launders the error while producing a number that reads like diligence. The PR body and the history record both carried "8/8" on that basis; the PR body is corrected, and since history is append-only, a new record corrects the old one rather than rewriting it.

**Fixed at the root, per your item 2.** Everything now cites `PMID:39313810`. All 50 snippets in the file are machine-verified, including the two that were wrong. Worth noting for anyone reading later: the PubMed abstract uses ASCII hyphens in those ranges where the publisher text uses en-dashes, so the snippets are genuinely exact against the PMID cache rather than passing loosely.

**On `skip_prefixes`:** `DOI` sits in that list between `dbgap` and `gtex`, under a comment reading *"These are dataset accession identifiers, not literature references, so snippet validation does not apply to them."* `DOI` and `MONDO` are the two entries that don't match their own justification. `CLAUDE.md` states the reference validator checks DOIs. I've noted it in [#9840](https://github.com/monarch-initiative/dismech/issues/9840) rather than changing the config in a curation PR, since removing `DOI` would put every existing DOI-cited snippet in the KB under validation at once and that diff needs its own review.

## Items 3–8

**3. `mappings`.** Added, `skos:narrowMatch` to MONDO:0008967 following the `Adult_Refsum_Disease` precedent. The justification says why not `exactMatch`: MONDO names the disorder from its bile acid defect and that concept is already the `BASD Type 4` subtype elsewhere in the KB, whereas this entry covers the whole enzyme deficiency including the pristanic-acid arm, which is not a bile acid phenotype at all.

**4. Pathograph → phenotypes.** Twelve edges added, and you were right that the retinal arm had no mechanism node at all despite RP being the leading adult presentation. There is now a `Retinal Photoreceptor Degeneration` node, placed downstream of phytanic/pristanic accumulation with `INDIRECT_UNKNOWN_INTERMEDIATES` — the Refsum parallel motivates the placement but the mechanism hasn't been shown in AMACR deficiency specifically.

**5. Duplication.** The pathophysiology node is now `Peripheral Nerve Axonal Degeneration` (TISSUE) pointing `DIRECT` at the `Sensorimotor Peripheral Neuropathy` phenotype. Same fix pattern you got on #9834.

**6. Gallstones.** Added, 2/12, quoting the sentence with both ages. The 15-year-old cholecystectomy is worth having — juvenile cholelithiasis is unusual enough to be a diagnostic prompt on its own.

**7. Liver fibrosis/cirrhosis.** Now its own node between hepatic injury and HCC, which also gives the HCC phenotype a mechanistic parent instead of leaving it floating next to the drug.

**8. Genetics.** `functional_impact_category: LOSS_OF_FUNCTION` added (on the pathophysiology node — `Genetic` rejects `genetic_context`). The c.154T>C variant is added **with an explicit limit**: I did not assert it is the commonest allele. That claim reaches me only through the deep-research report citing MedlinePlus, and in the cached cohort paper it appears inside a patient table the cache renders as `Malec.154T > Cc.` — unquotable. The variant's `description` says so. p.Leu107Pro I could not source from any cached PMID at all, so it is not in the entry.

## Suggestions

Cell types added (hepatocyte, photoreceptor); Seizure re-evidenced from PMID:35428665's patient rather than the mouse paper's background sentence, with the mouse sentence kept as corroboration graded `OTHER`; glucosuria added as `VERY_RARE` with a `PARTIAL` grade since it's one incidental finding; the description's mimicry claim corrected — two of the three original patients were mimics, the third was a child without neuropathy.

The pristanic/phytanic **ratio** and fever-as-trigger I left out: neither is stated in a cached source in a form I can quote, and after the events above I would rather leave a gap than reconstruct a claim.

## One more shared-file note

Re-fetching PMID:39313810 upgraded its cache from `abstract_only` to `full_text_xml` (PMC green OA). `Inborn_Disorder_of_Bile_Acid_Synthesis` also cites it, so I revalidated that entry — 45/45, unaffected, since the new text is a superset. Flagging because a `fetch-reference` on a tracked cache file is a repo-wide change, not a local one.

## Validation

50/50 machine-verified (was 35/43 + 8 hand-checked, of which 2 were wrong). `validate`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-environmental-evidence`, `check-snippet-length`, `validate-history` all pass.

---

matentzn:
Pushed `4f197b80a`. Both IMPORTANT items are fixed, and five of the seven suggestions are taken.

Your framing of the two blocking items is the useful part of this review: they are both **side-effects of a fix**. Re-fetching PMID:39313810 by PMID upgraded its cache from `abstract_only` to PMC full text, and that made one statement in the entry obsolete and contradicted another. Neither would have been caught by re-running any gate, because neither is a validation failure — they are claims about a source that stopped being true when the source got bigger.

**1. Zygosity.** Dropped `zygosity: HOMOZYGOUS` and `allele_type: VARIANT`. You're right on both counts: three of twelve are compound heterozygous and two siblings carry a whole-gene deletion, so neither slot describes the cohort, and `ZygosityEnum` has no `biallelic` member to narrow to correctly. Rather than leave a silent omission, `genetic_context` now carries a `description` saying which patients the enum can't express and why both slots are unset — otherwise the next curator "fixes" it back.

**2. The stale limitation note.** Fixed. The note said c.154T>C could only be reached through a run-on patient table, which was true of the abstract-only cache and false of the one committed in the same commit. The snippet is now *"three patients were homozygous for the c.154T > C (p.Ser52Pro) variant"*, graded `SUPPORT` rather than the weak `PARTIAL` stand-in.

Your generalisation is the one I want to record: **a claim about what a source does *not* contain has to be re-read whenever the source changes.** That is a second instance of the round-1 lesson — there, a verification claim outlived the check that produced it; here, a *negative* claim about a source outlived the source. Both are assertions about evidence that no gate can see.

Taking the rest of round-1 item 8 off the same sentence: **c.364C>T (p.His122Tyr)** is now curated. It is homozygous in three patients — the same number as Ser52Pro — and, unlike Ser52Pro, was novel to this cohort, which is exactly why secondary sources name Ser52Pro as *the* AMACR allele and omit this one. The **5p13.3 deletion** is curated as its own mechanism class (enzyme absent, not catalytically dead), with the contiguous-gene caveat that phenotype in those two siblings cannot be attributed to AMACR alone — the index sibling was ascertained through oculocutaneous albinism.

**Suggestions taken:** Gallstones has an incoming edge from `C27-Bile Acid Intermediate Accumulation`, graded `INDIRECT_KNOWN_INTERMEDIATES` because the biliary-composition step is established general bile-acid physiology, not something measured in AMACR deficiency. Coagulopathy rebound `HP:0001892` → `HP:0003256`; you're right that the quoted word is the cascade abnormality and bleeding is its consequence. Hepatic Fibrosis added at 1/12, with both the ascertainment caveat (all patients screened by ultrasound, most by FibroScan, so this is a screened rate) and the paediatric negative quoted. Thalamic MRI moved to `FREQUENT` — 9/12 is 75%, inside the band. The bare `Congenital bile acid synthesis defect 4` synonym is gone, since a synonym asserts precisely the identity `skos:narrowMatch` declines to assert.

**Not taken, with reasons:**

- **Glucosuria keeps no incoming edge.** It is an incidental screening finding in one child and no cited source proposes a mechanism. An edge would be inventing one.
- **Cirrhosis is not a phenotype.** The cohort states the fibrosis→cirrhosis→HCC risk as a surveillance rationale; no cohort patient is reported cirrhotic. It stays modelled at the mechanism node, where a risk belongs.
- **The seven upgraded cache files (suggestion 8) stay.** You verified them harmless and I re-verified `Inborn_Disorder_of_Bile_Acid_Synthesis` at 45/45. Reverting shared cache files to reduce diff noise carries more risk than the noise does. Agreed it is out-of-scope churn; the cause is that `fetch-reference` upgrades whatever it touches, and there is no way to fetch one reference without improving its neighbours.
- **Suggestion 9 noted, and it changes the story.** If `normalize_text` strips non-word characters before matching, then the en-dash difference was never what broke anything — the `DOI` entry in `skip_prefixes` was the whole defect, and my round-1 self-check was chasing a red herring *and* getting the red herring wrong. Worth knowing.

**Validation on this head:** `just validate` passes with **57/57 snippets verified** (was 50/50); `check-entity-refs`, `check-duplicate-keys`, `check-environmental-evidence`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens` all clean, and `validate-history` passes on the new record. Working tree clean after validation — this commit added no cache rows, since `HP:0003256` and `HP:0001395` were both already cached.

---

matentzn:
Pushed `764abe836`, acting on your in-progress note about whitespace substitutions before the review landed.

Rather than fix the four you found, I ran the same strict check across **all 57 snippets** and classified every divergence. That turned out to matter: **42 byte-exact before, 47 after, and one of the divergences was not whitespace at all** — so checking only what was reported would have left it.

| Class | Count | Action |
|---|---|---|
| Typographic spaces in HGVS notation | 4 | **fixed** — all mine, added last round |
| Sentence-initial capital | 1 | **fixed** — the one non-whitespace divergence |
| Line-wrap newlines in the cache rendering | 10 | **deliberately left** — see below |

**The HGVS spaces.** The source writes the variants with a thin space around the operator and a non-breaking space after the position: `c.154T`U+2009`>`U+2009`C`, and `c.364`U+00A0`C`U+2009`>`U+2009`T`. I typed ASCII spaces. Now using the source characters. Worth flagging loudly because they are *invisible in an editor* — a later curator who tidies them back to plain spaces would be reintroducing the divergence, not fixing a typo, so the history record names the codepoints.

**The one you would not have found from the four.** `PMID:34267495`'s snippet began `for all three` where the source sentence begins `For all three` — lowercased at some point to read as an embedded clause. Not whitespace, so a whitespace-normalising comparison misses it entirely. That is the argument for auditing the whole file rather than the reported instances.

**The ten I am not changing, and why.** The cached markdown wraps long paragraphs, so a sentence reads `retinitis pigmentosa \nwas detected` where the published sentence has a single space. Quoting that as a space is the *faithful* transcription — the newline is an artifact of how the cache file is rendered, not a character in the paper. Embedding a raw newline in a YAML snippet to match it would make the quote less faithful, and would break the moment the cache is regenerated with different wrapping. I have written the reasoning into the history record so this class does not get re-reported as a defect each round.

**The gate point, which is the interesting one.** None of the 14 was visible to anything: `normalize_text` collapses all whitespace before substring matching, so the file verified **57/57 both before and after**. That is precisely the blind spot behind the round-1 en-dash finding — and your round-2 observation that the dashes never actually blocked anything is the same fact from the other side. The reference validator answers *"is this quote in this source, modulo formatting"*, which is the right question for catching fabrication and the wrong one for catching transcription drift. Both are worth knowing; only the first has a gate.

**Validation on this head:** `just validate` passes with **57/57 snippets verified**; `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence` all clean; `validate-history` passes on the new record. Working tree clean — no cache churn.

---

matentzn:
Pushed `3c6bdca15`, which **reverts most of my previous commit**. Your round-3 analysis landed while `764abe836` was in flight, and it settles the exactness question against me on evidence I did not have.

Your three points, in the order that actually decided it:

**Byte-exactness is unattainable by construction.** The caches hard-wrap at ~80 columns, so any snippet longer than one line necessarily contains a newline the published sentence does not. There is no version of this file — or any file — in which every snippet is byte-exact. So "exact source substring" cannot mean byte-exact; it can only mean exact modulo `normalize_text`. That reframes the whole thing: I was chasing a target that does not exist.

**The census settles the convention.** 51,939 byte-exact against 78,726 matching only after normalisation. ASCII transcription is what the KB does. Four unicode-space snippets in one entry would have been the anomaly, not the correction.

**And the cost I had not weighed at all: greppability.** A snippet containing U+2009 cannot be found by a curator searching for `c.154T > C`. That is a real, recurring loss of utility, traded for a form of exactness the first point shows is unreachable. That argument alone would have been enough.

So all four unicode-space substitutions are reverted. I kept the sentence-initial capital, since `For all three` matches the source and costs nothing either way — **the net delta against the approved `4f197b80a` is now one character.**

What survives from the audit is the method, not that conclusion: checking all 57 snippets rather than the four reported is what surfaced the capital, which is not whitespace and which a normalising comparison cannot see. Auditing the whole file was right even though one of its conclusions was wrong.

One housekeeping note, because history records are append-only: the previous record's Class 1 paragraph explicitly instructs a future curator *not* to tidy those spaces back to ASCII. That instruction is now wrong. The new record withdraws it by name and gives the reasoning, so the pair reads correctly in sequence rather than the stale instruction sitting there unqualified.

**Validation on this head:** `just validate` passes with **57/57 snippets verified**; `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens` all clean; `validate-history` passes on the new record. Working tree clean — no cache churn.

Your two remaining suggestions I've deliberately not actioned, and I'd rather say why than leave them silently open. The `Ataxia`/`Cognitive Decline` reparenting is a genuine judgement call and I think you're probably right — the evidence sentence is the chronic adult arm, not episode sequelae — but it is a structural change to the pathograph on an entry that has now been through four rounds, and it isn't a defect. The `parents:` addition is the same shape: the relationship is already recorded and machine-findable in `mappings`. Both are better as a follow-up than as a fifth round here. Happy to be told otherwise.

---

PR #9841, opened 2026-08-28
curate: Polycystic Kidney Disease 2 (PKD2, MONDO:0013131)

### LINKED ISSUE #9837: Curate polycystic kidney disease 2 (MONDO:0013131)
Curate a dismech entry for **polycystic kidney disease 2** ([MONDO:0013131](https://monarchinitiative.org/MONDO:0013131)) — the PKD2/polycystin-2 form of autosomal dominant polycystic kidney disease, in which loss of the TRPP2 channel drives progressive renal cyst growth with a substantially later onset and slower decline than the PKD1 form.

**Preflight already done.** `kb/disorders/Autosomal_Dominant_Polycystic_Kidney_Disease.yaml` exists and is bound to the *parent* term MONDO:0004691. It carries **no `has_subtypes` block**, so PKD2 is not currently covered anywhere in the KB, and MONDO:0013131 appears in no entry. The proposal is a distinct `kb/disorders/` Disease: PKD2 has its own MONDO leaf term, its own causal gene (`hgnc:9009` PKD2), and a genuinely different natural history (median ESKD roughly two decades later than PKD1), which is the clinically actionable distinction.

Stub: `stubs/Polycystic_Kidney_Disease_2.yaml`

aleixpuigb:
Adds `kb/disorders/Polycystic_Kidney_Disease_2.yaml` for PKD2-related ADPKD.

**Why this is a new entry and not a duplicate.** `kb/disorders/Autosomal_Dominant_Polycystic_Kidney_Disease.yaml` is bound to the parent term MONDO:0004691 and carries **no `has_subtypes` block**, so PKD2 was not covered anywhere in the KB — MONDO:0013131 appears in no entry on `main`. PKD2 has its own MONDO leaf term, its own causal gene, and a clinically actionable difference in natural history.

**What is specific to this entry.** The mechanism is shared with PKD1 and the entry says so — four pathophysiology nodes conform to the `renal_cystogenesis` module. The substance here is the *timing*:

| | PKD2 | PKD1 |
|---|---|---|
| Median age at ESRD | ~78 y | ~54 y |
| ESRD by age 60 | 9.8% | — |
| Hypertension vs PKD1 | OR 0.25 | — |
| Haematuria vs PKD1 | OR 0.59 | — |

A somatic second-hit node is added to account for why cyst formation is focal on a heterozygous background.

**Evidence discipline — the part worth reviewing.** PKD2-specific cohort numbers are used wherever they exist, in preference to ADPKD-wide figures. Two claims could only be sourced ADPKD-wide, and both are graded `PARTIAL` with the reason stated in the explanation rather than presented as measured PKD2 findings:

- **Intracranial aneurysm frequency** — GeneReviews reports a fivefold elevation for ADPKD without stratifying by genotype.
- **Tolvaptan efficacy** — TEMPO 3:4 enrolled ADPKD patients and reported no PKD2 subgroup. Since most PKD2 patients progress slowly, comparatively few meet the rapid-progression criteria the trial's population was calibrated on.

Two knowledge gaps are recorded: whether tolvaptan benefit transports to PKD2, and whether any PKD2 allelic effect on renal outcome is reproducible — the two largest genotype-phenotype studies disagree, and Genkyst's own intraclass analysis attributes 34.1% of eGFR variance to pedigree but only 7.1% to the shared variant.

**Deep research.** One `openscientist` report, committed under `research/`. One thing reviewers should know: **the report shipped with no `reference_validation` block at all**, despite the run passing `--validate-references`, and nothing in the report signals that. Validation was retro-fitted with `just validate-research-reference` and came back clean — 47/47 resolved, 1/1 quotes valid, 46/47 on topic. `just preflight-dr … MONDO:0013131` returned `WARN` for rival gene PKD1 at 41% of PKD2 mentions; I assessed that as a false positive, since PKD2 is defined clinically by contrast with PKD1 and that comparison is the point of the entry.

**Validation**

- `just validate` — passes, **38/38 snippets verified** against cached references
- `just check-entity-refs` — clean
- `just check-duplicate-keys` — clean
- `just validate-history` — clean

Deletes `stubs/Polycystic_Kidney_Disease_2.yaml`. `references_cache/` carries only the four PMIDs this entry newly cites; four further PMIDs fetched while exploring were removed rather than swept in.

Closes #9837

---
_Generated by [Claude Code](https://claude.ai/code/session_013f2JB3rJzdbjprkmvfeMTZ)_

### DIFF
--- a/kb/disorders/Polycystic_Kidney_Disease_2.yaml
+++ b/kb/disorders/Polycystic_Kidney_Disease_2.yaml
@@ -536,14 +536,13 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       PKD2 patients were less likely to have hypertension (odds ratio 0.25 [95%
-      CI 0.15-0.42]), a history of urinary-tract infection
+      CI 0.15-0.42]), a history of urinary-tract infection (0.50 ... or
+      haematuria (0.59
     explanation: >-
-      The same sentence continues to give haematuria an odds ratio of 0.59, but
-      the reference validator strips bare bracketed spans such as [0.35-0.98]
-      before matching, so the clause naming haematuria cannot be quoted
-      verbatim. The quoted portion establishes the head-to-head design and the
-      direction of the PKD2-versus-PKD1 comparison; the haematuria value is
-      given in the description above.
+      Head-to-head comparison establishing that haematuria is significantly less
+      frequent in PKD2 than PKD1, odds ratio 0.59. The ellipsis bridges a
+      confidence interval the reference validator strips before matching; the
+      surrounding text is quoted verbatim.
 - category: Renal
   name: Cyst infection
   frequency: OCCASIONAL
@@ -594,15 +593,18 @@ phenotypes:
       failure in PKD2.
 - category: Renal
   name: Decreased glomerular filtration rate
-  frequency: OCCASIONAL
+  frequency: FREQUENT
   description: >-
-    GFR does decline in PKD2, but late and in a minority at any given age. Most
-    patients still have preserved function at diagnosis - 81.4% had an eGFR
-    above 60 mL/min/1.73 m2 - and the proportion above 30 mL/min/1.73 m2 is
-    still 73.1% in the sixth decade. The frequency is graded OCCASIONAL rather
-    than FREQUENT for that reason: at the point most PKD2 patients are seen,
-    reduced GFR is the exception. Men have lower kidney function than women at
-    equivalent age, without a corresponding difference in renal survival.
+    GFR declines late in PKD2 but in most patients who live long enough. The
+    grading is lifetime, not cross-sectional, which matters here because the two
+    give opposite answers: at diagnosis 81.4% still have an eGFR above 60
+    mL/min/1.73 m2, yet the proportion above 30 mL/min/1.73 m2 falls to 73.1%,
+    58.8% and 43.5% across the sixth, seventh and eighth decades. FREQUENT is
+    also the only grading consistent with Stage 5 chronic kidney disease being
+    FREQUENT below, since stage 5 is a subset of reduced GFR and a subset cannot
+    be commoner than the set that contains it. Men have lower kidney function
+    than women at equivalent age, without a corresponding difference in renal
+    survival.
   phenotype_term:
     preferred_term: Decreased glomerular filtration rate
     term:
@@ -649,13 +651,12 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       PKD2 patients were less likely to have hypertension (odds ratio 0.25 [95%
-      CI 0.15-0.42]), a history of urinary-tract infection
+      CI 0.15-0.42]), a history of urinary-tract infection (0.50 ... or
+      haematuria (0.59
     explanation: >-
-      Establishes that urinary-tract infection is significantly less frequent in
-      PKD2 than PKD1. The quote stops before the odds ratio itself: the
-      reference validator strips bare bracketed spans such as [0.31-0.83] before
-      matching, so the full sentence cannot be quoted verbatim. The value 0.50
-      is given in the description above.
+      Gives the PKD2-versus-PKD1 odds ratio for urinary-tract infection, 0.50.
+      The ellipsis bridges a confidence interval the reference validator strips
+      before matching; the surrounding text is quoted verbatim.
 - category: Renal
   name: Nephrolithiasis
   frequency: OCCASIONAL


aleixpuigb:
Thanks — this was a genuinely useful review. Pushed `3ed9e030b`. Every item is addressed; two of your suggested quotes turned out not to be quotable, and I've said why below rather than silently substituting.

### Blocking

**1. `check-snippet-length`** — fixed, both now quote the full sentence. The root cause is worth naming: I ran `just validate`, `check-entity-refs` and `check-duplicate-keys` before pushing but **not** the four baseline gates. All six now pass on all three PRs in this run.

### Evidence corrections — you were right on all three

**2. `Preclinical studies` graded `HUMAN_CLINICAL`.** Correct, and it's the clearest error in the entry. It's the TEMPO 3:4 background describing other groups' animal work. All three copies regraded `MODEL_ORGANISM`. Your note that all three being *identically* wrong is what let it past `check-snippet-grading` is the interesting part — that gate catches divergence, so a uniformly wrong grade is invisible to it.

**3. Decreased GFR.** Fixed, and it changed the phenotype. The sentence you pointed to describes *preserved* function in 81.4% of patients, which — as you noted — sits badly under `FREQUENT`. I dropped it to `OCCASIONAL` and rewrote the description: at the age most PKD2 patients are seen, reduced GFR is the exception. Catching a wrong frequency by way of a wrong citation is a nice result.

**4. cAMP node.** Fixed with `PMID:23736843`, which the `renal_cystogenesis` module already cites — so this was available to me and I didn't use it. It now carries both the elevation itself and the fact that blocking V2R lowers cystic-tissue cAMP, which is the interventional half.

### Completeness — all added

`diagnosis` (unified ultrasound criteria), GeneReviews counselling/agents-to-avoid/surveillance, ACEi-ARB treatment, `clinical_trials` (TEMPO 3:4 + REPRISE), UTI and nephrolithiasis phenotypes, `progression` including the 44% concomitant-nephropathy finding, and `animal_models` + `experimental_models` with `modeled_mechanisms` links.

Item 6 was the one that stung most and you were right to lead with it: the unified criteria exist *because* PKD1-derived criteria under-perform in PKD2 through reduced sensitivity. For an entry whose whole argument is "PKD2 differs in ways that change management", omitting a genotype-driven diagnostic threshold was the wrong thing to leave out.

Item 11's tolvaptan datum is now in the knowledge gap — only 25 of 293 Genkyst patients met rapid-progression criteria. That turns the gap from an inference into a PKD2 measurement.

### Two quotes I could not use

The Hateboer odds-ratio sentence you recommended twice **does not validate**. The reference validator strips bare bracketed spans before matching, so `a history of urinary-tract infection (0.50 [0.31-0.83])` becomes `(0.50 )` and the match fails. `[95% CI 0.15-0.42]` survives — presumably a `literal_bracket_patterns` entry — but `[0.31-0.83]` and `[0.35-0.98]` do not.

So both snippets stop before the stripped spans, the odds ratios are stated in the descriptions, and each `explanation` records the reason. I did not edit `conf/reference_validator_config.yaml` to force it through — that seemed like the wrong lever for one entry.

### Suggestions

12, 14, 16 applied. **13** — agreed, `HP:0012330` Pyelonephritis is a different entity, and the description was already arguing they differ, so the term contradicted its own prose; `term:` dropped, free-text `preferred_term` kept. **15** — `progression` added. **17** — you're right, thank you; the parent does carry a `genetic:` entry for PKD2. My PR body overstated it. What was absent is the PKD2-specific natural history, and I've kept the claim to that.

### The lump/split call

Taking this seriously rather than deflecting, because your framing is the stronger one and I want to say exactly where I get off the train.

You're right that the mechanism is shared, that all four nodes conform to the same module nodes as the parent, and that ClinGen curates PKD2 against MONDO:0004691. If the criterion is "a discrete pathomechanism", PKD2 fails it — this is one pathomechanism at a different tempo, and I say so in the entry.

Where I land differently is that the timing difference isn't a quantitative footnote, it's the thing that changes what you do:

- Median ESRD ~78 y vs ~54 y — for many PKD2 patients, kidney failure is not the thing that kills them
- **The diagnostic criteria are literally different**, and were derived because the PKD1 ones fail in PKD2
- 25 of 293 patients meet tolvaptan rapid-progression criteria
- Early ESRD in PKD2 should trigger a search for a second diagnosis — 44% of such patients had one

A `has_subtypes` row carries `description` plus `subtype:`-scoped phenotypes/prevalence/progression. It cannot carry a `diagnosis` section, a `clinical_trials` section, or model links — and after this round the entry has all three, each PKD2-specific.

That said, I hold this loosely, and it's a genuine cost that the pathograph is now duplicated across two entries. If the project prefers the subtype disposition I'd rather fold it now than after merge. Tagging @cmungall for a ruling, as you suggested.

### Validation after the fixes

```
just validate                  ✓ 63/63 snippets verified
just check-snippet-length      ✓ clean   (was ✗ — the blocking gate)
just check-title-snippets      ✓ clean
just check-snippet-grading     ✓ clean
just check-folded-hyphens      ✓ clean
just check-entity-refs         ✓ clean
just check-duplicate-keys      ✓ clean
just validate-history          ✓ clean
```

One housekeeping note: re-fetching `NCT00428948` rewrote its cache file with cosmetic churn only (quote style, a `full_text_attempted` key, trailing newline; body byte-identical). I reverted it rather than ship the noise.

Four tooling gaps found during this run are filed separately: #9844, #9845, #9846, #9847.

---

aleixpuigb:
Pushed `60d4c2c25`. Both items fixed.

### 1 — you were right, and I was wrong in the KB itself

I checked `_split_parts` (`reference_snippet_audit.py:554`) before touching anything: it splits on `\s*\.{2,}\s*` and matches each part independently, exactly as you said. Your ellipsis form validates at 63/63.

So my `explanation` text asserting the clause "cannot be quoted verbatim" was false, and it shipped. I got the mechanism right — brackets are stripped — and then drew the wrong conclusion from it and wrote that conclusion into the knowledge base as if it were a fact about the tooling. **That's the worse half of this finding**, and I'm glad you insisted on correcting the explanations rather than just the snippets: a future curator hitting the same wall would have read my note and worked around it unnecessarily.

Both items now quote through `or haematuria (0.59`, and the urinary-tract-infection phenotype cites its own OR 0.50 directly.

The thing worth noting for the record: the truncated snippet **verified perfectly and supported nothing**. It stopped one clause short of the word "haematuria" while attached to a haematuria claim. No gate catches that — `check-snippet-length` passed it because the truncation was still long enough, and nothing compares what a snippet says against what it's cited for. That's the same failure shape as the stenting problem on #9843.

Filed as #9848, per your suggestion — with the ellipsis convention, the misleading error message, and a proposal that a bare numeric confidence interval should probably be `literal_bracket_patterns` by default, since it's source text rather than an editorial gloss.

### 2 — the frequency inconsistency

Correct, and it's incoherent independent of the numbers: Stage 5 CKD is a subset of reduced GFR, so grading the subset `FREQUENT` and the superset `OCCASIONAL` cannot be right whatever the underlying data say. Now `FREQUENT`.

The underlying error was mine — I read a cross-sectional at-diagnosis figure as a lifetime frequency. `FrequencyEnum` is over patients who ever manifest the feature, and the same paper gives eGFR >30 in 73.1% / 58.8% / 43.5% across the sixth to eighth decades. The at-diagnosis nuance is kept in the `description`, where it's informative rather than driving the enum.

No apology needed for the churn — your previous comment pointed at the right sentence, and the frequency call was mine to get right.

### Suggestions

Not taken this round, deliberately:

- **`conforms_to` coverage** — the cyst-expansion node does span two module nodes. Splitting it is the right fix rather than adding a second `conforms_to`, and that's a structural change I'd rather not make in a round that's otherwise two narrow corrections. Happy to do it if you'd like it before approval.
- **Distal pathograph join** — same reasoning, though I note I *did* add these on #9842 when it was raised there. If you want consistency across the run, say so and I'll add them here too.
- **Kidney replacement therapy** term mismatch — real; the entry describes dialysis and transplantation but binds only the transplantation term.
- **Cardiac valvular / LVH** — covered by the PKD2-specific scope note, but I'll add an explicit line to `notes` if you'd prefer the omission stated rather than inferred.

Tell me which of these you want and I'll fold them in; otherwise this is my last substantive round on the entry.

### Validation

```
just validate                  ✓ 63/63 snippets verified
just check-snippet-length      ✓ clean
just check-snippet-grading     ✓ clean
just check-entity-refs         ✓ clean
```

---

PR #9849, opened 2026-08-28
review(Alzheimer): add missing cell-biology mechanisms and current hypotheses

cmungall:
Cell-biology review of the Alzheimer Disease entry. The entry modeled
amyloid, tau, glial, vascular, glymphatic, viral and adaptive-immune
mechanisms in depth but had no representation of several cellular
mechanisms with strong, largely human evidence. Notably, Synaptic
Dysfunction — the change that correlates best with cognitive decline —
had no cellular mechanism for how synapses are actually lost, and the
entry had no terminal cell-death node despite describing parthanatos.

Adds seven mechanistic_hypotheses groups and nine pathophysiology nodes:

- Complement-Mediated Microglial Synapse Elimination (C1q/C3/CR3)
- Neuronal Endosomal-Lysosomal Acidification Failure (APP-bCTF/v-ATPase,
  enlarged Rab5 endosomes, PANTHOS) — the "inside-out" amyloid model
- Selective Vulnerability of RORB+ Entorhinal Excitatory Neurons
- Necroptotic Neuronal Death (RIPK1/RIPK3/MLKL in GVD bodies)
- Senescent Cell Accumulation
- Oligodendrocyte and Myelin Dysfunction
- Microglial Lipid Droplet Accumulation (ACSL1, APOE4/4)
- Interneuron Dysfunction and Network Hypersynchrony (Nav1.1/SCN1A)
- Tau-Induced Nucleocytoplasmic Transport Failure (Nup98)

Declares conformance to the disabled_macroautophagy, cellular_senescence
and cns_myelin_failure modules. The myelin node also gives the entry's
existing oligodendrocyte precursor cell proteomic age gap biomarker a
mechanism to attach to, and the RORB node supplies the selective
vulnerability premise the HSV-1 node was relying on implicitly.

Also adds a Seizures phenotype (HP:0001250), human astrocyte-reactivity
evidence on Neuroinflammation, meningeal-lymphatic evidence on
Glymphatic Clearance Failure, and three discussions recording open
questions: which cell type is senescent in human brain (mouse studies
say glia, the largest human survey says >97% excitatory neurons),
whether plaques originate inside neurons or from secreted peptide, and
the complement-versus-MFG-E8 mouse-human opsonin mismatch.

Fixes two dangling free-text `consequences` on PARP1-Mediated
Parthanatos ("Neuronal Cell Death", "Mitochondrial Dysfunction") that
named no node in the entry.

All 27 new references fetched via `just fetch-reference`; every snippet
verified as an exact substring of its cached source (333/339 checked,
6 skipped by prefix). Schema, term, entity-ref and duplicate-key checks
pass.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01UTs2Aa1qpr7oww9VmhrH4U

### DIFF
--- a/kb/disorders/Alzheimer_Disease.yaml
+++ b/kb/disorders/Alzheimer_Disease.yaml
@@ -2957,13 +2957,15 @@ pathophysiology:
   - reference: PMID:37652017
     reference_title: "Human astrocytes and microglia show augmented ingestion of synapses in Alzheimer's disease via MFG-E8."
     supports: PARTIAL
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: IN_VITRO
     snippet: "Inhibiting interactions of MFG-E8 rescues the elevated engulfment of AD synapses by astrocytes and microglia without affecting control synapse uptake."
     explanation: >-
-      The human counterpart of the rescue experiment, but through MFG-E8 rather
-      than complement, and with astrocytes participating alongside microglia. The
-      cell biology of glial synapse ingestion is confirmed in human tissue; the
-      complement pathway specifically is not.
+      The counterpart of the rescue experiment, but through MFG-E8 rather than
+      complement, and with astrocytes participating alongside microglia. Graded
+      IN_VITRO because this sentence reports the cultured-cell arm — a phagocytosis
+      assay in primary human astrocyte and microglia cultures with a blocking
+      antibody. The same study's postmortem-tissue observation is curated
+      separately as HUMAN_CLINICAL on the hypothesis group.
   downstream:
   - target: Synaptic Dysfunction
     description: >-
@@ -3922,8 +3924,8 @@ pathophysiology:
   - preferred_term: Inhibitory synaptic transmission
     modifier: DECREASED
     term:
-      id: GO:0050804
-      label: modulation of chemical synaptic transmission
+      id: GO:0051932
+      label: synaptic transmission, GABAergic
   locations:
   - preferred_term: Hippocampus
   - preferred_term: Cerebral Cortex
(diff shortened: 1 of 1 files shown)


cmungall:
Both review findings are fixed, plus the two suggestions, and the failing check is addressed. Pushed in bb8bf1a.

**The `evidence_source` catch was right, and I had it backwards.** The MFG-E8 sentence from PMID:37652017 follows "In culture," in the abstract — it is a phagocytosis assay in primary human astrocyte and microglia cultures with a blocking antibody, not a postmortem observation. Regraded to `IN_VITRO`, and the explanation now states why. The same paper stays `HUMAN_CLINICAL` on the hypothesis group for its postmortem-tissue sentence, so the split now falls where it should.

**Missing titles added.** Both new trial-registration items (NCT03486938, NCT04063124) now carry `reference_title`, copied verbatim from the cache including the source's own "AGB101on" with no space.

**Both suggestions taken.** `GO:0050804` was a regulation-of term standing in for a claim about the transmission itself, so it is now `GO:0051932` (synaptic transmission, GABAergic), which is more specific and already cached. `target_phenotypes` added to the SToMP-AD trial; HOPE4MCI already had one.

**On the failing `test (3.13)` check:** it was not a test failure. The curation-scope guard rejected one file, `templates/hypothesis_deep_research_datasets.md`, because `templates/` is not in the curation allowlist. I have added the `scope-override` label, which is what the guard asks for when the change is intentional.

It is intentional, and worth explaining rather than just labelling. That template is what generated the seven deep-research reports in this PR — it takes the standard hypothesis variables plus a `candidate_datasets` block, and requires each report to state, per dataset, whether the data can address the hypothesis at all, the concrete analysis, and the discriminating prediction stated in advance. Dropping it from this PR would leave those seven reports with no in-repo record of how they were produced. The alternative is to split it into its own infrastructure PR; happy to do that instead if you would rather keep this one strictly curation-scoped — just remove the label and say so.

All deterministic gates re-run clean after the fixes: `validate` exit 0, 362/368 snippets verified with 6 prefix-skipped, plus entity-refs, duplicate-keys, folded-hyphens, snippet-grading, title-snippets and snippet-length.

---

Issue #9729, opened 2026-08-27
OpenScientist CMTX report suggests MONDO:0010674 (Hunter syndrome) as the CMTX1 term


Phillip-a-richmond:
`research/Charcot-Marie-Tooth_Disease_Type_X-deep-research-openscientist.md` recommends the wrong MONDO ID for CMTX1, in two places:

- line 620, in the disease-information table: `| MONDO | **MONDO:0010674** |`
- line 816, in the ontology-suggestions summary: `**Ontology term suggestions (summary).** MONDO:0010674; OMIM 302800.`

```
MONDO:0010674   mucopolysaccharidosis type 2          <- Hunter syndrome
MONDO:0010549   Charcot-Marie-Tooth disease X-linked dominant 1   <- correct
```

Both are already in `cache/mondo/terms.csv`, so this is verifiable offline in one grep.

Note the OMIM number on the same line (302800) *is* correct for CMTX1 — so the line is internally inconsistent, with a right OMIM ID sitting next to a MONDO ID for an unrelated lysosomal storage disease.

## Why it needs recording rather than just knowing

This passes every automated check we have. The CURIE is real, resolves in MONDO, and is a valid member of the `diseaseterm` dynamic enum. `just validate-terms` compares the `label` you write against the ontology's canonical label — so it catches the error only if the curator writes the *CMTX1* label alongside the *Hunter syndrome* ID. A curator who copies the ID and lets the label follow from the ontology gets a clean validation run and a silently wrong binding.

It is the fourth Named Entity Confusion found in this one report. The others:

| What | Caught by |
|---|---|
| PMID:42017539 cited for a serum-NfL claim in CMTX — it is a *NEFL*-CMT paper | curation, before use |
| MONDO:0010674 offered as the CMTX1 term | review |
| the ">90% of males" frequency, present only in the report's own summary table with no traceable source sentence | curation |

The pattern worth noting: the report's *citations* were validated (26/26 verified, 0% confabulation), but its *ontology suggestions* and *summary tables* are not validated by anything. Those are exactly where three of the four errors are. A report can be citation-clean and still carry a wrong CURIE in a summary table.

## Suggested actions

1. Nothing to fix in the report itself — it is a provider artifact and should not be hand-edited.
2. `MONDO:0010549` is the correct CMTX1 binding for whenever subtype `disease_term` bindings are added (none of the four CMT compartment entries bind subtype terms today, so this is a house-wide gap rather than a regression).
[...]

Issue #8432, opened 2026-08-13
Adopt deep-research-client 0.2.9: validate report references during generation, not after


caufieldjh:
## Problem

Right now, checking whether a deep-research (DR) report's citations are real is a
wholly separate, after-the-fact chore. `just research-disorder <provider> <Disorder>`
drops a report in `research/`, and the curator then has to hand-pick every PMID/DOI
out of it, run `just fetch-reference` on each one, quote a snippet into the KB YAML,
and only *then* discover — via `just validate-references` — that a citation was
confabulated or a quote was invented. By that point the entry is already half-built
around a source that does not exist.

The dismech anti-hallucination SOP (CLAUDE.md §2a) says as much: DR outputs are
*leads*, not ground truth, and every identifier/snippet must be independently
verified. That's still true. But the verification currently happens as late as
possible, per-claim, instead of once at report-generation time.

## What changed upstream

`deep-research-client` **v0.2.9** (we pin `>=0.2.7`) can validate a report's
references as part of generating it. It's backed by `linkml-reference-validator` —
the same library dismech already uses — via a new `validation` extra:

```
deep-research-client[validation]  ->  linkml-reference-validator>=0.2.1
```

Two entry points:

1. **Inline, during generation** — `research ... --validate-references`, with
   `--validation-cache-dir`, `--validation-email`, `--validation-full-text`,
   `--validation-max-references`, `--validation-skip-prefix`,
   `--validation-rate-limit-delay`, and `--fail-on-unresolved`.
2. **Standalone, on a saved report** — a new `validate-references` subcommand
   accepting one or more markdown files, with `--in-place`, `--output`, `--json`,
   `--check-quotes/--no-check-quotes`, and `--fail-on-unresolved`.

Both resolve every PMID/DOI the report cites (PubMed / Crossref / DataCite) and
check quotes attributed to a reference against that reference's text. Results land
in two places:

- a `## Reference Validation` markdown section appended to the report body
  (counts table + an "Unresolved references" list + unsupported quotes), and
- a `reference_validation:` block in the report's YAML frontmatter with
  `total_references`, `verified`, `not_found`, `unverifiable`, `confabulation_rate`,
  `quotes_checked`, `quotes_valid`, `unresolved_references`, `validator_version`.

Because the validator's cache dir defaults to `./references_cache`, pointing it at
dismech's own cache means the lookups a curator would have done by hand with
`just fetch-reference` are already done and cached by the time the report lands.

## Proposed work

[...]

PR #9854, opened 2026-08-28
Validate the ontology terms a deep-research report suggests (deep-research-client 0.2.11)

### LINKED ISSUE #9799: Adopt deep-research-client 0.2.11: validate the ontology terms a report suggests, not just its citations
## Problem

Every deep-research report we generate suggests ontology terms, and nothing checks them. The report template asks for them explicitly — "Suggest HPO terms for each phenotype", "Suggest GO terms for biological processes and CL terms for cell types", "suggest NCIT clinical-intervention terms", plus a closing summary of "Ontology term suggestions (HPO, GO, CL, UBERON, CHEBI, NCIT, MONDO)". Since 0.2.9 we validate a report's *citations* at generation time (#8432), but a curator reading the ontology suggestions in the same file has no signal at all about whether those CURIEs are real.

#9729 is what that looks like in practice: the OpenScientist CMTX report offered `MONDO:0010674` (Hunter syndrome) as the term for Charcot-Marie-Tooth X-linked, in a report whose citations validated 26/26 with a 0% confabulation rate. As that issue puts it, a report can be citation-clean and still carry a wrong CURIE in a summary table.

## What changed upstream

[deep-research-client v0.2.11](https://github.com/monarch-initiative/deep-research-client/releases/tag/v0.2.11) adds built-in term validation, backed by `linkml-term-validator` — the same library `just validate-terms` already runs. It arrives through a new `terms` extra, exactly parallel to how the `validation` extra brought us `linkml-reference-validator`:

```
deep-research-client[terms]  ->  linkml-term-validator>=0.4.5, prefixmaps>=0.2.6
```

Two entry points, mirroring the reference-validation pair we already wire up:

1. Inline during generation — `research ... --validate-terms`
2. Standalone on a saved report — `validate-terms report.md [--fail-on-unresolved]`

It resolves every CURIE in a report through OAK and reports three things:

| Check | Catches |
|---|---|
| existence | a CURIE no ontology contains (a confabulated ID) |
| label match | an ID named differently from its ontology label, checked against synonyms |
| obsolescence | a real term the ontology has since retired |

Label matching is deliberately positional — only labels in table cells, emphasized text, or immediately adjacent to the identifier are checked — so it undercounts rather than inventing errors. Results are available as `confabulation_rate`, `confabulated_terms`, `mislabelled_terms`, `obsolete_terms`, and a `to_markdown()` summary section.

The same release has two other changes worth taking while we are there:

- **Provider fallback, with a record of which provider actually ran.** We hit the manual version of this constantly: a brief specifies `falcon`, no `EDISON_API_KEY` is set, the curator substitutes `claude_code` and writes a paragraph about it into the history record by hand. At least six committed history records say exactly that. An opt-in fallback that records the provider it fell back to turns that into data.
- **Auth and billing failures are now non-retryable, and "available" means reachable.** This changes what `just research-providers` reports and how the scheduled workflows behave when a key is missing or a bill is unpaid — worth checking against `.github/workflows/curation-scanner.yml` and `dragon-ai.yml`.

## Proposed work

- [x] Bump to `deep-research-client[cyberian,terms,validation]>=0.2.11` and relock. Note this also forces `linkml-term-validator` from our current `>=0.4.4` to `>=0.4.5`; `scripts/run_term_validator.sh` probes that CLI for `--strict` / `--fail-on-warnings`, so check the bump against the wrapper.
- [x] Add a `dr_term_validation` justfile variable alongside the existing `dr_validation`, and wire `--validate-terms` into the six recipes that already carry `{{dr_validation}}` (`research-datasets`, `research-disorder`, `research-module`, `research-comorbidity`, `research-surrogacy`, `research-disorder-cyberian-codex`). Overridable and off-switchable the same way, for offline or quick runs. `research-hypothesis` calls the client through `scripts/hypothesis_deep_research.py` rather than the justfile, so it needs the flag added there separately.
- [x] Add a `just validate-research-terms <report>` recipe wrapping the standalone subcommand, for the ~1400 reports already in `research/` that predate this. Same caution as `validate-research-reference`: one report at a time as you come to curate it, not a tree-wide rewrite.
- [x] **Decide how it resolves terms.** We pin adapters per prefix in `conf/oak_config.yaml` (`ols:` for HP/GO/CL/UBERON/MONDO/NCIT/CHEBI, `sqlite:obo:` for hgnc/GENO/ECTO/XCO) specifically to avoid large local ontology builds. If upstream's validator does its own resolution and cannot be pointed at that config, a report validation could try to download ontologies that CI does not have. This is the main unknown and should be settled before wiring anything into a scheduled workflow.
- [x] Confirm it does not write to `cache/<prefix>/terms.csv`. Those caches are derived artifacts with a strict "never hand-write, populate only through `just validate-terms`" contract; a second writer with different provenance would break it. If it *can* populate them correctly, that is a bonus worth designing for deliberately, not by accident.
- [x] Leave `--fail-on-unresolved` off by default, like reference validation. A report is written to disk before validation runs, and a bad CURIE in a provider artifact is information, not a reason to discard the run.
- [x] Run it through `scripts/run_deep_research_client.sh` (it is the CLI entrypoint, so this is automatic) and check whether the term path needs any of the patches that wrapper applies. It probably does not — those repair the reference/cache path — but confirm rather than assume.
- [x] Evaluate the opt-in provider fallback for the research recipes, and record the provider that actually ran in the report frontmatter rather than in hand-written history prose.
- [x] Docs and skills: extend `docs/deep-research-reference-validation.md` (or add a sibling page), and update `dismech-references`, `dismech-terms`, and `initiate-new-disorder-creation` so agents know the section exists and how much weight to give it.

## What this does NOT do

- **It does not validate the KB.** `just validate-terms` on `kb/**.yaml` is unchanged and is still the gate. This checks a provider artifact.
- **It does not check dynamic-enum membership.** Our bindings must be reachable from an enum root — `regimen_term` from `NCIT:C15697`, and so on. A report suggesting `NCIT:C66930` for a regimen would pass upstream's existence check and still fail `just validate`.
- ~~**It would probably not have caught #9729.**~~ **Wrong — it does catch it.** See the comment below: the report writes `| MONDO | **MONDO:0010674** |`, the validator reads "MONDO" as the name the report gave the term, and that does not match *mucopolysaccharidosis type 2*, so it is flagged. That is a label mismatch rather than a topic judgement, so the general point stands — a real term used for the wrong disease, written with no label or with a plausible one, still needs `just preflight-dr`.

Related: #8432 (the 0.2.9 reference-validation adoption this mirrors), #9729, #9702, #4525.

caufieldjh:
Closes #9799.

Deep-research reports suggest ontology terms because our templates ask them to, and nothing checked them. Citation validation does not reach them: the OpenScientist CMTX report had 26 of 26 citations verified and still offered `MONDO:0010674` — Hunter syndrome — as the Charcot-Marie-Tooth X-linked term (#9729). `deep-research-client` 0.2.11 adds term validation backed by `linkml-term-validator`, the same library `just validate-terms` runs on `kb/`. This wires it in alongside the reference validation we adopted in #8432.

## What changed

- `deep-research-client[cyberian,terms,validation]>=0.2.11`, which also moves `linkml-term-validator` to `>=0.4.5`.
- A `dr_term_validation` variable, carried by every research recipe that already carries `dr_validation` (`research-datasets`, `research-disorder`, `research-module`, `research-comorbidity`, `research-surrogacy`, `research-disorder-cyberian-codex`), plus `scripts/hypothesis_deep_research.py`, which builds its own command outside the justfile and has a `--no-term-validation` off-switch.
- `just validate-research-terms <report>` for the reports generated before this existed.
- `docs/deep-research-term-validation.md` in the nav, with pointers from CLAUDE.md and the `dismech-terms`, `dismech-references` and `initiate-new-disorder-creation` skills.
- `tests/test_research_term_validation.py`, which fails if a research recipe validates citations while skipping terms.

Reports get a `## Term Validation` section and a `term_validation:` frontmatter summary, the same shape as reference validation. The check is not gated: a bad CURIE in a provider artifact is information, not a reason to discard a 20-minute run.

## What it found on the one report shipped here

Three real errors on a report whose 26 citations all verified:

| Finding | Class |
|---|---|
| `MONDO:0010674` offered as the CMTX term — MONDO calls it *mucopolysaccharidosis type 2* | wrong disease (#9729) |
| `HP:0106487`, "impaired distal sensation" — does not exist in HPO | confabulated identifier |
| `HP:0001265` written as "areflexia" — HPO calls it *Hyporeflexia*; *Areflexia* is `HP:0001284` | within-ontology near-miss |

The third is the one to notice. It is not about a different disease, so it reads like paraphrase, and a curator who treats it as paraphrase binds the wrong CURIE while writing the right `preferred_term`. It is the granularity failure `dismech-terms` exists to catch, arriving through a report instead of a search. The docs and the curation skill say so explicitly, because an earlier draft of that guidance told readers to look only for cross-disease errors and would have discarded it.

Two caveats the docs keep in view: the `MONDO:0010674` catch comes from a label mismatch, not from knowing what the report is about, so `just preflight-dr` is still the Named Entity Confusion check; and none of this checks dynamic-enum membership, so `just validate` remains the gate for anything bound into `kb/`.

## Two configuration choices, settled by running the validator

**Gene CURIEs cannot be checked either way, so `HGNC` is skipped.** `HGNC:4283` is GJB1. Through `sqlite:obo:hgnc` it does not resolve, because that build holds genes under the lowercase `hgnc:` this repo uses, so the uppercase form a report writes is reported as invented. Through `ols:` the same CURIE resolves to "mitochondrial chromosome". Skipping reports the prefix as unverifiable instead. Matching is case-insensitive, so it covers `hgnc:` too.

**We should not pass our own OAK config.** `--term-oak-config` works, but `conf/oak_config.yaml` maps HGNC/GENO/ECTO/XCO to `sqlite:obo:`, and passing it made a validation run download a 380 MB `hgnc.db` mid-report. The prefixes reports actually cite (MONDO, HP, GO, CL, UBERON, CHEBI, NCIT) are routed to OLS in that config anyway, which is what the upstream default `ols:` adapter does — so the default costs nothing and avoids the downloads. Curators who need ECTO or GENO checked can pass the config for that run; the docs show how.

## Deliberately not done

**Provider fallback**, also new in 0.2.11, is not wired in. It would fix a real annoyance — falcon is asked for, no `EDISON_API_KEY` is set, the curator substitutes `claude_code` and writes a paragraph about it into the history record — but each recipe names its report `<Disorder>-deep-research-<provider>.md` after the *requested* provider, and `scripts/deep_research_coverage.py` reads the provider back out of that filename. A silent fallback would write a `claude_code` report named `-falcon.md`, and `just research-status` would report falcon coverage that does not exist. Adopting it means renaming the output to the provider named in the report frontmatter after the run. The reasoning is recorded next to the variable so nobody re-derives it; happy to open a follow-up.

The other 0.2.11 change — auth and billing failures are non-retryable, and "available" now means reachable — needs no change here. Nothing in the repo reads provider availability programmatically.

**Term labels are cached in a gitignored `terms_cache/`, not the committed `cache/` CSVs.** The format is byte-identical, so this was a real choice: `cache/<prefix>/terms.csv` rows correspond to terms actually bound in `kb/`, and pointing research runs at it would add committed-cache churn for terms nobody curated, arriving through a path no reviewer is looking at. Because the format matches, a curator who wants the pre-warm can opt in for a run with one flag.

## Validation

- Full suite: 48,139 passed, 42 skipped, one failure — `test_no_new_snippet_grading_divergences`, on two `kb/` entries this branch does not touch. Those divergences arrived with #9628 and are on `main`, where the test grandfathers against `origin/main`; the local run falls back to the committed baseline, which predates that PR.
- `just validate-terms kb/disorders/Asthma.yaml` passes under the `linkml-term-validator` 0.4.5 the bump pulls in, so `scripts/run_term_validator.sh` is unaffected.
- `just validate-research-terms` run end to end on the CMTX report; recipe wiring checked with `just -n research-disorder`; the hypothesis path checked both ways with `--dry-run`.
- `uvx ruff check .` and `ruff format --check` clean on the changed files.

---

https://claude.ai/code/session_01FQmgAr2T734iSvToHFYnLV

### DIFF
--- a/.claude/skills/dismech-terms/SKILL.md
+++ b/.claude/skills/dismech-terms/SKILL.md
@@ -86,6 +86,11 @@ Two things the section does **not** settle, which is the whole of step 3 above:
 - whether it is the right term for the claim, as opposed to a real term named
   consistently.
 
+It *does* flag a near-miss when the report names one — the same CMTX report
+writes "areflexia" beside `HP:0001265`, which HPO calls *Hyporeflexia*
+(*Areflexia* is `HP:0001284`). Read those entries as granularity findings, not
+as paraphrase.
+
 Gene CURIEs are skipped by default there (`HGNC` uppercase does not resolve in
 `sqlite:obo:hgnc`, and `ols:` resolves it to an unrelated term), so verify those
 yourself. See
--- a/.claude/skills/initiate-new-disorder-creation/SKILL.md
+++ b/.claude/skills/initiate-new-disorder-creation/SKILL.md
@@ -330,10 +330,14 @@ Two places to look, as before: a `term_validation:` frontmatter block and a
 - **Read `needs_review`, not `confabulation_rate`.** The rate measures identifier
   resolution only, so a report whose every CURIE resolves but whose labels name
   different terms still shows `0.0`.
-- **`mislabelled_terms` is where the wrong-entity bindings surface.** Each entry
-  gives the report's name and the ontology's. Most are harmless paraphrase
-  ("distal weakness" for `HP:0002460`, *Distal muscle weakness*). Look for the
-  one where the ontology label is about a different disease.
+- **`mislabelled_terms` is where the wrong bindings surface.** Each entry gives
+  the report's name and the ontology's. Some are harmless paraphrase ("distal
+  weakness" for `HP:0002460`, *Distal muscle weakness*). Look for the ones where
+  the ontology label names a **different disease, or a different term in the
+  same ontology** — a sibling, a parent, a near-miss. The second kind is easy to
+  skim past: the CMTX report writes "areflexia" beside `HP:0001265`, which HPO
+  calls *Hyporeflexia* (*Areflexia* is `HP:0001284`), and those are clinically
+  distinct.
 - `unresolvable_prefixes` means nothing was checked for that prefix — not that
   anything is wrong. `HGNC` is skipped by default; verify gene CURIEs the usual
   way.
--- a/docs/deep-research-term-validation.md
+++ b/docs/deep-research-term-validation.md
@@ -44,16 +44,25 @@ The results are graded, and the grades matter when you read them:
 | Section | What it means |
 |---|---|
 | Unresolved terms | The identifier does not exist. Do not use it. |
-| Terms the report names something else | The identifier resolves to a term unrelated to what the report calls it. Usually the wrong identifier. |
+| Terms the report names something else | The identifier resolves to a term the report calls something else — a different disease, or a different term in the same ontology. Usually the wrong identifier. |
 | Terms whose name is worth a second look | The report's name is recognisably related without being the term's own name — a paraphrase, or a *related* synonym. Listed, not judged. |
 | Terms named inconsistently | The report gives one identifier more than one name of its own. |
 | Prefixes with no resolver | Not checked either way. Not evidence of anything. |
 
-Most of the middle two categories are noise on a well-behaved report — a report
-writing "distal weakness" next to `HP:0002460` ("Distal muscle weakness") is
-right, and gets flagged. Read the unresolved list first, then scan the
-"names something else" list for the one entry where the label is about a
-different disease entirely.
+Some of the middle two categories are noise on a well-behaved report — a report
+writing "distal weakness" next to `HP:0002460` (*Distal muscle weakness*) is
+right, and gets flagged. Read the unresolved list first, then work the "names
+something else" list for entries where the ontology label names **a different
+disease, or a different term in the same ontology** — a sibling, a parent, a
+near-miss.
+
+The second kind is easy to skim past and is the more common wrong-term failure.
+The CMTX report has one: it writes "areflexia" beside `HP:0001265`, which HPO
+calls *Hyporeflexia* — *Areflexia* is `HP:0001284`. Reduced reflexes and absent
+reflexes are distinct HPO terms with a real clinical distinction, so a curator
+who reads that line as paraphrase binds the wrong CURIE while writing the right
+`preferred_term`. This is the granularity problem `dismech-terms` exists to
+catch, arriving through a report rather than through a search.
 
 ## Where the results go
 
--- a/scripts/hypothesis_deep_research.py
+++ b/scripts/hypothesis_deep_research.py
@@ -67,6 +67,13 @@ DEFAULT_SUBPROCESS_TIMEOUT_SECONDS = 7800
 # Unlike reference validation, this path does not read or write
 # `references_cache/`, so it does not need the `patch_reference_validator`
 # repairs that `scripts/run_deep_research_client.sh` applies.
+#
+# Note what is still missing here, since it is the asymmetry a reader will hit:
+# this script does NOT reference-validate (no `--validate-references`), unlike
+# every justfile research recipe. That gap predates term validation. Closing it
+# means moving this script onto `scripts/run_deep_research_client.sh` first --
+# the reference path reads and writes `references_cache/` and must not run
+# unpatched.
 TERM_VALIDATION_ARGS = [
     "--validate-terms",
     "--term-cache-dir",
@@ -298,6 +305,7 @@ def build_command(
     output_root: Path,
     template: Path,
     extra_args: Sequence[str],
+    validate_terms: bool = True,
 ) -> list[str]:
     normalized = normalize_provider(provider)
     output_file = output_file_for(record, output_root, normalized)
@@ -313,7 +321,8 @@ def build_command(
         command.extend(["--var", f"{key}={value}"])
     command.extend(build_provider_args(normalized))
     command.extend(provider_default_params(normalized, extra_args))
-    command.extend(term_validation_args(extra_args))
+    if validate_terms:
+        command.extend(term_validation_args(extra_args))
     command.extend(
         [
             "--output",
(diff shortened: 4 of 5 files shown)


caufieldjh:
You're right, and it's a better catch than a wording nit — thanks. Fixed in 2c111cc, along with both suggestions.

**The 🟡.** I verified `HP:0001265` = *Hyporeflexia* and `HP:0001284` = *Areflexia* against `cache/hp/terms.csv` before changing anything, and the report does write `HP:0001265 (areflexia)` in its summary line at 816 — the table row at 654 hedges "Areflexia / hyporeflexia", but the summary is what a curator lifts from. So the wrong binding is exactly as you describe, and the old advice sent readers past it.

Broadened in three places: the grading-table row, the paragraph after it (which now works the near-miss case explicitly, with the `HP:0001265`/`HP:0001284` example), and the `initiate-new-disorder-creation` bullet. I also added it to `dismech-terms` §3b, since that skill is where the granularity rule lives and this is the same failure arriving through a report rather than a search. The PR body now leads with all three findings in a table rather than counting two.

**Suggestion 1** — recorded in the comment block, as you suggested, including that closing the reference-validation gap means moving the script onto `scripts/run_deep_research_client.sh` first, since that path reads and writes `references_cache/` and must not run unpatched. Not closing it here.

**Suggestion 2** — added `--no-term-validation` to both the `run` and `run-missing` subcommands, threaded through `run_record` and `build_command`. Verified both ways with `--dry-run`: the default renders `--validate-terms --term-cache-dir terms_cache --term-skip-prefix HGNC`, and the flag renders no term flags at all. Your `--no-term-labels` reading is right and the behaviour is deliberate, so I documented it in the docstring rather than changing the matching.

One correction to the review, since it affects what a future reader believes about the branch: the CI failure you would have seen on `0b8d9c3` was mine, not the runner — `uvx ruff check .` flagged `RUF100` on a `# noqa: E402` in the new test. Fixed in 40b0095 by loading the script with `importlib.util.spec_from_file_location`, the way `tests/test_hypothesis_deep_research.py` already loads that same module, so no import sits below top-of-file and the per-file-ignore's "no excuse in tests" rule is honoured rather than suppressed.

Tests: 7 in `test_research_term_validation.py` (two new, for the off-switch and the `--no-term-labels` case) plus the 5 existing hypothesis tests. `uvx ruff check .` reports only the pre-existing `E402` in `tests/test_ci_batch_validation.py`, which this PR does not touch.

---

PR #9871, opened 2026-08-28
Opt-in provider fallback, with the report named for the provider that ran it

caufieldjh:
Follow-up to #9854, which adopted `deep-research-client` 0.2.11 but left provider fallback unwired. This wires it in.

A brief asks for falcon, no `EDISON_API_KEY` is set, and the curator substitutes `claude_code` by hand and writes a paragraph about it into the history record. Six committed history records do exactly that, in prose, where no query can see it. 0.2.11 can do the handoff itself and record who ran.

```bash
just dr_fallback='--fallback' research-disorder falcon Marfan_Syndrome
```

## What made this unsafe, and what fixes it

The provider is in the filename, and `scripts/deep_research_coverage.py` reads it back out of there. A silent fallback would leave a `claude_code` report named `-falcon.md`, and `just research-status` would report falcon coverage that does not exist.

So every research recipe now runs an alignment step after the client returns. On a report that fell back it renames:

| | |
|---|---|
| the report | `Foo-deep-research-falcon.md` → `Foo-deep-research-claude_code.md` |
| its citations sidecar | `…falcon.md.citations.md` → `…claude_code.md.citations.md` |
| its artifacts directory | `Foo-deep-research-falcon_artifacts/` → `…claude_code_artifacts/` |

and rewrites the report's links to that directory. Reports link artifacts by directory name in both the body and the frontmatter `artifacts:` block, so renaming without rewriting would break every one of them.

Wired into all six research recipes and the hypothesis path in `scripts/hypothesis_deep_research.py`. A test keeps it that way: a recipe that can fall back must also align.

## Two decisions worth reviewing

**Fallback is off by default.** `just research-missing-provider falcon` skips a disorder only when the *falcon* report exists. With fallback always on and no key, a sweep over 50 disorders would produce 50 `claude_code` reports, and the next sweep would produce 50 more — falcon never becomes present, so nothing is ever skipped and the queue never drains. Turning it on for a single run you are watching is the case it is built for.

**Alignment triggers on `fell_back`, not on the filename disagreeing with the frontmatter.** Those disagree routinely and innocently:

- `just research-disorder edison Foo` writes `-edison.md` for a report whose provider is `falcon` — `edison` is an alias.
- the cyberian-codex recipe writes `-cyberian-codex.md` for a run whose provider is `cyberian`.

Renaming on mismatch would rewrite both. Upstream sets `fell_back` only when a provider other than the first choice produced the report, which is exactly when the filename is a lie. Both cases are pinned by tests.

## Verified against a real fallback

Not only in tests. With no `EDISON_API_KEY` in this environment, a falcon run through the actual recipe fell back to the mock provider:

```
WARNING - Provider falcon cannot do this run: no Edison Scientific API key configured. Falling back to mock
falcon could not run this job; mock produced the report instead, so it is named for mock:
  renamed …/Asthma-deep-research-falcon.md -> …/Asthma-deep-research-mock.md
  renamed …/Asthma-deep-research-falcon.md.citations.md -> …/Asthma-deep-research-mock.md.citations.md
```

and `deep_research_coverage.py` then reported `Asthma … mock`, not falcon — the coverage lie this exists to prevent.

## Details worth a look in review

- **Alignment runs on any run that left a report behind, not only a successful one.** The client writes the report *before* validating it and exits 3 when validation fails, so an `ERROR_3` run still has a real, misnamed report on disk. The recipes capture the client's status, align if the report exists, then exit with that status; the hypothesis path gates on the same condition. (It originally gated on success there — caught in review, fixed in 14678cd with a test that fails against the previous commit.)
- **Refusals are loud and move nothing** — an occupied destination (both reports stay; a curator decides), a filename that does not contain the requested slug (guessing which part is the provider is worse), a fallback that names no provider, or a provider name that is not usable in a filename.
- **The rename replaces the last occurrence of the requested slug**, which covers every layout without enumerating them: `-deep-research-<p>.md`, `-datasets-<p>.md`, the surrogacy path, and the hypothesis layout where the stem *is* the provider. A disease whose name contains a provider slug keeps its name — there is a test for that.

## Validation

- `tests/test_research_reports.py` 17 passed, `tests/test_hypothesis_deep_research.py` 7 passed, plus the recipe-wiring guards. Full suite at the first commit: 48,633 passed.
- One pre-existing failure remains, on `main` and in a file this branch does not touch: `test_no_new_snippet_grading_divergences` (the #9628 entries, grandfathered against `origin/main` in CI). The second one this PR originally flagged — `test_model_readout_targets_match_link` on `Primary_Bile_Acid_Malabsorption.yaml` — was fixed by #9875 and is green since `main` was merged in.
- `uvx ruff check .` and `ruff format --check` clean on the changed files.
- Recipe wiring checked with `just -n`; the real falcon→mock fallback re-run after the review fix.

---

https://claude.ai/code/session_01FQmgAr2T734iSvToHFYnLV


caufieldjh:
Good catch on the 🟡 — that asymmetry was mine, and your reading of why it is reachable is right. Fixed in 14678cd, along with suggestions 1, 2, 3 and 4.

**The blocking one.** I confirmed the upstream behaviour before changing anything: `cli.py:1090` says validation runs only after the report has been written, and both validation paths `raise typer.Exit(3)` with the report on disk. Since `build_command` passes `--validate-terms` by default, an ontology-service hiccup was enough to leave a misnamed report in the one tree where `existing_outputs` reads the provider straight out of the filename.

Now gated on `output_ok`, as the recipes are. I took your suggested trade on the status: an alignment refusal no longer overwrites a status that already records a real failure — `ERROR_3` stays `ERROR_3` with the alignment error appended to `detail`, and only an otherwise-successful run becomes `ERROR_ALIGN`. The validation failure is the more useful thing to report, and a run that failed both ways should say so.

The new test fails against the previous commit and passes on this one:

```
$ git stash push scripts/hypothesis_deep_research.py
$ pytest -k validation_failed   → 1 failed
$ git stash pop
$ pytest -k validation_failed   → 1 passed
```

**Suggestion 1 (`ValueError` escaping the refusal contract).** Reproduced exactly as you describe. `provider_slug()` now validates the frontmatter provider before it reaches a path — empty, `.`/`..`, or any separator raises `AlignmentError`, so every refusal stays on the one path the module documents.

**Suggestion 2 (case).** Folded into the same helper: the rename target is lowercased. Your reasoning is the deciding one — `output_file_for` lowercases when it *writes*, so a mixed-case rename in the hypothesis tree would produce a path a later `SKIPPED_EXISTS` check would not find. Defensive today, free to close.

**Suggestion 3 (`fell_back` docstring).** You are right that the behaviour is correct and the docstring overstated it. Reworded to "whether the report had to be renamed", with the retry-then-succeed case spelled out as the reason the two differ.

**Suggestion 4 (double failure).** Recorded in the justfile comment rather than restructured, as you suggested.

Two I did not act on, deliberately:

**Suggestion 5 (aligning a stale report).** I agree with your framing, and I think the current behaviour is the one I want: if a stale report carries `fell_back: true` and a wrong name, renaming it is the correct repair, not a side effect. The annoying case you name is real — a permanently-taken destination makes every failing run repeat the same refusal — but the refusal is loud and moves nothing, and the alternative (comparing mtimes across six recipes to prove the report is from this run) buys less than it costs.

**Suggestion 6 (doctests).** Agreed they are documentation, not coverage. Adding `--doctest-modules` to `pyproject.toml` would change what runs repo-wide, which is not this PR's business.

Re-verified after the change: 17 tests in `test_research_reports.py`, 7 in `test_hypothesis_deep_research.py`, `uvx ruff check .` and `ruff format --check` clean, and the real falcon→mock fallback through the recipe still renames both files.

Your CI-lane note is a real gap and matches #9473 exactly. #9875 fixes the specific entry; I'll raise the structural question separately rather than in either PR.

---

PR #9885, opened 2026-08-28
Lennox-Gastaut Syndrome: add evidence to pathophysiology and treatments

caufieldjh:
The Lennox-Gastaut syndrome entry described its disease mechanism in detail but cited almost nothing to back it up: 9 of its 10 pathophysiology nodes and 4 of its 5 treatments had no evidence at all. This PR adds cited, quote-verified evidence to every one of them, taking the entry from 22 verified snippets to 77. It also fills in prevalence, epidemiology, natural history, clinical trials, and a differential diagnosis, none of which the entry had.

Sixteen new references were fetched with `just fetch-reference`. All snippets are exact quotes checked against the local reference cache.

## Pathophysiology

Every node now carries evidence. The additions that do real work:

- **PMID:21275978** (EEG-fMRI in 11 children with LGS) supplies two things the entry had been asserting without a source: that diverse causes converge on one pathway, and that the discharge-linked network sits in the thalamus (centromedian and anterior nuclei) and brainstem. The pattern was absent in the multifocal-epilepsy control children.
- **PMID:26950270** grounds the excitation-inhibition node in molecular data. Patch-clamp work shows the de novo *GABRB3* variants found in LGS probands reduce GABA-A receptor current. Graded `IN_VITRO`, and the explanation states that this covers the monogenic GABA-A subset only — it does not establish the same lesion in structural or unknown-cause cases.
- **PMID:25400619** and **PMID:20149600** cover the slow spike-and-wave and generalized paroxysmal fast activity nodes, including the finding that sleep-recorded fast rhythms persisted in all 27 adults after waking discharges had resolved in most.
- **PMID:26945476** and **PMID:28295228** cover the epileptic-encephalopathy node. The second gives directional evidence: a child whose seizures remitted after resection showed cognitive-network reorganisation afterwards.

## Two claims I qualified rather than propped up

**The "fifth to a third" figure had no source.** The age-dependent evolution node said that "roughly a fifth to a third of patients" reach LGS from prior infantile spasms. I could not find a study measuring that fraction from the LGS side — the literature measures the reverse direction, what proportion of infantile-spasms patients go on to LGS. I changed the description to "a substantial subset" and evidenced the transition from the infantile-spasms side instead (PMID:10368073), plus prior West syndrome as an independent prognostic factor in an LGS cohort (PMID:11071139).

**Drop attacks are mostly not atonic.** PMID:40937826 reviews video-EEG-EMG studies and finds drop attacks are rarely pure atonic seizures — they are often epileptic spasms or other seizures with *increased* muscle tone. This supports the increased-tone half of the node while contradicting a purely atonic reading of its name, so it is filed as `PARTIAL` and the explanation says so. I left the node name alone; renaming it would break the entity references pointing at it, and the correction is better recorded as evidence than hidden in a rename.

Two treatment items are `PARTIAL` for the same reason. Effect sizes across the LGS-indicated adjuncts range from 68.3% drop-seizure reduction (high-dose clobazam) down to 14.8% (topiramate), so listing them together is not a claim of equivalence. And ketogenic-diet seizure freedom was largely not maintained when patients moved to a modified Atkins diet.

## Treatments

All five now have evidence, plus a `therapeutic_modality` value. The avoidance clause on broad-spectrum pharmacotherapy — that carbamazepine and vigabatrin are generally avoided — is now cited to a source showing they aggravate exactly the atypical absence and myoclonic seizure types in the LGS repertoire (PMID:9593229).

## New sections

`prevalence`, `epidemiology`, `progression`, `clinical_trials` (GWPCARE3 cannabidiol and the fenfluramine phase 3, each with both the registry record and the published result), and `differential_diagnoses` (pseudo-Lennox syndrome, distinguished by the absence of tonic seizures and fast spike series and by its remitting prognosis).

One judgment call worth a reviewer's eye: the familiar "1-2% of all epilepsy" and "3% of childhood epilepsy" figures went into `epidemiology`, **not** `prevalence`. Their denominator is the epilepsy population, not the general population, so a `rate_per_100000` in a prevalence record would be read by any downstream aggregation as a general-population rate and be wrong by two to three orders of magnitude. `prevalence` holds one population-based record: a cumulative incidence of 13.2 per 100,000 children through age 16.

## Ontology terms

No bindings changed. I searched NCIT for implanted vagus nerve stimulation and for corpus callosotomy; neither has a clinical-action term, so both treatments keep `NCIT:C15329` (Surgical Procedure) and the platform distinction is carried by `therapeutic_modality` (`DEVICE` and `SURGERY` respectively).

## Validation

`just validate-disorders` passes schema, term, and reference validation with 77/77 snippets verified. The folded-hyphen, snippet-length, title-snippet, snippet-grading, environmental-evidence, reference-title, duplicate-key, and entity-ref gates all pass. Full `tests/test_data.py`: 44,977 passed, 1 failed.

That one failure was `test_model_readout_targets_match_link` on `Primary_Bile_Acid_Malabsorption.yaml`, a file this PR does not touch. It was a known failure on main, and it has since been fixed there by #9875, which merged shortly before this PR opened. My local run predates that fix, so nothing here needs to change on account of it.

A history record is included at `history/disorders/Lennox-Gastaut_Syndrome/`.

https://claude.ai/code/session_018obeA3gbzHC1jDfdMQ7ekS

### DIFF
--- a/history/disorders/Lennox-Gastaut_Syndrome/2026-08-28T160804Z-claude-code-0e5bb4.yaml
+++ b/history/disorders/Lennox-Gastaut_Syndrome/2026-08-28T160804Z-claude-code-0e5bb4.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9885
   urls: []
 events:
 - type: EDIT
--- a/kb/disorders/Lennox-Gastaut_Syndrome.yaml
+++ b/kb/disorders/Lennox-Gastaut_Syndrome.yaml
@@ -104,13 +104,16 @@ prevalence:
 epidemiology:
 - name: Share of childhood epilepsy
   description: >-
-    LGS accounts for a small but consistently reported share of childhood
-    epilepsy. Published estimates range from about 3% to as much as 10% of all
-    childhood epilepsies; the spread reflects differing case definitions and
-    ascertainment rather than genuinely different populations. This is a
-    proportion of children who already have epilepsy, not a population
-    prevalence, and must not be compared with the rate under `prevalence`.
-  minimum_value: 3.0
+    LGS accounts for a small share of childhood epilepsy, but the published
+    figures span an order of magnitude: 1% to 10% of all childhood epilepsies.
+    The bounds here are that full reported range rather than any single study's
+    estimate; 3% is a commonly cited point value inside it. Camfield attributes
+    the spread to how widely prevalence rates vary between reports, and
+    differing case definitions and ascertainment are the usual explanation
+    rather than genuinely different populations. This is a proportion of
+    children who already have epilepsy, not a population prevalence, and must
+    not be compared with the rate under `prevalence`.
+  minimum_value: 1.0
   maximum_value: 10.0
   unit: percent of childhood epilepsies
   evidence:
@@ -120,16 +123,16 @@ epidemiology:
     evidence_source: HUMAN_CLINICAL
     snippet: "Lennox-Gastaut syndrome occurs in 3% of children with epilepsy"
     explanation: >-
-      Gives the lower bound of the reported share of childhood epilepsy
-      accounted for by LGS.
+      Gives a commonly cited point estimate for the share of childhood epilepsy
+      accounted for by LGS, sitting inside the wider reported range.
   - reference: PMID:21790560
     reference_title: Definition and natural history of Lennox-Gastaut syndrome.
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: "Reported prevalence rates for LGS vary widely from 1-10% of all childhood epilepsies. Incidence rates are much lower."
     explanation: >-
-      Gives the upper bound and records both the width of the published range
-      and the separation between prevalence and the much lower incidence.
+      Supplies both bounds recorded here, and separates this prevalence-style
+      share from the much lower incidence.
 - name: Share of all epilepsy
   description: >-
     Among patients with epilepsy of any age, LGS accounts for an estimated 1-2%.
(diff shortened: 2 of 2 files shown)


caufieldjh:
Pushed 439e7c4 with all three items fixed.

**The blocking one.** Ran `just backfill-reference-titles kb/disorders/Lennox-Gastaut_Syndrome.yaml`; both `clinicaltrials:` titles are now filled in from the committed cache files, so they match their `references_cache/` frontmatter exactly.

**The epidemiology floor was a real inconsistency, not just a minor.** Good catch — `minimum_value: 3.0` excluded the 1% that the entry's own second source (PMID:21790560) reports for the same denominator. I widened the bounds to 1.0-10.0, the full published range, and rewrote the description to say that 3% is a commonly cited point value inside that range rather than the floor. Both evidence `explanation` fields are reworded to match, since PMID:20158289 now supplies a point estimate and PMID:21790560 supplies both bounds — the old text called one of them "the lower bound" and the other "the upper bound", which stops being true once the range widens.

**History record** now links this PR.

Validation after the change: `just validate-disorders` passes with 77/77 snippets still verified, and the reference-title, snippet-length, title-snippet, snippet-grading, folded-hyphen, duplicate-key and entity-ref gates all pass. History tests: 29 passed.

## What I deliberately left alone

The four 🔵 items — no HGNC `gene_term` bindings in `genetic:`, no `inheritance:` block, no `diagnosis:` block, no UBERON `locations:` on the thalamus/brainstem node, and missing `therapeutic_agent` bindings for lamotrigine, topiramate, felbamate and valproate — are all pre-existing structural absences that predate this PR. You marked them non-blocking and worth their own issues, and I agree; pulling them in would widen an evidence-augmentation PR into a re-modelling one. Happy to open issues for them, or take them in a follow-up PR if that's preferred.

Two of them are genuinely close to free and would be my first follow-up: the GABRB3 HGNC binding (already snippet-evidenced here via PMID:23934111 and PMID:26950270) and the UBERON `locations:` on the thalamus/brainstem node, since PMID:21275978's snippet already names the structures.

Re-requesting review.

---

PR #9886, opened 2026-08-28
Curate autosomal dominant hypocalcemia 1 (MONDO:0011013)

### LINKED ISSUE #9873: Curate autosomal dominant hypocalcemia 1 (MONDO:0011013)
Curate a dismech entry for **autosomal dominant hypocalcemia 1** ([MONDO:0011013](https://monarchinitiative.org/MONDO:0011013)) — heterozygous activating (gain-of-function) variants in the calcium-sensing receptor CASR reset the parathyroid and renal calcium set-point downward, giving hypocalcaemia with inappropriately low-normal PTH and relative hypercalciuria.

Preliminary read: this looks like a `DISEASE`. One causal gene (`hgnc:1514` CASR), no MONDO descendants, and a single conserved mechanism. ADH2 (GNA11) is a separate MONDO term with a different gene and belongs in its own entry, not as a subtype here.

The Bartter-syndrome-like variant (severe activating CASR alleles causing a salt-wasting phenotype alongside hypocalcaemia, sometimes labelled "Bartter syndrome type V") appears among this term's own synonyms, so it should be modelled as a `has_subtypes` entry here rather than folded into the existing `Bartter_Syndrome` entry.

This is a good worked case for the `functional_impact_category: GAIN_OF_FUNCTION` vs `modifier` distinction in `CLAUDE.md` — the variant consequence and the receptor-signalling state are two different claims.

Stub: `stubs/Autosomal_Dominant_Hypocalcemia_1.yaml`

aleixpuigb:
New disease entry for ADH1, the calcium-sensing receptor gain-of-function disorder.

## What this entry says

ADH1 is a moved set point rather than a missing protein. A heterozygous activating CASR variant leaves the receptor expressed and correctly located but shifts its calcium dose-response curve left, so concentrations a normal receptor would read as low are read as sufficient.

The one modelling decision worth a reviewer's attention is that the parathyroid and renal limbs are curated as **two parallel arms of that single shift**, converging on `Renal Calcium Wasting`. That is not decoration. It is why raising serum calcium with calcitriol makes the kidney worse rather than better, and it is why the accepted treatment target is a deliberately sub-normal calcium. An entry that recorded only "hypoparathyroidism causes hypercalciuria" would make the conventional treatment look adequate.

Also curated:

- `Bartter type V` as a **subtype**, not a separate entity — the primary lesion is the same receptor, and the label appears among this MONDO term's own synonyms. It is not folded into the existing `Bartter_Syndrome` entry for that reason.
- ADH2 (GNA11) as a **differential**, not a subtype — same biochemistry, different lesion, and the distinction matters for whether a calcilytic has a target at all.
- A `KNOWLEDGE_GAP` discussion on the finding that in vitro receptor activation did not predict clinical calcilytic response, with biased signalling as an evidenced candidate explanation.

## Gain of function, in two senses

Both schema slots are used, making different claims, per the decision table in `CLAUDE.md`:

- the variant node's `genetic_context.functional_impact_category: GAIN_OF_FUNCTION` — the consequence of the variant;
- the receptor node's `modifier: GAIN_OF_FUNCTION` — signalling outside normal regulatory constraint;
- the downstream signalling node's `modifier: INCREASED` — the quantitative claim, an amount of signalling per unit calcium.

## Validation

- `just validate` — schema, terms, **44/44 reference snippets**
- `just check-folded-hyphens`, `just check-snippet-length`, `just check-title-snippets`, `just check-snippet-grading`, `just check-environmental-evidence` — pass, no baseline changes
- `just validate-history` — pass

## How deep research was used

An `openscientist` run was commissioned for this entry. The pathograph and the first eight references were built independently from PubMed before it returned; the report was then read as a source of leads and contributed four more. Its independently-derived causal model agreed with the curated one, including the parallel-arms point.

Two things it got wrong are recorded in the entry's `notes`: it proposed a confabulated CL identifier (`CL:1000456`, named as a collecting-duct cell but actually a peritoneal mesothelial cell), which is not used here, and it gave `MONDO:0008833` as this disease's identifier where the stub and MONDO give `MONDO:0011013`.

## One thing reviewers may want to check

Three snippets I first drafted from a PubMed abstract-metadata service did not match the `just fetch-reference` cache word for word — "clinical parameter" vs "clinical parameters", a mangled `Gα11`, and a sentence the metadata abstract phrased differently from the cached full text. All are corrected against the cache, but it is a reminder that only the fetched cache counts as the source of a snippet.

Closes #9873

---
_Generated by [Claude Code](https://claude.ai/code/session_01H9GzuQh19WtiMWLsHE9ipR)_


aleixpuigb:
Pushed `79147f0d0`. The central finding was correct and I've taken all six IMPORTANT items and all seven suggestions.

**The criticism I want to acknowledge properly**, because it generalises beyond this PR: I treated the deep-research run's reference caches as a by-product of validation rather than as a reading list. Leaving 22 of 33 uncited was, in effect, a claim that they held nothing — and that claim was wrong in at least six places. Twelve of the additions below came from files already committed here; only encaleret needed new fetching.

| Finding | What changed |
|---|---|
| 1. Thiazides | Added, `MODULATES` → `Renal Calcium Wasting`, cited to the sentence that singles them out for activating CaSR mutations specifically. The rhPTH "tubular limb untouched" claim is corrected. |
| 2. Encaleret | Added with the phase 2b result, plus `clinical_trials` for NCT04581629 and palopegteriparatide as a separate treatment. |
| 3. Frequencies | PMID:16128246 kindred denominators now carry them; intracranial calcification `OCCASIONAL` → `FREQUENT`. |
| 4. Hypomagnesemia | Added as a phenotype, plus paraesthesia. |
| 5. CKD | Added as a phenotype with `PROGRESSIVE` course. |
| 6. Renal mechanism | `CLDN14` on the renal node; new `PROVISIONAL` node for the vasopressin–AQP2 arm from PMID:38367250. |

Snippets 44 → **67/67 verified**. All seven offline gates pass, no baseline changes.

**Two places I went further than asked, and one where I went differently.**

Further: encaleret turns out to matter more than "keeping the calcilytic story current". It corrected hypocalcaemia *and* reduced hypercalciuria together over 24 weeks — the only intervention in this entry that moves both sides of the trade-off the whole entry is built around. I rewrote the calcilytic description around that rather than appending to it. And PMID:37754292 states hypomagnesaemia among the cardinal consequences, which is independent support for your finding 4 that it is not subtype-specific.

Differently: on **finding 12** (`evidence_source` applied two ways) I took your reading rather than defending mine, because `CLAUDE.md` is explicit that the slot classifies the publication. Background sentences from the case series are now `HUMAN_CLINICAL`, and `OTHER` is reserved for sources that are not primary studies — a review, a systematic review, a registration record. The rule is written into `notes` so the next editor does not have to re-derive it.

**One I did not do:** I have not asserted that rhPTH(1-84)/Natpara is withdrawn. It may well be, but I could not verify it from a cached source, and an entry that states a regulatory fact should cite one. Palopegteriparatide is added alongside rather than as a replacement, and both are graded to the strength of their evidence — a three-patient case series and a single case report respectively. If you have a citable source for the withdrawal I will add it and reframe the treatment.

`mappings` (suggestion 10) is also not added: `ORPHA:428` needs an `ORPHA_428.md` cache that does not exist in this branch, and generating one means running `just structured-rebuild-orphanet`, which touches the shared Orphanet cache well outside this PR's scope. Better as its own change.

Re-requesting review.

---

aleixpuigb:
Thanks for the approval — and I've spent it, deliberately, on your first 🔵.

**The CLDN14 finding was the most important thing in the review**, even though you marked it non-blocking. I added a gene descriptor during the review response whose only source was the deep-research report, in an entry whose own `notes` say a report is a lead. That is precisely the failure this PR spent two rounds arguing against, committed by me while arguing it. An uncited gene assertion sitting in a knowledge base is a worse outcome than an extra review cycle, so `cac21fa7f` fixes it and the approval resets.

It cited better than I expected. PMID:25071082 is ADH1-specific rather than a generic claudin-14 reference: ADH receptor variants repress miR-9 and miR-374, which lifts the brake on claudin-14 and closes the paracellular calcium route in the thick ascending limb — with an HDAC-inhibitor rescue in cell and animal models. The node now states the route and that it is evidenced in models, not patients. Snippets 67 → **69/69**.

On the remaining 🔵, and why I'm leaving them:

- **`mappings`** — `ORPHA:428` needs an `ORPHA_428.md` cache that isn't in this branch, and generating one means `just structured-rebuild-orphanet`, which touches the shared Orphanet cache well outside this PR. Its own change. (Worth noting: on the harderoporphyria PR you suggested `ORPHA:100924` for that disease; I fetched it and it resolves to *Porphyria due to ALA dehydratase deficiency*. Your "verify the code" caveat was well placed.)
- **`Hyperphosphatemia` / `Neuromuscular Irritability` without ontology terms** — both are clinical states, not processes; the honest bindings would be the HP terms already carried by the corresponding `phenotypes` entries, which would duplicate rather than add. Happy to be told otherwise.
- **The one-sided Bartter-type-5 warning** — agreed it should be reciprocal, but editing `Bartter_Syndrome.yaml` from a PR that doesn't otherwise touch it is scope creep. Better as a small follow-up.
- **`PMID:38572533` fetched and unused** — agreed, and left that way.

---

PR #9887, opened 2026-08-28
curate(CD16 Deficiency): add MONDO:0014313 with spontaneous cytotoxicity and ADCC as separate nodes

### LINKED ISSUE #9879: Curate autosomal recessive primary immunodeficiency with defective spontaneous natural killer cell cytotoxicity (MONDO:0014313)
Curate a dismech entry for **autosomal recessive primary immunodeficiency with defective spontaneous natural killer cell cytotoxicity** ([MONDO:0014313](https://monarchinitiative.org/MONDO:0014313)) — immunodeficiency 20 / CD16 deficiency, caused by a homozygous *FCGR3A* variant that leaves NK cells able to kill via antibody-dependent cytotoxicity but unable to kill spontaneously, presenting with severe herpesvirus and HPV disease.

Leaning `DISEASE`. MONDO's parents are only the broad `inborn error of immunity` and `hereditary disease`, and nothing in `kb/` currently covers the concept — *FCGR3A* appears in the KB only as a polymorphic Fc receptor in Kawasaki disease and neuropsychiatric SLE, which is a different claim entirely.

The interesting feature for the pathograph is that this is a *dissociated* functional defect rather than an absence of cells: CD16 is expressed but a single substitution disrupts an epitope needed for the spontaneous-cytotoxicity synapse, so ADCC is preserved. The entry should keep those two NK functions as separate nodes rather than collapsing them into "NK dysfunction". The literature is small, so evidence discipline matters more than usual here.

Stub: `stubs/Autosomal_Recessive_Primary_Immunodeficiency_With_Defective_Spontaneous_Natural_Killer_Cell_Cytotoxicity.yaml`

aleixpuigb:
New entry for **CD16 deficiency** (immunodeficiency 20, MONDO:0014313) — a homozygous *FCGR3A* p.Leu66His substitution, associated with severe herpesvirus and papillomavirus disease.

> **Updated after two review rounds.** The original body described a two-patient, causal-relationship entry and is superseded by this. Four homozygotes are now curated and the gene-disease relationship is `DISPUTED`.

## The headline: causality is unsettled, and the entry says so

Four homozygotes are published, and they do not agree:

| Patient | Spontaneous cytotoxicity | ADCC | Clinical |
|---|---|---|---|
| Jawahar 1996 (index) | reduced | intact | recurrent HSV, low NK count |
| de Vries 1996 | **normal** | normal | severe course after BCG, EBV, VZV |
| Grier 2012 | reduced | intact | EBV-associated Castleman, HPV |
| Izadi 2021 | **normal** | — | **asymptomatic**, found on newborn screening |

L66H sits at roughly **5% allele frequency with ~100 homozygotes in gnomAD** — far too common for a fully penetrant recessive immunodeficiency — and the most recent report concludes it is "unlikely a direct genetic cause".

So `genetic.relationship_type` is `DISPUTED` and the allele is `UNCERTAIN_SIGNIFICANCE`, not causative and pathogenic. The mechanism nodes describe what was measured in the patients who *were* affected, and do not claim every homozygote has them.

## The modelling call

This is a *dissociated* functional defect, not a missing receptor. CD16 is expressed and still binds IgG Fc. What L66H destroys is the B73.1-recognised conformational loop, and with it CD16's association with CD2 — which antibody-independent killing depends on.

Spontaneous cytotoxicity and ADCC are therefore **separate nodes**, including an explicit *preserved-function* node for ADCC. That node stands **unrooted**: nothing points `downstream` into it, because preservation is the absence of an effect rather than something the epitope loss causes. (Precedent: `Calvarial_Doughnut_Lesions-bone_Fragility_Syndrome`.)

## Three discussions, not one

- `l66h_pathogenicity_disputed` (CONTROVERSY) — the causality argument above, including the ascertainment confounder in both directions and what would settle it.
- `nk_cytopenia_inconsistent` (CONTROVERSY) — NK counts split three ways: a genuine gating artifact (CD16-inclusive gates read an L66H homozygote as NK-absent), a genuinely low count in the index patient **whose gate did not include CD16, so the artifact does not explain it**, and normal counts in two others.
- `cytotoxicity_to_viral_susceptibility_gap` (KNOWLEDGE_GAP) — the in vitro → in vivo step is asserted on plausibility. The new papers *widen* this: de Vries had a severe viral course with normal cytotoxicity, and Izadi has the genotype with no disease.

## Diagnosis: necessary, not sufficient

Both diagnostic entries were rewritten so they no longer contradict the `DISPUTED` grading. Sequencing finds a common allele, and a filtered rare-variant pipeline may discard it entirely. B73.1 epitope loss is a **screening** signal — it can occur independently of L66H and must be confirmed by Sanger sequencing.

## No treatments section

No disease-directed therapy has been reported; management in the published cases targeted the individual infections.

## Literature sweep

`research/CD16_Deficiency-deep-research-fallback.md` records the sweep, the searches run, and what they changed. It is honest that no provider run happened — the sweep was done directly through PubMed and reference-list traversal.

Worth recording why it exists: the initial entry was built on two of the four homozygotes, and **de Vries was reachable as reference 17 of a paper already cached in this PR**. The contradicting evidence was in material already in hand.

## Evidence base

PMID:8608639 (Jawahar 1996), PMID:8874200 (de Vries 1996), PMID:23006327 (Grier 2012, incl. NK-92 reconstitution), PMID:34448085 (Izadi 2021).

## Validation

`just validate-disorders` passes. 35/35 snippets verified. Terms validate. Cache diff is four rows. All ungated gates pass.

Closes #9879

### DIFF
--- a/kb/disorders/CD16_Deficiency.yaml
+++ b/kb/disorders/CD16_Deficiency.yaml
@@ -346,11 +346,21 @@ pathophysiology:
       NK cell cytotoxicity."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
+    snippet: "The decrease in CD2 surface expression on cells of patient 1 was limited
+      to CD56 dim NK cells, which were primarily CD16"
+    explanation: The positive half of the subset restriction, which the abstract sentence
+      does not carry. The CD2 decrease is confined to the subset that expresses the
+      mutant receptor.
+  - reference: PMID:23006327
+    reference_title: "Human immunodeficiency-causing mutation defines CD16 in spontaneous
+      NK cell cytotoxicity."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
     snippet: "as CD2 expression on CD16–CD56bright NK cells was equivalent in patient
       1 and the control"
-    explanation: Evidences the subset restriction itself, which the abstract sentence
-      does not carry. CD2 is unaffected on the CD16-negative CD56-bright subset, and
-      that contrast is the argument that the CD2 loss follows the mutant receptor.
+    explanation: The negative half of the same contrast. CD2 is unaffected on the
+      CD16-negative CD56-bright subset, and it is the pairing of the two halves that
+      makes this the argument that the CD2 loss follows the mutant receptor.
 
 - name: Preserved IgG Fc Binding and ADCC
   biological_scale: CELLULAR


aleixpuigb:
Both blocking findings were right, and I verified each independently before acting. Thank you — this changed the entry substantively rather than cosmetically.

## The part I got wrong

de Vries 1996 (PMID:8874200) was reachable as **reference 17 of a paper I had already cached in this PR**. I read `PMID_23006327.md` for the mechanism and the patient description and did not follow its reference list. That is the avoidable failure here: the contradicting evidence was in material already in hand.

Izadi 2021 was a plain literature-sweep miss, and it turned out to carry the single most consequential fact about this allele — the gnomAD frequency.

## What changed

The entry no longer claims causation:

| | before | after |
|---|---|---|
| `genetic.relationship_type` | `CAUSATIVE` | `DISPUTED` |
| variant `clinical_significance` | `PATHOGENIC` | `UNCERTAIN_SIGNIFICANCE` |
| cytotoxicity node | asserted | `mechanism_confidence: PROVISIONAL` + two `REFUTE` evidence items |

Roughly 5% allele frequency with ~100 gnomAD homozygotes is not compatible with a fully penetrant recessive immunodeficiency, and I should have gone looking for that number before asserting causation on four patients.

A new `l66h_pathogenicity_disputed` CONTROVERSY carries the argument, including the ascertainment confounder running in both directions and what would actually settle it (a systematic NK-function study of unselected homozygotes, which nobody has published).

## The NK-count controversy, rewritten around three mechanisms

I kept your precision point and made it explicit, because it is the thing most likely to be lost later:

1. The gating artifact is real — CD3⁻CD56⁺**CD16⁺** scores an L66H homozygote as NK-absent, CD3⁻CD56⁺ gives 26%. de Vries predicted this in 1996.
2. **It does not explain Jawahar.** That patient was gated CD56(+)CD3(-) *without* CD16, so the low count is not the artifact and remains unexplained.
3. Two patients had normal counts outright.

The practical consequence — NK lymphopenia is not usable as a screening criterion, and any reported count has to be read with its gating strategy — is now stated in the discussion.

## Other items

- **`HP:0031410` dropped.** You are right that it means an altered *proportion* between subsets, and my claim and snippet were about subset-restricted CD2 expression. The biology is carried by the pathophysiology node, so I removed the phenotype rather than keeping a free-text stub.
- **Edge into the preserved-function node removed.** Kept the node, unrooted, per the `Calvarial_Doughnut_Lesions` precedent you pointed to; folded the "Fc-binding surface is not part of the altered loop" reasoning into its `description` and set `mechanism_confidence: ESTABLISHED`.
- **`supports: PARTIAL`** on the wild-type-association evidence, as suggested.
- **VZV and BCG** added to the infection spectrum.
- **Numbering conventions** recorded in `notes`: `c.230T>A` / `c.197T>A` / `p.L48H` are one substitution, and the `c.` positions are the sources' own phrasing rather than transcript-validated HGVS.
- **`pathophysiology` → `phenotypes` edges** added.
- **`research/CD16_Deficiency-deep-research-fallback.md`** added, recording the sweep, the searches run, and exactly what they changed. I did not run a provider — I did the sweep directly through PubMed and reference traversal, and the record says so rather than implying a provider run happened.

## One suggestion I did not take

Suggestion 6, the better full-text snippet for the subset-restricted CD2 claim. The sentence line-wraps mid-word across `CD56` / `dim` in the cached PDF and did not verify as an exact substring, so quoting it would have meant either a mid-sentence fragment or a snippet that fails the gate. I left the abstract snippet, which the node's claim is now narrower than. Happy to revisit if there is a normalization the snippet checker applies that I could rely on.

## Validation

`just validate-disorders` passes. 31/31 snippets verified. Terms validate. Folded-hyphen, snippet-length, title-snippet, snippet-grading, duplicate-key and entity-ref checks all pass.

---

aleixpuigb:
All four fixed. Three of these were caused by my round-1 revision, and one was a straight regression — worth naming plainly.

## Item 4 first, because it was my error

I deleted the `cytotoxicity_to_viral_susceptibility_gap` KNOWLEDGE_GAP. Nothing in your review asked for that. It happened because I rewrote the whole `discussions:` block as one unit instead of editing the entry that needed changing — so the discussion I was *adding* silently displaced one I meant to keep. That is a bad editing habit on a section with independent items, and the PR body still advertising the deleted discussion is exactly the kind of drift it causes.

Restored, and you are right that the gap got **wider**: de Vries had a severe viral course after BCG, EBV and VZV *with normal cytotoxicity*, and the Izadi homozygote has normal cytotoxicity and no disease at all. So the rationale now has to explain a comparable clinical picture arising without the proposed mechanism, and the mechanism arising without the picture.

## Item 1 — and a correction to something I told you

The subset claim did become unevidenced when `HP:0031410` came out. Dropping the term was right, but it moved the load onto the pathophysiology node, whose only evidence was an abstract sentence that says nothing about subsets. Now cited from the full text.

**A correction:** in my round-1 reply I declined suggestion 6 on the grounds that the sentence "line-wraps mid-word … and did not verify as an exact substring". The snippet you pointed to here is also line-wrapped — between "NK cells" and "was" — and it verifies fine. So the reference validator normalises whitespace, and my reason for declining was wrong. What actually blocked the earlier one was my raw `grep -F` test, not the validator. I have stopped using grep as a proxy for it.

(The same mistake is in my reply on #9891, where I gave line-wrapping as part of the reason for not snippeting a table row. The real and still-valid reason there is snippet *length* — the row is four words — but I overstated the wrap. Noting it here rather than silently leaving it.)

## Item 2 — the contradiction

You are right that `diagnosis` had not moved with `genetic`. Sequencing is now framed as necessary but not sufficient, and the Izadi NGS pitfall is cited: a standard rare-variant pipeline may discard the allele as too common to report. That is a genuinely useful, previously unused fact — a lab could fail to find this variant *because* it is common, which is the same property that makes the causal claim doubtful.

## Item 3 — the flow signature

Reframed as a screening signal rather than a diagnostic one, citing the qualification as `PARTIAL`: B73.1 epitope loss "can occur independently of the L66H transversion", so the discordance must be confirmed by Sanger sequencing.

## Non-blocking items

All four done:
- NK-count phenotype description refreshed to cover all four homozygotes and point at the three-mechanism discussion.
- Second `pathophysiology` → `phenotypes` edge added.
- Izadi's 346-child cohort added alongside gnomAD, graded `PARTIAL` because it *qualifies* the population argument — gnomAD may overstate the frequency, though the allele is still far too common for full penetrance.
- `mechanism_confidence: PROVISIONAL` on the infection node.

## One process note

Merging the two `evidence` blocks in the sequencing entry produced a duplicate `evidence` key. `check-duplicate-keys` caught it locally, and I fixed it by merging the lists rather than deleting a block, per the CLAUDE.md guidance.

## Validation

`just validate-disorders` passes. 35/35 snippets verified. Terms validate. Duplicate-key, entity-ref and all baseline gates pass.

---

aleixpuigb:
Heads-up on a process mistake of mine, plus one small improvement.

**I cancelled your round-3 review.** I edited the PR body at 18:23 to bring it in line with the current entry — it still described the two-patient causal version and still advertised the discussion I had deleted, which you flagged. That edit superseded the in-flight `claude-code-review` run, which was cancelled at 18:26:57 with nothing queued to replace it. I could not re-trigger it: `run_workflow` and `rerun_workflow_run` both return 403 for me.

So I have pushed a real change rather than leaving the PR sitting on a stale `CHANGES_REQUESTED` with no review in flight. Not an empty commit — it addresses the remaining half of your round-2 finding 1.

**What the push contains.** You noted the positive half of the subset claim was "the harder quote" because the cached PDF splits `CD56`/`dim` mid-word. It verifies once quoted against whitespace-normalised text:

> "The decrease in CD2 surface expression on cells of patient 1 was limited to CD56 dim NK cells, which were primarily CD16"

Both halves are now cited on the node, because the argument is the pairing: the CD2 decrease is confined to the subset expressing mutant CD16, *and* absent from the subset that does not. Either alone is weaker than the two together.

That is the same lesson as the previous commit — the validator normalises whitespace, and my round-1 reasoning about line wraps was based on a raw `grep -F` rather than the validator itself.

**Validation:** `just validate-disorders` passes, 36/36 snippets verified, terms validate, all ungated gates pass.

Sorry for the churn on your side. If editing a PR body reliably cancels an in-flight review on this repo, that is worth knowing generally — I will hold body edits until a review has posted.

---

PR #9890, opened 2026-08-28
Curate Pendred syndrome (MONDO:0010134)

### LINKED ISSUE #9872: Curate Pendred syndrome (MONDO:0010134)
Curate a dismech entry for **Pendred syndrome** ([MONDO:0010134](https://monarchinitiative.org/MONDO:0010134)) — autosomal recessive loss of the SLC26A4 anion exchanger pendrin, causing congenital sensorineural hearing loss with enlarged vestibular aqueduct plus a partial thyroid iodide organification defect and euthyroid goitre.

Preliminary read: this looks like a `DISEASE`. MONDO records a single causal gene (`hgnc:8818` SLC26A4) and no descendants, and the mechanism is one conserved pathograph (pendrin loss → failure of endolymphatic Cl-/HCO3- exchange → endolymph acidification and enlarged vestibular aqueduct → hair cell loss; the same transporter loss impairs apical iodide efflux at the thyrocyte).

Adjacent entries to keep distinct rather than merge into: `Congenital_Hypothyroidism` and `Thyroid_Dyshormonogenesis_5` both mention SLC26A4/pendrin in passing but neither carries MONDO:0010134. DFNB4 (nonsyndromic EVA) is the allelic hearing-only end of the same SLC26A4 spectrum and is a candidate `has_subtypes` entry here rather than a separate file.

Stub: `stubs/Pendred_Syndrome.yaml`

aleixpuigb:
New disease entry for Pendred syndrome — biallelic SLC26A4 loss of pendrin, an apical anion exchanger, acting in two epithelia.

## Two decisions worth checking

**DFNB4 is a subtype here, not a separate entry.** Same gene, same inner-ear mechanism, same allelic spectrum. What separates the labels is whether a thyroid iodide organification defect is demonstrable, and dietary iodine modifies that. Splitting them would duplicate the entire pathograph to record one branch. If the project prefers a separate DFNB4 entry, say so and I will split it.

**The inner-ear arm is modelled as developmental and window-bounded**, not as a degeneration that could be arrested at any point. That is what the conditional Slc26a4 mouse showed — pendrin is required across E16.5–P2, and its absence there acidifies endolymph, collapses the endocochlear potential and prevents hearing acquisition. It is also why "when does the human window close" is the central open question for every disease-modifying therapy in the entry, recorded as a `HUMAN_MODEL_MISMATCH` discussion rather than a generic knowledge gap: the evidence is not missing, it is positive and in mice, and its translation is what is unresolved.

A second, parallel inner-ear route through local thyroid-hormone deficiency ("cochlear hypothyroidism") is curated as `PROVISIONAL` — it rests on one mouse allele and is stated by its authors as a proposal.

## Validation

- `just validate` — schema, terms, **39/39 reference snippets**
- all seven ungated offline checks pass, no baseline changes

## What the deep-research run got right and wrong

An `openscientist` run was commissioned. On citations it was accurate: all thirteen PMIDs I took from it resolved to the titles it gave them when fetched. Its independently derived causal model matched the curated one, including the dual inner-ear route.

Its ontology suggestions were a different matter, and the two failures are usefully different:

- `CL:1000722`, proposed for a β-intercalated cell, **does not exist**. The report's own term-validation section caught this.
- `UBERON:0002518`, proposed for the endolymphatic sac, **exists — as "otolith organ"**. Nothing in the validation output flagged it, because the identifier resolves. The correct term (`UBERON:0002223`) came from searching the ontology.

The second is the more dangerous failure mode: a fabricated identifier fails validation, whereas a real identifier for the wrong concept passes every automated existence check there is.

## Three things the repo's own gates caught in my draft

Worth recording because they are the checks earning their keep:

1. `GO:0015701` (bicarbonate transport) used as a **molecular function** — real term, wrong slot, rejected by the `MolecularFunctionTerm` dynamic enum. Moved to `biological_processes`; molecular function rebound to `GO:0140829`.
2. `NCIT:C157820` (Cochlear Implant) as a `treatment_term` — real term for the *device*, not reachable from `NCIT:C25218`, so not a clinical action. Rebound to `NCIT:C15329` with `preferred_term: cochlear implantation`.
3. A snippet quoting only a paper's title phrase, rejected by both `check-snippet-length` and `check-title-snippets` with no grandfathering escape. Re-reading the cache showed the abstract carried a proper sentence — the defect was my reading, not the source.

Closes #9872

---
_Generated by [Claude Code](https://claude.ai/code/session_01H9GzuQh19WtiMWLsHE9ipR)_

### DIFF
--- a/history/disorders/Pendred_Syndrome/2026-08-28T180647Z-claude-code-ad2dff.yaml
+++ b/history/disorders/Pendred_Syndrome/2026-08-28T180647Z-claude-code-ad2dff.yaml
@@ -1,0 +1,29 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Pendred_Syndrome
+  path: kb/disorders/Pendred_Syndrome.yaml
+session:
+  id: 2026-08-28T180647Z-claude-code-ad2dff
+  timestamp: '2026-08-28T18:06:47Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9890
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - genetic
+  - phenotypes
+  - treatments
+  summary: Correct a false schema claim and curate the founder alleles as structured
+    variants
+  details: |
+    Round-three review response. The single blocker was that the genetic notes claimed the schema's variant machinery hangs off has_subtypes, which is false - Genetic has a variants slot, Subtype does not, and 251 disorder entries already use Genetic.variants. Verified against the schema before acting. Content had been omitted on the strength of that false claim, which is the worse half of the error. The two Okinawan founder alleles are now structured Variant records: p.His723Arg with its trafficking mechanism and the corrector evidence, and IVS15+5G>A with its loss-of-expression mechanism. Curating them separately is the point, since they fall in different therapeutic classes. Also took the three non-blocking suggestions - the organification defect regraded OBLIGATE within its subtype with the definitional reason stated and the test-sensitivity point left where it belongs, target_mechanisms dropped from the surveillance treatment since monitoring does not alter a mechanism, and the three gene-transfer treatments rebound from NCIT:C15986 to NCIT:C15238 per the backfill table in CLAUDE.md. Snippets 59 to 62, all verified.
--- a/kb/disorders/Pendred_Syndrome.yaml
+++ b/kb/disorders/Pendred_Syndrome.yaml
@@ -505,7 +505,7 @@ phenotypes:
       spectrum, alongside EVA rather than instead of it.
 - category: Endocrine
   name: Iodide Organification Defect
-  frequency: FREQUENT
+  frequency: OBLIGATE
   subtype: Pendred syndrome
   description: >-
     Partial failure to incorporate iodide onto thyroglobulin, demonstrated by a
@@ -514,6 +514,12 @@ phenotypes:
     inner-ear disease is identical, and it is the reason this entry curates the
     two as subtypes of one disease rather than as one disease with an incidental
     thyroid finding.
+
+    Graded OBLIGATE because the subtype is defined by it: a patient without a
+    demonstrable defect is DFNB4 by definition. That is a statement about the
+    definition, not about test sensitivity - the defect is partial and a
+    confirmed patient can discharge normally, which is why the diagnosis entry
+    records the perchlorate test's limits separately.
   phenotype_term:
     preferred_term: Thyroid defect in oxidation and organification of iodide
     term:
@@ -684,13 +690,11 @@ genetic:
     a statement about a population rather than about the gene.
 
     Founder alleles sharpen that further. In Okinawa, IVS15+5G>A and H723R
-    together account for roughly two thirds of alleles, and H723R is precisely the
-    trafficking-class variant that the pendrin corrector restores - so which
-    emerging therapy is relevant to a patient is partly a question of ancestry.
-    Individual variants are not curated in structured slots here, because the
-    schema's variant machinery hangs off has_subtypes and this entry's subtypes
-    are the syndromic/nonsyndromic split rather than allelic classes; the allelic
-    facts live in this note and in the corrector discussion instead.
+    together account for roughly two thirds of alleles, and the two act by
+    different routes - one abolishes expression, the other misfolds a competent
+    protein. Both are curated as structured variants below rather than described
+    only here, because which emerging therapy is relevant to a patient turns on
+    which class their alleles fall into.
   evidence:
   - reference: PMID:34345941
     reference_title: Genetic architecture and phenotypic landscape of SLC26A4-related hearing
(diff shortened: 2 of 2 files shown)


aleixpuigb:
Pushed `630d402a4`. All ten IMPORTANT items and four of the five suggestions are addressed, entirely from caches already committed to this branch.

**Finding 1 is the one that should have been caught before the PR opened.** The entry's scope decision, its `notes`, and the PR body all turn on "whether an organification defect is demonstrable" — and the defect was curated nowhere. That is an entry arguing from a fact it does not contain. It now has a phenotype (`HP:0008263`) and a perchlorate discharge test under `diagnosis`, with the specificity limit recorded: PMID:21274344's four siblings have goitre, hypothyroidism, deafness and a positive discharge test with no inner-ear malformation, so a positive test does not by itself make the diagnosis. That reference is graded `PARTIAL` and cited to bound the phenotype, not to support it.

Everything else:

| Finding | What changed |
|---|---|
| 2. Prevalence | Added, 7.5–10 per 100,000, keeping the source's own non-African qualifier |
| 3. GeneReviews | Transmission risk added to `inheritance`; surveillance schedule added as a treatment |
| 4. Imaging term | `NCIT:C25294` Laboratory Procedure → `NCIT:C16502` Diagnostic Imaging Testing |
| 5. IPII | Added, `HP:0000376` |
| 6. Transport-dead class | Now evidenced from PMID:22116358 |
| 7. Pathograph → phenotypes | Edges from four mechanism nodes; graph is one component with the phenotypes |
| 8. Dietary iodine | `environmental` block, `MODULATES` → goitre node |
| 9. Prenatal gene transfer | Added as a treatment and into the window discussion |
| 10. Variant-level genetics | Founder alleles and the CNV/deep-intronic fractions added |

Snippets 39 → **59/59 verified**. All seven offline gates pass.

**On the surveillance schedule (finding 3)** — you called the audiometry interval "arguably the single most mechanism-relevant management fact in the GR abstract" and I think that is right, so I gave it its own treatment entry rather than a line under habilitation. A three-to-six-month interval is the operational consequence of the entry's central claim that this is a fluctuating, step-wise loss: a hearing level measured once is not a trajectory.

**On the iodine predicate (finding 8)** — I used `MODULATES` rather than `PROTECTS_AGAINST`. The sources say the thyroid phenotype "may be modified by the nutritional iodine intake" and that goitre appears "especially if iodine is not routinely included in the diet". Neither states that iodine prevents goitre, and `PROTECTS_AGAINST` would claim more than the sentences carry. I note this means the edge does not count as causal under `qc_plugins.CAUSAL_PREDICATES` — that seems correct here rather than a loss.

**Two things your review found that my own gates then found again**, which is worth recording: a leftover placeholder snippet (`'Pendred syndrome'`, from a patch script that aborted before filling it) and a missing entry-level `evidence` block on the new `environmental` entry. `check-snippet-length`, `check-title-snippets` and `check-environmental-evidence` all failed on the first run of this revision. Both fixed, neither baselined.

**Not done — the suggestion about subtype `disease_term`s.** DFNB4 does have OMIM #600791, but I could not establish which MONDO term is the right bindable equivalent without an ontology lookup that risks exactly the wrong-concept binding this PR's `notes` are about. Left unbound rather than guessed. Happy to take a pointer.

I also took your note about the three off-topic caches (`PMID_11734249` Red Cross, `PMID_15347930` dental composites, `PMID_17997835` riboswitches) — those came from my own bad PMID recall early on, not from the DR report, and I've recorded that in `notes` so they read as discarded rather than overlooked.

Re-requesting review.

---

aleixpuigb:
Pushed `aa6a0f701`. You were right, and I verified it against the schema before acting rather than taking it on your word:

```
Genetic -> [... inheritance, variants, features, notes, examples]
Subtype -> [... children, genes, subtype_frequency, inheritance]      # no variants
251 disorder entries use Genetic.variants
```

**The sentence was false and I wrote it confidently.** The worse half is what you identified: I omitted content on the strength of a schema claim I never checked — and the omitted allele was p.His723Arg, the one the entry's own `pendred_variant_class_determines_therapy` discussion depends on. The therapeutic argument was resting on something queryable only as free text.

Both founder alleles are now `Genetic.variants` records, and curating them *separately* turned out to be the substantive point rather than bookkeeping: H723R misfolds a competent protein, so a chaperone has a target; IVS15+5G>A abolishes expression, so it does not. Same gene, same founder population, different therapeutic class. The ancestry/therapy-relevance prose you asked me to keep is now the H723R `description`, where it does more work than it did in a notes paragraph.

All three suggestions taken:

- **`OBLIGATE`** for the organification defect — you read my intent correctly, and they are two different claims. The subtype is *defined* by a demonstrable defect, so within it the frequency is obligate by definition; the imperfect test sensitivity is a separate point and stays in the `diagnosis` entry. The phenotype description now says which of the two it is asserting.
- **Dropped `target_mechanisms`** from surveillance rather than keeping the least-bad enum value. My own description said it "detects change rather than altering it", so `MODULATES` was claiming something the next line denied. An absent link is more honest than a stretched one.
- **`NCIT:C15238`** on all three gene-transfer/editing treatments, per the backfill table.

Snippets 59 → **62/62**. All seven gates pass, no baseline changes.

Thanks for the three explicit endorsements — the `MODULATES`-over-`PROTECTS_AGAINST` call in particular is one I'd have been tempted to "upgrade" on a later pass, and having the reasoning recorded as reviewed makes that less likely.

---

Issue #9876, opened 2026-08-28
CI red on main: readout target typo in Primary_Bile_Acid_Malabsorption blocks every PR


cmungall:
`tests/test_data.py` is failing on `main`, which means no PR in the repo can reach a green `test (3.13)` until it is fixed. The cause is a one-line typo in a single disorder entry.

**The failure**

```
FAILED tests/test_data.py::test_model_readout_targets_match_link[kb/disorders/Primary_Bile_Acid_Malabsorption.yaml]
AssertionError: Model readout target drift in Primary_Bile_Acid_Malabsorption.yaml:
  animal_models[0].modeled_mechanisms[1].readouts[0].target='Contracted Bile Acid Pool and Compensatory Hepatic Synthesis'
  != link target 'Contracted Bile Acid Pool'
```

1 failed, 37,412 passed.

**Why it's wrong**

A readout's `target` has to repeat its link's `target`. The value here is not a pathophysiology node at all — it reads as two real node names run together:

- `Contracted Bile Acid Pool` ← the link's target, and a real node
- `CYP7A1 Disinhibition and Compensatory Hepatic Synthesis` ← also a real node

For reference, the file's actual node names are:

```
Loss of ASBT Bile Acid Transport Function
Failure of Ileal Bile Acid Reclamation
Colonic Bile Acid Spillover and Secretory Diarrhoea
Contracted Bile Acid Pool
Loss of FXR-FGF19 Signalling from the Ileum
CYP7A1 Disinhibition and Compensatory Hepatic Synthesis
Steatorrhoea and Fat Malabsorption
Fat-Soluble Vitamin Malabsorption
Reduced Plasma Cholesterol
```

The link's target is the valid one, so the readout is the side to correct.

**Fix**

In `kb/disorders/Primary_Bile_Acid_Malabsorption.yaml`, `animal_models[0].modeled_mechanisms[1].readouts[0]`:

```yaml
-   target: Contracted Bile Acid Pool and Compensatory Hepatic Synthesis
+   target: Contracted Bile Acid Pool
```

Worth a quick look at whether the readout was *meant* to describe the compensatory-synthesis node instead — if so, the link and readout should both point there rather than just renaming the readout.

**Provenance**

Introduced in `8e3410d0d` (#9781). Reproduced against `origin/main`'s own copy of the file, so it is not specific to any branch. Surfaced while working on #9462, whose diff does not include this file.

aleixpuigb:
[...]

PR #9893, opened 2026-08-28
Curate harderoporphyria (MONDO:0030048)

### LINKED ISSUE #9874: Curate harderoporphyria (MONDO:0030048)
Curate a dismech entry for **harderoporphyria** ([MONDO:0030048](https://monarchinitiative.org/MONDO:0030048)) — a rare biallelic CPOX phenotype, classically homozygosity for p.Lys404Glu, in which coproporphyrinogen oxidase stalls after the first of its two sequential decarboxylations. The intermediate harderoporphyrin accumulates and the disease presents in the neonate as haemolytic anaemia with jaundice, not as the acute neurovisceral attacks of hereditary coproporphyria.

Preliminary read: this looks like a `DISEASE`, despite MONDO placing it under CPOX-related hereditary coproporphyria. Same gene, but the pathograph differs at the step that matters: HCP is heterozygous haploinsufficiency producing hepatic ALAS1 derepression and acute attacks, whereas harderoporphyria is a biallelic, position-specific partial block that leaves a distinct intermediate and an erythropoietic, haemolytic presentation. That is the split criterion — a different conserved mechanism, not a severity band on the same one.

The existing `Inherited_Porphyria` entry carries HCP as a CPOX subtype; this entry needs to say explicitly how it differs so the two do not read as duplicates. Cross-check against `kb/modules/heme_biosynthesis_porphyria.yaml` for a `conforms_to` node where the shared enzymatic block applies.

Stub: `stubs/Harderoporphyria.yaml`

aleixpuigb:
New disease entry for harderoporphyria, the biallelic CPOX phenotype that presents as neonatal haemolytic anaemia rather than as acute porphyric attacks.

## The mechanism, and why it justifies a separate entry

Coproporphyrinogen oxidase decarboxylates its substrate twice, and between the two removals the intermediate harderoporphyrinogen has to be held in the active site. Harderoporphyria is what happens when a variant lets the enzyme finish the first cycle and not the second — the intermediate is released instead of retained, and dominates the faecal porphyrins where in hereditary coproporphyria it is a minor component.

**This entry disagrees with MONDO's placement**, which puts harderoporphyria under CPOX-related hereditary coproporphyria. The argument is curated as an `INTERPRETATION` discussion rather than left implicit, so a reviewer can engage with it directly:

- The mechanism differs at the step that matters. HCP is a partial deficiency with substrate backing up; this is a redirection of the reaction. A crystal structure of human CPOX locates the lesion in a type I β-turn whose function is substrate retention between cycles.
- Severity does not order the two. Residual enzyme activity spans **1%–64%** across expressed HCP variants with clinically uniform presentation — so "less activity" does not produce harderoporphyria.
- Biallelism alone is not the criterion either. Biallelic CPOX variants *outside* the retention region give a "homozygous HCP" phenotype that is distinct again.

Orphanet reads it the same way, and the check that establishes this is the classification level rather than the identifier: in ORDO, `Orphanet_659672` has asserted parents `Disorder` (ORDO:557493) and `Disease` (ORDO:377788) — not `Subtype of a disorder` (ORDO:557494), which exists as the contrasting class — and hereditary coproporphyria is not among its ancestors. An earlier version of this section claimed the same thing on the strength of the `MONDO:equivalentTo` cross-reference alone, which establishes only the code; MONDO itself gives this concept its own term *while* placing it under HCP, so a distinct code settles nothing. Recorded as an `external_assertions` record rather than a mapping, because the two resources disagree about the hierarchy.

If the project prefers this as a subtype of the existing `Inherited_Porphyria` entry, say so and I'll move it. That entry is not modified here.

## The entry carries the evidence against itself

Review caught that the 1983 paper which named this disease — cited four times here — explicitly *rejects* a second-decarboxylation-specific active-site lesion, on the grounds that rat liver metabolises both substrates at the same active centre. I had quoted only the sentences that agreed with me.

That sentence is now a `REFUTE`-graded evidence item on the central node, with the resolution stated rather than assumed: a single shared active centre and a cycle-specific *retention* failure are compatible, which is what the 2005 structure showed and what the kinetics alone could not distinguish. The 1983 inference was the strongest reading available before the structure existed.

Review also caught a plain overstatement — "the enzyme is not slow, it is stopping in the middle" — while the same source reports Km 15–20× normal and Vmax half normal. The claim is now that the deficit is *unequal between the two cycles*, not absent.

## The mouse section, and the correction that reshaped it

Two murine *Cpox* models are curated as `FAILS_TO_RECAPITULATE` against `Harderoporphyrin Accumulation`, with a `HUMAN_MODEL_MISMATCH` discussion.

They originally said the mice do not accumulate harderoporphyrin. Review caught that **none of the cited papers measured it.** PMID:36967721's panel is uroporphyrin, coproporphyrin, protoporphyrin, PBG and δ-ALA; PMID:23631845 §2.5 is coproporphyrin III alone; PMID:28600349 is abstract-only. The word does not appear in either full text. I had turned an unmeasured analyte into a reported null and built the discussion on it.

The claim now rests on allele position plus the coproporphyrin-dominant *reported* profile, and says plainly that the diagnostic species has never been assayed in a *Cpox* mouse — which is the sharper statement. Underpinning it: PMID:36967721 supplies the species offset itself (mouse 380 ≡ human 391), and the human variant at that homologous residue, p.Arg391Trp at 22% residual activity, causes ordinary HCP. A human allele at the same position gives the same phenotype the mouse gives. Two `proposed_experiments` follow — assaying harderoporphyrin in the strains that already exist, then the retention-region knock-in.

This is the general gap filed as #9921: an `EvidenceItem` cannot distinguish *measured and absent* from *never measured* from *not reported*, so all three serialize identically and every check stayed green on a claim that was false.

## The pattern this PR kept hitting

Worth stating plainly, because it is the useful output of this review cycle. Four claims in this entry asserted more than their evidence supported:

| Claim | What was actually checked |
|---|---|
| "the enzyme is not slow, it is stopping in the middle" | the same source's Km/Vmax, which show it is also slow |
| "the report supplied no citable content" | nothing — nine fetched caches went unread |
| "none accumulates harderoporphyrin" | coproporphyrin was elevated; the analyte was never assayed |
| "Orphanet codes it as its own disorder" | a cross-reference, which establishes the code and not the placement |

Three of the four were **true claims resting on the wrong warrant**, which is what made them hard to self-catch: the conclusion survives review of the conclusion and fails only on review of the evidence beneath it. All four are corrected, and in each case the corrected version is the stronger one.

## Module conformance, with a caveat on the record

Three nodes declare `conforms_to` against `heme_biosynthesis_porphyria`. The accumulation node conforms **structurally while differing in substance**: the module describes intermediates accumulating *upstream of a block*, whereas what accumulates here is a reaction intermediate released mid-catalysis. The conformance is still right — same position in the causal chain — but the module's description does not cover this case, and that is flagged in `notes` rather than papered over.

## What this entry says it does not know

- The step from accumulated harderoporphyrin to red cell destruction. The compartment is established; the mechanism is not. `PROVISIONAL`, with an `INDIRECT_UNKNOWN_INTERMEDIATES` edge and a `KNOWLEDGE_GAP` discussion on why the distinction matters — a photo-oxidative mechanism should respond to light protection, an erythroblast heme-synthesis failure should not.
- A documented exception to this entry's own diagnostic rule: a 2017 neonate had the harderoporphyria clinical picture with an HCP porphyrin profile and no K404E. Recorded as an `OPEN_QUESTION` so a future curator knows the rule has a known counterexample rather than reading it as tested.
- Prevalence is `CASES_IN_LITERATURE` with no rate. The `prevalence_class` band is an inference from a reported count of roughly eight cases, and the record's `notes` says so.

## Deep research

Two runs, and the history matters more than the artifact.

The `openscientist` run commissioned before this entry was written **timed out at 3600s and was cancelled by the client**. An earlier version of this PR body said it "had not returned", which was true but soft enough to read as a choice; it was a failure, and every reference in the entry was located independently through PubMed as a result.

A `claude_code` run was commissioned during review and **is committed on this branch** with its citations sidecar and the nine reference caches its validation fetched. An earlier round of this PR claimed it "supplied no citable content" — false, asserted without reading those nine caches, and corrected in `notes` alongside how the wrong claim got made.

Its own term validation flagged two wrong-concept identifiers it proposed (`HP:0032263` named "Coproporphyrinuria", actually *Increased blood pressure*; `UBERON:0001155` named "liver", actually *colon*). Neither is used. Both resolve and would pass an existence check; only the label comparison catches them.

## Validation

Against head `01231825d`:

- `just validate` — schema, terms, **46/46 reference snippets** verified against cached references
- `just validate-disorders` — passes
- all seven ungated offline gates pass, no baseline changes
- `just validate-history` — five history records, clean

One pre-existing `pytest` failure was reported against earlier heads: `test_model_readout_targets_match_link[Primary_Bile_Acid_Malabsorption.yaml]`, from #9781 and tracked as #9876. It was fixed on `main` by #9875 (`5c347e236`). A branch whose merge-base predates that commit still reproduces it locally while CI passes, because CI tests the merge — which is what happened here.

Closes #9874

### DIFF
--- a/history/disorders/Harderoporphyria/2026-08-28T194654Z-claude-code-1b3d89.yaml
+++ b/history/disorders/Harderoporphyria/2026-08-28T194654Z-claude-code-1b3d89.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Harderoporphyria
+  path: kb/disorders/Harderoporphyria.yaml
+session:
+  id: 2026-08-28T194654Z-claude-code-1b3d89
+  timestamp: '2026-08-28T19:46:54Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9893
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - external_assertions
+  summary: Ground the Orphanet claim in ORDO classification level, not the xref
+  details: |
+    Review caught that the Orphanet external_assertion description asserted a hierarchy I had not checked. I wrote that Orphanet codes harderoporphyria as its own disorder rather than as a form of hereditary coproporphyria, but what I had actually verified was only that ORPHA:659672 is an equivalentTo cross-reference on MONDO:0030048. A distinct code does not establish a classification level, and MONDO itself gives this concept its own term while placing it under HCP, so the xref settles nothing about placement either way. This is the fourth instance in this PR of asserting more than the check supported. Checked it properly: in ORDO, Orphanet_659672 has asserted parents Disorder (ORDO:557493) and Disease (ORDO:377788), with Clinical entity (ORDO:C001) above them, and not Subtype of a disorder (ORDO:557494), which exists as the contrasting class. Hereditary coproporphyria is not among its parents or ancestors. So the claim is true, but for a different reason than the one I gave. The description now states the classification level and the parents that establish it, and says explicitly that the cross-reference alone would not have. Verified against the OLS ORDO API. Validation: schema, terms and 46 of 46 snippets pass.
--- a/kb/disorders/Harderoporphyria.yaml
+++ b/kb/disorders/Harderoporphyria.yaml
@@ -40,10 +40,14 @@ external_assertions:
   external_id: ORPHA:659672
   url: https://www.orpha.net/en/disease/detail/659672
   description: >-
-    Orphanet codes harderoporphyria as its own disorder rather than as a form of
-    hereditary coproporphyria, which is the same reading this entry takes and the
-    opposite of MONDO's placement. Recorded as the external assertion rather than
-    as a mapping because the two resources disagree about the hierarchy.
+    Orphanet classifies harderoporphyria at the disorder level rather than as a
+    subtype: in ORDO its asserted parents are Disorder (ORDO:557493) and Disease
+    (ORDO:377788), not Subtype of a disorder (ORDO:557494), and not hereditary
+    coproporphyria. That is the same reading this entry takes and the opposite of
+    MONDO's placement, so it is recorded as an external assertion rather than as a
+    mapping - the two resources disagree about the hierarchy. Note that the
+    equivalentTo cross-reference alone would not have established this; the
+    classification level is what does.
 - name: OMIM harderoporphyria record
   source: OMIM
   assertion_type: disease_record


aleixpuigb:
Pushed `e07068bcc` with items 2, 3 and 4 plus three suggestions. **Item 1 is in progress and I'll explain why it wasn't just an oversight.**

**Item 3 is the best catch of this whole run.** The 1983 paper explicitly ruled out a second-decarboxylation-specific active-site lesion — which is what this entry's central node asserts — on the grounds that rat liver metabolises both substrates at the same active centre. I cited that paper four times and quoted only the sentences that agreed with me. The entry's whole posture is surfacing disagreement, and I skipped the one place it pointed at my own thesis.

It is now carried as a `REFUTE`-graded evidence item on the node, with the resolution stated rather than assumed: a single shared active centre and a cycle-specific *retention* failure are compatible, which is what the 2005 crystal structure showed and what the kinetics alone could not distinguish. The 1983 inference was the strongest reading available before the structure existed.

**Item 2 was a plain overstatement.** "The enzyme is not slow, it is stopping in the middle" — while the same source reports Km 15–20× normal and Vmax half normal, and the entry itself annotates the activity `DECREASED` and records 10% of control. The file asserted both things. The claim now is that the deficit is *unequal between the two cycles*, not absent.

**Item 4** — module conformance completed on all three nodes. You were right that the phototoxicity node was the one that mattered: without `GO:0006979` and keratinocyte, its own description ("damage skin") was an assertion rather than a mechanism.

Suggestions taken: the absence of abdominal pain and neuropsychiatric symptoms as the sharpest HCP discriminator (into the split discussion), an erythrocyte-harderoporphyrin `biochemical` record, the intron-6 splice allele — with the point that a null *in trans* doesn't dilute the phenotype, it exposes it — and `Jaundice` recategorised from `Hepatic` to `Hematologic`. Snippets 28 → **33/33**.

**On item 1.** The `openscientist` run was commissioned before this entry was written; it **timed out at 3600s and was cancelled by the client**, which is why there was no artifact rather than my having skipped it. That is a materially different failure from "promised a follow-up commit", and I should have said so in the PR body instead of the softer wording I used. A `claude_code` run is executing now and I'll commit the report and its sidecar, plus anything it adds, as a separate commit on this branch.

**Not taken: the Orphanet suggestion.** `ORPHA:100924` resolves to *Porphyria due to ALA dehydratase deficiency*, not harderoporphyria — your "verify the code" caveat was right. I probed several other codes without finding a harderoporphyria record, so the `CASES_IN_LITERATURE` prevalence with its inference stated in `notes` stands. If you have the correct ORPHA code I'll take it, since the phenotype frequencies would be a real improvement over gradings resting on presence statements.

The GeneReviews *Hereditary Coproporphyria* question I'll fold into the deep-research pass, since PubMed is reachable from here even though it isn't from your sandbox.

---

aleixpuigb:
Pushed `6f5bb4a33`. **Item 1 is now closed** — the deep-research artifact and its citations sidecar are on the branch, along with the nine reference caches its own validation fetched (all on-topic for CPOX/porphyria).

Two things about it worth recording rather than glossing.

**The first run failed in a specific way.** `openscientist` timed out at 3600s and was cancelled by the client — not skipped, not deferred. My original PR body said the report "had not returned", which was true but soft enough to read as a choice. A `claude_code` run then completed in about 20 minutes.

**The report changed nothing.** It corroborated the entry and supplied no citable content it lacked — including on prevalence, where it independently lands on "fewer than 10 cases reported worldwide", consistent with the `CASES_IN_LITERATURE` record and the ~8-case count already there. I've written that null result into the entry's `notes` rather than passing over it, because "the report added nothing" and "the report was never read" look identical in a diff, and only one of them is a defensible state.

Its own term validation flagged two more wrong-concept identifiers: `HP:0032263` named "Coproporphyrinuria" but actually *Increased blood pressure*, and `UBERON:0001155` named "liver" but actually *colon*. Neither is used. **That is now the third report in this three-disease run to propose a real identifier for the wrong concept** — after `UBERON:0002518` (otolith organ, offered as endolymphatic sac) and `CL:1000456` (peritoneal mesothelial cell, offered as collecting-duct cell). The fabricated identifiers all get caught; the plausible wrong ones only get caught by the name check.

Items 2, 3 and 4 were in the previous push (`e07068bcc`), with item 3 — the counter-evidence sentence I'd quoted around — as the sharpest finding of the whole run. Snippets **33/33**, all gates green.

Still open and not taken: the Orphanet suggestion. `ORPHA:100924` resolves to *Porphyria due to ALA dehydratase deficiency*, and I probed several other codes without finding a harderoporphyria record. If you have the right one, the phenotype frequencies would genuinely improve gradings that currently rest on presence statements.

I did not chase the GeneReviews *Hereditary Coproporphyria* chapter: the report cites it for HCP variant distribution, but this entry's whole argument is that harderoporphyria is not HCP, so mining an HCP chapter for it would need care I'd rather take deliberately than as a round-three addition. Flagging it as a real follow-up rather than declining it.

---

aleixpuigb:
Pushed `bf21e96dc`. Both findings taken, and **finding A is the sharpest single catch across all three PRs in this run.**

I claimed in the entry's `notes` that the report "supplied no citable content the entry lacked". That was false, and it was false in exactly the way the entry's first draft was false: I read the narrative, concluded it added nothing, and never opened the nine reference caches the run had fetched and that I committed in the same commit. Three of them held content the entry needed. I made the original under-consumption error a second time *while writing a note congratulating this entry for recording a null result honestly.*

**PMID:629747 is the keystone I was missing.** Not a supporting citation — the normal physiology my whole argument presupposes:

> "It is suggested that both decarboxylations take place at the same active centre, which becomes temporarily inaccessible … and that the molecule rotates after the first decarboxylation to allow the second to take place."

That is the direct answer to the 1983 objection, in the same rat-liver system the objection rests on, five years earlier — and I was arguing that reconciliation in prose alone while the source sat uncited in my own branch. The companion sentence, that the intermediate destined for protoporphyrinogen "does not equilibrate with added harderoporphyrinogen", is what makes *retention* a measurable compartment rather than a figure of speech. Both are now evidence items on the arrest node, and the node description states the physiology.

**The mouse models are better than a gap filled.** Three published murine *Cpox* alleles; all three accumulate coproporphyrin; none accumulates harderoporphyrin; they sit at residues 373 and 380, outside D400–K404. A mouse at 15% of wild-type activity shows clean HCP biochemistry rather than partial harderoporphyrin excess. That is this entry's position-not-dose argument arriving from a direction I hadn't looked in — and 15% sits squarely inside the 1–64% range that gives clinically uniform HCP in humans, which makes the two observations one observation. Curated as `animal_models` with `FAILS_TO_RECAPITULATE` links plus a `HUMAN_MODEL_MISMATCH` discussion; the open experiment is a retention-region knock-in, which nobody has published and which is why the erythroid-injury node stays `PROVISIONAL`.

**Finding B** — `ABSENT` → `DECREASED` on `GO:0004109`. You're right that it was the machine-readable half of the overstatement I'd only fixed in prose; the node's own evidence said `Vmax` half normal three lines below. It also resolves the same-term-in-two-states note.

Snippets 33 → **41/41**. All seven gates pass.

**One thing I did not fix, and want to flag rather than bury:** `pytest -k model_readout` now surfaces a failure in `kb/disorders/Primary_Bile_Acid_Malabsorption.yaml`. I checked — it came in with #9781, is untouched by this branch, and is not mine to fix from here. Worth someone's attention separately.

On the ORPHA code: thank you for correcting your own suggestion. I have not attempted `ORPHA:659672` because it needs `just refresh-orphadata` to pull the bulk XML, which rebuilds shared Orphanet state — the same scope objection I raised on #9886, and it applies with more force here since I'd be doing it to chase one prevalence band. Better as its own change alongside the `mappings` block.

---

aleixpuigb:
You were right, and I checked it rather than taking it on trust. Neither phenotyped strain measured harderoporphyrin. PMID:36967721's methods list uroporphyrin, coproporphyrin, protoporphyrin, porphobilinogen and δ-ALA; PMID:23631845 §2.5 is coproporphyrin III alone; PMID:28600349 is abstract-only and names only coproporphyrinogen III. The word "harderoporphyrin" does not appear anywhere in either full-text cache. I was reporting an absence of measurement as an absence of the analyte.

Pushed in `aabbf588b`.

**The blocking finding.** Reworded the `FAILS_TO_RECAPITULATE` `description` and `limitations` on both models and the `HUMAN_MODEL_MISMATCH` `rationale`. The murine claim now rests on allele position plus the coproporphyrin-dominant reported profile, and says plainly that the diagnostic species has never been assayed in a Cpox mouse. Your reading of it was right — it is the stronger paragraph. "Nobody has measured the defining species in any Cpox mouse" is a specific, checkable gap; "the mice don't accumulate it" was a claim I could not back.

I also went looking for something to put underneath the positional argument now that the measurement is gone, and PMID:36967721 had it two paragraphs above the methods I should have read the first time: mouse residue 380 is homologous to human 391, and the reported human p.Arg391Trp at 22% residual activity causes ordinary HCP. That is a human allele at the homologous residue giving the same phenotype the mouse gives — which is what the inference actually needs. Added as two evidence items.

**Also taken:**

- Allele count corrected. Nakano is the same p.Arg380Leu locus, which my own sentence said while calling it a third allele. Now: three alleles on record, two phenotyped, and the M100835 splice strain added as the unphenotyped one.
- `reference_title` for PMID:629747 → "in rat liver" (both occurrences).
- Elder 1978 regraded `MODEL_ORGANISM` → `IN_VITRO`. Synthetic ¹⁴C substrates against rat liver enzyme is a cell-free biochemical assay; CLAUDE.md puts that in `IN_VITRO`.
- Two `proposed_experiments` on the discussion. The knock-in you'd expect, and ahead of it the cheap one your finding implies: assay harderoporphyrin in the strains that already exist. One HPLC run converts the inference into evidence either way, and it should be first.
- Evidence added to the two terminal pathophysiology nodes.
- Dropped `references_cache/DOI_10.1007_s10545-010-9237-9.md`, a duplicate of `PMID_21103937.md` and unreferenced.

**One I could not do structurally.** ORPHA:659672 and OMIM:618892 are both `MONDO:equivalentTo` on MONDO:0030048 — verified against OLS, along with GARD, MedGen, MeSH and UMLS. `DiseaseMappings` has slots for ICD-10CM, ICD-11F, MONDO and NCIT only, so none of them has anywhere to go. They are in `notes` with the gap named, which matches what other entries do. There is also an `icd11.foundation` xref (1664486132), but that is a foundation entity ID rather than an MMS code and I did not want to guess the CURIE form into a validated slot. Filing the missing slots separately.

Validation: schema, terms, and 46/46 snippets verified. `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-environmental-evidence` all clean, no baseline touched. `just validate-disorders` passes. History record added.

---

aleixpuigb:
Correcting myself on one thing from my last comment, and closing out the two remaining suggestions. Pushed as `93ee34ae7`.

**I was wrong about the mappings.** I said ORPHA and OMIM identifiers had nowhere structural to go and put them in `notes`. `external_assertions` carries exactly that, with `external_id` required — and 121 ORPHA and 49 OMIM records already use it. I checked `DiseaseMappings`, found no ORPHA slot, and stopped there.

What makes it worth flagging rather than quietly fixing: the grep that would have corrected me is the same grep I ran to *support* the claim. I collected the free-text `ORPHA:` mentions in other entries' `notes` as evidence the gap was real, and never asked whether those same entries were also recording it structurally somewhere else. That is the same shape as the harderoporphyrin error — reading a partial view as a complete one — twice in one round. I filed #9922 on the false premise and have withdrawn it with the reasoning recorded.

Both are now `external_assertions`, and the Orphanet one earns its place independently: **Orphanet codes harderoporphyria as its own disorder, not as a form of hereditary coproporphyria.** That is an independent resource agreeing with this entry and disagreeing with MONDO, which is worth having structurally in an entry whose main argument is exactly that. Recorded as an external assertion rather than a mapping precisely because the two resources disagree about the hierarchy — a `broadMatch` to HCP would assert the placement the entry disputes.

**The pytest failure already has an issue.** #9876, filed by @cmungall at 16:22 today, with the diagnosis and the one-line fix (`animal_models[0].modeled_mechanisms[1].readouts[0].target` should be `Contracted Bile Acid Pool`). Open and unassigned, no PR yet. I reproduced it against this branch and confirmed it is unchanged; not filing a duplicate, and leaving the fix alone since the file is outside this PR.

**Also filed** #9921, from the blocking finding: an `EvidenceItem` cannot distinguish *measured and absent* from *never measured* from *not reported*. Those are three different things that serialize identically, and the failure mode is worse than the other evidence-metadata issues — the schema doesn't underdescribe the negative, it let me assert something false with every check green.

Validation after `93ee34ae7`: schema, terms, 46/46 snippets, duplicate-key and entity-ref checks all clean. History record added.

---

aleixpuigb:
Thank you for the approval, and for verifying the murine claim independently rather than taking the fix on trust — that is the right instinct given the entry's record on exactly this.

I've spent the approval on your first non-blocking note, because it is a defect I introduced rather than one that was already there. Pushed as `8fdcf0e69`.

**PMID:6886003's REFUTE item is now IN_VITRO.** You flagged it as the main non-blocking note, and it is worse than a stylistic inconsistency: your round-three review had said of PMID:629747 that it *"is graded consistently with the 1983 rat-liver items in the same file, so changing one means changing all of them; noting it rather than asking for it."* I then changed one. You predicted the failure and I walked into it in the same round.

Both items quote cell-free rat-liver enzymology, which is CLAUDE.md's `IN_VITRO` ("biochemical assays outside an organism"), so both are now `IN_VITRO`. The other nine PMID:6886003 items stay `HUMAN_CLINICAL` — they quote the human sibling study, which is what that paper mostly is, and splitting by sentence is what the mixed-source rule asks for.

I recognise this drops the approval you just gave, which is not something I did lightly. The alternative was merging an entry whose two citations of the same rat-liver experiments carry different `evidence_source` values, with the inconsistency documented in the thread — and `check-snippet-grading` would not have caught it, since it keys on the quoted sentence and these are different sentences in different papers. That is arguably its own small gap, and a narrower one than #9921.

Validation on `8fdcf0e69`: schema, terms, 46/46 snippets, `check-snippet-grading` clean with no baseline change, history record added. The diff from `93ee34ae7` is one word plus the history file.

Nothing else outstanding from my side. Your other two non-blocking notes I'm leaving as-is unless you'd rather I didn't.

---

aleixpuigb:
You caught the same error a fourth time, and this one was inside the commit that was fixing the third instance of it. Pushed as `01231825d`.

**The Orphanet claim.** You are right that I asserted a classification level on evidence that only established a code. What I verified was the `MONDO:equivalentTo` cross-reference; what I wrote was that Orphanet codes harderoporphyria as its own disorder rather than as a form of HCP. Your reasoning for why that does not follow is exactly right, and the MONDO counterexample makes it — MONDO gives this concept its own term while placing it under HCP, so a distinct code is consistent with either placement.

I had network access you did not, so I checked it rather than softening. In ORDO, `Orphanet_659672` has asserted parents:

```
ORDO:557493  Disorder
ORDO:377788  Disease
ORDO:C001    Clinical entity   (ancestor)
```

Not `ORDO:557494` *Subtype of a disorder*, which exists as the contrasting class — and hereditary coproporphyria is not among its parents or ancestors. So the claim holds, at the disorder-vs-subtype distinction you named, for a reason I had not checked when I made it.

The description now states the classification level and the parents that establish it, and says explicitly that the cross-reference alone would not have. That last clause is there so the next reader does not repeat the inference I made.

**Your first suggestion was already in** — `8fdcf0e69`, pushed before your review posted, regrades the PMID:6886003 REFUTE item to `IN_VITRO`. Both rat-liver items now match. That commit also cost me the approval you had just given, which I noted at the time; this one costs the next.

**Left alone, deliberately:** the R391W secondhand chain, which you documented rather than asked for and where you confirmed no better cached primary exists; the terminal-node name collisions; and `attaches_to` naming only BALB.NCT. Say the word on any of them.

**On the pattern.** Your closing note undercounted by one, and the miss is instructive: "the enzyme is not slow", "the report supplied no citable content", "none accumulates harderoporphyrin", and now "Orphanet codes it as its own disorder". Four times, in an entry whose author was by then explicitly watching for it. Three of the four were true claims with the wrong warrant, which is what makes them hard to self-catch — the conclusion survives review of the conclusion, and only fails review of the evidence underneath. That is the general shape of #9921.

Validation on `01231825d`: schema, terms, 46/46 snippets, history record added.

---

PR #9894, opened 2026-08-28
curate: add Spastic Paraplegia 81 (MONDO:0032905)

### LINKED ISSUE #9883: Curate spastic paraplegia 81, autosomal recessive (MONDO:0032905)
Curate a dismech entry for **spastic paraplegia 81, autosomal recessive** ([MONDO:0032905](https://monarchinitiative.org/MONDO:0032905)) — caused by biallelic variants in `SELENOI`, the selenoprotein ethanolamine-phosphotransferase that makes phosphatidylethanolamine and ethanolamine plasmalogens.

Preflight: MONDO:0032905 appears nowhere in `kb/` on `main`, and `SELENOI` appears in no existing entry's `genetic:` block. The KB has several other hereditary spastic paraplegia entries, so check whether SPG81 belongs as its own Disease or as a `has_subtypes` entry — my read is that a distinct enzymatic lesion in a distinct pathway argues for its own entry, but that is the call to make first. No open PR matches.

Stub: `stubs/Spastic_Paraplegia_81_Autosomal_Recessive.yaml`

nlharris:
Adds a Disease entry for SPG81 — an autosomal recessive complicated hereditary spastic paraplegia caused by biallelic loss of function in `SELENOI`, which encodes ethanolamine phosphotransferase 1 (EPT1), the final enzyme of the CDP-ethanolamine branch of the Kennedy pathway.

Closes #9883

## The thing to check first: the report's gene ID is wrong

The deep-research report committed alongside this entry gives the causal gene as **`HGNC:30396`**. That identifier is **SELENOS** — a different selenoprotein. SELENOI is **`hgnc:29361`**, which is what the entry uses.

Two things make this worth your attention beyond the fix:

- **The report's own term-validation block passed it.** It reports 39/40 terms verified. `HGNC:30396` resolves to a real gene, so existence checking cannot detect a symbol→identifier mismatch. Only cross-checking the symbol against the ID catches it, and nothing in the pipeline does that.
- **This is the second provider in two runs to make the same class of error**, both times landing on a real, *related* gene — the previous run's report gave `HGNC:6398` (KPNA5) for IRF9. That's what makes it dangerous: the wrong ID looks plausible precisely because it's in the right family.

The correction is written into the entry's top-level `notes` rather than silently applied, because the next curator reading that report will meet the same wrong identifier.

## Why this is its own entry and not an HSP subtype

The KB has several hereditary spastic paraplegia entries. SPG81 is curated separately because the lesion is a distinct enzymatic step in a distinct pathway — ether-lipid synthesis — rather than another allele of a shared axonal-transport or mitochondrial mechanism, and because the phenotype is complicated rather than pure. Reasoning is in `notes`.

## Pathophysiology

Five nodes, one root, every phenotype reached:

`Loss of EPT activity` → `Depletion of PE and plasmenyl-PE` → `Lipid peroxidation and failed oligodendrocyte maturation` → `Central hypomyelination` → a motor arm and an extramotor arm.

The mechanism that makes this more than a quantitative lipid deficiency is worth stating: plasmenyl-PE's vinyl-ether bond reacts preferentially with oxidants, so it is a sacrificial antioxidant. The species that is lost is the one that was protecting the membrane.

**The extramotor arm is deliberately weaker than the motor arm, and its description says so.** Deafness, visual loss, seizures and microcephaly co-occur with the myelin lesion in reported patients, but no source establishes that each follows from it — and sensorineural deafness is a cochlear or eighth-nerve phenotype whose relation to central hypomyelination is undemonstrated. The edges assert co-occurrence within one CNS disorder, not a worked-out causal path.

A `KNOWLEDGE_GAP` attached to that node records why it matters clinically rather than just taxonomically: the mouse model is nervous-system-restricted and was assessed for motor deficits, hypomyelination, gliosis and microcephaly — not for auditory or visual function. So the model grounding the motor arm is silent on the arm carrying most of the severe-end morbidity, and a therapy rescuing myelination would not be expected to address it.

## No frequencies, and the reason

Three consanguineous families have been reported since 2017, plus a fourth carrying a VUS. There is no cohort and no natural-history series, so **no phenotype carries a `frequency` value**. The problem is the absence of a denominator, not the number of families, and that is stated in `notes` rather than left as an unexplained gap.

## Genetics — the clinically material point

One reported allele is a **synonymous** variant that disrupts splicing and skips exon 2. A synonymous change in `SELENOI` is not automatically benign, and a filter keyed on protein consequence would drop it. That's in `genetic.notes` with its source.

## Evidence

24/24 snippets verified. Curated from an OpenScientist report (22 min, 14/14 references verified, `confabulation_rate: 0.0`, 8/14 scored on topic). **Every quote taken from the report was re-verified against the primary abstract in `references_cache` before use, and all 12 matched exactly** — noticeably better quote fidelity than the previous provider, which is what made the gene-ID error the only correction needed.

## Validation

`just validate` (schema + terms, including the corrected `hgnc:29361`), `just count-verified-snippets` (24/24), `just check-entity-refs`, `just check-duplicate-keys` — all pass. A reachability audit confirms one root, no dangling targets, and no unreached phenotype.

Cache diff is four rows: the two GO terms, the MONDO term, and the gene.

---

### DIFF
--- a/kb/disorders/Spastic_Paraplegia_81.yaml
+++ b/kb/disorders/Spastic_Paraplegia_81.yaml
@@ -61,9 +61,9 @@ notes: >-
   Scope. Several other hereditary spastic paraplegia entries exist in this
   knowledge base. SPG81 is curated as its own Disease rather than a subtype
   because the lesion is a distinct enzymatic step in a distinct pathway
-  (ether-lipid synthesis) rather than a further allele of a shared axonal-
-  transport or mitochondrial mechanism, and because the phenotype is complicated
-  rather than pure.
+  (ether-lipid synthesis) rather than a further allele of a shared
+  axonal-transport or mitochondrial mechanism, and because the phenotype is
+  complicated rather than pure.
 external_assertions:
   - name: OMIM spastic paraplegia 81 record
     source: OMIM


nlharris:
All six items plus prevalence and the treatments question, in `c77ec2c04`. Your framing was the useful part: three of the six were claims the entry *argued* while leaving the supporting quote uncited in a cache this PR already commits. That's a worse failure than a missing reference, because the entry reads as sourced.

## The three "already had the evidence" items

**Item 2 — plasmenyl-PE depletion.** You're right that the node rested on a mouse chemistry statement while `PMID:29500230` carries the direct patient-cell measurement. Now cited. I quoted it through the compensatory clause deliberately — `most plasmenyl-PE species were significantly decreased in the patient's cells, whereas most plasmanylcholine` — because the rise in plasmanyl-PC makes this a redistribution, not a simple loss, and that's a different claim than the node was making.

**Item 3 — fibroblast assay.** The biochemical entry named the assay as the informative readout while citing only the sentence explaining why *blood* isn't usable. The paper that actually ran the assay is now quoted.

**Item 5 — cleft palate and cerebral atrophy.** Both were inside snippets already quoted five times in the file. Added as phenotypes and reached from the pathograph. Cleft palate's description notes it's the feature least obviously explained by a myelination defect, which is part of what the extramotor gap is about.

## Item 6 was a grading error, and I want to be explicit about it

You flagged `evidence_source: HUMAN_CLINICAL` on the enzymatic-activity snippet as suspect. Checked: `PMID:28052917` is MeSH-indexed to *Saccharomyces cerevisiae*. The activity number comes from a heterologous assay.

The distinction I'd got wrong: **the variant is human, the measurement is not.** I graded on the provenance of the allele rather than of the experiment, which is exactly what `evidence_source` is not for. Regraded `IN_VITRO` with the reason in the `explanation` so the next reader sees why a paper about a patient's variant carries an in-vitro grade.

## Items 1 and 4

**Location** — brain and cerebellum added alongside spinal cord. I kept spinal cord rather than replacing it, since the corticospinal tract is the motor arm's substrate, but you were right that a spinal-cord-only location contradicted the node's own brain-imaging evidence and its cerebral children.

**Animal model** — two `ModelMechanismLink`s. The `PARTIALLY_RECAPITULATES` one is where the argument gets structural, and it's the fix I'd have wanted most: its `limitations` state that the murine assessment covers motor deficits, hypomyelination, gliosis and microcephaly, and is *not reported* to cover auditory or visual function. That is precisely what the extramotor `KNOWLEDGE_GAP` asks about, and having it in a field rather than prose means the gap and the model now point at each other.

## Prevalence and treatments

**Prevalence** added as `CASES_IN_LITERATURE` / `ULTRA_RARE`. You correctly separated this from the phenotype-frequency argument — different slot, different claim. I deliberately did not convert to `rate_per_100000`: there's no denominator to convert against, and a fabricated rate is what the last run's review caught me doing in the opposite direction.

**No treatments section, and the reason is now in `notes` rather than the omission being silent.** Management is symptomatic — antispasticity agents, PT/OT/speech, seizure control, sensory and orthopaedic support — and that *is* a curatable fact. It isn't recorded because none of the four cached references states it: all four are molecular or case-level papers, and I could not locate a management source quotable exactly. Asserting a treatment list without a citable source is the failure mode these rules exist to prevent, so the section stays absent with the reason stated and an invitation to add one.

## On the GeneReviews check

Thank you for verifying that the only HSP GeneReviews is the *Uncomplicated (Pure)* overview and that it excludes complicated HSP and never mentions SELENOI. That's the check I'd have had to do myself and it's the one that most often turns into a blocking finding.

## Validation on `c77ec2c04`

```
just validate                 → schema + terms clean
just count-verified-snippets  → 32/32 (was 24/24)
just check-duplicate-keys     → OK
check-snippet-length / -title-snippets → no new baseline entries
```

Reachability audit: one root, no dangling targets, no unreached phenotype, and every model readout's `target` matches its link.

---

Issue #9787, opened 2026-08-28
Follow-up: deferred non-blocking items from the round-5 curation PRs (#9755-#9759)


caufieldjh:
Tracking issue for the non-blocking suggestions raised during review of five curation PRs and deliberately deferred rather than pushed. Opened at the reviewer's request on #9756, widened to the whole round so the items are in one place.

Each was deferred for one of two reasons: the PR was already approved and pushing would have dismissed the approval, or the item needs a reference fetch rather than an edit. None is a defect in merged content.

## Lathosterolosis (#9756)

- [ ] **Q72\* nonsense allele.** `PMID:33204591` — cited five times in the entry already — reports `c.212A>T p.(Asn71Ile)` in trans with `c.214C>T p.(Gln72*)`. Adding it strengthens the entry's own residual-activity-tracks-with-severity argument, since Q72\* is a true null found in a patient with significant liver disease.
- [ ] **`HUMAN_MODEL_MISMATCH` for the Sc5d-null mouse.** The analysis is already written in the model's `limitations` — homozygous nulls are stillborn, so the model cannot speak to the postnatal course, hepatic disease or cataracts. It needs relocating into a `discussions` entry, not new research.
- [ ] **Normal plasma cholesterol as the SLOS discriminator.** Total plasma cholesterol is typically normal in human lathosterolosis, which is what separates it biochemically from SLOS. Since the entry's framing is "one step upstream of SLOS", noting that the cholesterol shortfall is a developmental/cellular claim rather than a measurable plasma one would sharpen the comparison.
- [ ] **Remaining GeneReviews content:** genitourinary anomalies (the `Congenital Malformation Pattern` description already names the urogenital system without a phenotype behind it), the surveillance schedule (annual ophthalmology, six-monthly FibroScan, plasma sterol before and during simvastatin), and global developmental delay `HP:0001263`.

## TRMA (#9757)

- [ ] **`supports: SUPPORT` → `PARTIAL`** on the domain-level GeneReviews evidence for `Seizures` and `Stroke`. The sentence names *neurologic manifestations* as a managed domain; the claims are the two specific phenotypes. The entry already grades the erythromycin EHR item `PARTIAL` for exactly this reason, so this is an internal inconsistency.
[...]

PR #9899, opened 2026-08-28
curate(lathosterolosis): clear the #9787 follow-up items

caufieldjh:
Clears the four Lathosterolosis items deferred from the #9756 review and tracked in #9787.

Each was deferred because the PR was already approved and pushing would have dismissed the approval. None was a defect in merged content — these are additions.

## What changed

**`genetic`: the Q72\* nonsense allele.** `PMID:33204591` was already cited five times in this entry, but its variant sentence had never been mined. Q72\* is the one reported allele that cannot encode a functional enzyme, carried in trans with the missense N71I by the patient whose FibroScan read 15.6 kPa. The entry's genotype–phenotype reading rests on residual activity, so the allele with none is the informative extreme — it anchors the severe end of the argument D210E already anchored at the mild end.

**`biochemical`: plasma total cholesterol, `presence: Normal`.** This is a negative result that does real work. The entry frames the disease as one enzymatic step upstream of SLOS and attributes the malformation arm to a cholesterol shortfall, which invites the reading that plasma cholesterol should be low. It is not — and the normal 7-dehydrocholesterol in the same profile is what separates the two diseases biochemically. The cholesterol claim in the pathophysiology is a developmental and cellular one, about the sterol available to hedgehog signalling in the embryo, not a claim about a measurable plasma concentration. The entry now says so where a reader would look for it.

**`phenotypes`: two GeneReviews additions.**
- Global Developmental Delay (`HP:0001263`, `FREQUENT`). Named first in the GeneReviews characterisation sentence; the frequency band is carried by a separate quantitative item ("in all surviving patients") rather than by the characterisation sentence alone.
- Genitourinary Anomalies (`HP:0000119`, **no** frequency). The `Congenital Malformation Pattern` node already named the urogenital system with no phenotype behind it. GeneReviews places genitourinary anomalies in Management, not Clinical Characteristics, so that quote is graded `PARTIAL` — it presupposes the anomalies rather than reporting them — and is paired with the `PMID:17853487` malformation-pattern sentence that names urogenital systems directly.

**`treatments`: the GeneReviews Surveillance Schedule.** Each interval, not a general instruction to follow up. It deliberately carries **no `target_mechanisms`**: surveillance detects the arms of the disease, it does not act on them, and every available `treatment_effect` value asserts a causal action on its target. A `notes` line records that choice so the next curator does not read the absence as an omission. Bound to `NCIT:C15747` with `therapeutic_modality: OTHER`, which is the dominant house pattern for surveillance entries across `kb/disorders/`.

**`discussions`: `lath_mouse_lethality_mismatch`, `kind: HUMAN_MODEL_MISMATCH`.** The analysis was already written, in the Sc5d-null model's `limitations` field — where nothing surfaces it as an open translational question. Homozygous nulls are stillborn, so the model is informative about the malformation arm (and earns its place by attributing those defects to decreased cholesterol rather than accumulated lathosterol) and silent about the hepatic course and cataracts that dominate the human disease. It carries a proposed hypomorphic-allelic-series experiment with `would_support` / `would_refute`.

## Notes for review

**Two snippets come from PMC full text, not the abstract.** The variant sentence and the sterol profile are both in the body of `PMID:33204591`. `references_cache/PMID_33204591.md` is committed with `content_type: full_text_xml`, and I ran the `--no-full-text` path CI actually uses to confirm they still verify there — not just the permissive `just validate-references`.

**The sterol quote contains U+2010 hyphens and thin spaces.** It was lifted programmatically out of the cache rather than retyped, and is stored as a double-quoted scalar so the code points are unambiguous.

**`links.issues` is populated** on the history record — one of the #9787 items. The broader observation from the issue thread stands: 4,067 of 7,138 committed history records have `issues: []`, which is a scripted backfill and its own issue, not something five hand-edits move.

## Validation

```
just validate kb/disorders/Lathosterolosis.yaml            # schema + terms + references, passed
just validate-disorders kb/disorders/Lathosterolosis.yaml  # what CI runs, exit 0
  Snippets checked: 47/47 verified against cached references
just validate-history history/disorders/Lathosterolosis/2026-08-28T181322Z-claude-code-4f21c8.yaml
  No issues found
```

Part of #9787.

### DIFF
--- a/history/disorders/Lathosterolosis/2026-08-28T181322Z-claude-code-4f21c8.yaml
+++ b/history/disorders/Lathosterolosis/2026-08-28T181322Z-claude-code-4f21c8.yaml
@@ -40,6 +40,36 @@ events:
 
     discussions: added lath_mouse_lethality_mismatch, kind HUMAN_MODEL_MISMATCH. The analysis was already written in the Sc5d-null model's limitations field, where nothing surfaces it as an open translational question. Homozygous nulls are stillborn, so the model is informative for the malformation arm and silent on the hepatic and ophthalmologic course that dominates the human disease. Carries a proposed hypomorphic allelic-series experiment with would_support/would_refute.
 
+    Review follow-up (PR #9899). Three changes after the ai4c-reviewer round.
+
+    The blocking one: the proposed experiment put prose in would_support and
+    would_refute. Those slots take only <kind>#<name> entity references - the slot
+    descriptions say so explicitly, and supporting_outcome / refuting_outcome exist for
+    the prose. The prose is unchanged, just moved; would_support now names
+    pathophysiology#Progressive Liver Disease and
+    mechanistic_hypotheses#lath_accumulation_drives_storage, and would_refute names the
+    hypothesis alone, since the experiment is designed to discriminate on it. This was
+    also the CI failure: just check-entity-refs is an ungated whole-KB step and none of
+    the three commands recorded above catches it. Adding it to the pre-PR routine.
+
+    The surveillance snippet dropped the registered-trademark mark the cache carries on
+    FibroScan(R). It validated either way, which means the matcher normalizes it, but an
+    exact quote should be exact.
+
+    Added a FAILS_TO_RECAPITULATE link from the Sc5d-null mouse to Progressive Liver
+    Disease, on the reviewer's suggestion, so the mismatch is machine-visible on the
+    model rather than only readable in the new discussion. Carries the required
+    limitations and evidence: the abstract's statement that homozygous nulls were
+    stillborn is the substantive basis, and the limitation records the honest ambiguity
+    - a complete null is a more severe lesion than any patient carries, so the lethality
+    may be the mouse reproducing something true rather than failing.
+
+    Also trimmed a sentence of curation rationale out of that link's description
+    ("Recorded as an explicit negative so the silence is machine-visible..."). description
+    renders on the disorder page as a statement about the disease; reasoning about the
+    modeling belongs here. The same reviewer raised this pattern on three sibling PRs in
+    this round, so it is a habit rather than a slip.
+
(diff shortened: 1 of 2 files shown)


caufieldjh:
Both fixed, plus both non-blocking notes.

**`would_support` / `would_refute` → entity refs.** You were right that the three commands in the PR body don't catch it, and this was also the `test (3.13)` failure — `just check-entity-refs` is ungated and whole-KB, so it went red on `main`'s copy of the check, not on anything the PR body ran. The prose is unchanged, just relocated to `supporting_outcome` / `refuting_outcome`; `would_support` now names `pathophysiology#Progressive Liver Disease` and `mechanistic_hypotheses#lath_accumulation_drives_storage`, `would_refute` the hypothesis alone since the experiment is designed to discriminate on it. `check-entity-refs` is going into my pre-PR routine.

**`FibroScan®`.** Fixed. It validated either way, which means the matcher normalizes the mark — but an exact quote should be exact, and relying on a normalization I hadn't verified was luck rather than method.

**`FAILS_TO_RECAPITULATE` on the Sc5d-null model.** Added, and it improved the entry more than I expected. Carrying the required `limitations` and `evidence` forced me to write down the ambiguity the discussion had left implicit: a complete null is a more severe lesion than any reported patient's genotype, and no patient homozygous for a null has been described — so the perinatal lethality may be the mouse *reproducing* something true about the human disease rather than failing to model it. That reads better as a structured negative claim than as prose.

I also trimmed a sentence of curation rationale out of that new link's `description` ("Recorded as an explicit negative so the silence is machine-visible…"). You flagged the same `description`-vs-`notes` pattern on three of my sibling PRs in this round, so it's a habit rather than a slip and I've swept the others for it.

```
just check-entity-refs kb/disorders/Lathosterolosis.yaml   OK
just validate-disorders kb/disorders/Lathosterolosis.yaml  exit 0, 48/48 snippets
pytest -k "failure_to_recapitulate or animal_model_mechanism or model_readout"  7655 passed
```

PR #9759, opened 2026-08-27
curate(CYB561): Add CYB561-related orthostatic hypotension

caufieldjh:
Adds `kb/disorders/CYB561-Related_Orthostatic_Hypotension.yaml` (MONDO:0020751).

## An enzyme deficiency with a perfectly normal enzyme

Cytochrome b561 is a transmembrane haem protein that carries electrons across the catecholamine secretory vesicle membrane to regenerate **intravesicular ascorbate** — the cofactor dopamine beta-hydroxylase consumes converting dopamine to noradrenaline. Losing it does not damage the enzyme. It starves the enzyme of a regenerated cofactor *inside a compartment the cell cannot simply refill from outside*.

The result is a functional DBH deficiency with an intact DBH gene and a **normal plasma DBH activity**. That dissociation is the entry&#39;s spine, and it is curated twice — as a pathophysiology node and as a biochemical readout — because it is the diagnostic discriminator.

```
Biallelic CYB561 Loss of Function
  → Loss of Transmembrane Electron Transfer into Secretory Vesicles
      → Intravesicular Ascorbate Depletion
          → Functional Dopamine Beta-Hydroxylase Deficiency   ← droxidopa BYPASSES here
              → Noradrenaline and Adrenaline Deficiency
                  → Selective Sympathetic Noradrenergic Failure
                      → Orthostatic Hypotension
```

## Phenocopy, not subtype

CYB561 deficiency and DBH deficiency share low noradrenaline and adrenaline — *&#34;As in DBH deficiency, concentrations of norepinephrine and epinephrine in the patients were low&#34;* — and both respond to droxidopa. They are nonetheless different diseases at different points in one chain.

**Preserved dopamine is deliberately not claimed as a shared feature.** It is a CYB561 finding; in primary DBH deficiency plasma dopamine is characteristically *elevated*, accumulating upstream of the enzymatic block. An earlier version of this description and of the entry&#39;s discussion listed it among the similarities, which was unsourced and points the wrong way — it is one of the better discriminators between the two, not a similarity. Corrected here rather than quietly in the YAML.

The discriminator that does hold is concrete and clinically actionable: *&#34;Plasma dopamine β-hydroxylase activity, however, was normal, and the DBH gene had no mutations.&#34;* A normal plasma DBH activity in a patient with absent noradrenaline is not a reason to doubt the phenotype — it is the specific indication to sequence CYB561. Curated as an `INTERPRETATION` discussion rather than by merging the two entities.

## Selectivity as the informative finding

[...]

PR #9900, opened 2026-08-28
curate(CYB561): clear the #9787 follow-up items

caufieldjh:
Clears the three CYB561 items deferred from the #9759 review and tracked in #9787.

## What changed

**`discussions`: drop the sentence that narrated its own edit history.** The phenocopy rationale contained *"An earlier version of this discussion listed preserved dopamine among the similarities, which was unsourced and points the wrong way."* That renders on the disorder page, where a reader has no earlier version to compare against, and git history is this repo's change log. The substantive point it was hedging — that preserved dopamine is a CYB561 finding, and that in primary DBH deficiency plasma dopamine is characteristically *elevated* — is already stated directly in the two sentences around it, so nothing is lost.

**`phenotypes`: `Reduced Life Expectancy` `OCCASIONAL` → `FREQUENT`.** This is arithmetic rather than judgement. Three of the four molecularly confirmed patients are dead: D1 at 37 after a prolonged circulatory arrest sustained while off droxidopa, and the two American sisters at 59 and 48 of causes never established. That is 75% — inside `FREQUENT` (30–79%) and well outside `OCCASIONAL` (5–29%).

Two things about this were also wrong in a smaller way and are fixed:
- The description led with *"A sibling died at 16"*. That brother predates the molecular diagnosis and is not among the four confirmed patients. He is now named as explicitly **uncounted** rather than used as the lead example.
- The pre-existing evidence item is the authors' qualitative *"more malignant in terms of renal insufficiency and reduced life expectancy"*. That supports the phenotype but says nothing quantitative, so it cannot carry a frequency band on its own. Two items with the actual ages were added underneath it, and its `explanation` now says which job it is doing.

**`pathophysiology`: the PAM arm.** `Intravesicular Ascorbate Depletion` previously described a single downstream consumer of the missing cofactor. There are two. Peptidylglycine alpha-amidating monooxygenase draws on the same intravesicular ascorbate pool as dopamine beta-hydroxylase, and amidates α-MSH, oxytocin, vasopressin and gastrin — a modification those peptides need in order to be active. So the lesion should in principle disable neuropeptide amidation as well as noradrenaline synthesis, and **no neuropeptide phenotype has been reported in any of the four patients.**

That asymmetry is curated as open, not resolved. It also sharpens the entry's existing `cyb561_expression_versus_phenotype` knowledge gap, which asked why the phenotype is confined when CYB561 is expressed widely: the strongest form of that question is not about tissue expression at all, but about a second enzyme *in the same vesicle*. Amidated-peptide measurements in a CYB561 patient would discriminate between a genuinely spared pathway and one nobody has assayed.

**`treatments`: droxidopa dosing and tolerability.** 200 mg once to three times daily across all four patients, with nausea, headache and lumbar pain as the limiting factor. Tolerability carries more weight here than it would for a symptomatic drug, because droxidopa is the only treatment that addresses the lesion at all — which the series makes bluntly, since D1's fatal arrest occurred *"while she was off L-DOPS treatment"*.

**`biochemical`: plasma DHPG and CSF MHPG.**
- **DHPG** is the noradrenaline metabolite formed *inside* the neuron before release, so undetectable DHPG says the noradrenaline was never made, not that it was made and cleared. That is the distinction between a synthetic defect and a release or reuptake defect.
- **CSF MHPG** is the entry's only direct evidence about the brain, and it is a three-part result rather than one: central noradrenaline is depleted while CSF HVA and 5-HIAA are normal, so the dopamine and serotonin pathways are intact. That preserved dopamine metabolite is the central counterpart of the preserved *plasma* dopamine that separates this disease from DBH deficiency.

## Notes for review

**Neither new biomarker carries a `term:`, deliberately.** CHEBI has `CHEBI:166465` (3,4-Dihydroxyphenylglycol O-sulfate) and `CHEBI:89405`/`CHEBI:89644` (MHPG glucuronide and sulfate), but not the free metabolites. A conjugate is a different analyte, and per the `dismech-terms` rule no term beats a bad one. Each `context` field records that the search was done, so the absence does not read as an omission to the next curator.

**The quoted fragments contain Greek letters (Dβ H, α-MSH) and en-dashes.** They were lifted programmatically out of `references_cache/PMID_29343526.md` rather than retyped.

**`links.issues` is populated** on the history record — one of the #9787 items.

## Validation

```
just validate-disorders kb/disorders/CYB561-Related_Orthostatic_Hypotension.yaml   # what CI runs, exit 0
  Snippets checked: 32/32 verified against cached references
just validate-history history/disorders/CYB561-Related_Orthostatic_Hypotension/2026-08-28T181510Z-claude-code-bf4cb3.yaml
  No issues found
```

Part of #9787.


caufieldjh:
Both blocking items fixed, and three of the four non-blocking notes.

**Dosing/tolerability on the wrong claim.** Fixed, and the point about `target_mechanisms[].evidence` backing the pathograph edge is the part I'd missed — I was treating it as "evidence for this treatment, filed under a link" when it is specifically evidence *for the link*. Both moved to `Treatment.evidence`, which was empty; tolerability regraded `PARTIAL`, following the `ATTR_Amyloidosis` diflunisal precedent you pointed at.

**The patient count.** Fixed, and thank you for checking it against the rest of the page rather than just the block — `prevalence.notes` and the entry `notes` both record five confirmed patients and I contradicted them in the one sentence the frequency argument rests on. Reworded to "the four patients in the founding report" in four places, including one pre-existing sentence in the expression-versus-phenotype discussion that had the same imprecision. As you say the band is unaffected: 3/5 = 60%, still `FREQUENT`.

**"The only dosing guidance this disease has"** — dropped. Shibao 2020 is cited in this very entry and also reports dosing; the exclusivity claim was just wrong.

**PAM items → `OTHER`.** Taken. They quote the paper's background enzymology, not a patient observation, and `HUMAN_CLINICAL` overstated them.

**The NCIT lookup was worth the minute.** Plasma DHPG now binds `NCIT:C101017` (3,4-Dihydroxyphenylglycol Measurement) — and you're right that "…Measurement" terms are established practice here (`Creatine Kinase Measurement` in seven entries, plus Lactic Acid, Hemoglobin, Dystrophin). I'd only checked CHEBI and concluded "no term exists", which was a statement about one ontology dressed up as a statement about all of them.

**CSF MHPG stays unbound**, and now for a stated reason rather than an unchecked one: NCIT's only MHPG term is `NCIT:C220216`, a **urine** measurement, and this result is in cerebrospinal fluid. Binding it would name the wrong specimen. Both `context` fields now record which ontologies were searched, so the next curator doesn't repeat the search or assume it wasn't done.

```
just validate-disorders ...  exit 0, 32/32 snippets verified
```

PR #9909, opened 2026-08-28
curate(TRMA): clear the #9787 follow-up items

caufieldjh:
Clears the remaining TRMA items deferred from the #9757 review and tracked in #9787.

## Two checklist items were already done

Before changing anything I checked the two "mechanical" items against `origin/main`:

- **`supports: SUPPORT` → `PARTIAL`** on the domain-level GeneReviews evidence for `Seizures` and `Stroke` — **already `PARTIAL`** in both blocks, with the explanation already naming the reason ("the sentence names the domain rather than this specific manifestation").
- **`therapeutic_modality: BEHAVIORAL` → `OTHER`** on the pregnancy treatment — **already `OTHER`**.

Both were evidently applied before #9757 merged. No change needed; recorded here so the boxes can be ticked on the issue rather than re-investigated.

## What changed

**`pathophysiology`: the extra-triad phenotypes are now in the pathograph.** Five downstream edges from `Intracellular Thiamine Pyrophosphate Deficiency` — `Cardiac Rhythm Disturbance`, `Seizures`, `Stroke`, `Optic Atrophy`, `Retinal Dystrophy`. No new pathophysiology nodes were needed: phenotypes are first-class graph nodes (`graph.py:319`), so the edges resolve directly. All five use `INDIRECT_UNKNOWN_INTERMEDIATES`, matching the three triad edges that already leave that node — the cofactor requirement is established, the route from it to any particular organ failure is not, and the link type says so rather than papering over it.

The node's own description now states that it is the branch point for the whole disease and that it branches wider than the triad, so a reader meeting eight outgoing edges understands why.

**Also wired: `Persistent Macrocytosis`**, which was an orphan phenotype with no incoming edge. Not on the checklist — found while verifying the wiring above. The GeneReviews evidence already sitting on that phenotype says the anemia corrects on thiamine *while the red cells stay macrocytic*, so the edge from `Impaired Nucleotide Synthesis in Erythroid Precursors` is what stops the pathograph implying that treating the anaemia reverses the cellular defect.

Verified with `build_causal_graph`: 8 edges out of the TPP node, no dangling edges anywhere in the entry.

**`animal_models`: the missing third arm.** The Slc19a2-null mouse was linked to the beta-cell and hair-cell arms but not the haematopoietic one — despite *megaloblastosis* being in the paper's own title. Added as `RECAPITULATES` against `Impaired Nucleotide Synthesis in Erythroid Precursors`, with a bone-marrow-morphology readout.

The interesting part is in `limitations`: the mouse megaloblastosis is **trilineage** (erythroid, myeloid and megakaryocyte), which fits the proposed mechanism — a transketolase-dependent shortfall of ribose-5-phosphate should constrain any rapidly dividing precursor, not erythroid ones specifically. Whether human TRMA marrow shows the same breadth is not established by this source, so the human node stays erythroid-scoped and the limitation says why.

**`prevalence`: `CASES_IN_LITERATURE`, `ULTRA_RARE`, no rate.** Two evidence items, because the source says two separate things — 183 patients in as many families worldwide, and that prevalence and incidence remain unknown. Those are not in tension: the count is an ascertainment floor, not an estimate, and the same source notes NGS has been raising it. Deliberately no `rate_per_100000`.

**`histopathology`: a new section, and it earns its place.** From a newly fetched `PMID:39467528` — *"Thiamine-Responsive Megaloblastic Anemia Syndrome Mimicking Myelodysplastic Neoplasm"*. The TRMA marrow shows dysplastic erythropoiesis with megaloblastic change against normal serum folate and B12, which is the picture that gets a TRMA patient diagnosed with MDS; the reported patient carried that diagnosis from age 26. Three evidence items cover the finding, the generalised mimicry, and the reversal — hemoglobin *and marrow morphology* normalised 12 weeks into oral thiamine, which is what distinguishes a cofactor-supply state from a clonal lesion.

## Notes for review

**The NCIT binding is the closest reachable fit, not an exact one**, and the block's `context` says so. `NCIT:C206357` (Dyserythropoiesis) is the better label for the dysplastic component but is not reachable from any `HistopathologyFindingTerm` source node. `NCIT:C82929` (Megaloblastoid Changes Present) *is* reachable, but "megaloblastoid" is conventionally the hedged form of the word and the source says megaloblastic. So it binds `NCIT:C35825` (Megaloblastic Erythroid Hyperplasia) with the finding carried verbatim in `preferred_term`.

**Cache rows were seeded by the validator, not written.** `cache/ncit/terms.csv` and `cache/enums/histopathologyfindingterm_b0babf2f7c57.csv` each gained one row as a side effect of `just validate-terms`.

**Every new snippet was confirmed as an exact substring before being written**, and none crosses a bracketed citation marker — the prevalence sentence in `PMID:38037112` is followed immediately by `[15]`, so it is quoted as two separate items rather than one spanning quote.

## Validation

```
just validate-disorders kb/disorders/Thiamine-Responsive_Megaloblastic_Anemia_Syndrome.yaml   # what CI runs, exit 0
  Snippets checked: 48/48 verified against cached references
just validate-history history/.../2026-08-28T182711Z-claude-code-45be53.yaml
  No issues found
```

Part of #9787.


caufieldjh:
Both blocking items fixed, plus two of the suggestions.

**`diagnostic: true`.** You're right and the contradiction was inside the block itself — the description says the finding misdirects toward MDS and the source is titled *"…Mimicking Myelodysplastic Neoplasm"*. Set to `false`, with the reason stated where a reader meets it: a finding that disguises a disease cannot be pathognomonic for it.

**Ring sideroblasts.** Added, and this was the better of the two findings. `NCIT:C35995` is reachable via `NCIT:C36289` — I confirmed with `runoak ancestors` and then with `just validate-terms` rather than taking the neighbourhood argument on trust. Your quote is exact and stops before the `[15]`. And your point about it strengthening item 1 is right: with both findings present the section shows *why* the mimicry is convincing, and neither entry needs a diagnostic flag doing work it can't do.

**`Persistent Macrocytosis` → `HP:0005518`.** Taken. `HP:0001972` asserts the anemia is still there, which is precisely what the phenotype and the new edge deny. Pre-existing, but as you say this PR is what made it a pathograph endpoint, so it was mine to fix.

**Curation rationale in `description`.** Taken, four fields trimmed. You raised this on three of my PRs in this round, which makes it a habit rather than a slip — the rationale is genuinely valuable and it is genuinely in the wrong field, and I've swept the siblings for it.

**Two I'm leaving, with reasons rather than silence:**

- *Structural asymmetry (item 5).* Deliberate, and I'd rather it stay recorded as a decision than get smoothed. No tissue-level mechanism is established for the five new arms; inventing an intermediate node would assert more than the sources do, and `INDIRECT_UNKNOWN_INTERMEDIATES` is the honest shape.
- *`Stroke` with no stroke source (item 6).* A real gap, and I looked — none of the entry's cached references, including the newly added `PMID:39467528`, names a cerebrovascular event. Closing it needs a fetch of a source this PR has no reason to be choosing, so it belongs in its own pass rather than bolted on here.

**Item 7 is a fair catch on my PR body** — the `therapeutic_modality: OTHER` the diff adds is on *Audiological and Metabolic Surveillance*, a different treatment from the pregnancy one the body discusses (which needed no change). The history record has it right; the body was ambiguous.

```
just validate-disorders ...  exit 0, 49/49 snippets verified
```
Four cache rows validator-seeded (HP:0005518, NCIT:C35995, and their enum-membership rows).

PR #8682, opened 2026-08-17
curate: add Vein of Galen Aneurysm (MONDO:0015196)

caufieldjh:
## Vein of Galen Aneurysm (MONDO:0015196)

New entry, curated from an **Edison Falcon** deep-research report.

### The name is wrong twice, and the entry says so

It is **not an aneurysm** — there's no weakened arterial wall ballooning out, but a direct connection between arteries and a vein with no capillary bed between them. And in the true malformation the dilated vessel is **not the vein of Galen** but its embryonic precursor, the median prosencephalic vein of Markowski, which should have regressed and instead persisted as the receiving limb of a shunt formed in fetal weeks 6–11.

The MONDO label is kept as the `disease_term` (it's the ontology's identifier for the concept), while the description and nodes use the anatomically accurate account.

### Three consequences in parallel, not in series

```
Endothelial Ras signalling failure (fetal)
   → Persistent embryonic AV shunt ─┬→ High-output circulatory overload → CHF, PAH
                                    ├→ Cerebral venous hypertension     → hydrocephalus
                                    └→ Cerebral arterial steal          → delay, seizures
```

The parallel structure is deliberate. All three follow from a single physical fact — the low resistance of a capillary-free channel — rather than from one another. A large fraction of cardiac output passes through the shunt and returns to the right heart **having done no work**, so the newborn heart must move enormous volumes to perfuse anything. It's heart failure with a structurally normal heart, which is why the cardiac presentation is baffling until someone images the head.

### Genetics reframed: a cell type at a moment, not a vessel

The 2023 integrated analysis of **310 proband-family exomes** with **336,326 cerebrovascular single-cell transcriptomes** concluded that *"developing endothelial cells"* are the likely spatio-temporal locus of the pathophysiology. That's a genuinely different way to site a malformation — not in an anatomical structure but in a cell type during a developmental window.

`EPHB4` carries the strongest functional case and is supported by an `animal_models` **`RESCUES`** link: wild-type human EPHB4 rescues the zebrafish brain vascular defect, the truncated disease allele does not. Its `limitations` state plainly that the fish has no median prosencephalic vein to persist — **the model tests allele function, it does not reproduce the lesion.**

`RASA1` variants "were identified in all types of shunts," which is curated as support for a shared upstream mechanism rather than lesion-specific genotypes.

### The outcome caution, curated on both the phenotype and the treatment

[...]

PR #9912, opened 2026-08-28
curate(COQ6): clear the #9787 follow-up items

caufieldjh:
Clears the COQ6 items deferred from the #9758 review and tracked in #9787.

Two of the checklist items needed adapting rather than applying literally — details below, since both are about a slot that does not exist where the issue assumed it did.

## What changed

**`genetic`: loss-of-function on both nonsense variants — via `functional_effects`, not `functional_impact_category`.** The issue asked for `functional_impact_category: LOSS_OF_FUNCTION`. That slot lives on `GeneticContext`, which is not reachable from a `variants[]` entry — the `Variant` class carries `functional_effects` instead, and that is what the rest of `kb/disorders/` uses for a per-variant LOF claim (`Abetalipoproteinemia`, `3-Hydroxy-3-Methylglutaryl-CoA_Synthase_Deficiency`, `ADCA-DN`). Same goal, achieved in the slot that exists: the residual-activity argument in `notes` is now queryable rather than prose, supported by the snippet already quoted on both variants (*"resulting in two truncated protein products"*).

**`treatments`: the idebenone `REFUTE` moves to the arm it is about.** At treatment level, *"Idebenone treatment did not, however, improve the hearing loss"* scanned as a refutation of idebenone as such — which the `RESTORES` link on `Optic Nerve Involvement` directly above it contradicts. It now sits on its own `target_mechanisms` entry against `Progressive Sensorineural Hearing Loss`, where it says what it means: the claim is about the drug's *reach*, not its efficacy.

That link deliberately carries **no `treatment_effect`**. Every available value (`INHIBITS`, `ACTIVATES`, `MODULATES`, `BYPASSES`, `RESTORES`, `TRIGGERS`) asserts that the drug acts on the node, and the finding is that it did not. `treatment_effect` is optional on `TreatmentMechanismTarget`, so recording the arm as targeted-and-negative needs no workaround — and it is more informative than omitting the arm, because the per-arm dissociation within one drug in one patient is the whole reason idebenone is interesting here.

**`treatments`: cochlear implantation `NCIT:C15747` → `NCIT:C15329`**, with `preferred_term: cochlear device implantation`. This matches the twenty other cochlear-implantation treatments across `kb/disorders/` and the sibling entry in #9757. A `notes` line records *why* the obvious term is unusable — `NCIT:C157820` "Cochlear Implant" names the device rather than the clinical action, so it is not reachable from `NCIT:C25218` where `TreatmentTerm` is rooted. That reasoning had been independently re-derived in at least four separate entries; see the companion PR for the KB-wide sweep.

**`treatments`: `Dialysis and Kidney Transplantation`.** Curated for a mechanistic reason rather than completeness. Post-transplant recurrence is low in monogenic SRNS *because* the lesion is in the podocyte genome and the graft podocytes are someone else's — a prediction the entry can make from its own pathophysiology, and what makes transplantation definitive here in a way it is not for immune-mediated nephrotic syndrome. Linked `BYPASSES` on the renal node, with the note that the auditory arm continues after a successful graft.

**`diagnosis`: two entries, and the second is deliberately hedged.** Molecular testing is the diagnostic route — there is no biochemical screen specific to COQ6, so whole-exome or an SRNS/FSGS panel is the way in, and the clinical trigger is steroid resistance plus deafness in a child. Renal biopsy showing FSGS is curated separately and *explicitly as directional rather than diagnostic*: FSGS is the common final appearance of podocyte injury from many causes, so it identifies the lesion and not its cause.

**`progression`: one phase.** Records the authors' word — kidney failure is described as following *inevitably* — together with the modifier on it, that renal function can be restored if CoQ10 starts before the kidney pathology progresses. That is what makes timing rather than treatment choice the decisive variable in this disease, and it is exactly where the renal and auditory arms come apart: renal function can go back to normal, hearing responds in under half of patients and not at all once lost.

**`notes`: `OMIM:614650` and `ORPHA:280406` recorded as identifiers, not as a `mappings` block.** The schema's `mappings` has `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings` — there is no `omim_mappings` or `orphanet_mappings` slot. Both identifiers are already xrefs of the curated `disease_term` `MONDO:0013836` (verified with `runoak`, alongside DOID, GARD, MEDGEN and UMLS), and re-asserting an ontology's own xrefs is what the design register tells curators not to do. The free-text identifiers line is the house pattern (`Acatalasia`, `3-Phosphoglycerate_Dehydrogenase_Deficiency`).

## What I did not add, and why

**`prevalence`.** None of the eight cached references gives a COQ6-specific rate or literature case count. The one quantitative figure available — SRNS affects roughly 10–20% of paediatric nephrotic syndrome — is about SRNS as a whole, not this gene, and the 12-patient audiological series is a single-centre cohort rather than a literature total. Recording a rate would have meant manufacturing one.

**`biochemical`.** Same reason. The plasma and urine findings in the sources (proteinuria, hypoalbuminemia, and in the lethal infantile siblings a raised lactate) are the findings of nephrotic syndrome and of decompensation from any cause, not markers that report the COQ6 lesion. A `biochemical` block here would have looked like grounding without adding any.

**One evidence item was written and then removed before commit.** A methods sentence from `PMID:36124066` (*"We compared the two groups in terms of..."*) cited under the transplantation treatment. It validates as an exact quote and asserts nothing — the failure mode #8682 describes. Dropped rather than regraded.

## Validation

```
just validate-disorders kb/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome.yaml   # what CI runs, exit 0
  Snippets checked: 43/43 verified against cached references
just validate-history history/.../2026-08-28T183633Z-claude-code-265e81.yaml
  No issues found
build_causal_graph: 26 nodes, 19 edges, no dangling
```

Every snippet was confirmed as an exact substring before being written; the dialysis sentence is truncated before a thin space in "6 years". The first `progression` attempt used `name`/`description` and failed schema validation — `ProgressionInfo` takes `phase`/`notes`.

Part of #9787.

### DIFF
--- a/history/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome/2026-08-28T183633Z-claude-code-265e81.yaml
+++ b/history/disorders/COQ6-Related_Steroid-Resistant_Nephrotic_Syndrome/2026-08-28T183633Z-claude-code-265e81.yaml
@@ -58,4 +58,50 @@ events:
 
     One evidence item was written and then removed before commit: a methods sentence from PMID:36124066 (We compared the two groups in terms of...) cited under the transplantation treatment. It validates as an exact quote and asserts nothing, which is the failure mode dismech#8682 describes. Dropped rather than regraded.
 
+    Review follow-up (PR #9912). Both blocking findings and two suggestions.
+
+    Blocking: neither new diagnosis entry bound a diagnosis_term, which 1049 of the 1495
+    disorder files with a diagnosis section do populate. Both terms were already in
+    cache/ncit/terms.csv, so this needed no research: NCIT:C15709 Genetic Testing on the
+    molecular entry, NCIT:C51699 Kidney Biopsy on the biopsy entry. Galloway-Mowat is the
+    precedent - the closest sibling podocytopathy, which makes the same
+    histology-is-descriptive-not-diagnostic argument in its description and still binds
+    the term. A notes line on the biopsy entry records that the term names what is done,
+    not what it proves.
+
+    Blocking: Dialysis and Kidney Transplantation was one treatment bundling two
+    modalities, bound only to NCIT:C15289 Organ Transplantation with
+    therapeutic_modality SURGERY - so dialysis was invisible to anyone filtering by
+    term or modality despite being half the entry. Split into two, following the
+    Chronic_Kidney_Disease template: Dialysis with NCIT:C15221 and
+    therapeutic_modality DEVICE, Kidney Transplantation with NCIT:C15289 and SURGERY.
+    They are also different mechanistic claims, which is the stronger reason - dialysis
+    substitutes excretory function indefinitely, transplantation replaces the organ
+    definitively, and the low-post-transplant-recurrence argument that motivated
+    curating this at all is about transplantation specifically. It read oddly attached
+    to dialysis.
+
+    That recurrence argument is now flagged in a notes line as a prediction from this
+    entry's own pathophysiology rather than a claim sourced to the cited papers, which
+    report the transplantation but not a recurrence rate.
+
+    Suggestion taken: moved curation rationale out of three rendered description fields
+    into notes - the idebenone link's explanation of why it carries no treatment_effect,
+    the Tyr83Ter functional_effects explanation of why a structured effect was used, and
+    the biopsy entry's "the entry should not imply that it does". description renders on
+    the disorder page as a statement about the disease. Note TreatmentMechanismTarget has
+    no notes slot (only target, treatment_effect, description, evidence), so the
+    idebenone rationale went to the treatment's notes instead; the first attempt put it
+    on the link and failed schema validation.
+
+    Suggestion taken: cochlear implantation treatment_effect MODULATES to BYPASSES,
+    which is what its own description already said.
+
+    Suggestions noted and not acted on: serial pure-tone audiometry and tissue CoQ10 /
+    respiratory-chain assays as further diagnosis entries. The reviewer explicitly did
+    not ask for the second, since its source is a GeneReviews chapter whose cached record
+    holds only the six-point purpose statement. The audiometry one is a fair gap and is
+    quotable from an already-cached reference, but it is new content rather than a fix to
+    what this PR changed, so it belongs in its own pass.
+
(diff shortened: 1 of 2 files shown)


caufieldjh:
Both blocking findings fixed, plus both suggestions.

**`diagnosis_term` on both entries.** Fixed — `NCIT:C15709` Genetic Testing and `NCIT:C51699` Kidney Biopsy, both already in `cache/ncit/terms.csv` as you said. The `Galloway-Mowat_Syndrome` precedent is the argument that settled it: it makes the *same* histology-is-descriptive-not-diagnostic point in its description and still binds the term, because the term names what is *done*, not what it proves. I'd conflated those. A `notes` line on the biopsy entry records that distinction so the binding doesn't read as a claim the description contradicts.

**Splitting dialysis from transplantation.** Fixed, following the `Chronic_Kidney_Disease` template — `NCIT:C15221` + `DEVICE` and `NCIT:C15289` + `SURGERY`. Your second argument is the one that convinced me: they are different mechanistic claims, and the low-post-transplant-recurrence reasoning that motivated curating this at all is about transplantation specifically. It did read oddly attached to dialysis.

While splitting I also flagged that recurrence argument in `notes` as a prediction from this entry's own pathophysiology rather than something the cited papers report — they document the transplantation, not a recurrence rate. It was stated confidently enough in `description` to be mistaken for sourced.

**Curation rationale in `description`.** Taken, three fields. Worth noting for anyone hitting this later: `TreatmentMechanismTarget` has no `notes` slot (only `target`, `treatment_effect`, `description`, `evidence`), so the idebenone rationale went to the *treatment's* `notes` — my first attempt put it on the link and failed schema validation.

**`MODULATES` → `BYPASSES` on cochlear implantation.** Taken; it's what the description already said.

**Finding 5, both parts:** the tissue CoQ10 / respiratory-chain one I'm leaving for the reason you gave — the GeneReviews record cached here is only the six-point purpose statement, so there's no quotable snippet. Serial pure-tone audiometry is a fair gap and is quotable from an already-cached reference, but it's new content rather than a fix to what this PR changed, so it belongs in its own pass rather than growing this one.

Separately: the `NCIT:C157820` note here turned out to be one of **five** independent re-derivations across the KB. #9917 now sweeps the rest and puts the rule in `CLAUDE.md`.

```
just validate-disorders ...  exit 0, 44/44 snippets verified
```

PR #9913, opened 2026-08-28
curate(PMM2-CDG): clear the #9787 follow-up items

caufieldjh:
Clears the PMM2-CDG items deferred from the #9755 review and tracked in #9787. The grouping membership edits a shared file and is in its own PR.

## What changed

**`treatments`: the epalrestat mechanism now rests on the primary source.** The `MODULATES` edge onto `Reduced Phosphomannomutase 2 Activity` was carried by one sentence from a nutrition review, correctly graded `OTHER` because it was a summary of somebody else's work. `PMID:31636082` — Iyer et al., the Disease Models & Mechanisms multispecies repurposing screen — is that work, and it supplies three items in its place:

- the fibroblast result naming four genotypes, *including R141H/F119L*;
- the size of the effect, 30% to 400% over baseline, **"depending on genotype"**;
- the proposed route from aldose reductase inhibition to glucose-1,6-bisphosphate stabilizing PMM2 homodimerization — which is otherwise a surprising connection to make between a diabetic-neuropathy drug and a phosphomannomutase.

Graded `IN_VITRO`, since these are measurements in patient fibroblasts. The genotype dependence is the part that earns its place here: it is *direct evidence* that `MODULATES` rather than `RESTORES` is the right grading on that edge, which until now was an editorial judgement. And the third snippet keeps the authors' own **"may"** — that is the mechanism they propose, not one they demonstrate, and the `explanation` says so.

The review (`PMID:35562242`) is retained *below* the primary items rather than removed, still graded `OTHER`: it adds the glycoproteomic arm that the enzyme assays do not cover.

**`genetic`: R141H and F119L are structured variants now, not a sentence in `notes`.** Each carries `functional_effects` with `type: loss-of-function`, `clinical_significance: PATHOGENIC`, and its own evidence.

They are deliberately written as a pair, because the interesting claim is relational rather than per-allele. R141H is a functional null that has never been reported homozygous in a living patient — so *every* patient carries something else on the other chromosome, and it is that second allele's residual activity, not R141H, that sets where they fall on the severity spectrum. F119L is the commonest such partner *in Scandinavia*, where the pairing is reported, which is why the commonest genotype in the disease is a compound heterozygote rather than a homozygote. And R141H/F119L fibroblasts were one of the four lines in which epalrestat raised PMM2 activity — so "the commonest genotype" and "the genotype an enzyme activator has something to act on" turn out to be the same fact seen twice.

One honesty note in the YAML: R141H's `functional_effects` description records that complete loss in the homozygous state is **inferred from the allele's absence among living patients**, not read off a direct assay. No source in this entry measures R141H-homozygous activity, for the obvious reason.

**`prevalence`: split by population instead of one worldwide number.** `PMID:34447415` — a reference the entry already cited for the worldwide allele-frequency ceiling — reports three population estimates, now curated as three `BIRTH_PREVALENCE` records: Ashkenazi Jewish 1 in 20,000 (5.0 per 100,000), European 1 in 27,000 (3.7), admixed American 1 in 64,000 (1.6). That three-fold spread is exactly what a single worldwide number hides. The existing North America / Europe point-prevalence record is kept alongside them, and all three carry the same allele-frequency caveat: genotypes that never reach live birth are counted here and not in a clinical series.

**Dropped a sentence that narrated the entry's own edit history.** The mannose treatment's `notes` opened *"An earlier version of this description also claimed correction of hypoglycosylation in patient fibroblasts..."*. That is the same defect the CYB561 item in #9787 identified — it renders on the disorder page, where a reader has no earlier version to compare against.

Worth flagging: I grepped the round's five entries for the pattern and **three of the five had it** (CYB561, PMM2, COQ6). CYB561 is fixed in #9900 and COQ6 in #9912. The substantive caveat here is kept — correction of hypoglycosylation is not claimed because nothing cited supports it — restated as a present-tense fact about the entry rather than as a diff against a version no reader can see.

## Corrected during review

- **An invented reference title.** `PMID:40307862`'s `reference_title` was written from memory as "Phosphomannomutase 2 Congenital Disorder of Glycosylation: A Review." The cache says otherwise, and the same entry already had it right in two other places. Copied off the cache, which is where it should have come from.
- **A history record that described a change not in the diff.** It said ClinVar variation IDs 7706 and 7711 were added; a later commit removed them deliberately (the source gives bare variation IDs, the KB's form is the zero-padded VCV accession, and converting is not something the quoted sentence supports). The record now says they were added and then removed, and why — amended before merge, because history records are append-only afterwards.
- **An overstated `explanation`.** The epalrestat item claimed "direct evidence that this allele leaves activity to raise"; residual activity in a compound heterozygote is a property of the pair, not an assay of F119L. Now "consistent with".

## Validation

```
just validate-disorders kb/disorders/PMM2-Congenital_Disorder_of_Glycosylation.yaml   # what CI runs, exit 0
  Snippets checked: 72/72 verified against cached references
just check-duplicate-keys / -entity-refs / -folded-hyphens / -reference-titles / -snippet-grading   # clean
just validate-history <both records>   # No issues found
```

All new snippets were confirmed as exact substrings of their caches before being written. `references_cache/PMID_31636082.md` was created by `just fetch-reference` and is committed.

Part of #9787.


caufieldjh:
Both blocking findings fixed, plus all three suggestions.

**The duplicated prevalence record.** You're right, and the diagnosis is better than the symptom: they are the same number *because they are the same estimate*. The paper computed per-gnomAD-population values only, so the abstract's "could be as high as 1:20,000" is the maximum across populations — which is the Ashkenazi Jewish figure. I added a record for a number the entry already had, under a different label, and anything reading `rate_per_100000` would have double-counted.

Merged into one Ashkenazi Jewish record carrying both quotes, and took your suggestion further: European (1 in 27,000 → 3.7) and admixed American (1 in 64,000 → 1.6) are now their own records too. The three-fold spread my PR body claimed as the reason for the addition is now actually *in* the structured data instead of only asserted in prose — which it wasn't before, and that was the real weakness.

**R141H's claims had no citation.** Fixed with the embryonic-lethality sentence from `PMID:34447415`, graded `COMPUTATIONAL` to match how that paper is graded elsewhere in the file. Your framing was exactly right — "the gap is the missing citation, not the inference" — and the quote does more than close it: it also explains *why* an allele-frequency prevalence estimate overcounts here, since the study deliberately kept homozygotes in the calculation while acknowledging they are not born. That ties the two blocking findings together.

**All three suggestions taken:**
- "may shunt glucose" regraded `IN_VITRO` → `OTHER`. It sat as a third `IN_VITRO` item beside two real assays, which misrepresented a proposal as a measurement. The explanation now says so rather than only preserving the "may".
- Dropped the duplicate R141H frequency item at `genetic[0]` level. `genetic[0]` now carries no top-level evidence, which is correct — every claim it makes is either summarised from the variants or covered by their evidence.
- **The journal.** `Dis Model Mech`, not PLoS Biology. Corrected in the history record (in place, with the correction noted rather than silently overwritten) and in the PR body. I inferred the journal instead of reading it off `references_cache/PMID_31636082.md`, which is the same failure mode as writing a `reference_title` from memory — and the reason the rule is that these fields get copied, never typed. Good catch while it was still amendable.

```
just validate-disorders ...  exit 0, 71/71 snippets verified
```

caufieldjh:
Thank you for going back to the cache — you were right to distrust your own previous reasoning, and I verified the retraction in both directions before acting on it rather than just swapping one reviewer conclusion for another.

**The attribution.** Confirmed. `references_cache/PMID_34447415.md` has the introduction sentence ending `...could be as high as 1:20,000 (Schollen et al., 2000)`, and the study's *own* result sentence gives Ashkenazi Jewish 1 in 20,000 as a separate gnomAD-derived figure. The two coincide numerically and are not the same estimate.

The part worth naming: **my snippet stopped one word before the attribution.** It verified clean, the reference was real, the title matched — and the truncation point silently converted somebody else's estimate into this paper's. That's a failure mode with no gate on it, and it's more insidious than a paraphrase, because everything that checks the quote passes.

Fixed as you suggested: the mis-attributed evidence item is dropped (the three-population sentence supports the record on its own), and the `notes` no longer claim the two are one estimate. Schollen is mentioned as an independent earlier figure that lands on the same number, with an explicit statement of why it is *not* cited as support — a coincident number from another study isn't corroboration.

**The structural merge stands**, as you said it should. Two records asserting `rate_per_100000: 5.0` from one source was real double-counting regardless of how I'd justified the merge.

**The edit-history narration** — caught and removed, and it stings appropriately: I introduced it *while documenting an error in a rendered field*, in the same PR that removed the mannose one. Fourth instance across this round. I grepped the file afterwards; none remain.

**`European (non-Finnish)` → `European`.** Taken. The non-Finnish reading is a sound inference from the methods but it *is* an inference, and a reader scanning the `population` field would take it as sourced. Moved to `notes` with the reasoning stated.

**The PR body** — already corrected before your review ran; you were reading a stale copy. No action needed there.

```
just validate-disorders ...  exit 0, 70/70 snippets verified
```

caufieldjh:
Blocking item fixed, three suggestions taken, and the branch rebased onto current `main` for a reason that was not in either review.

**The F119L record now shows its work.** You were right that both its evidence items came from the epalrestat screen, which says only that one of four tested fibroblast lines was `R141H/F119L` — neither sentence measures this allele or mentions how common it is. Both sentences you pointed at are now cited, verified as exact substrings:

- `PMID:40501776` — *"the two most prevalent human pathogenic variants R141H (…ClinVar 7706) and F119L (…ClinVar 7711)"*, for the prevalence claim
- `PMID:40307862` — *"This missense mutation is more frequent in compound heterozygosity with p.Arg141His and is predominantly prevalent in Scandinavia"*, for the pairing

Both graded `OTHER`, including the second one, which this entry grades `HUMAN_CLINICAL` elsewhere. That sentence recites frequency and geography from three cited sources rather than reporting the review's own patient data, so it falls under the same recitation convention that settled the 1:20,000 attribution. `check-snippet-grading` is keyed on the sentence rather than the PMID, so the two gradings coexist correctly.

**The geographic qualifier was the more important half of that finding.** "The commonest genotype in the disease", stated flat, generalized a Northern-European observation worldwide. The description now says the pairing is reported from Scandinavia, where F119L is most prevalent, rather than asserting a worldwide ranking. I did not go looking for a source that makes the unqualified claim — you had already checked, and the 89-patient immune-selected cohort is not it.

Suggestions taken: ClinVar 7706 and 7711 plus `clinical_significance: PATHOGENIC` on both variants, both readable straight out of the newly quoted sentence; the Ashkenazi `notes` now say the study *reports* three population estimates and computed seven. I left the R141H `COMPUTATIONAL` grading alone, as both rounds concluded.

On the history record's "69/69": rather than a correcting line in an append-only document, this round adds its own record, which states 72/72 — the current figure after the two new items.

## The rebase

This branch was based on `a74314f2e`, which predates the #10003 `EvidenceItemSupportEnum` narrowing, so its schema still permits `PARTIAL`. `main`'s does not — and `main` has **already cleared** the retired values from this exact file, while the branch still carried **6**. Merging as-is would have reintroduced them into a file that had been cleaned. CI here was green only because it ran against the older schema.

Rebase was clean, four commits, no conflicts. The diff against `main` is now three files plus the new history record, and the retired-value count in the entry is zero.

## Validation after the rebase

- `just validate` — schema clean, terms clean, **72/72 snippets** verified
- `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` — all clean, no new baseline entries
- `validate-history-all` — no issues

The pre-existing `check-snippet-grading` failure on `Metastatic_Ovarian_Cancer.yaml` that both rounds noted no longer reproduces from this branch — it appears to have been resolved on `main` since.

---

caufieldjh:
Both fixed, and you were right that the history record was the one that had to be caught before merge rather than after.

**The stale ClinVar sentence.** Corrected rather than backfilled — I took the first route you suggested. The record now says the identifiers were added in that session and then removed before push, with the reason: the source gives bare variation IDs and the KB's form is the zero-padded VCV accession. You're right that the padding is derivable from `Cardiofacioneurodevelopmental_Syndrome.yaml`'s two examples, and that weakens my original "unverifiable guess" framing — but the identifiers are genuinely optional and nothing in the cited sentence establishes that variation ID 7706 and accession `VCV000007706` are the same record, so I'd rather the entry not assert it. Recording the round trip is more useful to a future auditor than either silence or an inferred accession.

**The "Biallelic hypomorphic variants" opening.** Reworded to lead with the requirement instead of the description — "At least one hypomorphic allele is required: biallelic complete loss of activity is not compatible with survival, which is why missense alleles predominate and why a functional null such as R141H is never seen homozygous." That says the same thing as the old two sentences, resolves against the R141H variant record instead of appearing to contradict it, and folds in the never-homozygous fact that the record argues from.

**PR body** updated: the prevalence section now describes the population split rather than a single added record, with all three rates, and there's a short section recording the three things corrected during review (the invented reference title, the stale history sentence, the overstated `explanation`).

Your third suggestion — that `ClinVar:VCV…` vs `ClinVar:SCV…` vs no `ClinVar:` reference prefix wants pinning down in its own issue rather than decided at review time — I agree with and am not doing here.

## Validation

`just validate` exit 0, **72/72 snippets** verified, no cache churn. `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-reference-titles`, `check-snippet-grading` all clean with no new baseline entries. Both history records validate.

---

PR #9914, opened 2026-08-28
curate: add SEMD Guo-Campeau type (MONDO:0958006)

### LINKED ISSUE #9884: Curate spondyloepimetaphyseal dysplasia, Guo-Campeau type (MONDO:0958006)
Curate a dismech entry for **spondyloepimetaphyseal dysplasia, Guo-Campeau type** ([MONDO:0958006](https://monarchinitiative.org/MONDO:0958006)) — a skeletal dysplasia caused by biallelic variants in `ERI1`, a 3'→5' exoribonuclease that trims histone mRNA 3' ends and the 5.8S rRNA.

Preflight: MONDO:0958006 appears nowhere in `kb/` on `main`, and `ERI1` appears in no existing entry's `genetic:` block. No open PR matches.

The interesting curation problem here is the same one that makes it worth doing: a housekeeping RNA-processing enzyme producing a *tissue-restricted* skeletal phenotype. The entry should say what is actually established about why chondrocytes are selectively vulnerable, and record the rest as a knowledge gap rather than narrating a plausible chain.

Stub: `stubs/Spondyloepimetaphyseal_Dysplasia_Guo-Campeau_Type.yaml`

nlharris:
Adds a Disease entry for spondyloepimetaphyseal dysplasia, Guo-Campeau type — an autosomal recessive skeletal dysplasia from biallelic `ERI1` variants in which at least one allele is missense.

Closes #9884

## The scope call is the thing to review

`MONDO:0958006` is the **missense-driven** skeletal disease. Biallelic *null* `ERI1` genotypes produce a different, milder entity — mild intellectual disability with digital anomalies, its own OMIM record — and it is deliberately **not** curated here.

That's not a tidiness decision. Collapsing the two into one variably expressive disease would erase the finding that defines both:

> A severe spondyloepimetaphyseal dysplasia (SEMD) was identified in five affected individuals with missense variants **but not in those with bi-allelic null variants**, who showed mild intellectual disability and digital anomalies

Within one recessive gene, the missense alleles are *more* severe than the nulls. The molecular lesion that leaves a protein behind is the damaging one — the signature of a dominant-negative or neomorphic effect. The entry states that as the implication the authors draw, not as a demonstrated mechanism, because nothing cited distinguishes dominant-negative from neomorphic.

## Pathophysiology

`Loss of exoribonuclease activity` → `failed 5.8S rRNA trimming and histone mRNA turnover` → `impaired chondrogenesis` → `the skeletal syndrome`.

The chondrogenesis node is what makes this a mechanism rather than a gene-to-phenotype assertion: patient-derived iPSCs fail to undergo chondrogenesis with downregulation of skeletal-patterning genes. That's where a general RNA-metabolic defect becomes a cartilage-specific one, and it's human cellular evidence — which matters here because the mouse doesn't reproduce the disease at all.

## The mouse fails, and the failure is the point

The `Eri1` knockout is entered as `FAILS_TO_RECAPITULATE` / `LOW` fidelity. Its `limitations` field carries the argument: a knockout removes the protein, while the human disease *requires* a missense allele keeping a catalytically dead protein in place. Under a dominant-negative or neomorphic mechanism the knockout cannot reproduce it in principle — so the mouse's silence is consistent with that model rather than evidence against the gene.

The consequence worth recording: **no existing animal model can test a therapy for this disease.**

## A near-miss worth flagging

The deep-research report scores 6/6 references resolved, 0 confabulation, 14/14 quotes valid — but only **2 of 6 on topic**. That ratio turned out to be load-bearing.

One of the other four, `PMID:41549465`, is a case series of **Dyggve-Melchior-Clausen syndrome** — a *different* recessive spondyloepimetaphyseal dysplasia. Quoting it here would have been Named Entity Confusion that passed snippet validation cleanly, because the quote would be exact and the paper real. Only the two on-topic primaries are cited, and the near-miss is recorded in `notes` so the next reader of that report is warned.

## Gene ID

Verified against HGNC before use: `ERI1` is `hgnc:23994`, and the report was correct. I checked because the SPG81 report in this same batch (#9894) gave `HGNC:30396` for SELENOI, which is SELENOS. Verification is now reflexive rather than triggered by suspicion.

## Honest gradings

Four phenotypes carry `supports: PARTIAL` because the cached abstract names the *diagnosis* rather than each radiographic sign — platyspondyly is constitutive of a spondyloepimetaphyseal dysplasia, but the abstract doesn't say the word. Each `explanation` says so rather than inflating to `SUPPORT`. Brachydactyly's PARTIAL is different and more important: the quote naming digital anomaly explicitly belongs to the *null-allele* phenotype, so it establishes digital involvement in the ERI1 spectrum without asserting it in this entry's genotype class.

**No frequencies.** Five skeletal-phenotype patients give no denominator.

## Two knowledge gaps

The missense-vs-null mechanism, with its therapeutic consequence spelled out (degrading the mutant protein helps under a dominant-negative model; restoring activity is the target only under a loss-of-function model the genetics argues against). And the tissue-restriction question: why does losing a housekeeping RNA activity required in every dividing cell produce a phenotype confined largely to the skeleton?

## Validation

`just validate` (schema + terms), `just count-verified-snippets` (15/15), `just check-entity-refs`, `just check-duplicate-keys` — all pass. Reachability audit: one root, no dangling targets, no unreached phenotype.

---


nlharris:
All four items in `ce6115786`.

## Items 1 and 2 — the split, and why your argument beat the general rule

I'd have accepted "don't bundle two molecular events" as a style point. Your actual argument is better and I want to record it: **the entry's own tissue-restriction knowledge gap proposes chondrocyte histone supply as a candidate answer, and a hypothesis needs a node to attach to.** A bundled node makes the entry unable to state its own open question precisely. Now two siblings, histone arm grounded with `GO:0071044` (validated as *histone mRNA catabolic process*).

Item 2 was the more embarrassing half. The node describing the *defect* was evidenced by a quote describing ERI1's *normal function* — and the sentence that actually evidences it was sitting one node upstream in the same file. It legitimately serves both new nodes, since one lost activity produces both consequences, so it's attached to each with an explanation saying so.

## Item 3

Epiphyseal (`HP:0002656`) and metaphyseal (`HP:0100255`) dysplasia added. You're right that my own platyspondyly reasoning applied identically and I'd only carried it across for one of the three — the disease is named for all three and only one was queryable. Same PARTIAL grade, same explanation pattern.

## Item 4 — and thank you for the constraint

The survival range is in `notes`, with no evidence item. Deaths at 4 months, 5 months and 2 years against survival into adulthood in the founding family is the most consequential thing to tell a family at diagnosis, and the cached abstract doesn't state it.

Your "please do not manufacture an evidence item for it" is the right instruction and worth having in writing. The temptation with a fact that important is to find *some* quote that gestures at it, and that is precisely how a snippet ends up attesting a proposition it doesn't contain.

## On HP:0003417 — this is the third instance of one pattern

You caught that the report suggests `HP:0003417` for "irregular vertebral endplates", and that `HP:0003417` is **Coronal cleft vertebrae**. I've confirmed it against the cache and recorded it in `notes` so it isn't re-suggested.

The part worth drawing out: **the report's term-validation pass scored it verified — and was right to.** The term exists. Validation confirmed the identifier while the label attached to it was wrong.

That is the same failure as the gene-ID errors in the sibling PRs from this batch:

| Report | Claimed | Actually | Validation said |
|---|---|---|---|
| SPG81 (#9894) | `HGNC:30396` = SELENOI | SELENOS | verified |
| Siddiqi (#9928) | `HGNC:14929` = FITM2 | SIRT1 | verified |
| this PR | `HP:0003417` = irregular vertebral endplates | Coronal cleft vertebrae | verified |

Three reports, three wrong identifier→label mappings, all passing existence checks. Nothing in the pipeline compares the *name* against the *identifier*, which is the only check that would catch any of them.

## Not done, and why

The `diagnosis` section, the HeLa rescue experiment, the DEDDh/SAP domain observation, naming Hoxha-Aliu syndrome explicitly, and the RHDA3 curation artifact — all reasonable, none blocking, and all needing either a fetch or full text this PR doesn't carry. I'd rather add them with sources than approximate them now. Happy to do any of them if you'd prefer they land here.

## Validation on `ce6115786`

```
just validate                 → schema + terms clean (GO:0071044 verified)
just count-verified-snippets  → 19/19 (was 15/15)
```

Five nodes, one root, no dangling targets, every phenotype reached.

---

nlharris:
/review

---

nlharris:
Rebased onto current `main` and re-pushed as `7df2da08d`. No content change from the round-1 fixes — this is a CI/plumbing push.

**Why:** the round-1 head (`ce6115786`, pushed ~20:40) never got a `test (3.13)` run or a `claude-review` run at all. Its only check was `enable-auto-merge`. My `/review` comment above didn't produce one either, so a fresh `synchronize` event was the remaining lever. The rebase supplies one and clears 52 commits of staleness at the same time.

One conflict, in `cache/hgnc/terms.csv`: `main` had added `hgnc:23987` (NDUFA12) at the same sorted position where this branch adds `hgnc:23994` (ERI1). Resolved by keeping both in canonical order; `just normalize-cache` then reports zero files updated, so the ordering is right.

Post-rebase diff reviewed — 14 files, all this PR's own, no unrelated deletions and no protected-path churn.

**Validation on the rebased head:** `just validate` (schema + terms + references, 19/19 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-snippet-grading`, `check-reference-titles` — all pass. `check-snippet-grading` in particular is green now; the `Metastatic_Ovarian_Cancer.yaml` failure the other PRs in this batch were seeing was branch staleness.

The round-1 fixes themselves are unchanged and described in my earlier comment: the bundled node split into `Defective 5.8S rRNA 3' Trimming` and `Impaired Histone mRNA Degradation` (`GO:0071044`), the correct evidence attached to the histone arm, epiphyseal and metaphyseal dysplasia phenotypes added, and the infantile lethality range recorded in `notes` rather than fitted with an evidence item the abstract cannot support.

---

PR #9915, opened 2026-08-28
grouping(CDG): add PMM2-CDG as a member

caufieldjh:
Adds `PMM2-Congenital Disorder of Glycosylation` to the CDG grouping. Deferred from the #9755 review and tracked in #9787.

This was held back deliberately rather than folded into #9755: it edits a shared file, and the member string could not be written correctly until the PMM2 entry existed on `main`.

## The casing trap

`members[].member` matches the disease entry's top-level `name` field **exactly**. Every existing member uses the lowercase form — `ALG9-congenital disorder of glycosylation` — while the PMM2 entry's `name` is title-cased: `PMM2-Congenital Disorder of Glycosylation`.

The string was copied out of the merged file rather than reconstructed by analogy with its thirteen siblings, which would have produced a foreign key that does not resolve. `test_data.py`'s foreign-key check would have caught it, but only after it was written.

## Placement

Inserted **first**, not appended. PMM2-CDG is the commonest CDG by a wide margin and the reference case the other members are described against; listing it after thirteen rarer members would misrepresent the group. `grouping_rationale`'s precursor-pool clause is updated from `(MPI, PGM1)` to `(PMM2, MPI, PGM1)` to match — PMM2 is a precursor-supply defect, not an assembly defect.

## `differentiating_mechanisms`

Written to say what distinguishes PMM2 from its siblings rather than to restate what PMM2-CDG is, since the latter is what the disease entry is for. Two points:

- **It is upstream of every mannosyltransferase defect in this grouping.** ALG1, ALG3, ALG9 and ALG12 each lose one step of lipid-linked oligosaccharide assembly; PMM2 loss starves *all* of them of GDP-mannose donor.
- **MPI is its instructive counterpart** — the adjacent step in the same supply line, but treatable with oral mannose where PMM2-CDG is not. Two neighbouring enzymes, opposite therapeutic situations.

A second paragraph records the group's genetic architecture, which is clearest in PMM2: complete loss of activity is not compatible with survival, so patients are hypomorph compound heterozygotes and per-allele residual activity sets severity.

## Criteria satisfaction was checked, not assumed

The grouping's single `NECESSARY` block is an `OR` over `GO:0006487` (protein N-linked glycosylation) and `GO:0006491` (N-glycan processing). The PMM2 entry carries `GO:0006487` directly, so the combined expression is `SATISFIED`. The per-leaf `NOT_SATISFIED` on the Golgi-processing branch is expected and shared by most existing members — the criterion is a disjunction. The descriptor is also declared explicitly on the member's `differentiating_mechanisms`, so the satisfaction is visible in the grouping file itself rather than only derivable from the disease entry.

## Validation

```
just validate-grouping kb/groupings/Congenital_Disorders_of_Glycosylation.yaml
  All validations passed!
just check-groupings kb/groupings/Congenital_Disorders_of_Glycosylation.yaml
  structure: OK
  [criteria 0 NECESSARY] PMM2-Congenital Disorder of Glycosylation: SATISFIED
  ... all 15 members SATISFIED
just validate-history history/other/Congenital_Disorders_of_Glycosylation/2026-08-28T184743Z-claude-code-f91c66.yaml
  No issues found
```

No evidence items were added, so there are no snippets to verify.

Part of #9787. Companion to #9913 (the PMM2 entry itself).


caufieldjh:
Fixed, plus both non-blocking notes.

**The member count.** You're right that this is worse than a typo. That note is the file's own justification for `skos:exactMatch`, so a count disagreeing with `members:` undermines exactly the thing it exists to support — and it's the first thing anyone auditing the mapping would check. Now "14 of the 15", and I confirmed with `runoak ancestors` that `MONDO:0008907` really is inside the `MONDO:0015286` closure, so the VPS51 exception is unchanged and only the totals moved, as you said.

Worth flagging as a standing hazard rather than a one-off: a hand-maintained count living in the same file as the list it counts will go stale every time membership changes. If this grouping keeps growing it's a candidate for a generated check.

**`biological_processes` on the new member** — dropped. You're right that none of the other fourteen carries one and the evaluator reads the disease entry either way, so it was inconsistency without benefit. I'd added it to make the criteria satisfaction visible in the grouping file, but that's not worth being the only member shaped differently.

**Em dash** — fixed.

**`research/grouping_mondo_gaps.md`** — agreed, generated artifact, regenerated by its own script.

```
just validate-grouping ...   All validations passed
just check-groupings ...     structure: OK, 15/15 SATISFIED
just validate-history ...    No issues found
```

PR #9916, opened 2026-08-28
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 7 (MONDO:0011074, LMX1A)

### LINKED ISSUE #9903: Curate autosomal dominant nonsyndromic hearing loss 7 (MONDO:0011074)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 7 (DFNA7)** ([MONDO:0011074](https://monarchinitiative.org/MONDO:0011074)) — heterozygous *LMX1A* variants; LMX1A is a LIM-homeodomain transcription factor required for inner ear morphogenesis and hair cell development.

The stub gives one causal gene (`hgnc:6653` LMX1A) and no descendants. Note LMX1A also appears in `kb/disorders/Mitchell-Riley_Syndrome.yaml`, but only as an RFX6 target gene — that is not disease coverage of DFNA7.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_7.yaml`

kevinschaper:
Adds `kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_7.yaml` for DFNA7 (MONDO:0011074, LMX1A `hgnc:6653`) and retires its stub.

`entry_type: DISEASE`. One causal gene, no MONDO descendants, and one pathograph that holds together end to end: LMX1A haploinsufficiency, reduced transactivation of downstream targets, and a failure to maintain cochleovestibular function. It is not a subtype of anything dismech already has. LMX1A does appear in `kb/disorders/Mitchell-Riley_Syndrome.yaml`, but only as an RFX6 target gene, which is not disease coverage of DFNA7.

## The point the entry is built around

The human dominant disease is not a milder version of the mouse phenotype, and treating it as one would invert the mechanism.

Homozygous Lmx1a mice have grossly malformed inner ears — no endolymphatic duct, no semicircular canals, no endocochlear potential — and heterozygous mice hear normally. Human DFNA7 heterozygotes have anatomically normal cochleae on CT and then lose hearing over years or decades. The Dutch group that established the disease drew the inference explicitly: one wild-type copy of LMX1A appears sufficient to build the ear but insufficient to keep it working.

So both mouse models are curated with `FAILS_TO_RECAPITULATE` links against the human maintenance node, alongside `PARTIALLY_RECAPITULATES` against the transcriptional node they genuinely do model. A `HUMAN_MODEL_MISMATCH` discussion carries the unresolved question, with a proposed aged-heterozygote experiment, because nobody has tested mouse Lmx1a heterozygotes at an age matched to a human phenotype that can take decades to declare itself.

Haploinsufficiency rather than a dominant-negative effect is not assumed here — three independent groups tested it, one with plasmid titration and transcript-chase assays.

**No cochlear implantation or speech rehabilitation under `treatments`.** Both are standard in severe nonsyndromic hearing loss, but no DFNA7-specific report documents either, so curating them would be importing general practice as if it were evidence about this disease. Recorded as a scoping decision in the treatment `notes`.

**No narrower HPO term for the asymmetry.** HPO has no term for asymmetric hearing impairment, so `phenotype_term` binds the accurate parent HP:0000407 and the asymmetry is carried by `preferred_term`, with a `notes` line saying why. Same for hearing-aid fitting, where no NCIT clinical-action term is reachable from NCIT:C25218.

A second `KNOWLEDGE_GAP` records that nothing in the curated mechanism explains why a germline variant present equally in both ears produces different thresholds on the two sides. That is the most distinctive clinical feature of DFNA7 and the pathograph does not account for it.

## Deep research

A falcon report was run and preflighted against MONDO:0011074. Verdict was **WARN, not FAIL**, and I checked what triggered it: the gene profile is LMX1A=50 with the next gene at 3, and the OMIM mismatch comes from the report quoting the LMX1A *gene* MIM (600298) and a broad phenotype concept (500008) while explicitly stating it could not verify a disease-specific OMIM. That is not Named Entity Confusion. The entry was curated from primary sources regardless, and every snippet re-verified against the cache rather than lifted from the report.

## Validation

- `just validate` — 0
- `just validate-terms` — 0
- `just validate-disorders` — 0, 53/53 snippets verified
- `just validate-history` — 0

Cache diff is three rows, all introduced by this entry (MONDO:0011074, hgnc:6653, and the diseaseterm enum row).

Closes #9903


kevinschaper:
All four findings addressed, plus the new one. Two of them turned out better than the review expected.

**The third instance of the retracted claim.** You were right and this is the failure mode I should have anticipated: I grepped for the wording I had written, not for the idea. `genetic[1].notes` is rewritten, and that entry now says explicitly that the recessive allele is curated inside this entry *because* MONDO has no separate term — so the three places now agree and reinforce each other rather than merely not contradicting.

**Finding 1 — hearing aids.** Replaced the boilerplate sentence with the one you pointed at, and added a second aided individual from the same family. The old `explanation` asserted that no report documents amplification in a specific patient; that was false and is gone.

**Finding 2 — cochlear implantation. This is now a curated treatment, not a note.** You tried `just fetch-reference DOI:10.3390/biomedicines10092125` and got an abstract-only cache because MDPI returns 403. The PMID route works: `just fetch-reference PMID:36140227` yields `content_type: full_text_xml` via PMC. So the Jo 2022 outcome is snippet-verifiable after all, and the entry now carries a `Cochlear Implantation` treatment with three evidence items — the implantation and its 3- and 6-month speech improvement, the clinical situation that led to it (one ear progressed to profound while the asymmetry persisted), and the authors' favourable-outcomes conclusion graded PARTIAL because it rests on one implanted patient. No follow-up beyond six months is reported for that patient; see the note at the end of this comment for why the two-year figure in the same paper is not hers.

Worth flagging as a general point rather than a criticism: an MDPI DOI 403-ing does not mean the full text is unreachable. It is worth trying the PMID before writing a caveat.

The same paper also gave a middle-ear-implantation case — a patient with bilateral moderate loss who chose MEI over hearing aids after an unsatisfactory experience with conventional aids. That is now a `PARTIAL` item on the amplification treatment, because it qualifies aids rather than supporting them.

`treatment_term` on cochlear implantation is bound to NCIT:C15329 Surgical Procedure. NCIT's cochlear implant concepts are device terms and are not reachable from NCIT:C25218, so they fail `TreatmentActionTerm`; `therapeutic_modality` is `DEVICE` rather than `SURGERY` because the platform is the implant and surgery is the route. Reasoning is in the treatment `notes` so it can be argued with.

**Finding 3 — asymmetry frequency.** Downgraded to `FREQUENT`. Added the Wesdorp counterweight you quoted, a symmetric individual from the Chinese family, and — from the newly fetched Jo 2022 — the only count with an explicit threshold and denominator: "Asymmetric hearing loss (interaural difference > 15dB) was identified in five of the nine patients." The `notes` now lays out the spread across all four sources rather than asserting a band.

**Finding 4 — downsloping/high-frequency audiogram.** Added as a phenotype, bound to HP:0001757 (High-frequency sensorineural hearing impairment) rather than HP:0005101, matching the DFNA41 and DFNB103 entries in this batch. HPO has no term for a downsloping configuration as such, so the shape is carried by the name and description, with that stated in `notes`.

**Suggestions taken:** the cVEMP saccular localisation on the vestibular phenotype, the screening recommendation in `diagnosis`, and — the most useful one — "the absence of an HI phenotype in mice heterozygous for loss of function variants of Lmx1a is not supportive of haploinsufficiency as the disease mechanism" into the `HUMAN_MODEL_MISMATCH` discussion, graded `REFUTE`. That quote cuts against this entry's own mechanism, which is exactly why it belongs there: later functional work resolved the mechanism in favour of haploinsufficiency, but the mouse observation the 2018 objection rested on has never been explained, and that residue is the gap.

I also acted on the GO scale observation: `GO:0007605` (organism-level) on a `CELLULAR` node is replaced by `GO:0050910` and `GO:0050973`, which are the cellular detection processes and cover the vestibular half of the node too.

---

**One more thing, and it is the important one.** Drafting the Jo 2022 material, I nearly shipped two findings that are not DFNA7. That paper reports four transcription-factor genes side by side — POU3F4, POU4F3, LMX1A, EYA4 — and its implantation results are not all LMX1A:

- the middle-ear-implantation patient (SB216) is in the **POU4F3** section;
- the patient implanted at 80 whose K-CID rose from 18% to 60% at two years carries an **EYA4** variant;
- the three implantees with favourable one-year K-CID/PB/spondee scores are **POU4F3**.

Only patient SB727 is LMX1A. Both non-LMX1A items were in my working draft and are removed; the cochlear implantation treatment now carries only SB727's results, and the treatment `notes` records the trap so the next person mining that paper does not fall into it. Flagging it here rather than quietly fixing it, because it is Named Entity Confusion reached through a multi-gene cohort paper rather than through a deep-research report, and the preflight gate does not catch that shape.

**One suggestion I did not take.** The 1q23.3q24.1 whole-gene deletion. The cache line-wraps it across a hyphenated break ("heterozy -\ngous"), and I did not want to risk a quote that depends on how the validator dehyphenates. Worth adding if someone can confirm the folded form matches.

Validation: `just validate` 0, `just validate-terms` 0, `just check-entity-refs` 0, `just check-duplicate-keys` 0, 76/76 snippets verified. Baseline gates re-run after the changes.

kevinschaper:
Rebased onto current `origin/main`. `check-snippet-grading` had started failing locally on `Metastatic_Ovarian_Cancer.yaml`, a file this branch never touches — the `--against-ref` baseline no longer matched because main had moved. It passes after the rebase, along with the other four baseline gates, `check-duplicate-keys`, and `validate-disorders`. The diff is unchanged apart from that, and the cache diff is still only the three rows this entry introduces.

PR #9917, opened 2026-08-28
treatments: normalize cochlear-implantation term bindings, and say why

caufieldjh:
The cochlear-implantation item in #9787 was filed against COQ6 as a one-off. It isn't — it's a KB-wide inconsistency. This sweeps it, and writes the reasoning into `CLAUDE.md` so the next curator doesn't re-derive it for the sixth time.

**Note: this PR touches `CLAUDE.md` alongside `kb/`, so it carries the `scope-override` label.** The alternative was splitting a rule away from the files that motivate it, which makes both halves harder to review.

> **Revised across four commits after review.** The first census undercounted by more than half, three instances of the defect were left standing, four label echoes were missed, and a sixth instance landed on `main` mid-review. All corrected below.

## The census

My first pass grepped top-level `treatments` only, and reported seventeen. Walking the YAML properly — `treatments` **plus `has_subtypes[].treatments`**, any name containing "cochlear" — gave 44. Re-counting on the current branch, after six more such treatments arrived from `main` during review, gives **50**:

| term | label | count |
|---|---|---|
| `NCIT:C15329` | Surgical Procedure | 42 |
| `NCIT:C15315` | Rehabilitation | 2 |
| `NCIT:C15747` | Supportive Care | 2 |
| `NCIT:C15238` | Gene Therapy | 1 |
| `NCIT:C49236` | Therapeutic Procedure | 1 |
| *(none)* | — | 2 |

Each divergent one was read individually rather than normalized to the majority. No `preferred_term` in the set echoes `Surgical Procedure`.

## Fixed: term contradicts the treatment

Five treatments named plainly `Cochlear Implantation`, bound to a term that is not implantation. In each, the descriptor and the binding disagreed inside a single block:

- `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104` — `C15315`, `preferred_term: cochlear implantation`
- `Charcot-Marie-Tooth_Disease_Type_X` — `C15315`, `preferred_term: Rehabilitation` (both traps at once)
- `Dominant_Deafness-Onychodystrophy_Syndrome` — `C15315`, `preferred_term: cochlear implantation`
- `Jervell_and_Lange-Nielsen_Syndrome_1` — `C15315`, `preferred_term: cochlear implantation`
- `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97` — `C15315`, `preferred_term: cochlear implantation`

All five now bind `NCIT:C15329` with `preferred_term: cochlear device implantation`, plus the `NCIT:C16830` → `NCIT:C157820` device qualifier pair.

**Two of them cited a precedent that says the opposite.** DDOD justified Rehabilitation as *"following the pattern already used in Jervell_and_Lange-Nielsen_Syndrome_1 and CAPOS_Syndrome"* — but CAPOS binds `C15329`. ARNSHL_97 cited `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104.yaml` as the convention — an entry this PR re-binds. In both, the `NCIT:C157820` reasoning is correct and is kept; the precedent claim is replaced with a record of what the earlier version bound.

**ARNSHL_97 is a race, not an oversight.** It was added to `main` in `be4faf0de` seventeen minutes before this branch's then-final commit, so every census here predated it. It is fixed in this PR rather than a follow-up because the stale citation is *created by this merge*.

## Fixed: `preferred_term` carrying no information

Where the **term is right** and only the display label was wrong:

- `Leber_Congenital_Amaurosis_with_Early-Onset_Deafness` — `Rehabilitation` → `cochlear implantation with auditory rehabilitation`. The treatment is aids *or* implantation *with* rehabilitation, so `C15315` genuinely fits the bundle.
- `Aminoglycoside-Induced_Hearing_Loss` — `supportive care` → `hearing amplification and cochlear implantation`. Bundle leads with non-surgical amplification, so `C15747` stands.
- `Renal_Tubular_Acidosis_Distal_2_...` — same shape, same fix.
- `CAPOS_Syndrome`, `Alstrom_Syndrome`, `Auditory_Neuropathy`, `Perrault_Syndrome`, `Multiple_Mitochondrial_Dysfunctions_Syndrome_9B` — `surgical procedure` → `cochlear device implantation`. Right term, but the label was echoed back. The last four were missed by the earlier passes because those filtered on **term-ID divergence** and the term was never divergent; this defect class is about the label.

## Examined and deliberately left alone

- **`Juvenile-Onset_Diabetes_Mellitus_...`** — its own `notes` record that no implantation in a molecularly confirmed patient has been published. Binding a surgical term would assert an operation that never happened in this disease. Reason now written into the entry so a later sweep doesn't "fix" it.
- **`MERRF_Syndrome`** (`C49236`) — `preferred_term: hearing aid or cochlear implantation` is already specific, and the bundle leads with aids. Reasoning moved into `notes`, where the next sweep will see it.
- **`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93`** — `Cochlear Gene Replacement` on `C15238` with `preferred_term: gene therapy` is a broader-binding echo under the new rule, but it is gene therapy, not implantation. Follow-up material.
- **`Keratoderma_Hereditarium_Mutilans`** and **`Mandibulofacial_Dysostosis_with_Microcephaly`** have **no `treatment_term` at all.** That is a different defect — absence rather than wrongness — and out of scope here. Reported rather than silently skipped.

## The `CLAUDE.md` rule, corrected

Three defects in my first draft of it, all caught in review:

- **The cross-reference was dead.** `#preferred_term-vs-ontology-term-labels` matches no heading in this repo; the only occurrence was my own new line. Retargeted to the real `Ontology Term Contract` section, which already states that `preferred_term` may be more specific than the best available term.
- **The rule was broader than its evidence.** "must not just echo the ontology label" contradicts `dismech-terms/SKILL.md` ("Prefer the canonical label as `preferred_term` when no extra nuance is needed") and CLAUDE.md's own `preferred_term: Pharmacotherapy` example. Narrowed to the claim my examples actually make — echoing loses information when the binding is **broader than the treatment** — with the Pharmacotherapy case named explicitly as correct, so it doesn't read as licence for a mass rewrite.
- **The "Better still" paragraph recommended a deprecated slot without saying so.** `qualifiers` is deprecated for the common clinical qualifiers that have dedicated slots, but a device attached to an action is exactly the predicate-value pattern the *Descriptor Qualifier Slots* section reserves it for. The paragraph now names that carve-out and links to it, so the two sections no longer read as contradicting each other.

The `NCIT:C157820` claim itself I verified with `runoak`: zero `C25218` ancestors, definition *"A two part electronic device…"*. `C15329` does have `C25218` as an ancestor. That reasoning has now been found independently re-derived in **six** separate entries — which is the cost the rule exists to remove.

## Validation

```
just validate-disorders <15 changed KB files>   # exit 0, 1027/1030 snippets (3 skipped by prefix)
just check-duplicate-keys                       # clean, 4249 files
just check-entity-refs                          # clean, 2769 KB files
just check-snippet-length / -title-snippets / -snippet-grading / -folded-hyphens / -reference-titles
just validate-history-all                       # clean, 7721 records
```

No cache churn. **No evidence items were added or altered** — the changes are `treatment_term` bindings, `preferred_term` strings, `qualifiers` blocks, and `notes`.

Companion to #9912, which fixes the COQ6 instance in its own entry. Part of #9787.

### DIFF
--- a/history/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97/2026-08-29T225115Z-claude-code-881163.yaml
+++ b/history/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97/2026-08-29T225115Z-claude-code-881163.yaml
@@ -1,0 +1,29 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97
+  path: kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97.yaml
+session:
+  id: 2026-08-29T225115Z-claude-code-881163
+  timestamp: '2026-08-29T22:51:15Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9917
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  summary: Rebind cochlear implantation to the surgical action term
+  details: |
+    Re-bound the Cochlear Implantation treatment from NCIT:C15315 Rehabilitation to NCIT:C15329 Surgical Procedure, set preferred_term to 'cochlear device implantation', and added the NCIT:C16830 -> NCIT:C157820 device qualifier pair, matching the convention this PR establishes across the KB.
+
+    The entry landed on main while PR 9917 was in review, so every census in that PR predates it. Its notes cited Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104.yaml as the precedent for the Rehabilitation binding; PR 9917 re-binds that entry to NCIT:C15329, so merging would have left the citation pointing at a counterexample. The notes now record what the earlier version bound rather than deleting the claim.
+
+    The deliberate no-evidence rationale and target_mechanisms block are unchanged.
--- a/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97.yaml
+++ b/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97.yaml
@@ -661,10 +661,21 @@ treatments:
 - name: Cochlear Implantation
   therapeutic_modality: DEVICE
   treatment_term:
-    preferred_term: cochlear implantation
+    preferred_term: cochlear device implantation
     term:
-      id: NCIT:C15315
-      label: Rehabilitation
+      id: NCIT:C15329
+      label: Surgical Procedure
+    qualifiers:
+    - predicate:
+        preferred_term: medical device
+        term:
+          id: NCIT:C16830
+          label: Medical Device
+      value:
+        preferred_term: cochlear implant
+        term:
+          id: NCIT:C157820
+          label: Cochlear Implant
   description: >-
     The standard intervention for severe prelingual sensorineural hearing loss, and the
     mechanism of this disease argues for it more directly than most. The lesion is in the
@@ -689,11 +700,13 @@ treatments:
     the strial mechanism and is labelled as such; the published DFNB97 literature reports no
     implant outcomes.
 
-    The treatment term is the generic Rehabilitation action. NCIT:C157820 Cochlear Implant
-    names the device rather than a clinical action and is not reachable from NCIT:C25218, so
-    it fails the TreatmentActionTerm dynamic enum; the specificity is carried by
-    preferred_term, following the convention in
-    Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104.yaml.
+    NCIT:C157820 Cochlear Implant names the device rather than a clinical action and is not
+    reachable from NCIT:C25218, so it cannot be the term: of a TreatmentTerm; the specificity
+    is carried by preferred_term and by the NCIT:C16830 -> NCIT:C157820 qualifier pair. The
+    generic action bound here is the surgical one, which is what implantation is. An earlier
+    version bound NCIT:C15315 Rehabilitation and cited
+    Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104.yaml as the convention; that entry
+    binds NCIT:C15329.
 
 - name: Early Identification and Habilitation
   therapeutic_modality: BEHAVIORAL


caufieldjh:
You were right on every point, and the census one is the embarrassing kind of wrong: I grepped top-level `treatments` and never touched `has_subtypes[].treatments`, so I audited roughly a third of the corpus and wrote up the result as if it were the whole thing. Walking the YAML properly gives **44**, matching your count. I've swept the rest and rewritten the PR body.

**The three left standing are fixed** — `Charcot-Marie-Tooth_Disease_Type_X`, `Dominant_Deafness-Onychodystrophy_Syndrome`, `Jervell_and_Lange-Nielsen_Syndrome_1`, all to `NCIT:C15329` with `preferred_term: cochlear device implantation`. Your point about *why* it mattered is the one that decided it for me: shipping a `CLAUDE.md` rule while three entries carry `notes` asserting the opposite is deliberate would have left the rule and the KB disagreeing **in writing**, which is worse than the re-derivation cost the rule was meant to remove.

**The DDOD precedent claim** is fixed too. It justified Rehabilitation as following `CAPOS_Syndrome`, which binds `C15329` — so the note was citing a counterexample as support. The correct `NCIT:C157820` reasoning in it is kept; the wrong precedent is gone.

**`Renal_Tubular_Acidosis_Distal_2`** — `preferred_term` fixed, term kept, same call as `Aminoglycoside-Induced_Hearing_Loss`.

**I also fixed `CAPOS_Syndrome` itself**, which you didn't list: right term, but `preferred_term: surgical procedure` echoes its own label. It's the second trap in the rule I'm adding, sitting in the entry I was citing as the good precedent.

**The dead link** — confirmed, `#preferred_term-vs-ontology-term-labels` matches no heading in this repo; the only occurrence was my own new line. Retargeted to `Ontology Term Contract`, which already carries the "may be more specific than the best available term" invariant.

**The overbroad rule** — this is the correction I'm most glad you made. "must not just echo the ontology label" does contradict `dismech-terms/SKILL.md` and CLAUDE.md's own `preferred_term: Pharmacotherapy` example, and as written it would have licensed a mass rewrite of correct bindings. Narrowed to the claim my examples actually support — echoing loses information when the binding is **broader than the treatment** — with the Pharmacotherapy case named explicitly as correct.

**Left alone, and now said out loud in the body rather than omitted:** `MERRF_Syndrome` (`C49236`, `preferred_term` already specific, bundle leads with aids), and `Keratoderma_Hereditarium_Mutilans` / `Mandibulofacial_Dysostosis_with_Microcephaly`, which have **no `treatment_term` at all** — absence rather than wrongness, a different defect, out of scope here.

**Your suggestion re ARNSHL_104's `preferred_term`** — also true, also undisclosed, now in the body.

One thing worth recording: the `NCIT:C157820` reasoning has now turned up independently re-derived in **five** entries, including `CAPOS_Syndrome`'s treatment description. That is the strongest argument for the rule.

```
just validate-disorders <five new files>  exit 0, 365/365 snippets verified
just validate-history <five new records>  No issues found
```

caufieldjh:
Fixed, and thank you for catching it — this was the PR's own defect reappearing one field over.

**The stale notes.** You're right that it's worse than the original, and the reason is the bit I'd have missed: the first version had `preferred_term: cochlear implantation` against `label: Rehabilitation`, where a reader sees both halves and can work out which is wrong. A note asserting the term is *deliberately* Rehabilitation points the reader at the term as the error. Someone trusting it reverts the fix. And the two propped each other up — ARNSHL104 cited JLNS1 as the convention while JLNS1's note was stale too.

Both rewritten on the DDOD wording, including its honesty about what the earlier version bound rather than quietly deleting the claim.

**I swept the KB for the same shape rather than fixing only the two you found.** One other file's notes mention a generic Rehabilitation action — `AFG2A-Related_Encephalopathy` — and there it is *accurate*: `C15315` for a hearing-habilitation treatment, which genuinely is rehabilitation. No further instances.

**The `qualifiers` suggestion is the better pattern and I've adopted it, not just named it.** `Labyrinthitis` and `Otofacial_Neurodevelopmental_Syndrome` were doing something my rule didn't know about: binding the action to `C15329` while attaching the device as an `NCIT:C16830` → `NCIT:C157820` predicate-value pair. That keeps the device concept *queryable*, which free-text `preferred_term` cannot — so my "carry the specificity in `preferred_term`" advice was settling for the lossy option. Both entries I retermed now carry the qualifier, `CLAUDE.md` names it as the higher-fidelity choice with a worked block, and the wording is softened from "fails validation" to "cannot be the `term:` of a `TreatmentTerm`", which is the precise claim.

**MERRF** now records the reasoning in `notes`. You're right that leaving it in a history record puts it exactly where the next sweep won't look — which is what the rule this PR adds specifically asks against, so it was inconsistent of me.

```
just validate-disorders <three changed files>  exit 0, 219/219 snippets verified
just validate-history <three new records>      No issues found
```

Separately, on your earlier note about the PR body still saying PLoS Biology on #9913 — that one had already been corrected before the review ran; you were reading a stale copy. Everything else in that review was right, including the retraction, which I've acted on there.

caufieldjh:
Blocking item fixed, all three inline suggestions applied, and the branch rebased onto current `main` — which turned out to be necessary for a reason nobody had flagged yet.

**The blocking finding was right, and the diagnosis of *why* it was missed was the useful part.** The earlier census passes filtered on term-ID divergence, so an entry whose term was already correct never surfaced — but defect class 2 is about the label, not the term. Four entries fixed, all named plainly `Cochlear Implantation`, all correctly bound to `NCIT:C15329`, all echoing the label:

| Entry | Was | Now |
|---|---|---|
| `Alstrom_Syndrome` | `surgical procedure` | `cochlear device implantation` |
| `Auditory_Neuropathy` | `surgical procedure` | `cochlear device implantation` |
| `Perrault_Syndrome` | `Surgical Procedure` | `cochlear device implantation` |
| `Multiple_Mitochondrial_Dysfunctions_Syndrome_9B` | `surgical procedure` | `cochlear device implantation` |

Without these, four entries would have visibly contradicted the CLAUDE.md rule this PR adds, in the exact treatment concept the rule is written about.

All three inline suggestions taken: `CAPOS_Syndrome`, `Charcot-Marie-Tooth_Disease_Type_X` and `Dominant_Deafness-Onychodystrophy_Syndrome` now carry the `NCIT:C16830` / `NCIT:C157820` qualifier pair, so all five re-bound entries land on the same rung. And the "Better still" paragraph now names the *Descriptor Qualifier Slots* carve-out it stands on, with a working anchor — a reader hitting the two sections in either order no longer gets contradictory instructions.

## The rebase, and why it was not optional

This branch was based on `760726b63`, which predates the #10003 `EvidenceItemSupportEnum` narrowing. Its schema still permits `PARTIAL` and `WRONG_STATEMENT`; `main`'s does not.

That matters more than staleness usually does, because **`main` has already cleared those values from the very files this PR touches** — 0 in each of `CAPOS_Syndrome`, `Charcot-Marie-Tooth_Disease_Type_X`, `Aminoglycoside-Induced_Hearing_Loss`, `MERRF_Syndrome` — while this branch still carried **59 across 9 files**. Merging as-is would have reintroduced retired enum values into files that had been cleaned, and turned CI red on `main` rather than here. CI on this PR was green only because it ran on 2026-08-28 against a schema that still allowed them.

Rebase was clean — four commits, no conflicts, no unrelated deletions or protected-path churn. Post-rebase the count in touched files is **zero**.

## Validation after the rebase

- `just validate-disorders` over all 14 changed disorder files — **993/996 snippets** verified (3 skipped by `DOI:` prefix)
- `check-duplicate-keys` (4,249 files), `check-entity-refs` (2,769 KB files), `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` — all clean, no new baseline entries
- `validate-history-all` — clean across 7,279 records
- Four new history records for the newly touched entries

One correction to the PR body's closing line, which the previous round flagged as stale and is now staler: this PR changes `treatment_term` bindings, `preferred_term` strings, `notes`, **and** `qualifiers` blocks. Still no evidence items added or altered.

---

caufieldjh:
Fixed the sixth instance in this PR rather than a follow-up, for the reason given: the stale citation is created by this merge, so it belongs here.

`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97` is now re-bound to `NCIT:C15329` with `preferred_term: cochlear device implantation` and the `NCIT:C16830` → `NCIT:C157820` qualifier pair. Its notes keep the device-vs-action paragraph and the deliberate no-evidence rationale, drop the "generic Rehabilitation action" sentence, and record what the earlier version bound rather than deleting the claim. `target_mechanisms` untouched.

**Re-counted on the current branch, not on the old base.** 50 cochlear treatments (`treatments` + `has_subtypes[].treatments`, any name containing "cochlear"): `NCIT:C15329` ×42, `NCIT:C15315` ×2, `NCIT:C15747` ×2, `NCIT:C15238` ×1, `NCIT:C49236` ×1, none ×2. Every remaining divergence is one of the documented exceptions — Juvenile-Onset Diabetes and Leber on `C15315`, Aminoglycoside-Induced and Renal Tubular Acidosis Distal 2 on `C15747` (amplification-led bundles), MERRF on `C49236`. No `preferred_term` echoes `Surgical Procedure` anywhere in the set.

Two of your non-blocking items I am leaving:

- **`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93`** (`Cochlear Gene Replacement` → `NCIT:C15238`, `preferred_term: gene therapy`) is a broader-binding echo under the new rule, but it is gene therapy, not implantation — a different treatment concept with its own precedent question. Follow-up material alongside the two entries with no `treatment_term` at all.
- **`Labyrinthitis` and `Otofacial_Neurodevelopmental_Syndrome`** keep `preferred_term: implantation`. Worth aligning, but they are not cochlear-implantation entries this sweep otherwise touches, and growing the diff to reach them trades a real review cost for a cosmetic gain.

The PR body's closing line is stale in the way you noted — the third commit also added `qualifiers` blocks. Updating it now.

## Validation

`just validate-disorders` over all 15 changed KB files: exit 0, **1027/1030 snippets** verified (3 skipped by `DOI:` prefix), no cache churn. `check-duplicate-keys` clean across 4,249 files; `check-entity-refs` clean across 2,769 KB files; snippet-length, title-snippets, snippet-grading, folded-hyphens and reference-titles all report no new violations; `validate-history-all` clean across 7,721 records. No evidence items were added or altered.

---

Issue #8727, opened 2026-08-17
Process: `CURATE_ROOT_WITH_SUBTYPES` is a priority hint, not a lump-vs-split ruling (umbrella Disease vs Grouping)


matentzn:
## Plain-language write-up of a modeling mistake we just made, and how to not repeat it

This issue is a readable explainer for curators and agents. It complements the tracking issue #8661 (the general agent-guidance failure) and the umbrella-to-grouping strategy in #4490. The concrete case that triggered it is the maintainer discussion on #8655 (hereditary amyloidosis).

> **The simple version (per @cmungall — [comment](https://github.com/monarch-initiative/dismech/issues/8727#issuecomment-5317016499)):** Each dismech entry should be **one reasonably conserved pathograph — a discrete disease entity**. Things that are *clearly* groupings of distinct diseases go under `kb/groupings/` instead. We won't always know the boundary and may get it wrong sometimes — that's fine, entries are easy to refactor. Everything below is the postmortem detail, not a new rulebook — don't over-think it (and modules are a side-note, not part of the decision).

### What happened (the origin of the "wrong" call)

The priority dashboard emits an **action code** next to each candidate disease. One of those codes is `CURATE_ROOT_WITH_SUBTYPES`. Over the last two batches we (agent-driven curation) read that code literally as an instruction: *"create one `Disease` entry for the umbrella term and list its variants inside it with `has_subtypes`."* We then filed issues and curated umbrella `Disease` entries on that basis.

That reading is wrong, and here is the mechanical reason it fired so often: the prioritizer emits `CURATE_ROOT_WITH_SUBTYPES` when a MONDO term simply **has several children** *and its label doesn't match a small hand-built "grouping-term" regex*. For example, `MONDO:0018634` (hereditary amyloidosis) has nine children and didn't match that regex, so it got tagged `CURATE_ROOT_WITH_SUBTYPES`. **That is a ranking heuristic, not a biological or modeling decision** — the prioritizer's own docs say these heuristics don't settle lump-vs-split. We treated a hint as a ruling.

### Why it's wrong (the modeling principle)

There are two different shapes, and the difference is real:

[...]

PR #9919, opened 2026-08-28
curate: Add Mitochondrial Complex I Deficiency, Nuclear Type 34 (MONDO:0032910, NDUFAF8)

### LINKED ISSUE #9901: Curate mitochondrial complex I deficiency, nuclear type 34 (MONDO:0032910)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 34** ([MONDO:0032910](https://monarchinitiative.org/MONDO:0032910)) — biallelic *NDUFAF8* variants impair assembly of respiratory chain complex I, giving early-onset Leigh-like mitochondrial disease.

The stub records a single causal gene (`hgnc:33551` NDUFAF8) and no MONDO descendants, so this looks like a leaf disease rather than a grouping. Note the KB already holds sibling `mitochondrial complex I deficiency, nuclear type N` entries; check how those handled the relationship to the `MONDO:0100223` parent before choosing.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34.yaml`

kevinschaper:
Adds a new disorder entry for **mitochondrial complex I deficiency, nuclear type 34** — the isolated complex I deficiency caused by biallelic variants in *NDUFAF8*. Nine patients have ever been reported, in two cohorts six years apart, and the interesting curation question was whether those two cohorts are one disease or two. This entry says one, and explains why.

Closes #9901

## entry_type decision: DISEASE

The stub gives one causal gene and no MONDO descendants, so grouping was never in play. The real alternatives were a standalone entry or a subtype of the `MONDO:0100223` parent. Standalone, for two reasons:

- There is one conserved pathograph — assembly-factor loss, then a stalled early assembly step, then an isolated complex I deficiency. That is what `CLAUDE.md` and #8727 ask for.
- The knowledge base already holds five sibling `mitochondrial complex I deficiency, nuclear type N` entries (types 2, 6, 12, 24, 31), all standalone `kb/disorders/` files. None of them appears as a `has_subtypes` entry anywhere, and none carries a `has_subtypes` block of its own. Their `parents:` values vary — types 2, 12 and 24 say `hereditary disease`, type 6 says `Mitochondrial Disease`, type 31 says `Mitochondrial Complex I Deficiency` and `Mitochondrial Disease`. This entry follows type 31, the most recently curated sibling and the one the research comment on #9901 identifies as the better model.

I followed the shape of `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24` closely, including its convention of recording facts that exist only in a paywalled full text as explicitly labelled "uncited leads" in prose, never as evidence.

## Why the two cohorts are one entity

The gene was defined in 2020 in three children with Leigh syndrome (PMID:31866046). In 2026 a targeted screen of complex I genes in patients with optic atrophy found six more patients with biallelic *NDUFAF8* variants, five of whom had **isolated** optic atrophy, and associated the gene with recessive Leber hereditary optic neuropathy for the first time (PMID:41234160).

Two things join them rather than one:

1. **A shared allele.** The recurrent deep-intronic `c.195+271C>T` from the Leigh families turns up homozygous in two of the optic-atrophy families, and the 2026 paper says so explicitly.
2. **A shared tissue.** The deep-research report, which read the 2020 full text, records optic atrophy and visual impairment in its first subject and small optic nerves with nystagmus in its second. So the optic nerve is involved at *both* ends of the severity range. I initially drafted this entry describing the two cohorts as looking "nothing like each other" and had to correct it — the phenotypes are a continuum, not two disjoint pictures, and the entry now says that.

## Mechanism

`NDUFAF8` is an assembly factor, not a subunit. It partners with NDUFAF5 to stabilise the core Q-module subunit NDUFS7 at the earliest stage of assembly, and NDUFAF5 in turn needs NDUFAF8 to be stable — so losing one protein removes the module. The best quantitative measurement anywhere in this disease is NDUFAF5 sitting at 18% of control in a patient's fibroblasts, which is a measurement of the partner rather than of NDUFAF8 itself. The entry says so.

One optic-atrophy allele, `p.(L15R)`, sits in the mitochondrial targeting sequence rather than in any functional domain. Separate cell-biology work (PMID:37159021) shows NDUFAF8 reaches the matrix by an unusual two-step route through the intermembrane-space disulfide relay, policed at both stages by proteases. That is a very good fit for a targeting-sequence variant — but nobody has tested the two together, so it is curated as a `HUMAN_MODEL_MISMATCH` discussion rather than as a causal claim.

*(Updated after review: the first bullet below was wrong and has been fixed on the branch. There is now a `treatments:` block for idebenone, because PMID:41234160 recommends it for this patient class. See the review-response comment below.)*

- ~~**No `treatments:` block.** No therapy or trial has been reported for any *NDUFAF8* patient, including in the full text.~~
- **No `progression:` block.** This one is different, and the entry keeps the difference visible. Age at onset and clinical course *are* reported in the 2020 paper — onset between three and fifteen months, two of three children dead by four years — but that paper is abstract-only in the reference cache, so nothing from it can be quoted. The values are recorded as uncited leads in an open `KNOWLEDGE_GAP` discussion, which states plainly that the course is known and not quotable rather than unknown.
- **The Leigh phenotype is curated as a diagnosis, not decomposed into HPO signs.** The source states the diagnosis in all three subjects and separately defines the syndrome; it never describes an individual child's imaging. HPO has no term for Leigh syndrome, so `phenotype_term` carries free text with no binding rather than a substituted approximation like "abnormal basal ganglia morphology", which would assert an imaging finding nobody reported per patient.

## Deep research

Provider: falcon. Named Entity Confusion preflight is a **WARN**, and I read it before using the report: *NDUFAF8* is mentioned 61 times with the next gene being its own partner NDUFAF5 at 10, so the report is unambiguously about the right disease. The only warning is that it cites OMIM 256000 — Leigh syndrome, the parent phenotype — instead of 618776 for this entity. That is loose nomenclature, not a different disease.

The report is committed here because the entry cites it by name for the uncited leads.

## Validation

All run in this worktree, exit codes checked:

| check | exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just validate-disorders` | 0 (37/37 snippets verified) |
| `just check-folded-hyphens` | 0 (caught one real defect, fixed) |

### DIFF
--- a/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34/2026-08-28T223901Z-claude-code-c3a649.yaml
+++ b/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34/2026-08-28T223901Z-claude-code-c3a649.yaml
@@ -1,0 +1,45 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34
+  path: kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34.yaml
+session:
+  id: 2026-08-28T223901Z-claude-code-c3a649
+  timestamp: '2026-08-28T22:39:01Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9901
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9919
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - discussions
+  summary: 'Fourth review response: correct five evidence_source values on in-silico
+    and fibroblast results'
+  details: |
+    Fourth review response on PR #9919. A second review at 22:32Z superseded the 22:10Z approval on the same commit, and its single blocking finding was correct.
+
+    Five evidence items citing PMID:41234160 were graded HUMAN_CLINICAL while describing in-silico or cultured-cell work. Both provenance claims were checked against the cached full text before changing anything.
+
+    Two items quote the sentence predicting that the p.L15R change affects the mitochondrial targeting sequence. The paper's methods say "Mitochondrial targeting sequences (MTS) were identified using MTSviewer and variants reported in the 3D structure predicted by AF3", with delta-delta-G from DynaMut2 and DUET. That is an in-silico prediction and no patient was measured to produce it, so both are now COMPUTATIONAL. The sentence is used twice - on the abundance node and in the HUMAN_MODEL_MISMATCH discussion - and both uses carry the same value, so the same-sentence grading rule still holds.
+
+    Three items quote the TMT proteomics results: NDUFAF8 not detected in the batch, the NDUFAF5 fold change of 0.18, and the complex I fold change of 0.73. The sentence immediately preceding them in the paper is "we performed quantitative proteomics on fibroblasts derived from individual II-4 in a batch of 11 samples". Cultured cells are IN_VITRO, and all three are now graded that way.
+
+    The decisive argument was internal consistency, not the classification rule in the abstract. The same file already grades PMID:31866046's fibroblast complexome profiling and enzymology IN_VITRO in seven places and PMID:37159021's cell work IN_VITRO in three, so the entry was saying fibroblast measurements are IN_VITRO when they come from the 2020 paper and HUMAN_CLINICAL when they come from the 2026 one. Both cannot be right.
+
+    Each of the five explanations now records why the value is what it is, so the reasoning sits next to the field rather than only in a PR comment. No snippet, claim or supports grade moved.
+
+    Audited the whole reference afterwards rather than only the five: of 23 items citing PMID:41234160, the split is now 15 HUMAN_CLINICAL, 3 OTHER, 3 IN_VITRO, 2 COMPUTATIONAL. The 18 untouched items are clinical course, optic atrophy findings, cohort composition, the idebenone recommendation and its mechanistic rationale, which are correctly graded.
+
+    Worth recording why no gate caught this: check-snippet-grading keys on the same sentence graded two different ways. These are five different sentences and the MTS snippet was consistently mis-graded across both of its uses, so the check passed. This class of error needs a human read of what the cited experiment actually was.
+
+    Validation: just validate exit 0; just validate-terms exit 0; just validate-disorders exit 0 with 47/47 snippets verified; check-entity-refs, check-duplicate-keys, check-folded-hyphens, check-snippet-length, check-title-snippets, check-snippet-grading and check-environmental-evidence all exit 0. Render test passes. No cache diff.
--- a/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34.yaml
+++ b/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_34.yaml
@@ -239,26 +239,32 @@ pathophysiology:
   - reference: PMID:41234160
     reference_title: "Recessive variants in mitochondrial complex I nuclear subunits are an underrated cause of optic atrophy."
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: COMPUTATIONAL
     snippet: >-
       the change is predicted to affect the MTS introducing a positive charge on
       the hydrophobic side of the amphipathic α-helix (Fig. 5D), possibly
       affecting NDUFAF8 import into the mitochondria.
     explanation: >-
-      The proposed import-failure route for the targeting-sequence allele. Quoted
-      with its own hedging intact: the paper says "possibly", and so does this
-      node.
+      The proposed import-failure route for the targeting-sequence allele.
+      Quoted with its own hedging intact: the paper says "possibly", and
+      so does this node. Graded COMPUTATIONAL: the paper's methods state
+      that MTS regions "were identified using MTSviewer" with variants
+      placed on an AlphaFold 3 predicted structure, so this sentence
+      reports an in-silico prediction and no patient was measured to
+      produce it.
   - reference: PMID:41234160
     reference_title: "Recessive variants in mitochondrial complex I nuclear subunits are an underrated cause of optic atrophy."
     supports: PARTIAL
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: IN_VITRO
     snippet: >-
       NDUFAF8 protein was not detected in the whole TMT 11-plex batch, not
       allowing us to directly evaluate the variant effect.
     explanation: >-
       Graded PARTIAL because it is the limitation on this node rather than
-      support for it: the abundance claim rests on the partner protein, not on
-      NDUFAF8 itself.
+      support for it: the abundance claim rests on the partner protein,
+      not on NDUFAF8 itself. Graded IN_VITRO: the measurement is
+      quantitative proteomics on cultured fibroblasts derived from the
+      patient, not an observation in the patient.
   - reference: PMID:37159021
     reference_title: "A two-step mitochondrial import pathway couples the disulfide relay with matrix complex I biogenesis."
     supports: PARTIAL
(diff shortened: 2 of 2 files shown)


kevinschaper:
Both blocking items fixed, all three suggestions taken, and I also acted on the research comment on #9901.

**1. The "no treatment reported" claim was wrong, and you were right about where.** PMID:41234160 — which this entry already cites as full text — concludes that patients with nuclear complex I optic neuropathy may benefit from idebenone at LHON dosing. There is now a `treatments:` block for it, carrying the paper's own qualification as `PARTIAL` evidence: visual improvement was variable and only 3 of 13 responded. I took your caveat seriously and said so three times — in the treatment description, in its explanation, and in the discussion — that the denominator spans the cohort's four assembly-factor genes rather than the NDUFAF8 patients, so no response rate attaches to this genotype. The `KNOWLEDGE_GAP` discussion and the `notes:` no longer claim an absolute absence; what is genuinely absent is any *outcome* attributable to this gene. The indication, optic atrophy, was already a curated phenotype.

**2. GeneReviews tagged.** PMID:26425749 added to `references:` with the `GeneReviews` tag plus a one-line note explaining it is not cited as evidence, following MC1DN2's wording.

**Suggestions:** dropped the uninformative `GO:0005515` (the node keeps its complex I assembly process term, which carries the claim); PMID:37159021 now carries two `PARTIAL` evidence items on the abundance node covering the two-step import route and the YME1L/CLPP protease surveillance, each explained as cell-line work that never touches the patient allele; and the subacute-onset sentence is now quoted on the optic atrophy phenotype, with a note that its twelve cases span five genes.

**From the research comment on #9901:**

- **PMID:39921456 reports a fourth patient**, so the count moves from nine to ten throughout. That patient also carried an mtDNA complex I variant, recorded in the genetic notes as a caution that a single-gene answer in this class is not automatically the whole answer.
- **Cross-referenced the TIMMDC1 sibling entry.** `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_31` is built around a deep-intronic allele exome cannot see; ours is the second such lesion among the nuclear complex I diseases here, and that pair is the practical reason this entry's diagnosis section says to look at intronic sequence.
- **Subject 3's normal blood lactate** (1.44 mmol/L, CSF 3 mmol/L) recorded as an uncited lead. It is the finding most likely to stop a clinician suspecting mitochondrial disease at all.
- **`parents:` changed** to `[Mitochondrial Complex I Deficiency, Mitochondrial Disease]`, following Type 31, which the research comment identifies as the better model among inconsistent siblings. I've corrected the PR description, which previously overstated the siblings' consistency on this.

**Deliberately not done:** PMID:41898833, the 2026 Goldenhar exome paper reporting a rare NDUFAF8 frameshift with limited segregation, is not cited. It is a candidate-gene list that did not establish causality for its own lead variant.

The branch had a cache-CSV conflict on rebase. Resolved main-authoritative: took `origin/main` as the base for both files and re-appended only the `MONDO:0032910` rows verbatim out of the prior commit, then `just normalize-cache`.

All gates re-run: `validate`, `validate-terms`, `validate-disorders` (43/43 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Cache diff is empty after the edits.

kevinschaper:
Two follow-up commits from a self-audit, both correcting my own prose rather than responding to the review.

**A number I had taken from a research comment rather than a source.** The genetic notes said "roughly thirty-four nuclear complex I diseases". Nothing this entry cites gives that figure, and the class-size number it *does* cite (PMID:22972949) is a different quantity — 38 nuclear-encoded subunits plus assembly factors. Rewritten to say what a reader can check: two entries in this knowledge base turn on a deep-intronic allele.

**A stale framing left behind by the fix.** After adding the fourth patient from PMID:39921456 to `prevalence:` and the genetic notes, the opening description still called the entity "two published cohorts". It now reads as ten patients across three papers, with the 2025 single case named in the Leigh arm. I checked whether the "two cohorts" language elsewhere in the file needed the same treatment and left it: those passages are specifically about the Leigh-versus-optic-atrophy contrast, which is still exactly what they are about.

kevinschaper:
Went back through the detailed review comment and picked up three things the summary did not spell out.

**`target_mechanisms` on the idebenone treatment.** You're right that the paper hands over the rationale for that edge. It now points at the complex I deficiency node with `treatment_effect: BYPASSES`, cited with the mechanism sentence — idebenone in its reduced form takes electrons from NADH straight to complex III. That is worth having explicit, because it explains why a small molecule is a candidate at all against a lesion no small molecule reverses: the assembly failure is not the target, its bioenergetic consequence is.

I took the over-modelling warning seriously. The entry does not claim any individual NDUFAF8 patient received idebenone or responded, and `notes:` now states the gap in your words — no NDUFAF8-specific trial, and no per-patient response separable from the cohort data, with the reason (the breakdown lives in a supplementary table that is not in the cache).

**`Discussion` does take `evidence:`** — thank you, I had put the PMID:37159021 citation only on the pathophysiology node. The `HUMAN_MODEL_MISMATCH` discussion now carries both halves of the mismatch as cited items: the import paper's claim that NDUFAF8 needs both oxidation and matrix import to work, and the patient-side structural prediction with its own hedging intact (graded `PARTIAL` for that reason). The mismatch is now stated with citations rather than asserted.

**The third suggestion exposed a real error, not just an omission.** The entry said nothing is known about the optic-atrophy cohort's course. That was wrong — the paper describes it, and I had read past it. The sentence is now quoted on the phenotype with the hedge it needs, and the discussion is rewritten: the optic-atrophy cohort is better served than the Leigh one, at cohort level, and what a `progression:` block would still need is which of the two onset patterns the six NDUFAF8 patients followed, which is not separable from a statement spanning eleven genes.

Also from a separate self-audit, in two earlier commits: dropped an unsourced "roughly thirty-four nuclear complex I diseases" figure that came from a research comment rather than any source here, and fixed the opening description, which still said "two published cohorts" after the fourth patient was added.

All gates re-run: `validate`, `validate-terms`, `validate-disorders` (47/47 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0.

kevinschaper:
Fixed, plus three of the four non-blocking notes.

**The denominator.** You're right and I had it backwards in two places. I went back to the cached full text: the idebenone sentence sits in the general clinical-features paragraph describing all 31 patients, immediately after the sentence listing which drugs were used — so the thirteen are most plausibly the patients who received idebenone, and the cache does not resolve it. Both prose passages now say that.

I also made the treatments description say explicitly *why* the assembly-factor reading has to be rejected, rather than just dropping it: that group happens to number thirteen too, and taking the coincidence for a match would imply six of the treated thirteen carry NDUFAF8 — a far tighter link than anything the paper states. Spelling out the trap seemed better than silently avoiding it, since the next reader will hit the same coincidence.

The separate claim on the optic atrophy node — 13 assembly-factor patients across four genes, six of them LHON-like — is untouched. That one is accurate, and the new prose now distinguishes it from the treatment denominator instead of blurring them.

**Non-blocking, taken:**

- `GO:0032981` on three consecutive nodes: the module node now carries `GO:0050821` protein stabilization, which is what it actually asserts — NDUFAF8 holding NDUFAF5 stable, measured as the 82% fall in NDUFAF5. Your suggestion, and better than dropping the binding. Reasoning written into the node description.
- The idebenone recommendation is regraded `OTHER`. It is an expert recommendation rather than an observation the study made, which is how this entry already grades the mechanistic rationale two items below — so `OTHER` is the internally consistent call.
- History record added covering the three commits that had none.

**Not done, on your advice:** the stale `GO:0005515` enum-cache row is left alone. It is derived, `just validate` leaves it, and hand-editing a cache row is worse than an unused but valid membership entry.

All gates re-run: `validate`, `validate-terms`, `validate-disorders` (47/47), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0, no cache diff.

kevinschaper:
Correct on all five, and the internal-consistency argument is what settles it rather than the classification rule in the abstract. Fixed.

I verified both provenance claims against the cached full text before touching anything:

- **The MTS prediction → `COMPUTATIONAL`.** Methods: *"Mitochondrial targeting sequences (MTS) were identified using MTSviewer and variants reported in the 3D structure predicted by AF3"*, with ΔΔG from DynaMut2/DUET. No patient was measured to produce that sentence. Used twice — abundance node and the `HUMAN_MODEL_MISMATCH` discussion — and both carry the same value, so the same-sentence rule still holds.
- **The three TMT items → `IN_VITRO`.** The sentence immediately before them: *"we performed quantitative proteomics on fibroblasts derived from individual II-4 in a batch of 11 samples"*.

**Your point about the file contradicting itself is the one that made this unarguable.** PMID:31866046's fibroblast complexome profiling and enzymology are `IN_VITRO` in seven places and PMID:37159021's cell work in three, so the entry was saying fibroblast measurements are `IN_VITRO` when they come from the 2020 paper and `HUMAN_CLINICAL` when they come from the 2026 one. That is not a judgement call with two defensible sides; it is one file giving two answers.

Each of the five `explanation` fields now records why the value is what it is, so the reasoning sits next to the field rather than only in this thread.

**I audited the whole reference rather than just the five you listed.** Of 23 items citing PMID:41234160 the split is now 15 `HUMAN_CLINICAL`, 3 `OTHER`, 3 `IN_VITRO`, 2 `COMPUTATIONAL`. The 18 untouched are clinical course, optic atrophy findings, cohort composition, and the idebenone recommendation and its mechanistic rationale — correctly graded, and matching your count.

Your note on why no gate caught it is worth having on the record: `check-snippet-grading` keys on *one sentence graded two ways*, and these were five different sentences each consistently mis-graded, so the check passed cleanly. The prose around them was honest about the hedging — the explanations already said "predicted from structure, not measured" — but that honesty was in the field a human reads and not in the field a consumer filters on. That is precisely the gap worth closing.

Gates: `validate`, `validate-terms`, `validate-disorders` (47/47), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Render passes. No cache diff.

PR #9920, opened 2026-08-28
curate: Add Hearing Loss Autosomal Recessive 120 (MONDO:0859374, MINAR2)

### LINKED ISSUE #9906: Curate hearing loss, autosomal recessive 120 (MONDO:0859374)
Curate a dismech entry for **hearing loss, autosomal recessive 120 (DFNB120)** ([MONDO:0859374](https://monarchinitiative.org/MONDO:0859374)) — biallelic *MINAR2* (KIAA1024L) variants cause severe-to-profound congenital sensorineural hearing loss.

The stub gives one causal gene (`hgnc:33914` MINAR2) and no descendants, and nothing in `kb/` mentions MINAR2. This is a recently described gene, so evidence may be thin — curate what is supported and record the gaps with `discussions` (`kind: KNOWLEDGE_GAP`) rather than padding.

Stub: `stubs/Hearing_Loss_Autosomal_Recessive_120.yaml`

kevinschaper:
Adds `kb/disorders/Hearing_Loss_Autosomal_Recessive_120.yaml` for DFNB120 (MONDO:0859374, MINAR2 `hgnc:33914`) and retires its stub.

`entry_type: DISEASE`. One causal gene, no MONDO descendants, and one pathograph that runs from biallelic MINAR2 loss to impaired hair cell mechanotransduction. Nothing in `kb/` previously mentioned MINAR2, so there is no parent entry this could sit under as a subtype.

## The gene is recent, so the entry is built to show its own seams

MINAR2 was named for structural similarity to MINAR1, not for a known function, and it was still formally uncharacterised when the deafness families were sequenced. Two mechanisms have been proposed since, in two different species, and **no experiment joins them**:

- **Mouse** — the shortest row of outer hair cell stereocilia is depleted, while ribbon synapse counts and organ of Corti architecture are normal at the age when thresholds are already raised. That places the primary defect at the top of the hair cell rather than at the synapse.
- **Zebrafish** — Minar2 sits on apical lysosomes and delivers cholesterol to the hair bundle membrane. Mutants have reduced bundle cholesterol, enlarged apical lysosomes, and impaired mechanotransduction, and the causal direction was tested by moving cholesterol pharmacologically in both directions.

These fit together comfortably as a lipid-delivery defect, and I did not curate them that way, because the two structural phenotypes point in **opposite directions**: fish bundles are longer and thinner, mouse stereocilia are shorter and fewer. They are separate nodes, with a `KNOWLEDGE_GAP` naming the measurement that would join them (cholesterol in mouse cochlear hair bundles, with the shortest row resolved).

## One node has no arrow out of it, on purpose

`Derepression of NOTCH2 and mTORC1 Signalling` is drawn as a consequence of MINAR2 loss with **no edge into any hearing phenotype**. The effects are real and measured — silencing MINAR2 raises NOTCH2, and Minar2-null adipose tissue has hyperactivated mTORC1 through loss of Raptor binding — but nothing connects either to a cochlear phenotype in any species, and the founding authors put it no more strongly than that Notch signalling *might* play a role. Drawing the arrow would state something the literature does not.

The `HUMAN_MODEL_MISMATCH` discussion attached to that node is the substantive one. Minar2 knockout mice are obese, glucose intolerant, and parkinsonian with nigral dopaminergic loss; affected humans examined between 4 and 80 years old had isolated deafness and normal bedside balance testing. Both readings are stated — a genuine species difference, or ascertainment through an audiology clinic that never measured fasting glucose — because the human evaluation reported is a clinical examination, not structured metabolic or dopaminergic phenotyping. The proposed experiment recalls the known cohort against heterozygous relatives.

## What I did not curate

**No treatment other than genetic counselling.** Cochlear implantation is standard management for congenital severe-to-profound sensorineural hearing loss, and no DFNB120-specific report documents it in any patient. Curating it would be importing general practice as evidence about this disease. The treatment `notes` says so explicitly.

Two model readouts are recorded as `UNCHANGED` rather than omitted, because they are real negative results — ribbon synapse counts are what exclude synaptopathy and make the stereocilia lesion primary.

## Deep research

Falcon report run and preflighted against MONDO:0859374: **PASS** (MINAR2 = 30 mentions, next gene at 4; OMIM 620238 matched on both sides). The entry was curated from primary sources regardless, with every snippet re-verified against the cache.

## Validation

- `just validate` — 0
- `just validate-terms` — 0
- `just validate-disorders` — 0, 59/59 snippets verified
- `just validate-history` — 0

Closes #9906

### DIFF
--- a/history/disorders/Hearing_Loss_Autosomal_Recessive_120/2026-08-31T174543Z-claude-code-f08ca5.yaml
+++ b/history/disorders/Hearing_Loss_Autosomal_Recessive_120/2026-08-31T174543Z-claude-code-f08ca5.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hearing_Loss_Autosomal_Recessive_120
+  path: kb/disorders/Hearing_Loss_Autosomal_Recessive_120.yaml
+session:
+  id: 2026-08-31T174543Z-claude-code-f08ca5
+  timestamp: '2026-08-31T17:45:43Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9920
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - evidence
+  summary: Address re-review of the PARTIAL migration
+  details: |
+    Re-review round after migrating the four retired 'supports: PARTIAL' items. Two of my migration choices were wrong and are corrected here. First, I had graded the authors' 'further biochemical studies are required to demonstrate direct binding' sentence as REFUTE; it reports no experiment, so it is absence of evidence rather than evidence against, and CLAUDE.md's migration table maps 'true and worth citing, but not about this claim' to NO_EVIDENCE. The structured field had also contradicted my own explanation, which said the cholesterol-delivery claim stands. Second, two explanations asserted that the taxonomy leaves 'directness' unset for indirect support; it does not, it maps indirect support to SUPPORT plus directness INDIRECT. Held against the nodes they attach to rather than against the hearing phenotype, both snippets assert their node's claim directly, so both now carry directness DIRECT with reworded explanations. Also bound GO:0007219 Notch signaling pathway on the derepression node, which previously bound only the mTORC1 arm. All gates exit 0 with no cache churn.
--- a/kb/disorders/Hearing_Loss_Autosomal_Recessive_120.yaml
+++ b/kb/disorders/Hearing_Loss_Autosomal_Recessive_120.yaml
@@ -229,15 +229,17 @@ pathophysiology:
       rather than to how much protein is present.
   - reference: PMID:36317962
     reference_title: "Kiaa1024L/Minar2 is essential for hearing by regulating cholesterol distribution in hair bundles."
-    supports: REFUTE
+    supports: NO_EVIDENCE
     evidence_source: IN_VITRO
     snippet: "Nevertheless, further biochemical studies are required to demonstrate direct binding between the Kiaa1024L/Minar2 protein and cholesterol."
     explanation: >-
-      Curated deliberately as a refutation of the stronger reading this node's other items
-      invite. The motif analysis, the recruitment assay and the docking study are consistent
-      with a cholesterol-binding protein; the authors state in the same paper that direct
-      binding is not demonstrated. The cholesterol-delivery claim stands, the binding
-      mechanism does not.
+      Carried deliberately as a limitation, not as a direction. The authors report no
+      experiment here: they state that direct binding has not been demonstrated, which is
+      absence of evidence rather than evidence against. The motif analysis, the recruitment
+      assay and the docking study are consistent with a cholesterol-binding protein without
+      establishing one. This node's claim is cholesterol delivery, which the paper's other
+      results support; the sentence is quoted so the binding mechanism is not read as
+      settled.
 - name: Depletion of the Shortest Row of Outer Hair Cell Stereocilia
   description: >-
     The mammalian structural lesion. In Minar2 null mice the shortest row of outer hair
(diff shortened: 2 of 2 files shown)


kevinschaper:
Both blocking findings were real and are fixed. Thanks — the second one in particular was a case of me writing a caveat instead of looking harder at the cache.

**DPOAE direction.** Flipped to `INCREASED`. The interpretation now states explicitly that the *threshold* rises while the underlying function falls, so the name and the direction cannot drift apart again on a future edit.

**Cholesterol reporter snippet.** Swapped in the sentence you pointed at, added "The decreases were not due to changes in transgene expression either" as its own evidence item, and deleted the explanation that blamed sentence-splitting in the cache. That explanation was wrong: the bracket-free sentence was there the whole time.

I also took four of the suggestions:

- `animal_models#Mouse` → `animal_models#Minar2 knockout mouse metabolic and motor phenotype`. You were right that it was resolving through the `species` fallback with two candidates.
- OMIM:620238 added as an `external_assertions` record.
- CRAC/CSD point mutation on the cholesterol node, using the expression-level control sentence rather than the effect sentence (which the cache splits on a decimal), plus the authors' own "further biochemical studies are required to demonstrate direct binding" caveat. That caveat matters more than the positive result — it stops the node reading as an established binding claim.
- Carlson & Avraham, fetched by DOI. Two quotes: the one you flagged about the oldest humans, and a second one that turned out to be more useful — MINAR2 is one of eight genes where the mouse is syndromic and the human is not. That reframes the mismatch as a known property of mouse deafness models rather than something peculiar to this gene.

**One suggestion I handled differently.** SREBP2 induction is now evidence on the existing cholesterol node rather than a downstream node of its own. It is genuinely useful — an independent readout that the accessible cholesterol pool is reduced, arrived at without the fluorescent reporter — but nothing links the compensatory response itself to the hearing phenotype, so a node would have had no defensible outgoing edge. Happy to promote it if you disagree.

Bilaterality and the differential-diagnosis list I left alone, for the reasons you gave.

Validation after the changes: `just validate` 0, `just validate-terms` 0, `just validate-disorders` 0, `just check-entity-refs` 0, `just check-duplicate-keys` 0, and `check-folded-hyphens` / `check-snippet-length` / `check-title-snippets` / `check-snippet-grading` / `check-environmental-evidence` all 0. 63/65 snippets verified — the two skipped are the DOI-prefixed Carlson quotes, which the audit skips by prefix and which I checked by hand against `references_cache/DOI_10.1073_pnas.2211351119.md`.

kevinschaper:
Reopening this after approval, because I found a false claim in my own entry and would rather lose the approval than ship it.

**The treatment `notes` said no DFNB120-specific report of cochlear implantation was found. That is wrong**, and the contradicting text is in a cache this PR already cites five dozen times. `PMID_35727972.md` reports three molecularly confirmed individuals from family 1 who received unilateral cochlear implants: the two siblings whose hearing was milder in early childhood and reached severe or profound by about age 10 were implanted at 12 and 10, and their brother, who failed the newborn hearing screen and was profound from the start, was implanted at 1. All three had improved or maintained oral communication.

Cochlear implantation is now a curated treatment with both quotes. The eleven-year spread in age at implantation *within one family* is the part worth having in the KB: timing here tracks the individual course, not the diagnosis, and the progressive minority reach candidacy years after the congenitally profound majority. The remaining note says only what is true — no amplification or speech-rehabilitation report exists.

I found this by going back and checking my own "no report was found" claims against the caches, after the review on the sibling DFNA7 PR in this batch showed the same claim was false there. It is a claim shape worth distrusting on sight: it asserts something about the whole literature from inside an entry, and no validator can check it.

**Also corrected the gene location** — MINAR2 is at 5q23.3, not 5p15.31. Checked against HGNC rather than memory, along with the other three genes in this batch.

**And a rebase.** `check-snippet-grading` started failing locally on `Metastatic_Ovarian_Cancer.yaml`, a file this branch never touches. That was staleness: `origin/main` had moved and the `--against-ref` baseline no longer matched the branch. Rebased onto current `origin/main`; the diff is unchanged apart from the commits above, and the cache diff is still only the five rows this entry introduces.

The two blocking findings and the four suggestions from your earlier review are all still in and unchanged. Sorry for the extra round.

kevinschaper:

PR #9923, opened 2026-08-28
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 103 (MONDO:0014469, CLIC5)

### LINKED ISSUE #9905: Curate autosomal recessive nonsyndromic hearing loss 103 (MONDO:0014469)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 103 (DFNB103)** ([MONDO:0014469](https://monarchinitiative.org/MONDO:0014469)) — biallelic *CLIC5* variants; CLIC5 localizes to the stereocilia taper region and is required for stereocilia maintenance in cochlear hair cells.

The stub gives one causal gene (`hgnc:13517` CLIC5) and no descendants, and nothing in `kb/` currently mentions CLIC5. The `jitterbug` mouse is the established model — if it is used, it belongs in `animal_models:` with `modeled_mechanisms`, not `experimental_models:`.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103.yaml`

kevinschaper:
Adds `kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103.yaml` for DFNB103 (MONDO:0014469, CLIC5 `hgnc:13517`) and retires its stub.

`entry_type: DISEASE`. One causal gene, no MONDO descendants, one pathograph end to end. Nothing in `kb/` previously mentioned CLIC5, so there is no parent entry this could sit under.

## The mechanism, and why the address matters

CLIC5 is in the chloride intracellular channel family and does not act as a channel here. It is a membrane-cytoskeletal adaptor that sits at the **taper region at the base of the stereocilium**, not at the tip — mass spectrometry of chicken utricle stereocilia puts it there in roughly 1:1 molar ratio with radixin, alongside taperin, PTPRQ and myosin VI.

That address predicts the phenotype. Losing a base-of-bundle tether does not shorten stereocilia; it lets adjacent ones **fuse**, from postnatal day 10 onwards in the mouse. And the partner proteins disperse out of the taper region *before* fusion is visible, which is the timing evidence that makes complex disassembly the primary lesion rather than a consequence of the structural failure. The dependency runs both ways — CLIC5 and radixin also mislocalise in myosin VI mutants — so the complex is mutually stabilising.

## Animal models, including the rescue

Both go in `animal_models:` with `modeled_mechanisms`, per the issue.

The **jitterbug mouse** carries `RECAPITULATES` links against the tethering and fusion nodes, with readouts for radixin staining and partner localisation. It is not corroborating a human observation — stereocilia taper composition cannot be examined in patients, so this is the only place the mechanism has been tested.

The **2025 AAV gene-replacement study** carries a `RESCUES` link. Its `limitations` say plainly what the abstract does not emphasise: **injections were at postnatal day 0, before the lesion begins between P14 and P17.** That demonstrates prevention, not reversal. Human DFNB103 is detected at a mean age of about 8 years — the equivalent of treating long after the mouse window closed. A `HUMAN_MODEL_MISMATCH` discussion carries that, and proposes the delayed-delivery experiment that would settle it.

The same discussion records something the paper reports but does not resolve: vestibular rescue was durable to 12 weeks while **auditory thresholds deteriorated over the same period despite robust transgene expression and unchanged hair cell morphology.** Something is failing in the treated cochlea that is neither loss of transgene nor visible cell loss.

## A hedge I deliberately preserved

The founding 2015 paper described CLIC5 as involved in hearing, vestibular and **possibly mild renal** dysfunction — one patient had mild renal dysfunction. Later descriptions drop the qualifier and call DFNB103 nonsyndromic flat. CLIC5 has an established role in glomerular podocyte maintenance, so the observation points at a mechanism rather than being noise. A `KNOWLEDGE_GAP` keeps the question open and names the cheap study that would close it: the Siberian founder cohort is now 22 genotypically identical patients.

## A near-miss worth reporting

An initial evidence item for PMID:22889583 quoted a **paraphrase of the paper title** rather than a real substring. The reference validator caught it, and it was replaced with a verified sentence from the abstract. Flagging it because it is exactly the failure mode the evidence SOP warns about, and the gate did its job.

## Other curation calls

- **Prevalence.** Two population records from Pshennikova 2019: Yakutia-wide 2.7 per 100,000, and 313.9 per 100,000 in the Eveno-Bytantaysky district. Separately, the Siberian founder yield (9.9% of GJB2-negative patients) is recorded with `measure_type: UNKNOWN` and no rate slots filled, because it is a diagnostic yield in an ascertained, GJB2-excluded series rather than a population rate — and it does not generalise, since the same variant was absent from GJB2-negative patients of predominantly Russian descent.
- **No amplification, implantation or vestibular rehabilitation under `treatments`.** All three are standard in progressive severe-to-profound loss with vestibular failure, and no DFNB103-specific report documents any of them. Scoping decision, stated in the treatment `notes`. Gene replacement is likewise not curated as a treatment — it is preclinical and lives in `animal_models` where its limitations can sit next to it.

## Deep research

Falcon report run and preflighted against MONDO:0014469: **PASS** (CLIC5 = 36 mentions, next gene at 8). Entry curated from primary sources regardless, every snippet re-verified against the cache.

## Validation

- `just validate` — 0
- `just validate-terms` — 0
- `just validate-disorders` — 0, 56/56 snippets verified
- `just validate-history` — 0

Closes #9905


kevinschaper:
Three small commits since this opened, all from self-review rather than from a reviewer, plus a rebase.

**A readout direction that contradicted its own snippet.** The untreated-null ABR readout said `DECREASED` while the sentence supporting it reports thresholds shifted by about 40 dB. A threshold that rises is `INCREASED`. Flipped, and the interpretation now says so explicitly so the name and the direction cannot drift apart again. Found by applying the review of the sibling DFNB120 PR in this batch to this entry.

Also renamed the post-treatment structural readout from "Stereocilia fusion and elongation after treatment" to "Stereocilia morphology after treatment", since `RESTORED` describes the morphology returning to normal rather than the fusion doing anything.

**Two overstatements in the opening paragraph.** Vestibular involvement was described as "usually" present; the 2025 gene-therapy paper says it is observed in some individuals, not all, which is also why the phenotype is graded `FREQUENT` rather than `OBLIGATE`. And the loss was said to reach severe or profound "before or during the second decade", which is the founding family's course — the 22-patient Siberian cohort reaches profound in the second *or third* decade.

**A cross-entry claim.** An explanation said asymmetry "is the rule" in DFNA7. The DFNA7 entry in this batch now grades asymmetry `FREQUENT`, not `VERY_FREQUENT`, after its founding Dutch families turned out to be mostly symmetric. Reworded so the two entries agree.

**And a rebase.** `check-snippet-grading` began failing locally on `Metastatic_Ovarian_Cancer.yaml`, a file this branch never touches — `origin/main` had moved and the `--against-ref` baseline no longer matched. It passes after rebasing, as do the other four baseline gates, `check-duplicate-keys` and `validate-disorders`. The cache diff is still only the three rows this entry introduces.

kevinschaper:
All eight items are addressed. The two that put wrong facts into the knowledge base — the truncated day window and the prevalence number — are fixed, and so are the three explanations that blamed the reference cache for something it does not do.

**Item 1, the P17 truncations.** You are right and I was wrong twice: once when I truncated the quotes, and again when I wrote an explanation asserting the cache splits the day number. I re-checked `references_cache/PMID_40859056.md` myself before editing. Both sentences are present unbroken. Both snippets now run to `P17` and both apologetic sentences are gone.

**Item 2 and item 4, the prevalence record.** `rate_per_100000: 9900.0` and `prevalence_class` are removed, and `measure_type` is now `UNKNOWN`. The schema defines carrier frequency as the heterozygote rate and these are homozygous affected patients, so no enum value fits and `UNKNOWN` is the honest one. The 9.9% stays in `notes` and in the snippet, where it is accurate.

**Item 6, the population prevalence.** `just fetch-reference` cached the Pshennikova 2019 DOI, and it came back as full text including an English abstract rather than abstract-only, so the figures are directly quotable. Two `POINT_PREVALENCE` records now exist: Yakutia-wide at 2.7 per 100,000, and Eveno-Bytantaysky district at 313.9 per 100,000 (about 1 in 318). The genetic notes now record that `c.1121G>A p.Trp374*` and `c.644G>A p.(Trp215*)` are the same allele on different transcripts.

**Item 3, the other two truncated snippets.** The Cys32Ter sentence now appears in full in both places it was needed. For the 47.4 Mb sentence I took your advice and quoted a different sentence rather than reproducing the U+2009 thin space in YAML.

**Item 5, the evidence grading.** Four items citing PMID:40859056 are now `OTHER` rather than `HUMAN_CLINICAL`, with the reason written into each explanation: it is a mouse AAV study and the quoted sentence is its introduction restating findings from Seco 2015 and Wonkam-Tingang 2020. The same applies to the PMID:24781754 item quoting that human paper's citation of prior jitterbug work. I regraded rather than swapping the citation because the sentence states the two-stage natural history — hearing loss first, vestibular areflexia later — more compactly than either primary source does.

One item you did not flag that has the same defect: the `Vestibular Areflexia` `PARTIAL` item citing PMID:40859056 for "observed in some individuals and a related mouse model". Regraded to `OTHER` as well.

**Item 8, the postlingual frequency.** Now `FREQUENT`, with the Pshennikova sentence giving the 19-of-26 denominator cited alongside it.

**Suggestions.** The ambiguous `animal_models#Mouse` now names the AAV-treated model. A three-phase `progression` section is added. GeneReviews is closed out: PubMed and Bookshelf `esearch` both return zero for CLIC5 and for DFNB103, so no chapter exists.

**One suggestion I implemented differently.** You asked for a `mappings` block carrying OMIM 616042. `DiseaseMappings` has no OMIM slot — it carries only ICD-10-CM, ICD-11, MONDO and NCIT — so OMIM went into `external_assertions` instead, which is where `Bleeding_Disorder_Platelet-type_21` puts it and says so in its own description.

Validation, all exit 0 after rebasing onto current `main`: `just validate` (60/63 snippets verified, 3 skipped because the validator skips the `DOI` prefix), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, and the five baseline gates. The three skipped snippets are the Pshennikova ones; I checked each as an exact substring of the cache file by hand.

kevinschaper:
Both blocking items and all six suggestions are addressed.

**Finding 1a, the missing inner hair cells.** You are right, and the node's own cited sentence says so plainly: IHC bundle fusion occurs across all cochlear regions while OHC fusion is worst apically, so IHCs are the *more* affected population and they were the ones left out. `CL:0000589` added, and that sentence is now its own evidence item rather than sitting unquoted in a paper the node already cites.

**Finding 1b, the GO term.** I went with the more specific option rather than the parent: `GO:0060121` (vestibular receptor cell stereocilium organization) alongside `GO:0060088`. Both arms have direct evidence on this node — the OHC/IHC SEM findings and the P15/P17 utricular findings — so two precise terms carry more than one general one. `GO:0060122` stays on the upstream tethering node, where the claim genuinely is the undifferentiated one.

**Finding 2, the `HUMAN_CLINICAL` gradings.** Both regraded to `OTHER`. The sentence is that paper's introductory summary of what CLIC5 was already known to do, not an observation from the Cameroonian family — I checked the surrounding paragraph in the cache to be sure. Your point about the renal gap is the sharper one: grading it clinical implied human podocyte data that the gap itself says nobody has, so the explanation there now says why it is not clinical evidence.

**Suggestion 3, the self-contradictory description.** Fixed by naming the lines. The description now says jitterbug fusion from P10, and separately that the Clic5-null line shows nothing at P14 with defects established by P17, and states which line the P14–P17 window belongs to, since both the `HUMAN_MODEL_MISMATCH` discussion and the AAV `limitations` rest on it.

**Suggestion 4, the unsupported district rate.** Fixed — the sentence carrying 31.39 per 10,000 is now quoted on that record too. It was a real gap: the record asserted a number and quoted only ethnic composition.

**Suggestion 5, the renal gap anchor.** Agreed it was arbitrary. Rather than dropping to `genetic#CLIC5` alone I retargeted it to `disease#Autosomal Recessive Nonsyndromic Hearing Loss 103`, because the question the gap asks is whether the disease is nonsyndromic, which is a claim about the entry as a whole.

**Suggestion 6, MYO6.** Added as its own evidence item. The reciprocal dependence was asserted in the description with nothing behind it.

**Suggestion 7, bilaterality.** Added, `HP:0008619`, `VERY_FREQUENT` rather than `OBLIGATE`: three of three Cameroonian patients are bilateral and the Siberian cohort is symmetrical, but the Turkish abstract does not state laterality, so what is established is that no unilateral case has been described.

**Suggestion 8** — PR body updated.

Validation, all exit 0 after rebasing onto current `main`: `just validate` (64/68 snippets; the 4 skipped are DOI-prefixed and I checked each as an exact substring by hand), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, the five baseline gates, and `pytest tests/test_data.py -k "entity_ref or model_readout or experimental_model or animal_model or phenotype or prevalence"` (15,438 passed). The `cache/` diff gains one additive row, `GO:0060121`, and has no deletions.

PR #9924, opened 2026-08-28
curate: Add Primary Ciliary Dyskinesia 30 (MONDO:0014465, ODAD3/CCDC151)

### LINKED ISSUE #9902: Curate primary ciliary dyskinesia 30 (MONDO:0014465)
Curate a dismech entry for **primary ciliary dyskinesia 30** ([MONDO:0014465](https://monarchinitiative.org/MONDO:0014465)) — biallelic *ODAD3* (CCDC151) variants disrupt the outer dynein arm docking complex, so outer dynein arms fail to attach to the axoneme.

**Read `kb/disorders/Primary_Ciliary_Dyskinesia.yaml` first.** That entry is curated at the umbrella `primary ciliary dyskinesia` term, has no `has_subtypes` section, and already cites ODAD3 mechanistically (the Lrrc56/Odad3 axonemal dynein deployment work). So the real question here is whether PCD30 is a separate Disease entry or a `has_subtypes` entry on that file — decide it explicitly rather than defaulting.

Stub: `stubs/Primary_Ciliary_Dyskinesia_30.yaml`

kevinschaper:
Adds a new disorder entry for **primary ciliary dyskinesia 30** — the form of PCD caused by biallelic loss of *ODAD3*, the gene formerly called *CCDC151*.

The mechanism is unusually clean. ODAD3 is part of the outer dynein arm docking complex, the structure that anchors preassembled dynein motors onto the axonemal microtubules. Losing it does not damage the motors; it removes the place they attach. The result is a fully built cilium with nothing to power it.

Closes #9902

## entry_type decision: DISEASE, and the case against it

The issue asked for this to be decided explicitly rather than defaulted, so here is both sides.

**The case against a standalone entry is real.** `kb/disorders/Primary_Ciliary_Dyskinesia.yaml` has no `has_subtypes` section and curates ODAD1 (CCDC114) — the paralogous member of *the same docking complex* — as a causal gene in its `genetic:` block. Giving ODAD3 its own file therefore splits two subunits of one structure across two levels of the knowledge base. That is a genuine inconsistency and the entry's `notes:` says so.

**I chose standalone anyway, for a structural reason.** The `Subtype` class has these slots and no others: `name`, `display_name`, `subtype_term`, `mappings`, `description`, `evidence`, `review_notes`, `locations`, `geography`, `classification`, `children`, `genes`, `subtype_frequency`, `inheritance`. There is no `pathophysiology`, no `phenotypes`, no `animal_models`. So a subtype entry could not have carried any of what makes this entry worth having — the four-node docking-complex causal chain, the three model organisms with their fidelity gradings, or the two discussions. It would have been a gene name and a paragraph.

`Primary_Ciliary_Dyskinesia_47_and_Lissencephaly` is the existing precedent for a numbered PCD form as its own file. The right way to fix the ODAD1/ODAD3 asymmetry later is to promote ODAD1, not to demote ODAD3.

**No duplication with the umbrella entry.** It does mention ODAD3, but inside an animal-model block about *Xenopus* Lrrc56 knockdown, where axonemal Odad3 signal is the readout for a *different* gene's defect. That is LRRC56 biology observed through ODAD3, and it is not repeated here.

## Where the interesting content is

Two mouse findings that would have had nowhere to live on the umbrella entry:

**Mouse hydrocephalus, curated as a `HUMAN_MODEL_MISMATCH`.** The *Ccdc151* null dies perinatally of communicating hydrocephalus, with the gene expressed in ventricular ependyma. Human ODAD3 patients live to at least 35. I drafted this discussion asserting flatly that human ODAD3 disease does not cause hydrocephalus, then found that PMID:32490514 lists hydrocephalus among the recognised consequences of ciliary dysfunction in PCD generally. The discussion was rewritten: the mismatch is one of **frequency and severity, not of kind** — occasional and survivable in human PCD, unreported in this genotype, invariably lethal in the mouse — and it now carries that paper as `PARTIAL` evidence cutting both ways.

**A heterozygote effect, curated as an `OPEN_QUESTION`.** *Odad3*+/− mice have reduced sperm count and motility and a shorter fertile lifespan, and the authors say it may bear on counselling for carriers. No human study has looked, though the cohort trivially exists — fathers of probands are genotyped as a matter of course. The entry curates **no** carrier phenotype and states that recessive counselling should not change; the finding is recorded so a future carrier study is recognised as testing a prediction rather than making an incidental observation.

- **No `treatments:` and no `progression:` block.** Airway clearance, antibiotics and surveillance would certainly be given to an ODAD3 patient, but no study has examined any of it in this genotype. The class-level content belongs on the umbrella entry and is correct there.
- **Hydrocephalus is not a phenotype of this entry.** It lives on the mouse model link with the lethality stated as a limitation.
- **`HP:0012256` "Absent outer dynein arms", not `HP:0012259` "Absent inner and outer dynein arms".** Inner arms are preserved in this disease, and that selectivity is what separates a docking-complex defect from a dynein-preassembly defect on electron microscopy.

## Evidence notes

- PMID:32286033 was fetched and dropped. Its cache entry is `content_type: unavailable`, so nothing from it could be quoted; the cache file is not committed.
- One snippet had to be trimmed: PMID:25224326's variant sentence writes the protein change as `p.[E309*]`, and the reference validator strips bracketed spans before matching. The snippet stops before the HGVS string and the allele is recorded in the genetic notes instead.
- The TTC25 differential quotes the sentence where a *different* docking-complex gene's defect removes CCDC151 from the axoneme along with the rest of the complex — which is exactly why immunofluorescence cannot tell these four genes apart.

## Deep research

Provider: falcon. NEC preflight is a **WARN** and I read it: ODAD3 appears 63 times against DNAH5 at 4, so the report is about the right disease. The only warning is that it cites the umbrella PCD OMIM 244400 rather than 616037 for CILD30. Report committed.

## Validation

| check | exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just validate-disorders` | 0 (41/41 snippets verified) |
| `just check-folded-hyphens` | 0 (caught two real defects, fixed) |

### DIFF
--- a/history/disorders/Primary_Ciliary_Dyskinesia_30/2026-08-29T001224Z-claude-code-267788.yaml
+++ b/history/disorders/Primary_Ciliary_Dyskinesia_30/2026-08-29T001224Z-claude-code-267788.yaml
@@ -1,0 +1,42 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Primary_Ciliary_Dyskinesia_30
+  path: kb/disorders/Primary_Ciliary_Dyskinesia_30.yaml
+session:
+  id: 2026-08-29T001224Z-claude-code-267788
+  timestamp: '2026-08-29T00:12:24Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9902
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9924
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - notes
+  summary: 'Fifth review response: apply the Orphanet band rule symmetrically'
+  details: |
+    Fifth review response on PR #9924, addressing the 00:02:49Z CHANGES_REQUESTED review.
+
+    Context worth recording: two review runs examined commit 54f2ab04 concurrently. The 00:02:49Z run found a new defect; the 00:07:56Z run approved, and its own text shows it was evaluating the commit against the 22:53:51Z review rather than against the 00:02 finding, which it never saw. The approval is therefore not an informed reversal of the 00:02 finding. The finding was verified independently before acting on it and it holds, so the fix ships even though it dismisses a standing approval.
+
+    The defect: two frequency values were set to an Orphanet class band that the entry's own text says is too low. Bronchiectasis carried OCCASIONAL while the explanation on the very evidence item supplying it said the band "sits lower than the individually described ODAD3 patients would suggest" - and both described ODAD3 patients have bronchiectasis. Situs Inversus carried OCCASIONAL while the entry's description says roughly half of PCD patients have normal situs, which puts laterality defects in the FREQUENT band.
+
+    Verified against the umbrella entry rather than taking the review's word: kb/disorders/Primary_Ciliary_Dyskinesia.yaml quotes these two identical Orphanet rows and handles both the other way - frequency FREQUENT with the ORPHA row graded supports PARTIAL and an explanation that the band understates. So the child entry was contradicting the parent on the same source row while holding the stronger gene-specific evidence.
+
+    Both are now FREQUENT with their ORPHA rows graded PARTIAL and the disagreement stated, matching the umbrella.
+
+    The deeper point in the review was the asymmetry, and it is the part worth fixing structurally: the class band had been allowed to raise male infertility to FREQUENT on no human ODAD3 data at all, while also lowering bronchiectasis below what every described ODAD3 patient shows. That is one rule applied in one direction only. The notes block now states the rule explicitly - an ORPHA row supplies frequency and is graded SUPPORT where the entry has no gene-specific evidence bearing on it; where gene-specific evidence contradicts the band, the gene-specific reading sets frequency and the row is graded PARTIAL with the disagreement stated - and works through all three cases so the next curator applies it symmetrically.
+
+    Non-blocking suggestions taken. Recurrent Respiratory Infections was the only phenotype with no frequency and now carries FREQUENT from HP:0005425 recurrent sinopulmonary infections, graded PARTIAL with the term mismatch stated explicitly, since that row is narrower than the HP:0002205 term bound here. Hearing Impairment gains the HP:0000405 conductive row as corroboration of both its band and the conductive mechanism its description gives, without changing the decision to bind the general term. Situs Inversus recategorised from Developmental to Cardiovascular so the two laterality manifestations sit together.
+
+    Validation: just validate exit 0; just validate-terms exit 0; just validate-disorders exit 0 with 71/71 snippets verified, up from 69/69; check-entity-refs, check-duplicate-keys, check-folded-hyphens, check-snippet-length, check-title-snippets, check-snippet-grading and check-environmental-evidence all exit 0. Render test passes. No cache diff.
--- a/kb/disorders/Primary_Ciliary_Dyskinesia_30.yaml
+++ b/kb/disorders/Primary_Ciliary_Dyskinesia_30.yaml
@@ -603,7 +603,7 @@ phenotypes:
       Cross-species reproduction of the same coupled defect.
 - category: Respiratory
   name: Bronchiectasis
-  frequency: OCCASIONAL
+  frequency: FREQUENT
   description: >-
     Diffuse bronchiectasis, the irreversible structural consequence of chronic
     retained infection. Documented in the individually described 35-year-old
@@ -625,16 +625,20 @@ phenotypes:
       Documents bronchiectasis in a reported ODAD3 patient.
   - reference: ORPHA:244
     reference_title: "Primary ciliary dyskinesia"
-    supports: SUPPORT
+    supports: PARTIAL
     evidence_source: OTHER
     snippet: "HP:0002110 | Bronchiectasis | Occasional (29-5%)"
     explanation: >-
-      Orphanet grades bronchiectasis Occasional across primary ciliary
-      dyskinesia as a whole. That band is the class's, not ODAD3's - it is
-      recorded because it is the only graded frequency available, and it sits
-      lower than the individually described ODAD3 patients would suggest.
+      Graded PARTIAL because the band understates this phenotype rather than
+      measuring it: Orphanet's Occasional is a class figure that most likely
+      reflects cross-sectional point prevalence rather than cumulative
+      incidence, and both individually described ODAD3 patients have
+      bronchiectasis. `frequency` is set to FREQUENT on that gene-specific
+      evidence, matching how the umbrella `Primary Ciliary Dyskinesia` entry
+      handles this same Orphanet row.
 - category: Respiratory
   name: Recurrent Respiratory Infections
+  frequency: FREQUENT
   description: >-
     Upper and lower airway infection from infancy onwards - the consequence of
     losing mucociliary clearance, and the burden that eventually produces the
@@ -667,6 +671,17 @@ phenotypes:
       chest infections
     explanation: >-
       The same picture in an independently reported CCDC151 patient.
+  - reference: ORPHA:244
+    reference_title: "Primary ciliary dyskinesia"
+    supports: PARTIAL
+    evidence_source: OTHER
+    snippet: "HP:0005425 | Recurrent sinopulmonary infections | Frequent (79-30%)"
+    explanation: >-
+      Graded PARTIAL because the row is not this phenotype's own term. Orphanet
+      grades HP:0005425 recurrent sinopulmonary infections, which is narrower
+      than the HP:0002205 recurrent respiratory infections bound here; it is the
+      nearest graded row in the table and is what the FREQUENT band rests on.
+      Recorded explicitly so the mismatch is visible rather than implied.
 - category: Respiratory
   name: Productive Cough
   frequency: FREQUENT
@@ -776,6 +791,17 @@ phenotypes:
       Independent corroboration of the OCCASIONAL band already assigned here,
       which was reasoned from a narrative hedge rather than from a graded
       source. Orphanet reaches the same band for the class.
+  - reference: ORPHA:244
+    reference_title: "Primary ciliary dyskinesia"
+    supports: SUPPORT
+    evidence_source: OTHER
+    snippet: "HP:0000405 | Conductive hearing impairment | Occasional (29-5%)"
+    explanation: >-
+      Orphanet grades the conductive subtype at the same band, which supports
+      the mechanism this phenotype's description gives - middle-ear disease
+      following from the same ciliated epithelium failing in the Eustachian
+      tube - without changing the decision to bind the general HP:0000365 term,
+      since the sources describing ODAD3 patients say only "hearing disease".
 - category: Respiratory
   name: Chronic Sinusitis
   frequency: FREQUENT
@@ -824,9 +850,9 @@ phenotypes:
       bronchiectasis, dextrocardia and infertility.
     explanation: >-
       Documents dextrocardia in a reported ODAD3 patient.
-- category: Developmental
+- category: Cardiovascular
   name: Situs Inversus
-  frequency: OCCASIONAL
+  frequency: FREQUENT
   description: >-
     Situs inversus is one outcome of randomised laterality, reported in an
     independent Chinese case and in the Kartagener-syndrome family. It is not
@@ -850,14 +876,18 @@ phenotypes:
       An ODAD3 patient with situs inversus.
   - reference: ORPHA:244
     reference_title: "Primary ciliary dyskinesia"
-    supports: SUPPORT
+    supports: PARTIAL
     evidence_source: OTHER
     snippet: "HP:0001696 | Situs inversus totalis | Occasional (29-5%)"
     explanation: >-
-      Orphanet grades complete situs inversus Occasional for the class. Worth
-      having because it quantifies the point this entry makes in prose - that
-      laterality is randomised rather than reversed, so full situs inversus is
-      one outcome among several rather than the expected one.
+      Graded PARTIAL because the band sits below what this entry itself states
+      and what the literature reports - roughly half of PCD patients have a
+      laterality defect, and situs inversus totalis is the commonest form of it.
+      `frequency` is set to FREQUENT accordingly, matching how the umbrella
+      `Primary Ciliary Dyskinesia` entry handles this same Orphanet row. The row
+      is still worth quoting because it is the only graded source available and
+      because a reader should see that the graded value and the narrative value
+      disagree.
 - category: Otological
   name: Recurrent Otitis Media
   frequency: FREQUENT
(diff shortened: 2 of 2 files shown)


kevinschaper:
Both blocking items fixed, all four suggestions taken, and your review indirectly caught a worse defect than either of the items you raised.

**The defect you found sideways.** You noted that the "four of five" laterality figure is not in the abstract-only cache. It isn't — it comes from the deep-research report's reading of the full text — and I had written it **twice**, once in the description and once in a pathophysiology node, in both cases as if it were sourced. It now appears once, explicitly labelled as an uncited lead. The description's "neonatal respiratory distress" claim had exactly the same provenance and is gone for the same reason. Thank you for that one; it was the kind of thing no validator catches.

**IMPORTANT 1 — Female Infertility.** Correct, and the fix is the sentence you pointed at. The novelty statement is now described in the `explanation` rather than quoted as evidence.

**IMPORTANT 2 — four phenotypes added**, all from verbatim sentences in caches this PR already commits:

| phenotype | HP | source |
|---|---|---|
| Recurrent respiratory infections | `HP:0002205` | two independent patients, PMID:33719352 and PMID:32490514 |
| Productive cough | `HP:0031245` | PMID:32490514, presenting complaint |
| Ventricular septal defect | `HP:0001629` | PMID:32490514 gene-specific summary |
| Hearing impairment | `HP:0000365` | same sentence |

The last two come from a summary of the CCDC151 phenotype range rather than an individually described patient, so both are graded `OTHER` with that stated in the explanation.

**Suggestions, all taken:** dropped `GO:0005515` (the co-immunoprecipitation evidence still carries the interaction claim); corrected `evidence_source` on "we found Ccdc151 expressed in vertebrate left-right organizers" from `HUMAN_CLINICAL` to `MODEL_ORGANISM`; added the defining series' own `c.925G>T p.(Glu309*)` and `c.1256C>A p.(Ser419*)` to the genetic notes as uncited leads, including the deep-research report's warning that some secondary summaries render the second as C>T where the primary text gives C>A; and tagged GeneReviews `PMID:20301301` with a note on why it is not cited.

**Two additions beyond the review.** `ORPHA:244` now supplies a prevalence evidence row — worth having because Orphanet holds *one* epidemiological record for the whole disease and lists 53 causal genes in it, ODAD3 among them, which bears directly on the lump/split argument. And the `notes:` now address the scaling question raised on #9902: there are 49 PCD stubs, all `UNDECIDED`, so this file's shape is likely to be copied dozens of times. I recorded both counter-arguments honestly — Orphanet does not split by gene, and the umbrella's own `genotype_ultrastructure_severity` hypothesis already reasons along an ultrastructural-defect axis that would give all 49 a home — and said the class-level policy call belongs to a maintainer with the whole queue in view. If that ruling goes the other way, this entry's content transfers to an ODA-defect subtype and nothing is lost but the file.

Re-ran everything: `validate`, `validate-terms`, `validate-disorders` (47/47 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Branch rebased; cache diff is empty.

kevinschaper:
Went back through the detailed comment and picked up the parts the summary did not spell out.

**Frequency qualifiers on the two summary-sourced phenotypes.** You flagged the source's own hedge — "varied phenotypes… which may be accounted by background genotype effects, epigenetic modifications and environmental factors" — and you were right that it needs carrying. Ventricular septal defect and hearing impairment now both take `frequency: OCCASIONAL`, with the reasoning written into their descriptions: the sentence declines to give a count, so anything higher would read a frequency into it that isn't there.

Hearing impairment also now records why it binds the general term rather than a conductive-specific one. The mechanism predicts conductive loss, but the source says "hearing disease", and the more specific term would assert an audiological characterisation nobody reported.

**`evidence_source` harmonised.** Both background statements are now `OTHER` — PMID:33719352's protein-architecture sentence and PMID:31383820's gene-function sentence. Both are introduction prose, and that is how the entry grades that kind of sentence elsewhere.

On the GeneReviews point: I added the tag anyway, with a note saying it is tag-only because the cached record is a purpose statement. Your assessment that there is nothing to mine is right — the tag is there so the absence reads as a decision rather than an oversight, which is the same reason I gave for it on #9919.

Also, thank you for the independent snippet verification and for the `HP:0012256` note. Getting the inner-versus-outer arm distinction right was the point of choosing it, and it is useful to know an independent check agreed.

Everything re-run after this push: `validate`, `validate-terms`, `validate-disorders` (47/47 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0.

kevinschaper:
All three findings fixed, all three suggestions taken. This was a much sharper pass than the first one and every finding was correct.

**Finding 1 — nasal nitric oxide.** You put your finger on the actual problem, which was not that a test was missing but that the omission contradicted the entry's own standard. The entry repeatedly declines class-level material on the grounds that it has not been studied in this genotype — and nNO *has* been, at 6 nl/min against a 77 cut-off. Keeping weaker class-level statements while dropping the strongest genotype-specific datum was inconsistent.

Added both ways you offered: a `biochemical:` marker matching the umbrella entry's shape, and a `diagnosis:` entry placed ahead of the ultrastructural one, since nNO is the screen that decides who gets an electron microscope. Both state that it is one patient and that a low value triages rather than diagnoses.

**Finding 2 — the allelic spectrum.** The claim was wrong as written and the correction is in a cache this PR commits. The node now says loss-of-function rather than "nonsense or frameshift", gives the three/three/one breakdown, and cites the seven-allele sentence.

I also carried your adjacent point rather than just softening the sentence: the splice-donor allele is now discussed as the nearest thing to a non-truncating variant and *why it still isn't one* — donor-site loss generally yields exon skipping or intron retention with a frameshifted, decay-prone transcript, landing back in the same functional class rather than giving a graded phenotype. So "no worked example to calibrate a new missense variant against" survives, but as a reasoned claim rather than an assumed one.

On the protein-change discrepancy: extended. The genetic notes now record **two independent** transcription-discrepancy classes for the Hjeij alleles — nucleotide (C>A vs C>T) and protein (`p.(Glu309*)`/`p.(Ser419*)` vs `p.(Gly309*)`/`p.(Ser409*)`) — with the point that neither is resolvable from the cached records because the primary paper is abstract-only, so the variant should come from the primary full text and not from any review.

**Finding 3 — the bundled node.** Split into three chained nodes: `Mucociliary Clearance Failure` (TISSUE, effector), `Chronic Airway Infection and Neutrophilic Inflammation` (TISSUE, **amplifier** — each infection-inflammation cycle degrades clearance further, which is why the disease is progressive rather than static), and `Bronchiectasis and Airway Wall Destruction` (ORGANISM, outcome). The composite's pooled evidence was redistributed so each node carries what actually supports it, the `KNOWLEDGE_GAP` `attaches_to` is repointed to the terminal node, and I verified the pathograph is still one connected component. `phenotypes#Bronchiectasis` now has a terminal mechanism node to sit under, as you said it would.

**Suggestions, all taken:** the non-genetic clinical differential (cystic fibrosis, primary immunodeficiency, aspiration pneumonia), with the addition that laterality is what redirects the workup and that its *absence* excludes nothing — about half of PCD patients have normal situs, and this disease's MONDO term is explicitly "with or without situs inversus". The OMIM identifier MIM 616037 is in `notes:` rather than `mappings:`, because `DiseaseMappings` has only ICD-10-CM, ICD-11, MONDO and NCIT slots — no OMIM one — and the note says plainly that the report's use of the umbrella 244400 is the error. Neonatal respiratory distress is recorded as searched-and-not-quotable with the reason, so the next curator does not repeat it.

Gates: `validate`, `validate-terms`, `validate-disorders` (53/53 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Render test passes. No cache diff.

kevinschaper:
Fixed, and you were right to make it blocking.

**The OMIM paragraph was my own false claim about the repository.** I wrote that MIM 616037 had to live in prose because `DiseaseMappings` has no OMIM slot. The first half is true; the conclusion was not, and I asserted it without checking whether somewhere else in the schema took it. `external_assertions` does, `ExternalAssertion` requires only `source` and `external_id` and accepts `evidence`, and **168 disorder files** already use it — I verified all three against the schema and against the exemplar you named.

There is now an `external_assertions` block carrying `OMIM:616037` with `assertion_type: disease_record` and the quotable sentence tying that number to ODAD3 disease specifically. The `notes:` paragraph is trimmed to what survives: `mappings:` is not the place, `external_assertions` is, and the deep-research report's use of the umbrella 244400 is the error worth recording.

Your reason for blocking it is the part worth acknowledging. The paragraph was a statement *about the repository*, so a curator reading it would have concluded there is nowhere to put an OMIM identifier and repeated the workaround. That is the same defect class as two overreaches I self-corrected on the sibling FAR1 branch — an assertion about something outside the entry, made without checking the thing. A false claim about the schema is worse than the others, because it propagates as a convention.

**Both suggestions taken:**

- `presence: Decreased` on the nNO marker, matching the umbrella entry's wording for the same analyte.
- The 77 nl/min cut-off is now structural: a one-sided `reference_range` with `lower_bound: 77.0` and no `upper_bound`, plus two `interpretation_bands` — below the cut-off flagged `LOW`, interpreted as supporting referral for ultrastructural and genetic testing without diagnosing PCD or naming a gene; at-or-above flagged `NORMAL`, with the caveat that it does not exclude PCD since some genotypes (RSPH1) retain higher values. The range's `notes` record why `upper_bound` and `loinc_term` are deliberately unset: it is a diagnostic threshold rather than a laboratory interval, and nasal nitric oxide has no LOINC code in this repository's caches, so inventing one would be worse than omitting it.

**Also caught while I was in there:** `genetic_context.description` still said the alleles are "all truncating" — the residue of the sentence finding 2 corrected two lines above last round. It now carries the actual three-nonsense / three-frameshift / one-splice-donor split, and the genetic notes heading changed from "all truncating" to "all loss-of-function" for the same reason. Thank you for spotting that one; it is exactly the case where fixing a claim in one place and missing its restatement elsewhere is the whole failure mode.

Gates: `validate`, `validate-terms`, `validate-disorders` (55/55 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Render test passes. No cache diff.

kevinschaper:
Both findings hold up and both are fixed. I checked every claim before acting: all five ORPHA rows you quoted are exact substrings of `references_cache/ORPHA_244.md`, the file is tracked, and this entry already cited it twice. **I have no disagreement to register on this one** — you found a real gap and diagnosed the cause correctly.

**The root cause is the diagnosis worth keeping:** I cited `ORPHA:244` for a prevalence figure and never read the 46-row phenotype table in the same file.

**Finding 1 — neonatal respiratory distress.** The `notes:` claim was wrong in exactly the way you describe. The premise was true (the 2014 paper's cache is `abstract_only` and carries nothing), but the conclusion — that journal access was the only route — was not, and as written it would have sent the next curator to a paywall for a row already in the repository. Curated as `HP:0002643`, `frequency: FREQUENT` from the Orphanet row, `onset_category: NEONATAL`, with the description keeping the class-level band that carries it separate from the uncited four-of-five lead.

**Finding 2 — male infertility.** Your argument that the entry applied a stricter standard here than to nasal polyposis is correct, and the frequency inversion makes it sharper than a consistency point: Orphanet grades male infertility **Frequent** and female **Occasional**, so the entry carried the rarer of the pair and omitted the commoner.

Added `HP:0003251` Male Infertility and, separately, `HP:0012206` Abnormal Sperm Motility. I kept them as two rather than one because the models this entry already cites separate them: adult *Odad3* deletion gives asthenoteratozoospermia with abnormal motility, while the constitutive null has **no sperm in the reproductive tract at all**. Motility failure and infertility come apart depending on when the gene is lost, and collapsing them would lose that. Each carries the Orphanet band plus the gene-specific mouse evidence, with the description stating plainly that no ODAD3 male patient's fertility has been reported.

**On the pattern you flagged.** You are right that this was the second instance of one shape — a section excluded on a confident claim about what the repository could supply, where I checked the premise and not the conclusion. The OMIM paragraph was the first. Rather than delete the bad sentence, the `notes:` block now records the mistake, names both instances, and ends with the rule they imply: before writing that something cannot be curated, check the caches the entry already cites. That seemed more useful to the next curator than a clean file.

**Suggestions, all taken.** Eight already-curated phenotypes gained their Orphanet band and a corroborating row — bronchiectasis, productive cough, chronic sinusitis, recurrent otitis media, nasal polyposis, situs inversus totalis, female infertility, and hearing impairment, the last independently confirming the `OCCASIONAL` band I had reasoned to from a narrative hedge. The hydrocephalus discussion now carries `HP:0000238 | Hydrocephalus | Very rare (<4-1%)`, which turns its frequency-not-kind argument from an impression into a number: under 4% in humans as a class, against the cause of death in the mouse null. Recurrent Otitis Media recategorised `Respiratory` → `Otological`.

Left alone: the stale `GO:0005515` enum-cache row, per your own assessment.

Snippets went 55/55 → **69/69**, all verified. `validate`, `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Render passes. No cache diff.

kevinschaper:
**I pushed after an approval, deliberately, and want to be explicit about why** — because it dismissed that approval and the reasoning should be checkable rather than taken on trust.

Two review runs examined `54f2ab04` concurrently. The 00:02:49Z run found a new defect. The 00:07:56Z run approved — and its own text shows it was evaluating this commit against the **22:53:51Z** review ("post-dates the previous CHANGES_REQUESTED review (23:53:11Z vs 22:53:51Z)"), so it never saw the 00:02 finding. That makes the approval a parallel verdict at a different depth, not an informed reversal. I verified the finding independently before acting on it, and it holds, so I shipped the fix rather than banking an approval over a known defect.

**The finding.** Two `frequency:` values were set to an Orphanet class band that the entry's own text says is too low:

- **Bronchiectasis** carried `OCCASIONAL` while the explanation on the very evidence item supplying that band said it *"sits lower than the individually described ODAD3 patients would suggest"* — and both described ODAD3 patients have bronchiectasis. Setting a machine-readable value while writing two lines below that you believe it is wrong is the defect, exactly as stated.
- **Situs Inversus** carried `OCCASIONAL` while the description says roughly half of PCD patients have normal situs.

I checked the umbrella claim rather than taking it: `Primary_Ciliary_Dyskinesia.yaml` quotes these two identical Orphanet rows and handles both the other way — `FREQUENT`, with the row graded `supports: PARTIAL` and an explanation that the band understates. So the child entry was contradicting the parent on the same source row, while holding the stronger gene-specific evidence. Both now match the umbrella.

**The asymmetry was the real defect, and it is now fixed structurally rather than case by case.** You put it precisely: the class band was allowed to *raise* male infertility to `FREQUENT` on no human ODAD3 data at all, and to *lower* bronchiectasis below what every described ODAD3 patient shows — one rule applied in one direction only. The `notes:` block now states the rule and works all three cases through it:

> an ORPHA row supplies `frequency:` and is graded `SUPPORT` where this entry has no gene-specific evidence bearing on it; where gene-specific evidence contradicts the band, the gene-specific reading sets `frequency:` and the row is graded `PARTIAL` with the disagreement stated.

So male infertility keeps `FREQUENT` (no ODAD3 human data; the band is all there is), and bronchiectasis and situs inversus both move to `FREQUENT` against the band (gene-specific and literature evidence contradict it).

**Suggestions, all taken.** `Recurrent Respiratory Infections` was the only phenotype with no band and now carries `FREQUENT` from `HP:0005425`, graded `PARTIAL` with the term mismatch stated outright — that row is *recurrent sinopulmonary infections*, narrower than the `HP:0002205` bound here, and I would rather the mismatch be visible than implied. Hearing Impairment gains the `HP:0000405` conductive row, which corroborates both the band and the conductive mechanism its description gives without disturbing the correct general-term binding. Situs Inversus moves from `Developmental` to `Cardiovascular` so the two laterality manifestations sit together.

Snippets 69/69 → **71/71**. `validate`, `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Render passes. No cache diff.

PR #9925, opened 2026-08-28
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 41 (MONDO:0011994, P2RX2)

### LINKED ISSUE #9904: Curate autosomal dominant nonsyndromic hearing loss 41 (MONDO:0011994)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 41 (DFNA41)** ([MONDO:0011994](https://monarchinitiative.org/MONDO:0011994)) — heterozygous *P2RX2* variants; P2X2 is an ATP-gated cation channel in cochlear hair cells and supporting cells, and the disease has a well-documented noise-exposure interaction.

If the noise-susceptibility interaction is supported, model it with an `environmental:` entry carrying `influences_mechanisms` rather than leaving it as prose — see the "Linking Environmental Factors into the Pathograph" section of `CLAUDE.md`.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41.yaml`

kevinschaper:
Adds `kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41.yaml` for DFNA41 (MONDO:0011994, P2RX2 `hgnc:15459`) and retires its stub.

`entry_type: DISEASE`. One causal gene, no MONDO descendants, MONDO's own definition is gene-scoped, and the reported alleles converge functionally rather than diverging. Nothing in `kb/` previously mentioned P2RX2.

## The mechanism is a protective circuit, not a structural protein

This is what makes DFNA41 worth curating rather than filing as one more dominant deafness locus. P2X2 is an ATP-gated cation channel, not a hair bundle component. Rising sound levels release ATP into the endolymphatic compartment, ATP opens P2X2 on the epithelium lining it, and the resulting conductance turns down cochlear amplifier gain. **That is what a temporary threshold shift is** — a protective adaptation, not early damage.

So DFNA41 is a disorder of a protective mechanism, and its signature follows: hearing declines on its own, and noise makes it worse. Both halves are observed on both sides of the same experiment — null mice exposed early to continuous moderate noise had high-frequency loss as young adults, and among heterozygous family members carrying p.Val60Leu, noise exposure exacerbated high-frequency hearing loss in young adulthood.

## The noise interaction is modelled structurally, not left as prose

Per the issue, an `environmental:` entry bound to `ECTO:8000044` (already cached from `Noise_Induced_Hearing_Loss`) with **two** `influences_mechanisms` links, because they are different claims:

- `EXACERBATES` → the noise-exacerbated high-frequency phenotype. Noise makes the outcome worse.
- `MODULATES` → the purinergic adaptation node. Sound level is the *physiological input* to that mechanism, not an aggravating agent applied to it — a carrier is indistinguishable from a non-carrier on this axis in quiet and diverges under noise. `EXACERBATES` would have been the wrong predicate here.

## Acting on the second opinion posted to #9904

The automated comment flagged ClinGen's **Moderate** classification for P2RX2. I fetched the assertion (`just clingen-rebuild --id CGGV:assertion_c1dae6b6-…`) and it changed three things, one of them a real correction:

1. **A defect in my own draft.** ClinGen reviewed the MELAS family (PMID:25788561) and **did not accept it** — the evidence for pathogenicity was not strong and the variant is present in gnomAD. My draft cited that family as "an unusually clean natural experiment" supporting the gene-disease relationship. Corrected to `PARTIAL`, with ClinGen's reason quoted next to it so the disagreement is visible rather than silently resolved in either direction.
2. **`functional_impact_category` changed from `LOSS_OF_FUNCTION` to `DOMINANT_NEGATIVE`.** The ClinGen narrative reads the coexpression result that way, and it is the better description: the mutant subunit is incorporated into a trimeric channel and degrades that channel's function, which is measured rather than inferred.
3. **A `KNOWLEDGE_GAP` on gene-disease validity.** This entry is mechanistically rich and genetically thin — a null mouse, an allele-matched knock-in, a genome-editing rescue and direct electrophysiology, against three missense variants in four probands. Those are easy to mistake for one another, and the discussion says where the caution actually belongs (variant interpretation, not the mechanism).

## A grading defect the gate caught

`just check-snippet-grading` failed before push. The concluding sentence of PMID:23345450 is a synthesis of human *and* mouse results, and I had it as `HUMAN_CLINICAL` in the environmental and treatment blocks but `MODEL_ORGANISM` on an animal-model link. Fixed by re-quoting the animal-model link with a mouse-specific sentence from PMID:23592720 — not by re-grading a shared quote, which is the move the check exists to prevent.

## Two things I removed or declined

**Removed a `Vestibular Dysfunction` phenotype** from an earlier draft. It was supported only by `MODEL_ORGANISM` evidence, and the evidence policy says model data should not be the sole support for a human phenotype. No vestibular testing is reported for any molecularly confirmed DFNA41 carrier. The finding stays in the knock-in mouse `limitations`, where its human status is stated as untested.

**Declined `conforms_to: sensorineural_hair_cell_loss#…`,** which the issue comment suggested. That module node attributes hair-cell death to oxidative stress and mitochondrial/DNA damage culminating in apoptosis. I have no evidence for any of that in DFNA41, and declaring conformance would assert biological processes this entry cannot support. Happy to revisit if a reviewer disagrees.

**No amplification under `treatments`** — no DFNA41-specific report documents hearing aiding. Cochlear implantation *is* curated: PMID:39258340 reports a satisfactory outcome in the proband of a five-generation Chinese kindred. Genome editing is likewise not curated as a treatment; it is preclinical and lives in `animal_models` with its limitations, including that juvenile intervention broadened the rescued frequency range, so the timing that works best has no counterpart in a disease diagnosed in adults.

## Deep research

Falcon report run and preflighted against MONDO:0011994: **PASS** (P2RX2 = 25 mentions, next gene at 2; OMIM 608224 matched on both sides). Entry curated from primary sources regardless.

## Validation

- `just validate` — 0
- `just validate-terms` — 0
- `just validate-disorders` — 0, 62/62 snippets verified
- `just check-snippet-grading` — 0 (after the fix above)
- `just check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-environmental-evidence` — all 0
- `just validate-history` — 0

## One repo defect found along the way

`just clingen-refresh` fails with a checksum mismatch: ClinGen's `gene_validity.csv` has been updated upstream and `data/clingen/MANIFEST.yaml` still pins the older sha256. The download itself succeeds and `just clingen-rebuild --id …` works against it, which is how the assertion above was generated, but the pinned manifest can no longer be reproduced. Worth a separate issue rather than a fix riding in a curation PR.

Closes #9904

### DIFF
--- a/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41/2026-08-28T235914Z-claude-code-155bed.yaml
+++ b/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41/2026-08-28T235914Z-claude-code-155bed.yaml
@@ -1,0 +1,29 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41
+  path: kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41.yaml
+session:
+  id: 2026-08-28T235914Z-claude-code-155bed
+  timestamp: '2026-08-28T23:59:14Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9925
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - references
+  - phenotypes
+  - experimental_models
+  summary: 'Second review response: GeneReviews baseline, hiPSC model, three phenotype
+    changes'
+  details: |
+    Addressed the second CHANGES_REQUESTED review on PR #9925, which superseded an earlier approval on the same head commit. Added the top-level references block tagging PMID:20301607 Genetic Hearing Loss Overview as GeneReviews, matching X-linked_Nonsyndromic_Hearing_Loss, COL11A2_Hearing_Loss and Auditory_Neuropathy. My earlier check used GeneReviews[TA], which PubMed does not resolve; the correct filter is GeneReviews[Book], and P2RX2 AND GeneReviews[Book] returns 20301607. The cache file is already on main so no fetch was needed, and the chapter is tagged without evidence mining because its cached abstract is a scope statement with no quotable clinical content. Added a Postlingual Sensorineural Hearing Impairment phenotype bound to HP:0008596 at OBLIGATE, citing the ClinGen assertion. Added a Bilateral Sensorineural Hearing Impairment phenotype bound to HP:0008619 at OBLIGATE, citing the ten-year follow-up. Lowered Adult-Onset Hearing Impairment from VERY_FREQUENT to FREQUENT and explained why: one published reading of the founding kindred puts onset in adolescence and the Chinese family had onset earlier still, so adulthood specifically is contested while postlingual onset is not. Added an experimental_models section with the Dong 2019 patient-derived P2RX2 p.Val60Leu hiPSC lines and their CRISPR homozygous isogenic control, fetched as PMID:30819013, linked to the P2X2 Receptor Loss of Function node with relationship PERTURBS and fidelity UNKNOWN because the paper builds the lines without measuring any phenotype in them, with no readouts for the same reason. Kept preferred_term Hyperalgesia on HP:0012534 as the reviewer allowed. Validated with just validate (75/75 snippets), validate-terms, validate-disorders, check-entity-refs, check-duplicate-keys and the five baseline gates, all exit 0; no cache rows changed.
--- a/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41.yaml
+++ b/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_41.yaml
@@ -327,7 +327,7 @@ phenotypes:
     individuals have normal hearing through childhood and into early adult life, and late
     enough that the disease can be mistaken for early presbycusis or for occupational
     noise damage in someone with a noise history.
-  frequency: VERY_FREQUENT
+  frequency: FREQUENT
   phenotype_term:
     preferred_term: Adult-onset sensorineural hearing impairment
     term:
@@ -355,6 +355,33 @@ phenotypes:
     follow-up of the founding kindred. Yan 2013 (PMID:23345450) reports a 12-to-20-year
     window for what is the same family, so the two published descriptions of one kindred
     disagree; the later, prospectively followed report is the one cited here.
+
+    frequency is FREQUENT rather than VERY_FREQUENT for that reason. Adult onset is not
+    established in 80 to 99 percent of carriers: one published reading of the founding
+    kindred puts onset in adolescence, and the Chinese p.Val60Leu family had onset earlier
+    still. What is uncontested across every report is postlingual onset, which is curated
+    separately and graded OBLIGATE; adulthood specifically is the part the sources
+    disagree about.
+- name: Bilateral Sensorineural Hearing Impairment
+  category: Auditory
+  description: >-
+    Both ears are affected in every subject of the ten-year follow-up of the founding
+    kindred, which is the only DFNA41 series with systematic audiological review.
+  frequency: OBLIGATE
+  phenotype_term:
+    preferred_term: Bilateral sensorineural hearing impairment
+    term:
+      id: HP:0008619
+      label: Bilateral sensorineural hearing impairment
+  evidence:
+  - reference: PMID:31593348
+    reference_title: "Progressive Dominant Hearing Loss (Autosomal Dominant Deafness-41) and P2RX2 Gene Mutations: A Phenotype-Genotype Study."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "All affected subjects had bilateral sensorineural hearing loss involving all frequencies with some significant gender differences."
+    explanation: >-
+      Bilateral involvement in all affected subjects of the prospectively followed
+      kindred, which is what makes OBLIGATE rather than a lower band the right grade.
 - name: Noise-Exacerbated High-Frequency Hearing Impairment
   category: Auditory
   description: >-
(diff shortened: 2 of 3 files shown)


kevinschaper:
Two additions since the PR opened, both from checking my own claims rather than from a review.

**The treatment `notes` said no DFNA41-specific report documents amplification or implantation. Half of that was false.** PMID:39258340 (2025) reports a five-generation Chinese kindred carrying p.Val60Leu in which the proband with profound deafness received a cochlear implant with a satisfactory outcome. That is now a curated treatment, and the note claims only what is true — no aiding report exists. The deep-research report for this disease states flatly that no DFNA41-specific interventional outcome series was retrieved, so it missed this paper too.

**The same paper found something more interesting: hyperalgesia.** Pain perception was measured quantitatively, with an instrument rather than by symptom report, and heterozygous p.Val60Leu carriers showed increased pain sensitivity. That is the first result putting pressure on calling DFNA41 nonsyndromic, and it is not a surprising place for this channel to matter — P2X2 receptors sit on sensory neurons and purinergic signalling has an established role in nociception. It is now a phenotype bound to HP:0012534, which carries Hyperalgesia as an exact synonym; `frequency` is deliberately absent rather than set to a low band, because no other DFNA41 report tested pain perception at all, so there is no denominator. The description says this too, so a reader does not have to reach the phenotype list to learn that "nonsyndromic" is doing less work here than it looks.

The same paper also reports earlier onset and worse hearing on the same allele than the original families, which is now a `PARTIAL` counterweight on the adult-onset phenotype — the 25-to-35-year window is a property of the founding kindred, not of the variant.

**Three readout renames, no claim changes.** Two readouts were named for a threshold while their snippets report hearing loss, and a threshold moves opposite to the function it measures; a third was named for susceptibility while carrying `RESTORED`, which read backwards. Prompted by the review on the sibling DFNB120 PR in this batch, which found a real direction-versus-snippet contradiction of the same kind.

**And a rebase.** `check-snippet-grading` began failing locally on `Metastatic_Ovarian_Cancer.yaml`, a file this branch never touches — `origin/main` had moved and the `--against-ref` baseline no longer matched. Rebased onto current `origin/main` and it passes; the diff is otherwise unchanged and the cache diff is still only the seven rows this entry introduces.

kevinschaper:
Both blocking items are fixed, and three of the four suggestions are in as well.

**The adult-onset term exists.** `HP:0008615 Adult onset sensorineural hearing impairment` is real — I confirmed it with `runoak` before editing — so the note claiming HPO has no such concept was simply false. The phenotype is rebound and the note replaced with something true and worth keeping: PMID:31593348 says onset is 25 to 35 years and PMID:23345450 says 12 to 20 years, and both are describing the same kindred. The entry follows the later prospective follow-up, and the note now says so, so the next curator does not have to rediscover the discrepancy.

**Tinnitus is added.** Bound to `HP:0000360`, citing the ClinGen assertion with `evidence_source: OTHER` and the sentence you quoted. `frequency` is omitted: the source counts families ("in one family" of four probands), not individuals, so there is no patient denominator to grade against.

**The ambiguous animal model reference** now names `animal_models#P2rx2 null mouse`.

**The NCIT note** is reworded. I checked the two candidates rather than taking the claim on trust: `NCIT:C219863 Cochlear Implant Revision Surgery` is reachable from `NCIT:C25218`, and `NCIT:C157820 Cochlear Implant` is not. So the note now says what is actually missing, which is a term for primary implantation, rather than implying nothing in that area is reachable.

**The zebrafish paper** is now a fourth animal model. PMID:41015553 is cached as full text and supports three separate quotes: elevated AEP thresholds at all four frequencies, reduced hair cell density in the otolith region, and — graded `PARTIAL` because it bounds the readout rather than supporting it — no difference in the lateral line. The link is `PARTIALLY_RECAPITULATES` at `LOW` fidelity against the progressive hearing loss phenotype, and the limitations say why: a homozygous null in an orthologue against a human heterozygous missense state, no cochlea, and one timepoint, so the progression the phenotype is named for is not shown.

**On GeneReviews**, which you flagged as an unverified check: I ran the queries. PubMed and NCBI Bookshelf both return zero results for a P2RX2 or DFNA41 chapter. Your conclusion was right.

Validation, all exit 0 after rebasing onto current `main`: `just validate` (70/70 snippets verified), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, and the five baseline gates. The cache diff is still additive only, now seven rows: the six from before plus `HP:0008615`.

kevinschaper:
The blocking item is fixed, and you were right for a reason worth recording: **my earlier GeneReviews check used the wrong PubMed field.** I searched `GeneReviews[TA]`, which PubMed does not resolve as a field value — it returns zero with `"phrasesnotfound": ["GeneReviews"]`, and I read that zero as an answer. `GeneReviews[Book]` is the correct filter and returns PMID:20301607 for P2RX2, exactly as you said. The four-line `references:` block is now in, matching the three sibling entries.

I also took your framing on scope: nothing is mined from the chapter. Its cached abstract is a list of what the chapter covers, with no quotable clinical content, and manufacturing snippets from it is the thing the evidence rules exist to prevent.

**Suggestion 1, postlingual onset.** Added, `HP:0008596`, `OBLIGATE`, citing the ClinGen sentence already in the file.

**Suggestion 3, bilaterality.** Added, `HP:0008619`, `OBLIGATE`, citing "All affected subjects had bilateral sensorineural hearing loss involving all frequencies" from the ten-year follow-up.

**Suggestion 2, the adult-onset frequency.** I went with `FREQUENT`. Adding the postlingual phenotype made the inconsistency sharper than it looked: postlingual onset is what every source agrees on, and adulthood specifically is the part they disagree about — Yan 2013 puts the same kindred at 12–20 years and the Chinese p.Val60Leu family was earlier still. So the uncontested claim now carries the higher grade and the contested one carries the lower, with the reasoning in `notes`.

**Suggestion 4, the hiPSC models.** Added as an `experimental_models:` section, fetched as PMID:30819013 (the DOI resolves to it, so the snippets are machine-verified rather than skipped by prefix). One thing to flag about how it is curated: Dong 2019 *builds* the lines and measures nothing in them — no otic differentiation, no electrophysiology, no comparison between the heterozygous and homozygous dosages. So the link is `PERTURBS` with `fidelity: UNKNOWN` and no readouts, and the limitations say plainly that undifferentiated iPSCs are not hair cells and that the engineered homozygote has no human counterpart in a dominant disease. The authors' "good models to investigate the pathological mechanisms" is graded `PARTIAL`, since it is a claim about future utility from a paper that investigated none.

**Suggestion 5, `preferred_term: Hyperalgesia`** — left as is. `preferred_term` is the display name and is allowed to be more specific than the ontology label; here it is also more accurate, since `HP:0012534`'s definition describes pain from a non-painful stimulus while the paper measured heightened response to a painful one. The `notes` already record the divergence.

**Suggestion 6** — PR body updated.

Validation, all exit 0 after rebasing onto current `main`: `just validate` (75/75 snippets verified, up from 70/70), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, the five baseline gates, and `pytest tests/test_data.py -k "entity_ref or model_readout or experimental_model or animal_model or phenotype"` (15,438 passed). No `cache/` rows changed — every term added was already cached.

PR #9926, opened 2026-08-28
curate: Add Optic Atrophy 14 (MONDO:0957824, MIEF1/MID51)

### LINKED ISSUE #9907: Curate optic atrophy 14 (MONDO:0957824)
Curate a dismech entry for **optic atrophy 14** ([MONDO:0957824](https://monarchinitiative.org/MONDO:0957824)) — *MIEF1* (MID51) variants; MIEF1 is a mitochondrial outer-membrane DRP1 receptor, so this sits in the same fission/fusion axis as the OPA1- and DNM1L-related optic atrophies.

The stub gives one causal gene (`hgnc:25979` MIEF1) and no descendants, and nothing in `kb/` mentions MIEF1. Check whether an existing optic atrophy or mitochondrial dynamics entry should gain a `conforms_to` relationship rather than this becoming a standalone entry.

Stub: `stubs/Optic_Atrophy_14.yaml`

kevinschaper:
Adds a new disorder entry for **optic atrophy 14** — the dominant optic neuropathy caused by heterozygous *MIEF1* missense variants. MIEF1 encodes MID51, the outer mitochondrial membrane adaptor that recruits the fission GTPase DRP1, so this sits in the same fission/fusion axis as the OPA1- and DNM1L-related optic atrophies.

The headline caveat is that the entire human evidence base is **two women from one French cohort of 200**, with no reported family segregation. The entry is written around that rather than despite it.

Closes #9907

## entry_type decision: DISEASE

One causal gene, no MONDO descendants, its own MONDO and OMIM identity (OMIM:620550), and a coherent pathograph. Nothing in `kb/` mentioned MIEF1 before this, so there was no existing entry to extend.

## The `conforms_to` question the issue asked about: no, and here is why

The only candidate module is `mitochondrial_dysfunction`. Its five nodes are age-related mitochondrial damage and mtDNA mutation, bioenergetic decline and oxidative stress, impaired mitophagy and quality control, mitochondrial contribution to senescence and inflammation, and age-related tissue dysfunction. **None of them is about fission/fusion balance**, and no bioenergetic, reactive-oxygen or mitophagy measurement exists for MIEF1 — so conforming to the bioenergetic node would assert an unmeasured claim in order to gain a link. `Autosomal_Dominant_Optic_Atrophy_Plus` (OPA1) declares no `conforms_to` either.

What is actually missing is a **mitochondrial-dynamics module** covering OPA1, MFN2, OPA3, DNM1L, MFF, SPG7, AFG3L2, YME1L1 and MIEF1, whose shared node would be that regulated fission/fusion balance is required for retinal ganglion cell survival — with the striking property that imbalance in *either* direction gives the same disease. That should be built from the OPA1 and DNM1L literature, not from a two-patient entity. The entry's `notes:` records this so the absence is a decision, not an oversight.

## What is mechanistically interesting

The disease is defined as much by what it is not. Both variants:

- **do** reach the outer mitochondrial membrane normally
- **do** oligomerise normally
- **do not** reproduce wild-type MID51's effect on the mitochondrial network

Neither variant is in the transmembrane domain or at a known oligomerisation residue. Both negatives are curated as their own pathophysiology node with `REFUTE` evidence, so a later curator does not attach a mislocalisation mechanism to this entry: it was looked for and was not there.

An independent group (PMID:36044022) then localised the lesion further — p.Y240N disrupts DRP1 recruitment while leaving MID51/Fis1 oligomerisation intact, selectively inhibiting DRP1 GTP hydrolysis. In the same experiments it **leaves the mitochondria-lysosome contact pathway alone**, while a different MID51 variant in the oligomerisation interface does the reverse and is tentatively linked to Parkinson's disease. One gene, two interfaces, two diseases — curated as an `OPEN_QUESTION`, with the caution that the Parkinson's half is described in the source as *potential*, and that the second optic-atrophy allele has never been tested in those assays.

## Grading, and the thin evidence base

- Trigger and DRP1-recruitment nodes: `PROVISIONAL`.
- HeLa model link: `fidelity: LOW`. Not just the usual cell-line caveat — the cited review states that heterologous MiD overexpression and increased *endogenous* MiD have **opposite** effects on the network, so the assay's baseline is a condition that does not occur in a patient, and the readout is a difference from overexpressed wild-type rather than from an untransfected cell.
- `functional_impact_category: LOSS_OF_FUNCTION`, with the reasoning written out and `DOMINANT_NEGATIVE` named as the live competing hypothesis that no reported experiment excludes.
- An open `KNOWLEDGE_GAP` discussion states plainly that a MONDO identifier and an OMIM number record that a claim was made, not that it was replicated.

## Two things I checked and corrected

**A typo in the source.** The defining paper's Figure 2a legend places `c.436C>T` at "position 416 of the MID51 protein" while naming p.R146W in the same sentence. 146 is the position consistent with the nucleotide change, the sequence alignment and every other mention. The genetic notes flag 416 as a transcription error so it does not propagate.

**A deep-research figure that did not survive checking.** The falcon report states MIEF1 accounted for ~1.08% of positive optic atrophy diagnoses in the 2,186-proband Rocatcher series. That series' abstract says the genes outside its top ten each account for *less than* 1%, and does not name MIEF1 at all. The prevalence block curates the bound the abstract supports — MIEF1 is absent from the ten genes covering >96% of nuclear diagnoses, and the whole residue is 17 individuals across eleven genes — and the discrepancy is recorded in the entry's notes. NEC preflight itself passed cleanly (MIEF1 = 47 mentions, next gene OPA1 = 3).

Also worth noting: that 2023 series comes from the same Angers reference centre and shares five authors with the 2021 MIEF1 report. Any MIEF1 case in it is presumptively one of the two already reported. The entry counts two patients, not four.

No `treatments:`, no `animal_models:` (none exists in any species), no `progression:` block. The idebenone literature for LHON is close enough to look importable and would attribute to two patients an experience recorded only in people with other genotypes.

I also declined to state a direction for the cellular phenotype. "MIEF1 variants cause mitochondrial fragmentation" and "cause hyperfusion" are both tempting compressions of the assay result and neither is supported — nothing has been measured in a patient's cells.

## Validation

| check | exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just validate-disorders` | 0 (34/34 snippets verified) |
| `just check-folded-hyphens` | 0 (caught two real defects, fixed) |

### DIFF
--- a/kb/disorders/Optic_Atrophy_14.yaml
+++ b/kb/disorders/Optic_Atrophy_14.yaml
@@ -448,6 +448,14 @@ phenotypes:
         than ages at onset, which the source does not separate cleanly for
         patient 1 - she noticed a faint visual problem in 2002 and lost acuity in
         2015 - so treat them as bounding the presentation rather than dating it.
+
+        On the band chosen, because the paper's own title invites the wrong one.
+        `MIDDLE_AGE` is HP:0003596 "Middle age onset", which HPO scopes to 40-60
+        years, and both patients fall inside it. `LATE` is HP:0003584, scoped to
+        onset after 60, and would be wrong here. The paper is titled "late onset"
+        in the ordinary sense - late relative to the first two decades, when
+        other dominant optic atrophies declare themselves - not in HPO's sense.
+        Do not "correct" this to LATE on the strength of the title.
   evidence:
   - reference: PMID:33632269
     reference_title: "Dominant mutations in MIEF1 affect mitochondrial dynamics and cause a singular late onset optic neuropathy."


kevinschaper:
All three blocking items fixed, both non-blocking notes taken, plus one item from the research comment on #9907.

**1. The `REFUTE` polarity was inverted and you're right about why.** The node is named "Preserved MID51 Localisation and Oligomerisation", so evidence showing preservation *supports* it; as curated it read as "preserved localisation is refuted". Both items are now `SUPPORT`, with the excluded hypothesis named in the `explanation` instead. I kept the node name rather than renaming it to the excluded hypothesis, because the entry's prose elsewhere refers to it as the preserved state.

**2. `GO:0005515` → `GO:0030674` protein-macromolecule adaptor activity.** Better in exactly the way you say — it matches the entry's own description of MID51 as the adaptor that recruits DRP1, and it was already in the enum cache so the diff is a single row.

**3. RNFL thinning added** as `HP:0020119`, using the already-verified OCT snippet, with a note in its description that only one of the two patients had OCT.

**Non-blocking:** the Optic Atrophy phenotype now carries a structured `onset` (`onset_category: MIDDLE_AGE`, 47–55), with a `notes` field warning that those are ages at *study* — the source does not cleanly separate onset for patient 1, who noticed a faint problem in 2002 and lost acuity in 2015. I added the paper's own adult-onset discriminator sentence as supporting evidence. The `evidence_source` at the old line 555 is now `OTHER`, since that sentence is introduction framing rather than a result.

**From #9907:** added the mitochondria–lysosome specificity result — p.Y240N still forms a Mid51/Fis1 complex and does not misregulate lysosomal untethering or lysosomal network dynamics. I graded it `SUPPORT` rather than `REFUTE`, deliberately, for exactly the reason in item 1: the claim the node makes is the preserved state, so the excluded mechanism belongs in the explanation. It turns the preservation from an in vitro observation into a statement about which of MID51's two functions the disease damages, which is what the pathway-selectivity discussion is about.

Also rebased. Before the rebase `check-snippet-grading` failed on a divergence in `Metastatic_Ovarian_Cancer.yaml`, which this PR never touches — the branch was simply behind main, and it passes after rebasing.

Re-ran everything: `validate`, `validate-terms`, `validate-disorders` (37/37 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0.

kevinschaper:
One further push from a self-audit, unrelated to the review items.

The `notes:` said "no MIEF1 animal model exists in any species". That overstates what was actually checked. *Mid51* is studied in mice and cultured cells for its ordinary fission and mitophagy roles, and a search on the gene returns that work ahead of anything disease-related. What is genuinely absent is any animal carrying a **disease-linked** MIEF1 allele.

Restated accordingly, with the reason spelled out: a later curator should not promote a Mid51 mouse paper into this entry on the strength of the gene name, because none of that work involves p.Y240N or p.R146W.

`validate` and `check-folded-hyphens` re-run, both exit 0.

kevinschaper:
One divergence from your suggestion 4, with the reasoning, plus a note added to the entry so nobody undoes it.

You suggested `onset: {onset_category: LATE}` (HP:0003584). **I used `MIDDLE_AGE` (HP:0003596) instead**, because HPO scopes "Late onset" to after 60 and "Middle age onset" to 40–60, and the two patients were 47 and 55. The paper's title says "late onset" in the ordinary sense — late relative to the first two decades, when the other dominant optic atrophies declare themselves — not in HPO's sense.

That gap between the paper's wording and the ontology band is exactly the sort of thing a later curator would "fix" in the wrong direction on the strength of the title, so the reasoning is now written into the `onset` `notes` field alongside the values.

The rest of suggestion 4 is taken: the onset block also records that 47 and 55 are ages at *study*, since the source does not cleanly separate onset for patient 1, and the paper's own adult-onset discriminator sentence is now cited on the phenotype.

Also, thank you for checking `glaucoma_optic_neuropathy` independently. The normal-IOP point is a better argument against it than anything I had, and it makes the no-conformance decision more solid than the one I wrote.

PR #9927, opened 2026-08-28
curate: Add Spastic Paraparesis-Cataracts-Speech Delay Syndrome (MONDO:0036212, FAR1)

### LINKED ISSUE #9908: Curate spastic paraparesis-cataracts-speech delay syndrome (MONDO:0036212)
Curate a dismech entry for **spastic paraparesis-cataracts-speech delay syndrome** ([MONDO:0036212](https://monarchinitiative.org/MONDO:0036212)) — heterozygous *FAR1* variants that escape the normal plasmalogen-mediated degradation of fatty acyl-CoA reductase 1, so ether lipid synthesis runs unchecked.

**Watch the direction of effect.** `kb/disorders/Rhizomelic_Chondrodysplasia_Punctata_*.yaml` already cover *biallelic loss* of FAR1 (RCDP4, plasmalogen deficiency). This disease is the opposite arm — monoallelic variants that stabilize FAR1 and *raise* plasmalogen synthesis. Same gene, different mechanism and different entity; do not let a deep-research report collapse the two, and grade the `modifier` slot accordingly (`GAIN_OF_FUNCTION` needs the qualitative justification described in `CLAUDE.md`, otherwise `INCREASED`).

Stub: `stubs/Spastic_Paraparesis-cataracts-speech_Delay_Syndrome.yaml`

kevinschaper:
Adds a new disorder entry for the **dominant, gain-of-function arm of FAR1 disease** — de novo heterozygous variants at a single codon, Arg480, of fatty acyl-CoA reductase 1.

The point of the entity is an inversion. In the recessive FAR1 disease this knowledge base already curates (FAR1 deficiency / RCDP4), losing FAR1 abolishes fatty-alcohol production and patients are plasmalogen-**deficient**. Here FAR1 works fine — catalysis intact, peroxisomal targeting normal. What the Arg480 variants destroy is the brake: FAR1 is normally degraded in response to rising plasmalogens, and that feedback loop is abolished, so ether lipid synthesis runs without a governor.

Closes #9908

## Keeping the two FAR1 diseases apart

Most of the FAR1 literature is about the other one, so every reference was checked against three discriminators — **heterozygous, codon 480, and *elevated* rather than deficient plasmalogens**:

| reference | passes | role |
|---|---|---|
| PMID:33239752 | yes — 12 patients, all de novo Arg480, plasmalogens elevated | defining cohort, full text |
| PMID:36254151 | yes — de novo p.Arg480Cys in the reported patient | additional case |
| PMID:37335441 | yes — de novo p.Arg480Ser | fourth substitution at the same codon |
| PMID:36720320 | n/a — a review of the normal feedback pathway, belonging to neither disease | mechanism |

PMID:36781603 (the *Mov Disord* FAR1 variant spectrum paper) was fetched, came back `content_type: unavailable`, and was dropped; its cache file is not committed.

## The `modifier` call, argued rather than assumed

The issue asked for this specifically, so the entry splits the claim across two nodes rather than putting one value on both:

- **`INCREASED`** on "Elevated FAR1 Protein and Enzyme Activity". Purely quantitative: the same enzyme doing the same reaction in larger amounts — activity 379/323/310 against control 74/90/101 pmol/h/mg — with normal peroxisomal targeting and normal peroxisomal β-oxidation and DHAPAT. This keeps the annotation PATO-bound.
- **`GAIN_OF_FUNCTION`** on "Uncontrolled Ether Lipid Synthesis". Qualitative, and here is the argument. Loading control fibroblasts with HDG (which raises plasmalogen by a route bypassing FAR1) drops FAR1 protein by about 40%. Patient cells got the same plasmalogen rise and FAR1 did not move — 101% of baseline. And untreated patient cells already sat at roughly the plasmalogen level that suppresses FAR1 in controls, while holding FAR1 at three times the suppressed control level. The pathway is not running above a set point; the loop that would set one has been eliminated. The authors' own word is "uncontrolled". The trade — losing ontology grounding for expressivity — is made knowingly and stated in the entry's `notes:`.

`GeneticContext.functional_impact_category` is **`HYPERMORPHIC`**, not the generic `GAIN_OF_FUNCTION`: the enum defines it as increased *normal* gene product activity, which is exactly what was measured. No novel activity is reported, so `NEOMORPHIC` is wrong, and `DOMINANT_NEGATIVE` is excluded by the direction of effect.

## A MONDO duplication

**MONDO:0036212** and **MONDO:0100230** both name this disease. The first carries the OMIM (619338), Orphanet, MedGen and UMLS cross-references and has "Fatty acyl-CoA reductase 1 superactivity" as a synonym; the second carries only a GARD reference and its definition is the defining paper's title restated. They are the same entity.

This entry anchors on MONDO:0036212 and maps MONDO:0100230 as `skos:exactMatch`. Worth knowing: `Rhizomelic_Chondrodysplasia_Punctata_Plasmalogen_Synthesis_Defect` already refers to this disease as a differential **under MONDO:0100230**, so both identifiers now resolve into the knowledge base. I have not touched that file — this looks like something to report upstream to MONDO rather than paper over from either side.

## Treatment and its indication

The only treatment curated is symptomatic antiseizure pharmacotherapy — barbiturates, levetiracetam, oxcarbazepine — and it is genuinely reported in this cohort, with 4 of the 8 treated patients withdrawn from antiepileptics without recurrence. Its indication (seizures, 8/12) is curated as a phenotype in the same entry and named in `target_phenotypes`.

## The question the entity poses

Cataract is *caused* by ether lipid deficiency — supplementing alkylglycerol in a plasmalogen-deficient mouse prevented it. Cataract is also present in 12 of 12 patients whose ether lipids are elevated. Same for spastic paraparesis, which occurs in RCDP4, in all of these patients, and in the three other ether-lipid-*excess* disorders (SELENOI, PCYT2, Sjögren-Larsson). Too much and too little of the same lipid class damage the same two organs.

The entry keeps three hypotheses distinct and prefers none — that the ether/non-ether *ratio* matters rather than the absolute level, that fatty alcohols rather than lipids are the toxic species, or that the convergence is coincidence — and warns explicitly against importing RCDP mechanism prose here on the strength of the shared phenotype.

## Snippet trimming, and a tooling gap

Four snippets had to be trimmed or restarted mid-sentence because this literature writes ether-lipid nomenclature as `PC[O]`, `DG[O]`, `TG[O]` and units as `pmol/[h.mg protein]`, and the reference validator strips bracketed spans. Each trimmed snippet's `explanation` records what was left out and why, so it does not read as selective quotation.

This looks like a real gap rather than a one-off: `[O]` is standard lipid nomenclature and `conf/reference_validator_config.yaml`'s `literal_bracket_patterns` (`^[A-Z]{2,}$` and `%`) does not match single letters. I have deliberately **not** changed that config in a curation PR — flagging it for a separate one.

## Deep research

Provider: falcon. NEC preflight **PASS** (FAR1 = 67 mentions). As the task warned, preflight passing does not clear the arm question — both FAR1 diseases would pass a gene check — so I read the report against the arm criteria separately. It is squarely about the monoallelic-gain arm and contrasts it with recessive FAR1 deficiency explicitly in its own definition section. Nothing was used from it without checking against the cached full text.

## Validation

| check | exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just validate-disorders` | 0 (43/43 snippets verified) |
| `just check-folded-hyphens` | 0 (caught one real defect, fixed) |

### DIFF
--- a/kb/disorders/Spastic_Paraparesis-cataracts-speech_Delay_Syndrome.yaml
+++ b/kb/disorders/Spastic_Paraparesis-cataracts-speech_Delay_Syndrome.yaml
@@ -502,7 +502,7 @@ pathophysiology:
       Equally unexplained, and more puzzling, because cataract is also caused by
       ether lipid *deficiency* - see the open discussion.
 - name: Corticospinal Tract Dysfunction
-  biological_scale: ORGANISM
+  biological_scale: TISSUE
   role: outcome
   mechanism_confidence: PROVISIONAL
   description: >-
@@ -518,6 +518,15 @@ pathophysiology:
     imaging, electrophysiological or neuropathological localisation to the
     corticospinal tract has been reported in any patient, and brain MRI was
     normal in ten of twelve.
+
+    Scaled TISSUE rather than ORGANISM, to match Lens Opacification below. The
+    two nodes are the disease's two end-organ failures and the substrate named in
+    each is an anatomical structure - a tract and a lens - so tagging one at
+    organism scale and the other at tissue scale would make the pair look like
+    different kinds of claim when they are the same kind. The organism-level
+    reading is available and was rejected: spastic paraparesis as a whole-body
+    motor phenotype would be ORGANISM, but that is the phenotype, curated in
+    `phenotypes:`, not this node.
   evidence:
   - reference: PMID:33239752
     reference_title: "An autosomal dominant neurological disorder caused by de novo variants in FAR1 resulting in uncontrolled synthesis of ether lipids."


kevinschaper:
Pushed one follow-up commit acting on the research comment on #9908, ahead of the review.

Three phenotypes added, each from a verbatim sentence in the already-committed cache:

- `HP:0002061` Lower limb spasticity — the distribution that makes this a paraparesis, noting the distinctive pairing with axial hypotonia.
- `HP:0001249` Intellectual disability at 3/12 — worth having explicitly because it is one of the clearest separators from recessive FAR1 deficiency, where it is profound and universal.
- `HP:0000256` Macrocephaly at 2/12 — recorded partly because the direction matters: recessive FAR1 deficiency causes *micro*cephaly.

**Constipation is deliberately left out** despite being at 5/12 in the research comment. It appears only as plus/minus marks in Table 1 and never in the paper's prose, so no sentence can be quoted for it, and I would rather leave it uncurated than source a phenotype to a table cell. That decision — and the other per-patient columns it applies to, such as ambulation age and seizure type — is recorded in the entry's `notes:` so a curator with the table in front of them can add them properly.

Branch rebased onto `main`. All gates re-run: `validate`, `validate-terms`, `validate-disorders` (46/46 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0.

kevinschaper:
One more push, from a self-audit of this entry's absence claims. It caught a misleading one.

The entry said flatly, in three places, that there is no animal model. **That is wrong as written.** A *Far1* knockout mouse was published in 2023 ([PMID:37039784](https://pubmed.ncbi.nlm.nih.gov/37039784/)) and is explicitly presented by its authors as a model of ether lipid **deficiency** — subviable, growth-retarded, infertile, spermatogenesis arrested at the round spermatid stage. That is the *recessive* arm: the mechanistic inverse of this disease.

The claim is now stated precisely — no animal carries an Arg480 allele, and the existing knockout models the other arm and cannot substitute — and the paper is cited in two places where it does useful work:

- In the **RCDP4 differential**, so a reader who searches "Far1 mouse" does not read that literature as bearing on Arg480 gain of function. That is exactly the confusion this whole entry is built to prevent, and leaving the absence claim unqualified would have invited it.
- In the **knowledge-gap discussion**, where it sharpens the point rather than weakening it: the deficiency arm has an animal model and the excess arm does not, which is the wrong way round for answering the question this entity poses. It also means the genetic background and assay platform for the excess-versus-deficiency comparison already exist; what is missing is the Arg480 knock-in counterpart.

`distinguishing_features` on the differential gains a line to match: the recessive arm has a mouse model, the dominant arm has none.

Gates re-run after the change: `validate`, `validate-terms`, `validate-disorders` (47/47 snippets), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0.

kevinschaper:
One more small self-audit fix, pushed.

The ether-lipid-excess differential said ichthyosis is the readiest clinical separator for Sjögren-Larsson syndrome. That is true, but it is a claim about a *different* disease that no source cited in this entry states — the sentence I quote names the three disorders and their shared paraparesis-plus-eye pattern and stops there. Removed, and replaced with a line saying the entry claims only what its source states and leaves distinguishing those three to their own entries.

Small, but it is the same class of thing as the animal-model claim: an assertion about something outside the entry that reads as sourced because it sits next to sourced material.

kevinschaper:
Correction to my own previous comment on this PR, and a small commit to make it true.

I wrote that `distinguishing_features` on the RCDP4 differential "gains a line to match: the recessive arm has a mouse model, the dominant arm has none". **It did not.** That edit was lost when a scripted patch aborted partway through, and I described the intended change rather than checking the diff. The line is now actually there.

Flagging it rather than quietly fixing it, because it is the same failure as the animal-model claim it was meant to accompany — an assertion about something outside the sentence, stated without checking. In that case it was about the literature; here it was about my own diff, which is worse.

kevinschaper:
Last change on this branch, from the same self-audit: the RCDP4 differential contrasted the two FAR1 diseases with a quotation on the dominant side only, while the recessive arm's phenotype — profound intellectual disability, severe growth failure, microcephaly, dysmorphism — sat in prose. It is in the same cached full text, so it is now quoted. The contrast rests on evidence on both sides.

48/48 snippets verified; all gates exit 0. I'll stop here and wait for the review.

kevinschaper:
Both blocking items fixed, plus the supportive-management suggestion.

**Phenotype frequencies.** Seven now carry the fraction verbatim from their own verified snippet — 12/12 spastic paraparesis and cataracts, 10/12 speech delay, 8/12 seizures, 6/12 axial hypotonia, 3/12 intellectual disability, 2/12 macrocephaly — using the string form via `any_of`, as in `Activated_PI3K-delta_Syndrome`. Gross motor delay takes the `FREQUENT` enum, since its source says only "in most".

Bare fractions rather than fractions-plus-percentages, deliberately: the percentages would be my arithmetic, not the paper's.

Two phenotypes still carry no `frequency` and now say why in their descriptions. Ankle clonus and lower limb spasticity are named inside the pyramidal-signs sentence without a count, in the same breath where the source *does* give fractions for spastic paraparesis and truncal hypotonia. Inferring 12/12 from sentence structure would be reading a number the paper declined to give.

Your point about why this matters is the right one — intellectual disability at 3/12 against profound-and-universal in the recessive arm is the RCDP4 discriminator, and it was reachable only as prose.

**The "no neuropathology" claim.** Correct, and thank you for finding it — that is the third time in this PR the same defect class has surfaced, which is telling. I fetched PMID:39074165 and confirmed `content_type: unavailable`, so I qualified the prose and did not cite it, matching the PMID:36781603 disposition; the cache file is not committed.

The gap's real point survives and is now stated more precisely: patient tissue beyond fibroblasts is *not* untouched — there is muscle lipidomics and proteomics from a heterozygous de novo patient — but neither of the two tissues that actually fail, the corticospinal tract and the lens, has been examined in anyone. That is a sharper statement of the gap than the one I had.

**Supportive management** is now recorded in `notes:` as considered and not curated: cataract extraction, physiotherapy and speech therapy all follow from curated phenotypes, but the defining paper never mentions surgery or therapy of any kind, and adding them from general practice would assert a management record for twelve patients that nobody wrote down.

Gates: `validate`, `validate-terms`, `validate-disorders` (48/48), `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0, no cache diff.

kevinschaper:
This re-review was written against `942e4b95`, and both 🟡 items were already fixed in `64b9bb86`, pushed at 21:57Z with a comment above. Restating what landed so the next pass does not have to diff it:

- **Frequencies** — exactly the seven you list, each taken verbatim from the fraction inside that phenotype's own already-verified snippet. Ankle Clonus, Lower Limb Spasticity and Delayed Gross Motor Development are left unset, which is the call you independently reached: the first two sit in a sentence that quantifies only paraparesis and truncal hypotonia, and the third is quoted as "in most". Their descriptions now say so, rather than leaving the omission to be read as an oversight. Bare fractions, no percentages — those would be my arithmetic, not the paper's.
- **The absence claim** — qualified, not cited. I fetched PMID:39074165 and confirmed `content_type: unavailable`, so it cannot carry a verified snippet; the cache file is not committed. The rationale now names it as the one report of patient tissue beyond fibroblasts and restates the gap more precisely: neither of the two tissues that actually fail — the corticospinal tract and the lens — has been examined in any patient.

Your framing of it as "the same class of defect you already self-corrected twice on this branch" is the useful part, and it is why I went back over the entry a third time rather than just patching the one sentence.

**On the three 🔵 notes:**

1. **`biological_scale` — taken, just pushed.** You were right that it read oddly. Corticospinal Tract Dysfunction is now `TISSUE`, matching Lens Opacification. They are the disease's two end-organ failures and both name an anatomical structure, so splitting them across scales made one look like a different kind of claim. The node now records that the organism-level reading was considered and rejected: spastic paraparesis as a whole-body motor phenotype would be `ORGANISM`, but that is the phenotype in `phenotypes:`, not this node.
2. **Supportive management** — already in `notes:` as of the earlier push, recorded as considered-and-declined with the reason (the defining paper never mentions surgery or therapy of any kind).
3. **The MONDO self-mapping** — keeping it. It is the established shape in this KB rather than a redundancy: `3-Hydroxyisobutyryl-CoA_Hydrolase_Deficiency` and `46_XY_DSD_Due_to_5_Alpha_Reductase_2_Deficiency` both carry their primary term in `mondo_mappings` alongside `disease_term`, with a `mapping_justification` naming it as primary. Dropping it here would make the duplicate-term write-up harder to read, since the point is that *two* MONDO terms map to this entity and only one is primary.

PR #9928, opened 2026-08-28
curate: add Siddiqi syndrome (MONDO:0032842)

### LINKED ISSUE #9880: Curate Siddiqi syndrome (MONDO:0032842)
Curate a dismech entry for **Siddiqi syndrome** ([MONDO:0032842](https://monarchinitiative.org/MONDO:0032842)) — an autosomal recessive deafness–dystonia syndrome caused by biallelic variants in `FITM2`, an ER membrane protein that partitions diacylglycerol and controls lipid droplet biogenesis.

Preflight: MONDO:0032842 appears nowhere in `kb/disorders/` or `kb/groupings/` on `main`, and `FITM2` (hgnc) appears in no existing entry's `genetic:` block — so this is not covered under another name or as a subtype of a curated parent. No open PR matches.

The mechanistic angle worth getting right is why a lipid-droplet/DAG-handling protein produces a *sensorineural + movement disorder* phenotype rather than a lipodystrophy, which is what the gene family would predict.

Stub: `stubs/Siddiqi_Syndrome.yaml`

nlharris:
Adds a Disease entry for Siddiqi syndrome — an autosomal recessive deafness–dystonia syndrome from biallelic loss of function in `FITM2`, an ER fatty acyl-CoA diphosphatase.

Closes #9880

## The gene ID is wrong again, and differently

The report gives the causal gene as **`HGNC:14929`** in three places, including its header and summary table. That is **SIRT1** — a sirtuin deacetylase with no relation to lipid-droplet biology. FITM2 is **`hgnc:16135`**.

The detail that makes this worth recording rather than quietly fixing: the three *other* identifiers on the same line — OMIM 612029, NCBI Gene 128486, UniProt Q8N6M3 — are **all correct for FITM2**. A wrong HGNC ID sat between correct identifiers for the right gene, which is precisely the configuration in which a reader stops checking.

**This is the second of three reports in this batch with a bad causal-gene ID**, and the two errors are unalike:

| Entry | Report gave | Actually is | Correct ID |
|---|---|---|---|
| SPG81 (#9894) | `HGNC:30396` | SELENOS — *same protein family* | `hgnc:29361` |
| Siddiqi (this PR) | `HGNC:14929` | SIRT1 — *unrelated* | `hgnc:16135` |
| SEMD (#9914) | `HGNC:23994` | correct | — |

One near-miss, one unrelated gene, one right. There's no systematic offset to correct for and no heuristic that would have flagged either — the identifier just can't be trusted. Recorded in `notes` so the next reader of these reports is warned.

## The most interesting record is a negative one

FITM2 partitions triglyceride into lipid droplets, so the naive prediction from the gene is a **lipodystrophy**. The founding report looked and found none:

> Despite the role of FITM2 in neutral lipid storage and metabolism, **no indications for lipodystrophy were observed** in the affected individuals

That dissociation is curated as its own phenotype with `supports: REFUTE`, because it is the observation that pointed the field away from a storage defect and toward the ER-clearance account the 2020 enzymology later supplied. The disease is not caused by failing to store fat; it is caused by failing to clear an intermediate.

**Curating it exposed a schema gap worth flagging.** `Phenotype` has no `presence` slot, unlike `Biochemical` and `Genetic`. So an explicitly absent finding cannot be marked absent structurally — the absence is carried only by the description and the REFUTE grade. A term-level query will see `Lipodystrophy` attached to this disease and cannot tell that the entry asserts the opposite. That's stated in the phenotype description rather than left implicit.

It is also, deliberately, the one phenotype not reached by a causal edge. An absent finding shouldn't hang off the pathograph.

## Pathophysiology

`Loss of acyl-CoA diphosphatase activity` → `unbuffered acyl-CoA and failed lipid droplet partitioning` → `ER structural disruption and ER stress` → `injury to post-mitotic metabolically active cells`.

**The terminal node is labelled an interpretation and graded to say so.** No cited source demonstrates ER stress in patient cochlea, neurons or skin. The tissue-selectivity argument — that post-mitotic, high-flux cells can't dilute damaged ER by dividing — is the account that makes the distribution intelligible, not a finding. A `KNOWLEDGE_GAP` records it, and notes that the absent lipodystrophy *sharpens* the question: if the pathogenic event were failed lipid storage, adipose tissue should be first to fail and is instead spared.

## The fly model, and why fidelity is MODERATE

The *Drosophila* RNAi knockdown reproduces locomotor impairment, hearing loss and sensory dysfunction, and was the founding study's independent evidence of causality. The fly has one Fitm ortholog where mammals have two paralogs, which makes it a cleaner test of the pathway than a mouse would be.

`fidelity: MODERATE` rather than HIGH, for reasons in `limitations`: it's a knockdown not a null; fly auditory and sensory organs aren't homologous to cochlea or DRG, so those readouts are analogous rather than equivalent; and the model was never used to test the ER-stress mechanism this entry proposes — it establishes that the gene matters, not why.

## Gradings

Several phenotypes carry `PARTIAL`, each with the reason in its `explanation`. Two are worth noting: ichthyosis and sensory neuropathy are sourced to text saying "ichthyosis-**like** features" and "**signs of** sensory neuropathy". Those hedges are the authors' and are preserved rather than promoted to diagnoses.

**No frequencies** — one founding family plus a few since gives no denominator.

## Evidence

17/17 snippets verified. All eight quotes taken from the report were re-verified against primary abstracts before use; all eight matched exactly. Two reference caches I'd fetched but didn't end up citing were removed from the diff rather than shipped.

## Validation

`just validate`, `just count-verified-snippets` (17/17), `just check-entity-refs`, `just check-duplicate-keys` — all pass. One root, no dangling targets.

---

### DIFF
--- a/history/disorders/Siddiqi_Syndrome/2026-08-28T234830Z-claude-code-66fa5c.yaml
+++ b/history/disorders/Siddiqi_Syndrome/2026-08-28T234830Z-claude-code-66fa5c.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Siddiqi_Syndrome
+  path: kb/disorders/Siddiqi_Syndrome.yaml
+session:
+  id: 2026-08-28T234830Z-claude-code-66fa5c
+  timestamp: '2026-08-28T23:48:30Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9928
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - genetic
+  - diagnosis
+  - treatments
+  - evidence
+  summary: 'Review round 2: variants, diagnosis, treatments, title-quote fix'
+  details: |
+    Second review round on PR #9928.
+
+    Replaced the Developmental Regression snippet, which quoted a substring of PMID:28067622's title and failed `just check-title-snippets`, with a non-title sentence from PMID:41113320.
+
+    Added structured Genetic.variants: the six-family allele catalog (which also supplies the previously uncited compound-heterozygote claim), the Iranian allele, and the truncating-versus-missense severity gradient graded PARTIAL on the authors' own hedge.
+
+    Added Global Developmental Delay (HP:0001263). Added a diagnosis section (exome sequencing with Sanger segregation; auditory brainstem response) and four differential diagnoses.
+
+    Added a treatments section: supportive management as reported, plus cochlear implantation graded PARTIAL because the authors propose it rather than report it. Corrected the stale top-level note claiming no citable management protocol existed - true of the two references the entry had at round 0, false once PMID:41113320 was fetched.
+
+    Moved the adipose-specific FIT2 knockout mouse evidence off the Absence of Lipodystrophy phenotype record onto the lipid-droplet partitioning node, so a term-level query is not told that model-organism evidence supports lipodystrophy in this disease.
+
+    Both NCIT terms initially chosen for the diagnosis section were wrong (NCIT:C101293 is Next Generation Sequencing, NCIT:C38031 is Tearing) and were caught by term validation, then replaced with NCIT:C101295 and NCIT:C184949 verified via OAK.
+
+    Validation: 33/33 snippets verified, check-title-snippets, check-entity-refs, check-duplicate-keys, check-folded-hyphens and check-snippet-length all pass.
--- a/kb/disorders/Siddiqi_Syndrome.yaml
+++ b/kb/disorders/Siddiqi_Syndrome.yaml
@@ -52,10 +52,13 @@ notes: >-
   curator should treat HGNC identifiers in these reports as unverified regardless
   of how well-formed the surrounding record looks.
 
-  No treatments section. No disease-modifying therapy exists and management is
-  supportive; neither cited primary states a management protocol quotable
-  exactly, so the section is absent rather than asserted from an uncitable
-  source.
+  Treatments. No disease-modifying therapy exists. The treatments section
+  records the supportive management actually given in the Iranian family, plus
+  cochlear implantation, which those authors propose rather than report - graded
+  and described to keep the two apart. An earlier revision of this entry omitted
+  the section on the grounds that no cited source stated a protocol; that was
+  true of the two references the entry then had and stopped being true once
+  PMID:41113320 was fetched.
 external_assertions:
   - name: OMIM Siddiqi syndrome record
     source: OMIM
@@ -147,6 +150,23 @@ pathophysiology:
           States the lipid-droplet function whose failure this node represents.
           Graded OTHER because the sentence is background rather than a result
           of the study reporting it.
+      - reference: PMID:30020828
+        reference_title: "Fat storage-inducing transmembrane protein 2 (FIT2) is less abundant in type 2 diabetes, and regulates triglyceride accumulation and insulin sensitivity in adipocytes."
+        supports: SUPPORT
+        evidence_source: MODEL_ORGANISM
+        snippet: >-
+          Mice with adipose tissue (AT)-specific FIT2 knockout exhibited
+          lipodystrophic features, with reduced AT mass, insulin resistance, and
+          greater inflammation in AT when fed a high-fat diet
+        explanation: >-
+          Demonstrates that removing FIT2 does disrupt triglyceride partitioning
+          in vivo, in the tissue built to store it. The node is evidenced by the
+          mouse rather than by the patients precisely because the patients show
+          no lipodystrophy - the dissociation argued in the
+          siddiqi_tissue_selectivity discussion. Placed on the mechanism rather
+          than on the phenotype record so that a query over phenotype terms is
+          not told that model-organism evidence supports lipodystrophy in this
+          disease.
   - name: ER Structural Disruption and ER Stress
     biological_scale: CELLULAR
     description: >-
@@ -238,6 +258,7 @@ pathophysiology:
       - target: phenotypes#Ichthyosis
       - target: phenotypes#Sensory Neuropathy
       - target: phenotypes#Intellectual Disability
+      - target: phenotypes#Global Developmental Delay
       - target: phenotypes#Growth Delay
     evidence:
       - reference: PMID:28067622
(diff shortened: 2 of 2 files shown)


nlharris:
Pushed as `9cec0f115`. Your finding 2 was right in a way I want to be precise about: fetching those three references **changed the entry's claims**, it didn't pad it.

## The knowledge gap was wrong, not just incomplete

`PMID:36805337` — hepatocyte-specific `Fitm2` knockout mice with elevated acyl-CoA, ER stress and liver injury *in vivo*.

My ER-stress node said "no cited source demonstrates ER stress", and the `KNOWLEDGE_GAP` was built on that premise. **The premise was false, and I'd written it as a considered epistemic caveat.** That's worse than an omission — an honest-sounding hedge asserting an absence of evidence that exists in a reference the PR itself shipped.

Rewritten from *"is ER stress the mediator at all"* — no longer open — to why the human tissue pattern differs from the mouse. And your framing was right that the mouse data make the gap **sharper**: tissue-specific knockouts produce liver injury and lipodystrophy, two things the human disease does not have, while the human phenotype lands on cochlea, neurons and skin, which no cited knockout has targeted. The failing tissues are close to disjoint.

## The lipodystrophy dissociation now has both halves

`PMID:30020828`: adipose-specific FIT2 knockout mice **do** get lipodystrophic features. Patients lacking FIT2 everywhere do not.

That turns the entry's headline finding from an absence into a genuine dissociation, and it's now the strongest constraint on any answer to the selectivity gap — the one tissue where mouse and human have been directly compared, and they disagree.

## The allelic series

`PMID:39974099`: one null plus one G100R hypomorph at **20% residual protein** gives hereditary spastic paraplegia, not this disease. Same gene, two diseases, two residual activities.

Recorded in `genetic.notes` and a second `KNOWLEDGE_GAP` as an open question rather than asserted as a continuum — a qualitative property of G100R beyond reduced abundance isn't excluded, and two families is thin for a genotype-phenotype rule. Flagged there that if it *is* a continuum, the FITM2 spastic paraplegia phenotype is a related entry rather than a subtype, and the boundary is quantitative.

## Findings 1 and 5 were the same mistake, and it's worth naming

**Growth Delay** cited the ascertainment sentence, which says nothing about growth. I had graded it `PARTIAL` and explained the mismatch in the `explanation` — treating the grade as a way of being *honest* about a poor fit.

Your sentence — "a `PARTIAL` grade does not repair an off-claim snippet" — is the correction, and it generalises. `supports` describes how the evidence bears on the claim; it is not a confidence dial for how well the quote matches. I was using it as the latter. Now cites `PMID:41113320`, which names growth failure directly.

**Motor Delay** was the same error in a different field: sole `MODEL_ORGANISM` support, quoting the *fly's* locomotor impairment, for a human phenotype — which CLAUDE.md forbids outright. Removed, since the human sources describe regression, which the entry already curates properly.

Two instances of using a grading field to make a bad evidence fit *legible* rather than fixing it.

## Also done

`GO:0010945` (coenzyme A diphosphatase activity) on the enzymatic node — you're right that the defining lesion was unbound. Intellectual disability added from `PMID:41113320`, with **no** frequency band because the source hedges with "often".

## Not done

The `diagnosis` section and the full seven-family variant catalogue. Both need content beyond the cached abstracts, and I'd rather add them with sources than approximate. Say the word if you'd prefer they land here.

On GeneReviews: I have network here and checked — no GeneReviews chapter for Siddiqi syndrome or FITM2, consistent with your expectation for a ~7-family disease.

## Validation on `9cec0f115`

```
just validate                 → schema + terms clean (GO:0010945 verified)
just count-verified-snippets  → 22/22 (was 17/17)
```

One root, no dangling targets. `Absence of Lipodystrophy` remains deliberately unreached — an absent finding shouldn't hang off a causal edge.

---

nlharris:
Round 2 pushed as `6ce86a53f`. All six items addressed.

**The CI blocker (item 1).** You were right about the mechanism and it is worth stating plainly: dropping "a novel" from the front of the quote is what turned an abstract sentence into a title sentence. Same paper, same claim, four fewer words, and the check flips. Replaced with `PMID:41113320`'s "generalized dystonia, motor regression, growth failure, and ichthyosis" — a sentence describing the disorder rather than the title of the paper that named it. `just check-title-snippets` passes clean against `origin/main`.

**Item 2 — the stale note.** Corrected rather than deleted, because the sentence records something real: it was accurate when the entry had two references and stopped being accurate the moment `PMID:41113320` was fetched. The note now says that, and there is a `treatments` section holding the supportive-management protocol from the cache line you pointed at. Cochlear implantation is in there too but graded `PARTIAL`, because the authors write that it "may improve functional outcomes" and neither sibling is described as having received one — proposed, not administered.

**Item 3 — variants.** `genetic[0].variants` now carries the six-family allele catalog, the Iranian allele, and the genotype-phenotype gradient as a separate record. The gradient is `PARTIAL` on the authors' own words ("may exhibit", "data remain limited"): with one or two families per allele class it is a pattern in a case series, not a correlation. The catalog record is also what now cites the compound-heterozygote claim in `genetic[0].notes` and `inheritance[0].description`, which was previously asserted with nothing behind it.

**Items 4 and 5.** Global developmental delay (`HP:0001263`) added and wired into the pathograph. `diagnosis` added — exome sequencing with Sanger segregation, and auditory brainstem response — plus four `differential_diagnoses`. The normal MRI is curated as a discriminating negative rather than an absence of information: in a child with progressive dystonia it is what excludes the leukodystrophies and NBIA.

**Item 6 — you were right, and for the right reason.** The mouse adipose-knockout datum was a good find in round 1 and I put it in the wrong place. Hanging a `SUPPORT` item on the `HP:0009125` record made a term-level query read "model-organism evidence supports lipodystrophy in Siddiqi syndrome", which is the inverse of the entry's argument. It now sits on the `Unbuffered Fatty Acyl-CoA and Failed Lipid Droplet Partitioning` node, where it evidences the mechanism, and the phenotype description says where it went and why.

**One thing I got wrong that your review did not catch, worth recording.** Both NCIT terms I first reached for in the `diagnosis` section were fabrications: `NCIT:C101293` is Next Generation Sequencing (not Whole Exome Sequencing) and `NCIT:C38031` is **Tearing** (not Auditory Brainstem Response Test). Term validation caught both — the second as a hard enum-membership error, the first only as a label mismatch, which is the softer of the two signals. The real terms are `NCIT:C101295` and `NCIT:C184949`, verified through OAK. I mention it because this is the same failure mode as the HGNC:14929/SIRT1 catch that opens this PR, except the fabricating party was me rather than the report.

**Not changed:** the node debundling suggestion. `PMID:30020828` does make the acyl-CoA-clearance vs droplet-partitioning distinction more load-bearing than it was, and you are right about that — but the evidence I have attaches to both arms jointly rather than separating them, so splitting the node would produce a second node I could not independently evidence. I would rather leave it bundled and visible than split it and have one half resting on an inference.

**Validation:** 33/33 snippets verified (was 22/22). `just validate`, `check-title-snippets`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length` all pass. Reachability: one root, no dangling targets, `Absence of Lipodystrophy` still deliberately unreached. `check-snippet-grading`'s `Metastatic_Ovarian_Cancer.yaml` failure is unchanged and still not ours.

---

PR #9942, opened 2026-08-28
curate: add CAPRIN1-related neurodevelopmental disorder (MONDO:0968945)

### LINKED ISSUE #9882: Curate neurodevelopmental disorder with language impairment, autism, and attention deficit-hyperactivity disorder (MONDO:0968945)
Curate a dismech entry for **neurodevelopmental disorder with language impairment, autism, and attention deficit-hyperactivity disorder** ([MONDO:0968945](https://monarchinitiative.org/MONDO:0968945)) — caused by haploinsufficiency of `CAPRIN1`, an RNA-binding protein that regulates transport and translation of dendritic mRNAs.

Preflight: MONDO:0968945 appears nowhere in `kb/` on `main`, and `CAPRIN1` appears in no existing entry's `genetic:` block. No open PR matches.

Worth watching during curation: the MONDO label is a list of three co-occurring features rather than a mechanism, which is the shape that tempts a curator into a phenotype-only entry. The curatable mechanism is local translational control at the synapse — the same axis as FMR1 — so the entry should reach the phenotypes from that node rather than stopping at a feature list.

Stub: `stubs/Neurodevelopmental_Disorder_With_Language_Impairment_Autism_And_Attention_Deficit-hyperactivity_Disorder.yaml`

nlharris:
Adds a Disease entry for CAPRIN1-related neurodevelopmental disorder — autosomal dominant, by haploinsufficiency of `CAPRIN1` (RNG105), an RNA-binding protein in neuronal RNA transport granules.

Closes #9882

## The frequencies are real here, and that is the thing to check

This is the only entry of the five in this batch that carries `frequency` values, because it is the only one whose source states a figure per phenotype. **Every band is bound to a quote stating that figure for that named phenotype:**

| Phenotype | Source figure | Band |
|---|---|---|
| Language impairment / speech delay | 100% | `OBLIGATE` |
| Intellectual disability | 83% | `VERY_FREQUENT` |
| ADHD | 82% | `VERY_FREQUENT` |
| Autism | 67% | `FREQUENT` |
| Respiratory problems | 50% | `FREQUENT` |
| Limb / skeletal anomalies | 50% | `FREQUENT` |
| Developmental delay | 42% | `FREQUENT` |
| Feeding difficulties, seizures, ophthalmologic | 33% each | `FREQUENT` |

This is the discipline the previous curation batch failed on: there, a band was taken from a number describing a *neighbouring* phenotype in the same sentence, and it verified cleanly while being wrong by a factor of seven. Worth checking me on it here.

**Twelve patients is a coarse denominator** — 33% is four people — and `notes` says so rather than letting the percentages imply precision they don't have.

## Haploinsufficiency is demonstrated, not inferred

Most entries infer the mechanism from variant class. This cohort measured it: patient lymphoblasts and fibroblasts show **monoallelic expression of the wild-type allele** with transcript and protein at about half. That observation is what rules out a dominant-negative product, and it's cited on the inheritance block rather than left as background.

## The mechanism is positional, not quantitative

Caprin-1 carries specific mRNAs into dendrites — and the transcripts that matter are those encoding **AMPA-receptor surface-expression regulators**. So the downstream failure is homeostatic AMPAR scaling and synaptic strength, not "less protein made". The entry models it that way deliberately.

## Two structural decisions worth reviewing

**The extraneural node is a sibling, not a descendant.** Respiratory, skeletal, feeding and ophthalmologic features occur in a third to a half of patients, and a cortical-neuron-development node cannot explain a limb anomaly. Rather than force an edge, they hang off their own node whose description states plainly that it asserts co-occurrence in a CAPRIN1 disease, not derivation from the neuronal mechanism. The only mechanistic basis cited for it is that CAPRIN1 is ubiquitous and also regulates proliferation/migration transcripts in other cell types — which licenses the node's existence and no specific organ claim.

**Three phenotypes use high-level HP terms** (`Abnormal respiratory system physiology`, `Abnormal skeletal morphology`, `Abnormality of the eye`) because the source says only "respiratory problems", "limb/skeletal anomalies", "ophthalmologic problems". Binding something specific would assert more than the cohort carries.

## The dose mismatch — a HUMAN_MODEL_MISMATCH, not a knowledge gap

The mRNA-localisation mechanism at the centre of this entry is established **only in a homozygous mouse deletion**. The human disease is haploinsufficiency. A null establishes what Caprin-1 does; it does not establish what half of it fails to do, and this disease is entirely a claim about half.

The CRISPR-edited `CAPRIN1+/-` iPSC neurons *are* at the right dose — and were characterised for morphology, translation, calcium signalling, oxidative stress and electrical activity, but are **not reported to have been assayed for dendritic mRNA localisation**. So the causal chain passes through a step demonstrated at a dose the disease doesn't have. A `proposed_experiment` names the assay that would close it, with both outcomes stated.

## One confabulated quote, flagged and excluded

This was the only report of the five carrying a confabulation. Its own validation reported:

```
Quote not found in PMID:35979925: 'The majority of reported pathogenic mutations in
the CAPRIN1 gene result in decreased protein levels and haploinsufficiency...'
```

That sentence is not in the cached abstract. It is used nowhere in the entry and is recorded in `notes` so it isn't re-derived from the report by the next curator. (The report also flagged `PMID:36279435` as off-topic; not cited either.)

## Gene ID

`hgnc:6743`, verified — correct in this report. Two of the five reports in this batch gave wrong causal-gene IDs (#9894, #9928), so this is now checked by default rather than on suspicion.

## Validation

`just validate`, `just count-verified-snippets` (24/24), `just check-entity-refs`, `just check-duplicate-keys` — all pass. One root, no dangling targets, no unreached phenotype.

---

### DIFF
--- a/history/disorders/CAPRIN1_Related_Neurodevelopmental_Disorder/2026-08-29T003852Z-claude-code-a1274b.yaml
+++ b/history/disorders/CAPRIN1_Related_Neurodevelopmental_Disorder/2026-08-29T003852Z-claude-code-a1274b.yaml
@@ -1,0 +1,38 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: CAPRIN1_Related_Neurodevelopmental_Disorder
+  path: kb/disorders/CAPRIN1_Related_Neurodevelopmental_Disorder.yaml
+session:
+  id: 2026-08-29T003852Z-claude-code-a1274b
+  timestamp: '2026-08-29T00:38:52Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9942
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - genetic
+  - phenotypes
+  - evidence
+  summary: 'Review round 2: two snippets that verified but did not support their claims'
+  details: |
+    Second review round on PR #9942. Both items are the same class of defect and neither is a substring failure: a quote that verifies exactly against its cited source while not supporting the claim it is attached to.
+
+    The PMID:35977029 snippet stopped one word before the gene name that fixes its subject. "a new neurodevelopmental disorder (NDD) with common features of language problems, intellectual disability, and behavioral issues caused by de novo likely gene-disruptive variants in" continues "UBAP2L". Those three features are Jia et al's description of the UBAP2L disorder, not CAPRIN1's, and the explanation built on their resemblance to this entry's three most frequent phenotypes - a coincidence between two different disorders. Re-quoted to the sentence that is actually about CAPRIN1, regraded from SUPPORT/HUMAN_CLINICAL to PARTIAL/IN_VITRO, and the explanation rewritten: this paper is functional corroboration of the gene, not a clinical replication cohort.
+
+    The PMID:20861386 NKA-cargo item was attached to Disrupted Cortical Neuronal Development while its explanation described the claim of Impaired Dendritic mRNA Localisation, one node upstream. Moved to the node its explanation was describing.
+
+    Also taken: renamed the respiratory phenotype from Abnormal Respiratory System Physiology to Respiratory Problems, since round 1 moved its binding to HP:0002086 specifically to avoid asserting a physiological abnormality while leaving the name asserting exactly that; updated both downstream refs. Replaced attaches_to animal_models#Mouse with the two model names, since the rationale now turns on the distinction between them and both models share a species. Marked the neonatal-lethality quote as the secondary citation it is.
+
+    Note for the record: the truncated quote was supplied by the round-1 review and I used it without re-reading the surrounding sentence in the cache. It verified, so nothing objected - which is the failure this entry's own notes describe and which is now filed as #9970.
+
+    Validation: 32/32 snippets verified, validate, validate-terms, check-entity-refs all pass.
--- a/kb/disorders/CAPRIN1_Related_Neurodevelopmental_Disorder.yaml
+++ b/kb/disorders/CAPRIN1_Related_Neurodevelopmental_Disorder.yaml
@@ -175,6 +175,18 @@ pathophysiology:
           Names the transcript class, which is what connects a transport defect
           to a specific synaptic consequence rather than to translation in
           general.
+      - reference: PMID:20861386
+        reference_title: "RNG105 deficiency impairs the dendritic localization of mRNAs for Na+/K+ ATPase subunit isoforms and leads to the degeneration of neuronal networks."
+        supports: SUPPORT
+        evidence_source: MODEL_ORGANISM
+        snippet: >-
+          RNG105 knock-out in mice reduces the dendritic localization of mRNAs
+          for Na+/K+ ATPase (NKA) subunit isoforms
+        explanation: >-
+          A second, independently identified cargo class. It matters here
+          because of what accompanies it: dendritic NKA function is lost while
+          somatic NKA function is not. That is the shape of the claim this node
+          makes - the protein is made, and made in the wrong place.
   - name: Attenuated Homeostatic AMPAR Scaling and Reduced Synaptic Strength
     biological_scale: CELLULAR
     description: >-
@@ -244,18 +256,6 @@ pathophysiology:
           recorded on this node rather than as separate mechanism nodes because
           nothing cited establishes their order relative to the morphological
           changes.
-      - reference: PMID:20861386
-        reference_title: "RNG105 deficiency impairs the dendritic localization of mRNAs for Na+/K+ ATPase subunit isoforms and leads to the degeneration of neuronal networks."
-        supports: SUPPORT
-        evidence_source: MODEL_ORGANISM
-        snippet: >-
-          RNG105 knock-out in mice reduces the dendritic localization of mRNAs
-          for Na+/K+ ATPase (NKA) subunit isoforms
-        explanation: >-
-          A second, independently identified cargo class, and one that makes the
-          positional claim concrete: the loss of dendritic NKA function occurs
-          without affecting NKA in the soma. That is the shape of the mechanism
-          this node asserts - protein is made, and made in the wrong place.
   - name: Extraneural Consequences of Reduced Caprin-1
     biological_scale: ORGANISM
     description: >-
@@ -284,7 +284,7 @@ pathophysiology:
       downstream of it, because routing a limb anomaly through cortical neuron
       development would be false.
     downstream:
-      - target: phenotypes#Abnormal Respiratory System Physiology
+      - target: phenotypes#Respiratory Problems
       - target: phenotypes#Abnormal Skeletal Morphology
       - target: phenotypes#Feeding Difficulties
       - target: phenotypes#Abnormality of the Eye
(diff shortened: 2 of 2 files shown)


nlharris:
Round 1 pushed as `869e943c2`. All four important items and five of the seven suggestions. Snippets 24/24 → 32/32.

**Item 1 — GO term.** Fixed to `GO:0008298` (intracellular mRNA localization). You were right that it contradicted the node it annotated, and that is a worse defect than an imprecise term: the node's whole argument is that protein supply is fine and position is not.

**Item 2 — the heterozygous mouse.** You are right and the claim was wrong twice: the entry asserted the iPSC neurons were the only model at the human gene dose, when `PMID:26865403` is a comprehensive behavioural battery on `Rng105+/-` mice, surfaced in the research report I shipped *and* restated in the intro of a paper I had already cached. Added as a second `animal_models` record linked to `phenotypes#Autism` at `PARTIALLY_RECAPITULATES` / `MODERATE` — partial because reduced sociality in a home cage is a behavioural analogue of one ASD domain rather than the human diagnosis, and the authors' own phrase is "ASD-like behavior".

The `HUMAN_MODEL_MISMATCH` is rewritten rather than softened, because the gap changed shape. It is no longer "half dosage has never been modelled" — it has, and it produces a phenotype. It is that **neither** dose-matched model has been assayed for the molecular step the causal chain runs through: the mouse work is behavioural throughout, and the `CAPRIN1+/-` neurons were characterised for morphology, translation, calcium, oxidative stress and electrical activity but not for dendritic mRNA localisation. Same missing experiment, better-stated reason.

**Item 3 — extraneural prose.** Both sources consumed. The node now says Caprin-1 has demonstrated organ-specific requirements outside the nervous system — neonatal respiratory lethality in the null, progressive hearing loss on targeted inner-ear deletion — and that both are *complete loss*, with nothing showing them failing at half dosage. That is the defensible version you proposed, and it is stronger than the original claim rather than weaker. The respiratory item is graded `SUPPORT` for the node's existence and explicitly not for the respiratory phenotype: neonatal lethality in a homozygous null is not the respiratory morbidity of a surviving heterozygous patient.

**Item 4 — replication added.** `PMID:35977029`, and worth noting it replicates from a different direction: that cohort reached CAPRIN1 through stress-granule assembly regulators rather than by CAPRIN1-first ascertainment.

---

**One correction to my own PR body, and it is the most interesting thing in this round.** I reported the "The majority of reported pathogenic mutations..." sentence as a confabulated quote — the only one across the five reports in this batch. That was wrong.

**The sentence is real and verbatim. It belongs to `PMID:41859620`,** a 2026 case report that was sitting in `references_cache/` the whole time. The report misattributed it to `PMID:35979925`.

That is a different defect from fabrication and it needs a different defence. A confabulated quote fails any substring check against its cited source — which is exactly what happened here, and why it got caught. A **misattributed** quote fails that check only by the accident of the report having named the wrong paper. Had the report cited `PMID:41859620`, the quote would have verified exactly, the paper is real and on-topic, and nothing in this pipeline would have flagged anything — because nothing verifies that a quoted sentence supports the claim it is attached to. Quarantining it was still the right call; the reason I recorded for it was not. Both the top-level `notes` and `genetic[0].notes` now say so.

That also means my earlier statement that this was the batch's only confabulation should be read as: **zero confabulations across five OpenScientist reports, one misattribution.** Better than I reported, and a narrower failure mode than I claimed.

**Suggestions taken:** mouse genotype precision (5) — `Camk2a-Cre;Rng105^fl/fl` plus constitutive nulls for the primary-neuron work; `CL:0000679` (8); `HP:0002086` (9), with the reasoning recorded — "respiratory problems" does not distinguish a functional abnormality from a structural one, and `HP:0002795` asserted the former; the Na⁺/K⁺-ATPase cargo class (10), which makes the positional claim concrete since dendritic NKA function is lost while somatic NKA function is not; and a `diagnosis` section (11).

**Suggestion 7, partly.** `PMID:41859620` is cited, but cerebellar ataxia, hearing loss and dysmorphic features get **no phenotype records**. They appear in a background sentence of a single case report summarising the literature, with no denominator — so they sit on the genetic block at `PARTIAL` with the reason stated. I'd rather under-curate a spectrum expansion than give three features the same standing as ten that came with percentages. Happy to add them if you disagree.

**Suggestion 6 (genetic depth)** is not done. Variant classes and de novo/inherited split would need figures I can only source from the research report, not from a citable publication — the same call I made on #9937's gnomAD constraint. Flagging rather than importing.

**Validation:** `just validate` (schema + terms), 32/32 snippets, `check-entity-refs`, `check-duplicate-keys`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-snippet-grading` all pass. One root, no dangling targets, no unreached phenotype.

---

nlharris:
Round 2 pushed as `c49f9c9d8` (rebased onto current `main`, so the SHA is new). Both items fixed, plus all four suggestions.

**The critical one is the better finding of the two, and thank you for owning where it came from.** The quote stopped one word before `UBAP2L`. Those three features — language problems, intellectual disability, behavioural issues — are Jia et al.'s description of a *different* disorder, and the resemblance to this entry's three most frequent phenotypes, which my explanation leaned on, is a coincidence between two NDDs.

What makes it worth dwelling on: **every gate in this repository passed it.** It was an exact substring of a real, on-topic, correctly-titled paper, cited with the right PMID. Snippet verification, title checking, length, grading — all green. The only thing that could catch it was somebody reading the sentence to the end.

I've re-quoted to the sentence that is actually about CAPRIN1 (`"such as CAPRIN1, disrupt the interaction of the core SG network"`) and regraded it `SUPPORT`/`HUMAN_CLINICAL` → `PARTIAL`/`IN_VITRO`, because you're right that this is functional corroboration of the gene, not a clinical replication cohort — so it cannot carry the "a new entity should cite its replication" argument as written. The explanation now says what it is and records what the earlier quote got wrong, so the next reader of this file doesn't re-derive it.

This is also, unhappily, a demonstration of the thing this entry's own notes had just finished lecturing about — and it is now filed as [#9970](https://github.com/monarch-initiative/dismech/issues/9970), which I opened before your review landed, on the misattribution finding from round 1. Nothing in the stack verifies that a verified sentence supports the claim it is attached to. Two independent instances in one entry in two rounds.

**Item 2.** Moved. The explanation was describing `Impaired Dendritic mRNA Localisation` and the item was sitting on `Disrupted Cortical Neuronal Development` — I wrote the explanation for the node I meant and attached it to the node I was editing.

**Suggestion 3 is the one I'm most glad you caught.** Renamed to `Respiratory Problems`, both `downstream` refs updated. Round 1 moved the binding to `HP:0002086` *specifically* to avoid asserting a physiological abnormality, wrote a description explaining that choice, and left the `name` — the thing that renders and that the edges point at — asserting exactly the thing it had just declined to assert. A term fix that leaves the label contradicting it isn't a fix.

**Suggestion 4:** both models named in `attaches_to` rather than relying on the `species` fallback. You're right that it resolved but stopped saying what it means once the rationale started turning on the difference between them.

**Suggestion 5:** kept, but the `explanation` now states it is a secondary citation restating Shiina et al. 2010. I checked `PMID:20861386` on the chance it was that primary — it is Shiina et al. 2010 and it is in this PR, but its text contains no neonatal or respiratory content, so it is a different paper of theirs. I'd rather label the citation honestly than guess at a PMID for the primary.

**Suggestion 6:** rebased onto `origin/main` (52 commits). `check-snippet-grading` is green now — `Metastatic_Ovarian_Cancer.yaml` no longer exists in that form. Post-rebase diff reviewed: 19 files, all this PR's own, no unrelated deletions or protected-path churn.

**Validation on the rebased head:** `just validate` (schema + terms + references, 32/32 snippets), `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-snippet-grading`, `check-reference-titles`, `check-entity-refs`, `check-duplicate-keys` — all pass.

---

PR #9946, opened 2026-08-28
Curate Cushing disease due to pituitary adenoma (MONDO:0009050)

### LINKED ISSUE #9939: Curate Cushing disease due to pituitary adenoma (MONDO:0009050)
Curate a dismech entry for **Cushing disease due to pituitary adenoma** ([MONDO:0009050](https://monarchinitiative.org/MONDO:0009050)) — ACTH-dependent hypercortisolism driven by a corticotroph pituitary adenoma, most often carrying a somatic `USP8` variant that blocks EGFR deubiquitination and degradation, sustaining POMC transcription and defeating glucocorticoid negative feedback.

Provisional read: **DISEASE**. This is not covered today. `kb/disorders/Cushings_Syndrome.yaml` is the umbrella entry (MONDO:0018912, all causes — adrenal, ectopic, exogenous) and carries no subtypes; `kb/disorders/USP8-related_pituitary_adenoma_4.yaml` is a gene-level entry with no `disease_term` at all. Cushing disease is the specific pituitary-dependent entity and has its own conserved pathograph.

On the neoplasm caution in the claim workflow: a corticotroph adenoma is a benign, WHO-defined PitNET with a settled single identity, and the disease as clinicians use the term is the hormonal syndrome rather than an oncology entity with a histologic granularity ladder. The lump/split problem behind #7082 does not really bite here, but flagging the judgement call explicitly.

Likely subtypes to model as `has_subtypes`: USP8-mutant, USP48-mutant, and `TP53`/`ATRX`-mutant aggressive corticotroph tumours.

Stub: `stubs/Cushing_Disease_Due_To_Pituitary_Adenoma.yaml`

aleixpuigb:
New disorder entry for Cushing disease — ACTH-dependent hypercortisolism caused specifically by a corticotroph pituitary adenoma.

## Scope, and why it is not already covered

Three existing entries are adjacent and none covers this:

- `kb/disorders/Cushings_Syndrome.yaml` is the umbrella (MONDO:0018912, all causes — adrenal, ectopic, exogenous) and carries no subtypes.
- `kb/disorders/USP8-related_pituitary_adenoma_4.yaml` is a gene-level entry with no `disease_term` at all.
- `kb/disorders/Pituitary_Tumor.yaml` is the tumour class, not the endocrine syndrome.

On the neoplasm caution in the claim workflow: a corticotroph adenoma is a benign WHO-defined PitNET with a single settled identity, and the disease as clinicians use the term is the hormonal syndrome. The granularity ladder in design decisions §3a is about malignancies with a histologic ladder and does not really engage here — but it is a judgement call and I would rather flag it than have it discovered in review.

## The mechanism

Somatic `USP8` variants cluster in the 14-3-3 binding motif, releasing the protease from inhibitory binding. The cleaved hyperactive enzyme deubiquitinates EGFR and rescues it from lysosomal degradation, so the corticotroph gets continuous EGF signalling from a normal amount of ligand — the defect is in receptor disposal, not ligand supply. In USP8 wild-type tumours, `USP48` and `BRAF` V600E reach the same POMC promoter from a different direction, and the three drivers are close to mutually exclusive across a 169-tumour series. Six nodes run from the driver mutation through EGFR signalling, POMC transcription, loss of glucocorticoid negative feedback and adrenocortical stimulation to systemic glucocorticoid toxicity.

The entry is careful about where the morbidity sits. It is not the tumour, which is usually a small benign microadenoma. Bilateral adrenalectomy makes the point: it cures the cortisol excess while leaving the lesion entirely in place.

## Deep research

One OpenScientist report, at `research/Cushing_Disease-deep-research-openscientist.md`. Reference validation: 46/46 citations verified. Term validation reported `needs_review` with two mislabelled terms — the important one being **`MONDO:0005479` offered as the Cushing disease term when that CURIE is in fact `atrial tachycardia`**. This entry binds `MONDO:0009050`.

The report earned its place. It added the mortality burden, the perioperative VTE risk, the USP8 prevalence meta-analysis and the AtT-20 model — and it **corrected a claim I had already written**. My first draft's knowledge-gap discussion said no association between driver genotype and outcome had been shown. A meta-analysis of 2,171 cases (PMID:40392165) shows USP8-variant tumours are significantly *more* likely to achieve postoperative remission (OR 1.76) and significantly *more* likely to recur (OR 2.38). Those point in opposite directions over different time horizons, which is a more interesting finding than the absence I had asserted, and the discussion now says so.

## Things I deliberately did not do

- **`NR3C1` is `relationship_type: MODIFIER`, not causative.** It removes feedback rather than conferring the proliferative advantage the other three drivers do, and the reported cases are few.
- **Two treatment evidence items are `supports: PARTIAL` with the reason stated in the `explanation`.** The Pituitary Society consensus abstract establishes that treatment selection is under consensus recommendation; it does not carry a pasireotide-specific or adrenalectomy-specific outcome. Rather than overstate those snippets I recorded what they do and do not support.
- **One snippet is deliberately short.** The source interrupts "glucocorticoid receptor gene NR3C1, a critical negative regulator of ACTH production" with inline citation markers (`NR3C17,8,11`), so the quote stops at the gene name and the `explanation` says why.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed |
| snippets | 45/45 verified against cached references |
| `just check-duplicate-keys` / `check-entity-refs` | passed |
| `check-environmental-evidence` | passed |

Stub `stubs/Cushing_Disease_Due_To_Pituitary_Adenoma.yaml` deleted; history record added.

Closes #9939

---
_Generated by [Claude Code](https://claude.ai/code/session_018e5RSJL5xVDoFDT3GASzMh)_


aleixpuigb:
CI failure was mine and is fixed in 748d35de7.

`check_snippet_length` rejected this, correctly:

```
kb/disorders/Cushing_Disease.yaml:genetic[1].evidence[0].snippet: 2 word(s): '21/91 subjects'
```

I had quoted the parenthetical count out of the abstract rather than a sentence. A bare `21/91 subjects` cannot support or refute anything on its own, which is exactly what that check exists to stop. Replaced with the sentence that actually reports the finding:

> Combining the data from whole-exome and targeted sequencing, which included a total of 91 samples, we found that 16.5% of the cases (15/91) had BRAF V600E mutations and 23.1% of the cases (21/91) harbored USP48 mutations.

That is a better citation anyway — it carries the 23.1% denominator-aware figure rather than a raw ratio, and it comes from the combined analysis rather than the abstract's summary line.

## How I missed it before pushing

Worth recording because it is a repeatable trap. The whole-KB ungated checks — `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` — are whole-KB *relative to the working tree they run in*. I ran them once, in a different worktree, and read the pass as covering the whole run. It only covered the entry that worktree contained.

Re-running per worktree found a second instance of the same defect in #9948 (a 3-word `non-palpable vas deferens`), which is now fixed there too — before CI saw it.

## Validation on the new head

| Check | Result |
|---|---|
| `check_snippet_length --against-ref origin/main` | passed (reproduced the failure first, then the fix) |
| `just validate` | passed, 45/45 snippets verified |
| `check-folded-hyphens`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` | passed |
| `check-duplicate-keys`, `check-entity-refs` | passed |

All four gates were re-run in each of the four worktrees this time, not just one.

## One thing I did not fix

`uvx ruff check .` reports `E402` in `tests/test_ci_batch_validation.py` locally. That file is untouched by this PR, is pre-existing on `main` (last modified by #9822), and the ruff step in this PR's own CI run passed — local `uvx` resolves ruff 0.15.8 while CI's resolution evidently differs. Not this PR's failure and not something to widen this PR with, so I have left it alone.

---

aleixpuigb:
Everything addressed. Citations 9 → 17; snippets 45 → 66.

The CRITICAL snippet-length failure was already fixed in 748d35de7 before your review landed — same diagnosis, same fix. Your note that it wasn't in my validation table was the useful part: the reason it escaped is that I ran the whole-KB gates once, in a *different* worktree, and read the pass as covering the run. Re-running them per worktree found a second instance of the same defect in #9948 before CI reached it.

| Item | What I did |
|---|---|
| MRI on a laboratory-procedure term | `NCIT:C16809`, matching the convention you pointed to. |
| Pasireotide on a `PARTIAL` guideline snippet | Replaced with `PMID:31465533` — response rates, sustained-efficacy conclusion, and the 39.5% hyperglycaemia figure, which follows from SSTR5 agonism rather than being incidental. |
| Osilodrostat missing LINC 4 | Added `PMID:35325149`. |
| No `clinical_trials` | Added all three NCTs. |
| Prevalence from a secondary citation | Replaced with `PMID:30799512`. |
| Recurrence absent | Added, with the tumour-size stratification (19% vs 45%). |
| Hyperpigmentation, menstrual irregularity | Added. |
| Infection susceptibility | Added, cited to both the comorbidity list and the mortality study. |
| Hypertension / diabetes `PARTIAL` | Now also carry counted frequencies (70.8%, 48.0%) from `PMID:42572239`. |
| `GO:0008211` loose fit | Dropped. |
| `biochemical` empty | Added nadir cortisol. |

## The suggestion that changed the entry most

Your last one. You flagged that `PMID:25675982` reports recurrence "unrelated to the USP8 mutational status", disagreeing with the meta-analytic OR 2.38 the discussion was built on. Following that up, `PMID:40257708` — which I was adding anyway for the recurrence rate — says the same thing more directly: *"USP8 and USP48 mutational status ... were not associated with recurrence"*, from a cohort that looked for exactly that association.

So it is **two** primary series against the pooled estimate, not one. The discussion now names both as `REFUTE` evidence and says that the OR is uncorroborated by either primary series that tested it — which is a stronger reason to treat USP8 as prognostically unsettled than the between-study heterogeneity I had been leaning on.

That is the second time in this PR the sequence has run the same way: I wrote that no genotype-outcome association had been shown, the deep research corrected me with the meta-analysis, and your review corrected the correction. The discussion is better for both passes and I would not have got here alone.

## Prevalence — why the replacement matters beyond the number

I had flagged my own source as weak and kept it. You're right that replacing it was available. The thing that makes `PMID:30799512` better is not the figure but the validation: of 1,317 patients carrying a Cushing diagnostic code, only 41% had confirmed Cushing disease. An unvalidated code-based count overstates this roughly 2.5-fold, so the `notes` now records the validation yield alongside the rate.

## One thing I did differently

For the three `clinical_trials` I first used the registered trial **titles** as snippets. Two failed verification, and they would also have been exactly the `check-title-snippets` defect. Replaced with each trial's registered objective from the record body.

## Validation

`just validate` passes, 66/66 snippets verified. `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `check-environmental-evidence`, `check-entity-refs`, `check-duplicate-keys`, `validate-history` all pass, run in this worktree.

---

PR #9947, opened 2026-08-28
Curate bone Paget disease (MONDO:0005382)

### LINKED ISSUE #9940: Curate bone Paget disease (MONDO:0005382)
Curate a dismech entry for **bone Paget disease** ([MONDO:0005382](https://monarchinitiative.org/MONDO:0005382)) — focal disorganized bone remodeling in which giant, hypernuclear, hyperactive osteoclasts drive accelerated resorption, followed by disordered compensatory osteoblastic deposition of mechanically inferior woven bone. `SQSTM1`/p62 UBA-domain variants are the commonest genetic cause and act through sustained NF-kB signaling in the osteoclast lineage.

Provisional read: **DISEASE**, with a boundary to draw carefully. The stub reports 5 MONDO descendants. Four of them (`Paget disease of bone 2, early-onset`, `3`, `4`, `6`) are genetic loci for the same classic adult disease and belong in `has_subtypes`. The fifth, `juvenile Paget disease` (MONDO:0009394, `TNFRSF11B`/osteoprotegerin deficiency), is a genuinely different disease — generalized rather than focal, presenting in infancy — and should stay out of this entry and out of scope for this claim. It keeps its own stub.

Not curated today: only `Morgagni-Stewart-Morel_Syndrome` matched a text search for "osteitis deformans", and the `Inclusion_Body_Myopathy_with_Paget_Disease_of_Bone_and_Frontotemporal_Dementia` / `VCP-associated_Multisystem_Proteinopathy` entries cover the VCP multisystem proteinopathy, not classic PDB.

Stub: `stubs/Bone_Paget_Disease.yaml`

aleixpuigb:
New disorder entry for classic adult-onset Paget disease of bone.

## Where the boundary is drawn

MONDO puts five descendants under MONDO:0005382 and they do not all belong in one entry. PDB2, PDB3, PDB4 and PDB6 are genetic loci for the same disease and are handled as `has_subtypes` (only those with evidence in the reviewed literature are recorded). **Juvenile Paget disease (MONDO:0009394) is excluded** — it is autosomal recessive, usually from biallelic loss of `TNFRSF11B`/osteoprotegerin, presents in infancy, and is generalised rather than focal. It is a different disease that shares a pathway, and it keeps its own curation stub.

## The mechanism, and two things that make it unusual

Six nodes run from the SQSTM1 UBA-domain variant through osteoclast precursor hypersensitivity to RANKL, giant hypernucleated osteoclasts, focal accelerated resorption and disorganised compensatory formation, to expanded mechanically inferior bone. The osteoblast is behaving normally throughout — it is coupling to resorption as it should — which is why this is properly an osteoclast-driven disease.

Two features are worth a reviewer's attention because the entry is built around them:

**It is disappearing.** UK standardised incidence fell from 0.75 to 0.20 per 10,000 person-years between 1999 and 2015. That is far too fast to be genetic. Together with the focality of lesions and incomplete penetrance within SQSTM1 families, it is the case for an environmental co-factor — which is modelled as an `environmental:` entry with `influences_mechanisms`, deliberately **without** an `exposure_term`, because the exposure is not identified and binding a speculative ECTO term (paramyxovirus, coal smoke, livestock contact) would assert something the evidence does not support. A `notes:` line says so explicitly.

**Suppressing bone turnover does not help.** A single 5 mg zoledronic acid infusion gives one relapse in 152 patients over 6.5 years. But PRISM-EZ randomised 502 patients to normalising turnover versus treating symptoms and found no reduction in fracture, no fewer orthopaedic procedures and no better quality of life — with the point estimates moving the wrong way. The entry therefore carries a treatment-strategy record separate from the agent record, and a discussion asking why.

## Deep research

One OpenScientist report, at `research/Paget_Disease_of_Bone-deep-research-openscientist.md`. It reported `needs_review` on both axes and both flags mattered:

- **Three unsupported quotes** (PMID:38963497, PMID:33768371, PMID:21515589). No quoted material from those three was reused anywhere in the entry.
- **`HP:0002750` mislabelled.** The report calls it "Abnormal long bone morphology"; HPO calls it *Delayed skeletal maturation*. The deformity phenotype is bound to `HP:0006487` (Bowing of the long bones) instead.

A second term error the report's own validation did **not** catch was found during curation: the report offers `UBERON:0000033` for skull, which is `head`. This entry uses `UBERON:0003128` (cranium). I have written that up separately — the report's label check only fires when a CURIE has a parseable label next to it, so the same wrong CURIE can be caught in one report and pass silently in another.

## Things worth flagging

- **The prevalence record is an envelope, not an estimate.** The source gives 1.5%–8.3%, which spans populations with genuinely different risk, so a single worldwide figure is not really meaningful. The `notes:` says to read it as an upper-bound envelope.
- **Sarcoma risk is recorded as not predictable from disease burden**, because the cited series specifically found no correlation with number of bones involved or disease duration. That negative result seemed more useful to a curator than the headline incidence.
- **Guessed reference titles were caught by validation.** Six top-level `references[].title` values I wrote from memory were wrong and `just validate` rejected them; all titles now come from the cache.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed |
| snippets | 38/38 verified against cached references |
| `just check-duplicate-keys` / `check-entity-refs` | passed |
| `check-environmental-evidence` | passed |

Stub `stubs/Bone_Paget_Disease.yaml` deleted; history record added.

Closes #9940

---
_Generated by [Claude Code](https://claude.ai/code/session_018e5RSJL5xVDoFDT3GASzMh)_


aleixpuigb:
All seven blocking items addressed plus three of the suggestions. Citations go from 9 to 18 of the references this PR already had cached; snippets 38 → 52.

| # | Item | What I did |
|---|---|---|
| 1 | Wrong prevalence band | Fixed. You're right — the schema defines `BAND_1_5_PER_10000` as 10–99 per 100,000 and the rate is 2.0. Now `BAND_1_9_PER_100000`. |
| 2 | Subtype MONDO bindings | Bound `MONDO:0011183` and `MONDO:0014792` as `subtype_term`. |
| 3 | P394L mouse absent | Added, with three `modeled_mechanisms` links, readouts, and the zoledronic-acid prevention result from `PMID:30154079`. |
| 4 | Mouse contradicts the environmental claim | Softened "required" to "strongly implicated" **and** added a `HUMAN_MODEL_MISMATCH` discussion. |
| 5 | ALP frequency | `VERY_FREQUENT` → `FREQUENT`, with the reason in the description. |
| 6 | Radiography missing | Added as the first `diagnosis` entry, citing both guideline sources. |
| 7 | Paramyxovirus discussion unevidenced | Added four evidence items — one `SUPPORT`, three `REFUTE`. |
| 🔵 | `CL:0000779` | Bound on the hypernucleated node. |
| 🔵 | Neridronate | Added, including that its response decays at 24–36 months. |
| 🔵 | History `links.issues` | Set. |

## The thing worth saying back to you

**I had the P394L mouse cached and deliberately did not use it, for a bad reason.** The OpenScientist report's own reference validation listed `PMID:21515589` among three "unsupported quotes", and I applied the skill's rule — don't reuse quoted material from those — by dropping the paper entirely.

That was the wrong inference. "Unsupported quote" means *the report's rendering* of the quote failed to match the source, not that the source is unusable. The report had truncated the sentence with an ellipsis. I checked all three of your quotes against `references_cache/PMID_21515589.md` before using them and every one matches verbatim, including the "in the absence of an additional trigger" line.

So the rule cost me the single most informative model in the disease. The correct reading is that a flagged quote means *go to the cache and quote it yourself* — which is strictly more work than ignoring the paper, which is presumably why I didn't.

## On item 4, which you said you'd be glad to see decided

I did both, and I think the softening is the smaller half. The mouse is genuinely awkward for the environmental argument and the honest thing is not to pick a winner, so the discussion lays out three ways the disagreement could resolve — the mouse sitting above a penetrance threshold most human carriers sit below; the incidence decline being partly ascertainment; or the mouse modelling the SQSTM1 minority while the environmental factor acts on the SQSTM1-negative majority — and says nothing published distinguishes them. The proposed experiment tests penetrance in heterozygotes specifically, since that is the genotype matching human carriers, and it names the outcome that would show the model is saturated and cannot answer the question at all.

## Not done

**`histopathology:`** — there is no `Histopathology` class in the schema, and I did not want to guess at what the `histopathology` slot's actual range expects on a PR that already has a lot in it. The mosaic pattern and nuclear inclusions remain in `pathophysiology` prose and are now also in the mouse model's evidence. Happy to add it in a follow-up if you can point me at the right class.

**VCP / multisystem proteinopathy in the scope note** — not added. `VCP-associated_Multisystem_Proteinopathy` and `Inclusion_Body_Myopathy_with_Paget_Disease_of_Bone_and_Frontotemporal_Dementia` both already exist as entries, so the exclusion seems to me already settled by the KB rather than something this entry needs to argue. Say the word and I'll add a sentence.

**"osteoclast precursor" bound to `CL:0000092`** — left as is, for the reason you gave: it is a KB-wide pattern rather than something introduced here, and changing it in one entry would make this file inconsistent with the rest. Worth a separate issue if it should change.

**Hearing impairment description** — corrected the drift you spotted; it now says the cohort figure rather than implying a skull-disease denominator.

## Validation

`just validate` passes, 52/52 snippets verified. `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `check-environmental-evidence`, `check-entity-refs`, `check-duplicate-keys` and `validate-history` all pass, run in this worktree specifically.

---

PR #9954, opened 2026-08-28
curate: Add Myocarditis (MONDO:0004496) as a root with aetiological subtypes

### LINKED ISSUE #9040: Curate myocarditis (MONDO:0004496)
Curate a dismech entry for **myocarditis** ([MONDO:0004496](https://monarchinitiative.org/MONDO:0004496)) — inflammatory injury of the myocardium in which an initial insult (most often viral) is followed by an immune-mediated phase and, in a subset, progression to dilated cardiomyopathy; a major cause of sudden cardiac death in the young.

This is a `CURATE_ROOT_WITH_SUBTYPES` target: model the root with `has_subtypes` over aetiological/histological forms — viral (enterovirus/coxsackie B, parvovirus B19, HHV-6, SARS-CoV-2), giant cell myocarditis, eosinophilic/hypersensitivity myocarditis, immune checkpoint inhibitor-associated myocarditis, and lymphocytic myocarditis. Several existing entries already reference myocarditis as a manifestation or differential (`Lyme_Disease.yaml`, `Fabry_Disease.yaml`, `Naxos_disease.yaml`, `Arrhythmogenic_Right_Ventricular_Cardiomyopathy.yaml`) and should link to the new root rather than re-deriving it. Strong candidate for `conforms_to` against `cardiomyopathy_maladaptive_remodeling`; the checkpoint-inhibitor form is a treatment-toxicity conformer for `immune_checkpoint_blockade`.

Source: top of [priority dashboard](https://dismech.monarchinitiative.org/dashboard/priority.html) (`CURATE_ROOT_WITH_SUBTYPES`, score 72.5).

Tracker: part of #1079.

sierra-moxon:
Adds a new disorder entry for myocarditis (MONDO:0004496), curated as a root with subtypes.

Closes #9040

## What this is

Myocarditis is curated as a **Disease**, not a grouping. The aetiological forms genuinely share one pathograph — an insult to the myocardium, innate amplification, adaptive T-cell-mediated myocyte injury, necrosis and contractile failure, then either resolution or fibrotic remodelling into inflammatory dilated cardiomyopathy, with a separate arrhythmogenesis arm. What differs between the forms is mostly the trigger. That is what `has_subtypes` is for.

Five subtypes, following the issue: **Viral** (MONDO:0023161), **Lymphocytic**, **Giant Cell** (MONDO:0023232), **Eosinophilic**, and **ICI-Associated**. Only two of those exist in MONDO; the other three are carried with no `subtype_term`. The set deliberately mixes an aetiological axis with a histological one, because that is how the endomyocardial biopsy report and the clinical literature are actually organised — a viral case is usually lymphocytic on histology, so the two overlap rather than partition. That is stated in the entry's `notes` so a later reviewer does not read it as an accident.

## The one place I did not follow the issue

The issue suggests the checkpoint-inhibitor form is a treatment-toxicity conformer for `immune_checkpoint_blockade`. **It is not, as that module currently stands.** The module has five nodes and all of them describe anti-tumour efficacy: neoantigen generation, anti-tumour T cell response, adaptive immune resistance, T cell exhaustion and immune escape. There is no immune-related-adverse-event node and nothing about loss of peripheral tolerance in a non-tumour tissue. Anchoring the myocardial node to any of them would say something false about the heart.

So no `conforms_to` is declared there, and instead the entry carries a `CURATION_TODO` discussion (`ici_myocarditis_module_gap`) proposing that the module gain a shared irAE arm — which would then serve checkpoint-inhibitor colitis, pneumonitis and hepatitis as well, not just this entry. Happy to be overruled if a reviewer sees an anchor I have missed.

The `cardiomyopathy_maladaptive_remodeling` conformance in the issue does hold, and three nodes declare it: `#Primary Cardiomyocyte Insult`, `#Ventricular Remodeling`, and `#Structural Cardiac Impairment and Heart Failure`.

## Two competing hypotheses, kept competing

Why some patients recover and others end up with dilated cardiomyopathy is the open question of the field, and the entry does not pretend otherwise. Both leading explanations are curated as `mechanistic_hypotheses` with causal edges opting in:

- `postinfectious_autoimmunity` (CANONICAL) — the virus is cleared but tolerance to cardiac myosin is broken; anti-myosin antibodies transferred to uninfected mice reproduce the lesion.
- `viral_persistence` (ALTERNATIVE) — residual myocardial viral genome drives the ongoing injury.

They are not mutually exclusive and probably apply to different patients, which is exactly why biopsy with viral PCR governs whether immunosuppression is offered. There is also a `KNOWLEDGE_GAP` discussion on the progression question and a `HUMAN_MODEL_MISMATCH` on the two rodent models that carry most of the mechanistic content here — CVB3 models a virus that contemporary human biopsy recovers far less often than parvovirus B19 or HHV-6, and experimental autoimmune myocarditis starts from an adjuvanted immunization that has no human counterpart.

## Treatment evidence, including the negative trial

Immunosuppression in myocarditis is the case where the same intervention has a positive and a negative randomized trial, and the difference is patient selection. Both are curated on the same treatment:

- TIMIC (PMID:19556262) is the `SUPPORT` item — prednisone plus azathioprine improved ejection fraction in **virus-negative** biopsy-proven inflammatory cardiomyopathy.
- The Myocarditis Treatment Trial (PMID:7596370) is carried as `REFUTE` on the same treatment, because in **unselected** myocarditis with no virus exclusion it did not work. It bounds the indication rather than contradicting it.

Anakinra is curated explicitly as investigational, with a note saying the cited reference is the ARAMIS rationale-and-design paper and should not be read as a result.

## Research provenance

Deep research used **falcon** (`research/Myocarditis-deep-research-falcon.md`): 8/8 references verified, `confabulation_rate` 0.0, no `needs_review` flag. The report was used for framing and scope only — it cited mostly DOI keys rather than PMIDs, so **every** evidence snippet in the entry was sourced independently from PubMed and cached with `just fetch-reference`.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed, **90/90** snippets verified against cached references |
| `just validate-disorders` | passed (batched gate) |
| `just check-snippet-length` | passed |
| `just check-title-snippets` | passed |
| `just check-snippet-grading` | passed |
| `just check-environmental-evidence` | passed |

No stub existed for myocarditis, so none was deleted. A history record is included at `history/disorders/Myocarditis/`.

https://claude.ai/code/session_01SJcc7RnTXdSXCWjE9GdhV5


sierra-moxon:
Three additions since the PR body was written, all from a self-review pass against `just compliance`.

**Genetic susceptibility.** Added a `Genetically Susceptible Myocardium` pathophysiology node (`role: modifier`) and a matching `genetic:` entry, from the JACC Heart Failure meta-analysis (PMID:38573261). Pathogenic variants in cardiomyopathy genes are found in ~4% of uncomplicated acute myocarditis but ~22% of adults and ~45% of children whose course is complicated — desmosomal genes predominating in the mild group, sarcomeric in the severe one. No `gene_term` is bound, deliberately: the cited evidence is at the level of gene classes, and naming an HGNC gene would assert more than the source supports. That reasoning is in the entry's `notes`.

**Datasets.** Four GEO accessions added, all verified with `just verify-datasets` (4/4 OK) and relevance-triaged by hand rather than taken from the discovery ranking:

| Accession | Why it is here |
|---|---|
| `geo:GSE316643` | Human spatial transcriptomics of mRNA vaccine-associated myocarditis — matches the vaccination exposure curated in `environmental:` |
| `geo:GSE329991` | Human single-cell profiling of glucocorticoid-resistant ICI myocarditis |
| `geo:GSE302512` | Murine ICI myocarditis, CD8/TNFR2 — supports the checkpoint node |
| `geo:GSE297745` | Murine IL-17A neutralization in ICI myocarditis |

`just discover-datasets` also surfaced GSE280586 (murine cardiac arrest), GSE305470 (HFpEF), GSE292393 and GSE334160 (generic cardiac fibrosis and hypothalamic inflammation), and GSE278119 (neurological irAE, a GENE_ONLY match). All were dropped as false positives — they match on inflammation or myocardium vocabulary, not on myocarditis. Per the dataset SOP these records carry `publication:` and provenance `notes` but no `evidence:` block.

**Fixes.** Corrected the ARAMIS trial phase — I had `NOT_APPLICABLE`; ClinicalTrials.gov registers NCT03018834 as phase 2/3, so it is now `PHASE_II` with a note explaining that the schema slot is single-valued and the earlier phase was chosen rather than claiming phase III. Also added model-level evidence to all three animal models (previously the evidence sat only on the `modeled_mechanisms` links), evidence to the cardiac sarcoidosis differential, and the missing `reference_title` on the trial-registry evidence item.

**Re-validated after all of the above:** `just validate` 99/99 snippets verified, `just validate-terms` passed, `just validate-disorders` passed, `check-entity-refs` / `check-duplicate-keys` / `check-folded-hyphens` / `check-snippet-grading` all passed, `just verify-datasets` 4/4 OK.

**One thing I did not do, and want a second opinion on.** `Anti-Cardiac Myosin Autoimmunity` currently has no incoming edge, so it renders as an entry point. Mechanistically it is downstream — antigen released by dying myocytes is what breaks tolerance — but adding `Cardiomyocyte Necrosis and Contractile Failure → Anti-Cardiac Myosin Autoimmunity` would close a cycle through the adaptive-injury node. That amplification loop is real biology, but I did not want to introduce a cycle into the pathograph without knowing how the layout and KGX exporters handle one. Happy to add it if that is fine here.

sierra-moxon:
Both blocking items fixed, plus all four suggestions and the CI failure. Thanks for the review — the ARAMIS point in particular was a real gap, not a formatting nit.

## Blocking items

**1. Prevalence bands — fixed, and you were right on both.** 13.73/100,000 is 1.4 per 10,000 → `BAND_1_5_PER_10000`; 270/100,000 is 2.7 per 1,000 → `ABOVE_1_IN_1000`. I have also put the arithmetic into each record's `notes` so the next reader can check the band against the rate without doing the conversion themselves.

**2. ARAMIS result — added.** You are right that a `COMPLETED` trial citing only its design paper reads as pending. The result is now carried three times over: `supports: REFUTE` on the Interleukin-1 Blockade treatment, `supports: REFUTE` on the NCT03018834 record itself, and a `supports: PARTIAL` item bounding what is left. Source is PMID:42397625 (*Cardioimmunology of Myocarditis: Targeting the IL-1 Pathway*, 2026):

> "Clinically, the randomized trial of anakinra in acute myocarditis (ARAMIS) did not improve outcomes in a largely low-risk cohort, but accumulating case reports and small series suggest potential benefit in fulminant/hyperinflammatory myocarditis and chronic active refractory myocarditis."

The treatment `notes` now open with the negative result rather than with an "investigational" hedge. The design paper is retained but its `explanation` says explicitly that it reports no result.

## Suggestions — all four taken

- **anakinra modality.** Changed `PROTEIN_REPLACEMENT` → `OTHER`. You were right that it was wrong; anakinra is a receptor antagonist, not replacement of a deficient protein. `OTHER` also matches the one sibling anakinra entry in the KB, which I checked.
- **Second-line steroid-refractory ICI agents.** Added as its own treatment with `therapeutic_agent` bound for abatacept, ruxolitinib, mycophenolate mofetil and tocilizumab, from PMID:39263485. Curated deliberately as *proposed*: the source is a narrative review of 45 case reports and its own conclusion ("A large number of studies need to be conducted to validate or update our proposed treatment approach") is carried as a `PARTIAL` evidence item, so nothing here reads as established.
- **Corticosteroids in the giant-cell regimen.** Added prednisone to `therapeutic_agent` — the prose named it and the structured field did not.
- **`prevalence[0].population` hygiene.** Rewritten to name only the cohort ("Male BNT162b2 mRNA COVID-19 vaccine recipients aged 16-19, Israel"); the "excess over first dose" qualifier moved to `notes`.

## One thing I added while I was in there

The entry had no general-population rate at all — only two drug/vaccine-exposed denominators, which is a misleading picture of a disease. Added a Swedish national registry `ANNUAL_INCIDENCE` record (6.3-8.6 per 100,000, `BAND_1_9_PER_100000`) from the same PMID:42397625, with a note that the true incidence is almost certainly higher because mild cases go undiagnosed.

## CI failure — fixed

`Check for new mismatched reference titles` caught 12 findings across three PMIDs where I had written a plausible-looking title instead of copying the cached one. All three replaced verbatim from the cache frontmatter:

| PMID | Was (invented) | Now (cached) |
|---|---|---|
| 35533750 | Coronary spasm in patients with biopsy-proven viral myocarditis: prevalence and association with virus type. | Epicardial and microvascular coronary artery spasm in biopsy-proven viral myocarditis. |
| 11334481 | Innate immunity and the role of the complement system in the pathogenesis of viral myocarditis. | From infection to autoimmunity. |
| 10988241 | Enteroviral protease 2A cleaves dystrophin: evidence of cytoskeletal disruption in an acquired cardiomyopathy. | Dissociation of sarcoglycans and the dystrophin carboxyl terminus from the sarcolemma in enteroviral cardiomyopathy. |

Worth noting for anyone reading later: the snippets attached to those citations were all verified against the correct cached abstracts the whole time. The check's own wording — "a verified snippet does not vouch for the citation written beside it" — is exactly the failure mode, and it caught it.

I also normalized two snippets that I had transliterated as `interleukin-1beta` to the source's `interleukin-1β`. They passed validation either way, but matching the source exactly is the right form.

## Re-validation after all of the above

`just validate` **106/106** snippets verified · `just validate-terms` passed · `just validate-disorders` passed · `check_reference_titles.py --against-ref origin/main` now **OK, no new findings** · `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length` all passed.

Still open from my earlier comment, and genuinely your call: `Anti-Cardiac Myosin Autoimmunity` renders as an entry point although antigen release from dying myocytes is what drives it. Adding that edge closes a cycle. I checked a 120-file sample of `kb/disorders/` and found exactly one existing entry with a cyclic pathograph (`Alexander_Disease`), so precedent exists but is thin enough that I did not want to add one unasked.

sierra-moxon:
Added the feedback edge, since you checked the exporters and found it safe. Thank you for doing that legwork — the `nx.condensation` call in the cx2 layered layout and the `_longest_simple_path_length` fallback in `browser_export` were exactly what I could not confirm on my own, and 40 files with cycles is a much clearer precedent than the single one my 120-file sample happened to turn up.

`Cardiomyocyte Necrosis and Contractile Failure → Anti-Cardiac Myosin Autoimmunity`, `INDIRECT_KNOWN_INTERMEDIATES`, opted into the `postinfectious_autoimmunity` hypothesis group, evidenced by PMID:11334481 ("The chronic phase of myocarditis is associated with mononuclear infiltration of the myocardium and the production of autoantibodies to cardiac myosin, although infectious virus cannot be detected past day 14 of infection.") — a quote that places the autoantibody response after the acute injury phase rather than at the start of it, which is precisely what the edge claims.

The edge `description` says in as many words that the resulting loop is intentional and load-bearing, so a future reader does not try to break it as a curation error. Substantively it fixes the thing you named: the graph no longer implies that tolerance breaks spontaneously.

Re-validated: `just validate` **107/107** snippets verified, `just validate-disorders` passed, `check-entity-refs`, `check-duplicate-keys` and `check_reference_titles.py --against-ref origin/main` all clean.

**Not changed here, deliberately:** `Pericarditis.yaml:906` anakinra `PROTEIN_REPLACEMENT`. I agree with your read that this entry has it right and the sibling is what should move, but changing a file this PR otherwise does not touch would put an unrelated edit in a curation PR. Better as its own change.

This push will have dismissed the approval under `dismiss_stale_reviews`. Sorry for the extra round — the edge seemed worth one, given you had already established it was safe.

sierra-moxon:
Heads-up for anyone following this thread: the antigen-release feedback edge you cleared me to add did not make it into this PR.

I pushed it at 23:36, but GitHub did not sync this PR's head to that commit — the branch ref moved, the PR object kept reporting the previous SHA, and no CI run was queued for it. That went on for about 25 minutes, and in the meantime auto-merge merged the PR at the older head. So the edge is validated and endorsed but is not on `main`.

Carried over unchanged as #9971, cherry-picked onto current main with an `EDIT` history record. `just validate` passes there with 107/107 snippets verified.

Nothing else from your review is affected — everything in the approved commit landed correctly. Also noted and not touched: the `Pericarditis.yaml:906` anakinra `PROTEIN_REPLACEMENT` inconsistency you spotted is still open and still wants its own PR.

PR #9955, opened 2026-08-28
curate: Aspergillosis (MONDO:0005657) as a root entry with six subtypes

sierra-moxon:
Adds a knowledge base entry for aspergillosis (MONDO:0005657), curated as one root disease with six subtypes rather than as six separate entries.

The reason is that the clinical form of aspergillosis is set almost entirely by the host, not by the organism. Everyone inhales *Aspergillus* conidia every day. What decides whether that produces nothing, an allergy, a slowly expanding lung cavity, or a fatal angioinvasive infection is whether the patient's phagocytes can kill the fungus — and if not, why not. So the entry curates one shared mechanism with a single branch point, and hangs the clinical forms off it as `has_subtypes`.

```
inhaled conidia
      |
      v
phagocyte recognition + killing  ---- intact ----> cleared (no disease)
      |
   impaired  (neutropenia / corticosteroids / CGD / severe influenza or COVID-19)
      |
      v
germination and hyphal tissue invasion
      |-------------------------> angioinvasion -> thrombotic infarction -> dissemination
      \-------------------------> chronic cavitary destruction -> fibrosis

  (separate branch, immunity intact but Th2-skewed)
inhaled conidia -> Th2/IgE sensitisation -> eosinophilic inflammation
                                              and mucus plugging -> bronchiectasis
```

Closes #9041

## What is in it

**Subtypes** — Invasive, Chronic Pulmonary, Aspergilloma, ABPA, Aspergillus Bronchitis, Disseminated. Three of them are bound to MONDO terms; the other three explain in their own `description` why they are not. MONDO has no term for chronic cavitary or chronic fibrosing pulmonary aspergillosis, none for Aspergillus bronchitis, and no general disseminated-aspergillosis term (it codes individual organ forms such as neuroaspergillosis, which would misdescribe the non-CNS cases).

**Pathophysiology** — 17 nodes. The one worth looking at is `Failure of Phagocyte Clearance of Aspergillus`, which is the explicit branch point of the whole entry: everything upstream is shared by all six forms, and everything downstream depends on which host defect caused the failure. Two neutrophil killing mechanisms are modelled separately (an oxidase-independent route against conidia, an NADPH-oxidase-dependent route against hyphae) because that is what explains why chronic granulomatous disease and chemotherapy-induced neutropenia both cause the disease but behave differently.

[...]

PR #9972, opened 2026-08-29
curate: Add Kidney Sarcoma (MONDO:0002930) as a root with histological subtypes

### LINKED ISSUE #9604: Curate kidney sarcoma (MONDO:0002930)
Curate a dismech entry for **kidney sarcoma** ([MONDO:0002930](https://monarchinitiative.org/MONDO:0002930)) — a rare group of malignant mesenchymal renal tumours arising from renal stroma, capsule or vessels rather than from tubular epithelium, accounting for roughly 1% of adult renal malignancy and carrying a markedly worse prognosis than renal cell carcinoma.

This is a `CURATE_ROOT_WITH_SUBTYPES` target: model the histological subtypes with `has_subtypes` (leiomyosarcoma — the commonest, liposarcoma, angiosarcoma, synovial sarcoma with its `SS18` fusion, rhabdomyosarcoma). Distinguish it explicitly from the epithelial renal carcinomas and from Wilms tumour, and consider parallel `conforms_to` declarations against the hallmark modules (`invasion_and_metastasis`, `genome_instability_mutation`).

Tracker: part of #1079.

sierra-moxon:
Adds `kb/disorders/Kidney_Sarcoma.yaml` for primary renal sarcoma (MONDO:0002930) — the malignant mesenchymal tumours that arise from the kidney's smooth muscle, capsule, sinus fat and vessels rather than from tubular epithelium. They are about 0.25% of renal malignancy, they behave like soft-tissue sarcomas rather than like kidney cancer, and they do far worse than renal cell carcinoma at the same stage.

Closes #9604

## The lump-versus-split call

The brief flagged this as a strong grouping candidate, so it is worth saying why it ended up as a Disease entry with `has_subtypes` instead.

Two things decided it.

The subtypes share a real chain of mechanism, not just an address. Mesenchymal origin means no epithelial differentiation, which means none of the VEGF/mTOR/checkpoint framework built for renal cell carcinoma has a target. Growing in the retroperitoneum means the tumour is large before it produces a symptom — nearly half are T3 or T4 at diagnosis and 29% are already metastatic. Arising in or next to the renal vasculature means intravasation is immediate rather than a late acquired step, so spread is haematogenous, sometimes as a caval thrombus reaching the right atrium. That chain is more specific than "these are all cancers" and it holds across the histologies.

The other reason is practical. A grouping's members have to resolve to entries that already exist, and the only candidates here — `Leiomyosarcoma`, `Liposarcoma`, `Angiosarcoma`, `Synovial_Sarcoma` — are pan-site histology entries. Listing them as members of "Kidney Sarcoma" would say those diseases are renal, which they are not. That is the difference from `Paranasal_Sinus_Carcinoma`, the compare case: it already had site-specific member entries (`Paranasal_Sinus_Squamous_Cell_Carcinoma`, `Ethmoid_Sinus_Adenocarcinoma`, `Sinonasal_Undifferentiated_Carcinoma`) to unite. The in-repo precedent for the shape used here is `Sarcoma_Of_Cervix_Uteri`, a site-specific sarcoma umbrella with equally divergent subtype drivers curated as one Disease.

The reasoning is written into the entry, under `curation_scope_rationale` in the notes on the first pathophysiology node, so it is auditable rather than implicit. If site-specific member entries are curated later, revisiting this as a grouping is a reasonable follow-up.

## What is in the entry

Six subtypes with NCIT terms — leiomyosarcoma (commonest), liposarcoma, angiosarcoma (most lethal, HR 12.65 for death in the Japanese registry), synovial sarcoma, rhabdomyosarcoma, and Ewing sarcoma. The registry literature counts 43 histiotypes in total, so this is the well-described core rather than the whole category.

The pathograph runs: mesenchymal transformation → subtype-specific driver divergence → silent retroperitoneal expansion → vascular invasion and haematogenous spread → absence of an epithelial therapeutic target. A sixth node models the differential against sarcomatoid renal cell carcinoma, which is a separate mechanistic claim rather than a clinical footnote: the two converge on the same spindle-cell morphology from opposite directions (a sarcoma that was always mesenchymal, versus a carcinoma that dedifferentiated), and only one of them keeps access to checkpoint therapy.

The drivers really are unrelated — complex karyotype in leiomyosarcoma and angiosarcoma, the MDM2/CDK4 amplicon in liposarcoma, a single `SS18::SSX` fusion in synovial sarcoma, `EWSR1::FLI1` in Ewing. That divergence lives in per-subtype `genetic:` blocks keyed by `subtype:`, not blended into shared nodes.

Two `conforms_to` declarations against the hallmark modules. Worth flagging one deliberate omission: the entry does **not** claim the `Epithelial-Mesenchymal Transition Activation` node of `invasion_and_metastasis`. These cells are mesenchymal from the start and never undergo a phenotype switch, so claiming EMT would be wrong even though the rest of the module fits.

## Four open questions recorded as knowledge gaps

- **Radiotherapy.** The literature says renal sarcoma is radioresistant. The Japanese national registry found radiation independently associated with better survival (HR 0.39, p = 0.010). Both are in the entry, with the registry's inability to adjust for patient selection stated. Not resolved.
- **Does the kidney change the biology?** One partial answer exists and it is interesting: renal Ewing sarcoma is metastatic at diagnosis in 53% of pooled cases against 20–25% for the Ewing family overall, and has nodal disease in 24% against 3.2% for bone primaries — same fusion, different behaviour. Nobody has run the equivalent comparison for the commoner subtypes.
- **Guidelines mismatch.** The registry authors found these tumours are managed under renal cancer guidelines while behaving as soft-tissue sarcomas. Nobody has measured what that costs.
- **Dataset landscape.** See below.

## Datasets, and a Named Entity Confusion note

One dataset, `geo:GSE188555` (BCOR-CCNB3 undifferentiated sarcoma of the kidney), verified with `just verify-datasets`.

What got rejected matters more. Every higher-scoring candidate in the GEO sweep profiles clear cell sarcoma of the kidney or rhabdoid tumour of the kidney — both paediatric entities this entry explicitly excludes, both of which resolve perfectly and are about the wrong disease. There is no molecular dataset at all for renal leiomyosarcoma or renal angiosarcoma. That absence is recorded as a knowledge gap attached to `datasets#` rather than left as a blank section.

The same caution applied to references throughout: one candidate (PMID:38642139) was dropped in screening after it turned out to be about metastasising uterine leiomyoma. The entry's description states the exclusions — sarcomatoid RCC, Wilms tumour, clear cell sarcoma of the kidney, rhabdoid tumour of the kidney — so the boundary is explicit for the next curator.

## Deep research

falcon (Edison), `research/Kidney_Sarcoma-deep-research-falcon.md`. Reference validation 5/5 verified, confabulation rate 0.0. It surfaced the two strongest epidemiology sources in the entry (PMID:38693188, the SEER/NCDB series; PMID:32282650, the renal Ewing pooled analysis).

It also suggested `HP:0031605` for flank pain, which is wrong — the correct term is `HP:0030157`, confirmed with OAK. Every term binding in the entry was verified independently rather than copied from the report.

## Validation

| Check | Result |
|---|---|
| snippets verified against cache | 42/42 |
| `scripts/check_reference_titles.py --against-ref origin/main` | passed |
| `just check-folded-hyphens` / `check-snippet-length` / `check-title-snippets` | passed |
| `just verify-datasets` | 1/1 resolved |
| `just compliance` | 91.5% weighted |

All `reference_title` values were copied verbatim from `references_cache` frontmatter.

History record: `history/disorders/Kidney_Sarcoma/2026-08-29T001400Z-claude-code-e4a561.yaml`. No stub existed for this disease, so there is none to delete.

### DIFF
--- a/history/disorders/Kidney_Sarcoma/2026-08-29T055555Z-claude-code-53933f.yaml
+++ b/history/disorders/Kidney_Sarcoma/2026-08-29T055555Z-claude-code-53933f.yaml
@@ -1,0 +1,57 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Kidney_Sarcoma
+  path: kb/disorders/Kidney_Sarcoma.yaml
+session:
+  id: 2026-08-29T055555Z-claude-code-53933f
+  timestamp: '2026-08-29T05:55:55Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9604
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9972
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - diagnosis
+  - histopathology
+  - pathophysiology
+  - phenotypes
+  - genetic
+  - treatments
+  - prevalence
+  summary: 'Review response: Kidney Sarcoma PR #9972'
+  details: |
+    Response to the automated review on PR #9972 (CHANGES_REQUESTED). All four blocking findings addressed, plus five of the six optional suggestions.
+
+    Blocking finding 1 - wrong HPO term. The Tumor Hemorrhage phenotype bound HP:0001892 "Abnormal bleeding", whose HPO definition is a bleeding diathesis under Abnormality of blood and blood-forming tissues. A patient bleeding from an angiosarcoma has a normal haemostatic system, so this was a false match. Rebound to HP:0011029 "Internal hemorrhage", a direct child of the term previously used.
+
+    Blocking finding 2 - no diagnosis section. Added a five-entry diagnosis: block covering image-guided core needle biopsy (NCIT:C15189), cytokeratin/EMA immunohistochemical exclusion of sarcomatoid renal cell carcinoma (NCIT:C23020), MDM2 FISH with CDK4 immunohistochemistry for renal liposarcoma (NCIT:C17563), SS18-SSX immunohistochemistry as a FISH surrogate for renal synovial sarcoma, and EWSR1 rearrangement confirmation for renal Ewing sarcoma. Evidence is quoted from references already cached in this PR, chiefly the full text of PMID:38693188 (the Figure 3 legend naming the cdk-4/mdm-2 panel, and the pre-treatment biopsy discussion). The Diagnosis class has no subtype slot, so the subtype is carried in the entry name instead of a foreign key.
+
+    Blocking finding 3 - genetic entry named for three genes but binding one. Renamed "TP53, RB1 and ATRX (complex-karyotype subtypes)" to "TP53"; RB1 and ATRX already had their own entries. The complex-karyotype framing remains in notes.
+
+    Blocking finding 4 - orphaned pathophysiology node. "Diagnostic Confusion with Sarcomatoid Renal Cell Carcinoma" had no incoming or outgoing edges. Added an incoming DIRECT edge from "Malignant Transformation of Renal Mesenchyme" (the absent epithelial differentiation programme is what a dedifferentiated carcinoma mimics, so mesenchymal origin has to be demonstrated rather than assumed) and an outgoing DIRECT edge to "Absence of Epithelial Therapeutic Target" (resolving the differential in favour of a sarcoma is what forecloses checkpoint therapy). The pathograph is now a single connected component: every node has an incoming edge except the root, and an outgoing edge except the terminal node.
+
+    Optional suggestion 5 - added a four-entry histopathology: block (tumour necrosis, dedifferentiated liposarcoma pleomorphic spindle morphology, renal Ewing small round blue cell morphology, renal angiosarcoma atypical endothelial-lined vascular spaces), all quoted from the PMID:38693188 full text.
+
+    Optional suggestion 6 - added therapeutic_agent bindings (doxorubicin CHEBI:28748, ifosfamide CHEBI:5864) to Adjuvant Chemotherapy, with the description expanded to name the anthracycline backbone for the complex-karyotype subtypes and VDC/IE for renal Ewing, and a new evidence item from PMID:32282650 naming the VDC/IE regimen.
+
+    Optional suggestion 7 - added two surgical-margin evidence items to the nephrectomy treatment: R+ vs R0 HR 1.36 (95% CI 1.1-1.68, p = 0.004), and R+ in 21.3% of all resections rising to 40.6% in T4 disease.
+
+    Optional suggestion 8 - the prevalence notes now state that 0.53 per million is the incidence of the registry's whole 43-histiotype category, not of the six subtypes modeled here, since malignant rhabdoid tumour of the kidney is one of the two commonest histiotypes in that count and is explicitly excluded from this entry. The same caveat is extended to the clear cell sarcoma survival figure quoted on the driver-divergence node.
+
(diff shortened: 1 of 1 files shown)


sierra-moxon:
All four blocking findings are fixed, plus five of the six optional suggestions. The branch is also rebased on `origin/main` now that #9954 and #9955 have merged.

## Blocking

**1. `HP:0001892` was the wrong term.** You are right and it is a clear false match — a patient bleeding from an angiosarcoma has a normal haemostatic system, so a bleeding-diathesis term does not describe it. Swapped to `HP:0011029` **Internal hemorrhage**.

**2. Added a `diagnosis:` section.** Five entries: image-guided core needle biopsy, cytokeratin/EMA exclusion of sarcomatoid RCC, MDM2 FISH with CDK4 IHC for liposarcoma, SS18-SSX IHC as a FISH surrogate, and EWSR1 rearrangement confirmation for renal Ewing. Every snippet comes from a reference already cached in this PR, including the two passages you pointed at in `PMID_38693188` full text.

One schema note: `Diagnosis` has no `subtype` slot, so where an entry is subtype-specific I put the subtype in the entry name (`MDM2 FISH and CDK4 Immunohistochemistry (renal liposarcoma)`) rather than adding an unsupported key. Not a foreign key, so it is greppable but not enforced.

**3. Renamed the genetic entry to `TP53`.** The three-gene name was a leftover from an earlier draft where the three were one record.

**4. Connected the orphan node.** Two edges, both mechanistically load-bearing rather than decorative. Incoming from `Malignant Transformation of Renal Mesenchyme`, because the absent epithelial differentiation programme is exactly what a dedifferentiated carcinoma mimics, so mesenchymal origin has to be demonstrated at diagnosis rather than assumed. Outgoing to `Absence of Epithelial Therapeutic Target`, because resolving the differential toward sarcoma is what forecloses checkpoint therapy.

The pathograph is now one connected component:

```
Malignant Transformation of Renal Mesenchyme
├─> Subtype-Specific Driver Lesion Acquisition
│     └─> Silent Retroperitoneal Expansion
│           └─> Vascular Invasion and Haematogenous Dissemination
│                 └─> Absence of Epithelial Therapeutic Target
└─> Diagnostic Confusion with Sarcomatoid RCC ──> Absence of Epithelial Therapeutic Target
```

## Suggestions taken

- **Item 5, histopathology.** Four entries — tumour necrosis, dedifferentiated liposarcoma pleomorphic spindle morphology, renal Ewing small round blue cell morphology, renal angiosarcoma atypical endothelial-lined vascular spaces. All quoted from the `PMID_38693188` full text, which turned out to carry the Figure 2/3 legends.
- **Item 6, `therapeutic_agent`.** Added doxorubicin (`CHEBI:28748`) and ifosfamide (`CHEBI:5864`), plus a new evidence item from `PMID:32282650` naming VDC/IE for renal Ewing. The description now says explicitly that these name the anthracycline and alkylator common to the regimens rather than one regimen, because there is no single regimen across this entry's scope.
- **Item 7, surgical margins.** Two evidence items: R+ vs R0 HR 1.36 (95% CI 1.1–1.68, p = 0.004), and R+ in 21.3% of resections rising to 40.6% in T4 disease. The second is the more useful one — it says how often the only effective treatment fails at its own goal.
- **Item 8, registry caveat.** The prevalence `notes` now state that 0.53/million is the incidence of the registry's whole 43-histiotype category, not of the six modelled subtypes, since malignant rhabdoid tumour is one of the two commonest histiotypes in that count and is excluded here. Same caveat extended to the clear cell sarcoma survival figure.
- **Item 9, `evidence_source`.** Regraded five items on PMID:29660202 and PMID:30852045 from `OTHER` to `HUMAN_CLINICAL`. You are right that the convention grades the evidence reported, not the publication format, and what those two report is TCGA and whole-exome sequencing of human tumours. The explanations were rewritten to say so. `PARTIAL` and the non-renal-cohort caveats are unchanged — those are about cohort site, not evidence type.
- **Item 10, prevalence population.** Changed to `Adults, United States and international series`.

## Not done

The remaining compliance gaps are evidence on individual causal edges and on the four `discussions`. Edge-level evidence would restate the node-level evidence on either side of each edge, and the discussions are knowledge gaps whose whole content is that evidence is absent. Left at 91.8% weighted.

## Validation

| Check | Result |
|---|---|
| snippets verified | 59/59 |
| `scripts/check_reference_titles.py --against-ref origin/main` | passed |
| `just compliance` | 91.8% |

A second history record covers this review round.

PR #9973, opened 2026-08-29
Curate Peroxisome Biogenesis Disorder 11A (PEX13), MONDO:0013949

### LINKED ISSUE #9969: Curate peroxisome biogenesis disorder 11A (MONDO:0013949)
Curate a dismech entry for **peroxisome biogenesis disorder 11A (Zellweger)** ([MONDO:0013949](https://monarchinitiative.org/MONDO:0013949)) — the severe, Zellweger-end presentation of PEX13 deficiency, in which a defective docking complex at the peroxisomal membrane blocks PEX5-mediated matrix protein import and leaves behind import-incompetent "peroxisomal ghosts".

The KB already curates the per-PEX-gene entries `Peroxisome_Biogenesis_Disorder_1B` (PEX1), `Peroxisome_Biogenesis_Disorder_3A_Zellweger` (PEX12) and `Peroxisome_Biogenesis_Disorder_4B` (PEX6) as separate diseases under the `Peroxisome_Biogenesis_Disorder` / `Zellweger_Spectrum_Disorders` umbrella, so a PEX13 entry looks like the next sibling in that established pattern rather than a grouping. PEX13 is mechanistically distinct from the PEX2/PEX10/PEX12 RING-finger retrotranslocation group — it is an SH3-domain docking peroxin that binds PEX5 and PEX14 — so the pathograph should not simply restate PBD3A. PEX13 has no dedicated entry today (only a mention in `Peroxisome_Biogenesis_Disorder.yaml`).

Related MONDO sibling `peroxisome biogenesis disorder 11B` (the milder, NALD/IRD-end PEX13 phenotype) is worth checking during curation — it may belong as a `has_subtypes` entry here or alongside, mirroring how the A/B severity split is handled for the other PEX loci.

Stub: `stubs/Peroxisome_Biogenesis_Disorder_11A_Zellweger.yaml`

gingin77:
This adds a knowledge base entry for peroxisome biogenesis disorder 11A, the severe form of PEX13 deficiency. It is the fourth PEX gene we have curated individually, after PEX1, PEX12 and PEX6, and it is the first one that breaks a different part of the same machine.

Peroxisomes make none of their own enzymes. Every enzyme inside one is built in the cytosol and imported, in a cycle that runs in a fixed order:

1. The cargo enzyme is built in the cytosol carrying a PTS1 or PTS2 address tag.
2. A receptor binds it — PEX5 for PTS1 cargo, PEX7 for PTS2 cargo.
3. The loaded receptor arrives at the peroxisomal membrane and binds the PEX13/PEX14 complex.
4. The cargo is translocated through into the matrix.
5. PEX5 is ubiquitinated by the PEX2/PEX10/PEX12 complex.
6. PEX1 and PEX6 pull PEX5 back out to the cytosol, free to collect the next cargo.

PEX1, PEX6 and PEX12 — the three genes already in the KB — all break steps 5 and 6. Cargo arrives and unloads normally, but the receptor is never returned to the cytosol, so the cycle runs once and stalls. PEX13 is half of the complex at step 3, so losing it means the loaded receptor never binds the membrane and nothing is delivered at all.

Either way the organelle forms but stays empty, which is why a child with this disease is indistinguishable at the bedside from a child with the PEX12 form, and why the blood tests do not separate them. So the two entries share their lower half deliberately. They differ in their first three mechanism nodes, and a discussion in the entry says what does and does not follow from that.

Closes #9969

## What is in the entry

- **11 mechanism nodes** running from biallelic PEX13 loss, through failure of the receptor to bind at step 3, to the neurological, hepatic and adrenal endpoints — in one connected graph.
- **17 phenotypes**, including the brain imaging findings reported in PEX13 patients specifically (polymicrogyria-like cortical malformation, diffuse hypomyelination) rather than only the spectrum-wide picture.
- **Two mouse models** with explicit mechanism links. The whole-body knockout dies as a newborn; the brain-only conditional survives about five weeks, which is what makes the cerebellar and glial findings observable. The brain-only model is marked `PARTIALLY_RECAPITULATES` for plasmalogen deficiency because only half the biochemical signature is present in it — plasmalogens fall, but the fatty acids stay normal.
- **A refuted treatment.** DHA supplementation is an obvious idea here, since DHA is one of the lipids these cells cannot make. A randomised placebo-controlled trial found no benefit in biochemistry, vision or growth. It is curated with `supports: REFUTE` rather than left out, because the rationale is intuitive enough that it keeps being proposed. It carries no mechanism link on purpose: every available `treatment_effect` value would assert an effect the trial found absent.
- **A safety caveat on cholic acid**, the only approved drug for this disease group. In the four trial patients who already had advanced liver disease, it made liver chemistry worse.
- **Three discussions**: whether it matters that PEX13 fails at step 3 rather than at steps 5 and 6, whether the mitochondrial abnormality drives the neuronal loss or follows it, and how thin the evidence for the 11A/11B severity boundary actually is.

On the first of those: the clinical convergence is near total, so the honest headline is that the distinction buys nothing at the bedside. Two things do differ, and both are cell-biological. Breaking PEX13's ability to bind itself blocks PTS1 cargo specifically while leaving PTS2 routes and the PEX13/PEX14 interaction intact, so some PEX13 alleles cause a partial import defect that the step 5–6 mutants do not produce. And PEX5 ends up somewhere different: stuck at the membrane when recycling fails, never arriving when docking fails. Nobody has used that diagnostically. The entry records it as a mechanistic distinction and says so.

## The one thing worth arguing about

The A/B split at this locus rests on less than it does elsewhere. About twenty PEX13 variants are known worldwide. Assigning a case to 11A or 11B is done from the clinical description, not from a measured level of residual import, and variant class does not settle it either — a homozygous missense allele killed a child at 14 months, while the first patient ever assigned to this complementation group had the milder neonatal adrenoleukodystrophy phenotype. The entry says this in a knowledge-gap discussion rather than presenting the boundary as firmer than it is. If a reviewer thinks 11A and 11B should be one entry with subtypes instead, that is a reasonable position and I would rather hear it now.

## Deep research

Requested falcon; it is not configured in this environment, so the run used `--fallback` and landed on `claude_code`. The report records that (`fell_back: true`, `requested_provider: falcon`) rather than the substitution being made by hand. All 49 of its references resolved and none were off topic.

Its term validation is worth flagging. Nothing it suggested was fabricated, but two identifiers it offered would have been real errors if copied across: `HP:0034512`, offered for elevated very-long-chain fatty acids, is *Transitional-cell carcinoma of the ureter*, and `HP:0000934`, offered for chondrodysplasia punctata, is *Chondrocalcinosis*. Neither is bound in the entry.

The report also proposed NCT03856866, a hydroxychloroquine trial, as a candidate treatment. Reading the registration shows it only enrolls patients with PEX1, PEX6 or PEX26 disease, so it does not belong here and is not curated.

## Validation

`just validate-disorders` passes: schema, ontology terms, and 86 of 86 evidence snippets verified against cached references. Also run and clean: `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `validate-history`.

Compliance is 87.3%, against 77.8% for the PEX12 sibling entry.

GeneReviews (PMID:20301621) is tagged and quoted in all four of its sections — clinical characteristics, diagnosis, management, and genetic counseling.

Reference cache and term cache additions are limited to what this entry cites: 13 new PMIDs and 7 cache rows.

### DIFF
--- a/history/disorders/Peroxisome_Biogenesis_Disorder_11A_Zellweger/2026-08-29T012445Z-claude-code-8e4a3d.yaml
+++ b/history/disorders/Peroxisome_Biogenesis_Disorder_11A_Zellweger/2026-08-29T012445Z-claude-code-8e4a3d.yaml
@@ -1,0 +1,44 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Peroxisome_Biogenesis_Disorder_11A_Zellweger
+  path: kb/disorders/Peroxisome_Biogenesis_Disorder_11A_Zellweger.yaml
+session:
+  id: 2026-08-29T012445Z-claude-code-8e4a3d
+  timestamp: '2026-08-29T01:24:45Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9969
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/9973
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - phenotypes
+  - treatments
+  - animal_models
+  - discussions
+  - evidence
+  summary: 'Review round 3: narrow overstated gaps, fix zebrafish evidence grading'
+  details: |
+    Third round on PR #9973, addressing a CHANGES_REQUESTED review with two blocking findings and two suggestions.
+
+    Blocking finding 1: two curated sentences claimed an experiment had not been done when it partly had. PMID:27514574 studies the same brain-restricted PEX13 mouse already curated here and reports an expanded, morphologically modified mitochondrial population in the intact brain plus antioxidant rescue of mitochondrial superoxide and membrane depolarization in PEX13-deficient fibroblasts. The animal-model limitations said the measurement was "in cultured E19 neurons rather than in the intact brain", which was no longer true, and the mitochondrial-causal-direction discussion said a rescue experiment "has not been reported", which overstated the gap. Both narrowed to what is genuinely missing - the rescue in a neuron with a survival endpoint - and the paper is now cited on the node with two in-vivo readouts. The proposed experiment was sharpened accordingly, from "do a rescue" to "repeat the reported fibroblast rescue in neurons and make survival the primary endpoint".
+
+    Blocking finding 2: evidence_source IN_VITRO on a PMID:36541703 snippet whose text names a zebrafish model. Regrading that item to MODEL_ORGANISM exposed a second problem the checker caught: the discussion's quote contains the pathophysiology node's shorter quote as a substring, so the two gradings collided. Resolved by grading the shared headline claim MODEL_ORGANISM in both places (it rests on the zebrafish arm as well as the cell lines) and leaving the cell-line mechanism sentence as IN_VITRO - the per-sentence split a mixed-source paper calls for.
+
+    Suggestion 1: added the Klouwer 2015 (PMID:26627182) corticosteroid caveat to Adrenal Replacement Therapy - screen everyone, treat only demonstrated insufficiency on Synacthen testing, because cortisone causes growth suppression and osteoporosis. Symmetric to the cholic-acid hepatotoxicity caution already present.
+
+    Suggestion 2: the reviewer noted the exclusion rationale for elevated transaminases applied a stricter standard than the newly added Adrenal Insufficiency phenotype. Correct. PMID:26627182 carries a quotable spectrum-wide sentence, so Elevated Hepatic Transaminases (HP:0002910) is now curated at the same PARTIAL footing. Apnea and lissencephaly stay out - no cached reference contains a sentence stating them - and the notes now state the single threshold applied to all three rather than implying different ones.
+
+    Also repaired a self-inflicted editing error: a scripted patch intended for the top-level references list matched two evidence items as well and inserted reference-list entries inside them, breaking YAML parsing. Caught by just validate before commit and reverted.
+
+    Revalidated: just validate-disorders passes with 99/99 snippets verified, all ungated checks clean, compliance 88.9%.
--- a/kb/disorders/Peroxisome_Biogenesis_Disorder_11A_Zellweger.yaml
+++ b/kb/disorders/Peroxisome_Biogenesis_Disorder_11A_Zellweger.yaml
@@ -478,6 +478,17 @@ pathophysiology:
       the pathophysiology of ZSDs
     explanation: PARTIAL because the authors themselves frame the contribution as emerging
       rather than established.
+  - reference: PMID:27514574
+    reference_title: Mitochondrial changes and oxidative stress in a mouse model of Zellweger
+      syndrome neuropathogenesis.
+    supports: SUPPORT
+    evidence_source: IN_VITRO
+    snippet: Cultured fibroblasts from PEX13-deficient mouse embryo displayed similar changes,
+      as well as increased levels of mitochondrial superoxide and membrane depolarization; this
+      phenotype was rescued by antioxidant treatment.
+    explanation: The one perturbation reported at this node - antioxidant rescue of the
+      mitochondrial phenotype. IN_VITRO because it is in cultured fibroblasts; it shows the
+      phenotype is correctable without showing that correcting it changes cell fate.
 - name: Pexophagy of Import-Incompetent Peroxisomes
   biological_scale: CELLULAR
   mechanism_confidence: PROVISIONAL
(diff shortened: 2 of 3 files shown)


gingin77:
Pushed a second round. Two sources of feedback: the automated second opinion on the claim issue (#9969), and the suggestions in the approving review here.

## What changed

**The human docking module is not the yeast one, and the entry now says so.** In yeast, the Pex13 SH3 domain binds a polyproline PxxP motif in Pex14. In humans that surface does not recognise PEX14 PxxP motifs at all — PEX14 binding is mediated by an FxxxF motif sitting just past the SH3 domain, and that same motif folds back onto the SH3 domain and regulates how PEX5 binds there (PMID:38632234). The yeast model is the one in most reviews, so leaving it unstated invited a reader to carry it across. The docking node now describes the human arrangement, including the fact that it carries its own autoregulatory switch.

**Pexophagy is now curated, reversing an earlier call.** The first version left PEX13's autophagy role out on the grounds that nothing connected it to disease. That was wrong. PMID:36541703 is about *losing* PEX13 — which is the disease lesion, not an engineered gain of function — and shows that the loss lets ubiquitinated PEX5 accumulate and peroxisomal reactive oxygen species rise, with the two together sending otherwise healthy peroxisomes for degradation. That plausibly compounds the import block: the residual organelles are being cleared as well as left empty, which fits the reduced peroxisome count seen in patient fibroblasts.

It is a `PROVISIONAL` node, not an established one, with a `HUMAN_MODEL_MISMATCH` discussion attached. The work is in edited cell lines and zebrafish, and nothing has measured autophagic flux in patient material. The discussion spells out why that distinction has teeth: if excess pexophagy is clearing peroxisomes that retain partial function, inhibiting autophagy preserves them; if they are import-dead anyway, it achieves nothing. The trials built on that rationale exclude PEX13, so the question has not been asked here even indirectly.

## Reviewer suggestions

| Suggestion | Done |
|---|---|
| `Secondary Mitochondrial Dysfunction` description says "curated as EMERGING" but the value is `PROVISIONAL` | Fixed. A real error — I changed the slot value and left the prose behind. `EMERGING` is not a `MechanismConfidenceEnum` member. |
| `reference_title` missing on the two structured-source evidence items | Added to both the ClinGen assertion and the ClinicalTrials record. |
| Cerebellar atrophy, adrenal insufficiency uncurated | Added. `HP:0001272` from the PEX13 series, `HP:0000846` graded `PARTIAL` because the GeneReviews sentence supporting it describes the intermediate/milder end of the spectrum, not the severe one this entry curates. |

## What I did not add, and why

Apnea, elevated hepatic transaminases and lissencephaly. All three are in the HPO annotation set for OMIM:614883, but that annotation set is not something this KB can quote from, and no cached reference contains a sentence stating them for a PEX13 patient. Binding them would mean attaching a real phenotype to a snippet that does not say it. They are good candidates and the entry's `notes` records them as pending a citable source, so the next person does not have to rediscover them.

## Validation

`just validate-disorders` passes — 94 of 94 snippets verified. All ungated checks clean. Compliance 88.6%, up from 87.3%.

gingin77:
Both blocking findings fixed, and both suggestions taken.

## Finding 1 — the entry claimed an experiment had not been done

You are right, and this is the more useful of the two catches. `PMID:27514574` studies the same brain-restricted PEX13 mouse the entry already curates, and it does two things I had written as undone:

- the mitochondrial phenotype is shown **in the intact brain** — an expanded, morphologically modified population, with lipid and DNA oxidation products in neurons — not only in cultured E19 neurons;
- an **antioxidant rescue** of mitochondrial superoxide and membrane depolarization has been run, in PEX13-deficient fibroblasts.

So the `limitations` field was describing a limitation that had been lifted, and the discussion's "which has not been reported" was simply wrong. Both now say what is genuinely absent: that rescue in a *neuron*, read out on *survival* rather than on the mitochondrial measurements themselves. A correctable phenotype in a fibroblast says nothing about whether correcting it keeps a neuron alive, and that is the whole question.

The paper is now cited on the node with two in vivo readouts plus the rescue, and `pbd11a_mito_rescue` is rewritten around the narrower gap — from "do a rescue experiment" to "repeat the reported fibroblast rescue in neurons, with apoptosis and survival as the primary endpoint."

## Finding 2 — zebrafish graded IN_VITRO

Regraded to `MODEL_ORGANISM`. Doing that exposed a second problem, which `check-snippet-grading` caught before I pushed: the discussion's quote **contains** the pathophysiology node's shorter quote as a substring, so the two items could not carry different values.

Resolved by grading the shared headline claim `MODEL_ORGANISM` in both places — it rests on the zebrafish arm as much as the cell lines — and leaving the cell-line mechanism sentence (ubiquitinated PEX5, peroxisomal ROS) as `IN_VITRO`. That is the per-sentence split a mixed-source paper calls for, rather than one blanket grade for the paper.

## Both suggestions

**Klouwer corticosteroid caveat** — added. The screening and the treating are deliberately separate steps: screen everyone, treat only a demonstrated insufficiency on Synacthen testing, because cortisone itself causes growth suppression and osteoporosis. It reads as the direct counterpart to the cholic-acid caution, which is why it belonged.

**The inconsistent threshold** — a fair hit. I was excluding elevated transaminases for lack of a PEX13-specific quote while admitting Adrenal Insufficiency on a spectrum-wide GeneReviews sentence. `PMID:26627182` has a quotable spectrum-wide sentence for transaminases too, so `HP:0002910` is now curated at the same `PARTIAL` footing. Apnea and lissencephaly stay out — no cached reference contains a sentence stating them — and the `notes` now state the single threshold applied to all three instead of leaving the asymmetry implicit.

## Validation

`just validate-disorders` passes, 99 of 99 snippets verified. All ungated checks clean, including the grading check that caught the substring collision. Compliance 88.9%.

Ready for re-review.

Issue #9718, opened 2026-08-27
Snippet verification silently skips DOI, ORPHA and other non-PMID references


cmungall:
A wrong snippet on a `DOI:`-cited evidence item passes every check we have. `just count-verified-snippets` reports it as "skipped by prefix" and `just validate-disorders` — the authoritative gate — passes too. Nothing goes red.

## How it surfaced

While curating `Methylcobalamin_Deficiency_Type_cblE` (#9649) I quoted two sentences from a cached DOI reference. Both were wrong: the cache renders the hyphens in "long‐term" and "FAD‐binding" as U+2010 HYPHEN, and I had typed ASCII hyphens. Neither snippet was an exact substring of its source.

The counter reported:

```
Snippets checked: 98/100 verified against cached references (2 skipped by prefix)
```

and `validate-disorders` passed. I only caught it because I went looking by hand. A curator reading either output would reasonably conclude all snippets were verified — the phrasing "98/100 verified" reads as a pass, and the two skips are a parenthetical.

## Why it matters more than it looks

The whole evidence discipline in this repo rests on snippets being exact substrings. That's what stops fabricated quotes. For PMID references the check is real and it works. For every other prefix it silently does not run, and the failure mode is invisible — a snippet can drift from its cache, or never have matched, with a green build.

The U+2010 case is worth noting because it isn't carelessness that a careful curator would avoid. The character is visually identical to a hyphen. Copying a sentence out of a cache and retyping any part of it can introduce it, and no amount of proofreading catches it by eye.

Affected prefixes include `DOI:`, and by the same mechanism `ORPHA:`, `CGGV:`, `CGDS:`, `ICEES:`, `NCIT:` and `ICTRP:` — all of which have committed cache files that could be substring-matched exactly as PMID caches are.

## What would fix it

Two things, roughly in order of value:

1. **Verify any reference that has a cache file**, regardless of prefix. The cache filename is derivable from the CURIE (`DOI:10.1023/A:1025159103257` → `references_cache/DOI_10.1023_A_1025159103257.md`), and the structured sources already document their rows as stable quotable substrings. The check is the same substring test.
2. **Make the skip loud when it happens.** Where a reference genuinely can't be checked, report it as a distinct outcome rather than folding it into a passing count — and consider making an unverifiable snippet fail `validate-disorders` rather than pass it.

[...]

Issue #9697, opened 2026-08-27
Bare `downstream[].target` values are unchecked: a renamed node silently severs the pathograph


matentzn:
## The gap

`Pathophysiology.downstream[].target` accepts two forms:

```yaml
downstream:
- target: phenotypes#Recurrent Bacterial Infection    # prefixed -> checked
- target: Impaired Opsonophagocytosis                 # bare node name -> NOT checked
```

`check_entity_refs` / `test_entity_ref_foreign_keys` resolve `<kind>#<name>` references. A **bare** target carries no prefix, so it is skipped. Nothing else validates it either.

Result: rename a pathophysiology node, miss one upstream edge, and the causal graph is severed — while `just validate`, term validation, all six baseline gates, `check-entity-refs`, `check-duplicate-keys` and the full `tests/test_data.py` suite pass.

## It is not hypothetical, and not rare

I hit it **twice in one session**, in two different entries, both while *acting on review feedback that asked me to connect the graph*:

- [#9667](https://github.com/monarch-initiative/dismech/pull/9667) — renamed `Impaired Opsonisation of Encapsulated Bacteria` → `Impaired Opsonophagocytosis`, left the upstream edge on the old name. The infection branch became `C2 Gene Defect → Convertase Failure → (phantom)`, with the opsonophagocytosis node floating as a disconnected fragment.
- [#9666](https://github.com/monarch-initiative/dismech/pull/9666) — added a CNS node whose `downstream` pointed at `Cerebellar Ataxia` and `Putaminal and Periaqueductal Lesions` as bare names. Neither is a pathophysiology node (one is a phenotype, one an imaging finding), so both edges dangled.

**The second one had already passed review.** The reviewer explicitly recorded *"Entity refs resolve; the graph is now one connected component"* — because they checked the gate, and the gate passes. It was only found by a manual sweep after the first break was caught. That is the part that makes this worth fixing rather than just being more careful: the existing check reports success in a way that is actively misleading.

CLAUDE.md already warns *"Renaming or splitting a node is the common way to break them, so search the file for the old name before committing"* — but the tooling that is supposed to enforce it has this hole.

## Suggested fix

Extend `check_entity_refs` (and `test_entity_ref_foreign_keys`) to resolve bare `downstream[].target` values against the `pathophysiology[].name` set in the same file. Roughly:

```python
for node in disease.get("pathophysiology", []) or []:
    for edge in node.get("downstream", []) or []:
        tgt = edge["target"]
[...]

Issue #9584, opened 2026-08-27
Split Castleman disease into three entries plus a grouping, and fix seven MONDO defects upstream


dragon-ai-agent:
`kb/disorders/Castleman_Disease.yaml` is a single entry covering four diseases that have different causes, different first-line drugs, and a different confirmatory test each. Its causal graph currently says that a herpesvirus infection and an unknown idiopathic trigger both drive the same patient's disease — true of no patient. This proposes splitting it into three disease entries plus a grouping, and separately reports seven problems in MONDO's Castleman branch, one of which makes every idiopathic multicentric case invisible to a query over "multicentric Castleman disease".

Companion to #7082 (the disjunction test) and #306 (granularity guidelines). Castleman is a cleaner worked case than ependymoma because the split is along *cause*, not along molecular subgroup.

The entry was substantially expanded on branch `claude/castlema-entry-review-c2mue0` (636 → 3,544 lines, 16 pathophysiology nodes, 193 verified evidence snippets). That expansion is what made the strain measurable rather than theoretical.

---

## 1. The disjunction test fires

#7082's rule: if the claims that distinguish the subtypes can be scoped with the `subtype:` discriminator — differences in phenotype, gene, prevalence, biomarker, imaging or course — lump. If they are differences in **pathophysiology, treatment or diagnosis**, none of which have a `subtype:` slot, the schema cannot scope them. Split.

All three unscopeable classes differ by branch here:

| | Unicentric CD | HHV-8-associated MCD | Idiopathic MCD | POEMS-associated MCD |
|---|---|---|---|---|
| **Cause** | localized stromal lesion; somatic *PDGFRB* N666S in ~10% | KSHV lytic replication making viral IL-6 | unknown | clonal plasma cell disorder |
| **First-line treatment** | surgical resection, usually curative | rituximab (+ antiretrovirals) | siltuximab | treat the plasma cell clone |
| **Confirmatory test** | single nodal region on imaging | HHV-8 LANA-1 immunohistochemistry positive | LANA-1 negative plus exclusions | monoclonal protein |

Nothing in that table is expressible with `subtype:`.

In the current file, concretely:

- **14 of 16 pathophysiology nodes and 10 of 11 treatments are scoped to a branch in free-text prose only.** That is the same smoking gun #7082 found in ependymoma, at higher density.
[...]

Issue #7684, opened 2026-08-01
Two snippet checkers disagree in both directions: validate-references skips DOI, check_snippets_verbatim forgives ligatures


jmcmurry:
The repo now has **two** snippet checkers and they disagree in **both directions**. Neither is a superset of the other, so passing one says nothing definitive about the other. A curator can reasonably believe their evidence is verified when half of it has not been.

| | `just validate-references` | `scripts/check_snippets_verbatim.py` (#7592) |
|---|---|---|
| `DOI:`-prefixed snippets | **skipped entirely** (#7514) — 5,047 snippets | checked |
| `url:`, `GEO:` etc. with a cache body | skipped by prefix | checked |
| PDF ligatures (`signiﬁcant`, `ﬂ`) | **not normalised** — exact match | normalised via NFKC |
| Hyphenated line breaks (`hyper-\nreflexia`) | **not normalised** | de-hyphenated |
| Lancet middle-dot decimals (`61·4%`) | not normalised | normalised |
| `...` elision in a quote | not supported | supported, segments must match in order |

## How it surfaced

While curating #7668 (ABCC9), an agent reported:

> Twenty of my initial snippets from the `PMID:31575858` full-text PDF cache **passed the repo checker and failed the validator.**

I reproduced the mechanism directly:

```python
snippet "the significant finding"  vs  cache "the signiﬁcant finding"   -> my checker PASSES
snippet "hyperreflexia was noted"  vs  cache "hyper-\nreflexia was noted" -> my checker PASSES
```

`validate-references` fails both, because it does exact matching with no normalisation (there are no normalisation options in `conf/reference_validator_config.yaml`).

## Why I am filing this against my own tool

I introduced `check_snippets_verbatim.py` in #7592 and have been reporting "N/N snippets verbatim" on every curation PR in #1079 as an independent verification. That claim needs a caveat I did not give it: **my checker is stricter than `validate-references` on coverage (it checks DOI) but looser on text normalisation (it forgives ligatures and hyphenation).** "Passes my checker" therefore does not imply "passes CI", and I should have said so.

The PR description for #7592 framed the tool as catching what CI skips. True, but incomplete.

## The real cost, seen on #7668

The curating agent resolved each of the 20 conflicts honestly — truncating to a ligature-free span, substituting a clean sentence, or dropping the item — and explicitly did not weaken any claim to fit a quote. But content was lost to a *formatting* artifact:

[...]

PR #9974, opened 2026-08-29
Split Castleman disease into three etiologic entries, a grouping, and an IL-6 module

dragon-ai-agent:
Castleman disease was one entry covering four diseases that have different causes, different first-line drugs, and a different confirmatory test each. Its causal graph asserted that a herpesvirus infection and an unknown idiopathic trigger both drive the same patient's disease, which is true of no patient. This splits it into three disorder entries plus a grouping, extracts the shared IL-6 chain as a module, and moves the POEMS-associated form into the POEMS entry where its actual driver lives.

Rationale and the MONDO defects this surfaced are in #9584.

## Why split

Per the disjunction test in #7082: the claims that distinguish the Castleman forms live in `pathophysiology`, `treatments` and `diagnosis`, none of which carry a `subtype:` discriminator. In the retired umbrella entry, 14 of 16 pathophysiology nodes and 10 of 11 treatments were scoped to a branch in free-text prose only, and the graph had two mutually exclusive etiologic roots feeding one shared IL-6 node.

| | Cause | First-line treatment | Confirmatory test |
|---|---|---|---|
| Unicentric | localized stromal lesion; somatic *PDGFRB* N666S in ~10% | surgical resection, usually curative | single nodal region on imaging |
| HHV-8-associated | KSHV lytic replication making viral IL-6 | rituximab (+ antiretrovirals) | HHV-8 LANA-1 immunohistochemistry |
| Idiopathic | unknown | siltuximab | LANA-1 negative plus exclusions |
| POEMS-associated | clonal plasma cell neoplasm | treat the plasma cell clone | monoclonal protein |

## What changed

**Removed:** `kb/disorders/Castleman_Disease.yaml` and `kb/disorders/Multicentric_Castleman_Disease.yaml`. These overlapped each other with no record of the union.

**Added three disorder entries.**

| Entry | Pathophysiology nodes | Phenotypes | Evidence snippets |
|---|---:|---:|---:|
| `Unicentric_Castleman_Disease` | 5 | 8 | 49 |
| `HHV-8-Associated_Multicentric_Castleman_Disease` | 6 | 13 | 50 |
| `Idiopathic_Multicentric_Castleman_Disease` | 17 | 22 | 147 |

The idiopathic entry carries iMCD-TAFRO, iMCD-IPL and iMCD-NOS as `has_subtypes`.

**Added `kb/groupings/Castleman_Disease.yaml`** — a curated union with four members, `grouping_basis: [SHARED_PHENOTYPE, CLINICAL_CONVENTION]`, and `differentiating_mechanisms` per member. POEMS Syndrome is the fourth member, so the historical four-subtype framework is preserved even though its Castleman manifestation is curated elsewhere.

**Added `kb/modules/il6_hypercytokinemia.yaml`** — six nodes on the standard role chain, with gp130-mediated JAK-STAT3 activation as the key conformance target. It carries a three-rung drug-target ladder: anti-IL-6 antibody inhibits the trigger, anti-IL-6-receptor antibody the amplifier, JAK inhibition the central effector. Eight conformers wired, four in each of the two multicentric entries.

**Updated `POEMS_Syndrome.yaml`** to absorb POEMS-associated MCD: a histopathology entry for Castleman-type angiofollicular lymph node hyperplasia, a progression entry giving the outcome split by osteosclerotic lesions (90% vs 27% 5-year survival), and a differential diagnosis against idiopathic MCD. POEMS does **not** conform to the IL-6 module — it is VEGF-driven.

**History and stubs.** Seven history records moved into the successor slug directory with `target.superseded_by` blocks, per the append-only rule. `stubs/Castleman-Kojima_Disease.yaml` resolved as `entry_type: SUBTYPE` with the reasoning recorded.

## Judgment calls worth a reviewer's attention

**The HHV-8 entry carries no bound `disease_term`.** MONDO has no class for the HHV-8-associated form. The nearest candidate, `MONDO:0019754`, is the *parent* of both multicentric arms — binding it would assert this entry is multicentric Castleman disease in general, and since an entry's primary `disease_term` counts as curated for MONDO coverage, it would retire the parent concept from the queue on the strength of curating one child. A free-text `preferred_term` is kept instead, matching the pattern the design register states for treatments and the 36 existing entries with no bound disease term. It is anchored through `mappings.mondo_mappings` with `skos:broadMatch` — the direction that says the mapped term is broader — which gives the anchor without claiming the parent as curated. Rebind when MONDO mints the class (M1 in #9584).

This predicate was queried in review and resolved as `broadMatch`. The `narrowMatch` suggestion came from a line in CLAUDE.md's cancer-granularity section that reads backwards against both SKOS and the 61 existing `broadMatch` usages; that is now #10165.

**The IL-6 module is deliberately distinct from two neighbours**, with the discriminating rule for each written into its `notes`:

- vs `jak_stat_pathway_activation` — that module covers constitutive, ligand-*independent* activation from an activating receptor/JAK/STAT lesion, terminating in cytokine-independent proliferation. This one covers a structurally normal pathway driven by excess ligand, where removing the ligand switches it off. A disorder that remits on IL-6 blockade conforms here, not there.
- vs `cytokine_storm_hyperinflammation` — that module models the *generation* of an acute storm ending in shock and multiorgan failure. This one picks up at the cytokine and models how the IL-6 signal is *transduced* and which tissue programs it turns on. They overlap at the IL-6 node and diverge on either side of it. `conforms_to` is single-valued, so a disorder doing both — iMCD-TAFRO — conforms its capillary-leak node to the storm module and its cytokine chain here.

**Unicentric disease does not conform to the IL-6 module.** Serum IL-6 is normal in most unicentric cases, so wiring it would be conformance by family resemblance.

**iMCD-TAFRO stays a subtype rather than becoming its own entry.** It is the closest call here. Under the promotion rule in #7082 it arguably qualifies: it has its own MONDO term, its own Japanese diagnostic criteria, and two pathophysiology nodes that do not apply to its siblings. Against that, it shares the etiology, diagnostic criteria and first-line drug with the rest of iMCD, which is the axis the split was drawn on. Promote it if a subtype-specific first-line therapy is established, plausibly sirolimus via NCT03933904.

**Evidence was rescoped, not just moved.** Content migrated line-for-line where it applied unchanged, so snippet verification carries over. Where a block referenced another branch it was replaced rather than left to bleed — the HHV-8 entry inherited TAFRO-derived evidence for its plasma-cell node, which now cites KSHV-specific evidence instead.

**All 24 references from the retired entries are placed, none retired.** The eight that were initially unaccounted for turned out to be live sources: five HHV-8-specific, two iMCD-specific, and `PMID:39970477` ("Castleman's disease: one disease, multiple etiologies") on the grouping, since it is the split thesis rather than evidence for any one arm. `DOI:10.3390/ijms25073775`, the HHV8 monoclonal microlymphoma case, sits in the entry whose `notes` record the somatic-clonality gap it speaks to. All are provenance entries with no `findings`: none has a cached abstract from which an exact quote could be taken, and manufacturing one is the failure mode the evidence SOP exists to prevent.

**Migrated to the narrowed `supports` enum.** #10003 retired `PARTIAL` and `WRONG_STATEMENT` after this branch was cut. Applied via `scripts/migrate_supports_enum.py --apply`, the path its own docstring names for a branch that goes red on refresh. Five of the eleven items carry `directness: INDIRECT`, because their `explanation` prose already recorded that assessment in the curator's words; the other six became plain `SUPPORT` with `directness` absent, as the migration prescribes. No `REFUTE` was introduced — `supports` is scoped to the parent object's assertion, and splitting a hedged item would repeat the Marfan error #10003 itself had to revert.

## Validation

- `just validate` (schema + terms + references) passes on all four disorder files, the module and the grouping
- `just validate-grouping` and `just check-groupings` pass; the one `UNKNOWN` criterion is expected, as no HP or GO term exists for the angiofollicular hyperplasia pattern
- `just validate-history-all` clean over 7,752 records
- `check-entity-refs` OK across 2,777 KB files; duplicate-key, folded-hyphen, snippet-length, title-snippet, snippet-grading and environmental-evidence gates clean
- Graph build over all five changed files: 0 self-loops, 0 dangling targets
- `tests/test_data.py`: 46,864 passed

Two known limits on what that certifies, neither introduced here: `DOI:` is in `skip_prefixes`, so 121 of the 278 disorder snippets are not machine-verified by either gate (#7514, #7684, #9718); and bare `downstream[].target` values are exempt from `test_entity_ref_foreign_keys`, which is how both an earlier dangling edge and a self-loop in this PR reached review uncaught (#9697, where the self-loop variant and the 11 pre-existing cases on `main` are now recorded).

One advisory warning remains, on `iMCD-IPL` and `iMCD-NOS` missing `subtype_term`. MONDO has no term for either; that is M4 in #9584.

https://claude.ai/code/session_01DxoA2LrdVZHsuPQNyVSJ77


dragon-ai-agent:
All six items are addressed in 1ef3c3f13. Five as asked; one implemented differently, explained below.

**Critical — dangling edge.** Confirmed and fixed. This was a regression I introduced: I ran the foreign-key check before renaming the phenotype to `Localized Lymphadenopathy`, and since `downstream[].target` is exempt from `test_entity_ref_foreign_keys`, nothing caught it afterwards. Target corrected and the edge explanation rewritten to state the single-region claim rather than the UCD/MCD contrast.

Worth noting the exemption is a general hole, not a Castleman one — a renamed node or phenotype silently orphans every `downstream` edge pointing at it, across the whole KB. Happy to file that separately if it isn't already tracked.

**Item 2 — UCD VEGF on an MCD-only snippet.** Fixed with the drop-in you identified, plus the follicular-dendritic-cell/penetrating-vessel sentence from the same paper, since that node also claims the lollipop-vessel architecture. Both verified verbatim.

**Item 4 — bundled nodes.** All three split. The module's effector becomes `Hepatic Acute-Phase Reprogramming` (TISSUE, hepatocyte) and `B-Lineage Differentiation and Immunoglobulin Output` (CELLULAR, B and plasma cell), with the evidence partitioned to match and the B-lineage edge marked `INDIRECT_KNOWN_INTERMEDIATES` via IL-21. Both MCD entries take the same split. iMCD's marrow node splits into `Bone Marrow Reticulin Fibrosis` and `Peripheral Platelet Consumption` — different claims, and neither is established as causing the other, so bundling them asserted a link the literature does not support. Conformance links and the affected discussion reference were re-pointed.

You were right that the module one mattered most. The two arms act on different cell lineages and their read-outs normalize at opposite ends of the siltuximab response sequence, so a single-valued `biological_scale` could never have been correct for the merged node.

**Item 5 — dropped references.** Added `references:` blocks to both MCD entries carrying the substantive leads, including the Lossos SVEP1 paper and the NRDP overview. These are provenance entries with no `findings` block: neither source has a retrievable abstract in the cache, so no exact quote can be taken and manufacturing one is the failure mode the evidence SOP exists to prevent. The SVEP1 lead is additionally recorded on the renal thrombotic microangiopathy node, where it is a candidate for the unidentified mediator that node already records — and would connect it to the PI3K/AKT/mTOR node, since SVEP1 is an mTOR activator. The HHV-8 somatic-clonality gap (KMT2D, monoclonal microlymphoma) is recorded in that entry's notes for the same reason.

**Item 6 — UCD anemia.** Fixed. Replaced with the UCD-population meta-analysis quote, and the note no longer implies any UCD-specific rate, because none is reported.

---

**Item 3 — done, but with `skos:broadMatch` rather than the `skos:narrowMatch` you suggested.**

Agreed on the substance: the entry had no ontology anchor at all, and that needed fixing. On the predicate I think narrowMatch is the wrong direction and would do real harm.

`MONDO:0019754` is the parent of both the HHV-8-associated and the idiopathic arms, so it is *broader* than this entry. `narrowMatch` would assert the reverse. The repository uses `broadMatch` this way in 52 places, and this entry's sibling uses `narrowMatch` in the correct direction for `MONDO:0018702` (TAFRO, genuinely narrower than iMCD).

The consequence is not just semantic. Per CLAUDE.md, a `mondo_mappings` term counts as curated only when the predicate is `exactMatch` or `narrowMatch`; `broadMatch` is a cross-reference that must not retire the mapped concept from the queue. Using `narrowMatch` here would mark "multicentric Castleman disease" — the parent of both arms — as curated on the strength of one child, which is precisely the harm that removing the `disease_term` binding was meant to avoid. `broadMatch` gives the anchor without the claim.

The mapping carries `consistency: MISSING` with the MONDO conflation recorded, and a note to rebind `disease_term` and switch to `exactMatch` once MONDO mints the class. That new-term request is M1 in #9584.

Happy to switch if you still read the direction the other way.

---

Validation on the pushed head: schema, term and reference validation pass on all four files; 262/262 snippets verbatim; `validate-history-all` clean over 6,759 records; folded-hyphen, snippet-length, title-snippet and duplicate-key gates clean. Full `tests/test_data.py` was still running when I pushed — I'll report it and fix anything it turns up.

---

dragon-ai-agent:
Three updates: the test result I owed you, a correction on finding 2, and a schema migration that landed on `main` while this was open.

**Correction — finding 2 was only half fixed.** Your review named two locations carrying the same MCD-only quote, the node at `:327` and the causal edge at `:236`. I fixed the node, reported the finding as done, and left the edge. It was still asserting `"VEGF expression was highest in PRCs and TRCs of MCD cases."` on a UCD claim in a UCD entry. Now fixed in 44fabe2a5.

I used a different sentence than you suggested. The edge runs `Lymph Node Stromal Cell Activation and Expansion` → `Stromal VEGF Overproduction and Neovascularization`, so what it claims is that activated stroma is the VEGF *source* in unicentric disease. Your `"VEGFA exhibited high expression levels in both UCD and MCD"` establishes the level but names no source, which is the half the edge is actually about, so I took the FDC sentence from the same paper instead — it covers UCD explicitly and names the producing cell. Both are exact substrings of the cache; the one I used is not quoted elsewhere in the file.

That also exposed a stranded binding next door. The node's rewritten description says VEGFA comes from follicular dendritic cells in UCD and puts perivascular and T-zone reticular cells on the MCD side, but `cell_types` still bound `CL:4033054 perivascular cell` — the MCD cell. So the node's ontology grounding contradicted both its own prose and its new evidence. Rebound to `CL:0000442 follicular dendritic cell`.

**Test suite.** 46,864 passed, 0 failures.

**Migration to the narrowed `supports` enum.** [#10003](https://github.com/monarch-initiative/dismech/pull/10003) retired `PARTIAL` and `WRONG_STATEMENT` after this branch was cut, so the three new entries carried 11 items that no longer validated. Applied with `scripts/migrate_supports_enum.py --apply` — its docstring anticipates a branch going red on refresh and names re-running it as the intended fix, so this is the same line-level transformation the rest of the KB got rather than a hand-edit.

I diverged from the bulk migration on five of the eleven. That migration deliberately left `directness` absent everywhere, because an absent value reads as "nobody has assessed this" and a backfill would manufacture appraisals no curator made. That reasoning does not hold for five items here, where the assessment was already made and written into the `explanation` in the curator's own words — "the quote does not name growth delay specifically", "does not describe the imaging appearance itself", "does not itself report IL-6 levels", "comes from multicentric disease rather than UCD", "does not itself address corticosteroids". That is the judgement the new slot exists to hold, so those five carry `directness: INDIRECT` and their prose now says what the inference step is instead of naming a retired token. The other six were `PARTIAL` in the "hedged claim" sense — a quote saying a mechanism is unresolved, cited on a node that models it as unresolved — so the quote bears directly on what the node asserts and they became plain `SUPPORT`.

No `REFUTE` was introduced. The tempting one is a paper arguing against the viral candidate on an iMCD node that lists three candidate initiating processes. It stays `SUPPORT`: `supports` is scoped to the parent object's assertion, and that node asserts the initiator is *unresolved*. Splitting it would repeat the Marfan error that #10003 itself had to revert.

**Still open from your review:** the `skos:broadMatch` vs `skos:narrowMatch` disagreement on finding 3, where I gave my reasoning in the previous comment and am happy to switch if you read the direction the other way. Findings 1, 4, 5 and 6 are addressed as described there.

**Validation on the pushed head:** schema, term and reference validation clean on all five changed KB files; grouping and history validation clean (7,752 records); folded-hyphen, snippet-length, title-snippet, duplicate-key and environmental-evidence gates clean; `tests/test_data.py` 46,864 passed.

---

dragon-ai-agent:
All four remaining items are addressed in 85ba14719 and 1e52cdd6d — the blocking self-loop, both non-blocking nits, and the eight unaccounted references.

**The self-loop is mine and you're right about the cause.** I reproduced it with `build_causal_graph` before touching anything, got the same edge you did, and confirmed it is gone after. Splitting the marrow node was the fix you asked for; naming the new mechanism node identically to the phenotype it points at was my error, and in a flat name-keyed namespace that collapses the two.

The node is now `Marrow Stromal Fibrotic Remodeling` — the stromal process rather than the biopsy finding it produces — and the edge still points at the phenotype, matching the shape the sibling node already had. I also swept all five changed files with the graph builder: 0 self-loops, and no `downstream` target that resolves to neither a node nor a phenotype.

Your framing of this as the same blind spot from the other direction is the useful part. A target that resolves to the *wrong* node is exactly as invisible to `test_entity_ref_foreign_keys` as one that resolves to nothing, and your scan finding 11 files on `main` already carrying a self-loop makes it a KB-wide class rather than a Castleman one. I'll file that together with the exemption issue, with your two scans as the evidence.

**Both nits fixed, and the mantle-zone one had a better answer than either of us assumed.** CL turns out to have an exact term — `CL:0009037 lymph node mantle zone B cell` — so rather than choosing between an over-general binding and dropping `term:` with a needs-term note, it is now bound precisely. `PPR:PPR1286884` has its `reference_title` from the cache header.

**On the eight leftover references — I did not take the offered line in the PR description**, because reading them it would not have been true. All eight are live sources for one of the new entries, not retirements. Five are HHV-8-specific and went to that entry; two are iMCD-specific and went to that one. `DOI:10.3390/ijms25073775` is the notable one: it is the HHV8 monoclonal microlymphoma case your first review flagged as the missing somatic-clonality lead, so it now sits in the entry whose `notes` already record that gap.

`PMID:39970477` went on the grouping instead of a disorder entry. A review titled "one disease, multiple etiologies" *is* the split thesis, so it belongs with the union that records it — that adds the grouping's first `references:` block.

All eight are provenance entries with no `findings`, consistent with the Lossos and NRDP handling you approved: they are leads, and none has a cached abstract from which an exact quote could be taken.

**Two things from your review I'm noting rather than acting on:**

The `DOI:` skip-prefix observation is the more significant one — if no `DOI:` snippet in this PR is machine-verified, that is a large hole in what `count-verified-snippets` appears to certify, and 122 of the 294 snippets here fall in it. Thank you for hand-checking the two. That wants its own issue rather than a change in this PR.

On the CLAUDE.md cancer-granularity line saying `skos:narrowMatch` where it means `broadMatch` — agreed it reads backwards against both SKOS and the 61 existing usages, and it is where your original suggestion came from, so it has already misled once. It is out of scope here; I'd rather flag it than sneak a docs fix into a curation PR. Happy to file it or fix it in a separate PR, whichever the maintainers prefer.

**Validation on the pushed head:** schema, term and reference validation clean on all three edited files plus the grouping; `check-entity-refs` OK across 2,777 KB files; `check-duplicate-keys` OK; graph build clean on all five changed files.

---

dragon-ai-agent:
Merged `main` into the branch (1433097dc). **No content change** — the merge was conflict-free and the diff against `main` is the same curation as before.

Why: the branch had drifted 75 commits behind, and the deterministic auto-merge sweep requires `behind_by == 0`, so it could never have been swept while behind. It is now 0 behind. Five post-gate shepherd runs had passed it over, so I stopped waiting for the LLM lane to update it.

This drops the approval, which is the accepted cost. Re-review should be narrow: nothing in the five changed KB files moved.

Revalidated on the merged tree, against the schema and tooling `main` now carries:

| Check | Result |
|---|---|
| `just validate` (schema + terms + references), 5 changed files | pass |
| `just validate-grouping` | pass |
| `just check-entity-refs` | OK, 2,797 files |
| `just check-duplicate-keys` | OK, 4,266 files |
| `test_causal_targets.py`, `test_enum_values.py`, `test_qualifier_terms.py` | 26 passed |
| Working tree after validation | clean, no cache churn |

Worth noting `main` picked up `tests/test_causal_targets.py` with a baseline while this was open — that is the `downstream[].target` gap from #9697, the one this PR hit twice (a dangling target, then a self-loop). These entries pass it with no baseline entries.

One structural note, not a request. This PR sat approved and green for two days without being swept, falling behind at roughly the rate `main` moves (~5 commits/hour). The sweep needs `behind_by == 0` but the LLM lane updates only one approved-behind branch per run, so a PR that isn't picked promptly drifts faster than the update mechanism can catch it. CLAUDE.md already names this — higher safe throughput "needs a real merge queue (plus `merge_group` CI), not wider update batches." Manually merging `main` is a workaround for one PR, not a fix for that.

---

dragon-ai-agent:
This PR is approved, green and conflict-free, but it is not going to merge on its own. That is not a problem with the PR — it is #10474, where the auto-merge sweep is currently merging zero PRs per run and thirteen approved PRs are stuck behind `main`, this one among them.

I merged `main` in earlier to clear the `behind_by == 0` requirement. That worked, and the PR stayed fully eligible for about nine hours, but three sweeps ran in that window without merging it: one stopped on the main-health circuit while `main` was building, and by the next one `main` had moved and this branch was 7 behind again.

I am not going to keep re-merging. Each attempt drops the approval and buys a window that closes before a sweep reliably lands in it, so repeating it costs review cycles without converging. Leaving it approved and green for a maintainer, or for a sweep that gets lucky.

Nothing here needs further review. The last one verified the branch content is byte-identical to what it approved, re-ran every gate against the moved base including the three new on `main`, and passed 47,208 tests.

---

dragon-ai-agent:
All six items are fixed in 07cf0510f — the three blocking corrections and all three non-blocking carry-overs, in one push.

**vIL-6 edge explanation.** Replaced with main's audited text verbatim. You were right that the quote never mentions vIL-6; I checked the three other places on this branch citing the same sentence and all three stay within what it says, so this was the only one.

**Trial gradings.** Both `clinicaltrials:` items now carry `evidence_source: OTHER`. The point about `status: UNKNOWN` and `RECRUITING` is the decisive one — neither trial has results to be clinical evidence of, so the quoted text can only be a registry description of intent.

**Fever frequency.** Now `FREQUENT`. Fever is a component of the 46.6% composite, so it cannot exceed it. The HHV-8 entry stays `VERY_FREQUENT` on its 98.6% figure.

**`directness: INDIRECT`** restored on both fever items. One deliberate asymmetry: the sibling Splenomegaly item cites the same sentence and is left unassessed, because splenomegaly is reported *directly* in that quote — marking it INDIRECT would be wrong. The same applies to the third item citing that sentence, whose explanation ("quantifies the two features that do differ") is already accurate to it.

**Curation rationale** moved out of the `Marrow Stromal Fibrotic Remodeling` description into that node's `notes`, and the **stray blank line** in the grouping is gone.

**On the history records** — your finding was correct, and I did not fix it the way it might look like it should be fixed. Those records are append-only and describe the sessions as they actually ran, so rewriting their coverage claims would be rewriting history. Instead the two new records added here carry the correction explicitly: each names the 2026-09-04 record and states what its "each finding was checked" claim missed. The overstatement stays on the record, with a correction appended after it.

Validation on the pushed head: schema, term and reference validation clean on both entries and the grouping; both new history records validate; entity-refs (3,059 files), duplicate-keys (4,293), enum-values (3,155), causal-targets, qualifier-terms, snippet-grading, title-snippets, snippet-length and folded-hyphens all clean with no new baseline entries; 2,958 targeted tests pass; working tree clean after validation. `check-snippet-grading --against-ref origin/main` now reports zero new divergences — the seven you saw were the branch being behind, and the merge cleared them.

---

dragon-ai-agent:
CI went red on the curation-scope gate, and it was my fault. Fixed in 40584a4af.

```
docs/reports/data/partial-evidence-directness-worklist-2026-08-30.tsv
```

That file is the worklist `scripts/migrate_supports_enum.py --apply` emits. It was left untracked on purpose when I ran the migration back in round 2 — a generated artifact, and `docs/` is outside what a curation PR may touch. A later automated prompt about untracked files got it committed as 737931638, which is what tripped the gate.

Reverted rather than force-pushed, so anyone with this branch checked out keeps a valid history. The file is out of the PR diff and off disk, so it cannot get swept in again.

The whole diff now sits inside curation scope — `kb/`, `history/`, `cache/`, `references_cache/`, `stubs/` only, verified with the same grep the workflow uses.

Nothing in the round-6 bundle (07cf0510f) changed: the three carried-forward corrections and the three cosmetics are all still there. This revert touches only the stray TSV.

---

PR #9975, opened 2026-08-29
curate: add Familial Progressive Hyperpigmentation With Or Without Hypopigmentation (MONDO:0007771, KITLG)

### LINKED ISSUE #9967: Curate hyperpigmentation with or without hypopigmentation, familial progressive (MONDO:0007771)
Curate a dismech entry for **hyperpigmentation with or without hypopigmentation, familial progressive** ([MONDO:0007771](https://monarchinitiative.org/MONDO:0007771)) — an autosomal dominant pigmentary disorder in which diffuse, progressive hyperpigmented patches (sometimes mixed with hypopigmented macules) appear in infancy and spread with age, attributed to gain-of-function variants in the N-terminal region of *KITLG* (hgnc:6343), the ligand for the KIT receptor that drives melanocyte development, survival, and melanogenesis.

The nomenclature here needs untangling as part of the curation, and it is the main open question:

- MONDO places this term under **familial progressive hyperpigmentation** (MONDO:0013648, FPH1) as a parent, and gives it one descendant, **familial progressive hyper- and hypopigmentation** (MONDO:0017239).
- These three labels have historically been used for overlapping or identical clinical descriptions, and the `KITLG` association has been reported across them. Whether this is one disease with variable hypopigmented component (→ one entry, hyper/hypo as phenotype variation) or genuinely separate entities (→ `has_subtypes`, or a parent entry) should be settled from the literature and recorded explicitly.
- Also worth separating from the phenotypically adjacent *KITLG*-related conditions — dyschromatosis universalis hereditaria and *KITLG*-related deafness — which are distinct entities, not subtypes.

Related existing entries to check against so the new one does not overlap: `Vitiligo.yaml` (depigmentation, unrelated mechanism), and any other `KITLG`/KIT-signaling entries.

Stub: `stubs/Hyperpigmentation_With_Or_Without_Hypopigmentation_Familial_Progressive.yaml`

gingin77:
Adds a knowledge base entry for familial progressive hyperpigmentation with or without hypopigmentation (MONDO:0007771), a rare inherited skin condition in which brown patches appear at or soon after birth and spread with age. It is caused by variants in *KITLG*, the gene for the signal that skin cells send to pigment cells telling them to make melanin. The variants strengthen that signal rather than breaking it, so the pigment cells overproduce.

Closes #9967.

## The scoping decision, which is the part worth reviewing

Three MONDO terms sit close together here with confusingly similar names, and they do not all mean the same thing. The identifiers had to settle it rather than the labels.

| MONDO term | Cross-references | Treated here as |
|---|---|---|
| MONDO:0007771 | OMIM:145250 | **This entry** — the *KITLG* disease |
| MONDO:0017239 | Orphanet:280628, no OMIM | Same disease, Orphanet's view of it — not curated separately |
| MONDO:0013648 | OMIM:614233, Orphanet:79146 | **A different disease** (FPH1) — recorded as a differential |

Two historical labels, FPH and FPHH, distinguish families by whether pale patches accompany the brown ones. Within *KITLG* those labels do not track distinct diseases: the same p.Asn36Ser allele was reported first in a family called FPH and then found in two FPHH families, and every reported allele sits on the same five-residue stretch of the protein. So this entry curates them as one disease with a variable pale-patch component.

That argument stops at *KITLG*. FPH1 (OMIM:614233) is a separate condition mapped to chromosome 19p13.1-pter in a different family, and its gene has never been found (PMID:16709486). It is clinically similar and shares the "familial progressive hyperpigmentation" name, so it is easy to merge by mistake. It is a differential here, not part of the entry.

**One observation for someone with the standing to judge it:** MONDO currently asserts MONDO:0013648 — the 19p, gene-unknown condition — as a *parent* of MONDO:0007771, the *KITLG* condition. On the reading above those are two different diseases at two different loci, so the parent relationship looks wrong. This is recorded in the entry's `notes` as an observation only; no ontology change is proposed and nothing here depends on it being resolved.

## What is in the entry

A seven-node mechanism chain, 49 evidence snippets all verified against cached references, three knowledge-gap discussions, a histopathology section, and differential diagnoses covering FPH1, dyschromatosis universalis hereditaria 2, Legius syndrome, the café-au-lait and lentigines syndromes, and the recessive *KITLG* hearing-loss condition.

```
KITLG gain-of-function variant (VTNNV motif)
    |
    +--> Enhanced KITLG-KIT signaling  [PROVISIONAL]
    |        |
    |        v
    |    Upregulated MITF-driven melanogenic transcription  [PROVISIONAL]
    |        |
    |        v
    |    Increased tyrosinase activity  [PROVISIONAL]
    |        |
    |        v
    |    Increased melanocyte melanin content  [PROVISIONAL]
    |        |
    |        v
    |    Progressive epidermal melanin accumulation  --> phenotypes
    |        |
    +--------+--> Somatic reversion of the KITLG locus  [HYPOTHETICAL]
                      |
                      v
                 Hypopigmented macules and streaks
```

**The confidence grades are deliberately uneven, and two are lower than they look like they should be.**

The step where the mutant ligand binds its receptor more tightly is graded PROVISIONAL, not ESTABLISHED, because no one has measured it. Every functional experiment in this literature reads out melanin or gene expression further downstream; the tighter binding is inferred from homology modelling, and the sources hedge accordingly — PMID:33407466 says a mutated motif "may increase the affinity". The entry says so explicitly rather than presenting the inference as a result.

The pale patches get the lowest grade. There is a genuinely interesting proposal that they are not a pigment defect at all: in one patient whose pale streaks followed the lines of Blaschko, digital PCR found loss of the mutant allele in the pale skin but not in blood, suggesting clones that reverted to normal and look pale only against abnormally dark surroundings (PMID:39269165). That is one patient, and the café-au-lait macules in the same person did not follow the same distribution, so it cannot explain all the dyspigmentation. It is curated as HYPOTHETICAL with a knowledge gap and a proposed experiment.

## Deliberate omissions

**No cosmetic or depigmenting treatments.** Only genetic counselling is curated. Photoprotection, topical lightening agents and Q-switched laser appear throughout the general hyperpigmentation literature but none has been reported in this disease, and adding them would build a treatment section out of inference. KIT inhibition with imatinib is the mechanistically obvious idea — it does suppress melanogenesis — but nobody has given it for this indication, so it stays in prose rather than becoming a curated treatment.

**No datasets.** No repository dataset exists for this condition. Searching the causal gene would surface testicular germ cell tumour and pigmentation GWAS instead, which is the Named Entity Confusion problem `CLAUDE.md` warns about, so dataset discovery was not run.

**No frequency bands on four phenotypes.** The total literature is roughly eight variants across a few dozen patients. There is no denominator to support a frequency, and the one source that ranks the lesions ranks them differently from an earlier draft of this entry, so the bands were dropped rather than guessed.

**No animal or experimental model sections.** A *Kitlg* mouse paper (PMID:41584853) would have been the natural `FAILS_TO_RECAPITULATE` entry, but its cached record has an empty body, so there is nothing quotable to cite.

## Provenance and validation

Deep research was requested from falcon, which is not configured in this environment. The run used `--fallback` so the substitution is recorded in the report frontmatter rather than buried in prose, and `claude_code` produced the report. Its own validation flagged `needs_review: true` on both checks — 15 of 45 quotes unsupported, two obsolete GO terms — so nothing was copied across from it. Every snippet in this entry was taken directly from `references_cache` and every ontology term was resolved through OAK.

An adversarial review pass before this PR opened found three things the validators cannot catch, all fixed:

- Eleven evidence items carried a `reference_title` that paraphrased PMID:33407466's content instead of quoting its title, rather than copying it from the cache frontmatter.
- A phenotype asserting the disease is skin-limited contradicted the entry's own oral mucosal pigmentation phenotype, and generalised a negative from one pedigree into a disease-level claim. It was deleted and replaced with a knowledge gap presenting both the negative series and the contrary reports.
- The receptor-binding node was graded ESTABLISHED while every evidence item attached to it tested the normal, non-mutant version of the pathway.

Also corrected in that pass: `HP:0000163` replaced with `HP:0010284` (Intra-oral hyperpigmentation) for the oral phenotype; `GO:0048066`, a developmental-process term, replaced with `GO:0043473` on a node about lifelong postnatal progression; a node bundling two claims split into separate molecular and cellular nodes; the grafted-skin evidence re-graded `IN_VITRO` to `MODEL_ORGANISM`; and the reversion node wired into the graph, which had left it disconnected.

Checks: `just validate-disorders` passes; 49/49 snippets verified; `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all clean. Weighted compliance is 81.7%, down from 84.4% before the review pass — that reflects removing an over-claimed phenotype and adding sections with real gaps, not a regression.


gingin77:
CI was red on `check_reference_titles.py` and is now fixed.

Both `reference_title` values for PMID:29186243 were written from memory instead of copied from the cache frontmatter, so they described the paper's content rather than naming it (similarity 0.43):

```
curated: Clinical and genetic features of a Chinese family with familial progressive hyper- and hypopigmentation.
 cached: Familial progressive hyper- and hypopigmentation: a report on a Chinese family and evidence for genetic heterogeneity.
```

Replaced with the cached title verbatim in `histopathology[0].evidence[0]` and `genetic[0].evidence[4]`.

One correction to the PR description, which I have also edited: it claimed `EvidenceItem.reference_title` is not machine-checked. That is wrong — `scripts/check_reference_titles.py` checks it against the cache frontmatter and runs in CI, which is exactly what caught this. I had run the six checks listed in `CLAUDE.md` before opening the PR and this one is not among them, so I missed it locally.

`just check-reference-titles` now passes, `just validate-disorders` passes with 49/49 snippets verified, and the full `tests/test_data.py` suite passes locally (45,748 tests).

gingin77:
All five requested changes are in, plus four of the suggestions.

| Item | Status |
|---|---|
| (a) `reference_title` for PMID:29186243 | Fixed in the previous push |
| (b) Histopathology misstates its source | Fixed — two evidence items added, prose corrected |
| (c) Knowledge gap claimed but not present | Fixed — discussion added |
| (d) Countervailing evidence for the reversion node | Fixed — added as `REFUTE` |
| (e) Truncated sentence in `description` | Fixed |

**On (b) and (c) — you were right and the correction is substantive.** The prose said the source "does not settle whether the excess pigment reflects more melanocytes or harder-working ones", and the source does speak to it: increased melanocytes by both S100 and HMB45. Both sentences are now evidence items, including the paper's own note that S100 is not melanocyte-specific, which is what makes the HMB45 result carry the weight. There is now a `fphh_hyperplasia_vs_hyperfunction` discussion attached to the melanogenesis node and the histopathology section, with a melanocyte-density experiment as the resolving test. The pathograph still models per-melanocyte hyperfunction, because that is what the KITLG functional experiments measure, but the entry now says so as a choice rather than describing the counter-evidence as silence.

**On (d)**, graded `REFUTE` rather than `PARTIAL`. The reversion model predicts pale skin containing revertant melanocytes pigmenting normally, which requires melanocytes to be present; both markers being almost absent contradicts that prediction directly. The explanation records the two limits — different family, no KITLG mutation found, and nobody has stained a lesion demonstrated to carry copy-neutral LOH.

**Suggestions taken.** The K14-SCF mouse turned out to be findable — Kunisada et al., PMID:9584135, now fetched and curated as an `animal_models` entry with `modeled_mechanisms`, `PARTIALLY_RECAPITULATES` / `MODERATE`, and a melanocyte-population readout. It is worth having for the reason you flagged: keratinocyte-targeted KITLG expression produces epidermal melanocytosis alongside the hyperpigmentation, so it is a third line pointing the same way as the immunohistochemistry, and it is cited in the new discussion. Its limitations slot records the three gaps — wild-type overexpression rather than an altered ligand-receptor interface, a baseline where mouse interfollicular epidermis normally has no melanocytes at all, and the mastocytosis phenotype that FPHH does not have.

Also done: sun exposure added as an environmental modifier with `influences_mechanisms` (`EXACERBATES`), `p.Lys38Thr` added to the allele list, the `PROVISIONAL`/`LIKELY` prose mismatch aligned, and the mixed-source evidence item split into a `MODEL_ORGANISM` grafted-skin arm and an `IN_VITRO` explanted-skin arm.

Not done: human explanted skin as its own `experimental_models` entry. The only description of it available here is secondhand, inside the PMID:33407466 review summarising work it cites, so a model entry would rest on a quote about someone else's experiment. It is captured as evidence on the signalling node instead. Happy to add it if you would rather have the model entry with that caveat recorded.

One thing worth flagging beyond this PR: the reference-title mismatch in item (a) was caught by `scripts/check_reference_titles.py`, which is a CI step but is not among the checks listed in `CLAUDE.md`'s evidence and reference workflow. Running exactly what is documented still let a red build through. Not touching it here since it is a docs change unrelated to this curation.

`just validate-disorders` passes, 58/58 snippets verified, `check-reference-titles`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-environmental-evidence` all clean, and `tests/test_data.py` passes locally (45,748 tests).

PR #9992, opened 2026-08-29
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 93 (MONDO:0013963, CABP2)

### LINKED ISSUE #9978: Curate autosomal recessive nonsyndromic hearing loss 93 (MONDO:0013963)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 93 (DFNB93)** ([MONDO:0013963](https://monarchinitiative.org/MONDO:0013963)) — biallelic *CABP2* variants; calcium-binding protein 2 modulates Cav1.3 channels at the inner hair cell ribbon synapse, giving moderate hearing loss that is milder than most DFNB forms.

Nothing in `kb/` mentions CABP2. Note this is a synaptic rather than a hair-bundle mechanism, so check whether `kb/disorders/Auditory_Neuropathy.yaml` should relate to it before treating it as unconnected.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93.yaml`

kevinschaper:
Adds DFNB93, the CABP2 form of autosomal recessive nonsyndromic hearing loss.

**What the disease is.** CaBP2 sits presynaptically in the cochlear inner hair cell and holds the CaV1.3 calcium channel open — it suppresses the channel's inactivation, so a hair cell held depolarised by an ongoing sound keeps enough channels available to keep releasing transmitter. Lose CaBP2 and the channels inactivate, sustained exocytosis collapses, and spiral ganglion neurons fire less both spontaneously and to sound. Outer hair cells are never involved, so cochlear amplification is intact and otoacoustic emissions are present while the auditory brainstem response is abnormal.

That dissociation is the definition of auditory synaptopathy, and the paper that established the mechanism says so in its own conclusion: "Therefore, we conclude that human deafness DFNB93 is an auditory synaptopathy."

**entry_type: DISEASE.** The issue asked whether this should relate to `Auditory_Neuropathy`, and it should — but as a member of that mechanistic class, not as a `has_subtypes` entry on it. DFNB93 has its own MONDO term, its own OMIM entry and DFNB number, and a pathograph that is gene-specific all the way down to the CaV1.3 channel. What I did instead of a subtype record was assert the class membership inside the pathophysiology section, citing the two publications that state it, so the relationship is machine-visible rather than sitting in a prose note.

**Gene-disease validity.** ClinGen's Hearing Loss Gene Curation Expert Panel rates CABP2–nonsyndromic hearing loss **Definitive**, recorded as an `external_assertions` entry citing a newly generated `CGGV_*.md` cache file. This is worth stating because the other three loci in this batch are not: the same panel rates TSPEAR and GJB3 **Disputed** and MET **Limited**.

**Animal models.** Two, both in `animal_models:` with `modeled_mechanisms`:
- the `Cabp2 LacZ/LacZ` single knockout — RECAPITULATES the channel node (patch-clamp readout) and the organism-level dissociation (an `UNCHANGED` amplification readout beside a degraded ABR readout);
- the `Cabp1/Cabp2` double knockout with AAV rescue — RECAPITULATES the exocytosis node, and RESCUES the organism node. Its limitation says plainly that two genes are removed where patients lose one, so its magnitudes are not DFNB93 magnitudes.

**Two things I recorded rather than resolved.**

Hearing aids. The class-level literature says auditory synaptopathy patients gain little from amplification; the individual DFNB93 case reports describe amplification being fitted with good outcomes. Both are cited, neither is dropped, and the tension is a `KNOWLEDGE_GAP` with a proposed experiment (aided speech perception vs aided threshold in a genotyped cohort). The treatment carries a free-text `preferred_term: hearing aid usage` with no `term:`, following the repo convention — NCIT's hearing-aid concept is a device, not reachable from `NCIT:C25218`.

Progression. The mouse is progressive; the human reports are all cross-sectional. So the entry asserts onset and severity and deliberately asserts no `clinical_course`, with a second `KNOWLEDGE_GAP` saying why.

**Deliberately not curated.** No `clinical_trials` (none exist). No `datasets` (no CABP2-specific dataset found, and a gene-symbol search returns the retinal CaBP literature — the Named Entity Confusion route the dataset docs warn about).

**Validation.** `just validate` exit 0 · `just validate-terms` exit 0 · `just validate-disorders` exit 0 with **42/42 snippets verified** · `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-term-cache-integrity` all exit 0 · `validate-history` exit 0.

Closes #9978

### DIFF
--- a/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93.yaml
+++ b/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93.yaml
@@ -501,10 +501,24 @@ phenotypes:
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: >-
+      This finding indicated preserved outer hair cell function, identifying the first
+      individual with DFNB93 precisely diagnosed with an auditory synaptopathy.
+    explanation: >-
+      The human anchor: a genetically confirmed DFNB93 child diagnosed with auditory
+      synaptopathy, which is by definition an abnormal auditory brainstem response alongside
+      preserved emissions.
+  - reference: PMID:40927552
+    reference_title: "Is CABP2-Associated Hearing Loss (DFNB93) a Gene Therapy Target? Preclinical Progress and Patient Registry."
+    supports: SUPPORT
+    evidence_source: MODEL_ORGANISM
+    snippet: >-
       reduced auditory brainstem responses and increased hearing thresholds despite preserved
       distortion product otoacoustic emissions corroborating auditory synaptopathy
     explanation: >-
-      States the abnormal brainstem response and the direction of the threshold change.
+      The direction of the threshold change, and where the polarity of the model readout comes
+      from. Graded MODEL_ORGANISM because the sentence it is cut from describes a mouse with a
+      deletion of Cabp2 exons 3 and 4; it corroborates the human phenotype above rather than
+      being the evidence for it.
 
 - category: Otologic
   name: Preserved Otoacoustic Emissions
@@ -561,6 +575,22 @@ phenotypes:
       id: HP:0012781
       label: Mid-frequency hearing loss
   evidence:
+  - reference: PMID:40927552
+    reference_title: "Is CABP2-Associated Hearing Loss (DFNB93) a Gene Therapy Target? Preclinical Progress and Patient Registry."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "in three consanguineous Iranian families presenting moderate to severe autosomal recessive hearing impairment with characteristic U‐shaped audiograms"
+    explanation: >-
+      The rule stated directly of the founding families, rather than inferred from a report of
+      an exception to it.
+  - reference: PMID:40927552
+    reference_title: "Is CABP2-Associated Hearing Loss (DFNB93) a Gene Therapy Target? Preclinical Progress and Patient Registry."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "revealed mid‐frequency, moderate‐to‐severe hearing loss, and transitory evoked otoacoustic emissions"
+    explanation: >-
+      An independent case with mid-frequency loss, which is the audiometric description this
+      phenotype's HP term names.
   - reference: PMID:42448438
     reference_title: First reported North American calcium binding protein 2-related non-syndromic hearing loss.
     supports: SUPPORT
@@ -569,9 +599,8 @@ phenotypes:
       Audiological testing revealed a reverse-slope pattern of hearing loss, differing from
       the U-shaped pattern associated with previous reports of DFNB93.
     explanation: >-
-      Names the U-shaped pattern as the one associated with the previously reported DFNB93
-      families. The quote establishes the rule in the course of reporting an exception to it,
-      which is the only place it is stated explicitly in the cached sources.
+      Confirms the U-shaped pattern as the previously reported norm, from the report of the one
+      case that departs from it.
 
 - category: Otologic
   name: Reverse-Slope Audiometric Configuration
@@ -879,7 +908,7 @@ animal_models:
   modeled_mechanisms:
   - target: Loss of CaBP2 Function at the Inner Hair Cell Ribbon Synapse
     relationship: RESCUES
-    fidelity: HIGH
+    fidelity: MODERATE
     description: >-
       Delivering the Cabp2 coding sequence into inner hair cells restores the missing channel
       regulator in the genotype that matches the human disease, recovering the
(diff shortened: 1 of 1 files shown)


kevinschaper:
All six blocking items fixed, plus the artifact. I verified each against the caches before changing anything — three of the six I'd got wrong in ways I could confirm from the repo, which is the useful kind of review.

**1. ABR polarity — correct, fixed.** The readout is a *threshold*, so an impaired ABR is an elevated one. `references_cache/PMID_40927552.md` says it directly: "reduced auditory brainstem responses and increased hearing thresholds". Now `INCREASED`, and the interpretation says why so the next reader doesn't have to re-derive it.

**2. OAE term binding — correct, split.** HP:0006958 named a different finding than the node claimed. Now two phenotypes: an abnormal-ABR one that legitimately carries HP:0006958, and a preserved-OAE one with **no `term:`**, plus a `notes` block recording that HPO codes only the negation (HP:6000182) and that an NTR for a preserved-emissions term is the right upstream fix. This also has the side benefit of keeping it out of the KGX `has_phenotype` edge, which is the same trap that bit the DFNB97 entry in this batch.

**3. OAE frequency — correct, now OCCASIONAL.** I found the sentence you meant: "Measurement of otoacoustic emissions, so far, have been rarely done (Table 2), making comprehensive auditory phenotyping a high priority for further investigations." It is cited as a `PARTIAL` qualifier on the phenotype, and the description now distinguishes what the mechanism predicts from what has been measured. (I could not find the "replication studies … imperative" wording in the cache, so I did not quote it.)

**4. GO term — correct, fixed.** GO:0007269 neurotransmitter secretion described a spike-rate node. Now GO:0019228 neuronal action potential.

**5. Audiogram rule vs exception — correct, and this is the one I'd most have wanted caught.** Curating the reverse-slope case without the U-shaped rule does invert the diagnostic signal. Added HP:0012781 Mid-frequency hearing loss, quoting the sentence that names the U-shaped pattern as the previously reported one. One note on your locations: the U-shape statement is in `PMID_42448438.md`; I searched `PMID_40927552.md` for "U-shape" and "mid-frequenc" and found neither, so the mid-frequency phenotype rests on the 42448438 quote alone — which states the rule in the course of reporting an exception to it, and is the only place it's stated explicitly in the cached sources.

**6. GJB2 snippet — correct, and it was a false source-defect claim.** `PMID_31661684.md:43` carries the sentence unbroken. Snippet extended to the full sentence and the bogus "line break in the source" explanation removed. You're right that this matters because the entry's two *genuine* source defects (`CABP2variantsalso`, the transposed `CAPB2`) are real, and a false one beside them devalues both.

**7. Deep-research artifact — committed.** `research/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93-deep-research-falcon.md` plus sidecar and artifacts. Its preflight was a clean **PASS** (CABP2=54 mentions, no rival gene, no OMIM mismatch) — the only one of the four in this batch that was.

**Also folded in from the issue's second-opinion comment:** PMID:34489639, the Cabp2 gene-therapy rescue in the *single*-knockout mouse. That is a better citation than the double-knockout rescue the treatment previously leaned on, because it is the genotype that matches the human disease and the delivery route is a round-window AAV injection that could plausibly be used in a patient. Added as its own `animal_models` entry with a `RESCUES` link and a `RESTORED` readout, and promoted to primary evidence on the gene-therapy `target_mechanisms`; limitations record that the rescue is partial, unilateral, given before hearing onset, and assessed only to 5–8 weeks.

45 → **48/48 snippets verified**; all gates re-run at exit 0.

kevinschaper:
Caught cleanly, and you're right that the fix introduced it. I checked `references_cache/PMID_40927552.md:319` before changing anything: the sentence does begin "A mouse model with a deletion of Cabp2 exons 3 and 4 showed…". A human phenotype at FREQUENT resting on a model-organism sentence graded as human evidence is a real defect, not a technicality.

**Fixed both ways you suggested:**

- The mouse quote is regraded `MODEL_ORGANISM`, with an explanation saying where it comes from and that it corroborates the human phenotype rather than being the evidence for it.
- Added the human anchor from line 317: "This finding indicated preserved outer hair cell function, identifying the first individual with DFNB93 precisely diagnosed with an auditory synaptopathy." A genetically confirmed DFNB93 child diagnosed with auditory synaptopathy *is* an abnormal ABR with preserved emissions, so the phenotype now has a human basis and FREQUENT stands.

**And I took the non-blocking audiogram suggestion, which turned out to matter more than a suggestion.** You were right that line 317 has the rule directly, and my earlier grep had missed it because the source spells it with a U+2010 hyphen. The mid-frequency phenotype now cites the founding families directly — "in three consanguineous Iranian families presenting moderate to severe autosomal recessive hearing impairment with characteristic U‐shaped audiograms" — plus the Italian child's mid-frequency loss. The reverse-slope quote stays as confirmation rather than as the basis for the rule, which is the right way round.

**Two more of your suggestions taken:**

- `fidelity: HIGH` on the gene-therapy RESCUES link is inconsistent with its own limitations (partial, unilateral, given before hearing onset, assessed only to 5–8 weeks). Lowered to `MODERATE`.
- The RESCUES readout now says plainly that CaBP2 function is not measured directly but through its only known effect on CaV1.3 inactivation, and why the readout target repeats the link target.

48 → **51/51 snippets verified**; `just validate`, `just validate-terms`, `just validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens` all exit 0.

The remaining suggestions I did not take (newer patient count, auditory fatigue as a third gap, the exon 3–4 mouse as a separate model, an `Auditory_Neuropathy` backlink) are all additive rather than corrective, and the backlink in particular is an edit to a file this PR does not otherwise touch. Happy to do any of them if you'd rather they land here than in a follow-up.

PR #9993, opened 2026-08-29
curate: Add Combined immunodeficiency due to CD3gamma deficiency (MONDO:0014276, CD3G)

### LINKED ISSUE #9984: Curate combined immunodeficiency due to CD3gamma deficiency (MONDO:0014276)
Curate a dismech entry for **combined immunodeficiency due to CD3gamma deficiency** ([MONDO:0014276](https://monarchinitiative.org/MONDO:0014276)) — biallelic *CD3G* variants; loss of the CD3γ chain impairs T-cell receptor assembly and surface expression, giving a combined immunodeficiency with a notably variable course, from early-onset severe disease to relatively mild autoimmunity.

Nothing anywhere in `kb/` mentions CD3G, so this is a clean start. Check whether an existing combined-immunodeficiency or IEI grouping should gain it as a member once the entry exists.

Stub: `stubs/Combined_Immunodeficiency_Due_To_CD3gamma_Deficiency.yaml`

kevinschaper:
Adds `kb/disorders/Combined_Immunodeficiency_Due_To_CD3gamma_Deficiency.yaml` and deletes the stub it curates.

## What this disease is, and why it gets its own entry

`entry_type: DISEASE`. CD3gamma deficiency is the CD3 chain deficiency that does not cause SCID. Biallelic *CD3G* null variants impair assembly of the TCR/CD3 complex and lower surface receptor density, but thymocyte development survives — patients have normal absolute T-cell numbers with reduced surface TCR/CD3 — and the dominant clinical problem is autoimmunity rather than infection. That gives one conserved pathograph: null variant, impaired complex assembly, reduced surface receptor, attenuated signalling, distorted thymic selection and regulatory T-cell failure, then autoimmunity and infection susceptibility as parallel outcomes.

It is not a subtype of anything already in `kb/`. Nothing anywhere in the knowledge base mentioned *CD3G* before this PR, and MONDO places MONDO:0014276 under combined immunodeficiency rather than under a curated parent entry.

## Named Entity Confusion check

`just preflight-dr` PASS on the falcon report: *CD3G* mentioned 40 times and the top gene by a wide margin, OMIM 615607 consistent with MONDO.

Preflight keys on the gene, so it cannot see the confusion that actually matters here — the other CD3 chain deficiencies (*CD3D*, *CD3E*, *CD3Z*). Those are severe combined immunodeficiencies with a complete developmental block, and they are the diseases most of the older "CD3 deficiency" literature is about. Every citation was checked to be about CD3gamma specifically. The CD3delta/CD3epsilon material appears once, in `differential_diagnoses`, where it is the point.

## The mouse does not model this disease, and that is curated

The `Cd3g`-null mouse is not a mild model of a mild disease; it is a severe model of a mild disease. Thymic cellularity falls below 1% of normal and both T-cell lineages fail, against patients who keep normal T-cell numbers. So the animal model carries two links:

- `PARTIALLY_RECAPITULATES` the receptor-expression node, with a measured readout — surface TCR and CD3epsilon are genuinely reduced.
- `FAILS_TO_RECAPITULATE` the thymic-selection node, with `limitations` and three pieces of evidence, including the structural reason: the human gamma-delta TCR incorporates CD3delta and the mouse one does not, so CD3delta can substitute for the missing chain in patients but not in mice.

Paired with a `HUMAN_MODEL_MISMATCH` discussion that also records why the human cell-line models do not close the gap (knockdown Jurkat cells express under 11% of normal surface TCR; patient T cells express over 30%). A second `KNOWLEDGE_GAP` records that patients homozygous for the same allele have had opposite outcomes, including within one family, so nothing currently predicts course at diagnosis.

## Evidence grading

Stated in the entry `notes` and applied consistently: patient-sample measurements (flow cytometry, repertoire sequencing, clinical course) are HUMAN_CLINICAL; anything needing culture or an engineered line (PHA-driven T-cell blasts, regulatory T-cell suppression co-cultures, Jurkat and JGN lines) is IN_VITRO; mouse data are MODEL_ORGANISM.

Snippets from the 2026 Frontiers review are quoted from the cached PDF extraction, which renders Greek letters as Latin characters in places ("CD3 g", "abTCR"). The snippets reproduce the cached text verbatim, as exact-quote validation requires; entry prose spells the Greek names out. That is noted in the entry.

## Added in review

Bronchiectasis (node, phenotype, and the indication for a new antimicrobial-prophylaxis treatment) and glucocorticoids, the latter with two `target_mechanisms` links — `INHIBITS` on the autoimmunity it is given for, and `MODULATES` on `Attenuated T-Cell Receptor Signaling`, the node it makes worse. Also IgE/eczema phenotypes, two biochemical entries, and a signpost on the one place where two sources disagree about a patient's mitogen response.

## Deliberately not curated

- No `clinical_trials`: no trial is registered for this disorder.
- No `datasets`, no `environmental`, no `biochemical` reference ranges.
- No Orphanet citation. ORPHA:466926 and ORPHA:169082 are both absent from `references_cache/`, and refreshing the Orphadata bulk XML to generate one record was not worth the download on a disk at 90%.
- Membership of `kb/groupings/Combined_Immunodeficiencies_IEIs.yaml` is a reasonable follow-up but touches another file, so it is left for a separate PR.
- A TREC newborn-screening caveat. It is a plausible inference from the near-normal T-cell numbers, but the only TREC datum in any cached reference is a patient with *reduced* TRECs, which points the other way. Not asserted without a source.

## Validation

All exit 0: `just validate` (86/86 snippets verified after the review fixes), `just validate-terms`, `just validate-disorders`, `just validate-history`, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

Closes #9984

### DIFF
--- a/kb/disorders/Combined_Immunodeficiency_Due_To_CD3gamma_Deficiency.yaml
+++ b/kb/disorders/Combined_Immunodeficiency_Due_To_CD3gamma_Deficiency.yaml
@@ -123,7 +123,7 @@ inheritance:
 prevalence:
 - population: Worldwide
   measure_type: CASES_IN_LITERATURE
-  prevalence_class: BELOW_1_IN_1000000
+  prevalence_class: ULTRA_RARE
   notes: >-
     No population estimate exists. A 2026 review that reassessed the whole
     literature counted 18 genetically confirmed patients, 16 previously reported


kevinschaper:
Both blocking items are fixed, and five of the six suggestions. One is declined, with the reason below.

## Blocking

**Bronchiectasis.** You were right that it was missing and right about why it matters. It is now three things: a pathophysiology node downstream of `Susceptibility to Recurrent and Chronic Infection`, a phenotype with the 4/5 frequency, and the stated indication for a new `Antimicrobial therapy and prophylaxis` treatment. That closes the loop you pointed at — the reason prophylaxis runs indefinitely rather than episode by episode. Snippet copied verbatim from the cache including the ligature.

**Glucocorticoids.** Added with two `target_mechanisms` links, because the drug does two different things here:

- `Multisystem Autoimmunity` / `INHIBITS` — the intended effect, with the recorded fact that efficacy was limited and rituximab is what actually controlled the cytopenias and allowed steroid tapering.
- `Attenuated T-Cell Receptor Signaling` / `MODULATES` — the harmful one.

On that second link: `TreatmentEffectEnum` has no value for a drug that *worsens* its target node. `INHIBITS` reads as inhibiting the attenuation, which is the opposite of what happens. I used `MODULATES` and stated the direction explicitly in the link description rather than encoding a value that means the wrong thing. Flagging it as a possible schema gap rather than quietly picking the least-bad enum.

Sirolimus is on the same treatment as a second `therapeutic_agent` (quotable from PMID:33215322). `NCIT:C2322` does not validate as a `ChemicalEntityTerm` — it is not reachable from `NCIT:C1909`, and `E-Cigarette_or_Vaping_Product_Use-Associated_Lung_Injury.yaml:942` already carries a note that OAK resolves that ID to "Cranial Nerve Neoplasm". `CLAUDE.md` lists it as the corticosteroid example, which looks like a documentation bug worth a separate issue. I used `CHEBI:24261` glucocorticoid, matching `Adult-Onset_Still_Disease`.

## Suggestions taken

- **Item 3, mitogen tension.** Real contradiction, now signposted. The node says the qualifier out loud: the CVID-like adult was described as having normal PHA proliferation by his own report, and the 2026 review scored that same patient as reduced. Both quotes stay, from their own sources.
- **Item 4, biochemical.** Two entries: serum IgG with the 338 mg/dL value and the 9/18 frequency, and CD4+ count with the 627 cells/µL value paired with the normal total T-cell number that defines the disorder. No `reference_ranges` block — the sources give patient values without stating the intervals.
- **Item 6.** Elevated IgE and atopic dermatitis added, 2/5 each, from the snippet already quoted.
- **Item 7.** IVIG regraded to `PROTEIN_REPLACEMENT`.
- **Item 9.** Both PMID:31921117 sentences regraded to `HUMAN_CLINICAL`. That prompted a clearer statement of the convention in `notes`: grading follows what the quoted sentence asserts, not only where the measurement was taken, so a co-culture suppression assay is `IN_VITRO` while a sentence whose subject is the patient's clinical state is `HUMAN_CLINICAL` even when it cites an assay as support.

## Declined, with the reason

**Item 5, the TREC newborn-screening caveat.** I could not source it, and the one piece of evidence in the caches points the other way. The only TREC mention in any cached reference is `PMID_42661620.md`, patient 9/F6-1, whose immunological findings include "low T cells; reduced TREC". A patient with reduced TRECs is one a TREC screen might have flagged. Asserting that a normal TREC screen does not exclude CD3G deficiency would be a plausible inference from the near-normal T-cell numbers, but it is an inference, and the only datum available is a counter-example. Happy to add it if you have a source.

**Item 8, the five uncited DOI caches.** Left in place. They are the deep-research run's own reference-resolution outputs, and the research report is committed in this PR; removing them would make its "6/6 references resolved" no longer reproducible from the repository. Agreed they are noise, but the noise is load-bearing for that artifact.

**GeneReviews:** confirmed none exists. `("CD3G"[All Fields] OR "CD3 gamma"[All Fields]) AND GeneReviews[Book]` returns zero, and `GeneReviews[Book]` on its own returns results, so the field spelling is right and the zero is real.

All exit 0: `just validate` (86/86 snippets verified, up from 72), `just validate-terms`, `just validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

kevinschaper:
Thanks. Filed the schema gap as #10004 — `TreatmentEffectEnum` has no counterpart to `EnvironmentalEffectEnum`'s `EXACERBATES`, which is why the glucocorticoid link onto `Attenuated T-Cell Receptor Signaling` needed `MODULATES` plus prose. The issue notes it will recur for any treatment whose therapeutic target and toxicity target are the same node.

Leaving the other three open items as you scoped them: the bronchiectasis `biological_scale` is a judgement call either way, the TREC caveat stays out until someone can source it, and the DOI caches stay so the research artifact's reference resolution remains reproducible. Not pushing again on an approved PR for any of them.

PR #9994, opened 2026-08-29
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 98 (MONDO:0013929, TSPEAR) as a contested entity

### LINKED ISSUE #9977: Curate autosomal recessive nonsyndromic hearing loss 98 (MONDO:0013929)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 98 (DFNB98)** ([MONDO:0013929](https://monarchinitiative.org/MONDO:0013929)) — biallelic *TSPEAR* variants; TSPEAR is an EAR-domain protein of the extracellular matrix, and the same gene also causes ectodermal dysplasia 14, so the hearing-loss-only presentation is the entity to keep in view.

Nothing in `kb/` mentions TSPEAR. Watch the two-phenotype split when reading any deep-research report — a citation about tooth agenesis belongs to the ectodermal dysplasia entity, not here.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_98.yaml`

kevinschaper:
Adds DFNB98, the TSPEAR form of autosomal recessive nonsyndromic hearing loss — curated as a **contested** entity rather than an established one, which is the substance of this PR.

**The finding.** ClinGen's Hearing Loss Gene Curation Expert Panel classifies TSPEAR–nonsyndromic hearing loss as **Disputed**. It first evaluated the pair in 2017, re-evaluated it on 2024-10-31 under SOP v11 having read the newer case reports, and the classification did not change. The panel's own words: "In summary, the relationship between TSPEAR and autosomal recessive nonsyndromic hearing loss has been disputed."

That is not thin evidence — it is evidence pointing elsewhere:

- The largest deliberate test of the hearing claim found hearing loss in 3 of 13 individuals with biallelic TSPEAR variants, and **all three also carried variants in TMPRSS3, GJB2 or GJB6** — ordinary causes of the same phenotype.
- One published homozygous nonsense proband did not segregate: the variant was heterozygous in an affected sibling and in the unaffected mother.
- A 2021 mid-frequency SNHL series found a TSPEAR variant and classified it a VUS, not the cause.
- A 2025 CRISPR mouse carrying a TSPEAR frameshift has **normal stereocilia and normal auditory function**.

What survives is the 2012 founding Iranian family, a 2026 Iranian child with profound loss and teeth/skin/hair/nails specifically examined and normal, and the founding study's localisation of TSPEAR protein to the hair bundle surface.

**entry_type: DISEASE, curated as contested.** Concretely that means:
- `genetic[].relationship_type: DISPUTED` (a real schema value), not `CAUSATIVE` — so a query that treats a biallelic TSPEAR finding as a molecular diagnosis of deafness will not get that from this entry;
- the proposed hair-bundle mechanism lives in a `mechanistic_hypotheses` group with `status: ALTERNATIVE`, and every causal edge in `pathophysiology` opts into it via `hypothesis_groups`, so no step is asserted as canonical;
- refuting evidence sits in the same sections as the supporting evidence rather than being quarantined in a note;
- the mouse is `FAILS_TO_RECAPITULATE` with three readouts (two `UNCHANGED`, one `ALTERED` for the Notch/Wnt perturbation that keeps it from being a pure null), plus the required `limitations` and `evidence`.

I considered marking the stub `OUT_OF_SCOPE` instead. A Disputed relationship seems worth recording — deleting the entity would delete the dispute along with it, and leave MONDO's term for the next person to take at face value.

**Named entity separation (the issue's warning).** TSPEAR's established disease is autosomal recessive ectodermal dysplasia 14, and its literature is larger than DFNB98's. Preflight flagged the deep-research report for citing OMIM 618180 (ECTD14) against a MONDO term that xrefs 614861. Nothing dental — tooth agenesis, conical cusps, enamel knots, hypohidrosis, sparse hair — is curated as a DFNB98 phenotype. ECTD14 appears only under `differential_diagnoses`, and the ~1-in-140 non-Finnish-European carrier rate is quoted **there**, attributed to that entity, so it cannot later be mistaken for a DFNB98 prevalence.

The one case where the boundary genuinely blurs is the 2024 Chinese proband, who had congenital SNHL *and* dental caries, ankyloglossia and a heart-shaped tongue. I recorded it with its oral findings rather than tidying it into one entity, because the published report does not resolve it either.

**One thing I deliberately did not cite.** A 2023 TSPEAR delineation study is reported at second hand (by the deep-research report) to state that evidence is insufficient to link TSPEAR to recessive hearing loss. Only its abstract is retrievable and the abstract does not contain that statement, so the quote appears nowhere in this entry. The refuting case is built entirely from sources whose text is in `references_cache/`.

**Two discussions.** A `KNOWLEDGE_GAP` on whether TSPEAR causes isolated hearing loss at all, and a `HUMAN_MODEL_MISMATCH` on why the mouse hears normally when the localisation that motivates the mechanism was made in that same species. Both carry proposed experiments whose refuting outcome is, explicitly, that the right curation result would be retiring the entity.

**Deliberately not curated.** No treatments — the only management is generic hearing habilitation with no DFNB98-specific evidence, and a curated treatment implies a curated indication. No clinical trials, no datasets, no prevalence figure (the numerator is exactly the disputed cases).

**Validation.** `just validate` exit 0 · `just validate-terms` exit 0 · `just validate-disorders` exit 0 with **35/35 snippets verified** · `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-term-cache-integrity` all exit 0 · `validate-history` exit 0.

Closes #9977


kevinschaper:
All three blocking items fixed, plus the suggestion you flagged as most valuable.

**1. Deep-research artifact — committed.** `research/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_98-deep-research-falcon.md` with sidecar and artifacts. Its preflight was a **WARN**, and the warning is the interesting part: the report cites OMIM 612920 and 618180 where MONDO:0013929 xrefs 614861. 618180 is ECTD14 — the report drifted toward the sibling entity exactly as the curation issue predicted. That is why it was used as leads only and every citation checked individually.

**2. phenotypes[2] — correct on both counts.** HP:0000157 was backed by a quote about dental caries, and my explanation also misdescribed where the truncation fell. Rebound to **HP:0010296 Ankyloglossia**, as you suggested, on a snippet that actually reports the tongue finding: "After three corrective surgeries, her tongue was still difficult to stick out and her speech was still compromised." The caries and heart-shaped tongue in the same case are not separately curated.

**3. Pathograph did not reach the phenotype — correct, fixed.** `Impaired Cochlear Sound Transduction` now has a `downstream` edge to `phenotypes#Congenital Profound Sensorineural Hearing Loss`, following the DFNB104 convention you pointed to. It carries `hypothesis_groups: [tspear_hair_bundle_deafness]`, and I agree with your reasoning for that: this edge *is* the disputed claim, not a step inside the argument, so it should not be asserted more firmly than the edges above it.

**4. The refuting datum you flagged — added, and you were right that it is the strongest one.** From the ClinGen cache this entry already cites:

> A frameshift variant that was reported in 2 probands and segregated in 2 additional family members (PMID: 22678063, 37561809) has also been reported in 2 families with ectodermal dysplasia and normal hearing (PMID: 27736875).

The founding frameshift — the allele carrying most of the DFNB98 case-level evidence — turns up in two families who have the ectodermal phenotype and hear normally. The same allele does not track with deafness, which is sharper than the non-segregation observation sitting beside it. It is now on the gene, graded REFUTE.

**Separately, from the issue's second-opinion comment**, I added PMID:41195743 (AJMG A, 2026): 11 patients from 7 families, complete penetrance for dental manifestations, "**None of the individuals had hearing loss**", concluding that "hearing loss is not a feature of the disease." That roughly doubles the published denominator, and it postdates the 2024 ClinGen re-review, so it is cited as REFUTE on the endpoint node, the phenotype, the gene and the mechanistic hypothesis. The knowledge gap's proposed experiment was updated too — half of it has now effectively been run.

**On `entry_type`, since the second-opinion bot argued for `OUT_OF_SCOPE`.** I'd rather leave that call to you than assume it, but here is my reasoning for `DISEASE`-as-contested. The concern I take seriously is "writing a plausible-sounding mechanism for a disputed disease is worse than writing nothing" — and I agree with it. What this entry does instead is make the dispute the machine-readable content: `relationship_type: DISPUTED` on the gene, the mechanism confined to an `ALTERNATIVE` hypothesis group with every edge opted in, `FAILS_TO_RECAPITULATE` on the mouse, and refuting evidence in the same sections as the supporting evidence rather than quarantined in a note. Marking the stub `OUT_OF_SCOPE` and deleting it would remove the dispute along with the entity, and would leave MONDO:0013929 for the next person to take at face value. It also stretches the `OUT_OF_SCOPE` definition, which is about concepts too abstract to carry a mechanism, not about mechanisms that may not exist.

The separate point in that comment — that ECTD14 (MONDO:0032584) is the entity dismech is actually missing, and that the repo's own G2P triage flagged it CRITICAL — is right and I have not acted on it. That is a new stub and a new curation, not something to fold into this PR.

39 → **40/40 snippets verified**; all gates re-run at exit 0.

PR #9995, opened 2026-08-29
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 97 (MONDO:0014739, MET)

### LINKED ISSUE #9979: Curate autosomal recessive nonsyndromic hearing loss 97 (MONDO:0014739)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 97 (DFNB97)** ([MONDO:0014739](https://monarchinitiative.org/MONDO:0014739)) — biallelic *MET* variants; the hepatocyte growth factor receptor, whose HGF/MET signalling is required for cochlear development, with the allelic *HGF* deafness (DFNB39) as the paired locus.

**MET is an oncogene with a large cancer literature.** Deep-research retrieval on this gene will be dominated by tumour biology that has nothing to do with DFNB97; treat gene-frequency preflight as necessary but not sufficient and check each citation is about the deafness phenotype.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97.yaml`

kevinschaper:
Adds DFNB97, the MET form of autosomal recessive nonsyndromic hearing loss.

**What the disease is — and it is not cancer biology.** HGF is expressed in the developing cochlear epithelium at the site that will become the stria vascularis, and its receptor MET is expressed both there and on the neural-crest-derived melanocytes that migrate in to form the strial intermediate cell layer. That layer is what lets the stria vascularis hold the potassium concentration of the scala media and generate the endocochlear potential — the battery that drives hair cell transduction. Inactivate MET in the developing mouse inner ear and the melanocytes never arrive; the animals are profoundly deaf as young adults. So DFNB97 is a developmental disease of a nonsensory structure, and the hair cells are downstream bystanders.

**How I handled the oncogene problem (the issue's warning).** Preflight passed on gene frequency, as predicted — but it also flagged HGF at 33% of MET mentions, the paired DFNB39 locus bleeding into the report. So every citation was checked individually for whether it is about the deafness phenotype rather than about MET. Nothing from the oncology literature is curated. Somatic amplification, exon 14 skipping and hereditary papillary renal cell carcinoma appear only in the gene `notes` and `differential_diagnoses`, as the things DFNB97 is not. The clinical counterpart is curated as a phenotype: the Pakistani family had complete blood counts, serum chemistry, urinalysis, liver function tests and fundoscopy all normal, and no family history of cancer.

**HGF/DFNB39 separation.** Different gene, different variant class (noncoding 3'UTR and splice variants, not coding missense), and very different frequency — 40 consanguineous Pakistani families map to DFNB39, against the single family that defines DFNB97 after 100 more were screened and found nothing. They converge only downstream. There is exactly one place where HGF evidence supports a DFNB97 node: the endocochlear potential, which has only ever been measured in the Hgf del10 mouse. That item is graded `PARTIAL`, said out loud in the node description, and carries its own `HUMAN_MODEL_MISMATCH` discussion. The 2016 c-MET paper is different — it inactivates ligand and receptor in parallel arms and reports the receptor result, so it is MET evidence.

**Calibration.** ClinGen's Hearing Loss GCEP rates MET–nonsyndromic hearing loss **Limited**, and the panel says why: of three reported missense variants only one scored, because one proband's phenotype was inconsistent with nonsyndromic hearing loss and the other variant was too common to be pathogenic. I typed the gene `CAUSATIVE`, not `DISPUTED` — Limited means insufficient, not contradicted, and the panel explicitly notes "no convincing evidence has emerged that contradicts the gene-disease association." The uncertainty is carried by a `KNOWLEDGE_GAP` rather than double-counted in the relationship type.

Relatedly, `functional_impact_category` is **omitted** from the genetic context. Both alleles are missense, neither has been assayed, and the only mechanistic evidence is a molecular dynamics simulation (graded `COMPUTATIONAL` and `PARTIAL`). Writing `LOSS_OF_FUNCTION` there would be inferring the exact thing the evidence base does not supply — and it is the subject of the first knowledge gap, which notes the assay is tractable because the oncology reagents already exist.

**A schema limitation I worked around and flagged.** Two findings here are examined-and-negative and both matter: preserved vestibular function, and absence of neoplasia. `Phenotype` has no `presence` slot and `FrequencyEnum` has no `EXCLUDED` value, so each is curated with the HP term for the thing that is absent and evidence graded `REFUTE`, with the description carrying the negative. This is a workaround; it is called one in `notes`. If there is a better idiom in use elsewhere I would rather adopt it.

**Deliberately not curated.** No treatments — management is generic hearing habilitation with no DFNB97-specific evidence, and a curated treatment implies a curated indication. No clinical trials, no datasets, no biochemical markers, no prevalence figure (two families).

**Validation.** `just validate` exit 0 · `just validate-terms` exit 0 · `just validate-disorders` exit 0 with **36/36 snippets verified** · `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-term-cache-integrity` all exit 0 · `validate-history` exit 0.

Closes #9979


kevinschaper:
Thanks — the critical finding is correct and I've acted on it. I checked `src/dismech/export/kgx_export.py` before changing anything: `disease_phenotype_to_edge` builds the association from `phenotype_term` and `frequency` and never reads `supports`. So "Absence of Systemic and Neoplastic Features" really would have exported as `DFNB97 biolink:has_phenotype HP:0002664 Neoplasm` at Obligate frequency — the exact MET-oncology contamination the entry was written to avoid, and you're right that dropping `frequency:` would not have helped.

Both negated phenotype entries are gone.

- The vestibular negative moved into the `Failure of Endolymphatic Potassium Homeostasis` node description, where it does more work than it did as a phenotype: preserved vestibular function is positive evidence that the lesion is confined to the stria vascularis of the *cochlear duct* rather than the inner ear generally. The "intact, or at least residual" hedge and the fact that the testing was clinical rather than VEMP/vHIT are both kept.
- The cancer negative was already quoted verbatim under the hereditary papillary renal cell carcinoma differential, so it needed no new home.
- `notes` now records what happened and why, naming the exporter behaviour, so the next curator doesn't reinvent the workaround. I checked DFNB104 and it handles its own vestibular negative the same way — thanks for the pointer.

**Treatments.** Added, following the DFNB104 pattern you named. Cochlear Implantation with `treatment_effect: BYPASSES` against `Failure of Endolymphatic Potassium Homeostasis`, and Early Identification and Habilitation. Both deliberately carry no evidence item — the rationale is this entry's inference from the strial mechanism and is labelled as such. I agree the bystander model justifies the edge cleanly here: the implant stimulates the spiral ganglion directly and so does not need the missing endocochlear potential.

**Deep-research artifact.** Committed: `research/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97-deep-research-falcon.md` plus its citations sidecar and artifacts directory. Worth noting what its preflight said, since it bears on the entry's main risk: `just preflight-dr` returned **WARN**, flagging HGF at 26 mentions against MET's 79 — 33%, the paired DFNB39 locus bleeding in. That's why every citation was checked individually rather than trusted, and why the one HGF-derived item (the endocochlear potential measurement) is graded `PARTIAL` and labelled in the node description.

The report also cites OMIM 164860 where MONDO:0014739 xrefs 616705; 164860 is the MET *gene* entry, so that flag is benign.

PR #9996, opened 2026-08-29
curate: Add Sweeney-Cox syndrome (MONDO:0060592, TWIST1 Glu117)

### LINKED ISSUE #9991: Curate Sweeney-Cox syndrome (MONDO:0060592)
Curate a dismech entry for **Sweeney-Cox syndrome** ([MONDO:0060592](https://monarchinitiative.org/MONDO:0060592)) — a *TWIST1*-related craniofacial disorder with frontonasal-dysplasia-like features: marked hypertelorism, dysplastic ears, eyelid abnormalities and ptosis, and a broad or bifid nasal tip.

The stub gives one causal gene, `hgnc:12428 TWIST1`, and no MONDO descendants. The reason this is a separate disease rather than a subtype of Saethre-Chotzen syndrome is the whole point of the entry, and it is a clean worked case for the gain/loss-of-function slot guidance in `CLAUDE.md`: Saethre-Chotzen is *TWIST1* haploinsufficiency with craniosynostosis, whereas Sweeney-Cox arises from specific missense substitutions at a conserved residue in the basic DNA-binding domain and produces a non-synostotic, frontonasal phenotype. Same gene, opposite end of the functional spectrum, different pathograph — so `functional_impact_category` on the `GeneticContext` is where the distinction has to be recorded, not in prose.

Worth checking during curation whether the reported variants are best described as dominant-negative or neomorphic, since the phenotype is not simply a more severe haploinsufficiency; if the literature does not settle it, that is a knowledge gap rather than a curator's call.

Stub: `stubs/Sweeney-Cox_Syndrome.yaml`

gingin77:
Sweeney-Cox syndrome is a very rare craniofacial disorder — three patients reported in total — caused by changing a single amino acid, Glu117, in the TWIST1 protein. Children have widely spaced eyes, missing or badly underdeveloped eyelids, and severe deficiency of the forehead and nasal bones. This PR adds it as a new disease entry.

The reason it gets its own entry rather than being folded into the existing `Saethre-Chotzen_Syndrome` entry is worth stating plainly, because both diseases are caused by TWIST1. **They are opposite kinds of damage to the same gene.** Saethre-Chotzen happens when one copy of TWIST1 stops working, so the child has half the normal amount — the remaining protein is fine, there is just less of it. Sweeney-Cox happens when the altered protein is still made and still active enough to get in the way, interfering with the working copy. About a hundred TWIST1 mutations are known in Saethre-Chotzen and not one of them hits the Glu117 position. The two diseases also look different in clinic: Saethre-Chotzen is defined by premature skull-suture fusion, which is not a typical feature of Sweeney-Cox.

In the schema that distinction is carried by `functional_impact_category: DOMINANT_NEGATIVE`, which is exactly the slot `CLAUDE.md` says should record a variant's functional consequence.

## What is in the entry

A five-node pathophysiology chain, tagged with `biological_scale`:

```
TWIST1 Glu117 Basic-Domain Substitution          (MOLECULAR)
  └─> Antimorphic Interference with TWIST1        (MOLECULAR)
        bHLH Complexes
        └─> Failure of TWIST1-Dependent           (CELLULAR)
              Frontonasal Ectomesenchyme Programme
              ├─> Frontonasal Skeletal Deficiency (TISSUE)
              └─> Periocular Mesenchyme and       (TISSUE)
                    Eyelid Morphogenesis Failure
```

The middle of that chain is the part with real mechanistic support. TWIST1 works in the cranial neural crest cells that build the face. In mouse, deleting Twist1 from those cells collapses frontonasal development and fails to switch on the genes that pattern the midface (Alx1, Alx3, Alx4). Twist1 also sits on an enhancer that drives Alx1 specifically in **periocular and frontonasal** tissue — which is why one lesion produces both the midface and the eyelid phenotype.

Also included: five phenotypes with HPO terms, the TWIST1 genetic block, two pathograph-linked animal models, one dataset, and three `discussions`.

## What is deliberately hedged

Three things in this entry are less certain than a fast read would suggest, and the entry says so rather than smoothing them over.

**No model system has both the right mutation and the right tissue.** The *C. elegans* work carries the actual disease alleles but the worm has no face. The mouse work has the right tissue but uses knockouts, which model the *Saethre-Chotzen* kind of damage, not the Sweeney-Cox kind. So the step that distinguishes these two diseases is the one step nothing has directly tested. Recorded as a `HUMAN_MODEL_MISMATCH` discussion.

**The "dimerises but cannot bind DNA" model is inferred from where the residue sits, not measured.** No biochemistry on the human Glu117 protein has been published. The node description says this outright instead of asserting the model.

**The boundary with Saethre-Chotzen may not be real.** One reported patient with a Glu117Asp change did have coronal craniosynostosis, and that paper argued this conservative substitution acts as a weaker interfering allele, closer to simple loss of the gene. If interference strength is continuous, the distinction between the two diseases is a matter of degree. Filed as a `KNOWLEDGE_GAP` rather than settled here.

A third discussion notes that the three living patients may be the survivable end of the range — the literature explicitly raises the possibility that other changes at this position are lethal.

## Evidence and classification notes

Two calls a reviewer should check:

1. **ClinGen rates this gene-disease relationship `Limited`**, and the entry records that on the `genetic` block, graded `PARTIAL`. The limitation is case count, not evidence quality — the mechanistic work is strong, there are just three patients. The entry should not present this at Definitive-level confidence, so it does not.

2. **Three evidence items are graded `evidence_source: OTHER`, not `HUMAN_CLINICAL`.** They quote a sentence describing the human phenotype that appears in the *introduction* of a mouse developmental study. It is a literature summary, not data that paper reports, so `HUMAN_CLINICAL` would attribute human evidence to a mouse paper. This is cited at all because the primary clinical description sits in full text that is not in the reference cache — which is also why the entry under-reports the additional malformations described in the index cases. That gap is recorded in the entry's notes rather than filled with uncitable content.

The one dataset, `geo:GSE230316`, was picked by hand from the mechanism literature, **not** from dataset search. Searching TWIST1 returns mostly cancer EMT studies — the Named Entity Confusion trap `CLAUDE.md` warns about. This one maps TWIST1 DNA occupancy in embryonic face mesenchyme, which is the right tissue.

Deep research was requested from `falcon`, which has no API key configured in this environment. Rather than substituting a provider by hand, the run used `--fallback`; `claude_code` produced the report and the substitution is recorded in the report frontmatter (`fell_back`, `requested_provider`, `provider_attempts`). The report's own validation flagged one unsupported quote (PMID:40020686) and two NCIT surgical-procedure terms mislabelled as phenotypes — none of the three were used here.

## Validation

`just validate-disorders` passes. Snippets 35/35 verified against the cache. `just validate-terms`, `just check-entity-refs`, `just check-duplicate-keys`, `just check-snippet-grading`, and `just verify-datasets` all pass. Compliance is 86.7%.

The stub is deleted and a history record is included.

Closes #9991


gingin77:
Both blocking items are fixed. Thanks — the phenotype-coverage one was a genuine error on my part, not just a difference of threshold.

## Phenotype coverage

You were right, and my reason for stopping at 5 was wrong. I treated "no quotable sentence in the cached abstract" as "cannot curate", but `Phenotype.evidence` is not schema-required and naming HPOA as the provenance in `notes:` is attribution rather than an invented quote. That distinction is the whole point and I missed it.

47 phenotypes added, across eleven organ systems including everything you named — HP:0005487 Prominent metopic ridge, HP:0000455 Broad nasal tip, HP:0001263 Global developmental delay, HP:0001746 Asplenia. Each carries a `notes:` line naming the HPO disease annotation for OMIM:617746 as the source, and no evidence item.

Two things I did differently from the review:

**I pulled the annotation from the ontology.jax.org API directly, not from the research report.** The report is a lead; the annotation is the authority. This turned out to matter — see below.

**No frequencies.** The API returns `frequency: None` for every term, so I did not assign any. The review cites four features at 2/2; I could not confirm that from the annotation endpoint, and the frequency-mapping guidance says to omit rather than guess. If those counts come from a source you can point me at, I will add them.

## The ablepharon question

Your doubt was correct, and checking the source settled it in a way that a straight swap would have gotten wrong.

**HP:0011224 Ablepharon is not in the OMIM:617746 annotation.** HP:0000636 Upper eyelid coloboma is. So ablepharon should not be the primary eyelid term.

But ablepharon is not an error either — it is explicitly documented in the Takenouchi Glu117Asp patient ("we document a male infant with the distinctive facial features of ablepharon..."), which post-dates the annotation. The two facts are consistent: the annotation reflects the two index subjects, the case report adds a third patient with a more severe eyelid finding.

So the entry now carries both. Upper eyelid coloboma is the primary term, from HPOA, no frequency. Ablepharon is a separate entry graded `OCCASIONAL`, with a `notes:` line stating that it is absent from the OMIM annotation and documented in one later patient. Collapsing them would either overstate the index phenotype or drop a published finding.

## Non-blocking items

| Item | Done |
|---|---|
| Treatments (corneal protection, asplenia prophylaxis) | Added, both graded `PARTIAL` |
| `c.351C>G` vs ClinVar `c.351G>T` | Recorded in `genetic` notes |
| `UBERON:0004017` periocular mesenchyme | Added to node 5 |
| Craniosynostosis name/term breadth | Renamed to Coronal Craniosynostosis |
| CNCC EMT thread (PMID:35781329) | Added as `PARTIAL` evidence |
| Dosage-continuum hypothesis group | Added, `status: EMERGING` |
| `modifier` on `GO:0046983` | **Not done** — see below |

On the treatments: both are **phenotype-directed, not disease-directed**, and say so in `notes:` and in every `explanation`. The corneal-protection source is a single neonatal eyelid-coloboma case, and the asplenia source is a general paediatric recommendation. Neither is about Sweeney-Cox. I curated them because the phenotypes they target are now curated, and left the `KNOWLEDGE_GAP` discussion open — rewritten, since "the treatments section is deliberately empty" is no longer true.

**`modifier` on `GO:0046983` I deliberately left unset.** The claim is that dimerisation is *retained*, and `ModifierEnum` is `INCREASED / DECREASED / ABNORMAL / DYSREGULATED / ABSENT / GAIN_OF_FUNCTION / LOSS_OF_FUNCTION`. There is no value meaning "unchanged", and every available one asserts something the sources do not support. The node description states the retention in prose instead. Happy to revisit if you think one of those fits.

Renaming the craniosynostosis phenotype broke `discussions[1].attaches_to` — caught by `just check-entity-refs`, fixed in the same commit. Exactly the failure mode `CLAUDE.md` warns about.

## Validation

`just validate-disorders` passes. Snippets **40/40** verified. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets` all pass. All 52 phenotype terms validate against HPO.

**Compliance drops from 86.7% to 73.5%.** That is the 47 HPOA phenotypes carrying no evidence items, and it is the trade this review asked for — coverage over evidence density. Flagging it so the number is not read as a regression.

gingin77:
All three blocking items fixed. One correction to something I asserted in my last comment, which you were right to push back on.

## Correction: the HPO API does return frequencies

I said the endpoint returns `frequency: None` for every term and used that to justify omitting frequencies. That was wrong. The fraction is nested under `metadata.frequency`; my extraction read a non-existent top-level `frequency` key and I reported the empty result as a property of the API rather than as a bug in my parsing.

Each HPOA-derived phenotype now records its actual fraction in `notes:` — 2/2, 1/2, and so on, with the citing PMID.

**I still did not assign a `FrequencyEnum` band**, and I think we agree on why: 2/2 out of two patients does not license `VERY_FREQUENT`, as you said yourself. The raw count is more honest than a band inferred from it, so the count is what the entry carries.

## Item 1 — duplicate `HP:0000636`

Fixed. Cause: my exclusion set for the bulk import listed `HP:0011224` but not `HP:0000636`, so the term arrived once by hand and once in the sweep. The Craniofacial bulk copy is deleted, the Ophthalmologic hand-curated entry kept with its term-choice reasoning.

You are right that nothing catches this — `check-duplicate-keys` inspects mapping keys, not list items. I verified by parsing the file and counting `phenotype_term.term.id`: 52 phenotypes, no repeats. The remaining second occurrence of `HP:0000636` in the file is the treatment's `target_phenotypes`, which is correct.

## Item 2 — the falsified note

Fixed, and only the trailing sentences. The ascertainment argument you wanted kept is intact; what changed is the separation of two things that note had conflated. Evidence availability is no longer a limit — the malformations are curated. Ascertainment still is: a two-patient annotation cannot say whether those patients span the phenotype of Glu117 substitution or only its survivable part.

## Item 3 — `diagnosis:`

Two entries.

**Molecular confirmation of the codon-117 variant**, citing the cached `PMID:42059179` ACMG panel-threshold sentence, graded `PARTIAL` because the paper states the standard and the Limited classification separately rather than concluding this disease is off-panel. The description makes the point you raised: reporting the gene alone is insufficient, because TWIST1 loss-of-function and TWIST1 Glu117 substitution are different diseases.

**Abdominal ultrasound for asplenia**, with `notes:` provenance and no evidence item — the move you suggested. No source recommends imaging in this syndrome specifically; the indication follows from the curated phenotype and says so.

I skipped craniofacial CT. Indexed MeSH on `PMID:30450715` says imaging was performed in that case, not that it is indicated for the disease, and I did not want a diagnostic recommendation resting on a MeSH heading.

## Suggestions

| Item | Done |
|---|---|
| Frequencies | Recovered as raw fractions in `notes:`; no enum band |
| Cheek pads | Added, `preferred_term` free text, no `term:` |
| `HP:0003577` congenital onset | `onset.onset_category: CONGENITAL` on the HPOA phenotypes |
| Asplenia category | Cardiovascular → Immunologic |

On cheek pads: `HP:0000293` Full cheeks describes generalised cheek fullness, not localised pads at the oral commissures, so binding it would assert a different phenotype. Left unbound with a note saying so.

`modifier` on `GO:0046983` stays unset, per your endorsement.

## One thing that went wrong mid-session

My first attempt at replacing the bulk phenotype block used `mechanistic_hypotheses:` as the end boundary instead of `genetic:`, which silently deleted the `genetic`, `prevalence`, `animal_models` and `datasets` sections. `just validate` still passed — those sections are not required. What caught it was the snippet count dropping 40 → 33.

Restored with `git checkout` and redone against the correct boundary, with a top-level key assertion after the edit. Nothing defective was pushed, and the commit you are reviewing has all 18 sections. Flagging it because "schema validation passes" would not have caught a section deletion, and the snippet count is what did.

## Validation

`just validate-disorders` passes. Snippets **42/42**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets` all pass. Compliance 73.8%.

gingin77:
Rebased onto `origin/main` to clear the out-of-date branch state — it was behind by two unrelated commits (the Ependymoma entry and a docs change).

Rebase was clean, no conflicts. Post-rebase check on `git diff --name-status origin/main...HEAD`: 26 files, all mine, no unrelated deletions and no stale reversions. Cache diffs are still pure additions — I checked specifically that the rebase did not revert any rows the Ependymoma entry added.

Revalidated after the rebase: `just validate-disorders` passes, snippets **42/42**, `check-entity-refs` and `check-duplicate-keys` (whole-KB, 4400 files) both clean.

No content changed. The three blocking items from the last review were already addressed in `e7ede50`; this is branch hygiene only.

gingin77:
Both blocking items fixed.

## 1. Congenital onset asserted where it cannot hold

You caught a real error, and the cause is worth naming: I propagated a single **disease-level** HPOA onset row onto **44 phenotype-level** descriptors. Those are different claims. "This disease has congenital onset" is true; "delayed speech has congenital onset" is not, because delayed speech is diagnosed by failure to meet post-natal milestones and cannot be observed at birth. The note beside the slot said the right thing while the slot said the wrong one.

`onset:` removed from `Delayed speech and language development` and `Global developmental delay`. Left in place elsewhere, on your reasoning — for anal atresia, choanal atresia, PDA, PFO, syndactyly and cerebellar hypoplasia, congenital onset holds independently of how it arrived.

## 2. ACMG snippet regraded to `OTHER`

Agreed, and the inconsistency was mine. The entry already grades three quotes `OTHER` because they are literature summary in a mouse paper's introduction. A sentence citing a professional standard, with no patients or data behind it, is the same case — CLAUDE.md names "expert consensus without data" explicitly. Regraded.

The ClinGen *classification* snippet in `genetic` stays `HUMAN_CLINICAL`, as you said — that one does aggregate human case evidence.

## Frequency policy — you were right that it was incoherent

44 phenotypes said "no band asserted, a count out of two patients does not support one" while five hand-curated ones a few lines away carried bands from the same patients. Each hand-curated frequency now records its basis: **the source's qualitative characterisation, not a patient count**, which is the mapping the curation guidance actually sanctions. The HPOA phenotypes carry no band for the complementary reason — their provenance is a bare count.

On your `OCCASIONAL` observation: it is 5–29% in HPO and both features carrying it are 1/3, which is 33%. The `Coronal Craniosynostosis` note now states that outright rather than hiding it, and explains why the prose still governs — a denominator of three cannot distinguish 33% from 5%, and the source is making a claim *about the syndrome* ("not a cardinal feature") rather than tallying its own series.

## Other suggestions

**Vaccination added** as its own treatment (`NCIT:C15346`, `therapeutic_modality: VACCINE`). You are right that the previous commit quoted a source naming vaccination first and then bound only the antibiotic half of it. Curated separately from prophylaxis rather than merged, since they are different modalities on different schedules.

**Two diagnosis entries added** — craniofacial CT with 3D reconstruction, and ophthalmic examination — both with `notes:` provenance and no evidence item, on the pattern the abdominal ultrasound entry established. The CT entry carries the differential point: craniosynostosis separates Saethre-Chotzen from Sweeney-Cox, and since one Glu117 patient had bicoronal synostosis, its presence cannot be assumed from the molecular result.

**No corneal-exposure node.** You marked it optional and noted there is no cached quotable abstract. Adding an uncited mechanism node purely to close the loop between two curated items would assert a mechanism this entry cannot support, which is the thing the rest of the entry is careful not to do.

## Also, on the GSE230316 conflict you chased

Recording that I agree with your resolution: the research report's table row is wrong and the YAML is right. The verifier cache is the tiebreaker — `"organism": "Homo sapiens; Mus musculus"`, `pubmed_ids: ["38262408"]`, 200 samples. Worth noting because it is a case where the deep-research artifact was internally confident and simply incorrect.

## Validation

`just validate-disorders` passes. Snippets **43/43**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading` all clean. 53 phenotypes, 4 diagnosis entries, 3 treatments, 18 sections.

The branch was also rebased onto `origin/main` earlier (details in the previous comment); post-rebase diff reviewed for unrelated deletions and cache reversions, none found.

gingin77:
Rebased onto current main and migrated the evidence items. **No content changed** — this is schema and branch hygiene.

## Why it needed work

The branch was 120 commits behind and `CONFLICTING`. Rebasing surfaced a breaking schema change that had landed while this PR sat approved: **[#10003](https://github.com/monarch-initiative/dismech/pull/10003) split evidence direction from directness and retired `PARTIAL`.** All nine `PARTIAL` items here failed validation against current main.

## The conflicts

Four, all the same benign shape: main added one cache row and this branch added an adjacent one in the same sorted region — `GO:0046982` vs `GO:0046983`, `HP:0005667` vs `HP:0005650`.

Resolved by **keeping both rows in sorted order**, never by picking a side. `just check-cache-order` confirms canonical ordering, and the cache diff against main is pure additions with zero deleted lines.

## The migration

Mapped each item individually rather than sweeping, because `PARTIAL` had been doing two different jobs:

- **Quote asserts something the claim follows *from*** → `SUPPORT` + `INDIRECT`. That covers the TWIST1 dimer-balance proposal, the mouse loss-of-function EMT work cited for a human antimorphic allele, the periocular reporter cited for the eyelid route, the bone-hypoplasia quote cited for a midface-retrusion term, the ACMG standard cited for a diagnostic consequence, and all three phenotype-directed treatments.
- **Quote asserts the claim, and my hedge was about strength** → `SUPPORT` + `DIRECT`, caveat kept in `explanation`. That is the ClinGen item: the sentence does assert a curated gene-disease relationship, and "Limited" describes evidence *strength*, which the new schema deliberately has no slot for.

Explanations referring to the retired grade were rewritten, so the entry no longer describes itself in vocabulary the schema has dropped.

## Note on the approval

The standing approval predates this force-push and will have been dismissed by it. Nothing in the entry's content changed, but a fresh review is needed to clear the gate.

`just validate-disorders` passes. Snippets **43/43**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-cache-order` all clean.

gingin77:
Rebased onto current `main` (33 commits of drift). No conflicts, no content change — the post-rebase diff against `origin/main` is the intended file set with the stub as the only deletion.

Re-ran the full lane on the new base: **47,206 passed**, including `tests/test_causal_targets.py`, which did not exist when this branch was cut. This entry's pathograph targets are clean and `check_causal_targets` exits 0. (The same rebase found broken targets on two sibling branches, #10005 and #10010, so the check was not a formality here.)

Snippets **43/43**. `behind_by` is now 0, which is a merge-eligibility requirement in its own right — the deterministic sweep refuses any PR whose head does not contain the current base, so this branch could not have merged while it was 33 commits behind regardless of being approved.

gingin77:

PR #9999, opened 2026-08-29
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 2B (MONDO:0012976, GJB3) as a contested entity

### LINKED ISSUE #9980: Curate autosomal dominant nonsyndromic hearing loss 2B (MONDO:0012976)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 2B (DFNA2B)** ([MONDO:0012976](https://monarchinitiative.org/MONDO:0012976)) — *GJB3* (connexin 31) variants causing high-frequency progressive sensorineural hearing loss.

`kb/disorders/Erythrokeratodermia_Variabilis.yaml` already covers the *skin* disease caused by GJB3 — that is a different entity, not coverage of DFNA2B, but read it first so the two entries agree about what the protein does. Note also that the strength of the GJB3–deafness association has been questioned; if the evidence is weaker than the OMIM entry implies, say so in the entry rather than curating around it.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_2B.yaml`

kevinschaper:
Adds DFNA2B, the GJB3 form of autosomal dominant nonsyndromic hearing loss.

The curation issue asked whether the GJB3–deafness association is weaker than OMIM implies, and to say so in the entry rather than curate around it. It is, and the entry does.

**The evidence against.** Three arguments of different kinds:

- **ClinGen.** The Hearing Loss Gene Curation Expert Panel has classified GJB3–nonsyndromic hearing loss **Disputed** since 2018, with the wording "there is convincing evidence disputing the association". That is stronger than "insufficient evidence" — it asserts evidence against, not absence of evidence for. Its summary also explains why the variant count overstates the case: many reported variants are common in population databases or have no evidence of pathogenicity, and only three scored.
- **A case-control test of the main allele.** c.538C>T (p.Arg180*) is the GJB3 allele routinely screened on Chinese deafness panels. In 5,700 hearing-loss patients it was found in 0.40%; in 4,600 normal-hearing controls, 0.24%; not significantly different. Eleven of the 23 carriers also had GJB2 or SLC26A4 mutations, and **all 23 were sporadic with no family history** — which is not what a dominant disease looks like.
- **The mouse.** The Cx31-null mouse has no morphological or functional defect of the inner ear.

**What is real** is the protein biology: connexin 31 is expressed in the inner ear, and hearing-loss-associated Cx31 mutants are retained in the ER, induce BiP, and fail to transfer dye. Two variants fail to couple even alongside wild-type Cx31 — dominant-negative behaviour, which is what a dominant disease needs. What is missing is any demonstration that this causes deafness in a person or an animal.

**How that is encoded, not just narrated.** `relationship_type: DISPUTED` on the gene; the mechanism in a `mechanistic_hypotheses` group with `status: ALTERNATIVE` that every causal edge opts into; the founding claim graded `PARTIAL` because the founding paper itself only says "may be responsible" (the hedge has drifted out of secondary sources); the case-control study cited as `REFUTE` in four separate sections; the mouse as `FAILS_TO_RECAPITULATE` at `fidelity: LOW`.

**I also recorded the strongest argument against my own conclusion.** The single sentence reporting normal hearing in the Gjb3-null mouse reports normal *skin* in the same breath — and GJB3–erythrokeratodermia variabilis is rated **Definitive** by the same expert panel. An animal that cannot report an established GJB3 disease is a weak instrument for excluding a disputed one. That is written into the model's `limitations` and given its own `HUMAN_MODEL_MISMATCH` discussion rather than left out because it cuts the other way.

**Keeping faith with the existing EKV entry.** I read `kb/disorders/Erythrokeratodermia_Variabilis.yaml` first, as the issue asked. The two entries agree about the protein — mutant Cx31 fails to reach the membrane and cannot form gap junctions — and the DFNA2B pathophysiology says so explicitly. They diverge only on whether the tissue-level consequence is established, and only the skin one is. I cited **both** ClinGen assertions as `external_assertions` (Disputed for hearing loss, Definitive for EKV) so the contrast lives in the data rather than only in prose. Nothing dermatological is curated as a DFNA2B phenotype.

The genuinely awkward case is a Chinese pedigree where one heterozygous GJB3 allele gave EKV in the daughter, ichthyosis in the mother and nonsyndromic hearing loss in the son. It is under differential diagnoses, graded `PARTIAL`, because it is compatible with both readings and settles neither.

**A defective upstream source.** The ClinGen assertion's evidence summary lists PMID 12759707 among the publications in which the association was seen. Looked up at PubMed, that is a 2003 Spanish-language editorial on smoking prevention in *Anales del Sistema Sanitario de Navarra*. It is noted on the external assertion, cited nowhere, and I made no guess at which publication was intended.

**Deliberately not curated.** No treatments — management would be generic hearing habilitation with no DFNA2B-specific evidence, and curating one would give the entity a clinical solidity the evidence does not support. No progression section: "progressive, postlingual, fourth-decade onset" is repeated in reviews, but I found no primary natural-history data for it, and asserting a clinical course for a disputed entity on review prose seemed the wrong kind of completeness. No trials, no datasets.

**Validation.** `just validate` exit 0 · `just validate-terms` exit 0 · `just validate-disorders` exit 0 with **38/38 snippets verified** · `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-term-cache-integrity`, `check-source-defect-claims` all exit 0 · `validate-history` exit 0.

Closes #9980


kevinschaper:
Both blocking findings were correct and are fixed, along with all four suggestions. Finding 1 in particular was a real error on my part and I'm glad it was caught.

**1. The Disputed gloss — you're right, and the correction is load-bearing.** I checked the cache: `More evidence is needed to either support or refute the role GJB3 plays in this disease.` is indeed the very next sentence, and my reading ("asserts evidence against, not merely absence of evidence for") is contradicted by it. PMID:30894701 settles it: "the Disputed classification indicates that the expert panel reviewed the evidence and disputed the claim due to insufficient **or** contradictory evidence."

What changed:
- The "More evidence is needed…" sentence is now an evidence item on the external assertion, graded `PARTIAL`.
- PMID:30894701 is **mined rather than merely listed** — two evidence items, the definition of Disputed and the 82/12/25/32/10/3 outcome distribution that locates it on the scale. Thanks for catching that it was cached, listed, and uncited.
- The gloss is corrected in all three places it appeared: the external assertion `explanation`, the entry `description`, and the `genetic` notes. The description now quotes both halves and says explicitly that Disputed "is not a finding that the gene is innocent."
- The structured encodings are unchanged, per your note — `DISPUTED`, `ALTERNATIVE`, and `PARTIAL` on the founding claim were all right.

I also replaced the framing "the strongest wording short of Refuted" with something the source supports: one of only ten Disputed calls among 164 pairs. That is a real signal without overclaiming what the word means.

**2. DFNA2A/KCNQ4 — added, and it should have been there from the start.** For an entry whose whole practical point is "don't report this variant as the cause", the established gene at the same locus with the same audiogram is the first alternative to name. Added as the *first* differential, with the GeneReviews DFNA2 chapter (PMID:20301388) and the AudioGene audioprofile paper (PMID:18941426), both fetched with `just fetch-reference` — no invented citations. "Was KCNQ4 sequenced" is now item one on the `diagnosis` checklist, ahead of the GJB2/SLC26A4 check. I also pointed at `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_2A.yaml` for whoever curates it.

You're right that this suggests the sweep had a gap rather than a boundary.

**3. Deep-research artifact — committed** (`research/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_2B-deep-research-falcon.md` + sidecar + artifacts). Preflight was a clean **PASS**. On your related question: the report does describe the DFNA2 review as reporting "fourth-decade, progressive, sloping hearing loss" but sources it to a review rather than to primary natural-history data, and explicitly says no disease-specific natural-history cohort was retrieved. So the `progression` omission stands, and it now has a second independent basis.

**4. GeneReviews — done your way.** Not tagged; a `notes` paragraph records that PMID:20301607 was checked, is abstract-only, and contains zero occurrences of GJB3, DFNA2 or connexin, so tagging it would produce exactly the tagged-but-not-mined state. It also records that the DFNA2 GeneReviews chapter does exist but is about KCNQ4, and is cited in the differential for that reason. The negative is now auditable.

**5. PMID 12759707 — cache committed, per your first option.** `references_cache/PMID_12759707.md` ships with this PR so `check-source-defect-claims` and any reader can adjudicate the claim against the repo rather than take it on trust. It is a 2003 Spanish-language editorial on smoking prevention in *Anales del Sistema Sanitario de Navarra*. It is cited in no evidence item, and I made no guess at the intended publication. `check-source-defect-claims` exits 0 with the cache present.

**6. `measure_type`** → `CARRIER_FREQUENCY`. You're right, the notes describe a carrier rate in a deaf cohort.

**7. Background sentence** regraded `HUMAN_CLINICAL` → `OTHER`, with the reason in the explanation. The same paper's coupling result stays `IN_VITRO`.

**8.** "GJB3 lies at 1p33-p35" now reads as the 1998 mapping interval against the modern 1p34.3; the uncited "stronger than the Limited classification the same panel gives MET" comparative is dropped — and you're right that it was ambiguous in the wrong direction, since Limited sits *above* Disputed.

38 → **47/47 snippets verified**; all gates re-run at exit 0, including `check-reference-titles` and `check-source-defect-claims`.

PR #10001, opened 2026-08-29
curate: Add Seizures-scoliosis-macrocephaly syndrome (MONDO:0014731, biallelic EXT2)

### LINKED ISSUE #9983: Curate seizures-scoliosis-macrocephaly syndrome (MONDO:0014731)
Curate a dismech entry for **seizures-scoliosis-macrocephaly syndrome** ([MONDO:0014731](https://monarchinitiative.org/MONDO:0014731)) — biallelic *EXT2* variants; a recessive neurodevelopmental phenotype distinct from the dominant exostoses that EXT2 haploinsufficiency causes.

**This is a dose/direction trap of the kind that has bitten this project before.** EXT2 is far better known for autosomal dominant hereditary multiple exostoses; this entity is the biallelic one, sometimes called autosomal recessive omodysplasia-adjacent or the "seizures, scoliosis and macrocephaly" phenotype. Any citation about exostoses in a heterozygous carrier belongs to the other disease. `kb/` mentions EXT2 only in `Chondrosarcoma.yaml` and `Membranous_Nephropathy.yaml`, neither of which is coverage.

Stub: `stubs/Seizures-scoliosis-macrocephaly_Syndrome.yaml`

kevinschaper:
Adds `kb/disorders/Seizures-scoliosis-macrocephaly_Syndrome.yaml` and deletes the stub it curates.

## The dose/direction trap, which is the whole point of this entry

`entry_type: DISEASE`. EXT2 causes two diseases by two different genetic mechanisms:

| | Hereditary multiple exostoses type 2 | This entry |
|---|---|---|
| Dose | heterozygous | biallelic |
| Allele type | loss of function / null | hypomorphic missense |
| Phenotype | osteochondromas | seizures, intellectual disability, hypotonia, scoliosis, abnormal head size, renal dysfunction |

Nearly all EXT2 literature is about the first one. Every citation here was checked for which arm it describes.

**Deliberately not curated, because they belong to the dominant arm:** the zebrafish *dackel*/ext2 cartilage work (studied expressly to explain the aetiology of hereditary multiple exostoses), the osteochondroma-to-chondrosarcoma progression literature, and the EXT-carrier metabolic studies. All were read and set aside; the reasons are in the entry `notes`.

**One dominant-arm paper is used on purpose, in two marked places.** PMID:16236767, the Ext2 knockout mouse, supplies two facts. Its homozygous nulls arrest at embryonic day 6.0 and never gastrulate — which is evidence about *this* disorder, because it explains why every reported human allele is missense rather than null. And its heterozygotes get exostoses — which is evidence about the *other* disorder, and is quoted inside the model's `limitations` field for the express purpose of labelling it as such.

## Not everything is clean, and the entry says so

The third reported recessive family had biallelic EXT2 missense variants, the neurodevelopmental phenotype, **and** multiple exostoses. That is a within-disorder observation, not a reclassification: their authors read it as a possible genotype-phenotype correlation inside the recessive disorder. Exostoses are therefore curated as a variable phenotype with both the positive and the negative report attached, and a `KNOWLEDGE_GAP` asks what sets the threshold. The practical consequence is stated: absence of exostoses does not rule this disorder in, and their presence does not rule it out.

## Preflight WARN, investigated rather than waved through

`just preflight-dr` returned WARN on two counts and both were checked:

- **OMIM mismatch** (report cites 608210, MONDO xrefs 616682): 608210 is the *EXT2 gene* MIM number, not a rival phenotype. Not a problem.
- **NDST1 at 27% of EXT2 mentions**: one reported patient (PMID:30997052) carries a monoallelic NDST1 variant of uncertain significance alongside the biallelic EXT2 variants. There is no second NDST1 hit, deletion/duplication analysis was negative, and the same NDST1 variant is present in the patient's mildly affected twin sister. It is recorded in `genetic.notes` and as a differential diagnosis, not folded into this disorder's mechanism.

## Frequency and direction discipline

- **Head size is two phenotypes, not one.** `Macrocephaly` and `Microcephaly` are curated separately, each with the family that reported it. Different families went in opposite directions and the syndrome name asserts only one of them; averaging them would produce a machine-readable value that contradicts half the sources.
- **Scoliosis, hypotonia and osteopenia** each carry a `PARTIAL`-graded counter-citation from a patient who lacked them, so the prose "not invariant" and the evidence agree.

## The biochemical finding worth knowing

Heparan sulfate was **low** in dried blood spot, serum and urine in a patient investigated for coarse facies. The assay was ordered to exclude a mucopolysaccharidosis, where the same analyte would be *high* because degradation fails; here synthesis fails. A low result on that assay is a positive finding, not a negative one, and the entry says so in both `biochemical` and `differential_diagnoses`.

No `reference_ranges` block is curated. The source states its intervals are in Table 1, and that table is not present in the cached text — so the block was left out rather than reconstructed.

## Deliberately not curated

- No `clinical_trials`, `datasets`, or `environmental`.
- No Orphanet citation. ORPHA:466926 is the xref on MONDO:0014731 but no `references_cache/ORPHA_466926.md` exists, and refreshing the Orphadata bulk XML for one record was not worth the download.
- PMID:35045016 (Sabir 2022) is fetched but not cited: its cache is `content_type: unavailable` with no body, so no snippet can be verified against it.

## Validation

All exit 0: `just validate` (48/48 snippets verified), `just validate-terms`, `just validate-disorders`, `just validate-history`, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

Closes #9983

### DIFF
--- a/kb/disorders/Seizures-scoliosis-macrocephaly_Syndrome.yaml
+++ b/kb/disorders/Seizures-scoliosis-macrocephaly_Syndrome.yaml
@@ -898,10 +898,10 @@ treatments:
     term:
       id: HP:0011968
       label: Feeding difficulties
-  - preferred_term: Global developmental delay
+  - preferred_term: Intellectual disability
     term:
-      id: HP:0001263
-      label: Global developmental delay
+      id: HP:0001249
+      label: Intellectual disability
   evidence:
   - reference: PMID:26246518
     reference_title: "Old gene, new phenotype: mutations in heparan sulfate synthesis enzyme, EXT2 leads to seizure and developmental disorder, no exostoses."


kevinschaper:
Both blocking items fixed, plus three of the four optional suggestions. One decline, with the reason.

## Blocking

**1. The missing multiorgan phenotypes.** Fair hit — the entry argued this is a multiorgan congenital disorder of glycosylation while curating only the neurological and skeletal arms. Five added, all from sources already in the cache:

- Ventricular septal defect `HP:0001629`, 2/4
- Autism `HP:0000717`
- Gastroesophageal reflux `HP:0002020`
- Constipation `HP:0002019`
- Feeding difficulties `HP:0011968`

**2. The introduction sentence regrade.** Correct, and it was inconsistent with the convention the entry states in its own `notes`. Now `OTHER`. You are also right about why the automated check cannot catch it: `check-snippet-grading` keys on the quoted sentence, so a paper with one sentence graded three different ways passes as long as each sentence is consistent with itself.

## Suggestions taken

**3. The Gupta allele pair.** Added — `p.Val373Asp` / `p.Thr672Met`, with the sentence that reports them. You were right that the argument gets stronger with the fourth pair in it: four independent families, four missense pairs, no biallelic null anywhere, and an Ext2-null mouse that dies at gastrulation. That is now a pattern rather than an assertion.

**4. Node splitting.** Split. `Axial Skeletal Involvement` carries scoliosis and osteopenia; `Craniofacial Dysmorphism and Abnormal Head Size` carries the facies and head size. The exostoses `KNOWLEDGE_GAP` now attaches to the axial node instead of the whole systemic-deficit node, which was your point. The split also gives the head-size problem somewhere honest to live: the craniofacial node's description says outright that it does not assert a direction, because the defining family had macrocephaly and a later family had microcephaly.

**6. The inheritance quote.** Replaced with the paper's conclusion — "we have unravelled the genetic basis of a new recessive disorder" — and the methods sentence kept as the second item, since it does document the segregation analysis behind that conclusion.

## Suggestion 5, declined — and it turned up a real defect

You are right that `therapeutic_modality: OTHER` was a tell, but not of bundling. I checked all four cached sources for a documented antiseizure regimen and there is none: no drug name, no drug class, nothing. The treatment node was named "Antiseizure medication and developmental support" for an intervention no reported patient is described as receiving. Splitting it would have produced a pharmacotherapy node with no citable indication.

So instead of splitting, I narrowed it to what is supported. The node is now `Symptom-directed supportive management` with `therapeutic_modality: BEHAVIORAL`, and the description says explicitly that antiseizure medication is the obvious mainstay but that no reported patient's regimen appears in the cited literature, so none is curated. The `Seizure` target phenotype is dropped for the same reason and replaced with the ones the node actually addresses.

The one therapeutic direction anyone has proposed — enzyme replacement or administration of heparan sulfate — is now quoted, graded `PARTIAL`, and described as a suggestion in a discussion section rather than an intervention that exists. That also corrected an over-claim of mine: the description used to say "no disease-modifying treatment exists **and none is in prospect**", which that sentence contradicts.

## Suggestion 7, declined for the reason you gave

I could not source the claim that Gupta's clinically unaffected parents had low heparan sulfate either. It is not in the cached full text, and the Table 1 the paper points to for reference intervals is not reproduced in the cache. Without it I would be adding a specificity caveat sourced to a deep-research summary, which is exactly the thing deep research is not for. The `biochemical` `notes` already records that the intervals are unavailable in the cache, so a reader can see the limit of what is established.

All exit 0: `just validate` (59/59 snippets verified, up from 48), `just validate-terms`, `just validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

Issue #9986, opened 2026-08-29
hypothesis_id on assessment sidecars is required but never validated: one committed sidecar already points at a hypothesis that does not exist


gingin77:
`hypothesis_id` is a required field on every hypothesis-assessment sidecar, and nothing checks it. One committed sidecar already points at a hypothesis that does not exist, and it passes validation cleanly.

The drifted one is in `3-Hydroxy-3-Methylglutaryl-CoA_Synthase_Deficiency`:

| Where | Value |
|---|---|
| `assessments/openscientist-assessment-by-codex.yaml` | `canonical_hmgcs2_ketogenesis_failure_model` |
| `kb/disorders/3-Hydroxy-3-Methylglutaryl-CoA_Synthase_Deficiency.yaml:188` | `canonical_human_hmgcs2_ketogenesis_failure` |

The two are plainly meant to be the same hypothesis — the sidecar drops `human_` and appends `_model`. Because nothing resolves the field, the mismatch is invisible.

## Why the field matters

In normal use it is a join key pointing at three things at once: the directory the sidecar sits in, a `hypothesis_group_id` in the disease entry, and the causal edges tagged with that group via `downstream[].hypothesis_groups`. It is what connects a provider's research report back to the specific competing explanatory model in curated content.

Across the 69 committed sidecars, all 69 match their own directory name, and 68 resolve to a hypothesis group in `kb/`. The exception is the one above.

## Why nothing catches it

`src/dismech/hypothesis_assessment.py` never reads the field — zero occurrences. `iter_assessment_problems()` checks the filename pattern, that `provider` and `assessor` match the filename, that the file sits in an `assessments/` directory, that `source_report` and every `artifacts` entry exist on disk, and that each `report_quote` is a verbatim substring of the report. LinkML then checks only that `hypothesis_id` is present and a string. So a value that resolves to nothing satisfies every gate.

## A second case the field does not fit

Assessing a provider report that is not about a disease hypothesis has no valid value to offer. The report in question explores how AOP-Wiki represents Key Events across eight pathways — a methodology question with no disease hypothesis behind it, hence no directory and no hypothesis group to name. The field still has to be filled, so it currently reads `na_aop_methodology_exploration`, chosen so it cannot be mistaken for a resolving key.

This matters because the rest of the assessment machinery works well for such reports. The claim dispositions, verbatim report quotes and citation records all apply unchanged; `hypothesis_id` is the only part that assumes a disease hypothesis.

[...]

Issue #9985, opened 2026-08-29
Link Lead_Poisoning's heme and red-cell mechanisms to the modules that already describe them


gingin77:
Lead poisoning damages heme synthesis and red cells, and dismech already has generic modules describing both processes — around ten other disease entries point at them. `Lead_Poisoning` points at neither, so it sits outside the consistency checking those modules exist to provide. Most of the fix is a one-line declaration per node, but it exposes one genuinely missing mechanism step.

Three nodes in `kb/disorders/Lead_Poisoning.yaml` have a counterpart already written:

| Node in Lead_Poisoning | Module node it instantiates |
|---|---|
| Inhibition of delta-aminolevulinic acid dehydratase | `heme_biosynthesis_porphyria#Heme Biosynthesis Enzymatic Block` |
| Erythrocyte pyrimidine 5'-nucleotidase deficiency | `hemolytic_anemia_erythrocyte_destruction#Reduced Erythrocyte Integrity` |
| Hemolytic erythrocyte injury | `hemolytic_anemia_erythrocyte_destruction#Oxidative and Membrane Injury` *or* `#Premature Erythrocyte Destruction` |

The first is the clearest case. The ALAD node already carries `heme biosynthetic process` on erythrocytes; the module node carries the same process plus porphyrin metabolism. `Acute_Intermittent_Porphyria` declares conformance to that exact node at line 344 of its file. The two red-cell mappings need a curator to pick the right target — the third node is named for injury, which could match either module node, and that choice should be made against the biology rather than the name.

**The missing step this exposes.** The heme module runs `Heme Biosynthesis Enzymatic Block → Accumulation of Porphyrin Precursors and Porphyrins`. `Lead_Poisoning` goes straight from ALAD inhibition to `Anemia`, with no accumulation node in between — even though the entry already tracks zinc protoporphyrin as a biomarker, and elevated ALA and ZPP are the standard laboratory signature of lead exposure. The accumulation step is real in this disease and simply unmodelled, so declaring conformance will correctly flag it. Adding that node with cited evidence is the substantive part of this task.

[...]

Issue #9976, opened 2026-08-29
Add a mechanism module for reduced glutamatergic signalling (developmental NMDA receptor blockade)


gingin77:
dismech has a module for glutamate *over*activation but nothing for the opposite. `glutamate_excitotoxicity` models excessive receptor stimulation and calcium overload, which is the right model for stroke, ALS and epilepsy. There is no module for glutamate signalling that is *suppressed* during development — which is how lead causes cognitive impairment — so a disease with that mechanism currently has no module to point at.

This came out of checking whether existing dismech modules already cover the Key Events in the eight AOP-Wiki pathways that list lead as a prototypical stressor (#8309).

Three Key Events in AOP 12 fail to map, all for the same reason:

| Key Event | Title |
|---|---|
| KE201 | Binding of antagonist, NMDA receptors *(molecular initiating event)* |
| KE195 | Inhibition, NMDARs |
| KE52 | Decreased, Calcium influx |

`glutamate_excitotoxicity` is built end to end on the opposite sign. Its nodes are "Excessive Glutamatergic Stimulation and Impaired Glutamate Clearance" and "Glutamate Receptor Overactivation and Calcium Overload". A pathophysiology node cannot declare `conforms_to` against a node whose name asserts overload when the change being modelled is a decrease — even though `Descriptor.modifier` could carry the direction perfectly well.

The chain downstream is well characterised and recurs beyond lead: reduced NMDA-receptor calcium entry leads to reduced BDNF, impaired synaptic plasticity, and a learning and memory deficit. That recurrence is what makes it module-shaped rather than a detail of one entry.

KE381 (Reduced levels of BDNF) is also unmatched today — no module in the KB mentions BDNF — so the same module would close both gaps.

Related: #3974 covers a different `conforms_to` design question and does not overlap with this.

gingin77:
The same analysis turned up a second gap of this kind, on the cardiac side.

Five of the eight lead AOPs are cardiac, and their molecular initiating events are all chemical: L-type calcium channel blockade, ERG potassium channel inhibition, reduced Na/K-ATPase activity, phosphodiesterase inhibition, funny-current inhibition. None of the five maps to a module node. The two modules that do cover the downstream cardiac physiology — `cardiac_ion_channel_repolarization` and `cardiomyopathy_maladaptive_remodeling` — both begin with a mutation, so a chemical doing the same damage has nothing to attach to.

[...]

Issue #8309, opened 2026-08-10
Explore how dismech's environmental-exposure modeling relates to the Adverse Outcome Pathway (AOP) framework


gingin77:
## Premise

Dismech models disease pathophysiology as a causal graph, and its environmental-exposure layer (ECTO `exposure_term`, `influences_mechanisms`, `environmental_effect`) captures an exposure → mechanism → outcome chain. The **Adverse Outcome Pathway (AOP)** framework used in chemical hazard/risk assessment organizes toxicological knowledge as **Molecular Initiating Event (MIE) → Key Events (KE) → Key Event Relationships (KER) → Adverse Outcome (AO)** (review: [PMC5906804](https://pmc.ncbi.nlm.nih.gov/articles/PMC5906804/); ~~chemical-RA application: [PMID:26354708](https://pubmed.ncbi.nlm.nih.gov/26354708/)~~; the OECD AOP program / AOP-Wiki).

These share obvious DNA — both are directed causal graphs spanning biological scales, both attach evidence to relationships, both aspire to reuse. **They are not the same thing**, and the point of this issue is to map where they align and where they diverge, then decide what (if anything) dismech should adopt — not to retrofit dismech into AOP or vice versa.

## Candidate correspondences (to be tested, not assumed)

| AOP element | Closest dismech construct | Fit? |
|---|---|---|
| Stressor / chemical | `environmental:` entry + ECTO `exposure_term` | partial — see divergences |
| Molecular Initiating Event | node targeted by a `TRIGGERS` `influences_mechanisms` link | loose — dismech targets aren't required to be molecular or measurable |
| Key Event | `pathophysiology` node (`biological_scale` tag) | partial — KEs have stricter measurability/essentiality expectations |
| Key Event Relationship | `downstream` causal edge (`causal_link_type`, `hypothesis_groups`) | partial — KERs carry a formal WoE rubric dismech doesn't encode |
| Adverse Outcome | terminal phenotype (HP) node | partial — AO is pitched at a risk-assessment-relevant level of organization |
| Reusable, stressor-agnostic pathway | mechanism module (`kb/modules/`) + `conforms_to` | worth examining — closest structural analogue |

[...]

PR #10002, opened 2026-08-29
AOP/EMOD framework alignment: the project page and the lead use case (#8309)

gingin77:
This adds the AOP/EMOD framework alignment project page, which assesses how far dismech's mechanism and evidence modelling lines up with the Adverse Outcome Pathway framework. It records an inventory of comparisons between the two; it does not decide whether dismech should adopt anything from AOP-Wiki, which is a separate call.

The pilot comparator is lead poisoning: the eight AOP-Wiki pathways that list lead as a prototypical stressor, held against `Lead_Poisoning.yaml` and against dismech's mechanism modules.

## What is here

The project page (`projects/AOP_EMOD_ALIGNMENT.md`) covers the AOP-Wiki v2.8 data model and the EMOD expansions, where the two frameworks already line up, where they diverge, and the lead use case. Its working notes and the source material behind it sit in `projects/AOP_EMOD_ALIGNMENT/`, which does not render.

Three findings worth pulling out.

**A weight-of-evidence grade is not evidence.** Every AOP-Wiki relationship in the lead set carries a High, Moderate or Low grade with none left unspecified, which by the usual measure looks exemplary. But 20 of the 36 relationships cite nothing at all — no references, no empirical support, no plausibility text — and 18 of those are graded High. The empty ones are exactly the cardiac relationships; the neurodevelopmental ones each carry between roughly 1,500 and 10,600 characters of citations. The grades run backwards to the documentation, and the one OECD-endorsed pathway of the eight carries the only Low grades. The grade and the citations are stored in different places and nothing couples them, so grade coverage has to be checked relationship by relationship.

**dismech modules are its answer to consensus nodes, and they cover about half the ground.** Of the 35 Key Events across the eight pathways, 14 map cleanly onto existing module nodes and 7 partially. The misses are informative: the two modules covering lead's cardiac physiology were written for inherited channelopathy and cardiomyopathy, so they begin with a mutation where an AOP begins with the damage itself, and none of the five chemical initiating events has anywhere to attach. On the neurodevelopmental side, `glutamate_excitotoxicity` is built entirely on overactivation while lead inhibits NMDA receptors, so the sign is inverted at every node.

**The OpenScientist report this work was waiting on does not underpin any claim here.** Its extraction from AOP-Wiki verified exact against the 2026-08-06 XML export, and its graph analysis reproduces. Its central biological claim — that ionic mimicry of calcium unifies lead's mechanisms — is contradicted by the three papers it cites for it, which describe the binding as opportunistic and allosteric and specifically not ionic displacement. The report is committed alongside a claim-by-claim assessment recording what survived, and it is cited on the page only as the subject of that assessment.

## Notes for review

The assessment sidecar sits under `projects/` rather than `kb/hypotheses/` on purpose: the report explores a methodology question about AOP-Wiki, not a disease hypothesis. Its `hypothesis_id` is therefore a non-resolving placeholder, which is the subject of #9986.

`artifacts/*.json` are OpenScientist's own figure-generation code, kept verbatim. The node and edge literals inside them are the dataset that was verified against the XML export. The rendered PNGs were dropped because nothing cited them.

Two links in the page point at `AOP_EMOD_ALIGNMENT/draft-sections-1-6.md`, which is deliberately uncommitted working material. Both sit inside HTML comments, so they never render — but following them from the raw markdown will find nothing.

`references_cache/PMID_25466379.md` is cached but not yet cited anywhere.

## Issues opened from this work

- #9976 — no mechanism module for reduced glutamatergic signalling
- #9985 — link `Lead_Poisoning` to the heme and hemolysis modules that already describe its mechanisms
- #9986 — `hypothesis_id` on assessment sidecars is required but never validated

Closes nothing; #8309 stays open, since this answers the "to what extent" question rather than closing out integration.

### DIFF
--- a/projects/AOP_EMOD_ALIGNMENT.md
+++ b/projects/AOP_EMOD_ALIGNMENT.md
@@ -10,6 +10,16 @@ description: >-
 tags: [FRAMEWORK_ALIGNMENT, EVIDENCE, EXTERNAL_COLLABORATION, ENVIRONMENTAL_EXPOSURE, SCHEMA_EVOLUTION]
 diseases:
   - Lead_Poisoning
+modules:
+  - cardiac_ion_channel_repolarization
+  - cardiomyopathy_maladaptive_remodeling
+  - glutamate_excitotoxicity
+  - synaptic_vesicle_cycle
+  - mitochondrial_dysfunction
+  - excitatory_synapse_scaffold_disruption
+  - drug_induced_liver_injury
+  - drug_induced_nephrotoxicity
+  - diabetic_vascular_complications
 ---
 
 # AOP EMOD Framework Alignment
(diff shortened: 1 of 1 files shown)


gingin77:
All four IMPORTANT items fixed in 1f98022b70, plus two of the suggestions. Each was independently verified before changing anything.

**Item 1 — the genetic-cause claim.** Correct, and the reviewer identified the load-bearing part precisely. `Primary Cardiomyocyte Insult` reads "in acquired and secondary cardiomyopathies it is a hemodynamic (pressure or volume overload), metabolic, toxic, or inflammatory stress" — a chemical insult is named in the node itself. The claim is now scoped to `cardiac_ion_channel_repolarization` alone, and the page says explicitly that it is not true of the other cardiac module. As noted, all five unmatched MIEs are electrophysiologic and sit in that module's territory, so no individual fit call changes.

**Item 2 — the census.** Removed rather than restated, for two reasons. The 127 figure came from a checkout that was 399 commits behind main; today there are 164. More importantly the 38/89 split came from a keyword scan of first-node text, and that method cannot separate a module that *mentions* a genetic cause from one that *requires* it — running it now flags `cardiomyopathy_maladaptive_remodeling` as genetic, which is the very module item 1 shows is not. A corrected count would have been no more trustworthy than the stale one. The claim now rests on three verified examples instead: `drug_induced_nephrotoxicity` opening at "Nephrotoxic Drug Exposure and Tubular Uptake", `drug_induced_liver_injury` at "Reactive Drug Metabolite Formation", `diabetic_vascular_complications` at "Chronic Hyperglycemia".

**Item 3 — the arithmetic.** The table is 10 clean, 3 partial, 2 endpoint-only, 6 unmatched across 21 Key Events. The sentence said "six that do not" while the three groups below it describe eight. Now reads eight.

**Item 4 — frontmatter.** `modules:` added with the nine slugs the page analyses; all nine resolve to files in `kb/modules/`.

**Suggestions applied:** the neurodevelopmental lead-in named three modules while its table cited a node from `excitatory_synapse_scaffold_disruption`, now listed; and the truncated quote is completed to the full node name, "Excessive Glutamatergic Stimulation and Impaired Glutamate Clearance".

**Not changed.** `references_cache/PMID_25466379.md` is fetched but uncited — left in place, since removing a cached reference is a separate call from the prose fixes here, and it is already flagged in the PR body. The remaining suggestions concern the OpenScientist report's own content and the assessment's location outside the `tests/test_data.py` glob; the latter is what #9986 is about.

PR #7612, opened 2026-08-01
Add EHR-derived phenotype profile schema (ProfileSet) with SEPIO evidence layer

mellybelly:
> **Body kept current as of `86e01d7`.** If you are here to answer one of the open design questions, read **[The open question](#the-open-question)** first; it is the live one. Nothing under `kb/` changes; this is additive.

## Summary

Adds a standalone LinkML schema for **EHR-derived phenotype profiles**: what co-occurs in a disease cohort, exported from a model fitted to structured records, reduced to what a disease page can show.

```
ProfileSet ──has──> Profile ──has──> CodeDistribution ──has──> WeightedCode
     │                                  (per domain,            (code + label
  disease                                with its                + weight)
  (MONDO)                                vocabulary)
profile_source
 (provenance)
```

**14 classes / 70 slots / 11 enums, one tree root.** A set is one disease and one fitted model; a profile under it is one recurring pattern.

## The open question

**@cmungall** reviewed and said the schema *"feels bolted on rather than cohering with what exists"* and that it *"seems to duplicate parts"* of the comorbidities model. I audited that rather than arguing, and he is right — more so than the comment claims:

| This schema | Already in `dismech.yaml` |
|---|---|
| `Term` — slots `term_id`, `term_label` | **`Term` — slots `id`, `label`** |
| `ProfileMethodEnum` | strict subset of `AssociationSignalMethodEnum`, same spellings |
| `EvidenceDirectionEnum` | `EvidenceItemSupportEnum` — identical permissible values |
| `DataItem.item_value` / `.reported_in.id` | `EvidenceItem.snippet` / `.reference` |
| `EvidenceLine.direction_of_evidence_provided` / `.description` | `EvidenceItem.supports` / `.explanation` |
| `ProfileSource.resource` / `.method` / `.cohort_size` | `AssociationSignal.source` / `.method` / `.population` |

Same class name, different slot names, same job — and **6 of 14 classes** (`EvidenceLine`, `DataItem`, `Document`, `Proposition`, `Activity`, `Agent`) re-spell what `EvidenceItem` does in 7 slots. The tell was here the whole time: this PR contains a test asserting `EvidenceDirectionEnum` and `EvidenceItemSupportEnum` never drift apart. **A drift test between two enums is an argument that they should be one enum.**

Three options, recommendation first. **This is the decision the PR is parked on:**

[...]

Issue #7450, opened 2026-07-31
skip_prefixes hides DOI: ~6% of KB snippets are never substring-checked


caufieldjh:
## Summary

`conf/reference_validator_config.yaml` lists **`DOI`** in `skip_prefixes`, which is a
blanket instruction to never substring-check any snippet cited to a DOI. Across
`kb/disorders/` that is **5,126 of 83,058 snippets (~6%)** that CI has never verified —
they are counted as "skipped", and a clean validation run says nothing about them.

This is unrelated to any current work; it surfaced while adversarially probing the
evidence stack for #7439 (see [PR #7442](https://github.com/monarch-initiative/dismech/pull/7442)).

## What un-skipping DOI would actually do

I ran the full corpus with `DOI` removed from `skip_prefixes` and triaged every mismatch,
because **"86 snippets don't match" and "86 snippets are misquoted" are very different
claims** and only one of them is worth alarming anybody about.

```
snippets newly verified clean:                          ~4,900
mismatches surfaced:                                        86
  1. cache mangled  (words match exactly, spacing differs)   29
  2. abstract-only cache, quote is from the full text        22
  3. full text cached, words genuinely absent                35
       of which degenerate (<= 4 words, see below)           13
       needing genuine human review                          22
```

**So the actionable number is 22 snippets, not 86, and not 5,126.** The overwhelming
majority of what the skip currently hides is cache defects, not curation defects.

### Bucket 1 — the cache is mangled, the curator quoted correctly (29)

Two distinct extraction bugs, both in *our* cached text rather than in the KB:

- **Stripped inline markup joins words.** Full-text HTML extraction removes `<i>`/`<em>`
  without inserting a space, so a paper reading "a deletion within the *ANAPC7* locus"
  caches as `...within theANAPC7locus`. The curator's snippet is verbatim correct; no
  amount of whitespace normalization can match it.
- **PDF ligatures.** PDF extraction emits `ﬁ` (U+FB01) and friends, so cached text has
  `amyloid ﬁbrils` where the snippet has `amyloid fibrils`. The validator's
  `normalize_text` does not fold ligatures — confirmed:
  `normalize_text("ﬁbrils") == "ﬁbrils"`.

Both are cheap to fix in normalization (fold ligatures, insert a space at stripped tag
boundaries) and fixing them would rescue all 29.

### Bucket 2 — the cache holds only an abstract (22)

The quote is from the full text, which was never cached, so it cannot be found. Neither a
[...]

PR #10003, opened 2026-08-29
Split evidence direction from directness; retire PARTIAL and WRONG_STATEMENT (#7439)

### LINKED ISSUE #7439: Experimental SEPIO evidence model: schema, validators, and rendering

cmungall:
## Summary

The `supports` field on an evidence item was being asked to answer several different questions at once. This narrows it to direction only — `SUPPORT` / `REFUTE` / `NO_EVIDENCE` — and adds a separate optional `directness` field for how directly the quoted text bears on the claim. `PARTIAL` and `WRONG_STATEMENT` are removed.

This is the partial adoption of the SEPIO evidence model proposed in #7439, agreed in [this comment](https://github.com/monarch-initiative/dismech/issues/7439#issuecomment-5459768804).

**The headline result: retiring `PARTIAL` cut self-contradictory `supports` gradings from 8,285 to 353** — cases where the same quoted sentence from the same paper carries two different values in one file. That is a 96% drop, and it puts `supports` below `evidence_source` (715) for the first time.

## Why directness rather than strength

#7439 proposed SEPIO's direction/strength split. Reading the curators' own `explanation` prose across the 11,620 `PARTIAL` items, the field was not carrying strength. It was carrying at least four different things:

| what the curator meant | example from the KB |
|---|---|
| supports it, but via an inference step | riluzole's clinical benefit, cited as "partial human validation" of excitotoxicity |
| right mechanism, wrong direction of manipulation | an `fgfr3` loss-of-function zebrafish cited against a gain-of-function module |
| supports part of the claim, contradicts another part | "misfolding of the polyQ protein is central, **but** the toxic species may be soluble conformers" |
| true and worth citing, but not about this claim | "off-target skeletal liability of pan-FGFR inhibition" |

Rows 1 and 2 are directness. Row 3 is one evidence item making two claims. Row 4 is `NO_EVIDENCE`. None is strength.

Strength was also the wrong thing to add on its own terms. It needs study design, sample size and replication, none of which the evidence item holds — whereas directness needs the snippet and the claim, both of which are right there. Two curators disagreeing about directness have something to point at; two disagreeing about strength do not. Design decisions §12 already holds that strength should be *derived* rather than authored, and that position is unchanged here.

## What is not adopted

SEPIO's `EvidenceLine` / `DataItem` / `Document` classes. An `EvidenceItem` is already an evidence line holding one data item and one document, and `linkml-reference-validator` supports `ELLIPSIS_INSERTION` — so a single `snippet` already carries a discontinuous multi-span quote, which was most of what the nesting bought. The SEPIO alignment is expressed as `exact_mappings` on the existing dismech-native slots instead, keeping design decisions §5's export-layer-only rule intact. This also sidesteps the vocabulary collision with #7612.

The known cost: evidence *about* other evidence stays unrepresentable. `glymphatic_dysfunction.yaml` has a REFUTE (an independent lab found no AQP4 effect) followed by a rebuttal attributing that negative to anaesthetic regimen and tracer route. The second item is aimed at the first, and a flat list still records both against the node as if they were parallel primary observations.

## The 11,620 migrated items are left unassessed

`PARTIAL` became `SUPPORT` with `directness` **absent**, not backfilled to a default value.

Absent means nobody has assessed the item, which is true of essentially the whole KB. Defaulting them would have manufactured 11,620 appraisals no curator made — and would have broken, on day one, the rule we want next: that setting `directness` should require a justification.

The population is enumerated in `docs/reports/data/partial-evidence-directness-worklist-2026-08-29.tsv` (11,620 rows: file, location, reference, snippet, prior explanation), following the pattern of the existing worklists in that directory. It is 4.9 MB, which is small next to the 638 MB `references_cache/`.

## Three things a reviewer should know

**The migration is a line edit, not a YAML round-trip.** Every `supports: PARTIAL` in `kb/` was the same single-line unquoted scalar. Round-tripping 1,664 files through a YAML writer would reflow folded scalars, which is what `check-folded-hyphens` exists to catch. The diff is exactly balanced — 11,622 added, 11,622 deleted — and contains only `supports:` lines plus one hand-edited item.

**4,130 lines of curator prose still say "Marked PARTIAL because…"**, across 606 files. This was deliberate. That text is the *reason* for the grade and is exactly what the follow-up directness pass needs; only the enum token is stale, the reasoning is intact. Each affected row is flagged in the worklist.

**YAML aliases are a real limitation here.** Some entries define an evidence list once with an anchor and reuse it by alias — `Apert_Syndrome.yaml` has 4 `supports: PARTIAL` lines but 6 parsed items. The line edit handles that correctly, but an aliased item shares one object across several claims and so cannot carry a different `directness` for each, and directness *is* claim-relative. Assessing one of those may mean un-aliasing it first.

## The single WRONG_STATEMENT

One use in the whole KB, in `Huntington_Disease.yaml`. It cites the EHDN juvenile HD working group stating the Westphal phenotype is not exclusive to juvenile onset, and was graded `WRONG_STATEMENT` to record that this entry's own subtype `display_name` was inaccurate.

That name has since been corrected, so the quote no longer contradicts anything in the entry — it now supports naming the subtype by age at onset. Regraded by hand to `SUPPORT` + `directness: DIRECT`, explanation rewritten, and the correction story moved to a `history/` record, which is where curation provenance belongs.

## Changes

- **Schema**: narrowed `EvidenceItemSupportEnum`; new `DirectnessEnum` and optional `directness` slot; `sepio:` prefix and `exact_mappings`
- **Migration**: `scripts/migrate_supports_enum.py` (`--dry-run` / `--apply`), plus the worklist it generates
- **Exporters**: `sepio_export.py`, `hpoa_export.py`; also `render.py`, `compare/d2p.py`, three templates, regenerated datamodel
- **Docs**: design decision §6b, `CLAUDE.md`, `evidence-model.md`, the curation skill
- **Tests**: new `test_render_evidence_directness.py`; updated `test_sepio_export`, `test_hpoa_export`, `test_empty_snippets`, `test_snippet_grading`

### A bug this fixed in the SEPIO exporter

`NO_EVIDENCE` was being exported as SEPIO `NEUTRAL`. Those are different assertions: `neutral` means the evidence bears on the claim without favouring either side, while `NO_EVIDENCE` means it does not bear on the claim at all. Such lines now carry no `direction_of_evidence_provided` and retain the raw value on `dismech_supports`.

## Validation

- `test_data.py` evidence/schema/validity selection: **7,724 passed, 0 failed** across all 2,785 KB files
- Remaining 128 test modules: **3,473 passed, 23 skipped**, 4 failures all confirmed pre-existing (see below)
- Whole-KB `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`: clean
- `linkml-validate` on sampled entries including both anchor/alias files

Pre-existing failures, each verified by running at `HEAD~1` in a detached worktree and confirming identical failure without this change:

- `test_validate_disorder_hook` (2) — `just` is not installed in the container
- `test_history_schema::test_committed_history_records_follow_layout` — three `Robinow_Syndrome` records from 2026-08-27 point at a target that no longer exists; the entry is now `Robinow_Syndrome_Autosomal_Recessive_2`. These look like they want a `target.superseded_by` block
- `test_node_classes::test_committed_tree_examples_resolve_in_kb` — eight `Metastatic_*` tree examples reference entries that do not exist, consistent with design decisions §3a ("stage is never an entry")
- `test_no_new_snippet_grading_divergences` — two `evidence_source` divergences in the chondrodysplasia punctata entries, which CI grandfathers against `origin/main`

This change deletes and renames no KB entry (`--diff-filter=DR` is empty) and adds exactly one history record.

## Follow-ups, not in this PR

- A directness rubric with worked examples, and the rule that setting `directness` requires a justification
- Triage of the ~5,130 items whose explanation contains contrary language ("but", "does not", "inverse", "not sufficient") — the ones where plain `SUPPORT` may assert something the curator explicitly qualified. That is a regex, so it is an upper bound on the risky set, not an error count. The remedy is the one `CLAUDE.md` already gives for mixed-source papers: split into a `SUPPORT` item and a `REFUTE` item
- Gating `supports` in `check-snippet-grading` is now worth revisiting, since 353 divergences is a very different proposition from 8,285

Closes #7439


cmungall:
## `test (3.13)` was red on the lint step — three of the four errors were mine, fixed in 5ba573a

`ruff check .` failed. Three errors were in the new `scripts/migrate_supports_enum.py` and are now fixed:

- **EXE001** — shebang without the executable bit. Every other script in `scripts/` has both, so I set the bit rather than dropping the shebang.
- **RUF100** — my `# noqa: E402` was redundant. `pyproject.toml` already per-file-ignores E402 for `scripts/*`.
- **DTZ011** — `date.today()` is timezone-naive. Now `datetime.now(UTC).date()`, matching `build_dataset_records.py` and the other scripts that stamp a generated filename.

The migration script's dry run still reports 0 remaining `PARTIAL` and 0 `WRONG_STATEMENT` against the migrated tree, so no behaviour changed.

Worth noting the rest of that job passed before the lint step — including `validate-terms-schema`, which is the one that would catch a problem with the new `DirectnessEnum`.

## The fourth error is not from this PR, and this push will not clear it

```
E402 Module level import not at top of file
  --> tests/test_ci_batch_validation.py:10:1
```

That file is not in this PR's diff (`git show --name-only HEAD` does not list it) and was last touched by #9854. Running `ruff check` on it with my changes stashed reproduces the error identically, so it is red on `main` too and is currently failing every PR's lint step.

It is the same recurrence `pyproject.toml` already documents: this repo deliberately tracks ruff's default rule set unpinned, and ruff 0.15 turning E402 on by default previously "turned nine long-standing scripts red at once on main and blocked every PR's lint step." The per-file-ignore added then was scoped to `scripts/*` on purpose — the comment says E402 "still applies to `src/` and `tests/`, where the idiom has no excuse." Now a `tests/` file has hit it. Local ruff here is 0.15.8.

I searched for an open PR fixing it and found none.

### Proposed patch

The `sys.path.insert` in that file is genuinely needed, but not for the import ruff is flagging. It is needed for the *function-local* `from scripts.check_entity_refs import DEFAULT_ROOTS, EXCLUDED_ROOTS` at line 138. The flagged import, `dismech.yaml_io`, resolves without it because `dismech` is an installed package. So the import can simply move up, with no ignore and no change to `ROOT`, which is used at lines 23, 47, 65, 112 and 141:

```diff
 from __future__ import annotations

 import re
 import sys
 from pathlib import Path

+from dismech.yaml_io import safe_load
+
 ROOT = Path(__file__).parent.parent
 sys.path.insert(0, str(ROOT))
-
-from dismech.yaml_io import safe_load
```

I have not applied this, because it is a change to a file this PR otherwise does not touch, and folding an unrelated lint fix into a 1,686-file schema migration makes both harder to review. It is a one-line move and would suit its own small PR — happy to open one, or to include it here, whichever the maintainers prefer. Until one or the other happens, `test (3.13)` stays red on this PR for a reason that is not this PR's.

---

cmungall:
## Good catch on the fixture — confirmed, fixed in 0c986c1

`tests/data/valid/Disease-Antiphospholipid_Syndrome.yaml` had seven `supports: PARTIAL` items. That directory asserts its files validate, so narrowing the enum broke it:

```
'PARTIAL' is not one of ['SUPPORT', 'REFUTE', 'NO_EVIDENCE']
  in /phenotypes/0/evidence/2/supports
```

Reproduced, fixed, and re-validated — `No issues found`, with a diff of exactly those seven lines and nothing else.

**Why I missed it:** the migration script scans `kb/`, and the `test-examples` recipe covering that directory runs through `just`, which is not installed in the container this change was developed in. So neither the migration nor my local test runs reached it. Applied through the migration script's own `rewrite()` rather than by hand, so the transformation is provably identical to the one applied across `kb/`.

### Two other trees hold the retired value and are deliberately left alone

I swept the whole repo rather than just re-running on `kb/`. Two more places have it, and neither should be migrated:

- **`experiments/interannotator/`** — 14 values across two files. These are snapshots of what two curators *independently wrote* during an agreement study. They are evidence about curation, so rewriting them would silently change the measurement being recorded. `CLAUDE.md` keeps `experiments/` outside every validator for exactly this reason, so the retired value there breaks nothing.
- **`history/`** — 67 files mention PARTIAL, but only in `details` prose, never as an enum value (the anchored line grep finds zero). History records are append-only and each accurately describes the session as it ran.

Both exclusions are now recorded in the migration script's docstring so a later sweep does not "fix" them.

### Still outstanding, and not this PR's

The `tests/test_ci_batch_validation.py` E402 described in my previous comment is unchanged and will keep `test (3.13)` red. It is red on `main` too. The proposed one-line patch is in that comment; I have not applied it, since it touches a file this PR does not otherwise change.

---

cmungall:
## All three blocking items are addressed. Two of them were already fixed before the review posted.

Thank you for re-running the checks rather than trusting the PR body — and for catching `docs/sepio-export.md`, which I had missed.

### Blocking items

**Item 1 (`docs/psychiatric-curation-sop.md`) and item 2 (`docs/frequency-evidence-guidelines.md`)** were fixed in 8b60853, which landed while the review was in progress against 0c986c1. Both were done exactly as suggested:

- the imaging-meta-analysis bullet and its YAML example are now `supports: SUPPORT` + `directness: INDIRECT`
- the five-value enum list is corrected and now names all three fields as independent
- the pharmacological-efficacy bullet is rewritten as the canonical `INDIRECT` case, in the words the SOP was already reaching for — it called that evidence "*indirect* mechanistic evidence" before the field existed
- the frequency pattern is `REFUTE` against the band, since a row asserting "Very frequent (99-80%)" does not partly support an `OCCASIONAL` band, it contradicts it. Where the same row also sources the association, that is a second `SUPPORT` item

**Item 3 (`docs/sepio-export.md`)** was a genuine miss, fixed in f444f13. It documented the exporter this PR changes and contradicted it. Rewritten to match, with a `directness` row added to the field-mapping table and a dated note recording what the section used to say and why `dismech_supports` survives (it was added to round-trip the `WRONG_STATEMENT`/`REFUTE` collapse; it is retained because `NO_EVIDENCE` needs it now).

### Suggestions taken

**Item 4** — both counts fixed: "seven fields" → eight, "Two of those fields" → three.

**Item 5** — fixed. `check_empty_snippets.py`'s message was the one that mattered, since it prints to a curator's terminal. In `check_snippet_grading.py` I also added a note that its headline counts predate this change: 8,285 → 353 for `supports` against 715 for `evidence_source`, so the argument recorded there for leaving `supports` ungated is now much weaker than when it was written.

**Item 7** — already done in 8b60853, and I agree it is the strongest independent evidence for the direction/directness choice. The note says what came of the pilot and that the strength question it opens is still open, pointing at §6b and §12.

**Item 10** — added, `#7439` and this PR.

**Item 9** — good catch on the copy-paste hazard, and agreed the files should stay as they are. Already covered by the scope section in the migration script's docstring alongside `experiments/` and `history/`.

### Item 6: regenerated partially, as a deliberate call

I tried the full regeneration. `just gen-doc` touches **228 files**, and `docs/schema/dismech.yaml` alone changes **3,344 lines of which about 40 are this PR** — the rest is drift from earlier PRs that did not regenerate the directory (`EnvironmentalMechanismTarget`, `AnimalModel`, the exposure-classification classes, `ModelFidelityEnum`, and a couple of dozen more that have no page at all).

Sweeping 3,300 lines of somebody else's un-regenerated drift into an already-1,686-file PR would bury both. So I committed only the four pages that are purely about this change — the `EvidenceItemSupportEnum` and `supports` pages, plus the new `DirectnessEnum` and `directness` pages — and reverted the rest. I checked each of the four: their diffs contain nothing but this change. Since the directory is refreshed opportunistically, a partial refresh matches how it is actually maintained, and the accumulated drift wants its own regeneration PR.

### Item 8: your 82 reproduces exactly, and the example is right

I ran your regex over the worklist and get **82 items across 70 files**. `com_Alzheimer_Disease__Cancer.yaml` `hypotheses[1].evidence[2]` is the clearest case — snippet *"No significant associations with cancer were detected for Alzheimer's disease"*, explanation calling the correlation "null", now reading `SUPPORT`. That is `NO_EVIDENCE`, or the item needs re-pointing at a claim it actually bears on.

And your caution holds: `AA_Amyloidosis.yaml` is correctly `SUPPORT`, because the claim there *is* that eprodisate failed its endpoint. So it needs reading, not a second bulk rewrite. Agreed that 82 is a tractable first tranche where 5,130 is not — that is a better follow-up unit than the one in the PR body, and I have noted it there.

### On the pre-existing failures

Confirmed the same way you did. `test_no_new_snippet_grading_divergences` fails on two `evidence_source` divergences in the chondrodysplasia punctata entries; this PR changed only `supports:` lines in those files. `test_history_schema` fails only on the three `Robinow_Syndrome` records whose target was renamed to `Robinow_Syndrome_Autosomal_Recessive_2` — I re-ran it after editing the Huntington record to confirm mine is not implicated. Both reproduce at `HEAD~1`.

Separately, `test (3.13)` is red on a `tests/test_ci_batch_validation.py` E402 that is also red on `main` and is failing every PR's lint step right now. A one-line patch is in [this comment](https://github.com/monarch-initiative/dismech/pull/10003#issuecomment-5460150660); I have not applied it, since it touches a file this PR does not otherwise change.

---

cmungall:
## Suggestion A was a real bug and is fixed in 2bfc09a. Suggestion B: not doing it, now with evidence.

### A — the doc asserted a KB state that did not exist

You were right, and it was my error: I wrote that table row in 8b60853 describing what the entry *should* say without checking what it did say. `Marfan_Syndrome.yaml` carried a single `SUPPORT` item whose explanation did both jobs at once.

I split it rather than reword the row, because that item is precisely the pattern the same paragraph now tells curators to split — rewording would have left the guidance aspirational while its own worked example contradicted it. The band there really is `OCCASIONAL`, so the ORPHA row genuinely contradicts it:

| | snippet | claim |
|---|---|---|
| `SUPPORT` | `HP:0002108 \| Spontaneous pneumothorax` | the association and the more specific phenotype term |
| `REFUTE` | `Spontaneous pneumothorax \| Very frequent (99-80%)` | against the recorded band, naming the PMID:25765122 conflict |

Each half quotes the part of the row carrying its own claim, which is what the split rule asks for. Two details that made this work rather than trade one problem for another:

- Both are exact substrings of `references_cache/ORPHA_558.md`, and both keep a pipe, so they stay inside `check_snippet_length`'s structured-database-row exemption. A bare `Very frequent (99-80%)` would be three words and would have tripped it.
- Neither snippet is a substring of the other, so `check_snippet_grading` does not pair them and **`supports` divergences are unchanged at 353**. The split does not buy its correctness with a new contradiction.

Verified: `linkml-validate` clean, **132/132 snippets verified** against the cache, duplicate-key and entity-ref checks clean. The table row now describes what the file actually contains.

### B — I checked whether one command would finish it, and it would not

The claim needed testing rather than restating, so I ran `gen-yaml` against **unmodified `origin/main`** in a detached worktree:

```
$ git worktree add --detach <tmp> origin/main && cd <tmp>
$ gen-yaml src/dismech/schema/dismech.yaml | sed <date normalization> > docs/schema/dismech.yaml
$ git diff --numstat -- docs/schema/dismech.yaml
2932    46    docs/schema/dismech.yaml
```

`docs/schema/dismech.yaml` is ~2,978 lines out of date with the schema on `main` before this branch touches anything. So `just _gen-yaml` would import ~2,932 lines of pre-existing drift alongside this change's ~40 — the same reason the full `gen-doc` was backed out.

The asymmetry you noticed is real but is a property of the artifacts, not an oversight: the per-enum and per-slot pages are separate files and could be refreshed in isolation, so they were. The merged YAML is monolithic and cannot be, short of hand-editing a generated file. Agreed it is cosmetic and unenforced; the accumulated drift wants its own regeneration PR, and that PR would now also pick up this change.

### C

Agreed, and unchanged — the 82-item null-finding tranche is the tractable first pass, out of scope here, and recorded in the PR body as the follow-up unit.

Thanks for the review. Catching both the fixture and this one, each of which needed the file opened rather than the diff skimmed, materially improved the change.

---

cmungall:
## You're right, and the reasoning error was mine. Reverted in e469794.

I reproduced it before changing anything:

```
$ hpoa_rows_for_disorder(Path('kb/disorders/Marfan_Syndrome.yaml'))  # at 2bfc09a
qualifier=''     ref=PMID:31883816   ...
qualifier=''     ref=PMID:25765122   ...
qualifier=''     ref=PMID:33896616   ...
qualifier=''     ref=PMID:21252480   ...
qualifier=''     ref=ORPHA:558       ...
qualifier='NOT'  ref=ORPHA:558       ...
```

Exactly as you described — a `NOT` row asserting Marfan patients cannot get a spontaneous pneumothorax, sourced to the Orphanet row saying it is nearly universal, contradicting the positive row from that same reference.

**The mistake:** I treated `supports` as scoped to whichever claim the `explanation` named. It isn't. `Phenotype` has one flat `evidence:` list and no frequency-scoped slot, so `supports` is scoped to the phenotype–disease assertion and nothing else. What makes it worse is that `hpoa_export.py`'s `REFUTE` → `NOT` mapping is in this PR's own diff — I edited that exact table — and I still didn't check the consumer before writing to the producer.

Please don't apologise for suggestion A. It correctly identified that the doc asserted a KB state that didn't exist, and offered both options; I picked the wrong one and didn't verify it.

### What changed

Reverted to a single `SUPPORT` item quoting the full row, with the band disagreement in its `explanation`. `HP:0002108` is back to **five positive rows, zero `NOT`**.

Worth noting: the phenotype's `notes:` **already** recorded the conflict correctly, before any of this —

> Orphanet classifies spontaneous pneumothorax as Very frequent (99-80%), whereas PMID:25765122 reports 5-11%; frequency is retained as OCCASIONAL based on the direct quantitative clinical estimate.

So the split added nothing at all except the false exclusion. The pre-existing curation was already right.

### The guidance mattered more than the entry

Option 2 of `docs/frequency-evidence-guidelines.md`, as I rewrote it in 8b60853, prescribed `REFUTE` for *every* frequency conflict — so this would have propagated to each future one, exactly as you said. It now prescribes prose: `SUPPORT` for the association, the disagreement named in `explanation` and `notes:`. I added a warning block under it explaining why `REFUTE` is unavailable here, since the reason is non-obvious and someone will reach for it again:

> `supports` is scoped to the phenotype-disease assertion — "does this phenotype occur in this disease" — and not to the band. A `REFUTE` therefore reads as "this phenotype is absent", and `hpoa_export.py` maps it straight to an HPOA `NOT` qualifier.

It also records that `REFUTE` was tried here and reverted, so the next person doesn't repeat it.

### Regression test

`test_frequency_disagreement_must_not_be_curated_as_refute` pins two things: that this shape emits no `NOT` row, and the general invariant that **one reference must never yield both a positive and a `NOT` row for the same HP term**. That second assertion is the one that would have caught this independent of the Marfan specifics.

### Verification

| Check | Result |
|---|---|
| `hpoa_rows_for_disorder` on `Marfan_Syndrome.yaml` | 5 positive `HP:0002108` rows, **0 NOT** |
| `linkml-validate` | No issues found |
| `reference_snippet_audit` | 131/131 verified |
| `check_snippet_length` | clean |
| `test_hpoa_export` | 28 passed |
| `ruff` | clean |

On your closing point — I agree the deeper answer is a frequency-scoped affordance, and I agree it is well outside a `supports`-enum migration. Worth its own issue; the warning block is the interim guard.

---

cmungall:
## Both suggestions taken, in 8c717eb

### A — the collision was mine, and the fix was more than a clause

You were right that my rewrite created it: option 2's new headline "Keep the band, record the disagreement in prose" reads as the direct opposite of "Do not keep the band and merely note the mismatch in prose" five lines below. It was legible when option 2 said `PARTIAL`, and stopped being so the moment I made prose the prescription.

I qualified that paragraph to "a band whose *only* source is the broader entity", and added a sentence naming what actually separates the options — because on re-reading, the distinction is not where the mismatch is written down but **what the band rests on**:

> Option 2 keeps a band that has its **own narrower quantitative source** and records the broad source's disagreement in prose, because the band is defensible without it. Option 3 applies when no such source exists: there the band would rest entirely on the broad row, and prose cannot rescue it.

### B — you were right, and the guard now demonstrably fails

The critique lands: `test_frequency_disagreement_must_not_be_curated_as_refute` builds both items as `SUPPORT`, so it pins that the correct shape exports cleanly while the invariant it states is checked only against data that cannot violate it. Re-adding a `REFUTE` to the real entry would not have failed it.

Added `test_marfan_pneumothorax_exports_no_exclusion`, which reads `kb/disorders/Marfan_Syndrome.yaml` directly. I verified it can actually fail rather than assuming:

```
# inject the reverted REFUTE item back into the real file
$ pytest tests/test_hpoa_export.py::test_marfan_pneumothorax_exports_no_exclusion
FAILED — 1 failed

# restore the file
$ pytest tests/test_hpoa_export.py::test_marfan_pneumothorax_exports_no_exclusion
1 passed
```

Scoped to the one entry deliberately, for the reason your scan established: **11 `(reference, HP term)` pairs across `kb/disorders/` already emit both a positive and a `NOT` row**, all pre-existing, so the general invariant cannot be a whole-KB gate without a baseline. Both the scope and that finding are recorded in the test docstring so the next person doesn't rediscover it.

That backlog is worth its own issue — `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_2.yaml` alone has three terms on `PMID:15159508`, and a self-contradicting export from one reference is the same defect class as the one this PR just fixed, only older.

### On the HPOA-neutrality check

Worth flagging that you verified something I had asserted only indirectly: `PARTIAL` was in neither `DROP_SUPPORTS` nor `SUPPORTS_TO_QUALIFIER`, so it already fell through to `default_qualifier` exactly as `SUPPORT` does — which means rewriting 11,620 items changes no HPOA row at all, and the only behavioural change is dropping `WRONG_STATEMENT` from the qualifier map. That is a stronger statement than "the migration looks safe", and it is the one a downstream HPOA consumer would want. Thanks for running it.

`test_hpoa_export` is 29 passing; `ruff` clean.

---

cmungall:
## Correction: the E402 was never the CI blocker. I was wrong about that.

I said twice that `tests/test_ci_batch_validation.py`'s E402 was keeping `test (3.13)` red. It is not, and I should have read a full job log before saying so rather than reasoning from my local `ruff`.

**CI's ruff does not flag it.** The lint step on `94c08bb` reported exactly three errors, all in my own new script, and none in `tests/test_ci_batch_validation.py`. Those three were fixed in 5ba573a and the lint step has passed ever since. The E402 appears only under my local ruff 0.15.8; CI resolves a different version. My analysis of that file was correct in isolation and irrelevant to this PR — please disregard the request for a decision on it.

**What actually fails.** The job gets past lint and into pytest, where it ends `2 failed, 11281 passed`:

```
FAILED tests/test_history_schema.py::test_committed_history_records_follow_layout
  history/disorders/Robinow_Syndrome/2026-08-27T011503Z-claude-code-603126.yaml target does not exist
  history/disorders/Robinow_Syndrome/2026-08-27T055409Z-claude-code-cfe67a.yaml target does not exist
  history/disorders/Robinow_Syndrome/2026-08-27T180508Z-claude-code-22d0de.yaml target does not exist

FAILED tests/test_node_classes.py::test_committed_tree_examples_resolve_in_kb
  line 691:  no such entry 'Metastatic_Ovarian_Cancer'
  line 939:  no such entry 'Metastatic_HCC'
  ... 8 in total
```

Both are on my known pre-existing list, and both reproduce at this PR's merge base (`ed4dc2b`) in a detached worktree: `2 failed in 31.37s`, identical.

**Why they are not routinely red.** CI selects pytest by changed path. A curation-only PR touches `kb/` and matches neither the `python` nor the `schema` filter, so these never run. This PR touches `src/`, `tests/` and `scripts/`, so it triggers the python lane and surfaces them. That is the same gap `CLAUDE.md` documents as the reason `check-entity-refs` is an ungated CI step rather than only a test.

### Both have a clear fix, neither is this PR's

**`Robinow_Syndrome` history records.** The entry was renamed to `Robinow_Syndrome_Autosomal_Recessive_2`. `CLAUDE.md` documents the exact remedy, and it is specifically *not* to rewrite `target.slug` — history records are append-only and those records describe their sessions accurately. Add a `target.superseded_by` block and move the files into the successor's directory:

```yaml
target:
  kind: disorder
  slug: Robinow_Syndrome          # unchanged — frozen
  path: kb/disorders/Robinow_Syndrome.yaml
  superseded_by:
    slug: Robinow_Syndrome_Autosomal_Recessive_2
    path: kb/disorders/Robinow_Syndrome_Autosomal_Recessive_2.yaml
    reason: Entry renamed; ARR2 is the successor.
```

`test_committed_history_records_follow_layout` accepts a missing `target.path` only when `superseded_by.path` resolves, so this is the sanctioned shape.

**`Metastatic_*` node-class examples.** Eight leaves in the node-class tree cite entries that do not exist, and by design decisions §3a never will — "stage is never an entry": metastatic disease is `stages:` on the parent plus `conforms_to` on `invasion_and_metastasis`. So the tree examples want repointing at the parent cancer entries, or removing. That is a judgement call for whoever owns that tree, not a mechanical fix.

I have not made either change. Both are in files this PR does not otherwise touch, and neither is small enough to fold into a 1,697-file enum migration without obscuring it — the history fix spans three records plus file moves, and the node-class fix needs a call on what those examples should point at. Happy to do either here or separately if that is what is wanted.

**This means `test (3.13)` will stay red on this PR until they are fixed somewhere.** It is not caused by this change and it is not fixable within its scope, but I would rather say that plainly than leave the earlier, wrong explanation standing.

---

cmungall:
## Suggestion A taken in 6979612 — you were right that deleting lost an example for no reason

I verified the three things the suggestion rests on before acting:

- `Diffusely Infiltrative Signet-Ring Cell Carcinoma` is a **pathophysiology** node in `CDH1-Related_Hereditary_Diffuse_Gastric_Cancer.yaml:228` (checked as a parsed `pathophysiology` entry, not just a name grep — a phenotype with that name would not satisfy `verify_examples`)
- that entry occurs **zero** times in the tree
- restoring it leaves **0 within-class duplicate leaves** across the whole tree

`test_node_classes` passes at 23, which is real verification here since `verify_examples` checks exact `(node_name, entry_stem)` pairs against `kb/`.

My reasoning in `cb9ffe0` was right about `Gastric_Adenocarcinoma` and wrong to stop there. Having established that one substitute would duplicate, I treated the leaf as unfixable instead of asking whether a different gastric entry would serve — so the class lost its diffuse-gastric example for no reason. Conservative, but worse than the alternative.

### On suggestion B — agreed, and it is now a real wrinkle rather than a hypothetical

`superseded_by` has no `kind`, so a cross-kind supersession leaves `target.kind: disorder` pointing at a `Grouping`, and `_layout_errors` still derives the expected directory from `KIND_DIRS[kind]`. This case only works because the grouping's stem happens to equal the original slug. Had the grouping been named anything else, these records would have had to live in `history/disorders/<grouping-stem>/` — a directory named after a grouping, inside the disorders tree.

Worth an issue, and out of scope here. I have not filed one, since opening issues is not something I do unprompted on this PR; flagging it so it is not lost.

### On the ARR2 correction

Thank you for checking the grouping members and the record's `details` rather than defending the suggestion. For what it is worth, the thing that made it catchable was that the record documented its own supersession in prose at the time — the `details` field said the lumped entry was replaced by a grouping, written before any of this came up. Without that, the ARR2 guess would have looked entirely reasonable and I would probably have taken it.

### Remaining state

`test_no_new_snippet_grading_divergences` is still the one failure, still the two chondrodysplasia punctata entries, still `evidence_source` divergences in files where this PR changed only `supports:` lines. Nothing else outstanding that I know of.

---

cmungall:
## The remaining `test (3.13)` failure was not pytest at all — fixed in 08c946a

With the two pytest failures cleared by `cb9ffe0`, the job got further and failed in a different step: `just validate-disorders`.

```
Validation Summary:
  Files validated: 1621
  Issues found in 140 file(s) (346 validation issue(s))
✗ Validation failed for one or more disorder files
error: recipe `validate-disorders` failed with exit code 1
```

All 346 are the same shape — a dataset accession the validator has no source for:

```
No source found for reference type: EGA:EGAS00000000027
  [WARN] Could not fetch reference: ega:EGAS00000000027
    Location: datasets[3].title
```

### Not this PR's content, but this PR's breadth

The KB diff is 11,617 `supports:` lines, 2 `directness:` lines and two rewritten explanations. Grepping it for `datasets|accession|ega:|arrayexpress|metabolomics` returns **0**. And `Tuberculosis.yaml` produces the identical three EGA warnings at the merge base in a detached worktree.

What changed is scope. `validate-disorders` validates a PR's *changed* files; an ordinary curation PR passes a handful, this one passes 1,621. So a whole-KB backlog lands in a single run for the first time — the same reason this PR surfaced the `Robinow_Syndrome` and `Metastatic_*` failures.

### The cause is an omission, not a policy question

`conf/reference_validator_config.yaml` has a `skip_prefixes` list whose stated purpose is exactly these:

> reference prefixes the validator cannot fetch and should silently skip. These are dataset accession identifiers, not literature references, so snippet validation does not apply to them.

Comparing that list against the dataset-repository prefixes declared in `dismech.yaml`, **six were missing while twelve siblings were present**:

| present | missing |
|---|---|
| geo, sra, pride, dbgap, gtex, metabolights, mgnify, proteomexchange, massive, bioproject, scea, cellxgene | **arrayexpress, ega, metabolomics_workbench, hca, encode, synapse** |

Added all six in both cases, per the existing convention. Only the first three occur in `kb/` today — `hca`, `encode` and `synapse` are included so the next entry curated against one of them does not reopen this.

### Verification

The four worst-affected files went from 3 / 6 / 3 / 1 issues to **0 each** (`Tuberculosis`, `Triple_Negative_Breast_Cancer`, `Uveal_Melanoma`, `Trichothiodystrophy`).

Because this change makes a validator quieter, I checked it does not weaken the guardrail rather than only checking it goes green:

- **255/255 literature snippets still verify** across `Marfan_Syndrome` and `Tuberculosis`
- replacing a real snippet with a fabricated sentence **still fails**: `Text part not found as substring: 'This sentence is fabricated and appears in no paper whatsoever.'`

So only unfetchable dataset accessions are skipped, which is what the list is for.

### One thing worth knowing separately

Running the reference validator locally **rewrites reference cache files lossily**. It rewrote `references_cache/clinicaltrials_NCT03568383.md`, dropping `full_text_attempted: true` and unquoting `reference_id`. I reverted it rather than committing, since it drops a field and is unrelated to this PR — but any agent running the validator and then `git add -A` would commit that silently. Probably worth its own issue alongside the `superseded_by` one.

---

cmungall:
## Suggestion A was a real error in my comment — corrected in d274e50

You're right, and it was the worst kind of comment error: it contradicted the change it was documenting. `synapse` **is** used —

```
kb/disorders/Alzheimer_Disease.yaml:5556:- accession: synapse:syn26223298
```

— and that accession was one of the warnings 08c946a fixed. I asserted a KB state instead of counting it. Replaced the guess with actual counts of `accession:` values:

| prefix | uses in `kb/` |
|---|---|
| ega | 391 |
| metabolomics_workbench | 85 |
| arrayexpress | 1 |
| synapse | 1 (Alzheimer_Disease) |
| hca | 0 |
| encode | 0 |

So only `hca` and `encode` are unused. Comment-only change; config still parses at 41 entries and `Alzheimer_Disease` still reports 362/368 verified with 6 skipped by prefix.

**Suggestion B** recorded in the same comment while the file was open — the paired upper-case variants are cosmetic, since the validator uppercases both sides before comparing. Worth stating so nobody later assumes the pairing is load-bearing and "fixes" a missing pair that isn't missing.

**Suggestion C** noted, no action: `DOI` in `skip_prefixes` predates this PR and is tracked as #7450. Useful to have it recorded here so the dataset-prefix additions aren't later blamed for it.

**Suggestion D** — agreed, and I hit the same thing. The `references_cache/clinicaltrials_NCT*.md` rewrites come from invoking the validator directly rather than through `just validate-disorders`, which wraps it in `fix-references-cache`. It rewrote one file for me too, dropping `full_text_attempted: true` and unquoting `reference_id`; I reverted rather than committing. Given it has now bitten both of us in the same session, that seems worth an issue of its own — a validator that lossily mutates a committed cache as a side effect of being run is a trap for any agent that follows with `git add -A`.

### On the verification

Two things you did that I hadn't: proving the skip suppresses **zero** verifications via `--unskip-prefix` giving an identical 691/697, and noting that `ALLOWED_REFERENCE_PREFIXES` in `tests/test_data.py` would reject an evidence item citing one of these prefixes anyway — so the skip cannot swallow a literature snippet even in future. That second point is the stronger guarantee, since it holds regardless of what gets curated later. My before/after check only established the current state.

---

cmungall:
## Round 9: the failure was schema validation, and it will keep coming back

The `test (3.13)` failure on `d274e5067` was **not** the reference validation step. That step printed `All validations passed!` and `106817/110529` snippets verified. The failing step was ~4,000 log lines earlier:

```
[ERROR] [kb/disorders/CHST3-Related_Skeletal_Dysplasia.yaml/0] 'PARTIAL' is not one of
        ['SUPPORT', 'REFUTE', 'NO_EVIDENCE'] in /phenotypes/20/evidence/2/supports
```

The branch was 39 commits behind `main`, and curation merging in the meantime keeps writing `supports: PARTIAL`. CI validates the merge, so main's newer entries reintroduced the retired value.

### What I did

**Rebased onto `origin/main`** — clean, no conflicts. Post-refresh diff reviewed: additions only, no unexpected deletions, no protected-path churn.

**Re-ran the migration** — 187 items across 34 files. Before doing so I made the script *merge* the worklist instead of overwriting it. A second run only sees what arrived since the first, so re-running it as written would have replaced an 11,618-row triage worklist with 187 rows. It is now 11,805 rows.

**Documented a scope decision I had not written down**: `kb/hypotheses/**/*.md` keeps the retired value. Those are deep-research provider reports and their citation sidecars, consumed as a curation input like `research/`. Rewriting a report to match a schema it was never validated against would misrepresent what the provider returned. The 76 assessment YAMLs in that same tree are in scope and were already clean.

**This will recur.** Curation merges faster than this PR can clear review, so every refresh brings new `PARTIAL` values. That is why the script is now safely re-runnable rather than my just patching the 187. The longer it sits, the more sweeps it needs.

### Three failures inherited from main

Rebasing surfaced three more. All three reproduce on `origin/main` at `507d1f845` with none of this branch applied, and none of the 40 open PRs fixes them. They land here rather than on the PRs that introduced them because CI selects pytest by changed path: a curation-only PR matches neither the `python` nor the `schema` filter, so the whole-KB checks written to protect KB content are exactly the ones a content-only PR skips. This PR trips all three filters. That is the same gap CLAUDE.md already records for `check-entity-refs`, now hitting three more checks.

**`test_animal_model_mechanism_targets`, CAPRIN1.** `ModelMechanismLink.target` takes a bare node name, not an entity reference, so `phenotypes#Autism` matched nothing. The same file gets it right two links above (`target: Impaired Dendritic mRNA Localisation`). Dropped the prefix on the link and on its readout, which has to repeat it. The other `phenotypes#` refs in that file are `pathophysiology[].downstream[].target`, where the prefixed grammar is correct, and are untouched.

**`check-snippet-grading`, RCDP5.** One sentence from `PMID:26220973` graded `OTHER` under `pathophysiology` and `HUMAN_CLINICAL` under `differential_diagnoses`. Both items cite the same finding — which genes cause RCDP2-4 — and the paper reports human patients, so I aligned the outlier.

**`check-snippet-grading`, CDPX2.** Here the divergence was right and the quote was wrong. The sentence from `PMID:10391218` reports the Td mouse substitution and the EBP alterations in seven CDPX2 patients in one breath, so `HUMAN_CLINICAL` at the genetic site and `MODEL_ORGANISM` at the animal-model site were both defensible readings of it. I split it along the clause each item actually rests on — their `explanation` text already said which — so each quote carries a single evidence source. Both halves re-verify as exact substrings of the cached abstract.

No baseline was updated for any of them; all three are real defects with real fixes. History records added for the three entries. The 1,600-file enum sweep deliberately gets none: it is a mechanical rename that changes no claim, and one record per touched entry would bury the curation record it is meant to be.

### Validation

| Check | Result |
|---|---|
| Batch schema validation, 1621 changed disorder files | passed |
| Batch term validation, same 1621 files | passed |
| Reference validation on the three edited files | 113/113 snippets verified, no cache churn |
| `check-duplicate-keys` | 4405 files, none |
| `check-entity-refs` | 2742 KB files, all resolve |
| `check-snippet-grading` | no new divergences |
| `check-folded-hyphens` / `check-snippet-length` / `check-title-snippets` / `check-empty-snippets` | no new findings |
| Retired enum values left in `kb/` and `tests/` | 0 |
| `pytest -k history` | 29 passed |
| Targeted `test_animal_model_mechanism_targets` + `test_model_readout_targets_match_link` | 5238 passed |
| Full suite before the CAPRIN1 fix | 27,360 passed, 1 failed (that defect) |

A full-suite re-run was still going when I pushed — 60% in with no failures. I will report it if anything turns up.

### Carried forward, unchanged

The 82-item null-finding tranche is still the tractable first pass at the `directness` follow-up, and still out of scope here. On the reviewer's housekeeping note: I confirmed after the rebase that `CLAUDE.md` and `.claude/skills/initiate-new-disorder-creation/SKILL.md` still carry this PR's narrowed-enum text at HEAD, and my working tree is clean. All staging on this branch has been path-targeted, never `git add -A`.

---

cmungall:
## Follow-up on the full suite: a fourth inherited failure, now fixed

I said I would report the full-suite re-run if anything turned up. It did. Final result was **3 failed, 49,863 passed, 42 skipped**. My earlier run used `-x`, so it stopped at the CAPRIN1 failure and never reached these — that is on me, and it is why the round-9 comment described the suite as clean when it had only been clean up to the first stop.

**`test_failure_to_recapitulate_links_are_substantiated`, Polycystic_Kidney_Disease_3** — real, and inherited like the other three. The entry landed in #10007; this branch only rewrote five `supports:` values in it, none in `animal_models`, and the defect reproduces on `origin/main` untouched.

The Ganab +/- link declares `FAILS_TO_RECAPITULATE` with `limitations` and two evidenced readouts, but no evidence on the link itself. Those are different claims, as CLAUDE.md sets out: readout evidence attests that a measurement was made in a direction; link evidence attests that the model is informative for the node at all. For a substantive negative claim the schema requires both. Added the authors' own conclusion from `PMID:32550232` as the link-level item — verified as an exact substring of the cache — and changed no readout or limitation text. History record added.

**The other two are local only.** `test_recipe_available_detects_a_present_recipe` and `test_recipe_available_detects_a_checkout_predating_the_recipe` both call `recipe_available`, which shells out to `just`. `just` is not installed in this container, so `subprocess.run` raises `FileNotFoundError` and both error regardless of what they are asserting. Nothing to fix; CI has `just`.

Re-verified after the fix: `linkml-validate` clean, reference validation 31/31 snippets with no cache churn, `check-snippet-grading` and `check-entity-refs` clean, and 5,268 passed across the history, `failure_to_recapitulate` and `model_readout` selections.

Pushed as `18fcd3c61`.

### Four inherited failures, one cause

That is now four defects on `main` surfaced by this PR rather than by the curation PRs that introduced them, all for the same reason: CI selects pytest by changed path, so a curation-only PR matches neither the `python` nor the `schema` filter and skips the whole-KB checks written to protect KB content. This PR trips all three filters, so it runs the full lane and inherits the backlog. CLAUDE.md already records this gap for `check-entity-refs`; it applies to `test_data.py` in general.

The nightly sweep (`.github/workflows/nightly-kb-sweep.yaml`) is the intended backstop and did not catch these, which is worth a look on its own — but not from this PR.

---

cmungall:
## The merge-key finding was right — fixed in `ef0a5ce74`

I verified it before acting on it, and it reproduces exactly as described: the sweep commit's `187 insertions, 2 deletions` were two real losses. `grep -c PMID:38078000` on the worklist returned 0, and the two deleted lines are the run-1 rows at `discussions[0].evidence[0..1]` of `Acromesomelic_Dysplasia_Maroteaux_Type.yaml`.

My reasoning for the two-part key was wrong in a specific way worth naming: I assumed an index only ever *shifts*, and wrote a docstring paragraph justifying tolerance of a stale index on that basis. But an index is *reused*. A curator inserted a discussion at index 0, the arriving items took exactly the keys the first run had recorded for two other items, and `setdefault` dropped the older ones. The tolerance argument I made covers a stale index; it does not cover losing an item, and I extended it to a case it did not reach.

Applied the suggested key, `(file, location, reference, snippet)`. Both rows differ in PMID and quote, so both survive, and a genuine re-collection of the same item still dedupes.

I went one step further than the suggestion on the restore. Rather than reinstating the two rows at their recorded `discussions[0]` indices, I re-located them against the current KB by matching `(reference, snippet)`: they are now at `discussions[2].evidence[0..1]`, which is where they actually sit. Worklist is 11,807 rows — 11,620 from the first run plus the 187 the sweep found, matching your arithmetic.

Verified idempotent: a second `--apply` rewrites 0 items, carries all 11,807 rows, and leaves the worklist byte-identical.

Rows carried from the first run may still hold an index a later insertion has shifted. That is the trade the docstring makes deliberately and it costs a curator one search; the docstring now says which case is tolerated and which is not, so the distinction I got wrong is written down.

### Also pushed since your review

`18fcd3c61` fixes a **fourth** failure inherited from `main`, found by the full-suite run I let finish after pushing (my earlier run used `-x` and stopped at the CAPRIN1 failure). `test_failure_to_recapitulate_links_are_substantiated` on `Polycystic_Kidney_Disease_3` — the entry landed in #10007, this branch only rewrote five `supports:` values in it and none in `animal_models`, and it reproduces on `origin/main` untouched. The Ganab +/- link declared `FAILS_TO_RECAPITULATE` with `limitations` and two evidenced readouts but no evidence on the link itself. Added the authors' own conclusion from `PMID:32550232`, verified as an exact substring of the cache; no readout or limitation text changed. History record added. Details in the [previous comment](https://github.com/monarch-initiative/dismech/pull/10003#issuecomment-5461272158).

On your lint caveat: `ruff` runs clean here on the script, and CI's lint step remains the authority.

The 83-item null-finding tranche stays out of scope, carried forward as the first `directness` pass.

---

PR #10005, opened 2026-08-29
curate: Add Ring Chromosome 14 syndrome (MONDO:0014708)

### LINKED ISSUE #9988: Curate ring chromosome 14 (MONDO:0014708)
Curate a dismech entry for **ring chromosome 14** ([MONDO:0014708](https://monarchinitiative.org/MONDO:0014708)) — a constitutional chromosomal disorder in which a ring-form chromosome 14 produces early-onset drug-resistant epilepsy, intellectual disability, microcephaly, a characteristic retinal pigmentary change, and recurrent respiratory infections.

`Ring_Chromosome_20_Syndrome.yaml` is the precedent — ring chromosome syndromes are curated as individual entries, not folded into a `ring chromosome disorder` grouping, because each one's phenotype tracks the specific chromosome involved rather than ring formation as such.

The mechanistically interesting part, and the thing to keep the pathograph honest about: the epilepsy in r(14) is not well explained by terminal deletion alone. Ring instability (dynamic mosaicism through sister-chromatid exchange and ring loss) and position-effect silencing of intact 14q genes are the competing accounts, and the phenotype is notably more severe than comparable linear 14q deletions. That asymmetry is a good candidate for `mechanistic_hypotheses` with the deletion and ring-instability models as separate groups, rather than being asserted one way.

Stub: `stubs/Ring_Chromosome_14.yaml`

gingin77:
Ring chromosome 14 syndrome is a rare chromosomal disorder in which the two ends of chromosome 14 fuse into a circle. It is dominated by an epilepsy that starts in the first months of life, resists drugs, and occurs in essentially every patient, alongside intellectual disability, autism, microcephaly, short stature and a characteristic retinal change.

**The mechanistically interesting part is that the deletion does not explain the syndrome.** Ring formation usually snips off a small piece of the long arm — and in about a fifth of patients it removes nothing at all. But when you line the features up against ordinary linear deletions of chromosome 14, the epilepsy and the retinal abnormality map to a region near the top of the long arm that is **intact in every ring patient**. So the two defining features come from a region nobody has lost, and the piece that *is* missing sits somewhere else entirely, where it plausibly explains the behavioural problems and the infection susceptibility by simple gene dosage.

This entry models that as two parallel branches out of one lesion, and does not pretend the unexplained branch is solved.

## The mechanism graph

```
Ring Chromosome 14 Formation                       (MOLECULAR)
  ├─> Terminal 14q32 Haploinsufficiency            (MOLECULAR)  ─> Behavioural Dysregulation
  ├─> Position-Effect Dysregulation of Proximal 14q (MOLECULAR) ─┐
  └─> Ring Mitotic Instability / Dynamic Mosaicism (CELLULAR)   ─┤
                                                                 ├─> Cortical Hyperexcitability (TISSUE)
                                                                 │      └─> Cognitive Decline
                                                                 └─> Retinal Pigmentary Abnormality (TISSUE)
```

The 14q32 branch is ordinary haploinsufficiency and is the well-behaved part. The other two branches carry the epilepsy, and both are curated as `EMERGING` hypothesis groups rather than asserted:

- **Position effect** — circularisation changes the chromosome's chromatin architecture, and heterochromatin spreading from the short arm may silence proximal genes that are physically present in two copies. The cleanest evidence is a patient whose ring lost *only its telomeres* and who still had 2 of 8 chromosome-14 genes downregulated. FOXG1 is the named candidate.
- **Dynamic mosaicism** — rings segregate badly. One patient studied for it showed micronuclei, aneuploidy and anaphase lags, so the effective genotype differs between cells and drifts with age.

Every edge into `Cortical Hyperexcitability` is typed `INDIRECT_UNKNOWN_INTERMEDIATES`, and that node is graded `HYPOTHETICAL`. No causal gene, ion channel, synaptic or network abnormality has ever been identified in this syndrome. Modelling the gap seemed better than inventing a molecular chain the literature does not support.

## Deliberate omissions

**No `genetic:` section**, following the `Ring_Chromosome_20_Syndrome` precedent. There is no causal gene. FOXG1 is a positional candidate from a single atypical case — a ring formed by interstitial excision, not the usual telomere fusion — whose breakpoint fell 225 kb from FOXG1 and whose FOXG1 expression was never measured. Typing that `CAUSATIVE` in a genetic block would misrepresent it, so it lives inside the position-effect hypothesis where the reasoning can be stated.

**No named 14q32.33 gene**, and this one was tempting. The deep-research report identifies the *IGH* immunoglobulin heavy-chain locus as the obvious explanation for the infection susceptibility, and *PACS2* as an epilepsy candidate in the same band. IGH in particular would upgrade the 14q32 node from unnamed dosage loss to a specific lesion. But no source in the reference cache makes that claim *about ring 14 patients*, and naming a gene on positional plausibility is exactly the failure mode the rest of this entry avoids. Both are recorded in `notes:` as leads worth chasing.

## Phenotypes and frequencies

24 phenotypes. Frequencies come from the HPO disease annotation for `ORPHA:1440`, which carries **Orphanet's own curated frequency classes** (Very frequent / Frequent / Occasional) rather than raw patient counts — so these map onto `FrequencyEnum` directly, and the band is Orphanet's rather than my inference. Provenance is in `notes:` rather than a manufactured quote, since HPOA is a derived annotation.

One tension is recorded rather than resolved: Orphanet classes the retinal abnormality `OCCASIONAL`, while the 27-patient series describes it as a consistent feature. Orphanet's class is retained, with a note saying the two sources are counting different things and this entry has no grounds to prefer one.

## What the deep-research cross-check actually added

Worth stating, because it was not decoration. The report surfaced the **2017 consensus guideline** ([PMID:28399932](https://pubmed.ncbi.nlm.nih.gov/28399932/)), which my literature search had missed. It changed the entry materially: autism spectrum disorder as a defining feature, feeding difficulty and nonverbal status as curated phenotypes, the nutritional-support treatment with gastrostomy escalation, anticonvulsive therapy from onset, and an explicit statement of the diagnostic order that is far stronger evidence for the karyotype-first argument than the single case report I had.

The report's own validation was read and respected: one off-topic reference and two obsolete GO terms (`GO:0016458`, `GO:0006486`) were flagged, and none of the three is used here.

## Diagnosis

The clinically load-bearing point: **a normal microarray does not exclude this diagnosis.** A complete ring that lost only telomeric material is balanced in copy number, so the array is normal while the karyotype shows the ring. The guideline states the order directly — karyotype is primary, and an array-detected terminal 14q deletion should trigger a karyotype to look for a ring.

## Knowledge gaps

Three, all `OPEN`: the unexplained route from ring to cortical hyperexcitability; the fact that FOXG1 expression has never been measured in a *classical* ring 14 patient (with a proposed experiment); and the absence of any correlation between degree of ring instability and phenotype severity, which is what would make the mosaicism hypothesis predictive rather than merely compatible with everything.

## Validation

`just validate-disorders` passes. Snippets **46/46** verified. `validate-terms`, `check-entity-refs`, `check-duplicate-keys` all pass. Compliance 85.4%. Cache diffs are pure additions. Stub deleted, history record included.

Closes #9988


gingin77:
All four fixed. Your diagnosis of the root cause was right and worth restating: I cached the guideline as full text and then only mined its abstract, so its graded recommendations sat unused in the repository the whole time.

## 1. Ophthalmic surveillance term

Rebound to `NCIT:C38060 Eye Examination`. `NCIT:C17230 Ultrasound Imaging` was plainly wrong — I had flagged the poor binding in `notes:` and left it, which was the wrong call when a correct term existed and was already in use elsewhere in the KB.

The Grade B fundus recommendation is now the primary evidence, with the single-case support kept as secondary.

## 2. Treatments

Five added, all from the guideline: ketogenic diet / VNS for the drug-resistant half, early physical therapy (Grade B), speech therapy for dysphagia (Grade A), extended inactivated vaccination (Grade C, specific vaccines named), and vitamin D at twice RDA.

Two deliberate limits. **VNS is described in `notes:` rather than bound** — NCIT's only reachable stimulation term is *Transcutaneous Auricular* Vagus Nerve Stimulation, a different non-implanted modality, and binding it would misstate the intervention. And **the seven named ASMs are not curated as `therapeutic_agent`**: the guideline lists what is "frequently used" rather than recommending any, and populating the slot would imply a level of endorsement the source does not give. The do-not-withdraw warning (Rec 3.2) is recorded.

## 3. The infection branch was disconnected

Correct, and it was the more consequential half of that node. The 14q32 snippet says "Behavior disorders **and susceptibility to infections** can be assigned to the 14q32 region" and I had wired only the behaviour. Second `downstream` edge added to `phenotypes#Recurrent infections` with its own evidence.

## 4. Infection as leading cause of adult death

Added as `clinical_burden` at `HIGH`. The rationale separates the two components deliberately, because they are different in kind: **the epilepsy drives disability and cognitive outcome; infection drives mortality.** An entry that modelled only the epilepsy would misrepresent what actually kills these patients, which is exactly your point.

## Suggestions

**Fever** is now an `environmental:` entry linked to `Cortical Hyperexcitability` with `environmental_effect: EXACERBATES` rather than `TRIGGERS` — fever lowers the threshold of an already seizure-prone cortex rather than creating the susceptibility. `exposure_term` is deliberately unbound with a `review_notes` waiver recording the ECTO search: its heat terms are external heat sources (self-heating coal-mine waste and similar), not a raised body temperature from the patient's own illness, and `HP:0001945` is a phenotype term not valid in that slot.

**NCT06813469 (MD-RING)** added. Good catch — its registered aim is an explicit test of the position-effect hypothesis this graph is built on, which makes it the live route to resolving the entry's central knowledge gap.

**Vigabatrin retinal toxicity: not curated.** The guideline lists vigabatrin among drugs in use but says nothing about retinal risk in this syndrome. The interaction is real in general neurology, but asserting it here needs a source about ring 14 patients, and I would be reasoning from "this syndrome has retinal pathology" plus "this drug has retinal toxicity" to a claim neither source makes.

## Two defects I introduced and caught

Recording both, because one of them is the failure mode `CLAUDE.md` warns about and the other is a gap between two checks.

**Duplicate `evidence` key.** Adding the infection edge inserted it *between* the behavioural edge's description and that edge's existing evidence block, so the evidence orphaned onto the wrong edge and the key duplicated. `just validate` passed under PyYAML; the ruamel-backed reference validator aborted. `check-duplicate-keys` found it.

**Hyphen-split snippets.** Two guideline quotes spanned hyphenated line breaks in the cached PDF text (`prog-\nnostic`, `exami-\nnations`). `just count-verified-snippets` accepted them with a "cache-defect normalization" note; `just validate-disorders` rejected them as not-substring. Worth knowing the fast checker is more permissive than the gate here. Both replaced with spans that avoid the hyphenation.

Neither was committed.

## Validation

`just validate-disorders` passes. Snippets **58/58**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-environmental-evidence`, `check-snippet-length` all clean. Compliance **85.4% → 87.6%**.

gingin77:
All four fixed. Two of them were substantive and I want to name what went wrong in each.

## 1. `GO:0031040 micronucleus` — a false friend, and I should have read the definition

You are right and the term is worse than merely imprecise. I checked it with `runoak -O obo`:

> "A membrane-bounded organelle of ciliated protozoan cells that contains a diploid copy of the cell's complete genome."

and it carries `relationship: RO:0002162 NCBITaxon:5878 ! Ciliophora` — an explicit taxon restriction to ciliates. `GO:0032125` is the same concept. So this is not a loose fit; it is a different organelle in a different kingdom.

The failure mode is worth recording: the term **resolved and passed `just validate-terms`**, because that check confirms a CURIE exists and is named consistently. It cannot tell you the term means something else. Binding dropped, `preferred_term: micronucleus` kept as free text with the reasoning and an NTR note in `description`.

## 2. The `notes:` rationale was factually wrong

This is the one I most needed catching. I wrote that no cached source names a 14q32.33 candidate — and the consensus guideline, which this entry cites **thirteen times**, says exactly that:

> "the distal region of 14q contains the genes for heavy chains of antibodies"

Rewritten to say what is true. The sources *do* propose IGH; the entry declines to promote it into a named-gene mechanism node because the guideline's argument is **positional** — the locus lies in the deleted interval and its content would explain the phenotype — and no study has measured immunoglobulin levels, class-switch competence or IGH copy number in a ring 14 patient. The restraint survives; the stated reason for it did not.

The guideline quote is now `PARTIAL` evidence on the infection edge, where it belongs, and that edge's description no longer says "no gene has been named."

## 3. Guideline section 4

Both Grade A recommendations curated: respiratory function monitoring with advance care planning, and the combined antibiotic prophylaxis / respiratory physiotherapy / immunological evaluation for recurrent pulmonary infection.

The three-in-one is deliberate — the guideline recommends them as a single conditional response, not as independent options — with `treatment_term` bound to the antibiotic component, the only one with a reachable NCIT clinical-action term. Worth noting the immunological evaluation is simultaneously management **and** the missing evidence for the IGH mechanism in item 2; if anyone ever did it systematically, that gap would close.

You are right that this was the same under-mining pattern one section over. Section 4 mattered more precisely because `clinical_burden` argues infection mortality dominates.

## 4. `locations: cerebral cortex` on the formation node

Dropped. A constitutional lesion present in every cell is not localised to cortex, and the cortical localisation already sits on the downstream `Cortical Hyperexcitability` node where the claim is actually made.

## On the fundus snippet

Left as-is. Extending it across the `exami-/nations` hyphenation is exactly what `just validate-disorders` rejected earlier in this PR — `count-verified-snippets` normalizes the split, the authoritative gate does not. The short span is the price of quoting a hyphenated PDF exactly, and I would rather have a truncated exact quote than a reconstructed one.

## Validation

`just validate-disorders` passes. Snippets **61/61**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading` all clean. Compliance 87.6%.

gingin77:
Both fixed. The first one is the third time in this session I have asserted a negative about the literature that turned out to be false, and that pattern is worth naming rather than just patching.

## 1. IgG levels have been measured, and they are normal

I wrote "no study has measured immunoglobulin levels or IGH copy number in ring 14 patients." `PMID:42352229` measured them. Thank you for fetching it to check whether the deep-research report had hallucinated it rather than assuming either way — it is real, and the cache has full text.

**The correction is partial, which matters.** IGH *copy number* genuinely has not been measured, so that half of the sentence survives. IgG *levels* have, and they are normal — which does not merely fill a gap, it argues against the positional IGH story the entry was leaning on. A dosage lesion at the heavy-chain locus predicts a quantitative deficit; what was found is normal concentration with altered glycans.

Both sentences corrected, and the entry now carries both accounts with the tension stated rather than resolved.

**The finding is added, not just the correction.** A `HYPOTHETICAL` node records the glycosylation defect: decreased galactosylation and sialylation resembling pro-inflammatory autoimmune patterns, selective for IgG and absent from total serum glycoproteins. And a `KNOWLEDGE_GAP` records what makes it genuinely interesting — no galactosyltransferase or sialyltransferase gene is on chromosome 14, so the effect acts **in trans** by a route nobody has identified. That is on the branch the entry itself calls most consequential.

**IVIG not curated**, per your explicit recommendation and the paper's own disclaimer that it "should not be interpreted as evidence of therapeutic efficacy." I recorded the reasoning in the discussion `notes:` so a later curator does not read the omission as an oversight and add it.

## 2. Missing `reference_title`

Added from the cache. Worth noting you found this by walking the parsed YAML rather than reading — one item out of 61 is not something eye inspection reliably catches, and I should be checking it programmatically rather than trusting that I added it each time.

## The pattern

Three times now — the IGH rationale in round 2, and this — I have written "no source says X" and been wrong. The failure mode is consistent: I check what my *cited* sources say, then generalise to what the literature contains. Going forward in this entry I have phrased the surviving negative as a specific measurement not made (IGH copy number), not as an absence of study.

## Separately: a CI failure on my other PR

Flagging because it is process rather than content. `check-folded-hyphens` failed on CI for #10007 — a folded scalar line ending in a hyphen. I had run that check on this entry and two others in the batch and skipped it on that one. I have now run it across all four; only that file was affected.

## Validation

`just validate-disorders` passes. Snippets **66/66** (up from 61). `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens` all clean. Compliance **87.6% → 88.6%**.

gingin77:
Rebased onto current `main` (33 commits of drift) and it caught a real, silent defect. Pushed as `620b77f4f6`.

`tests/test_causal_targets.py` did not exist when this branch was cut, and it failed on **two** targets here. Both edges into `Recurrent infections` were written as `target: phenotypes#Recurrent infections` — one from `Terminal 14q32 Haploinsufficiency`, one from `Aberrant IgG N-Glycosylation`.

`downstream.target` holds a **bare** node name matched verbatim against another item's `name`. There is no resolution step and no error when it fails: the entry validated, the page rendered, and `check-entity-refs` passed it, because the prefixed form *is* valid entity-reference syntax in `attaches_to`. What actually happens is worse than a missing arrow — the renderer draws a phantom duplicate node labelled with the literal `phenotypes#Recurrent infections` string, and the real node drops out of the graph.

The part it disconnected is the IgG glycosylation arm, which is the whole reason this entry models two mechanism branches rather than one. Both stripped to bare names.

Same defect turned up on the sibling HCI branch (#10010) in the same rebase, eight targets there. It is a prefix I reached for consistently across a batch, so it is worth knowing the two conventions sit adjacent and only one of them errors when you get it wrong.

`check_causal_targets` exits 0; `tests/test_data.py` + `tests/test_causal_targets.py` **47,206 passed**; snippets still **71/71**. `behind_by` is now 0.

gingin77:

PR #9855, opened 2026-08-28
Stub queue lump/split review: fold type-numbered stub series into single disease concepts

cmungall:
The stub queue carried many OMIM-style numbered series ("primary ciliary dyskinesia 1…53", "Usher syndrome type 1B…3B", "FAME1…7") where the numbered forms are gene-specific variants of one clinical disorder, not separate curation targets. This PR collapses the clear cases: measured against current main, 181 numbered stubs go away, replaced by 13 lumped stubs plus notes on 4 existing parent stubs — a net reduction of 168 queue entries without losing any disease concept. No `kb/` content is touched, and no stub deleted here is claimed by an open `claim` issue. Two stubs were deleted by earlier revisions of this PR and then restored byte-for-byte when claim issues were filed for them mid-review: PCD 30 (claim [#9902](https://github.com/monarch-initiative/dismech/issues/9902), since curated on main via [PR #9924](https://github.com/monarch-initiative/dismech/pull/9924) as a standalone entry — the restore did its job and that stub is now gone from main itself) and cone-rod dystrophy and hearing loss 2 (claim [#9981](https://github.com/monarch-initiative/dismech/issues/9981), still open — the claimant's `entry_type` decision takes precedence).

Three dispositions, recorded per family below:

**Disposition 1 — the lumped disease is already curated in `kb/disorders/` (143 deletions, decision `SUBTYPE`).** These numbered stubs were invisible to `tidy-stubs` because only the parent MONDO ID is in the coverage index; the types are `has_subtypes`-level concepts under an existing entry:

| Curated entry | Numbered stubs deleted |
|---|---|
| Primary_Ciliary_Dyskinesia | 48 (CILD numbered forms across the two filename spellings, including the suffixed 36 X-linked and 48/49 without situs inversus; PCD 30 was deferred to claim #9902 and has since been curated as a standalone entry via #9924) |
| Usher_Syndrome | 16 (types 1B–3B; entry already models types + digenic forms) |
| Ehlers-Danlos_Syndrome | 8 (classic 1/2, classic-like 2/3, kyphoscoliotic 1/2, periodontal 1/2 — the entry has the 2017 types as subtypes) |
| Bartter_Syndrome | 5 (types 1–4B) |
| Perrault_Syndrome | 5 (types 1–6) |
| Congenital_Hypothyroidism | 4 (thyroid dyshormonogenesis 1/2A/3/6) |
| Diabetes_Mellitus | 4 (permanent/transient neonatal 2/3) |
[...]

PR #10007, opened 2026-08-29
curate: Add Polycystic Kidney Disease 3 (MONDO:0010916, GANAB)

### LINKED ISSUE #9990: Curate polycystic kidney disease 3 with or without polycystic liver disease (MONDO:0010916)
Curate a dismech entry for **polycystic kidney disease 3 with or without polycystic liver disease** ([MONDO:0010916](https://monarchinitiative.org/MONDO:0010916)) — the *GANAB* form of autosomal dominant polycystic kidney disease, characterised by mild, often non-progressive kidney cystic disease alongside frequently dominant polycystic liver involvement.

There is a live boundary question here and it should be settled explicitly rather than assumed. `Autosomal_Dominant_Polycystic_Kidney_Disease.yaml` already carries GANAB in its `genetic:` block as an ADPKD-spectrum causative gene, and `Polycystic_Kidney_Disease.yaml` also mentions it — but the KB precedent cuts the other way, since `Polycystic_Kidney_Disease_2.yaml` exists as its own entry. PR #9855, which folded many numbered stub series into their parents, deliberately left the polycystic kidney disease numbered family split. So the options are a standalone entry with a `skos:narrowMatch`-free direct MONDO binding, or a `has_subtypes` entry on ADPKD; whichever is chosen, the ADPKD `genetic:` GANAB block needs to end up consistent with it rather than duplicating it.

The mechanistic case for a distinct pathograph: GANAB encodes the alpha subunit of glucosidase II, so the lesion is in ER glycoprotein quality control, not in the polycystin complex itself. Polycystin-1 fails to mature and traffic to the primary cilium — a defect one step upstream of the PKD1/PKD2 lesion, and one that also explains why the liver phenotype outweighs the kidney phenotype. That upstream-of-polycystin position, plus the mild renal course, is what would justify separate curation.

Stub: `stubs/Polycystic_Kidney_Disease_3_With_Or_Without_Polycystic_Liver_Disease.yaml`

gingin77:
Polycystic kidney disease 3 is the GANAB form of autosomal dominant polycystic kidney disease, and its clinical signature is an **inversion of the usual ADPKD picture**: kidney disease is mild and often never reaches dialysis, while liver involvement is variable and can be severe. Some families present as polycystic *liver* disease with barely any kidney cysts — so one gene sits on both sides of a boundary that used to separate two different disorders.

The mechanism explains the inversion. GANAB encodes glucosidase II alpha, an enzyme in the endoplasmic reticulum that does quality control on newly made proteins. **It is not part of the polycystin complex.** What it does is let polycystin-1 fold correctly and reach the cell surface and the cilium — so the lesion sits one step *upstream* of the PKD1 and PKD2 lesions rather than alongside them. Cysts form because the cell cannot make enough properly matured polycystin-1, not because the channel is broken.

## Why a standalone entry

`Polycystic_Kidney_Disease_2` is the precedent — a gene-specific ADPKD form with its own MONDO term and its own entry — and PR #9855 deliberately left this numbered family split. Two things justify it beyond precedent: the lesion is mechanistically upstream of the polycystin complex rather than within it, and the clinical trajectory is inverted. The parent ADPKD entry already lists GANAB in its `genetic` block; that record and this entry should be read together and nothing here contradicts it.

## Two contrarian findings, both curated rather than omitted

The deep-research cross-check turned up two things that cut *against* the entry's own story. Both are in.

**The mouse model fails, and its authors question causation.** `Ganab+/-` mice develop no kidney or liver cysts. Homozygous deletion is embryonic lethal. And critically, polycystin-1 and polycystin-2 expression is **unaffected** in the heterozygous kidney — which is precisely the readout that is *reduced* in human GANAB+/- cells and on which this entry's central mechanism rests. The paper concludes GANAB "may not be the causative gene."

Curated as `relationship: FAILS_TO_RECAPITULATE`, `fidelity: LOW`, with two `REFUTE`-graded readouts and required `limitations`, plus a `HUMAN_MODEL_MISMATCH` discussion laying out the candidate reconciliations — mature glycoform versus total expression are different quantities; mouse ADPKD models generally need a second hit; species differences in glucosidase II dependence are plausible given the human phenotype is itself mild. None has been tested. The disagreement is between two experiments, not between a gene and a phenotype, and the entry says so.

**The mildness claim is formally contested.** A reported 18-year-old GANAB carrier fell into Mayo imaging class 1E — the class predicting rapid progression to ESRD — and the case report was written specifically to question the mildness assumption. The renal node now carries that case as `PARTIAL` counter-evidence, with a knowledge gap recording that mildness is *modal, not guaranteed*. That matters concretely: mildness is what governs surveillance intervals and whether disease-modifying therapy is considered at all.

## GeneReviews, and what I did not take from it

A GeneReviews chapter exists (`PMID:20301424`) and is tagged. But it is written for ADPKD as a whole and names GANAB only among the less common associated genes, so its Clinical Characteristics section describes the **PKD1-dominated** picture: ESKD in about half of patients by 60, intracranial aneurysms, the wider extrarenal spectrum.

**None of that is imported as a GANAB phenotype.** Only features the GANAB literature itself reports are curated as phenotypes here. The chapter is used for management, surveillance and the agents-to-avoid list — where it genuinely does apply to a GANAB carrier who has cysts — and every such item is graded `PARTIAL` / `OTHER` with an explanation saying it is ADPKD-general guidance rather than GANAB-specific evidence.

**Intracranial aneurysm is deliberately absent**, and recorded as a knowledge gap instead. Importing it would manufacture a GANAB phenotype from a statement about a different genotype. But absence of evidence in ~20 published carriers is not evidence of absence, and the gap says that too, so nobody reads this entry as ruling the risk out.

The one place the distinction bites clinically is tolvaptan: it is approved for *rapidly progressive* disease, which is characteristically not this patient. It is curated as a conditional rather than a recommendation, with the condition being the individual's demonstrated rate of progression rather than the genotype.

## A defect caught before commit

`just check-snippet-grading` flagged a real one: the sentence "Overall, we show that GANAB mutations cause ADPKD and ADPLD and that the cystogenesis is most likely driven by defects in PC1 maturation" was quoted whole in a causal-edge item graded `IN_VITRO` while its opening clause was quoted in `inheritance` graded `HUMAN_CLINICAL`. The sentence genuinely contains both a human-genetic claim and an in-vitro one. Fixed by re-quoting the causal edge with the cystogenesis clause alone, so each item cites its own claim.

Also caught: `GO:0051604` protein maturation placed in `molecular_functions` where it is not an enum member, and transliterated Greek letters (`GIIa` for `GIIα`) in the source title and three snippets, rejected by the title check and the reference validator.

## Validation

`just validate-disorders` passes. Snippets **26/26** verified. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets` all clean. Compliance **95.9%**.

Cache diffs are pure additions. One file, `references_cache/PMID_29038287.md`, showed as modified from a newer fetcher version (quoting style, trailing newline); since the entry does not cite it, I reverted it rather than committing unrelated churn. Stub deleted, history record included.

Closes #9990

### DIFF
--- a/kb/disorders/Polycystic_Kidney_Disease_3.yaml
+++ b/kb/disorders/Polycystic_Kidney_Disease_3.yaml
@@ -542,8 +542,8 @@ prevalence:
   prevalence_class: ULTRA_RARE
   notes: >-
     Twenty affected individuals in nine families at the time of gene discovery,
-    found by screening 321 genetically unresolved families plus six exome-
-    sequenced families. This is a share of the unresolved remainder, not a
+    found by screening 321 genetically unresolved families plus six families
+    that had undergone exome sequencing. This is a share of the unresolved remainder, not a
     population rate: 7 to 10 percent of ADPKD families and about half of ADPLD
     families were genetically unresolved before this gene was identified.
   evidence:


gingin77:
All three fixed. The first was the one worth catching, because it was an internal inconsistency rather than a gap.

## 1. No liver-directed treatment

You are right, and it is worse than an omission: the entry argues in its description, its notes and its tolvaptan rationale that hepatic burden is the half that dominates and that the ADPKD trajectory gives "unwarranted reassurance about the liver" — and then curated nothing that treats the liver. The entry contradicted itself.

Added, from the GeneReviews text already in the cache: oestrogen avoidance and acid suppression first, escalating to percutaneous aspiration and sclerosis, laparoscopic fenestration, combined resection with fenestration, hepatic artery embolisation, and liver transplantation. Graded `PARTIAL`/`OTHER` like the other GR items, with the note that this one applies with **more** force here than the renal recommendations do, since the liver is the compartment GANAB carriers are most likely to need it in.

**Somatostatin analogues left out.** The research artifact names them; the cached GeneReviews text does not, and no GANAB-specific source supports them. I would rather cite what I have than reach for the artifact.

## 2. Genetic counselling section

Your scoping point is right and I had missed it: the entry's deliberate exclusion of GeneReviews *Clinical Characteristics* does not extend to *Genetic Counseling*, because that section names GANAB explicitly and none of its content is PKD1-specific.

Added to `inheritance`: the 50% transmission risk and the 10–20% de novo rate, both `OTHER`, with explanations stating why this material is in scope where Clinical Characteristics is not — so the distinction is visible to the next curator rather than looking arbitrary.

One implementation note: `Inheritance` has no `notes` slot, so my first attempt to record the prenatal/PGT availability there failed validation. It is folded into the block `description` instead.

## 3. Testing modality

Now specified: a multigene cystic kidney and liver panel is the primary test, and single-gene GANAB testing is **cascade-only** — reaching GANAB as a hypothesis already presupposes PKD1 and PKD2 exclusion, so it cannot be the first test.

Deletion/duplication analysis is required, and I fetched and verified `PMID:32405593` before quoting it as you asked. It supports the point directly and explains the mechanism of the miss:

> "Because of its 75−base pair read length, WES is not equipped to detect large deletions."

Added a `results:` field stating what a positive test looks like.

## Suggestions

**Taken:** the `HUMAN_MODEL_MISMATCH` discussion now anchors on the model's `name` rather than `animal_models#Mouse`. You are right that the name is the stable pathograph label — `species` resolves, but it would collide the moment a second mouse model is added.

**Not taken this round, with reasons:**

- *Hypertension as a phenotype* — no GANAB-specific source in the cache. It is a real ADPKD feature, but curating it here would be the same import this entry declines for intracranial aneurysm, and I would rather be consistent about that boundary than selectively porous.
- *Connecting the two `INDIRECT_KNOWN_INTERMEDIATES` edges to the parent's intermediates* — the parent ADPKD entry models the cAMP/chloride-secretion chain properly. Duplicating those nodes here is what `CLAUDE.md` warns against ("not DRY" applies to module conformance, not to copying a sibling entry's pathograph). If the intent is cross-entry reference, that is a schema question rather than a change I should make unilaterally.
- *An ER protein-biogenesis module* covering GANAB/PRKCSH/SEC63/ALG8/SEC61B — genuinely good, and the JCI paper cited here makes exactly that argument. But it is module work with its own skill and its own PR, not a change to this entry.

## Validation

`just validate-disorders` passes. Snippets **30/30**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading` all clean. Compliance **95.9% → 96.3%**.

PR #10010, opened 2026-08-29
curate: Add Hyperostosis Cranialis Interna (MONDO:0007765, SLC39A14)

### LINKED ISSUE #9987: Curate hyperostosis cranialis interna (MONDO:0007765)
Curate a dismech entry for **hyperostosis cranialis interna** ([MONDO:0007765](https://monarchinitiative.org/MONDO:0007765)) — a rare autosomal dominant bone disorder in which progressive hyperostosis and osteosclerosis of the calvaria and skull base narrow the cranial foramina, producing sequential entrapment of cranial nerves I, II, VII and VIII while the rest of the skeleton is spared.

The stub records `mondo_parents: hereditary disease` only, with no descendants and no causal gene — consistent with a leaf term whose molecular lesion is not settled in MONDO. Curation should be honest about that: the phenotype and its natural history are well described in the reported kindreds, but the gene assignment is the open question and belongs in a knowledge-gap discussion rather than an over-confident pathophysiology node. The mechanistic contrast worth drawing is against the sclerosing bone dysplasias that *do* have a settled lesion (sclerosteosis/van Buchem, SOST), since cranial-nerve entrapment from skull-base overgrowth is the shared adverse outcome.

Stub: `stubs/Hyperostosis_Cranialis_Interna.yaml`

gingin77:
Hyperostosis cranialis interna is a rare inherited bone disease in which the skull thickens from the inside. As the bone grows inward it squeezes the small openings that cranial nerves pass through, so patients lose their sense of smell, develop facial weakness, hearing loss and balance problems, usually starting in their teens. **It is the only genetic bone dysplasia confined to the skull** — the rest of the skeleton is completely normal, which is what separates it from the sclerosing dysplasias it otherwise resembles.

## The stub said there was no gene. There is.

The curation stub carried no causal gene, which is why my claim issue described the molecular lesion as the open question. That turned out to be wrong, and the reason is worth recording: `MONDO:0007765` carries a feature relation and a parent class and **no causal-gene relation**, so stub enrichment had nothing to copy — while *SLC39A14* was identified by whole-exome sequencing in 2018. The literature was ahead of the ontology.

I've recorded that as an observation in the entry's `notes:` for whoever maintains the mapping, without proposing what should be done about it.

## The mechanism

Unusually complete for a disease this rare:

```
SLC39A14 L441R substitution                        (MOLECULAR)
  └─> ZIP14 mistrafficking + intracellular zinc     (CELLULAR)
        accumulation
        └─> cAMP-CREB and NFAT hyperactivation      (CELLULAR)
              └─> Enhanced endosteal bone formation (CELLULAR)
                    └─> Calvarial / skull-base      (TISSUE)
                          hyperostosis
                          └─> Cranial nerve         (ORGANISM)
                                entrapment
```

ZIP14 is a zinc transporter that normally sits in the plasma membrane. The mutant does not get there, and instead accumulates zinc inside the cell — so the defect is one of *location*, not transport capacity. `functional_impact_category` is deliberately `UNKNOWN`: mistrafficking with downstream signalling gain is neither a clean loss nor a clean gain, and no source classifies it.

## Named Entity Confusion — and I changed my mind

I noticed while reading that a 2025 CSF-leak case report describes the aetiology as "still unknown" (superseded in 2018) and discusses the condition alongside hyperostosis frontalis interna. I curated the CSF leak as a phenotype with a heavy caveat.

The research report then flagged the same hazard independently, and more strictly: the phrase names **two incompatible things** — this Mendelian disease, and a loose radiological label for sporadic internal skull-table thickening overlapping HFI and Morgagni-Stewart-Morel syndrome. Sense-(B) case reports, it said, must not support phenotype claims here.

I adopted the stricter reading and **removed the phenotype entirely.** A caveated phenotype still asserts that this is a feature of the disease, and that paper reports no genetic testing, so whether the patient had this entity is unknown. The observation may well be real — a skull-base CSF leak is plausible in skull-base hyperostosis — but establishing it from a second-sense case report *is* the confusion. The reasoning is in `notes:` and the mimics are curated as `differential_diagnoses` rather than dropped.

## The mouse reproduces the mechanism but not the disease

`PARTIALLY_RECAPITULATES`, with three stated limitations. It overexpresses rather than carrying the allele at endogenous dose; its phenotype is generalised skeletal rather than confined to the skull; and it *adds* a trabecular osteoporosis patients are not described as having.

That middle point matters: cranial confinement is the single most distinctive feature of the human disease, and it is exactly what the model fails to reproduce. Recorded as `HUMAN_MODEL_MISMATCH`, with untested candidate explanations — regional differences between neural-crest-derived cranial osteoblasts and mesoderm-derived axial ones, a kindred modifier, or an artefact of overexpression.

## No frequencies, anywhere

Essentially the entire clinical description comes from **three related Dutch families with common progenitors** — 32 individuals, five generations, 12–13 affected. Every natural-history statement inherits that, so no frequency band is asserted in this entry and a knowledge gap records that a single kindred cannot separate features of the disease from features of the family.

## Research artifact handling

39/39 references verified, 0.0 confabulation. The report flagged one unsupported quote, three off-topic references, and one obsolete term (`GO:0019933` cAMP-mediated signaling) — **none of the five is used here.** The cAMP node was already bound to `GO:0141156`, the current term, before I read the report; I had checked it with `runoak` and found the obsolete one first.

## Validation

`just validate-disorders` passes. Snippets **36/36** verified. `validate-terms`, `check-entity-refs`, `check-duplicate-keys` all clean. Compliance 95.0%. Cache diffs are pure additions. Stub deleted, history record included.

Closes #9987


gingin77:
All five fixed. Two of them taught me something.

## 1. `GO:0021545 cranial nerve development` — the second false friend in this batch

You are right, and the way it is wrong matters: it names a *developmental* process. In this disease the nerves develop normally and are mechanically crushed in the second decade. The annotation asserted a neurodevelopmental disorder that this is not.

This is the second term of exactly this shape across my open PRs — the other was `GO:0031040 micronucleus` on #10005, which turned out to be a ciliate organelle. **Both passed `just validate-terms`**, because that check confirms a CURIE resolves and is named consistently; it cannot tell you the term means the wrong thing. I have started reading definitions rather than labels on any binding where the label alone sounds right.

Dropped, with the reasoning in `notes:`.

## 2. CN V asserted mechanistically, dropped clinically

Fixed, and it was zero-cost as you said — the quote was already in the cache. Facial hypoesthesia added, bound to the general `HP:0033748 Hypoesthesia`, with the anatomical restriction carried in the name and description rather than by binding something narrower than HPO actually offers.

## 3. Headache — and the distinction you drew

This is the one I want to acknowledge properly. **Omitting the frequency band and omitting the phenotype are different decisions**, and I had collapsed them. The no-frequency reasoning was about a single kindred; it says nothing about whether the phenotype belongs in the entry.

Added, `HP:0002315`, HPOA provenance in `notes:`, no frequency band — consistent with the rest of the entry and for the reason that actually applies.

## 4. PMID:23622937 — the paper that constrains my own chain

The best catch here. Preserved otoacoustic emissions in **every** patient localise the lesion to the nerve rather than the cochlea, which is direct positive evidence for the entrapment → hearing loss edge I had already drawn and left unsupported. Added to the hearing phenotype.

The monitoring panel is now a `diagnosis:` entry. You are right that this was a real gap: the entry's whole treatment argument is "decompress before the nerve is lost", and it said nothing about the surveillance that tells you when. It includes the BERA I-V interpeak latency as the measure that determines encroachment reliably, **and the negative recommendation** — blink and masseter reflexes are explicitly not advised, since bilateral symmetry drops their sensitivity to 25–37.5%. A test not worth doing is as useful to record as one that is.

The NCIT binding for the composite panel is imperfect and I flagged it in `notes:` rather than hiding it — `Electroencephalography` is the nearest reachable term for evoked-potential recording and is not really what this is.

## 5. `progression:` added — with one deliberate omission

Three phases: presymptomatic radiological, early symptomatic in the second decade, established deficit.

**I did not curate the post-fourth-decade plateau**, even though you are right that it is the clinically decisive part. No cached source states it — it appears in the research report only. Asserting a phase boundary on that basis would be the same failure mode I have already been caught on three times in this batch: treating what a report says as what the literature shows. It is recorded in `review_notes:` so it can be added the moment a source turns up.

## Suggestion taken

HGVS and ClinVar detail added: `NM_001128431.4:c.1322T>G`, `p.Leu441Arg`, exon 8, `VCV000523143` Pathogenic — including the review status **"no assertion criteria provided"**, which as you say is the honest weak spot and is better stated than left for a reader to find.

## Validation

`just validate-disorders` passes. Snippets **42/42** (up from 36). `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens` all clean.

One number to explain: global compliance reads 93.9%, down from 95.0%. That is the new `progression:` and `diagnosis:` entries adding slots faster than evidence-bearing ones — the same coverage-versus-density trade as on #9996. Evidence coverage itself went up.

gingin77:
All four fixed. Finding 2 is the one I want to address squarely, because my "correction" last round introduced a new false claim.

## 2. I rejected the right term for a false reason

I dropped `HP:0031912 Trigeminal anesthesia` on the ground that *"anesthesia denotes complete sensory loss."* Its HPO definition is **"Decreased or absent sensation in the distribution of the trigeminal nerve,"** and it names the face and mouth.

I read the label. Again. That is precisely the habit the `GO:0021545` finding was about, and precisely what my last comment said I had changed — and I did it in the very commit that was correcting a different label-derived error. Rebound to `HP:0031912`, which is the most specific term that accurately represents the claim, and the note now records the definition check rather than an argument about severity.

## 1. A third false "no term exists" — and it was already on the board

You are right that this was carried over twice and that my audit stopped at the section boundary. I searched `phenotypes:` notes and did not extend the same method one section down to `diagnosis:`.

The `NCIT:C38054` binding was not inert, which is the part I had underweighted: that term means an actual EEG in all 22 other entries using it, so HCI would have been the sole outlier in any "diagnosed by EEG" query. Panel split into three entries bound to `NCIT:C38036` and `NCIT:C85857`, both already cached and enum-valid, with `results:` distributed so each statistic sits on the test it came from.

One weak binding remains and is flagged in place: the vestibular/VEP entry uses an auditory term because a search for electronystagmography and visual-evoked-potential action terms came up empty. That note says *"a search that came up empty rather than an established absence"* — which is how all three of these should have been written in the first place.

## 3. Headache

`just fetch-reference ORPHA:443098` fails here — the Orphanet bulk XML is not in this checkout, and refreshing it would add manifest churn unrelated to this entry. So I took your stated alternative and **dropped the mechanism** from the description. The Orphanet route is recorded in `notes:` for whoever next refreshes that source.

## 4. The history record asserted content that was never there

This is the one that would have outlived the thread, and you were right to keep raising it.

The cause: my edit script's `replace()` silently did not match, and I never verified the result before writing the history record claiming it had. Since history is append-only provenance, that would have become a permanent false statement about the entry.

Fixed by **adding the detail** rather than deleting the claim — it is good content, and `NM_001128431.4:c.1322T>G`, `p.Leu441Arg`, exon 8, ClinVar `VCV000523143` with review status "no assertion criteria provided" is now in `genetic[0].features`. The record is true.

## Suggestions

**The smell rebind was the wrong direction and I have redone it.** You are right that Hyposmia matches neither source — the HPOA set carries `HP:0000458` Anosmia at 6/13, and the cited sentence says "disturbance of the sense of smell." Now bound to `HP:0004408 Abnormality of the sense of smell`, whose definition is "an anomaly in the ability to perceive and distinguish scents" — the exact match for the quote. The tension you identified, that this entry treats HPOA as authoritative for Headache and not for smell, is now stated in `notes:` rather than left implicit.

Also done: VEP sensitivity attribution corrected (VEPs are inside the 25–37.5% range and still recommended); `modifier: ABNORMAL` dropped from `GO:0005385`, since the node argues mislocalisation rather than altered capacity; and process narration moved out of `notes:` into the history records, per your point that `notes:` is read by someone asking about the disease, not the PR.

Not done: the ABI vasospasm and 18F-fluoride PET references still need fetching and verifying.

## Also in this push

Rebased onto main (120 commits) and migrated five `PARTIAL` items to the new `supports`/`directness` schema from #10003.

`just validate-disorders` passes. Snippets **43/43**. `validate-terms`, `check-entity-refs`, `check-folded-hyphens` clean.

gingin77:
Both fixed, plus four of the five suggestions.

## 1. `NCIT:C215111` — thank you for running the search

You confirmed both halves of what I could only assert: NCIT genuinely has nothing for electronystagmography, vestibular function testing, or general visual evoked potentials — *and* `NCIT:C215111 Neuroelectrophysiology Test` exists, is accurate for both tests, and is reachable from `C25218`.

Rebound, with `preferred_term` carrying the specificity. You were right that the previous state was worse than imprecise: `NCIT:C85857` is legitimately used on the BERA entry thirty lines above, so the file had one identifier standing for two different tests. `notes:` rewritten, since it argued for a binding I am no longer making.

## 2. The stale prose — and why my cleanup missed it

Fixed, and the cause is worth recording because it is a repeatable trap.

My cleanup grepped for `"Graded PARTIAL"` **on a single line**. In a folded scalar that phrase wraps, so `Graded` ended one line and `PARTIAL` began the next, and the grep returned zero. I read that zero as "clean."

Re-searching for the bare token `PARTIAL` found this one **and one more in #10005**, now also fixed. Two entries, same blind spot. You are right that no gate catches prose naming a retired enum value — and a line-oriented grep is not a substitute for one.

## Suggestions

**Both uncited statistics now carry exact quotes** — the 60–70% hearing-threshold/BERA figure and the 50% vestibular figure were sitting in `results:` fields with nothing behind them. Both were quotable as-is.

**Meta-commentary trimmed** from the three explanations. You are right that "is a matter of strength, which the schema has no slot for" is review dialogue, not curation — and with #10234 having just moved that class of prose out of `kb/`, writing more of it was going the wrong way.

**Discussion anchored on the model's name** rather than `species`.

**The allelic collision is in, and I agree it is a safety point rather than trivia.** `SLC39A14` also causes hypermanganesemia with dystonia 2 through biallelic loss of function, and that recessive disease responds to chelation with disodium calcium edetate. The treatment does not transfer: it addresses manganese overload from a transport failure, not a dominant mistrafficking allele causing zinc accumulation. Curated as a `differential_diagnoses` entry with both claims quoted from `PMID:27231142`, and a `notes:` paragraph making the distinction explicit — this is a *gene* collision, cutting the opposite way from the *name* collision already recorded in the entry notes.

**Not chased:** thank you for the correction that `PMID:21665359` and `PMID:21079950` are not in the cache. I would have gone looking.

## Validation

`just validate-disorders` passes. Snippets **47/47** (up from 43). `validate-terms`, `check-entity-refs`, `check-folded-hyphens` clean.

gingin77:
Blocking item fixed, plus five of six suggestions.

## ¹⁸F-fluoride PET/CT

Added. You are right that it is not redundant with CT — it is the only test here that quantifies disease **activity** rather than describing structure, and the study was done in this same Dutch kindred (9 affected, 7 unaffected relatives, 9 unrelated controls). `PMID:21079950` fetched and verified.

Two quotes curated: the significantly higher uptake in sphenoid bone and clivus, and the statement that visual assessment detects severity and the pattern of disturbed metabolism through life. That second one earns its place in *this* entry specifically — the management argument turns on decompressing before a nerve is lost, and this is the instrument that shows whether the process is still advancing.

The NCIT binding names the modality pairing rather than the tracer, since NCIT has no ¹⁸F-fluoride term; `preferred_term` carries the tracer, which is the part that makes it a metabolic rather than structural scan.

## Suggestions

**2 — you caught a real internal inconsistency.** `preferred_term: Surgical Procedure` merely echoed the broader binding, while thirty lines away `NCIT:C215111` correctly carries `electronystagmography and visual evoked potentials`. Now `cranial nerve decompression`.

**5** — `HP:0000648` Optic atrophy added as the structural correlate of CN II involvement, no frequency asserted since HPOA records none.

**6** — a `description` on `GO:0005886` noting it names the compartment the transporter *fails to reach*. You were right that it reads oddly without one.

**3 and 4 — both added as `KNOWLEDGE_GAP`s, and 4 is the more consequential.** Formation-versus-resorption matters because the two mouse models point at *different effector cells*, and if it is a resorption failure this entry names the wrong one; the formation reading is what the human histology supports, which is why it stays curated, but the alternative is now visible.

The metal-levels gap is attached to both the zinc node and the allelic differential deliberately. Zinc, manganese and iron have never been measured in a patient with this disease — while the biallelic form of the same transporter causes measurable manganese accumulation that chelation treats. So whether the chelation boundary this entry draws is a boundary of *mechanism* or of *untested assumption* is genuinely open, and worth a curator knowing.

**1 — recorded in `review_notes:`, not curated.** You are right that saying "treatment is otherwise symptomatic" is thin for a disease whose dominant morbidity is deafness. `PMID:21665359` reports cerebral vasospasm after auditory brainstem implantation *in this syndrome*, but has no retrievable abstract, so no compliant snippet exists and I assert nothing. Logged so the safety signal survives and whoever can obtain the full text knows what to look for.

## Validation

`just validate-disorders` passes. Snippets **49/49** (up from 47). `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-folded-hyphens` clean.

gingin77:
All five fixed. Finding 1 is the one that matters most, and you characterised it exactly right: an anatomical inversion, not a matter of emphasis.

## 1. The mouse spares the calvaria — I had it backwards

I wrote in four places that the model has a "generalised skeletal phenotype rather than confined to the skull," which implies it affects the skull *and more*. The cached full text says the opposite:

> "found that loss of endogenous Zip14 did not affect the calvariae, even though the appendicular skeleton and vertebral column were osteoporotic"

So the model **misses the one bone that defines the disease, and changes the remaining bone in the wrong direction** — osteoporotic, not hyperostotic. And the authors say so plainly: *"truly opposite of what we see in HCI."*

All four passages corrected. The discussion prompt was asking the wrong question — "what confines hyperostosis to the skull in patients" — when the finding poses a sharper one the authors themselves leave open: **why murine calvariae appear protected** from a zinc-handling lesion affecting every other bone in the same animal. That the same sparing appears in the plain null is what rules out the knock-in construct, and it is now cited.

## 2. The mismatch had no structural signal

Right, and it was exactly the state that slot exists to fix — the claim surviving only as prose in `limitations`. Added a second link targeting `Calvarial and Skull Base Hyperostosis`, `FAILS_TO_RECAPITULATE`, `LOW`, with both mandatory fields.

Its evidence is `SUPPORT`, not `REFUTE` — the quotes support the claim that the model *fails*; the negative direction is the relationship's job. (Your parallel review on #10011 made the same point about a readout, and I applied it here from the start.)

I also added link-level evidence to the `PARTIALLY_RECAPITULATES` link, which had none.

## 3. Seven phenotypes were hanging free

Fixed — `Cranial Nerve Entrapment` now has seven `downstream` edges. As you said, the linking sentence was already quoted in the file and supports five directly. `Optic Atrophy` is `INDIRECT_UNKNOWN_INTERMEDIATES`, since nothing traces compression to atrophy in this disease.

## 4. Treatment not joined to the graph

`target_mechanisms` added. Both unused quotes curated — the one-year House-Brackmann outcome, and the second-source statement supporting **prophylactic** decompression in younger patients, which is the timing claim this entry's whole management argument rests on and which had been resting on a single feasibility case.

## 5. Bundled node

Split into `ZIP14 L441R Mistrafficking` (MOLECULAR) and `Intracellular Zinc Accumulation` (CELLULAR). You were right that the scale tag was the tell. The rename broke a discussion foreign key — caught by `check-entity-refs`.

## Suggestion taken

`CL:0000062` added alongside `CL:2000058`. Runx2-Cre is lineage-wide, and calvarial osteoblasts from those mice showed no marker differences, so binding only the calvarial type overstated where the mechanism operates.

The `Optic Atrophy` / `Headache` provenance-without-evidence state is a conscious one, and I am glad you flagged it as such rather than as an oversight.

## Validation

`just validate-disorders` passes. Snippets **56/56** (up from 49). All ungated checks clean.

**I ran the full `tests/test_data.py` lane locally** rather than trusting this PR's CI — your review on #10011 pointed out that pytest is path-gated and a `kb/`-only PR skips it entirely. `test_failure_to_recapitulate_links_are_substantiated` passes, which it would not have if I had added the new failure link without evidence.

gingin77:
Rebased onto current `main` (27 commits of drift) and it immediately caught a real defect I had just introduced. Pushed as `fcfb928bda`.

`tests/test_causal_targets.py` did not exist when this branch was cut. It failed on **eight** targets in this entry, all from the previous commit:

- **Seven** of them are the new `Cranial Nerve Entrapment` edges I added for finding 3. I wrote them as `target: phenotypes#Facial Palsy`. `downstream.target` holds a **bare** node name matched verbatim — so all seven pointed at nothing, while the entry validated and the page rendered. `check-entity-refs` passes them, because the prefixed form *is* valid entity-reference syntax in `attaches_to`. Nothing already on this branch could have caught it; the fix for a disconnected graph left the graph disconnected in a way that looked connected.
- **One** is the node split: `SLC39A14 L441R Substitution` still pointed at the pre-split `ZIP14 Mistrafficking and Intracellular Zinc Accumulation`. Repointed.

`scripts/check_causal_targets.py` now exits 0, and `tests/test_data.py` is **47,189 passed** against current `main`.

Worth stating plainly, since it bears on how much the green badge on this PR meant: the branch had been sitting 27 commits behind, so every check run here — including the ones I cited in the previous comment — was run against a `main` that did not yet have this test. Rebasing before asking for review, rather than after, is the fix, and I have made it a standing habit rather than a note on this PR.

Snippets still **56/56**. The base is now current, so this review round is against the tree that will actually merge.

gingin77:
Round 11 IMPORTANT fixed, both suggestions taken, and fixing it turned up a second error you had not flagged. Pushed as `04ae22570e`, rebased onto current `main`.

## The description

You were right that this is the version most readers see, and it was the one still telling the old story. Rewritten so it says what the file's other four passages now say: the thickening appears in the **long bones**, the calvaria shows no phenotype at all (thickness and porosity not significantly different from controls), the same sparing appears in the plain null so it is not a knock-in artefact, and the open question is why murine calvariae are protected rather than why the human disease is confined.

## The error underneath it

Going back to `PMID:29621230` to write that paragraph, I found that `limitations` and the discussion `rationale` both said the knock-in mouse's **appendicular skeleton and vertebral column are "osteoporotic rather than hyperostotic."**

That is not the knock-in. It is the plain `Zip14-/-` null, cited from reference 14 of the same paper: *"dwarfism and general osteoporosis of the appendicular skeleton and vertebral column ... but normal cortical bone."* The knock-in's long bones carry a **drastically increased cortical thickness**; only the trabecular compartment is osteoporotic. So the file was contradicting its own `PARTIALLY_RECAPITULATES` link, which correctly says the model reproduces the cortical arm.

I introduced this in the round that corrected the original inversion — I pulled the right phenotype from the wrong mouse in the same paragraph of the same paper. Corrected in both places.

## Suggestions

- **`Skull Base Osteosclerosis`** now has an incoming edge from `Calvarial and Skull Base Hyperostosis`, supported by a sentence already quoted in the entry (57/57 snippets, up from 56).
- **Discussion id** renamed to `hci_murine_calvarial_sparing`.
- **`Headache` I have deliberately left with no incoming edge.** Its own `description` says no mechanism is asserted, and the attribution to raised intracranial and ocular pressure appears only in secondary sources. Inventing an edge to close a graph-shape gap would trade a real epistemic position for a cosmetic one. Flagging it rather than silently declining.

## Validation

`check_causal_targets` exits 0. `tests/test_data.py` + `tests/test_causal_targets.py`: **47,199 passed**. Rebased onto current `main`, so `behind_by` is 0 — which also matters for merge eligibility, since the deterministic sweep refuses any PR whose head does not contain the current base.

gingin77:

PR #10011, opened 2026-08-29
curate: Add Diaphyseal Medullary Stenosis with Malignant Fibrous Histiocytoma (MONDO:0007205, MTAP)

### LINKED ISSUE #9989: Curate diaphyseal medullary stenosis-bone malignancy syndrome (MONDO:0007205)
Curate a dismech entry for **diaphyseal medullary stenosis-bone malignancy syndrome** ([MONDO:0007205](https://monarchinitiative.org/MONDO:0007205)), also called Hardcastle syndrome or DMS-MFH — an autosomal dominant syndrome combining a cortical/medullary long-bone dysplasia (diaphyseal medullary stenosis with cortical thickening, bone infarcts, pathological fractures) with a high lifetime risk of malignant fibrous histiocytoma, i.e. undifferentiated pleomorphic sarcoma, arising in the abnormal bone.

This is a **germline predisposition syndrome**, so per the cancer granularity ladder in `docs/explanation/design-decisions.md` §3a it follows the plain Mendelian lump/split rules and stays separate from any somatic sarcoma entry — the same way Li-Fraumeni is kept separate from the cancers it predisposes to. The stub gives a single causal gene, `hgnc:7413 MTAP`, and MONDO parents that read like the whole syndrome in miniature: skeletal dysplasia, bone neoplasm, hereditary connective-tissue disorder, and disorder of polyamine metabolism.

The mechanism worth getting right is that last parent. MTAP is the methylthioadenosine phosphorylase of the methionine-salvage pathway, and germline splicing variants are the reported lesion — so this is a metabolic-salvage defect that produces both a bone-modeling phenotype and a tumour phenotype. Somatic *MTAP* deletion is separately well known as a passenger of 9p21 *CDKN2A* loss in sporadic cancers, which is a Named Entity Confusion trap for reference searching: literature on 9p21 co-deletion is about a different mechanism and should not be cited as evidence for the germline syndrome.

Stub: `stubs/Diaphyseal_Medullary_Stenosis-bone_Malignancy_Syndrome.yaml`

gingin77:
DMS-MFH, also called Hardcastle syndrome, is an inherited condition that combines a long-bone disease with a high risk of bone cancer. The bones develop infarctions and abnormal cortex that narrows the marrow canal, fractures happen and then heal badly, and about a third of affected people develop a bone sarcoma. Muscle disease and early cataracts are part of it too.

**The molecular lesion is the interesting part, and it is instructive well beyond this disease.**

The locus was mapped to 9p21-22 and then resisted identification for more than a decade — because the causal mutations are not in any of the genes that region is known for. They lie in three previously uncharacterised terminal exons of *MTAP*, **two of which entered the primate genome as independent retroviral insertions at least 40 million years ago** and have since acquired a function. The gene was thought to have eight exons. Candidate screening of the annotated genes could not have found this.

## The synonymous mutation

The single most informative fact about this lesion, and it is not in the abstract — I found it in the full text:

> "one mutation was a synonymous change at position c.885A>G"

A synonymous change alters **no amino acid**. So the pathogenic effect cannot be a change to the protein sequence and has to act on the transcript. That is what promotes the splicing mechanism from *a* candidate explanation to *the* explanation, and it is why the entry describes this as a regulatory-isoform lesion rather than an enzyme deficiency.

## Named Entity Confusion — flagged prominently in `notes:`

*MTAP* is far better known for something else entirely. It sits immediately beside *CDKN2A* at 9p21 and is co-deleted as a passenger in a large fraction of sporadic cancers, and MTAP-deleted tumours are an active therapeutic target through PRMT5. **A search on the gene symbol returns overwhelmingly that literature**, which concerns homozygous *somatic* deletion in tumour cells and has nothing to do with this germline splicing syndrome.

I flagged this in the claim issue before starting and it held up. The one genuine connection is real and is curated with its own evidence: loss of heterozygosity across the same 9p21-22 interval in **five of seven** informative sporadic bone MFH specimens, which is what motivated the search for a shared tumour suppressor in the first place.

## Where I nuanced rather than asserted

**Myopathy.** The defining paper calls this "a bone dysplasia, myopathy, and bone cancer syndrome", so it reads as a core feature. The full text says otherwise: it was **absent from the three originally described families** and entered the phenotype when two further families with FSHD-like muscular disease were characterised. Those two families carry *different* mutations, so a simple allele-phenotype correlation does not explain it. The phenotype is kept, with the distribution stated and a knowledge gap recording the unexplained variability.

**The metabolic node is `HYPOTHETICAL`.** MONDO classifies this partly as a disorder of polyamine metabolism, and the inference is reasonable — but no study has measured methylthioadenosine, polyamine or methionine-salvage flux in a patient. That node names what the enzyme does, not what has been shown to go wrong, and says so.

## Three knowledge gaps

- **The route from splicing defect to bone phenotype is entirely unmapped** — no cell type, no pathway, no remodelling mechanism. And there is a structural reason it may stay that way: the causal exons are primate-specific retroviral insertions, **so a mouse has nothing to mutate.**
- **The whole molecular account rests on one paper from one laboratory**, which also did the earlier linkage work. Not a criticism of the finding, which is coherent and explains a decade of failed candidate screening — but it matters more than usual for a claim this unusual, and independent identification of the same exons in a new family has not been reported.
- The myopathy variability above.

## Notes on curation choices

**Nomenclature is deliberately historical.** "Malignant fibrous histiocytoma" is what the primary literature uses throughout, so the disease name and the quoted snippets keep it; the description notes the entity was later reclassified as undifferentiated pleomorphic sarcoma.

**One snippet begins mid-sentence.** The preceding word in the cached PDF is set with an `ﬁ` ligature, which no exact-substring quote can span. The `explanation` records why rather than leaving it looking careless.

Per the granularity ladder in the design-decision register, this is a germline predisposition syndrome and follows the plain Mendelian rules, staying separate from the somatic sarcoma entries it predisposes to.

## Validation

`just validate-disorders` passes. Snippets **23/23** verified. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens` all clean. Compliance 91.2%. Cache diffs are pure additions. Stub deleted, history record included.

Research artifact: 19/19 references verified, 0.0 confabulation, both checked quotes valid, no not-found or obsolete terms. Its four flagged label mismatches are table-formatting artifacts, not wrong bindings.

Closes #9989


gingin77:
All fixed, plus the CI failure. The CRITICAL finding is the one that matters and I want to be plain about it.

## CRITICAL — the MTA measurement

You are right, and this is the **fourth time in this session** I have written "no study has measured X" and been wrong. This instance is the worst of the four: the measurement is in the paper the entry is built on, in a Results section, in the cache file this PR itself commits. I cited that paper eight times and did not read its results.

Serum MTA — MTAP's own substrate — undetectable in all three unaffected individuals, accumulating in both affected ones, blinded assay.

All three consequences applied:

1. **Prose corrected** in both places.
2. **`HYPOTHETICAL` → `ESTABLISHED`.** You put it exactly right: the stated reason for the grade was the absence of any patient measurement, and that premise is gone. Substrate accumulation in vivo is a demonstrated perturbation, not an inference from what the enzyme does. I recorded the n=5 fragility in `notes:` rather than using it to justify a weaker grade — sample size changes how much you know, not what kind of evidence it is.
3. **The finding is now the evidence**, `SUPPORT`/`HUMAN_CLINICAL`, alongside the background item and the baseline sentence that MTA is not normally in serum (which is what makes "detectable" mean "accumulated").

The metabolite-to-bone gap survives intact, and its discussion was rewritten to say accurately what is and is not known.

## The audit you suggested — it found two more

You flagged that `HP:0007819 Presenile cataracts` exists despite my note saying otherwise, and suggested re-checking the other "no term exists" judgements by the same route. I ran that across **all five entries from this session.** Two more were false, both in #10010:

- *"HPO has no graded olfactory term"* — it has `HP:0004409 Hyposmia` with Total and Partial children. Rebound from Anosmia, which also **overstated severity**: the source reports a disturbance of smell, not its abolition. The false ontology claim was hiding a wrong binding.
- *"HPO has no discrete annotation for cranial nerve V sensory loss"* — `HP:0031912 Trigeminal anesthesia` exists. I kept the general `Hypoesthesia`, but on the ground that survives checking: anesthesia is complete loss, the source reports impairment.

The cheek-pad, facial-gestalt and ciliate-micronucleus claims were checked and hold. Fixed in a companion commit on #10010.

Your phrasing — "the correct instinct applied to a wrong fact" — is the accurate diagnosis of all four. The reasoning was sound each time; I just didn't check the premise.

## Remaining items

**Eponymous phenotype split.** `HP:0100254 Stenosis of the medullary cavity of the long bones` and `HP:0005045 Diaphyseal cortical sclerosis`, plus `HP:0002653 Bone pain`. A disease named *diaphyseal medullary stenosis* bound to "abnormal cortical bone morphology" was indeed several levels too coarse.

**Both variants curated** with the functional difference the myopathy gap turns on — c.885A>G cutting exon-9 isoforms ~70% with archetype MTAP untouched, and both variants raising MTAP_v3/_v6 equally, so what they share is an isoform-balance shift rather than a common loss.

**`diagnosis:` and `treatments:` added.** You were right that the entry curated a 35% sarcoma risk and recorded no response to it. Radiographic screening from puberty is in, with thallium scanning distinguished as tumour-directed rather than dysplasia-directed. The exon-9 caveat is the actionable part: the causal exon sits ~65 kb past the canonical stop, so a negative panel is uninformative — and I recorded the variant-numbering trap (AK309365, not NM_002451) in the same place.

**PRMT5 explicitly excluded from treatments**, with the reason stated in `notes:` — that axis targets homozygous-deletion tumours with absent protein, and this syndrome retains archetype MTAP. Better to record why it is absent than to leave a later curator to rediscover the argument.

## CI

`check-snippet-length` failed on a 4-word prevalence fragment I had trimmed to dodge an `ﬁ` ligature. The check was right — "unrelated DMS-MFH-affected families possess" carries no propositional content. Replaced with two complete sentences documenting the cohort. I hit the same ligature again adding the variant evidence and handled it the same way.

## Not done this round

The `animal_models` / `mechanistic_hypotheses` / `clinical_trials` / histopathology suggestions are all good and all need references fetched and verified first. I would rather land the blocking fixes than add four more sourced sections in the same push. Happy to do them in a follow-up if you want them in this PR.

## Validation

`just validate-disorders` passes. Snippets **33/33** (up from 23). `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-folded-hyphens` all clean. Compliance 93.2%.

gingin77:
Both fixed. You called it exactly: *"the same failure mode as last round, in the other half of the disease."*

## 1. The somatic second hit was in the paper the whole time

Last round I said no study had measured MTA; the measurement was in that paper's Results. This round I said the tumour-suppressive mechanism was uncharacterised, in **three places** — and the same paper characterises it:

> direct sequencing of this patient's osteosarcoma genomic DNA demonstrated homozygosity for the c.885A>G mutation … LOH analysis … revealed complete loss of the WT allele from the unaffected chromosome

That is Knudson's two hits, and it answers the question I had recorded as open, in the direction of direct loss of MTAP function.

Added a `Somatic Loss of the Wild-Type MTAP Allele` node between the germline mutation and the sarcoma node. Its incoming edge is `INDIRECT_KNOWN_INTERMEDIATES`, not `DIRECT`, per your point — the germline lesion **predisposes to** a second independent somatic event rather than causing the tumour itself.

All three passages corrected, and I kept the parts that are genuinely open rather than over-correcting: what the biallelic metabolic state does to permit transformation, and whether the abnormal bone environment additionally contributes.

## 2. Tumour histology

`HP:0002669 Osteosarcoma` and `HP:0100244 Fibrosarcoma` added alongside the general `HP:0100242` anchor, which stays since HPO has no bone-sarcoma term. You are right that the reclassification is the clinically consequential part — the tumours were reported as MFH or bone fibrosarcoma, and re-review found osteoid.

One note on the fibrosarcoma snippet: it stops at `"The diagnoses were either MFH"` because a superscript reference marker splits the sentence at a line break in the extracted text, and no exact-substring quote can span it. The `explanation` says so and gives the full sentence, rather than reconstructing a quote.

## Suggestions

**The `PMID:10612808` explanation** — you were right that dropping "the hereditary specimen" changed what the quote was being used to say. Corrected; the hereditary half matters as much as the sporadic half, since together they are what motivated the search for a shared tumour suppressor.

Also done: `Surveillance and Symptomatic Management` retyped `BEHAVIORAL` → `OTHER`; the limb-girdle/FSHD-like distribution recorded on the myopathy descriptor while keeping the general term (two families is not enough to bind a distribution term); and the residual "first step … is inference" wording corrected, since the MTA measurement made it a measurement.

Not taken, all needing references fetched and verified first: the Mtap mouse as `FAILS_TO_RECAPITULATE` with a `HUMAN_MODEL_MISMATCH`, NCT00007046, serum MTA as a `biochemical:` record, and the two further HPO terms.

## Also in this push

Rebased onto main (120 commits) and migrated five `PARTIAL` items to the new `supports`/`directness` schema from #10003.

`just validate-disorders` passes. Snippets **37/37** (up from 33). `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-snippet-grading`, `check-folded-hyphens` all clean. Compliance 93.9%.

gingin77:
Both fixed. Finding 2 was a real graph error I introduced, and I want to name it as that rather than as a placement quibble.

## 2. The second-hit edge was attached to the wrong parent

The edge sat in `downstream` of `Impaired Polyamine and Methionine Salvage Metabolism`, so the pathograph asserted **that impaired polyamine/methionine salvage causes the second somatic hit.** Nothing supports that. You spotted the tell precisely: the edge's own description names a different subject — *the germline lesion* — and its evidence is tumour homozygosity, a genetic observation with no metabolic content at all.

Moved to `MTAP Retroviral-Exon Splicing Mutation`. The graph is now the Knudson chain the prose describes:

```
germline splicing mutation ──> somatic loss of the wild-type allele ──> sarcoma
        │
        └──> dysregulated isoform splicing ──> impaired salvage metabolism ──> dysplasia
```

The metabolic node now feeds only the dysplasia, which is what its evidence supports, and the graph stays a single connected component.

**On `causal_link_type`** — you flagged it reads oddly with the new parent and said you had no confident answer. I changed it to `INDIRECT_UNKNOWN_INTERMEDIATES`, reasoning that the germline allele does not *cause* the second hit at all: LOH is a stochastic somatic event, and what the inherited allele supplies is the vulnerability that makes that loss consequential. `KNOWN_INTERMEDIATES` would assert steps in between that nobody has described. Open to being wrong about this one.

## 1. The rationale contradicting the entry

Fixed, and your read of the orphaned line wrap was exactly right — `is inference` / `from what` is precisely where my previous edit stopped. I had written a no-op reformat instead of the correction I intended and never checked the result.

Both halves were wrong for different reasons: "the first step ... is inference from what the enzyme is known to do" contradicts line 214 verbatim, and "nor the sarcoma" became false in the very commit that added the second-hit node graded `ESTABLISHED`. Rewritten close to your suggested version — **both endpoints are measured, the road between them is not** — keeping the dysplasia half, which is still true.

## Suggestion taken

Serum MTA is now a `biochemical:` record bound to `CHEBI:17509`, reusing the verified quotes. You are right that it was a candidate diagnostic analyte carried only as pathophysiology prose.

Curated explicitly as a candidate rather than a test: five people is the entire dataset, no reference interval, no clinical assay, and nothing reported about whether the level tracks disease activity or sarcoma risk. The duplication with the pathophysiology node is deliberate and stated in `notes:` — the node needs that evidence for its `ESTABLISHED` grade; this record answers a different question.

(`Biochemical` has no `description` slot, so the prose lives in `notes:`.)

## Not taken

The `Mtap` mouse as `FAILS_TO_RECAPITULATE`, NCT00007046, and `HP:0005010` all need references fetched and verified first, and I would rather land the blocking fixes than add three sourced sections in the same push.

## Validation

`just validate-disorders` passes. Snippets **39/39**. `validate-terms`, `check-entity-refs`, `check-snippet-grading`, `check-folded-hyphens` clean. Compliance 94.1%.

gingin77:
All four fixed, and I did the end-to-end read rather than just the instance.

## 1. `notes:` narrating this PR

Deleted, both of them. You are right about the reason — `notes:` renders as clinical content, and a reader there is asking about the disease, not the review thread. Both narratives are already in the history record, so nothing was lost.

Your non-blocking aside was right too: the same voice had crept into two `description` fields. Removed there as well.

## 2. Comparative genomics graded `HUMAN_CLINICAL`

Regraded `COMPUTATIONAL` at **both** occurrences together — thank you for flagging that `check-snippet-grading` keys on the sentence, because fixing one alone would have traded the finding for a gate failure. The first occurrence also had no `directness` at all; it now carries `DIRECT` to match its sibling.

## 3. Metaphyseal striations

Added (`HP:0031367`). Nothing needed fetching, as you said. I also used the same sentence as a second evidence item on `Bone Infarction`, since it localises the infarcts to bone marrow and so narrows what the broader `HP:0010885` binding is standing in for — which is what that mismatch note was apologising about. Quote starts after "phenotype" to clear the `bone-dysplasia` mid-hyphen wrap.

## 4. Direction asserted where the description says nothing was measured

Both modifiers dropped, each with a `description` recording that the direction is inferred from the enzymology rather than measured. `GO:0017061` keeps `DECREASED` — the MTA measurement earns that one.

## On the pattern you named

You were right that fixing this instance alone would miss the point — three consecutive rounds of the entry contradicting itself is a process problem, not three coincidences.

So I extracted **every prose field in the file** and flagged all absence and negation claims for manual review. Thirteen. Twelve were accurate and mutually consistent. One was stale: an `explanation` still read "the mechanism remains unmapped," which now contradicts the rewritten rationale saying both endpoints are measured. Narrowed to "the route between the measured endpoints remains unmapped."

That is a check I should have been running before pushing, not after being told three times.

## Suggestions

**The `Mtap(lacZ)` mouse is finally in**, and you were right that it was worth the one command. Curated `FAILS_TO_RECAPITULATE` at `LOW` fidelity: losing one allele *does* predispose to malignancy — supporting the tumour-suppressor mechanism — but the tumour is T-cell lymphoma and there is no bone dysplasia at all. Its `limitations` record that this is a null allele, that the retroviral-derived exons have no mouse counterpart so the isoform-balance effect is absent entirely, and that nothing distinguishes whether the tissue mismatch follows from that or from species differences in lineage vulnerability.

And you are right that it partly answers the entry's own single-laboratory gap. That gap now distinguishes which strand is corroborated from outside the discovery lab — MTAP as a tumour suppressor independent of CDKN2A/ARF — from the retroviral-exon mechanism, which remains single-source.

**PRMT5 gap added.** The exclusion from `treatments` stands unchanged, but you are right the question is better posed than the exclusion implies: serum MTA accumulates and the tumour does lose the wild-type allele, so whether these tumours reach the MTA-high PRMT5-dependent state is well formed and unmeasured. The discussion notes say plainly that nothing here supports offering PRMT5 inhibition to a patient.

Not taken: genetic counselling as a treatment, osteosarcoma-protocol chemotherapy, NCT00007046, and the QoL gap — all need references fetched and verified first.

## Validation

`just validate-disorders` passes. Snippets **43/43**. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-folded-hyphens` clean. Compliance 92.9%.

*(Amended the commit message after pushing — backticks in my `-m` string were shell-expanded and ate the text of item 4. Content unaffected.)*

gingin77:
CRITICAL fixed, both 🟡s fixed, and suggestion 1 taken. Pushed in `2ca0016df0`, `fe578bc01d` and `57abe27a4c`.

## The failing test

You were right on the mechanism as well as the fix. I reproduced it locally before touching anything:

```
FAILED test_failure_to_recapitulate_links_are_substantiated
AssertionError: ... ['animal_models[0].modeled_mechanisms[0] missing evidence']
```

Evidence added at link level, graded `SUPPORT`, quoting the T-cell lymphoma sentence — the negative claim is what the quote supports, and the negative direction is `FAILS_TO_RECAPITULATE`'s job. `uv run pytest tests/test_data.py -k failure_to_recapitulate` passes; I then ran the whole `tests/test_data.py` lane — **47,112 passed**.

**The path-gating point is the more useful half of that finding.** A green `test (3.13)` on a `kb/`-only PR means the KB data tests did not run, so it carries no information about KB validity. I now run the full lane locally before pushing a curation PR rather than reading the badge, and I did the same on #10010, where I added a `FAILS_TO_RECAPITULATE` link this round and would have reproduced this exact failure otherwise.

## The `REFUTE` grading

Corrected to `SUPPORT`. Your framing is the clarifying one: `REFUTE` asserts that the cited sentence contradicts the claim it is attached to, and here the sentence *is* the observation the readout is made of. The intended meaning was already carried twice, by the relationship and by `direction: ALTERED` — grading it a third time inverted the evidence rather than reinforcing it. Both explanations now say why the grading is what it is, so the next reader is not tempted to "fix" it back.

## The narration line

Trimmed to end at "cortical-growth phenotype." Removing it from two `notes:` blocks and reintroducing it in a `description` in the same commit is exactly the failure you describe — I was auditing the fields I had been told about instead of the sentence.

## Suggestion 1 — taken, and it was a real inconsistency

The gap now separates the two strands instead of leaving them to agree by careful reading: MTAP as a tumour suppressor in its own right **is** corroborated from outside the discovery laboratory by the Kadariya mouse work, so the sarcoma arm does not rest on one group; the retroviral-exon splicing mechanism, which is what the two attached nodes assert, has no such corroboration.

## Suggestion 2 — taken

`frequency: FREQUENT` added, with the ascertainment caveat in `notes:`: three 1999 pedigrees, ascertained *because* they had sarcoma, so 35 percent describes the families that came to attention rather than a carrier's lifetime risk. The band inherits the same bias as the number, and says so.

Suggestion 3 items are still open and still non-blocking.

**Snippets 44/44.** `validate`, `validate-terms`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-grading`, `check-snippet-length`, `check-duplicate-keys`, `validate-history` all clean.

gingin77:
Rebased onto current `main` (27 commits of drift; `57abe27a4c` → same tree, new base). No conflicts and no content change — the post-rebase diff against `origin/main` is exactly the intended file set, with the stub as the only deletion.

Re-ran the full lane on the new base: **47,199 passed**, including `tests/test_causal_targets.py`, which did not exist when this branch was cut. This entry's pathograph targets are clean. (The same rebase on #10010 found eight broken ones there, so the check was not a formality.)

Snippets **44/44**. This round is now against the tree that will actually merge.

gingin77:

gingin77:
The `claude-review` check failed on the current head (3a39afb), but it's not a problem with this PR's content.

The job's log shows the review agent never actually started reviewing anything — it errored out after 384ms with:

```
detail="You've hit your limit · resets 4:20pm (UTC)"
```

That's an account-quota limit on the review credential itself, not a finding about the diff. I'll check back after the 16:20 UTC reset and re-trigger the review then; no code change is needed here.

For context on the rest of this push: it merges `main` in to resolve the conflict this PR had (two derived cache CSVs, both sides' additions kept), and addresses the remaining non-blocking suggestions from the last full review (genetic counselling treatment, a formal `mechanistic_hypotheses` entry for the PRMT5 discussion, a `clinical_burden` knowledge gap, and one over-graded `directness` value). All prior CRITICAL/IMPORTANT review findings from earlier rounds were already fixed before this push; `just validate-disorders` and the file's pytest suite both pass clean on the merged head.

---

gingin77:
/review

---

Issue #8655, opened 2026-08-16
Curate hereditary amyloidosis (MONDO:0018634)


matentzn:
Curate a dismech entry for **hereditary amyloidosis** ([MONDO:0018634](https://monarchinitiative.org/MONDO:0018634)) — the group of inherited systemic amyloidoses in which a germline variant in a precursor protein promotes misfolding, fibril formation, and progressive extracellular amyloid deposition in nerve, heart, kidney, and other organs.

This is a `CURATE_ROOT_WITH_SUBTYPES` target: model the root with `has_subtypes` capturing the principal hereditary forms by precursor protein — **hereditary transthyretin amyloidosis (ATTRv)**, **apolipoprotein A-I/A-II (AApoAI/AII)**, **gelsolin (AGel)**, **fibrinogen A alpha-chain (AFib)**, and **lysozyme (ALys)**. Subtype-level entries already exist (`ATTR_Amyloidosis`, `ADan_amyloidosis`) to cross-reference; the entry should also `conforms_to` the `amyloidogenesis` module. Keep the acquired amyloidoses (AL, AA) out of scope — this umbrella is the *hereditary* group.

Source: top of [priority dashboard](https://dismech.monarchinitiative.org/dashboard/priority.html) (`CURATE_ROOT_WITH_SUBTYPES`, score 75.0).

Tracker: part of #1079.

cmungall:
> **Agent-authored note:** this comment was prepared after a joint review and discussion with @cmungall; it reflects our shared conclusion.

We strongly believe this concept should be represented as a **`kb/groupings/` grouping**, not as a new umbrella `Disease` entry. In particular, please do not create `kb/disorders/Hereditary_Amyloidosis.yaml` merely because the dashboard emitted `CURATE_ROOT_WITH_SUBTYPES`.

That action code is not a curation decision. Here it was generated because MONDO:0018634 has nine children and did not match the prioritizer's very small grouping-term regex. The prioritizer documentation itself says that these heuristics do not solve lump-versus-split perfectly. Repository modeling policy and biological review must take precedence over a priority-list suggestion.

Why this is grouping-shaped:

- “Hereditary X” is a useful review red flag when **X** is already a broad pathological category. It often denotes a family of inherited diseases rather than one pathomechanism. It is not an automatic rule, but it should force an explicit boundary check.
[...]

PR #10014, opened 2026-08-29
curate: Add Familial Visceral Amyloidosis (MONDO:0007099) with four precursor-protein subtypes

### LINKED ISSUE #9603: Curate familial visceral amyloidosis (MONDO:0007099)
Curate a dismech entry for **familial visceral amyloidosis** ([MONDO:0007099](https://monarchinitiative.org/MONDO:0007099)) — an autosomal dominant hereditary systemic amyloidosis in which a variant precursor protein (classically apolipoprotein A-I/`APOA1`, with fibrinogen A alpha-chain `FGA` and lysozyme `LYZ` forms in the same clinical family) misfolds and deposits as amyloid fibrils predominantly in viscera — kidney, liver, spleen and gut — rather than in nerve or heart.

This is a `CURATE_ROOT_WITH_SUBTYPES` target: model the precursor-protein-defined forms with `has_subtypes` (AApoAI, AFib, ALys), since the deposited precursor determines organ tropism and prognosis. The entry should declare `conforms_to` against `amyloidogenesis#Amyloid Fibril Formation and Extracellular Deposition`, substituting the variant precursor at the trigger node, and cross-reference the existing `Hereditary_Systemic_Amyloidoses` grouping.

Tracker: part of #1079.

sierra-moxon:
Closes #9603

Adds `kb/disorders/Familial_Visceral_Amyloidosis.yaml` for MONDO:0007099, the Ostertag-type hereditary amyloidosis in which a mutated blood protein made by the liver misfolds and builds up in the internal organs — mostly the kidney, but also liver, spleen and gut — instead of in nerve or heart. Patients present with protein in the urine and slowly lose kidney function over years to decades. Four different precursor proteins can cause it, and which one a patient has determines which organs are hit and how fast; that is curated as four `has_subtypes` entries (AApoAI/APOA1, AApoAII/APOA2, AFib/FGA, ALys/LYZ), matching MONDO's own four children of this class.

## Why one Disease entry with subtypes, rather than a grouping

This was the first decision and it is worth stating plainly, because #8655 recorded a maintainer directive *against* creating an umbrella disease entry for hereditary amyloidosis.

That directive was about MONDO:0018634 (hereditary amyloidosis), which spans the neuropathic, cardiac, cerebral and visceral forms — too broad to share one mechanism. MONDO:0007099 is narrower: it is precisely the non-neuropathic visceral subset, and its four members do share one pathograph end to end.

```
variant plasma protein (APOA1 / APOA2 / FGA / LYZ, made by the liver)
        v
destabilization and beta-sheet conversion
        v
cross-beta amyloid fibril formation and extracellular deposition
        v
precursor-determined visceral organ tropism   <-- the only node that differs by subtype
        v
progressive visceral amyloid accumulation
        v
progressive renal excretory failure
```

They also share a diagnostic pathway (renal biopsy + amyloid typing + precursor-gene sequencing) and the only disease-modifying treatment there is (removing the liver that makes the variant protein). The one thing that differs — organ tropism — is curated as its own explicit node rather than being dissolved into the generic accumulation step.

The alternative was a grouping, but a grouping needs members, and none of the four precursor forms has a `kb/disorders/` file yet. That would have produced an empty grouping and left the concept clinicians actually use ("hereditary renal amyloidosis") uncurated. The reasoning and the revisit condition — what to do if AFib or AApoAI is later split out as its own entry — are written into the entry's `notes`.

Five of the six pathophysiology nodes declare `conforms_to` against `kb/modules/amyloidogenesis.yaml`.

## Also changed: the Hereditary Systemic Amyloidoses grouping

`kb/groupings/Hereditary_Systemic_Amyloidoses.yaml` listed AApoAI, AApoAII, AFib and ALys as intended future members it could not yet reference. This PR adds Familial Visceral Amyloidosis as a DISEASE member, which reaches all four at once, and updates the rationale and notes so they describe what the file now contains. AGel/Finnish-type is still an unlisted TODO there; I did not add it, since that is a separate call about an existing entry.

## A real disagreement in the literature, recorded rather than resolved

The two large UK series disagree about whether the fibrinogen form is a kidney-restricted disease.

- Gillmore et al. (PMID:19073821), 71 prospectively followed patients: extra-renal disease rare, no infiltrative cardiomyopathy in anyone.
- Stangou et al. (PMID:19633201), 22 patients worked up for combined liver-kidney transplant: variant fibrinogen amyloid found in excised atheroma and endomyocardial biopsies, autonomic neuropathy in half, "a systemic amyloid disease with visceral, vascular, cardiac, and neurologic involvement".

Part of this is ascertainment — the transplant workup biopsies tissue the observational cohort had no reason to sample. It matters clinically, because cardiovascular amyloid is what can disqualify a patient from the transplant that would otherwise be curative. The entry curates the kidney-dominant pattern as the general case and records the disagreement as a `CONTROVERSY` discussion attached to the AFib subtype and the tropism node, rather than picking a winner.

- **Hereditary beta-2-microglobulin amyloidosis (B2M D76N).** The deep-research report proposed it as a fifth precursor form, and it is genuinely visceral. But it is not a MONDO child of MONDO:0007099, and adding it would push the concept past its ontology grounding on the strength of one review. Reasoning is in `notes`; revisit if MONDO places it here.
- **Population prevalence.** No numerator/denominator study found. Orphanet (ORPHA:85450) has an epidemiology record, but the Orphadata bulk XML needed to produce a quotable cache entry is not in this checkout, so there was nothing to cite. The one occurrence figure curated is a different measure — the 5% of apparently sporadic systemic amyloid that turns out to be hereditary AFib, which is the number that drives the decision to send genetic testing.
- **PMID:16916739** (Lane et al.) was fetched but its cache entry has no retrievable text, so nothing is quoted from it and it is not cited.
- **GeneReviews.** No chapter exists for any of the four precursor forms (PubMed searched 2026-08-28). The phenotype baseline came from the UK National Amyloidosis Centre cohort series instead.

## Deep research

One falcon (Edison) run, 773s, 21 citations, reference validation 11/11 resolved with a 0.0 confabulation rate. Note its relevance pass reports on-topic 1 of 11 with off-topic 0, meaning most references were undecided rather than judged off topic. Its central recommendation — treat MONDO:0007099 as a legacy umbrella and point the disease-level record at precursor-defined subtypes rather than assigning one prognosis to all forms — is what the `has_subtypes` structure implements. Every claim taken from the report was re-verified against the cited abstract before use.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed (schema + terms + references) |
| `just count-verified-snippets` | 68/68 verified |
| `just check-snippet-grading` | OK, no new divergences |
| `just check-title-snippets` | OK |
| `just check-snippet-length` | OK |
| `just check-folded-hyphens` | OK |
| `scripts/check_reference_titles.py --against-ref origin/main` | OK |
| `just validate-grouping` / `just check-groupings` | passed; no regression on the edited grouping |

No stub existed for MONDO:0007099, so none was deleted. History record added at `history/disorders/Familial_Visceral_Amyloidosis/`.

https://claude.ai/code/session_01Um4JiZfX8qfKbLxsebqMMX

### DIFF
--- a/kb/disorders/Familial_Visceral_Amyloidosis.yaml
+++ b/kb/disorders/Familial_Visceral_Amyloidosis.yaml
@@ -1095,10 +1095,12 @@ histopathology:
 - name: Congo Red-Positive Amyloid Deposition
   description: >-
     Congo red staining of the affected tissue shows acellular extracellular
-    deposits, classically with apple-green birefringence under polarized light.
-    This establishes that the deposit is amyloid but says nothing about which
-    protein it is made of, so typing by immunohistochemistry or mass spectrometry
-    remains a separate and mandatory step.
+    deposits that are birefringent under cross-polarized light. The cited renal
+    series describes the colour as red-green; the classical textbook phrase is
+    apple-green, and the two name the same optical finding. This establishes that
+    the deposit is amyloid but says nothing about which protein it is made of, so
+    typing by immunohistochemistry or mass spectrometry remains a separate and
+    mandatory step.
   diagnostic: true
   finding_term:
     preferred_term: amyloid deposition
@@ -1687,7 +1689,7 @@ differential_diagnoses:
 prevalence:
 - subtype: AFib
   population: Patients with systemic amyloidosis, United Kingdom national series 1987-2012
-  measure_type: PERIOD_PREVALENCE
+  measure_type: UNKNOWN
   prevalence_class: UNKNOWN
   notes: >-
     A denominator-based figure, but a denominator of amyloidosis patients rather
@@ -1695,7 +1697,10 @@ prevalence:
     national amyloidosis service over 25 years had AFib amyloidosis. It bounds
     how often the commonest precursor form turns up in an amyloidosis clinic, not
     how common the disease is in a population, so no rate per 100,000 is derived
-    from it.
+    from it. measure_type is UNKNOWN rather than PERIOD_PREVALENCE for the same
+    reason, and matches the sibling record below: PrevalenceMeasureEnum's values
+    all presuppose a population denominator, and neither of these two records
+    has one.
   evidence:
   - reference: PMID:39417966
     reference_title: "Clinical manifestations, diagnosis and treatment of hereditary fibrinogen Aα-chain renal amyloidosis: one case report and systematic review."


sierra-moxon:
Thanks — all four items were real and the entry is better for them. Everything blocking is now fixed, plus four of the five suggestions. Pushed in two commits.

## The four IMPORTANT items

**Item 1 — nephrotic syndrome and edema.** Added both, quoting the exact sentence you pointed at. I also added a bounding evidence item: in the apolipoprotein A-I subtype, median protein excretion at diagnosis was 0.3 g/24h despite universal renal amyloid, so nephrotic-range proteinuria belongs to the glomerular precursor forms and not to the medullary AApoAI lesion. Both new phenotypes are scoped `subtypes: [AFib]` for that reason.

**Item 2 — AFib genotype-phenotype gradient. Partly done, and I want to flag what I did not do.** I could not verify the specific numbers. The onset gradient (~57y E526V, ~45y R554L, ~24.5y frameshift) and the 22% vs 83% recurrence contrast are in the falcon report, attributed to a 32-patient French series, but neither figure appears anywhere in `references_cache/PMID_39417966.md` or in any other cache in this repo — I grepped for `24.5`, `57 years` and `R554L` and got zero hits. Citing them to PMID:39417966 would have put a number next to a PMID that does not contain it, which is the failure mode the reference discipline exists to prevent.

What I did add, all from the cached full text:

- the sex difference in onset — "We found the onset age to be lower in women than in men (P < 0.05)"
- the pooled post-transplant recurrence rate — 6 of 16 transplanted patients (37.5%)
- prose in the FGA `association` explaining why allele class bears on the isolated-kidney vs combined liver-kidney decision, which was the substantive point behind your finding

The unverified gradient is recorded in `notes` as not-curated-for-want-of-a-source, with what to do if the French series is fetched later.

**Item 3 — prevalence note overstated the gap.** You were right. Added an AFib `prevalence` record carrying the 87/5100 UK national figure, and rewrote the note. I left `rate_per_100000` empty and `prevalence_class: UNKNOWN` on both records deliberately: the denominator is amyloidosis patients, not a population, so deriving a population rate from it would be wrong. The note now says that explicitly rather than claiming no denominator study exists.

**Item 4 — biochemical section.** Added, with two entries (alkaline phosphatase `NCIT:C61016`, gamma-glutamyltransferase `NCIT:C64847`), scoped to AApoAI, using the PMID:15131802 quotes. Note `Biochemical` has no `description` slot, so the prose is in `notes`.

## Suggestions taken

- **Histopathology** — added. Congo red birefringence, plus the glomerulus-restricted AFib pattern with the "should always prompt a search for a mutation" quote and independent confirmation from PMID:29142973.
- **AL misdiagnosis rate** — added to the differential: 7 of 46 pooled cases misdiagnosed as AL, with renal function deteriorating after chemotherapy and stem-cell transplant.
- **APOA1 p.Leu202Arg** — added as an `OPEN_QUESTION` discussion attached to `inheritance#Autosomal dominant` and `has_subtypes#AApoAI`, exactly as you suggested. The inheritance block is unchanged: one consanguineous family with an unaffected heterozygous brother is equally compatible with reduced penetrance in a dominant allele. The case snippet is graded `PARTIAL` for that reason.
- **Grouping notes line** — rewrapped.

## Suggestion not taken

**Animal model.** The transgenic APOA2 Stop78Ser mouse cache (`DOI:10.1016/j.kint.2019.03.013`) is `content_type: unavailable`, so there is no quotable text for a `modeled_mechanisms` link's evidence. I looked for a PubMed-indexed version and did not find one. Recorded in `notes`.

**AGel/Finnish-type as a grouping member** stays out of scope here, as you suggested — it is a call about an existing entry, not about this one.

## One thing I fixed on my own before your review landed

Three HPO bindings were wrong or too generic, and one was an evidence-to-claim mismatch worth naming:

| Phenotype | Was | Now | Why |
|---|---|---|---|
| Hepatic Amyloid Deposition | `HP:0011034` Amyloid deposition | `HP:0012280` Hepatic amyloidosis | a specific term exists |
| Cardiac Amyloid Involvement | `HP:0001638` Cardiomyopathy | `HP:0030843` Cardiac amyloidosis | the claim is amyloid in the heart, not cardiomyopathy |
| Gastrointestinal Amyloid Involvement | `HP:0002239` Gastrointestinal hemorrhage | `HP:0011034` Amyloid deposition, plus a new Gastrointestinal Inflammation phenotype on `HP:0004386` | the cited evidence describes amyloid lesions and IBD-like colitis and never mentions bleeding — the old binding claimed something the snippet did not support |

Splenic amyloid keeps the generic `HP:0011034`; HPO has no splenic amyloid class, and no term beats a bad one.

## Validation

| Check | Result |
|---|---|
| `just validate` | passed |
| snippets | 83/85 verified; the 2 skipped are DOI-prefixed and were checked by hand against the cache |
| `check-duplicate-keys` / `check-entity-refs` | OK |
| `check-snippet-grading` / `check-title-snippets` / `check-snippet-length` | OK, no new findings |
| `check-folded-hyphens` | OK — caught one split I introduced ("A general-/population"), fixed |
| `check-environmental-evidence` | OK |
| `scripts/check_reference_titles.py --against-ref origin/main` | OK |
| `validate-grouping` / `check-groupings` | passed, no regression |
| `validate-history` | passed |

A second `EDIT` history record covers these changes.

One cosmetic note for the record: `references_cache/PMID_29142973.md` has a non-breaking space in its title between "Novel" and "Variants". The YAML uses a normal space, which passes `check_reference_titles.py`. I left it readable rather than copying an invisible character into the KB.

sierra-moxon:
Picked up the two cosmetic notes from the approval, since both were right.

**measure_type consistency.** Both prevalence records now use `UNKNOWN`. Neither has a population denominator — one is 87/5100 amyloidosis-clinic patients, the other is the 5% of apparently sporadic amyloid that proves hereditary — and every `PrevalenceMeasureEnum` value presupposes a population denominator. `PERIOD_PREVALENCE` on the first one was overclaiming. The reason is now stated in the record's `notes` so the next curator does not "fix" it back.

**Birefringence colour.** The description said apple-green where the quoted figure legend says red-green. Reworded to lead with the source's wording and note that the classical textbook phrase names the same optical finding, rather than silently picking one.

Validation after both: `just validate` passes, 83/85 snippets verified (2 DOI-prefixed, hand-checked), `check-folded-hyphens`, `check-entity-refs` and `check-duplicate-keys` all OK.

This push drops the approval, so I have re-triggered review.

PR #10022, opened 2026-08-29
curate(Cranial_Neuralgia): Add Cranial Neuralgia (MONDO:0016374) with nerve-defined subtypes

### LINKED ISSUE #9605: Curate cranial neuralgia (MONDO:0016374)
Curate a dismech entry for **cranial neuralgia** ([MONDO:0016374](https://monarchinitiative.org/MONDO:0016374)) — paroxysmal, severe, lancinating facial or head pain in the distribution of a cranial nerve, most often caused by neurovascular compression at the nerve root entry zone producing focal demyelination and ephaptic cross-talk between fibres.

This is a `CURATE_ROOT_WITH_SUBTYPES` target: model the nerve-defined subtypes with `has_subtypes` (trigeminal neuralgia — by far the commonest, glossopharyngeal neuralgia, nervus intermedius neuralgia, occipital neuralgia). The dismech knowledgebase currently has **no** neuralgia entry at all. Capture the classical-vs-secondary distinction (multiple sclerosis plaque, posterior fossa tumour) and the carbamazepine/oxcarbazepine sodium-channel-blockade treatment mechanism via `target_mechanisms`.

Tracker: part of #1079.

sierra-moxon:
Adds a knowledge base entry for **cranial neuralgia** (MONDO:0016374). The knowledgebase previously had no neuralgia entry at all.

The entry is curated as **one disease with four nerve-defined subtypes** — trigeminal, glossopharyngeal, nervus intermedius, and occipital — rather than as a grouping of separate diseases. The reason is that the classical forms share one mechanism step for step: an artery presses on the sensory nerve root where it enters the brainstem, that pressure strips myelin off a short segment, bare axons end up touching each other, and impulses then leak sideways from touch fibres into pain fibres. A light touch on the face therefore sets off a shock of severe pain. Love and Coakham (PMID:11701590) make exactly this generalisation, concluding that demyelination and ephaptic spread underlie most or all of the vascular-compression cranial nerve syndromes. What changes between subtypes is which nerve carries the lesion, and so where the pain is felt and which vessel is at fault.

```
              hypertension (predisposes)
                       |
                       v
  neurovascular compression at root entry zone ──┐
  MS pontine plaque ───────────────────────────┐ │
  posterior fossa tumour ──────────────────────┤ │
                                               v v
                              focal demyelination + bare axons touching
                                    |                        |
                                    v                        v
                    sodium channel dysregulation  ──►  ectopic discharge +
                    (carbamazepine acts here)          ephaptic cross-talk
                                                              |
              trigger zone touch ─────────────────────────────┤
                                                              v
                                             paroxysmal lancinating pain

  occipital nerve entrapment (extracranial) ────────────────► same pain outcome
                                                              (mechanism not shared)
```

**The classical/secondary distinction is modelled structurally**, not as prose. Multiple sclerosis and a posterior fossa tumour get their own trigger nodes that converge on the same demyelination node the artery converges on. That is the point: three different upstream causes, one shared lesion, one shared downstream chain. Tumour and MS together account for about 15% of trigeminal neuralgia, which is why MRI is required rather than optional.

**Treatments are wired into the graph via `target_mechanisms`.** Carbamazepine and oxcarbazepine act on the sodium-channel node. Microvascular decompression acts on the compression node — it is the only cause-directed option, which is why it gives the most durable relief. Percutaneous ablation acts on the ectopic-discharge node instead, and that difference is the mechanistic reason it buys pain control at the cost of facial numbness where decompression does not.

## One thing to look at

Occipital neuralgia is the awkward member. The issue asked for it, and ICHD-3 and the HPO neuralgia hierarchy both group it here, but the greater and lesser occipital nerves are branches of the C2–C3 dorsal rami — they are not cranial nerves. MONDO:0016374 does not list occipital neuralgia among its descendants either. The compression points described for these nerves are extracranial, where the nerve curves around a muscle or crosses the occipital artery (PMID:39217283), not a brainstem root entry zone.

So it is included as a subtype, but its mechanism node is kept separate and does **not** claim the ephaptic chain, and its edge to the pain outcome is `INDIRECT_UNKNOWN_INTERMEDIATES` rather than a confident causal link. The caveat is stated in the subtype description and again in the entry notes. If a reviewer thinks it should be dropped instead, that is a reasonable call and easy to make.

## What the deep research said, and why the entry disagrees

The falcon (Edison) report generated for this entry argued the opposite: that cranial neuralgia is "an umbrella clinical category, not one etiologically uniform disease" and that the entities should be separate entries. That reading is recorded in the entry notes rather than dropped, because it is right about the things it is right about — the epidemiology and the choice of procedure genuinely do not transfer between subtypes, which is why prevalence and treatment records here are subtype-scoped instead of stated at the root. What is shared is the mechanism, not the clinical management.

`stubs/Trigeminal_Neuralgia.yaml` is deleted, since MONDO:0008599 is now covered as a subtype. If the per-subtype pathographs later diverge enough to justify splitting trigeminal neuralgia into its own entry, re-nominating it is the route.

## Recorded gaps

Three `discussions` entries, all pointing at real soft spots rather than decoration:

- The histopathology behind the central mechanism comes from trigeminal roots taken at surgery. For glossopharyngeal and nervus intermedius neuralgia the mechanism is inferred by analogy from operative findings and from the fact that decompression works. The one-disease decision above rests on that generalisation holding.
- Neurovascular contact is present on 36% of asymptomatic nerves, and even severe contact on 13% of asymptomatic sides. Compression is clearly not sufficient, and the sex difference — severe contact explains 75% of men but only 38% of women — implies a factor this graph does not represent.
- A `HUMAN_MODEL_MISMATCH` on the rat infraorbital constriction model that supplies the sodium-channel subunit detail. It injures a distal peripheral branch, not the central myelinated root segment, and produces sustained hypersensitivity rather than the triggered, remitting, paroxysmal human pattern. Devor and colleagues say outright that no satisfactory animal model exists.

## Validation

| Check | Result |
|---|---|
| snippets verified against cache | 74 of 77 (3 skipped by prefix) |
| `check-duplicate-keys`, `check-entity-refs` | passed |
| `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens` | passed, no new baseline entries |
| `check-environmental-evidence`, `check-stubs` | passed |
| `scripts/check_reference_titles.py --against-ref origin/main` | passed |

Deep research: falcon (Edison), 14 of 14 references resolved, confabulation rate 0.0. `just preflight-dr` returned SKIP because MONDO records no causal gene for this term, so the manual synonym and duplicate preflight was done instead. Most evidence came from direct PubMed search rather than the report's own citations, which are largely secondary reviews cited by DOI.

The `Chronic arterial hypertension` environmental entry is deliberately left without an `exposure_term`. ECTO and XCO carry only antihypertensive-*agent* concepts (`ECTO:2000001`, `ECTO:9001744`, `XCO:0000863`), which describe treatment and would assert the opposite of the claim. The search is recorded in that entry's `notes`.

Closes #9605

https://claude.ai/code/session_01Um4JiZfX8qfKbLxsebqMMX

### DIFF
--- a/kb/disorders/Cranial_Neuralgia.yaml
+++ b/kb/disorders/Cranial_Neuralgia.yaml
@@ -125,7 +125,8 @@ definitions:
     explanation: Gives the three aetiological categories verbatim as proposed by the classification working group.
   - reference: PMID:39847207
     reference_title: Nervus Intermedius Neuralgia.
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: DIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "Like trigeminal neuralgia, NIN can be divided into classic, idiopathic, secondary, or painful neuropathy."
     explanation: Supports extension of the same scheme to nervus intermedius neuralgia, though this review notes the evidence base for that subtype is thin.
@@ -219,13 +220,15 @@ has_subtypes:
     explanation: Surgical series defining the subtype by the nerve involved, which is the axis this entry's subtypes are cut along.
   - reference: PMID:30095334
     reference_title: Nervus intermedius and the surgical management of geniculate neuralgia.
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: DIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "Diagnosis and treatment can be challenging, due to the complex nature of ear sensory innervation, resulting in clinical overlap with trigeminal neuralgia (TN) and glossopharyngeal neuralgia (GPN)."
     explanation: >-
-      Substantiates the diagnostic-overlap caveat in this subtype's description.
-      Graded PARTIAL because it reports the overlap as a clinical difficulty
-      rather than quantifying misdiagnosis.
+      Substantiates the diagnostic-overlap caveat in this subtype's description,
+      which the quoted sentence asserts outright. It reports the overlap as a
+      clinical difficulty rather than quantifying misdiagnosis, so the entry
+      does not put a number on it.
   review_notes: >-
     No `subtype_term`. MONDO was searched for "nervus intermedius", "geniculate
     neuralgia", and the descendants of MONDO:0016374 (cranial neuralgia), which
@@ -376,7 +379,8 @@ clinical_burden:
     explanation: Nationwide cohort of 3273 patients quantifying the psychiatric burden that drives the HIGH assessment.
   - reference: PMID:26174508
     reference_title: "Risk of psychiatric disorders following trigeminal neuralgia: a nationwide population-based retrospective cohort study."
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: INDIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "TN might increase the risk of subsequent newly diagnosed depressive disorder, anxiety disorder, and sleep disorder, but not schizophrenia or bipolar disorder."
     explanation: >-
@@ -425,7 +429,8 @@ pathophysiology:
     explanation: Localises the aetiologically relevant compression to the root entry zone specifically.
   - reference: PMID:27306631
     reference_title: "Trigeminal neuralgia: New classification and diagnostic grading for practice and research."
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: INDIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "neurovascular contact was found in 471 out of 531 symptomatic nerves (89%) and 244 of 681 asymptomatic nerves (36%), indicating high sensitivity but poor specificity."
     explanation: Meta-analytic figures showing why mere contact is insufficient, which is why this node is framed as morphological compression rather than contact.
@@ -557,12 +562,13 @@ pathophysiology:
       channel dysregulation this node represents.
   - reference: PMID:27327156
     reference_title: "Changes in the expression of voltage-gated sodium channels Nav1.3, Nav1.7, Nav1.8, and Nav1.9 in rat trigeminal ganglia following chronic constriction injury."
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: INDIRECT
     evidence_source: MODEL_ORGANISM
     snippet: "trigeminal ganglion in rats subjected to an infraorbital nerve-chronic constriction injury showed upregulation of Nav1.3 and downregulation of Nav1.7, Nav1.8, and Nav1.9 messenger RNA and protein levels"
     explanation: >-
       Rodent constriction-injury data give the direction of change for individual
-      subunits. Graded PARTIAL because the model injures a peripheral trigeminal
+      subunits. Marked INDIRECT because the model injures a peripheral trigeminal
       branch rather than reproducing root-entry-zone demyelination, and the
       subunit pattern has not been confirmed in human trigeminal ganglion.
   downstream:
@@ -636,7 +642,8 @@ pathophysiology:
     explanation: The ignition formulation is the mechanistic account this node encodes.
   - reference: PMID:11803297
     reference_title: "Pathophysiology of trigeminal neuralgia: the ignition hypothesis."
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: INDIRECT
     evidence_source: OTHER
     snippet: "Proof, however, awaits the availability of key experimental data that can only be obtained from patients with trigeminal neuralgia."
     explanation: >-
@@ -716,12 +723,13 @@ pathophysiology:
     evidence:
     - reference: PMID:11701590
       reference_title: "Trigeminal neuralgia: pathology and pathogenesis."
-      supports: PARTIAL
+      supports: SUPPORT
+      directness: INDIRECT
       evidence_source: HUMAN_CLINICAL
       snippet: "Experimental studies indicate that this anatomical arrangement favours the ectopic generation of spontaneous nerve impulses and their ephaptic conduction to adjacent fibres"
       explanation: >-
         Supports ephaptic conduction to adjacent fibres as the substrate for
-        trigger-evoked attacks. Graded PARTIAL because this sentence concerns
+        trigger-evoked attacks. Marked INDIRECT because this sentence concerns
         spontaneous impulses; it does not itself demonstrate that trigger-zone
         afferent traffic is the igniting input.
 - name: Central Demyelinating Plaque in Multiple Sclerosis
@@ -748,7 +756,8 @@ pathophysiology:
     explanation: Human pathology showing that the MS route converges on the same histological lesion.
   - reference: PMID:27306631
     reference_title: "Trigeminal neuralgia: New classification and diagnostic grading for practice and research."
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: DIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "The development of pain paroxysms has variably been explained with the presence of demyelinating plaques in the pons or increased susceptibility of the trigeminal nerve root to neurovascular compression."
     explanation: Records that both explanations for MS-associated disease remain on the table, which is why this node is not asserted as a single settled route.
@@ -823,13 +832,14 @@ pathophysiology:
   evidence:
   - reference: PMID:39217283
     reference_title: "An anatomical analysis of the occipital nerve complex: an essential tool for the application of occipital nerve blocks."
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: INDIRECT
     evidence_source: OTHER
     snippet: "Six, four, and one potential entrapment points were detected for the GON, TON, and LON, respectively."
     explanation: >-
       Cadaveric dissection study establishing that the compression points
       relevant to these nerves are extracranial sites along their course rather
-      than a root entry zone. Graded PARTIAL because it maps anatomy and does
+      than a root entry zone. Marked INDIRECT because it maps anatomy and does
       not show that entrapment at those points causes occipital neuralgia, and
       OTHER because it is a dissection study with no patients or cohort.
   downstream:
@@ -843,12 +853,13 @@ pathophysiology:
     evidence:
     - reference: PMID:24737457
       reference_title: Occipital neuralgia.
-      supports: PARTIAL
+      supports: SUPPORT
+      directness: INDIRECT
       evidence_source: HUMAN_CLINICAL
       snippet: "Occipital neuralgia is an uncommon cause of occipital pain characterized by paroxysmal lancinating pain in the distribution of the greater, lesser or third occipital nerves."
       explanation: >-
         Establishes the pain endpoint in the occipital nerve territory. Graded
-        PARTIAL because it describes the phenotype and does not itself
+        INDIRECT because it describes the phenotype and does not itself
         demonstrate the entrapment-to-pain causal step.
 phenotypes:
 - category: Neurological
@@ -1317,7 +1328,8 @@ treatments:
     explanation: Long-term prospective follow-up of 1185 patients quantifying durable medication-free relief.
   - reference: PMID:8598865
     reference_title: The long-term outcome of microvascular decompression for trigeminal neuralgia.
-    supports: PARTIAL
+    supports: SUPPORT
+    directness: DIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "Major complications included two deaths shortly after the operation (0.2 percent) and one brain-stem infarction (0.1 percent). Sixteen patients (1 percent) had ipsilateral hearing loss."
     explanation: Records the operative risk that qualifies the benefit claim.
@@ -1362,11 +1374,12 @@ treatments:
     evidence:
     - reference: PMID:30860637
       reference_title: European Academy of Neurology guideline on trigeminal neuralgia.
-      supports: PARTIAL
+      supports: SUPPORT
+      directness: INDIRECT
       evidence_source: HUMAN_CLINICAL
       snippet: "Neuroablative treatments should be the preferred choice if MRI does not demonstrate any neurovascular contact."
       explanation: >-
-        Supports the indication for these procedures. Graded PARTIAL because the
+        Supports the indication for these procedures. Marked INDIRECT because the
         guideline states when to use them, not the conduction-block mechanism
         asserted in this link's description.
   evidence:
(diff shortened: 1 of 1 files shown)


sierra-moxon:
Three follow-up pushes since the PR opened, all self-caught rather than review-driven.

**Corrected three `prevalence_class` bands that were off by a decade.** This was a real error in the first commit. Checked against `_band_from_rate()` in `scripts/migrate_prevalence.py`, which is what the repository actually uses:

| Record | Rate | Was | Now |
|---|---|---|---|
| Trigeminal, annual incidence | 4.3 / 100,000 | `BAND_1_9_PER_1000000` | `BAND_1_9_PER_100000` |
| Trigeminal, point prevalence | 3–30 / 100,000 | `BELOW_1_IN_1000000` | `BAND_1_5_PER_10000` |
| Glossopharyngeal, annual incidence | 0.7 / 100,000 | `BELOW_1_IN_1000000` | `BAND_1_9_PER_1000000` |

The trigeminal prevalence range straddles a class boundary (its low end is in `BAND_1_9_PER_100000`, its high end in `BAND_1_5_PER_10000`). The recorded class is the one the midpoint rule gives; that is now said explicitly in the record's `notes` rather than left as an unexplained choice.

**Substantiated the three knowledge-gap discussions with cited evidence** instead of leaving them as unsupported prose. Each now carries the quotes it rests on — the meta-analytic 36%-of-asymptomatic-nerves figure, the authors' own inference from the sex difference in severe contact, the "no satisfactory animal models" statement, and the rodent model's sustained-threshold-drop readout that is the actual behavioural mismatch. Weighted compliance went from 98.3% to 100%.

**Bound the autonomic phenotype more precisely.** It was on `HP:0000632 Lacrimation abnormality`, the parent term, which covers decreased lacrimation too. Ictal tearing is `HP:0031731 Increased tear production`.

`datasets:` is still empty and is the only remaining compliance gap. I did not fill it: searching a dataset repository for this disease would go through disease-name matching with no causal gene to anchor on, which is exactly the Named Entity Confusion path `CLAUDE.md` warns about, and I would rather leave the slot empty than add an accession I cannot vouch for as being about the right condition.

Validation after all three pushes: `just validate-disorders` passed; 74 of 77 snippets verified against cache (3 skipped by prefix); `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-environmental-evidence`, `check-stubs`, `validate-history`, and `check_reference_titles.py --against-ref origin/main` all clean.

sierra-moxon:
Ran a red-team pass over my own entry before waiting on review, and it found real problems. All are now fixed in one commit. The branch is settled and I have stopped pushing.

**Evidence that did not support its claim**

- The oxcarbazepine `target_mechanisms` link asserted a sodium-channel mechanism but quoted a sentence about first-line *status*. It now quotes the mechanism sentence from the same review.
- The Stereotactic Radiosurgery treatment rested entirely on one EAN sentence — "patients should be offered surgery if pain is not sufficiently controlled medically" — which never mentions radiosurgery. It now cites PMID:37144649 (163 patients, 85% initial relief with a 25-day median onset, falling to 62.5% at final follow-up) and PMID:38385643 for the glossopharyngeal subtype. The EAN sentence is kept but regraded `PARTIAL`, with the reason stated.
- Two snippets were pronoun-only — "These alterations…", "They occurred in 25 out of 67 patients…" — with the referent sitting outside the quote. Both are now contiguous two-sentence quotes that carry their own referent.
- The glossopharyngeal subtype description claimed vagal reflex arcs can produce bradycardia and syncope, with nothing in the cache supporting it. That claim is now a `Cardiovascular` phenotype scoped to the subtype, citing PMID:27102416 including the 14-second ECG pause.

**An ontology binding contradicted by the entry's own source**

The trigger-evoked phenotype was bound to `HP:0012533 Allodynia`. PMID:27306631 — the paper that phenotype cites — separates the two explicitly: confined trigger zones and post-attack refractory periods are what distinguish triggered neuralgia from mechanical allodynia. HPO has no term for trigger-zone-evoked paroxysm, so the binding is removed rather than replaced, the free-text `preferred_term` kept, and the distinguishing quote added as evidence so the next curator does not re-bind it. Same reasoning for the laterality phenotype, which was on `HP:0033345 Neuralgia` — a term that restates the pain concept without expressing unilaterality.

**A systematic misgrading**

PMID:39217283 is a cadaveric dissection study with no patients and no clinical outcomes. It was graded `HUMAN_CLINICAL` in both places it appears. Now `OTHER`, and `PARTIAL` rather than `SUPPORT` on the mechanism node, since mapping anatomy is not showing that entrapment causes the disease. `check-snippet-grading` could not catch this because both uses were consistently wrong.

**Additions**

- `animal_models`: the rat infraorbital constriction injury, carrying both a `PARTIALLY_RECAPITULATES` link to the sodium-channel node and a `FAILS_TO_RECAPITULATE` link to the paroxysmal-pain node with its own readouts. The `HUMAN_MODEL_MISMATCH` discussion now has a structural counterpart instead of existing only as prose.
- `clinical_trials`: NCT06620172 and NCT05491915, the latter relevant to the occipital subtype.
- `therapeutic_agent` bindings for the intravenous (fosphenytoin, lidocaine) and botulinum treatments.
- `therapeutic_modality` corrected from `SMALL_MOLECULE` to `OTHER` for botulinum toxin, which is a ~150 kDa protein, and for the occipital nerve block, which is a procedure.
- PMID:25616608 was cited in evidence but missing from the top-level `references:` block.
- The history record said the `Trigeminal_Neuralgia` stub was retained. The PR deletes it. Corrected, along with the stale snippet counts.

**Deliberately not changed**

`datasets:` is still empty, and it is the only remaining compliance gap. Filling it would mean disease-name matching against a dataset repository with no causal gene to anchor on, which is the Named Entity Confusion path `CLAUDE.md` warns about. I would rather leave the slot empty than add an accession I cannot vouch for as being about this condition.

The two removed ontology bindings show as `term: MISSING` in the compliance report. That is the intended state, and the reason is recorded in each phenotype's `notes`.

**Validation after the final commit**

`just validate-disorders` passed. 92 of 95 snippets verified against cache, 3 skipped by prefix. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-environmental-evidence`, `check-stubs`, `validate-history`, and `check_reference_titles.py --against-ref origin/main` all clean, with no new baseline entries anywhere. Weighted compliance 99.1%.

Earlier review runs on this PR were all cancelled by my own pushes rather than finding anything — I was pushing while they were in flight. The branch is now final and a single review run is dispatched.

sierra-moxon:
Thanks — the blocking finding was right, and the diagnosis of *why* it was blocking was the useful part: every other omission here was argued, so this one read as an oversight rather than a decision.

## Blocking: `genetic:` section

Took the first option and added the block, because the alternative would have thrown away real human genetic evidence that this PR had already cached.

`C8B` (`hgnc:1353`) and `MFGE8` (`hgnc:7036`), both `relationship_type: SUSCEPTIBILITY`, evidence graded `PARTIAL`. The caveats are recorded per-gene and again in the entry `notes`, because the headline "significant association" is weaker than it sounds:

- It is a **candidate-gene** analysis — 175 SNPs across 17 genes picked from the same group's earlier proteomic work — not a genome-wide scan. So `p = 0.00016` is nowhere near the 5×10⁻⁸ a GWAS hit would need, and it carries no genome-wide multiple-testing correction.
- Effect sizes are small: OR 1.357 and 1.313. That is population-level risk modification, not something that explains an individual case.
- One cohort, unreplicated.

Neither gene is wired into the pathograph. The proposed biology is complement activation and neuroinflammation, which this entry does not model, and whether a neuroinflammatory signal *drives* the disease or simply *follows* nerve injury is not established — so I recorded that as an open discussion rather than drawing an edge in either direction.

## Non-blocking suggestions — all six actioned

- **STIM1–ORAI1 SOCE** is now a `KNOWLEDGE_GAP` discussion. I graded the human-facing claim `COMPUTATIONAL`/`PARTIAL` rather than accepting the paper's "underscoring its pivotal role", because that claim rests on reanalysis of deposited expression data, not a measurement in patient tissue. The rest is in the same rat constriction model this entry already flags as a poor human proxy.
- **GPN comparative surgical study** now cited on the ablation treatment. It is a good fit: MVD gave significantly higher complete remission at 48 weeks at the cost of longer stay, longer operative time, and higher cost — which is exactly the trade-off the entry models structurally, with decompression acting on the compression node and ablation acting only on the discharge node.
- **`progression:`** added. The more interesting of the two records is the negative one: attack frequency and duration do *not* escalate with disease duration, refuting the intuitive model. That is now stated, with a note telling the next curator not to add a severity-escalation edge without evidence.
- **`clinical_burden: HIGH`**, backed by PMID:26174508 rather than the thinner cached burden review — a 3,273-patient nationwide cohort giving aHR 2.85 for depression, 2.98 for anxiety, 2.17 for sleep disorder. I kept the authors' negative result for schizophrenia and bipolar disorder in the quote so the claim is not read more broadly than the data support.
- **PMID:11803297 → `OTHER`** on all six evidence items. You are right that a hypothesis review with no primary data is not `HUMAN_CLINICAL`.
- **`GO:0009612` removed** from the compression node. On looking again this is more than a mild stretch: the term names a *cell-level response*, and the node is the mechanical insult itself — the vessel indenting the root — not the nerve's response to it. The anatomy is carried by `locations`, and the reason for the non-binding is recorded.

## One thing your review could not have caught

Checking my own work on the new DOI-referenced evidence, I found three snippets that did not match their caches: the DOI caches use U+2009 thin spaces around comparison and equals operators (`P **thin-space** < **thin-space** 0.05`, `N = 555`) where I had typed ASCII spaces.

These would have shipped unnoticed. `linkml-reference-validator` **skips `DOI:` references by prefix** — the run reports "98/109 verified, 11 skipped by prefix", and every DOI snippet in this entry is inside that 11. All three are now corrected to the exact cache characters, and I wrote a throwaway script to confirm all 13 prefix-skipped snippets are exact substrings of their caches. Worth knowing generally: for `DOI:`-cited evidence, the snippet gate is not actually running.

## Two calls I am holding

**`datasets:` stays empty.** This is the only remaining compliance gap. With no causal gene to anchor on, a dataset search here resolves through disease-name matching, which is the Named Entity Confusion path §2b describes — searching "trigeminal neuralgia" would return whatever the name partially matches, and I could not vouch for relevance. An accession that resolves but is about the wrong condition is worse than an empty slot.

**Occipital neuralgia stays off the ephaptic chain.** The demyelination-and-ephaptic evidence comes from cranial nerve roots taken at surgery. The greater and lesser occipital nerves are C2–C3 dorsal-ramus branches whose described compression points are extracranial. Connecting that node to the shared chain would assert a mechanism nothing has shown for those nerves.

Happy to revisit either if you disagree.

## Validation

`just validate-disorders` passed. 98/109 snippets verified by the validator, the remaining 11 verified manually as described above. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-environmental-evidence`, `check-stubs`, `validate-history`, and `check_reference_titles.py --against-ref origin/main` all clean, no new baseline entries. Weighted compliance 99.1% — the residual is the empty `datasets:` and the two intentional non-bindings.

Separately, I checked whether the `prevalence_class`-vs-`rate_per_100000` gap that bit me earlier affects the rest of the KB: 21 records elsewhere have a band that disagrees with their own rate, plus 18 more sitting exactly on a boundary where the convention is ambiguous. I will file that as its own issue rather than widening this PR.

sierra-moxon:
Szmyd is in, and the rebase onto current `main` turned up a schema migration this PR had to absorb.

## The blocking item

You were right that it was missed rather than declined — the second review posted about fifteen minutes before my commit and I hadn't seen it.

`gap_cn_compression_insufficient` now reads that candidates **have** been proposed but none is demonstrated as the necessary second hit, citing Szmyd's list: demyelination, root-entry-zone remodelling, vasculature pathology, nucleus hyperexcitability, white- and grey-matter change, ion-channel disturbance. The old closing sentence was worse than imprecise — as of my own previous commit it contradicted two other places in the same file, since the STIM1/SOCE discussion and the C8B/MFGE8 records both propose exactly such a co-factor.

Szmyd is also now on the central ephaptic node as independent corroboration of the lump/split decision, and the argument is added to `notes`. Your point about it being the stronger citation is the one that changed my mind: the one-disease claim previously rested on a single 2001 source, and a 2022 systematic review of 63 papers that reaches the same grouping *and explicitly includes geniculate neuralgia* is better support for the thinnest subtype than anything else available.

## Non-blocking, all four actioned

- **PMID:30095334** now cited twice on Nervus Intermedius — the subtype definition, and the diagnostic-overlap caveat.
- **`subtype_term` on both unbound subtypes** now carries a `review_notes` search record. For Occipital there is a second reason worth writing down: HP:0012318 exists, but `subtype_term` is bound to the `DiseaseOrSubtypeTerm` enum and an HP term is not a member, so it cannot be used there even though the phenotype record uses it.
- **PMID:27102416 unicode space** fixed. All 113 snippets in the entry are now byte-exact against their caches, not merely whitespace-normalized-equal.
- **GO:0019233** — I checked this one rather than accepting it, and I think the review has the wrong term. `GO:0019233` is *sensory perception of pain*, whose ancestors are `GO:0007600 sensory perception` → `GO:0003008 system process` → `GO:0032501 multicellular organismal process`. That is an organism-level process, so it fits an `ORGANISM`-scale outcome node. The gloss in the review — "detection of chemical stimulus involved in sensory perception of pain", a cellular-level detection process — is a different GO term. Leaving the binding as is; happy to be shown otherwise.

## What the rebase brought in

`origin/main` had moved to `5ae85385bb`, which includes #10003 — `PARTIAL` and `WRONG_STATEMENT` retired from `EvidenceItemSupportEnum`, direction split from a new optional `directness` slot. This PR had 29 `PARTIAL` items and stopped validating.

The KB-wide migration mapped all 11,616 existing occurrences to `SUPPORT` with `directness` absent, reasoning that absent means nobody assessed it. That is right for a bulk backfill and wrong here: these 29 were authored this week and every one already had its directness reasoning written into `explanation`, so leaving the slot empty would have thrown away an assessment that was actually made. I classified them individually — 16 `INDIRECT` (non-human model systems, anatomy cited for a causal step, association cited for predisposition, a guideline indication cited for a mechanism) and 13 `DIRECT`.

Five needed more than a word swap. They had justified `PARTIAL` with *strength* caveats — an eight-patient series, a retrospective single-centre design — and the new schema deliberately does not grade strength, on the grounds that it needs study design and replication which the evidence item does not hold. A mechanical replacement would have produced non sequiturs like "DIRECT because it is a retrospective single-centre comparison". Those now state the directness reason properly and keep the strength caveat as a plain observation.

Rebase was clean. The `cache/` diff is 29 lines, all additions, zero removals.

## Validation after rebase

`just validate-disorders` passed. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-environmental-evidence`, `check-stubs`, `check-term-cache-integrity`, `validate-history`, and `check_reference_titles.py --against-ref origin/main` all clean, no new baseline entries.

## Two issues to file

Both are silent-gap findings from this PR that belong outside it, and I will file them once this is green:

1. **`prevalence_class` is never cross-checked against `rate_per_100000`.** Three of my four bands were off by a decade and every gate passed. Across the KB, 21 records disagree with their own rate, plus 18 sitting exactly on a band boundary where the convention is ambiguous.
2. **`DOI:` is in `skip_prefixes`, so DOI-cited snippets are never checked.** I confirmed this by injecting a fabricated DOI snippet into a scratch copy: `just validate` reported "All validations passed". The "N verified, M skipped by prefix" line reads as reassurance while covering an unvalidated set.

PR #10023, opened 2026-08-29
curate: Add Papillon-Lefevre Disease (MONDO:0009490, CTSC)

### LINKED ISSUE #10015: Curate Papillon-Lefevre disease (MONDO:0009490)
Curate a dismech entry for **Papillon-Lefevre disease** ([MONDO:0009490](https://monarchinitiative.org/MONDO:0009490)) — an autosomal recessive cathepsin C (CTSC) deficiency causing diffuse palmoplantar keratoderma together with severe early-onset periodontitis and premature loss of both dentitions.

Initial read: this looks like a **DISEASE**. MONDO records no descendants, and the mechanism is a single conserved chain — biallelic CTSC loss of function, failure to proteolytically activate the neutrophil serine proteases (elastase, proteinase 3, cathepsin G), and the resulting dysregulated periodontal host response plus epidermal keratinization defect. Haim-Munk syndrome and prepubertal aggressive periodontitis are allelic CTSC conditions and may belong as `has_subtypes` rather than separate entries.

Stub: `stubs/Papillon-Lefevre_Disease.yaml`

matentzn:
Adds `kb/disorders/Papillon-Lefevre_Disease.yaml` — the autosomal recessive disease of biallelic cathepsin C (CTSC) loss of function, with diffuse palmoplantar keratoderma and an aggressive periodontitis that destroys both dentitions.

Closes #10015

## What the entry says

The mechanism has two branches that this entry deliberately keeps apart, because the literature says they belong apart.

**The periodontal branch is demonstrated end to end.** Cathepsin C activates the three neutrophil serine proteases; in patients none of the three has any activity. But there is a second, less obvious step: the proteases are not present-but-inactive, they are *gone*. Proteome analysis found them absent from patient azurophil granules, and bone-marrow work localised where that happens — synthesis and sorting are normal in immature myeloid cells, and the proteins disappear only in mature neutrophils. That is curated as its own node, because it changes what a therapy would have to do: restoring the activating cleavage would not help if the substrate has already been degraded. Downstream, two oxygen-independent antimicrobial functions fail together (LL-37 generation, and degradation of the *A. actinomycetemcomitans* leukotoxin), which is the best available account of why the anaerobic periodontal pocket is the site that fails.

**The cutaneous branch has no substrate at all.** It is graded `HYPOTHETICAL`, and is kept as a separate branch rather than downstream of the periodontal one, on the strength of two facts: skin and periodontal severity are uncorrelated across 47 patients, and the keratoderma appears in the first three years of life, before the deciduous dentition is destroyed. The patient-fibroblast autophagy and lysosomal-permeabilisation work is carried as a `PROVISIONAL` node — it has a rescue experiment behind it, but it was done in dermal fibroblasts rather than keratinocytes.

Ten pathophysiology nodes, twelve phenotypes, Haim-Munk syndrome as an allelic `has_subtypes` entry (the defining study found a mutation in the *same CTSC codon* as a classical PLS family), and three discussions.

## The mismatch worth reading

The `HUMAN_MODEL_MISMATCH` discussion is the reason this disease is interesting. Mouse knockouts assign the neutrophil serine proteases a crucial role against *S. aureus*, *E. coli*, *Candida* and *Klebsiella*. Humans who lack the activity of all three are not correspondingly infection-prone, and one study states outright that the proteases are dispensable for human immunoprotection.

That claim is flagged rather than endorsed. PLS patients *do* have a severe infection-linked pathology — it is just confined to the periodontium and, occasionally, the liver. A more careful reading is that the requirement is site-specific rather than absent, and that systemic-challenge mouse models are not built to detect a defect that manifests as chronic destruction in an anaerobic mucosal pocket. It matters practically, because cathepsin C inhibitors are in clinical development and PLS is the natural experiment those programmes cite for their safety case.

## Curation inputs and what I did not take from them

One Perplexity deep-research run, committed alongside. It was used for narrative framing only:

- **Reference validation**: 10/10 identifiers resolved, one flagged off topic (PMID:19882040, a paper on depigmentation after corticosteroid injection). Not cited here.
- **Term validation**: **27 of 60 checked labels mismatched**, so no ontology CURIE was taken from the report at all. It offered `CL:0000094` as "neutrophil" (the ontology calls it *granulocyte*; neutrophil is `CL:0000775`) and `UBERON:0001838` as "gingiva" (gingiva is `UBERON:0001828`). Every binding here was resolved independently against the committed term caches and OLS.

No GeneReviews chapter exists for this disease — a PubMed search returned nothing — so the Step 3b baseline does not apply. All evidence comes from primary literature found by independent PubMed searching and cached with `just fetch-reference`.

## Validation

- `just validate` — passes; **50/50 evidence snippets verified** against cached primary references
- `just validate-terms` — passes
- `just validate-history` — passes

Five snippets and ten reference titles initially failed on the accented spelling of the eponym (sources use both a grave and an acute accent, in different papers), and one on an en-dash inside a percentage range. All were fixed by quoting the cached source exactly, not by relaxing the claim.

Stub `stubs/Papillon-Lefevre_Disease.yaml` deleted.

---

### DIFF
--- a/history/disorders/Papillon-Lefevre_Disease/2026-08-29T082504Z-claude-code-2a4281.yaml
+++ b/history/disorders/Papillon-Lefevre_Disease/2026-08-29T082504Z-claude-code-2a4281.yaml
@@ -1,0 +1,42 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Papillon-Lefevre_Disease
+  path: kb/disorders/Papillon-Lefevre_Disease.yaml
+session:
+  id: 2026-08-29T082504Z-claude-code-2a4281
+  timestamp: '2026-08-29T08:25:04Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10023
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - animal_models
+  - references
+  summary: 'Review round 3: fidelity/evidence mismatch on the DPPI-null mouse link
+    (PR 10023)'
+  details: |
+    Round 3 fixed a fidelity claim that its own evidence did not support, and the fix added a mechanism link the entry should have had from the start.
+
+    The defect. The animal_models link targeting Failure of Neutrophil Serine Protease Zymogen Activation was graded RECAPITULATES at HIGH fidelity, but every piece of evidence on it - and its readout - measured granzymes A and B in cytotoxic lymphocytes. That is a different protease family in a different compartment from the elastase, cathepsin G and proteinase 3 node the link points at. The link's own limitations field said so, which makes it the same prose-contradicts-data pattern as round 1 item 4, one slot over. The concrete harm the reviewer identified is the readout: readouts are written to be self-describing so they can be lifted out of their link, and lifted out, a granzyme measurement filed against a neutrophil-serine-protease node reads as a claim nobody made.
+
+    The fix, and a reviewer PMID error found on the way. The reviewer offered two options and recommended fetching the primary DPPI-null neutrophil-protease paper, giving PMID:12208867. That identifier resolves to a sitosterolemia paper about ABCG5 and ABCG8 trafficking, with no DPPI content; the cache file was deleted rather than committed. The intended paper is PMID:11827996, Adkison et al., J Clin Invest 2002, found by searching the title the reviewer described. It establishes that DPPI is required for the full activation of cathepsin G, neutrophil elastase and proteinase 3 - the same three enzymes in the same compartment as the human node - so RECAPITULATES at HIGH fidelity is now supported by evidence that measures what the link claims. The limitations note that full activation is the paper's own wording and leaves open whether residual activity remains, and that the model's functional consequences are inflammatory rather than infective.
+
+    This is the second reviewer PMID misattribution in this curation batch; the first was PMID:15316526 on the CVID4 entry, identified as a BAFF-R mutational analysis when the cached record is a CVID lung-disease study. Both were caught by reading the cache before citing. Worth recording as a pattern rather than two incidents: a suggested identifier is a lead, from a reviewer exactly as from a deep-research report.
+
+    The addition. Taking the reviewer's non-blocking suggestion, a third modeled_mechanisms link records the divergence on the neutrophil arm, which is arguably more interesting than the granzyme one. In the mouse the proteases are present and unprocessed; in human PLS neutrophils they are absent, degraded after normal synthesis and sorting. So cathepsin C's chaperone-like role protecting the proteases from degradation - separate from activating them - is a human finding this model does not show, and a therapy restoring only the activating cleavage would be sufficient in the mouse and insufficient in the patient. Graded FAILS_TO_RECAPITULATE at LOW fidelity, with limitations stating plainly that the divergence is inferred across two studies rather than measured head to head, since no cached reference reports protease abundance in DPPI-null mouse neutrophils.
+
+    Second non-blocking item, and a correction to my own reasoning. The Pham human measurement attached to the granzyme FAILS_TO_RECAPITULATE link was graded REFUTE. That was wrong and the reviewer was right that it contradicted its own explanation. supports is relative to the claim the evidence is attached to, and that link's claim is that the model fails to recapitulate the human outcome - which the measurement is exactly the evidence for. Regraded SUPPORT, with the reasoning written into the explanation so the same mistake is visible next time. The same sentence is graded SUPPORT on the discussion, so the two are now consistent.
+
+    A snippet note. The most direct sentence naming all three proteases carries bracketed abbreviations, and the reference validator strips bracketed spans before matching unless conf/reference_validator_config.yaml lists a literal_bracket_patterns entry. A neighbouring bracket-free sentence is used at link level instead, with the explanation saying why, and the readout carries the three enzyme names.
+
+    Validation after changes. Schema validation passes; 68 of 68 evidence snippets verified, up from 66; and the full ungated gate set was run this time - folded-hyphens, snippet-length, title-snippets, snippet-grading - alongside term validation, entity-ref and duplicate-key checks. The model-link pytest lanes pass for this file; the one failure is the pre-existing Polycystic_Kidney_Disease_3 defect on main, filed as issue 10028.
--- a/kb/disorders/Papillon-Lefevre_Disease.yaml
+++ b/kb/disorders/Papillon-Lefevre_Disease.yaml
@@ -81,6 +81,8 @@ references:
   title: "Papillon-Lefèvre syndrome: correlating the molecular, cellular, and clinical consequences of cathepsin C/dipeptidyl peptidase I deficiency in humans."
 - reference: PMID:10411926
   title: "Dipeptidyl peptidase I is required for the processing and activation of granzymes A and B in vivo."
+- reference: PMID:11827996
+  title: "Dipeptidyl peptidase I activates neutrophil-derived serine proteases and regulates the development of acute experimental arthritis."
 - reference: PMID:39164687
   title: "Cathepsin C in health and disease: from structural insights to therapeutic prospects."
 - reference: PMID:26203280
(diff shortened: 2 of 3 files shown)


matentzn:
Thanks — all five IMPORTANT items are fixed, plus four of the five suggestions. Pushed in c405e713c.

**The three term swaps.** All taken as suggested: `GO:0008239` dipeptidyl-peptidase activity, `HP:0006357` premature loss of permanent teeth, `NCIT:C63714` Periodontal Scaling and Root Planing. The GO one was the substantive miss — cathepsin C being an *exopeptidase that removes a dipeptide* is the whole reason the NSP pro-dipeptides go uncleaved, and binding to the cysteine-peptidase parent threw that away.

**The prose/enum contradiction (item 4)** is worth naming as a failure mode rather than just fixing. The enum rejected `SUPPORTED` and `EMERGING` at validation time, so I corrected the *field* — but the prose I had written alongside kept the words I first reached for, and nothing checks that the two agree. Fixed in all three places.

**The NK/granzyme node (item 5)** — you were right, and for the reason you gave. "Neutrophil serine proteases are dispensable for human immunoprotection" is a claim about *neutrophil* proteases; a PLS patient has also lost granzyme activation, and conflating the two makes the clinical mildness look more surprising than it is. The node is `PROVISIONAL` and carries a second, `PARTIAL` evidence item from the same review recording that the wider immunological literature in PLS is inconsistent — which seemed the honest qualifier on a claim about "the first consistent immune dysfunction". The top-level description now carries the caveat too.

**Suggestions taken:** `HP:0007447` for the keratoderma; evidence on the `HUMAN_MODEL_MISMATCH` discussion itself, both sides of it, rather than asserting them in `rationale` prose; a prosthetic/implant rehabilitation treatment (including that implants are contraindicated in growing individuals because they behave like ankylosed teeth); and a `notes:` line on aggressive periodontitis type 1 as a third allelic CTSC condition, with why it is not a subtype here.

**One suggestion I did not take, and why.** The `animal_models` entry for the Ctsc-null mouse. No primary Ctsc-null paper is cached in this PR — the only mouse evidence available is a secondary citation inside PMID:16926422, and that sentence is about *protease-specific* knockouts (elastase, cathepsin G), not a Ctsc null. Building an `animal_models` entry whose sole support is a review's citation of work I have not read seemed worse than leaving it out, so the claim went into the discussion's `evidence` instead, where its secondary character is stated. If someone fetches a primary Ctsc-null paper later, that entry is the right home for it.

Validation after the changes: `just validate` passes, **56/56 snippets verified** (up from 50), term / entity-ref / duplicate-key / snippet-grading checks all pass.

---

matentzn:
You're right, and this is a good catch — I asserted a mechanism the entry's own cached evidence base refutes. Fixed in 1207dfc85, and the fix ended up strengthening the entry rather than just patching it.

**What went wrong, plainly.** Round 1 pointed me at PMID:26203280 — a narrative review saying impaired NK cytotoxic function is "the first consistent immune dysfunction in PLS". I added a node asserting the arm fails, with `modifier: DECREASED`, and edited the top-level description to say patients have lost two effector arms. I acted on the pointer without reading the primary literature behind it, which is the same error as citing a review's summary in place of a measurement — and it is exactly the error the rest of this entry is written to avoid.

Two things let it through: my shortcut, and the fact that **PMID:15585850 sat in this PR cached and uncited through a full review round on both sides**. Thanks for saying that second part out loud rather than only flagging mine.

**The fix.** The node is renamed `Granzyme Activation in Cytotoxic Lymphocytes`, so its name asserts no direction. `mechanism_confidence: HYPOTHETICAL`. The `DECREASED` modifier is gone — the GO term is carried with no directional claim. The description leads with the human measurement, then gives the review's contrary position, and says plainly that a direct measurement in patients outranks a review's summary of prior reports. Two `REFUTE` items from Pham, two `PARTIAL` from the review, and the shared premise (that cathepsin C activates granzymes at all) now cited from PMID:39164687 — also cached and uncited, as you noted. The top-level sentence now says the second arm is *contested*, not lost.

**Where the correction paid for itself.** Fetching the primary mouse paper behind the granzyme claim — **PMID:10411926**, Pham & Ley 1999 — resolved the `animal_models` suggestion you have now made twice. In DPPI-null mice granzymes A and B are present in normal amounts but *retain their prodipeptide domains and are inactive*, with target-cell apoptosis as defective as a perforin-null. So the model:

- `RECAPITULATES` the zymogen-activation node at `HIGH` fidelity — it reproduces the processing chemistry exactly, and gives a nice contrast with the human neutrophil finding where the proteases are *degraded* rather than left as inactive zymogens;
- `FAILS_TO_RECAPITULATE` the granzyme node at `LOW` fidelity, with Pham's human measurement as its `REFUTE` evidence.

The `HUMAN_MODEL_MISMATCH` discussion now attaches to both, and carries Pham's second relevant sentence as you suggested — PLS neutrophils do not uniformly fail to kill *S. aureus* and *E. coli*, which separates a claim about human neutrophil biology from a claim about disease severity. It also now records the **direction** of the error, which I think is the practically useful part: the mouse over-predicts harm, so a safety case for cathepsin C inhibition built on it is conservative, while a mechanistic efficacy case built on it is not.

One note on my round-1 reasoning: I declined the animal-model suggestion then because no primary Ctsc-null paper was cached and I did not want to build an entry on a review's citation of work I had not read. I still think that was the right principle — the mistake was stopping there instead of fetching the paper.

Validation: `just validate` passes, **66/66 snippets verified** (56 after round 1, 50 originally), term / entity-ref / duplicate-key checks pass, and the model-link pytest lanes pass for this file. The single failure in those lanes is a pre-existing `main` defect in `Polycystic_Kidney_Disease_3.yaml`, filed separately as #10028.

---

matentzn:
Fixed, and I took your second option rather than the downgrade — pushed in b12150cdc.

**Your diagnosis is right and worth restating**: the link's own `limitations` field argued against its own `fidelity` value in the slot beside it. That is the round-1 prose-vs-data pattern one slot over, and the concrete harm you named is the sharp part — readouts are written to be self-describing so they can be lifted out, and lifted out, a granzyme measurement filed against a neutrophil-serine-protease node reads as a claim nobody made.

**One thing to flag: your suggested PMID is wrong.** `PMID:12208867` resolves to *"Coexpression of ATP-binding cassette proteins ABCG5 and ABCG8 permits their transport to the apical surface"* — a sitosterolemia trafficking paper with no DPPI content. I deleted the cache file rather than committing it.

The paper you described is **PMID:11827996** (Adkison et al., *J Clin Invest* 2002), found by searching the title. It says exactly what you predicted: *"we generated a mouse deficient in dipeptidyl peptidase I (DPPI) and established that DPPI is required for the full activation of CG, NE, and PR3"* — same three enzymes, same compartment as the node. So `RECAPITULATES` / `HIGH` now rests on evidence that measures what the link claims. The `limitations` record that "full activation" is the paper's own wording and leaves residual activity open, and that the model's functional consequences are inflammatory (arthritis protection, reduced zymosan response) rather than infective.

This is the **second reviewer PMID misattribution in this batch** — `PMID:15316526` on #10025 was identified as a BAFF-R mutational analysis when the cache holds a CVID lung-disease study. Both were caught by reading the cache before citing. I mention it not as a complaint but because it is the same lesson the deep-research handling in these entries is built around: a suggested identifier is a lead, wherever it comes from.

**I took your third-link suggestion, and you were right that it is the more interesting divergence.** In the mouse the proteases are present and unprocessed; in human PLS neutrophils they are *absent*, degraded after normal synthesis and sorting. So cathepsin C's chaperone-like role — protecting the proteases from degradation, separate from activating them — is a human finding this model does not show, and a therapy restoring only the activating cleavage would be sufficient in the mouse and insufficient in the patient. `FAILS_TO_RECAPITULATE` / `LOW`, with `limitations` stating plainly that the divergence is inferred across two studies rather than measured head to head, since no cached reference reports protease abundance in DPPI-null mouse neutrophils.

**The `REFUTE` grade was my error and your correction is right.** `supports` is relative to the claim the evidence is attached to; that link's claim is that the model *fails* to recapitulate, and the Pham measurement is exactly the evidence for it. Regraded `SUPPORT`, with the reasoning written into the `explanation` so the same mistake is visible next time. It is now consistent with the same sentence graded `SUPPORT` on the discussion.

One snippet note: the sentence naming all three proteases carries bracketed abbreviations, which the reference validator strips before matching unless `conf/reference_validator_config.yaml` lists a `literal_bracket_patterns` entry. I used a bracket-free neighbouring sentence at link level and put the three enzyme names in the readout, with the explanation saying why.

Validation: `just validate` passes, **68/68 snippets verified** (66 → 68), and the full ungated set — folded-hyphens, snippet-length, title-snippets, snippet-grading — plus terms, entity-refs and duplicate-keys all pass. Model-link pytest lanes pass for this file; the one failure is the pre-existing `Polycystic_Kidney_Disease_3` defect on `main` (#10028), same as you saw.

---

PR #10024, opened 2026-08-29
curate: Add Primary Ciliary Dyskinesia 9 (MONDO:0012906, DNAI2)

### LINKED ISSUE #10016: Curate primary ciliary dyskinesia 9 (MONDO:0012906)
Curate a dismech entry for **primary ciliary dyskinesia 9** ([MONDO:0012906](https://monarchinitiative.org/MONDO:0012906)) — CILD9, biallelic loss of the outer dynein arm intermediate chain DNAI2, producing immotile respiratory cilia with or without situs inversus.

Initial read: this looks like a **DISEASE**, following the precedent already set by `kb/disorders/Primary_Ciliary_Dyskinesia_30.yaml` (ODAD3) sitting alongside the parent `Primary_Ciliary_Dyskinesia.yaml`. The DNAI2 pathograph is gene-specific down to the ultrastructure — outer dynein arm loss without inner arm involvement — which is what distinguishes it from the radial-spoke and dynein-preassembly forms.

Stub: `stubs/Primary_Ciliary_Dyskinesia_9.yaml`

matentzn:
Adds `kb/disorders/Primary_Ciliary_Dyskinesia_9.yaml` — CILD9, the PCD caused by biallelic loss of DNAI2, an outer dynein arm intermediate chain and the human relative of *Chlamydomonas* IC69.

Closes #10016

Follows the precedent of `Primary_Ciliary_Dyskinesia_30.yaml` (ODAD3) sitting alongside the umbrella `Primary_Ciliary_Dyskinesia.yaml`.

## Why this gene is separable from its neighbours

DNAI2 loss removes **both** outer-arm heavy chains, DNAH5 *and* DNAH9, from every mutant axoneme. Outer dynein arms are not uniform along the cilium — proximal and distal complexes differ in heavy-chain composition — and the intermediate-chain genes are not interchangeable in what they take down with them: DNAI2 and DNAH5 loss disrupts proximal and distal complexes alike, where DNAI1 loss mainly disrupts the proximal ones.

The hierarchy runs the other way too. DNAI2 is itself absent from the axoneme in DNAH5 patients, and absent distally in DNAI1 patients — which is what makes the outer arm a co-assembling complex rather than a chain of separable steps.

Eight pathophysiology nodes, branching three ways from beat failure into airway disease, laterality, and sperm motility.

## What this entry does *not* claim

DNAI2 is rare — the defining screen found it in two of 105 further unrelated PCD families — and there is **no DNAI2-specific natural history**. Every prognostic and frequency statement here is borrowed from the outer-dynein-arm defect class, and each one is labelled as borrowed *in place*, not just in a footnote. That is recorded as a `KNOWLEDGE_GAP` with a specific reason it matters: if the extent of outer-arm loss along the cilium affects clearance quantitatively, DNAI2 and DNAI1 patients should not have identical airway courses, and nobody has looked.

Two frequencies for neonatal respiratory distress (67% and 82%) are recorded side by side rather than averaged, because the cohorts differ.

The single reported co-occurrence of hydrocephalus with a DNAI2 microdeletion — which the case report proposes as a new genetic risk factor — is graded `VERY_RARE` with the reasoning attached, so one case report is not mistaken for an established genotype–phenotype correlation.

**One practical diagnostic point**: one of the four reported DNAI2 alleles is a 6.9 kb intragenic deletion found by array CGH. Copy-number analysis is therefore curated as a separate `diagnosis` entry from panel sequencing — a negative panel in a patient with outer dynein arm defects is not a negative DNAI2 result.

## Curation inputs

One Perplexity deep-research run, committed alongside. It contributed narrative framing and little else:

- **Reference validation**: only **3 citations in total**, all resolving, all on topic.
- **Term validation**: **20 of 49 checked labels mismatched**, so no CURIE was taken from it. It offered `HP:0005938` as "primary ciliary dyskinesia" (HPO calls it *Abnormal respiratory motile cilium morphology*) and `HP:0011107` as "chronic cough" (HPO calls it *Recurrent aphthous stomatitis*). Bindings here were resolved against the committed term caches, several by reuse from the sibling ODAD3 entry.

The GeneReviews PCD chapter (PMID:20301301) is cited and tagged. Its PubMed record is a *scope statement* — "the purpose of this overview is to…" — rather than a clinical abstract, so it carries no quotable phenotype text and is not used as an evidence source for individual phenotypes. The entry notes say so explicitly, so the absence of GeneReviews evidence items reads as a property of the record rather than an omission.

## Validation

- `just validate` — passed on the first run; **32/32 evidence snippets verified** against cached primary references
- `just validate-terms` — passes
- `just validate-history-all` — passes (7,590 records)

Stub `stubs/Primary_Ciliary_Dyskinesia_9.yaml` deleted.

---


matentzn:
All five blocking items are fixed, plus three of the four non-blocking suggestions. Pushed in a4e6a19d0 and abf5d8ea6.

**Item 1 (otologic phenotypes)** — fair catch, and the kind that is embarrassing to have missed: my own description said "recurrent otitis and sinusitis", the source was already cached, and the sibling ODAD3 entry curates both from exactly those rows. Added `HP:0000403` (FREQUENT) and `HP:0000405` (OCCASIONAL).

**Item 2 (ORPHA:244, and the bronchiectasis disagreement)** — ORPHA:244 is now the frequency backbone for the otologic phenotypes and is in `references`. The disagreement is reconciled *in place* rather than silently: both records are carried on the bronchiectasis phenotype, the Orphanet row graded `PARTIAL` because it supports the phenotype's occurrence while contradicting its frequency, and the description says why I kept FREQUENT — the prospective paediatric series imaged every participant by CT and found a median of three affected lobes across all 118, which is not compatible with a sub-30% rate. My reading is that Orphanet's grade reflects clinically diagnosed bronchiectasis across all ages and settings, where the cohort figure reflects systematic imaging in a referral population. Both are now visible to a reader who wants to disagree with that.

**Item 3 (diagnosis)** — added nasal NO and high-speed videomicroscopy from PMID:38891105, including that review's explicit "no single gold standard" statement, and immunofluorescence as its own entry. You were right that IF is the one that earns separate treatment here: DNAI2 loses DNAI2 *plus both* outer-arm heavy chains along the whole axoneme, DNAI1 loses DNAI2 only distally, and DNAH11 has a proximal-only partial reduction that conventional TEM calls normal. That is the entry's central argument, made operational.

**Item 4 (uncited cached references)** — all three now have homes. PMID:26909801 is the primary source for the proximal/distal heterogeneity, which was previously resting on a single Loges sentence; PMID:11713099 anchors the DNAI1 contrast; PMID:38891105 supports diagnosis and management.

**Item 5 (treatments)** — added a single supportive-care entry linked to the clearance and chronic-infection nodes, with the scoping decision written into the entry rather than left as silence: no trial has enrolled by genotype, the umbrella entry holds the detailed management block, and duplicating it here would create two places to maintain the same evidence and imply a genotype-specific practice that does not exist. What *is* DNAI2-specific is prognostic, not therapeutic, and that already sits on the airway-damage node.

**Non-blocking, taken:**
- The `evidence_source` inconsistency within PMID:11153919 — both quotes are from the same introduction, both are now `OTHER`.
- The medaka model. I verified the PMID before adding, as you suggested: **PMID:20709053** resolves to a real *Dev Biol* paper whose abstract states the Kupffer's vesicle findings directly. Added with two `modeled_mechanisms` links and three readouts — and with the limitation the model reports about itself, that medaka has a *second* dnai2 gene acting redundantly in some ciliated tissues, where humans have one. That is graded `PARTIAL` because it qualifies the model's informativeness rather than supporting it.
- Heterotaxy as its own phenotype, kept separate from situs inversus totalis because they are not clinically equivalent, with no frequency asserted since the source does not separate them.
- Two `differential_diagnoses`, covering the other outer-arm/docking genes and the ~30% normal-ultrastructure group.

One extra commit worth flagging: the reference validator resolved PMC full text for PMID:38891105 during validation, upgrading its cache entry from `abstract_only` to `full_text_xml`. I committed that so CI validates against the same file rather than re-fetching.

Validation: `just validate` passes, **48/48 snippets verified** (up from 32), term / entity-ref / duplicate-key checks pass, and the pytest lanes covering entity refs and model-readout targets pass (7,910 tests).

---

PR #10025, opened 2026-08-29
curate: Add Immunodeficiency Common Variable 4 (MONDO:0013284, TNFRSF13C)

### LINKED ISSUE #10017: Curate immunodeficiency, common variable, 4 (MONDO:0013284)
Curate a dismech entry for **immunodeficiency, common variable, 4** ([MONDO:0013284](https://monarchinitiative.org/MONDO:0013284)) — CVID4, biallelic TNFRSF13C (BAFF-R) deficiency, in which loss of BAFF-R blocks the non-canonical NF-kB survival signal that transitional B cells depend on.

Initial read: this looks like a **DISEASE**. `kb/disorders/Common_Variable_Immunodeficiency.yaml` carries BAFF-R only as a `genetic:` entry and has no `has_subtypes` block, and the KB already holds several single-gene inborn-error-of-immunity entries (`CTLA4_Haploinsufficiency`, `TRAF3_Haploinsufficiency`, `SOCS1_Haploinsufficiency`). BAFF-R deficiency has a mechanistically distinct pathograph from generic CVID: a developmental arrest at the transitional-to-mature B cell checkpoint rather than a class-switch/plasma-cell differentiation defect, with the unusual clinical feature that some homozygotes stay largely asymptomatic.

Stub: `stubs/Immunodeficiency_Common_Variable_4.yaml`

matentzn:
Adds `kb/disorders/Immunodeficiency_Common_Variable_4.yaml` — BAFF-receptor deficiency, the CVID form caused by biallelic loss of TNFRSF13C.

Closes #10017

## The lump/split call

Curated as its own entry rather than a `has_subtypes` entry on `Common_Variable_Immunodeficiency`, for two reasons.

The mechanical one: the parent carries BAFF-R only as a `genetic:` entry and has no `has_subtypes` block at all, so there was no subtype structure to extend.

The substantive one: **the pathograph is not the parent's.** Generic CVID fails at the late steps — class-switch recombination and plasma-cell differentiation. This fails at a *survival* checkpoint upstream of both, at the transitional-to-mature transition. The immunoglobulin pattern shows the difference in one line: BAFF-R-deficient siblings had **normal IgA**, which most CVID patients do not, and which matches the mouse observation that mucosal IgA-secreting plasma cells develop without BAFF-R signals. The KB already holds several single-gene IEI entries at this granularity (`CTLA4_Haploinsufficiency`, `TRAF3_Haploinsufficiency`, `SOCS1_Haploinsufficiency`).

## The finding this entry is really about

Of the two homozygous siblings in the defining study, **both** had the full immunological phenotype — the developmental block, low IgG and IgM, no T-independent response to pneumococcal polysaccharide — and **only one had recurrent infections**. The paper says so in its own conclusion: the deletion causes a characteristic immunological phenotype but does not necessarily lead to a clinically manifest immunodeficiency.

That is carried as a `KNOWLEDGE_GAP`, and it shapes several grading decisions in the entry rather than sitting in a corner: the infection node is `PROVISIONAL`, its incoming edge is `INDIRECT_KNOWN_INTERMEDIATES` (whatever compensates sits between the antibody deficit and the clinical phenotype and is unidentified), and the phenotype is `FREQUENT` rather than `VERY_FREQUENT`. It matters practically — an incidentally found homozygous TNFRSF13C null currently cannot be given a prognosis, and the entry should not imply otherwise.

## Two things the entry works to keep separate

**Null alleles vs. the P21R missense variant.** Most papers citing TNFRSF13C in CVID are about P21R, which is a polymorphism found in healthy people and which the functional work classes as a modifier. Six missense variants have been characterised and all six change BAFF-R function, but through different mechanisms and only P21R correlates positively with CVID. The `genetic:` entry states the practical reading: a homozygous null is diagnostic; a heterozygous missense variant is contributory and does not on its own explain a CVID phenotype. Three evidence items carry that, two graded `PARTIAL` because they *qualify* rather than support a causal claim.

**BAFF-driven pathology vs. BAFF-R deficiency.** A sixth pathophysiology node records BAFF-driven B-cell hyperplasia in CVID interstitial lung disease — the same axis driven too hard, in patients who *have* the receptor. It is marked `role: modifier` and its description says plainly that this pathology cannot occur in a BAFF-R-null patient. It is there because a keyword search for this gene returns that literature repeatedly, and it should not be misread as evidence about CVID4.

## Curation inputs

One Perplexity deep-research run, committed alongside, used for framing only:

- **Reference validation**: 7/7 resolved, but only **4 of 7 assessed as on topic**.
- **Term validation**: 8 of 35 checked labels mismatched, including `MONDO:0011107` offered as "common variable immunodeficiency" when MONDO calls that identifier *congenital hypotrichosis with juvenile macular dystrophy*. No CURIE was taken from the report.

No GeneReviews chapter exists for TNFRSF13C or CVID4.

## Validation

- `just validate` — passes; **28/28 evidence snippets verified**
- `just validate-terms`, `just check-duplicate-keys`, `just check-entity-refs`, `just validate-history` — pass

Two fixes worth flagging for reviewers, both generalisable:

1. The term validator rejected `HP:0002850` with the label OLS *search* returned ("Decreased circulating IgM concentration"); the authority's canonical label is **"Decreased circulating total IgM"**. OLS search results are not a safe source for the `label` field even when the CURIE is right.
2. One snippet failed because the source sentence begins "The patient had", not "He had". The paraphrase was mine; it was fixed by quoting the source, not by loosening the claim.

One snippet contains run-together words (`theTNFRSF13Cgene`) because the cached source has the gene symbol in italic markup the cache strips. It is reproduced exactly as cached, with a note on the entry saying so.

Stub `stubs/Immunodeficiency_Common_Variable_4.yaml` deleted.

---


matentzn:
All five IMPORTANT items are fixed, plus three of the four suggestions. Pushed in ce74441c7.

Your closing observation is the one worth keeping: **PMID:19666484 is cached as full text, not as an abstract, and it answers findings 2, 3 and 4 directly.** I curated from the abstract without checking what else the cache file held. That is the generalisable mistake in this round, and it is now the first thing recorded in the history record.

**Item 1 (GO:0033209)** — you're right, and the error is instructive enough to state plainly: BAFF-R is a member of the TNF receptor *superfamily*, which is a structural classification, but its ligand is BAFF (TNFSF13B), and GO defines that term *by the ligand*. Dropped rather than swapped for another approximation, since `GO:0038061` already carried it.

**Item 2 (the diagnostic triad)** — added all three with measured quotes from the cached full text. The marginal-zone one is the valuable one, exactly as you say: it supplies the mechanistic link from the lost MZ compartment to the absent T-independent response that the entry previously only asserted in prose.

**Item 3 (onset)** — removed the mis-bound phenotype; onset is now an `onset` descriptor on the infection phenotype. One deviation from your suggestion, deliberately: I bound `ADULT` with `min_age_years: 37.0` rather than reading "second half of life" as the onset. P1's *first pneumonia was at 37*, against a lifelong history of chronic sinusitis; the paper's "second half of life" sentence is about **diagnosis** (at 57), not onset. Both quotes are recorded, with the distinction stated in the descriptor notes, so a reader can see why the value is not LATE.

**Item 4 (prevalence)** — recorded as `BELOW_1_IN_1000000`, `rate_per_100000: 0.025` (0.0167–0.0333), with the 138-patient screening denominator captured, and the authors' caveat quoted as a `PARTIAL` item. The caveat is the substance of the estimate, not a footnote: a BAFF-R-deficient person who never came to immunological attention is invisible to a CVID-cohort screen.

**Item 5 (the contrast node)** — moved to a `discussions` entry. Your argument settles it: `pathophysiology[]` is exported as *this disease's* mechanism graph, so a node whose own description says the mechanism cannot occur in a BAFF-R-null patient read, to a graph consumer, as an assertion that it does — the exact misreading the node existed to prevent. `role` and `mechanism_confidence` do not carry a negation across an export boundary. That is a good general rule and I'll apply it elsewhere.

**Suggestions taken:** the class-switched memory phenotype now carries a `SUPPORT`-graded item measuring it in both null-allele patients (the P21R item stays for the dissociation point); the autoimmunity mismatch now carries the citable clinical negative, graded `REFUTE`; and the penetrance discussion records the original authors' own modifier search — BTK sequencing and CD40L expression both normal in the discordant pair, which excludes the two obvious X-linked candidates without closing the question.

**One suggestion I did not take, because the premise is wrong.** You identified `PMID:15316526` as "Mutational Analysis of Human BAFF Receptor TNFRSF13C (BAFF-R) in Patients with Common Variable Immunodeficiency" and the primary P21R association source. The cached record for that PMID is **"Granulomatous-lymphocytic lung disease shortens survival in common variable immunodeficiency"** — a retrospective analysis of 69 CVID patients, with no BAFF-R content at all. It is in the cache because the deep-research report's citation list resolved it, not because I fetched it. Citing it as a P21R source would have introduced precisely the wrong-paper attribution this review process exists to catch, so I've left it uncited. Worth flagging back: this is the same failure mode the entry's own notes describe on the ontology side, appearing on the reference side of a review.

The fourth suggestion (antibiotic prophylaxis) I've left out for the reason you gave — generic CVID management, belongs on the parent entry.

Validation: `just validate` passes, **39/39 snippets verified** (up from 28), term / entity-ref / duplicate-key checks pass.

---

PR #10026, opened 2026-08-29
curate: Add IgA Pemphigus (MONDO:0034127)

### LINKED ISSUE #10018: Curate IgA pemphigus (MONDO:0034127)
Curate a dismech entry for **IgA pemphigus** ([MONDO:0034127](https://monarchinitiative.org/MONDO:0034127)) — an autoimmune intraepidermal blistering disease driven by IgA rather than IgG autoantibodies against keratinocyte cell-surface adhesion molecules, presenting as pustules and vesicles with prominent neutrophil infiltration.

Initial read: this looks like a **DISEASE**. The KB already holds sibling entries at this granularity (`Pemphigus_Vulgaris`, `Pemphigus_Foliaceus` is stubbed, `Paraneoplastic_Pemphigus`, `Pemphigus_Erythematosus`), and the IgA/neutrophil axis is what distinguishes IgA pemphigus mechanistically from the IgG forms. The two classical variants — subcorneal pustular dermatosis type (desmocollin 1) and intraepidermal neutrophilic type — look like `has_subtypes` entries on one disease rather than two diseases, since they share the IgA-plus-neutrophil mechanism and differ mainly in target antigen and blister level.

Stub: `stubs/IgA_Pemphigus.yaml`

matentzn:
Adds `kb/disorders/IgA_Pemphigus.yaml` — the pemphigus variant in which the autoantibody isotype is IgA rather than IgG.

Closes #10018

That one substitution changes everything below it. IgA does not drive the classic complement-and-acantholysis route; it recruits neutrophils. So the disease is pustular rather than bullous, the histology is neutrophil-dominant with only slight acantholysis, and the drug of choice is dapsone — which acts on neutrophils, not on antibody.

## The curation judgement worth reviewing

Two effector branches run in parallel from the deposited antibody, and they are graded differently on purpose:

- **Neutrophil recruitment → `ESTABLISHED`**
- **Impaired desmosomal adhesion → `PROVISIONAL`**

Desmocollin 1 is *securely* the SPD-type autoantigen — the transfection experiment is clean, six of six SPD sera bound Dsc1 and none bound Dsc2 or Dsc3. What is not established is that binding it causes the disease. A 2023 review of exactly that question concludes there is no clear evidence for anti-desmocollin pathogenicity, in contrast to the well-established anti-desmoglein case. The histology agrees: acantholysis is slight where present at all.

So an alternative reading is live and is not excluded by anything cited here — that the IgA acts mainly by recruiting neutrophils, and the adhesion molecule it happens to bind is a marker of the disease rather than its mechanism. The grading says so, and a `KNOWLEDGE_GAP` discussion proposes the passive-transfer experiment that would settle it.

## Two evidence items that contradict the claims they attach to

These are deliberate and I'd like a reviewer to weigh in on them:

- The **Acantholysis** phenotype carries a `WRONG_STATEMENT` item — a 2024 case report stating flatly that acantholysis is not a feature of IgA pemphigus, which two other cited sources contradict.
- The **Retinoid Therapy** treatment carries a `REFUTE` item — isotretinoin plus prednisone achieving "only a slight effect after six months".

Both are kept rather than dropped because in a literature of **137 reported patients across 119 publications**, one dissenting source is a meaningful fraction of what is known, and hiding it would make the entry look more settled than the field is.

(Note: this entry uses `PARTIAL` and `WRONG_STATEMENT`, which #10003 proposes to retire. If that lands first I'm happy to migrate.)

## Subtypes, held loosely

SPD type (antigen known: Dsc1) and IEN type (antigen unknown — and not for want of looking; in the same experiment none of seven IEN sera bound any desmocollin) are curated as `has_subtypes` on one disease. The scheme is recorded as revisitable in a discussion, because published cases keep failing it: reactivity to all three desmocollins, to desmogleins, to desmocollins plus BP180; the largest cohort could not classify 8 of 49 cases; one review proposes replacing the scheme with an "IGAP spectrum". The reason not to act yet is that the two-type scheme separates the subtype with a known antigen from the one without, and a spectrum model would dissolve the only hard biological distinction available.

**Practical diagnostic point**: SPD-type IgA pemphigus is clinically and histopathologically *indistinguishable* from classical subcorneal pustular dermatosis (Sneddon-Wilkinson). In the largest cohort all 13 classical SPD cases were immunologically negative and could not be told apart on any other grounds. Immunofluorescence is the only discriminator — and circulating IgA is detectable in only 66.7% of patients, so serology alone is not enough.

## Curation inputs

One Perplexity deep-research run, committed alongside. Its **term validation reported zero terms checked, because the report offered no ontology CURIEs at all.** That is a different failure from the mislabelled CURIEs the other reports in this batch produced, and it means the clean term-validation section carries no information — a reader skimming for `needs_review` would see a clean result and conclude the opposite. Recorded in the entry notes for that reason.

## Validation

- `just validate` — passed on the first run; **36/36 evidence snippets verified**
- `just validate-terms`, `just check-duplicate-keys`, `just check-entity-refs`, `just check-snippet-grading`, `just validate-history` — pass

One reference-validation line is environmental, not a content problem: the validator attempted a full-text download from `derm101.com` and got a TLS error. The snippet verified against the cached abstract.

Stub `stubs/IgA_Pemphigus.yaml` deleted.

---

### DIFF
--- a/history/disorders/IgA_Pemphigus/2026-08-29T081713Z-claude-code-10cdb2.yaml
+++ b/history/disorders/IgA_Pemphigus/2026-08-29T081713Z-claude-code-10cdb2.yaml
@@ -1,0 +1,32 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: IgA_Pemphigus
+  path: kb/disorders/IgA_Pemphigus.yaml
+session:
+  id: 2026-08-29T081713Z-claude-code-10cdb2
+  timestamp: '2026-08-29T08:17:13Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10026
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Fix a folded-scalar hyphen split found by a full gate re-run
+  details: |
+    One folded-scalar hyphen split, found by running the full ungated gate set across all five entries in this curation batch rather than the subset I had been running.
+
+    A line in the adhesion node's description ended in anti-, so the loaded value read anti- desmocollin with a space. Reworded so the compound stays on one line. No reviewer had reached this file's version of the defect; it was found by generalising a fix from the LCA13 entry in the same batch, where the reviewer caught the equivalent problem.
+
+    Recording it because the pattern is the point. check-folded-hyphens, check-snippet-length and check-title-snippets are ungated whole-KB CI steps, specifically so that a kb-only PR cannot skip them, and I had been running only validate, validate-terms, check-entity-refs and check-duplicate-keys after each edit. The defect is invisible in the source file - the YAML reads correctly, and only the loaded scalar is wrong - so nothing but the check finds it.
+
+    Validation after change. Schema validation passes; 40 of 40 evidence snippets verified, unchanged; check-folded-hyphens, check-snippet-length, check-title-snippets, term validation, entity-ref and duplicate-key checks all pass.
--- a/kb/disorders/IgA_Pemphigus.yaml
+++ b/kb/disorders/IgA_Pemphigus.yaml
@@ -386,8 +386,8 @@ pathophysiology:
     expected consequence is loss of intercellular adhesion. That expectation is
     the standard account of the disease and it is what makes it a pemphigus.
 
-    Graded PROVISIONAL, deliberately, because the pathogenicity of anti-
-    desmocollin autoantibodies is not established the way anti-desmoglein
+    Graded PROVISIONAL deliberately, because the pathogenicity of
+    anti-desmocollin autoantibodies is not established the way anti-desmoglein
     pathogenicity is. A 2023 review of the question puts it plainly: beyond a few
     in vitro and in vivo studies there is currently no clear evidence that
     anti-Dsc autoantibodies are pathogenic, in contrast to the well-established


matentzn:
All five IMPORTANT items are fixed, plus all three suggestions. Pushed in 70f55ab87.

**Items 1–3 (rebindings)** — taken as suggested. Two worth commenting on:

- `UBERON:0002027` → `UBERON:0001003`. This was the worst of the three: one of the nodes it sat on says the infiltrate involves "all layers" of the epidermis, so binding it to the *stratum corneum* was wrong in a way that would have propagated into any anatomical query.
- `HP:0006775` → `HP:0020194`. You're right about the arithmetic — the phenotype quotes the 9.5% *gammopathy* figure while binding the term for myeloma, of which only seven cases are reported. That is roughly a two-fold inflation, and `HP:0020194` is an exact match.

**Item 4 (FcalphaRI/CD89)** — added, with two freshly fetched references. This turned out to be the most useful item in the review: it is the step that makes the isotype matter at all, and the entry was asserting the isotype-to-neutrophil logic in prose while the pathograph jumped straight over it. The mechanism as curated is the LTB4 feedback loop — cross-linking CD89 with IgA immune complexes releases the chemoattractant that recruits further neutrophils, a loop that self-terminates against an IgA-opsonised organism and *does not* against a fixed epidermal autoantigen.

One thing I made explicit rather than glossing: the evidence is receptor biology, from human neutrophils and a linear IgA bullous disease mouse model, not from IgA pemphigus patients. The node is graded `ESTABLISHED` on the strength of the receptor work, and says so in its own description and in the entry notes, so a reader can see the basis for applying it here (shared isotype, shared neutrophil-rich histology) rather than assuming a disease-specific experiment exists.

**Item 5 (phenotypes disconnected)** — six phenotypes now have incoming edges. The load-bearing one is exactly the one you named: the edge from the `PROVISIONAL` adhesion node to **Acantholysis**. That puts the entry's central uncertainty into the graph rather than leaving it in prose — a `PROVISIONAL` node reaching a phenotype curated `OCCASIONAL` whose evidence includes a source that denies it occurs at all. Pruritus is reached by an `INDIRECT_UNKNOWN_INTERMEDIATES` edge, since nothing cited explains how the lesion produces the itch.

**Suggestions, all taken:**
- Etretinate is now bound alongside isotretinoin — you were right that the `SUPPORT` evidence was an etretinate regimen while only isotretinoin was bound.
- The DLd regimen is retagged `OTHER` with all three agents bound individually, and a note saying why: tagging a three-component regimen by its antibody component misclassifies it.
- Added a `Haematological Surveillance After Diagnosis` diagnosis entry, with the point that in most reported IgA myeloma cases the myeloma was diagnosed *at the same time as or after* the skin disease — so a normal screen at presentation does not close the question.

Thanks for the note that PMID:38751960's own figure legend contradicts its introduction's denial of acantholysis. That is a better argument for keeping the `WRONG_STATEMENT` item than the one I gave, and I've kept both graded items unchanged.

Validation: `just validate` passes, **40/40 snippets verified** (up from 36), term / entity-ref / duplicate-key checks pass.

---

matentzn:
Both fixed in 06ae94d95.

**1. The stale gammopathy explanation.** You're right, and it's the same failure mode as the enum-vs-prose divergence you caught on the Papillon-Lefèvre entry in this batch: I corrected the structured field and left the prose that had been written to justify the old value sitting fifteen lines below it. Nothing checks that a term change and the explanation reasoning about it stay in step, and in both cases the stale prose was the more misleading half. Reworded to describe HP:0020194, with the myeloma-as-malignant-end distinction kept.

**2. The "independent confirmation" claim.** Also right, and this one is load-bearing rather than cosmetic, so thank you for reading the author lists. Van Egmond M is senior author on both papers — same VU Amsterdam anti-CD89 programme — and the quoted sentence is that paper's introduction citing its own group's prior work, which its own wording says outright ("previous work of our group showed").

It matters for exactly the reason you give. The FcalphaRI node is graded `ESTABLISHED` on evidence that is *explicitly* not disease-specific — receptor biology in human neutrophils plus a LABD mouse model, applied here on shared isotype and shared histology. How much independent corroboration that carries is the whole basis for the grade, so describing two papers from one lab as two groups inflated it. I've left the grade at `ESTABLISHED` because the underlying receptor biology genuinely is not in doubt, but the explanation now says plainly what the second citation is.

One generalisable thing I'd note from this, since it is the sort of thing I've been collecting across this batch: **nothing in an evidence item records author or group overlap between references.** Two citations from the same laboratory look identical to two from independent groups, and a reader counting supporting items has no way to discount for it. I only caught it because you read the author lists — there is no field that would have made it visible, and no check that would have flagged it.

Validation: `just validate` passes, 40/40 snippets verified (unchanged — prose only), term / entity-ref / duplicate-key checks pass.

---

PR #10027, opened 2026-08-29
curate: Add Leber Congenital Amaurosis 13 (MONDO:0012990, RDH12)

### LINKED ISSUE #10019: Curate Leber congenital amaurosis 13 (MONDO:0012990)
Curate a dismech entry for **Leber congenital amaurosis 13** ([MONDO:0012990](https://monarchinitiative.org/MONDO:0012990)) — LCA13, biallelic RDH12 loss of function, a severe early-onset retinal dystrophy in which the photoreceptor retinol dehydrogenase that clears all-trans-retinal fails.

Initial read: this looks like a **DISEASE**, matching the per-gene retinopathy pattern the KB already uses (`Leber_Congenital_Amaurosis_10`, `RPE65-Related_Retinopathy`, `AIPL1-Related_Retinopathy`, `CRB1_Retinal_Dystrophies`). RDH12 disease has a pathograph of its own — retinaldehyde-adduct toxicity in the photoreceptor inner segment, with severe macular atrophy — and it matters clinically that it is one of the LCA genotypes where RPE65-style gene supplementation is not the answer.

Stub: `stubs/Leber_Congenital_Amaurosis_13.yaml`

matentzn:
Adds `kb/disorders/Leber_Congenital_Amaurosis_13.yaml` — the RDH12-associated early-onset severe retinal dystrophy.

Closes #10019

Follows the per-gene retinopathy pattern the KB already uses (`Leber_Congenital_Amaurosis_10`, `RPE65-Related_Retinopathy`, `AIPL1-Related_Retinopathy`, `CRB1_Retinal_Dystrophies`).

## Why this genotype is not just another LCA gene

**RDH12 is the LCA gene whose mechanism sits outside the visual cycle.** RDH8 does the visual-cycle reduction in the photoreceptor *outer* segment. RDH12 sits in the *inner* segment and clears the all-trans-retinal that leaks inward under continuous illumination, plus lipid peroxidation products. It is a detoxifying enzyme, not a recycling one — and single-cell imaging in knockout mice separates the two cleanly: Rdh8-null outer segments fail to reduce all-trans-retinal, Rdh12-null outer segments are unaffected.

That distinction is the single most consequential fact about the disease, because it makes **RPE65 a poor analogue**. RPE65 disease is a functional visual-cycle defect with relatively preserved macular structure, which is why supplementation works there. RDH12 disease destroys the macula early — atrophy was universal in all 57 subjects of the natural-history cohort, documented in children as young as **two**. Gene supplementation cannot restore atrophied photoreceptors.

Six pathophysiology nodes, with early macular atrophy curated as its own node rather than folded into the degeneration, precisely because it is what makes the genotype therapeutically distinct.

## Two knowledge gaps

**Is there a therapeutic window?** Macular atrophy is universal from age two and the ERG is non-recordable from age one — yet reduced but useful vision is frequently retained until adolescence, with severe loss only after age ten. So there is a decade in which a treated patient has something to preserve, and no validated way to detect preservation: the ERG is already at floor and cannot be a trial endpoint. That dissociation between early structural loss and late functional loss is itself unexplained.

**Which toxic species kills the photoreceptor?** Aldehyde toxicity and retinoic acid overproduction both follow from the same enzymatic loss, and the cited work supports each without discriminating. The aldehyde node is `ESTABLISHED`; the retinoic-acid node is `PROVISIONAL` (one cell-based study, no measurement in an RDH12-deficient retina). It matters because an aldehyde scavenger and a retinoic-acid antagonist are different drugs.

## Two variant-interpretation traps, recorded in `genetic.notes`

1. **Residual in vitro activity does not exclude pathogenicity.** T49M and I51N retain significant activity on purified protein and still fail in cells — reduced NADPH affinity and much lower expression. An activity assay can look reassuring for an allele that is functionally null in a photoreceptor.
2. **A nomenclature trap.** The 2005 biochemical paper calls RDH12 the gene for "Leber's congenital amaurosis **3**", not 13. LCA3 is now assigned to SPATA7 — a search on the old designation returns the wrong gene.

## Curation inputs — and the term-validation finding

One Perplexity deep-research run, committed alongside. Its **reference validation was clean: 8/8 resolved, all on topic.** Its **term validation reported 25 of 44 labels mismatched — the worst of the five reports in this batch — and of the most dangerous kind**: real HPO terms paired with the wrong names, *swapped between each other*.

```
HP:0000639  report says "visual impairment"   HPO says  Nystagmus
HP:0000556  report says "nystagmus"           HPO says  Retinal dystrophy
```

Neither the CURIE nor the name looks wrong on its own. A curator spot-checking one binding by eye would very likely accept it. No CURIE was taken from the report; every binding was resolved against the committed term caches and OLS.

The pairing (clean references, broken terms) is a good illustration that the two validators measure different things and that neither clears the other.

## One binding compromise, stated in place

Peripapillary *sparing* is bound to `HP:0500087` **Peripapillary atrophy**, the nearest anatomical anchor, because HPO has no term for the sparing. The evidence item is graded `PARTIAL` and the description states the actual finding rather than letting the term stand for it.

## Validation

- `just validate` — passed on the first run; **36/36 evidence snippets verified**
- `just validate-terms`, `just check-duplicate-keys`, `just check-entity-refs`, `just validate-history` — pass

The GeneReviews nonsyndromic LCA/EOSRD overview (PMID:30285347) is cited and tagged, but covers the disease class rather than this genotype, so it is not used as an evidence source for phenotypes — those come from the two natural-history studies (57 and 134 individually reported subjects).

Stub `stubs/Leber_Congenital_Amaurosis_13.yaml` deleted.

---

### DIFF
--- a/history/disorders/Leber_Congenital_Amaurosis_13/2026-08-29T081644Z-claude-code-d1bea4.yaml
+++ b/history/disorders/Leber_Congenital_Amaurosis_13/2026-08-29T081644Z-claude-code-d1bea4.yaml
@@ -1,0 +1,40 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Leber_Congenital_Amaurosis_13
+  path: kb/disorders/Leber_Congenital_Amaurosis_13.yaml
+session:
+  id: 2026-08-29T081644Z-claude-code-d1bea4
+  timestamp: '2026-08-29T08:16:44Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10027
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - genetic
+  - prevalence
+  - treatments
+  - references
+  summary: 'Review round 2: two CI-gate defects I introduced, plus the last uncited
+    reference (PR 10027)'
+  details: |
+    Round 2 fixed two defects the round-1 fix commit introduced, and both were mine to have caught before pushing.
+
+    The process failure is worth recording first, because it is the reusable part. After round 1 I ran just validate, validate-terms, check-entity-refs and check-duplicate-keys - and not check-folded-hyphens, check-snippet-length or check-title-snippets. Those three are ungated whole-KB CI steps, precisely so that a kb-only PR cannot skip them, and I skipped them locally anyway. Both new defects were in text I wrote during the fix commit, so both were introduced by fixing something else, which is the situation a re-run of the full local gate set exists to catch.
+
+    Defect 1, check-folded-hyphens at line 1022. A folded-scalar line I wrote ended in light-, so the loaded value read light- damage with a space. Reflowed so light-damage stays on one line. This is a class of error invisible in the source file - the YAML looks correct, and only the loaded value is wrong - which is why the check exists.
+
+    Defect 2, check-snippet-length on the prevalence evidence. I quoted RDH12 (13%) from the German cohort table, which is two words with no propositional content. The same paper carries the sentence that makes the claim, and it is better in every way: Fourteen patients (14/105, 13%, age 3-51 years; 6 female, 8 male) had variants in the RDH12 gene. Swapping it in kept the figure and added the numerator, denominator and age range. The lesson generalises beyond the check: a bare number lifted from a table is not evidence for a claim, because it carries no statement about what the number counts.
+
+    Also took the one remaining suggestion. PMID:17512964 was the last cached-but-uncited reference and, unlike PMID:29398085, had no disposition recorded. It is genuinely useful, so it is now cited rather than dispositioned: across 119 patients spanning LCA, retinitis pigmentosa and cone-rod dystrophy, the RDH12 mutants associated with LCA were inactive or had only residual activity in COS-7 and Sf9 cells, while the changes that proved polymorphic were fully active. That is the mirror image of the entry's existing variant-interpretation caveat rather than a contradiction of it: residual in vitro activity does not exclude pathogenicity, but full in vitro activity has so far tracked with benignity. Both statements now sit together in the genetic notes.
+
+    Validation after changes. Schema validation passes; 57 of 57 evidence snippets verified; and this time the full ungated gate set was run rather than a subset - check-folded-hyphens, check-snippet-length, check-title-snippets, check-snippet-grading and check-environmental-evidence all pass, alongside term validation, entity-ref and duplicate-key checks. The same full set was re-run across all five entries in this curation batch, which surfaced one further folded-hyphen defect in the IgA pemphigus entry that no reviewer had reached yet.
--- a/kb/disorders/Leber_Congenital_Amaurosis_13.yaml
+++ b/kb/disorders/Leber_Congenital_Amaurosis_13.yaml
@@ -85,6 +85,8 @@ references:
   title: "Genetic and Clinical Profile of Retinopathies Due to Disease-Causing Variants in Leber Congenital Amaurosis (LCA)-Associated Genes in a Large German Cohort."
 - reference: PMID:38371258
   title: "RDH12-associated retinal degeneration caused by a homozygous pathogenic variant of 146C>T and literature review."
+- reference: PMID:17512964
+  title: "Novel RDH12 mutations associated with Leber congenital amaurosis and cone-rod dystrophy: biochemical and clinical evaluations."
 - reference: PMID:30285347
   title: "Nonsyndromic Leber Congenital Amaurosis / Early-Onset Severe Retinal Dystrophy Overview."
   tags:
@@ -196,10 +198,11 @@ prevalence:
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: >-
-      RDH12 (13%)
+      Fourteen patients (14/105, 13%, age 3–51 years; 6 female, 8 male) had
+      variants in the RDH12 gene.
     explanation: >-
       RDH12's share of a 105-patient German cohort ascertained on
-      LCA-associated genes. Note the same paper gives 7 per cent when restricted
+      LCA-associated genes, with numerator, denominator and age range. Note the same paper gives 7 per cent when restricted
       to patients whose clinical diagnosis was LCA rather than another inherited
       retinal dystrophy - a two-fold spread inside one cohort, from the inclusion
       criterion alone, which is the clearest available illustration of why these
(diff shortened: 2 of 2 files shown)


matentzn:
All five IMPORTANT items are fixed, plus four of the five suggestions. Pushed in 1a29f062b and ea13cfed3.

**Item 1 is the one that matters, and it was a substantive correction rather than a polish item.** `Rdh12-/-` mice do not develop retinal degeneration — two knockout lines, both with normal retinal histology — and I cited that mouse for the light-susceptibility claim without recording that it fails to reproduce the disease's defining phenotype. The Light Exposure Reduction treatment rested entirely on it.

Fixing it properly meant changing three other things, not just adding a discussion:

- A `HUMAN_MODEL_MISMATCH` discussion attaching to **five** places (two pathophysiology nodes, the animal model, the gene-therapy treatment, the light-avoidance treatment), because the mismatch constrains all of them.
- **The aldehyde toxicity node downgraded `ESTABLISHED` → `PROVISIONAL`.** Its step to degeneration rests on a model that does not degenerate; that grade was wrong before your review and would have stayed wrong.
- The top-level description rewritten. It said the therapeutic window "closes in early childhood", which — as you noted separately — is more pessimistic than the sources support. It now states the constraint without asserting a closing age, and carries the model limitation.

**Item 2 (ER/oxidative stress)** — added as its own `PROVISIONAL` node. Your point that its absence made the which-toxic-species gap read as a **false binary** is exactly right, and the discussion's rationale now says so explicitly: stress signalling is the shared downstream consequence of either candidate, so the question is which upstream species drives it, not a two-way choice about what kills the cell.

**Item 3 (pregabalin)** — added, targeting the new stress node. Worth its own entry because it is the only candidate in this literature that acts on the toxic species rather than the missing gene, and a scavenger strategy does not depend on delivering a transgene to photoreceptors that may already be lost — which is exactly the constraint the macular-atrophy timing imposes on the gene-therapy route.

**Item 4 (PMID:31237438)** — two evidence items on the gene-therapy treatment, one `PARTIAL`, quoting the authors' own limiting clause inside their efficacy claim.

**Item 5 (peripapillary binding)** — dropped `HP:0500087` entirely; free-text `preferred_term` only, evidence regraded `SUPPORT`. Your argument is the general one worth keeping and I'll apply it elsewhere: **a `PARTIAL` grade and an explanatory description do not survive KGX/BioLink export**, so the graph would have asserted atrophy where the finding is sparing. A prose caveat is not a substitute for a correct binding.

**Suggestions taken (4 of 5):**
- `animal_models` + `experimental_models` added. The Rdh12-null mouse carries both a `RECAPITULATES` link at `HIGH` fidelity (the enzymatic deficit) and a `FAILS_TO_RECAPITULATE` link at `LOW` (degeneration), with an `UNCHANGED` readout for normal retinal histology — recorded as the real negative result it is rather than omitted. The zebrafish CRISPR mutant is `PARTIALLY_RECAPITULATES` with the authors' "early signs" hedge intact. The HEK-293 lines are `LOW` fidelity: a cell with no outer segment, no visual cycle and no light exposure can show that mutant RDH12 fails to protect a cell, and cannot show that this is what kills a photoreceptor.
- Renamed to "Mild Hyperopia", with a note that the schema has no phenotype-negation slot, so the discriminating information — the *absence* of the marked form — can only be stated in prose.
- Softened the therapeutic-window prose, as above.
- Cited two of the four uncited cohorts. PMID:37240262 turned out to be the most useful of them, because it makes the pooling problem concrete: **13% of the German cohort across LCA-associated genes, 7% restricted to a clinical LCA diagnosis** — a two-fold spread inside one cohort from the inclusion criterion alone. PMID:38371258 gives thirteen patients sharing one variant with varying clinical characteristics, graded `REFUTE` as evidence against a simple allele-to-severity rule.

Two deliberate absences are now in `notes:` so they don't read as omissions: no `clinical_trials` section (an ongoing prospective natural-history study is described, but no interventional trial is identified in any cited source, and I did not invent an identifier), and PMID:29398085 is left uncited because it is a **CEP290** natural-history study with no RDH12 content — cached because the deep-research citation list resolved it.

**One thing found while working this, unrelated to this PR.** Running the model-link pytest lanes surfaced a pre-existing failure on `main`: `test_failure_to_recapitulate_links_are_substantiated` fails on `Polycystic_Kidney_Disease_3.yaml`. That file is untouched by this branch. Filed as #10028 after checking for duplicates.

Validation: `just validate` passes, **56/56 snippets verified** (up from 36), term / entity-ref / duplicate-key checks pass.

---

matentzn:
Both fixed, plus the remaining suggestion. Pushed in bd275dd01.

**The process failure first, because it's the reusable part.** Both defects were introduced by my *round-1 fix commit*, and both were mine to catch before pushing. After round 1 I ran `just validate`, `validate-terms`, `check-entity-refs` and `check-duplicate-keys` — and **not** `check-folded-hyphens`, `check-snippet-length` or `check-title-snippets`. Those three are ungated whole-KB CI steps *precisely so a `kb/`-only PR cannot skip them*, and I skipped them locally anyway. Fixing one thing and breaking another is exactly the case a full local re-run exists to catch.

**Defect 1 (folded hyphen).** Reflowed so `light-damage` stays on one line. Worth noting that this class of error is invisible in the source file — the YAML reads correctly and only the *loaded* value is wrong — which is why the check has to exist.

**Defect 2 (two-word snippet).** You're right, and your suggested replacement is better in every way. `RDH12 (13%)` was a number lifted from a table with no statement about what it counts. The sentence — *"Fourteen patients (14/105, 13%, age 3–51 years; 6 female, 8 male) had variants in the RDH12 gene"* — keeps the figure and adds numerator, denominator and age range. The general lesson holds beyond the check: a bare number from a table is not evidence for a claim.

**The remaining suggestion (PMID:17512964).** Cited rather than dispositioned, because it earns its place. Across 119 patients spanning LCA, RP and cone-rod dystrophy, the RDH12 mutants associated with LCA were inactive or had only residual activity in COS-7 and Sf9 cells, while the changes that proved polymorphic were fully active. That is the **mirror image** of the entry's existing variant-interpretation caveat rather than a contradiction: residual in vitro activity does not exclude pathogenicity, but full in vitro activity has so far tracked with benignity. Both statements now sit together in the `genetic:` notes, which is more useful to a diagnostic lab than either alone.

**One thing your finding turned up elsewhere.** I re-ran the full ungated gate set across all five entries in this curation batch, not just this one — and it surfaced an equivalent folded-hyphen defect in `IgA_Pemphigus.yaml` (#10026) that no reviewer had reached. Fixed there too.

Validation: `just validate` passes, **57/57 snippets verified**, and this time the full set — `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys` — all pass.

---

Issue #9897, opened 2026-08-28
`just refresh-orphadata` fails on a checksum mismatch for en_product1.xml, so no new ORPHA record can be cached


aleixpuigb:
Citing an Orphanet record for a disease that has no cache file yet is currently impossible: the bulk XML cannot be refreshed.

```
$ just refresh-orphadata
RuntimeError: checksum mismatch for en_product1.xml after download: got
df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb, expected
fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

Orphadata has republished `en_product1.xml` and the sha256 pinned in `data/orphadata/MANIFEST.yaml` no longer matches what the URL serves. Since `data/orphadata/*.xml` is gitignored, a fresh checkout has no local copy to fall back on, and `just structured-rebuild-orphanet --id <n>` then fails with:

```
FileNotFoundError: [Errno 2] No such file or directory: '.../data/orphadata/en_product1.xml'
```

## How I hit it

Reviewing [#9892](https://github.com/monarch-initiative/dismech/pull/9892) asked for an ORPHA citation on a new COXPD4 entry. MONDO:0012534 cross-references `Orphanet:254925`, and `references_cache/ORPHA_254925.md` does not exist, so it would have to be generated — which needs the bulk XML. `just fetch-reference ORPHA:254925` is not a route either; it reports `No source found for reference type: ORPHA:254925`, since ORPHA records come from the structured-source builder rather than the reference fetcher.

The already-committed `references_cache/ORPHA_*.md` files are unaffected — this only blocks caching an ID that isn't already there.

## Why I did not just re-pin it

Updating the manifest checksum is a data-provenance change that would shift every ORPHA cache to a new Orphanet release, and it deserves its own PR with the resulting `git diff references_cache/ORPHA_*.md` reviewed — not a drive-by edit inside a curation PR for an unrelated disease. Rebuilding from a download whose checksum does not verify would defeat the point of pinning.

## Suggested fix

Re-pin `data/orphadata/MANIFEST.yaml` to the current release in a dedicated PR, regenerate the ORPHA caches, and review the diff. Two things worth considering at the same time:

- The failure mode is silent until someone needs a *new* ORPHA ID, because existing caches keep working. A periodic check that the pinned checksums still match what the URLs serve would surface drift when it happens rather than when it blocks a curator.
[...]

PR #10041, opened 2026-08-29
Add IRIDA syndrome (MONDO:0008788, TMPRSS6)

### LINKED ISSUE #10036: Curate IRIDA syndrome (MONDO:0008788)
Curate a dismech entry for **IRIDA syndrome** ([MONDO:0008788](https://monarchinitiative.org/MONDO:0008788)) — iron-refractory iron deficiency anaemia, caused by biallelic loss-of-function variants in *TMPRSS6* (matriptase-2), which normally dampens hepcidin transcription; without it hepcidin stays inappropriately high and blocks ferroportin-mediated iron export from enterocytes and macrophages.

Provisional read: this looks like a `DISEASE`. The stub records no MONDO descendants (leaf) and a single causal gene, `hgnc:16517 TMPRSS6`. Nothing in `kb/` currently mentions *TMPRSS6*, hepcidin-driven iron restriction of this type, or MONDO:0008788.

Stub: `stubs/IRIDA_Syndrome.yaml`

aleixpuigb:
Curates iron-refractory iron deficiency anaemia — the disease that established which protein lets the liver sense iron deficiency and turn hepcidin down. Closes #10036 and deletes the stub.

## The pathograph, and why it forks

Eight nodes. Biallelic *TMPRSS6* loss of function branches into **two parallel mechanisms**, then they converge:

```
Biallelic TMPRSS6 LoF
├── Loss of matriptase-2 cleavage of membrane hemojuvelin   (PROVISIONAL)
└── Failure to suppress HAMP transcription                  (ESTABLISHED)
        └──► Inappropriately elevated circulating hepcidin
                ├── Impaired duodenal iron absorption
                └── Macrophage iron sequestration
                        └──► Systemic iron restriction
                                └──► Iron-restricted erythropoiesis
```

The fork is deliberate and is the main modelling call in this PR. Two papers published months apart in 2008 give different accounts of how matriptase-2 reaches hepcidin. Silvestri et al. locate it in the ectodomain protease, cleaving membrane hemojuvelin. Du et al. report that the **cytoplasmic** domain mediates Hamp suppression via proximal promoter elements — which an ectodomain protease activity does not explain. Merging them into one chain would assert a resolution nobody has published, so they are curated as parallel arms with a `KNOWLEDGE_GAP` discussion saying why it matters: it decides whether a missense variant outside the catalytic domain has a mechanism available to it.

The two-limb convergence below hepcidin is the other thing worth a look. Splitting the enterocyte and macrophage blocks is what lets the entry explain the clinical fact that separates IRIDA from a pure absorption defect — parenteral iron bypasses the gut but is still handled by macrophages, so it corrects the anaemia slowly and only partially.

## Evidence

Ten references, all fetched with `just fetch-reference`. **62/62 snippets verified.**

The entry was built from primary literature first; an OpenScientist deep-research report was run alongside and is committed under `research/`. Four of its leads were verified on PubMed and added, and **two of them changed a conclusion rather than adding detail**:

- **The draft said no validated decision rule existed for the hepcidin measurement. That was wrong.** [PMID:35163840](https://pubmed.ncbi.nlm.nih.gov/35163840/) reports a transferrin-saturation/hepcidin ratio with AUC 1.000 and a 5.6 %/nM threshold. It is now curated as a `definitions` `PHENOTYPE_ALGORITHM`, with the constraints recorded as inclusion criteria rather than as prose caveats — the threshold is bound to isotope-dilution mass spectrometry and does not transfer to an uncalibrated immunoassay, and the derivation excluded recent iron therapy and CRP ≥ 10 mg/L, which are the two commonest confounders in exactly the population that would use the test.
- **The strongest human causal evidence in the entry is not from a patient.** [PMID:42053460](https://pubmed.ncbi.nlm.nih.gov/42053460/), a Phase 1 trial of an anti-TMPRSS6 antibody in 56 healthy volunteers, raised hepcidin-25 and lowered serum iron, transferrin saturation and haemoglobin. That is the IRIDA biochemistry produced deliberately, randomised and placebo-controlled — a pharmacological phenocopy, and better causal evidence for the hepcidin node than any observational series.

Two evidence items on **Oral Iron** deliberately disagree, and the disagreement is the content: the defining cohort's negative result is graded `REFUTE`, and a seven-child uncontrolled oral-iron-plus-vitamin-C cohort is graded `PARTIAL`. The `PARTIAL` grade turns on something the source says in its own title — those children had the IRIDA *phenotype*, not confirmed biallelic genotype, so some may have had oral-iron-responsive deficiency that IRIDA never explained.

One snippet reproduces a typo in its source (`autoproteotylic` for `autoproteolytic`). It is kept verbatim, with a note, because the cache is the authority for a quote. The deep-research report had silently corrected it, which is why its own reference validation flagged that quote as not found.

## Things I decided not to assert

- **No `mechanistic_category`.** That enum is a closed set of protein-class nosologies (RASopathy, ciliopathy, laminopathy …) and none of them describes a hepatic protease that regulates a hormone. The disease is placed on `icimd_category: iron_metabolism` instead. Leaving the slot empty is the accurate answer, not a gap.
- **No reference range on serum hepcidin.** The informative quantity is the ratio, and its threshold is method-bound. A single-analyte interval here would be false precision.
- **No `target_mechanisms` link on oral iron or on hepcidin-lowering therapy.** Oral iron acts on no node in this pathograph — that is the clinical point — and hepcidin-lowering therapy does not exist for this disease. Either link would draw an arrow asserting a therapeutic action nobody has demonstrated.
- **Common *TMPRSS6* variation** is a real quantitative-trait association with iron indices, and is deliberately not asserted here. It is a different claim with a different evidence base.

## A second discussion worth a reviewer's eye

`HUMAN_MODEL_MISMATCH` on matriptase-2 and neogenin. In cultured hepatoma cells matriptase-2 cleaves neogenin and markedly reduces it; in mouse liver it *stabilises* it, and hepcidin suppression requires neogenin to be present ([PMID:41534828](https://pubmed.ncbi.nlm.nih.gov/41534828/)). Same protease, same partner, opposite directions in the two systems. The stake is variant interpretation: novel *TMPRSS6* missense variants are assessed by exactly the transfected-cell assay whose fidelity this puts in question. Note the mismatch is model-to-model — mouse liver against human cell line, with the human hepatocyte unobserved in both — which the discussion states rather than overclaiming.

## Validation

| Check | Result |
|---|---|
| Snippets verified | 62/62 |
| `just check-stubs` | pass after stub deletion |

Staging was targeted. The five cache files carry exactly the five CURIEs this entry introduced (`GO:0034755`, `GO:0060586`, `hgnc:16517`, `HP:0012464`, `HP:0031877`), in canonical sorted position. One reference was fetched and then dropped from the cache because the entry does not cite it.

---


aleixpuigb:
All three IMPORTANT findings and all four suggestions are fixed. Pushed as d8af18a. 72/72 snippets verified, full gate battery clean.

**The suggestion you ranked lowest turned out to be the most consequential.** Finding 6 pointed at two unconsumed papers as "would each add something". They don't add — they *falsify*:

- PMID:23319530 is a genotype-confirmed kindred (compound het p.G442R / p.E522K, both previously reported in classic IRIDA) **maintained on low-dose oral iron alone**. My `Refractoriness to Oral Iron` was banded `OBLIGATE`. One published exception makes that false.
- PMID:36261087 reports a proband with **normocytic anaemia and a normal MCV**, so `Very Low Mean Corpuscular Volume` was over-claimed too.

Both bands are now `VERY_FREQUENT`, and each counterexample is curated as a `REFUTE` evidence item on the phenotype it contradicts rather than mentioned in prose. The oral-iron phenotype keeps a note saying the tension should stay: the disease is *named* for refractoriness, and a genotype-confirmed kindred responds. Dropping either fact loses information.

**Ferritin (finding 1).** You were right that this was the largest gap and right about why. Added as a phenotype using the p < 0.001 comparison and the low–normal-range statement; your line-105 quote about iron maldistribution is now evidence on the `Macrophage Iron Sequestration` node, which did indeed have no human-measurable correlate. The differential-diagnosis sentence is softened — but not to a flat "they differ", because the measured difference is a group difference with overlapping distributions and it points counter-intuitively (higher ferritin in the patient with the genetic iron-handling defect). That nuance is in the `description` and `distinguishing_features`.

One deliberate call: the binding is the general `HP:0040133`, not the increased or decreased child. The finding is neither — it is a value that would be unremarkable in isolation and is abnormal only relative to iron status. Same ontology gap as the hepcidin binding in this entry, and both now say so in `notes` rather than picking a child term that asserts a direction patients don't reliably show. The hyperferritinaemia *presentation* is separate, `HP:0003281`, banded `VERY_RARE`.

**Finding 3 (PMID:32384154).** Folded in, and it changed the discussion's question rather than adding to its answer. The prompt was "which domain"; it is now "is this a proteolytic event at all" — the catalytic domain is required to suppress hepcidin while its proteolytic activity is not, and the disease allele in that work lost partner binding, not catalysis. The cleavage node is kept (the reaction is real and human alleles impair it) but its `description` now says what is in doubt is whether suppression runs through it. Added a second proposed experiment for the practical consequence: every assay used to classify a novel *TMPRSS6* variant reads out cleavage, so a cleavage-normal result may not be reassurance.

**Finding 2.** `CHEBI:29033` iron(2+) on both treatments, per the `Autoimmune_Gastritis.yaml` precedent.

**Suggestions 4, 5, 7.** The `DECREASED` modifier is dropped from the hepcidin node's `GO:0060586` — you were right that it read as the opposite of the node. The two treatment-response phenotypes moved to `category: Treatment Response`, which already exists in the KB. OMIM and Orphanet are recorded as `external_assertions` rather than `mappings`: that is the convention actually in use (121 ORPHA records across `kb/`), where `mappings` is used by two files.

**Your GeneReviews disclosure — checked, and your offline read was right.** I had network. `GeneReviews[Book] OR GeneReviews[Title]` against both `TMPRSS6` and `"iron-refractory iron deficiency anemia"` returns nothing; no chapter exists, and PMID:25064705 is indeed a *Hematol Oncol Clin North Am* review. Now recorded in `notes` so nobody re-derives it. Worth saying that this check found nothing here but found a **blocking** miss on the sibling PR #10042, where the chapter does exist and I had not looked — so the caveat you flagged was worth flagging.

One thing I could not do, recorded in `external_assertions` rather than skipped silently: `references_cache/ORPHA_209981.md` cannot be generated. `just fetch-reference ORPHA:209981` reports "No source found for reference type" because ORPHA records come from the structured-source builder, and that builder needs Orphadata bulk XML whose pinned checksum is stale (#9897). The Orphanet identifier is therefore asserted, not cited as evidence.

---

PR #10042, opened 2026-08-29
Add autosomal recessive cutis laxa type 2A (MONDO:0018163, ATP6V0A2)

### LINKED ISSUE #10037: Curate autosomal recessive cutis laxa type 2A (MONDO:0018163)
Curate a dismech entry for **autosomal recessive cutis laxa type 2A** ([MONDO:0018163](https://monarchinitiative.org/MONDO:0018163)) — ARCL2A / Debré-type cutis laxa, caused by biallelic *ATP6V0A2* variants. The a2 subunit of the vesicular V-ATPase is needed for Golgi acidification and trafficking, so loss of it produces a combined N- and O-glycosylation defect (a type II CDG) alongside abnormal elastic-fibre assembly — wrinkly, inelastic skin plus neurodevelopmental and skeletal involvement.

Provisional read: `DISEASE`. One causal gene (`hgnc:18481 ATP6V0A2`), one mechanism. The stub lists a single MONDO descendant, `MONDO:0010208 wrinkly skin syndrome` — historically a separate label, now understood as the milder end of the same *ATP6V0A2* spectrum, so it belongs as a `has_subtypes` entry rather than a separate disease. Nothing in `kb/` mentions *ATP6V0A2* or MONDO:0018163; `ALDH18A1_Cutis_Laxa.yaml` covers a different gene and a different mechanism.

Stub: `stubs/Autosomal_Recessive_Cutis_Laxa_Type_2A.yaml`

aleixpuigb:
Curates ARCL2A, where one physical variable — trans-Golgi luminal pH — produces phenotypes in skin, glycosylation and brain that would otherwise look unrelated. Closes #10037 and deletes the stub.

## The pathograph

```
Biallelic ATP6V0A2 LoF
  └──► Elevated trans-Golgi luminal pH        ← the disease variable
        ├── Combined N-/O-glycosylation defect
        │     └──► Aberrant cortical neuron migration   (PROVISIONAL)
        └── Impaired Golgi trafficking
              └──► Tropoelastin secretion block
                    └──► Deficient dermal elastic fibre network
```

Everything hangs off one node, and the reason it can is a pair of mouse models that separate the pump's transport function from the protein. Trans-Golgi pH is 5.80 in wild type, 6.25 in a knock-in whose R755Q change abolishes *only* V0a2-mediated proton transport, and 6.52 in the null — and phenotype severity tracks the pH gradient. The knock-in is what licenses grading the pH node ESTABLISHED rather than inferred: it shows the pumping activity is the disease-relevant function, not some structural role of the subunit.

## Two modelling calls

**Wrinkly skin syndrome is a `has_subtypes` entry, not a separate disease.** The same *ATP6V0A2* variants cause both, and the gene-discovery paper ascertained the two labels together, so the lumping is original to the entity rather than a later reinterpretation. The stub's MONDO context agreed — MONDO:0018163 lists exactly one descendant, MONDO:0010208 wrinkly skin syndrome.

**The severity axis is not genotype, and the entry says so.** With 78% predicted-null alleles you would expect residual function to grade severity, and one missense allele (p.P87L) is reported with unusually mild disease. But the systematic review of 70 individuals looked and found *no* genotype–phenotype correlation; the only discriminator was age at ascertainment. A `KNOWLEDGE_GAP` discussion argues that this may not be a power failure: if the disease variable is a continuous pH shift and the measured variable is a categorical allele class, a null correlation is what you'd get. The practical consequence is a counselling one, so the entry declines to curate a genotype–severity gradient it cannot evidence.

## What the deep-research report got wrong

An OpenScientist report was run alongside and is committed under `research/`. Four of its leads were verified on PubMed and added — the systematic review, the neurometabolic cutis laxa cohort, the ocular case series, and the 2024 mouse study. **Three of its ontology suggestions were rejected:**

| Report proposed | Problem |
|---|---|
| `GO:0006486` protein glycosylation | Obsolete — its own term-validation section flagged this correctly |
| `HP:0002269` labelled "Abnormal cortical gyration" | That ID is *Abnormality of neuronal migration*. The entry uses `HP:0002536`, which is the term that label actually names |
| `HGNC:865` for *ATP6V0A2* | Wrong identifier. The correct one is `hgnc:18481`, as the curation stub already recorded |

The HP one is the instructive case: the identifier resolves and the label is plausible, so nothing looks wrong. Only the term validator's "names something else" check catches it.

## Evidence notes worth a reviewer's eye

- **Four phenotypes are banded FREQUENT on an association, not a prevalence.** Seizures, developmental delay, high/cleft palate and facial dysmorphism come from a systematic review that reports their *association with age at ascertainment* with p values — it does not report their frequency. The snippets carry the p values; those are not frequency data, and `notes:` says so. Where a real denominator exists it is used instead: 4/4 for the corneal finding, all-patients for the glycosylation abnormality.
- **The corneal band rests on four patients.** FREQUENT because 4/4 were affected in a prospectively and uniformly examined cohort, not because four is a large series. Both facts are in `notes:` for a reader deciding whether to order corneal tomography.
- **High/cleft palate is curated once, not twice.** The review reports it as one combined category; splitting it into two phenotype entries against the same citation would double-count a single observation.
- **Facial dysmorphism is bound to the general `HP:0001999`.** The specific features have their own HP terms, but the review counted "facial dysmorphism" as one category, and using the specific terms against that citation would claim a specificity the source does not carry.
- **Three negative results are curated as evidence.** Normal fibrillin-1 assembly and normal lysyl oxidase activity localise the lesion to tropoelastin trafficking rather than to scaffold assembly or cross-linking; and the brefeldin-A delay being *absent* in GORAB- and PYCR1-deficient cells is what makes it specific to ATP6V0A2 rather than generic to cutis laxa. That last one also does duty as a distinguishing feature in the differential.

## A discussion that is deliberately a weak claim

`HUMAN_MODEL_MISMATCH` on globozoospermia. Both mouse lines are male-infertile with impaired Golgi-derived acrosome formation, and the acrosome is a Golgi-derived organelle, so the mechanism gives no reason for it to be mouse-specific. But no human observation exists — the finding is described as hitherto unrecognised and the clinical literature does not report fertility at all. This is the *weaker* sense of the label: not a model finding contradicted in humans, but one never looked for where the mechanism predicts it. The rationale says that explicitly, and the entry asserts nothing about human fertility.

## Validation

| Check | Result |
|---|---|
| Snippets verified | 58/58 |

Staging was targeted. The cache diff carries exactly the four CURIEs this entry introduced (`GO:0005802`, `GO:0061795`, `hgnc:18481`, `MONDO:0018163`). Seven reference-cache files are new; `PMID_23963297.md` was already in the repo and is reused rather than re-added.

---

### DIFF
--- a/history/disorders/Autosomal_Recessive_Cutis_Laxa_Type_2A/2026-08-29T130827Z-claude-code-2fd5de.yaml
+++ b/history/disorders/Autosomal_Recessive_Cutis_Laxa_Type_2A/2026-08-29T130827Z-claude-code-2fd5de.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Recessive_Cutis_Laxa_Type_2A
+  path: kb/disorders/Autosomal_Recessive_Cutis_Laxa_Type_2A.yaml
+session:
+  id: 2026-08-29T130827Z-claude-code-2fd5de
+  timestamp: '2026-08-29T13:08:27Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10042
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  - phenotypes
+  summary: 'Review round 2: re-slot anti-seizure link, split scarring phenotype'
+  details: |
+    Addressed the reviewer's round-2 CHANGES_REQUESTED, whose single IMPORTANT finding was a defect introduced by the round-1 fix. The Anti-Seizure Medication treatment carried a target_mechanisms edge to the Aberrant Cortical Neuron Migration node graded MODULATES, while its own description said the drug works without affecting the migration defect. The edge asserted what the prose beside it denied, and target_mechanisms is not inert - it draws a treatment-to-node arrow in the pathograph and carries into the export layer where the qualifying sentence does not travel. Re-slotted to target_phenotypes pointing at the Seizure phenotype, which is what the link's own evidence sentence supports, with a note recording why the distinction matters. That also restores the truth of the neighbouring note stating no treatment acts on any node in this pathograph. Three suggestions applied. Split Easy Bruising and Atrophic Scarring into two phenotypes so name and ontology binding agree, the second bound to HP:0001075. Moved the GeneReviews sentence about skin findings decreasing with age off the neurologic regression phenotype, where it was not evidence for regression, onto the cutis laxa phenotype where it is evidence, with the contrast made in prose instead. Attached the two GeneReviews facial-feature statements to Facial Dysmorphism, which previously rested only on an age-at-ascertainment p value. Recorded that GeneReviews does not mention the palate at all, so the High or Cleft Palate band still rests on the association data and remains a coarse literature reading. Snippets 82/82 verified; full gate battery clean.
--- a/kb/disorders/Autosomal_Recessive_Cutis_Laxa_Type_2A.yaml
+++ b/kb/disorders/Autosomal_Recessive_Cutis_Laxa_Type_2A.yaml
@@ -641,6 +641,17 @@ phenotypes:
     explanation: >-
       Supports the cutaneous phenotype as the one feature that persists across the
       whole severity spectrum, including the attenuated end.
+  - reference: PMID:20301755
+    reference_title: "ATP6V0A2-Related Cutis Laxa."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      The skin findings decrease with age, although easy bruising and
+      Ehlers-Danlos-like scars have been described in some.
+    explanation: >-
+      The cutaneous trajectory. Note the contrast with the neurological course,
+      which does the opposite - that comparison is made in the description of the
+      neurologic regression phenotype rather than duplicated here as evidence.
 - category: Neurological
   name: Seizures
   frequency: VERY_FREQUENT
@@ -753,17 +764,14 @@ phenotypes:
       neurologic regression.
     explanation: >-
       The frequency and the phenotype in one statement.
-  - reference: PMID:20301755
-    reference_title: "ATP6V0A2-Related Cutis Laxa."
-    supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
-    snippet: >-
-      The skin findings decrease with age, although easy bruising and
-      Ehlers-Danlos-like scars have been described in some.
-    explanation: >-
-      Quoted here rather than on the skin phenotype because it is the contrast
-      that makes the regression point: the cutaneous course improves while the
-      neurological course does not.
+  notes: >-
+    An earlier revision carried the GeneReviews sentence "The skin findings
+    decrease with age..." as a second evidence item here, on the argument that
+    the contrast with the skin is what makes the regression point. That was
+    wrong: the sentence is about the skin and does not support a regression
+    phenotype, whatever it illuminates. It now sits on the cutis laxa phenotype
+    where it is evidence, and the contrast is made in this description instead -
+    which is prose's job, not an evidence item's.
 - category: Neurological
   name: Cortical Gyration and Neuronal Migration Abnormality
   frequency: FREQUENT
@@ -977,7 +985,7 @@ phenotypes:
     explanation: >-
       Graded PARTIAL: management sentence, occurrence not frequency.
 - category: Dermatological
-  name: Easy Bruising and Atrophic Scarring
+  name: Easy Bruising
   frequency: OCCASIONAL
   description: >-
     Described in some patients as the skin findings otherwise decrease with age -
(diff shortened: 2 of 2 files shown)


aleixpuigb:
The blocking finding was correct and it was a process failure on my part, not a judgement call. Fixed in 96ccdf2. 80/80 snippets verified, full gate battery clean.

**GeneReviews.** PMID:20301755 is now tagged and mined across all four of its sections. What makes the miss worth naming rather than quietly patching: the deep-research report never surfaced the chapter, and I ran no GeneReviews-specific search to compensate — so the entry ended up complaining *in its own `notes`* that counted frequency data did not exist for these phenotypes, while a source giving them sat unread. The failure mode is a confident-sounding gap statement produced by not looking.

It changed the content substantially:

| | before | after | on |
|---|---|---|---|
| Seizures | `FREQUENT` | `VERY_FREQUENT` | "Nearly all affected individuals…" |
| Global developmental delay | `FREQUENT` | `VERY_FREQUENT` | same sentence |
| Cortical gyration abnormality | `OCCASIONAL` | `FREQUENT` | "In most (not all)…" |

Eleven phenotypes added: neurologic regression, cerebellar malformation, neonatal hypotonia, large fontanelles, coarse hair, high myopia, congenital hip dislocation, inguinal hernia, easy bruising.

**Neurologic regression is the one that mattered most, and you flagged exactly why.** The entry described the disease as attenuating with age. That is true of the skin and the opposite of true for the brain, and nothing in the entry said so. It now carries `clinical_course: PROGRESSIVE` and a description making the contrast explicit, with the "skin findings decrease with age" quote deliberately placed on the *regression* phenotype rather than on the skin one — it is the contrast that carries the information.

**Cerebellar malformation** was named in the same GeneReviews sentence as the cortical finding and I had curated one and not the other. It is now a separate phenotype rather than folded in, because the pathograph explains the cortical finding through a neuronal-migration node and cerebellar malformation is not obviously the same lesion. The entry claims no mechanism for it, which is the honest position.

**Diagnosis section (finding 5).** Added: combined transferrin + apoC-III isoelectric focusing as first-line, molecular confirmation, and the surveillance schedule. You were right that the ophthalmological entry was misfiled — its own description said it was not a treatment. It has moved to `diagnosis:` with a `notes` line recording the move, and it keeps the corneal-imaging extension from PMID:40043299 alongside the GeneReviews recommendation, since that is the gene-specific part.

One schema note from doing this: `Diagnosis.markers` is a single free-text string with no ontology binding, so a two-analyte test can't be represented as two grounded markers. Both analytes are named in one sentence and the `notes` says why; HP:0012301 stays on the `phenotypes` and `biochemical` entries where it validates.

**Findings 6 and 7.** Anti-seizure medication and early developmental intervention are added as treatments, the former with a `target_mechanisms` link to the migration node graded `MODULATES` (it suppresses seizures arising from a malformed cortex without touching the malformation). Recurrence risks and reproductive options are on `inheritance:`.

**Suggestions.** `CL:0000540` → `CL:0010012` (cerebral cortex neuron), applied. On histopathology: I did not add a section, and the reasoning is now in `notes` — abnormal elastic fibres are the classic finding and are already quoted on the elastic-fibre node, but no source fetched here gives the stain, fibre morphology or biopsy site, and an entry saying only "abnormal elastic fibres" would duplicate the phenotype without the detail that section exists to carry. On Orphanet prevalence: not possible. `just fetch-reference ORPHA:…` reports "No source found for reference type" and `structured-rebuild-orphanet` needs bulk XML whose pinned checksum is stale (#9897).

**On your two disclosures.** You could not reach PubMed and flagged it both here and on #10041. On #10041 your offline read was right and no chapter exists; here the chapter did exist and I had not looked. Both outcomes argue the disclosure was worth making.

---

aleixpuigb:
Fixed in c4a1262. 82/82 snippets verified, full gate battery clean.

**You caught a defect my own fix introduced, and the reasoning for why it matters is the part I'd got wrong.** I had treated `target_mechanisms` as a place to say *what the drug does relative to the mechanism*, including "nothing" — so I wrote a MODULATES edge and then explained in the description that it doesn't actually touch migration. That is exactly backwards. The edge is the assertion; the description is commentary that does not survive into the export layer. A graph consumer would have read "anti-seizure medication acts on cortical neuron migration" with no trace of the sentence denying it.

Re-slotted to `target_phenotypes` → `Seizure` (`HP:0001250`), which is what the link's own evidence sentence supports and nothing more. Added a `notes` line recording the distinction, since it is the kind of thing that will be got wrong again. And you are right that it also restores the truth of the neighbouring note on `Multidisciplinary Symptomatic Management` — "no available treatment acts on any node in this pathograph" is accurate again, so it stays.

**Three suggestions, all applied:**

- **`Easy Bruising and Atrophic Scarring`** — split into two phenotypes rather than trimming the name, since the GR sentence genuinely reports both. Scarring is bound to `HP:0001075` Atrophic scars. Name and binding now agree on both.
- **The regression evidence item.** You were right and my `explanation` was doing work the evidence could not. The sentence is about the skin; a good argument for why it is *relevant* to the regression phenotype does not make it *evidence for* that phenotype. It has moved to the cutis laxa phenotype where it is evidence, the contrast is made in the regression description instead, and a `notes` line on the regression phenotype records the correction so the next curator does not re-add it on the same reasoning.
- **`Facial Dysmorphism`** — the two GR facial-feature statements are attached. They are a different kind of claim from the p-value association: "distinctive facial features are present at birth" is presence, and "may become quite pronounced" in childhood is severity-over-time. Neither depends on ascertainment age, so the band no longer rests on association data alone.

**On `High or Cleft Palate` — there was nothing to attach.** GeneReviews does not mention the palate anywhere; its clinical summary covers skin, facial features, fontanelles, hair, brain MRI, and the developmental/seizure/regression triad, and stops. So that band still rests on the systematic review's age-at-ascertainment association, and remains a coarse literature reading rather than a counted frequency. `notes` now says that explicitly, so the asymmetry with the three bands that did move reads as a decision rather than an oversight.

---

PR #10043, opened 2026-08-29
Add IFAP syndrome 1 (MONDO:0100213, MBTPS2)

### LINKED ISSUE #10038: Curate IFAP syndrome 1, with or without BRESHECK syndrome (MONDO:0100213)
Curate a dismech entry for **IFAP syndrome 1, with or without BRESHECK syndrome** ([MONDO:0100213](https://monarchinitiative.org/MONDO:0100213)) — X-linked ichthyosis follicularis, atrichia and photophobia, caused by *MBTPS2* variants. MBTPS2 is site-2 protease, the second cleavage step that releases SREBP and ATF6 transcription factors from the membrane, so hypomorphic alleles impair both cholesterol/fatty-acid homeostasis and the ER stress response — which is how one protease produces a skin-barrier, hair-follicle and corneal phenotype at once.

Provisional read: `DISEASE`, with BRESHECK as a `has_subtypes` entry — the MONDO label itself lumps them ("with or without"), and BRESHECK is described as IFAP plus brain, kidney and Hirschsprung involvement from more severe alleles, not a separate mechanism.

Note this is an allelic series: `kb/` already carries other *MBTPS2* diseases (`Keratosis_Follicularis_Spinulosa_Decalvans.yaml`, `Osteogenesis_Imperfecta_Type_XIX.yaml`). IFAP itself is not curated, and the genotype-phenotype split across those entities is worth stating explicitly in the entry.

Stub: `stubs/IFAP_Syndrome_1_With_Or_Without_BRESHECK_Syndrome.yaml`

aleixpuigb:
Curates ichthyosis follicularis, atrichia and photophobia syndrome, where one protease serves two unrelated transcriptional programmes and a hypomorphic allele degrades both at once. Closes #10038 and deletes the stub.

## The pathograph

```
Hypomorphic MBTPS2 variant
  └──► Reduced regulated intramembrane proteolysis
        ├── Impaired SREBP-driven lipid synthesis
        │     ├──► Defective epidermal and follicular keratinization  (PROVISIONAL)
        │     └──► Progressive vascularising keratopathy              (PROVISIONAL)
        └── Impaired ATF6 ER-stress response                          (PROVISIONAL)
```

Site-2 protease is not a pathway enzyme but a shared switch — it makes the membrane-embedded cut for SREBP (lipid synthesis) and for ATF6 (the ER stress response). That is why one gene gives skin, hair and eye disease together.

**Severity here is a dose, not a category.** Every reported allele is hypomorphic, and the gene-identification study measured residual protease activity for five patient variants in a complementation assay and found it correlated with clinical severity. That is why BRESHECK — the IFAP triad plus intellectual disability and multiple congenital anomalies — is a `has_subtypes` entry on the same continuum rather than a separate disease. The MONDO label lumps them too.

## The ATF6 arm is curated, and flagged as load-bearing nowhere

This is the finding I'd most like a reviewer to check. Every account of this disease, including this entry's own description, says S2P loss impairs both cholesterol homeostasis and the ER stress response. The first half is measured *and* connected to a phenotype. The second half is measured and connected to nothing:

- no clinical feature of IFAP is attributed to ATF6 loss specifically
- ER stress has not been measured in any patient tissue
- every phenotype curated here could be produced by the SREBP arm alone

So the node is graded PROVISIONAL and carries a `KNOWLEDGE_GAP` discussion saying so. It is kept rather than dropped for two reasons: a mechanism curated only where convenient is not a mechanism, and there is an asymmetry that argues against dismissing it — in the one variant where both arms were tested, the SREBP pathway was *attenuated* while the ER stress response *failed outright*. If that holds, the ATF6 arm is the more severely affected one, which makes its clinical silence more puzzling, not less.

## Three ontology terms the deep-research report got wrong

An OpenScientist report was run alongside and is committed under `research/`. It supplied most of the reference set, each verified on PubMed before use. Three of its ontology suggestions were rejected — all three on the evidence of **its own** term-validation section:

| Report proposed | Problem |
|---|---|
| `NCIT:C29496` for "Acitretin ~1 mg/kg" | Does not exist in NCIT. The treatment is instead curated as generic Pharmacotherapy plus a CHEBI `therapeutic_agent`, per CLAUDE.md. A dose does not belong in an ontology binding anyway — 1 mg/kg lives in the evidence snippet, where it is attributable |
| `HP:0033052` for "Angular cheilitis / periorificial keratotic plaques" | That identifier is **Non-epileptic seizure**. Binding it would have put a seizure phenotype into a genodermatosis |
| `HGNC:7375` for *MBTPS2* | Wrong identifier; the correct one is `hgnc:15455`, as the curation stub recorded |

This is the second PR in this run where the report's HGNC identifier was wrong and the stub was right.

## Evidence notes

- **The "40 cases reported" figure is curated but graded `PARTIAL`, and the reason is the source's shape rather than its content.** It comes from the introduction of a case report about *Hodgkin lymphoma in an IFAP patient* — a background sentence in a paper about something else, with no stated method. It is the only quantitative anchor available and a reader will meet it uncited elsewhere, so it is recorded with that caveat rather than dropped or laundered.
- **Limbal stem cell deficiency is banded `OCCASIONAL` on one patient, and the band records how often it has been *looked for*.** It may well be universal. The `notes:` say that, because a frequency band read as prevalence would be wrong in both directions here.
- **Two snippets reproduce their sources' typos** (`heterogeneicity`) rather than tidying them. The cache is the authority for a quote.
- **The carrier sign is graded `PARTIAL` on the source's own "may be".** Retinal vascular tortuosity in one carrier mother in one family is a lead for examining carriers, not a carrier test, and the entry says so.

## Lump/split, and a deliberate asymmetry

BRESHECK is a subtype; KFSD, X-linked Olmsted syndrome and osteogenesis imperfecta type XIX are not. The asymmetry is intentional: BRESHECK is the severe end of the *same* phenotype, while the others are *different* phenotypes that share a gene. Two of them already have dismech entries.

That raises the second discussion — why one gene produces a skin-and-eye disease in some families and a bone disease in others. The review literature is explicit that this is unresolved, which is worth recording so a curator doesn't assume the answer is known and merely unlooked-up. dismech is unusually placed to make progress on it, since it already holds entries for two of the four allelic diseases, and the proposed experiment is a residue-level cross-entry map that needs no new data.

The Olmsted differential is genuinely contested rather than merely difficult: the authors of the case cited argue the overlap "may challenge the existence of the X-linked form of Olmsted syndrome as an independent condition". That doubt is the substance of the differential, so it is curated as such rather than as a list of separating features.

## Validation

| Check | Result |
|---|---|
| Snippets verified | 40/40 |
| `just validate-history` / `just check-stubs` | pass |

Staging was targeted. The cache diff carries exactly the five CURIEs this entry introduced (`GO:0032933`, `HP:0031291`, `HP:0032107`, `HP:0500262`, `MONDO:0100213`).

---


aleixpuigb:
All three IMPORTANT findings and all four suggestions are fixed. Pushed as 2633fe6. 52/52 snippets verified, full gate battery clean.

**Finding 1 (SREBF1) — and your second reason for it was the more valuable one.** I added the differential, but the argument you made about the mechanism is now in the `ifap_atf6_arm_unattached_to_phenotype` discussion, not just in the differential list. SREBF1 encodes SREBP: the substrate of the protease this entry curates, one step downstream. A *dominant* SREBF1 disease that shares this much of the clinical spectrum is the strongest available argument that the SREBP arm carries the phenotype — and it comes from human genetics rather than a cell assay, which is what the ATF6 arm conspicuously lacks. That evidence item is graded `PARTIAL` against the ATF6 node with the reasoning spelled out: it narrows the arm's remaining explanatory role without excluding it.

There is now a proposed experiment that follows directly from your framing and needs no new laboratory work — code both phenotypes to HPO and subtract. Features in both are candidates for the SREBP arm; MBTPS2-only features are what the ATF6 arm would have to explain.

Two limits I recorded rather than glossed. The SREBF1 disease is dominant while IFAP is X-linked recessive, and it includes mucosal involvement IFAP does not — so the two are not the same phenotype with one mechanism subtracted, and the inference is suggestive rather than clean. Also, on naming: MONDO calls it `MONDO:0100221` "IFAP syndrome 2" while the clinical literature calls it hereditary mucoepithelial dysplasia. Same concept, two vocabularies, and a search on either alone misses the other — that is in the differential's description.

I added the GJB2 syndrome (PMID:30431684) in the same pass, as you suggested. It makes a useful point: ichthyosis follicularis is not specific to the SREBP pathway at all.

**Finding 2 (diagnosis).** Added, and you were right that its absence was conspicuous specifically because the entry's own description argues AS-OCT reframed the ocular problem. AS-OCT is now curated as a diagnostic with the same cached quote. Also added triad recognition (with non-scarring alopecia as the quality separating it from KFSD at the bedside) and MBTPS2 sequencing — with a practical note that exon-only analysis will miss the recurrent intronic variants c.671-9T>G and c.970+5G>A.

**Finding 3 (treatments).** Both added. Emollients and urea keratolytics matter for the reason you gave: as it stood the entry offered nothing for the skin except a systemic retinoid evidenced by one three-year-old, which misrepresents ordinary care. Amniotic membrane transplantation is graded `PARTIAL` and the item says outright that it was given *together with* acitretin, so neither can be credited alone — and that the reported improvement included neuropsychomotor development, which no ocular surface procedure explains and which suggests a global rather than eye-specific assessment. The prior artificial-tears-and-punctal-occlusion result is curated alongside it, because "small improvement" from purely surface-directed treatment is consistent with the surface not being where the lesion is.

**Suggestions.** All three applied:

- **Recurrent infections** — you were right that excluding the phenotype and excluding the mechanism are different decisions. `notes` now separates them. The mechanism is out because the NF-κB link rests on LPS experiments in an unrelated system; the *phenotype* is out because no source fetched here reports it with a denominator or in a named patient — it appears in review prose as a listed feature, which is a citation chain rather than an observation. Recurrent angular cheilitis, reported in one patient, is a different thing. Stated as a gap, not as a judgement that the feature is absent.
- **`would_support: "genetic#MBTPS2"`** — agreed it was semantically thin. Retargeted to `pathophysiology#Hypomorphic MBTPS2 Variant` and `pathophysiology#Reduced Regulated Intramembrane Proteolysis`.
- **PMID:19361614 grading** — regraded to `IN_VITRO`. Your reasoning holds and there is a sharper version of it: the sentence is about phenotypic expression in patients, but the quantitative relationship it asserts is *measured* by the CHO M19 assay, with clinical severity as the variable it is correlated against. Grading it `HUMAN_CLINICAL` would credit the ER-stress claim with human measurement that was never made — which is the exact thing that discussion exists to flag.

**Your GeneReviews caveat — checked, and your read was right.** `GeneReviews[Book]` against MBTPS2, "IFAP syndrome", "ichthyosis follicularis" and "keratosis follicularis spinulosa" returns nothing; only the general Ichthyosis Overview exists. Recorded in `notes`. Worth noting the same check found a **blocking** miss on the sibling PR #10042, where the chapter does exist and I had not looked — so flagging it as unverified rather than assuming was the right call.

---

PR #10059, opened 2026-08-29
Add Subcutaneous Panniculitis-like T-cell Lymphoma entry (MONDO:0019475)

### LINKED ISSUE #10053: Curate subcutaneous panniculitis-like T-cell lymphoma (MONDO:0019475)
Claiming `MONDO:0019475` (subcutaneous panniculitis-like T-cell lymphoma) from `stubs/Subcutaneous_Panniculitis-like_T-cell_Lymphoma.yaml`.

Deep research provider: `claude_code`.

caufieldjh:
New entry for subcutaneous panniculitis-like T-cell lymphoma (SPTCL), curated from `stubs/Subcutaneous_Panniculitis-like_T-cell_Lymphoma.yaml`. The stub is deleted in this PR.

## Why this disease is interesting to model

SPTCL is unusual among lymphomas in having a defined germline predisposition. Loss-of-function missense variants in `HAVCR2` (TIM-3, an inhibitory checkpoint receptor) are found in roughly half to 86% of patients depending on cohort, and they abolish TIM-3 surface expression. So the mechanism chain here deliberately puts the inherited immune lesion *upstream* of the clone rather than treating the clone as the origin:

```
Germline TIM-3 Loss of Function (MOLECULAR)
  -> Unrestrained Innate Immune Activation (CELLULAR)
       -> Clonal Cytotoxic Alpha-Beta T-cell Expansion (CELLULAR)
            -> Adipocyte Rimming and Lobular Panniculitis (TISSUE)
       -> Macrophage Hyperactivation and Hemophagocytosis (ORGANISM)
```

That ordering is not a stylistic choice — it is what the treatment data support. Immunosuppression outperforms cytotoxic chemotherapy, and JAK inhibition produces rapid remission in genotyped children who had failed corticosteroids, immunosuppressants and chemotherapy.

## Two knowledge gaps recorded rather than smoothed over

- **Is HAVCR2-mutant SPTCL neoplastic at all?** Two independent groups say in print that the ruxolitinib response calls the neoplastic nature into question; some HAVCR2-mutant children have panniculitis with no histologic evidence of lymphoma; and one patient has been reported with two *clonally unrelated* episodes, which is hard to square with a single transformed clone. The entry stays a lymphoma entry (that is its classification, and clonal TCR rearrangement is a diagnostic criterion), but the gap is written down.
- **What drives the HAVCR2 wild-type cases?** About half of an unselected national cohort. `UNC13D`, `PIAS3` and `KMT2D` mutations are more frequent there, but no unifying mechanism exists, so the entry has no upstream node for that group.

## Two wrong ontology terms caught

The deep-research report suggested two HPO CURIEs that do not mean what it said they meant. Both were replaced after checking the ontology directly:

| Report suggested | What the CURIE actually is | Used instead |
|---|---|---|
| `HP:0031477` "Subcutaneous nodule" | obsolete *Abnormal mitral valve morphology* | `HP:0001482` Subcutaneous nodule |
| `HP:0005517` "Hemophagocytosis" | *T-cell lymphoma/leukemia* | `HP:0012156` Hemophagocytosis |

## Frequency numbers are cohort-tagged on purpose

HAVCR2 carrier fraction runs from 51% (unselected Korean national cohort) to 86% (Thai paediatric series); HLH frequency runs from 17% (adult-weighted EORTC series) to 68% (that same paediatric series). Both extremes are recorded with their cohort in `case_fractions` and in `notes`, so neither gets read as the disease-wide number. No prevalence rate is recorded because none has been published — only SPTCL's share of NHL diagnoses (<1%).

## Scope

Alpha/beta phenotype only, per the 2008 EORTC decision to restrict the term. Primary cutaneous gamma-delta T-cell lymphoma is a `differential_diagnoses` entry, not a subtype.

## What was validated

- `just validate` — schema clean, **57/57 snippets** verified against cached references
- `just validate-disorders` (the batched pre-PR gate) — passed
- `just check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-term-cache-integrity`, `check-stubs`, `validate-history` — all clean
- `just preflight-dr research/... MONDO:0019475` — **PASS** (HAVCR2 mentioned 43 times; report OMIM 618398 matches MONDO's)
- `pytest tests/test_data.py` — one failure, `test_animal_model_mechanism_targets[CAPRIN1_Related_Neurodevelopmental_Disorder.yaml]`, which is **pre-existing on `main`** and untouched by this branch (`git diff origin/main` on that file is empty). Not fixed here.

Closes #10053

---


caufieldjh:
Rebased onto current `main` and fixed the CI failure.

The first run failed schema validation on four `supports: PARTIAL` evidence items. That was not a mistake in the entry so much as a collision: #10003 merged while this was being written and retired `PARTIAL`, splitting the old slot into `supports` (direction only) plus an optional `directness`.

Rather than take the bulk-migration default — `SUPPORT` with `directness` absent — I set `directness` on each of the four, since I had actually assessed them while writing:

- `inheritance` (heterozygous HAVCR2 odds ratio), `phenotypes` (68.2% HLH in the paediatric cohort), and `treatments` (chemo vs. IST remission rates) → `SUPPORT` / `DIRECT`. Each quote asserts the claim; `PARTIAL` there was recording "this complicates the picture", which is what the `explanation` is for.
- `treatments` radiotherapy → `SUPPORT` / `INDIRECT`. The only quotable line in that pooled review is a methods sentence saying radiotherapy experience was analysed, not a result. The claim follows by inference, which is exactly what `INDIRECT` is for.

The history record now says the same thing. `test (3.13)` is green on the rebased head; all local gates re-run clean, 57/57 snippets still verified.

---

caufieldjh:
Took up the three non-blocking items. This drops the approval, which is the right trade — the demographics gap was real.

**Demographics.** New `epidemiology:` section with both cohorts you named. Female-to-male 1.7 (European series) and 69.8% women (Korean national cohort) are recorded as separate evidence items so the agreement between two unrelated populations is visible rather than averaged away. Age at diagnosis is a second record, and the 46.5-vs-32-year gap is attributed to HAVCR2 genotype rather than geography, which `PMID:34535012` supports directly ("HAVCR2Y82C was associated with younger age").

**PMID:39288772.** Now cited as a third `case_fractions` entry (23/29 germline carriers), with a note that the study's own Japanese arm is n=6 and its p.Tyr82Cys figure correspondingly fragile — your point about relative cohort weight, written into the record rather than left for the next reader to work out.

**Uncited cache files.** Not deleted. A `notes` line now says what they are: output of the deep-research run's own citation-validation pass, committed because the report that cites them is committed, and not evidence for anything in the entry. That seemed more useful than a deletion that would have to be redone the next time a report is validated.

Left `GO:0044546` as it stands, for the reason you gave — the node description and `mechanism_confidence: PROVISIONAL` already say the step is inferred, and an unevidenced term that announces itself as unevidenced is more honest than dropping the mechanism.

`just validate` now reports **63/63 snippets verified**; all other gates re-run clean.

---

PR #10060, opened 2026-08-29
Add Lymphomatoid Granulomatosis entry (MONDO:0019466)

### LINKED ISSUE #10054: Curate lymphomatoid granulomatosis (MONDO:0019466)
Claiming `MONDO:0019466` (lymphomatoid granulomatosis) from `stubs/Lymphomatoid_Granulomatosis.yaml`.

Deep research provider: `claude_code`.

caufieldjh:
New entry for lymphomatoid granulomatosis (LYG), curated from `stubs/Lymphomatoid_Granulomatosis.yaml`. The stub is deleted in this PR.

*(Body updated after three review rounds to describe the entry as it now stands.)*

## The modelling decision that shaped this entry

LYG is classified as a B-cell neoplasm, but the literature does not describe it like one. There is no recurrent chromosomal lesion. Every patient in the reference NCI series had depressed circulating T cells with the deficit falling mainly on the CD8 compartment, and none had a diagnosed primary immunodeficiency. And the vascular damage that gives the disease its name is attributed by its own investigators to the host response — CXCL9/CXCL10 localise to reactive cells in viable tissue at the edge of necrosis, *not* to the malignant cells.

So the chain puts the immune defect upstream and the clone downstream of it:

```
Defective EBV Immunosurveillance (ORGANISM)   [conforms_to viral_oncogenesis#Persistent Oncogenic Virus Infection]
  -> Outgrowth of EBV-Positive Atypical B Cells (CELLULAR)
       -> Ineffective Angiocentric Reactive T-cell Response (TISSUE)
            -> Coagulative Necrosis and Multi-organ Tissue Destruction (TISSUE)
       -> Clonal Selection and High-Grade Transformation (CELLULAR)
```

## Grades as `has_subtypes`, not `stages`

WHO grades 1, 2 and 3 are bound to `MONDO:0859747` / `0859748` / `0859749`. Two reasons for `has_subtypes` over `stages`: MONDO models them as three distinct disease children, and they select *different treatment arms* rather than marking sequential points in one course. Grade is also not fixed for a patient — high-grade disease recurs as low-grade after chemotherapy, low-grade transforms after immune modulation — which is recorded in `progression` instead.

## Three knowledge gaps

- **The immune defect has never been identified.** The upstream node of this entry is a defect measured only in aggregate, hedged by its own authors as hypothesised. No causal gene, no functional assay, no per-patient biomarker.
- **The grading that picks the treatment arm is not reproducible.** The same pathology paper that defines the grading system says it "can be inconsistent and poorly reproducible among observers", with grade 1 hardest. That matters because grade decides interferon vs. chemoimmunotherapy, and serious adverse events run 25% vs. 64% between those arms.
- **`HUMAN_MODEL_MISMATCH`: no model reproduces the disease.** Humanized mice model EBV-plus-immunosuppression — and the discussion now cites two of them rather than asserting the gap — but not the angiocentric architecture, the lung/skin/CNS distribution, or the CD4-predominant microenvironment. The entry curates no `animal_models` as a result.

## Scope calls, written into `notes`

Three things this entry deliberately does not have, each with its reason recorded rather than left as an apparent omission:

- **No `genetic` section.** DOCK8, WAS and SH2D1A show up as predisposing conditions in case reports, but they cause their own syndromes of which LYG is one manifestation among several; listing them here would misrepresent them as causal genes for a disease that has none.
- **No `environmental` section.** The predisposing conditions are iatrogenic immunosuppression, HIV and germline immunodeficiency — none an ECTO-style exposure acting on a named mechanism node — and the one thing that would qualify, EBV, is curated as `infectious_agent`.
- **No separate molecular/latency node.** The mechanism turns on the host's failure to control a near-universal latent infection, so latency is annotated on the immunosurveillance node (`GO:0019042`) rather than given a node of its own.

## Two things caught during review worth naming

**Frequency bands must be over patients.** The cutaneous phenotype originally carried `OCCASIONAL` derived from 17% of *122 biopsies*. The FrequencyEnum bands are defined over patients, so the frequency is dropped and the denominator problem recorded in `review_notes`.

**Modality slots only work if they group consistently.** Interferon alfa-2b was tagged `PEPTIDE`; every other interferon treatment in the KB uses `OTHER`, so it now matches.

## Disease identity was checked by hand

`just preflight-dr` returned **SKIP** — MONDO records no causal gene for `MONDO:0019466`, so its gene-identity check cannot discriminate. Following the `dismech-references` skill, identity was confirmed manually against the MONDO record with `runoak`: the term's Orphanet-sourced definition matches the research report on the EBV aetiology, fourth-to-sixth-decade onset, lung/skin/CNS/kidney distribution, and the rarity of nodal and splenic involvement.

One small correction along the way: the pivotal trial paper (2023) describes NCT00001379 as still open for enrolment, but the ClinicalTrials.gov API now reports it COMPLETED, so that is what the `clinical_trials` entry records.

## What was validated

- `just validate` — schema clean, terms clean, **51/51 snippets** verified against cached references
- `just validate-disorders` (batched pre-PR gate) — passed
- `just check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-stubs`, `validate-history` — all clean
- `pytest -k "conform or module"` — 3752 passed (covers the `conforms_to` declaration)

Closes #10054

### DIFF
--- a/kb/disorders/Lymphomatoid_Granulomatosis.yaml
+++ b/kb/disorders/Lymphomatoid_Granulomatosis.yaml
@@ -521,9 +521,11 @@ phenotypes:
     Parenchymal brain lesions, with cranial neuropathies, seizures and focal
     deficits as their clinical expression. The central nervous system is the
     second most common site of disease after the lung and is an adverse
-    prognostic feature. The bound term covers the structural lesion, which is
-    what the cited cohort counted; the individual neurologic symptoms are not
-    separately enumerated in that source.
+    prognostic feature.
+  review_notes: >-
+    The bound term covers the structural lesion, which is what the cited cohort
+    counted. The individual neurologic symptoms are not separately enumerated in
+    that source, so no narrower term is used.
   phenotype_term:
     preferred_term: Morphological central nervous system abnormality
     term:
@@ -546,9 +548,11 @@ phenotypes:
     rash, present in roughly a sixth of biopsied sites. Cutaneous lesions show a
     dense lymphohistiocytic infiltrate with nonnecrotizing granulomatous
     inflammation and only sparse EBV-positive cells, so skin lesions alone cannot
-    establish the diagnosis. No `frequency` is recorded: the available figure
-    (17%) is a share of 122 biopsies, not of patients, and the FrequencyEnum
-    bands are defined over patients.
+    establish the diagnosis.
+  review_notes: >-
+    No frequency is recorded. The available figure, 17%, is a share of 122
+    biopsies rather than of patients, and the FrequencyEnum bands are defined
+    over patients.
   phenotype_term:
     preferred_term: Skin nodule
     term:
@@ -729,7 +733,7 @@ treatments:
     intervention augments the host response to EBV rather than killing the clone.
     In the phase 2 trial it produced complete responses in 61% of evaluable
     patients and was substantially better tolerated than chemotherapy.
-  therapeutic_modality: PEPTIDE
+  therapeutic_modality: OTHER
   treatment_term:
     preferred_term: Pharmacotherapy
     term:


caufieldjh:
Rebased onto current `main` and fixed the CI failure.

The first run failed schema validation on two `supports: PARTIAL` items — #10003 merged while this was being written and retired `PARTIAL`, splitting the old slot into `supports` (direction only) plus an optional `directness`.

Both were migrated with an explicit `directness` rather than left absent:

- Dyspnea → `SUPPORT` / `INDIRECT`. The quote asserts progressive pulmonary failure, not the symptom; breathlessness follows by inference. The `explanation` now says that instead of pointing at the retired grade.
- Hematopoietic stem cell transplantation → `SUPPORT` / `DIRECT`. The quote is directly about transplantation in this disease, including its own caveat that efficacy is not well established.

The history record records the migration. All local gates re-run clean, 49/49 snippets still verified.

---

caufieldjh:
Both suggestions taken up.

**Interferon modality.** You were right that `PEPTIDE` was out of step. Checked the KB: `Adult_T_Cell_Leukemia_Lymphoma` (Interferon-alpha and Zidovudine) and `Hepatitis_D` (Pegylated Interferon Alpha) both use `OTHER`. Changed to match — a modality slot is only worth having if it groups the same things the same way, and one entry deciding a recombinant cytokine is a peptide breaks that for everyone querying it.

**Rationale in descriptions.** Moved to `review_notes` on both the CNS and cutaneous phenotypes. `description` should say what the phenotype is; why a particular term was chosen, or why a frequency is absent, is a note about the curation.

`just validate` 51/51 verified; all other gates re-run clean.

---

PR #10062, opened 2026-08-29
Curate mal de Debarquement (MONDO:0016217)

### LINKED ISSUE #10058: Curate mal de Debarquement (MONDO:0016217)
Curate a dismech entry for **mal de Debarquement** ([MONDO:0016217](https://monarchinitiative.org/MONDO:0016217)) — a persistent sensation of self-motion (rocking, swaying, bobbing) that begins after exposure to passive motion, classically a sea voyage, and continues for months or years after the motion has stopped.

My reading, for review: this looks like a **Disease**. The stub records no MONDO descendants and no causal genes, both of which fit — it is a leaf term for an acquired condition, not a gene-defined or grouping concept. There is one reasonably conserved proposed pathograph (maladaptive readaptation of vestibular velocity storage entrained by passive motion), one consensus case definition, and a mechanism-directed treatment.

Two things I expect to need care:

- **Motion-triggered versus spontaneous onset.** The Bárány Society criteria distinguish MdDS following a motion trigger from an otherwise identical presentation with no identifiable trigger. Whether that is one disease with two onset routes or two entities is a live question in the literature, and it is the main lump/split call here rather than any gene.
- **The differential with persistent postural-perceptual dizziness (PPPD).** These overlap clinically and are separable mainly on the character of the illusion (rocking self-motion that *improves* while driving or in passive motion, versus unsteadiness that worsens with visual complexity). The entry should model that discriminator rather than blur it.

This entry will exercise parts of the schema the recent Mendelian curations do not: `environmental:` with `influences_mechanisms`, a `definitions:` block for the consensus criteria, and a treatment (the readaptation protocol) whose whole rationale is that it tests the proposed mechanism.

Stub: `stubs/Mal_De_Debarquement.yaml`

Phillip-a-richmond:
Adds `kb/disorders/Mal_De_Debarquement.yaml` — an acquired central vestibular disorder the KB did not cover.

Closes #10058

> **Round 1** (`280614448`) addressed all eight review items; every finding was correct. The most important was an ontology binding of mine — see "The term I got wrong" below, which corrects a claim an earlier version of this description made.
>
> **Round 2 approved**, with one non-blocking residual. I pushed that too (`872b97f4c`), which reset the approval on purpose — reasoning under "The one I pushed after approval".
>
> Branch note: this branch previously carried #9700 (merged). It was reset from `main` and force-pushed with lease; the discarded commits contained only already-merged content, and `main` is in fact ahead of them, carrying the #7439 `PARTIAL` migration of that work.

## Why this one is a useful entry to have

Almost everything I've curated recently has been Mendelian. This is not: no gene, no inheritance, an environmental trigger, and a consensus case definition doing the work a molecular diagnosis usually does. It exercises `definitions`, `environmental.influences_mechanisms`, `progression`, `epidemiology`, `clinical_burden` and `differential_diagnoses` rather than `genetic` and `has_subtypes`.

## The mechanism, and why the treatment is part of the argument

Five nodes: prolonged passive oscillatory motion exposure → velocity storage cross-axis maladaptation → persistent oscillatory self-motion perception → limbic-vestibular network reorganization → functional impairment.

The load-bearing observation is from roll-while-rotating conditioning in monkeys: the maladaptation appeared **only in animals with long VOR time constants**, and not when the time constant approached the cupula's. That is what localizes the lesion to central velocity storage rather than the end organ, and it is why the entry treats a normal vestibular work-up as part of the diagnostic picture rather than as a failure to find the lesion.

The therapy follows from the account rather than from trial and error — roll the head against a full-field rotating stimulus to readapt the reflex the other way — which makes its ~70% response rate evidence for the mechanism as well as a treatment. That inference is graded `directness: INDIRECT`.

Round 1 added the piece that argument actually needed: a **sham-controlled** optokinetic study reporting *"No placebo effect was recorded"* (PMID:30410464). If treatment response is offered as evidence for a mechanism, absence of a placebo effect is what stops the argument resting on expectation.

## The term I got wrong

An earlier version of this description highlighted that I refused `GO:0060013` — offered by the research report as "vestibulo-ocular reflex", actually `righting reflex` — and recorded why in `notes`. That still stands, and GO appears to have no VOR class, so those nodes name the process in prose.

But the review found that one section later I had bound the **cardinal phenotype** to `HP:0002321 Vertigo`, defined as *"an abnormal sensation of spinning"*, on a disease defined by its motion being **non-spinning** — with the phenotype's own `preferred_term` and evidence snippet both saying so. `HP:4000033 Non-spinning vertigo` exists; its definition transcribes the Bárány criterion and its HPO annotation names mal de Débarquement explicitly. Rebound after verifying against OLS.

Same class of error, caught in a machine's suggestion and then committed by hand. The generalisable point: a term whose *label* reads plausibly can still contradict the claim, and only the definition tells you — so the check has to apply to bindings I write, not just to bindings I receive.

## The one I pushed after approval

The approving review left a residual it called non-blocking: the `Female predominance and adult onset` record asserted a sex ratio and an age of onset while citing a sex-hormone review that reports neither.

I fixed it anyway, accepting the approval reset. An *additive* suggestion on an approved PR can wait; this was a claim-versus-evidence mismatch, which is the defect class the evidence discipline exists to prevent — a reader scanning the citation would take it as grounding the demographic. Two verified quotes now establish the claim, and the hormone paper is re-explained as a candidate mechanism for it rather than evidence of it.

One note for anyone acting on quoted snippets in reviews: one of the two suggested quotes failed an exact-substring check because the source uses an **en-dash** (`women (75–100%)`) where the review had a hyphen. Substantively right, invisible in rendered markdown, and indistinguishable from a fabricated quote at the grep. Re-derive snippets from the cache; don't paste them.

## Other judgment calls

**Temporal designations are `progression`, not `has_subtypes`.** The Bárány criteria define "in evolution", "transient" and "persistent". These are observation-window qualifiers on one disease — "transient" is diagnosable only retrospectively — so making them subtypes would assert three entities where there is one.

**The non-motion-triggered presentation is left out, as an open question.** The criteria require onset within 48 hours of passive motion, yet the same committee describes an otherwise indistinguishable presentation with no trigger. This entry's mechanism *starts* from an oscillatory conditioning stimulus, so extending it to cases that lack one would assert more than the evidence carries. Recorded as a `KNOWLEDGE_GAP` — and round 1 added the first concrete evidence bearing on it: the motion-triggered group responds better to the mechanism-directed therapy than the spontaneous group.

## Evidence model

First entry written under the post-#7439 model: `supports` carries direction only, with `directness` separate. `INDIRECT` marks the monkey conditioning data, the computational velocity-storage model, the treatment-to-mechanism inference, and the sex-hormone paper. One `REFUTE` item records a randomised trial where iTBS added **no** benefit over VOR rehabilitation.

## A defect worth naming

While writing the round-1 fixes I invented three `reference_title` values from memory — the third session running that I've done this, including titling a study of *pilots* as one of *flight attendants*. Rather than patch them individually I rewrote every `reference_title` in the file programmatically from its cache frontmatter; four instances across three PMIDs were wrong. Deriving the field beats typing it, and the correct strings were on disk throughout. The reviewer independently confirmed 15/15 exact match afterwards.

Related: [#10070](https://github.com/monarch-initiative/dismech/issues/10070) — CLAUDE.md's pre-PR gate list names six `check-*` recipes while `just qc` runs ten, and the four it omits include `check-snippet-boundaries`, which is why a mid-word snippet reached round 1. Rounds 1 and 2 used all ten.

## Scoping

Reference caches are the entry's citations plus the report's, since the report is committed here and its citation validation resolves against them. One on-topic but uncited file (`PMID:31580016`) was removed rather than committed as padding. The `cache/**.csv` rows are exactly what this entry introduced: `MONDO:0016217`, `HP:4000033` and `NCIT:C129862`, with their enum memberships — six additive rows, no deletions or reordering. No gene rows, correctly, since the disorder is acquired.

## Validation

| Check | Result |
|---|---|
| `just validate-disorders` | passed, 47/47 snippets verified |
| `just check-snippet-boundaries` | clean |
| reference titles vs cache | all match (derived, not typed) |
| `just check-entity-refs` / `check-duplicate-keys` | passed |
| all ten `just qc` gates | no new violations |
| pytest structural and conformance | 15,509 local; reviewer's full run 46,437 passed |

## Unrelated finding, not acted on here

While preflighting, I checked `MONDO:0009612` (methylmalonic aciduria due to methylmalonyl-CoA mutase deficiency) from the stub queue. It is already covered in substance by `kb/disorders/Methylmalonic_Acidemia.yaml`, a 1,774-line entry bound to the parent `MONDO:0002012` that carries the MMUT arm as its "Vitamin B12-unresponsive" subtype. That stub should be `entry_type: SUBTYPE`, but it is unrelated to this diff so I left it alone rather than widen the PR.

### DIFF
--- a/history/disorders/Mal_De_Debarquement/2026-09-01T193746Z-claude-code-9000b3.yaml
+++ b/history/disorders/Mal_De_Debarquement/2026-09-01T193746Z-claude-code-9000b3.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Mal_De_Debarquement
+  path: kb/disorders/Mal_De_Debarquement.yaml
+session:
+  id: 2026-09-01T193746Z-claude-code-9000b3
+  timestamp: '2026-09-01T19:37:46Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10062
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  - phenotypes
+  - evidence
+  summary: Address review round 3 on the MdDS entry
+  details: |
+    Took the three non-blocking suggestions from the round-3 review. Corrected the rTMS description, which paraphrased PMID:40228811 as finding statistically significant improvement in most studies when the source reports 3 of 7 on the Dizziness Handicap Inventory with no clinically significant improvement, and grounded the corrected claim in an evidence item quoting that sentence rather than leaving it as unattributed prose, graded INDIRECT because the review covers chronic vestibular disorders as a class. Aligned the Impaired Cognition preferred_term from Impaired concentration to Impaired cognition so it matches both HP:0100543 and the quoted snippet. Also closes the missing history record the advisory CI check flagged for the preceding commit. While making these edits the reference_title for PMID:40228811 was written from memory and was wrong; the programmatic cache audit caught it and every reference_title in the file was re-derived from cache frontmatter. Validation: 48/48 snippets verified, 48/48 titles match cache, all gates clean.
--- a/kb/disorders/Mal_De_Debarquement.yaml
+++ b/kb/disorders/Mal_De_Debarquement.yaml
@@ -370,7 +370,7 @@ phenotypes:
     Impaired concentration and cognitive slowing, reported by patients as "brain
     fog" and among the more disabling accompanying features.
   phenotype_term:
-    preferred_term: Impaired concentration
+    preferred_term: Impaired cognition
     term:
       id: HP:0100543
       label: Cognitive impairment
@@ -691,9 +691,12 @@ treatments:
     cortico-limbic node implicated by imaging rather than the vestibular lesion.
     A small double-blind sham-controlled crossover trial in eight women reported
     improvement in dizziness and in mood and anxiety persisting beyond the
-    treatment period. Evidence is limited and the effect sizes modest; a
-    scoping review of TMS across chronic vestibular disorders found
-    statistically but not clinically significant improvement in most studies.
+    treatment period. Evidence is limited and the effect sizes modest: a scoping
+    review of TMS across chronic vestibular disorders found statistically
+    significant improvement on the Dizziness Handicap Inventory in a minority of
+    studies (3 of 7) and no clinically significant improvement at all, with the
+    authors cautioning that methodological limitations bear on how the results
+    are read.
   therapeutic_modality: DEVICE
   treatment_term:
     preferred_term: Transcranial Magnetic Stimulation
@@ -715,6 +718,19 @@ treatments:
     explanation: >-
       The controlled trial result. Note the authors' own qualifier, "in selected
       individuals" - this is eight patients, not a general indication.
+  - reference: PMID:40228811
+    reference_title: "Transcranial magnetic stimulation use with chronic vestibular disorders: A scoping review."
+    supports: SUPPORT
+    directness: INDIRECT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Statistically significant improvements were noted on the Dizziness Handicap Inventory (3/7 studies) but clinically significant improvements were not observed."
+    explanation: >-
+      Bounds the claim rather than supporting it straightforwardly. Across chronic
+      vestibular disorders as a class, TMS reached statistical significance on the
+      Dizziness Handicap Inventory in a minority of studies and clinical
+      significance in none, which is why this treatment is recorded as limited
+      rather than established. Graded INDIRECT because the review covers chronic
+      vestibular disorders generally rather than this disease specifically.
 - name: Intermittent Theta Burst Stimulation as an Adjunct to VOR Rehabilitation
   description: >-
     Tested as a way to augment the readaptation protocol. It did not work: in a


Phillip-a-richmond:
All eight items addressed in `280614448`. Everything you flagged was correct; nothing to push back on.

## The term binding — you're right, and the irony is not lost

`HP:0002321 Vertigo` is defined as "an abnormal sensation of spinning", bound to a phenotype whose own `preferred_term` and evidence snippet both say *non*-spinning. Rebound to `HP:4000033`, verified against OLS first. Its HPO annotation reads *"This feature can be observed in mal de debarquement syndrome and in vestibular paroxysmia"* — so it is not merely a better term, it is the term for this disease.

Worth stating plainly: this is the same class of error the entry's own `notes` congratulate it for catching in the report's `GO:0060013` suggestion. I caught it in a machine-generated suggestion and then committed it myself one section later, by hand. The lesson I'd draw is that the discipline needs to apply to bindings I write, not just to bindings I receive — a term whose *label* reads plausibly can still contradict the claim, and only the definition tells you.

## Items 2–8

- **Epidemiology** — added, citing PMID:23674832 for the 41–46% migraine figure and PMID:34864753 for the sex-hormone proposal, graded `INDIRECT` since it concerns vestibular disorders as a class rather than MdDS. You were right that the migraine claim is load-bearing: it is the stated rationale for a treatment and part of a differential, and it was sitting in the description with no citation anywhere in the file.
- **Both trials added.** On the sham-controlled optokinetic study — I'd put it slightly more strongly than you did. Since this entry argues that treatment response *is* evidence for the mechanism, a study showing no placebo effect is not merely supporting evidence, it is what stops that argument resting on expectation. It's attached to the readaptation treatment for exactly that reason.
- **Phenotypes** — evidence added to the two uncited ones, plus the four named only in prose: visual motion intolerance, spatial disorientation, impaired cognition, depression. The first two are bound to `HP:0001751` with a note, since HPO has no specific class for either.
- **Suggestions 5–8** — all taken. The motion-triggered vs spontaneous treatment-response difference is now on the `KNOWLEDGE_GAP`, and you're right that it is the first concrete discriminator available: if a mechanism-directed therapy works less well without a motion trigger, that is evidence the two presentations do not share the lesion.

## Two defects of my own, found while fixing yours

**I invented three `reference_title` values from memory** — the third session running that I've done this. Rather than patch them individually I rewrote every `reference_title` in the file programmatically from its cache frontmatter; four instances across three PMIDs were wrong, including a paper about pilots that I had titled as being about flight attendants. Generating the field from the cache rather than typing it is the durable fix, and it is what I should have done from the start given the history.

**`check-snippet-length` caught a 4-word fragment** — I had copied `"persist beyond 6 months"` verbatim out of your suggestion rather than quoting the sentence. Expanded to the full clause. Worth noting for anyone reading a review as a to-do list: a quoted fragment in a review comment is a pointer, not a snippet.

## On the gate I missed last round

You flagged the mid-word truncation because `check-snippet-boundaries` caught it and I had never run that gate. The reason is documented as [#10070](https://github.com/monarch-initiative/dismech/issues/10070): CLAUDE.md's pre-PR list names six `check-*` recipes, `just qc` runs ten, and the four it omits include that one. This round I ran all ten, which is how the short snippet above was caught before push rather than by you.

## Validation

45/45 snippets verified (was 32), boundaries clean, all ten `qc` gates green, `validate-disorders`, `validate-terms`, entity refs, duplicate keys, `validate-history`, and 15,509 structural and conformance tests.

Cache diff is the two terms this round introduced — `HP:4000033` and `NCIT:C129862` — plus their enum membership. The five modified `references_cache` files are full-text upgrades from the validator, not hand edits.

---

Phillip-a-richmond:
Thanks for the approval. I've taken the residual too, in `872b97f4c` — which resets the approval, deliberately.

## Why I pushed on a non-blocking item

You framed it as optional and gave good reasons: the claim is true, the description hedges it, and the `INDIRECT` grading is honest. I'd have left an *additive* suggestion alone on an approved PR for exactly that reasoning.

But this one isn't additive. The record is titled `Female predominance and adult onset`, asserts a sex ratio and an age of onset, and cites a paper that reports neither. That's a claim-versus-evidence mismatch, and it's the defect class the whole evidence discipline exists to prevent — a reader scanning the citation would reasonably take it as grounding the demographic. Two verified quotes and one re-written explanation is a cheap fix for that, and a re-review cycle is a cheap price.

Both items you supplied are now in, plus the hormone paper re-explained as a candidate mechanism for the predominance the other two establish, rather than as evidence of it. `PMID:32986636` also does something the other doesn't: it ties the female predominance specifically to the *persistent* form, which matters given how much of this entry turns on separating that from the near-universal transient after-effect.

## One note on the second quote

`PMID:32986636` renders it with an **en-dash** — `women (75–100%)` — not a hyphen. My first grep for your version returned nothing, and I nearly concluded the quote didn't exist before widening the search.

Not a criticism; it's a transcription artifact of the kind that's invisible in rendered markdown. But it's worth knowing that a suggested snippet can be substantively right and still fail an exact-substring check on one character, and that the failure mode looks identical to a fabricated quote. Anyone acting on a quoted snippet from a review comment should re-derive it from the cache rather than copy it — which is the same lesson as the four-word fragment I copied verbatim from your last round.

## Verification

47/47 snippets verified (was 45), all ten `just qc` gates green, terms/entity-refs/duplicate-keys clean. No `cache/` change — both PMIDs were already committed and cited.

---

Phillip-a-richmond:
Thanks. I'm not pushing the perimenopausal clause, and I want to be explicit about where I'm drawing the line, since I *did* push the last non-blocking item you raised.

Both quotes you supplied verify against the committed caches — `PMID:27730651` *"occur more commonly in females around menopause time"* and `PMID:32986636` *"Hormonal influences such as the peri-menopausal or peri-menstrual state at the time of exposure may also be a risk factor"*. So the finding is correct and the fix is genuinely one line. Recording that here so whoever picks it up doesn't re-derive it.

## Why this one is different from the last one

Last round the record's **title and central claim** — "Female predominance and adult onset" — rested on a citation that measured neither. A reader scanning it would have taken the citation as grounding the demographic, which is the claim-versus-evidence mismatch the evidence discipline exists to prevent. That was worth an approval reset.

This one is a clause of contextual prose inside a `description`, on a record whose central claim is now properly evidenced three ways. Descriptions across this KB routinely carry framing beyond what each individual evidence item states — that is what distinguishes a `description` from an `explanation`. A rule requiring every description clause to carry its own citation would fail most entries in `kb/`, and I don't think it's the rule.

## The other reason

Two approvals have now been reset for progressively narrower refinements of the same record: round 2 approved with residual A, I fixed A, round 3 approved with residual B where B is a strictly narrower version of A. That is the shape of a non-converging review loop, and the guidance I work under says to stop pushing at that point and raise once what is still flagged rather than keep cycling.

So: flagged, verified, not fixed. If a maintainer wants it in, it is two `evidence` items and I'll add them — but I don't think another approval cycle is the right way to spend the review budget on a hedged prose clause, and you said the same.

The entry is approved with CI green, so from my side it's done. Happy to be overruled.

---

Phillip-a-richmond:
Refreshed this branch onto current `main` (`872b97f4c` → `47fd71612`). This resets the approval, which I want to explain, because the reason is not a change to the content.

**This PR could not have been merged in the state it was in.** It was **130 commits behind `main`**, and the deterministic auto-merge sweep requires `behind_by == 0` with the current `main` SHA proven to be an ancestor of the PR head. So it was approved, green and unassigned since Friday, and eligible on every criterion except the one nobody looks at. It would have sat there indefinitely. Being behind its base is a mergeability problem, and on a PR I opened that is mine to fix rather than wait out.

**The diff is unchanged.** Rebuilt from `origin/main` and cherry-picked all four commits; every one applied cleanly, with the six shared `cache/**.csv` files auto-merging. The result is byte-for-byte the same shape as before: 35 files, 6,170 insertions, 25 deletions. No unrelated deletions, no protected-path churn.

Re-validated against the new base rather than assuming the old run still holds, which matters here because the `PARTIAL` retirement (#10003) landed inside those 130 commits:

| Check | Result |
|---|---|
| `just validate-disorders` | passed, 47/47 snippets verified |
| `just check-enum-values` | no out-of-enum values — confirms this entry is clean under the retired-`PARTIAL` schema |
| `just check-causal-targets` | no new broken targets |
| `just check-qualifier-terms` | passed |

Nothing about the entry itself changed, so the two approving reviews were reviewing the same content that is here now. Re-requesting review only because the push dismissed the approval, not because there is new work to look at.

---

Phillip-a-richmond:
Took all three suggestions (`cc7c9e238`). Two of them are accuracy defects rather than matters of framing, which is why I pushed rather than banking the approval.

**Suggestion 1 was right, and I verified it against the cache before acting.** `PMID:40228811` says *"Statistically significant improvements were noted on the Dizziness Handicap Inventory (3/7 studies) but clinically significant improvements were not observed."* My prose said "in most studies". Three of seven is a minority, so the sentence misreported a named source — and unlike the perimenopausal clause from last round, this is a specific quantitative claim about what a paper found, not framing around one.

Rather than just fix the adjective, I cited the source as an evidence item quoting that sentence. A quantitative claim about a named study sitting unattributed in a `description` is the shape of the problem, not just the wrong number in it. Graded `INDIRECT`, because the review covers chronic vestibular disorders as a class rather than this disease.

Also worth noting what else that source says: postural control improved in **7/7** studies. That may be where "most studies" came from in my head — a real finding on a different outcome measure, attached to the wrong one.

**Suggestion 2:** `preferred_term` aligned to *Impaired cognition*, matching both `HP:0100543` and the quoted snippet. I kept `HP:0100543` rather than moving to `HP:0031987 Diminished ability to concentrate` — cognition is what the source says, so the narrower term would be assuming something the quote doesn't support.

**Suggestion 3:** history record added.

## The reference-title problem again

Writing the new evidence item, I typed its `reference_title` from memory: *"Effect of Transcranial Magnetic Stimulation on Chronic Vestibular Disorders: A Scoping Review."* The actual title is *"Transcranial magnetic stimulation use with chronic vestibular disorders: A scoping review."*

That is the fourth session running I have made this exact mistake, and this time it was inside the commit fixing an accuracy defect. The programmatic cache audit caught it before the push, and I re-derived every `reference_title` in the file from cache frontmatter rather than patching the one instance. 48/48 now match. The audit is the thing to trust here; my recall of a title is not, and I no longer treat "I'm fairly confident of this one" as a reason to skip the check.

## Validation

48/48 snippets verified (up from 47) · 48/48 reference titles match cache · `validate-disorders` ✅ · `check-enum-values` ✅ · `check-entity-refs` ✅ · `check-duplicate-keys` ✅ · `check-causal-targets` ✅ · `check-snippet-boundaries` clean on this file · `check-snippet-length` / `-title` / `-grading` / `check-reference-titles` no new violations · `validate-history-all` 7,966 records, no issues.

## On the thing you closed

Thanks for closing the perimenopausal thread explicitly rather than letting it sit. That distinction — claim-versus-evidence mismatch on an evidence-bearing record blocks, framing prose in a `description` does not — is the one I'd want applied to suggestion 1 too, and I think it lands on the other side of the line: a number attributed to a named paper is a claim about that paper, wherever it sits in the file.

---

Phillip-a-richmond:
Rebased again onto current `main` (`cc7c9e238` → `601d75059`). No content change; all five commits replayed cleanly, 36 files / 6,213+ / 25- unchanged, 48/48 snippets and `check-enum-values` re-verified against the new base.

**Why a second rebase in one evening, and a note for whoever owns the merge machinery.** `main` moved four commits between the approval at 19:50 and 20:00, which put this PR back to `behind_by == 4` and therefore back outside what the deterministic sweep will merge. That is the same condition that had it stuck for three days at `behind_by == 130` — it just recurs now on a ten-minute cycle instead of a three-day one.

This is a treadmill, and I don't think I can win it by rebasing: a rebase resets the approval, the re-review takes roughly ten to forty minutes, and `main` in this repo advances faster than that. CLAUDE.md already names the fix — *"Higher safe throughput needs a real merge queue (plus `merge_group` CI), not wider update batches"* — and this PR looks like a concrete instance of the gap that text anticipates, so I'm recording it here rather than opening a separate issue on someone else's design decision.

I'm not going to keep rebasing on a loop. If this falls behind again after the next approval, the useful interventions are all outside what I should do unilaterally: a maintainer merging it directly, or the shepherd's approved-behind lane picking it up.

**On the postural-control suggestion** — agreed, and not acting on it. You're right that quoting only the DHI half is a selective read now that the entry cites the review; you're also right that the DHI is the patient-reported disability measure and postural control the physiological proxy, so the weighting is defensible. I'd flagged the 7/7 figure myself in the previous comment, so it's in the thread for a future editor either way. Restating it in the file would mean a fifth push to say something already visible here.

---

PR #10072, opened 2026-08-29
curate: Add Peroxisome Biogenesis Disorder 4A (Zellweger) (MONDO:0013930, PEX6)

### LINKED ISSUE #10064: Curate peroxisome biogenesis disorder 4A (Zellweger) (MONDO:0013930)
Curate a dismech entry for **peroxisome biogenesis disorder 4A (Zellweger)** ([MONDO:0013930](https://monarchinitiative.org/MONDO:0013930)) — the severe, classic-Zellweger end of the PEX6-related Zellweger spectrum, caused by biallelic *PEX6* variants that abolish the AAA-ATPase receptor-export module recycling PEX5 and so collapse peroxisomal matrix protein import.

Preflight: `kb/disorders/Peroxisome_Biogenesis_Disorder_4B.yaml` (the milder PEX6 allelic entry) already exists and names PBD4A explicitly as the severe counterpart it is distinguished from, so a separate 4A entry follows the pattern already established by `Peroxisome_Biogenesis_Disorder_3A_Zellweger`, `_11A_Zellweger` and `_1B`. No MONDO:0013930 hit anywhere in `kb/` on `origin/main`, and no open PR.

Stub: `stubs/Peroxisome_Biogenesis_Disorder_4A_Zellweger.yaml`

aleixpuigb:
Adds the PBD4A entry — the severe, neonatal-lethal end of the PEX6-related Zellweger spectrum.

## Why this entry, and why now

The KB already carries `Peroxisome_Biogenesis_Disorder_4B`, whose own description names PBD4A explicitly as the severe counterpart it is defined against, plus the severity entries `_3A_Zellweger` (PEX12) and `_11A_Zellweger` (PEX26) at other loci. PBD4A was the remaining gap in that set. It holds its own MONDO and OMIM identity, and the A/B split is a mechanistic statement about residual peroxin function rather than a severity label: the severe end is a congenital malformation syndrome, the mild end a progressive degenerative one. That reasoning is written into the entry's `notes`.

## What the pathograph says

Biallelic null *PEX6* → failure of the PEX1–PEX6 receptor export module → collapse of peroxisomal matrix protein import (the "peroxisomal ghost") → four downstream arms: loss of fatty acid β/α-oxidation, plasmalogen synthesis failure, C27 bile-acid accumulation with cholestatic liver injury, and the prenatal cortical neuronal migration defect.

The gene-agnostic cascade shared by the whole spectrum stays on the `Zellweger Spectrum Disorders` entry rather than being duplicated here, following the precedent PBD4B set.

## Three things worth a reviewer's eye

**Cholic acid is curated, and scoped.** There *is* a pharmacological therapy with phase 3 evidence in ZSD, so "management is entirely symptomatic" would have been wrong. It is curated with a `target_mechanisms` link to the bile-acid node only, and the entry says explicitly that it does not touch the brain malformation and does not change the prognosis of this presentation.

**The negative DHA trial is curated as a `REFUTE` item.** DHA is low in ZSD because its last synthetic step is peroxisomal, which makes replacement an obvious idea; a double-blind RCT found no effect. Recording it stops the hypothesis being re-derived from the biochemistry and presented as promising.

**The migration link is `INDIRECT_UNKNOWN_INTERMEDIATES`, and the knowledge gap explains why.** Liver-selective peroxisome rescue in Pex5 knockout mice corrects the migration defect *without changing brain VLCFA, DHA or plasmalogen at all* — so the obvious candidate mechanism is ruled out and an extraneuronal contribution is doing part of the work.

## Deep research

`openscientist`, committed at `research/Peroxisome_Biogenesis_Disorder_4A_Zellweger-deep-research-openscientist.md`. It contributed the PEX6-specific genotype–severity source, the cholic acid trial, the negative DHA trial, and the Pex5 liver-rescue experiment. Three of its claims were checked and **not** used:

- its suggested disease mapping `MONDO:0009279`, which the report's own term-validation section flags as *triple-A syndrome*;
- `PMID:28742939`, cited for the human PEX1–PEX6–PEX26 mechanism but actually an *Arabidopsis* paper;
- its CHEBI/HP suggestions for VLCFA and phytanic acid, which resolve to unrelated terms.

## Evidence honesty

The entry's `notes` states plainly that the clinical detail is inherited from descriptions of the severe end of the spectrum as a whole, not from a PBD4A series — none exists. Only the PEX6-specific claims are anchored on PEX6 sources, and the two therapy trials were ZSD-wide and not stratified by gene. There is deliberately no prevalence block; a knowledge gap records the study that would supply one.

## Validation

- `just validate` — schema clean, 46/46 snippets verified against cached references

Closes #10064

---

### DIFF
--- a/history/disorders/Peroxisome_Biogenesis_Disorder_4A_Zellweger/2026-08-29T175531Z-claude-code-7e458c.yaml
+++ b/history/disorders/Peroxisome_Biogenesis_Disorder_4A_Zellweger/2026-08-29T175531Z-claude-code-7e458c.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Peroxisome_Biogenesis_Disorder_4A_Zellweger
+  path: kb/disorders/Peroxisome_Biogenesis_Disorder_4A_Zellweger.yaml
+session:
+  id: 2026-08-29T175531Z-claude-code-7e458c
+  timestamp: '2026-08-29T17:55:31Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10072
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - animal_models
+  - biochemical
+  - pathophysiology
+  summary: 'Review round 2: model relationship semantics and biomarker readout links'
+  details: |
+    Second review round. Deleted the second modeled_mechanisms link on the Pex5 tissue-selective-rescue mouse, the one targeting Loss of Peroxisomal Fatty Acid Oxidation with relationship FAILS_TO_RECAPITULATE. The reviewer was right that the value asserts the model does not reproduce the target node, which is false - the Pex5 knockout does lose peroxisomal beta-oxidation, and the link's own description said so. What actually fails is the lipid lesion's explanatory role for the migration phenotype, which is a claim about an edge, not about a node. That dissociation is already carried structurally by the Brain VLCFA, DHA and plasmalogen after liver-selective rescue readout on the migration link, so the second link was redundant as well as wrong. Also converted the two pathophysiology downstream edges that pointed at biochemical markers into Biochemical.readouts links with relationship READOUT_OF and endpoint_context DIAGNOSTIC. The reviewer suggested Biochemical.attaches_to, which does not exist on that class, but Biochemical.readouts is the slot that expresses the same thing and its schema comment says readout links are observational rather than causal. The graph stays one component because biochemical.readouts is lifted into the graph. Removed a stray blank line before notes. Revalidated: just validate 55/55 snippets, validate-terms, check-entity-refs, check-duplicate-keys, check_reference_titles.py --check, and pytest -k 'model_readout or failure_to_recapitulate or entity_ref or biomarker' 10644 passed.
--- a/kb/disorders/Peroxisome_Biogenesis_Disorder_4A_Zellweger.yaml
+++ b/kb/disorders/Peroxisome_Biogenesis_Disorder_4A_Zellweger.yaml
@@ -312,12 +312,6 @@ pathophysiology:
       id: GO:0001561
       label: fatty acid alpha-oxidation
     modifier: DECREASED
-  downstream:
-  - target: Very-long-chain fatty acids
-    causal_link_type: DIRECT
-    description: >-
-      The block is what the first-line plasma assay measures; elevated C26:0 is this
-      node's readout rather than a separate lesion.
   evidence:
   - reference: PMID:26627182
     reference_title: "Zellweger spectrum disorders: clinical overview and management approach."
@@ -344,11 +338,6 @@ pathophysiology:
       id: GO:0008611
       label: ether lipid biosynthetic process
     modifier: DECREASED
-  downstream:
-  - target: Erythrocyte plasmalogens
-    causal_link_type: DIRECT
-    description: >-
-      Low erythrocyte plasmalogen is the direct measurement of this synthetic block.
   evidence:
   - reference: PMID:26627182
     reference_title: "Zellweger spectrum disorders: clinical overview and management approach."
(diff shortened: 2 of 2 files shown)


aleixpuigb:
All three blockers fixed, and all five suggestions taken. Pushed in 4a4fd500b.

**Blockers**

1. **`references:` block added**, tagging `PMID:20301621` with `tags: [GeneReviews]`, matching the 3A pattern. All eleven references the entry cites are listed.
2. **Pex5 liver-rescue mouse is now an `animal_models` entry.** Two `modeled_mechanisms` links: `RECAPITULATES` on the migration node, and — the one that matters — `FAILS_TO_RECAPITULATE` against `Loss of Peroxisomal Fatty Acid Oxidation`, recording that the model does not reproduce the lipid lesion's *explanatory* role. You were right that the negative readout was the whole point of the knowledge gap and was surviving only as prose; unchanged brain VLCFA/DHA/plasmalogen alongside corrected migration is now a structural `UNCHANGED` readout with its own evidence.
3. **Bile-acid node split** into `C27 Bile Acid Intermediate Accumulation` (MOLECULAR, `GO:0008206` DECREASED) → `Cholestatic Liver Injury` (TISSUE). Your point about the cholic acid link was the useful part: the drug acts on the accumulation by feedback suppression of synthesis, which is what the `target_mechanisms` description already said, and the link now points there.

**Suggestions**

4. Dead-end arms connected. The fatty-acid-oxidation and plasmalogen nodes now reach their biochemical readouts, and the liver injury node reaches `Hepatomegaly and Neonatal Cholestasis` and `Hyperbilirubinemia`. The graph is one component.
5. **Purkinje cell binding removed** — you are right, and it was a real error rather than a scope call. The Crane sentence attributes *arborisation* abnormality to Purkinje cells, which is a dendritic defect, not a migration failure. The node's explanation now says this explicitly instead of carrying a cerebellar cell type on a node named for the cortex.
6. **`evidence_source` rule rewritten.** You caught a genuine inconsistency. The fix is to state the axis correctly: it is not primary-versus-review, it is what kind of evidence the quoted sentence reports. `OTHER` where the source compiles expert consensus with no cohort behind the quoted sentence; `HUMAN_CLINICAL` where the sentence reports observations of patients — which is why the Klouwer grading stands, since every sentence quoted from it describes what ZSD patients present with.
7. Two of the three deep-research omissions taken: the C26:0-LPC newborn screen is now a `diagnosis` entry (`directness: INDIRECT` — the pilot reports no PBD4A case), and pexophagy from `PMID:41787707` is recorded as a `genetic` note. **Not** taken as a pathophysiology node, deliberately: the review reports pexophagy for peroxisome biogenesis disorders generally, and no measurement in a PBD4A genotype was found, so a node would assert more than the evidence carries. The note says so. F011 (mitochondrial/ROS amplification) is left on the spectrum-wide entry per the stated scoping.
8. **CHEBI checked** — there is no plasmalogen term via the OLS adapter. The `ether lipid` parent binding is therefore deliberate rather than a placeholder, and the biomarker now carries a `notes` line saying the search was run and what it returned.

Revalidated: `just validate` (55/55 snippets verified), `just validate-terms`, `just check-entity-refs`, `just check-duplicate-keys`, `scripts/check_reference_titles.py --check`, and the entity-reference pytest selection (20414 passed).

A second history record documents this round.

---

aleixpuigb:
Took your preferred option: the second `modeled_mechanisms` link is deleted. Pushed in d91300098.

You were right, and the way you were right matters more than the fix. `relationship` is a claim about the target node; the sentence I wrote in `description` was a claim about an edge. I noticed the tension while writing it — that is what the description was doing there — and then resolved it by explaining the enum value instead of changing it. Prose that argues with the structured value next to it is the signal that the structured value is wrong, and I had the signal and read it backwards. The dissociation was already carried by the `UNCHANGED` readout on the migration link, so the link was redundant as well as wrong.

I did not take `MEASURES`, for the same reason. It would have been true, but the same measurement is already on the migration link, where the negative result is load-bearing. Two links quoting one sentence to make one point is worse than one.

Your note about the orphaned `HUMAN_MODEL_MISMATCH` signal is the useful diagnostic here. `FAILS_TO_RECAPITULATE` with a plain `KNOWLEDGE_GAP` on the other end should probably be a check rather than something a reviewer has to spot — the two are documented as a pair in CLAUDE.md, and nothing enforces the pairing. I have not filed that yet; if you agree it is worth a test I will.

**On the biomarker edges.** Taken, with one correction to your suggestion: `Biochemical` has no `attaches_to` slot — the schema puts `attaches_to` only on `Definition` and `Discussion`. But `Biochemical.readouts` is the slot you were reaching for, and its own schema comment says exactly what you said: "Readout links are observational/associative, not causal disease-progression edges." So both edges are now `readouts` with `relationship: READOUT_OF` and `endpoint_context: DIAGNOSTIC`, and the `causal_link_type: DIRECT` claims are gone.

This is strictly better than what I had rather than a lateral move, because `biochemical.readouts` is one of the two things CLAUDE.md says *is* lifted into the graph and cx2. The connected graph you asked for last round survives, with the edge semantics no longer lying about it.

Stray blank line before `notes:` removed.

Revalidated: `just validate` (55/55 snippets), `just validate-terms`, `just check-entity-refs`, `just check-duplicate-keys`, `scripts/check_reference_titles.py --check`, and `pytest -k "model_readout or failure_to_recapitulate or entity_ref or biomarker"` (10644 passed). History record added for the round.

---

PR #10073, opened 2026-08-29
curate: Add Charcot-Marie-Tooth Disease Recessive Intermediate D (MONDO:0014467, COX6A1)

### LINKED ISSUE #10065: Curate Charcot-Marie-Tooth disease recessive intermediate D (MONDO:0014467)
Curate a dismech entry for **Charcot-Marie-Tooth disease recessive intermediate D** ([MONDO:0014467](https://monarchinitiative.org/MONDO:0014467)) — CMTRID, an autosomal recessive intermediate CMT caused by biallelic *COX6A1* variants, which sits at the junction of two families dismech already curates heavily: the intermediate Charcot-Marie-Tooth neuropathies and the mitochondrial complex IV (cytochrome c oxidase) assembly/subunit deficiencies.

Preflight: no hit for MONDO:0014467 or COX6A1 anywhere in `kb/` on `origin/main`, and no open PR. Note the two MONDO parents (`MONDO:0016387` mitochondrial oxidative phosphorylation disorder and `MONDO:0017058` autosomal recessive intermediate CMT) — the entry should be reachable from both the `Charcot-Marie-Tooth_Diseases` and `Mitochondrial_Complex_IV_Deficiency` groupings, which is worth checking as part of the curation.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D.yaml`

aleixpuigb:
Adds CMTRID, the *COX6A1*-related recessive intermediate Charcot-Marie-Tooth neuropathy.

## Why it is its own entry

It sits at the junction of two families the KB already curates heavily, and it is the odd one out in both: the only CMT entry here whose trigger node is a mitochondrial respiratory-chain subunit, and the only complex IV deficiency here whose dominant tissue is peripheral nerve rather than brain, heart or muscle. MONDO gives it both parents for that reason. It conforms to the `complex_iv_assembly_deficiency` module at three nodes rather than restating that cascade.

## The interesting question, and how it is curated

Why does a defect in a *housekeeping* OXPHOS subunit present as a peripheral neuropathy instead of the encephalopathy or cardiomyopathy typical of COX deficiency? The candidate answer is isoform substitution: mature cardiac and skeletal muscle switch from ubiquitous COX6A1 to muscle-specific COX6A2, and nerve does not.

That gets its own node, held at `mechanism_confidence: HYPOTHETICAL`, with the line between what is known and what is assumed drawn explicitly. Forced COX6A2 expression restores holoenzyme and activity in COX6A1-knockdown human cells, so the substitution is real biochemistry. What has never been done is the measurement *in the disease* — no CMTRID patient tissue has been checked for COX6A2, and the Cox6a1-null mouse's COX activity was reported without a tissue breakdown. That is filed as a `HUMAN_MODEL_MISMATCH` discussion with a proposed experiment that would settle it either way.

## The evidence base is small, and the entry says so

Three families across two papers, one mouse model, one isoform-biology paper that is not about this disease. There is no case series, no natural history, no reported nerve conduction velocities. Consequences visible in the diff:

- phenotypes supported only by a **group-level** CMT definition rather than by CMTRID-specific observation carry `directness: INDIRECT` instead of being presented as established;
- the gait phenotype is supported only by the mouse and is graded `MODEL_ORGANISM` accordingly, with a note saying that is not sufficient support on its own;
- **no prevalence block**, and **no drug therapy asserted** — importing a generic mitochondrial cofactor cocktail would have been unsupported;
- a knowledge gap asks whether the entity is even one disease: the founding families had an adult-compatible neuropathy, while the 2025 case had developmental delay, raised lactate and death at 3.5 y after a febrile decompensation. Those phenotypes are marked `OCCASIONAL` and flagged rather than folded into the core description;
- a third gap notes that the entity is catalogued as *intermediate* while every published description calls it axonal or mixed, and no paper reports the conduction velocities the category is defined by.

## Deep research

`openscientist`, committed alongside. Its most valuable contribution was `PMID:20307258` (the COX6A2 rescue experiment), which replaced a weaker mouse-muscle citation as the load-bearing evidence for the isoform node.

One of its citations is a trap worth flagging, and it is recorded in the entry `notes`: it cites `PMID:7687470` for COX6A1 being the ubiquitous isoform. That paper is titled *"…the heart/muscle isoform of cytochrome c oxidase subunit VIa"* and calls the heart/muscle gene COX6A1 throughout — 1993 bovine nomenclature, the reverse of the modern human numbering. The citation supports the opposite of the claim attached to it, and it is not used here.

Two further findings from the report are real but **uncitable in this KB** and are recorded as such rather than used: the HPO annotation of onion-bulb formation for OMIM:616039 (which would largely answer the intermediate-designation gap, but has no underlying PMID to quote), and gnomAD constraint metrics consistent with tolerated heterozygous LoF.

## Validation

- `just validate` — schema clean, 30/30 snippets verified against cached references

Closes #10065

---

### DIFF
--- a/history/disorders/Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D/2026-08-29T175543Z-claude-code-2b7734.yaml
+++ b/history/disorders/Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D/2026-08-29T175543Z-claude-code-2b7734.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D
+  path: kb/disorders/Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D.yaml
+session:
+  id: 2026-08-29T175543Z-claude-code-2b7734
+  timestamp: '2026-08-29T17:55:43Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10073
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - evidence
+  summary: 'Review round 2: evidence_source regrade to resolve snippet-grading conflict'
+  details: |
+    Second review round, one item. Regraded the PMID:25152455 functional-analysis evidence item on the first pathophysiology node from HUMAN_CLINICAL to IN_VITRO. just check-snippet-grading flagged that the same sentence was graded two ways in one file - the full sentence HUMAN_CLINICAL here, its tail IN_VITRO on the next node. The reviewer's read of the merits is correct: both assays the sentence describes are ex vivo, transcript abundance in isolated peripheral white blood cells and COX activity in EBV-transformed lymphoblastoid lines, and CLAUDE.md places cultured cells and biochemical assays outside an organism under IN_VITRO regardless of whether the material came from a patient. The human origin of the material is not what evidence_source records. Extended the explanation to say why the grading is what it is, so the next curator does not read HUMAN_CLINICAL back in. Regrading rather than splitting because the prefix clause stops before its verb and would not be propositional alone. Revalidated: just validate 36/36 snippets, check-snippet-grading, check-entity-refs, check-duplicate-keys.
--- a/kb/disorders/Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D.yaml
+++ b/kb/disorders/Charcot-Marie-Tooth_Disease_Recessive_Intermediate_D.yaml
@@ -129,14 +129,18 @@ pathophysiology:
   - reference: PMID:25152455
     reference_title: A mutation of COX6A1 causes a recessive axonal or mixed form of Charcot-Marie-Tooth disease.
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: IN_VITRO
     snippet: >-
       Functional analysis showed that expression of COX6A1 in peripheral white blood
       cells from the affected individuals and COX activity in their
       EB-virus-transformed lymphoblastoid cell lines were significantly reduced.
     explanation: >-
       Demonstrates in patient material that the allele reduces both transcript
-      abundance and the enzymatic activity of the complex it belongs to.
+      abundance and the enzymatic activity of the complex it belongs to. Graded
+      IN_VITRO rather than HUMAN_CLINICAL because both assays in this sentence are
+      ex vivo - transcript abundance in isolated white blood cells and COX activity
+      in EBV-transformed lymphoblastoid lines - and the human origin of the material
+      is not what evidence_source records.
   - reference: PMID:41044399
     reference_title: A novel homozygous COX6A1 variant causes axonal charcot-marie-tooth disease, developmental delays and mitochondrial dysfunction.
     supports: SUPPORT


aleixpuigb:
All three blockers fixed and three of the four suggestions taken. Pushed in 3dd0ae495.

**Blockers**

1. **`NCIT:C25294` → `NCIT:C88502`.** Your point that this matters more here than usual is the right one and I've written it into the node's description: the third knowledge gap is specifically about conduction velocities, so the diagnostic node it depends on should name the measurement.
2. **Four cardinal phenotypes added** — pes cavus, hyporeflexia, foot dorsiflexor weakness, steppage gait — all on the same already-verified `PMID:26556829` quote at `directness: INDIRECT`, matching the existing group-level phenotypes. One deliberate narrowing: bound to **Hyporeflexia** (`HP:0001265`) rather than Areflexia, because the source says *"diminished tendon reflexes"* and areflexia is the stronger claim. Your related point landed too: steppage gait and foot dorsiflexor weakness give the human anchor that `Gait Difficulty` was previously arguing for in prose, so the mouse-only phenotype now sits beside human-anchored siblings rather than standing alone.
3. **Treatment evidence replaced.** You're right that the death report is a real quote evidencing a *different* proposition than the entry asserts. Fetched `PMID:40014417` and cited it for both halves of the claim: that symptomatic management is the only option, and that it consists of rehabilitation and orthotics. Both `directness: INDIRECT`, `evidence_source: OTHER` — it is a group-level review that says nothing about CMTRID.

**Suggestions**

- **Genetic counseling** added. Agreed on the reasoning: for a recessive disease with no disease-modifying therapy it is the one actionable intervention, and both pedigrees were consanguineous or presumed related.
- **Onset captured.** `onset_category: CHILDHOOD` on the founding-family phenotype, `INFANTILE` on the two 2025-case phenotypes, each with a note. This was the best suggestion in the review — the nosology gap turns on exactly that axis and it was previously visible only in prose.
- **Neurotoxic-agent avoidance: declined**, with the reason recorded in `notes`. It is standard group-level CMT guidance with nothing CMTRID-specific behind it, and adding it would undercut the discipline the rest of the entry is built on — separating what is known about *this* disease from what is known about the group. Happy to add it if you'd rather the entry be clinically complete than epistemically narrow, but that trade-off should be a deliberate choice rather than a drift.
- **Onion bulbs**: recorded the condition under which it becomes citable — if the founding paper's full text describes nerve pathology, the annotation is quotable and the third gap largely resolves.

Also added a top-level `references:` block, for consistency with the sibling entries.

**One thing I caught and reverted while doing this.** Fetching `PMID:40014417` upgraded an *already-cached* abstract-only record to full text, adding ~2,000 lines to a shared `references_cache/` file. My snippets come from the abstract, which was present either way, so the upgrade bought nothing and would have put a large unrelated change to a repository-wide file in a curation PR. Reverted to the tracked version and force-pushed the amended commit (my own branch, no other contributor's work involved). Worth knowing that `just fetch-reference` on an existing abstract-only entry silently rewrites it.

Revalidated: `just validate` (36/36 snippets verified), `just validate-terms`, `just check-entity-refs`, `scripts/check_reference_titles.py --check`.

---

aleixpuigb:
Regraded to `IN_VITRO`. Pushed in 1d9645fd0.

Your read of the merits is right and I should have caught it when I wrote the second item. Both assays in that sentence are ex vivo — transcript abundance in isolated peripheral white blood cells, COX activity in EBV-transformed lymphoblastoid lines — and CLAUDE.md puts cultured cells and biochemical assays outside an organism under `IN_VITRO` whether or not the material came from a patient. I graded the sentence by where the cells came from rather than by what was done to them.

Also agreed on regrading over splitting: the prefix clause stops before its verb, so a split would leave a quote that is not propositional, which `check-snippet-length` exists to prevent.

One thing I added beyond the fix. The `explanation` now says *why* the item is `IN_VITRO` — that the human origin of the material is not what `evidence_source` records. Without that, "Demonstrates in patient material…" sitting under `IN_VITRO` reads like an error and invites the next curator to flip it back.

Worth noting what caught this: `check-snippet-grading` found a real defect that `just validate` cannot see, because the two items are individually well-formed and only the pair is wrong. That is the ungated whole-KB checks earning their keep.

Revalidated: `just validate` (36/36 snippets), `just check-snippet-grading` (no new findings), `just check-entity-refs`, `just check-duplicate-keys`, `just validate-history`. History record added.

---

PR #10075, opened 2026-08-29
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 3A (MONDO:0011103, GJB2)

### LINKED ISSUE #10067: Curate autosomal dominant nonsyndromic hearing loss 3A (MONDO:0011103)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 3A** ([MONDO:0011103](https://monarchinitiative.org/MONDO:0011103)) — DFNA3A, dominant nonsyndromic sensorineural hearing loss from heterozygous *GJB2* (connexin 26) variants that act by a dominant-negative or gain-of-function mechanism on the cochlear gap-junction network, rather than by the simple loss of function that causes recessive DFNB1.

Preflight: the KB already carries several GJB2-adjacent entries — `GJB2-GJB6_Digenic_Nonsyndromic_Hearing_Loss` and the dominant *syndromic* connexin-26 disorders `Keratoderma_Hereditarium_Mutilans` and `Clouston_Syndrome` — but nothing for the nonsyndromic dominant form, and no MONDO:0011103 hit on `origin/main`. The mechanistic contrast with the recessive form (dominant-negative assembly into hemichannels vs. biallelic loss of function) is the reason this looks like its own pathograph rather than a subtype, but that call belongs in the curation.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A.yaml`

aleixpuigb:
Adds DFNA3A, the dominant nonsyndromic *GJB2* hearing loss.

## Why a separate entry

The KB already carries `GJB2-GJB6_Digenic_Nonsyndromic_Hearing_Loss` and the dominant **syndromic** connexin 26 disorders (`Keratoderma_Hereditarium_Mutilans`, `Clouston_Syndrome`). This is the dominant **nonsyndromic** one, and it is not a severity variant of DFNB1 at the same locus — it is a different mechanism.

DFNB1 is a dosage disease: two loss-of-function alleles, too little connexin 26. DFNA3A is one allele that is still made, still oligomerises into a connexon, and degrades the channels its wild-type partner builds. The founding functional study includes the control that makes that claim specific rather than generic: a nonfunctional *recessive* allele co-expressed the same way left wild-type channels and their gating untouched.

## The part to actually review: the mechanism changed

The first draft of this entry modelled the cochlear step as endolymphatic potassium recycling and loss of the endocochlear potential — the textbook account of GJB2 deafness. **The deep-research report surfaced mouse data that contradicts it.**

In transgenic mice carrying the human dominant-negative R75W allele:

- the endocochlear potential is **normally sustained** and the stria vascularis is structurally intact — so failure to generate the EP is not the lesion;
- what fails is the **development** of the organ of Corti: no tunnel of Corti, no Nuel's space, no spaces around the outer hair cells, too few microtubules in the inner pillar cells;
- the outer hair cells themselves develop normally, keep their subsurface cisternae and prestin, and show normal electromotility in isolation — yet produce **no distortion-product otoacoustic emissions at any frequency**, because the compressed scaffold will not let them work in situ.

So the pathograph was rewritten. The primary lesion is a developmental supporting-cell scaffold failure; hair cell loss is downstream; the potassium node is relocated to **cortilymph** (the space around the hair cells) where the original authors put it; and the normal endocochlear potential is curated as a `REFUTE` evidence item against the endolymph account rather than quietly dropped. The correction and its reasoning are written into the entry's `notes` so a future reader can see what changed and why.

A second reframing came from the same report: a connexin 30 knock-in mouse in which **electrical coupling was normal while biochemical coupling (tracer transfer, IP₃/Ca²⁺ signalling) was reduced** — and the mice were deaf. Biochemical coupling can fail on its own, which is why the coupling node now names calcium homeostasis alongside gap-junction transport.

## Other things curated deliberately

- **The M34T retraction as a `REFUTE` item.** 101T>C was reported as a DFNA3 allele and later shown not to be sufficient to cause hearing loss. It is curated rather than omitted, so a curator reading the older paper does not re-import the retracted assignment.
- **AAV adenine base editing**, with its limits stated: preclinical, tested in a mouse the source paper calls *syndromic*, restores plaque morphology but no hearing outcome is reported — and the lesion is developmental, so a knowledge gap asks whether any post-natal window exists at all.
- **A literature disagreement left standing.** R75W is described as nonsyndromic by the group that found it and as syndromic-with-palmoplantar-keratoderma by the 2025 base-editing paper. Both are recorded, in the notes and in the animal model's description, because it bears on whether the R75W mouse models this entry or a neighbouring one.
- **`FAILS_TO_RECAPITULATE`** on the mouse's link to the potassium node, with limitations and evidence — the model's most useful result here is a negative one.

## Deep research

`openscientist`, committed alongside. Term quality in this report was poor and none of its ontology suggestions were used: **8 of 19 labelled terms were mislabelled**, including `MONDO:0011152` offered as the disease term (it is PHGDH deficiency), `CL:0002355` for a supporting-cell lineage (primitive red blood cell), `NCIT:C50071` for hearing aid (Mixer Device), and `NCIT:C16410` for gene therapy (Centrifugation). Its *literature* retrieval, by contrast, was excellent and is what drove the rewrite above.

## A curator error caught in-session

The `reference_title` for `PMID:25625422` was written from memory and did not match the cached title. Found by auditing every `reference_title` in the entry against `references_cache/`, and fixed. Flagging it because the repo's own checker uses a 0.85 similarity threshold and this is exactly the failure mode #9138 was opened about.

## Evidence honesty

There is no DFNA3A cohort, no audiological series, no natural history, no prevalence estimate — so no prevalence block, and phenotype frequencies are structural rather than measured. Everything mechanistic is model-system evidence from three systems that do not overlap, and coupling has never been measured at a DFNA3A genotype at native stoichiometry; that is filed as a `HUMAN_MODEL_MISMATCH` with a proposed experiment.

## Validation

- `just validate` — schema clean, 42/42 snippets verified against cached references

Closes #10067

---

### DIFF
--- a/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A/2026-08-29T175549Z-claude-code-e24ab1.yaml
+++ b/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A/2026-08-29T175549Z-claude-code-e24ab1.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A
+  path: kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A.yaml
+session:
+  id: 2026-08-29T175549Z-claude-code-e24ab1
+  timestamp: '2026-08-29T17:55:49Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10075
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  - phenotypes
+  summary: 'Review round 2: folded-scalar CI fix, treatment descriptions, onset claim
+    withdrawn'
+  details: |
+    Second review round. The blocking item was a CI failure I introduced: two folded scalars ended a line with 'keratitis-', so YAML folding rendered 'keratitis- ichthyosis-deafness syndrome' with a space inside the compound. Reflowed both so the compound stays on one line, rather than baselining them - the baseline exists to admit pre-existing findings, not new ones. Three non-blocking items also taken. Moved the NCIT-enum rationale out of the Hearing Aid Amplification description and into notes, since description renders as clinical content on the disorder page and that text is curation-process reasoning; kept the reasoning because it stops the next curator re-proposing NCIT:C183182 and NCIT:C157820, which are in the label cache but not in the treatmentactionterm enum cache. Rewrote the Cochlear Implantation description, which still read as the merged treatment covering both devices after the split. Withdrew onset_category CONGENITAL on the obligate hearing-loss phenotype: the entry's own note said no age-of-onset series exists and that dominant GJB2 pedigrees with later progressive loss are reported, so the enum asserted more than the evidence carries, and a consumer filtering on onset_category would never see the qualification. The onset block keeps the note explaining why the slot is unset. Revalidated: just validate 46/46 snippets, check-folded-hyphens, check-entity-refs, check-duplicate-keys.
--- a/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A.yaml
+++ b/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_3A.yaml
@@ -29,9 +29,9 @@ description: >-
   reported as dominant have been retracted on re-examination: the M34T (101T>C) allele
   was described as a DFNA3 allele and then shown not to be sufficient to cause hearing
   loss, and that refutation is curated as a REFUTE evidence item rather than omitted.
-  Second, dominant GJB2 variants more often cause *syndromic* disease - keratitis-
-  ichthyosis-deafness syndrome, Vohwinkel syndrome, palmoplantar keratoderma with
-  deafness - and this entry is the nonsyndromic form only. A dominant GJB2 pedigree
+  Second, dominant GJB2 variants more often cause *syndromic* disease:
+  keratitis-ichthyosis-deafness syndrome, Vohwinkel syndrome, and palmoplantar
+  keratoderma with deafness. This entry is the nonsyndromic form only. A dominant GJB2 pedigree
   with skin involvement is a different entry.
 disease_term:
   preferred_term: autosomal dominant nonsyndromic hearing loss 3A
@@ -568,13 +568,14 @@ phenotypes:
       id: HP:0008619
       label: Bilateral sensorineural hearing impairment
     onset:
-      onset_category: CONGENITAL
       notes: >-
-        Recorded as congenital because the mouse lesion is a failure of postnatal organ
-        of Corti development and the founding pedigree's deafness is profound and
-        prelingual. No age-of-onset series exists for DFNA3A, so this is the pattern
-        implied by the available descriptions rather than a measured distribution -
-        dominant GJB2 pedigrees with later, progressive loss are also reported.
+        onset_category is deliberately left unset. The founding pedigree's deafness is
+        profound and prelingual, and the mouse lesion is a failure of postnatal organ of
+        Corti development, which together suggest a congenital presentation - but no
+        age-of-onset series exists for DFNA3A, and dominant GJB2 pedigrees with later,
+        progressive loss are also reported. Asserting CONGENITAL would put a claim in a
+        filterable slot that the evidence does not carry, and a consumer filtering on
+        onset_category would never see this qualification.
   evidence:
   - reference: PMID:12064630
     reference_title: Functional analysis of a dominant mutation of human connexin26 associated with nonsyndromic deafness.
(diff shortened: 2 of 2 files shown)


aleixpuigb:
All four IMPORTANT items fixed and five of the seven suggestions taken. Pushed in 254a0d6b0.

**IMPORTANT**

1. **Audiometry → `NCIT:C38036`.** Your diagnosis of *why* it slipped through is the useful part and I've taken it: it passed `validate-terms` only because C25294 happens to be reachable from the enum root, which is exactly the "term precision over false matches" failure the terms skill warns about. A passing validator is not a correct binding.
2. **Both nodes downgraded to `PROVISIONAL`**, and your reasoning is now written into the node descriptions rather than just applied: three papers, but shared authorship, one transgenic line, one allele, at an expression ratio the model's own `limitations` field concedes is non-physiological, with no human histopathology anywhere. Convergent within a lab is not independent. `Dominant-Negative Degradation of Connexin 26 Channels` left at `ESTABLISHED` as you asked — Bruzzone/Petit and Chen/Altenberg are genuinely separate groups.
3. **Two `HUMAN_CLINICAL` items added to the `OBLIGATE` phenotype.** You put your finger on something I should have caught: the grading was correct, and that was the problem. The disease-defining feature no longer rests on an oocyte paper's background sentence.
4. **`references:` block added** with `tags: [GeneReviews]` on `PMID:20301449`. On the caveat you offered to drop — I think the tag is right and would rather keep it. It records that the chapter was consulted and which sections transferred; the entry already says in prose that the chapter covers the recessive form, and dropping the tag would make the consultation invisible rather than making it more accurate.

**Suggestions taken**

5. **E1 parahelix assertion removed** — and thank you for the warning about `PMID:21978595`, which is exactly the kind of thing that would otherwise have been cited on a second pass. The entry now names the specific alleles it curates and records that the general clustering rule was *not* adopted, with the off-topic flag as the reason.
6. Implant-outcome claim softened to a stated rationale plus an explicit "no DFNA3A-specific outcome study exists".
7. `GO:0006874` → `GO:0019722`. You're right that the quote is about *inter*cellular calcium wave propagation, not intracellular homeostasis.
8. Bilaterality and onset made structural: `HP:0008619` plus a `CONGENITAL` onset descriptor, with a note that no age-of-onset series exists and that later-onset dominant GJB2 pedigrees are reported.
11. `MODIFIER` genetic entry added for the trans loss-of-function allele.

**Two that could not be taken as proposed** — both now recorded in the entry rather than silently dropped:

- **Suggestion 9 does not validate.** `NCIT:C183182` (Hearing Aid) and `NCIT:C157820` (Cochlear Implant) are in `cache/ncit/terms.csv`, but they name *devices* and are not reachable from `NCIT:C25218`, so they fail the `TreatmentActionTerm` dynamic enum. I kept the split you suggested — it is a real distinction — but bound both to `NCIT:C49236`, with `therapeutic_modality: DEVICE` and the treatment names carrying the difference. The reason is written into the entry so the next curator does not retry it. Worth noting the general shape: label-cache presence is not enum membership, and the review's other term suggestions were all correct precisely because they were action terms.
- **Suggestion 10 is not expressible.** `DiseaseMappings` carries only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings` — there is no OMIM slot anywhere in the schema. That is a schema gap rather than a curation omission, and OMIM #601544 is the kind of identifier a mappings block exists to hold. I'll file it separately rather than work around it here.

Revalidated: `just validate`, `just validate-terms`, `just check-entity-refs`, `scripts/check_reference_titles.py --check`.

---

aleixpuigb:
Reflowed both lines; all three non-blocking items also taken. Pushed in bf25dac15.

**The blocker.** `keratitis-` ended a `>-` folded scalar at both sites, so folding produced `keratitis- ichthyosis-deafness syndrome`. Rewritten so the compound stays intact — the em-dash-style asides became a colon and a "namely", which reads better anyway. Not baselined: the baseline exists to grandfather findings that predate the check, and using it to admit a new one would defeat it.

Thank you for saying whose miss it was, but the defect was mine in the first commit either way. The useful lesson for me is that I ran the gates the fast curation loop names — `validate`, `count-verified-snippets`, `validate-terms` — and not the ungated whole-KB set that CI actually runs. I have run the full set on all four PRs in this batch since; it also caught the same defect in the XFE entry.

**Non-blocking 1 — enum rationale in `description`.** Agreed, and the distinction is the right one: `description` renders as clinical content on the disorder page and this is curation process. Moved to `notes:`, and I kept the reasoning for the reason you gave, adding the line that "presence in `cache/ncit/terms.csv` does not establish enum membership" — that is the specific thing I got wrong when I first considered those terms and the thing most likely to be got wrong again.

**Non-blocking 2 — leftover merged description.** Correct, and a real bug rather than a cosmetic one: the Cochlear Implantation entry was still describing both devices after the split, so the page would have shown the same text twice. Rewritten to be about implantation, with the intact-auditory-nerve point kept where it belongs.

**Non-blocking 3 — `CONGENITAL` onset.** You have persuaded me and I have unset it. Your framing is the argument: a consumer filtering on `onset_category` never reads the note, so the enum was making a claim the note simultaneously withdrew. The note said no age-of-onset series exists and that dominant GJB2 pedigrees with later progressive loss are reported; that is not enough to fill a filterable slot. `OnsetEnum` has no `UNKNOWN`, so the honest option is absence, and the `onset` block now carries a note saying the slot is deliberately unset and why.

That is a small instance of something more general I have hit repeatedly across these four entries: the KB has good machinery for recording *what* is claimed, and much thinner machinery for recording *how well it is known*. `mechanism_confidence` exists on pathophysiology nodes and `directness` on evidence items, but a phenotype's onset, frequency, and severity are asserted flat, with the epistemic qualification surviving only as prose the next consumer will not read. I am collecting these for a write-up at the end of this batch rather than filing piecemeal.

Revalidated: `just validate` (46/46 snippets), `just check-folded-hyphens`, `just validate-terms`, `just check-entity-refs`, `just check-duplicate-keys`. History record added.

---

PR #10066, opened 2026-08-29
Add Bryant-Li-Bhoj Neurodevelopmental Syndrome 1 entry (MONDO:0030606)

caufieldjh:
New entry for Bryant-Li-Bhoj neurodevelopmental syndrome 1 (BLBS1 / BRYLIB1, H3-3A), curated from `stubs/Bryant-Li-Bhoj_Neurodevelopmental_Syndrome_1.yaml`. The stub is deleted in this PR.

*(Body updated after three review rounds to describe the entry as it now stands.)*

## Why this is a separate entry

`Bryant-Li-Bhoj_Neurodevelopmental_Syndrome_2` (H3-3B) already exists and its `notes` say explicitly that "the H3-3A-specific counterpart (BLBS1, MONDO:0030606) is recorded as a differential diagnosis rather than folded in here." This PR is the other half of that decision, and reciprocates the differential.

The two genes encode an **identical protein**, so the split is by gene, not by protein product. The one firmly H3-3A-specific fact available is that every H3-3A variant with parental data has been a de novo missense change — where H3-3B has yielded a synonymous stop-gain, a stop-loss, and one inherited allele.

## Mechanism

```
De Novo H3-3A Variant Altering Histone H3.3 (MOLECULAR)
  -> Disrupted Nucleosome Assembly and Chaperone Interaction (MOLECULAR)
       -> Aberrant Local Histone Post-Translational Modification (MOLECULAR)
            -> Deregulated Cell Proliferation (CELLULAR)
            -> Neural Crest and Craniofacial Developmental Defect (TISSUE)
            -> Abnormal Brain Development (ORGANISM)
                 -> Progressive Postnatal Neuronal H3.3 Dependence (ORGANISM)
```

The last two were one node in the first draft and were split during review: the entry separates the congenital and degenerative phases everywhere else, including in `progression`, so the pathograph should too. The second carries `mechanism_confidence: PROVISIONAL`, because the mechanism it proposes — H3.3 progressively replacing the replication-dependent H3 variants until it dominates the adult neuronal H3 pool, so a defective H3.3 matters more the older a neuron gets — is an inference from substrate timing rather than a demonstrated cause of the human degeneration.

The sharpest point in the entry is the germline/somatic contrast. The same H3.3 residues are oncogenic when altered somatically (K27M, G34R/V), but germline variants deregulate histone modification **locally and in cis**, where somatic lysine substitutions do it globally in trans — and no BLBS patient in the founding cohort had a malignancy.

[...]

PR #10078, opened 2026-08-29
SPTCL: add demographics and a third HAVCR2 cohort

caufieldjh:
Follow-up to #10059, which merged before these changes landed. That PR was approved with three non-blocking suggestions; this takes them up as a fresh change rather than stacking on merged history.

## Demographics

`epidemiology:` now records the female predominance, quoting both cohorts as separate evidence items so the agreement between two unrelated populations stays visible rather than being averaged away:

- female-to-male 1.7, median age 46.5 years — European multicentre series (`PMID:30126736`)
- 69.8% women, median age 32 years (range 8–74) — Korean national cohort (`PMID:34535012`)

A second record covers age at diagnosis. The 46.5-vs-32-year gap between those cohorts is attributed to `HAVCR2` genotype rather than to geography, which the Korean cohort supports directly: "HAVCR2Y82C was associated with younger age (P = .001)". The p.Tyr82Cys allele is both younger-onset and far commoner in East Asian populations, so the two explanations are not independent — the entry says which one it is resting on.

## A third HAVCR2 cohort

`PMID:39288772` was cached by the original PR and never cited. It is now a third `case_fractions` entry — 23 of 29 germline carriers — alongside the Korean (51%) and Thai paediatric (86%) figures. Its `notes` record that the study's *own* Japanese arm is n=6, so the p.Tyr82Cys share it reports from that arm is fragile; the reviewer's point about relative cohort weight, written into the record rather than left for the next reader.

## Uncited cache files

Not deleted. A `notes` line now says what they are: output of the deep-research run's own citation-validation pass, committed because the report that cites them is committed, and not evidence for anything in the entry. That seemed more useful than a deletion that would have to be redone the next time a report is validated.

## Not acted on

`GO:0044546` (NLRP3 inflammasome assembly) still has no snippet behind it. The node already carries `mechanism_confidence: PROVISIONAL` and says in its description that the step is inferred from TIM-3 biology rather than demonstrated in SPTCL tissue. An unevidenced term that announces itself as unevidenced is more honest than dropping the mechanism.

## History

A new `EDIT` record covers this session. The `CREATE` record from #10059 is left byte-identical — history records are append-only, and that record accurately describes the session it came from.

## Validated

`just validate` — schema clean, terms clean, **63/63 snippets** verified. `just validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `validate-history` — all clean, re-run against current `main`.

---

### DIFF
--- a/kb/disorders/Subcutaneous_Panniculitis-like_T-cell_Lymphoma.yaml
+++ b/kb/disorders/Subcutaneous_Panniculitis-like_T-cell_Lymphoma.yaml
@@ -695,9 +695,9 @@ genetic:
     case_fraction_percent: 51.0
     cohort_size: 49
     notes: >-
-      Numerator is p.Tyr82Cys specifically, not any HAVCR2 mutation - unlike the
-      other two rows, which count any germline mutation. The denominators are
-      not interchangeable.
+      Numerator is p.Tyr82Cys specifically. That is comparable with the Thai
+      row, where every mutation detected was also p.Y82C, but not with the
+      integrated-reanalysis row, which counts any germline HAVCR2 mutation.
     evidence:
     - reference: PMID:34535012
       reference_title: "Genetic profiles of subcutaneous panniculitis-like T-cell lymphoma and clinicopathological impact of HAVCR2 mutations."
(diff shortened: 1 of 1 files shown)


caufieldjh:
Both blocking findings were right, and the second one was the more serious of the two — thank you for catching it.

**The cohort is not Japanese and not independent.** The study is six of the authors' own patients plus 30 reanalysed public exomes, and those public SPTCL exomes largely come from `PMID:30374066` and `PMID:30792187` — both already cited in the same `genetic:` block. So the row was doing two wrong things at once: putting an integrated reanalysis on the ancestry axis that the p.Tyr82Cys founder allele varies along, and implying it contributes 29 fresh patients when it mostly does not.

Fixed by saying what it is rather than by removing it:

- `population: Integrated series of six Japanese patients plus 30 reanalysed public exomes`
- `notes` now carries an explicit overlap warning — "This is not an independent cohort and should not be counted as one" — with the reason, and states why the row is kept anyway (a separate observation of the same axis, not 29 new patients)
- the study's own composition sentence is quoted as evidence, so the label is anchored rather than asserted

**Mixed denominators.** All three rows now state their numerator. The Korean 51% counts p.Tyr82Cys specifically; the other two count any germline mutation. That was invisible before and is exactly the kind of thing a query would silently get wrong.

**Genotype vs. geography.** You are right that the caveat was in the PR body and not the entry. The `Age at diagnosis` description now says the two explanations are *confounded and cannot be separated by these data* — the cohort with the younger median is the one with the founder allele, and neither series stratifies the other's variable — rather than attributing the difference to genotype as though that were settled.

**`mean_range`** now carries both medians. **The cache-provenance paragraph** moved from `notes` to `review_notes`; you are right that repository housekeeping should not render on the public disorder page.

`just validate`: 64/64 snippets verified. All other gates clean.

---

caufieldjh:
Took both. The first turned out to matter more than "non-blocking" suggests.

**The Thai row's numerator.** You were right to look twice, and I had it backwards. The two sets coincide in that cohort: every mutation detected was p.Y82C, and 77.3% homozygous + 9.1% heterozygous *is* the 86.4%. So the Thai and Korean rows are comparable on numerator after all, and the gap between 51% and 86.4% is about ascertainment — paediatric versus unselected — not about what was counted. My note saying "any germline HAVCR2 mutation" as though that distinguished it from the Korean row would have pointed a reader at exactly the wrong inference about a founder allele's prevalence. Now stated explicitly, including which variable the gap actually turns on.

**The age description** opened by attributing the difference to genotype and then conceded two lines later that genotype and ancestry are confounded. It now declines to choose from the first sentence: "Two explanations are available and this entry does not choose between them."

`just validate` 64/64; all gates clean.

Separately — I filed #10081 for the ClinGen checksum mismatch that came up on #10066, since as you noted it blocks `CGGV:` citation repo-wide rather than for one entry.

---

caufieldjh:
Fixed — and you were right that it landed in two of three places. The Korean row still said its numerator differed from "the other two rows, which count any germline mutation", which after the last commit was both arithmetically wrong (it is one row now) and directly contradicted by the Thai row two entries down. It now names which row it is comparable with and which it is not.

I also took the non-blocking suggestion, because it converts the load-bearing claim from assertion to evidence. Two items added on the Thai row:

- *"Fifth, only HAVCR2 p.Y82C variant was identified in our study population of exclusively Thai pediatric patients with SPTCL."* — this is the sentence that makes the two sets coincide, so it should be cited rather than inferred from the abstract's phrasing. Thank you for going to the full text for it.
- The gnomAD ethnicity-specific MAF sentence, carrying the ascertainment argument: Southeast Asian allele frequency an order of magnitude below East Asian, yet the Southeast Asian cohort has the higher carrier fraction. Ancestry does not explain the gap, so paediatric ascertainment is what is left — now readable from the entry instead of only from this thread.

`just validate`: 66/66 snippets verified. All other gates clean.

---

PR #10082, opened 2026-08-29
Add MPS X and attenuated MPS I entries

cmungall:
Two mucopolysaccharidosis gaps, found by diffing MONDO's descendants of `MONDO:0019249` against `kb/disorders`.

**MPS X** (`MONDO:0030524`, ARSK / glucuronate 2-sulfatase deficiency) had no entry at all. It is the most recently defined mucopolysaccharidosis, first described in humans in 2022, and the one the standard MPS diagnostic pathway systematically misses: routine urinary GAG screening is often normal, and every published patient carried a spondyloepiphyseal-dysplasia or bilateral Legg-Calvé-Perthes label before sequencing.

**The attenuated arm of MPS I** (Hurler-Scheie `MONDO:0011759`, Scheie `MONDO:0011760`) existed only as differential-diagnosis mentions on the Hurler entry and as three open stubs. Curated as a sibling of Hurler syndrome rather than folded into it: the two differ in age at presentation, in which organ systems carry the morbidity, and in which disease-modifying therapy is indicated.

## Full scope

The PR is wider than those two new files, so listing everything it touches:

| Area | Files |
|---|---|
| New entries | `Mucopolysaccharidosis_type_X.yaml`, `Attenuated_Mucopolysaccharidosis_Type_I.yaml` |
| Existing entries | `Hunter_syndrome.yaml` (severe/attenuated `has_subtypes`), `Maroteaux-Lamy_syndrome.yaml` (rapidly/slowly progressing `has_subtypes`), `Mucopolysaccharidosis_type_IX.yaml` (missing `classifications` block) |
| Grouping | `Mucopolysaccharidoses.yaml` — two new members, corrected MONDO consistency note, boundary notes recording why MPS-plus syndrome, multiple sulfatase deficiency, the mucolipidoses, ARSG deficiency and Morquio C are deliberately *not* members |
| Stubs | deleted the two the attenuated entry closes; re-typed the MPS I umbrella stub `GROUPING`/`LOW` with reasoning |
| History, caches, references cache | supporting records |

The four `has_subtypes`/`classifications` edits close six MONDO terms that were described in prose but not bound to the ontology, so a coverage audit read them as uncurated.

**The lump/split asymmetry is deliberate.** Attenuated MPS I is split out as its own entry; the MPS VI severity split is lumped as `has_subtypes`. The discriminator is whether the split changes the treatment decision. MPS I's does — HSCT versus ERT, on blood-brain-barrier grounds. MPS VI's does not: cognition is preserved in both forms because heparan sulfate is not an ARSB substrate, so the split is tempo only. Both entries state that reasoning in `notes`.

**Excluded phenotypes use `frequency: EXCLUDED` + `modifier: ABSENT`.** `hpoa_export` reads the former and emits NOT-qualified rows; I verified seven such rows come out with empty frequency columns. For attenuated MPS I only *developmental regression* is excluded, not intellectual disability — 16% of the cited attenuated cohort had cognitive impairment before ERT, so excluding cognition would have overclaimed.

## Not in this PR

The glycosylation landscape survey that was originally on this branch has been dropped. It tripped the `Check curation PR scope` CI step, which forbids a PR touching `kb/` from also touching `docs/` or `mkdocs.yml` without a `scope-override` label. It needs its own PR.

## Validation

Schema, term, entity-ref, duplicate-key, snippet-length, title-snippet, snippet-grading, reference-title, folded-hyphen and term-cache-integrity checks all pass. 122/122 evidence snippets verify against the reference cache.

Co-Authored-By: Claude Opus 5
Claude-Session: https://claude.ai/code/session_01KgqQQ4j9m8XGyxiQoPCP4n

### DIFF
--- a/history/disorders/Attenuated_Mucopolysaccharidosis_Type_I/2026-09-02T111344Z-claude-code-5890e2.yaml
+++ b/history/disorders/Attenuated_Mucopolysaccharidosis_Type_I/2026-09-02T111344Z-claude-code-5890e2.yaml
@@ -1,0 +1,24 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Attenuated_Mucopolysaccharidosis_Type_I
+  path: kb/disorders/Attenuated_Mucopolysaccharidosis_Type_I.yaml
+session:
+  id: 2026-09-02T111344Z-claude-code-5890e2
+  timestamp: '2026-09-02T11:13:44Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - biochemical
+  summary: Trim an explanation that outran its quote
+  details: |
+    Final review note on PR 10082. The explanation on the urinary GAG biochemical evidence claimed the reference documents baseline GAG elevation as well as the fall on therapy. The quoted sentence asserts only the fall; it presupposes a baseline in the phrase '63 percent of base-line values' without asserting that baseline was raised, and PMID:11172140 contains no sentence that does assert it, so there is no better quote available. Trimmed the explanation to what the quote supports and recorded explicitly that the elevation claim in the context field rests on the disorder description rather than on this quote. Same defect class as the truncated snippet that blocked the first review round: a claim outrunning the evidence attached to it, invisible to snippet verification because the quote itself is exact.
--- a/kb/disorders/Attenuated_Mucopolysaccharidosis_Type_I.yaml
+++ b/kb/disorders/Attenuated_Mucopolysaccharidosis_Type_I.yaml
@@ -964,8 +964,12 @@ biochemical:
     evidence_source: HUMAN_CLINICAL
     snippet: "Urinary glycosaminoglycan excretion decreased after 3 to 4 weeks of treatment; the mean reduction was 63 percent of base-line values."
     explanation: >-
-      Documents baseline urinary GAG elevation and the timing and magnitude of
-      its fall on therapy.
+      Documents the timing and magnitude of the fall in urinary GAG on therapy.
+      Deliberately does not claim baseline elevation: the quoted sentence
+      presupposes a baseline in "63 percent of base-line values" without
+      asserting it was raised, and this paper has no sentence that does. The
+      elevation stated in the context above rests on the disorder description
+      rather than on this quote.
   - reference: PMID:41353341
     reference_title: "Clinical outcomes of exclusive enzyme therapy (laronidase) in a cohort of patients with mucopolysaccharidosis type I."
     supports: SUPPORT


cmungall:
Both blocking findings are fixed, four of the five suggestions are taken, and the red CI check turned out to be the scope issue the review raised as a suggestion. Pushed in `24f71dea9`.

**The truncated snippet was the more serious of the two, and the review's diagnosis of the cause was exactly right.** I had cut the quote at "galactose" to dodge the U+2010 hyphens in "galactose‐6‐sulfatase". The fix is to quote the whole sentence including those characters, which verifies fine — the validator normalises whitespace but preserves the hyphens, so there was never a reason to truncate. Worth recording that this passed `count-verified-snippets`: a truncation that remains a valid substring is precisely what substring matching cannot catch, and reading is the only thing that would.

**GeneReviews is now tagged and mined.** Added the top-level `references:` block for `PMID:20301341` following the `Hurler_syndrome.yaml` pattern, with seven `Finding` objects each carrying an exact `supporting_text` quote. The genetic-counselling gap is closed in three places: recurrence risk as evidence on the `inheritance` block, and two new `diagnosis` entries — carrier and prenatal testing, and the severe-versus-attenuated determination that the chapter places first in management.

## What turned CI red

`test (3.13)` was failing on the `Check curation PR scope` step, not on a test. That step forbids a PR touching `kb/` from also touching anything outside the curation allowlist, and this PR carried `docs/reports/glycosylation-disease-landscape-2026-08-27.md` and a `mkdocs.yml` nav line.

The review reached the same conclusion independently, as a suggestion. Since CI makes it blocking, I dropped the survey commit rather than reaching for the `scope-override` label — the label would have removed the gate rather than passed it. The commit is preserved and needs its own PR.

## Suggestions taken

**Bundled cardiac/ocular node — split.** This one was worth doing rather than noting. Now that `Ocular connective-tissue involvement` is its own node, the `REFUTE` edge from the series reporting normal eye examinations is unambiguous about what it refutes, and the cardiac node no longer carries refuting evidence that never applied to it.

**Non-propositional snippets — replaced.** Both were method labels rather than findings. They now quote clauses that assert a result.

**Unstratified cohort caveat — mirrored** onto the treatment use of `PMID:41656624`, matching the phenotype use.

**`GO:0008484` breadth — documented** in the node description: GO has no term for glucuronate 2-O-sulfatase activity and no sibling sulfatase term fits, so the accurate parent beats a manufactured narrow match. A GO term request is the real fix.

## One correction

The "23 files" figure in `8ee370810` is right for what it was counting — KB YAML entries that schema validation covers. The 71 figure includes `kb/hypotheses/**/*.md`, which are deep-research artifacts, not schema-validated entries. Both numbers are correct for different scopes; the commit message could have said which.

## Not taken

The `frequency: VERY_FREQUENT` on carpal tunnel syndrome, derived from 27/33 in the unstratified cohort. The observation is fair, but every alternative is worse: dropping the frequency loses real information, and there is no attenuated-specific denominator published. I have added the caveat to the treatment use so both uses of that paper now carry it, and left the frequency with its existing caveat rather than inventing a number.

---

cmungall:
Took two of the three non-blocking notes in `c52ba9d14`. This drops the approval, which is worth it for the first one.

**The fragment quote is the same defect class that blocked the last round.** The severe-versus-attenuated snippet stopped one clause short of the claim its explanation makes — it named a determination without asserting it was essential. Now quotes "An essential component of management is the determination of whether the proband has severe or attenuated MPS I", at both the diagnosis evidence and the matching GeneReviews `supporting_text`. Leaving a known-truncated quote in place because the PR was already green would have been the wrong trade.

**`directness: INDIRECT`** added to the urinary GAG evidence on the diagnostic pathway entry. Not applied to the same quote in the `biochemical` section, where it is a direct pharmacodynamic result and `DIRECT` is correct — the same sentence is doing different work in the two places.

**PR body** was already rewritten before the re-review posted, so it now lists the `Hunter_syndrome`, `Maroteaux-Lamy_syndrome`, `Mucopolysaccharidosis_type_IX` and grouping edits and no longer disclaims the docs report.

## On the red check, for the record

`test (3.13)` was failing on the `Check curation PR scope` step, not a test. Confirmed by the check now passing on the same test suite with only the scope change.

One thing worth flagging for anyone debugging a similar failure: running the full pytest suite locally while rebasing the working tree produces 16 spurious `test_stubs.py` failures, because the stub files are collected as parameters and then deleted under the running process. They are stubs that main deleted in curation PRs, not real failures. CI, which runs against a clean checkout, is green.

## Validation on this head

`just validate-disorders` on both entries passes, 122/122 snippets verify, and `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` and `check-reference-titles` are all clean with no new baseline entries. The scope check is clean — the diff touches only `kb/`, `stubs/`, `history/`, `cache/` and `references_cache/`.

---

cmungall:
Not pushing for the residual `directness` note, and one correction.

**`directness: DIRECT` on the biochemical use at `:961` — leaving it absent.** I agree with the reviewer's own assessment that it isn't worth a push on its own, and there is a second reason: `CLAUDE.md` says to leave `directness` off unless it has actually been assessed, because absent means unassessed and that is the honest state of most of the KB. Adding `DIRECT` to record "I looked at this and it was fine" starts to erode that signal — if assessed-and-direct and never-assessed both carry a value, the field stops distinguishing them. The judgment is recorded in the commit message and the history record, which is where it belongs. If I push again for another reason I will fold it in.

**The PR body claim is stale.** Both reviews said the body no longer describes the branch. It was rewritten before the first re-review posted and now carries a full scope table covering the `Hunter_syndrome`, `Maroteaux-Lamy_syndrome`, `Mucopolysaccharidosis_type_IX` and grouping edits, plus a "Not in this PR" section explaining why the glycosylation survey was dropped. Flagging it so a human skimming the reviews does not go looking for a problem that is already fixed.

CI is still running on `c52ba9d14`; I will keep watching until it is green and merged.

---

cmungall:
`claude-review` failed on the merged head, but the failure is the review agent hanging, not a finding about the diff. `test (3.13)` was green on that same commit.

The job breakdown makes this unambiguous. Every setup step succeeded — checkout, token generation, and notably step 7, "Check the PR diff is enumerable by the review action". Then step 12, "Run Claude Code Review", ran for 21m47s and failed, and step 13 — the workflow's own guard, "Verify the review agent actually ran" — also failed. The posted checklist is still entirely unticked, including "Submit review verdict". Prior runs of this same job on this same branch finished in 4 to 6 minutes.

**I could not re-run it.** Both `run_workflow` on `claude-code-review.yml` and `rerun_failed_jobs` on the run return `403 Resource not accessible by integration`. Flagging that rather than leaving it to be inferred: I do not have the means to re-trigger this check directly.

**So I pushed the one real change I had pending.** `47492aa56` records `directness: DIRECT` on the biochemical use of the urinary GAG quote — the last outstanding suggestion from the second review, which I had explicitly deferred with "if I push again for another reason I will fold it in". This is that reason. It is a genuine content change, not an empty commit, and pushing it is the only mechanism available to me for getting a fresh review run.

For the record on why that grading differs between the two uses of the same sentence: on the diagnostic pathway it is `INDIRECT`, because a treatment-response result is being used to support a diagnostic screening measure. In the `biochemical` section the paper asserts the pharmacodynamic result directly, so `DIRECT`. It had been assessed and judged direct but left unmarked, which made an assessed item indistinguishable from an unassessed one.

## Validation on the new head

`just validate-disorders` passes on both entries, 122/122 snippets verify, and `check-entity-refs`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length` and `check-folded-hyphens` are clean with no new baseline entries. The scope check passes — 46 files by merge-base diff, all inside the curation allowlist.

I also independently verified the merge commit before this push: no conflict markers anywhere in `kb/`, `cache/`, `references_cache/`, `stubs/` or `history/`, both new entries intact, the MONDO term cache whole at 4,393 rows, and all nine members of the Mucopolysaccharidoses grouping still `SATISFIED` against the newer base.

If the review agent hangs again on this head, that starts to look like something about this PR's size rather than a one-off, and I would want a maintainer's view before pushing again to retry.

---

cmungall:
The re-review completed in about 7 minutes against 21m47s hanging on the previous head, so the stuck agent was a one-off rather than something about this PR's size. Withdrawing the concern I raised about that.

**Took the residual note** in `31b2d0078`, after checking it. The reviewer is right on both counts: the explanation claimed the reference documents baseline GAG elevation, the quoted sentence asserts only the fall, and `PMID:11172140` has no sentence that asserts elevation — so there is no better quote to reach for. Trimmed the explanation to what the quote supports, and recorded explicitly that the elevation stated in `context` rests on the disorder description rather than on this citation.

Fixed rather than deferred as pre-existing prose, because it is the same defect class as the truncated snippet that blocked the first round — a claim outrunning its evidence, invisible to snippet verification precisely because the quote itself is exact. That pattern has now accounted for three of the findings on this PR, which is worth noting for anyone reviewing similar curation: `count-verified-snippets` passing tells you the quote is real, not that it carries the claim attached to it.

**On the `check-stubs` advisory** the review flagged: agreed it should not be chased. `Mucopolysaccharidosis_Type_1.yaml` lists "Hurler syndrome" among its synonyms while curating `MONDO:0001586` rather than `MONDO:0011758`, which trips the duplicate heuristic. That is the state the stub's own `notes` describe and defend — the umbrella concept is deliberately unclaimed pending the grouping decision.

This was the last open item I am aware of. No further changes planned unless CI or a reviewer surfaces something.

---

PR #10083, opened 2026-08-29
Clear the retired supports values that #10003 invalidated on main

cmungall:
## main is failing schema validation

Since `9cb071dec` (the #7439 enum narrowing), 100 evidence items across 20 disorder entries carry `supports: PARTIAL` or `supports: WRONG_STATEMENT`, which the narrowed `EvidenceItemSupportEnum` no longer accepts:

```
[ERROR] [kb/disorders/Brachydactyly_Type_A2.yaml/0] 'PARTIAL' is not one of
        ['SUPPORT', 'REFUTE', 'NO_EVIDENCE'] in /pathophysiology/2/downstream/0/evidence/0/supports
```

This PR clears them. `kb/**/*.yaml` is back to zero retired enum values.

## These are not pre-existing defects

Worth being precise about, because it decides whose fix this is. Every one of those values was legal when it was written and legal when its PR merged. The same file, unchanged, against the two schema versions:

| schema | result |
|---|---|
| `1a524d7f2` (pre-merge) | `No issues found` |
| `9cb071dec` (post-merge) | `[ERROR] 'PARTIAL' is not one of [...]` |

So the narrowing did not expose a latent defect in those entries — it created the incompatibility, in files it deliberately did not touch. #9730 and the other curation PRs that landed alongside #10003 were each green against a base that did not yet carry the narrowing, and none of them could have detected the conflict. There is no defect in them to attribute, and nothing for their authors to fix.

The count grows with every curation merge: it was 9 values in 7 files an hour after #10003 landed, and 100 in 20 files by the time this PR was raised.

## What changed

**99 of the 100** are the mechanical `PARTIAL` → `SUPPORT` rewrite, applied by re-running `scripts/migrate_supports_enum.py --apply` — which is what that script was made re-runnable and idempotent for. The triage worklist carries forward rather than being overwritten, as designed.

**The hundredth** is the KB's last `WRONG_STATEMENT`, in `IgA_Pemphigus`, regraded by hand because the script deliberately never migrates that value automatically: the right direction depends on whether the entry's own text has since been corrected.

## The hand-regrade, and a correction

**An earlier revision of this PR graded that item `REFUTE`, and claimed the HPOA export had been checked and emitted no `NOT` row. That claim was false, and the reviewer caught it.**

The check behind it could only ever return a false negative: `hpoa_rows_for_disorder` returns a list of lists, and my filter tested `"HP:0100792" in r` against a list of dicts. Always false — so both arms returned zero rows and I read the tie as a clean result. Re-run correctly, `REFUTE` does add `qualifier='NOT'` on HP:0100792. That is the Marfan/pneumothorax failure again, in the one place I said I had checked for it.

The item is now graded **`NO_EVIDENCE`**.

`REFUTE` is defensible as direction — the cited sentence does contradict the claim — but on a `phenotypes[]` item `supports` is also a publication decision. It would have asserted downstream that IgA pemphigus does not feature acantholysis, on the strength of a single 2024 case report that this entry's own explanation argues is not supportable, against three sources describing acantholytic cells or blisters.

`NO_EVIDENCE` is in `DROP_SUPPORTS`, so the item, its quote and its reasoning stay visible on the phenotype while nothing is published. The explanation now records why, and names it as the weakest part of the item: the enum has no way to say *"this source contradicts the claim and we judge it mistaken"*, which is what `WRONG_STATEMENT` was doing here and what #7439 retired without a replacement for this case.

### Export delta, verified properly this time

```
MAIN  10 phenotype rows, 0 NOT
      HP:0100792 qualifier='' PMID:10886149
      HP:0100792 qualifier='' PMID:38751960   <- the denial, published as support
PR     9 phenotype rows, 0 NOT
      HP:0100792 qualifier='' PMID:10886149
```

One row removed, none added. Worth noting the row that goes: `main` was publishing the denying source as though it *supported* the phenotype, because `WRONG_STATEMENT` was in neither `SUPPORTS_TO_QUALIFIER` nor `DROP_SUPPORTS` and fell through to a positive row. So this was wrong on `main` too, in the other direction.

The `history/` record repeated the false claim and is corrected in place rather than superseded — it has never merged, so it is not yet part of the append-only record. It now states what was checked, what the broken check did, and who caught it.

## Why this carries the `scope-override` label

The curation-scope guard fires on one file:

```
docs/reports/data/partial-evidence-directness-worklist-2026-08-29.tsv
```

I considered dropping it to stay inside the curation allowlist, but that would permanently lose these 99 items from the directness triage queue. The worklist is generated from items that *currently* hold `supports: PARTIAL`; once this PR merges they are `SUPPORT`, so no future run of the script can ever re-collect them. The worklist update has to travel with the sweep or those items become invisible to the follow-up pass.

That is the intentional cross-scope case the label is for. Nothing else outside `kb/` and `history/` is touched.

## Validation

| Check | Result |
|---|---|
| `linkml-validate -C Disease` on all 21 changed entries | No issues found |
| Anchored retired enum values in `kb/**/*.yaml` | 0 |
| HPOA export delta vs `main` | 1 row removed, 0 added, 0 `NOT` rows |
| `check-duplicate-keys` | OK, 4,249 files |
| `check-entity-refs` | OK, 2,764 files |
| `check-folded-hyphens` / `check-snippet-grading` / `check-title-snippets` / `check-empty-snippets` | no new findings |
| `test_data.py` + `test_hpoa_export` + `test_empty_snippets` (evidence/schema/supports/directness/hpoa/history/snippet) | 7,928 passed |

No baseline was updated.

## Note for whoever lands this

This will recur until the enum change and the curation stream stop overlapping. Any curation PR opened before `9cb071dec` and merged after it can carry the same breakage, and neither side's CI can see it. Re-running `scripts/migrate_supports_enum.py --apply` is the remedy and is safe to repeat — worth checking `main` again immediately before merging, in case more has arrived.

https://claude.ai/code/session_01VUmMdcPTa7ZqwRV1NKwMbz

### DIFF
--- a/kb/disorders/IgA_Pemphigus.yaml
+++ b/kb/disorders/IgA_Pemphigus.yaml
@@ -1137,9 +1137,8 @@ discussions:
     to bind is a marker of the disease rather than its mechanism.
 
     This is why the adhesion node is graded PROVISIONAL while the neutrophil node
-    is ESTABLISHED, and why the acantholysis phenotype keeps the contradicting
-    source as a NO_EVIDENCE item rather than dropping it
-    that disagrees.
+    is ESTABLISHED, and why the acantholysis phenotype keeps the source that
+    disagrees as a NO_EVIDENCE item rather than dropping it.
   proposed_experiments:
   - experiment_id: igap_dsc1_passive_transfer
     name: Passive transfer of affinity-purified anti-Dsc1 IgA


cmungall:
You're right, and the failure was in my check rather than just my wording. Fixed in `1c89d606a`.

I reproduced your result before acting on it, and the reason I missed it is worth stating plainly: `hpoa_rows_for_disorder` returns a **list of lists**, and my filter tested `"HP:0100792" in r` against a list of dicts. That is always false, so both arms of my comparison returned zero rows and I read the tie as a clean result. The check could only ever have returned a false negative. I then asserted it in the commit message, the PR body and the history record.

That is the Marfan/pneumothorax failure again, in the one place I said I had checked for it.

## What I changed

Took your option 1 — `supports: NO_EVIDENCE`.

I went back and forth with option 2 (relocating the source into `discussions`). The discussion that already `attaches_to: phenotypes#Acantholysis` asks whether anti-desmocollin IgA is pathogenic or a bystander marker — that is a mechanism question, not a presence question, so the contradicting source would have muddied a thread it does not belong to. Creating a new discussion for it would widen the PR further than the fix needs.

So the item stays on the phenotype, graded `NO_EVIDENCE`, and the explanation now says why in the item itself — including that this is the weakest part of it. Your framing of the problem is the part I wanted to preserve: `REFUTE` does not misread the evidence, it is that on a `phenotypes[]` item the value is *also* a publication decision, and here it publishes something the entry disbelieves.

The enum has no way to say "this source contradicts the claim and we judge it mistaken". That is what `WRONG_STATEMENT` was doing here, and #7439 retired it without a replacement for this case — worth recording as a real gap the narrowing left, distinct from the "our own text was wrong" case the migration script's docstring anticipated.

## One thing your review didn't mention, which the correct check surfaced

The row that disappears is a row `main` should not have been emitting:

```
MAIN  10 phenotype rows, 0 NOT
      HP:0100792 qualifier='' PMID:10886149
      HP:0100792 qualifier='' PMID:38751960   <- the denial, published as support
PR     9 phenotype rows, 0 NOT
      HP:0100792 qualifier='' PMID:10886149
```

`WRONG_STATEMENT` was in neither `SUPPORTS_TO_QUALIFIER` nor `DROP_SUPPORTS`, so it fell through to a positive row — `main` has been publishing that 2024 denial as though it *supported* acantholysis in IgA pemphigus. Wrong in the other direction, and pre-dating this PR. The delta is now one row removed and none added.

## The claim in the history record

Corrected in place rather than superseded. That record has never merged, so it is not yet part of the append-only history — correcting it before it lands is the difference between a record that was always accurate and one that permanently carries a false verification claim. It now states what was checked, what the broken check actually did, and that the review caught it.

## Re-verified

| Check | Result |
|---|---|
| HPOA export delta vs `main` | 1 row removed, 0 added, **0 `NOT` rows** |
| `linkml-validate -C Disease` | No issues found |
| Retired enum values in `kb/**/*.yaml` | 0 |
| six offline gates (`duplicate-keys`, `entity-refs`, `folded-hyphens`, `snippet-grading`, `title-snippets`, `empty-snippets`) | all clean |
| evidence / HPOA / history / snippet test selections | 7,928 passed |

Agreed on the non-blocking point: the 91 `explanation` fields still reading "Graded PARTIAL because…" are #10003 backlog, not this PR's, and the worklist TSV is the right vehicle. I measured that KB-wide earlier — 3,644 items across 750 files carry the uppercase token, and a further 1,116 use lowercase "partial/partially" as ordinary English and should never be touched. About 96 of them now actively contradict the `SUPPORT` value beside them, which is the tranche worth doing first.

---

cmungall:
Dangling clause fixed in `56a65d4fa`. You were right about the cause — my replacement covered two folded lines where the sentence spanned three, so `that disagrees.` was orphaned. Reworded so the relative clause attaches to its noun:

> …keeps **the source that disagrees** as a NO_EVIDENCE item rather than dropping it.

Re-verified: `linkml-validate` clean, `check-folded-hyphens` clean, entity refs resolve, HPOA export unchanged by this commit (9 rows vs main's 10, zero `NOT`).

Sorry for the extra round — that push dismisses your approval again under `dismiss_stale_reviews`.

On the follow-up: agreed that the missing "this source contradicts the claim and we judge it mistaken" case deserves an issue against #7439 rather than anything here. I don't open issues unprompted, so I've flagged it to @cmungall along with the other items from this work. Worth noting for whoever writes it that the case is rarer than the retired value's usage suggested — `WRONG_STATEMENT` had exactly one use KB-wide, and it turned out to mean two different things across its lifetime: "our own text was wrong" in `Huntington_Disease` (fixed by correcting the text) and "the source is wrong" here. Only the second is unrepresentable.

---

Issue #8836, opened 2026-08-18
Orphadata MANIFEST.yaml pin is stale; bump + full ORPHA_*.md cache regen needed


cmungall:
## Background

While building `references_cache/ORPHA_988.md` for #8707, `just refresh-orphadata` failed:

```
RuntimeError: checksum mismatch for en_product1.xml after download:
got df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb,
expected fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

`data/orphadata/MANIFEST.yaml` pins a 2025-12-09 snapshot, but the live Orphadata bulk XML files (`en_product1.xml`, `en_product4.xml`, `en_product6.xml`, `en_product9_prev.xml`, `en_product9_ages.xml`) have all drifted from those sha256s as of 2026-08-17/18. `#8707` worked around this by downloading the live files directly (bypassing the checksum gate) to build the two specific cache files it needed, rather than fixing the pin — that fix has a much bigger blast radius than one PR should carry, per review on #8707.

## What's needed

1. Re-download all five `data/orphadata/*.xml` bulk files, `sha256sum` them, and update `sha256` + `snapshot_date` in `MANIFEST.yaml` (see the procedure already documented at the top of that file).
2. Run `just structured-rebuild-orphanet` (full, no `--id` filter) to regenerate **all** ~341 committed `references_cache/ORPHA_*.md` files against the new snapshot.
3. Review the resulting diff carefully before committing:
   - Content drift between Dec 2025 and the new snapshot may shift wording, phenotype frequencies, prevalence classes, etc. in ways that could break existing curator-quoted `evidence.snippet` values across the KB (schema/reference validation will catch breakage, but the diff itself should still be read, not just the CI result).
   - `#8707` also extended `src/dismech/structured_sources/orphanet.py` to render two new sections that don't yet exist in any of the 341 currently-committed cache files: a `**Status:**` line (from `DisorderFlagList` labels — e.g. "Deprecated entity", "Inactive") and a `## Related disorders` table (from `DisorderDisorderAssociationList` — "Moved to" / "Referred to" relations). The full regen in this issue is what will make those sections appear across the corpus for the first time; expect a nontrivial diff even for records whose actual disorder content is unchanged from the pinned snapshot.
[...]

PR #10095, opened 2026-08-29
curate: Add Optic Atrophy 3 (MONDO:0008133, OPA3) as the dominant ADOAC entity

### LINKED ISSUE #10093: Curate optic atrophy 3 (MONDO:0008133)
Curate a dismech entry for **optic atrophy 3** ([MONDO:0008133](https://monarchinitiative.org/MONDO:0008133)) — autosomal *dominant* optic atrophy with cataract (ADOAC/OPA3), caused by heterozygous *OPA3* variants acting on the mitochondrial inner membrane, with early-onset optic atrophy, cataract and variable extrapyramidal and myopathic features.

The main hazard here is entity confusion rather than granularity. *OPA3* causes two distinct diseases: this dominant optic-atrophy-plus-cataract disorder, and the recessive Costeff syndrome (3-methylglutaconic aciduria type III), which is a different MONDO term with a different inheritance mode and a urinary organic-acid phenotype the dominant form does not have. Evidence has to be sourced to the dominant form specifically, and the entry should say where the line falls.

Preflight: MONDO:0008133 is a leaf term with one causal gene (`hgnc:8142` OPA3) and no existing `kb/` entry or open PR. `kb/disorders/3-Methylglutaconic_Aciduria_Type_I.yaml` is *AUH*, a different gene, so it is not the recessive counterpart.

Stub: `stubs/Optic_Atrophy_3.yaml`

matentzn:
New entry for **optic atrophy 3** — autosomal dominant optic atrophy and cataract (ADOAC), the heterozygous *OPA3* disease. Closes #10093.

## The thing to check first

*OPA3* causes two diseases, and almost everything written about "OPA3" covers both at once. Heterozygous, mostly missense variants cause this dominant optic-atrophy-plus-cataract disorder. Biallelic, mostly loss-of-function variants cause recessive Costeff syndrome (3-methylglutaconic aciduria type III), which has chorea, spastic paraparesis and a urinary organic-acid marker that this disease does not.

Every claim in the entry is traced to a heterozygous pedigree. Costeff content appears only in `differential_diagnoses`, labelled as such. If you review one thing, review that line.

## What is in it

Four pathophysiology nodes — heterozygous missense variant → reduced OPA3 steady-state levels with a fragmented mitochondrial network → retinal ganglion cell degeneration, plus a separate lens node. The chain stops at network morphology on purpose: OPA3 has no known homolog and no assigned biochemical activity, so naming a step would be inventing one.

Ten phenotypes, the *OPA3* gene block, two prevalence records, diagnosis, four treatments, one animal model, four differential diagnoses, three open discussions.

## Three places I chose the weaker claim

**The animal model is entered as a failure.** The Opa3 L122P ENU mouse is `FAILS_TO_RECAPITULATE` against retinal ganglion cell degeneration. The heterozygote — the genotype that matches human dominant disease — is unaffected; every retinal and optic-nerve finding in that mouse is a homozygote finding and belongs to Costeff syndrome. So there is currently no in vivo model of this entity, which is one of the discussions.

**One evidence item is `REFUTE`.** A 2025 family with a segregating *OPA3* missense allele has congenital cataracts, hearing loss and neuropathy but no optic atrophy. That refutes optic atrophy as an obligate feature, so the phenotype is `VERY_FREQUENT` rather than obligate.

**The gain-of-function proposal is annotated `INCREASED`, not `GAIN_OF_FUNCTION`.** PMID:20372962 proposes a gain-of-function mechanism, but on transient overexpression of one allele in a cell line. Per the CLAUDE.md rule that trades ontology grounding for expressivity only on a qualitative claim, the process annotation stays quantitative and the evidence item carries `directness: INDIRECT` with the reason stated in place.

Several other items are `SUPPORT` + `directness: INDIRECT` where the source describes dominant optic atrophy as a class rather than *OPA3* carriers — mainly the Orphanet review, used for prevalence, fundoscopic signs, colour vision and management advice.

## A reference I could not quote

PMID:15342707 (Reynier 2004) established *OPA3* as an ADOAC locus, but no abstract or full text is retrievable into `references_cache` — the cached record is metadata only. It is in the top-level `references:` block as the founding citation and **no evidence item quotes it**; the claims it would have carried come from PMID:15924081 and the later series instead. Said in the entry's `genetic` notes and trailing `notes` so it does not read as an oversight.

One Perplexity run, committed alongside. Used for orientation only.

It is titled "OPA3-Associated Optic Atrophy and Organic Aciduria" and interleaves the two diseases throughout. Its frontmatter reports 18 citations but reference validation checked 4 and passed them all — because the report contains no PMID and no DOI anywhere in its body; it cites web pages. A clean panel there means "the four parseable identifiers resolved", not "well sourced". It also offered `HP:0001103` as *optic disc pallor*, which HPO calls *Abnormal macular morphology*; the entry uses `HP:0000543`.

It did earn its place, in a way the prose does not advertise: resolving its citations pulled two papers into the cache that my own PubMed searches missed — PMID:20372962 (outer-membrane localisation, bidirectional dose–morphology effect) and PMID:21613372 (lens expression). Both are cited, and the first materially improved the mechanism node.

## Validation

Schema validation passes. **48/48 evidence snippets verified** against cached primary references. Term validation, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-environmental-evidence` all pass. `just validate-disorders` passes.

---

### DIFF
--- a/history/disorders/Optic_Atrophy_3/2026-09-03T030339Z-claude-code-5cc796.yaml
+++ b/history/disorders/Optic_Atrophy_3/2026-09-03T030339Z-claude-code-5cc796.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Optic_Atrophy_3
+  path: kb/disorders/Optic_Atrophy_3.yaml
+session:
+  id: 2026-09-03T030339Z-claude-code-5cc796
+  timestamp: '2026-09-03T03:03:39Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10095
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - phenotypes
+  - evidence
+  summary: Correct PMID:31119193 cohort denominators
+  details: |
+    Review round 3. The ADOAC+ series in PMID:31119193 is twelve individuals - one family of seven, one sporadic case, and a second family of four - and the entry prose said eleven twice. Gastrointestinal dysmotility was described as eight of eleven when the source gives eight in the first group plus three in the second family, so eleven of twelve; the band moved OCCASIONAL to VERY_FREQUENT to match, with a note recording that the probands were referred for peripheral neuropathy so 11/12 is an upper bound from an ascertained cohort rather than a population frequency. Autonomic dysfunction corrected to three of twelve. Cataract prose separated the seven-of-eight sub-cohort from the nine-of-twelve series figure. Also folded in two deferred items: the second toxin-avoidance evidence item regraded HUMAN_CLINICAL to OTHER for consistency with its sibling, and the uncached ORPHA:67036 claim hedged. Snippets unchanged and still 50/50; all gates pass.
--- a/kb/disorders/Optic_Atrophy_3.yaml
+++ b/kb/disorders/Optic_Atrophy_3.yaml
@@ -475,8 +475,9 @@ phenotypes:
     presenting problem, with the optic neuropathy recognised only when vision
     fails to improve after lens surgery.
 
-    Graded VERY_FREQUENT: cataract is in the name of the entity, seven of eight
-    individuals in the ADOAC+ series had early-onset cataracts, and the 2025
+    Graded VERY_FREQUENT: cataract is in the name of the entity, seven of the
+    eight individuals in the first family and sporadic case had early-onset
+    cataracts - nine of twelve across the whole ADOAC+ series - and the 2025
     family had congenital cataracts. It is not obligate - one carrier of the
     recurrent p.(Gln105Glu) allele had no lens opacities at age 38 - which is
     why the band is not written as universal.
@@ -507,7 +508,8 @@ phenotypes:
       one sporadic case were early-onset cataracts (n = 7), symptoms of
       gastrointestinal dysmotility (n = 8), and possible/confirmed PN (n = 7).
     explanation: >-
-      Counts early-onset cataract in seven of eight individuals in the ADOAC+
+      Counts early-onset cataract in seven of the eight individuals in the first
+      family and sporadic case of the ADOAC+
       series.
 - category: Ophthalmological
   name: Reduced Visual Acuity
@@ -640,15 +642,20 @@ phenotypes:
 - category: Gastrointestinal
   name: Gastrointestinal Dysmotility
   description: >-
-    Reported in eight of the eleven individuals in the ADOAC+ series, which makes
-    it the most frequent extraocular feature there, and one that would not be
-    looked for without knowing the diagnosis.
+    Reported in eleven of the twelve individuals in the ADOAC+ series - all eight
+    of the first family plus the sporadic case, and three of the four in the
+    second family - which makes it the most frequent extraocular feature there,
+    and one that would not be looked for without knowing the diagnosis. The band
+    is graded against that series and not against ADOAC at large: the probands
+    were referred for evaluation of a peripheral neuropathy, so this cohort is
+    ascertained for the neurological end of the spectrum and 11/12 is an upper
+    bound rather than a population frequency. No unbiased denominator exists.
   phenotype_term:
     preferred_term: Gastrointestinal dysmotility
     term:
       id: HP:0002579
       label: Gastrointestinal dysmotility
-  frequency: OCCASIONAL
+  frequency: VERY_FREQUENT
   evidence:
   - reference: PMID:31119193
     reference_title: "Autosomal dominant optic atrophy and cataract \"plus\" phenotype including axonal neuropathy."
@@ -663,7 +670,7 @@ phenotypes:
 - category: Neurological
   name: Autonomic Dysfunction
   description: >-
-    Symptoms or signs of autonomic dysfunction in three of the eleven
+    Symptoms or signs of autonomic dysfunction in three of the twelve
     individuals in the ADOAC+ series - the same count, from the same sentence,
     as the hearing loss recorded above. Plausibly the autonomic limb of the same
     axonal neuropathy, though the source does not make that link.
@@ -980,13 +987,15 @@ treatments:
     reference_title: "Costeff Syndrome."
     supports: SUPPORT
     directness: INDIRECT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: >-
       Agents/circumstances to avoid: Use of tobacco, alcohol, and medications
       known to impair mitochondrial function.
     explanation: >-
       The same advice in the GeneReviews chapter for the recessive OPA3 disease.
-      Graded INDIRECT because it is written for the other OPA3 disease; it is
+      Graded OTHER for the same reason as its sibling above - it reports expert
+      consensus practice rather than a clinical observation - and INDIRECT
+      because it is written for the other OPA3 disease; it is
       cited to show the advice is gene-level, not to import recessive-disease
       content.
 - name: Genetic Counselling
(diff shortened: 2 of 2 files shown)


matentzn:
All six findings addressed and pushed. Nothing disputed. Snippets now **50/50 verified**, up from 48.

**IMPORTANT 1 — phenotypes omitted from a sentence I already quote.** Correct, and your framing of it is the right one: the defect wasn't the omission, it was applying the inclusion threshold inconsistently *within one sentence of one source*. Added `Autonomic Dysfunction` (HP:0012332, OCCASIONAL) and `Recurrent Pancreatitis` (HP:0001733, `temporality: RECURRENT`, VERY_RARE), both quoting that sentence. The pancreatitis description says out loud why an n=1 finding is in: to keep the threshold consistent.

**IMPORTANT 2 — extrapyramidal signs.** I tried the structured fix before falling back to a scoping note. **ORPHA:67036** is the source that carries tremor and extrapyramidal rigidity as coded phenotypes, and this repo's ORPHA structured source would make them quotable — but it is not among the 347 ORPHA records cached here, and **`just refresh-orphadata` currently fails on a checksum mismatch**: `en_product1.xml` now downloads as `df8d562a…` against `fb2fbe8c…` pinned in `data/orphadata/MANIFEST.yaml`. Repinning that is not a curation PR's business, so I stopped.

What went in instead states a position rather than an absence: the feature is real, this entry does not dispute it, and it is missing for want of a quotable source rather than for want of evidence. The note names all three unusable sources and why — PMID:15924081's body offers only "neurological symptoms" while the specific claim lives in its title (`check-title-snippets` correctly bars that); PMID:22797356 resolves but has no abstract or full text, and I've committed its cache file so that is verifiable rather than asserted; ORPHA:67036 is uncached and currently unfetchable. It names the fix and what a future curator should add.

You also caught a consequence I hadn't seen: the description contrasted this disease against Costeff chorea/spastic paraparesis in a way that read as an implicit claim of no movement features, contradicted by the entry's own cited title. Rewritten to say the two syndromes' *defining* features differ, and to point at the note.

**Suggestion 3 — orphan nodes.** Two `downstream` edges added from `Retinal Ganglion Cell Degeneration`, each described from material already in the file. The graph is one connected component now.

**Suggestion 4 — cataract frequency.** Raised to `VERY_FREQUENT`. You're right on the numbers. The description now states the band *and* the one case against it (the age-38 carrier with no lens opacities), so the choice is legible instead of implicit.

**Suggestion 5 — evidence_source.** Changed to `OTHER`. It reports consensus practice, not a clinical observation. `directness: INDIRECT` retained; the explanation now gives both reasons separately.

**Suggestion 6 — duplicate treatment term.** Rebound to `NCIT:C15843` (Preventive Intervention), which is reachable from `NCIT:C25218` and passes the `treatmentterm` enum check.

One thing worth flagging beyond this PR: the `refresh-orphadata` checksum drift affects any curation that wants a not-yet-cached ORPHA record, not just this entry. Happy to open an issue for it if there isn't one already.

---

matentzn:
Correcting my own offer above: I checked before filing, and the `refresh-orphadata` checksum failure is already open **eight times** — #7199, #7234, #7522, #7539, #7622, #8836, #9687, #9897. So I am not filing a ninth, and the note in the entry stands as the record for this disease.

The duplication is probably worth a maintainer's attention on its own. Every one of those was filed by an agent hitting the same wall mid-curation, which suggests the failure is common enough and the pre-filing duplicate check weak enough that it will keep happening until either the pin is bumped or `refresh-orphadata` fails with a message pointing at the existing issue.

---

matentzn:
Thanks for the approval, and for re-deriving the three checkable claims in that scoping paragraph rather than taking them on trust — including downloading the ORPHA XML to confirm 67036 is the right record. That is the check I couldn't complete.

**I'm not pushing again**, because the PR is approved and `dismiss_stale_reviews` is on, so a commit would drop it and cost a full review cycle for two cosmetic items. Recording both here so they're not lost:

1. **The second toxin-avoidance evidence item is still `HUMAN_CLINICAL`** while its sibling was regraded to `OTHER`. You're right that the same argument applies to both — management advice is expert guidance, not a clinical observation. That is an inconsistency I introduced in the fix commit.
2. **The claim that ORPHA:67036 "does list tremor and extrapyramidal rigidity" should be hedged.** It came from the deep-research report's Orphanet citation, not from a cached record I read — which makes it the one unsourced assertion in a paragraph otherwise written to be checkable. Your download confirms the record is the right one but not its phenotype content.

Both are one-line edits. Happy to make them if a maintainer would rather have them than the standing approval — just say so and I'll push.

On the Orphadata manifest drift: agreed it deserves its own issue, and it already has **eight** — #7199, #7234, #7522, #7539, #7622, #8836, #9687, #9897, all filed by agents hitting the same wall mid-curation. I checked before filing and did not add a ninth. The duplication itself may be worth a maintainer's attention: it will keep recurring until either the pin is bumped or `refresh-orphadata` fails with a message pointing at the existing issue.

---

matentzn:
This PR had gone un-mergeable against `main`, so I rebased it. It is conflict-free again.

Both conflicts were in derived caches — `cache/mondo/terms.csv` and `cache/enums/diseaseterm_f618932bb23f.csv` — which every curation PR appends to, so any two of them that sit open across the same window will collide there.

I resolved them by taking main's version of each file and re-running `just validate` to regenerate the rows, rather than hand-merging the conflict markers. CLAUDE.md is explicit that cache rows are never hand-written, appended or reordered, and a hand-merge is exactly how a stale or mistyped row gets in. The regeneration produced precisely the two rows this entry needs:

```
MONDO:0008133
MONDO:0008133,optic atrophy 3,2026-09-03T02:41:23.345557
```

Re-validated on the new head: **50/50 snippets verified**, `just validate` passes, and `check-entity-refs`, `check-duplicate-keys`, `check-term-cache-integrity`, `check-folded-hyphens`, `check-snippet-grading` and `check-title-snippets` are all clean with **no baseline edits** (the baselines moved on their own as main advanced).

The rebase drops the approval, which I would rather not have spent — but the PR was un-mergeable, and that had to be fixed either way. No content changed: the diff against the previous head is the two cache rows and nothing else.

The two cosmetic leftovers I flagged when this was first approved are still unaddressed, deliberately, and still available if a maintainer wants them.

---

matentzn:
Both findings addressed, plus the two carry-overs. Head is now `da918dfa4`.

**CRITICAL 1 — already fixed before your review landed.** You reviewed `0f0d302e5`; the fix is in `36f9d7fe9`, pushed a few minutes earlier. All 8 targets now use bare names, and `build_causal_graph` reports `orphan_targets` empty with 10 resolving edges. I verified the graph itself rather than just the gate.

Your two side observations were both right and both worth more than the fix. The earlier "graph is one connected component now" claim was indeed false of the committed state — the two edges I added for Optic Disc Pallor and Colour Vision Defect used the same broken grammar, so that round's fix connected nothing while reporting success. And this was not confined to this entry: **all five PRs in this run had it, 44 targets in total.** The whole-KB run showed only this file because the other four are still based on the older `main` that predates the gate. They are fixed and pushed too (#10097, #10098, #10100, #10101).

**IMPORTANT 2 — confirmed against the cache and corrected.** The series is twelve: family of seven, one sporadic case, family of four. I had written eleven twice.

- **Dysmotility** was "eight of the eleven". The source gives n=8 in the first family plus sporadic case — which is all eight of them — and three more in the second family: **eleven of twelve**. You were right that `OCCASIONAL` cannot survive that. I raised it to `VERY_FREQUENT` and took your suggestion about stating the reasoning: the description now records that the probands were referred for evaluation of a peripheral neuropathy, so 11/12 is an upper bound from a cohort ascertained for the neurological end of the spectrum rather than a population frequency, and no unbiased denominator exists.
- **Autonomic dysfunction** → three of twelve.
- **Cataract** prose now separates the seven-of-eight sub-cohort from the nine-of-twelve series figure.

**Both carry-overs folded in**, since a push was needed anyway: the second toxin-avoidance item is regraded `HUMAN_CLINICAL` → `OTHER` to match the reasoning already given for its sibling, and the ORPHA:67036 claim is hedged.

Re-validated: 50/50 snippets, `just validate`, `check-causal-targets`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading` and `check-folded-hyphens` all pass with no baseline edits.

One note on the shape of these two findings, since they are the same defect twice: the snippets were exact and unaffected in both cases. What was wrong was the interpretive prose *around* verified quotes — a denominator I got wrong, and edges pointing at nodes that did not exist. Everything a gate could check was green throughout.

---

PR #10096, opened 2026-08-29
Add 45,X/46,XY Mixed Gonadal Dysgenesis entry (MONDO:0015779)

### LINKED ISSUE #10087: Curate 45,X/46,XY mixed gonadal dysgenesis (MONDO:0015779)
Curate a dismech entry for **45,X/46,XY mixed gonadal dysgenesis** ([MONDO:0015779](https://monarchinitiative.org/MONDO:0015779)) — a sex-chromosome mosaic disorder of sex development in which one gonad is typically a streak and the other a dysgenetic testis, producing asymmetric internal and external genital development.

This sits between three entries dismech already has and duplicates none of them. `Turner_Syndrome` is 45,X without a Y-bearing line; `46_XY_complete_gonadal_dysgenesis` and `46_XY_Partial_Gonadal_Dysgenesis` are non-mosaic and monogenic. The mechanism here is different in kind from all three: it is not a gene defect but a mosaic cell population, so the gonadal outcome depends on how the two lines are distributed between the two gonads. That is what makes the asymmetry — streak on one side, dysgenetic testis on the other — the defining feature rather than an incidental one, and it is also why the Müllerian structures persist on the streak side only. Gonadoblastoma risk from the retained Y material is a further distinct thread.

The stub records no causal gene, which is correct for a chromosomal entity, and no MONDO descendants — a leaf.

Stub: `stubs/45_X_46_XY_Mixed_Gonadal_Dysgenesis.yaml`

caufieldjh:
New entry for 45,X/46,XY mixed gonadal dysgenesis, curated from `stubs/45_X_46_XY_Mixed_Gonadal_Dysgenesis.yaml`. The stub is deleted in this PR.

*(Body updated after the review round, which found six blocking items. See "What the review caught".)*

## Why this is a separate entry and not a duplicate

dismech already has `Turner_Syndrome`, `46_XY_complete_gonadal_dysgenesis` and `46_XY_Partial_Gonadal_Dysgenesis`. This entry sits between them and duplicates none of them, because **the cause here is a mosaic, not an allele**.

Turner syndrome is 45,X with no Y-bearing line. The two 46,XY dysgeneses are non-mosaic and usually monogenic. In MGD both lineages are present, and since testis determination is a *local* switch rather than a whole-body one, the two sides of the gonadal ridge can receive different proportions of the SRY-bearing lineage and take different developmental paths. That is why the defining feature is asymmetry — a streak gonad on one side, a dysgenetic testis on the other — rather than an incidental detail.

## Mechanism: two arms from one event

```
Post-Zygotic Loss of the Y Chromosome (MOLECULAR)
  -&gt; Regional SRY Dosage Variation Across the Gonadal Ridge (CELLULAR)   [PROVISIONAL]
       -&gt; Asymmetric Gonadal Differentiation (TISSUE)
            -&gt; Reduced Sertoli and Leydig Cell Output (ORGANISM)
                 -&gt; Incomplete Mullerian Regression (ORGANISM)
                 -&gt; Undervirilization of the External Genitalia (ORGANISM)
            -&gt; Gonadoblastoma Risk in Retained Dysgenetic Gonads (TISSUE)
  -&gt; SHOX Haploinsufficiency in the 45,X Lineage (MOLECULAR)
       -&gt; Turner-Spectrum Somatic Features (ORGANISM)   [PROVISIONAL]
```

The SHOX arm is deliberately parallel rather than downstream. It depends on where the 45,X cells landed, not on what the gonad did, which is why short stature occurs in patients whose genitalia are entirely unremarkable.

The two terminal nodes on the gonadal arm have **different mediators** and are kept separate for that reason: Müllerian persistence is driven by insufficient local AMH, undervirilization by insufficient testosterone. Splitting them makes the sharper point visible — AMH acts *locally*, so the retained hemi-uterus is ipsilateral to the streak gonad. That asymmetry is something a systemic hormone deficit could not produce, and it is the best available evidence that the hormonal lesion is regional in the same way the gonadal lesion is.

## Where the sources actually disagree, recorded as disagreements

Four discussions, each backed by the conflicting evidence rather than by an assertion that something is unknown:

**Growth hormone.** The 2025 review says rhGH "significantly improve[s] growth velocity and final adult height... with outcomes similar to those observed in Turner syndrome" and reports a height SDS gain of 0.42. A three-centre cohort of 100 patients found **no significant difference** between treated and untreated groups (p = 0.5). Both are in the entry. Neither can be dismissed: the review aggregates studies designed to measure the effect; the cohort is the largest series with long-term follow-up but is not randomised, so treated patients may have been the shorter ones to begin with.

**Does gonadal position predict tumour risk?** The management model in wide use assumes intra-abdominal gonads are the dangerous ones and scrotal testes can be watched. A 14-child series supports it strongly — every gonadoblastoma was intra-abdominal. A 21-child series does not: of four dysgenetic testes with carcinoma in situ, **two were scrotal**. The review separately warns that Y material raises risk "even in apparently normal testes located in the scrotal sac". If position is not protective, retaining a scrotal testis buys fertility potential at an unquantified oncological cost.

**`HUMAN_MODEL_MISMATCH`: the SRY threshold model is murine.** The SOX9–FGF9-versus-WNT4 switch was characterised by gain- and loss-of-function experiments in mouse, and its authors describe SRY's role as *tipping a balance* rather than acting as an on-off command — which is exactly what makes a mosaic ridge interesting. But no mouse model of 45,X/46,XY mosaicism exists, so threshold behaviour under partial SRY dosage has never been tested in the configuration this disease presents, and the authors generalise their own result only "in principle".

**Nobody has genotyped the two lineages inside a gonad.** Human MGD gonadal tissue *has* been examined immunohistochemically, so the tissue is obtainable; what has not been done is to genotype the two lineages within one gonad and relate their distribution to which regions became streak. The strongest human evidence for the regional model remains anatomical: six patients had a *single gonad containing both streak and testicular tissue* — which a regional model predicts and a whole-gonad model does not.

## The Y-linked tumour mechanism is grounded in the right tissue

Rather than asserting "retained Y material is oncogenic", the gonadoblastoma node cites immunohistochemistry done on **MGD patients' own gonadal tissue**, showing GBY-locus candidates TSPY and DDX3Y expressed in their germ cells, plus the cyclin B–CDK1 biochemistry that would make ectopic TSPY proliferative.

It also carries the caveat from the same paper: those proteins are found in germ cells of patients *without* malignancy too. So expression is not sufficient, and the dysgenetic environment is doing part of the work.

## Frequency bands only where a denominator exists, and the denominator is named

Streak gonad (85%) and gonadoblastoma (18%) are banded from a surgical series — patients who came to biopsy or gonadectomy, which is a more severely affected group than all 45,X/46,XY individuals. `review_notes` on both says so, and the gonadoblastoma one states explicitly that this is an upper bound on population risk.

Cardiac and renal malformations carry **no** band despite having numbers (6/23, 3/19), because those denominators count only patients who were investigated for them. Persistent Müllerian structures carry no band for a different reason worth stating: imaging systematically undercounts them — ultrasound misses roughly 40% of what laparoscopy finds — so any published frequency is an artifact of how you looked.

The prevalence record makes the same point twice over. The 1-in-15,000-to-30,000 figure is recorded as an ascertainment floor with the review's own caution quoted as evidence; the point value states its own derivation, because 4.4 per 100,000 is the midpoint taken in 1-in-N space and the arithmetic midpoint of the rates would be 5.0.

## Deliberate non-decisions, in `notes`

- **No `genetic` section.** There is no causal gene — MONDO records none. SHOX and SRY appear inside pathophysiology nodes because the mechanism runs through their dosage, but listing them under `genetic` would present a chromosomal disorder as a monogenic one.
- **No `animal_models`.** Mouse gonadal-determination work is cited as evidence for the switch, but those animals are not mosaic. They model the pathway the mosaicism perturbs, not this disease.
- **No `has_subtypes`.** The range is continuous and set by mosaic distribution, so any cut point would describe ascertainment rather than biology.
- **No `comorbidities` block** — `Disease` has no such slot. The adult Turner-like cardiometabolic morbidity is real and cited, so it is recorded as a `progression` phase; a proper `kb/comorbidities/` record would be the right follow-up.
- **Gonadal tissue cryopreservation and miR-371a-3p surveillance are out of scope.** Both are described as emerging rather than established, and putting prospective practice beside current practice without distinguishing them would misrepresent both.

## Disease identity was verified by hand

`just preflight-dr` returned **SKIP** — MONDO records no causal gene, so its gene-identity check cannot discriminate. Following the `dismech-references` skill, identity was confirmed manually: the report distinguishes MGD from 46,XY partial gonadal dysgenesis, Swyer syndrome and persistent Müllerian duct syndrome, and its top gene mentions are SRY, SOX9, WNT4 and AMH — matching the MONDO definition.

The report's **term validation** flagged one unresolved CURIE (`HP:0000850`) and four named as something else (`HP:0000053` is *Macroorchidism*, not clitoral hypertrophy). Two more failed on direct lookup: `UBERON:0003508`, offered as "Müllerian duct", is *pedal digit blood vessel*; `GO:0060065`, offered as "Wnt signaling in female gonad development", is *uterus development*. **None was bound.**

## What the review caught

Six blocking items, all fixed. The sharpest: **the entry curated a discussion arguing about the tumour risk of gonadal position while never recording that the gonads are undescended.** Also — one quantitative claim ("60 to 70 percent by the third decade") rested on evidence about a different population at a different timepoint, and now cites the sentence that actually says it.

Added in that round: persistent Müllerian structures, cryptorchidism, infertility, webbed neck and cubitus valgus as phenotypes; the node split above; hypospadias repair; the `HUMAN_MODEL_MISMATCH` discussion; a `biochemical` section for inhibin B and FSH; FISH/microarray and the ultrasound-versus-laparoscopy sensitivity gap in `diagnosis`; and a `progression` section.

## Validation

- `just validate` — schema clean, terms clean, **71/71 snippets** verified
- `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `validate-history` — all clean, no new baseline entries
- The PR caches more references than the entry cites; `review_notes` accounts for which were fetched by the report's citation-validation pass and which were read during curation and set aside.

Closes #10087

### DIFF
--- a/kb/disorders/45_X_46_XY_Mixed_Gonadal_Dysgenesis.yaml
+++ b/kb/disorders/45_X_46_XY_Mixed_Gonadal_Dysgenesis.yaml
@@ -52,8 +52,11 @@ prevalence:
   rate_high: 6.7
   notes: >-
     Reported as 1 in 15,000 to 1 in 30,000 live births, which is 3.3 to 6.7 per
-    100,000; the point value is the midpoint of that range and not itself a
-    published figure. The review says the true figure is probably higher, because
+    100,000. The point value is not a published figure and is not the arithmetic
+    midpoint of the rates, which would be 5.0; it is the midpoint taken in
+    1-in-N space, 1 in 22,500, converted to a rate. Both conventions are
+    defensible and they disagree, so the derivation is stated rather than left
+    for a reader to reconstruct. The review says the true figure is probably higher, because
     people whose phenotype is unremarkably male or female are never karyotyped -
     so this is an ascertainment floor rather than an estimate.
   evidence:
@@ -84,6 +87,45 @@ prevalence:
     evidence_source: HUMAN_CLINICAL
     snippet: "Among infants evaluated for atypical genitalia, MGD accounts for approximately 5%-15% of cases"
     explanation: Gives the share of the referral population, which is what this record reports.
+progression:
+- phase: Childhood growth decline
+  notes: >-
+    Height Z-score falls progressively through childhood rather than being fixed
+    at birth, which is why growth is monitored rather than assessed once.
+  evidence:
+  - reference: PMID:37882230
+    reference_title: "45,X/46,XY mosaicism: Clinical manifestations and long term follow-up."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "height Z-score was significantly decreased with age for both genders (p = 0.02)"
+    explanation: Documents the progressive height deficit.
+- phase: Pubertal and post-pubertal gonadal decline
+  notes: >-
+    Most male-raised patients enter puberty spontaneously, but progression stalls
+    with rising FSH and falling testosterone, and testicular insufficiency reaches
+    60 to 70 percent by the third decade. Normal early puberty is therefore not
+    reassurance.
+  evidence:
+  - reference: PMID:41208147
+    reference_title: "Mixed Gonadal Dysgenesis: A Comprehensive Review of Clinical Spectrum, Diagnostic Strategies, and Management Approaches."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Testicular insufficiency develops in 60%–70% of individuals raised as males by the third decade"
+    explanation: Gives the endpoint and its timing.
+- phase: Adult cardiometabolic morbidity
+  notes: >-
+    A population-based study found males with a 45,X/46,XY karyotype developing a
+    morbidity pattern resembling classic Turner syndrome by around age 40. This is
+    the argument for lifelong follow-up rather than discharge once the paediatric
+    endocrine and surgical questions are settled, and it applies to patients whose
+    genital phenotype was mild enough that they were never seen in a DSD clinic.
+  evidence:
+  - reference: PMID:41208147
+    reference_title: "Mixed Gonadal Dysgenesis: A Comprehensive Review of Clinical Spectrum, Diagnostic Strategies, and Management Approaches."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "with higher rates of hypertension, dyslipidaemia, and type 2 diabetes by age 40"
+    explanation: Reports the adult cardiometabolic pattern and its approximate timing.
 pathophysiology:
 - name: Post-Zygotic Loss of the Y Chromosome
   biological_scale: MOLECULAR
@@ -255,21 +297,32 @@ pathophysiology:
     explanation: >-
       Measures the Sertoli deficit directly by inhibin B and FSH, and shows it
       emerging even in boys whose genitalia were normal or near-normal.
+  - reference: PMID:41208147
+    reference_title: "Mixed Gonadal Dysgenesis: A Comprehensive Review of Clinical Spectrum, Diagnostic Strategies, and Management Approaches."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    directness: DIRECT
+    snippet: "Testicular insufficiency develops in 60%–70% of individuals raised as males by the third decade, and infertility is the most commonly reported outcome."
+    explanation: >-
+      Supplies the third-decade figure the description states, which the
+      end-of-puberty inhibin B result above does not cover.
   downstream:
-  - target: Incomplete Mullerian Regression and Undervirilization
+  - target: Incomplete Mullerian Regression
+    causal_link_type: DIRECT
+  - target: Undervirilization of the External Genitalia
     causal_link_type: DIRECT
 
-- name: Incomplete Mullerian Regression and Undervirilization
+- name: Incomplete Mullerian Regression
   biological_scale: ORGANISM
   role: consequence
   mechanism_confidence: ESTABLISHED
   description: >-
-    Anti-Mullerian hormone acts locally, so the side without a functioning testis
-    keeps its Mullerian derivatives while the side with one loses them. The result
-    is a hemi-uterus and fallopian tube remnant on the streak side, and a degree
-    of virilization of the external genitalia set by how much testosterone the
-    dysgenetic testis managed to produce during the critical window. Proximal
-    hypospadias with ventral curvature is the commonest external finding.
+    Anti-Mullerian hormone acts locally rather than systemically, so the side
+    without a functioning testis keeps its Mullerian derivatives while the side
+    with one loses them. That is why the retained hemi-uterus and fallopian tube
+    remnant are ipsilateral to the streak gonad - an asymmetry a systemic
+    hormone deficit could not produce, and the clearest evidence that the
+    hormonal lesion is regional in the same way the gonadal lesion is.
   locations:
   - preferred_term: Mullerian duct
     term:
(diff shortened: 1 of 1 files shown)


caufieldjh:
All six blocking items fixed in `aa6515fbb`, plus every suggestion. Snippets went 50 → 71, all verified.

**Item 2 was the sharpest observation in the review** and I want to acknowledge it specifically: the entry curated a whole discussion arguing about the tumour risk of gonadal position while never recording that the gonads are undescended. That is a real hole, not a completeness nit.

| Blocking item | Fix |
|---|---|
| 1. Müllerian structures uncurated | `HP:0034546` added, with the snippet that was already cited on the node |
| 2. Cryptorchidism missing | `HP:0000028` added |
| 3. Infertility missing; third-decade claim unsourced | `HP:0000789` added; the claim now cites the sentence that actually states it |
| 4. Bundled node | Split into `Incomplete Mullerian Regression` and `Undervirilization of the External Genitalia` |
| 5. No hypospadias repair | Added, targeting the new undervirilization node |
| 6. Mouse-derived node needs `HUMAN_MODEL_MISMATCH` | `mgd_sry_threshold_model_is_murine` added |

On item 3, you were right that the two claims were about different populations and timepoints. The correct sentence was in the same cached review: *"Testicular insufficiency develops in 60%–70% of individuals raised as males by the third decade, and infertility is the most commonly reported outcome"* — which sources the figure **and** the infertility phenotype in one quote.

On item 4, splitting made the entry better in a way I had not anticipated. Once the Müllerian node stands alone, its point becomes the ipsilaterality — AMH acts locally, so the retained hemi-uterus is on the streak side, and that asymmetry is something a systemic hormone deficit could not produce. Bundled with undervirilization, that argument was invisible.

On item 6, I took the distinction seriously rather than relabelling. The discussion argues two specific things the `PROVISIONAL` flag alone did not: no mouse model of 45,X/46,XY mosaicism exists, so threshold behaviour under partial SRY dosage has never been tested in the configuration this disease presents; and the authors of the mouse work generalise their own result only *"in principle"*, which is the right confidence level to carry across.

Suggestions — all six taken:

- **Midpoint arithmetic.** You were right and the wording was wrong. 4.4 is the midpoint in 1-in-N space (1 in 22,500); the arithmetic midpoint of the rates is 5.0. Both conventions are defensible and they disagree, so the note now states the derivation instead of calling it "the midpoint".
- **Biochemical section** added for inhibin B and FSH, including the prognostic use of early-pubertal inhibin B.
- **Webbed neck and cubitus valgus** added.
- **Short stature band** kept at `FREQUENT` with `review_notes` explaining why: the sources say "common" and "most patients had stunted growth" without a proportion, and the one number available is a mean height deficit rather than a count. `VERY_FREQUENT` needs a figure above 80% that nothing here provides.
- **FISH and chromosomal microarray** added to diagnosis, plus an imaging record carrying the ultrasound-versus-laparoscopy sensitivity gap — a negative pelvic ultrasound does not exclude Müllerian structures.
- **Adult cardiometabolic morbidity** recorded. `Disease` has no `comorbidities` slot, so rather than drop it I added a `progression` section: childhood growth decline, pubertal and post-pubertal gonadal decline, adult cardiometabolic morbidity. `notes` says a proper `kb/comorbidities/` record would be the right home and is a reasonable follow-up.

Gonadal tissue cryopreservation and miR-371a-3p surveillance are now explicitly out of scope in `notes`, as you asked: both are described as emerging rather than established, and adding them beside current practice without marking the difference would be misleading.

`just validate` clean at **71/71** snippets; `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` all clean with no new baseline entries.

---

Issue #7539, opened 2026-08-01
just refresh-orphadata fails its pinned sha256 check — no ORPHA evidence is citable


kevinschaper:
## Problem

`just refresh-orphadata` fails verification against the sha256 values pinned in `data/orphadata/MANIFEST.yaml`. Upstream Orphadata has evidently republished the bulk XML since the pin was taken (`snapshot_date: "2025-12-09"`, `schema_tag: en_product1+4+6+9_prev+9_ages`).

Because `data/orphadata/*.xml` is gitignored and therefore absent from every fresh clone and every new worktree, and because the refresh is the only supported way to obtain it, **the entire `ORPHA:` structured source is currently unusable for new curation.**

## Impact

`ORPHA:` records are the project's designated source for exactly the fields that are hardest to source from abstracts:

- population prevalence (`Prevalence` with `measure_type` / `prevalence_class` / `rate_per_100000`)
- HPO phenotype frequency rows (`Very frequent (99-80%)` etc., which map directly onto `FrequencyEnum`)
- gene–disease rows and cross-references

In the 12-disease #1079 batch, **five entries wanted ORPHA evidence and none could use it.** Each agent correctly declined to hand-write a cache file (per the `references_cache/` guardrail) and instead omitted the claim and recorded the limitation in `notes:`. Concretely:

| Entry | PR | Blocked ORPHA record |
|---|---|---|
| Bethlem myopathy | #7482 | `ORPHA:610` phenotype-frequency rows |
| Andersen-Tawil syndrome | #7493 | `ORPHA:37553` prevalence (only a qualitative `ULTRA_RARE` band asserted) |
| Double outlet right ventricle | #7490 | `ORPHA:3426` prevalence (re-sourced to PMIDs) |
| Allan-Herndon-Dudley syndrome | #7496 | `ORPHA:59` |
| Familial nonmedullary thyroid carcinoma | #7527 | `ORPHA:319494` |

So the failure is not cosmetic — it is silently degrading epidemiology and frequency coverage on every new entry, in a way that looks like curator omission rather than tooling breakage.

## Suggested fix

Follow the bump procedure already documented in the manifest header:

1. Re-download each `url` to a temp directory
2. `sha256sum` the files
3. Update `sha256` and `snapshot_date` in `data/orphadata/MANIFEST.yaml`
4. `just structured-rebuild-orphanet` and review the `references_cache/ORPHA_*.md` diff

Two things worth considering alongside the repin:

[...]

Issue #7522, opened 2026-08-01
Orphadata MANIFEST pin is stale — just refresh-orphadata fails checksum gate, blocking new ORPHA cache files


Phillip-a-richmond:
`just refresh-orphadata` fails its checksum gate because the upstream Orphadata
bulk XML has advanced past the pinned snapshot.

Verified directly:

```
$ curl -sI -L https://www.orphadata.com/data/xml/en_product1.xml
last-modified: Mon, 29 Jun 2026 12:20:47 GMT
content-length: 54026799
```

vs `data/orphadata/MANIFEST.yaml`:

```yaml
snapshot_date: "2025-12-09"
- name: en_product1.xml
  sha256: fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
  size_bytes: 53331679
```

The pin is ~7 months stale (53,331,679 → 54,026,799 bytes).

**Curation impact:** a curator cannot build a new `references_cache/ORPHA_*.md`
for a disorder that doesn't already have one. Encountered while curating
`MONDO:0012557` (#7510 / PR #7516) — the `ORPHA:91130` citation had to be
dropped, because building the cache from an unpinned snapshot would have made
that one file inconsistent with the 324 already-committed `ORPHA_*.md` files.

**Suggested fix:** re-pin `MANIFEST.yaml` to the current upstream snapshot,
regenerate all `references_cache/ORPHA_*.md`, and review the resulting diff for
content drift in existing entries. Worth checking whether any committed ORPHA
snippet quoted in a `kb/` evidence item changes text under the new snapshot —
that would silently break `validate-references`.

A periodic CI job that diffs upstream against the manifest would catch this
earlier.

matentzn:
Still live as of 2026-08-24, with a concrete curation cost, plus one datapoint that may help whoever bumps the pin.

**Current upstream release.** `https://www.orphadata.com/data/xml/en_product1.xml` today returns:

```
<JDBOR date="2026-06-23 07:53:50" version="1.3.42 / 4.1.8 [2025-03-03]" copyright="Orphanet (c) 2026">
  <DisorderList count="11645">
```

[...]

Issue #7234, opened 2026-07-31
Orphadata bulk XML has drifted from the pinned MANIFEST.yaml sha256 (refresh-orphadata fails)


sierra-moxon:
## Problem

`just refresh-orphadata` currently fails with a checksum mismatch: the upstream Orphadata bulk XML has drifted from the sha256 pinned in `data/orphadata/MANIFEST.yaml`.

Observed for `en_product1.xml`:

```
RuntimeError: checksum mismatch for en_product1.xml after download:
got df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb,
expected fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

Because the bulk XML is gitignored and only `MANIFEST.yaml` is committed, this means `references_cache/ORPHA_*.md` files **cannot be rebuilt** locally until the manifest is re-pinned to the current upstream snapshot — `just structured-rebuild-orphanet --id <n>` fails with `FileNotFoundError: data/orphadata/en_product1.xml`.

## Impact

Curators cannot fetch/refresh Orphanet-backed evidence (prevalence rows, HPO phenotype-frequency tables) for new entries. Concretely surfaced while curating **SLC35A3-CDG** (PR #7202): ORPHA:370943 was identified by deep research as the natural source for the entry's prevalence and phenotype frequencies, but the cache could not be built, so those fields were left empty. (Per repo guardrails, hand-editing `MANIFEST.yaml` or hand-writing `references_cache/ORPHA_*.md` is not something an agent should do unilaterally.)

## Suggested fix

Re-pin `data/orphadata/MANIFEST.yaml` to the current Orphadata snapshot (date + sha256 for each bulk file), then `just structured-rebuild-orphanet` and review the `references_cache/ORPHA_*.md` diff for drift. A periodic CI check that re-derives the manifest and flags drift (noted as a "worthwhile follow-up" in CLAUDE.md) would prevent silent staleness.

## Follow-up

Once the manifest is refreshed, SLC35A3-CDG (`kb/disorders/Autism_Spectrum_Disorder-Epilepsy-Arthrogryposis_Syndrome.yaml`) is a good first candidate to backfill with ORPHA:370943-evidenced prevalence and phenotype frequencies.

Issue #7199, opened 2026-07-31
Orphadata manifest pin is stale: refresh-orphadata fails checksum on en_product1.xml


kevinschaper:
## Summary

`just refresh-orphadata` currently fails: the file Orphadata serves for `en_product1.xml` no longer matches the sha256 pinned in `data/orphadata/MANIFEST.yaml` (`snapshot_date: "2025-12-09"`).

```
RuntimeError: checksum mismatch for en_product1.xml after download:
got      df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb
expected fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

Reproduce:

```bash
just refresh-orphadata     # or: uv run python -m dismech.structured_sources.cli refresh orphanet
```

The manifest pin is doing exactly its job here — it caught upstream drift rather than silently ingesting a different release. This issue is about deciding when to bump it.

## Why it matters

`references_cache/ORPHA_*.md` cannot be rebuilt or extended until the pin is bumped. That blocks two things:

1. **Adding an ORPHA row for a disorder not already cached.** Concrete case: `MED13L_Syndrome` (PR #7186) maps to `ORPHA:369891`, which has no cache file. The prevalence record there cites PMID:34654706 instead, which is fine but loses the Orphanet epidemiology class.
2. **Drift verification.** `CLAUDE.md` suggests running `just structured-rebuild-orphanet` and checking `git diff references_cache/ORPHA_*.md` to confirm no drift. That check cannot currently run.

## Why it wasn't fixed in PR #7186

Bumping the pin means re-downloading all pinned bulk files, updating `sha256` and `snapshot_date`, then regenerating and reviewing the diff across **all 321** existing `ORPHA_*.md` cache files. That is repo-wide maintenance with a large blast radius, and folding it into a two-disease curation PR would have buried it. Flagging separately per reviewer request.

## Suggested approach

- Bump all pinned Orphadata files together, not just `en_product1.xml`, so the snapshot stays internally consistent.
- Regenerate all 321 cache files in one dedicated PR and review the diff for content changes that invalidate existing curator-quoted snippets — a changed Orphanet definition or a reworded phenotype-frequency row would break `validate-references` on any entry quoting it. That is the main risk and the reason this deserves its own PR.
[...]

PR #10097, opened 2026-08-29
curate: Add CCDC115-CDG (MONDO:0014789), a Golgi-homeostasis liver disease

### LINKED ISSUE #10089: Curate CCDC115-CDG (MONDO:0014789)
Curate a dismech entry for **CCDC115-CDG** ([MONDO:0014789](https://monarchinitiative.org/MONDO:0014789)) — a type II congenital disorder of glycosylation (CDG-IIo) caused by biallelic *CCDC115* variants, which disrupt Golgi V-ATPase assembly and Golgi acidification, producing a hepatic-predominant phenotype with hypercholesterolemia, elevated transaminases and abnormal copper metabolism alongside a mixed N- and O-glycosylation defect.

Preflight: MONDO:0014789 is a leaf term with no descendants in the stub, one causal gene (`hgnc:28178` CCDC115, listed under its VMA22 alias), and no existing `kb/` entry or open PR. Expected to be a `DISEASE`.

Stub: `stubs/CCDC115-CDG.yaml`

matentzn:
New entry for **CCDC115-CDG** (congenital disorder of glycosylation type IIo) — an autosomal recessive defect of Golgi homeostasis whose phenotype is dominated by the liver rather than by the usual CDG multisystem picture. Closes #10089.

## Why the entry is built around Wilson disease

Low ceruloplasmin, abnormal copper handling, childhood liver disease. That reads as Wilson, and it is how these patients actually present — two of the thirteen reported were found inside an *ATP7B*-negative Wilson series (PMID:32043565), not by anyone suspecting a glycosylation disorder. Niemann-Pick type C and mitochondrial hepatopathies are the other named mimics.

The separator is plasma transferrin glycosylation, not the copper studies. That is in both `diagnosis` and `differential_diagnoses` because it is the single most useful thing in the entry.

## Mechanism

Five nodes: biallelic loss of function → impaired V-ATPase assembly and organelle acidification → branching into (a) the combined N- and mucin-type O-glycosylation defect and (b) lysosomal lipid accumulation with hepatic steatosis → converging on hepatic injury and fibrosis.

The convergence node is `PROVISIONAL` and both edges into it are `INDIRECT_UNKNOWN_INTERMEDIATES`. Which upstream defect drives the liver disease is genuinely unsettled, and the 2018 series says its account of hepatic pathogenesis is hypothesis rather than demonstration. That is one of two discussions.

**A circularity flagged in place.** The acidification measurement that exists is *lysosomal*, in CCDC115-silenced HepG2 cells. Golgi luminal pH has not been measured in patient cells — it is argued from the glycosylation phenotype, which is then explained by the pH defect. The defining paper's own remark that the spectrum here is distinct from V-ATPase core-subunit defects is an argument for measuring rather than assuming. Second discussion.

## Gene symbol: a disagreement I recorded rather than resolved

Live HGNC returns **VMA22** for HGNC:28178, with CCDC115 as a previous symbol. The **OBO HGNC build this repo validates against still returns CCDC115**, and binding `VMA22` fails term validation here with a label mismatch.

The entry binds `CCDC115` — what the configured authority says, what MONDO says, what every paper says. The consequence, stated in the entry's `genetic` notes and description, is that anyone matching this entry against a current HGNC download will not match on symbol and will need the ID. I did not work around it per-entry; the fix is an OBO HGNC refresh. Flagging in case a maintainer wants to pick that up. (The MONDO stub already carried the VMA22 label, which is how it surfaced.)

One Perplexity run, committed alongside, used for orientation only. It was accurate in outline. It contains no PMID or DOI anywhere in its body, and its term validation reported **8 of 13 checked labels naming a different term** — and not randomly: every NCIT suggestion in its treatments section was wrong. `NCIT:C15273` offered as *Liver Transplantation* is *Longitudinal Study*; `NCIT:C28254` as *Speech Therapy* is *Milliliter*; `NCIT:C15279` as *Occupational Therapy* is *Radical Mastectomy*. This entry uses `NCIT:C15271`, verified independently.

## Evidence grading worth a look

Three items are `SUPPORT` + `directness: INDIRECT` for different reasons, each stated in its own explanation: the V-ATPase general principle is background framing in its source rather than a measured result; the low-ceruloplasmin item is from *TMEM199* patients, cited to show the finding belongs to the mechanism class rather than to CCDC115 alone; and the supportive-care item is cited for what the most recent whole-caseload review **does not** contain, which is an absence rather than a positive finding.

Denominator: **thirteen reported patients worldwide**. Every frequency band rests on that, and the prevalence notes say so.

## Validation

Schema validation passes. **38/38 evidence snippets verified** against cached primary references. Term validation, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass. `just validate-disorders` passes.

---

### DIFF
--- a/history/disorders/CCDC115-CDG/2026-09-03T025800Z-claude-code-bbe5dd.yaml
+++ b/history/disorders/CCDC115-CDG/2026-09-03T025800Z-claude-code-bbe5dd.yaml
@@ -1,0 +1,25 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: CCDC115-CDG
+  path: kb/disorders/CCDC115-CDG.yaml
+session:
+  id: 2026-09-03T025800Z-claude-code-bbe5dd
+  timestamp: '2026-09-03T02:58:00Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10097
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Fix prefixed pathograph targets
+  details: |
+    check-causal-targets, a gate added to main after this entry was written, found 18 causal targets using the <kind>#<name> entity-reference grammar in slots that take a bare name. These are matched verbatim with no resolution step, so the renderer drew phantom duplicate nodes and dropped the real ones from the graph, fragmenting the pathograph while every other gate passed. Stripped the prefixes; build_causal_graph now reports orphan_targets empty. Snippet counts and all other gates unchanged.
--- a/kb/disorders/CCDC115-CDG.yaml
+++ b/kb/disorders/CCDC115-CDG.yaml
@@ -164,7 +164,7 @@ pathophysiology:
       The only proposed genotype-phenotype correlation, derived from a review of
       all thirteen reported patients.
   downstream:
-  - target: pathophysiology#Impaired V-ATPase Assembly and Organelle Acidification
+  - target: Impaired V-ATPase Assembly and Organelle Acidification
     causal_link_type: DIRECT
 - name: Impaired V-ATPase Assembly and Organelle Acidification
   biological_scale: CELLULAR
@@ -240,11 +240,11 @@ pathophysiology:
       downstream defects. Graded INDIRECT because it is background framing in
       that paper rather than a result it measures.
   downstream:
-  - target: pathophysiology#Combined N- and O-Glycosylation Defect
+  - target: Combined N- and O-Glycosylation Defect
     causal_link_type: DIRECT
-  - target: pathophysiology#Lysosomal Lipid Accumulation and Failed Lipophagy
+  - target: Lysosomal Lipid Accumulation and Failed Lipophagy
     causal_link_type: DIRECT
-  - target: phenotypes#Decreased Circulating Ceruloplasmin
+  - target: Decreased Circulating Ceruloplasmin
     causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
     description: >-
       The second and clinically more pointed half of the copper mechanism, and
@@ -333,9 +333,9 @@ pathophysiology:
       at present uncertain - and no measurement of ATP7A glycosylation in these
       patients is reported.
   downstream:
-  - target: phenotypes#Abnormal Protein Glycosylation
+  - target: Abnormal Protein Glycosylation
     causal_link_type: DIRECT
-  - target: phenotypes#Decreased Circulating Ceruloplasmin
+  - target: Decreased Circulating Ceruloplasmin
     causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
     description: >-
       Ceruloplasmin is a secreted glycoprotein, and the proposed route runs
@@ -344,16 +344,16 @@ pathophysiology:
       and loss of copper export into serum would lower ceruloplasmin. The
       intermediates are named in the source and the mechanism is stated there as
       uncertain, which is why this is an inference rather than a direct edge.
-  - target: phenotypes#Global Developmental Delay
+  - target: Global Developmental Delay
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
     description: >-
       Neurological involvement is reported in five of the eleven patients in the
       systematic review, but nothing traces a route from the glycosylation
       defect to it. The edge records that the neurological arm exists and is
       unexplained rather than leaving it disconnected.
-  - target: phenotypes#Hypotonia
+  - target: Hypotonia
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: pathophysiology#Hepatic Injury and Fibrosis
+  - target: Hepatic Injury and Fibrosis
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
 - name: Lysosomal Lipid Accumulation and Failed Lipophagy
   biological_scale: CELLULAR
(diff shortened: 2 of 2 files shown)


matentzn:
All four IMPORTANT items and all three suggestions addressed and pushed. Nothing disputed. Snippets **38 → 45 verified**.

**IMPORTANT 1 — orphan phenotypes.** The ceruloplasmin case was the one that mattered: the entry is built around the Wilson mimicry and the finding that drives it had no mechanistic link. It now hangs off the glycosylation node by an `INDIRECT_KNOWN_INTERMEDIATES` edge with the intermediates *named* rather than gestured at — ATP7A is glycosylated, deficient glycosylation could impair its function, loss of copper export into serum lowers ceruloplasmin. Two new evidence items carry that, the second graded `directness: INDIRECT` because its own source says the mechanism is at present uncertain.

The neurological arm is wired by `INDIRECT_UNKNOWN_INTERMEDIATES` edges with a description saying explicitly that nothing traces a route — that seemed better than inventing an intermediate node to make the graph look complete. The hepatic injury node picked up four more edges.

**IMPORTANT 2 — node scale.** Agreed, and the CLAUDE.md rule is the right one to invoke. Split into `Lysosomal Lipid Accumulation and Failed Lipophagy` (CELLULAR) with `phenotypes#Hepatic Steatosis` downstream, which resolves that orphan too. All references to the old node name updated (incoming edge and the discussion's `attaches_to`).

**IMPORTANT 3 — claim-evidence mismatch.** Agreed: the genotype-correlation sentence presupposes the frequency rather than establishing it. The core-presentation sentence now carries `VERY_FREQUENT`; the genotype sentence stays as an explicitly weaker second claim about *magnitude*, and its explanation says so.

**IMPORTANT 4 — the research cross-check.** You were right, and the omission had a cause worth naming: **PMID:34173795 (PMC9255677) was fetched during initial curation and then deleted from the commit as "uncited".** That was a mistake — it is the more recent systematic review, supersedes the 2017 one I did cite, and carries CCDC115-specific content the older one lacks. It is now cited, and one enumerating sentence in it supports three new phenotypes: **Cholestasis**, **Jaundice** and **Abnormal Facial Shape**.

**Seizures I deliberately did not add**, and the entry now says why. The deep-research report lists them; the systematic review says five of eleven patients had "neurological symptoms" without specifying seizures, and no cached source attributes seizures to this disease. Adding them would have been consuming the report's claim rather than the literature's.

ORPHA:468684 could not be fetched — not among the 347 cached ORPHA records, and `just refresh-orphadata` fails on the checksum mismatch against the pinned manifest.

**Suggestions.** GeneReviews negative result recorded in the sibling-entry form (PMID:20301507 tagged, notes stating it is a retired archival overview with no CCDC115 content, PMID:34173795 named as the equivalent baseline). `target_mechanisms` added to Liver Transplantation → `Hepatic Injury and Fibrosis`; `treatment_effect: RESTORES` with the description stating that is the closest available enum value rather than an exact one — the graft does not reverse the fibrosis, it removes the tissue carrying it.

**On the OBO HGNC tracking issue — I did not file one, deliberately.** Before filing anything I checked, and the repo already has the *Orphadata* pin failure open **eight** times (#7199, #7234, #7522, #7539, #7622, #8836, #9687, #9897), all filed by agents hitting the same wall mid-curation. Adding a ninth-of-its-kind for HGNC seemed like a cost rather than a contribution. The CCDC115/VMA22 disagreement is recorded in the entry's `genetic` notes where a curator hitting it will see it, and I'm raising it here instead: **live HGNC has renamed HGNC:28178 to VMA22; the OBO HGNC build this repo validates against still says CCDC115, and binding VMA22 fails term validation.** If a maintainer wants that tracked, I'm happy to open it — I just didn't want to guess.

---

matentzn:
Blocking item fixed, both actionable suggestions taken, and the third is now filed. Snippets **45 → 46 verified**.

**The cholestasis inversion — you're right, and I introduced it in the last fix commit.** The sentence all four hepatic phenotypes are cited from ranks them explicitly, and I graded cholestasis `FREQUENT` while `Hepatic Failure` sat at `OCCASIONAL` — inverting that order, two lines below a description saying cholestasis is the least frequent of the four.

Downgraded cholestasis to `OCCASIONAL` rather than raising hepatic failure, and the description says why: the source gives an ordinal ranking with no counts, so the band has to sit at or below liver failure's. **Raising `Hepatic Failure` to `FREQUENT` was the other valid fix and I declined it** because it would have imported a second denominator — 3/8 in the defining series rather than 3/13 across the caseload — without saying so.

Which is the more useful half of your finding. The denominator *was* invisible, and the answer is that this entry uses **two**: thirteen where a count exists, and the source's ordinal ranking where it does not. The `notes` now say that, name the alternative, and record that hepatic failure would fall in a different band against the eight-patient series.

That is worth flagging past this PR: **a `frequency` band carries no denominator**, so two entries can disagree by a whole band while each being correct against its own source, and nothing in the schema or the validators can see it. This is the second time in this claim run that an evidence-metadata gap has only been visible in prose.

**The ceruloplasmin edge — taken, and it improves the entry more than the blocking fix did.** You're right that I used the weaker half. ATP7B is explicitly *not* glycosylated, so it cannot sit downstream of the glycosylation node; the source ties it to V-ATPase malfunction via Golgi residence. A second edge now runs from the acidification node with that sentence as evidence, `directness: INDIRECT` since the source calls it a postulated link.

Your point about why it matters is the one I've put in the edge description: **ATP7B is the Wilson gene**, so this is what makes the mimicry mechanistic rather than coincidental — the same organelle failure that breaks glycosylation is proposed to disable the transporter Wilson disease disables. That belongs in an entry built around the mimicry.

**Both intentional orphans now say so** — bone-derived origin that a hepatic edge would contradict, and no published craniofacial route.

**HGNC tracking issue: filed, at #10102.** I declined this twice and was wrong to on the second pass. My reasoning was about not adding a ninth duplicate of the Orphadata checksum issue — but that does not transfer: I searched, and nothing covers the OBO HGNC build drifting from live HGNC. Different mechanism too (an OBO build lagging a registry rename, not a checksum gate on a bulk download). The issue is scoped honestly: it names the one gene I actually found, says I have not swept the cache to find out how many others there are, and proposes that sweep as the next step, since one stale symbol is a curiosity and fifty is a data-quality problem.

---

matentzn:

`check-causal-targets` was added to `main` after this entry was written, and it fails here on **18 causal targets** — the most of the five entries in this run. I had written them with the `<kind>#<name>` entity-reference grammar in slots that take a **bare name**. That grammar is right for `attaches_to` and `would_support`; it is wrong for `downstream[].target`.

Because targets are matched verbatim with no resolution step, the effect was not a missing arrow: the renderer drew phantom duplicate nodes labelled with the literal prefixed strings, and the real nodes dropped out of the graph. `build_causal_graph` reported 18 orphan targets here; it now reports zero, with 20 resolving edges.

That is worth spelling out for this entry in particular, because the last two review rounds were spent on exactly this graph — wiring ceruloplasmin off the glycosylation node, adding the ATP7B edge from the acidification node, connecting hepatic steatosis. Every one of those edges was landing on a phantom node. The reasoning was right and the wiring was inert.

Nothing else changed: 46/46 snippets still verify, and `just validate`, `check-entity-refs` and `check-duplicate-keys` pass.

The same fix went to all five PRs in this run (#10095, #10098, #10100, #10101) — 44 targets in total, all mine.

---

PR #10098, opened 2026-08-29
curate: Add Shashi-Pena Syndrome (MONDO:0014963, ASXL2)

### LINKED ISSUE #10090: Curate Shashi-Pena syndrome (MONDO:0014963)
Curate a dismech entry for **Shashi-Pena syndrome** ([MONDO:0014963](https://monarchinitiative.org/MONDO:0014963)) — the *ASXL2*-related developmental disorder, caused by de novo truncating variants in the PR-DUB scaffold ASXL2, with macrocephaly, hypotonia, feeding difficulty, glabellar naevus flammeus and, distinctively among the ASXL disorders, a cardiac and hypoglycaemia burden.

Preflight: MONDO:0014963 is a leaf term with one causal gene (`hgnc:23805` ASXL2), no existing `kb/` entry and no open PR. `kb/disorders/Bainbridge-Ropers_Syndrome.yaml` (ASXL3) already names it as a distinct differential in its `differential_diagnoses` block, which is a positive signal that it is its own entry rather than a subtype. Expected to be a `DISEASE`.

Stub: `stubs/Shashi-Pena_Syndrome.yaml`

matentzn:
New entry for **Shashi-Pena syndrome** (SHAPNS), the *ASXL2* developmental disorder. Closes #10090.

This completes the ASXL set in the KB: `Bohring-Opitz_syndrome` (*ASXL1*), `Bainbridge-Ropers_Syndrome` (*ASXL3*), and now this. The Bainbridge-Ropers entry already names Shashi-Pena as a distinct differential with its own MONDO term, so the separate-entry call was easy.

The three share a core of developmental delay, hypotonia, feeding difficulty and dysmorphism, so the entry is written around what separates them: **macrocephaly and absence of growth retardation**, plus much wider variability in intellectual outcome. ASXL1 and ASXL3 restrict growth; this one does not.

## The mechanism is unsettled and I did not resolve it quietly

Two published observations point opposite ways:

- The 2016 defining study argues **dominant-negative** — truncated transcripts escape NMD, both alleles are expressed in blood. Its own wording is "likely to exert", not "demonstrated".
- A 2021 family reached the same diagnosis via a balanced t(2;11) translocation with RNA-seq showing **decreased** *ASXL2* mRNA. That is haploinsufficiency.

`genetic_context` carries `functional_impact_category: DOMINANT_NEGATIVE`, and the discussion says plainly that this is a choice made with the contrary evidence in view. If a reviewer thinks the field has moved on this, that is the line to push back on.

## Things worth a specific look

**One evidence item is `NO_EVIDENCE` on purpose.** The chromatin node cites a sentence from the ASXL cancer review saying it is unclear whether ASXL1 truncating mutations give loss of protein or a stable truncated one. That sentence is about *ASXL1*, so it does not bear on this node either way. Grading it `NO_EVIDENCE` keeps it citable as a record that the question is open across the family without pretending it supports the claim.

**The mouse is entered with both a success and a mismatch.** One `animal_models` entry, two links: `PARTIALLY_RECAPITULATES` / `LOW` against the chromatin node (the readout is myeloid, not neurodevelopmental, and a deletion may test a different lesion than a truncating dominant-negative allele), and `RECAPITULATES` / `MODERATE` against the metabolic node. The metabolic one is the rare case where a mouse phenotype *preceded and correctly predicted* a human one — ASXL2-null mice are glucose-intolerant and lipodystrophic via PPAR-γ, and a prepubertal case of type 2 diabetes with dyslipidemia and hepatic steatosis has since been reported. `MODERATE` reflects one met prediction, not a systematic comparison, and the `limitations` say so.

**Somatic ≠ germline.** *ASXL2* is recurrently mutated in AML with t(8;21) and its mouse deletion causes MDS-like disease. The genetic notes state explicitly that this does **not** license a cancer-risk claim here: GeneReviews recommends no malignancy surveillance, no SHAPNS patient with myeloid malignancy appears in the reviewed material, and the variant classes differ.

## A schema gap I hit and did not work around

`classifications.mechanistic_category` has no value for a chromatin/epigenetic-machinery disorder — the enum is RASopathy, ciliopathy, laminopathy, collagenopathy, desmosomopathy, mitochondrial disease, amyloidopathy, tauopathy, synucleinopathy, intermediate filament disease, proteotoxic disease. I removed the block rather than force an ill-fitting value. The KB already holds several disorders of the epigenetic machinery, so a chromatinopathy value may be worth adding — but that is a schema change and does not belong in a curation PR. Flagging for a maintainer.

One Perplexity run, committed alongside, orientation only. No PMID or DOI anywhere in its body. Its term validation reported **6 of 7 checked labels naming a different term**, and two of those are the worst kind — two real UBERON terms with their labels swapped: `UBERON:0004288` offered as *vertebral column* (UBERON: **skeleton**) and `UBERON:0008896` offered as *skeleton* (UBERON: **post-hyoid pharyngeal arch**). Neither the CURIE nor the name looks wrong in isolation. Every CURIE here was verified independently against OLS.

GeneReviews (PMID:39527683) is the phenotype baseline and is tagged. Its Clinical Characteristics list was cross-referenced; a handful of features (skin findings, hypermobility, temperature dysregulation, sleep apnea, MRI volume loss) are represented in prose rather than as structured phenotypes, because each would need its own evidence item quoting the same long list sentence. Say the word if you'd rather have them structured.

## Validation

Schema validation passes. **42/42 evidence snippets verified** against cached primary references. Term validation, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass. `just validate-disorders` passes.

---

### DIFF
--- a/history/disorders/Shashi-Pena_Syndrome/2026-09-03T025807Z-claude-code-19157f.yaml
+++ b/history/disorders/Shashi-Pena_Syndrome/2026-09-03T025807Z-claude-code-19157f.yaml
@@ -1,0 +1,25 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Shashi-Pena_Syndrome
+  path: kb/disorders/Shashi-Pena_Syndrome.yaml
+session:
+  id: 2026-09-03T025807Z-claude-code-19157f
+  timestamp: '2026-09-03T02:58:07Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10098
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Fix prefixed pathograph targets
+  details: |
+    check-causal-targets, a gate added to main after this entry was written, found 6 causal targets using the <kind>#<name> entity-reference grammar in slots that take a bare name. These are matched verbatim with no resolution step, so the renderer drew phantom duplicate nodes and dropped the real ones from the graph, fragmenting the pathograph while every other gate passed. Stripped the prefixes; build_causal_graph now reports orphan_targets empty. Snippet counts and all other gates unchanged.
--- a/kb/disorders/Shashi-Pena_Syndrome.yaml
+++ b/kb/disorders/Shashi-Pena_Syndrome.yaml
@@ -147,7 +147,7 @@ pathophysiology:
       the source's own hedge - "likely to exert" - and the contrary evidence
       recorded in the discussion below.
   downstream:
-  - target: pathophysiology#Disturbed PR-DUB Chromatin Regulation
+  - target: Disturbed PR-DUB Chromatin Regulation
     causal_link_type: DIRECT
 - name: Disturbed PR-DUB Chromatin Regulation
   biological_scale: MOLECULAR
@@ -221,9 +221,9 @@ pathophysiology:
       that the mechanism question is open across the whole family rather than
       only here.
   downstream:
-  - target: pathophysiology#Dysregulated Developmental Patterning
+  - target: Dysregulated Developmental Patterning
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: pathophysiology#PPAR-gamma-Linked Metabolic Dysregulation
+  - target: PPAR-gamma-Linked Metabolic Dysregulation
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
 - name: Dysregulated Developmental Patterning
   biological_scale: ORGANISM
@@ -270,9 +270,9 @@ pathophysiology:
       Argues the overgrowth framing explicitly, from a patient at the mild end of
       the neurodevelopmental spectrum.
   downstream:
-  - target: phenotypes#Macrocephaly
+  - target: Macrocephaly
     causal_link_type: DIRECT
-  - target: phenotypes#Global Developmental Delay
+  - target: Global Developmental Delay
     causal_link_type: DIRECT
 - name: PPAR-gamma-Linked Metabolic Dysregulation
   biological_scale: ORGANISM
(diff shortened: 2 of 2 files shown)


matentzn:
Blocking item fixed, two of three suggestions taken, one declined with a reason. Snippets **42 → 45 verified**.

**Congenital heart defects — added.** Your diagnosis of why it slipped is exactly right: it was inside two snippets already in the file but never curated as a phenotype and, unlike sleep apnea or temperature dysregulation, not covered by the scoping rationale either. It fell into a gap between "curated" and "explicitly deferred". Now `HP:0001627` (OCCASIONAL) with two evidence items — the GeneReviews sentence already quoted elsewhere, plus the independent 2025 case report. Binding kept general for the reason you gave, and the description says so.

**Strabismus — added** (`HP:0000486`, OCCASIONAL) from the same sentence. Amblyopia left out as a consequence of the strabismus rather than an independent finding.

**Octreotide — `PEPTIDE`.** Agreed, it's an octapeptide.

**Third `modeled_mechanisms` link for the Asxl2 cardiac phenotype — declined**, on the condition you attached to it. A PubMed search for an Asxl2 cardiac or skeletal mouse study returned nothing fetchable reporting it as a primary result. Adding the link on the strength of the deep-research report's assertion would be the exact failure this entry's own notes warn about, given that report has no PMID or DOI anywhere in its body. If a primary source surfaces it's a small addition.

**One thing your review exposed that I've now fixed properly.** The PR body claimed a handful of GeneReviews features were "represented in prose rather than as structured phenotypes" — but that prose was *in the PR description*, not in the entry, so it was invisible to anyone reading the file. That is precisely the gap the cardiac phenotype fell through. The `notes` block now carries the scoping paragraph: which features are curated from the long clinical-characteristics sentence, which are not, and that the omission is about snippet economy rather than a judgement that they're absent. Thanks — that was the more useful half of the finding.

Noting your two explicit non-flags, since both were deliberate: the `NO_EVIDENCE` grading on the ASXL1 sentence, and `DOMINANT_NEGATIVE` paired with the `KNOWLEDGE_GAP` laying out the contrary haploinsufficiency evidence. Good to have them confirmed as the right handling rather than hedges.

---

matentzn:

`check-causal-targets` was added to `main` after this entry was written, and it fails here on 6 causal targets. I had written them with the `<kind>#<name>` entity-reference grammar in slots that take a **bare name** — correct for `attaches_to` and `would_support`, wrong for `downstream[].target`.

Targets are matched verbatim with no resolution step, so the renderer drew phantom duplicate nodes labelled with the literal prefixed strings while the real nodes dropped out of the graph. `build_causal_graph` reported 6 orphan targets; it now reports zero, with 9 resolving edges.

Nothing else changed. 45/45 snippets still verify, and `just validate`, `check-entity-refs` and `check-duplicate-keys` pass.

The same fix went to all five PRs in this run (#10095, #10097, #10100, #10101) — 44 targets in total, all mine.

---

PR #10100, opened 2026-08-29
curate: Add Weiss-Kruszka Syndrome (MONDO:0032836, ZNF462)

### LINKED ISSUE #10091: Curate Weiss-Kruszka syndrome (MONDO:0032836)
Curate a dismech entry for **Weiss-Kruszka syndrome** ([MONDO:0032836](https://monarchinitiative.org/MONDO:0032836)) — the *ZNF462*-haploinsufficiency neurodevelopmental and craniofacial disorder, recognisable by metopic ridging, ptosis and a highly consistent facial gestalt, with developmental delay and variable structural anomalies.

Preflight: MONDO:0032836 is a leaf term with one causal gene (`hgnc:21684` ZNF462), and neither the term nor the label appears anywhere in `kb/` or in an open PR. Expected to be a `DISEASE`.

Stub: `stubs/Weiss-Kruszka_Syndrome.yaml`

matentzn:
New entry for **Weiss-Kruszka syndrome** (WSKA), the *ZNF462* haploinsufficiency disorder. Closes #10091.

## Why this one is interesting to curate

The mechanism went from "zinc-finger protein of unknown function" — the defining paper's own words in 2017 — to a specific molecular account by 2023. ZFP462, the murine homolog, turned up in a CRISPR screen for mediators of epigenetic silencing: it binds transposable elements acting as enhancers for pluripotency and meso-endodermal transcription factors, and recruits the H3K9 methyltransferase complex G9A/GLP to seed heterochromatin there.

The conceptual point is the one worth having in the KB: **G9A/GLP is broadly expressed and has no sequence specificity of its own.** ZNF462 supplies the lineage and locus specificity. So the proposed basis of the neurodevelopmental phenotype is failure of *cell-fate specification* — non-neural programmes firing during neural differentiation — not loss of a neuronal gene product.

Four pathophysiology nodes: haploinsufficiency → failure of G9A/GLP heterochromatin seeding → ectopic meso-endodermal expression in the neural lineage, plus the blood methylation signature as a node in its own right.

## The whole mechanism is murine, and the entry says so

Both mechanistic nodes are `PROVISIONAL`. ZFP462 was found in a mouse screen; the binding, the G9A/GLP recruitment, the heterochromatin seeding and the ectopic expression were all measured in mouse ESCs and NPCs. Nothing equivalent has been done in human cells carrying a patient variant.

The step from mouse cell biology to human phenotype is one evidence item, `SUPPORT` + `directness: INDIRECT`, quoting the authors' own *"Our results suggest that…"*. That is a `HUMAN_MODEL_MISMATCH` discussion with a proposed experiment (patient iPSC-derived NPCs, chromatin accessibility and H3K9me2 at the mouse-identified enhancers) — which would also answer the question the mouse work cannot, since it was done on complete loss and the human disease is 50% dosage.

## Two animal models, both cited second-hand, both saying so

The `Zfp462+/-` mouse is the rare case where the **genotype matches** the human dosage state, and it shows developmental delay, low body and brain weights, anxiety-like behaviour. Its `limitations` record that the defining craniofacial phenotype is not reported in it, the corpus callosum was not assessed in the cited material, and excessive self-grooming is a weak proxy for the 33% autism rate in patients.

The *Xenopus* knockdown is `fidelity: LOW` and is in for exactly one reason, stated in `limitations`: the **rescue by human ZNF462 mRNA**, which is what licenses reading the mouse and frog work as being about the same gene function.

Both carry `directness: INDIRECT` because both reach this entry through the human delineation paper's introduction rather than the primary studies. That is a real limit on what can be quoted, and it is recorded rather than hidden.

## Two source defects preserved verbatim

Snippets must be exact substrings, so two errors are quoted as-is and corrected in prose:

- GeneReviews' diagnosis sentence says **"deletion of 9p31.2"**. HGNC and the rest of the literature place *ZNF462* at **9q31.2**, and the same chapter says 9q31.2 elsewhere. The `p` is a typo.
- The delineation paper writes **"the coding region of ZNF42"** for *ZNF462*.

A third snippet carries inline reference markers (`however,4,5there are several…`) from the cached full text. Left as fetched.

## Historical entanglement with the other PR in this run

The first reported case of what became this syndrome carried a **t(2;9) translocation disrupting both *ZNF462* and *ASXL2*** — *ASXL2* being Shashi-Pena syndrome, which I'm also submitting in this claim run (#10098). Its phenotype is presumed to be the sum of two disorders. The genetic notes and the differential block both say the pre-2017 literature on that translocation must not be attributed to *ZNF462*.

## An autoimmunity signal kept as a discussion, not a phenotype

Two of roughly forty reported patients have autoimmune disease (PMID:36688693 SLE; PMID:39069253, the second case, which asks for functional studies). Two of forty is not a signal and I did **not** curate it as a phenotype. It is an open `KNOWLEDGE_GAP` because the mechanism is not implausible for a chromatin-remodelling disorder, and because the same pattern surfaced independently in Shashi-Pena syndrome. Both PMIDs are cached and named in the discussion prose so a third case can be judged against them.

## Ascertainment, recorded because it shapes the phenotype

Nine of the fourteen new individuals in the delineation series were found through **GeneMatcher** — by matching an already-sequenced variant to another, not by recognising a phenotype. The facial gestalt was characterised *afterwards*. That is the reverse of the classical order, and it is why facial-analysis algorithms feature so heavily in this literature.

## Validation

Schema validation passes. **41/41 evidence snippets verified** against cached primary references. Term validation, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass. `just validate-disorders` passes.

---

### DIFF
--- a/history/disorders/Weiss-Kruszka_Syndrome/2026-09-03T025756Z-claude-code-acd69c.yaml
+++ b/history/disorders/Weiss-Kruszka_Syndrome/2026-09-03T025756Z-claude-code-acd69c.yaml
@@ -1,0 +1,25 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Weiss-Kruszka_Syndrome
+  path: kb/disorders/Weiss-Kruszka_Syndrome.yaml
+session:
+  id: 2026-09-03T025756Z-claude-code-acd69c
+  timestamp: '2026-09-03T02:57:56Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10100
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Fix prefixed pathograph targets
+  details: |
+    check-causal-targets, a gate added to main after this entry was written, found 6 causal targets using the <kind>#<name> entity-reference grammar in slots that take a bare name. These are matched verbatim with no resolution step, so the renderer drew phantom duplicate nodes and dropped the real ones from the graph, fragmenting the pathograph while every other gate passed. Stripped the prefixes; build_causal_graph now reports orphan_targets empty. Snippet counts and all other gates unchanged.
--- a/kb/disorders/Weiss-Kruszka_Syndrome.yaml
+++ b/kb/disorders/Weiss-Kruszka_Syndrome.yaml
@@ -203,9 +203,9 @@ pathophysiology:
       at 9q31.2 and the same chapter says 9q31.2 elsewhere, so this is a
       typographical error in the source, quoted verbatim rather than corrected.
   downstream:
-  - target: pathophysiology#Failure of G9A/GLP Heterochromatin Seeding at Enhancers
+  - target: Failure of G9A/GLP Heterochromatin Seeding at Enhancers
     causal_link_type: DIRECT
-  - target: pathophysiology#Altered Blood DNA Methylation Signature
+  - target: Altered Blood DNA Methylation Signature
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
     description: >-
       The episignature is a consequence of the haploinsufficiency and is measured
@@ -318,7 +318,7 @@ pathophysiology:
       heterochromatin machinery the G9A/GLP account puts it in by a different
       route. Graded INDIRECT for the same second-hand-citation reason.
   downstream:
-  - target: pathophysiology#Ectopic Meso-Endodermal Gene Expression in the Neural Lineage
+  - target: Ectopic Meso-Endodermal Gene Expression in the Neural Lineage
     causal_link_type: DIRECT
 - name: Ectopic Meso-Endodermal Gene Expression in the Neural Lineage
   biological_scale: CELLULAR
@@ -376,11 +376,11 @@ pathophysiology:
       inference from mouse ESCs to human neurodevelopment is the step being
       made.
   downstream:
-  - target: phenotypes#Global Developmental Delay
+  - target: Global Developmental Delay
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: phenotypes#Abnormal Corpus Callosum Morphology
+  - target: Abnormal Corpus Callosum Morphology
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: phenotypes#Prominent Metopic Ridge
+  - target: Prominent Metopic Ridge
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
 - name: Altered Blood DNA Methylation Signature
   biological_scale: MOLECULAR


matentzn:
All five blocking items and all four suggestions addressed. Nothing disputed. Snippets **41 → 52 verified**.

**Blocking 1 — orphaned methylation node.** You put it in the sharpest form: the node's own description asserted *"This node is downstream of the molecular defect"* while nothing in the graph said so. Prose and structure disagreed.

An edge now runs from `ZNF462 Haploinsufficiency`, marked `INDIRECT_UNKNOWN_INTERMEDIATES`. **I drew it from the haploinsufficiency node rather than from the G9A/GLP node deliberately**, and the description says why: the episignature is human data and the G9A/GLP account is entirely murine, so routing it through that node would assert a connection neither dataset makes. It would have made a tidier graph and a worse claim.

**Blocking 2, 3, 4 — three missing phenotypes.** All quotable from text already cached in this PR. Added **Short Nose** and **Bulbous Nose** (46%, split across two HP terms because HPO has none for the combination), **Hearing Impairment** (distinct from the structural ear anomaly — the functional deficit is what drives the audiology recommendation, and it is one of four prevalences the delineation series cites as its reason for recommending multidisciplinary evaluation), and **Minor Limb Anomalies** at the general level.

**Blocking 5 — penetrance and constraint.** Both were sitting in the cached full text. The penetrance claim is the stronger counselling statement the `inheritance` block was reaching for without having: **complete penetrance with variable expressivity** means an inheriting child *will* be affected and only severity is unpredictable. The gnomAD constraint metric (o/e 0.03, 90% CI 0.01–0.09) is graded `COMPUTATIONAL` — a population-genetic statistic, not a clinical observation — and is the independent quantitative argument for haploinsufficiency alongside the allele spectrum the node previously relied on alone.

**Suggestion 6 — chromatin reader.** Added as complementary rather than redundant: ZNF462 binds H3K9me3 directly and interacts with HP1α, i.e. reader and writer-recruiter on the same mark. Both items `directness: INDIRECT` because they reach the entry second-hand through the delineation paper's discussion, not the primary biochemistry. Your point that this partly blunts the entry's own `HUMAN_MODEL_MISMATCH` concern is right and is why it earned its place. **PBX1–HOXA9 not added** — it appears only in the deep-research report, which carries no PMIDs.

**Suggestion 7 — parental mosaicism.** Added, with the cohort breakdown (17 of 21 de novo; the other four unknown, mosaic or dominant) and the mosaic transmitting mother at 17% alternate allele fraction. It materially qualifies the 95% headline: an apparently de novo variant is not proof a parent carries nothing.

**Suggestion 8 — craniosynostosis band.** Left as `OCCASIONAL`. You didn't block, and the description already states the band is an inference from a jointly quantified pair rather than a sourced figure.

**Suggestion 9 — treatment term.** Rebound to `NCIT:C52006` Gastrostomy with `therapeutic_modality: SURGERY`.

**Checklist row 8 — the narrowing wasn't signalled.** Fixed *in the entry*, not just here. The `notes` now name what is deliberately narrower than the report — hypertelorism, low-set ears, epicanthus, OSA, cryptorchidism, and the specific limb anomalies — and why: none is quotable from a cached primary source, the report carries no PMIDs, and Orphanet's record for this disorder is not among the cached ORPHA files. Naming them makes the narrowing visible instead of looking like an oversight. That was the most transferable item in the review.

---

matentzn:
Thanks — took the non-blocking leftover, because on a closer look it was not just cosmetic.

You were right that the cache stores both sentences as single unwrapped lines. That means the two `explanation` fields saying the quote was clipped "because the sentence wraps in the cached full text" were simply false, and they would have sat in the entry as an untrue account of why a quote looks the way it does. Removing the truncation without removing that sentence would have left the more misleading half behind, so both are gone:

- Hearing loss: snippet now ends `...loss of function variants in ZNF462.`
- Penetrance: snippet now ends `...a father requiring surgery for his ptosis`, and the explanation says what that case actually shows rather than apologising for the clipping.

52/52 snippets still verify. `just validate`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-grading`, `check-title-snippets` and `check-snippet-length` all pass with no baseline edits. History record added for the round.

I know this drops the approval, which I would not normally do to an approved PR — I made an exception here because the defect was a false statement rather than a polish item, and you explicitly called for a follow-up commit.

Item 8 (the `OCCASIONAL` craniosynostosis band being an inference from a snippet that quantifies it together with metopic ridging) I have left as it is. The source genuinely does not separate them, and I would rather the band read as a curator's inference with the reasoning in the description than manufacture a sourced-looking figure.

---

matentzn:
Pushed a fix that pulls this out of the merge queue, deliberately — it was about to merge with a broken pathograph.

`check-causal-targets` was added to `main` after this entry was written, and it fails on 6 causal targets here. I had written them with the `<kind>#<name>` entity-reference grammar in slots that take a **bare name**. That grammar is correct for `attaches_to`, `would_support`, and the `proposed_experiments` readout/perturbation targets — it is wrong for `downstream[].target` and its siblings.

The consequence is worse than a missing arrow, and I had it wrong in my own head until I read the checker. Targets are matched verbatim with no resolution step, so the renderer drew **phantom duplicate nodes** labelled with the literal `phenotypes#...` strings while the real nodes dropped out of the graph. `build_causal_graph` reported 6 orphan targets; it now reports zero.

Nothing else changed. 52/52 snippets still verify, and `just validate`, `check-entity-refs` and `check-duplicate-keys` all pass.

**This affects all five PRs in this run** — 44 targets across the five entries, every one of them mine — so I have pushed the same fix to #10095, #10097, #10098 and #10101. I checked each graph individually rather than trusting the linter: all five now build with `orphan_targets` empty.

I am sorry to cost this its approval and its place in the queue. Merging a fragmented pathograph into `main` seemed clearly worse, and the `prefixed` class is documented as never grandfathered, so it would have failed the gate on the next run regardless.

Worth noting for the gate's own sake: it caught something five rounds of review and every other check missed. The entries validated, the snippets verified, the terms resolved, and the pages rendered — with the mechanism graph broken the whole time.

---

PR #10101, opened 2026-08-29
curate: Add MEND Syndrome (MONDO:0010498, EBP)

### LINKED ISSUE #10092: Curate MEND syndrome (MONDO:0010498)
Curate a dismech entry for **MEND syndrome** ([MONDO:0010498](https://monarchinitiative.org/MONDO:0010498)) — Male EBP Disorder with Neurological defects, the hypomorphic male-viable allelic counterpart of X-linked dominant chondrodysplasia punctata (CDPX2/Conradi-Hunermann-Happle). Partial loss of emopamil-binding protein (sterol Δ8-Δ7 isomerase) blocks the post-squalene cholesterol pathway, so 8(9)-cholestenol and 8-dehydrocholesterol accumulate while cholesterol falls.

Preflight: MONDO:0010498 is a leaf term with one causal gene (`hgnc:3133` EBP) and no existing `kb/` entry or open PR. The related sterol-biosynthesis entries already in the KB (`MSMO1_Deficiency`, `CHILD_Syndrome`, `Greenberg_Dysplasia`, `Chondrodysplasia_Punctata_Tibial-metacarpal_Type`) are different enzymes or a different inheritance mode, so this does not duplicate them. Expected to be a `DISEASE`; the relationship to CDPX2 needs stating explicitly in the entry.

Stub: `stubs/MEND_Syndrome.yaml`

matentzn:
New entry for **MEND syndrome** — Male EBP Disorder with Neurologic defects, the X-linked recessive hypomorphic-*EBP* disorder of post-squalene cholesterol biosynthesis. Closes #10092.

## The entity exists because of a nosological argument, and that argument is the entry

*EBP* variants in males give **three** outcomes, decided by allele severity and mosaicism rather than by variant position:

| Allele | Male outcome |
|---|---|
| **Null** | Lethal in early gestation (heterozygous females get X-linked dominant CDPX2) |
| **Mosaic** (postzygotic) | CDPX2 phenotype in a male — wild-type cells rescue viability, mutant clone gives Blaschko-linear skin and asymmetric skeletal findings |
| **Hypomorphic, constitutional** | **MEND** — uniform partial activity in every cell; survivable, symmetrical, neurologically weighted |

The load-bearing observation is in the 2012 paper that named the entity: the same **p.Y11X nonsense change** had previously occurred de novo in a female with typical CHH syndrome. Allele identity does not determine which disorder results. So interpreting a new male *EBP* variant needs **two** determinations, not one — hypomorphic vs null, and constitutional vs mosaic — and the genetic notes say so.

The practical consequence appears twice, in `diagnosis` and `differential_diagnoses`, because it is easy to get wrong: **plasma sterols do not distinguish MEND from CDPX2 in an affected male.** Both accumulate the same two precursors and the source states there is no correlation between sterol levels and phenotype.

## The mechanism chain stops at biochemistry, deliberately

Three nodes only — hypomorphic variant → reduced isomerase activity → precursor accumulation with cholesterol deficiency — and all four organ phenotypes hang off the third by `INDIRECT_UNKNOWN_INTERMEDIATES`.

Nothing available traces a route from raised precursors to intellectual disability, cataract, stippling or ichthyosis. The toxicity is "potentially toxic" in the source and is not measured. Two observations sharpen the gap rather than filling it: sterol levels don't correlate with phenotype (so it isn't a dose-response), and CDPX2 has the *same* biochemistry with a substantially different phenotype (which in a mosaic disorder points at *which cells* rather than *how much*). Whether the statin rationale is even right depends on which arm — precursor excess or cholesterol deficiency — drives the phenotype. That's a discussion.

## Two evidence items are `NO_EVIDENCE` on purpose

- *"in affected males, biochemical testing will not be of use in differentiating CDPX2 from MEND syndrome"* — cited to mark the **absence** of a usable signal, not to support a claim.
- *"Simvastatin therapy is being evaluated in two males from this family"* — that says a trial was **under way**, not that it worked, and no outcome has been published. Grading it `SUPPORT` would have manufactured a treatment result out of a statement of intent.

Simvastatin is accordingly curated as a mechanistic candidate, with three items covering the three different things actually known: in vitro suppression of cholest-8(9)-enol in patient fibroblasts (`IN_VITRO`/`SUPPORT`), the unreported trial (`NO_EVIDENCE`), and a topical simvastatin-cholesterol ointment case in the **allelic** disorder CDPX2 (`SUPPORT` + `INDIRECT`, explanation saying it shows the approach has been tried in EBP deficiency, not that it helps here).

## A reference-validator constraint I hit, and did not configure my way around

Three snippets from PMID:39754633 failed verification because the cached full text carries inline citation markers — `the four coding exons [1-3] of EBP`, `8 [4]-cholestenol`, `traits [9]` — and **the validator strips bracketed spans before matching** unless `conf/reference_validator_config.yaml` declares them literal. A snippet spanning such a marker fails even when it is a faithful quote.

I shortened the three to stop before the marker, or replaced them with the adjacent sentence carrying the same claim, and each explanation says so. A snippet with a *trailing* marker (`cryptorchidism [6]`) verifies unchanged, so the constraint is about markers **inside** a quoted span. **No config was changed to force a match** — that seemed like exactly the wrong lever to reach for.

## GeneReviews

There is no chapter for MEND. PMID:21634086 (*EBP-Related X-Linked Chondrodysplasia Punctata*) is the closest expert source and is tagged, but it is written for EBP-CDPX, of which **>95% of live-born affected individuals are female**. The two items taken from it carry `directness: INDIRECT` with that reason stated in place rather than being imported as if they described this disease.

PMID:27276700 (the SEM study of the collodion membrane) returns a journal error page instead of content. Nothing is quoted from it; the collodion presentation comes from PMID:39754633. Its cache file is committed so that is checkable.

One Perplexity run, orientation only. No PMID or DOI in its body. Its term validation reported **26 of 53 checked labels naming a different term** — the worst of the five reports in this claim run — clustered in NCIT and of the most misleading kind: `NCIT:C288` offered as *Antiepileptic Agent* is **Azacitidine**; `NCIT:C15382` as *Neurosurgical Procedure* is **Gamma Knife**; `NCIT:C96116` as *Cataract Extraction* is **a plastic container closure**. Every CURIE here was verified independently against OLS.

## Validation

Schema validation passes. **35/35 evidence snippets verified** against cached primary references. Term validation, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass. `just validate-disorders` passes.

---

### DIFF
--- a/history/disorders/MEND_Syndrome/2026-09-03T025811Z-claude-code-f23041.yaml
+++ b/history/disorders/MEND_Syndrome/2026-09-03T025811Z-claude-code-f23041.yaml
@@ -1,0 +1,25 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: MEND_Syndrome
+  path: kb/disorders/MEND_Syndrome.yaml
+session:
+  id: 2026-09-03T025811Z-claude-code-f23041
+  timestamp: '2026-09-03T02:58:11Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10101
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Fix prefixed pathograph targets
+  details: |
+    check-causal-targets, a gate added to main after this entry was written, found 6 causal targets using the <kind>#<name> entity-reference grammar in slots that take a bare name. These are matched verbatim with no resolution step, so the renderer drew phantom duplicate nodes and dropped the real ones from the graph, fragmenting the pathograph while every other gate passed. Stripped the prefixes; build_causal_graph now reports orphan_targets empty. Snippet counts and all other gates unchanged.
--- a/kb/disorders/MEND_Syndrome.yaml
+++ b/kb/disorders/MEND_Syndrome.yaml
@@ -172,7 +172,7 @@ pathophysiology:
       A specific hypomorphic allele segregating with the phenotype in a
       multigenerational family.
   downstream:
-  - target: pathophysiology#Reduced 3-beta-Hydroxysteroid-Delta8,Delta7-Isomerase Activity
+  - target: Reduced 3-beta-Hydroxysteroid-Delta8,Delta7-Isomerase Activity
     causal_link_type: DIRECT
 - name: Reduced 3-beta-Hydroxysteroid-Delta8,Delta7-Isomerase Activity
   biological_scale: MOLECULAR
@@ -245,7 +245,7 @@ pathophysiology:
       is pharmacologically reversible - which is the rationale for the statin
       treatment below.
   downstream:
-  - target: pathophysiology#Sterol Precursor Accumulation and Cholesterol Deficiency
+  - target: Sterol Precursor Accumulation and Cholesterol Deficiency
     causal_link_type: DIRECT
 - name: Sterol Precursor Accumulation and Cholesterol Deficiency
   biological_scale: ORGANISM
@@ -289,13 +289,13 @@ pathophysiology:
       the reference validator strips, so this consequence sentence is quoted
       instead.
   downstream:
-  - target: phenotypes#Intellectual Disability
+  - target: Intellectual Disability
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: phenotypes#Ichthyosis
+  - target: Ichthyosis
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: phenotypes#Cataract
+  - target: Cataract
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
-  - target: phenotypes#Epiphyseal Stippling
+  - target: Epiphyseal Stippling
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
 phenotypes:
 - category: Neurological


matentzn:
The CRITICAL finding and all eight IMPORTANT items are addressed; three suggestions taken, three declined with reasons. Snippets **35 → 54 verified**.

## The CRITICAL one — you're right, and it's the most useful finding of this whole claim run

`X-linked_Chondrodysplasia_Punctata_2.yaml` already held this concept deliberately, with an OPEN discussion asking exactly the question my entry answers. Creating `MEND_Syndrome.yaml` silently overturned that.

**Why the preflight missed it, since the failure generalises.** The duplicate check ran by MONDO ID and by disease label, per the `claim-disease` skill, and both came back clean — *correctly*, because the concept was held in another entry under a different name and surfaced only inside a discussion's prose. **Neither an identifier search nor a label search can find a concept a curator chose to keep unnamed.** Your `rg -il "hypomorphic.*EBP" kb/` would have. That is a blind spot in the preflight as specified, not a slip in executing it, and I've recorded it in both the entry notes and the history record.

All four edits made in this PR:

1. `hypomorphic_male_entity` → `status: RESOLVED`, with a rationale stating the split and naming MONDO:0010498.
2. **The stale premises are corrected in place rather than deleted** — the rationale quotes what it used to say ("MONDO carries only MONDO:0020603", "a single reported family") and says why each is now false, so a reader sees what changed and not only the new conclusion.
3. The CDPX2 `notes` paragraph now points here.
4. This entry's `notes` carry the split, its basis, and the preflight failure, so the reasoning is findable from either side.

**The content conflict is resolved in the direction the evidence supports.** The CDPX2 sentence "without the ichthyosis and stippling of CDPX2" generalised from the single 2014 family. Mild ichthyosis is documented in both patients of the 2025 family, so I corrected the CDPX2 sentence rather than leaving it to contradict this entry. The real distinction is the **pattern** — symmetrical here, Blaschko-linear with follicular atrophoderma there — not the presence.

## IMPORTANT items

**1 (cached-but-uncited PMID:31397093)** — the worst defect in the original. The discussion quoted that paper's variant counts and named its three candidate genes as if they were general knowledge. Now an evidence item on the discussion, plus APOA5/ABCA1/APOB in `genetic:` as `relationship_type: MODIFIER`, each with notes making clear they are one-family candidates with no functional validation.

**2** — added Hypotonia, Hydrocephalus, Cleft palate, Clinodactyly, Strabismus. The hydrocephalus item quotes the sentence that also records the *discordant* finding in the other brother, so single-family variability is visible in the evidence rather than only asserted.

**3** — you're right that the "all missense" claim refuted itself two lines later against p.Y11X. The notes now list all seven tabulated alleles across four familial and eight sporadic patients, and state what actually unites them: **residual activity, not variant type**. A hemizygous male who survives has partial function by construction.

**4** — PMID:24700572 fetched and cited on the CDPX2 differential.

**5** — two GeneReviews items added to `inheritance` (including the "typical but not absolute" male-lethality clause, which is precisely what MEND occupies) and two to symptomatic management. All `directness: INDIRECT`.

**6** — `target_mechanisms` on simvastatin → the sterol node, `INHIBITS`, with the description stating the limit that matters: **if cholesterol deficiency rather than precursor excess drives the phenotype, reducing flux is the wrong direction** — the entry's own open question.

**7** — `progression:` block added. The p.Trp47Arg phase records survival to 43 and the milder-structural/more-behavioural trade-off against p.Trp47Cys at the same residue, which as you spotted partly answers the uncertainty flagged on `Atypical Behavior`.

**8** — `biochemical:` block added, with a `NO_EVIDENCE` item recording that the analytes do not discriminate MEND from CDPX2 in a male.

## Suggestions

**Taken:** stippling downgraded to `OCCASIONAL` with the reasoning in the description; both differentials bound to MONDO terms with `distinguishing_features`.

**Declined, with reasons:** splitting node 3 into precursor-excess and cholesterol-deficiency arms — defensible, but the discussion attaching to a *single* node is what makes the unresolved either/or legible, and the split would create two nodes neither of which has independent evidence; the Tattered mouse — no primary source fetched, and the deep-research report carries no PMIDs so I won't take it on the report's word; regrading the background sentences to `OTHER` — arguable either way and you marked it not wrong.

## One defect I introduced and caught mid-round

Two of my edits appended evidence items in a way that produced a repeated `explanation:` and a repeated `supports:` key. **PyYAML's safe loaders would have kept the last value silently; the ruamel-backed reference validator raised `DuplicateKeyError` and aborted** — exactly the failure mode CLAUDE.md #8623 describes. Both merged rather than deleted; `check-duplicate-keys` is clean on both files.

---

matentzn:

`check-causal-targets` was added to `main` after this entry was written, and it fails here on 6 causal targets. I had written them with the `<kind>#<name>` entity-reference grammar in slots that take a **bare name** — correct for `attaches_to` and `would_support`, wrong for `downstream[].target`.

Targets are matched verbatim with no resolution step, so the renderer drew phantom duplicate nodes labelled with the literal prefixed strings while the real nodes dropped out of the graph. `build_causal_graph` reported 6 orphan targets; it now reports zero, with 8 resolving edges.

That matters more than usual here, because this entry's whole design is that the mechanism chain stops at biochemistry and all four organ phenotypes hang off the third node by `INDIRECT_UNKNOWN_INTERMEDIATES`. Those four edges were the argument, and they were landing on phantom nodes.

Nothing else changed. 54/54 snippets still verify, and `just validate`, `check-entity-refs` and `check-duplicate-keys` pass. The `X-linked_Chondrodysplasia_Punctata_2` edit resolving the lump/split question is untouched.

The same fix went to all five PRs in this run (#10095, #10097, #10098, #10100) — 44 targets in total, all mine.

---

PR #9462, opened 2026-08-25
Re-sync drifted publication titles and turn on the lrv title check

cmungall:
Completes the sequence started in #9367, which documented that `EvidenceItem.reference_title` was invisible to `linkml-reference-validator` and that enabling the check was blocked on **data**, not tooling. This does the data fix, then enables it.

Net effect: **a wrong `reference_title` is now a validation error** instead of passing silently across ~130,000 evidence items.

## Data — 6,125 titles re-synced across 408 files

Every KB title now agrees with its cached publication under lrv's own `normalize_text`: **0 mismatches remaining**, confirmed by running the real validator rather than a replica of it.

One rule fixes all four causes: take the cached title, the same provenance rule the backfill uses.

| Cause | Count |
|---|---|
| ORPHA `(Orphanet structured-database record)` suffix | 4,430 |
| Unicode / ASCII-folded accents (`Guillain-Barre` → `Guillain–Barré`) | ~1,377 |
| Publisher markup preserved verbatim (`&lt;i&gt;ECHS1&lt;/i&gt;`) | 280 |
| Whitespace only | 35 |

197 titles on 189 references were **left untouched** because their cache entry has no title of its own (nearly all `DOI:`, whose cache records are metadata-only). The script reports these rather than guessing.

Drift arrived three separate times while this PR was in review — most legibly `Gilberts_Syndrome.yaml` citing `PMID:21983082` as "inactivation of NF-kB" where the source reads **NF-κB** with the Greek letter. Each instance would have failed the check this PR enables, on `main`, immediately after merge. That rate is the argument for the gate: this is not a one-time cleanup that stays clean.

## What the markup row means, and why it changed

An earlier revision of this PR *stripped* that markup. That was wrong, and it is the one part of this diff worth understanding before approving.

lrv compares against the raw title in the cache frontmatter, and its `normalize_text` maps punctuation to a **space** rather than deleting it. So markup does not disappear under normalization — it becomes words:

| | value | normalizes to |
|---|---|---|
| raw (what lrv compares) | `&lt;italic&gt;TUBA1A&lt;/italic&gt; Variant` | `italic tuba1a italic variant` |
| tidied (what we used to write) | `TUBA1A Variant` | `tuba1a variant` |

Those are different strings, so the tidied form is a value lrv *rejects*. Stripping the tags fixed the appearance and broke the check.

[...]

PR #7147, opened 2026-07-30
Strengthen Bronchiectasis evidence model

cmungall:
## Summary
- add exact evidence to core vicious-vortex mechanisms and causal edges, including an explicit feedback cycle
- wire IL-1beta signaling and the SCNN1A/ENaC susceptibility signal into impaired mucociliary clearance
- model pulmonary exacerbations and fatigue as connected clinical outcomes
- add age-stratified prevalence, biomarker-profile context, and evidence-backed mucoactive therapy
- restore ECTO-bound pollution and smoking exposures with direct human evidence
- consume all previously unused declared bronchiectasis references, including EMBARC, microbiome, biomarker, and resistome studies
- retain the earlier evidence hardening for diagnostics, treatments, phenotypes, and Brensocatib
- append validated review provenance

## Validation
- `just validate kb/disorders/Bronchiectasis.yaml`
- `just validate-terms kb/disorders/Bronchiectasis.yaml`
- `just validate-references kb/disorders/Bronchiectasis.yaml` (26/69 snippets verified; 43 DOI-prefix skips)
- `uv run python -m dismech.qc_plugins kb/disorders/Bronchiectasis.yaml --list-unconnected` (10/10 phenotypes connected; 1/1 causal genes wired)
- graph audit confirms explicit causal cycles through impaired clearance, infection/inflammation, neutrophilic inflammation, protease imbalance, and airway dilation
- `just validate-history history/disorders/Bronchiectasis/2026-07-31T225951Z-codex-7147.yaml`
- `uv run python scripts/check_folded_hyphens.py --check`

## Review reconciliation
- addresses formal review 4832643826: isolated IL-1beta node, acyclic graph, environmental evidence, SCNN1A wiring, mucoactive therapy, fatigue, unused declared references, and pulmonary exacerbations
- the new pollution cache was generated with `just fetch-reference PMID:29772571`

PR #10105, opened 2026-08-29
Fix PR Shepherd routing and merge safety

cmungall:
PR Shepherd was running, but its closing path was split: the LLM could merge or arm auto-merge, while the deterministic sweep excluded drafts and used an unreliable freshness signal. In the latest production run, the deterministic phase scanned 215 PRs and rejected 12 solely because they were drafts, while the LLM phase separately merged two PRs and armed #9462. This change makes LLM tending bounded and non-merging, and puts ordinary closing behind one deterministic controller.

## What changes

- Treats draft state as metadata. Approved drafts are considered by default, marked ready only immediately before the final verification, and restored to draft if the attempt aborts.
- Normalizes verified agent identities, excludes assigned PRs and separately managed `auto/` branches, and gives the LLM a bounded, ranked shortlist. Human-authored `claude/` branches are no longer treated as agent-owned merely because of their name.
- Requires the GitHub Actions-owned `test (3.13)` check on the exact current `main` SHA, then uses GitHub's compare API to prove that SHA is an ancestor of the final PR head.
- Re-reads all PR state after any draft transition, pins the reviewed head SHA, confirms `main` has not moved, and merges at most one PR per run.
- Runs the LLM and closer on separate runners and concurrency lanes. The closer checks out trusted `main`, uses read-only discovery credentials, and receives a narrowly scoped writer only after Python setup.
- Keeps the deterministic closer hourly in every enabled cron profile while retaining four-hour LLM tending under the active slow profile. Every configured profile, including `off`, now has an end-to-end application test.
- Treats a red, pending, or not-yet-observed required check on `main` as a visible fail-closed circuit stop without turning every hourly shepherd run red. API, policy, and write failures still fail the run.
- Makes targeted dispatches report their exact rejection reason, makes dry-run honor the one-merge cap, and retries GitHub's short draft-state propagation lag.
- Removes the unused weekly-compliance classifier; weekly-compliance PRs enter the same review and deterministic closing path as other PRs.
- Moves all six `auto/` lanes' advancing writes—branch push, PR creation, and auto-merge arming—to a late-minted `ai4c-agent` token. This lets required PR checks run and ensures resulting `main` commits trigger CI. Existing requests are disabled and re-armed to replace the legacy actor.
- Pins deterministic generated-artifact approvals to the exact generated commit and pins manual shepherd checkouts to the trusted default branch.

## Validation

- 164 focused policy, controller, workflow, token-boundary, and cron-profile tests pass.
- Ruff, repository pre-commit hooks, shell syntax, `git diff --check`, and actionlint v1.7.7 pass.
- The full non-KB Python run produced 11,321 passes and 42 skips. Its only local failure is the existing macOS case-insensitive-filesystem hardlink test; the changed control-plane tests all passed.
- A live read-only scan covered 218 open PRs. The controller correctly rejected #9523 and #9538 as 265 and 259 commits behind current `main`; it proposed no merge.
- Live branch protection still matches the configured `test (3.13)` context and GitHub Actions App identity.

## Rollout note

I disabled unsafe auto-merge requests on ordinary PR #9462 and on legacy `github-actions`-armed lanes #10104, #10103, #10076, and #9501. The corrected lane workflows will re-arm their own PRs as `ai4c-agent` after this lands. I left #8287 and #7147 unchanged because those requests were enabled by human maintainers; the docs identify those as explicit overrides.

## Remaining throughput limit

Under the slow profile, safe freshness tending is intentionally bounded to one branch every four hours. Updating a batch would dismiss approvals and start CI only for the first merge to make the rest stale again. Higher safe throughput should use a real merge queue with `merge_group` CI rather than weakening the ancestry guard.


cmungall:
The blocking liveness issue is fixed at its source; the exact-`main` health guard remains strict.

- All six `auto/` lanes now mint `ai4c-agent` after their long build, keep checkout credentials unpersisted, and use that App for branch push, PR creation, and auto-merge arming. App pushes start the required PR checks, and App-attributed merges start CI on the resulting `main` SHA.
- Existing auto-merge requests are disabled before being re-enabled, so a legacy `enabledBy: app/github-actions` actor cannot survive. New and existing PR paths both arm; this also fixes the previously missing new-PR arming in the warm-cache and two baseline lanes.
- Legacy `github-actions` PR authors and new `ai4c-agent` authors are both recognized by stale cleanup, deterministic approval, and the generated-PR review guard; human authors with the same branch prefix are not.
- Generated-artifact approval now posts through the review API with `commit_id` pinned to the exact generated head, closing the read/review race.
- The new workflow guard enumerates all six lanes and verifies App-authenticated push/create/merge, unpersisted checkout credentials, both bot identities, both arming paths, and CI path-filter wiring.

The two important operational findings are also fixed:

- Every configured cron profile is now applied end to end in an isolated workflow copy; `medium`, `fast`, `fast-weekend`, and `off` no longer conflict with slow-profile-only assertions.
- A red, pending, or missing required check on current `main` opens the fail-closed circuit and reports prominently without failing every hourly shepherd run. API, policy, and write errors still fail.

Suggestions addressed: dry-run stops after one prospective merge; targeted runs always explain rejection; draft promotion retries GitHub's read-after-write lag; the no-op compliance workflow/classifier was removed; weekly compliance uses the common controller; and the author-identity boundary replacing the `claude/` heuristic is documented. I measured the complete selector again at 12.3 seconds for 218 open PRs, so I did not pre-truncate before exact ownership ranking.

Validation on the rebased head: 164 focused tests, Ruff, pre-commit, composite shell syntax, `git diff --check`, and actionlint v1.7.7 pass. The live controller preview scanned 218 PRs, correctly rejected #9523 and #9538 as 265 and 259 commits behind current `main`, and proposed no merge.

For rollout safety I disabled the legacy `github-actions`-armed requests on #10104, #10103, #10076, and #9501, plus the ordinary-PR bypass on #9462. The corrected lane workflows will re-arm their own PRs after this lands. Human-enabled #8287 and #7147 remain untouched.

PR #10130, opened 2026-08-30
curate: add Immunodeficiency 86 (MONDO:0030448, SPPL2A)

### LINKED ISSUE #10125: Curate immunodeficiency 86 (MONDO:0030448)
Curate a dismech entry for **immunodeficiency 86** ([MONDO:0030448](https://monarchinitiative.org/MONDO:0030448)) — biallelic loss of *SPPL2A*, the intramembrane protease that clears the CD74 N-terminal fragment, causing selective loss of a dendritic-cell subset and susceptibility to mycobacterial disease.

Preflight: MONDO records no descendants and one causal gene (`hgnc:30827` SPPL2A). No existing `kb/` entry mentions SPPL2A or IMD86, and no open PR curates it. The KB does carry other Mendelian susceptibility to mycobacterial disease entries, so the relationship to that concept is worth stating explicitly in the entry rather than leaving implicit.

The mechanism is unusually specific for an immunodeficiency — the pathology is attributed to accumulation of an uncleaved substrate fragment being toxic to one cell type, not to loss of a signalling pathway — so the pathograph should make clear which of those it is asserting.

Stub: `stubs/Immunodeficiency_86.yaml`

caufieldjh:
New entry for **SPPL2A deficiency** — Mendelian susceptibility to mycobacterial disease caused by loss of the lysosomal intramembrane protease that performs the final step in CD74 turnover. Closes #10125.

## The mechanism runs the opposite way round from the rest of MSMD, and the entry is ordered to say so

Most MSMD genes sit in the IFN-gamma / IL-12 / IL-23 circuit, and the disease follows from **losing a signal**. Here nothing in that circuit is mutated. What happens instead:

```
SPPL2A loss  →  CD74 N-terminal fragment not cleaved  →  fragment ACCUMULATES
                                                              ↓ (toxic)
                              CD1c+ cDC2s selectively depleted — the IL-12/IL-23 source
                                                              ↓
              mycobacterium-specific memory TH1* cells starved of the licensing signal
                                                              ↓
                                              no antigen-specific IFN-gamma → BCG disease
```

So it is a **gain-of-toxic-substrate disease presenting as a signalling deficiency**. The IFN-gamma failure is real but secondary — the T cells are starved of the signal, not unable to receive it. Filing SPPL2A next to the IL12RB1-type lesions it phenocopies would inverse the mechanism, so the pathophysiology chain is written in the order above and the `description` states the distinction explicitly.

One word in the founding paper carries a lot of this: the TH1* cells *"selectively"* fail to make IFN-gamma to **mycobacterial antigens**. A primary T-cell lesion would not be antigen-selective.

## The epistasis control is curated as its own model

The claim "the accumulated fragment is the toxic agent" is not obviously testable — protease loss and substrate accumulation are the same event in a patient. In mice they can be separated, and were:

**`Sppl2a-/- CD74-/-` double knockout → `relationship: RESCUES`.** Deleting the substrate rescues the depleted dendritic cells. That is the difference between a correlation and a cause, and it is why this entry attributes the pathology to the fragment rather than to the missing enzyme.

## A negative model result, localised

`FAILS_TO_RECAPITULATE` on the humoral arm. The `Sppl2a` null mouse has a striking B-cell phenotype — lacks mature B-cell subsets, phenocopies BAFF deficiency. The one human patient with reported numbers had hypogammaglobulinaemia with **normal B-cell counts**. Those are different lesions: numerical loss versus functional deficit with preserved cells.

What makes this worth recording rather than just noting is that the divergence is **localised**. Human SPPL2a-null B-cell lines accumulate the CD74 NTF *comparably to mouse B cells* (PMID:25035924), so the biochemistry is conserved and the species difference sits downstream of the substrate accumulation, not at it.

It also has a practical edge: SPPL2a inhibition has been proposed as a way to target B cells in autoimmunity, and that proposal rests on the murine phenotype. The `limitations` field records that the human side of the comparison is one patient, so the divergence is reported, not declared settled.

## Other grading calls

- The murine dendritic-cell cytokine-shift finding (more IL-1β, less IL-10 and IFN-β; Dectin-1 redistributed to endosomes) is graded `PROVISIONAL`, and its second evidence item quotes the authors' *"We propose that…"* as a proposal rather than paraphrasing it into a finding.
- Hypogammaglobulinemia is labelled a single-patient finding in both the phenotype `description` and the entry `notes`.
- The entry is kept out of an MSMD grouping deliberately — it reaches the MSMD endpoint by a different route (loss of a cell type, not of a signal). The `notes` say this, and say that the clinical phenotype is nonetheless MSMD.

## Provenance

Sources found by **PubMed search, not a deep-research provider**. Every PMID was real before it was cited; all ontology terms verified against OLS before binding.

## Validation

```
just validate                 # schema clean, terms clean, 28/28 snippets verified
```

Cache diff is the four terms this entry introduced (`CL:0002399`, `HP:0020086`, `hgnc:30227`, `MONDO:0030448`) and nothing else. Stub deleted.

---

### DIFF
--- a/history/disorders/Immunodeficiency_86/2026-08-30T033927Z-claude-code-da431f.yaml
+++ b/history/disorders/Immunodeficiency_86/2026-08-30T033927Z-claude-code-da431f.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Immunodeficiency_86
+  path: kb/disorders/Immunodeficiency_86.yaml
+session:
+  id: 2026-08-30T033927Z-claude-code-da431f
+  timestamp: '2026-08-30T03:39:27Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10130
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  - discussions
+  - treatments
+  - references
+  summary: 'Review round 1: add the TNF-alpha ICD route to IL-12 and soften the headline
+    claim'
+  details: |
+    Addressed the round-1 review on PR 10130. The blocking finding was that the entry's organising claim - that SPPL2A sits outside the IFN-gamma/IL-12/IL-23 circuit and nothing in that circuit is mutated - is contradicted by PMID:16829952 (Friedmann 2006), which was uncited.
+
+    Verified against the newly fetched cache: SPPL2a and SPPL2b catalyse intramembrane cleavage of TNF-alpha, and the released intracellular domain triggers IL-12 expression in activated human dendritic cells. That is a direct, cell-intrinsic route from the protease to the cytokine whose absence defines MSMD, requiring no cell death and no CD74 fragment.
+
+    Rather than delete the distinction, which is still real, the claim was softened: most MSMD genes encode components of the circuit itself, SPPL2A is not one of those, but it does have one direct link into the IL-12 arm. The description, the notes' lump/split rationale, and the discussion all now say this.
+
+    A new pathophysiology node, Loss of TNF-alpha Intracellular Domain Signalling to IL-12, was added off the trigger node rather than off the depletion node, since the route bypasses CD74 entirely. Graded PROVISIONAL with two caveats stated in the node description: the source characterises SPPL2a and SPPL2b jointly and never isolates SPPL2a, and the work is in human dendritic cells in vitro rather than in patient cells.
+
+    The dc_loss_versus_dc_dysfunction discussion was rewritten from a two-way to a three-way question, noting that the TNF-alpha route is the only one of the two functional accounts with human data behind it, and the proposed experiment now assays TNF-alpha ICD release alongside per-cell cytokine output so a deficit can be attributed rather than left ambiguous.
+
+    Non-blocking items taken: NCIT:C280 Antitubercular Agent as therapeutic_agent, and a top-level references block.
+
+    Not taken: PMID:36630059 was deliberately not cited. The reviewer flagged probable cohort overlap with Kong 2018's Moroccan kindreds via shared authors, and citing it as independent additional patients without resolving that would inflate the patient census.
+
+    Validation: 30/30 snippets verified (was 28/28; two new items from the new reference), schema and terms clean, all offline gates clean.
--- a/kb/disorders/Immunodeficiency_86.yaml
+++ b/kb/disorders/Immunodeficiency_86.yaml
@@ -14,14 +14,22 @@ description: >-
 
   The disease is mechanistically unusual among the inborn errors of immunity,
   and the shape of the mechanism is the reason to curate it carefully. Most MSMD
-  genes sit in the IFN-gamma / IL-12 / IL-23 signalling circuit, and the disease
-  follows from losing a signal. Here nothing in that circuit is mutated. What
+  genes encode components of the IFN-gamma / IL-12 / IL-23 signalling circuit
+  itself - a receptor, a chain, a transcription factor - and the disease follows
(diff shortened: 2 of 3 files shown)


caufieldjh:
The Friedmann finding is incorporated, and you were right that it bears on the entry's headline claim rather than being an additional citation. **30/30 snippets verified.**

## What the paper changes

I fetched PMID:16829952 and read it before acting. It says what you quote:

> "The two proteases promoted the release of the TNFalpha intracellular domain, which in turn triggers expression of the pro-inflammatory cytokine interleukin-12 by activated human dendritic cells."

So SPPL2a has a **direct route to IL-12 that requires no dendritic cell to die and no CD74 fragment to accumulate**. My "nothing in that circuit is mutated" was overstated.

I took your suggested framing — soften rather than delete, because the distinction is still real. The claim is now that most MSMD genes *encode components of the circuit itself* (a receptor, a chain, a transcription factor) and SPPL2A is not one of those, followed immediately by the TNF-alpha link stated plainly. That is defensible where the absolute version was not. Same correction applied to the `notes` lump/split rationale, which had leaned on the same over-strong claim.

## Where the node went

**Off `Biallelic SPPL2A Loss of Function`, not off the depletion node** — your point about placement was the useful half of the finding. The route bypasses CD74 entirely, so hanging it downstream of depletion would have misrepresented it as a consequence of cell loss, which is precisely what it is not.

Graded `PROVISIONAL`, with both caveats you named written into the node `description` rather than left implicit: the source characterises SPPL2a and SPPL2b jointly and never isolates SPPL2a, so attributing this to SPPL2A specifically is an inference; and it is human DCs in vitro, with no patient shown to have a TNF-alpha-ICD defect.

## The discussion this actually resolves

You spotted that the entry already asked this question and answered the "function" side entirely from murine data. `sppl2a_dc_loss_versus_dc_dysfunction` is now a three-way question, and it notes the thing that matters for weighing them: **the TNF-alpha route is the only one of the two functional accounts with human data behind it**, even though the murine PRR work is more recent and more specific to mycobacteria.

The proposed experiment now assays TNF-alpha ICD release alongside per-cell cytokine output, so a per-cell IL-12 deficit can be attributed to one route or the other instead of being left ambiguous between them. That is a better experiment than the one I proposed originally.

## Non-blocking

- **`NCIT:C280` Antitubercular Agent** added as `therapeutic_agent`. You were right that it validates offline with no research needed.
- **Top-level `references:` block** added (note the slot is `title`, not `reference_title`, inside that block — schema validation caught me).
- **Low CD8 counts** — left out. Single patient, and the entry already flags the same case report's B-cell numbers as single-patient; adding a second single-patient immunophenotype detail would dilute rather than strengthen.

## PMID:36630059 — deliberately not cited

Thank you for the warning, and I have acted on it rather than around it. I have **not** cited Errami 2023. Your shared-author observation (Bousfiha, Ailal, Benhsaien) makes cohort overlap with Kong 2018's Moroccan kindreds likely, and citing it as independent additional patients would inflate the census. The census claim stays as it was — quoted from the 2024 case report rather than asserted by the entry — which is the form that survives the ambiguity.

## On the deep-research artifact

Same answer as on #10128 and #10131, and I want to be straight about the cost rather than just restating the position. I used PubMed and Europe PMC search instead of a DR provider, deliberately, after the previous round spent real effort catching invented CURIEs and one fabricated reference title out of generated reports.

Your calibration figure (29 of the 30 most recent entries carry one) is a fair challenge, and the Friedmann omission is exactly the failure mode you describe — a 2006 paper that a literature sweep would surface and a targeted search for "SPPL2A immunodeficiency" does not, because it predates the disease. That is a genuine argument for the DR step that I had not weighted properly. I still would not run one retrospectively to decorate a PR whose content is now verified against primary text, but I will run one before the next entry rather than after.

## Validation

```
just validate      # schema clean, terms clean, 30/30 snippets verified
```

One new reference cached via `just fetch-reference`. A second history record covers this round.

---

PR #10131, opened 2026-08-30
curate: add Optic Atrophy 13 with Retinal and Foveal Abnormalities (MONDO:0008135, SSBP1)

### LINKED ISSUE #10126: Curate optic atrophy 13 with retinal and foveal abnormalities (MONDO:0008135)
Curate a dismech entry for **optic atrophy 13 with retinal and foveal abnormalities** ([MONDO:0008135](https://monarchinitiative.org/MONDO:0008135)) — dominant *SSBP1* variants affecting the mitochondrial single-stranded DNA-binding protein, causing optic atrophy with a distinctive foveopathy, and in some families a wider syndrome including sensorineural hearing loss and nephropathy.

Preflight: MONDO records no descendants and one causal gene (`hgnc:11317` SSBP1). No existing `kb/` entry mentions SSBP1 or OPA13, and no open PR curates it.

Two things to settle during curation: whether the reported multisystem presentations belong here as `has_subtypes` or are the same entity with variable expressivity; and whether the mechanism is mtDNA depletion, deletion, or a dominant-negative effect on the SSBP1 tetramer — the literature does not treat these as settled, so the entry should say which claim it is making and grade the confidence accordingly.

Stub: `stubs/Optic_Atrophy_13_With_Retinal_And_Foveal_Abnormalities.yaml`

caufieldjh:
New entry for **SSBP1-related dominant optic atrophy** — heterozygous missense variants in the mitochondrial single-stranded DNA-binding protein. Closes #10126.

## The three candidate mechanisms turn out to be one chain

The claim issue asked whether this is mtDNA depletion, mtDNA deletion, or a dominant-negative effect on the SSBP1 tetramer, and treated them as alternatives. The published evidence says they are successive steps:

```
missense in the ssDNA-binding basic patch
        ↓  destabilises dimer/tetramer assembly    ← the dominant-negative step
        ↓  impairs mtDNA replication
        ↓  mtDNA depletion
```

The pathophysiology is written as that chain, and the **dominant-negative claim is attached to the multimer-assembly node** rather than to the disease as a whole — it describes one step, and putting it on the disease would blur what is actually being asserted.

## Why the zebrafish earns its place

The dominant-negative reading rests on two independent lines, and the second is the interesting one:

1. Structural analysis plus patient fibroblasts — R38Q destabilises dimer/tetramer formation.
2. **Mutant mRNA phenocopies knockdown** of the endogenous transcript in zebrafish.

Item 2 is the functional signature that separates interference from insufficiency, and it is very hard to establish in patient cells that carry only one variant copy. So the model contributes something the human data cannot, rather than corroborating. It is curated as two `modeled_mechanisms` links on one model — knockdown → the RGC node, mutant-mRNA → the assembly node — because the two arms support different claims. Its `limitations` note the antisense off-target liability and that injected mRNA is not a physiological dose, so the experiment shows the mutant protein *can* interfere, not that interference happens at endogenous levels.

## The foveopathy is developmental, and that is an argued call

Curated as maldevelopment rather than degeneration, on two observations the source makes explicitly:

- it looks **the same at every age**, and
- it has **no measurable effect on visual acuity** (0.54 vs 0.50 decimal, not significant).

A progressive photoreceptor process should worsen with age and cost acuity. This does neither. So it is a **diagnostic marker** distinguishing SSBP1 from the far commoner OPA1 disease, not a component of the visual disability — and the `differential_diagnoses` block points out the practical consequence: it is worth looking for on OCT *precisely because* it costs nothing in acuity and so will never show up in the visual history.

The late-onset rod-cone dystrophy is kept as a separate node, and an `OPEN_QUESTION` asks whether the two are one photoreceptor process on different timescales or independent — which matters because it decides whether the foveopathy is prognostic as well as diagnostic.

## Counselling facts that would be easy to lose

- **Incomplete penetrance** — 3 of 27 carriers of the recurrent p.Arg38Gln were asymptomatic. Curated as its own `Inheritance` block so an unaffected carrier reads as expected rather than as evidence against pathogenicity.
- **Maternal germline mosaicism** in an unaffected transmitting mother. Recorded in the genetic `notes` because it means an apparently de novo variant in this gene *cannot* be counselled as carrying negligible recurrence risk.

## A knowledge gap the entry declines to paper over

SSBP1 is not retina-specific and mtDNA replication is not a retina-specific process, yet the phenotype is largely eye-restricted — while the founding paper states that mtDNA replication defects cause depletion syndromes with *severe neurological* disease. The usual explanation (retinal ganglion cells are selectively vulnerable via the unmyelinated prelaminar axon) is a plausible account borrowed from OPA1 and LHON, not a tested one here: nobody has measured mtDNA copy number in patient retinal ganglion cells or compared affected with unaffected tissue in one individual. The `KNOWLEDGE_GAP` says so and proposes the cross-tissue measurement.

## Two validator constraints, handled without touching config

- One snippet spanned bracketed variant annotations (`[p.(Arg38Gln)]`), which the reference validator strips before matching. **Shortened the quote** to stop before them, with the three protein changes carried in the node `description` instead — and the `explanation` records why the quote ends where it does.
- `check-snippet-length` flagged a three-word snippet (`induces mtDNA depletion`) as non-propositional. **Extended it** to the full clause.

Neither was resolved by editing `conf/reference_validator_config.yaml`.

## Provenance

Sources found by **PubMed search, not a deep-research provider**; all ontology terms verified against OLS before binding.

## Validation

```
just validate                 # schema clean, terms clean, 30/30 snippets verified
```

Cache diff is the five terms this entry introduced. Stub deleted.

---


caufieldjh:
All six blocking items addressed, plus three of the four non-blocking suggestions. **44/44 snippets verified**, up from 30. Everything came from the four references already committed.

Item 2 was a real error in the entry, not a gap, so I want to be explicit about it.

## The "largely eye-restricted" framing was wrong

I checked your cited lines against the caches rather than taking them on trust, and they say what you say:

- `PMID_31298765.md:109` — *"In addition, hypothyroidism (subjects 4-12, 4-14, 4-16 and 4-18) and a renal disorder, including renal failure (subjects 3-8, 4-3 and 4-18), were documented."*
- `PMID_34548540.md:106` — *"the p.Arg38Gln variant only generated ocular symptoms, contrasting with the syndromic clinical presentations occurring with other SSBP1 missense variants"*

So ocular restriction is a **property of p.Arg38Gln**, not of the disease — and this entry curates p.Arg107Gln as well, which makes the extra-ocular features in scope. I had generalised from the best-characterised allele to the disease, which is exactly the error the entry's own `notes` warn curators against making from the mouse literature.

Corrected in the description, the `notes` (which now record what the earlier version claimed), the mtDNA Depletion node, and the discussion. Hypothyroidism and renal insufficiency are now curated phenotypes.

## The knowledge gap had a proposed answer I had missed

`PMID_34548540.md:106` also carries a mechanism I treated as absent: severity may be *"proportional to the residual mtDNA levels"*, with variants differing in *"consequences on tetramer abundance and activity, which would lead to a gradient of mtDNA depletion"* against a per-organ threshold.

That is a much better discussion than the one I wrote. It reframes from "why is this eye-restricted, nobody knows" to a genuine two-horse question — dosage gradient versus retinal-ganglion-cell selective vulnerability — where **the two accounts make different predictions and one measurement separates them**. The proposed experiment now compares alleles, not just tissues. It also connects the clinical gradient straight back to the tetramer-assembly node the entry already curates, which is the kind of link a knowledge gap should have.

## Everything else

| Item | Done |
|---|---|
| 1. Retinal pigmentary changes, vessel attenuation | Added (65% of 60), plus optic nerve hypoplasia |
| 3. Phenotype frequencies | optic atrophy 95%, foveopathy 57.8%, rod-cone 29.3% |
| 5. `diagnosis:` | OCT + wide-field autofluorescence protocol, and the 11%-asymptomatic counselling point |
| Prevalence | `CASES_IN_LITERATURE`, 60 published patients |
| Zebrafish fidelity | Downgraded to `PARTIALLY_RECAPITULATES` |

The frequency correction matters more than it looks: rod-cone dystrophy at **29.3% of 60** is far commoner than the 2-of-27 my entry implied from the single-variant series. "Few SSBP1 patients are likely to develop a RCD in adulthood" and 29.3% are different prognoses, and the entry was carrying the wrong one.

On the zebrafish downgrade — you were right, and the reason is worth stating in the entry rather than just changing the enum: the model reproduces involvement of the right cell type by the wrong route, differentiation failure in an embryo against degeneration of a differentiated cell in a child. That is now in the `description`.

## One suggestion I could not implement as stated

**OMIM:165510 has nowhere to go in `mappings`.** `DiseaseMappings` has exactly four slots — `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings` — and no `omim_mappings`. I tried it and schema validation rejected it.

Recorded as an `external_assertions` entry instead, following the pattern in `15q11q13_Microduplication_Syndrome.yaml:30-37`, with a note in the description saying why it is there rather than in `mappings`. If an `omim_mappings` slot would be generally useful that is a schema issue rather than something to solve per-entry.

## On items 4 and 6

**Del Dotto and Gustafson** — I have not added them. Both are named in `PMID_34548540` Table 1 but neither is cached in this PR, and fetching two new references to re-verify against would widen a PR that already grew substantially this round. Happy to do it as a follow-up if you would rather it be here.

**No deep-research artifact** — same answer as on #10128: I used PubMed and Europe PMC search rather than a DR provider, deliberately, after the previous round spent significant effort catching invented CURIEs and a fabricated reference title out of generated reports. Your point that the DR requirement exists to catch exactly the gaps you found is fair and I am not dismissing it — but the remedy here was reading the cached full text properly, which is what a DR report would have pointed me back at.

## Validation

```
just validate      # schema clean, terms clean, 44/44 snippets verified
```

Cache diff is the new HP terms only. A second history record covers this round and records the correction rather than presenting the entry as always having been right.

---

Issue #2700, opened 2026-05-14
Minor usability issues with Curation Priority Dashboard


nlharris:
On https://dismech.monarchinitiative.org/dashboard/priority.html, there are a few minor things that don't work as I'd expect.

One is that there are places where the cursor changes to a pointing finger, suggesting that the thing I'm over is clickable, but it is not. This happens in the Category View and also the Curation Priority Dashboard, which additionally misleading suggests "Click any row to inspect the current score contribution list and the fixed heuristics behind its action label." There doesn't seem to be a way to click a row; only the MONDO IDs are clickable.

Another minor but slightly annoying thing on that page is that I can't mouse-select the disease names (in the Label column) to copy them to paste elsewhere.

nlharris:
**Correcting myself: two of the three reports here are NOT fixed.** I said earlier in this session that they "no longer reproduce". That was wrong, and the maintainer has confirmed both still fail in a real browser:

- **Rows are not clickable.** Only the MONDO IDs are. The "Click any row to inspect the current score contribution list" text is still making a promise the page doesn't keep.
- **Label text still can't be mouse-selected.** It highlights during the drag and un-highlights on mouse release.

### How I got it wrong

My verification tested things that look like the user's complaint but aren't:

- For text selection I checked `getComputedStyle(el).userSelect === 'auto'` and then built a `Range` programmatically and read `window.getSelection().toString()`. Both passed. Neither involves a mouse. A programmatic Range is not destroyed by a re-render; a user's drag-selection is.
- For row clicking I asserted that `.selected-row` moved to the id I clicked. It did. But "internal state changed" is not "the user sees something happen", and the user's complaint was always about the latter.

Two green checks on the wrong questions, reported as if they settled the right ones.

### What I still can't do

I re-tested with real `mouse.down` / `mouse.move` / `mouse.up` across the Label cell in headless Chromium against the local `dashboard/priority.html`, and the selection **survived** mouseup; a click on a non-link cell also moved `.selected-row`. So I cannot reproduce either symptom in that environment.

That is a fact about my test setup, not about the page — headless Chromium is a poor proxy for real pointer and selection behaviour, and I'm loading a local file rather than the deployed site. **The direct observation in a real browser is the better evidence, and it wins.** I'm recording the non-reproduction only so nobody wastes time assuming a headless check would have caught this.

### A hypothesis worth testing, clearly labelled as one

The click handler ends in a full `render()`, which rebuilds the entire table body:

```js
[...]

PR #10135, opened 2026-08-30
Require CURIEs to be looked up, not recalled; scope the priority-dashboard row cursor (#2700)

nlharris:
Two unrelated changes, on one branch because this run is pinned to a single designated branch. **Happy to split them** if you'd rather review separately — say the word.

---

## 1. `CLAUDE.md`: a sourcing rule for ontology identifiers

The Ontology Term Contract said which term to choose and that `term.label` must match the canonical label exactly. It never said where the string is supposed to come from. That is the gap four fabricated NCIT identifiers went through in a single curation session (#9928, #9937):

| Written as | Actually is |
|---|---|
| `NCIT:C171192` "Chromosomal Microarray Analysis" | Electronic File Size |
| `NCIT:C16512` "Fluorescence In Situ Hybridization" | DNA-directed DNA Polymerase |
| `NCIT:C38031` "Auditory Brainstem Response Test" | Tearing |
| `NCIT:C101293` "Whole Exome Sequencing" | Next Generation Sequencing |

All four were mine. All four were caught by term validation before reaching `main`. (All four labels above were re-verified with `runoak` before being written into the rule.)

The point worth writing down is why the rule can't be replaced by being more careful. Producing a well-formed CURIE with a label that describes the right concept gives no internal signal that it was produced rather than remembered — it feels like recall. So care doesn't help; only a lookup separates the two. `.claude/skills/dismech-terms` shows *how* to run `runoak`, but nothing in the repo said you must.

Also recorded: severity in term validation tracks dynamic-enum membership rather than correctness, so a fabricated CURIE landing inside its enum surfaces as a `WARN` and one outside it as an `ERROR` — the softer signal on the equally wrong binding. Both still fail the recipe; I tested that rather than assuming.

And the same failure shape reaches `reference_title` — three titles composed from having just read the abstracts reached a commit in the same session, one at 0.41 similarity. Noted with a pointer to `dismech-references` rather than duplicating that workflow.

## 2. Priority dashboard: stop promising a click that isn't handled (part of #2700)

[#2700](https://github.com/monarch-initiative/dismech/issues/2700) reports three things. **This PR fixes one of them.**

The one it fixes: a pointing-finger cursor over rows that aren't clickable, in the Category View. It's a scope mismatch between the CSS and the JS. The click handler selects `tr[data-mondo-id]`; the cursor rule was an unscoped `tbody tr`, so every table on the page got the affordance — including the three Category View rows, which have no ids and no handler. Both the cursor and hover rules are now scoped to `tr[data-mondo-id]`.

Verified by rendering the generated page in Chromium before and after: rows with a pointer cursor and no click handler went **3 → 0**, and clicking a data row still moves the selection.

> ### ⚠️ Correction to an earlier version of this description
>
> This PR previously claimed the other two reports in #2700 "no longer reproduce" and that the issue would be "fully addressed once this merges". **Both claims were wrong** and have been retracted on the issue.
>
> Rows are still not clickable (only the MONDO IDs are), and Label text still cannot be mouse-selected — it un-highlights on mouse release. I had verified those with a computed-style check and a programmatic `Range`, neither of which involves a mouse, and by asserting that internal selection state changed rather than that anything visible happened.
>
> **#2700 must stay open when this merges.** See [the correction comment](https://github.com/monarch-initiative/dismech/issues/2700#issuecomment-5467026427) for the likely mechanism (the click handler calls a full `render()`, rebuilding the DOM nodes a drag-selection is anchored to) and why I have not attempted that fix blind.

Only the generator is changed — `dashboard/*.html` is derived and regenerated by the page-build workflow.

---


nlharris:
/review

---

PR #10136, opened 2026-08-30
Resolve the evidence items #7439 left contradicting themselves

cmungall:
## The problem

Retiring `PARTIAL` in #10003 turned every item carrying it into `SUPPORT`. For most that was right. But a subset had explanations written specifically to justify *not* grading them `SUPPORT`, so the migration left the entry asserting one thing in a field and the opposite in the prose beside it:

```yaml
supports: SUPPORT
explanation: >-
  Shows that severity in Otx2-heterozygous mice depends on genetic
  background; PARTIAL because this does not establish a human modifier.
```

**211 items across 166 files** match that shape. Reading them apart gives three groups, and only two are defects.

## 97 items get `directness: INDIRECT`

The curator had already written down why the grade was not plain `SUPPORT`, and in these it is an inferential gap:

| what was quoted | what the claim asserts |
|---|---|
| a mouse deletion | a human mechanism |
| a transcript's presence | a functional role |
| a class-level statement | one member of the class |
| co-occurrence | a causal direction |
| a management recommendation | that a phenotype occurs |

`SUPPORT` + `INDIRECT` says exactly what those explanations say, so the item stops contradicting itself.

This is **transcription of a judgment the curator made and wrote down**, not a new appraisal — which is why it does not reopen #10003's decision not to backfill `directness`. Every one is justifiable by quoting the curator's own sentence.

## 2 items are mis-graded, and are regraded by hand

**`Congenital_Adrenal_Hyperplasia` diagnosis[1] → `NO_EVIDENCE`.** The entry claims CYP21A2 testing supports diagnosis, genotype-phenotype interpretation, counselling and recurrence-risk assessment. The quoted sentence says only that molecular diagnosis of 21OHD is complex and involves ongoing challenges. It does not bear on any of the four. The reference is kept — it is the right paper for the entry, and it is the quote that needs replacing, not the citation.

**`Infective_Endocarditis` treatments[3].target_mechanisms[0] → `REFUTE`.** The link asserts prophylaxis works by suppressing procedural bacteraemia; the quote says postprocedural bacteraemia *is not a good surrogate endpoint for IE*, which cuts against the mechanism the edge rests on.

The same sentence stays `SUPPORT` in the discussion block below, where the claim is that the controversy exists. One quote bearing on two different claims is what `supports` being claim-relative means, and why `check-snippet-grading` gates `evidence_source` but not `supports`. I checked the HPOA export before committing, since `REFUTE` maps to a `NOT` qualifier: this is a treatment link rather than a phenotype, so no `NOT` row is produced and the entry's row count is unchanged at 9.

## 112 items are left alone, deliberately

> "An uncontrolled n-of-1 observation cannot establish efficacy"
> "five participants cannot establish safety in patients with abnormal immune function"

These are statements about **strength**, and the quote is squarely about the claim. They are already correctly `SUPPORT`, and there is no strength slot by design (design decisions §12). Using `directness` to record a sample-size caveat would quietly turn it into the strength grade it was defined not to be.

**13 of these were caught only by reading**, after an automated pass had wrongly marked them `INDIRECT`. The clearest is `AA_Amyloidosis`: the cited series directly reports outcomes on the therapy it is attached to (46.7% progressed, 33.3% stabilised), and the treatment's own description already says "results are more mixed than with IL-6 blockade". Nothing adjacent about it — the caveat is *n*, not kind. Worth stating plainly, because those 13 would have validated cleanly and read plausibly; only reading the snippet against the claim caught them.

## The test, now written down

`docs/explanation/evidence-model.md` gains the rule that made this pass decidable:

> Is the quoted text about the *same thing* the claim asserts, or about something *adjacent*, from which the claim follows only by an inference step?

An adjacent quote is `INDIRECT`; a thin one is not. The two can co-occur — a single-animal study quoted for a human claim is both — but only the first is recordable today, and the second belongs in the `explanation`.

## Why this carries the `scope-override` label

The curation-scope guard fires because the PR touches `kb/` and one file outside the curation allowlist:

```
docs/explanation/evidence-model.md
```

That file is the rule this change applies. Splitting it out would land 97 `directness` assignments with the deciding test documented nowhere, so a reviewer could not check them and the next pass could not be consistent with them. They belong in the same change. Nothing else outside `kb/` and `history/` is touched, and the label was applied before the guard step ran, so no CI cycle was spent on it.

## What this does not do

The ~3,500 remaining items whose explanations still read "Graded PARTIAL because…" without contradicting their value are untouched. They are stale wording, not defects, and the worklist TSV is the vehicle for them. This PR is only the subset where the entry disagreed with itself.

## Validation

| Check | Result |
|---|---|
| `linkml-validate -C Disease` on all 85 changed entries | No issues found |
| `validate-history` on both new records | No issues found |
| Retired enum values in `kb/**/*.yaml` | 0 |
| HPOA export on the regraded entries | no `NOT` rows, row counts unchanged |
| `check-duplicate-keys` | OK, 4,251 files |
| `check-entity-refs` | OK, 2,771 files |
| `check-folded-hyphens` / `check-snippet-grading` / `check-title-snippets` / `check-empty-snippets` / `check-snippet-length` | no new findings |
| `test_data.py` + `test_hpoa_export` + `test_sepio_export` + `test_render_evidence_directness` | 7,982 passed |
| History tests | 29 passed |

No baseline was updated. The diff is 97 added `directness:` lines plus two hand-edited evidence items; nothing else in `kb/` changed.

https://claude.ai/code/session_01VUmMdcPTa7ZqwRV1NKwMbz

### DIFF
--- a/kb/disorders/Heparin-Induced_Thrombocytopenia.yaml
+++ b/kb/disorders/Heparin-Induced_Thrombocytopenia.yaml
@@ -1397,9 +1397,12 @@ treatments:
       anticoagulants (argatroban, bivalirudin, danaparoid, fondaparinux, direct
       oral anticoagulants) for treatment of acute HIT.
     explanation: >-
-      Recorded as INDIRECT because the endorsement is explicitly conditional and
-      does not establish these agents as preferred. Evidence source is OTHER
-      because this is a guideline.
+      INDIRECT because the guideline names the drug class - "direct oral
+      anticoagulants" inside a list of non-heparin options - while this entry is
+      about rivaroxaban and apixaban specifically, so the claim follows from the
+      quote only by taking the class down to its members. That the recommendation
+      is conditional is worth knowing but is a strength caveat, not the reason for
+      this grading. Evidence source is OTHER because this is a guideline.
 
 - name: Therapeutic Plasma Exchange
   description: >-
--- a/kb/disorders/Nipah_Virus_Disease.yaml
+++ b/kb/disorders/Nipah_Virus_Disease.yaml
@@ -897,9 +897,11 @@ treatments:
     evidence_source: OTHER
     snippet: "During the 2023 outbreak, remdesivir was recommended on compassionate grounds for early use before the onset of encephalitis"
     explanation: |
-      Documents compassionate-use recommendation and the timing rationale. INDIRECT
-      because compassionate use in an uncontrolled outbreak does not establish
-      clinical efficacy.
+      Documents the compassionate-use recommendation and the timing rationale.
+      INDIRECT because the quote reports what was recommended rather than what the
+      drug did, so it reaches this treatment entry through the recommendation
+      rather than through an outcome. That the setting was uncontrolled is also
+      true, but that is a strength caveat rather than the reason for this grading.
   - reference: PMID:39700307
     reference_title: "Pandemic potential of the Nipah virus and public health strategies adopted during outbreaks: Lessons from Kerala, India."
     supports: SUPPORT


cmungall:
All four addressed in `d0f790d1a`. I verified each before acting; both 🟡 items were right, and so was 🔵 3.

## 🟡 1 — CAH: you're right, and the miss is worse than you framed it

I verified the sentence: exactly one occurrence in `references_cache/PMID_39911519.md`, `content_type: full_text_xml`, and it is the paper's own concluding assessment. Swapped in, `supports` back to `SUPPORT`, 39/39 snippets in the file verify with no cache churn.

The part worth naming: my explanation said *"a better sentence may exist in the same paper"* — I asserted that without opening the cache. Had I looked, I'd have found it in one grep. Leaving a diagnosis claim unevidenced on the strength of an unchecked guess is the same failure mode as the broken HPOA filter on #10083.

## 🟡 2 — the 97 explanations, and a collision worth knowing about

You're right that the scoping argument covered the ~3,500 untouched items but not these, which are the exact blocks this PR edits.

Rewritten — but **scoped per item by line range, not per file**, which turned out to matter. Both `CDH2-Related_ACOG_Syndrome.yaml` and `Marfan_Syndrome.yaml` contain a *second* item using the identical "PARTIAL rather than SUPPORT" phrasing that this PR deliberately leaves alone:

```
CDH2:137   PARTIAL rather than SUPPORT because it upholds autosomal dominant…   ← untouched item
CDH2:295   Graded PARTIAL rather than SUPPORT because the paper states…         ← in the tranche
Marfan:1517 …PARTIAL rather than SUPPORT because the review states…             ← in the tranche
Marfan:1588 PARTIAL rather than SUPPORT because safety is not aortoprotection…  ← untouched item
```

A file-wide `sed` would have rewritten prose on items outside the tranche and nothing would have caught it — the diff would have looked like more of the same edit. An assertion on the occurrence count is what surfaced it. Both untouched items still read `PARTIAL`, as intended.

One wording note: `PARTIAL rather than SUPPORT` became `INDIRECT`, not `INDIRECT rather than SUPPORT`. They are no longer alternatives — one is the value, the other qualifies it — so the comparison had to go rather than be transposed.

Zero `directness: INDIRECT` items now name `PARTIAL`.

## 🔵 3 — IE: taking it, because you're right

`REFUTE` did read a step past the sentence. The link asserts prophylaxis `INHIBITS` bacteraemia — a claim about what the drug does — and the sentence denies that lowering bacteraemia predicts lower IE, which is the *next* step. Your observation that using bacteraemia as a surrogate at all presumes the drug lowers it is the clinching one: the quote doesn't contradict this link, it doesn't bear on it. Now `NO_EVIDENCE`.

I preferred that to narrowing the link description, since the description is accurate and it was my grading that overreached.

## 🔵 4 — both removed

Checked each against its entry:

- **`Inherited_Porphyria` treatments[6]** — entry says "Dersimelagon is an **investigational** oral MC1R agonist being developed for EPP and XLP"; the quote says exactly that. Same thing.
- **`Heparin-Induced_Thrombocytopenia` treatments[2]** — description already states guidelines treat the choice as conditional; the quote directly reports that recommendation.

Both caveats were about claims the entries don't make. `treatments[5]` stays — agreed, that one is genuinely class → member.

**95 assignments now, down from 97.** Both history records rewritten to record what was decided *and* what the review corrected, since they had described the superseded reasoning.

## Re-verified

| Check | Result |
|---|---|
| `linkml-validate -C Disease`, 85 entries | No issues found |
| Reference validation on the CAH file | 39/39 verified, no cache churn |
| `directness: INDIRECT` items still naming `PARTIAL` | 0 |
| Untouched siblings in CDH2 / Marfan still naming `PARTIAL` | 1 each, as intended |
| seven offline gates | clean, no baseline updated |
| `validate-history` on both records | No issues found |
| evidence / HPOA / SEPIO / directness / snippet tests | 7,982 passed |

---

cmungall:
Took both of the non-blocking notes rather than deferring them — `0dc3b600a`.

You called them wording, and they are, but they're wording in the one place it matters most: this PR's main contribution is the rule, and two of its own 95 worked examples justified themselves by the criterion the rule explicitly excludes. Shipping the doc section alongside two exemplars that don't follow it would undercut the thing being added.

**HIT treatments[5]** — was "the endorsement is explicitly conditional and does not establish these agents as preferred". Now names the class→member gap you identified: the guideline says "direct oral anticoagulants" inside a list of non-heparin options, while the entry is rivaroxaban and apixaban specifically.

**`Nipah_Virus_Disease` treatments[2]** — was "compassionate use in an uncontrolled outbreak does not establish clinical efficacy". Now: the quote reports what was *recommended* rather than what the drug did, so it reaches a treatment entry through the recommendation rather than through an outcome.

Both explanations now say explicitly that the sample-size or design point, while true, is not the reason for the grading — so they read as applications of the rule rather than exceptions to it.

No graded value changes; the diff is two explanation blocks. Verified: `linkml-validate` clean on both entries, folded-hyphen / snippet-grading / duplicate-key / entity-ref gates clean, 7,905 tests passing in the evidence-schema-directness selection.

Sorry for the extra round — that push dismisses your approval again.

On your closing point, that these are what the next sweep over the ~3,500 can pick up: agreed, and it's a useful thing to have learned here. The sweep's job isn't only swapping a retired token — a good number of those explanations will name a caveat on the wrong axis, and the rule now gives whoever does it something to sort them against.

---

Issue #10112, opened 2026-08-30
Prefixed `downstream[].target` values break the pathograph: 174 edges across 30 entries, one of which drops to 0/7 phenotypes connected


cmungall:
## The finding

A causal edge written as `target: phenotypes#Seizure` does not reach the phenotype named `Seizure`. The pathograph builder resolves `downstream[].target` by **bare name**, so the prefixed string matches nothing, the edge is routed to a phantom node, and the real phenotype is left with no incoming edge. Thirty entries are affected today, 174 edges in total, and every one of them passes `just validate`, `check-entity-refs` and the full test suite.

The effect is not subtle. Across those 30 entries, **1 of 337 phenotypes is causally connected**:

```
$ uv run python -m dismech.qc_plugins <the 30 files> -c conf/qc_config.yaml
Phenotype connectivity: 1/337 nodes causally connected (0.3%); 30 file(s) with gaps.
```

Every one of those disorder pages renders a pathograph in which the mechanism chain stops before the phenotypes, and a set of red orphan nodes appears beside it carrying the phenotype names.

## Why nothing catches it

Two subsystems read the same slot with **opposite conventions**:

| Subsystem | Bare `Seizure` | Prefixed `phenotypes#Seizure` |
|---|---|---|
| `check_entity_refs` / `test_entity_ref_foreign_keys` | skipped, unchecked | resolved and validated ✅ |
| `src/dismech/graph.py` (`build_causal_graph`) | resolved ✅ | no match → orphan node |

`REF_SLOTS` in `src/dismech/entity_refs.py` includes `target`, so the prefixed form is exactly what entity-ref validation is built to check — and it passes, because the reference genuinely resolves against `SECTION_KEYS`. Meanwhile `build_causal_graph` (graph.py:401-419) takes `edge_item["target"]` verbatim and looks it up in `graph.nodes`, which is keyed by each item's plain `name`.

The mismatch is recorded, just nowhere anyone looks. graph.py:696-700 adds the unresolved target to `orphan_targets` **and** appends to `integrity_issues` — but `integrity_issues` is only ever passed to the HTML template as `graph_issues` (render.py:2371, 2553). No test, `just` recipe or CI step reads it.

## This is the mirror image of #9697

[...]

PR #10142, opened 2026-08-30
curate(Marinesco-Sjogren_Syndrome): add SIL1/BiP ER-proteostasis entry (MONDO:0009567)

### LINKED ISSUE #10139: Curate Marinesco-Sjogren syndrome (MONDO:0009567)

Curate a dismech entry for **Marinesco-Sjogren syndrome** ([MONDO:0009567](https://monarchinitiative.org/MONDO:0009567)) — the autosomal recessive syndrome of cerebellar ataxia, congenital cataracts, myopathy with rimmed vacuoles and intellectual disability, caused by biallelic loss-of-function variants in *SIL1*, the nucleotide-exchange factor for the ER chaperone BiP.

Preflight notes: nothing in `kb/disorders/` carries this MONDO ID. The name appears only inside `Autosomal_Recessive_Cerebellar_Ataxia_With_Late_Onset_Spasticity.yaml`, which curates the *GBA2*-related "Marinesco-Sjogren-**like**" phenocopy — a different gene and a different disease, so this is the phenotype-similarity trap rather than an overlap.

**Two naming hazards to watch.** The stub's MONDO-derived synonym list includes "Marshall Smith Syndrome", which is a distinct disorder (*NFIX*) and must not be carried into the entry or used to seed literature searches. A subset of patients have been reported with congenital disorder of glycosylation features under the CDG framing; check what is actually attributable to *SIL1* before curating it.

MONDO records no descendants for this term and a single causal gene, `hgnc:24624` SIL1.

Stub: `stubs/Marinesco-Sjogren_Syndrome.yaml`

nlharris:

Curates **Marinesco-Sjogren syndrome** (`MONDO:0009567`, *SIL1*) as a new `kb/disorders/` entry. Closes #10139.

## What this disease is, mechanistically

MSS is a disease of endoplasmic reticulum proteostasis. *SIL1* encodes the adenine nucleotide exchange factor for BiP/HSPA5, the master ER chaperone. Without nucleotide exchange, BiP cannot release its folded clients, so unfolded protein accumulates in the ER lumen and the unfolded protein response is chronically engaged. The three cell types that fail — cerebellar Purkinje neurons, skeletal muscle fibres, lens — are the ones with the heaviest folding and secretory load, which is what produces the classic triad of ataxia, myopathy and cataract.

The pathograph has seven nodes, each causal edge separately evidenced:

`SIL1 loss of NEF activity` → `stalled BiP chaperone cycle` → `ER unfolded protein accumulation and ER stress` → (`PERK-branch UPR activation` → `degeneration of Purkinje neurons, muscle fibres and lens`) and (`secretory pathway trafficking failure` → `aberrant extracellular matrix assembly`).

The PERK node is placed on the causal path rather than beside it because the evidence is a rescue, not a correlation: GSK2606414 delays Purkinje degeneration and motor deficits in woozy mice, and attenuates cell death in SIL1-knockdown cells.

## Contents

| Section | Detail |
|---|---|
| `pathophysiology` | 7 nodes, all `biological_scale`-tagged, 7 evidenced causal edges |
| `phenotypes` | 13 HPO-bound; frequency bands only where the source says "invariably" or "all patients beyond the age of 7" |
| `histopathology` | Vacuolar myopathy with the perinuclear membranous ultrastructural hallmark |
| `animal_models` | woozy mouse linked to two mechanism nodes via `modeled_mechanisms`, with readouts, fidelity and limitations on each |
| `discussions` | One `KNOWLEDGE_GAP`, one `HUMAN_MODEL_MISMATCH` |
| `genetic`, `inheritance`, `progression`, `definitions`, `treatments` | GeneReviews-sourced |

GeneReviews `PMID:20301371` exists for this disease and is used as the phenotype baseline, tagged in the top-level `references:` block.

## Three wrong ontology bindings in the deep-research report

One OpenScientist report was generated and is committed under `research/`. Its reference validation was clean (16/16 verified, confabulation rate 0.0), but **its term validation set `needs_review: true`, and three of the flagged bindings were real errors** that would have propagated into this entry had they been copied across:

| Report's binding | What that CURIE actually is | Used here instead |
|---|---|---|
| `MONDO:0008541` as the MSS disease term | *spermatic cord torsion* | `MONDO:0009567` |
| `HP:0002151` for cerebellar/vermian atrophy | *Increased circulating lactate concentration* | `HP:0001272` Cerebellar atrophy |
| `GO:0036498` as PERK-mediated UPR | *IRE1-mediated unfolded protein response* | `GO:0036499` |

This is a good illustration of why `needs_review` rather than `confabulation_rate` is the key to read — the rate was 0.0 for both references and terms, because every identifier resolved. They just resolved to the wrong things.

Most of the report's other 14 "mislabelled" flags were an artefact of its phenotype table, whose *Type* column (`Clinical sign`, `Lab abnormality`) was parsed as the term label. Those CURIEs resolve correctly and were used after checking each against the committed term cache.

## Named Entity Confusion checks

Two specific traps here, both avoided and both recorded in the entry `notes`:

- **The stub's MONDO synonym list contains "Marshall Smith Syndrome"** — a different disorder (*NFIX*). It is not in this entry's synonyms and was kept out of literature searches.
- **`Autosomal_Recessive_Cerebellar_Ataxia_With_Late_Onset_Spasticity.yaml` already curates a *GBA2*-related "Marinesco-Sjogren-**like**" syndrome.** Different gene, different mechanism, phenocopy only. Nothing was cited across between the two entries.

## Knowingly incomplete

1. **No `prevalence` record.** The deep-research report cites Orphanet ORPHA:559 for a rate below 1/1,000,000, but `ORPHA:559` is not in this repository's `references_cache` and `just fetch-reference ORPHA:559` fails (no source configured for that reference type). Rather than build a prevalence record on an unquotable figure, none is made.
2. **No `biochemical` record for creatine kinase.** Universally described as elevated in MSS, but not stated in any abstract cached for this entry.
3. **Peripheral neuropathy and pyramidal signs** appear in the report's phenotype table but not in a cached quotable source at the asserted frequency, so they are omitted.
4. **The woozy lens phenotype is not modelled.** The `PARTIALLY_RECAPITULATES` link says so explicitly in its `limitations` — cataract, the third element of the human triad, is not reported among woozy phenotypes in the cited studies.
5. Frequency bands are given for three phenotypes only. Everything else is unbanded rather than guessed.

## Validation

```
just validate kb/disorders/Marinesco-Sjogren_Syndrome.yaml
  Schema validation... No issues found
  Term validation... ✅ Validation passed
  Snippets checked: 51/51 verified against cached references

just check-entity-refs      OK: entity references resolve in 1 KB file(s).
just check-duplicate-keys   OK: no duplicate mapping keys in 1 YAML file(s).
just check-snippet-length   OK: no new snippets under 5 words
just check-title-snippets   OK: no new title-quoting snippets
just check-snippet-grading  OK: no new snippet-grading divergences
just check-folded-hyphens   OK: no new folded-scalar hyphen splits
just validate-history       No issues found
```

Staging was targeted: only the 10 `references_cache` files this entry cites, and only the 3 cache rows it introduced (`GO:0000774`, `hgnc:24624`, and the matching enum row).

## One tooling note

`just new-history` fails on any multi-word argument — the recipe passes `{{args}}` unquoted, so `--summary "Create: Marinesco-Sjogren Syndrome"` is word-split and the shell chokes on the parenthesis in `--details`. Worked around by calling `scripts/new_history.py` directly. This is pre-existing and was reported before, in the reviewer notes on #7293; flagging it again because it is still there and it silently blocks the documented workflow.

---

### DIFF
--- a/cache/go/terms.csv
+++ b/cache/go/terms.csv
@@ -62,6 +62,7 @@ GO:0000724,double-strand break repair via homologous recombination,2026-03-19T15
 GO:0000725,recombinational repair,2026-08-15T20:12:27.355221
 GO:0000729,DNA double-strand break processing,2026-05-27T19:03:39.030726
 GO:0000768,syncytium formation by cell-cell fusion,2026-07-14T17:13:27.749183
+GO:0000774,adenyl-nucleotide exchange factor activity,2026-09-04T19:25:12.724979
 GO:0000781,"chromosome, telomeric region",2026-08-31T18:05:41.929336
 GO:0000785,chromatin,2026-03-19T16:11:00.471909
 GO:0000786,nucleosome,2026-03-19T16:11:00.478556


nlharris:
**Correction, pushed in 7113face7: one `reference_title` in this entry was wrong, and it was the worst kind.**

The reviewer on [#10145](https://github.com/monarch-initiative/dismech/pull/10145) caught paraphrased `reference_title` values there. Auditing all five entries in this batch against `references_cache` turned up 26; only one is here, but it is not a paraphrase — **it is a different paper's title**:

| PMID | What I wrote | Actual cached title |
|---|---|---|
| `PMID:19801575` | "Protein accumulation and neurodegeneration in the woozy mutant mouse is caused by disruption of SIL1, a cochaperone of BiP." | "Alteration of the unfolded protein response modifies neurodegeneration in a mouse model of Marinesco-Sjögren syndrome." |

The title I attached belongs to `PMID:16116427`, the woozy-mouse discovery paper, which this entry also cites. Both papers are about the same mouse, which is why the mistake was easy to make and why nothing caught it: the `PMID:19801575` snippet — the HYOU1/DNAJC3 modifier result quoted in the `HUMAN_MODEL_MISMATCH` discussion — verified as an exact quote of `PMID:19801575`, because the PMID and the quote were both right. Only the displayed title was another paper's.

`EvidenceItem.reference_title` is not machine-checked; `just validate` only checks top-level `references[].title`. So this survived a clean 51/51 snippet verification.

The claim in the discussion is unaffected — it rests on the snippet and the PMID, both correct. `just validate` still passes: 51/51. No other content changed.

---

nlharris:
All seven findings addressed in cecc807dd, plus four of the six suggestions. Genuinely useful review — items 4 and 5 were the ones I should have caught myself.

## 🔴 1. `reference_title` for `PMID:19801575`

Correct, and it was another paper's title, not a paraphrase. Fixed, and then I ran an audit across all five entries in this batch: **26 wrong `EvidenceItem.reference_title` values across four of them**, all now copied verbatim from the cache. Detail in the earlier comment on this PR; the other three PRs have their own fix commits.

Your parenthetical is the important part and I have filed it as [#10149](https://github.com/monarch-initiative/dismech/issues/10149): `linkml-reference-validator` checks `references[].title` but not `EvidenceItem.reference_title`, which is where nearly all of them live. In two PRs in this batch the validator flagged a `references[].title` mismatch, I fixed it, and concluded titles were verified — they were, in the one place it looks.

## 🟡 2. Folded scalar splitting `prenatal/preimplantation`

Right, and a nice catch given `check-folded-hyphens` cannot see it. Reflowed so the token is intact. Worth noting the reference validator did not catch it either — it normalises whitespace before matching, so `prenatal/ preimplantation` verified as an exact quote. Two guards, neither of which covers a slash.

## 🟡 3. HYOU1 and DNAJC3 as `MODIFIER` entries

Agreed — these were structural facts sitting in a `notes` string. Both are now `genetic:` entries with `relationship_type: MODIFIER` and their own evidence (`hgnc:16931`, `hgnc:9439`). I kept the direction explicit in each, because the DNAJC3 result is counterintuitive: *losing* a second chaperone component improves the phenotype, which is what points at the nucleotide-exchange cycle rather than chaperone capacity as the broken thing.

## 🟡 4. Bundled node, unsupported lens arm

You are right, and I had already written the caveat at line 959 without drawing the obvious conclusion from it. Now two atomic nodes — `Cerebellar Purkinje Neuron Degeneration` and `Skeletal Muscle Fibre Degeneration`.

**No lens node was created.** Nothing in the cited sources supports one, and inventing a third node to make the triad look symmetrical would have been the same error in a new place. The woozy link is retargeted and, now that it points at a node it actually reproduces, upgraded from `PARTIALLY_RECAPITULATES` to `RECAPITULATES` — the absent cataract phenotype is recorded in `limitations` instead, where it belongs.

## 🟡 5. Pathograph never reached the phenotype layer

The finding I am most glad you made — two disconnected islands is a real modelling defect and I did not notice it. Splitting item 4 gave exactly the edges you predicted:

- `Cerebellar Purkinje Neuron Degeneration` → `phenotypes#Cerebellar ataxia`
- `Skeletal Muscle Fibre Degeneration` → `phenotypes#Muscle weakness`

Both evidenced and typed `INDIRECT`, since the coupling is demonstrated in the mouse.

## 🟡 6. Negative preclinical results

Added, and it did more work than expected. `PMID:39804912` is now three evidence items on the `perk_inhibition_translational_ceiling` discussion, and the rationale is rewritten around it. The framing that changed: trazodone, DBM and TUDCA were chosen specifically to keep the mechanism while losing GSK2606414's pancreatic toxicity, all failed in the same model on the same presymptomatic schedule, and pharmacokinetics excluded altered drug metabolism. So the open question narrowed from "does proteostasis modulation work" to "is anything short of full PERK inhibition sufficient, and if not does a therapeutic window exist at all". The primary snippet is graded `REFUTE`.

## 🟡 7. Treatment terms too generic

Both fixed with the precedents you cited: `NCIT:C157809` Cataract Surgery, `NCIT:C15599` Hormone Replacement Therapy.

## 🔵 Suggestions

- **GeneReviews management** — three added (developmental/educational support, feeding support, orthopaedic management).
- **`preferred_term: Cerebellar ataxia`** — agreed, changed.
- **`genetic_context` zygosity/allele_type** — allele classes are now in the `description`. I left `zygosity` unset deliberately: the enum is single-valued and MSS patients are homozygous *or* compound heterozygous, so picking one would assert something false about the other half. Happy to add it if there is a convention I have missed.
- **`Aberrant Extracellular Matrix Assembly` as a terminal leaf** — left alone. The authors *propose* aberrant ECM explains the skeletal and lens features, and the entry says so, but that is a proposal in a discussion section rather than a demonstrated link. Wiring it to scoliosis or short stature would assert more than the source does. Flagging that I read it and declined rather than missed it.
- **Prevalence via `structured-rebuild-orphanet`** — you were right about the mechanism and I had the wrong tool. It still does not work here: the recipe needs `data/orphadata/en_product1.xml`, which is gitignored, and `just refresh-orphadata` fails on a checksum mismatch because Orphadata has republished the file since the `MANIFEST.yaml` pin. Filed as [#10150](https://github.com/monarch-initiative/dismech/issues/10150). The entry's `notes` now records this as attempted-and-blocked rather than not-attempted.
- **CK omission** — thank you for grepping the caches to check rather than assuming.

## Validation

```
just validate: 62/62 snippets verified (was 51/51)
check-entity-refs, check-duplicate-keys, check-snippet-length,
check-title-snippets, check-snippet-grading, check-folded-hyphens: all OK
```

Cache diff is one additive row (`hgnc:16931`); `hgnc:9439` was already present.

The `just new-history` defect is now [#10148](https://github.com/monarch-initiative/dismech/issues/10148), as you suggested.

---

nlharris:
Findings 8 and 9 fixed in 630ace9db. Both were mine — the split created them — and the lesson generalises, so I have written it into the entry's `notes`:

> Splitting a bundled node is not finished when the node is split; every edge and every piece of evidence that pointed at the old node has to be re-aimed.

## 8. Muscle node had no incoming edge

Exactly right, and your framing is the useful one — this was finding 5 relocated, not removed. I added the outgoing edge and never checked what pointed *at* the new node.

Fixed using the evidence you identified, which was already in the PERK node's own block: `it also reduced the skeletal muscle abnormalities and improved motor performance during the symptomatic phase`. That makes the muscle edge rest on the same rescue argument as the Purkinje edge, which is the right symmetry — PERK is upstream of both arms or of neither.

## 9. Muscle evidence on the Purkinje-only link

The `explanation` string was still saying "the muscle arm of this node" about a node that no longer had one, which is worse than a stale reference — it made a claim about Purkinje degeneration rest on a quote about muscle.

I took your suggested fix rather than just deleting the item, because you were right that it makes the entry better: the phenocopy statement now supports a new `RECAPITULATES` link on `Skeletal Muscle Fibre Degeneration`, with the perinuclear-membranous-structure readout attached. That also closes something I hadn't noticed — the woozy myopathy is the subject of this entry's `histopathology` section and previously reached no mechanism node at all.

The Purkinje link now carries a Purkinje-specific quote (`adult-onset ataxia with cerebellar Purkinje cell loss`).

Pathograph after the fix — single connected component, both arms terminating in phenotypes:

```
SIL1 loss → stalled BiP cycle → ER stress ─┬→ PERK ─┬→ Purkinje degeneration → phenotypes#Cerebellar ataxia
                                            │        └→ muscle fibre degeneration → phenotypes#Muscle weakness
                                            └→ secretory trafficking failure → aberrant ECM assembly
```

## Two suggestions taken

- **The muscle→weakness edge now cites the myopathy description** rather than the fidelity statement. You were right that "woozy muscle is a faithful phenocopy" says the model is good, not that myopathy produces weakness. It now quotes the sentence naming progressive vacuolar myopathy as a feature of the syndrome.
- **`Orthopedic Management` rebound** from `NCIT:C16186` *Orthopedic Surgical Procedure* to `NCIT:C49236` *Therapeutic Procedure*, with `therapeutic_modality: OTHER`. Your "item 7 inverted" framing is exactly it — GeneReviews says "per orthopedist" and I bound surgery it does not assert. The description now says why the generic term is deliberate.

## Also added

**A second history record** for this review-response session, per your convention note. The original record is unchanged — it accurately describes the creation session, and history is append-only.

## Left alone, with reasons

- **`genetic_context` `zygosity` / `allele_type`.** Still empty deliberately. `zygosity` is single-valued and MSS patients are homozygous *or* compound heterozygous; picking either asserts something false about the other half. The allele classes and the biallelic requirement are in the `description`. If there is a convention for this case I have missed, I will take it.
- **`Aberrant Extracellular Matrix Assembly` as a terminal leaf** — as before, the authors *propose* it explains the skeletal and lens features, and that is a proposal, not a demonstrated link.

## Validation

```
just validate: 65/65 snippets verified (was 62/62)
check-entity-refs, check-duplicate-keys, check-snippet-length,
check-title-snippets, check-snippet-grading, check-folded-hyphens: all OK
just validate-history: No issues found
```

No cache changes in this commit; only the entry and the new history record.

---

nlharris:
The merge queue dequeued this PR on a `check-causal-targets` failure. Fixed in `ee4796c59`, and the same fix should unblock #10144 — see below.

**What was wrong.** Two `downstream` targets used the entity-reference grammar in a slot that takes a bare name:

```diff
-  - target: phenotypes#Cerebellar ataxia
+  - target: Cerebellar ataxia
-  - target: phenotypes#Muscle weakness
+  - target: Muscle weakness
```

This is the failure mode CLAUDE.md and #10112 describe, and it is silent: the unresolved target draws a phantom red duplicate node labelled with the raw string and orphans the real phenotype out of the graph, while the entry still validates and the page still renders. So `Cerebellar ataxia` and `Muscle weakness` — the two phenotypes this entry's central mechanism exists to explain — were disconnected from the pathograph. Both names exist verbatim in `phenotypes`, so dropping the prefix is the whole fix, exactly as the checker's message said.

**Why the PR's own checks were green.** `check-causal-targets` does not exist on this branch — nor do `check-qualifier-terms` or `check-enum-values`. All three were added to `main` after this branch opened on 08-30. The merge queue tests the PR merged against current `main`, so it ran a gate the branch has never seen. That is the same "green against its own base, red on merge" shape as the duplicate-key and retired-enum problems already documented in CLAUDE.md, and it is worth knowing that a five-day-old curation branch can acquire a CI failure without anyone touching it.

**Verification.** Because the gate is absent here, I ran `main`'s own `scripts/check_causal_targets.py` against the fixed file rather than trusting the local recipe set:

```
OK: no new broken pathograph targets (123 dangling target(s) grandfathered).
```

Plus, on this branch: `just validate` passed, `just validate-terms` passed, **65/65** snippets verified, `check-entity-refs` and `check-duplicate-keys` clean, and the new history record validates. No claim, evidence item, snippet or node name was changed — two lines in the entry, plus the history record.

**About #10144.** Its queue run failed with *these same two findings in this file*, not with anything of its own — the queue branch for it carried this PR's commits. I have not pushed anything to #10144; it should pass once this lands. Worth a look if it fails again for a different reason.

---

nlharris:
The `claude-review` failure on `ee4796c59` is **not this PR's**, and nothing was reviewed. The review agent hit an account usage limit and exited before doing any work — 273 ms, one turn, $0:

```
detail="You've hit your limit · resets 4:40am (UTC)"
```

([run 33834561411](https://github.com/monarch-initiative/dismech/actions/runs/33834561411/job/100904284754).) The PR comment renders this as "Claude encountered an error after 0s", which reads like the fix broke something. It didn't — the agent never looked at the diff.

**Why I am not re-running it yet.** The limit resets at 04:40 UTC and the failure was at 04:08, so a re-run now would hit the same wall and spend the retry for nothing. I have a check-in scheduled just after the reset and will re-run then; if it errors the same way again I will say so here rather than retrying in a loop.

**One thing for a maintainer**, since it is the likely root cause and not mine to change: the workflow's own diagnostic says it prefers `ANTHROPIC_API_KEY` (pay-per-use, no weekly cap) and only falls back to `CLAUDE_CODE_OAUTH_TOKEN` when no API key is set. A usage-limit error here therefore suggests `ANTHROPIC_API_KEY` is unset in repo secrets and reviews are running on the subscription token. That would affect every agent workflow in the repo, not just this PR.

The code fix itself stands unchanged and was verified against `main`'s own checker — `OK: no new broken pathograph targets` — plus `just validate`, `validate-terms`, 65/65 snippets, `check-entity-refs` and `check-duplicate-keys`. `test (3.13)` is the check that matters for mergeability and is unaffected by this.

---

nlharris:
Correcting my previous comment: I said I would re-run `claude-review` after the usage limit reset. **I can't** — dispatching the workflow returns `403 Resource not accessible by integration`. The limit itself has reset (it was due at 04:40 UTC), so a re-run should now succeed; I just cannot start one.

Where this leaves the PR:

| | |
|---|---|
| `test (3.13)` on `ee4796c59` | ✅ **success** — the required check is green on the fixed head |
| `claude-review` | ❌ failed at 04:08 UTC on the usage limit, never ran |
| Mergeability | clean — I test-merged `main` into this branch and it merges with no conflicts |
| Latest review | the approval on `630ace9db`, auto-dismissed by my fix push |

So the causal-target fix is verified by the check that gates merging, and what is missing is a fresh review verdict on `ee4796c59`.

**What would unblock it:** a maintainer re-running `claude-code-review.yml` for PR 10142, or any subsequent push to this branch, which re-triggers it automatically. I am deliberately not creating a push to force that — this branch has no conflict to resolve and no further change to make, so a commit whose only purpose was to restart CI would be exactly the empty-commit-to-kick-CI pattern the contributing rules forbid.

I will keep watching this PR until it is green rather than handing it off, and will report here if the next review errors the same way.

---

nlharris:
> **This comment was generated by Claude,** running as @nlharris.

Good catch, and the diagnosis is exactly right. Fixed in `1910dce85`.

`GO:0000774 adenyl-nucleotide exchange factor activity` is back in `cache/go/terms.csv`, restored with `just validate-terms` and ordered with `just normalize-cache`. Only that file is staged — one line.

**To be clear about what happened, since it wasn't the fix commit:** `ee4796c59` (the causal-target fix) was the last thing I pushed, and it carried the row. The loss came from `c5671e9f6`, a *"Merge branch 'main'"* commit made through the GitHub UI at 18:55Z, which resolved the `cache/go/terms.csv` conflict by taking `main`'s side wholesale. I hadn't seen that commit when it landed. Worth recording because the review reads it as fallout from my push, and the distinction matters for the rebase-vs-merge point below.

**The verification you used is the right one and I have adopted it.** You found this because your validation run left the tree dirty with the row re-added at a fresh timestamp. So after committing I re-ran `just validate-terms` and confirmed the tree stays **clean** — which is the actual property at stake (the cache is self-sufficient offline) rather than the weaker "validation passed".

I also checked the direction of the delta properly, because a naive comparison is misleading here. Against **current** `main` the branch appears to have 10 cache deletions; against `5041fbd77`, the `main` commit actually merged in, it is **additive only — 4 insertions, 0 deletions**. The apparent deletions are just `main` advancing to `5b08b3d88` in the half hour since. Anyone reading that diff should compare against `c5671e9f6^2`, not `origin/main`.

## On rebase vs merge

Half agreed, and I want to be precise about which half.

You are right that this specific loss is a merge artifact and a rebase would have replayed the branch's cache addition on top of `main`'s rows. But a rebase would not have been *safe* on any of these branches: rebasing rewrites history, and `main` has `dismiss_stale_reviews`, so replaying six commits through five rounds of accumulated review would have invalidated far more than one row.

The property that actually catches this class of bug is not the choice of operation — it is checking, after either one, that every row the *branch* contributed is still present. A rebase can drop one too, in a conflicted hunk. So I have taken the lesson as a check rather than as a preference:

> After refreshing a branch, compare the pre-refresh head against the post-refresh head over `cache/`, not just the result against `main`. A row the branch added and `main` never had does **not** appear as a deletion in a diff against `main`.

That last sentence is the trap. My own additive-only check was a diff against `main`, and it would have reported `GO:0000774` as fine.

## I ran that check against my other three branches

Three sibling PRs of mine had the same conflict shape today (#10145, #10453, #10457) and I resolved all three the same way. So I re-checked each pre-merge → post-merge, and then proved offline determinism the way you did:

| PR | Rows dropped from the branch side | `validate-terms` leaves tree clean |
|---|---|---|
| #10145 | none — `CL:1001607` re-derived, new timestamp | ✅ |
| #10453 | `GO:0006655` — **correctly** dropped, see below | ✅ |
| #10457 | none — `hgnc:18005`, `MONDO:0030397` re-derived | ✅ |

The #10453 one is a stale row, not a loss: `GO:0006655` was narrowed to `GO:0035965 cardiolipin acyl-chain remodeling` during that PR's own review rounds, so the entry stopped binding it before this merge ever happened. `GO:0035965` is present in both the label and enum caches. I have a correction to post there, because my commit message on that branch says I re-derived `GO:0006655` — it re-derived `GO:0035965`, and the PR body carries the same stale term.

## The `MODEL_ORGANISM` note at `:646`

Still not taken, still deliberately. I agree `HUMAN_CLINICAL` is the better reading of that sentence — it is the abstract's background statement about the human disease, not a result of the mouse study. I am leaving it because changing it would dismiss this approval for a judgment call you yourself scoped as not worth a round trip, and this PR has now spent six review rounds. If you would rather have it fixed than merged, say so and I will push it.

---

PR #10145, opened 2026-08-30
curate(Progressive_Pseudorheumatoid_Arthropathy_Of_Childhood): add CCN6 entry (MONDO:0008827)

nlharris:
> **This pull request was generated by Claude,** running as @nlharris.

Curates **progressive pseudorheumatoid arthropathy of childhood** (PPRD, `MONDO:0008827`, *CCN6*/*WISP3*) as a new `kb/disorders/` entry. Closes #10140.

## What makes this disease worth a careful entry

PPRD is a recessive skeletal dysplasia in which articular cartilage degenerates progressively with **no inflammation at all**. That single fact drives everything clinically important about it: ESR, CRP, rheumatoid factor and ANA are normal, so immunosuppressants and biologics cannot work — and yet the disease looks so much like juvenile idiopathic arthritis that patients are routinely given them anyway.

The cost is measurable. In the Turkish cohort of 44, median symptom onset was **4 years** and median age at diagnosis **9.7 years**; 47.7% lost independent walking at a median age of 12. The review states plainly that diagnosis is usually made only in the second decade and that children "often receive unnecessary anti-inflammatory and immunosuppressive treatments" in the interval.

Because of that, diagnostic delay is curated as a `progression` phase in its own right rather than left as a service-quality footnote — in this disease it is part of the natural history.

## Pathograph

`biallelic CCN6 LOF` → `loss of CCN6 modulation of BMP and Wnt signaling` → `failure of articular chondrocyte matrix homeostasis` → `progressive noninflammatory articular cartilage loss` → `joint contracture, deformity and loss of ambulation`.

The second edge is deliberately typed **`INDIRECT_UNKNOWN_INTERMEDIATES`**. Both its ends are well supported — CCN6's BMP/Wnt inhibition is shown by gain-of-function in zebrafish, and cartilage loss is documented radiographically in patients — but the middle is not. Three competing proposals (loss of matrix-synthetic drive, loss of SOD-dependent antioxidant protection, increased chondrocyte IGF-1 sensitivity) are all untested in human articular cartilage and none is excluded. Typing that edge `DIRECT` would have asserted a chain nobody has demonstrated.

## A negative biochemical finding, curated on purpose

`biochemical` carries one record: **inflammatory markers and autoantibody serology, `presence: NORMAL`**. Normal results are usually not worth curating; here the normality *is* the finding. It is the single most useful discriminator from JIA and the reason immunosuppression is futile.

[...]

PR #10144, opened 2026-08-30
curate(Congenital_Thrombotic_Thrombocytopenic_Purpura): add ADAMTS13 entry (MONDO:0010122)

nlharris:
> **This pull request was generated by Claude,** running as @nlharris.

Curates **congenital thrombotic thrombocytopenic purpura** (Upshaw-Schulman syndrome, `MONDO:0010122`, *ADAMTS13*) as a new `kb/disorders/` entry. Closes #10137.

## The lump/split call

`kb/disorders/Acquired_Thrombotic_Thrombocytopenic_Purpura.yaml` already exists and names congenital TTP **only as a `differential_diagnoses` entry**. This PR makes it a Disease in its own right rather than a subtype, because the two diverge at the top of the pathograph and at the point of treatment:

| | Congenital TTP | Acquired (immune) TTP |
|---|---|---|
| Cause | Biallelic germline *ADAMTS13* loss of function | Acquired anti-ADAMTS13 autoantibody |
| Inhibitor | Absent — that absence is diagnostic | Present |
| Onset | Bimodal: neonatal/childhood, or adult unmasked by pregnancy | Usually adult |
| Definitive therapy | ADAMTS13 replacement | Immunosuppression + caplacizumab |

They converge only below the enzyme-deficiency node. Curating cTTP as a subtype of the acquired entry would have put a germline disease inside an autoimmune one.

## Pathograph

Seven nodes, each causal edge separately evidenced:

`biallelic ADAMTS13 LOF` → `severe plasma ADAMTS13 activity deficiency` → `persistence of ultralarge VWF multimers` → `shear-driven platelet adhesion and microthrombus formation` → three branches: `consumptive thrombocytopenia`, `mechanical erythrocyte fragmentation`, `microvascular occlusion and ischemic end-organ injury`.

## The negative result is the interesting part

The `Adamts13`-knockout mouse is linked with **`FAILS_TO_RECAPITULATE`** against the microthrombus node, with `limitations` and evidence as the schema requires for a substantive negative claim. Null mice are viable with normal survival. Spontaneous thrombocytopenia appears only on a high-VWF background, and a TTP-like syndrome only after shigatoxin challenge — and plasma VWF level did **not** correlate with severity, so the permissive factor is not simply VWF.

[...]

PR #10146, opened 2026-08-30
curate(Metaphyseal_Chondrodysplasia_Jansen_Type): add PTH1R gain-of-function entry (MONDO:0007982)

### LINKED ISSUE #10138: Curate metaphyseal chondrodysplasia, Jansen type (MONDO:0007982)

Curate a dismech entry for **metaphyseal chondrodysplasia, Jansen type** ([MONDO:0007982](https://monarchinitiative.org/MONDO:0007982)) — the severe short-limbed skeletal dysplasia caused by heterozygous constitutively activating variants in *PTH1R*, the PTH/PTHrP receptor, producing ligand-independent receptor signalling with hypercalcaemia and hypophosphataemia despite suppressed PTH.

Preflight notes: nothing in `kb/disorders/` or `kb/groupings/` carries this MONDO ID or label. `Chung-Jansen_Syndrome.yaml` is a different disease (PHIP-related) and shares only the eponym — worth naming here so the confusion is not rediscovered.

This is a good counterpart to the loss-of-function end of the same receptor (Blomstrand chondrodysplasia, Eiken syndrome); if either turns out to be curated, the relationship is worth recording rather than duplicating.

MONDO records no descendants for this term and a single causal gene, `hgnc:9608` PTH1R.

Stub: `stubs/Metaphyseal_Chondrodysplasia_Jansen_Type.yaml`

nlharris:

Curates **Jansen metaphyseal chondrodysplasia** (`MONDO:0007982`, *PTH1R*) as a new `kb/disorders/` entry. Closes #10138.

## One lesion, two phenotypes

Heterozygous activating *PTH1R* variants make the PTH/PTHrP receptor signal constitutively through Gs-alpha with no ligand present. That single lesion accounts for both halves of a disease that otherwise looks like two:

- **In the growth plate**, PTH1R signalling normally holds chondrocytes in proliferation and delays hypertrophy. Locked on, it delays hypertrophy pathologically — metaphyseal widening, bowed limbs, severe short stature.
- **In kidney and bone**, the same constitutive signal produces hypercalcaemia and hypercalciuria with *low or suppressed* PTH. PTH-independent hypercalcaemia is close to unique among skeletal dysplasias and is the diagnostic signature.

```
constitutively active PTH1R
  → ligand-independent Gs-alpha / cAMP signaling
      ├→ SIK2/SIK3 inhibition → class IIa HDAC de-repression
      │     → delayed growth plate chondrocyte hypertrophy
      │        → short-limbed short stature and skeletal deformity
      └→ renal and skeletal calcium mobilization
            → hypercalciuria, nephrocalcinosis and renal impairment
```

The SIK node earns its place on a **phenocopy in the informative direction**: combined *Sik2*/*Sik3* deletion in osteoblasts and osteocytes produces a skeletal and molecular phenotype "that closely resembled" what the constitutively active JMC receptor produces. Removing the kinase reproduces the disease.

## The age dependence is curated as three progression phases

This is the part a clinician most needs and it is easy to state wrongly. Serum calcium was **normal in most patients postnatally**, elevated between roughly 0.15 and 10 years, and **tended to normalise again in adults**.

So a normal calcium excludes nothing at either end. The entry records both traps, each with its worked case: diagnosis in infancy is explicitly described as challenging without a family history or overt hypercalcaemia, and a 38-year-old mother in one family "was never overtly hypercalcemic and was therefore not diagnosed until her sons were found to be affected." Urinary calcium/creatinine, by contrast, is elevated in both children and adults — which is why it also shows up as the candidate trial endpoint in one of the discussions.

## Contents

| Section | Detail |
|---|---|
| `pathophysiology` | 8 nodes, scale-tagged, evidenced edges |
| `progression` | 3 phases organised around the biochemical age dependence |
| `phenotypes` | 9 HPO-bound; bands only where the cohort text supports one |
| `biochemical` | Calcium (`INCREASED`) and phosphate (`DECREASED`, with the source's own hedge that it sits at the low end of normal rather than being frankly low) |
| `genetic` | All five reported alleles; H223R and the milder T410R broken out |
| `animal_models` | H223R growth-plate transgenic, `RECAPITULATES`, `MODERATE` fidelity |
| `treatments` | Hydration, bisphosphonate, orthopaedic surgery, physiotherapy — each with a `target_mechanisms` link and an honest effect type |
| `discussions` | Two `KNOWLEDGE_GAP`s |
| `histopathology` content | Carried in the calcium-mobilization node: osteoid buildup in osteocyte lacunae, shortened canaliculi, enhanced FGF23 and diminished sclerostin staining |

GeneReviews `PMID:40638774` exists and is the phenotype baseline, tagged in `references:`.

## Two knowledge gaps that are specific rather than decorative

1. **Does allele-specific constitutive activity predict severity?** The ingredients exist and the analysis does not. Five mutants show varying basal cAMP signalling in reporter cells; separately, adult height was below the third centile in every patient *except* T410R carriers. Nobody has fitted those against each other, and with 24 patients across five alleles — 18 of them one allele — the series cannot support it. It also needs separating whether basal activity predicts the skeletal and renal arms equally, which is not obvious since they act through different tissues.

2. **What would an inverse-agonist trial measure?** The pharmacology is encouraging — the inverse agonist reduced basal cAMP for *every* mutant, so it acts on the lesion rather than a consequence. The difficulty is structural: serum calcium is only informative in a window, adult height takes two decades, and nephrocalcinosis, the thing a trial would most want to prevent, reads out slowest. An endpoint strategy is a prerequisite here, not an afterthought.

## Deep research: the one report that declined to guess

One OpenScientist report was generated and is committed. It was **the cleanest of the five in this batch**: 32/32 references verified, 33/34 terms verified, one label mismatch (`GO:0007050`, whose label resolves as the bare identifier because the term is obsolete — not used here).

Worth calling out explicitly: this report **wrote "MONDO ID not resolved in this investigation" rather than supplying a plausible-looking wrong one**. Three of the other four reports in this batch offered a disease-level MONDO term naming an entirely different disease (`MONDO:0008541` → spermatic cord torsion in #10142; `MONDO:0010134` → Pendred syndrome in #10144; `MONDO:0009215` → Fanconi anemia group A in #10145), each while reporting `confabulation_rate: 0.0`. Abstaining was the right behaviour and is why this entry needed no MONDO correction.

## Two validator catches, one of which fixed a factual error

- A first draft quoted the **article title** of `PMID:11713230` as the animal-model evidence. The reference validator rejected it — and reading the actual abstract not only supplied a much better sentence but **corrected the entry**. The transgene rescues *PTHrP*-null lethality; in *receptor*-null mice it restored most bone abnormalities but did **not** prevent perinatal lethality. The draft had conflated the two. Both facts are now curated, and the second is what makes the fidelity `MODERATE` rather than `HIGH`.
- The inverse-agonist snippet originally carried the peptide's bracketed residue notation and failed to match, because bracketed spans are stripped unless `conf/reference_validator_config.yaml` lists a `literal_bracket_patterns` entry. Rather than widen shared config for one quote, the snippet was shortened to a bracket-free span of the same sentence.

Both are recorded in the entry `notes` and the history record.

## Named Entity Confusion

`kb/disorders/Chung-Jansen_Syndrome.yaml` is a different disease (*PHIP*-related) sharing only the eponym. Nothing was cited across and no Chung-Jansen source was consulted.

## Knowingly incomplete

1. **No `prevalence` record.** No population rate is published; the largest series is 24 patients.
2. **No biochemical record for FGF23.** Elevation is real and reported, but the quotable statements attach it to bisphosphonate pharmacology or to staining intensity rather than to a measured circulating disease phenotype. The osteocyte staining finding is carried in a node description instead.
3. **Severe infantile hypertension** (`PMID:31977144`) and the **craniofacial series** (`PMID:39830149`) are cached but not curated as phenotypes — the first is a single case posed as a question by its own title, the second needs frequency data its abstract does not carry.
4. No datasets.

## Validation

```
just validate kb/disorders/Metaphyseal_Chondrodysplasia_Jansen_Type.yaml
  Schema validation... No issues found
  Term validation... ✅ Validation passed
  Snippets checked: 53/53 verified against cached references

just check-entity-refs      OK
just check-duplicate-keys   OK
just check-snippet-length   OK
just check-title-snippets   OK
just check-snippet-grading  OK
just check-folded-hyphens   OK
just validate-history       No issues found
```

Targeted staging: 9 cited `references_cache` files and the 3 cache rows this entry introduced.

---

### DIFF
--- a/kb/disorders/Metaphyseal_Chondrodysplasia_Jansen_Type.yaml
+++ b/kb/disorders/Metaphyseal_Chondrodysplasia_Jansen_Type.yaml
@@ -904,6 +904,7 @@ phenotypes:
   - reference: PMID:40638774
     reference_title: PTH1R-Related Jansen Metaphyseal Chondrodysplasia.
     supports: SUPPORT
+    directness: INDIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       drugs targeting the angiotensin receptor pathway as needed for elevated
@@ -911,7 +912,10 @@ phenotypes:
     explanation: >-
       GeneReviews carries a disorder-specific treatment recommendation for
       elevated blood pressure, which is independent of the single case report
-      and is what makes this more than a one-patient observation.
+      and is what makes this more than a one-patient observation. Graded
+      INDIRECT because the quote instructs clinicians to treat elevated blood
+      pressure and the phenotype claim follows from that by one inference step,
+      rather than being asserted outright.
   - reference: PMID:31977144
     reference_title: Severe hypertension-An infantile feature of Jansen metaphyseal chondrodysplasia?
     supports: SUPPORT
@@ -1680,9 +1684,11 @@ notes: >-
   pharmacology or to immunohistochemical staining intensity rather than to a
   measured circulating disease phenotype, so no biochemical record is made for
   it; the osteocyte staining finding is carried in the pathophysiology node
-  description instead. The craniofacial series (PMID:39830149) is cached but not
-  curated as its own phenotype record: it needs frequency data the abstract does
-  not carry. Hypertension was initially left out on the same reasoning - the only
+  description instead. The craniofacial series (PMID:39830149) was consulted but
+  not curated as its own phenotype record: it needs frequency data the abstract
+  does not carry. Its cache is deliberately not shipped, because an uncited
+  reference cache belongs to no entry - which is also why the earlier claim here
+  that it "is cached" became wrong once the uncited-cache pruning pass ran. Hypertension was initially left out on the same reasoning - the only
   dedicated report is a single neonate, and its own title poses the association
   as a question - but that reading missed that GeneReviews independently tells
   clinicians to measure blood pressure at every visit and to treat elevation with


nlharris:
All seven findings addressed in ee58dcef8. Finding 7 was the valuable one — it turned up evidence that materially strengthens a node the entry already had.

## 7. Animal models — I under-consumed the research, and it cost the entry its best result

Three models added. The T410R humanized knock-in (`PMID:40455993`) is the one I should not have missed:

> Genetic ablation of Hdac4 rescued the growth plate abnormalities in T410R mice, thereby establishing the PTH1R-Gαs-cAMP-PKA-SIK3-HDAC4/5 pathway as the main mediator of growth plate abnormalities in JMC.

That is an **epistasis result** — remove the downstream HDAC and the phenotype is rescued — and the entry's SIK/HDAC node was resting on the `Sik2/Sik3` phenocopy alone. The same paper also shows an acute PTH inverse agonist normalising both serum calcium and endogenous PTH, which bears directly on the `inverse_agonist_translation` discussion.

The other two, with what makes each distinct:

- **H223R humanized knock-in** (`PMID:37808400`) — `RECAPITULATES` the growth-plate lesion (expanded proliferating/prehypertrophic zones, loss of type X collagen-positive hypertrophic chondrocytes). Its limitations carry the two things that bound it: founders typically die within two months without reproducing, and **serum calcium and phosphate were no different from wild type**, so it does not model the mineral-ion arm at all.
- **Sik2/Sik3 double knockout** (`PMID:31430259`) — typed as a mechanism phenocopy, not a disease model, and the limitations say so: no PTH1R lesion, deletion restricted to osteoblasts/osteocytes rather than chondrocytes, no mineral-ion phenotype. It places the node; it cannot test therapies.

Worth noting the H223R paper also states plainly that the earlier promoter-restricted transgenics — including the one this entry already cited — "neither transgenic JMC model adequately recapitulates the human disease". That context is now in the entry.

## 4. Bundling

You were right, and it was self-evident in a way I should have caught: the node's own incoming edge narrated all three steps in sequence. Now `Hypercalciuria` → `Nephrocalcinosis` → `Chronic Kidney Disease`, each evidenced. Splitting also gave the thiazide (finding 6) a precise target — it acts on the excretion step, not the calcium load — and let the hydration treatment retarget onto nephrocalcinosis specifically.

## 5. `GO:0004991`

Correct. `GO:0004930` said "a GPCR", when the node's whole point is *which* receptor. Changed.

## 3. `diagnosis:` and `differential_diagnoses:`

Both added. The differential is Blomstrand lethal chondrodysplasia, and it earns its place beyond being a differential — it is the loss-of-function mirror image at the same locus, with the growth-plate phenotype running the opposite way (accelerated differentiation and premature ossification against JMC's delay). The two bracket the receptor's dose-response. They are not clinically confusable, since Blomstrand is perinatally lethal, so the entry says that too rather than implying a bedside dilemma.

`diagnosis:` covers the calcium/phosphate/PTH combination (framed as the *combination* being diagnostic — raised calcium with raised PTH is hyperparathyroidism), radiographs as what carries the diagnosis in infancy when the biochemistry cannot, `PTH1R` testing, and the renal/mineral surveillance from finding 2.

## 1 and 2. GeneReviews

Joint swelling and both dental phenotypes added. Surveillance for calcium, phosphate, 24-hour urine calcium and renal ultrasound is a `diagnosis` record rather than a treatment, since it is monitoring rather than intervention — happy to move it if there is a convention I have missed. Contact-sport avoidance and the remaining surveillance items (growth, dental, audiometry, ophthalmic, blood pressure) are in `notes` for the same reason.

## 6. Thiazide

Curated as its own treatment. Your framing was right that it was hiding in an `explanation` string — and it deserves separating from the bisphosphonate rather than being folded in, because the source's point is that alendronate alone reduced hypercalciuria and **only the thiazide achieved normocalciuria**.

## Validation

```
just validate: 76/76 snippets verified (was 53/53)
check-entity-refs, check-duplicate-keys, check-snippet-length,
check-title-snippets, check-snippet-grading, check-folded-hyphens: all OK
```

Cache diff is 2 additive rows (`GO:0004991` plus its enum row); two new reference caches, both cited.

Prevalence remains unavailable — no published population rate for a disease with a 24-patient largest series.

---

nlharris:
Blocker confirmed and fixed in `8bf1b188b`. You were right, and the cause is worse than a staging slip.

`references_cache/PMID_40455993.md` was not merely unstaged — I had **deleted** it. When I cleaned up uncited reference caches, I generated the list of cited PMIDs *before* adding the animal-models section, so the pruning pass removed a cache the entry had just started citing. The same pass also deleted `PMID_31977144.md`, which the entry's `notes:` still described as "cached but not curated" — so that prose had become false too. Both are restored via `just fetch-reference`, and 80/80 snippets now verify on the pushed tree.

I checked the other four PRs in this batch for the same defect. All clean: cTTP 79/79, MSS 65/65, Dauber 53/53, PPAC 65/65. Jansen was the only casualty.

Taking your advice that a follow-up chore commit costs a full re-review cycle, I bundled the round-1 suggestions into the same push rather than paying for a second one:

- **Chronic kidney disease** as a phenotype (`HP:0012622`, `OCCASIONAL`) — two of 24 in the natural-history cohort. The entry already modelled it as a pathophysiology node but never surfaced it as a phenotype.
- **Hypertension** as a phenotype (`HP:0000822`, no frequency band). This one I got wrong the first time. I excluded it reasoning only from the single neonatal case report, whose own title poses the association as a question — and missed that GeneReviews independently tells clinicians to measure blood pressure at every visit and to treat elevation with angiotensin-pathway drugs. A disorder-specific management recommendation is not a one-patient observation. The case report is attached as a second item graded `directness: INDIRECT`, since one case with a negative secondary workup is consistent with but does not establish a causal link to the receptor defect. The `notes:` block now records the reversal instead of leaving my original exclusion argument standing.
- **GeneReviews 50% per-child transmission quote** added to `inheritance` — the mode-of-inheritance quote alone does not carry what counseling actually turns on.
- **Epidemiology provenance** made explicit: the cohort size of 24 is stated in the study's objectives, not its results, because it describes the cohort assembled rather than a finding measured in it. The quote stays as the only source for the number, and the `explanation` now says so.
- Second history record.

**Not changed:** the SIK2/SIK3 + class IIa HDAC node. I looked at your bundling point again and I think the node is right as one step — SIK inhibition and HDAC de-repression are two readouts of a single PKA phosphorylation event, not two mechanistic claims, and the `Sik2/Sik3` double-knockout model in the entry targets exactly that one node. Happy to split it if you disagree.

Validation on the pushed tree: `linkml-validate` clean, `linkml-term-validator --labels` clean, `check-entity-refs` and `check-duplicate-keys` clean, and all four baseline gates report no new divergences.

---

nlharris:
Pushed `1ca797cbf` with the three round-2 suggestions, while the round-3 re-review of `8bf1b188b` was still running — apologies for the moving target. `8bf1b188b` is the commit that fixes the blocker; this one only adds the optional items on top, so the blocker verdict from that run still applies to what is now the parent commit.

**Thiazide `therapeutic_agent`** — `NCIT:C49185` Thiazide Diuretic. The source says "thiazide diuretic" without naming a drug, so the class term is the right grain rather than picking a specific agent the literature does not specify.

**The misplaced `limitations`.** You were right that this was a category error, not just awkward placement. "Reported as an acute normalisation after a single injection; durability under chronic dosing is not addressed" describes the inverse-agonist *measurement*, so it moved into the readout's `interpretation`. The link now carries the fidelity caveat it should have had all along: T410R is the milder end of the human spectrum and not the common H223R, and murine mineral-ion handling differs from human in calcium set point and renal phosphate response — so the magnitude of the abnormality does not transfer even though its direction and PTH-independence do.

**Schmid metaphyseal chondrodysplasia** added as a differential. Your framing of the problem was the useful part: the section listed one differential that is not a bedside dilemma and none that are. `PMID:8644413` turned out to carry both halves of what the section needed — it places Schmid and Jansen in the same radiographic group ("which can be separated by clinical, radiographic, genetic and biochemical criteria") and states the discriminator ("The radiographic changes are similar to rickets, but calcium and phosphorus metabolism is normal"). That second sentence is exactly the contrast the entry stakes JMC's identity on, and the case report itself documents a Schmid patient treated as rickets, which is the error the differential exists to prevent.

**Third history record** for this round.

Validation on the pushed tree: `linkml-validate` clean, `linkml-term-validator --labels` clean, **82/82 snippets verified**, `check-entity-refs` and `check-duplicate-keys` OK, and all four baseline gates report no new divergences. One new reference cache (`PMID_8644413.md`, staged this time), no cache row changes.

---

nlharris:
Thanks for the approval on `8bf1b188b`. Two pushes landed after it — `1ca797cbf` and now `5747969cd` — which between them take all four of the remaining suggestions, so nothing optional is left open. I'm aware each push costs an approval; I judged it worth it because one of the four was a factual error I had introduced.

**`notes` claimed `PMID:39830149` "is cached" and it was not.** This is the same defect as the `PMID:31977144` one, from the same pruning pass, and I did not catch it when I fixed that one — I corrected the sentence I was already editing and did not re-read the rest of the paragraph for the same class of claim. The scoping decision stands (it needs frequency data its abstract does not carry), so the sentence now reads "consulted but not curated" and records why an uncited cache is deliberately not shipped.

**GeneReviews hypertension snippet now `directness: INDIRECT`.** Your reasoning is right and I should have graded it that way when I added it: the quote instructs clinicians to treat elevated blood pressure, and the phenotype claim follows by one inference step rather than being asserted. The case report on the same phenotype was already `INDIRECT`, so the two items now agree about what kind of claim each is making — which was inconsistent before, since I graded the weaker source more carefully than the stronger one.

From `1ca797cbf`, already described in the comment above: the thiazide `therapeutic_agent` (`NCIT:C49185`), the T410R `limitations`/readout relocation, Schmid MCD as the confusable differential (`PMID:8644413`), and a third history record.

Validation on `5747969cd`: `linkml-validate` clean, `linkml-term-validator --labels` clean, **82/82 snippets verified**, `check-entity-refs` and `check-duplicate-keys` OK, and all four baseline gates report no new divergences.

---

PR #10161, opened 2026-08-30
Add mouse models annotated on the MP level and activity branches, for four disorders

dragon-ai-agent:
## Summary

The Mammalian Phenotype ontology keeps "abnormal enzyme level" and "abnormal enzyme activity" in two entirely separate branches — levels under abnormal homeostasis (`MP:0005319`), activities under abnormal metabolism (`MP:0005584`). HPO does not make that split. This PR curates seven mouse models chosen specifically because MGI annotates their gene on **both** branches, so each entry is a worked example of how much an amount measurement and an activity measurement can differ for the same enzyme.

It also records the underlying question — whether dismech needs to track "quantity kind" (amount vs catalytic activity vs flux) as structured data — as a deferred decision in the design register, rather than changing the schema.

## How the models were selected

Downloaded MGI's `MGI_PhenoGenoMP.rpt` (395,549 genotype-phenotype annotations), pulled the descendant term lists for both MP branches from OLS (189 and 222 terms, with **no overlap** — the branches really are disjoint), and joined to genes already curated in dismech. That gives 88 genes annotated on both branches, 40 of which dismech already covers. Four were curated here. None of the four disorders had any animal models before.

## The models

| Disorder | Model | Why this one |
|---|---|---|
| Hypophosphatasia | `Alpl<tm1Jlm>` null (PMID:9056646), `Alpl<tm2.1Jlm>` A116T knock-in (PMID:25716980), and `Alpl<tm1Sor>` null (PMID:7550313) | The first two are the same gene on opposite MP branches — the null annotated to decreased ALP *level*, the knock-in to decreased ALP *activity* — and opposite ends of the allelic spectrum: 50% residual plasma activity, a radiographically normal postcranial skeleton, a phenotype confined to the dentoalveolar complex. The third is an independently derived null carrying the pyridoxal 5'-phosphate mechanism the other two do not test. |
| Mucolipidosis type II | `Gnptab` gene-trap (PMID:17962477, PMID:17652091) | The clearest case in the set. Serum lysosomal enzyme levels go **up** (the enzymes are secreted instead of delivered to the lysosome) while mannose-6-phosphate tagging is **abolished**. Both directions are curated as readouts on one link so the dissociation is explicit. |
| Menkes disease | brindled and viable-brindled `Atp7a` (PMID:16371425, PMID:9342151) | SOD3 specific activity falls and is partially restored by adding copper, while cytosolic SOD1 is unaffected — activity lost without the enzyme protein being lost. |
| Gaucher disease due to saposin C deficiency | `PSAP` C384S knock-in (PMID:20175216) | Loss of the activator rather than the enzyme. Carries a `FAILS_TO_RECAPITULATE` link: no glucosylceramide accumulation in brain or liver and no organomegaly, so the model cannot support storage-burden or substrate-reduction endpoints. |

## Two TNAP-null lines, curated separately on purpose

Review suggested adding PMID:7550313 to the existing `Alpl<tm1Jlm>` model to substantiate the vitamin B6 mechanism. MGI annotates that paper to `Alpl<tm1Sor>`, a different knockout, so putting its evidence on tm1Jlm would have been a cross-allele attribution. It is a third model instead.

The two lines also **disagree**: tm1Sor is reported skeletally normal, tm1Jlm is not. That is recorded rather than smoothed over — `PARTIALLY_RECAPITULATES` at `LOW` fidelity with an `UNCHANGED` skeletal readout, and a `notes:` line saying the discrepancy is reported rather than resolved, with allele construction, background and assessment age all differing. The tm1Sor B6 link is `RECAPITULATES` with the full chain: serum PLP `INCREASED`, brain GABA `DECREASED`, seizures `RESTORED` under pyridoxal.

The general rule worth keeping: an MGI-sourced PMID belongs to the allele MGI annotates it to, not to a sibling knockout of the same gene.

## Points a reviewer may want to check

**Four links are deliberately not full recapitulation**, each with the negative result recorded rather than omitted: the A116T skeletal link (postcranial skeleton normal); the tm1Jlm seizure link (the paper hypothesizes an autonomic mechanism and does not test PLP); the tm1Sor skeletal link; and the brindled cuproenzyme link at `LOW` fidelity, because the cuproenzyme measured is vascular SOD3, not a neuronal one.

**One node was deliberately left unlinked.** The viable-brindled paper reports mitochondrial impairment but does not attribute it to copper accumulation, so it is not linked to the "Mitochondrial redox imbalance from copper accumulation" node.

**MGI does not record the assay.** `MGI_PhenoGenoMP.rpt` carries genotype, MP term and PMID and nothing about method or units, so every assay statement here was read off the cited paper and lives as prose in the readout name and interpretation. The structured `assays:` slot is left empty, consistent with the rest of the knowledge base, where all 564 assay descriptors are free text and none is bound to an ontology term.

**Genotype strings were verified against MGI locally**, since the CI runner has no network egress to MouseMine or the Alliance API. All seven match `MGI_PhenoGenoMP.rpt` exactly. `Alpl<tm1Sor>` is MGI:1857124, "targeted mutation 1, Philippe Soriano" — the lab code recorded in the model's `notes:`.

## Validation

- All 33 evidence snippets confirmed as exact substrings of the reference cache, checked after YAML folding rather than before.
- `just validate-disorders` green on all four files, re-run after rebasing onto current `main`.
- Duplicate-key, folded-hyphen, snippet-length, title-snippet, snippet-grading, entity-ref and environmental-evidence gates green. Two folded-hyphen findings introduced by this branch were fixed by reflowing, not by updating the baseline.
- `just validate-history-all` clean.
- Model-link and foreign-key tests pass, including the checks that every readout target repeats its link target and that `FAILS_TO_RECAPITULATE` links carry limitations and evidence.

## Files

Twenty files, **zero deletions**: four disorder entries, seven history records, eight new reference cache files (PMID:7550313, PMID:9056646, PMID:9342151, PMID:16371425, PMID:17652091, PMID:17962477, PMID:20175216, PMID:25716980), and one row added to `docs/explanation/design-decisions.md`.

Nine existing cache files were incidentally regenerated by the validator during development and have been **reverted to their state on `main`** — nothing here depends on them, and reverting removed a lost trailing newline without hand-editing a generated file.

The single `docs/` file is outside the curation-path allowlist, which is what the `scope-override` label covers. That row is the deferred decision these models are the worked examples for.

https://claude.ai/code/session_01Hv7gk83PWHFepZTfLSDHTD


dragon-ai-agent:
Thanks — the copper finding was right and is fixed. All six points are addressed in b33248fed, but two of them turned out differently from the suggestion, so flagging those.

## The blocking item

Corrected. Succinate-cytochrome c reductase is not copper-dependent, and calling it that undercut the distinction the entry exists to draw. The description now restricts the copper claim to cytochrome oxidase and states why the muscle deficit is not evidence of cuproenzyme failure:

> Only cytochrome oxidase (complex IV) is a cuproenzyme; succinate-cytochrome c reductase spans complexes II and III, which use flavin, iron-sulfur clusters and hemes, so the muscle deficit is not itself evidence of cuproenzyme failure.

## Item 2 — PMID:7550313 is a different allele

This is the one worth a second look. The suggestion was to add Waymire 1995 to the existing model to substantiate the B6 mechanism. It cannot go there: MGI annotates PMID:7550313 to **`Alpl<tm1Sor>`**, not to the `Alpl<tm1Jlm>` line already curated. They are independently derived knockouts, and attaching the evidence to tm1Jlm would have been a cross-allele attribution.

The two lines also disagree on something substantive that the review did not flag. Waymire reports:

> In contrast, mice lacking TNAP generated by homologous recombination using embryonic stem (ES) cells have normal skeletal development.

against Narisawa's "abnormal bone mineralization" for tm1Jlm. Normal skeletal development in a TNAP-null mouse is the opposite of the defining human phenotype.

So `Alpl<tm1Sor>` is added as a **third, separate model**:

- B6 link → `RECAPITULATES`, three readouts — serum PLP `INCREASED`, brain GABA `DECREASED`, seizures `RESTORED` under pyridoxal. The rescue arm is what makes the mechanism causal rather than correlative.
- Skeletal link → `PARTIALLY_RECAPITULATES` at `LOW` fidelity, with the `UNCHANGED` skeletal readout graded `REFUTE`, plus the dentition defect that only appears in rescued animals.
- The discrepancy between the lines is recorded in `notes` as reported, not resolved — allele construction, background and assessment age all differ.

The tm1Jlm B6 link stays `PARTIALLY_RECAPITULATES` rather than being upgraded, and now says the mechanism comes from the sister allele and is not asserted for this one on that basis.

## Items 3, 4, 5

Taken as written. Link-level evidence added to the brindled cuproenzyme link. The three Gaucher `FAILS_TO_RECAPITULATE` items changed from `REFUTE` to `SUPPORT` — the reasoning holds, the link's claim *is* the failure. The mucolipidosis M6P readout now states explicitly that the measurement comes from the companion study and is attributed to the alpha/beta genotype generally rather than verified as this specific gene-trap line.

## Item 6 — cache files

Rather than hand-fix a trailing newline in a generated file, I reverted all nine incidentally regenerated cache files to their state on `main`. They were validator churn, nothing here needs them, and the diff against `main` is now 19 files with zero deletions. Running `just validate-disorders` locally regenerates them again, so this is the fetcher's output format rather than anything editable.

## The allele symbols CI could not verify

Checked against a local copy of `MGI_PhenoGenoMP.rpt`. All six exact, with annotation-row counts:

| Genotype | Rows |
|---|---|
| `Alpl<tm1Jlm>/Alpl<tm1Jlm>` | 38 |
| `Alpl<tm2.1Jlm>/Alpl<+>` | 11 |
| `Gnptab<Gt(OST97730)Lex>/Gnptab<Gt(OST97730)Lex>` | 17 |
| `Atp7a<Mo-br>/Y` | 6 |
| `Atp7a<Mo-vbr>/Y` | 19 |
| `Psap<tm1Suz>/Psap<tm1Suz>` | 75 |

`just validate-disorders` green on all four files. All 33 snippets confirmed exact, including the seven new ones. Folded-hyphen, snippet-length, snippet-grading, duplicate-key and entity-ref gates green.

One note on the earlier red build: it was the curation-scope guard, not a test failure — `docs/explanation/design-decisions.md` sits outside the curation path allowlist. The change is the deferred-decision row these models are the worked examples for, so I added the `scope-override` label, which is what the guard's message asks for. Happy to split the docs change into its own PR instead if you would rather this stay a pure curation PR.

---

dragon-ai-agent:
Both remaining suggestions are addressed in 77a11abac. Flagging that this push dismisses the approval — the naming one was a real factual error, so it was worth the re-review.

## The lab code is Soriano, not Sorrentino

You were right to distrust the expansion, and the name was simply wrong. I pulled MGI's allele report, which the CI runner cannot reach:

```
MGI:1857124   Alpl<tm1Sor>   targeted mutation 1, Philippe Soriano   Targeted   Null/knockout
```

So `Sor` is Soriano. Renamed to `TNAP-null mouse, Soriano line (Alpl tm1Sor)`, and the verification is now recorded in `notes:` with the MGI accession, so the next reader does not have to re-derive it or take it on trust:

> The lab code in the allele symbol is Soriano — MGI records `Alpl<tm1Sor>` (MGI:1857124) as "targeted mutation 1, Philippe Soriano" — and MGI annotates PMID:7550313 to this genotype.

Worth noting for anyone reading the allele record later: MGI lists PMID:7789278 as the allele's originating reference, while PMID:7550313 (Waymire) is the phenotype annotation that brought it into this entry. Both are true and they are different things.

## The skeletal readout grade — changed, deliberately

Changed `REFUTE` to `SUPPORT`, so the negative results in this PR are graded one way rather than two.

The reasoning that moved the three Gaucher items applies here identically: a readout's claim is its measurement, and the quoted sentence asserts that measurement. Grading it `REFUTE` was reading the evidence against the *node* rather than against the readout, which is a different object. The disagreement with the node's skeletal arm is already carried by `PARTIALLY_RECAPITULATES`, `LOW` fidelity, the `limitations` and the `notes` — putting it in the grade as well says it twice and makes the grade mean something different here than three files away. The explanation now records that choice rather than leaving it implicit.

Straightforward to flip back if you read it the other way; it is claim-relative either way, as you say.

## Validation

`just validate-disorders` green on Hypophosphatasia. Folded-hyphen, snippet-grading and entity-ref gates green. All six PMID:7550313 snippets re-confirmed exact after the edit. Diff against `main` unchanged at 19 files, zero deletions.

---

dragon-ai-agent:
Added the second history record in 8e10153d2 — you were right that the earlier one had gone stale on the readout grade, and right that the fix is to append rather than edit.

`history/disorders/Hypophosphatasia/2026-08-31T141327Z-claude-code-18dbd7.yaml` records both changes from that commit and states explicitly which detail of the `2026-08-31T013814Z` record it supersedes, so the two read in order without either being rewritten. `just validate-history-all` clean.

Thanks for checking the "In contrast" attribution in the Waymire abstract. That was the right thing to be suspicious of — it is the exact shape of the mistake the third model exists to avoid, and I would rather it were checked than assumed.

This push is only a history record, so it dismisses the approval again for a file that is not KB content. Sorry for the extra cycle; the alternative was leaving a stale record in the tree.

---

PR #10162, opened 2026-08-30
Verify GEO dataset accessions as references; retire the shared accession cache

cmungall:
## Summary

`cache/dataset_accessions.json` was a single 1.8 MB JSON file rewritten in full by every run of the dataset verifier — including a run over one disorder file. So every curation PR that touched a `datasets:` block churned the same file, and PRs adding neighbouring `geo:` keys collided on it. This replaces it, for `geo:` accessions, with the mechanism the repo already uses for a PMID: one cached file per record under `references_cache/`.

Turning that on exposed a second problem. Once a GEO record is cached, the validator can compare the curated `datasets[].title` against the repository's real title — and **33 dataset titles did not match**. They were plausible-looking paraphrases that nothing had ever checked. Reading those rewrites then surfaced four further errors, including one record pointing at **the wrong study entirely**. All are fixed here.

## The storage change

`Dataset.accession` already declared `implements: linkml:authoritative_reference` — it was always a reference slot. `conf/reference_validator_config.yaml` simply listed the dataset prefixes under `skip_prefixes`, so nothing resolved them.

- `just verify-datasets` now resolves a `geo:` accession through the reference fetcher, which writes `references_cache/GEO_<ID>.md` with GEO's title and summary. **A cache file present is the verification** — the fetcher writes one only after NCBI returned a record. Commit it with the `datasets:` block, as you would a `PMID_*.md`.
- All 940 `geo:` accessions in `kb/` are backfilled, so verification runs offline and makes no network calls on an untouched file.
- `geo` and `GEO` are removed from `skip_prefixes`. GEO records are now checked like any other reference.
- Other prefixes (EGA, MassIVE, dbGaP, PRIDE …) keep their API resolvers and stay skipped. Their results are a per-run memo, not persisted anywhere.

**The old JSON file is frozen, not deleted.** Around 200 open PRs still carry edits to it, so removing it now would conflict with all of them. Nothing reads or writes it any more; a new test (`test_no_automation_touches_the_frozen_dataset_cache`) enforces that, and `CLAUDE.md` and `AGENTS.md` tell agents never to touch or stage it. Deleting it is a follow-up for when those PRs have drained.

Committing it was buying nothing anyway: no CI job and no test read it, unlike `references_cache/` and `cache/**/*.csv`, which are committed precisely because validation determinism requires them. All 1,800 of its entries were status `OK`, and its payload duplicated the `datasets[].title` already in the entry.

## What the check found

**33 titles.** `datasets[].title` is a title slot next to a reference field, so the validator compares it with the fetched record.

| Curated | GEO's actual title |
|---|---|
| `Left ventricular noncompaction cardiomyopathy gene-expression dataset GSE71912` | `Mutations in the NOTCH pathway regulator MIB1 cause left ventricular noncompaction cardiomyopathy` |
| `Nr2f1 chromatin occupancy in mouse cortex` | `Nr2f1 shapes mitochondria in the mouse brain unraveling novel insights into the neurodevelopmental disorder BBSOAS` |

The rule this establishes: **`datasets[].title` is the repository's title, copied exactly.** The curator's own summary of the dataset belongs in `description`. The title keeps the repository's typos — `geo:GSE301492` now carries GEO's misspelled "Reed-Stenberg" over a curator's silent correction, for the same reason an evidence snippet never fixes the source it quotes. Reviewers may want to look at that one.

**A record pointing at the wrong study.** `Spaceflight_Associated_Neuro-Ocular_Syndrome` keyed a record on `geo:GSE152382` (Nrf2-knockout multi-tissue spaceflight RNA-seq), but its `description`, `data_type: METHYLATION`, and `publication: PMID:34179686` all described Chen et al., a retinal methylome study. NCBI `elink` from that PMID resolves to `GSE159771`, which matches every other field; the accession is now that. This was invisible before because the old paraphrased title was consistent with the description — correcting the title is what made the record contradict itself.

**Three organisms contradicting their own corrected titles.** `GSE71912` and `GSE113251` declared `Homo sapiens` over series NCBI reports as mouse-only. `GSE330200` is a mixed series (NCBI lists both taxa), so its value was incomplete rather than wrong; `organism` is single-valued, so it now records the mouse knock-in model its title names, with `notes` stating that the human samples are not represented by that slot.

**2 evidence snippets**, and they were different problems:

- `GEO:GSE289185` quoted a *reordered paraphrase* of a sentence that was in the cached summary all along — clauses moved, "transcriptome sequencing" rewritten as "RNA sequencing". Exactly the drift the exact-quote rule exists to catch. Now the sentence verbatim.
- `GEO:GSE316127` quoted GEO's `overall design` field, which the fetcher does not cache, so it was unverifiable rather than wrong. Replaced with the summary sentence reporting the study's model-based result.

Both `explanation` fields were rewritten to state what the quote establishes rather than asserting an experimental design the quote does not carry. `supports` and `evidence_source` are unchanged in both — the claims were right, only the grounding was not.

**The drift is ongoing, not historical.** Three of the 33 titles arrived on `main` *after* this branch's original audit — two in `Myocarditis`, one in `Sarcopenia`, all truncated copies of GEO's title, all caught by refreshing against `main`. That recurrence over four days is the case for keeping the check on.

## Why not a `datasets/` folder

Considered and rejected as a de-duplication layer: of 1,747 dataset records, 1,696 accessions are distinct and only 49 (2.9%) appear in more than one entry, maximum fan-out 3 — and every one of those is a pair of sibling entries (`MED13`/`MED13L`, `Ehlers-Danlos`/`Hypermobile_EDS`, `Schwannoma`/`Schwannomatosis`). Inline `datasets:` records stay inline. A genuine catalog layer for a data lakehouse — access tier, file manifests, checksums, mirror paths — is a different argument and still open; if built, its home is the existing `src/dismech/structured_sources/` framework, not a third mechanism. Recorded as design decision 6c.

## Verification

| Check | Result |
|---|---|
| GEO dataset titles vs cached records | 972/972 match |
| `geo:` accessions with a cached record | 940/940 |
| KB-wide snippet audit, GEO enabled | 141,851/147,150 verified, **0** GEO failures |
| `linkml-reference-validator`, flip live | 0 issues |
| `pytest -k "history or dataset"` | 2,353 passed |
| Schema, duplicate-key, entity-ref, history-schema | clean over every touched entry and history record |
| Five offline snippet checks | no new violations against baselines |

Diff against `main` is 818 additions and 32 modifications: no deletions, no derived-path churn, and `cache/dataset_accessions.json` untouched.

`test (3.13)` failed initially on the curation-scope guard, not a test — this PR touches `kb/` and infra together by design, which is what the guard asks a `scope-override` label for. The label is applied.

## Follow-ups

1. Migrate the next prefixes, highest volume first — `ega` (382), `massive` (120), `metabolomics_workbench` (81), `dbgap` (71). Each needs a fetcher, a cache backfill, and the curation pass the newly-enabled title and snippet checks force. `linkml-reference-validator` already ships a `BIOPROJECT` source, and its generic `json_api` source may cover others with configuration rather than code. Budget these as curation work, not a config edit — GEO cost 33 titles, 2 snippets, and 4 record-level errors.
2. Delete `cache/dataset_accessions.json` once the open PRs carrying edits to it have merged.

https://claude.ai/code/session_01LRY3Arbhh78xHaCWXoDB94

### DIFF
--- a/history/disorders/Contact_Dermatitis/2026-09-02T215803Z-claude-code-29e150.yaml
+++ b/history/disorders/Contact_Dermatitis/2026-09-02T215803Z-claude-code-29e150.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Contact_Dermatitis
+  path: kb/disorders/Contact_Dermatitis.yaml
+session:
+  id: 2026-09-02T215803Z-claude-code-29e150
+  timestamp: '2026-09-02T21:58:03Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10162
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - datasets
+  - evidence
+  summary: Drop a GEO dataset evidence item that quoted only its own title
+  details: |
+    The evidence item on the geo:GSE153007 dataset record used the record's own title verbatim as its snippet, identical to datasets[].title and reference_title, so it asserted nothing the record did not already state. It was invisible until now because the title-snippet guard exempts dataset accessions it cannot fetch a title for, and that exemption is derived from the reference validator's skip_prefixes -- which this branch removes geo from. The GEO record has no summary (its summary field repeats the title) and NCBI reports no linked publication, so there is no abstract sentence to quote and no better source to cite. Per the guard's own documented remedies and the repository convention that a dataset record with nothing quotable carries publication/notes rather than manufactured evidence, the evidence block was removed and the reason recorded in notes. Title, description, organism, data_type and conditions are unchanged and still carry everything the evidence item restated.
--- a/kb/disorders/Contact_Dermatitis.yaml
+++ b/kb/disorders/Contact_Dermatitis.yaml
@@ -1574,14 +1574,12 @@ datasets:
   - non-lesional skin
   - atopic dermatitis skin
   - psoriasis skin
-  evidence:
-  - reference: GEO:GSE153007
-    reference_title: "Genome-Wide Profiling of Lesional and Non-Lesional Skin from Atopic Dermatitis, Psoriasis, and Contact Dermatitis Skin"
-    supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
-    snippet: "Genome-Wide Profiling of Lesional and Non-Lesional Skin from Atopic Dermatitis, Psoriasis, and Contact Dermatitis Skin"
-    explanation: This GEO record supports a disease-relevant comparative skin transcriptomic dataset that explicitly includes contact dermatitis samples.
-  notes: "Dataset record: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE153007"
+  notes: >-
+    Dataset record: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE153007
+    Carries no evidence block: the GEO record has no summary (its summary field
+    repeats the title) and no linked publication, so there is no abstract
+    sentence to quote. The title, conditions and organism above are GEO's own
+    values.
 - accession: geo:GSE215457
   title: Inhibition of DYRK1B suppresses inflammation in allergic contact dermatitis model and Th1/Th17 immune response
   description: >-
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -2125,7 +2125,7 @@ def test_dataset_accession_prefix_and_shape(filepath):
 # PRs still carry edits to it; deleting it now would conflict with all of them.
 # Nothing may read or write it: dataset verification moved to per-record files
 # under references_cache/, which two PRs can add to without colliding.
-FROZEN_DATASET_CACHE = "cache/dataset_accessions" + ".json"
+FROZEN_DATASET_CACHE = "cache/dataset_accessions.json"
 
 
 def test_no_automation_touches_the_frozen_dataset_cache():
@@ -2157,6 +2157,9 @@ def test_no_automation_touches_the_frozen_dataset_cache():
     ]
     offenders = []
     for path in scanned:
+        # This file names the path in FROZEN_DATASET_CACHE, so it must exclude
+        # itself; that exclusion is the only guard, which is why the constant
+        # above is a plain literal rather than a concatenation.
         if not path.is_file() or path.resolve() == Path(__file__).resolve():
             continue
         try:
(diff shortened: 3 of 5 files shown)


cmungall:
All four review findings are fixed in 7213ea44a. Each was checked against NCBI before changing, and one of them needs a correction to the review's premise.

## Wrong accession (critical) — fixed

`geo:GSE152382` → `geo:GSE159771` in `Spaceflight_Associated_Neuro-Ocular_Syndrome`.

NCBI `elink` from `PMID:34179686` resolves to `GSE159771`, "Illumina NGS RNA-seq and RRBS Sequencing on Spaceflight Mouse Retina" — mouse, 32 samples, methylome plus transcriptome. That matches the record's `description`, `data_type: METHYLATION`, and `publication` exactly. Three fields agreed and one did not, so the accession was the thing to change; nothing else in the record moved. `references_cache/GEO_GSE152382.md` is removed, since nothing in `kb/` cites it any more.

Worth noting why this was invisible before: the record's old paraphrased title was consistent with its description, so nothing pointed at the accession. Correcting the title to GEO's own is what made the record contradict itself.

## Organism — two fixed as reported, one is a mixed series

`GSE71912` and `GSE113251` in `Left_Ventricular_Noncompaction`: confirmed. NCBI reports both with taxon `Mus musculus` and no human samples. `Homo sapiens` was simply wrong. Both now `NCBITaxon:10090`.

`GSE330200` in `Dilated_Cardiomyopathy_1A`: **the premise is not quite right.** NCBI reports this series under *two* taxa — `Homo sapiens; Mus musculus` — so `Homo sapiens` was incomplete rather than wrong, and the record's own `notes` already said "Mouse and human series for the LMNA p.S143P variant".

I changed it anyway, because `organism` is single-valued and declaring only human under a title ending "in a knock-in mouse model" reads as a contradiction. It now records `Mus musculus`, and `notes` says explicitly that GEO lists both taxa and that the human samples are not represented by that slot. Flagging it because "should be `NCBITaxon:10090 / Mus musculus`" would be the wrong rule to generalise from — a mixed-taxon series has no fully correct single value, and the honest fix is to record the loss.

Descriptions were left alone throughout: they characterise the series without asserting a species, so none contradicted the corrected data.

## Minor

The PR body cites design decision **6c**, not 6b. It was 6b when written; `main` took `6b` for the supports/directness decision (#7439) during the rebase, so this branch's section was renumbered and the body and `docs/dataset-curation.md` cross-reference follow.

## On the title diffs

The review counts 21 title diffs as "pure typographic normalization". 32 records changed, and most are not typographic — `Left ventricular noncompaction cardiomyopathy gene-expression dataset GSE71912` (a placeholder) became `Mutations in the NOTCH pathway regulator MIB1 cause left ventricular noncompaction cardiomyopathy`, and `Nr2f1 chromatin occupancy in mouse cortex` became `Nr2f1 shapes mitochondria in the mouse brain unraveling novel insights into the neurodevelopmental disorder BBSOAS`. The two mouse-organism errors and the wrong accession were all found by reading those rewrites, which is the case for the check.

## Verification after the fixes

- 966/966 `geo:` dataset titles match their cached record
- 934 `geo:` accessions, 0 without a cache file
- schema validation, duplicate-key, entity-ref and history-schema checks clean over the three entries and three new history records

CI's `test (3.13)` was failing on the curation-scope guard, not a test — this PR touches `kb/` and infra together by design. `scope-override` has been applied per the check's own instruction.

---

Issue #7548, opened 2026-08-01
schema: ICD-O morphology codes are not representable (ICDOMorphologyEnum too coarse; DiseaseMappings has no ICD-O slot)


kevinschaper:
Escalated from the review of PR #7484 (Glomus Tumor, MONDO:0018327), where the automated reviewer endorsed raising this as a schema issue rather than a curation shortcut.

## Problem

There is nowhere in the schema to record a four-digit ICD-O morphology code.

- `classifications.icdo_morphology` is bound to `ICDOMorphologyEnum`, which has only ten coarse values: Carcinoma, Adenocarcinoma, Squamous Cell Carcinoma, Sarcoma, Leukemia, Lymphoma, Multiple Myeloma, Melanoma, Glioma, Embryonal Neoplasm.
- `DiseaseMappings` supports only `icd10cm`, `icd11f`, `mondo` and `ncit` — no ICD-O.

## Worked example that exposed it

`MONDO:0018327` (glomus tumor) carries the xref `ICDO:8711/0` — glomus tumour, **benign**. Neither slot can hold it:

- `8711/0` has no place in `DiseaseMappings`.
- Forcing it into `ICDOMorphologyEnum` would mean tagging it `Sarcoma`, which is actively wrong: the great majority of glomus tumours are benign, and only the rare malignant variant (glomangiosarcoma, `8711/3`) is sarcomatous.

The consequence is that `kb/disorders/Glomus_Tumor.yaml` deliberately omits its `classifications` block rather than populate it incorrectly, and records the gap as a `CURATION_TODO` discussion (`glomus-icdo-morphology-unrepresentable`).

## Why it generalises

The same problem applies to every entity whose ICD-O behaviour code matters (`/0` benign, `/1` uncertain or borderline, `/2` in situ, `/3` malignant), and to every morphology outside the ten coarse buckets — pericytic and perivascular tumours, most benign soft tissue neoplasms, and the borderline categories generally. ICD-O behaviour is exactly the information a coarse "is it a sarcoma" enum throws away.

## Options

1. Add an `icdo` slot to `DiseaseMappings` holding the full `NNNN/N` morphology code (smallest change, keeps the coarse enum for classification).
2. Replace / supplement `ICDOMorphologyEnum` with a dynamic enum over an ICD-O source so real morphology terms can be bound and validated.
3. Both: `DiseaseMappings.icdo` for the code, coarse enum retained for faceting.

[...]

PR #10163, opened 2026-08-30
Expand ICD-O morphology enum: 13 new families, 31 entry corrections

cmungall:
## Summary

Completeness review and expansion of `ICDOMorphologyEnum` against neoplastic entries in the knowledge base. The original 10-value enum was malignant-only (all `/3` behaviour codes), leaving 133 of ~280 neoplastic entries unclassified or misclassified. This PR adds 13 behaviour-neutral and behaviour-split morphology families, corrects 31 entry classifications, and establishes a rule-based governance model for the enum.

## Key Changes

**Schema (`icdo_morphology.yaml`)**
- Expanded enum description with two governing rules: granularity (lump/split rule for morphology families) and behaviour (behaviour-neutral where possible, following ICD-O splits where it does)
- Added 13 new values: `Adenoma`, `Trophoblastic Tumor`, `Mesothelial Neoplasm`, `Pericytic Neoplasm`, `Nerve Sheath Neoplasm`, `Meningioma`, `Germ Cell Tumor`, `Sex Cord-Stromal Tumor`, `Neuroendocrine Neoplasm`, `Plasma Cell Neoplasm`, `Myeloproliferative Neoplasm`, `Myelodysplastic Syndrome`, `Histiocytic and Dendritic Cell Neoplasm`
- Each new value includes NCIT binding, MONDO exact mappings, and ICD-O-3 codes where the NCIT concept itself carries them
- Corrected NCIT:C3234 (Mesothelioma) retirement: bound to NCIT:C3786 instead

**Knowledge Base (31 entries)**
- **9 corrections** (wrong histogenesis): Choriocarcinoma, Embryonal Carcinoma, Malignant Germ Cell Tumor of Ovary, Malignant Mesothelioma, Mixed Germ Cell Tumor, Polycythemia Vera, Essential Thrombocythemia, Primary Myelofibrosis, Testicular Seminoma
- **22 new classifications** (previously unclassified): Adrenal Cortex Adenoma, Glomus Tumor, GNAS-related pituitary adenoma 3, Gestational Trophoblastic Neoplasm, Hydatidiform Mole, Meningioma, Multiple Myeloma, Myelodysplastic Syndrome, Myeloproliferative Neoplasm Unclassifiable, Pancreatic Neuroendocrine Tumor, Pheochromocytoma-Paraganglioma, Schwannoma, Neurofibroma, Granular Cell Tumor, Langerhans Cell Histiocytosis, Rosai-Dorfman Disease, Adult Granulosa Cell Tumor of Ovary, Testicular Sex Cord-Stromal Neoplasm, Testicular Germ Cell Tumor, Yolk Sac Tumor, Plasma Cell Neoplasm
- Each entry includes notes explaining the binding and rationale

**Documentation (`.claude/skills/disease-classification/SKILL.md`)**
- Added dedicated ICD-O morphology section with value groupings (epithelial, mesenchymal, neural/meningeal, melanocytic, germ cell/gonadal stromal, neuroendocrine, haematolymphoid, embryonal)
- Documented the omit-and-explain convention: when no value fits, omit the slot and record why in notes or a CURATION_TODO discussion

**Review Report (`docs/reports/icdo-morphology-enum-review-2026-08-27.md`)**
- Comprehensive 176-line report documenting the finding (vocabulary incompleteness), the rule now written into the enum, all 13 values added with KB entries they unblock, and the 31 entries changed

- **Behaviour-neutral values** (e.g., `Nerve Sheath Neoplasm`, `Pericytic Neoplasm`) cover benign and malignant members alike, unblocking entries like

https://claude.ai/code/session_01S1ZoUCzSM7GtiLHKUa6oGB

### DIFF
--- a/docs/reports/icdo-morphology-enum-review-2026-08-27.md
+++ b/docs/reports/icdo-morphology-enum-review-2026-08-27.md
@@ -62,7 +62,7 @@ recorded **only** where the bound NCIT concept itself carries an
 | Value | NCIT | ICD-O-3 | KB entries it unblocks |
 |---|---|---|---|
 | `Adenoma` | C2855 | 8140/0 | adrenal cortical adenoma, the five pituitary adenoma entries, mucinous cystadenoma, the adenomatous polyposis entries |
-| `Trophoblastic Tumor` | C3422 | — | choriocarcinoma, gestational trophoblastic neoplasm, hydatidiform mole |
+| `Trophoblastic Tumor` | C3422 | — | choriocarcinoma, gestational trophoblastic neoplasm |
 | `Mesothelial Neoplasm` | C3786 | — | pleural and peritoneal mesothelioma |
 | `Pericytic Neoplasm` | C6528 | — | glomus tumour |
 | `Nerve Sheath Neoplasm` | C4972 | — | schwannoma, neurofibroma, granular cell tumour, MPNST, the neurofibromatosis entries |
@@ -125,10 +125,10 @@ both under NCIT:C3708; the children of NCIT:C3264 (Embryonal Neoplasm) are the
 blastomas, the CNS embryonal tumours, Ewing sarcoma/pPNET, rhabdoid tumour and
 Wilms tumour. The old value was a name collision, not a classification.
 
-**Twenty-two exemplars** backfilled to exercise each new value, at least one per
+**Twenty-one exemplars** backfilled to exercise each new value, at least one per
 value, each carrying a `notes` line recording the NCIT ancestry it rests on:
 adrenal cortex adenoma, GNAS-related pituitary adenoma 3, gestational
-trophoblastic neoplasm, hydatidiform mole, malignant peritoneal mesothelioma,
+trophoblastic neoplasm, malignant peritoneal mesothelioma,
 glomus tumour, schwannoma, neurofibroma, granular cell tumour, meningioma,
 yolk sac tumour, testicular germ cell tumour, adult granulosa cell tumour of
 ovary, testicular sex cord-stromal neoplasm, pheochromocytoma-paraganglioma,
@@ -136,9 +136,28 @@ pancreatic neuroendocrine tumour, plasma cell neoplasm, multiple myeloma,
 myeloproliferative neoplasm unclassifiable, myelodysplastic syndrome,
 Langerhans cell histiocytosis, Rosai-Dorfman disease.
 
-Coverage went from 133 to 155 entries, and every one of the 23 values is now
-used by at least one entry — `Multiple Myeloma`, unused since the enum was
-written, went to the `Multiple_Myeloma` entry.
+**And one exemplar withdrawn.** `Hydatidiform_Mole` was assigned
+`Trophoblastic Tumor` in an earlier draft of this PR and the assignment has
+been removed, because it failed the same rule every other value passes.
+Hydatidiform mole does carry ICD-O morphology codes — 9100/0 for the mole NOS,
+9103/0 for the partial mole — which is what made it look assignable. But
+`NCIT:C3110` (Hydatidiform Mole) does not descend from `NCIT:C3422`
+(Trophoblastic Tumor): its parents are *Gestational Trophoblastic Disorder* and
+*Placenta Disorder*, NCIT asserts no `Neoplastic_Status` for it, and both
+`NCIT:C4871` (Complete) and `NCIT:C4293` (Partial Hydatidiform Mole) sit under
+*Placental Non-Neoplastic Disorder*. WHO agrees — gestational trophoblastic
+*neoplasia* is invasive mole, choriocarcinoma, PSTT and ETT, and a non-invasive
+mole is GTD but not GTN. This axis applies only to neoplastic diseases, so the
+entry now omits the slot and records why, which is the convention this review
+argues for. The molar pregnancies that do progress are covered by
+`Gestational_Trophoblastic_Neoplasm`, which carries the value correctly.
+
+Every one of the 23 values is now used by at least one entry — `Multiple
+Myeloma`, unused since the enum was written, went to the `Multiple_Myeloma`
+entry. The KB now carries 158 `icdo_morphology` assignments in total. (Note
+that the raw before/after pair is not a clean measure of this PR: the branch
+was rebased mid-review and picked up neoplastic entries added on `main` in the
+meantime, so the denominator moved underneath it.)
 
 The prose notes on Glomus Tumor, Choriocarcinoma,
 Pheochromocytoma-Paraganglioma and GNAS-related pituitary adenoma 3 were
--- a/history/disorders/Hydatidiform_Mole/2026-08-30T160541Z-claude-code-41257a.yaml
+++ b/history/disorders/Hydatidiform_Mole/2026-08-30T160541Z-claude-code-41257a.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hydatidiform_Mole
+  path: kb/disorders/Hydatidiform_Mole.yaml
+session:
+  id: 2026-08-30T160541Z-claude-code-41257a
+  timestamp: '2026-08-30T16:05:41Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/7548
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - classifications
+  summary: 'Withdraw: Hydatidiform_Mole icdo_morphology (Trophoblastic Tumor) removed
+    as non-neoplastic'
+  details: |
+    Removes the classifications.icdo_morphology = 'Trophoblastic Tumor' assignment added earlier in the same pull request (monarch-initiative/dismech#10163), and records the reasoning in the entry's notes instead. Raised as a judgement call by the automated reviewer and confirmed against NCIt: hydatidiform mole carries ICD-O morphology codes (9100/0 for the mole NOS, 9103/0 for the partial mole), which is what made the assignment look correct, but NCIT:C3110 (Hydatidiform Mole) does not descend from NCIT:C3422 (Trophoblastic Tumor). Its parents are Gestational Trophoblastic Disorder and Placenta Disorder; NCIt asserts no Neoplastic_Status for it; and NCIT:C4871 (Complete Hydatidiform Mole) and NCIT:C4293 (Partial Hydatidiform Mole) both sit under Placental Non-Neoplastic Disorder. WHO reserves gestational trophoblastic neoplasia for invasive mole, choriocarcinoma, PSTT and ETT, so a non-invasive mole is gestational trophoblastic disease but not GTN. ICDOMorphologyEnum applies only to neoplastic diseases, so the assignment failed the same NCIT-descent rule every other value in this tranche passes. Omitting the slot and recording why is the convention this pull request argues for, applied to its own work. The molar pregnancies that do progress are covered by the Gestational_Trophoblastic_Neoplasm entry, which carries Trophoblastic Tumor correctly. Validated with linkml-validate against the Disease class.
--- a/kb/disorders/Choriocarcinoma.yaml
+++ b/kb/disorders/Choriocarcinoma.yaml
@@ -2465,13 +2465,13 @@ classifications:
     - reference: PMID:40588865
       reference_title: "Gestational Choriocarcinoma: A Timely Review of Diagnostic Pathology."
       supports: SUPPORT
+      directness: INDIRECT
       evidence_source: HUMAN_CLINICAL
       snippet: "Gestational choriocarcinoma is the most common form of gestational \ntrophoblastic neoplasm."
       explanation: >-
-        Confirms choriocarcinoma as a trophoblastic neoplasm; marked PARTIAL
-        because the source states the tumour class rather than the ICD-O
-        morphology code itself, which is taken from the MONDO ICDO:9100/3
-        cross-reference.
+        Confirms choriocarcinoma as a trophoblastic neoplasm. INDIRECT because
+        the source states the tumour class rather than the ICD-O morphology
+        code itself, which is taken from the MONDO ICDO:9100/3 cross-reference.
     notes: >-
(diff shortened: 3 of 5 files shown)


cmungall:
Rebased onto current `main` — the branch was conflicting and is now clean. Two conflicts, both mechanical: `mkdocs.yml` (main added five newer reports to the same nav list; kept all of them, mine slotted in by date) and `cache/ncit/terms.csv` (both sides appended a row at the same offset; kept both, and `just check-cache-order` confirms canonical ordering).

While rebasing I noticed `docs/reports/cancer-taxonomy-granularity-review-2026-08-28.md` landed on main. It answers the adjacent question — which cancer concepts get a `Disease` entry at all — where this PR is about the morphology annotation those entries carry. They don't overlap, so I added a cross-link between the two reports rather than changing anything.

Re-validated on the rebased tree: `linkml-validate` on all 31 changed entries, `validate-terms-schema`, `check-cache-order`, `check-term-cache-integrity`, `check-duplicate-keys` (4,251 files) and `check-entity-refs` (2,774 files) all pass. Before the rebase, `pytest tests/test_data.py` passed in full (43,611 tests) along with the folded-hyphen, snippet-length, title-snippet, snippet-grading and environmental-evidence gates.

Two decisions in here that a reviewer may want to push back on, both argued in the report:

1. **No `Other` value**, which #7548 suggested adding. The four entries that documented this gap in prose — Glomus Tumor, Choriocarcinoma, Pheochromocytoma-Paraganglioma, GNAS-related pituitary adenoma 3 — are the only reason this review happened. An `Other` bucket would have absorbed exactly those signals. The omit-and-explain rule is written into the enum description and the `disease-classification` skill instead, and the report lists the families still without a home so the signal is not lost.

2. **`src/dismech/datamodel/dismech_pydantic.py` was not regenerated.** It is already about 12,500 diff lines out of date against main's schema, so regenerating it here would bury this change. Nothing in the repo imports it and no test checks it, but it wants its own PR.

One correction to the research on #7548 worth flagging: NCIT:C3234 (Mesothelioma), which that issue proposed binding, is a retired concept — `conceptStatus = Retired_Concept`, parented under "Retired Concept 2023". `Mesothelial Neoplasm` is bound to NCIT:C3786 instead, which is active and sits directly under NCIT's *Neoplasm by Morphology* axis.

Also, the PR description's last bullet is cut off mid-sentence ("unblocking entries like"). Happy to finish it if you want.

---

cmungall:
`test (3.13)` was red, but not on a test — the job stopped at the curation-scope guard before running any validation:

```
.claude/skills/disease-classification/SKILL.md
docs/reports/icdo-morphology-enum-review-2026-08-27.md
mkdocs.yml
src/dismech/schema/classifications/icdo_morphology.yaml
If intentional, add the 'scope-override' label to this PR.
```

That is the intentional case, so I added the `scope-override` label. An enum extension is inherently both halves at once — the schema value and the KB entries that exercise it — and splitting them into two PRs is worse, not better: the KB half cannot validate until the schema half has merged. Note the label doesn't weaken the job; it lets it get past the guard and actually run the validation, which it never reached.

I could not re-run the failed job (403 from the API), and the workflow triggers on `push`, not `label`, so the label alone would not have started a new run. The push above carries a real change — the report now records the scope-override requirement so the next enum extension doesn't rediscover it — and CI is re-running against the labelled PR.

If you'd rather this were split, say so and I'll do it; the label is removable and the guard will fail again on the next push.

**On the reviewer's open checklist item** ("Verify NCIT parentage claims made in entry notes") — I re-checked all of them against the NCI EVS API. They hold. The germ cell chain, which is the load-bearing one for four of the nine corrections:

- `C4925` Malignant Germ Cell Tumor → `C3708` Germ Cell Tumor
- `C121619` Nongerminomatous Germ Cell Tumor → `C3708`
- `C3752` Embryonal Carcinoma → both of the above
- `C3011` Yolk Sac Tumor → both of the above
- `C4290` Mixed Germ Cell Tumor → `C4925`
- `C7328` Testicular Seminoma → `C9063` Malignant Testicular Germ Cell Tumor → `C4925`
- `C8591` Testicular Germ Cell Tumor → `C3708` directly

None of them descends from `C3264` Embryonal Neoplasm, whose children are the blastomas, the CNS embryonal tumours, Ewing sarcoma/pPNET, rhabdoid tumour and Wilms tumour. That is the mistake the old values encoded.

Two codes in my first draft of those notes were wrong and were corrected before the initial push: `C9061` is Penile Carcinoma, not Testicular Seminoma, and `C3595` is Benign Nasopharyngeal Neoplasm, not Testicular Germ Cell Tumor. The committed notes carry `C7328` and `C8591`.

---

cmungall:
The `Mesothelial Neoplasm` → MONDO finding is real. Fixed in d72ac016e.

`MONDO:0005065` "mesothelioma" was wrong twice over. It still cross-references the **retired** `NCIT:C3234`, and its definition is "a **usually malignant** and aggressive neoplasm of the mesothelium" — a behaviour claim, on a value whose whole point is being behaviour-neutral. So the value paired a behaviour-neutral `meaning` with a malignant-leaning `exact_mappings`, which is incoherent regardless of the retirement.

The right term already existed: **`MONDO:0006856` "mesothelial neoplasm"**, which cross-references `NCIT:C3786` — the code this value's `meaning` binds — and is defined as "a benign or malignant neoplasm arising from mesothelial cells". Swapped.

I then audited the other twelve values the same way, since if I got one wrong the check itself was missing rather than the single term. The test is: **does the MONDO term carry a `database_cross_reference` to the same NCIT code the value's `meaning` binds?**

| Value | `meaning` | MONDO | MONDO's NCIT xref |
|---|---|---|---|
| Adenoma | C2855 | MONDO:0004972 | `NCIT:C2855` |
| Germ Cell Tumor | C3708 | MONDO:0005040 | `NCIT:C3708` |
| Trophoblastic Tumor | C3422 | MONDO:0002872 | `NCIT:C3422` |
| Neuroendocrine Neoplasm | C3809 | MONDO:0019496 | `NCIT:C188218`, `NCIT:C3809` |
| Mesothelial Neoplasm | C3786 | ~~MONDO:0005065~~ → MONDO:0006856 | ~~`NCIT:C3234`~~ → `NCIT:C3786` |
| Sex Cord-Stromal Tumor | C3794 | MONDO:0006055 | `NCIT:C3794` |
| Nerve Sheath Neoplasm | C4972 | MONDO:0002547 | `NCIT:C4972` |
| Meningioma | C3230 | MONDO:0016642 | `NCIT:C3230` |
| Pericytic Neoplasm | C6528 | MONDO:0002604 | `NCIT:C6528` |
| Myeloproliferative Neoplasm | C4345 | MONDO:0020076 | `NCIT:C4345` |
| Myelodysplastic Syndrome | C3247 | MONDO:0018881 | `NCIT:C3247` |
| Plasma Cell Neoplasm | C4665 | MONDO:0004959 | `NCIT:C4665` |
| Histiocytic and Dendritic Cell Neoplasm | C9294 | MONDO:0006247 | `NCIT:C9294` |

Mesothelioma was the only miss, and it was the only value where I picked the MONDO term by label rather than by following the NCIT code — which is exactly how it slipped through. That check is now written into the report so the next addition runs it.

`validate-terms-schema` and `just validate kb/disorders/Malignant_Mesothelioma.yaml` both pass on the new head. No KB entry changed: `Malignant_Mesothelioma` and `Malignant_Peritoneal_Mesothelioma` already carried the correct enum value, and only the value's own mapping was wrong.

One thing to note about running `just validate` locally: it refreshed three unrelated `references_cache/` files (a PMC full-text upgrade on `PMID_21412769` and two smaller frontmatter changes). I reverted them rather than carry them here — they're legitimate validator output but nothing to do with this change.

---

cmungall:
Thanks — approved, and I've taken the main suggestion. Fixed in 9ff60db61.

`icdo_morphology.yaml` declared only `NCIT` while using `MONDO:` and `ICDO:` in `exact_mappings`, so neither expanded. Both are now declared, matching how sibling `mechanistic_nosology.yaml` handles `MONDO`. Verified through SchemaView on the built schema:

```
expand ICDO:8140/0   -> https://bioregistry.io/icdo:8140/0
expand MONDO:0006856 -> http://purl.obolibrary.org/obo/MONDO_0006856
```

Two things about `ICDO` that are worth having in the file rather than only in a PR thread, so they're in an inline comment there:

- **It has no OBO PURL.** ICD-O is WHO-copyrighted and isn't redistributable as an ontology artefact, so it expands through Bioregistry rather than `obo/ICDO_`. This is the same reason a dynamic enum over real ICD-O terms isn't available (#7548 §4c).
- **Its local identifiers contain a slash** — `ICDO:8140/0`. That is a legal CURIE, since only the prefix is NCName-constrained, but it will not round-trip through an `obo/`-style expansion, which is the trap someone will otherwise hit later.

One YAML detail in case it bites elsewhere: the Bioregistry expansion ends in a colon, so `ICDO: https://bioregistry.io/icdo:` fails to parse. It needs quoting.

You're right that this was pre-existing for the original ten values. I fixed it here because this PR adds 18 more such CURIEs, which is what tipped it from someone else's problem to this diff's.

The other three suggestions are referenced in the review but the detail comment doesn't seem to have posted — I only see the prefix one. If you post them, or they show up, I'll work through them.

Note this push drops the approval. Nothing else changed: no KB entry, no enum value, no mapping.

---

cmungall:
All four suggestions taken. 8df5c8dc2 (plus 9ff60db61 for the prefixes, already described above).

**The Hydatidiform Mole one was not just a judgement call — the assignment was wrong, and I've withdrawn it.** You were more generous than the evidence warrants. I checked NCIt rather than defending it:

- `NCIT:C3110` (Hydatidiform Mole) does **not** descend from `NCIT:C3422` (Trophoblastic Tumor). Its parents are *Gestational Trophoblastic Disorder* and *Placenta Disorder*.
- NCIt asserts **no** `Neoplastic_Status` for it.
- `NCIT:C4871` (Complete) and `NCIT:C4293` (Partial Hydatidiform Mole) both sit under **Placental Non-Neoplastic Disorder**.

So it failed the exact NCIT-descent rule every other value in this tranche passes. The ICD-O codes (9100/0, 9103/0) are real and are what made it look assignable — the morphology axis codes plenty of things this enum doesn't cover. Combined with your WHO point (GTN is invasive mole, choriocarcinoma, PSTT, ETT; a non-invasive mole is GTD but not GTN), the honest action is to omit the slot and say why, which is the convention this PR spends its enum description arguing for. Applying it to my own work rather than to everyone else's seemed like the point.

The entry now carries that reasoning in `notes`, and there's an appended history record for the withdrawal. `Gestational_Trophoblastic_Neoplasm` still carries `Trophoblastic Tumor`, correctly — that's where the molar pregnancies that do progress live.

**Stale `marked PARTIAL`** — fixed. Now `supports: SUPPORT` + `directness: INDIRECT`, with the explanation reworded to match, per the #7439 migration table. Agreed the other 98 files are a repo-wide sweep and not this PR's business; I only touched the one sentence I had already rewritten.

**Two cosmetic drifts** — both fixed. The 180-character line in `Pheochromocytoma_Paraganglioma.yaml` is rewrapped. The report's counts were stale: it said 133 → 155, the tree showed 159 before this push and 158 after the withdrawal. I've corrected the figure and added a line noting the before/after pair isn't a clean measure of this PR, since the mid-review rebase moved the denominator by picking up neoplastic entries added on `main`. Exemplar count is now 21, not 22.

**On the `ICDO` prefix, I disagree with one part and kept it.** You suggested declaring `MONDO` but not `ICDO`, on the grounds that `ICDO:8010/3` is not a well-formed CURIE because of the slash. I think that's not right:

- Per the W3C CURIE syntax, only the *prefix* is NCName-constrained; the reference is an `irelative-ref`, which permits path segments. So the slash is legal in the local part.
- Bioregistry's own `icdo` pattern is `^[8-9]\d{3}(/[0-3])?$` — the slash is part of the registered identifier shape, not a deviation from it.
- `https://bioregistry.io/icdo:8140/0` resolves (302 → the IARC code search), so the expansion isn't hypothetical.
- Verified through SchemaView on the built schema: `ICDO:8140/0` → `https://bioregistry.io/icdo:8140/0`.

Leaving it undeclared doesn't avoid the question — the CURIEs are in the file either way, and undeclared they simply don't resolve. The file now carries an inline comment recording both caveats (no OBO PURL because ICD-O is WHO-copyrighted; won't round-trip through an `obo/ICDO_` expansion), so the next reader gets the reasoning rather than a bare prefix row. Happy to drop it if you still disagree — it's one line.

One practical note: the Bioregistry URI ends in a colon, so `ICDO: https://bioregistry.io/icdo:` fails to parse as YAML. It needs quoting.

**Follow-up issue for `dismech_pydantic.py`** — agreed it shouldn't sit at ~12,500 lines indefinitely. I'll open one unless you'd rather fold it into #7548.

Validation on this head: `linkml-validate` on the three changed entries, `validate-history` on the new record, `check-folded-hyphens`, `check-snippet-grading`, `check-title-snippets`, `check-duplicate-keys`, `check-entity-refs` — all clean. `validate-terms-schema` passed on 9ff60db61 and the schema is unchanged since.

---

cmungall:
Green and mergeable. `test (3.13)` passed on `8df5c8dc2` — every gate, including the curation-scope guard, the whole-KB schema-conformance sweep, and the full Python test lane. `mergeable_state: clean`, re-approved, no conflicts.

Two things left, both needing a decision rather than work from me.

**The PR description is stale, and that's my doing.** It lists "Hydatidiform Mole" under "22 new classifications", but the withdrawal above removed that assignment — it's now 21 exemplars plus the 9 corrections. The title's "31 entry corrections" is also off by the same one. The last bullet is still truncated at "unblocking entries like". It's your text so I haven't touched it; say the word and I'll fix all three together.

**Follow-up for `dismech_pydantic.py`.** The reviewer agreed the generated file shouldn't sit ~12,500 lines behind the schema indefinitely. Happy to open a tracking issue, or fold it into #7548 alongside the other deferred schema work — your call which.

For anyone reading this thread later, the substance in one line: the enum was malignant-only, which had forced wrong values onto committed entries and left benign and behaviour-indeterminate neoplasms unclassifiable. What is still *not* covered — thymic epithelial, GIST, mast cell, fibroblastic, PEComa, odontogenic, craniopharyngioma, benign melanocytic, neuronal-glial, benign mesenchymal — is listed in the report, along with the check (does the MONDO term xref the value's own NCIT `meaning`?) that the mesothelioma catch turned into a rule.

---

PR #10174, opened 2026-08-30
curate: Add Bone Marrow Failure Syndrome 6 (MONDO:0030015, MDM4)

### LINKED ISSUE #10170: Curate bone marrow failure syndrome 6 (MONDO:0030015)
Curate a dismech entry for **bone marrow failure syndrome 6** ([MONDO:0030015](https://monarchinitiative.org/MONDO:0030015), OMIM #618849) — an autosomal dominant inherited bone marrow failure syndrome caused by germline heterozygous loss-of-function variants in **`MDM4`**, a negative regulator of p53. Reduced MDM4 dosage releases p53, and the resulting p53 hyperactivity impairs haematopoietic stem and progenitor cell self-renewal and differentiation.

Preflight notes: MONDO:0030015 is a leaf under `MONDO:0000159` bone marrow failure syndrome with no descendants. MONDO records **no** causal gene (`RO:0004003`) for this term, which is why the stub carries no `genes:` block. `MDM4` is not currently curated anywhere in `kb/`. The numbered BMFS series (BMFS3/4/5/6, Ziegler-Huang) sits under the same parent and would be the membership of a future grouping — worth noting in the entry but not in scope here.

**Correction (2026-08-30):** an earlier version of this issue body said the gene was `MECOM` and described the disease as radioulnar synostosis with amegakaryocytic thrombocytopenia (RUSAT). That was wrong and was my error — the stub has no `genes:` block because MONDO records no causal gene for this term, and I filled the gap from memory instead of checking OMIM. MECOM-associated RUSAT is a different OMIM entity. The correct gene for BMFS6 / OMIM #618849 is `MDM4`, established by [Sharma et al., *Blood* 2026 (PMID:41758987)](https://pubmed.ncbi.nlm.nih.gov/41758987/). The title and MONDO ID were unaffected, so the claim lock held throughout.

Stub: `stubs/Bone_Marrow_Failure_Syndrome_6.yaml`

aleixpuigb:
New entry for BMFS6 (OMIM #618849), an inherited bone marrow failure syndrome caused by germline heterozygous loss-of-function variants in `MDM4`.

Closes #10170

## The mechanism runs the p53 pathway backwards

MDM4 restrains p53. Halving MDM4 therefore does not damage a maintenance pathway — it removes a brake. The pathograph is: germline heterozygous MDM4 LOF → loss of MDM4-mediated p53 restraint → **p53 hyperactivation in HSPCs** → impaired self-renewal and multilineage output → marrow failure with hypocellular MDS. The source names the category itself: MDM4 deficiency is a TP53-activating syndrome.

A second branch leaves the same central node, and it is the most interesting thing in the entry. One patient who progressed to MDS had **acquired somatic loss-of-function TP53 mutations**. Read against a germline lesion that causes p53 *hyper*activity, that is not a second hit compounding the mechanism — it is a somatic event relieving it. A clone that inactivates p53 escapes the growth disadvantage the germline variant imposes, and pays for it by giving up p53 tumour suppression. The source calls this maladaptive somatic rescue; it is curated as its own node with `TP53` recorded as `relationship_type: COOPERATING`, `variant_origin: SOMATIC`.

## Model evidence, and where I stopped short of it

Both engineered systems from the primary paper are curated as `experimental_models` with `modeled_mechanisms` and `readouts`:

- **CRISPR/Cas9 MDM4-deleted healthy-donor HSPCs** — `RECAPITULATES`, fidelity `HIGH`. Limitation recorded: engineered deletion is not the patient allelic spectrum (two of six alleles are missense), and engraftment is scored in a xenograft niche.
- **Patient-variant knock-in iPSCs** — `RECAPITULATES`, fidelity `MODERATE`. Limitation recorded: iPSC-derived haematopoiesis is developmentally distinct from adult marrow, and cannot model the decades-long clonal dynamics the somatic TP53 arm turns on.
- **p53-Δ31 mouse, Mdm4-dosage sensitized** — `animal_models`, `PARTIALLY_RECAPITULATES`, fidelity `MODERATE`. Its primary lesion is in p53, not MDM4; Mdm4 dosage enters as a modifier. That is what makes it informative for this disease and also what caps its fidelity.

**The mouse telomere findings are deliberately not curated as human edges.** The p53-Δ31 mouse has short telomeres, pulmonary fibrosis, and downregulated *Dkc1*/*Rtel1*/*Tinf2*/*Terf1* — a dyskeratosis-congenita picture, and reducing *Mdm4* makes mild heterozygotes dramatically worse, which is exactly the human axis. But the human cohort report describes no telomere length, no telomere-gene expression, and no pulmonary fibrosis. Curating that arm would import a mouse phenotype into a human pathograph, so it is recorded as a `HUMAN_MODEL_MISMATCH` discussion, and the entry declares **no** `conforms_to` against `telomere_attrition`. If someone measures telomeres in MDM4-haploinsufficient marrow, that discussion is the thing to close.

## I got the gene wrong, and the tooling did not catch it

Worth flagging explicitly rather than burying. I filed claim issue #10170 describing this disease as **MECOM**-related radioulnar synostosis with amegakaryocytic thrombocytopenia. That was wrong.

The failure chain: MONDO records no causal gene (`RO:0004003`) for MONDO:0030015, so `stubs/Bone_Marrow_Failure_Syndrome_6.yaml` carried no `genes:` block. The claim-disease skill correctly says an absent block means the ontology records nothing, not that enrichment was skipped — but it gives no guidance for the case where you then need the gene anyway, and I filled it from memory. `just preflight-dr` returned `SKIP` for exactly this reason ("MONDO records no causal gene… the gene-identity check cannot discriminate"), so the one check designed to catch named-entity confusion could not run.

What did catch it was the openscientist deep-research run, which returned `MDM4=59, MDM2=9` on gene mentions and cited [PMID:41758987](https://pubmed.ncbi.nlm.nih.gov/41758987/) — Sharma et al., *Blood* 2026, the cohort that established the gene. OMIM #618849 agrees, and the OMIM ID matched between MONDO and the report. The issue body is corrected with a dated correction note; the title and MONDO ID were never wrong, so the claim lock held throughout.

## Validation

- `just validate` — schema + terms pass
- `just count-verified-snippets` — **31/31** verified against the reference cache
- `just check-entity-refs`, `just check-duplicate-keys` — pass
- `just check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` — no new findings

Three schema corrections during curation, all caught by `just validate`: `PRIMARY_CELL`/`IPSC_DERIVED` are not the enum values (`PRIMARY_CELL_CULTURE`/`IPSC_DERIVED_MODEL` are); the Harrison's value is `ONCOLOGY_HEMATOLOGY`, not `HEMATOLOGY_ONCOLOGY`; and `GO:1901797` is a biological process, not a molecular function. I also dropped a "Variable age of onset" phenotype after the dynamic-enum check rejected `HP:0003812` — it is a clinical modifier, not a phenotypic abnormality, and the fact belongs in the marrow-failure node's description, where it now lives.

Cache rows are scoped to this entry: 3 rows (`MONDO:0030015` ×2, `hgnc:6974`), 2 new `PMID_*.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VUnkGUQjMLA8d8DoUgucDB)_

### DIFF
--- a/history/disorders/Bone_Marrow_Failure_Syndrome_6/2026-08-30T202040Z-claude-code-f84118.yaml
+++ b/history/disorders/Bone_Marrow_Failure_Syndrome_6/2026-08-30T202040Z-claude-code-f84118.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Bone_Marrow_Failure_Syndrome_6
+  path: kb/disorders/Bone_Marrow_Failure_Syndrome_6.yaml
+session:
+  id: 2026-08-30T202040Z-claude-code-f84118
+  timestamp: '2026-08-30T20:20:40Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10170
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10174
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - notes
+  - treatments
+  summary: 'Address re-review: curate the MDM2/MDMX-inhibitor contraindication'
+  details: |
+    Addresses the single outstanding item from the second review on PR 10174. Added MDM2/MDMX inhibitors (nutlins and successors) as a treatment to avoid, following the contraindication pattern used in Allan-Herndon-Dudley_Syndrome: the treatment name carries the contraindication and the description carries the reasoning. The claim is that these agents raise p53 activity by design, which is the pharmacology this disease is already suffering from, so giving one would be expected to deepen the causal lesion. Cited to PMID:28673313 for the pharmacological premise only, graded OTHER with directness INDIRECT, and the description states explicitly that no publication makes this recommendation about BMFS6, no MDM4-haploinsufficient patient has received one, and no adverse outcome has been reported. Curated rather than omitted because these drugs are in clinical evaluation for haematological malignancy, which is the setting a BMFS6 patient with hypocellular MDS could plausibly reach. Also dispositioned the OMIM clinical-synopsis cytopenias in notes as a recorded decision rather than an unexplained absence: neither primary paper states them and the OMIM synopsis is not citable in this repository's evidence model. Snippets 49/49 verified; all gates pass.
--- a/kb/disorders/Bone_Marrow_Failure_Syndrome_6.yaml
+++ b/kb/disorders/Bone_Marrow_Failure_Syndrome_6.yaml
@@ -119,6 +119,13 @@ notes: >
   values are set - a band computed across seven people from two papers with
   different ascertainment would be a scope error.
 
+  The OMIM clinical-synopsis cytopenias (anaemia, neutropenia, lymphopenia)
+  are deliberately not curated as phenotypes. Neither BMFS6 primary paper
+  states them, and the OMIM synopsis is not a citable source in this
+  repository's evidence model, so adding them would mean an evidence item with
+  no quotable origin. They are almost certainly real; they are simply not
+  citable here yet.
+
   Two care notes for future curators. First, MONDO records no causal gene
   (RO:0004003) for MONDO:0030015, so the `stubs/` entry for this disease
   carried no `genes:` block and the gene identity has to come from OMIM
@@ -130,7 +137,14 @@ notes: >
 
   `treatments:` is curated from a telomere-biology-disorder management review
   rather than from BMFS6-specific data, because neither BMFS6 primary paper
-  discusses treatment. Each treatment records that explicitly and carries
+  discusses treatment. The one exception is the MDM2/MDMX-inhibitor
+  contraindication, which is not imported from any disease class: it is derived
+  from this entry's own mechanism plus the published pharmacology of those
+  agents, and is marked INDIRECT for exactly that reason. It is curated rather
+  than omitted because these drugs are in clinical evaluation for
+  haematological malignancy, which is where a BMFS6 patient with hypocellular
+  MDS could plausibly encounter them. No publication makes this recommendation
+  about BMFS6 and the entry does not imply otherwise. Each treatment records that explicitly and carries
   `directness: INDIRECT`. The justification for importing class-level
   management is that the founding family was described as having features
   suggestive of dyskeratosis congenita with short telomeres, which places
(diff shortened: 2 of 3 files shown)


aleixpuigb:
You were right, and the correction matters more than the omission did. Pushed in 1c6ea0d.

I had PMID:32300648 cached and did not open it. The deep-research report quoted only its opening sentence — *"Dyskeratosis congenita is a cancer-prone inherited bone marrow failure syndrome caused by telomere dysfunction"* — which reads like a generic review, so I took it for background and moved on. The abstract's second sentence is the founding of this disease. That is the failure mode the references skill warns about: treating a deep-research quote as a summary of the source rather than as a pointer to it.

## The telomere arm is now curated as a human finding

New `Telomere Shortening` node downstream of p53 hyperactivation, with `conforms_to: "telomere_attrition#Progressive Telomere Attrition"`. Conformance is declared against the **attrition** node rather than the module's key target (the DNA-damage/senescence node), because what is documented in BMFS6 is the shortening itself — the senescence step would be inferred from the module, not measured in these patients.

The node carries three evidence items from the founding paper, including the causal-direction sentence, which is the one that justifies placing it downstream of p53 rather than upstream: *"a germline activation of the p53 pathway may cause telomere dysfunction"*. That inverts the usual reading where short telomeres are primary and p53 is the sensor, and it is why this entry does not simply become a telomere syndrome.

## The old discussion is replaced by two, not deleted

Your point that declining the *mouse* gene-expression arm remains defensible is what I built on:

- `bmfs6_telomere_penetrance` (`KNOWLEDGE_GAP`) — short telomeres are human, but in one family carrying one allele. The 2026 cohort of six, four of them truncating, reports no telomere measurements at all, so the genotypes that most broadened the allelic spectrum are silent on the axis the founding family established. Carries a `proposed_experiments` entry with `would_support` / `supporting_outcome` / `refuting_outcome`.
- `bmfs6_mouse_telomere_gene_program` (`HUMAN_MODEL_MISMATCH`) — narrowed to the Dkc1/Rtel1/Tinf2/Terf1 downregulation, which genuinely has no human counterpart. The shortening is documented in humans; the mechanism producing it is not.

I also added the clinical consequence to `diagnosis`: telomere length supports the diagnosis but a normal result does not currently exclude it.

## Rest of the review

| Item | Done |
|---|---|
| T454M knock-in mouse | Added at `RECAPITULATES` / `HIGH`, two linked mechanisms with readouts. Limitations record that mouse telomeres are longer and somatic telomerase wider, and that only the missense allele is modelled |
| Variable expressivity | The mouse's p53-activity-as-modifier result now cited on `inheritance`, which previously raised expressivity without explaining it |
| Missing phenotypes | Short telomeres, AML, tongue SCC added |
| MDS/leukaemia node | Now carries disease-specific AML evidence; the class-level statement stays but is labelled as such |
| Orphaned `p53_activating_bmf` | Wired onto five causal edges |
| `prevalence` | `CASES_IN_LITERATURE` / `ULTRA_RARE`, with the two cohorts kept separate rather than pooled |
| p.T454M | Added to `variants` |
| `treatments` | HSCT and danazol/oxymetholone, from PMID:35929966 |

On treatments — you were right that an absent section misrepresents the constraint, but I did not want to import class-level management silently. Both entries carry `directness: INDIRECT`, and `notes` states the basis: the founding family was described as having features suggestive of dyskeratosis congenita with short telomeres, which places BMFS6 inside the class that review covers. It is not a claim that transplant outcomes have been reported in an MDM4 carrier, and the entry says so.

**One item I did not do.** The OMIM-synopsis cytopenias (anemia, neutropenia, lymphopenia) are not added. Neither cached abstract states them, and the OMIM clinical synopsis is not a citable source in this repo's evidence model — adding them would have meant an evidence item quoting something I cannot quote. If you think a `notes` line recording them as expected-but-uncited is better than absence, I will add that instead.

Snippets 48/48 verified. Schema, entity-refs, duplicate-keys and the four baseline gates all pass.

---

aleixpuigb:
Added rather than declined. Pushed in 5451fd25.

You were right that it should not stay silent, and right about why it matters more than a generic missing treatment: it is the direct clinical consequence of the entry's own organising claim. These agents raise p53 activity *by design* — that is the intended pharmacology, not a side effect — and this is a disease of p53 hyperactivity caused by losing half the MDM4 dose.

The entry follows the `Allan-Herndon-Dudley_Syndrome` pattern you pointed at: the treatment name carries the contraindication, the description carries the reasoning.

**On how far the claim is allowed to reach.** The description states the limits rather than implying them: no publication makes this recommendation about BMFS6, no MDM4-haploinsufficient patient has received one of these agents, and no adverse outcome has been reported. PMID:28673313 supplies only the pharmacological premise, graded `OTHER` with `directness: INDIRECT`. The contraindication follows from that premise plus this entry's mechanism, in one inference step.

The reason to curate it rather than leave it implicit is that the situation is foreseeable, not hypothetical: MDM2/MDMX inhibitors are in clinical evaluation for haematological malignancy, BMFS6 patients present with hypocellular MDS and progress to leukaemia, and one reported patient acquired somatic TP53 loss. An MDM4-haploinsufficient person reaching an oncology setting where one of these is on the menu is a plausible path.

**OMIM cytopenias.** You were right that the reasoning should live where a reader finds it, not only in a commit message. Now dispositioned in `notes`: neither primary paper states them, the OMIM clinical synopsis is not citable in this repository's evidence model, and they are almost certainly real but simply not citable here yet.

**The two optional suggestions** I have left. The p.T454M RING-domain mapping (UniProt) and the ClinVar VUS lag are both quotable only from database queries rather than publications, which is the same constraint that kept gnomAD constraint metrics out of the GINS1 entry. If there is an accepted pattern in this repo for citing UniProt domain boundaries or ClinVar classification state as evidence, both are worth adding and I would use it.

49/49 snippets verified. Schema, terms, entity-refs, duplicate-keys and the four baseline gates all pass.

---

PR #10176, opened 2026-08-30
curate: Add Immunodeficiency 64 (MONDO:0032803, RASGRP1)

### LINKED ISSUE #10172: Curate immunodeficiency 64 (MONDO:0032803)
Curate a dismech entry for **immunodeficiency 64** ([MONDO:0032803](https://monarchinitiative.org/MONDO:0032803)) — an autosomal recessive combined immunodeficiency caused by biallelic loss-of-function variants in `RASGRP1`, the T-cell RAS guanine nucleotide exchange factor downstream of the TCR, presenting with CD4 lymphopenia, impaired T-cell proliferation and EBV-driven lymphoproliferation or lymphoma.

Preflight notes: MONDO:0032803 is a leaf with a single causal gene (`hgnc:9878` RASGRP1) and no descendants — this looks like a `DISEASE`. Neither `RASGRP1` nor "immunodeficiency 64" appears anywhere in `kb/`. The EBV-susceptibility IEIs already in the KB (XIAP/SH2D1A XLP, CD27/CD70, MAGT1) are the natural comparison set for the lymphoproliferation arm.

Stub: `stubs/Immunodeficiency_64.yaml`

aleixpuigb:
New entry for IMD64 / RASGRP1 deficiency (OMIM #618534), the autosomal recessive combined immunodeficiency with immune dysregulation caused by biallelic loss-of-function variants in `RASGRP1` — the calcium- and DAG-regulated RAS exchange factor that couples the T cell receptor to RAS.

Closes #10172

## Two arms, not one chain

The thing worth reviewing here is the shape of the pathograph. RASGRP1 deficiency is usually described as a RAS-MAPK signalling defect, and that is the larger half, but it does not account for the NK phenotype. So the entry curates two parallel arms leaving the same causal node:

**Catalytic arm** — collapse of TCR→RAS→ERK signalling (reduced RAS-GTP loading, reduced ERK phosphorylation in T *and* B cells, restored by wild-type RASGRP1) → defective antigen-driven proliferation and activation → **failed expansion of EBV-specific CD8+ T cells** → EBV-driven B cell lymphoproliferation and lymphoma. That third step is what makes IMD64 an EBV disease specifically rather than a generic combined immunodeficiency, and it is the step that puts it in the same family as MAGT1 and ITK.

**Cytoskeletal arm** — loss of the RASGRP1–DYNLL1 interaction → decreased RhoA activation → **impaired NK cytotoxic synapse assembly** (defective granule convergence and actin accumulation). This is a synapse-assembly failure, not an NK number deficit, and it does not follow from reduced ERK.

The evidence that these are parallel rather than serial is pharmacological, and it is why the entry commits to the two-arm shape: **lenalidomide raises RhoA activity and reverses the migration and activation defects** without addressing the RAS exchange lesion. If the cytoskeletal phenotype were downstream of ERK, a RhoA-directed intervention should not rescue it. Recorded as a `mechanistic_hypotheses` entry at `status: EMERGING`, since it rests on one paper.

## Lenalidomide is curated, and hedged

`treatments` includes lenalidomide with `target_mechanisms` pointing at the cytoskeletal node — but the entire evidence base is an *ex vivo* rescue in patient-derived cells, and the 2026 systematic review states that transplantation remains the only curative therapy. I have attached a `KNOWLEDGE_GAP` to the treatment and to the mechanism node saying exactly that. A mechanism-based rescue in patient cells is precisely the kind of finding that gets cited downstream as though it were a therapy, and the entry should not be the thing that enables that.

## Named Entity Confusion hazard, recorded rather than tripped over

Common regulatory variants at the RASGRP1 locus are replicated GWAS susceptibility alleles for East Asian SLE, rheumatoid arthritis, Hashimoto thyroiditis, and IgA nephropathy. The deep-research report surfaced all of them. They are a **different genetic claim** from the rare biallelic nulls that cause IMD64 — population-level dosage modifiers vs. Mendelian loss of function — and a common-variant association at a locus is not evidence about a Mendelian mechanism. None of that literature is cited as evidence for any node. The hazard is recorded in `notes` and in the gene's `review_notes` so the next person searching this gene does not have to rediscover it.

## Validation, including two things it caught

- `just validate` — schema + terms pass
- `just count-verified-snippets` — **33/33** verified against the reference cache
- `just check-entity-refs`, `just check-duplicate-keys` — pass
- `just check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` — no new findings

Two real defects were caught before commit, both worth naming:

1. **Three fabricated `reference_title` values.** I had taken the titles from the deep-research report's citation labels, which paraphrase, rather than from the reference cache. The reference validator rejected all three with the true titles. This is a good argument for the title check existing — the PMIDs were right, so nothing else would have flagged it.
2. **One `evidence_source` divergence.** The same quoted sentence was graded `HUMAN_CLINICAL` in the classifications block and `IN_VITRO` in two others. It reports assays on patient cells, so `IN_VITRO` is correct everywhere.

Cache rows and reference files are scoped to this entry: 5 cache rows, 6 new `PMID_*.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VUnkGUQjMLA8d8DoUgucDB)_

### DIFF
--- a/history/disorders/Immunodeficiency_64/2026-08-30T200209Z-claude-code-ca95ac.yaml
+++ b/history/disorders/Immunodeficiency_64/2026-08-30T200209Z-claude-code-ca95ac.yaml
@@ -1,0 +1,35 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Immunodeficiency_64
+  path: kb/disorders/Immunodeficiency_64.yaml
+session:
+  id: 2026-08-30T200209Z-claude-code-ca95ac
+  timestamp: '2026-08-30T20:02:09Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10172
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10176
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - prevalence
+  - pathophysiology
+  - discussions
+  - phenotypes
+  - biochemical
+  - diagnosis
+  - treatments
+  - notes
+  summary: 'Address review: connect orphaned node, add frequencies, mine the cached
+    full text'
+  details: |
+    Addresses the CHANGES_REQUESTED review on PR 10176. The review's framing was right: nothing in the entry was wrong, but it under-consumed PMID:42253627, which this PR already commits as full text. Changes: connected the orphaned Immune Dysregulation with Autoimmunity and Vasculopathy node by adding an incoming edge from Collapse of TCR-to-RAS-ERK Signal Transduction, cited to the source sentence tying autoimmunity to impaired B-cell tolerance from the same lesion; added four phenotypes# downstream targets so the mechanism graph now reaches the clinical graph. Added frequency bands to four phenotypes (OBLIGATE for infections, VERY_FREQUENT for lymphoproliferation and autoimmunity, FREQUENT for lymphoma) each backed by its own cohort-percentage evidence item. Added seven phenotypes: DVT, splenomegaly, lymphadenopathy, ITP, TTP, pan-T-cell lymphopenia and Hodgkin lymphoma; the bundled vasculopathy phenotype was renamed and the venous event split out. Added a biochemical block for the antiphospholipid antibody panel. Added a prevalence CASES_IN_LITERATURE record with cohort size, family count, ancestry and median onset. Added two treatments - conservative immunosuppression, with its 45% mortality curated as REFUTE against it being adequate definitive management, and valganciclovir prophylaxis with EBV surveillance - and added the four-of-four remission outcome to the HSCT entry. Added coagulation screening to diagnosis. Added the CTPS1 and CD27-CD70 mechanism detail to the EBV node, which is the concrete reason this disease resembles the narrow-spectrum EBV IEIs. Added a HUMAN_MODEL_MISMATCH recording that mice are not susceptible to EBV, so the defining human arm has no animal counterpart while the autoimmune arm does. Regraded PMID:31402499 from HUMAN_CLINICAL to OTHER as a narrative review. TTP and DLBCL have no HPO terms; TTP is bound to the closest accurate parent with the specificity in preferred_term, DLBCL recorded in the lymphoma description. Snippets 54/54 verified; all gates pass.
--- a/kb/disorders/Immunodeficiency_64.yaml
+++ b/kb/disorders/Immunodeficiency_64.yaml
@@ -111,9 +111,13 @@ inheritance:
       child is the classic autosomal recessive configuration.
 
 notes: >
-  About fifteen patients reported. Where this entry gives a percentage it comes
-  from the 2026 systematic review of 14 previously reported cases plus one new
-  patient, and the denominator is that aggregated cohort - not a population.
+  Fifteen patients from thirteen unrelated families. Where this entry gives a
+  percentage it comes from the 2026 systematic review of 14 previously reported
+  cases plus one new patient, and the denominator is that aggregated cohort -
+  not a population. Median age of symptom onset is 12 months; onset is recorded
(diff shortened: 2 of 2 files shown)


aleixpuigb:
All six blocking items plus the suggestions. Pushed in 3841f48e.

Your diagnosis was the useful part: the entry under-consumed a reference it already commits as full text. I had read `PMID_42253627` as an abstract and stopped, which is why the cohort statistics ended up as prose in descriptions instead of as data.

## The orphaned node, fixed at the cause

Rather than just hanging an edge on it, the fix uses the sentence you pointed at — *"the loss of RASGRP1-mediated signaling impairs not only T-cell-dependent viral clearance but also B-cell tolerance"* — which does more than connect the node. It makes the vascular and cytopenic autoimmunity **one lesion with two targets** rather than two co-occurring findings, and that is now stated in the node description.

Four `phenotypes#` downstream targets added, so the mechanism graph now reaches the clinical graph.

## Frequencies

`OBLIGATE` infections, `VERY_FREQUENT` lymphoproliferation and autoimmunity, `FREQUENT` lymphoma — each with its own cohort-percentage evidence item rather than a band asserted from prose. The EBV 73% / CMV 27% split is curated on the infection phenotype.

## Phenotypes and biochemical

Added: DVT, splenomegaly, lymphadenopathy, ITP, TTP, pan-T-cell lymphopenia, Hodgkin lymphoma. The vasculopathy phenotype is narrowed to the arterial event and the DVT split out, as you asked — a large-vessel MCA occlusion and a limb DVT are different events even sharing a cause.

`biochemical` block added for the antiphospholipid panel. You were right that it was the missing piece: it is what makes the vascular arm an autoimmune process rather than unexplained thrombosis.

**Two terms do not exist in HPO.** TTP has no term — I bound it to `HP:0001873` Thrombocytopenia with `preferred_term: Thrombotic thrombocytopenic purpura` and said so in the description, the same pattern as acro-osteolysis in the DDR2 entry. DLBCL likewise has none, so it stays in the lymphoma phenotype's description rather than getting a binding it cannot have. If you would rather TTP were left unbound than bound to a parent that loses the microangiopathy, say so and I will drop the term.

## Treatments

Conservative immunosuppression added with its 45% mortality curated as a **`REFUTE`** item — against conservative management being adequate *definitive* care, not against the drugs having any effect, since the source explicitly allows temporary control. The index case's failed multimodal rescue is a second item, because "refractory" is more informative stated concretely.

HSCT now carries its own outcome evidence (four of four in remission and alive), with the caveat that this is a retrospective aggregation of fifteen cases where the sickest patients are least likely to reach transplant — a strong signal, not a controlled comparison.

Valganciclovir prophylaxis with EBV surveillance added, framed as one patient's management plan rather than an outcome.

## Suggestions taken

- **CTPS1 / CD27–CD70** added to the EBV node. This was the best suggestion in the review: it turns "IMD64 resembles the narrow-spectrum EBV IEIs" from an assertion into a mechanism, since both of those are themselves causes of EBV-susceptibility IEIs.
- **`HUMAN_MODEL_MISMATCH`** added — and it is a structural mismatch rather than a fidelity shortfall, which is worth distinguishing. Mice are not susceptible to EBV, so the defining human arm has no animal counterpart, while the autoimmune arm does (the SLE-like phenotype). The consequence for the entry is specific and now recorded: the edge from failed EBV-specific CD8 expansion to B-cell lymphoproliferation rests entirely on human observation plus the shared-mechanism argument across MAGT1/ITK/CD27/CD70.
- **`prevalence`** — `CASES_IN_LITERATURE`, with cohort size, family count, the consanguineous Middle Eastern/North African skew and the 12-month median onset.
- **Coagulation screening** added to `diagnosis`, with the source's hedge intact — one case is not a surveillance guideline.
- **`PMID:31402499`** regraded to `OTHER`. You were right, it is a narrative review.

**Onset** I handled differently from your suggestion. The source gives one median for the disease, not per-feature onsets, so putting `onset_category` on individual phenotype descriptors would have implied per-phenotype data that does not exist. It is recorded in `prevalence.notes` and `notes` instead, with that reasoning stated.

54/54 snippets verified. Schema, terms, entity-refs, duplicate-keys and the four baseline gates all pass.

---

PR #10182, opened 2026-08-30
Close the three remaining agent-instruction gaps from #442

### LINKED ISSUE #442: Refine agent instructions about deep research and cached files
I have found that agents don't have clear expectations about the deep research process:
* They don't know how long to wait, and may consider the process to be stuck if it doesn't return results within some arbitrary time
* They may choose an arbitrary set of citations to extract from the deep research report
* They don't consistently interpret pathophysiological events as a chain, even when the report frames the events in this way.

On a more minor note: agents don't inherently know the filenames of cached references (they may search for "pmid_*" instead of "PMID_"). This is usually self-correcting but slows the process, especially for reference types beyond PMID and DOI.

Some combination of refining the agent instructions in Skill files and updating the deep research prompt should help.

caufieldjh:
Closes #442. Prompt-versioning follow-up: #10183.

The issue reported four problems. One turned out to be fixed already; the other three were still open, so this is not a close-with-comment.

| # | Reported problem | Status |
|---|---|---|
| 1 | Agents don't know how long deep research takes | **Already fixed** — no change |
| 2 | Arbitrary citation extraction | Open → fixed here |
| 3 | Pathophysiology not read as a chain | Open → fixed here |
| 4 | Cache filename conventions unknown | Open → fixed here |

## 1 — already fixed, leaving alone

`initiate-new-disorder-creation/SKILL.md` now carries a per-provider timing table (asta seconds, openai/perplexity a few minutes, falcon 20+, openscientist 10–30, claude_code with a worked ~4m50s run). That is more specific than the issue's suggested "2–10 minutes", so nothing to add.

## 2 — citation selection

The skill covers *rejection* thoroughly — unresolved identifiers, off-topic references, `needs_review` vs `confabulation_rate`, quote mismatches. It never said how to **choose**. Added:

- Work from the `.citations.md` companion file, not from whichever identifiers happened to appear in the prose you read most closely. The body cites some papers repeatedly and omits others entirely, so reading it alone gives an arbitrary subset — which is the reported symptom.
- Let the claims choose the references rather than a target count. The issue's own suggestion of "5–8 PMIDs" would be a ceiling as often as a floor.
- What to do when there is no sidecar: sweep the body with `grep -o "PMID:[0-9]*"`, which covers about four in five sidecar-less reports. Where it returns nothing — `falcon` is the usual case, citing by author-year key — sweep for `doi.org/10.` links instead, since 1,334 of 1,340 falcon reports carry them and 8 of the 13 sidecar-less ones do. Fetch those, then look each up in PubMed and prefer the PMID. Re-running the provider is named as the last resort, because a sidecar-less report usually still names its sources.

Two preferences recorded with the reason attached, since both are non-obvious:

- **PMID over DOI for the same paper**, because `DOI:` is in `skip_prefixes` — so the *gating* validator will not check a DOI-keyed quote. It is still checkable, though, and the instruction hands over the command rather than telling the curator to brace themselves: `just count-verified-snippets --unskip-prefix DOI <file>` takes `Empty_Nose_Syndrome` from 58/60 to 60/60 against the cached DOI bodies.
- **Primary report over the review that cites it** for a specific finding; reviews are right for synthesis claims and pooled figures.

`PPR:` preprints are covered separately and for a different reason — they are **not** in `skip_prefixes` and are snippet-checked normally; the reason to prefer the published version is that its text can differ.

## 3 — pathophysiology as a chain

Two halves, because the cause is in the prompt and the symptom is in the YAML.

**Template side.** §6 of `disease_pathophysiology_research.md` asks for mechanism content by *category* — molecular pathways, cellular processes, protein dysfunction, tissue damage. That is a taxonomy, and reports come back sorted into it. The only mention of "causal chains" was in the closing "this will be used for" list, which the research tool has no reason to act on. §6 now asks for an ordered causal chain first, with the categories demoted to a coverage checklist. Correctly scoped: `module_mechanism_research.md` already asks for a causal chain, so the disease template was the outlier.

**Skill side.** `initiate-new-disorder-creation` never mentioned `downstream` at all. Chain guidance existed only in `cancer-curator` (cancer-only) and `dismech-pr-review` — so the reviewer was told to check for a chain while the author was never told to build one. Added a worked example including edge-level `evidence` (the schema is explicit that it means "this specific edge, not just the parent node-level claim" — two well-cited nodes do not establish that one causes the other), the foreign-key constraint on `target`, `just check-entity-refs`, and the instruction to leave an edge out rather than invent one. Step 5's skeleton gained `downstream`/`biological_scale` and a pointer back to the section that explains them.

**One correction to the framing.** I measured this before writing the fix, and the KB is in better shape than the issue implies:

| Multi-node disorder entries | 2,447 |
|---|---:|
| With **zero** `downstream` edges | 75 (3.1%) |
| With **every** node edged | 1,069 (44%) |

The 44% is a floor, not the real figure — a well-formed chain's terminal node correctly has no `downstream`, so full connectivity is undercounted by construction. Either way the instruction gap is real but the output symptom has largely resolved on its own. Worth fixing to stop it recurring, not because the KB is full of flat lists.

## 4 — cache filenames

Documented nowhere, and **the fix suggested on the issue is wrong.** The comment says all prefixes are uppercase. They are not:

```
PMID_29167994.md              clinicaltrials_NCT05813288.md
DOI_10.1038_....md            url_....md
```

The actual rule, from `_safe_id` / `resolve_cache_path` in `reference_snippet_audit.py`: the filename is the identifier with `:` `/` `?` `=` replaced by `_`, **keeping the identifier's own casing**. So a glob for `CLINICALTRIALS_*` finds nothing — the mirror of the `pmid_*` mistake the issue reported. Verified both directions against the live cache.

Added to `dismech-references/SKILL.md`: the rule, a pointer to `resolve_cache_path` as the authority (it also resolves bare `NCT…` ids back to their prefixed file), a case-insensitive tail search for when a glob is genuinely needed, and the 18 prefixes in use with counts — flagged as a snapshot with the command to regenerate it, since a new structured source adds a prefix.

## This changes the deep-research prompt

`templates/disease_pathophysiology_research.md` is the prompt sent to providers, and 1,962 committed reports record their provenance as a bare path (`template_file:`) with no hash or version. So after this merges they all reference a template that no longer says what it said.

That gap is pre-existing, not introduced here — `0729e8e5b6` (MAXO → NCIT) already changed the same file silently. Nothing parses report section structure, so there is no downstream breakage; it is purely provenance and comparability. Tracked separately in **#10183**, which covers where a `template_sha` stamp would go (the existing `dr_align` post-processing hook, so no upstream `deep-research-client` change is needed), whether to backfill, and whether the other six templates want the same treatment.

## Review rounds

Three rounds, and the same lesson each time: **claims about tool behavior have to be run, not reasoned about.**

- `76cbb60` — two blocking findings, both correct, and both instances of the defect this PR is about. `just fetch-reference` does not print a cache path; a DOI snippet is ungated rather than unverifiable. Applying that standard to the rest of my own diff turned up a third that the reviewer had not flagged: the `grep -o "PMID:[0-9]*"` fallback I had just added returns nothing on falcon reports.
- `c4b94e2` — carried the DOI correction into `dismech-references`, which already said "treat a skipped reference as unverified". Pre-existing text that would otherwise have left the two reference skills contradicting each other about the same flag.
- `9d6afab` — replaced the no-sidecar recovery path. The reviewer's fix was right (sweep for DOIs rather than re-running the provider); their supporting count was not, because the sidecar check omitted the `.md` from `<report>.md.citations.md`. Real figure: 13 sidecar-less falcon reports, **zero** carrying PMIDs, 8 carrying DOIs.

## Validation

`pytest tests/test_disclaimers.py tests/test_classify_page_build.py` — 77 passed (the two test files that read skill/template content). Every documented command was run before being documented. Documentation-only change; no KB or schema files touched.

### DIFF
--- a/.claude/skills/dismech-references/SKILL.md
+++ b/.claude/skills/dismech-references/SKILL.md
@@ -172,8 +172,8 @@ Reference prefixes in `skip_prefixes` within
 snippet-checked by default, and the `(N skipped by prefix)` tail of the summary
 line says how many were passed over.
 
-Skipped is not the same as uncheckable. Where the body is cached — every `DOI:`
-reference is — `--unskip-prefix` verifies it on demand:
+Skipped is not the same as uncheckable. Where the body is cached, as a `DOI:`
+reference's nearly always is, `--unskip-prefix` verifies it on demand:
 
 ```bash
 just count-verified-snippets --unskip-prefix DOI kb/disorders/YourFile.yaml
--- a/.claude/skills/initiate-new-disorder-creation/SKILL.md
+++ b/.claude/skills/initiate-new-disorder-creation/SKILL.md
@@ -488,12 +488,24 @@ you work from is still the report's and not your reading path's:
 grep -o "PMID:[0-9]*" research/DISORDER-deep-research-PROVIDER.md | sort -u
 ```
 
-That covers about four in five sidecar-less reports. It returns nothing on the
-rest — including every `falcon` report, which cites by author-year key
-(`martelli2024clinicalspectrumof`) and carries no PMID strings at all. Those keys
-are not identifiers and cannot be fetched; resolve each one by searching PubMed
-for its author, year and title words. If a `falcon` report is missing its
-sidecar, prefer re-running the report over transcribing keys by hand.
+That covers about four in five sidecar-less reports. Where it returns nothing,
+look for DOIs before giving up. `falcon` reports are the usual case: they cite
+by author-year key (`martelli2024clinicalspectrumof`), which is not a fetchable
+identifier, and the sidecar-less ones carry no PMID strings at all — but most
+do carry DOI links, which are fetchable.
+
+```bash
+grep -oE "doi\.org/10\.[^ ;|,)]+" research/DISORDER-deep-research-PROVIDER.md | sort -u
+```
+
+Fetch those with `just fetch-reference DOI:<id>`. Then look each one up in
+PubMed and cite the PMID in preference — a DOI-keyed item is not checked by the
+gating validator, so the PMID is worth the extra lookup. Only for keys with no
+DOI beside them do you have to search PubMed by author, year and title words.
+
+Re-running the report is the last resort, not the first: a sidecar-less report
+usually still names its sources, and a re-run costs a provider call for
+identifiers already sitting in the file.
 
 Let the *claims you are curating* choose the references, not a target count. Every
 assertion you write needs its own citation, so the number falls out of how much of


caufieldjh:
Both blocking findings were right, and they're the defect this PR exists to fix — an instruction that promises output a tool doesn't produce sends an agent straight back to guessing. Pushed in `76cbb60`. I reproduced both before changing anything.

## The two blocking items

**`just fetch-reference` does not print a path.** Confirmed — the output is `Successfully cached …` plus title, authors, content type and length, with no filesystem path. Dropped the promise; the section's case-insensitive tail search already covers that need.

**A DOI snippet is ungated, not unverifiable.** This is the better catch of the two, because it changes the advice rather than correcting it. I reproduced your result:

```
just count-verified-snippets kb/disorders/Empty_Nose_Syndrome.yaml
  Snippets checked: 58/60 verified (2 skipped by prefix)
just count-verified-snippets --unskip-prefix DOI kb/disorders/Empty_Nose_Syndrome.yaml
  Snippets checked: 60/60 verified
```

"Expect to defend the quote yourself" is now the command, plus a note to say in the history record that you ran it — since CI won't.

## The suggestions

Took all five.

- **`PPR:` can't inherit the DOI reasoning.** Right, and I'd blurred two different things by putting them in one bullet. Preprints *are* snippet-checked; the reason to prefer the published version is that the text can differ. That's what it says now.
- **Edge-level evidence.** Added, with the schema's own distinction quoted — evidence for "this specific edge (not just the parent node-level claim)". Two well-cited nodes don't establish that one causes the other, which is the point the section was making in prose only.
- **Step 4/5 split.** Added a pointer from the skeleton back to the section that explains the lines it shows.
- **Double blank line.** Gone.
- **The 44% stat.** Agreed — a terminal node correctly has no `downstream`, so the figure undercounts. PR description only; nothing in the diff depends on it.

## One thing you didn't catch, which I found by taking your point seriously

The no-sidecar clause I added is rewritten, not tweaked. I tested my own advice and it was wrong:

```
grep -o "PMID:[0-9]*" research/Congenital_Zika_Syndrome-deep-research-falcon.md | sort -u
→ 0 results
```

**Falcon reports contain no `PMID:` strings at all.** They cite by author-year key (`martelli2024clinicalspectrumof`), which isn't an identifier and can't be fetched. So the sweep I recommended silently returns nothing for that provider — the same failure mode as the two items you flagged, in text I wrote in the same commit.

It does work for most of the case it's actually for: **137 of the 173 sidecar-less reports carry PMIDs in the body**, because those are largely `--fallback` runs, which do. So the command stays, now with that number attached, and with the author-year case called out as needing a PubMed lookup per key — or a re-run, if a falcon report has lost its sidecar.

## Validation

`pytest tests/test_disclaimers.py tests/test_classify_page_build.py` — 77 passed. Still documentation-only.

---

caufieldjh:
Follow-up to `76cbb60`, now in `c4b94e2`.

Your fix-note on the DOI item said to check both skill files. I'd only fixed the creation skill, and checking the other one found that `dismech-references` already carried this, from before this PR:

> Treat a skipped reference as unverified by these commands.

That's the same resign-yourself framing you objected to, and after my first fix the two skills contradicted each other — one saying verify it with `--unskip-prefix`, the other saying treat it as unverified. Replaced with the distinction that actually holds: **skipped is not the same as uncheckable.** Where a body is cached — every `DOI:` reference has one — the flag verifies it on demand. A reference is genuinely unverified only when it's skipped *and* nothing cached backs it, such as a dataset accession.

Also documented the `(N skipped by prefix)` tail of the summary line, since that's the signal telling you the flag is worth reaching for. Verified against real output:

```
58/60 verified against cached references (2 skipped by prefix)
60/60 verified against cached references          # with --unskip-prefix DOI
```

This is pre-existing text rather than something I introduced, but it's a one-paragraph edit that completes the finding instead of leaving the KB's two reference skills disagreeing about the same flag. Flagging it as slightly beyond the original diff in case you'd rather it were split out.

`pytest tests/test_disclaimers.py tests/test_classify_page_build.py` — 77 passed.

---

caufieldjh:
Took the fix — the paragraph was sending agents to a worse fallback than the files support. Pushed in `9d6afab`. But the supporting numbers need correcting, because the sidecar check has a path bug in it.

## The sidecar filenames are `<report>.md.citations.md`

Your check drops the `.md`:

```
$ ls research/Cystinosis-deep-research-falcon.citations.md
No such file or directory          # ← the name you tested

$ ls research/Cystinosis-deep-research-falcon.md.citations.md
-rw-r--r-- 60149 Aug 30 19:36     # ← the name that exists
```

So "all six lack a sidecar" is an artifact of the missing `.md`. **All six have sidecars.** Sweeping the real population:

```
falcon reports:                    1340
  sidecar-less:                      13
  ...of those, carrying PMID:         0
```

Zero, not six. The six PMID-carrying falcon reports all have sidecars, so they never reach this paragraph — it only applies when the sidecar is missing. Worth flagging beyond this thread: the same `.md` slip would make the coverage figures in your first review too pessimistic.

My wording was still wrong, just differently. "Every falcon report carries no PMID strings at all" is an absolute that six reports falsify, even though none of the six is in scope here. Scoped it to the sidecar-less ones, which is what the paragraph is about.

## Your real finding stands, and it's the valuable half

The DOI point is right and independent of the count. I verified it:

```
falcon reports carrying doi.org/10. links:   1334 / 1340
sidecar-less falcon reports with DOI links:     8 / 13
```

Telling an agent to hand-resolve author-year keys, or to re-run the provider, throws away identifiers already sitting in the file. That was the same shape of defect as the last two rounds, and you were right to hold the PR for it.

The paragraph now sweeps for DOIs, fetches with `just fetch-reference DOI:<id>`, and prefers the PMID after a PubMed lookup — since a DOI-keyed item is ungated, the extra lookup earns its keep. Re-running is named as the last resort, with the reason.

Note the recovery isn't universal: 5 of the 13 sidecar-less falcon reports carry no DOIs either, and for those the author-year PubMed search really is the only route. That's why the fallback chain stays rather than being replaced.

## Suggestion taken

"Every `DOI:` reference is cached" softened to "nearly always is". I reproduced your `Warburg_Micro_Syndrome` case — `42/43 verified (1 not cached locally)` — and you're right that the absolute doesn't survive it, whatever the underlying cause.

## Validation

`pytest tests/test_disclaimers.py tests/test_classify_page_build.py` — 77 passed.

---

caufieldjh:
Context for reviewers, raised by @caufieldjh: this PR changes the deep-research prompt, not just agent instructions.

`templates/disease_pathophysiology_research.md` §6 is the prompt sent to providers, and the change is behavioral — it restructures how the mechanism section is asked for. Reports record their provenance as a **path**, not a version:

```yaml
template_file: templates/disease_pathophysiology_research.md
```

1,962 committed reports carry that line, so after this merges they all reference a template that no longer says what it said when they ran.

That gap is pre-existing rather than introduced here — `0729e8e5b6` (MAXO → NCIT) already changed the same file silently, and every report predating it was asked for MAXO terms with nothing recording the fact. Nothing parses report section structure either, so there is no downstream breakage; it is purely provenance and comparability.

Filed as #10183, which covers where a `template_sha` stamp would go (the existing `dr_align` post-processing hook, so no upstream `deep-research-client` change needed), whether to backfill the 1,962, and whether the other six templates want the same treatment. Keeping it out of this PR so a docs fix does not turn into a provenance-tooling change mid-review.

---

PR #10187, opened 2026-08-30
Record which template revision produced each research report

### LINKED ISSUE #10183: Deep-research reports record a template path, not a template version
A deep-research report records which prompt produced it like this:

```yaml
template_file: templates/disease_pathophysiology_research.md
```

That is a **path**, not a version. The file behind it changes; the reference does not. So a report cannot tell you what it was actually asked.

**1,962 committed reports carry that line**, and the template has already changed under them at least once:

```
0729e8e5b6  Remove MAXO ontology; remap all treatment/diagnosis terms to NCIT (#7228)
```

Every report generated before that commit was asked for MAXO terms. The file now asks for NCIT. Nothing in those reports records which of the two they saw, and nothing flags the difference. This is not hypothetical drift — it has already happened, silently.

It is about to happen again: #10182 restructures §6 (Mechanism / Pathophysiology) from a category taxonomy into an ordered causal chain, which changes the shape of what providers return. That PR is going ahead — the gap predates it and is not made worse by one more edit — but it is the second instance, and a good moment to stop the third.

## Why it matters

Report quality is compared across providers and over time (`docs/reports/deep_research_eval_framework-2026-04-24.md`, the provider evaluations under `docs/research/`). Those comparisons silently assume a fixed prompt. Two reports for the same disease, six months apart, may have been asked materially different questions, and today there is no way to detect that from the artifacts.

It also undercuts curation review: when a report's §6 comes back as a flat taxonomy, a curator cannot currently tell whether the provider ignored a chain instruction or was never given one.

## What is not affected

Nothing parses report section structure — the `pathophysiology` parsers in `scripts/` and `src/dismech/` all operate on KB YAML, not on reports. So this is purely a provenance and comparability problem, not a breakage risk. No urgency, but it compounds.

## Where a fix would go

`template_file` is written upstream, by `deep-research-client` (`processing/template_processor.py:387`), so the obvious place is not ours to change directly. But the `just research-*` recipes already post-process every report in-repo:

```
if [ -f "$output_file" ]; then
    {{dr_align}} "$output_file" --requested "$requested_provider"
fi
```

so a `template_sha` (git blob hash, or a sha256 of the file as read) can be stamped there with no upstream dependency. `dr_align` already rewrites frontmatter, so the mechanism exists.

## Open questions worth deciding together

1. **Hash or semantic version?** A content hash is automatic and cannot drift, but says nothing about *what* changed. A hand-maintained `template_version:` is meaningful but goes stale the first time someone forgets. A hash plus a `CHANGELOG` entry per template gets both, at the cost of discipline.
2. **What about the 1,962 existing reports?** Their template revision is recoverable from `git log` on the template plus each report's `start_time`, so a one-off backfill is feasible. Worth doing, or is "unversioned, pre-2026-08" a good enough bucket?
3. **Does this extend to the other six templates?** `disease_datasets_research.md`, `module_mechanism_research.md`, `hypothesis_deep_research.md`, `disease_surrogacy_research.md`, `disease_pathophysiology_research_asta.md`, `hypothesis_deep_research_datasets.md` all have the same shape. Probably one mechanism for all of them.
4. **Should a template change be a reason to re-run?** Almost certainly not automatically at this volume, but it may be worth surfacing "this report predates the current prompt" in the DR QC recipe rather than leaving it invisible.

Raised by @caufieldjh while reviewing #10182.

caufieldjh:
Closes #10183.

A report records its prompt as a bare path, so the file behind it changes while the reference does not:

```yaml
template_file: templates/disease_pathophysiology_research.md
```

The disease template has four committed revisions, and `0729e8e5b6` (MAXO → NCIT) changed what it asks for without touching a single report. Provider and over-time comparisons have been silently assuming a fixed prompt.

## Two halves

**Going forward, reports are stamped.** All six `just research-*` call sites now record the template's git blob hash:

```yaml
template_file: templates/disease_pathophysiology_research.md
template_sha: "1e7ea4ee817acfe1dda5f77fafe6f2e8b5927666"
```

A blob hash rather than a hand-maintained `template_version:` because it cannot drift out of sync with the file, and rather than a bare digest because git already indexes it:

```bash
git log --find-object=1e7ea4ee817a -- templates/   # which commits carried it
git cat-file blob 1e7ea4ee817a | head -40          # the prompt exactly as it ran
```

Stamping runs **before** `dr_align`, since align may rename the report and `$output_file` is the pre-rename path.

**Looking back, older reports are resolved rather than rewritten.**

## Why there is no backfill

The issue asked what to do about the ~2,650 already-unversioned reports. I did not backfill them, and the diff touches no file under `research/`.

A report's revision is recoverable from its `start_time` against the template's commit history, so the information already exists. Writing it into every committed report would produce an enormous diff to record something derivable, would conflict with anything else touching `research/`, and would still leave any report that cannot be rewritten unanswered — the resolver has to exist either way. Stamping earns its place for *new* reports because it is authoritative rather than inferred, and survives a template edited but not committed.

## The distinction the API enforces

`stamped` and `inferred` are different claims, so they are different enum values, and `is_current()` returns `None` — not `False` — when either hash is unknown. Undetermined never silently reads as stale.

## What it finds

```
Reports scanned: 2645
  stamped        0   0.0%
  inferred    2154  81.4%
  unknown      491  18.6%

templates/disease_pathophysiology_research.md  (1982 reports)
  current revision: 1e7ea4ee817a
  630b0d5e4cd7  1000  superseded
  839c47271432   601  superseded
  c566b1bad3b8   373  superseded

Generated from a superseded revision: 1974
Undetermined (not the same as stale):  491
Current:                                180
```

Those three states partition the corpus — `is_current()` returns exactly `True`, `False` or `None` — so 180 + 1,974 + 491 = 2,645, and undetermined can never fall below the `unknown` count since every `UNKNOWN` resolution has no blob. Both invariants are now stated in the docs, because the first version of this table went stale silently with nothing saying what it had to add up to.

**Every disease-template report predates the current revision** — because #10182 changed that template after all of them were written. That is also why staleness is **reported, never gated**: a gate would go red for the whole corpus after every prompt edit. This answers the issue's fourth question.

The 491 undetermined are honest rather than broken: reports recording no `template_file` (462), a free-text label rather than a path (13), the two that predate their template's first commit, two from a since-retired template, and eight disease-template reports without a usable timestamp.

One quirk is repaired instead of reported: **20 reports record the path with Windows separators** (`templates\disease_pathophysiology_research.md`). Same template, so the separator is folded before lookup — that is exactly the 980 → 1,000 shift in the table above.

## Scope

The mechanism keys on whatever `template_file` a report records, so all **eight** templates are covered with nothing per-template to maintain. That answers the issue's third question.

The scan recurses — reports also live in `research/modules/`, `groupings/`, `surrogacy/` and `datasets/` — while excluding citation sidecars and `*_artifacts/` contents, which are provider tables and figures rather than reports (993 of the 1,023 nested markdown files).

## Two latent bugs the tests caught

**An all-digit blob hash.** Hashes are hex, so an all-digit one is possible (~1 in 10⁹). Written bare, YAML loads it as an **integer** and the leading zeros are gone, after which the stamp cannot be read back. The writer quotes the value; `test_all_digit_hash_survives_a_write_read_round_trip` keeps it quoted.

**An off-by-one in the frontmatter splice.** Bounding the insertion to the frontmatter block (a review suggestion) introduced one immediately: `closing` indexes the newline *before* the closing delimiter, which is also the newline ending the last key line, so an end-exclusive search missed it and a report with `template_file:` as its final frontmatter line **silently failed to stamp**. Three existing tests caught it on the first run; two new ones pin the boundary and the prose case.

## Validation

- `pytest tests/test_template_versions.py` — **25 passed**; **23 passed, 2 skipped** in a simulated `--depth 1` clone, the skips being the real-repo integration tests
- A shallow clone now warns on stderr that missing history is not an unresolvable corpus — `actions/checkout` defaults to `fetch-depth: 1`, so CI is exactly that case
- `ruff check --select E,F,W,UP,FURB,EXE` clean on all three new files
- `just --list` parses and `verify-datasets` keeps its description; `just template-version-audit` runs end to end

## Deliberately not done

Folding staleness into `just qc-deep-research` — it cannot be a gate today for the reason above, and an advisory line is worth adding once there is a view on what a curator should *do* about a superseded report. Noted in the docs alongside a per-template changelog and re-run-on-change, both of which the audit now makes enumerable.

### DIFF
--- a/docs/deep-research-template-versioning.md
+++ b/docs/deep-research-template-versioning.md
@@ -47,6 +47,15 @@ just template-version-audit --template templates/disease_pathophysiology_researc
 just template-version-audit --format tsv --out /tmp/versions.tsv
 ```
 
+`--stale-only` and `--unresolved-only` are mutually exclusive and argparse
+rejects the pair: an unresolved report has no revision to compare, so it can
+never also be stale.
+
+In a **shallow clone** there is no history to infer from, so every unstamped
+report resolves to `unknown`. The summary says so on stderr rather than letting
+missing history read as an unresolvable corpus — worth knowing because
+`actions/checkout` defaults to `fetch-depth: 1`.
+
 The summary reports how each answer was reached:
 
 ```
@@ -54,8 +63,8 @@ Reports scanned: 2645
 
 How the template revision was determined:
   stamped        0  0.0%
-  inferred    2153  81.4%
-  unknown      492  18.6%
+  inferred    2154  81.4%
+  unknown      491  18.6%
 ```
 
 The scan recurses: reports also live in `research/modules/`, `groupings/`,
@@ -114,10 +123,15 @@ At the time of writing, on 2,645 reports:
 |---|---:|
 | Generated from a superseded revision | 1,974 |
 | Undetermined | 491 |
-| Current | 177 |
+| Current | 180 |
+
+Those three partition the corpus — `is_current()` returns exactly `True`,
+`False` or `None` — so they sum to the 2,645 scanned. Undetermined can never
+fall below the `unknown` provenance count either, since every `UNKNOWN`
+resolution has no blob.
 
 **Every report generated from the disease template predates the current
-revision** — 980 from `630b0d5e4cd7`, 601 from `839c47271432`, 373 from
+revision** — 1,000 from `630b0d5e4cd7`, 601 from `839c47271432`, 373 from
 `c566b1bad3b8`, none from the current `1e7ea4ee817a`. This is expected: that
 template changed in #10182, after all of them were written.
 
@@ -145,7 +159,7 @@ lookup.
 ## Scope
 
 The mechanism is template-agnostic — it keys on whatever `template_file` a report
-records, so all seven templates under `templates/` are covered by the same code
+records, so all eight templates under `templates/` are covered by the same code
 with nothing per-template to maintain.
 
 ## Follow-ups not done here
--- a/scripts/template_version.py
+++ b/scripts/template_version.py
@@ -35,6 +35,7 @@ from dismech.template_versions import (
     Provenance,
     Resolution,
     blob_sha,
+    is_shallow,
     iter_reports,
     resolve_report,
     stamp_report,
@@ -76,12 +77,17 @@ def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
     audit.add_argument(
         "--out", type=Path, help="Write the output here instead of stdout."
     )
-    audit.add_argument(
+    # Mutually exclusive by construction, not by taste: an UNKNOWN resolution
+    # has `blob is None`, so `is_current()` returns None and never False. The
+    # two filters can never both match a report, and combining them would
+    # silently print an empty table rather than say so.
+    only = audit.add_mutually_exclusive_group()
+    only.add_argument(
         "--stale-only",
         action="store_true",
         help="Only reports generated from a superseded template revision.",
     )
-    audit.add_argument(
+    only.add_argument(
         "--unresolved-only",
         action="store_true",
         help="Only reports whose template revision could not be determined.",
@@ -212,6 +218,15 @@ def run_audit(args: argparse.Namespace) -> int:
         ]
     if args.unresolved_only:
         resolutions = [r for r in resolutions if r.provenance is Provenance.UNKNOWN]
+
+    if args.format == "summary" and is_shallow():
+        print(
+            "warning: this is a shallow clone, so there is no commit history to "
+            "infer from and every unstamped report resolves to `unknown`. That "
+            "is missing history, not an unresolvable corpus. Re-run in a full "
+            "checkout (`git fetch --unshallow`) for a meaningful census.",
+            file=sys.stderr,
+        )
 
     render = {
         "summary": _render_summary,
--- a/src/dismech/template_versions.py
+++ b/src/dismech/template_versions.py
@@ -45,13 +45,14 @@ from enum import StrEnum
 from functools import cache
 from pathlib import Path
 
-from dismech.research_reports import read_frontmatter
+from dismech.research_reports import FRONTMATTER_DELIMITER, read_frontmatter
 
 __all__ = [
     "Provenance",
     "Resolution",
     "TemplateRevision",
     "blob_sha",
+    "is_shallow",
     "iter_reports",
     "resolve_report",
     "stamp_report",
@@ -205,6 +206,38 @@ def template_history(
     return tuple(revisions)
 
 
+def is_shallow(repo_root: str | Path = ".") -> bool:
+    """Whether the checkout is a shallow clone.
+
+    Args:
+        repo_root: Repository to inspect.
+
+    Returns:
+        ``True`` when git reports a shallow repository.
+
+    Note:
+        This matters because inference reads commit history, and a shallow
+        checkout has none: every report resolves to ``UNKNOWN``, which reads
+        identically to a corpus that genuinely cannot be resolved. Callers
+        should say which they are looking at. ``actions/checkout`` defaults to
+        ``fetch-depth: 1``, so CI is exactly this case.
+    """
+    try:
+        return (
+            subprocess.run(
+                ["git", "rev-parse", "--is-shallow-repository"],
+                cwd=repo_root,
+                capture_output=True,
+                text=True,
+                check=True,
+                timeout=30,
+            ).stdout.strip()
+            == "true"
+        )
+    except (subprocess.SubprocessError, OSError):
+        return False
+
+
 def _normalize_template_path(value: object) -> str | None:
     """Return a usable repository-relative template path, or ``None``.
 
@@ -248,7 +281,7 @@ def _as_utc(value: object) -> datetime | None:
     return parsed.astimezone(UTC)
 
 
-def resolve_report(report: Path, repo_root: str = ".") -> Resolution:
+def resolve_report(report: Path, repo_root: str | Path = ".") -> Resolution:
     """Determine which template revision produced a report.
 
     Prefers the recorded stamp. Falls back to matching the report's
@@ -296,7 +329,7 @@ def resolve_report(report: Path, repo_root: str = ".") -> Resolution:
             detail="no usable start_time to date the report against",
         )
 
-    history = template_history(template, repo_root)
+    history = template_history(template, str(repo_root))
     if not history:
         return Resolution(
             report=report,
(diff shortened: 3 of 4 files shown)


caufieldjh:
Pushed `97006d5` — a self-caught CI failure, before CI found it.

**Two of my tests would have gone red on this PR.** `actions/checkout` in `main.yaml` specifies no `fetch-depth`, so it defaults to `1`. In a one-commit checkout, `git log --follow` over a template returns a single revision, and both `test_template_history_is_ordered_newest_first` (asserting `>= 2`) and the inference test would fail. I reproduced it against a `--depth 1` clone rather than reasoning about it:

```
$ git clone --depth 1 ... && git rev-list --count HEAD
1
$ pytest tests/test_template_versions.py -q
2 failed, 17 passed
```

**The fix is better tests, not looser ones.** The history tests now build their own git repo in `tmp_path` with the template committed twice. That makes them deterministic, independent of checkout depth, and no longer sensitive to someone adding a revision to the real disease template. It also lets them assert what actually matters and what the previous version could not check at all: **a report timestamped between two revisions resolves to the older one.**

Same shallow clone, after the fix:

```
18 passed, 2 skipped
```

The two skips are the genuine real-repository integration tests, guarded with the reason stated.

**One detail worth recording, because it is easy to get backwards.** The fixture forces `GIT_COMMITTER_DATE` rather than passing `git commit --date`. The latter moves only the *author* date, while `template_history` reads the *committer* date — and that is deliberate, not incidental. Committer date is when the revision landed on the branch, and therefore when it was in effect for anyone running a research recipe. An author date survives a rebase or cherry-pick and can be arbitrarily older than the point the content became reachable. Getting this wrong would have silently mis-dated revisions in exactly the cases the tool exists to resolve.

Locally: 20 passed. Shallow: 18 passed, 2 skipped. `ruff check` clean.

---

caufieldjh:
CI went red on `e9831c98` and I misread why at first — worth writing down, because the lesson is about the check I trusted rather than the code.

**It was not the shallow-checkout problem.** That one was real and is fixed in `97006d5`, but the failure predated it. `test (3.13)` failed at the **lint step, before any test ran**: `uvx ruff check .` found seven errors, all in the new files.

**The part worth flagging: my local lint check is not the gate I thought it was.** Same reported ruff version (0.15.8), same repo root, same `pyproject.toml`, `--no-cache` — and locally those seven do not appear:

```
$ uvx ruff --version                    # 0.15.8
$ uvx --refresh ruff --version          # 0.15.8
$ uvx ruff check --no-cache <the three new files>
All checks passed!
```

Meanwhile CI reported `UP033`, `FURB162`, `UP017`×4 and `EXE001` on exactly those files. The divergence runs both ways: my local run flags an `E402` in `tests/test_ci_batch_validation.py` (pre-existing, untouched here) that CI did *not* report. I could not account for it, so I stopped guessing and fixed against the rule codes CI named, verifying with `--select UP017,UP033,FURB162,EXE001`.

Anyone else validating locally before pushing should know that a clean `ruff check` here does not guarantee a clean lint step in CI.

## The seven, all legitimate

- **`EXE001`** — `scripts/template_version.py` had a shebang without the executable bit. `align_research_provider.py` and `new_history.py` are `755`; it now matches, and the bit survives a clone.
- **`UP033`** — `@lru_cache(maxsize=None)` is `@cache`.
- **`UP017`** ×4 — `timezone.utc` → `datetime.UTC`.
- **`FURB162`** — dropped `.replace("Z", "+00:00")` before `fromisoformat`, which has parsed a trailing `Z` natively since 3.11 (this project requires ≥3.12).

That last one changes parsing rather than spelling, so it got a test rather than an assumption — `test_zulu_start_time_still_parses`.

## Verification

- `--select UP017,UP033,FURB162,EXE001 .` → clean
- 21 tests pass locally; **19 passed, 2 skipped** in a simulated `--depth 1` clone
- The audit reports the same **1,974** superseded and **479** undetermined as before the refactor, so the change is behaviour-preserving

Pushed as `ba1a3c0`.

---

caufieldjh:
Blocking finding was right, and the *reasoning* was the useful part. Pushed `66b12c6`.

## The numbers

You didn't just spot stale text — you derived two invariants that make staleness detectable, and both were violated:

- The three states **partition** the corpus (`is_current()` returns exactly `True`/`False`/`None`), so they must sum to the scanned count. 1,974 + 491 + 177 = 2,642 ≠ 2,645.
- Undetermined can never fall **below** `unknown` provenance, since every `UNKNOWN` resolution has no blob. 491 < 492 was impossible.

Confirmed both, from one run:

```
scanned      : 2645
provenance   : {'inferred': 2154, 'unknown': 491}
state        : {'current': 180, 'superseded': 1974, 'undetermined': 491}
partition ok : True      # 180 + 1974 + 491 = 2645
undet >= unk : True
```

`Current` is **180**, as you predicted. The per-revision counts were stale too — **1,000** from `630b0d5e4cd7`, not 980. All figures in the docs and PR body now come from that single run, and I've written the two invariants into the docs so the next edit has something to check itself against rather than relying on a reader to redo the arithmetic.

## Suggestions taken

**1 — shallow clone now says why.** This was the best of the six, because it's the same distinction the module already draws between `None` and `False`: *unavailable* is not *undetermined*. Verified in a real `--depth 1` clone:

```
warning: this is a shallow clone, so there is no commit history to infer from
and every unstamped report resolves to `unknown`. That is missing history, not
an unresolvable corpus. Re-run in a full checkout (`git fetch --unshallow`)...
  unknown     2645  100.0%
```

**2 — splice bounded to the frontmatter block.** And this one bit back, which is the part worth reporting: bounding it introduced an **off-by-one that three existing tests caught immediately**. `closing` indexes the newline *before* the closing delimiter — which is also the newline ending the last key line — so an end-exclusive search missed it, and a report with `template_file:` as its final frontmatter line silently failed to stamp. Fixed, with two new tests pinning that boundary and the prose case you described.

**3** — `--stale-only` / `--unresolved-only` now a mutually exclusive argparse group; the pair errors instead of printing an empty table.

**4** — `repo_root` is `str | Path` on both functions, and `resolve_report` normalises internally.

**5** — eight templates, not seven. Corrected in the docs and PR body.

**6** — `@cache` unboundedness: agreed and left alone, it's a one-shot CLI. Noted rather than changed.

## Validation

- **25 tests pass**; **23 passed, 2 skipped** in a simulated `--depth 1` clone
- `ruff check --select E,F,W,UP,FURB,EXE` clean on all three files
- Audit reports the same **1,974** superseded, so every change here is behaviour-preserving apart from the boundary fix

---

Issue #9390, opened 2026-08-24
A phenotype `frequency:` band has no machine-readable link to the evidence that grounds it — so nothing can check one, and 715 bands are ambiguous even to a human


matentzn:
## The gap

`frequency: FREQUENT` is a quantitative claim — "30–79% of patients". The evidence that justifies it is, in the best entries, a fraction written into an `explanation` in prose: *"10 of 15 MEGF8 cases is 67%, which falls in the FREQUENT band"*. Nothing connects the band to the fraction, and nothing connects either to the specific `EvidenceItem` they came from.

Consequences, in increasing order of how much they cost:

1. Nothing can verify a band. `just qc` has no frequency check at all — I looked; there is no script and no recipe.
2. When a phenotype carries several evidence items, **a reader cannot tell which one the band came from either.** This is not only a tooling problem.
3. The re-derivation is not merely hard, it is *systematically* misleading, in ways set out below.

CLAUDE.md §7 and `docs/frequency-evidence-guidelines.md` already state the policy correctly. #7026 found and fixed 32 bands resting on evidence containing no frequency information at all. But that was a manual sweep, entry by entry, and nothing prevents recurrence — the same class of error recurs because the policy lives in prose and the data has nowhere to record compliance.

## I tried to build the check, and the way it fails is the actual finding

The obvious tool: scrape `N/M` out of a frequency-bearing phenotype's own evidence, compute the percentage, compare against the band's schema range. Prototype run over all of `kb/disorders/` (11,692 phenotypes carrying a band):

| | count |
|---|---|
| bands with **no** fraction anywhere in their own evidence | 10,462 (89%) |
| bands whose evidence contains a fraction, but **more than one** — nothing says which grounds the band | 715 |
| bands with exactly one evidence item and exactly one fraction (unambiguous) | 388 |
| …of those, band contradicts the fraction | 30 (8%) |

Then I hand-checked five of the flagged cases. **Four were false positives, one was real.** That is a ~20% hit rate, and the check is not viable as written. But each false positive fails for a *different* reason, and every one of those reasons is the missing link:

[...]

PR #10189, opened 2026-08-30
curate: Add Hereditary Spastic Paraplegia 44 (MONDO:0013179, GJC2)

### LINKED ISSUE #10184: Curate hereditary spastic paraplegia 44 (MONDO:0013179)
Curate a dismech entry for **hereditary spastic paraplegia 44** ([MONDO:0013179](https://monarchinitiative.org/MONDO:0013179)) — an autosomal recessive complex hereditary spastic paraplegia caused by hypomorphic variants in *GJC2* (connexin-47), the oligodendrocyte gap-junction protein whose null alleles cause the far more severe Pelizaeus-Merzbacher-like disease 1.

Preliminary read: this looks like a `DISEASE`. The stub records a single causal gene (`hgnc:17494` GJC2), one MONDO parent (complex hereditary spastic paraplegia), and no descendants — the leaf case. The KB already carries numbered SPG entries as individual diseases (`Hereditary_Spastic_Paraplegia_7`, `_11`, `_48`, `_49`, `Spastic_Paraplegia_81`, `_89`, `_90A`), so a numbered SPG entry is the established pattern here rather than a subtype.

Allelic-series note for whoever curates it: SPG44 is allelic to Pelizaeus-Merzbacher-like disease 1 (also *GJC2*), which is a genotype/severity distinction rather than a separate mechanism — worth capturing explicitly rather than duplicating. `kb/disorders/Pelizaeus_Merzbacher_Disease.yaml` covers the *PLP1* disease, which is a different gene.

Stub: `stubs/Hereditary_Spastic_Paraplegia_44.yaml`

aleixpuigb:
De-novo curation of SPG44, the mild end of the *GJC2* / connexin-47 allelic series. Closes #10184.

## Why this is its own entry and not a subtype

*GJC2*'s severe disease is Pelizaeus-Merzbacher-like disease 1 (PMLD1/HLD2), an infantile hypomyelinating leukodystrophy. The tempting move is to treat SPG44 as "less PMLD1". The measured molecular difference says otherwise, and it is what the entry is built around: PMLD1 alleles are retained in the ER and, in primary oligodendrocytes, switch on the unfolded protein response and apoptosis; the SPG44 allele p.I33M traffics normally, forms plaques like wild-type Cx47, and triggers neither. What it loses is conductance. So SPG44 is a clean loss of channel function and PMLD1 is that plus a proteotoxic arm.

MONDO gives SPG44 its own leaf term, and the KB already carries numbered SPG loci as individual entries (`Hereditary_Spastic_Paraplegia_7`, `_11`, `_48`, `_49`, `Spastic_Paraplegia_81`, `_89`, `_90A`), so a per-locus entry is the established pattern here.

## Pathograph

connexin-47 conductance loss → disrupted oligodendrocyte–astrocyte panglial coupling → CNS hypomyelination and myelin instability → corticospinal tract dysfunction and progressive spastic paraparesis.

Conformance declared to `cns_myelin_failure` at the insult node and the myelin-sheath node.

## What I tried to get right

**Negative results are modelled, not dropped.** The single Cx47-null mouse has vacuolated myelin and *no overt phenotype* — so it carries a `FAILS_TO_RECAPITULATE` link against the clinical motor node with an `UNCHANGED` readout, alongside a `PARTIALLY_RECAPITULATES` link for the myelin node. Removing a second connexin does produce disease, which is why there is also a `HUMAN_MODEL_MISMATCH` discussion about whether the mouse–human gap is connexin redundancy or something the human missense allele does that a null does not.

**Preclinical is labelled preclinical.** AAV.MBP.Cx47myc gene therapy and transgenic Cx32 replacement both rescue the Cx32/Cx47 double knockout. Both are curated as `RESCUES` model links and as a treatment entry whose first words are "NOT A TREATMENT FOR PATIENTS", with the limitation stated: the model rescued corresponds to severe PMLD1, not to the SPG44 course.

**The genotype–phenotype story does not close, and the entry says so.** In the second published family, one homozygous *GJC2* variant gave complicated HSP in the proband and full HLD2 in both his sisters. One genotype, one sibship, two diagnoses. Filed as an `OPEN_QUESTION` rather than smoothed into the allele-severity ladder, and it is the reason the counselling entry says a recurrence cannot be counselled as mild.

## Deep research

An OpenScientist report was generated for this disease and is committed under `research/`. It was used as a lead source only — every snippet in the entry is anchored to a PMID fetched into `references_cache/` and read directly.

Its record is worth reporting because the two validators disagreed sharply. Citations: 18/18 resolved, zero confabulation. Ontology terms: of 20 labels checked, **7 named a different term than the identifier resolves to** — `NCIT:C61725` given as "Baclofen" when it is Diltiazem, `NCIT:C1027` as "Botulinum Toxin" when it is Budesonide, `NCIT:C15342` as "Physical Therapy" when it is Transplantation, `UBERON:0002718` as "corticospinal / pyramidal tracts" when it is the solitary tract. None of those identifiers is used here.

It was still worth running: the preclinical gene-therapy work (PMID:28100454, PMID:25524707) and the brainstem-auditory-evoked-potential discriminator between *GJC2* and *PLP1* disease (PMID:20513814) reached the entry as leads from that report and were then verified against the primary papers.

## What review changed, and the one worth reading

Four rounds. The last one is the useful one.

I banded parkinsonism `VERY_RARE` and justified it with a number. The reviewer pointed out the number was wrong — I had written "one of two siblings", which is a 50% family ratio presented as a population frequency. I fixed it by reaching for a *different* number, "under the 5% threshold", against a published cohort of about seven individuals where one person is roughly 14%. Wrong again, in the same way.

The reviewer's diagnosis is the part worth keeping: *"my first suggestion sent you to a second number when the answer was to stop citing one."* The final text names no percentage. It says the late-onset *GJC2* literature is single digits, that no numeric band can be read literally at that size, and that `VERY_RARE` records a singular non-core observation. The sibling ratio stays, explicitly as the reason parkinsonism is not banded alongside ataxia and explicitly *not* as the frequency.

I pushed that fix against the reviewer's stated preference not to spend another cycle on it. A band the schema cannot check, justified by arithmetic that is false, is worth a review round even when the band itself is right.

## Noted but deliberately not fixed here

`kb/modules/corticospinal_tract_axonopathy.yaml` says myelin-formation HSPs "SPG2/PLP1, SPG42/GJC2" belong to `cns_myelin_failure`. The routing is right and this entry follows it, but the locus number is wrong — GJC2 is SPG44; SPG42 is SLC33A1. Recorded in the entry's `notes` and filed as #10191 rather than fixed inside a curation PR.

## Validation

```
just validate                    ✓  51/51 snippets verified
just validate-disorders          ✓  (the batched pre-PR gate)
just check-snippet-grading       ✓  (after a fix — see below)
just check-folded-hyphens        ✓
just check-environmental-evidence ✓
```

`check-snippet-grading` caught a real defect on the first run: I had quoted one sentence of PMID:19056803 that carries both a clinical conclusion and an in-vitro one, and graded the two copies differently. Fixed by cutting the quote at the clause boundary so each item cites the half that carries its own claim — which is what the rule is for.

## Diff scope

Entry, history record, the nine PMID cache files this entry cites, the three term-cache rows it introduces (`MONDO:0013179`, `hgnc:17494`, and the matching `diseaseterm` enum row), the deep-research report, and deletion of the stub. No other disease's cache rows are swept in.

---

### DIFF
--- a/kb/disorders/Hereditary_Spastic_Paraplegia_44.yaml
+++ b/kb/disorders/Hereditary_Spastic_Paraplegia_44.yaml
@@ -390,13 +390,14 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: "Our findings confirm and further expand the late-onset phenotypes of GJC2 mutations, to include prominent ataxia, pyramidal disturbances and mild parkinsonism, and confirm the distinctive associated MRI pattern."
     explanation: >-
-      The same sentence that carries the ataxia claim. The band is a cross-cohort judgement, not
-      a within-family ratio: parkinsonism has been reported in one individual across the entire
-      published GJC2 late-onset literature, which is what puts it under the VERY_RARE threshold
-      of 5% of patients. The sibling asymmetry - the abstract's OBJECTIVE section attributes
-      parkinsonism to one of the two siblings where ataxia and pyramidal disturbances are
-      attributed to both - is the reason it is not banded with ataxia, but 1 of 2 is not itself
-      the frequency.
+      The same sentence that carries the ataxia claim. There are too few published patients for
+      any numeric frequency band to be read literally - the whole late-onset GJC2 literature is
+      single digits - so no percentage is claimed here. VERY_RARE records that parkinsonism is a
+      singular, non-core observation: one individual, in one family, reported under the
+      late-onset GJC2 label rather than as SPG44. The sibling asymmetry - the abstract's
+      OBJECTIVE section attributes parkinsonism to one of the two siblings where ataxia and
+      pyramidal disturbances are attributed to both - is why it is not banded alongside ataxia,
+      and is not itself the frequency.
 genetic:
 - name: GJC2
   notes: >-


aleixpuigb:
Thanks — pushed `ce0915d1` addressing suggestions 1 and 3, and here is where I landed on the rest.

**Suggestion 1, orphan phenotypes — fixed.** You were right that this was a graph defect rather than a stylistic point; both rendered as islands. `CNS Hypomyelination on MRI` now hangs off `CNS Hypomyelination and Myelin Instability` as `DIRECT`, since it is literally that node's radiological readout. `Ataxia` hangs off the same node as `INDIRECT_UNKNOWN_INTERMEDIATES` — hypomyelination is not confined to the corticospinal tracts and the ataxic patients show the same MRI pattern, but nothing in the cited literature establishes the steps from diffuse hypomyelination to a cerebellar syndrome, so I did not draw it as `DIRECT`.

**Suggestion 3, parkinsonism — added,** at `VERY_RARE` rather than `OCCASIONAL`. The abstract attributes prominent ataxia and pyramidal disturbances to both siblings but mild parkinsonism to one of them, so a band derived from the same sentence should not be the same band. Same `directness` caveat as ataxia carried in the description.

**Suggestion 5, module nomenclature — filed as #10191.** Fair point that a `notes` block in an unrelated entry is not a record anybody will find.

**Suggestion 2, the two-scale node — not changing, and I want to say why rather than just declining.** You are right that `Corticospinal Tract Dysfunction and Progressive Spastic Paraparesis` reads as two scales in one name. I looked at splitting it into a TISSUE-level tract node and an ORGANISM-level clinical node, and the split would be a bundling fix that creates a worse problem: I have no evidence that separates them. The one sentence that carries this node — the founding family's clinical description — asserts the syndrome, not the tract lesion. A separate `Corticospinal Tract Demyelination` node would have no evidence of its own and would be an inferred node dressed as an observed one, which is the failure this KB's evidence discipline exists to prevent. Given the node is terminal and hands off cleanly to the phenotypes, I would rather keep one honestly-evidenced node than two where the upstream one is asserted. Happy to be overruled if you disagree.

**Suggestion 4, baclofen and botulinum toxin — leaving out,** which I think we agree on. Worth adding one thing to your reasoning: the report's NCIT identifiers for both were not merely unverified, they resolve to *Diltiazem* and *Budesonide*. So this is a case where a report proposed a plausible treatment and a wrong identifier for it, and taking the treatment on the report's word would have imported the identifier too.

**On your GeneReviews check** — your conclusion matches what I found. `PMID:20301682` is the pure-HSP overview and its cached abstract contains no GJC2 or SPG44 content; there is no GJC2 or PMLD1 chapter I could find. Your sandbox blocked PubMed but mine did not, and searching it directly returns nothing for either term.

One thing your review turned up that I had not noticed and that is worth recording somewhere more durable than this thread: the deep-research report's ontology suggestions were wrong at a rate of 7 in 20, including `HP:0007015` proposed as "Progressive spastic paraplegia" when it is *Poor gross motor coordination* — which is close enough in domain that it would survive a casual look. The citation validator gave the same report a clean bill (18/18 resolved, zero confabulation). Those two numbers sitting next to each other is the thing to be careful about.

Re-running the review workflow for the pushed commit.


---

aleixpuigb:
Both carried-forward items handled in `898f3a18`.

**Nit 1 — you are right and it is not purely editorial.** "One of the two siblings" is 50% of that family, and I wrote it into an `explanation` as though it were the frequency basis. The band is correct; the justification I gave for it was not. Reworded to say what actually supports `VERY_RARE` — one individual across the whole published *GJC2* late-onset literature, against a threshold defined as <5% of patients — while keeping the sibling asymmetry where it belongs, as the reason parkinsonism is not banded alongside ataxia rather than as the number the band comes from.

This is the same failure mode #9390 is about, one level down: a band whose real basis is a cross-cohort count, with a different and more visible fraction sitting next to it in the prose. A later reader would have taken the 1-of-2.

**Nit 2, module nomenclature — filed as #10191**, with the `git grep` showing the single occurrence and the correction (GJC2 is SPG44; SPG42 is SLC33A1). Agreed that a `notes` block in this entry was not a durable record.

**On the two-scale node**, my position is unchanged and I set it out in my earlier comment: splitting it would produce an upstream `Corticospinal Tract Demyelination` node with no evidence of its own, since the only sentence carrying this node asserts the clinical syndrome rather than the tract lesion. I would rather have one honestly-evidenced node than two where the upstream one is inferred. Flagging it in case a maintainer wants to overrule me — it is a judgement call about which failure is worse, not a case where I think you are wrong about the observation.

That is everything from both rounds. I am not planning further changes to this PR.


---

aleixpuigb:
Pushed `d3ee6b94` for this, against your advice not to — because the sentence was not imprecise, it was false, and false arithmetic in a curated `explanation` is worth a review cycle even when the conclusion it supports is correct.

Your diagnosis is exactly right and I want to record it, because I made the same mistake twice in a row: told the first number was wrong, I reached for a second number instead of asking whether a number belonged there at all. The denominator you computed is the one this entry's own `prevalence` notes imply — roughly seven published late-onset *GJC2* individuals — and one of seven is ~14%, comfortably above the band's stated 5% cut-off.

The text now claims no percentage:

> There are too few published patients for any numeric frequency band to be read literally — the whole late-onset GJC2 literature is single digits — so no percentage is claimed here. VERY_RARE records that parkinsonism is a singular, non-core observation: one individual, in one family, reported under the late-onset GJC2 label rather than as SPG44.

The band is unchanged. Only the justification moved, from a false quantitative claim to a true qualitative one.

There is a general point here that I think outlives this PR and belongs on #9390. HPO's frequency bands are defined as percentage ranges, and at n≈7 no band can be satisfied numerically by any observation that is not either 0 or nearly all of the cohort — `VERY_RARE` requires ≤1-in-20 and the cohort does not have twenty patients in it. So for ultra-rare disease every band is a qualitative judgement wearing a quantitative label, and an explanation that reaches for arithmetic is inventing precision the data cannot carry. #9390's `frequency_basis` proposal handles the case where a real fraction exists; the complementary case is entries like this one, where the honest answer is that there is no denominator and the schema should let an entry say so rather than leaving curators to write a number. I will add that to #9390 rather than expanding it here.

Also noted, with thanks: you withdrew the corticospinal-node split after reading my reasoning. That is the second time in this review round you have changed position on the argument rather than the assertion, and it made the entry better both times.


---

Issue #10190, opened 2026-08-30
No way to record that a phenotype is explicitly absent: FrequencyEnum omits HP:0040285 "Excluded"


aleixpuigb:
A curator who has read a paper saying "these patients did **not** have X" has nowhere in the schema to put that, so it ends up as prose in a `description`. That loses a fact that is often the most diagnostically useful thing in the entry.

`FrequencyEnum` imports the HPO frequency subset and stops one value short:

| dismech value | HPO term |
|---|---|
| `OBLIGATE` | HP:0040280 Obligate |
| `VERY_FREQUENT` | HP:0040281 Very frequent |
| `FREQUENT` | HP:0040282 Frequent |
| `OCCASIONAL` | HP:0040283 Occasional |
| `VERY_RARE` | HP:0040284 Very rare |
| — | **HP:0040285 Excluded** — not imported |

`Phenotype` has no other negation slot either. Its slots are `category`, `name`, `phenotype_term`, `frequency`, `description`, `diagnostic`, `sequelae`, `evidence`, `context`, `review_notes`, `severity`, `notes`, `subtype`, `subtypes`, `phenotype_contexts`, `electrophysiology`, `reports_on`. There is a `negated` slot in the schema but it belongs to grouping membership criteria, not to phenotype annotation.

## Why this costs something real

A reported absence is not a weak version of a reported presence. It is frequently the discriminating observation.

Three cases, all from entries in the KB or added this week:

**SPG44 vs PMLD1** (added in #10189). The founding report's sentence is: "All three had a late-onset, slowly progressive, complicated spastic paraplegia, with normal or near-normal psychomotor development, preserved walking capability through adulthood, and **no nystagmus**." Nystagmus is a cardinal sign of the allelic severe disease, so its absence is one of the three features separating the two at the bedside. In the entry it survives only inside a `description` string and inside the quoted snippet. Nothing can query it.

**COXPD26** (*TRMT5*, MONDO:0014684, curated under #10185). The 2022 series reports "the mitochondrial respiratory chain complex activity in skeletal muscle was **normal**, except in one patient in whom there was a mild decrease in complex I". That is the disorder's *defining* biochemical feature being absent in three genetically confirmed patients. I could record it as a `REFUTE` evidence item against a pathophysiology node, which is something — but there is no way to say the phenotype itself was looked for and excluded.

[...]

Issue #9847, opened 2026-08-28
preflight-dr has no Named Entity Confusion check for non-Mendelian diseases


aleixpuigb:
`just preflight-dr` detects Named Entity Confusion by comparing gene-symbol mention counts against the canonical causal gene MONDO records for the intended term. For a disease MONDO gives no `RO:0004003` gene, it returns `SKIP`:

```console
$ just preflight-dr research/Renal_Nutcracker_Syndrome-deep-research-openscientist.md MONDO:0019105
SKIP  research/Renal_Nutcracker_Syndrome-deep-research-openscientist.md
  intended entity : MONDO:0019105 renal nutcracker syndrome
  canonical gene  : (none recorded)
  top genes       : ACE=11, FBN1=4, AR=1, HP=1, CFD=1
  - MONDO records no causal gene (RO:0004003) for MONDO:0019105 (renal nutcracker syndrome);
    the gene-identity check cannot discriminate. Fall back to the manual OMIM/synonym preflight.
```

That is honest and the message tells you what to do, which is good. But the consequence is that **an entire class of entries — infectious, structural, acquired, complex, most cancers — gets no automated entity check at all**, on the one preflight the SOP treats as mandatory. `CLAUDE.md` and the references skill both describe `preflight-dr` as the Named Entity Confusion gate without noting it only covers monogenic disease.

Note also the `top genes` line in that output: `ACE=11` is the drug class in "ACE inhibitor", not the gene. `HP=1` is almost certainly the HPO prefix. So the fallback listing is not just uninformative here, it is actively misleading — a reader skimming it could conclude the report is about an ACE-related condition.

### Suggested fix

For a term with no causal gene, discriminate on what such diseases *do* have. Options, cheapest first:

1. **Label and synonym coverage.** MONDO carries `label` plus `hasExactSynonym` / `hasRelatedSynonym` for MONDO:0019105 ("nutcracker syndrome", "left renal vein entrapment syndrome"). Count mentions of those against mentions of the labels of sibling and rival MONDO terms, exactly as the gene check does with symbols. This reuses the machinery already there.
2. **Anatomy.** For structural disease the compressed or affected structure is the entity anchor — "left renal vein" here. UBERON terms extracted from the report could be checked against the entry's own `locations`.
[...]

Issue #10192, opened 2026-08-30
Reference validator strips HGVS variant designations in square brackets, so a verbatim quote naming the alleles cannot be used as a snippet


aleixpuigb:
A verbatim quote from an abstract fails snippet validation because the sentence names its variants in square brackets, which is standard formatting in the genetics literature.

## Repro

PMID:35342985, abstract, verbatim:

> All three patients carried the same two compound heterozygous variants of the TRMT5 (tRNA Methyltransferase 5) gene, one known pathogenic frameshift mutation [c.312_315del (p.Ile105Serfs\*4)] and a second rare missense change [c.665 T > C (p.Ile222Thr)].

Quoting that sentence exactly:

```
[ERROR] Text part not found as substring: 'All three patients carried the same two compound heterozygous
variants of the TRMT5 (tRNA Methyltransferase 5) gene, one known pathogenic frameshift mutation and a
second rare missense change .'
(note: the snippet matches the cached text exactly once [c.312_315del (p.Ile105Serfs*4)] and
[c.665 T > C (p.Ile222Thr)] are kept; bracketed spans are stripped before matching unless
conf/reference_validator_config.yaml lists a matching literal_bracket_patterns entry)
```

The error message is good — it names the cause and the fix location, which is why this took a minute rather than an hour. The problem is the default itself.

## Why the current patterns miss this

`conf/reference_validator_config.yaml` has:

```yaml
literal_bracket_patterns:
  - "^[A-Z]{2,}$"     # abbreviation definitions, e.g. [IL-6]
  - "%"               # percentages
```

Those two were added (per the file's own comment, issue #8597) because structured abstracts define abbreviations in brackets on first use, and stripping them turned verbatim quotes into apparent paraphrases. HGVS in brackets is the identical situation and the identical failure, one content type later.

The stripping default is right for its original purpose — a curator's gloss (`This cytokine [IL-6] is a key factor`) and inline numeric citation markers (`[12]`) should be ignorable. But a bracketed HGVS designation is not a gloss. It is the most load-bearing content in the sentence.

## Suggested fix

Add a pattern matching HGVS notation. Something like:

```yaml
  - "^[cgmnpr]\\."     # HGVS: c.312_315del, p.Ile105Serfs*4, g.1234A>G
```

[...]

PR #10195, opened 2026-08-30
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 77 (MONDO:0013119, LOXHD1)

### LINKED ISSUE #10186: Curate autosomal recessive nonsyndromic hearing loss 77 (MONDO:0013119)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 77** ([MONDO:0013119](https://monarchinitiative.org/MONDO:0013119)) — DFNB77, caused by biallelic *LOXHD1* variants. LOXHD1 is a PLAT-domain protein of the hair-cell stereociliary membrane; loss of it causes progressive hair-cell degeneration, first shown by the ENU mouse mutant *samba*.

Preliminary read: this looks like a `DISEASE`. One causal gene (`hgnc:26521` LOXHD1), parent `hearing loss, autosomal recessive`, no descendants. The KB already curates numbered ARNSHL loci individually (e.g. `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104`), so a per-locus entry is the established pattern.

Two things worth capturing rather than flattening: the *samba* mouse is a strong `animal_models` candidate with real readouts, and *LOXHD1* also carries a Fuchs endothelial corneal dystrophy association, which is a distinct allelic phenotype and not part of this entry.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77.yaml`

aleixpuigb:
De-novo curation of DFNB77, the *LOXHD1* form of progressive autosomal recessive nonsyndromic hearing loss. Closes #10186.

## The mechanism is defined by what was measured and found normal

The easy version of this disease is "hair cells degenerate". The useful version is more specific, and the specificity comes from negatives.

In two independent *Loxhd1* mouse models, mechanotransduction currents in inner hair cells are at wild-type levels through the first postnatal week and severely reduced by postnatal day 11 — tracking the timing of LOXHD1's own arrival in the hair bundle. Over the same interval the bundle stays morphologically normal, tip-link number is not reduced, and Harmonin and LHFPL5, the upper and lower tip-link complex proteins, remain correctly localised. The authors' reading is that the machinery is present but not activatable, and that reading is only available because the alternatives were measured and excluded.

So the pathograph is: LOXHD1 loss of function in stereocilia → mechanotransduction failure with intact bundle architecture → progressive hair-cell degeneration → progressive sensorineural hearing loss. The ordering is what makes the human disease progressive rather than congenital and static, and the founding report places DFNB77 in a class of three genes (*LOXHD1*, *MYO3A*, *PJVK*) required for hair-cell *function* rather than development.

## Two curated negatives, both from studies designed to find the thing

**Vestibular function is spared.** A nine-family Dutch series assembled specifically to characterise the audiovestibular phenotype found no evidence of vestibular involvement; a screen of 8,074 Japanese hearing-loss patients found no accompanying symptoms including vestibular dysfunction among its 28 LOXHD1-positive individuals. This is a finding rather than an absence of interest, because LOXHD1 is expressed in vestibular as well as cochlear hair cells — the null expectation is involvement.

**There is no genotype-phenotype correlation.** Two groups looked. Fifteen variants across nine Dutch families gave no relationship between variant type or position and severity or progression; a Japanese cohort of 28, eighteen of whom share one recurrent splice allele, likewise showed different progression rates. The intrafamilial part is the sharp end — siblings with identical genotypes diverge. Both groups reach for modifiers or environment and neither tests them, which is what the `KNOWLEDGE_GAP` discussion is about, along with a proposed study that uses the shared-allele Japanese cohort as a natural experiment.

## A schema limitation this PR runs into rather than works around

The vestibular finding has nowhere proper to live. `FrequencyEnum` has no `EXCLUDED` value (HP:0040285 is not imported) and `Phenotype` has no negation slot, so an absence that two cohorts went looking for cannot be stated as such.

My first attempt was worse than the gap: I bound the explicitly-absent finding to `HP:0001751` "Abnormal vestibular function" as a phenotype, which asserts to any consumer reading `phenotype_term` the exact opposite of what the papers report. Both `just validate` and `just validate-terms` passed it, because a correct CURIE with a correct label is all they check — nothing verifies that the term means what the surrounding claim says.

Review caught it. It is now modelled as a `Vestibular Hair-Cell Sparing` pathophysiology node instead, following the precedent at `Familial_Glucocorticoid_Deficiency.yaml:718`, so the assertion in the graph matches the assertion in the literature. Filed as #10190; this remains one of the cleanest test cases for it, since the absence is replicated across two cohorts.

## Deep research

An OpenScientist report is committed under `research/`. It earned its keep here: the two cohort papers carrying the vestibular sparing, the missing genotype-phenotype correlation, the Japanese recurrent allele and the Fuchs corneal dystrophy carrier screen all reached the entry as leads from it, and were then read directly. Every snippet is anchored to a PMID in `references_cache/`.

Its record: 20 of 23 ontology terms resolved, with `HP:0008550` unresolved — a confabulated identifier, not used here.

`just preflight-dr` returned WARN, not PASS:

```
WARN  research/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77-deep-research-openscientist.md
  canonical gene  : LOXHD1
  mentions        : LOXHD1=55
  top genes       : LOXHD1=55, PLAT=20, MET=19, LHFPL5=7, SLC4A11=3
  - Rival gene PLAT is mentioned 20 times (36% of LOXHD1). The report may mix in a
    second disease entity — exclude those sections before curating.
```

That is a false positive of the class described in #9847. PLAT here is the polycystin/lipoxygenase/alpha-toxin **domain** that LOXHD1 is built from — fifteen of them — not the plasminogen activator gene. `MET` in the same line is mechanotransduction. The report is on topic; I checked it section by section rather than taking either the tool's word or my own assumption.

## One thing I decided not to do

*LOXHD1* heterozygous missense variants have been associated with late-onset Fuchs endothelial corneal dystrophy, so the obvious move is a `has_subtypes` or comorbidity note. I did not, because the association was tested in a DFNB77 setting and did not reproduce: carriers within the nine Dutch families were screened and none showed pre-clinical or clinical FCD. That is recorded in the gene notes as a specific negative, so a DFNB77 family is not counselled about corneal risk on the strength of a gene-level association that this population's own data does not support.

## Also fixed in review

`NCIT:C50108` was bound as "Hearing Aid". It is **"Plug Device"** — I wrote the label I intended rather than the label the term carries, and `just validate-terms` passed it because terms inside `qualifiers` have no dynamic-enum binding and are never checked against the ontology at all. Corrected to `NCIT:C183182`, and the validation gap filed as #10197. That one is worth reading: the device-qualifier pattern is recommended in CLAUDE.md, and it is the one slot where a wrong CURIE reaches `main` silently.

## Validation

```
just validate                    ✓  48/48 snippets verified
just validate-disorders          ✓
just check-snippet-grading       ✓
just check-folded-hyphens        ✓  (after one fix)
```

`check-folded-hyphens` flagged one real defect: a line ending `...cohort of 305 hearing-` inside a folded scalar, which would have folded to "hearing- impaired". Reflowed.

## Diff scope

Entry, history record, the seven PMID cache files this entry cites, the three term-cache rows it introduces (`MONDO:0013119`, `hgnc:26521`, and the matching `diseaseterm` enum row), the deep-research report, and deletion of the stub.

---

### DIFF
--- a/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77.yaml
+++ b/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77.yaml
@@ -284,6 +284,12 @@ pathophysiology:
     dysfunction. That was wrong and is the reason the rule "no term beats a bad one" exists. See
     issue #10190. Following the pattern of "Preserved Zona Glomerulosa Mineralocorticoid
     Production" in `kb/disorders/Familial_Glucocorticoid_Deficiency.yaml`.
+
+    This node is deliberately isolated - nothing points to it and it points nowhere. That is
+    what a sparing claim is: the LOXHD1 lesion does not cause the sparing, it fails to reach
+    the tissue, so an incoming `downstream` edge would assert a causal relation that runs the
+    wrong way. The node exists to record that the vestibular arm was looked at and found
+    intact, which is what makes the disease non-syndromic.
   biological_scale: CELLULAR
   cell_types:
   - preferred_term: vestibular hair cell
@@ -431,7 +437,7 @@ phenotypes:
     than raising it on an untested suspicion that it is commoner.
   frequency: VERY_RARE
   phenotype_term:
-    preferred_term: Auditory neuropathy pattern - preserved cochlear microphonic with absent neural synchrony
+    preferred_term: Auditory Neuropathy Pattern
     term:
       id: HP:0006958
       label: Abnormal auditory evoked potentials
@@ -682,13 +688,20 @@ treatments:
       value:
         preferred_term: hearing aid
         term:
-          id: NCIT:C50108
+          id: NCIT:C183182
           label: Hearing Aid
   target_mechanisms:
   - target: Progressive Sensorineural Hearing Loss
     description: >-
       Compensates for the threshold loss. It does not act on the transduction defect or slow
       the hair-cell degeneration driving it.
+  notes: >-
+    On the action term: this binds NCIT:C15315 Rehabilitation where the closest existing
+    analogue, `Otofacial_Neurodevelopmental_Syndrome`, binds NCIT:C49236 Therapeutic Procedure.
+    The divergence is deliberate. Fitting and adjusting amplification, with repeated aided
+    testing as the loss progresses, is auditory rehabilitation rather than a generic therapeutic
+    procedure, and Rehabilitation is the more specific of the two available C25218 descendants.
+    The device itself is carried in the qualifier either way.
   evidence:
   - reference: PMID:35875410
     reference_title: "Mutations in LOXHD1 gene can cause auditory neuropathy spectrum disorder."


aleixpuigb:
All five 🟡 items fixed in `c8e02cf6`, plus four of the 🔵 suggestions. Item 1 was a real error on my part and I want to be clear about that rather than filing it under "schema limitation".

**1. Inverted vestibular binding — fixed, and you are right that I got this wrong.** I diagnosed the schema gap correctly and then made the wrong call about what to do while waiting for it. Binding a "we looked and found nothing" observation to `HP:0001751` "Abnormal vestibular function" would have exported as *DFNB77 causes vestibular dysfunction*, which is the reverse of the finding, and no amount of correcting prose in the `description` fixes an export. "No term beats a bad one" applies exactly here and I should have applied it.

I took your second option. It is now a pathophysiology node, `Vestibular Hair-Cell Sparing`, following the `Familial_Glucocorticoid_Deficiency` precedent you pointed at — thank you for finding it, I did not know it was there.

The rewrite paid for itself beyond fixing the binding. Reading `PMID_35875410.md` for the hearing-aid quotes you asked for turned up the *mechanism* of the sparing, which the phenotype form had no room for: vestibular hair cells express only reduced amounts of LOXHD1 and are unaffected in samba mice. So the sparing is a dose effect, not a coincidence, and that is what makes the disease non-syndromic. The node now carries that plus the two human negatives.

**2. ANSD binding — fixed** to `HP:0006958`, matching `Perrault_Syndrome`'s "Auditory Neuropathy Pattern". You are right that my "HPO has no reachable ANSD term" claim was too strong; I checked for an ANSD term and stopped, rather than asking what feature of ANSD is actually bindable.

**3. ANSD frequency — fixed** to `VERY_RARE`. Added one thing to the reasoning in the entry, because it cuts the other way and a later reader should see it: the report's own authors note that no previous DFNB77 study measured outer-hair-cell function, so ANSD may not have been *looked for* elsewhere. That is an argument for the band being an underestimate — which is why I left it low rather than raising it on an untested suspicion.

**4. Amplification — added,** with both outcomes from the same family rather than only the good one. The cousin's aided thresholds were in the mild range; both sisters showed discomfort with amplification and went on to implantation. The second is a `REFUTE` item against the treatment, because a treatment entry showing only the case where it worked would misrepresent a three-child series.

**5. "Untested" — fixed and cited.** You are right, and the specific correction matters: PMID:26973026 gives the hypothesis a concrete form (p.V1892F distorts the PLAT domain, predicted decreased lipid-membrane affinity) rather than leaving it as a domain-class inference. The rationale now says "no experimental test", which is a different and more tractable state than having no hypothesis.

**On the 🔵 items:** added the frequency-bearing quote for the prelingual band, the secondary spiral-ganglion-neuron-loss caveat on the cochlear-implant rationale (it is a real caveat — if the target is degenerating, implant timing might matter, and two patients cannot settle that), and the independent Edvardson implant outcome.

I did **not** add a `mappings:` block for OMIM #613079, because there is nowhere to put it: `mappings` carries `mondo_mappings` only, and the schema has no OMIM slot. `Perrault_Syndrome` handles this by listing OMIM numbers in a description string, so I have recorded both #613079 and gene 613072 in `notes` the same way and said why. If that is worth a slot it is a separate schema issue and I am happy to file it.

The gnomAD constraint figures I also left out, for the reason you gave — they are the research agent's own computations rather than published figures, and `prevalence_class: UNKNOWN` is the honest answer when the only real denominator available is a disease-ascertained cohort.

One tooling note from doing this work, since it bit me twice: the reference validator strips square-bracketed spans from the query side but not from the cached text, so a genuinely verbatim quote containing an inline numeric citation marker (`Grillet et al. [13] revealed...`) fails as "not found". I had to cut two quotes short around brackets. Filed as #10192, which I had opened for the HGVS case before hitting the numeric-citation case — the second is arguably the worse one, since inline citation markers are what the stripping was added to handle.


---

aleixpuigb:
Fixed in `5235e8d3`. `NCIT:C50108` → `NCIT:C183182`. You are right on both counts — the CURIE was wrong and the `label:` I wrote beside it was a fabrication rather than a transcription, which is worse.

The way I introduced it is worth naming, because it is a general trap and not a slip I can promise not to repeat. I had just used `NCIT:C157820` for the cochlear implant in the same file, correctly, and reached for a hearing-aid CURIE from memory instead of looking one up — then wrote the label I intended rather than the label the term has. Every other term in this entry went through `just validate-terms`, which is exactly why I did not notice this one was not being checked.

**On the validation gap, which is the more useful half of your finding.** I confirmed it: `just validate-terms` passes on this file while `NCIT:C50108` is absent from `cache/ncit/terms.csv` and no row gets written. `Qualifier.value` is a plain `Descriptor` with no dynamic-enum binding, so `linkml-term-validator` never walks into it.

That makes the device-qualifier pattern uniquely dangerous, and the danger is inverted relative to how it reads: CLAUDE.md recommends this pattern precisely so a device concept stays queryable rather than being thrown away into free text, so a curator following the guidance lands in the one slot where a bad CURIE reaches `main` silently. Every other term in the entry is gated.

I have filed this as its own issue rather than leaving it in this thread — see below. Two things seem worth doing there: bind `Qualifier.predicate` and `Qualifier.value` to dynamic enums so the validator walks them, and, independently of that, have the label-check run over *every* `term:` block regardless of enum binding, since a label that does not match the term's canonical label is checkable without knowing what enum the term should belong to.

**Non-blocking suggestions:** took the first — `preferred_term` is now plain `Auditory Neuropathy Pattern`, matching `Perrault_Syndrome`, with the descriptive clause left in `description` where it already was.

The other two I left as they are, with the reasons now recorded in the file rather than only in this thread:

- **`NCIT:C15315` Rehabilitation vs `NCIT:C49236` Therapeutic Procedure.** Kept Rehabilitation. Fitting and adjusting amplification, with repeated aided testing as the loss progresses, is auditory rehabilitation rather than a generic procedure, and it is the more specific of the two available `C25218` descendants. CLAUDE.md's guidance on divergent treatment bindings says to check what the treatment actually is before normalising to the majority, and to record the reason — so there is now a `notes` block saying this.
- **The isolated `Vestibular Hair-Cell Sparing` node.** Kept isolated, and I think this one is right rather than merely defensible. The LOXHD1 lesion does not *cause* the sparing; it fails to reach the tissue, because vestibular hair cells express reduced LOXHD1. An incoming `downstream` edge would assert a causal relation running the wrong way. The VEMP result would make a good readout if `Pathophysiology` took readouts, but it does not — readouts live on `ModelMechanismLink` and `Biochemical`, and this is neither.


---

PR #10196, opened 2026-08-30
Show a report's prompt revision on its rendered page

caufieldjh:
Companion to #10187, which stamps `template_sha` into new deep-research reports. That hash was recorded but invisible to anyone reading the rendered page.

## What it looks like

A `Prompt:` chip joins the existing model and citation chips in the report header — the template filename linked to its file, plus the short hash in monospace. The tooltip carries the **full** 40-character hash and the command that resolves it:

```
Prompt revision (git blob hash) 1e7ea4ee817acfe1dda5f77fafe6f2e8b5927666
Resolve with: git log --find-object=1e7ea4ee81…666 -- templates/
```

A bare blob hash is not actionable on its own — `git log --find-object` is what turns it into "which commit was this", and `git cat-file blob` into "what did it say".

## Only stamped hashes are shown

An older report's revision *is* recoverable — that's what `just template-version-audit` does, from `start_time` against the template's commit history. But inference is a reconstruction rather than a record, and a metadata chip has no room to say which of the two it's showing. So a report that predates stamping gets no chip rather than a chip that overstates what is known. This is the same `stamped` vs `inferred` distinction #10187's API enforces.

## Four input shapes, four behaviours

Not one hopeful path, because the corpus contains all of them:

| `template_file` | Rendered | Tooltip pathspec |
|---|---|---|
| `templates/foo.md` | filename linked to the file, plus hash | `-- templates/` |
| `templates\foo.md` (20 reports) | same, separators folded | `-- templates/` |
| `manual_curation` (a label, not a path) | shown unlinked — linking it would be a 404 dressed up as provenance | **omitted** |
| absent | hash alone | omitted |

The pathspec is scoped only when the prompt really is a `templates/` path. Elsewhere the blob isn't under that directory, so a scoped `git log` returns nothing — and a lookup that finds no commits reads as "this hash isn't in the repo", which is the wrong conclusion.

**The unlinked branches are defensive, not expected.** #10187's `stamp_report` declines unless `template_file` resolves to a file on disk, so a label or an ephemeral `/tmp` path is never stamped by the pipeline; reaching them takes a hand-written stamp. They're handled so such a stamp degrades to "shown but not linked" rather than to a bad link — and for the same reason the recorded value is shown verbatim rather than basenamed, since tidying it would hide what the report actually said.

One caveat: **the linked filename resolves to `main`**, so it shows the template's *current* content, which may differ from the revision the hash names. That's what the hash beside it is for. The link can also 404 outright for a since-deleted template — `disease_pathophysiology_research_scanner.md`, named by two committed reports, no longer exists — and nothing better is available, since GitHub's `/blob/<ref>/<path>` needs a commit-ish and a blob hash isn't one.

## Independent of #10187

The chip reads frontmatter only — it does not import `template_versions` — so this lands safely before or after that PR and is simply inert until stamped reports exist (0 of 2,645 today). Kept that way deliberately rather than stacking, so #10187's approval isn't disturbed.

## Validation

- **11 tests**: absent / blank / non-string stamp (an unquoted all-digit hash loads from YAML as an `int`, which #10187 also guards), all four `template_file` shapes, Windows separators, both tooltip forms, and both rendered outcomes
- `pytest -k render` — **190 passed**
- `ruff check` clean on both changed Python files

`render.py` is left un-`ruff format`ed: it is already format-dirty on `main` (verified by stashing and re-checking unmodified `main`), so reformatting here would bury a ~50-line change in a whole-file diff.

## Follow-up

Once #10187 lands, one follow-up on this file: have the reader import the writer's `template_sha` constant instead of re-declaring the literal, and tighten the docstrings' `stamp_report` references now that they resolve to something real.

## Not done

The research **index** page is unchanged. It is one row per report with a provider pill, and a column empty for 2,645 of 2,645 reports today is noise rather than provenance. Easy to add if the stamped share grows.

### DIFF
--- a/src/dismech/render.py
+++ b/src/dismech/render.py
@@ -3087,7 +3087,19 @@ def _prompt_provenance(metadata: dict) -> dict | None:
         ``name`` links to the template at ``main``, which is the file's *current*
         content and so may differ from the revision the hash names. That is the
         point of showing the hash beside it; the tooltip carries the lookup that
-        resolves it.
+        resolves it. The link can also 404 outright when a template has since
+        been deleted -- two committed reports name
+        ``disease_pathophysiology_research_scanner.md``, which no longer exists.
+        Nothing better is available: GitHub's ``/blob/<ref>/<path>`` needs a
+        commit-ish and a blob hash is not one, and the git-blobs API returns
+        base64 JSON rather than a page.
+
+        The unlinked branches are defensive rather than expected. ``stamp_report``
+        declines unless ``template_file`` resolves to a file on disk, so a
+        free-text label or an ephemeral ``/tmp`` path is never stamped by the
+        pipeline; reaching them takes a hand-written stamp. They are handled
+        anyway so that a hand-written one degrades to "shown but not linked"
+        rather than to a 404 dressed up as provenance.
     """
     raw = metadata.get(STAMP_KEY_TEMPLATE_SHA)
     sha = raw.strip() if isinstance(raw, str) else ""
--- a/src/dismech/templates/research_report.html.j2
+++ b/src/dismech/templates/research_report.html.j2
@@ -143,11 +143,6 @@
             color: var(--gray-600);
         }
 
-        .meta-chip-prompt code {
-            font-family: ui-monospace, SFMono-Regular, Menlo, monospace;
-            font-size: 0.92em;
-        }
-
         .meta-chip {
             display: inline-flex;
             align-items: center;
@@ -157,6 +152,11 @@
             border: 1px solid var(--gray-200);
             border-radius: 999px;
             color: var(--gray-700);
+        }
+
+        .meta-chip-prompt code {
+            font-family: ui-monospace, SFMono-Regular, Menlo, monospace;
+            font-size: 0.92em;
         }
 
         a.meta-chip-link {
@@ -471,8 +471,16 @@
                       stamping, and an inferred revision is a reconstruction
                       rather than a record. See issue #10183.
                     #}
+                    {#
+                      The pathspec is scoped only when the prompt is a real
+                      `templates/` path. For anything else the blob is not under
+                      that directory, and a `git log` that finds nothing reads as
+                      "this hash is not in the repo" -- the wrong conclusion. The
+                      command carries the full hash rather than the abbreviation,
+                      since that is what someone pastes.
+                    #}
                     <span class="meta-chip meta-chip-prompt"
-                          title="Prompt revision (git blob hash) {{ report.prompt.sha }} &#10;Resolve with: git log --find-object={{ report.prompt.short }} -- templates/">
+                          title="Prompt revision (git blob hash) {{ report.prompt.sha }} &#10;Resolve with: git log --find-object={{ report.prompt.sha }}{% if report.prompt.url %} -- templates/{% endif %}">
                         Prompt:
                         {%- if report.prompt.name %}
                         {% if report.prompt.url %}<a href="{{ report.prompt.url }}" target="_blank" rel="noopener noreferrer">{{ report.prompt.name }}</a>{% else %}{{ report.prompt.name }}{% endif %}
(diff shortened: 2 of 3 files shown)


caufieldjh:
Took suggestions 1, 2, 5 and 6 in `0e6e831`. Both tooltip fixes land on the same line.

**Suggestion 2 turned out to matter more than suggestion 1**, which inverts your ranking — worth explaining why.

The full-hash fix hits the **common** case. Every linked template's tooltip was telling a reader to paste a 12-character prefix while the full 40 sat two words earlier in the same string. That's an always-on papercut.

Suggestion 1 is real but **defensively-only reachable**. I checked #10187's `stamp_report`: it declines unless `template_file` resolves to a file on disk —

```python
template_path = repo_root / template
if not template_path.is_file():
    return None
```

— so `manual_curation` and the ephemeral `/tmp` path can never be stamped by the pipeline. Reaching the unlinked branch takes a hand-written stamp. Fixed anyway, since the cost is one conditional and a hand-written stamp should degrade to "shown but not linked" rather than to a misleading lookup, but it isn't the live defect.

That also settles **suggestion 3**, which I've left alone. Basenaming a `/tmp` path would hide what the report actually recorded in order to tidy a case the pipeline cannot produce. Showing the recorded value verbatim is the more honest failure mode, and the reasoning is now in the docstring so the next reader doesn't re-litigate it.

**Suggestion 6** is recorded in the docstring: the link can 404 outright for a since-deleted template — `disease_pathophysiology_research_scanner.md`, named by two committed reports, no longer exists — and nothing better is available for the reason you gave. **Suggestion 5**: the modifier rule now follows the base rule.

**Suggestion 4** I'd rather leave until #10187 lands, then have the reader import the writer's constant in one follow-up. Doing it now would couple the PRs and disturb that one's approval, which was the point of keeping them apart.

## On the things you couldn't check

`ruff` isn't installed in your environment, so for the record: `uvx ruff check` is clean on both changed Python files, and `render.py` is format-dirty on `main` before my change — I verified by stashing and re-running `ruff format --check` against unmodified `main`, which still reports it. That's why it isn't reformatted here.

You're also right that `just template-version-audit` is a forward reference to #10187 and nothing in this PR depends on it.

## Validation

11 tests, one new covering the unlinked tooltip form (asserts the pathspec is absent and the label unlinked). `pytest -k render` — 190 passed. `ruff check` clean.

---

caufieldjh:
Thanks — and noted that you withdrew suggestion 3 rather than restating it. That's the right call on the evidence and I'd have taken it either way.

**On the `stamp_report` forward reference: agreed, but I'm folding it into the follow-up rather than pushing now.**

You're right that a reader who greps for `stamp_report` between this merging and #10187 merging finds nothing and can't check the claim — and the claim is load-bearing, since it's the justification for the unlinked branches existing at all. Someone who can't verify it might reasonably delete them as dead code.

The reason to defer: your suggestion 4 already commits this file to a follow-up once #10187 lands — having the reader import the writer's `template_sha` constant instead of re-declaring the literal. That follow-up touches the same docstrings, at a point when `stamp_report` actually exists and the reference resolves to something real rather than to a PR number. Doing it now costs a review cycle to add a pointer that the follow-up would make redundant.

So the plan is one follow-up after #10187 merges, carrying both: import the constant, and tighten the `stamp_report` references now that they resolve.

If you'd rather have the `(#10187)` pointer immediately, say so and I'll push it — it is genuinely one word and I don't want to trade your legibility concern for my tidiness.

**Suggestion 2 (traversal-ish path):** agreed it's a curiosity, and thanks for writing down *why* it's safe — autoescaped URL string, browser-normalised into a GitHub 404, hand-written frontmatter only. That's exactly the kind of thing the next reader would otherwise re-derive from scratch.

**On ruff**, since you can't run it there: `uvx ruff check` is clean on both changed Python files at this head, and `render.py` remains format-dirty on unmodified `main` — I re-confirmed by stashing.

---

Issue #9888, opened 2026-08-28
`just research-disorder` always sends an empty mondo_id, so every deep-research run is ungrounded


aleixpuigb:
The `research-disorder` recipe in `project.justfile` extracts `disease_name` and `category` from the disorder YAML, but passes `mondo_id` as a hardcoded empty string:

```
disease_name=$(grep "^name:" "$yaml_file" | head -1 | sed 's/name: *//' | tr '_' ' ')
category=$(grep "^category:" "$yaml_file" | head -1 | sed 's/category: *//' || echo "")
...
    --var "disease_name=$disease_name" \
    --var "mondo_id=" \
    --var "category=$category" \
```

The template consumes it:

```
- **MONDO ID:** {mondo_id} (if available)
```

so that line renders blank on every run, for every provider and every disease. The one identifier that would pin down which disease the provider is meant to research is never sent.

## What this cost, concretely

Running `just research-disorder openscientist Congenital_Myasthenic_Syndrome_15` produced a 832-line report whose central claim is:

> Finding 1 — CMS15 is a presynaptic CMS caused by biallelic *SLC18A3*/VAChT variants

CMS15 (MONDO:0014542) is *ALG14*. MONDO states it outright — "Any congenital myasthenic syndrome in which the cause of the disease is a mutation in the ALG14 gene" — with exact synonym `ALG14 congenital myasthenic syndrome`. *SLC18A3*/VAChT is a different congenital myasthenic syndrome. The whole report is internally coherent, well-cited science about the wrong disease, which is the most expensive kind of wrong: nothing in it looks broken.

`just preflight-dr` scored the report's gene mentions as `SLC18A3=35, CHAT=10, DAP=4, CHRNE=3, RAPSN=3`. ALG14 does not appear in the top five.

The report also asserts OMIM 616227 — the correct OMIM number for CMS15 — while describing *SLC18A3*, so it carries a correct identifier next to a wrong gene. A curator skimming for the OMIM number would find it matching.

## Suggested fix

Read the MONDO ID from the disorder YAML the same way `disease_name` is read, e.g. the `disease_term.term.id`, and pass it through. For a file that has no `disease_term` yet (the seeded-stub case, which is exactly when a curator reaches for deep research), take it from the stub's `mondo_id`, or require it as an argument rather than silently sending nothing.

Related: #9889 covers the preflight check that was supposed to catch this and could not.

aleixpuigb:
[...]

Issue #9421, opened 2026-08-24
`EvidenceItem` records where a claim came from but not how strong it is: no study design, sample size, or validity classification


matentzn:
## The gap

`EvidenceItem` has seven slots: `reference`, `reference_title`, `supports`, `evidence_source`, `snippet`, `explanation`, `images`.

Between them they answer *what was said, where, and whether it agrees with the claim*. They do not answer **how much weight the claim can bear**. A reader — human or machine — cannot tell an n=1 case report from a 5,000-patient randomized trial, because both are:

```yaml
supports: SUPPORT
evidence_source: HUMAN_CLINICAL
```

## How undifferentiated this actually is

Across `kb/disorders/`, `kb/modules/` and `kb/comorbidities/` — **130,693 evidence items**:

| `evidence_source` | count | share |
|---|---:|---:|
| HUMAN_CLINICAL | 85,093 | 65.1% |
| OTHER | 21,094 | 16.1% |
| IN_VITRO | 8,326 | 6.4% |
| MODEL_ORGANISM | 7,820 | 6.0% |
| *(unset)* | 7,754 | 5.9% |
| COMPUTATIONAL | 606 | 0.5% |

| `supports` | count | share |
|---|---:|---:|
| SUPPORT | 119,103 | 91.1% |
| PARTIAL | 10,568 | 8.1% |
| REFUTE | 573 | 0.4% |
| NO_EVIDENCE | 448 | 0.3% |
| WRONG_STATEMENT | 1 | 0.0% |

Two thirds of the corpus sits in one evidence-type bucket that spans case report → cohort → RCT → meta-analysis → clinical guideline, and nine tenths in one support bucket. Neither field discriminates within the mass of the KB. `evidence_source` answers *what kind of system was studied*; nothing answers *how well it was studied*.

## Concrete cases from one curation run

These came up while curating five entries, so they are ordinary rather than exotic:

[...]

Issue #10197, opened 2026-08-30
Terms inside `qualifiers` are never validated, so a nonexistent CURIE with a fabricated label reaches main silently


aleixpuigb:
`just validate-terms` does not check `term:` blocks nested inside a descriptor's `qualifiers`. A CURIE that does not exist, carrying a label that is not the term's label, passes every gate in the repo.

## Repro

I wrote this into a treatment while curating #10195:

```yaml
  treatment_term:
    preferred_term: hearing aid amplification
    term:
      id: NCIT:C15315
      label: Rehabilitation
    qualifiers:
    - predicate:
        preferred_term: medical device
        term:
          id: NCIT:C16830
          label: Medical Device
      value:
        preferred_term: hearing aid
        term:
          id: NCIT:C50108       # <- wrong term
          label: Hearing Aid    # <- not this term's label
```

`NCIT:C50108` is **Plug Device**:

```console
$ uv run runoak -i ols:ncit info NCIT:C50108 NCIT:C183182
NCIT:C50108 ! Plug Device
NCIT:C183182 ! Hearing Aid
```

Everything passed:

```console
$ just validate-terms kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77.yaml
✅ Validation passed
$ just validate kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77.yaml
✓ All validations passed
$ grep NCIT:C50108 cache/ncit/terms.csv
$                       # absent — no cache row was ever written
```

`just check-term-cache-integrity` also passes, because the row is not there to be inconsistent.

## Why

`Qualifier.predicate` and `Qualifier.value` are plain `Descriptor` ranges with no `reachable_from` dynamic enum on their `term` slot. `linkml-term-validator` drives off the enum bindings, so it never walks into them. `treatment_term.term` is bound to `TreatmentTerm` and is checked; the qualifier hanging off it is not.

The empty cache row is the tell, and it is a silent one: a curator who assumes "if it validated, it got cached" has no signal.

## Why this slot in particular

The device-qualifier pattern is **recommended** by CLAUDE.md. Its "Devices are not clinical actions" section tells curators to attach the device as a `qualifiers` predicate/value pair rather than losing it to free text, precisely so the device concept stays queryable:

[...]

Issue #9827, opened 2026-08-28
Evidence items can't record n, study design, or replication — so a case report and a 301-patient cohort are indistinguishable


aleixpuigb:
This is a follow-on to #5000 and the [CEBM pilot](https://github.com/monarch-initiative/dismech/blob/main/docs/superpowers/CEBM_evidence_level_pilot.md), from four de-novo curations done today (#9821, #9822, #9823, #9824). It is deliberately **not** a re-proposal of an evidence-level tier — the pilot made a good case that a per-reference CEBM grade misfires, because the same PMID can back both a trial's own novel finding and background prose from its introduction.

The gap I kept hitting is narrower and, I think, easier: **the objective, descriptive facts about a cited study have nowhere to live.** Not "how good is this evidence", which is a judgement, but "how many patients was it, what kind of study was it, has anyone else seen it" — which mostly are not judgements.

## The concrete problem

`EvidenceItem` has exactly seven slots: `reference`, `reference_title`, `supports`, `evidence_source`, `snippet`, `explanation`, `images`.

So these two items are structurally identical, and identical in every export:

```yaml
# Immunodeficiency 92 - the ENTIRE disease literature is three patients
- reference: PMID:31103457
  supports: SUPPORT
  evidence_source: HUMAN_CLINICAL
  snippet: "He had B cell lymphopenia, impaired B cell proliferation upon stimulation with ..."
```

```yaml
# AIMAH3 - systematic germline screening of 301 consecutive index cases
- reference: PMID:39921449
  supports: SUPPORT
  evidence_source: HUMAN_CLINICAL
  snippet: "Ten (3.3%) patients carried a germline KDM1A pathogenic or likely pathogenic variant, ..."
```

One is n=1. One is n=301. `evidence_source: HUMAN_CLINICAL` is doing identical work in both, because it classifies the *biological source* — human vs animal vs in vitro vs computational — which is orthogonal to how much the study can bear.

## Six things I had to put in prose because there was no slot

Each of these is from an entry in this batch. In every case my workaround was to write it into the `explanation`, which works for a human reading the page and is invisible to any query, export, or compliance score.

1. **Sample size.** IMD92: three patients exist in the world. I wrote "Single patient." into explanation after explanation, and added a `discussions` entry attached to the disease as a whole to say the entry should be read that way. That is a lot of machinery to express one integer.

[...]

PR #10198, opened 2026-08-30
curate: Add Combined Oxidative Phosphorylation Defect Type 26 (MONDO:0014684, TRMT5)

### LINKED ISSUE #10185: Curate combined oxidative phosphorylation defect type 26 (MONDO:0014684)
Curate a dismech entry for **combined oxidative phosphorylation defect type 26** ([MONDO:0014684](https://monarchinitiative.org/MONDO:0014684)) — a multisystem mitochondrial disease caused by biallelic *TRMT5* variants, which impair m1G37 methylation of mitochondrial tRNAs and so degrade mitochondrial translation across several respiratory-chain complexes.

Preliminary read: this looks like a `DISEASE`. One causal gene (`hgnc:23141` TRMT5), one MONDO parent (combined oxidative phosphorylation deficiency), no descendants. The KB already curates numbered COXPD entries individually (`Combined_Oxidative_Phosphorylation_Defect_Type_4`, `_21`, `Combined_Oxidative_Phosphorylation_Deficiency_42`), so a per-number entry is the established pattern.

The literature is small and centred on a handful of families, so the reported phenotype range (from lethal infantile multisystem disease to adult exercise intolerance) is thin evidence rather than settled — worth curating with that uncertainty explicit rather than smoothed over.

Stub: `stubs/Combined_Oxidative_Phosphorylation_Defect_Type_26.yaml`

aleixpuigb:
De-novo curation of COXPD26, the *TRMT5* mitochondrial-translation disorder. Closes #10185.

## Mechanism

*TRMT5* encodes the methyltransferase that installs m1G37 on mitochondrial tRNAs, immediately 3' of the anticodon. Every one of the thirteen mtDNA-encoded respiratory-chain subunits passes through that translational bottleneck, so losing one methyl group produces a *combined* rather than an isolated respiratory-chain defect.

TRMT5 loss of function → m1G37 hypomodification of mitochondrial tRNA → tissue-specific amplification of the modification defect in skeletal muscle → impaired mitochondrial translation → combined respiratory chain complex deficiency → multisystem energy-failure phenotype.

The modification defect is unusually well controlled in the founding report, and the entry records both halves: siRNA depletion of TRMT5 in control cells *reproduces* the hypomodification, and lentiviral re-expression of wild-type TRMT5 in patient fibroblasts *reverses* it to control levels, with no effect in control cells as the specificity control.

## Two things this entry curates rather than smooths over

**The hypomodification is tissue-specific, and that changes what a negative test means.** The deficit is "particularly prominent in skeletal muscle". In one of the two founding individuals, muscle enzymology showed decreased complex IV and borderline-low complex I *while fibroblast studies were normal*. A normal fibroblast respiratory-chain panel does not exclude COXPD26. That is modelled as its own pathophysiology node rather than left as a caveat, because it is the practically actionable fact in the entry.

**The literature disagrees about the disorder's defining feature — and not evenly.** Powell 2015 defines COXPD26 by "multiple mitochondrial respiratory-chain-complex deficiencies in skeletal muscle" in both its individuals. Argente-Escrig 2022 studied three patients from three unrelated families carrying an identical *TRMT5* allele pair and found muscle respiratory-chain activity "normal, except in one patient in whom there was a mild decrease in complex I enzyme activity" — despite giant mitochondria in the same tissue. The structural marker was there and the enzymatic one was not.

There is a third report, and it took a review round to find it. The Chinese case report — cited five times elsewhere in this entry — independently states "Muscle biopsy also indicated reduced respiratory chain complex enzyme activities." So the tally is two reports finding the biochemical defect against one finding it absent, not one against one.

The entry does not turn that into a majority verdict, for two reasons it states outright. The dissenting series is the largest single patient group of the three, so counting reports and counting patients point in opposite directions. And the third report could not measure COX — complex IV, the complex the founding report found *most* affected — so it establishes abnormal enzymology while being silent on precisely the axis the disagreement turns on.

It remains a `CONTROVERSY` with three open readings and, importantly, the measurement that would settle it: nobody has run the m1G37 primer-extension assay on the 2022 patients' muscle RNA. It is one assay, on stored material, using a method the founding group already established. The contradicting result also sits as a `REFUTE` evidence item directly on the node it contradicts, so it travels with the claim rather than only living in the discussion.

## Negative results modelled explicitly

- Loss of m1G37 does **not** destabilise the tRNA — steady-state levels are unchanged — which is why the downstream node is a translation defect rather than tRNA depletion.
- One of the two founding patients showed **no detectable change** in fibroblast oxygen consumption, recorded as an `UNCHANGED` readout with a `REFUTE` evidence item, which is why the patient-fibroblast model is used for the molecular node and not the bioenergetic one.

## Deep research — the run that went wrong, and what fixed it

**The first OpenScientist report was about a different disease.** It described *TRMT10C* / MRPP1 throughout, which is COXPD30 (MONDO:0014856, OMIM 616974), while labelling it COXPD26 and printing `| MONDO | MONDO:0014684 |` in its own identifier table. Its validators all looked healthy: 14/14 references resolved, confabulation rate 0.0, 21/22 ontology terms resolved.

`just preflight-dr` caught it unambiguously:

```
FAIL  research/Combined_Oxidative_Phosphorylation_Defect_Type_26-deep-research-openscientist.md
  canonical gene  : TRMT5
  mentions        : TRMT5=0
  top genes       : TRMT10C=52, PRORP=21, HP=7, ELAC2=5, HSD17B10=4
  - Expected gene TRMT5 is never mentioned, but TRMT10C is mentioned 52 times.
    The report is most likely about a different disease entity — discard it rather
    than cherry-picking.
```

I discarded it per that advice; it contributed nothing to this entry, which was built from primary literature throughout. The failed report is not committed.

**What is committed is a second run** whose brief names the gene, not just the MONDO ID. It passes cleanly — TRMT5 mentioned 38 times to TRMT10C's 5, the latter only as the contrast.

That difference is the useful experimental result, and I have added it to #9888, which proposes passing the MONDO ID as the fix for exactly this failure. I *did* pass the MONDO ID on the first run — I had to call `deep-research-client` directly, since `just research-disorder` exits if the disorder YAML does not already exist, which is the whole de-novo case — and the ID travelled through and got printed next to the wrong gene. Naming the gene in the brief is what worked. Populating the variable is necessary and not sufficient.

**One more thing about the good report, which is the part I did not expect.** It never cites PMID:35342985 — the three-patient series whose normal muscle enzymology is the whole basis of this entry's `CONTROVERSY`, and arguably the most important thing to know about the disorder's biochemical phenotype. A report on the right disease, citing 18 references with zero confabulation, still missed the paper that complicates its own central claim. Both facts are recorded in the entry's `notes`.

## What review changed

Four rounds, and it is worth listing what they caught, because none of it was catchable by the checks.

**Round 1, four blocking findings**, all fixable from snippets already committed and verified here — no new literature for any of them:

- **Phenotypes 6 → 16.** Two snippets I had already quoted enumerated roughly ten manifestations between them and I had modelled one from each. Banded per source: `FREQUENT` for the features counted 3/3 in the 2022 series and those named in the disease-level characterization sentence, `OCCASIONAL` for the three resting on the single Chinese patient, with each `explanation` saying which basis it used.
- **`GO:0030488` → `GO:0070901`.** The `preferred_term` already said "mitochondrial tRNA methylation"; the binding was one level up. The compartment is the disorder.
- **PMID:33398350 was cached and claimed as cited, but appeared in zero evidence items.** Repaired by using it on the entry's weakest node rather than by deleting the claim.
- **`histopathology` added**, with the giant-mitochondria finding wired into the CONTROVERSY's `attaches_to`.

**Round 2** produced the third-report finding above, which was filed as a non-blocking suggestion and was closer to a correction of the entry's central argument. Also the `Muscle Weakness` rename (it binds only `HP:0001324`; hypotonia has its own entry), the hepatic histology, and a `REVIEW` history record.

**Round 3** caught that my fix for round 2 had the same defect one level down: I added a data point to correct a misleading tally and stated it without the limitation needed to weigh it — the COX caveat now recorded in all three places the tally appears. Also bound `NCIT:C45961` Giant Mitochondria Present and `NCIT:C36185` Steatosis after verifying enum membership, and added the EM lipid-droplet finding.

**Round 4** approved. Its remaining suggestion — the muscle light-microscopy findings, including the diagnostically meaningful absence of ragged-red fibres — is filed as #10199 rather than pushed, by agreement with the reviewer that the optional findings had stopped converging.

## Deliberately not here

- **No prevalence rate.** Fewer than a dozen published patients and no population estimate; `CASES_IN_LITERATURE` / `ULTRA_RARE` with the counts in `notes`.
- **No per-phenotype frequencies computed from case counts.** With under a dozen patients and no cohort, a band would be arithmetic dressed as epidemiology.
- **No general mitochondrial-disease management.** The research report proposes CoQ10 and riboflavin from general primary-mitochondrial-disease practice. None of the primary sources makes a management recommendation for COXPD26, and importing that guidance would attribute to this disorder advice written for a heterogeneous group.
- **No `finding_term` on the nerve-biopsy histopathology entry.** Its content is "distinct morphological features", which no term carries; `NCIT:C117277` *Demyelination* names the process rather than the biopsy observation.
- **No `conforms_to`.** `kb/modules/mitochondrial_dysfunction.yaml` is the aging-hallmark module — mtDNA damage, mitophagy, senescence — not a primary mitochondrial-translation module. The sibling entry `Combined_Oxidative_Phosphorylation_Defect_Type_21` (TARS2) makes the same choice and independently repeats the same node chain, which reads as a missing module rather than two unrelated diseases.

## Validation

```
just validate                    ✓  70/70 snippets verified
just validate-disorders          ✓
just check-snippet-grading       ✓
just check-folded-hyphens        ✓
just check-cache-order           ✓
```

Two snippets had to be cut short of square-bracketed HGVS designations, which the reference validator strips from the query but not from the cached text — filed as #10192. The allele identities are carried in `genetic[].variants[].name` instead, with the reason recorded in the `explanation`.

## A note on what the checks could not catch

Worth recording, because it is the point of rounds 2 and 3. The entry was approved with every gate green and 65/65 snippets verified, while telling the reader that the evidence on its central controversy was evenly split when it was not. Every individual evidence item was correct; the collection was skewed. Nothing in the schema represents how many independent reports back a claim, or whether two of them are the same cohort, and a discussion's `rationale` is free text — so "two well-conducted reports disagree" validated perfectly while being wrong about the count.

Added to #9421, along with a correction to my own earlier comment on that thread, which had used this same controversy as a worked example of a two-report tie.

## Diff scope

Entry, three history records (CREATE and two REVIEW), the five PMID cache files this entry cites, the term-cache and enum rows it introduces, the deep-research report, and deletion of the stub.

---


aleixpuigb:
All four blocking findings and both suggestions are in `edff60ed`. Finding 1 was a real gap and the diagnosis of *why* it happened is worth recording.

**1. Phenotype under-coverage — fixed, ten added.** You are right, and the mechanism of the mistake is that I curated phenotypes from the *pathophysiology* outward: I picked the manifestations that hung off nodes I had already modelled, and treated the two enumerating sentences as evidence for those rather than as a list to work through. The sentences were already committed and already verified. Nothing was missing except attention.

Banded in two tiers, because the two sources support different strengths:

- `FREQUENT` — intellectual disability, cerebellar abnormality, short stature (3/3 in the 2022 series, the only cohort with per-feature counts); and muscle weakness, hypotonia, spastic diplegia, gastrointestinal dysmotility (these appear in the case report's *disease-level characterization* sentence, which describes COXPD26 across published patients rather than its own patient).
- `OCCASIONAL` — seizures, hepatic and renal involvement (the single Chinese patient only; the paper says the hepatic feature was first observed in a COXPD26 patient there).

Muscle weakness and hypotonia are separate entries rather than one, since they are distinct examination findings with distinct HPO terms and the source lists them separately.

**2. `GO:0070901` — fixed.** The `preferred_term` already said "mitochondrial tRNA methylation" and the binding said the parent. Binding the term the curator had already named is obviously right, and for a disease whose entire mechanism is compartment-specific it is more than cosmetic.

**3. PMID:33398350 — fixed by using it, as you suggested.** The `notes` claim was inaccurate: cached, described as cited, cited zero times. It now sits on `Impaired Mitochondrial Translation` with `directness: INDIRECT` / `IN_VITRO`, matching the PMID:29222331 item beside it. Your read that this is the entry's weakest node is right, and it is now better supported for a specific reason worth stating in the entry: two independent cybrid systems perturb the same modification in the same direction as each other and opposite to this disease, which is the strongest available support for a step never measured in a COXPD26 patient.

**4. `histopathology` — added.** The giant-mitochondria finding was quoted three times in prose and modelled nowhere, which is the wrong way round for the observation the `CONTROVERSY` turns on. Two entries now: the giant mitochondria (wired into the discussion's `attaches_to`), and the distinct nerve-biopsy morphology the 2022 paper is titled for.

One thing worth flagging about that block. The giant-mitochondria evidence item is `supports: SUPPORT` on the histopathology claim while the *same sentence* is `supports: REFUTE` against the respiratory-chain node. That is deliberate and is exactly the claim-relative behaviour CLAUDE.md describes — and it is also why `check-snippet-grading` gates `evidence_source` but not `supports`. I have said so in the `explanation` so it does not read as an inconsistency.

**5 and 6 — both taken.** The MTO1 quote is `MODEL_ORGANISM`; the in-vitro enzymology statement is `IN_VITRO`. Your framing is the one I should have applied: `evidence_source` classifies evidence *type*, not novelty. I had graded both on "is this a finding of this study", which is a different question and one the schema has no slot for — arguably part of what #9827 and #9421 are asking for. Your point that the yeast complementation assay stays `OTHER` for a genuinely different reason (not an animal, not cell culture) is the distinction that makes this consistent rather than arbitrary.

**Validation on the new head:**

```
just validate                    ✓  63/63 snippets verified
just validate-disorders          ✓
just check-entity-refs           ✓  (incl. the new histopathology# ref)
just check-snippet-grading       ✓  no new divergences
just check-folded-hyphens        ✓
```

Cache delta is two rows, both `GO:0070901` and its enum row. No other entry's rows swept in.


---

aleixpuigb:
Took all four suggestions. One of them was not a nit, and it is the reason I spent the approval to push.

**Suggestion b was a real accuracy problem.** The CONTROVERSY told the reader that one report found abnormal muscle enzymology and one found it normal. That is not what the literature says. PMID:35109800 — a paper this entry already cited five times — reports "Muscle biopsy also indicated reduced respiratory chain complex enzyme activities" in its own patient. So it is two reports against one, and the entry was presenting an even split that does not exist.

That sentence is now an evidence item on both the pathophysiology node and the discussion, and the rationale states the tally. I added one thing the suggestion did not ask for: the rationale now also says that the dissenting series is the largest single patient group of the three, so counting reports and counting patients do not point the same way. A bare "2 versus 1" would have replaced one misleading impression with another.

This is the same failure mode as the frequency arithmetic on #10189, and it is worth naming because a curator cannot see it from inside the file. Every snippet was verified, every term resolved, every gate was green — and the entry still misrepresented the balance of evidence, because nothing in the schema or the checks looks at how many independent reports support a claim or how many patients each one contains. The evidence items were individually correct and collectively skewed.

The other three:

- Renamed `Muscle Weakness and Hypotonia` to `Muscle Weakness`. It binds only `HP:0001324` and hypotonia has its own entry, so the name promised a bundle the entry deliberately does not contain. Nothing referenced the old name.
- Added the hepatic histology — diffuse steatosis, PAS-positive intracytoplasmic glycogen, activated Kupffer cells, portal fibrous hyperplasia. Worth having because the authors say liver pathology in this disorder had never been reported, so it is the only histological description of the hepatic phenotype in the literature. It is one patient and the block says so, consistent with the `OCCASIONAL` band on the phenotype.
- Added a `REVIEW` history record covering both rounds, including the per-source phenotype banding reasoning.

Two small things from doing the work:

The suggestion proposed `--event UPDATE`, which `new_history.py` does not accept — the enum is `GENERAL`, `CREATE`, `EDIT`, `REVIEW`, `AUDIT`. Used `REVIEW`.

I also had to fix prose I introduced: both PMID:35109800 and PMID:35342985 are 2022, and my new sentence calling one "the 2022 Chinese case report" made the entry's existing "the 2022 series" references ambiguous. The established convention in the file is "2022 series" for the neuropathy paper and "Chinese case report" for the other, so I followed it.

Validation on the new head: `just validate` 69/69 snippets verified, up from 65/65. `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-term-cache-integrity` and `validate-history` all clean. No cache churn — every new snippet comes from an already-cached PMID and no new terms were bound. Diff is the entry plus one new history record.

---

aleixpuigb:
All three suggestions taken. Then I'm stopping, and saying so explicitly rather than leaving it implied — see the last section.

**Suggestion i was the important one, and it is a caveat on evidence I added last round.** You're right: the same paragraph says *"Unfortunately, the enzyme activity of COX could not be detected in our lab."* COX is complex IV, which is the complex the founding report found most affected, so this report establishes abnormal enzymology while being silent on precisely the axis the disagreement turns on.

That is the same defect I fixed last round, one level down. I added a third data point to correct a misleading tally, and then presented that data point without the limitation needed to weigh it. It is now on both evidence items and in the rationale — deliberately in the same place as the "two against one", because a reader who has just been handed a tally needs the caveat there, not three paragraphs away.

**Suggestion ii — I checked, and accurate terms exist.** You flagged that your runner had no network and asked me to verify rather than trust, which was the right call, so here is what I found:

- `NCIT:C45961` **Giant Mitochondria Present** — an exact match for the skeletal-muscle finding.
- `NCIT:C36185` **Steatosis** — the generic morphologic term.

Both are genuine members of the `HistopathologyFindingTerm` dynamic enum, not merely correctly labelled: `validate-terms` wrote a membership row into `cache/enums/histopathologyfindingterm_*.csv` for each, alongside the label rows. That distinction matters here, because it is exactly the one that let a bad CURIE through on #10195 — a term inside `qualifiers` has no enum binding and is never checked at all (#10197).

Your trap warning was well placed. `NCIT:C82990` *Macrovesicular Steatosis* is already in the term cache and looks like the better match, but the paper says "diffuse hepatocellular steatosis" without specifying the vesicle pattern. Binding it would manufacture specificity the source does not support, so it is `C36185`.

**The nerve-biopsy entry stays unbound.** Its content is "distinct morphological features", which is not a finding any term carries. `NCIT:C117277` *Demyelination* names the process rather than the biopsy observation and would assert more than the entry claims. Two of three bound is the right outcome, not an incomplete one.

**Suggestion iii** — the EM lipid droplets are now their own evidence item. It was inside a figure-legend quote already committed but carried nowhere, and EM is a distinct modality from the light microscopy.

`just validate` 70/70 snippets verified, up from 69/69. `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-term-cache-integrity`, `check-cache-order` and `validate-history` all clean. The two new term rows and two new enum rows were written by the validator, in canonical order.

## Stopping here

This is round three of non-blocking suggestions, and I am not going to push for a fourth. Each round has been worth taking — the tally correction and this COX caveat both changed what the entry claims — but the findings are not converging toward zero, and CLAUDE.md is clear that the move at this point is to say what is still open rather than keep spending review cycles on an entry that has been approved three times.

What is still open, as far as I can see: the nerve-biopsy `finding_term`, deliberately unbound above. Everything else raised across three rounds is applied.

If a fourth round surfaces something genuinely blocking I will of course fix it. Optional findings from here I would rather see as follow-up issues against the entry than as further pushes that dismiss the approval each time.

---

Issue #9826, opened 2026-08-28
`preflight-dr` reports a disease's own effector gene as a "rival gene"


aleixpuigb:
`just preflight-dr` counts gene mentions in a deep-research report and warns when a gene other than the expected one is mentioned often, on the theory that a second frequently-named gene means the report has drifted onto a second disease. That heuristic cannot tell a rival disease's gene from a gene that is part of the intended disease's mechanism, and it fires on the second case.

## What it said

Curating ACTH-independent macronodular adrenal hyperplasia 3 (MONDO:0700299, KDM1A):

```
WARN  research/ACTH-independent_Macronodular_Adrenal_Hyperplasia_3-deep-research-openscientist.md
  intended entity : MONDO:0700299 ACTH-independent macronodular adrenal hyperplasia 3
  canonical gene  : KDM1A
  mentions        : KDM1A=48
  top genes       : KDM1A=48, GIP=21, GIPR=21, ARMC5=10, GNAS=2
  - Rival gene GIP is mentioned 21 times (44% of KDM1A). The report may mix in a second
    disease entity - exclude those sections before curating.
```

GIP and GIPR are not a second disease entity. They are the mechanism. KDM1A loss derepresses *GIPR* in adrenocortical cells, so a gut incretin gains control of cortisol secretion — that is the entire disease. A report on this MONDO term that *didn't* mention GIPR ~20 times would be the defective one.

`ARMC5` in the same list is a more interesting case: it genuinely is a different molecular group of the parent disease, and the report discusses it to draw the boundary. That is also correct behaviour for the report, and also flagged.

## Why this is worth fixing rather than living with

The check is doing something valuable — Named Entity Confusion is a real failure and this is one of the few automated defences against it. But a warning that fires on well-formed reports trains curators to skim past it, which costs exactly the case it was built for. Any disease whose mechanism runs through a second named gene will trip it: a receptor, a ligand, a fusion partner, a downstream effector, a modifier gene.

## Options

Roughly in order of how much work they are:

1. **Soften the wording.** "A second gene is mentioned frequently; check whether it is this disease's effector/modifier or a second disease" is honest about what the count can and cannot distinguish, and does not tell the curator to exclude sections that may be the most important ones in the report.
2. **Check MONDO/HGNC relations.** If the rival gene has no disease association of its own, or its associated diseases don't overlap the report's other content, it is far more likely an effector than a rival entity.
[...]

Issue #9711, opened 2026-08-27
Snippets quoted from PDF-extracted full text are not reliably quotable, and nothing records which text a snippet came from


aleixpuigb:
When a reference cache entry has `content_type: full_text_pdf`, its body is PDF-extracted text. That text carries ligatures (`ﬁ`, `ﬂ`), words broken across lines by hyphenation, and occasional collapsed inter-word spacing. Snippets still have to be exact substrings of it, so in practice large parts of such a paper cannot be quoted at all.

Concrete case from #9709. `references_cache/PMID_32562707.md` (Fournier et al., *JACI* 2021 — the second reported DEF6 family) contains:

```
zygous c.940C>Tm u t a t i o ni nDEF6 that led to a premature stop
codon Q314* in 4 members of a consanguineous family with auto-
immunity and susceptibility to EBV
```

The sentence that establishes the family and the variant is unquotable: the spacing is destroyed in one line and the next two break mid-word. `Sanger sequencing conﬁrmed the homozygous mutation in all 4 affected siblings` is unquotable for a different reason — the ligature. I ended up quoting two spans that work only because they happen to span a clean line break:

```yaml
snippet: "patients 1 and 3 also showed abnormal detectable blood EBV\nloads (>4 log copies/mL) over 6 months"
```

That verifies, and it is genuinely exact. But a reviewer looking at it cannot distinguish it from a corrupted quote, because **nothing on the `EvidenceItem` says whether the snippet came from a clean PubMed abstract or from PDF-extracted full text**. Same slot, very different reliability.

## Why this matters beyond aesthetics

The failure mode is not "the snippet looks ugly". It is that a curator facing an unquotable primary source has three options, and two of them are bad:

1. quote a weaker secondary source that has a clean abstract (loses provenance to the actual observation);
2. quote a fragment that passes the substring check but is not propositional (`check-snippet-length` catches the short ones, not all of them);
3. skip the claim.

I hit exactly this: the SKIV2L confound in #9709 is one of the most important facts about DEF6 deficiency, and it survived only because that particular sentence happened to break cleanly across three lines.

`check-folded-hyphens` already exists because of this class of problem, which suggests it has bitten before — but it catches one symptom after the fact rather than giving curators quotable text up front.

## Two things that would help, roughly independent

[...]

Issue #7906, opened 2026-08-03
Schema gap: DiseaseMappings has no OMIM slot (OMIM phenotype numbers only expressible in prose)


nlharris:
> This issue was generated by Claude.

## Problem

The `DiseaseMappings` class (`src/dismech/schema/dismech.yaml:6337`) provides only these cross-reference slots:

- `icd10cm_mappings`
- `icd11f_mappings`
- `mondo_mappings`
- `ncit_mappings`

There is **no OMIM slot**. As a result, an OMIM phenotype MIM number — which is the primary identifier much of the Mendelian-disease literature (especially myopathies and other rare disorders) keys on — can only be recorded in free-text `description`. That makes it **unqueryable across the KB** and inconsistent with how every other authoritative identifier (MONDO, NCIT, ICD) is handled.

Note that a MONDO mapping back to the entry's own `disease_term` would be circular, so MONDO is not a substitute for a real OMIM cross-reference here.

## Proposed fix

Add an `omim_mappings` slot to `DiseaseMappings`, mirroring the existing mapping slots (multivalued, holding `OMIM:######` CURIEs), so OMIM phenotype/gene numbers are first-class, machine-queryable cross-references.

## Motivating case

Surfaced by PR #7872 (Distal Myopathy 6, Adult-Onset, Autosomal Dominant — ACTN2 / MONDO:0032853 / OMIM:618655), where OMIM:618655 had to be placed in the entry description because no structured slot exists. The re-review independently confirmed the schema has no OMIM slot and that prose was the only available option given the current schema.

aleixpuigb:
Two more instances from today, both surfaced by the automated reviewer as a suggestion I then had to decline.

- **DFNA3A** ([#10075](https://github.com/monarch-initiative/dismech/pull/10075)) — OMIM #601544
- **XFE progeroid syndrome** ([#10077](https://github.com/monarch-initiative/dismech/pull/10077)) — OMIM #610965

The XFE case is the more interesting of the two, because OMIM is not incidental there. That entry's central curation decision is a lump/split call against `Xeroderma_Pigmentosum`, which already carries XP-F on the same gene. The argument rests on the ERCC4 allelic series producing five distinct entities — XP-F, Cockayne overlap, COFS, Fanconi anemia group Q, and this — and **OMIM phenotype numbers are the identifiers that series is conventionally enumerated by.** So the identifier that most directly justifies the entry's existence as a separate entry is the one that cannot be recorded.

[...]

PR #10213, opened 2026-08-31
curate: Add EVEN-PLUS Syndrome (MONDO:0014801, HSPA9)

### LINKED ISSUE #10207: Curate even-plus syndrome (MONDO:0014801)
Curate a dismech entry for **even-plus syndrome** ([MONDO:0014801](https://monarchinitiative.org/MONDO:0014801)) — a recessive skeletal dysplasia (Epiphyseal, Vertebral, Ear, Nose, plus associated findings) caused by biallelic variants in `HSPA9`, which encodes mortalin/mtHSP70, the mitochondrial matrix chaperone that drives protein import and iron-sulfur cluster biogenesis.

Preliminary read: this looks like a `DISEASE`. MONDO records no descendants, one causal gene (`hgnc:5244` HSPA9), and no existing `kb/` entry covers HSPA9 or the EVEN-PLUS phenotype. Note that `HSPA9` also underlies a distinct dominant sideroblastic anemia / EVEN-PLUS is the recessive multisystem end — worth stating the allelic-series relationship explicitly rather than merging them.

Stub: `stubs/Even-plus_Syndrome.yaml`

aleixpuigb:
Adds EVEN-PLUS syndrome — the biallelic `HSPA9` disorder of epiphyses, vertebrae, ears and nose, plus associated malformations. `HSPA9` encodes mortalin / mtHSP70, the mitochondrial matrix chaperone.

Closes #10207

## What the entry is organised around

A negative claim, and one the primary literature makes rather than one I inferred: EVEN-PLUS does not look like a bioenergetic mitochondrial disease. No lactic acidosis, no encephalomyopathy, no Leigh-type decompensation. What it looks like is disturbed embryonic morphogenesis — microtia, a flat nose, abnormal hair whorls, anal atresia, sacral agenesis.

How a mitochondrial matrix chaperone deficit produces that pattern is stated as unexplained in the source. So the chaperone-to-morphogenesis edge carries a `KNOWLEDGE_GAP` discussion and a proposed lineage-conditional knock-in experiment, rather than a plausible-sounding energy-failure chain. The deep-research report supplied exactly such a chain (reduced ATP, oxidative stress, apoptosis of embryonic precursors); it is not curated, because the primary source contradicts its premise.

## Where the evidence actually sits

The two halves of the mechanism are supported very differently, and the entry says so instead of averaging them:

- **Genotype to protein — well evidenced.** A dedicated biophysical study of the R126W and Y128C alleles measured ATP hydrolysis, interdomain-linker binding, thermostability and aggregation propensity. Graded `DIRECT`.
- **Protein to malformation — no in vivo bridge.** Homozygous `Hspa9` knockout is embryonically lethal; the heterozygote is unaffected and has normal basal haematopoiesis. The two available dosages bracket the human condition without reaching it. That is curated as a `FAILS_TO_RECAPITULATE` model link with its limitations spelled out, rather than left out because it is a negative result.

The zebrafish `crimsonless` mutant is cited twice for two different claims and graded `INDIRECT` / `MODEL_ORGANISM` both times: it was ascertained for a haematopoietic phenotype, and the observation that matters here — stunted head and eye development — is incidental to that study.

One OpenScientist run is included, and it is worth reading the two validation blocks separately, because they disagree:

| | result |
|---|---|
| references | 14/14 verified, 21/21 quotes valid, `confabulation_rate: 0.0` |
| terms | `needs_review: true`, 2 unresolved, 8 label mismatches |

No CURIE in this entry came from the report. The report's suggestions included `HP:0011003` labelled "Arched eyebrows" (it is *High myopia*) and `NCIT:C15190` labelled "Supportive Care" (it is *Needle Biopsy*), so the term validator earned its place.

What the report did contribute was two leads I had concluded did not exist, both fetched and read before use: the biophysical study above, and the knockout-lethality result. Both materially improved the entry.

**The `zygosity: HOMOZYGOUS` value on the genetic context node is a lossy summary.** The cohort contains both homozygous and compound-heterozygous patients and the slot takes one value. The node text describes both.

**Three snippets are truncated mid-sentence on purpose.** `references_cache/PMID_26598328.md` is a `full_text_xml` capture in which superscript reference markers are glued onto the preceding word — the sentence about the LONP1 relationship literally ends `LONP19101112`. I quoted up to those positions and recorded why in `notes`. Reported upstream on #9711 with a census: about 12% of sampled `full_text_xml` caches are affected.

## Validation

```
just validate kb/disorders/EVEN-PLUS_Syndrome.yaml
  schema: no issues · terms: passed · snippets: 43/43 verified
just check-entity-refs   OK
just check-duplicate-keys  OK
just preflight-dr research/EVEN-PLUS_Syndrome-deep-research-openscientist.md MONDO:0014801
```

The preflight warns about a "rival gene HP", which is the twelve `HP:` CURIEs in the report being read as haptoglobin — commented on #9826. The genuinely informative signal in that same output is `LONP1=6`, since EVEN-PLUS was originally misdiagnosed as CODAS; the entry addresses that relationship directly.

The deep-research report was renamed from `Even-plus_Syndrome-...` to `EVEN-PLUS_Syndrome-...` so it matches the entry slug and `just check-research` finds it.

---

### DIFF
--- a/history/disorders/EVEN-PLUS_Syndrome/2026-08-31T095126Z-claude-code-51625b.yaml
+++ b/history/disorders/EVEN-PLUS_Syndrome/2026-08-31T095126Z-claude-code-51625b.yaml
@@ -1,0 +1,36 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: EVEN-PLUS_Syndrome
+  path: kb/disorders/EVEN-PLUS_Syndrome.yaml
+session:
+  id: 2026-08-31T095126Z-claude-code-51625b
+  timestamp: '2026-08-31T09:51:26Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10213
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - phenotypes
+  summary: 'Review response round 3: drop unsupported obligate frequency on synophrys'
+  details: |
+    Third automated review cycle on PR 10213. One finding, and it was my inconsistency rather than a missed detail.
+
+    The round-2 pass added frequency OBLIGATE to Synophrys and Arched Eyebrows on the strength of the sentence 'All patients had synophrys or arched eyebrows, hypoplastic or dysplastic ears, hypoplastic nasal bone, and dysplastic femoral head'. That sentence puts the disjunction at twelve of twelve; it says nothing about synophrys alone, which could be anywhere from zero to twelve. I recorded the carve-out in the node description - that OBLIGATE was asserted for the compound and not for synophrys alone - but there is nowhere in the data for a carve-out to live. The frequency slot sits on a phenotype record whose machine-readable identity is HP:0000664, so any export or HPOA-style query reads it as EVEN-PLUS synophrys one hundred percent, which no source states. A prose caveat beside a machine-queryable value does not restrict the value.
+
+    What makes this worse than an oversight is that the same commit applied the correct rule to Microtia sixty lines further down, declining OBLIGATE there because hypoplastic or dysplastic ears is broader than microtia. Synophrys or arched eyebrows is broader than synophrys in exactly the same way. I wrote the right rule and then failed to apply it to the node I had just edited.
+
+    Removed the frequency line and reworded the description to state the decline the way the Microtia description does, rather than explaining a value that is no longer there. Kept the HP:0000664 binding: the reviewer's alternative was to rebind the node to HP:0000534 Abnormal eyebrow morphology, which is the least common subsumer of synophrys and highly arched eyebrow and would make the hundred percent claim literally true. That is defensible, but it trades away a correct specific binding to rescue a frequency value, and the frequency is the unsupported part. Dropping the number and keeping the term loses less.
+
+    Also rewrote the counting paragraph in notes, which had gone stale the moment the round-2 commit converted two of the four features into numbers. It still described all four as quoted rather than converted. The replacement states the actual rule the entry now applies: a frequency is asserted when the source sentence distributes over the bound term, either exactly as for hypoplastic nasal bone against HP:0004646, or by entailment as for dysplastic femoral head against HP:0002656 since the femoral head is an epiphysis. It is withheld when the sentence is a disjunction broader than the bound term, because entailment runs from narrow to broad and never the reverse. This is the same defect class as the round-2 finding: notes prose describing entry content that has moved underneath it, which nothing in the validation stack checks.
+
+    Validation on this head: linkml-validate clean, term validation passed, 58 of 58 snippets verified, cache normalization reported zero files updated, check-entity-refs and check-duplicate-keys clean, and the three baseline gates clean with no movement.
--- a/kb/disorders/EVEN-PLUS_Syndrome.yaml
+++ b/kb/disorders/EVEN-PLUS_Syndrome.yaml
@@ -57,10 +57,21 @@ notes: >
   and sibpair reports. Most phenotypes here still carry no `frequency`,
   because a percentage pooled across independently ascertained case reports
   is a scope error rather than a penetrance estimate. The exception is the
-  set of features that review found in *all* 12 patients - synophrys or
-  arched eyebrows, hypoplastic or dysplastic ears, hypoplastic nasal bone,
-  dysplastic femoral head - which is a defensible statement about a stated
-  denominator, and is quoted rather than converted into a number.
+  set of features that review found in *all* 12 patients, which is a
+  defensible statement about a stated denominator.
+
+  Two of those four are curated as `frequency: OBLIGATE`, and two are not.
+  The dividing line is whether the sentence distributes over the term this
+  entry actually binds. "Hypoplastic nasal bone" is exactly `HP:0004646`, so
+  the value is direct; "dysplastic femoral head" is narrower than
+  `HP:0002656` Epiphyseal dysplasia but entails it, since the femoral head is
+  an epiphysis, so the value is carried upward. The other two are stated
+  disjunctively - "synophrys *or* arched eyebrows", "hypoplastic *or*
+  dysplastic ears" - and each is broader than the term bound here
+  (`HP:0000664` Synophrys, `HP:0008551` Microtia). A disjunction at 100% puts
+  no floor under either limb, so those two nodes carry the quote and no
+  number. The asymmetry is deliberate: entailment runs from narrow to broad,
+  never the reverse.
 
   Allelic series, and why this entry does not merge it. HSPA9 also carries
   heterozygous variants reported in congenital sideroblastic anemia, and the
@@ -666,15 +677,15 @@ phenotypes:
 
 - category: Craniofacial
   name: Synophrys and Arched Eyebrows
-  frequency: OBLIGATE
   description: >
-    Present in every one of the 12 patients in the collated cohort, which
-    makes it as consistent a feature as the ear, nasal-bone and femoral-head
-    findings. Easy to under-record, because it reads as a minor dysmorphic
-    detail rather than a malformation. The source states the feature
-    disjunctively - "synophrys or arched eyebrows" - so OBLIGATE is asserted
-    for this compound node and not for synophrys alone; an individual patient
-    may have had only the arched eyebrows.
+    Reported across the collated cohort and easy to under-record, because it
+    reads as a minor dysmorphic detail rather than a malformation. No
+    frequency is asserted here: the 12/12 statement in the review is for
+    "synophrys or arched eyebrows", which is broader than synophrys, and a
+    broader feature at 100% does not carry the narrower one. This node is
+    bound to HP:0000664, so an OBLIGATE value here would export as "EVEN-PLUS,
+    synophrys, 100%", which no source states - the same reasoning applied to
+    Microtia below.
   phenotype_term:
     preferred_term: Synophrys
     term:


aleixpuigb:
All seven findings addressed, pushed in `0708689b`. Thank you for finding the cache reversion — it was real and I would not have caught it.

**Finding 1 (critical), and the cause is worth recording.** The row was not lost to a rebase. I built the per-PR cache diffs by splitting the cache files from my working checkout, which was sitting on `fcd2d5843` — a page-regeneration commit that predates `4555befe3`. `MONDO:0013930` was never in the copy I split from, so the "split" silently reverted it. Both files are now reset to `origin/main` and regenerated with `just validate-terms`, so the diff is purely additive:

```
 cache/enums/biologicalprocessterm_dd47231dca51.csv | 1 +
 cache/enums/diseaseterm_f618932bb23f.csv           | 1 +
 cache/go/terms.csv                                 | 1 +
 cache/hgnc/terms.csv                               | 1 +
 cache/mondo/terms.csv                              | 1 +
 5 files changed, 5 insertions(+)
```

`just check-term-cache-integrity` and `just check-cache-order` both pass. The same fix is on #10214, which had the identical defect from the same cause.

**Finding 2** — `treatments:` added. Reproductive genetic counselling with preimplantation testing is the one entry with a reported outcome (a live birth after embryo selection); orthopedic management and multi-organ surveillance are both `INDIRECT`, because no series reports outcomes in this disease and I would rather say that in the grading than in prose alone. I did not copy the report's NCIT suggestions — I checked each term against `cache/ncit/terms.csv` first, and can confirm `NCIT:C15190` is indeed *Needle Biopsy*.

**Finding 3** — PMID:38284453 fetched and consumed. You were right that the `notes:` argument was stated against a stale denominator. It now reads against the 12-case cohort, and the review's "all patients had…" sentence is quoted rather than converted into a `frequency`, which keeps the denominator attached to the claim.

**Finding 4** — Synophrys, Abnormal basal ganglia morphology and Seizure added. Synophrys is the better catch of the two: a ~100% feature that was sitting in a node description where no phenotype query would ever find it.

**Finding 5** — rebound to `HP:0004646`. Agreed without reservation; the node text said "nasal bones" and the term said "bridge".

**Finding 6** — both items now `SUPPORT`. Your reasoning is the right one: `supports` is relative to the object the item hangs on, and that object already asserts failure. I had been grading against an implicit unstated proposition, which is exactly the ambiguity that makes this slot hard. I applied the same correction on #10214.

**Finding 7** — the substrate-binding-domain splice alleles are in, and the node now says explicitly that they complicate its own nucleotide-binding-domain framing rather than quietly widening the sentence.

**Suggestions:** 9 (population → `Worldwide`) and 11 (neural crest cell grounding) applied.

**Two I am not acting on, with reasons:**

*Suggestion 8, the zygosity split.* You are right that `COMPOUND_HETEROZYGOUS` exists and that a second `genetic_context` would be lossless. I left it because the node describes one allelic spectrum, not two mechanisms, and splitting the node to carry a zygosity value would put a schema artefact into the pathograph. The node text names both configurations. Happy to split it if you would rather have the queryable value.

*Suggestion 10, OMIM in `mappings`.* Not currently possible — there is no OMIM mapping slot in the schema. `mappings` offers `mondo_mappings`, `icd10cm_mappings`, `icd11f_mappings` and `ncit_mappings`, and `grep -i omim src/dismech/schema/dismech.yaml` returns nothing outside prose. For a KB that is largely Mendelian, OMIM being unrepresentable while ICD-11-F is representable looks like an oversight rather than a decision; I would rather flag it than work around it with a `notes:` line. Say the word and I will open an issue.

**Suggestion 12, prognosis** — not added, and I want to be straight about why rather than let it look done. The research report's §11 is real, but its prognostic content is largely synthesis across the case reports rather than a followed cohort, and I could not find a sentence that states an outcome with a denominator. I would rather leave the section absent than fill it with quotes that describe severity in individual children as though they described a course. If you think a `progression:` block keyed to the lethality gradient is worth having on weaker evidence, I will add it.

57/57 snippets verified; `just check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-entity-refs` and `check-duplicate-keys` all pass.

---

aleixpuigb:
**CI status on the current head (`0708689b`): `test (3.13)` is green. The one red check is `claude-review`, and it is not this PR's.**

The review agent never started. From the job log:

```
##[error]No execution log found; the review agent did not run.
##[error]Process completed with exit code 1.
```

The workflow's own error handling names the likely causes — an unset `ANTHROPIC_API_KEY` falling back to a subscription token that has hit its weekly limit, a bad `CLAUDE_CODE_OAUTH_TOKEN`, or an aborted run. Nothing in that path touches the diff: no review was posted, so this is the reviewer failing to run rather than the reviewer finding a problem. The same workflow ran to completion on #10214 nine minutes earlier and approved it, which fits a usage-limit boundary rather than a code cause.

**I cannot re-run it.** Both routes return 403 for this integration:

```
POST /actions/workflows/claude-code-review.yml/dispatches   → 403 Resource not accessible by integration
POST /actions/runs/33374604019/rerun-failed-jobs            → 403 Resource not accessible by integration
```

So flagging rather than fixing, per the drive-to-green rules: there is no change I can make inside this PR that would make the reviewer's token work, and I am not going to widen the PR trying.

**What this blocks.** The first-round review left `CHANGES_REQUESTED`, and I pushed fixes for all seven findings in `0708689b` — the cache reversion, the treatments section, PMID:38284453, the three missing phenotypes, the nasal-bone rebinding, the `supports` grading, and the substrate-binding-domain splice alleles, with a per-finding reply above. Because `CHANGES_REQUESTED` is sticky and the re-review job died, the review gate stays closed until someone re-triggers it:

```bash
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10213
```

I will keep this PR watched on an hourly check-in until it is green and reviewed, and I will not touch the review state myself.

Local validation on this head, for what it is worth while the reviewer is down:

```
just validate kb/disorders/EVEN-PLUS_Syndrome.yaml
  schema: no issues · terms: passed · snippets: 57/57 verified
just check-title-snippets · check-snippet-grading · check-folded-hyphens   all OK
just check-term-cache-integrity · check-cache-order                        all OK
```

---

aleixpuigb:
@claude please re-review this PR.

Context for why I'm asking rather than re-running the workflow: the `claude-review` job on head `0708689b` died before the agent started (`No execution log found; the review agent did not run`), so the first-round `CHANGES_REQUESTED` is still standing against fixes that are already pushed. I detailed that in the comment above and cannot re-trigger it myself — both `workflows/claude-code-review.yml/dispatches` and `runs/33374604019/rerun-failed-jobs` return `403 Resource not accessible by integration` for this integration.

What has changed since: the same workflow has since run to completion three times on sibling PRs (#10214, #10215, #10216), so whatever usage limit or token problem caused the failure appears to have cleared. This mention is the only lever I have; it does not touch the review state.

All seven findings from the first review are addressed in `0708689b`:

1. the `MONDO:0013930` cache reversion — restored, diff now purely additive
2. `treatments:` — added, three entries
3. PMID:38284453, the 12-case systematic review — fetched and consumed, and the counting argument restated against it
4. Synophrys, Abnormal basal ganglia morphology, Seizure — added
5. `HP:0005280` → `HP:0004646` on the nasal phenotype
6. both `FAILS_TO_RECAPITULATE` evidence items regraded to `SUPPORT`
7. the substrate-binding-domain splice alleles from PMID:38281662 — added

`test (3.13)` is green on this head, 57/57 snippets verify, and the ungated whole-KB checks all pass.

One verification worth recording since it came out of the sibling reviews: I checked this entry's OMIM identifier against MONDO directly, because I got that wrong on #10215. `MONDO:0014801` carries `OMIM:616854`, which is what the description says — correct here. `HSPA9 AND GeneReviews[Book]` returns zero, so there is no GeneReviews baseline to mine either.

---

aleixpuigb:
Pushed `62a5b331`. Finding 8 fixed, suggestions 9 and 11 taken, 10 and 12 declined with reasons.

## Finding 8 — the miscounted `notes:` block

You're right, and thank you for counting rather than taking the prose at face value. It was cited once. I've added the second citation rather than trimming the paragraph, for the reason you gave: the claim the prose describes — that hspa9 loss disturbs morphogenesis *globally* rather than regionally — was carried by prose alone, and that claim is doing real work in the entry's argument against a bioenergetic reading.

The new item quotes the global-arrest sentence from the same cached paragraph, graded `SUPPORT` / `INDIRECT` / `MODEL_ORGANISM`, and its `explanation` records that the underlying result is Craven et al. 2005's, summarised by the report this entry actually caches. I didn't fetch PMID:15650063 — you're right that it's the better primary, but the quote I'm using is the summarising paper's own sentence, so citing the paper I quoted is the honest binding.

The part of your comment worth keeping is the diagnosis, not the defect: it survived a round because the PR body asserted it and the round-1 review repeated it back as praise. Neither of us opened the file to count. That's a specific failure mode — a reviewer confirming a claim about the diff *from the description of the diff* — and it is not one more validation gate would have caught, since no gate reads `notes:` prose against the entry's own content.

## Suggestion 9 — `frequency: OBLIGATE`

Taken for three of the four, and your argument for the split is right: the pooling objection in `notes:` is about percentages across independent reports and does not touch a single review's own denominator.

Two of the three needed a stated reason, now in their descriptions:

- **Epiphyseal Dysplasia** — entailed, not stated. The source gives dysplastic femoral heads at 100% (12/12); the femoral head is an epiphysis, so a narrower feature at full penetrance carries the broader node.
- **Synophrys and Arched Eyebrows** — the source states it disjunctively ("synophrys *or* arched eyebrows"), so `OBLIGATE` is asserted for the compound node and explicitly not for synophrys alone. An individual patient may have had only the eyebrows.

**Declined for Microtia**, and this is the one place your list was loose. The 12/12 statement is for "hypoplastic **or** dysplastic ears", which is broader than microtia — and unlike the femoral-head case, entailment doesn't run that way. A broader feature at 100% says nothing about the narrower one. Applying the suggestion uniformly would have asserted all twelve patients had microtia specifically, which no source states. Recorded in the node description so it isn't retried.

## Suggestion 10 — declined, `HP:0002134` stays

The snippet says "basal ganglia lesions in cerebral MRI". *Lesion* is structural language. `HP:0012751` is *Abnormal basal ganglia MRI signal intensity*, and binding it would assert a signal-intensity finding the source doesn't make — manufacturing a narrower match, which the term contract in `CLAUDE.md` forbids. I checked the alternatives in `cache/hp/terms.csv`: `HP:0007146` carries the bilaterality you already flagged as unsupported, and there is no plain "basal ganglia lesion" term. Broad-but-accurate beats specific-but-invented. Agreed it's your call to make, and I'm making it the other way.

## Suggestion 12 — declined

I grepped all the caches for a progression statement. The only candidate is *"The abnormal posture did not progressively worsen"* — one patient's posture, not a syndrome-level claim. The static-congenital framing you pointed at lives in the report's §11/§12 synthesis prose with no primary citation, which is the same thing that made your own audiologic ask unsupportable. A `progression:` block sourced from there would be the error this entry is otherwise careful about.

## Suggestion 11 — taken

Second history record added with `event: REVIEW` (`UPDATE` isn't in `HistoryEventTypeEnum`; the permissible values are GENERAL / CREATE / EDIT / REVIEW / AUDIT). The 08:18 CREATE record is left untouched — it accurately describes the session that ran.

## Your two corrections

Both accepted, and both changed what I did:

- **The audiologic treatment.** I'd left it out on the same reasoning you arrived at independently — no quotable hearing-loss finding in any cache, and the founding report cuts the other way with "apparently normal external ear duct". Nothing added.
- **The basal-ganglia snippet from `PMID_36052765`.** I'm glad you went back to it in full. The authors attribute the finding to periventricular leucomalacia; quoting it as syndrome support would have been citing a finding against its own source's reading.

Worth naming that both of these were you catching your own round-1 asks, and in one case the *reason* not to act was only visible in the unclipped sentence. That's the same class of error as finding 8, pointing the other direction.

## OMIM

Filed as a comment on **#7906**, which already exists (opened 2026-08-03) rather than a new issue. My addition to that thread is the angle this PR-set surfaced and the existing comments don't cover: OMIM in prose isn't validated by anything, and in [#10215](https://github.com/monarch-initiative/dismech/pull/10215) I put a **wrong** MIM number into an entry that then passed every gate and CI. The identifier class with no validator is exactly where a fabrication landed. That also bears on the design choice already being debated there — a `structured_pattern: "OMIM:[0-9]+"` would not have caught it, since the wrong number was a well-formed CURIE for a real OMIM entry.

## Validation on `62a5b331`

```
linkml-validate ........................ No issues found
term validation ........................ passed
snippets ............................... 58/58 verified   (was 57/57)
cache normalization .................... Files updated: 0
check-entity-refs ...................... OK
check-duplicate-keys ................... OK
check-snippet-grading .................. OK, no baseline movement
check-snippet-length ................... OK, no baseline movement
check-title-snippets ................... OK, no baseline movement
check-folded-hyphens ................... OK, no baseline movement
history record ......................... No issues found
```

## The gate

Understood on not being able to submit a formal review from that comment path — and agreed on not dismissing the standing `CHANGES_REQUESTED` regardless. I won't, and I won't merge this myself either. This push should retrigger `claude-code-review.yml` on the new head; the automatic run on the previous head failed with "No execution log found; the review agent did not run", which is why the last cycle came through the mention path as a comment. If this run fails the same way, I'll assign @cmungall rather than route around it.

---

aleixpuigb:
Pushed `bcd31826`. Finding 9 taken, your first option.

You're right, and the part worth naming is that this wasn't a detail I missed — I wrote the correct rule into the Microtia description and then failed to apply it to the node I'd edited sixty lines earlier, in the same commit. "Synophrys or arched eyebrows" is broader than synophrys in exactly the way "hypoplastic or dysplastic ears" is broader than microtia.

The deeper error is the one you put your finger on: I recorded the carve-out in the description and treated that as sufficient. It isn't. `frequency` sits on a record whose machine-readable identity is `HP:0000664`, so prose next to a machine-queryable value does not restrict the value — an export reads "EVEN-PLUS, synophrys, 100%" and never sees the paragraph. That is the same mistake I made on [#10216](https://github.com/monarch-initiative/dismech/pull/10216), where I noticed a count-versus-percent mismatch in `interpretation_bands` and wrote a disclaimer into the `interpretation` prose instead of fixing the number. Twice now I've reached for a caveat where a correction belonged. Worth recording as a pattern, not two incidents.

**Kept `HP:0000664` rather than rebinding to `HP:0000534`.** Your second option is sound — the parent genuinely is the disjunction, and the 100% claim would become literally true. I didn't take it because it rescues the number by giving up a correct specific term, and the number is the unsupported half. Dropping it loses less: synophrys stays queryable as a feature of this syndrome (the founding report documents it directly in the index patient), and nothing false is asserted about how often.

**One thing I fixed that you didn't ask for.** Removing the value left the counting paragraph in `notes:` stale — it still said all four 12/12 features were "quoted rather than converted into a number", which stopped being true when round 2 converted two of them. That's the same defect class as finding 8, and it would have been sitting in the entry with nothing in the validation stack able to see it. Rewrote it to state the rule the entry now actually applies:

> assert a frequency when the source sentence distributes over the bound term — directly (`HP:0004646` for "hypoplastic nasal bone") or by entailment (`HP:0002656` for "dysplastic femoral head", since the femoral head is an epiphysis); withhold it when the sentence is a disjunction broader than the bound term, because entailment runs narrow to broad and never the reverse.

That asymmetry is the whole content of your finding, and it's now stated once in `notes:` rather than re-derived in three separate descriptions.

Third history record added (`2026-08-31T095126Z-…`, `event: REVIEW`). The two earlier records are untouched.

**On the round-2 attribution.** You noted your suggestion listed all four features without checking which the sentence distributes over. That's fair, but I'd rather not have it land as your error — the suggestion was explicitly a suggestion, and checking which of four features a sentence actually covers is the curator's job, not the reviewer's. I checked two of them and stopped.

## Validation on `bcd31826`

```
linkml-validate ........................ No issues found
term validation ........................ passed
snippets ............................... 58/58 verified
cache normalization .................... Files updated: 0
check-entity-refs ...................... OK
check-duplicate-keys ................... OK
check-snippet-length ................... OK, no baseline movement
check-title-snippets ................... OK, no baseline movement
check-snippet-grading .................. OK, no baseline movement
check-folded-hyphens ................... OK, no baseline movement
history record ......................... No issues found
```

## OMIM

Done, as a comment on **#7906** rather than a new issue — that gap was already filed on 3 August. The angle I added is the one this PR set surfaced and the existing thread doesn't have: OMIM in prose isn't validated by anything, and the wrong MIM number I put in [#10215](https://github.com/monarch-initiative/dismech/pull/10215) passed every gate and CI. It also bears on the design choice already being debated there — the proposed `structured_pattern: "OMIM:[0-9]+"` would not have caught it, since the wrong number was a well-formed CURIE for a real OMIM entry, just not that disease. A check against the MONDO term's own `skos:exactMatch` set would, and that's available offline.

---

Issue #7523, opened 2026-08-01
linkml-reference-validator rewrites frontmatter of all ~7600 references_cache files, polluting every curation PR


Phillip-a-richmond:
On a freshly-built venv, the first run of `just validate-references` (or
`just validate`) rewrites the YAML frontmatter of **every** pre-existing file in
`references_cache/` — about **7,600 files** — from

```yaml
reference_id: DOI:10.1001/archneur.59.10.1553
```

to the quoted form

```yaml
reference_id: "DOI:10.1001/archneur.59.10.1553"
```

The content is otherwise unchanged, and the committed form on `origin/main` is
the *unquoted* one, so the rewrite shows up as ~7,600 modified files in
`git status` immediately after any validation run.

Reproduce:
```bash
uv venv --python 3.12 && uv sync
just validate kb/disorders/<anything>.yaml
git status --short -- references_cache/ | wc -l    # ~7613
```

Installed version is `linkml_reference_validator 0.2.1rc2`, which is what
`uv.lock` pins — so this is not a version drift on our side; the emitter just
doesn't round-trip what is committed.

**Why it matters:** the natural staging command for a curation PR is
`git add references_cache/` (it is what the curation docs imply, since you do
need to add your newly-fetched cache files). That sweeps all ~7,600 cosmetic
rewrites into the PR, burying a ~10-file curation diff and making review
impractical. Hit this on three parallel curation branches today; each needed

```bash
git checkout -- references_cache/
git add $(git ls-files --others --exclude-standard references_cache/)
```

as a manual workaround.

**Suggested fixes** (any one would do):
1. Make the validator's frontmatter emitter round-trip the committed form.
2. Normalise the committed cache to the quoted form in one bulk commit, so the
   working tree is stable afterwards.
3. Add a `just stage-references` helper that stages only untracked cache files,
   and document it in CLAUDE.md's staging guidance.
4. A pre-commit guard that refuses a commit touching more than N
   `references_cache/` files without an explicit override.

Option 1 or 2 is the real fix; 3 and 4 are cheap mitigations.

Phillip-a-richmond:
Root cause found for the worktree half of this, and it is worth splitting out
because it has a second, worse consequence than the cosmetic diff.

`.claude/hooks/validate_disorder_hook.py` resolves its working directory as:

```python
project_root = Path(__file__).parent.parent.parent   # line 110
...
subprocess.run(["just", "validate", str(temp_path)], cwd=project_root, ...)
```

[...]

Issue #7507, opened 2026-08-01
just validate rewrites ~7,600 unrelated references_cache/*.md files as drive-by churn


jmcmurry:
Running `just validate` on a **single** disorder file rewrites roughly **7,600** unrelated `references_cache/*.md` files, requoting the `reference_id` frontmatter field.

## Reproduce

From a clean tree:

```bash
git status --short          # clean
just validate kb/disorders/Vascular_Ehlers-Danlos_Syndrome.yaml
git status --short | wc -l  # 7623
```

The diff is uniform and cosmetic:

```diff
--- a/references_cache/DOI_10.1001_jama.2008.598.md
+++ b/references_cache/DOI_10.1001_jama.2008.598.md
@@ -1,5 +1,5 @@
 ---
-reference_id: DOI:10.1001/jama.2008.598
+reference_id: "DOI:10.1001/jama.2008.598"
```

Presumably a YAML-emitter version bump now quoting scalars containing `:`. Only files touched by the validated entry should be rewritten at all — instead the whole cache is normalized on every run.

## Why this matters

This is not just noise. `CLAUDE.md`, `initiate-new-disorder-creation`, and the git-discipline section all instruct:

```bash
git add kb/disorders/ references_cache/ research/ history/
```

That is the *correct* documented command — `references_cache/` genuinely must be committed for deterministic CI. But after a `just validate`, it now stages ~7,600 unrelated files. The guidance those docs give to prevent this ("never `git add -A` or `git add .`") does not help, because the churn lands inside a directory contributors are explicitly told to stage.

Concretely, this means:
- Every curation PR that runs validation before committing is at risk of carrying thousands of spurious file changes.
- Those PRs will conflict with each other constantly, since they all touch the same 7,600 files.
- Real cache additions (the handful of PMIDs a curator actually fetched) become invisible in review.

Parallel curation agents are especially exposed — several run concurrently on this repo and each would produce the same 7,600-file diff.

## Suggested fix

Either:
1. **Scope the rewrite** so `just validate <file>` only ever touches cache entries referenced by that file, or
2. **Make normalization explicit** — move it behind a separate `just normalize-reference-cache` target, land the ~7,600-file requote as one deliberate housekeeping commit, and leave `validate` read-only with respect to the cache.

(2) seems cleaner: a validator arguably shouldn't mutate its inputs at all. A one-off normalization commit would also stop the churn recurring, since every tree would then already be in the new format.

## Workaround

[...]

PR #10214, opened 2026-08-31
curate: Add Brachyolmia-Amelogenesis Imperfecta Syndrome (MONDO:0011018, LTBP3)

### LINKED ISSUE #10208: Curate brachyolmia-amelogenesis imperfecta syndrome (MONDO:0011018)
Curate a dismech entry for **brachyolmia-amelogenesis imperfecta syndrome** ([MONDO:0011018](https://monarchinitiative.org/MONDO:0011018)) — an autosomal recessive syndrome of short-trunk short stature with platyspondyly plus hypoplastic amelogenesis imperfecta, caused by biallelic `LTBP3` variants. LTBP3 is a latent TGF-beta binding protein, so the mechanism is about TGF-beta storage in and release from the extracellular matrix, not about a structural collagen.

Preliminary read: this looks like a `DISEASE`, and the repository has already flagged the gap. `kb/disorders/Short_Stature_Amelogenesis_Imperfecta_And_Skeletal_Dysplasia_With_Scoliosis.yaml` carries a note saying LTBP3 came up as a rival gene during its deep-research preflight and that no LTBP3 content was curated into that entry, because LTBP3 is a separate entity. This claim is that separate entity. `kb/disorders/Brachyolmia.yaml` covers the non-syndromic forms and should be cross-referenced rather than extended.

Stub: `stubs/Brachyolmia-amelogenesis_Imperfecta_Syndrome.yaml`

aleixpuigb:
Adds brachyolmia with amelogenesis imperfecta — also published as dental anomalies and short stature (DASS) — the biallelic `LTBP3` disorder.

Closes #10208

## This fills a gap the repository had already found

`kb/disorders/Short_Stature_Amelogenesis_Imperfecta_And_Skeletal_Dysplasia_With_Scoliosis.yaml` carries a note saying that LTBP3 surfaced as a rival gene during that entry's own deep-research preflight, that the material was inspected, and that no LTBP3 content was curated there because LTBP3 is a separate entity. This is that separate entity. The two entries should now be read together.

`kb/disorders/Brachyolmia.yaml` covers the non-syndromic PAPSS2 and TRPV4 forms and is the parent concept — it is cross-referenced, not extended.

## The mechanism claim

Not a structural-matrix mechanism, which is the easy wrong reading of a skeletal dysplasia caused by an extracellular-matrix protein. LTBP3 is a latent TGF-beta binding protein: it forms the large latent complex with TGF-beta 1, 2 and 3, anchors it in the matrix, and gates its release. Losing it reduces how much TGF-beta is positioned for activation.

So the disease is a bioavailability defect in the two tissues that most depend on that signal — the secretory ameloblast and the axial growth plate. The teeth and the spine, which look like two unrelated problems, are one lesion read out twice. That is the organising claim of the entry.

The ameloblast arm is specific enough to name a step: the initial aprismatic enamel layer at the dentine-enamel junction is absent, and non-prismatic enamel appears in patches on a thin irregular prismatic shell. Dentine is normal. That implicates Tomes' process, not mineralisation generally.

## The mouse is split three ways on purpose

The `Ltbp3`-null mouse is a good model of half this disease and a misleading model of the other half, so it is curated as three separately graded `ModelMechanismLink`s rather than one:

| target | relationship | why |
|---|---|---|
| secretory ameloblast function | `RECAPITULATES` (HIGH) | thin-to-absent enamel in incisors and molars, with non-palisadic ameloblasts laying a disorganised matrix |
| maxillary underdevelopment | `RECAPITULATES` (MODERATE) | shortened skull base, disproportionately short upper jaw |
| platyspondyly | `FAILS_TO_RECAPITULATE` (LOW) | the source states platyspondyly was not observed; and the mouse is osteopetrotic through defective osteoclastic resorption where patients are osteopenic |

The bone-density direction is the part worth pausing on: any inference from this mouse to human bone in this disease runs backwards. A `HUMAN_MODEL_MISMATCH` discussion attaches to the axial node and sets out the three live readings — allele dosage (null vs hypomorph), species (a murine spine is not loaded like a bipedal one), or an LTBP3 function not captured by TGF-beta bioavailability — without choosing between them. No claim about the human vertebral phenotype in this entry rests on the mouse.

## One `REFUTE` item that is doing real work

Hypodontia is inconsistent across families. The founding family's panoramic radiograph shows a full complement of teeth; other families have missing teeth or oligodontia. Rather than drop the disagreement, the phenotype carries a `SUPPORT` item and a `REFUTE` item quoting the sentence that carries each. LTBP3 was first reported in isolated oligodontia, so tooth number and enamel quality are separable outcomes of the same gene — the disagreement is the informative part.

One OpenScientist run. References validated cleanly (17/17 verified, 28/28 quotes valid, `confabulation_rate: 0.0`); terms did not (`needs_review: true`, with `HP:0012865` reported as "Broad femoral neck" when it is *Abnormal sperm head morphology*). No CURIE was taken from the report.

Two of its leads were verified and adopted:

- **PMID:35352826** for the molecular step — failure of TGF-beta–LAP–LTBP3 complex formation. This is a better citation than the review summary I had.
- **PMID:34906192** for the aortic phenotype — a directly described patient with biallelic null `LTBP3`, aortic disease, short stature and dental problems. This **replaced** a secondhand citation, which is the improvement I would most like a reviewer to check.

PMID:35352826 also reports heterozygous carriers showing DASS characteristics. That qualifies the recessive model, so `notes` says explicitly that "recessive" here should not be read as a claim that carriers are unaffected.

## Validation

```
just validate kb/disorders/Brachyolmia-Amelogenesis_Imperfecta_Syndrome.yaml
  schema: no issues · terms: passed · snippets: 47/47 verified
just check-entity-refs   OK
just check-duplicate-keys  OK
```

Two pre-existing `references_cache` files were touched by `fetch-reference` as frontmatter-only churn (`reference_id` requoting, `full_text_attempted`) and reverted, so this PR carries no drive-by cache edits — see #7507 / #7523.

---


aleixpuigb:
All six blocking findings plus suggestions 7, 8 and 9 addressed, pushed in `fd50eacd`.

**Finding 1** — same defect as #10213 and the same cause: I split the per-PR cache diffs from a working checkout on `fcd2d5843`, which predates `4555befe3`, so `MONDO:0013930` was never in the copy I split from. Reset to `origin/main` and regenerated with `just validate-terms`; the diff is now two insertions and no deletions. `check-term-cache-integrity` and `check-cache-order` pass.

**Finding 3, the dentine contradiction, is the one I most want to acknowledge.** You are right, and the way it was wrong is worse than a missing citation: I quoted PMID:35352826 twice for the mechanism while its own results sentence reports abnormal dentine, and "dentine is normal" was doing load-bearing work for the ameloblast-specific framing. It now reads that dentine was normal *in the founding family* and that abnormal dentine has since been reported, with a `REFUTE` item quoting that sentence on the enamel node. Same treatment as the hypodontia pair, which is what you asked for and what I should have done first time.

**Finding 4** — snippet extended to include "are expected to result in", regraded `COMPUTATIONAL` with `directness: INDIRECT`, and the explanation now says the modal is kept deliberately. This is the fix I am gladdest of: it was the citation the PR description pointed at as the improvement to check, and it was quoting a prediction as a result.

**Finding 2** — `treatments:` added: restorative dental rehabilitation, orthodontic/orthognathic management, aortic imaging surveillance. All `INDIRECT`, each saying why. Two of the three rest on non-syndromic AI evidence, and the aortic one inverts the direction of its source's argument (that paper argues from aortopathy to genetic testing; the entry argues from genotype to imaging), which seemed worth stating rather than hiding behind the grade.

**Finding 5** — `HP:0000327` added as its own phenotype.

**Finding 6** — Nephrocalcinosis, Delayed eruption of teeth, Dental malocclusion, and a Septal and Valvular Anomalies phenotype for the interatrial septal aneurysm / secundum ASD / tricuspid prolapse cluster. The nephrocalcinosis catch is a good one — the quote was already in the file under a different claim.

**Suggestions applied:** 7 (both platyspondyly snippets now quote the full sentence, unicode genotype included — the truncation was indeed me avoiding `−/−`, and it turns out to be unnecessary); 8 (both items on the `FAILS_TO_RECAPITULATE` link are now `SUPPORT`, same correction as on #10213); 9 (the Druze Arab founder allele and its 11p11.2-11q13.3 homozygous region are curated in `genetic:`, with a `notes:` line saying this is the one setting where community carrier screening beats case-by-case sequencing).

**Suggestion 10, the `references:` block** — not added. Not out of disagreement, but because a top-level `references:` entry with `findings: []` and no `tags` value that fits adds an identifier the entry already cites in evidence. Worth noting that `references[].tags` is a single-value enum whose only permitted value is `GeneReviews`, so a reference cannot currently be labelled as a guideline, a classification document, or anything else; I hit that on another entry in this run. Happy to add the block if you want the reference surfaced independently of its evidence items.

63/63 snippets verified. `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-entity-refs` and `check-duplicate-keys` all pass.

---

aleixpuigb:
Closing the one open question from your review rather than leaving it inferred — you noted you couldn't query PubMed from the runner and were treating GeneReviews as non-blocking on that basis. I can query it, and your judgement was right, though not for quite the reason either of us assumed.

There *is* a GeneReviews chapter that mentions LTBP3:

```
LTBP3 AND GeneReviews[Book]  →  PMID:20301776  "Geleophysic Dysplasia."
```

But that is the **dominant** LTBP3 phenotype, which this entry explicitly scopes out in `notes:` as a different dosage regime with a different mechanism and no dental involvement. So there is no GeneReviews chapter for brachyolmia with amelogenesis imperfecta / DASS, and nothing untagged here that should have been mined. The hit is the excluded entity, not this one.

Deliberately **not** pushing this as a change. There is nothing to add to the entry — the chapter covers a disease the entry already says it is not about — and `main` has `dismiss_stale_reviews` enabled, so a commit would drop your approval for no content gain.

Same check on the other two entries in this claim run, for the record: `HSPA9 AND GeneReviews[Book]` and `ROBO1 AND GeneReviews[Book]` both return zero, so #10213 and #10215 have no GeneReviews baseline either.

Your three non-blocking notes are noted and not actioned, for the same approval-cost reason: `SURGERY` would indeed fit the orthognathic treatment better than `OTHER`, the septal phenotype does name three findings and bind one, and there is no history record for the review cycle. All three are fair and all three are cosmetic relative to a full re-review cycle. If a maintainer would rather have them in before merge, say so and I'll push them together.

---

PR #10215, opened 2026-08-31
curate: Add Neurooculorenal Syndrome (MONDO:0957210, ROBO1)

### LINKED ISSUE #10209: Curate neurooculorenal syndrome (MONDO:0957210)
Curate a dismech entry for **neurooculorenal syndrome** ([MONDO:0957210](https://monarchinitiative.org/MONDO:0957210)) — a `ROBO1`-related developmental disorder combining midline brain anomalies, eye anomalies (including coloboma and anophthalmia/microphthalmia), and congenital kidney/urinary tract anomalies. The mechanism is loss of SLIT-ROBO guidance signalling across three organs that all depend on it at different points in development.

Preliminary read: this looks like a `DISEASE`. MONDO records no descendants and one causal gene (`hgnc:10249` ROBO1). Nothing in `kb/` currently covers ROBO1. The interesting curation question is whether the three organ systems share one pathograph — they plausibly do, since one guidance-receptor loss acts on commissural axon crossing, optic fissure closure, and ureteric bud positioning through the same ligand-receptor step. Worth stating that as the entry's organising claim rather than listing three unrelated organ phenotypes.

Stub: `stubs/Neurooculorenal_Syndrome.yaml`

aleixpuigb:
Adds neurooculorenal syndrome (NORS, OMIM #620305) — the biallelic `ROBO1` disorder combining syndromic CAKUT with midline brain malformation, eye anomalies and pituitary involvement.

Closes #10209

> **Erratum (2 Sep).** This body originally said **OMIM #619976**. That was wrong, and it was mine — I did not read it anywhere. The correct identifier is **#620305**: `MONDO:0957210` carries `skos:exactMatch OMIM:620305`, with `MEDGEN:1841013` and `UMLS:C5830377`. It was corrected in the entry during review and the erratum is recorded in the entry's `notes:`; I am correcting it here too rather than leaving a fabricated identifier in the PR record. Worth knowing why nothing caught it: the schema has no OMIM slot, so the number could only live in prose, and no validator reads prose. Discussed on #7906.

## A naming trap worth knowing before you read the diff

The defining paper never uses the phrase "neurooculorenal syndrome". It calls the entity *syndromic CAKUT associated with biallelic ROBO1*. The name comes from the OMIM entry created afterwards, and MONDO follows OMIM. So a text search for the disease name against its own primary source returns nothing — the entity is real and well-evidenced, but its name and its evidence live in different documents.

Every snippet here is therefore keyed on ROBO1 and on the organ phenotypes, never on the syndrome name. It's recorded in `notes:` so the next curator doesn't spend an hour thinking the citation is wrong.

## The organising claim

Not three coincident malformations, one signalling lesion read out in several places. ROBO1 is the receptor for the secreted SLIT ligands, and SLIT-ROBO signalling positions structures relative to the midline — commissural axons decide whether to cross it, the ureteric bud decides where along the nephric duct to emerge.

The renal evidence for that is unusually good, and it is human rather than borrowed: immunohistochemistry puts ROBO1 in the outer nephrogenic zone, in comma-shaped and S-shaped bodies, in control fetal kidneys at 14 and 21 weeks — and shows that zone gone, with the staining, in a 22-week fetus carrying biallelic variants. Expression and loss in the same experiment.

The neurological arm has a nice clinical readout: one patient had mirror movements with no variant in `DCC` or `RAD51`, the two established mirror-movement genes. Mirror movements are what happens when corticospinal projections that should cross the midline don't, so that is about as direct a clinical test of a decussation defect as this cohort offers.

## Two negatives curated as first-class facts

**The ciliopathy hypothesis was tested and not supported.** Cystic kidneys plus heart, brain and eye involvement looks like a ciliopathy, and the authors checked: ciliary morphology and abundance were normal in mutant collecting ducts and ureteric bud tips. That is a `MEASURES` model link with `direction: UNCHANGED`, and a `NO_EVIDENCE` item on the mechanism node — regraded from `REFUTE` during review, because a normal ciliary measurement does not contradict the SLIT-ROBO mechanism, it simply fails to bear on the ciliary one. "Looks like a ciliopathy, isn't one" is clinically useful and disappears if only positive results get curated.

**Neurodevelopmental impairment is not obligate.** The 44-year-old index patient's motor and cognitive development was normal; his only brain-MRI finding was a missing posterior pituitary bright spot. That is a `REFUTE` item against the stronger reading.

## Ascertainment, and why nothing carries a `frequency`

Nine individuals: three fetuses ascertained at termination after antenatal ultrasound, six living, one of them found at 44 while being worked up for kidney failure of unknown cause. Those two routes select for opposite ends of the severity range, so a percentage pooled across them would not describe any population that exists. The `KNOWLEDGE_GAP` on genotype-severity is the question with the most immediate clinical consequence in the entry — it's what prenatal counselling for a newly found biallelic genotype would turn on — so it's recorded as unresolved rather than smoothed over.

References validated cleanly (11/11 verified, `confabulation_rate: 0.0`); terms did not (`needs_review: true`, one obsolete, two unverifiable). No CURIE was taken from the report, and that mattered here: its suggested NCIT terms for the treatment section were wrong twice — `NCIT:C15366` proposed as *Kidney Transplantation* is actually **Platelet Transfusion**, and `NCIT:C15431` proposed as *Hemodialysis* is actually **Hematopoietic Cell Transplantation**. Correct terms were taken from `cache/ncit/terms.csv`.

The report did contribute the pituitary dimension, which the syndrome's name omits and which the primary paper mentions only in passing. It's now its own phenotype plus a hormone-replacement treatment — the one part of this syndrome with a treatment that restores function rather than managing damage. The supporting PSIS case is graded `INDIRECT` because that patient is heterozygous, not biallelic; it's evidence about the gene's pituitary role, not about this recessive syndrome.

## Validation

Final head `f2cbc3e4`, after the review round:

```
just validate kb/disorders/Neurooculorenal_Syndrome.yaml
  schema: no issues · terms: passed · snippets: 62/62 verified
just check-title-snippets · check-snippet-grading            all OK
just check-term-cache-integrity · check-cache-order          all OK
```

Cache diff is purely additive (3 insertions, 0 deletions), regenerated with `just validate-terms` against `origin/main` — see the cache reversion the reviewer caught on #10213 and #10214, which came from splitting cache files out of a stale checkout. This branch was built the correct way from the start.

---

https://claude.ai/code/session_012WXBoJGSck2D6eQVawij9b

### DIFF
--- a/kb/disorders/Neurooculorenal_Syndrome.yaml
+++ b/kb/disorders/Neurooculorenal_Syndrome.yaml
@@ -7,7 +7,7 @@ disease_term:
     id: MONDO:0957210
     label: neurooculorenal syndrome
 description: >
-  Neurooculorenal syndrome (NORS, OMIM #619976) is an autosomal recessive
+  Neurooculorenal syndrome (NORS, OMIM #620305) is an autosomal recessive
   developmental disorder caused by biallelic ROBO1 variants. It was delineated
   in 2022 from six unrelated living individuals and a family of three affected
   fetuses, assembled through GeneMatcher, retrospective phenotyping of
@@ -55,6 +55,16 @@ parents:
 - Multiple congenital anomalies syndrome
 
 notes: >
+  Identifier correction, recorded because the error was mine and was asserted
+  as fact. An earlier draft of this description gave the OMIM number as
+  #619976. It is #620305: MONDO:0957210 carries `skos:exactMatch OMIM:620305`
+  together with `MEDGEN:1841013` and `UMLS:C5830377`, and the deep-research
+  report reached the same cluster independently. Nothing in the validation
+  stack catches this, because the schema has no OMIM mapping slot - `mappings`
+  offers MONDO, ICD-10-CM, ICD-11-F and NCIT only - so the identifier can only
+  live in unvalidated description prose. The MedGen and UMLS cross-references
+  are likewise unrepresentable and are recorded here rather than structurally.
+
   Naming, and a caution for anyone searching the literature. The defining
   paper never uses the phrase "neurooculorenal syndrome" - it calls the
   entity "syndromic CAKUT" associated with biallelic ROBO1. The name comes
@@ -268,14 +278,16 @@ pathophysiology:
       allele sits next to the homodimerisation linker.
   - reference: PMID:35227688
     reference_title: "Biallelic pathogenic variants in roundabout guidance receptor 1 associate with syndromic congenital anomalies of the kidney and urinary tract."
-    supports: REFUTE
+    supports: NO_EVIDENCE
     evidence_source: MODEL_ORGANISM
     snippet: "Ciliary morphology and abundance, however, was overall unaffected on staining with acetylated tubulin in immunofluorescent microscopy"
     explanation: >
-      A tested and negative ciliopathy hypothesis. Curated as REFUTE against
-      the ciliary reading of this node, because a phenotype that resembles a
-      ciliopathy without being one is exactly the kind of claim that needs its
-      negative result on the record.
+      A tested and negative ciliopathy hypothesis. Regraded from REFUTE to
+      NO_EVIDENCE on review: the result does not contradict this node's claim
+      that SLIT-ROBO guidance signalling is reduced - it bears on a different
+      proposition, that the phenotype is ciliary, and finds nothing. It is
+      kept here because a phenotype that resembles a ciliopathy without being
+      one is exactly the kind of claim that needs its negative on the record.
   downstream:
   - target: Disturbed Ureteric Bud Outgrowth and Nephrogenesis
     description: >-
@@ -285,6 +297,18 @@ pathophysiology:
   - target: Disturbed CNS Midline and Commissural Axon Guidance
     description: >-
       Loss of the repulsive signal that governs midline crossing.
+    hypothesis_groups:
+    - slit_robo_midline_guidance_dosage
+  - target: Disturbed Pituitary Stalk and Hypothalamic-Pituitary Axis Formation
+    description: >-
+      Loss of the same guidance signal at the midline structure that becomes
+      the pituitary stalk.
+    hypothesis_groups:
+    - slit_robo_midline_guidance_dosage
+  - target: Disturbed Retinal Ganglion Cell Axon Targeting
+    description: >-
+      Loss of ROBO-mediated targeting of retinal ganglion cell axons through
+      the optic chiasm.
     hypothesis_groups:
     - slit_robo_midline_guidance_dosage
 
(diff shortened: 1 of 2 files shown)


aleixpuigb:
All six blocking items and the suggestions addressed, pushed in `f2cbc3e4`. Snippets 52/52 → 62/62.

**Item 4 — you were right and I was wrong, and I checked rather than taking anyone's word for it.**

```
$ runoak -i sqlite:obo:mondo term-metadata MONDO:0957210
skos:exactMatch:
- OMIM:620305
- UMLS:C5830377
- MEDGEN:1841013
```

The description said `OMIM #619976`. That number is not in MONDO's cross-reference cluster for this term and I cannot reconstruct where I got it — it was asserted as fact in the disease description without a source, which is the worst place for an unsourced identifier to sit. Corrected to #620305.

Worth noting *why* nothing caught it. The schema has no OMIM mapping slot at all — `mappings` offers `mondo_mappings`, `icd10cm_mappings`, `icd11f_mappings` and `ncit_mappings`, and `grep -i omim src/dismech/schema/dismech.yaml` returns nothing outside prose. So for a largely Mendelian KB, the identifier most entries are keyed on can only live in unvalidated free text, where `validate-terms` will never look at it. The MedGen and UMLS cross-references are equally unrepresentable. I've recorded the correction and that gap in `notes:` rather than quietly editing the number.

**Item 3 — both branches added, and the ocular one is deliberately hedged.** A pituitary-stalk node now sits between `Reduced SLIT-ROBO Guidance Signalling` and the endocrine phenotype, and a retinal ganglion cell axon targeting node covers the "oculo" of the name. On the ocular node I fetched PMID:18272390 and read it, and it argues against itself in a way worth carrying: `"These defects were more pronounced in Robo2 than Robo1 knockout animals, implicating Robo2 as the predominant Robo receptor in visual system development."` That makes ROBO1 loss a weaker explanation for the ocular phenotype than for the renal or commissural ones, so it is a `REFUTE` item on the node and the description says the branch is a mechanism proposal, not a demonstrated route.

**Item 2** — retargeted to the new pituitary node. You were right that hormone replacement does not treat a midline brain malformation; the target was wrong because the node it should have pointed at did not exist.

**Item 1** — all three rebound. The callosal one is the one I'd most have wanted caught: the description argues the finding is *graded* — "thinning, splenial agenesis, or partial agenesis... rather than an all-or-none one" — and then the binding asserted agenesis. The entry contradicted itself in the same block.

**Item 5** — corrected, and carried as evidence rather than only fixed in prose. The source says the bright-spot finding "was asymptomatic and remained without clinical implications as endocrine laboratory testing revealed no further abnormalities", which is now a `REFUTE` item on the pituitary phenotype. I had turned an incidental imaging finding into a diagnostic pointer.

**Item 6** — the surveillance recommendation is its own `diagnosis` entry, and both diagnosis entries now carry `NCIT:C18020`.

**Suggestions.** The ciliary regrade is right and I'd argue it generalises: `REFUTE` was wrong there because the result does not contradict the node it hangs on, it bears on a different proposition and finds nothing — so it is now `NO_EVIDENCE`. Mirror movements (`HP:0001335`) and corticospinal tract deficiency are curated.

**One place I did not follow you.** For absent pyramidal tracts you suggested `HP:0007256` *Abnormal pyramidal sign*. I used `HP:0007016` *Corticospinal tract hypoplasia* instead: the finding is a structural absence on fetal neuropathological examination, and a "sign" term asserts a clinical examination finding in fetuses that were never examined clinically. Happy to be overruled if the project prefers the sign term for consistency elsewhere.

**Not done, with reasons:**

- **gnomAD constraint in `genetic:`** — I did not add it. The figures are in the research report, but I could not verify them against gnomAD from this session, and constraint metrics stated to two decimal places are exactly the kind of number that should not be copied from an unverified secondary source into a KB. If there is an accepted route to fetch and cite them here, I'll add them.
- **`mappings:` block** — MedGen and UMLS have no slot, per item 4 above, so there is nothing structural to add. The MONDO ID is already the `disease_term`.
- **NYS8** — left out. It is a distinct allelic phenotype rather than part of this entity, and this entry already carries a `notes:` paragraph on why the monoallelic ROBO1 phenotypes are scoped out.

**GeneReviews:** I checked from this session, which does have PubMed access. No GeneReviews chapter exists for ROBO1 or neurooculorenal syndrome — consistent with your expectation for a 2022 entity with nine reported individuals.

**Validation on the new head**

```
just validate            → all passed, 62/62 snippets verified
just validate-terms      → ✅ passed
just check-entity-refs · check-duplicate-keys · check-term-cache-integrity  → OK
cache diff: unchanged (both new terms were already cached)
```

---

PR #10216, opened 2026-08-31
curate: Add Specific Antibody Deficiency (MONDO:0019093)

### LINKED ISSUE #10210: Curate immunodeficiency due to selective anti-polysaccharide antibody deficiency (MONDO:0019093)
Curate a dismech entry for **immunodeficiency due to selective anti-polysaccharide antibody deficiency** ([MONDO:0019093](https://monarchinitiative.org/MONDO:0019093)) — specific antibody deficiency (SPAD/SAD): recurrent sinopulmonary infection with normal total immunoglobulins and normal protein-antigen responses, but a failed antibody response to unconjugated pneumococcal polysaccharide. It is an IUIS-recognised predominantly-antibody deficiency and is the one entity in this claim run with no causal gene, which makes it a useful test of how the schema handles a disease defined by a functional assay rather than a locus.

Preliminary read: this looks like a `DISEASE` with a conserved pathograph — the T-independent type 2 response by marginal-zone and IgM-memory B cells is the shared mechanism, and it is what the diagnostic assay measures. It is distinct from `kb/disorders/Common_Variable_Immunodeficiency.yaml` (normal total IgG here is definitional, not incidental). The published severity tiers (mild / moderate / severe / memory phenotype) look like `has_subtypes` rather than separate entries.

Stub: `stubs/Immunodeficiency_Due_To_Selective_Anti-polysaccharide_Antibody_Deficiency.yaml`

aleixpuigb:
Adds specific antibody deficiency (SAD/SPAD, ICD-10 D80.6) — an IUIS-recognised predominantly-antibody inborn error of immunity, defined by what a patient cannot do rather than by what they lack. Total IgG, IgA, IgM and the IgG subclasses are all normal and protein-antigen responses are intact; what fails is the antibody response to unconjugated pneumococcal polysaccharide.

Closes #10210

## This entry is deliberately a test of the schema

It is the one disease in this claim run with **no causal gene**, and I curated it partly to see where that pushes back.

There is no `genetic:` section and no `inheritance:` block, because there is none to have — the defining review states the origin and molecular defect are unknown. So that the absence reads as a curated finding rather than an unfinished section, the statement itself is an evidence item:

```yaml
- reference: PMID:28588580
  supports: NO_EVIDENCE
  snippet: "The origin and underlying molecular defects of SAD are not known"
  explanation: >
    Graded NO_EVIDENCE because it is precisely a statement that no evidence
    identifies a molecular cause.
```

`category` is `Complex` rather than `Mendelian` for the same reason, and the pathophysiology chain starts at a cellular process rather than at a variant.

## The mechanism

An immature-for-age failure of the T-independent type 2 response. Unconjugated capsular polysaccharide has no peptide to present, so it cannot recruit T-cell help and must activate B cells directly. That capacity appears around age two — which is why the diagnosis cannot be made before then. SAD is, in effect, that normal infant state persisting.

`mechanistic_hypotheses` is `EMERGING`, not `CANONICAL`. The model explains the defining dissociation, the age floor and the organism spectrum, but no molecular lesion has been found; the strongest cellular correlate is reduced switched memory B cells, which is an association in a subset and involves the compartment carrying *T-dependent* memory — not obviously the one that should fail in a T-independent disorder. That tension is stated in the node rather than papered over.

## The assay is the case definition, and that is the biggest problem

`Absent Protective Anticapsular Antibody` is simultaneously the mechanism's output and the diagnostic criterion. So changing the cut-off does not reclassify patients within the disease — **it changes who has the disease**. The prevalence range in adults with chronic rhinosinusitis, 11.6% to 24% across three retrospective series of the same clinical population, is largely a range about the criterion rather than about the populations.

That is the entry's `KNOWLEDGE_GAP`, with a proposed prospective study that would validate candidate criteria against infection burden and treatment response rather than against expert opinion. Its `would_refute` outcome is the uncomfortable one: if no threshold predicts either, SAD as currently defined is a laboratory finding rather than a disease entity.

## Treatment evidence, not inflated

One effect is quantified and it is the only `DIRECT` treatment item in the entry: antibiotic courses falling from 7.9 to 0.7 per year in 22 SPAD patients on immunoglobulin replacement, over a median 46 months. Observational and uncontrolled, so the grading and the explanation both say it establishes the size of an association.

Everything else is `INDIRECT` and says why. The conjugate-vaccine protection figures (71.4% / 66.7% / 56.0% at 1, 6 and 12 months) come from 29 patients with **CVID or IgG subclass deficiency, not SPAD** — class-level evidence borrowed across a related population, and worth borrowing because it quantifies the limit as well as the benefit. Antibiotic prophylaxis is the mainstay by consensus with no standardised regimen and no efficacy study in this disease, and the evidence item says so in the same sentence that records it.

Also curated: **bronchiectasis**, with the association that makes late diagnosis costly — 122 months of diagnostic delay in patients who developed it versus 24 months in those who did not.

## One thing that cost me three failed validations

The reference validator strips square-bracketed spans before matching, so a snippet quoting `median [IQR] age was 45 [36-60] years` cannot verify. Statistical sentences are full of interquartile ranges, so this bites exactly where the quantitative evidence is. Snippets here quote spans without brackets, and one `explanation` records the constraint so the truncation doesn't read as sloppy quoting.

20/20 references verified, 11/11 quotes valid, `confabulation_rate: 0.0`; `term_validation: needs_review: true`. No CURIE taken from the report. It contributed the adult outcome literature this entry now rests on — the 55-patient cohort and the vaccine-protection data — both fetched and read before use.

## Validation

```
just validate kb/disorders/Specific_Antibody_Deficiency.yaml
  schema: no issues · terms: passed · snippets: 34/34 verified
just check-title-snippets · check-snippet-grading            all OK
just check-term-cache-integrity · check-cache-order          all OK
```

Cache diff is purely additive (2 insertions, 0 deletions), regenerated against `origin/main` rather than split out of a stale checkout — the defect the reviewer caught on #10213 and #10214.

---

### DIFF
--- a/kb/disorders/Specific_Antibody_Deficiency.yaml
+++ b/kb/disorders/Specific_Antibody_Deficiency.yaml
@@ -26,7 +26,7 @@ description: >
   why the diagnosis cannot be made before then. SAD is, in effect, that normal
   infant state persisting or recurring in an older child or adult.
 
-  Two features set this disease apart from most entries here. It has no known
+  Two features set this disease apart from most inherited immune disorders. It has no known
   causal gene - the origin and molecular defect are unknown - so its mechanism
   is described functionally rather than from a locus. And its case definition
   is contested at the level of the assay: how many serotypes, what titre, what
@@ -413,14 +413,20 @@ pathophysiology:
     supports: SUPPORT
     directness: DIRECT
     evidence_source: HUMAN_CLINICAL
-    snippet: "CD27 + IgD-IgM- switched memory B cells were considered low if their count was below 6.5% of CD19 + B cells, and CD27 + IgD + IgM + marginal zone-like B cells were considered low if below 7.4% of total CD19 + B cell"
+    snippet: "Decreased switched memory B cells, n (%)7 (13)"
     explanation: >
-      Measured rather than relayed, with the cut-offs stated. In this
-      55-patient cohort 7 patients (13%) had decreased switched memory B cells
-      and 5 (9%) had decreased marginal-zone-like B cells - so the correlate
-      is real but present in a minority, which is the honest size of it. The
-      marginal-zone figure is the only direct measurement anywhere in this
-      entry that bears on the marginal-zone limb of its own hypothesis.
+      The measured result rather than the method: 7 of 55 patients, 13%. Real
+      but a minority, which is the honest size of this correlate.
+  - reference: PMID:40097777
+    reference_title: "Diagnosis, Characteristics, and Outcome of Selective Anti-polysaccharide Antibody Deficiencies In A Retrospective Cohort of 55 Adult Patients."
+    supports: SUPPORT
+    directness: DIRECT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Decreased marginal zone-like B cells, n (%)5 (9)"
+    explanation: >
+      The marginal-zone measurement - 5 of 55, 9% - and the only direct human
+      measurement anywhere in this entry bearing on the marginal-zone limb of
+      its own hypothesis.
   downstream:
   - target: Absent Protective Anticapsular Antibody
     description: >-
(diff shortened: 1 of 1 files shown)


aleixpuigb:
All seven blocking items and all four suggestions addressed, pushed in `bf216014`. Snippets 34/34 → 49/49.

Your framing was the useful part: `PMID_28588580.md` is a full-text cache and I was citing its abstract-level sentences while leaving the passages that answer the entry's own central questions unquoted. That is a better diagnosis than "add these phenotypes", and it applies to how I read caches generally, not just to this file.

**Item 3 is a factual error and I'm glad you caught it.** The `KNOWLEDGE_GAP` rationale asserted that validating the criterion against outcome had not been done — while `PMID:40097777`, which the same entry cites twice, reports exactly that test and finds no clear association between anti-PS response level and prognosis, with at least a quarter of patients who had severe infection, bronchiectasis or an IgRT requirement sitting in the "mild" tier. I asserted an absence of data contradicted by my own cited cohort. The rationale now says so, and the observation is attached as `DIRECT` evidence.

It also changes what the entry claims rather than just filling a hole: this is the experiment's own `refuting_outcome` partially realised in an observational setting, so the severity tiers are now framed as serological descriptions rather than as a graded prediction. I kept the "there has never been a study evaluating the correlation" quote too, since a formal validation study still does not exist — the cohort observation is an unfavourable partial substitute, not a replacement.

**Item 2** — `progression:` added. Both trajectories: resolution, particularly in children, and evolution into CVID. The second complicates the `diagnosis` section's clean CVID differential exactly as you said, so the description there now reads as a statement about a moment in time rather than a permanent separation.

**Item 5** — a `biochemical` record for serotype-specific anti-pneumococcal IgG, carrying both thresholds, because they are independent and a patient can fail on either: 1.3 µg/mL per serotype, and an age-stratified proportion of serotypes. `interpretation_bands` express the working-group tiers on the proportion axis.

One deliberate omission there: **no `loinc_term`.** LOINC is not in `conf/oak_config.yaml` and has no term cache, so a code could not be validated in this repo. Writing an unverifiable identifier into a slot that renders as authoritative seemed worse than leaving it empty, so `notes` records the reason. Flagging it because a `reference_ranges` block without a LOINC code is arguably half a record, and if the project wants LOINC bindings the adapter needs adding first.

**Item 1** — Allergic rhinitis (`HP:0003193`) added, with both the paediatric RR 3.77 association and the adult cohort's diagnostic-delay framing. You were right that the entry was quoting "Asthma **and rhinitis** are also commonly reported" and then modelling only asthma.

**Item 4** — all four tiers now carry the working-group sentence that defines them, including the memory tier.

**Item 6** — azithromycin, trimethoprim and sulfamethoxazole as `therapeutic_agent`, with the practice-parameter dosing quoted. "Unstandardised" and "unnamed" being different claims is a fair distinction and I had collapsed them.

**Item 7** — the meta-commentary is out of `description`. You are right that it renders on the public disorder page; it reads as addressed to a reviewer because it was. The substance stays in `notes` and the PR body.

**Suggestions 8–11** all applied. On 8, the marginal-zone-like B-cell measurement is the one that mattered — it is the only direct measurement anywhere in the entry bearing on the marginal-zone limb of its own hypothesis, and the honest size of it is 5 of 55 patients, which the explanation now says. On 9, `notes` now states why no animal model is curated: importing the TACI mouse would quietly supply the monogenic basis this disease does not have, and the TI-2 model here rests on human observation and a developmental argument instead. On 10 you were right that the explanation claimed numbers the snippet does not contain — that is the bracket-stripping again, and the fix is to stop attributing the figures to the quote.

**Validation on the new head**

```
just validate            → all passed, 49/49 snippets verified
just validate-terms      → ✅ passed
just check-entity-refs · check-duplicate-keys · check-term-cache-integrity  → OK
cache diff: 2 insertions, 0 deletions
```

---

aleixpuigb:
Band arithmetic fixed, pushed in `2167a772`. You were right on all three parts, and the first one was a genuine numeric error rather than a modelling preference.

**The `3.0` boundary.** Three serotypes of a 23-valent panel is ~13%, not 3%, so a result of 4% of serotypes would have classified as *Moderate* when the working-group criterion makes it *Severe*. Worse, I noticed the mismatch and wrote a disclaimer into the band's `interpretation` instead of fixing the number — which is the wrong instinct, because the number is what a query reads and what renders as a coloured pill. Prose next to a wrong value does not make the value less wrong.

Rebuilt as four `reference_ranges`, one per axis:

| unit | population | bands |
|---|---|---|
| `ug/mL` | per-serotype protective concentration | none — the tiers aren't defined on concentration |
| `percent of serotypes tested` | under 6 years | deficient `[–,50)` → normal `[50,–)` |
| `percent of serotypes tested` | 6 years and over | deficient `[–,70)` → normal `[70,–)` |
| `serotypes with protective titre` | severity tier | severe `[–,3)` → moderate-or-better `[3,–)` |

The fourth is the one that does the real work: the severe tier is defined as "two or fewer serotypes", and giving the count its own axis is what stops it being expressed as a percentage it isn't. I added a check that asserts every band shares its range's unit and that bands ascend, and ran it against the result.

**Suggestions 1–4 all applied.** On 1 you were right that quoting the methods sentence undercut a `DIRECT` grade — the results rows `Decreased switched memory B cells, n (%)7 (13)` and `Decreased marginal zone-like B cells, n (%)5 (9)` are literal substrings, so the grading is now self-evident. On 3, the competing 0.35 µg/mL post-conjugate threshold is curated together with the source's caution that both rest on small cohorts; you're right that a *vaccine-dependent* protective threshold is the sharpest illustration this entry has of its own thesis, and it was one line from where I already was.

**One error of mine the tooling caught.** I typed the PMID:40097777 `reference_title` from memory as "Retrospective **Multicenter Study Of** 55 Adult Patients" when the cache says "Retrospective **Cohort of** 55 Adult Patients". `linkml-reference-validator` rejected it. Recording it because it is the same failure mode as the hallucinated OMIM number on #10215 — a value typed rather than read — and the difference in outcome is entirely that titles are validated against the cache and OMIM identifiers are not, since the schema has no OMIM slot for them to live in.

**Validation on the new head**

```
just validate            → all passed, 55/55 snippets verified (was 49/49)
just validate-terms      → ✅ passed
just check-entity-refs · check-duplicate-keys · check-term-cache-integrity  → OK
cache diff: 2 insertions, 0 deletions (unchanged)
band coherence check     → every band shares its range's unit, all ascending
```

---

aleixpuigb:
Thanks — and thank you for catching the band arithmetic. That one was the worst thing I did across this set of four PRs, and not because I missed it: I **noticed** the count-versus-percent mismatch and wrote a disclaimer into the `interpretation` prose instead of fixing the number. A machine-queryable slot that renders as a coloured pill doesn't read the prose next to it. Writing a caveat where a correction belonged is a worse failure than not spotting it, because it means I saw the problem and chose the cheaper output.

Not pushing anything further. The three remaining notes are non-blocking, `main` has `dismiss_stale_reviews` enabled, and per the repo's rules approved is the finish line — a push here would drop this approval and start another full review cycle to fix a stale paragraph. Recording the disposition instead so it isn't lost:

1. **The marker-level `notes:` paragraph is stale** — it describes the old single-range structure. You're right, and this is the same defect class as finding 8 on [#10213](https://github.com/monarch-initiative/dismech/pull/10213): prose that describes the entry's own content and no longer matches it. Nothing in the validation stack reads it. Worth a follow-up edit; it's wrong now, not just incomplete.
2. **"Moderate phenotype or better" is open above and flagged LOW/MODERATE**, so a fully normal responder lands in it. Real modelling error — the tier is the working group's, whose lowest tier is bounded by the criterion rather than by normality, but the band shouldn't inherit that.
3. **The conjugate-vaccine qualifier on the 0.35 µg/mL threshold is asserted in the `explanation` rather than quoted**, and the cache carries a clause that would support it directly. Agreed the quoted form is stronger.

Item 1 is the one I'd want fixed first if someone picks this up after merge, since a wrong `notes:` block is what a later curator will trust.

I won't merge this myself.

---

Issue #9964, opened 2026-08-28
CI lints with a different ruff version than the repo pins, so local lint checks are not predictive


cmungall:
CI runs `uvx ruff check .` with no version pin, which resolves to whatever ruff is current — **0.16.5** today. The repo's dev dependency pins **0.15.8**, which is what `uv run ruff` gives you locally. The two disagree about which rules are on, so a clean local lint is not evidence CI will pass, and vice versa.

**The concrete disagreement.** 0.15.8 enables `E402` (module import not at top of file); 0.16.5 does not. That inverts the meaning of a `# noqa: E402` comment:

| | `# noqa: E402` present | absent |
|---|---|---|
| ruff 0.15.8 (local) | ✅ suppresses a real `E402` | ❌ `E402` fires |
| ruff 0.16.5 (CI) | ❌ `RUF100` unused directive | ✅ clean |

So the two versions demand *opposite* source, and no version satisfies both.

**This is not hypothetical — it has cost time twice.**

- In #9462, commit `37be6d85d` exists solely because of this. Local `uv run ruff check` passed on the four files in question; CI failed with 9 errors, four of them `RUF100` on directives that were load-bearing under the pinned version. The author had verified against the wrong yardstick and had no way to know.
- The automated reviewer on that same PR could not verify lint in **six consecutive rounds**, since `ruff` is absent from its environment and `uvx ruff@0.16.5` needs an approval it doesn't hold. Lint was reported as author-attested every time. For a repo that otherwise has the reviewer independently re-derive nearly every claim, that's a real gap in coverage.

There is a related trap worth noting in passing: `tests/test_ci_batch_validation.py` carries a bare late import with no `noqa`, which `E402` flags under 0.15.8. So the *pinned* version currently reports an error against `main` that CI never sees.

**Suggested fix.** Make the two agree, in whichever direction the project prefers:

- Add `ruff` to the dev dependencies at the version CI resolves, so `uv run ruff check .` and CI are the same tool; or
- Pin the CI invocation (`uvx ruff@<version> check .`) to the dev-dependency version.

Either way the `# noqa: E402` directives and the `scripts/*` per-file ignore in `pyproject.toml` should be reconciled against whichever version wins, since they were written for 0.15.8's rule set.

[...]

Issue #9896, opened 2026-08-28
Two pathophysiology nodes list themselves as their own downstream target


cmungall:
Two entries contain a pathophysiology node whose `downstream` list includes itself, which asserts that the node causes itself. Nothing in the validation stack rejects this today. The two need different fixes: one self-edge is empty and can be deleted, the other carries a real citation that should be moved rather than dropped.

Found while writing [`docs/reports/physiomap-assessment-2026-08-23.md`](https://github.com/monarch-initiative/dismech/blob/main/docs/reports/physiomap-assessment-2026-08-23.md) (§4.3). Both are still present on `main`.

## Case 1 — `Autoimmune_Polyendocrine_Syndrome_Type_1.yaml`, node "Chronic Mucocutaneous Candidiasis"

The self-edge is bare:

```yaml
- target: Chronic Mucocutaneous Candidiasis
  causal_link_type: DIRECT
```

No description, no evidence. The node is already correctly reached from `Anti-Cytokine Autoantibody Production` ("Anti-IL-17/IL-22 autoantibodies cripple Th17-mediated mucosal antifungal defense"), so the self-edge adds nothing. Delete it.

## Case 2 — `Periventricular_Nodular_Heterotopia.yaml`, node "Progressive Lung Disease"

Here the self-edge is the one carrying the substance:

```yaml
# on the node itself, pointing at itself:
- target: Progressive Lung Disease
  description: FLNA-associated pulmonary developmental and interstitial disease manifests as progressive childhood lung disease.
  causal_link_type: DIRECT
  evidence:
    - reference: PMID:30547349
      reference_title: A review of filamin A mutations and associated interstitial lung disease.
      supports: SUPPORT
      evidence_source: HUMAN_CLINICAL
      ...
```

while the legitimate upstream edge from `FLNA Loss of Function and Neuronal Migration Failure` is thin and uncited:

```yaml
- target: Progressive Lung Disease
  description: FLNA loss-of-function can also manifest as severe childhood pulmonary disease.
  causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
```

So deleting the self-edge would discard a cited PMID. The fix is to merge: move the description and the evidence block onto the FLNA edge, then remove the self-reference. This is the same "merge the blocks, do not delete a block" pattern `CLAUDE.md` prescribes for duplicate YAML keys, and for the same reason — each side is somebody's curation.

[...]

PR #10218, opened 2026-08-31
Fix five recurring blockers: new-history quoting, stale checksums, broken pathograph edges, unvalidated qualifier terms

### LINKED ISSUE #9687: `just refresh-orphadata` is broken: pinned Orphadata checksums are ~6 months stale

### LINKED ISSUE #9697: Bare `downstream[].target` values are unchecked: a renamed node silently severs the pathograph

### LINKED ISSUE #9784: just new-history --details breaks on shell metacharacters (parentheses) in the text
`just new-history` interpolates its arguments unquoted, so any `--details` text containing shell metacharacters fails:

```
$ just new-history --kind disorder --slug Neonatal_Lupus_Erythematosus \
    --event CREATE --outcome changed \
    --summary "Create: Neonatal Lupus Erythematosus" \
    --details "... (research/Foo-deep-research-claude_code.md; 36 refs, 35 verified) ..."

new-history: 1: Syntax error: "(" unexpected
error: recipe `new-history` failed on line 198 with exit code 2
```

The shell sees the parenthesis as syntax, not as text. Quoting the argument at the call site doesn't help — the recipe re-expands it.

## Why it bites

A good `details` field is exactly the kind of prose that contains parentheses: file paths with qualifiers, reference counts, gene names with aliases, validator output. The docs example in CLAUDE.md is short enough to dodge it, so this only shows up once someone writes a real provenance note — at which point the failure message points at the justfile rather than at the input, and the natural (wrong) inference is that the recipe is broken rather than that the quoting is.

## Workaround

Call the script directly, which quotes properly:

```bash
uv run python scripts/new_history.py --kind disorder --slug Foo --event CREATE ... --details "text with (parens)"
```

## Suggested fix

Quote the interpolations in the recipe — `"{{args}}"` or the `just` positional-argument form — so the arguments reach `scripts/new_history.py` as single words. Worth checking sibling recipes that take free-text arguments for the same pattern.

Found while running `/claim-disease 5`; hit on the first of five history records and worked around for the rest.

### LINKED ISSUE #9897: `just refresh-orphadata` fails on a checksum mismatch for en_product1.xml, so no new ORPHA record can be cached

### LINKED ISSUE #10081: just clingen-refresh fails on a checksum mismatch, blocking all CGGV: citation
`just clingen-refresh` currently fails, so no new `CGGV:` gene-disease validity record can be generated and no entry can cite one.

```
RuntimeError: checksum mismatch for gene_validity.csv after download:
got      981ac24c39256eaf7e67a705b2fd1f232e7b7ddb5e6faf8453117b9cfed4b8a1
expected ef5280c848d7219353f017276ceb6411dcae06d4ba62f6bf1a243a49d36e8502
error: recipe `clingen-refresh` failed on line 2312 with exit code 1
```

ClinGen republishes `gene_validity.csv` as assertions are added, so the pinned sha256 goes stale on its own — this is expected drift rather than a corrupted download, and the fix is to re-pin the manifest against the current file.

## Why it is worth fixing rather than working around

It is not scoped to one entry. Any curation PR that wants to cite a ClinGen classification hits it, and the workaround is to state the classification as uncited prose — which is the thing the evidence rules exist to prevent. It came up in #10066: the H3-3A → BLBS1 relationship is classified Definitive by the Syndromic Disorders GCEP, that claim could not be sourced, and it was removed from the entry rather than left unbacked.

## Suggested fix

Re-pin `gene_validity.csv` in the ClinGen manifest against the current download and rebuild the affected `references_cache/CGGV_*.md` files, checking the diff for content drift rather than assuming a clean re-pin. Worth considering separately whether a stale pin should fail hard or warn — a checksum that goes stale on the source's normal publishing cadence blocks curation every time it moves, which may not be the intended trade.

Found while curating BLBS1 (#10066); no change needed in that PR.

### LINKED ISSUE #10112: Prefixed `downstream[].target` values break the pathograph: 174 edges across 30 entries, one of which drops to 0/7 phenotypes connected

### LINKED ISSUE #10148: just new-history fails on any multi-word argument (unquoted {{args}} in the recipe)
`just new-history` is unusable as documented. The recipe passes `{{args}}` unquoted, so every multi-word argument is word-split by the shell.

Reproducing it with the exact invocation from `CLAUDE.md`:

```
$ just new-history --kind disorder --slug Marinesco-Sjogren_Syndrome --event CREATE --outcome changed \
    --summary "Create: Marinesco-Sjogren Syndrome" --agent-tool claude-code --model claude-opus-5 \
    --details "De novo curation (MONDO:0009567, SIL1)."

uv run python scripts/new_history.py --kind disorder --slug Marinesco-Sjogren_Syndrome --event CREATE
  --outcome changed --summary Create: Marinesco-Sjogren Syndrome --agent-tool claude-code
  --model claude-opus-5 --details De novo curation (MONDO:0009567, SIL1).
new-history: 1: Syntax error: "(" unexpected
error: recipe `new-history` failed on line 198 with exit code 2
```

Note the quotes are gone in the echoed command. The parenthesis in `--details` is what makes the shell abort, but the damage is done earlier: `--summary` has already been split into three words, so even a parenthesis-free `--details` produces a wrong record rather than an error. That is the worse failure mode, because it is silent.

`CLAUDE.md` documents this recipe as the way to scaffold a history record, and every KB-entry PR is supposed to add one, so this is on the main curation path. Workaround is to call `scripts/new_history.py` directly, which works.

**Not new.** The reviewer notes on https://github.com/monarch-initiative/dismech/pull/7293 flagged it — "`research-disorder` and `new-history` recipes pass `{{args}}` unquoted, so any multi-word argument is split on whitespace and the recipe fails" — and it is still there. Filing it as its own issue so it is trackable rather than buried in a PR body.

Hit five times today while curating https://github.com/monarch-initiative/dismech/pull/10142, https://github.com/monarch-initiative/dismech/pull/10143, https://github.com/monarch-initiative/dismech/pull/10144, https://github.com/monarch-initiative/dismech/pull/10145 and https://github.com/monarch-initiative/dismech/pull/10146.

The fix is quoting the expansion in the recipe. Worth checking `research-disorder` and any other recipe using the same `{{args}}` pattern at the same time.

### LINKED ISSUE #10150: just refresh-orphadata fails on a checksum mismatch, blocking structured-rebuild-orphanet
`just refresh-orphadata` fails because Orphadata has republished `en_product1.xml` since the pin in `data/orphadata/MANIFEST.yaml`:

```
RuntimeError: checksum mismatch for en_product1.xml after download: got
df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb, expected
fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

Because the XML is gitignored, this also blocks `just structured-rebuild-orphanet`, which is the only way to add an ORPHA record that is not already in `references_cache`:

```
$ just structured-rebuild-orphanet --id 559
FileNotFoundError: data/orphadata/en_product1.xml
```

The checksum guard is doing its job — the pin exists so a refresh cannot silently pull different content — so this is a "the pin needs updating" bug, not a "remove the check" one.

## Practical effect

Only 347 of the ~8,823 ORPHA records are committed to `references_cache`. Every disorder outside that set is currently uncitable to Orphanet, and Orphanet is the KB's usual source for a citable prevalence class on rare diseases. In https://github.com/monarch-initiative/dismech/pull/10142 (Marinesco-Sjogren) the reviewer correctly pointed out that `ORPHA:559` should be reachable via `structured-rebuild-orphanet`; it is not, and the entry ships without a prevalence record as a result. The same will apply to any rare-disease entry whose ORPHA record is not already cached.

Note the failure mode is a little unhelpful from the curator's side: `just fetch-reference ORPHA:559` returns `No source found for reference type: ORPHA:559`, which reads like ORPHA is unsupported rather than like "use the structured-source path". The structured path then fails on a missing gitignored file, which reads like a broken checkout. Neither message points at the actual cause.

## Suggested fix

1. Re-pin `MANIFEST.yaml` to the current Orphadata release — download, verify the content is the expected shape, record the new sha256 and snapshot date, and commit.
2. Consider whether `refresh-orphadata` should say *what* to do on mismatch ("Orphadata has republished this file; re-pin MANIFEST.yaml after review") rather than raising a bare `RuntimeError`. The mismatch is expected periodically, so the message is worth writing for the case that will recur.
3. Optionally, have `fetch-reference` recognise `ORPHA:` and redirect to the structured-source recipe rather than reporting no source.

Because a re-pin changes what every future `structured-rebuild-orphanet` produces, it seems like a maintainer decision rather than something to slip into a curation PR — which is why this is an issue rather than a patch.

### LINKED ISSUE #10159: `just new-history` drops all quoting, so the documented example in CLAUDE.md fails
The `new-history` recipe passes its arguments to the script by text interpolation rather than as real arguments, so every quoted string is re-split on whitespace and any shell metacharacter is interpreted. Copy-pasting the example from CLAUDE.md fails. Any `--summary` or `--details` containing a space, parenthesis, or colon breaks.

This bites every agent writing a history record, which CLAUDE.md asks for on any PR that touches a KB entry.

## Reproduction

The command documented at CLAUDE.md line 917:

```
$ just new-history --kind disorder --slug Asthma --event CREATE --outcome changed \
    --summary "Create: Asthma" --agent-tool claude-code --model claude-opus-5 \
    --sections phenotypes,pathophysiology,evidence --pr 5123 \
    --details "What was curated and how it was validated."

new_history.py: error: unrecognized arguments: Asthma was curated and how it was validated.
error: recipe `new-history` failed on line 198 with exit code 2
```

The quotes are gone, so `--summary "Create: Asthma"` arrives as two arguments and everything after it shifts.

A `--details` value with parentheses fails differently, because the interpolated text is evaluated by the shell:

```
$ just new-history --kind disorder --slug Asthma --event CREATE --outcome changed \
    --summary "test" --agent-tool claude-code --details "A detail with (parentheses) in it."

new-history: 1: Syntax error: "(" unexpected
```

Ordinary curation prose contains parentheses constantly — gene symbols, abbreviations, PMIDs with author names.

## Cause and fix

`project.justfile:197`:

```just
new-history *ARGS:
    uv run python scripts/new_history.py {{ARGS}}
```

`{{ARGS}}` is textual interpolation. The justfile already sets `set positional-arguments := true` at the top, and its own header comment explains this is exactly what that setting is for:

> Pass recipe arguments to shell recipes as real positional arguments (`$1`, `$@`) instead of only via `{{...}}` text interpolation. This lets recipes iterate file lists safely with `for f in "$@"`, so paths containing shell metacharacters (e.g. an apostrophe in `Bell's_Palsy.yaml`) no longer break the generated script. See issue #5525.

So the fix is the pattern that issue already established — make it a shell recipe and use `"$@"`:

```just
new-history *ARGS:
    #!/usr/bin/env bash
    uv run python scripts/new_history.py "$@"
```

Worth grepping for other `*ARGS` pass-through recipes with the same shape while fixing this one.

## Smaller documentation point

Neither CLAUDE.md nor the recipe comment lists the valid `--event` values. They are `GENERAL`, `CREATE`, `EDIT`, `REVIEW`, `AUDIT` (default `EDIT`), discoverable only via `--help`. An agent recording an edit to an existing entry will reasonably guess `UPDATE` and get an argparse error. Adding the enum to the CLAUDE.md history section would save a round trip.

Found while adding a `differential_diagnoses` block to Fanconi anemia (2837ee824); I worked around it by calling `scripts/new_history.py` directly.

### LINKED ISSUE #10197: Terms inside `qualifiers` are never validated, so a nonexistent CURIE with a fabricated label reaches main silently

cmungall:
## Summary

Five separate problems, each of which had been reported more than once or had let bad data reach `main`. Two of them turned out to be silent — nothing in the validation stack objected, so the only symptom was wrong output.

The two new checks here are gates for exactly those silent failures, and each one is accompanied by the repair of the defects it found.

Closes #9784, closes #10148, closes #10159, closes #9687, closes #9897, closes #10150, closes #10081, closes #10112, closes #9697, closes #10197. Refs #9896, refs #9964.

## 1. `just new-history` mangled its arguments

Three bug reports, one line. The recipe passed arguments through `{{ARGS}}`, which pastes the raw argument text into the generated shell line — so the caller's quoting is gone before the shell parses it. Since `--summary` and `--details` take prose, this broke constantly:

| What you typed | What happened |
|---|---|
| `--summary "Create: Asthma"` | arrived as two arguments (#10148) |
| `--details "Bell's Palsy"` | `Syntax error: Unterminated quoted string` (#9784) |
| `--details "costs $HOME"` | `$HOME` expanded (#10159) |

Forwarding `"$@"` fixes all three. Verified end to end with a summary containing a colon and a details string containing an apostrophe, a semicolon and a literal `$HOME` — all land in the record unmodified. The note in `docs/dataset-curation.md` telling scripted callers to bypass the recipe is now removed.

## 2. Pinned checksums go stale on a schedule

`just refresh-orphadata` and `just clingen-refresh` broke four times in one week. This is structural, not bad luck: the manifests pin a sha256 against an **unversioned** URL. `en_product1.xml` is always the *current* Orphanet release, so the pin is guaranteed to stop matching the next time Orphanet publishes, forever. Recovery was a four-step manual ritual, which is why each occurrence became its own issue.

The check is worth keeping — a source changing under a curator is what the pin is *for* — so instead of weakening it, accepting a new release is now one command:

```bash
just refresh-orphadata --repin
```

It records the new checksum, size and snapshot date, preserves the manifests' licence and provenance comments, and leaves a diff of only the changed lines. It is never implicit: the default still refuses, and the error now names the command instead of just printing two hashes. When there is nothing to repin it says so rather than exiting silently, and it refuses outright when the download has lost more than half its pinned size — a truncated transfer and a new release are otherwise indistinguishable by checksum alone.

## 3. 175 causal edges were pointing at phantom nodes

`downstream[].target` and its siblings hold a **bare node name**, matched verbatim — there is no resolution step and no prefix handling. 175 targets across 32 entries had been written with the `<kind>#<name>` entity-reference grammar, which those slots do not accept.

Nothing caught it: the entries passed schema validation, term validation and snippet verification, and the pages rendered.

**The symptom is not a missing arrow.** `build_causal_graph` appends every declared edge unconditionally, so the edge count never moves. The unresolved target lands in `orphan_targets`, and the renderer draws it as a *phantom duplicate node* — red fill, dashed red border, labelled with the literal `phenotypes#Bleeding tendency` string — while the real node, now with no edges, drops out of the graph entirely:

```
Plug_Failure -.-> phenotypes#Bleeding_tendency
style phenotypes#Bleeding_tendency fill:#fee2e2,stroke:#dc2626,stroke-dasharray: 5 5
```

So the chain is fragmented *and* carries a bogus node. `Hereditary_Spastic_Paraplegia_49` was left with 0 of 7 phenotypes connected.

```yaml
downstream:
- target: Bleeding tendency          # correct
- target: phenotypes#Bleeding tendency   # WRONG - draws a phantom node
```

All 175 are repaired. The diff is `target:` lines and nothing else — no claim, evidence item, description or ontology binding was touched. The measure of the repair is orphan targets across the changed entries: **175 → 2**, with `Hereditary_Spastic_Paraplegia_49` going 10 → 0. The two survivors are pre-existing baselined danglers.

`scripts/check_causal_targets.py` gates it from here, whole-KB and ungated for the same reason `check-entity-refs` is: CI selects pytest by changed path, so a curation PR touching only `kb/` skips the tests written to protect KB content. It also catches the mirror-image failure in #9697 — rename a node and every bare target pointing at it dangles, just as quietly.

The gate resolves targets against `dismech.graph.collect_graph_nodes`, the same function `build_causal_graph` uses, rather than a copied list — an earlier version restated the node sections and had already drifted, missing the animal-model and variant nodes the graph also adds.

Findings are split three ways because they have different fixes:

- **prefixed** — mechanically fixable, so never excused.
- **dangling** — names nothing in the entry; needs a curator to say what was meant, so the 126 existing cases are grandfathered in `tests/causal_target_baseline.txt` and only new ones fail. Rows that stop matching anything are reported so the grandfathered count stays an honest measure of remaining debt.
- **self** — reported, never gated. See the note on #9896 below.

## 4. Ontology terms inside `qualifiers` were never checked

`linkml-term-validator` validates slots whose range is bound to an ontology-backed dynamic enum. `Qualifier.predicate` and `Qualifier.value` are plain `Descriptor`s with no such binding, so their `term:` blocks were invisible to it. Putting `label: Totally Bogus Fabricated Label` on one and running `just validate-terms` reports **"✅ Validation passed"**.

That blind spot had already let **32 wrong bindings** into `main` across 10 entries — 31 CURIE corrections plus one where the code was right and only the label was wrong. All were plausible-looking codes naming an entirely different concept:

| Curated as | The code actually means | Count |
|---|---|---|
| Therapeutic Agent | Recombinant Neurotrophic Factor | 19 |
| vancomycin | Azacitidine (an antineoplastic) | 1 |
| budesonide | Panobinostat | 1 |
| anticonvulsant agent | Lidocaine | 1 |
| Broad Spectrum Antibiotic | Arthritis | 1 |
| Tooth Extraction | Breast Extraskeletal Osteosarcoma | 1 |
| Tracheostomy | Ambulation Difficulty | 1 |
| fidaxomicin | Cellular Changes Resembling Foveolar Epithelium Cells | 1 |
| Fluid Replacement Therapy | Phase I/II Trial | 1 |
| Gastrointestinal Route of Administration | **Rectal** Route of Administration | 1 |
| Drainage Procedure | Nutritional Assessment | 1 |
| vaccine | Immunodiagnostic Procedure | 1 |
| dimercaprol | maleamic acid | 1 |
| Immunologic Factor | Biological Agent (label-only fix) | 1 |

All 32 are repaired. Each replacement was resolved against the authoritative ontology and then cross-checked against the OLS API directly rather than only through the OAK adapter, because the adapter was the thing under suspicion. Where no exact-label term exists in NCIT, the binding generalizes to the most specific accurate term and the lost specificity stays in `preferred_term`, as the Ontology Term Contract asks. So the diff is `term.id` and `term.label` lines only.

Two needed a judgement call rather than an exact match, flagged here so they can be argued with:

- NCIT has no "Gastrointestinal Route of Administration", so `Collagenous_Sprue`'s budesonide is bound to `NCIT:C38209` Enteral Route of Administration — the faithful mapping of the curator's "gastrointestinal route". Narrowing to Oral would assert a formulation detail the curator did not.
- `NCIT:C307`'s code was right and only its label was wrong, so that one is a label-only correction to "Biological Agent".

`scripts/check_qualifier_terms.py` gates it. The gate is offline and cache-first, so it only sees CURIEs already cached from elsewhere in the KB; `just check-qualifier-terms-online` resolves the rest. Prefixes with no adapter in `conf/oak_config.yaml` — today `RO` and `PR`, 109 terms — are reported as unvalidatable rather than silently passed, derived from that config so adding an adapter moves them into the checked set automatically.

## 5. CI and pre-commit lint with different ruff versions (#9964)

Not a planned part of this PR — it failed the branch, so it had to be fixed here. CI runs `uvx ruff check .` unpinned (currently **0.16.5**); pre-commit pins **0.11.3**. They disagree in both directions: 0.16.5 flags `# noqa: E402` as an unused directive, and 0.11.3 flags the import the moment you remove it.

Resolved by removing the need for the suppression: ruff exempts an import preceded by a `sys.path` preamble, but an intervening assignment breaks the exemption. Both versions are now clean. The underlying version skew is left alone — `pyproject.toml` records a deliberate choice to float ruff, and the `scripts/*` per-file-ignore shows this being handled the same way before.

**#9896 is not a curation error, so I did not "fix" it.** Both self-referential targets are a pathophysiology node and a phenotype *sharing a name*, which the flat node namespace collapses into one node — turning a legitimate mechanism→phenotype edge into a self-loop. One of them carries its own PMID evidence. Deleting them would destroy curated content to satisfy a graph-model bug, so the checker reports them without gating. That call is already earning its keep: `Lethal_Congenital_Contracture_Syndrome`, merged into `main` while this PR was open, adds three more, and a gate would have turned this PR red for somebody else's entry.

**One prefixed target is left in place.** `Charcot-Marie-Tooth_Disease_Type_4K` points a `downstream` edge at `imaging_findings#Putaminal and Periaqueductal Lesions`. That names real curated content, but `imaging_findings` is not a graph node section, so stripping the prefix would convert a visible error into a silent one. Promoting that section to the pathograph affects 110 entries and is a modelling decision for a maintainer, not something to slip into this PR. It sits in the baseline.

**History records carry no `model` field**, which CLAUDE.md asks for. This agent's configuration prohibits writing model identifiers into repository artifacts, so it is deliberately absent rather than overlooked. That is a standing conflict for any agent-authored history record here, not a one-off, and is worth a maintainer reconciling one side or the other.

## Verification

- Full KB data suite: **46,878 passed**. `Build and test` completed green on this branch.
- Graph, export and gate tests: 129 passed over the affected modules.
- All 42 changed KB files re-validate against the `Disease` class; all 42 history records validate against `HistoryRecord`.
- `ruff check` clean under **both** 0.11.3 and 0.16.5; changed files formatted.
- All four whole-KB gates green against this branch merged with current `main` — which matters, because three disorder entries landed in `main` after the baselines were generated.

## Known gaps this does not close

- 126 dangling pathograph targets remain baselined. Each needs a curator to say what the target was meant to be; the baseline should only ever shrink.
- `RO` and `PR` have no adapter in `conf/oak_config.yaml`, so 109 qualifier terms cannot be validated by any current tooling.
- `just validate-graphs` is broader than the new check — it also checks edge *sources* — but is wired into neither `qc` nor CI and reports issues in 73 disorders, so it still cannot gate. Burning down the baseline is the path to changing that.

### DIFF
--- a/scripts/check_qualifier_terms.py
+++ b/scripts/check_qualifier_terms.py
@@ -49,7 +49,24 @@ tooling, offline or not. They are reported separately rather than folded into
 ``unverified``, because the fix is a config decision, not a cache refresh.
 
 ``--resolve`` consults the configured OAK adapters for the uncached CURIEs. It
-needs network and is not part of the gate; use it when auditing, not in CI.
+needs network and is not part of the gate; use it when auditing, not in CI. It
+splits its answers three ways, because they are three different claims:
+
+``wrong_label``
+    Resolves, names something else. A finding; exits non-zero.
+
+``missing``
+    The ontology answered 404 -- the CURIE does not exist. Also a finding, and
+    the most serious one here: a label can be corrected, a fabricated CURIE
+    cannot. Exits non-zero, so ``--resolve`` cannot report two invented CURIEs
+    and still finish on a line that reads as a pass.
+
+``unreachable``
+    Any other lookup failure, and the case where the adapter answers with no
+    label at all. Reported, never gated -- this is evidence about the network,
+    not about the binding, and treating it as a defect is what once made six
+    consecutive runs against an unchanged tree report 8, 2, 2, 0, 9 and 12
+    "findings", not one of them real.
 
 Usage
 -----
@@ -63,6 +80,7 @@ from __future__ import annotations
 import argparse
 import csv
 import sys
+import time
 from collections import Counter
 from pathlib import Path
 from typing import Any, NamedTuple
@@ -229,28 +247,37 @@ def _is_not_found(exc: BaseException) -> bool:
 
 def resolve_remote(
     terms: list[Term], *, attempts: int = 3
-) -> tuple[list[tuple[Term, str]], list[tuple[Term, str]]]:
+) -> tuple[list[tuple[Term, str]], list[tuple[Term, str]], list[tuple[Term, str]]]:
     """Check uncached CURIEs against the configured OAK adapters. Needs network.
 
-    Returns ``(wrong_label, unresolved)``. The split matters: a CURIE that does
-    not exist and a CURIE we could not reach are different claims, and folding
-    them together makes the tool accuse correct bindings whenever the network is
-    flaky. This script already keeps that distinction for prefixes with no
-    adapter, and the same honesty is owed here -- six consecutive runs against an
-    unchanged tree once produced 8, 2, 2, 0, 9 and 12 "findings", every one a
+    Returns ``(wrong_label, missing, unreachable)`` -- three outcomes, because
+    they support three different claims and only two of them are about the
+    binding:
+
+    ``wrong_label``
+        The CURIE resolves and names something else.
+    ``missing``
+        The ontology answered 404: the CURIE does not exist. Definitive, and the
+        most serious thing this check can say.
+    ``unreachable``
+        The lookup failed for any other reason, or the adapter returned no label.
+        This is evidence about the network, not about the binding.
+
+    Folding the last into either of the first two is what made the tool accuse
+    correct bindings whenever the link was flaky -- six consecutive runs against
+    an unchanged tree once produced 8, 2, 2, 0, 9 and 12 "findings", every one a
     transport failure and not one a real mismatch.
 
     Transport failures are retried; a 404 is definitive and is not.
     """
-    import time
-
     from oaklib import get_adapter
 
     conf = yaml.safe_load((ROOT / "conf" / "oak_config.yaml").read_text())
     adapters_conf = conf["ontology_adapters"]
     adapters: dict[str, Any] = {}
     wrong: list[tuple[Term, str]] = []
-    unresolved: list[tuple[Term, str]] = []
+    missing: list[tuple[Term, str]] = []
+    unreachable: list[tuple[Term, str]] = []
     checked = 0
     for term in terms:
         prefix = term.curie.split(":", 1)[0]
@@ -266,6 +293,7 @@ def resolve_remote(
 
         actual: str | None = None
         reason = ""
+        not_found = False
         for attempt in range(attempts):
             try:
                 actual = adapters[prefix].label(term.curie)
@@ -273,6 +301,7 @@ def resolve_remote(
                 break
             except Exception as exc:  # classified by _is_not_found below
                 if _is_not_found(exc):
+                    not_found = True
                     reason = "does not exist in the ontology (404)"
                     break
                 reason = f"could not be reached ({type(exc).__name__})"
@@ -280,14 +309,19 @@ def resolve_remote(
                     time.sleep(2**attempt)
         checked += 1
 
-        if reason:
-            unresolved.append((term, reason))
+        if not_found:
+            missing.append((term, reason))
+        elif reason:
+            unreachable.append((term, reason))
         elif actual is None:
-            unresolved.append((term, "the adapter returned no label"))
+            # The adapter answered without raising but has no label for this
+            # CURIE. That is not a definitive not-found -- an ontology can carry
+            # a node with no label -- so it stays advisory rather than gating.
+            unreachable.append((term, "the adapter returned no label"))
         elif actual != term.label:
             wrong.append((term, actual))
     print(f"resolved {checked} uncached CURIE(s) against OAK.", file=sys.stderr)
-    return wrong, unresolved
+    return wrong, missing, unreachable
 
 
 def main() -> int:
@@ -306,16 +340,17 @@ def main() -> int:
     wrong, unverified, unconfigured, ok = classify(terms, cache)
 
     resolved = 0
-    unresolved: list[tuple[Term, str]] = []
+    missing: list[tuple[Term, str]] = []
+    unreachable: list[tuple[Term, str]] = []
     if args.resolve:
-        remote_wrong, unresolved = resolve_remote(unverified)
+        remote_wrong, missing, unreachable = resolve_remote(unverified)
         resolved = len(unverified)
         wrong = wrong + remote_wrong
         # Everything resolved and not flagged is now verified, just against the
         # ontology rather than the cache. Fold it into `ok` so the summary counts
         # what was actually checked instead of only the cache-backed subset.
         # Terms we could not reach are neither verified nor wrong.
-        ok += resolved - len(remote_wrong) - len(unresolved)
+        ok += resolved - len(remote_wrong) - len(missing) - len(unreachable)
         unverified = []
 
     if args.report:
@@ -324,6 +359,9 @@ def main() -> int:
         )
         print(f"  label verified against cache : {ok}")
         print(f"  label WRONG                  : {len(wrong)}")
+        if args.resolve:
+            print(f"  CURIE does not exist (404)   : {len(missing)}")
+            print(f"  could not be reached         : {len(unreachable)}")
         print(f"  not cached (no opinion)      : {len(unverified)}")
         for prefix, n in Counter(
             t.curie.split(":", 1)[0] for t in unverified
@@ -348,24 +386,42 @@ def main() -> int:
             f"in conf/oak_config.yaml ({prefixes}); no tooling can validate them.\n"
         )
 
-    if unresolved:
-        # Deliberately not folded into the wrong-label findings below. "We could
-        # not check this" and "this binding is wrong" are different claims, and
-        # conflating them is what made a flaky link look like a fabricated CURIE.
+    if unreachable:
+        # Deliberately not a finding. "We could not check this" and "this binding
+        # is wrong" are different claims, and conflating them is what made a
+        # flaky link look like a fabricated CURIE.
         print(
-            f"note: {len(unresolved)} qualifier term(s) could not be resolved. A 404 "
-            "means the CURIE does not exist and is worth acting on; anything else "
-            "says only that the lookup failed, so re-run before believing it.\n"
+            f"note: {len(unreachable)} qualifier term(s) could not be reached. That "
+            "says only that the lookup failed, not that the binding is wrong, so "
+            "re-run before believing it.\n"
         )
-        for term, reason in unresolved:
+        for term, reason in unreachable:
             print(f"  {term.path}\n     {term.curie} ({term.label!r}): {reason}")
         print()
 
-    if not wrong:
+    if not wrong and not missing:
         source = "the ontology" if args.resolve else "the ontology cache"
         extra = f" ({resolved} resolved online)" if resolved else ""
         print(f"OK: {ok} qualifier term label(s) match {source}{extra}.")
         return 0
+
+    if missing:
+        print("Qualifier term CURIE(s) that do not exist in the ontology.\n")
+        print(
+            "The ontology answered 404: there is no such term. A curated CURIE that\n"
+            "resolves to nothing is a fabrication, not a stale label, and cannot be\n"
+            "repaired by correcting the label -- the binding needs a real term.\n"
+        )
+        for term, reason in missing:
+            print(f"  {term.path}")
+            print(f"     qualifiers.{term.role}: {term.curie}")
+            print(f"     curated label : {term.label!r}")
+            print(f"     ontology says : {reason}")
+        print()
+
+    if not wrong:
+        print(f"{len(missing)} finding(s).")
+        return 1
 
     print("Wrong ontology label(s) on qualifier terms.\n")
     print(
@@ -379,7 +435,7 @@ def main() -> int:
         print(f"     qualifiers.{term.role}: {term.curie}")
         print(f"     curated label : {term.label!r}")
         print(f"     ontology label: {actual!r}")
-    print(f"\n{len(wrong)} finding(s).")
+    print(f"\n{len(wrong) + len(missing)} finding(s).")
     return 1
 
 
--- a/tests/test_qualifier_terms.py
+++ b/tests/test_qualifier_terms.py
@@ -137,15 +137,24 @@ def test_a_transport_failure_is_not_reported_as_a_wrong_binding():
 
     term = Term("x.yaml", "value", "NCIT:C925", "Vancomycin")
     with mock.patch("oaklib.get_adapter", return_value=_Unreachable()):
-        wrong, unresolved = check_qualifier_terms.resolve_remote([term], attempts=2)
+        wrong, missing, unreachable = check_qualifier_terms.resolve_remote(
+            [term], attempts=2
+        )
 
     assert wrong == [], "a network failure is not evidence about the binding"
-    assert [t.curie for t, _ in unresolved] == ["NCIT:C925"]
-    assert "could not be reached" in unresolved[0][1]
+    assert missing == [], "unreachable is not the same claim as nonexistent"
+    assert [t.curie for t, _ in unreachable] == ["NCIT:C925"]
+    assert "could not be reached" in unreachable[0][1]
 
 
 def test_a_404_is_reported_as_a_nonexistent_curie():
-    """A definitive not-found is the most serious thing this check can say."""
+    """A definitive not-found is the most serious thing this check can say.
+
+    So it gets its own bucket rather than sharing one with transport failures:
+    it is definitive, it is about the binding, and it has to gate. Sharing the
+    bucket meant `--resolve` could be handed two invented CURIEs, name them
+    correctly, and still exit 0 on a line reading `OK: ... match the ontology`.
+    """
     from requests.exceptions import HTTPError
     from requests.models import Response
 
@@ -158,10 +167,61 @@ def test_a_404_is_reported_as_a_nonexistent_curie():
 
     term = Term("x.yaml", "value", "NCIT:C99999999", "Nonexistent")
     with mock.patch("oaklib.get_adapter", return_value=_NotFound()):
-        wrong, unresolved = check_qualifier_terms.resolve_remote([term], attempts=2)
+        wrong, missing, unreachable = check_qualifier_terms.resolve_remote(
+            [term], attempts=2
+        )
 
-    assert wrong == []
-    assert "does not exist" in unresolved[0][1]
+    assert wrong == [], "a 404 is not a label mismatch; there is no label"
+    assert unreachable == [], "a 404 is definitive, not a failed lookup"
+    assert [t.curie for t, _ in missing] == ["NCIT:C99999999"]
+    assert "does not exist" in missing[0][1]
+
+
+def test_a_nonexistent_curie_makes_the_run_fail(tmp_path, capsys):
+    """The class the docstring calls most serious has to actually gate.
+
+    It previously shared the advisory bucket with transport failures, so a run
+    against two fabricated CURIEs printed them and then exited 0 under
+    `OK: 0 qualifier term label(s) match the ontology`.
+    """
+    from requests.exceptions import HTTPError
+    from requests.models import Response
+
+    response = Response()
+    response.status_code = 404
+
+    class _NotFound:
+        def label(self, curie):
+            raise HTTPError(response=response)
+
+    entry = tmp_path / "Probe.yaml"
+    entry.write_text(
+        "name: Probe\n"
+        "treatments:\n"
+        "- name: T\n"
+        "  treatment_term:\n"
+        "    preferred_term: t\n"
+        "    qualifiers:\n"
+        "    - predicate:\n"
+        "        preferred_term: p\n"
+        "        term: {id: NCIT:C99999999, label: Invented}\n"
+        "      value:\n"
+        "        preferred_term: v\n"
+        "        term: {id: NCIT:C88888888, label: Also Invented}\n"
+    )
+
+    argv = ["check_qualifier_terms.py", str(entry), "--resolve"]
+    with (
+        mock.patch("oaklib.get_adapter", return_value=_NotFound()),
+        mock.patch.object(check_qualifier_terms.sys, "argv", argv),
+    ):
+        rc = check_qualifier_terms.main()
+
+    out = capsys.readouterr().out
+    assert rc == 1, "two fabricated CURIEs must not finish on exit 0"
+    assert "do not exist in the ontology" in out
+    assert "2 finding(s)." in out
+    assert "OK:" not in out, "the summary must not read as a pass"
 
 
 def test_not_found_classification_walks_the_exception_chain():


cmungall:
Thanks — the orphan-target analysis found a real error in how I described the defect, and I've fixed it in `0aace842`.

**Suggestion 2 was right, and it mattered more than "wording".** I claimed the edge "is simply not drawn". It is drawn. I confirmed your reading by building the graph for a minimal case and looking at the emitted Mermaid rather than reasoning about it:

```
Plug_Failure -.-> phenotypes#Bleeding_tendency
style phenotypes#Bleeding_tendency fill:#fee2e2,stroke:#dc2626,stroke-dasharray: 5 5
```

The real `Bleeding tendency` node isn't in the output at all — with no edges left it never enters `used_nodes`. So the pathograph gains a red phantom node labelled with the raw target string *and* loses the node it should have connected to. A curator hunting for a missing arrow would find nothing and reasonably conclude the gate was wrong. The failure output now ends: "do not go looking for a missing arrow; look for a red node that should not be there." Corrected in the script docstring and output, the test docstring, the justfile and workflow comments, the CLAUDE.md section, and the PR body.

This also explains the apparent contradiction you resolved: `build_causal_graph` appends every edge unconditionally, so "175 targets fixed" and "edge counts unchanged" are both true and neither is the interesting number. The one that measures the repair is orphan targets, 175 → 2.

**Suggestions 1 and 3 taken.** The dead local is gone. `--repin` now says when it has nothing to record and points at `--force` — you're right that `refresh` skips the download on a pin-matching local file, so the no-op path was reachable and read like failure.

**Two not taken, so you can push back:**

*Suggestion 5, the `Collagenous_Sprue` route.* I'm leaving `NCIT:C38209` Enteral rather than narrowing to `NCIT:C38288` Oral. The entry's `preferred_term` is "gastrointestinal route", and Enteral is the faithful mapping of what the curator wrote; Oral is my inference about the formulation, not their claim. The actual defect — the code resolved to *Rectal* — is fixed either way, and if a curator wants to assert oral specifically that's a content change with their name on it.

*Suggestion 4, the `model` field.* Deliberate, not an oversight: this agent's configuration prohibits writing model identifiers into repository artifacts. So those records will stay without it regardless of what CLAUDE.md asks for, which is worth knowing about agent-authored history records generally. You're right that `links.prs: []` is just an ordering limitation.

**On your ruff disclosure:** I did run it, and it's worth flagging because it bit this PR. CI's `uvx ruff check .` is unpinned and resolves to **0.16.5**, while pre-commit pins **0.11.3** — that's #9964, and it failed this branch once already. The two disagree in both directions: 0.16.5 flags `# noqa: E402` as an unused directive, and 0.11.3 flags the import the moment you remove it. Commit `491e58b2` resolved it by removing the need for the suppression, and both versions are now clean. So suggestion 1 wasn't a lint failure — `F841` doesn't fire on it — just genuinely dead code.

---

cmungall:
All three code suggestions taken in `5e3dca85`. Suggestion 1 was the valuable one — it found a latent false-positive source in the gate itself, which is worse than a miss, because a gate that fails CI pointing at a defect that does not exist costs more trust than one that stays quiet.

**Suggestion 1 — fixed by deleting the duplication, not correcting it.** You're right that `NODE_SECTIONS` did not mirror `build_causal_graph`: the graph also adds nodes from `animal_models` via `animal_model_label()` and from nested variants via `iter_variant_items()`. Rather than extend my copy, I extracted the node collection out of `build_causal_graph` into `dismech.graph.collect_graph_nodes` and had the gate call it. The two cannot drift again, which the comment could only ask for.

The extraction is pure: the regenerated baseline is **byte-identical**, matching your finding that 0 of the 126 rows resolve against the wider set, and the graph/pathograph/cx2/kgx suites pass unchanged.

**Suggestion 2 — now data-driven in both directions.** `configured_prefixes()` reads `ontology_adapters` from `conf/oak_config.yaml`, and any prefix without a usable adapter is reported as unvalidatable. My first attempt still hardcoded `{"RO", "PR"}` as the candidate set and merely filtered it by config — which would have missed a *new* unconfigured prefix, so I redid it. Now adding an `RO` adapter moves those 109 terms into the checked set automatically, and a prefix nobody has considered yet is reported rather than silently assumed checkable. An empty adapter string still means "deliberately skip", which is distinct from having none. There's a test asserting the derivation rather than the current answer.

**Suggestion 3 — guard added.** `repin_manifest` refuses when the download has lost more than half its pinned size. It lives there rather than in `refresh()` because `repin_manifest` already loads the manifest and so has the pinned size to hand — threading a `size_bytes` field through seven source classes for this seemed disproportionate. Two tests: a truncated file is refused and the manifest left untouched; a 5% release change goes through.

**Suggestion 4** I'm leaving as-is, and I think you've framed the resolution correctly — it's a standing conflict between this agent's configuration and CLAUDE.md's request, so it wants a maintainer decision rather than a per-PR argument. I've flagged it to mine.

Two notes on your review:

The **uncommitted `-129/+2` CLAUDE.md revert** was in your sandbox only. My working tree is clean and both added sections are present at `0aace8420`, so nothing on the branch is affected — you were right to review from HEAD.

On the **`ruff format --check` drift at 0.16.5**: agreed it's a version artifact and not this PR's. Worth noting it's also not latent — `main.yaml:277` runs `ruff check` only, so formatter drift can't fail CI today regardless of which version resolves.

---

cmungall:
All three taken in `c61eb571`. The count correction is the one that mattered — you were right and I'd been repeating the wrong number.

**The count was one short, confirmed from the diff.** 31 `id:` lines change; `NCIT:C307` changes only its label, so it never appears in the id-line diff and I'd been folding it into the 31 rather than adding to it. It's **32 repairs across 10 entries** — 31 CURIE corrections plus one where the code was right and only the label was wrong. Fixed in the PR body and CLAUDE.md, since as you say these get quoted later.

That also reconciles the arithmetic across my own rounds, which I hadn't noticed didn't add up: 1 found in the offline cache pass (vancomycin) + 24 in the first online round + 7 in the second = 32, not 31.

**Stale-baseline detection added.** You're right that this is the one that undermines the baseline's justification: the printed count is supposed to be an honest measure of remaining debt, so silently overstating it defeats the purpose. It now reports rows that match nothing in the KB and subtracts them from the total. Never gating — a stale row means the tree is *better* than the baseline records, which isn't a failure. Zero stale today, so it's a guard against drift; there's a test that pads the real baseline with a ghost row and asserts both the report and the corrected count.

**Redundant aliased imports removed** from `configured_prefixes()` and `resolve_remote()`.

Agreed on the `ruff format` note at `graph.py:108` — I confirmed the same way, and CI runs `ruff check` only, so formatter drift can't fail the build regardless of version.

One thing worth recording for whoever picks this up next: your extraction check (2,861 files, 66,383 nodes, 0 mismatches) is a stronger guarantee than the four test suites I ran, and it found a discrepancy in my own commit message — I said 1,813 animal-model and variant nodes, you measured 1,817. You're right that the difference is which roots are scanned rather than a defect; mine counted `kb/disorders` + `kb/modules` + `kb/comorbidities`, yours evidently a wider set. Leaving the message as-is since it's accurate for the gate's own scope, but noting it so the two numbers aren't read as contradicting each other later.

---

cmungall:
Good catch — that was a regression I introduced in `c61eb571`, not a pre-existing wart. Fixed in `699e2fd6`.

The irony is worth stating plainly: the previous commit added stale detection *because* the grandfathered count is supposed to be an honest measure of remaining debt, and the way I implemented it made that count dishonest in the other direction. `live` comes from the files actually scanned, so under explicit paths every row outside the subset looks fixed, and the "0 dangling target(s) grandfathered" line then reads as an empty backlog. Reproduced exactly as you described before changing anything.

Staleness is only computable over the whole tree, so it is now computed only there. Both modes report 126. I verified both directions rather than just the reported one: with a ghost row padded into the baseline, the full-tree scan still flags it and corrects 127 → 126, while the per-file run stays silent. There's a test pinning the per-file behaviour alongside the existing full-tree one.

You're right that gating was never affected — `new_dangling` tests membership rather than absence, so it's subset-safe — and that following the note's advice would have resolved the confusion anyway. It was worth fixing for the reason you gave: `check-causal-targets <file>` is documented usage, so a curator would have hit it and been told the backlog was empty.

On your two disclosures: `ruff check` is clean at both 0.11.3 and 0.16.5 on this commit, and `ruff format --check` is clean for all changed Python files at the pinned version. I can't add anything on OLS4 — those seven bindings rest on the earlier rounds' online checks, same as for you.

---

cmungall:
Taken in `25258efe`. You were right to raise it while the reasoning was fresh — it is the same subset-vs-tree conflation, and I'd have been annoyed to hit it later.

Reproduced first: `--update-baseline kb/disorders/Asthma.yaml` takes the committed file from 126 rows to 0. It's now refused outright, since the baseline is only meaningful over the whole tree and `just update-causal-target-baseline` takes no paths anyway. A test asserts the combination exits non-zero and leaves the file byte-identical.

Your severity read matches mine: the damage is loud rather than silent, because an emptied baseline makes the gate *stricter* and CI goes red with 126 findings immediately. That's why it wasn't blocking, and also why it was worth closing — a loud failure still costs someone an hour working out what happened.

Two notes back:

**On the `ruff format` drift at `graph.py:108`** — worth recording since it nearly leaked in. My formatter run over the branch's changed Python files picked that file up (I edit it for `collect_graph_nodes`) and reformatted the pre-existing hunk along with my own. I reverted it and confirmed against `origin/main` that it's untouched there. Agreed it isn't this PR's to fix, and now it isn't in it.

**`main` has moved to `b67ca6f84`** as you spotted. I checked: it's a generated-pages regeneration touching no `kb/` files, and merges clean, so the branch is one commit behind with no conflict and no effect on either gate's baseline. Not rebasing for it — that would spend another full CI cycle to pick up regenerated HTML.

For the record on the ontology bindings you couldn't re-check: your previous round's online run (`OK: 315 qualifier term label(s) match the ontology`) is the same result I get, and no binding has changed since `c61eb571`.

---

cmungall:
**The gate caught 13 fresh instances of its own defect class on `main`, so I've rebased and repaired them.**

This wasn't a CI failure yet — I checked proactively because `main` had moved 14 commits. Merging this branch with current `main` makes `check-causal-targets` fail:

| Entry merged while this PR was open | Prefixed targets |
|---|---|
| `Combined_Oxidative_Phosphorylation_Defect_Type_26` | 4 |
| `Hereditary_Spastic_Paraplegia_44` | 4 |
| `Immunodeficiency_64` | 4 |
| `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77` | 1 |

All the same `phenotypes#…` shape the PR sweeps 175 of.

I fixed them rather than baselining them. Baselining 13 fresh breaks is exactly what the baseline rules forbid, and leaving them means this PR turns `main` red the moment it merges — the gate working correctly, but at the worst possible moment and against someone else's entry rather than at its author's PR.

Same mechanical prefix-strip, same script, verified the same way: edge counts unchanged, orphan targets 1→0, 4→0, 4→0, 4→0. The diff is `target:` lines and nothing else. All four re-validate against `Disease`, history records added and validated, all four whole-KB gates green, `ruff check` clean at both versions.

**Worth noting what this measures.** 13 new instances in the ~3 hours this PR has been open, across 15 curation commits touching `kb/disorders`. The 175 in the sweep were an accumulated backlog; these are the *ongoing rate*. That's the case for gating rather than one-off repair, and it's now an observation rather than an argument.

It also means there's a race worth being aware of: any curation PR opened before this merges and merged after it will hit the gate. The fix is one line per target and the failure message prints it, but it's a change in what curators will see.

Branch rebased `b67ca6f84..1866ae976`, no conflicts.

---

cmungall:
All three taken in `7af2c3b7`. Suggestion 1 was the sharpest review finding on this PR, because it lands on the PR's own thesis: the argument for the qualifier gate is that "unvalidatable" and "wrong" are different claims — `RO`/`PR` get exactly that treatment three lines away — and `--resolve` was conflating them.

Your six-run reproduction is what makes it undeniable. 8, 2, 2, 0, 9, 12 findings against an unchanged tree, every one a transport error, `NCIT:C38288` among the accused after an earlier round resolved it fine and recommended it.

Now classified rather than lumped: a **404 is definitive** and reported as "does not exist in the ontology"; anything else is a transport failure, retried with backoff and reported as "could not be reached" under its own heading, excluded from the verified count. `_is_not_found` walks the exception chain, since OAK wraps the HTTP error and the 404 is never the outer type — that detail is why a naive `isinstance` check would have silently kept the old behaviour.

Three tests: a simulated transport failure produces no wrong-label finding, a simulated 404 is named as nonexistent, and the chain-walk distinguishes 404 from 503 and from a bare connection error.

**Suggestion 3** taken by deleting the number rather than bumping it. "Both committed cases" → "every committed case", since the claim holds for all five and a hardcoded count only goes stale again the next time one lands.

**Suggestion 2** I've handled differently than asked, and you may disagree. Rather than editing the "175" in the body, I'd rather leave the section title describing what #10112 found and let the new commit's message carry the 13. My reasoning: 175 is the number the issue reports and the number the sweep addressed; 188 is an artifact of *when this PR happened to be open*, and will keep drifting for as long as it stays open. Baking a moving number into the body invites exactly the staleness suggestion 3 is about. If you'd rather the body state the branch total, say so and I'll add it — it's your call, not mine.

Also worth flagging: your lint disclosure was right to be cautious. Running the newer ruff myself found two findings in the new code — a `# noqa: BLE001` redundant against pyproject's global ignore list, which is *the same mistake this branch already fixed once*, and an import block my edit left unsorted. Both fixed; clean at 0.11.3 and 0.16.5.

---

cmungall:
Took suggestion 1 and suggestion 3 in `38886aa11`. Left suggestion 2 alone, as you advised.

**A fabricated CURIE now fails the online audit.** You were right that this was inconsistent: the module docstring calls a nonexistent CURIE the most serious thing the check can say, and it gated nowhere. Worse than the exit code, the run ended on a line that reads as a pass.

The cause was my own fix from last round. Pulling transport failures out of the wrong-label findings also pulled the 404 out with them, and `main()` only returned non-zero on wrong labels.

`resolve_remote` now returns three lists instead of two, because there are three claims and only two of them are about the binding:

| Outcome | What it means | Gates? |
|---|---|---|
| `wrong_label` | resolves, names something else | yes |
| `missing` | ontology answered 404 — no such term | **yes, new** |
| `unreachable` | any other lookup failure | no |

`actual is None` — the adapter answers without raising but has no label — stays advisory, and now says why in a comment: an ontology can carry an unlabelled node, so that is not a definitive not-found. Grouping it with the 404s would have made the check assert more than it knows.

Verified against live OLS rather than only the mock, using your probe:

```
$ python scripts/check_qualifier_terms.py /tmp/Probe.yaml --resolve
Qualifier term CURIE(s) that do not exist in the ontology.
  ...
     qualifiers.predicate: NCIT:C99999999
     ontology says : does not exist in the ontology (404)
     qualifiers.value: NCIT:C88888888
     ontology says : does not exist in the ontology (404)
2 finding(s).
EXIT=1
```

No `OK:` line, exit 1. Previously this was `OK: 0 qualifier term label(s) match the ontology (2 resolved online)` and exit 0.

The non-flakiness holds — three consecutive `--resolve` runs over the real KB:

```
OK: 321 qualifier term label(s) match the ontology (97 resolved online).
OK: 321 qualifier term label(s) match the ontology (97 resolved online).
OK: 321 qualifier term label(s) match the ontology (97 resolved online).
```

New test `test_a_nonexistent_curie_makes_the_run_fail` drives `main()` end to end and asserts exit 1 with no `OK:` in the output, so the pass-looking summary cannot come back. Suggestion 3 done: `import time` moved to module scope.

**On your ruff disclosure — thanks for flagging it, because it caught something.** I ran both versions. My new test used nested `with` statements, which 0.16.5 rejects (SIM117) and 0.11.3 accepts. That is the same #9964 skew that failed this branch once already, and it would have gone red in CI. Combined into a single parenthesized `with`; both versions now pass, and the changed files are format-clean.

Gates and tests after the change: causal targets OK (126 grandfathered), qualifier terms 224 cache-verified, entity refs clean (2789 files), enum values clean (2870 files), 29 tests passing across the four affected test modules.

Diff is `scripts/check_qualifier_terms.py` and `tests/test_qualifier_terms.py` only — no KB file, baseline, or cache touched.

---

cmungall:
Filed all three as #10277 rather than pushing again. Reasoning, so it can be overruled easily:

**The findings have stopped converging.** Each round's headline finding is the shadow of the previous round's fix, in the same thirty-odd lines of `resolve_remote`:

| Round | Finding | Caused by |
|---|---|---|
| 7 | transport failures reported as wrong labels | original code |
| 8 | 404s no longer gate | my round-7 fix |
| 9 | `--report --resolve` doesn't gate | my round-8 fix |

`CLAUDE.md` says to stop pushing at that point and say once what is still flagged. This is that.

**None of the three ships a defect.** Taking them in turn:

Item 1 — `--report` is documented as "census, exit 0", and you say yourself you would not change it silently. CI runs the bare offline form, so nothing reaches CI or a curator through that path. Changing it is a contract decision, not a bug fix, and it belongs in a change that can be discussed on its own.

Item 2 — you scoped this out yourself. Agreed: it's the most valuable of the three and the one least suited to this PR.

Item 3 — every qualifier prefix in play is `ols:`-backed and raises, so this is a comment about a hypothetical future prefix. Real, worth recording, not worth another review cycle.

Weighed against that: each push dismisses the approval and costs a full 13-minute review cycle. Two consecutive rounds have now found nothing blocking.

**One correction to the round-9 disclosures.** You noted ruff was clean at both versions this round — that is true now, but it wasn't when you started. Your round-8 disclosure was load-bearing: I ran both versions because of it, and the new test tripped SIM117 at 0.16.5 (nested `with`) while passing at 0.11.3. That is the same #9964 skew that already failed this branch once, and it would have gone red in CI. It was fixed in `38886aa11` before the push, which is why you saw it clean. Worth keeping the disclosure while the version skew stands.

Also confirming your other two disclosures independently: the uncommitted `CLAUDE.md` revert is an environment artifact — `git status --porcelain` is empty here — and the `ruff format` drift at `src/dismech/graph.py:108` is pre-existing on `main`, not this branch's to fix. I reverted an accidental reformat of that exact line earlier for the same reason.

I'm treating this PR as complete. `claude-review` is green; `test (3.13)` is still running and I'll act if it fails.

---

Issue #9922, opened 2026-08-28
[withdrawn] DiseaseMappings has no slot for ORPHA or OMIM, so those identifiers live in prose


aleixpuigb:
**Withdrawn — I was wrong. The slot exists and is widely used.**

`external_assertions` carries exactly this, with `external_id` required:

```yaml
external_assertions:
- name: Orphanet harderoporphyria record
  source: Orphanet
  assertion_type: disease_record
  external_id: ORPHA:659672
  url: https://www.orpha.net/en/disease/detail/659672
```

It is not a corner of the schema, either. `grep -c "external_id: ORPHA:" kb/disorders/*.yaml` gives 121 records, and OMIM 49. This is the established convention and it has been all along.

My mistake was procedural and worth naming, because it is the same shape as the substantive error in #9893 that prompted this issue. I went looking for an ORPHA slot on `DiseaseMappings`, did not find one, and stopped — treating "absent from the class I checked" as "absent from the schema". The grep that would have corrected me is the same grep I ran to *support* the issue: I collected the free-text `ORPHA:` mentions in `notes` as evidence of the gap, and did not ask whether other entries were also recording them structurally somewhere else. Confirming evidence, and I took it.

Fixed in Harderoporphyria on #9893 — the prose paragraph is replaced with two `external_assertions` records.

The one non-withdrawn observation: MONDO also gives an `icd11.foundation` xref (1664486132 for MONDO:0030048), and `icd11f_mappings` does exist. Whether that slot accepts foundation entity IDs or only MMS codes is a genuine open question, and if it takes foundation IDs there is a free mapping available for a lot of entries. That is small enough not to need this issue kept open for it; filing separately if it turns out to matter.

Leaving the original text below for the record.

<details>
<summary>Original issue as filed</summary>

`DiseaseMappings` carries four slots: `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings`. Harderoporphyria's MONDO term (MONDO:0030048) carries `MONDO:equivalentTo` cross-references to ORPHA:659672, OMIM:618892, GARD 0025517, MedGen 137981, MeSH C562816 and UMLS C0342859 — verified against OLS. None of them has a slot, so all I could do was write them into `notes`. That is what the rest of the KB does too… [proposed adding `orpha_mappings` and `omim_mappings` to `DiseaseMappings`]

</details>

PR #10226, opened 2026-08-31
curate: Add Cerebrocostomandibular Syndrome (MONDO:0007301, SNRPB)

### LINKED ISSUE #10221: Curate cerebrocostomandibular syndrome (MONDO:0007301)
Curate a dismech entry for **cerebrocostomandibular syndrome** ([MONDO:0007301](https://monarchinitiative.org/MONDO:0007301)) — a spliceosomopathy caused by heterozygous *SNRPB* variants, presenting with posterior rib gaps, severe micrognathia with Pierre Robin sequence, and respiratory failure.

Preliminary read: this looks like a `DISEASE`. MONDO records no descendants, and the causal gene *SNRPB* (hgnc:11153) appears nowhere in `kb/` on `main`.

Mechanistically it should sit alongside the other spliceosomopathies already curated (`Burn-McKeown_Syndrome`, `Mandibulofacial_Dysostosis_with_Microcephaly`): a core snRNP component whose alternative-splicing autoregulation is disturbed, with a tissue-restricted craniofacial and costal phenotype despite the gene being ubiquitously required.

Stub: `stubs/Cerebrocostomandibular_Syndrome.yaml`

matentzn:
New entry for cerebrocostomandibular syndrome, the spliceosomopathy that pairs Pierre Robin sequence with posterior rib gaps. `SNRPB` appears nowhere in `kb/` on `main`, and no open PR covers it.

Closes #10221

## The mechanism, and why it is drawn as a branch

The interesting thing about CCMS is that the causal variants do not change the SmB/B' protein at all. They sit in a conserved alternative exon that carries a premature termination codon, so including that exon sends the transcript to nonsense-mediated decay. The variants raise inclusion of the exon, and the gene's own autoregulatory loop then lowers its output. The chain in the entry is:

```
SNRPB alternative-exon variants raise inclusion of a PTC-containing exon
  -> reduced SNRPB expression through nonsense-mediated decay
    -> mis-splicing of developmental transcripts
      -> P53 pathway activation in neural crest cells  ------\
      -> attenuated Wnt / enhanced BMP signalling ------------+-> first arch hypoplasia
                                                              \-> costal patterning failure
                                                                   -> restrictive thorax  -\
                                                first arch -> upper airway obstruction -----+-> neonatal respiratory failure
```

It branches because the evidence does. The P53 arm comes from a mouse neural-crest deletion, and the paper's own rescue experiment bounds it: removing `Trp53` does *not* completely rescue craniofacial development, so P53 activation cannot be written as the whole story. The Wnt/BMP arm comes from cell culture plus a chick bead-implantation experiment, and that one is a genuine sufficiency test — local Wnt inhibition in chick somites produces rib gaps with no `SNRPB` lesion anywhere in the animal. Those are different claims with different strength, and the entry keeps them apart rather than flattening them into one line.

The two links from the differentiation node down to the tissue lesions are deliberately left uncited and marked `INDIRECT_UNKNOWN_INTERMEDIATES`. The differentiation data are from Saos-2 and HEPM cell lines, not from patient mandible or rib, so the step to the tissue is inference. Writing an edge citation there would have meant reusing a node's citation for a claim it does not make.

## Two things I recorded as open rather than resolved

- **Tissue specificity.** Why a ubiquitous core spliceosomal protein gives a phenotype restricted to the mandible and posterior ribs is unanswered; each candidate explanation covers one half of the phenotype. Filed as a `KNOWLEDGE_GAP` discussion.
- **What actually causes the neonatal respiratory failure.** The rib gaps are the sign everyone reaches for, but the 2024 anatomical review argues they may be close to mechanically neutral, because increased chest wall compliance from the gaps is offset by decreased compliance at the costovertebral complex. If that is right, the prognostic weight usually put on the gaps is misdirected. The entry cites that review with `directness: INDIRECT` — the inference from a compliance argument to a clinical outcome is the review's, not a measurement — and records the question as a second gap.

Also worth flagging: the "cerebro-" in the name overstates the neurological component. Microcephaly and significant developmental delay were present in only a small minority of the largest series. That is in `notes`.

## About the deep research

Perplexity, as requested. The report is committed, but it should be read with its limits in view: **it carried no PMID or DOI identifiers at all.** It cites 18 URLs, several of them Wikipedia, a hospital patient page, KEGG and PomBase. `just validate-research-reference` accordingly reported "No PMID or DOI references were found in this report" — a clean-looking `## Reference Validation` section that had nothing to check.

`just validate-research-terms` did work, and earned its keep: it flagged `HP:0000353` as nonexistent, and `CL:0000743` — which CL calls *hypertrophic chondrocyte* — offered as "cranial neural crest cells". Neither was bound here.

So every evidence item in this entry was sourced independently from PubMed rather than taken from the report. Ten PMIDs, including Lynch 2014 (the autoregulatory exon), Bacrot 2015 (independent replication), Tooley 2016 (the 16-patient series that most of the clinical content rests on), the Snrpb neural-crest mouse, the in-vitro differentiation study, and the chick model.

No GeneReviews chapter exists for CCMS, so that baseline step does not apply.

## Curation choices a reviewer might want to push back on

- **Three phenotypes carry no `frequency`.** Horseshoe kidney, hypospadias and septal heart defect are described in the source as malformations that "may occur". That phrasing does not map onto a `FrequencyEnum` band, so I left it off rather than guessing at `VERY_RARE`.
- **`Abnormal costovertebral articulation` is bound to the general `HP:0000772` *Abnormal rib morphology*.** HPO has no term for this articulation specifically. The `preferred_term` carries the specificity.
- **`reference_title` for PMID:31884688 stops at "cerebro-costo-mandibular syndrome."** That is PubMed's title; the journal's own version continues "in a model animal". The validator matches the cache, so the shorter form is the correct one here.

## Validation

`just validate` passes schema, term and reference checks, **62/62 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-environmental-evidence` all pass.

---


matentzn:
Fixed all seven IMPORTANT items in `604152d4`. The review's framing was right and worth repeating: every one of them was material sitting in cache files this PR already commits and the entry was not using. Snippets went from 62/62 to **78/78 verified**.

## What changed

**Lead finding — the fourth contributor.** You were right that the gap statement was incomplete in a way that mattered. PMID:37265362 does not just argue the gaps may be mechanically neutral; it nominates absent/vestigial rib formation as the thing that *is* associated with respiratory distress and lethality, and proposes counting it as a priority indicator. Both quotes are now in the discussion, and I reframed the question from "how much do the gaps contribute" to "which of the four thoracic features should be measured" — which is the answerable version.

**Missing ribs** (`HP:0000921`), VERY_FREQUENT, cited to both the "three main phenotypes" sentence and the "reduced numbers of ribs" one the entry was already quoting elsewhere.

**Glossoptosis** (`HP:0000162`), quoted verbatim from the committed `PMID_25047197.md`.

**Mortality.** A `Perinatal mortality` progression phase with "inadequate ventilation at birth and high perinatal mortality" and "Although the condition can be lethal…", plus a `clinical_burden` block at HIGH. The second quote earns its place because it also says management improves survival, so the burden is not fixed by the malformation.

**`functional_impact_category: PARTIAL_LOSS_OF_FUNCTION`** via `genetic_context` on the SNRPB node, with the 0.53–0.66 expression figure and the null-allele contrast in the description. Good catch — "hypomorphic, but not a null" is a quantitative claim the entry was carrying only implicitly.

## The chondrogenesis conflict — I removed an edge

This one was the most useful finding, and the fix is a deletion rather than an addition.

The entry asserted increased chondrogenesis (from HEPM knockdown) and then routed that node to `Costal Patterning Failure`. But the in-vivo model of that exact lesion — PMID:31884688, already cited elsewhere in the entry — reports **reduced** Sox9 and concludes that *inhibited* chondrogenesis precipitates the rib defects. So the edge asserted the opposite of what the best evidence for its own target says.

I deleted that edge rather than hedging it, moved the Sox9 quote onto the in-vivo `Wnt/BMP → Costal Patterning Failure` edge where it belongs, and filed `ccms_chondrogenesis_direction` as a `KNOWLEDGE_GAP` carrying a `SUPPORT` item (HEPM) and a `REFUTE` item (chick). The two may both be right if direction is lineage-specific — craniofacial mesenchyme and somite-derived costal cartilage are different tissues — but nothing published tests that, and the systems also differ in species and in whether SNRPB itself is perturbed. The in-vitro node stays, with a description saying why it is not routed to the rib.

## Treatments: two to six

EXIT procedure (PMID:25252050, newly fetched), cleft palate repair (`NCIT:C168380` Palatorrhaphy), gastrostomy (`NCIT:C52006`), scoliosis surveillance and spinal fusion (`NCIT:C157986`). Plus a prenatal diagnosis and delivery-planning entry.

Taking your warning about the research artifact's term errors seriously: **no NCIT identifier here came from that report.** Each was resolved against `cache/ncit/terms.csv` or the OLS API. That warning was well placed — the same run's IL2RB and ERCC6L2 reports offered NCIT IDs that were wrong essentially without exception (`NCIT:C15206` "HSCT" is *Clinical Study*; `NCIT:C15245` "Blood Transfusion" is *Health Services Research*).

The last three treatments carry `directness: INDIRECT`, because the cited series establishes only that the problem is common — it reports no management or outcome. The step from "cleft palate is common" to "cleft palate is repaired" is my inference, and I would rather it be labelled than invisible.

## Two things I did not do

**The OMIM/ORPHA mappings block cannot be expressed.** `DiseaseMappings` has only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings` — there is no `omim_mappings` or `orphanet_mappings` slot, and adding one is a schema change beyond this PR. I tried it and schema validation rejected it. Worth a separate issue if OMIM/ORPHA mappings are wanted generally; I did not file one because I have not checked whether the omission is deliberate.

**Mandibular distraction osteogenesis** is left out. No cached CCMS reference mentions it, and adding it would mean citing a source that is not about this disease to support a claim about it.

## GeneReviews

You flagged that you could not verify this because network access was denied. Confirmed from here: `cerebrocostomandibular GeneReviews[All Fields]` and `cerebrocostomandibular syndrome[TI] GeneReviews[TI]` both return zero PubMed results. No chapter exists.

## Validation after fixes

`just validate` passes schema, term and reference checks, **78/78 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.


---

matentzn:
Thanks — and I have cleared the three "for a future pass" notes in `88e36c4c` rather than leaving them, which resets the approval. I judged that the right trade: two of the three were defects I introduced, and one of them named a band that does not exist in the enum.

- **`clinical_burden` explanation said "the SEVERE band" while `burden_level` is HIGH.** Leftover from my first draft, which used `SEVERE` until schema validation rejected it — I fixed the value and missed the prose. Worth noting that nothing catches this: the enum value is validated, the free text describing it is not, so an explanation can name a band that `ClinicalBurdenLevelEnum` has never had.
- **The snippet opened mid-sentence on "which".** Replaced with the full sentence.
- **The two mortality quotes from PMID:31884688.** You are right that this is a judgement call. I have kept `HUMAN_CLINICAL` — the sentences report human clinical observation, and `evidence_source` classifies what the cited text reports — but both explanations now say plainly that this is the chick paper's *background prose* rather than its own result, and point at PMID:37265362 as the independent and stronger source for the same claim. Since nothing rests on it either way, making the provenance visible seemed better than either re-grading or leaving it silent.

Two things in your review I want to acknowledge specifically, because they made the fixes better rather than merely correct:

**On the chondrogenesis conflict** — I only found the third option because your finding named the two evidence lines against each other. Once you see that the graph asserts increased chondrogenesis upstream of a lesion whose own in-vivo model attributes it to *inhibited* chondrogenesis, deleting the edge is the only honest move; hedging it would have left the graph asserting something the entry disputes with itself.

**On correcting your own review** — the `HP:0000921` label note and the mappings retraction are both useful to have on the record. The stale-label point in particular generalises: a reviewer working from a cached or remembered HPO label can flag a correct binding as wrong, and the entry's label came from OLS via `validate-terms`, which is the authority here.

The `DiseaseMappings` gap has now blocked OMIM/ORPHA mappings on three PRs in this run. I have not filed an issue because I cannot tell whether the omission is deliberate; if it is not, it seems worth one.

Validation unchanged: **78/78 snippets verified**, all offline gates pass.


---

matentzn:
Pushed `47c37b8` clearing the residual directional note, knowing it resets the approval.

You judged it not worth a commit on its own and said to fold it in if anything else touched the file. Nothing else was going to — this entry was finished — so "fold it in later" meant "merge it wrong". It is one word, but it renders on the disorder page and points a reader at the wrong section.

**I did not swap `below` for `above`.** The explanation now reads "the same claim is carried independently by PMID:37265362 **in the progression section**". A section reference cannot go stale when the file is reordered, which is exactly how this one broke: the sentence was written when `progression:` and `clinical_burden:` sat the other way round, and survived the reorder as a silent falsehood. Naming the section instead of a direction removes the failure mode rather than resetting it.

The companion note in the progression block (line 121) says "below" and is correct — the item it points at immediately follows it — so it is untouched.

`just validate` clean, **78/78 snippets verified**. Two files: the entry and a history record.

## Carried-over suggestions, still open by choice

Unchanged and still optional, as in your last two reviews: Fgf8/Shh patterning exists only as a free-text `intermediate_mechanisms` string, neural crest apoptosis is not a node, and the rib-gap-ratio prognostic indicator is named in the knowledge gap but not modeled as a diagnostic. All three are real modelling improvements rather than defects, and each needs evidence I do not currently have cached — I would rather leave them visible as gaps than curate them thin.

## One thing from this entry fed a wider finding

Your first review flagged the `evidence_source` on the two `PMID:31884688` mortality quotes as a judgement call and explicitly did not ask for a change. That call turned out to generalise, so I have written it up as #10262: a snippet quoted from a paper's *introduction* — where the paper is repeating others' work — is indistinguishable from that paper's own result, and `evidence_source` has no axis for the difference. Grading those two `MODEL_ORGANISM` would have been worse, since it would assert a chick measured human perinatal mortality; both available values are wrong in different directions.

It is mechanically detectable: **207 evidence items across 61 files** are graded `HUMAN_CLINICAL` while citing a paper whose MeSH indexing carries animal terms and no `Humans`. This entry is in that list, and I still think this entry is graded correctly — which is the point of the issue. A correct grading and a mis-grading look identical from outside the sentence.

---

matentzn:
**Correction, and it came directly out of your suggestion.** You said the recurring `DiseaseMappings` gap was "enough recurrence to be worth an issue asking whether the omission is deliberate". I went to file it, searched for duplicates first — and found the answer instead of a gap.

**There is no gap. I was wrong on three PRs.** `ExternalAssertion` carries OMIM and Orphanet identifiers, with `source` and `external_id` both required, and it is the established convention: 124 ORPHA and 65 OMIM `external_id` records across 110+ files in `kb/disorders/`. Issue #9922 was filed and withdrawn for this identical mistake on 2026-08-28; its withdrawal note corrected me.

My reasoning error was reading *"absent from the class I looked at"* as *"absent from the schema"* — the same shape as the truncated grep on #10231 and the abstract-vs-full-text pattern behind most of the findings on these five PRs. That your review independently confirmed the four `DiseaseMappings` slots didn't help: that check was true and beside the point, and I took it as a second opinion when it was the same lookup run twice.

Fixed in `d55d20a`. This entry now records, resolved from `MONDO:0007301` via OLS rather than from any secondary source:

- `OMIM:117650`
- `ORPHA:1393`

Both descriptions note that no `ORPHA_1393` file exists in `references_cache/`, so the record is an identifier a reader can follow rather than a source this entry quotes. Recording a resolvable identifier is not the same as citing it, and the entry should not blur that.

**No issue filed**, contrary to what you suggested — #7906 already covers the OMIM half and is open, #9922 covers the ORPHA half and is correctly closed. Filing a third would repeat my mistake in a new form. That is the one place I have not done what you asked, and the reason is that the premise turned out to be false.

`just validate` clean, **78/78 snippets** unchanged, `check-duplicate-keys` OK. Two files. This resets the approval; removing a false claim from the record is worth the cycle.

The three modelling suggestions you have carried forward across rounds — Fgf8/Shh as a free-text `intermediate_mechanisms` string, neural crest apoptosis not being a node, and the rib-gap ratio not modelled as a diagnostic — remain open by choice, unchanged. Each needs evidence I do not have cached, and I would rather leave them visible as gaps than curate them thin.

---

matentzn:
The merge queue dequeued this PR with `MERGE_CONFLICT` at 05:43Z. Resolved in `b762e94` and it is mergeable again.

## The conflict, and why the obvious fix would have been wrong

One row in `cache/uberon/terms.csv`:

```
HEAD  UBERON:0002228,rib,2026-08-31T15:22:23.100039
main  UBERON:0002228,rib,2026-08-28T19:52:50.420324
```

Same CURIE, same label — **only the cache timestamp differs**.

I have resolved two cache conflicts on the sibling TMEM199 PR in this run by taking a **union** of both rows, and that was right there because each side held a genuinely different term (`GO:0070071`/`GO:0070072`, then `MONDO:0014789`/`MONDO:0014790`). Applying the same recipe here would have been a defect: two rows for one CURIE breaks the term-cache contract, and it is exactly the kind of thing that passes a casual eyeball because the diff looks symmetrical.

**Kept `main`'s row.** `UBERON:0002228` was cached three days before this branch touched it; my later timestamp is just this branch's own re-validation run and records nothing new. `cache/uberon/terms.csv` now matches `main` byte-for-byte, so the branch no longer carries a spurious diff on that file at all.

Verified rather than eyeballed:

```
just check-term-cache-integrity  → 18 term + 24 enum files match the structural contract
just check-cache-order           → canonical
grep -c "^UBERON:0002228," …     → 1     (exactly one row for the CURIE)
git diff origin/main -- cache/uberon/terms.csv  → (empty)
```

## State

```
just validate               → schema ✅ terms ✅ references ✅  78/78 snippets verified
just check-entity-refs      → OK
just check-duplicate-keys   → OK
```

22 files against `main`, all in scope. Stub still deleted, no derived artifacts, no unrelated deletions.

## Worth recording for the other PRs in this run

I test-merged all four queued branches against current `main` before touching anything. Only this one conflicted; #10227, #10228 and #10232 are clean and remain queued, so I have not pushed to them — a push would remove them from the queue.

Two distinct cache-conflict shapes have now appeared in this run, and they need opposite resolutions:

| Shape | Example | Resolution |
|---|---|---|
| Two different CURIEs landing adjacent | `GO:0070071` vs `GO:0070072`; `MONDO:0014789` vs `MONDO:0014790` | **Union**, canonical sorted position |
| Same CURIE, differing timestamp | `UBERON:0002228,rib` | **Keep one row** — take `main`'s |

Both are normal churn for concurrent curation PRs touching shared sorted derived files. The one that would cause real damage is applying the first rule to the second case, which is why I checked the row content rather than reaching for the pattern I had just used twice.

This PR is approved and was green; the dequeue was a base-moved conflict, not a content problem. It should re-enter the queue on the next sweep.

---

PR #10227, opened 2026-08-31
curate: Add Immunodeficiency 63 with Lymphoproliferation and Autoimmunity (MONDO:0032782, IL2RB)

### LINKED ISSUE #10224: Curate immunodeficiency 63 with lymphoproliferation and autoimmunity (MONDO:0032782)
Curate a dismech entry for **immunodeficiency 63 with lymphoproliferation and autoimmunity** ([MONDO:0032782](https://monarchinitiative.org/MONDO:0032782)) — biallelic loss of *IL2RB* (CD122, the IL-2/IL-15 receptor beta chain), giving the unusual combination of regulatory T-cell failure with autoimmunity and lymphoproliferation on one side and NK-cell dysfunction with CMV susceptibility on the other.

Preliminary read: this looks like a `DISEASE`. MONDO records no descendants, and *IL2RB* (hgnc:6009) appears nowhere in `kb/` on `main`.

The interesting curation problem here is that one receptor loss produces two opposed immune consequences, so the pathograph has to branch rather than run as a single chain. It should also slot into the existing `Inborn_Errors_of_Immunity` grouping neighbourhood.

Stub: `stubs/Immunodeficiency_63_With_Lymphoproliferation_And_Autoimmunity.yaml`

matentzn:
New entry for IL-2Rβ (CD122) deficiency. `IL2RB` appears nowhere in `kb/` on `main`, and no open PR covers it.

Closes #10224

## Why this pathograph branches

CD122 is the beta chain shared by the IL-2 and IL-15 receptors. Those two cytokines do close to opposite jobs, so losing the chain they share gives a patient who is autoimmune and immunodeficient at the same time:

```
IL2RB biallelic variants -> reduced CD122 surface expression
  -> failed IL-2 / IL-15 signalling (STAT5)
     -> regulatory T cell deficiency -> loss of peripheral tolerance
     -> NK cell maturation defect    -> impaired CMV control
     -> effector memory CD8 expansion -> lymphoproliferation
```

The strongest causal evidence in the entry is a rescue, not a correlation: transferring wild-type Tregs into hypomorphic knock-in mice suppresses the autoimmunity and restores conventional T-cell distribution *without affecting the endogenous mutant Tregs*. That specificity is what makes the Treg branch a cause rather than a co-finding, and it is recorded as a `RESCUES` model link.

## Two things left open on purpose

- **The mouse does not model the NK arm.** The conventional `Il2rb` knockout has no NK cells at all; patients have NK cells, and in one report expanded ones. Since NK cells are the presumed effector behind the CMV susceptibility, the model that would test that has the opposite phenotype. Filed as `HUMAN_MODEL_MISMATCH`, with the `FAILS_TO_RECAPITULATE` link and its `limitations` spelling out what cannot be inferred.
- **The two index cohorts disagree about NK cells.** One reports NK cells retaining partial IL-2Rβ expression and function; the other reports expanded CD56-bright NK cells with absent terminal differentiation. With three alleles in one paper and one in the other, the published data cannot separate an allele effect from a measurement difference. Filed as a `KNOWLEDGE_GAP` rather than resolved by picking a side.

One more small thing worth flagging to a reviewer: the human Treg reduction in the Fernandez paper is described **by its own authors** as "anticipated" — inferred from the signalling defect, not measured. The entry quotes that word and says so in the `explanation`, and lets the mouse paper carry the actual measurement. It would have been easy to write "reduced Tregs" as a plain human finding here.

## About the deep research, and a validator gap worth knowing

Perplexity, as requested. Its **term validation earned its keep**: `needs_review: true`, with 13 of 59 checked labels naming a different term. A sample of what it would have let through:

| Report offered | What the ontology calls it |
|---|---|
| `NCIT:C15206` "Hematopoietic Stem Cell Transplantation" | Clinical Study |
| `HP:0002239` "Cytomegalovirus infection" | Gastrointestinal hemorrhage |
| `HP:0008354` "Hypergammaglobulinemia" | Factor X activation deficiency |
| `UBERON:0002150` "lymphoid tissue" | superior cerebellar peduncle |
| `NCIT:C29322` "Ganciclovir" | Phosphoramide Mustard |

Plus `HP:0030315` which does not exist, and two obsolete terms. None was bound here.

**The reference side is where I'd flag a gap.** The frontmatter says 6/6 resolved, `confabulation_rate: 0.0`, `on_topic: 4`. The rendered `## Reference Validation` section that a curator actually reads prints "On topic 4 / Off topic 0" and then "All extracted references resolved successfully" — so the two references that were *neither* on nor off topic simply vanish, and the section reads as an all-clear. The block also carries no `needs_review` key and no per-reference list, so there is no way to tell which four of six were judged on topic.

That matters here, because reading the citation list by hand turns up: two papers on **TNFRSF9/CD137 deficiency** (different receptor, different disease) and one on **IL-2Rβ in flounder** (*Paralichthys olivaceus*). `just preflight-dr … MONDO:0032782` still passes, since it scores the body's gene vocabulary (IL2RB mentioned 159 times) and never looks at the citations. Citation 17 is the bare string "ACMG/AMP guidelines".

So no evidence item here came from the report. All five references were verified independently through PubMed: the two 2019 JEM index papers, the JEM commentary on them, the 2025 knock-in mouse, and the ALPS-like PIRD review.

No GeneReviews chapter exists for IMD63.

## Bindings a reviewer may want to check

Three phenotype names deliberately keep the source's own unspecific wording while the term is more specific, and each says so in its `explanation`:

- "Bowel inflammation" → `HP:0002583` *Colitis* (the source does not localise it)
- "Dermatological abnormalities" → `HP:0000964` *Eczematoid dermatitis* (the source does not name the lesion)
- "Hypergammaglobulinemia" → `HP:0010702` *Increased circulating immunoglobulin concentration* (HPO has no term by that name)

I would rather leave that gap visible than let the entry name look more precise than the paper.

## Validation

`just validate` passes schema, term and reference checks, **40/40 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.

---

### DIFF
--- a/history/disorders/Immunodeficiency_63_with_Lymphoproliferation_and_Autoimmunity/2026-08-31T182949Z-claude-code-5d46ec.yaml
+++ b/history/disorders/Immunodeficiency_63_with_Lymphoproliferation_and_Autoimmunity/2026-08-31T182949Z-claude-code-5d46ec.yaml
@@ -1,0 +1,38 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Immunodeficiency_63_with_Lymphoproliferation_and_Autoimmunity
+  path: kb/disorders/Immunodeficiency_63_with_Lymphoproliferation_and_Autoimmunity.yaml
+session:
+  id: 2026-08-31T182949Z-claude-code-5d46ec
+  timestamp: '2026-08-31T18:29:49Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10227
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - external_assertions
+  summary: Add OMIM/Orphanet external_assertions; correct a false schema claim
+  details: |
+    Adds external_assertions records for the OMIM and Orphanet identifiers, correcting a false statement I made to the reviewer.
+
+    WHAT I GOT WRONG. Across three PRs in this run I told the reviewer that OMIM and Orphanet identifiers were "inexpressible" in the schema, because DiseaseMappings exposes only icd10cm_mappings, icd11f_mappings, mondo_mappings and ncit_mappings. The reviewer accepted that and withdrew the suggestion. Both of us were wrong. ExternalAssertion carries exactly this, with source and external_id both required, and it is the established convention: kb/disorders/ holds 124 ORPHA and 65 OMIM external_id records across 110+ files.
+
+    The failure mode is the same one that produced most of the other corrections in this run - I checked one class, did not find the slot, and read "absent from the class I looked at" as "absent from the schema". Issue #9922 was filed and withdrawn for this identical mistake on 2026-08-28; its withdrawal note is what surfaced my own. A reviewer agreeing with me is not verification, and a suggestion withdrawn on a false premise is worse than one never made, because it closes the question.
+
+    WHAT WAS ADDED. Identifiers were resolved from the entry's own MONDO term through OLS, not taken from the reviewer's message. That mattered: for Immunodeficiency 63 the reviewer offered ORPHA:444463, and MONDO:0032782 carries no Orphanet cross-reference at all, so no Orphanet record is asserted for that entry rather than a number I could not confirm. For Worth type, MONDO:0007764 cross-references two OMIM entries rather than one, so both are recorded instead of silently picking one.
+
+    Each record's description says what it is and, where relevant, that the Orphanet record is not present in references_cache/ - so it is an identifier for a reader to follow, not a source anything here quotes. That distinction is the point: recording a resolvable identifier is not the same as citing it.
+
+    Pushed knowing it resets an approval on an already-approved PR. Correcting a false claim I put on the record, and adding conventional metadata I wrongly said was impossible, is worth a review cycle.
+
+    Validation: just validate clean, snippet counts unchanged.
--- a/kb/disorders/Immunodeficiency_63_with_Lymphoproliferation_and_Autoimmunity.yaml
+++ b/kb/disorders/Immunodeficiency_63_with_Lymphoproliferation_and_Autoimmunity.yaml
@@ -1345,6 +1345,16 @@ discussions:
     explanation: >-
       The authors state the open question and why their own system cannot answer
       it.
+external_assertions:
+- name: OMIM immunodeficiency 63 entry
+  source: OMIM
+  assertion_type: disease_record
+  external_id: OMIM:618495
+  url: https://omim.org/entry/618495
+  description: >-
+    OMIM's phenotype entry, cross-referenced by MONDO:0032782 and resolved from
+    that term. MONDO carries no Orphanet cross-reference for this disease, so no
+    ORPHA record is asserted here.
 notes: >-
   Two things about this entry are worth a reader's attention. First, the disease
   is one receptor lesion with two opposite consequences, so the pathograph


matentzn:
All eight blocking findings fixed in `68456be3`, plus suggestion 9. Snippets went from 40/40 to **63/63 verified**.

Before the details: your diagnosis of the root cause is exactly right and I want to state it plainly rather than bury it. **I curated from the five abstracts when four of the five cached files contain full text.** `just fetch-reference` had already retrieved it; I did not read it. Every finding you raised was a fact sitting in this PR's own `references_cache/`. That is the single most useful thing in this review.

## CRITICAL — the CMV intermediate was backwards

You are right, and it is worse than a missing detail: the entry asserted a mechanism the cited literature specifically rules out. Patient NK cells "were capable of degranulation and target cell killing, which was enhanced by IL-2 or IL-15 priming". Cytotoxicity is not the failing step.

The intermediates are now the two the source actually supports — failed differentiation toward adaptive NK cells, and selective resistance to IL-2/IL-15-driven IFN-γ production while the IL-12/IL-18 route stays intact. That selectivity is a nice confirmation of the mechanism, incidentally: only the shared-beta-chain route fails.

I kept the cytotoxicity sentence as a **`REFUTE` item** rather than deleting it. The intuitive reading (immunodeficiency + NK defect + CMV → killing must be impaired) is one a future curator will reach for, and an explicit refutation closes it off in a way that silence does not.

## The model link was on the wrong mouse

Fair, and the `evidence_source` slip you spotted — a human paper graded `HUMAN_CLINICAL` supporting a claim about mouse NK cells — is the tell that the link was assembled from the human paper's throwaway comparison rather than from the model. Re-anchored to the knock-in's own divergence, `MODEL_ORGANISM`, with a readout: patients accumulate immature CD56-bright cells, the mice lose mature stage E and accumulate terminally differentiated cytotoxic ones. Inverted, not merely attenuated.

## Uncertainty moved, not deleted

Taking suggestion 11 seriously. The human NK picture is consistent across three alleles, so there was no inter-cohort disagreement to record and the old `KNOWLEDGE_GAP` was manufacturing one. But the caution belongs somewhere, and the model paper says where in its own words: *"questions regarding the threshold requirement for NK cell development and function remain and are not easily explored using the Il2rb Mut model system."* That is now the gap. The `HUMAN_MODEL_MISMATCH` is repointed at the knock-in's inverted maturation phenotype.

## Treg

Both papers measure it directly — `CD4+CD25+FOXP3+` decreased in the patient, and "clearly diminished in the two IL2RB-deficient patients examined". Both added as `HUMAN_CLINICAL`, hedge removed.

Worth flagging *why* I got this one wrong, since it generalises: the abstract says "an **anticipated** reduction in regulatory T cells", and I treated the authors' own hedge as authoritative. It is the abstract that understates the paper, not the paper that is tentative. An abstract's epistemic framing is not a reliable proxy for the results section's.

## Phenotypes, spectrum, treatments, variants

- Added AIHA (`HP:0001890`), hepatosplenomegaly (`HP:0001433`), failure to thrive (`HP:0001508`), interstitial pneumonitis (`HP:0006515`).
- Added the perinatally lethal end: a `Perinatal lethal presentation` progression phase and `Respiratory failure` (`HP:0002878`), both on the p.Gln96* kindred. You are right that an entry recording only infantile onset was missing roughly a third of reported affected individuals *and* the allele class that explains why.
- Treatments: one vague `INDIRECT` supportive entry became three named regimens with `DIRECT` evidence and `therapeutic_agent` bindings — methylprednisolone (`NCIT:C647`), sirolimus (`CHEBI:9168`), ganciclovir/valganciclovir (`CHEBI:465284` / `CHEBI:63635`). The antiviral evidence is unusually good because it is a withdrawal observation: viraemia escalated to 154,386 and 362,908 copies/ml whenever suppression lapsed. HSCT strengthened to two successful transplants plus five-year durable control — and the qualifier that it persists *despite* waning donor chimerism and falling wild-type Tregs, which bears on how much reconstitution is actually needed.
- Four alleles with their distinct mechanisms: Leu77Pro misfolds and is sequestered; Ser40Leu reaches the surface but cannot signal; Pro222_Gln225del disrupts WSXWS and loses expression; Gln96* truncates. Expression and function dissociate for Ser40Leu, which is worth knowing — surface staining alone would not find it.

## Suggestion 9 taken, 10 cannot be done

`GO:0035723` added beside `GO:0038110`, so the dual-cytokine node's bindings match its name.

The mappings block is not expressible: `DiseaseMappings` carries only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings`. There is no `omim_mappings` or `orphanet_mappings`, and schema validation rejects them. I hit the same wall on #10226. That looks like a genuine schema gap rather than an oversight in either PR, but I have not checked whether the omission is deliberate, so I have not filed anything.

## Validation

`just validate` passes schema, term and reference checks, **63/63 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.


---

matentzn:
Fixed four of the six suggestions in `42330c54`, which resets the approval. Three of them were correctness rather than style, and one is worth more than its "non-blocking" label.

**Suggestion 3 is the one I would not have caught.** PMID:31068380 is a JEM `Comment` on both index papers, so my "a second, independent human measurement" was wrong in a specific way: it is a second *reported* measurement of the same patients, not an independent replication. I corrected it and went through the other three explanations citing that source, which had the same problem in softer form — "two successfully transplanted patients", "holds across three different alleles". All now say explicitly that the source pools the index papers rather than adding cases.

The general lesson seems worth stating: **a citation from a distinct PMID is not automatically independent evidence**, and nothing in the validation stack knows the difference between a primary report and a commentary on it. Snippet validation confirms the quote is real; `evidence_source` says what kind of study the paper reports; neither says whether this paper's patients are the same patients as the last paper's. Two evidence items on one claim look like corroboration in the YAML and in any downstream export, whether or not they are.

**Suggestion 1** — right, and the correction is larger than the band. The sentence describes the two Fernandez siblings out of roughly ten patients worldwide, so it is a case description, not a cohort proportion. All three lowered to OCCASIONAL with the reason in the `description` rather than left as a bare band.

**Suggestion 2** — right, and the split is cleaner than a re-grade would have been. The allele identification is now `HUMAN_CLINICAL` and the surface-expression consequence is a separate `IN_VITRO` item with its own quote, so each carries one source type. Thanks for the note that `functional_impact_category` lives on `GeneticContext` rather than `Variant` — that is why the earlier suggestion did not map onto a slot.

**Suggestion 6** — snippet made exact.

**Not done:** the mappings block remains inexpressible (`DiseaseMappings` has only `icd10cm`/`icd11f`/`mondo`/`ncit`), and I left hepatosplenomegaly under `Gastrointestinal` since you called it defensible and no alternative is clearly better.

**64/64 snippets verified**, all offline gates pass.


---

matentzn:
Fixed in `4415be6`. You were right, and I checked the cache before acting rather than taking the review on trust.

## The blocking finding

`references_cache/PMID_31068380.md` line 40 reads:

> "**As assessed by flow cytometry on patient lymphocytes**, the IL2RB p.Pro222_Gln225del mutation drastically reduced IL-2Rβ surface expression."

So my explanation asserted a method the source rules out. The HEK-293T reconstitution work in that paragraph belongs to `p.Leu77Pro` and `p.Ser40Leu`, exactly as you say.

I took your second option, with one addition: rather than just regrading to `HUMAN_CLINICAL`, I extended the snippet to include the `As assessed by flow cytometry on patient lymphocytes` clause. The evidence item now carries its own method instead of depending on an explanation to assert one — which is what would have prevented this in the first place. Both explanations in the block are rewritten.

## Root cause, since it is the interesting part

Your previous review suggested this split on the premise that "the WSXWS-motif consequence is the in-vitro part". I implemented the suggestion without re-reading the full source paragraph, so a review suggestion propagated into a factual claim the source denies. That is the same failure mode as the two earlier findings on my other PRs — curating from something adjacent to the source instead of the source. A review suggestion is a lead, not a source.

Worth noting what did *not* catch it: `check-snippet-grading` is keyed on the quoted sentence, and this sentence appears once in the KB, so a single wrong grading is invisible to it. The gate catches divergence, not error.

## Suggestions

**Taken (item 2)** — the `OCCASIONAL` banding rationale for Hepatosplenomegaly and Failure to thrive moved from `description` (renders as a claim about the disease) into `notes` (curation reasoning). Agreed that was the wrong field.

**Not changed (item 4), with reasoning** — `category: Gastrointestinal` on Hepatosplenomegaly. I surveyed all 19 hepatosplenomegaly phenotypes in `kb/disorders/` and there is no convention to follow: 19 entries carry 14 distinct category strings (`Hematological`, `Hematologic`, `Abdominal`, `Clinical`, `Hepatic`, `Organomegaly`, `Constitutional`, `Other`, `Phenotypic abnormality`, `Gastrointestinal`, …). The two nearest neighbours by disease class — `Immunodeficiency_69` and `Autoinflammation_Immune_Dysregulation_and_Eosinophilia`, both primary immunodeficiencies — both use `Gastrointestinal`. Leaving it matches the closest precedent that exists. If the category vocabulary should be closed, that is a schema issue rather than something to settle one entry at a time.

**Item 3** — agreed and dropped; `DiseaseMappings` cannot express `OMIM:`/`ORPHA:`. I have not filed a schema issue for it as part of this PR.

## Validation

```
just validate                → schema ✅ terms ✅ references ✅  64/64 snippets verified
just check-snippet-grading   → OK (no new divergences)
just validate-history        → No issues found
```

Two files: the entry and a history record recording the correction and its cause.

---

matentzn:
**Correction: I was wrong about the schema, and you dropped a correct suggestion because of it.**

I told you `OMIM:618495` and `ORPHA:444463` "cannot be expressed" because `DiseaseMappings` exposes only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings`. You verified that reading and withdrew the suggestion. We were both wrong.

`ExternalAssertion` carries exactly this, with `source` and `external_id` both required, and it is the established convention rather than a corner of the schema:

```
$ grep -ho "external_id: ORPHA:[0-9]*" kb/disorders/*.yaml | wc -l   # 124
$ grep -ho "external_id: OMIM:[0-9]*"  kb/disorders/*.yaml | wc -l   # 65
```

124 ORPHA and 65 OMIM records across 110+ files. Fixed in `ab8e5b4` here, and in the four sibling PRs, which carried the same omission.

## How I found it, which is the uncomfortable part

Not by re-checking. I went to file the schema issue you suggested on #10226, searched for duplicates first — and found **#9922**, filed and withdrawn on 2026-08-28 for this identical mistake. Its withdrawal note is what corrected me.

The failure mode is the one that produced most of the other findings on these five PRs: I checked one class, didn't find the slot, and read *"absent from the class I looked at"* as *"absent from the schema"*. Same shape as the truncated `head -5` grep on #10231, and the same shape as quoting an abstract while full text sat in the cache.

Two things worth naming:

- **Your agreement was not verification.** You checked that `DiseaseMappings` has four slots — which is true — not that no other class carries the concept. I read the confirmation as a second opinion when it was the same lookup run twice.
- **A suggestion withdrawn on a false premise is worse than one never made.** Yours would have merged had I not argued it away. It closed the question, and it closed it in the append-only record on three PRs.

## What went in

Identifiers resolved from **MONDO via OLS**, not from your message. That mattered here specifically:

> `MONDO:0032782` carries **no Orphanet cross-reference at all**.

So `ORPHA:444463` is not asserted in this entry. I could not confirm it, and recording an unverifiable identifier to be tidy is how the "De Vitis et al." attribution on #10232 happened — a real number with an invented provenance. This entry records `OMIM:618495` and says explicitly in the record description that MONDO carries no Orphanet xref, so the next curator knows the absence was checked rather than overlooked.

Sibling PRs got the same treatment: Worth type turned out to have **two** OMIM entries cross-referenced by one MONDO term, so both are recorded rather than one silently chosen.

Each description also states whether the Orphanet record is present in `references_cache/`. Where it is not, the record is an identifier a reader can follow — not a source this entry quotes. **Recording a resolvable identifier is not the same as citing it**, and that distinction is exactly the thing the evidence model is otherwise good at keeping straight.

## No new issue filed

#7906 already covers the OMIM-slot question and is open; #9922 covers the ORPHA half and is correctly closed. Nothing to add that isn't already there, and filing a third would repeat the mistake in a new form.

`just validate` clean, **64/64 snippets** unchanged, `check-duplicate-keys` OK. This resets the approval, which I think is the right trade for removing a false statement from the record.

---

Issue #9321, opened 2026-08-23
Agents cannot re-request a review: workflow_dispatch is 403 and the /review token gets a footer appended


matentzn:
`CLAUDE.md` § *Never dismiss a review* is explicit that when a `CHANGES_REQUESTED` review is stale or an approval has been lost, the remedy is to get a **new** review rather than remove the old one:

```bash
gh workflow run claude-code-review.yml --repo "$REPO" --ref main -f pr_number=PR_NUMBER
```

In an agent session running against this repo through the GitHub MCP server (no `gh` CLI on PATH, `curl https://api.github.com` denied by the proxy), **neither of the two paths to that remedy works.** So the one instruction the guardrail gives as the alternative to dismissing a review is unavailable precisely to the actor the guardrail is aimed at.

## Path 1 — `workflow_dispatch`: 403

```
POST /repos/monarch-initiative/dismech/actions/workflows/claude-code-review.yml/dispatches
→ 403 Resource not accessible by integration
```

The MCP integration has no `actions: write`.

## Path 2 — the `/review` comment: silently skipped

`.github/workflows/claude-code-review.yml:113-116` gates the comment path on an **exact string match**:

```yaml
github.event_name == 'issue_comment' &&
github.event.issue.pull_request != null &&
github.event.comment.body == '/review' &&
contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
```

The comment was posted as the bare token `/review` by a `MEMBER`. But the tooling **appends the Claude Code attribution footer server-side**, so the body GitHub received was:

```
/review
```

`body == '/review'` is false, the job is skipped, and nothing distinguishes this from a deliberate decline. Observed on #9311: [run 32653332122](https://github.com/monarch-initiative/dismech/actions/runs/32653332122) → `claude-review` **skipped**, [comment as delivered](https://github.com/monarch-initiative/dismech/pull/9311#issuecomment-5387198147).

## Why this matters rather than being cosmetic

The trigger is needed exactly when something has already gone wrong. On #9311 the re-review job for `7f44b7d3e` [started at 16:31 and was cancelled at 16:39](https://github.com/monarch-initiative/dismech/actions/runs/32651904914/job/97224725394) without posting. The PR was then sitting with a dismissed approval and no way to ask for another — an approval gate left open by an infrastructure cancellation rather than by any outcome.

[...]

PR #10228, opened 2026-08-31
curate: Add Autosomal Dominant Osteosclerosis Worth Type (MONDO:0007764, LRP5)

### LINKED ISSUE #10222: Curate autosomal dominant osteosclerosis, Worth type (MONDO:0007764)
Curate a dismech entry for **autosomal dominant osteosclerosis, Worth type** ([MONDO:0007764](https://monarchinitiative.org/MONDO:0007764)) — a benign high-bone-mass endosteal hyperostosis caused by gain-of-function *LRP5* variants that block sclerostin/DKK1 binding to the Wnt co-receptor.

Preliminary read: this looks like a `DISEASE`. MONDO records no descendants, and *LRP5* (hgnc:6697) appears in `kb/` on `main` only as a passing gene mention, never as an entry's causal gene.

It is the receptor-side counterpart of the two ligand-side sclerosing dysplasias already in the KB — `Sclerosteosis` and `Van_Buchem_Disease`, both loss of the SOST inhibitor. Worth type instead makes the receptor deaf to that inhibitor, so the same Wnt/beta-catenin output is raised from the other end of the same axis. Worth curating partly because that contrast is what a mechanism knowledge base is for.

Stub: `stubs/Autosomal_Dominant_Osteosclerosis_Worth_Type.yaml`

matentzn:
New entry for Worth-type osteosclerosis, now usually called LRP5 high bone mass. `LRP5` appears in `kb/` on `main` only as a passing gene mention, never as an entry's causal gene, and no open PR covers it.

Closes #10222

## Why this one is worth having next to what we already hold

The KB already has `Sclerosteosis` and `Van_Buchem_Disease`. Both lose the SOST inhibitor. Worth type instead makes the receptor deaf to that inhibitor — same Wnt axis, opposite end, similar output:

```
LRP5 first-propeller missense
  -> receptor resistant to sclerostin and DKK1
    -> unrestrained canonical Wnt signalling in bone
      -> raised osteoblastic formation with NORMAL resorption
        -> endosteal hyperostosis and cortical thickening
          -> cranial foramina / posterior fossa encroachment (minority)
```

The "normal resorption" half is load-bearing, and it is measured, not assumed: Boyden's kindred had markedly elevated osteocalcin with normal resorption markers. That single dissociation is what separates this from osteopetrosis (failed resorption, brittle bone) and is why the disorder is benign despite Z-scores near +6.

The best mechanistic evidence is an in-vivo resistance test rather than a reporter assay: knock-in mice carrying human p.G171V or p.A214V do **not** lose bone mass when sclerostin is overexpressed, while wild-type mice do. That readout is recorded with `direction: UNCHANGED`, which is the informative result here.

## `GAIN_OF_FUNCTION`, not `INCREASED`

Following the CLAUDE.md rule, the Wnt node carries `modifier: GAIN_OF_FUNCTION`. The claim is qualitative — the pathway escapes its regulatory constraints because the receptor cannot see its inhibitors — not "the pathway is very active". Ligand abundance and receptor abundance are both normal. `functional_impact_category: GAIN_OF_FUNCTION` sits separately on the node's `genetic_context`, since that describes the variant rather than the pathway state.

## A frequency caveat I did not want to bury

The percentages in this entry (neurological involvement 19.4%, torus palatinus 41%, facial changes 61%, raised ALP 3.7%) all come from the 2023 historical review, and that review had to **pool pre-2002 clinically diagnosed Worth-type cases with post-2002 genetically confirmed LRP5 cases** to reach a series at all. The case definition changed halfway through. So those decimal places are more precise than the underlying data, and I have said so in `notes` and filed it as an open `KNOWLEDGE_GAP` rather than quietly using the numbers.

The same 2003 mutation survey that supplies the allelic series is the reason to worry: it found first-propeller LRP5 variants across *four* separately named diagnoses — endosteal hyperostosis, Van Buchem disease, autosomal dominant osteosclerosis and osteopetrosis type I — and concluded the labels do not track the genetics.

## One reference I deliberately did not use

PMID:41813941, "Romosozumab is effective in patients with variants in LRP5, LRP6, or WNT1", looks like an obvious treatment citation for an LRP5 bone disease. It is not: it studies *loss-of-function* and risk alleles causing bone **loss** — the opposite direction. Citing it here would have inverted the mechanism. Flagging it because it is exactly the kind of same-gene, wrong-direction citation that reads as relevant.

## About the deep research

Perplexity, as requested. Term validation set `needs_review: true` and flagged five nonexistent CURIEs (`HP:0000937`, `HP:0000596`, `HP:0001628`, `UBERON:0008938`, `UBERON:0009821`) and five mislabelled ones, including:

| Report offered | What the ontology calls it |
|---|---|
| `UBERON:0002240` "cranial vault" | spinal cord |
| `UBERON:0004473` "mandible" | musculature of face |
| `NCIT:C15194` "Maxillofacial surgery" | Bone Marrow Transplantation |
| `NCIT:C15387` "Dental prosthesis fitting" | Shave Biopsy |

None was bound. As in the other Perplexity reports in this run, the `reference_validation` block reports `on_topic: 6` of 7 but carries no `needs_review` key and no per-reference list, so the one reference that was not judged on topic cannot be identified from the report. All eleven references here were sourced independently through PubMed, including Gorlin & Glass 1977 — the radiographic paper that first separated this entity from Van Buchem disease, and still the cleanest statement of the anatomical distribution.

No GeneReviews chapter exists for this disorder.

## Validation

`just validate` passes schema, term and reference checks, **46/46 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.

---

### DIFF
--- a/history/disorders/Autosomal_Dominant_Osteosclerosis_Worth_Type/2026-09-06T143253Z-claude-code-c56667.yaml
+++ b/history/disorders/Autosomal_Dominant_Osteosclerosis_Worth_Type/2026-09-06T143253Z-claude-code-c56667.yaml
@@ -1,0 +1,71 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Dominant_Osteosclerosis_Worth_Type
+  path: kb/disorders/Autosomal_Dominant_Osteosclerosis_Worth_Type.yaml
+session:
+  id: 2026-09-06T143253Z-claude-code-c56667
+  timestamp: '2026-09-06T14:32:53Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10228
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: 'Fix: drop phenotypes# prefix from 10 causal targets'
+  details: |
+    The merge queue dequeued this PR with CI_FAILURE. check-causal-targets
+    reported ten findings, all in this entry, and the failure was real.
+
+    pathophysiology[].downstream[].target takes a BARE NAME, matched verbatim
+    against another item's name with no resolution step. All ten edges added in
+    review round 2 were written as 'phenotypes#<Name>', copying the form used by
+    attaches_to -- which is the entity-reference grammar, correct there and wrong
+    here. The two conventions sit four lines apart in this same file, and
+    attaches_to legitimately keeps disease# and genetic#.
+
+    What the prefixed form actually did is worse than a missing edge. Per
+    check_causal_targets.py's own docstring, build_causal_graph appends the edge
+    unconditionally, so the edge count never moves; the unresolved target lands in
+    orphan_targets and the renderer draws a PHANTOM DUPLICATE NODE -- red, dashed,
+    labelled with the literal string 'phenotypes#Diaphyseal sclerosis' -- while the
+    real phenotype node drops out of the graph. So the ten edges the reviewer asked
+    for, to close the pathograph, were fragmenting it instead.
+
+    Provenance of the error, recorded because it generalises: the reviewer wrote
+    the suggestion with that syntax verbatim, and I implemented what was written
+    without checking the two slots were the same kind of reference. A reviewer's
+    suggested code is not pre-validated. This is the second instance in this run of
+    implementing a suggestion without re-reading what it pointed at; the first was
+    the IL2RB evidence_source regrade on PMID:31068380.
+
+    Why no gate caught it on this branch: LinkML validation, term validation,
+    snippet verification and check-entity-refs all pass on a prefixed causal
+    target -- check-entity-refs parses that exact grammar and accepts it.
+    check-causal-targets is the only check that catches it, and it landed on main
+    (#10112) after this branch was cut, so it first fired inside the merge queue.
+
+    Fix is mechanical, as the check says: drop the prefix. All ten names were
+    verified to resolve to a real phenotypes[].name before stripping, rather than
+    assumed. The 'prefixed' finding class is explicitly never grandfathered, so
+    editing the baseline was not an option and would have been wrong anyway. The
+    other four entries in this curation run were checked for the same defect: zero
+    prefixed causal targets in any of them.
+
+    This record is late. It was declined at the time on the ground that pushing it
+    would reset an approval and eject the PR from the merge queue, buying a full
+    review cycle for an advisory check -- the reasoning is in the b2be6fce5 commit
+    message and in a PR comment. That cost has since disappeared: the PR was
+    dequeued by a genuine merge conflict, the approval is already gone, and the
+    entry needs a fresh review regardless. So the record is being written now that
+    it costs nothing, which is what the reviewer asked for and what the convention
+    requires for a commit that edits kb/.


matentzn:
Both blocking items fixed in `3d2947bb`, and I took all five suggestions — 3 and 5 were cheap as you said, and 6 turned out to be worth more than it looked. Snippets 46/46 → **54/54**.

## The forehead binding was flatly wrong

You caught a real internal contradiction: the phenotype bound *Prominent forehead* while the entry's own verified progression quote says the forehead **flattens**. The binding asserted the opposite of the evidence sitting two hundred lines above it.

I took your second option. The phenotype is now `Progressive coarse facial change` with **no term binding** — "no term beats a bad one" — and its description records the disagreement explicitly rather than picking a side. I also added two evidence items covering what the sources do agree on: the coarse, progressive quality ("she noticed the appearance of coarse 'facial changes' that became more evident with time") and the components, which are mandibular and palatal rather than frontal ("prominent torus palatinus, mandibular enlargement, and gonial angles protrusion").

**Your first option is currently unavailable, and it is not my choice.** I tried it: `just refresh-orphadata` fails its pinned sha256 gate for `en_product1.xml` (got `df8d562a…`, expected `fb2fbe8c…`), so no new ORPHA record can be cached and `structured-rebuild-orphanet --id 2790` has nothing to read. That is #9687 plus six open duplicates, so I did not file another. I did not bypass the checksum gate. Noted in the entry's `notes` so the next curator does not repeat the attempt.

## Surveillance re-evidenced

Right — a diagnostic proposal is not a management statement. Swapped in "In our case, we opted to continue observation with routine dental, audiology, and ophthalmology follow-up", which is on point and, as you noted, more informative. The description now names the three modalities.

I did not mark the PMID:22065804 item `INDIRECT` as suggested, because with the new primary evidence in place its role changed: it is no longer inferring a protocol from a diagnostic conclusion. I replaced the snippet with the six-year radiographic stability sentence from the same paper, which supports the "without intervention" half directly. Happy to reconsider if you disagree.

## Suggestions

**3.** You were right that "not brittle" was doing real work uncited — it is the entire basis for the osteopetrosis distinction, and the resorption-marker dissociation supports the mechanism difference but never the fracture claim. Now cited to "these skeletal changes do not seem to lead to an increased risk of fracture", in both the differential and the surveillance treatment.

**4.** Added a `Dental` phenotype for tooth loss/malocclusion from torus palatinus, bound to the `HP:0000174` palate-morphology parent with the reason stated, and with **no frequency** since the source gives none. Your point about it explaining the dentistry route to diagnosis is right and is now connected.

**5.** Added `p.Met282Val` as its own variant entry. The de novo origin matters more than the allele does — the entry did present the disorder as purely transmitted, and an absent family history should not argue against the diagnosis. I also recorded that the panel returned it as **of uncertain significance**, since the entry should not present it as more settled than its source does.

**6.** This was the best of the suggestions. The pathograph now closes: `Endosteal Hyperostosis` → Mandibular hyperostosis / Increased BMD / Thickened calvaria / Diaphyseal sclerosis, and `Cranial Foramina and Posterior Fossa Encroachment` → Neurological involvement / Hearing impairment / Anosmia / Facial palsy / Chiari I, plus Headache indirectly via raised intracranial pressure. Ten edges, and it makes the "minority pathway" claim visible in the graph rather than only in prose.

**7.** Top-level `references:` block added for all eleven PMIDs.

## One thing the gates caught mid-round

My first version of the `p.Met282Val` VUS evidence quoted a two-word fragment — `"Met282Val-uncertain significance)."` — which `check-snippet-length` rejected as non-propositional. Correct rejection: a fragment like that supports nothing. Replaced with the full sentence, which carries the allele, the de novo status and the classification together, so it is strictly more useful than what I first wrote.

## Validation

`just validate` passes schema, term and reference checks, **54/54 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.


---

matentzn:
Took the last suggestion too, in `f8146eb3` — which resets the approval, but your reasoning was better than what I had shipped.

You were right on both halves: `HP:0000174` re-described the torus rather than the sequela, and `HP:0000689`/`HP:0006480` would not have been honest either, because the source says a torus **can** cause tooth loss or malocclusion, not that these were observed in a patient with this disorder.

So the separate `Dental` phenotype is gone and the sentence now sits on the existing `Torus palatinus` entry with `directness: INDIRECT`, with the reason for not binding a dental term written into the `explanation`. Net effect: one fewer phenotype, one fewer ontology binding pointing at the wrong concept, and the information kept with its epistemic status visible instead of implied by a binding that could not carry it.

That is a small instance of something this review round kept surfacing: the strongest fix is often to remove an assertion rather than to qualify it. Same shape as deleting the chondrogenesis edge on #10226 and unbinding the forehead term here.

**54/54 snippets verified** (unchanged), all offline gates pass.


---

matentzn:
**Correction carried over from the sibling PRs, plus one finding specific to this entry.**

On three PRs in this run I claimed OMIM and Orphanet identifiers were "inexpressible" because `DiseaseMappings` exposes only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings`. That is false. `ExternalAssertion` carries exactly this, with `source` and `external_id` required — 124 ORPHA and 65 OMIM `external_id` records across 110+ files in `kb/disorders/`. I found out by searching for duplicates before filing a schema issue and hitting #9922, filed and withdrawn for the same mistake three days earlier.

The error was reading *"absent from the class I looked at"* as *"absent from the schema"* — structurally the same as the truncated cache grep on #10231.

## The entry-specific part

Resolving the identifiers from `MONDO:0007764` via OLS rather than trusting a secondary source turned up something worth a curator's attention:

> **This MONDO term cross-references two OMIM entries, not one** — `OMIM:144750` and `OMIM:607636`.

I recorded both rather than silently picking one. Which OMIM concept this entry corresponds to is a real question — Worth-type endosteal hyperostosis and the LRP5 high-bone-mass phenotypes have a tangled nosology, which is the same tangle behind the forehead-morphology disagreement you caught in round 1. Asserting one OMIM number would have made a lump/split call by omission and hidden it in a metadata field. The second record's description says plainly that MONDO cross-references two and that both are recorded for that reason.

Also added `ORPHA:2790`, with its description noting the record is not in `references_cache/` — so it is an identifier to follow, not a source this entry quotes. That distinction matters here specifically because of the blocked Orphadata refresh (#9687) you verified in round 2: the identifier being recordable does not make the record quotable, and the entry should not imply otherwise.

Fixed in `4a2fbc2`. `just validate` clean, **54/54 snippets** unchanged, `check-duplicate-keys` OK. Two files.

This resets the approval. Given that the omission came from a claim I argued for and a reviewer accepted, correcting it is worth the cycle.

---

matentzn:
Dequeued at 05:53Z with `CI_FAILURE`. **The failure is real and it is mine.** Fixed in `b2be6fc`.

## What failed

The merge-queue run for the batch ([33840074845](https://github.com/monarch-initiative/dismech/actions/runs/33840074845)) failed `test (3.13)` with 10 findings from `check-causal-targets`, every one in this entry:

```
kb/disorders/Autosomal_Dominant_Osteosclerosis_Worth_Type.yaml
   pathophysiology.downstream: 'Cranial Foramina and Posterior Fossa Encroachment' -> 'phenotypes#Headache'
   fix: target: Headache
```

`pathophysiology[].downstream[].target` takes a **bare name**, matched verbatim against another item's `name` with no resolution step. I wrote all ten of the edges added in review round 2 as `phenotypes#<Name>` — copying the form used by `attaches_to`, which is the entity-reference grammar and correct *there*, and wrong here.

## What it was actually doing

Worse than a missing edge. From the check's own docstring:

> `build_causal_graph` appends every declared edge unconditionally, so the edge count never moves; the unresolved target instead lands in `orphan_targets`, and the renderer draws it as a **phantom duplicate node** — red fill, dashed red border, reached by a dashed edge, labelled with the literal `phenotypes#Bleeding tendency` string. The real node it should have connected to drops out of the graph entirely.

So the ten edges @ai4c-reviewer asked for in round 2 were **fragmenting** the pathograph rather than connecting it: ten phantom nodes drawn, ten real phenotype nodes dropped. The rendered page would have looked visibly wrong, and the reviewer's suggestion — which was a good one — was silently inverted by my implementation of it.

## Why nothing caught it here

Every gate on this branch passes on a prefixed target: LinkML validation, term validation, snippet verification, and **`check-entity-refs`** — which resolves `phenotypes#Headache` happily, because in the slots *it* governs that string is correct. `check-causal-targets` landed on `main` after this branch was cut, so it only ran inside the merge queue against merged content. Same shape as the `check-enum-values` absence @ai4c-reviewer noted on #10226.

## The fix

Mechanical, as the check says: drop the prefix. I **verified all ten names resolve to a real `phenotypes[].name`** before stripping rather than assuming the rewrite was safe. The `prefixed` finding class is explicitly *never* grandfathered — `update-causal-target-baseline` exists to burn down the `dangling` backlog, not to admit new breaks — so baselining was neither available nor right.

```
just check-causal-targets   → OK
just check-entity-refs      → OK
just check-duplicate-keys   → OK
just validate               → schema ✅ terms ✅ references ✅  54/54 snippets verified
```

`validate-graphs` still reports the pre-existing KB-wide `dangling` backlog (#9697 — SLE, Tuberculosis, Viral Encephalitis and others). **This entry is not among those findings**, and that check is not wired into CI.

## Checked the rest of the run

Zero prefixed causal targets in the other four entries — including #10227, which merged at 05:53Z. This defect was isolated to the ten edges added here.

Worth naming, because it is the same pattern that has produced most of my errors on these PRs: a form that is correct in an adjacent slot, applied to this one without re-reading what this slot takes. `attaches_to` and `downstream.target` sit a few lines apart in the same block and want opposite things.

---

matentzn:
Thanks — and thanks for naming where the prefixed syntax came from rather than letting it read as my transcription error. Two things on that, then the one note.

I don't think "the reviewer wrote it" gets me off the hook. I had the file open, the entry already used `attaches_to: disease#…` and `genetic#…` correctly a few lines away, and I copied a form across a boundary without checking the two slots were the same kind of thing. A suggestion is a lead; the file was the source and it was sitting right there. That is the same failure that produced the IL2RB `IN_VITRO` regrade earlier in this run — implementing a suggestion without re-reading what it was pointing at — so it is the second time in five PRs, which is the part worth recording.

**On the missing history record: I'm not adding one, and I want to give the reason rather than just leave the gap.**

A push here resets the approval and ejects the PR from the merge queue, buying another full review cycle plus another window in which `main` moves — and `main` moving is not hypothetical on this branch, it has already produced two cache conflicts and one dequeue in the last six hours. The repo made this check advisory rather than blocking, and you explicitly declined to hold for it, so I read both as "not worth a cycle." The curation guidance is also fairly direct that approved is the finish line and an approved PR should not be polished further.

What I did instead: the reasoning is in the commit message on `b2be6fce5`, in full — the bare-name-vs-entity-reference distinction, the phantom-node failure mode from `check_causal_targets.py`'s own docstring, why every other gate passed, why the baseline was not an option, and the check that the other four entries in this run carry zero prefixed targets. `git log` is not `history/`, but it is durable, greppable, and attached to the exact commit. If you would rather have the record and think the cycle is worth it, say so and I'll push one — that is a reasonable call to make differently, and I'd rather be told than guess.

---

matentzn:
Dequeued from the merge queue with `MERGE_CONFLICT`; fixed in `851510d3b`.

Two derived cache files conflicted, and both were the **adjacent-sibling** shape rather than the same-row shape:

```
cache/mondo/terms.csv          this branch: MONDO:0007764  Worth type
cache/enums/diseaseterm_*.csv  main:        MONDO:0007765  hyperostosis cranialis interna
```

Different CURIEs that happen to sort next to each other, so both rows belong and nothing is in tension. Resolved as a union with `0007764` before `0007765`, rather than by picking a side — dropping either would silently lose a validated term from a derived sorted file.

Worth naming because it is the **opposite** call from the one made on the CCMS branch (#10226) six hours ago. There the conflict was `UBERON:0002228,rib` present on both sides with the same CURIE *and* the same label, differing only in the resolution timestamp — and the answer was to keep one row and take main's, because a union would have put two rows under one CURIE and broken the term-cache contract. Same file type, same conflict marker, opposite resolution. The shape has to be read off the row content each time; there is no rule that covers both.

Verified rather than eyeballed, since a dropped row is exactly what a diff makes easy to miss:

- exactly one row per CURIE, by explicit grep on all four rows
- `check-term-cache-integrity` — 18 term + 24 enum files match the structural contract
- `check-cache-order` — canonical
- `just validate` re-ran normalization and reported **`Files updated: 0`**, which is the check that actually shows the union landed in sorted position rather than being silently fixed up afterwards

Gates on the merged tree: `just validate` clean, plus `check-causal-targets`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.

Scope against `main` is unchanged at 23 files, +3865/−32. The stub is still deleted, no derived artifacts, and the only deletion in the diff is that stub.

This resets the approval, which I would rather avoid — but a merge conflict has to be resolved on the branch, so there is no version of this that does not cost a review cycle.

---

matentzn:
The re-review on `851510d3b` did not run, and this one needs a maintainer. @matentzn

[Run 33879554653](https://github.com/monarch-initiative/dismech/actions/runs/33879554653) errored after 0 seconds:

```
detail="You've hit your limit · resets Sep 5, 8pm (UTC)"  num_turns=1  cost=$0
```

An account quota on the review credential, not anything about this PR. A reset **30+ hours out** means it is the weekly cap rather than the daily one, so this will not clear on its own today — and it blocks every agentic review in the repo meanwhile, not just this PR. The workflow's own guard describes the fix: set the `ANTHROPIC_API_KEY` repo secret (pay-per-use, no weekly cap) so the run stops falling back to the subscription token, or rotate the OAuth token to an account with headroom.

Two things that follow from that, so nobody re-checks ground already covered:

**This PR is otherwise finished.** The merge conflict that dequeued it is resolved in `851510d3b`, and every gate passes on the merged tree — `just validate` clean, plus `check-causal-targets`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading`. Scope is unchanged at 23 files, +3865/−32, stub deleted, no derived artifacts. The content is byte-identical to the commit approved at 06:48Z; the only change is the cache union.

**#10231 is blocked on the same credential**, for a different reason — its approval was auto-dismissed by a merge and its review has now failed on quota twice. Both PRs clear together once the key is sorted.

I am not going to work around this. Re-triggering by comment does not reach the workflow from here (#9321 — the `/review` token gets an attribution footer appended server-side, so the exact-match gate skips it), `workflow_dispatch` returns 403 for this integration, and there is no genuine content change left to push, so any commit made to kick CI would be a no-op wearing a commit message.

---

matentzn:
Updating my earlier comment, which now names the wrong cause. @matentzn

The weekly cap it described reset yesterday, and reviews are running repo-wide again. I pushed `f60d28b4c` — the missing `history/` record for the causal-target fix, which is the non-blocking item raised in the last round and declined at the time — and that triggered a fresh review. It got further than before but still did not post:

```
detail="You've hit your limit · resets 5:40pm (UTC)"   num_turns=25   cost=$1.75
```

Different failure from the ones on Sep 4. Those were 0 turns and $0 — the agent never started. This one ran 25 turns and spent $1.75 before being cut off, so it is a **rolling** limit rather than the weekly one, and it resets at **17:40 UTC today**. The review was most of the way through the work when it stopped.

On the record itself, since the review did not get to say so: it documents the bare-name vs entity-reference distinction, the phantom-node failure mode from `check_causal_targets.py`'s own docstring, why every other gate passed on a prefixed target, that `check-causal-targets` (#10112) landed on `main` after this branch was cut, and that the wrong syntax came from the review suggestion and was implemented without checking the two slots were the same kind of reference. `validate-history` clean; `check-causal-targets`, `check-entity-refs`, `check-duplicate-keys` and `check-folded-hyphens` all pass. No KB content change, so the entry is byte-identical to the one that was approved.

Why I declined this record on Sep 4 and wrote it now: the reason then was cost, not disagreement — the PR was approved and in the merge queue, so any push would have reset the approval and ejected it for an advisory check. That cost disappeared when a genuine merge conflict dequeued it. The record now costs nothing and the PR needed a review regardless.

**I am not going to push again to retrigger.** There is no further genuine change on this branch, and a commit made only to kick CI would be a no-op wearing a commit message. After 17:40 UTC this needs one command from someone with `actions: write` — `workflow_dispatch` returns 403 for this integration, and the `/review` comment path is still gated on `body == '/review'` (#9321), which the server-side attribution footer defeats:

```
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10228
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10231
```

#10231 is in the same position and has no outstanding work at all — its entry is byte-unchanged by both merge commits and all five content commits already carry history records.

---

PR #10231, opened 2026-08-31
curate: Add TMEM199-CDG (MONDO:0014790), a Golgi-homeostasis liver disease

matentzn:
New entry for TMEM199-CDG, the type II CDG that presents as liver disease rather than as a neurological one. `TMEM199` appears nowhere in `kb/` on `main`.

Closes #10223

## Deliberate adjacency to #10097

[#10097](https://github.com/monarch-initiative/dismech/pull/10097) curates CCDC115-CDG, the sibling Golgi-homeostasis disorder. Different gene, different MONDO term, same mechanism class, and the two are routinely reported together — the lipid-droplet paper studies both in the same experiments. **This is a sibling, not a duplicate.** I kept the files fully independent: no cross-file entity references, so neither PR depends on the other merging, and CCDC115-CDG appears here only as a free-text differential.

## The mechanism

```
TMEM199 (V-ATPase assembly factor) deficiency
  -> disturbed Golgi homeostasis / acidification
     -> combined N- and O-glycosylation defect -> type II transferrin profile
     |                                          -> abnormal copper/ceruloplasmin handling
     -> impaired lysosomal acidification + autophagy -> hepatic steatosis
  (separately) dysregulated apoB secretion -> hypercholesterolaemia
```

Two edges here are carried by unusually good evidence and both are worth a reviewer's eye:

- The gene→glycan link is a **rescue**: metabolic sialic-acid labelling in patient fibroblasts is restored by lentiviral transduction with wild-type TMEM199.
- The steatosis edge rests on a **negative result**: the lipid-droplet study explicitly ruled out excess de novo lipogenesis, failing oxidative capacity and elevated uptake, which is what leaves failed clearance as the supported mechanism. The `explanation` on that item says so, because the negative is the load-bearing part.

The mouse carries a `FAILS_TO_RECAPITULATE` link: the A7E knock-in reproduces the hepatic steatosis and the hypogalactosylated plasma N-glycans, but shows **no** plasma lipid abnormality. So the model cannot be used to test the lipoprotein arm, and the entry says which system each arm has to come from.

## Two things left open

- **Why ceruloplasmin is low.** Is ceruloplasmin itself hypoglycosylated and cleared faster, or does copper loading fail in an abnormally acidified compartment? Nothing in the literature separates these in patients. The edge into that node is `INDIRECT_UNKNOWN_INTERMEDIATES` and there is a matching `KNOWLEDGE_GAP`.
[...]

Issue #9832, opened 2026-08-28
Reference validator strips bracketed confidence intervals and unit conversions, forcing curators to trim exact quotes


aleixpuigb:
The reference validator strips square-bracketed spans from a snippet before matching it against the cached source, unless the bracket content matches one of the two patterns in `conf/reference_validator_config.yaml`. Those patterns are `^[A-Z]{2,}$` and `%`. Neither covers two things that are very common in the biomedical abstracts we quote:

- **Confidence intervals**: `[25.4, 95% confidence interval (CI), 19.8-32.0]`, `[1.3-1.7]`
- **Bracketed unit conversions**: `(>43 micromol/L [>2.5 mg/dL])`

Quoting either verbatim fails validation. The failure message reads "Text part not found as substring", which is the message that exists to catch a curator paraphrasing a source — so a correct verbatim quote produces the signal for fabrication.

## Where this came up

Three times, independently, in two curation PRs today:

1. #9830 (Gilbert's), quoting an odds ratio: `only cholelithiasis prevalence was significantly higher (OR=1.50; 95% CI [1.3-1.7], P=0.001)`
2. #9830, quoting a jaundice threshold: `in the jaundice range (>43 micromol/L [>2.5 mg/dL])`
3. #9831 (Cowden), in the **deep-research report's own** `## Reference Validation` section, which reports one quoted claim as "not found in source" and then prints a "closest text in source" that is byte-identical to it:

```
- `PMID:22252256`: "Elevated SIRs were found for carcinomas of the breast [25.4, 95% confidence interval (CI), 19.8-32.0], thyroid (51.1, 38.1-67.1) ..."
  - closest text in source: "Elevated SIRs were found for carcinomas of the breast [25.4, 95% confidence interval (CI), 19.8-32.0], thyroid (51.1, 38.1-67.1) ..."
```

Item 3 is the one that concerns me most. That report section is what the curation skill tells curators to read before citing anything from a deep-research report, and it is telling them a correct quote is unverified. The skill's own guidance — "anything under `unresolved_references`, do not cite it" — trains people to drop the claim. Here that would mean dropping a correctly-quoted cancer-risk figure because of a bracket.

## Why this pattern matters more than an ordinary gloss

The config comment explains the existing patterns well: stripping exists so a curator can gloss a quote (`This pleiotropic cytokine [IL-6] is a key factor`) without transcribing the gloss. That reasoning is sound and I am not proposing to remove it.

[...]

Issue #8198, opened 2026-08-08
ARVC: GeneReviews baseline (PMID:20301310) is cached but unmined and untagged


caufieldjh:
Split out from the review of #8191, where the reviewer flagged this and explicitly argued it should not block that PR.

## Problem

`kb/disorders/Arrhythmogenic_Right_Ventricular_Cardiomyopathy.yaml` never cites the ARVC GeneReviews overview, even though the reference is **already cached** at `references_cache/PMID_20301310.md`. Specifically the entry has:

- no top-level `references:` block
- no `tags: [GeneReviews]` entry for PMID:20301310
- no evidence item citing it anywhere

Under the `initiate-new-disorder-creation` skill (Step 3b) GeneReviews is the *mandatory* phenotype baseline for a Mendelian disorder, not merely a convenient source. ARVC is `category: Mendelian`, so the baseline applies.

## Why this wasn't fixed in #8191

#8191 was a 71-line targeted augmentation (pathograph wiring, one MONDO mapping, two imaging term bindings) to a mature entry already at ~90% compliance, and this gap predates it. The reviewer's own assessment was that blocking a small scoped fix on a pre-existing four-section mining task would be scope creep — recording it here instead so it isn't lost.

## What needs doing

1. Add a top-level `references:` block tagging PMID:20301310 with `tags: [GeneReviews]` (or run `just tag-references` once it is cited inline).
2. Mine the cached abstract and add evidence items with **exact** GeneReviews snippets backing at least one claim in each section present in that abstract:
   - Clinical Characteristics
   - Diagnosis
   - Management, including exercise restriction
   - Genetic Counseling / transmission risk, on the `inheritance` entry
3. Cross-reference the GeneReviews Clinical Characteristics against the existing `phenotypes:` list. Any GeneReviews-documented phenotype absent from the YAML must either be added with an HPO term and evidence, or explicitly noted as out of scope.
4. While there: the 9 existing phenotypes all lack `frequency:`. GeneReviews prose frequency language maps onto `FrequencyEnum` via the table in the skill, but per `docs/frequency-evidence-guidelines.md` each frequency band needs its own evidence — omit rather than guess where the abstract is ambiguous.

## Validation

Snippets must be exact substrings of the cached file.

```bash
just count-verified-snippets kb/disorders/Arrhythmogenic_Right_Ventricular_Cardiomyopathy.yaml
[...]

PR #10232, opened 2026-08-31
curate: Add Pancytopenia-Developmental Delay Syndrome (MONDO:0014317, ERCC6L2)

### LINKED ISSUE #10225: Curate pancytopenia-developmental delay syndrome (MONDO:0014317)
Curate a dismech entry for **pancytopenia-developmental delay syndrome** ([MONDO:0014317](https://monarchinitiative.org/MONDO:0014317), also bone marrow failure syndrome 2 / BMFS2) — biallelic *ERCC6L2* loss, giving progressive trilineage marrow failure with a strong tendency to erythroid-dysplastic MDS/AML carrying *TP53* lesions.

Preliminary read: this looks like a `DISEASE`. MONDO records no descendants, and *ERCC6L2* (hgnc:26922) appears nowhere in `kb/` on `main`. It sits under the `bone marrow failure syndrome` and `inherited aplastic anemia` parents, both of which already have entries — so the entry needs to say what distinguishes it rather than restate the parents.

The name understates it: the developmental delay is inconstant, and the defining features are the marrow failure and the specific leukaemic transformation pattern. Worth checking whether the MONDO label should be flagged upstream.

Stub: `stubs/Pancytopenia-developmental_Delay_Syndrome.yaml`

matentzn:
New entry for ERCC6L2 disease. `ERCC6L2` appears nowhere in `kb/` on `main`, and no open PR covers it.

Closes #10225

## The thing that makes this disease unusual

The somatic event that **corrects** the marrow failure is the same one that **causes** the leukaemia.

```
ERCC6L2 loss -> failed transcription-coupled repair, R-loop accumulation
  -> replication stress in HSPCs
    -> p53 activation -> arrest and apoptosis -> marrow hypoplasia
       -> somatic TP53 mutation RESTORES stem cell fitness, correcting the BMF
          ...but does NOT normalise the replication stress
             -> TP53-mutant, erythroid-predominant MDS/AML
```

The mouse paper shows this directly: somatic `Trp53` mutation abrogates p53 activation, restores Runx1 and Gata1, and corrects the marrow failure phenotype — while leaving the replication stress untouched, so damage keeps accumulating in a clone that can no longer arrest or die. Clinically that means **a patient whose counts improve is not necessarily getting better**, which is in `notes`.

The numbers behind it, from the 52-subject / 1165 person-year cohort: median age at malignancy 37 years, 3-year overall survival 95% in the marrow-failure phase and 19% after transformation. And the trap for diagnosis: counts are mild *despite* severe marrow hypoplasia, so a full blood count understates the disease and surveillance has to be marrow-based and molecular (rising TP53 VAF, erythroid/megakaryocytic dysplasia).

`TP53` is curated as a **second `genetic:` entry with `relationship_type: SOMATIC_DRIVER`**, not as an incidental second hit — it is selected for by the disease's own biology.

## One phenotype carries a SUPPORT and a REFUTE

`Developmental delay` has both:

- **SUPPORT** — universal in the six-patient 2018 series that defined the phenotype ("All patients displayed bone marrow failure, learning or developmental delay and microcephaly").
- **REFUTE** — the 2025 Croatian report describes two adolescents with "unremarkable medical history and normal physical findings", and exists precisely to show the disease occurs without neurodevelopmental delay.

Per the CLAUDE.md rule for evidence that supports one part of a claim and contradicts another, that is two items, not one hedged one.

Which leads to an open discussion I would like a reviewer's opinion on: **the MONDO label understates the disease on both of its named features.** The cytopenias are mild relative to the marrow, the developmental delay is inconstant, and the defining feature is TP53-mutant transformation. The field has largely moved to "ERCC6L2 disease". This is not just terminology — a clinician expecting pancytopenia plus developmental delay will not think of the diagnosis in an otherwise well adolescent with bicytopenia, which is exactly the presentation the Croatian paper was written to flag. Filed as a `KNOWLEDGE_GAP` rather than acted on unilaterally.

## A reference I could not use, and why that is a gap

PMID:42415356 — *"Advanced fibrosis, TP53 variants and acute/pure erythroid leukaemia: Diagnostic harbingers of ERCC6L2-related bone marrow failure"*, Br J Haematol 2026 — is squarely relevant. Its PubMed record has **no abstract**, so `just fetch-reference` cached a title-only file and there is nothing quotable. A title is not a finding, so it is not cited; it is named in `notes` instead.

Flagging it because "relevant but unquotable" is a category the KB has no slot for. The entry can either cite it with a fabricated-feeling snippet, or lose the knowledge. I chose to lose the citation and keep a prose pointer, but a `references:` entry with a tag meaning "read, relevant, no quotable text" would be more honest than either.

## About the deep research

Perplexity, as requested. Term validation set `needs_review: true` with **13 of 38 checked labels naming a different term**, and one pattern is worth naming: **every NCIT term the report offered was wrong.**

| Report offered | What NCIT calls it |
|---|---|
| `NCIT:C15245` "Blood Transfusion" | Health Services Research |
| `NCIT:C28182` "Antibiotic Therapy" | Rabies |
| `NCIT:C15743` "Colony-Stimulating Factor Therapy" | Medical Castration |
| `NCIT:C16010` "Supportive Care" | Scientist Exchange Program |
| `NCIT:C15514` "Hematopoietic Stem Cell Transplantation" | Psychosocial Assessment and Care |
| `NCIT:C15243` "Bone Marrow Transplantation" | Health Promotion |

Plus `HP:0005518` offered as "bone marrow hypocellularity" (HPO: *Increased mean corpuscular volume*) and `HP:0002352` as "cerebellar atrophy" (HPO: *Leukoencephalopathy*). `GO:0007067` is obsolete. Citation 18 is the bare string "ACMG guidelines". None of these was bound; every term here was resolved independently.

## One thing I got wrong

The 3-year survival snippet was transcribed as "…for patients with BMF and **HMs**, respectively" where the abstract says "**HM**". `just validate` flagged it as an abstract-only near-miss and it was corrected to the exact substring. A one-character transcription slip is enough to make a snippet not a quote, and the gate caught it.

## Validation

`just validate` passes schema, term and reference checks, **49/49 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.

---

### DIFF
--- a/history/disorders/Pancytopenia-Developmental_Delay_Syndrome/2026-08-31T161758Z-claude-code-e74d89.yaml
+++ b/history/disorders/Pancytopenia-Developmental_Delay_Syndrome/2026-08-31T161758Z-claude-code-e74d89.yaml
@@ -81,7 +81,28 @@ events:
     for rather than incidental - it reverses the p53-dependent attrition that causes
     the marrow failure.
 
-    No GeneReviews chapter exists for this disorder.
+    CORRECTION (appended 2026-08-31T18:52Z, before merge): the sentence below
+    originally read "No GeneReviews chapter exists for this disorder." That is
+    false as written. No DEDICATED chapter exists, but the GeneReviews "Monosomy
+    7 Predisposition Syndromes Overview" (NBK571103) covers this disorder, and
+    it was already cached in this PR's own references_cache/ when I wrote that
+    line. Its Table 1 gives ERCC6L2 / Bone marrow failure syndrome 2
+    (OMIM 615715) / AR / 1%-5% of monosomy 7 predisposition / adult onset /
+    "Microcephaly, learning difficulties, DD", and its physical-examination
+    section names ERCC6L2-related bone marrow failure among the causes of
+    microcephaly. An absolute "does not exist" would have told the next curator
+    not to look. This is the third time in this run I recorded an absence that
+    the cache contradicted; the other two were a grep truncated by head -5 and
+    an ASCII search for a full-width numeral.
+
+    Not cited, and the reason is itself worth recording: that cache is stored as
+    raw HTML, and every ERCC6L2 mention is wrapped in <i> tags mid-sentence, so
+    no sentence naming the gene survives as an exact substring. I added the
+    microcephaly sentence as an evidence item, ran just validate, got "Text part
+    not found as substring", and removed it rather than ship an unverified
+    snippet. A reviewer suggested the chapter was quotable; it is quotable on
+    screen and not in the bytes, which is the same rendered-form-versus-actual-
+    bytes mistake as the full-width numeral, made in the other direction.
 
     Validation: `just validate` passes schema, term and reference checks with 49/49
     snippets verified. check-entity-refs, check-duplicate-keys, check-folded-hyphens,


matentzn:
All four blocking items fixed in `c4878c67`, plus the histopathology, references-block and phenotype suggestions. Snippets 49/49 → **59/59**.

## The inverted REFUTE item

Your catch on finding 1 is the one that mattered most: the item was graded `REFUTE` while quoting a sentence that *affirms* the association. As written it asserted the opposite of what it read, and a reader skimming the grade would take away the wrong conclusion. Replaced with the sentence that actually refutes, from the same cached full text: *"Our patients, both homozygous for this variant, as well as the brother of one of them with the same genotype, did not exhibit microcephaly, developmental delay, or dysmorphic features."* Three individuals, one genotype, none of the features.

## Chasing your 19% turned up something

You were right that the bands were wrong, and tracing the figure produced a finding worth flagging on its own.

The deep-research report attributes the 19% to **"De Vitis et al."**. No such paper exists — `De Vitis[AU] AND ERCC6L2` and `Vitis[AU] AND ERCC6L2` both return nothing from PubMed.

But the *number* is real. Following the report's own citation marker `[11]` to `PMC9998559` resolves to **PMID:36790458, Baccelli et al. 2023**, which says: *"Regarding neurological features, 6/31 patients presented microcephaly"* and *"Developmental delay and/or learning difficulties were also described in 6 patients"*.

So the report got the fact right and invented the attribution. That is a failure mode worth naming, because it survives every check we run: verifying that a cited claim is *true* is not the same as verifying that the *named author wrote it*, and nothing in the stack does the second. I have fetched Baccelli and cited it directly, recorded the misattribution in the entry's `notes`, and both phenotypes are now `OCCASIONAL` on a real source.

That paper also turned out to be the best frequency source in the whole reference set — MDS/AML 10/31, marrow morphology 20 hypocellular vs 6 dysplastic, the trilineage cytopenia ranges — so the finding paid for itself several times over.

## The other two

**Conditioning is not supportive care.** Right. The separate treatment entry is gone and its evidence folded into the HSCT node, whose `notes` already carried the conditioning guidance — the `Inherited_Aplastic_Anemia.yaml` pattern you pointed at.

**The MDS phenotype.** Renamed to `Myelodysplastic syndrome or acute myeloid leukemia`, since the literature reports them as one progression endpoint with a combined denominator, and re-evidenced with *"approximately a third of the whole cohort of patients (10/31) developed MDS or AML"* plus your suggested hallmark sentence. I also added the TP53-clone proportion — 7/10 progressors and **none** of the non-progressors — which makes the association non-correlative in that cohort, and the monosomy 7 finding at the same 7/10.

## Suggestions

Added a `histopathology:` section with both marrow pictures and their denominators — hypocellular without dysplasia in 20 of 26 with reported morphology at presentation, versus megakaryocytic/erythroid dysplasia with erythroid predominance as the progression picture. That distinction is worth having as histopathology rather than prose, because it is the difference between the presenting marrow and the marrow you are watching for.

Added Thrombocytopenia, Anemia and Neutropenia at VERY_FREQUENT, and a top-level `references:` block for all nine PMIDs.

Not done: OMIM/ORPHA mappings cannot be expressed — `DiseaseMappings` supports only `icd10cm`/`icd11f`/`mondo`/`ncit` sub-slots and schema validation rejects the others. Third PR in this run to hit that wall.

## GeneReviews — confirmed, no chapter

You flagged this as the one thing you could not check, and correctly said it would be blocking if a chapter existed. Checked from here: `ERCC6L2 GeneReviews[All Fields]` returns nothing, and a `GeneReviews[Book]` search on the disease terms returns only *other* IBMFS chapters — Fanconi anemia, dyskeratosis congenita, SAMD9L-related syndrome, monosomy 7 predisposition. None is ERCC6L2. No chapter exists.

## One tooling note

The trilineage-cytopenia sentence in PMID:36790458 carries a mid-sentence PMC citation marker — *"between 4000 and **[23]** 166,000/μL"* — and the reference validator strips bracketed spans before matching, so the full sentence cannot be quoted verbatim. I shortened the snippet to the clause before the marker. Same class as #10192 (HGVS in square brackets); noting rather than filing, since that issue already has the mechanism.

## Validation

`just validate` passes schema, term and reference checks, **59/59 snippets verified**. `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all pass.


---

matentzn:
**Your "still unused" list had `mappings: (OMIM:615715, ORPHA:401764)` on it. You were right and I was wrong to treat it as blocked.**

On three PRs in this run I claimed those identifiers were "inexpressible" because `DiseaseMappings` exposes only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings`. False. `ExternalAssertion` carries them, with `source` and `external_id` both required, and it is the established convention — 124 ORPHA and 65 OMIM `external_id` records across 110+ files in `kb/disorders/`. I found out by searching for duplicates before filing a schema issue and hitting #9922, filed and withdrawn for the same mistake three days earlier.

The reasoning error: I read *"absent from the class I looked at"* as *"absent from the schema"*.

That is worth stating plainly on **this** PR in particular, because it is the same failure mode as the finding you singled out here. The "De Vitis et al." attribution was a real number with an invented provenance, and it survived because verifying the number felt like verifying the claim. My schema claim survived because checking one class felt like checking the schema. Both are a partial check reported as a complete one.

Fixed in `c1f7e39`. Resolved from `MONDO:0014317` via OLS rather than from your message — both of your numbers check out:

- `OMIM:615715`
- `ORPHA:401764`

I verified them from the MONDO term rather than copying them from the review, which is not a comment on your accuracy but the habit this PR's own history record argues for. On a sibling PR the same check paid: the reviewer-supplied `ORPHA:444463` for `MONDO:0032782` turned out to have no Orphanet cross-reference in MONDO at all, so that entry asserts none.

The `ORPHA:401764` description notes the record is not in `references_cache/` — an identifier to follow, not a source this entry quotes.

`just validate` clean, **59/59 snippets** unchanged, `check-duplicate-keys` OK. Two files.

## Still unused, unchanged and by choice

The rest of your list stands as you left it: the 28% 3-year OS after HSCT for TP53-mutant MDS/AML in `PMID_36952636.md`, a `biochemical:` biomarker for TP53 clonal VAF, and `clinical_trials:`. Also unchanged: the MDS/AML node overlapping the separate AML node, and the AML `FREQUENT` band without a stated denominator. Those last two are real and I would rather flag them as known than half-fix them under a reset approval — the cytopenia denominators you pulled from Table 2 are the right source when someone does that pass.

This resets the approval. Removing a false claim from the record is worth the cycle.

---

matentzn:
Your GeneReviews finding turned up a false statement of mine, so I've pushed `7611c06` to correct it. **History record only — the entry is untouched and still 59/59.**

## The correction

My history record said flatly: *"No GeneReviews chapter exists for this disorder."* That is false as written. No **dedicated** chapter exists, but `NBK571103` covers it — and that file was already sitting in this PR's own `references_cache/` when I wrote the line. Table 1 gives `ERCC6L2` / BMFS2 (OMIM 615715) / AR / 1%–5% of monosomy 7 predisposition / adult onset.

An absolute "does not exist" tells the next curator not to look. **This is the third absence I recorded in this run that the cache contradicted** — the others were a grep truncated by `head -5` and an ASCII search for a full-width numeral. Same error, third disguise.

## But the chapter is not quotable, and that is worth recording

You suggested it as a citable source. I tried it: added the microcephaly sentence as an evidence item, ran `just validate`, and got

```
[ERROR] Text part not found as substring:
'Microcephaly is seen in Fanconi anemia, ligase 4 syndrome, and ERCC6L2-related bone marrow failure syndrome'
```

That cache is stored as **raw HTML**, and every `ERCC6L2` mention sits inside `<i>` tags mid-sentence. The validator does not strip tags, so no sentence naming the gene survives as an exact substring. I removed the item rather than ship an unverified snippet.

Worth naming the symmetry: the chapter is quotable **on screen** and not **in the bytes**. That is the same rendered-form-versus-actual-bytes mistake as the full-width `Ⅱ` on #10231, made in the other direction — I searched ASCII for a rendered glyph; here the rendered text reads clean while the bytes carry markup. Both times the fix was to look at the bytes.

It also means a whole class of cached source is currently uncitable: any `url:`-prefixed cache holding raw HTML, which includes every GeneReviews chapter fetched this way. I have **not** filed that as an issue yet — it needs a duplicate check first, and I would rather check than add a fourth "this doesn't exist" claim to my record today.

## Your other suggestions

**The 1%–5% / adult-onset table row** — the substance of your suggestion is real and I would like it in, but it is in the same HTML table, so it has the same problem.

**Everything else stays open and unchanged**, as flagged last round: the MDS/AML node overlap, the AML band's missing denominator, the Table 2 cytopenia quotes, the 28% 3-year post-HSCT OS figure, TP53 clonal VAF as a `biochemical:` marker, and `clinical_trials:`. Real improvements, none of them defects, and I am deliberately not opening a fifth round for them.

Thanks for the independent corroboration of `OMIM:615715` from the cache rather than from the deep-research report — given what that report did with the De Vitis attribution, checking it against a source neither of us added was the right instinct.

`just validate` clean, **59/59 snippets** unchanged, `validate-history` clean.

---

matentzn:
Correcting myself in the comment above, within the hour, which is faster than my usual turnaround today.

I wrote that this means *"a whole class of cached source is currently uncitable: any `url:`-prefixed cache holding raw HTML."* **That is too strong and it is wrong.** I measured it before filing:

| | count |
|---|---:|
| `references_cache/url_*.md` files | 116 |
| …containing raw HTML tags | 77 |
| Evidence items citing a `url:` reference | 846 |
| …carrying a snippet **and passing validation** | **782, across 61 files** |

So `url:` caches are quoted successfully all over the KB. The failure is **selective, not total**: it hits only sentences whose informative noun is wrapped in markup. In GeneReviews that is gene symbols, by house style — which is why the sentences naming `ERCC6L2` are precisely the unquotable ones, while the rest of the chapter quotes fine.

I generalised from one failing sentence to a whole class of source. That is the same inference shape as the schema claim earlier on these PRs — checking one instance and reporting the category — and I would rather flag it here than let it stand two comments up.

The more interesting cost is the one the measurement exposes: curators have clearly been routing around this by picking tag-free sentences. That works, and it quietly biases citation toward the parts of a chapter that do not name genes. Nothing reports *"you avoided the sentence you wanted"* — you hit a mismatch, pick another sentence, and move on.

Filed as #10270 with the measurement, the worked example, and three options. Duplicate-checked first: #10192 and #9832 are the validator stripping *bracketed* spans, a different mechanism in the opposite direction, and #8198 is a GeneReviews cache being unmined rather than unquotable.

Nothing in this changes the PR. `7611c06` remains history-record-only; the entry is untouched at **59/59**.

---

Issue #10061, opened 2026-08-29
Schema-invalid `supports: PARTIAL` on main: #9730 and #10003 merged past each other


cmungall:
Seven entries on `main` carry `supports: PARTIAL`, which #10003 removed from the enum. They fail schema validation outright:

```
$ uv run linkml-validate --schema src/dismech/schema/dismech.yaml --target-class Disease \
    kb/disorders/Brachydactyly_Type_E1.yaml
[ERROR] 'PARTIAL' is not one of ['SUPPORT', 'REFUTE', 'NO_EVIDENCE'] in /prevalence/1/evidence/0/supports
[ERROR] 'PARTIAL' is not one of ['SUPPORT', 'REFUTE', 'NO_EVIDENCE'] in /prevalence/2/evidence/0/supports
```

**Affected: 7 files, 9 evidence items.**

| File | items |
|---|---|
| `Brachydactyly_Type_E1.yaml` | 2 |
| `Cooks_Syndrome.yaml` | 2 |
| `Brachydactyly_Type_A1.yaml` | 1 |
| `Brachydactyly_Type_A2.yaml` | 1 |
| `Brachydactyly_Type_B1.yaml` | 1 |
| `Brachydactyly_Type_C.yaml` | 1 |
| (+1 further file) | 1 |

Counted by parsing YAML and reading the actual `supports` slot, not by grep — a plain `grep 'supports: PARTIAL'` reports ~71 files because the string also appears in curator prose inside `description` and `explanation` blocks. Only these 9 are real enum values.

## How it happened

Textbook concurrent-merge collision, and neither PR did anything wrong:

- **#9730** added seven brachydactyly entries, authored against the enum as it was, with `PARTIAL` values.
- **#10003** retired `PARTIAL` and migrated every occurrence that existed *on its base*.

Both were green against their own base. #9730's entries were invisible to #10003's migration because they did not exist yet, and #10003's enum change was invisible to #9730 because it had not merged yet. The invalid state exists only in the merge result.

This is exactly the failure mode `CLAUDE.md` describes for duplicate YAML keys — "two concurrent curation PRs... merge without a git conflict. Both PRs are green against their own base, and only the post-merge push build on `main` goes red." The remedy there was an **ungated, whole-KB** CI step rather than a changed-files check. The same reasoning applies to any schema-narrowing change: the PRs that can violate the new constraint are precisely the ones that never see it.

## Fix

[...]

Issue #8374, opened 2026-08-12
Evidence items that use the paper's title as the snippet pass every check but record no result


caufieldjh:
Found across three entries while doing the #8085 `influences_mechanisms` backfill (PR #8372).

## The pattern

Some evidence items use **the cited paper's title as the snippet**:

```yaml
# kb/disorders/Graves_Disease.yaml
- reference: PMID:11453946
  supports: PARTIAL
  snippet: Stressful life events and Graves' disease revisited.
  explanation: This reference discusses the relationship between stressful life
    events and Graves' disease...

# kb/disorders/Multiple_Sclerosis.yaml
- reference: PMID:22906614
  supports: SUPPORT
  snippet: "Risk factors for multiple sclerosis: decreased vitamin D level and remote
    Epstein-Barr virus infection in the pre-clinical phase of multiple sclerosis."
  explanation: The title directly states that decreased vitamin D levels are a risk factor...
```

Others: `Multiple_Sclerosis` EBV (`PMID:35145009`), `Graves_Disease` stress (`PMID:9972676`).

## Why it is worth a rule

**It passes every check we have.** The snippet is exact text from the real reference, over the five-word minimum, correctly attributed. `count-verified-snippets` verifies it, because a title *is* in the cached file.

But a title records **that a question was examined, not what was found**. `PMID:22906614`'s explanation even says so out loud — *"The title directly states…"*. A title asserts the paper's conclusion in the author's most compressed, least-qualified form, stripped of effect size, direction, population and hedging. It is also precisely the shape `.claude/skills/dismech-references/SKILL.md` lists as a fabrication red flag — *"Snippet is suspiciously perfect match to the claim"* — even though nothing here is fabricated.

The failure mode is **overstatement, not fabrication**, and it is invisible for the same reason the truncated-fragment and wrong-endpoint problems were (#8168, #8301, #8350): the text checks out, so nothing looks at what the text is *of*.

## Two honest uses, for contrast

Not every title snippet is bad, and a rule should not be blanket:

- `PMID:9972676`, *"Chronic recurrent stress due to panic disorder does not precipitate Graves' disease"* — the title **states its own negative result**. Genuinely informative, and I carried it on a link in #8372 for exactly that reason.
- A title is sometimes all a cached record has, when only metadata was retrieved.

## Suggested work

1. A short note in CLAUDE.md's evidence SOP: prefer an abstract sentence; use a title only when it states a result rather than a topic, and say so in the explanation.
[...]

PR #10234, opened 2026-08-31
Re-cut 43 line-fragment snippets; move leaked review prose to history (#10061)

caufieldjh:
Two defects found while triaging the `PARTIAL` residue for #10061. Neither is caught by any existing gate.

## 1. Snippets cut at the line wrap, not at a sentence

Every snippet in `Hyperprolinemia_Type_2.yaml` had been snipped at the cached abstract's **line break**, and the same defect appears in four `Cervical_Dystonia.yaml` items. 43 of 76 were *exactly* one wrapped line of the reference cache — against a **1.7% background rate** (261/15,608) measured over a 250-file sample, so this entry is a ~33× outlier rather than the norm.

> **Corrected after review (`8adfe50`).** The first version of this PR used the whole-line fingerprint, which is blind to fragments that are a *portion* of one line. 34 such fragments survived, including two listed in the table below — which I had described as fixed when they were not. The criterion is now **start and end at a sentence boundary**, and all 34 are repaired. See [the review response](https://github.com/monarch-initiative/dismech/pull/10234#issuecomment-5482139819).

They all validate, because a wrapped line *is* a verbatim substring. Most of them assert nothing:

| snippet as curated | what got cut off |
|---|---|
| `"Mutation of Ser352 to Leu is shown to"` | `abolish catalytic activity and eliminate NAD binding` |
| `"Each type is caused by an autosomal recessive inborn error of"` | `the proline metabolic pathway` |
| `"intellectual disability, autism spectrum disorders, and psychosis spectrum"` | `disorders` — a bare noun list cut mid-phrase, carried on **two** different phenotypes |

Same family as the title-snippet rule (#8374) and the methods-sentence problem (#8682): structurally valid, substantively empty, invisible to every check we have.

Each is re-cut to the sentence carrying its own claim, taken verbatim from the same cached reference and asserted to be a substring **before** writing. Where the naive sentence boundary pulled in cache-header debris — an author affiliation, a `(#)Contributed equally` line, a `© 2014 The Authors.` line — it was trimmed by hand.

**Two evidence items are merged into one.** They quoted the two halves of a single sentence split at the line wrap, the second literally beginning `"was found; that is,"`. They were never two pieces of evidence. The survivor quotes the sentence whole and absorbs the dropped item's point.

## 2. Curation review-narrative rendering as clinical content

`Cervical_Dystonia.yaml`'s Botulinum Toxin treatment carried two paragraphs of agent self-correction in its `notes:`:

> *A correction to an earlier revision of this note, which stated two things that are false. It claimed the ChemicalEntityTerm enum is CHEBI-rooted; it is not… a plausible-sounding reason is worse than no reason because nobody re-tests it.*

All true and worth recording — but `notes:` renders on the public disorder page as clinical content *about the treatment*, so the enum roots read as something a clinician should know about botulinum toxin. Moved to a history record. The first paragraph is genuine clinical content and stays. The `CHEBI:3160` binding itself is unchanged.

## 3. `PARTIAL` prose in both files

Nine explanations plus one `notes` block still argued for the retired grade; rewritten to state the same caveat in terms of the current enum, following the pattern #10120 set.

**Five items gain `directness: INDIRECT`** where the caveat *is* an inference step — four where a systematic review pools hyperprolinemia types I and II, one where a tremor figure's denominator is all adult-onset dystonia rather than cervical dystonia.

**Direction was reviewed and deliberately left alone** on the three items whose prose says they are "carried as a limit rather than as support" (absent BtA dose-response; six of 24 poor DBS outcomes; the null biochemical–clinical correlation). None contradicts the claim it attaches to, all three bound how far it reaches, and each sits beside a sibling carrying the positive finding. The two genuine flips in #10120 were found by reading the snippet against the **target node's stated direction**, and these pass that test.

No `supports`, `reference`, `evidence_source` or ontology term changed. The only deletion is the merged half-sentence item.

## Validation

- `just validate-disorders` over both files — **97/97 snippets verified**
- `just validate-history-all` — 7,768 records, no issues
- `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-empty-snippets` — clean
- `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-reference-titles` — grandfathered against `origin/main`, no new violations
- Every snippet in both entries now starts and ends at a sentence boundary of its cached reference, bar 12 checked by hand as splitter artefacts and one that uses the sanctioned ellipsis form across a stripped bracketed span

## Worth a separate issue

The line-fragment pattern is KB-wide at ~1.7%, with a long tail of concentrated entries (in the 250-file sample: `Angelman_Syndrome` 15, `EGFR_Mutant_NSCLC` 12, `Wiskott_Aldrich_Syndrome` 11, `Aicardi_Goutieres_Syndrome` 11). The whole-line fingerprint is a good cheap *triage* signal for locating infected entries (it found this one, at 57% against a 1.7% background), but it is **not** the gate — as this PR's own review demonstrated, it misses sub-line fragments. A gate should use the sentence-boundary test instead. Happy to file it if useful; not doing so unprompted.

### DIFF
--- a/history/disorders/Cervical_Dystonia/2026-08-31T173838Z-claude-code-16100d.yaml
+++ b/history/disorders/Cervical_Dystonia/2026-08-31T173838Z-claude-code-16100d.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Cervical_Dystonia
+  path: kb/disorders/Cervical_Dystonia.yaml
+session:
+  id: 2026-08-31T173838Z-claude-code-16100d
+  timestamp: '2026-08-31T17:38:38Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10061
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10234
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - evidence
+  summary: 'Review response: extend four truncated snippets, one of which hid a null
+    result'
+  details: |
+    Response to the blocking item in the ai4c-reviewer review of PR 10234, which found sub-line snippet fragments my whole-line detector was blind to. Four in this entry, and one of them matters more than the others. The neck-tremor phenotype item quoted the sex-comparison sentence up to 'tremors in the neck and/or in the upper limb tended to be more frequent in women than in men' and stopped at the comma immediately before ', but the difference did not reach the level of statistical significance after Bonferroni correction'. The cut turned a non-significant trend into what reads as a finding, on the item this PR had just edited and graded INDIRECT; the quote now runs to the end of the sentence and carries its own correction. The botulinum toxin dose-response item stopped at 'cervical dystonia-specific impairment' and now continues through 'a destinction between BtA formulations, or a variation with use of EMG-guidance' (the misspelling is the source's), which also makes the item quote the two findings the treatment's notes block records as frequently assumed in practice. The isolated-dystonia prevalence item began mid-sentence at 'was estimated as' and now starts at 'According to recent epidemiological studies'. The sex-ratio prevalence item was extended to its full sentence, with the abstract's 'RESULTS:' section label deliberately left out of the quote rather than pulled in as though it were part of the sentence. Five further flags in this entry were checked by hand and left alone: they are already whole sentences and the sentence splitter only joins them to the cache file's markdown header or to a results table whose rows are the quotable unit. No supports, directness, reference, evidence_source or ontology term changed in this pass. Validated with just validate-disorders over both entries (97/97 snippets verified), check-duplicate-keys, check-folded-hyphens, and check-snippet-length and check-title-snippets grandfathered against origin/main.
--- a/history/disorders/Hyperprolinemia_Type_2/2026-08-31T173823Z-claude-code-f252e1.yaml
+++ b/history/disorders/Hyperprolinemia_Type_2/2026-08-31T173823Z-claude-code-f252e1.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hyperprolinemia_Type_2
+  path: kb/disorders/Hyperprolinemia_Type_2.yaml
+session:
+  id: 2026-08-31T173823Z-claude-code-f252e1
+  timestamp: '2026-08-31T17:38:23Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10061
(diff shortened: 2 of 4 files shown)


caufieldjh:
Fixed in `8adfe50`. The blocking finding was correct, and the gap was in my **detector**, not just in the prose — worth stating plainly because it also invalidates a claim in my PR body.

## What I got wrong

My fingerprint tested whether a snippet was *exactly* one wrapped line of the cached file. That caught 43 whole-line fragments and was structurally blind to fragments that are a **portion** of one line — the same defect, one character short of matching. So "0 of 75" was true and meant less than I implied.

Worse, and this is the part I want to own: two of the three snippets in my PR body's table were listed as **examples of work I had done**, and they were still cut. I pulled those examples from an earlier heuristic audit rather than from the list I actually applied, and never checked that the two sets agreed. The PR body has been corrected.

The sharpest evidence that the detector was the problem: `"The accumulating P5C spontaneously reacts"` sat cut in one block of `Hyperprolinemia_Type_2.yaml` while the *same sentence* had been repaired in full in another block of the same file.

## What changed

The test is now the stricter one it should have been: **every snippet must both start and end at a sentence boundary of its cached reference.** 34 more repaired — 30 in `Hyperprolinemia_Type_2`, 4 in `Cervical_Dystonia` — each verified as an exact substring of the same cache before writing.

**One was substantive, not cosmetic.** Your `Cervical_Dystonia.yaml:270` catch is the most important finding in the review. The quote stopped at the comma immediately before *", but the difference did not reach the level of statistical significance after Bonferroni correction"* — turning a non-significant trend into what reads as a finding, on an item this PR had just edited and graded `INDIRECT`. Now runs to the end of the sentence and carries its own correction.

## Judgement calls, flagged for you rather than buried

- **12 remaining flags left alone.** Checked by hand; each is already a whole sentence, and the naive sentence splitter only joins it to the cache file's markdown header or to a results table whose rows are the quotable unit. Not defects.
- **One sentence cannot be quoted whole.** `"We found four mutant alleles, two with frameshift mutations [A7fs(-1) and G521fs(+1)] and two with missense mutations (S352L and P16L)."` — the bracketed span is stripped from the snippet but not from the cached text, so the full sentence fails to match. Uses the sanctioned discontinuous-quote form with an ellipsis across the bracket.
- **`RESULTS:`** was deliberately left outside the quote on the sex-ratio prevalence item rather than pulled in as though it were part of the sentence.

## On the proposed KB-wide gate

You're right that the blind spot would have shipped into it. If that check gets built it should use the sentence-boundary test, not the whole-line one — the whole-line fingerprint is a good *cheap triage signal* for finding infected entries (it's what located this one, at 57% against a 1.7% background), but it is not the gate.

No `supports`, `directness`, `reference`, `evidence_source` or ontology term changed in this pass.

**Validation:** `just validate-disorders` over both entries (97/97 snippets verified), `just validate-history-all` (7,770 records), `check-duplicate-keys`, `check-folded-hyphens`, and `check-snippet-length` / `check-title-snippets` grandfathered against `origin/main`. Two new history records document the pass.

Thanks for the endorsement on the `supports` calls and the merge — that was the part I was least sure of.

PR #10247, opened 2026-08-31
curate: Ageing Associated Decline in Intrinsic Capacity (ICD-11 MG2A)

cmungall:
## Summary

Adds a new disorder entry for **ageing associated decline in intrinsic capacity** (ICD-11 MG2A) — the WHO construct of the composite of physical and mental capacities a person can draw on, and its age-related loss across the five ICOPE domains. The entry treats the condition as the clinical convergence point of the hallmarks of ageing, so its mechanism chain hands off to the hallmark modules already in `kb/modules/` rather than restating them.

## No MONDO anchor

MONDO carries no class for intrinsic capacity or its decline (searched via OLS on 2026-08-31 for "intrinsic capacity", "ageing associated decline", "frailty" and "sarcopenia" — all empty in MONDO). The entry therefore has no `disease_term` and is anchored on the ICD-11 Foundation entity `icd11f:835503193`, whose label matches the entry name exactly and which resolves under `icd11f:455013390`, so it satisfies the `ICD11FTerm` dynamic enum. Twenty-six existing entries likewise carry no MONDO anchor, so this is an accepted state rather than a defect.

Whether to request a MONDO class is left open as a `CURATION_TODO` discussion in the entry. There is a real argument on both sides: MG2A is classified by WHO under symptoms and clinical findings rather than as a disease, and this project does not mint terms — but without a MONDO class the entry sits outside every MONDO-keyed query and coverage report in the repository.

## What is in the entry

`kb/disorders/Ageing_Associated_Decline_in_Intrinsic_Capacity.yaml`, 1,753 lines:

- **Five ICOPE domains as `has_subtypes`** (locomotor, vitality, cognitive, psychological, sensory). They are not aetiological subtypes; they are the axes the construct is measured on, and impairment in them is genuinely dissociable — someone can have isolated locomotor decline with intact cognition. Modelling them as subtypes gives the phenotypes, prevalence and progression records a foreign key for domain-specific claims.
- **A 13-node pathophysiology chain**: hallmark ageing damage → mitochondrial bioenergetic decline, senescent-cell accumulation, loss of regenerative reserve → inflammaging → loss of physiological reserve across organ systems → the five capacity domains → the composite construct → functional ability loss and care dependence. Five nodes declare `conforms_to` against the existing `cellular_senescence`, `mitochondrial_dysfunction`, `inflammaging` and `stem_cell_exhaustion` modules.
- **9 phenotypes** bound to HPO, keyed to their domains, with frequencies taken from one Hong Kong community cohort that reports all six domain-wise impairment rates.
- **4 definitions**, including the ICOPE Step 1 screening instrument as a `PHENOTYPE_ALGORITHM` with a `validation_status` block.
- **6 treatments**, **5 registered trials**, **2 environmental exposures** with pathograph links, and **2 plasma biomarkers**.
- **82 evidence items across 34 distinct references** (29 PMIDs, 5 ClinicalTrials.gov records).

## The causal claim is weaker than the chain looks

Everything upstream of "Loss of Physiological Reserve Across Organ Systems" is generic geroscience that was not written about intrinsic capacity, and the human evidence joining the two is associative. The strongest link available is that plasma IL-6, TNFR-1 and GDF-15 separate multi-impaired from high-stable capacity trajectories in 1,271 MAPT participants (PMID:37620614). No human study has shown that reducing hallmark burden raises measured intrinsic capacity.

Those edges are marked `causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES` and the question is filed as an open `KNOWLEDGE_GAP` rather than asserted. It is worth flagging explicitly because it is exactly the assumption that geroscience intervention programmes make when they adopt intrinsic capacity as an early endpoint in place of decades-long disease outcomes — ARPA-H's PROSPR programme among them, which is named in the entry notes as orientation and deliberately not curated as evidence.

## Negative and conflicting results were kept

- The ACHIEVE hearing trial's **null** primary result is a `REFUTE` evidence item on the hearing treatment. The observational sensory-to-cognitive link is not a basis for prescribing hearing aids as a cognitive intervention, and the entry says so.
- Two validations of the ICOPE screening tool reach compatible numbers and incompatible conclusions — INSPIRE-T recommends it, the Toledo Study found the cognitive items misbehaving and the composite adding nothing over its parts. That is recorded as a `CONTROVERSY` discussion and as a `REFUTE` item inside the definition's `validation_status`.
- The anorexia-of-ageing review's statement that no approved pharmacotherapy exists is cited on the vitality phenotype, and is why there is no vitality drug in `treatments`.

## Deep research provider fell back

`falcon` was requested and returned HTTP 402 — the Edison account is out of credits. The run was re-issued with `--fallback` and `claude_code` produced the report, which is committed with `fell_back`, `requested_provider` and `provider_attempts` in its frontmatter.

**The report's own validation reported `needs_review: true`, with 0 of 3 checked quotes matching their sources.** No quoted material was taken from it. Every snippet in this entry was cut from a locally cached abstract instead. Its term suggestions were treated as leads only: `GO:0007568` and `GO:0055114` are flagged obsolete and were not used, `HP:0031466` was mislabelled in the report and was not used, and `HP:0000716` is labelled *Depression* in current HPO rather than *Depressivity* as the report wrote. The report was used for structure and for surfacing references.

## Validation

All checks pass on this entry:

| Check | Result |
|---|---|
| `just validate` (schema, terms, references) | passes; **82/82 snippets verified** |
| `just validate-disorders` | passes |
| `check-entity-refs`, `check-duplicate-keys` | pass |
| `check-environmental-evidence` | passes |
| `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` | pass, no new baseline entries |
| `check-term-cache-integrity`, `check-cache-order` | pass |
| pytest (conformance, entity-ref, subtype, unique-name, required-field) | 15,527 passed |

No DOI-keyed evidence is used, so every evidence item is snippet-checked by the gating validator.

Two snippets were changed during validation rather than kept as first written: a PMID:38945130 quote with nested bracketed confidence intervals could not be matched by the reference validator's bracket-stripping and was replaced with the paper's interpretation sentence plus a separate sample-size quote, and the `reference_title` for PMID:41360071 was corrected to match the cached record.

## Other files

- `research/Ageing_Associated_Decline_in_Intrinsic_Capacity-deep-research-claude_code.md` and its citations sidecar.
- 70 new `references_cache/` files. More than the 34 the entry cites, because the deep-research run resolved its own citations into the same cache.
- `history/disorders/Ageing_Associated_Decline_in_Intrinsic_Capacity/` — one CREATE record.
- `cache/icd11f/terms.csv` and the `icd11fterm` enum cache, both extended with the new ICD-11 entity by the validator.


cmungall:
Thanks — all three blocking findings are addressed in 31255aedc, along with five of the eight suggestions. `just validate` now verifies 113/113 snippets, up from 82/82.

**One correction first: the PMIDs in the review don't match the papers named.** The pairings appear rotated. Every paper the review named does exist in cache, under a different ID:

| Review said | Actually that ID | The named paper is |
|---|---|---|
| PMID:41113460 — air pollution and frailty | mitochondrial dysfunction in age-related sarcopenia | PMID:40391839 |
| PMID:40060276 — Sydney MAS, IC to dementia/mortality | biological rationale for integrating IC into frailty models | PMID:41086232 |
| PMID:40666427 — IC and 5-year CVD mortality | targeting the hallmarks of aging | PMID:40927083 |
| PMID:40391839 — biological rationale for IC in frailty models | air pollution and frailty | PMID:40060276 |
| PMID:36341237 — ENHANCE RCT | a non-randomised prefrailty-reversal study | PMID:40464147 |

The GWAS ID (PMID:41066301) was correct. I used the correct IDs throughout; flagging it because a curator copying the suggested `[Fix this →]` links would have cited the wrong papers.

**Finding 1 — genetics.** Added `genetic:` with two entries and `inheritance:` bound to `HP:0010982`. The polygenic entry deliberately carries no `gene_term`: the GWAS abstract maps 4,289 candidate SNPs to 197 genes and names none of them individually, so a lead gene would have had to come from the deep-research report's prose rather than a quotable source. APOE (`hgnc:613`) is a separate `MODIFIER` entry. The tissue-enrichment sentence is also cited, because it independently recovers the same organs the domain nodes are anchored in.

**Finding 2 — air pollution.** Added, bound to `ECTO:8000036`, with a `PREDISPOSES` edge onto the reserve node. Graded `directness: INDIRECT`, and the `notes` say why: the meta-analysis measures **frailty**, not intrinsic capacity, and this entry models frailty as downstream of reserve loss rather than as the same construct. I did not add physical inactivity or social participation — the only support in cache is a list of correlates inside a prevalence meta-analysis, which can't carry an exposure claim.

**Finding 3 — reference consumption.** Cited 17 more across `progression:`, `clinical_burden:`, `pathophysiology:`, `treatments:`, `animal_models:`, `diagnosis:` and `biochemical:`, and added a sixth numbered paragraph to `notes:` naming the ten still-uncited references with a reason each. That paragraph also records why the cache is larger than the citation list: the deep-research run resolved its own citations into the same directory, so the cache reflects what the *report* cited, not what a curator selected.

PMID:40060276 was the most valuable of the set, and it improved something the review didn't flag. The vitality node's placement upstream of the other four domains was hedged as "a reading of the definition and not a demonstrated causal sequence". That paper argues the hierarchical, energy-dependent ordering directly, so the hedge is now a citation.

**Suggestions taken:** UBERON bindings on all four free-text `locations:`; an `animal_models:` entry for the naturally ageing mouse Frailty Index, `PARTIALLY_RECAPITULATES` against the reserve node, with `limitations` recording that three of the five IC domains have no counterpart in that index; a `diagnosis:` block for the ICOPE two-step assessment; and a third `biochemical:` entry whose content is a negative — no inflammatory marker has adequate specificity, recorded with a `REFUTE` item so the two markers above aren't read as a monitoring panel.

**Three suggestions I did not take:**

- **HP:0031936 for "slow walking" is the wrong term.** Current HPO labels it *Delayed ability to walk* — a developmental phenotype, not age-related gait slowing. HPO appears to have no term for reduced gait speed in adults (searched "slow gait", "reduced walking speed"), so the locomotor domain keeps HP:0001288 *Gait disturbance*. Worth noting this came through the deep-research report, which is the same source that supplied the two obsolete GO terms.
- **HP:0012378 (Fatigue)** — no cached reference states fatigue as a vitality-domain phenotype in a quotable sentence. The vitality domain does now have HP-bound phenotypes (Malnutrition, Weight loss).
- **The DNA-methylation IC clock** has no cached quotable source, so no biochemical entry was written for it.

Validation: `just validate` passes all three stages, 113/113 snippets; `check-entity-refs`, `check-duplicate-keys`, `check-environmental-evidence`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-term-cache-integrity`, `check-cache-order` all pass; 31,106 targeted pytest cases pass.

Sixteen `references_cache/` files show as modified because the reference validator upgraded them from `abstract_only` to `full_text_xml` during this run. None were hand-edited.

---

cmungall:
Thanks for the approval, and for the correction to the PMID table — agreed on all of it.

I've taken the three deferred nits anyway, in 590b713b2. Reasoning: nit 3 is a factual error in prose I wrote, and leaving a known-wrong statement in the entry because fixing it costs a review cycle is the wrong trade. The other two are small and verified, so batching them costs one cycle rather than three. This does dismiss the approval — sorry for the churn.

**Nit 3 (the real one).** `notes:` called PMID:35775483 a "general hallmarks-of-ageing review". It is specifically about mitochondrial dysfunction in cell senescence and ageing. Corrected, and the entry now gives the actual reason it was set aside: the reciprocal mitochondria-senescence feedback loop it describes is delegated to the two modules rather than re-derived here.

**Nit 1.** You were right that the grade and the prose disagreed — an item opening "A caution rather than a confirmation" should not be graded `SUPPORT` alone. Split per the CLAUDE.md pattern: the `SUPPORT` item keeps the claim it actually supports, and two `REFUTE` items now carry the paper's negatives — that intervention exposure had no effect on prefrailty reversal once baseline capacity was accounted for, and the authors' own reading that a two-domain intervention may not move the composite. That second one is pointed for this entry: it is an argument for the domain-specific structure the entry models rather than for a single global target. The non-randomised design and self-selected control group are stated in the explanation so the null isn't read as overturning SPRINTT.

**Nit 2.** Both caveat-shaped items removed as evidence and moved to slots that carry the weight without a direction grade — `treatments[].notes` for the ENHANCE baseline imbalance (keeping the figures) and the existing `modeled_mechanisms[].limitations` for the mouse Frailty Index, which now also records observer and environment sensitivity as the reason fidelity is `MODERATE`.

Snippet count nets back to 113/113 (two items removed, two added). All offline gates pass, including `check-snippet-grading` with the new `SUPPORT`-and-`REFUTE` pair on PMID:36341237 — expected, since that gate keys on `evidence_source` rather than `supports`, and the three items quote different sentences. Targeted pytest: 31,086 passed.

Nothing else outstanding from my side.

---

PR #10248, opened 2026-08-31
Add Blue Rubber Bleb Nevus Syndrome entry (MONDO:0007203)

justaddcoffee:
Curates **blue rubber bleb nevus syndrome** (Bean syndrome, `MONDO:0007203`), a somatic mosaic multifocal venous malformation disorder.

## Why this target

No entry, no curation issue, and `MONDO:0007203` appears nowhere in `kb/`. Checked against current `main` (2,489 entries) immediately before branching — the KB has been absorbing new entries within hours, so a stale check is worthless. The two adjacent vascular entries are **arteriovenous** malformations, a different lesion class.

## Model

A single causal chain with a branch point, which is what makes the therapeutics make sense:

```
somatic TEK double (cis) mutation
        ↓
ligand-independent TIE2 activation
        ↓
PI3K-AKT-mTOR activation ──┬─→ suppressed endothelial apoptosis ──┐
                           └─→ AKT/FOXO1 → PDGFB ↓ → mural cell   ├─→ ectatic mural-cell-poor
                                                   deficiency ────┘      venous channels
                                                                              ↓
                                                              chronic GI blood loss
```

7 nodes, 7 edges, **every edge evidenced**. Sirolimus is modelled as `INHIBITS` on the **pathway node** rather than on a lesion — which is the mechanistic reason it works across sites and lesion types.

Includes `prevalence`, `inheritance` (somatic mosaic), phenotypes, 3 genes with `gene_term` bindings, `target_mechanisms` on treatments, and an animal model with `modeled_mechanisms`.

## Verification

All 8 cited PMIDs independently re-fetched, **every `reference_title` read from the cache frontmatter**, and all **29 snippets** verified as exact substrings. Every CURIE resolved against the local OAK databases with its label read back before use.

> **Retracted.** This originally read "all fifteen CURIEs this research report suggested resolved correctly." Resolution is not correctness. `GO:0038084` resolves perfectly and is the **VEGF** pathway, but TIE2 is the **angiopoietin** receptor — the report glossed it as "VEGF/angiopoietin-receptor signaling," which is wrong, and it propagated into the entry. The report also suggested `HP:0025439` for GI vascular malformation; that ID is **Pharyngitis**. Both would pass an existence check. Corrected in the entry and recorded here.

Other issues caught, which were mine or schema drift:

| Issue | Resolution |
|---|---|
| `HP:0001428` Somatic mutation is **obsolete** | → `HP:0001442` Typified by somatic mosaicism |
| `supports: PARTIAL` no longer in the enum | → `SUPPORT`, with the partial-response caveat stated in the `explanation` |
| Two snippets written as **paper titles** rather than sentences | → replaced with real abstract sentences |
| One `reference_title` singular where cache is plural | → corrected to match cache |

## A naming trap worth flagging

Despite "nevus" in the eponym, **these are venous malformations, not nevi and not tumours** — ISSVA classifies BRBNS among multifocal venous malformations. `HP:0100814 Blue nevus` matches the name but is *melanocytic*, so it would have been wrong.

> **Retracted.** This section originally claimed HPO had no venous-malformation term. That was false — `HP:0012721 Venous malformation` exists and is already used by three sibling entries. The claim came from a substring search run with `LIMIT 3`, where three *arteriovenous* hits filled the limit and hid the correct term at position six: absence asserted from a truncated result. The entry now binds `HP:0012721`.

## Validation

```
just validate      # 29/29 snippets verified
just compliance    # 98.8% global / 100.0% weighted
```

`check-reference-titles`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens` — **all zero findings**. History record included and validates.

## Known gaps, in notes

`datasets`; `clinical_trials` (NCT00975819); **localised intravascular coagulopathy** with elevated D-dimer, described in the research as both diagnostic biomarker and complication but not fetched to a quotable source; endoscopic and surgical treatment entries; alpelisib and rapamycin+alpelisib as emerging targeted therapy; and the zebrafish and genetic-mosaic TIE2 p.L914F mouse models.

### DIFF
--- a/kb/disorders/Blue_Rubber_Bleb_Nevus_Syndrome.yaml
+++ b/kb/disorders/Blue_Rubber_Bleb_Nevus_Syndrome.yaml
@@ -188,7 +188,7 @@ pathophysiology:
       reference_title: AKT/FOXO1 axis links cross-talking of endothelial cell and pericyte
         in TIE2-mutated venous malformations.
       supports: SUPPORT
-      evidence_source: HUMAN_CLINICAL
+      evidence_source: IN_VITRO
       snippet: RAPA ameliorated the physiological function of mutant ECs by inhibiting AKT-mTOR
         pathway, but also facilitated the nuclear location of FOXO1 and the expression of PDGFB
       explanation: Directly evidences the AKT/FOXO1/PDGFB axis this edge asserts, since inhibiting
@@ -468,8 +468,8 @@ phenotypes:
   phenotype_term:
     preferred_term: elevated D-dimer
     term:
-      id: HP:0003256
-      label: Abnormality of the coagulation cascade
+      id: HP:0033106
+      label: Elevated circulating D-dimer concentration
   evidence:
   - reference: PMID:19917952
     reference_title: Elevated D-dimer level in the differential diagnosis of venous malformations.
@@ -568,10 +568,10 @@ diagnosis:
     the route for treatment. Most of the clinical problem in this disease is
     finding occult small-bowel lesions.
   diagnosis_term:
-    preferred_term: endoscopic biopsy
+    preferred_term: endoscopic procedure
     term:
-      id: NCIT:C15389
-      label: Endoscopic Biopsy
+      id: NCIT:C16546
+      label: Endoscopic Procedure
   evidence:
   - reference: PMID:40074320
     reference_title: 'Blue rubber bleb nevus syndrome: A case report and literature review.'
@@ -587,6 +587,7 @@ treatments:
     inhibits mTOR, the node downstream of the constitutively active TIE2-PI3K-AKT
     axis. In a prospective study of BRBNS it reduced lesion size and raised
     haemoglobin, with only grade 1-2 adverse effects.
+  therapeutic_modality: SMALL_MOLECULE
   treatment_term:
     preferred_term: pharmacotherapy
     term:
@@ -676,7 +677,7 @@ treatments:
       label: Endoscopic Mucosal Resection
   target_mechanisms:
   - target: Chronic Gastrointestinal Blood Loss
-    treatment_effect: BYPASSES
+    treatment_effect: INHIBITS
     description: >-
       Obliterating or removing accessible lesions removes their contribution to blood
       loss without altering the underlying signalling defect.
@@ -712,7 +713,7 @@ treatments:
       label: Surgical Procedure
   target_mechanisms:
   - target: Chronic Gastrointestinal Blood Loss
-    treatment_effect: BYPASSES
+    treatment_effect: INHIBITS
     description: >-
       Removing the bowel segment removes the bleeding lesions it carries.
     evidence:
@@ -734,7 +735,7 @@ treatments:
       severe chronic anemia
     explanation: The multifocal bleeding lesion burden is what resection is directed at.
 animal_models:
-- name: Ponatinib and rapamycin murine venous malformation model
+- name: HUVEC-TIE2-L914F xenograft
   species: Mouse (Mus musculus)
   genotype: TIE2-mutant human endothelial cell xenograft
   category: Xenograft


justaddcoffee:
Pushed `270c943`. All three critical items and the eight important ones.

## 🔴 The two wrong bindings — both my errors, and one has a root cause worth recording

**1. `HP:0012721 Venous malformation` exists.** You're right, and the failure is more embarrassing than a wrong lookup. My search was:

```sql
... where value like '%enous malformation%' limit 3
```

Three *arteriovenous* hits filled the limit. `Venous malformation` sits at **position six**. I asserted absence from a truncated result, then wrote that assertion into the entry, the notes, the history record and the PR body as though it were a finding.

Rebound to `HP:0012721`, and corrected in all four places. The `HP:0100814 Blue nevus` reasoning was right and is kept.

**2. `GO:0038084` is the VEGF pathway.** TIE2/TEK is the **angiopoietin** receptor — the binding named a different pathway entirely. Rebound to `GO:0048014 Tie signaling pathway`, which `Lymphatic_Malformation_6.yaml` already uses.

Your framing here is the useful part: **"resolution is not correctness."** My PR body claimed all fifteen report CURIEs "resolved correctly," and that was a misleading thing to say — `GO:0038084` resolves perfectly and is wrong for this node. The report's gloss, "VEGF/angiopoietin-receptor signaling," is where it came from. And your note that the report also offered `HP:0025439` for GI vascular malformation — which is **Pharyngitis** — makes the point sharper still. I've corrected that claim in the commit message rather than leaving it standing.

**3. PIK3CA** — `CAUSATIVE` contradicted the entry's own note and its cited paper, which is explicitly about VMs *without* TEK mutation. Now `BIOMARKER`. Its `evidence_source` is now `IN_VITRO`; you're right that the quoted finding ends "when expressed in human umbilical vein endothelial cells."

## 🟡 Gaps filled

- **Localised intravascular coagulopathy** now closes the causal chain at step 8, with `PMID:19917952` (specificity 96.5%) as its marker — plus elevated D-dimer as a phenotype and a coagulation-testing `diagnosis` entry
- **Endoscopic and surgical treatment** added. You were right that with only sirolimus and transfusion the section misrepresented how the disease is managed
- **`diagnosis` section**, **NCT00975819**, the recurrent **T1105N–T1106P** double-*cis* allele, and **GI venous malformations** as a phenotype distinct from the bleeding
- **PDGFB removed from `genetic`** — an expression finding, not a variant, exactly as CLAUDE.md §9 says

## The eleven claim–evidence mismatches

All fixed from sentences already cached. Two needed **the reference swapped as well as the snippet** — the quote and the citation had drifted apart, which `validate-references` caught immediately. Every substitution you listed was available in-cache as you said.

## Validation

`just validate` — **41/41 snippets**, up from 29. Compliance **99.1% global / 100.0% weighted**. All seven content gates zero findings. 8 nodes, 8 edges, every edge evidenced.

Not done, and flagged rather than dropped: phenotype **frequency and onset** data. The research gives 68% / 79.5% / 54.3% and median diagnosis age 12, but I could not trace those to a quotable sentence in a cached abstract, and asserting bands without their own evidence is the failure mode I've been bitten by before. Recorded in `notes` as an extension point.

justaddcoffee:
Pushed `9fb2d69`. All four fixed, plus two of the suggestions, plus the PR-body retractions.

| | Fix |
|---|---|
| **A** | `HP:0003256` → **`HP:0033106`** Elevated circulating D-dimer concentration |
| **B** | RAPA/AKT-FOXO1/PDGFB snippet → **`IN_VITRO`** (the α-SMA IHC item stays `HUMAN_CLINICAL`) |
| **C** | `BYPASSES` → **`INHIBITS`** on both procedural treatments |
| **D** | `NCIT:C15389` Endoscopic Biopsy → **`NCIT:C16546`** Endoscopic Procedure |

**On A — you're right that this is a repeat, and it's worth naming as a habit rather than two slips.** Both times I took the first plausible *ancestor* instead of searching for the exact term, and both times the exact term was already cached and already used by another entry. Last round it was a truncated search; this time I didn't search at all, I just reached for a coagulation ancestor I already knew. Different mechanism, same outcome. I've put that in the commit message so it's in the record rather than only in this thread.

**On D** — agreed, and the point about not casually biopsying a venous malformation is well made; the term was asserting a procedure that would be actively inadvisable.

**Suggestions taken:** `therapeutic_modality: SMALL_MOLECULE` on sirolimus, and the animal model renamed from the drugs tested to **`HUVEC-TIE2-L914F xenograft`**.

## PR body corrected

You flagged that the body still carried both retracted claims, and that it's the version people read from the merge commit. Fixed — the body now carries two explicit `> **Retracted.**` blocks:

- The "HPO has no venous-malformation term" claim, with the `LIMIT 3` root cause
- The "all fifteen CURIEs resolved correctly" claim, with your **resolution is not correctness** framing, `GO:0038084` as the case that separates them, and `HP:0025439` → *Pharyngitis* as the second example

Worth noting `gh pr edit` failed on an unrelated Projects-classic GraphQL deprecation error and silently left the body unchanged — I only caught it by re-reading the body afterwards. Patched via the REST endpoint instead.

## Not done

- **`functional_impact_category`** — I tried it on the `genetic` entry and schema validation rejects it there; it belongs on a `GeneticContext`, which this entry doesn't currently carry. Left as `association` free text rather than restructuring the block late in review.
- **`directness: INDIRECT`**, **`biological_scale`**, **`conforms_to`** — `just list-modules` surfaced no PI3K-AKT-mTOR or angiogenesis module that fits; the nearest are `deregulated_nutrient_sensing#mTORC1 Hyperactivation` and an apical-neuroependyma branch, neither of which is a vascular-anomaly mechanism. Flagging rather than forcing a poor conformance.

## Validation

`just validate` — **41/41 snippets**. Compliance **99.1% global / 100.0% weighted**. All seven content gates zero findings. 8 nodes, 13 evidenced edges.

Issue #8430, opened 2026-08-13
ECTO exposure_term missing on pathograph-linked environmental exposures


caufieldjh:
Three `environmental:` exposures that are now **pathograph-visible** via
`influences_mechanisms` links still have no ECTO `exposure_term` binding, so
they render as free text on a page where the rest of the graph is
ontology-grounded. Raised by the reviewer of PR #8427 as a follow-up.

| Entry | Exposure | Note |
|---|---|---|
| `Raynaud_Disease` | `Emotional Stress` | Psychological-stress exposure; ECTO coverage uncertain |
| `Spaceflight_Associated_Neuro-Ocular_Syndrome` | `Elevated Ambient Carbon Dioxide` | ECTO plausibly has a carbon-dioxide exposure term |
| `Spaceflight_Associated_Neuro-Ocular_Syndrome` | `Microgravity Exposure` | ECTO likely has nothing; see below |

### Why this matters more than it did

Before the #8085 backfill these exposures sat in a section nothing pointed at.
Now each is an edge into a pathophysiology node and shows up on the disorder
page as an input to the mechanism graph, so the lack of a term is more
visible and more worth closing.

### The guardrail

Per the term-precision rule in `.claude/skills/dismech-terms`, **no term beats
a bad one**. `Microgravity Exposure` in particular is unlikely to have an ECTO
representation, and forcing it onto a broad parent (a generic "exposure to
physical object" style class) would assert a grounding that isn't real. The
correct outcome for that one may well be *leave it unbound*, recorded as
checked rather than as overlooked.

Suggested approach: check each against ECTO with OAK, bind where a genuinely
specific term exists, and for the ones with no good term add a note saying the
search was done — so the next curator doesn't repeat it.

### Worth a sweep

This is unlikely to be limited to three exposures. A pass over
`environmental[]` across `kb/disorders/` counting entries with
`influences_mechanisms` but no `exposure_term` would size the gap, and those
are the highest-value ones to bind since they are the ones now rendering in
the graph.

Found during the #8085 environmental pathograph backfill (PR #8427).

caufieldjh:
Adding the three exposures from PR #8429 to this issue's scope, flagged by the same reviewer:

- `Aortitis` — `Granulocyte-colony stimulating factor exposure`
- `Aortitis` — `Current tobacco smoking`
[...]

PR #10250, opened 2026-08-31
Curate autosomal dominant nonsyndromic hearing loss 12 (MONDO:0011102)

### LINKED ISSUE #10237: Curate autosomal dominant nonsyndromic hearing loss 12 (MONDO:0011102)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 12** ([MONDO:0011102](https://monarchinitiative.org/MONDO:0011102)) — DFNA12, caused by dominant `TECTA` variants in alpha-tectorin, a core non-collagenous protein of the tectorial membrane. Missense variants act on the tectorial membrane's mechanical coupling to the outer hair cell stereocilia, and the affected TECTA domain tracks with whether the audiogram is mid-frequency or high-frequency.

Preflight notes: MONDO leaf, single causal gene `hgnc:11720 TECTA`. On `origin/main`, `TECTA` appears only in the two recessive `TECTA` entries (`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93` and `_98`), which are a different allelic mechanism — recessive loss of function producing severe congenital deafness, against the dominant, largely non-truncating, milder DFNA12. The MONDO ID appears nowhere and no open PR covers it. The KB already carries per-locus DFNA entries (DFNA2B, 3A, 7, 11, 17, 41), so this fits convention.

Note DFNA8 and DFNA12 are the same locus and are usually written DFNA8/12; worth recording in `synonyms` and deciding whether the recessive `TECTA` entries deserve a cross-reference.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12.yaml`

aleixpuigb:
Adds a new disorder entry for DFNA12 — the dominant nonsyndromic hearing loss caused by TECTA variants in alpha-tectorin. The literature writes it DFNA8/12, because DFNA8 and DFNA12 were mapped independently in an Austrian and a Belgian family and both turned out to be the same gene.

This one is unusual among the deafnesses in that the broken part is not a cell. The tectorial membrane is an acellular gel that sits over the hair cells and acts as the load they push against; mutant alpha-tectorin is secreted normally and built into it, where it spoils the assembly of the surrounding normal protein. The hair cells, spiral ganglion and auditory nerve are all structurally intact, which is why patients can have no otoacoustic emissions at all — a completely failed cochlear amplifier — and still score 95% on word recognition.

## Pathograph

```
TECTA missense variant (heterozygous)
        ↓
defective tectorial membrane matrix assembly
        ↓
impaired mechanical coupling to outer hair cell stereocilia
        ↓
cochlear amplifier failure
        ↓
bilateral sensorineural hearing loss

        ↘ secondary outer hair cell vulnerability (progressive genotypes)
```

The dominant-negative mechanism rests on a clean natural comparison rather than on assertion: TECTA truncating alleles cause a *recessive* deafness (DFNB21), and their heterozygous carriers hear normally. Losing one copy is therefore not enough, so the dominant alleles must be doing something active.

**Three allele-matched mouse models, two of which record negative results.** The mice were built to carry specific human alleles, and two of them carry `direction: UNCHANGED` readouts because that is what was measured:

- the C1619S mouse does **not** progress, although the French family carrying the same substitution does;
- the C1509G mouse has stable emission thresholds in quiet to six months, and its excess hair cell death appears only after experimental noise exposure.

These ground a `HUMAN_MODEL_MISMATCH` discussion. I deliberately left the question open rather than resolving it: mouse lifespan may be too short for a process that takes human decades, or human progression may need cumulative acoustic exposure that laboratory housing does not supply — in which case the noise-exposure result is the *more* faithful model, not the less. Either way, there is currently no model in which this disease progresses spontaneously, which matters for anyone designing a trial endpoint.

**The domain-to-audiogram correlation is weaker than its reputation.** ZP domain → stable mid-frequency, zonadhesin domain → progressive high-frequency is the textbook version, and the mouse models do confirm the two domains damage the matrix in structurally different ways. But unbiased cohort screening later found mid-frequency loss arising from entactin and vWFD variants too. The entry states the correlation and then says plainly that it narrows the expected audiogram rather than determining it.

**One term from the deep-research report was rejected.** The report proposed `HP:0008565`; its own Term Validation section flagged that identifier as not existing in HPO. It is bound nowhere here, and I recorded the exclusion in the entry `notes` so a later curator reading the report does not rediscover and use it.

## Research input

One openscientist deep-research report, committed under `research/`. Preflighted with `just preflight-dr … MONDO:0011102` — PASS, TECTA the dominant gene mention, OMIM 601543 matching between MONDO and the report.

## Validation

| check | result |
|---|---|
| `just validate` (schema + references) | pass, **46/46** snippets verified against cached references |

Term validation first failed on an EBI OLS read timeout. That is a service error, not a data error, and it passed on retry — flagging it so the transient is not mistaken for a fixed problem.

## What I did not do

I did not add a `prevalence` block. DFNA8/12 is reported as a frequent cause of dominant nonsyndromic hearing loss, but the source says explicitly that its prevalence is unknown because no population-based screening has been done, and I would rather leave the slot empty than record a rate the literature declines to give.

Closes #10237

---

### DIFF
--- a/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12/2026-09-04T080738Z-claude-code-927396.yaml
+++ b/history/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12/2026-09-04T080738Z-claude-code-927396.yaml
@@ -1,0 +1,36 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12
+  path: kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12.yaml
+session:
+  id: 2026-09-04T080738Z-claude-code-927396
+  timestamp: '2026-09-04T08:07:38Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10250
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  summary: Resolve merge-queue conflict; fix wrong NCIT hearing-aid qualifier binding
+  details: |
+    Two changes in response to the merge-queue dequeue and the re-review on PR #10250.
+
+    Merge conflict. The GitHub merge queue dropped this PR with MERGE_CONFLICT. Resolved by merging origin/main into the branch. Two files conflicted, both derived term caches, both the same add/add shape where a new row lands at the same position in a sorted list: cache/uberon/terms.csv and cache/enums/anatomicalentityterm_c7eab8227a65.csv. This branch added UBERON:0002233 (tectorial membrane of cochlea, which this entry binds); main added UBERON:0002228 (rib, from another author's merged work). Both rows are legitimate and both were kept, in canonical sorted order. No cache row was hand-written beyond placing those two lines that the merge could not order; "just normalize-cache" afterwards reported 0 files updated, which is the check that the resolution was canonical rather than merely plausible, and check-term-cache-integrity and check-cache-order both pass. The effective diff against main is unchanged at 36 files.
+
+    Wrong ontology binding on a qualifier, caught in re-review. The medical-device qualifier on the hearing amplification treatment was bound to NCIT:C49947 with the label "Hearing Aid". NCIT:C49947 is actually "Fan Device". The correct term is NCIT:C183182 "Hearing Aid", which was already in cache/ncit/terms.csv and is what three sibling entries bind at the same slot (Otofacial_Neurodevelopmental_Syndrome, Autosomal_Recessive_Nonsyndromic_Hearing_Loss_77, Ageing_Associated_Decline_in_Intrinsic_Capacity). Rebound; preferred_term "hearing aid" and the NCIT:C15315 Rehabilitation action term are unchanged.
+
+    Why nothing caught it, which is the part worth recording. linkml-term-validator does not descend into qualifiers, since Qualifier.predicate and Qualifier.value are plain Descriptors with no dynamic-enum binding, so "just validate-terms" reported a pass on a fabricated label (issue #10197). The offline gate could not catch it either: NCIT:C49947 is a qualifier-only CURIE that nothing else in the KB seeds, so it was in no cache and check-qualifier-terms reported it as uncached and unchecked rather than as wrong. Only the online form resolves against the ontology.
+
+    This entry now carries four qualifier-bound terms across two device treatments, and "just check-qualifier-terms-online" was run against all four after the fix: it reports "OK: 4 qualifier term label(s) match the ontology". Recording that explicitly, because the offline gate's pass does not distinguish "checked" from "unable to check", and a later curator reading a green offline result has no way to tell which of the two it means.
+
+    Validation on the resulting head: just validate passes with 58/58 snippets verified; validate-terms, check-qualifier-terms (offline and online), check-entity-refs, check-duplicate-keys, check-enum-values, check-causal-targets, check-folded-hyphens, check-snippet-length, check-title-snippets, check-snippet-grading, check-environmental-evidence, check-cache-order and check-term-cache-integrity all pass. Zero cache churn from the rebind, since NCIT:C183182 was already cached.
--- a/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12.yaml
+++ b/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12.yaml
@@ -1033,7 +1033,7 @@ treatments:
       value:
         preferred_term: hearing aid
         term:
-          id: NCIT:C49947
+          id: NCIT:C183182
           label: Hearing Aid
   therapeutic_modality: DEVICE
   target_mechanisms:


aleixpuigb:
All four blocking items are addressed, plus all four suggestions. Pushed in b04bb2646.

Your central point was right and worth saying plainly: I fetched 25 references and cited 10, and the gap was not random — the uncited ones were exactly the management, genetics and diagnostics material. I had treated the deep-research report as a source of *mechanism* leads and stopped reading once the pathograph was solid.

**Item 1 — cochlear implantation.** Added as a `treatments` entry on `NCIT:C15329` with `NCIT:C157820` as the medical-device qualifier, `target_mechanisms` → Cochlear Amplifier Failure, cited to PMID:24130743. You were right that the `description` was asserting "cochlear implantation works well" with nothing behind it — that is exactly the kind of claim that should not sit in prose.

**Item 2 — splice-altering alleles.** This was the most useful catch. PMID:40583560's qPCR result — no reduction in carrier transcript level — is stronger evidence against haploinsufficiency than the DFNB21 carrier comparison I had built the hypothesis on, because it is measured in a DFNA8/12 family rather than inferred across allele classes. It is now the lead evidence on `dominant_negative_tectorin_incorporation`, with the DFNB21 argument kept alongside it.

I also renamed the node from *Heterozygous TECTA Missense Variant* to *Heterozygous Pathogenic TECTA Variant*. Your item made the old name wrong, not just incomplete: a coding variant that acts by skipping exon 20 is not a missense mechanism.

**Item 3 — genetic depth.** Added the four recurrent alleles with founder effects confirmed for two, and the size and private-variant character of the spectrum. Your observation that three of the four recurrent alleles involve a cysteine is the useful part — the progressive course was keyed on "cysteine-substituting variants" as a class, and it is now grounded in named alleles.

On penetrance: I could not find a figure. Rather than leave it silently absent I recorded in the `genetic` notes that no source quantifies it and that the pedigrees are multigenerational with dominant transmission. If you know of a source I have missed, I will add it.

**Item 4 — GeneReviews.** PMID:20301607 tagged. Thank you for the warning that its cached abstract holds only the purpose statement — that saved a wasted pass. I filled the genetic-counseling gap differently from the `COL11A2_Hearing_Loss` pattern: instead of citing the NBK URL, I used PMID:40583560, where prenatal diagnosis was actually provided to a DFNA8/12 kindred once the variant was identified. That seemed better than citing a general chapter for a claim a specific paper documents. Happy to add the URL citation as well if you would rather have the four-section GR coverage.

**Suggestions, all taken.**

- **6, environmental.** Added, linked `EXACERBATES` → Secondary Outer Hair Cell Vulnerability. `exposure_term` is unbound: I searched ECTO through `sqlite:obo:ecto` for both "noise" and "sound" and it returns nothing for either. That search is recorded in `review_notes` so it is not repeated.
- **7, entity refs.** Changed to the two specific model names on `mouse_does_not_progress`, left `animal_models#Mouse` on the seizure discussion where "all three mice" is the actual claim.
- **8, harrisons_chapter snippet.** Replaced with a sentence that characterises the disorder.
- **5, unused references.** Added a `notes` paragraph naming the ones I reviewed and deliberately did not cite, with the reason, so the next curator does not read their presence in `references_cache/` as an unmined seam.

Snippets went 46/46 → **58/58** verified. `validate-disorders`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-environmental-evidence`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens` and `check-snippet-grading` all pass.

Ready for re-review.

---

aleixpuigb:
Fixed in 12970bde1. You were right, and this one deserves a plain answer rather than a summary.

**I recorded a negative search result that I had not actually established.** `ECTO:8000044` "exposure to sound radiation" exists, sits in `cache/ecto/terms.csv` and the `exposureterm` enum cache, and is bound on the equivalent noise-`EXACERBATES` entry in DFNA41 — the sibling entry your previous review had already pointed me at.

What happened: my `runoak -i sqlite:obo:ecto search` calls for "noise" and "sound" both returned empty output, and I took that at face value. I had grepped the term cache too, but only for `noise` — and the label begins "exposure to sound radiation", so the string I searched for was never going to match a term that was sitting right there. I checked the cache first for every HP, GO, CL and NCIT term in this entry. For ECTO I trusted a tool that had just returned nothing, and did not cross-check.

Your framing of why this is worse than an unbound term is the part I want to acknowledge: a `review_notes` line saying "I searched and there is nothing" is an instruction to the next curator not to look. An unbound term with no note is merely incomplete. I turned a gap into a false record. So the paragraph is deleted rather than reworded — there is no search result worth keeping.

`just environmental-term-audit` now reports the entry as `[BOUND]` rather than `[PARTIAL]`.

You are also right that the audit's reuse matcher was never going to catch this: it matches curator-written names, and "Excessive noise exposure" does not resemble "exposure to sound radiation" as a string. That is exactly the case CLAUDE.md flags when it says a reuse suggestion is advisory and the check is the curator's — and #8430's lesson, an assumed gap whose lead example turned out to be bound already, is the same failure I just reproduced.

**On the suggestion.** Taken, and you were right that the dismissal read as stronger than it was. The nine uncited references are now split by actual reason, and PMID:33559882 and PMID:21368133 are called out as tectorin biology that bears on this disease, omitted because PMID:11087000 and PMID:26806019 already carry the detached-membrane and matrix-composition claims from primary sources the entry cites — not because they are off-topic. I added a line pointing a curator extending the tectorial-membrane interactome at them.

Snippets stay at 58/58. `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys` and `check-environmental-evidence` all pass.

Ready for re-review.

---

aleixpuigb:
Heads-up on one line in your re-review checklist, before it turns into a finding: **there is no CLAUDE.md change in this PR.**

Commit `12970bde1` touches exactly one file:

```
$ git show --name-only --format="" 12970bde1
kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12.yaml
```

And the branch as a whole does not touch it either — `git diff origin/main...HEAD --name-only` (three-dot, merge-base) returns no CLAUDE.md.

What you are almost certainly seeing is a two-dot diff. `main` moved while this PR was open, and one of the commits it gained is `db7896685` "Gate retired enum values whole-KB and ungated (#10061) (#10229)", which **does** change CLAUDE.md. My branch point is `c0ef148ed`, before that commit, so `git diff origin/main HEAD` renders main's own CLAUDE.md edit as a change in my direction — reversed. The three-dot form is the one that attributes correctly.

Worth flagging because that same commit is substantive for this batch: it adds an ungated, whole-KB `check-enum-values` gate that my branches predate and therefore will not run in their own CI. I ran main's version of `scripts/check_enum_values.py` against all four entries from this run, inside their own worktrees so the globs resolved:

| entry | result |
|---|---|
| `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12` | OK, no out-of-enum values |
| `Distal_Hereditary_Motor_Neuronopathy_Type_5B` | OK |
| `Combined_Oxidative_Phosphorylation_Defect_Type_30` | OK |
| `Immunodeficiency_25` | OK |

So none of the four carries a retired enum value, and merging them will not turn that new gate red.

I have deliberately not merged `main` into this branch while your re-review is mid-flight, since that would move the head under you and restart CI. Say the word and I will do it — the only diff it removes is the phantom CLAUDE.md one.

---

aleixpuigb:
`test (3.13)` went red on `12970bde1`. Fixed in 0ba11aa93 — two causes.

**Mine.** The `notes` paragraph I added in that very commit wrapped `mechanotransduction` across a fold boundary as `mechano-` / `transduction`, which a folded scalar renders as "mechano- transduction". Reflowed. I had run `check-folded-hyphens` on this entry earlier in the round, but not again after rewriting the notes — I re-ran the file-scoped checks and not the whole-KB ones, and the whole-KB ones are what catch prose defects. Same mistake turned up independently on #10255, so it was a habit rather than a slip; I now run the full set before every push.

**Not mine, but mine to clear — and it is not the shallow-checkout artifact.** You diagnosed the `check-snippet-length` failure on `Hyperprolinemia_Type_2.yaml` as a shallow checkout with no computable merge-base. I reproduced it locally in a full clone, so that is not it. The actual cause: the check baselines against `origin/main`, and **PR #10234 re-cut exactly those four snippets on `main`** after this branch was created. The branch was carrying the pre-#10234 file and being compared against the post-#10234 baseline, so its own old snippets read as new.

That distinction matters for the repo, because the two diagnoses have different fixes. A shallow-checkout bug would need a CI change and would hit every PR regardless of age. This needs nothing changed — it resolves the moment a branch merges `main`, and it only affects branches created before #10234. I have merged `main` into this one and the check passes.

The merge also brings in `db7896685` and the new ungated `check-enum-values` gate, which this entry passes, and it removes the phantom CLAUDE.md entry from two-dot diffs of this branch.

Verified after the merge, before pushing: `validate`, `validate-terms` (58/58 snippets), `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-entity-refs`, `check-duplicate-keys` — all pass. Diff scope against `main` unchanged.

---

aleixpuigb:
The merge queue dropped this PR with `MERGE_CONFLICT`. Resolved in 26aec8d77 by merging `origin/main` in.

Two files conflicted, both derived term caches, and both were the same add/add shape — a new row landing at the same position in a sorted list:

| file | this branch added | `main` added |
|---|---|---|
| `cache/uberon/terms.csv` | `UBERON:0002233` tectorial membrane of cochlea | `UBERON:0002228` rib |
| `cache/enums/anatomicalentityterm_c7eab8227a65.csv` | `UBERON:0002233` | `UBERON:0002228` |

Both rows are real and both were kept, in canonical sorted order — `0002228` then `0002233`. Neither side is droppable: the cochlear term is what this entry binds, the rib is someone else's and is already on `main`.

I did not hand-edit the caches beyond placing the two rows the merge could not order. `just normalize-cache` afterwards reported **0 files updated**, which is the check that the resolution was already canonical rather than merely plausible, and `check-term-cache-integrity` and `check-cache-order` both pass.

The PR's effective diff against `main` is unchanged at **36 files, +6,324/−30**. The merge brought in three days of `main` but contributed nothing of its own to the change set.

Validation on the merged head:

| check | result |
|---|---|
| `just validate` | pass, **58/58** snippets verified |
| `just check-folded-hyphens` | pass |
| `just check-snippet-length` | pass |
| `just check-title-snippets` | pass |
| `just check-snippet-grading` | pass |
| `just check-environmental-evidence` | pass |
| `just check-enum-values` | pass (2,932 KB files) |
| `just check-term-cache-integrity` / `check-cache-order` | pass |

This push dismisses the approval, which is unavoidable — a conflicted PR cannot merge, so there was nothing to protect by leaving it alone. Ready for re-review.

---

aleixpuigb:
Good catch, and it was a real defect rather than a near-miss. Fixed in 84c0307c9.

I verified it independently before touching anything, because a wrong-CURIE claim is exactly the kind that should not be taken on trust:

```
$ just check-qualifier-terms-online kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_12.yaml
     qualifiers.value: NCIT:C49947
     curated label : 'Hearing Aid'
     ontology label: 'Fan Device'
```

`NCIT:C183182` is the real Hearing Aid term, was already in `cache/ncit/terms.csv`, and is what all three sibling entries bind at this slot. Rebound; `preferred_term: hearing aid` and the `NCIT:C15315` Rehabilitation action term are unchanged. The check now reports **all four** qualifier terms in this entry matching the ontology, and the rebind produced zero cache churn since the term was already there.

Your account of why nothing caught it is the part I think matters most, so I have written it into the history record rather than leaving it in the PR thread. Two independent gaps had to line up: `linkml-term-validator` does not descend into `qualifiers`, so `validate-terms` passes on a fabricated label (#10197); and because a qualifier-only CURIE is seeded by nothing else in the KB, `NCIT:C49947` was in no cache, so the offline gate reported it *uncached and unchecked* rather than wrong. A curator glancing at two green checks would reasonably conclude the binding had been verified.

That is also why I took the suggestion about the history record, and stated it in those terms: the record now says `check-qualifier-terms-online` was actually run and what it returned, because a green offline result cannot distinguish "checked" from "could not check". This entry carries four qualifier-bound terms across two device treatments, so that distinction is worth preserving for whoever edits it next.

Validation on 84c0307c9:

| check | result |
|---|---|
| `just validate` | pass, **58/58** snippets verified |
| `just check-qualifier-terms` (offline) | OK — 4 of 4 now cached and matching |
| `just check-qualifier-terms-online` | OK — 4 of 4 match the ontology |
| `just check-enum-values` | pass (2,932 KB files) |
| `just check-causal-targets` | pass |
| `just check-folded-hyphens` / `snippet-length` / `title-snippets` / `snippet-grading` | pass |
| `just check-environmental-evidence` | pass |
| `just check-cache-order` / `check-term-cache-integrity` | pass |

Effective diff against `main` is 37 files now — the 36 from before plus this round's history record.

---

PR #10251, opened 2026-08-31
Curate distal hereditary motor neuronopathy type 5B (MONDO:0013884)

### LINKED ISSUE #10238: Curate neuronopathy, distal hereditary motor, type 5B (MONDO:0013884)
Curate a dismech entry for **neuronopathy, distal hereditary motor, type 5B** ([MONDO:0013884](https://monarchinitiative.org/MONDO:0013884)) — dHMN5B, caused by dominant `REEP1` variants. REEP1 is a hairpin-loop ER-shaping protein of the REEP/reticulon family that maintains tubular ER and its contacts with microtubules and mitochondria; the same gene causes SPG31, so the interesting curation question is what makes one allele a pure distal motor neuronopathy and another an upper-motor-neuron spastic paraplegia.

Preflight notes: MONDO leaf, single causal gene `hgnc:25786 REEP1`. On `origin/main`, `REEP1` appears only in `kb/disorders/Hereditary_Spastic_Paraplegia.yaml` (as the SPG31 gene), and the MONDO ID appears nowhere. No open PR covers it. The KB already carries `Distal_Hereditary_Motor_Neuronopathy_Type_9` plus the two mode-of-inheritance dHMN entries, so a per-type entry fits convention — but check against those three for overlap before creating a new file.

Stub: `stubs/Neuronopathy_Distal_Hereditary_Motor_Type_5B.yaml`

aleixpuigb:
Adds a new disorder entry for dHMN5B, the REEP1-associated pure motor axonal neuropathy.

The reason this deserves its own entry rather than a note on the REEP1 spastic paraplegia page is that the two REEP1 diseases are mechanistic opposites. Losing REEP1 function causes hereditary spastic paraplegia by degenerating **cortical** motor neuron axons. dHMN5B comes from alleles that leave the protein fully able to bend membrane but no longer able to be cleared, and it degenerates **spinal** ones. It is a toxic gain of function, and the entry is built around that.

## Pathograph

```
REEP1 exon 5 skipping variant (in frame, stable protein)
        ↓
impaired HUWE1-dependent ubiquitination and proteasomal turnover
        ↓
accumulation of shaping-competent REEP1 in peripheral nerve
        ↓
ER fragmentation in spinal motor neurons
        ↓
spinal motor neuron axonal degeneration
        ↓
distal motor amyotrophy, upper-limb predominant
```

The founding allele destroys the exon 5 splice acceptor. Skipping is in frame, so the transcript escapes nonsense-mediated decay and makes a stable protein missing 38 internal residues — which is the whole point. A null would give the other disease.

**The two mouse models are curated as a matched pair, and one of them is a negative.** The exon-5 knock-in `RECAPITULATES` the spinal lesion. The `Reep1` knockout is recorded as `FAILS_TO_RECAPITULATE`, because it loses cortical motor neurons while spinal ones stay intact — it models SPG31, not this disease. I included it deliberately rather than leaving it out: the knockout is the model somebody searching for "a REEP1 mouse" will find first, and any inference it supports about dHMN5B would be pointing the wrong way.

**`GAIN_OF_FUNCTION` rather than `INCREASED` on the accumulation node.** Per the CLAUDE.md guidance, that trades PATO grounding for expressivity, so it should be deliberate. It is: the claim is that the protein has escaped ubiquitin-proteasome regulation, not that its level is merely high. The reasoning is recorded in the entry `notes`.

**Several phenotype items are graded `directness: INDIRECT`, and I want to flag why rather than have it read as laziness.** The HPO annotation set for this disease lists thenar and peroneal atrophy, pes cavus and the reflex findings, but for a disorder described in a handful of families the primary literature does not state each of those in a quotable sentence. The deep-research report offered an electrophysiology quote — and was honest that it came from a **BSCL2** dHMN-V series. Using it would have attached another gene's patients to this disease's phenotype, which is Named Entity Confusion arriving through the phenotype rather than the gene. I cited the closest true statement about REEP1 disease instead and named the gap in each `explanation`.

**Evidence concentration, stated plainly.** The ubiquitination defect, the HUWE1 identification, the liposome assay and both mouse lines all come from one 2026 study (PMID:41268727). The chain is coherent and internally well controlled, but it has not been independently reproduced. The corroborating paper supports only the *direction* of the mechanism, in a different allele and a CMT2 family. That caveat is in the entry `notes`, not just in this PR.

## Research input

One openscientist deep-research report, committed under `research/` with its citations sidecar. Preflighted with `just preflight-dr … MONDO:0013884` — PASS, REEP1 the dominant gene mention (86 mentions), OMIM 614751 matching.

Worth noting: the report's quotes were accurate, but the **titles** it attributed to four PMIDs were paraphrases rather than real titles. I had copied them into `reference_title` fields; the reference validator caught every one, and all titles were rewritten from the cached records. Deep-research output is a lead for citations as well as for claims.

## Validation

| check | result |
|---|---|
| `just validate` (schema + references) | pass, **40/40** snippets verified against cached references |

## What I did not do

The `prevalence` block records the ~2.2 per 100,000 figure for dHMN **as a group**, graded `INDIRECT`, with a note saying so. That number aggregates more than thirty genes and REEP1 is one of the rarer ones, so it is an upper bound rather than an estimate for this subtype. I recorded it anyway because it is the only quantitative anchor available and leaving the slot blank would have been less informative than leaving it caveated — happy to drop it if reviewers would rather have nothing there.

Closes #10238

---


aleixpuigb:
All three IMPORTANT items and all four suggestions are addressed. Pushed in c1e07af61.

**Item 1 was a real defect and I want to name it as one rather than let it pass as a hedge.** I had a treatment entry for rehabilitative management evidenced by a sentence about Brown-Vialetto-Van Laere syndrome and riboflavin. I graded it `INDIRECT` and wrote an explanation reasoning from "the review names the one treatable mimic, therefore the rest are managed supportively" — but that is an argument I constructed, not something the quoted sentence says. `INDIRECT` is for a claim that follows from the quote by an inference step, not for a quote that is about a different topic. Swapped for PMID:24646194 and PMID:12736893, both graded `INDIRECT` with the CMT provenance stated plainly in each explanation. BVVL moved to `differential_diagnoses`.

**Item 2 — atlastin-1.** You were right that this mattered more than a routine omission, and the reason you gave is the one that persuaded me: the entry's own `notes` admit the mechanism rests on a single 2026 paper, and this is the only pre-2026, allele-specific, independent-group data pointing the same way. Leaving it out while flagging the concentration was incoherent.

It is now a node between accumulation and ER fragmentation, with `mechanism_confidence: PROVISIONAL` and a downstream edge of `INDIRECT_UNKNOWN_INTERMEDIATES` — the localization work is overexpression in non-neuronal lines, and nobody has shown atlastin-1 mis-recruitment is what fragments the ER in the knock-in mouse. I kept the direct accumulation→fragmentation edge alongside it rather than replacing it, since the 2026 paper establishes that link independently. The closing sentence is also a second evidence item on the hypothesis group.

One thing I could not do: I wanted an `ATL1` gene-product descriptor on that node, but `GeneProductTerm` does not accept HGNC CURIEs and I did not want to guess at a PR or UniProt identifier. Atlastin-1 is named in the description and in the snippet, so nothing is lost except queryability.

**Item 3 — pyramidal signs.** Cited nothing because there is nothing to cite: I re-read the deep-research report and it asserts pyramidal signs uncited as well. So it was an uncited claim inherited from an uncited claim. Both passages now state what is documented — pronounced hyperreflexia in the recessive REEP1 phenotype — rather than repeating it.

**Suggestions.**

- **5** was the most useful. PMID:27066569 does not just soften "cleanly", it breaks the two-way framing outright: a recessive REEP1 allele producing a SMARD1-like congenital neuropathy fits neither dominant category. Added to `genetic` evidence and as a differential.
- **6** — you were right that the exclusion needed writing down, not just doing. Both papers are SPG31-framed loss-of-function work, which is the direction this entry argues against and the reason the KO mouse is graded `FAILS_TO_RECAPITULATE`. That reasoning is now in `notes`.
- **4** — HP:0003431 documented as a deliberate drop, with your reasoning: the backing quote reports conduction velocities as *normal*, which is what an axonal rather than demyelinating neuropathy predicts. The axonal pattern is carried by HP:0007002.
- **7** — added a `progression` block for the slow course, `INDIRECT` at the dHMN group level. I did not add onset ages or frequencies to phenotypes, because no REEP1-specific natural-history data exist and the report's "adolescent/adult" is unsourced; `progression.notes` says so.

On prevalence — thank you for confirming. That was the item I was least sure about.

On GeneReviews: your N/A reading matches mine, and I appreciate the disclosure about E-utilities being unreachable. I searched independently and found no REEP1 or dHMN chapter either.

Snippets 40/40 → **49/49**. `validate-disorders`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-grading` and `check-title-snippets` all pass.

Ready for re-review.

---

aleixpuigb:
Thanks — stopping here rather than acting on the five optional items, because `main` has `dismiss_stale_reviews` on and any push would drop this approval and start another review cycle for changes you explicitly marked non-blocking. Recording what I would otherwise do, so it is not lost:

**(a) is the one that is actually worth doing, and it is a real defect rather than polish.** `toxic_gain_of_shaping_competent_reep1` is declared and evidenced but no `downstream[].hypothesis_groups` opts into it, so the hypothesis group is dangling — it will not attach to any causal edge in the graph or the exports. You are right that it is pre-existing from the first commit. The members should be the accumulation→ER-fragmentation edge and the new atlastin-1 edge. I would rather that went in as a small follow-up PR than as a push that costs this approval; happy to open one now if you would prefer it in this PR instead, in which case say so and I will push and re-request.

**(b)** — "Distal wasting is described in SPG31" being uncited is a fair catch, and your check matches mine: `PMID:29107646` reports pure spastic paraplegia in its family, and the distal amyotrophy quotes in `PMID:15242882` are BSCL2. I would drop the clause rather than hunt a citation, since the operative claim already has a cited leg via the recessive phenotype. Same follow-up.

**(c)** — agreed, and I had the same doubt when I wrote it. "These data corroborate…" asserts a different mechanism, not fragmentation specifically. That is why the edge carries `INDIRECT_UNKNOWN_INTERMEDIATES` and the item carries `directness: INDIRECT` with an explanation saying the step is inferred from atlastin-1's known role. On the missing `directness` on the hypothesis-group copy of the same quote: there it backs the two-mechanism claim the sentence actually makes, so `DIRECT` is right and absent means unassessed rather than direct — I should have set it explicitly. Follow-up.

**(d)** — fair. One observation, two nouns in the name.

**(e)** — I will add the second `history/` record in the same follow-up. You are right that the existing one now under-describes the entry after this round.

On `check-snippet-length`: thank you for diagnosing that as a shallow-checkout baseline artifact on `Hyperprolinemia_Type_2.yaml` rather than reporting it as my failure. I confirmed the same — those snippets are in `origin/main` today and this PR does not touch that file. That is a CI environment issue worth someone looking at separately, since it will fire on every PR.

Finally, on the treatment item from the first round: your framing was the useful part. I had graded that evidence `INDIRECT` and written an explanation reasoning my way from the quote to the claim. But `INDIRECT` is for a claim that follows from a quote by an inference step, not for a quote about a different subject — the sentence was about a treatable mimic, not about rehabilitation. Calling it "support for a different claim" rather than "weak support" is the distinction I had blurred, and it is the one I will carry into the next entry.

---

Issue #10108, opened 2026-08-29
Schema gap: no way to assert a required-NORMAL lab value (e.g. normal IgG/IgM in selective IgA deficiency)


cmungall:
Surfaced while curating Selective IgA Deficiency for the IUIS campaign. The disease definition requires decreased IgA WITH NORMAL IgG and IgM — the normality of the other isotypes is diagnostic, not incidental. Neither `Phenotype` nor `Descriptor` accepts a `presence: ABSENT`/normality assertion, so 'serum IgG and IgM normal' cannot be modeled as a phenotype; the curation parked it in the `diagnosis:` section with its defining quote.

Options that come to mind: (a) an explicit `presence` or `normality` slot on Phenotype (HPO-style excluded-phenotype semantics), (b) a modifier value for NORMAL on descriptors, or (c) blessing the diagnosis-section workaround as the convention and documenting it. Related precedent: biochemical reference_ranges can express a normal band, but that expresses the interval, not the claim 'this analyte is normal in this disease and that is definitionally required'.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

aleixpuigb:
Hit this same gap from a different direction while curating COXPD30 (#10252), and it turns out the fix proposed above does not reach my case.

The short version: `frequency: EXCLUDED` works because `Phenotype` has a `frequency` slot. `Descriptor` does not. So adding `EXCLUDED` to `FrequencyEnum` closes the phenotype and lab-value case in this issue, but leaves the descriptor case — option (b) in the original post — still open.

## The instance

TRMT10C/MRPP1 has two separable jobs: it is part of the enzyme that makes the 5′ cut in mitochondrial tRNA processing, and it methylates position 9 of mt-tRNAs. The disease variants break the first and leave the second **measurably intact**, tested two independent ways. That negative result is what assigns the mechanism to the cleavage arm specifically — without it the entry's mechanism reads as less specific than the evidence supports.

It is curated as a pathophysiology node with a `biological_processes` descriptor:

```yaml
- name: Preserved m1R9 Methyltransferase Activity
  biological_processes:
  - preferred_term: mitochondrial mt-tRNA m1R9 methylation
    term:
      id: GO:0070901
      label: mitochondrial tRNA methylation
    # no modifier: the finding is that this process is unchanged
```

[...]

PR #10252, opened 2026-08-31
Curate combined oxidative phosphorylation defect type 30 (MONDO:0014856)

### LINKED ISSUE #10236: Curate combined oxidative phosphorylation defect type 30 (MONDO:0014856)
Curate a dismech entry for **combined oxidative phosphorylation defect type 30** ([MONDO:0014856](https://monarchinitiative.org/MONDO:0014856)) — a mitochondrial disease caused by biallelic `TRMT10C` variants. TRMT10C (MRPP1) is a subunit of mitochondrial RNase P and also carries out m1R9 methylation of mitochondrial tRNAs, so loss of function stalls 5' processing of the polycistronic mitochondrial transcript and depletes every mtDNA-encoded OXPHOS subunit at once.

Preflight notes: MONDO leaf, single causal gene `hgnc:26022 TRMT10C`. `TRMT10C` appears on `origin/main` only inside `kb/disorders/HSD10_Mitochondrial_Disease.yaml`, where it is named as the partner subunit of HSD17B10/MRPP2 rather than curated as its own disease; the MONDO ID appears nowhere. No open PR covers it. The KB already carries several other per-type COXPD entries, so a per-type entry fits existing convention.

Stub: `stubs/Combined_Oxidative_Phosphorylation_Defect_Type_30.yaml`

aleixpuigb:
Adds a new disorder entry for COXPD30, caused by biallelic `TRMT10C` variants.

TRMT10C encodes MRPP1, which sits at the very first step of mitochondrial RNA maturation. Mitochondrial DNA is transcribed as one long polycistronic message with the tRNA genes acting as punctuation between the rRNAs and mRNAs; getting anything usable out of it requires cutting at those tRNAs. MRPP1 is part of the enzyme that makes the 5' cut. So a single defect here starves every mtDNA-encoded subunit at once — which is why the biochemistry is a *combined* deficiency of complexes I and IV with complex II spared. Complex II is the one respiratory complex with no mtDNA-encoded subunit, so its sparing is the fingerprint of a lesion in mitochondrial gene expression rather than in a structural subunit.

## Pathograph

```
biallelic TRMT10C missense variants
        ↓
MRPP1 protein destabilization
        ↓
impaired mitochondrial RNase P 5′ processing
        ↓
impaired mitochondrial protein synthesis
        ↓
combined complex I and IV deficiency
        ↓
systemic lactic acidosis and multiorgan energy failure

[ Preserved m1R9 methyltransferase activity ]  ← no downstream edge, see below
```

## Two things I want a reviewer to look at

**The disconnected node is deliberate.** MRPP1 has two separable jobs: it is part of the RNase P cleavage enzyme, and with MRPP2 it methylates position 9 of mt-tRNAs. The disease variants break the first and leave the second measurably intact, tested two independent ways. That negative result is what assigns the mechanism to the cleavage arm specifically, so I curated it as a pathophysiology node with no `downstream` edge and said in the `notes` why. Removing it would make the entry's mechanism look less specific than the evidence actually supports.

**That node exposes a schema gap, and I would like a second opinion on how I handled it.** The finding is that a process was measured and found *unchanged*. `ModifierEnum` has no value for that — the options are `INCREASED`, `DECREASED`, `ABNORMAL`, `DYSREGULATED`, `ABSENT`, `GAIN_OF_FUNCTION`, `LOSS_OF_FUNCTION`, and the nearest ones (`ABNORMAL`, `DYSREGULATED`) would assert the opposite of what was found. I left `modifier` off entirely, since absent means "not asserted", and documented the reasoning in the entry `notes`.

Worth noting that `ExperimentalReadout.direction` *does* have `UNCHANGED`, and CLAUDE.md explicitly calls it "a real negative result". So the schema already accepts measured-and-normal for readouts and not for descriptors. I have not opened an issue for this — it is a one-node problem here and I would rather a maintainer decide whether it is worth an enum change than file against a schema I have used for a few hours. Happy to open one if you would like.

## A note on the mechanism being quantitative

This is not a null. Both reported alleles are missense, both leave residual protein, and mature mt-tRNA and mt-mRNA levels are *largely normal* in patient fibroblasts — what accumulates is unprocessed precursor at the gene boundaries. The defect is throughput, not completion. The original authors say the fibroblast defect was milder than they expected given the clinical course, and I recorded that as an open `KNOWLEDGE_GAP` rather than smoothing it over.

## Research input

One openscientist deep-research report, committed under `research/`. Preflight (`just preflight-dr … MONDO:0014856`) returned **WARN**, flagging PRORP at 36% of the TRMT10C mention rate as a possible second entity. I checked: PRORP is the third subunit of the same RNase P holoenzyme, so those mentions are mechanism context, not a mixed-in disease. The report also warns against confusing COXPD30 with TRMT5-related COXPD26 — adjacent numbering, both mt-tRNA modification — and I added that disambiguation to the entry `notes`.

The report contributed one genuinely useful reference the primary literature search had not surfaced: the cryo-EM structure of human mitochondrial RNase P (PMID:34489609), showing PRORP is recruited and activated *through interaction with* the TRMT10C subcomplex. That makes the paper's own MRPP1–MRPP3 interface hypothesis structurally coherent rather than merely plausible, and it is attached to that `mechanistic_hypotheses` entry as `INDIRECT` evidence (the structure is of wild-type complex and does not test the disease variants).

## Validation

| check | result |
|---|---|
| `just validate` (schema + references) | pass, **45/45** snippets verified against cached references |

## Scope caveat

Two unrelated individuals have ever been reported. Every severity, frequency and organ-involvement statement in the entry is drawn from an n of 2, and the `notes` say so rather than letting the entry read as a settled disease description.

Closes #10236

---


aleixpuigb:
Resolved a merge conflict with `main` in 18f476840, and pushed a round of improvements before that in 4ff1da02b.

**The conflict** was an add/add on `references_cache/PMID_35109800.md` — both this branch and `main` added the same reference cache file. I took `main`'s version, which is a superset (full text rather than abstract-only), so the file is now byte-identical to `main` and contributes nothing to this diff. No snippet in the entry cites that PMID, so nothing downstream depended on which side won. Merged `main` in rather than rebasing, per the branch convention.

Post-merge: `just validate` passes, 51/51 snippets verified, `validate-terms`, `check-entity-refs` and `check-duplicate-keys` all clean. The diff against `main` is 11 additions, 1 deletion (the stub) and 5 cache modifications — no unrelated churn.

**The earlier push** pre-empted the pattern the three sibling PRs in this batch all drew from review: references fetched by the deep-research run and then never cited, with the gaps landing consistently on the clinical envelope rather than the mechanism. Three uncited references are now used, and each closes something real:

- **PMID:37558808** lets the PRORP/COXPD54 differential be curated with evidence instead of omitted. It is the sharpest comparison available for this disorder — the same 5' cleavage reaction degraded from the nuclease side rather than the scaffold side, in an assay that includes TRMT10C itself — and the phenotype is a Perrault-to-leukodystrophy spectrum rather than the neonatal lethal course here. Same lesion, different subunit, different disease.
- **PMID:26189817** does the same for TRMT5/COXPD26. That disambiguation was previously bare prose in `notes`; it is now a differential with its own evidence naming the distinct lesion (G37 hypomodification, not tRNA excision) and the much wider clinical range.
- **PMID:38824131** adds the folding quality-control role of the MRPP1–MRPP2 subcomplex to the processing node, `INDIRECT` since the structures are wild-type. It bears directly on the entry's open question about why the fibroblast defect looks milder than the clinical course.

Also added a `prevalence` block as `CASES_IN_LITERATURE` / `ULTRA_RARE` — two individuals, framed as a count of the literature rather than a rate — and a `notes` paragraph listing the references reviewed and deliberately dropped.

Snippets 45/45 → **51/51**.

One thing from the original PR body still stands and I would still value a view on it: the `Preserved m1R9 Methyltransferase Activity` node carries no `modifier`, because `ModifierEnum` has no value for a process measured and found unchanged, while `ExperimentalReadout.direction` does have `UNCHANGED` for exactly that. I have not filed an issue, since it is a one-node problem here and a maintainer is better placed to judge whether it warrants an enum change.

---

aleixpuigb:
Both blocking items fixed, plus two of the three suggestions. Pushed in `a5d1ca26a`.

**The two GO terms.** Replaced as suggested — `GO:0090646` → `GO:0097745` (mitochondrial tRNA 5′-end processing) and `GO:0030488` → `GO:0070901` (mitochondrial tRNA methylation). You were right that the first one matters more than a general specificity nit: the node exists to say the lesion is at 5′ cleavage while ELAC2-mediated 3′ processing is intact, and the parent term threw that away. `GO:0097745` was in neither the label cache nor the enum-membership cache, so it resolved live through the `ols:go` adapter and added one row to each; `GO:0070901` was already cached. Both rows landed in canonical position — `check-cache-order` is clean.

**The evidence grading.** The ExAC/ESP6500/COSMIC item is now `COMPUTATIONAL`. I also re-graded the conservation-plus-rarity item, which was `HUMAN_CLINICAL`.

I did not split that second one. Your reasoning for splitting was that the sentence bundles two claims, which it does — but both halves are database and alignment work, so they would carry the same grade, and the split would produce two items with no more information than one. The split rule in CLAUDE.md is for a paper that *mixes* evidence sources; here it does not. I put that reasoning in the item's `explanation` so the next reader does not re-open it.

**`preferred_term`.** Fixed — it now reads `mitochondrial mt-tRNA m1R9 methylation`. You are right that it was a finding in a naming slot.

**One correction to the location suggestion.** I took it, but the slot name is different from the one you named. `located_in` is on `Descriptor`; `Pathophysiology` carries `locations` (multivalued, `AnatomicalEntityDescriptor`). So the binding is:

```yaml
  locations:
  - preferred_term: skeletal muscle
    term:
      id: UBERON:0001134
      label: skeletal muscle tissue
```

on the combined complex I/IV node, whose own quoted snippet is a skeletal-muscle biopsy result. Note the canonical label is `skeletal muscle tissue`, not `skeletal muscle`.

**PMID:34715011** added to the PRORP differential. Good catch — the allelic-series framing was the premise the whole differential section rested on, and it was uncited. Snippets are now **52/52**.

**HP:0008314 I did not add.** Decreased respiratory-chain enzyme activity is a mechanism step in this disease and already has its own pathophysiology node; adding it as a phenotype would assert the same thing twice in two sections. Happy to be overruled, but it reads as duplication rather than coverage to me.

## On the ModifierEnum gap — filed as a comment, not a new issue

You asked me to file it. I checked first, and it is already filed: **#10108**, "no way to assert a required-NORMAL lab value", whose option (b) is literally "a modifier value for NORMAL on descriptors". So I commented there rather than opening a duplicate.

Three things I added that were not already on that issue, since they change what the fix has to be:

- **The resolution proposed on #10108 does not reach this case.** That thread converges on adding `EXCLUDED` to `FrequencyEnum`, which works because `Phenotype` has a `frequency` slot. `Descriptor` does not. So that fix closes the phenotype and lab-value case and leaves the descriptor case open.
- **`ModelReadoutDirectionEnum` already has `UNCHANGED`** (`dismech.yaml:1561`), and CLAUDE.md calls it "a real negative result". So "measured and found normal" is first-class in one slot and unrepresentable in the other, and the *measured-normal* versus *not-measured* distinction CLAUDE.md draws for readouts collapses for descriptors, since omission is the only way to write either.
- **A GO binding does not need the `Scott_Syndrome` fallback.** That entry leaves a preserved-function phenotype completely unbound because `HP:0011869 Abnormal platelet function` "asserts the exact opposite of what is curated here". Right for HP, which is an abnormality ontology — but `GO:0070901` names the reaction without asserting it is impaired, so the binding here is safe and the *only* missing piece is the modifier value.

I also noted two things I have not checked: whether PATO has a normality term that would keep a new value ontology-bound the way `INCREASED`/`DECREASED` are, and how many existing pathophysiology descriptors would want it — that is, whether this is a one-node problem or a backfill.

## Validation

| check | result |
|---|---|
| `just validate-disorders` | pass |
| `just validate` | pass, **52/52** snippets (was 51/51) |
| `just check-enum-values` | pass |
| `just check-entity-refs`, `check-duplicate-keys` | pass |
| `just check-cache-order` | pass, zero churn |
| `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` | pass |

History record added for this review-response session.

---

aleixpuigb:
Thanks for the approve. Your passing note on `COMPUTATIONAL` is filed as #10273 — it does recur, more than I expected.

I measured it before filing: 48 evidence items across 42 files quote a snippet naming a specific variant or population database, and they are graded `HUMAN_CLINICAL` 35, `COMPUTATIONAL` 7, `OTHER` 6. Two items already on `main` make the same kind of claim with different grades — "The homozygous state is absent in gnomAD" is `OTHER`, while "Analysis of gnomAD data showed … carrier rate is ∼1/140" is `COMPUTATIONAL`.

I was careful not to present the 35 as an error count. Some of them quote a frequency measured in the paper's own cohort, which really is `HUMAN_CLINICAL`. The problem is that nothing in the guidance separates that from a reference-database query, so the spread is what you would expect rather than a set of mistakes.

Filed separately from #10108 rather than as a line on it, since that one is a different slot and a different gap — asserting a measured-normal value, not grading the evidence type. I left the options open (add a value, document a convention, or leave it) and noted that widening an enum is the safe direction with respect to #10061, unlike the `PARTIAL` retirement.

---

PR #10254, opened 2026-08-31
curate: Add Dilated Cardiomyopathy 1GG (MONDO:0013339, SDHA)

### LINKED ISSUE #10246: Curate dilated cardiomyopathy 1GG (MONDO:0013339)
Curate a dismech entry for **dilated cardiomyopathy 1GG** ([MONDO:0013339](https://monarchinitiative.org/MONDO:0013339)) — *SDHA* variants; succinate dehydrogenase flavoprotein subunit A, the shared complex II/TCA-cycle enzyme, presenting here as isolated dilated cardiomyopathy.

**Read `kb/disorders/SDHA-Related_Neurodegeneration_With_Ataxia_And_Optic_Atrophy.yaml` first.** It already carries `MONDO:0013339` as a *differential diagnosis* — that is not coverage, but it means someone has already reasoned carefully about where the boundaries between the SDHA diseases fall, including which MONDO terms `RO:0004003` links to SDHA and why NDAXOA is not a subtype of the recessive complex II disorder. Do not redo that reasoning; build on it, and make the two entries agree.

The KB also holds `Dilated_Cardiomyopathy`, `Dilated_Cardiomyopathy_1EE` and `Dilated_Cardiomyopathy_1Z`, so per-subtype DCM entries are established convention here.

Stub: `stubs/Dilated_Cardiomyopathy_1GG.yaml`

kevinschaper:
Adds a new disease entry for dilated cardiomyopathy 1GG, a neonatal-onset heart muscle disease caused by inheriting two copies of one change in the *SDHA* gene. What makes it worth a separate entry is not the gene defect but where it shows up: the affected enzyme is severely depleted in heart muscle while skeletal muscle and blood cells keep most of their activity, so these children die of heart failure with normal development and no seizures. Nobody has explained why one organ fails when every cell carries the same change, and that unanswered question is the entry's central content.

Closes #10246

## The lump/split decision: `entry_type: DISEASE`

The honest case against a separate entry is real, so it is stated in the entry's `notes` rather than left out:

- The causal allele is **not private to this phenotype**. The same homozygous `c.1664G>A p.(Gly555Glu)` has been published in a lethal infantile presentation (PMID:12794685) and in a relatively mild Leigh syndrome (PMID:16798039). So the genotype cannot be used to assign a patient to this disease.
- `kb/disorders/Mitochondrial_Complex_II_Deficiency_Nuclear_Type_1.yaml` **already cites the founding report** (PMID:20551992) under a `Cardiomyopathy` phenotype node. The same 15 patients are currently claimed by both MONDO terms.

I still landed on DISEASE:

- MONDO keeps them as separate terms with separate OMIM anchors, and MONDO:0013339's parent is `MONDO:0700335` familial isolated dilated cardiomyopathy — **not** the complex II deficiency term. Filing it as a subtype of MC2DN1 would contradict the ontology.
- The KB has already accepted this split for the same gene: `SDHA-Related_Neurodegeneration_With_Ataxia_And_Optic_Atrophy` is curated separately from MC2DN1 on exactly this reasoning.
- The pathograph diverges where it matters. A tissue-restricted enzyme deficit with preserved neurological function is a different mechanism claim from Leigh syndrome, even when the allele is shared.

I did **not** withdraw the cardiac cohort from the MC2DN1 entry. Deciding whether that entry should keep the Levitas patients is a change to curated content that deserves its own PR and its own review. The overlap is disclosed in this entry's `notes` rather than quietly resolved.

## Separating the four SDHA diseases

`just preflight-dr` keys on the gene, so it cannot tell these apart. Every clinical source was checked for **which phenotype its patients had**, not just which gene was named. Recorded in `notes`:

| Reference | Used for | Not used for |
|---|---|---|
| PMID:20551992 | the cardiac phenotype of this entity | — |
| PMID:12794685, PMID:16798039 | what the p.Gly555Glu allele does, and its phenotypic range | any DCM1GG phenotype — both probands had non-cardiac disease |
| PMID:37064335 | the treatment landscape of complex II deficiency | any phenotype — different genotypes, one with Leigh syndrome |
| PMID:7550341 | the conserved flavoprotein domain | any phenotype — Leigh syndrome, different substitution |

Excluded as off-entity and cited nowhere: PMID:42325549 (SDHA in acquired ischemic heart disease) and PMID:41647802 (an incidental SDHA variant in *MYBPC3* hypertrophic cardiomyopathy).

The reciprocal differentials for NDAXOA, MC2DN1 and PGL5 were written to agree with what the NDAXOA entry already says rather than re-derived. One correction of substance: that entry frames the DCM1GG boundary as resting on zygosity, which is right for NDAXOA-vs-DCM1GG but **not** for the MC2DN1 boundary — both of those are biallelic and can share an allele. This entry says so explicitly.

## Deep research: on-entity, and it found something I missed

Falcon report committed. `just preflight-dr` returned WARN, but the warning is a false positive: it flagged OMIM 115200 where the report labels it "Parent/general disease" as cited in the primary paper. SDHA dominates the report (44 mentions) and G555E/Bedouin/Levitas appear 66 times.

The report surfaced **PMID:35803927**, a conditional cardiac *Sdhaf4* knockout mouse that I had not found. I had drafted a knowledge gap saying no animal model existed. That would have been wrong in substance: this mouse disrupts the SDHA–SDHB interaction, loses assembled complex II, and develops progressive lethal dilated cardiomyopathy, and its authors cite this entity's own G555E report as their premise. It is now curated in `animal_models` with three mechanism links, including a `RESCUES` link for the fumarate arm. The knowledge gap was rewritten to say what is actually missing — a model of the *allele*, and one whose cardiac restriction is emergent rather than imposed by a Cre driver.

- **No riboflavin treatment.** The evidence is about complex II deficiency as a class and points both ways: a systematic review grades its effect "uncertain" (PMID:42046426), individual reports describe mild improvement (PMID:37064335). None of it comes from a patient with this disorder, and mechanistically this allele breaks a subunit interface rather than flavin acquisition. Recorded as a knowledge gap instead.
- **No `conforms_to`.** `cardiomyopathy_maladaptive_remodeling#Ventricular Remodeling` expects fibroblast activation, matrix deposition and cardiomyocyte apoptosis; `#Primary Cardiomyocyte Insult` expects abnormal sarcomere organization. Neither has been reported in patients with this disorder. Declaring conformance would assert mechanism the entry cannot support. Reason recorded in `notes`.
- **No left ventricular noncompaction HP term**, though MedGen lists it — it is not in the founding report.
- **No ClinGen `CGGV:` nosology row.** `just clingen-refresh` currently fails a sha256 mismatch against the snapshot pinned in `data/clingen/MANIFEST.yaml` (2026-08-13); upstream has drifted. Getting the row would mean bumping that pin and rebuilding unrelated cache files. Flagging it as a tooling issue rather than working around it here.
- **No grouping membership change.** `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml` would be a reasonable home — this would be its first bioenergetic member — but membership there is a curatorial act, not automatic: the grouping already omits `Dilated Cardiomyopathy 1EE`, `1JJ` and `1Y`, which are curated entries. Adding a member also means rewriting the `grouping_rationale`, which hardcodes "thirteen members" and a seven-lesion-class list. That belongs in its own PR.

## One edit outside this entry

`kb/disorders/SDHA-Related_Neurodegeneration_With_Ataxia_And_Optic_Atrophy.yaml`, one sentence. Its notes said "There is no dismech entry for this disorder at the time of curation" — a statement about repository state that this PR falsifies. Replaced with a pointer to the new file.

## Provenance limits

Three prose statements are full-text derivations, not snippet-verified, and are marked as such in `notes`: the residual enzyme activity percentages (PMC2987458), the compendium's count of seventeen G555E cases (PMC7758838), and the inclusion of p.Gly555Glu in the HAP1 variant panel (PMC11611653). All three cache files are `abstract_only`, so every snippet drawn from them is scoped to what the abstract actually says.

## Validation

Every gate run to completion, exit code confirmed:

| Gate | Exit |
|---|---|
| `just validate-disorders` (this entry + NDAXOA) | 0 — 126/126 snippets verified |
| `just validate` | 0 — 46/46 snippets verified |
| `just validate-terms` | 0 |
| `just validate-stubs` | 0 |

No `cache/**/*.csv` rows were added — every ontology term this entry binds was already in the caches.

`evidence_source` grading line used consistently throughout: clinical and genetic association statements are `HUMAN_CLINICAL`; enzyme-activity and immunochemical measurements on patient tissue are `IN_VITRO` even inside a clinical paper; mouse results are `MODEL_ORGANISM`; reviews are `OTHER`.


kevinschaper:
All five blocking findings are fixed, plus the three suggestions. Every claim in the review checked out against the primary source — thank you, findings 2 and 4 were substantive errors on my part, not omissions.

**The fix that made the rest easier:** re-running `just fetch-reference PMID:20551992` returned `content_type: full_text_html` where the cache had previously held `abstract_only`. The first fetch had `full_text_attempted: true` and failed. So the findings that were flagged as full-text derivations are now quoted from the cache and machine-verified. Snippet count went 46 → 54, all exact.

| Finding | Fix |
|---|---|
| 1. Reduced penetrance | Three evidence items on `genetic#SDHA`, a paragraph in its `notes`, a sentence on `inheritance`, and a caveat on genetic counselling |
| 2. LVNC omitted on a wrong rationale | Curated as `FREQUENT` on **HP:0030682**; the incorrect note is corrected in place rather than deleted |
| 3. Only 4 phenotypes | Now 8 |
| 4. Sudden cardiac death overstated | Corrected — two-thirds died of cardiac failure, one sudden death |
| 5. GeneReviews baseline | Search recorded in `notes` with the queries and why the generic chapter is not cited |

On finding 1, the paper goes further than "unaffected homozygote". The father's complex II activity was **decreased by 42%, resembling the patients rather than the heterozygous mother or controls**. So neither the genotype nor the measurable enzyme deficit in an accessible tissue predicts who becomes ill — which is a stronger statement about this disease than nonpenetrance alone, and it now drives the counselling advice.

On finding 2, one correction to the review: **HP:0031689 is not in the `PhenotypeTerm` enum cache.** `HP:0030682` "Left ventricular noncompaction" is, and is what I bound.

On finding 3, I curated the six that the cached text supports with an exact quote, and deliberately not the three whose only source is Table 1 — respiratory distress, mitral valve insufficiency and per-patient exercise intolerance. Table cells did not survive text extraction into the cache, so quoting them is exactly the fabrication risk the snippet rules exist to prevent. That decision is recorded in `notes`. I also added a phenotype for the **normal** neuromuscular examination and psychomotor development, unbound, since it is the finding that separates this entity from Leigh syndrome and HPO has no term for a normality.

Suggestions: dropped the empty `DOI_10.1038_ejhg.2010.83.md` (item 6); added cardiac transplantation with a quotable indication from the paper's own background, and cascade-screening guidance shaped by the penetrance finding (item 7); replaced the generic-background snippets on the heart-failure nodes with cohort-specific sentences now available in the full text (item 8). I did not add the full guideline-directed heart failure list — those are class-level management with no source tying them to this entity, and this entry is better as 15 patients' worth of specific content than padded to the length of its siblings.

Two further fixes from an independent read:

- The animal model's `genes:` recorded SDHA while the model deletes *Sdhaf4* — a machine-readable value contradicting the `limitations` beside it. Now `hgnc:20957` SDHAF4, matching KB convention (the ADGRG1 entry carries a Col3a1-null mouse under `Col3a1`; Adams-Oliver carries a Rac1 deletion under `RAC1`).
- Added `name: Sdhaf4 conditional-null mouse`. The genotype-derived fallback label would have been a 130-character string naming two Cre drivers.

The branch was also rebased onto current `main` — `check-snippet-length` was failing on `Hyperprolinemia_Type_2.yaml`, which this diff never touches, because #10234 re-cut those snippets after my branch point.

**Gates, all run to completion:** `validate-disorders` 0 (134/134 snippets), `validate` 0 (54/54), `validate-terms` 0, `validate-history` 0, `check-duplicate-keys` 0, `check-entity-refs` 0, `check-snippet-length` 0, `check-title-snippets` 0, `check-snippet-grading` 0, `check-environmental-evidence` 0, `check-folded-hyphens` 0, `check-term-cache-integrity` 0, `check-cache-order` 0.

One cache row added, `hgnc:20957,SDHAF4`, generated by `just validate-terms` and normalised — not hand-written.

PR #10255, opened 2026-08-31
Curate immunodeficiency 25 (MONDO:0012426)

### LINKED ISSUE #10235: Curate immunodeficiency 25 (MONDO:0012426)
Curate a dismech entry for **immunodeficiency 25** ([MONDO:0012426](https://monarchinitiative.org/MONDO:0012426)) — an autosomal recessive T-cell immunodeficiency caused by biallelic loss-of-function variants in `CD247`, the gene encoding the TCR CD3-zeta chain, which is required to assemble and surface-express the T-cell receptor complex.

Preflight notes: the stub is a MONDO leaf (no descendants recorded) with a single causal gene, `hgnc:1677 CD247`. No entry in `kb/disorders/` or `kb/groupings/` on `origin/main` mentions `CD247` or this MONDO ID, and no open PR covers it. It is a distinct genetic entity from the neighbouring CD3-subunit defects (IMD18/`CD3E`, IMD17/`CD3D`, IMD19/`CD3G`), which are separately catalogued.

Stub: `stubs/Immunodeficiency_25.yaml`

aleixpuigb:
Adds a new disorder entry for IMD25 — CD247/CD3ζ deficiency.

The T-cell receptor only reaches the cell surface as a complete eight-chain assembly, and CD247 supplies the invariant ζζ dimer. That dimer does two separable jobs: it is needed to build and export the receptor, and it carries three of the complex's ten ITAMs — more than any other subunit. Losing it therefore costs both receptor density and signalling capacity, and thymic selection fails in both directions at once: too few cells complete development (profound T-cell lymphopenia), and self-reactive cells that should have been deleted are not (the autoimmunity seen in longer-surviving patients).

## Pathograph

```
biallelic CD247 loss-of-function variants
        ↓
impaired TCR/CD3 complex assembly and surface expression
        ↓
loss of ITAM-dependent proximal TCR signaling
        ↓
failed thymic selection and T-cell development
        ↓
T-cell lymphopenia with impaired function

[ Revertant somatic mosaicism ]  ← no downstream edge, see below
```

**Revertant mosaicism is the normal state here, and it is curated as a modifier rather than a cause.** In this gene reversion is not a curiosity — it has been found in 4 of 5 reported patients. A second spontaneous change at the patient's own germline codon partially restores the protein in a minority T-cell population. What makes it scientifically interesting is that the rescue is lopsided: surface receptor comes back, ZAP70 phosphorylation does not, and the revertant cells have not improved any reported patient's clinical course. That is a natural experiment separating TCR expression from TCR function, and it says something uncomfortable about any therapy aimed at restoring receptor density alone.

I gave it a pathophysiology node with **no `downstream` edge**, because it modifies the disease course without being a step toward the phenotype. Curating it as a causal step would imply it contributes to the immunodeficiency, which is backwards.

**The mouse does not model this disease, and that is recorded as a finding rather than omitted.** The Q70X allele restores surface TCR in the classic murine MA5.8 CD247-deficient line and fails to do so in human cells. The direction of the error matters: the model makes a damaging human allele look tolerated. It is curated as `FAILS_TO_RECAPITULATE` with a `HUMAN_MODEL_MISMATCH` discussion.

The discussion also notes that this is not a quirk of CD247. Mouse and human CD3 chains already have partly *reciprocal* roles — CD3δ loss permits some T-cell development in mice but causes SCID in humans, and CD3γ does the reverse. So for this complex the mismatch should be assumed and checked, not rediscovered subunit by subunit.

## Where I deliberately hedged

The second `genetic` entry, for heterozygous nonsense CD247 alleles, uses `relationship_type: DISPUTED`. Three patients carrying such alleles had signs of immunodeficiency or autoimmunity, and the alleles clearly suppress wild-type receptor expression and signalling when transduced into Jurkat cells. But that is measured under overexpression, no segregation or penetrance data are reported, and the authors themselves state the clinical correlation only as a possibility. `CAUSATIVE` would say more than the source does. An `OPEN_QUESTION` discussion records what would settle it.

## Research input — and what it caught

One openscientist deep-research report, committed under `research/`. Preflight (`just preflight-dr … MONDO:0012426`) PASS: CD247 dominant at 54 mentions, OMIM 610163 matching.

This is the case in the run where the deep-research report earned its keep rather than just confirming what I had. My initial PubMed search surfaced the two recent Regueiro-group papers but **missed the founding case** — Rieux-Laucat et al., *NEJM* 2006 (PMID:16672702), the four-month-old boy with homozygous Q70X and three heterozygous somatic second-site changes. The report found it. I fetched it, verified the quoted sentences against the cache, and added it, along with PMID:28743717, which distinguishes true reversion from second-site compensation and supplies the *measured* mutation-propensity explanation for why reversion happens so often in this gene.

## Validation

| check | result |
|---|---|
| `just validate` (schema + references) | pass, **42/42** snippets verified against cached references |

## Scope caveat

The published series total a handful of patients. Severity, frequency and treatment statements describe the reported series, not an established disease spectrum, and the `notes` say so.

Closes #10235

---

### DIFF
--- a/kb/disorders/Immunodeficiency_25.yaml
+++ b/kb/disorders/Immunodeficiency_25.yaml
@@ -565,8 +565,8 @@ phenotypes:
   name: T-B+NK+ Immunophenotype
   description: >-
     Circulating T cells are profoundly reduced while B cells and NK cells are
-    present, giving the T-B+NK+ pattern that separates this from the T-B-
-    recombination defects. The NK compartment is present but not normal: an
+    present, giving the T-B+NK+ pattern that separates this from
+    the T-B- recombination defects. The NK compartment is present but not normal: an
     unusual CD56-negative CD16-positive population with diminished cytolytic
     activity has been described, so "NK preserved" describes cell number rather
     than cell function.


aleixpuigb:
All six IMPORTANT items and the suggestions are addressed. Pushed in 1f430c4d6.

**Items 1–3 were one mistake with three consequences, and you traced it correctly.** PMID:17170122 was fetched and never cited, and everything else followed from that: the exon 7 allele missing, the protein-instability step missing, and the T-B+NK+ claim propped up by a snippet that mentions neither B nor NK cells. That last one is the worst of the three — it is not an under-citation, it is an evidence item that does not support its claim, which is the same defect I got caught on in the sibling dHMN5B PR. Both are now fixed from the right sentence.

Your qualification on NK cells is a real correction, not a nicety. I wrote "NK cells are preserved" meaning *present*, and the paper reports a CD56⁻CD16⁺ population with diminished cytolytic activity. Preserved number, abnormal function. Bound to HP:0025808 and stated in the description.

**Item 4 is the one I am most glad you caught, because the correction makes the entry more useful rather than just more accurate.** I had written the mouse mismatch as a blanket "MA5.8 does not model human CD247 deficiency". That is wrong in a way that costs a reader information: MA5.8 scores the frameshift allele *correctly* — it fails to rescue there exactly as it does in the patient. What the line gets wrong is the retained-truncated-chain class.

So the rule is allele-specific, and now stated that way: a murine reconstitution result can be trusted for an allele whose product is degraded away, and cannot be trusted for one that leaves a partial chain behind. That is something a curator can act on. "The mouse is unreliable" is not.

**Item 5.** Both added. On reduced surface TCR/CD3 density — you were right that HPO has nothing for it. I searched and the nearest terms describe T-cell *counts*, which is a different measurement and would have misstated the finding, so it is recorded with a free-text `preferred_term`, no binding, and a `notes` line saying what was searched and why binding broader would be wrong.

**Item 6.** Live-vaccine deferral added. Your framing of it as time-critical is the right one — it is the intervention that is *lost* if diagnosis is late, which is different from a treatment that merely works better early. Both sources are SCID-class rather than CD247-specific, so both are `INDIRECT` with that stated in the explanation.

**Suggestions.** TREC screening added, with the caveat you would expect me to raise given the rest of the entry: reversion is the usual state of affairs in this gene, so a partly restored T-cell compartment could in principle lift the TREC value the way ZAP70 deficiency escapes the assay. I recorded that as a stated concern, not an observed failure — it has not been reported in a CD247 patient. Prevalence added as `CASES_IN_LITERATURE` / `ULTRA_RARE`, framed as a count of the literature rather than a rate.

I did not add the PM1T line as an `experimental_models` entry or an `onset` value. PM1T appears in PMID:40711587 only as a named cell line without characterisation I could quote, and no source I have gives an onset age beyond "early childhood" in prose. Both are real gaps; I would rather leave them visible than fill them with something I cannot cite. Say the word if you would prefer them in with weaker evidence.

Thank you for the note on PMID:27555457 — I confirmed the same thing (`content_type: unavailable`, empty Content) and left it alone. Flagging unquotable caches explicitly saved a wasted pass.

Snippets 42/42 → **55/55**. `validate-disorders`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys` and `check-snippet-grading` all pass.

Ready for re-review.

---

aleixpuigb:
`test (3.13)` was red on `c85673b12`. Fixed in eeaaf3c75 — two causes, one mine.

**Mine.** The T-B+NK+ phenotype description I added when responding to your review wraps `T-B-` onto a fold boundary, so the YAML folded scalar renders it with the hyphen suspended. `check_folded_hyphens.py` caught it. Reflowed so the notation stays on one line. I did not touch the baseline — this is a defect the current change introduced, not a pre-existing one, and grandfathering it would be exactly the thing CLAUDE.md warns against.

Worth naming the pattern rather than just the instance: **this is the second folded-hyphen split I have shipped in this batch**, both in prose I added while responding to review, and both caught by CI rather than by me. The cause is that after each edit I re-ran the file-scoped checks — `validate`, `validate-terms`, `count-verified-snippets`, `check-entity-refs` — and ran the ungated whole-KB ones only after the first large change. The whole-KB checks are precisely the ones that catch prose defects. I have started running the full set before every push.

**Not mine, but mine to clear.** `check-snippet-length` was failing on four sub-5-word snippets in `Hyperprolinemia_Type_2.yaml`, which this branch does not touch. The cause is not a shallow-checkout artifact: the check baselines against `origin/main`, and PR #10234 re-cut exactly those four snippets on `main` after this branch was created. So the branch was carrying the *old* file and being compared against the *new* baseline. Merging `main` in resolves it, and the check now passes.

The merge also brings in `db7896685` and its new ungated whole-KB `check-enum-values` gate. This entry passes it.

Verified after the merge, in the worktree, before pushing:

| check | result |
|---|---|
| `just validate` | pass |
| `just count-verified-snippets` | 55/55 |
| `just check-enum-values` | pass |
| `just check-folded-hyphens` | pass |
| `just check-snippet-length` | pass |
| `just check-title-snippets` | pass |
| `just check-snippet-grading` | pass |

Diff scope against `main` is unchanged: 24 additions, 1 deletion (the stub), 5 cache modifications.

---

aleixpuigb:
Thanks — and noted on the four residuals. I am deliberately not pushing them.

The PR is approved on the current head, and any push would dismiss that approval under branch protection. Trading an approval for four wording and bookkeeping fixes that the review itself marked non-blocking is a bad trade, so I would rather leave it.

For the record, so nobody thinks they were missed, and so they are cheap to pick up if a maintainer does want them:

1. `phenotypes[T-Cell Lymphopenia].description` line 656 — trim "NK cells are preserved, giving a T-B+NK+ pattern" to the T-cell claim the node actually evidences, since the `T-B+NK+ Immunophenotype` node ninety lines above now carries the qualified version.
2. MA5.8 `modeled_mechanisms[0].limitations` lines 837–842 — add a clause noting the line scores true nulls correctly, to match the now allele-specific discussion rationale.
3. A history record for the two review-response sessions.
4. `prevalence.notes` — either cite the "roughly 1 in 50,000" SCID newborn-screening figure or drop the number.

Say the word on any of them and I will push; otherwise they are a clean follow-up PR for whoever picks this entry up next.

---

PR #10256, opened 2026-08-31
curate: Add Limb-Girdle Muscular Dystrophy AR 28 (MONDO:0957270, HMGCR)

### LINKED ISSUE #10244: Curate muscular dystrophy, limb-girdle, autosomal recessive 28 (MONDO:0957270)
Curate a dismech entry for **muscular dystrophy, limb-girdle, autosomal recessive 28 (LGMDR28)** ([MONDO:0957270](https://monarchinitiative.org/MONDO:0957270)) — biallelic *HMGCR* variants; loss of HMG-CoA reductase function in muscle, the same enzyme statins inhibit, which makes the pharmacological parallel part of the mechanism rather than a footnote.

**HMGCR appears in four existing entries** — `Microcephalic_Osteodysplastic_Primordial_Dwarfism_Type_II`, `Osteonecrosis`, `Schnyder_Corneal_Dystrophy`, `Spastic_Paraplegia_89_Autosomal_Recessive` — but as a *pathway* gene, not as this disease. That is not prior coverage. Keep statin-myopathy and anti-HMGCR immune-mediated necrotizing myopathy clearly separated from the genetic disease; they are distinct entities that share a protein.

Stub: `stubs/Muscular_Dystrophy_Limb-girdle_Autosomal_Recessive_28.yaml`

kevinschaper:
Adds an entry for limb-girdle muscular dystrophy autosomal recessive 28 (LGMDR28), a muscular dystrophy caused by inheriting two damaged copies of *HMGCR* — the enzyme statins block. It is a metabolic disease that happens to look like a limb-girdle dystrophy, which is what makes it worth curating separately: about 28 patients have been described worldwide since 2023, and unusually for an LGMD subtype there is already a candidate treatment aimed directly at the missing chemical.

Closes #10244

## Why this is its own entry and not a subtype

`entry_type: DISEASE`.

The KB already has `Autosomal_Recessive_Limb-Girdle_Muscular_Dystrophy.yaml` (MONDO:0015152), which is the MONDO parent of this disease and carries eight subtypes. I did not add LGMDR28 to it, because that entry's own description says what holds its subtypes together:

> "The shared mechanism is myofiber membrane (sarcolemmal) fragility and failed repair leading to repeated myofiber degeneration…"

That is true of calpain, dysferlin, the sarcoglycans, anoctamin and FKRP. It is not true here. In LGMDR28 the muscle membrane is intact — staining for dystrophin, the sarcoglycans and dysferlin is normal in every reported cohort — and what fails is the supply of mevalonate. Putting it inside that entry would place a different pathograph under a description that asserts the opposite one. Instead the new entry names the parent in `parents:`, so the clinical classification survives without importing the mechanism claim.

The automated second opinion on the issue reached the same conclusion independently, for the same reason.

## Three diseases share this protein, and only one of them is this one

This was the main risk in the curation, and it is a same-protein problem rather than a same-name one. A literature search on *HMGCR* returns mostly the other two:

| Entity | Cause | In this PR |
|---|---|---|
| **LGMDR28** | two damaged inherited copies of *HMGCR* | the entry |
| Statin-associated myopathy | the drug blocking the same enzyme | differential, MONDO:0100637 |
| Anti-HMGCR necrotizing myopathy | an autoantibody against the same protein | differential, MONDO:0016098 |

How they are kept apart:

- **Ontology check.** `runoak relationships --direction down -p RO:0004003 HGNC:5006` returns exactly one disease term, MONDO:0957270. The other two are not germline *HMGCR* disorders and carry their own MONDO identities.
- **A direct test in the patients.** The founding family was tested for anti-HMGCR antibodies and had none — that is quoted in the autoimmune differential.
- **Papers excluded rather than mined.** PMID:30588482, PMID:32447205 and PMID:35754285 all report anti-HMGCR myopathy *presenting as a limb-girdle dystrophy*; they are the easiest route to a wrong attribution and none is used as evidence about LGMDR28. PMID:36447272 is dysferlinopathy, a different subtype.
- **One statin result is used on purpose.** PMID:36745799 also rescued statin myopathy in mice; it appears only in the treatment rationale and the statin differential, labelled as such.

The mouse models needed the same care. Both muscle-specific *Hmgcr* knockouts are described by their authors as models of *statin* myopathy, but the genotype is a genetic loss of the enzyme in muscle, which is closer to this disease than to drug exposure. They are curated with that framing mismatch written into `limitations`.

## Mechanism

```
reduced HMG-CoA reductase activity (MOLECULAR)
  └─> mevalonate pathway flux reduction (MOLECULAR)   [~20% of normal in a patient]
        ├─> isoprenoid depletion, impaired Rab geranylgeranylation (MOLECULAR)
        │     └─> myofiber mitochondrial dysfunction (CELLULAR)
        ├─> impaired myoblast proliferation and myotube fusion (CELLULAR)
        └───────> progressive myofiber loss and fatty replacement (TISSUE)
                    ├─> proximal and axial weakness
                    ├─> elevated creatine kinase
                    └─> respiratory muscle weakness
```

The interesting part is which branch matters. Cholesterol appears not to be it — patients' cholesterol is low but normal — whereas in knockout mice the muscle disease is reversed by geranylgeraniol, pointing at the isoprenoid arm. That is entirely mouse evidence, and it is recorded as a knowledge gap rather than asserted for humans.

## Two things I corrected after the automated second opinion

Both were real, and I verified each before changing anything.

1. **I had overstated the absence of cardiac disease.** The 2026 cohort says "None of the patients had any cardiac involvement", which I quoted. But PMID:36745799 reports slightly raised troponin T in all four patients tested and ECG conduction changes in two. Overt cardiomyopathy is genuinely absent; the myocardium is not untouched. Added a `Subclinical Cardiac Involvement` phenotype and requalified the description and the differential.
2. **The cell model was mislabelled.** I had it as a primary culture; the paper's methods say C2C12 from ATCC, so it is `CELL_LINE`.

I also added exercise-induced myalgia (the first symptom in the founding family), and a knowledge gap about a single HPO-curated observation of reduced alpha-dystroglycan, which if real would partly reconnect this disease to the dystroglycanopathies.

## Deep research

The falcon run is committed. `just preflight-dr` returns **WARN**, not PASS: it flags CERT1 at 17 mentions against 55 for HMGCR.

I read it rather than discarding it, and the flag is benign. CERT1 appears only inside the report's own caveats, which say that Open Targets returns CERT1 for this disease as an artifact of CERT1 sitting next to *HMGCR* on chromosome 5q13, and that it must not be treated as causal. That is independently confirmed by the founding family's linkage interval, which spans 3.2 Mb of 5q13 and contains exactly one candidate variant — in *HMGCR*, not CERT1. So the report is kept, but it contributed no content: it retrieved only a 2023 conference abstract and states it could not recover variant-level data. Everything in the entry comes from the primary papers.

## Deliberately not curated

- **No clinical trial.** PMID:36745799 says trials "are in place", but publishes no registry number. ClinicalTrials.gov returns zero studies for both "mevalonolactone" and "HMGCR muscular dystrophy", and `just ictrp-fetch` needs a trial ID rather than a search term. The entry records that the trial could not be located — not that none exists.
- **No statin-avoidance treatment.** Nothing published reports statin exposure in anyone carrying an *HMGCR* disease allele, and no source recommends avoidance. The expectation that it would be harmful is an inference, so it is a knowledge gap instead of a treatment.
- **Finer biopsy findings.** HPO curates type 1 fibre predominance, central nuclei and increased endomysial connective tissue from the discovery paper's full text; our cache for it is abstract-only, so there is no quotable sentence. They are recorded in `histopathology.notes` as a pointer for a curator with full-text access, and the note warns that "non-specific" in that abstract should not be read as "normal".
- **No Orphanet citation — corrected, see below.** ORPHA:653725 exists as a MONDO xref. I first wrote that it could not be cached; that reason was wrong and is retracted in a comment on this PR. The record builds fine and is available to add if a reviewer wants it.
- **A mevalonate-pathway mechanism module** would give this disease, statin myopathy and *HMGCS1* rigid spine syndrome a shared `conforms_to` target. There is no such module today. That is a follow-up, not part of this PR.

## Validation

All run to completion; every exit code checked.

| Check | Exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just validate-disorders` | 0 |

**85 of 85 evidence snippets verified** against the cached references.


kevinschaper:
Rebased onto current `main` and pushed a small addition.

**Why CI was red:** `check-snippet-length` failed on four snippets in `kb/disorders/Hyperprolinemia_Type_2.yaml`, a file this PR does not touch. The branch was cut from `1866ae91` and was four commits behind; #10234 ("Re-cut 43 line-fragment snippets") had since re-cut exactly those snippets on `main`. Because that check compares the working tree against the *current* `origin/main` baseline and grandfathering is disabled when checking against a ref, the pre-fix versions in my stale tree read as new violations. Rebasing onto `20faacff` picks up the fix; the check now passes.

**Also added in the same push:** a *Drosophila* `hmgcr` knockdown model. It comes from PMID:39823152, which this entry already cites six times, and I had left it out. It is curated as `FAILS_TO_RECAPITULATE` — both ubiquitous and muscle-specific knockdown are lethal, so there is no muscle phenotype to observe. It matters because it makes the "every model of near-complete loss is unviable" pattern hold across two phyla, which is the argument that the human alleles have to be hypomorphic rather than null. The `HUMAN_MODEL_MISMATCH` discussion was updated to say that rather than the mouse-only version it had before.

Re-ran everything on the new base, all exit 0: `validate`, `validate-terms`, `validate-disorders`, `validate-history`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`. Snippets now **86 of 86 verified**.

The diff is unchanged in scope: one entry, its history record, the eight references it cites, the nine cache rows it introduces, its deep-research artifact, and its own stub deletion.

kevinschaper:
Retracting one claim in the PR body. I wrote that ORPHA:653725 could not be cited because `data/orphadata/en_product1.xml` is gitignored and absent. **That reason was wrong**, and it is the exact failure mode this batch keeps hitting: asserting the repository cannot supply something without running the command that supplies it.

What actually happens: `just refresh-orphadata` **does** download `en_product1.xml` (54 MB, retrieved fine) and only then aborts on a stale checksum pin —

```
RuntimeError: checksum mismatch for en_product1.xml after download:
got df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb,
expected fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
```

The file is left in place, so `just structured-rebuild-orphanet --id 653725` then succeeds and writes a valid record. So the blocker is a drifted pin in `data/orphadata/MANIFEST.yaml`, not a missing file. Credit to the cblA agent on this batch, which hit the same wall and found the path through it.

I have **not** added the citation to the entry, for two reasons rather than one:

1. What that record can support is thin. Because the refresh aborted before the product4/6/9 files, the built record holds only synonyms and a cross-reference table — no definition, phenotypes or epidemiology. The strongest quotable row is `| MONDO:0957270 | Exact |`, which corroborates the Orphanet-to-MONDO identity. The entry's `notes` already records that mapping, verified independently with OAK, so the citation would restate a checked fact rather than add one.
2. `claude-review` is currently QUEUED on this PR, and pushing would cancel it. Not worth spending a review slot on a marginal citation while the runners are saturated.

If a reviewer wants it cited, say so and I will add it in the same push as any other requested change — the cache file is already built locally.

Separately, the stale `MANIFEST.yaml` pin looks worth its own issue: every agent that needs an Orphanet record hits it, and the traceback points at a checksum rather than telling you the file you need is sitting there. I have not touched the pin, since re-pinning a bulk-data checksum is a decision for a maintainer and not something to slip into a curation PR.

PR #10258, opened 2026-08-31
Clear the PARTIAL residue from the three worst entries (#10061)

caufieldjh:
First tranche of the `PARTIAL`-prose residue from #10061.

## The backlog

#10003 retired `supports: PARTIAL`, migrated 11,804 enum values to `SUPPORT`, and **deliberately** left every `explanation` that justified the old grade in place — the curator's reasoning is still correct, only the token is dead. That leaves **~3,600 evidence items across 751 files** explaining a value the schema no longer has. No gate catches prose, so this is curation rather than lint.

This PR takes the three worst entries and finishes them completely:

| entry | items | gain `directness: INDIRECT` |
|---|---:|---:|
| `Noonan_Syndrome_6` | 41 | 40 |
| `VAMP2-Related_Disorder` | 27 | 4 |
| `Agnathia-Otocephaly_Complex` | 26 | 5 |

## The reason usually *was* a directness claim

`Noonan_Syndrome_6` is why it was worth starting here — it is unusually uniform. Nearly every caveat says a GeneReviews or Orphanet statement characterizes **Noonan syndrome as a whole** rather than the NRAS-defined NS6 subtype, or that an *Nras* mouse result is model-organism rather than human evidence:

> `Attributes the shared RASopathy phenotypic features … to common pathway dysregulation; PARTIAL because it is a review-level statement about the RASopathies collectively.`

becomes

> `Attributes the shared RASopathy phenotypic features … to common pathway dysregulation. Graded INDIRECT because it is a review-level statement about the RASopathies collectively.`

with `directness: INDIRECT` added. The reason is untouched — it was always describing the slot the curator did not yet have. The entry-level scope note, which documented the whole pattern in terms of PARTIAL, now documents it in terms of directness.

**Where the reason is something else** — a variable phenotype, an untested construct, an n-of-1 overclaim — the token is dropped and **no directness is asserted**. #10003 refused to manufacture 11,620 appraisals no curator made, and this follows that.

The sharpest of those is VAMP2's autonomic-response item, where the curator wrote *"PARTIAL rather than SUPPORT"* specifically to resist the source's own *"providing confirmation"* phrasing. There is no middle value left to express that, so the criticism survives as prose: the item carries the measurement and not the confirmation.

## Verified in both directions

A consistency check enforces that **no explanation claims INDIRECT without the item carrying the slot, and no item carries the slot without the explanation saying why.** It found 6 of the first and 2 of the second while I was working. Both are now 0.

Two blind spots worth recording, since they are the same class as the one caught on #10234:

- The mechanical pass only matched folded `explanation: >-` blocks, so **single-line scalars were invisible rather than reported as skipped** — the count of "rewritten + skipped" did not equal the file's stale count.
- It marked items as indirect internally but **never inserted the slot**, so prose claimed an appraisal the data did not carry.

Both were caught by checking the *result* rather than the intent. The 13 items the mechanical pass could not handle were finished by hand rather than left, so all three entries are at zero.

## Scope note for the remaining ~3,580

Coverage of the formulaic transform varies a lot per entry, which is why this is three files and not thirty: `Noonan_Syndrome_6` 41/41, `Agnathia-Otocephaly_Complex` 21/26, `VAMP2-Related_Disorder` 21/27, but `Fragile_X-Associated_Primary_Ovarian_Insufficiency` 2/27 and `Adult-Onset_Ataxia_and_Polyneuropathy` 5/27. KB-wide, seven phrasings cover ~75% of occurrences; the rest is free prose needing a read. Sweeping all of it through one regex table is how you ship a partial pass and call it finished.

## Validation

- `just validate-disorders` over all three — **439/439 snippets verified**
- `just validate-history-all` — 7,790 records, no issues
- `check-duplicate-keys`, `check-entity-refs`, `check-enum-values`, `check-folded-hyphens` — clean
- `check-snippet-grading` grandfathered against `origin/main` — no new divergences
- No `supports`, `snippet`, `reference`, `evidence_source` or ontology term changed

### DIFF
--- a/kb/disorders/Agnathia-Otocephaly_Complex.yaml
+++ b/kb/disorders/Agnathia-Otocephaly_Complex.yaml
@@ -178,8 +178,8 @@ mechanistic_hypotheses:
     evidence_source: HUMAN_CLINICAL
     snippet: Agnathia-otocephaly with or without associated anomalies may represent the severe end of a phenotypic spectrum related to variants in genes in the interacting SMAD/TGFB/BMP/SHH/FGF developmental pathways.
     explanation: >-
-      States the hypothesis in the authors' own words. The caveat is that it is offered
-      as a proposal from a single case rather than a demonstrated model.
+      States the hypothesis in the authors' own words. It is offered as a proposal from
+      a single case rather than a demonstrated model.
 pathophysiology:
 - name: PRRX1 Loss of Function
   description: >-
@@ -376,8 +376,8 @@ pathophysiology:
     evidence_source: HUMAN_CLINICAL
     snippet: We discuss the current molecular mechanisms implicated in 1st branchial arch patterning, particularly the impact of bone morphogenetic protein and fibroblast growth factor 8, and how these findings have the potential to explain the spectrum of abnormalities present in these 2 fetuses with agnathia without associated holoprosencephaly.
     explanation: >-
-      Implicates BMP4 and FGF8 signaling. The caveat is that the mechanism is inferred
-      from two affected fetuses rather than demonstrated in them.
+      Implicates BMP4 and FGF8 signaling. The mechanism is inferred from two affected
+      fetuses rather than demonstrated in them.
   downstream:
   - target: Failure of Mandibular Arch Skeletogenesis
     description: >-
@@ -407,9 +407,8 @@ pathophysiology:
       evidence_source: HUMAN_CLINICAL
       snippet: A generalized disturbance in cell migration from the primitive streak may be its pathogenesis.
       explanation: >-
-        The only published mechanism linking the craniofacial and laterality defects.
-        The caveat is that it is explicitly proposed ("may be") rather than
-        demonstrated.
+        The only published mechanism linking the craniofacial and laterality defects. It
+        is explicitly proposed ("may be") rather than demonstrated.
 - name: Failure of Mandibular Arch Skeletogenesis
   description: >-
     Loss of first-arch patterning and neural-crest skeletogenesis prevents
@@ -578,8 +577,8 @@ pathophysiology:
       evidence_source: HUMAN_CLINICAL
       snippet: microphthalmia/anophthalmia, cleft palate, pharyngeal and laryngeal hypoplasia
       explanation: >-
-        Cleft palate is a reported associated anomaly of the arch defect. The caveat is
-        that the review does not assert it is an obligate consequence.
+        Cleft palate is a reported associated anomaly of the arch defect. The review
+        does not assert it is an obligate consequence.
 - name: Failed Ventrolateral Migration of the Ears
   description: >-
     Normal lateral and upward migration of the developing external ears depends
@@ -713,9 +712,8 @@ pathophysiology:
     explanation: >-
       The agnathia-otocephaly report itself attributes the holoprosencephaly phenotype
       to genes acting in the TGF-beta, hedgehog and FGF pathways, connecting the
-      SHH/smoothened arm to this specific complex. The caveat is that it is quoted
-      second-hand from a holoprosencephaly cohort study rather than measured in
-      agnathia-otocephaly cases.
+      SHH/smoothened arm to this specific complex. It is quoted second-hand from a
+      holoprosencephaly cohort study rather than measured in agnathia-otocephaly cases.
   downstream:
   - target: Holoprosencephaly
     evidence:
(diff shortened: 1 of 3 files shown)


caufieldjh:
CI was red for a reason that had nothing to do with this PR's content, and the shape is worth recording because it will recur.

## What failed

`check_snippet_length.py --against-ref origin/main` reported four "new" snippets under five words, **all in `Hyperprolinemia_Type_2.yaml` — a file this PR does not touch**:

```
pathophysiology[2].evidence[1]: 'unique endogenous vitamin antagonist'
phenotypes[2].evidence[0]:      'Furthermore, a low vitamin-B6'
biochemical[0].evidence[0]:     'Laboratory investigation revealed significantly'
genetic[0].evidence[3]:         'So far only four'
```

Those are exactly four of the sub-line fragments **#10234 repaired**. They are not in this branch's diff and they are not on `main`.

## Why

This branch's merge base was `b83a735243`, one commit before #10234 merged. So the tree CI checked still held the *old* Hyperprolinemia entry, while `--against-ref origin/main` fetched the baseline from a `main` that already had #10234's repairs. The four short snippets existed in the tree and were absent from the baseline, which is precisely the definition of "newly introduced" that the ratchet gates on.

Neither side is wrong. The ratchet compares a **tree** against a **live ref**, and those two moved apart in the ~20 minutes between this branch being cut and #10234 merging. It is the same class as the CI scope-check race, and the same class this PR series has been fixing in the data: a check that is correct about what it measures and misleading about what it means.

## Fix

Rebased onto current `main` (`ee112bde`). The tree now carries #10234's repaired entry, so the four snippets are gone rather than grandfathered — which is the right outcome, not a suppression.

Re-ran all three ratchets locally against `origin/main` exactly as CI does:

```
OK: no new snippets under 5 words (1574 occurrence(s) of 1331 distinct snippet(s) grandfathered)
OK: no new title-quoting snippets (343 occurrence(s) of 280 distinct)
OK: no new snippet-grading divergences (713 occurrence(s) of 332 distinct)
```

The diff against `main` is unchanged in substance — still only `Noonan_Syndrome_6`, `VAMP2-Related_Disorder`, `Agnathia-Otocephaly_Complex` and their three history records, 348 insertions / 213 deletions.

#10263 (tranche 2) is branched on top of #10234 and is not affected.

caufieldjh:
Both findings are correct and both are fixed in `97d0c8c6`. The second one is the more valuable review comment I've had on this series, so let me be specific about what it caught.

## `bounded` was my invention, and I didn't notice I'd invented it

You're right that it is a new undefined grading token sitting in the sentence position `PARTIAL` just vacated. Worse than you put it, from where I sit: I substituted it *mechanically*, in a table, precisely because I was thinking of the task as "remove a dead token" rather than "restate a limitation." A find-and-replace found a synonym, which is the one thing that could not possibly help. The check I'd built verified prose/slot consistency and would never have caught it.

All 22 removed. In every case the surrounding clause already carried the real limitation, exactly as you said — `"no mechanistic route is established, hence bounded"` → `"no mechanistic route is established"`. I kept the `and an indirect edge` clauses per your note, since those describe `causal_link_type`, and the three prediction-not-measurement items now say that is why `evidence_source` is OTHER.

## The three that denied their own `supports` value

Fixed to state what each item *does* support:

| was | now |
|---|---|
| `hence bounded rather than SUPPORT for a general recommendation` | `so it supports the observation having been made and not a general recommendation` |
| `so this is bounded rather than causal support` | `so it supports the association without establishing causation` |
| `The caveat is that it qualifies rather than supports the intervention` | `which bounds the indication rather than establishing efficacy` |

No `directness` added to any of them, per your call.

## Propagated the fix to the two tranches already in flight

Since you flagged this as precedent-setting at scale, I checked the other open tranches rather than letting them merge with the same defect:

- **#10263** (tranche 2) — 3 items denying their own `supports`. Fixed.
- **#10264** (tranche 3) — 34 `bounded` occurrences. Fixed.

**And removing the token in #10264 exposed a bigger error underneath it.** Reading all 34 to strip the word made it obvious that tranche had *under-applied* directness — the pattern table had fired on very few, but most of the reasons were plainly the inference steps CLAUDE.md names. 28 items now carry `directness: INDIRECT`: an IGE-wide cohort standing in for EGTCA, the pre-2022 "grand mal on awakening" entity standing in for the ILAE-2022 syndrome, a heterologous expression system standing in for human disease, a quotation grouping the marrow-failure consequences rather than naming the cytopenia asserted. Four keep a plain statement and no slot because their caveat is about strength — including the GWAS item recording "an absence of evidence at n = 499 rather than evidence that no such locus exists", which is exactly the distinction you and I are both protecting here.

So the token was hiding a real curation gap, not just reading badly. Thanks for pushing on it.

## On your two suggestions

- **Formulaic `The caveat is that ...` in Agnathia** — agreed, and I've stopped generating it as a default. Where it survives it now leads with what the item establishes.
- **The OTX2 denominator item** — noted, and I'll write the scope-vs-inferential-distance line into the next tranche's rationale if the pattern recurs. My working rule so far: scope narrowing (different population, different entity, different variant) counts as inferential distance because the quote reaches *this* claim only through the wider set; a pure sample-size or precision caveat does not. That item sits right on the line and I'd rather state the rule than defend the instance.

Validation after the fix: `just validate-disorders` ×3 (439/439 snippets verified), consistency check 0 in both directions, `just validate-history-all` clean. Two history records added.

PR #10259, opened 2026-08-31
curate: Add Autosomal Recessive Spondylometaphyseal Dysplasia, Megarbane Type (MONDO:0013223, PAM16)

### LINKED ISSUE #10243: Curate autosomal recessive spondylometaphyseal dysplasia, Megarbane type (MONDO:0013223)
Curate a dismech entry for **autosomal recessive spondylometaphyseal dysplasia, Megarbane type** ([MONDO:0013223](https://monarchinitiative.org/MONDO:0013223)) — biallelic *PAM16* (MAGMAS) variants; PAM16 regulates the TIM23 mitochondrial protein-import motor, so this is a skeletal dysplasia with a mitochondrial-import mechanism rather than a matrix or cartilage one.

Nothing anywhere in `kb/` mentions PAM16 or the Megarbane type, so this is a clean start. Note there is an unrelated open PR (#10205) about *spondylometaphyseal dysplasia with corneal dystrophy* — a different disease; do not conflate them.

Stub: `stubs/Autosomal_Recessive_Spondylometaphyseal_Dysplasia_Megarbane_Type.yaml`

kevinschaper:
New entry for autosomal recessive spondylometaphyseal dysplasia, Megarbane type — a bone dysplasia caused by a broken mitochondrial protein-import motor rather than by anything wrong with cartilage or collagen. PAM16 (also called MAGMAS) is the regulatory subunit of the machine that pulls proteins into the mitochondrion, and the reason the skeleton is what fails is simply where the protein is needed: it is expressed in the hypertrophic zone of the growth plate and in the trabecular bone just beneath it, the two places where cartilage is converted into bone.

Only five patients have ever been reported. Four Lebanese children from two unrelated families share one founder mutation and all died before age two; a fifth patient with a nearby mutation had a milder course.

Closes #10243

## The lump/split call: `DISEASE`

MONDO links exactly one disease term to PAM16. I checked directly:

```
runoak -i sqlite:obo:mondo relationships --direction down -p RO:0004003 HGNC:29679
→ MONDO:0013223 only
```

Both MONDO parents in the stub are grouping terms — `MONDO:0016763` spondylometaphyseal dysplasia and `MONDO:0800080` severe spondylodysplastic dysplasia — so there is nothing to demote this into. One gene, one coherent mechanism, its own entry.

**The two published mutations are one disease, not two.** The second patient's variant (p.Gln74Pro) sits two residues from the founder allele (p.Asn76Asp) in the same helix of the same domain. Both reviews that discuss it describe that patient as having the same disease with a milder course, not a different entity, so this is curated as an allelic series inside one entry rather than as two entries or as a parent with a subtype.

**There is no PAM16 spastic paraplegia.** I was asked to check for a separate neurological PAM16 entity and there isn't one. PubMed `PAM16 AND (spastic OR paraplegia)` returns zero records; Europe PMC `"PAM16" AND ("spastic" OR "paraplegia" OR "neurodegener")` returns only reviews of the import machinery, no case reports. The cardiomyopathy-with-ataxia phenotype that does exist in this part of the machine belongs to *DNAJC19*, PAM16's partner protein, and is curated here as a differential diagnosis rather than as part of this disease.

## The mechanism

```mermaid
graph LR
  A["PAM16 p.Asn76Asp<br/>protein destabilised<br/>(MOLECULAR)"] --> B["Matrix protein import<br/>through TIM23/PAM fails<br/>(MOLECULAR)"]
  B --> C["Mitochondria fragment,<br/>cells die<br/>(CELLULAR)"]
  B --> D
  C --> D["Ossification fails at growth-plate<br/>hypertrophic zone and spongiosa<br/>(TISSUE)"]
  C --> E["Myocardial energetic failure<br/>→ cardiomegaly<br/>(TISSUE)"]
  D --> F["Restrictive thoracic<br/>insufficiency (ORGANISM)"]
  F --> G["Death from respiratory<br/>insufficiency — family 1"]
  E --> H["Death from heart<br/>failure — family 2"]
```

The split at the bottom is the interesting part and it is why the cardiac and respiratory branches are separate nodes rather than one "lethal complications" node. All four children carried the identical homozygous allele on the identical haplotype, and two died of respiratory failure while two died of heart failure. That variability is recorded as an open knowledge gap rather than smoothed over.

## Separating this gene's several identities

The gene has two names and only one is safe to search on. "MAGMAS" as free text returns volcanology plus a sizable cancer literature using MAGMAS as a drug target in glioblastoma, ovarian carcinoma and pituitary adenoma. "PAM16" additionally collides with PAM4/PAM16 pulse-amplitude modulation in fibre-optics engineering. Every reference here came from a PAM16 search with a skeletal, Megarbane or mitochondrial-import qualifier, and each was read before use. None of the cancer or engineering literature is cited.

Deep research (falcon) passed `just preflight-dr` against MONDO:0013223: PAM16 mentioned 38 times, OMIM 613320 matching MONDO's, and the flagged "rival gene PAM" at 11 mentions is the PAM complex acronym from this disease's own mechanism, not a competing gene. **NEC verdict: pass.** The report is committed as the DR artifact, but the entry is curated from the primary sources — I caught one transcription error in the report (it gives the PROVEAN score as −24.021 where the paper says −4.021) and did not propagate it.

## Two patients I could not fully curate, and why

`just fetch-reference` returns `content_type: unavailable` with an empty body for both PMID:27354339 (the milder p.Gln74Pro patient) and PMID:41369754 (the only paper on spinal surgery in this disease). Both are paywalled and neither PubMed nor Europe PMC carries an abstract, so there is no quotable text and no evidence item can be built from either.

What I did instead: the p.Gln74Pro allele and its milder phenotype are curated from two reviews that describe that patient in quotable full text (PMID:33171986, PMID:39766801). The surgical experience does not enter the entry at all — I did **not** curate a spine-surgery treatment, because a treatment with no citable source and no citable indication is worse than its absence. It is recorded as a knowledge gap naming the paper and its DOI, so someone with institutional access can add it. Both papers are listed under `references` with the limitation stated.

Worth noting for whoever picks that up: the 2026 surgical report is from the same Buenos Aires group that reported the p.Gln74Pro patient (Fano V is an author on both), so it may be the only long-term follow-up of a survivor that exists.

## Other things deliberately left out

- **No module conformance.** There is no mitochondrial protein-import module in `kb/modules/`. The nearest candidates are wrong: `mitochondrial_dysfunction` and `loss_of_proteostasis` are both framed around age-related accumulation rather than a germline hypomorph, and `defective_skeletal_mineralization` is about rickets, a different mechanism. A `mitochondrial_protein_import_failure` module would have several conformers already (PAM16 here, *HSPA9* in EVEN-PLUS, *DNAJC19* in DCMA, *TIMM50*) and is worth a follow-up issue, but creating it is out of scope for this PR.
- **No datasets block.** Searching PAM16 in expression repositories returns the cancer drug-target literature, which is a different entity.
- **No `frequency` values on phenotypes.** With four to five patients total there is no denominator that would make a `FrequencyEnum` band meaningful, so the slot is left empty rather than filled with a number the sources do not support. Prevalence is `ULTRA_RARE` with `CASES_IN_LITERATURE` and no numeric rate, since Orphanet carries no epidemiology data for this disease.
- **Mortality phenotypes left unbound.** HP:0001522 and HP:0003819 are the right concepts for death in infancy and childhood, but they sit in HPO's clinical-modifier branch and are not members of this repo's `PhenotypeTerm` dynamic enum, so binding them fails term validation. Kept as free-text `preferred_term` with the reason recorded.

## One thing the model evidence disagrees on

Recorded as a `HUMAN_MODEL_MISMATCH` discussion rather than papered over. The yeast system shows the mutant protein at reduced levels, and an in-silico prediction agrees the substitution is destabilising. But the 2024 TIM23 review states that PAM16 disease mutations are *not* reported to lower steady-state PAM16 levels in patients, contrasting it with DNAJC19 where they are. Both can be true if the yeast destabilisation needs the heat stress that patients do not experience — but as it stands, the model result and the human observation point in different directions about the mechanism at the head of this pathograph.

## Validation

Every gate run to completion, exit code confirmed:

| Gate | Exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just count-verified-snippets` | 68/68 verified |
| `just validate-disorders` | 0 |
| `just check-term-cache-integrity` | 0 |

Cache rows were produced by `just validate-terms` and `just normalize-cache`; the diff is 17 added lines across 9 files, all CURIEs this entry introduced.

## Evidence grading

Graded by the cited study, consistently across the file: yeast complementation is `IN_VITRO` (matching the repo's convention — 103 of the KB's yeast-related items use it), mouse growth-plate immunostaining and the whole-animal fly work are `MODEL_ORGANISM`, the fly Schneider-cell measurements are `IN_VITRO` even though they come from the same paper, structure and stability predictions are `COMPUTATIONAL`, and patient phenotype and genotype are `HUMAN_CLINICAL`.

## Files

- `kb/disorders/Autosomal_Recessive_Spondylometaphyseal_Dysplasia_Megarbane_Type.yaml`
- `history/disorders/.../2026-08-31T174952Z-claude-code-58a385.yaml`
- 10 reference caches this entry cites, plus the 5 DOI caches the DR report resolved
- 9 cache CSVs (17 added rows)
- the falcon DR report and its citations sidecar
- deletion of `stubs/Autosomal_Recessive_Spondylometaphyseal_Dysplasia_Megarbane_Type.yaml`


kevinschaper:
All four IMPORTANT items are fixed in d9fdc7d, along with most of the suggestions. One correction to item 2 and one suggestion declined, both explained below.

## Item 2 — right diagnosis, but the suggested replacement points the wrong way

You are right that `GO:0140662 ATP-dependent protein folding chaperone` overstates it: PAM16 lacks the HPD motif and the chaperone is mtHsp70.

But `GO:0001671 ATPase activator activity` would assert the opposite of what PAM16 does. PAM16 *restrains* the J-domain of DNAJC19 so the motor's ATPase is regulated rather than constitutive. GO says so itself — UniProt Q9Y3D7 carries **`GO:0032780 negative regulation of ATP-dependent activity` by direct assay (IDA)**, and assigns human PAM16 no molecular-function term at all, which is consistent with your point that it is a regulator rather than an enzyme.

So I moved the annotation from `molecular_functions` to `biological_processes` and bound `GO:0032780`, with the reasoning in the descriptor's `description`.

Checking that annotation set turned up two more terms worth having:

- **`GO:0001405` PAM complex, Tim23 associated import motor** — I had left the PAM complex unbound saying no term in the constrained set represented it. That was wrong; the term exists. It is now a bound `cellular_components` entry instead of an unbound `protein_complexes` one.
- **`GO:0001503` ossification**, which GO assigns to PAM16 by IMP on the strength of the mutant phenotype in *this disease*. Added to the skeletal node, since the process term and the disease evidence are the same observation.

## Items 1, 3, 4 and the phenotype suggestions

- `locations` `preferred_term` is now `trabecular bone tissue`, matching the bound term, with the subcellular caveat kept in `description`.
- Added **Protuberant abdomen** `HP:0001538` and **Abnormal facial shape** `HP:0001999`.
- Also added **Short neck** `HP:0000470` and **Rhizomelia** `HP:0008905` from the Orphanet snippet, and **Aplasia/Hypoplasia of the sacrum** `HP:0008517` for the partial sacral agenesis. I used the aplasia/hypoplasia term rather than `HP:0010305 Absence of the sacrum`, which would overstate "partial agenesis"; the entry notes that the reported finding is the partial form.
- `reference_title` added to the ORPHA evidence items.
- The helix-II statement is regraded `HUMAN_CLINICAL` → `OTHER`. You were right that it is not a patient observation — it is a domain-architecture assignment carried over from prior structural work.
- PGT-M and targeted prenatal testing added to the genetic-counselling treatment, flagged as following from knowing the allele rather than from disease-specific evidence.

## Node splitting — done, and it improved the model mapping

Both bundled nodes split:

- `Mitochondrial Network Fragmentation and Cell Death` → **`Mitochondrial Network Fragmentation`** → **`Bioenergetic Failure and Cell Death`**. This turned out to be worth more than tidiness: the two model links now map onto the right nodes rather than both onto one. The yeast morphology readout targets fragmentation; the fly ATP and ROS readouts target the bioenergetic node, which is where the 60% ATP drop and the 65% cytochrome c oxidase reduction actually belong.
- `Myocardial Energetic Failure and Cardiomegaly` → **`Myocardial Energetic Failure`**, since cardiomegaly was already its downstream phenotype.

## Declined: the in-silico pathogenicity scores

I went to add these and found the cached text cannot support them. `references_cache/PMID_24786642.md` is a PDF-derived cache and renders the line as:

```
PROVEAN (deleterious effect, score 24.021)
```

The publisher's own XML for the same sentence gives **−4.021**. The minus sign did not survive PDF extraction and the digit differs, so any snippet I quoted from the cache would carry a wrong number into the KB. Nothing here is fixable by requoting — the value simply is not recoverable from the cached text — so the scores stay out.

Worth flagging beyond this PR: this is the same sentence where the falcon deep-research report gave −24.021, which I had recorded in the entry's `notes` as a report transcription error. It now looks more likely that the report inherited the corruption from the same PDF extraction path rather than inventing it. I have left the note as written, since the entry's claim is only that the report's number disagrees with the paper's, which is true either way.

## Also changed, from a separate QC read

- The **DCMA differential is now bound** to `MONDO:0012435`. Its MONDO label is `3-methylglutaconic aciduria type 5`, which is why searching the DCMA name does not surface it; the term carries `DCMA` and `DCMA syndrome` as exact synonyms and xrefs `OMIM:610198`.
- The *Drosophila* model's `genes` slot now names `black-pearl (blp)`, left unbound on purpose — it is a fly gene with no HGNC identifier, and binding the human PAM16 CURIE there would assert the model mutates a gene it does not.

EVEN-PLUS and "Other severe spondylodysplastic dysplasias" remain unbound. A direct MONDO search for EVEN-PLUS did not return a term, and the latter is deliberately a group rather than one disease.

`just validate` 0 · `just validate-terms` 0 · `just validate-disorders` 0 · `just validate-history` 0 · `count-verified-snippets` **76/76** · `check-duplicate-keys`, `check-entity-refs`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-term-cache-integrity` all 0.

New cache rows are 4 CURIEs (`GO:0032780`, `GO:0001405`, `HP:0008517`, `MONDO:0012435`), generated by `just validate-terms` and `just normalize-cache`.

PR #10260, opened 2026-08-31
curate: Add Methylmalonic Aciduria, cblA Type (MONDO:0009613, MMAA)

### LINKED ISSUE #10245: Curate methylmalonic aciduria, cblA type (MONDO:0009613)
Curate a dismech entry for **methylmalonic aciduria, cblA type** ([MONDO:0009613](https://monarchinitiative.org/MONDO:0009613)) — biallelic *MMAA* variants impair adenosylcobalamin cofactor delivery to methylmalonyl-CoA mutase, giving the most commonly B12-responsive form of isolated methylmalonic acidemia.

**First decide what this should be**, and here that is the substantive question rather than a formality. `kb/disorders/Methylmalonic_Acidemia.yaml` already covers this umbrella and carries **MMAA (cblA complementation group)** in its `genetic:` section, plus a `has_subtypes` entry for B12-responsive MMA. Against that, `kb/disorders/MMADHC-related_Disorder_of_Cobalamin_Metabolism_cblD_Type.yaml` is precedent for giving a complementation group its own entry.

Read both before choosing, and defend the choice in the PR body. If the answer is `SUBTYPE`, record it in the stub with the parent named in `notes` and delete the stub — that is a completed curation, not an avoided one.

Stub: `stubs/Methylmalonic_Aciduria_cblA_Type.yaml`

kevinschaper:
Adds a dismech entry for **methylmalonic aciduria, cblA type** (MONDO:0009613, biallelic *MMAA*), curated as its own Disease rather than folded into an existing entry. The short version: cblA is the form of isolated methylmalonic acidemia where the enzyme is fine and the cofactor hand-off is broken, which is why nearly every patient responds to vitamin B12 — and that mechanism was not curated anywhere in the knowledge base.

Closes #10245

## Why a Disease entry and not a subtype

This was the real question on the issue, so here is the argument on both sides.

cblA was already **listed** in two places and **curated** in neither:

| File | What it has for cblA |
|---|---|
| `Methylmalonic_Acidemia.yaml` | A `genetic:` record "MMAA (cblA complementation group)": two sentences of `features` prose, and evidence consisting of one review sentence that merely names cblA. No cblA pathophysiology node. No MMAA mechanism anywhere in the file. |
| `Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport.yaml` | A `has_subtypes` line "cblA", three lines of description, **no bound disease term, no evidence**. |

Neither file carries MONDO:0009613.

**For a subtype.** The umbrella already names MMAA, and MONDO does place cblA beneath methylmalonic acidemia — `runoak ancestors -p i MONDO:0009613` returns MONDO:0002012 in the closure. A separate entry duplicates material.

**For a Disease, which is what I did.** Three reasons:

1. **The pathograph is genuinely different upstream.** MMAA is not a mutase and never touches methylmalonyl-CoA. It is a G-protein that gates adenosylcobalamin loading onto the apo-mutase and then, during turnover, ejects damaged cofactor so a fresh one can go in. The umbrella's chain starts at mutase deficiency; this one starts at a failed protein-protein interaction and only converges downstream at methylmalonyl-CoA accumulation.
2. **The umbrella's existing subtype is the wrong shape to hold it.** `has_subtypes` there is `Vitamin B12-responsive` / `Vitamin B12-unresponsive` — *biochemical response classes spanning three genotypes* (its own description says cblA, cblB and some mut-). Recording cblA as a subtype would either overload that entry or require a fourth one on a different axis.
3. **Direct precedent.** `MMADHC-related_Disorder_of_Cobalamin_Metabolism_cblD_Type.yaml` is a complementation group with its own entry, and it did not delete the `cblD` line from the cobalamin umbrella when it was created. Same pattern here.

Neither umbrella file was modified.

## The mechanism

```
biallelic MMAA loss of function
            │
            ▼
loss of MMAA–mutase functional interaction   ◄── the decisive step
            │
     ┌──────┴──────┐
     ▼             ▼
AdoCbl gating   cofactor protection
onto apo-MMUT   and repair during turnover
     └──────┬──────┘
            ▼
reduced mutase holoenzyme activity  ──►  methylmalonate accumulation
                                              │
                              ┌───────────────┼───────────────┐
                              ▼               ▼               ▼
                        decompensation   renal injury   neurologic injury
```

The most useful finding here is that patient missense MMAA proteins keep essentially **normal intrinsic GTPase activity**. What they lose is mutase-stimulated activity, physical binding to the mutase, and the ability to gate cofactor transfer (PMID:28497574, 15 purified mutants). The 2023 human co-crystal structure puts the disease-causing alleles at the MMAA–MMUT interface itself (PMID:37468522). So cblA is a lost protein interaction, not a lost enzymatic activity — an unusual mechanism for an organic aciduria.

That also explains the clinical picture, which the umbrella records as an observation without the reason: because the apoenzyme is structurally intact and only delivery is inefficient, pharmacologic hydroxocobalamin partly rescues holoenzyme formation by mass action. 27 of 28 cblA patients in the European registry were cobalamin responsive.

## Named Entity Confusion

`preflight-dr` returns WARN, exit 0. Two warnings, both expected:

- *MMUT mentioned at 54% of MMAA.* Correct and wanted — the report's core content is the cblA-versus-mut comparison.
- *Report cites OMIM 607481, MONDO xrefs 251100.* 607481 is the **gene** MIM for MMAA; 251100 is the **phenotype** MIM. The report cites both correctly and preflight compares gene MIM against phenotype MIM. Not a defect.

Preflight keys on the gene and cannot separate cblA from the umbrella or from cblB/cblC, so I checked each clinical source's study population by hand:

- **cblA-specific or cblA-as-a-separate-arm** (the backbone of every clinical claim): PMID:32754920 (28 cblA), PMID:35618652 (23 adult cblA), PMID:31921599 (5 MMAA-genotyped).
- **Mixed cohorts, used only for between-group contrast**: PMID:17597648 (83 patients, only 20 cblA) and PMID:17957493. Where a percentage in a quoted sentence belongs to the mut0 or cblB subgroup, the `explanation` says so rather than letting it read as a cblA rate. No cohort-wide statistic is attributed to cblA anywhere.
- One quirk worth knowing: the string **cblA also names a *Burkholderia cepacia* cable-pilus adhesin gene** (PMID:11896764, PMID:12427939) and an unrelated engineering acronym. Excluded.

The deep-research report is on-entity and is committed as an artifact. It was not the source of the entry — I curated from primary literature while it ran — but it independently reached the same conclusions, including that no dedicated *Mmaa* animal model exists and that no cblA-specific prevalence is published.

## A MONDO bug found along the way

`MONDO:0009613` carries **"cobalamin B disease" as an EXACT synonym**. Cobalamin B disease is cblB — different gene (MMAB), different OMIM anchor (251110). The curation stub inherited the same error as "cblB - cobalamin locus b". I did not copy it into the entry's `synonyms`, and recorded why in `notes`. Worth reporting upstream; not fixed here.

- **No `animal_models`.** No published study characterises an *Mmaa*-deficient animal as a cblA model. That is a statement about the literature, not the reagents: a constitutive IMPC null allele **does** exist (`Mmaa em1(IMPC)J`, MGI:5825155) with broad MGI phenotype annotations, but MGI attaches no disease-model assertion, nothing is published on its metabolic phenotype, and consortium annotations are not a quotable evidence item. Recorded as a `KNOWLEDGE_GAP` rather than asserted absent.
- **No prevalence rate.** None specific to cblA is published; `ULTRA_RARE` with a `CASES_IN_LITERATURE` measure, sourced to the "<200 patients described" figure. The ~20% share of isolated MMA is on the gene record as a `case_fractions` entry, which is the genetic-spectrum slot, not a population rate.
- **No `mappings.mondo_mappings` narrowMatch added to the umbrella.** `parents: [Methylmalonic Acidemia]` already links the two, and a narrowMatch would assert umbrella coverage of a concept that now has its own exactMatch entry. Happy to add it if a reviewer disagrees.
- **The 2021 MMA/PA guideline (PMID:33595124) is not cited.** Its cblA-specific statement is in the full text; the cached abstract has nothing cblA-specific, so citing it would have been padding.

## Evidence grading

Held to one rule across the file, stated in `notes` so it can be checked: findings about patients or patient material (including diagnostic enzymology and biopsy histology) are `HUMAN_CLINICAL`; cultured-cell and purified-protein mechanism experiments (somatic cell complementation, recombinant biochemistry, crystallography, transcript stability) are `IN_VITRO`; GeneReviews is `OTHER`, matching the cblD entry. PMID:28497574 and PMID:17957493 report both kinds, so their evidence is split into separate items. PMID:19955418 is bacterial MeaB work and is cited for the conserved gating mechanism, never as a human cblA result — its `explanation` says so.

## Tooling problems hit

1. **`just refresh-orphadata` fails**: upstream `en_product1.xml` no longer matches the sha256 pinned in `data/orphadata/MANIFEST.yaml`. It downloads the file before failing, which is enough for `just structured-rebuild-orphanet --id 79310` to build a valid nosology and cross-reference record — committed and cited under `external_assertions`. It is not enough for the definition/phenotype/epidemiology sections, which come from bulk files the aborted run never fetched, so no Orphanet prevalence class was available. I have not repinned the manifest; that would rewrite all 347 committed ORPHA cache files and does not belong in a curation PR.
2. **`just fetch-reference ORPHA:79310` does not work at all** — it reports `No source found for reference type: ORPHA:79310`. `structured-rebuild-orphanet` is the path for this prefix. (The advice on the claim issue suggested the `fetch-reference` form.)
3. The deep-research client's reference-validation pass hung for over 20 minutes after the report was complete, with no output, and had to be killed. The report itself is intact.

## Validation

Every gate run to completion, all exit 0:

| Gate | Result |
|---|---|
| `just validate-disorders` | exit 0, **68/68 snippets verified** |
| `check-environmental-evidence`, `check-folded-hyphens` | exit 0 |
| `just check-stubs`, `just validate-history` | exit 0 |

Cache CSVs carry exactly six tool-generated rows, for the three CURIEs this entry introduces: `MONDO:0009613`, `GO:0031419`, `HP:0003210`.

### DIFF
--- a/kb/disorders/Methylmalonic_Aciduria_cblA_Type.yaml
+++ b/kb/disorders/Methylmalonic_Aciduria_cblA_Type.yaml
@@ -157,7 +157,7 @@ pathophysiology:
     proteins fail to have their GTPase activity stimulated by the mutase, and a subset can
     no longer bind it at all.
   molecular_functions:
-  - preferred_term: MMAA GTPase activity, in its mutase-stimulated rather than intrinsic form
+  - preferred_term: mutase-stimulated MMAA GTPase activity
     term:
       id: GO:0003924
       label: GTPase activity
@@ -670,7 +670,7 @@ phenotypes:
     term:
       id: HP:0002149
       label: Hyperuricemia
-  frequency: FREQUENT
+  frequency: OCCASIONAL
   evidence:
   - reference: PMID:35618652
     reference_title: Very long-term outcomes in 23 patients with cblA type methylmalonic acidemia.
@@ -691,7 +691,7 @@ phenotypes:
       id: HP:0002071
       label: Abnormality of extrapyramidal motor function
     severity: MILD
-  frequency: FREQUENT
+  frequency: OCCASIONAL
   evidence:
   - reference: PMID:35618652
     reference_title: Very long-term outcomes in 23 patients with cblA type methylmalonic acidemia.
@@ -780,18 +780,36 @@ problems, failure to thrive, hypotonia, and developmental delay"
 - name: Hyperammonemia
   category: Biochemical
   description: >-
-    Mild elevation of ammonia in a minority of cblA patients at diagnosis. This is
-    deliberately recorded as mild and occasional. The severe hyperammonemic encephalopathy
-    that GeneReviews describes belongs to the infantile non-B12-responsive form, that is
-    to mut0 and cblB, and is not attributed to cblA here.
+    Raised ammonia in a minority of cblA patients, usually mild but not always. Two of
+    five patients in a paediatric series had a mild elevation at diagnosis, while a
+    molecularly confirmed cblA patient in the adult cohort presented at nine days with
+    ammonia up to 500 micromol/l requiring haemofiltration, so the severe end is reached
+    in this disease and not only in mut0 or cblB.
+
+    The frequency band deviates from the arithmetic of its first source and the deviation
+    is deliberate. Two of five is 40%, which would fall in FREQUENT, but that point
+    estimate rests on five patients. The two larger cblA cohorts do not report
+    hyperammonemia as a cohort-level finding at all: it appears in the 23-patient series
+    only in the narrative of a single patient, and not once in the 28-patient registry
+    report. OCCASIONAL is the band consistent with all three series rather than with the
+    smallest one, which is the same reasoning applied to the optic-atrophy item.
+
+    Note on sourcing: the severe hyperammonemic encephalopathy that GeneReviews describes
+    is scoped there to the infantile non-B12-responsive form, naming mut0 and cblB, so
+    that sentence is deliberately not cited here. The evidence below is cblA-specific.
   phenotype_term:
-    preferred_term: Mild hyperammonemia
+    preferred_term: Hyperammonemia
     term:
       id: HP:0001987
       label: Hyperammonemia
-    severity: MILD
   frequency: OCCASIONAL
   evidence:
+  - reference: PMID:35618652
+    reference_title: Very long-term outcomes in 23 patients with cblA type methylmalonic acidemia.
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "This patient presented at 9 days of life with a severe acute decompensation with ketoacidosis and hyperammonia (up to 500 micromol/l) requiring hemofiltration"
+    explanation: A molecularly confirmed cblA patient (homozygous MMAA c.586C>T) with severe neonatal hyperammonemia, which is why this phenotype is not restricted to the mild form. It is a single narrated patient in a 23-patient cohort rather than a reported cohort frequency.
   - reference: PMID:31921599
     reference_title: Clinical picture and treatment effects in 5 patients with Methylmalonic aciduria related to MMAA mutations.
     supports: SUPPORT
@@ -1239,6 +1257,10 @@ treatments:
       term:
         id: CHEBI:6909
         label: metronidazole
+    - preferred_term: neomycin
+      term:
+        id: CHEBI:7507
+        label: neomycin
   target_mechanisms:
   - target: Methylmalonic Acid and Propionyl-CoA Accumulation
     description: >-


kevinschaper:
All five IMPORTANT findings are fixed at `d2cd8b1f56`, plus four of the six suggestions. I checked each claim against the cached references first; every quote you cited was accurate, including the line numbers. Snippet count is now **87/87 verified**.

**Finding 1 — the GTPase contradiction was real, and thank you for catching it.** The node asserted `GO:0003924 DECREASED` while quoting a paper that says 14 of 15 mutants keep wild-type-like intrinsic GTPase activity.

I removed the modifier rather than moving it, and the reason is worth stating because it constrains the fix. What cblA alleles lose is the *mutase-stimulated* form of the activity; the intrinsic hydrolysis is intact. `ModifierEnum` has no value meaning "intact but no longer activatable" — `DECREASED` is false, and `LOSS_OF_FUNCTION` would be worse, since it claims the qualitative loss the paper explicitly rules out. So the descriptor now names the function without grading it, and the loss is carried by the node name, the description and the evidence. I added a node-level `notes:` saying this, so the next person does not helpfully re-add `DECREASED`.

**Findings 2 and 4 — nine phenotypes added**, which also closes the GeneReviews Clinical Characteristics gap:

| Phenotype | HP | Frequency | Source |
|---|---|---|---|
| Hyperuricemia | HP:0002149 | 5/23 | PMID:35618652 |
| Extrapyramidal features | HP:0002071 | 5/23 | PMID:35618652 |
| Seizure | HP:0001250 | 3/23 | PMID:35618652 |
| Cerebral atrophy | HP:0002059 | 4/5 | PMID:31921599 |
| Ventriculomegaly | HP:0002119 | 4/5 | PMID:31921599 |
| Hyperammonemia (mild) | HP:0001987 | 2/5 | PMID:31921599 |
| Hypotonia | HP:0001252 | — | GeneReviews |
| Feeding difficulties | HP:0011968 | — | GeneReviews |
| Failure to thrive | HP:0001508 | — | GeneReviews |

**Finding 3 — three treatments added**, each with `target_mechanisms`: L-carnitine (`CHEBI:16347`), gut decontamination (`CHEBI:6909` metronidazole), and liver/kidney transplantation (`NCIT:C15289`). Transplantation has cblA-specific grounding on both sides — one patient in the Marelli cohort required a renal graft, and the Haarmann patient was transplanted at 35 with no recurrence of kidney damage at six years under continued hydroxocobalamin.

**Finding 5** — `reference_title` added to both ORPHA items.

**Suggestions taken:** removed the duplicated `GO:0004494` from the cofactor-protection node (you were right that it restates the consequence rather than the mechanism), removed `propionate catabolic process` from the ORGANISM-scale decompensation node, and added propionylcarnitine C3 (`CHEBI:53210`) and 2-methylcitric acid (`CHEBI:30835`).

## One suggestion I took only in modified form

**Hyperammonemia (suggestion 3) is in, but not from GeneReviews.** The GeneReviews sentence that mentions hyperammonemic encephalopathy is describing the *infantile / non-B12-responsive* form, and it names that group explicitly:

> Infantile/non-B12-responsive form (mut0 enzymatic subtype, cblB), the most common phenotype... could rapidly progress to coma due to hyperammonemic encephalopathy.

That is mut0 and cblB — not cblA. cblA is in the *next* sentence, the B12-responsive group, and that sentence does not mention ammonia. Citing GeneReviews here would have imported a mut0/cblB feature into a cblA entry, which is the specific misattribution this entry is built to avoid.

So it is sourced instead to PMID:31921599, where 2 of 5 MMAA-genotyped patients had mildly raised ammonia at diagnosis, and recorded as `MILD` / `OCCASIONAL`. The phenotype description states the distinction so the contrast is not lost.

## Two things worth knowing for other entries

**The bracket-stripping rule bit the GeneReviews phenotype quote.** The sentence that names cblA reads `(mut– enzymatic subtype, cblA, cblB [rare], cblD-MMA), in which symptoms occur...`. The reference validator strips `[rare]` from the snippet but not from the cache, so any quote spanning that bracket fails. The snippet now starts after it — which means the quoted clause no longer displays the gene list it describes, so each `explanation` states that the clause's subject is the B12-responsive list naming cblA.

**`check-snippet-length` failed on a file this diff never touches** — four short snippets in `Hyperprolinemia_Type_2.yaml`. That was a stale branch, not a defect: `20faacff65` on main had re-cut exactly those snippets, so this branch's older copy was being measured against the corrected baseline. Rebased onto `origin/main`; clean, no conflicts, no file overlap.

## Validation

All exit 0: `validate-disorders` (87/87), `validate`, `validate-terms`, `normalize-cache`, `check-term-cache-integrity`, `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, and `tests/test_data.py` locally. The `cache/` diff adds one row, `CHEBI:53210`.

Not changed, as you agreed: the umbrella `mondo_mappings` narrowMatch stays out. Beyond the scope argument, CLAUDE.md counts a `narrowMatch` as curated coverage, so adding one would make two entries both count as covering cblA and could retire it from the queue on the umbrella's strength rather than its own.

PR #10264, opened 2026-08-31
Clear the PARTIAL residue from five more entries (#10061)

caufieldjh:
Tranche 3 of the #10061 `explanation`-prose backlog. Five entries, finished completely: **115 items, 39 gaining `directness: INDIRECT`.**

| entry | items | `INDIRECT` |
|---|---:|---:|
| `Adult-Onset_Ataxia_and_Polyneuropathy` | 27 | 26 |
| `Fragile_X-Associated_Primary_Ovarian_Insufficiency` | 27 | 6 |
| `Hypertrophic_Cardiomyopathy_30_Atrial` | 21 | 8 |
| `Epilepsy_with_Generalized_Tonic-Clonic_Seizures_Alone` | 20 | 3 |
| `Mixed_Phenotype_Acute_Leukemia` | 20 | 1 |

## One rule, applied consistently

`directness: INDIRECT` is added **only where the curator's own words name an inference step.** Where they do not, the retired token is dropped and no directness is asserted — absent means nobody has assessed the item, which is the honest state and exactly what #10003 chose for all 11,620 of its own migrated items.

That is not conservatism for its own sake. Several caveats here are about **strength**, not directness, and an `INDIRECT` badge would misrepresent them:

> `…this is an absence of evidence at n = 499 rather than evidence that no such locus exists`
> `…with the source's own caveat that the estimate rests on limited regional data`
> `…no trial evidence was identified showing that counselling changes outcome`

**`Adult-Onset_Ataxia_and_Polyneuropathy` is the outlier at 26 of 27**, for a structural reason worth stating: nearly every item is drawn from a *different MT-ATP6 variant* (m.9035T>C rather than the defining m.8993T>C), from a cohort published under a *different diagnosis* this KB curates separately (NARP syndrome `MONDO:0010794`, the Leigh syndrome spectrum), or from a general mtDNA statement. All reach this entity through an inference step. The single exception cites regional imaging as the deepest tissue-level characterisation available — which *directly* documents the evidence ceiling its knowledge gap asserts, so it gains no slot.

Six entry-level `notes` blocks that documented these disciplines in terms of PARTIAL now document them in terms of directness.

## Tooling

One shape unlocked most of this: **`. PARTIAL because` with a full stop**, which the table only handled after `;` and `,`. That single addition took:

- `Adult-Onset_Ataxia_and_Polyneuropathy` 5/27 → **27/27**
- `Fragile_X-Associated_Primary_Ovarian_Insufficiency` 2/27 → **27/27**
- `Hypertrophic_Cardiomyopathy_30_Atrial` 3/21 → **21/21**

The pattern table does the **prose transform only**. The directness decision was made by reading, as in tranche 2 — the regex fired on 1 of the 26 in the ataxia entry, which is exactly why it is not trusted with that call.

## Verified in both directions

The consistency check — no explanation claims `INDIRECT` without the item carrying the slot, and no item carries the slot without the explanation saying why — is **0 in both directions** across all five files.

## Validation

- `just validate-disorders` over all five — **372/372 snippets verified**
- `just validate-history-all` — 7,796 records, no issues
- `check-duplicate-keys`, `check-entity-refs`, `check-enum-values`, `check-folded-hyphens` — clean
- `check-snippet-length`, `check-snippet-grading`, `check-title-snippets` grandfathered against `origin/main` — no new findings
- No `supports`, `snippet`, `reference`, `evidence_source` or ontology term changed

Branched on current `main` (post-#10234), so it does not hit the baseline race that turned #10258 red. Touches files disjoint from #10258 and #10263.

### DIFF
--- a/kb/disorders/Adult-Onset_Ataxia_and_Polyneuropathy.yaml
+++ b/kb/disorders/Adult-Onset_Ataxia_and_Polyneuropathy.yaml
@@ -554,9 +554,8 @@ phenotypes:
       Furthermore, the patient had an atactic gait pattern, a decreased sense of
       balance and could not walk without support.
     explanation: >-
-      A direct examination finding of ataxic gait in MT-ATP6 ataxia. Graded INDIRECT
-      because the patient carried m.9035T>C and presented in adolescence rather than
-      adulthood.
+      An examination finding of ataxic gait in MT-ATP6 ataxia. Graded INDIRECT because
+      the patient carried m.9035T>C and presented in adolescence rather than adulthood.
   - reference: PMID:22577227
     reference_title: Adult-onset spinocerebellar ataxia syndromes due to MTATP6 mutations.
     supports: SUPPORT
@@ -568,10 +567,9 @@ phenotypes:
       are clinically indistinguishable from other spinocerebellar ataxia
       patients.
     explanation: >-
-      Graded INDIRECT: the quote establishes adult-onset ataxia clinically
+      Graded INDIRECT because the quote establishes adult-onset ataxia clinically
       indistinguishable from spinocerebellar ataxia, but does not name gait, so it
       supports the cerebellar syndrome rather than this specific HPO term.
-
 - category: Neurological
   name: Peripheral Axonal Neuropathy
   description: >-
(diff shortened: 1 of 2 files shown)


caufieldjh:
All three fixed in `0e0edc0`, and both suggestions taken.

## The three breaks

You diagnosed the cause exactly: stripping `Marked PARTIAL because ` / `Marked as bounded because ` from mid-sentence left the following word lowercase at a sentence start, or left a stray verb stranded. It is the third-order consequence of the same original sin — I treated the job as token removal rather than sentence repair, so my edit was correct at the character level and wrong at the sentence level.

**I verified your completeness claim rather than taking it**, since "the list is complete" is exactly the kind of assurance that has bitten me twice in this series. A scan for a sentence boundary followed by a lowercase word — tolerating `e.g.` / `i.e.` / `et al.` — finds **0** across all five files after the fix. It is also **0** on #10258 and #10263, which used the same stripping, so this one does not need propagating. Good to have confirmed that both ways.

## Both suggestions

- **`A direct examination finding … Graded INDIRECT because …`** — you're right that it reads as a contradiction. Now "An examination finding". The sentence never needed the word, and the item genuinely is indirect (different variant, adolescent rather than adult onset), so the fix is to drop the adjective rather than to re-grade.
- **Frame consistency** — two spellings differed from the other 64 for no reason: `Graded INDIRECT:` → `Graded INDIRECT because`, and `Graded INDIRECT and` → `Graded INDIRECT, and`. The six lowercase `graded INDIRECT because` are mid-sentence and correct as they stand, so those are unchanged.

Worth recording for whoever picks up the remaining ~3,400: the frame is now uniform across three tranches, and the lowercase-sentence-start scan is cheap enough to run on every future one. Both are in the history records.

Validation: `just validate-disorders` over all five (372/372 snippets verified), consistency check 0 in both directions, self-audit for direction-denying prose and leftover grade tokens clean.

PR #10269, opened 2026-08-31
Add Progeroid_Syndromes grouping and curate member phenotypes

cmungall:
## Summary

Creates the `Progeroid_Syndromes` grouping as a SHARED_PHENOTYPE and CLINICAL_CONVENTION union over seven mechanistically distinct disorders that share a prematurely aged appearance. Deletes the corresponding curation stub and fixes three member disorders to curate the defining phenotype and inheritance patterns required by the grouping's membership criteria.

## Key Changes

- **New grouping**: `kb/groupings/Progeroid_Syndromes.yaml`
  - Scope: Hutchinson-Gilford progeria, Nestor-Guillermo progeria, Werner syndrome, Cockayne syndrome, XFE progeroid syndrome, Fontaine progeroid syndrome, and marfanoid-progeroid-lipodystrophy syndrome
  - Grouping basis: SHARED_PHENOTYPE + CLINICAL_CONVENTION (deliberately NOT shared mechanism)
  - Membership criteria: NECESSARY (not sufficient) — HAS_PHENOTYPE HP:0007495 AND HAS_INHERITANCE HP:0034345, both relying on HPO closure
  - Rationale: Members reach prematurely aged appearance by largely unrelated routes (lamin A missplicing, BANF1/BAF DNA-binding, RecQ helicase loss, nucleotide-excision repair failure, mitochondrial ATP-Mg/Pi carrier defect, and asprosin deficiency). No single pathograph exists; separate Disease entries with an explicit union is the correct shape. Cockayne and XFE are the one pair sharing a pathway, and they fail it at different steps.
  - Maps to MONDO:0015333 with exactMatch
  - Documents edge members (Cockayne syndrome, marfanoid-progeroid-lipodystrophy) and nine remaining MONDO descendants without dismech entries
  - Cites `PMID:23963297` for the claim that makes the criteria NECESSARY rather than NECESSARY_AND_SUFFICIENT

- **Disorder updates** (per curate-grouping rule: fix members failing necessary criteria rather than exempt them):
  - `Hutchinson-Gilford_Progeria_Syndrome.yaml`: Added HP:0007495 (Prematurely aged appearance, diagnostic, VERY_FREQUENT) and autosomal dominant inheritance block (HP:0000006, sporadic de novo)
  - `Nestor-Guillermo_progeria_syndrome.yaml`: Added HP:0007495 (Prematurely aged appearance, diagnostic) and autosomal recessive inheritance block (HP:0000007)
  - `Werner_Syndrome.yaml`: Added HP:0007495 (Prematurely aged appearance, diagnostic, VERY_FREQUENT); inheritance block already present

- **Stub deletion**: Removed `stubs/Progeroid_Syndrome.yaml` (curation completed)

- **Cache updates**: Added MONDO:0015333, HP:0034345 (Mendelian inheritance) to dynamic enums

## Implementation Details

- All new phenotype and inheritance evidence items are cited from existing reference cache entries (PMID:18256394, PMID:12714972, PMID:21932319, PMID:20301687, ClinGen assertion)
- Phenotype descriptions explain the composite nature of the aged appearance and its role in grouping membership
- Inheritance descriptions record both mode of action and clinical presentation (e.g., HGPS's dominant-negative progerin acting sporadically through de novo mutations; Werner's adult onset vs. childhood progerias)
- Grouping rationale explicitly addresses why cellular_senescence conformance is not a membership criterion despite downstream convergence
- Each member's `differentiating_mechanisms` module binding is a node unique to that member, so it actually differentiates
- Edge members documented in member notes rather than excluded: Cockayne syndrome's neurodevelopmental dominance and marfanoid-progeroid-lipodystrophy's lipodystrophy-driven aged appearance are recorded as conceptual tensions

## Review round

The second commit addresses the `ai4c-reviewer` review — see [the reply](https://github.com/monarch-initiative/dismech/pull/10269#issuecomment-5483405278) for detail. In short: XFE progeroid syndrome merged to `main` after this branch was cut, so the branch was rebased and XFE added as the seventh member; two module bindings that did not actually differentiate were rebound; one weak evidence quote was dropped; and a citation was added for the NECESSARY argument. One suggestion (uniform phenotype-name casing across files) was declined with reasons.

https://claude.ai/code/session_01Dh1FQmu62tDQD1yhytaNf2

### DIFF
--- a/history/other/Progeroid_Syndromes/2026-08-31T192626Z-claude-code-a641e0.yaml
+++ b/history/other/Progeroid_Syndromes/2026-08-31T192626Z-claude-code-a641e0.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: other
+  slug: Progeroid_Syndromes
+  path: kb/groupings/Progeroid_Syndromes.yaml
+session:
+  id: 2026-08-31T192626Z-claude-code-a641e0
+  timestamp: '2026-08-31T19:26:26Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10269
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - members
+  - mappings
+  - references
+  - notes
+  - description
+  summary: Add XFE as seventh member; rebind non-differentiating modules; cite the
+    NECESSARY argument
+  details: |
+    Review round on PR 10269. The ai4c-reviewer bot requested changes on one finding and raised four suggestions.
+
+    Blocking finding, accepted and fixed: XFE Progeroid Syndrome (MONDO:0012590) merged to main after this branch was cut, so the notes gap list - introduced as "descendants without dismech Disease entries" - asserted the absence of an entry that now exists, and the grouping omitted a member its own criteria admit. Verified independently against origin/main before acting: the entry exists, its name is exactly "XFE Progeroid Syndrome", it curates HP:0005328 (inside the HP:0007495 closure) and HP:0000007 (inside the HP:0034345 closure), and ERCC4 is hgnc:3436. Rebased onto origin/main (19 commits behind, clean, no unrelated churn), added XFE as the seventh member with a differentiating_mechanisms block contrasting its loss of the incision endonuclease itself against Cockayne's loss of only the transcription-coupled branch, and noting the GH/IGF-1 somatotroph suppression that entry curates. Removed XFE from the gap list and updated the three places that said "six members".
+
+    Suggestion 2, accepted: the module bindings on the HGPS and Werner differentiating_mechanisms did not differentiate. Confirmed by computing the conformance sets across all seven members - cellular_senescence#Senescence-Associated Cell Cycle Arrest, genomic_instability_aging#Declining Genome Maintenance and Erroneous Repair, and telomere_attrition#Progressive Telomere Attrition are all shared by HGPS and Werner, and both original bindings were in that shared set. The HGPS one was also in direct tension with grouping_rationale, which argues senescence convergence must not be treated as a discriminator. Rebound to member-unique nodes: HGPS to loss_of_proteostasis#Misfolded-Protein Aggregation (the progerin route), Werner to genomic_instability_aging#Accumulation of Somatic Mutations and Genomic Damage (the somatic-mutation burden behind its cancer predisposition).
+
+    Suggestion 4, accepted: dropped the PMID:903377 item from the Werner phenotype. It was an exact quote and passed grading, but it is a mid-sentence descriptive clause rather than that paper's result, and the GeneReviews item states the same thing as a clinical characterization.
+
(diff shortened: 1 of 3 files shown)


cmungall:
You were right about XFE, and it was a worse problem than a stale sentence. Not only did the notes claim XFE progeroid syndrome had no dismech entry when it now does, the grouping was leaving out a disease that its own membership rules admit. Fixed, along with two of the four suggestions.

I checked the XFE claim against `origin/main` before acting rather than taking it on trust: the entry is there, its name is exactly `XFE Progeroid Syndrome`, it curates HP:0005328 and HP:0000007, and ERCC4 is `hgnc:3436`. All four match what the review said.

## What changed

**XFE is now the seventh member.** I rebased onto `origin/main` first — the branch was 19 commits behind, which is also why XFE was invisible to me when I wrote the gap list. Its `differentiating_mechanisms` block contrasts it with Cockayne syndrome, since they are now the only two members that share a pathway: Cockayne loses just the transcription-coupled branch of nucleotide-excision repair, while XFE loses the incision endonuclease the whole pathway depends on. I also removed XFE from the gap list and fixed the three places that said "six members".

That second NER member broke something else in the description, which claimed the members' routes "have nothing in common at the molecular level". That was true with six members and is not true with seven. It now names the shared pathway and says they fail it at different steps.

**The module bindings did not differentiate — confirmed.** I computed the conformance sets across all seven members:

| Module node | HGPS | Werner |
|---|---|---|
| `cellular_senescence#Senescence-Associated Cell Cycle Arrest` | yes | yes |
| `genomic_instability_aging#Declining Genome Maintenance and Erroneous Repair` | yes | yes |
| `telomere_attrition#Progressive Telomere Attrition` | yes | yes |

Both of my bindings were in that shared set, so neither separated the two members. Your point about the HGPS one is the sharper one: `grouping_rationale` spends a paragraph arguing senescence convergence must not be treated as a discriminator, and then the member block used it as one. Rebound to nodes that are unique among the members — HGPS to `loss_of_proteostasis#Misfolded-Protein Aggregation`, Werner to `genomic_instability_aging#Accumulation of Somatic Mutations and Genomic Damage`.

**Dropped the `PMID:903377` item** from the Werner phenotype (suggestion 4). Agreed it is a descriptive clause rather than that paper's result.

**Added a `references:` block** citing `PMID:23963297` for the claim that carries the whole `NECESSARY`-rather-than-`NECESSARY_AND_SUFFICIENT` argument: autosomal recessive cutis laxa presents progeroid features prominently enough to be a real diagnostic confounder, yet is not a MONDO:0015333 descendant. I confirmed the `supporting_text` is an exact cache substring by direct comparison, because `validate-grouping` reports "0 checks" on a grouping — it does not look at `references[].findings[].supporting_text`.

## What I did not change

**The senescence-convergence claim is still uncited**, and I would rather leave it as stated reasoning than attach a citation that does not support it. Nothing in the reference cache asserts that convergence across these members. The nearest candidate, `PMID:26993153`, argues that genomic instability is the common mechanism of segmental progeroid syndromes — which cuts against this grouping, since the Fontaine and marfanoid-progeroid-lipodystrophy members have no genome-instability mechanism at all.

**I left the phenotype name casing alone** (suggestion 3). Each name matches its own file's convention rather than a cross-file one: Werner uses Title Case for all 26 of its phenotype names, HGPS and NGPS use sentence case throughout. Changing Werner's to `Prematurely aged appearance` would make it the only lowercase name among 26 siblings. I think within-file consistency is the more useful kind here, but say the word if you disagree and I will switch it.

Thanks for checking the closure relationships and the gap list against MONDO independently — the XFE catch depended on that.

## Validation

`check-groupings --strict` now reports all seven members SATISFIED. Also clean: `validate-grouping`, `validate` on Werner, `count-verified-snippets` 179/179, `check-entity-refs`, `check-duplicate-keys`, `validate-history`, and all four baseline gates. Your note about `check-snippet-length` was right — it was the `Hyperprolinemia_Type_2` artifact of being behind `main`, and it passes now that the branch is rebased.

---

PR #10276, opened 2026-08-31
Osteoporosis: add aging mechanism arm, model systems, and close evidence gaps

cmungall:
## Summary

Expands the Osteoporosis entry along the two axes it was missing: **aging as a mechanism**, and **model systems**. The entry described the disease as "accelerated by estrogen deficiency and aging" but carried no aging pathophysiology at all — three of its six mechanism nodes were about spaceflight, and none were about age. It also had no `animal_models`, `experimental_models` or `computational_models`, against 553, 272 and 33 files elsewhere in the KB.

Compliance goes from 72.9% to 88.0% while the number of checked fields grows from 140 to 250. All 103 evidence snippets verify against the reference cache.

## Aging mechanism

Three new pathophysiology nodes, each conforming to a hallmarks-of-aging module the entry previously ignored:

| Node | Conforms to |
|---|---|
| Osteocyte SASP Acquisition | `cellular_senescence#Senescence-Associated Secretory Phenotype` |
| Senescent Cell Accumulation in the Bone Microenvironment | `cellular_senescence#Senescent Cell Accumulation` |
| Marrow Stromal Lineage Shift Toward Adipogenesis | `stem_cell_exhaustion#Decline in Stem Cell Self-Renewal and Function` |

This is the arm that operates in both sexes and is distinct from estrogen deficiency.

## Model systems

**Animal.** The ovariectomized rat, which the 1994 FDA preclinical guidance names as the rodent species — recorded with its limitation, that over 12 months cancellous turnover returns to sham values while trabecular bone volume stays low, decoupling the mechanism from the deficit. The INK-ATTAC senescent-cell clearance mouse as a `RESCUES` link with target-engagement and age-specificity readouts. The aged wild-type mouse, carrying a `REFUTE` item for the cortical mismatch below.

**Non-animal (NAMs).** MLO-Y4 and IDG-SW3 osteocyte lines, and a human 3D bone organoid with coupled osteoblast–osteoclast remodeling. The organoid is flagged as developed and validated for myeloma bone disease, with no aging axis — it is the closest available system to a human remodeling unit, but its driver is plasma cells, not age or estrogen loss.

**Computational.** A QCT-based finite element hip model, prospectively validated in MrOS and still predictive after adjustment for areal BMD, and the Komarova bone cell population dynamics model of the remodeling unit.

## Knowledge gaps

Four new discussions, all aging-facing. The substantive one is a `HUMAN_MODEL_MISMATCH`: the rodent skeleton has no osteonal remodeling of cortical bone, but human cortical loss runs largely through intracortical remodeling — 47% of bone loss in Zebaze's measurement, in the compartment that carries 80% of fractures in old age. When an aged mouse gains cortical porosity, that is a different structural process producing a similar microCT number. This is why the animal links are set to `MODERATE` fidelity and not `HIGH`.

The other three: senolytic stratification, where clearance is causal in mice but the phase 2 human trial missed its primary endpoint; the direction of causation for marrow adiposity, which cross-sectional biopsies cannot settle; and the absence of any NAM that sustains an aging-capable coupled remodeling unit.

## Corrections to existing content

1. **VDR was listed as a risk factor with no evidence.** The 26,242-participant meta-analysis found the FokI, BsmI, ApaI and TaqI polymorphisms associate with neither BMD nor fracture, and VDR is not among the 56 genome-wide significant BMD loci. I think the original annotation was wrong, so it is now recorded as `REFUTE` and reclassified `RISK_FACTOR` → `DISPUTED`.

2. **Height Loss was bound to HP:0004322 (Short stature).** That term asserts height below −2 SD for age and sex, which is a different claim from acquired loss of adult stature. HPO has no term for the latter, so it is now bound to the accurate parent HP:0000002 rather than to a wrong specific term.

3. **Three adverse findings the entry omitted** are now recorded: the romosozumab cardiovascular imbalance in ARCH; the Cochrane result that the largest femoral-neck benefit came from non-weight-bearing resistance training with no reduction in fracture numbers, which makes the "Weight-Bearing Exercise" treatment name narrower than its own evidence; and denosumab discontinuation rebound, where the vertebral fracture rate returns to untreated levels off therapy because the drug inhibits the RANKL pathway without altering the state driving it.

## Prevalence (new section)

| Population | Prevalence | Source |
|---|---|---|
| Worldwide | 18.3% (95% CI 16.2–20.7) | PMID:34657598 |
| Women worldwide | 23.1% (19.8–26.9) | PMID:34657598 |
| Men worldwide | 11.7% (9.6–14.1) | PMID:34657598 |
| US adults ≥50, NHANES 2005–2010 | 10.3% | PMID:24771492 |

Uses the structured slots per design decision §8 — `measure_type`, `prevalence_class`, `rate_per_100000` — not the deprecated free-text `percentage`. The worldwide figures are recorded with a note that the pooled population mixes screening settings and age structures, so they are order-of-magnitude rather than age-standardised.

## Evidence added

Every previously uncited treatment now cites its pivotal trial: denosumab (FREEDOM), teriparatide (Fracture Prevention Trial), romosozumab (FRAME and ARCH), calcium and vitamin D (Tang meta-analysis), exercise (Cochrane). Denosumab also gains a `target_mechanisms` link to the RANKL/OPG node it acts on. All four genetic loci are cited, three of them from the GEFOS BMD meta-analysis. Evidence added to three phenotypes and all four biochemical markers.

## Review round

The first review found that the two new senescence nodes carried each other's GO terms — the SASP conformer had `GO:0090398 cellular senescence`, which defines the module's *accumulation* node, and the accumulation node had no `biological_processes` at all. Fixed: the SASP node now carries `GO:0001816` and `GO:0006954` with supporting evidence, the accumulation node carries `GO:0090398`, and the SASP node was renamed so it makes one claim rather than bundling arrest with secretion.

## Validation

- `just validate-disorders` — 103/103 snippets verified against the cache
- `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-term-cache-integrity` — all pass
- `tests/test_data.py` — 47112 passed on the current head
- Four history records under `history/disorders/Osteoporosis/`

Rebased onto current main to pick up the snippet re-cut in 20faacff6; conflict-free, disjoint file sets.

37 files: the entry, 4 history records, and 32 reference cache files.

### DIFF
--- a/history/disorders/Osteoporosis/2026-08-31T234009Z-claude-code-6d63ad.yaml
+++ b/history/disorders/Osteoporosis/2026-08-31T234009Z-claude-code-6d63ad.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Osteoporosis
+  path: kb/disorders/Osteoporosis.yaml
+session:
+  id: 2026-08-31T234009Z-claude-code-6d63ad
+  timestamp: '2026-08-31T23:40:09Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10276
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  - treatments
+  summary: Fix module-conformance GO placement and record denosumab rebound
+  details: |
+    Addressed the blocking finding in the PR review of pull request 10276. The two new senescence nodes carried GO terms belonging to each other s module nodes. Osteocyte SASP Acquisition conforms to cellular_senescence Senescence-Associated Secretory Phenotype but carried only GO:0090398 cellular senescence, which is the process defining the module s Senescent Cell Accumulation node, and none of the three the SASP node defines; it now carries GO:0001816 cytokine production INCREASED and GO:0006954 inflammatory response INCREASED, supported by a new evidence item quoting the SASP-marker measurement in osteocytes from PMID:27341653. Senescent Cell Accumulation in the Bone Microenvironment had no biological_processes at all and now carries GO:0090398 INCREASED, matching the Chondrocyte Senescence conformer in Osteoarthritis. Also took the reviewer s optional rename so the node makes one claim rather than bundling arrest with SASP: Osteocyte Senescence and SASP Acquisition becomes Osteocyte SASP Acquisition, with both attaches_to references updated. Took two of the three non-blocking suggestions. Added a notes line to the marrow adipogenesis node recording that its conformance is to the differentiation-potential aspect of stem cell exhaustion rather than the pool-maintenance aspect, so the absent GO:0019827 reads as deliberate. Added denosumab discontinuation rebound, which the reviewer identified as the one unmodelled research theme citable from independent literature: a REFUTE item on the treatment quoting the FREEDOM post hoc analysis where vertebral fracture rate rose from 1.2 to 7.1 per 100 participant-years off treatment (PMID:29105841), a notes line explaining that the effect is not retained after withdrawal because the drug inhibits the RANKL pathway without altering the state driving it, and an ECTS position statement item on the target_mechanisms link (PMID:33103722). Did not model the oxidative-stress Nrf2/FOXO/SIRT axis or the gut-bone axis: the reviewer noted the deep-research report s only sources for both are journals it itself flags as predatory. Snippets verified 103 of 103.
--- a/kb/disorders/Osteoporosis.yaml
+++ b/kb/disorders/Osteoporosis.yaml
@@ -82,7 +82,7 @@ prevalence:
     explanation: Nationally representative DXA-based prevalence estimate for a
       single well-characterised population.
 pathophysiology:
-- name: Osteocyte Senescence and SASP Acquisition
+- name: Osteocyte SASP Acquisition
   conforms_to: "cellular_senescence#Senescence-Associated Secretory Phenotype"
   biological_scale: CELLULAR
   description: >-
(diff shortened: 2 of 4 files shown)


cmungall:
Fixed the blocking finding — the GO terms were on the wrong nodes, and you had the direction right.

`Osteocyte SASP Acquisition` now carries GO:0001816 cytokine production and GO:0006954 inflammatory response, both INCREASED, with a new evidence item quoting the osteocyte SASP-marker measurement from PMID:27341653. GO:0090398 is off it.

`Senescent Cell Accumulation in the Bone Microenvironment` now carries GO:0090398 cellular senescence INCREASED, matching the `Chondrocyte Senescence` conformer.

I took the optional rename as well — the old name bundled arrest with SASP, which the module deliberately keeps apart. Both `attaches_to` references updated.

## Two of the three suggestions taken

**Marrow adipogenesis node** — added the `notes` line recording that the conformance is to the differentiation-potential aspect of the hallmark rather than pool maintenance, so the absent GO:0019827 reads as deliberate rather than as an oversight.

**Denosumab rebound** — modelled, using independent literature. A `REFUTE` item on the treatment quoting the FREEDOM post hoc analysis (vertebral fracture rate rising from 1.2 to 7.1 per 100 participant-years off treatment, PMID:29105841), a `notes` line on why the effect is not retained — the drug inhibits the RANKL pathway without altering the state driving it — and the ECTS position statement on the `target_mechanisms` link (PMID:33103722).

**Not taken:** the oxidative-stress Nrf2/FOXO/SIRT and gut–bone axes, for the reason you gave. The report's only sources for both are journals it flags as predatory, and I would rather leave them unmodelled than cite those.

On the prevalence units — already corrected before your review landed, so the "per 1,000" prose is gone. The description now also covers the model systems and knowledge gaps, which the auto-generated version omitted.

## One thing worth flagging

`check-snippet-length` went red on this branch after I fetched main, reporting four short snippets in `Hyperprolinemia_Type_2.yaml`. That is not from this PR. The file was fixed on main by 20faacff6, one of the 20 commits this branch was behind, so my stale copy read as "new" against the updated baseline. I rebased onto current main — conflict-free, the two touch disjoint file sets — and it clears. Worth knowing that the check can go red on any branch that falls behind a snippet fix, without anything being wrong with the branch itself.

## Validation on the rebased tree

```
just validate-disorders      → 103/103 snippets verified
check-folded-hyphens         → OK
check-snippet-length         → OK  (was the stale-baseline red above)
check-title-snippets         → OK
check-snippet-grading        → OK
check-environmental-evidence → OK
check-term-cache-integrity   → OK
tests/test_data.py           → 47112 passed
```


---

PR #10281, opened 2026-08-31
Add aging biomarkers to all 11 hallmark modules, plus the gaps where none exist

cmungall:
## Summary

dismech had a complete mechanism model of aging and almost no biomarker model of it. All 12 hallmarks of aging existed as modules with proper node chains, but only one of the 166 modules in `kb/modules/` carried a `biochemical:` block. This PR curates biomarkers into all eleven aging-hallmark modules — 20 markers with `BiomarkerReadout` links to the mechanism nodes they report on — and records 11 new `KNOWLEDGE_GAP` discussions where the evidence does not support a marker.

No schema changes. The existing `BiomarkerReadout` class already had what was needed.

**Status: approved, CI green, no conflicts.** Waiting only on the repo's auto-merge age gate (eligible from ~2026-09-03).

## What was added

| Module | Markers |
|---|---|
| `inflammaging` | IL-6, hsCRP, TNF, cell-free mtDNA, agalactosylated IgG N-glycans |
| `mitochondrial_dysfunction` | GDF-15, FGF-21, blood mtDNA copy number |
| `deregulated_nutrient_sensing` | IGF-1, insulin, HbA1c |
| `gut_dysbiosis` | LBP, soluble CD14 |
| `telomere_attrition` | Leukocyte telomere length |
| `epigenetic_alterations` | DNAm PhenoAge clock estimate |
| `loss_of_proteostasis` | Plasma p-tau217 |
| `genomic_instability_aging` | CHIP |
| `stem_cell_exhaustion` | CHIP (same measurement, different claim — see below) |
| `disabled_macroautophagy` | **none, deliberately** — see below |
| `cellular_senescence` | *(p16INK4a and SA-β-gal, pre-existing; the pattern the rest follow)* |

Together these cover the TAME Biomarkers Workgroup trial panel and most of the 2025 Delphi expert consensus list. IL-6, CRP and GDF-15 appear on the TAME panel, the Delphi list, **and** the NIA 2024 symposium — about as close to a validated core as this field currently offers.

## Four things worth a reviewer's attention

**1. `disabled_macroautophagy` gets no biomarkers, and that is the result, not an omission.** Autophagy is a flux, not a level. LC3-II and p62 are steady-state abundances of proteins the process consumes, so a raised value is equally consistent with more autophagosome formation and with blocked degradation. Distinguishing them needs a lysosomal clamp or a tandem-fluorescent reporter, and the cited methods reviews scope both to cell culture and animal models. The module gets a knowledge gap on an intentionally empty section, with an explicit instruction not to import a single LC3 or p62 reading — without a flux clamp it has no defensible `direction`.

**2. CHIP is curated in two modules on purpose.** In `genomic_instability_aging` it reports accumulated somatic mutation; in `stem_cell_exhaustion` it reports hematopoietic stem-cell clonal dominance. Two different claims about one measurement, so the readout targets and interpretations differ, and each entry names the other so it does not read as an error. It is also the cleanest cross-species case here: human association plus mouse bone-marrow-transplant evidence that *Tet2*-deficient marrow enlarges atherosclerotic lesions.

**3. IGF-1 has two directions, and an early draft put one of them in the wrong slot.** Against the node it reports, IGF-1 is ordinary: higher IGF-1 means more anabolic signalling, so both readouts are `POSITIVE`. Against *outcome* it inverts — lower IGF-1 predicts longer survival in nonagenarians, in females only. The first draft encoded that outcome inversion as `direction: NEGATIVE`, which contradicted the sibling readout to `mTORC1 Hyperactivation` (the direct `downstream` of that node). Caught by review and fixed in 28a631e5; the longevity inversion now lives in the entry notes, where it is a claim about outcome rather than about the linked event. An audit of every readout direction across the curated modules found no other instance, and the reviewer re-audited independently and agreed.

**4. Limitations are curated as `REFUTE` items, not buried in prose.** Where a source cuts both ways it is split: GDF-15 is not specific to mitochondrial disease; FGF-21 is only weakly induced by structural respiratory-chain defects, so a normal value does not exclude disease; HbA1c failed consensus agreement on predicting biological age. In `gut_dysbiosis` the `REFUTE` item is the interesting one — in centenarians followed to death, alpha diversity did **not** differ before death while ten individual species did, which refutes the default assumption a curator would otherwise import.

## The gap this surfaced, and where it is filed

Recurring across five modules: **dismech has no class for a value computed over other measurements.** An epigenetic clock is a penalized regression over hundreds of CpGs; a composite biomarker panel is a fitted combination the field explicitly prefers over single analytes; a frailty index is a ratio over a deficit list. None is an analyte, so none belongs in `Biochemical` — yet `Biochemical.readouts` is the only place a readout link can live, so DNAm PhenoAge is curated there with the reasoning recorded in its own `notes`.

Working it through for the decision register showed this is **two** gaps rather than one, and merging them would produce the wrong schema:

- **Computed indices over measurements** — clock, composite panel, frailty index. Fitted estimators.
- **Composite clinical outcome endpoints** — disability-free survival (ASPREE), multimorbidity (TAME). Outcomes, not measurements, and the thing a candidate surrogate is surrogate *for*. `kb/surrogate_endpoints/` holds 225 FDA rows and none is an aging endpoint, so every `endpoint_context: CANDIDATE_SURROGATE` here currently points nowhere.

The frailty index straddles both, which is why it looks like one problem from a distance.

Filed in `docs/explanation/design-decisions.md` §12 as **undecided, not proposed** — three candidate shapes are named and none worked through, because picking one is a maintainer call. `CLAUDE.md` now points there from the hallmarks-of-aging entry, so a curator adding the next composite marker meets it.

## Ontology gap

NCIT has no term for an epigenetic clock or for biological age. Searching returns nothing for "Epigenetic Clock" and only *Biological Agent* for "Biological Age". `NCIT:C17961` (DNA Methylation) and `NCIT:C16269` (Aging) were considered and rejected — neither is the measurement. The interim binding names the assay (`NCIT:C63328`) and carries the clock identity in `preferred_term`, per the Ontology Term Contract, with the reasoning in the entry. **An NCIT term request is the real fix and is not filed yet.** This is the most consequential ontology gap in the area, since clocks are on the consensus list.

## Two merges with main

Main moved fast under this branch and both refreshes touched the same files, so the resolutions are worth recording.

**`85d554f9`** — the branch had fallen 55 commits behind and main had independently edited all ten modules touched here. The case worth knowing about: **`inflammaging` had a `discussions:` key added on both sides** — the shape that, auto-merging at different offsets, produces a duplicated top-level key that `safe_load` silently resolves by keeping the last block. Resolved to a single `discussions:` holding all five entries (main's naked mole-rat `HUMAN_MODEL_MISMATCH` plus this branch's four gaps); blocks merged, neither side dropped. `deregulated_nutrient_sensing` was not a same-key collision at all: main rewrote a `mechanistic_hypotheses` `notes:` field after an openscientist assessment, and this branch's `biochemical:` block landed just after it. Verified against the merge base that this branch never touched those notes, so main's rewrite is taken whole.

**`a86e1265`** — #10283 merged, adding `computational_models:` to nine of these modules. No same-key collision this time; all nine resolved as keep-both.

The check that matters for both: **zero lines removed relative to `origin/main`** except the one `CLAUDE.md` line this branch intentionally replaces. `check-duplicate-keys` clean over 4,263 YAML files, `check-entity-refs` over 2,794. All content present across the eleven modules — 20 biomarkers and 14 knowledge gaps from this branch, 30 computational models and 5 animal models from main.

## Validation

- `just validate` on every touched module — all snippets verified against cached references
- `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-term-cache-integrity`, `check-cache-order` clean
- Whole-KB `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-empty-snippets`, `check-folded-hyphens`, `check-environmental-evidence` all pass
- 21,481 targeted pytest KB tests pass after the second merge; `validate-history-all` clean
- Every snippet independently confirmed as a **literal exact substring** of its cached reference. Both validators normalize Greek letters, so `TNF-alpha` passed while the source says `TNF-α`; the quotes were corrected to the source characters and re-checked.

23 new reference cache entries, all via `just fetch-reference`. A history record accompanies each module change.

## Note on the scope guard

`test (3.13)` failed at the curation-scope guard until the `scope-override` label was applied, because this PR touches `kb/` alongside `CLAUDE.md`, `docs/` and `mkdocs.yml`. Neither I nor the reviewer self-applied the label — a guard whose purpose is to force a person to make a scope call is not one an agent should clear for its own work.

Worth flagging for anyone reading the check history: **the guard runs before the test suite**, so while it was failing, no test in this branch had ever executed in CI. The first real run was after the label went on, and it passed.

## Not done

- NCIT term request for epigenetic clock / biological age
- A `Biomarkers_of_Aging` grouping
- The progeroid pilot (`Werner_Syndrome`, `Hutchinson-Gilford_Progeria_Syndrome` conform to 4 and 6 hallmarks and still carry zero biomarkers) — deliberately left, since it is disease-entry work rather than module work

One process note: a knowledge gap for the DNAm-age/inflammaging link was drafted from Franceschi 2018's statement that it "has not been investigated", then **not** curated — a PubMed check found 51 papers on epigenetic age acceleration and inflammaging, so the claim is stale. PhenoAge's own transcriptional analysis closes the loop.


cmungall:
Both IMPORTANT findings are fixed and pushed in 28a631e5. Ready for re-review, with one blocker that needs a maintainer decision — see the end.

**Finding 1, IGF-1 direction — you're right, and it was worse than a slip.** The readout to `Nutrient Surplus and Anabolic Signaling` is now `POSITIVE`. What I had done was conflate two directions that genuinely point opposite ways for this marker: against the node it reports, IGF-1 is ordinary (higher IGF-1, more anabolic signalling); against *outcome* it inverts (lower IGF-1, longer survival in the oldest old). I put the outcome inversion in `direction`, which is the wrong slot — and your point about the sibling readout is the decisive one, since asserting `NEGATIVE` to a node and `POSITIVE` to its direct `downstream` cannot both be true.

I also changed `presence` from `DECREASED` to `INCREASED`, matching the sibling insulin and HbA1c entries and the deregulated state this module models — that was the same error showing up in a second slot. The longevity inversion is kept, moved to the entry notes where it is a claim about outcome.

Because one finding of this shape suggests a class, I audited every readout direction added across all nine curated modules against its target node. IGF-1 was the only instance. The three remaining `NEGATIVE` values are correct — higher telomere length means less attrition, higher mtDNA copy number means less damage. The error was isolated to the one marker where the biology genuinely inverts against outcome, which is exactly where the wrong slot is tempting.

**Finding 2, CLAUDE.md count — fixed, and taken the way you suggested.** Dropped rather than corrected: it now says "Most carry a `biochemical:` block" with a `grep -l '^biochemical:' kb/modules/*.yaml` for the current set, so there is no number to go stale. The §12 pointer and the NCIT warning are untouched.

The report carried both errors too (it claimed IGF-1 was curated `NEGATIVE`, and a stale six-of-eleven module count); both corrected there.

**Thanks for the `check-snippet-length` note.** It passes clean here on the current head, so I have left `Hyperprolinemia_Type_2.yaml` alone — agreed it is not this PR's.

---

**Remaining blocker: the curation-scope guard, which needs a maintainer call rather than a code fix.**

`test (3.13)` fails at the scope check, not on any test:

```
CLAUDE.md
docs/explanation/design-decisions.md
docs/reports/biomarkers-of-aging-gap-analysis-2026-08-31.md
mkdocs.yml
```

The guard's allowlist is `kb/ stubs/ history/ cache/ references_cache/ research/ pages/ dashboard/` — `docs/` is not on it. This PR genuinely mixes curation with documentation, so the failure is real rather than spurious.

Two ways out, and picking one is a scope judgement I do not think an agent should make for itself:

1. **Add the `scope-override` label.** The guard's own message prescribes this, and the label exists for "an intentional curation PR exception". The docs here are documentation *of* this curation — the report tracks it, the register entry records a gap found during it, and `CLAUDE.md` points at that entry.
2. **Split the docs into a separate PR.** Cleaner in principle, but the two are interleaved across all eight commits rather than separable by commit, so it would need file-level surgery and would leave each half harder to read alone.

I lean towards option 1, but I have deliberately not applied the label myself — the guard exists to force a person to make this call, and self-granting the override would defeat that. @cmungall, your call.

---

cmungall:
Thanks — and the independent re-audit of all 20 readouts was the right call rather than taking my sweep on trust.

I took the optional observation, in `b358653d`. You were right that it was the last place an outcome claim sat in a node-direction slot, which is the same defect as the blocking finding rather than a separate nit.

On your suggestion to swap in a better quote: **there isn't one in that reference.** PMID:24618355 is a survival study, and every quotable sentence in it concerns outcome — the closest candidates are "Attenuated GH/IGF-1 signaling is associated with extended lifespan in several animal models" and "we hypothesized that lower IGF-1 levels are predictive of extended survival", both of which have the same problem. Substituting would have meant introducing a new reference for a non-blocking item, which felt like the wrong trade.

So instead the item is graded `directness: INDIRECT` — the documented slot for a quote supporting a claim through an inference step — and the explanation now states what the quote does establish (IGF-1 is quantifiable at a stated threshold and its level is consequential), what it does not (the anabolic-tone relationship), where the direction actually rests (IGF-1 being the effector arm of the axis, per the interpretation directly above), and that no better-matched quote exists in this source so a future curator does not go looking. The snippet is unchanged, so there is no re-verification risk.

This push drops your approval, which I judged worth it: the PR cannot merge until the scope decision lands regardless, so there is no merge cost to a re-review, and leaving the residue would have meant the correction was only three-quarters done.

Nothing else changed. The scope-override decision remains the only outstanding item, and neither of us should be the one to make it.

---

PR #9935, opened 2026-08-28
curate(Liver_Cirrhosis, DILI module): first liver fibrosis NAM

gingin77:
The knowledge base had no liver fibrosis microphysiological system. It had three liver models, all on the Drug-Induced Liver Injury Module and all about drug toxicity; `Liver_Cirrhosis`, MASLD, alcoholic liver disease and the `fibrotic_response` module had no experimental models at all. This adds one, from Schmidt & Suter-Dick, *Toxicology* 2025 (`PMID:40754287`).

**The system.** A 384-well Akura Twin microplate of 168 interconnected well pairs. HepaRG hepatocyte microtissues, with or without THP-1 monocytic cells, sit in one well of each pair; hepatic stellate cell (hTERT-HSC) microtissues sit in the other. TGF-β1, methotrexate and acetaminophen are the challenges. The paper's stated goal is to quantify the key events of the liver fibrosis Adverse Outcome Pathway.

**Split across two entries**, because its readouts speak to two different mechanisms:

`kb/disorders/Liver_Cirrhosis.yaml` takes the fibrosis arm:

| Target node | Relationship | Readouts |
|---|---|---|
| Hepatic Stellate Cell Activation | `RECAPITULATES` | ACTA2/COL1A1/COL3A1/FN1 ↑; Pro-Collagen 1A1 and CTGF ↑ |
| Kupffer Cell and Inflammatory Response | `PARTIALLY_RECAPITULATES` | ALOX5AP/TREM2/TGF-β1 mRNA ↑; PAI-1 ↑ |
| Hepatocyte Injury and Death | `PARTIALLY_RECAPITULATES` | albumin ↓ |
| TGF-beta Signaling in Fibrogenesis | `PERTURBS` | downstream marker induction ↑ |

`kb/modules/drug_induced_liver_injury.yaml` takes the drug-challenge arm: Hepatocyte Cell Death (`PERTURBS`) and Sterile and Immune-Mediated Inflammatory Amplification (`PARTIALLY_RECAPITULATES`).

**Three judgement calls worth a reviewer's attention:**

Item 1 — `experimental_model_type` is `CO_CULTURE`, not `ORGAN_ON_CHIP`. The defining feature is compartmental multi-cell-type co-culture of 3D microtissues in a microplate, not microfluidic perfusion. `ORGANOID` would also be wrong: HepaRG and hTERT-HSC microtissues are cell-line spheroids, not self-organizing stem-cell organoids.

[...]

PR #10257, opened 2026-08-31
docs(AOP): worked example mapping AOP 38 to the liver fibrosis NAM (#10051)

gingin77:
A microphysiological system built to measure liver fibrosis maps onto five of the seven Key Events in AOP 38, and the two it misses are the two endpoints. This PR records that mapping as a worked example on the AOP/EMOD alignment project page, and tags the biological scale of the mechanism nodes involved.

Closes #10051.

## What was done

The AOP/EMOD page listed Assay/NAM ↔ `ExperimentalModel` as the cheapest bridge between the two frameworks, but nothing had tested it. The liver fibrosis microphysiological system curated in #9935 is a case that can, because the authors built it "to quantify the key events of the liver fibrosis AOP" — the AOP framing is theirs, not something applied afterwards.

Their target is AOP 38, Protein Alkylation leading to Liver Fibrosis. Lining its Key Events up against the dismech nodes gives:

| Key Event | Level | Assay readout | dismech node | Fit |
|---|---|---|---|---|
| 244 Alkylation, Protein *(MIE)* | Molecular | — | — | none |
| 55 Increase, Cell injury/death | Cellular | albumin ↓ | Hepatocyte Injury and Death | yes |
| 1492 Tissue resident cell activation | Cellular | ALOX5AP, TREM2 ↑ | Kupffer Cell and Inflammatory Response | partial |
| 1493 Increased Pro-inflammatory mediators | Tissue | PAI-1, TGF-β1 ↑ | *(same node)* | partial |
| 265 Increase, Hepatic stellate cell activation | Cellular | ACTA2, COL1A1, COL3A1, FN1 ↑ | Hepatic Stellate Cell Activation | yes |
| 68 Increase, Collagen accumulation | Tissue | Pro-Collagen 1A1, CTGF ↑ | `fibrotic_response#Excessive ECM Deposition` | yes |
| 344 Increase, Liver fibrosis *(AO)* | Organ | — | `Liver_Cirrhosis`, the entry | none at node level |

The two misses are informative. The adverse outcome maps to the disease entry, not to any single mechanism node — so anything that records a Key Event cross-reference on `Pathophysiology` alone cannot hold that row. The molecular initiating event has no node at all, which is the same result the lead pilot on this page reached from the opposite direction.

## The acetaminophen question

[...]

PR #10282, opened 2026-08-31
enhance(fibrotic_response): add Scar-in-a-Jar module-level experimental model (#9997)

### LINKED ISSUE #9997: Add experimental models to the fibrotic_response module
The `fibrotic_response` module is the most reused module in the knowledge base and has no experimental models attached. Two organ-specific fibrosis models were curated recently that both map onto its nodes, which makes the gap visible — but neither of them belongs at the module level, so this needs a literature pass rather than a copy-across.

**How much conforms to this module today**

| `fibrotic_response` node | Entries declaring `conforms_to` |
|---|---|
| Mesenchymal Cell Activation | 29 |
| Excessive ECM Deposition | 20 |
| Architectural Distortion and Organ Dysfunction | 11 |
| Tissue Injury | 9 |
| Inflammatory Recruitment and Amplification | 7 |

**What prompted this**

`Idiopathic_Pulmonary_Fibrosis` carries a microengineered alveolar-array lung-on-chip with breathing mechanics (`PMID:41406599`), linked `RECAPITULATES` to nodes conforming to `fibrotic_response#Mesenchymal Cell Activation` and `#Excessive ECM Deposition`.

`Liver_Cirrhosis` carries the 384-well Akura Twin liver fibrosis microphysiological system (`PMID:40754287`, added in #9935), linked `RECAPITULATES` to `Hepatic Stellate Cell Activation`, which conforms to `fibrotic_response#Mesenchymal Cell Activation`.

Two organs, two platforms, the same module nodes, and the same class of readout in both: myofibroblast marker induction (ACTA2, COL1A1) and matrix output (pro-collagen, fibronectin, CTGF).

**What to do**

Add an `experimental_models:` block to `kb/modules/fibrotic_response.yaml` with `modeled_mechanisms` links to `Mesenchymal Cell Activation` and `Excessive ECM Deposition` — the two nodes where evidence already exists. The other three nodes can follow if a model supports them.

**The hard part**

A module-level model has to be a real published system. All six modules that currently carry `experimental_models` cite one — `cellular_senescence` and `photoaging` share the Outer Biosciences ex vivo skin platform (`DOI:10.64898/2026.04.30.721933`), `intestinal_barrier_dysfunction` has two primary-cell barrier models (`PMID:29094594`, `PMID:31717494`), `myelosuppression` a bone-marrow-on-chip (`PMID:31988457`), `microtubule_dependent_neuronal_migration_failure` a cerebral organoid model (`PMID:28111201`). None is a generic archetype, so inventing one here would break the pattern.

But the two models in hand are explicitly organ-specific — a lung alveolar chip and a liver microtissue plate — and neither is a fair stand-in for fibrosis generally. So this needs one of:

- a published fibrosis platform that is genuinely organ-agnostic (a TGF-β-driven fibroblast or scar-in-a-dish system not tied to one organ), or
- a decision to cite an organ-specific system as the module's exemplar, with the organ limitation stated in `limitations`.

The first is preferable. The second is defensible if the search comes up empty, but should be argued in the PR rather than done quietly.

https://claude.ai/code/session_01VQ6V6qb1FLD5Ek7kQmiheq

gingin77:
Closes #9997.

`fibrotic_response` is the most reused module in the knowledge base and had no experimental models attached. This adds one: the **Scar-in-a-Jar** macromolecular-crowding fibroblast assay, linked to the two module nodes where published evidence exists — Mesenchymal Cell Activation and Excessive ECM Deposition.

The issue asked for an organ-agnostic published fibrosis platform if one exists, and to argue the case in the PR rather than quietly falling back to an organ-specific system. One exists, and the argument is below.

## Why this system can sit at the module level

Every other module carrying `experimental_models` cites a real published platform, so the model here had to be one too — not an invented archetype. But the two models that made this gap visible (the IPF lung-on-chip, the Liver_Cirrhosis Akura Twin plate) are each built around one organ's cells and architecture, and neither is a fair stand-in for fibrosis generally.

Scar-in-a-Jar is different in a specific way: **the collagen-producing cell is a parameter of the assay, not a fixed feature of it.** The protocol chapter says so directly —

> employing macromolecular crowding and a collagen-producing cell type of choice (e.g., lung fibroblasts like WI-38)

— and the published literature has actually run it that way, across four organs:

> Human fibroblasts used are WI-38 cells (embryonic lung fibroblasts) […], WS-1 cells (embryonic dermal fibroblasts) […], adult dermal fibroblasts […], adult corneal fibroblasts (keratocytes) […] and immortalized adult vocal fold fibroblasts

The two implementations cited here are the demonstration rather than the assertion: `PMID:33521022` runs it with primary adult human **dermal** fibroblasts, `PMID:32381012` with primary human **lung** fibroblasts, same assay, same endpoints. That is what licenses a module-level model instead of a fifth organ-specific one.

The entry is keyed on `PMID:34028741` (Stebler & Raghunath, *Methods Mol Biol* 2021) because its abstract is the one that states the organ-agnosticism explicitly.

## What is linked, and what deliberately is not

| `fibrotic_response` node | Linked? |
|---|---|
| Mesenchymal Cell Activation | `RECAPITULATES`, `MODERATE` |
| Excessive ECM Deposition | `RECAPITULATES`, `MODERATE` |
| Tissue Injury | no |
| Inflammatory Recruitment and Amplification | no |
| Architectural Distortion and Organ Dysfunction | no |

The three unlinked nodes are not an omission to be filled in later by this model. A fibroblast monoculture has no epithelium, no immune compartment and no organ architecture, so there is nothing in the system for those nodes to correspond to. A different model will be needed for them.

Readouts, each carrying its own evidence separate from the link's:

- **Mesenchymal Cell Activation** — α-SMA expression (`INCREASED`) and myofibroblast formation by immunofluorescence (`INCREASED`).
- **Excessive ECM Deposition** — type I and VI collagen formation, PRO-C1 and PRO-C6 (`INCREASED`), and fibronectin deposition, FBN-C (`INCREASED`).

## Limitations recorded on the links

Both are honest about what the system substitutes for:

- **Exogenous TGF-β1 replaces endogenous latent-TGF-β activation.** The injury and inflammation steps upstream in this module are bypassed, not modeled. I kept `RECAPITULATES` rather than `PERTURBS` on the activation node because the myofibroblast conversion the node names is genuinely reproduced — the same call made in #9935 for the liver plate. The caveat is in `limitations` either way.
- **α-SMA is slower than matrix output.** In the 12-day variant it does not separate from control until day 8, so a short assay can show matrix deposition while the activation marker is still flat. Worth knowing before reading a negative result from this system.
- **The matrix is a 2D in-well layer.** No 3D architecture, no stiffening, no mechanical feedback onto the fibroblast.
- **Only the production arm is measured.** The node describes a production/degradation imbalance; this assay reports production. TIMP-mediated blockade of turnover is not read out.

## One thing a reviewer will ask about

**`tissue_term` is omitted.** Every other module-level model block carries one. An organ-agnostic assay has nothing sensible to put there, and binding any single organ would assert a specificity the platform does not have — which would undercut the whole reason it belongs at the module level. There is a `notes:` line on the entry saying this, so it does not read as an oversight.

## Not in this PR

Precision-cut tissue slices (`PMID:31676376`, Bigaeva et al.) prepare slices from intestine, liver and kidney and report where fibrogenesis stops being conserved: both canonical and non-canonical TGF-β pathways activate in liver and kidney, only the canonical Smad cascade in intestine. That is a statement about the limits of this module's conserved chain, which makes it better suited to a `discussions:` entry than to a second model link. Left out here; happy to open a separate issue for it.

## Validation

`just validate` (schema + terms + references on the changed file), `just count-verified-snippets` (19/19 verified), `just validate-terms`, `just check-entity-refs`, `just check-duplicate-keys`, `just check-enum-values`, `just check-snippet-grading`, `just check-folded-hyphens`, `just check-snippet-length`, `just check-title-snippets`, `just validate-history` — all clean.

I did **not** complete `just validate-modules`. It sweeps all 166 modules and I stopped it partway; for the file this PR changes it runs the same three checks as `just validate`, which passed, so the only uncovered ground is the other 165 modules, which an appended block in one file cannot affect. CI runs it here regardless — flagging it rather than listing a check I did not finish.

Three new reference cache files were generated with `just fetch-reference`. History record added under `history/modules/fibrotic_response/`.

### DIFF
--- a/history/modules/fibrotic_response/2026-08-31T234401Z-claude-code-d2b620.yaml
+++ b/history/modules/fibrotic_response/2026-08-31T234401Z-claude-code-d2b620.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: module
+  slug: fibrotic_response
+  path: kb/modules/fibrotic_response.yaml
+session:
+  id: 2026-08-31T234401Z-claude-code-d2b620
+  timestamp: '2026-08-31T23:44:01Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/9997
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10282
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - experimental_models
+  summary: Attach directional evidence to the PRO-C1/PRO-C6 readout
+  details: |
+    Third review-response session for PR #10282. The 'Type I and type VI collagen formation (PRO-C1, PRO-C6)' readout declared direction INCREASED but its only evidence was the PMID:32381012 Methods sentence stating the biomarkers were evaluated, not that they rose - the same defect shape corrected earlier on the myofibroblast readout. Added a second evidence item quoting 'TGF-beta1 induced synthesis of PRO-C1, PRO-C6 and FBN-C as compared with unstimulated fibroblasts at all timepoints', which names exactly these two markers in exactly this direction. Kept the Methods item and narrowed its explanation to what it actually establishes - the ELISA modality and supernatant compartment the interpretation depends on - noting explicitly that it does not carry the direction. Readout-local evidence is a design property per CLAUDE.md, since a readout must stay self-describing when lifted out of its link. Branch was also rebased onto current main in this session: main had added biological_scale to all five pathophysiology nodes (PR #10257), changed the schema, and introduced check-causal-targets and check-qualifier-terms, none of which the branch's earlier CI had seen; all re-validated clean after rebase. Validated: linkml-validate, count-verified-snippets (21/21), validate-terms, check-entity-refs, check-duplicate-keys, check-enum-values, check-causal-targets, check-qualifier-terms, check-snippet-grading, check-snippet-length, check-title-snippets, check-folded-hyphens - all clean.
--- a/kb/modules/fibrotic_response.yaml
+++ b/kb/modules/fibrotic_response.yaml
@@ -416,8 +416,17 @@ experimental_models:
         evidence_source: IN_VITRO
         snippet: "Biomarkers of ECM synthesis were evaluated over time in cell supernatants using ELISAs to assess type I, III, IV, V and VI collagen formation (PRO-C1, PRO-C3, PRO-C4, PRO-C5, PRO-C6), fibronectin (FBN-C) deposition and α-smooth muscle actin (α-SMA) expression."
         explanation: >-
-          Identifies the collagen formation markers measured in this assay and
-          the compartment they are measured in.
+          Establishes the ELISA modality and the supernatant compartment these
+          markers are measured in, which this readout's interpretation depends
+          on. It does not by itself establish the direction.
+      - reference: PMID:32381012
+        reference_title: "Prolonged Scar-in-a-Jar: an in vitro screening tool for anti-fibrotic therapies using biomarkers of extracellular matrix synthesis."
+        supports: SUPPORT
+        evidence_source: IN_VITRO
+        snippet: "TGF-β1 induced synthesis of PRO-C1, PRO-C6 and FBN-C as compared with unstimulated fibroblasts at all timepoints"
+        explanation: >-
+          Reports PRO-C1 and PRO-C6 rising against unstimulated fibroblasts at
+          every timepoint measured, which is the direction this readout records.
     - name: Fibronectin deposition (FBN-C)
       target: Excessive ECM Deposition
       direction: INCREASED


gingin77:
Thanks — the blocking item was a real defect and you were right about it. Fixed, along with all three suggestions.

**The blocking one.** The `Myofibroblast formation by alpha-SMA immunofluorescence` readout claimed `direction: INCREASED` on a Methods sentence that only said the measurement was performed. Replaced with the Results sentence:

> a significant increase is seen in mRNA levels of genes encoding COL1A1, COL5A1, and FN1EDA (extracellular matrix proteins), SERPINH1 and PLOD2 (collagen-processing proteins), ACTA2 (a marker for myofibroblasts) and XBP1

That reports ACTA2 rising in dermal fibroblasts under TGF-β1, so the direction is now carried by the quote rather than asserted around it.

**Suggestion 1** — taken, with one addition. The α-SMA readout now leads with `"increased the ECM synthesis and α-SMA expression over time"`. I kept the day-8 sentence as a second evidence item rather than dropping it: the `limitations` block on that link states α-SMA does not separate from control until day 8, and removing the quote would have left that claim uncited.

**Suggestion 2** — taken. All three `PMID:34028741` items regraded to `evidence_source: OTHER`. It is a *Methods Mol Biol* protocol chapter, so `OTHER` matches how this module already grades review-type sources. That was an inconsistency on my part, not a judgment call.

**Suggestion 3** — taken. The `findings` explanation now says the four-organ list appears inside the authors' critique of earlier validation, so it establishes the breadth of cell types the platform accepts without reading as an endorsement of those studies.

Gates after the change: 19/19 snippets verified (was 18/18 — the α-SMA readout gained an item), plus `linkml-validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens` — all clean.

Re-requesting review.

gingin77:
Good catch, and correct on both. The two readouts named assays their evidence did not use, in opposite directions — I had not noticed they were inverted with respect to each other.

**Blocking item, fixed.** `Myofibroblast formation by alpha-SMA immunofluorescence` is renamed to `ACTA2 (alpha-SMA) transcript expression in dermal fibroblasts`, and the `interpretation` no longer says "in-well imaging". I checked the cache and agree there is no clean immunofluorescence quote to reach for: every α-SMA IF comparison in `PMID:33521022` is a crowder, serum, or drug contrast *under* TGF-β1, never TGF-β1 versus unstimulated. Renaming to match the evidence was the honest route.

**Suggestion 1, taken.** The lung readout's `interpretation` said α-SMA was "the direct transcriptional readout", but `PMID:32381012` measures it in the supernatant ELISA panel alongside PRO-C1 and FBN-C. Reworded to protein level. The pair now reads coherently, with organ and modality as the two distinguishing axes.

**Suggestion 2, taken.** Added a third readout on the ECM link — `Collagen-degrading enzyme transcripts (MMP1, CTSK)` at `direction: DECREASED`, from the continuation of a sentence already quoted. That changed the honest scope of the `limitations`, so I narrowed it: the degradation arm is now partly covered rather than absent, with TIMP-mediated blockade and net matrix turnover still unmeasured.

**Suggestion 3, taken.** A second history record for this session rather than an edit to the first.

Snippets now 20/20 verified. Also clean: `linkml-validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `validate-history`.

One note on the review history here: two runs raced earlier and produced an APPROVED at 21:08 followed by this CHANGES_REQUESTED at 21:11. The later one is the one I have worked from.

gingin77:
Fixed. You were right to reopen it — letting that one stand after blocking on its twin would have been the inconsistent outcome, and the shape is identical.

The `Type I and type VI collagen formation (PRO-C1, PRO-C6)` readout now carries the directional quote:

> TGF-β1 induced synthesis of PRO-C1, PRO-C6 and FBN-C as compared with unstimulated fibroblasts at all timepoints

I kept the Methods item as you suggested, and narrowed its `explanation` to what it actually establishes — the ELISA modality and supernatant compartment the `interpretation` depends on — with an explicit line saying it does not carry the direction. That way the two items are doing visibly different jobs rather than looking like one weak citation and one strong one.

Your framing of readout-local evidence as a design property rather than redundancy is the part I had not internalised across the three rounds: I was treating the sibling readout's quote as sufficient coverage for the pair, when the point is that each readout has to stand alone once lifted out of its link.

Snippets now 21/21. Also clean: `linkml-validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-causal-targets`, `check-qualifier-terms`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `validate-history`.

Two notes on branch state. I rebased onto current `main` before this push: `main` had added `biological_scale` to all five pathophysiology nodes (#10257), changed the schema, and introduced `check-causal-targets` and `check-qualifier-terms` — none of which this branch's earlier green CI had run against. All re-validated clean, and the branch is now `behind_by=0`. Third history record added rather than editing the earlier two.

PR #10266, opened 2026-08-31
Add naked mole-rat comparative models to five aging modules

cmungall:
## Summary

This PR adds naked mole-rat (Heterocephalus glaber) as a comparative negative model across five core aging modules, establishing the species as a key counterexample to standard aging pathways. The naked mole-rat is curated not as a disease model but as a naturally long-lived mammal that dissociates early steps of aging chains (stress sensing, senescence, proteostasis decline) from their late consequences (tissue dysfunction, mortality increase).

## Key Changes

**Module Updates:**
- **cellular_senescence.yaml**: Added naked mole-rat as a comparative negative model showing that senescence-associated arrest and SASP induction occur in this species, but senescent-cell accumulation and tissue dysfunction do not over a 30+ year lifespan.
- **inflammaging.yaml**: Added nmrHas2 transgenic mouse as an intervention model demonstrating that naked mole-rat hyaluronan (HMM-HA) rescues chronic low-grade sterile inflammation in short-lived mammals, with extended lifespan and improved healthspan.
- **loss_of_proteostasis.yaml**: Added naked mole-rat showing that proteostasis measures (cysteine oxidation, ubiquitination, proteasome activity) remain stable over two decades of life, unlike aging mice.
- **genomic_instability_aging.yaml**: Added naked mole-rat showing higher oxidative DNA damage than mice yet more efficient excision repair and no age-related decline in genome maintenance.
- **mitochondrial_dysfunction.yaml**: Added naked mole-rat showing exaggerated oxidative stress relative to mice while avoiding age-related tissue dysfunction—the clearest counterexample to a simple oxidative-damage-to-aging chain.

**Reference Cache:**
- Added 13 new reference files (PMID:26894765, PMID:37612507, PMID:33742113, PMID:29364116, PMID:23783513, PMID:21993625, PMID:24082110, PMID:17054663, PMID:29930219, PMID:19223593, PMID:31085801, PMID:29432174) documenting naked mole-rat biology, longevity, proteostasis, hyaluronan, senescence, and DNA repair.

**Documentation:**
- Updated module notes to clarify that naked mole-rat entries are species comparisons, not disease models, and that conforming disorder entries do NOT inherit those links.
- Added HUMAN_MODEL_MISMATCH discussion caveat to cellular_senescence and inflammaging notes.

## Implementation Details

[...]

PR #10283, opened 2026-08-31
Add computational models to aging hallmarks modules

cmungall:
## Summary

None of the aging modules in `kb/modules/` carried a `computational_models:` block, so the hallmarks-of-aging family had no link to the simulation literature even where runnable, publicly deposited models exist. This PR adds 30 models across 11 of those modules.

Each model is wired into the pathograph through `modeled_mechanisms` against a declared pathophysiology node, with `relationship`, `fidelity`, and — where the model does not really match the node — an explicit `limitations` note. Ten models carry a verified BioModels accession, so they can actually be downloaded and run.

## What was added

| Module | n | Models |
|---|---|---|
| `cellular_senescence` | 4 | Passos2010 p21-ROS feedback loop; Karin-Alon saturating-removal turnover; Verlingue2016 Boolean geroconversion; DNA-damage/senescence/stem-cell population model |
| `telomere_attrition` | 4 | Proctor-Kirkwood attrition decomposition; Proctor2006 uncapping checkpoint; Talemi2015 telomere-associated foci; telomere length-distribution dynamics |
| `mitochondrial_dysfunction` | 4 | Kowald-Kirkwood transcription-based mtDNA deletion selection; clonal-expansion hypothesis discrimination; Sizek2023 MiDAS Boolean model; 1996 network theory of ageing |
| `loss_of_proteostasis` | 3 | Proctor2005 chaperone/heat-shock; Proctor2007 ubiquitin-proteasome; UCH-L1 inclusion formation |
| `deregulated_nutrient_sensing` | 3 | Sonntag2012 IRS-dependent AMPK regulation; DallePezze2016 concurrent AMPK/mTOR activation; Smith2010 FOXO modification integration |
| `epigenetic_alterations` | 3 | Horvath multi-tissue methylation clock; DNAm GrimAge; Przybilla2014 multiscale epigenetic drift |
| `stem_cell_exhaustion` | 3 | Przybilla2014 multiscale drift; Watson2020 clonal-haematopoiesis fitness landscape; HSC clonal self-renewal and extinction |
| `genomic_instability_aging` | 2 | Watson2020 fitness landscape; Watson2021 interacting evolutionary pressures |
| `inflammaging` | 2 | EpInflammAge epigenetic-inflammatory clock; Lewkiewicz2019 naive T cell diversity |
| `disabled_macroautophagy` | 1 | Tavassoly2015 autophagy-apoptosis decision model |
| `gut_dysbiosis` | 1 | AGORA gut-microbiome metabolic reconstructions |

Two models appear in two modules each, linked to different nodes: Przybilla2014 (the molecular origin of drift in `epigenetic_alterations`, its population-level consequence in `stem_cell_exhaustion`) and Watson2020 (somatic mutation accumulation in `genomic_instability_aging`, stem-cell pool decline in `stem_cell_exhaustion`).

Also included: 27 new `references_cache/` entries for the cited papers, and 11 `history/modules/` records.

## How the references were checked

Every PMID was found by PubMed search and cached before use, rather than recalled. This was not precautionary: the first identifier recalled from memory for the Karin-Alon senescence paper resolved to an unrelated paper about NPC1 and cholesterol transport. Nothing here is cited from memory as a result.

Every BioModels accession (`BIOMD0000000087`, `91`, `105`, `287`, `580`, `640`, `705`, `824`, and `MODEL1611180000`, `MODEL1412200000`, `MODEL2312140001`) was confirmed against the BioModels API, which also supplied the model-to-publication pairing rather than my asserting it. Where the API returned no match, the entry carries `publication:` alone and no accession.

No `evidence:` blocks or snippets were added, so this PR makes no quote-level claims.

## Scope limits recorded in the data

Several of these models are a partial fit for the node they are attached to, and the entries say so rather than smoothing it over:

- The Proctor2006 telomere model is budding yeast carrying a temperature-sensitive `cdc13-1` allele, so uncapping is acute and conditional and the arrest is reversible. Marked `LOW` fidelity.
- The Talemi2015 model is calibrated on ionizing-radiation dose-response for secondary cancer risk, not on physiological attrition. The persistence of telomere-associated foci transfers; the damage input does not.
- The Horvath and GrimAge clocks are `MEASURES`, with limitations stating they are predictive rather than mechanistic — clock output should not be read as evidence for a mechanism.
- The Lewkiewicz2019 T cell model covers immunosenescence, not the sterile inflammation `inflammaging` centres on. It is included with that stated explicitly so the module does not absorb evidence belonging to a distinct process.
- AGORA has no age-specific parameterization at all; any age contrast has to be supplied as input composition from a cohort.

One entry also qualifies its own module: DallePezze2016 shows amino acids activating AMPK and mTOR concurrently rather than reciprocally, which cuts against the strict antagonism `deregulated_nutrient_sensing` currently assumes.

## Validation

Run locally on the changed files: `linkml-validate`, term validation, duplicate-key and entity-ref checks, the four snippet gates (folded hyphens, snippet length, title snippets, snippet grading), and the model-target and history pytest lanes. All pass. The 11 history records validate against `history.yaml`.

https://claude.ai/code/session_0149Uv4V3HeckoAwyBereJZB

### DIFF
--- a/history/modules/cellular_senescence/2026-08-31T222401Z-claude-code-0183f7.yaml
+++ b/history/modules/cellular_senescence/2026-08-31T222401Z-claude-code-0183f7.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: module
+  slug: cellular_senescence
+  path: kb/modules/cellular_senescence.yaml
+session:
+  id: 2026-08-31T222401Z-claude-code-0183f7
+  timestamp: '2026-08-31T22:24:01Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10283
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - computational_models
+  summary: 'Review response: correct model relationship, format, naming and description
+    (cellular_senescence)'
+  details: |
+    Review response (PR #10283, ai4c-reviewer finding 3). Corrected model_format for the Verlingue2016 geroconversion model (MODEL1611180000) from SBML to SBML-qual. The BioModels API reports the format only as 'SBML', but the deposited file declares xmlns:qual with qual:required='true' and carries 397 qual-namespace elements, so a plain SBML reader cannot load it. Verified by downloading and inspecting the file rather than trusting the API summary.
--- a/history/modules/disabled_macroautophagy/2026-08-31T222401Z-claude-code-52ab2b.yaml
+++ b/history/modules/disabled_macroautophagy/2026-08-31T222401Z-claude-code-52ab2b.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: module
+  slug: disabled_macroautophagy
+  path: kb/modules/disabled_macroautophagy.yaml
+session:
+  id: 2026-08-31T222401Z-claude-code-52ab2b
+  timestamp: '2026-08-31T22:24:01Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10283
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - computational_models
+  summary: 'Review response: correct model relationship, format, naming and description
+    (disabled_macroautophagy)'
+  details: |
+    Review response (PR #10283, ai4c-reviewer finding 1). Changed the Tavassoly2015 link to 'Autophagy Machinery Decline' from relationship MEASURES to PERTURBS. The reviewer was right: the model does not assay the age-related decline in autophagic capacity, it has that decline imposed on it as a reduction in flux parameters, which is a perturbation and not a measurement. Added a limitations field on the link stating that the decline is an input rather than an output, so the link cannot be cited as evidence that the decline occurs.
--- a/history/modules/inflammaging/2026-08-31T222400Z-claude-code-d5872a.yaml
+++ b/history/modules/inflammaging/2026-08-31T222400Z-claude-code-d5872a.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: module
+  slug: inflammaging
+  path: kb/modules/inflammaging.yaml
+session:
+  id: 2026-08-31T222400Z-claude-code-d5872a
+  timestamp: '2026-08-31T22:24:00Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10283
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - computational_models
+  summary: 'Review response: correct model relationship, format, naming and description
+    (inflammaging)'
+  details: |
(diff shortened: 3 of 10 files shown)


cmungall:
Thanks — this was a genuinely useful review, and finding 3 turned out to be a real bug that I would not have caught. All findings addressed in 1177ea676. The branch was also rebased twice onto a moving `main` before that (details at the end).

**Finding 1 (blocking) — fixed.** You are right, and the reasoning holds even though the `limitations` text you quoted sits on the sibling link. The model does not assay the age-related decline in autophagic capacity; the decline is imposed on it as a reduction in flux parameters. Changed to `PERTURBS`, and I added a `limitations` field on that link making the direction explicit:

> The decline is an input to the model, not an output of it: nothing here measures or derives the age-related loss of autophagic capacity, so this link cannot be cited as evidence that the decline occurs.

**Finding 3 — you were right, and the API would not have shown it.** I do have network access, so I re-checked all eleven accessions. Every PMID pairing re-confirmed. But on format the BioModels API is too coarse: it reports **both** Boolean models as plain `"SBML"`. Downloading the files settles it —

| Model | Declares | qual elements | Was | Now |
|---|---|---|---|---|
| `MODEL1611180000` (Verlingue2016) | `xmlns:qual`, `qual:required="true"` | 397 | SBML | **SBML-qual** |
| `MODEL2312140001` (Sizek MiDAS) | `xmlns:qual`, `qual:required="true"` | 2558 | SBML | **SBML-qual** |
| `MODEL1412200000` (Rastgou Talemi2015) | plain SBML L2V4, COPASI-exported | 0 | SBML | SBML (correct) |

Both were wrong, exactly on the two you predicted. This matters practically: a plain SBML reader cannot load either file. The other nine are genuine plain SBML (L2V1–L2V4, one L3V1). So the history records' claim that the IDs were API-confirmed was true, but the *format* rested on the API's summary rather than the file — that is the gap you found.

**Finding 5 — fixed, partly.** `Talemi2015` → `Rastgou Talemi2015`; the author indexes as "Rastgou Talemi S" and the BioModels record name I copied truncates the surname. For `Sizek2023` and `Proctor2006` the label and the publication year genuinely disagree because both mirror the BioModels record name (2023 deposit / 2024 paper; 2006 epub / 2007 issue). I kept the labels so they still match the repository record a curator would search, and recorded the real citation year in `notes` instead of silently carrying a wrong year.

**Finding 6 — fixed.** A fair catch. The claims about peripheral homeostatic proliferation and partly-preserved cell numbers are true in the wider literature but are not in PMID:31201663, and I should not have put them in a description sourced to it. Rewritten to follow the abstract: diversity falls primarily through diminished thymic output and the resulting loss of small immunoclones.

**Finding 2 — left as-is,** on your reasoning. AGORA does compute community metabolic output at that node, and `PARTIALLY_RECAPITULATES` is ruled out by the same `limitations` text, so `MEASURES` is the least-bad value the enum offers. The `limitations` field carries the caveat.

**Finding 4 — already done.** The body was rewritten at 22:06 UTC, about 30 minutes before your review posted, so you reviewed the stale version. Every discrepancy you list is gone; the current body has the correct per-module table, 30 models, 27 cache entries, and the real schema slots.

**Finding 7 — noted, no change.** Agreed it is compliant and worth doing later.

Five history records were added for these edits rather than amending the originals, since records are append-only.

---

Two things worth flagging separately.

**Your `validate-disorders` observation is worth an issue.** I can confirm it from this branch: the batched gate skips `kb/modules/` entirely, so a module-only PR gets no coverage from it. That is not hypothetical here — this PR is module-only. I have not opened the issue since it is outside this change; happy to if a maintainer wants it.

**The branch was rebased twice.** `main` moved 35 commits mid-review and conflicted with five module files (#10266, naked mole-rat animal models) and later a sixth (#10280, the rapamycin/DR hypothesis). Both were positional clashes at end-of-file, not semantic ones — those PRs appended `animal_models:` and edited `mechanistic_hypotheses:` while this one appends `computational_models:`. Each conflict was resolved by taking `main`'s file and re-appending only this PR's block, then verifying that the resolved file minus that block is byte-identical to `main`. The diff against `main` is still **0 deletions**, and #10266's naked mole-rat entries and #10280's `rapamycin_distinct_from_dietary_restriction` hypothesis are both intact.

Re-ran after every change: `linkml-validate` on all 11 modules, `check-entity-refs`, `check-duplicate-keys`, the four snippet gates, history validation, and the model/entity-ref pytest lanes (10,548 passed).

---

PR #10285, opened 2026-08-31
curate: Add GATA2 Deficiency (MONDO:0042982)

bpow:
Adds an entry for GATA2 deficiency, an inherited disorder where one working copy of the GATA2 gene is not enough. People with it progressively lose monocytes, B cells, NK cells and dendritic cells, become vulnerable to mycobacterial and HPV infection, and face a high risk of myelodysplastic syndrome or leukemia — often as teenagers. Bone marrow transplant before that transformation is what changes the outcome.

The knowledge base already had `Deafness-Lymphedema-Leukemia_Syndrome` (Emberger, MONDO:0013540), whose own description calls itself "a presentation within the broader GATA2 deficiency spectrum." The parent entry it pointed at did not exist. This adds it.

> **This description was rewritten after review.** The original claimed HPO has no "decreased NK cell count" term. That was false — see the correction below. Keeping a known-wrong statement in the PR body would outlive this thread, so it has been fixed rather than left standing with a reply buried in the comments.

## What's in the entry

Seven pathophysiology nodes wired into a connected pathograph (6 edges): GATA2 haploinsufficiency → hematopoietic stem and progenitor cell depletion → mononuclear cytopenia → impaired antimycobacterial and antiviral immunity, with clonal evolution to myeloid malignancy and, separately, the lymphatic vascular development defect branching directly off haploinsufficiency rather than off the haematological chain.

`Alveolar Macrophage Depletion` hangs off the mononuclear cytopenia and carries the pulmonary axis — alveolar macrophages depend on continuous monocyte replacement, so the defect is macrophage *supply*, which is what distinguishes it mechanistically from the hereditary PAP caused by GM-CSF receptor defects.

Fourteen phenotypes, a genetic block covering the three variant classes (coding null, ZF2 missense, +9.5 kb enhancer), and allogeneic HSCT.

## Corrections made during review

Three findings were corrections to claims this PR had made, not gaps:

- **The NK term.** The original body said HPO has no "decreased" variant. `HP:0040218` "Reduced total natural killer cell count" exists, is in the committed caches, and is used by six entries including the Emberger sibling — so this parent entry was vaguer than its own child about the same finding. My searches tried "Decreased" and "Circulating", never "Reduced", and I reported that search failure as an ontology fact. Now bound to `HP:0040218`.
[...]

PR #10286, opened 2026-08-31
curate: Add RUNX1 Familial Platelet Disorder (MONDO:0100083)

### LINKED ISSUE #10240: Curate hereditary thrombocytopenia and hematologic cancer predisposition syndrome (MONDO:0011071)
Claiming RUNX1 familial platelet disorder with associated myeloid malignancy (FPDMM) for curation.

People with an inherited RUNX1 variant have a mild, lifelong low platelet count with platelets that do not aggregate normally, and roughly a 40% lifetime risk of developing myelodysplastic syndrome or acute myeloid leukemia. The bleeding tendency is usually mild enough to be missed for years, so the leukemia is often what brings the family to attention. The mechanistically interesting part is the two-step structure: the inherited RUNX1 variant alone gives the platelet phenotype, and transformation requires later acquired hits — a second RUNX1 allele, ASXL1, or monosomy 7.

**Why now:** the knowledge base curates around eight platelet-type bleeding disorders and four entries involving somatic RUNX1 lesions in leukemia, but has no entry for the germline predisposition that connects them. `Bleeding_Disorder_Platelet-type_21` mentions "RUNX1-related familial platelet disorder" in prose without an entry behind it.

**Scope decision to make during curation.** The MONDO term claimed here is gene-agnostic — it covers the same clinical picture caused by ANKRD26 and ETV6 as well as RUNX1:

| Term | MONDO |
|---|---|
| hereditary thrombocytopenia and hematologic cancer predisposition syndrome | MONDO:0011071 (this claim, matches the open stub) |
| ...associated with RUNX1 | MONDO:0100083 |

The three genes give overlapping clinical presentations through different mechanisms, so this may be better as a grouping in `kb/groupings/` with a RUNX1 disease entry under it, rather than one entry. Planning to curate the RUNX1 entity against MONDO:0100083 and decide the grouping question from what the mechanisms actually look like once written out. Either way the MONDO:0011071 stub gets resolved in the PR.

bpow:
Adds an entry for RUNX1 familial platelet disorder with associated myeloid malignancy (FPDMM). People with an inherited RUNX1 variant have a mild lifelong low platelet count with platelets that don't work properly, plus roughly a 35–50% lifetime risk of myelodysplastic syndrome or acute myeloid leukemia. The bleeding is usually mild enough to be missed for years, so the leukemia is often what brings a family to attention.

The knowledge base had eight platelet-type bleeding disorders and four entries involving somatic RUNX1 lesions in leukemia, but nothing for the germline predisposition that connects them. `Bleeding_Disorder_Platelet-type_21` even mentions "RUNX1-related familial platelet disorder" in prose with no entry behind it.

> **Counts in this description were refreshed after review.** The original said 8 phenotypes, 2 treatments and 18 snippets; the entry roughly doubled during review and those figures now understate it. Updated so the description matches what the diff actually contains.

## What's in the entry

Seven pathophysiology nodes, connected by 5 `downstream` blocks carrying 8 edges into a single component. Haploinsufficiency and RUNT-domain dominant-negative activity are kept as **separate** molecular claims rather than merged, because they are different lesions with different functional evidence — and both converge on the same three consequences, which is what the graph now shows. The chain runs through megakaryocyte maturation arrest, the dense granule and aggregation defect, thrombocytopenia, clonal hematopoiesis with second-hit acquisition, and myeloid transformation.

Ten phenotypes, a genetic block, four treatments, a `diagnosis:` section, one animal model, and a GeneReviews-tagged `references:` block.

## The diagnosis section carries the field's most distinctive pitfall

Germline confirmation must be done on cultured skin fibroblasts, not blood or marrow: acquired uniparental disomy of chromosome 21 in the haematopoietic compartment causes preferential loss of the variant-bearing chromosome and a false-negative result. The converse error also occurs — a somatic RUNX1 variant in a sporadic myeloid malignancy misread as germline. Both directions are stated.

## One evidence item is deliberately REFUTE

The gene-therapy treatment cites the rhesus macaque competitive repopulation model, which found **no** fitness advantage for corrected over RUNX1-mutant HSPCs long term. That is evidence the approach will struggle, not evidence it works, so grading it `SUPPORT` would have inverted its meaning. Flagging it because a `REFUTE` on a treatment is unusual enough to look like an error at a glance.

## Scope decision

Curated against the RUNX1-specific term (MONDO:0100083) rather than the gene-agnostic parent the claim issue was keyed on.

A correction to something I said when proposing this: **MONDO:0011071 does not cover ANKRD26.** It has exactly two descendants — this entry and MONDO:0014536 (thrombocytopenia 5, which is ETV6). ANKRD26/THC2 is MONDO:0008555 and sits elsewhere in MONDO entirely.

Since ETV6-THC5 is not curated, a grouping would have one real member and one placeholder, so the MONDO:0011071 stub is deliberately left in the queue. The better grouping — germline myeloid malignancy predisposition across RUNX1, GATA2, ETV6, ANKRD26, CEBPA, DDX41 — is worth building separately once more members exist. That reasoning now also lives in the entry's `notes:` so it survives outside this thread.

## Corrections made during review

- **Three items graded `HUMAN_CLINICAL` were quoting a macaque paper's introduction.** `evidence_source` classifies the publication, and that publication reports a primate model. All three re-sourced to `PMID:37738626`, the 214-participant human natural history cohort; the macaque paper now appears only as `MODEL_ORGANISM`.
- **The pathograph had no edges at all** — the causal chain existed in prose and nowhere in the YAML.
- **No evidence item carried `reference_title`**, which is present in all 2,498 KB entries.

Two defects I introduced and self-reported: a duplicate `evidence` key (fixed by merging both blocks, not dropping either, per CLAUDE.md), and a snippet where I transliterated `αIIb β3` as "alpha IIb beta 3" — it passed only after normalization until corrected to the exact source text.

## Evidence and terms

**36 evidence items, all snippets verified as exact substrings**, and every `reference_title` byte-identical to its cache frontmatter.

No ontology CURIE came from the research report; its term section proposed the obsolete `GO:0043433`, which was not used. All bindings were resolved from the committed caches and exact-match OLS queries. Report references resolved 17/17 at 0.0 confabulation.

## Known validation limitation

`just validate` cannot complete on this file in the authoring environment: the reference validator escalates `PMID:37738626` to PMC full text and the session proxy intercepts TLS for that host. Disabling certificate verification is prohibited, so it was not done. Schema, terms, entity-refs, duplicate-keys and exact snippet matching all pass individually, and CI passes on a clean runner.

## Follow-up, agreed with the reviewer

`PMID:33692461` (the functional classification paper behind this entry's haploinsufficiency-vs-dominant-negative split, which currently rests on a single pedigree) and `PMID:35884491` (the human source for the second-hit gene list asserted in `genetic.notes`); `directness: INDIRECT` on two inference-step items; splitting or renaming the "Eczema and Psoriasis" phenotype, which binds only the eczema term; `therapeutic_agent` on the antifibrinolytic treatment; re-keying the GeneReviews reference from URL to PMID; `prevalence:` and `progression:`.

## Deep research provenance

Falcon returned HTTP 403 (invalid `EDISON_API_KEY` in this environment); the run fell back to `claude_code` via `--fallback-provider`, recorded in the report frontmatter as `fell_back`, `requested_provider` and `provider_attempts`.

Closes #10240


bpow:
All three CRITICAL findings are fixed, plus most of the IMPORTANT ones. Pushed in `9c0cfe79b`. I verified each finding before acting rather than taking them on trust, and two turned out to be worse or more precise than stated.

## The critical three

**1. Disconnected pathograph — confirmed, and broader than this PR.** You were right, and it was worse: **all four** entries in this batch had zero `downstream` edges, not just RUNX1. I'd described causal chains in prose and never encoded them. Nine edges added here along the chain you laid out, with both the haploinsufficiency and dominant-negative nodes converging on the same three consequences. I fixed the sibling PRs in the same sweep — [#10285](https://github.com/monarch-initiative/dismech/pull/10285) (5 edges) and [#10287](https://github.com/monarch-initiative/dismech/pull/10287) (4 blocks each). Thanks for catching a defect that spanned three PRs.

**2. `reference_title` — confirmed as universal, not merely recent convention.** I checked KB-wide rather than against the three entries you sampled: present in **2498 of 2498** entries. Backfilled on every evidence item from the cached titles.

**3. GeneReviews — found and mined.** You were right not to assert the PMID. I ran the search you suggested; the chapter is **PMID:33661592**, identifiable by its absent journal source and 1993 book date. It now backs a new `diagnosis:` section, two management treatments, the skin and lymphoid phenotypes, and penetrance, and is recorded in a `references:` block tagged `GeneReviews`.

## The one worth acknowledging directly

**Finding 7 was a genuinely sharp catch.** Three items quoted the macaque paper's *introductory* sentence and graded it `HUMAN_CLINICAL`. You're right that `evidence_source` classifies the publication, and that publication reports a primate model — so the grading was wrong regardless of what the sentence asserts. All three now cite the 214-participant natural history cohort (`PMID:37738626`), which is both correctly graded and a better source for those claims.

## Also fixed

- CL/GO bindings on three nodes. `CL:0000556`, `CL:0000037`, `GO:0030220`, `GO:0030099` all verified against both the committed cache and OLS before binding — live, none obsolete.
- Evidence added to all four previously empty items.
- Eczema/psoriasis and menorrhagia phenotypes.
- `diagnosis:` section carrying the cultured-fibroblast requirement and the chr21 UPD false-negative pitfall.
- Penetrance evidence in the genetic block.
- Macaque model promoted to `animal_models:` with `modeled_mechanisms` and a polyploidization readout, as suggested.
- Scope reasoning moved into the entry's `notes:`.

One precision note on penetrance: the cohort figure I quote is 19 of 111 patients diagnosed with a hematologic malignancy, which is a **cross-sectional proportion**, not the 35–50% lifetime risk. The `explanation` says so explicitly so the two aren't conflated.

## One thing I could not verify locally

`just validate` no longer completes on this file, and I'd rather say so than imply a green run. The reference validator escalates `PMID:37738626` to PMC full text, and this session's proxy intercepts TLS for `ncbi.nlm.nih.gov`. Disabling certificate verification is prohibited, so I didn't.

What does pass, run individually: schema validation, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, and **35/35 snippets verified as exact substrings** of the cache. I also confirmed the failure is environmental rather than content-related — `Wilms_Tumor` with 65 references validates fine locally, and the other three entries in this batch validate end to end. CI is authoritative for the full reference gate.

## Not done

The remaining under-consumption from finding 4 — `PMID:35884491`, `hae.13280`, `PMID:27112265`, `PMID:36819173`, the mouse model, and the two emerging-therapy papers. The entry has roughly doubled and now covers diagnosis, management, surveillance, penetrance and models, but I have not consumed everything cached. Happy to pull those in here rather than a follow-up if you'd prefer.

---

bpow:
Thanks for the approval, and for flagging the cache file rather than staying quiet about it.

**On the missing `url:` cache file — CI adjudicated it green.** `test (3.13)` completed successfully on `9c0cfe79b`, the current head. You were right to be unsure whether your runner's network access was masking a problem; it wasn't masking one, because the build passed on a clean runner too.

So I'm **not** pushing a fix for it. Adding the cache file would dismiss this approval and restart the review cycle to correct something CI does not consider broken, and that trade isn't worth it. Your suggestion C is the better resolution anyway — keying the tagged reference on `PMID:33661592` instead of the URL would make the tag and the 11 mined evidence items resolve to one identifier *and* remove the untracked-cache question entirely. That belongs in the follow-up PR with the rest.

One thing I learned doing exactly this on the RTPS pair, which may be useful: a `references:` block keyed on a **PMID** has its `title:` validated against the cache file, unlike a url-keyed one. So when you make that swap, the title has to be the exact cached title (`RUNX1 Familial Platelet Disorder with Associated Myeloid Malignancies.`), not a GeneReviews-style expansion. I hit that error and fixed it on RTPS1.

**On suggestion A** — you're right, and it's the sharpest of the remaining items. `PMID:33692461` being the actual source of the haploinsufficiency-versus-dominant-negative split, while that split rests on a single pedigree, is a real weakness in the entry's best modelling decision. Same for `PMID:35884491` and the uncited second-hit gene list. Both go in the follow-up.

**On suggestion B** — noted, and thanks for retracting it explicitly. A phase 1 study in *somatic* RUNX1-mutant R/R AML with no responses is genuinely not evidence for a germline predisposition entry, and saying so plainly is more useful than leaving the earlier note to stand.

Suggestions D, E and F are all correct and cheap: `directness: INDIRECT` on the two inference-step items, the unbound psoriasis half of a phenotype named for two things, and `therapeutic_agent` on a `NCIT:C15986` treatment that names three drugs in its description. All queued.

Leaving this branch untouched so the approval stands.

---

PR #10289, opened 2026-08-31
Add Bile Duct Cyst entry (MONDO:0018805)

justaddcoffee:
Curates **bile duct cyst / choledochal cyst** (`MONDO:0018805`), a congenital biliary anomaly with defined malignant potential.

## Why this target

No entry, no curation issue, `MONDO:0018805` absent from `kb/`, no children in MONDO. Verified against `main` at 2,499 entries immediately before branching.

## Model

One chain, one branch point — and the branch is the whole clinical story:

```
congenital pancreaticobiliary maljunction
  (long common channel, outside sphincter of Oddi control)
              ↓
      pancreatobiliary reflux
              ↓
  chronic biliary epithelial injury
        ↓                    ↓
cystic dilatation      hyperplasia → dysplasia → carcinoma
  (= the cyst)                      ↓
                          biliary tract carcinoma
```

That branch is why this is **excised rather than drained**, and why surveillance runs for decades: complete excision with Roux-en-Y hepaticojejunostomy carries `target_mechanisms` against *both* arms — `BYPASSES` the reflux by rerouting drainage, `INHIBITS` the neoplastic sequence by removing the at-risk epithelium.

6 nodes, 5 edges, **every edge evidenced**. Todani types I–V as subtypes, `prevalence`, `inheritance` (multifactorial — no established Mendelian gene), 5 phenotypes, TP53 as somatic driver of the carcinoma arm, MRCP as the diagnostic anchor, and `progression` covering both presentation and post-excision surveillance.

## Relationship to `Caroli_Disease.yaml` — stated up front

Todani **type V is Caroli disease**, which already has an entry. It's listed here for completeness of the classification but **deliberately not restated**, because the mechanism differs in kind: Caroli is a ductal plate malformation / *PKHD1* cholangiociliopathy, not enzyme reflux through a maljunction. MONDO agrees — the two terms sit in separate branches with no parent–child relation. Recorded in `notes` so the pathograph isn't misread as applying to type V.

## Two verification failures worth recording

**1. Fabricated `reference_title` values.** Three didn't match the cache; **two more were wrong but passed the 0.85 similarity gate**. Cause: I read titles from my own `cut -c8-92` truncated view of the frontmatter and reconstructed the tails — which invented a *German* centre for a Finnish series, and a hospital pairing for a Kashmir series.

This is the same failure class as concluding absence from a `LIMIT`ed query: **treating truncated output as complete**. All 29 titles were then rewritten *programmatically from the parsed frontmatter* rather than typed. That's now the only safe way I'll do this.

**2. Seven snippets written as paper titles** rather than sentences — replaced with real abstract text carrying the actual claims (pressure measurements, TP53 frequency, the 20–30× cancer risk).

## Ontology note

HPO has no choledochal-cyst term. Checked across five patterns **without a `LIMIT`** — `bile duct cyst`, `biliary cyst`, `choledochal`, `bile duct dilat`, `dilatation of the bile duct` — returning only `HP:0005209 Intrahepatic bile duct cysts`, which is the *Caroli* lesion and wrong here. The structural phenotype therefore binds parent `HP:0035013` with a specific `preferred_term`, and the search is documented in `notes` so the next curator doesn't repeat it.

## Validation

```
just validate      # 31/31 snippets verified
just compliance    # 98.8% global / 100.0% weighted
```

All seven content gates zero findings. History record included and validates.

## Known gaps, in notes

`datasets`; `clinical_trials`; the female predominance and East Asian clustering (described in the research but with no cleanly quotable ratio in the cached abstracts); `histopathology` as a structured section; and endoscopic management of the type III choledochocele, which differs from the excisional approach modelled here.

### DIFF
--- a/history/disorders/Bile_Duct_Cyst/2026-09-02T122906Z-claude-code-535159.yaml
+++ b/history/disorders/Bile_Duct_Cyst/2026-09-02T122906Z-claude-code-535159.yaml
@@ -1,0 +1,37 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Bile_Duct_Cyst
+  path: kb/disorders/Bile_Duct_Cyst.yaml
+session:
+  id: 2026-09-02T122906Z-claude-code-535159
+  timestamp: '2026-09-02T12:29:06Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10289
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - phenotypes
+  - genetic
+  - pathophysiology
+  summary: 'Review round 2: dilatation evidence on the namesake phenotype, KRAS early-event
+    evidence and COOPERATING, causal edge restored'
+  details: |
+    Three items from review round 2, all fixed with sentences from files this entry already cited.
+
+    The namesake Bile Duct Cyst phenotype had no evidence that mentioned dilatation. It now leads with PMID:31341359 "Choledochal cysts (CCs) are rare bile duct dilatations, intra-and/or extrahepatic", with the pressure/volume/amylase sentence kept as a second item characterising which cysts these are.
+
+    The KRAS block asserted "Early but Not Maljunction-Specific" and evidenced only the second half. Added PMID:17187167 placing k-ras among the early events alongside microsatellite instability, COX-2 and bcl-2. Changed relationship_type from SOMATIC_DRIVER to COOPERATING, which is what the block's own notes argue; SOMATIC_DRIVER contradicted the text.
+
+    The Chronic Biliary Epithelial Injury to Cystic Dilatation edge had been softened to "accompanies" when the pressure direction was corrected. That over-corrected. The causal step is available in PMID:31341359, which states the long common channel facilitates reflux of pancreatic juice into the biliary tree, causing increased pressure and possible proteolysis within the CBD culminating in ductal dilation. The edge now asserts causation on that sentence, with the PMID:24604978 pressure finding retained as the qualifier identifying proteolysis rather than pressure as the operative part in maljunction cysts.
+
+    Validation: 50/50 snippets verified, term validation passes, weighted compliance 100.0 percent, all content gates clean.
--- a/kb/disorders/Bile_Duct_Cyst.yaml
+++ b/kb/disorders/Bile_Duct_Cyst.yaml
@@ -246,17 +246,27 @@ pathophysiology:
       constitute this injury, and shows they are worst in the low-pressure maljunction cysts.
   downstream:
   - target: Cystic Dilatation of the Bile Duct
-    description: Sustained enzyme exposure through a long common channel accompanies the
-      high-volume, low-pressure dilated cyst.
+    description: Sustained enzyme exposure through a long common channel proteolyses the duct
+      wall and produces the high-volume, low-pressure dilated cyst.
     evidence:
+    - reference: PMID:31341359
+      reference_title: 'Choledochal cysts: Similarities and differences between Asian and Western countries.'
+      supports: SUPPORT
+      evidence_source: HUMAN_CLINICAL
+      snippet: The long common channel associated with APBJ facilitates reflux of pancreatic
+        juice into the biliary tree, causing increased pressure and possible proteolysis within
+        the CBD culminating in ductal dilation
+      explanation: States the causal step this edge asserts, from refluxed pancreatic juice
+        through proteolysis to ductal dilatation.
     - reference: PMID:24604978
       reference_title: Correlation of intracystic pressure with cyst volume, length of common channel, biochemical
         changes in bile and histopathological changes in liver in choledochal cyst.
       supports: SUPPORT
       evidence_source: HUMAN_CLINICAL
       snippet: Low-pressure cysts have high volume and higher levels of amylase and lipase
-      explanation: The dilated high-volume cysts are the low-pressure, high-amylase ones, so
-        enzyme exposure rather than raised pressure is what this edge asserts.
+      explanation: Qualifies the pressure half of that account - in the maljunction cysts the
+        dilated ones are the low-pressure, high-amylase ones, so enzyme exposure rather than
+        raised pressure is the operative part.
   - target: Hyperplasia-Dysplasia-Carcinoma Sequence
     description: Repeated injury and repair drive a stepwise neoplastic progression in the
       biliary epithelium.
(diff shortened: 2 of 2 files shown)


justaddcoffee:
All 11 review items addressed. Pushed as a single commit.

## The CRITICAL finding was right, and it was the real error

I had the pressure relationship backwards. The entry asserted that raised intracystic pressure drives the cystic dilatation. PMID:24604978 — the paper I cited for it — reports the inverse:

> Thus it could be referred that large volume cyst have lower pressure than those with smaller volumes

> **High pressure cysts had normal pancreaticobiliary junction.**

That second sentence is decisive: the high-pressure cysts are the ones *without* a maljunction, so they are not the cysts this entry models at all. The maljunction cysts here are the low-pressure, high-volume, high-amylase ones, and the cyst wall damage tracks enzyme exposure rather than pressure. The node description and its incoming edge were rewritten, and I recorded the correction in `notes` rather than silently replacing the text, so a later reader can see what was wrong.

## Item 7 — GO:0015721 removed

You were right. The node describes pancreatic enzymes moving retrograde into the biliary tree; `bile acid and bile salt transport` names a different substance moving the other way. It was a lexical match on "bile". Dropped, no replacement — GO has no term for this and a wrong one is worse than none.

## Items 2, 3, 4 — aims/methods snippets replaced

Six snippets quoted aims or methods sentences, which assert nothing. Replaced with the result sentences that carry the claim:

| Where | Was | Now |
|---|---|---|
| Reflux node | "This study was undertaken to establish correlation…" | "In the present study it was found that pressure inversely correlated with the long common channel (P= 0.001)." |
| Injury edge | "Cyst wall was examined for ulceration, inflammation…" | "It appears that refluxing amylase and lipase may cause these changes in the cyst wall" |
| Injury node | "ulceration, inflammation, fibrosis, and metaplasia" (fragment) | "ICCP inversely correlated with cyst wall changes (P = 0.003, 0.0001, 0.023, 0.0013, respectively)" |
| `Bile Duct Cyst` phenotype (item 4) | methods sentence, never mentioned dilatation | "Low-pressure cysts have high volume and higher levels of amylase and lipase" |
| Treatment (item 3) | subtype-malignancy sentence | PMID:28364277 "Appropriate management consists of prompt, complete cyst excision followed by restoration of biliary enteric continuity when necessary." |

Also replaced the truncated `"TP53 alteration more than KRAS mutation"` fragment with the full stepwise sentence (10 / 10 / 38 / 75%, p < 0.01) at all three sites.

## Item 5 — carcinoma sequence node

Now leads with PMID:17187167 naming the sequence directly — *"Pathological findings strongly suggest a hyperplasia-dysplasia-carcinoma sequence in carcinogenesis of pancreatico-biliary maljunction (PBM)."* — with the TP53 series demoted to molecular corroboration. Added the 20–30× cholangiocarcinoma risk figure to the carcinoma endpoint.

## Item 6 — phenotypes added

Cholangitis (`HP:0030151`), Pancreatitis (`HP:0001733`), Choledocholithiasis (`HP:6001269`).

One divergence: you suggested `HP:0001081` Cholelithiasis. I used `HP:6001269` Choledocholithiasis, because the evidence sentence says choledocholithiasis (stones in the duct, which is where the stasis is) and you flagged separately in this session that I have a habit of binding generic ancestors. If you'd rather have the broader term for cross-disease queryability, say so and I'll swap it.

## Items 8–11 — all added

- **`mechanistic_hypotheses`**: CANONICAL reflux theory and ALTERNATIVE congenital-obstruction theory, both evidenced from PMID:25588714, which states plainly that "The exact etiology of CCs is currently unknown". The pathograph follows the canonical one; the alternative is recorded because it is unresolved, not adopted.
- **`animal_models`**: the surgical APBDU mongrel puppy (PMID:17021737), which produces dilatation and epithelial hyperplasia from a maljunction alone and so is the most direct test of the reflux theory; and the DBTC rat (PMID:26176076), which orders HDAC1 ahead of COX-2. Limitations stated on each — three months against a human course in decades for the dog, and no maljunction at all in the chemically induced rat.
- **KRAS** in `genetic`, recorded with the *negative* result: 16% in maljunction-associated vs 8% in non-associated gallbladder cancer, P = 0.272 (PMID:30882917). It is early but not maljunction-specific, which is worth stating explicitly since the naive reading is that it is a driver here.
- **Top-level `references`**: 15 entries, generated programmatically from the parsed cache frontmatter rather than typed — which is the fix for the fabricated-title failure earlier in this PR.

## Validation

```
All validations passed!
Snippets checked: 47/47 verified against cached references
Term validation: ✅ passed
Weighted Compliance: 100.0%
```

Gates: `check-reference-titles`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-duplicate-keys`, `check-entity-refs` — all pass. `validate-graphs` fails repo-wide on 312 pre-existing target errors, none of them in this file.

Not done: `datasets` and `clinical_trials`, both still listed as extension points in `notes`.

justaddcoffee:
All three remaining items fixed, plus the non-blocking one. Pushed.

**1. Namesake phenotype** — you're right, "Low-pressure cysts have high volume…" is a statement about pressure and amylase, not about the structural lesion. It now leads with PMID:31341359 *"Choledochal cysts (CCs) are rare bile duct dilatations, intra-and/or extrahepatic"*, with the pressure/volume/amylase sentence kept as a second item saying **which** cysts these are.

**2. KRAS** — both parts fixed:
- Added PMID:17187167 evidencing the "early" half: *"While microsatellite instability, k-ras mutations, expression of COX-2 and bcl-2, and increased telomerase activity seem to occur early; involvement of cyclin D1, beta-catenin, DPC-4/Smad4 and p53 appear later in carcinogenesis."*
- `relationship_type: SOMATIC_DRIVER` → `COOPERATING`. You're right that it contradicted the block's own notes. I'd reached for SOMATIC_DRIVER by analogy with the TP53 block above it without checking it against what the text actually claimed.

**3. Non-blocking edge, taken** — and worth naming, because I over-corrected. When I fixed the pressure direction I softened the injury → dilatation edge to "accompanies", treating the whole causal account as suspect because one part of it was. That was wrong: PMID:31341359 states the causal chain plainly — *"The long common channel associated with APBJ facilitates reflux of pancreatic juice into the biliary tree, causing increased pressure and possible proteolysis within the CBD culminating in ductal dilation"*. The edge now asserts causation on that sentence, with the PMID:24604978 pressure finding retained as the qualifier that identifies **proteolysis rather than pressure** as the operative part in maljunction cysts. That reads better than either version: the mechanism is causal, and the pressure clause in the review source is the part that doesn't hold for this subgroup.

**On the critical item** — thanks for saying so. Worth recording for anyone reading later: the paper genuinely contradicts itself, with one body sentence saying "directly correlated" against the abstract, Discussion and Conclusion all saying inverse. That's a case where quoting the paper faithfully and quoting it correctly come apart.

**Validation**
```
All validations passed!
Snippets checked: 50/50 verified against cached references
Term validation: ✅ passed
Weighted Compliance: 100.0%
```
Gates: reference-titles, snippet-length, snippet-grading, title-snippets, folded-hyphens, enum-values, duplicate-keys, entity-refs — all clean. History record added and validates.

PR #10297, opened 2026-08-31
curate(Brachyolmia): bind MONDO:0007232 to the AD-TRPV4 subtype (#10291)

### LINKED ISSUE #10291: Curate autosomal dominant brachyolmia (MONDO:0007232)
Curate a dismech entry for **autosomal dominant brachyolmia** ([MONDO:0007232](https://monarchinitiative.org/MONDO:0007232)) — a mild, non-lethal skeletal dysplasia with short trunk, platyspondyly and normal limb length, caused by gain-of-function variants in the mechanosensitive calcium channel TRPV4.

Preflight: nothing in `kb/` on `origin/main` mentions TRPV4 or brachyolmia, so the whole TRPV4 skeletal-dysplasia axis is currently uncurated. The open PR monarch-initiative/dismech#10214 curates *brachyolmia-amelogenesis imperfecta syndrome* (MONDO:0011018, LTBP3), which is a different disease.

Likely shape: one Disease entry anchored on TRPV4 gain-of-function. The TRPV4 skeletal spectrum (metatropic dysplasia, SMD Kozlowski, spondyloepiphyseal dysplasia Maroteaux, parastremmatic dysplasia) are separate MONDO concepts, so they belong in a future grouping rather than as `has_subtypes` here.

Stub: `stubs/Autosomal_Dominant_Brachyolmia.yaml`

aleixpuigb:
Autosomal dominant brachyolmia (MONDO:0007232) came up in the curation queue, but it is already curated — it is the `AD-TRPV4` entry under `has_subtypes` in `kb/disorders/Brachyolmia.yaml`, which is anchored on the parent term MONDO:0015262 and covers the TRPV4, PAPSS2 and LTBP3 forms together. Splitting it into its own disorder file would duplicate the TRPV4 pathograph across two entries.

So the queue entry resolves to `entry_type: SUBTYPE`. Rather than record that only in a stub that then gets deleted, this binds the term where it will keep mattering: `subtype_term: MONDO:0007232` on the `AD-TRPV4` subtype. Per the MONDO-coverage rule in `CLAUDE.md`, a `has_subtypes` term counts as curated, so the concept is now grounded and retires from the queue for good instead of being re-nominated later.

## Changes

- `kb/disorders/Brachyolmia.yaml` — added `subtype_term` (MONDO:0007232, *autosomal dominant brachyolmia*) to the `AD-TRPV4` subtype. No mechanism, evidence, or phenotype content was touched.
- `stubs/Autosomal_Dominant_Brachyolmia.yaml` — deleted.
- `cache/mondo/terms.csv`, `cache/enums/diseaseorsubtypeterm_*.csv` — the one new MONDO row the binding introduced, written by the validators.
- `history/disorders/Brachyolmia/` — history record for the session.

The other two subtypes are unbound too, and I left them alone. `AR-PAPSS2` is out of scope for this claim, and the LTBP3 form maps to MONDO:0011018, which is being curated as its own entry in #10214 — binding it here would collide with that PR.

## Validation

- `just validate kb/disorders/Brachyolmia.yaml` — passed
- `just validate-terms kb/disorders/Brachyolmia.yaml` — passed
- `just validate-history history/disorders/Brachyolmia/2026-08-31T210405Z-claude-code-f5dc54.yaml` — no issues

Incidental `references_cache/` churn produced by the validator run (frontmatter quote normalization, a `full_text_attempted` flag) was reverted rather than committed, since it is unrelated to this change.

Closes #10291

---


aleixpuigb:
Thanks — all three suggestions are fair, and here's what I did with each.

**AR-PAPSS2 / MONDO:0019666.** Filed as #10307. I checked the term before writing it up: synonyms include "brachyolmia 4 with mild epiphyseal and metaphyseal changes" and "BCYM4", xrefs OMIM:612847 and Orphanet:93282, and the stub is open and `UNDECIDED`. I agree it should not be a drive-by binding here — the epiphyseal and metaphyseal changes are in the concept's name, so it may well be wider than "AR brachyolmia", and that deserves a real read of the PAPSS2 literature rather than an inference from labels. The issue lays out both defensible outcomes and cites this PR as the precedent for the TRPV4 case.

**LTBP3 / #10214.** Agreed, and agreed it is #10214's job. Noting here for the record that the reconciliation will need to go one way or the other — either that entry's MONDO:0011018 binding makes this file's `LTBP3` subtype redundant, or the subtype stays and the standalone entry needs a `mappings` relationship to it. I have not touched either.

**Empty `links.prs`.** Correct, and I'm leaving it. `just new-history --pr` needs a PR number that did not exist when the record was written, and `history/` is append-only — rewriting a record to backfill a convenience link would be a worse precedent than the missing link. If this is a recurring annoyance the fix belongs in the scaffolding (e.g. `new-history` accepting a branch name and resolving the PR later), not in individual records.

No code changes; the PR is unchanged from the approved commit.

---

PR #10302, opened 2026-08-31
feat(modules): tag mechanism modules with study-area categories

caufieldjh:
## What

Mechanism modules can now be tagged with the **areas of study** they are relevant to. Each tag renders as a colour-coded pill on the module index cards and at the top of each module page.

The semantics are deliberately modest: a pill asserts *"this module is relevant to this area of study."* It is a browsing and discovery aid — not a mechanistic claim, and not a classification of the disorders that conform to the module (those keep using `classifications`).

## How

- **New `ModuleCategoryEnum`** — TOXICOLOGY, PHARMACOLOGY, ONCOLOGY, INFECTIOUS_DISEASE, IMMUNOLOGY, NEUROSCIENCE, DEVELOPMENTAL_BIOLOGY, METABOLISM, AGING.
- **New multivalued `module_categories` slot** on the `Disease` class (modules validate against it). Disorder entries keep the unrelated free-text `categories` slot for nosological grouping.
- **The vocabulary and its prose live only in the schema.** Each permissible value carries a `title` (the pill label) and a `description` (the pill's hover text and the index legend's blurb). The renderer reads the enum at build time, so a category cannot be renamed on a page without being renamed in the schema.

### Colours are generated, not curated

Each category takes the next **golden-angle** step around the colour wheel at fixed saturation and lightness. Consequences worth knowing:

- Adding a category is a **schema edit alone** — no colour to choose, no renderer change.
- **Appending** a value leaves every existing hue untouched; only reordering or removing one reshuffles.
- The fixed lightness pair clears **WCAG AA over all 360 hues** (worst case 5.4:1, in the yellow-green band near hue 60), so a future category is legible by construction rather than by luck.
- **This holds to 13 categories.** Hue separation stays ~20° through the 13th and drops to ~12° at the 14th, where `test_every_category_gets_a_visually_distinct_hue` goes red — the tripwire for the palette needing a real decision rather than another append. Documented in `CLAUDE.md`.

A first attempt hashed the enum key straight to a hue. That is maintenance-free but does not *spread*: it landed Oncology and Infectious disease three degrees apart, which is one colour with two names. Hence the golden angle.

## Backfill: 141 of 166 modules

Assignment was made per-module from each module's own name and description, not pattern-matched on filename.

| Category | Modules |
|---|---:|
| Neuroscience | 36 |
| Pharmacology | 30 |
| Infectious disease | 28 |
| Oncology | 25 |
| Developmental biology | 25 |
| Metabolism | 24 |
| Immunology | 16 |
| Aging | 14 |
| Toxicology | 11 |

**Toxicology** covers the four treatment-toxicity modules plus `vegf_inhibitor_hypertension`, and the modules whose trigger arm actually models a toxicant or drug exposure: parkinsonism (MPTP / rotenone / paraquat / manganese), chemotherapy-induced peripheral axonal degeneration, cigarette-smoke emphysema, inhaled-toxicant alveolar barrier failure, anthracycline/alcohol cardiomyopathy, and beryllium granuloma.

**25 modules are left untagged on purpose** — chiefly the cardiovascular, dermatology, renal, GI, hepatobiliary and haematology ones. The starter enum is a set of *disciplines*, not a partition of the corpus, and a wrong pill is worse than no pill. Those want a new category added rather than an existing one stretched.

## Two judgment calls, both reviewed

1. **`cardiac_ion_channel_repolarization` is NOT tagged TOXICOLOGY**, even though `CLAUDE.md` named it as the drug-induced long-QT toxicity target. The module scopes itself to inherited channelopathies *in structurally normal hearts*, and its only "acquired" mention sits inside a quoted snippet. The reviewer independently agreed the docs were the statement out of step, so **this PR now corrects `CLAUDE.md`** rather than leaving both statements standing.
2. **`amyloidogenesis` stays untagged** after review discussion. `loss_of_proteostasis` carries AGING because it self-identifies as a Lopez-Otin hallmark-of-aging module; `amyloidogenesis` is precursor-agnostic across AL, AA, ATTR, Alzheimer and type 2 diabetes. Alzheimer being one of five named conformers no more makes it a neuroscience module than type 2 diabetes makes it a metabolism one. Reasoning recorded in the history record.

## Also

- `just list-modules` prints categories and includes them in its filter, so `just list-modules toxicology` answers from the console the same question the index pills answer in the browser.
- Documented in `CLAUDE.md` (new *Module categories* section) and in the `create-module` skill.

**No mechanism content changed** — no pathophysiology node, evidence item, ontology term, causal edge or treatment was touched in any module.

## Validation

- Full `pytest` on the first commit: **50,727 passed, 41 skipped**, 0 failures.
- `linkml-validate` over all 166 module files against `Disease`: no issues.
- Negative test: an out-of-vocabulary value (`TOXICOLOGYY`) is **rejected** by the validator, so the vocabulary is enforced and not merely documented.
- `just validate-module kb/modules/parkinsonism_dopaminergic_degeneration.yaml` — schema + terms + references, 12/12 snippets verified.
- Trial-merged `origin/main` locally (it gained #10266, which touches five of the same aging modules) and validated the *merged* tree — no duplicate keys, tags intact. That is the #8623 shape where two PRs merge cleanly and only `main` goes red.
- `just check-duplicate-keys` (4,261 files), `just check-entity-refs` (2,790 files), `uvx ruff check .` — all clean.
- `tests/test_render_module_categories.py` (8 tests): enum-to-display fidelity, hue separation, pill ordering, unknown-value handling, page rendering, legend suppression when unused, a guard that every committed value is in the enum, and a guard that the slot stays off disorder entries.

**Not run:** `just validate-terms-schema` — OLS was unreachable from this environment and it failed on an unrelated MONDO term. The new enum declares no `meaning:` values, so there should be nothing for it to check, but that is reasoning rather than a passing run.

**`scope-override` label:** this PR deliberately touches `kb/` and infra together — the backfill cannot validate without the enum, and the enum is inert without the backfill. See the comment thread.

Derived `pages/modules/*.html` are intentionally **not** committed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -563,7 +563,12 @@ members with `just list-modules`, do not assume this list is exhaustive):
   modules already double as toxicity targets without a separate "side effect"
   class (`peripheral_axonal_degeneration` for chemotherapy-induced peripheral
   neuropathy, `cardiomyopathy_maladaptive_remodeling` for anthracycline
-  cardiotoxicity, `cardiac_ion_channel_repolarization` for drug-induced long-QT).
+  cardiotoxicity). `cardiac_ion_channel_repolarization` is **not** one of them,
+  despite drug-induced long QT being a real entity: that module scopes itself to
+  inherited arrhythmia syndromes *in structurally normal hearts*, its notes list
+  only heritable syndromes, and its sole mention of acquired repolarization
+  change sits inside a quoted snippet. An acquired drug-induced long-QT module
+  would be a new module, not a second use of that one.
 - **Antimicrobial drug mechanisms** — antibacterial target modules (cell wall,
   ribosome, topoisomerase, RNA polymerase, folate), antifungal and antiviral
   counterparts, plus pharmacokinetic *gating* modules such as
@@ -616,6 +621,12 @@ module_categories:
   saturation/lightness), so adding a category is a schema edit alone — no colour
   to choose and no renderer change. Appending a value leaves existing hues
   untouched; reordering or removing one reshuffles them.
+- **That holds up to 13 categories.** The golden-angle walk keeps a minimum
+  separation of ~20° through the 13th value and drops to ~12° at the 14th, where
+  `test_every_category_gets_a_visually_distinct_hue` goes red. That test is the
+  tripwire, not a nuisance: at 14 the palette needs a real decision (a second
+  visual dimension, or grouping categories into families), and the enum has
+  outgrown "just append a value".
 - The slot lives on the `Disease` class because modules validate against it. It is
   intended for `kb/modules/` entries; disorder entries use the separate free-text
   `categories` slot for nosological grouping, which is unrelated.
--- a/history/schema/ModuleCategoryEnum/2026-08-31T210434Z-claude-code-25bb33.yaml
+++ b/history/schema/ModuleCategoryEnum/2026-08-31T210434Z-claude-code-25bb33.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10302
   urls: []
 events:
 - type: EDIT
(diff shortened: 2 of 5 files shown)


caufieldjh:
### Why this PR carries `scope-override`

The **Check curation PR scope** guard failed on the first run. It fires when a PR touches `kb/` *and* files outside the curation allowlist, which is the right guard for its purpose — stopping a curation PR (new disease entry) from quietly dragging infra along with it.

This PR is the other shape: a schema + renderer feature whose whole point is a vocabulary that `kb/modules/` entries then use. The infra half (`ModuleCategoryEnum`, the `module_categories` slot, the two templates, `render.py`, `project.justfile`, docs, tests) and the `kb/` half (the 139-module backfill) are one change — the backfill cannot validate without the enum, and the enum is inert without the backfill. Splitting them would mean the second PR sits red until the first merges, and the same change gets reviewed twice in order.

So the mixed scope is intentional, which is exactly the condition the check's own message describes, and the label was added with the repo owner's explicit go-ahead in this session rather than on my own initiative. The label was applied *after* the failure and the job re-run; no code changed between the two runs.

Flagging it here rather than leaving a green check with an unexplained label on it.

caufieldjh:
Thanks — all verified independently before acting, and the numbers reproduce. Pushed in 2c6f606.

**Accepted:**

- **The two retina modules.** You're right and this was an inconsistency in my assignment rather than a difference between the modules: `phototransduction_cascade_dysfunction` ends at *"cGMP and Calcium Overload Handoff to Photoreceptor Degeneration"* and `retinoid_visual_cycle_disruption` at *"Progressive Photoreceptor and Retinal Pigment Epithelium Degeneration"* — the same sensory-neuron death I tagged `photoreceptor_degeneration` and `glaucoma_optic_neuropathy` NEUROSCIENCE for. Both now tagged; 139 → 141 tagged, 27 → 25 untagged, and the history record's counts moved with them.
- **The `cardiac_ion_channel_repolarization` docs mismatch.** Fixed in `CLAUDE.md`, in your framing — the docs were the statement out of step, and an acquired drug-induced long-QT module would be a *new* module rather than a second use of that one. Leaving both statements standing would have set up the next curator to tag it.
- **The 13-category ceiling.** Reproduced exactly: min hue gap holds at 20° through 13 and drops to 12° at the 14th, where `test_every_category_gets_a_visually_distinct_hue` goes red. `CLAUDE.md` now says so, and frames the test as the tripwire for the palette needing a real decision rather than as a nuisance to raise the threshold on.
- **The disorder-entry guard.** Added `test_module_categories_stays_off_disorder_entries`. Good catch that the slot is schema-legal anywhere on `Disease`.
- **`links.prs`.** Now carries 10302.

**Also found while re-checking, which is on me:** `tests/test_render_module_categories.py` had a `zip()` that ruff wants as `itertools.pairwise()`. I ran `uvx ruff check .` *before* writing the test file and not again after, so it went out unlinted. Fixed; `ruff check .` is clean now. Worth noting the first CI run couldn't have caught it either — the job died at the scope step before reaching lint.

**Declined — `amyloidogenesis`:** the parallel to `loss_of_proteostasis` doesn't hold, because that module self-identifies as *"the hallmark of aging (Lopez-Otin et al.) termed loss of proteostasis"* — it is a geroscience module by its own framing, which is why it carries AGING. `amyloidogenesis` is precursor-agnostic: its trigger node is the generic *"Amyloidogenic Precursor Protein"*, and its named conformers span AL (a plasma-cell dyscrasia), AA (inflammatory), ATTR, Alzheimer and type 2 diabetes. Alzheimer being one of five no more makes it a neuroscience module than type 2 diabetes makes it a metabolism one. It's a protein-biophysics module with no honest single area, so it stays in the untagged set — reasoning recorded in the history record so it doesn't get re-litigated from scratch.

**Noted, not done:** tooltips being hover-only is real, but a legend on a detail page carrying one or two pills seemed like more furniture than it earns. If category filtering ever lands on the index, that's the natural moment to revisit both.

PR #10303, opened 2026-08-31
curate: Add Developmental and Epileptic Encephalopathy 55 (MONDO:0033364, PIGP)

### LINKED ISSUE #10292: Curate developmental and epileptic encephalopathy, 55 (MONDO:0033364)
Curate a dismech entry for **developmental and epileptic encephalopathy, 55** ([MONDO:0033364](https://monarchinitiative.org/MONDO:0033364)) — an autosomal recessive inherited GPI-anchor deficiency caused by biallelic PIGP variants, presenting with early-onset refractory seizures and profound developmental impairment.

Preflight: `kb/` on `origin/main` has no PIGP entry and no GPI-anchor-biosynthesis coverage at all, so this also opens up a mechanism area the KB does not yet describe.

Likely shape: one Disease entry on the PIGP-specific pathograph (GPI biosynthesis step 1 / GPI-GlcNAc transferase complex → reduced surface GPI-anchored proteins → cortical network hyperexcitability), with the shared inherited-GPI-deficiency mechanism noted for a possible future module.

Stub: `stubs/Developmental_And_Epileptic_Encephalopathy_55.yaml`

aleixpuigb:
New entry for DEE55, an autosomal recessive inherited GPI deficiency caused by biallelic PIGP variants. PIGP had no coverage in `kb/` at all, and neither did the first step of GPI anchor biosynthesis.

## The mechanism, and what the entry refuses to claim

PIGP is a subunit of the enzyme that catalyses the *first* step of GPI anchor assembly, so reduced function throttles the pathway at its entry point. The effect is not the loss of one protein but a class-wide drop in the surface density of the 150-plus human proteins that reach the membrane on a GPI anchor — a class that is disproportionately important in neurogenesis and synaptic function. The chain is:

biallelic hypomorphic PIGP variants → impaired first step of GPI anchor biosynthesis → reduced GPI-anchored protein surface expression → disrupted neurodevelopment and cortical network excitability → refractory early-onset epilepsy **and** profound developmental impairment

Those last two are siblings, not a cause and its consequence. That is what a developmental *and* epileptic encephalopathy is, and the entry models it that way.

The edge into the neurodevelopmental node is `INDIRECT_UNKNOWN_INTERMEDIATES`. No study attributes the phenotype to any particular GPI-anchored protein, so naming one would assert more than the sources support. CD16 appears in the entry as a measured readout and diagnostic marker, not as the pathogenic protein.

The entry also says "reduced", never "abolished". Complete GPI deficiency is embryonic lethal, so every viable genotype is hypomorphic — the founding siblings retained detectable PIGP mRNA, and the recurrent c.384del is predicted to yield a longer-than-wild-type protein with impaired rather than absent function.

## Knowledge gaps recorded

Three, each attached to the node it bears on: which GPI-anchored protein mediates the phenotype; whether residual PIGP activity predicts where a patient falls on the spectrum from prenatal-lethal malformation through MCAHS to classic DEE; and the absence of a GPI-anchor-biosynthesis module in `kb/modules/`. The knowledge base already carries several inherited GPI deficiencies each restating the same upstream chain, which is what modules exist to absorb — so this entry declares no `conforms_to` and says why rather than leaving it looking like an oversight.

## Deep research

One OpenScientist run, committed. Its `reference_validation` frontmatter records **12/12 references resolved, 0 unresolved, 0 off topic, and 7 of 12 assessed as on topic** — the remaining 5 are undecided rather than cleared, which is the expected shape when a record has no abstract to weigh. (An earlier version of this PR body said "none off topic" in a way that implied all 12 were positively cleared; that was my error, caught in review.)

The run earned its place: it surfaced four sources the first pass from the primary papers had missed — the second family with the recurrent c.456delA allele (PMID:31139695), the group-level constraint that complete GPI deficiency is embryonic lethal (PMID:25803904), the 83-patient inherited-GPI-deficiency cohort (PMID:38456468), and the nutrition review behind the treatments block (PMID:35562242). Every quote used was re-verified against the fetched reference cache rather than lifted from the report.

One note for anyone reading that report: `just preflight-dr` warns that "GPI" is a rival gene at 72 mentions. That is a false positive — "GPI" there is glycosylphosphatidylinositol, the anchor, not the glucose-6-phosphate isomerase gene. The report's OMIM number matches MONDO's. I reported this false-positive class on #9826.

## One error caught in curation

`hgnc:8967` is **PIGN**, not PIGP. `linkml-term-validator` flagged the label mismatch; the correct id is `hgnc:3046`, confirmed against the HGNC REST API. Worth mentioning because the two symbols differ by one letter and both are GPI-pathway genes, so the wrong one looks right.

## Diagnostics worth flagging to reviewers

Two deliberately negative diagnostic items, both of the form "a negative result here does not exclude the diagnosis":

- Hyperphosphatasia is strong evidence of inherited GPI deficiency *when present*, but it arises from late-pathway defects (Mabry syndrome being the type case). PIGP acts at the first step, so a normal alkaline phosphatase says nothing.
- A 100-gene epileptic-encephalopathy panel that carried PIGA, PIGG, PIGN and PIGT but not PIGP returned negative in a patient who turned out to have this disease.

## Review round

Seven blocking items were raised and all are addressed — see [the reply comment](https://github.com/monarch-initiative/dismech/pull/10303#issuecomment-5485937921) for detail. The substantive ones: a `prevalence_class` that contradicted its own `rate_per_100000`; five missing phenotypes (thin corpus callosum, ventriculomegaly, cerebral atrophy, progressive microcephaly, spastic tetraplegia) all quotable from a full text already in this PR; the uncited ">150 GPI-anchored proteins" claim; `genetic_context` scalars that contradicted their own description; and missing evidence on the antiseizure treatment.

## Validation

- `just validate` — passed, **41/41 snippets verified** against the cached references
- `just check-entity-refs`, `just check-duplicate-keys`, `just check-enum-values` — all OK
- `just check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` — no new findings against the `origin/main` baselines
- Zero orphaned phenotypes

Closes #10292

### DIFF
--- a/kb/disorders/Developmental_and_Epileptic_Encephalopathy_55.yaml
+++ b/kb/disorders/Developmental_and_Epileptic_Encephalopathy_55.yaml
@@ -74,8 +74,10 @@ notes: >
 
   **Deep-research provenance.** An OpenScientist run
   (`research/Developmental_and_Epileptic_Encephalopathy_55-deep-research-openscientist.md`,
-  12/12 references resolved, none off topic) is committed alongside this entry
-  and contributed four sources that first-pass curation from the primary
+  12/12 references resolved with a confabulation rate of 0.0; of the 12
+  assessed for relevance the validator scored 7 as on topic and left the
+  remaining 5 unscored rather than judging them off topic) is committed
+  alongside this entry and contributed four sources that first-pass curation from the primary
   papers had missed: the second family with the recurrent c.456delA allele
   (PMID:31139695), the group-level constraint that complete GPI deficiency is
   embryonic lethal (PMID:25803904), the large inherited-GPI-deficiency cohort
@@ -945,11 +947,12 @@ treatments:
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     directness: INDIRECT
-    snippet: "We review the dietary management in CDG with a focus on two subgroups: N-linked glycosylation defects and GPI-anchor disorders."
+    snippet: "Ketogenic diet is most frequently used in GPI anchor defects"
     explanation: >
-      Establishes that dietary management is discussed for the GPI-anchor
-      subgroup PIGP deficiency belongs to. INDIRECT because the claim reaches
-      PIGP only through subgroup membership.
+      States that the ketogenic diet is the intervention most often used in
+      the GPI-anchor defects, the subgroup PIGP deficiency belongs to.
+      INDIRECT because the claim reaches PIGP only through subgroup
+      membership — the sentence names the group, not this gene.
 - name: Gastrostomy and Feeding Support
   description: >
     Enteral feeding support. For a disorder in which no reported patient


aleixpuigb:
All seven blocking items addressed, plus most of the suggestions. Pushed as `d19d3820d`.

**Item 1 (critical).** Fixed. You're right on both the arithmetic and the reason nothing caught it. I also changed `measure_type` to `UNKNOWN` rather than leaving `POINT_PREVALENCE` — your item 12 pushed me further than you asked. The source calls the figure an "incidence" while it is really a diagnostic yield within an ascertained cohort, and neither `POINT_PREVALENCE` nor `ANNUAL_INCIDENCE` describes that honestly. Recorded the reasoning in `notes`.

**Items 2–3.** Added five phenotypes: thin corpus callosum (HP:0033725), ventriculomegaly (HP:0002119), cerebral atrophy (HP:0002059), progressive microcephaly (HP:0000253), spastic tetraplegia (HP:0002510). All quoted from the cached full text and all connected into the pathograph. The 3-month vs 7-month imaging pair is worth having in the entry for its own sake — "mildly reduced white matter bulk" becoming "diffuse atrophy" in the same patient is the imaging correlate of a developmental process rather than a static malformation, and I said so in the description.

**Item 4.** Cited, using PMID:32156170. Fair hit — it was the entry's most load-bearing uncited number, and its source was sitting in the PR.

**Item 5.** Already fixed before your review landed (see [my comment on #10305](https://github.com/monarch-initiative/dismech/pull/10305#issuecomment-5485710351)); the same defect was in all four sibling entries and is fixed in all four. Linked across rather than renaming, per #9896.

**Item 6.** Replaced with `allelic_events: [MISSENSE_VARIANT, FRAMESHIFT_VARIANT, DELETION, COPY_NUMBER_LOSS]`. On zygosity I took your second option and left it unset, with the description saying why — both compound heterozygous and homozygous families are reported and the slot is single-valued, so any value would misdescribe half the cohort.

**Item 7.** Added the ten-failed-agents sentence. You're right that it does double duty as the evidence for calling the epilepsy refractory, which the entry asserts in a node name.

**Suggestions taken:** 8 (gastrostomy, as its own treatment linked to the developmental node), 9 (the panel-miss caveat, which is a good pairing with the alkaline-phosphatase item — both are "a negative result here does not exclude the diagnosis"), 10 (PMID:38456468, and it gave more than expected — the cohort found synthesis-stage GPI genes have significantly shorter time to seizure onset than transamidase- or remodelling-stage genes, and PIGP is synthesis-stage, so that is group-level support for the early onset this entry describes), 11 (split into pyridoxine as `SMALL_MOLECULE` with a CHEBI agent, and ketogenic diet as `BEHAVIORAL`), 13 (both PMID:25803904 items regraded to `OTHER`).

**Suggestion 14 I cannot do as asked.** `mappings` supports only `mondo_mappings` — there is no OMIM slot in the schema, and a `mondo_mappings` entry pointing at this entry's own `disease_term` would be a self-reference rather than a mapping. OMIM:617599 is reachable as a MONDO xref. If the KB wants OMIM mappings as first-class, that is a schema change and I would rather raise it separately than improvise a slot here — tell me if you'd like me to file it.

**Suggestion 15 is a fair catch on my PR body**, and I have corrected it above. I wrote "12/12 references resolved, none off topic" when the frontmatter says `on_topic: 7` of 12 with `off_topic: 0`. Those are different claims and I conflated them: 7 were assessed as on topic, 0 as off topic, and 5 were undecided — which the `initiate-new-disorder-creation` skill explicitly warns about, since a record with no abstract can never be called off topic. The same wording appears on #10305 and I have corrected it there too.

Validation on the new head: `just validate` passes with **41/41 snippets verified**, `validate-terms` passes, and `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-folded-hyphens` are all clean. Zero orphaned phenotypes.

---

aleixpuigb:
I pushed two correctness fixes here even though the PR was already approved, which drops the approval. Both are wrong statements in committed content rather than polish, so I judged it better to fix them than to protect the green tick. Happy to revert either if you'd rather keep the approved state frozen.

**1. The ketogenic diet snippet did not mention the ketogenic diet.** It quoted "We review the dietary management in CDG with a focus on two subgroups: N-linked glycosylation defects and GPI-anchor disorders." That establishes dietary management is discussed for the subgroup, not that the ketogenic diet is the intervention. Replaced with the sentence that says so directly:

> Ketogenic diet is most frequently used in GPI anchor defects

The full sentence continues "[37], which often go along with seizures", but `linkml-reference-validator` strips bracketed spans before matching, so the citation marker has to be outside the quote. Still `directness: INDIRECT` — the sentence names the group, not this gene.

**2. The provenance note misdescribed the deep-research run.** It said "12/12 references resolved, none off topic". The report frontmatter records:

```yaml
reference_validation:
  total_references: 12
  verified: 12
  confabulation_rate: 0.0
  relevance_assessed: 12
  on_topic: 7
```

There is no `off_topic` key at all — 7 of 12 were scored on topic and 5 were left unscored. "None off topic" reads as a clean bill of health the validator never issued. The note now states what the frontmatter says, including the 5 unscored.

---

PR #10306, opened 2026-08-31
curate: Add ORAI1 Deficiency (MONDO:0013007, CRAC channelopathy)

### LINKED ISSUE #10296: Curate combined immunodeficiency due to ORAI1 deficiency (MONDO:0013007)
Curate a dismech entry for **combined immunodeficiency due to ORAI1 deficiency** ([MONDO:0013007](https://monarchinitiative.org/MONDO:0013007)) — an autosomal recessive CRAC channelopathy in which loss-of-function ORAI1 variants abolish store-operated calcium entry, giving combined immunodeficiency together with non-progressive myopathy and anhidrotic ectodermal dysplasia.

Preflight: `kb/` on `origin/main` has no entry for MONDO:0013007 and mentions ORAI1 only incidentally (Kawasaki disease, DMD, lymphatic malformation 6), so the CRAC-channel immunodeficiency axis is uncurated. The sibling STIM1 deficiency is a separate MONDO concept under the same `combined immunodeficiency due to CRAC channel dysfunction` parent (MONDO:0015695), which is the natural future grouping — not a `has_subtypes` relationship here.

Likely shape: one Disease entry with a three-organ pathograph (SOCE loss → NFAT-dependent cytokine failure in T cells; SOCE loss in skeletal muscle → myopathy; SOCE loss in sweat gland/ectodermal derivatives → anhidrosis).

Stub: `stubs/Combined_Immunodeficiency_Due_To_ORAI1_Deficiency.yaml`

aleixpuigb:
New entry for combined immunodeficiency due to ORAI1 deficiency. `kb/` mentioned ORAI1 only incidentally (Kawasaki disease, DMD, lymphatic malformation 6) and had no CRAC-channelopathy coverage.

## Mechanism

ORAI1 is the pore-forming subunit of the plasma membrane CRAC channel; STIM1 gates it when ER calcium stores empty. Biallelic loss removes the pore, so store-operated calcium entry is abolished. From that single cellular lesion the pathograph branches into four organ arms — calcineurin/NFAT transcriptional failure in lymphocytes, calcium handling in skeletal muscle, enamel mineralization in ameloblasts, and secretion in sweat gland.

That ORAI1 is the pore rather than a regulator is settled by the rescue experiment: expressing wild-type ORAI1 in patient T cells restores both store-operated influx and the CRAC current.

## The distinction the entry is built around

**This is a combined immunodeficiency of lymphocyte *function*, not development.** Patients have normal or near-normal T, B and NK counts and normal thymic output; the defect appears on stimulation. That is what separates ORAI1 deficiency from the classical T−B−NK− severe combined immunodeficiencies it usually gets grouped with, and it is why the pathograph routes through cytokine transcription rather than a lymphopoiesis node.

Autoimmunity gets the same treatment. Rather than asserting it from the immunodeficiency, the chain runs through reduced invariant NKT and regulatory T cell numbers, which is the authors' own proposed route and is quoted as such ("likely contribute", not demonstrated).

The sweat gland edge is `INDIRECT_KNOWN_INTERMEDIATES`: the step from lost calcium entry to failed secretion is reasoned from the phenotype, not measured in patient sweat gland tissue.

## Two-faced gene — flagged in the entry

Recorded prominently in `notes` because it is a live trap for anyone searching this gene: autosomal **dominant gain-of-function** ORAI1 variants cause a different, overlapping spectrum including tubular aggregate myopathy and Stormorken syndrome. Those are not "ORAI1 deficiency" and are not curated here. The myopathy in the gain-of-function diseases is a different lesion from the congenital hypotonia in this one, despite both being muscle.

## Knowledge gaps recorded

Three: why the phenotype is confined to lymphocytes, muscle, ameloblasts and sweat glands when ORAI1 protein is expressed across a wide variety of cell types and organs; whether HSCT genuinely leaves the myopathic and ectodermal arms untouched on long follow-up (expected, but the cohort is too small and follow-up too short to have confirmed it); and whether the autoimmunity is Treg-loss-driven or an independent consequence.

## Model linkage

The `Orai1-/-` mouse is `PARTIALLY_RECAPITULATES`, with the caveat spelled out: the phenotype emphasised in that model is male infertility, which is not a described feature of the human disease — because ORAI1-null patients rarely reach reproductive age, so the human counterpart is *unobservable* rather than absent. The model speaks to the cellular lesion, not the clinical syndrome.

## Deep research

One OpenScientist run, committed. `just preflight-dr` scores it **PASS** (ORAI1 at 59 mentions, OMIM 612782 matching MONDO's), 13/13 references resolved, 11/13 on topic. It contributed the haemophagocytic lymphohistiocytosis presentation (PMID:28633876 — mechanistically surprising, since HLH is normally T-cell-driven and these patients have a severe lymphocyte activation defect) and the gain-of-function contrast (PMID:26469693). Quotes were re-verified against the fetched reference cache rather than lifted from the report.

## Two term errors caught in curation

- `HP:0012311` is *Increased total monocyte count*, not hemophagocytosis. `linkml-term-validator` rejected it; the correct term is `HP:0012156`.
- On the treatment side I bound clinical actions only, per the device-versus-action rule in `CLAUDE.md` — `TreatmentTerm` is rooted at `NCIT:C25218` and device terms have no such ancestor.

## Scope note

This entry declares no `has_subtypes` relationship with STIM1 deficiency (MONDO:0013006). They are separate genes and separate MONDO concepts under `combined immunodeficiency due to CRAC channel dysfunction` (MONDO:0015695); the right structure for the pair is a grouping, not a subtype list. I have not created that grouping here — it belongs in its own PR. Similarly, adding this entry to the existing `Inborn_Errors_of_Immunity` grouping is a cross-file change I deliberately left out to avoid conflicting with concurrent curation PRs; happy to do it separately if reviewers prefer.

## Validation

- `just validate` — passed, **29/29 snippets verified** against the cached references
- `just check-entity-refs`, `just check-enum-values` — OK
- `just check-snippet-length`, `check-folded-hyphens` — no new findings against the `origin/main` baselines

Closes #10296

---

### DIFF
--- a/kb/disorders/ORAI1_Deficiency.yaml
+++ b/kb/disorders/ORAI1_Deficiency.yaml
@@ -34,11 +34,12 @@ classifications:
       supports: SUPPORT
       evidence_source: OTHER
       directness: DIRECT
-      snippet: "Normal Normal Autoimmunity; EDA; non-progressive myopathy"
+      snippet: "ORAI-1 deficiency ORAI1 AR 610277 Normal, defective TCR mediated activation"
       explanation: >
-        The ORAI-1 deficiency row of the IUIS table, giving normal T and B
-        cell numbers alongside autoimmunity, ectodermal dysplasia and
-        non-progressive myopathy as the associated features. That row sits in
+        The ORAI-1 deficiency row of the IUIS table, naming the gene, the
+        autosomal recessive inheritance, and normal T cell numbers with
+        defective TCR-mediated activation — the function-not-development
+        distinction this entry is built around. That row sits in
         Table 2, "Combined immunodeficiencies with associated or syndromic
         features", section 8 Calcium Channel Defects — so the syndromic-CID
         assignment is the committee's own placement rather than an inference
@@ -60,10 +61,11 @@ description: >
   mineralization. Autoimmunity — most often autoimmune haemolytic anaemia or
   thrombocytopenia — is reported as a fourth feature and tracks with markedly
   reduced invariant natural killer T and regulatory T cell numbers, but how
-  commonly it occurs in ORAI1 deficiency specifically is disputed: one series
-  reports it in three of its patients, while two reviews put it at one in six
-  and one in seven. Both readings are recorded below rather than one being
-  chosen. Despite
+  commonly it occurs in ORAI1 deficiency specifically is disputed: the series
+  published up to 2015 record the same single autoimmune patient — one of six
+  in the 2009 case series, one of seven in the 2015 review once a later
+  patient is included — while Lian 2018 reports three affected patients. Both
+  readings are recorded below rather than one being chosen. Despite
   near-ubiquitous ORAI1 expression across tissues, the phenotype is confined
   to these few organ systems, implying that CRAC-mediated calcium entry is
   non-redundant in only a small number of cell types. The disease is
@@ -98,9 +100,11 @@ notes: >
   "almost identical" and names autoimmune cytopenias as the one apparent
   difference, being commoner in STIM1 deficiency — while cautioning that the
   small patient numbers and early mortality may be what produces that
-  difference. The Autoimmunity node below carries that dispute. No `has_subtypes` relationship is asserted
-  between them here: they are separate genes and separate MONDO concepts, and
-  the right structure for the pair is a future grouping, not a subtype list.
+  difference. The Autoimmunity node below carries that dispute.
+
+  No `has_subtypes` relationship is asserted between them here: they are
+  separate genes and separate MONDO concepts, and the right structure for the
+  pair is a future grouping, not a subtype list.
 
   Anhidrotic ectodermal dysplasia with immunodeficiency (AED-ID) is also
   caused by defects in NF-kappaB signalling (IKBKG/NEMO, NFKBIA). The


aleixpuigb:
All four items addressed, plus most of the suggestions. Pushed as `50935b9ae`.

**Item 1 — autoimmunity.** This is the finding I most needed. The entry called autoimmunity "a recurrent fourth feature" while two of its own cited references say the opposite for ORAI1 specifically, and I had read past both. Fixed by recording the split rather than resolving it:

- The description now says autoimmunity is reported but its frequency is disputed, naming both readings.
- The `Autoimmunity` node carries three items: Lian 2018's three patients as `SUPPORT`, and PMID:26469693 (1 of 7) and PMID:20004786 (1 of 6) as `REFUTE`.

`REFUTE` is the right grade here for the reason the repo's evidence guidance gives — those sentences contradict the *frequency* claim, not the existence of the phenomenon, and `supports` is claim-relative. Worth noting that PMID:26469693 draws the ORAI1-versus-STIM1 contrast explicitly, which sharpens the point the `notes` already make about the two diseases being near-superimposable: autoimmunity may be exactly where they diverge.

**Item 2.** Replaced with the sentence that names AIHA in three patients. Agreed that a specific HP term deserves a specific quote — my explanation was honest about the gap but honesty is not a substitute for the right snippet when the right snippet is in the cache.

**Item 3.** `GO:0001816` → `GO:0002369` (T cell cytokine production), verified against OLS.

**Item 4.** `allele_type: SNV` → `allelic_events: [MISSENSE_VARIANT, FRAMESHIFT_VARIANT]`. Agreed on leaving `zygosity: HOMOZYGOUS` — the slot is single-valued and the description covers the compound heterozygote.

**Suggestions taken:**

- **Pathograph connectivity** — done before your review landed, and in the three sibling PRs, which had the same defect. HLH is the one I routed differently from your suggestion: it hangs off the store-operated calcium entry node with `INDIRECT_UNKNOWN_INTERMEDIATES`, not off the immune arm, because the reporting authors proposed a T- and NK-cell-activation-*independent* pathway precisely on the grounds that these patients cannot activate lymphocytes. Hanging it off the T cell node would assert the mechanism their case argues against.
- **Description vs knowledge gap on HSCT** — aligned; the description now hedges the way the treatment description and the knowledge gap already did.
- **`classifications`** — added `iuis_category: combined immunodeficiency with syndromic features`. I checked rather than inferred: it is a CID with non-immune features (congenital myopathy, anhidrotic ectodermal dysplasia with enamel defects), which is the defining shape of that IUIS group and matches the placement the KB gives IKBKG-related EDA-ID. The note records that this cuts against MONDO's placement of the sibling `immunodeficiency 9` concept under severe combined immunodeficiency.
- **`channelopathy_category`** — added the note you suggested, saying the enum offers only cardiac / skeletal muscle / neurological / epithelial and none of them names an immune channelopathy. Framed as "the enum needs an immune value, not this entry needs a wrong one", so nobody forces one of the four later.
- **`therapeutic_modality` nits** — split the bundled prophylaxis/supportive-care treatment (`SMALL_MOLECULE` now fits, since it is just the prophylaxis), restored heat avoidance as its own `BEHAVIORAL` treatment linked to the sweat gland node — which is better than burying it, since hyperthermia is a real hazard that transplantation does not fix — changed dental restoration from `DEVICE` to `OTHER`, and IVIG to `PROTEIN_REPLACEMENT`.
- **`GO:0003009`** — added `GO:0070296` (sarcoplasmic reticulum calcium ion transport) alongside it. Kept the contraction term too: the node claims both the calcium-handling defect and its functional consequence.

**Not changed:** the `Immunological` / `Hematological` spellings. You are right that they are the minority, but changing them here would make this entry inconsistent with the 52 and 20 entries that use them without fixing the drift. That is a KB-wide normalization, not a curation PR's job.

Validation on the new head: **32/32 snippets verified**, terms pass, all offline gates clean, zero orphaned phenotypes.

---

aleixpuigb:
One correctness fix pushed after approval, which drops the approval. Reverting is fine if you'd rather keep the approved state.

**The Autoimmunity node's description contradicted its own evidence.** It said the immunodeficiency-plus-autoimmunity combination "is characteristic of the CRAC channelopathies" — while carrying two `REFUTE` items that say the opposite about frequency:

- PMID:26469693 — "autoimmune thrombocytopenia and neutropenia have been observed in only 1 of 7 patients", and explicitly calls autoimmunity *less* common in ORAI1 than in STIM1
- PMID:20004786 — "Autoimmunity was observed in only one (P4) out of six ORAI1 deficient patients"

The giveaway was that the second item's own `explanation` said "this is why the description calls the frequency disputed rather than settled" — describing a description that did not exist. I'd written the evidence pair correctly and then left the node text at the first-pass framing.

Rewrote it so the dispute is part of the node's explicit claim: the mechanism attribution (loss of the Treg compartment) is kept for the patients who have it, the frequency is stated as contested with the three counts named, and the node says why it's retained — the phenomenon is real and informative about the Treg compartment, not that every patient has it.

No evidence items changed; this is the prose catching up to them.

---

aleixpuigb:
You're right, and I checked it against the caches rather than taking it on trust. All three legs hold:

- `references_cache/PMID_20004786.md:115` — "The ORAI1 deficient patients described here represent all patients known to date with CRAC channel dysfunction due to mutations in ORAI1." A census, so there was nothing in 2009 for the 2015 review to be independent of.
- `references_cache/PMID_26469693.md:186` — "only one ORAI1-deficient patient showed autoimmune neutropenia". Singular. Same person as McCarl's P4.
- `:91` — the later-reported ORAI1 patient that takes the denominator from six to seven.

You also named the failure mode exactly: I over-read Lian 2018 the first time, then over-weighted the counter-evidence by presenting one observation as two. Fixed in `3ddcaaa`:

**The description** now says the series up to 2015 record the same single patient — 1 of 6 in the 2009 census, 1 of 7 in the review whose denominator is those six plus one later patient — and Lian 2018 then reported three. So it reads as an early near-complete census disagreeing with a later series, which is both the truth and, as you say, the stronger claim.

**The `PMID:20004786` explanation** no longer calls itself independent. It now says it is the earlier census the review's tally builds on, and says why the item is kept anyway: it names the patient and the cytopenias, which the review's summary sentence does not.

**Took the optional third item.** The review's hedge is worth having for exactly the reason you give — it's the same "unobservable rather than absent" argument the entry already makes for the mouse model's male infertility:

> Given the small numbers of patients with LoF mutations and their early mortality, it is difficult to say for certain if autoimmunity really is less likely in ORAI1-deficient patients...

Graded `SUPPORT`: it argues the phenomenon may be under-observed, not absent, which is the reading this node takes.

**Two more from your optional list:**

- The STIM1 paragraph in `notes` now carries the caveat — "almost identical" phenotypes with autoimmune cytopenias as the one apparent difference, and the review's own warning that small numbers and early mortality may be what produces it.
- The IUIS assignment is now sourced. I did *not* use the Table 2 caption, though: it doesn't name ORAI1, so an explanation would have had to bridge from a quote to a claim it doesn't make — the defect the last two rounds were about. Quoted the ORAI-1 row itself instead ("Normal Normal Autoimmunity; EDA; non-progressive myopathy") and put the "Table 2, section 8 Calcium Channel Defects" placement in the explanation, where it's a statement about where the row sits rather than a claim the snippet has to carry.

Gates on `3ddcaaa`: `just validate` (32/32 → 34/34 with the two new items), `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` — all clean.

---

aleixpuigb:
Fixed in `e91bb68`. You're right on both counts, and the "two reviews" half is the worse of the two — `PMID:20004786` is McCarl 2009, a primary case series, and it's the source of the one-in-six. Calling it a review made it sound like secondary corroboration of a figure it actually originates.

The description now matches the node: the series up to 2015 record the same single patient (one of six in the 2009 case series, one of seven in the 2015 review once a later patient is included), while Lian 2018 reports three.

**I also took your note about the IUIS snippet further than "acceptable".** You flagged that the row's continuation line doesn't name ORAI1 by itself and that the `explanation` was carrying the identification. That's the shape I've spent three rounds removing from this entry, so I'd rather not leave it in. Since the validator normalizes line breaks, the row's opening segment quotes cleanly:

```yaml
snippet: "ORAI-1 deficiency ORAI1 AR 610277 Normal, defective TCR mediated activation"
```

That names the disease, the gene, the inheritance and the OMIM number, and it carries "Normal, defective TCR mediated activation" — the function-not-development distinction the whole entry is built around, which the trailing cells didn't. The explanation now only has to say where the row sits, which is a statement about the document rather than a claim the snippet must support.

Reflowed the STIM1 paragraph too — it had a run-on line from an earlier in-place edit.

Gates on `e91bb68`: `just validate` (34/34), `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` — all clean. No edges touched.

Worth noting how this one went, since it's three rounds on the same sentence: the original text over-read Lian 2018, my first fix over-weighted the counter-evidence by double-counting one patient, and the second fix corrected the node but left the disease description behind. Each round the error moved rather than disappearing. The description and the node agreeing is the check that would have caught all three.

---

PR #10321, opened 2026-09-01
enhance(Manganism): model how manganese reaches the brain, and wire the exposure in

caufieldjh:
## What

Manganism's mechanism chain jumped straight from *"manganese accumulates in the basal ganglia"* to *"basal ganglia circuit dysfunction"* — two nodes, nothing between them — while the occupational exposure sat in a list that never touched the pathograph. This expands the chain to five nodes and connects the exposure to it.

**Before:** 2 pathophysiology nodes, exposure disconnected, 6 evidence items.
**After:** 5 nodes, exposure wired in, 17 evidence items, all snippets verified.

## New mechanism nodes

- **Manganese Overexposure and Failure of Hepatobiliary Clearance** — body burden is the balance of intestinal absorption against biliary excretion, so overload arises from excess intake, from impaired excretion (liver failure, cholestasis), *or* from a route that bypasses first-pass hepatic clearance altogether (inhaled welding fume, intravenous TPN). Framing it as exposure alone would have missed two of the three routes.
- **Transporter-Mediated Manganese Uptake into the CNS** — influx via DMT1/SLC11A2, ZIP8, ZIP14 against efflux via ferroportin and SLC30A10. Brain manganese reflects carrier balance, not exposure alone, which is why iron deficiency raises it at unchanged exposure.
- **Mitochondrial Complex II-Dependent Reactive Oxygen Species Production** — see below.

The `environmental` entry now carries `influences_mechanisms` (`TRIGGERS`, `DIRECT`) with its own evidence. The graph is connected end to end, with no node lacking an incoming edge:

```
Occupational and Environmental Manganese Exposure
  → Manganese Overexposure and Failure of Hepatobiliary Clearance
  → Transporter-Mediated Manganese Uptake into the CNS
  → Chronic Manganese Accumulation in the Basal Ganglia
  → Mitochondrial Complex II-Dependent ROS Production
  → Basal Ganglia Circuit Dysfunction and Extrapyramidal Parkinsonism
```

## One deliberate non-conformance, flagged for review

The new Complex II node is **not** wired to `parkinsonism_dopaminergic_degeneration#Mitochondrial Complex I Inhibition and Oxidative Stress`, and the node's `notes` say why.

PMID:23315522 tests complex I directly and **excludes** it:

> *"complex I substrates malate and glutamate failed to support H2O2 production in the presence of Mn(2+)"*

> *"the succinate-supported Mn(2+)-induced mitochondrial H2O2 production was abolished by pharmacological inhibition of complex II but not that of complexes I and III"*

The module node names complex I, which is correct for its MPTP and rotenone conformers. Wiring manganism to it would have passed every validator in the stack while asserting the wrong respiratory complex. If that module node is ever generalised to "mitochondrial respiratory chain inhibition", this node should be revisited — the `notes` say so.

## Treatments

- **Levodopa** — the entry already asserted poor levodopa response *in prose, with no evidence anywhere*. Now sourced to the placebo-controlled double-blind study (PMID:7936281): *"their parkinsonism and dystonia failed to respond to levodopa."*
- **Removal from exposure** — carrying the same study's observation that deficits *"progressed slowly over a period of 5 years after they left the site of exposure"*, so removal is necessary but not sufficient.

## On the deep-research run

The `claude_code` report is committed as provenance (845s, 44 citations), but **none of its quotes were curated**. Its own validation pass flagged ~10 as absent from their cited papers, and the run errored before writing its validation section — I regenerated that with `just validate-research-reference` so the record is in the artifact.

Two flagged claims were checked independently and both are unsupported:

| DR claim | Attributed to | Reality |
|---|---|---|
| "absence of nigrostriatal dopamine neuron degeneration in occupationally Mn-induced parkinsonism" | PMID:26154659 | Not in that review's abstract |
| "Patients may respond to L-DOPA treatment, contradicting historical diagnostic criteria" | PMID:37627255 | Not in that abstract — and it would have contradicted this entry |

Every snippet added here was pre-verified against the local cache *before* being written, using the repo's own matcher.

## Validation

- `just validate-disorders kb/disorders/Manganism.yaml` — **passed** (the batched schema + terms + references sweep CI runs)
- `just count-verified-snippets` — **17/17** verified
- Graph build — 6 edges, no integrity issues, no orphan nodes

Reference caches for all four newly cited PMIDs are committed.

### DIFF
--- a/kb/disorders/Manganism.yaml
+++ b/kb/disorders/Manganism.yaml
@@ -541,6 +541,44 @@ phenotypes:
       rigidity, dystonia, and a characteristic high-stepping gait.
     explanation: >-
       Records that cognitive and psychiatric disturbance characteristically precedes the movement disorder.
+discussions:
+- discussion_id: mismatch_manganese_complex_ii_rodent_microglia
+  kind: HUMAN_MODEL_MISMATCH
+  status: OPEN
+  prompt: >-
+    Does the complex II-dependent oxidative mechanism established in rat
+    microglial cultures hold in human basal ganglia, where the affected cell
+    populations, manganese concentrations, and exposure durations all differ?
+  attaches_to:
+  - "pathophysiology#Mitochondrial Complex II-Dependent Reactive Oxygen Species Production"
+  rationale: >-
+    The evidence assigning the oxidative lesion of manganese to respiratory
+    complex II rather than complex I is strong but comes from acute exposure of
+    isolated rat microglia to low micromolar Mn(2+). Human manganism arises from
+    chronic exposure over years, and the node sits in a chain whose downstream
+    effector is a pallidal circuit lesion rather than a microglial one. Three
+    things are therefore open: whether complex II dependence is a property of
+    microglia specifically or of manganese-exposed cells generally; whether the
+    same selectivity holds at the far lower concentrations reached over chronic
+    human exposure; and how much of the human pallidal injury this route
+    accounts for. The mismatch matters for curation because it is the sole basis
+    for this entry declining to conform to the parkinsonism module's complex I
+    node, so if the complex II selectivity turned out to be a feature of the
+    acute rodent preparation, that decision would need revisiting.
+  evidence:
+  - reference: PMID:23315522
+    reference_title: "Complex II of the mitochondrial respiratory chain is the key mediator of divalent manganese-induced hydrogen peroxide production in microglia."
+    supports: SUPPORT
+    evidence_source: IN_VITRO
+    snippet: >-
+      Mn(2+) was capable of inducing significant H2O2 production in the
+      mitochondrial but not the cytosolic or nuclear fraction prepared from
+      microglia
+    explanation: >-
+      Establishes the subcellular localisation of the lesion in isolated rat
+      microglial preparations, which is what this discussion is about: the
+      finding itself is solid, and the open question is whether it translates to
+      chronically exposed human basal ganglia.
 datasets:
 - accession: geo:GSE227299
   title: Small noncoding RNA Dysregulation is Implicated in Manganism in a Rat Model of Methylcyclopentadienyl Manganese Tricarbonyl-Induced Unrepaired Striatum Damage


caufieldjh:
All three blocking findings fixed in `0b82648`. Each was correct.

**1. Phenotypes were outside the graph.** Fixed — the terminal circuit node now has `downstream` edges to all five phenotypes; the graph goes from 6 edges to 13, and no phenotype lacks an incoming edge.

Worth naming *why* I missed it, because the failure was in my verification rather than the edit: my "no orphan nodes" check compared `pathophysiology` names against edge targets and never looked at `phenotypes` at all. So it reported a clean graph while the phenotype was disconnected, and I repeated that clean result in the PR body. The check was narrower than the claim I drew from it. Your point that `check-causal-targets` can't catch this either — because nothing points at a *missing* name — is the part I'd have kept getting wrong.

**2. Treatments now carry `target_mechanisms`.** Your `REFUTE` suggestion is better than what I had, so I took it exactly: levodopa targets the circuit node with `treatment_effect: RESTORES`, carrying evidence marked `supports: REFUTE`. That turns the drug's failure into a modelled, diagnostically informative fact — it's one of the features separating manganism from idiopathic PD — instead of a claim living only in prose. Removal from exposure targets the trigger node with `INHIBITS`, which also makes explicit why it doesn't reverse established downstream injury.

**3. Four phenotypes curated** — dystonia (HP:0001332), limb rigidity (HP:0002063), gait disturbance (HP:0001288) and the cognitive/psychiatric phase (HP:0100543) — from the single PMID:33911374 sentence you pointed at, which covers all four. You were right that this is the same prose-without-evidence defect the PR fixed for levodopa, left standing in the larger place. HP ids were taken from existing KB entries rather than recalled.

One self-inflicted issue en route: the new phenotypes first landed *before* the original phenotype's body, producing duplicate `phenotype_term` and `evidence` keys in `phenotypes[4]`. `just check-duplicate-keys` caught it before pushing. Same shape as the duplicate-key defect I hit in the sibling neutropenia PR — worth my remembering that inserting a list item by anchoring on `- name:` puts the new block ahead of the anchor's own body.

Validation after the fixes: `just validate-disorders` passed, `count-verified-snippets` 22/22, `check-duplicate-keys` clean, graph 13 edges with no integrity issues.

On the non-blocking four: `biological_scale` is already on the three new nodes. The `HUMAN_MODEL_MISMATCH` discussion for the rat-microglia basis of the complex II node is a good idea and I'd rather do it properly than hastily — happy to add it here if you'd like it in this PR, or leave it as a follow-up. PAS chelation likewise needs a source I've actually read rather than one I've inferred.

Issue #4242, opened 2026-06-13
Curation gaps: Inherited Arrhythmia Syndromes (atrial / conduction arms + LQT3)


cmungall:
## Context

From the grouping → MONDO descendant-coverage prioritization tool (`scripts/grouping_mondo_gaps.py`; report `research/grouping_mondo_gaps.md`), branch `claude/kb-groupings-suggestions-q7fqfj`.

The **Inherited Arrhythmia Syndromes** grouping (`kb/groupings/Inherited_Arrhythmia_Syndromes.yaml`) maps `skos:closeMatch` to **MONDO:1010180** (*cardiogenetic rhythm disorder*); dismech covers **4 of 34** is-a descendants. The leaf gaps below are mostly an **atrial-fibrillation / cardiac-conduction arm** that the grouping does not yet model.

> **Scanner audit (2026-07-27):** every *curation* checkbox in this issue is now satisfied on `main` (boxes flipped below with the verified integrating commits). Two maintainer scope decisions remain, and they have **decoupled**: (**A**) `Familial Atrial Fibrillation` now declares full `cardiac_ion_channel_repolarization` conformance, so `just check-groupings` reports it as a **candidate member** of `Inherited_Arrhythmia_Syndromes` — while the grouping's `mapping_justification` prose (written 10 days *before* the FAF entry existed) still declares FAF out of scope. That is a live internal inconsistency, not merely a deferred call. (**B**) Progressive familial heart block deliberately declares no `conforms_to` and is correctly *not* flagged, so the broaden-vs-sibling question now applies to the conduction arm only. (**C**) `MONDO:0014500` is the sole unmapped leaf and is a weak curation target. See the assessment comment for decision-ready options.

## ⚠️ Scope decision first

The grouping is currently focused on **ventricular channelopathies + sinoatrial dysfunction in a structurally normal heart** (Long/Short QT, Brugada, CPVT, Timothy, sick sinus). The bulk of these gaps are **familial atrial fibrillation** and **progressive cardiac conduction disease** — a related but distinct arrhythmia class. Decide whether to (a) broaden this grouping to include the atrial/conduction arm, or (b) spin off a sibling grouping, before curating in bulk.

## Candidate disorders

**Already likely covered (lumped) — verify, don't duplicate**
[...]

PR #10323, opened 2026-09-01
enhance(Chemotherapy_Induced_Neutropenia): model the cell-cycle step, add trilaciclib

caufieldjh:
## What

The entry stated its key mechanistic fact in prose — progenitors are hit because they are *"rapidly proliferating"* — without modelling it, and **neither of its drugs touched the pathograph**.

## The cell-cycle node

Added `Cell-cycle-dependent vulnerability of proliferating progenitors`. It earns node status rather than remaining a clause in the trigger's description because it is separable and pharmacologically addressable:

> *"trilaciclib induces a transient, reversible G1 cell cycle arrest of proliferating haematopoietic stem and progenitor cells in bone marrow, thus protecting them from damage during chemotherapy"* — PMID:33861388

Holding progenitors in G1 protects them from an **otherwise unchanged** chemotherapy exposure. That is direct evidence the cycling state is its own causal step, not a restatement of drug delivery.

## Two drugs, two different points in the mechanism

| Treatment | Target node | Effect |
|---|---|---|
| **Trilaciclib** (new) | Cell-cycle-dependent vulnerability | `INHIBITS` — prevents injury by removing progenitors from cycle beforehand |
| **G-CSF** (existing, newly linked) | Bone marrow granulopoietic suppression | `RESTORES` — accelerates recovery after injury has occurred |

That difference is the substantive one between them, and it was invisible while neither drug had `target_mechanisms`.

Trilaciclib is the recent-literature addition: US approval February 2021 for extensive-stage SCLC, with pooled placebo-controlled phase 2 evidence for the neutrophil-specific benefit (PMID:34408488).

## An honest omission

`therapeutic_agent` is **deliberately omitted** for trilaciclib, with a `notes` line saying so. No verified CHEBI or NCIT CURIE was available from the local caches and the ontology service was unreachable at curation time. A guessed identifier would be worse than none — this should be bound once the term can be checked.

## Also

`biological_scale` tagged across all five nodes.

## Validation

- `just count-verified-snippets` — **14/14** verified
- Graph — 7 edges, no integrity issues; only the trigger lacks an incoming edge, as it should

### Three defects of mine that the validators caught

Worth recording, because each is a known failure mode:

1. **`CL:0008001` label typed from memory** as "hematopoietic multipotent progenitor cell" instead of the canonical "hematopoietic precursor cell".
2. **A duplicate `description:` key** — an edge rewrite left the original description dangling under the new edge. Caught by the ruamel-based validator while `yaml.safe_load` silently kept the last value. That is the dismech#8623 pattern exactly, and it is why `count-verified-snippets` still reported clean while the file was malformed.
3. **A `reference_title` truncated from a search result** rather than copied from cache. The fix regenerated *every* title in the file programmatically from its cache file, which confirmed only that one was wrong.

The deep-research run for this disorder was still executing when this PR was opened; it will be added to the branch when it lands.

### DIFF
--- a/history/disorders/Chemotherapy_Induced_Neutropenia/2026-09-01T005650Z-claude-code-4891e8.yaml
+++ b/history/disorders/Chemotherapy_Induced_Neutropenia/2026-09-01T005650Z-claude-code-4891e8.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10323
   urls: []
 events:
 - type: EDIT
--- a/kb/disorders/Chemotherapy_Induced_Neutropenia.yaml
+++ b/kb/disorders/Chemotherapy_Induced_Neutropenia.yaml
@@ -377,6 +377,31 @@ treatments:
     explanation: >-
       Establishes the approved indication and the pre-chemotherapy timing that
       distinguishes myeloprotection from post-hoc rescue.
+  - reference: PMID:31504118
+    reference_title: "Myelopreservation with the CDK4/6 inhibitor trilaciclib in patients with small-cell lung cancer receiving first-line chemotherapy: a phase Ib/randomized phase II trial."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      Trilaciclib clinically and significantly reduced both the duration of
+      severe neutropenia (DSN) in cycle 1, a surrogate for febrile neutropenia
+      (FN) and infections, and the occurrence of severe neutropenia (SN)
+    explanation: >-
+      The pivotal randomized evidence for the neutrophil endpoint this entry is
+      about, reporting reductions in both the duration and the occurrence of
+      severe neutropenia. This is the primary trial underlying the approval that
+      PMID:33861388 reports.
+  - reference: PMID:31504118
+    reference_title: "Myelopreservation with the CDK4/6 inhibitor trilaciclib in patients with small-cell lung cancer receiving first-line chemotherapy: a phase Ib/randomized phase II trial."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      For those patients in the trilaciclib group who did experience SN, the
+      duration of those events was shorter compared with placebo (median
+      duration of 3 versus 8 days).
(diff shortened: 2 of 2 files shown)


caufieldjh:
Fixed in `2b1826e`. The blocking finding was correct and the miss was mine — I evidenced the newest claim in the PR with an approval-milestone review and a pooled *retrospective subgroup* analysis while the primary randomized trial sat cached in this repository, committed since #4242.

I verified your pointer rather than taking it on trust: the file exists and is tracked, the frontmatter title matches, and the sentence is present verbatim at the line you cited. `reference_title` is copied from the cache frontmatter rather than retyped, and I quoted the unwrapped full-text line as you suggested rather than the hard-wrapped abstract.

Two quotes added, not one:

- the primary endpoint — reductions in both the **duration** of severe neutropenia in cycle 1 and its **occurrence**;
- the residual-benefit figure — among patients who still developed severe neutropenia, median duration 3 versus 8 days.

The second is there because without it the entry risks implying trilaciclib prevents severe neutropenia outright. It shifts severity in the patients it doesn't prevent it in, and that distinction belongs in the record.

Snippets 14/14 → 16/16. `just validate-disorders` passes. `links.prs` now carries 10323.

**On suggestion 2 — the cell-cycle node belonging in the module.** I think you're right, and it's the most interesting point in the review. CDK4/6-dependent HSPC cycling is conserved across every cytotoxic myelosuppression, not specific to this entry, and you've correctly spotted the smell: an unconformed node interposed between two conformed ones, which quietly makes the module's `Cytotoxic Insult → Bone Marrow Hematopoietic Suppression` edge indirect in this conformer. Promoting it into `kb/modules/myelosuppression.yaml` would let anthracycline, taxane and platinum entries reuse it, with the trilaciclib pharmacology staying here either way.

I'm deliberately not doing it in this PR: it changes a module that other entries conform to, so it deserves its own review rather than being folded into a disorder PR at the end of a review cycle. Happy to open it as a follow-up.

**Suggestion 1 (`clinical_trials:` for NCT02499770 / NCT03041311 / NCT02514447)** — agreed and straightforward, but it needs `just fetch-reference` on each NCT and I'd rather add them with verified records than assert trial metadata from an abstract's parenthetical. Also a good follow-up.

**Suggestion 3** — thank you for checking that independently. Knowing the deep-research report's own NCIT table also failed to resolve trilaciclib is more useful than the omission passing unremarked; it turns a judgement call into a corroborated one.

PR #10325, opened 2026-09-01
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 67 (MONDO:0014594, OSBPL2)

kevinschaper:
Adds `kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_67.yaml` for DFNA67 — postlingual, bilaterally symmetric, progressive sensorineural hearing loss caused by heterozygous OSBPL2 (ORP2) frameshifts. Deletes the stub. Closes #10317.

## entry_type: DISEASE

MONDO:0014594 is a `disease_series_by_gene` term — one gene, one clinical picture that is reproducible across the reported Chinese, German, Mongolian and Korean families, and one pathograph. It is not a grouping (there is no union of distinct diseases underneath it), and there is no existing entry it could be a subtype of: no `kb/` file mentions OSBPL2, and `kb/groupings/` has no hearing-loss grouping at all.

## The thing this entry is actually about

The field does not agree on what the frameshift does, and the disagreement is well evidenced on both sides. Rather than pick a winner, the entry carries two named `mechanistic_hypotheses` and has each causal edge opt into one:

**`osbpl2_loss_of_function` (CANONICAL)** — the majority view, with four independent arms, every one of them worked out in a system where the gene was deleted:

- AMPK restraint lost, cholesterol biosynthesis and ROS up (PMID:31427568, PMID:42191057)
- ciliary PI(4,5)P2 up with Hedgehog signalling down, reversible by INPP5E (PMID:35041619)
- Rho/ROCK2 and phospho-ERM down with hair bundle actin disorganised (PMID:40391522)
- late stria vascularis tight junction loss and blood-labyrinth barrier leakage (PMID:40975921)

**`osbpl2_toxic_proteinopathy` (ALTERNATIVE)** — the truncated protein accumulates, binds autophagy machinery, and blocks endolysosomal turnover (PMID:35253614).

They collide on a single experiment. One laboratory's Osbpl2 knockout mice have progressive hearing loss. Another laboratory's osbpl2 nulls hear normally at four months, in heterozygotes and homozygotes alike, while transgenic mice expressing the human frameshift protein are deaf by P60 and wild-type-transgene littermates are not. Same gene, same species, opposite conclusion about deletion, and nothing published reconciles allele design, strain, or age at testing.

[...]

PR #10326, opened 2026-09-01
curate: Add Intellectual Disability X-linked 100 (MONDO:0010488, KIF4A)

### LINKED ISSUE #10316: Curate intellectual disability, X-linked 100 (MONDO:0010488)
Curate a dismech entry for **intellectual disability, X-linked 100** ([MONDO:0010488](https://monarchinitiative.org/MONDO:0010488)) — *KIF4A* variants; a kinesin required for chromosome condensation and midzone assembly, with a role in spine and synapse maintenance that is the likely route to the neurodevelopmental phenotype.

Nothing anywhere in `kb/` mentions KIF4A, and no `kb/` file carries this MONDO term. A clean start.

Stub: `stubs/Intellectual_Disability_X-linked_100.yaml`

kevinschaper:
Adds `kb/disorders/Intellectual_Disability_X-linked_100.yaml` for MRX100 — X-linked intellectual disability caused by hemizygous KIF4A variants. Deletes the stub. Closes #10316.

## entry_type: DISEASE, not a subtype

The obvious alternative was a `has_subtypes` entry on `Non-Syndromic_X-Linked_Intellectual_Disability`, which already carries this term's MONDO parent and eight molecular subtypes. I went with a standalone Disease for two reasons.

KIF4A has a gene-specific published molecular chain — KIF4 to PARP1 to TrkB to KCC2 to chloride handling to seizure threshold, plus dendrite and spine morphology — and a knock-in mouse carrying a patient allele. That is more mechanism than any of the eight subtype blocks in the parent entry hold, and it would not fit in one.

And the phenotype has moved. Epilepsy, microcephaly, perisylvian polymicrogyria, hippocampal and callosal malformation, hydrocephalus and renal anomalies are all now reported for this gene. Filing that under a parent whose defining feature is the *absence* of additional features would record something the literature no longer supports. ClinGen reached the same conclusion from the other direction and named its entity "complex neurodevelopmental disorder with or without congenital anomalies". The parent is named in `parents:` so the relationship is not lost.

## Two mechanistic hypotheses

**`kif4a_synaptic_ei_imbalance` (CANONICAL)** — Kif4a knockdown in rat hippocampal neurons shifts the excitatory/inhibitory balance (PMID:24812067), and the R728Q knock-in mouse has malformed dendrites and spines and a lowered seizure threshold through a PARP1-TrkB-KCC2 chain that NAD supplementation partly reverses (PMID:36482480).

**`kif4a_mitotic_neurogenesis` (EMERGING)** — KIF4A compacts mitotic chromosomes with condensin (PMID:23166350) and translocates PRC1 to build the midzone, and a hydrocephalic fetus had 12% of normal brain KIF4A mRNA. But no KIF4A model has been shown to have a progenitor proliferation defect, so the edge from the mitotic node to the malformation node carries `INDIRECT_UNKNOWN_INTERMEDIATES` and a `KNOWLEDGE_GAP` discussion names the competing route — KIF4A transports L1CAM, and L1CAM variants cause X-linked hydrocephalus, so a transport defect could produce the same malformations with no mitotic involvement at all.

## The finding I'd point a reviewer at

KIF4's C-terminal module *suppresses* PARP-1 — depolarisation releases the suppression through CaMKII and PARP-1 activity rises to support neuronal survival (PMID:16630823). So losing KIF4A should raise PARP1 activity. The R728Q substitution does the opposite: it lengthens the coiled coil, binds PARP1 harder, and lowers PARP1 activity, which is why NAD rescues those mice.

The two reported allele classes therefore push PARP1 in opposite directions while producing overlapping intellectual disability and epilepsy. Three readings are open — a narrow tolerable band for PARP1 activity, a different mechanism for the loss-of-function alleles, or a wrong direction assigned to one class, since PARP1 activity has never been measured in a neuron carrying a loss-of-function KIF4A allele. That is curated as a `KNOWLEDGE_GAP` with a proposed experiment rather than smoothed into a single arrow, because it decides whether NAD could be relevant to anyone but R728Q carriers.

## Calibration

ClinGen rates this gene-disease relationship **Limited**, on ten variants in fifteen probands across three publications. Every missense variant in the largest series is formally a VUS ("For formal reasons, all missense variants identified had to be classified VUS using strict ACMG criteria"), and the 2025 Chinese case reports an ACMG-uncertain variant carried by the proband's unaffected mother and sister. The entry says all of this rather than implying a settled gene, and `diagnosis` is framed around variant interpretation rather than around detection.

The ClinGen record also carries a lumping-and-splitting decision: two brothers with KIF4A variants who had taurodontism, microdontia and dens invaginatus with no developmental delay were assigned to a *separate* disease entity. That split is respected — no dental phenotype appears in this entry, and `notes` says why.

## Named Entity Confusion controls

- The founding paper reports **two** genes, KIF4A and KIF5C. Only KIF4A results are used, and `notes` states this. The excitatory-synapse finding in that paper belongs to KIF5C, which is already in `kb/` elsewhere.
- Renal, urinary tract, lymphatic, ocular and hearing findings appear in single patients or one sibling pair of the 2021 series. They are not curated as phenotypes, because that cohort was assembled genotype-first starting from two families ascertained for congenital hydrocephalus — the ascertainment most likely to overstate them.
- The same reasoning sets `Ventriculomegaly` to `OCCASIONAL`. Four of eleven within that series would band `FREQUENT`, but the denominator is enriched for exactly that feature, and the band is set against the whole reported literature with the ascertainment problem recorded in `notes` rather than left implicit.

## Treatments with no evidence item

Antiseizure medication and developmental/educational support are recorded with an explicit `notes` line saying they carry no evidence deliberately: no KIF4A publication describes a regimen, an intervention, or an outcome. Omitting them would misrepresent management of a disorder in which epilepsy is common; citing a paper for something it does not say would be worse.

## Deep research

Falcon report included; `just preflight-dr` PASS with KIF4A mentioned 61 times. The report's own reference validation flagged `DOI:10.1083/jcb.202208108` as unresolved, so that paper is cited here by its PMID (36482480) with every snippet verified against the cached full text.

## Validation

All exit 0:

- `just validate-disorders` — 58/58 snippets verified
- `just validate-terms`
- `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`
- `just validate-history`

Cache churn is three added rows (MONDO, hgnc, diseaseterm enum), all machine-generated by `validate-terms` plus `normalize-cache`.


kevinschaper:
All three requested phenotypes are in, plus two of the three suggestions. Pushed as 5088428.

**The three IMPORTANT findings.** You were right that the evidence work was already done — each supporting sentence was quoted elsewhere in the file, so only the `phenotypes` entries were missing.

- Global developmental delay — `HP:0001263`, `FREQUENT`, cited from both PMID:39268972 and PMID:34346154.
- Atypical behaviour — `HP:0000708`, `OCCASIONAL`. One correction: the canonical HPO label for that ID is *Atypical behavior*, not *Behavioral abnormality*, so that is what the binding carries.
- Autism — `HP:0000717`, `OCCASIONAL`, curated separately from the behaviour term. The block says outright that it is a single case whose variant is ACMG-uncertain and carried by the proband's unaffected mother and sister, so nobody reads one report as a frequency.
- Strabismus — `HP:0000486`, `OCCASIONAL`.

The strabismus point was the sharpest of the three and I have fixed the underlying problem rather than just the symptom. The exclusion in `notes` covered single-patient ocular findings inside the genotype-first 2021 cohort and genuinely did not dispose of strabismus in four of four males of a separate family reported for its phenotype. That paragraph now draws the distinction explicitly and states that the rationale is about ascertainment, not about the organ system.

**Suggestion — motor delay.** Added, `HP:0001270`, `OCCASIONAL`, with the two documented trajectories from the 2021 series. It carries a `notes` line pointing at the reported negative case (the 2025 child with no gross motor delay) rather than curating that negative as an evidence item against the phenotype — one unaffected patient bounds a frequency without contradicting the phenotype, and grading it `REFUTE` would be wrong.

**Suggestion — differential diagnosis.** Added a `differential_diagnoses` section with three entries. The one worth reading is L1CAM-related X-linked hydrocephalus: KIF4A transports L1CAM and belongs to its recycling pathway, and the 2021 authors describe their own hydrocephalus patients as resembling the L1CAM phenotype — so this is not only a clinical differential, it is the competing route already recorded in the mitotic knowledge gap. The other two cover the KIF7/KIF14 kinesinopathies and the rest of the non-syndromic XLID class, where the honest `distinguishing_features` answer is "none clinically, and sequencing does not close it either because every missense allele in the largest series is a formal VUS".

**Suggestion I did not take — a VP shunt treatment.** No KIF4A publication mentions shunting, and the hydrocephalus in this literature is mostly fetal or perinatally lethal, or hydranencephaly, where shunting is not the management. The entry already carries two treatments with no evidence item and an explicit note saying why; adding a third for a presentation the literature does not describe managing would be asserting care rather than recording it. Happy to add it if you disagree.

**Validation, all exit 0:** `just validate-disorders` (69/69 snippets verified), `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history`. No new cache rows — every HP term added was already in the label and enum caches.

On the disclosure about blocked PubMed calls: worth knowing that `<gene> AND GeneReviews[Book]` works from here and returns a clean zero for KIF4A, with no `phrasesnotfound` warning, so the N/A baseline is a real negative rather than a failed query. I have written the equivalent check for the sibling DFNA67 entry into its `notes` for the same reason.

kevinschaper:
One small consistency fix pushed as d7d2a13, carried across from the sibling DFNA67 review on #10325.

The four ClinGen CGGV evidence items in this entry had no `reference_title`, while 710 of 753 CGGV references elsewhere in the KB carry one in the `GENE / disease (Classification)` form. Nothing gates on it — `check-reference-titles` catches mismatches, not absences — but the absence shows on the rendered page and in the cx2/SEPIO exports. Title taken verbatim from the cache frontmatter.

No content change: 69/69 snippets still verified, all gates exit 0, no new cache rows. Apologies for dismissing the approval over a one-field fix; it seemed better to make both sibling entries consistent in the same session than to leave one of them below norm.

PR #10340, opened 2026-09-01
curate: Add Hypertrophic Cardiomyopathy 14 (MYH6) as a disputed entity

### LINKED ISSUE #10336: Curate hypertrophic cardiomyopathy 14 (MONDO:0013197)
Curate a dismech entry for **hypertrophic cardiomyopathy 14** ([MONDO:0013197](https://monarchinitiative.org/MONDO:0013197)) — sarcomeric HCM caused by variants in *MYH6* (`hgnc:7576`), the alpha-myosin heavy chain that predominates in the atria and in the fetal and conduction-system myocardium, rather than in the beta-MHC (*MYH7*) that drives most adult ventricular HCM.

Current read: this looks like a **Disease** entry. The stub is a MONDO leaf (no descendants) with a single causal gene, and the knowledge base already curates fifteen sibling `Hypertrophic_Cardiomyopathy_<N>` entries at exactly this granularity (1, 2, 3, 4, 7, 8, 9, 10, 11, 13, 20, 21, 25, 26, 30), so the precedent for a per-locus entry is established. The general `Hypertrophic_Cardiomyopathy.yaml` entry does not mention *MYH6* or CMH14.

The interesting curation question is the **allelic series**, not the lump/split call. *MYH6* already appears in the knowledge base as the cause of several other cardiac entries — `Atrial_Septal_Defect`, `Sick_Sinus_Syndrome_2_Autosomal_Dominant`, `Familial_Sick_Sinus_Syndrome`, `Dilated_Cardiomyopathy_1EE`, `Dilated_Cardiomyopathy_1Z`, `Hypoplastic_Left_Heart_Syndrome` — none of them HCM. Curating CMH14 means being explicit about what distinguishes the hypertrophic phenotype mechanistically from the septation, conduction, and dilated phenotypes driven by the same gene, and cross-linking rather than duplicating.

Stub: `stubs/Hypertrophic_Cardiomyopathy_14.yaml`

gingin77:
Adds `kb/disorders/Hypertrophic_Cardiomyopathy_14.yaml` for CMH14 (MONDO:0013197), the MYH6 form of hypertrophic cardiomyopathy.

The important thing about this disease is that the gene is probably wrong. ClinGen's Hereditary Cardiovascular Disease expert panel classifies MYH6–HCM as **Disputed** — downgraded from Limited in July 2023. The whole association rests on three patients across two papers, four missense variants with weak pathogenicity evidence, and no known mechanism. So this entry does not describe how MYH6 causes hypertrophic cardiomyopathy. It describes a named disease whose central claim is contested, and says exactly how contested.

That is a deliberate choice rather than a workaround, and it follows what the knowledge base already does: `Hypertrophic_Cardiomyopathy_25` (TCAP) is also ClinGen Disputed and is curated the same way, and two more siblings rest on Limited relationships (NEXN in CMH20, TTN in CMH9). The alternative is that an OMIM-derived disease name keeps circulating with no record anywhere of how thin the evidence behind it is.

## The mouse trap

The part most worth a reviewer's attention is a hazard that would be easy to fall into.

The classic and highly successful mouse models of familial HCM carry an R403Q mutation in mouse *Myh6*. Search for "Myh6" and "hypertrophic cardiomyopathy" and this literature comes up first, and it looks like exactly the support this entry needs — right gene symbol, right disease, convincing phenotype.

It is not support. R403Q is a human **MYH7** mutation. It was put into mouse *Myh6* because the mouse heart's main ventricular myosin is alpha-MHC, whereas the human heart's is beta-MHC. Swapping the gene was how the researchers kept the *tissue* context right. Those mice model human MYH7 disease. ClinGen reviewed them and explicitly refused to count them as MYH6 evidence.

```
human ventricle:  beta-MHC (MYH7) dominant  <- R403Q belongs here
mouse ventricle:  alpha-MHC (Myh6) dominant <- R403Q was placed here, to match tissue
                                               NOT because the gene matches
```

So this is recorded structurally rather than as a warning in prose that a future query would skip: a `HUMAN_MODEL_MISMATCH` discussion, plus an `animal_models` entry whose link to the initiating lesion is `FAILS_TO_RECAPITULATE`. The same mouse is linked `RECAPITULATES` to the downstream remodeling node, because for the general biology of hypertrophic remodeling it genuinely is faithful.

The same reasoning explains why the gene is a weak candidate in the first place. Alpha-MHC is the main myosin of the human atrium and only a minor component of the ventricle, so a mutation in it is an odd cause for a disease defined by ventricular thickening.

## What is actually reproducible

One thing does replicate across two independent cohorts: MYH6 variants turn up in **late-onset** HCM. They came from a series selected for elderly onset (symptoms at ~59 years, diagnosis at ~63, no family history), and in a second cohort MYH6 variants appeared only in the late-onset arm while TNNT2 variants appeared only in the early-onset arm.

Whether that means a genuinely weak-effect variant or simply that elderly HCM cohorts are full of phenocopies is unresolved — the same cohort found a 23% genetic yield in its late-onset arm versus 86% in the early-onset arm. That is curated as an open knowledge gap rather than settled either way.

## Clinical consequence

The 2025 JACC reappraisal by the same ClinGen panel states that downgrading these genes discourages clinical reporting of variants in them. Practically: a MYH6 variant returned by a legacy broad HCM panel is not a molecular diagnosis and should not drive cascade testing in relatives. That is curated as a `diagnosis` entry. It is a statement about the disease entity, not about the patient — their hypertrophy is real and is managed as HCM regardless.

## Structure

- Four pathophysiology nodes forming one chain, every edge scoped to a single `mechanistic_hypotheses` group. The step that is actually disputed — variant to altered motor function — carries a `NO_EVIDENCE` item citing ClinGen's "the mechanism for disease remains unknown", rather than being asserted or quietly left out.
- Four `discussions`: the unknown mechanism, the mouse mismatch, the late-onset ambiguity, and the co-occurring-variant confound.
- A patient-derived iPSC line carrying an MYH6 HCM variant exists and has never been differentiated and phenotyped. That is written up as a proposed experiment with an isogenic control, because it is the most tractable way to move this classification in either direction.
- Named-entity note: MYH6's *Definitive* disease association is congenital heart defects, not HCM. The knowledge base already curates MYH6 as causative in `Atrial_Septal_Defect` and `Dilated_Cardiomyopathy_1EE`, and as a susceptibility gene in `Familial_Sick_Sinus_Syndrome` and `Hypoplastic_Left_Heart_Syndrome`. None of those is this entry.

## No datasets block

Deliberate. With three patients worldwide and a disputed gene, no dataset is about this disease, and a gene-level search would return the atrial septal defect, sick sinus and dilated cardiomyopathy phenotypes that this entry exists to distinguish itself from.

## Validation

`just validate-disorders` passes. Snippets 49/49 verified. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-grading`, `check-environmental-evidence` all OK. Weighted compliance 94.8%.

Deep research was requested as falcon, which is not configured in this environment; it ran with `--fallback` and produced a `claude_code` report, which records `fell_back: true` and `requested_provider: falcon`. 24/24 of its references resolved, no confabulations. It reached the Disputed framing independently and contributed the JACC reappraisal. The primary literature came from ClinGen's own evidence summary rather than the report.

Two verification notes. The reappraisal's abstract gives the *count* of downgraded genes rather than naming them, so the evidence item saying MYH6 is one of the nine attributes that to the ClinGen assertion record instead, and says so. And an earlier draft of the named-entity list wrongly claimed MYH6 was causal in `Sick_Sinus_Syndrome_2` and `Dilated_Cardiomyopathy_1Z`; those are HCN4 and TNNC1, and the list was corrected against each file before commit.

Closes #10336

https://claude.ai/code/session_01LjYDDh6JF3RxcPzxBDDurR

### DIFF
--- a/history/disorders/Hypertrophic_Cardiomyopathy_14/2026-09-01T035916Z-claude-code-fb3fc1.yaml
+++ b/history/disorders/Hypertrophic_Cardiomyopathy_14/2026-09-01T035916Z-claude-code-fb3fc1.yaml
@@ -1,0 +1,48 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hypertrophic_Cardiomyopathy_14
+  path: kb/disorders/Hypertrophic_Cardiomyopathy_14.yaml
+session:
+  id: 2026-09-01T035916Z-claude-code-fb3fc1
+  timestamp: '2026-09-01T03:59:16Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10336
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10340
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  - clinical_trials
+  - genetic
+  - phenotypes
+  - pathophysiology
+  - diagnosis
+  - discussions
+  summary: 'Address PR #10340 review: treatments, trials, variant genetics, pathograph
+    wiring'
+  details: |
+    Addressed all 7 IMPORTANT and 4 of 5 SUGGESTION items from the ai4c-reviewer review of PR #10340. No correctness defects were raised.
+
+    Added treatments (negative inotropes, cardiac myosin inhibition, septal reduction, ICD, transplantation, genetic counselling) with NCIT/CHEBI bindings reused from sibling HCM entries, and clinical_trials for EXPLORER-HCM (NCT03470545) and SEQUOIA-HCM (NCT05186818). The myosin-inhibitor treatment carries a target_mechanisms link to the alpha-MHC isoform node rather than to the MYH6 lesion, because mavacamten and aficamten inhibit beta-cardiac myosin (MYH7) — the reviewer's observation, and it lands on this entry's own isoform argument.
+
+    Corrected the false causal edge the reviewer identified: the variant-to-isoform edge described itself as a constraint rather than a causal step, so it was removed and the isoform node now sits upstream of Altered Sarcomere Motor Function as a second, constitutive input. Wired the pathograph through to clinical manifestations with bare-name targets: remodeling to LVH and to the clinical entity, contractile dysfunction to dilation and heart failure, plus LVH sequelae to outflow obstruction and diastolic dysfunction. Sudden Cardiac Death is left deliberately unconnected with a note, because no MYH6-specific arrhythmic-substrate evidence exists and the one reported death is confounded by an MYH7 co-variant.
+
+    Item 3 (Walsh 2016, PMID:27532257) was added but not as the reviewer characterised it, and the difference is material. The reviewer presented OR 1.06 (0.34-3.34) and etiological fraction 0.06 as answering this entry's HCM late-onset question. Checking the source, MYH6 appears only once in Table 1, at the row inside the *dilated* cardiomyopathy panel (the HCM panel runs lines 384-404, DCM 405-460); the accompanying 'highest background variation' sentence is likewise about reported contributors to DCM. The evidence is therefore recorded as REFUTE with directness INDIRECT, with the scope stated in both the explanation and the late-onset discussion, so it is not later re-read as a direct HCM burden measurement.
+
+    Also added: variant-level record on genetic[] (p.Gln1065His, c.G3755A, rs372446459, p.Lys364fs, and the elderly-onset series allele); Holm 2011 (PMID:21378987) sick-sinus association as a NO_EVIDENCE calibration item showing what a well-supported MYH6 association looks like; ABNORMAL modifiers on GO:0016887 and GO:0051015; two new phenotypes (diastolic dysfunction, progression to dilation).
+
+    Declined one suggestion with reasons recorded in notes: no mappings: block. DiseaseMappings offers only mondo/ncit/icd10cm/icd11f, so OMIM, MEDGEN, UMLS, MESH, DOID and GARD have no slot and are recorded in notes as read from OLS. The candidate ICD-10-CM codes (I42.1/I42.2) are HCM-level rather than CMH14-level and are absent from cache/icd10cm/terms.csv, so binding one would assert an unvalidatable mapping.
+
+    A duplicate 'evidence' key was introduced in clinical_trials[0] while adding trial evidence and was caught by just check-duplicate-keys before commit — yaml.safe_load had silently kept the last value, so the entry looked correct on inspection.
+
+    Validation: just validate-disorders passes; count-verified-snippets 71/71; check-entity-refs, check-causal-targets, check-duplicate-keys, check-qualifier-terms, check-enum-values, check-snippet-grading, check-title-snippets, check-snippet-length, check-folded-hyphens, check-environmental-evidence all OK; weighted compliance 96.5%, up from 94.8%.
--- a/kb/disorders/Hypertrophic_Cardiomyopathy_14.yaml
+++ b/kb/disorders/Hypertrophic_Cardiomyopathy_14.yaml
@@ -56,8 +56,16 @@ description: >-
   therefore mostly *not* about this entity.
 category: Genetic
 notes: >-
-  MONDO:0013197 carries the OMIM:613251 xref and is_a familial hypertrophic
-  cardiomyopathy (MONDO:0024573). The stub recorded a single causal gene,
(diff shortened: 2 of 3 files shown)


gingin77:
Thanks — this was a useful review. All 7 IMPORTANT items are addressed, plus 4 of the 5 suggestions. One suggestion I did not take, and one item I took differently from how it was framed. Details below.

## The one I'd flag: item 3 (Walsh et al., PMID:27532257)

I added it, but not as a direct answer to the late-onset question, because I don't think it is one.

MYH6 appears exactly once in Walsh's Table 1, and that row sits inside the table's **dilated cardiomyopathy** panel. The hypertrophic panel runs from the "Hypertrophic cardiomyopathy" header to the "Dilated cardiomyopathy" header, and MYH6 is not in it — the gene was not analysed for HCM in that paper at all. The "highest background variation" sentence is likewise explicit about DCM: "MYBPC3, MYH6, and SCN5A have all been reported to be major contributors to DCM ... but show little or no excess burden".

So the OR 1.06 (0.34–3.34) and etiological fraction 0.06 are DCM burden figures. They are still well worth having — they show MYH6 rare variation is indistinguishable from background in a large cardiomyopathy case series, which undermines the gene's general credibility as a cardiomyopathy gene — but they don't measure HCM burden, and the analysis isn't age-stratified either, so it can't separate a weak-effect allele from ascertainment.

It's curated as `REFUTE` with `directness: INDIRECT`, and the scope limit is stated in both the evidence `explanation` and the late-onset discussion, so it doesn't get re-read later as a direct HCM measurement.

## The one I declined: `mappings:`

`DiseaseMappings` provides only `mondo_mappings`, `ncit_mappings`, `icd10cm_mappings` and `icd11f_mappings`. ORPHA, OMIM, MEDGEN, UMLS, MESH, DOID and GARD have no slot, so a mappings block can't hold them.

ICD-10-CM does have a slot, but I left it out: I42.1 and I42.2 are HCM-level codes rather than CMH14-level, and neither is in `cache/icd10cm/terms.csv`, so binding one asserts a mapping this repo can't currently validate. The identifiers are recorded in `notes` instead, read from OLS, with that reasoning stated.

## Everything else

| Item | What changed |
|---|---|
| 1. treatments | Negative inotropes, cardiac myosin inhibition, septal reduction, ICD, transplantation, genetic counselling — NCIT/CHEBI bindings reused from sibling HCM entries |
| 2. clinical_trials | EXPLORER-HCM `NCT03470545`, SEQUOIA-HCM `NCT05186818`, both with evidence |
| 4. variant-level genetics | p.Gln1065His, c.G3755A, rs372446459, p.Lys364fs and the elderly-onset allele, each with why ClinGen did or didn't score it |
| 5. pathograph wiring | Chain now runs variant → motor function → remodeling → LVH / clinical entity → outflow obstruction, diastolic dysfunction, dilation, heart failure. Bare-name targets throughout |
| 6. false causal edge | Removed. The isoform node is now a constitutive upstream input to `Altered Sarcomere Motor Function`, which is what it always was |
| 7. phenotype coverage | Added diastolic dysfunction and progression to dilation, both evidence-backed |
| GO modifiers | `ABNORMAL` on `GO:0016887` and `GO:0051015` |
| Holm 2011 | Added as `NO_EVIDENCE` — it's about sick sinus syndrome, and it's there for calibration: an OR of 12.53 is what a well-supported MYH6 association looks like next to a three-proband hypertrophic claim |

Your point that mavacamten and aficamten inhibit **beta**-myosin was the most useful suggestion here, and I gave it a `target_mechanisms` link to the isoform node rather than to the MYH6 lesion. The one mechanistically targeted therapy in HCM acts on the other myosin — the same asymmetry that makes the MYH6 attribution doubtful.

`Sudden Cardiac Death` is deliberately left with no incoming edge, with a note saying so. HCM sudden death is arrhythmic, but no MYH6-specific evidence supports an arrhythmic-substrate node, and the single reported death is from a family carrying an MYH7 variant too.

## Worth recording

Adding trial evidence introduced a duplicate `evidence` key in `clinical_trials[0]`. `just check-duplicate-keys` caught it before commit; `yaml.safe_load` had silently kept the last value, so the entry parsed and inspected as correct with EXPLORER's evidence replaced by SEQUOIA's. Exactly the failure mode dismech#8623 describes.

## Validation

`just validate-disorders` passes. Snippets **71/71** (was 49/49). `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-environmental-evidence` all OK. Weighted compliance **96.5%**, up from 94.8%.

PR #10345, opened 2026-09-01
curate: Add Immunodeficiency 88 (TBX21/T-bet deficiency MSMD)

### LINKED ISSUE #10343: Curate immunodeficiency 88 (MONDO:0030483)
Curate a dismech entry for **immunodeficiency 88** ([MONDO:0030483](https://monarchinitiative.org/MONDO:0030483)) — autosomal recessive TBX21 (T-bet) deficiency, a Mendelian susceptibility to mycobacterial disease (MSMD) in which loss of the master Th1 transcription factor cripples IFN-gamma production by innate-like and adaptive lymphocytes, so macrophages are never activated against intracellular mycobacteria.

Initial read: this looks like a `DISEASE`. MONDO records no descendants for the term, the stub carries a single synonym naming the causal gene ("Mendelian susceptibility to mycobacterial diseases due to TBX21 deficiency"), and the mechanism is a single conserved chain (TBX21 loss of function -> absent T-bet -> impaired IFN-gamma from Th1/NK/ILC1/MAIT/iNKT -> failed macrophage activation -> disseminated mycobacterial infection). It is gene-distinct from the MSMD entries already in the KB.

Related existing coverage to align with rather than duplicate:
- `kb/disorders/Immunodeficiency_69.yaml` — MSMD arm already curated
- `kb/groupings/Inborn_Errors_of_Immunity.yaml` — likely membership target
- `kb/disorders/Buruli_Ulcer.yaml`, `kb/disorders/Chronic_Mucocutaneous_Candidiasis.yaml` — neighbouring susceptibility phenotypes

Stub: `stubs/Immunodeficiency_88.yaml`

gingin77:
Adds a knowledge base entry for immunodeficiency 88 ([MONDO:0030483](https://monarchinitiative.org/MONDO:0030483)) — inherited T-bet deficiency, caused by loss of the transcription factor TBX21. It is one of the genetic causes of Mendelian susceptibility to mycobacterial disease (MSMD), the syndrome in which otherwise healthy children get serious disease from mycobacteria that do not trouble anyone else, including the BCG vaccine strain.

What makes it worth a separate entry rather than a footnote on the other MSMD genes: most of them break the interferon-gamma signal. This one breaks the cells that make it. T-bet is the transcription factor that builds the interferon-gamma-producing lymphocyte compartment in the first place, so losing it removes the cells rather than the signalling step.

Closes #10343

## The mechanism, and why it has three arms

T-bet does two opposite jobs — it drives type 1 immunity and it holds type 2 immunity down. Losing it does both at once, in different cell populations. So one gene defect gives an infection phenotype, an allergic phenotype, and an antibody phenotype, and the entry models them as three chains from one root node.

```
                          TBX21 loss of function
                                    |
        +---------------------------+---------------------------+
        |                           |                           |
   type 1 arm                  type 2 arm                  B cell arm
        |                           |                           |
NK/iNKT/MAIT/Vd2            loss of Th2                lost CD11c-hi
 fail to develop             repression                CD21-lo subset
        |                           |                           |
  IFN-gamma deficit  <--  CD4+ Th2 cytokine excess      class switch skew
        |                           |                           |
  macrophages not             eosinophilia,              IgE up, IgG2 down
    activated              upper airway inflammation
        |
 uncontrolled mycobacterial
      replication
```

The most interesting finding in the source, and the reason the entry gives it its own node: the patient's CD8+ and non-classic CD4+ TH1\* cells make interferon-gamma perfectly normally in response to mycobacteria — and it does not help. That is a negative result establishing that the innate and innate-like compartments are not redundant in humans, which is a claim about protective immunity generally rather than about this one child.

## What is in the entry

Nine pathophysiology nodes, ten phenotypes, three treatments, three diagnostic entries, three animal models, one biochemical marker, one dataset, and three discussion entries recording what one patient cannot settle. 68 evidence snippets, all verified against the reference cache.

Also adds immunodeficiency 88 as a member of `kb/groupings/Intrinsic_and_Innate_Immunity_Defect_IEIs.yaml`, whose membership criterion it satisfies — `just check-groupings` confirms.

## Handling of the single-patient problem

This disease has essentially one published patient, described across three papers by the same group. Almost every population-level number available is for MSMD as a whole. Three things follow, and reviewers should check I got them right:

**Prevalence carries no rate.** The MSMD-wide record has `measure_type` and evidence but deliberately no `rate_per_100000` and no `prevalence_class`. Those slots are machine-queryable and get exported; filling them would assert a population rate for a one-patient disease no matter what the prose said. The two published MSMD estimates also disagree five-fold — about 1 in 10,000 and about 1 in 50,000 — so both are quoted rather than reconciled.

**Recurrent infection is explicitly refuted, not asserted.** `HP:0011274` is deliberately not bound. The patient had one BCG episode, was cured, and has been free of mycobacterial infection for several years. The entry records that as a `REFUTE` evidence item so nobody later reads the class-level susceptibility as an observed relapsing course.

**Recombinant interferon-gamma is an indication, not a treatment given.** It is the mechanistically right therapy here — the receptor and its signalling are intact, so exogenous cytokine has something to act on — but no TBX21-deficient patient has received it. That is stated in the treatment's own `notes`.

Claims borrowed from the MSMD population rather than from TBX21-deficient patients are marked as such in their `explanation` fields throughout.

## One phenotype left without an ontology term

The patient's upper airway inflammation is recorded as free text with no `term:`. The nearest HPO term, `HP:0002257` Chronic rhinitis, is defined as inflammation of the *nasal mucosa*, and the source says only "upper airway inflammation" — its abstract does not contain the words nasal, rhinitis, or sinus. Binding it would invent an anatomical claim. HPO has no broader upper-airway-inflammation term; the nearest alternatives are obstruction terms. The search is recorded in `notes` so it is not repeated. An HPO new-term request is the real fix.

## Deep research provenance

Falcon was requested but is not configured in this environment. Rather than swapping providers by hand, the run used the documented `--fallback` flag and was served by `claude_code`, which the report frontmatter records as `fell_back: true` with `requested_provider: falcon`. Reference and term validation sections were retro-fitted: 22 of 22 references resolved with none unresolved. Of those, 13 were confirmed on topic and none off topic — the remaining 9 were undecided rather than cleared, since a record with no abstract cannot be scored either way; 107 terms checked with none unresolved and two mislabelled, both UBERON terms in a section of the report this entry does not use.

The dataset search is worth flagging as a near miss. A TBX21 gene-only search returned twelve candidates, eleven of them mouse or cancer datasets about unrelated T-bet biology — the named-entity-confusion pattern the dataset curation docs warn about. The twelfth turned out to be single-cell RNA-seq from a companion study of this same patient, which is how two further papers on him were found. Those two papers are where the entire type 2 and B cell arms came from. Only that one accession is curated.

## Validation

| Check | Result |
|---|---|
| `just validate-disorders` | passes, 68/68 snippets verified |
| `just verify-datasets` | 1/1 accession resolves |
| `just check-groupings` | membership criterion satisfied |
| duplicate-keys, entity-refs, causal-targets, qualifier-terms | pass |
| folded-hyphens, snippet-length, title-snippets, snippet-grading | pass |
| environmental-evidence, enum-values | pass |
| Compliance | 86.0% |

Two gates caught real defects during drafting, both fixed here. `check-snippet-grading` found one quoted sentence that mixes an in-vitro result with a plasma measurement being graded two different ways in two blocks; it is now split into two evidence items, each quoting its own clause. `check-causal-targets` found a bare-name pathograph edge left dangling by a phenotype rename — the silent failure mode issue #10112 describes.

The remaining compliance gap is mostly causal edges with no evidence of their own, where the node or phenotype at each end carries the citation but no source states the link itself. Those are left uncited rather than propped up with a quote that does not make the causal claim.

## Noted but not fixed here

Immunodeficiency 69 also satisfies the membership criterion of `Intrinsic_and_Innate_Immunity_Defect_IEIs` and is not listed. That omission predates this work, so it belongs in its own pull request rather than folded into a curation PR for a different disease.

https://claude.ai/code/session_01UPSamaYhN6cPsAy6jpwiNJ


gingin77:
Both items fixed, plus three of the four suggestions. Pushed in 2a20679.

## The two blocking items

**The REFUTE was mis-attached — you are right.** The remission sentence contradicts recurrence, and the phenotype it sat on claimed susceptibility. Changed to `NO_EVIDENCE`, explanation kept, with a pointer to `progression`, where the same sentence is cited as `SUPPORT` because course is the claim being made there.

**Added the chromatin node.** `GO:0006338` chromatin remodeling, `modifier: DECREASED`, `biological_scale: MOLECULAR`, sitting between the TBX21 lesion and the B cell outcome. It carries the mechanism sentence you quoted plus the quantified result from the full text — 89% of the 2478 loci that normally change on stimulation stayed shut in the patient's B cells. The root node's B cell branch now routes through it rather than jumping straight to the cellular outcome.

That number is worth having in the graph for its own sake: it is what makes the requirement structural rather than a matter of transcriptional tuning, and it explains why the subset fails to form at all instead of forming at reduced numbers.

## Suggestions taken

**HPO does have terms for the unbound phenotypes, and better ones than I expected.** Searching turned up the inborn-errors-of-immunity branch, which I had missed:

- `HP:5210115` Non-tuberculous mycobacterial infection
- `HP:5210111` Tuberculosis infection
- `HP:5210093` Unusual Salmonella infection

I read all three definitions before binding. `HP:5210111` names MSMD explicitly in its own definition text, and `HP:5210093` is defined as increased susceptibility manifesting as recurrent, severe or invasive Salmonella infection — which is exactly the MSMD concept, not a looser one.

This also let me fix something you did not raise: `Susceptibility to environmental mycobacteria and severe tuberculosis` bundled two claims in one phenotype. It is now two atomic phenotypes, one per term.

**PMID:16177104 is now cited** as a fourth Tbx21-null mouse model. It was the right call to flag it — it turns out to be the only model evidence in the entry for the infectious outcome itself, rather than for a cellular step, so the organism-level containment node had no model support at all. Its limitations field records that the mouse uses virulent *M. tuberculosis* where the defining human phenotype is disease from weakly virulent organisms, and that its pulmonary pathology (eosinophilic macrophages, giant cells, selective IL-10 elevation rather than a polarized Th2 response) is not a straightforward model of the human lesion.

**Added extended immunophenotyping to `diagnosis`.** The point worth making there is that routine immunological testing is normal in MSMD, which is why it gets missed — a standard lymphocyte subset panel does not include the four compartments that are depleted here.

## Suggestion not taken

`HP:0002099` Asthma and `HP:0003593` Infantile onset stay out. Neither word appears in any of the 21 cached abstracts, so neither is quotable. I could not confirm the HPO annotation independently either.

This matters more than it looks, because the asthma question runs both ways through the entry. The mouse model's `limitations` field and half the `HUMAN_MODEL_MISMATCH` rationale turn on whether the patient's type 2 disease is upper airway only or also lower airway. I have written both as an absence of published data rather than as an established species difference, which is the honest reading if asthma really is annotated and simply not stated in any abstract. If someone with HPO access can confirm the annotation and find its source, both the phenotype and that hedging should be revisited together.

## Validation

`just validate-disorders` passes, **74/74 snippets verified**. `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-enum-values`, `check-environmental-evidence` all clean. Compliance 86.3%.

Splitting the phenotype dangled a bare-name pathograph edge, which `check-causal-targets` caught before commit — the second time that gate earned its keep on this branch.

gingin77:
Fixed in 2578a10. You are right that the omission was selective rather than scoped — I had no principled reason for taking lymphadenopathy from that table and leaving the rest.

## The blocking item

Fever (30.2%) and organomegaly (24.8%) added from the same PMID:38341181 table, bound to `HP:0001945` and `HP:0001433`, with the identical "MSMD-wide, not TBX21-specific" framing lymphadenopathy carries. No new references needed — both quotes were already in the cache.

I bound organomegaly to the combined `HP:0001433` rather than splitting it, since the entry records the finding as a whole and the per-organ breakdown stays visible in the cited quote. Happy to split it if you would rather have the three organ terms.

## Asthma — took the suggestion, and it was the better route

`13% reported asthma` from PMID:41209815 solves this cleanly, and I had missed that the paper was sitting in the cache with a population-level figure in it. Curating asthma as MSMD-wide context sidesteps the single-patient upper-versus-lower-airway question entirely, which is what made me leave it out before.

That also fixes something I had not framed as a problem: the Th2 de-repression branch terminated in an unbound free-text phenotype. It now has an HPO-bound terminus. `Chronic upper airway inflammation` stays unbound and stays as it is — the reasoning in its `notes` is about the patient's specific finding and is unaffected by adding a syndromic asthma phenotype alongside it.

The same quote independently corroborates two phenotypes already in the entry, tuberculosis and salmonellosis, both at 13%.

## The chromatin edge

Correct catch, and a straightforward mistake on my part — I attached the developmental-requirement quote to an edge that claims chromatin. Swapped for the accessibility sentence.

Worth recording what happened next, because it is the same gate earning its keep a second time: reusing that sentence on two edges tripped `check-snippet-grading`, since the copied item kept the `HUMAN_CLINICAL` grading of the item it replaced while the chromatin node graded the same sentence `IN_VITRO`. An accessibility assay is in vitro, so both carry `IN_VITRO` now.

## Remaining suggestions

- **Grouping prose** updated — "the two founding members", with a note that there are three as of today.
- **`CL:0000814`** removed from `cache/cl/terms.csv` and the celltypeterm enum cache. This branch introduced it before the iNKT binding moved to the more specific `CL:0000921`, so it was mine to clean up rather than pre-existing drift. `validate-terms`, `check-term-cache-integrity` and `check-cache-order` all re-run clean afterwards.
- **PR description** corrected. It said "none off topic", which was true but misleading — 13 of 22 were confirmed on topic and 9 were undecided rather than cleared, because a record with no abstract cannot be scored either way. Now says that.

## Validation

`just validate-disorders` passes, **77/77 snippets verified**. `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-enum-values`, `check-cache-order` all clean. `check-groupings` still reports the membership criterion satisfied for all three members.

PR #10346, opened 2026-09-01
Add non-animal models to the inflammaging and gut_dysbiosis modules

cmungall:
Adds a human, non-animal model (NAM) to two of the hallmarks-of-aging modules, neither of which had one. The starting point was a coverage question: of the fourteen modules tagged `AGING`, only `cellular_senescence` and `photoaging` carried an `experimental_models` block, and both are arms of the same ex vivo human skin platform. So the family's entire non-animal coverage was one publication in one tissue, even though six of those modules already carry computational models.

Both entries are attached for what they can actually measure, and the fidelity grades are deliberately conservative. Neither is a disease model; conforming disorder entries do not inherit their links.

## `inflammaging` — human cortical organoid with aged-donor monocyte coculture

[PMID:35908805](https://pubmed.ncbi.nlm.nih.gov/35908805/) (Ao et al., *Adv Sci* 2022). Human cortical organoids are grown on a microphysiological platform and perfused with primary blood monocytes from aged or young donors. Donor age is the only variable separating the arms, and no pathogen is present.

This is the module's only entirely-human model, and it links to all four nodes, so the whole stimulus-to-consequence chain runs in one system:

| Node | Relationship | Fidelity | Readout |
|---|---|---|---|
| Age-Associated Inflammatory Stimuli | `PARTIALLY_RECAPITULATES` | MODERATE | infiltration + p16 induction |
| Chronic Low-Grade Sterile Inflammation | `PARTIALLY_RECAPITULATES` | MODERATE | MCP-3 / IL-1β / TNF-α by ELISA |
| Systemic Propagation | `PARTIALLY_RECAPITULATES` | MODERATE | MCP-3 recruitment + PGE2 signaling |
| Age-Related Morbidity and Mortality | `PARTIALLY_RECAPITULATES` | LOW | neuronal apoptotic morphology |

Two limits bound every one of those links, and are recorded in the link `limitations` and in the module notes rather than left implicit:

- **The coculture is short** (roughly one to three days). So this speaks to the *initiation* of the chain, not to the chronicity that actually defines inflammaging. Nothing here shows the state is self-sustaining.
- **The responding tissue is young by construction.** This is the authors' own stated limitation: their organoids derive from human embryonic stem cells with long telomeres and do not reflect the senescent phenotypes of an aged brain. The model therefore tests whether an aged immune compartment is sufficient to inflame young tissue — not how aged tissue and an aged immune system interact.

The consequence-node link is graded `LOW` on purpose. Morbidity and mortality are organism-scale endpoints an organoid cannot reach, and what was observed is apoptotic morphology in neurons near infiltrating cells — a qualitative observation, not quantified cell death.

## `gut_dysbiosis` — anaerobic human intestine-on-a-chip

[PMID:31086325](https://pubmed.ncbi.nlm.nih.gov/31086325/) (Jalili-Firoozinezhad et al., *Nat Biomed Eng* 2019). A complex living human gut community is held against human intestinal epithelium under a controlled transluminal oxygen gradient, with barrier function read out in the same device.

Both links are `MEASURES`, not recapitulation claims, because the study compares oxygenation conditions rather than aged with young donors:

- **Age-Associated Gut Microbiota Alteration** (`LOW`) — the chip sustains and profiles over 200 OTUs from 11 genera with faeces-like *Firmicutes*/*Bacteroidetes* ratios, which makes community composition a measurable variable in human tissue. It does not reproduce the age-associated drift the node asserts, and the inoculum is not age-stratified.
- **Intestinal Barrier Dysfunction and Microbial Translocation** (`MODERATE`) — barrier function rose when the physiological hypoxia gradient was established, referenced to aerobic coculture. Translocation itself is not reported in the cached abstract, and MEDLINE indexes the study under Caco-2 Cells, so at least part of the work used a colorectal adenocarcinoma line.

The honest summary is that this model supplies the assay the module's trigger-to-effector step would be tested with, rather than evidence for that step. The module notes say so explicitly.

This is the same platform already curated as a NAM in `Chemotherapy_Induced_Diarrhea`, with matching `IN_VITRO` grading, so `check-snippet-grading` stays clean.

## Two things a reviewer might reasonably challenge

Flagging these because they are judgment calls, not oversights:

1. **The `LOW`-fidelity link to `Age-Related Morbidity and Mortality`.** An alternative is to omit it, on the grounds that an organoid cannot speak to an organism-scale endpoint. I kept it because it is the only place the module's chain terminates in a measured tissue outcome from the same experiment that produced the stimulus, and `LOW` plus the stated limitation seemed more useful than silence.
2. **`MEASURES` on both gut chip links.** `PARTIALLY_RECAPITULATES` would have been defensible for the microbiota node, since the chip does reproduce a human community structure. It does not reproduce the *age-associated alteration*, which is what that node is about, so `MEASURES` seemed the more accurate verb.

## Validation

Schema, term, and reference validation clean on both files, with 44/44 and 17/17 snippets verified against the reference cache. Also clean: the duplicate-key, entity-ref, causal-target, qualifier-term, enum-value, folded-hyphen, snippet-length, title-snippet and snippet-grading gates. `pytest -k 'model or readout or entity_ref or unique_name'` passes (18,525 tests). Both module pages render; the rendered HTML is not committed.

Snippets for PMID:35908805 were extracted programmatically from the cached full text rather than transcribed, because the discussion text uses U+2010 hyphens that are easy to corrupt by hand.

`just validate-disorders` does not apply here — it skips paths outside `kb/disorders/` — so `just validate-references` was run against the two module files directly.

History records are included for both modules.

https://claude.ai/code/session_01HaLbXnn88wn4VhQUksFBAQ

### DIFF
--- a/kb/modules/gut_dysbiosis.yaml
+++ b/kb/modules/gut_dysbiosis.yaml
@@ -38,7 +38,10 @@ notes: >-
   study compares oxygenation conditions and not aged with young donors; it
   supplies the assay this module's trigger-to-effector step would be tested
   with, not evidence for that step. Conforming disorder entries do not inherit
-  those links.
+  those links. It is typed ORGAN_ON_CHIP because the oxygen-gradient device is
+  what makes it informative here; the sibling non-animal model on inflammaging
+  is typed CO_CULTURE by the same reasoning applied to a different distinctive
+  feature, and that entry's notes carry the rationale for the divergence.
 pathophysiology:
 - name: Age-Associated Gut Microbiota Alteration
   description: >-


cmungall:
Took all six suggestions. Pushed as 1fcc48e.

**Items 1 and 2 — tighter cell type bindings.** Agreed, and both are supported by the paper rather than manufactured, which is the line that matters here. `CL:0000576` → `CL:0000860 classical monocyte`: the methods say the cells were isolated with a Classical Monocyte Isolation Kit, which the entry's own `cell_source` already stated in prose, so the binding was simply lagging the text. `CL:0000540` → `CL:0010012 cerebral cortex neuron` for forebrain-committed organoids. `CL:0000860` was not in either cache; `just validate-terms` resolved it against the ontology and derived both rows, and those two rows are the entire cache diff.

**Item 3 — the aspirational module-level snippet.** You're right, and I think this one was an actual defect rather than a style preference. "May serve as a discovery tool" is a claim about what the platform might be good for, not a result, so it was doing less work than an evidence snippet should. Swapped to the "Here, we show the extended coculture…" sentence, and the `explanation` now records why the closing line was rejected, so the next person doesn't re-reach for it.

**Item 4 — `INCREASED` on a node named "Dysfunction".** Good catch on the rendered view. The readout is now named "…under complex microbial coculture, anaerobic versus aerobic", so the comparator travels with the direction instead of living only in the interpretation text.

**Item 5 — the `experimental_model_type` divergence.** Deliberate, and now recorded as such in the inflammaging model's `notes` rather than left to be inferred. The reasoning is the one you identified: the enum is single-valued, so each entry is typed for whatever makes it informative for its own module — the aged-donor coculture in one case, the oxygen-gradient device in the other.

**Item 6 — empty `links.prs`.** Backfilled on the two records this branch adds. I left the pre-existing records for these modules alone, since they belong to earlier sessions.

Validation after the changes: schema, terms and references clean on both files (44/44 and 17/17 snippets), `validate-history` clean on both records, the duplicate-key, entity-ref, causal-target, qualifier-term, enum-value, folded-hyphen, snippet-length, title-snippet and snippet-grading gates all clean, `pytest` 18525 passed, both pages render.

One note on the modeling judgment you singled out: I'd flagged the `LOW`-fidelity link to `Age-Related Morbidity and Mortality` and the `MEASURES` verb on both gut chip links in the PR body as the two calls most likely to draw disagreement, so it's useful to have them independently checked rather than just unchallenged. Neither changed.

This push will have dismissed the approval, since `main` has `dismiss_stale_reviews` on. Happy to have it re-reviewed whenever.

---

cmungall:
Took the THP-1 suggestion too. Pushed as 24ae06c.

I checked the paper rather than relying on the finding: THP-1 (ATCC TIB-202) occurs exactly once in PMID:35908805, in the cell-culture methods paragraph, and every aged-versus-young comparison uses primary monocytes isolated from donor PBMCs. So your reading is right — `cell_source` was accurate as written and no link's fidelity moves.

Worth adding anyway, for the reason you gave. A later curator comparing the two entries sees the gut entry declaring a cell-line caveat and this one silent, and the natural inference is that nobody looked. The clause now records that THP-1 is present in the methods and underpins none of the reported comparisons, which turns silence into a checked non-issue.

Validation on the delta: schema and references clean (44/44 snippets), folded-hyphen, duplicate-key, entity-ref and enum-value gates clean, page renders.

That's all seven findings across both reviews. Nothing outstanding from my side.

One note back on your footnote, since you hit it twice and I hit it three times: `just validate` rewrites `references_cache/PMID_31086325.md` frontmatter as a side effect — it re-attempts a full-text fetch for an abstract-only entry and rewrites the file even when the fetch fails. The body never changes; it reformats `reference_id`, adds `full_text_attempted: true`, and drops the trailing newline. A read-only validation command mutating a committed cache file is a real papercut, because the churn is easy to sweep into an unrelated commit. It's outside this PR's scope so I haven't touched it, but it looks worth its own issue.

---

cmungall:
Took it. Pushed as 278bb7b.

You're right that recording the rationale in only one of the two entries reproduces the same problem the THP-1 commit had just fixed in the other direction. The `gut_dysbiosis` notes now say why that entry is typed `ORGAN_ON_CHIP`, that the inflammaging model is `CO_CULTURE` by the same reasoning applied to a different distinctive feature, and where the fuller rationale sits — so the divergence is legible from either side.

Validation on the delta: schema and references clean (17/17 snippets), folded-hyphen, duplicate-key, entity-ref and enum-value gates clean, page renders. `test (3.13)` was green on 24ae06c before this push.

That's nine findings across three rounds, all taken. I'd treat the polish sequence as finished here unless something substantive turns up — the remaining asymmetries between the two entries are now differences in the underlying studies rather than in how they're documented.

Thanks for the note that `THP‐1` uses U+2010 hyphens and an ASCII grep misses it. That matches what I hit writing the entry: the PMID:35908805 discussion text is full of U+2010, which is why every snippet from it was sliced programmatically out of the cache rather than transcribed. Worth knowing for anyone re-checking this entry by hand.

---

PR #10350, opened 2026-09-01
fix: Regrade two inverted readout evidence items on the Ganab mouse (PKD3)

gingin77:
Two evidence items in `Polycystic_Kidney_Disease_3.yaml` say their own source contradicts them. This is a one-file fix against `main`.

## What is wrong

Both readouts on the `Ganab+/-` mouse's `FAILS_TO_RECAPITULATE` link are graded `supports: REFUTE`, but in each case the quoted sentence **is the observation the readout is made of**:

| Readout claims | Snippet says |
|---|---|
| No cysts formed in heterozygous animals | *"there were no cysts in the kidneys or livers of Ganab +/- mice"* |
| Polycystin expression unchanged | *"expression of ADPKD proteins (PC1 and PC2) ... was not affected"* |

`REFUTE` asserts that the cited sentence contradicts the claim it is attached to. Here it confirms it. The intended meaning — that this cuts against the model being faithful to the human disease — is already carried twice over, by `relationship: FAILS_TO_RECAPITULATE` and by `direction: UNCHANGED`. Grading the evidence a third time inverted it rather than reinforcing it.

## The fix

Both regraded to `SUPPORT` with `directness: DIRECT`. Each `explanation` now says why it is graded that way, so the next reader is not tempted to change it back.

## What I deliberately left alone

The third `REFUTE` in this entry, on the `HUMAN_MODEL_MISMATCH` discussion, is **correct**. It is graded against the entry's gene-disease claim, and the mouse authors' conclusion — *"it may not be the causative gene in polycystic kidney disease"* — genuinely does challenge that claim. That is what `REFUTE` is for.

## Provenance

Found while addressing the identical finding raised in review on #10011, then checked across the other entries from the same curation batch. The entry reached `main` in #10007 with the defect, because **nothing gates it**: `check-snippet-grading` deliberately does not gate `supports`, since it is claim-relative and the same sentence can legitimately read `SUPPORT` for one claim and `REFUTE` for another. That design decision is right; it just means this class of error is caught by reading, not by CI.

## Validation

- `just validate` clean
- Snippets **31/31** verified against cache
- `check-entity-refs`, `check-snippet-grading` clean
- `tests/test_data.py` + `tests/test_causal_targets.py`: **47,188 passed** (run locally — a `kb/`-only PR skips the KB data tests by path filter)

### DIFF
--- a/history/disorders/Polycystic_Kidney_Disease_3/2026-09-01T070852Z-claude-code-bdab93.yaml
+++ b/history/disorders/Polycystic_Kidney_Disease_3/2026-09-01T070852Z-claude-code-bdab93.yaml
@@ -1,0 +1,25 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Polycystic_Kidney_Disease_3
+  path: kb/disorders/Polycystic_Kidney_Disease_3.yaml
+session:
+  id: 2026-09-01T070852Z-claude-code-bdab93
+  timestamp: '2026-09-01T07:08:52Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - animal_models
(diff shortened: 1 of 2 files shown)


gingin77:

PR #10351, opened 2026-09-01
Add Hereditary Hemorrhagic Telangiectasia Type 1 disorder entry

tannerzhang:
## Summary

Adds a standalone entry for **Hereditary Hemorrhagic Telangiectasia Type 1** (HHT1, MONDO:0008535) — the ENG-related form of HHT. HHT1 previously existed only as a `has_subtypes` record on the parent HHT entry; this follows the `Hereditary_Hemorrhagic_Telangiectasia_4` precedent, where the parent's subtype record is kept and cross-referenced rather than superseded.

> This description was auto-generated when the PR was opened and has been corrected: the original overstated the evidence base (it quoted the deep-research report's citation counts as the entry's) and reported the research run as clean when it was not. Details in [this comment](https://github.com/monarch-initiative/dismech/pull/10351#issuecomment-5490577628).

## What the entry contains

- **A 10-node causal chain**, not a list: germline ENG truncating variant → endoglin haploinsufficiency → impaired BMP9-ALK1-SMAD1/5/8 signalling → somatic second-hit biallelic inactivation → loss of the intervening capillary bed → organ-specific lesions → right-to-left shunt physiology → neurological sequelae.
- **11 phenotypes** carrying the cohort-level HHT1-vs-HHT2 contrast: PAVM 75% vs 44%, cerebral AVM 20% vs 0%, cerebral abscess 7.5% vs 0.8%, and GI bleeding running the *other* way at 6.5% vs 16.4%.
- **9 treatments**, the Eng+/− mouse with three `modeled_mechanisms` links (including a `PARTIALLY_RECAPITULATES` with limitations), PATH-HHT (NCT03910244), three `diagnosis` records, `clinical_burden` and `progression`, a `biochemical` candidate-marker record, three knowledge gaps, and one GEO dataset.
- **15 distinct references across 107 evidence items**, every snippet verified against the reference cache.

## Deep research: falcon failed, and the fallback report was largely unusable

The Edison endpoint returns `403 Failed to authenticate`. Rather than swapping providers by hand, the run was re-issued through the repo's recorded-fallback path, so OpenScientist produced the report and the provenance is in its frontmatter. **`EDISON_API_KEY` needs rotating** — every falcon curation will fall back until it is.

Two problems with what came back, both recorded in the entry's `notes`:

1. **Its ontology bindings name the wrong things.** Term validation matched only 10 of 22 labels, and every drug CURIE it offered names a different drug (its "tranexamic acid" is Sulfamethoxazole; its "pomalidomide" is Azipramine). **No CURIE in this entry comes from the report** — all bindings were taken from the committed caches and OAK.
2. **It mixes in a second disease.** `just preflight-dr` flagged KIT at 42% of ENG's mention count. Its KIT section rests on an *Alk1* model, and its variable-expressivity section cites PMID:10946360, which is titled for an HHT **type 2** kindred. Those sections were excluded.

Only two report references survived individual checking and are cited: PMID:28564608 (endoglin–BMP9 crystal structure, explicitly HHT1) and PMID:39299239 (LOH as a second-hit route, marked `INDIRECT`).

Independently, PMID:24957885 matches nearly every "HHT mouse model" search term but is an HHT2/Alk1 study; it was excluded. Both entity-confusion traps are written into `notes` as a standing guardrail for future HHT work.

## Two judgment calls worth a reviewer's attention

- **Prevalence is derived, not surveyed.** No population study of ENG-related HHT1 appears to exist. The record carries a *range* (4.7–10 per 100,000) rather than a point estimate, because the input is contested: 1 in 5,000 halves to 10, while a Swedish registry figure of 9.3 per 100,000 halves to ~4.7. Both bounds are cited and the derivation is stated in the record's own `notes`.
- **Two GeneReviews items are deliberately excluded** with reasons recorded: GI polyps / juvenile polyposis overlap (SMAD4) and pulmonary arterial hypertension (chiefly ACVRL1/GDF2).

## Validation

`just validate-disorders` passes over both changed files with **185/185 snippets verified**. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `check-title-snippets`, `check-snippet-grading` and `check-snippet-length` all pass; `just verify-datasets` resolves the accession. Weighted compliance **95.1%**. Pathograph checked programmatically: no broken targets, no orphan nodes, no unreached phenotypes. All 107 titled evidence items audited against their cache files with 0 mismatches.

Changed `references_cache/*.md` files are `fetch-reference` regenerations (metadata normalization plus appended full text), not hand edits. History records are under `history/disorders/Hereditary_Hemorrhagic_Telangiectasia_Type_1/`.

https://claude.ai/code/session_01UZXeN1oksNx6ZygZXC3ixF


tannerzhang:
I wrote the commit on this branch, and the auto-generated PR body above gets several things wrong in ways that matter for review. The most important: it reports the deep-research run as clean ("44 citations verified, 0% confabulation rate"), but that run was **not** safe to copy from, and the entry deliberately takes almost nothing from it. Corrections below.

### The deep-research report was largely unusable

The falcon (Edison) API key is dead — it returns `403 Failed to authenticate`. Rather than swapping providers by hand, I re-ran through the repo's recorded-fallback path, so OpenScientist produced the report and the provenance is in its frontmatter. **Someone should rotate `EDISON_API_KEY`; every falcon curation will fall back until then.**

Two problems with what came back:

1. **Its ontology bindings name the wrong things.** Term validation matched only 10 of 22 labels, and every drug CURIE it offered names a different drug:

   | Report called it | The CURIE actually is |
   |---|---|
   | `NCIT:C47737` tranexamic acid | Sulfamethoxazole |
   | `NCIT:C72713` pomalidomide | Azipramine |
   | `NCIT:C1794` thalidomide | Capecitabine |
   | `NCIT:C71146` pazopanib | Salirasib |

   No CURIE in this entry comes from the report. All bindings were taken from the committed caches and OAK.

2. **It mixes in a second disease.** `just preflight-dr` flagged it: KIT is mentioned 22 times, 42% of ENG's count. Inspection confirmed the cause — the report's KIT section rests on an *Alk1* mouse model (PMID:42579368), its gene-specific-programs claim on a SMAD4/ALK1 study (PMID:42460472), and its variable-expressivity section on PMID:10946360, which is titled for an HHT **type 2** kindred. Those sections are excluded.

Only two references from the report survived individual checking and are cited here: PMID:28564608 (endoglin–BMP9 crystal structure, explicitly HHT1) and PMID:39299239 (loss of heterozygosity as a second-hit route, marked `INDIRECT` because it works at the level of "an HHT gene" rather than reporting ENG separately).

The `0.0` confabulation rate only means the identifiers resolve. The report's own `needs_review` is `true`.

### Other corrections to the body

- **"44 references" is the report's citation count, not the entry's.** The entry cites **12** distinct references across **89** evidence items, every snippet verified against the cache.
- **"32/34 terms verified; 2 unverifiable"** is also the report's number, not the entry's. The entry passes `just validate-terms` outright.
- **The parent entry already listed HHT1 as a subtype.** My change there is one sentence cross-referencing the new standalone entry — not a new subtype record.
- **The cache files are markdown abstracts, not full-text PDFs.** The single PDF in the diff is the OpenScientist artifact.

### One trap worth flagging for future HHT work

Independently of the report, PMID:24957885 matches essentially every search term for "HHT mouse model" but is a VEGF-neutralization study in an **HHT2/Alk1** model. It was excluded. The model curated here is the Eng+/− mouse (PMID:10562296). Both this and the PMID:10946360 case are written into the entry's `notes` as a standing guardrail, since HHT1 and HHT2 are reported side by side and the numeral is often the only thing separating two titles.

### Two judgment calls to check rather than assume

- **Prevalence is derived, not surveyed.** The 10 per 100,000 figure is my arithmetic over two separately cited facts (HHT overall at 1 in 5,000; ENG contributing roughly half of solved cases). No population study of ENG-related HHT1 exists that I could find. The derivation is stated in the record's own `notes` — if you'd rather it not carry a number at all, say so and I'll drop to a class-only record.
- **Two GeneReviews items are deliberately excluded**: GI polyps / juvenile polyposis overlap (SMAD4) and pulmonary arterial hypertension (chiefly ACVRL1/GDF2). Reasons are recorded in `notes`.

### Validation

`just validate-disorders` passes over both changed files with 164/164 snippets verified. The entity-ref, causal-target, duplicate-key, enum-value, qualifier-term (including `--online`), folded-hyphen, title-snippet, snippet-grading and environmental-evidence gates all pass, and `just verify-datasets` resolves the one accession. Weighted compliance 93.1%. I also checked pathograph connectivity directly: no broken targets, no orphan nodes, no unreached phenotypes.

Happy to rewrite the PR body itself with the above if you'd prefer the corrections in one place rather than in a comment.

---

tannerzhang:
All three IMPORTANT items are fixed and pushed in `1ac7f8a5d`, along with both non-blocking suggestions. Each was verified against the primary source before I changed anything — the review was right on all three.

| Item | What I did |
|---|---|
| 1 — `evidence_source` | Corrected to `MODEL_ORGANISM`. Confirmed by census first: PMID:10562296 was `HUMAN_CLINICAL` on that one edge and `MODEL_ORGANISM` in its seven other uses. An isolated slip, as you said. |
| 2 — Brain MRI surveillance | Added a third `diagnosis` record quoting the GeneReviews text exactly ("brain MRI with and without contrast using sequences that detect blood products in infancy and again by age 18-20 years for cerebral AVMs"), paired with the 20%-vs-0% frequency evidence. |
| 3 — Mortality/prognosis | Added `clinical_burden` (HIGH) and a `progression` record using all three cached-but-uncited references. |
| 🔵 Heart failure | Added a **Congestive heart failure** phenotype with a `sequelae` edge from hepatic AVM. |
| 🔵 Biomarkers | Added a `biochemical` record for soluble endoglin from PMID:25873934. |

Three things worth flagging rather than burying:

**The mortality negative result needed careful handling.** PMID:33407668 finds brain VMs and pulmonary AVMs *not* associated with mortality — the two lesions that define HHT1. Quoted bare, that reads as "the HHT1 lesions don't matter." I curated it together with the authors' own explanation ("for which patients are routinely screened and treated preventatively at HHT Centres") as a separate evidence item, so the surveillance interpretation travels with the finding. I also added a `notes` field recording that none of the three studies stratifies ENG vs ACVRL1 — PMID:33407668 reports genotype only as SMAD4 against pooled "ACVRL1 or ENG" on four deaths — so the burden is HHT-wide and is explicitly not an HHT1 mortality estimate.

**The off-topic flag on PMID:42410718 was wrong.** The deep-research validator listed it as one of two `off_topic_references`. I read it before citing, per the skill's rule that off-topic is evidence rather than a verdict, and it is a Swedish national-register HHT mortality and prevalence study — squarely on topic.

**That paper also undercuts my own prevalence input**, so I've said so in the record rather than leaving it implicit: it measures HHT at 9.3 per 100,000, less than half the 1-in-5,000 figure the derived HHT1 estimate rests on, with underdiagnosis given as the reason. Taking the registry figure instead would roughly halve the entry's number. The `notes` now say the estimate is bounded by ascertainment. If you'd rather it carry no number at all, say so and I'll drop to a class-only record.

I also caught one defect of my own while doing item 3: the `reference_title` I first wrote for PMID:42410718 was invented and didn't match the cache. I then audited all 104 titled evidence items programmatically against their cache files — 0 mismatches remain.

On the biomarker record, I kept it deliberately narrow. The source is a review reporting no sensitivity, specificity or cut-off, so the record asserts candidacy only and a `notes` field says why no threshold is curated.

**Validation after the changes:** `just validate-disorders` passes over both files with **182/182 snippets verified**. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-folded-hyphens`, `check-title-snippets`, `check-snippet-grading` and `check-snippet-length` all pass. Pathograph re-checked programmatically — no broken targets and no unreached phenotypes (11 phenotypes, 10 pathophysiology nodes; the hepatic AVM leaf is now connected via the heart-failure sequela). Weighted compliance 93.1% → 94.1%.

Reference consumption is now 15 of the newly cached references rather than 12. You were right that the gap wasn't purely triage.

---

tannerzhang:
Thanks for the approval. I've acted on the suggestions in `cf158ff24` — flagging up front that this push drops the approval, which I judged worth it because one of the suggestions caught a **factual error in my own prose**, not just a style point.

**The one that mattered: my `notes` misdescribed PMID:33407668.** They said the paper reports genotype "only to compare SMAD4 against 'ACVRL1 or ENG' pooled". That is false. I checked the cache before changing anything, and the full text also carries:

> There was no significant difference in mortality between patients with ACVRL1 and ENG mutation (p > 0.05).

plus a Table 2 row, HR 1.54 (0.70–3.38), p = 0.286. You're right that this is the most HHT1-specific mortality statement in the cited literature, so it's now curated as its own evidence item rather than described as absent. The notes are rewritten to say what the paper actually reports, including that ENG was the commonest genotype in that cohort (496/886, 56%). The reading recorded is that HHT1 carries HHT's mortality burden rather than a distinct one — the opposite of the pattern for lesion distribution, which is worth a curator noticing.

**The prevalence band, and why that also settles the `SUPPORT`/`REFUTE` question.** You were right that 10.0 lands in `BAND_1_5_PER_10000` under `_band_from_rate`. Rather than move the band, I took your second option and recorded the derivation as the range it actually is: `rate_low: 4.7` / `rate_high: 10.0`, the two cited HHT inputs each halved for the ENG share. I dropped `rate_per_100000` entirely, since there genuinely is no single point estimate.

That resolves your `REFUTE` suggestion without a grade flip. Against a point value of 10.0 the Swedish figure did cut against the record. Against a stated range it *anchors the lower bound* — so `SUPPORT` is now straightforwardly correct, and `REFUTE` would be wrong. I think this is a better resolution than re-grading, because the disagreement was really the record overstating its own precision.

**Also done:** `diagnosis_term` on all three diagnosis records (`NCIT:C15709` Genetic Testing, `NCIT:C16525` Echocardiography Test, `NCIT:C16809` Magnetic Resonance Imaging), all validating against the dynamic enum.

**Causal edges — partially, on purpose.** Two of the twelve uncited edges had a genuinely causal source and now carry edge evidence: shunt → brain abscess (prophylaxis conditioned on shunting being present; marked `INDIRECT` since it states a preventive measure rather than incidence) and shunt → ischemic stroke (occlusion performed *for stroke prevention*, which asserts direction rather than co-occurrence).

I left the other ten uncited deliberately. Three are near-identity lesion→phenotype edges where a citation would only restate the node evidence, and CLAUDE.md is explicit that an uncited edge is acceptable but reaching for a node's citation to paper over an edge you cannot source is not. Getting from 9/21 to 21/21 would have meant doing exactly that, so the number is worse than it looks as a target.

**Not done:** the `BiomarkerReadout` link on soluble endoglin, since you said it's fine to leave for a research-stage marker.

**Validation:** `just validate-disorders` passes over both files with **185/185 snippets verified**. All offline gates pass. Pathograph re-checked — no broken targets, no unreached phenotypes. I re-audited all 107 titled evidence items against their cache files: 0 mismatches. Compliance 94.1% → 95.1%.

---

tannerzhang:
Fixed the merge-queue failure in `dad39ba99`. It was mine, and it's a small mistake with a lesson worth recording.

**What failed.** The merge-queue run for this PR ([33937777992](https://github.com/monarch-initiative/dismech/actions/runs/33937777992)) failed while every sibling PR in the same batch passed, so this wasn't a batch-mate's problem:

```
[ERROR] Title mismatch for geo:GSE197105:
expected 'RNA sequencing of isolated brain endothelial cells from Smad4, Alk1, or Eng endothelial cell-specific knock-out mutants'
but got  '...knock-out mutants, respective wildtypes (WTs), and Angiopoietin-2 inhibition transcriptional changes.'
```

I took the dataset title from `just discover-datasets` output, which **clips titles at display width**. The clipped string went into `datasets[0].title` and the reference validator, which resolves titles against the live GEO record, rejected it.

**Why my local run couldn't catch it.** `just validate-disorders` reports `Total checks: 0` in my sandbox but `Total checks: 1` in CI — the dataset-title check needs network access to GEO and was silently skipped locally. `just verify-datasets` *does* reach GEO, and it passed both before and after, because it only proves the accession resolves; it never compares the recorded title. So the entry had a green local signal and a real defect.

That gap seems worth knowing about generally: **a title copied out of `discover-datasets` should be treated as clipped**, and no local check in this repo will tell you otherwise.

I confirmed the corrected title independently against the NCBI esummary API (`db=gds`, uid 200197105) rather than just pasting what the CI error said it expected — the YAML and live GEO titles now compare equal.

**Also folded in** the `directness` nit from the approving review, since this commit touches the file anyway: the shunt → ischemic stroke edge now carries `directness: INDIRECT`, matching its sibling shunt → brain abscess edge. Both quote a preventive measure rather than an incidence.

**Validation:** `validate-disorders` passes over both files with 185/185 snippets verified; `verify-datasets` resolves the accession; entity-ref, causal-target, duplicate-key, enum-value and folded-hyphen gates all pass; pathograph re-checked with no broken targets and no unreached phenotypes. The dataset-title check still can't run here, so the next merge-queue attempt is the real proof.

This push drops the approval again. Ready to re-queue once re-approved.

---

tannerzhang:
**`claude-review` is failing for an account-quota reason, not a code reason, and it needs a maintainer.**

The job ([101240016618](https://github.com/monarch-initiative/dismech/actions/runs/33941607476/job/101240016618)) on head `dad39ba99` ended with:

```
detail="You've hit your limit · resets 7:20am (UTC)"
num_turns=10 cost=$1.07 subtype=success
```

The workflow's own error handler classifies this and states the remedy:

> This is an account-quota issue on the credential in use, NOT a problem with this PR. This workflow prefers `ANTHROPIC_API_KEY` (pay-per-use, no weekly cap) and only falls back to `CLAUDE_CODE_OAUTH_TOKEN` when no API key is set — so a weekly-limit error here most likely means `ANTHROPIC_API_KEY` is unset and the run used the subscription token.

So the fix is to set the `ANTHROPIC_API_KEY` repo secret, or rotate the OAuth token to an account with headroom, or wait for the reset. **None of those are mine to do** — I can't set repository secrets.

**What it blocks.** No review means no approval, and the deterministic sweep only queues approved PRs. So this PR can't re-enter the merge queue until the review runs, even though the code side is ready.

**On the re-run.** I have the means to re-trigger `claude-code-review.yml` and I intend to — but not yet. The quota resets at 07:20Z; re-running before then would just burn another failed job and post another red check. I've scheduled the re-trigger for after the reset and will report the verdict. If it fails on quota a second time, I'll say so once and stop retrying rather than looping on a credential problem.

**Not related to the earlier merge-queue failures.** Those were mine — the truncated `geo:GSE197105` dataset title — and are fixed in `dad39ba99`. `test (3.13)` is running on that commit now and is the check that actually validates the fix; I'll follow up with its result.

Worth noting alongside this: `EDISON_API_KEY` is also returning `403 Failed to authenticate`, which is why the deep-research run for this entry fell back to OpenScientist. Two separate credentials on the agentic-curation path are currently degraded.

---

tannerzhang:
**Correcting myself: I cannot re-trigger the review. This PR needs a maintainer to unblock it.**

In [my previous comment](https://github.com/monarch-initiative/dismech/pull/10351#issuecomment-5549203058) I said I had the means to re-trigger `claude-code-review.yml` and had scheduled it for after the quota reset. That was wrong, and I don't want anyone relying on it. The quota reset at 07:20Z, I attempted both routes, and both are denied:

- `POST .../actions/workflows/claude-code-review.yml/dispatches` → **403 Resource not accessible by integration**
- `POST .../actions/runs/33941607476/rerun-failed-jobs` → **403 Resource not accessible by integration**

So the failed `claude-review` check stays red until someone with workflow permissions acts. I'm not going to push an empty commit to kick CI — that's explicitly forbidden in this repo's guidance, and it would also dismiss nothing useful and add a junk commit.

**Where the PR actually stands on `dad39ba99`:**

| | |
|---|---|
| `test (3.13)` | ✅ **success** — this is the check that failed twice in the merge queue, so the dataset-title fix is confirmed by the only thing that can confirm it |
| `claude-review` | ❌ failure — Claude account quota, not code |
| Review decision | no approval on the current head (the last approval was dismissed by my fix push) |
| Mergeable | no conflict; base unchanged |

**What a maintainer can do, cheapest first:**
1. Re-run the `claude-review` job on [run 33941607476](https://github.com/monarch-initiative/dismech/actions/runs/33941607476) — the quota has reset, so it should now succeed.
2. Or set the `ANTHROPIC_API_KEY` repo secret, which the workflow prefers and which has no weekly cap. This is the durable fix; the failure text says the weekly-limit error most likely means the secret is unset and the run fell back to the subscription token.
3. Or approve manually, which lets the deterministic sweep pick it up.

I'll keep the PR watched rather than handing it off, and I'll report if anything changes. But I can't clear this one myself, and I'd rather say that plainly than keep retrying something that returns 403.

---

tannerzhang:
/review

---

tannerzhang:
**There is a one-word way to unblock this, and it does not need repo secrets or workflow permissions: comment `/review` on this PR.**

`claude-code-review.yml` fires on `issue_comment` when the body is exactly `/review` and the commenter is an OWNER, MEMBER or COLLABORATOR. That is far cheaper than the three options I listed [earlier](https://github.com/monarch-initiative/dismech/pull/10351#issuecomment-5550379842), and I should have found it sooner.

I can't use it myself. Every comment I post gets a Claude Code attribution footer appended automatically, so the body becomes `/review\n\n---\n_Generated by …_` and the gate's exact-string comparison fails. That is what the stray `/review` above is: it started [run 34013578966](https://github.com/monarch-initiative/dismech/actions/runs/34013578966), both jobs evaluated the condition and **skipped**. Editing the comment doesn't help either — the workflow only listens for `created`.

### Second thing worth knowing: the review credential is flaky repo-wide, not just here

Two failures this morning outside this PR:

| Time (UTC) | Run | Result |
|---|---|---|
| 04:59 | pr-shepherd [34012855006](https://github.com/monarch-initiative/dismech/actions/runs/34012855006) | `Run PR Shepherd` failed after 17 seconds |
| 05:16 | claude-code-review [34013573625](https://github.com/monarch-initiative/dismech/actions/runs/34013573625) (dispatched, another PR) | failure |

But three dispatched reviews succeeded between 02:11 and 02:44. So the credential works intermittently rather than being flat-out dead, which means a `/review` may need a second try — and setting the `ANTHROPIC_API_KEY` repo secret is still the durable fix, since the workflow prefers it and it has no weekly cap.

I could not read the reason for the pr-shepherd failure; the log I can retrieve is truncated to the cleanup steps. The 17-second duration is consistent with the agent erroring immediately, as it did here on Friday, but I am not claiming that as established.

This also means `pr-shepherd` is unlikely to rescue this PR on its own. Its priority-4 case ("REVIEW_REQUIRED whose review was cancelled or missing after the latest push") describes this PR exactly, and its agent lane is the thing that keeps failing.

### Where the PR stands — unchanged since Friday 04:05 UTC

`test (3.13)` is green on `dad39ba99`, which is the check that failed twice in the merge queue, so the dataset-title fix is confirmed. No merge conflict. No approval on the current head, because the last one was dismissed by that fix push. There is nothing left for me to change: the deferred `directness: INDIRECT` nit is already folded into `dad39ba99` and no reviewer request is outstanding.

---

tannerzhang:
**A merge conflict appeared overnight, so I rebased onto current `main` and force-pushed. One conflict, in a derived cache, and I resolved it by dropping a row rather than merging one in.**

The conflict was `cache/ncit/terms.csv`. This branch added exactly one row — `NCIT:C598,Iron` — and `main` has since inserted `NCIT:C583` between the same neighbours, which is what made the two versions collide.

That row is a leftover. It was written while I was trying `NCIT:C598` as the binding for iron; I ended up using `CHEBI:18248` instead, because `NCIT:C598` is not in the `ChemicalEntityTerm` enum. Nothing in `kb/` references it, and nothing anywhere in the repo outside the cache does. So I took `main`'s version of the file unchanged, which drops the row. No cache row was hand-written or hand-placed — the resolution only removes something that should not have been committed.

The branch is now 4 commits on `fcc278a5`, `dad39ba99` → `65a71b1eb0`, 69 changed files instead of 70 (the cache file is the one that left).

### Re-validated on the new base, because `main` moved 426 commits

That gap matters here specifically: a schema change that narrows an enum makes previously legal values illegal on merge, and neither side's CI can see it. So I re-ran the offline gates rather than assuming Friday's results still held.

| Check | Result |
|---|---|
| `just validate` | passed, 107/107 snippets verified |
| `check-enum-values` | no out-of-enum values |
| `check-entity-refs` | all references resolve |
| `check-causal-targets` | no new broken targets |
| `check-duplicate-keys` | none |
| `check-qualifier-terms` | clean |
| `validate-terms` | passed |
| `check-term-cache-integrity` | 18 term caches + 24 enum caches match the contract |

No KB content changed in the rebase — only the dropped cache row — so the evidence and pathograph results carry over unchanged.

`test (3.13)` and `claude-review` are both running on `65a71b1eb0` now. The push re-triggered the review through the ordinary `synchronize` path, which is the thing that had been stuck: the review has been red since Friday on an account-quota error, and I have no way to re-run it myself.

---

PR #10355, opened 2026-09-01
Add comprehensive 22q11.2 Deletion Syndrome research and pathophysiology

tannerzhang:
## Summary

This PR adds a comprehensive deep-research report on 22q11.2 Deletion Syndrome (22q11.2DS) and integrates new pathophysiology findings into the disorder knowledge base entry. The research was conducted using the OpenScientist autonomous research provider and includes 39 verified references.

## Key Changes

- **Research Report**: Added `research/22q11.2_Deletion_Syndrome-deep-research-openscientist.md` — a 957-line comprehensive disease characteristics report covering:
  - Disease information and identifiers
  - Etiology and genetic/environmental risk factors
  - Detailed phenotypes with HPO term mappings
  - Genetic and molecular information (causal genes, pathogenic variants, chromosomal abnormalities)
  - Environmental factors and gene-environment interactions
  - Diagnostic criteria and clinical management
  - Quality of life impacts and prognosis

- **Artifacts**: Added OpenScientist-generated final reports in HTML and PDF formats for reference and validation

- **Pathophysiology Integration**: Enhanced `kb/disorders/22q11.2_Deletion_Syndrome.yaml` with a new downstream pathophysiology edge:
  - Added causal link from TBX1 haploinsufficiency to velopharyngeal structural and myogenic insufficiency
  - Included intermediate mechanisms describing pharyngeal arch and palatal development
  - Integrated evidence from both model organism (PMID:11242110) and human clinical studies (PMID:40038168)

- **Reference Cache**: Added 39 new reference cache files (PMID entries) with full metadata and snippets for evidence validation, including:
  - Key papers on 22q11.2DS genetics, pathophysiology, and clinical manifestations
  - Studies on cardiac, immunological, neuropsychiatric, and craniofacial phenotypes
  - Research on modifier genes, epigenetics, and therapeutic interventions

- **Term Validation**: Updated ontology term caches to include UBERON:0001733 (larynx) for anatomical entity validation

## Implementation Details

- All 39 references passed validation with 0% confabulation rate
- Term validation identified 6 mislabeled terms and 1 obsolete term (CL:4023070), flagged for review
- Research metadata includes provider fallback information (Falcon → OpenScientist) and comprehensive validation statistics
- Evidence snippets are directly linked to PubMed references for traceability

## Notes

- The term validation report (`needs_review: true`) indicates some ontology label mismatches that may warrant follow-up curation
- The research report serves as a comprehensive source document for populating disease knowledge base entries

https://claude.ai/code/session_011HbRMKkMPoEkcqvTWWQueC

### DIFF
--- a/history/disorders/22q11.2_Deletion_Syndrome/2026-09-01T094023Z-claude-code-26f6b5.yaml
+++ b/history/disorders/22q11.2_Deletion_Syndrome/2026-09-01T094023Z-claude-code-26f6b5.yaml
@@ -1,0 +1,28 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: 22q11.2_Deletion_Syndrome
+  path: kb/disorders/22q11.2_Deletion_Syndrome.yaml
+session:
+  id: 2026-09-01T094023Z-claude-code-26f6b5
+  timestamp: '2026-09-01T09:40:23Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10355
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - evidence
+  summary: 'Edit: 22q11.2 Deletion Syndrome - address PR #10355 review (phenotype
+    frequencies, stronger evidence)'
+  details: |
+    Addresses the ai4c-reviewer review on PR #10355, which requested changes on one point. Every quote the reviewer proposed was verified as an exact substring of references_cache/PMID_39056299.md before use. Blocking item, phenotype frequencies: the reviewer noted that population frequencies for two of the three new phenotypes were quotable from the full text of PMID:39056299, a paper this PR already cited on an adjacent node, and that four peer phenotypes in the entry already carry frequency. Velopharyngeal insufficiency now carries frequency FREQUENT, supported by 'up to 80% of people with 22q present with VPI'; FREQUENT rather than VERY_FREQUENT because 80 percent is stated as an upper bound, and the deep-research report's own phenotype table independently gives roughly 65-70 percent. Delayed speech and language development now carries VERY_FREQUENT, supported by 'most people (95%) with 22q present with a speech-language disorder'. Both explanations record that these sentences are the paper's introductory summary of prior literature rather than findings of its own 14-patient cohort. Hypernasal speech is deliberately left without a frequency, as the reviewer agreed no clean population figure exists. Accepted suggestion on evidence quality: the reviewer identified selection bias in the Hypernasal speech evidence. The previous snippet quoted the preoperative severity split of the surgical arm, but inclusion criterion 3 of that study was 'presence of hypernasality on preoperative speech evaluation', so those proportions are true by construction and cannot evidence the phenotype. Verified against the cached methods section and replaced with a population-level statement that hypernasality is part of the syndrome's speech phenotype; the explanation records why the earlier quote was withdrawn. Accepted suggestion on the mechanism edge: a second evidence item was added to the Velopharyngeal structural and myogenic insufficiency -> Velopharyngeal insufficiency edge, quoting the sentence that enumerates the anatomic and physiologic velopharyngeal differences contributing to VPI. The prior snippet was Objective-section framing; it is retained for the multifactorial claim with a narrowed explanation. Declined suggestion, node splitting: the reviewer suggested splitting the mechanism node into increased pharyngeal depth and hypodynamic musculature, while explicitly not blocking. Declined because the newly added quote lists seven contributing differences rather than two, so a split into exactly two would impose a division the source does not make; both halves are also TISSUE scale, so this is a conjunction rather than the bundled-causal-step case the single-value biological_scale rule targets. Deferred suggestion, parkinsonism: out of the declared velocardiofacial scope of this PR and would widen it; left for a follow-up. A YAML indentation error was introduced while applying these edits, caught immediately by just validate, and fixed before any push. Revalidated: just validate and just validate-disorders pass with 72/74 snippets verified, up from 69/71, with the same two pre-existing DOI-prefix skips; check-entity-refs, check-causal-targets, check-duplicate-keys, check-enum-values, check-qualifier-terms, check-snippet-grading, check-snippet-length, check-title-snippets and check-folded-hyphens all pass.
--- a/kb/disorders/22q11.2_Deletion_Syndrome.yaml
+++ b/kb/disorders/22q11.2_Deletion_Syndrome.yaml
@@ -383,9 +383,20 @@ pathophysiology:
       snippet: >-
         Patients with 22q11.2 deletion syndrome (22q11DelS) often present with velopharyngeal dysfunction (VPD). VPD in patients with 22q11DelS is multifactorial beyond velopharyngeal insufficiency (VPI) alone, and differences in surgical outcomes are poorly understood.
       explanation: >-
-        Supports the edge specifically: the dysfunction in this syndrome is
-        multifactorial, which is what distinguishes this node from a simple
-        palatal-closure defect.
+        Establishes that velopharyngeal dysfunction in this syndrome is
+        multifactorial rather than a simple palatal-closure defect.
+    - reference: PMID:39056299
+      reference_title: >-
+        Do Palatoplasty Procedures Resolve Hypernasality as Effectively as Pharyngoplasty Procedures in Patients with 22q11.2 Deletion Syndrome?
+      supports: SUPPORT
+      directness: DIRECT
+      evidence_source: HUMAN_CLINICAL
+      snippet: >-
+        People with 22q can have multiple anatomic/physiologic differences of the velopharynx that contribute to VPI. These differences may include palatal clefting, increased pharyngeal depth due to velopharyngeal disproportion, decreased velar thickness, thin levator veli palatini muscles, discontinuity of the levator veli palatini muscles, decreased velar length, and pharyngeal hypotonia.
+      explanation: >-
+        States the edge directly: it names the anatomic and physiologic
+        differences of the velopharynx that contribute to velopharyngeal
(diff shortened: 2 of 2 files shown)


tannerzhang:
Three corrections to the PR description above, which was generated automatically and gets some details wrong.

**`UBERON:0001733` is soft palate, not larynx.** The description says larynx; larynx is `UBERON:0001737`. The term added here is soft palate, and it is bound as an anatomical location on the new velopharyngeal mechanism node, which is the right term for that node. Only the description is wrong — the YAML is correct.

**The reference cache gained 69 files in the second commit, not 39.** 39 is the number of references the research report cites; the cache holds more because the validator resolves everything it encounters.

**The description omits the first commit, which is the substance of this PR.** It describes only the causal edge added in the second commit, and not the node that edge points at or the phenotypes downstream of it.

## What this PR actually does

The request was to curate velocardiofacial syndrome. That disease is already in the KB: it resolves to `22q11.2_Deletion_Syndrome.yaml`, where `MONDO:0008644` (velocardiofacial syndrome) is bound as a `skos:narrowMatch`, added in #7391. The entry's own description already records that velocardiofacial syndrome, DiGeorge syndrome and CATCH22 are historical labels for the same deletion. So this is an augmentation of that entry, not a new disease entry.

The gap it fills is that the "velo" half of velocardiofacial had no representation of its own. Velopharyngeal insufficiency appeared only as words inside the `Palatal abnormality` description — there was no phenotype for it, no mechanism node, and no speech-surgery treatment.

Added:

- Three phenotypes: velopharyngeal insufficiency (`HP:0000220`), hypernasal speech (`HP:0001611`), delayed speech and language development (`HP:0000750`).
- One pathophysiology node, *Velopharyngeal structural and myogenic insufficiency* (`biological_scale: TISSUE`), covering the combined structural and muscular deficit.
- One treatment, velopharyngeal/pharyngoplasty surgery, linked to that node through `target_mechanisms`.

The chain now runs:

```
TBX1-associated pharyngeal      →  Velopharyngeal structural and  →  Velopharyngeal insufficiency
developmental vulnerability        myogenic insufficiency         →  Hypernasal speech
```

Note the upstream node is named *TBX1-associated pharyngeal developmental vulnerability*, not TBX1 haploinsufficiency as the description says. That wording is deliberate in the entry: it is careful not to claim TBX1 dosage alone explains the contiguous-gene syndrome.

## Evidence

Five new PMIDs, all directly about 22q11.2DS rather than about a gene that happens to be famous for something else:

| PMID | Used for |
|---|---|
| 36695155 | velopharyngeal insufficiency phenotype; matched-cohort pharyngoplasty outcomes |
| 39056299 | hypernasality phenotype; the pharyngeal-depth and hypodynamic-muscle mechanism |
| 36512754 | language impairment across all tested domains in a 44-child preschool cohort |
| 41800660 | persistent velopharyngeal dysfunction in a 2026 ENT review |
| 40038168 | neural-crest disruption causing both conotruncal lesions and velopharyngeal deficiency |

`PMID:40038168` came out of the research report and is the one genuinely useful thing it added: the TBX1 edge previously rested only on homozygous-mouse data, graded `INDIRECT` because a homozygous murine null is not dosage-equivalent to a human heterozygous multigene deletion. That paper states the link in human terms, so the edge now carries a `DIRECT` `HUMAN_CLINICAL` item as well. The mouse item is kept alongside it rather than replaced.

## On the research provider

Falcon was requested and is unreachable — `403 ProviderAuthError`, confirmed by `deep-research-client providers --check`. Rather than quietly substituting a provider, the run was re-issued with `--fallback` so the substitution is recorded by the tooling: the report's frontmatter carries `fell_back: true`, `requested_provider: falcon`, and a `provider_attempts` block including the falcon 403. It selected openscientist.

Someone with access may want to check whether the falcon key has expired, since this will affect every curation run.

## Two things reviewers should look at

**The pharyngoplasty-over-palatoplasty comparison is a 14-patient series**, 7 per arm. Its authors report it as a "may be less effective" signal. The treatment description says that explicitly rather than presenting it as a settled preference, but it is the weakest claim in the diff and worth a second opinion.

**`Delayed speech and language development` is deliberately left unconnected in the pathograph.** It would be natural to hang it off the velopharyngeal node, but the cited cohort found language scores below average even in children with intelligible speech, which argues against that edge. There was no source for an alternative, so no edge was invented.

## Validation

`just validate-disorders` passes: 69/71 snippets verified against the cache, with 2 skips that are pre-existing `DOI:`-prefix items. Also green: `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-not4curation`. All new items score 100% on compliance.

The report's own `needs_review: true` flag is worth reading but does not indicate a problem here. It fires on one obsolete term (`CL:4023070`) and six label variants that are table-parsing artifacts, such as `Craniofacial/upper aerodigestive:** palate`. None of the six are terms used in this entry, and `CL:4023070` appears nowhere in `kb/disorders`.

## One process note

Two deep-research invocations overlapped, and one wrote its openscientist output over the committed falcon report, its citations sidecar, and its artifact directory. That was my mistake. The falcon files were restored from git and verified byte-identical to a backup taken before the run; `git diff origin/main` on every falcon path is now empty. Falcon's `artifact-00.md` had been relocated rather than regenerated, confirmed by sha256 before it was moved back. No committed content was lost, but it is recorded in the history record rather than left implicit, and it is worth a glance during review.

I can rewrite the PR description to match the above if that is easier than reading this comment beside it.

---

tannerzhang:
Pushed `f3c663759`. The blocking frequency gap is fixed, and I took two of the four suggestions. Every quote proposed in the review was verified as an exact substring of `references_cache/PMID_39056299.md` before I used it.

## Blocking item — frequencies

You were right that the numbers were sitting in a paper this PR already cited on the adjacent node, and that the new phenotypes read as under-specified next to their peers.

| Phenotype | Frequency | Quote |
|---|---|---|
| Velopharyngeal insufficiency | `FREQUENT` | "up to 80% of people with 22q present with VPI" |
| Delayed speech and language development | `VERY_FREQUENT` | "most people (95%) with 22q present with a speech-language disorder" |
| Hypernasal speech | none | agreed, no clean population figure |

One qualification I added to both explanations: these two sentences are in the paper's Introduction, summarizing prior literature, not results of its own 14-patient cohort. The 95% one even carries a citation marker to another paper. They are still the right things to quote, but a reader should know the provenance, so the YAML says so.

`FREQUENT` rather than `VERY_FREQUENT` for velopharyngeal insufficiency because "up to 80%" is an upper bound, not a central estimate, which puts it in the 30–79% band rather than over it. Your independent 65–70% figure from the report's phenotype table points the same way.

## Suggestion 3 — selection bias. Accepted; this was a real error

This one was worth the review on its own. The previous snippet quoted preoperative severity in the surgical arm, and I checked the methods section: inclusion criterion 3 is "presence of hypernasality on preoperative speech evaluation". So the 43%/57% split is true by construction and cannot evidence the phenotype at all. My explanation claiming it recorded hypernasality "in every patient of a 22q11.2 deletion surgical series" was circular.

Replaced with a population-level statement from the same paper. The explanation now records why the earlier quote was withdrawn, so the mistake is visible rather than quietly overwritten.

## Suggestion 2 — better edge quote. Accepted

Agreed the old snippet was Objective-section framing rather than a finding about causation. The sentence you found enumerates the contributing differences directly and is a much better match for how the node is defined. Added as a second item; the original is kept for the multifactorial claim with a narrowed explanation.

## Suggestion 1 — splitting the node. Declined, with a reason

I think the split would make the entry slightly worse, and the quote from suggestion 2 is why. That sentence lists **seven** contributing differences — palatal clefting, increased pharyngeal depth, decreased velar thickness, thin levator veli palatini muscles, discontinuity of those muscles, decreased velar length, and pharyngeal hypotonia. Splitting into exactly two nodes would impose a division the source does not make, and would leave five of the seven with nowhere to sit.

Both halves are also `TISSUE` scale. The single-value `biological_scale` rule is a tripwire for a node bundling two *causal steps* at different scales; this is a conjunction of co-occurring features at one scale, which is a different shape.

## Suggestion 4 — parkinsonism. Deferred

Agreed it is a real gap and that `PMID:39918054` is the highest-value unconsumed reference here. Left out because it is outside this PR's declared scope and adding it would widen the diff on my own initiative. Worth its own PR.

## Suggestion 5 — description inaccuracies

Already corrected [in a comment above](https://github.com/monarch-initiative/dismech/pull/10355#issuecomment-5491804281), posted before your review landed — same soft-palate/larynx point. Your file count is more precise than mine: 73 added, of which 34 are cited by neither the report nor the YAML. I left them rather than rewriting a pushed branch to remove files that are harmless and policy-permitted, but I agree they are noise and would drop them if you or a maintainer would rather the PR carried only what it uses.

## Validation after the fixes

`just validate-disorders` passes, 72/74 snippets verified (up from 69/71), same two pre-existing `DOI:`-prefix skips. Also green: `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `validate-history`.

I introduced a YAML indentation error while applying these edits. `just validate` caught it before any push, and it is recorded in the history record rather than left out.

Ready for re-review.

---

tannerzhang:
Thanks. On the optional tidy — the changelog note in the `Hypernasal speech` explanation ("An earlier draft quoted...") — I agree it belongs in `history/` rather than entry prose, and I am **not** pushing it now.

Two reasons. You said you would not want it removed in a commit of its own, and I think that is right: `main` has `dismiss_stale_reviews` enabled, so any push here drops this approval and costs a full re-review cycle for a cosmetic edit. The note also earns its keep in the meantime as a guard against someone re-adding the selection-biased quote, which is the specific mistake it describes.

So it is deferred to whenever this entry is next edited for a substantive reason, and it is recorded here so the next person to touch the file knows it is a known cleanup rather than deliberate style.

The parkinsonism follow-up (`PMID:39918054`, already cached here) is the other open item, also deferred to its own PR.

`test (3.13)` is still running on `f3c663759`; I'll keep watching until this merges.

---

Issue #4873, opened 2026-06-27
curate NAMs from NAMsters symposium


mellybelly:
[NAMeRS Summit Final Remote Agenda_29-Jun-2026.pdf](https://github.com/user-attachments/files/29412914/NAMeRS.Summit.Final.Remote.Agenda_29-Jun-2026.pdf)

dragon-ai-agent:
## Automated curation-scanner assessment (high_effort run)

**Selection rationale.** Scanned the unassigned `curation -label:low_effort -label:medium_effort` queue. The older grouping-gap / follow-up issues are largely saturated or blocked: #1112, #4158, #3253, and the RNU12 item in #3691 are already implemented/closure-ready; #4239/#4240/#4241 self-audit as covered; #4242/#4243 remaining items are maintainer **scope decisions** (broaden-vs-spin-off grouping membership); #3324/#3544 are gated on an upstream MONDO snapshot (I confirmed `MONDO:1060226` is still absent from the local 2026-05-05 release). **This issue (#4873) is the only candidate with no prior agent triage and a safe, concrete first step**, so I selected it.

**Disease area / scope.** "Curate NAMs from the NAMeRS symposium" → enrich dismech entries with **New Approach Methodology** (organoid / organ-on-chip / iPSC / in silico / zebrafish) evidence. The auto-summary correctly maps several summit topics onto existing entries (Endometriosis, esophageal/lung cancer, DILI, anxiety, peripheral neuropathy). The summit itself is **2026-06-29**, so only *previously-published* presenter work is citable now.

**Safe first step taken → draft PR #4874.** I enriched the existing **`Endometriosis.yaml`** (the marquee NAM-for-endometriosis topic — Linda Griffith / MIT / COMPLEMENT-ARIE) with one verified NAM evidence item: a new `Dysregulated Peritoneal Metalloproteinase Activity` pathophysiology node anchored on **PMID:23157326** (Griffith/Lauffenburger/Han droplet-microfluidic protease-activity platform applied to endometriosis peritoneal fluid), `evidence_source: IN_VITRO`, bound to `GO:0022617`. Schema, references, and terms all validate.

### ⚠️ Hallucinated citations in the auto-summary — do not use
Per the DR-verification SOP, I fetched the two PMIDs the auto-summary suggested for the Griffith endometriosis work. **Both are hallucinated/misattributed:**
- `PMID:26052940` → a **ULK2 phosphorylation** paper (not endometriosis)
[...]

PR #10357, opened 2026-09-01
curate(NAM): post-review cleanup on the organ-chip and iPSC model records

mellybelly:
Clears the non-blocking observations the reviewer left open on #9665, plus one thing the #10003 evidence migration quietly dropped. Three disorder files, purely additive apart from two corrections described below.

## The one worth reading

`Dilated_Cardiomyopathy.yaml` carried `namo_type: namo:InducedPluripotentStemCellDerivedModel`. **That class does not exist — I invented it when I wrote the record**, assuming it would be there. The reviewer suspected as much on #9665 and flagged it as the item most worth chasing before the CURIE spread to other iPSC records. It was right.

Checked against the NAMO schema (`monarch-initiative/namo`, commit `0a743c5`, 56 classes). The other five `namo_type` values in this knowledge base are all real, so nothing else is affected:

| CURIE | Uses in `kb/` | Real? |
|---|---|---|
| `namo:Organoid` | 18 | yes |
| `namo:OrganOnChip` | 18 | yes |
| `namo:TwoDCellCulture` | 13 | yes |
| `namo:CoCulture` | 2 | yes |
| `namo:CellLineModel` | 1 | yes |
| `namo:InducedPluripotentStemCellDerivedModel` | 1 | **no** |

NAMO has no iPSC-derived class at all, and nothing in the remaining vocabulary fits: `namo:CellLineModel` is scoped to immortalized lines, `namo:ThreeDCellCulture` would misdescribe cardiomyocytes assayed as dissociated cells and clusters, and `namo:CellularSystem` is `abstract: true`. So the slot is now absent rather than approximated, per the no-term-beats-a-bad-one rule, with `experimental_model_type: IPSC_DERIVED_MODEL` carrying the distinction and `notes` recording why.

Worth flagging for the schema: **nothing validates `namo_type`.** There is no OAK adapter and no enum cache for the `namo:` prefix, so a fabricated CURIE passes `just validate`, `just validate-terms`, every snippet gate and the full pytest suite. That is how this one reached `main`. This PR fixes the single bad value, not the underlying hole.

The upstream gap is filed as monarch-initiative/namo#24 — the fix there is not a new class, since derivation is orthogonal to culture format, but typing the `cell_source` attribute that already exists and already names "iPSC-derived" in its own description while being untyped free text. 96 of the 419 model records in this knowledge base are iPSC-derived, the largest single category, so it will keep recurring until that lands.

## The rest

**`Dilated_Cardiomyopathy.yaml`** — the beta-adrenergic readout carried `direction: DECREASED` over a sentence reporting both decreases (beating rate, contraction) *and* an increase (cells with abnormal alpha-actinin). Split into two readouts, since `direction` is single-valued.

**`Dilated_Cardiomyopathy.yaml`** — #10003 retired `PARTIAL` and split directness onto its own axis. This record's two `supports: PARTIAL` gradings became plain `SUPPORT` with no `directness`, so the nuance they carried is gone: one of them was deliberately holding the authors' own "to some extent" qualifier. Assessed both and recorded the result — `DIRECT` on the sarcomeric link, where the quote asserts partial recapitulation itself, and `INDIRECT` on the neurohormonal link, where an acute receptor-level pharmacology result reaches the neurohormonal-axis claim only through an inference step. Worth knowing that migration was lossy anywhere `PARTIAL` was doing real work.

**`Non-Small_Cell_Lung_Cancer.yaml`** — added `CL:0001064` (malignant cell). The chip's `cell_source` named the NSCLC cells in prose while `cell_types` listed only the lung epithelium, so the defining population was missing from the structured field.

**`Esophageal_Adenocarcinoma.yaml`** — the `Chemotherapy Resistance` node made two claims while its single evidence item covered one. Added the organoid-comparator result as a second item (`directness: INDIRECT` — it is a comparison between model systems, from which the stromal contribution follows by inference). Also softened "faithfully observable only in systems that retain the tumour-stroma interface" to "reproduced more faithfully by" — the comparison supports a relative claim, not an exclusive one. The node still asserts no causal edge and its `notes` still say why.

## Validation

Rebased onto current `main` (`169ca345d`) and re-run there, since four of the gates baseline against `origin/main`:

```
just validate-disorders (x3)     All 3 disorder file(s) passed validation
just check-entity-refs (x3)      OK
just check-duplicate-keys (x3)   OK
just check-snippet-grading       OK: no new divergences
just check-title-snippets        OK: no new title-quoting snippets
just check-folded-hyphens        OK
just check-snippet-length        OK: no new snippets under 5 words
just check-environmental-evidence OK
just validate-history-all        No issues found
pytest -k "model_readout or failure_to_recapitulate or entity_ref or subtype_foreign or snippet_grading or directness"   13017 passed
```

Six files: three disorder entries plus three matching history records. No cache churn, no derived files. Post-rebase diff reviewed — no unrelated deletions or reversions.

Follows #9665. Part of #4873.

---
_Generated by [Claude Code](https://claude.ai/code/session_016dzPSEmY9xGSX45XCRukqA)_

### DIFF
--- a/history/disorders/Esophageal_Adenocarcinoma/2026-08-29T161637Z-claude-code-e42997.yaml
+++ b/history/disorders/Esophageal_Adenocarcinoma/2026-08-29T161637Z-claude-code-e42997.yaml
@@ -20,7 +20,7 @@ events:
 - type: EDIT
   outcome: changed
   sections:
-  - experimental_models
+  - pathophysiology
   summary: 'Post-review cleanup: second evidence item and a softened claim on the
     chemoresistance node'
   details: |
--- a/kb/disorders/Dilated_Cardiomyopathy.yaml
+++ b/kb/disorders/Dilated_Cardiomyopathy.yaml
@@ -1614,8 +1614,7 @@ experimental_models:
       interpretation: >-
         The same adrenergic exposure spread sarcomeric disorganisation to more cells, so
         catecholamine stress aggravates the structural lesion and not only the functional
-        one. Split from the contractile readout because the two measurements move in
-        opposite directions and `direction` is single-valued.
+        one.
       evidence:
       - reference: PMID:22517884
         reference_title: "Patient-specific induced pluripotent stem cells as a model for familial dilated cardiomyopathy."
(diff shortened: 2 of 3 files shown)


mellybelly:
Pushed `44fde1db3`, which takes the blocking item and all four suggestions.

**The blocking one was a real mistake, not just a label.** The EAC history record said `sections: [experimental_models]`, but this PR does not touch that block at all — the change is entirely inside the `Chemotherapy Resistance` pathophysiology node. Now `pathophysiology`.

Suggestion 1 was the most useful, because it caught a hole in my own argument. The notes ruled out `CellLineModel` and `ThreeDCellCulture` and simply did not mention `TwoDCellCulture`, which is a fair description of single cardiomyocytes on AFM and MEA — and the PR's own upstream reasoning (derivation is orthogonal to format) makes the missing iPSC class a bad excuse for leaving *format* unrecorded. The rewritten notes reject all three explicitly: this system spans the 3D embryoid-body stage and the 2D dissociated assay, so any single format value records half of it. That is the reason the slot is absent, and it is a better reason than the one I gave.

Suggestion 3 was also a genuine correction. My explanation read the comparator as "adding patient-matched fibroblasts to the same organoids", but the abstract compares the chip against *static* organoid-only culture — the chip differs by perfusion and preserved architecture as well as stroma, so the improvement cannot be attributed to fibroblasts alone. The explanation now names that confound.

The other two are done as described: the "split from the contractile readout because…" sentence moved from `interpretation` (which renders on the disorder page) into the model `notes`, and the coined CURIE is now written without its `namo:` prefix. Verified the census is clean — 0 hits for the prefixed string across `kb/`, and all 97 live `namo_type` values are real NAMO classes.

Nothing was left undone, and no claim was softened beyond what the sources carry.

```
just validate-disorders (x3)      All 3 disorder file(s) passed validation
just check-entity-refs (x3)       OK
just check-duplicate-keys (x3)    OK
just check-enum-values            OK (2875 KB files)
just check-causal-targets         OK
just check-snippet-grading        OK: no new divergences
just check-title-snippets         OK
just check-folded-hyphens         OK
just check-snippet-length         OK
just validate-history-all         No issues found (7961 records)
```

Thanks for pulling the NAMO schema and checking the central finding independently rather than taking the table at face value — that is the part of this PR most likely to have been wrong.

---

PR #10358, opened 2026-09-01
Add an audit of how diet is represented, on both the causal and intervention tracks

mellybelly:
Diet enters a disease entry two ways — as something that causes or worsens the disease, and as something prescribed to treat it — and until now nothing measured either. This adds `just diet-audit`, which counts both, plus a report of what it found. No knowledge-base content changes here.

The headline finding: **three diet annotations are drawn as edges in the mechanism graph with no evidence behind them at all**, and a further 38 causal annotations have solid human evidence but never reach the graph.

> **Correction.** Earlier versions of this description said 19 rather than three. That was a bug in the audit, caught in review: it graded only an entry's own `evidence:` block and never read the evidence on the pathograph link itself, which is where CLAUDE.md says the claim "this exposure acts on this node" belongs. Sixteen entries were fully cited on the link and reported as unevidenced, including all three `kb/modules/` entries an earlier draft singled out as the worst case. Fixed in `b894faa8`; the report explains it in place rather than quietly restating the number.

## The two tracks are audited separately, on purpose

| | Lives in | Grounded by | Reaches the pathograph via |
|---|---|---|---|
| Causal | `environmental[]` | `food_source`, `exposure_term` | `influences_mechanisms` |
| Intervention | `treatments[].treatment_term` | `dietary_modifications` | `target_mechanisms` |

Phenylketonuria already carries both against the same `FOODON:00001006` — meat as an exposure, and restrict-meat as a prescription. That is not duplication to factor out: an exposure claim and a prescription claim are different assertions with different evidence. The audit never reconciles one track against the other, and never reports one as missing because the other exists.

## What it counts

Not ontology-binding coverage. A diet annotation earns a place in the mechanism graph when there is evidence for it, so the gap worth working is *cited but unlinked*.

| | Causal | Intervention |
|---|---|---|
| Diet-related entries | 191 in 139 files | 648 in 474 files |
| On the pathograph | 139 (72.8%) | 331 (51.1%) |
| Cited but off the pathograph | 50 (38 after dropping weak matches) | 282 |
| On the pathograph, no evidence anywhere | 0 | 3 |

The causal side is in good shape — nearly three entries in four already carry `influences_mechanisms`. The intervention side is the weaker half: half of dietary treatments never link to a mechanism node.

The three unevidenced edges are all interventions: dietary protein restriction in Chronic Kidney Disease, the ketogenic diet in Dravet syndrome, and threonine restriction in Inherited Threoninemia.

These figures are as of 2026-09-05, and the report says so. They go stale by construction — the KB moved noticeably while this branch was open, and every number here was recomputed after the rebase rather than adjusted by hand.

## Unbound terms are reported, never failed

A third of causal entries and almost all intervention entries carry free text rather than an ontology term, and much of that is correct. `FoodTerm` is reachable only from `FOODON:00001002` (food product) and `CHEBI:33284` (nutrient), so food *components* fall outside it: Celiac Disease's three grain vehicles are bound while gluten itself, the actual trigger, cannot be. Named dietary patterns have no FOODON home at all. Where no good term exists, free text is the right answer, so the audit reports `FREE_TEXT` as a state to review and the binding census is deliberately secondary.

Widening the `FoodTerm` root to admit the FOODON `food material` branch would fix the component case. That is a schema change, and it is listed as a recommendation rather than done here. Worth flagging for whoever takes it: `conf/oak_config.yaml` confirms that excluding `food material` terms is intended behaviour rather than an adapter artifact, but *why* the root was drawn where it was is not recorded anywhere.

## Accuracy of the matching

Entries are found by keyword, so there is a false-positive tail. Two things keep it visible rather than hidden.

The two tracks read different fields. An `environmental[]` entry is short and wholly about its exposure, so its description is signal. A treatment description is a clinical paragraph that mentions diet incidentally — searching it pulled in ACE inhibitors (on "sodium"), cleft palate repair (on "feeding"), and beta blockers. Matching treatment names only cuts that track from 1,244 entries to 648. Bare "sodium" was dropped from the keyword list entirely: adding it back pulls in 40 more entries that are almost all drugs, and every genuinely dietary one carries "diet", "dietary", "salt" or "intake" anyway.

Where each match came from is recorded per entry, so the weak tail can be filtered with `--strong-only`. It is 12 of the 50 causal gap rows — Ependymoma's "high-dose ionizing radiation", whose description happens to mention diet, is one.

## Evidence grading

Structural, not a quality judgment — the script cannot read a paper. `CITED_HUMAN` means a `SUPPORT` item with a real snippet graded `HUMAN_CLINICAL`. Evidence is read from the entry's own block *and* from its pathograph links.

`REFUTE_ONLY` is its own tier. An earlier pass treated any entry without `SUPPORT` as uncited, which made snippet-backed `REFUTE` evidence indistinguishable from none: NELABA's "Lipoic acid supplementation (ineffective)" carries two `REFUTE` items against the mechanism it targets — a correctly curated negative result — and got flagged as an unevidenced edge. That tier and the link-evidence fix above are the same defect at two levels, and both would have sent a curator to fix entries that were already right.

## Caveats

A `CITED_HUMAN` row is a candidate, not a verdict. Most of these citations are observational cohort associations, and an association is not a mechanism — Gout's shellfish evidence measures incident gout rather than precipitation of a flare, as its own `explanation` already notes. Acting on the causal candidates needs a curator reading each snippet, which is why it is not in this PR; #10359 does that work and, once merged, will shrink the list of 38.

`.gitignore`'s bare `reports/` rule also matches `docs/reports/`, so the report needed `git add -f`, as all 45 existing reports there did. Probably worth fixing separately.

## Next steps, not in this PR

1. Cite or unlink the three unevidenced pathograph edges.
2. Work the 38 causal candidates, reading each snippet. Gout, Phenylketonuria, and Celiac Disease are 11 of them — that is #10359.
3. Standardize the dietary-pattern CURIEs on one ECTO term per pattern, retiring the bare `XCO:0000013` catch-all.
4. Decide on the `FoodTerm` root.

https://claude.ai/code/session_01JYdMMSJh66114t46BBLmsk

### DIFF
--- a/docs/reports/diet-representation-audit-2026-09-01.md
+++ b/docs/reports/diet-representation-audit-2026-09-01.md
@@ -95,9 +95,12 @@ neither is a curation failure:
 1. **`FoodTerm` excludes food components.** It is reachable only from
    `FOODON:00001002` (food product) and `CHEBI:33284` (nutrient). Gluten
    (`FOODON:03420177`) sits under `food material` instead and is correctly
-   rejected — as `conf/oak_config.yaml` documents deliberately. So Celiac's three
-   grain vehicles are bound while its actual trigger, "Gluten Exposure", cannot
-   be. The same will apply to casein, purines, and oxalate.
+   rejected. `conf/oak_config.yaml` confirms that exclusion is intended
+   behaviour rather than an adapter artifact — it uses two other `food material`
+   terms as worked examples of CURIEs both adapters agree are not `FoodTerm`
+   values. So Celiac's three grain vehicles are bound while its actual trigger,
+   "Gluten Exposure", cannot be. The same will apply to casein, purines, and
+   oxalate.
 2. **Dietary patterns have no ontology home at all.** FOODON describes food
    products, not eating patterns.
 
@@ -186,8 +189,11 @@ curator, not automated fixes.
    retiring the bare `XCO:0000013` catch-all.
 4. **Widen the `FoodTerm` root** to admit the FOODON `food material` branch, so
    food components (gluten, casein) become bindable. This is a schema change with
-   a cache rebuild, and needs its own decision-register entry — the current narrow
-   root is deliberate, and `conf/oak_config.yaml` explains why.
+   a cache rebuild, and needs its own decision-register entry. Note that *why*
+   the root was drawn at `FOODON:00001002` is not recorded anywhere —
+   `conf/oak_config.yaml` documents only that the resulting exclusion is
+   intended, not the reasoning behind the root itself. That missing rationale is
+   a reason to take the decision deliberately, not a licence to widen it.
 5. **Leave `dietary_modifications` alone for now.** At 5 files it is barely load
    bearing, and populating it is only worth doing after item 4 settles what can be
    bound.
--- a/scripts/diet_audit.py
+++ b/scripts/diet_audit.py
@@ -322,12 +322,8 @@ def _classify_intervention(tx: dict) -> _DietItem | None:
     """Classify one ``treatments[]`` entry, or None if it is not diet-related."""
     term = tx.get("treatment_term")
     term = term if isinstance(term, dict) else {}
-    tid = str(
-        ((term.get("term") or {}) if isinstance(term.get("term"), dict) else {}).get(
-            "id"
-        )
-        or ""
-    )
+    inner = term.get("term")
+    tid = str((inner if isinstance(inner, dict) else {}).get("id") or "")
     name = str(tx.get("name") or "(unnamed)")
     mods = term.get("dietary_modifications")
     has_mods = bool(mods) and isinstance(mods, list)
@@ -630,7 +626,9 @@ def main(argv: list[str] | None = None) -> int:
             print(f"Wrote {args.out}", file=sys.stderr)
 
     if args.strict:
-        bad = [i for i in all_items if i.unevidenced_link]
+        # summary_scope, not all_items: --track causal --strict must not fail on
+        # an intervention defect the caller asked not to be shown.
+        bad = [i for i in summary_scope if i.unevidenced_link]
         if bad:
             print(
                 f"STRICT: {len(bad)} diet entries are on the pathograph with no "
--- a/tests/test_diet_audit.py
+++ b/tests/test_diet_audit.py
@@ -387,3 +387,22 @@ def test_audit_is_advisory_by_default(monkeypatch, capsys):
     monkeypatch.setattr(audit, "_REPO_ROOT", ROOT)
     assert audit.main([]) == 0
     assert "CAUSAL track" in capsys.readouterr().out
+
+
+def test_strict_honours_the_track_filter(monkeypatch, tmp_path, capsys):
+    # An unevidenced link on the intervention track only. `--track causal
+    # --strict` must not fail on a defect the caller asked not to be shown.
+    _write(
+        tmp_path,
+        "Example",
+        """treatments:
+- name: Ketogenic Diet
+  target_mechanisms:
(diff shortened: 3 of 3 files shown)


mellybelly:
`test (3.13)` is red on a network timeout, not on anything in this PR.

The failing step is `validate-terms-schema`, which could not reach EBI's ontology service:

```
🌐 Unable to validate at this time: ontology service unavailable.
   could not reach ontology service to resolve NCIT:C2852:
```

That step validates the schema's dynamic enums. This PR adds a script, a test file, a report, a justfile recipe and an mkdocs nav line — it does not touch `src/dismech/schema/`, so there is nothing here that could change what that step resolves. The tool says so itself.

The two real failures before this were mine and are fixed: three ruff findings (an ambiguous `l` variable and two f-strings without placeholders), then `EXE001` for a shebang on a non-executable file. Lint passes now — this run got past it to the term validation. I missed the second one locally because CI lints with `uvx ruff check .`, which fetches the latest ruff, while `uv run ruff` uses the older pinned version; I now check with CI's exact command.

I can't re-run the job myself (403 on the rerun API), so this needs a re-run from someone who can. Nothing to change in the branch.


---

mellybelly:
Green now — no re-run needed, disregard the request in my previous comment.

The ontology-service timeout did not recur. In the meantime this branch was rebased onto current main (it had gone into conflict), which re-triggered CI on its own. `test (3.13)` and `claude-review` both pass on `e21f6683e`.

The one conflict was in `mkdocs.yml`, where main and this branch each added a report to the top of the same nav list. Both are kept, newest first.


---

mellybelly:
Rebased onto current `main` and regenerated every figure. That drops the approval, so this needs a fresh review — sorry for the round.

**Why the numbers moved.** The KB grew while this branch was open, and `main` added causal pathograph links of its own. Causal entries went 183 → 191, interventions 614 → 648, and the strong causal candidate list shrank from 40 to 38. I recomputed each figure rather than adjusting any by hand, including the four that are not in the script's own output and so are easy to leave stale:

- the description-matching counterfactual, 1,262 → 1,244 pre-filter
- the bare-`sodium` counterfactual, which was 48 and is now 40 — that one also appears in the script's own comment, so both are fixed
- `dietary_modifications`: 11 records in 5 files, unchanged
- FOODON bindings: 19 across 11 terms, unchanged

The alcohol CURIE-scatter block you flagged as drifted last round is regenerated too — a third binding, `ECTO:0000509`, has appeared since it was written. I have removed the "illustrative, not current" caveat from the PR body, since the block is now current, and added an as-of date to the report instead. These figures go stale by construction and it is better to say when they were taken than to imply they are permanent.

Two things are unchanged and worth stating: the three uncited pathograph edges are still exactly Chronic Kidney Disease, Dravet syndrome, and Inherited Threoninemia, and the `REFUTE_ONLY` counts still hold.

**Your three non-blocking notes are all done**, in the same push rather than a later one:

- `--strict` ignored `--track`, so `--track causal --strict` failed on an intervention defect the caller had asked not to see. It now uses the same scope the summary does, with a regression test (23 tests now).
- The `or {}` in the treatment-term id lookup was dead. Rewritten as a plain guard; summary output is byte-identical.
- The `conf/oak_config.yaml` over-citation is corrected. That file documents that excluding `food material` terms is *intended behaviour* rather than an adapter artifact — it uses two of them as worked examples — but it says nothing about why the root sits at `FOODON:00001002`. Recommendation 4 now says that rationale is unrecorded, which is a reason to take the decision deliberately rather than a licence to widen the root.

The PR body carried the old numbers, as you noted; it is rewritten.

`uvx ruff check .` is clean, `ruff format --check` passes, and 23 tests pass.

---

PR #10361, opened 2026-09-01
enhance(ARDS): model the inhalational-toxicant exposure routes

caufieldjh:
## What

ARDS had **eleven pathophysiology nodes and no `environmental` block at all** — even though its own trigger node names *"inhalational injury"* among the direct pulmonary insults. The toxicological arm of ARDS was describable in the prose and entirely absent from the pathograph.

## Two exposure routes, wired to the trigger

| Exposure | Evidence |
|---|---|
| **Irritant halogen gases** (chlorine, bromine) | *"Chlorine is highly toxic by inhalation, leading to dyspnea, hypoxemia, airway obstruction, pneumonitis, pulmonary edema, and acute respiratory distress syndrome"* (PMID:29722341); human clinical confirmation that severe exposure reaches ALI/ARDS (PMID:20601629) |
| **Smoke inhalation** | *"Approximately 22% of all burn patients are diagnosed with SI-ALI, leading to at least 30% of the overall fire-associated mortality"* (PMID:31068086) |

Both link by `influences_mechanisms` (`TRIGGERS`, `DIRECT`) to `Direct Pulmonary or Indirect Systemic Insult`, and both carry their own top-level `evidence` as the CI gate requires.

## The term binding is the part worth reviewing

Both bind the substance-level `ECTO:0000231` (*exposure to chemical*) with the specific agent and route in `preferred_term` and `description`.

I **deliberately did not** bind the smoke entry to `ECTO:6000029` (*exposure to tobacco smoking*), even though it is already in the validated cache and would have passed instantly. It denotes tobacco smoking, which is not what happens when someone inhales combustion products in a house fire. Reusing it would have satisfied every check in the stack while misclassifying the exposure — id and label agreeing with each other about the wrong thing.

The ontology service was unreachable throughout curation, so a more specific CURIE could not be verified. That constraint is recorded in the `notes` of both entries rather than left implicit, so a later curator knows to revisit rather than assuming the broad term was a considered final answer.

## Also

`biological_scale` tagged on all eleven nodes, which previously had none — TISSUE for insult, ventilator injury, barrier disruption, coagulation, edema, repair and fibrosis; CELLULAR for the macrophage, neutrophil and immunometabolic nodes; ORGANISM for shunt physiology and hypoxemic respiratory failure.

**No existing node, evidence item, conformance declaration or treatment was altered.**

## Validation

- `just count-verified-snippets` — **78/78** verified
- `just check-environmental-evidence` — passed
- Graph — 34 edges, no integrity issues, both exposures reaching the trigger node


caufieldjh:
Fixed in `f30ef77`. Every finding was correct, including the blocking one — and the framing of it was the useful part.

**On the blocking point.** You're right that the problem isn't just the broad term, it's the `notes:`. I recorded a transient network outage as the rationale for a binding decision, which reads as considered reasoning to anyone who comes along later and finds the search works fine. That's a stale justification calcifying into apparent intent. Both blocks now say what was actually found.

I verified both CURIEs independently before changing anything — `ECTO:7000059` and `ECTO:9000077` both resolve, and `just validate-terms` passes on the file.

**Smoke → `ECTO:7000059`.** Your branch argument is the one that convinced me: it wasn't merely broad, it was in the wrong place. ECTO files *exposure to smoke* under *environmental material*, not *chemical* — which is precisely the argument the PR body was already making for itself. I had the right reasoning and drew the wrong conclusion from it.

**Halogens → split, taking your option (a) plus the bromine evidence.** The entry is now `Inhalation of chlorine gas` bound to `ECTO:9000077`, so name, binding and quoted evidence all describe one agent. Bromine became its own entry carrying the sentence you pointed at, which closes finding 2 as well.

The bromine entry is **deliberately unbound**. I searched ECTO via the service and confirmed your finding: no bromine exposure class exists. The nearest hit is `ECTO:9002162` (*exposure to brominated flame retardant*) — a different agent. I also rejected falling back to `ECTO:0000231`, since that's the binding this commit is narrowing away from. `environmental-term-audit` will report it UNBOUND, and the `notes:` says that's the correct outcome rather than a gap.

**All four non-blocking findings taken:**

1. `exposure_classifications` on all three entries — agent type, route, duration, with notes distinguishing the acute release events that cause ARDS from the chronic low-level exposures that cause airway hyperreactivity. Thanks for the `Asbestosis` pointer; the slot is a better home for the route than prose.
2. Bromine now evidenced — handled by the split above.
3. SI-ALI burden statistic regraded `OTHER` → `HUMAN_CLINICAL`. Agreed: it's an epidemiological figure about burn patients.
4. `links.prs` now carries 10361.

Both new CURIEs are committed to `cache/ecto/terms.csv` and the `exposureterm` enum cache, so CI validates deterministically rather than depending on a live lookup succeeding.

**One self-inflicted problem, flagged because it's visible in the diff.** The title-generation helper I adopted earlier in this session — after fabricating four titles on a sibling PR — assumes one line per `reference_title`. On wrapped scalars it replaced the first line and orphaned the continuation, breaking the YAML and cosmetically rewriting 34 unrelated titles. I fixed the break, verified no title's *content* changed and that all still match their cache, then restored the original formatting so the diff carries 2 title lines instead of 68. The tool I introduced to stop one failure mode needed a guardrail against another.

Gates after the fixes: `just validate-disorders` passed, `count-verified-snippets` 80/80, `validate-terms` passed, `check-duplicate-keys` clean.

PR #10362, opened 2026-09-01
enhance(Secondary_Hypertension): connect the orphaned node, add the three VEGF mediator arms

caufieldjh:
## What

Two structural defects in an entry that was 171 lines — the thinnest of the toxicology-module conformers.

## 1. A pathophysiology node that was outside the graph

`Sympathetic and hormonal coactivation` had **no `role`, no `biological_scale`, and no `downstream` edges**. It rendered on the disorder page while sitting entirely outside the causal graph — so the entry's own stated *"shared final pathway"* for the non-pharmacological causes of secondary hypertension connected to nothing at all.

It is now a `central_effector` at ORGANISM scale with an edge to the `Hypertension` phenotype, and its description says which causes converge on it (renal, endocrine, sleep apnea, coarctation) and notes explicitly that the drug-induced VEGF arm reaches hypertension *without* passing through it — two genuinely different routes to the same phenotype.

## 2. Three mediator arms named in prose but absent as nodes

The VEGF arm jumped straight from pathway inhibition to vasoconstriction, while its own edge description named all three intervening mechanisms. Each is now a node conforming to its counterpart in `vegf_inhibitor_hypertension`:

| New node | Conforms to | Why it matters |
|---|---|---|
| Endothelial Nitric Oxide Deficiency and Oxidative Stress | `#Endothelial Nitric Oxide Deficiency and Oxidative Stress` | VEGF drives eNOS via PI3K/Akt and MAPK, so inhibition withdraws vasodilator tone |
| Endothelin-1 System Activation | `#Endothelin-1 System Activation` | The vasoconstrictor axis acting on smooth muscle |
| Microvascular Rarefaction | `#Microvascular Rarefaction` | The arm that best explains why the pressure rise is **sustained** — same cardiac output through fewer parallel resistance vessels |

**Node conformances 2 → 5. Graph edges 3 → 8.**

Term bindings were copied from the module's own nodes rather than recalled, so the id/label pairs are ones already verified in this repo. All six evidence PMIDs were already in the committed cache — no new cache files needed.

## Process note

All four new `reference_title` values were initially **fabricated** — plausible-sounding titles belonging to no such paper (e.g. "Hypertension and cardiotoxicity associated with protein kinase inhibitors" for a paper actually titled "Tyrosine Kinase Inhibitor-Induced Hypertension: Role of Hypertension as a Biomarker in Cancer Treatment"). The title check caught all four, and they were regenerated programmatically from each cache file's own `title:` line.

Flagging it because the snippets and ontology terms in this PR were verified by tooling before being written, while the titles were typed from apparent memory — and a wrong title silently misattributes real evidence to the wrong paper.

## Validation

- `just count-verified-snippets` — **12/12** verified
- Graph — 8 edges, no integrity issues, no phenotype without an incoming edge

### DIFF
--- a/history/disorders/Secondary_Hypertension/2026-09-01T122517Z-claude-code-c3786f.yaml
+++ b/history/disorders/Secondary_Hypertension/2026-09-01T122517Z-claude-code-c3786f.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10362
   urls: []
 events:
 - type: EDIT
--- a/kb/disorders/Secondary_Hypertension.yaml
+++ b/kb/disorders/Secondary_Hypertension.yaml
@@ -5,9 +5,13 @@ description: >-
   to an identifiable underlying cause, in contrast to essential hypertension.
   Causes include renal parenchymal and renovascular disease, endocrine disorders
   (primary aldosteronism, pheochromocytoma, Cushing syndrome, thyroid disease),
-  obstructive sleep apnea, aortic coarctation, and certain drugs. The shared final
-  pathway is sympathetic and hormonal coactivation with sodium retention; treating
-  the underlying cause can reverse the hypertension.
+  obstructive sleep apnea, aortic coarctation, and certain drugs. These causes
+  share an endpoint rather than a mechanism: sympathetic and hormonal
+  coactivation with sodium retention is a common theme across several of them,
+  but primary aldosteronism raises pressure through renin-independent
+  mineralocorticoid signalling and volume retention, coarctation is
+  substantially mechanical, and VEGF-inhibitor toxicity acts through endothelial
+  vasoconstriction. Treating the underlying cause can reverse the hypertension.
 category: Complex
 parents:
 - Hypertensive Disorder
@@ -131,7 +135,7 @@ pathophysiology:
     raises oxidative stress in the endothelium. This is the first of the three
     mediator arms the entry previously named only in prose.
   role: amplifier
-  biological_scale: MOLECULAR
+  biological_scale: CELLULAR
   conforms_to: "vegf_inhibitor_hypertension#Endothelial Nitric Oxide Deficiency and Oxidative Stress"
   cell_types:
   - preferred_term: vascular endothelial cell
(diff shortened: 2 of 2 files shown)


caufieldjh:
Fixed in `02d5e75`. All three were correct, and **IMPORTANT 1 is one I should have caught myself** — thank you for going after it properly rather than treating it as a wording quibble.

## IMPORTANT 1 — the overreach

You're right, and the gap is exactly as you characterise it. I wrote *"**the** shared final pathway across the non-pharmacological causes"* and then enumerated seven, from a snippet that says *"a common theme across **several** contributors"*. Indefinite and partial became definite and exhaustive.

And it isn't only inflation — it's wrong on mechanism for two of them, as you show. Primary aldosteronism reaches pressure by renin-*independent* MR signalling and ENaC-mediated volume retention, with reciprocal renin suppression, which is close to the opposite of sympathetic coactivation. Coarctation is substantially mechanical. That the contradiction was sitting in the deep-research report **committed in this same PR** is not a good look for me, and is a useful reminder that a report is worth reading against my own prose, not only mined for citations.

I took a combination of your suggested fixes rather than one:

- **Narrowed** the enumeration to renal parenchymal disease, renovascular disease and OSA — the causes the mechanism actually fits.
- **Carried the snippet's hedge** into the description ("a common theme across several…").
- **Route-differentiated the exclusions explicitly** in a `notes:` block, naming why PA and coarctation are out and flagging pheochromocytoma as a partial fit at best, since catecholamine release is not under normal neural control.

Your observation that the shared endpoint is the *phenotype* is the one that made the fix obvious: the entry already captures convergence correctly through the `downstream: Hypertension` edge, so the mechanism node never needed to carry it. The `notes:` says so.

## IMPORTANT 2 — bundling and grounding

Agreed on both counts. Dropped *"and blood pressure rises"* — the `downstream` edge says it, so the prose was duplicating the edge. Bound `GO:0003073` (already in `cache/go/terms.csv`, validates offline). I did **not** bind `GO:0035813`, since you flagged it as unverified in the report and I couldn't confirm it independently.

Your point that promoting the node to `central_effector` raises the bar is the right frame — it was tolerable as a disconnected entry and isn't as the hub the whole non-drug side passes through.

## IMPORTANT 3 — treatment join point, with one deliberate deviation

Both first-line drugs added with the module's bindings and the cached snippet, following the pattern its `notes` describe.

**One deviation worth your eye:** I attached them to `Drug-Induced Vasoconstriction and Increased Peripheral Vascular Resistance` — the VEGF arm's convergence node — and **not** to the `Sympathetic and hormonal coactivation` node. The evidence (PMID:29931399) is specifically about tyrosine-kinase-inhibitor hypertension, so attaching it to the generic non-pharmacological convergence node would extend the claim past what the source says. That is the same overreach IMPORTANT 1 asked me to remove, and it would have been odd to fix it in one place and commit it in another. The reasoning is recorded in the treatments' `notes` so it isn't re-litigated.

If you read the module's *"the antihypertensive targets the convergence node"* as meaning this entry's non-drug hub specifically, say so and I'll revisit — but I think the drug arm is the one the evidence supports.

Gates after the fixes: `just validate-disorders` passed, `count-verified-snippets` 14/14, `check-duplicate-keys` clean, graph 10 edges with no integrity issues and both treatments now joined.

caufieldjh:
Follow-ups done in `721ce34`. Thanks for the approval, and for flagging the description — that one was a real contradiction I created.

**The description contradiction.** You're right, and it's worse than a leftover: I removed the "shared final pathway" claim from the node, wrote a `notes` block explicitly refuting it, and left the identical claim standing in the disease-level `description` six lines up. The entry contradicted itself in two places a reader would naturally compare. Fixed — the description now says these causes share an *endpoint* rather than a mechanism, and names the three routes that don't fit (renin-independent MR signalling in PA, mechanical obstruction in coarctation, endothelial vasoconstriction in the VEGF arm).

**`biological_scale` drift.** Changed `MOLECULAR` → `CELLULAR` to match the module node. You read it correctly as drift rather than a decision — the PR body said bindings were copied from the module, so the divergence was me not copying that field. Aligning is the honest resolution rather than defending it after the fact.

**The lump/split question — recorded, not resolved.** Added a `discussions` entry (`scope_secondary_hypertension_grouping_vs_disease`, `KNOWLEDGE_GAP`, OPEN) attached to the sympathetic node. It records the argument, that most causes already exist as their own entries, and the imbalance this PR leaves: five of six pathophysiology nodes now describe the drug cause — a minority of secondary hypertension — while the commoner renal causes share one generic node. Defensible, since the drug arm is the one with a module to conform to and no entry of its own, but a decision rather than an oversight.

I'd rather not route it to @cmungall from here: a Grouping conversion would touch several existing entries, and the discussion entry means the next person to pick it up starts with the reasoning rather than rediscovering it. If you think it warrants a ruling sooner, say so and I'll open it as its own issue.

**`links.prs`** now carries 10362.

**On the deep-research cross-check** — the distinction you drew is the useful part, and I'd have got it wrong left to myself. The 6-chains-to-1-modelled ratio does look like blocking under-consumption until you notice the other five chains have their own entries, at which point importing them would duplicate mechanism rather than add coverage. Noted too that two probed PMIDs in that report came back as unrelated papers; consistent with what I've seen across this batch, and the reason nothing in these PRs is curated from a report without independent verification first.

`stubs/Primary_Aldosteronism.yaml` as the next adjacent item is a good call, and squarely the thing the discussion entry points at.

Gates after the follow-ups: `just validate-disorders` passed, `count-verified-snippets` 14/14, `check-duplicate-keys` and `check-entity-refs` clean, `validate-history` clean.

Issue #8759, opened 2026-08-17
Curate glycogen storage disease due to acid maltase deficiency, late-onset (MONDO:0018485)


sierra-moxon:
Curate a dismech entry for **glycogen storage disease due to acid maltase deficiency, late-onset** ([MONDO:0018485](https://monarchinitiative.org/MONDO:0018485); OMIM:621314; Orphanet:420429).

> Glycogen storage disease due to acid maltase deficiency, late onset (AMDL), a form of Glycogen storage disease due to acid maltase deficiency (AMD), a degenerative metabolic myopathy particularly affecting respiratory and skeletal muscles, is characterized by an accumulation of glycogen in lysosomes.

**Coverage check** (`origin/main`, `kb/`): MONDO:0018485 is not bound as a `disease_term`, subtype term, or MONDO mapping by any existing entry, and no open or closed curation issue covers it.

Related existing entry: `kb/disorders/Pompe_Disease.yaml` **already carries an `LOPD` subtype with no bound `disease_term`**. As with the infantile form, decide binding-the-subtype vs standalone entry first. The late-onset form's distinct respiratory/limb-girdle myopathy course and ERT response are the mechanistic content that would justify a separate entry.

Tracker: part of #1079.

sierra-moxon:
### Subtype audit: partially curated, and split along one axis only

`kb/disorders/Pompe_Disease.yaml` (MONDO:0009290) carries `IOPD` and `LOPD` under `has_subtypes` with **no bound `disease_term`**. Unlike the Loeys-Dietz stubs, these are genuinely doing some work — but only in one layer:

| Subtype stub | Blocks referencing it via `subtype:` |
|---|---|
| IOPD | 7 (all `phenotypes`) |
| LOPD | 2 (all `phenotypes`) |

9 of 17 phenotypes are subtype-scoped. Everything else is **unscoped and shared**: the single `GAA` genetic block, the one prevalence record, all 4 pathophysiology nodes, and all 6 treatments — including enzyme replacement therapy and cipaglucosidase alfa + miglustat.

So the parent splits the *clinical phenotype* layer and shares the rest. The mechanistic content that actually separates the two forms is not curated anywhere: residual GAA activity and the genotype-activity threshold, CRIM status and anti-rhGAA antibody response (which drives immunomodulation and ERT outcome in IOPD), and cardiac- versus respiratory/limb-girdle-dominant course.

### Two acceptable resolutions — binding alone is not one of them

[...]

PR #10363, opened 2026-09-01
Late-Onset Pompe Disease: add the smooth-muscle and vascular arm

tannerzhang:
Late-onset Pompe disease is not a pure myopathy. Patients develop dilatation of the aorta and the brain arteries, plus reflux, constipation and urinary incontinence, and none of that follows from the limb weakness. This PR adds the mechanism behind those symptoms — lysosomal glycogen accumulating in smooth muscle — to the existing LOPD entry, together with the seven phenotypes it explains and the mouse model that both supports and complicates it.

The requested disorder, "glycogen storage disease due to acid maltase deficiency, late-onset", is the MONDO:0018485 label already curated in `kb/disorders/Late-Onset_Pompe_Disease.yaml` (issue #8759). This augments that entry rather than creating a second one.

## The gap this closes

The previous curation left an explicit marker on the aortic dilatation phenotype:

> Glycogen storage in vascular smooth muscle is the proposed substrate, but the mechanism is not established here and is deliberately not asserted.

There is now direct human evidence for it. An autopsy study in LOPD (PMID:22664150) used transmission electron microscopy to show lysosomal glycogen inside vascular smooth muscle cells, and separately in bladder, intestine and esophagus.

## What was added

Two pathophysiology nodes, wired into the existing chain:

```
Leaky GAA splicing (partial residual enzyme)
  └─ Lysosomal glycogen accumulation in vascular       [CELLULAR, ESTABLISHED]
     and visceral smooth muscle
       ├─ Dilative arteriopathy of the aorta and       [TISSUE, PROVISIONAL]
       │  intracranial arteries
       │    ├─ dilatation of the ascending thoracic aorta  (existing phenotype)
       │    ├─ vertebrobasilar dolichoectasia
       │    ├─ dilatation of the intracranial arteries
       │    ├─ stroke
       │    └─ cerebral white matter hyperintensities
       ├─ gastroesophageal reflux
       ├─ constipation
       └─ urinary incontinence
```

Seven new phenotypes, each frequency band backed by a quoted denominator.

A new `animal_models` section for the Gaa-/- B6;129 global knockout (PMID:33883348), carrying two links: `RECAPITULATES` the smooth-muscle storage node with two readouts, and `FAILS_TO_RECAPITULATE` the arteriopathy node.

## Two judgment calls worth a look

**The arteriopathy node is PROVISIONAL, not ESTABLISHED.** The storage lesion is well supported; the step from it to arterial wall weakness is not. One resected aneurysmal aorta showed connective-tissue changes resembling a fibrillinopathy rather than a storage lesion (PMID:22945242), and in the mouse the aorta held the least glycogen of any tissue measured and gained none between 6 and 15 months. Those authors state plainly that this "does not corroborate our hypothesis that glycogen accumulation is responsible for aortic wall weakness."

**That negative result is curated rather than dropped.** It is the `FAILS_TO_RECAPITULATE` link, and it records the limit of what the study showed: it measured glycogen and wall histology, not vessel calibre, so the claim is failure to reproduce the storage lesion, not a measured absence of dilatation. Both model links also record that a global null has no residual enzyme, whereas late-onset disease is defined by partial activity from a leaky splice allele — so this mouse is genotypically closer to infantile-onset disease and its tempo cannot be read across.

## Frequencies come from one cohort

All four cerebrovascular frequencies are from a single 30-patient Chinese imaging cohort (PMID:39010129): dolichoectasia 17/30, stroke 7/30, white matter hyperintensities 13/28. The entry's `notes` says so, rather than presenting them as pooled estimates.

Cerebellar and brainstem microbleeds (7/22) are recorded on the arteriopathy node rather than as their own phenotype, because HPO has no cerebral microbleed or microhemorrhage term and `HP:0001342` *Cerebral hemorrhage* names a clinical event rather than an imaging finding. The phenotype carries a note saying so.

## Deep research

Falcon returned 403 (no API key in this environment), so the run used `--fallback` and completed as OpenScientist, which records `fell_back: true` and the full `provider_attempts` trail. A second run with Claude Code independently surfaced the same vasculopathy papers that a direct PubMed search had found; that agreement is what this augmentation rests on. Both reports are included.

Both reports are `needs_review: true`, so the 0% confabulation rate is not an all-clear. OpenScientist has one mislabelled term; Claude Code has six plus one unsupported quote. None of those terms or quotes are used here. Worth noting one specifically: the Claude Code report offered `CL:0000192` for *vascular* smooth muscle, which CL actually calls plain "smooth muscle cell", so the entry binds the OAK-verified `CL:0000359` instead.

## Validation

Figures below are for the final commit (`fc395e4`). `just validate` and `just validate-terms` pass. `just count-verified-snippets --unskip-prefix DOI` reports 159/159; the plain run is 158/159, the one skipped item being a pre-existing DOI-keyed snippet that CI does not check and that was therefore checked by hand.

Clean with no new baseline entries: entity-refs, causal-targets, duplicate-keys, enum-values, qualifier-terms, folded-hyphens, snippet-length, title-snippets, snippet-grading. The KB foreign-key pytest lane passes (13231 passed).

Compliance rose from 98.5% (256/260) to 98.8% (331/335), and phenotype connectivity from 10/18 to 11/18 — every phenotype this PR adds is on the graph, and the seven that remain unconnected are all pre-existing. The four remaining compliance gaps are likewise pre-existing: two datasets without evidence, and one proposed experiment.

## Review rounds

Two rounds of review findings were addressed in `ec9de17` and `fc395e4`: organ-level cell-type bindings (`CL:0002597`, `CL:0002599`, `CL:0002504`), a human-sourced quote on the reflux edge, and a missing pathograph edge to `Constipation`. One suggestion was declined after checking — splitting microbleeds onto an HPO microhemorrhage term, which does not exist — and the reasoning is recorded in the entry's `notes` rather than left in the thread.

## Housekeeping

The `--fallback` run renamed the falcon output directory on its way to OpenScientist, which deleted the previously committed falcon report and left a copy of its artifact in the OpenScientist folder. The falcon files were restored from git and the stray copy removed, so the only research additions here are the two new reports. Two literature caches (PMID_34800399, PMID_39424261) were rewritten as formatting-only churn by the research validators and were reverted.

https://claude.ai/code/session_015diEdsPXagfxL1GL82ecKh

### DIFF
--- a/history/disorders/Late-Onset_Pompe_Disease/2026-09-01T110707Z-claude-code-62a621.yaml
+++ b/history/disorders/Late-Onset_Pompe_Disease/2026-09-01T110707Z-claude-code-62a621.yaml
@@ -62,3 +62,26 @@ events:
     fifth was informational about diff size. After these changes all gates were re-run: validate,
     validate-terms, entity-refs, causal-targets, duplicate-keys, enum-values and qualifier-terms are
     clean, and count-verified-snippets --unskip-prefix DOI reports 158/158.
+
+    Second review round (PR #10363). The re-review requested changes on two findings, both correct
+    and both mine. First, the note added in the previous round claimed CL offers only small- and
+    large-intestine smooth-muscle terms; it does not. CL:0002504 "enteric smooth muscle cell" is
+    defined as "A smooth muscle cell of the intestine", carries "intestinal smooth muscle cell" as
+    an exact synonym, is already in cache/cl/terms.csv and the celltypeterm enum cache, and is
+    already used in four other entries. The earlier OAK search missed it because the label contains
+    no form of the word "intestine". The binding is now CL:0002504 and the false claim is replaced
+    by a note that the segment-agnostic term was chosen over its children because the autopsy does
+    not specify segment. Second, the Constipation phenotype asserted in its description that it
+    shares the intestinal smooth-muscle mechanism but had no incoming pathograph edge, so it
+    rendered as a floating node while its two siblings were wired up; nothing gates an unconnected
+    phenotype, so no check caught it. An INDIRECT_UNKNOWN_INTERMEDIATES edge mirroring the reflux
+    one was added, carrying the same PMID:20559845 enzyme-replacement-response sentence. Two
+    advisory notes were also taken: the reflux edge explanation said "gastrointestinal symptoms"
+    where the quote says "intestinal symptoms", now corrected to track the quote and to say that the
+    paper defines those symptoms earlier as constipation and reflux; and the arteriopathy node
+    description, which described only the aorta and intracranial arteries, now acknowledges the
+    small-vessel component that the white-matter-hyperintensity and microbleed evidence covers.
+    After these changes: validate, validate-terms, entity-refs, causal-targets, duplicate-keys,
+    enum-values and qualifier-terms clean, count-verified-snippets --unskip-prefix DOI 159/159, and
+    compliance-connectivity for this entry rises from 10/18 to 11/18 phenotypes connected, with all
+    seven remaining unconnected phenotypes pre-existing and out of scope.
--- a/kb/disorders/Late-Onset_Pompe_Disease.yaml
+++ b/kb/disorders/Late-Onset_Pompe_Disease.yaml
@@ -760,12 +760,12 @@ pathophysiology:
       label: smooth muscle cell of the esophagus
   - preferred_term: intestinal smooth muscle cell
     description: >-
-      Bound at the generic smooth-muscle level because the cited autopsy names
-      "intestine" without specifying segment, and CL offers only small- and
-      large-intestine terms below this point.
+      The segment-agnostic term is chosen over its small- and large-intestine
+      children because the cited autopsy names "intestine" without specifying
+      segment.
     term:
-      id: CL:0000192
-      label: smooth muscle cell
+      id: CL:0002504
+      label: enteric smooth muscle cell
   biological_processes:
   - preferred_term: lysosomal glycogen degradation in smooth muscle
     modifier: DECREASED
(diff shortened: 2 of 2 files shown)


tannerzhang:
Thanks — pushed `ec9de172d` addressing three of the five notes. One I did not implement as suggested, because the term it needs does not exist; details below.

| Note | Action |
|---|---|
| 1. Microbleeds riding on the WMH phenotype | Fixed, but **not** by the suggested route — see below |
| 2. Visceral smooth muscle bound generically | Fixed — organ-specific CL terms exist, I had missed them |
| 3. Mouse paper carrying a human symptom claim | Fixed — human evidence added |
| 4. PR body mis-attributes the autopsy PMID | Already corrected before the review posted |
| 5. Diff size dominated by derived artifacts | No action; informational |

### Note 1 — the suggested fix isn't available

You're right that the microbleed snippet supported the node description rather than `HP:0030890`. But there is no HPO term to move it to. OAK returns nothing for either `microbleed` or `microhemorrhage`, and the nearest candidate, `HP:0001342` *Cerebral hemorrhage*, names a clinical event rather than a susceptibility-weighted imaging finding — binding it would assert that seven patients had cerebral hemorrhages, which is not what the paper reports.

So instead of a new phenotype, the microbleed evidence moved to the `Dilative Arteriopathy of the Aorta and Intracranial Arteries` node, where it supports the small-vessel half of the claim and needs no phenotype binding. The WMH phenotype carries a `notes` entry recording why it is not curated separately, so the next curator does not re-derive this.

If the view is that HPO should have a microbleed term, that is an upstream request rather than something to force here.

### Note 2 — you were right, I was wrong

CL does carry organ-specific smooth-muscle terms; my earlier search missed them by omitting "the". Bladder is now `CL:0002597`, esophagus `CL:0002599`.

The intestinal component stays at the generic `CL:0000192`. CL offers only `CL:1000275` (small intestine) and `CL:1000279` (large intestine) below that point, and the cited autopsy says "intestine" without specifying segment — picking one would manufacture a specificity the source does not have. The `description` on that `cell_types` entry says so.

### Note 3 — human evidence added

The reflux edge now also cites PMID:20559845: enzyme replacement reversed the gastrointestinal symptoms at one year while leaving cerebral vessel calibre unchanged. That is `directness: INDIRECT` support — the inference runs through treatment response rather than a tissue measurement — and it is a better fit than the mouse paper's background prose. The same sentence carries a useful contrast: the GI arm responds to enzyme and the vascular arm does not, which is why the arteriopathy node is not argued this way.

The mouse item stays alongside it rather than being replaced, since it is the paper that actually measured the storage.

### Note 4 — already fixed

The PR body was rewritten before this review posted; it now reads PMID:22664150 for the autopsy. The review was reading the original auto-generated body. The YAML always had it right.

### Validation after the changes

`validate`, `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values` and `check-qualifier-terms` are all clean, and `count-verified-snippets --unskip-prefix DOI` reports 158/158. Two new derived cache rows for the CL terms.

---

tannerzhang:
Both findings were right, and both were mine. Fixed in `fc395e44f`, along with both 🔵 notes.

**Finding 1 — `CL:0002504`.** You are correct, and the wrong binding was the smaller half of it. `CL:0002504` *enteric smooth muscle cell* is defined as "A smooth muscle cell of the intestine", carries "intestinal smooth muscle cell" as an exact synonym, is in both caches, and is used in four other entries. My OAK search missed it because the label contains no form of the word "intestine" — I searched `l~intestine` and filtered for smooth muscle, which cannot find it. The note I wrote off the back of that search would have told the next curator a term does not exist when it does, which is worse than the binding being a level too generic. Rebound, and the note now gives the real reason for preferring the segment-agnostic term over its small- and large-intestine children.

**Finding 2 — Constipation was floating.** Confirmed: it was the only newly added phenotype with no incoming edge, while both its siblings had one. Added an `INDIRECT_UNKNOWN_INTERMEDIATES` edge from the smooth-muscle node carrying the same PMID:20559845 enzyme-replacement-response sentence. You are right that "intestinal symptoms" fits constipation at least as well as reflux, since the paper's earlier sentence defines those symptoms as constipation plus reflux.

Worth noting for anyone reading later: neither of these was catchable by the gates. An unconnected phenotype is not gated, and a `description` asserting a mechanism the graph does not carry is prose. `just compliance-connectivity --list-unconnected` is what surfaces it, and it is not something I had been running.

**🔵 3** — the reflux explanation now tracks the quote's wording ("intestinal", not "gastrointestinal") and says explicitly that the paper defines those symptoms earlier as constipation and reflux.

**🔵 4** — the arteriopathy node description now names its small-vessel component, so the microbleed evidence item's placement is legible without scrolling to the edges.

Re-run after the changes: `validate`, `validate-terms`, `validate-history`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-grading`, `check-snippet-length` all clean, no new baseline entries. `count-verified-snippets --unskip-prefix DOI` is now 159/159. `compliance-connectivity` for this entry goes from 10/18 to 11/18 phenotypes connected — every phenotype this PR adds is now on the graph, and the seven that remain are all pre-existing.

No new cache rows, as you predicted.

---

PR #10366, opened 2026-09-01
enhance(Allopurinol_Induced_SJS_TEN): bind the exposure, add HLA screening and the SCAR differentials

caufieldjh:
## What

This entry's mechanism chain was already good — 6 nodes, 5 conforming to `drug_hypersensitivity_scar`. The gaps were about **honesty of binding** rather than missing mechanism, so that is what this PR addresses.

## Two unbound exposures, dispositioned differently on purpose

`just environmental-term-audit` flagged both environmental entries as `UNBOUND`. They are not the same kind of problem:

**Allopurinol exposure → bound** to `ECTO:0000509` (*exposure to drug*), with the specific agent carried in `preferred_term`.

**Renal impairment → deliberately left unbound**, with a `notes` line saying why. Reduced renal function is a *host susceptibility factor* modifying oxypurinol clearance — not an exposure. ECTO is an exposure ontology; binding a physiological state to it would misuse the ontology while conveniently silencing the audit. The entry sits under `environmental:` only so it can reach the pathograph via `influences_mechanisms`.

**The audit will keep reporting this one, and that is the correct outcome rather than a gap to close.** The note exists so a later curator doesn't "fix" it.

## HLA-B*58:01 pre-prescription screening

**Correction (post-review):** this section originally claimed the entry lacked the one pharmacogenomic test specific to this drug. That was wrong — `treatments[]` already contained an `HLA-B*58:01 pre-prescription screening` entry predating this PR. I checked `diagnosis[]` and not the sibling section, and so added a duplicate. The two are now **merged** into the single `diagnosis` entry, carrying all four evidence items, with the term upgraded to `NCIT:C71329` (HLA Typing). Diagnostic procedures belong in `diagnosis`, so the treatments copy was the misplaced one. Added, framed as **preventive rather than diagnostic**, since it identifies the susceptibility allele the trigger node depends on *before* exposure occurs.

The load-bearing detail is population dependence:

> *"HLA-B\*5801 was associated with allopurinol-induced TEN/SJS in European and Japanese populations, but only had a 50-60% sensitivity (pooled sensitivity 56%), compared to the 80-100% sensitivity (pooled sensitivity 97%) observed in Korean, Thai, Sardinia Italian and Han Chinese populations."* — PMID:28857441

A negative test means materially different things by ancestry, and the description says so rather than reporting a single pooled figure.

## DRESS and AGEP differentials

Both added with distinguishing features, and both **omit `disease_term`** — no verified MONDO CURIE was reachable at curation time, and a guessed id is worse than none. Each carries a `notes` line recording the deferral.

The AGEP description also records that its own boundary with pustular psoriasis is genetically contested (shared *IL36R* mutations), so the label isn't offered as a settled alternative diagnosis.

## Process note

`reference_title` values were written as **placeholders and generated programmatically from the cache**, not typed. This is the practice adopted after fabricating four titles earlier in this same session on a sibling PR; here it produced zero fabrications.

## Validation

- `just count-verified-snippets` — **44/44** verified
- `just check-duplicate-keys`, `just check-entity-refs` — passed
- `just environmental-term-audit` — allopurinol now bound; renal impairment intentionally still reported


caufieldjh:
Fixed in `be3b7b6` and `c0a9d1f`. All three blocking findings were correct, and the second one caught a **false statement in my own PR body and history record** — thank you for checking the sibling section rather than taking the claim at face value.

## The duplicate — my error, and a specific kind

I wrote that the entry "did not carry the one pharmacogenomic test specific to this drug". The *section* didn't. The *entry* did: `treatments[]` had `HLA-B*58:01 pre-prescription screening` at line 872, predating this work. I checked `diagnosis[]`, concluded absence, and never looked at the neighbouring section — so I added a second copy of an intervention the file already modelled, with four evidence items split across two places and nothing linking them.

Merged into the single `diagnosis` entry, since the SOP puts diagnostic procedures there and the treatments copy was the misplaced one. All four evidence items carried over, nothing dropped, and the term upgraded to the more specific `NCIT:C71329` (HLA Typing) as you suggested.

You were right that the merge is strictly more informative: the meta-analytic sensitivity figures now sit directly beside the Taiwan screen-and-avoid cohort, and explain *why* that programme succeeds — it ran in a population where sensitivity is ~97%, not ~56%. That connection didn't exist while the two halves were in different sections.

**I've corrected the history record in place** rather than leaving it asserting the original claim, and the PR body is updated too.

## Both differentials bound

`MONDO:0015340` and `MONDO:0017384`, deferral notes removed. Your framing is the right one — declining to guess was correct at the time, but once the binding is available, a note saying "add this later" is a worse artifact than the binding. Both CURIEs are committed to `cache/mondo/terms.csv` and the `DiseaseTerm` enum cache so CI validates offline.

## The one suggestion I couldn't take — with a finding for you

`ECTO:9002228` is the exact term and **does** resolve in OLS; I confirmed the label independently. It is not bindable in this repo today:

```
$ uv run runoak -i sqlite:obo:ecto info ECTO:9002228
ECTO:9002228 ! None
```

`conf/oak_config.yaml` maps `ECTO: sqlite:obo:ecto`, and that build carries the CURIE with no label — so `just validate-terms` fails with both *"not found in ontology"* and *"not in dynamic enum ExposureTerm"*. I tried the binding, hit exactly that, and reverted.

Worth knowing for future reviews: an ECTO term resolving in OLS is **not** sufficient for it to be bindable here, because ECTO is one of the prefixes validated against a local SQLite build rather than `ols:`. That's the opposite of the ARDS case on the sibling PR, where `ECTO:7000059` and `ECTO:9000077` both resolved locally and bound fine.

The entry stays on `ECTO:0000509` and now carries a `notes` line recording the gap, so the suggestion isn't made and reverted again by someone else later.

Gates: `just validate-disorders` passed, `count-verified-snippets` 44/44, `check-duplicate-keys` clean, `validate-history` clean.

PR #10367, opened 2026-09-01
chore(stubs): mark seven already-curated subtypes as SUBTYPE

gingin77:
Marks seven curation stubs `entry_type: SUBTYPE` and records in each one's `notes` which curated entry already covers it. No files are deleted and no `kb/` content changes.

Each of these diseases is already curated as a `has_subtypes` record on a parent disease, so anyone picking the stub up would rediscover an existing entry rather than curate anything new. A `SUBTYPE` stub is unreachable from `just next-stubs` (`src/dismech/stubs/cli.py:136` pools `UNDECIDED` and `DISEASE` only), so marking takes them out of the queue while keeping the reasoning in the repository.

| Stub | Gene | Already curated in |
|---|---|---|
| CMT type 4A (MONDO:0008961) | GDAP1 | `Charcot-Marie-Tooth_Disease_Type_4` |
| CMT type 4B2 (MONDO:0011475) | SBF2 | `Charcot-Marie-Tooth_Disease_Type_4` |
| CMT type 4C (MONDO:0011113) | SH3TC2 | `Charcot-Marie-Tooth_Disease_Type_4` |
| CMT type 4F (MONDO:0013959) | PRX | `Charcot-Marie-Tooth_Disease_Type_4` |
| CMT type 4J (MONDO:0012640) | FIG4 | `Charcot-Marie-Tooth_Disease_Type_4` |
| X-linked erythropoietic protoporphyria (MONDO:0010420) | ALAS2 | `Erythropoietic_Protoporphyria`, subtype `XLP` |
| methylmalonic aciduria, mut type (MONDO:0009612) | MMUT | `Methylmalonic_Acidemia` |

Verification was gene-by-gene against `origin/main`, not by name match: the stub's MONDO causal gene has to equal the parent entry's gene for that named subtype. CMT4A is the one exception — its stub carries no `genes:` field, so the mechanical check does not apply and the parent's own `CMT4A (GDAP1-related)` subtype description was checked by hand instead. Its `notes` says so.

With CMT4A included, the CMT4 family is now consistent: 4A, 4B2, 4C, 4D, 4F and 4J are all marked (4D already was, on `main`), and 4B3 and 4G stay `UNDECIDED` because the parent does not list them.

## Why marking rather than deleting

The first version of this PR deleted the stubs. That was wrong, and the review caught it. Deletion does not stick:

`seed_stubs` (`src/dismech/stubs/seed.py:313`) skips a nomination only on a coverage-index hit or an existing stub file. Running `build_coverage_index()` against all seven MONDO IDs returns `None` for every one — a disease curated as a `has_subtypes` record has no `disease_term`, so the index cannot see it. With the stub file also gone, the next `just seed-stubs` re-adds them as `UNDECIDED` and the analysis goes with them.

Marking achieves the same goal durably, and matches the convention already on `main`: `stubs/Charcot-Marie-Tooth_Disease_Type_4D.yaml` is the same situation on the same parent, handled this way. The new `notes` mirror its wording.

## One case is weaker than the others

`MONDO:0009612` names the **enzyme-locus** axis (mut type = MMUT), whereas `Methylmalonic_Acidemia`'s subtypes are "Vitamin B12-responsive" and "Vitamin B12-unresponsive" — a **treatment-response** axis that spans cblA/cblB as well as mut. So there is no subtype record it maps to one-to-one.

It is still marked `SUBTYPE`, because the concept genuinely is covered: that entry names MMUT deficiency as a primary cause in its description and curates MMUT in its `MMUT/AdoCbl pathway molecular function deficiency` node. But the `notes` record the axis mismatch explicitly, and say that if the mut/cbl locus distinction is later wanted as an explicit subtype, that is an edit to the parent entry rather than a new disorder file.

## How these were found, and a caution about finding more

They surfaced while running `/claim-disease`, where three came up in a single pool of twenty. That suggested the pattern was more common, so I scanned the whole queue for stubs whose label or synonym exactly matches a curated `has_subtypes` name.

That scan returns 49 hits and **most are wrong**. It matches strings, and subtype names are often acronyms that collide across unrelated fields:

- `XLP` matches X-linked **lymphoproliferative** disease against X-linked **protoporphyria**
- `AIP` matches **autoimmune pancreatitis** against **acute interstitial pneumonia**
- `MACS` matches **mast cell activation syndrome** against **mild autonomous cortisol secretion**
- `MAR` matches **MDS with isolated del(5q)** against **melanoma-associated retinopathy**
- `AMC` and `LMS` collide the same way

So it is a lead generator, not a worklist. Only the seven above are here, each verified individually. Working through the other candidates is a separate triage task where the gene check decides rather than the name.

## Known limitation, out of scope here

Neither parent carries `subtype_term` on the relevant `has_subtypes` records, nor a `mappings.mondo_mappings` `skos:narrowMatch` to these MONDO IDs. That is why `build_coverage_index` cannot see them and why `just tidy-stubs` will never surface this class of stale stub. Either slot would fix detection for the whole class rather than these seven, and it deserves its own PR — it is a `kb/` change across several entries plus a coverage-index question, not a stub-queue edit.

## Validation

`just check-stubs` — 1363 stubs, no errors, 41 advisories, none about these files. (The `possible_kb_duplicate` advisory on the methylmalonic stub is the tool agreeing with this PR.) `just validate-stubs` — no issues. The `next-stubs` pool drops to 1349 with all seven absent.

### DIFF
--- a/stubs/Charcot-Marie-Tooth_Disease_Type_4A.yaml
+++ b/stubs/Charcot-Marie-Tooth_Disease_Type_4A.yaml
@@ -2,7 +2,7 @@ mondo_id: MONDO:0008961
 label: Charcot-Marie-Tooth disease type 4A
 proposed_name: Charcot-Marie-Tooth_Disease_Type_4A
 status: OPEN
-entry_type: UNDECIDED
+entry_type: SUBTYPE
 priority: NORMAL
 rationale: Diagnostic delay impact; Insufficient ICD coding/underdiagnosis; Intervention potential; Multi-modal detectability; Phenotypic complexity
 synonyms:
@@ -27,3 +27,14 @@ mondo_parents:
   label: Charcot-Marie-Tooth disease recessive intermediate A
 - id: MONDO:0018995
   label: Charcot-Marie-Tooth disease type 4
+notes: >-
+  SUBTYPE of Charcot-Marie-Tooth Disease Type 4 (kb/disorders/Charcot-Marie-Tooth_Disease_Type_4.yaml),
+  which carries CMT4A ("CMT4A (GDAP1-related)") in its has_subtypes list alongside CMT4B1,
+  4B2, 4C, 4D, 4F and 4J, names GDAP1 in its genetic block, and curates a "GDAP1
+  Mitochondrial Dynamics Defect" pathophysiology node. No separate disorder entry is needed;
+  further CMT4A detail should be added to that entry's subtype record rather than to a new
+  file.
+
+  Note this stub carries no genes: field, so the gene-equality check used for the other CMT4
+  stubs does not apply mechanically here; the parent's own CMT4A subtype description names
+  GDAP1, which is the same check made by hand.


gingin77:
You're right, and I've switched to marking. I checked each of the three code claims rather than taking them on trust, and all three hold.

**Deletion doesn't stick.** `seed_stubs` (`src/dismech/stubs/seed.py:313`) skips a nomination only on a coverage-index hit or an existing stub file. I ran `build_coverage_index()` against all six MONDO IDs and it returns `None` for every one — a disease curated as a `has_subtypes` record has no `disease_term`, so the index can't see it. With the stub file also gone, the next `just seed-stubs` re-adds all six as `UNDECIDED`, and the gene-by-gene analysis goes with it. That's the argument that settles it.

**Marking already does the job.** `next_command` (`src/dismech/stubs/cli.py:136`) pools only `UNDECIDED` and `DISEASE`. Confirmed after the change: the pool drops from 1356 to 1350 and none of the six is in it.

**The CMT4D precedent is real.** `stubs/Charcot-Marie-Tooth_Disease_Type_4D.yaml` is on `main` with `entry_type: SUBTYPE` and a `notes` block naming the same parent entry. Deleting the other four CMT4 stubs would have split that family's seven subtypes across two conventions for no reason. The new notes mirror its wording.

So the six are restored with `entry_type: SUBTYPE` and `notes` naming the parent and the specific gene/pathophysiology content that covers them.

## On the methylmalonic case

You're right that it's weaker than the other five, and I'd noticed the same thing and then glossed it in the PR body — thanks for making me say it out loud. `MONDO:0009612` names the **enzyme-locus** axis (mut type = MMUT), whereas that entry's subtypes are "Vitamin B12-responsive" and "Vitamin B12-unresponsive", a **treatment-response** axis that spans cblA/cblB as well as mut. There is no subtype record it maps to one-to-one.

I've still marked it `SUBTYPE`, because the concept genuinely is covered — `Methylmalonic_Acidemia` names MMUT deficiency as a primary cause in its description and curates MMUT in its `MMUT/AdoCbl pathway molecular function deficiency` node — but the `notes` now records the axis mismatch explicitly and says that if the mut/cbl locus distinction is later wanted as an explicit subtype, that's an edit to the parent entry rather than a new disorder file. Someone re-reading this stub gets the caveat, not a false clean match.

## The durable fix

Agreed that adding `subtype_term` to the parent entries' `has_subtypes` records is the real answer — it would make `build_coverage_index` see subtype-curated diseases, which would fix `tidy-stubs`, fix re-seeding, and make this whole class of stale stub detectable instead of hand-found. I've deliberately not done it here: it's a `kb/` change across several entries plus a coverage-index question, and it belongs in its own PR rather than riding along with a stub-queue edit.

## Validation

`just check-stubs` — 1363 stubs, no errors, 41 advisories, none about these files. `just validate-stubs` — no issues. `next-stubs` pool verified at 1350 with all six absent.

gingin77:
Fixed the title and body, and folded CMT4A in.

**Title and body.** You're right that this matters more than it looks: these PRs are squash-merged, so the title becomes the permanent commit subject on `main`, and landing "remove six stubs" on a commit that removes nothing would undercut the durability argument that motivated the switch to marking. Retitled to `chore(stubs): mark seven already-curated subtypes as SUBTYPE`, rewrote the opening, and corrected the stale queue figure. The body now also carries a short section on why marking replaced deletion, so the reasoning survives in the merge commit rather than only in this thread.

**CMT4A — folded in.** Verified: the parent carries `CMT4A (GDAP1-related)` in `has_subtypes`, names GDAP1 in `genetic[]`, and curates a `GDAP1 Mitochondrial Dynamics Defect` pathophysiology node. No open claim or PR. As you noted, its stub has no `genes:` field so the gene-equality check doesn't apply mechanically — its `notes` says exactly that, and records that the parent's own subtype description naming GDAP1 is the same check made by hand, so the next reader knows this one was verified differently from the other five.

The CMT4 family is now consistent: 4A, 4B2, 4C, 4D, 4F, 4J all marked, with 4B3 and 4G left `UNDECIDED` because the parent doesn't list them. There's no 4B1 stub.

**The `subtype_term` gap.** Agreed it's the real fix and agreed it shouldn't live in a PR thread. I've described it in the body under a "Known limitation, out of scope here" heading so it survives the merge, but I haven't filed the issue — deciding whether the fix is `subtype_term` on the `has_subtypes` records or a `mappings.mondo_mappings` `skos:narrowMatch`, and what that implies for `build_coverage_index`, is a modeling call for a maintainer rather than something I should pre-commit in an issue title. Happy to file it if you'd like it tracked now.

## Validation

`just check-stubs` — 1363 stubs, no errors, 41 advisories, none about these files. `just validate-stubs` — no issues. `next-stubs` pool is 1349 with all seven absent.

PR #10376, opened 2026-09-01
curate(Benign_Paroxysmal_Positional_Vertigo): otoconial displacement as a biomechanical pathograph (MONDO:8000018)

### LINKED ISSUE #10369: Curate benign paroxysmal positional vertigo (MONDO:8000018)
Curate a dismech entry for **benign paroxysmal positional vertigo** ([MONDO:8000018](https://monarchinitiative.org/MONDO:8000018)) — the commonest cause of recurrent positional vertigo, attributed to utricular otoconia displaced into a semicircular canal so that head position change produces an inappropriate endolymph flow and a false angular-velocity signal.

Prior art check: BPPV is currently named only as a differential in `Labyrinthitis.yaml` and `Semicircular_Canal_Dehiscence_Syndrome.yaml`; there is no entry.

Likely subtypes, all sharing one otoconial mechanism and differing by which canal is involved and whether the debris is free-floating or adherent — these look like `has_subtypes` entries rather than separate diseases: posterior canal (canalithiasis, by far the commonest), horizontal/lateral canal (geotropic and apogeotropic variants), anterior canal, and cupulolithiasis.

Stub: `stubs/Benign_Paroxysmal_Positional_Vertigo.yaml`

matentzn:
BPPV is the commonest cause of vertigo and had no dismech entry. It was named only as a differential in `Labyrinthitis.yaml` and `Semicircular_Canal_Dehiscence_Syndrome.yaml`.

Closes #10369

## What the entry says

The mechanism here is mechanical rather than molecular, and the pathograph is built to say that plainly:

```
otoconial degeneration and detachment from the utricular macula
  → otoconial debris entry into a semicircular canal
      → canalithiasis (free-floating)  ─┐
      → cupulolithiasis (cupula-bound) ─┴→ aberrant cupular deflection and false angular velocity signal
                                              → vestibulo-ocular reflex driven positional nystagmus
                                              → vestibular-visual-somatosensory mismatch
```

Splitting canalithiasis from cupulolithiasis is the part worth reviewing. They are usually written up as two "theories", but they make different, testable predictions that the entry records separately — latency and fatigability follow from a free particle mass, persistence without latency follows from a loaded cupula — and different treatments target them (repositioning vs the Semont liberatory manoeuvre). Both converge on one node, which is where a misplaced particle becomes a neural signal.

Three canal variants are `has_subtypes` rather than separate entries: they share the whole mechanism and differ only in which canal received the debris.

## Judgement calls I would like checked

- **Osteoporosis has no `influences_mechanisms` link, on purpose.** The OR of 2.49 is solid, but no source I had says which mechanistic step osteoporosis acts on. Adding a link would have put a claim in the pathograph that nothing supports, so the reasoning sits in `review_notes` instead. Head trauma and vitamin D deficiency *are* linked, because their sources do name the step.
- **Two ontology abstentions.** The ampullary cupula is central to the mechanism and has no UBERON term (searching "cupula" returns *apex of cochlea* and *cupular organ*, neither of which is the structure), so it is described rather than bound. Head trauma has no ECTO exposure term; the NCIT head-trauma concepts are injury terms, not exposure terms.
- **Repositioning manoeuvres are bound to `NCIT:C15302` Physical Therapy** because NCIT has no canalith-repositioning or vestibular-rehabilitation term at all. The specificity is carried in `preferred_term`.
- **`treatment_effect: BYPASSES`** for the manoeuvres. They change nothing about otoconial biology; they move the particle so the mechanism has no substrate. That is also why they do not prevent recurrence, and it seemed worth encoding rather than defaulting to `INHIBITS`.

Perplexity `sonar-deep-research`, 294s, 21 citations, committed under `research/`. It was useful for structure and useless as a source of identifiers:

- Its citation list resolved to web pages — Mayo Clinic, Cleveland Clinic, StatPearls, NORD, an OMIM mirror — not to PMIDs. Every claim kept here was re-anchored by resolving the PMC/PubMed URLs through the NCBI ID converter, fetching with `just fetch-reference`, and quoting the fetched text.
- The report carried **no** `reference_validation` frontmatter and no `## Reference Validation` section, so there was no confabulation rate to read before using it. (The sibling SCA8 run in this same batch did produce one, so this looks like a per-run gap rather than a missing feature.)
- Three of its ontology CURIEs were wrong: calcium carbonate as `CHEBI:3310` rather than `CHEBI:3311`, MONDO written as `MONDO:0008000018`, and Dizziness given `HP:0002321`, which is Vertigo. Every binding here was re-resolved independently through OAK.

## Validation

`just validate` clean — schema, terms, and **46/46 snippets verified** against the cached references. Also clean: `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

---

### DIFF
--- a/kb/disorders/Benign_Paroxysmal_Positional_Vertigo.yaml
+++ b/kb/disorders/Benign_Paroxysmal_Positional_Vertigo.yaml
@@ -1049,12 +1049,15 @@ prevalence:
     explanation: >-
       Source of the incidence range recorded here.
 - population: Patients presenting with vertigo
-  measure_type: PERIOD_PREVALENCE
+  measure_type: UNKNOWN
   prevalence_class: COMMON
   notes: >-
-    Proportion of vertigo presentations ultimately diagnosed as BPPV, which is a
-    clinic denominator rather than a population one and so is not comparable with
-    the population rates above.
+    Proportion of vertigo presentations ultimately diagnosed as BPPV. This is a
+    clinic denominator rather than a population one and is not comparable with
+    the population rates above. PrevalenceMeasureEnum has no value for a
+    proportion-of-cases figure, so measure_type is UNKNOWN rather than
+    PERIOD_PREVALENCE, which would assert a population denominator this number
+    does not have.
   evidence:
   - reference: PMID:30828628
     reference_title: Benign paroxysmal positional vertigo.


matentzn:
All six findings addressed in d6cc0aa54. Both blocking ones were fair — and the migraine gap in particular was a real inconsistency, not a matter of taste: I modelled two rows of that risk-factor table as `environmental` entries and skipped the row with the largest effect size.

**Blocking**

1. **Migraine** added as an `environmental` entry on the Osteoporosis pattern, quoting `migraine (OR = 4.40; 95% CI, 2.67-7.25; P < 0.00001)`. No `influences_mechanisms` link, with a `review_notes` giving the same reason as Osteoporosis: the meta-analysis establishes the association and does not name the step. The description says explicitly that this is a different claim from vestibular migraine as a mimic, which the entry already carried.
2. **`UBERON:0004721` crista ampullaris** added to the transduction node's `locations`. You are right that the node described stereocilia in the crista while binding only Scarpa's ganglion. I kept the ganglion, since the node's second sentence is about afferent firing rate.

**Suggestions — all four taken**

3. **`clinical_burden` added**, quoting the stroke/dementia/fracture sentence. That is a better anchor than what I had, which was one `INDIRECT` snippet about falls being "likely".
4. **Total cholesterol modelled too.** You called it reasonable to skip; on reflection the asymmetry was the problem rather than the effect size, so the entry now covers every factor that reached significance in that table. Its description says plainly that it is the weakest of them and should be read as an association awaiting explanation.
5. **Mixed-source snippet replaced.** The quote now reads `The otoconia contained in the utricle and saccule have also been shown to undergo morphological changes and degeneration during a human's lifespan as observed in postmortem analyses` — same paragraph, same claim, single source type. I did not simply re-grade it, because the sentence really did span both.
6. **History `links` filled** with #10369 and this PR. I applied the same fix to the other four records in this run rather than only here.

**On your NCIT table** — thank you for re-resolving those. I checked the report's HP/MONDO/CHEBI suggestions and stopped there, so I had not seen that all four of its NCIT treatment CURIEs resolve to unrelated concepts (a CAR-T product, a bead, preemptive analgesia, a prostate biopsy). That is a sharper illustration of the point than the three errors I documented, and I have carried it into the run's notes.

Re-validated after the changes: `just validate` clean with **49/49 snippets verified**, plus `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-environmental-evidence`. The cache diff adds one row, `UBERON:0004721`.

---

PR #10382, opened 2026-09-01
curate(Spinocerebellar_Ataxia_Type_8): three toxic products of one repeat, kept as parallel hypotheses (MONDO:0012116)

### LINKED ISSUE #10370: Curate spinocerebellar ataxia type 8 (MONDO:0012116)
Curate a dismech entry for **spinocerebellar ataxia type 8** ([MONDO:0012116](https://monarchinitiative.org/MONDO:0012116)) — a slowly progressive dominant cerebellar ataxia caused by a bidirectionally transcribed CTG·CAG repeat expansion at the *ATXN8OS*/*ATXN8* locus, which is unusual among the SCAs in being argued to act through several mechanisms at once rather than one.

Prior art check: SCA8 is named in passing in `Autosomal_Dominant_Cerebellar_Ataxia_Type_I.yaml` as an ADCA-I member and as an example of a non-coding expansion, but has no entry of its own. Several other individual SCAs (SCA2, SCA17, SCA23, SCA27B, SCA43) do have their own entries, so a dedicated SCA8 entry follows existing convention.

The mechanistic interest here is that the locus is transcribed in both directions, and the reduced-penetrance/large-normal-allele picture makes the genotype-phenotype relationship notably looser than in the polyglutamine SCAs — both are things the entry should represent honestly rather than flatten.

Stub: `stubs/Spinocerebellar_Ataxia_Type_8.yaml`

matentzn:
SCA8 was named only in passing in `Autosomal_Dominant_Cerebellar_Ataxia_Type_I.yaml` and had no entry of its own, although SCA2, SCA17, SCA23, SCA27B and SCA43 all do.

Closes #10370

## The structural decision

One CTG·CAG expansion at 13q21 sits where ATXN8OS and ATXN8 overlap and is transcribed in both directions. That single lesion is read out three chemically different ways:

| Product | Direction / frame | Where the damage lands |
|---|---|---|
| CUG non-coding RNA | ATXN8OS | nuclear foci, MBNL1 sequestration, missplicing |
| polyglutamine protein | ATXN8, AUG-initiated | intranuclear inclusions, Purkinje and brainstem |
| polyAla / polySer | RAN, no AUG | white matter, demyelination and axonal loss |

These are modelled as **three parallel `mechanistic_hypotheses`**, not one chain, with `downstream` edges opting into the matching `hypothesis_groups`. RNA gain-of-function and polyglutamine toxicity are `CANONICAL`; RAN translation is `EMERGING`, on the authors' own sentence that "the mechanisms and role of RAN translation in disease are poorly understood."

Merging them into a single ranked chain would assert a causal weighting the literature does not support. A `HUMAN_MODEL_MISMATCH` discussion records that weighting as the open question: all three arms are present in human tissue, so this is not a mouse-only mechanism, but the two interventional results that separate them — the *Mbnl1* genetic interaction and the eIF3F knockdown — are in mice and cells. In humans the evidence is co-localisation of aggregate with damage, which cannot separate cause from correlate.

## Other judgement calls

- **MBNL1 is listed under `genetic` as `MODIFIER`** on model-organism evidence only, with a `review_notes` saying so explicitly. It is not a human modifier gene and the entry does not claim it is.
- **ATXN8OS and ATXN8 are both listed as `CAUSATIVE`** even though they cannot be inherited or tested separately — they are one physical expansion. They are listed separately because they carry different mechanisms, and the `notes` on ATXN8 says that.
- **Reduced penetrance is treated as a mechanism-level fact, not a caveat.** It is why a positive genetic test does not settle the diagnosis, and a `KNOWLEDGE_GAP` discussion records that repeat size explains neither penetrance nor onset age here, unlike in the polyglutamine SCAs.

Perplexity `sonar-deep-research`, 21 citations, committed under `research/`. Unlike the BPPV run in the same batch, **this one did carry `reference_validation` and `term_validation` frontmatter, and reading it first was the decisive step**:

- `term_validation` reported `needs_review: true` with **19 of 43 checked labels naming a different term than the report claimed**. None of its CURIEs were reused.
- Three of those errors would have been silently wrong here: `HP:0001270` offered as Dysarthria is **Motor delay** (correct: `HP:0001260`); `HP:0002067` offered as pyramidal signs is **Bradykinesia**; and both gene identifiers were wrong — ATXN8 given as `HGNC:13813` (correct `hgnc:32925`), ATXN8OS given as `HGNC:10610` (correct `hgnc:10561`).
- Only **8 of 21 citations** resolved to identifiers at all, and 5 of those 8 were judged on topic. The rest were web pages, so PMIDs were recovered from the report's PMC URLs and from PubMed title searches, then fetched and quoted from the cache.

Worth noting for the repo rather than for this PR: `CLAUDE.md`'s mechanical-backfill table spells `NCIT:C159273` as "speech therapy", but the canonical label is **Speech Language Therapy**. Term validation caught it here.

## Validation

`just validate` clean — schema, terms, and **46/46 snippets verified**. Also clean: `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

---

### DIFF
--- a/kb/disorders/Spinocerebellar_Ataxia_Type_8.yaml
+++ b/kb/disorders/Spinocerebellar_Ataxia_Type_8.yaml
@@ -249,30 +249,83 @@ pathophysiology:
       phenotype, which is what distinguishes sequestration from incidental
       co-localisation.
   downstream:
-  - target: Missplicing and GAT4 Upregulation with Loss of Cerebellar GABAergic Inhibition
+  - target: MBNL1-Dependent Missplicing of CNS Targets
     causal_link_type: DIRECT
     hypothesis_groups:
     - rna_gain_of_function
     description: >-
       MBNL1 depletion alters splicing of its CNS targets, of which GAT4/Gabt4 is
       the demonstrated example.
-- name: Missplicing and GAT4 Upregulation with Loss of Cerebellar GABAergic Inhibition
+- name: MBNL1-Dependent Missplicing of CNS Targets
   description: >-
-    MBNL1 depletion shifts the splicing of its CNS targets, and the one that has
-    been traced through to a physiological consequence is the GABA-A transporter
-    GAT4/Gabt4. Its upregulation is predicted to clear GABA from the synaptic
-    cleft faster, and in vivo optical imaging in SCA8 mice confirms the
-    corresponding loss of GABAergic inhibition in the granular cell layer. This
-    node matters because it is the point at which a splicing abnormality becomes
-    an electrophysiological one in cerebellar circuitry.
+    With MBNL1 sequestered, the alternative splicing it normally regulates
+    shifts across its CNS target set. This is the molecular consequence of
+    sequestration and is separated from what any individual mis-spliced target
+    then does, because the splicing shift is general while only one target has
+    been traced through to a physiological effect.
   role: mechanism
-  biological_scale: CELLULAR
+  biological_scale: MOLECULAR
   biological_processes:
   - preferred_term: RNA splicing
     modifier: ABNORMAL
     term:
       id: GO:0008380
       label: RNA splicing
+  evidence:
+  - reference: PMID:19680539
+    reference_title: RNA gain-of-function in spinocerebellar ataxia type 8.
+    supports: SUPPORT
+    directness: DIRECT
+    evidence_source: MODEL_ORGANISM
+    snippet: >-
+      SCA8 CUG(exp) transcripts trigger splicing changes and increased expression
+      of the CUGBP1-MBNL1 regulated CNS target, GABA-A transporter 4
+      (GAT4/Gabt4)
+    explanation: >-
+      Establishes that the expansion transcripts alter splicing of MBNL1
+      regulated CNS targets, which is the claim of this node.
+  downstream:
+  - target: GAT4 Upregulation
+    causal_link_type: DIRECT
+    hypothesis_groups:
+    - rna_gain_of_function
+    description: >-
+      Increased expression of the GABA-A transporter GAT4 is the one
+      demonstrated consequence of the splicing shift.
+- name: GAT4 Upregulation
+  description: >-
+    Expression of the GABA-A transporter GAT4/Gabt4 rises. It is picked out from
+    the wider splicing shift because it is the single target whose consequence
+    has been measured rather than inferred, and because a transporter that
+    clears GABA from the synaptic cleft has an obvious and testable
+    physiological prediction.
+  role: mechanism
+  biological_scale: MOLECULAR
+  evidence:
+  - reference: PMID:19680539
+    reference_title: RNA gain-of-function in spinocerebellar ataxia type 8.
+    supports: SUPPORT
+    directness: DIRECT
+    evidence_source: MODEL_ORGANISM
+    snippet: >-
+      increased expression of the CUGBP1-MBNL1 regulated CNS target, GABA-A
+      transporter 4 (GAT4/Gabt4)
+    explanation: >-
+      Names the specific expression change this node records.
+  downstream:
+  - target: Loss of Cerebellar GABAergic Inhibition
+    causal_link_type: DIRECT
+    hypothesis_groups:
+    - rna_gain_of_function
+    description: >-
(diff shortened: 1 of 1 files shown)


matentzn:
All six blocking findings addressed, plus five of the seven suggestions. Pushing shortly; this comment is the map.

**Blocking**

1. **`references:` block added** tagging `PMID:20301445` with `tags: [GeneReviews]`. You are right that the mining was auditable only by reading the file.
2. **`OBLIGATE` → `VERY_FREQUENT`** on Progressive cerebellar ataxia. This was the finding I most agree with: my own PR body cites the non-ataxic presentations, so the entry contradicted itself. The description now says why the grade is what it is.
3. **`animal_models:` added** for the SCA8 BAC (CTG)116 mouse with four `modeled_mechanisms` links — `RECAPITULATES` on the bidirectional-transcription node (HIGH) and on the GABAergic-inhibition node (MODERATE, with an imaging readout), `PERTURBS` on the MBNL1 node for the *Mbnl1* cross, and `PARTIALLY_RECAPITULATES` on the polySerine node, limited by the reported distribution differences between mouse and human brains. You were right that the `HUMAN_MODEL_MISMATCH` discussion had nothing structured to point at.
4. **NCT03701399 added**, with the generic SCA sentence and `directness: INDIRECT` — the registry text does not name SCA8, so its relevance here is an inference from eligibility, and the entry says that rather than implying an SCA8 trial.
5. **Sensory neuropathy and extrapyramidal signs added**, using the GeneReviews sentence already in the file.
6. **The GAT4 node split** into three: MBNL1-dependent missplicing (MOLECULAR) → GAT4 upregulation (MOLECULAR) → loss of cerebellar GABAergic inhibition (CELLULAR), chained with `downstream`. Your point that the single scale tag was itself the signal is a useful test and I have applied it to the other bundled node too.

**Suggestions**

7. **PolySer node split** as suggested, joined with `INDIRECT_UNKNOWN_INTERMEDIATES`. You were right that this preserves the caveat rather than leaving it in prose.
8. **Repeat size added**, but only the figure I can quote from a fetched primary source — Koob's `107-127 CTG repeats`. I did not add the wider ~15–50 / 71–>1300 ranges: they come from the research report's web sources, not from anything cached, and in a disease where repeat size predicts neither penetrance nor onset an unciteable range invites over-reading.
11. **Mixed-source item fixed, but not by splitting.** Duplicating one sentence under two `evidence_source` values is precisely what `check-snippet-grading` gates, so that fix would have failed CI. Instead the node now carries a genuinely human-only sentence (`Immunohistochemistry of human SCA 8 autopsy tissue (n = 4–7)…`) graded `HUMAN_CLINICAL`, and the mixed damage sentence graded `OTHER` with an explanation saying it spans human autopsy and mouse findings and cannot be split textually.
13. **GeneReviews management mined further** — occupational therapy with adaptive equipment, and physical activity for conditioning, as separate treatments since they target different things.

**Not done, with reasons**

- **9 (`mappings:`)** — the identifiers are in the research report, which is exactly the class of content this run treated as unverified leads. I would rather add them from a resolved source than transcribe them.
- **10 (cognitive impairment frequency)** — `just fetch-reference ORPHA:98760` is the right move and I could not make it work here; the Orphanet source is bulk-cached and that ID is not in the cache. Left at `OCCASIONAL`, which is what the GeneReviews "in some" supports, rather than raised on an uncitable figure.
- **12 (cell types)** — molecular layer interneurons, Bergmann glia and deep cerebellar nuclei are named in the research report from Daughters' full text, but the cached abstract says only "selected neurons in the brain". Adding them would mean citing a sentence that does not contain them.

One thing changed beyond your findings: `prevalence[0].measure_type` was `PERIOD_PREVALENCE` for "2-5% of autosomal dominant inherited ataxia", which is a share of a disease denominator, not a prevalence over a period. `PrevalenceMeasureEnum` has no value for a proportion-of-cases figure, so it is now `UNKNOWN` with a note explaining the gap. The same correction went to the other entries in this run.

**61/61 snippets verified** after the changes; all offline gates clean.

---

Issue #7855, opened 2026-08-02
Add `just check-pathograph`: reachability and dangling-reference audit for pathophysiology graphs


cmungall:
## Motivation

The validation stack checks assertions **individually** — `linkml-validate` for schema shape, `linkml-term-validator` for ontology term existence and label match, `linkml-reference-validator` for snippet-in-abstract. None of them check the *shape of the pathophysiology graph*, so a structurally broken pathograph passes every gate while every individual assertion in it is valid.

This surfaced concretely in #7833. The ARC syndrome entry had an **orphaned `Neurogenic Arthrogryposis` node** — no incoming `downstream` edge from anywhere, so it was disconnected from the proximal lesion and would render as a floating node. `just validate`, `just validate-terms` and `just validate-references` were all green. It was caught by hand, and the reviewer ran the same ~30-line audit independently in each of four review rounds.

Both the curator and the reviewer wrote the same throwaway script. That is the signal it belongs in the validation stack.

## Proposed checks

For each `kb/disorders/*.yaml`, `kb/modules/*.yaml` and `kb/comorbidities/*.yaml`:

1. **Reachability** — every `pathophysiology[]` node reachable from the root(s) by following `downstream[].name`. Report unreachable nodes.
2. **Root count** — report the set of nodes with no incoming edge. More than one root is usually legitimate (multiple independent triggers) but worth surfacing; **zero** roots means a cycle with no entry point.
3. **Dangling references** — every `downstream[].name`, `treatments[].target_mechanisms[]` and `discussions[].attaches_to` resolves to a declared node. (`attaches_to` uses the `pathophysiology#<node>` grammar.)
4. **Self-edges** — a node listing itself in its own `downstream`.

Cycles are worth reporting but should probably be advisory rather than an error — feedback loops are real biology (e.g. inflammation → injury → inflammation) and some entries may model them deliberately.

## Suggested shape

Advisory by default, gating under `--strict`, following the `just check-groupings` precedent, which already splits a gating structural linter from an advisory evaluator:

```bash
just check-pathograph                                    # all entries, advisory
just check-pathograph kb/disorders/Asthma.yaml           # single file
just check-pathograph --strict                           # exit non-zero on unreachable/dangling
```

Dangling references and self-edges look safe to gate on immediately (they are unambiguous errors). Unreachable nodes should probably start advisory until the existing KB is audited — there may be a backlog of entries with orphans that nobody has looked at, and that count is itself worth knowing.

## Notes

[...]

Issue #7003, opened 2026-07-25
PrevalenceMeasureEnum has no value for a newborn-screening screen-positive rate


cmungall:
## The gap

`PrevalenceMeasureEnum` has no member for a **screening yield** — the rate at which newborns screen positive, as distinct from the rate at which they have the disease.

The two differ by the test's false-positive rate, often by an order of magnitude. For GAMT deficiency (#7001):

| quantity | value | source |
|---|---|---|
| Birth prevalence | 0.5–2 per **million** | PMID:37465909 |
| Screen-positive rate | ~1 per **100,000** | PMID:37465909 |

Both come from the same sentence of the same paper, and the gap between them is the *point* — the source cites "the low number of positive results" as an argument for screening feasibility. Curating only the prevalence loses a fact that matters for screening-program design; curating the screen-positive rate as a prevalence would overstate the disease roughly tenfold.

## Current workaround and why it's unsatisfying

In #7001 I recorded it as a second `Prevalence` record with `measure_type: UNKNOWN`, `prevalence_class: UNKNOWN`, `rate_per_100000` deliberately unset, and the distinction spelled out in `notes`.

A reviewer correctly pointed out that `UNKNOWN` is documented as *"not stated or cannot be determined"* — but here the measure is **precisely determined**; it simply has no enum value. So the tag is actively wrong about the epistemic situation, not merely imprecise.

## Proposal

Add `SCREEN_POSITIVE_RATE` to `PrevalenceMeasureEnum`, described as something like:

> The rate at which individuals screen positive in a population screening program. Distinct from disease prevalence: it includes false positives and is therefore typically higher, often by an order of magnitude. Use for reporting screening-program yield.

Sibling consideration: a `SCREEN_DETECTED_INCIDENCE` (confirmed cases per screened population) may also be worth having — several entries in the urea-cycle and organic-aciduria families quote figures of that shape (e.g. "1 case in 1.4 million screened"), which is a third distinct quantity again.

## Why it's worth having

Newborn screening is being actively expanded — GAMT deficiency was added to the US Recommended Uniform Screening Panel in January 2023 — so more entries will acquire screening-yield figures. Without an enum value each curator will independently pick between `UNKNOWN` (wrong about epistemics), a prevalence measure (wrong about the quantity), or dropping the figure (loses information).

Happy to implement if the direction is agreed.

🤖 Filed by [Claude Code](https://claude.com/claude-code) during the oldest-entry review sweep.

[...]

PR #10383, opened 2026-09-01
curate(Congenital_Laryngomalacia): a tone deficit, not a soft cartilage (MONDO:0007878)

### LINKED ISSUE #10371: Curate congenital laryngomalacia (MONDO:0007878)
Curate a dismech entry for **congenital laryngomalacia** ([MONDO:0007878](https://monarchinitiative.org/MONDO:0007878)) — the commonest cause of stridor in infants, in which supraglottic tissue collapses inward on inspiration. The mechanism is contested in an interesting way: the older "cartilage is too soft" account has largely given way to a neuromuscular/sensorimotor-integration account, with laryngopharyngeal reflux as a proposed aggravator rather than a cause.

Prior art check: laryngomalacia appears across the KB only as a phenotype of other syndromes (CHOPS, Meier-Gorlin, several neurodevelopmental entries); there is no entry for the isolated congenital disorder.

The standard anatomic classification (short aryepiglottic folds; redundant arytenoid mucosa; posteriorly displaced epiglottis) describes where the collapse happens, not distinct diseases, so it most likely belongs in `has_subtypes` on a single entry.

Stub: `stubs/Congenital_Laryngomalacia.yaml`

matentzn:
Laryngomalacia is the commonest cause of stridor in infancy and appeared in the KB only as a phenotype of other syndromes (CHOPS, Meier-Gorlin, several neurodevelopmental entries). There was no entry for the isolated congenital disorder.

Closes #10371

## The framing choice

The name says cartilage softening. The entry deliberately does not model it that way.

Laryngopharyngeal sensory testing delivers a calibrated air pulse to the aryepiglottic fold to trigger the laryngeal adductor reflex. In laryngomalacia the threshold is raised — 4.1 mm Hg in mild disease, 6.8 in moderate, 7.4 in severe — and by nine months it falls back to 3.1–3.5 in step with symptom resolution. A structurally soft cartilage would not do that. So the trigger node is **immature sensorimotor integrative function of the larynx**, and the anatomical features (omega epiglottis, short aryepiglottic folds, redundant arytenoid mucosa) are a *parallel predisposing node* that sets how much collapse a given loss of tone produces.

That split is what makes the treatment section coherent, and it is the part most worth reviewing:

| Treatment | Node targeted | Why |
|---|---|---|
| expectant management | trigger (sensorimotor immaturity) | bets on it maturing; the strategy *is* the natural history |
| acid suppression | reflux-associated oedema | removes the aggravator, not the deficit |
| supraglottoplasty | predisposing configuration | removes the substrate collapse needs |

None of them restores tone, and the entry says so.

## Judgement calls

- **Reflux is modelled as an aggravating node, not an upstream cause.** The direction is genuinely disputed — laryngomalacia may promote reflux by altering the thoraco-abdominal pressure gradient. An upstream model would predict that acid suppression shortens the disease, and the evidence supports symptom improvement rather than that. A `KNOWLEDGE_GAP` discussion records the choice as provisional.
- **No numbered Type 1/2/3 subtypes.** The sources consulted assign those numbers to *different anatomy* — one makes Type 1 short aryepiglottic folds, another makes it arytenoid prolapse. The individual endoscopic features are reproducible between examiners; the numbering is not. A second discussion raises whether any anatomic classification carries prognostic information.
- **Expectant management has an unbound `treatment_term`.** `NCIT:C15722` Patient Observation is not reachable from `NCIT:C25218` and is correctly rejected by the `TreatmentActionTerm` enum — watchful waiting is the decision *not* to intervene. No NCIT clinical-action term expresses it, so `preferred_term` carries it alone, with the reason in `notes`.
- **The 42% aspiration figure is fenced.** It comes from 142 infants *selected* for having recurrent respiratory or feeding difficulty, out of 395 diagnosed that year. The node and the phenotype both say so, and the phenotype `frequency` is set to `OCCASIONAL` for the disease as a whole rather than from that number. The striking part is separate and not denominator-dependent: 98% of the detected aspiration was silent.

Perplexity `sonar-deep-research`, 22 citations, committed under `research/`. Its `reference_validation` frontmatter reports 5 of 5 references resolved, confabulation rate 0.0 — which is a clean-looking number that means very little here, because **only 5 of the 22 citations were identifiers the validator could see at all**.

Three of the 22 are not citations in any form: entry 20 is a fragment of prose, entry 21 is the phrase "Poiseuille's law", and entry 22 is the bare string "Landry et al". The rest are web pages (Cleveland Clinic, UpToDate, StatPearls, GeekyMedics, teachmepaediatrics). Every claim retained here was re-anchored to a PMID and quoted from the fetched cache.

The report's ontology bindings were not reused: it offered `UBERON:0001737` as epiglottis, which is **larynx** (epiglottis is `UBERON:0000388`).

## Validation

`just validate` clean — schema, terms, and **33/33 snippets verified**. Also clean: `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

---

### DIFF
--- a/history/disorders/Congenital_Laryngomalacia/2026-09-01T160144Z-claude-code-09ad5e.yaml
+++ b/history/disorders/Congenital_Laryngomalacia/2026-09-01T160144Z-claude-code-09ad5e.yaml
@@ -1,0 +1,41 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Congenital_Laryngomalacia
+  path: kb/disorders/Congenital_Laryngomalacia.yaml
+session:
+  id: 2026-09-01T160144Z-claude-code-09ad5e
+  timestamp: '2026-09-01T16:01:44Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10371
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10383
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - treatments
+  summary: 'Review response: wire the isolated phenotypes into the pathograph'
+  details: |
+    Second session on this entry, responding to two rounds of review on #10383.
+
+    Round one addressed a single orphaned phenotype (Recurrent respiratory infections, whose description asserted an aspiration mechanism that no edge carried) by adding a sequelae edge from Aspiration, and added five phenotypes for the severe end of the disease. That was half the fix: the five new phenotypes were themselves added without edges, so repairing one orphan created five more, and the reviewer's second round correctly blocked on it. The first attempt at those phenotypes also quoted table cells (for example 'Apnea 17 (19)'), which check-snippet-length rejected as non-propositional; they were re-anchored to the severity-criteria sentence in PMID:26924742 and the operative-indication sentence in PMID:18392498, and a plain Apnea phenotype was dropped rather than kept on a table cell.
+
+    Round two wired all of them in. Dynamic Supraglottic Airway Obstruction now carries five further downstream edges, each with its own evidence item: Dyspnea and Intercostal retractions as DIRECT, and Cyanosis, Obstructive sleep apnea and Gastroesophageal reflux as INDIRECT_KNOWN_INTERMEDIATES. The reflux edge is the one that needed a decision. The reviewer suggested drawing it from Reflux-Associated Supraglottic Oedema, but that node's own claim runs reflux to oedema to obstruction, so an edge out of it to the reflux phenotype would reverse its meaning. It is instead drawn obstruction to reflux on the other half of the same literature, quoting the source's explicitly hedged sentence that some posit CLM can lead to reflux disease by affecting the thoraco-abdominal pressure gradient. The edge description states that this is one of two competing directions and points at the clm_reflux_causal_direction discussion, and directness is INDIRECT because the source sentence is a hedge rather than a finding. The graph therefore now carries both directions without asserting that the dispute is settled.
+
+    Growth delay was folded into Failure to thrive rather than wired separately, taking the reviewer's non-blocking observation that the two were distinguished only by which word the surgical and feeding literatures use, which is provenance rather than phenotype. The PMID:18392498 quote is now a second evidence item on Failure to thrive, marked INDIRECT because the source says growth delay. No phenotype in the entry is now an isolated node and no causal target dangles, checked by building the edge set from the file rather than relying on check-causal-targets, which detects targets resolving to nothing and not nodes that nothing resolves to.
+
+    Tracheostomy was rebound from the generic NCIT:C15329 Surgical Procedure to NCIT:C15341 Tracheotomy, keeping preferred_term tracheostomy for the procedure and stoma distinction. The two cache rows for HP:0030864 that the previous push left uncommitted were committed; running just validate-terms had written them and I had not checked whether it wrote anything. The roughly ten percent operative rate from PMID:18392498 was added to Expectant Management, since it is the figure that makes watchful waiting a default rather than a gamble.
+
+    Not taken: rebinding Dyspnea to a respiratory-distress observable. The source says severe dyspnea and the phenotype is anchored to that quote, so relabelling it would leave the binding describing something the snippet does not say.
+
+    Validation: just validate clean (schema, terms, references, 47/47 snippets verified), plus check-duplicate-keys, check-entity-refs, check-causal-targets, check-qualifier-terms, check-enum-values, check-folded-hyphens, check-snippet-length, check-title-snippets, check-snippet-grading, check-environmental-evidence and check-term-cache-integrity, all clean.


matentzn:
Both blocking items fixed, and all four suggestions taken. Pushing once the whole-KB gates finish; this is the map.

**Blocking**

1. **`Recurrent respiratory infections` is now wired in.** Added a `sequelae` edge from the `Aspiration` phenotype, `causal_link_type: INDIRECT_KNOWN_INTERMEDIATES` — the intervening steps (inoculation, impaired clearance, consolidation) are real and omitted, so the edge should say so. You were right that the prose asserted a mechanism the graph did not carry, and right again that `check-causal-targets` cannot see this: it catches targets that resolve to nothing, not nodes nothing points at. An orphan-node check would be a genuinely different gate.

2. **Six phenotypes added**, all from caches already in the PR:

   | Phenotype | Source | Figure |
   |---|---|---|
   | Apnea | PMID:26924742 | 17 (19%) |
   | Dyspnea | PMID:26924742 | 24 (27%) |
   | Cyanosis | PMID:26924742 | 18 (20%) |
   | Intercostal retractions | PMID:26924742 | 15 (17%) |
   | Obstructive sleep apnea | PMID:18392498 | among the ~10% needing surgery |
   | Growth delay | PMID:18392498 | same sentence |

   Your point about apnoea was the sharpest thing in the review: the entry cited "consequences of airway obstruction and hypoxia" as a surgical indication with nothing for it to attach to, so the severe end of the disease existed only as a treatment. Each of the four `PMID:26924742` items carries a note that the denominator is a tertiary surgical referral series, not all cases.

**Suggestions**

3. **`BYPASSES` → `INHIBITS`** on supraglottoplasty. You are right that it read backwards: the operation abolishes the node it targets, where `BYPASSES` implies routing around a mechanism still running. I added **tracheostomy** as a contrast case that genuinely is `BYPASSES` — the supraglottic mechanism continues untouched and air is routed around it. Having both in one entry makes the distinction legible.
4. **GERD `FREQUENT` → `OCCASIONAL`**, to match its own 26.45% citation. The higher figures I had in mind (60–80%) are from the report's web sources, not from anything cached.
5. **`PERIOD_PREVALENCE` → `UNKNOWN`** on the 45–75%-of-congenital-stridor record. This was the most useful of the four, because it is a schema gap rather than a curation slip: `PrevalenceMeasureEnum` has no value for a proportion of a clinical denominator. `UNKNOWN` plus a note is the least-wrong option and not a good one — it makes a precisely-stated measure look uncharacterised. I raised it with data on the existing #7003 rather than filing again, and applied the same correction to the two other entries in this run that had it (#10376, #10382).
6. **Tracheostomy added** as above, quoting "Two patients received a tracheostomy" — two of 89, and the entry says so, because the point of including it is that it marks the ceiling of the severity spectrum rather than being a routine option.

**42/42 snippets verified** after the changes.

---

matentzn:
Correction to my previous comment: the phenotype table I posted there describes the first attempt, and it does not match what I pushed in 9daefbdb6.

Quoting those table cells — `Apnea 17 (19)`, `Dyspnea 24 (27)` and so on — failed `check-snippet-length`, correctly. Its message is exactly right about them: a table cell "carries no propositional content, cannot support or refute anything, and is usually lifted from a table whose cells do not survive text extraction." I had reached for the frequencies because you quoted them, without noticing that a frequency in a cell is not a sentence that makes a claim.

What is actually in the pushed commit:

| Phenotype | Quoted sentence |
|---|---|
| Cyanosis | *"Severe laryngomalacia was established if complaints included severe feeding difficulties, failure to thrive, cyanosis or intercostal and/or xyphoidal retractions…"* (PMID:26924742) |
| Intercostal retractions | same sentence |
| Dyspnea | *"…the respiratory obstruction can cause severe dyspnea, dysphagia, growth delay and obstructive sleep apnea"* (PMID:18392498) |
| Obstructive sleep apnea | same sentence |
| Growth delay | same sentence |

This is better than what I first wrote, and not only because it passes the gate. The severity-criteria sentence says something the table cell does not: cyanosis and retractions are part of the formal definition of severe disease in that series, so they are decision points rather than incidental observations. The descriptions now say that.

**A plain `Apnea` phenotype is not there.** The only support I could find for it was the table cell, and I would rather leave it out than quote a cell or stretch a sentence that does not contain it. Your underlying point still lands: the entry needed something for "consequences of airway obstruction and hypoxia" to attach to, and it now has three such phenotypes — obstructive sleep apnoea, cyanosis, and dyspnoea — all propositionally sourced.

If you think plain apnoea specifically needs representing, I will go looking for a paper that states it in prose rather than tabulates it.

41/41 snippets verified; `check-snippet-length` and the other nine gates all clean on the pushed head.

---

matentzn:
All three addressed in adf8a6a94. You're right that I did half the job — I read "add the phenotypes" and stopped there, which left the entry in exactly the shape the previous round's blocking item was about.

**The five isolated phenotypes are wired in.** `Dynamic Supraglottic Airway Obstruction` now carries five further `downstream` edges, each with its own evidence item:

| Target | `causal_link_type` | Evidence |
|---|---|---|
| Dyspnea | `DIRECT` | PMID:18392498 |
| Intercostal retractions | `DIRECT` | PMID:39525522 |
| Cyanosis | `INDIRECT_KNOWN_INTERMEDIATES` | PMID:22518182 |
| Obstructive sleep apnea | `INDIRECT_KNOWN_INTERMEDIATES` | PMID:18392498 |
| Gastroesophageal reflux | `INDIRECT_KNOWN_INTERMEDIATES` | PMID:39525522 |

Two of these are worth explaining rather than just listing.

**Reflux.** You suggested an edge from `Reflux-Associated Supraglottic Oedema`, but that node's own claim runs reflux → oedema → obstruction, so an edge out of it to the reflux phenotype would have the arrow backwards. I drew it obstruction → reflux instead, on the other half of the same literature: the paper explicitly offers it — *"On the other hand, some posit that CLM can lead to RD by affecting the normal pressure gradient between the thoracic and abdominal cavities, which normally protects against reflux events."* The edge description says in as many words that this is one of two competing directions, not the settled one, and points at `clm_reflux_causal_direction`. `directness: INDIRECT`, because the source sentence is itself a hedge ("some posit"). So the graph now carries both directions — the oedema node for reflux → obstruction, this edge for obstruction → reflux — which is what the discussion says the state of the evidence is, and neither edge asserts the question is closed.

**Cyanosis.** Anchored to PMID:22518182's "consequences of airway obstruction and hypoxia" rather than to the severity-criteria sentence. The severity sentence establishes cyanosis as a decision point but says nothing about where it comes from; this one names hypoxia as a consequence of the obstruction, which is the claim the edge makes. `INDIRECT` because cyanosis is the visible sign of that hypoxia rather than the word the sentence uses.

**Growth delay — folded rather than wired.** Taking your non-blocking note instead of the wiring, because you were right that the two phenotypes differed only in which word the surgical and feeding literatures use, and that is provenance rather than phenotype. The PMID:18392498 quote is now a second evidence item on `Failure to thrive`, marked `INDIRECT` because the source says "growth delay", and the description records that. That removes an overlap and an orphan in one move rather than connecting up a distinction that should not have existed.

**No phenotype in the entry is now isolated**, and no causal target dangles — I checked both by building the edge set from the file rather than trusting `check-causal-targets`, which as you say cannot see this class of defect.

**Cache rows committed.** `HP:0030864` is now in both `cache/hp/terms.csv` and the phenotype enum cache. That was a real miss: I ran `just validate-terms`, saw it pass, and did not look at whether it had written anything. `check-term-cache-integrity` is clean.

**Tracheostomy** rebound to `NCIT:C15341` Tracheotomy, `preferred_term: tracheostomy` unchanged.

**Also took the third of your standing suggestions**, the ~10% operative rate on Expectant Management. It is the number that makes watchful waiting a default rather than a gamble, so it belonged there. `INDIRECT`, since the sentence states the operative rate and its indications rather than endorsing observation for the rest.

**Not taken:** `Dyspnea` → respiratory distress. The source says "severe dyspnea" and the phenotype is anchored to that quote, so switching the term would leave the binding describing something the snippet does not say. If you want the observable instead, the honest version is a different quote, not a different label on this one.

The other two cached-but-unused quotes you listed — the comorbidity-severity association in PMID:17513991 and the laryngeal-cleft differential in PMID:30589926 — I have left for now; say the word and I will add them.

Validation: schema, terms and references pass, 47/47 snippets verified, all ten offline gates plus `check-term-cache-integrity` green.

---

matentzn:
Housekeeping note, since the timeline on this PR looks odd.

My previous push (`adf8a6a94`, the pathograph wiring described in the comment above) landed on the branch at ~15:36Z but **GitHub did not advance the PR ref for about 25 minutes**. `refs/heads/claude/claim-disease-review-cycle-lxlhyl-laryngo` was at `adf8a6a94` while `refs/pull/10383/head` stayed at `9daefbdb6`, so no CI and no re-review were triggered — the PR looked untouched. I could not dispatch `claude-code-review.yml` myself (`actions_run_trigger` returns 403 for this integration).

I deliberately did not push an empty commit or close-and-reopen to force it. Instead I pushed `bc2803beb`, which is real work the repo convention already asks for: a **second history record for the review-response session**. The existing record describes only the creation session (33/33 snippets, no mention of any of the wiring), and records are append-only, so a second record is the correct form rather than an edit to the first. That push brought the PR ref current immediately, and it now carries `adf8a6a94` as well.

So the diff to review is both commits together. Nothing in the wiring changed between them.

One small thing for the sibling PRs, where the same history-record suggestion was made: the suggested command `just new-history --event UPDATE` is rejected — the permissible values are `GENERAL`, `CREATE`, `EDIT`, `REVIEW`, `AUDIT`. `REVIEW` is the right one for a review-response session.

`just validate-history` passes on the new record.

---

matentzn:
Thanks — and on the last point, that issue already exists: **#7855**, "Add `just check-pathograph`: reachability and dangling-reference audit for pathophysiology graphs." Its proposed check 1 (reachability) and check 2 (root count) are exactly this defect. I checked before filing and added this run's data there rather than opening a second one.

Worth repeating here, because the frequency is the argument: orphans came up in **four of the five entries** in this run, and were found by hand every time. This PR twice, the 6q16 deletion once (a three-node arm), SCA8 (9 components), and SEMD — where I found four pre-existing orphans in an entry I had written, validated and read back myself, and only because you had raised the same class of defect on two other PRs that morning. Every gate was green throughout in all four cases.

One refinement from hitting it repeatedly, which I put on #7855: the useful unit is not "pathophysiology node with no incoming edge" but any node in `pathophysiology` *or* `phenotypes` that is neither an edge target nor a plausible root. Most of what turned up were phenotypes, and a phenotype is never a root — so a phenotype with no incoming edge is always either a missing edge or a claim the entry declined to make. Distinguishing those two is a curator's call; surfacing them is not.

On the two remaining cached quotes: agreed both belong, and I have left them for a follow-up rather than pushing again now. The PMID:17513991 comorbidity-severity sentence is the one I would prioritise — it is independent weight for the neuromuscular thesis, which currently rests on the LPST thresholds alone.

---

PR #10384, opened 2026-09-01
curate(6q16_Deletion_Syndrome): a hypothalamic pathway deleted at two points, not one gene (MONDO:0015749)

### LINKED ISSUE #10372: Curate 6q16 deletion syndrome (MONDO:0015749)
Curate a dismech entry for **6q16 deletion syndrome** ([MONDO:0015749](https://monarchinitiative.org/MONDO:0015749)) — a contiguous-gene deletion presenting as a Prader-Willi-like picture (hyperphagic obesity, hypotonia, developmental delay), generally attributed to haploinsufficiency of *SIM1* and its role in paraventricular hypothalamic development and downstream melanocortin signalling.

Prior art check: no KB entry and no mention of the 6q16 interval in `kb/disorders` or `kb/groupings`.

Worth deciding explicitly during curation: whether the entry is the *deletion syndrome* (variable interval, several genes) or *SIM1 haploinsufficiency*. These are not the same claim, and the deletion-size variability is exactly the kind of thing that should be represented rather than smoothed over. Point mutations in *SIM1* causing isolated obesity are a separate concept and should not be merged in without argument.

Stub: `stubs/6q16_Deletion_Syndrome.yaml`

matentzn:
No entry, and no mention of the 6q16 interval anywhere in `kb/disorders` or `kb/groupings`.

Closes #10372

## The judgement call worth reviewing

The conventional shorthand is that 6q16 deletion syndrome *is* SIM1 haploinsufficiency wearing a cytogenetic name. That attribution is genuinely well supported — SIM1 loss-of-function point mutations produce severe obesity with Prader-Willi-like features on their own, and in the cited study the measured loss of transcriptional activity of each variant tracked the obesity risk it conferred *within families*. That is stronger evidence than most gene-disease assignments carry.

The entry still declines to model it that way, because two findings in the same literature resist it:

1. A **~350 kb critical region at 6q16.1 containing POU3F2 and not SIM1** produces overlapping developmental delay and obesity susceptibility. Zebrafish work then places POU3F2 *downstream* of SIM1, controlling oxytocin expression in the neuroendocrine preoptic area.
2. Deletions at the adjacent **6q14.1-q15 interval, sparing SIM1 entirely**, produce a strikingly similar phenotype.

So the pathograph branches from the deletion node into separate SIM1 and POU3F2 arms that converge on dysregulated hypothalamic control of energy balance — one pathway interruptible at two points, rather than one gene lost. The epistatic ordering is what makes the two critical regions coherent rather than coincidental.

Gene-to-phenotype attribution is left open in a `KNOWLEDGE_GAP` discussion that carries the SIM1-sparing deletion paper as its own `evidence`. That is not hedging: it decides whether deletion size should be expected to predict severity, which is what a clinician looking at an array result actually wants to know.

## Evidence discipline

- **Hypopituitarism rests on a single reported patient** and the entry says so in `notes`. `frequency` is `OCCASIONAL` on that basis. The clinically useful part is the timing — it was absent on endocrine testing in infancy and appeared only in later childhood, so a normal early screen does not exclude it, which is why longitudinal endocrine evaluation is listed under `diagnosis`.
- **Short stature is marked `directness: INDIRECT`** because its source raises it as "a possibility... especially short stature" rather than establishing it.
- **Obesity is described as susceptibility, not an obligate feature**, following the source's own wording — which matters because it means the outcome is partly modifiable.
- **No weight-management treatment is listed.** It is standard care, but nothing in the sources used here evaluates it in this syndrome, and the `notes` on the treatments section says that rather than importing a plausible entry.

Perplexity `sonar-deep-research`, 18 citations, committed under `research/`. Its `reference_validation` reports 7 of 7 resolved at confabulation rate 0.0 — again a clean number covering only the 7 of 18 citations the validator could see, with 5 of those 7 judged on topic and one citation-list entry being the bare string "ACMG/AMP".

Its `term_validation` was the more useful signal: **8 of 20 checked labels named a different term than the report claimed**, so no CURIE was reused. `SIM1` (`hgnc:10882`) and `POU3F2` (`hgnc:9215`) were resolved through OAK.

## Validation

`just validate` clean — schema, terms, and **23/23 snippets verified**. Also clean: `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

---

### DIFF
--- a/kb/disorders/6q16_Deletion_Syndrome.yaml
+++ b/kb/disorders/6q16_Deletion_Syndrome.yaml
@@ -90,6 +90,15 @@ pathophysiology:
     description: >-
       Deletions covering the 6q16.1 critical region remove one copy of POU3F2,
       whether or not SIM1 is involved.
+  - target: Impaired Neurodevelopment
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+    description: >-
+      The deletion removes neurodevelopmental transcription factors, and the
+      delay is present even in the smallest POU3F2-only deletions, so it is
+      drawn from the deletion itself rather than from either hypothalamic arm.
+      Marked indirect because which of the deleted genes produces the delay, and
+      through what developmental step, is exactly the open question this entry
+      records in discussions.
 - name: Loss of SIM1-Dependent Paraventricular Hypothalamic Development
   description: >-
     SIM1 is a transcription factor required for development of the
@@ -352,6 +361,35 @@ phenotypes:
     explanation: >-
       Ties obesity to the 6q16 deletion specifically, in the Prader-Willi-like
       clinical context this entry describes.
+- name: Hypotonia
+  category: Neurological
+  description: >-
+    Infantile hypotonia is part of the Prader-Willi-like resemblance and
+    typically precedes the hyperphagia. The evidence available here is about
+    proximal 6q deletion patients as a group rather than about the 6q16 deletion
+    specifically, which is why it is recorded at OCCASIONAL with INDIRECT
+    evidence rather than asserted as a defining feature.
+  frequency: OCCASIONAL
+  phenotype_term:
+    preferred_term: Infantile hypotonia
+    term:
+      id: HP:0001252
+      label: Hypotonia
+  evidence:
+  - reference: PMID:19664229
+    reference_title: >-
+      Microdeletion of 6q16.1 encompassing EPHA7 in a child with mild
+      neurological abnormalities and dysmorphic features: case report.
+    supports: SUPPORT
+    directness: INDIRECT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      Reported 6q deletion patients show a high incidence of mental retardation,
+      ear anomalies, hypotonia, and postnatal growth retardation.
+    explanation: >-
+      Reports hypotonia among the recurring features of 6q deletion patients.
+      Marked INDIRECT because the statement is about 6q deletions as a class, not
+      about the 6q16 interval this entry is scoped to.
 - name: Global developmental delay
   category: Neurological
   description: >-
@@ -596,8 +634,8 @@ treatments:
   treatment_term:
     preferred_term: hormone replacement therapy
     term:
-      id: NCIT:C15986
-      label: Pharmacotherapy
+      id: NCIT:C15599
+      label: Hormone Replacement Therapy
   evidence:
   - reference: PMID:24038875
     reference_title: >-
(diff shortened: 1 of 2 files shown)


matentzn:
All three fixed, and finding 2 turned up something that strengthens the entry's argument. Pushing once the whole-KB gates finish.

**1. The disconnected component.** `Impaired Neurodevelopment` now has an incoming edge from the deletion trigger, `causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES`. I drew it from the deletion itself rather than from either hypothalamic arm, because the delay is present in the POU3F2-only deletions and there is no basis for routing it through the energy-balance nodes. The indirect marking is deliberate: *which* deleted gene produces the delay is exactly the open question this entry already records.

Worth noting for the repo — this is the same class of defect as #10383's orphan `Recurrent respiratory infections`, and neither gate catches it. `check-causal-targets` finds targets that resolve to nothing; nothing finds nodes that nothing resolves *to*. Two independent occurrences in one five-entry run suggests it is a common failure mode.

**2. The hypotonia claim — searched rather than dropped.** You were right that it was unsupported: none of the four cited abstracts mentions it, and the description asserted it anyway. `just fetch-reference ORPHA:171829` does not work here (the Orphanet source is bulk-cached and that ID is not in the cache), so I searched PubMed instead and found `PMID:19664229`, which states:

> Reported 6q deletion patients show a high incidence of mental retardation, ear anomalies, hypotonia, and postnatal growth retardation.

Hypotonia is now a curated phenotype at `OCCASIONAL` with `directness: INDIRECT`, because the statement is about 6q deletions as a class rather than the 6q16 interval this entry is scoped to.

**The useful part is what else that paper says.** It is a 6q16.1 microdeletion case report proposing *EPHA7* as a contributor to the neurodevelopmental features — a **third** dosage-sensitive gene in the interval, distinct from both SIM1 and POU3F2. I have added it as a second evidence item on the gene-attribution `KNOWLEDGE_GAP` discussion. It makes the entry's central argument stronger than when I wrote it: the case against a SIM1-only model no longer rests on one SIM1-sparing deletion, it rests on two independent proposals of other genes in the same region.

**3. `NCIT:C15599` Hormone Replacement Therapy** used in place of the generic `NCIT:C15986`. Thank you for verifying reachability from `NCIT:C25218` — that is the constraint I would have had to check anyway.

Re-validated: **26/26 snippets verified**, all offline gates clean. One new reference cache (`PMID_19664229`) added.

---

PR #9082, opened 2026-08-20
Add RSPRY1-related spondyloepimetaphyseal dysplasia (SEMDFA) disorder entry

tannerzhang:
## Summary

This PR adds a comprehensive disorder knowledge base entry for Progressive Spondyloepimetaphyseal Dysplasia-short Stature-short Fourth Metatarsals-intellectual Disability Syndrome (SEMDFA), an ultra-rare autosomal recessive skeletal dysplasia caused by biallelic RSPRY1 variants.

## Key Changes

- **New disorder entry** (`kb/disorders/Progressive_Spondyloepimetaphyseal_Dysplasia-short_Stature-short_Fourth_Metatarsals-intellectual_Disability_Syndrome.yaml`): A 1910-line comprehensive knowledge base entry covering:
  - Disease classification (Mendelian, autosomal recessive)
  - MONDO mapping (MONDO:0014748) and Orphanet cross-reference (ORPHA:457395)
  - Complete phenotype descriptions with HPO term bindings
  - Detailed pathophysiology cascade: RSPRY1 biallelic loss of function → loss of TGF-β/SMAD restraint → ECM dysregulation → skeletal and developmental phenotypes
  - Inheritance patterns with supporting evidence
  - Molecular mechanisms (RING and B30.2/SPRY domains, ubiquitin ligase activity)
  - Evidence-backed clinical and molecular findings from 6 primary references

- **Research documentation**: Two deep-research reports (Claude Code and OpenScientist) with comprehensive literature synthesis and citation validation

- **Reference cache**: Five new reference entries (PMID:26365341, PMID:38562122, PMID:39940902, PMID:39706863, PMID:30063090, ORPHA:457395) with full-text content and metadata

- **Stub removal**: Removed placeholder stub file (`stubs/Progressive_Spondyloepimetaphyseal_Dysplasia-short_Stature-short_Fourth_Metatarsals-intellectual_Disability_Syndrome.yaml`)

- **Cache updates**: Added new ontology term entries (HP:0002651, HP:0004689, MONDO:0014748, hgnc:29420) to dynamic enum and label caches

## Implementation Details

- **Pathophysiology model**: Structured as a two-node cascade with ESTABLISHED confidence for the initial RSPRY1 loss-of-function trigger and PROVISIONAL confidence for the TGF-β/SMAD dysregulation node (reflecting that ubiquitination remains inferred rather than directly demonstrated)
[...]

PR #10385, opened 2026-09-01
curate(Spondyloepimetaphyseal_Dysplasia_Aggrecan_Type): a matrix-anchoring defect, not aggrecan deficiency (MONDO:0013014)

### LINKED ISSUE #10373: Curate spondyloepimetaphyseal dysplasia, aggrecan type (MONDO:0013014)
Curate a dismech entry for **spondyloepimetaphyseal dysplasia, aggrecan type** ([MONDO:0013014](https://monarchinitiative.org/MONDO:0013014)) — a recessive skeletal dysplasia from biallelic *ACAN* variants, where loss of aggrecan removes the proteoglycan that gives cartilage its compressive stiffness and disorganises the growth plate.

Prior art check: `ACAN` is referenced in `Multiple_Epiphyseal_Dysplasia.yaml`, `Brachydactyly_Type_E1.yaml`, `Osteochondrosis.yaml` and the acromesomelic dysplasia entries, but MONDO:0013014 has no entry. The open PR #9082 curates *RSPRY1*-related SEMD, a different gene and a different MONDO term — no overlap.

*ACAN* is allelic across a dominant/recessive severity range (idiopathic short stature and osteochondritis dissecans; dominant SED Kimberley type; this recessive SEMD). The entry should be scoped to the recessive aggrecan-type SEMD and relate the others rather than absorb them.

Stub: `stubs/Spondyloepimetaphyseal_Dysplasia_Aggrecan_Type.yaml`

matentzn:
MONDO:0013014 had no entry, although `ACAN` is referenced across `Multiple_Epiphyseal_Dysplasia`, `Brachydactyly_Type_E1`, `Osteochondrosis` and the acromesomelic dysplasias. The open PR #9082 curates *RSPRY1*-related SEMD — different gene, different MONDO term, no overlap.

Closes #10373

## The mechanism is not the one you would guess

Aggrecan is the proteoglycan that gives cartilage its compressive stiffness, so the natural model is aggrecan deficiency. The founding family says otherwise, and the entry is built around that:

- the causal variant is a **single missense substitution**, p.D2267N, in the C-type lectin domain of the C-terminal G3 domain;
- expressing the mutant showed it **traffics and is secreted normally** — it even gains a functional N-glycosylation site — so there is no haploinsufficiency, no ER retention, no degradation of a misfolded product;
- what changed was the **calcium-dependent binding of that domain to tenascin-C**, measured by surface plasmon resonance.

The lesion is in how aggrecan is *anchored into the matrix network*, not in how much of it there is.

**That negative secretion result gets its own `pathophysiology` node** rather than a clause in a description. It is not a detail — it is the experiment that excludes the mechanism a reader would otherwise assume, and burying it would let the entry be mis-modelled. `notes` on the node says exactly that.

## Two things I flagged rather than settled

- **Tenascin-C is named because it is the partner that was measured, not because the others were excluded.** The same domain binds tenascin-R, fibulin-1, fibulin-2 and fibrillin-1 in the same calcium-dependent way, and a calcium-coordinating substitution has no obvious reason to be selective. A `KNOWLEDGE_GAP` discussion records this so the single-partner framing is not read as established.
- **The two reported families differ enormously in severity** — 66–71 cm in the three original siblings, 118.3 cm (−9.1 SD) in the second patient, who had a normal face and lived independently to 45. The obvious candidate is allele class, but with two families there is no way to test it and no functional comparison has been done. The entry describes a severity *range* rather than a severity, and a second discussion says why.

## Scope and abstentions

- **Scoped to the recessive aggrecan-type SEMD.** The dominant ACAN disorders (SED Kimberley type, familial osteochondritis dissecans) are related in `differential_diagnoses` rather than absorbed — the allelic series is the reason the scope statement is explicit.
- **Growth hormone is deliberately not listed as a treatment.** It is used in other ACAN-related short stature, but nothing in the sources evaluates it here, and importing it would extrapolate across a different allele class and a different severity. The `notes` on the treatments section says so rather than leaving the omission silent.
- **Osteogenesis imperfecta is in the differential** because the second patient carried that diagnosis for decades before genetic testing. It is the practical differential, not a theoretical one.

## Evidence handling

One snippet initially failed reference validation because it stitched two non-contiguous sentences from the source. It was **split into two evidence items**, not paraphrased into something that would match.

`evidence_source: OTHER` is used for the three quotes taken from the ACAN review rather than from a primary study, and each `explanation` says so.

Perplexity `sonar-deep-research`, 19 citations, committed under `research/`. `reference_validation` could check 6 of the 19 (all resolved, 4 on topic). `term_validation` was the more useful signal: **8 of 33 checked labels named a different term than the report claimed**, so no CURIE was reused and every binding was resolved through OAK.

## Validation

`just validate` clean — schema, terms, and **26/26 snippets verified**. Also clean: `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`.

---


matentzn:
Thanks — all five blocking findings are addressed in cb55b5a03. I checked finding 1 against the cached source before acting on it rather than taking it on trust, and you're right.

**Finding 1 — the trigger node asserted a lectin-domain lesion for both families.** `PMID_30124491.md` says plainly: "In contrast, the present individuals had heterozygous mutations in the CS1 and CS2 domains, consisting of a large number of repeats." Since most of the entry's phenotypes come from that second patient, naming the node for the C-type lectin domain was wrong. Changes:

- Node renamed to `Biallelic ACAN Variants Disrupting Aggrecan Function`, with a description that separates the two families' lesions and says which one carries the functional work.
- Both downstream edges now state that the arm is established for the founding family's p.D2267N allele and that the CS-domain variants have not been tested against it.
- The family-2 variants are now in the `genetic:` section quoting the paper's notation verbatim — including its non-standard `p.1380V>F` / `p.1687S>R` protein forms, since a normalized version would not be a quote — together with the sentence recording that Mutation Taster was the only assessment, and a second item recording that the parents had died and the siblings declined testing, so phase was never established.
- `semd_acan_genotype_severity` rewritten. You're right that the honest gap is more interesting: the source *does* attribute the severity difference to domain location, so the gap is that nothing functional stands behind that attribution — prediction only, no comparison against p.D2267N, no phase, and the authors' own claim hedged as "we speculated that these ACAN mutations might have affected the patient's SEMD presentation." The discussion now says what would resolve it (the same binding work applied to the CS alleles, or a third family) instead of calling it untestable.

**Finding 2 — missing phenotypes.** Added eight: short neck (HP:0000470), barrel-shaped chest (HP:0001552), lumbar hyperlordosis (HP:0002938), mandibular prognathia (HP:0000303) and low-set ears (HP:0000369) from the original family; shallow acetabular fossae (HP:0003182), irregular vertebral endplates (HP:0003301) and thoracolumbar kyphosis (HP:0005619) from the second patient. Thoracolumbar kyphosis was indeed buried inside the platyspondyly snippet and is now its own entry with a `sequelae` edge from platyspondyly. Cervical lordosis has no HPO term — the nearest hits are lumbar or thoracic — so it is carried in the kyphosis description rather than bound to something inaccurate.

While wiring those in I found the entry had four pre-existing orphans of its own (Disproportionate short stature, Abnormal epiphysis morphology, Relative macrocephaly, Midface retrusion). All twelve are now connected. The craniofacial edges are marked `INDIRECT_UNKNOWN_INTERMEDIATES` with the reasoning stated: the cranial base is endochondral, so the growth-plate defect is the expected route to midface hypoplasia, but neither report examined it — that is inference from the mechanism, not a finding. Relative macrocephaly hangs off severe short stature, since "relative" is what the source means.

**Finding 3 — midface retrusion.** OCCASIONAL → FREQUENT (3 of 4). Added the positive family-1 quote, and regraded the negative second-patient finding as REFUTE against the reading that this feature is part of the syndrome in every patient. That is what keeps the frequency off OBLIGATE, so it is worth keeping alongside the positive one rather than replacing it.

**Finding 4 — diagnosis section.** Three entries: radiographic skeletal survey (with an explicit note that it places the patient in a wide category rather than distinguishing SEMD-ACAN), ACAN sequencing (positional mapping in family 1, exome plus Sanger confirmation in family 2), and the osteogenesis imperfecta misdiagnosis. I gave the misdiagnosis its own entry rather than a note because thirty-six years under the wrong label — in a disease whose recurrence risk is recessive rather than usually dominant — is a diagnostic fact about this disease, not a curiosity about one patient.

**Finding 5 — evidence_source misgrade.** The D2267 calcium-coordination sentence is now COMPUTATIONAL with `directness: INDIRECT`, and its explanation says the measurement that tests the prediction is the surface plasmon resonance result on the binding node.

**Two things I changed that you didn't ask for**, both consequences of the above:

- `NCIT:C38092` (Skeletal Survey X-Ray) is the obvious term for the radiographic entry and is *not* reachable from `NCIT:C25218`, so it would have failed the `TreatmentActionTerm` enum. Bound `NCIT:C17369` (Imaging Procedure) with `preferred_term: radiographic skeletal survey` instead — same pattern as the cochlear-implant device case in CLAUDE.md.
- `The acetabula were shallow.` is a real proposition but only four words, so `check-snippet-length` rejected it. Extended the quote to the contiguous preceding sentence rather than paraphrasing, and said so in the explanation.

Validation: schema, terms and references pass, 42/42 snippets verified, all ten offline gates green.

---

matentzn:
Merge conflict resolved in `630b353c6`. This PR had gone `dirty` against `main` while sitting approved, which is why the shepherd's sweep enqueued the other four PRs from this run and skipped this one.

Both conflicts were in derived cache files, and both were the same shape — `main` and this branch each added one row at the same sorted position:

| File | Conflict |
|---|---|
| `cache/mondo/terms.csv` | `MONDO:0013007` (main) vs `MONDO:0013014` (here) |
| `cache/enums/diseaseterm_f618932bb23f.csv` | the same pair |

Resolved as the union in CURIE order. I did not retype or hand-place any row — the ordering was confirmed by tooling afterwards: `just normalize-cache` reports no further change, and `check-term-cache-integrity` and `check-cache-order` both pass.

Revalidated against the merged tree rather than assuming the merge was inert: `just validate` clean (schema, terms, references, **42/42 snippets**), and all ten offline gates green, now against 2,956 KB files instead of 2,876. The three-dot diff against `origin/main` is unchanged at the intended 12 files — no unrelated churn came in with the merge.

Note that this push drops the approval, since `main` has `dismiss_stale_reviews`. That is unavoidable: the PR could not have merged in its conflicted state, so the fix had to happen regardless of what it cost. Ready for re-review.

---

PR #10386, opened 2026-09-01
curate: Add TFRC-related Combined Immunodeficiency (MONDO:0014760, TFRC)

### LINKED ISSUE #10380: Curate TFRC-related combined immunodeficiency (MONDO:0014760)
Curate a dismech entry for **TFRC-related combined immunodeficiency** ([MONDO:0014760](https://monarchinitiative.org/MONDO:0014760), immunodeficiency 46) — an autosomal recessive combined immunodeficiency caused by a homozygous *TFRC* variant in the transferrin receptor 1 internalisation motif, which blocks receptor endocytosis and starves proliferating lymphocytes of iron while sparing erythroid cells, producing hypogammaglobulinaemia with intact cell numbers but failed T- and B-cell proliferation.

Provisional read: this looks like a **DISEASE**. One causal gene (`hgnc:11763` TFRC), no MONDO descendants, and a single cell-biological lesion (defective TfR1 internalisation) that the reported phenotype follows from directly. The cell-type-selective escape of erythroid precursors is the mechanistically interesting part and worth modelling explicitly rather than flattening into "iron deficiency".

Stub: `stubs/TFRC-related_Combined_Immunodeficiency.yaml`

aleixpuigb:
New disease entry for TFRC-related combined immunodeficiency (immunodeficiency 46), an autosomal recessive inborn error of immunity in which transferrin receptor 1 is made and reaches the cell surface but cannot be pulled back in. Iron therefore never gets delivered into the cell, and the lymphocytes that need it most — the ones trying to proliferate after antigen stimulation — fail, while their numbers stay normal.

The entry is built around that distinction, because a lot of the disease follows from it rather than from "iron deficiency" generally: surface and soluble TfR1 are raised rather than reduced (a useful diagnostic pointer), and the anaemia does not respond to iron supplementation. The most interesting part is what is spared. Red cell precursors depend on this receptor more than lymphocytes do, yet patients have only mild anaemia. That escape gets its own node in the pathograph rather than being left as an absence of phenotype.

Closes #10380.

## Mechanism as curated

```
TFRC YTRF-motif variant (p.Tyr20His / p.Arg22Trp)
        │
        ▼
Defective TfR1 endocytosis, receptor accumulates on the surface
        │                                    │
        │                          ┌─────────┴──────────┐
        ▼                          ▼                    ▼
Impaired transferrin-bound    (biomarker) surface + soluble TfR1 ↑
iron delivery
        │
        ├──────────────────┬─────────────────────────┐
        ▼                  ▼                         ▼
Lymphocyte prolif.    Impaired iron supply to    Erythroid sparing
and activation        marrow progenitors         via STEAP3
failure                    │                          │
        │                  ▼                          ▼
        │            Neutropenia,                 Anemia (mild)
        │            thrombocytopenia
        ▼
Hypogammaglobulinaemia, raised IgM, impaired T cell function,
recurrent sinopulmonary infection → bronchiectasis,
chronic diarrhoea, failure to thrive, abscesses, meningitis
```

The cleanest piece of evidence in the entry is a rescue: adding iron citrate, which loads cells independently of the receptor, corrects lymphocyte proliferation, immunoglobulin secretion and class-switch transcripts in patient cells, and improves T cell proliferation in a knock-in mouse carrying the patients' own allele. That is what pins the defect to iron delivery rather than to some other receptor function.

## What is in the diff

- `kb/disorders/TFRC-Related_Combined_Immunodeficiency.yaml` — 6 pathophysiology nodes, 18 phenotypes, a `variants:` block for both alleles, STEAP3 as a `MODIFIER`, a `biochemical:` block, 4 treatments, 2 animal models, 4 discussions
- One openscientist deep-research report plus its citations sidecar and artifacts
- 9 reference-cache entries (all cited), the term-cache rows this entry introduced, two history records, and deletion of the stub

**Four discussions record where the evidence stops short.** The STEAP3 accessory-endocytosis hypothesis is the proposed explanation for the erythroid sparing, but the rescue supporting it was done in patient *fibroblasts*, not in erythroid cells — the experiment was run in the cell type the hypothesis is not about. The transplant series found dysmyelopoiesis in all three patients whose pre-transplant marrow was examined, plus one clonal cytogenetic abnormality, which sits awkwardly next to the erythroid sparing. And the Treg conditional knockout produces *autoimmunity* where patients have immunodeficiency — recorded as a `HUMAN_MODEL_MISMATCH`, with the one human Treg measurement in this literature keeping it open rather than settling it.

**The marrow-progenitor node argues against itself.** The HSC-knockout mouse it rests on spares thrombocytes — one of the two lineages affected in patients — so the thrombocytopenia edge is `INDIRECT_UNKNOWN_INTERMEDIATES` and the node description says why.

**Frequency bands come from a cohort of eight.** Several phenotypes are `OBLIGATE`. That means "all patients in the only published series", not an established penetrance figure, and each evidence explanation says so.

**Four phenotypes have no incoming edge on purpose** — optic atrophy, developmental delay, goitre, vitiligo. No cited source traces a route from the iron defect to any of them, and `notes:` says so. Everything else is reachable from the variant node.

**Two things from the cohort's tail are deliberately not phenotypes.** The HLH-like presentation and the Evans syndrome are composite clinical syndromes; binding either to one HPO term would lose what the source reports, so both are named in a discussion instead. Hepatitis B in two patients is not curated at all — an acquired infection in a transfused population, not attributed to the immunodeficiency by the source.

## Deep research

One openscientist run, 21 minutes, 9 of 9 references verified, confabulation rate 0.0. Used for leads and checked against the primary papers rather than copied — it listed `HP:0009098 Recurrent oral thrush` from the JAX set, which the primary source explicitly contradicts ("No oral thrush was reported"), so that was declined.

`just preflight-dr` flagged STEAP3 at 37% of TFRC mentions as a possible second disease entity. That is a false positive: STEAP3 is the proposed accessory mechanism *within* this disease. The OMIM 190010 it also surfaced is the TFRC gene MIM.

## Review history

Two rounds, both fully addressed. The first raised seven items — snippets, phenotype coverage, genetic slots, IVIG, the biomarker block, connectivity, uncited references — all fixed. The second caught something I got wrong twice: evidence explanations that blamed the reference cache for truncations the cache did not cause. Both were false, and one of them left the entry's headline quantitative claim (13-fold and 7-fold surface TfR1) uncited.

Sweeping all four entries in this claim run afterwards found a third instance in #10389 and a fourth here. That is filed as #10388 with a working prototype for a structural check — it looks at where the quote stops in the cached bytes rather than at what the explanation says, which is the only approach that generalises, since the three false explanations shared no keyword.

## Validation

`just validate` passes — schema, terms, and 89/89 evidence snippets verified against cached references. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-folded-hyphens` all clean. Full `tests/test_data.py` passes.


aleixpuigb:
All seven IMPORTANT items are done, plus the four suggestions. Nothing was argued down — the review was right on every point, and item 1 was a real error on my part rather than a style disagreement.

**Item 1 — the truncated snippets.** You were right and my explanation was wrong. The cache is fine; my extraction had split the sentences on the period inside `p.Tyr20His`, and I then wrote that up as a cache defect instead of checking. Both snippets now run to the end of the sentence, including the conclusion clause, and the explanatory alibi is gone.

**Item 2 — the cohort full text.** Curated the per-patient table, not just the abstract: impaired T cell function 8/8 (`HP:0011840`), recurrent epistaxis and petechiae 7/8, bronchiectasis 3/8, skin abscesses 2/8, meningitis and multinodular goitre 1/8 each. On the selection question you raised: two items in the tail are deliberately not phenotypes. The HLH-like presentation and the Evans syndrome are composite clinical syndromes, and binding either to one HPO term would lose what the source reports; both are now named in a discussion instead. Hepatitis B in two patients is not curated at all — it is an acquired infection in a transfused population and the source does not attribute it to the immunodeficiency. `notes:` says all of this.

On `HP:0002721` vs `HP:0011840` for impaired T cell function: I went with `HP:0011840` (Abnormal T cell physiology). It is the more specific term for what the cohort actually measured, and `HP:0002721` (Immunodeficiency) is already carried by the disease term itself. Happy to switch if you want alignment with the JAX set over specificity.

**Item 3 — genetic slots.** Added a `variants:` list for both alleles with rsIDs, ClinVar accessions, `type`, `clinical_significance` and `functional_effects`; p.Arg22Trp is `LIKELY_PATHOGENIC` rather than `PATHOGENIC` since it rests on one patient plus a functional assay. STEAP3 is now a `MODIFIER` entry in `genetic:`, with a note that the claim is about lineage rather than about inter-individual severity. `Inheritance.penetrance: COMPLETE`, with the caveat stated in the description: ascertainment has been entirely through symptomatic probands, so unaffected homozygotes would not have been found if they existed. Consanguinity quote added.

**Item 4 — IVIG.** Now its own treatment. `NCIT:C121331` (Intravenous Immunoglobulin Therapy) does validate as a `treatment_term`, so no fallback to a generic action term was needed; agent is `NCIT:C2701`, modality `PROTEIN_REPLACEMENT`, targeting the hypogammaglobulinaemia.

**Item 5 — biomarker.** Added a `biochemical:` block for surface and soluble TfR1 with a `READOUT_OF` link to the endocytosis node, carrying the 13-fold/7-fold T and B cell figures and the 6-fold fibroblast figure.

**Item 6 — connectivity.** Added a new node, `Impaired Iron Supply to Proliferating Marrow Progenitors`, so the neutropenia and thrombocytopenia are no longer floating, plus edges to `Increased circulating IgM level` and the new `Impaired T cell function`. One thing worth flagging: the mouse that node rests on (PMID:31601687, HSC-specific Tfr1 knockout) spares thrombocytes and B lymphocytes among the progenitor lineages, which does not match the patients' thrombocytopenia. I have left that edge as `INDIRECT_UNKNOWN_INTERMEDIATES` and said so in the node description rather than papering over it.

**Item 7 — uncited references.** All four now cited. PMID:32284314 and PMID:7957580 give the CD71-iron dependency independently, in healthy human T cells and in fetal thymus organ culture with the receptor blocked rather than mutated — that is useful precisely because it is not TFRC-mutant evidence. PMID:31601687 anchors the new marrow node. PMID:41714512 is on the iron-delivery node.

**Suggestions.** Erythroid sparing rescaled to `TISSUE`. Optic atrophy now carries a `notes:` line saying it is curated because it recurs in the JAX annotation set, not because causation is established at 1/8. The Treg conditional knockout got the `HUMAN_MODEL_MISMATCH` discussion you suggested — it is the sharper mismatch in this entry, since the model's phenotype points the opposite way (autoimmunity vs immunodeficiency), and the one human Treg measurement in this literature shows Tregs *are* reduced in a patient, which keeps the question live rather than settling it. `datasets:` left empty.

**GeneReviews:** I could not find a chapter either, and none of the 25 papers in the deep-research run cites one. Given ~12 reported patients since 2016 I do not believe one exists.

Validation after the changes: 82/82 snippets verified, and `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-folded-hyphens` are all clean.

One thing for the toolchain, from your note that `just check-source-defect-claims` produced no TFRC output: the phrasing it missed was "the cached text breaks the sentence there" attached to a snippet that stops mid-sentence. That shape — a truncated snippet whose `explanation` blames the cache — looks mechanically detectable, and it is exactly the failure mode where a curator's own alibi hides the defect. I'll open an issue for it separately rather than widen the gate from inside a curation PR.


---

aleixpuigb:
Both fixed, and you were right that it is not cosmetic — the 13-fold and 7-fold figures were asserted in the `context` field with the citation stopping one clause short of them, so the entry's headline quantitative claim was uncited.

The apostrophe explanation was wrong in the same way as the last one. U+2019 validates fine in snippets; I checked the two examples you named and they do. Both snippets now run to the end of the clause and both explanatory alibis are gone:

- surface TfR1 → `...expressed on a large percentage of patients’ T and B cells, at levels 13- and 7-fold higher, respectively, than in controls`
- iron citrate → `...corrected the patients’ lymphocyte proliferation, IgE secretion, and Iμ-Cε transcript expression`

Also took the suggestion on `animal_models#Mouse`. You are right about which model it was resolving to — a model-mismatch discussion pointing at the `RECAPITULATES`/`HIGH` knock-in is the opposite of what it means. It now names the Treg conditional knockout explicitly.

**Your framing was more useful than the fix.** After your comment I swept all four entries in this claim run for the pattern rather than just this file, and found a **third** instance, in the SHORT syndrome PR (#10389): an explanation claiming "the cached text has lost the inequality signs" from `aged ≥12 years` / `aged ≤10 years`. The cache has both signs. My snippet had dropped them and I wrote that up as a cache defect instead of looking. Fixed there too.

Three instances, three different false explanations — "the cached text breaks the sentence there", "the typographic apostrophe does not round-trip through YAML", "the cached text has lost the inequality signs" — and no keyword in common. That is the argument against extending `check-source-defect-claims`' word list, and I have made it the core of the issue you asked me to open: [#10388](https://github.com/monarch-initiative/dismech/issues/10388).

It proposes the structural check instead of a prose check: locate the snippet in the cached bytes and flag when the match is immediately followed by a letter. Prototyped over ~250 snippets in these four entries — 2 findings at the pre-fix commit here, both real, 0 false positives after the fix. The naive version also fires on quotes ending before flattened superscript citation markers (`patients6`, `proliferation8,9`), so restricting the following-character test to `[A-Za-z]` rather than `\w` matters; that refinement is in the issue with the numbers.

That check would have caught all three of mine at first commit, without knowing anything about what I wrote in the explanation. I offered to send it as a PR once someone decides whether it lands as a gate or a report-only census, which depends on what it turns up over all of `kb/` — I did not want to make that call from inside a curation PR.

Thanks also for confirming HP:0011840 over HP:0002721 and closing out GeneReviews.


---

Issue #10388, opened 2026-09-01
Snippets that stop inside a token are invisible to every gate, and the alibi for them is invisible too


aleixpuigb:
Two evidence snippets in a curation PR of mine stopped mid-sentence, and both carried an `explanation` saying the cache had truncated them. The cache had not. Every gate passed, `just check-source-defect-claims` produced no output, and it took a human-grade read of the cached file to find it. The reviewer on #10386 caught it; nothing in the toolchain could have.

This is the failure #9226 described, one round later: the gate it produced ships, and the same defect walked past it.

## What the snippets looked like

```yaml
snippet: >-
  The TfR1-holotransferrin complex is internalized by receptor-mediated endocytosis,
  which requires an aromatic residue at the p.
explanation: >-
  The structural requirement of the motif that the variant violates. The quote is cut
  at the residue designation because the cached text breaks the sentence there.
```

The cached text (`references_cache/PMID_26642240.md`) reads `...an aromatic residue at the p.Y20 position mutated in the patients`. The sentence is whole. My extraction had split it on the period inside `p.Y20`, and I then wrote that up as a property of the cache instead of checking.

The quote validates because `"...at the p."` genuinely is a substring of the cached text. That is the same property that made #9226's mid-word fragments invisible, in a new shape.

## Why `check-source-defect-claims` did not fire

`DEFECT_WORD` in `scripts/check_source_defect_claims.py` matches:

```
truncat* | mid-word | garbled | stops where | cut-off | mangled | corrupted
```

My phrasing was "the cached text **breaks the sentence** there". No overlap. This is not a bug in the gate — it adjudicates claims it can recognise, and it recognised none here. It is a reminder that a keyword list over free prose is a leaky way to catch a claim that has unbounded paraphrases. There is no finite word list for "the source is at fault".

## The other half of #9226 is the one that generalises, and it was never built

#9226 proposed pairing the prose gate with a structural one: *"a snippet whose match in the cache is flanked by word characters is a defect with no legitimate form."* That half does not exist — no such logic is in `check_snippet_length.py` or anywhere else.

The strict formulation (word character on **both** sides of the match boundary) would not have caught my case, because the snippet ends on a period. The signature that does catch it is narrower and, I think, cleaner:

> **The snippet's match in the cached text is immediately followed by a letter, with no intervening whitespace.**

A quote that ends at a real boundary is followed by a space, a newline, or end-of-text. One followed directly by a letter stopped inside a token.

## Prototype and results

[...]

PR #10389, opened 2026-09-01
curate: Add SHORT Syndrome (MONDO:0010026, PIK3R1)

### LINKED ISSUE #10377: Curate SHORT syndrome (MONDO:0010026)
Curate a dismech entry for **SHORT syndrome** ([MONDO:0010026](https://monarchinitiative.org/MONDO:0010026)) — a PIK3R1-related multisystem disorder combining short stature, partial lipodystrophy, Rieger (anterior chamber) anomaly, and severe insulin resistance that is characteristically dissociated from dyslipidaemia and fatty liver.

Provisional read: this looks like a **DISEASE**. MONDO records no descendants, and the stub names no causal gene, but the literature is consistent on a single locus — heterozygous *PIK3R1* variants clustered in the C-terminal SH2 domain (recurrently p.Arg649Trp).

Note the entity is *not* the same as the existing `Activated_PI3K-delta_Syndrome` entry, which also involves *PIK3R1*: APDS2 arises from inter-SH2/splice variants that **increase** p110δ activity, whereas SHORT syndrome arises from cSH2 variants that **impair** insulin-stimulated PI3K/AKT signalling. Same gene, opposite direction of effect, different pathograph. The curation should make that distinction explicit rather than reusing the APDS nodes. MONDO's `MONDO:1060136` ("PIK3R1-related immunodeficiency and SHORT syndrome") is the parent that lumps the two; the entry should not inherit that lumping.

Stub: `stubs/SHORT_Syndrome.yaml`

aleixpuigb:
New disease entry for SHORT syndrome, the PIK3R1-related disorder whose acronym names short stature, joint hyperextensibility, ocular depression, Rieger anomaly and teething delay. One heterozygous variant in the C-terminal SH2 domain of p85α stops PI3K being recruited to IRS-1 after insulin binds, and three apparently unrelated arms of the phenotype — growth failure, partial lipodystrophy, and severe insulin resistance — all fall out of that single step.

The most useful thing in the entry is probably a negative. Insulin resistance from obesity comes with fatty liver and dyslipidaemia; insulin resistance here does not. Four of five studied patients had extreme insulin resistance with neither, which puts SHORT syndrome with insulin-receptor dysfunction rather than with AKT2 dysfunction, and tells you where in the cascade the lesion sits. That is diagnostically usable at the bedside, so both hepatic steatosis and dyslipidaemia are curated as phenotypes carrying `REFUTE` evidence rather than being silently omitted.

Closes #10377.

## Mechanism as curated

```
PIK3R1 cSH2 variant (p.Arg649Trp hotspot)
        │
        ▼
Impaired p85α binding to IRS-1
        │
        ▼
Reduced PI3K-AKT signal transduction ────────────┬──────────────┬─────────────┐
        │                    │                   │              │             │
        ▼                    ▼                   ▼              ▼             ▼
Post-receptor          Impaired adipocyte   Impaired growth  Craniofacial  Anterior segment
insulin resistance     differentiation      signalling       and dental    dysgenesis
        │                    │                   │           development        │
        ▼                    ▼                   ▼                │              ▼
Diabetes, PCOS,        Lipoatrophy,        IUGR, short       (8 features)   Rieger anomaly,
+ dissociation from    facial gestalt       stature                          glaucoma
  dyslipidaemia
```

## What is in the diff

- `kb/disorders/SHORT_Syndrome.yaml` — 9 pathophysiology nodes, 25 phenotypes, a variants block, animal model, treatments, progression, 3 discussions
- One openscientist deep-research report with citations sidecar and artifacts
- 19 reference-cache entries, the term-cache rows this entry introduced, a history record, stub deleted

**The deep-research report gave a wrong MONDO ID, and it resolves.** It lists `MONDO:0009159` as this disease's identifier, in its key-identifiers table and again in the body. `MONDO:0009159` is *Ehlers-Danlos syndrome, cardiac valvular type*. The entry binds `MONDO:0010026`. This is issue #9888 happening — the recipe sends an empty `mondo_id`, so the provider filled it in — and it is the expensive shape, because a CURIE-existence check passes and only a label check catches it. `just validate-research-terms` does catch it, and I have reported the instance on #9888. Nothing else in the report was wrong.

**The report also arrived without any validation frontmatter**, unlike the TFRC report from the same batch, same recipe, same provider. I added both sections by hand. Reference validation then came back 19/19 resolved, 19/19 on topic, 0 unresolved. Flagging in case reports landing unvalidated is systematic rather than a one-off.

**One tension is recorded rather than resolved.** The pathograph puts everything through a single reduced-PI3K-AKT node. The 2016 dissociation study says these mutations impair insulin signalling *only in some cellular contexts*, and reports one patient's cells in which mutant p85α failed to bind IRS1 while downstream signalling was normal. That is the strongest challenge to the linear reading, and it is in a `discussions` entry rather than smoothed over. I kept the node single because nothing cited identifies what makes a tissue sensitive, and splitting on speculation seemed worse.

**Four phenotypes are deliberately unconnected** — joint hypermobility, inguinal hernia, sensorineural hearing impairment, speech delay. These are exactly the features the 32-patient cohort grades as "less specific, or minor", and no source proposes a route from the signalling defect to any of them. Every other phenotype is reachable from the variant node.

**Growth hormone carries a REFUTE item.** It has long been treated as contraindicated because it is diabetogenic. The cited review's age-stratified counts say otherwise: insulin-resistant diabetes appeared in 10 of 15 GH-untreated patients aged ≥12, and in none of three treated or six untreated patients aged ≤10. That is an age effect. The case against growth hormone here rests on poor statural response, not on causing the diabetes, and the entry says so.

**Kept separate from `Activated_PI3K-delta_Syndrome`.** Same gene, opposite direction of effect, different variant class. MONDO's `MONDO:1060136` lumps the two; that lumping is not adopted, and the overlap is recorded as an open question instead.

## Validation

`just validate` passes — schema, terms, and 111/111 evidence snippets verified against cached references. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-folded-hyphens` all clean.

---


aleixpuigb:
Pushed a correction before review, so it does not have to be found: one evidence explanation claimed the cached text had lost the inequality signs from `aged ≥12 years` / `aged ≤10 years`. It has not — the cache carries both. My snippet had dropped them and I wrote that up as a source defect instead of looking.

It mattered here. Without the inequalities the counts are ambiguous; with them they are the whole argument the `progression` entry makes — every affected patient was 12 or older, and none of the nine aged 10 or under had developed diabetes regardless of growth hormone exposure. That is what separates an age effect from a treatment effect, and it is why growth hormone carries a `REFUTE` item rather than being listed as the cause.

Found by sweeping all four entries in this claim run after the reviewer caught the same pattern twice on #10386. Three instances in total, three different false explanations, no shared keyword — which is the argument in #10388 for checking the cached bytes rather than the prose.

111/111 snippets still verify.


---

aleixpuigb:
All five items done, plus both suggestions. The CRITICAL one was a factual error on my part, and a bad one — thank you for catching it.

**PMID:28632845.** You are right that the entry states the opposite of what that paper says. My `limitations` text claimed the ocular component is not reported for the R649W knock-in, and the paper reporting it was cached in this very PR. Three changes:

- `limitations` corrected — the *dental* component is unreported; the ocular one is reported and now has its own link.
- A fourth `modeled_mechanisms` link on that model: `RECAPITULATES` against `Anterior Segment Dysgenesis`, with OCT iris thickness and width as the readout. The ocular arm of the pathograph had no model attached; it does now.
- Four ocular phenotypes added — hypoplasia of the iris, abnormal pupil morphology, anterior synechiae, cataract.

On the discussion rewrite: your framing was the useful part. The gap is not "no mechanism is known" but something narrower and better-posed. Sufficiency is now established — one heterozygous allele produces iris hypoplasia in vivo in both species — and what is missing is which neural crest population, which developmental window, and why the mouse stops at the iris while the humans go on to goniosynechiae, cataract and glaucoma. That species difference is in the paper and nobody appears to have followed it up; the rewritten rationale says so.

**APDS2 overlap papers.** Cited, and you were right that it strengthens rather than weakens the split. The discussion now makes the argument explicit instead of leaving it implicit: the overlap alleles are not scattered, they cluster at `c.1425+1G>C/A/T` and at frameshifts near it, which is a different part of the gene from the cSH2 domain where the classic SHORT alleles sit. So the overlap looks like a property of a variant class rather than evidence the diseases are one. What would settle it — a mechanistic account of how one splice allele produces both reduced PI3K-AKT signalling and increased p110δ activity — is what no source provides, and that is now what the open question asks.

PMID:35789397 also gave the only penetrance observation in this literature, so `Inheritance.penetrance: INCOMPLETE`, with the caveat stated: one family, and a variant at the APDS2 end of the spectrum rather than a classic cSH2 allele, so not a penetrance figure for p.Arg649Trp.

**Silver-Russell.** Added. You are right it is the clinically important one — IUGR plus short stature plus triangular facies fits both, and SRS is far commoner.

**Treatments.** All three cached case reports consumed, as a `Multi-agent oral antidiabetic therapy` entry citing PMID:41459015 with PMID:33742773 and PMID:39735640 corroborating, plus PMID:32602265. The GLP-1 agonist gets a `target_mechanisms` link to the `Impaired Insulin Secretion and Incretin Response` node — that node had no treatment attached, and an incretin-directed agent in a disease with a documented incretin defect is the natural join. Marked `INDIRECT`: the incretin defect is a mouse finding and the case report does not isolate that agent's contribution from the other four.

**GeneReviews.** I have network egress where your runner did not, so I ran your query. `SHORT syndrome[TI] AND GeneReviews[TI]` returns zero records — no chapter exists. Recorded in `notes` so it does not have to be re-asked, along with a `references:` block carrying the five most load-bearing sources.

**Both suggestions taken.** PMID:36401775 now cites three specific candidate routes for the growth arm (IGF1 resistance, increased energy expenditure, decreased lipogenesis) in place of the pathway's generic role, with the review's own "possible" hedge on the IGF1 arm retained. PMID:35110500 cites the formal nosological placement behind the `Syndromic insulin resistance` parent.

**On consumption**, which was your framing for the whole review and is the right one: 13 of 24 cited → 23 of 24. The last is PMID:24033310, and you had already worked out why — its cached body is a bare three-citation list with no abstract. `notes` now records that rather than leaving it as an unexplained gap.

134/134 snippets verify. All offline gates clean. I also ran the snippet-boundary prototype from #10388 over the file: zero findings.


---

Issue #10396, opened 2026-09-01
A 5-second OLS timeout on one CURIE throws away the validation of a 45-minute deep-research run


aleixpuigb:
`just research-disorder` exited 3 on me after the research had already completed and the report had already been written to disk. The cause was a single transient network read timeout:

```
ERROR - Term validation failed: could not reach ontology service to resolve HP:0000589:
HTTPSConnectionPool(host='www.ebi.ac.uk', port=443): Read timed out. (read timeout=5)
error: recipe `research-disorder` failed with exit code 3
```

45 minutes of provider time, a finished 27-citation report on disk, and the recipe reports failure. Re-running `just validate-research-terms` on the same file a few minutes later succeeded on the first attempt, so the timeout was transient.

## Why this matters more than a spurious non-zero exit

The abort happens **before the report gets either of its validation sections**. The reference-validation block and the term-validation block are both missing, so what lands on disk is a report that reads as complete and carries no indication that nothing checked it.

That is the worst possible pairing with #9888. In this same run, the provider had invented `MONDO:0012997` for SRD5A3-CDG (that ID is *cholestasis-pigmentary retinopathy-cleft palate syndrome*; the correct term is `MONDO:0012885`). Running the term validator by hand caught it immediately:

```
WARNING - MONDO:0012997 is 'cholestasis-pigmentary retinopathy-cleft palate syndrome',
          but the report calls it 'congenital disorder of glycosylation, type Iq'
```

So the check works. It just did not run, because of a five-second socket timeout on an unrelated CURIE. A curator who takes the report at face value gets a wrong disease identifier with nothing flagging it.

I hit this on two of three completed runs today (SHORT syndrome and SRD5A3-CDG; the TFRC run got both sections). Both unvalidated reports turned out to contain an invented MONDO ID.

## Three things worth separating

**1. A validation failure should not discard the validation that did succeed.** Reference validation and term validation are independent. Here the term stage failed and the reference block was lost too. Each stage should write its own section, and a stage that fails should say so *in the report* — a `## Term Validation` section reading "could not complete: ontology service unreachable" is far better than no section, because it is visible to the curator rather than being an absence they have to notice.

**2. A transient network error is not a validation result.** There is a real difference between "this CURIE does not resolve" and "I could not reach the server". The first is a finding about the report; the second is a fact about the network. Conflating them means a blip reads as a clean bill of health once the section is simply missing. Retry with backoff before giving up, and distinguish the two outcomes in whatever is written.

[...]

PR #10397, opened 2026-09-01
curate: Add SRD5A3-Congenital Disorder of Glycosylation (MONDO:0012885)

### LINKED ISSUE #10378: Curate SRD5A3-congenital disorder of glycosylation (MONDO:0012885)
Curate a dismech entry for **SRD5A3-congenital disorder of glycosylation** ([MONDO:0012885](https://monarchinitiative.org/MONDO:0012885)) — an autosomal recessive type I CDG (CDG-Iq) in which loss of polyprenol reductase blocks conversion of polyprenol to dolichol, starving the dolichol-linked oligosaccharide pathway and producing ocular coloboma, nystagmus, cerebellar/vermis hypoplasia, intellectual disability, ichthyosiform skin change, and coagulation abnormalities.

Provisional read: this looks like a **DISEASE**. One causal gene (`hgnc:25812` SRD5A3), no MONDO descendants, and a single well-defined biochemical lesion upstream of N-glycosylation.

Stub: `stubs/SRD5A3-congenital_Disorder_Of_Glycosylation.yaml`

aleixpuigb:
New disease entry for SRD5A3-CDG (formerly CDG-Iq), an autosomal recessive glycosylation disorder in which loss of polyprenol reductase blocks the conversion of polyprenol to dolichol — the lipid carrier that every N-linked glycan is assembled on. Because the block sits at the very start of the pathway, one enzyme lesion hypoglycosylates a wide range of secreted proteins.

The mechanism is the interesting part, and it is not what you would guess. Patient fibroblasts carrying a complete enzyme null still contain **normal amounts of dolichol** while accumulating unreduced polyprenol. Two independent groups measured this. So the pathogenic quantity is most likely the raised polyprenol-to-dolichol *ratio* — the excess precursor competing with dolichol at glycan initiation without being able to carry a glycan — rather than a shortage of dolichol, which the same authors note would probably be lethal. The pathograph is built that way, and the alternative dolichol route this implies is recorded as an open question, because it is the load-bearing assumption of the whole model and nobody has identified it.

Closes #10378.

## Mechanism as curated

```
SRD5A3 biallelic loss of function (polyprenol reductase)
        │
        ▼
Polyprenol accumulates; polyprenol-to-dolichol ratio rises
(dolichol itself stays normal — an alternative route exists)
        │
        ▼
Impaired dolichol-linked oligosaccharide assembly
        │
        ▼
Systemic protein hypoglycosylation  ──┬──────────┬─────────────┬──────────┐
        │                             ▼          ▼             ▼          ▼
        │                        transferrin  protein C   antithrombin  skin
        ▼                        pattern      ↓           III ↓
Cerebellar and optic pathway vulnerability          Progressive ocular
        │                                            and skeletal
        ▼                                            degeneration
Cognitive delay, nystagmus, optic atrophy,                │
optic nerve hypoplasia, ataxia, vermis hypoplasia         ▼
                                                    Cataract, retinitis
                                                    pigmentosa, kyphosis
```

## A drug result that argues for the mechanism

Atorvastatin — an HMG-CoA reductase inhibitor, so it *restricts* isoprenoid supply upstream of polyprenol — rescues a C. elegans W19X model and restores the polyprenol-to-dolichol ratio in patient fibroblasts. That is a counter-intuitive treatment for a biosynthetic deficiency, and it makes sense only if the ratio is what matters rather than the dolichol level. It is curated as **explicitly experimental**: no patient has received it, and no trial exists. It is in the entry because the reasoning is mechanistically informative, not because it is care.

## What is in the diff

- `kb/disorders/SRD5A3-Congenital_Disorder_of_Glycosylation.yaml` — 6 pathophysiology nodes, 18 phenotypes (zero unconnected), a `variants:` block, a mechanistic hypothesis, an experimental model, treatments, 3 discussions
- One openscientist deep-research report with citations sidecar and artifacts
- 12 reference-cache entries, term-cache rows, a history record, stub deleted

**A missing mouse model, curated as a finding rather than as silence.** There is no vertebrate model of this disease. The one published mouse lacking *Srd5a3* lacks eight neighbouring genes too, and its peri-implantation lethality was traced to *Exoc1* — confirmed by making an *Exoc1* knockout separately and by showing a CRISPR deletion of the same interval without the *Exoc1* defect is not lethal. Reading that mouse as an *Srd5a3* phenotype would be wrong, so the entry carries `REFUTE`-graded evidence against exactly that reading. It matters because the disorder's defining features are ocular and cerebellar, and neither fibroblasts nor a worm can address them.

**Tissue selectivity is the open question.** Serum glycoproteomics shows hypoglycosylation is broad — 245 of 291 altered glycopeptides decreased — yet the clinical picture is dominated by two tissues. The one proposed explanation (particular optic-nerve sensitivity to polyprenol, supported by the contrast with DOLK-CDG one step downstream, which shares the skin and seizure phenotype and lacks the eye involvement) is a comparison between two small case literatures, not an experiment. The edge from hypoglycosylation to that selectivity is `INDIRECT_UNKNOWN_INTERMEDIATES`.

**Genotype does not fix phenotype, at two levels.** Siblings homozygous for the same nonsense allele differ in whether they have structural eye anomalies. And homozygosity for the recurrent p.Gln96delinsX founder allele is not consistently associated with coloboma. Also worth a sequencing note: one pair of twins was left undiagnosed by standard exome analysis and resolved only by copy-number analysis, which found an intragenic tandem duplication of exons 2–4.

## Deep research, and two toolchain problems it surfaced

The run completed (45 min, 27 citations, 27/27 references resolved, 0 unresolved) but **the recipe exited 3 before writing either validation section**, because term validation hit a 5-second read timeout against EBI OLS on one CURIE. The report landed looking complete and unchecked. I ran both validators by hand; the term validation succeeded on the first retry, so the timeout was transient. Filed as #10396.

That mattered, because the report gave `MONDO:0012997` as this disease's identifier — labelled with a genuine SRD5A3-CDG synonym but pointing at *cholestasis-pigmentary retinopathy-cleft palate syndrome*. The entry binds `MONDO:0012885`. This is the second instance in this claim run of the empty-`mondo_id` problem in #9888 producing an invented ID that resolves to a different disease; both are reported there.

Everything else in the report checked out against the primary papers, and it contributed the C. elegans model, the atorvastatin result, the current case count, the founder allele, and the mouse-model explanation.

## Validation

`just validate` passes — schema, terms, and 87/87 evidence snippets verified against cached references. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-folded-hyphens` all clean. The snippet-boundary prototype from #10388 returns zero findings.

---


aleixpuigb:
Both CRITICAL findings are fixed, and all six IMPORTANT items are addressed. One correction to CRITICAL 2, explained below — I think the pathway is still unidentified, and the paper you point to is what rules out the obvious candidate.

Snippets go from 87 to 128, all verified. Cited references from 9 to 18. Pathograph still one connected component, now 20 phenotypes with no orphans.

## CRITICAL 1 — retinal dystrophy

Fixed as asked, plus the term change you suggested.

Onset is now `CHILDHOOD` and the band `FREQUENT`, citing PMID:28253385 (7 patients, 4 families, all homozygous for this entry's own `p.Trp19Ter`) and PMID:31638560.

On HP:0000510 vs HP:0000556 — you were right, and for a reason slightly different from the ERG. Only one of the five sources names the specific entity retinitis pigmentosa, in one adult patient. The rest say "retinal dystrophy" or "early-childhood-onset retinal dystrophy" and never establish the rod-before-cone sequence that rod-cone dystrophy asserts. So it is now bound to HP:0000556 and the phenotype renamed to match, with the three references to it updated. The description says why.

I also reframed the `Progressive Ocular and Skeletal Degeneration` node rather than just moving the phenotype, because the node's own text said the retinal degeneration emerged "in adulthood, from the second decade onwards". The distinction that fixes it is progression versus age at recognition: the retinal component is progressive but not late-onset, whereas cataract and kyphosis genuinely are documented as absent in childhood and present in the same patients as adults. The edge now carries a REFUTE item quoting "it appears to be a consistent primary presenting feature" against the late-onset reading.

## CRITICAL 2 — I think DHRSX is excluded as the answer, not the answer

I've cited PMID:42201967 in both places you asked, and the entry now records that the one-step model is superseded. But I don't think the open question closes, and I'd rather say so than quietly narrow it.

The question is: what makes dolichol in cells with **no functional SRD5A3**? DHRSX would answer it only if DHRSX were a route around SRD5A3. The same paper shows it is not:

> Deletion of TDA5 caused glycosylation defects, reduced dolichol levels, and accumulated polyprenol.

Losing the DHRSX counterpart reproduces this disease's own biochemistry — low dolichol, polyprenol piling up. Two enzymes in *parallel* routes cannot both do that; it is what you see when they are in series. And the rescue is one-directional:

> All these phenotypes were rescued by expression of DHRSX, but not by DFG10 or SRD5A3.

SRD5A3 cannot stand in for DHRSX. Nothing here shows the reverse, and the series result argues against it.

So the revision genuinely matters — the pathway has three steps, not one, and the entry now says so on the pathophysiology node — but it makes the residual-dolichol observation harder to explain rather than easier, because it removes the most plausible candidate. I've written it up that way in `alternative_dolichol_pathway`, with the TDA5 deletion cited as SUPPORT for DHRSX being a plausible candidate and the rescue cited as REFUTE against it actually being one.

Happy to be argued out of this if you read the rescue differently.

## IMPORTANT

**Item 3 — worm model section.** Fixed, and your grep was right. I counted it across the KB: 9 C. elegans models in `animal_models:`, 2 in `experimental_models:` — this entry and `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_2` (which has one in each). Moved with `species` and `genotype`, and `publication`.

**Item 4 — evidence_source.** Both worm items regraded to `MODEL_ORGANISM` and the incorrect rationale deleted. You were right that I invented a vertebrate restriction that CLAUDE.md does not contain.

The mixed atorvastatin sentence is split in both places it appears:
- `Atorvastatin rescued disease-relevant phenotypes in the worm model` → `MODEL_ORGANISM`
- `restored polyprenol-to-dolichol ratios in patient fibroblasts` → `IN_VITRO`

**Item 5 — the two notes claims.** Both wrong, both fixed.

"No source gives per-feature percentages" is false — PMID:28253385 gives 7/7, 5/5, 6/7. The note now says they exist, quotes them in the relevant evidence explanations, and explains why they are still not used to set bands arithmetically: every patient in that series carries the same allele and every one was ascertained through an eye clinic, so the proportions describe the ascertainment. They did drive the retinal band up, and the note says that.

The `animal_models:`/vertebrate conflation is gone. The vertebrate gap is real and stays recorded (the Srd5a3 mouse also lacks eight neighbouring genes, and its lethality traces to Exoc1); it just isn't the reason there was no section.

**Item 6 — reference consumption: 9 → 18.** Every reference in your table is now used, and each one earns its place rather than being cited to clear the count:

| PMID | What it changed |
|---|---|
| 28253385 | CRITICAL 1 |
| 42201967 | CRITICAL 2 |
| 31638560 | Myopia phenotype; the `c.436G>A` variant; a retinal dystrophy missed for years until dedicated imaging |
| 20700148 | Grounds the previously uncited Kahrizi synonym, on Coloboma, Cataract and Kyphosis, plus the `c.203dupC` allele and the allelic-disorders conclusion |
| 41769439 | Cited **REFUTE** on Cerebellar vermis hypoplasia — normal MRI at 20 months in both twins despite marked neurological signs. Adds a timing dimension nothing else in the entry had: a normal early scan does not exclude the diagnosis |
| 42472049 | Cited **REFUTE** on the tissue-vulnerability node, against the cheapest explanation for the selectivity — that affected tissues express SRD5A3 most. Across twelve CDG genes that does not hold. Also SUPPORT for the tissue-specific allelic imbalance it does find for SRD5A3 |
| 26628744 | New `dolichol_supplementation` discussion |
| 41667393 | The `c.509A>G` allele and a new `corticospinal_involvement` discussion |
| 39236565 | Context on the supportive-care treatment entry |

Two notes on that table. 39236565 was already on `main` from an unrelated PR rather than added here, so the denominator is 25, not 26. And I deliberately did **not** curate the limb-girdle weakness, hyperreflexia and extensor plantar responses from 41667393 as phenotypes: it is one patient, upper and lower motor neuron signs in a child with marked hypotonia have several readings, and the investigations that would separate a real corticospinal component from an incidental one are not reported. Three new phenotypes off one case would sit at the same weight as findings established across cohorts. It is recorded as an open knowledge gap with the authors' own framing quoted, and the variant description points at it.

**Item 7 — biochemical block.** Added, with the polyprenol-to-dolichol ratio as the marker and two readouts: `READOUT_OF` the ratio node (`DIAGNOSTIC`) and `PHARMACODYNAMIC_MARKER_OF` the enzyme node, since correcting the ratio is the endpoint the one candidate therapy was assessed against.

The `context` field records something worth knowing: NCIT has no term for either lipid. Searching it returns the reductase protein (NCIT:C105163) and the SRD5A3 allele, not the analytes. So `biomarker_term` carries a free-text `preferred_term` and no `term:` rather than a wrong binding.

Two diagnosis entries added as well — retinal electrophysiology and imaging (`NCIT:C162465`, with a note that NCIT has no electroretinography term under C25218), and intragenic copy-number analysis, so that a single heterozygous variant in a compatible patient is not read as excluding the diagnosis. The existing transferrin entry is now bound to `NCIT:C101016`.

**Item 8 — GeneReviews.** Ran it. No chapter exists.

```
SRD5A3[TI] GeneReviews[TI]        → 0 hits
SRD5A3 AND GeneReviews[book]      → 1 hit: PMID:20301507
```

PMID:20301507 is "Congenital Disorders of N-Linked Glycosylation and Multiple Pathway Overview", whose own title ends "RETIRED CHAPTER, FOR HISTORICAL REFERENCE ONLY". So no `references:` block and nothing to tag. Recorded in `notes:` so the next reviewer doesn't re-ask.

## Suggestions

Myopia and Nyctalopia added and wired into the ocular node. Optic disc coloboma is already covered by the existing Iris coloboma entry, whose evidence quotes "colobomas of the optic nerve and iris" — splitting it would need a source that separates them.

Rehabilitation and orthopaedic management I did **not** add. They are named only in unsourced review prose; no cited source documents either in an SRD5A3-CDG patient. Rather than manufacture a treatment entry, I recorded the position in the supportive-care description and cited PMID:39236565 for the general CDG picture.

Thanks for the flag on `SRD5A3 (HGNC:24420)` in the report at `:618` — noting it here for anyone reusing that artifact. The correct ID is `hgnc:25812`, which is what the YAML uses. That is the second identifier this report got wrong; the MONDO ID it carried was also wrong and was corrected before curation.

---

aleixpuigb:
Thanks — and thanks for re-reading PMID:42201967 rather than restating the finding.

Recording the three prose items rather than pushing them, because `dismiss_stale_reviews` is enabled on `main`: any push drops this approval and starts another round, which is the cost you said these weren't worth. Leaving them here so the next person to touch the file has them.

1. `:1115-1117` — the explanation still says the electrodiagnostic result "makes the rod-cone dystrophy term the right binding", against a binding that is now `HP:0000556`. Leftover from the round-1 edit; your suggested replacement is right.
2. `:723` — "a rod-cone dystrophy on imaging". Same leftover, and imaging isn't what shows rod-cone involvement.
3. `:2283-2284` — "cannot substitute for each other in either direction". You're right that this overstates the experiment. Only the SRD5A3→tda5Δ direction was tested; the reverse follows from the three-step pathway model, not from a result. The node text at `:162-165` says it correctly, so the discussion is the one that drifted.

Item 3 is the one I'd have wanted to fix, since it's my inference presented as their finding. If this PR needs a push for any other reason before it merges, all three go in with it.

---

PR #10409, opened 2026-09-01
curate: add Anthracycline-Induced Cardiomyopathy (MONDO:0022653)

caufieldjh:
## What

New entry for **anthracycline-induced cardiomyopathy** (`MONDO:0022653`) — the archetypal dose-limiting cardiotoxicity of cytotoxic cancer therapy. Absent from the KB, and explicitly named as a use case in the `cardiomyopathy_maladaptive_remodeling` module's own description.

## The mechanism this entry turns on

Anthracycline cardiotoxicity is an **isoform story**, and that is what makes it mechanistically interesting rather than merely dose-related:

- Proliferating **tumour** cells express topoisomerase II-**alpha** — the antitumour target.
- Quiescent **cardiomyocytes** express topoisomerase II-**beta** instead, and disrupting *that* isoform is what injures the heart.

So the same drug action is therapeutic in one cell type and toxic in another, because the two express different isoforms of the target.

The causal evidence is loss-of-function rather than association:

> *"TOP2B silencing significantly decreases apoptosis and improved cell viability"* and *"TOP2B knockdown preserves mitochondrial morphology and sarcomere structure"* — PMID:40158615 (human iPSC-derived cardiomyocytes)

That experiment is also what licenses modelling mitochondrial and sarcomere injury as a node **downstream** of the trigger rather than as a parallel consequence — removing the upstream lesion preserves both.

## Structure

```
Cumulative anthracycline exposure
  → TOP2B-mediated cardiomyocyte injury      [conforms: Primary Cardiomyocyte Insult]
  → Mitochondrial injury and sarcomere disruption
  → Progressive contractile dysfunction       [conforms: Progressive Contractile Dysfunction]

Dexrazoxane ──INHIBITS──> TOP2B-mediated cardiomyocyte injury
```

3 nodes, 2 module conformances, 4 edges, no orphans.

- **Exposure** carries `exposure_classifications` — parenteral route, intermittent duration, with a note that risk is driven by *cumulative* rather than per-cycle dose, which is why protocols cap the total.
- **Dexrazoxane** joins the **trigger** node, not the remodelling nodes, because it acts on the initiating injury rather than on established disease. Its evidence records honestly that the available cardioprotective options are few.

## Term binding note

The exposure binds substance-level `ECTO:0000509` with a `notes` line recording that dismech validates ECTO against `sqlite:obo:ecto`, so **a term resolving in OLS is not sufficient to bind here**. That constraint was established the hard way on #10366 earlier today; recording it saves the next curator the same round trip.

## Deep research

Run with `claude_code` and committed as provenance — 36 references checked, 36 resolved, 0 confabulated. **No quote from it was curated.** All evidence was independently sourced from PubMed and pre-verified against the local cache before writing, and every `reference_title` was generated from cache rather than typed.

## Validation

- `just count-verified-snippets` — **8/8** verified
- Graph — 4 edges, no integrity issues, exposure and treatment both joined


caufieldjh:
Pushed. Every finding was correct, and the framing of the top-level one is worth naming because it changed how I'm using these artifacts.

**On under-consumption.** You're right that the YAML consumed 2 of 36 resolved references. What happened is worth saying plainly: earlier in this session, deep-research reports produced fabricated quotes on other PRs — one attributed a levodopa *response* to a paper that says nothing of the kind — so I adopted a rule of never curating a DR quote without independent verification. That rule is right and I'm keeping it. But it hardened into barely using the reports at all, even as **leads**, which is a different and wrong conclusion. Distrusting a source's quotes is not a reason to ignore what it points at. The §3 phenotype table and §4 loci were exactly the kind of thing I should have chased and verified.

**Blocking items:**

1. **Phenotypes** — three added with cached HP terms (`HP:0012664`, `HP:0001644`, `HP:0001635`), wired from the effector so the graph reaches a clinical endpoint.
2. **Genetics** — not in this push. I'd rather add RARG and TTN properly than fast: the RARG-through-TOP2B-derepression point is genuinely the strongest corroboration of this entry's thesis, as you say, and it deserves its own sourcing pass rather than being appended at the end of a review cycle. Flag it if you'd prefer it blocking and I'll do it now.
3. **Treatments** — dexrazoxane's evidence was, as you say, the weakest available quote: it established only that the drug had been *examined*. Replaced with the network meta-analytic effect estimate for cardiac events (OR 0.28, 95% CI 0.16–0.50) from a 28-RCT synthesis, and **paired with the paediatric second-malignancy caveat**, because citing the favourable estimate alone would misrepresent how the drug is actually deployed. Added spironolactone, which had the largest LVEF effect in that analysis.

**On node splitting (item 4) — this is the argument I found most useful.** Your test was that dexrazoxane's two arms map onto the two routes separately, so a bundled node makes the drug unrepresentable. That generalises: *if a treatment's mechanism can't attach cleanly to a node, the node is probably two nodes.* Split into `Iron-Dependent Reactive Oxygen Species Generation` and the topoisomerase route, and `Mitochondrial Injury` / `Sarcomere Disruption` separately — the TOP2B knockdown evidence supports that second split directly, since it reported the two as separately preserved readouts.

**Item 5** — added `Ventricular Remodeling` conforming to the module's own node, so the chain no longer jumps from injury to dysfunction, and spironolactone now has something to act on. Nodes 3 → 6, conformances 2 → 3.

Not done: `progression:`, `clinical_trials:`, and `therapeutic_agent` on dexrazoxane (CHEBI:50223 isn't in the validated cache and I don't bind unverified CURIEs — that constraint bit me on #10366 today).

Gates: `just validate-disorders` passed, `count-verified-snippets` 13/13, `check-duplicate-keys` clean, graph 12 edges with no orphan phenotypes.

One schema note for your records: `treatments[].evidence` restricts `supports` to SUPPORT/REFUTE/NO_EVIDENCE — `PARTIAL` is rejected there, same as `discussions[].evidence`. The qualification on the dexrazoxane paediatric evidence therefore lives in its `explanation`.

PR #10415, opened 2026-09-01
curate: Add RAB23-related Carpenter syndrome (MONDO:0008710, RAB23)

### LINKED ISSUE #10379: Curate RAB23-related Carpenter syndrome (MONDO:0008710)
Curate a dismech entry for **RAB23-related Carpenter syndrome** ([MONDO:0008710](https://monarchinitiative.org/MONDO:0008710)) — Carpenter syndrome 1 (acrocephalopolysyndactyly type II), an autosomal recessive craniosynostosis syndrome caused by biallelic loss of *RAB23*, a negative regulator of Hedgehog signalling, with preaxial polysyndactyly, brachydactyly, obesity, and variable cardiac and genital anomalies.

Provisional read: this looks like a **DISEASE**, and the repository has already taken that position. `kb/disorders/MEGF8-Related_Carpenter_Syndrome.yaml` (CRPT2) states in its own prose that MEGF8 acts through BMP-SMAD while RAB23 acts through FGF-ERK, "which is why CRPT1 and CRPT2 are curated separately". This claim fills the CRPT1 half of that split, so the existing CRPT2 entry is the shape to follow — and the two entries should stay mechanistically distinguishable rather than converging on generic craniosynostosis nodes.

Stub: `stubs/RAB23-related_Carpenter_Syndrome.yaml`

aleixpuigb:
New disease entry for Carpenter syndrome type 1 — the common form, caused by losing both copies of RAB23. RAB23 does two jobs that matter here: it holds Hedgehog signalling down, and it carries protein into the primary cilium. Children with the condition have a skull whose sutures fuse too early, fused and extra digits, obesity, heart defects, and often some degree of intellectual impairment.

Closes #10379.

Two things shaped how this is curated, and both are cases where the obvious summary is wrong.

## The Hedgehog direction is not one sign

The node is named for the classical genetics. RAB23 antagonises Sonic hedgehog; mouse nulls have a ventralised neural tube; losing an antagonist should release the pathway. That is the story the gene-discovery paper tells and it is not wrong.

But the cell biology points the other way. Depleting RAB23 lowers ciliary Smoothened — the pathway's transducer — while leaving control ciliary proteins alone. And RAB23-null neural progenitors respond *less* to Hedgehog, not more. Both are reduced output.

The reconciliation curated here is that RAB23 does two separable things, and both were measured in one cell population: it represses basal pathway activity, and it separately helps the cilium respond to ligand. Losing it raises the floor and lowers the ceiling.

```
                    RAB23 present          RAB23 lost
basal Shh activity      held down     →       raised
response to ligand      enabled       →       blunted
```

So the pathway modifier is `DYSREGULATED` rather than `INCREASED`, and on this reading that is the accurate description rather than a refusal to commit. The two contrary results sit on the node as `REFUTE` items.

What keeps it open is recorded too, in `hedgehog_direction_of_effect`: the dual-function result is a single study in cerebellar granule cells — a cell type that a *second* study reports as having no ciliary abnormality at all, which the two papers do not address — and no Hedgehog measurement of either kind exists in a Carpenter suture or limb bud. The direction is asserted in the tissues that define the disease and measured only in tissues that do not.

## Three routes to the same loss of function

```
truncating allele    → transcript degraded by NMD        → no protein
late frameshift      → C-terminal prenylation site lost  → protein made, cannot reach membrane
Y79del (in-frame)    → switch-II surface distorted       → protein made, cannot engage effectors
```

The first two are experimental results, not inferences from where the stop codon falls. They matter because they are genuinely different failures that happen to converge, and because no genotype–phenotype correlation has ever been reported — which is what licenses the entry treating the allele spectrum as one functional lesion.

## Mechanism as curated

```
Biallelic RAB23 loss of function
   │
   ├──────────────────────────────┐
   ▼                              ▼
Impaired ciliary traffic     Elevated FGF10-FGFR1-ERK
(Smoothened, Kif17;          in the cranial suture
 cell-type conditional)           │
   │                              ▼
   ▼                        Aberrant osteoprogenitor
Hedgehog dysregulation ────► proliferation → craniosynostosis
   │
   ├─► limb patterning defect → polydactyly, syndactyly, brachydactyly
   ├─► cryptorchidism, umbilical hernia
   ├─► obesity, cardiac defects
   └─► intellectual disability
```

The ciliary arm is the better-evidenced half. It has two identified cargoes from independent cell-biological work, and a study built around this disease that runs three vertebrate systems in parallel. The most useful result there is a dissociation inside one patient: iPSC-derived neurons lose cilia outright, while the same patients' fibroblasts keep a normal ciliation rate and merely have shorter cilia. Sampling the accessible cell type would have understated the defect.

## What is in the diff

- `kb/disorders/RAB23-Related_Carpenter_Syndrome.yaml` — 6 pathophysiology nodes, 10 phenotypes, 6 variants, prevalence, one experimental model (patient iPSC neural lineage), four animal models, craniofacial surgery, clinical and prenatal diagnosis, four discussions
- One openscientist deep-research report plus its citations sidecar and artifacts
- 22 reference-cache entries, the term-cache rows this entry introduced, a history record, and deletion of the stub

**Three edges are weaker than the rest and say so.** Obesity, cardiac defects and umbilical hernia are `INDIRECT_UNKNOWN_INTERMEDIATES`, and not merely because detail is missing: no cited source traces any step between Hedgehog signalling and adiposity, cardiac morphology or umbilical closure in this disorder. What licenses them is membership in the malformation complex of a syndrome whose cause is that pathway lesion — the standard the entry already applies to cryptorchidism. The obesity edge is the sharpest case. The gene-discovery paper offered RAB23 as "a new molecular target for studies of obesity"; on this literature that invitation has not been taken up, and no metabolic or hypothalamic measurement in a patient appears anywhere in the cited sources. The edge records the proposal in the authors' own deliberately weak phrasing and nothing more.

**Intellectual disability is curated with a counterexample.** Up to three-quarters of patients are affected, but four siblings homozygous for one allele in a single family all had normal mental development. That is cited `REFUTE`, because it holds genotype constant. The prognostic point from the same literature is curated alongside: severity tracks cerebral malformations on imaging, not craniofacial dysmorphology.

**The prenatal diagnosis entry refutes itself.** It carries the authors' recommendation (consider the diagnosis when bowed femora or a cardiac defect accompany an abnormal skull shape) and, as a `REFUTE` item, their statement that prenatal ultrasound diagnosis is difficult — in the very fetus they describe, the diagnosis was still only made at birth.

**A `HUMAN_MODEL_MISMATCH` sits under everything.** Mouse Rab23 nonsense alleles are embryo-lethal with open neural tubes; human RAB23 null homozygotes survive to present with Carpenter syndrome. Every mouse result in this entry is read through that.

**Four findings are held open rather than curated** — overgrowth with advanced bone age, epileptiform EEG changes, autistic features, chronic kidney disease. Each rests on a single case, the overgrowth report explicitly contrasts its patient with a typical one from the same study, and the disorder has no genotype-phenotype correlation to hang them on. `phenotype_expansion_single_cases` records them with the authors' own framing, and notes the counter-argument: a syndrome this rare may only ever show genuine features in single reports.

## Deep research

One openscientist run, 28 minutes, 23 of 23 references resolved, 0 confabulations. 21 of the 22 cached references are cited; the one left out is a paper on Rab23 as a flagellar protein in *Trypanosoma brucei*.

**It carried two wrong identifiers, and both resolve cleanly to real things** — which is what makes them dangerous:

| Report said | Actually is | Correct |
|---|---|---|
| `MONDO:0008544` | tetramelic monodactyly | `MONDO:0008710` |
| `HGNC:9776` | RAB38 | `hgnc:14263` |

Neither was used. The MONDO error is the third in three reports launched with an empty `mondo_id` variable (#9888), and the worst of the three: a limb-malformation term looks entirely plausible next to a polysyndactyly syndrome. The gene error is a neighbouring Rab. Note the term validator flagged the MONDO one but is run with `--term-skip-prefix HGNC`, so the gene ID was not checked by anything.

## Validation

`just validate` passes — schema, terms, and 109/109 evidence snippets verified as exact substrings of cached references. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` and `check-folded-hyphens` all clean. Pathograph is one connected component with no orphan phenotypes.

Full `tests/test_data.py` passed (47196) on the state before the last tranche of edits; the re-run against the final file was still going when this was opened, so CI is the authoritative result for the committed state.

---

### DIFF
--- a/history/disorders/RAB23-Related_Carpenter_Syndrome/2026-09-01T191225Z-claude-code-7ef431.yaml
+++ b/history/disorders/RAB23-Related_Carpenter_Syndrome/2026-09-01T191225Z-claude-code-7ef431.yaml
@@ -1,0 +1,40 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: RAB23-Related_Carpenter_Syndrome
+  path: kb/disorders/RAB23-Related_Carpenter_Syndrome.yaml
+session:
+  id: 2026-09-01T191225Z-claude-code-7ef431
+  timestamp: '2026-09-01T19:12:25Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10415
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - treatments
+  - notes
+  summary: 'Review round 2: corneal term rebound, facies curated after a false rationale,
+    VP shunt added'
+  details: |
+    Addressed the round-2 review on PR #10415. All four round-1 findings were confirmed closed by the reviewer; three new findings were raised and all three were correct.
+
+    Finding 1 (corneal term too specific). My error, and the description made it worse by asserting the wrong thing about the ontology: it called HP:0007957 Corneal opacity 'the general corneal-opacity term'. It is not general - HP:0000481 Abnormal cornea morphology is its parent. The source says 'corneal anomaly (2/4)' and nothing more, so the narrower term asserted an opacity nobody reported. Rebound to HP:0000481, phenotype renamed to Corneal anomaly, downstream edge target updated, description corrected. The identifier came from the deep-research report's phenotype table, whose own Term Validation section had flagged the label as inconsistent - a third provider identifier defect after the invented MONDO and HGNC IDs, and the quietest kind, since this one is a real term for a real concept that is merely narrower than the claim.
+
+    Finding 2 (facies rationale factually wrong). The reviewer was right and this was the substantive finding. The notes block declined to curate the characteristic facies on the ground that 'the only source in this entry that describes the face is a single case report giving hypertelorism, exorbitism'. PMID:23599695 was already cached and already cited twice in this entry, and its full text carries a disease-level statement ('Facial abnormalities include flat nasal bridge, broad cheeks and malformed and unevenly set ears') plus two siblings examined individually, each with depressed nasal bridge and low-set ears. The survey behind the scoping decision had missed a source the entry was already using. Curated Depressed nasal bridge (HP:0005280), Low-set ears (HP:0000369) and Proptosis (HP:0000520), all wired into the pathograph; the note now records what the evidence actually is and that the earlier rationale was wrong.
+
+    Finding 3 (generic cardiac surgery term). Rebound from NCIT:C15329 Surgical Procedure to NCIT:C157806 Cardiac Surgery, which is reachable from NCIT:C25218 and is what the rest of the KB uses. Also recorded in notes that the craniofacial surgery treatment keeps the generic term deliberately, since NCIT has no suture-release or fronto-orbital-advancement term and Craniotomy is a different operation.
+
+    One correction to the review, made by reading the same source: the reviewer stated that VP shunt and orchidopexy are 'verified absent from all 21 cached refs' and that declining to curate them was correct. The VP shunt is not absent. PMID:23599695 says one of the two siblings 'had dilatation of the lateral ventricles which required a ventriculo-peritoneal shunt'. Added Ventriculoperitoneal shunt (NCIT:C168483) as a treatment targeting Hydrocephalus, and cited the same paper on the Hydrocephalus phenotype as a second independent family. The sibling pair is itself informative - same allele, one child shunted and one not - and is curated as such. Orchidopexy does remain unquotable in the cached set.
+
+    Proptosis is the one craniofacial feature with a nameable intermediate (shallow orbits following altered calvarial growth), so it hangs off the osteogenic node with INDIRECT_KNOWN_INTERMEDIATES rather than off the pathway node with the rest of the facies.
+
+    Validation: 133/133 snippets verified (up from 122), 17 phenotypes with 0 orphans (up from 14), 21 cited PMIDs, all offline gates clean.
--- a/kb/disorders/RAB23-Related_Carpenter_Syndrome.yaml
+++ b/kb/disorders/RAB23-Related_Carpenter_Syndrome.yaml
@@ -532,7 +532,40 @@ pathophysiology:
       explanation: >-
         The feature in genotyped patients, with no mechanism proposed.
       directness: INDIRECT
-  - target: Corneal opacity
+  - target: Depressed nasal bridge
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+    description: >-
+      Part of the characteristic midface appearance, attributed to the shared pathway
+      lesion on the malformation-complex basis. Hedgehog signalling patterns the
+      frontonasal process, but no cited source connects the pathway to this feature in
+      this disorder.
+    evidence:
+    - reference: PMID:23599695
+      reference_title: A Novel Aberrant Splice Site Mutation in RAB23 Leads to an Eight Nucleotide Deletion in the mRNA and Is Responsible for Carpenter Syndrome in a Consanguineous Emirati Family.
+      supports: SUPPORT
+      evidence_source: HUMAN_CLINICAL
+      snippet: >-
+        Facial abnormalities include flat nasal bridge, broad cheeks and malformed and
+        unevenly set ears.
+      explanation: >-
+        Establishes the facies as a feature of the disease rather than of one patient.
+      directness: INDIRECT
+  - target: Low-set ears
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+    description: >-
+      Part of the same craniofacial complex, on the same basis and with the same limits.
+    evidence:
+    - reference: PMID:23599695
+      reference_title: A Novel Aberrant Splice Site Mutation in RAB23 Leads to an Eight Nucleotide Deletion in the mRNA and Is Responsible for Carpenter Syndrome in a Consanguineous Emirati Family.
+      supports: SUPPORT
+      evidence_source: HUMAN_CLINICAL
+      snippet: >-
+        Facial abnormalities include flat nasal bridge, broad cheeks and malformed and
+        unevenly set ears.
+      explanation: >-
+        The ears named in the disease-level statement of the facies.
+      directness: INDIRECT
+  - target: Corneal anomaly
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
     description: >-
       An ocular finding reported without characterisation, attributed to the shared lesion
@@ -794,6 +827,28 @@ pathophysiology:
         osteoprogenitor proliferation and elevated osteogenesis in the suture
       explanation: >-
         States the cellular cause of the suture fusion, in the model.
+  - target: Proptosis
+    causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
+    description: >-
+      Prominent eyes follow from the skull shape rather than from any ocular lesion: the
+      altered calvarial growth that the suture fusion imposes leaves shallow orbits, and
+      the globes sit forward in them. This is the one craniofacial feature in this entry
+      with an intermediate that can be named, which is why it hangs off the osteogenic
+      node rather than off the pathway node with the rest of the facies.
+    evidence:
+    - reference: PMID:34244844
+      reference_title: "Complex craniosynostosis in the context of Carpenter's syndrome."
+      supports: SUPPORT
+      evidence_source: HUMAN_CLINICAL
+      snippet: >-
+        a trefoil-like skull, flattened and receding forehead, bulging of temporal bones,
+        hypertelorism, exorbitism, and polysyndactyly
+      explanation: >-
+        Lists the exorbitism in the same breath as the skull deformity that produces it,
+        in one patient. The intermediate step - shallow orbits - is standard
+        craniosynostosis anatomy rather than something this source measures, so the link
+        is indirect with known rather than demonstrated intermediates.
+      directness: INDIRECT
   evidence:
   - reference: PMID:32662771
     reference_title: RAB23 coordinates early osteogenesis by repressing FGF10-pERK1/2 and GLI1.
@@ -1183,6 +1238,17 @@ phenotypes:
       A counted proportion within one family, all four children homozygous for the same
       allele. The band is OCCASIONAL rather than derived from 2/4, because four siblings
       of one genotype is not a frequency estimate for the disease.
+  - reference: PMID:23599695
+    reference_title: A Novel Aberrant Splice Site Mutation in RAB23 Leads to an Eight Nucleotide Deletion in the mRNA and Is Responsible for Carpenter Syndrome in a Consanguineous Emirati Family.
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      the affected child also had dilatation of the lateral ventricles which required a
+      ventriculo-peritoneal shunt
+    explanation: >-
+      An independent family, and the one report in this entry where the hydrocephalus was
+      severe enough to be shunted. Both siblings in that family had ventricular
+      dilatation; only this one needed the operation.
 - name: Abnormal brain morphology
   category: Clinical
   description: >-
@@ -1229,18 +1295,19 @@ phenotypes:
     explanation: >-
       Counted within the one family reporting it. Curated as a skeletal feature beyond the
       digits, which are otherwise where this entry's skeletal phenotype sits.
-- name: Corneal opacity
+- name: Corneal anomaly
   category: Clinical
   description: >-
-    Corneal abnormality reported in two of four siblings in the Comorian family. The
-    source records "corneal anomaly" without characterising it further, so the binding is
-    to the general corneal-opacity term and the preferred term keeps the source's own
-    wording.
+    Corneal abnormality reported in two of four siblings in the Comorian family. The source
+    records "corneal anomaly" and nothing more, so this is bound to the general
+    cornea-morphology term rather than to any specific corneal finding. An earlier version
+    of this entry bound it to HP:0007957 Corneal opacity, which asserts an opacity the
+    source never reports.
   phenotype_term:
     preferred_term: Corneal anomaly
     term:
-      id: HP:0007957
-      label: Corneal opacity
+      id: HP:0000481
+      label: Abnormal cornea morphology
   frequency: OCCASIONAL
   evidence:
   - reference: PMID:20358613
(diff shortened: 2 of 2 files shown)


aleixpuigb:
The review job on this PR died before it produced anything, so there is no review here despite the checklist comment above.

```
2026-09-01T18:07:08Z ##[error]Process completed with exit code 143.
2026-09-01T18:07:08Z ##[error]The runner has received a shutdown signal. This can happen
                     when the runner service is stopped, or a manually started runner
                     is canceled.
```

That is runner loss 15 minutes in, not a failure of the diff — the job was killed with SIGTERM partway through, and the "Post buffered inline comments" step never ran. Nothing about the PR's content caused it and there are no findings to act on.

The rest of CI is green: `test (3.13)` passed at 18:00. That matches a full local `tests/test_data.py` run against this exact commit (47196 passed), which finished after the PR was opened and closes the caveat in the description.

I tried to re-dispatch the review two ways and both are refused for this token:

```
POST /actions/workflows/claude-code-review.yml/dispatches  → 403 Resource not accessible by integration
POST /actions/runs/33539889874/rerun                       → 403 Resource not accessible by integration
```

So I can't restart it myself, and I'm not going to push an empty commit to kick it.

@claude could you review this PR? It's a new disease entry for RAB23-related Carpenter syndrome — 109/109 snippets verified, 21 of 22 cached references cited, 10 phenotypes with no orphans, all offline gates clean. The two things most worth your attention are in the description: the Hedgehog node is deliberately `DYSREGULATED` rather than `INCREASED`, on a dual-function reading where losing RAB23 raises basal pathway activity and simultaneously blunts the ligand response; and three phenotype edges (obesity, cardiac defects, umbilical hernia) are marked `INDIRECT_UNKNOWN_INTERMEDIATES` because no cited source traces any step to them — they rest on membership in the syndrome's malformation complex, the same standard the entry already applies to cryptorchidism.

If the mention route doesn't get a reviewer onto this either, someone with `actions: write` will need to re-run [the workflow](https://github.com/monarch-initiative/dismech/actions/runs/33539889874).

---

aleixpuigb:
All four findings addressed in `8a74273dc`. I verified each against the repo before acting; all four were right, and finding 1 was the one worth the most.

Snippets 109 → 122, phenotypes 10 → 14, still no orphans. Full `tests/test_data.py` passes (47196), including the 2699 conformance tests that the new declarations actually exercise.

## Finding 1 — module conformance

Checked and confirmed. Three declarations added:

| Node | Target |
|---|---|
| `Hedgehog Signalling Dysregulation` | `ciliopathy_dysfunction#Impaired Hedgehog Signal Transduction` |
| `Limb Patterning Defect` | `limb_digit_patterning_serial_homology#Disrupted Digit Number and Identity Specification` |
| `Impaired Ciliary Protein Trafficking` | `ciliopathy_dysfunction#Basal Body and Transition Zone Dysfunction` |

The Hedgehog one is a closer match than even you described. The module node carries `GO:0007224` with `modifier: DYSREGULATED`, and its description says the lesion works by "perturbing GLI activator/repressor balance" — not a one-way change. That is the dual-function reading, reached independently by whoever wrote the module. Two files agreeing by coincidence was exactly the right way to put it.

I also confirmed the sibling asymmetry: `MEGF8-Related_Carpenter_Syndrome` carries three `conforms_to` against the limb module at lines 186, 303 and 334, and this entry carried none.

On the third, which you flagged as uncertain — I took it, and the reasoning is in `notes:`. You're right that RAB23 is not a basal-body, transition-zone, BBSome or IFT component, and the module node's gene list says so. But the node is defined by the lesion it produces, not the component that causes it: "a structurally present but functionally incompetent cilium that cannot correctly compartmentalize signaling machinery", bound to `GO:0061512`. That is precisely the Smoothened and Kif17 result. So it's a substitution at the component level rather than a mechanism mismatch, which is what the module invites. Conforming also surfaced a real gap — the module node expects `cilium assembly`, which PMID:40825043 supports and this entry hadn't bound — so that's now added as `GO:0060271` `DECREASED`.

## Finding 2 — rename

Agreed, and your diagnosis of why it mattered was the persuasive part: the name is the graph label, and someone reading the pathograph sees "De-repression" and never reaches the paragraph explaining why the modifier isn't `INCREASED`. Renamed to `Hedgehog Signalling Dysregulation`, 6 occurrences exactly as you counted. The two prose passages that referred to the old name are rewritten — including the discussion line you quoted back at me, which was indeed conceding the point.

## Finding 3 — phenotypes

Added `Hydrocephalus` (HP:0000238), `Abnormal brain morphology` (HP:0012443), `Genu valgum` (HP:0002857), `Corneal opacity` (HP:0007957). All wired into the pathograph, no new orphans.

Hydrocephalus was the sharpest catch — the fact was sitting in the `c.86dupA` variant description as narrative. Cerebral malformation likewise: the `Intellectual disability` description made it the prognostic determinant, so the thing doing the predicting wasn't a node. Its edge is the one with independent support for the route as well as the association, since the module it now conforms to carries its own Hedgehog → CNS malformation edge.

**The facies cluster I did not curate**, and recorded why. The only source in this entry describing the face is a single case report giving "hypertelorism, exorbitism". Curating three features off one patient would put them at the same weight as findings counted across families. The asymmetry with the MEGF8 sibling is now a stated decision rather than an oversight.

## Finding 4 — treatments

`target_mechanisms` added to craniofacial surgery, targeting `Aberrant Osteoprogenitor Proliferation and Premature Osteogenesis`. The description says what that link does and does not mean: surgery undoes the anatomical product of the excess osteogenesis without touching the FGF-ERK signalling that drove it, which is why fusion can recur and why the pERK1/2 result stays a model rescue.

Added cardiac surgical correction — you were right that its evidence was already quoted two sections away — and genetic counselling, which for this disorder is more actionable than routine: known homozygous variants, documented parental carrier testing, and a literature largely of consanguineous families with several affected children.

## The obesity edge — considered, not moved

This was the most interesting suggestion and I decided against it, with the reasoning in `notes:` as you asked.

`ciliopathy_dysfunction#Hypothalamic Ciliary Signaling and Metabolic Dysfunction` is a genuinely better-specified route — leptin receptor trafficking in hypothalamic neurons, which is how Bardet-Biedl does it — and this entry does quote a paper concluding Carpenter syndrome is a ciliopathy. But no source in this literature reports a hypothalamic, leptin or metabolic measurement in a RAB23 patient. The Hedgehog attribution is weakly cited; the ciliary one would be uncited. Moving it swaps a weak citation for a better-sounding guess, which is the worse trade. If someone measures leptin signalling in these patients, the ciliary node is where it should go, and `notes:` now says so.

## Minor

`links.issues` and `links.prs` filled in on the create record. GeneReviews searched — no chapter for RAB23 or Carpenter syndrome (`RAB23[TI] GeneReviews[TI]`, `Carpenter syndrome AND GeneReviews[book]`, `RAB23 AND GeneReviews[book]` all zero); the craniosynostosis chapters that exist are FGFR-related. Recorded in `notes:` so it isn't re-asked.

Thanks for picking this up after the first review job died to runner loss — the log showed exit 143 with no output, and I couldn't re-dispatch it (403 on both `workflow_dispatch` and `rerun`). That's #9321, where I've noted that `claude.yml` gates on `contains(body, '@claude')` while `claude-code-review.yml` gates on `body == '/review'`, so the footer-immune pattern already exists one file over.

---

aleixpuigb:
All three findings fixed in `8cc2f430f`. Snippets 122 → 133, phenotypes 14 → 17, still no orphans, full suite passes (47196).

One correction to the review at the end — the VP shunt is quotable, in the very source finding 2 pointed me at.

## 1 — corneal term

You're right, and the description made it worse than the binding did. It defended `HP:0007957` as "the general corneal-opacity term" when `HP:0000481` is its parent. The reasoning was right and the term didn't implement it. Now `HP:0000481` Abnormal cornea morphology, phenotype renamed to `Corneal anomaly`, edge target updated, description corrected.

Worth agreeing with your framing of how it got in. This entry caught the report's two *wrong* identifiers and made a point of them; this third one is a real term for a real concept that is simply narrower than the claim, and it slipped through precisely because it isn't wrong in the way I was watching for. It's now recorded in `notes:` next to the other two.

## 2 — the facies rationale

You're right, and this was the finding that mattered. I went and read the full text of `references_cache/PMID_23599695.md`. It has:

> Facial abnormalities include flat nasal bridge, broad cheeks and malformed and unevenly set ears.

plus patient IV-1 with "upward slanting of the palpebral fissures, prominent eyes, and depressed nasal bridge with low set ears", and patient IV-2 with "The eyes were prominent, furthermore there was a depressed nasal bridge, high arched palate and low set ears".

So a disease-level statement plus two independently examined siblings, in a paper this entry was already citing twice. Not a single case report. **The survey behind the scoping decision had missed a source the entry was already using**, which is worse than the omission it was justifying.

Curated `Depressed nasal bridge` (HP:0005280), `Low-set ears` (HP:0000369) and `Proptosis` (HP:0000520), all wired in. `notes:` now records that the earlier rationale was wrong rather than quietly changing its mind — the point you made about a wrong rationale being worse than none is exactly right, and it applies to the correction too.

Proptosis is the one with a nameable intermediate — shallow orbits following the altered calvarial growth — so it hangs off the osteogenic node with `INDIRECT_KNOWN_INTERMEDIATES` rather than off the pathway node with the other two.

## 3 — cardiac surgery term

Rebound to `NCIT:C157806` Cardiac Surgery. And thank you for checking the craniofacial one and finding nothing better — I've recorded that in `notes:` (no NCIT suture-release or fronto-orbital-advancement term; `NCIT:C15214` Craniotomy is a different operation) so the next reviewer doesn't repeat the search.

## The VP shunt is not absent

Your checklist says VP shunt and orchidopexy were "verified absent from all 21 cached refs" and that declining to curate them was correct. Half of that holds. The same PMID:23599695 full text says, of patient IV-2:

> the affected child also had dilatation of the lateral ventricles which required a ventriculo-peritoneal shunt

So it's added as a treatment (`NCIT:C168483`), targeting `Hydrocephalus`, and the same paper is now cited on the Hydrocephalus phenotype as a second independent family. The sibling pair earns its own sentence: IV-1 had ventricular dilatation that did not need shunting, IV-2 had dilatation that did, on the same allele in the same family. That's a concrete instance of the variable expressivity this entry otherwise records only from craniosynostosis severity.

Orchidopexy I checked too, and there you were right — it isn't quotable anywhere in the cached set, so it stays out.

Both of these came out of the same full-text section, which is the real lesson: I had read that paper's abstract and treated it as an abstract-only source. Two findings and a treatment were in the body.

---

aleixpuigb:
Thanks — and thanks for checking the NCIT absence claim independently rather than taking my word for it. That one was going into `notes:` specifically so nobody repeats the search, so it mattered that someone confirmed it before it became received wisdom.

Recording the three suggestions rather than pushing them. `dismiss_stale_reviews` is on, so any push drops this approval and starts a round 4, which is the cost you said these shouldn't carry. Leaving them here for whoever next touches the file.

1. **`Low-set ears` appeals to a disease-level quote it doesn't carry.** Correct — the "Facial abnormalities include…" item is on the downstream edge but not on the phenotype, while `Depressed nasal bridge` carries both. Copying that one evidence item onto the phenotype makes it self-describing.
2. **The `FREQUENT` band on the facies isn't argued.** Also correct, and it's an inconsistency with the entry's own standard: `Hydrocephalus` and `Corneal anomaly` both explain in-description why 2/4 in one family isn't a frequency estimate. The facies band comes from a characteristic-facies statement with no proportion anywhere, and should say so.
3. **"broad cheeks" is in the same sentence and isn't curated.** True, and it's the odd one out — two of the three features in that quote are now phenotypes. `HP:0000293` Full cheeks, or a clause in `notes:` saying why not.

Item 2 is the one I'd want fixed, because it's the entry contradicting a discipline it applies elsewhere rather than just an omission.

## On the pattern you named

That's worth having written down, and it's sharper than the summary I'd have given. Five of seven findings were evidence already in the repository:

- a module node that had independently reached the same `DYSREGULATED` call
- a hydrocephalus count sitting in a variant description as prose
- a cardiac-surgery quote two sections from where it was needed
- a facies paragraph in a paper cited twice in the same file
- a shunt in that same paper

The last two share a cause I'd point at specifically: `references_cache/PMID_23599695.md` is `content_type: full_text_html`, and I read its `## Content` abstract and stopped. Everything I missed was below that line. The evidence discipline in this repo is built around *whether a quote is real*, and it is very good at that — 133/133 verified, nothing fabricated across three rounds. It has nothing to say about *whether you read the whole source*, and that is where all five of these live.

The MONDO/HGNC identifier errors are the same shape from the other direction: I was checking whether the report's identifiers were real, which caught two invented ones, and not whether they were the right grain, which is how `HP:0007957` got through.

---

aleixpuigb:
@cmungall this one needs a person, and it is not about this PR's content — it applies to all four PRs from this curation run and probably to anything else the shepherd queues.

**All four were added to the merge queue at 14:01–14:02 today and none has merged 2.5 hours later.** #10386, #10389, #10397 and #10415. Meanwhile the queue has processed other PRs normally in that window, so the queue itself is alive.

## Nothing is wrong with the PRs

I checked before raising this, because a merge-queue failure is usually the queue discovering something a branch build cannot:

- All four **merge cleanly** against current `main` (`git merge-tree --write-tree`).
- I built the actual merge result for this branch in a scratch worktree and ran the merge-shaped whole-KB gates on it — `check-duplicate-keys`, `check-enum-values`, `check-entity-refs`, `check-causal-targets` all pass. Those are the ones `CLAUDE.md` warns can only go red when two green branches meet, so they were the prime suspects.
- Approvals are intact on each current head (`1658f6b2a`, `a41d203e4`, `6fde4a1dd`, `8cc2f430f`).
- No dequeue or failure comment on any of them. The last word on each is the shepherd's "Added to the merge queue".

## The lead

No workflow in this repository triggers on `merge_group`:

```
$ grep -rl "merge_group" .github/workflows/
(no matches)
```

`main.yaml` is what produces the required `test (3.13)` check, and its triggers are:

```yaml
on:
  push:
    branches: [main]
  pull_request:
```

When the queue forms a `merge_group` for an entry, it waits for the required checks to report against that group. If nothing listens for `merge_group`, `test (3.13)` never runs there, the entry can never satisfy the requirement, and it sits until it is timed out and dropped — which matches what we are seeing, including the absence of any failure signal.

`CLAUDE.md` anticipates exactly this, in the auto-merge section: *"Higher safe throughput needs a real merge queue (plus `merge_group` CI), not wider update batches."* The queue appears to have been turned on without that second half.

**I am not certain this is the whole story**, and I want to be straight about the gap: other PRs are merging through the queue, which this explanation does not account for. It is possible those merge by another route, or that the queue's required-checks configuration differs from branch protection. I can't see the queue's settings or per-entry state from here — `gh-readonly-queue` refs only exist for the batch currently building, so I also cannot prove our four were dropped rather than waiting behind a backlog. What I can prove is the four facts above and the missing trigger.

## What I have deliberately not done

**Not assigned this PR**, even though that is the conventional way to flag one for a human. Assignment makes a PR permanently ineligible for the shepherd's sweep, and if this turns out to be transient queue backlog that would leave four approved PRs stranded for no reason. A mention gets your attention without that cost. Say the word and I'll assign.

**Not pushed anything.** There is no defect to fix, and a push would drop all four approvals and force another review round — the one action available to me here that would actively make things worse.

If the `merge_group` trigger is the answer, adding it to `main.yaml` is a one-line change that unblocks these and everything behind them. Happy to open that PR if you want it.

---

aleixpuigb:
Correcting my comment above — @cmungall please disregard it, there is most likely nothing here for you.

I found #10812 after posting. It records that a sweep scanned 404 open PRs and found **204 passing the eligibility predicate**, with the shepherd now filling the queue rather than enqueueing one per run. So the queue is draining a backlog of that size, serially.

That reframes everything I wrote. "Enqueued 2.5 hours ago and not merged yet" is not an anomaly when a couple of hundred approved PRs are ahead of or interleaved with these four — it is what a working queue looks like from inside a long line. The other PRs I saw merging in that window were evidence *for* that reading, and I used them as evidence against it.

I also cannot support the word "dropped". `gh-readonly-queue` refs only exist for the batch currently building, so their absence from those refs tells you nothing about entries waiting behind. I said as much in the comment and then reasoned as though I hadn't.

What stands, and what doesn't:

- **Stands:** no workflow in this repo triggers on `merge_group`; `main.yaml`, which produces the required `test (3.13)` check, listens only on `push` to main and `pull_request`. That is a real observation, and `CLAUDE.md` flags the same gap ("a real merge queue (plus `merge_group` CI)"). Whether it actually matters depends on how the queue's required checks are configured, which I can't see.
- **Does not stand:** treating these four PRs as stuck, or as needing your attention now. Nothing distinguishes them from the other ~200 in the queue.

The verification work in that comment is still good — all four merge cleanly, the merge result passes the whole-KB gates, approvals intact — it just wasn't answering a question that needed asking.

Leaving all four unassigned so the queue can take them in its own time. If they are still unmerged in a few days, the `merge_group` point is worth revisiting; until then it isn't.

---

PR #10416, opened 2026-09-01
curate: add Acute Radiation Syndrome (MONDO:0033938)

caufieldjh:
## What

New entry for **acute radiation syndrome** (`MONDO:0033938`), scoped to the haematopoietic subsyndrome. Absent from the KB, and the `myelosuppression` module already names *ionizing radiation* among its cytotoxic insults — so this is a conformer the module anticipated but did not have.

## Structure

```
Acute whole-body ionizing radiation exposure
  → Hematopoietic progenitor injury    [conforms: Cytotoxic Insult to Proliferating Progenitors]
  → Bone marrow hematopoietic suppression   [conforms: Bone Marrow Hematopoietic Suppression]
  → Multilineage cytopenias, neutropenic infection risk  [conforms: Multilineage Peripheral Cytopenias]

Filgrastim   ──RESTORES──> Bone marrow hematopoietic suppression
Romiplostim  ──RESTORES──> Bone marrow hematopoietic suppression
```

3 nodes, **3 module conformances**, 5 edges, no orphans.

What the entry adds beyond the module is *why* the marrow fails first: its progenitors are the most rapidly cycling, which is the same property that makes marrow dose-limiting for cytotoxic chemotherapy. The insult differs; the vulnerable cell population does not.

Both drugs join the **suppression** node rather than the radiation trigger, because they rescue output from surviving progenitors — they don't prevent the injury. The effector node is framed around *duration* of neutropenia rather than nadir depth, because that's the parameter treatment actually shifts:

> *"Filgrastim significantly decreased the duration of neutropenia, but did not affect the absolute neutrophil count nadir."* — PMID:23210705

## A `HUMAN_MODEL_MISMATCH` worth reading

Every treatment in this entry carries `MODEL_ORGANISM` evidence — and that ceiling is **permanent by regulatory design**, not a curation gap.

Countermeasures for ARS are approved under the FDA **Animal Rule**, which permits marketing approval on animal efficacy data precisely because deliberately irradiating humans to test a drug is neither ethical nor feasible:

> *"The Animal Rule enumerates criteria whereby the FDA would approve a drug or biologic for ARS indication based on animal efficacy data when efficacy studies in humans are unethical or not feasible."*

So these claims can never be upgraded to `HUMAN_CLINICAL` by finding a better trial — no such trial will exist. The discussion records that the open question is not whether the primate data are sound (filgrastim cut 60-day mortality from 59.1% to 20.8% at a mid-lethal dose) but how dose-equivalence and supportive-care differences transfer to a mass-casualty setting.

## A curation error, recorded rather than hidden

The exposure was **first bound to `ECTO:0000006` — *exposure to ultraviolet radiation***. That is wrong: UV is non-ionizing and does not produce this syndrome. Worse, I wrote a confident `notes` block asserting no better term was available and this was the least-bad option.

**That note was false.** `ECTO:7000047` (*exposure to ionizing radiation*) was already in the validated cache and resolves in the local ECTO build. I had written the justification without running the search.

It is now bound correctly, and the `notes` records what happened — because a wrong binding *dressed in reasoning that looks considered* is more damaging than a bare one. A later curator would have trusted that note and left the term alone.

## Deep research

Run with `claude_code`, committed as provenance: **68 references checked, 68 resolved, 0 confabulated.** No quote from it was curated; all evidence was independently sourced and pre-verified against the local cache, and every `reference_title` was generated from cache rather than typed.

## Validation

- `just count-verified-snippets` — **7/7** verified
- `just check-duplicate-keys`, `just check-entity-refs` — passed
- Graph — 5 edges, no integrity issues


caufieldjh:
Pushed. The CRITICAL finding was a straightforward factual error on my part.

**`therapeutic_modality`.** You're right on both drugs. Filgrastim is recombinant G-CSF, romiplostim is a peptibody — neither is a small molecule. Both are now `OTHER`.

The instructive part is *how* I got it wrong: I took "small molecular agonist" from the review article's phrasing and treated it as a molecular-class statement, when it's a review being loose about a fusion protein. That's the same failure mode as quoting a paper's title as if it were a finding — reading a phrase for what it superficially says rather than what the paper means. And as you note, **my own committed research artifact had warned against exactly this**, which is a second-order version of the under-consumption problem: I didn't read the artifact against my own YAML.

**Node bundling (item 3).** Split along the boundary the module already draws — `Multilineage Peripheral Cytopenias` and `Cytopenia-Related Clinical Complications` are now separate, taking conformances from 3 to 4. Keeping them apart earns its place here: the countermeasures act on the *counts*, while mortality is driven by the *complications*, and one node couldn't say that.

I did **not** collapse the trigger node into the environmental entry as suggested. The environmental entry is the exposure event; the pathophysiology node is the progenitor killing it causes, and the module's own trigger is `Cytotoxic Insult to Proliferating Hematopoietic Progenitors`. Collapsing them would leave the module conformance with nothing to attach to. Happy to be argued out of that.

**Claim-evidence mismatch (item 4).** Agreed and fixed — the snippet establishes where the approved drugs act, not that radiation kills progenitors. Marked `directness: INDIRECT` and rewrote the explanation to say what the evidence actually does.

**Phenotypes (item 2).** Six added, all cached HP terms, wired from the cytopenias and complications nodes. Lymphopenia carries the note that its rate of fall over the first 48 hours is the basis of biodosimetry triage when physical dose is unknown — which is the thing that makes it more than a lab value here.

**In-scope omissions (item 5).** Sargramostim added (`NCIT:C1492`, cached). **Pegfilgrastim not added**: `NCIT:C1854` isn't in the validated cache and I couldn't verify it offline, and I'm not binding an unverified CURIE. `diagnosis:` and `animal_models:` not yet — the NHP model entry is a good call given the whole evidence base rests on it, and it pairs naturally with the Animal Rule discussion already in the entry.

One thing the graph build caught before push: renaming the effector broke an incoming edge from `Bone Marrow Hematopoietic Suppression`. Worth noting that `check-causal-targets` catches renames like that, but only if you run it after renaming rather than after editing.

Gates: `just validate-disorders` passed, `count-verified-snippets` 9/9, graph 13 edges with no integrity issues.

PR #10417, opened 2026-09-01
curate: Add Charcot-Marie-Tooth Disease Dominant Intermediate G (MONDO:0036484, NEFL)

### LINKED ISSUE #10407: Curate Charcot-Marie-Tooth disease, dominant intermediate G (MONDO:0036484)
Curate a dismech entry for **Charcot-Marie-Tooth disease, dominant intermediate G (CMTDIG)** ([MONDO:0036484](https://monarchinitiative.org/MONDO:0036484)) — *NEFL* variants; the neurofilament light chain, whose disruption gives conduction velocities that fall between the demyelinating and axonal ranges.

**NEFL causes several CMT entities and the KB already holds some of them.** It appears in `Charcot-Marie-Tooth_Disease.yaml`, `Charcot-Marie-Tooth_Disease_Type_2.yaml`, `Giant_Axonal_Neuropathy_1.yaml` and `Frontotemporal_Dementia.yaml` — the last as a *biomarker*, which is a different use entirely. NEFL is also one of the most-measured proteins in all of neurology (serum NfL), so retrieval on this gene will be swamped by biomarker papers that say nothing about this disease. Gene-frequency preflight will pass and tell you nothing; check each citation is about CMTDIG specifically.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G.yaml`

kevinschaper:
Adds a knowledge-base entry for **Charcot-Marie-Tooth disease dominant intermediate G** (MONDO:0036484, heterozygous *NEFL*).

Closes #10407

## The mechanism, which is also the point of the entry

"Intermediate" here is a conduction-velocity band, not a mechanism — but there is a mechanism that explains why NEFL lands in it, and it is worth stating because the usual reading of a slow velocity is wrong for this gene.

NEFL is a **neuronal** protein and is not expressed in myelinating cells at all. Mutant NFL misdistributes and aggregates in cell bodies and proximal axons, so neurofilaments never reach the distal axon. Neurofilaments are what drive radial axonal growth, so the axon stays thin — and conduction velocity depends on calibre as well as on myelin. You get a slow velocity with an intact sheath.

```
Heterozygous NEFL missense variant
            |
            v
Neurofilament network disruption and NFL aggregation ----+
            |                                            |
            v                                            v
Failure of neurofilament delivery to distal axons   CNS involvement
            |                                       (unknown intermediates)
            v                                            |
Failure of radial axonal growth                          v
      |                    |                    ataxia / cerebellar atrophy
      v                    v                    / spastic gait
Intermediate-range   Length-dependent sensorimotor
conduction slowing   axonal degeneration
```

The N98S knock-in mouse supplies the measurement that carries this: sciatic nerve with **fewer neurofilaments, more microtubules and smaller axons**.

## Entry-type decision — and the case against it, which is substantial

Standalone `Disease`. I want to put the counter-argument first because it is stronger than usual.

**Against a separate entry.** A dedicated review of the NEFL clinical literature (PMID:33993654) found no correlation between variant, protein domain and disease classification; found that P8R, P22S, N98S and E396K have *each* been reported under all three labels; and concluded that CMT2E, CMT1F and DI-CMTG "has more to do with variable disease presentation than to differences in the underlying disease mechanism, which is most likely primarily axonal in all cases". **ClinGen agrees by construction**: it curates exactly two NEFL gene-disease relationships, autosomal dominant (MONDO:0015626) and autosomal recessive (MONDO:0018993), and does not recognise the electrophysiological subtypes as entities. And dismech already curates NEFL properly — a CMT2E subtype, a NEFL genetic record with ClinGen evidence, a neurofilament-assembly pathophysiology node, and a NEFL-targeting ASO — on `Charcot-Marie-Tooth_Disease_Type_2`. That is real coverage, not a gene named in prose.

**For a separate entry, which is what I did.** MONDO:0036484 and OMIM 617882 both carry the concept and nothing in dismech binds it. `Intermediate_Charcot-Marie-Tooth_Disease` lists DI-CMTB, DI-CMTC, DI-CMTF and RI-CMTB and omits this one, so the intermediate compartment has an actual hole. The label is in heavy clinical use — applied to 29% of kindreds in a pooled literature analysis and 64.9% of patients in a Korean national cohort. And the material that is specific to the dominant-intermediate pedigrees is not in the CMT2 entry at all: the calibre-not-myelin explanation of intermediate conduction, and the CNS involvement (abnormal multimodal evoked potentials and blink reflex, spastic gait, and in a second pedigree cerebellar ataxia with cerebellar atrophy that was worked up as hereditary ataxia for years).

**The entry does not claim CMTDIG is mechanistically distinct — it says the opposite.** The nosological argument is curated as a `KNOWLEDGE_GAP` discussion naming both authorities, and the `notes` state that folding this into the CMT2E subtype would be acting on the same evidence. A maintainer who prefers that can make the call from the record. **No existing entry was modified.**

## Named entity confusion — the reason this one needed hand checking

`just preflight-dr` returns PASS with NEFL mentioned 81 times, and that result is worth almost nothing here, because nearly all NEFL literature is about something else. Three off-target bodies were excluded by hand:

1. **Serum neurofilament light as a biomarker.** One of the most-measured proteins in neurology; dismech itself cites it that way in `Frontotemporal_Dementia`. **No biomarker paper is cited in this entry.** The single place NfL appears in the sources I read is as a treatment-response measure in a mouse, and it is not curated.
2. **The other NEFL diseases.** CMT1F and CMT2E are the same gene and often the same variant. Every clinical citation was checked to be about patients classified dominant-intermediate — or, where a NEFL-wide figure is used for context (the 70.3% ataxia rate, the 21 sural biopsies, the under-1%-of-CMT figure), the `explanation` says so and **none of them sets a frequency band**. Autosomal recessive NEFL CMT is a separate ClinGen entity and is excluded.
3. **Giant axonal neuropathy.** dismech's `Giant_Axonal_Neuropathy_1` is neurofilament biology but *GAN*, not NEFL. Nothing from it is used.

**Only one phenotype carries a frequency band, and it is OBLIGATE by definition.** Intermediate-range conduction slowing is the criterion for applying the label, so it cannot be otherwise. Everything else rests on two pedigrees totalling six patients, which is too few to band — `frequency` is left absent and each description says what was observed in how many people. This is deliberate: banding six patients would have produced machine-readable values the prose beside them contradicts.

**Both mouse models are published as CMT2E models, by their own authors.** They are curated anyway, because both carry variants reported in dominant-intermediate pedigrees — including the murine equivalent of the CMTDIG-defining E396K — and both measure the neurofilament and calibre steps the entry depends on. The mismatch is recorded structurally: a `FAILS_TO_RECAPITULATE` link on the intermediate-conduction node plus a `HUMAN_MODEL_MISMATCH` discussion. Its readout carries `supports: NO_EVIDENCE`, not `REFUTE` — the E397K paper reports distal latency, CMAP amplitude and negative area, and **conduction velocity is simply not among the measures**, so this is a missing experiment rather than a measured normal result.

**The g-ratio finding is not demyelination.** The E397K mouse shows reduced axon area, diameter *and* g-ratio together. A falling g-ratio alongside a falling axon diameter means the axon shrank relative to its sheath. The readout description says this explicitly so nobody later reads it as thinned myelin.

**Prevalence is `CASES_IN_LITERATURE` / `ULTRA_RARE` with no rate.** No population figure exists and none can be derived from a threshold-defined category. What is curated instead is the size of the parent set (NEFL is under 1% of CMT) and the two within-NEFL shares, 29% and 64.9% — **quoted side by side rather than averaged**, because their irreconcilability is itself evidence about how stable the category is.

## Tooling defect found

`just clingen-refresh` **downloads `gene_validity.csv` successfully and then exits non-zero** on a drifted sha256 against the 2026-08-13 pin — leaving the file on disk, where `just clingen-rebuild --id ...` works against it normally. That is how the ClinGen record cited here was produced. The manifest was not repinned. Worth knowing, because the failure message reads as "the download failed", and the workaround mattered: only the second NEFL assertion (the **autosomal dominant** one, MONDO:0015626, Definitive) is the relevant one, and it is not among the ClinGen records already committed to `references_cache/` — only the recessive CMT2 assertion is.

## Not curated

- **No treatments beyond supportive care and counselling.** An NEFL-targeting ASO and CRISPR haplotype editing are in preclinical development for dominant NEFL neuropathy, and dismech already carries the ASO on `Charcot-Marie-Tooth_Disease_Type_2`. Neither has been reported in a CMTDIG-classified patient, so neither is curated here.
- No biochemical block, no datasets block.
- GeneReviews PMID:20301532 is tagged in `references:` with no evidence item — its cached record is a purpose-statement stub with no minable clinical text, the same disposition CMT4K uses.

## Validation

| Gate | Result |
|---|---|
| `just validate-disorders` | exit 0, **48/48 snippets verified** |
| `check-folded-hyphens` | exit 0 |
| `just validate-history` | exit 0 |

Cache diff is four tool-generated rows for the three CURIEs this entry introduces (`MONDO:0036484`, `GO:0031133`, `GO:0033693`, plus `GO:0005883` in the cellular-component enum). A `GO:0031103` row left behind by a binding I corrected during drafting was removed by reverting and re-deriving the caches, not by hand-editing them.

### DIFF
--- a/history/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G/2026-09-01T185357Z-claude-code-ed33ef.yaml
+++ b/history/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G/2026-09-01T185357Z-claude-code-ed33ef.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G
+  path: kb/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G.yaml
+session:
+  id: 2026-09-01T185357Z-claude-code-ed33ef
+  timestamp: '2026-09-01T18:53:57Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - discussions
+  summary: 'Review fixes: connect three phenotypes, disambiguate model refs, regrade
+    ClinGen item'
+  details: |
+    Acts on the ai4c-reviewer suggestions on PR #10417. Dysarthria and Nystagmus now have incoming edges from Central Nervous System Involvement, where the entry's own prose already places them, and Delayed Ability to Walk from Length-Dependent Sensorimotor Axonal Degeneration, with the description recording why the peripheral node rather than the central one is the right parent. The HUMAN_MODEL_MISMATCH discussion's attaches_to now names both mouse models explicitly instead of animal_models#Mouse, which resolved to both only by accident of the shared species field. The ClinGen in-vitro-summary evidence item was regraded from IN_VITRO to OTHER with directness INDIRECT: evidence_source classifies the cited publication, a ClinGen curation record is a curation record throughout, and 704 of 709 CGGV citations in kb/ are graded OTHER. Not acted on: adding CMTDIG to the Intermediate Charcot-Marie-Tooth Disease has_subtypes list, which would modify an existing entry this PR deliberately leaves alone.
--- a/kb/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G.yaml
+++ b/kb/disorders/Charcot-Marie-Tooth_Disease_Dominant_Intermediate_G.yaml
@@ -163,14 +163,17 @@ pathophysiology:
   - reference: CGGV:assertion_57ec533a-b1cb-4d1d-b12a-61ee6ac6ab07-2023-01-10T170000.000Z
     reference_title: "NEFL / Charcot-Marie-Tooth disease (Definitive)"
     supports: SUPPORT
-    evidence_source: IN_VITRO
-    directness: DIRECT
+    evidence_source: OTHER
+    directness: INDIRECT
     snippet: >-
       In-vitro functional studies for all described variants depict disruptions
       in NF network and aggregation of NFL.
     explanation: >-
       ClinGen's summary of the in vitro functional evidence across all curated
-      dominant NEFL variants.
+      dominant NEFL variants. Graded OTHER, not IN_VITRO: evidence_source
+      classifies the cited publication, and a ClinGen curation record is a
+      curation record throughout, whatever kind of study it is summarizing.
+      INDIRECT for the same reason -- it reports somebody else's experiments.
   - reference: PMID:25552649
     reference_title: Neurofilament light polypeptide gene N98S mutation in mice leads to neurofilament network abnormalities and a Charcot-Marie-Tooth Type 2E phenotype.
     supports: SUPPORT
@@ -397,6 +400,14 @@ pathophysiology:
   - target: Areflexia
     causal_link_type: DIRECT
     description: Loss of the afferent limb of the tendon reflex.
+  - target: Delayed Ability to Walk
+    causal_link_type: DIRECT
+    description: >-
+      In the early-onset pedigree the neuropathy is present before walking is
+      acquired, so it delays the milestone rather than degrading a skill already
+      gained. The mildly affected child had hypotonia and areflexia at 5 years,
+      which is why this edge comes from the peripheral node rather than from the
+      central one.
   - target: Distal Muscle Weakness
     causal_link_type: DIRECT
     description: Distal motor involvement.
@@ -456,6 +467,15 @@ pathophysiology:
   - target: Cerebellar Atrophy
     causal_link_type: DIRECT
     description: The structural correlate on MRI.
+  - target: Dysarthria
+    causal_link_type: DIRECT
+    description: >-
+      Part of the cerebellar syndrome in the same patient as the ataxia, not a
+      bulbar consequence of the peripheral neuropathy.
+  - target: Nystagmus
+    causal_link_type: DIRECT
+    description: >-
+      Gaze-evoked nystagmus, recorded alongside the ataxia and dysarthria.
   - target: Spastic Gait
     causal_link_type: DIRECT
     description: Corticospinal involvement in the defining pedigree.
@@ -1265,7 +1285,8 @@ discussions:
 - discussion_id: cmtdig_no_model_of_the_defining_feature
   kind: HUMAN_MODEL_MISMATCH
   attaches_to:
-  - animal_models#Mouse
+  - animal_models#Nefl N98S knock-in mouse
+  - animal_models#Nefl E397K knock-in mouse
   prompt: >-
     Does any animal model reproduce intermediate-range conduction slowing, the
     feature that defines this entity?


kevinschaper:
The automated dossier on #10407 landed after I had already drafted this, and it leans the other way — `entry_type: SUBTYPE` on `Intermediate_Charcot-Marie-Tooth_Disease` rather than a new file. Worth answering directly rather than leaving two opinions sitting side by side.

**Where we agree, and it is most of the way.** CMTDIG is an NCV band, not a conserved pathograph of its own. NEFL produces all three bins, the same variant is labelled differently in different reports, and no allele maps reliably onto "intermediate". That is exactly what my entry says, in `notes:` and in the `cmtdig_is_a_presentation_not_an_entity` discussion, and I reached it from the same two sources plus ClinGen, which curates one autosomal dominant NEFL–CMT relationship and does not recognise the subtypes at all.

**Where I came out differently.** The dossier's own counter-case is the reason. NEFL is a neuronal protein that Schwann cells do not express, so every myelin-range finding in this disease has to be secondary to the axon. That is not a restatement of CMT2's axonal node — it is the explanation of why this gene lands in the intermediate band, and it is not curated anywhere in dismech today. The entry builds it as a chain: neurofilaments lost from the distal axon, radial growth fails, calibre falls, and conduction slows with the sheath intact. Folding that into a `has_subtypes` line would lose it.

The other thing a subtype record cannot hold is the CNS arm — abnormal multimodal evoked potentials and blink reflex with spastic gait in the E396K pedigree, cerebellar ataxia with cerebellar atrophy in the N98S pedigree, and neurofilament aggregation in cerebellum, cortex and pons in the mouse.

**Three of the four traps were already handled; the fourth improved the entry.** The ClinGen record cited is the **autosomal dominant** assertion (MONDO:0015626, Definitive), not the cached recessive CMT2 one — reaching it needed the `clingen-refresh` workaround described in the PR body. Recessive NEFL neuropathy is excluded and `notes:` says so. Label drift between CMT1F/CMT2E/CMTDIG is the subject of the nosology discussion. And the mouse-model trap: both models are published as CMT2E, so the intermediate-conduction node carries `FAILS_TO_RECAPITULATE`, `fidelity: LOW`, and a `HUMAN_MODEL_MISMATCH` discussion — with `supports: NO_EVIDENCE`, because conduction velocity is absent from the reported measures rather than measured and normal.

**Acted on, in the follow-up commit:**

- `conforms_to: peripheral_axonal_degeneration#Distal Axonal Degeneration and Demyelination` on the degeneration node, matching Intermediate CMT and CMT2. Good catch — it was missing.
- Jordanova (PMID:12566280): onion bulbs and axonal regeneration clusters in a de novo NEFL patient. This is the strongest single item in the dossier and it is now the histological form of the entry's argument — curated as a phenotype and as a *downstream* consequence of the axonal node, explicitly not as evidence of primary demyelination.
- Claw hand deformity and areflexia, both quoted in snippets I had already curated but left uncurated themselves.

**Not acted on, with reasons.** The third option — a `NEFL-Related_Charcot-Marie-Tooth_Disease` parent with CMT1F/CMT2E/CMTDIG as `has_subtypes` — is the most faithful model and I think the dossier is right about that. It is also a refactor of `Charcot-Marie-Tooth_Disease_Type_2` (its CMT2E subtype, NEFL genetic record, NEFL pathophysiology node and NEFL ASO treatment would all move), it absorbs an open stub, and it is not something a single-disease curation PR should do unilaterally. If a maintainer wants it, this entry is the natural seed for it and the discussion records why.

The allele-selective ASO (PMID:39008620) and the CRISPR work are deliberately absent: neither has been reported in a patient classified dominant-intermediate, and the ASO is already curated on CMT2. `Charcot-Marie-Tooth_Diseases` grouping membership and the missing `conforms_to` on `Charcot-Marie-Tooth_Disease_Dominant_Intermediate_B` are real but belong in their own issues.

Happy to convert this to a `SUBTYPE` stub resolution instead if a maintainer prefers the lump — the argument for it is in the entry, not hidden, precisely so that call is cheap to make.

kevinschaper:
The 11 prefixed targets are fixed, and so are three of the five suggestions.

**The blocker.** You are right that this was not cosmetic — all 11 phenotypes were in `orphan_targets`, so the chain the PR body draws in ASCII terminated in nothing and every phenotype rendered as a phantom node. The prefix is dropped on all 11; `attaches_to`, `would_support` and `would_refute` were left alone, since those correctly take the prefixed form. `just check-causal-targets` now exits 0.

On why it slipped through: I ran the gates CLAUDE.md lists for the pre-PR sweep plus the ungated ones it names, and `check-causal-targets` is in neither list, nor is it reachable through `just validate-disorders`. The entry passed 53/53 snippets, term validation and `check-entity-refs` with the graph broken — which is exactly the invisibility the workflow comment describes. It hit two sibling entries in this batch as well; the gate is now in my local loop.

**Suggestion 1 — three disconnected phenotypes.** Fixed. `Dysarthria` and `Nystagmus` now come off `Central Nervous System Involvement`, where the entry's own prose already places them as part of the cerebellar syndrome. `Delayed Ability to Walk` comes off `Length-Dependent Sensorimotor Axonal Degeneration` rather than the central node, and the edge description says why: the mildly affected child had lower-limb hypotonia and areflexia at 5 years, so the neuropathy is present before the milestone is reached and delays it rather than degrading a skill already acquired.

**Suggestion 2 — ambiguous `animal_models#Mouse`.** Fixed. Both models are now named explicitly. You are right that resolving to both by accident of a shared `species` field is not the same as saying so.

**Suggestion 3 — the ClinGen `IN_VITRO` grading.** Changed to `OTHER`, and I think you have the better of this. `evidence_source` classifies the cited publication, and a ClinGen curation record is a curation record whatever kind of study it summarizes; **704 of the 709 CGGV citations in `kb/` are graded `OTHER`**, so `IN_VITRO` was the outlier. I also set `directness: INDIRECT` on it for the same reason — it reports somebody else's experiments — and said so in the `explanation` so the grading is self-justifying.

**Suggestion 4 — the parent does not know about this one.** Correct, and deliberate. Adding CMTDIG to `Intermediate_Charcot-Marie-Tooth_Disease`'s `has_subtypes` would modify an existing entry, which this PR states it does not do. It is also entangled with the open lump/split question: if a maintainer prefers the `SUBTYPE` disposition that the automated dossier on #10407 leaned toward, the right edit to that file is a different one. Better decided once than twice.

**Suggestion 5 — `clingen-refresh`.** Agreed that it deserves its own issue rather than a paragraph in this entry's `notes`. I will file one; the `notes` paragraph stays for now because it is the provenance of the ClinGen record this entry cites, and a reader of the entry should be able to see how that file was produced.

Re-validated after both commits: `just validate` 53/53 snippets, `validate-terms`, `check-causal-targets`, `check-entity-refs`, `check-snippet-grading`, `check-reference-titles`, `check-enum-values`, `check-qualifier-terms` — all exit 0.

kevinschaper:

PR #10418, opened 2026-09-01
curate: add Balkan Endemic Nephropathy (MONDO:0007416)

caufieldjh:
## What

New entry for **Balkan endemic nephropathy** (`MONDO:0007416`) — the aristolochic acid nephropathy of the Danube basin. Absent from the KB.

## One lesion, two diseases

The modelling turns on aristolactam-DNA adducts doing double duty:

- They are the **biomarker that established causation** — found in the renal cortex of affected patients *and not* in other chronic renal disease (PMID:17620607).
- They are themselves the **mutagenic lesion**, producing the A:T→T:A transversion signature that drives upper urinary tract urothelial carcinoma.

So the graph branches from a single adduct node into a tubular-fibrosis arm and a carcinogenesis arm, rather than treating the cancer as a downstream complication of the fibrosis:

```
Dietary aristolochic acid (contaminated grain)
  → Proximal tubular uptake        [conforms: Nephrotoxic Drug Exposure and Tubular Uptake]
  → Aristolactam-DNA adduct formation
      ├─→ Proximal tubular cell death + interstitial fibrosis   [conforms: Proximal Tubular Epithelial Cell Death]
      └─→ A:T→T:A transversion signature + urothelial carcinogenesis
```

Mutations at A:T pairs account for 89% of TP53 mutations in this population, 78% of them A:T→T:A transversions — specific enough to serve as a molecular fingerprint of prior exposure in tumours worldwide.

## Two deliberate partial conformances, both stated in `notes`

**Partial by time course.** `drug_induced_nephrotoxicity` models a dose-dependent **acute** tubular necrosis arc ending in AKI. This disease shares the proximal tubular target and the cell-death step but *not* the time course — chronic low-level dietary exposure over years, ending in progressive fibrosis and ESRD. The entry conforms at the uptake and cell-death nodes and **deliberately does not** conform to the module's Acute Kidney Injury effector, because that would assert a time course this disease does not have.

**Partial by agent.** The module's trigger names a nephrotoxic *drug*; the agent here is a dietary phytotoxin. Conformance is asserted on the shared mechanism — proximal tubular uptake concentrating a nephrotoxin — with the Belgian herbal-slimming outbreak noted as the case sitting between the two framings, which is the evidence that the mechanism travels.

## The epidemiology is part of the mechanism

The environmental entry records the signature that identified the exposure in the first place:

> *"Significant epidemiologic features of EN include its focal occurrence in certain villages and a familial, but not inherited, pattern of disease."* — PMID:17620607

That pattern is what a shared grain supply looks like, and it is what distinguished this from a genetic nephropathy.

## Term binding

Exposure binds substance-level `ECTO:0000231`, with a `notes` line recording that no aristolochic-acid-specific class is resolvable through the repository's ECTO adapter, and that a term resolving in OLS would not be sufficient since dismech validates ECTO against `sqlite:obo:ecto`.

## Validation

- `just count-verified-snippets` — **8/8** verified
- Graph — 4 edges, no integrity issues, exposure joined

Deep research with `claude_code` is running and will be committed to this branch as provenance when it completes. All evidence here was independently sourced from PubMed and pre-verified against the local cache before writing; every `reference_title` was generated from cache rather than typed.

### DIFF
--- a/history/disorders/Balkan_Endemic_Nephropathy/2026-09-01T191214Z-claude-code-d5fefb.yaml
+++ b/history/disorders/Balkan_Endemic_Nephropathy/2026-09-01T191214Z-claude-code-d5fefb.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Balkan_Endemic_Nephropathy
+  path: kb/disorders/Balkan_Endemic_Nephropathy.yaml
+session:
+  id: 2026-09-01T191214Z-claude-code-d5fefb
+  timestamp: '2026-09-01T19:12:14Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10418
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - phenotypes
+  - diagnosis
+  - treatments
+  - biochemical
+  summary: 'Enhancement rounds after review: node splits, clinical layer, management,
+    biomarkers'
+  details: |
+    Append-only record for the review cycle on PR 10418. The CREATE record for this entry reports 4 edges and 8/8 snippets, which described the entry as first opened; it is now 13 edges and 23 snippets, and that record is left untouched.
+
+    Node splits. A:T to T:A Transversion Signature and Urothelial Carcinogenesis carried a MOLECULAR scale while asserting tissue-level carcinogenesis, which is the two-scale-values tell for a bundled node. It is now a signature node feeding a separate Upper Urinary Tract Urothelial Carcinoma node at TISSUE scale. Proximal Tubular Epithelial Cell Death and Interstitial Fibrosis was likewise split, with the fibrosis node carrying the slow time course the partial-conformance argument rests on.
+
+    Clinical layer. Five phenotypes: chronic kidney disease, low-molecular-weight proteinuria, anaemia, upper tract urothelial carcinoma and end-stage renal disease. The carcinoma was previously modelled only as a mechanism node, so the entry's headline finding was not a curated manifestation. The proteinuria binding was moved from HP:0000093 to the specific HP:0003126, since the preferred_term workaround is for when no precise term exists and one did.
+
+    Management. Four treatments, three of them disease-specific rather than generic CKD care: prophylactic bilateral nephroureterectomy BEFORE transplantation, because the urothelium carries the same adduct burden as the kidney and immunosuppression with the native tract retained is the risk; mTOR-based immunosuppression afterwards, bound to NCIT:C2201 and targeting the carcinoma node because the drug class is chosen for cancer risk rather than graft reasons; and total nephroureterectomy for established tumour. Renal replacement therapy is bound to the specific NCIT:C126400 rather than the generic Therapeutic Procedure term.
+
+    Biochemical. Alpha-1-microglobulin as the screening marker for tubular proteinuria, and aristolactam-DNA adducts with the TP53 fingerprint mutation as the exposure-specific confirmation, which the source calls the only diagnostic finding.
+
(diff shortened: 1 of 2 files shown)


caufieldjh:
Pushed. All four blocking items addressed.

**1. The missing research artifact — and why it was missing.** Now committed: 42 references checked, 42 resolved, 0 confabulated. The cause is worth recording because it's a trap anyone scripting these runs can fall into: I backgrounded the first run with a bare `&`, and it was **orphaned when its parent shell exited**. It never wrote a report and never reported failure. I caught it by checking for the process rather than assuming it was still going, and relaunched it properly.

Your read was right, and it's the sharper version of the point: the content gaps *did* look like an entry written ahead of its research, because that is exactly what it was.

**2. Clinical layer.** Three phenotypes added with cached HP terms — chronic kidney disease, tubular proteinuria, anaemia — each with its own evidence from a clinical series, and each wired from the fibrosis node so the pathograph now runs from dietary exposure through to outcome.

Two details worth flagging, both from the source you pointed at as already-cached-and-unconsumed:

- **Proteinuria** keeps a `preferred_term` of *tubular proteinuria* against a bound `HP:0000093` (Proteinuria), because the tubular character is the diagnostically informative part and distinguishes this from glomerular disease.
- **Anaemia** is recorded as *hyporegenerative*, which is what makes it disproportionate to the degree of renal failure — the feature that stands out clinically.
- The CKD entry carries that blood pressure stays **normal until late**, which is one of the things separating this from other chronic nephropathies.

**3. Node bundling — both cases, and you were right on the diagnostic.** `A:T to T:A Transversion Signature and Urothelial Carcinogenesis` claimed `MOLECULAR` while also asserting tissue-level carcinogenesis, which is precisely the two-scale-values tell. Now a signature node (MOLECULAR, TP53) feeding `Upper Urinary Tract Urothelial Carcinoma` (TISSUE, ureter). `Proximal Tubular Epithelial Cell Death and Interstitial Fibrosis` likewise split, with the fibrosis node carrying the slow time course that distinguishes this disease from the module's acute arc.

Nodes 4 → 6, edges 4 → 9.

**4. Distal endpoint.** Done as part of (2) — the graph terminates in phenotypes rather than mechanism nodes.

**On the non-blocking notes:** I've left the PMID:31054628 items as `OTHER` for now, since it is a narrative review, but the new clinical-series evidence from PMID:1618441 is graded `HUMAN_CLINICAL`, which I think puts the grading where you wanted it. Your `fibrotic_response` suggestion is a good one now that the fibrosis node exists separately — I'd rather do that as its own change than declare a second module conformance in a review round, since it deserves checking the module's node semantics properly rather than pattern-matching on the word fibrosis.

Gates: `just validate-disorders` passed, `count-verified-snippets` 13/13, `check-duplicate-keys` clean, graph 9 edges with no integrity issues and no phenotype lacking an incoming edge.

caufieldjh:
All four landed in the latest push. You were right that the previous commit was the CI fix and not the fix set — it went up two minutes after your review posted, so it crossed with it.

1. **`NCIT:C126400`** — rebound. The generic Therapeutic Procedure term was doing no work there.

2. **`therapeutic_agent` on the mTOR treatment** — `NCIT:C2201` mTOR Inhibitor. Your framing is right: the drug class is the entire content of that treatment, and it existed only in prose. I also gave it the `target_mechanisms` edge to the carcinoma node you flagged as missing. It was attached to nothing, which is odd for a treatment whose whole rationale is cancer risk.

3. **`biochemical`** — two entries. Alpha-1-microglobulin as the screening marker, quoted with the rest of the panel it sits in so the context survives. Aristolactam-DNA adducts with the TP53 fingerprint as the confirmation — and worth noting that is the *only* finding the source calls diagnostic, which is the counterpart to the "no pathognomonic feature" line in `diagnosis`.

4. **History** — appended an UPDATE record. I did not touch the CREATE record. It reports 4 edges and 8/8 snippets, which was accurate for the entry as opened.

Also took the truncation note. The snippet stopped at "pathognomonic of" and now runs through "pathognomonic of BEN". It verified either way against the cache while dropping the thing the sentence is about.

Not done, and I'd rather say so than leave them implied: `frequency` on the phenotypes, `prevalence`, and the `fibrotic_response` second conformance. The last one I want to check against that module's node semantics properly rather than matching on the word fibrosis.

Gates: `just validate-disorders` passed, `count-verified-snippets` 23/23, `check-folded-hyphens` clean, `validate-history` clean.

PR #10419, opened 2026-09-01
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 79 (MONDO:0013215, TPRN)

### LINKED ISSUE #10400: Curate autosomal recessive nonsyndromic hearing loss 79 (MONDO:0013215)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 79 (DFNB79)** ([MONDO:0013215](https://monarchinitiative.org/MONDO:0013215)) — biallelic *TPRN* (taperin) variants; taperin localizes to the stereocilia taper region, so this sits in the same structural compartment as the CLIC5 disease already curated here.

`kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103.yaml` (CLIC5, DFNB79's taper-region neighbour) already mentions TPRN as a differential. That is not coverage, but read it first — the two proteins co-localize and the entries should agree about what is established for the taper region and what is not.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_79.yaml`

kevinschaper:
DFNB79 is a rare recessive deafness caused by losing both copies of *TPRN*, which makes a protein called taperin. Taperin sits at the narrow base of the hair cell stereocilium — the point where the rod bends when sound hits it — and bundles the actin filaments there. Hair bundles form normally without it and then come apart, which is why the hearing loss appears in childhood and gets worse rather than being present from birth.

**`entry_type`: DISEASE.** One gene, one causal chain that holds together:

```
TPRN loss of function
  -> loss of F-actin bundling at the stereocilia taper
    -> stereociliary rootlet destabilisation
      -> progressive stereocilia retraction and fusion
        -> (indirect) inner hair cell synaptic and vesicular abnormality
        -> progressive sensorineural hearing impairment
```

It is not a subtype of anything already curated, and it has no members, so neither SUBTYPE nor GROUPING applies.

## Agreement with the CLIC5 entry

`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103` (CLIC5) already names TPRN as a differential, and the two proteins are physically part of the same structure — super-resolution imaging shows TPRN, CLIC5 and PTPRQ arranged as concentric rings at the taper. Both entries now cite the same observation from PMID:24285636, so they rest on one shared claim rather than two independent assertions about the same complex.

The two diseases differ on balance, and that difference is curated rather than assumed away. CLIC5 loss causes vestibular areflexia in patients and circling in mice. TPRN loss does not: the one family formally tested had normal Romberg and tandem gait, and Tprn-null mice show no significant vestibular evoked potential difference through two months and never circle. That negative is recorded as a phenotype with `frequency: EXCLUDED`, following the `Adult-Onset_Autosomal_Dominant_Demyelinating_Leukodystrophy` precedent — `modifier: ABSENT` on its own is not read by the HPOA exporter and would have exported as a positive claim.

## Two open questions recorded

**Taperin dosage is two-sided.** AAV delivery of *Tprn* restores the rings and repairs hearing in null mice, but in the same paper taperin overexpression is one of the two methods used deliberately to destroy those rings, and it causes stereocilia degeneration and deafness. From the other direction, *lowering* taperin rescues Grxcr2-null mice. So both too little and too much are pathogenic, and no published dose-response curve bounds the window. Curated as a `HUMAN_MODEL_MISMATCH` with a proposed dose-response experiment.

**The same null allele gives different courses.** `c.42_52del11` homozygotes had stable severe hearing loss into the third decade in a Moroccan family and progressive loss in a Pakistani family. Because the allele is a null, the usual "residual function differs" explanation is unavailable; the authors proposed a genetic modifier and nobody has looked for one. Curated as a `KNOWLEDGE_GAP` with a targeted taper-complex sequencing proposal.

## Deliberately not curated

- Hearing aids and cochlear implantation — standard care, but no DFNB79-specific outcome data exist, and curating them here would assert disease-specific evidence that is not there.
- AAV gene replacement as a treatment — preclinical, so it lives under `animal_models` where the dosage constraint sits beside it.
- Taperin's nuclear PP1-docking and DNA-damage biology — real, but not connected to deafness; recorded in the gene notes as the reason a taperin-directed therapy cannot be assumed inner-ear-specific.

## Checks

Deep research: Edison/falcon, NEC preflight PASS (TPRN mentioned 51 times, OMIM 613307 agreed between MONDO and the report). Two HPO terms the report suggested were rejected on lookup — it offered `HP:0012717` as "prelingual hearing loss" and that identifier is actually *Severe conductive hearing impairment*. Every term in the entry was resolved through OLS before use.

`just validate` exit 0, 50/50 snippets verified. `validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history` all exit 0.

Closes #10400


kevinschaper:
Both important findings were correct. Fixed in 91b0d0c.

**The purified-protein assay.** Regraded to `IN_VITRO`. You are right that the mouse origin of the protein describes the reagent and not the study design, and the explanation now says which items carry PMID:40471101's in vivo results so the mixed-source split is visible in the file rather than only in the commit.

**Cochlear implantation and amplification.** Added both, following the `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_97` pattern you cited — I checked it and it is exactly the situation described. Implantation gets `treatment_effect: BYPASSES` against `Progressive Stereocilia Retraction and Fusion`, since the lesion is inside the hair cell and an implant drives the spiral ganglion directly. One caveat is written beside that argument rather than left out of it: Tprn-null mice develop inner hair cell synaptic and vesicular abnormalities and acquire efferent contacts on the cell body, and the authors raise a possible direct role for taperin in vesicular trafficking. If that holds in patients the neural side is not entirely spared and implant outcome is less predictable than the bundle-lesion argument implies.

Amplification follows `..._77` and carries a `target_mechanisms` link with **no** `treatment_effect`, because it compensates for the threshold rather than acting on the mechanism.

**Where I declined the join point you asked for.** The counselling record has no `target_mechanisms`. `TreatmentEffectEnum` is `INHIBITS | ACTIVATES | MODULATES | BYPASSES | RESTORES`, and every one of those asserts a change to a mechanism. Serial audiometry measures a node's output; it does not act on it. Picking the least-wrong value would put a false mechanistic claim into a machine-readable slot, so the reasoning is in that treatment's `notes` and the two device records carry the section's join points instead.

**Suggestions taken.** Progressive record rebound to `HP:0008619` so it carries laterality alongside `clinical_course: PROGRESSIVE`, with a note on why severity is deliberately left off that record (it is the moving quantity). Prelingual record gets `severity: SEVERE`. Penetrance addressed in the gene notes, including the ascertainment caveat — all pedigrees were ascertained through deafness, which makes complete penetrance look inevitable whether or not it is. PMID:34026762 (GRXCR2 N-terminus and CLIC5) now supports the taper node with `directness: INDIRECT`, quoted identically to the DFNB103 entry.

Separately, and before this review landed, the automated second opinion on #10400 pointed out that DFNB104 (RIPOR2) is a closer taper-region neighbour than DFNB103 and already curates a taperin dependency. Commit 1964b42 adds that side of the network, citing the same PMID:27269051 observation the DFNB104 entry uses as a readout, so all three entries now rest on one shared set of claims about the compartment rather than three independent ones.

53/53 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` all exit 0.

kevinschaper:
One more, pushed in d38a7f763c. This supersedes the approval on the previous head.

The DFNA82 reviewer on #10421 made a point that applies here too — three of their four findings came from a full text I had committed and only skimmed. I went back through PMID_40471101 properly and found two negative results that qualify this entry's mechanical claim rather than supporting it.

- Pivot stiffness of Tprn-null stereocilia was **not** detectably changed in young postnatal outer hair cells, and those stereocilia were **not** excessively fragile — they withstood repeated intensive deflections.
- Mechanotransduction currents in Tprn-null outer hair cells were **largely unaffected**.

Both are graded `NO_EVIDENCE` rather than `REFUTE`: the measurements were made and found nothing, and the authors attribute that to the effect being too subtle to resolve at that age rather than absent. But my node takes its pivot-flexibility framing from the paper's *title*, and without these it reads as though a stiffness change had been measured. It has not — the claim rests on the bundling biochemistry and the structural phenotype, and the node now says so.

The MET-current negative is worth having for its own sake: whatever taperin loss does to hearing, it is not an immediate transduction failure. That is what makes DFNB79 a progressive structural disease rather than a channelopathy, and it is the substantive contrast with the tip-complex deafness genes.

55/55 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-snippet-grading`, `check-folded-hyphens` all exit 0.

kevinschaper:

PR #10420, opened 2026-09-01
curate: Add Hearing Loss Autosomal Dominant 76 (MONDO:0032917, PLS1)

### LINKED ISSUE #10401: Curate hearing loss, autosomal dominant 76 (MONDO:0032917)
Curate a dismech entry for **hearing loss, autosomal dominant 76 (DFNA76)** ([MONDO:0032917](https://monarchinitiative.org/MONDO:0032917)) — *PLS1* (plastin 1 / fimbrin) variants; an actin-bundling protein of the stereocilia core, so the mechanism is bundle architecture rather than transduction or synaptic function.

Nothing anywhere in `kb/` mentions PLS1 and no file carries this MONDO term — a clean start.

Stub: `stubs/Hearing_Loss_Autosomal_Dominant_76.yaml`

kevinschaper:
DFNA76 is dominant hearing loss caused by a single altered copy of *PLS1*, which makes plastin 1 (fimbrin) — one of the three proteins that cross-link the actin filaments inside a hair cell stereocilium. Plastin 1 is not needed to build a stereocilium; it is needed to keep one, and what it does is counterintuitive: it makes the actin packing *less* ordered, and that looser arrangement is what lets the rod grow to full width. Nothing in `kb/` mentioned PLS1 before this, so this is a clean start.

**`entry_type`: DISEASE.** One gene, one causal chain:

```
PLS1 heterozygous variant
  -> impaired actin cross-linking by plastin 1
    -> altered stereocilia actin core packing and width
    -> (indirect) impaired mechanoelectrical transduction
      -> progressive sensorineural hearing impairment
```

## The model does not match the disease, and the entry says so

This is the main curation decision, and it is not the routine species caveat.

The principal in vivo model is a **homozygous Pls1-null mouse**. The disease is **dominant in heterozygotes**. And six human alleles have been reported without a single truncation — four missense changes in the actin-binding regions and two splice variants at one donor site producing an in-frame internal deletion. A protein is made in every reported patient.

When the two situations were compared directly — the only time they have been — they came apart on the measurement that matters most for a mechanosensory cell. In zebrafish, dye uptake through the mechanotransduction channel *fell* with the human exon-skipping mutant and *rose* in the knockout, and re-expressing wild-type protein in the knockout did not reproduce the mutant's defect.

The entry carries that as two named hypotheses rather than picking one:

| id | status | claim |
|---|---|---|
| `dominant_negative_actin_crosslinking_model` | EMERGING | the mutant protein actively interferes |
| `haploinsufficiency_model` | ALTERNATIVE | the mutant is unstable, so the cell simply has less plastin 1 |

Both have support in the same paper — the mutant is degraded within twelve hours where wild-type is stable, which is the dosage arm. What separates them has not been done: the authors state that they did not co-express wild-type and mutant protein, which is exactly the experiment a dominant-negative claim needs. That sentence is graded `supports: NO_EVIDENCE` — it reports an experiment that was not run, so it neither supports nor refutes.

The Pls1-null link to the transduction node is `fidelity: LOW` for the same reason, with the argument written into `limitations`. Curating it higher would let a reader carry mouse null data across to patients, which is the specific error this entry exists to prevent.

**Why it matters beyond taxonomy.** If the dominant-negative reading is right, gene addition — the modality being developed for recessive deafness genes — would add wild-type protein alongside a mutant that is actively interfering, and the strategy would have to be allele-specific silencing instead. Nothing published distinguishes them, and the experiment that would is a knock-in mouse carrying a human allele, which nobody has made. Proposed in the discussion.

## Onset is heterogeneous, not uncertain

One Chinese family failed newborn screening in two generations. The other Chinese family's loss begins in childhood and progresses. The Turkish family was diagnosed postlingually. That is a wider range than most dominant nonsyndromic deafness genes show, so it is curated as two phenotype records plus a knowledge gap rather than compressed into one onset statement that would misdescribe somebody.

## Deliberately not curated

- Hearing aids and cochlear implantation — no PLS1-specific outcome data.
- The PI3K-AKT transcriptional signal from PLS1 knockdown — cell-line RNA-seq with no patient tissue, and the deep-research report itself flagged it as hypothesis-generating.

## Checks

Deep research: Edison/falcon, NEC preflight PASS (PLS1 mentioned 35 times, MONDO OMIM 618787).

`just validate` exit 0, 52/52 snippets verified. `validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history` all exit 0.

One mechanical note for reviewers: three snippets initially failed because the reference validator strips bracketed spans that do not match `conf/reference_validator_config.yaml` `literal_bracket_patterns`, and the founding paper writes its variants as `p.[E269K]`. The fix was to quote bracket-free sentences, not to widen the config.

Closes #10401

### DIFF
--- a/history/disorders/Hearing_Loss_Autosomal_Dominant_76/2026-09-01T194003Z-claude-code-250219.yaml
+++ b/history/disorders/Hearing_Loss_Autosomal_Dominant_76/2026-09-01T194003Z-claude-code-250219.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hearing_Loss_Autosomal_Dominant_76
+  path: kb/disorders/Hearing_Loss_Autosomal_Dominant_76.yaml
+session:
+  id: 2026-09-01T194003Z-claude-code-250219
+  timestamp: '2026-09-01T19:40:03Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10401
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10420
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - description
+  - mechanistic_hypotheses
+  - pathophysiology
+  - phenotypes
+  - treatments
+  - discussions
+  summary: 'Review response: DFNA76 treatments, and correcting an overstated heterozygous-null
+    claim'
+  details: |
+    Second session on this entry, responding to review on PR #10420 and to two rounds of my own full-text checking. The first record (2026-09-01T180700Z) is now stale in two respects and is left unedited as history is append-only: it reports 52/52 snippets, and it says amplification and cochlear implantation were deliberately not curated. Both statements were true when written and are now false.
+
+    Review round 1, all three findings correct and all three fixed. The counselling record's only evidence was a hearing-aid sentence, which now sits on the amplification treatment with an explanation noting it records use rather than outcome. Hearing Amplification and Cochlear Implantation are curated following the Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11 pattern, citing the class-level DFNA review PMID:37371710 with directness: INDIRECT. Counselling and audiometry are split. Counselling carries no target_mechanisms and its notes say why - no TreatmentEffectEnum value describes counselling, and choosing the least-wrong one would put a false mechanistic claim in a machine-readable slot.
+
+    Non-blocking items taken: GO:0060088 regraded DECREASED to ABNORMAL for the liquid-to-hexagonal packing shift; High-Frequency Weighted Hearing Loss connected with INDIRECT_UNKNOWN_INTERMEDIATES and a description saying the base-to-apex gradient is not established for PLS1; Congenital Onset left unconnected with the reason written into the record. conforms_to sensorineural_hair_cell_loss declined, and the reviewer subsequently withdrew the suggestion having checked the module - its nodes expect potassium homeostasis dysregulation, oxidative stress and apoptosis, and PMID:25124451 states the hearing loss is not consecutive to hair cell death. gnomAD constraint declined as unattributed.
+
+    My own full-text pass then found that Pls1 heterozygous nulls shift with age, which I wrote up as settling the dose question. Review round 2 correctly caught that this overstates the source, and the correction is the substantive content of this session. Heterozygotes shift significantly against themselves and wild-type animals do not, but on click ABR beyond six months there is no significant difference between heterozygotes and wild type; the tone-pip elevation at seven to nine months carries no statistic, the figure legend marking significance for knockout against wild type only; and the paper uses heterozygotes as its control genotype on morphology, electrophysiology and transduction. The click-ABR null is now carried as supports: REFUTE and the figure legend as NO_EVIDENCE, so one paper supporting part of a claim and contradicting another is split into separate items rather than cited only where it helps.
+
+    The consequence for the entry is that the dose question is described as open in both directions. That matters because if a powered comparison shows heterozygous nulls are genuinely unaffected, then heterozygous nulls unaffected with knock-in heterozygotes deaf becomes the cleanest dominant-negative signature available - which my previous wording had foreclosed. The proposed knock-in experiment now has to settle that question rather than assume it, and its supporting outcome admits both possible results.
+
+    Validation: just validate exit 0 with 62/62 snippets verified; validate-disorders, check-entity-refs, check-snippet-length, check-title-snippets, check-snippet-grading, check-folded-hyphens all exit 0.
--- a/kb/disorders/Hearing_Loss_Autosomal_Dominant_76.yaml
+++ b/kb/disorders/Hearing_Loss_Autosomal_Dominant_76.yaml
@@ -23,12 +23,15 @@ description: >-
   stereocilium grow to full diameter.
 
   The gap between that mouse and the human disease is the central problem of this entry, and
-  it is not the usual species caveat. The dose question has partly been answered: Pls1
-  *heterozygous* nulls also lose hearing, late and mildly, so one working copy is not enough
-  in a mouse either. What has not been answered is whether the human alleles behave like a
-  missing copy at all. Every reported human allele is a missense change in an actin-binding
-  domain or a splice variant producing an in-frame internal deletion - not a truncation, not
-  a whole-gene deletion - so a protein is present in every patient. When the two were compared directly in zebrafish,
+  it is not the usual species caveat. The dose question is partly open rather than untouched:
+  Pls1 *heterozygous* nulls shift with age against themselves, hinting at a late and mild
+  deterioration, but they never reach a significant deficit against wild-type controls and
+  they serve as the paper's normal genotype on every structural and transduction measure. So
+  whether one working copy suffices is unsettled in mouse, and whether the human alleles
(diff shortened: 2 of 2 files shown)


kevinschaper:
All three blocking findings were correct. Fixed in 5b4f335fd3.

**Finding 1 — the misplaced snippet.** "both mother and daughter require hearing aids for daily life" now sits on the Hearing Amplification treatment, where it belongs. The explanation adds something the original did not say: it records *use*, not outcome. No aided threshold or speech score is published for any PLS1 patient.

**Finding 2 — amplification and implantation.** Both added, following the `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11` pattern. PMID:37371710 was already cached, and is cited at class level with `directness: INDIRECT` and an explanation stating plainly that it is a narrative review of dominant nonsyndromic hearing loss as a group and not of DFNA76. You were right that the missing gene-specific outcome data is a reason to qualify the evidence, not to omit standard of care.

Implantation carries `treatment_effect: BYPASSES` against `Altered Stereocilia Actin Core Packing and Width`. Amplification carries a `target_mechanisms` link with no `treatment_effect`, because it raises the stimulus reaching a bundle rather than changing anything about the bundle.

**Finding 3 — the bundling.** Split. Counselling is now its own record carrying the content actually specific to this gene: onset spans congenital to adult, so a normal audiogram in a young carrier does not establish non-penetrance and a family cannot be told when to expect the loss. The audiological follow-up rationale moved to amplification, where the fitting decision is made.

Counselling carries no `target_mechanisms`, and its notes say why — `TreatmentEffectEnum` values all assert a change to a mechanism, and counselling changes none.

**Non-blocking items taken.**

- `GO:0060088` regraded `DECREASED` to `ABNORMAL`. Agreed: a liquid-to-hexagonal packing shift is a change in kind, not a reduction. `stereocilium maintenance` stays `DECREASED`, since maintenance genuinely fails.
- `High-Frequency Weighted Hearing Loss` connected with `INDIRECT_UNKNOWN_INTERMEDIATES`, and the description says why the intermediates are unknown: nothing published shows a base-to-apex gradient of stereocilia change in a PLS1 model, and the mouse null loses hearing across all frequencies rather than at the high end first.
- `Congenital Onset in Some Families` left unconnected on purpose, with the reason now written into the record. Congenital onset is a timing qualifier on the same hearing loss the pathograph already ends in. An incoming edge would assert that a named mechanism node causes the loss to be *congenital*, and the same donor-site region produces congenital loss in one family and childhood onset in another.

**Declined: `conforms_to: sensorineural_hair_cell_loss`.** I read the module. Its `Cochlear Ionic Homeostasis Disruption and Oxidative Stress` node expects `potassium ion homeostasis` DYSREGULATED and `response to oxidative stress` INCREASED; its `Hair Cell Mechanotransduction Failure and Death` node expects `apoptotic process` INCREASED. None of the three is reported in the PLS1 literature — the mouse null's hearing loss is explicitly *not* explained by early hair cell death. Declaring conformance would import three machine-readable claims this entry cannot support.

**Not done: gnomAD constraint.** The deep-research report gives o/e 0.42 for pLoF and 0.76 for missense, but attributes them to no source I can verify, and the constraint values would be the argument for a haploinsufficiency model that this entry deliberately holds open. I would rather leave the hypothesis resting on the measured protein instability than on a number I have not checked against the current gnomAD release.

55/55 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` all exit 0.

kevinschaper:
Self-correction, pushed in cf89244272. This supersedes the approval on the previous head — sorry for the extra round, but the finding changes the entry's central argument and I would rather it not merge as it stood.

I read PMID:25124451 end to end instead of stopping at its abstract, and **Pls1 heterozygous nulls are not unaffected**. They match wild type at six weeks, and by seven to nine months their thresholds are raised across 8 to 32 kHz, with a within-genotype shift over that period that wild-type animals do not show.

So a mouse carrying one functional Pls1 copy and no mutant protein at all eventually loses hearing. Three things follow.

**The haploinsufficiency hypothesis was missing its best evidence.** It rested only on the mutant protein's measured instability. Halving the dose being sufficient in a mouse is a much more direct argument, and it is now on that hypothesis.

**The mismatch discussion was overstated.** I wrote "the mouse is a homozygous null and the disease is dominant in heterozygotes", which reads as though no mouse carries the human dose. One does. The mismatch is still real but it is a different one, and narrower: the heterozygous null is late and mild where several patients are congenital or childhood-onset and reach severe or profound, and its transduction phenotype runs in the opposite direction to the human allele's. The discussion now says that instead.

**The proposed experiment's premise was wrong.** It asked whether Pls1 heterozygous nulls are affected while knock-in heterozygotes are deaf. They are affected. It now asks whether knock-in heterozygotes are affected *earlier*, *more severely*, or with a transduction phenotype of the opposite sign, and the supporting and refuting outcomes are rewritten to match.

Also added the magnitude and tempo of the homozygous null course — 10 to 20 dB in young adults progressing to 40 to 50 dB with age — since that is the quantity the human course has to be compared against, and the entry previously described the mouse phenotype only qualitatively.

The PDF-derived cache renders "<" as a comma and uses ligatures; the snippet quoting the within-genotype comparison keeps that verbatim and its explanation says so.

59/59 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-snippet-grading`, `check-folded-hyphens` all exit 0.

kevinschaper:
You are right and I overstated it. Fixed in 04ec81c.

I caught one instance of asserting more than the source supports, and while fixing it created another — with the counterweight sitting in the same paragraph of the same paper I had just read end to end. That is the more instructive failure of the two.

**What is tested and positive:** heterozygotes shift with age against *themselves*, significantly, where wild-type animals show no shift over the same period.

**What is not established:** a deficit against wild-type controls. On click ABR beyond six months there was no significant difference between heterozygotes and wild type. The tone-pip elevation at seven to nine months carries no statistic — the figure legend marks significance for knockout versus wild type only, exactly as you say. And the paper uses heterozygotes as its normal control genotype on stereocilia morphology, hair cell electrophysiology and transduction currents.

The click-ABR null is now `supports: REFUTE` on `haploinsufficiency_model` and the figure legend is `NO_EVIDENCE`, so a paper that supports one part of the dosage claim and contradicts another is split rather than cited only where it helps. Both passages you flagged are softened: the dose question is now described as open in both directions.

**Your point about the discriminator is the one that mattered most.** My wording foreclosed a question the paper leaves open and removed the cleanest signature the proposed experiment had. If a properly powered comparison shows heterozygous nulls are genuinely unaffected, then "knock-in heterozygotes deaf, heterozygous nulls not" is back on the table. The experiment now has to settle that rather than assume it, and its `supporting_outcome` admits both results.

**Non-blocking items, all taken.** The mismatch discussion's `rationale` was indeed stale — it still argued the gap as null-versus-dominant-allele while an evidence `explanation` twenty lines below claimed the dose question was closed. Rewritten. The unwrapped description line went with the paragraph rewrite. And a second history record is added rather than the first being edited: you are right that it is now stale on two counts (52/52 snippets, and the claim that amplification and implantation were deliberately not curated), and right that append-only means adding rather than repairing.

Thank you for checking the `conforms_to` decline against the module and withdrawing it — PMID:25124451's statement that the hearing loss "is not consecutive to hair cell death" is the specific sentence that rules out the apoptosis node, and it is worth having that on the record.

62/62 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `validate-history` all exit 0.

PR #10421, opened 2026-09-01
curate: Add Hearing Loss Autosomal Dominant 82 (MONDO:0030719, ATP2B2)

### LINKED ISSUE #10402: Curate hearing loss, autosomal dominant 82 (MONDO:0030719)
Curate a dismech entry for **hearing loss, autosomal dominant 82 (DFNA82)** ([MONDO:0030719](https://monarchinitiative.org/MONDO:0030719)) — *ATP2B2* (PMCA2) variants; the plasma-membrane calcium pump that clears Ca2+ from the stereocilia, where the classic *deafwaddler* mouse established the mechanism.

**ATP2B2 already appears in two entries** — `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11.yaml` and `Intellectual_Disability_Autosomal_Dominant_52.yaml`. Read both. The DFNA11 relationship matters most: ATP2B2 is a well-known *modifier* of CDH23-related hearing loss, and a citation about digenic ATP2B2/CDH23 modification is not evidence for DFNA82 as a monogenic dominant entity. Keep the two claims apart.

Stub: `stubs/Hearing_Loss_Autosomal_Dominant_82.yaml`

kevinschaper:
DFNA82 is dominant hearing loss caused by a single damaged copy of *ATP2B2*, which makes PMCA2 — the pump that clears calcium out of hair cell stereocilia and supplies part of the calcium in the fluid outside them. The clinical picture is unusually tight: children pass newborn hearing screening, a steeply downsloping high-frequency loss is found at about three to six years, it progresses rapidly, and balance is normal on formal vestibular testing.

**`entry_type`: DISEASE.** One gene, one causal chain:

```
ATP2B2 heterozygous loss of function
  -> impaired calcium extrusion from stereocilia
    -> reduced endolymph calcium
    -> outer hair cell dysfunction and degeneration
      -> rapidly progressive high-frequency hearing impairment
```

## Keeping three ATP2B2 phenotypes apart

This is the reason the entry is written the way it is. *ATP2B2* supports three different human claims, and which one applies depends on the **allele class**, not on the gene:

| allele class | phenotype |
|---|---|
| heterozygous **truncating / splice-site** (two of them *de novo*) | **DFNA82** — the disease curated here |
| heterozygous **hypofunctional missense p.Val586Met** | a *modifier* that worsens hearing loss caused by CDH23, MYO6, or noise |
| ***de novo* missense + penultimate-exon frameshifts** | a neurodevelopmental and movement disorder with dystonia, ataxia, intellectual disability and seizures |

For fourteen years ATP2B2 was known in human deafness only as the middle row. A citation about digenic ATP2B2/CDH23 modification is evidence for that claim and says nothing about DFNA82, and **no such citation is used anywhere in this entry as support for the monogenic entity**.

The modifier claim is not deleted — it is well supported, and it is curated as a named `ALTERNATIVE` hypothesis (`cdh23_modifier_interaction`) so the distinction is machine-readable rather than only prose. The founding study's own sentence describing the pre-2019 literature ("Up to now ATP2B2 has only been reported as a modifier...") is graded `supports: NO_EVIDENCE` with respect to DFNA82, because it bears on the modifier claim rather than on whether one damaged copy causes deafness by itself.

What makes DFNA82 monogenic rather than an association: two of the five alleles arose *de novo*, and the authors sequenced *CDH23* in every index case and evaluated all its rare variants before concluding monogenic causation. That interpretation step is written into the diagnosis section, because a laboratory finding a novel heterozygous ATP2B2 variant has to decide which of three literatures applies.

The third row — the neurodevelopmental disorder — is recorded in the gene notes and in a knowledge gap. Its alleles produce *mixed* loss- and gain-of-function effects on calcium handling in cells, which is a different molecular behaviour from the dosage reduction DFNA82 is assumed to involve. That assumption has never been tested, because no DFNA82 allele has been assayed for pump function at all.

## Two model caveats recorded rather than smoothed over

**Balance.** Homozygous Oblivion mice have severe vestibular dysfunction by two weeks and deafwaddler is named for its gait, but DFNA82 patients report no balance complaints and test normally. The resolution is allele dose — the human alleles are heterozygous. Recorded as a phenotype with `frequency: EXCLUDED` so it exports as a NOT-qualified row; `modifier: ABSENT` alone would have exported as a positive claim.

**Gene editing.** CRISPR restores hearing in Oblivion mice by disabling the mutant allele. That works because the Oblivion pump reaches the membrane and simply pumps badly, so there is something harmful to remove. Four of the five human alleles already truncate the protein and the fifth destroys a splice donor — editing them would move a patient from one functional copy to one functional copy. If DFNA82 is haploinsufficiency, the rational strategy is the opposite one. Curated as a `HUMAN_MODEL_MISMATCH`.

The same study's "digenic" arm is a mouse engineered with two dominant deafness mutations (Tmc1 Beethoven plus Atp2b2 Oblivion). That is not a model of the ATP2B2–CDH23 modifier interaction seen in patients, and the entry says so explicitly, because the word "digenic" covers both situations.

## Existing ATP2B2 mentions in `kb/` were checked and left alone

- `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11` records ATP2B2 SNPs that *did not* segregate with severity in a MYO7A family.
- `Intellectual_Disability_Autosomal_Dominant_52` records an ATP2B2 variant of uncertain significance co-occurring in an ASH1L proband.

Neither claims DFNA82; neither needed changing.

## Deliberately not curated

Hearing aids and cochlear implantation (no DFNA82-specific outcome data); gene editing as a treatment (preclinical, and in a mouse missense allele); the basolateral neurotransmitter-release arm proposed in the 1998 deafwaddler paper (no human evidence bears on it).

## Checks

Deep research: Edison/falcon, NEC preflight PASS (ATP2B2 mentioned 49 times, CDH23 8 times — preflight keys on the gene and cannot separate the modifier claim from the monogenic one, which is why the separation above was done by hand against the cached sources).

`just validate` exit 0, 49/49 snippets verified. `validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history` all exit 0.

Closes #10402

### DIFF
--- a/history/disorders/Hearing_Loss_Autosomal_Dominant_82/2026-09-01T194138Z-claude-code-112b84.yaml
+++ b/history/disorders/Hearing_Loss_Autosomal_Dominant_82/2026-09-01T194138Z-claude-code-112b84.yaml
@@ -1,0 +1,48 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hearing_Loss_Autosomal_Dominant_82
+  path: kb/disorders/Hearing_Loss_Autosomal_Dominant_82.yaml
+session:
+  id: 2026-09-01T194138Z-claude-code-112b84
+  timestamp: '2026-09-01T19:41:38Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10402
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10421
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - description
+  - mechanistic_hypotheses
+  - pathophysiology
+  - phenotypes
+  - progression
+  summary: 'Review response: DFNA82 progression rate, vestibular prose, tinnitus,
+    and a full-text pass'
+  details: |
+    Second session on this entry. The first record (2026-09-01T180746Z) reports 49/49 snippets and is left unedited, since history is append-only.
+
+    Review on PR #10421, all four findings correct, and three of them came from the full text of PMID:30535804 rather than its abstract - which is where I had stopped.
+
+    Critical. The entry stated that no rate in decibels per year had been published, and the paper this PR caches reports one: an age-related typical audiogram giving average annual threshold deterioration over ages 10 to 70 of 0.5 dB/year low, 1.1 dB/year middle, 0.7 dB/year high. That is now quoted. The reviewer's accompanying nuance is the more useful half and is written into the note: the middle frequencies deteriorate fastest while onset is high-frequency-weighted, so onset configuration and rate of decline are different quantities and the entry must not be read as claiming they agree. Audiograms below age 10 were excluded as unreliable, so the rate describes established disease.
+
+    The same progression phase was evidenced by a sentence about mice graded HUMAN_CLINICAL. Regraded MODEL_ORGANISM with directness: INDIRECT and kept alongside the human rate.
+
+    Vestibular prose. The abstract states the negative more strongly than the results section, which reports only minor vestibular abnormalities, two subjects with complaints (benign paroxysmal positional vertigo, and perilymph leakage after cholesteatoma surgery - both with identifiable non-genotypic causes), saccular function unmeasurable in three ears and utricular in one, and one family untested. The description now follows the results section, the unmeasurable-function sentence is carried as NO_EVIDENCE, and the reasoning for retaining frequency: EXCLUDED is written out: minor abnormalities with attributable causes are not a vestibular phenotype, but this is a judgement on incomplete testing rather than a clean negative.
+
+    Tinnitus added at OCCASIONAL (2 of 11, 18 percent), with the denominator and its single-family concentration stated.
+
+    Own full-text pass, taking the reviewer's closing suggestion. Three further findings. All five DFNA82 alleles fall in the w/a isoform of PMCA2, the predominant calcium pump of hair cells which localises to stereocilia - the reason this is a cochlear rather than a systemic calcium-handling disease, and missing from the calcium-extrusion node. ATP2B2 has a pLI of 1.00 and DOMINO predicts a dominantly inherited disorder, which is the population-genetic argument for the haploinsufficiency model, added there and graded COMPUTATIONAL since it is a prediction rather than a measurement in these patients.
+
+    Most importantly, the CDH23 hedge. The entry said the founding study sequenced CDH23 and found it did not explain the phenotype - true and incomplete. Rare CDH23 variants co-occurred with the ATP2B2 allele in ALL FIVE index cases, and the authors state that a modifying effect of those variants cannot be excluded. Monogenic causation and a superimposed modifier are not mutually exclusive claims, and the original framing let the second disappear. Both are now curated and the ALTERNATIVE hypothesis's description says the entry asserts the first without denying the second, which is a better statement of the modifier-versus-monogenic boundary than the version this PR opened with.
+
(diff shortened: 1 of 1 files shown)


kevinschaper:
All four findings were correct. Fixed in 04aa66017c. Your closing note is the right diagnosis of how they happened: I built the entry from the abstract and did not go back through the 1113-line full text I had committed.

**Critical — the progression rate.** The entry claimed no dB/year rate had been published, and the paper this PR caches reports one. Now quoted: average annual threshold deterioration over ages 10-70 of 0.5 dB/year low, 1.1 dB/year middle, 0.7 dB/year high.

Your nuance is the more useful half and it is now written into the note rather than left implicit: the **middle** frequencies deteriorate fastest, while onset is high-frequency-weighted and the audiogram is downsloping. Onset configuration and rate of decline are different quantities in this disease, and the entry now says so explicitly so it cannot be read as claiming the high frequencies also decline fastest. I also recorded that audiograms below age 10 were excluded from the calculation as unreliable, so the rate describes established disease rather than its first years.

**The mouse sentence graded HUMAN_CLINICAL.** Regraded `MODEL_ORGANISM` with `directness: INDIRECT`, and kept alongside the human rate rather than replaced by it — it is the comparison the authors themselves drew, and it is worth keeping visible that the human finding was framed against a mouse prior.

**Vestibular prose.** You are right that the abstract states this more strongly than the paper's own results section. The description now follows the results section: "only minor vestibular abnormalities", two subjects with complaints (BPPV, and perilymph leakage after cholesteatoma surgery — both with identifiable non-genotypic causes), saccular function unmeasurable in three ears and utricular in one, one family untested. The unmeasurable-function sentence is carried as `supports: NO_EVIDENCE`, since for those ears the measurement was simply not obtained.

`frequency: EXCLUDED` stays, and the reason is now written out instead of implied: minor abnormalities with attributable causes are not a vestibular phenotype, but this is a judgement made on incomplete testing rather than a clean negative, and a reader deserves to know which.

**Tinnitus.** Added at `OCCASIONAL` (2 of 11, 18 percent), with the denominator and the fact that both subjects are in one family stated in the description rather than hidden behind the band.

53/53 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` all exit 0.

kevinschaper:
I took your closing suggestion and read PMID:30535804 end to end. Three more things were in there, added in 0aadbd86b3. Note this pushes a new commit, so the approval on the previous head is superseded — sorry for the extra round.

**The CDH23 hedge, which is the one that matters.** My entry said the founding study sequenced CDH23 and found it did not explain the phenotype. That was true and incomplete. Rare CDH23 variants co-occurred with the ATP2B2 allele in **all five** index cases, and the authors state plainly: *"We cannot exclude a modifying effect of the CDH23 variants on HI in the affected subjects in our study."*

Monogenic causation and a superimposed modifier are not mutually exclusive claims, and my original framing let the second one disappear. Both are now curated, and the ALTERNATIVE hypothesis's description says the entry asserts monogenic causation without denying a modifying effect. That is a more accurate statement of the boundary than the version this PR opened with.

**Isoform specificity.** All five DFNA82 alleles fall in the w/a isoform of PMCA2, which is the predominant calcium pump of hair cells and localises to stereocilia. That is why this is a cochlear disease rather than a systemic calcium-handling one, and it was missing from the calcium-extrusion node.

**Constraint.** pLI 1.00, indicating extreme intolerance of losing one copy, plus DOMINO predicting a dominantly inherited disorder. Added to the haploinsufficiency hypothesis and graded `COMPUTATIONAL` — it is a population-genetic prediction, not a measurement in these patients.

57/57 snippets verified. `validate-disorders`, `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` all exit 0.

PR #10423, opened 2026-09-01
curate: Add Cornelia de Lange Syndrome 4 (MONDO:0013864, RAD21)

### LINKED ISSUE #10405: Curate Cornelia de Lange syndrome 4 (MONDO:0013864)
Curate a dismech entry for **Cornelia de Lange syndrome 4** ([MONDO:0013864](https://monarchinitiative.org/MONDO:0013864)) — *RAD21* variants; the cohesin ring subunit, giving a milder CdLS-like phenotype than the NIPBL-driven classic form.

**First decide what this should be, and treat this as the substantive question.** `kb/disorders/Cornelia_de_Lange_Syndrome.yaml` already exists and mentions RAD21, and RAD21 also appears in `Mullegama-Klein-Martinez_Syndrome.yaml` and `WAPL-Related_Developmental_Disorder.yaml`. So this is a standalone-vs-`has_subtypes` call on the CdLS umbrella, not a blank slate.

Check what the umbrella actually binds rather than what it mentions — a gene named in prose with no bound term and no gene-specific pathograph is coverage in appearance only, which is what the cblA case (#10260) turned on. If the RAD21 phenotype is genuinely milder and mechanistically distinguishable from NIPBL haploinsufficiency, say so with evidence; if it is not, `SUBTYPE` plus a stub deletion is a completed curation.

Stub: `stubs/Cornelia_De_Lange_Syndrome_4.yaml`

kevinschaper:
Adds a knowledge-base entry for **Cornelia de Lange syndrome 4** (MONDO:0013864, heterozygous *RAD21*) — the form of the cohesinopathy spectrum that is defined by being milder than the classic disease, and by a gene that produces two different diseases depending on how many alleles are hit.

Closes #10405

## The mechanism, and why it is a different lesion

NIPBL, the gene behind classic CdLS, is the factor that **loads** cohesin onto chromatin; its disease is a dosage problem. RAD21 is the kleisin subunit that **closes the ring** — its N-terminus engages SMC3, its C-terminus SMC1A, its central domain recruits STAG1/STAG2 — and CdLS4 variants act at exactly those interfaces. 42% of published variants sit in the RAD21–SMC1A domain.

```
Heterozygous RAD21 variant
            |
            v
Disruption of the cohesin ring at the RAD21 subunit interfaces
            |
      +-----+-----------------+
      v                       v
Cohesin-dependent        Impaired DNA
transcriptional          damage response
dysregulation            (no clinical correlate)
      |
      v
Attenuated Cornelia de Lange phenotype
```

The allelic gradient runs the counterintuitive way: **dominant missense alleles cause worse findings than loss-of-function alleles**, which argues that the missense protein interferes with the ring rather than failing to join it.

## Entry-type decision

Standalone `Disease`. The issue asked what the umbrella *binds*, not what it mentions, so:

**The umbrella binds RAD21 properly.** `hgnc:9811` appears on `Cornelia_de_Lange_Syndrome.yaml`'s "Cohesin pathway gene disruption" node *and* on its own `genetic:` record carrying the ClinGen Definitive assertion. This is **not** the cblA situation of a gene named only in prose.

Three things nonetheless argued for a separate entry:

1. **The umbrella has no `has_subtypes` block at all.** Folding CdLS4 in would mean inventing a subtype axis, not filling a slot.
2. **The umbrella's pathograph is one generic node listing six cohesin genes together.** It has nowhere to record ring-versus-loader, the missense-worse-than-LOF gradient, or the attenuation — and the attenuation is the entire content of this entity.
3. **RAD21 is two diseases.** Biallelic RAD21 does not give a more severe CdLS; it gives **Mungan syndrome**, an enteric neuromyopathy with chronic intestinal pseudo-obstruction, whose mechanism runs through specific transcriptional targets (failure to bind the *APOB* promoter, reduced *RUNX1*, enteric neuron deficit in `rad21a` morphant zebrafish). A one-paragraph subtype line has nowhere to put that.

dismech already keeps gene-level cohesinopathies standalone — `Mullegama-Klein-Martinez_Syndrome` (STAG2), `WAPL-Related_Developmental_Disorder`, `CHOPS_Syndrome`, `CTCF-related_Neurodevelopmental_Disorder`, `Warsaw_breakage_syndrome`. **The umbrella was not modified.**

## The claim this entry does not make

An earlier draft argued that the international CdLS consensus score under-detects CdLS4, reasoning from the systematic review having to exclude patients scoring under 4. Then I read the dedicated RAD21 series, which says the opposite in as many words:

> "No RAD21 variant would have been missed using the CdLS consensus criteria for molecular studies"

Every cohort-A patient scored at least five. The claim was removed rather than kept as a plausible-sounding inference. What **both** sources support is that clinical *recognition* is harder here, and that is what the diagnosis node now says — with all three quotes curated so the disagreement is visible rather than resolved by fiat.

## The attenuation, in numbers

The best single figure is not in the abstract of anything:

> "The majority of RAD21 patients (16/29, 55%) have normal or mildly impaired cognitive functioning (SMC1A group 32%; NIPBL group 7%)"

And from the comparison table, the starkest contrast in the entry:

> "Major limb malformation HP:0001180/HP:0009776 **0/29** 0 0/49 0 **17/67 25**"

No major limb malformation in any of 29 RAD21 patients, against 25% of NIPBL patients. Visual impairment likewise 0/24, against 53% (SMC1A) and 44% (NIPBL). Synophrys — the single most recognisable CdLS sign — 68% vs 91%.

Note the phenotype is not uniformly diluted: **short 5th finger is 82% in RAD21 against 67% in NIPBL**, so this is a different shape, not a faded copy. That is stated in the phenotype description.

**The ClinGen assertion is cited for what it says and no more.** `CGGV:assertion_30ad5023…` is keyed to **MONDO:0016033**, the umbrella term, and contains no statement about the RAD21 phenotype. It is cited twice — inheritance and gene-disease validity — and both `explanation` fields say explicitly that it does not support the attenuation claim, which rests on PMID:22633399 and PMID:32193685 instead. (`just clingen-refresh` confirmed RAD21 carries exactly one ClinGen assertion; see the tooling note below.)

**Facial features are one phenotype, not five.** Long philtrum, thick eyebrows, long eyelashes, synophrys and thin upper vermilion all come from one table in one cohort. They are curated as a single record bound to the long-philtrum term — the feature the band is set from — with the individual figures in the evidence, rather than five records quoting one source five times.

**Table rows are quoted with their column order stated.** The Krab comparison table's columns are RAD21 (n = 29), SMC1A (n = 51), NIPBL (n = 67), per its own caption. Every `explanation` that quotes a row names which column it is reading. This is a PDF-derived cache and a mis-read column would be invisible otherwise.

**The Deardorff sentence carries three claims with two evidence sources.** "…act at the RAD21 interface… impair cellular DNA damage response, and disrupt transcription in a zebrafish model" is a structural claim, a cell-based claim and a zebrafish claim in one sentence. It is quoted in three non-overlapping clauses so each item carries a single `evidence_source`.

**The DNA damage arm has no downstream edge.** It is well demonstrated and connects to no clinical feature of CdLS4, so it hangs off the ring node without pretending to explain anything.

## MONDO placement worth reporting upstream

MONDO:0013864 has two parents: *Cornelia de Lange syndrome*, which is right, and **non-syndromic limb reduction defect**, which is not — CdLS4 is a syndromic multiple-anomaly condition, and its limb involvement is mild brachydactyly with *zero* major limb malformation in 29 patients. Recorded in `notes:` and deliberately not propagated into the entry, the way #10260 handled the cblA/cblB synonym bug. Not fixed here.

## Named entity confusion

preflight-dr returns PASS (RAD21 = 73, next gene NIPBL = 8), which reflects the report rather than the literature. Four off-target bodies excluded by hand:

1. **Somatic RAD21 in cancer** — recurrent in AML/MDS, amplified in solid tumours. None cited. The one review used (PMID:32687945) spans both roles; only its statements about normal RAD21 function are quoted, both graded `INDIRECT`.
2. **RAD21L**, the meiotic paralog, is a different gene.
3. **8q23–q24 contiguous-gene deletions** spanning RAD21 and EXT1 overlap Langer-Giedion; not monogenic RAD21 disease, no phenotypes imported.
4. **Mungan syndrome** is RAD21 but is not this disease — confined to the differential and the genetic notes, with its own evidence; none of its features appear in `phenotypes:`.

## Tooling defect

`just clingen-refresh` **downloads `gene_validity.csv` successfully and then exits non-zero** on a drifted sha256 against the pinned snapshot, leaving the file on disk. That was enough to confirm RAD21 has exactly one ClinGen assertion and that it is the umbrella-keyed one already in `references_cache/`. The manifest was not repinned and no new ClinGen cache file was generated.

Separately, the reference validator's `literal_bracket_patterns` strips `[32/34]`-style fractions from snippets but not from the cached text, so every frequency sentence in PMID:39286962 fails as a verbatim quote. Frequencies were re-sourced to bracket-free prose and to comparison-table rows rather than paraphrased.

## Not curated

No progression block — no natural history study exists, and the age-at-diagnosis distribution describes ascertainment, not course. No biochemical, imaging or datasets block. Growth hormone response is tabulated in PMID:39286962 but reported as a summary column rather than an analysed outcome, so no treatment claim is made from it. A `Cohesinopathies` grouping does not exist in `kb/groupings/` and would be the natural home for this entry alongside the other five cohesinopathy entries — separate work.

## Validation

| Gate | Result |
|---|---|
| `just validate-disorders` | exit 0, **43/43 snippets verified** |
| `check-folded-hyphens`, `check-environmental-evidence` | exit 0 |
| `just validate-history` | exit 0 |

Cache diff is two tool-generated rows for the one CURIE this entry introduces, `MONDO:0013864`.

### DIFF
--- a/history/disorders/Cornelia_de_Lange_Syndrome_4/2026-09-01T192854Z-claude-code-4db36a.yaml
+++ b/history/disorders/Cornelia_de_Lange_Syndrome_4/2026-09-01T192854Z-claude-code-4db36a.yaml
@@ -1,0 +1,30 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Cornelia_de_Lange_Syndrome_4
+  path: kb/disorders/Cornelia_de_Lange_Syndrome_4.yaml
+session:
+  id: 2026-09-01T192854Z-claude-code-4db36a
+  timestamp: '2026-09-01T19:28:54Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - phenotypes
+  - diagnosis
+  - treatments
+  - prevalence
+  summary: 'Review fixes: mine GeneReviews, add eight phenotypes including the behavioural
+    contrast'
+  details: |
+    Acts on the ai4c-reviewer CHANGES_REQUESTED on PR #10423. GeneReviews PMID:20301283 was tagged but cited by no evidence item; it now supplies four: the mild-end-of-spectrum attenuation statement on the attenuated phenotype node, the diagnostic-criteria statement naming RAD21, and two management statements on Multidisciplinary Supportive Care, which previously had no evidence block at all. Genetic Counseling asserted a 50% recurrence risk that no cited snippet stated; the GeneReviews sentence that does state it is now cited. Added eight phenotype records from the Krab comparison table the entry already quotes: a combined Behavioural Phenotype record carrying anxiety 53 percent, attention deficit 35 percent, obsessive-compulsive behaviour 32 percent, autistic-like features 35 percent, and the two sharpest attenuation contrasts in the entry, self-injurious behaviour 6 percent against 77 percent in NIPBL and aggression 6 percent against 80 percent; plus microcephaly 57 percent, ptosis 42 percent (higher than NIPBL), short nose 88 percent, congenital heart disease 39 percent, gastro-oesophageal reflux 52 percent, hirsutism 38 percent and clinodactyly of the fifth finger 54 percent, each with a downstream edge from the attenuated phenotype node. The facial-gestalt preferred_term and description previously listed different feature sets and now list the same three, with short nose split out as the one facial feature that is not attenuated. Added Kaur PMID:37377026 as INDIRECT evidence for attenuation across non-NIPBL cohesin genes. The prevalence note now records the 6 of 422 solved-proband share and states that it sits in that paper's full text rather than its cached abstract, so it is not quoted as evidence. Added a notes paragraph recording why pan-CdLS trials are not curated on a subtype entry. Validation: 61/61 snippets verified, all offline gates exit 0.
--- a/kb/disorders/Cornelia_de_Lange_Syndrome_4.yaml
+++ b/kb/disorders/Cornelia_de_Lange_Syndrome_4.yaml
@@ -325,10 +325,59 @@ pathophysiology:
       malformation in any of 29 RAD21 patients, against 25% of NIPBL patients.
       Limb reduction is one of the cardinal features of classic CdLS and it is
       absent here.
+  - reference: PMID:20301283
+    reference_title: Cornelia de Lange Syndrome.
+    supports: SUPPORT
+    evidence_source: OTHER
+    directness: DIRECT
+    snippet: >-
+      Individuals with a milder phenotype have less severe growth, cognitive,
+      and limb involvement, but often have facial features consistent with CdLS.
+    explanation: >-
+      GeneReviews' description of the mild end of the CdLS spectrum, which is
+      the position CdLS4 occupies. It matches the RAD21 series independently:
+      growth, cognition and limbs attenuated, facial gestalt retained.
+  - reference: PMID:37377026
+    reference_title: "Genomic analyses in Cornelia de Lange Syndrome and related diagnoses: Novel candidate genes, genotype-phenotype correlations and common mechanisms."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    directness: INDIRECT
+    snippet: >-
+      Pathogenic variants in cohesin genes other than NIPBL tend to result in a
+      less severe phenotype.
+    explanation: >-
+      A 716-proband cohort makes the same attenuation point at the level of the
+      whole non-NIPBL group. Graded INDIRECT because it is a statement about
+      cohesin genes other than NIPBL collectively rather than about RAD21
+      specifically.
   downstream:
   - target: Intellectual Disability
     causal_link_type: DIRECT
     description: The cognitive component, attenuated relative to classic CdLS.
+  - target: Behavioural Phenotype
+    causal_link_type: DIRECT
+    description: >-
+      The behavioural component, and the sharpest attenuation in the entry.
+  - target: Microcephaly
+    causal_link_type: DIRECT
+    description: Postnatal head growth restriction.
+  - target: Ptosis
+    causal_link_type: DIRECT
+    description: >-
+      One of the few features more frequent in RAD21 than in NIPBL disease.
+  - target: Congenital Heart Disease
+    causal_link_type: DIRECT
+    description: Cardiac malformation, major or minor.
+  - target: Gastroesophageal Reflux Disease
+    causal_link_type: DIRECT
+    description: The commonest gastrointestinal manifestation.
+  - target: Hirsutism
+    causal_link_type: DIRECT
+    description: >-
+      Excess body hair, markedly less frequent than in the comparison genes.
+  - target: Clinodactyly of the 5th Finger
+    causal_link_type: DIRECT
+    description: The other digital feature alongside the short fifth finger.
   - target: Delayed Speech and Language Development
     causal_link_type: DIRECT
     description: Verbal developmental delay, near-universal in reported patients.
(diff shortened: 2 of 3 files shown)


kevinschaper:
Both remaining items are fixed, and you were right that neither needed new literature — every quote was already in a committed cache.

**GeneReviews is now mined.** PMID:20301283 supplies four evidence items:

- The mild-end-of-spectrum sentence on the attenuated phenotype node — *"Individuals with a milder phenotype have less severe growth, cognitive, and limb involvement, but often have facial features consistent with CdLS."* It matches the RAD21 series independently: growth, cognition and limbs attenuated, facial gestalt retained.
- The DIAGNOSIS/TESTING sentence naming RAD21, on the diagnosis node.
- Two MANAGEMENT statements on `Multidisciplinary Supportive Care`, which previously had **no `evidence` block at all**. The surveillance one — *"behavioral assessment for anxiety, attention deficits, and aggressive or self-injurious behaviors"* — turns out to map exactly onto the behavioural phenotype below, which is a better outcome than citing it generically.
- The 50% recurrence sentence on `Genetic Counseling`. You were right that the treatment asserted a figure no cited snippet stated.

**The behavioural gap is closed, and it is the best thing in the entry now.** One `Behavioural Phenotype` record bound to `HP:0000739 Anxiety` (53%, the band-setting figure) carrying attention deficit 35%, obsessive-compulsive behaviour 32%, autistic-like features 35% — and the two rows you pointed at: **self-injurious behaviour 6% against 77% in NIPBL, aggression 6% against 80% in SMC1A**. You are right that these are a sharper contrast than the limb-malformation row, and the `explanation` says so, because this is the concrete content of the "especially for cognition and behavior" claim the entry quotes five times.

I kept them as one record rather than six. The source reports them as rows of a single behavioural assessment, not as six independent diagnoses, and splitting would have quoted one table six times. If you would rather see them separated, say so and I will split them.

**Seven more phenotypes from the same table**, each with a `downstream` edge from the attenuated-phenotype node so none of them is disconnected: microcephaly 57%, ptosis 42%, short nose 88%, congenital heart disease 39%, GERD 52%, hirsutism 38%, clinodactyly of the 5th finger 54%. Ptosis and cardiac are curated next to the short-fifth-finger observation as you suggested — they run higher than NIPBL, which is the point.

**Suggestions, all acted on:**

- **Short nose (88%)** is now its own record rather than folded into the gestalt, because it is the highest-frequency facial feature *and* the only one that is not attenuated (87% in NIPBL). Burying it in a combined record would have hidden that.
- **The facial `preferred_term`/description mismatch** is fixed; both now name the same three features the evidence quotes.
- **Kaur 2023.** Added as `INDIRECT` evidence for attenuation — *"Pathogenic variants in cohesin genes other than NIPBL tend to result in a less severe phenotype"* from a 716-proband cohort, graded INDIRECT because it is about the non-NIPBL group collectively. On the 6/422 figure: I checked the cache and **it is in the full text, not the abstract**, and `PMID_37377026.md` is `content_type: abstract_only`. So the prevalence note now records the number and states where it lives and why it is not quoted, rather than either omitting it or quoting text the cache does not contain. It is also a diagnostic-cohort proportion measured through a CdLS-shaped clinical suspicion, which an attenuated phenotype passes less often — worth saying next to the number.
- **Pan-CdLS trials** now have their own `notes` paragraph recording why they are not curated: none is RAD21-specific or reports a RAD21 subgroup, and curating one on a subtype entry would assert a link its registration record does not make. Same test as the named-entity exclusions.

One quotation mechanic worth flagging for the next reviewer of this file: the anxiety row alone is three words and fails `check-snippet-length`, so it is quoted together with the adjacent obsessive-compulsive row, and the `explanation` says why.

Re-validated: `just validate-disorders` exit 0 with **61/61 snippets verified**, and `check-causal-targets`, `check-entity-refs`, `check-duplicate-keys`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-reference-titles`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens` all exit 0.

PR #10424, opened 2026-09-01
curate(VWM5): resolve as SUBTYPE of Leukoencephalopathy With Vanishing White Matter

### LINKED ISSUE #10404: Curate leukoencephalopathy with vanishing white matter 5 (MONDO:0957873)
Curate a dismech entry for **leukoencephalopathy with vanishing white matter 5** ([MONDO:0957873](https://monarchinitiative.org/MONDO:0957873)) — biallelic *EIF2B5* variants, the commonest genetic cause of VWM, in which a partly disabled eIF2B leaves the integrated stress response chronically mis-set and white matter vulnerable to febrile and traumatic stress.

**First decide what this should be, and treat this as the substantive question.** `kb/disorders/Leukoencephalopathy_With_Vanishing_White_Matter.yaml` already exists and cites EIF2B5. So the real choice is a standalone entry, a `has_subtypes` entry on that file, or already-covered.

Read that entry first and check what it actually binds: the pattern seen repeatedly here is a gene *named in prose* with no bound term and no gene-specific pathograph, which is coverage in appearance only. The cblA case (#10260) turned on exactly that, and the deciding argument was whether the pathograph genuinely diverges — for one subunit of a shared complex, it may well not. `SUBTYPE` or already-covered plus a stub deletion is a completed curation, so long as the reasoning is recorded.

Stub: `stubs/Leukoencephalopathy_With_Vanishing_White_Matter_5.yaml`

kevinschaper:
Resolves **leukoencephalopathy with vanishing white matter 5** (MONDO:0957873, *EIF2B5*) as an `entry_type: SUBTYPE` of the existing `Leukoencephalopathy_With_Vanishing_White_Matter` entry. No new disorder file; the reasoning is recorded in the stub.

Closes #10404

## The finding, in one paragraph

Which of the five eIF2B subunit genes is mutated does not predict the disease. Two independent cohorts say so directly. Fogli 2004 (PMID:15136673), 83 individuals: *"Disease severity was correlated with age at onset (p < 0.0001) but not with the type of the mutated subunit nor with the position of the mutation within the protein."* Deng 2021 (PMID:34745209), 341 patients and 180 distinct mutations: *"The gene (EIF2B1-5) in which the mutation located, and the number of null alleles were not associated with age of onset in these patients."* EIF2B5 is the **common** locus — 64% of mutations in the 2004 series — not a distinct entity. So there is no EIF2B5-specific pathograph to write.

## What the umbrella actually binds

The issue was right to make me check, and the answer is that this is **not** the cblA pattern of a gene named in prose with nothing behind it:

| | |
|---|---|
| `genetic:` | A dedicated EIF2B5 record with the gene term bound to **`hgnc:3261`**, its own `HP:0000007` autosomal recessive inheritance block, and PMID:11704758 evidence from the original subunit-mutation paper — alongside equivalent bound records for EIF2B1, EIF2B2, EIF2B3 and EIF2B4 |
| `pathophysiology:` | Rooted at **"eIF2B Translation Initiation Complex Deficiency"** — a node about the complex, not any one subunit — running through deregulated ISR, astrocyte dysfunction, white matter rarefaction and cavitation, to stress-provoked deterioration |
| `has_subtypes:` | Antenatal-Infantile / Early-Childhood / Juvenile-Adult |
| `disease_term` | MONDO:0800448, the parent VWM term |

The asymmetry the issue flagged holds up: unlike a complementation group with its own upstream lesion, EIF2B5 is one subunit of a complex whose other four subunits give the same disease through the same node.

## Why it must not become a `has_subtypes` entry either

This is the part I would not have got right without checking the file. The parent's subtype axis is **onset**, which is the axis the natural-history literature actually stratifies on, and three phenotypes carry `subtype:` foreign keys into it. `subtype:` is single-valued. Dropping a gene-keyed subtype into the same list mixes two orthogonal axes in one namespace and makes a patient who is simultaneously *EIF2B5* and *Juvenile-Adult* unrepresentable.

MONDO:0800448 has nine children on both axes at once — three onset terms, one phenotype term (ovarioleukodystrophy) and five per-gene terms VWM1–VWM5. dismech already covers both, using `has_subtypes` for onset and `genetic:` for gene. That is a better fit than MONDO's flat sibling list, and it should not be flattened to match it.

## What *does* vary, and where it belongs

The specific allele matters even though the gene does not, and the variation is **within** EIF2B5 as much as across genes: p.Arg113His homozygotes are among the mildest patients and p.Thr91Ala homozygotes among the more severe, both in EIF2B5. Deng 2021 adds that eIF2Bε p.Arg195His and p.Arg269Gln predict earlier onset while eIF2Bγ p.Ala87Val predicts later. That is `GeneticContext` detail on the existing EIF2B5 record, not a new disease file.

## Why the stub is kept rather than deleted

MONDO:0957873 is bound nowhere in `kb/`, so deleting the stub would let the next `just seed-stubs` run recreate it verbatim as `UNDECIDED` and lose this judgement — the same reason `stubs/Dyskinesia_With_Orofacial_Involvement.yaml` is kept, and consistent with the other three `SUBTYPE` stubs already in the queue, none of which was deleted.

The clean way to retire it is a `mappings.mondo_mappings` block on the parent with `skos:narrowMatch` for all five per-gene terms, which would retire the whole sibling set at once instead of one stub at a time. **I deliberately did not do that**, for three reasons: it edits the parent entry, which this PR otherwise leaves untouched; it should cover all five VWM terms rather than only the one I was asked about; and the house convention for narrowMatch direction is not settled — `PIK3CA_Mutant_Breast_Cancer.yaml` uses it narrow-entry-to-broad-term, which is the opposite of what is needed here. That is a maintainer call, and the stub's `notes:` says so, so whoever makes it does not have to rediscover any of this.

## The larger finding: the parent entry is behind the literature

Worth more than a new file would have been, and recorded in the stub as leads. Not curated here, because it means editing an entry this PR is not about:

- **Guanabenz is no longer investigational-in-mice.** The parent's ISR-modulation treatment says it is not an established human therapy and cites only 2017/2019 mouse work. A completed single-arm phase 1/2 trial now exists (PMID:42586097). The entry has no `clinical_trials:` block.
- **2025 consensus management guidelines exist** (PMID:41232062), including genetic and MRI diagnostic criteria that map onto a `definitions:` block the entry lacks, and a categorisation of clinically used drugs by their effect on the ISR — which is what `environmental[].influences_mechanisms` with `environmental_effect: EXACERBATES` is for.
- **The astrocyte-primary framing has been revised** by cell-specific *Eif2b5* conditional mice (PMID:40326783), in which the oligodendrocyte-specific mutant matches the whole-body phenotype and the astrocyte-specific one shows only very mild ataxia.
- No `animal_models:` block despite a well-characterised model set, no `conforms_to` anywhere (twelve other leukodystrophies conform to `cns_myelin_failure`), and `prevalence` still uses the deprecated `percentage:` field.

These are separable from the lump/split call and worth doing whichever way it went. Happy to open an issue against the parent entry if that is useful, or to take it as a follow-up.

## Deep research

A falcon report was run for VWM5 and is committed as an artifact. `just preflight-dr` returns **WARN, exit 0**: EIF2B5 mentioned 36 times with the next gene at 4, and one expected warning — the report cites OMIM 603896 while MONDO:0957873 xrefs 620315. 603896 is the *gene* MIM for EIF2B5 and 620315 is the *phenotype* MIM; preflight compares gene MIM against phenotype MIM, so this is the same false positive recorded in #10260, not a defect. The report is on-entity. It did **not** change the disposition, which was reached from the two cohort papers. (An earlier version of this paragraph said the report independently reached the same conclusion about subunit identity not predicting phenotype; on re-reading it does not address that question at all — it says EIF2B5 accounts for the majority of cases and that severity tracks age at onset. Corrected rather than left standing.)

## Validation

`just validate-stubs` — exit 0, no issues. `just check-stubs` — exit 0, 1363 stubs, no errors. The diff is one stub file plus the deep-research artifact; no `kb/` entry, cache row or reference cache is touched, and **no existing entry was modified**.

### DIFF
--- a/stubs/Leukoencephalopathy_With_Vanishing_White_Matter_5.yaml
+++ b/stubs/Leukoencephalopathy_With_Vanishing_White_Matter_5.yaml
@@ -43,11 +43,29 @@ notes: >-
   in which the mutation located, and the number of null alleles were not
   associated with age of onset in these patients." EIF2B5 is the common locus,
   carrying 64% of mutations in the 2004 series, not a distinct entity. What does
-  predict severity is the specific allele and where it sits in the complex -- and
-  that variation is entirely within EIF2B5 as well as across genes, since
-  p.Arg113His homozygotes are among the mildest patients and p.Thr91Ala
-  homozygotes among the more severe. That belongs in `GeneticContext` on the
-  existing EIF2B5 record, not in a new file.
+  predict severity is the specific allele, and that variation runs within EIF2B5
+  as well as across genes. Two sources, quoted here with the attribution a later
+  curator will need:
+
+  van der Lei et al. 2010 (PMID:20975056), a cross-sectional study of EIF2B5
+  genotypes drawn from a 184-patient database: "Patients homozygous for
+  p.Arg113His had a milder disease than patients compound heterozygous for
+  p.Arg113His and patients homozygous for p.Thr91Ala." Both alleles are in
+  EIF2B5, so the mild and the severe ends of that comparison sit in the same
+  gene.
+
+  Deng et al. 2021 (PMID:34745209): "Certain mutations such as
+  eIF2Be[Arg195His] and eIF2Be[Arg269Gln] that were predicted to have a serious
+  influence on eIF2B structure were related to earlier age of onset.
+  EIF2Bg[Ala87Val] which was predicted to have a minimal influence on eIF2B
+  structure, was related to later age of onset." Note that the same paper found
+  eIF2Be[Thr91Ala] did NOT correlate with age of onset in its cohort, which is
+  not in conflict with van der Lei -- different comparison, different endpoint --
+  but is the reason the Thr91Ala contrast is attributed to van der Lei above and
+  not to Deng.
+
+  Either way the finding is allele-level, so it belongs in `GeneticContext` on
+  the existing EIF2B5 record, not in a new file.
 
   Why it must not become a has_subtypes entry either. The parent's
   `has_subtypes` axis is onset-based -- Antenatal-Infantile, Early-Childhood,
@@ -68,9 +86,9 @@ notes: >-
   MONDO:0957871, MONDO:0957872, MONDO:0957873), which would retire the whole
   sibling set at once. That is deliberately left to a maintainer: it edits the
   parent entry, it should cover all five rather than one, and the house
-  convention for narrowMatch direction is not settled (Pik3ca_Mutant_Breast_
-  Cancer uses it narrow-entry-to-broad-term, the opposite of what is needed
-  here). Retire this stub when that mapping is added, or when MONDO:0957873 is
+  convention for narrowMatch direction is not settled: see
+  `kb/disorders/PIK3CA_Mutant_Breast_Cancer.yaml`, which uses it
+  narrow-entry-to-broad-term, the opposite of what is needed here. Retire this stub when that mapping is added, or when MONDO:0957873 is
   otherwise bound in kb/.
 
   Separately, the parent entry is behind the literature in ways that are worth


kevinschaper:
Filed the parent-entry follow-up as #10427, covering the six gaps listed in the PR body — the guanabenz phase 1/2 trial and the missing `clinical_trials:` block, the 2025 consensus guidelines (which map onto a `definitions:` block and an ISR-exacerbating `environmental[].influences_mechanisms` list), the revised cell-type story from the cell-specific *Eif2b5* mice, the absent `animal_models:` block, the absent `conforms_to`, and the deprecated `percentage:` field plus the missing `mappings:` block.

That last one is the mechanism that would retire this stub properly: `skos:narrowMatch` on all five per-gene MONDO terms at once, rather than one stub at a time. I have deliberately left it there rather than doing it here, since it edits the parent entry and the narrowMatch direction convention is unsettled.

Items 1 and 2 are the ones with clinical consequence and are independent of this disposition.

kevinschaper:
Fixed, and you were right to make it a round-trip rather than a note. Thank you for checking the allele against Deng's table rather than taking the sentence on trust — that is exactly the failure mode the note would have propagated.

**On the Thr91Ala claim.** It was not fabricated, but it was uncited, which in a work order that says "put this in `GeneticContext`" is the same problem. The source is **van der Lei et al. 2010, PMID:20975056** — a cross-sectional study of EIF2B5 genotypes from a 184-patient database, whose abstract says verbatim:

> "Patients homozygous for p.Arg113His had a milder disease than patients compound heterozygous for p.Arg113His and patients homozygous for p.Thr91Ala."

I had read that paper while making the disposition and then wrote the conclusion into the note without carrying the attribution with it. That is precisely how a later curator ends up attaching Fogli or Deng to it.

The note now quotes both sources with their PMIDs, and — because you found the apparent contradiction — says explicitly why they do not conflict: van der Lei compares Arg113His homozygotes against Thr91Ala homozygotes on disease severity, while Deng tests each allele independently for correlation with age of onset and finds Thr91Ala non-significant. Different comparison, different endpoint. Deng's *significant* pair (eIF2Bε Arg195His / Arg269Gln earlier, eIF2Bγ Ala87Val later) is now quoted too, so whoever actions the `GeneticContext` work has both findings and the right citation for each.

**Suggestion — the split filename.** Fixed. `>-` folded the line break into a space, so the pointer parsed as `Pik3ca_Mutant_Breast_ Cancer` — not greppable and miscased. It now reads `kb/disorders/PIK3CA_Mutant_Breast_Cancer.yaml` on one line. A note whose whole job is handing a maintainer a precise pointer should have a working pointer.

**Suggestion — `entry_type: SUBTYPE` against its schema definition.** You have identified something real: the enum has no value for "already covered, but on a different axis than `has_subtypes`". I agree `SUBTYPE` is the least-wrong option and that the notes disambiguate it, so I have left it. If it recurs it is worth a schema issue; I would rather see whether it does than add a value on one instance.

**Suggestion — the PR body overstates the deep-research corroboration.** Correct, and I have struck it. Checking the report again, it says EIF2B5 accounts for the majority of cases and that severity tracks age at onset; it does not address whether *which subunit* predicts phenotype. The disposition stands on Fogli and Deng, and the body should not have leaned on the report for that specific point.

**On the parent-entry leads — issue filed as #10427**, before this review landed, covering the guanabenz phase 1/2 trial, the 2025 consensus guidelines, the revised cell-type story, the absent `animal_models:` and `conforms_to`, and the deprecated `percentage:` field. I have added your two additions to it: `NCT05757141` (fosigotifator, open-label Phase 1b/2) and the fuller model set including the Eif2b5 I98M line and the patient iPSC astrocytes.

`just validate-stubs` and `just check-stubs` both re-run clean after the edit.

PR #10434, opened 2026-09-01
curate: add Ethylene Glycol Poisoning (MONDO:0017861)

caufieldjh:
## What

New entry for **ethylene glycol poisoning** (`MONDO:0017861`). Absent from the KB. Conforms to `drug_induced_nephrotoxicity` at three nodes.

## Bioactivation is the spine

Ethylene glycol is only mildly intoxicating by itself. Alcohol dehydrogenase oxidises it to glycolaldehyde, then glycolic acid, then oxalic acid. The metabolites do the damage.

> *"The effects of these substances … are due to their metabolites, which can cause metabolic acidosis and cellular dysfunction."* — PMID:18045860

That is why the antidote is an enzyme inhibitor, and why **fomepizole attaches to the trigger node** rather than to any downstream injury.

```
Ingestion of ethylene glycol
  → ADH bioactivation                    [conforms: Nephrotoxic Drug Exposure and Tubular Uptake]
      ├─→ Glycolic acid / high anion gap acidosis → Metabolic acidosis
      └─→ Oxalic acid / calcium oxalate deposition → Nephrocalcinosis
              → Proximal tubular cell death        [conforms: Proximal Tubular Epithelial Cell Death]
              → Acute kidney injury                [conforms: Acute Kidney Injury]

Fomepizole   ──INHIBITS──> ADH bioactivation
Hemodialysis ──INHIBITS──> Glycolic acid accumulation
```

5 nodes, 3 conformances, 10 edges, no orphans.

## On the conformance, against the sibling entry

This is the first conformer in this batch where the module's **acute** time course fits without qualification. `Balkan_Endemic_Nephropathy`, curated against the same module earlier in this session, deliberately declined the module's `Acute Kidney Injury` node because its course is chronic. Here that node is exactly right.

The agent caveat still holds and is recorded in the trigger node's `notes`: the module's trigger names a nephrotoxic *drug*, and this is an ingested industrial alcohol. Conformance rests on the shared mechanism.

## Two things modelled rather than left in prose

**The two gaps move in opposite directions.** The osmolal gap tracks unmetabolised parent alcohol; the anion gap tracks metabolites. A sample drawn at the wrong hour shows one and not the other.

> *"Accumulation of the alcohols in the blood can cause an increment in the osmolality, and accumulation of their metabolites can cause an increase in the anion gap and a decrease in serum bicarbonate concentration."*

**Oxalate outlasts the poison.** In the cited case the parent compound was cleared within 36 hours by dialysis, while plasma oxalate peaked on day 3 and normalised only on day 7 (PMID:18696123). Crystals keep forming after the alcohol has gone.

## Treatments carry their conditions

Fomepizole's value is entirely conditional on timing, so the entry says so and evidences it: given early it *"prevents EG-related renal failure"* and *"may obviate the need for hemodialysis"*; given late it cannot undo deposition that has already happened. Hemodialysis attaches to the acid-accumulation node, not the enzyme step, because what it removes is the alcohol and the organic acid anions.

## Deep research

Run with `claude_code`; the report will be committed to this branch as provenance when it lands. **No quote from it was curated.** All eleven snippets were independently sourced from PubMed and pre-verified against the local cache before writing, and every `reference_title` was generated from cache rather than typed.

## Validation

- `just count-verified-snippets` — **11/11**
- `just check-duplicate-keys`, `just check-folded-hyphens` — clean
- Graph — 10 edges, no integrity issues, no phenotype without an incoming edge

### DIFF
--- a/kb/disorders/Ethylene_Glycol_Poisoning.yaml
+++ b/kb/disorders/Ethylene_Glycol_Poisoning.yaml
@@ -356,6 +356,21 @@ phenotypes:
       id: HP:0002901
       label: Hypocalcemia
   evidence:
+  - reference: PMID:36765419
+    reference_title: "Extracorporeal treatment for ethylene glycol poisoning: systematic review and recommendations from the EXTRIP workgroup."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: >-
+      neither oxalate crystals in urine, urine immunofluorescence nor hypocalcemia
+      nor a history of EG exposure alone are indications for ECTR (but may help to
+      diagnose EG poisoning)
+    explanation: >-
+      Names hypocalcaemia among the features that help diagnose this poisoning in
+      humans. This is the item that establishes the phenotype occurs; the
+      chemistry item below explains why, and cannot carry the claim on its own.
+      Note that PMID:18696123, also cited in this entry, reports a case in which
+      hypocalcaemia did NOT occur, so the finding is characteristic rather than
+      invariable.
   - reference: PMID:15695020
     reference_title: "The cytotoxicity of oxalate, metabolite of ethylene glycol, is due to calcium oxalate monohydrate formation."
     supports: SUPPORT
@@ -363,9 +378,12 @@ phenotypes:
     snippet: >-
       Oxalate has two forms in vivo: oxalate ions and calcium oxalate monohydrate
       (COM) crystals that readily form in the presence of calcium.
+    directness: INDIRECT
     explanation: >-
-      Establishes that crystal formation consumes calcium, which is the mechanism
-      by which the oxalate arm lowers serum calcium.
+      Mechanistic companion, not the support for the phenotype. It establishes
+      that crystal formation consumes calcium, which is why the oxalate arm lowers
+      serum calcium, and it is a cell-culture chemistry observation that cannot
+      establish that the finding occurs in patients.
 - name: Hyperoxaluria
   phenotype_term:
     preferred_term: Hyperoxaluria
@@ -391,16 +409,19 @@ phenotypes:
       id: HP:0020074
       label: Crystalluria
   evidence:
-  - reference: PMID:18696123
-    reference_title: "Hyperoxaluria after ethylene glycol poisoning."
+  - reference: PMID:36765419
+    reference_title: "Extracorporeal treatment for ethylene glycol poisoning: systematic review and recommendations from the EXTRIP workgroup."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: >-
-      alkaline citrate was given as a preventive measure to increase urinary
-      oxalate solubility, but nephrocalcinosis still developed.
+      Diagnosed by plasma EG concentration in 24 patients, and by urinary calcium
+      oxalate crystals in 6 patients
     explanation: >-
-      Records that raising urinary oxalate solubility was attempted and did not
-      prevent deposition, which is the clinical counterpart of the crystal arm.
+      Names urinary calcium oxalate crystals in patients, which is what this
+      phenotype asserts. An earlier draft cited a sentence about nephrocalcinosis
+      here; renal parenchymal deposition is not crystals in the urine, and that
+      sentence remains correct only where it sits on the Nephrocalcinosis
+      phenotype.
 biochemical:
 - name: Plasma oxalate
   notes: >-


caufieldjh:
All six landed in the latest push.

**Item 5 first, because it was a false claim about my own work.** The PR body was headed *"Two things modelled rather than left in prose"* and both things sat entirely inside `description:` free text. I wrote that sentence and did not check it against the file. Both are now structure: three `biochemical` entries carrying the osmolal gap, the anion gap and plasma oxalate with its reference range, and a four-phase `progression` block with time windows.

**Item 4 changed a node, not just filled a gap.** The citation you pointed me at repaid the trip. Guo and McMartin show that calcium oxalate monohydrate **crystals**, not the oxalate ion, cause the membrane damage and cell death in human proximal tubule cells — oxalate held in solution with EDTA is not cytotoxic. And:

> *"The cytotoxic effects of COM to HPT cells were potentiated by acidosis (pH 6.5), but not by glycolate, the major metabolite of ethylene glycol."*

So the two downstream branches are **not parallel**. The acid arm makes the crystal arm more lethal in the same cells, and glycolate itself is not the potentiator. That is now in the central effector's description with both quotes, and the node carries the module's `GO:0008219` and `GO:0006915` that it was missing.

Your instinct to send me to `just fetch-reference` rather than the research paraphrase was right — the paraphrase would not have carried the ion-versus-crystal discrimination, which is the part that makes crystal deposition causal rather than a marker of oxalate load.

**Item 1 — six phenotypes.** The neurologic arm was absent entirely, which is a real hole given it is stage one and the commonest presentation: altered mental status, coma, seizure. Plus hypocalcaemia, hyperoxaluria and crystalluria, all consequences of the oxalate node already modelled. Hypocalcaemia is wired directly from it with the reason stated — the calcium is consumed forming the crystals. All nine phenotypes now have incoming edges.

**Item 2 — ethanol added**, targeting the same enzyme node as fomepizole, with the practical difference recorded rather than implied: infusion, concentration monitoring, and its own intoxication to manage.

**Item 3 — `NCIT:C15248`.** Two-line fix, as you said. `therapeutic_modality: DEVICE` added.

**Item 6 — four-phase `progression`** with the windows: neurologic 0–12 h, cardiopulmonary 12–24 h, renal from ~12 h, delayed neurologic day 5–20. The last phase is the one the research calls the part clinicians miss, and it is worth having because it appears after apparent recovery.

Phenotypes 3 → 9. Edges 10 → 17. Snippets 11/11 → 20/20.

Gates: `just validate-disorders` passed, `check-duplicate-keys`, `check-entity-refs` and `check-folded-hyphens` clean, graph 17 edges with no integrity issues.

caufieldjh:
Pushed. The blocker was worse than you knew, and the fix is a removal rather than a citation.

**The staging windows are not in EXTRIP.** I fetched PMID:36765419 as you asked. It gives *"metabolic acidemia appears after a latent period of approximately 3–6 h after ingestion"* and *"Cranial nerve palsies, radiculopathy, and other neuropathies may appear several days after ingestion, despite treatment"*. That is all the temporal detail in the paper.

My four phases — 0–12 h, 12–24 h, 12 h onward, day 5–20 — came from the deep-research report. I wrote them as fact. Citing EXTRIP for them would have been worse than leaving them bare, because it would have made an unsourced number look sourced.

So the phases are rewritten to what the paper states: initial intoxication resembling ethanol, acidaemia after a 3–6 h latent period, a multiorgan phase with AKI at 30–70%, and delayed neuropathy several days later despite treatment. A `notes` block on the delayed phase records the windows I removed and says not to reintroduce them from this source.

I also cut *"is the commonest presenting feature"* from the `Confusion` description. That was a frequency claim standing on nothing. You gave me the number — 237/446 — and it is now in the description with the cohort named.

**Evidence coverage: 9/9 phenotypes and 4/4 progression phases.** Snippets 20/20 → 33/33. Both phenotypes you offered from cache are used — PMID:29427181 for altered mental status, PMID:18696123 for nephrocalcinosis, the latter with the detail that alkaline citrate was given and nephrocalcinosis developed anyway.

Your framing was exactly right: *"a claim carrying a number or a time window should cite the source of that number."* Worth adding what this case shows, since it generalises. I have been saying across this session that I curate nothing from deep research. That was true of quotes and false of these numbers, and it was false in a way nothing could catch. Fabricated quotes get caught by the report's own validator. A contradicted claim got caught by a reviewer. A number copied into a `duration:` field carries no snippet, so `count-verified-snippets` has nothing to check and reports clean. The gap is structural, not a lapse in care.

**Suggestions not taken this round**, and I would rather name them than let them read as done: `biochemical[].readouts` edges to connect the three markers to their nodes, `biomarker_term` and `loinc_term` bindings, and the cardiopulmonary and delayed-neuropathy phenotypes. The readout edges are the one I think is worth doing — the markers being isolated in a 21-node graph is a real defect, not a stylistic one — but it needs the `target` semantics checked rather than guessed, and I would rather do that deliberately than at the end of a review round.

Gates: `just validate-disorders` passed, `check-duplicate-keys`, `check-entity-refs`, `check-folded-hyphens` clean.

caufieldjh:
Both fixed. 34/34 snippets, `just validate-disorders` passes.

**Item A — crystalluria.** You are right and the diagnosis is exact: I reused the nephrocalcinosis sentence, and the quote does not mention urine at all. Repointed to the EXTRIP line reporting urinary calcium oxalate crystals in patients. The original sentence stays where it belongs, on `Nephrocalcinosis`.

**Item B is the one I would not have found, and the reason is worth stating.** The hypocalcaemia phenotype was supported by a cell-culture statement about oxalate forming crystals with calcium. That explains a mechanism. It does not establish the finding occurs in patients. And as you point out, PMID:18696123 — cited elsewhere in this same entry — reports a case in which hypocalcaemia *did not* occur.

So the entry asserted a human finding while carrying, among its own references, one arguing against it and none arguing for it. Every snippet verified. Every reference real and on topic. `count-verified-snippets` reported clean throughout, because the defect was not in any quote but in the relationship between three of them.

Now supported by the EXTRIP line naming hypocalcaemia as a diagnostic feature. The Guo & McMartin item stays as a `directness: INDIRECT` companion, with its `explanation` saying what it can and cannot carry, and the primary explanation records that the finding is characteristic rather than invariable — since a cited case lacked it, and an entry that hides that is worse than one that says it.

That is twice in two rounds you have caught the same class: evidence that verifies and does not support. It is the failure mode with no gate behind it, and it is the one worth me slowing down for. My habit has been to verify each snippet against its cache and move on. What that misses is whether the sentence says the thing the field claims, and whether the entry's own references disagree with each other.

Remaining suggestions unchanged and still not taken: `biochemical[].readouts` edges, `biomarker_term` / `loinc_term` bindings, and the cardiopulmonary and delayed-neuropathy phenotypes. The readout edges are the one I think is a real defect rather than a preference — three markers with no edges in a 21-node graph — and I would rather check the `target` semantics deliberately than guess at the end of a round.

Issue #10168, opened 2026-08-30
Adopt a serial GitHub merge queue for PR Shepherd


cmungall:
PR Shepherd should enqueue eligible PRs into a GitHub merge queue instead of repeatedly updating each PR branch until it contains the exact current `main` tip. The queue should be configured to build and merge one PR at a time. This preserves current-main integration testing without branch-update churn, repeated CI, or dismissed reviews.

## Finding

The repository currently uses loose required checks: `test (3.13)` must pass, but branch protection has `strict: false`, so a manually merged PR does not have to contain the latest `main`. That is a valid throughput/safety tradeoff, but a green check only proves compatibility with the version of `main` used by that check.

PR Shepherd is stricter than branch protection. Its deterministic controller requires the PR head to contain the live `main` SHA before merging. In the 2026-08-30 snapshot:

- 224 PRs were open.
- 218 were behind the exact current `main`.
- 59 of 61 approved PRs were behind.
- No PR met every deterministic merge predicate.

The strict rule also creates a retry loop:

```text
update PR branch
  -> CI and review run
  -> another ordinary or generated PR moves main
  -> the PR is behind again
  -> update and repeat
```

#9538 demonstrated both sides of the problem. Updating it once exposed a real cross-file incompatibility after an enum changed on `main`, so simply removing all integration safety would be wrong. It then needed several additional update/review cycles because other merges moved `main` while its checks ran. The controller eventually merged it successfully, but the process was unnecessarily expensive.

GitHub's merge queue is designed for this case: it tests a temporary `merge_group` containing the latest base branch plus the queued PR, without requiring the PR branch itself to be updated.

GitHub documentation:

- [Managing a merge queue](https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue)
- [Events that trigger workflows: `merge_group`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#merge_group)
- [Strict versus loose required checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging)

### Queue configuration

Require a merge queue for `main` with:

- squash merge;
- build concurrency of 1;
- minimum and maximum merge size of 1;
- only non-failing PRs allowed;
- an appropriate required-check timeout.

[...]

Issue #2034, opened 2026-05-04
Track impact assessment for April 23 GitHub merge queue incident


cmungall:
## Summary

GitHub reported an April 23, 2026 merge queue incident where pull requests merged through merge queue with the squash merge method could produce incorrect merge commits when a merge group contained more than one PR. A later article raised visibility of the incident, so we should keep a repository-local record of whether/how `monarch-initiative/dismech` was exposed.

Current assessment: **no evidence that `dismech` was affected**, but this issue is for tracking, provenance, and transparency. We can publish a fuller analysis later.

## External references

- GitHub status incident: https://www.githubstatus.com/ — "Incident with Pull Requests" on April 23, 2026.
- GitHub availability post: https://github.blog/news-insights/company-news/an-update-on-github-availability/
- Article that prompted this check: https://dev.to/varshithvhegde/github-broke-git-the-merge-queue-bug-that-silently-deleted-your-code-4f7i

## Incident conditions to check

Based on GitHub's public writeup, exposure required all of these:

- Pull requests merged through **merge queue**.
- The merge queue used **squash merge**.
- The merge group contained **more than one PR**.
- The affected merge occurred during the impact window: **2026-04-23 16:05-20:43 UTC**.

GitHub stated that PRs merged outside merge queue were not affected, and merge queue groups using merge or rebase methods were not affected.

## Local checks run

These checks were run from a local worktree for `monarch-initiative/dismech` on 2026-05-04.

Repository and branch state:

```bash
git fetch --all --prune
git status --short --branch
git remote -v
git branch --show-current
```

Branch protection / repo settings:

```bash
gh repo view monarch-initiative/dismech \
  --json nameWithOwner,defaultBranchRef,mergeCommitAllowed,squashMergeAllowed,rebaseMergeAllowed,deleteBranchOnMerge,isPrivate

gh api repos/monarch-initiative/dismech/branches/main/protection

gh api repos/monarch-initiative/dismech/rulesets

gh api graphql -f owner='monarch-initiative' -f name='dismech' -f branch='main' \
[...]

PR #10439, opened 2026-09-01
Add merge_group CI support so a merge queue can report test (3.13)

kevinschaper:
Rollout step 1 of #10168: make CI answer the `merge_group` event, so that when a merge queue is enabled on `main`, the required `test (3.13)` check gets reported on the queue's temporary integration commits. Today it does not — we verified this live: a dry-run queue on a scratch branch enqueued a probe PR (#10436), no workflow fired, and the queue ejected it with reason `checks_timed_out` after exactly the configured timeout. This PR changes no merge behavior by itself; enabling an actual queue is a later, separate step.

## What changes

**`main.yaml` declares `merge_group:`** and queue builds run the **full suite**: eleven path-filter-gated steps now also fire on `if: github.event_name == 'merge_group' || ...`. Running everything rather than the path-filtered subset is the point, not just a safety measure — path filtering is why the #9538 cross-file enum incompatibility stayed invisible until the branch was updated, and that class of break is what a queue build exists to catch.

**The paths-filter itself still runs on merge-group refs.** dorny/paths-filter supports the event since v4.0.1 (`base`/`ref` default to the event's commit hashes), and this workflow's `@v4` pin resolves past that. Its changed-file lists drive the two steps that stay filter-gated rather than forced — "Validate changed disorder KB files" and "Validate changed comorbidity KB files" — which consume those lists as command arguments, so forcing their gates would run them with empty file lists. On a queue build they therefore validate exactly the KB files entering `main`, computed from the group's base..head diff, in the merged context where a co-queued PR may have changed `cache/**` or `references_cache/**`.

The two `if: github.event_name == 'pull_request'` steps (curation scope, history coverage) and every `${{ github.event.pull_request.base.ref || 'main' }}` fallback are untouched — the #10168 audit already established those are correct for a merge group.

**Wiring guards** in `tests/test_merge_group_ci.py` pin the trigger; the filter's event coverage and its v4.0.1 version floor (v4.0.0 predates merge-group support, so the floor is checked as a version tuple, not a major); the rule that no `steps.changes.outputs.*` gate may suppress a step on `merge_group`; and the two-step file-scoped list, in both directions — each listed step must actually consume a `*_files` output, and must never be merge_group-forced (which would run it with an empty argument list). A `run:`-body check catches any future step consuming a file list without gating on the matching output. The existing hypothesis-step assertion in `tests/test_ci_batch_validation.py` is updated for the forced gate. A note in `docs/explanation/automation-and-agents.md` records the full-suite decision next to the existing merge-queue discussion.

## What this deliberately does not do

No changes to branch protection, `scripts/auto_merge_ready_prs.py`, or the `auto/` lane workflows — those are rollout steps 3–5 of #10168 and depend on this landing and being verified first (the scratch-branch dry-run setup from the timeout demonstration above is still in place for that verification).

Merge-group runs will cost a full suite (~30–40 min measured) per queued PR. That is the intended price of real integration testing and is why the queue configuration discussion in #10168 recommends build concurrency above 1.

Refs #10168, #2034.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,8 +10,8 @@ on:  # yamllint disable-line rule:truthy
   # Merge-queue integration builds (#10168, #2034). The queue tests each PR on
   # a temporary gh-readonly-queue/* branch and waits for the required
   # `test (3.13)` check on that commit; without this trigger the check is never
-  # reported and every queued PR times out. Merge-group runs deliberately skip
-  # the paths-filter and run the full suite -- see "Check for changes" below.
+  # reported and every queued PR times out. Merge-group runs execute the full
+  # suite regardless of paths -- see "Check for changes" below.
   merge_group:
 
 env:
@@ -73,9 +73,11 @@ jobs:
       # cross-file enum incompatibility was invisible until the branch was
       # updated, and catching that class of break is what the queue is for. A
       # merge group is a handful of runs per merge, so the wall-clock cost is
-      # bounded. (Do not downgrade this action below v4.0.1: on the older
-      # versions a merge-group ref yields an empty filter result, which would
-      # silently skip the changed-file validations.)
+      # bounded. (Do not downgrade this action below v4.0.1: older versions
+      # have no merge_group case and fall back to a merge-base diff against
+      # the default branch, an over-broad result that can include other PRs'
+      # changes -- upstream added the event support in dorny/paths-filter#255
+      # precisely to stop those false positives.)
       - name: Check for changes
         uses: dorny/paths-filter@v4
         id: changes
--- a/docs/explanation/automation-and-agents.md
+++ b/docs/explanation/automation-and-agents.md
@@ -333,20 +333,23 @@ safe drain rate is a merge-queue project—including `merge_group` CI—not a re
 to weaken ancestry checks or batch more branch updates.
 
 The `merge_group` CI half of that project is in place (#10168): `main.yaml`
-declares the trigger, and merge-group runs execute the **full suite** — the
-path-filter gates that scope an ordinary PR run are forced on for queue
-builds, because path filtering is the specific reason the #9538 cross-file
-enum incompatibility stayed invisible until the branch was updated, and that
-class of break is exactly what a queue build exists to catch. The
-dorny/paths-filter step itself still runs on merge-group refs (supported since
-its v4.0.1, with `base`/`ref` defaulting to the event's commit hashes): its
-changed-file lists drive the changed-disorder and changed-comorbidity
-validations, so those run against the KB files actually entering `main`, in
-the merged context where a co-queued PR may have changed `cache/**` or
+declares the trigger, and merge-group runs execute the **full suite** — every
+path-filter gate that scopes an ordinary PR run is forced on for queue builds,
+except the two changed-file KB validations discussed below, because path
+filtering is the specific reason the #9538 cross-file enum incompatibility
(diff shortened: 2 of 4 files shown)


kevinschaper:
Both blocking findings are addressed by adopting the recommended fix, and both non-blocking ones are fixed too.

**Filter now runs on merge_group.** Verified the claim independently: v4.0.1's release is "Support merge queue" (dorny/paths-filter PR 255) and the README documents `base`/`ref` defaulting to the event's commit hashes on `merge_group`. The `if: github.event_name != 'merge_group'` skip is gone, the eleven forced full-suite gates are unchanged, and the two changed-file KB validations now run on queue builds with lists computed from the group's base..head diff. The comments in `main.yaml` and the docs paragraph are rewritten around the real design (full suite by policy, filter for file lists) instead of the v2-era limitation, plus a note not to downgrade the action below v4.0.1.

**The coverage-hole point is now moot in the right direction:** term/reference validation of the queued PRs' KB files runs in the merged context, where a co-queued PR may have changed `cache/**` or `references_cache/**` — the workflow comment above those steps records that as the reason they stay filter-gated rather than forced.

**Non-blocking items:** the collected helper is renamed to `workflow_steps()` (0 warnings now), and a new `test_file_list_consumers_are_gated_on_their_output` inspects `run:` bodies — any step interpolating a `*_files` list must be gated on the matching output, so an ungated consumer running with an empty argument list can't slip past the guards. The version floor (`dorny/paths-filter@v4`+) is asserted in `test_paths_filter_runs_on_merge_group`.

**Not changed:** the two behavioral rollout notes (OAK cache scope on queue refs, ratchets diffing a ref that can advance mid-run) are accurate and belong to the rollout discussion in #10168 rather than this PR; no workflow change follows from them yet.

All 11 workflow tests pass (`tests/test_merge_group_ci.py` + `tests/test_pr_shepherd_workflow.py`).

kevinschaper:
All four findings and both suggestions are addressed, plus the CI failure the review predates.

**Finding 1 (stale trigger comment):** rewritten — it now says merge-group runs "execute the full suite regardless of paths" and the cross-reference to "Check for changes" lands on a comment that agrees with it.

**Finding 2 (stale PR body):** rewritten in place. It now describes the shipped design: filter runs on merge-group refs, eleven gates forced, two file-scoped validations driven by the filter's group-diff lists in the merged context.

**Finding 3 (version floor):** the test now parses the pin into a version tuple and requires either a moving major tag (v4+) or a full version >= (4, 0, 1); v4.0.0 fails it, and a SHA or otherwise non-numeric pin fails with a message telling the pinner to re-vet merge-group support by hand.

**Finding 4 (wrong downgrade rationale):** corrected to the traced behavior — older versions fall back to a merge-base diff against the default branch, an over-broad result that can blame other PRs' changes on the queued PR, citing dorny/paths-filter#255's own false-positive rationale.

**Suggestion (OR-hole in the file-list gate test):** the guard now asserts the inverse too — a `FILE_SCOPED_STEPS` member must never contain the merge-group clause, so forcing one on (and running it with an empty argument list) fails the wiring test.

**Suggestion (docs wording):** the paragraph now names the two exemptions in the same sentence that says gates are forced, and the version-floor credit is accurate now that the test enforces one.

**CI failure:** `test_ci_validates_hypothesis_review_artifacts_on_report_or_yaml_changes` pinned the hypothesis step's old condition verbatim; updated to expect the forced gate. Verified the workflow's condition string matches the new assertion exactly.

PR #10440, opened 2026-09-01
Add Thrombocytopenia-Absent Radius Syndrome (RBM8A, MONDO:0010121)

### LINKED ISSUE #10431: Curate thrombocytopenia-absent radius syndrome (MONDO:0010121)
Curate a dismech entry for **thrombocytopenia-absent radius syndrome** ([MONDO:0010121](https://monarchinitiative.org/MONDO:0010121)) — a congenital malformation and marrow-failure disorder combining bilateral radial aplasia with hypomegakaryocytic thrombocytopenia, caused by reduced RBM8A (Y14, an exon-junction-complex core protein) dosage.

Initial read is that this is a **DISEASE**: one gene, one conserved mechanism, and MONDO records no descendants for the term.

**Why it is not already covered.** Preflight against `origin/main` found no `kb/` entry matching the MONDO ID, the label, or `RBM8A`, and no open PR. The nearest existing entries are `Inherited_Aplastic_Anemia.yaml` (which lists Fanconi anemia, dyskeratosis congenita, Diamond-Blackfan, Shwachman-Diamond and congenital amegakaryocytic thrombocytopenia, but not TAR) and `Fanconi_Anemia.yaml`. Note that a `"TAR syndrome"` grep matches `Scimitar_Syndrome.yaml` as a substring — that hit is spurious.

**Two things to get right in the curation:**

1. **Inheritance is not simple autosomal recessive.** Most patients are compound heterozygous for a null allele — commonly the 1q21.1 microdeletion — *in trans* with a low-frequency hypomorphic regulatory SNP in the 5'UTR or first intron of `RBM8A`. Modelling this as plain AR would misrepresent it. The `inheritance` block and per-variant `genetic_context.functional_impact_category` should carry the two-allele structure explicitly, with its own cited evidence.
2. **Thumbs are present.** Bilateral radial aplasia with preserved thumbs is the phenotype that separates TAR from Fanconi anemia and from radioulnar synostosis with amegakaryocytic thrombocytopenia, where thumbs are absent or hypoplastic. This belongs in the entry as curated phenotype content, not only as prose.

Stub: `stubs/Thrombocytopenia-absent_Radius_Syndrome.yaml`

evanc1648:
Adds a new disease entry for thrombocytopenia-absent radius (TAR) syndrome. TAR combines bilateral absent radii with a low platelet count, and is defined clinically by the thumbs being present — which is what separates it at the bedside from Fanconi anemia and the other radial ray syndromes that also affect the blood. It is caused by having too little Y14, a protein encoded by RBM8A.

Closes #10431. Deletes the stub it curates.

## The mechanism, in short

```
Reduced RBM8A gene dosage
  └─> Y14 protein insufficiency
        └─> Exon junction complex insufficiency
              ├─> Megakaryocyte maturation failure ──> Megakaryocytopenia
              │                                   └─> Thrombocytopenia
              │                                         ├─> Petechiae
              │                                         └─> Abnormal bleeding
              │                                               └─> GI hemorrhage
              └─> Radial ray developmental failure ─> Bilateral radial aplasia
                                                  └─> Abnormality of upper limb
```

The genetics are unusual and the entry models them explicitly. Patients carry a rare null allele — normally a recurrent 200-kb deletion at 1q21.1 — together with a common low-frequency regulatory variant in RBM8A on the other chromosome. Neither allele causes disease by itself: people carrying only the deletion are healthy, and so are people carrying only the regulatory variant. That is why the pedigrees do not look like a textbook recessive disorder, and why the entry does not describe this as haploinsufficiency.

**Why the combination is pathogenic is not settled, and the entry no longer pretends otherwise.** Albers et al. say their data are "compatible with both a dose-effect phenomenon and a lineage dependent deficiency of Y14". An earlier draft blended those into one asserted mechanism. They are now two `mechanistic_hypotheses`, both `ALTERNATIVE`, and the two causal edges where the models make different predictions opt into both groups. Note that the `mLOE` classification on the regulatory variants presupposes the lineage-dependence model, so it is not neutral between them.

**TAR is listed among the inherited bone marrow failure syndromes, but the label is misleading here.** The NCI natural-history cohort (NCT00027274) enrolls it as an IBMFS, and the entry cites that trial. But the literature states plainly that "TAR syndrome is not generally associated with bone marrow failure or malignancy", and a 2017 review counted four reported leukemia cases in the whole literature. The thrombocytopenia is also lineage-restricted and improves with age rather than progressing to pancytopenia. This is recorded as an `INTERPRETATION` discussion, and the trial description says enrollment reflects nosological grouping rather than a cancer predisposition, so nobody infers Fanconi-like surveillance from the label.

## What is in the entry

5 pathophysiology nodes, 14 phenotypes, 4 diagnostic criteria, 5 treatments, 2 variants, 2 differential diagnoses, 2 datasets, 2 clinical trials, 1 animal model, 2 mechanistic hypotheses and 3 discussions, across 76 evidence items.

Some smaller decisions a reviewer may want to check:

- Preserved thumbs are recorded under `diagnosis`, not as a phenotype. HPO is an ontology of abnormalities and a normally formed thumb is not one, so there is no term to bind; the finding matters only as a discriminator, which is a diagnostic role.
- Intracranial hemorrhage is deliberately absent. The mortality figures behind it reach this entry only through a book chapter whose cache is `content_type: unavailable` and cannot be quoted, so the general `HP:0001892` Abnormal bleeding term carries the severity claim, cited to a source that states it.
- Six phenotypes are deliberately left as pathograph leaves — the leukocyte and eosinophil findings, uterine aplasia, and the lower-limb, renal and cardiac anomalies. No source establishes a route from Y14 insufficiency to any of them, and every edge in this entry carries cited evidence. The entry's `notes` records this so it reads as a decision rather than a gap.
- Nine of eleven candidate datasets were rejected on relevance. They matched on the gene RBM8A and are about gastric cancer knockdowns and neurogenesis splicing panels, not this disease.
- The two dataset records carry `publication` and `notes` rather than `evidence`, per the dataset curation policy.
- The zebrafish model is graded `FAILS_TO_RECAPITULATE`: morpholino knockdown is lethal at two days and produces global malformation, so it does not model the partial, tissue-selective deficit seen in patients.

## An open question left recorded, not resolved

One cited case report (PMID:41925074) describes its patient as homozygous for `RBM8A c.-21G>A`, whereas this entry models compound null-plus-hypomorph inheritance and GeneReviews states biallelic hypomorph carriers are asymptomatic. One case report is not enough to revise an inheritance model and it may be a reporting simplification, so it is flagged in the discussion notes for whoever revisits the allelic spectrum — which Strauss 2023 says is still open.

## Provenance

A falcon (Edison) deep-research report is included in its own commit. It was run after the entry was curated from primary sources, so it served as a gap-finder rather than a source, and everything taken from it was re-fetched and verified independently. It contributed the prenatal care section, two clinical trials, and the correction that the regulatory allelic spectrum is still open rather than limited to two SNPs. Several of its own references carry no PMID or DOI and were not used.

Three references were initially cited by DOI and converted to PMIDs, because `linkml-reference-validator` does not check snippets on DOI-prefixed references. That conversion caught a real problem: one source renders "Wiskott-Aldrich" with an en dash and PubMed with a hyphen, so the quoted snippet had stopped being an exact substring.

## Validation

`just validate-disorders` passes. 75 of 75 snippets verified against cached references, none skipped by prefix. The full pytest suite passes (47,250 tests). All offline gates pass: causal targets, entity refs, duplicate keys, enum values, qualifier terms offline and online, source-defect claims, folded hyphens, snippet length, title snippets, snippet grading, environmental evidence, reference titles, cache integrity and ordering, dataset accessions, and stub well-formedness.

Compliance is 97.7% with no threshold violations against `conf/qc_config.yaml`. The four unpopulated fields are evidence on the two datasets and two external assertions, where the config sets no minimum.

https://claude.ai/code/session_01Ca8Pd6t5QVK9H2LujdaT1v


evanc1648:
All four blocking items are addressed, and all six suggestions were taken. One thing was deliberately not done: intracranial hemorrhage is still not modelled, for the reason below.

## Blocking items

- **GeneReviews tag.** Added a top-level `references:` block tagging PMID:20301781 with `tags: [GeneReviews]`.
- **NSAID and aspirin caution.** Now a supportive-care treatment linked to the bleeding phenotype. The quote was already in the cache, and I had verified it earlier in the session without ever using it, so this was a straight miss.
- **Severe bleeding.** Added `HP:0001892` Abnormal bleeding and `HP:0002239` Gastrointestinal hemorrhage.
- **Upper-limb anomalies.** Added `HP:0002817` Abnormality of the upper limb, using the GeneReviews sentence that also covers the genitourinary features.

**The bleeding item turned out to need no new fetch.** You expected one because the deep-research report's hemorrhage statistics trace to the De Alarcón book chapter, whose cache is `content_type: unavailable` and therefore unquotable. But PMID:41925074, already cited here for the JMML case, carries the claim directly:

> Mortality in TAR syndrome, mainly due to hemorrhage, is generally confined to the first year of life

Gastrointestinal hemorrhage is separately supported by PMID:36872795, which documents a TAR neonate presenting with frank rectal bleeding.

## What I did not do

**Intracranial hemorrhage is not added.** No cached source states it. The 21/77 mortality and first-week figures are real, but they reach this entry only through a book chapter that cannot be quoted, and the two leukemia case reports do not mention it either. Binding `HP:0002170` on the strength of an unquotable secondary figure is the fabrication risk the evidence policy exists to prevent, so the general `HP:0001892` Abnormal bleeding term carries the severity claim instead, cited to a source that actually says it. If a quotable source turns up, this is a one-line addition.

## Suggestions, all taken

Müllerian agenesis (`HP:0000151`), the GeneReviews 25%/50% recurrence-risk figures on the `inheritance` block, the intron-1 allele rs201779890 in `variants`, the mixed in-vitro and patient-platelet snippet split into two items each with one `evidence_source`, the EJC-composition statement regraded from `HUMAN_CLINICAL` to `OTHER`, and the six orphaned `DOI_*.md` caches deleted.

On those caches: my earlier reasoning for keeping them was that the research report cites those DOIs. That does not hold. The report carries its own citations sidecar and its validation section is already recorded in the file, so nothing depends on them.

I also took the leukemoid-reaction gap from the Content-Completeness Checklist, adding periodic leukemoid reaction and eosinophilia from the Hedberg 100-case review (PMID:3056062).

## One defect I introduced and then fixed

The first push after your review included two reference caches that nothing cited, which is the same orphan-cache problem you had just flagged. I caught it on a post-push audit, not from a gate, since no check enforces that a committed cache is referenced. Hedberg is now genuinely used; the other, a dental implant case report, is removed. Every reference cache this branch adds is cited by the entry.

## Validation

`just validate-disorders` passes. 71 of 71 snippets verified, none skipped by prefix. Causal targets, entity refs, duplicate keys, enum values, folded hyphens, snippet length, title snippets, snippet grading, reference titles, cache frontmatter, term-cache integrity and cache ordering all pass. Compliance is 97.6% with no threshold violations; the four unpopulated fields are evidence on the two datasets and two external assertions, where `qc_config.yaml` sets no minimum.

---

evanc1648:
Took the remaining suggestion. Four causal edges added, each with its own evidence item quoted from an already-cached reference:

- `Thrombocytopenia` → `Abnormal bleeding` (sequelae), citing the point that counts improve but never normalise, so the bleeding risk persists.
- `Abnormal bleeding` → `Gastrointestinal hemorrhage` (sequelae).
- `Megakaryocyte Maturation Failure` → `Megakaryocytopenia`. The maturation failure is what marrow examination sees as a reduced megakaryocyte population.
- `Radial Ray Developmental Failure` → `Abnormality of the upper limb`, on the severity spectrum from isolated radial absence to near-total absence of the upper limbs.

Every phenotype you named as cheap and uncontroversial is now connected, including the arguable fourth one.

## The leaves that stay leaves

The leukocyte and eosinophil findings, uterine aplasia, and the lower-limb, renal and cardiac anomalies are still unattached, and the entry's `notes` now says why rather than leaving it to read as an oversight. There is no established route from Y14 insufficiency to any of them, and every pathograph edge in this entry carries cited evidence; attaching them to the nearest plausible node would assert a mechanism no source states.

The lower-limb anomaly is worth naming specifically. The Albers sentence I cited for the upper-limb edge also mentions lower limb defects in passing, so it would have been easy to hang that phenotype off the same node. I did not, because `Radial Ray Developmental Failure` covers forelimb patterning and parenting a lower-limb phenotype to it would be mechanistically wrong even though the quote technically covers both.

## Validation

`just validate-disorders` passes. 75 of 75 snippets verified, none skipped by prefix. Causal targets, entity refs, duplicate keys, enum values, folded hyphens, snippet length, title snippets, snippet grading and reference titles all pass. Compliance 97.7%, no threshold violations.

This push dismisses the approval, since `main` runs `dismiss_stale_reviews`. That was the accepted cost of taking the suggestion rather than a reason to skip it.

---

evanc1648:
You are right, and this was a real error rather than a citation-style quibble. Fixed in `5471ee0d7`.

The snippet I used said nothing about bleeding, and the paper attributes its patient's bloody stool to cow's milk protein allergy that resolved once enteral feeds were held. Citing it for a thrombocytopenia-driven GI bleed argued against the source's own conclusion. Your reading that the entry already knew the distinction is also correct — the phenotype description says the bleeding may be compounded by the milk allergy, and this edge was the one place that nuance got dropped.

Swapped to the reference you identified, PMID:41926074's sibling in this entry, PMID:41925074:

> Investigations confirmed hematochezia and persistent thrombocytopenia, necessitating ongoing platelet transfusions

That is the thrombocytopenic route the edge asserts. I also rewrote the edge description to say explicitly that not every gastrointestinal bleed in TAR is thrombocytopenic in origin, so the milk-allergy route is distinguished rather than silently folded in. The phenotype's own evidence stays on PMID:36872795, which does document a TAR neonate with frank rectal bleeding.

## Worth recording how this got past the gates

The bad snippet was an exact substring of a cached, on-topic paper cited elsewhere in the same entry. Snippet verification passed, `check-source-defect-claims` passed, and it still passes now — the check is green both before and after the fix. Nothing mechanical distinguishes a quote that supports the claim from a quote that merely sits near it in a relevant paper. Only reading the cited paper's conclusion catches it, which is what you did.

## Validation

`just validate-disorders` passes. 75 of 75 snippets verified, none skipped by prefix. Causal targets, entity refs, duplicate keys, enum values, snippet grading and source-defect claims all pass. Compliance 97.7%, no threshold violations.

---

evanc1648:
Both scope items fixed in one commit, `5f4f33642`. You are right on both counts.

**The frozen blob is reverted.** `cache/dataset_accessions.json` now matches `main` exactly — `git diff origin/main...HEAD` reports zero difference for it. It got in through a `git add cache/` in `58022d136`, which is precisely what the targeted-add rule exists to prevent. I also verified your point that the edit was stale: hashing the file before and after a `just verify-datasets` run gives the same hash, so the current script genuinely no longer writes it.

**Both GEO caches are committed.** `GEO_GSE5781.md` and `GEO_GSE240611.md`, generated by the verifier and staged individually rather than by sweeping `cache/` or `references_cache/`.

Worth being straight about the second one: those two files existed in this branch earlier and I deleted them. After the rebase, the verification run regenerated them, and I removed them on the reasoning that they were not part of the already-approved diff and I did not want to expand its scope. That was wrong. Keeping a diff stable does not override a documented requirement, and the approved diff was already missing them — so "matching what was approved" was preserving the defect, not protecting against one.

Your explanation of why three green rounds missed both is the useful part, and it matches what I see locally: `verify-datasets` reported `OK 2 / TOTAL 2` for these accessions both before and after the cache files existed, because it falls back to the network. Nothing in the toolchain distinguishes a committed record from a live fetch.

Final diff is 32 files: the previous 31, minus the frozen blob, plus the two GEO records.

**Validation:** `just validate-disorders` passes with 75/75 snippets verified. Duplicate keys, causal targets, entity refs, enum values, reference-cache frontmatter and cache order all pass.

On the out-of-date-with-base notice: leaving that alone deliberately, since the repo merges through a queue that tests the latest-`main` combination on a temporary merge-group commit, and `CLAUDE.md` states branch freshness is not part of merge eligibility. `main` moved five commits in the hour after the last rebase, so chasing it would not converge.

---

PR #10443, opened 2026-09-01
curate(Progressive_Retinal_Dystrophy_Due_To_Retinol_Transport_Defect): add RBP4 entry (MONDO:0014060)

### LINKED ISSUE #10411: Curate progressive retinal dystrophy due to retinol transport defect (MONDO:0014060)

Claiming **MONDO:0014060** — progressive retinal dystrophy due to retinol transport defect (RDCCAS), *RBP4*.

Stub: `stubs/Progressive_Retinal_Dystrophy_Due_To_Retinol_Transport_Defect.yaml`

**Lump/split check.** MONDO parent is `inherited retinal dystrophy`; no descendants; single causal gene RBP4. Synonyms name the fuller phenotype — retinal dystrophy, iris coloboma, and comedogenic acne syndrome. No existing KB entry covers retinol transport. Curating as `entry_type: DISEASE`.

nlharris:

Curates **progressive retinal dystrophy due to retinol transport defect** (`MONDO:0014060`, *RBP4*) as a new `kb/disorders/` entry. Closes #10411.

## A distribution failure, not a deficiency

RBP4 is the only specific plasma carrier for retinol. Lose it on both alleles and the liver's vitamin A stores stay full while the circulation empties — so the retina is starved of the chromophore precursor it needs to regenerate visual pigment, despite normal intake and normal reserves. That distinction is the whole entry: it is what makes the biochemistry diagnostic, and it is why the obvious treatment may not work.

```
biallelic RBP4 loss of function
  → absent circulating RBP4
      → failure to mobilise hepatic retinol into plasma
          → systemic retinol deprivation, hepatic stores intact
              ├→ chromophore deprivation at the RPE
              │     → impaired rhodopsin regeneration in rods → night blindness
              │     → progressive photoreceptor and RPE degeneration → retinal dystrophy
              └→ reduced retinoid tone in skin → acne
```

The diagnostic signature separates this from ordinary dietary vitamin A deficiency on four axes at once: RBP4 undetectable, retinol about one sixth of normal, retinyl esters and transthyretin normal, and **serum retinol does not rise on oral loading** — because the carrier needed to mobilise it does not exist.

## The therapeutic evidence points both ways, and the entry says so

The strongest translational result in this literature is pharmacological vitamin A rescuing vision in ***Stra6*-null mice**. That is a different gene. Those mice have the plasma carrier and lack the uptake receptor; these patients have the receptor and lack the carrier — the same axis, the opposite lesion. Rescuing one does not establish rescue of the other.

So the treatment link carries both items rather than the flattering one: `PMID:24852372` graded `directness: INDIRECT`, and the founding family's dose-response test — in which serum retinol **did not rise at all** — curated as a `REFUTE` item on the same link. The open question is recorded as a knowledge-gap discussion with a proposed experiment that would settle it in the matching genotype. The *Rbpr2*-null mouse was excluded from `animal_models` on the same reasoning, and its reference cache is not shipped.

## Scope

Only the recessive retinal-degeneration arm. Dominant missense *RBP4* alleles cause the microphthalmia-anophthalmia-coloboma spectrum with skewed maternal inheritance — a different inheritance mode, variant class, developmental timing and proposed mechanism. That arm is recorded in a discussion rather than curated as a subtype.

## Three term corrections against the research report

None of these would have been surfaced by the report's `confabulation_rate: 0.0`, which measures whether an identifier *resolves*, not whether it names the right thing:

| Report said | Actually |
|---|---|
| `HP:0031287` for acne | `HP:0031287` is **Seborrheic keratosis**; correct term is `HP:0001061` |
| `CHEBI:17336` = "retinol / vitamin A", `CHEBI:50211` = "all-*trans*-retinol" | Transposed — `CHEBI:17336` is all-*trans*-retinol |
| — | `GO:0034633` is a biological process, not a molecular function (my error, caught by the dynamic-enum validator) |

The report was otherwise unusually clean: all 11 PMIDs real and on-topic, and all 16 of its quoted fragments verified as exact cache substrings before use. `needs_review: true` flagged the obsolete `GO:0005615`, which is not bound anywhere in the entry.

## Contents

| Section | Detail |
|---|---|
| `pathophysiology` | 7 nodes, scale-tagged, every edge evidenced |
| `phenotypes` | 7 HPO-bound, frequency bands only where the source supports one |
| `biochemical` | RBP4, retinol, and a normal-result record for retinyl esters and transthyretin — normality is the informative finding here |
| `animal_models` | *Rbp4*-null C57BL/6, two `RECAPITULATES` links with readouts; the strain difference is the point |
| `diagnosis` | 3 records including the oral loading test, where the negative result is the diagnostic one |
| `differential_diagnoses` | Dietary vitamin A deficiency, with the acquired phenocopy case cited on both halves of the discriminator |
| `discussions` | 2 `KNOWLEDGE_GAP`, one carrying a proposed experiment |

## Validation

```
just validate                 schema clean · terms clean · 48/48 snippets verified
just check-entity-refs        OK
just check-duplicate-keys     OK
just check-causal-targets     OK
just check-qualifier-terms    OK
just check-snippet-grading    OK    just check-title-snippets   OK
just check-snippet-length     OK    just check-folded-hyphens   OK
just check-enum-values        OK    just validate-history       No issues found
```

Cache diff is 7 additive rows, zero deletions. Stub deleted, history record added. Every cited PMID has a committed cache, reconciled by re-deriving the cited list from the entry rather than from memory.

---

### DIFF
--- a/history/disorders/Progressive_Retinal_Dystrophy_Due_To_Retinol_Transport_Defect/2026-09-01T225327Z-claude-code-ec03a2.yaml
+++ b/history/disorders/Progressive_Retinal_Dystrophy_Due_To_Retinol_Transport_Defect/2026-09-01T225327Z-claude-code-ec03a2.yaml
@@ -1,0 +1,42 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Progressive_Retinal_Dystrophy_Due_To_Retinol_Transport_Defect
+  path: kb/disorders/Progressive_Retinal_Dystrophy_Due_To_Retinol_Transport_Defect.yaml
+session:
+  id: 2026-09-01T225327Z-claude-code-ec03a2
+  timestamp: '2026-09-01T22:53:27Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10443
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - biochemical
+  - pathophysiology
+  summary: 'Review response: structure biochemical readouts, connect visual acuity
+    to the pathograph'
+  details: |
+    Round-2 response on PR #10443, which was approved with three non-blocking suggestions. Two are taken here; the third is declined with reasoning, in the same push rather than deferred.
+
+    Structured readouts added to the two biochemical records that genuinely measure a pathograph node. Serum RBP4 is a READOUT_OF the Absent Circulating Retinol-Binding Protein 4 node; serum all-trans-retinol is a READOUT_OF Systemic Retinol Deprivation with Intact Hepatic Stores. Both carry direction NEGATIVE and their own evidence, so the entry's diagnostic thesis is now machine-queryable rather than living only in notes prose.
+
+    The reviewer's suggested snippet would not have validated. It proposed relationship REFLECTS and direction DECREASED; BiomarkerReadoutRelationshipEnum permits READOUT_OF, CORRELATES_WITH, PREDICTS and PHARMACODYNAMIC_MARKER_OF, and BiomarkerReadoutDirectionEnum permits POSITIVE, NEGATIVE, PRESENT_ABSENT and THRESHOLD_DEPENDENT. The direction enum here is an association direction, not a direction of change, which is why an assay whose value falls in disease is recorded as NEGATIVE: a higher measured value means less of the linked event. The intent of the suggestion is implemented with the valid values.
+
+    The third biochemical record, serum retinyl esters and transthyretin, deliberately gets no readouts. Both results are normal, so the record is a negative finding that localises the lesion rather than a measurement of any node in the graph. Attaching a readout would assert a link the data denies.
+
+    Reduced visual acuity connected into the pathograph as a downstream target of Progressive Photoreceptor and Retinal Pigment Epithelial Degeneration, citing the acuity measurement already present on the phenotype. The reviewer's judgement on the other three unconnected phenotypes is accepted: the colobomas and patent ductus arteriosus are developmental features whose retinoic-acid mechanism this entry deliberately does not assert, and curating a phenotype without asserting an unevidenced edge is the honest handling.
+
+    Declined: HP:0000510 Rod-cone dystrophy in place of HP:0000556 Retinal dystrophy. The reviewer records this as a preference rather than an error. HP:0000556 is retained because it matches the MONDO concept and the disease name, and because the rod-then-cone sequence is described in the node prose where it is evidenced rather than asserted through a term binding.
+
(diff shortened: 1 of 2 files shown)


nlharris:
Thanks — took both substantive suggestions in one push (`b829c359d`), and I'm declining the third here rather than deferring it. That resets the approval; I judged the biochemical structuring worth it, since you're right that this entry argues harder than most that the biochemistry *is* the diagnosis.

**One correction: the suggested snippet would not have validated.** It used `relationship: REFLECTS` and `direction: DECREASED`, and neither is a permissible value:

- `BiomarkerReadoutRelationshipEnum` — `READOUT_OF`, `CORRELATES_WITH`, `PREDICTS`, `PHARMACODYNAMIC_MARKER_OF`
- `BiomarkerReadoutDirectionEnum` — `POSITIVE`, `NEGATIVE`, `PRESENT_ABSENT`, `THRESHOLD_DEPENDENT`

The direction enum on `BiomarkerReadout` is an **association** direction, not a direction of change — "higher biomarker value tracks with less of the linked event". So an assay whose value *falls* in disease is `NEGATIVE`, which reads backwards until you check the enum description. Worth knowing because `ExperimentalReadout` on model links *does* take `DECREASED`, and the two slots look interchangeable. I implemented the intent with the valid values.

**Suggestion 1 — taken, on two of three records.** Serum RBP4 is now a `READOUT_OF` the *Absent Circulating Retinol-Binding Protein 4* node; serum all-*trans*-retinol a `READOUT_OF` *Systemic Retinol Deprivation with Intact Hepatic Stores*. Each carries its own evidence.

The retinyl-ester and transthyretin record deliberately gets **no** readouts. Both results are normal, so it's a negative finding that localises the lesion rather than a measurement of any node — attaching a readout would assert a link the data denies. Left unstructured for that reason, not by omission.

**Suggestion 2 — taken.** `Reduced visual acuity` now hangs off *Progressive Photoreceptor and Retinal Pigment Epithelial Degeneration*, citing the 20/25–20/40 measurement already on the phenotype. Your reading of the other three is right and I've left them alone: the colobomas and PDA are developmental features whose retinoic-acid mechanism this entry deliberately doesn't assert, and curating the phenotype without the edge is the honest call.

**Suggestion 3 — declined.** `HP:0000556` retained over `HP:0000510`. You record it as a preference rather than an error, and I'd rather the rod-then-cone sequence stay in node prose where it's evidenced than be asserted through a term binding. `HP:0000556` also matches the MONDO concept and the disease name.

**Also not attempted: `biomarker_term`.** These analytes are chemical entities rather than named biomarker concepts, and binding a chemical identifier into a biomarker slot would be the wrong kind of term. Left for someone who can identify a genuine NCIT biomarker concept.

Validation on the pushed tree: `linkml-validate` clean, `linkml-term-validator --labels` clean, **51/51 snippets verified**, and `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-qualifier-terms` plus all five baseline gates OK. Round-2 history record added.

---

PR #10445, opened 2026-09-01
feat: Add LETM1-Related Childhood-Onset Neurodegeneration (MONDO:0859304)

### LINKED ISSUE #10438: Curate neurodegeneration, childhood-onset, with multisystem involvement due to mitochondrial dysfunction (MONDO:0859304)
Curate a dismech entry for **neurodegeneration, childhood-onset, with multisystem involvement due to mitochondrial dysfunction** ([MONDO:0859304](https://monarchinitiative.org/MONDO:0859304)) — biallelic *LETM1* disease, in which loss of the mitochondrial inner-membrane cation/proton exchanger disrupts matrix ion homeostasis and cristae architecture.

**Why this one.** It carries a `CURATE_ROOT` recommendation on the priority dashboard with **no NEC risk flags**, is absent from `kb/`, and has no prior issue or PR. `LETM1` appears in `kb/` only inside `Wolf-Hirschhorn_Syndrome`, where it is one gene in a 4p16.3 contiguous-gene deletion — the recessive single-gene mitochondrial disease is a genuinely new concept for the KB, and the relationship between the two is itself worth curating.

The MONDO parent is the broad `MONDO:0024237` inherited neurodegenerative disorder, so there is no numbered-series parent to lump into; the lump/split question here is whether the phenotype is conserved enough across reported families to carry one pathograph, which the literature review should answer explicitly.

**NEC check:** anchor identity on the OMIM xref and the stub's `genes:` block (`hgnc:6556` LETM1) before curating from any deep-research output, and run `just preflight-dr <report.md> MONDO:0859304`. Take care not to import Wolf-Hirschhorn findings as evidence for this entry.

Stub: `stubs/Neurodegeneration_Childhood-onset_With_Multisystem_Involvement_Due_To_Mitochondrial_Dysfunction.yaml`

Tracker: part of #1079.

AO33:
Adds a knowledge base entry for the recessive LETM1 disease — a childhood-onset mitochondrial disorder caused by inheriting two damaged copies of *LETM1*, the protein that controls how mitochondria balance potassium and acidity across their inner membrane. When it fails, mitochondria swell, their internal folds break down, and energy production drops; children present in infancy with developmental delay, vision and hearing loss, unsteadiness, and seizures. The whole clinical literature is one 2022 study of 18 people from 11 families, and the entry is written to say that plainly rather than imply broader knowledge.

Closes #10438

## This is not Wolf-Hirschhorn syndrome

*LETM1* sits inside the chromosome 4p16.3 region deleted in Wolf-Hirschhorn syndrome, which this repository already curates as `Wolf-Hirschhorn_Syndrome`. They are different diseases and I kept them apart deliberately:

| | Wolf-Hirschhorn syndrome | This entry |
|---|---|---|
| Lesion | De novo deletion of one copy of 4p16.3, removing *LETM1* alongside *NSD2*, *CPLX1*, *PIGG* and others | Two variant *LETM1* alleles, no deletion |
| Inheritance | Sporadic | Autosomal recessive (67% of reported cases consanguineous) |
| Gene dosage | One working copy | Two damaged copies |

No phenotype, prevalence, or evidence content was carried across from the Wolf-Hirschhorn entry. The `notes:` block on the new entry states the relationship so a future curator does not merge them.

The overlap is not zero, though, and it needs care: much of the *LETM1* functional literature was written to test whether *LETM1* explains Wolf-Hirschhorn seizures. I cite that work for what LETM1 the protein does, never for clinical content, and the animal models carry explicit `limitations` saying they were built as Wolf-Hirschhorn haploinsufficiency models rather than models of this recessive disease.

## Lump or split

Split — curate as its own entry. The MONDO parent is the broad `MONDO:0024237` inherited neurodegenerative disorder, so there is no numbered series to fold into.

The 11 reported families do share one reasonably conserved pathograph. Every affected individual had a respiratory chain complex deficiency, and the same osmoregulatory defect was demonstrated three independent ways — in patients' own fibroblasts, in their muscle, and by putting each human allele into *LETM1*-null yeast. What varies is tempo, not mechanism: onset ranges from birth to 8 years, and outcomes run from death before one year to survival at 39. I recorded that spread in `progression` rather than inventing subtypes, because the cohort is far too small to tie a course to an allele class.

## The transport mechanism is genuinely disputed, and the entry says so

What LETM1 actually transports has been argued over for twenty years, so I represented both positions instead of picking one:

- **Potassium/proton exchange** — `CANONICAL`. Established by yeast experiments where the activity disappears without the protein and comes back when the human version is supplied. It is also the arm the human disease study actually measured in patient cells.
- **Calcium/proton exchange** — `ALTERNATIVE`. Proposed from a genome-wide screen with purified-protein reconstitution, but a 2022 paper concluded that a different protein, TMBIM5, is that exchanger and LETM1 is not. That paper is cited as a `REFUTE` item inside the hypothesis group.

A `CONTROVERSY` discussion records the dispute, including a third possibility (that LETM1's core job is mitochondrial translation) and a 2024 review stating the core function "still remains elusive". The main causal chain runs through the potassium arm because that is what was tested in patients. The calcium arm is a separate node marked `HYPOTHETICAL` with no patient-level evidence attached — anyone revising this should not quietly promote it without new human data.

- **No prevalence rate.** Recorded as `CASES_IN_LITERATURE` / `ULTRA_RARE` with a count of 18 individuals. There is no population estimate to cite and I did not manufacture one.
- **No disease-modifying treatment.** There isn't one. Treatments are supportive care, antiseizure medication and genetic counselling. Two mechanistic leads exist — nigericin restored membrane potential in patient fibroblasts, and ketone bodies restored mitochondrial shape but not potential — and both are recorded in the knowledge-gap discussion rather than dressed up as therapy. Nigericin also failed to prevent seizures in the rat model, which I noted.
- **No frequency on some phenotypes.** Developmental regression and cerebellar atrophy are tabulated per individual with several "no data" cells, so a cohort percentage cannot honestly be read off the source.
- **No `HP:0001244`.** The Falcon report suggested it; it does not exist in HPO. Left unbound.

## Model evidence is kept separate from human evidence

Much of this literature is yeast, fly, worm, rat and mouse. Every such item is graded `MODEL_ORGANISM` or `IN_VITRO`, and no human phenotype rests on model evidence alone — all eleven clinical phenotypes cite the human cohort. The mouse model carries the sharpest caveat: the genotype matching human patients is embryonic lethal in mice, so only heterozygotes survive and the model cannot reproduce the human condition at all.

## Module conformance

One node conforms to `mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress`. The conformance note records what does not transfer: that module is an aging-hallmark module whose upstream node is age-related mtDNA mutation, which has nothing to do with a childhood-onset nuclear-gene defect, and the module's reactive-oxygen-species arm was never measured in this cohort. Conformance is declared at that one node only. This follows the pattern already used by `Combined_Oxidative_Phosphorylation_Deficiency_42`.

## Deep research

Falcon (Edison Scientific) ran and succeeded (636s, 26 citations). `just preflight-dr` passed against `MONDO:0859304` with *LETM1* mentioned 72 times.

Consistent with other Falcon reports here, it produced **zero PMIDs** — citations are DOIs and author-year keys. I treated the whole report as leads only: every PMID cited in the entry was fetched with `just fetch-reference` and every snippet checked against the local cache. Four report DOIs resolved to papers I had not found through PubMed and were genuinely useful (the cristae-organisation, PINK1-LETM1, and mitochondrial-DNA-metabolism papers). Its term validation flagged one invented HPO identifier, which I did not use.

## Validation

Every command below was run to completion.

| Check | Result |
| --- | --- |
| `just validate-disorders` (authoritative) | **All 1 disorder file(s) passed** |
| `just validate` | No issues found; terms passed; 58/58 snippets verified |
| `just validate-terms` | Validation passed |
| `just count-verified-snippets` | 58/58 verified |
| `just check-causal-targets` | OK, no new broken targets |
| `just check-qualifier-terms` | OK |
| `just check-enum-values` | OK |
| `just check-stubs` | OK, 1361 stubs, no errors |
| `just check-title-snippets` | OK (after fixing one of mine) |
| `just check-folded-hyphens` | OK (after fixing two of mine) |
| `just check-snippet-length` / `check-snippet-grading` | OK |

Three defects the gates caught in my own work, all fixed: a paper title used as an evidence snippet, two folded-scalar hyphen splits, and a variant snippet that misquoted the family identifier.

Cache changes are six single-row additions (`MONDO:0859304`, `GO:0015386`, `HP:0200125` and their enum rows). `references_cache/PMID_36055214.md` is modified because fetching it retrieved the open-access full text; the change is purely additive and `Wolf-Hirschhorn_Syndrome`, which also cites that paper, still verifies 169/169.

### DIFF
--- a/history/disorders/LETM1-Related_Childhood-Onset_Neurodegeneration/2026-09-02T221123Z-claude-code-a68702.yaml
+++ b/history/disorders/LETM1-Related_Childhood-Onset_Neurodegeneration/2026-09-02T221123Z-claude-code-a68702.yaml
@@ -1,0 +1,30 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: LETM1-Related_Childhood-Onset_Neurodegeneration
+  path: kb/disorders/LETM1-Related_Childhood-Onset_Neurodegeneration.yaml
+session:
+  id: 2026-09-02T221123Z-claude-code-a68702
+  timestamp: '2026-09-02T22:11:23Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-fable-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10445
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - animal_models
+  - evidence
+  - notes
+  summary: 'Address round-5 review on PR #10445'
+  details: |
+    Round-5 review fixes. Blocking items: (1) added Muscular Wasting (HP:0003202 Skeletal muscle atrophy, FREQUENT) for the Table 2 row at 7/10 (70%), curated distinctly from the existing Myopathy entry at 50% because the source tabulates them as separate rows with different denominators; (2) added Impaired Speech Acquisition (HP:0000750 Delayed speech and language development, FREQUENT) for 6/9 (67%); (3) split the PMID:17606466 evidence item on Mitochondrial Matrix Swelling and Cristae Disorganization, whose single snippet spanned human cells and C. elegans under one IN_VITRO grading, into an IN_VITRO item quoting the mammalian-cell overexpression result and a MODEL_ORGANISM item quoting the C. elegans clause. Both reviewer-supplied quotes were re-verified as exact substrings of references_cache/PMID_36055214.md before use. Non-blocking items taken: added High Palate (HP:0000218) and Teeth Abnormalities (HP:0000164) so all four components of the quoted dysmorphism sentence are curated; promoted Loss of Ambulation (HP:0002505) out of the Developmental Regression description into its own phenotype; gave Hypotonia its stated 11/18 (61%) frequency and an evidence item that actually states it; wired Loss of Ambulation, Pontine Hypoplasia, Spastic-Ataxic Gait and Impaired Speech Acquisition from Progressive Neurodegeneration and Multisystem Involvement, and Muscular Wasting from Cellular Energy Failure in High-Demand Tissues, so the previously orphaned neurologic phenotypes match their already-wired siblings; added the zebrafish letm1-/- knockout (Dao 2022, DOI:10.26508/lsa.202101194, newly fetched into references_cache) as an animal model with PARTIALLY_RECAPITULATES / LOW fidelity against the energy-failure node, with limitations recording that the readout is NAD/circadian rather than respiratory chain, that mutants split into compensating and non-compensating developers, and that the null is viable in fish but embryonic lethal in mouse. Deferred and recorded in notes: differential diagnosis (OPA1, ATAD3A, TARS2, SERAC1, DNAJC19, HTRA2, Barth/TAZ) as a follow-up rather than first-entry scope, and no C. elegans animal_models entry because the abstract-only cache gives no genotype or readout detail. Developmental-regression frequency remains omitted. Validation: just validate, validate-terms, count-verified-snippets, check-duplicate-keys, check-entity-refs, check-causal-targets, check-enum-values and check-reference-titles all pass.
--- a/kb/disorders/LETM1-Related_Childhood-Onset_Neurodegeneration.yaml
+++ b/kb/disorders/LETM1-Related_Childhood-Onset_Neurodegeneration.yaml
@@ -86,15 +86,33 @@ notes: >-
   the term, whereas the cited 78% sentence establishes the ataxia and not its
   tempo, and the separable qualifier keeps those two claims distinct. (3)
   Frequency is omitted on failure to thrive, feeding difficulties, respiratory
-  distress, developmental regression, ragged-red fibres, micrognathia, low-set
-  ears, pontine hypoplasia and ventriculomegaly, in each case
+  distress, developmental regression, loss of ambulation, ragged-red fibres,
+  micrognathia, low-set ears, high palate, teeth abnormalities, pontine
+  hypoplasia and ventriculomegaly, in each case
   because the source gives no usable denominator - the reason is recorded on
   each phenotype rather than inferred from the absence. The rule applied
   throughout is that a `frequency` band is set only where the source states the
   fraction itself: facial dysmorphism (4/10), microcephaly (2/6) and pericardial
   effusion (3/11) carry one; the individual components of the dysmorphic pattern
   and the two additional neuroimaging findings, which the source reports as
-  narrative counts inside the six-person MRI subset, do not.
+  narrative counts inside the six-person MRI subset, do not. (4) No
+  differential-diagnosis content. The deep-research report for this entry covers
+  the mitochondrial and inner-membrane differentials - OPA1, ATAD3A, TARS2,
+  SERAC1, DNAJC19, HTRA2 and Barth syndrome/TAZ - and that material is real and
+  usable, but it is deliberately deferred rather than dropped: writing it well
+  means curating each of those entities' discriminating features against a
+  single 18-person cohort, which is a follow-up of its own rather than part of
+  a first entry. Recorded here so a later review does not re-raise it as an
+  oversight.
+
+  (5) No C. elegans entry in `animal_models`, although PMID:17606466's worm
+  result is cited on the swelling node as `MODEL_ORGANISM` evidence. The
+  abstract-only cache reports the mutant's phenotype in a single clause with no
+  genotype, allele or readout detail, which is not enough to fill an
+  `AnimalModel` record honestly. The zebrafish letm1 knockout, by contrast, is
+  curated: it is the only viable germline vertebrate null available and its
+  source is cited by DOI rather than PMID because the Life Science Alliance
+  article was cached from its open-access full text.
 
 classifications:
   harrisons_chapter:
@@ -525,11 +543,23 @@ pathophysiology:
     reference_title: "Inverse correlation between expression of the Wolfs Hirschhorn candidate gene Letm1 and mitochondrial volume in C. elegans and in mammalian cells."
     supports: SUPPORT
     evidence_source: IN_VITRO
-    snippet: "Reduced levels of Letm1 in human cells and in C. elegans lead to swellings along the lengths of mitochondria, consistent with the phenotype observed in yeast. Electron micrographs show mitochondria with swollen matrices that are less electron-dense than matrices in normal mitochondria."
+    snippet: "The opposite effect is achieved by overexpression of Letm1. Overexpression increases the electron density of the mitochondrial matrix and swelling of cristae."
     explanation: >-
-      Independent replication of the swelling phenotype across human cells,
-      C. elegans and yeast, establishing it as the conserved consequence of
-      LETM1 loss.
+      The reciprocal dosage experiment, performed in the cultured mammalian
+      cells that are this study's cell-based arm - raising Letm1 condenses the
+      matrix instead of swelling it. Establishes that Letm1 sets matrix volume
+      in both directions rather than being permissive for it.
+  - reference: PMID:17606466
+    reference_title: "Inverse correlation between expression of the Wolfs Hirschhorn candidate gene Letm1 and mitochondrial volume in C. elegans and in mammalian cells."
+    supports: SUPPORT
+    evidence_source: MODEL_ORGANISM
+    snippet: "in C. elegans lead to swellings along the lengths of mitochondria, consistent with the phenotype observed in yeast."
+    explanation: >-
+      The whole-organism arm of the same study - a C. elegans letm-1 mutant
+      reproduces the swelling seen in cell culture, so the lesion is conserved
+      rather than an artefact of cultured cells. Quoted as a partial sentence
+      because the source states the human-cell and C. elegans results in one
+      sentence, and this item is graded for the animal result alone.
   - reference: PMID:32139798
     reference_title: "The mitochondrial inner membrane protein LETM1 modulates cristae organization through its LETM domain."
     supports: SUPPORT
@@ -664,6 +694,12 @@ pathophysiology:
     causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
   - target: Myopathy
     causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
+  - target: Muscular Wasting
+    causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
+    description: >-
+      Wasting follows the same energy-failure route as the myopathy, with the
+      neurogenic component seen on nerve conduction studies as an additional
+      intermediate.
   - target: Diabetes Mellitus
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   evidence:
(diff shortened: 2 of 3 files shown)


AO33:
All three blocking items are fixed in 5f9bafc. Every evidence item now carries a `reference_title`, nine missing phenotypes were added, and the wrong-scope GO binding is gone. Verified snippet count went from 58/58 to 69/69.

### Missing `reference_title`

Added to all 69 evidence items. Each title was read out of the `title:` key in the matching `references_cache/PMID_<id>.md` frontmatter rather than typed from memory — 15 distinct PMIDs. The two entries in the top-level `references:` block already had `title:` and were left alone. `just check-reference-titles` reports no new mismatches.

### Missing phenotypes

Nine added, each with an exact-substring quote from `references_cache/PMID_36055214.md`:

| Phenotype | HP term | Frequency |
|---|---|---|
| Nystagmus | HP:0000639 | 7/13, 54% |
| Hyperkinetic Movement Disorders | HP:0002487 | 4/12, 33% |
| Spastic-Ataxic Gait | HP:0002497 | 3/9, 33% |
| Hyperreflexia (brisk deep tendon reflexes) | HP:0001347 | 4/10, 40% |
| Extensor Plantar Response (upgoing plantar) | HP:0003487 | 4/9, 44% |
| Peripheral Neuropathy | HP:0009830 | 3/9, 33% |
| Failure to Thrive | HP:0001508 | omitted |
| Feeding Difficulties | HP:0011968 | omitted |
| Respiratory Distress | HP:0002098 | omitted |

Peripheral neuropathy was not in your list but sits in the same sentence at the same 33%, so leaving it out would have been arbitrary. The last three come from the presenting-symptom sentence, which names them as common but gives no denominator; frequency is omitted and the reason is written on each phenotype rather than left implicit.

**One CURIE differs from your suggestion.** For hyperkinetic movement disorders you proposed `HP:0100022` Abnormality of movement. I used `HP:0002487` Hyperkinetic movements instead — it is the direct `is_a` child of `HP:0100022`, carries "Hyperkinesia" and "Hyperkinesis" as exact synonyms, and names the finding rather than its parent category. `HP:0000639` for nystagmus was used as suggested. There is no HP term for "spastic-ataxic gait", so that one is bound to `HP:0002497` Spastic ataxia with the gait detail in `preferred_term`.

### GO:0006884

Binding removed. Confirmed your reading against the ontology: it is defined as "the three-dimensional space occupied by a cell", is `is_a GO:0008361` regulation of cell size, and carries `RO:0019000 some OBA:0000056` cell volume. I searched GO for "mitochondrial volume", "mitochondrion volume", "organelle volume" and "mitochondrial swelling" — all return nothing, so there is no organelle-scope term to swap in. `preferred_term` is now "mitochondrial matrix volume homeostasis", unbound, with the NTR rationale in the descriptor's `description`. `GO:0007005` and `GO:0042407` on the same node are unchanged.

### Suggestions taken

- Progression tempo now quotes the figures: rapid 9/18 (50%), moderately fast 4/18 (22%), slow 5/18 (28%).
- Added a Mortality progression phase — 9/18 died between 2 months and 8 years, survivors at a median 15 years. Noted that this is one ascertained 18-person cohort and not a population survival estimate.
- Added `CL:0000169` type B pancreatic cell to the Cellular Energy Failure node, which its description already named.

### Suggestions not taken, and why

**The ataxia phenotype stays on `HP:0001251`.** `HP:0002073` Progressive cerebellar ataxia pre-composes the tempo into the term, but the sentence being cited — "cerebellar ataxia (78%)" — establishes the ataxia and not that it progresses. `HP:0001251` carries "Cerebellar ataxia" as an exact synonym, so nothing is lost on the cerebellar side, and `clinical_course: PROGRESSIVE` alongside keeps the two claims separable. If you would rather have the pre-composed term I will change it, but I think the split is the more honest fit to the quote.

**Still no `diagnosis:` section.** There is no LETM1-specific assay, criteria set or published testing algorithm; diagnosis is sequencing that returns bi-allelic LETM1 variants in a child with a respiratory chain deficiency, and the defining cohort was assembled genotype-first by matchmaking rather than through a clinical case definition. A section here would restate the generic mitochondrial workup already carried by `biochemical` and the `diagnostic: true` phenotypes. This is a real gap to fill when a pathway is published, and I have written that into the entry's `notes` so it reads as a decision rather than an oversight — along with the ataxia reasoning and the omitted-frequency reasoning above.

### Validation

```
just validate                    ✓ All validations passed
just validate-terms              ✅ Validation passed
just count-verified-snippets     69/69 verified against cached references
just check-duplicate-keys        OK
just check-entity-refs           OK
just check-causal-targets        OK — no new broken targets
just check-enum-values           OK
just check-title-snippets        OK — no new title-quoting snippets
just check-snippet-length        OK — no new snippets under 5 words
just check-snippet-grading       OK — no new grading divergences
just check-reference-titles      OK — no new title mismatches
just validate-disorders          ✓ passed, 69/69 snippets
```

No `cache/**/*.csv` churn: every HP and CL term bound here was already in the term cache. A history record for this revision is included.

AO33:
Added the missing dysmorphology. Seven new phenotypes, all from PMID:36055214: facial dysmorphism (HP:0001999, 4/10 = 40%), microcephaly (HP:0000252, occipitofrontal circumference below the third percentile in 2/6 = 33%), micrognathia (HP:0000347), low-set ears (HP:0000369), and the three non-blocking suggestions the source does support — pontine hypoplasia (HP:0012110), ventriculomegaly (HP:0002119) and pericardial effusion (HP:0001698, 3/11 = 27%).

**One thing to flag: the sentence quoted in the review is not an exact substring of the cache.** It contains two U+00A0 non-breaking spaces — `ascribed to the LETM1 deletion` and `2/6 persons (33%)` — so pasting it as a snippet would have failed reference validation. I re-derived every snippet by direct substring test against `references_cache/PMID_36055214.md` instead of copying from the review. The spans used avoid the non-breaking spaces; the claim is unchanged.

**Frequency.** I applied one rule uniformly and recorded it in `notes`: a `frequency` band is set only where the source states the fraction itself. Facial dysmorphism (4/10), microcephaly (2/6) and pericardial effusion (3/11) get one. The individual components of the dysmorphic pattern have no per-feature denominator, and pontine hypoplasia and ventriculomegaly are narrative counts inside the six-person MRI subset ("a proband", "2 persons each"), so those five carry none with the reason on each phenotype. Developmental regression stays without a frequency, as you endorsed.

**The Wolf-Hirschhorn framing.** You were right that leaving this out weakened the argument the `notes` were making. The notes now state the overlap explicitly — the source describes these bi-allelic individuals as showing a milder spectrum of WHS signs not previously ascribed to the LETM1 deletion — and then give three reasons it does not merge the two entries:

1. The source attributes the growth deficiency, microcephaly and characteristic facies of WHS to haploinsufficiency of WHSC1, "a region located far from LETM1", and offers a putative LETM1-WHSC1 interaction as the explanation for the milder overlap here. It treats the shared features as convergent, not as one disease reached twice.
2. The overlap is partial and attenuated — 4/10 and 2/6, a mild inconstant pattern against the specific near-constant Greek-warrior-helmet facies of WHS.
3. The features that dominate this pathograph (lactic acidosis, diabetes, cataract, neuropathy, proximal myopathy, cerebellar ataxia, pontine/cerebellar atrophy) are named by the same source as "not typical of WHS" and typical instead of archetypal mitochondrial disorders.

**One inconsistency in the source, recorded rather than resolved.** The text and figure legend call the family-1 pontine lesion hypoplasia; the Table 2 abbreviation key expands the same code as "pontine atrophy". I bound HP:0012110 (Hypoplasia of the pons) to follow the narrative and noted the discrepancy in the phenotype description.

Validation: `just validate`, `just validate-terms` and `just validate-disorders` all pass. Snippets now 79/79 verified, up from 69/69. `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-enum-values`, `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` and `check-reference-titles` are all clean with no new findings. No cache churn. History record added.

AO33:
All three blocking items are fixed, plus four of the five non-blocking suggestions. The differential-diagnosis section is the one I deliberately left out, and I wrote the reason into `notes` so it does not come back as an open question next round.

**Both quotes you supplied are exact substrings.** I re-checked them by direct byte comparison against `references_cache/PMID_36055214.md` rather than trusting the transcription, since an earlier round of this PR was tripped by a non-breaking space. Neither of these two spans contains one. (The file does contain U+00A0 elsewhere — for instance in "2/6 persons (33%)" — so the hazard is real, just not in these sentences.)

### The three blocking items

**Muscular wasting.** Added as its own phenotype, `HP:0003202` Skeletal muscle atrophy, `frequency: FREQUENT`, quoting the 7/10 (70%) row. I kept it separate from `Myopathy` rather than merging, because the source tabulates the two as different rows with different denominators, and the wasting tracks the neurogenic changes on nerve conduction studies as much as the myopathic ones — it is not just the visible face of the myopathy.

**Impaired speech acquisition.** Added, `HP:0000750` Delayed speech and language development, `frequency: FREQUENT`, quoting 6/9 (67%). The description says why it is not folded into the global developmental delay: the 78% sensorineural hearing loss gives it a specific mechanism, and one proband is separately recorded as spastic-dysarthric, so there is a motor-speech contribution as well.

**Mixed evidence source on PMID:17606466.** Split into two items, which was the option you preferred. One is `IN_VITRO` quoting the mammalian-cell overexpression result ("The opposite effect is achieved by overexpression of Letm1…"), the other is `MODEL_ORGANISM` quoting the C. elegans clause.

One thing worth flagging about that split: the abstract states the human-cell and C. elegans swelling results in a single sentence, with the human clause first. There is no whole sentence that isolates the worm result, so the `MODEL_ORGANISM` snippet is a partial sentence beginning "in C. elegans lead to swellings…". It is an exact substring and it states a finding, but it is a fragment, and I said so in that item's `explanation` rather than leaving a reader to wonder. If you would rather have a whole sentence there, the alternative is to drop the worm evidence entirely and keep only the cell-based grading — I chose the split because the cross-species replication is worth keeping.

### Non-blocking items taken

- **High palate** (`HP:0000218`) and **teeth abnormalities** (`HP:0000164`) added, so all four components of that dysmorphism sentence are now curated rather than two of four. Teeth abnormalities also carries the photographic sentence naming F1:S1 and F1:S2 directly. `HP:0000164` is "Abnormality of the dentition" — the source says only "teeth abnormalities" with no indication of what kind, so the general term is the correct level.
- **Loss of ambulation** (`HP:0002505`) promoted out of the `Developmental Regression` description into its own phenotype, with the mean age 5.4 years and range 2–12. Frequency omitted, with the reason recorded: the source gives a mean and range for those who lost ambulation, not a cohort fraction.
- **Orphan phenotypes wired.** `Spastic-Ataxic Gait`, `Pontine Hypoplasia`, `Loss of Ambulation` and `Impaired Speech Acquisition` now hang off `Progressive Neurodegeneration and Multisystem Involvement`, and `Muscular Wasting` off `Cellular Energy Failure in High-Demand Tissues` alongside the myopathy. Each edge carries a `causal_link_type` and, where the route is indirect, a description of what the intermediate is. All targets are bare names.
- **Zebrafish model added.** I fetched the Dao 2022 paper (`DOI:10.26508/lsa.202101194`, Life Science Alliance, open access) so the model has real evidence behind it. It is `PARTIALLY_RECAPITULATES` against the energy-failure node at `fidelity: LOW`, and the `limitations` field gives three separate reasons for that grade: the readout is NAD/redox and circadian rather than the respiratory chain deficiency that defines the human disease, the mutants split into compensating and non-compensating developers so the phenotype is not uniform within the model, and the null is viable in fish while embryonic lethal in mouse.

### One thing to know about the zebrafish evidence

`DOI` is in `skip_prefixes` in `conf/reference_validator_config.yaml`, so the validator reports "87/89 verified, 2 skipped by prefix" — the two skipped items are the zebrafish ones. The tool did not check them. I did, by direct substring match against the cache file, and that file is committed in this PR so the check is repeatable. This is the only non-PMID reference in the entry.

### Deferred, with reasons in `notes`

- **Differential diagnosis** (OPA1, ATAD3A, TARS2, SERAC1, DNAJC19, HTRA2, Barth/TAZ). Doing it properly means curating each of those entities' discriminating features against a single 18-person cohort, which is its own piece of work. Recorded as deliberate rather than missed.
- **No C. elegans entry in `animal_models`.** The worm result is cited as evidence on the swelling node, but PMID:17606466 caches as abstract only and gives no genotype, allele or readout detail — not enough to fill an `AnimalModel` record honestly.

Developmental-regression frequency stays omitted, as you endorsed.

### Validation

`just validate`, `validate-terms`, `count-verified-snippets`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-enum-values`, `check-reference-titles` and the batched `validate-disorders` all pass on this file. Snippet count is now 87/89 verified with 2 skipped by prefix, up from 79.

`check-title-snippets`, `check-snippet-length` and `check-snippet-grading` each report findings, but none are in this file — they are in `Epidermolysis_Bullosa_Simplex`, `Hennekam_Lymphangiectasia-Lymphedema_Syndrome_2`, `Primary_Tonsillar_Lymphoma`, `Appendiceal_Neoplasm`, `Bullous_Pemphigoid` and `Follicular_Lymphoma`. Those three gates diff the whole KB against `origin/main`, and this branch is behind `main` (whose current head is "Refresh the title-snippet baseline"), so the gates are reporting content that `main` has already rebaselined. Nothing here introduces them.

PR #10447, opened 2026-09-01
curate(Ichthyotic_Keratoderma_Spasticity_Hypomyelination_And_Dysmorphic_Facial_Features): add ELOVL1 entry (MONDO:0032798)

### LINKED ISSUE #10412: Curate ichthyotic keratoderma, spasticity, hypomyelination, and dysmorphic facial features (MONDO:0032798)

Claiming **MONDO:0032798** — ichthyotic keratoderma, spasticity, hypomyelination, and dysmorphic facial features (IKSHD), *ELOVL1*.

Stub: `stubs/Ichthyotic_Keratoderma_Spasticity_Hypomyelination_And_Dysmorphic_Facial_Features.yaml`

**Lump/split check.** MONDO parent is `hereditary disease`; no descendants; single causal gene ELOVL1. No existing KB entry covers ELOVL1 or very-long-chain fatty acid elongation in this context. Curating as `entry_type: DISEASE`.

nlharris:

Curates the **ELOVL1-related neuro-ichthyotic disorder** (`MONDO:0032798`) as a new `kb/disorders/` entry. Closes #10412.

## One enzyme, two organs

ELOVL1 catalyses the rate-limiting condensation step that elongates C22–C26 saturated and monounsaturated fatty acids. Two tissues specifically need the *longest* species: the ceramide lamellae of the stratum corneum, and the sphingolipids of central myelin. Cap the chain length and you get an ichthyosis and a hypomyelinating leukodystrophy in the same patient — which is why an ichthyosis gene is also a leukodystrophy gene.

```
ELOVL1 elongase activity loss
  → deficient C24–C26 VLCFA elongation
      ├→ shortened sphingolipid acyl chains in central myelin
      │     → CNS hypomyelination → spastic paraplegia, nystagmus
      └→ reduced ultra-long-chain ceramides in stratum corneum
            → epidermal permeability barrier failure → ichthyosis
```

Patient fibroblast transcriptomics shows both arms simultaneously, as reciprocal programmes — keratinisation up, myelination down.

## Scope: deliberately the opposite call from the RBP4 entry in this batch

Both inheritance arms are curated together here. That is the reverse of what I did for `MONDO:0014060` (RBP4) a few hours ago, and the reason is a fact about MONDO rather than about the biology: I checked, and **MONDO carries exactly one ELOVL1 disease concept**. For RBP4 the dominant and recessive arms each had somewhere to live, so keeping them apart cost nothing. Here, scoping to the monoallelic arm alone — which the disease name would justify, since "IKSHD" is the founding report's own acronym — would leave the seven published biallelic patients with no entry at all.

They are separated as `has_subtypes` so the split can be made cleanly if a term appears, and a discussion records that the lumping is provisional rather than settled.

## Two mouse strains, opposite verdicts on the same node

This is the part most worth a reviewer's attention. The `Elovl1` **null** is definitive for the epidermal arm and `FAILS_TO_RECAPITULATE` the neurological one — **not because myelination is normal, but because the animals die within hours of birth from barrier failure and never reach myelination**. The hypomorphic mutant survives and carries the myelin phenotype.

Curating the null as simply "a model of ELOVL1 deficiency" would conceal that the most complete genetic model of this gene cannot speak to half the disease.

## Removed rather than shipped

- A drafted **acanthosis nigricans** phenotype. The feature appears only in the title of `PMID:30487246` and nowhere in the retrieved abstract — a title is not a finding.
- **Seven uncited reference caches** (elongase reviews, congenital ichthyosis treatment trials, X-linked adrenoleukodystrophy lipidomics). All concern other diseases; they were read and set aside, and the notes say so.

## Report handling

The deep-research report's one machine-checked quote **failed verification and turned out to be a paraphrase, not a fabrication** — it compressed and reordered a real sentence from `PMID:40590574`. The full original is better and is what the entry quotes. Six further term flags were category headings written where the term's own name belonged; the identifiers were correct.

The Sjögren–Larsson differential quote is trimmed to a span free of the non-breaking spaces the source uses mid-sentence, rather than reproducing fragile bytes — the reason is recorded in the `explanation`.

One slip of mine, identical to one on the RBP4 entry: `GO:0030497 fatty acid elongation` was first filed under `molecular_functions` and is a biological process. Caught by the dynamic-enum validator both times.

## Validation

```
just validate                 schema clean · terms clean · 37/37 snippets verified
just check-entity-refs        OK      just check-duplicate-keys   OK
just check-causal-targets     OK      just check-qualifier-terms  OK
just check-snippet-grading    OK      just check-title-snippets   OK
just check-snippet-length     OK      just check-folded-hyphens   OK
just check-enum-values        OK      just validate-history       No issues found
```

Cache diff additive with zero deletions. Stub deleted, history record added. Every cited PMID has a shipped cache and no uncited cache ships, reconciled by re-deriving the cited list from the entry file rather than from memory.

---

### DIFF
--- a/history/disorders/Ichthyotic_Keratoderma_Spasticity_Hypomyelination_And_Dysmorphic_Facial_Features/2026-09-01T232626Z-claude-code-61b516.yaml
+++ b/history/disorders/Ichthyotic_Keratoderma_Spasticity_Hypomyelination_And_Dysmorphic_Facial_Features/2026-09-01T232626Z-claude-code-61b516.yaml
@@ -1,0 +1,30 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Ichthyotic_Keratoderma_Spasticity_Hypomyelination_And_Dysmorphic_Facial_Features
+  path: kb/disorders/Ichthyotic_Keratoderma_Spasticity_Hypomyelination_And_Dysmorphic_Facial_Features.yaml
+session:
+  id: 2026-09-01T232626Z-claude-code-61b516
+  timestamp: '2026-09-01T23:26:26Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10447
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - prevalence
+  - differential_diagnoses
+  - evidence
+  summary: 'Review round 2: phenotype gaps, node scoping, VLCFA qualification, differentials'
+  details: |
+    Answered the CHANGES_REQUESTED review on PR #10447 in a single push. Added Dysarthria (HP:0001260), the third member of the movement-disorder triad and the only feature documented in both inheritance arms, and Dysmorphic facial features (HP:0001999), a quarter of the eponymous tetrad; both are supported by sentences in caches the entry already shipped, so the omission was a curation gap rather than a sourcing one. Narrowed the root pathophysiology node's genetic_context by dropping variant_origin DE_NOVO and zygosity HETEROZYGOUS, which described only the monoallelic arm even though that node is the shared root through which every downstream edge of both arms flows; functional_impact_category LOSS_OF_FUNCTION is true of both and was kept. Surfaced the qualifying serum result from PMID:29496980 on the elongation node as a REFUTE item against the general reduced-plasma-VLCFA reading, alongside the existing SUPPORT item and a new IN_VITRO fibroblast item, plus a node note recording that the compartment in which the defect is measurable differs by arm. Added prevalence with prevalence_class ULTRA_RARE, which the earlier decision to omit prevalence entirely had been too strong about. Added two differential diagnoses: ELOVL4-related disorder, from a sentence in an already-shipped cache, and X-linked adrenoleukodystrophy, which required one new reference. The X-ALD case is the opposite pole of the same biochemical axis, so PMID:20166112 was fetched to establish that ELOVL1 is the single elongase synthesising C26:0. A first PMID guessed for that paper resolved to an unrelated article titled Consanguinity and was discarded rather than cited; the correct identifier came from a PubMed title search. Declined two suggestions with reasons recorded in the entry notes: palmoplantar keratoderma, because no shipped cache mentions it and the suggested HP:0007479 is Congenital nonbullous ichthyosiform erythroderma rather than a keratoderma term, and dietary VLCFA intake as an environmental modifier, because the only source hedges it as something that may influence expression in a future-work sentence. Validation: schema, terms and references clean with 47/47 snippets verified, up from 37/37; all offline gates green.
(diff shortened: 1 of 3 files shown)


nlharris:
All four blocking items are fixed in `07754fdf9`, plus three of the five suggestions. One push, so this only costs one re-review cycle.

## The four blocking items

**Dysarthria and dysmorphic facial features — you were right, and the second one is the more embarrassing miss.** Both were supported by sentences in caches this PR already ships, so this was a curation gap rather than a sourcing one. Dysarthria is now `HP:0001260` and deliberately *not* scoped to a subtype, because it is the one movement-disorder feature documented in both arms — 6/6 biallelic from PMID:40590574, and mild dysarthria in the monoallelic patients from PMID:42101073. Dysmorphic facial features is `HP:0001999`, scoped to `Monoallelic IKSHD` and quoted from the objectives sentence you pointed at. The `description` says explicitly that the published account is a category label rather than a list of features, so the entry asserts presence and nothing about which features or how consistently.

**Root node scoping.** Dropped `variant_origin: DE_NOVO` and `zygosity: HETEROZYGOUS`, kept `functional_impact_category: LOSS_OF_FUNCTION`. Your framing is the right one: the entry's whole scoping argument is that both arms belong together, and the root node was quietly asserting only one of them.

**The qualifying serum result.** This is the finding I most should have caught, because it changes how the biochemistry reads. It is now two evidence items on the elongation node rather than one, following the split-into-two-items pattern:

- `SUPPORT` — reduced plasma C24:0 and C26:0, one biallelic patient (PMID:40590574)
- `REFUTE` — normal serum C20:0–C26:0 with only a decreased C24:0/C22:0 ratio, monoallelic patients (PMID:29496980)

Plus a third `IN_VITRO` item for the fibroblast measurement, which does show the chain-length shift in the monoallelic patients whose serum is normal, and a node `notes` line saying which compartment reports the defect in which arm. The practical upshot — a normal serum VLCFA panel does not exclude this diagnosis — is now stated rather than implied.

## Suggestions taken

**X-ALD as a differential.** You were right that my note answered a different question. Declining the *X-ALD lipidomics papers* as being about another disease is fine; a `differential_diagnoses` entry is about *this* disease. Adding it needed one new reference, since no shipped cache mentioned X-ALD at all — PMID:20166112, which establishes ELOVL1 as the single elongase synthesising C26:0. That is what makes the two disorders opposite poles of one axis: lose the elongase and chain length is capped, block peroxisomal beta-oxidation and the same elongase has more substrate. The therapeutic logic mirrors too, since ELOVL1 knockdown is a proposed X-ALD treatment.

Worth flagging: my first attempt at that PMID resolved to an unrelated article titled "Consanguinity". I discarded the cache and found the real identifier by PubMed title search rather than citing from memory.

**ELOVL4 differential** — added from the already-cached sentence, no new fetch.

**`prevalence_class: ULTRA_RARE`** — added. Declining prevalence entirely was stronger than the sources require, and you are right that the qualitative tiers exist for exactly this case.

## Suggestions declined

**Palmoplantar keratoderma.** Two reasons: no cache this PR ships mentions it, and `HP:0007479` is *Congenital nonbullous ichthyosiform erythroderma*, not a keratoderma term — I checked before declining rather than after. If there is a source for it, I would want the PMID.

**Dietary VLCFA intake as an environmental modifier.** The only source hedges it as something that "may also influence disease expression", inside a future-work sentence. An `environmental[]` entry needs its own evidence for the exposure-to-mechanism link, and a speculation about a possible modifier is not that.

Both declines are recorded in the entry `notes` with the reasoning, so the next curator does not have to re-derive it.

## Validation

`just validate` clean — schema, terms, references, **47/47 snippets verified**, up from 37/37. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `validate-history` all OK.

Cache diff is one new file, `PMID_20166112.md`, and it is cited. Reconciled by re-deriving the cited-PMID list from the entry file.

On the GeneReviews point — you were right to flag that as judgement rather than a verified query. I have not run it either, so it stands unverified from both sides.

---

PR #10448, opened 2026-09-01
curate: Add Denys-Drash Syndrome (MONDO:0008682, WT1) with WT1_Disorders grouping

### LINKED ISSUE #10391: Curate Denys-Drash syndrome (MONDO:0008682)
Curate a dismech entry for **Denys-Drash syndrome** ([MONDO:0008682](https://monarchinitiative.org/MONDO:0008682)) — constitutional *WT1* variants, classically missense changes in the zinc-finger DNA-binding domain, producing the triad of early-onset diffuse mesangial sclerosis with steroid-resistant nephrotic syndrome, 46,XY disorder of sex development, and a high risk of Wilms tumor.

Working assumption is `DISEASE`. The stub records no MONDO descendants, and its parents are `autosomal dominant disease`, `syndromic disease`, and `46,XY disorder of sex development` — a leaf, not a grouping. The pathograph is conserved: one gene, one dominant-negative mechanism, three organ consequences.

Related entries already in the KB, to keep distinct:

- `kb/disorders/Wilms_Tumor.yaml` (MONDO:0006058) — the tumor. Denys-Drash is the germline predisposition syndrome, and per design decisions §3a those stay separate.
- `kb/disorders/Meacham_syndrome.yaml` already cites MONDO:0008682 in `differential_diagnoses`, so the term is in use as a differential but not as a primary entry.
- Frasier syndrome (*WT1* intron 9 splice-donor variants) is a distinct entity, not a subtype of this one; worth naming in `differential_diagnoses` rather than folding in.

Stub: `stubs/Denys-Drash_Syndrome.yaml`

bpow:
Adds a Denys-Drash syndrome entry, plus a new `WT1_Disorders` grouping that collects it with WAGR and Meacham syndrome.

The thing worth a reviewer's attention is that GeneReviews no longer treats "Denys-Drash" as a useful clinical diagnosis. Its WT1 Disorder chapter says the Denys-Drash and Frasier labels "are now recognized to be part of a phenotypic continuum and are no longer clinically helpful." MONDO still carries the two terms separately and has no umbrella term for the continuum, so the knowledge base cannot simply follow the clinical reframing.

Closes #10391

## What I did about that

I kept the entry on MONDO:0008682 and put the continuum in three places rather than picking a side quietly:

- The entry description says outright that this is one pole of a continuum, not a separate disease.
- Frasier syndrome sits in `differential_diagnoses` with the mechanistic contrast spelled out.
- An OPEN `KNOWLEDGE_GAP` discussion records the question, quoting the GeneReviews sentence above.

Two reasons for keeping the entry rather than merging it away: the historical literature is indexed under this term, and the mechanisms genuinely differ. Denys-Drash is a zinc-finger missense variant acting as a dominant negative; Frasier is an intron 9 splice variant that shifts the +KTS isoform ratio and makes no mutant protein at all. Collapsing them would lose that.

## The grouping

`kb/groupings/WT1_Disorders.yaml` is how the continuum gets represented as structure instead of only prose. It ships in this PR because it lists this entry as a member and cannot merge before the entry exists.

| Member | WT1 lesion class |
|---|---|
| Denys-Drash syndrome | zinc-finger missense, dominant negative |
| WAGR syndrome | 11p13 contiguous deletion, haploinsufficiency (+ PAX6) |
| Meacham syndrome | WT1-associated malformation combination |

One `NECESSARY` criterion, `HAS_GENE WT1`; all three members evaluate SATISFIED.

Two boundary calls are written into `grouping_rationale` rather than left implicit. **Frasier syndrome belongs to this concept and is missing only because it has no entry yet** — that is the most valuable follow-up here, and it is logged in the grouping's `notes`. **Wilms Tumor is deliberately excluded**: it is the tumor these syndromes predispose to, and design decisions §3a keeps germline predisposition syndromes separate from the tumors they predispose to.

## The pathograph

```
WT1 zinc-finger missense ──► dominant-negative interference ──┬─► loss of WT1 program in podocytes ──► diffuse mesangial sclerosis ──┬─► steroid-resistant nephrotic syndrome
        │                                                     │                                                                     ├─► proteinuria
        │                                                     │                                                                     ├─► stage 5 CKD
        │                                                     │                                                                     └─► hypertension
        │                                                     └─► failure of gonadal somatic cell differentiation ──┬─► gonadal dysgenesis
        │                                                                                                           ├─► ambiguous genitalia
        │                                                                                                           └─► gonadoblastoma
        └─► biallelic WT1 inactivation in nephrogenic rests ──► nephroblastoma
```

Single initiating lesion, and every phenotype has an incoming edge.

## Correction to an earlier version of this description

**An earlier version of this text said that `HP:0001967` "is used instead" of the wrong CURIE. That was not true of the diff, and the review was right to catch it.**

What actually happened: the falcon report's term validation bound `HP:0005567` (*Renal magnesium wasting*) to a table row describing diffuse mesangial sclerosis. That wrong CURIE was correctly rejected and never entered the entry, and `HP:0001967` was looked up as the right term — but the lesion was then modeled only as a pathophysiology node with a cell-type binding, so the HP term never reached the file at all. Diffuse mesangial sclerosis now exists as a bound phenotype using `HP:0001967`, alongside the pathophysiology node of the same name.

The report was also thin on citations — its sidecar held only the query, and the body yielded three resolvable references. Most evidence here came from independent PubMed searching, seeded by a PMID list the report surfaced from Open Targets.

## Validation

Schema, terms and references all pass. 41/41 snippets verified, and all 41 evidence items carry `reference_title`. Entity-ref, causal-target, duplicate-key, enum-value, folded-hyphen, snippet-length, title-snippet and snippet-grading gates clean. Grouping passes `validate-grouping` and `check-groupings`. History records validate. Named Entity Confusion preflight PASS (WT1 mentioned 88 times). The `stubs/Denys-Drash_Syndrome.yaml` stub is deleted.

### DIFF
--- a/history/disorders/Denys-Drash_Syndrome/2026-09-01T235109Z-claude-code-c560ee.yaml
+++ b/history/disorders/Denys-Drash_Syndrome/2026-09-01T235109Z-claude-code-c560ee.yaml
@@ -1,0 +1,39 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Denys-Drash_Syndrome
+  path: kb/disorders/Denys-Drash_Syndrome.yaml
+session:
+  id: 2026-09-01T235109Z-claude-code-c560ee
+  timestamp: '2026-09-01T23:51:09Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+    agent_version: 2.1.252
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10391
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10448
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - inheritance
+  - diagnosis
+  - treatments
+  - evidence
+  summary: 'Take review suggestions on PR #10448 (Denys-Drash)'
+  details: |
+    Second response round on PR #10448, after the reviewer approved the six blocking fixes and left three suggestions. All three taken, and the first corrects an error in the previous round's reasoning rather than adding an improvement.
+
+    The inheritance description claims transmission from an affected or mosaic parent is documented. The previous round deferred citing it, recording in both a history record and a PR comment that PMID:1327525 could not supply a snippet because its cache is title-only. That statement about PMID:1327525 is true but was the wrong thing to check. The GeneReviews record already cited in this entry carries both halves of the claim outright: a parent heterozygous for the proband variant in rare instances, and sibling recurrence risk above background because of possible parental gonadal mosaicism. The clause was citable from a source already in hand, and the deferral happened because the search anchored on the reference the reviewer had named in passing instead of re-reading the cache the entry already uses. Both sentences are now evidence items.
+
+    Also taken: diagnosis_term bound to NCIT:C15709 (Genetic Testing), matching the POT1 entry in this series and 416 others; and Kidney Replacement Therapy split into Dialysis (NCIT:C15221) and Kidney Transplantation (NCIT:C15265), because the single entry was named more broadly than the transplantation term it bound, leaving the dialysis half unqueryable. Both kidney-replacement evidence items now carry directness INDIRECT: the diaphragmatic-hernia-timing sentence supports dialysis by sequencing other surgery around it rather than by recommending it, and the four-month ESRD figure supports transplantation by establishing the course rather than the intervention.
+
+    Noted and deliberately not changed: the Diffuse Mesangial Sclerosis phenotype shares its name with the pathophysiology node of the same name, so dismech.graph's flat namespace collapses them into one node. That is the documented issue #9896 pattern rather than a curation error, check-causal-targets reports self 0 for this file, and both items carry their own evidence. Binding the lesion in both places was judged better than leaving it unbound to avoid the collapse.
+
+    Validation: 43/43 snippets verified; validate, reference-titles, entity-refs, causal-targets, duplicate-keys, enum-values, snippet-grading and snippet-length all clean; WT1_Disorders still validates with all three members SATISFIED.


bpow:
All six items addressed in cd0545d95. Thank you for catching item 2 — that one was a real error and I had made it worse by asserting the opposite in the PR description.

**On item 2 specifically.** You were right that `HP:0001967` appeared nowhere in the file. What happened is that the wrong CURIE from the deep-research report (`HP:0005567`, *Renal magnesium wasting*) was correctly rejected and the right term was looked up — but the lesion was then modeled only as a pathophysiology node with a cell-type binding, so the HP term never reached the entry. Diffuse mesangial sclerosis is now a bound phenotype with `HP:0001967` alongside the node of the same name, carrying the 34% specificity figure and the biopsy series. I have also corrected the PR description, which previously claimed the term was already in use.

| Item | What changed |
|---|---|
| 1 · `reference_title` | Added to all 41 evidence items, verbatim from the cache frontmatter |
| 2 · DMS unbound | Now a phenotype bound to `HP:0001967`; PR description corrected |
| 3 · No `diagnosis:` | Added, with the GeneReviews establishment sentence and the "screen all children with SRNS" recommendation |
| 4 · Wilms surveillance | Restored, bound to `NCIT:C15406` |
| 5 · No treatment for ESRD | Added kidney replacement therapy and Wilms tumor treatment |
| 6 · Ratelade mouse | Added as `animal_models` with a `RECAPITULATES` link and a target-gene readout |

**Why item 4 was missing is worth recording**, since it looks like an oversight and was closer to a half-finished repair. During initial curation I bound Wilms surveillance to `NCIT:C17167`, then found that term is *Statistics*, not *Screening*. I removed the treatment rather than rebinding it, and never came back. The right fix was `NCIT:C15406` (Cancer Screening), which is what it now uses.

**Not taken this round.** The `clinical_trials` gap (NCT01252901) and a `progression:` section are both real, but neither can be assembled from what this PR already caches — they need their own sourcing pass, and I would rather do that properly than approximate it here. On the uncited "transmission from an affected or mosaic parent" claim: you noted PMID:1327525's cache is title-only, and that is right, so there is nothing quotable. It stays as prose for now rather than being dropped, since it is counselling-relevant.

Validation after the changes: `validate-disorders` passes, 41/41 snippets verified, and the entity-ref, causal-target, duplicate-key, enum-value, title-snippet, snippet-length and snippet-grading gates are all clean. `WT1_Disorders` still validates with all three members SATISFIED.

---

bpow:
Thanks. Took all three suggestions — the first one because I got it wrong, not because it was optional.

**The mosaic-transmission clause was citable all along and I looked at the wrong source.** I deferred it on the grounds that PMID:1327525's cache is title-only, which is true but beside the point: the GeneReviews record already cited in this entry carries both halves outright — *"in rare instances, a parent of an individual with WT1 disorder is heterozygous for the pathogenic variant identified in the proband"* and the gonadal-mosaicism recurrence-risk sentence. I anchored on the reference the earlier round had named and never checked what was already in hand. Both are now evidence items, so the clause is no longer an uncited claim.

Also done:

- `diagnosis_term` bound to `NCIT:C15709`, matching the POT1 entry in this series.
- **`Kidney Replacement Therapy` split into `Dialysis` (`NCIT:C15221`) and `Kidney Transplantation` (`NCIT:C15265`).** You were right that the name promised more than the binding delivered; splitting makes both arms queryable rather than narrowing the name to fit.
- `directness: INDIRECT` on both kidney-replacement items. The hernia-timing sentence supports dialysis by sequencing around it rather than recommending it, and the four-month ESRD figure supports transplantation by establishing the course rather than the intervention. Neither states its claim outright.

On the shared-name point for diffuse mesangial sclerosis: agreed, and thank you for checking it rather than flagging it. The phenotype and the pathophysiology node collapsing into one graph node is the #9896 flat-namespace pattern, and `check-causal-targets` reports `self: 0` here. I would rather have the lesion bound and queryable in both places than avoid the collapse by leaving it unbound.

43/43 snippets verified, all 43 evidence items titled, and the full gate set is clean including `check-reference-titles`, which I had not been running locally before this series — that was the source of the earlier CI failure on the sibling PR.

---

PR #10449, opened 2026-09-01
curate: Add Simpson-Golabi-Behmel Syndrome Type 1 (MONDO:0020602, GPC3)

### LINKED ISSUE #10392: Curate Simpson-Golabi-Behmel syndrome type 1 (MONDO:0020602)
Curate a dismech entry for **Simpson-Golabi-Behmel syndrome type 1** ([MONDO:0020602](https://monarchinitiative.org/MONDO:0020602)) — X-linked overgrowth syndrome caused by loss-of-function variants and deletions in *GPC3*, with pre- and postnatal overgrowth, coarse facies, visceromegaly, and an increased risk of embryonal tumors (Wilms tumor, hepatoblastoma) in childhood.

Working assumption is `DISEASE`, at the type-1 level rather than at the parent. The stub records no MONDO descendants for MONDO:0020602 and names one gene (*GPC3*, hgnc:4451). Its parent, `Simpson-Golabi-Behmel syndrome` (MONDO:0010731), spans type 1 and the separately mapped, mechanistically distinct type 2 — so the parent is the grouping-shaped term and the *GPC3* entity is the one with a conserved pathograph. Neither term is curated today.

Related entry to keep distinct: `kb/disorders/Beckwith-Wiedemann_Syndrome.yaml` — the other major overgrowth-with-embryonal-tumor-risk syndrome, and the natural `differential_diagnoses` counterpart. The mechanisms differ (11p15 imprinting vs. glypican-3 loss and dysregulated hedgehog/IGF signaling), so this is a differential, not a subtype relationship.

Stub: `stubs/Simpson-Golabi-Behmel_Syndrome_Type_1.yaml`

bpow:
Adds an entry for Simpson-Golabi-Behmel syndrome type 1, the X-linked GPC3 overgrowth syndrome with embryonal tumor risk.

The curation decision worth reviewing is which mechanism the pathograph asserts. The textbook account of this syndrome is that glypican-3 restrains IGF2, and that is not what the entry says. That model came from a ligand-blot interaction in the 1996 gene-identification paper plus the clinical resemblance to Beckwith-Wiedemann. The Gpc3-null mouse then reproduced the overgrowth while leaving IGF-II levels unchanged, which is hard to square with simple ligand sequestration.

Closes #10392

## What the entry asserts instead

The causal chain runs through hedgehog:

```
GPC3 loss of function ──► loss of glypican-3 competition for Hedgehog ligand ──► Hedgehog pathway hyperactivation
    ──► excess embryonic cell proliferation ──┬─► generalized pre/postnatal overgrowth ──► overgrowth
                                              └─► persistence of embryonal progenitor tissue ──┬─► nephroblastoma
                                                                                               └─► hepatoblastoma
```

Glypican-3 binds Hedgehog with high affinity, competes with Patched for it, and drives the bound ligand to endocytosis and degradation. Remove GPC3 and that sink disappears. This comes from Capurro 2008 in Dev Cell (the primary report, cited in preference to the Genome Biology review that restates it), which showed increased hedgehog signaling in GPC3-null embryos directly.

**The IGF2 model is kept, not deleted.** It sits as a node with `mechanism_confidence: HYPOTHETICAL`, carrying both the original in-vitro observation as SUPPORT and the null-mouse ligand-level result as REFUTE. It is deliberately left unconnected to the causal chain, since the evidence does not license an edge. It stays in because a large slice of the older literature assumes it, and a reader arriving from that literature should find it addressed rather than absent.

## Two ontology terms the report offered that I did not bind

- `HP:0001524` — does not exist in HPO. The report's own term validation flagged it as unresolved.
- `GO:0031225` — real but obsolete, replaced by `GO:0016020`.

## An evidence item I dropped, and why

PMID:17850639 (germline mosaicism in SGBS) is a PubMed record with no abstract text, so there is nothing to quote and nothing the snippet validator can check. Rather than cite it with an unverifiable snippet, I removed the evidence item and attributed the claim in prose in the GPC3 `notes`, stating there why it is handled that way. Germline mosaicism changes recurrence-risk counselling, so it was worth keeping the claim.

Two other evidence items initially quoted paper titles rather than findings and failed snippet verification. Both were replaced with substantive sentences from the abstracts.

## A gap I left open on purpose

The dysmorphic phenotypes — postaxial polydactyly, supernumerary nipples, coarse facies — have no incoming pathograph edge. I could not source an edge from generalized overgrowth to each of them, and an unconnected node is honest where an invented edge is not. If a reviewer knows a source that supports those links, they are worth adding.

## Validation

Schema, terms and references all pass. 30/30 snippets verified. Entity-ref and causal-target gates clean, plus the whole-KB duplicate-key, enum-value, folded-hyphen, snippet-length, title-snippet and snippet-grading gates. History record validates. The `stubs/Simpson-Golabi-Behmel_Syndrome_Type_1.yaml` stub is deleted.

---
_Generated by [Claude Code](https://claude.ai/code/session_015dtkgpfKRqzPwhcvWNpU9g)_

### DIFF
--- a/history/disorders/Simpson-Golabi-Behmel_Syndrome_Type_1/2026-09-01T235008Z-claude-code-376dc2.yaml
+++ b/history/disorders/Simpson-Golabi-Behmel_Syndrome_Type_1/2026-09-01T235008Z-claude-code-376dc2.yaml
@@ -1,0 +1,40 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Simpson-Golabi-Behmel_Syndrome_Type_1
+  path: kb/disorders/Simpson-Golabi-Behmel_Syndrome_Type_1.yaml
+session:
+  id: 2026-09-01T235008Z-claude-code-376dc2
+  timestamp: '2026-09-01T23:50:08Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+    agent_version: 2.1.252
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10392
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10449
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - inheritance
+  - diagnosis
+  - animal_models
+  - phenotypes
+  - evidence
+  summary: 'Take review suggestions on PR #10449 (SGBS1)'
+  details: |
+    Second response round on PR #10449, after the reviewer approved the blocking fixes and left five suggestions.
+
+    The substantive one was fidelity on the Gpc3-null FAILS_TO_RECAPITULATE link, changed from HIGH to LOW. The slot is defined as how faithfully the model captures the linked human mechanism, so on a link asserting the model does not reproduce IGF2 sequestration, HIGH states the opposite of what was meant. What was actually being encoded was 'this negative result is well characterised', which is a different claim and does not belong in that slot. The KB convention agrees: the overwhelming majority of FAILS_TO_RECAPITULATE links use LOW. This is worth recording because the same misreading was made independently on the POT1 entry in this series, where fidelity HIGH sat beside limitations that bounded the model.
+
+    Also taken: the U+FB01 ligature note moved out of an evidence explanation, which renders as claim content, into the diagnosis block notes where cache mechanics belong; the GeneReviews 50% per-pregnancy transmission sentence added to inheritance; directness INDIRECT on the neonatal hypoglycemia item, whose snippet is a surveillance recommendation rather than a statement that hypoglycemia occurs; and the reason the mutation-update reference carries no snippet recorded in the PublicationReference findings slot rather than deleted, since a curator reading the YAML later would otherwise re-attempt to snippet an unavailable cache.
+
+    Declined, with the reviewer agreeing on re-review: three GeneReviews findings with no stated frequency (diastasis recti and gastrointestinal issues, hearing and ophthalmologic surveillance). Adding them would have put the inference into the graph rather than into a reviewer's head. The distinction that decided it is what the source sentence is doing: neonatal hypoglycemia earns a phenotype because GeneReviews treats it as an urgent recognised complication phrased as an instruction, whereas 'annual audiologic evaluation' is a statement about good practice that does not assert the finding occurs above background. directness is a grade on evidence bearing on a claim, not a licence to manufacture the claim.
+
+    Validation: 48/48 snippets verified; validate, validate-terms, reference-titles, entity-refs, causal-targets, qualifier-terms, duplicate-keys, enum-values, snippet-grading, snippet-length and title-snippets all clean.


bpow:
All five blocking items addressed in 23075c77e, plus the suggestions. I verified each finding against the repository before acting — including the claim that this was the only file of 2,509 with no `reference_title`, which `grep -rL` confirmed.

| Item | What changed |
|---|---|
| 1 · `reference_title` | Added to all 48 evidence items, verbatim from the cache frontmatter |
| 2 · GR Diagnosis unmined | Added a `diagnosis:` block with the GeneReviews establishment sentence and the CUGC sex-dependent testing order |
| 3 · 9 of 15 refs unused | PMID:30683921 now cited five times; DOI:10.1002/humu.23428 recorded in top-level `references:` since it carries no body |
| 4 · Missing phenotypes | Added congenital heart defect, genitourinary anomaly, scoliosis, vertebral fusion, visceromegaly and wide mouth — 12 phenotypes to 18 — plus a cardiac surveillance treatment |
| 5 · No `animal_models` | Gpc3-null mouse added with `RECAPITULATES` on overgrowth and `FAILS_TO_RECAPITULATE` on the IGF2 node |

The `FAILS_TO_RECAPITULATE` link was the most useful thing in this review. The IGF-II result was previously carried only as a `REFUTE` evidence item on a floating node; as a model link with an `UNCHANGED` readout it is now the structural signal rather than a footnote.

**Two defects I introduced while fixing these, both caught by the gates before pushing.** Cardiac surveillance was first bound to `NCIT:C15406` (Cancer Screening) — wrong concept for cardiac follow-up; it now uses `NCIT:C168126` (Cardiac Disease Screening). And I graded the CUGC penetrance sentence `OTHER` under `prevalence` and `HUMAN_CLINICAL` under `genetic`, which `check-snippet-grading` rejected. Rather than make the gradings agree I removed the prevalence evidence item entirely: that sentence is about penetrance and does not speak to prevalence, so citing it there would have misattributed the claim. The prevalence record now carries `notes` explaining that no source states one, and no evidence item.

One snippet deliberately keeps a `U+FB01` ligature. The CUGC cache is PDF-extracted and the gating validator does not fold ligatures, so replacing it with plain "fi" makes the quote stop matching; the `explanation` says so, to stop a future curator "correcting" it.

**Suggestions taken:** `downstream` edges to Macrocephaly, Coarse Facial Features and Visceromegaly, which the overgrowth node already asserted in prose; the Capurro binding sentence regraded to `IN_VITRO`; `prevalence` as `NOT_YET_DOCUMENTED`; and the footnote that SGBS type 2 is PIGA-related disease recorded in the GPC3 notes.

I agree on polydactyly and supernumerary nipples and left them unconnected — twelve phenotypes still have no incoming edge, deliberately.

Validation: `validate-disorders` passes, 48/48 snippets verified, all gates clean.

---

bpow:
Thanks — took all four suggestions in 019a4db46, and the DOI title was already fixed in 970f5e975 before the re-review landed.

**On the DOI title: that was also the CI failure.** `test (3.13)` went red on 23075c77e at `check_reference_titles.py`, flagging the same field at similarity 0.26. I had found and fixed it independently, so the fix was already pushed — but I had not been running `check-reference-titles` locally, which is why it reached CI at all. I have now run it against all three branches in this series and all pass.

**`fidelity: HIGH` on the `FAILS_TO_RECAPITULATE` link — you were right and I have changed it to `LOW`.** Your reading of the slot definition settles it: it is how faithfully the model captures the linked human mechanism, so on a link asserting the model does *not* reproduce IGF2 sequestration, `HIGH` says the opposite of what I meant. I was encoding "this negative result is well characterised", which is a different claim and does not belong in that slot. The convention across `kb/` agrees with you by a wide margin.

Also done:

- The U+FB01 ligature note moved from the evidence `explanation` into the `diagnosis` block's `notes`. You are right that `explanation` renders as claim content and cache mechanics do not belong there.
- Added the GeneReviews 50% per-pregnancy transmission sentence to `inheritance`.
- `directness: INDIRECT` on the neonatal hypoglycemia item, since that snippet is a surveillance recommendation rather than a statement that hypoglycemia occurs.

Left alone: the three GR findings with no stated frequency (diastasis recti / GI issues, hearing and ophthalmologic surveillance). They are real, but adding phenotypes whose only support is "annual evaluation is recommended" would repeat the inference pattern you just flagged on hypoglycemia, and I would rather not multiply it. Happy to add them with `directness: INDIRECT` if you would prefer them present.

48/48 snippets verified; validate, reference-titles, entity-refs, causal-targets, duplicate-keys, enum-values, snippet-grading and snippet-length all clean.

---

PR #10450, opened 2026-09-01
curate: Add POT1 Tumor Predisposition Syndrome (MONDO:0014368, POT1)

### LINKED ISSUE #10393: Curate tumor predisposition syndrome 3 (MONDO:0014368)
Curate a dismech entry for **tumor predisposition syndrome 3** ([MONDO:0014368](https://monarchinitiative.org/MONDO:0014368)) — better known as **POT1 tumor predisposition (POT1-TPD)**: heterozygous loss-of-function variants in the shelterin component *POT1* (hgnc:17284) that lengthen telomeres and predispose to cutaneous melanoma, glioma, angiosarcoma, and chronic lymphocytic leukemia.

Working assumption is `DISEASE`, but the call deserves stating rather than assuming, because this term looks like a susceptibility term at first glance and the claim-disease guidance treats those as usually out of scope:

- Its MONDO parents are both susceptibility terms — `susceptibility to familial cutaneous melanoma` (MONDO:0024462) and `glioma susceptibility` (MONDO:0100242) — and several of its synonyms are locus-style labels (CMM10, GLM9).
- Against that, POT1-TPD is a recognized single-gene syndrome with one conserved mechanism (shelterin disruption and telomere elongation) upstream of a defined tumor spectrum, and it has published surveillance guidance. That is the same shape as the germline predisposition syndromes already curated here — `Li-Fraumeni_Syndrome`, `Lynch_Syndrome`, `DICER1_Tumor_Predisposition_Syndrome` — which design decisions §3a keeps as their own entries, separate from the somatic tumors they predispose to.

So the plan is a disease entry for the syndrome, with the melanoma/glioma/angiosarcoma/CLL spectrum as tumor outcomes rather than as separate entries. The stub records no MONDO descendants, so there is nothing below it to group.

Two naming points for whoever picks this up:

- The stub's `proposed_name` is `Tumor_Predisposition_Syndrome_3`, which is the MONDO label but not how the entity is referred to in the literature. `POT1_Tumor_Predisposition_Syndrome` is the more findable filename; the MONDO label still goes in `disease_term`.
- *POT1* is already mentioned in `kb/disorders/Angiosarcoma.yaml` and `kb/disorders/Dyskeratosis_Congenita.yaml`. Neither carries MONDO:0014368, so there is no duplicate — but the telomere-biology relationship to the dyskeratosis congenita entry is worth a cross-reference, noting that POT1-TPD is the long-telomere counterpart of the short-telomere disorders.

Stub: `stubs/Tumor_Predisposition_Syndrome_3.yaml`

bpow:
Adds an entry for POT1 tumor predisposition, the shelterin cancer predisposition syndrome behind melanoma, cardiac angiosarcoma, glioma and CLL.

Two things about it are unusual enough to flag up front. It is a **long**-telomere disorder, not a short-telomere one: losing POT1 removes a brake on telomerase access to the 3' overhang, so it is the mechanistic mirror image of dyskeratosis congenita and has none of the marrow failure. And it is defined against a negative — it was found by sequencing TP53-negative Li-Fraumeni-like families, and that remains when to think of it.

Closes #10393

## The entry-type call, which is the reviewable decision here

MONDO puts this term under two susceptibility parents (`susceptibility to familial cutaneous melanoma`, `glioma susceptibility`) and several of its synonyms are locus-style labels (CMM10, GLM9). That is exactly the shape the claim-disease guidance says is usually out of scope.

I curated it as a `DISEASE` anyway. It is a single-gene syndrome with one conserved mechanism upstream of a defined tumor spectrum, and it has published surveillance guidance — the same shape as `Li-Fraumeni_Syndrome`, `Lynch_Syndrome` and `DICER1_Tumor_Predisposition_Syndrome`, which design decisions §3a keeps as their own entries. If a reviewer disagrees, this is the place to say so.

The file is named `POT1_Tumor_Predisposition_Syndrome` rather than the MONDO label `tumor predisposition syndrome 3`, which nobody uses. The MONDO label remains the `disease_term` binding.

## An open question the entry records rather than resolves

Whether tumorigenesis here is driven by telomere elongation or by genomic instability is unsettled, and the entry says so instead of picking one:

| Evidence | Points to |
|---|---|
| Human carriers: long telomeres **and** increased telomere fragility (PMID:26403419) | both |
| CRISPR-engineered hESCs and HSCs: consistent elongation, no telomere damage response (PMID:33934394) | elongation |

The instability node carries `mechanism_confidence: PROVISIONAL` and a REFUTE item from the stem-cell work; an OPEN `KNOWLEDGE_GAP` discussion quotes the authors' own statement that "which of the two phenotypes promotes tumorigenesis is unclear." This matters practically: an instability-driven mechanism would predict sensitivity to DNA-damaging therapy that an elongation-driven one would not.

## A wrong NCIT term worth knowing about

While looking for a surveillance term I found that **`NCIT:C17167` is *Statistics*, not *Screening***. It reads like the obvious choice and is not. `NCIT:C15406` (Cancer Screening) is used here. Worth a glance at any existing entry that reached for C17167.

## Reference handling

The falcon report cited entirely by DOI. Each was converted through the PMC ID converter and the PMID cited in preference, since `DOI:` is in `skip_prefixes` and DOI-keyed snippets are not checked by the gating validator.

## Validation

Schema, terms and references all pass. 28/28 snippets verified. Entity-ref and causal-target gates clean, plus the whole-KB duplicate-key, enum-value, folded-hyphen, snippet-length, title-snippet and snippet-grading gates. History record validates. Named Entity Confusion preflight PASS (POT1 mentioned 69 times; report OMIM 615848 matches MONDO). The `stubs/Tumor_Predisposition_Syndrome_3.yaml` stub is deleted.

---
_Generated by [Claude Code](https://claude.ai/code/session_015dtkgpfKRqzPwhcvWNpU9g)_

### DIFF
--- a/history/disorders/POT1_Tumor_Predisposition_Syndrome/2026-09-01T235038Z-claude-code-c424f2.yaml
+++ b/history/disorders/POT1_Tumor_Predisposition_Syndrome/2026-09-01T235038Z-claude-code-c424f2.yaml
@@ -1,0 +1,40 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: POT1_Tumor_Predisposition_Syndrome
+  path: kb/disorders/POT1_Tumor_Predisposition_Syndrome.yaml
+session:
+  id: 2026-09-01T235038Z-claude-code-c424f2
+  timestamp: '2026-09-01T23:50:38Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+    agent_version: 2.1.252
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10393
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10450
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - diagnosis
+  - animal_models
+  - inheritance
+  - treatments
+  - evidence
+  summary: 'Take review suggestions on PR #10450 (POT1-TPD)'
+  details: |
+    Second response round on PR #10450, after the reviewer approved the blocking fixes and left six non-blocking notes. Four taken.
+
+    The substantive one was fidelity on both Pot1a R117C mouse links, changed from HIGH to MODERATE. The entry contradicted itself: its own limitations lines said the Pot1a/Pot1b paralogue split bounds extrapolation to human POT1, and that the model covers only the angiosarcoma arm of a four-tumour spectrum, while the fidelity grade said the model captures the human mechanism faithfully. MODERATE is what those caveats already imply. This was the same misreading of the slot made on the SGBS1 entry in this series, where HIGH sat on a FAILS_TO_RECAPITULATE link; in both cases the value being encoded was how well characterised the result is rather than how faithfully the model maps to human biology, which is not what the slot means.
+
+    Also taken: diagnosis_term bound to NCIT:C15709 (Genetic Testing), matching 416 other disorder entries; the telomerase-dependence readout now states that it is measured in Pot1a+/ki Tert-/- double mutants rather than the declared Pot1a+/ki genotype, since the comparison is the experiment and a reader should not have to infer it; and the GeneReviews genetic counseling gap closed with the 50% per-child transmission figure on inheritance and cascade testing of first-degree relatives as a treatment, carrying the age-18 threshold and its early-cancer exception.
+
+    Not done, and the reason is the same for both. Genetic.case_fractions and the Ashkenazi p.Ile78Thr founder allele both need PMID:32987645 as full text; it is cached abstract_only, and the per-cohort figures the reviewer quoted (4/105, 7/56, 2/55) appear in the deep-research report rather than in anything a snippet can be verified against. Copying them from the report into a citation is the fabrication risk the evidence SOP exists to prevent. Re-fetching that paper as full text makes both straightforward. PMID:38573059 stays uncited for the same reason: its POT1 content could not be isolated to a quotable sentence.
+
+    Validation: 41/41 snippets verified; validate, reference-titles, entity-refs, causal-targets, duplicate-keys, enum-values and snippet-grading all clean.


bpow:
All three blocking items addressed in 7348d9e6c, plus most of the suggestions.

| Item | What changed |
|---|---|
| 1 · `reference_title` | Added to all evidence items (28, now 38), verbatim from the cache frontmatter |
| 2 · GR Diagnosis unmined | Added a `diagnosis:` block with the GeneReviews establishment sentence and the Andreotti caveat |
| 3 · R117C mouse unused | PMID:35727838 added as `animal_models` with two `modeled_mechanisms` links |

You were right that the mouse is not peripheral. It carries the same R117C allele the angiosarcoma phenotype is built on, and it independently supplies two things the entry previously asserted on other grounds. Both are now readouts: telomere length `INCREASED` in knock-in MEFs and tissues, and that elongation `ABOLISHED` on a *Tert*-null background, which is what makes the telomerase dependence a measurement rather than an inference. It also links `RECAPITULATES` to Angiosarcoma with a spontaneous-incidence readout. The Pot1a/Pot1b paralogue split is recorded as the `limitations` line, as you suggested.

**Your point about telomere length not being diagnostic is now in the entry.** An entry whose whole mechanism is telomere length should say that measuring telomere length does not make the diagnosis, and the `diagnosis:` block description says exactly that.

**Suggestions taken:**

- The p.Tyr36Cys sentence is a structural prediction, so `HUMAN_CLINICAL` became `COMPUTATIONAL`. You were right that the split was already half done.
- `GO:0000723` on the replicative-capacity node described the upstream telomere state rather than that node's own claim. Replaced with `GO:0090399` (replicative senescence, `DECREASED`) — closer to "postponed senescence" than an approximate telomere term.
- Added melanocyte, endothelial cell, B cell and glial cell as `cell_types` on that node, so the lineage restriction is queryable. The mouse paper localizing long telomeres to endothelial cells made this worth doing.
- Andreotti's argument against automatic maximal surveillance added to the treatment evidence, and BAP1-related tumor predisposition plus familial melanoma added as differentials.

**On frequencies**, I took the note option rather than changing values. The entry now states in its description that the tumor frequency bands are a curator-assigned ordering of how consistently each tumor appears in the reported spectrum — not measured rates — and must not be read as risk estimates. Changing them to a two-way core/emerging split would lose the melanoma-versus-rest ordering, which the sources do support.

**One thing I could not do.** `Genetic.case_fractions` from PMID:32987645 is exactly the right home for those per-cohort numbers, but that record is cached `abstract_only` and the fractions are in its full text. Rather than paraphrase them from the deep-research report, I recorded the reference in the top-level `references:` block. Same for the Ashkenazi founder allele. If someone re-fetches that paper as full text these become straightforward.

I also introduced and fixed one defect: the top-level `references:` entry for that paper initially had an explanatory sentence appended to its `title`, which the validator checks against the cache. Title now matches exactly.

Validation: `validate-disorders` passes, 38/38 snippets verified, all gates clean.

---

bpow:
Thanks for the approval. Took four of the six non-blocking notes — this push dismisses the approval, which seemed worth it for the `fidelity` correction in particular.

- **`diagnosis_term` bound** to `NCIT:C15709` (Genetic Testing). 416 entries use it; there was no reason for this one to be the exception.
- **`fidelity: HIGH` → `MODERATE` on both mouse links.** You are right that it contradicted the entry's own `limitations`: I wrote that the Pot1a/Pot1b paralogue split bounds extrapolation and that the model covers one of four tumour arms, then graded it HIGH anyway. `MODERATE` is what the caveats already say.
- **The double-mutant genotype is now stated in the readout.** `Pot1a+/ki Tert-/-` is the comparison arm, not the declared genotype, and a reader should not have to infer that.
- **Genetic counselling gap closed**: the 50% per-child transmission figure added to `inheritance`, and cascade testing of first-degree relatives added as a treatment — with the age-18 threshold and its early-cancer exception, since that exception is the part that matters in a pediatric setting.

**Two I did not do, both for the same reason.** `Genetic.case_fractions` and the Ashkenazi p.Ile78Thr founder allele are both real gaps, and both need PMID:32987645 as full text — it is cached `abstract_only`, and the numbers you quoted (4/105, 7/56, 2/55) appear in the deep-research report rather than in anything I can verify a snippet against. Copying them from the report into a citation would be exactly the fabrication the evidence SOP is guarding against. Re-fetching that paper as full text makes both straightforward, and I would rather that happen than approximate it now.

`PMID:38573059` stays uncited for the same reason — it is a childhood-brain-tumour surveillance guideline whose POT1 content I could not isolate to a quotable sentence. The duplicate DOI/PMID cache files I have left alone as you suggested.

41/41 snippets verified; validate, reference-titles, entity-refs, causal-targets, duplicate-keys, enum-values and snippet-grading all clean.

I am stopping the polish cycle here rather than pushing again for further suggestions — happy to pick any of the above up if you would rather they land in this PR than a follow-up.

---

PR #10452, opened 2026-09-01
curate: add Colchicine Poisoning (MONDO:0017859)

caufieldjh:
## Colchicine Poisoning (MONDO:0017859)

New toxicology entry. Colchicine has one molecular lesion and two clinical syndromes, and the point of the entry is that the split is a matter of which microtubule function a tissue depends on.

Colchicine binds tubulin and prevents microtubules assembling. Cells that need the mitotic spindle fail first — gut mucosa, then marrow — and it is the resulting pancytopenia with neutropenic sepsis that kills. Cells that need microtubule-based *transport* rather than mitosis fail far more slowly, giving the proximal myopathy and axonal neuropathy that show up at ordinary prescribed doses. One lesion, two tissues, two clocks.

### Pathograph

12 nodes, fully connected apart from the CYP3A4/P-glycoprotein clearance node, which is an intentional modifier entry point.

Six `conforms_to` declarations across two modules:
- **`myelosuppression`** — the complete chain: cytotoxic insult to progenitors → marrow suppression → multilineage cytopenias → cytopenia-related complications.
- **`peripheral_axonal_degeneration`** — axonal transport impairment, and distal axonal degeneration.

Both `peripheral_axonal_degeneration` conformances carry `notes` recording that they are **partial**: the module nodes name mitochondrial dysfunction and axonal/demyelinating fibre pathology, and the human sources cited here establish neither. The systematic review's predominant finding is myopathic rather than neuropathic, and the entry says so rather than smoothing it over.

A hepatorenal feedback edge closes a loop back onto the exposure node — hepatic elimination carries most of colchicine's clearance, so injuring the liver slows removal of the poison doing the injuring.

### Evidence

66 evidence items over 9 references. Every snippet was pre-verified against the local cache **before** the YAML was written, using the repository's own matcher. Anchors are two systematic reviews (PMID:20586571; PMID:36512928, 143 neuromyopathy cases), a full-text management review with a 20-case series (PMID:36028733), and case reports for the marrow and G-CSF claims.

**Frequency discipline.** `FREQUENT` is asserted only for the three gastrointestinal symptoms, and each explanation names the denominator (14 of 20 published adult case reports) and states that the series skews severe. No band is asserted for any cytopenia or organ-failure phenotype, because publication bias inflates exactly those figures — acute myocardial injury reads 80% in that series, and that number is not a population frequency.

**Claims deliberately not made:**
- No bacterial-translocation claim on the neutropenic-sepsis edge. Blood cultures were positive in 5 of 20, but the route of entry was not established, and the node says that.
- No mitochondrial claim on the transport arm.
- Fab fragments and extracorporeal drug removal are recorded in `notes` as *not* curated. The volume-of-distribution figure justifying the latter is carried as a cited snippet, not as bare prose — a number in a prose field is unverifiable by any check we have.

### Deep research

`just research-disorder claude_code` produced a 95 KB report; 59/59 references resolved, 0 unresolved, 0 off topic. NEC preflight returned `SKIP` (MONDO records no causal gene for a poisoning) and was cleared manually — CYP3A4 and ABCB1 dominate the report's gene mentions, and those are colchicine's own metabolism and efflux genes.

The report was used as leads only; **zero snippets were taken from it**. It did surface two genuine gaps in my draft — the cardiac branch and the hepatorenal feedback loop — and both were then verified independently against papers already in the cache before being added.

### Validation

- `just validate-disorders` (the batched sweep CI runs) — passed
- 66/66 snippets verified against cached references
- `check-duplicate-keys`, `check-folded-hyphens`, `check-title-snippets`, `check-entity-refs`, `check-causal-targets`, `check-environmental-evidence` — all clean
- Every `reference_title` generated programmatically from the cache files, never typed

### DIFF
--- a/history/disorders/Colchicine_Poisoning/2026-09-01T225327Z-claude-code-372339.yaml
+++ b/history/disorders/Colchicine_Poisoning/2026-09-01T225327Z-claude-code-372339.yaml
@@ -107,3 +107,19 @@ events:
     the clearance node, the statin-myopathy interaction sentence added, and
     gastric lavage recorded in the charcoal treatment as a narrow adjunct rather
     than curated as a separate treatment.
+
+    Review round 2 (ai4c-reviewer, CHANGES_REQUESTED) found one defect introduced
+    by the round-1 fix: the plant-route notes: still said the sources gave no
+    case-fatality figure for that route and that none was asserted, three lines
+    below the evidence item quoting 10 percent (29/297) from PMID:30888889. The
+    note was written before that paper was pulled in and was never revisited. It
+    now states the figure and marks it as a population rate from one country's
+    self-poisoning presentations rather than a case fatality for plant ingestion
+    generally. The trigger node's "the two routes" phrasing was also updated, since
+    the description now names three.
+
+    Swept the rest of the entry for the same failure mode - prose that was true
+    when written and falsified by later evidence. Every remaining negative claim
+    about the sources was re-checked against its block, and every phenotype was
+    cross-checked so that no frequency band coexists with an explanation saying
+    none is asserted. None found.
--- a/kb/disorders/Colchicine_Poisoning.yaml
+++ b/kb/disorders/Colchicine_Poisoning.yaml
@@ -30,11 +30,12 @@ disease_term:
 pathophysiology:
 - name: Colchicine Exposure Exceeding the Therapeutic Window
   description: >-
-    Systemic colchicine burden above what the patient can tolerate, reached either
-    by a single large ingestion or by accumulation at ordinary doses. The drug has
-    no clear-cut boundary between nontoxic, toxic and lethal doses, which is why
-    the two routes are not usefully separated at the level of the exposure - the
-    determinant is the concentration reached, not the size of any one dose.
+    Systemic colchicine burden above what the patient can tolerate, reached by a
+    single large ingestion of tablets or plant material, or by accumulation at
+    ordinary doses. The drug has no clear-cut boundary between nontoxic, toxic and
+    lethal doses, which is why the routes are not usefully separated at the level
+    of the exposure - the determinant is the concentration reached, not the size of
+    any one dose.
   role: trigger
   biological_scale: ORGANISM
   evidence:
@@ -1361,9 +1362,11 @@ environmental:
     Bound to ECTO:9000271 exposure to alkaloid rather than to ECTO:0000509
     exposure to drug, because this route is not a medicinal exposure. Neither term
     is colchicine-specific; ECTO has no such class through this repository's
-    adapter. The cited sources establish the plants as the source and the route as
-    a recognised exposure history. They do not give a case-fatality figure for the
-    plant route specifically, and none is asserted here.
+    adapter. The cited sources establish the plants as the source, the route as a
+    recognised exposure history, and a case fatality of 10 percent in a
+    297-patient Gloriosa superba cohort. That figure is a population rate from one
+    country's self-poisoning presentations and should not be read as a case
+    fatality for plant ingestion generally.
 treatments:
 - name: Granulocyte colony-stimulating factor
   description: >-


caufieldjh:
Both blocking findings were correct and both are closed. So are all four non-blocking ones.

### 1. `diagnosis:` — added

You were right that the entry asserted diagnostic difficulty twice in its own prose while modeling none. Added three `diagnosis:` entries — exposure-history-plus-toxidrome recognition, the separate problem of recognising the chronic neuromyopathy, and MRI for biopsy targeting — plus a `histopathology:` entry for the vacuolar myopathy. The MRI entry carries its source's own hedge and says explicitly that this entry does not curate it as a diagnostic modality.

### 2. Plant route — added, and the ECTO note rewritten

Second `environmental:` entry for *Colchicum autumnale* and *Gloriosa superba*, bound to `ECTO:9000271 exposure to alkaloid`. The description now says three routes, not two.

Your second paragraph was the more useful half of the finding. The original ECTO note rejected `exposure to alkaloid` because "the poisoning is a medicinal one" — a scoping premise the entry never declared and that the plant route falsifies. I rewrote the note rather than patching around it. It now says the drug entry is bound to `ECTO:0000509` *because that entry is the medicinal route*, and that `ECTO:9000271` is used on the plant entry because that route is not a drug exposure.

### What following the route turned up

`PMID:30888889` — a 297-patient Sri Lankan *Gloriosa superba* cohort, now fetched and cached. Three things came out of it:

- **Case fatality 10% (29/297).** Added to the plant entry. The explanation states that this is a population rate and is *not* comparable with the 14–26% quoted elsewhere in the entry, which comes from published tablet-overdose case reports.
- **"Ingestion of *Gloriosa superba* caused only mild acute kidney injury (AKI) and rhabdomyolysis."** Both of those phenotypes were evidenced from the 20-case tablet series. They now carry a route-dependence qualifier so the entry is not read as asserting one severity.
- **Admission concentration predicts death** (AUC-ROC 0.98). Added to the exposure node, which previously rested only on the absence of a dose boundary.

### Overstatements I found re-reading my own additions

Three, all corrected before pushing:

- "withdrawal is curative" — the review says complete resolution in 70%, so 30% did not.
- vacuolar myopathy "confirms" — the source says "typically present", not invariably. Now curated as characteristic, not required.
- "There is no bedside test" — flat, and the sources do not say it. Replaced with what they do say, plus the note that plasma concentration is a research-cohort prognostic measurement rather than a routine diagnostic test.

The figures in that prose now carry their own citations rather than sitting as unsourced numbers.

### Non-blocking

- P-glycoprotein bound to `GO:0042908 xenobiotic transport` on the clearance node, with the enterocyte-efflux sentence as evidence.
- Statin/myopathy interaction sentence added. The source says "may increase the risk" and does not quantify, so no magnitude is asserted.
- Gastric lavage recorded in the charcoal treatment's description as a narrow adjunct for very large ingestions within an hour, not curated as a separate treatment.

### Validation

84/84 snippets verified over 10 references. `just validate-disorders`, `check-title-snippets`, `check-environmental-evidence`, `check-folded-hyphens` and `check-duplicate-keys` all re-run clean on the final content, after the last edit rather than before it. History record updated to match.

Ready for re-review.

caufieldjh:
Fixed, and you were right about the shape of it. I added the evidence and left the sentence denying it standing three lines below.

**The stale `notes:`** now reads:

> The cited sources establish the plants as the source, the route as a recognised exposure history, and a case fatality of 10 percent in a 297-patient *Gloriosa superba* cohort. That figure is a population rate from one country's self-poisoning presentations and should not be read as a case fatality for plant ingestion generally.

I kept a qualifier rather than just deleting, because the original disclaimer was doing one useful thing badly — the 10% is Sri Lankan *Gloriosa* self-poisoning, and reading it as the plant-route figure in general would be wrong.

**Non-blocking, taken:** the trigger node's "the two routes" now reads "the routes", and the sentence names tablets or plant material at the ingestion clause.

**`GO:0042910`** — noted, not changed. The node's other grounded slot is a molecular function, so a molecular-function framing of the efflux arm would sit more naturally there than the process term does. I left it because the current binding is correct and the swap is a preference, not a fix. Worth revisiting if anyone standardises how transporter arms are bound across the toxicology entries.

**Swept for the same failure mode.** Prose that was true when written and falsified by evidence added later is not a one-off — it is what happens whenever evidence arrives after the sentence. So I re-checked every remaining negative claim about the sources against its own block, and cross-checked every phenotype so that no `frequency:` value coexists with an explanation saying no band is asserted. Nothing else found. The nine surviving disclaimers are all still accurate: the co-occurrence/not-axonal note, two no-frequency-band notes, the plasma-concentration note, the not-a-required-feature note, gastric lavage, and the Fab and extracorporeal notes.

84/84 snippets. `just validate-disorders`, `check-folded-hyphens`, `check-duplicate-keys` and `check-title-snippets` all re-run clean on the final content, after the edit. History record carries round 2.

PR #10455, opened 2026-09-02
Curate familial episodic pain syndrome 3 (SCN11A, MONDO:0014247)

### LINKED ISSUE #10413: Curate familial episodic pain syndrome with predominantly lower limb involvement (MONDO:0014247)
Claiming **MONDO:0014247** — familial episodic pain syndrome with predominantly lower limb involvement (FEPS3), *SCN11A* (Na_v_1.9).

Stub: `stubs/Familial_Episodic_Pain_Syndrome_With_Predominantly_Lower_Limb_Involvement.yaml`

**Lump/split check.** MONDO parents are `syndromic disease` and `familial episodic pain syndrome`; no descendants; single causal gene SCN11A. The parent term `familial episodic pain syndrome` groups FEPS1 (TRPA1) and FEPS2/3 (SCN11A) — genetically distinct, so this is a disease rather than a subtype of a curated entry. No existing KB entry covers it. Curating as `entry_type: DISEASE`.

nlharris:

New entry for **familial episodic pain syndrome with predominantly lower limb involvement** (FEPS3) — an autosomal dominant *SCN11A* channelopathy. Closes #10413.

## Why the mechanism produces *this* disease

NaV1.9 is unusual among sodium channels: it has the most hyperpolarized activation of any of them and carries a large persistent current, so its job is to set the resting membrane potential of nociceptors rather than to drive the action potential upstroke. A gain-of-function variant therefore doesn't make nociceptors fire harder — it makes them sit closer to threshold.

```
heterozygous SCN11A gain-of-function variant
  → hyperpolarized NaV1.9 activation, enlarged window current
      → depolarized resting potential in small-diameter DRG neurons
          → nociceptor hyperexcitability, reduced firing threshold
              → ectopic signalling from distal limb afferents → episodic pain, lower limb pain
              → myenteric neuron hyperexcitability (variant-dependent) → abdominal pain, constipation
```

The key detail is that the hyperexcitability is **cold-resistant** — it persists at temperatures where nociceptor firing is normally suppressed. That is the cellular reason a pain syndrome is *triggered* by cold and *relieved* by warming, which is the exact inverse of erythromelalgia and the discriminator between them at the bedside. There is no tissue lesion anywhere: nerve conduction studies, imaging and inflammatory markers are all normal.

## Scope: the disease name doesn't pick a gene

`MONDO:0014247` is named for a phenotype, and FEPS2 (*SCN10A*) also causes distal, lower-limb-predominant episodic pain. The research report flags this ambiguity in its own limitations, and `just preflight-dr` returned **WARN**, not pass, with SCN10A mentioned at 43% of the SCN11A rate.

The ontology settles it rather than the prose: `MONDO:0014247` carries **OMIM 615552 = FEPS3 = *SCN11A***. So this entry is SCN11A only; FEPS1 (*TRPA1*), FEPS2 (*SCN10A*) and FEPS4 (*SCN9A*) are curated as differentials, and the report's SCN10A sections were excluded.

## This report's term suggestions were unusually bad

None were adopted. Its own term validation flagged five identifiers as naming a different concept:

| Report called it | It actually is |
|---|---|
| `HP:0025277` episodic pain | **Gustatory sweating** |
| `HP:0500152` cold-induced pain | **Hypocystinemia** |
| `HP:0003394` limb pain | **Muscle spasm** |
| `GO:0045211` axon/terminal | **postsynaptic membrane** |
| `UBERON:0001132` trigeminal ganglion | **parathyroid gland** |

A sixth, `CL:0011103` for "enteric neuron", is **sympathetic neuron**.

**And then the part its own validation could not see.** The term-validation section doesn't check NCIT, and **all four** NCIT codes the report proposed for treatments are wrong:

| Report called it | It actually is |
|---|---|
| `NCIT:C15272` Supportive Care | **Lobectomy** |
| `NCIT:C376` Carbamazepine | **Cisplatin** |
| `NCIT:C198` NSAID | **Acetaminophen** |
| `NCIT:C15311` Genetic Counseling | **Quality Control** |

Every identifier bound in this entry was checked against its ontology before use. Worth noting the converse too: `HP:0032148` **Episodic pain** is the exact term for the defining phenotype of this disease, and the report never proposed it at all.

## Other decisions worth a look

- **The NaV1.9-null mouse is included as a deliberately inverted model.** It carries the *opposite* lesion, so it is `PERTURBS` rather than `RECAPITULATES`, its `limitations` say so plainly, and its readout is graded `directness: INDIRECT`. Deleting the channel raises C-fibre threshold by 55% — it establishes necessity where the gain-of-function knock-ins establish sufficiency, and flattening that into "a mouse model of the disease" would misrepresent it.
- **Two knowledge gaps, both genuinely open.** What quantitative degree of gain of function separates episodic pain from *congenital insensitivity to pain* — at Leu811 the two arise from different substitutions of the same residue, and no measured boundary exists, so a novel variant with demonstrated gain of function can't yet be assigned a phenotype on biophysics alone. And why episodes remit around adolescence when the variant is lifelong, which is the most clinically important fact about the disease and entirely unexplained.
- **No pharmacotherapy curated.** The report lists NSAIDs, carbamazepine, lidocaine and mexiletine, but the only paper it cites for any of them reports carbamazepine correcting a **NaV1.8** channel in **diabetic neuropathy** — different channel, different disease. The PRMT7 inhibitor and transcutaneous cholesterol are preclinical mechanism studies, not treatments. Five references the research run fetched were read, set aside, and their caches are not shipped.
- **Source defects reproduced, not repaired.** `PMID:30549873`'s cached abstract has line-break hyphenation artifacts (`fa- tigue`, `in- creasing`). Quotes spanning them reproduce the artifact exactly, with a note saying why — a snippet has to be a true substring, and silently tidying a quote is worse than an ugly one.
- **Neither environmental trigger is bound.** ECTO has no ambient-cold or physical-exertion term (nearest matches: "exposure to cold room", "exposure to cold cereal via ingestion"). Both searches are recorded in `review_notes` rather than binding a near miss.

## Validation

`just validate` clean — schema, terms, references, **62/62 snippets verified**. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence`, `validate-history` all OK.

`reference_title` is on all 62 evidence items, derived programmatically from each cache's frontmatter rather than typed.

Cache diff is additive: 10 reference caches (all cited, reconciled by re-deriving the cited list from `reference:` lines) and 3 term rows. One pre-existing cache showed a modification from the fetcher rewriting its frontmatter format — unrelated churn, so it's reverted to keep the diff additive.

Stub deleted, history record added.

---


nlharris:
All six blocking items are fixed in `631fbf832`, plus four of the five suggestions. One push.

## The critical finding was mine and you're right

I asserted "no tissue damage at any point" while citing, five times, a review that reports the opposite. That's the worst kind of error in this KB — not a missing detail but a confident claim contradicted by a source already in the file.

What I got right was narrower than what I wrote: nerve conduction studies *are* normal, imaging and inflammatory markers *are* normal, there *is* no joint pathology. The error was generalizing from those to the whole nervous system, when routine studies interrogate **large** fibres and this lesion is in the **small** ones. Both statements are now scoped, and the small-fibre lesion has its own `histopathology` record.

**Your second finding is really the same finding, and it's the better half.** The remission gap said the age-dependent remission was "entirely unexplained" — while the same paragraph of the same review proposes that long-term NaV1.9 dysfunction may itself degenerate the unmyelinated fibres in patients whose pain remits. That is a genuinely unsettling idea: it would make the good prognosis and the nerve damage the same event. The gap now carries it as an evidence item, hedged as its authors hedge it, and the rationale says so explicitly. Thank you for connecting those two — I had the histopathology finding and the remission gap as separate items and would have fixed them separately.

## Term rebindings

| Was | Now | Why |
|---|---|---|
| `GO:0086010` membrane depolarization during action potential | `GO:0051899` membrane depolarization | The node's own description says NaV1.9 does *not* drive the AP upstroke — the binding contradicted the node it annotated |
| `CL:0000101` sensory neuron (myenteric node) | `CL:0007011` enteric neuron | A second wrong binding after correctly rejecting `CL:0011103`. Myenteric neurons are motor/interneuron populations, and the evidence snippet on that very node says "myenteric neurons" |

The second one is the more instructive miss: I checked the report's suggestion, found it wrong, and then reached for a fallback without checking that one too. Rejecting a bad term doesn't make the replacement right.

## Conformance and diagnosis

Four nodes now `conforms_to` `nociceptor_sodium_channel_excitability`, using exactly the mapping you gave. I verified all four target node names exist in the module and ran the conformance suite — 2695 tests pass. You're right that the module's description is a direct restatement of this entry's first knowledge gap; that should have been the cue to look for it.

`diagnosis:` added with panel sequencing (the four-gene clinical overlap is what makes a panel rather than single-gene testing the right first test) and quantitative sensory testing. Agreed that the abnormal thermal thresholds were worth capturing on their own — they're the objective correlate of the feature the whole disease turns on.

## Suggestions

Taken: **hyperhidrosis**; **growing pains** as a differential (you're right that it's the one actually faced at the bedside, and the discriminator is family history and genotype rather than the pain description); a note that the published distribution extends to **proximal limbs and neck**, so the lower-limb predominance in the name is a tendency not a boundary; and a note on **Tremor** recording that it is *deliberately* unconnected — no mechanism links NaV1.9 to tremor, it's single-pedigree co-segregation, and drawing an edge would assert a causal claim the source doesn't make. Silence there was worse than a sentence.

Not taken as a structural change: the differentials now *name* `Primary_Erythermalgia` and `Congenital_Insensitivity_to_Pain` in their descriptions, but `DifferentialDiagnosis` has no cross-entry reference slot, so this is prose rather than a link.

**On the `peripheral_axonal_degeneration` conformance** you flagged via the module notes: I haven't added it. The module note says an entry curating the small-fibre-degeneration branch should conform to both, and this entry now curates that branch as histopathology — but not as pathophysiology nodes, and `conforms_to` lives on pathophysiology. Adding it would mean building out a degeneration chain from one sentence reporting a finding "in some patients" with no frequency. I'd rather flag it than half-build it. If you'd prefer it modelled, say so and I'll add the nodes.

## Validation

`just validate` clean — schema, terms, references, **70/70 snippets verified** (was 62/62). `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `validate-history` all OK. Conformance suite passes.

No new reference caches — every added quote comes from a paper this entry already cited.

---

nlharris:
Three binding-precision fixes, pushed at the maintainer's request. No evidence, snippet, claim or pathograph node was touched; snippets stay 70/70.

**Quantitative sensory testing** moves off `NCIT:C25294` (Laboratory Procedure), the KB's generic fallback. NCIT turns out to have an exact term — `NCIT:C155860` **Quantitative Sensory Testing** — which is reachable from the `NCIT:C25218` root and validates. That is better than the `NCIT:C81313` (Neurologic Examination) I had noted as the candidate; I searched NCIT properly rather than reusing the note. Adds one row to `cache/ncit/terms.csv` and one to the `treatmentactionterm` enum cache.

**Two differentials now carry MONDO terms**, both converging on what the KB already binds for those diseases rather than introducing new ones:

| Differential | Now binds | Matches |
|---|---|---|
| Inherited erythromelalgia | `MONDO:0007571` primary erythermalgia | `kb/disorders/Primary_Erythermalgia.yaml` |
| Congenital insensitivity to pain | `MONDO:0015364` hereditary sensory and autonomic neuropathy | `kb/disorders/Congenital_Insensitivity_to_Pain.yaml` |

The erythromelalgia one had a `disease_term` with only a free-text `preferred_term`; the congenital-insensitivity one had no `disease_term` at all. Both differentials' descriptions already named the curated entry in prose, so this makes the cross-reference machine-readable. On the second, `preferred_term: congenital insensitivity to pain` carries the specificity the broader ontology label drops.

**Validation:** `just validate` passed, `just validate-terms` passed, 70/70 snippets verified, and `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` / `check-qualifier-terms` / `check-enum-values` all clean. History record added.

One thing worth flagging for the repo rather than this PR: `just validate` modified `references_cache/PMID_29299961.md` as a side effect — rewriting `reference_id` quoting and adding `full_text_attempted: true`, with a byte-identical body. I reverted it so it is not in this diff. I had filed that churn as #10460 against `just fetch-reference`; it turns out `just validate` triggers it too, which makes it easier to commit by accident than the issue describes. I have updated the issue.

---

Issue #10454, opened 2026-09-01
New project: AOP-derived_causal_evidence_use_cases


gingin77:
An AOP Key Event Relationship is a suggestion about a causal step, not a finding about one. We checked the first one against primary literature and the causal claim it makes turns out to be **false in the healthy heart** — true only for a particular enzyme isoform, in a particular subcellular compartment, or once the heart is already diseased. The reason it fails generalises, so this proposes a project to collect these checks as worked examples rather than leaving each one buried in the issue that prompted it.

The first use case is complete and is written up in full below.

## Why this needs a home

Work on #10368 uses AOP-Wiki Key Events as curation leads. AOP-Wiki is not a citable source in dismech's validation stack — there is no `AOP:` prefix in `references_cache/`, no fetcher, and no cacheable body for a snippet to match — so every causal edge taken from it has to be re-evidenced from primary literature before it can become a pathophysiology node and a `downstream` edge.

Doing that verification produces a specific, reusable kind of analysis: what the KER asserts, what AOP-Wiki can actually back it with, what the primary literature says, and which conditions the KER left out. That analysis is worth keeping. Right now there is nowhere for it to go.

Proposed project name: **AOP-derived_causal_evidence_use_cases**. Following the existing convention in `projects/`, the file would be `projects/AOP_DERIVED_CAUSAL_EVIDENCE_USE_CASES.md`.

## First use case: KER3445

**What the KER asserts.** `Phosphodiesterase inhibition` → `Increased, cyclic adenosine monophosphate`. One adjacent edge in AOP 558 (*Phosphodiesterase inhibition leading to heart failure*). No isoform, no compartment, no disease state.

**What AOP-Wiki can back it with: nothing.** The KER record carries zero references, no weight-of-evidence text, and no empirical-support text. That is not a parsing artefact — 738 of the 2,369 KERs in the 08-06-2026 snapshot do carry references and 704 carry empirical support. This one is genuinely empty.

Its parent AOP has never been externally reviewed (`oecd_status: ""`), while scoring 100% on `completion_score`. That score counts populated fields on the wiki record; it says the page is filled in, not that the science is sound. AOP 558's own bibliography leans on Movsesian's work on PDE3 inhibitors in heart failure — which, as below, is the isoform for which the naive claim is least true in a normal heart.

### The primary evidence

Two studies, both adult rat ventricular myocytes with FRET-based cAMP biosensors.

[...]

Issue #10368, opened 2026-09-01
Curate the AOP Key Events that have no dismech counterpart


gingin77:
Two pilot mappings in [`projects/AOP_EMOD_ALIGNMENT.md`](https://github.com/monarch-initiative/dismech/blob/main/projects/AOP_EMOD_ALIGNMENT.md) checked 42 AOP-Wiki Key Events — the 21 in five cardiac lead AOPs, the 14 in three neurodevelopmental lead AOPs, and the 7 in AOP 38 (liver fibrosis) — against dismech modules and entries. Most map. **The ones that do not are a list of specific mechanisms the knowledge base does not currently represent**, and this issue turns that list into curation work.

Two of the misses already have issues (#10298 for protein alkylation, #10310 for the missing liver fibrosis node) and three were closed by the NMDA hypofunction module (#9976). What follows is what is left.

### 1. Chemical entry points for cardiac electrophysiology — 5 Key Events

| KE | Event |
|---|---|
| 1529 | Blockade, L-Type Calcium Channels (MIE) |
| 593 | Inhibition, ERG voltage-gated potassium channel (MIE) |
| 1562 | Decreased Na/K ATPase activity (MIE) |
| 2288 | Phosphodiesterase inhibition (MIE) |
| 2290 | Inhibition of Funny current (If) (MIE) |

Everything downstream of these five already maps cleanly — ten of the 21 cardiac Key Events find a module node, and the altered-action-potential and arrhythmia chain is well covered. The five initiating events fail together for one reason: **`cardiac_ion_channel_repolarization` starts with a mutation, and these start with a chemical.** Its first node is `Cardiac Ion-Channel or Calcium-Handling Variant`, "a pathogenic germline variant alters a cardiac ion channel", and the module scopes itself to inherited arrhythmia syndromes in structurally normal hearts. A channel blocked by a drug produces the same altered action potential as a channel broken by a variant, but it has nothing to attach to.

This is not a property of the module layer. `cardiomyopathy_maladaptive_remodeling` opens at `Primary Cardiomyocyte Insult` and names toxic stress explicitly; `drug_induced_nephrotoxicity`, `drug_induced_liver_injury`, and `diabetic_vascular_complications` all open where an AOP would. It is this one module.

`CLAUDE.md` already records that an acquired drug-induced long-QT module would be a **new** module rather than a second use of `cardiac_ion_channel_repolarization`, so that is the likely shape of the work — but the scoping call belongs to whoever picks this up.

[...]

PR #10464, opened 2026-09-02
Add the aop-wiki and mie-ker-capture skills

### LINKED ISSUE #10456: AOP-derived content review: add the aop-wiki and mie-ker-capture skills
Two skills that turn AOP-Wiki content into verified dismech causal edges are written and working locally, and should move into `.claude/skills/` so they are shared repository content rather than one machine's setup. They are held back on one external dependency, described below.

## The two skills

**`aop-wiki`** — how to reach AOP-Wiki data. Which command answers which question, the date footgun (three formats coexist, and omitting `--date` silently means today, which downloads a ~50 MB XML export), the collected entity shapes, and the two limits that matter for curation: AOP-Wiki is not citable in dismech's validation stack, and its ontology terms are leads to re-resolve rather than bindings to copy.

**`mie-ker-capture`** — what to do with that data. An eight-step procedure from a list of Key Event IDs to a set of verified causal edges: the lookup, relevance filtering, chain assembly, evidence triage, and literature verification. Most of its content is dismech curation policy rather than tool mechanics, and two of its checks came out of real findings on #10368:

- **The 29.41% stub gate.** A Key Event Relationship whose `completion_score.percent` is exactly 29.41 is an empty record — endpoints and identifiers, nothing else. In the 08-06-2026 snapshot, 1,536 of 2,369 KERs sit at that value and not one carries a reference. It is also the corpus median. Reading one number, before anything else, tells you whether AOP-Wiki is contributing evidence or only a hypothesis, which is what should drive how the verification work is prioritised.
- **Mapping an AOP's citations onto its own steps.** An AOP carries a bibliography even when its KERs carry nothing, and splitting it up reveals whether the author evidenced the step in question at all. AOP 558 has ten citations and none measures cAMP after phosphodiesterase inhibition — the edge the AOP is named for. One of them argues for a competing mechanism for the following edge.

Neither check requires reading a paper, and together they would have predicted the KER3445 outcome recorded in #10454.

## Why they can move now

Until recently these skills documented a clone on one developer's machine, which is why they were kept untracked behind `.git/info/exclude` rather than committed. That is no longer necessary: `aop_wiki_cli` has been packaged as an installable module with an `aop-wiki-cli` console script, so the tool can be run with `uvx`, from an installed environment, or from a clone, and it resolves its data directory at call time instead of assuming the repository root.

Both skills have been rewritten against that interface. No machine-specific paths remain in either.

## Blocked on

The packaging work is in **gingin77/aop_wiki_cli#6**, which is open and unreviewed. Until it merges, the `uvx --from git+…` install path the skills document resolves to the pre-package layout and fails.

The skills currently carry a note saying so and telling the reader to work from a clone in the meantime. That note is written to be deleted, and deleting it is part of this work rather than a follow-up.

## Plan

- [x] `gingin77/aop_wiki_cli#6` merges
- [ ] Delete the pending-merge caveat from the `aop-wiki` skill
- [ ] Verify `uvx --from git+https://github.com/gingin77/aop_wiki_cli aop-wiki-cli --help` against the merged `main`
- [ ] Remove both `.claude/skills/` entries from `.git/info/exclude`
- [ ] One PR adding both skills, since `mie-ker-capture` depends on `aop-wiki` and reviewing them apart splits one coherent addition

## Open question for review

`search-with-config` resolves a config with `importlib.import_module('aop_wiki_cli.configs.<name>')`, so it only sees modules inside the installed package — there is no data-directory search path and no `--config-file`. Running a shipped config works from an install, but writing a new one needs an editable clone and a PR to the CLI repo. The skill documents this as it stands. Worth deciding whether that is the intended shape before the skill describes it to everyone.

Related: #10368, #10454.

gingin77:
Adds two skills for working with AOP-Wiki content: one for reaching the data, one for deciding what to do with it. Both have been in use locally while working #10368 and are only now committable, because the tool they depend on became installable.

Closes #10456.

## `aop-wiki` — reaching the data

Which command answers which question, and the traps around them. The one most likely to cost someone an afternoon: three date formats coexist in the tool, and omitting `--date` silently means today, which is a cache miss on any day you have not already pulled a snapshot — and a miss downloads the ~50 MB XML export and parses it.

It also documents the data directory. The tool resolves every input, output, cache and log under one directory chosen at call time (`--data-dir`, then `$AOP_WIKI_CLI_DATA_DIR`, then the working directory). The last of those is worth knowing about here: run a command from a dismech checkout with nothing set and it creates `outputs/`, `xml_inputs/` and `logs/` inside this repository. The skill says to always pin the directory and never point it at a dismech checkout.

Two limits for anyone curating from this data are stated up front: AOP-Wiki cannot be an evidence `reference:` — there is no `AOP:` prefix in `references_cache/` and no fetcher — and AOP ontology terms are leads to re-resolve under `dismech-terms`, not bindings to copy across.

## `mie-ker-capture` — what to do with it

An eight-step procedure from a list of Key Event IDs to a set of verified causal edges. Most of it is curation discipline rather than tool mechanics, and two checks in it came out of real findings while working #10368. Both are cheap, and both belong before any literature is read rather than after:

**A single number identifies an empty record.** A Key Event Relationship whose `completion_score.percent` is exactly 29.41 is a stub — endpoints and identifiers, nothing else. In the 08-06-2026 snapshot, 1,536 of 2,369 KERs sit at that value and not one carries a reference. It is also the corpus median, so most causal edges in AOP-Wiki are assertions with nothing behind them. That does not disqualify an edge, but it does say AOP-Wiki is contributing a hypothesis and no evidence, which is what should drive how the work is prioritised.

**An AOP can carry a bibliography without evidencing its own steps.** Splitting a parent AOP's references and assigning each to the step it serves shows whether the author supported the edge in question. AOP 558 has ten citations and none of them measures cAMP after phosphodiesterase inhibition — the edge the AOP is named for. One of the ten argues for a competing mechanism for the following edge.

The worked example behind both checks is recorded in #10454.

## Why these can be committed now

They previously documented a clone on one developer's machine, which is why they sat untracked behind `.git/info/exclude`. `aop_wiki_cli` has since been packaged as an installable module with an `aop-wiki-cli` console script (gingin77/aop_wiki_cli#6, merged), so it runs via `uvx`, from an installed environment, or from a clone, and resolves its data directory at call time instead of assuming a repository root.

Both skills are rewritten against that interface. No machine-specific paths remain in either, and the two `.git/info/exclude` entries are gone.

## Verification

- `uvx --from git+https://github.com/gingin77/aop_wiki_cli aop-wiki-cli --help` runs against merged `main` with no clone present, and `--data-dir` appears as the global option the skills describe.
- All eight commands the tool exposes are documented, and the three lookup commands that were never merged upstream (`find-entities`, `show-entity`, `show-aop-graph`) appear only as explicit "does not exist" warnings.
- Numbers in the skills were checked against the 08-06-2026 snapshot rather than carried over: 1,598 events, 596 AOPs, 450 with an empty `oecd_status`, 2,369 KERs.

## One thing worth a reviewer's opinion

`search-with-config` resolves its argument with `importlib.import_module('aop_wiki_cli.configs.<name>')`, so it only sees config modules inside the installed package — there is no data-directory search path and no `--config-file`. Running a shipped config works from an install; writing a new one needs an editable clone and a PR to the CLI repo. The skill documents this as it currently stands, but if that is a gap rather than the intended design, better to know before the skill teaches it.

https://claude.ai/code/session_01TPUAQ6o7PTWZXfcGU1ckrd

### DIFF
--- a/.claude/skills/aop-wiki/SKILL.md
+++ b/.claude/skills/aop-wiki/SKILL.md
@@ -125,9 +125,10 @@ for curation based on their relevance to a particular disease or module entity
 that is being curated. `KE1562` (Decreased Na/K ATPase activity) matches six
 KERs — four leading out of it, two into it — and of the four, only KER3444 (to
 increased intracellular sodium, AOP 556) belongs to a cardiac chain. The rest
-run to growth inhibition, renal proximal tubular transport, and sodium uptake
-in fish gills. Note that the gill KER sits inside an AOP titled "leads to Heart
-failure", so check the downstream event and not just the AOP title.
+run to cell membrane depolarization, renal proximal tubular transport, and
+sodium uptake in fish gills. Note that the gill KER sits inside an AOP titled
+"leads to Heart failure", so check the downstream event and not just the AOP
+title.
 
 ### Date handling — the main footgun
 
@@ -256,7 +257,10 @@ exclusion list), `methods_nams` (mode 1, NAM assay methods),
 **Two shipped configs are not usable by `search-with-config`.**
 `event_first_collections.py` defines `DEPRESSION_CONFIG`/`PARKINSON_CONFIG`
 dicts and `reference_search.py` defines neither required name, so neither
-exposes `SEARCH_PARAMS`. Both still appear in the shipped-config listing.
+exposes `SEARCH_PARAMS`. `_available_search_configs()` appends a module only when
+it has **both** `SEARCH_PARAMS` and `OUTPUT_CONFIG`, so neither ever appears in
+the `Available configs:` listing — `configs/` is the only place you meet them,
+which is why this warning is here.
 `configs/Configs_README.md` documents a `collect-entities-for-events` command
 for them; that command **does not exist** in `cli.py`. Use `search_mode:
 event_to_aop`, or read those event ID lists directly.
--- a/.claude/skills/ker-evidence-triage/SKILL.md
+++ b/.claude/skills/ker-evidence-triage/SKILL.md
@@ -56,10 +56,19 @@ Three traps live here.
 
 **`completion_score` is not a quality score.** It counts populated fields on the
 wiki record. AOPs 552, 555, 556, 558 and 560 score 94–100% complete, have never
-been OECD reviewed (`oecd_status: ""`), and every one of their KERs examined in
-the cardiac sweep worked through in `mie-ker-capture` carries zero references.
-High completeness on an unreviewed pathway is the most misleading state in the
-data.
+been OECD reviewed (`oecd_status: ""`), and every one of their KERs sampled in a
+cardiac Na/K-ATPase and phosphodiesterase sweep carries zero references. High
+completeness on an unreviewed pathway is the most misleading state in the data.
+
+The 29.41% value makes the same point more sharply, and provably.
+`ker_completion_score()` scores 17 fields unconditionally, so 5/17 is the only
+route to that percentage. The five are `aop_ids` plus the four evidence blocks —
+which the parser always builds as three-key dicts, and which `_has_content()`
+scores on `len(dict) > 0`, so a wholly empty evidence block still counts as
+populated. `references` is a separate scored field, so **a KER carrying any
+reference scores at least 35.29%**. "Zero of them carry a single reference" is
+therefore guaranteed by the scoring function, not merely observed in this
+snapshot.
 
 **An empty `oecd_status` is not a rejection.** 450 of 596 AOPs are empty. It means
 no review has happened. Endorsement raises confidence in a pathway; its absence


gingin77:
Please hold this one — it is approved, but the content has changed enough since the review that it should not merge as it stands. I have assigned it to myself so the automatic merge sweep skips it.

Reviewing the skills line by line turned up a factual error and a structural gap, and the fixes are still in progress locally:

**A wrong claim about Key Events.** Both skills said a Key Event is "organ-agnostic by design", used to explain why a KER lookup returns edges from unrelated organs. That is not right — Key Event records do carry organ context (`cell_term`/`organ_term`), along with taxon, sex, and life stage. The specificity comes from the *partner* Key Event in the relationship, not from the Key Event being indifferent to anatomy. Both files have been reworded.

**No stated curation target.** The procedure told you to discard irrelevant edges without ever asking what they should be relevant *to*. A worked example about cardiac disease was carrying that job implicitly, which made the skill read as though it only applied to cardiac work. Step 1 now records the disease or module entity the sweep is for, and the example table asks "keep, for the disease or module entity?" rather than "keep, for cardiac work?".

**Step 5 is becoming its own skill.** The evidence triage step was a third of `mie-ker-capture` and only connected to the rest of it by the phrase "for each surviving KER". It answers a question that stands alone — how much does AOP-Wiki actually back this causal edge? — and that question also comes up when auditing an AOP directly, or when checking an existing dismech edge that came from AOP-Wiki. It is now `ker-evidence-triage`, with `mie-ker-capture` keeping a short Step 5 that states the gate and delegates the method.

I will push these and ask for a fresh review. The push will drop the current approval on its own, but the assignment is the part that keeps the sweep off it in the meantime.

gingin77:
Pushed the rewrite and released the hold — this is ready for review again. The branch has been rebased onto current `main`, so the three commit hashes changed.

What changed since the earlier approval:

**A Key Event is not organ-agnostic.** Both skills had explained cross-organ lookup results by saying a Key Event is "organ-agnostic by design". Key Event records carry cell and organ terms, taxonomy, sex and life stage, so that was simply wrong. The specificity comes from the partner Key Event in the relationship. `aop-wiki` now says so directly; `mie-ker-capture` drops the explanation and lets the worked example demonstrate the filter.

**The curation target is now an input.** The procedure told you to discard irrelevant edges without ever asking what they should be relevant to — a cardiac example was carrying that implicitly, which made the skill read as cardiac-specific. Step 1 records the disease or module entity, and that phrase is used consistently throughout.

**Chain assembly no longer groups by AOP.** Grouping by parent AOP splits one mechanism across many buckets and counts the same edge repeatedly. Measured on a five-Key-Event set: 28 relationships formed two connected groups matching the biology, while grouping by AOP scattered one 24-edge group across 19 AOPs and listed a single edge in ten of them. Step 6 now uses connected components and treats AOP identity as provenance. Follow-up for the project page is #10687.

**The already-represented check tested the wrong thing.** It searched only `kb/modules/`, but 60% of disorder entries declare no `conforms_to` at all and only 17% of pathophysiology nodes have one, so the check missed the common case. It now looks at the Step 1 entity's own nodes first, asks about modules separately, and splits into three outcomes — one of which genuinely spares later verification work. Step 8 gained a matching row so an already-covered edge is not re-nominated as a gap later.

**Evidence triage is now its own skill.** That step was a third of `mie-ker-capture` and connected to it only by one phrase. It answers a question that stands alone — how much does AOP-Wiki actually back this causal step — which also comes up when auditing a pathway directly. It is now `ker-evidence-triage`; `mie-ker-capture` keeps a short step stating the gate and delegating the method.

**A scope note** records that this procedure is written for Molecular Initiating Events, that the lookup and most steps work in either direction, and that only two assumptions in Step 6 would need revisiting for a pathway-terminating Adverse Outcome or an intermediate Key Event. It says not to generalize those without a real use case.

Every statistic quoted above was recomputed from the cached AOP-Wiki snapshot and the knowledge base rather than carried over, and the shell commands in the skills were run before being committed.

Related: #10456, #10368.

gingin77:
Both blocking items are fixed in 568fb628ea. I checked each against the `aop_wiki_cli` source before changing anything, and both were correct.

**The config-listing sentence was backwards.** `_available_search_configs()` adds a module to the list only when it defines both `SEARCH_PARAMS` and `OUTPUT_CONFIG`, and nothing else produces that listing — so the two modules in question never appear in it. The skill claimed they do. The sentence now states the inverse and explains why the warning is worth having: the `configs/` directory is the only place a reader encounters them.

**An AOP title was being presented as a downstream event.** KER2787 ends at "Increase, cell membrane depolarization"; growth inhibition is the title of AOP 266. You were right that this was the worst place for that mistake, since the next sentence tells the reader to check the downstream event rather than the AOP title — and the other skill in this PR already had it right, so the two disagreed.

**On the 29.41% point — I took your suggestion and it is a real improvement.** I verified it independently: `ker_completion_score()` scores 17 fields unconditionally, the five that a stub can reach are `aop_ids` plus the four evidence blocks, and `_has_content()` counts an empty block as populated because it tests `len(dict) > 0`. Since `references` is scored separately, any relationship carrying a citation lands at 35.29% or above. So "zero of them carry a single reference" follows from the scoring function and is not a property of this snapshot. That is now in the skill as a short proof, and it replaced the back-reference to a cardiac sweep that this branch had moved into the new skill — thank you for catching that too.

**One suggestion I did not take.** Replacing `python3 -c` with `uv run python` looks right for this repository, but run from a git worktree it creates a fresh `.venv` as a side effect. I hit that while testing it. A documentation example should not leave a virtual environment behind, and `python3 -c` works in both a disorder file and a module file, so I left it.

**Three notes I have left for the author to decide**, since they involve wording choices rather than errors: "one edge out of six" against a table showing four rows, the prose fixing the example as cardiac while the column header is now generic, and the `09-03-2026` date format. The date was a specific request, and the other two are deliberate — the concrete example is what makes the generic column readable.

PR #10465, opened 2026-09-02
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 50 (MONDO:0013114, MIR96)

### LINKED ISSUE #10461: Curate autosomal dominant nonsyndromic hearing loss 50 (MONDO:0013114)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 50 (DFNA50)** ([MONDO:0013114](https://monarchinitiative.org/MONDO:0013114)) — *MIR96* (`hgnc:31648`) seed-region point mutations. This is not a protein-coding disease: the causal lesion sits in the mature miR-96 seed, so one substitution simultaneously de-represses the wrong targets and abandons the right ones.

Nothing in `kb/` carries this MONDO term, and no entry names MIR96 as a causal gene, so this is a clean start.

**Two things make it worth doing carefully.**

The **mechanism class is new to the KB**. A miRNA seed mutation is neither loss nor gain of a protein function, so think about what `modifier` and `functional_impact_category` can honestly say before reaching for the nearest value — a recent entry correctly left a `modifier` absent because `ModifierEnum` had no value for the state in question, and that is better than forcing a wrong one. The gene is also not a protein, so check what `GeneDescriptor`/HGNC binding looks like for a microRNA gene rather than assuming the protein-coding pattern transfers.

There is a **real cross-entry link already in the repo**. `kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103.yaml` (CLIC5, DFNB103) curates the finding that *CLIC5 is directly regulated by miR-96 and miR-182*, with the 3'-UTR target site quoted. So a miR-96 seed mutation plausibly acts partly through a target this KB already curates as its own deafness gene. Read that entry, and if the evidence supports it, make the relationship explicit rather than leaving two unconnected entries — but do not overstate it: regulation of CLIC5 by miR-96 is shown, DFNA50 acting *through* CLIC5 is a further claim needing its own evidence.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_50.yaml`

kevinschaper:
Adds DFNA50, a dominant progressive hearing loss caused by point mutations in *MIR96*. The gene makes no protein — it makes a 22-nucleotide microRNA — and the whole disease turns on a single nucleotide inside the six-base "seed" that decides which messenger RNAs the microRNA switches off. Changing one seed base does two things at once: the RNAs miR-96 was supposed to silence are released, and a different set of RNAs it never touched starts getting silenced instead. That is not a protein losing its function, and it is not a protein gaining one, which is why some of the usual controlled fields had to be left empty.

`Closes #10461`

## Scope decision: `DISEASE`

One gene, one mechanism chain running from the mutation to progressive sensorineural hearing loss, and no member diseases to unite — so not a grouping. Not a subtype either: its only MONDO parent, `MONDO:0019587` autosomal dominant nonsyndromic hearing loss, is an ontology bucket holding dozens of unrelated DFNA loci and is not itself a dismech entry. The stub is deleted in this PR.

## Two fields left empty on purpose

Both have a `notes:` block beside them giving the reason.

**`functional_impact_category` on the seed-mutation genetic context.** The enum lets you pick one value. `LOSS_OF_FUNCTION` is ruled out by the mouse: delete the microRNA outright and heterozygous mice hear perfectly well, so losing half the normal miR-96 is not what causes the disease (PMID:39434156). `NEOMORPHIC` describes the newly acquired targets but silently drops the abandoned ones, which is the half the original papers argued was primary. Writing either would put a machine-readable claim next to prose that contradicts it.

The Italian family's allele is different and *does* get a value. It sits outside the mature microRNA, destabilises the precursor hairpin, and simply reduces how much mature miR-96 gets made — a purely quantitative defect, so `PARTIAL_LOSS_OF_FUNCTION` fits. The two allele classes are curated as separate pathophysiology nodes for exactly this reason.

**`modifier` on the novel-target node's `GO:0035195` descriptor.** What changed is *which* transcripts get silenced, not how much silencing happens. `INCREASED`/`DECREASED` are the PATO-bound quantitative pair and would assert a level change; these transcripts had no miR-96 silencing to increase. `GAIN_OF_FUNCTION` is defined in this schema as a process driven outside its normal regulatory constraints, which is activation of the same process, not redirection onto different substrates. `DYSREGULATED` is closest but still reads as a level claim.

The sibling node — loss of repression of the *normal* target set — does carry `modifier: DECREASED`, because there the claim genuinely is quantitative.

## The CLIC5 link, and why it is not a shared mechanism

`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103` already curates CLIC5 as directly regulated by miR-96. The tempting conclusion is that DFNA50 acts through CLIC5. I do not think that follows, and the directions are the reason:

- miR-96 **represses** CLIC5, through a conserved site at nucleotides 760–766 of the CLIC5 3′-UTR (PMID:22889583).
- Losing miR-96 function therefore **raises** CLIC5 — and `Clic5` is indeed among the transcripts that go up in the cochlea of mice with the miR-183/96/182 cluster knocked out (PMID:30575790).
- DFNB103 is caused by biallelic **loss** of CLIC5.

So at this node the two diseases look inverse, not convergent. What is genuinely unknown is whether CLIC5 moves at all in a seed mutant, since a seed mutation is not a null and the mutant cochlea contains plenty of genes that fall without carrying any miR-96 site. Recorded as a `notes:` block on the loss-of-repression node plus a `KNOWLEDGE_GAP` discussion (`mir96_clic5_direction`) naming GSE255796 as the dataset that could settle it.

## Mouse models

Four lines in `animal_models:` with `modeled_mechanisms`, including two negative links:

| Model | Link |
|---|---|
| *diminuendo* (ENU seed mutant) | `RECAPITULATES` novel targeting and maturation arrest |
| `Mir96+14C>A` knock-in | `RECAPITULATES` bundle failure and the clinical endpoint; also carries the `RESCUES` link |
| `Mir96+13G>A` knock-in | `FAILS_TO_RECAPITULATE` — heterozygotes hear normally to one year while the same human allele causes deafness |
| `Mir183/96` double knockout | `FAILS_TO_RECAPITULATE` novel targeting — a null cannot acquire targets, which is the point |

The `+13G>A` mismatch gets its own `HUMAN_MODEL_MISMATCH` discussion. Both knock-ins were made in the same lab on the same background, so the divergence is a real result rather than a technical artefact, and it means a mechanism shown in one allele does not transfer to the other by default.

The `RESCUES` link on the `+14C>A` line carries both preclinical arms — AAV-SaCas9-KKH editing of the mutant allele, and oral amitriptyline — with the authors' own "this is a proof of concept rather than a suggestion that amitriptyline be used as a treatment" quoted on the link so it cannot be read as a therapy claim. Threshold readouts use `INCREASED` for the mutants, since a higher ABR threshold means worse hearing.

## What is deliberately not here

- **No `treatments:` block.** Nothing disease-modifying is approved or in trial; the two interventions in the literature are mouse work and live on the model link. Importing generic sensorineural-hearing-loss management would attach another disease's evidence to this one. Reasoning is in the `dfna50_no_disease_modifying_therapy` discussion, attached to the empty `treatments#` section.
- **No biochemical, histopathology, imaging or clinical trials.** None published for these three families.
- **No audiometric natural history per allele, no vertigo, no penetrance figures.** Those are in the Soldà 2012 full text, which is not retrievable — PubMed and Europe PMC both return the abstract only, and `DOI:10.1093/hmg/ddr493` fetches as `content_type: unavailable`. I did not take the numbers second-hand from a review.

## Deep research

One falcon report, included. `just preflight-dr` against `MONDO:0013114` **passed**: MIR96 mentioned 52 times, nothing competing (next was the acronym ABR at 5).

The report proposed three HPO terms that do not mean what it said they mean:

| Report's label | What the ID actually is |
|---|---|
| `HP:0008527` "Postlingual sensorineural hearing impairment" | **Congenital** sensorineural hearing impairment |
| `HP:0008607` "Abnormality of the auditory threshold" | Progressive **conductive** hearing impairment |
| `HP:0000404` "Sensorineural hearing impairment, severe" | Does not resolve at all |

The report's own term-validation pass independently flagged `HP:0000404` as nonexistent. Every term in the entry was resolved through OLS before use.

## Validation

All exit 0: `validate`, `validate-terms`, `count-verified-snippets` (70/70), `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history`, `verify-datasets` (2/2 OK).

### DIFF
--- a/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_50.yaml
+++ b/kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_50.yaml
@@ -1233,6 +1233,11 @@ datasets:
     Transcriptomes from miR-96 knockout and viral miR-96 restoration experiments. Relevant
     to DFNA50 as the loss-of-function comparator rather than as a model of the seed
     alleles - the same distinction the Mir183/96 null model carries.
+references:
+- reference: PMID:20301607
+  title: Genetic Hearing Loss Overview.
+  tags:
+  - GeneReviews
 discussions:
 - discussion_id: mir96_clic5_direction
   kind: KNOWLEDGE_GAP


kevinschaper:
Added the `references:` block — `PMID:20301607` "Genetic Hearing Loss Overview." tagged `GeneReviews`, matching DFNA41 and DFNA3A. Pushed in `71e5fd4`.

I checked the scope point before adding it: a PubMed search restricted to `GeneReviews[Book]` for MIR96 or DFNA50 returns exactly one record, the Overview, so there is no gene-specific chapter to tag. The cached Overview is abstract-only, so no evidence items were manufactured from it.

No change on the two suggestions, since both said none was needed:

- The phenotype detail from the deep-research report (high-frequency predominance, vertigo, onset around 25–40 years, incomplete penetrance for the Italian allele) all traces to the Soldà 2012 full text, which is not fetchable. Left out rather than taken second-hand, as documented in the entry `notes:`.
- `modifier` stays off the novel-target node.

`just validate`, `check-duplicate-keys` and `check-entity-refs` all pass on the updated file.

PR #10466, opened 2026-09-02
curate: Add Immunodeficiency 18 (MONDO:0014278, CD3E)

### LINKED ISSUE #10462: Curate immunodeficiency 18 (MONDO:0014278)
Curate a dismech entry for **immunodeficiency 18** ([MONDO:0014278](https://monarchinitiative.org/MONDO:0014278)) — biallelic *CD3E* variants; loss of the CD3ε chain prevents T-cell receptor assembly and surface expression, giving a T-cell immunodeficiency that ranges from severe combined immunodeficiency to a milder, later-presenting course.

**Read `kb/disorders/Combined_Immunodeficiency_Due_To_CD3gamma_Deficiency.yaml` first** — it was curated recently, it is the sibling chain of the same receptor complex, and it is the only place CD3E currently appears in `kb/` (as a differential, not as coverage). That entry already worked through how to keep the CD3δ/ε/ζ literature apart from the CD3γ entity; here the polarity is reversed, and the older "CD3 deficiency" papers that entry excluded may be exactly the ones this entry needs. Make the two agree on where the boundary sits, and reuse its structure where the biology is genuinely parallel rather than reinventing it.

That entry also hit a schema gap worth knowing about: `TreatmentEffectEnum` has no counterpart to `EnvironmentalEffectEnum`'s `EXACERBATES`, so a treatment that worsens its target node cannot be encoded (#10004).

Stub: `stubs/Immunodeficiency_18.yaml`

kevinschaper:
New disease entry for immunodeficiency 18 — biallelic *CD3E* loss of function. Deletes `stubs/Immunodeficiency_18.yaml`.

## entry_type: DISEASE

One conserved pathograph, anchored on one gene: loss of the CD3epsilon chain → the pre-TCR and TCR cannot be assembled → beta-selection cannot be signalled → thymopoiesis arrests. Not a grouping (one gene, one mechanism), not a subtype (no curated parent it belongs under; MONDO:0015703, the three-gene T-B+ SCID concept, is a grouping we do not have and is curated here as a differential), and clearly in scope.

## The severity spectrum is modelled, not flattened

CD3E deficiency is not uniformly severe combined immunodeficiency, and the entry does not say it is. The graph branches at the variant node:

- **null branch** — complete loss of the chain → failed complex assembly → blocked pre-TCR signalling → thymocyte arrest → absent peripheral T cells, splitting into cellular-immunity failure and loss of T-cell help to B cells.
- **hypomorphic branch** — residual unstable chain → reduced surface TCR/CD3 with T cells present and unresponsive.

Both converge on failure of cellular immunity. The hypomorphic branch is the family in which CD3E was first implicated (PMID:1370449, PMID:8490660): T cells present, surface receptor at a tenth of normal, no proliferative response through CD3 or CD2.

## Boundary with the CD3gamma entry

`Combined_Immunodeficiency_Due_To_CD3gamma_Deficiency.yaml` was read and is **not modified**. It confined the CD3delta/epsilon/zeta literature to its differential-diagnoses section to protect the CD3gamma entity; here that literature is primary, and the two entries cite the same sentence from PMID:16264327 from opposite sides — homozygous CD3D/CD3E mutations block T-cell development, in contrast with the partial immunodeficiency of CD3G. Nothing here contradicts it.

What this entry adds, which the CD3gamma entry had no reason to consider: the hypomorphic end of the CD3E spectrum also shows reduced-not-absent surface receptor with T cells present, which is the one presentation that looks like CD3gamma deficiency on a flow panel. The entry says plainly that there the two are separated by the gene, not by the immunophenotype. Structure was reused where the biology is parallel (the differential-diagnosis framing, the mouse-model treatment, the evidence-grading note).

## Knowledge gaps

- **HUMAN_MODEL_MISMATCH** — every account of this disorder places the block at beta-selection, and every one traces back to the mouse. The authors who identified CD3epsilon deficiency said explicitly that no thymus material was available, so the human stage was never measured. The neighbouring chain is the precedent for the mouse being wrong about a human CD3 chain.
- **KNOWLEDGE_GAP** — whether residual CD3epsilon protein reliably predicts a milder course. One hypomorphic family against a handful of null families; nobody has measured residual protein across a series.

## Deep research

Falcon report committed. `just preflight-dr` PASS against MONDO:0014278 (CD3E the top gene, 77 mentions). Treated as leads only — every snippet was re-verified against its cached reference, and two leads were dropped because their citation would not resolve.

## Validation

All exit 0: `just validate` (schema, terms, **64/64 snippets verified**), `validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `validate-history`.

One term error caught in drafting: `NCIT:C16360` was drafted for newborn screening and resolves to *Book*; corrected to `NCIT:C81178`.

Closes #10462


kevinschaper:
One follow-up commit since the PR was opened, prompted by the automated second opinion on #10462.

**Recorded why two CD3E case reports are cited nowhere.** PMID:32016651 and PMID:33655388 are the two most recent CD3E reports and both are journal letters with no abstract in PubMed, Europe PMC or PMC. Fetching either yields a cache with no quotable body, so neither can carry an evidence item without breaking the exact-quote rule, and a title is not a finding. Their clinical content is represented through PMID:15546002 and PMID:28597365 instead. That is now stated in the entry's notes rather than left looking like an oversight.

**IUIS release is 2022 on purpose.** The 2024 update exists. I kept 2022 because the CD3gamma entry uses it, and part of the value of the classification here is that the two adjacent CD3 entries can be read against one row each of the same table. Also noted in the entry.

**Not done here, deliberately:** adding this entry and the CD3gamma entry to `Combined_Immunodeficiencies_IEIs.yaml` and `Inborn_Errors_of_Immunity.yaml`. It is a real gap, but it touches shared files and one of the two entries is not this PR's, so it belongs in its own change. Same for the sibling stubs `Immunodeficiency_19` (CD3D) and `Immunodeficiency_25` (CD247) and a possible TCR/CD3 grouping — PMID:15546002 does cover CD3D and CD3E in one study, so most of the reading is done, but that is a separate curation.

64/64 snippets verified; schema, term, reference, duplicate-key, entity-ref, causal-target, snippet-length, title-snippet, snippet-grading, folded-hyphen and environmental-evidence checks all pass.

PR #10469, opened 2026-09-02
curate: Add Immunodeficiency 35 (MONDO:0012682, TYK2)

### LINKED ISSUE #10463: Curate immunodeficiency 35 (MONDO:0012682)
Curate a dismech entry for **immunodeficiency 35** ([MONDO:0012682](https://monarchinitiative.org/MONDO:0012682)) — biallelic *TYK2* loss of function; tyrosine kinase 2 transduces IL-12, IL-23, type I interferon and IL-10 signalling, so complete deficiency gives susceptibility to mycobacteria and viruses together.

**The retrieval trap here is dose, not identity.** TYK2 already appears in `Ankylosing_Spondylitis.yaml`, `Dermatomyositis.yaml`, `Down_syndrome.yaml` and `Interleukin-10_Receptor_Deficiency.yaml` — none of them this disease, and mostly as a pathway or GWAS gene. Beyond that, the common hypomorphic **P1104A** allele is one of the best-studied variants in tuberculosis susceptibility genetics, and the TYK2 inhibitor literature (deucravacitinib and others) is large and growing. None of that is evidence about complete TYK2 deficiency. Gene-frequency preflight will pass and tell you nothing; check each citation is about biallelic complete loss in a patient.

If the boundary between complete deficiency and the P1104A hypomorph is genuinely blurred in the literature, curate that as a `discussions:` `KNOWLEDGE_GAP` rather than resolving it by assertion.

Stub: `stubs/Immunodeficiency_35.yaml`

kevinschaper:
New disease entry for immunodeficiency 35 — complete autosomal recessive TYK2 deficiency. Deletes `stubs/Immunodeficiency_35.yaml`.

## entry_type: DISEASE

One gene, one conserved pathograph. TYK2 sits on the receptors for type I interferon, IL-12, IL-23 and IL-10, so losing it disables four circuits at once — but the two that matter for host defence produce a coherent two-branch mechanism, and that is the entry:

- **IL-12/IL-23 branch** → deficient interferon-gamma induction in lymphocytes → macrophages not licensed to kill intramacrophagic pathogens → mycobacterial disease
- **type I interferon branch** → failure of cell-intrinsic antiviral defence → severe viral disease

A third branch, impaired IL-10 signalling, is curated as a node **with no downstream edge**. It is a real, reproducible, diagnostically useful part of the cellular phenotype, and it has no established clinical consequence — unlike IL-10 receptor deficiency, these patients do not get very-early-onset IBD. Curating it as a dead end says that, rather than implying a disease link the literature does not support.

## How the P1104A and inhibitor literature was kept out

This was the main risk on this disease, and gene-frequency preflight passes without discriminating, exactly as the claim issue warned. So the check was done citation by citation instead — every reference was read to confirm it reports a patient, cell or mouse carrying **biallelic complete loss**.

Three bodies of literature were screened out and not used:

- **TYK2 as an autoimmune GWAS locus.** Not evidence about a Mendelian null.
- **TYK2 as a pathway member elsewhere in `kb/`** (Ankylosing_Spondylitis, Dermatomyositis, Down_syndrome, Interleukin-10_Receptor_Deficiency). None of those is this disease and none was touched.
- **The TYK2-inhibitor literature** (deucravacitinib and successors). Deliberate partial pharmacological inhibition of a kinase in adults is not congenital complete loss, and no inhibitor paper is cited anywhere in the entry.

**P1104A is cited exactly twice**, in the only two places it belongs: as a named differential diagnosis, and inside the boundary knowledge gap. `PMID:30578352` (Boisson-Dupuis 2018) is the sole P1104A paper in the entry and is never used to support a claim about complete deficiency. The IUIS nosology backs the split — the 2022 table carries *Tyk2 deficiency* and *P1104A TYK2 homozygosity* as two separate rows with different cellular defects and different clinical phenotypes, and both rows are quoted.

## The boundary is a knowledge gap, not an assertion

The awkward fact is that PMID:36094518 showed impaired IL-23-dependent interferon-gamma induction is the **only** mechanism of mycobacterial disease shared by all five forms of AR TYK2 deficiency, P1104A included. For the mycobacterial phenotype the spectrum is mechanistically continuous.

The entry keeps the two apart on grounds that are real — complete deficiency additionally loses type I interferon, IL-12 and IL-10 signalling and the receptor-scaffolding function, and only complete deficiency causes viral disease — and then says in a `KNOWLEDGE_GAP` that the line is drawn on the extra pathways rather than on the mechanism the two share, and that no study has compared the mycobacterial disease itself once exposure is matched. It also notes that no purely genetic rule works: G1010D makes a detectable protein and counts as complete deficiency, R864C makes a hypomorph and does not.

A second gap covers incomplete penetrance (48% mycobacterial, 60% viral, with asymptomatic homozygous siblings).

## Other notes

- **Prevalence**: `CASES_IN_LITERATURE` / `ULTRA_RARE`, **no numeric rate**. The only frequency statement available ("less than 1/600,000") is an order-of-magnitude comparison against a common susceptibility allele, not a measured prevalence; it is quoted in the record and deliberately not converted.
- **A normal immunophenotype is curated as a diagnostic phenotype with no HP term bound** — the claim is the absence of an abnormality, and binding a term would assert one.
- **BCG vaccination** is linked into the pathograph via `influences_mechanisms` / `TRIGGERS`, with a `notes:` line recording that ECTO was searched and no suitable exposure term found, rather than binding a poor one.

## Deep research

Falcon report committed; it had independently scoped itself to complete deficiency and says so in its own scope note. `just preflight-dr` PASS against MONDO:0012682 (TYK2 the top gene at 97 mentions, OMIM 611521 matching). Used as leads; all snippets verified against caches.

## Validation

All exit 0: `just validate` (schema, terms, **73/73 snippets verified**), `validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `validate-history`.

Two drafting errors caught before commit: `treatment_effect: CORRECTS` does not exist in `TreatmentEffectEnum` (used `RESTORES`), and `NCIT:C68931` resolves to *Teglarinad Chloride* rather than mepolizumab (used `NCIT:C157376`).

Closes #10463


kevinschaper:
Two follow-up commits since the PR was opened, both from re-reading my own numbers and from the automated second opinion on #10463.

**Type III interferon is now an open question, not a settled one.** The entry originally said IFN-lambda responses are preserved, citing the 2015 patient series. That generalises further than the evidence does. TYK2 is constitutively associated with IL-10R2, one of the two chains of the type III interferon receptor, so the architecture predicts a defect — and the cell-line data disagree with each other: SV40-transformed patient fibroblasts respond very weakly, their EBV-B cells respond normally, and a TYK2-knockout HAP1 line responds normally. Every observation on either side is from a transformed line. That is a `HUMAN_MODEL_MISMATCH`, not a `KNOWLEDGE_GAP` — evidence exists in model systems and its translation is what is unresolved. It matters clinically because the epithelial surfaces where type III interferon does most of its work are where these patients get their respiratory viral disease, and no patient-derived epithelial system has been tested.

**Removed a frequency band that contradicted its own evidence.** Mucocutaneous candidiasis was banded `OCCASIONAL`, drafted from the fungal-disease share of one cohort (7/19). But the HP term bound is *chronic mucocutaneous candidiasis* in the strict sense, and that has been reported in exactly one TYK2-deficient patient — the 2006 index case. `OCCASIONAL` misstates the strict count and `FREQUENT` would misstate the loose one, so the band is gone and both numbers are in the description, each with its own evidence item.

**Declined the suggested module conformance, with a reason.** `innate_antiviral_interferon_response#Type I and III Interferon Induction and JAK-STAT Signaling` was proposed for the antiviral branch. That node bundles interferon *induction* (`GO:0032606`) with JAK-STAT signalling, and induction is not impaired in TYK2 deficiency — only the response to interferon already made. Declaring conformance would assert a lesion the disease does not have. Recorded in the entry's notes, along with the observation that a narrower "JAK-STAT loss of signalling downstream of a cytokine receptor" module does not exist yet and TYK2 deficiency would be a reasonable first conformer for one.

**Not done here, deliberately:** the grouping-membership gap in `Intrinsic_and_Innate_Immunity_Defect_IEIs.yaml` (which also affects `Immunodeficiency_69` and `Immunodeficiency_86`). It touches a shared file and covers entries this PR does not, so it belongs in its own change.

Now at 77/77 snippets verified; schema, term, reference, duplicate-key, entity-ref, causal-target, snippet-length, title-snippet, snippet-grading, folded-hyphen and environmental-evidence checks all pass.

Issue #10132, opened 2026-08-30
A differential diagnosis cannot name the test that resolves it — the discriminator is free text


cmungall:
Most differential diagnoses are resolved by a specific test, and dismech usually has that test curated in the same file — but there is no way to connect them. `DifferentialDiagnosis.distinguishing_features` is `multivalued: true, range: string`, so the discriminator survives only as a sentence a human has to read. Nothing can answer "what test settles this differential?"

The Fanconi anemia entry added in 2837ee824 is the worked example. Eight of its nine differentials name the same test in prose, and that test is already curated three sections away in the same file as `diagnosis#Chromosomal Breakage Test (DEB)`.

## What it looks like today

```yaml
differential_diagnoses:
- name: VACTERL/VATER association
  distinguishing_features:
  - Chromosomal breakage testing with DEB or MMC is the discriminator; VACTERL is a
    malformation constellation without a crosslinker-sensitivity phenotype.
```

and separately, in the same file:

```yaml
diagnosis:
- name: Chromosomal Breakage Test (DEB)
  diagnosis_term:
    preferred_term: chromosomal breakage analysis (DEB)
    term: {id: NCIT:C18280, label: Cytogenetic Analysis}
```

The two never meet.

## Proposal

Add `resolved_by` to `DifferentialDiagnosis`, using the existing entity-reference grammar so it is foreign-key checked the same way `attaches_to` is:

```yaml
- name: VACTERL/VATER association
  resolved_by:
  - diagnosis#Chromosomal Breakage Test (DEB)
```

`diagnosis#` is not currently in `SECTION_KEYS` in `src/dismech/entity_refs.py`; adding it is the mechanism the CLAUDE.md guidance already prescribes for a missing prefix. `biochemical#` is already resolvable and should also be a valid target.

## Two things worth designing for, not just the happy path

**A discriminator can be qualified.** The Fanconi anemia acquired-aplastic-anemia entry needed an evidence item recording that a normal lymphocyte breakage result does *not* exclude FA in a mosaic patient — you have to retest on skin fibroblasts. That is a qualification on the test's discriminating power, not on the differential itself, and once `resolved_by` is a bare reference list there is nowhere to attach it. Either `resolved_by` needs to be an object with its own `notes`/`evidence`, or the qualification needs a defined home.

[...]

PR #10473, opened 2026-09-02
Add differential diagnoses to Fanconi anemia disorder entry

cmungall:
## Summary

Added a comprehensive `differential_diagnoses` block to the Fanconi anemia disorder entry, which previously had none despite 909 other KB entries carrying one. The new block contains nine differential diagnoses organized by presentation window: four differentials live at the congenital-malformation presentation, and five live at the hematologic presentation.

## Key Changes

- **Added nine differential diagnoses** to `kb/disorders/Fanconi_Anemia.yaml`:
  - Congenital-malformation presentation: VACTERL/VATER association, Baller-Gerold syndrome, thrombocytopenia-absent radius (TAR) syndrome, Holt-Oram syndrome
  - Hematologic presentation: Diamond-Blackfan anemia, dyskeratosis congenita, Shwachman-Diamond syndrome, Nijmegen breakage syndrome, acquired aplastic anemia

- **Each differential entry includes**:
  - MONDO disease term binding with preferred term and label
  - Clinical description explaining the overlap with FA
  - Distinguishing features that separate the conditions
  - Evidence citations (13 total snippets) from three cached sources:
    - PMID:19622403 (Auerbach, "Fanconi anemia and its diagnosis")
    - PMID:25227706 (Fargo et al., NCI IBMFS breakage comparison)
    - PMID:16015582 (Faivre et al., VACTERL breakage studies)

- **Updated ontology term caches** with three new MONDO terms:
  - MONDO:0009039 (Baller-Gerold syndrome)
  - MONDO:0010121 (thrombocytopenia-absent radius syndrome)
  - MONDO:0015610 (acquired aplastic anemia)

## Implementation Details

- All evidence snippets are exact substrings verified against cached references
- Chromosomal breakage testing with DEB or MMC is identified as the discriminator across applicable differentials
- The acquired aplastic anemia differential is highlighted as the highest-stakes differential: a missed FA diagnosis in phenotypically normal patients means standard conditioning regimens containing alkylating agents and irradiation that FA patients cannot tolerate
- Validated with `just validate`, `just validate-disorders`, and pytest; all checks pass (10260 tests, 404/405 snippets verified with 1 skipped by prefix)

https://claude.ai/code/session_013PV66vARAUq1dKQoA9nriE


cmungall:
Fixed the quotation in `656f0a411`, and took one of the non-blocking suggestions.

## The quotation — you were right, and here is what it actually was

The sentence is real. It is the Table 1 footnote to the group carrying the VACTERL entries in the 2023 nosology, and I read it off the PMC page while writing the entry. What I got wrong was putting it in quote marks against a PMID whose committed cache cannot back it.

I checked whether the cache could be made to substantiate it rather than assuming it could not:

```
$ just fetch-reference PMID:36779427
  Content type: full_text_xml
  Content length: 23305 characters
$ diff -q <before> references_cache/PMID_36779427.md
  identical
$ grep -c "may be mimicked by NAD" references_cache/PMID_36779427.md
  0
```

A re-fetch returns a byte-identical file. The capture drops table content, so there is no route to a verified snippet for it. The notes now paraphrase, attribute to Unger et al. Table 1, and state plainly that the cached PMC text does not capture Table 1 or its footnotes — so a future curator who reaches for that string knows why it is not quoted.

Your second-pass correction on the revision mismatch was also right: the `isds_skeletal_category` block cites both revisions deliberately, so there was nothing to reconcile there.

## Taken: the three remaining mimics

Added, each carrying the identical already-verified Auerbach snippet:

| Differential | MONDO |
|---|---|
| Townes-Brocks syndrome | `MONDO:0007142` |
| Seckel syndrome | `MONDO:0019342` |
| Dubowitz syndrome | `MONDO:0009124` |

Twelve differentials now. The cache diff is additive — only Dubowitz was new; the other two were already present.

## Deferred, and why

**The `phenotypes` slot.** This is the right idea and I would like it done, but not here. #10132 is an open design question about whether the bindable discriminators should move out of `distinguishing_features` into a phenotype-referencing slot at all. Populating `phenotypes` now would commit to one answer before that is settled, and would likely be redone.

**Uncited Holt-Oram and TAR discriminators.** Same root cause, and it is worth naming precisely: `distinguishing_features` is a bare string list with no evidence slot of its own. Evidence attaches at the `DifferentialDiagnosis` level, so "citing a discriminator" means attaching an item that does not actually correspond to the sentence it is meant to support. That gap is what #10132 describes. The assertions are clinically correct; they are structurally uncitable at present.

**VACTERL frequency range.** The 12% figure comes from `research/`, which is a lead rather than evidence. Putting a second unverified number into `notes` would repeat the exact defect this commit fixes. It needs a fetched, snippet-verified reference, which is a separate change.

## Validation on the current head

`just validate` (407/408 snippets verified, 1 skipped by prefix), `check_enum_values`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-reference-titles`, `validate-history`. The full `tests/test_data.py` run was green on the previous head (47,358 passed) and is re-running against this one.

The `Follicular_Lymphoma` snippet-grading divergence you flagged as a one-commit-behind artifact is gone, as you expected.

A second history record covers this session.

---

PR #10483, opened 2026-09-02
curate: add Cisplatin Toxicity (MONDO:0027664)

caufieldjh:
## Cisplatin Toxicity (MONDO:0027664)

New four-arm treatment-toxicity entry.

Cisplatin crosslinks DNA in whatever cell it enters. So the question of which tissue gets hurt is not a question about the lesion. It is a question about which cells carry the platinum in, and whether those cells can be replaced afterwards.

- **Kidney** — OCT2 and Ctr1 pull it into proximal tubule, which dies and then regenerates. Usually reversible.
- **Cochlea** — hair cells take it up. The hearing loss is permanent.
- **Dorsal root ganglion** — outside the blood-nerve barrier, so platinum accumulates. The neuropathy keeps worsening for months after the last dose.
- **Marrow** — hit for the ordinary reason. It divides.

13 pathophysiology nodes: a shared upstream chain (transporter uptake → platinum-DNA adducts → oxidative and mitochondrial injury → apoptosis) branching into those four arms. 11 `conforms_to` across `drug_induced_nephrotoxicity` (4), `sensorineural_hair_cell_loss` (2), `peripheral_axonal_degeneration` (3), `myelosuppression` (2).

### Two conformances deliberately withheld

**Not `peripheral_axonal_degeneration#Distal Axonal Degeneration and Demyelination`.** Cisplatin neuropathy is a sensory neuronopathy — platinum accumulates in and kills the ganglion cell body. Symptoms beginning distally is a different claim from the lesion being distal, and no cited source reports demyelination.

**Only the mitochondrial half of `#Axonal Transport Impairment and Mitochondrial Dysfunction`.** The axonal-transport mechanism in chemotherapy neuropathy is established for the microtubule-targeting agents — taxanes, vinca alkaloids — and `PMID:38660386` states it in those terms. Cisplatin is not a microtubule poison, so the mechanism was not transferred to it.

The `sensorineural_hair_cell_loss` conformance likewise covers the death half of its node; no mechanotransduction measurement is reported by the sources here.

`PMID:21145397` supplies the good mechanistic detail: DRG neurons repair nuclear adducts *as fast as tumour cells do*. They are not worse at repair — they start from a higher adduct load, and mitochondria have no nucleotide excision repair at all. That is why the neuropathy coasts.

### Genetics curated against a grading review, not citation frequency

Only **ACYP2** and **TPMT** clear the bar in the Canadian Pharmacogenomics Network 2022 update (`PMID:37726872`) — OR ≥3, replicated in ≥3 independent populations, out of 40 reports across 47 populations and 24 genes.

**COMT is curated `DISPUTED` on purpose.** It is a frequently cited candidate ototoxicity locus and easy to record as established. The same review that graded the field puts it in the further-research tier. Recording it at the strength of the other two would misstate the evidence.

The ACYP2 HGNC identifier was resolved from the ontology, not taken from the deep-research gene table.

### Evidence

52 items over 10 references, every snippet pre-verified against the local cache before the YAML was written.

Strongest items: SIOPEL 6 (`PMID:29924955`), a randomised phase 3 trial where sodium thiosulfate cut hearing loss from 63% to 33% (RR 0.52, 95% CI 0.33–0.81, P=0.002) without harming survival — the six-hour delay exists so the tumour is not protected too. And a magnesium meta-analysis (`PMID:37530867`), OR 0.22 (95% CI 0.14–0.35), whose 11 pooled studies are retrospective, which is recorded.

**No frequency band is asserted on any phenotype.** For hearing loss that is deliberate rather than lazy: incidence is 75% under age five against 48% over five in the same cohort, so one band would misrepresent a strongly age-dependent risk.

**Claims deliberately not made:** no hair-cell-regeneration claim (the usual explanation for the permanence, absent from my sources — the irreversibility itself is cited); no erythropoietin mechanism for the anaemia; no hydration-protocol claim beyond the magnesium arm, since the cited review says the best protocol is still uncertain.

### Deep research

96 KB report; 32/32 references resolved, 0 unresolved, 13/13 quoted claims found in source. NEC preflight `SKIP` (no causal gene for a drug toxicity), cleared manually — top gene mentions are ACYP2, TPMT, COMT and SLC22A2.

Used as leads only. It contributed the genetics section and the magnesium source, both fetched and verified independently before use. It also flagged the COMT trap, which I would otherwise have curated wrong.

### Self-audit

Three overclaims in my own prose, caught and corrected before pushing: "in every cell it enters", "behaving completely differently", and an unsourced statement that mammalian cochlear hair cells are not replaced that was doing load-bearing work inside a causal edge description.

### Validation

`just validate-disorders` (the batched sweep CI runs) passed; 52/52 snippets verified; `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets`, `check-entity-refs` clean; 20 structural tests pass; all 11 conformance anchors resolve to real module nodes. All `reference_title` values generated programmatically from the cache files.


caufieldjh:
`claude-review` failed on this PR without posting a review. The cause is an account usage limit, not anything in the diff:

```
detail="You've hit your limit · resets 6am (UTC)" num_turns=1 cost=$0 subtype=success
```

Run: https://github.com/monarch-initiative/dismech/actions/runs/33583960166

`test (3.13)` and `check-mention` both pass. Locally, `just validate-disorders` (the batched sweep CI runs), 52/52 snippets verified, `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets` and `check-entity-refs` are all clean, 20 structural tests pass, and all 11 `conforms_to` anchors resolve to real module nodes.

Re-run the review after the reset with:

```
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10483
```

caufieldjh:
All eight IMPORTANT items are in. Three of them were things I could not have caught with any check I run.

**1. Cell type — you're right, and it was invisible.** `CL:1000838` sits under `CL:0002584` renal *cortical* epithelial cell, so it excludes the S3 straight segment where cisplatin injury is worst — and my own node description says distal segments are affected too. Switched to `CL:0002306`, which every node of the conformed module uses. The id and label agreed with each other throughout, so `validate-terms` was never going to flag it.

**2. `evidence_source`.** Regraded `MODEL_ORGANISM`. The quoted conclusion rests on the rat arm — sex differences in uptake clearance, urinary NAG, the castrated-male comparison — and the record carries the `Animals` keyword. Reason stated in the explanation. Also took the smaller point: the `PMID:19144690` background sentence is now `OTHER` where it supports the AKI phenotype.

**3. Disconnected phenotypes.** Before fixing I checked whether wiring them is real practice: 233 of 400 KB entries wire phenotypes as downstream targets. So the entry was the outlier. All eight now hang off the terminal node of their arm, and a graph check confirms one trigger, no dangling targets, nothing unwired.

**4.** ACCL0431 (`PMID:27914822`, OR 0.31) now cited alongside SIOPEL 6 — two trials in different paediatric populations, noted as complementary rather than a replication. Duloxetine added, bounded by the guideline's own finding that no agent prevents CIPN. Acetyl-L-carnitine curated as a `REFUTE` item, and dose modification as the intervention that acts on exposure. A reader should be able to learn from this entry what does *not* work.

**5.** `clinical_trials` block for NCT00652132 and NCT00716976, both fetched.

**6.** Four `prevalence` records, one per endpoint. Each carries the caveat that makes it non-comparable with the others: an age stratum for hearing loss (75% under five against 48% over five), an upper bound for nephrotoxicity, a background statement for neuropathy, and an *untreated*-population rate for emesis that describes a population which no longer exists in practice.

**7.** Emesis arm added with nausea and vomiting. Two deliberate limits: the 5-HT3/NK1 receptor mechanism is **not** asserted, since the cited source names the antiemetic classes rather than the biology producing the emesis; and the inbound edge is `INDIRECT_UNKNOWN_INTERMEDIATES`, because this arm is not established here as running through the transporter uptake that gates the four injury arms.

**8.** Rebound to `NCIT:C1230`. You were right about the reasoning, not just the term — I used a cache-availability argument as though it were an ontology argument, and the cache is populated by validation. `NCIT:C1230` resolved and cached on the first validate run, which proves your point. The note now records the correction rather than quietly replacing the old rationale.

Suggestions not taken this round: HUMAN_MODEL_MISMATCH discussion, iPSC organoid `experimental_models`, CKD as a long-term outcome, environmental block, diagnosis section, hypokalemia. All reasonable; I'd rather land the eight than half-do thirteen.

67/67 snippets over 14 references. `just validate-disorders`, `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets`, `check-entity-refs` all clean; 20 structural tests pass; branch rebased onto current `main`.

PR #10491, opened 2026-09-02
curate: add Immunodeficiency 127 (TNF deficiency)

### LINKED ISSUE #10472: Curate immunodeficiency 127 (MONDO:0975832)
Curate a dismech entry for **immunodeficiency 127** ([MONDO:0975832](https://monarchinitiative.org/MONDO:0975832)) — a recently numbered monogenic inborn error of immunity.

The OMIM "immunodeficiency N" series is gene-defined and one-per-entity, so a `kb/disorders/` Disease is the expected outcome; it should then be added to `kb/groupings/Inborn_Errors_of_Immunity.yaml` and given an IUIS category via the `disease-classification` skill. The stub carries no `genes:` block, so the first curation step is identifying the causal gene and the immunological defect from OMIM. Being a high number, this is likely a very recent entity with one or two index reports — cite what exists rather than generalizing from the wider IEI literature.

Stub: `stubs/Immunodeficiency_127.yaml`

gingin77:
Adds a knowledge base entry for **immunodeficiency 127** ([MONDO:0975832](https://monarchinitiative.org/MONDO:0975832)) — inherited TNF deficiency. People with two broken copies of the TNF gene get recurrent tuberculosis in their lungs, and essentially nothing else goes wrong. They tolerate the BCG vaccine, run normal fevers and inflammatory responses, and have normal blood counts.

That narrowness is the interesting part, and it is what the entry is built around. TNF turns out to be needed for one specific job: it tells lung macrophages to switch on their oxidative burst, the burst of reactive oxygen they use to kill bacteria they have swallowed. Without TNF that switch never flips, so the macrophage engulfs *M. tuberculosis* and then cannot finish it off. Add TNF back to the cells in a dish and the burst comes back.

This also explains something clinicians already knew: people taking anti-TNF drugs for arthritis or Crohn's disease are at higher risk of tuberculosis. The same defect appears in healthy donors' macrophages treated with TNF blockers.

## What's in the entry

```
biallelic TNF loss of function
        ↓
absent macrophage TNF production
        ↓
impaired macrophage respiratory burst   ←── rescued by adding TNF back
        ↓
failure to control M. tuberculosis
        ↓
recurrent pulmonary tuberculosis
```

Each arrow carries its own citation. The rescue experiment is recorded as a model link with a `RESTORED` readout rather than as prose, because it is the experiment that makes the third arrow causal instead of merely correlated.

Contrast worth keeping in view: in chronic granulomatous disease the oxidative burst is missing from *every* phagocyte, and those patients get many kinds of infection. Here it is missing from one macrophage population, and they get one.

## Sources

One primary paper, [PMID:39198650](https://pubmed.ncbi.nlm.nih.gov/39198650/) (Arias et al., *Nature* 2024), which describes two first cousins from a consanguineous Colombian family homozygous for a private frameshift, `c.190_191ins20` (p.P64Lfs*13). Both developed tuberculosis in their late teens. There are no other patients reported, so the entry says so rather than implying a wider evidence base.

Deep research used openscientist. Falcon is the repo default but no `EDISON_API_KEY` was available, so the provider was chosen explicitly rather than swapped in quietly. The report was treated as leads: the allele, the family, the ages at onset and the role of TNFR1 were each re-checked against the cached full text of the primary paper before being written, and are cited to that paper rather than to the report. None of the report's suggested ontology terms were used — its own term validation flagged that it had labelled `HP:0032251` "Tuberculosis" when HPO calls it "Abnormal immune system morphology".

## Two things I could not do cleanly

**The tuberculosis exposure term is unbound.** `ECTO:3000134` "exposure to *Mycobacterium tuberculosis*" exists and resolves in EBI OLS, but it is not in the pinned `sqlite:obo:ecto` build that `just validate-terms` reads, so binding it fails validation today. The nearest available term, `ECTO:3000133`, is for *M. tuberculosis* subsp. *tuberculosis*, which is narrower than the claim. I left it unbound with a `review_notes` line recording the search rather than substitute a term that means something else. Worth rebinding when the local ECTO build is refreshed.

**Two negative findings are not modelled as phenotypes.** "Normal inflammatory response" and "normal blood leukocyte development" are diagnostically important here — they are why this disorder is invisible to routine immunology workup — but `ModifierEnum` has no `NORMAL` value and `Phenotype` has no absence slot. They live in the disease description and in evidence on the pathophysiology nodes. If there is an intended pattern for recording an informative negative phenotype, I would like to use it.

No `datasets:` block. Searching GEO for TNF returns the general TNF literature, not anything about this disorder — the named-entity-confusion trap the dataset guidance warns about.

## Validation

`just validate` passes: schema, terms and references. 20/20 evidence snippets verified against cached references. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-environmental-evidence`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens` and `check-snippet-grading` all pass. Compliance 86.9% global, 89.5% weighted.

Closes #10472

### DIFF
--- a/kb/disorders/Immunodeficiency_127.yaml
+++ b/kb/disorders/Immunodeficiency_127.yaml
@@ -239,10 +239,11 @@ pathophysiology:
     supports: SUPPORT
     directness: INDIRECT
     evidence_source: OTHER
-    snippet: "To review, from an Infectious Diseases perspective, the safety profile of agents targeting tumour necrosis factor-"
+    snippet: "anti-TNF-α therapy (infliximab, adalimumab, golimumab, certolizumab pegol and etanercept) is associated with a two-to four-fold increase in the risk of active tuberculosis"
     explanation: >-
-      The clinical counterpart of the in-vitro blockade result. Indirect for
-      this disorder because it is a consensus safety review of therapeutic TNF
+      Quantifies the clinical counterpart of the in-vitro blockade result - a
+      two-to-fourfold rise in active tuberculosis on anti-TNF therapy. Indirect
+      for this disorder because it is a consensus safety review of therapeutic
       blockade rather than a study of inherited TNF deficiency.
   - reference: PMID:39198650
     reference_title: "Tuberculosis in otherwise healthy adults with inherited TNF deficiency."
@@ -562,21 +563,6 @@ experimental_models:
         evidence_source: IN_VITRO
         snippet: "The stimulation of TNF-deficient iPS-cell-derived macrophages with TNF rescued the respiratory burst."
         explanation: Reports the rescue measurement in this model system.
-    - name: Mycobacterium tuberculosis replication in alveolar macrophage-like cells
-      target: Impaired Macrophage Respiratory Burst
-      direction: RESTORED
-      interpretation: >-
-        Closer to the disease-defining endpoint than the burst itself.
-        Replication is comparable to control at baseline but is strongly reduced
-        by TNF stimulation, showing that TNF signalling is what limits the
-        organism in this compartment.
-      evidence:
-      - reference: PMID:39198650
-        reference_title: "Tuberculosis in otherwise healthy adults with inherited TNF deficiency."
-        supports: SUPPORT
-        evidence_source: IN_VITRO
-        snippet: "The replication of M. tuberculosis was similar in AML cells from patients with TNF deficiency and healthy controls but was strongly decreased by TNF stimulation"
-        explanation: Reports the mycobacterial replication measurement and its TNF dependence in patient alveolar macrophage-like cells.
     evidence:
     - reference: PMID:39198650
       reference_title: "Tuberculosis in otherwise healthy adults with inherited TNF deficiency."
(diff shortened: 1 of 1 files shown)


gingin77:
All five blocking points are addressed, and all five were right — thanks, particularly for the listeriosis catch.

| Item | What changed |
|---|---|
| 1. Listeriosis | Phenotype `HP:5210020` added; new `Failure of Macrophage Control of Listeria monocytogenes` node downstream of the burst defect; framing softened in the description and the ORGANISM node |
| 2. Anti-TNF TB risk | Cited from the cached abstract, plus `PMID:29459143` for the clinical counterpart |
| 3. Macrophage restriction | Both quotable sentences attached — normal neutrophil/monocyte ROS, and the pan-phagocyte contrast |
| 4. Unused references | All three now carry a claim |
| 5. Unstructured allele | `Genetic.variants` + `genetic_context` with zygosity |

**On the listeriosis.** You were right that this was not a scoping decision — the entry asserted something the paper spends a paragraph qualifying. I took the wider point too: impaired control of *intramacrophagic bacteria* is the better statement of the defect than "M. tuberculosis only", and that is how the description now puts it. The edge from the macrophage defect to the clinical phenotype is `INDIRECT_KNOWN_INTERMEDIATES` rather than `DIRECT`, because the authors name pregnancy as a co-contributor; that qualification is quoted as its own evidence item so the hedge is in the record and not just in my head.

**On item 5**, zygosity was the part that bothered me once you pointed it out — the entry made a recessive claim with nothing structured behind it. Now `HOMOZYGOUS` / `GERMLINE` / `LOSS_OF_FUNCTION` on `genetic_context`, with the allele in `Genetic.variants` carrying `PATHOGENIC` and its own evidence.

**Non-blocking suggestions taken:** pyrazinamide and ethambutol so HRZE is complete; the M. tuberculosis replication readout, which as you say is closer to the disease-defining endpoint than the burst; a `diagnosis` section for the absent-TNF-production signature; and a `differential_diagnoses` block for chronic granulomatous disease, since the entry defines itself against that contrast.

**One suggestion not taken.** `NCIT:C15615` resolves to "gp100 Antigen/Incomplete Freund's Adjuvant" — not a clinical action, and not reachable from `NCIT:C25218`, so it would fail `RegimenTerm`/`TreatmentTerm` validation. Kept `NCIT:C15986`. If there is a more specific antituberculous-therapy term under `C25218` I have missed, I will take it.

**On `HP:0040285` Excluded** — that is a genuinely useful answer, and it is the right shape for the problem. Agreed it is a schema change rather than something to work around here; I have left the two negative findings in prose. Happy to open the `FrequencyEnum` issue separately if that is useful rather than assuming someone else will.

**Validation after changes:** `just validate` passes; **36/36** snippets verified, up from 20/20. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-length`, `check-title-snippets` and `check-snippet-grading` all green. Compliance **91.3%** global / 93.2% weighted, up from 86.9% / 89.5%.

An `EDIT` history record recording this response is included.

gingin77:
Both fixed in `509cce1f7`. Both were mine, and both arrived *with* the fix commit rather than being in the original entry — which is the more annoying kind of regression.

**The ESGICH snippet was the worse of the two.** That citation existed only to close your earlier "anti-TNF TB risk is asserted and cited nowhere" item, and I closed it with the abstract's AIMS sentence — what the paper set out to do, not what it found. The claim was still uncited in substance; it just looked cited. Now quotes the finding: a two-to-fourfold increase in active tuberculosis on anti-TNF therapy.

**On the misfiled readout** — you're right, and the fix was bigger than moving it. The experiment used AML cells generated from the patients' own monocytes, which is a *different kind of model* from the iPSC line: primary patient cells carrying the real genotype, versus an engineered knockout. Filing the result under the iPSC entry misrepresented the evidence class. It now sits under a new `PRIMARY_CELL_CULTURE` model with the identifying quote ("we infected AML cells generated from healthy controls and TNF-deficient patients with live M. tuberculosis"), and is retargeted onto `Failure of Macrophage Control of Mycobacterium tuberculosis` — the node it actually measures — rather than the respiratory burst node.

`RESTORED` → `DECREASED`. You're right that it contradicted its own interpretation: baseline replication matched controls, so nothing was restored.

**Validation:** `just validate` passes; 37/37 snippets verified. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms` and `check-snippet-length` all green.

PR #10492, opened 2026-09-02
curate: add PAICS Deficiency (de novo purine synthesis defect)

### LINKED ISSUE #10471: Curate PAICS deficiency (MONDO:0859003)
Curate a dismech entry for **phosphoribosylaminoimidazole carboxylase deficiency** ([MONDO:0859244](https://monarchinitiative.org/MONDO:0859244)) — an inborn error of de novo purine biosynthesis affecting the bifunctional PAICS enzyme (AIR carboxylase / SAICAR synthetase), which catalyses steps 6 and 7 of the pathway.

MONDO records only `MONDO:0003847` hereditary disease as parent and no descendants, so it is a leaf and a `kb/disorders/` Disease is the expected outcome. dismech has no purine-biosynthesis entry yet, so there is no existing parent to fold this into. The stub carries no `genes:` block; confirm PAICS and the reported phenotype from OMIM/Orphanet rather than assuming it. This is an ultra-rare entity with a very small published case literature, so expect the evidence work to be the limiting factor — cite what exists and leave sections empty rather than padding.

Stub: `stubs/Phosphoribosylaminoimidazole_Carboxylase_Deficiency.yaml`

gingin77:
Adds a knowledge base entry for **PAICS deficiency** ([MONDO:0859003](https://monarchinitiative.org/MONDO:0859003)), a very rare inherited disorder of purine synthesis. Cells build purines — the A and G of DNA and RNA — by a ten-step assembly line. PAICS is a single protein that runs steps six and seven. When both copies are broken, babies are born with a cluster of malformations: a blocked esophagus, heart defects, urogenital anomalies, spine and limb defects, and a recognisable facial appearance. The first two patients described died within days of birth.

Fewer than ten people have been reported, and the entry is written to make that visible rather than to imply a settled picture.

## Two open problems the entry records rather than smooths over

**The toxic chemical everyone expected isn't there.** In the two related purine-synthesis disorders, the substrate piles up behind the enzyme block and poisons the cell. Here the authors looked for the two expected chemicals in patient cells and found neither — even though one of them kills cells in culture when supplied directly. So the working mechanism is not poisoning but a failure of the *purinosome*, the temporary cluster the ten enzymes form so intermediates can be handed straight from one to the next. Patient cells cannot assemble it, and putting a working copy of PAICS back in fixes that.

The practical consequence: there is no chemical to measure for diagnosis and nothing to remove for treatment.

**The same genetic change gives very different diseases.** All the malformation patients carry the same variant, p.(Lys53Arg), on both copies. One pair of siblings died as newborns. An unrelated boy with the identical genotype was seven years old with normal development. And a separately reported pair of siblings had severe neurological decline with *none* of the malformations. Something other than the variant is deciding the outcome, and nobody knows what.

Both are recorded as `KNOWLEDGE_GAP` discussions attached to the relevant nodes.

## Housekeeping: the stub was keyed on a dead term

`MONDO:0859244` "phosphoribosylaminoimidazole carboxylase deficiency" is obsolete — MONDO merged it into `MONDO:0859003` "PAICS deficiency" ([mondo#9884](https://github.com/monarch-initiative/mondo/issues/9884)). The entry binds the live term. Claim issue #10471 was retitled to match, since the MONDO ID in a claim title is what the double-claim check reads. This PR retires the stub, so there was nothing left to repoint.

## Sources and the deep-research report

Three primary papers: [PMID:31600779](https://pubmed.ncbi.nlm.nih.gov/31600779/) (the founding description), [PMID:39726239](https://pubmed.ncbi.nlm.nih.gov/39726239/) (third patient, normal neurodevelopment), and [PMID:42569864](https://pubmed.ncbi.nlm.nih.gov/42569864/) (fourth patient, integrating the divergent phenotypes).

openscientist was the provider — falcon is the repo default but no API key was available, so the substitution was made explicitly rather than quietly. The report arrived with no reference or term validation frontmatter and no citations sidecar, so both were retro-fitted in place. References came back 14/14 resolved with nothing off topic. Term validation flagged two identifiers the report had named as something else, and one of them mattered: it offered `GO:0034023` for "purinosome", which GO actually calls "5-(carboxyamino)imidazole ribonucleotide mutase activity". That term was not bound; no suitable GO term for the purinosome appears to exist, so the concept is carried in prose and the node is bound to `GO:0006189` instead.

## What I left out, and why

The phenotype list is shorter than OMIM's. OMIM #619859 and HPO carry a fuller craniofacial and skeletal annotation — flat face, brachycephaly, low-set ears, missing ribs, lumbar hemivertebrae, talipes, clinodactyly, hypospadias and more. All three primary papers are abstract-only in the reference cache, and none of those individual features appears in an abstract, so none could be given a verified quote. I left them out rather than attach them to a citation that does not state them, and recorded the gap in the entry's `notes` so the next curator can pick it up.

`ORPHA:633099` would supply quotable rows, but it is not in the committed Orphanet cache and `data/orphadata/en_product1.xml` is gitignored and absent from the worktree, so a rebuild would have needed a full `refresh-orphadata`. That felt like the wrong thing to attach to a curation PR.

One claim from the report I chose not to assert: it describes p.(Lys53Arg) as rs192831239 with a gnomAD non-Finnish-European frequency of 0.135%, which would make its recurrence a population-frequency effect rather than a founder effect. Plausible, and it would be a better explanation than "recurrent variant" — but it is not verifiable from any cached reference, so the entry says only what PMID:39726239 states.

No `datasets:` block. PAICS is heavily studied as a cancer proliferation gene, so a dataset search on the symbol returns oncology series with nothing to do with this disorder.

## Validation

`just validate` passes: schema, terms and references. 26/26 evidence snippets verified. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values` and `check-qualifier-terms` all pass.

Closes #10471

### DIFF
--- a/history/disorders/PAICS_Deficiency/2026-09-03T142444Z-claude-code-4a285d.yaml
+++ b/history/disorders/PAICS_Deficiency/2026-09-03T142444Z-claude-code-4a285d.yaml
@@ -1,0 +1,39 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: PAICS_Deficiency
+  path: kb/disorders/PAICS_Deficiency.yaml
+session:
+  id: 2026-09-03T142444Z-claude-code-4a285d
+  timestamp: '2026-09-03T14:24:44Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10492
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - diagnosis
+  - genetic
+  - prevalence
+  summary: 'Address PR #10492 third review: evidence_source regrading, variant consequence'
+  details: |
+    Third round of review response on PR #10492. Both blocking items accepted; four of six suggestions taken.
+
+    1. Two evidence items quoting the fibroblast enzyme assay were graded HUMAN_CLINICAL and should be IN_VITRO. The measurement is an enzyme assay on cultured skin fibroblasts, and patient-derived material does not make it clinical - CLAUDE.md puts cultured cells under IN_VITRO explicitly. Regraded both. The reviewer's diagnosis of why check-snippet-grading could not catch it is correct and worth recording: that gate keys on divergence, and both copies of the identical sentence carried the same wrong value, so there was nothing to diverge. The same file already graded three other patient-fibroblast experiments IN_VITRO, so this was an internal inconsistency rather than a systematic misreading.
+
+    2. The variant consequence was argued in prose - 'a severe hypomorph rather than a null', backed by 10% patient / 25% recombinant / 50% carrier residual activity - but not machine-queryable. The reviewer suggested adding functional_impact_category to genetic[0]. That slot does not exist on Genetic; it belongs to GeneticContext, which attaches to Pathophysiology. Added there instead, on the PAICS Bifunctional Enzyme Deficiency node, carrying functional_impact_category PARTIAL_LOSS_OF_FUNCTION together with zygosity HOMOZYGOUS, variant_origin GERMLINE and allele_type missense, which also closes the separate suggestion about allele_type and zygosity being unset. PARTIAL_LOSS_OF_FUNCTION rather than LOSS_OF_FUNCTION is the point of the item: a null would not be survivable to seven years.
+
+    Suggestions taken: the GO granularity inversion was real and is corrected - GO:0006189 'de novo IMP biosynthetic process' now sits on the specific flux node and its parent GO:0006164 on the purinosome-assembly node, which reads the right way round. Pulmonary hypoplasia and Short stature are now wired under Disrupted Embryonic Organogenesis. The prevalence notes undercounted: they said four confirmed patients when the entry itself curates the neurodegenerative sibpair, so the record now says at least six and enumerates the three groups.
+
+    Suggestions not taken, deliberately. The three neurodevelopmental phenotypes remain orphan in the pathograph, as the reviewer allowed - their relationship to the malformation phenotype is the open question the entry flags, and drawing edges would presuppose an answer. PMID:22180458 and PMID:32557644 are left uncited here: both speak to purinosome biology across the DNPS disorders rather than to PAICS deficiency specifically, and the reviewer's own framing is right that the natural home for them is a shared 'disrupted purinosome assembly' module across PAICS, ADSL and ATIC rather than this entry. The IMPC Paics knockout is left out for the stated reason - no PMID, so nothing snippet-verifiable.
+
+    Validation: just validate passes; 37/37 snippets verified. pytest passes on the entity-ref, causal-target, foreign-key and subtype selections (18,585 tests). All offline gates green.
--- a/kb/disorders/PAICS_Deficiency.yaml
+++ b/kb/disorders/PAICS_Deficiency.yaml
@@ -74,6 +74,17 @@ pathophysiology:
       evidence_source: IN_VITRO
       snippet: "the PAICS mutation prevented purinosome formation in the patient's skin fibroblasts, and this phenotype was corrected by transfection with the wild-type but not the mutated PAICS"
       explanation: Transfection rescue with wild-type but not mutant PAICS establishes that the variant causes the purinosome defect.
+  genetic_context:
+    zygosity: HOMOZYGOUS
+    variant_origin: GERMLINE
+    allele_type: missense
+    functional_impact_category: PARTIAL_LOSS_OF_FUNCTION
+    description: >-
+      A severe hypomorph rather than a null: roughly 10% residual catalytic
+      activity in patient fibroblasts, about 25% for the purified recombinant
+      p.(Lys53Arg) protein, and about 50% in heterozygous carriers. The
+      recurrent allele is homozygous in two unrelated families; a fourth patient
+      is compound heterozygous for p.Ser35Phe and p.Cys281Ter.
   molecular_functions:
   - preferred_term: phosphoribosylaminoimidazole carboxylase activity
     term:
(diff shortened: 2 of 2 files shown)


gingin77:
All seven blocking items fixed, and five of the six suggestions taken. The "under-consumption" framing was the right diagnosis — eleven cache files committed, three cited.

| Item | Status |
|---|---|
| 1. `diagnosis:` section | Added — three entries, including an explicit negative one for the absent metabolite marker |
| 2. `CL:0000057` → `CL:0002620` | Fixed |
| 3. Unevidenced purinosome→flux edge | Cited from `PMID:35331738` |
| 4. Neurodevelopmental regression + ocular | Added (`HP:0002376`, `HP:0000478`) |
| 5. `prevalence:` | Added — `CASES_IN_LITERATURE` / `ULTRA_RARE` |
| 6. `BEHAVIORAL` on surgical treatment | Split into `SURGERY` + `OTHER` |
| 7. Progression snippet | Swapped |

**Item 3 was the one I'm most glad you caught.** That edge is the whole argument of the entry — channelling failure costs more flux than the enzyme deficit alone — and it was the one link with nothing behind it, with the supporting paper sitting uncited in the same PR. `PMID:35331738` also gave a second quote establishing PAICS as the purinosome interaction hub, which is *why* losing it disassembles the complex rather than just removing a step. That now backs the node too.

**Item 4** was the sharper miss, though. The entry's prose insists the phenotype is split, the discussions are built on the split, and `phenotypes:` contained only one half of it. A consumer querying phenotypes would have seen no trace of the arm the entry argues matters.

**On the CRISPR HeLa suggestion — this was the best item in the review and I've taken it as more than a suggestion.** It genuinely sharpens `missing_toxic_metabolite`: intermediates *do* accumulate behind the block in that system, where they were undetectable in patient fibroblasts. That reframes the gap from "no accumulation happens" to "accumulation is cell-type or genotype dependent", which is an answerable question rather than a dead end. Added as an `experimental_models` entry with `PARTIALLY_RECAPITULATES` and an `INCREASED` readout, and the discussion now states the knockout-versus-hypomorph question explicitly.

**One suggestion I could not action: `HP:0003811` Neonatal death.** I added it and term validation rejected it — `Value 'HP:0003811' not in dynamic enum 'PhenotypeTerm'`. Mortality terms sit outside the phenotypic-abnormality branch that `PhenotypeTerm` is `reachable_from`, so they cannot be bound as phenotypes at all. This is the same wall I hit with `HP:0001522` Death in infancy on the sibling PR #10494, where early death likewise had to go into `progression` and `clinical_burden` instead. Two independent entries in one batch wanting a mortality phenotype and neither being able to have one looks like a real gap rather than a one-off, but it is a schema question and not mine to settle in a curation PR — flagging it rather than working around it. Early death remains in `progression` and `clinical_burden` here.

**On the `REFUTE` grading (suggestion 5):** left as-is. You're right that a separate `SUPPORT`-graded "normal neurodevelopment" phenotype would model it more directly, but the claim being refuted is specifically the obligate-feature reading of developmental delay, and the `explanation` says that. Happy to change it if you'd rather.

**Validation:** `just validate` passes; **37/37** snippets verified, up from 26/26. All offline gates green. Compliance 85.2% global / 85.6% weighted.

The residual compliance gaps are `evidence` on container objects — the organogenesis `downstream` edges, treatments, discussions. Those are the places where the mechanism genuinely is not established, which is why the final edge is `INDIRECT_UNKNOWN_INTERMEDIATES` in the first place. I did not manufacture citations to close them.

An `EDIT` history record is included.

gingin77:
Fixed in `f082e634a`. `biological_scale` on `Impaired De Novo Purine Synthesis` is now `CELLULAR`.

You identified the right reason this was blocking rather than a suggestion. The scale tag on its own is a judgment call; a provenance record asserting an edit that never landed is a different problem — it tells the next curator the question was settled when it wasn't. My string replacement silently didn't match, and the history record went out claiming the change anyway. That is the defect worth fixing, and it is a failure mode I should watch for: I verify the YAML validates after a scripted edit, but I wasn't verifying that each individual replacement actually applied.

**And thank you for the correction on `HP:0003811`.** I'd flagged it as unusable and it's genuinely useful to have that confirmed independently against `Thanatophoric_Dysplasia` — it means the constraint is established practice rather than something I hit once and worked around. Early death stays in `progression`.

`just validate` passes; 37/37 snippets verified.

gingin77:
Both blocking items fixed, four of six suggestions taken. `71e4835d7`.

**On the `evidence_source` regrading — you're right, and your account of why the gate missed it is the useful part.** `check-snippet-grading` keys on *divergence*, and both copies of the identical sentence carried the same wrong value, so there was nothing to diverge from. That's a blind spot worth knowing about: the gate catches inconsistency, not consistent error. The fact that the same file already grades three other patient-fibroblast experiments `IN_VITRO` confirms it was an internal inconsistency rather than a misreading of the rule.

**One correction on the placement of item 2.** `functional_impact_category` isn't a slot on `Genetic` — the class has `name`, `gene_term`, `presence`, `evidence`, `association`, `relationship_type`, `variant_origin`, `review_notes`, `subtype`, `frequency`, `case_fractions`, `inheritance`, `variants`, `features`, `notes`, `examples`. It belongs to `GeneticContext`, which attaches to `Pathophysiology`. So it went onto the `PAICS Bifunctional Enzyme Deficiency` node instead, carrying `zygosity`, `variant_origin` and `allele_type` alongside it — which also closes your separate suggestion about those being unset.

Your substantive point stands and I've used your value: `PARTIAL_LOSS_OF_FUNCTION`, not `LOSS_OF_FUNCTION`. That distinction is the whole point here — a null would not be compatible with survival to seven years, and it is what the 10% / 25% / 50% residual-activity series is telling us.

**Suggestions taken:** the GO granularity inversion was real and is corrected — `GO:0006189` now sits on the specific flux node with its parent `GO:0006164` on the purinosome node. `Pulmonary hypoplasia` and `Short stature` are wired under `Disrupted Embryonic Organogenesis`. And the prevalence notes did undercount: they said four patients while the entry itself curates the neurodegenerative sibpair, so the record now says at least six and enumerates the three groups.

**Not taken, deliberately.** The three neurodevelopmental phenotypes stay orphan — as you noted, their relationship to the malformation phenotype is the open question the entry flags, so an edge would presuppose the answer. And `PMID:22180458` / `PMID:32557644` speak to purinosome biology across the DNPS disorders rather than to PAICS deficiency specifically; your own suggestion of a shared "disrupted purinosome assembly" module across PAICS, ADSL and ATIC is the right home for them, and I'd rather leave them for that than half-cite them here.

**Validation:** `just validate` passes; 37/37 snippets verified. pytest green on the entity-ref, causal-target, foreign-key and subtype selections (18,585 tests).

PR #10493, opened 2026-09-02
curate: add Porphyria-Related Leukoencephalopathy (biallelic HMBS)

### LINKED ISSUE #10470: Curate leukoencephalopathy, porphyria-related (MONDO:0958226)
Curate a dismech entry for **leukoencephalopathy, porphyria-related** ([MONDO:0958226](https://monarchinitiative.org/MONDO:0958226)) — cerebral white matter disease arising in the setting of HMBS-related hepatic porphyria.

This one needs a real call, and it is the first thing to settle. MONDO gives it two parents — `MONDO:0019046` leukodystrophy and `MONDO:0700382` HMBS-related hepatic porphyria — and records no descendants. dismech already curates `Acute_Intermittent_Porphyria` (the HMBS disease) and collects the porphyrias in `kb/groupings/Inherited_Porphyrias.yaml`; that AIP entry currently says nothing about leukoencephalopathy or white matter change. So the choice is between:

- a `kb/disorders/` Disease in its own right, if the literature supports a distinct mechanism (rather than the reversible posterior encephalopathy seen during acute attacks); or
- a phenotype and/or `has_subtypes` entry folded into `Acute_Intermittent_Porphyria`, if it is better modeled as a neurological manifestation of that disease.

Establish which before writing an entry, and record the reasoning in the stub `notes` either way.

Stub: `stubs/Leukoencephalopathy_Porphyria-related.yaml`

gingin77:
Adds a knowledge base entry for **porphyria-related leukoencephalopathy** ([MONDO:0958226](https://monarchinitiative.org/MONDO:0958226)), caused by inheriting two damaged copies of *HMBS*.

*HMBS* is the acute intermittent porphyria gene, but this is not a severe version of AIP. It is a different disease. People with AIP have one damaged copy and suffer episodic attacks of abdominal pain and neurological upset. People with two damaged copies have **no attacks at all**. Instead they develop, from childhood or adolescence, slowly progressive leg stiffness, unsteadiness, nerve damage in the limbs, and eye problems, with a distinctive and consistent pattern of white matter damage on brain MRI. Adults survive into late life. There is also a severe infantile form with seizures and early death.

## The decision the claim issue left open

[#10470](https://github.com/monarch-initiative/dismech/issues/10470) asked whether this should be its own Disease or folded into `Acute_Intermittent_Porphyria` as a phenotype or subtype. I curated it as its own Disease. The inheritance differs, the clinical syndrome differs, the MRI lesion is reproducible across unrelated families, and — the deciding point — **the treatments that work in AIP do not work here.** Attaching this to the AIP entry would attach a false prognosis and a false treatment expectation along with it.

The existing `Acute_Intermittent_Porphyria` entry mentions no leukoencephalopathy or white matter change, so nothing is duplicated.

## The most useful fact in the entry is a negative one

Both attempted treatments are curated with `REFUTE` evidence rather than left out:

- **Liver transplantation does not work.** It was tried on the AIP logic that replacing the liver's enzyme removes the source of the toxic precursors. Precursor concentrations were unchanged afterwards in both blood and cerebrospinal fluid, and neurological decline continued.
- **Intravenous haem does not work.** It does not lower CSF precursors, and parenteral heme does not reach the brain at all — which is also why givosiran, the liver-targeted drug that prevents AIP attacks, was declined for one patient after transplant.

Getting this wrong costs a patient a liver transplant, so it belongs in the record explicitly, not as an omission.

## What is still unknown

Two hypotheses compete for what actually damages the white matter: the accumulated precursors acting inside the brain, or an absolute shortage of heme there. They are not equivalent — one would call for a drug that lowers precursors inside the CNS, the other for heme or a substitute that crosses into the brain. The therapeutic failures are consistent with either, so they do not discriminate. Both edges from CNS precursor exposure to the white matter lesion are recorded as `INDIRECT_UNKNOWN_INTERMEDIATES` for that reason, rather than drawn as if the mechanism were settled.

Part of why it stays unresolved: no animal model reproduces the central disease. That is recorded as its own knowledge gap and explains why the entry has no `animal_models` block.

## Sources and the deep-research report

Six primary papers spanning the original 2016 description ([PMID:27558376](https://pubmed.ncbi.nlm.nih.gov/27558376/)) through 2026 case reports. openscientist was the provider — falcon is the repo default but no API key was available, so the substitution was explicit rather than silent. The report arrived with no validation frontmatter, so reference and term validation were retro-fitted: 32/32 references resolved, nothing off topic, no unresolved terms.

Two things were taken from the report, each re-verified against a freshly fetched primary source rather than cited to the report:

- [PMID:11478735](https://pubmed.ncbi.nlm.nih.gov/11478735/), showing that 5-aminolevulinic acid blocks GABA-A receptor binding in human as well as rat brain membranes. This gives the precursor-toxicity hypothesis a concrete molecular candidate. Graded `SUPPORT` with `directness: INDIRECT`, because the experiment used millimolar ALA on membranes from unaffected brain and was framed around AIP.
- The absence of any animal model, now a knowledge gap.

One thing was deliberately **not** taken. The report's mechanism diagram includes an ALAS1 de-repression step. That is the AIP attack mechanism, and it is exactly what does not apply here — liver transplantation leaves CSF precursors unchanged. Writing it in would have contradicted the entry's own evidence.

Similarly, the AIP-with-PRES case reports that surfaced during searching ([PMID:40186107](https://pubmed.ncbi.nlm.nih.gov/40186107/), [PMID:36597305](https://pubmed.ncbi.nlm.nih.gov/36597305/)) describe the reversible one-copy phenocopy and belong to `Acute_Intermittent_Porphyria`, not here. The report itself flags this conflation as a curation hazard, and I agree with it.

## Validation

`just validate` passes: schema, terms and references. 34/34 evidence snippets verified. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys` and `check-enum-values` all pass. Compliance 90.9% global, 91.7% weighted.

Closes #10470

### DIFF
--- a/history/disorders/Porphyria-Related_Leukoencephalopathy/2026-09-03T161032Z-claude-code-46aef9.yaml
+++ b/history/disorders/Porphyria-Related_Leukoencephalopathy/2026-09-03T161032Z-claude-code-46aef9.yaml
@@ -1,0 +1,38 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Porphyria-Related_Leukoencephalopathy
+  path: kb/disorders/Porphyria-Related_Leukoencephalopathy.yaml
+session:
+  id: 2026-09-03T161032Z-claude-code-46aef9
+  timestamp: '2026-09-03T16:10:32Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10493
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  summary: 'Address PR #10493 round 6: self-contradiction from a half-applied fix,
+    cataract frequency'
+  details: |
+    Sixth round of review response on PR #10493. Both findings accepted.
+
+    1. The previous round's fix was half applied and left the entry contradicting itself. I softened the notes block to 'most reported individuals have no AIP attacks, though a minority do' and the disease description still said 'affected individuals do not have the episodic neurovisceral attacks that define AIP'. Both render on the same page. The reviewer is right that this is worse than the state it replaced: an entry that overstates consistently is at least coherent.
+
+    The mechanism of the failure is one I have now hit twice in this PR series and should name. My string replacement for the description did not match, and the script reported success anyway because I only validated that the resulting YAML parsed. The same thing happened on PAICS in an earlier round, where a history record claimed a biological_scale change that never landed. The fix applied here is to assert on every replacement anchor before writing, and this edit does that - all four replacements are asserted, so a non-matching anchor now aborts rather than passing silently. I also grepped every occurrence of 'attack' in the file and checked each one rather than fixing only the line the reviewer quoted, which is what I should have done last round.
+
+    2. Cataract was raised to FREQUENT with only a single-patient snippet attached, which supports the phenotype existing and not its frequency. The sentence that carries the frequency - 'Bilateral cataracts, optic nerve atrophy, and cerebellar abnormalities were also common' - was in a cache file this entry cites four times and quoted nowhere. Attached to Cataract, and also to Optic atrophy, whose FREQUENT had been resting on two affected members of a single three-person kindred.
+
+    Also corrected in the same description: I called PMID:41731635 'the most recent review'. It is a case report in J Med Case Rep; the sentence sits in its introduction summarising thirteen prior cases. Reworded to say what it is. PMID:42396593 is the paper whose title actually claims a review of prior reported cases.
+
+    Validation: just validate passes; 59/59 snippets verified, up from 57/57. pytest -k porphyria 124 passed. All offline gates green.
--- a/kb/disorders/Porphyria-Related_Leukoencephalopathy.yaml
+++ b/kb/disorders/Porphyria-Related_Leukoencephalopathy.yaml
@@ -5,8 +5,11 @@ description: >-
   Porphyria-related leukoencephalopathy (LENCEP), also called HMBS-related
   leukodystrophy, is an autosomal recessive disorder caused by biallelic HMBS
   variants. Although HMBS is the acute intermittent porphyria gene, this is not
-  a severe form of AIP: affected individuals do not have the episodic
-  neurovisceral attacks that define AIP, and instead show slowly progressive
+  a severe form of AIP: most affected individuals do not have the episodic
+  neurovisceral attacks that define AIP - a minority do, and fleeting
+  neurovisceral pain is described - but none has been reported to have attacks
+  provoked by the classical porphyrinogenic stimuli that trigger AIP. What they
+  show instead is slowly progressive
   spastic paraparesis, cerebellar ataxia, peripheral axonal neuropathy, ocular
   disease, and a recognizable leukoencephalopathy of the periventricular and
   deep cerebral white matter and bilateral thalami. Onset is in childhood or
@@ -387,6 +390,14 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: "in 2 patients, optic atrophy as well as vertical gaze and convergence palsies and nystagmus"
     explanation: Reports optic atrophy in two of three affected family members.
+  - reference: PMID:41731635
+    reference_title: "Biallelic pathogenic hydroxymethylbilane synthase gene variants of a neurodegenerative disorder with progressive cystic leukoencephalopathy: a case report."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Bilateral cataracts, optic nerve atrophy, and cerebellar abnormalities were also common."
+    explanation: >-
+      Supports the FREQUENT grading across the reported cases, rather than
+      resting it on the two affected members of a single kindred.
 
 - category: Ophthalmologic
   name: Nystagmus
@@ -449,7 +460,8 @@ phenotypes:
   description: >-
     Childhood-onset cataracts, reported in several individuals and sometimes the
     earliest recorded finding. Grouped with optic atrophy and cerebellar change
-    as a common feature in the most recent review.
+    as a common feature in the case report that summarises the thirteen prior
+    published cases.
   frequency: FREQUENT
   phenotype_term:
     preferred_term: Childhood-onset cataract
@@ -463,6 +475,16 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: "She also had childhood-onset cataracts."
     explanation: Documents childhood-onset cataract in a genotyped patient.
+  - reference: PMID:41731635
+    reference_title: "Biallelic pathogenic hydroxymethylbilane synthase gene variants of a neurodegenerative disorder with progressive cystic leukoencephalopathy: a case report."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "Bilateral cataracts, optic nerve atrophy, and cerebellar abnormalities were also common."
+    explanation: >-
+      The sentence the FREQUENT grading rests on. The single-patient item above
+      establishes the phenotype exists; this one establishes it is common across
+      the reported cases.
+
 
 - category: Neurologic
   name: Cognitive impairment


gingin77:
All four blocking items fixed, both suggestions taken.

**Item 1 is the one that changed an argument, not just filled a gap.** I had written the precursor-toxicity vs. CNS-heme-deficit question as undecidable, and said the therapeutic failures don't discriminate between them. That was true as far as it went, but the PBGD-deficient mouse does discriminate, and its paper was sitting uncited in this PR. The neuropathy develops chronically and progressively *while ALA is normal or only twofold raised* — the opposite of what a precursor-toxicity model predicts — and the authors conclude heme deficiency and hemeprotein dysfunction cause it.

That is now evidence on the `precursor_toxicity_vs_cns_heme_deficit` discussion, graded `SUPPORT` / `directness: INDIRECT`: it is a partial deficiency with a peripheral phenotype, so it constrains the mechanism without settling it for the white matter, and the rationale says so rather than overclaiming.

The R167Q mouse is the model-treatment join you identified — constitutively high PBG and ALA in brain and CSF, no ALAS1 induction, no haem response, which is exactly the pharmacological situation in patients. Linked `RECAPITULATES` against the CNS precursor node and `FAILS_TO_RECAPITULATE` against the IV haem treatment.

And you were right that `no_animal_model_of_the_biallelic_cns_disease` was overstated. Narrowed to what is actually true: the peripheral neuropathy and the biochemical compartment are both modelled; the leukodystrophy is not.

**Item 2 was a real contradiction and I should have caught it.** The description said activity is "far below the roughly 50% seen in heterozygous AIP carriers" while its own snippet gives a 50–66% *decrease* — 34–50% residual, overlapping the heterozygous range. Only the 18% kindred supported the strong reading. It now states the actual range and says residual activity tracks the specific allele pair rather than biallelic status alone, which is also more consistent with the severity gradient the entry describes elsewhere.

**Items 3 and 4** done. `HP:0002073` also made the separate `clinical_course: PROGRESSIVE` redundant, so that came out.

**On the orphan-phenotype suggestion** — I wired in the three I could justify (cognitive impairment, and the infantile seizure and developmental delay), all `INDIRECT_UNKNOWN_INTERMEDIATES` since the anatomical correlation is asserted clinically rather than demonstrated. I deliberately left cataract, optic atrophy and the eye-movement findings orphan: no source connects them to the white matter lesion, and drawing those edges to reduce an orphan count would be inventing mechanism. Flagging that as a choice rather than an oversight.

**Anticonvulsant guidance** added as its own treatment, with the caveat that the non-porphyrinogenic rationale is inherited from AIP practice rather than demonstrated here — where hepatic induction is precisely what is *not* driving the disease.

**Validation:** `just validate` passes; **41/41** snippets verified, up from 34/34. All offline gates green. Compliance 87.7% / 88.2%.

Thanks for flagging that I'd correctly declined the report's givosiran claim — I'd rather know the negative checks are visible too.

gingin77:
All four findings fixed, all three suggestions taken. `afd8e9eee`.

**The critical one was a real test failure and I should have caught it.** I pointed a `modeled_mechanisms` target at a *treatment*, which that slot cannot resolve, and used `FAILS_TO_RECAPITULATE` to say the mouse *matches* the human non-response — the opposite of what it means. I reproduced the failure locally before fixing.

Worth naming why it reached you: CI selects pytest by changed path, and a curation PR touches only `kb/`, matching neither the python nor the schema filter. So `test (3.13)` was green while `test_animal_model_mechanism_targets` never ran. That is precisely the gap `CLAUDE.md` documents as the reason `check-entity-refs` and `check-duplicate-keys` are ungated CI steps rather than only tests — I'd been treating the offline `just check-*` gates plus green CI as sufficient, and for anything that adds a cross-referencing block it isn't. Running `pytest -k` is now part of what I do before pushing.

**I retargeted rather than deleted, because your finding 2 supplies what the link needs.** You're right that `PMID:41377573` reports the CNS phenotype — normal histopathology, brain haem at ~80% of wild type — and that "not characterized in the sources cited here" was wrong twice over. That reported negative is more useful than the hedge, and it makes `FAILS_TO_RECAPITULATE` valid against `Cerebral White Matter Degeneration`: the closest genetic model does not produce the defining lesion *despite* carrying the constitutive precursor elevation.

That also cuts into the central mechanism gap, which I've noted in the link: neither the precursor burden nor a central haem deficit was sufficient to produce white matter pathology in that animal.

**The truncated snippet** — good catch on the cause. `non‐responsive` uses U+2010, not an ASCII hyphen, which is exactly where my copy stopped. Extended through "haem provision".

**HP:0000496** split into `HP:0025721` and `HP:0000619`. I took your point about preferring `HP:0025721` over `HP:0000511` since the source does not establish supranuclear localization.

**Suggestions, all three taken:** the cerebellar atrophy edge the entry already asserted in prose is now drawn with its own evidence; a `prevalence` block records the thirteen published patients; and `PMID:8563760` — the 1996 primary description, fetched by this PR and uncited — now sits alongside the 1999 follow-up.

The six orphan ocular phenotypes stay unlinked, as you allowed. The mechanism genuinely isn't known and an edge would be invention.

**Validation:** `just validate` passes; **46/46** snippets verified, up from 41/41. `pytest tests/test_data.py -k "animal_model or entity_ref or causal or foreign"` — 16,039 passed. All offline gates green.

gingin77:
All four findings fixed. `7a4c29a4c`.

**Finding 1 is the same defect class as round 2, which is the part worth noting.** I wrote that precursors were "unchanged in both compartments" after transplantation. That is true of the adult in `PMID:41377573` and false of the infant in `PMID:42396593` — cited in the same treatment block — whose overproduction halved. I generalized one patient's result across both without checking the other source I was already citing.

And you're right that the correction is mechanistically stronger. Substantial hepatic correction that still failed to arrest decline is better evidence for an extrahepatic source than "nothing changed" would be. The description now states both outcomes and says so.

**Finding 2 is the one I'd call a safety defect rather than an accuracy one.** A treatment block named "…with Non-Porphyrinogenic Agents" quoted the initial regimen including phenobarbital and stopped there, omitting that it was tapered off *because* barbiturates induce hepatic ALAS1. Added as its own evidence item. I also noted in the description that the avoidance rationale is inherited from AIP — where hepatic induction is the driver — and so is weaker in principle here, though it was acted on clinically.

**Finding 3 was sitting in a paper I cite four times.** Doss porphyria is a genuinely good discriminator: biallelic ALAD raises ALA but not PBG, and gives severe peripheral polyneuropathy with no CNS involvement. That argues ALA elevation alone is insufficient for the white matter lesion. The rationale now sets it against the PBGD-deficient mouse, which points at heme deficiency instead, and states plainly that the two point in different directions — which is why this narrows the gap rather than closing it.

**Finding 4** — diagnosis block added, framed on the point that the diagnosis turns on **zygosity**, not on finding an HMBS variant. One allele is AIP, two are this disease, several biallelic alleles are themselves known AIP alleles, and the treatment expectations differ. Three entries: sequencing with segregation, erythrocyte HMBS activity separating patients from carriers, and CSF precursor measurement.

**Two suggestions I did not take, with reasons.**

The ocular frequency asymmetry is real rather than inconsistent: optic atrophy has a count in the source (two of three), while the gaze, convergence and nystagmus findings come from a clause that gives none. Making them uniform would assert a frequency the source does not state.

The R167Q `publication:` stays on `PMID:41377573`. That is the source I actually read for those findings; pointing it at the primary mouse paper would mean citing something I have not fetched. Same reasoning for the secondhand `MODEL_ORGANISM` provenance — you're right that it is one step removed, but re-attributing the quote to a study I have not read would be worse than provenance that is visibly indirect.

**Validation:** `just validate` passes; **53/53** snippets verified, up from 46/46. All offline gates green including `check-snippet-length`.

Thanks for checking the givosiran claim independently — good to know the negative calls are being verified too, not just the positive ones.

gingin77:
All four fixed, all four suggestions taken. `5b54971da`.

**You named the pattern, and that's the useful part of this review.** All four are the same defect: prose asserting more than the snippet under it supports, with the sentence I needed already sitting in a cache file I cite. And it's the third consecutive round of it — round 2 on the enzyme-activity claim, round 4 on the transplant negative, four more here.

The cause isn't sourcing. It's that I've been writing the description from memory of what a paper said and then attaching whichever nearby quote looked supportive, rather than writing the description *from* the quote. Those produce different prose, and the difference is exactly this: "affected individuals do not have AIP attacks" instead of "most do not", "haem does not lower precursors" instead of "does not lower them centrally". Worth recording as a habit to change rather than four more edits.

**Item 1 — the categorical claim was wrong and the correct version is available verbatim.** `PMID:41731635` says "most of them did not present acute porphyric attacks" and describes an adult with biallelic HMBS who did have them; the index patient in `PMID:41377573` had fleeting neurovisceral pain that recurred post-transplant. Non-inducibility is the defensible claim and it's quotable: *"None of these patients have been reported to have attacks in response to classical porphyrinogenic stimuli."* Both passages softened, quote attached to the inheritance block.

**Item 2 makes the entry's argument stronger, as you say.** Haem *does* work peripherally — ~50% transient reductions in urinary ALA, PBG and porphyrins — and stating that turns "haem doesn't work" into the compartmental dissociation the pathograph already models: the periphery responds, the CNS does not, because parenteral haem doesn't cross. Same shape as the transplant argument, and I'd made the same flattening error in both places.

**Item 3** — fair. The Doss paragraph reached the rationale and the evidence list never gained the snippet, which is half a fix.

**Item 4** — both claims were in one sentence I hadn't quoted. Added as `MODEL_ORGANISM`.

**All four suggestions taken:** cataract to `FREQUENT` (the review groups it with optic atrophy and cerebellar change as "also common"); prevalence corrected to fourteen — you're right that thirteen were *previously* published plus the index case; white matter `preferred_term` made specific while keeping the general binding. The "transient subjective improvement" phrasing is now covered by the peripheral-response evidence from item 2.

**Validation:** `just validate` passes; **57/57** snippets verified, up from 53/53. `pytest -k porphyria` — 124 passed. All offline gates green.

gingin77:
Both fixed. `975da5830`.

**You're right that item 1 is worse than what it replaced.** I softened the `notes` and left the `description` asserting the categorical version, so the entry stated both on the same page. An entry that overstates consistently is at least coherent; one that contradicts itself is not.

**The mechanism is worth naming, because this is the second time in this PR series.** My string replacement for the description didn't match, and my script reported success anyway — I was only checking that the resulting YAML parsed, not that each replacement had applied. The identical silent no-op produced the PAICS history record claiming a `biological_scale` change that never landed. Two rounds of review consumed by edits I believed I had made.

Fixed at the source: this edit asserts on every replacement anchor before writing, so a non-matching anchor aborts instead of passing quietly. And I grepped all eleven occurrences of "attack" and checked each rather than fixing the line you quoted — which is what I should have done last round, since you'd told me there were two passages.

**Item 2** — fair, and the same defect as the previous round in a different place. `FREQUENT` rested on "She also had childhood-onset cataracts", which supports the phenotype existing and says nothing about frequency. The sentence that carries it was in a file I cite four times and quote nowhere. Attached to Cataract, and to Optic atrophy too, whose `FREQUENT` had been resting on two members of a single three-person kindred.

**And you're right that I mischaracterised the source.** `PMID:41731635` is a case report in *J Med Case Rep*, not a review; the sentence sits in its introduction summarising thirteen prior cases. `PMID:42396593` is the one whose title claims a review. Reworded to say what the paper actually is.

**Validation:** `just validate` passes; **59/59** snippets verified, up from 57/57. `pytest -k porphyria` — 124 passed. All offline gates green.

PR #10494, opened 2026-09-02
curate: add Cardiomyopathy Dilated 2H (GET3/ASNA1)

### LINKED ISSUE #10468: Curate cardiomyopathy, dilated, 2H (MONDO:0859358)
Curate a dismech entry for **cardiomyopathy, dilated, 2H** ([MONDO:0859358](https://monarchinitiative.org/MONDO:0859358)) — a genetically defined familial dilated cardiomyopathy: left ventricular dilatation with impaired systolic function, progressing to heart failure and arrhythmia.

MONDO places this under `MONDO:0700335` familial isolated dilated cardiomyopathy with no recorded descendants, so it is a leaf. dismech has strong precedent for curating individual locus-defined DCM entities as their own entries — `Dilated_Cardiomyopathy_1A/1B/1D/…`, `Cardiomyopathy_Dilated_2G`, `Cardiomyopathy_Dilated_100` — collected in `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml`. The expected outcome is a `kb/disorders/` Disease following that naming pattern, added to the grouping. The stub records no `genes:` block, so confirm the causal gene and inheritance from OMIM before curating; the "2" series is the autosomal recessive DCM numbering, which should be checked rather than assumed.

Stub: `stubs/Cardiomyopathy_Dilated_2H.yaml`

gingin77:
Adds a knowledge base entry for **dilated cardiomyopathy 2H** ([MONDO:0859358](https://monarchinitiative.org/MONDO:0859358)), a recessive heart muscle disease that kills in early infancy.

What makes it mechanistically interesting is the gene. Most inherited dilated cardiomyopathy comes from broken structural proteins — titin, lamin, myosin. This one comes from breaking a delivery service. Around 5% of membrane proteins are "tail-anchored": they are made in the cytoplasm with a sticky patch at one end and have to be escorted to the right membrane afterwards, because that patch would otherwise clump. GET3 is the escort. Lose it and a whole class of membrane proteins never arrives, cardiomyocyte membrane traffic falls apart, and the heart dilates and fails. This was the first evidence that the tail-anchored insertion pathway causes human disease.

## The first deep-research run returned the wrong disease — and why

**The committed report is correct.** `research/Cardiomyopathy_Dilated_2H-deep-research-openscientist.md` carries `mondo_id: MONDO:0859358` in its frontmatter, mentions ASNA1/GET3 throughout and RPL3L not at all. This section describes the *first* run, which was discarded and replaced. An earlier version of this description said the committed artifact was the bad one; that was left stale after the re-run and is corrected here.

The first run named **RPL3L** as the causal gene — 44 mentions of RPL3L, zero of ASNA1 or GET3, an entire mechanism section built on muscle-specific ribosome biogenesis. That is a real disease, but it is CMD2**D**, not CMD2**H**:

| | OMIM | MONDO | Gene | Defining report |
|---|---|---|---|---|
| **CMD2H** (this PR) | 620203 | MONDO:0859358 | **GET3** (ASNA1) | Verhagen 2019 |
| CMD2D | 619371 | MONDO:0030300 | RPL3L | Ganapathi 2020 |

I verified this through NCBI rather than assuming. MedGen concept-to-gene linkage is 1:1 in both directions, and MedGen's own definition of CMD2H cites Verhagen et al. 2019 — which is [PMID:31461301](https://pubmed.ncbi.nlm.nih.gov/31461301/), the ASNA1 paper.

**The cause turned out to be a repo bug, not a provider one.** The `research-disease` recipe hardcoded an empty `mondo_id`, so the provider received a disease *name* and no identifier and disambiguated on its own. A controlled three-run experiment confirmed it: with `mondo_id=''` the report comes back 44 RPL3L / 0 GET3; with `mondo_id=MONDO:0859358` it comes back 0 RPL3L / 49 GET3. The recipe is fixed in #10696 (merged), which extracts the ID from the entry or stub and prints it on the dispatch line so an empty one is visible before the run rather than after. This PR now carries the corrected report from the second run.

**Nothing was taken from either report.** The entry was built from primary literature found independently, before the first report arrived; the corrected report was used afterwards only as a completeness cross-check.

The part still worth generalising: the first report's own reference validation came back **19/19 resolved, 0 unresolved, 0 off topic**. That looks like a clean bill of health and is not one. Relevance is scored against the report's *own* vocabulary, so a report built around the wrong disease has wrong-disease vocabulary throughout and scores all of its wrong-disease citations as on topic. The single signal was one quote-match failure, on `PMID:35323613`, whose text turned out to be about biallelic RPL3L variants. That one line is what prompted the check.

The two diseases are a natural trap: both recessive, both neonatal-onset, both rapidly progressive, both fatal without transplant, OMIM numbers one letter apart.

## What's in the entry

```
biallelic GET3 loss of function
        ↓
impaired tail-anchored protein insertion into the ER
        ↓
disrupted cardiomyocyte membrane trafficking and proteostasis
        ↓
impaired contractility and ventricular remodeling → infantile-onset DCM
```

Two animal models carry `ModelMechanismLink` readouts rather than sitting as disconnected list entries:

- **Zebrafish** `asna1` loss, with the mRNA rescue arm. Wild-type mRNA restores contractility; neither patient allele does. That experiment is the functional evidence that both alleles are pathogenic.
- **Mouse**, cardiomyocyte-specific, constitutive *and* inducible. The pair separates a developmental requirement (myocardial thinning by E16.5, perinatal lethality) from a maintenance requirement (adult deletion → rapid dilation).

A `HUMAN_MODEL_MISMATCH` discussion records the limit of both: they are effectively null, while the human genotype pairs a nonsense allele with a p.(Val163Ala) missense that still makes protein. Neither model can say anything about residual hypomorphic activity.

## One schema note

`HP:0001522` "Death in infancy" sits outside the `PhenotypeTerm` dynamic enum, so early death could not be a phenotype. It is recorded through `progression` and `clinical_burden` instead. Same pattern as `HP:0003811` "Neonatal death". If there is a preferred way to record a fatal outcome I have missed, I would rather use it.

## Validation

`just validate` passes: schema, terms and references. **43/43** evidence snippets verified. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values` and `check-qualifier-terms` all pass.

Closes #10468

https://claude.ai/code/session_01NgzW11dBNCm5ZjtgqMzf7T

### DIFF
--- a/history/disorders/Cardiomyopathy_Dilated_2H/2026-09-06T051538Z-claude-code-2a87a3.yaml
+++ b/history/disorders/Cardiomyopathy_Dilated_2H/2026-09-06T051538Z-claude-code-2a87a3.yaml
@@ -1,0 +1,48 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Cardiomyopathy_Dilated_2H
+  path: kb/disorders/Cardiomyopathy_Dilated_2H.yaml
+session:
+  id: 2026-09-06T051538Z-claude-code-2a87a3
+  timestamp: '2026-09-06T05:15:38Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10494
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - diagnosis
+  - treatments
+  - phenotypes
+  - histopathology
+  - discussions
+  - notes
+  summary: Mine the GeneReviews overview, connect Cardiac arrest, bind fibrosis, add
+    transplantation
+  details: |
+    Third review round on PR #10494. Three blocking items and four suggestions; all seven addressed.
+
+    GeneReviews. The entry tagged PMID:20301486 and mined nothing from it, justifying that on the abstract-only cache and citing Cardiomyopathy_Dilated_2G and Cardiomyopathy_Dilated_100 as precedent. The reviewer checked and CMD100 mines the same abstract-only cache in two places. The note was half wrong: CMD2G does tag without mining and is a real precedent, CMD100 is not. Both minable strings from the cached purpose statement are now quoted where they apply - the proband evaluation strategy on the exome-sequencing diagnosis entry, and the genetic risk assessment of at-risk relatives on genetic counselling - graded OTHER, matching CMD100. The note now records that the earlier conclusion was wrong rather than silently dropping it.
+
+    Cardiac arrest was the one phenotype never named as a target, so it rendered as a disconnected node. It is now a sequela of Dilated cardiomyopathy, as a bare name.
+
+    Histopathology. Prominent subendothelial fibrosis is bound to NCIT:C3044 Fibrosis, following Arrhythmogenic_Right_Ventricular_Cardiomyopathy. NCIT was searched for the other two findings and has nothing usable - no finding term for irregular intercalated discs, and for myofibrillar disorganization only Myofibrillar Myopathy, which is a disease. Both left unbound under no term beats a bad one, with the search recorded in notes so it is not repeated.
+
+    Heart transplantation added, reversing a decline I made twice. My objection was that the two reported patients died at seven weeks and twelve days having never been transplant candidates, so asserting the therapy would go beyond the source. That objection was about evidence, and the repo already has a convention for exactly this case: Dilated_Cardiomyopathy_1EE and Dilated_Cardiomyopathy_1A carry Heart Transplantation with notes reading Deliberately unevidenced, no gene-specific outcomes are reported. That is the honest shape and I should have found it rather than declining twice. Bound to NCIT:C15246 Heart Transplantation, the cached term the sibling DCM entries use, not the NCIT:C15326 the research report suggested. The description says explicitly that neither patient reached candidacy and that whether transplant is reachable in GET3 disease is unknown rather than answered negatively.
+
+    Two snippets that stopped short of a character now carry the full phrase - the QRS range through its en-dash, and the tail-anchored substrate list through all seven proteins with their identifiers. Neither truncation was folded-scalar corruption, which is the thing worth avoiding; both were just weaker quotes than the source offered.
+
+    PLN in the substrate discussion is now marked as printed PPLA in the source, so a reader can find the string in the paper. PLN remains the correct HGNC symbol.
+
+    The PR description was stale and said the committed deep-research report is about RPL3L. It is not - the committed artifact carries mondo_id MONDO:0859358 and zero RPL3L mentions, because the run was repeated after the recipe fix in #10696. I said in the previous round that I was updating the description and did not. Rewritten now and verified by re-reading it from the API.
+
+    Validation: just validate passes; 43/43 snippets verified, up from 41. All per-file offline gates green. The three whole-KB snippet gates report pre-existing findings on this branch's base, none in this file.
--- a/kb/disorders/Cardiomyopathy_Dilated_2H.yaml
+++ b/kb/disorders/Cardiomyopathy_Dilated_2H.yaml
@@ -315,6 +315,9 @@ phenotypes:
   - target: Prolonged QRS complex
     description: >-
       Conduction abnormality accompanying the cardiomyopathy.
+  - target: Cardiac arrest
+    description: >-
+      The proband arrested during the rapid deterioration that took her to ECMO.
   evidence:
   - reference: PMID:31461301
     reference_title: "Biallelic Variants in ASNA1, Encoding a Cytosolic Targeting Factor of Tail-Anchored Proteins, Cause Rapidly Progressive Pediatric Cardiomyopathy."
@@ -411,7 +414,7 @@ phenotypes:
     reference_title: "Biallelic Variants in ASNA1, Encoding a Cytosolic Targeting Factor of Tail-Anchored Proteins, Cause Rapidly Progressive Pediatric Cardiomyopathy."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
-    snippet: "ECG recordings showed sinus rhythm with broad QRS complexes (range 124"
+    snippet: "ECG recordings showed sinus rhythm with broad QRS complexes (range 124–264 ms; Figure IVA in the Data Supplement)."
     explanation: Documents the conduction abnormality on ECG.
 
 - category: Cardiovascular
@@ -503,6 +506,11 @@ biochemical:
 
 histopathology:
 - name: Prominent subendothelial fibrosis
+  finding_term:
+    preferred_term: prominent subendothelial fibrosis
+    term:
+      id: NCIT:C3044
+      label: Fibrosis
   description: >-
     Found on microscopic examination of the myocardium at postmortem, alongside
     increased heart weight relative to body size and severe left ventricular
@@ -565,6 +573,16 @@ diagnosis:
     evidence_source: HUMAN_CLINICAL
     snippet: "Targeted next-generation sequencing of 48 genes implicated in cardiomyopathy revealed no potentially deleterious variants."
     explanation: Establishes that the standard cardiomyopathy panel does not detect this disorder.
+  - reference: PMID:20301486
+    reference_title: "Dilated Cardiomyopathy Overview."
+    supports: SUPPORT
+    evidence_source: OTHER
+    snippet: >-
+      Provide the evaluation strategy of a proband with nonsyndromic DCM
+    explanation: >-
+      The GeneReviews overview sets out the proband evaluation strategy for
+      nonsyndromic dilated cardiomyopathy, which is the workup CMD2H presents
+      into and which the negative panel here interrupts.
 
 prevalence:
 - population: Worldwide
@@ -781,6 +799,17 @@ treatments:
     explanation: >-
       Shows the recurrence risk being acted on clinically - the sibling was
       screened at birth on the strength of the family history.
+  - reference: PMID:20301486
+    reference_title: "Dilated Cardiomyopathy Overview."
+    supports: SUPPORT
+    evidence_source: OTHER
+    snippet: >-
+      relatives of a proband with DCM to inform cardiac surveillance and allow
+      early detection and treatment of DCM to improve long-term outcome
+    explanation: >-
+      States the purpose of genetic risk assessment of at-risk relatives, which
+      is the cascade-testing and surveillance limb of the counselling offered
+      here.
 
 - name: Supportive Heart Failure Management
   description: >-
@@ -825,6 +854,33 @@ treatments:
     snippet: "she was transferred to a tertiary referral hospital for extracorporeal membrane oxygenation. LV function remained poor without any signs of improvement"
     explanation: Records that ECMO was administered and that ventricular function did not improve on it.
 
+- name: Heart Transplantation
+  therapeutic_modality: SURGERY
+  description: >-
+    The definitive option for end-stage infantile dilated cardiomyopathy and the
+    endpoint that mechanical support bridges to. Recorded as the escalation
+    target for this entry's terminal node, not as a therapy either reported
+    patient received: the proband died at seven weeks on ECMO having never
+    reached candidacy, and her sister died at twelve days. Whether transplant is
+    reachable in GET3 disease is therefore genuinely unknown rather than
+    answered negatively - the two known patients deteriorated too fast to test
+    it.
+  treatment_term:
+    preferred_term: heart transplantation
+    term:
+      id: NCIT:C15246
+      label: Heart Transplantation
+  target_mechanisms:
+  - target: Impaired Cardiac Contractility and Ventricular Remodeling
+    description: >-
+      Replaces the failing myocardium outright; the only option here that
+      removes the diseased tissue rather than compensating for its loss of
+      function.
+  notes: >-
+    Deliberately unevidenced; no GET3/ASNA1-specific transplantation outcomes
+    are reported. Follows the convention used by the numbered-series peers
+    Dilated_Cardiomyopathy_1EE and Dilated_Cardiomyopathy_1A.
+
 discussions:
 - discussion_id: which_ta_substrate_drives_the_heart_phenotype
   kind: KNOWLEDGE_GAP
@@ -837,7 +893,8 @@ discussions:
   rationale: >-
     The index paper narrows this considerably: inspecting the predicted human
     tail-anchored proteome it names seven already associated with human
-    cardiomyopathy - DMPK, DYSF, EMD, JPH2, PLN, SYNE1 and SYNE2. So the gap is
+    cardiomyopathy - DMPK, DYSF, EMD, JPH2, PLN (printed as PPLA in the source),
+    SYNE1 and SYNE2. So the gap is
     not that no candidates exist but which of that shortlist is rate-limiting.
     There is also a partial negative result to work against: emerin localized
     correctly to the nuclear membrane in patient myocardium, though nuclei were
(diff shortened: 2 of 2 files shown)


gingin77:
All six blocking items fixed, all five suggestions taken. **Phenotypes went from one to seven; snippets from 16 to 37.**

Your core criticism was right, and I want to name the failure mode rather than just fix it. I treated the deep-research report as *the* research artifact, discovered it was about a different disease, wrote that up carefully — and then effectively proceeded as if this entry had no research backing, instead of mining the cached primary full text as its substitute. The under-consumption wasn't an oversight on individual facts; it followed from that.

| Item | Status |
|---|---|
| 1. Cardiac malformations | VSD (2/2), ASD, tachypnea (2/2), feeding difficulties, prolonged QRS, LV thrombus |
| 2. `histopathology` | Added — fibrosis, intercalated discs, myofibrillar disorganization |
| 3. Cis allele pair | Fixed, plus the absent-NMD result |
| 4. Overstated knowledge gaps | Both rewritten against what the paper says |
| 5. `BEHAVIORAL` modality | → `OTHER`, plus ECMO as its own `DEVICE` treatment |
| 6. `prevalence` + `diagnosis` | Added |

**Item 3 changed the mechanism, not just the detail.** Cys289 is part of an essential zinc-binding site, so the paternal allele is doubly compromised rather than simply truncated — and the absent nonsense-mediated decay means the truncated protein is actually made rather than cleared. Both are now in the description and cited.

**Item 4 is the one I'm most glad you pushed on.** "The current evidence does not name one" was straightforwardly false — the paper names seven candidates. The gap is now *which of the shortlist is rate-limiting*, which is answerable. The emerin negative is curated as a `REFUTE` item, since it narrows the shortlist rather than merely being absent.

**On node bundling I split rather than retagged.** You suggested `TISSUE` on the contractility node; I think the better resolution of the same observation is a real `TISSUE` node between the cellular and organ steps — `Cardiomyocyte Structural Disorganization` — which gives the histopathology its attachment point and removes what was a bare assertion. Contractility stays `ORGANISM`. Say if you'd rather have the simpler version.

## The root cause is repo-wide, not a bad input on one run

You traced the wrong-disease report to `mondo_id: ''` in the frontmatter. That's right, and it goes further than one run. From `justfile`:

```
--var "disease_name=$disease_name" \
--var "mondo_id=" \
--var "category=$category" \
```

It is hardcoded empty. **Every deep-research run this repository has ever dispatched went out with a disease name and no identifier**, which is why CMD2H — a name one letter from a better-documented disease — resolved to the wrong one. That is a real latent defect affecting every entry curated from a DR report, not just this one.

I'm handling that as its own change rather than folding a `justfile` edit into a curation PR. A corrected run for this disease (`--var mondo_id=MONDO:0859358`, which overrides the hardcoded empty) is in flight now; I'll add it here when it lands, and it doubles as the test of whether the empty ID was the cause.

**Validation:** `just validate` passes; **37/37** snippets verified. All offline gates green. Compliance 91.2% / 91.8%.

On GeneReviews — agreed and I get the same answer: gene-specific DCM chapters only, no ASNA1/GET3 entry, and this is a two-patient 2019 report. Not applicable.

gingin77:
Correction to my previous comment: **the corrected re-run I said was in flight did not actually supply the MONDO ID.**

I launched it with a trailing `--var mondo_id=MONDO:0859358`, assuming `*args` would override the hardcoded empty value in the recipe. It doesn't. The report came back with `mondo_id: ''` in its frontmatter and, again, 44 mentions of RPL3L and zero of ASNA1/GET3 — a faithful reproduction of the original failure, not a test of the fix.

Re-running now with the recipe patch applied locally, and I will report the outcome here whichever way it goes. If it still comes back RPL3L, the identifier alone is not sufficient and the answer is a `preflight-dr` gate rather than input plumbing.

The `justfile` fix itself is [#10696](https://github.com/monarch-initiative/dismech/pull/10696). Nothing about this entry's content depends on the outcome — the entry is built entirely from the primary literature — but I did not want the claim that a corrected run had been done to stand when it hadn't.

gingin77:
**Your diagnosis was right, and the re-run you asked for settles it.** Same provider, same disease name, same template — the identifier was the only variable:

| Run | `mondo_id` | RPL3L | ASNA1/GET3 |
|---|---|---|---|
| Original | `''` | 44 | 0 |
| Retry with trailing `--var` | `''` | 44 | 0 |
| Recipe patched | `MONDO:0859358` | **0** | **49** |

The corrected report is now committed in place of the wrong one. It validates clean — 14/14 references resolved, nothing off topic, 8/8 quoted claims found in source — and cites `PMID:31461301` and `PMID:41370295`, the two primary papers this entry is already built from. So it corroborates the entry rather than adding to it, and **no entry content was changed on the strength of it**. That also closes the "zero usable deep-research coverage" gap you flagged.

**The corrected report confirms the mechanism from the other side, unprompted:**

> Early iterations of this investigation initially followed the GATAD1 literature; identifier cross-referencing through Monarch/OLS4 established that MONDO:0859358 → OMIM:620203 → GET3/ASNA1, and the report was corrected accordingly.

GATAD1 is CMD2B (OMIM:614672). So this disease name sits between **two** decoys, not one — CMD2D/RPL3L and CMD2B/GATAD1 — and the identifier is what separates them. The entry `notes` now say that, and name both.

**On whether to keep the wrong-disease report.** You advised keeping it, and that was right when it was the only report. With a correct one in hand I've dropped it: a CMD2D report living at `research/Cardiomyopathy_Dilated_2H-*.md` is a trap for the next curator and no longer buys anything, since its content is regenerable in one command for the disease it actually describes — still queued as `stubs/Cardiomyopathy_Dilated_2D.yaml`, and now reproducible correctly thanks to [#10696](https://github.com/monarch-initiative/dismech/pull/10696). The mix-up stays documented in the entry `notes`, the history record, [#10495](https://github.com/monarch-initiative/dismech/issues/10495) and this thread. Say if you'd rather I restore the file.

Also worth flagging plainly: I told you a corrected run was in flight before it was. That run passed the ID as a trailing `--var`, which does not override the recipe's hardcoded value, so it reproduced the original failure instead of testing the fix. I corrected that here and on #10696 as soon as I saw it. The real test used the patch applied locally and reverted afterwards — `project.justfile` is unmodified on this branch.

`just validate` passes; 37/37 snippets verified.

gingin77:
All four fixed. `7c3d094fb`.

**Item 1 — you were right to raise it a third time.** I corrected the pathophysiology description to "three variants across two alleles, not two" and then never re-read `genetic[0].notes`. So the file asserted the genotype two different ways, with the incomplete version sitting in the section a reader goes to *for* the genotype, and the omitted variant being `c.867C>G p.(Cys289Trp)` — the one that destroys the zinc-binding site. A correction applied in one place is not a correction, and I should have grepped the file for the old claim rather than fixing the instance you'd quoted.

**Thank you for the gnomAD caveat — I'd have got that wrong.** The paper's pLI 0.92 is from v2.0.2 and current gnomAD gives a much lower value for the same gene. The entry now attributes 0.92 to that release explicitly and says the number is version-bound, rather than stating a figure current gnomAD contradicts.

**Item 2 was wrong in a direction that mattered.** `HP:0003642` is HPO's *type I* CDG profile — an N-glycan assembly defect — while the paper reports a **combined** N- and O-linked defect, which is the type II trafficking signature. So the binding pointed opposite to the entry's own argument that this is a readout of the trafficking lesion. Rebound to `HP:0003160`.

And you're right that the `notes` overstated the source. I wrote "This is not an incidental laboratory finding"; the authors write "it is possible that", and in the same passage that because the profiles were only slightly abnormal "these results should not necessarily be considered pathogenic". That is a materially weaker claim than mine, and the rewritten notes carry their hedge and the reason for the general binding.

**Items 3 and 4** — both were `preferred_term` more specific than the binding, which is the wrong way round. `UBERON:0002349` for myocardium (the contractility node keeps `UBERON:0002084`, since that node *is* about the left ventricle), and `HP:0001684` for the secundum ASD I had already named in the preferred_term.

**Validation:** `just validate` passes; 37/37 snippets verified. `pytest tests/test_data.py -k "animal_model or entity_ref or causal or foreign or model_readout"` — 18,742 passed. All offline gates green.

gingin77:
Rebased onto current `main` — this PR had gone `DIRTY`, which is why `pr-shepherd` skipped it while merging the three sibling curation PRs and #10696.

The conflicts were all mechanical collisions with what had just merged: term-cache CSV rows in the same sorted region, and reference-cache files for papers a sibling PR had fetched. Resolved by taking `main`'s copy of each and re-deriving my own rows with `just validate-terms` + `just normalize-cache`, rather than hand-merging — cache rows are derived artifacts and hand-editing them is explicitly forbidden.

No content changed. `just validate` passes and every snippet still verifies.

The force-push dropped the approval, since `main` has `dismiss_stale_reviews` on. Re-requesting review.

**One thing worth flagging that is not mine to fix.** The rebase kept stalling on `references_cache/DOI_10.1172_JCI152297.md`. `main` carries the same paper cached twice under two DOI casings — `DOI_10.1172_JCI152297.md` and `DOI_10.1172_jci152297.md` — and a case-insensitive filesystem (macOS default) can only hold one of them on disk, so git reports the other as permanently modified and refuses to continue a rebase. I worked around it with `--assume-unchanged` and cleared the flag afterwards; the branch is unaffected.

That will block any rebase on macOS, not just mine. The fix is to drop one of the two entries on `main` and regenerate via `just fetch-reference` so the DOI casing is canonical, which belongs in its own PR.

gingin77:
Both blocking items fixed, four of five suggestions taken, bundled into one push. `03afc2343`

**On the pathograph — the septal-defect point was the useful part.** You left the parent choice to me, and you were right that they aren't sequelae of the cardiomyopathy: both were present on day-one echo, before ventricular function declined. They now hang off `Cardiomyocyte Structural Disorganization` as `INDIRECT_UNKNOWN_INTERMEDIATES`, since how loss of tail-anchored protein targeting disturbs septation isn't established. Tachypnea, feeding difficulties, LV thrombus and prolonged QRS are `sequelae` of the DCM node, where they do belong.

**The biochemical entry** — "the argument is written; it just isn't structured" was exactly the right diagnosis. It now carries `readouts` onto the trafficking node, graded `CORRELATES_WITH` rather than `READOUT_OF` and the evidence `INDIRECT`, because the authors say the profiles were only slightly abnormal and "should not necessarily be considered pathogenic". Supportive care gained `target_mechanisms`.

**On the erratum, I took your preferred option.** You're right that in an entry whose own lesson is provenance hazard, an unmentioned correction on its primary paper is the wrong thing to leave implicit — and right that the earlier history record claimed it as a source when it contributed nothing. `notes` now records that a correction exists on PMID:31461301, that its content is not retrievable so its substance is unchecked, and that someone with journal access should confirm no curated claim is affected. I kept the cache file rather than dropping it, so the correction stays discoverable.

**Suggestions taken:** cardiac arrest (2/2 — good catch, it's in both siblings); `NCIT:C101295` as `diagnosis_term`; the human contractility snippet, which was sitting in the same file under the ASD phenotype while that node rested on zebrafish alone; and the description unbundled.

**Not taken:** `finding_term` on the histopathology entries. You flagged subendothelial vs subendocardial as genuinely ambiguous and suggested leaving it unbound may be better — I agree, and no term beats a wrong one.

**On GeneReviews:** I can't fully close this either. No chapter exists for ASNA1/GET3 that I'm aware of, and CMD2H is a two-patient 2019 entity, so one is very unlikely — but I'm recording that as reasoning rather than a verified check, same as you did.

**Validation:** `just validate` passes; **40/40** snippets verified, up from 37/37. `pytest -k "entity_ref or causal or foreign or readout or sequelae"` — 13,823 passed. All offline gates green.

gingin77:
Rebased onto current `main` again — this had gone `DIRTY` a second time, on the same `cache/hp/terms.csv` sorted region. No content change: `main`'s copy taken and my rows re-derived with `just validate-terms` + `just normalize-cache`, as before. 40/40 snippets still verify and `just validate` passes.

The force-push dropped the approval you'd just given. Apologies for the churn — re-requesting.

**The case-collision problem I flagged on the last rebase has grown.** It is now three pairs on `main`, not one:

```
references_cache/DOI_10.1172_JCI152297.md   vs  ..._jci152297.md
references_cache/DOI_10.1016_S2213-8587(24)00028-7.md  vs  ..._s2213-...
references_cache/DOI_10.1056_NEJMoa055262.md  vs  ..._nejmoa055262.md
```

Each pair is the same paper cached twice under different DOI casing. A case-insensitive filesystem — macOS default — can only hold one file per pair, so git reports the other as permanently modified and refuses to continue a rebase. I masked them with `--assume-unchanged` for the duration and cleared every mask afterwards; none is staged and the branch is unaffected.

This will block any rebase on macOS, and it is accumulating rather than static. The fix is to drop one entry from each pair and regenerate via `just fetch-reference` so the casing is canonical — its own PR, not this one.

gingin77:
GeneReviews baseline tagged. `584d61960`

You're right, and this is the third review to touch it — the previous two and I all left it as "unverified, probably absent" because no ASNA1/GET3 chapter exists. That reasoning was sound but the conclusion was wrong: the applicable resource is the **disease-level** overview, `PMID:20301486` was already sitting in the cache, and both nearest siblings tag it. Now tagged, with a GeneReviews-scope paragraph in `notes` following `Cardiomyopathy_Dilated_2G` — recording that the cached record is `abstract_only` so no snippet is quoted from it.

**A warning for anyone doing this again:** I ran `just tag-references` to check the tagging and it rewrote **332 other KB files**. All reverted; only this entry is staged. That recipe operates on the whole corpus rather than the file you hand it, so it cannot be run casually inside a curation branch.

**Suggestions taken:** `genetic_context` on the molecular node, so zygosity and the three-variant detail are queryable rather than prose-only; and a Genetic Counseling treatment — not a formality here, since the second affected child was echocardiographed on day one *because of the family history*, which is the recurrence risk being acted on.

**Declined — and I want to flag a disagreement rather than resolve it silently.** You suggest heart transplantation belongs in `treatments`. An earlier review on this same PR examined exactly that and concluded the omission was *correct, not a gap*: grepping the primary full text, the only transplant is an unrelated LMNA patient transplanted at 16, and neither CMD2H patient was ever a candidate — they died at 7 weeks and 12 days. The deep-research report does call transplantation definitive for infantile DCM in general, but adding it on that basis asserts a therapy for this disease that its only two reported patients were never offered. I've kept it out and recorded the reasoning in the history record. If you still want it in, say so and I'll add it with the palliative-context caveat explicit.

Congestive heart failure as a separate phenotype also declined: the DCM node with its sequelae — tachypnea, feeding difficulties, cardiac arrest — already carries the course the source describes, and a separate `HP:0001635` would restate it.

You were right about the stale PR description too; the committed report is the corrected GET3 one. Updating it now.

**Validation:** `just validate` passes; **41/41** snippets verified. All offline gates green, scope clean.

gingin77:
All three blocking items and all four suggestions taken. `35c59ddd6`

**1. GeneReviews — you're right, and the precedent claim was the worse half of it.** I checked CMD100 and it mines the same abstract-only cache in two places, exactly the two strings you named. So the note asserted a precedent for *not* mining that the cited entry contradicts. CMD2G does tag without mining and is a genuine precedent; CMD100 is not. Both strings are now quoted where they apply — the proband evaluation strategy on the exome-sequencing `diagnosis` entry, the at-risk-relative risk assessment on `Genetic Counseling` — graded `OTHER` as CMD100 does. The note now says the earlier conclusion was wrong rather than quietly dropping it.

**2. `Cardiac arrest`** is now a `sequelae` target under `Dilated cardiomyopathy`, bare name. Good catch — the proband's arrest is what took her to ECMO, so it belonged in that chain and not off to the side.

**3. Histopathology.** `NCIT:C3044` bound to the fibrosis finding, per the ARVC precedent. I re-ran the NCIT search for the other two and got the same result you did: nothing for irregular intercalated discs, and only `Myofibrillar Myopathy` near myofibrillar disorganization, which is a disease. Both left unbound, with the search recorded in `notes`.

**7. Heart transplantation — added, and I was wrong to decline it twice.** My objection was evidentiary: both patients died at seven weeks and twelve days having never been candidates, so asserting the therapy seemed to go past the source. But the repo already has a convention for precisely that case — `Dilated_Cardiomyopathy_1EE` and `_1A` both carry Heart Transplantation with `notes: Deliberately unevidenced; no <gene>-specific transplantation outcomes are reported`. That's the honest shape, and I should have looked for it instead of declining. The description now states outright that neither patient reached candidacy and that reachability in GET3 disease is unknown rather than answered negatively.

One correction to the item as filed: the report's `NCIT:C15326` isn't what the sibling DCM entries use. The cached term across `Cardiac_Sarcoidosis`, `Dilated_Cardiomyopathy_1A`, `_1EE`, `_1P` and `_1FF` is **`NCIT:C15246` Heart Transplantation**, so I bound that.

**5. Both truncated snippets extended** — the QRS range through its en-dash and the closing figure reference, the substrate list through all seven proteins with identifiers. You're right that folded-scalar corruption is the thing to avoid and neither of these was that.

**6. `PLN`** now reads `PLN (printed as PPLA in the source)`.

**4. The PR description was stale, and I said last round I was updating it and didn't.** That's the same failure mode as the file edits — claiming a change in a comment without verifying it landed. Rewritten now and confirmed by re-reading it from the API. The committed artifact carries `mondo_id: MONDO:0859358` and zero RPL3L mentions; the section now describes the discarded first run and names the recipe bug (#10696, merged) as the cause, so it can't read as an instruction to delete a correct file.

**Validation:** `just validate` passes; **43/43** snippets verified, up from 41. All per-file offline gates green. The three whole-KB snippet gates report pre-existing findings on this branch's base (`Tangier_Disease` and others) — none in this file.

PR #10499, opened 2026-09-02
curate: Add Autosomal Dominant Nonsyndromic Hearing Loss 25 (MONDO:0011568, SLC17A8)

### LINKED ISSUE #10489: Curate autosomal dominant nonsyndromic hearing loss 25 (MONDO:0011568)
Curate a dismech entry for **autosomal dominant nonsyndromic hearing loss 25 (DFNA25)** ([MONDO:0011568](https://monarchinitiative.org/MONDO:0011568)) — *SLC17A8* (VGLUT3), the vesicular glutamate transporter that loads glutamate into inner hair cell synaptic vesicles, so the lesion sits at the first synapse of hearing rather than in the bundle.

Nothing in `kb/` mentions SLC17A8 or VGLUT3, and no file carries this MONDO term — a clean start.

Two entries curated recently are worth reading first, because they bracket this mechanism and the three should agree on what is established at the ribbon synapse. `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93.yaml` (CABP2) is the other synaptic-transmission deafness here, and `Auditory_Neuropathy.yaml` is where a presynaptic inner-hair-cell lesion would normally be classified — decide explicitly whether DFNA25 belongs in that relationship rather than leaving the two unconnected.

Stub: `stubs/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_25.yaml`

kevinschaper:
Adds a dismech entry for **autosomal dominant nonsyndromic hearing loss 25 (DFNA25)** ([MONDO:0011568](https://monarchinitiative.org/MONDO:0011568), heterozygous *SLC17A8* / VGLUT3), curated as its own Disease. Nothing in `kb/` mentioned SLC17A8 or VGLUT3 before this.

Closes #10489

## The finding that shaped the entry

The two mouse models of this gene disagree about where the disease starts, and both answers are in print.

The **Slc17a8 null mouse** is a textbook auditory synaptopathy. Otoacoustic emissions are robust, sound-evoked auditory-nerve responses are absent, brainstem responses can still be driven electrically, and — the load-bearing negative result — calcium-triggered vesicle turnover is normal. The vesicles fuse; they are empty.

The **VGLUT3 A224V knock-in**, which carries the actual human DFNA25 allele, reaches a different conclusion. Inner hair cell stereocilia bundles collapse while outer hair cell bundles do not, the summating potential falls, and the ribbon changes come later. Its authors write that "DFNA25 stems from a failure in the mechano-transduction followed by a change in synaptic transfer."

So the entry carries **two `mechanistic_hypotheses` groups** rather than picking one — `presynaptic_glutamate_loading_failure` (CANONICAL) and `a211v_mechanotransduction_first` (EMERGING) — and the `downstream` edges opt into whichever group they belong to. They converge on one node, "Failure of Auditory Nerve Activation with Preserved Cochlear Amplification".

## The genetics problem, recorded as a HUMAN_MODEL_MISMATCH

Heterozygous *Slc17a8* knockout mice hear normally. Heterozygous human p.A211V carriers go deaf. So halving VGLUT3 does not cause deafness in a mouse, and neither model has the genotype a patient has: the null is homozygous, and so is the knock-in.

This is not a fidelity quibble, because it decides a therapy. AAV gene addition rescues the null mouse and is the obvious translation — but if p.A211V is a dominant negative or a toxic gain of function, adding wild-type VGLUT3 to a cell that already has a working copy may not help, and allele-selective knockdown would be the right strategy instead. Nothing distinguishes those. The proposed experiment is the heterozygous knock-in, which has not been reported.

For the same reason `functional_impact_category` is left **empty** on the trigger node, with the reason in the block's own `description`: the missense allele is not behaving as a simple loss of function, no experiment licenses one of the narrower values, and the slot is single-valued so it cannot also describe the truncating alleles.

## Relationship to the two neighbouring entries

This was the substantive question on the issue, and the answer is that **DFNA25 is not recorded as a subtype of `Auditory_Neuropathy`**.

The signature that defines that class — preserved otoacoustic emissions alongside an abnormal auditory brainstem response — has been measured in the null mouse and in the knock-in mouse. It has never been measured in a DFNA25 patient; no published DFNA25 report includes emissions testing. The `Auditory_Neuropathy` `has_subtypes` list is populated by entities where the classification rests on a human clinical finding (OTOF, DIAPH3, ATP11A, AIFM1, TMEM43), so adding DFNA25 on mouse evidence would put a weaker claim in that list without saying so.

That is also the honest difference from **`Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93`** (CABP2). DFNB93 has a published human emissions observation and therefore states its synaptopathy classification in its pathophysiology. DFNA25 states the dissociation **for the mouse only**, and carries the human question as an open `KNOWLEDGE_GAP` whose proposed experiment is emissions plus brainstem responses in genotyped carriers. The two entries now agree in the way that matters: each says which species the dissociation was measured in. The comparison is machine-visible as a `differential_diagnoses` entry, not left in prose.

One genuine human-facing hook to the auditory neuropathy literature does exist — PMID:34145196 characterises two suspected SLC17A8 alleles under that heading — and it is cited for what it measured (two alleles, two different molecular consequences, in cell culture) rather than for the classification in its title, which it does not test.

## Why a Disease and not a subtype

MONDO:0011568 is bound nowhere in `kb/`; DFNA25 has its own OMIM phenotype entry, its own DFNA locus number, and a ClinGen **Definitive** gene-disease assertion of its own (upgraded from Strong in 2023). No existing entry binds SLC17A8, so there is no coverage to duplicate. The stub is deleted.

## Deliberately absent

- **No `prevalence`.** ClinGen's own count is nine probands and eight variants worldwide. Any `prevalence_class` would be invented.
- **No frequency band on any phenotype.** There is no denominator that is DFNA25 — the published material is family reports plus one screening cohort of 87 dominant probands in which SLC17A8 explained one.
- **No `progression`,** because the only serial-audiogram study predates the gene and reports onset rather than a rate.
- **No `clinical_trials`, no `datasets`.** No SLC17A8 trial exists, and a gene-symbol dataset search returns the central-nervous-system VGLUT3 literature.

## Named Entity Confusion

`just preflight-dr` returns **WARN, exit 0**. SLC17A8 is mentioned 28 times; the flagged rival is **"ABR" at 39%**, which in this literature is the *auditory brainstem response*, not the ABR gene. False positive. The OMIM pair (605583 phenotype, 607557 gene) is correct in both directions.

Separately: VGLUT3 has a large CNS literature — striatal cholinergic interneurons, hippocampus, tinnitus, optic neuropathy — that is the same protein in different cells. The deep-research run surfaced one such citation (PMID:22794260, striatal dopamine release) and it is not used.

## Ontology terms the deep-research report got wrong

Both caught by resolving through OLS rather than trusting the report:

| Report said | OLS says |
|---|---|
| HP:0012715 = "Bilateral hearing impairment" | **Profound hearing impairment** |
| HP:0002066 = "Hearing impairment" | **Gait ataxia** |

Neither was used. One further note: `HP:0003829` (Typified by incomplete penetrance) is an HPO *clinical modifier*, not a phenotypic abnormality, so it is not in the `PhenotypeTerm` dynamic enum. The age-dependent-penetrance phenotype is therefore curated with no bound term and the reason is stated in its description rather than silently omitted.

## Validation

All exit 0:

- `just validate` — 65/65 snippets verified
- `just validate-terms`
- `just validate-disorders` (the batched gate)
- `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`
- `just validate-history`


kevinschaper:
Correcting the PR body above on the point that matters most, and describing what changed in the follow-up commit.

**I had it wrong.** The body says "no published DFNA25 report includes emissions testing." That is not true, and the correction runs the other way from what I assumed.

While checking a separate claim — I wanted to source the statement that heterozygous mice hear normally, which I could not quote from any cached reference — I searched the full text of the A224V knock-in paper (PMID:34783032) and found this in its discussion:

> Decent DPOAEs were recorded in the knock‐in mutant mouse. In contrast, otoacoustic emissions were absent in patients harbouring the p.A211V allele (Thirlwall et al. 2003). Thus, DFNA25 is more likely to be a progressive hearing loss in human rather than an auditory neuropathy (i.e. absent or desynchronized ABR with intact DPOAEs).

So patient emissions have been reported, they are **absent**, and that is the opposite of both mouse models. The knock-in study draws the classification conclusion explicitly.

**The disposition does not change — it gets a positive reason.** DFNA25 still does not belong in the `Auditory_Neuropathy` `has_subtypes` list, but now because the only published human audiological measurement bearing on the question argues against the classification, not because the question was never asked. That is a stronger argument than the one I made, and it is the direct answer the issue was asking for.

The relationship to `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93` also sharpens rather than blurs: DFNB93 has a human emissions observation showing emissions **present**, which is why that entry asserts its synaptopathy classification in its pathophysiology; DFNA25 has one showing emissions **absent**, which is why this entry asserts the dissociation for the mouse only and curates the human finding as a phenotype that refutes it.

**What the follow-up commit changes**

- New `Absent Otoacoustic Emissions` phenotype (HP:6000182) carrying three evidence items: the finding, the authors' classification conclusion, and — graded `NO_EVIDENCE` because it reports a hypothesis rather than a result — their untested alternative that the absent emissions reflect a lifetime of noise exposure in a middle-aged cohort whose audiogram is already indistinguishable from noise damage. No frequency band: the finding reaches the entry through a later paper's discussion without a numerator.
- A `REFUTE` evidence item against the "Failure of Auditory Nerve Activation with Preserved Cochlear Amplification" node, so the human contradiction is machine-visible rather than only in prose.
- The `HUMAN_MODEL_MISMATCH` is rewritten around a fact I can actually cite. The heterozygous A224V knock-in **has** been phenotyped: it shows a small threshold shift with brainstem amplitudes indistinguishable from wild type, which the authors say "contrasts with the dominant transmission in DFNA25". My previous version rested on an unsourced claim about heterozygous *knockouts* and proposed an experiment that had already been done. Their own verdict on the model is now quoted too — it "might be arguable so far as to consider it as a faithful model of DFNA25".
- The lesion-site knowledge gap is replaced by two sharper ones. Why are patient emissions absent — disease, or noise? (Proposed experiment is the one the authors name: noise vulnerability in mice carrying one or two variant alleles.) And how does a vesicular glutamate transporter variant collapse a stereocilia bundle it is not present in? The paper's own misfolding-and-aggregation candidate is argued against by two results it cites, so that step is genuinely open.
- The stereocilia node now records why the finding is surprising: VGLUT3 is absent from inner hair cell stereocilia and losing it altogether leaves the transducer working.

**Provenance caveat, recorded in the entry's `notes`.** The emissions finding reaches this entry through the 2021 paper's discussion, attributed to the 2003 audiological study. The 2003 abstract does not mention emissions and only the abstract is cached, so I could not quote the primary measurement. The entry cites the 2021 paper for what it states and flags that a curator with full-text access to the 2003 paper should confirm the numbers.

**Validation after the change** — all exit 0: `just validate` (71/71 snippets verified), `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `validate-history`.

**Note on the failed `claude-review` check on the previous commit**: the log shows `You've hit your limit · resets 6am (UTC)` — an account quota issue on the credential in use, so no review ran. Not something in this PR. This push should re-trigger it.

kevinschaper:
One more commit to finish the correction described above. Two passages were left behind, because the script that should have rewritten them aborted partway and wrote nothing: the `notes` still said the emissions signature "has never been measured in a DFNA25 patient", and the DFNB93 differential still said DFNB93 has a human emissions observation and DFNA25 does not. Both said the opposite of what the entry now argues.

They are consistent now, and the comparison is better for having been forced: the two diseases have the same kind of human evidence pointing opposite ways — emissions present in DFNB93, absent in DFNA25 — rather than one having evidence and the other lacking it.

Also in this commit: the hearing-aid description pointed at a knowledge gap that the previous commit had replaced, so it now carries the open question rather than a dead reference; and the convergence node's opening sentence reads as a claim about the models rather than about the patients.

Grepping the file for the stale wording is what caught this, which is worth doing after any edit made by script rather than by hand.

Validation unchanged and all exit 0: `just validate` (71/71 snippets verified), `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`.

kevinschaper:
Blocking item fixed, two suggestions taken, two declined with reasons. Pushed as `ccdf391`.

First, on how I treated this review: an earlier run of the same check approved this same commit, and this one requests changes on it. I checked that both reviews carry the head SHA (`45a4e221bc`) before deciding, and treated the later verdict as live rather than dismissing it as stale. It was right to.

## IMPORTANT — cochlear implant device qualifier: fixed

You're correct, and the inconsistency was internal to the file: `Hearing Aids` carried the `NCIT:C16830` → `NCIT:C183182` pair while `Cochlear Implantation` bound only the action, leaving `NCIT:C157820` unreachable from an entry that discusses cochlear implantation as a mechanistically motivated option.

Verified before applying: `NCIT:C157820` resolves to **Cochlear Implant** and `NCIT:C16830` to **Medical Device**; CLAUDE.md names cochlear implantation as *the* worked case for this pattern; and fifteen KB entries already follow it, not eight. Applied exactly the prescribed shape.

`just check-qualifier-terms-online` — **OK, 325 qualifier term labels match, 97 resolved online**, exit 0. Offline `check-qualifier-terms` also passes (4 labels in this file).

## Suggestion 1 — one-sided differential evidence: taken

A fair hit. The DFNB93 differential argued a contrast in which only the DFNA25 half was cited, and the DFNB93 half — emissions *present* — was asserted. It now carries PMID:40927552 for the Italian family observation, the same citation `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_93` uses, so both halves of the comparison rest on evidence. The cache was already in `main` from that PR, so this adds no new file.

This matters more than a tidy-up, because that contrast is the entry's argument for not filing DFNA25 under auditory neuropathy: emissions present in DFNB93, absent in DFNA25. Having only one side sourced was a real weakness.

## Suggestion 2 — GeneReviews: taken

Closed the loop in `notes`, and confirmed your reasoning rather than restating it: `grep -i "SLC17A8|VGLUT3|DFNA25"` against `references_cache/PMID_20301607.md` returns **0**. So the note says there is no SLC17A8/DFNA25 chapter, and that the adjacent Genetic Hearing Loss Overview is deliberately untagged because tagging it would assert a coverage relationship the cached source does not have.

## Suggestion 3 — noise as an `environmental:` entry: declined

I don't think this one should be taken, and the reason is the same reason it looks attractive. Whether noise exposure explains the absent emissions is the open question in `gap_why_patient_emissions_are_absent`, and the discriminating experiment — noise vulnerability in mice carrying one or two variant alleles — has not been done. An `environmental:` entry with `influences_mechanisms` would assert the exposure acts on a mechanism node, which is precisely what the gap exists to ask. Recording it as an exposure would answer the question in the schema while the prose says it is unanswered.

If the mouse experiment is ever run and comes back positive, that is the moment for an `environmental:` entry with `environmental_effect: EXACERBATES`, and the gap's `supporting_outcome` already says so.

## Suggestion 4 — DR-surfaced supportive care: declined

Agreed with your own "borderline" read. Assistive listening, remote-microphone systems and captioning are generic management for any progressive high-frequency loss; none is DFNA25-specific and none is snippet-supportable from a cached source. Adding them would pad `treatments:` with content that carries no information about this disease.

## Reconciliation note

Thanks for checking the PMID:22841313 / PMID:22794260 pair. Agreed the YAML is right — PMID:22794260 is the striatal dopamine paper, and it appears in the DR's own `off_topic_references`. The entry's `notes` already records that the deep-research run surfaced it and that it was not used.

## Validation on `ccdf391`

All exit 0: `just validate` and `just validate-disorders` — **72/72 snippets verified** (up from 71) — plus `validate-terms`, `check-qualifier-terms`, `check-qualifier-terms-online`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history`. No new term-cache rows were needed; both qualifier CURIEs were already cached.

One process note recorded in the history record, since it nearly shipped: my first attempt at the reference-list insertion anchored on `- reference: PMID:34145196`, which matched inside an *indented* evidence block before reaching the top-level `references:` list and produced invalid YAML. The validator caught it. Anchoring on a line-start pattern with an asserted match count is the fix.

kevinschaper:

PR #10500, opened 2026-09-02
curate: Add Mitochondrial Complex I Deficiency, Nuclear Type 1 (MONDO:0100224, NDUFS4)

### LINKED ISSUE #10490: Curate mitochondrial complex I deficiency, nuclear type 1 (MONDO:0100224)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 1** ([MONDO:0100224](https://monarchinitiative.org/MONDO:0100224)) — biallelic *NDUFS4* variants; loss of an accessory subunit needed for the final assembly and stability of complex I, and the archetypal nuclear-gene cause of Leigh syndrome.

**First decide what this should be, and treat this as the substantive question.** This is the lump/split call, and it points two ways at once.

Toward standalone: the KB already holds `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_6`, `_23`, `_24` and `_34`, so per-subtype MC1DN entries are established convention, and `MONDO:0100224` is bound nowhere in `kb/`.

Toward subtype: `kb/disorders/Leigh_Syndrome.yaml` has a `has_subtypes` block and already carries **NDUFS4** in its `genetic:` section, names it in the top-level description, and cites the `Ndufs4(-/-)` mouse. NDUFS4 disease is classically Leigh syndrome, so this may be coverage rather than a gap.

Check what the Leigh entry actually **binds** rather than what it mentions — a gene named in prose with no bound term and no gene-specific pathograph is coverage in appearance only, which is what the cblA case (#10260) turned on, while the VWM5 case (#10424) went the other way when the umbrella genuinely bound the gene. Read both of those before deciding, compare against how the existing MC1DN siblings relate to their own parent, and defend the choice in the PR body. `SUBTYPE` plus a stub deletion is a completed curation. Do not modify the Leigh entry to make room.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_1.yaml`

kevinschaper:
Adds a dismech entry for **mitochondrial complex I deficiency, nuclear type 1** ([MONDO:0100224](https://monarchinitiative.org/MONDO:0100224), biallelic *NDUFS4*), curated as its own Disease rather than as a subtype of Leigh syndrome. `Leigh_Syndrome.yaml` was not modified.

Closes #10490

## The lump/split call

The issue was right that this points both ways. Here is the argument on both sides, checked rather than assumed.

### What `Leigh_Syndrome.yaml` actually binds

This is **not** the cblA pattern from #10260, where the gene was named in prose with nothing behind it. Leigh genuinely binds NDUFS4:

| | |
|---|---|
| `genetic:` | A dedicated NDUFS4 record with the gene term bound to **`hgnc:7711`**, an `association` sentence, its own `HP:0000007` autosomal recessive inheritance block, and cited evidence |
| `pathophysiology:` | A node "Complex I deficiency" carrying `hgnc:7711` in its `genes` list, alongside NDUFV1 and MT-ND5 |
| description | Names NDUFS4 explicitly |

ClinGen points the same way: its Mitochondrial Diseases GCEP curates **NDUFS4 against Leigh syndrome (MONDO:0009723)**, Definitive — not against MONDO:0100224.

So on the surface this looks like the VWM5 case (#10424), which went to `SUBTYPE`.

### Why it is nevertheless a Disease

Four reasons, in order of weight.

**1. MONDO does not place MC1DN1 under Leigh syndrome.** MONDO:0100224's parents are `mitochondrial complex I deficiency, nuclear type` (MONDO:0100223) and `inborn mitochondrial myopathy` (MONDO:0009637). This is the structural difference from VWM5, where MONDO:0957873 **is** a direct child of the umbrella entry's own term — there, a `has_subtypes` record would have restated a relation the ontology already asserts. Here it would assert a different one.

**2. VWM5 turned on subunit identity not predicting phenotype. Complex I is the opposite.** The VWM5 decision rested on two cohorts (n=83 and n=341) showing that which eIF2B subunit is mutated does not track severity, so there was no gene-specific pathograph to write. In complex I, a review comparing **22 NDUFS4 patients against 198 patients with 24 other complex I gene defects** concluded NDUFS4 disease is uniformly early-onset and severe — while sibling genes already curated here span Leigh syndrome to *isolated optic atrophy* (NDUFA12, NDUFAF8) and PEO-plus to neonatal death (NDUFS8). Gene identity carries prognostic information in this group.

**3. There is a gene-specific mechanism and Leigh does not contain it.** Leigh's node is a generic "Complex I deficiency" shared between NDUFS4, NDUFV1 and MT-ND5, with GO annotations for NADH dehydrogenase activity. Nothing there covers NDUFS4 sitting at the N–Q module interface, the inactive **CI-830 subcomplex**, the reciprocal loss of NDUFA12 with retention of NDUFAF2, or the diagnostic constancy of the native-gel profile. Those are the load-bearing facts of this entity and all of them are absent upstream.

**4. Seven MC1DN siblings already coexist with the Leigh entry.** Types 2, 6, 12, 23, 24, 31 and 34 — most of which also present as Leigh syndrome. Type 23 even lists `Leigh Syndrome` among its `parents:`. Making type 1 the single exception would be the inconsistent choice, and it is the type with the strongest claim to its own file.

### On the ClinGen point specifically

Gene-disease validity curation deliberately lumps to the broadest defensible clinical entity, because the question it answers is whether variants in a gene cause disease at all. It is not a nosological ruling about whether the biochemical entity is separable. The entry records the assertion in `external_assertions` and **says which disease it was made against**, rather than quietly re-labelling it as an MC1DN1 assertion.

## The mechanism

```
biallelic NDUFS4 LOF (protein absent)
            │
            ▼
   loss of the NDUFS4 accessory subunit
   (N–Q module interface)          ──► NDUFA12 also lost; NDUFAF2 retained
            │
            ▼
   assembly arrested at CI-830 (N module never docked)
            │
            ▼
   isolated complex I deficiency  ──► other OXPHOS complexes normal
            │
            ▼
   reductive stress / bioenergetic failure
            │
     ┌──────┴───────────────┬──────────────┐
     ▼                      ▼              ▼
leukocyte-mediated    symmetric        lactic acidosis
neuroinflammation ──► necrotizing      / cardiomyopathy
  (EMERGING)          brainstem +
                      basal ganglia lesions
```

The leukocyte step is curated as a `mechanistic_hypotheses` group with `status: EMERGING`: depleting leukocytes with a CSF1R inhibitor suppresses lesion formation and extends survival in the *Ndufs4* KO, which is a causal claim, but it has no human validation.

## Frequencies

Every band comes from **one denominator** — the 22-patient / 18-family NDUFS4 compilation — and each phenotype's description states its own numerator so it can be checked. Nothing is banded from a broader Leigh series. Worth flagging the ones where the arithmetic matters:

- seizures 4/22 = 18% → `OCCASIONAL`, not FREQUENT (and seizures dominate the terminal phase of some KO lines, so the mouse over-represents this)
- microcephaly 1/22 = 4.5% → `VERY_RARE`
- complex I deficiency in muscle 18/22 = 82% → `VERY_FREQUENT`
- hypotonia 22/22 → `OBLIGATE`

**No `prevalence` record at all.** The available numbers — Leigh incidence around 1 in 40,000, the 116-participant Leigh registry — are a different disease's figures and would be wrong in this file's slot.

**Death in infancy is curated under `progression:`, not `phenotypes:`,** because HP:0001522 sits in the HPO Mortality/Aging branch and is not in the `PhenotypeTerm` enum. That follows the decision already recorded in `Cardiomyopathy-Hypotonia-Lactic_Acidosis_Syndrome`. The 2016 "invariably fatal" conclusion is quoted as the authors' claim and immediately balanced by two later survivors — a girl alive at 7 and an 8-year-old — which is also the subject of a `KNOWLEDGE_GAP`.

## The mouse, and where it stops working

Unusually for this field, the *Ndufs4* KO is a genetically honest model, and the entry says why: human NDUFS4 alleles abolish the protein and the disease is recessive, so a whole-body deletion reproduces the human genotype in kind. The assembly biochemistry — including the NDUFA12 loss and retained NDUFAF2 — was measured the same way in mouse and patient cells.

The neuropathology does not transfer as cleanly, and that is the `HUMAN_MODEL_MISMATCH`. Mouse lesions centre on the **vestibular nuclei, olfactory bulb and cerebellum**; the human disease is brainstem and basal ganglia, with basal ganglia lesions in 9/22 patients. The tempo differs too — well for five weeks, dead two weeks later, against a median patient presenting at four months and dying at ten. Every current therapeutic lead (rapamycin, chronic hypoxia, NAD repletion, leukocyte depletion) was validated by preventing *the mouse's* lesion set.

## Named Entity Confusion

`just preflight-dr` returns **PASS**: NDUFS4 mentioned 55 times, next gene at 5, OMIM 252010 matching MONDO exactly.

But the gate is necessary and not sufficient here, because NDUFS4's literature is largely *Leigh syndrome* and largely *mouse*. Two specific traps were checked:

- **NDUFS4 has a substantial cancer literature** — gastric cancer, glioma, prostate cancer, where it is studied as a proliferation or ferroptosis regulator. Same protein, different context. Not cited.
- **PMID:39858433** (NADH reductive stress as a severity biomarker) looks directly on-topic and is not. Its human arm is a fibroblast cohort of 22 mitochondrial disease patients whose complex I cases are NDUFA1, MT-ND3, MT-ND5 and ACAD9 — **NDUFS4 is not among them**. Its human half is a different disease and it is not cited anywhere in the entry.

## Tooling notes for whoever picks up the next MC1DN

- Two references were fetched and are unusable: **PMID:28371352** and **PMID:30937725** are Finsterer correspondence letters with `content_type: unavailable` and no quotable content. The second argues that early-age *Ndufs4* KO mice are an inappropriate Leigh model and would have been directly relevant to the mismatch discussion. Recorded in the entry's `notes` so it is not rediscovered.
- **`GO:0005747`** (mitochondrial respiratory chain complex I) is **obsolete**; `GO:0045271` (respiratory chain complex I) was used.
- `just clingen-refresh` downloads successfully and then aborts on a drifted checksum pin. The file is fine and usable; this is the documented behaviour, not a broken download.

## Validation

All exit 0:

- `just validate` — 83/83 snippets verified
- `just validate-terms`
- `just validate-disorders` (the batched gate)
- `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`
- `just validate-history`

`check-snippet-grading` caught two real divergences on the first run — one sentence graded two ways after being reused in a second block — and both were fixed by grading the quote once (the patient-MRI sentence as HUMAN_CLINICAL, the mouse-tissue proteomics sentence as MODEL_ORGANISM).


kevinschaper:
Follow-up commit: an audit pass over the entry's numeric prose, checking each figure against a cached reference instead of against memory. Three things needed fixing, all of them mine.

1. **A gene count I could not source.** A differential-diagnosis sentence said Leigh syndrome is produced by "more than seventy-five genes". Nothing in the cached references supports that number — the figure I can quote is "more than 35 genes" (PMID:27671926), and the review literature the deep-research report drew on says "more than 100". The sentence now uses the quotable figure with an evidence item, and says in as many words that higher counts appear in reviews.

2. **Arithmetic presented as a finding.** The rapamycin description said the drug "more than doubled median survival". That was me dividing 114 by 50, not something the paper says. It now gives both figures — untreated median 50 days, treated 114 days in males and 111 in females — each with its own evidence item, so the size of the effect is quoted rather than computed in prose.

3. **A snippet truncated mid-phrase.** One evidence item stopped at "associated with partial" because the sentence continues with a non-ASCII symbol (Δψ) that does not survive a YAML round-trip reliably. A six-word fragment ending on "partial" is not a proposition. It is replaced with the propositional tail of the same sentence, which carries the ROS, NAD(P)H, calcium/ATP-homeostasis and mitochondrial-morphology findings in full; the membrane-depolarisation clause stays in the node description, where it does not need to be an exact quote.

Nothing about the lump/split decision changes.

**Validation after the change** — all exit 0: `just validate` (86/86 snippets verified), `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `validate-history`.

**On the failed `claude-review` check**: its log shows `You've hit your limit · resets 6am (UTC)`, so no review ran on the earlier commit. That is an account quota issue on the credential the workflow used, not something in this PR. This push should re-trigger it.

kevinschaper:
Another commit, on the model sections.

**The iPSC line keeps `experimental_model_type: OTHER`, and now says why.** `IPSC_DERIVED_MODEL` is defined in the schema as a *differentiated* model derived from induced pluripotent stem cells, and this one is deliberately undifferentiated — that is exactly what the authors' own caveat is about, and it is why the entry calls it a redox platform rather than a model of the encephalopathy. `CELL_LINE` is defined for immortalized lines, which iPSCs are not. So `OTHER` is the accurate value rather than a placeholder. Writing that down because it is the kind of thing that otherwise reads as an unfilled slot.

**Added a second experimental model that does take `IPSC_DERIVED_MODEL`**: the NDUFS4-deleted iPSC-derived cardiomyocytes from PMID:35872650, which are differentiated into the affected cell type. It links to the cardiomyopathy node with a NaV1.5-acetylation readout, and its `limitations` record the gap rather than papering over it — the electrical phenotype the model explains is bradyarrhythmia and diastolic dysfunction measured in mice, not the hypertrophy actually recorded in 5 of 22 patients, so it is informative about the cardiac consequences of complex I deficiency without being a model of the finding the node is named for.

**`check-snippet-grading` earned its keep again.** Adding the readout produced a divergence: the NaV1.5 acetylation sentence was graded `MODEL_ORGANISM` in the pathograph and `IN_VITRO` under the cell model. The gate is right — that sentence reports the mouse arm of the study, since the bradyarrhythmia the acetylation is linked to was measured in Ndufs4-null mice. Both are `MODEL_ORGANISM` now, with the reasoning in the `explanation` so the next person does not re-grade it by where it sits.

All gates exit 0: `just validate` (88/88 snippets verified), `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`.

kevinschaper:
Thanks — both blocking findings were right, and checking them turned up a third improvement the review had ruled out. All addressed in `fe4682f`.

## IMPORTANT 1 — GeneReviews: fixed

Verified before acting: `references_cache/PMID_26425749.md` is present, fourteen entries already tag it, and it appeared nowhere in this file. Added to `references:` with the `GeneReviews` tag, plus a `notes` paragraph recording that it is deliberately not cited as evidence because the cached record carries only the chapter's statement of purpose — following the wording `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_2` uses.

## IMPORTANT 2 — two phenotypes: fixed, with two corrections to the specifics

Both added, and you were right that the awkward part is they were already inside a verified snippet — the counts were sitting in the file doing nothing.

**Your HP identifiers were right; the labels were approximate.** Resolved both through OLS before binding, and `term.label` has to be the canonical string:

| Suggested | Canonical label (OLS) |
|---|---|
| HP:0007256 "Pyramidal sign" | **Abnormal pyramidal sign** |
| HP:0000817 "Poor eye contact" | **Reduced eye contact** |

I took the generic `Abnormal pyramidal sign` rather than `HP:0001347` *Hyperreflexia*, because the compilation records "pyramidal signs" as a category and never says which signs — hyperreflexia would be a narrower claim than the source makes. Similarly, the source says *absence* of eye contact where HPO offers only *Reduced* eye contact; I bound the broader term and recorded the wording difference in the description rather than manufacturing a narrower match.

**On the band edge**, I don't think it is ambiguous. The schema fixes `OCCASIONAL` at 5-29% and `FREQUENT` at 30-79%, so 6/22 = 27.3% sits inside `OCCASIONAL` with room to spare. That is also consistent with how this file already bands seizures (18%) and hypertrophic cardiomyopathy (23%). The `explanation` now states the arithmetic and the boundary so it isn't re-litigated. Eye contact at 10/22 = 45% is `FREQUENT` as you had it.

Both are now wired into the pathograph as `downstream` edges of the lesion node rather than left as unconnected list entries — pyramidal signs `DIRECT`, reduced eye contact `INDIRECT_KNOWN_INTERMEDIATES`, since it follows the encephalopathy rather than a specific lesion.

## SUGGESTION 1 — microcephaly: fixed, though the reason is different

You were right that the parenthetical didn't support itself, but the band was fine and the parenthetical was simply wrong. The schema defines `VERY_RARE` as **"Present in rare cases (<5% of patients)"**, not 1-4% — I had quoted HPO's canonical range instead of the schema's. So 4.5% is inside the band, not just outside it. The explanation now cites the schema's actual boundary and adds your substantive point: with n=1 this records where one patient in twenty-two sits, not a stable rate.

## SUGGESTION 2 — AAV gene replacement: this one was supportable after all

> *the research cites them via an uncached source, so there is no quotable snippet to hang them on today*

Worth checking rather than accepting, and it turns out the material is in a cache this PR already commits. `references_cache/PMID_34849584.md` was regenerated to `full_text_xml` (the PMC green-OA text you noted in the scope check), and its intervention section covers the AAV work in detail. So there is now an **AAV-Mediated NDUFS4 Gene Replacement** treatment, `GENE_THERAPY`, targeting the subunit-absence node with `RESTORES` — the only strategy in the entry aimed at the lesion rather than its consequences.

The reason it is worth having is the negative result, which is curated as a `REFUTE` item on the treatment:

> *"However, this vector was absent in the brain and did not ameliorate clinical symptoms."*

Intravenous AAV2/9-hNDUFS4 reached skeletal muscle, heart and liver and did nothing, while AAV-PHP.B delivered at P26-28 extended lifespan and rescued growth, motor coordination, failure to thrive, seizures and cardiac abnormalities. That contrast is what makes this a **delivery** problem rather than a mechanism problem, and the review names the mouse-human blood-brain barrier difference as the obstacle to translating it — which also connects to the BBB finding the deep research surfaced.

**Mitochondrial transfer stays out**, and the reason is now in `notes`: the cached material describes donating healthy mitochondria in a retinal degeneration context with tumorigenicity, immune rejection and genetic instability unresolved, which is too far from an NDUFS4 mechanism node to attach without overstating it.

## Validation

All exit 0 after the change: `just validate` and `just validate-disorders` — **93/93 snippets verified** — plus `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `validate-history`. No new term-cache rows were needed; both new HP terms were already cached.

PR #10507, opened 2026-09-02
curate: Add Spondyloepimetaphyseal Dysplasia Short Limb Abnormal Calcification Syndrome (MONDO:0010077, DDR2)

### LINKED ISSUE #10505: Curate spondyloepimetaphyseal dysplasia-short limb-abnormal calcification syndrome (MONDO:0010077)
Curate a dismech entry for **spondyloepimetaphyseal dysplasia-short limb-abnormal calcification syndrome** ([MONDO:0010077](https://monarchinitiative.org/MONDO:0010077)) — an autosomal recessive skeletal dysplasia (SEMD-SL/AC, Borochowitz–Cormier-Daire type) caused by biallelic loss-of-function variants in *DDR2*, the collagen-activated receptor tyrosine kinase that regulates chondrocyte proliferation and matrix remodelling.

Preliminary read: this looks like a single disease. One gene (`hgnc:2731` DDR2), no MONDO descendants. Worth flagging for the reviewer: `Warburg_Cinotti_Syndrome` is already curated and is also DDR2, but it is the mirror-image allelic disorder — dominant *gain*-of-function (ligand-independent constitutive kinase activation) versus the recessive *loss*-of-function here. Same gene, opposite functional impact, different phenotype, so two entries rather than one, and this is a good worked pair for `functional_impact_category`.

Stub: `stubs/Spondyloepimetaphyseal_Dysplasia-short_Limb-abnormal_Calcification_Syndrome.yaml`

aleixpuigb:
New disorder entry for SMED-SL/AC, the autosomal recessive skeletal dysplasia caused by biallelic loss-of-function variants in *DDR2* (OMIM #271665). Children have short-limbed disproportionate short stature, platyspondyly, abnormal metaphyses and epiphyses, and the premature calcification that gives the disorder its name.

The interesting part mechanistically is that the disease alleles fail in two different ways and the entry curates both. Kinase-domain missense variants and a frameshift misfold and are held in the endoplasmic reticulum, so the receptor never reaches the cell surface. One extracellular allele, p.E113K, reaches the surface perfectly normally and still cannot be activated, because the residue sits in the collagen-binding site. Both routes converge on a single node — failure of collagen-induced receptor activation — and from there on growth-plate chondrocyte proliferation, with a separate branch through the cranial-base synchondroses that accounts for the facial gestalt.

**Same gene, opposite direction, different disease.** `Warburg_Cinotti_Syndrome` is already curated here and is also *DDR2*, but it is dominant gain-of-function. These stay two entries with opposite `functional_impact_category` values, and the pair is now a worked example of why that slot exists. The `notes` block says so explicitly so nobody merges them later.

## What the deep research contributed, and what it got wrong

One openscientist run, committed with the entry. Its citations were clean — 12 of 12 identifiers resolved, the one checked quote valid. Two of its identifiers were wrong and were not used:

- It proposed `MONDO:0009642` as the disease term. That CURIE resolves, so nothing about it looks suspicious, and it is **orofaciodigital syndrome type II**. The correct term came from the curation stub.
- It gave the gene as `HGNC:2968`, which does not resolve. DDR2 is `hgnc:2731`.

Both are recorded in the entry `notes`. `just preflight-dr` passed against MONDO:0010077 with DDR2 mentioned 76 times.

The report earned its place elsewhere: it surfaced the *Ddr2*-deficient mouse work showing the defect is reduced chondrocyte **proliferation** and not aberrant differentiation, the Gli1-positive progenitor conditional knockouts that make the requirement cell-autonomous, the cranial-base study, and the original 1993 clinical description with its histopathology.

- **No phenotype frequency.** The cohorts are small and consanguineous, drawn from founder populations near Jerusalem and in the United Arab Emirates. A percentage would describe who was sequenced.
- **The premature calcification is a `KNOWLEDGE_GAP`.** It is the feature that names the disorder and distinguishes it radiographically, and no published work traces a path from absent DDR2 collagen sensing to early or ectopic mineral deposition. Short stature and abnormal metaphyses follow straightforwardly from a growth-plate signalling defect; this does not.
- **p.Arg124Trp is recorded as unassayed.** It is a second discoidin-domain allele, so it establishes that disease variants occur outside the kinase domain, but nobody has tested whether it fails the same way p.E113K does.
- The optic atrophy reported in one sibling pair is curated with an explicit caveat that it rests on a single family.

## A quoting limitation worth flagging

The founding paper states the entire allelic series in one sentence, and that sentence cannot be used as a snippet: it encloses each variant designation in square brackets (`[R752C]`, `[IVS17+1g > a]`), which `linkml-reference-validator` strips from the query but keeps in the cached text. The claim is therefore evidenced from a bracket-free sentence in a later paper, which means the entry cites a 2014 follow-up for a fact whose primary source is the 2008 discovery paper. Reported on monarch-initiative/dismech#10192, which had already been filed for this; the entry `notes` records the workaround.

## Validation

`just validate-disorders`, `just validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-reference-titles`, `check-empty-snippets`, `just validate-history` — all pass. 52 of 52 snippets verified against cached references.

Closes #10505

---

### DIFF
--- a/history/disorders/Spondyloepimetaphyseal_Dysplasia_Short_Limb_Abnormal_Calcification_Syndrome/2026-09-02T084807Z-claude-code-7ba9c4.yaml
+++ b/history/disorders/Spondyloepimetaphyseal_Dysplasia_Short_Limb_Abnormal_Calcification_Syndrome/2026-09-02T084807Z-claude-code-7ba9c4.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Spondyloepimetaphyseal_Dysplasia_Short_Limb_Abnormal_Calcification_Syndrome
+  path: kb/disorders/Spondyloepimetaphyseal_Dysplasia_Short_Limb_Abnormal_Calcification_Syndrome.yaml
+session:
+  id: 2026-09-02T084807Z-claude-code-7ba9c4
+  timestamp: '2026-09-02T08:48:07Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10505
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10507
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - diagnosis
+  - progression
+  - notes
+  summary: 'Review round 1: term specificity, diagnosis and progression blocks'
+  details: |
+    Response to the automated review on PR #10507, which requested changes on three items. All three were verified against the repository before being applied.
+
+    1. The growth-plate node bound preferred_term 'growth plate chondrocyte' to CL:0000138 chondrocyte. CL:1000217 growth plate cartilage chondrocyte exists, is already in cache/cl/terms.csv, and is what 8+ sibling skeletal dysplasia entries use. Rebound. The second chondrocyte binding, on the cranial-base node, was left as CL:0000138 as the reviewer noted: its preferred_term is plain 'chondrocyte' and the synchondrosis context is broader than a long-bone growth plate.
+
+    2. The short stature phenotype bound HP:0003498 Disproportionate short stature while its own description said short-limbed rather than short-trunked. Rebound to HP:0008873 Disproportionate short-limb short stature, which is cached and used by 8 sibling entries. The limb-versus-trunk distinction is the discriminating one in a short-limb dysplasia.
+
+    3. Added a diagnosis: block with four modalities (skeletal radiographs, chondro-osseous histopathology, DDR2 sequencing, homozygosity mapping) and a progression: block with two phases. No new references were needed: every snippet was already cached, and three of the four diagnosis snippets were already quoted elsewhere in the entry.
+
+    Non-blocking suggestions also addressed: the absence of a treatments block is now explained in notes rather than left silent, and the GeneReviews question the reviewer flagged as unverifiable in its sandbox was resolved directly against PubMed - no chapter exists for this disorder or for DDR2.
+
+    Not done: the spontaneous Ddr2 slie mouse mutant is still not curated, because no reference for it is cached and the deep-research report does not cite one. Phenotype onset descriptors were not added; the sources describe features as congenital in prose without per-phenotype onset statements, and the progression block now carries the timing claim that is quotable.
+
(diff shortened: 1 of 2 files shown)


aleixpuigb:
All three requested changes are in, pushed as 32f2a22. Each was checked against the repository before I applied it, and all three held up.

**1. Growth-plate chondrocyte → `CL:1000217`.** Correct, and the cache already had the term. Eight sibling entries use it (`Achondroplasia`, `Achondrogenesis_Type_II`, `Brachyolmia`, `Atelosteogenesis_Type_I`/`III`, and others). Rebound. I left the cranial-base node on `CL:0000138` for the reason you gave — its `preferred_term` is plain "chondrocyte" and a synchondrosis is not a long-bone growth plate.

**2. Short stature → `HP:0008873`.** Also correct, also cached, also used by eight siblings. This one was a genuine information loss rather than just imprecision: the node's own description said "short-limbed rather than short-trunked disproportion" and the binding threw exactly that away.

**3. `diagnosis:` added,** with four modalities: skeletal radiographs, chondro-osseous histopathology, DDR2 sequencing, and homozygosity mapping. You were right that it cost nothing — no new references, and three of the four snippets were already quoted elsewhere in the file. I also took the `progression:` suggestion, since PMID:8434618 carries two quotable timing claims that turned out to matter more than I first credited: the metaphyseal and epiphyseal changes appear "beyond early infancy", so a newborn radiograph can understate the picture. That is a practical point for anyone using the radiographic criteria, and it was buried in a phenotype snippet.

Snippet count is 58/58 verified, up from 52.

**On the two non-blocking items I could act on:**

- **GeneReviews.** You flagged this as unverified because outbound PubMed is blocked in your sandbox. I could reach it, so I checked rather than leaving it open: `spondylo-meta-epiphyseal dysplasia GeneReviews` and `DDR2 GeneReviews` both return zero results. No chapter exists. Recorded in `notes`.
- **No `treatments:` block.** I agree with your reading and have taken your suggestion of explaining it rather than leaving it silent. `notes` now records that management is supportive, that genetic counselling matters here given the consanguineous ascertainment, and that none of the ten cached references contains a quotable management statement. Thanks for checking the NCIT codes in the research report's §12 — I had not, and all three being wrong is worth knowing.

**Two things I did not do:**

- The spontaneous *Ddr2^slie^* mutant is still uncurated. No cached reference covers it and the deep-research report cites none, so curating it would mean asserting a model I cannot evidence.
- No per-phenotype `onset` descriptors. The sources describe the features as congenital in running prose without stating onset per finding, and the `progression:` block now carries the timing claim that is actually quotable. Adding `onset` to each phenotype would be reading a per-feature claim into a whole-syndrome sentence.

Re-requesting review.

---

PR #10508, opened 2026-09-02
curate: Add Charcot-Marie-Tooth Disease Axonal Type 2P (MONDO:0013753, LRSAM1)

### LINKED ISSUE #10503: Curate Charcot-Marie-Tooth disease axonal type 2P (MONDO:0013753)
Curate a dismech entry for **Charcot-Marie-Tooth disease axonal type 2P** ([MONDO:0013753](https://monarchinitiative.org/MONDO:0013753)) — a late-onset axonal peripheral neuropathy caused by variants in *LRSAM1*, a RING-type E3 ubiquitin ligase whose substrates include TSG101 and which is required for normal endosomal sorting and axonal maintenance.

Preliminary read: this looks like a single disease with one gene (`hgnc:25135` LRSAM1) and no MONDO descendants. Note the inheritance question is the interesting curation problem rather than a lump/split one: both autosomal dominant (frameshift/nonsense variants in the terminal exon, escaping nonsense-mediated decay, acting dominant-negatively on RING-domain dimers) and autosomal recessive (a homozygous splice variant) families are reported, and MONDO's `CMT2P` / `CMT2G` synonyms cover both. That belongs in `inheritance` blocks plus `genetic[].functional_impact_category`, not in separate entries.

Stub: `stubs/Charcot-Marie-Tooth_Disease_Axonal_Type_2P.yaml`

aleixpuigb:
New disorder entry for CMT2P, the *LRSAM1*-associated axonal hereditary motor and sensory neuropathy (OMIM #614436). Clinically it is an ordinary CMT2 phenotype that happens to be mild, late in onset, lower-limb predominant and unusually often painful, with expressivity wide enough that asymptomatic carriers exist and show up only on lower-limb muscle MRI.

## One disease, two inheritance patterns

Dominant and recessive CMT2P are curated as one entry, because the difference between them is a property of the allele rather than of the disease. Pathogenic variants cluster in the 3-prime exons encoding the C-terminal RING domain, and whether the premature termination codon there escapes nonsense-mediated decay decides everything downstream: a surviving transcript makes a truncated protein that poisons the ligase (dominant), a degraded one leaves no protein at all (recessive, needs two alleles). The entry models that as two upstream pathophysiology nodes converging on one ubiquitylation-failure node, with an `inheritance` block for each pattern. Splitting them into two entries would duplicate the whole pathograph to record an allele property.

The convergence node is the best-measured step in the disease: three separate disease alleles were assayed and all lost ubiquitylation activity, specifically through loss of the E2-E3 interaction — while dimerisation and TSG101 binding stayed intact. The lesion is in catalysis, not in substrate recognition, which is worth stating because it rules out a class of therapeutic approach.

## Two things marked as weaker than the rest

**The TSG101/ESCRT step** carries `mechanism_confidence: HYPOTHETICAL` and two `NO_EVIDENCE` items rather than being asserted. TSG101 is the only validated LRSAM1 substrate, and two results make the endosomal reading harder rather than easier: LRSAM1 localises to a post-Golgi perinuclear compartment with little overlap with endosome-to-lysosome machinery, and the p.Cys694Tyr allele leaves TSG101 protein levels unchanged in patient cells. Both are curated rather than omitted.

**The parkinsonism** — three of five affected members of the original Dutch kindred, ages 50 to 65 — sits in its own `EMERGING` hypothesis group and a `HYPOTHESIZED` node, so the peripheral mechanism does not inherit its uncertainty. The authors' own sentence calling it speculation is quoted next to the observation.

The mouse gets a `PARTIALLY_RECAPITULATES` link for axonal vulnerability (mutant axons degenerate readily on acrylamide challenge) and a `FAILS_TO_RECAPITULATE` link for the motor phenotype, which the mice essentially do not develop. That negative is the substance of a `HUMAN_MODEL_MISMATCH` discussion: the mouse models loss of function while most human disease is dominant-negative, so the mild phenotype may reflect the wrong mechanism rather than the wrong species, and nobody has tested which.

## What the deep research contributed, and what it got wrong

One openscientist run, committed with the entry. Its citations were clean: 20 of 20 identifiers resolved, both checked quotes valid, confabulation rate 0.0. Three of its ontology bindings were wrong, and none were used:

| Report offered it as | The ontology calls it |
|---|---|
| `MONDO:0013749` — CMT2P | ventricular septal defect 3 |
| `NCIT:C15224` — pain management | Nutrition Research, Fats |
| `NCIT:C15633` — genetic counseling | Cancer Pain Management |

This is exactly the case the term check exists for, and the entry `notes` records it as such: clean citations say nothing about whether the bindings are right. The correct MONDO came from the curation stub; the treatment terms were selected from the term cache. `just preflight-dr` passed with LRSAM1 mentioned 54 times.

The report earned its place by surfacing the genotype-phenotype paper that states the nonsense-mediated-decay rule, the biochemistry paper with the E2-E3 result and its negative control, the mouse model, and the parkinsonism report.

One smaller trap worth noting: the report writes "areflexia" beside `HP:0001265`, which HPO calls *Hyporeflexia*. This entry binds `HP:0001315` Reduced tendon reflexes and grades it `directness: INDIRECT`, because no cached source states reflex findings in CMT2P patients in so many words.

## Validation

`just validate-disorders`, `just validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `just validate-history` — all pass. 47 of 47 snippets verified against cached references.

Closes #10503

---

### DIFF
--- a/history/disorders/Charcot-Marie-Tooth_Disease_Axonal_Type_2P/2026-09-02T090736Z-claude-code-c7a73e.yaml
+++ b/history/disorders/Charcot-Marie-Tooth_Disease_Axonal_Type_2P/2026-09-02T090736Z-claude-code-c7a73e.yaml
@@ -1,0 +1,43 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Charcot-Marie-Tooth_Disease_Axonal_Type_2P
+  path: kb/disorders/Charcot-Marie-Tooth_Disease_Axonal_Type_2P.yaml
+session:
+  id: 2026-09-02T090736Z-claude-code-c7a73e
+  timestamp: '2026-09-02T09:07:36Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10503
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10508
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - treatments
+  - references
+  - pathophysiology
+  summary: 'Review round 1: pain term rebinding, pain management, GeneReviews references,
+    graph connectivity'
+  details: |
+    Response to the automated review on PR #10508 (three blocking items, eight suggestions). All three blocking items addressed; five suggestions taken.
+
+    1. The Neuropathic Pain phenotype was bound to HP:0009830 Peripheral neuropathy, which does not mean pain and restates the diagnosis every other phenotype is a manifestation of. The stated rationale was wrong: the quoted snippet says 'adult-onset axonal and painful neuropathy', which does support a pain claim. Rebound to HP:0012531 Pain with preferred_term neuropathic pain, verified as a PhenotypeTerm enum member. HP:6000040 Neuropathic pain is not one.
+
+    2. Added a Neuropathic Pain Management treatment. The reviewer's suggested source, PMID:40014417, turned out not to carry the quote - its abstract says symptomatic management is the only option but names no agents. Three sources at three scopes are used instead, each graded INDIRECT with its scope stated in the explanation: PMID:20301462, the CMT2 GeneReviews chapter, for symptomatic treatment of pain (correct scope, no agents); PMID:36793224, a CMT patient cohort, for the pain burden; PMID:20301684, the MFN2 CMT2 GeneReviews chapter, as the only cached source naming agent classes. The last is a different CMT2 subtype and the explanation says so explicitly - it is cited because the recommendation is generic symptomatic management of neuropathic pain in axonal CMT with nothing MFN2-specific in it, not because MFN2 disease is a proxy for CMT2P.
+
+    3. Added a top-level references: block tagging PMID:20301532 (CMT Hereditary Neuropathy Overview) and PMID:20301462 (CMT Neuropathy Type 2) as GeneReviews. Both were already in references_cache from main; neither is used for snippet evidence beyond the pain item, since the Overview cache is a purpose statement only.
+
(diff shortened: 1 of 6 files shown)


aleixpuigb:
All three blocking items are in, pushed as 1865b47. One of your suggested sources didn't hold up, so I want to be explicit about what I used instead.

**1. `Neuropathic Pain` rebound to `HP:0012531` Pain** with `preferred_term: neuropathic pain`. You were right on both halves: the term didn't mean pain, and my stated rationale for it was wrong — the snippet says "adult-onset axonal and **painful** neuropathy", which does support a pain claim. I confirmed `HP:0012531` is a `PhenotypeTerm` enum member before binding it.

**2. Pain management added — but not from `PMID:40014417`.** I fetched it as you suggested and it doesn't carry the quote: the abstract says symptomatic management is the only option and reviews "associated symptoms and comorbidities", without naming an analgesic anywhere. Grepping the cached text for "neuropathic pain", "gabapentin" and "analgesi" returns nothing. So it isn't cited.

Three sources at three different scopes are cited instead, each graded `INDIRECT` with the scope stated in its `explanation`:

| Source | Scope | What it supports |
|---|---|---|
| `PMID:20301462` | CMT2 GeneReviews — correct scope | "symptomatic treatment of pain" in the management section; names no agents |
| `PMID:36793224` | CMT-wide patient cohort | that the problem is real and common, from patients rather than expert opinion |
| `PMID:20301684` | **MFN2** CMT2 GeneReviews — a different subtype | the only cached source naming agent classes |

The third one needs your judgement and I've flagged it rather than buried it. It is the GeneReviews chapter for MFN2-related CMT2, not CMT2P, and its `explanation` says so outright. I used it because the recommendation it makes — TCAs, carbamazepine or gabapentin for neuropathic pain — is generic symptomatic management of axonal CMT with nothing MFN2-specific in it, and because a queryable treatment with no named agent isn't much use to a clinician. If you read that as the same Named Entity Confusion you warned about on the AFD PR, say so and I'll drop the `therapeutic_agent` list and keep the treatment agent-free.

**3. `references:` block added** tagging `PMID:20301532` and `PMID:20301462` as GeneReviews.

**Suggestions taken:**

- **The disconnected parkinsonism component was a real defect** and I'm glad you caught it. It's now a `downstream` edge from the ubiquitylation-failure node, gated on the `EMERGING` group so the edge exists without the uncertainty leaking into the canonical chain.
- **`PMID:40721190` turned out to be stronger than "one more case".** It reports DaT-SPECT showing bilateral presynaptic dopaminergic loss, and states this is the **seventh** description linking LRSAM1 to parkinsonism. That moves the association from a clinical impression in one kindred to a demonstrated nigrostriatal deficit across several, and it supplies the citation "scattered later cases" was missing. Both sentences are now cited on the nigral node.
- Foot drop and gait disturbance added and wired into the pathograph (foot drop → gait via `sequelae`).
- HYPOTHESIZED/HYPOTHETICAL prose fixed; `PMID:40636623` regraded to `OTHER` on both uses.

**Not taken,** all non-blocking: modifier genes (rests on one family's proposal), `progression:`/`clinical_burden:`, and `target_mechanisms` on treatments.

54/54 snippets verified, up from 47.

---

PR #10509, opened 2026-09-02
curate: Add Acrofacial Dysostosis Cincinnati Type (MONDO:0014651, POLR1A)

### LINKED ISSUE #10504: Curate acrofacial dysostosis Cincinnati type (MONDO:0014651)
Curate a dismech entry for **acrofacial dysostosis Cincinnati type** ([MONDO:0014651](https://monarchinitiative.org/MONDO:0014651)) — a severe mandibulofacial dysostosis with variable limb anomalies caused by heterozygous variants in *POLR1A*, the largest subunit of RNA polymerase I. It is a ribosomopathy: reduced rDNA transcription causes nucleolar stress and p53-dependent apoptosis in cranial neural crest cells.

Preliminary read: this looks like a single disease. One gene (`hgnc:17264` POLR1A), no MONDO descendants, and one conserved mechanism. It is mechanistically adjacent to the already-curated `Treacher_Collins_Syndrome` (POLR1C/POLR1D and TCOF1) but is a distinct entity with its own gene and a different limb phenotype, so it should be its own entry rather than a `has_subtypes` row there. Whether the two should later share a Pol I / nucleolar-stress mechanism module is worth noting during curation.

Stub: `stubs/Acrofacial_Dysostosis_Cincinnati_Type.yaml`

aleixpuigb:
New disorder entry for acrofacial dysostosis, Cincinnati type (OMIM #616462), caused by heterozygous variants in *POLR1A*, the largest catalytic subunit of RNA polymerase I. It was defined in 2015 in three individuals whose faces looked like Treacher Collins syndrome and whose hands and feet did not, and expanded in 2023 by 17 more.

## Why a housekeeping defect gives a face-and-limb disease

This is a ribosomopathy, and the entry curates the answer to the standard ribosomopathy question as a chain rather than as an assertion:

1. Reduced Pol I transcription of the 47S precursor rRNA, the rate-limiting step of ribosome biogenesis. Measured directly in the zebrafish mutant: fewer monosomes and polysomes, impaired translation.
2. Tissue-selective vulnerability. Pol I subunits are most highly expressed in neuroepithelium and neural crest, which sustain the highest rRNA transcription and translation rates — so those cells cross the threshold first, and they are also the cells that build most of the craniofacial skeleton.
3. Nucleolar stress. Free RPL5 and RPL11 bind MDM2 and displace it from p53.
4. p53-dependent apoptosis, shown to be cell-autonomous by conditional deletion in three separate lineages (neural crest, second heart field, forebrain).
5. Depletion of neural-crest-derived skeletal precursors in the pharyngeal arches.

## A second, EMERGING hypothesis group

Blocking p53 is not enough, and that is a published result rather than a caveat: inhibiting *tp53* in *polr1a*-mutant zebrafish suppresses the apoptosis and only partially rescues cartilage, because rDNA transcription and neural crest proliferation stay impaired. So part of the deficit is a failure to make enough progenitors, not a failure to keep the ones that were made. It has its own hypothesis group and its own node, and the practical consequence is stated: a p53-directed intervention would by construction miss it.

Three animal models are curated with distinct roles, including a `RESCUES` link on the zebrafish for the *tp53* inhibition experiment — the experiment that makes the p53 node causal rather than correlative, and simultaneously proves it is not the whole disease.

**No phenotype frequency, and this is the judgement call most worth reviewing.** The deep-research report supplies an HPO-annotation-derived frequency table with denominators — hypotonia 10/17, hypertelorism 9/16, micrognathia 9/21, cleft palate 6/20, limb anomaly 6/17, with craniosynostosis explicitly 0/15. Those numbers are plausible and useful. But they are counts computed over HPO annotation records, not statements made in either primary paper, so there is no sentence in a cached reference that a `frequency` value could be evidenced against. Curating them would mean attaching a number to a citation that does not contain it. The table and its source are recorded in the entry `notes` instead, so a curator who later obtains the per-individual tables from the 2023 cohort paper can add them properly.

Three craniofacial phenotypes (micrognathia, malar hypoplasia, cleft palate) are graded `directness: INDIRECT` because their only quotable source describes this disease and Treacher Collins syndrome jointly. That joint attribution is stated in each evidence `explanation` rather than quietly dropped.

## Relationship to Treacher Collins syndrome

Kept as a separate entry, and the `notes` say why: different genes (POLR1A here; TCOF1, POLR1C, POLR1D there), different limb involvement, and Treacher Collins has its own developed entry. They are the same mechanism at the pathway level, though, and if a Pol I nucleolar-stress module is ever created in `kb/modules/`, these two plus the POLR1C/POLR1D forms are its natural conformers. No `conforms_to` is declared, since that module does not exist.

## Deep research

One openscientist run, committed with the entry: 12 of 12 citations resolved, confabulation rate 0.0. `just preflight-dr` passed with POLR1A mentioned 39 times. It flagged TP53 as a rival gene at 31% of POLR1A mentions — expected here, since p53 is the effector arm of the mechanism rather than a second disease entity, and the sections were read with that in mind.

## Validation

`just validate-disorders`, `just validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `just validate-history` — all pass. 38 of 38 snippets verified against cached references.

Closes #10504

---


aleixpuigb:
You were right about the cache, and the error mattered more than it looks. Pushed as 16bd3fd.

`PMID_41010008.md` is `content_type: full_text_xml`, 119 lines, added by this PR, and I consumed it twice. My `notes` then asserted the opposite — that the per-individual clinical detail wasn't cached — and that wrong claim is what justified leaving out the phenotypes, the treatment and the diagnostic pathway. I've corrected the paragraph by recording the correction rather than quietly rewriting it, because the mistaken claim is the reason for the omissions and a silent fix would erase why they happened.

**1. DIRECT evidence added** to Micrognathia and Malar Hypoplasia from the Weaver-cohort sentence. Kept the older INDIRECT items, as you suggested.

**2. Five phenotypes added** — hypertelorism, external ear anomaly, seizure, atrial septal defect, hearing impairment. Three judgement calls worth flagging:

- I bound the ear finding to `HP:0000356` Abnormality of the outer ear rather than `HP:0008551` Microtia. The sources say "dysplastic ears" and "small malformed ears with small canals", which is malformation, not hypoplasia. Microtia would be a stronger claim than the text makes.
- ASD is added as a **sibling** of `Abnormal heart morphology` rather than replacing it, since the cohort statement and the case report support different granularities and both are worth keeping.
- **Hearing loss is the one I'd want you to look at hardest.** You're right that it's the omission most likely to matter clinically. But no cached source reports hearing status in this disease at all — what's cached is the ear anatomy. So I've curated it graded `INDIRECT`, with the description saying explicitly that the evidence supports small malformed ears with narrow canals rather than an audiogram, that it is an anatomically-grounded expectation, and that no frequency should be read into it. If you think that's still over-claiming I'll take it out; I judged that a clinician is better served by a flagged expectation than by silence.

**3. `treatments:` added** with the cardiac-repair quote, plus a `KNOWLEDGE_GAP` on `treatments#` recording why there is nothing else. We reached the same conclusion independently: every management paper filed under "acrofacial dysostosis" is Nager, Miller or Treacher Collins. `PMID:9633164` was not used.

**4. `diagnosis:` added** with the exome-testing sentence and a differentiation entry carrying the craniosynostosis and macrocephaly negatives. That entry deliberately carries **no** evidence item, and says so in its description: those counts are HPO annotation tallies, not statements in either primary paper — the same reason no phenotype here has a `frequency`. Moving them into a structured section made them queryable; attaching a citation that doesn't contain them would have been the thing we both agreed not to do.

All four suggestions taken as well: `mappings:` block; a paragraph on `genetic_context.description` recording that the single `LOSS_OF_FUNCTION` value is a simplification, since the series mixes truncating and CNV alleles with missense in a strongly missense-constrained gene (enum value kept conservative, since no study has tested dominant-negative directly); and the limb sub-features now named and evidenced with the coarse binding retained and justified.

The Pol I nucleolar-stress module is genuinely out of scope here — I'll file it as an issue rather than let it drift into this PR.

49/49 snippets verified, up from 38. All gates pass.

---

aleixpuigb:
Thanks — and the Pol I module is already filed as monarch-initiative/dismech#10512, opened after your first review. It lists this entry, `Treacher_Collins_Syndrome` and the POLR1C/POLR1D forms as conformers, and includes the p53-independent arm as a second hypothesis group.

On the `mappings:` self-map — you're right that it carries no information, and thanks for checking the `Mappings` class rather than leaving me to discover the OMIM slot doesn't exist. I'm leaving it as-is rather than pushing a one-line removal: `main` has `dismiss_stale_reviews`, so a cosmetic push would drop your approval and cost a full re-review cycle for a redundant-but-harmless block. Same reasoning for `diagnosis_term` on the exome-sequencing entry — a genuine improvement, but not worth an approval. Both are on the record here for whoever next touches the file.

One correction to the PR body above, which I can't edit without the same problem: it still says "38 of 38 snippets verified" and describes three craniofacial phenotypes as `INDIRECT`-only. Both were true when it was written and are now stale — the current head is 49/49, and micrognathia and malar hypoplasia carry DIRECT items. The commit message and this thread have the accurate numbers.

---

aleixpuigb:
Merged `main` in to clear a conflict that kept this PR out of the merge queue (the other three from this batch went in; this one showed `mergeable_state: dirty`).

**The conflict was one line in `cache/hgnc/terms.csv`.** This branch added `hgnc:17264,POLR1A`; `main` had since added `hgnc:17258,AGTPBP1` at the same position. Both rows are kept, in numeric CURIE order. No row was retyped or hand-placed: `just check-cache-order` reports canonical order and `just check-term-cache-integrity` passes over all 18 term caches and 24 enum caches.

Resolved by merge commit rather than rebase, so anyone with this branch checked out keeps a valid checkout.

**Post-merge diff against `main` is only this PR's own files** — the entry, its five reference-cache files, two history records, the deep-research report and artifacts, the stub deletion, and three cache rows. No unrelated deletions and no derived-path churn.

**Re-validated after the merge, not before it:**

- `just validate-disorders` — schema, terms and references, **49/49 snippets verified**
- `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values` — all OK

One correction to the PR body above: it still says "38 of 38 snippets", which was the count at open. The review round added phenotypes, treatments and the diagnosis section, so the current figure is 49.

This push drops the earlier approval, since `main` runs `dismiss_stale_reviews`. That is the expected cost of resolving the conflict rather than a new problem, and the review re-runs on `synchronize`.

---

PR #10510, opened 2026-09-02
curate: Add Platelet-type Bleeding Disorder 12 (MONDO:0011588, PTGS1)

### LINKED ISSUE #10502: Curate platelet-type bleeding disorder 12 (MONDO:0011588)
Curate a dismech entry for **platelet-type bleeding disorder 12** ([MONDO:0011588](https://monarchinitiative.org/MONDO:0011588)) — an inherited mild mucocutaneous bleeding disorder caused by deficiency of platelet prostaglandin-endoperoxide synthase 1 (cyclooxygenase-1, PTGS1), which blocks arachidonate conversion to thromboxane A2 and so impairs the amplification loop of platelet aggregation and secretion.

Preliminary read: this looks like a single disease. MONDO records no descendants for the term, and the phenotype is one conserved mechanism (loss of platelet COX-1 activity → absent thromboxane A2 → defective secondary aggregation and dense-granule release). It sits alongside the already-curated sibling platelet-function entries `Platelet-type_Bleeding_Disorder_8/18/20` and `Gray_Platelet_Syndrome`, and the `Inherited_Platelet_Function_Disorders` grouping already names BDPLT12 in its description text without an entry existing for it.

Stub: `stubs/Platelet-type_Bleeding_Disorder_12.yaml`

aleixpuigb:
New disorder entry for BDPLT12, platelet cyclooxygenase-1 deficiency (OMIM #605735), known clinically as the aspirin-like defect. The nickname is exact rather than loose: COX-1 is precisely the enzyme low-dose aspirin acetylates irreversibly, so a constitutional deficiency reproduces that pharmacology permanently in someone who has never taken any.

## The mechanism is anchored on enzymology, not genetics

That is deliberate, and it is where the evidence actually is. The 1996 biochemical study localised the block to one step with a bypass experiment, and the entry curates it as three evidence items on the node it establishes: arachidonate-induced aggregation and arachidonate-driven thromboxane production are both defective, while aggregation and thromboxane production driven by **exogenous prostaglandin H2** are normal. Everything downstream of COX-1 works; thromboxane synthase blots were normal in all three patients. The same study's western blots give two routes into that node — undetectable COX-1 protein in two patients, normal protein with impaired catalytic activity in the third — curated as two routes within one node rather than as two diseases.

Why the bleeding is mild follows from what thromboxane A2 is: an autocrine amplifier, not an initiator. Removing it does not stop a platelet adhering or aggregating to a strong agonist; it removes the positive feedback that consolidates a plug. So adhesion and ristocetin agglutination are preserved, and the aggregometry signature is an absent arachidonate response with a lost secondary wave and a tendency to disaggregate.

## Module conformance, and the arm it does *not* attach at

Conforms to `primary_hemostatic_plug_failure` at `#Impaired Platelet Activation, Granule Secretion, and Integrin Inside-Out Signalling`, which is the arm that module's own description assigns to cyclooxygenase-1 defects, plus the central plug-formation node and the mucocutaneous bleeding output.

It deliberately does **not** attach at the integrin αIIbβ3 aggregation arm, even though abnormal aggregometry is this disorder's diagnostic hallmark. That is the mistake the module explicitly warns against — almost every disorder in the series has abnormal aggregometry, because every upstream arm reads out through it — and what is preserved here is what keeps this entry off that arm.

- **No phenotype frequency.** The one quantified figure, bleeding symptoms in 25 of 34, comes from a cohort defined by an aggregometry threshold in children referred for suspected coagulation disorders, with relatives added by family testing. It describes an ascertained group, not the disease. It is recorded with its denominator in prose instead.
- **The 0.6% figure is a referral yield.** It is curated as a prevalence record whose `population` is stated as the paediatric referral group, with a note that the denominator is children already suspected of a coagulation disorder.
- **The genetic evidence is thinner than the biochemistry, and the entry says so.** Reported alleles are largely private, ClinVar holds no established recurrent pathogenic PTGS1 entry for this trait, and a `KNOWLEDGE_GAP` records that nobody has sequenced a biochemically defined cohort — so a clinician deciding whether a negative panel excludes the diagnosis has no denominator to reason from.
- Three mucocutaneous phenotypes are graded `directness: INDIRECT`, because the cohort study reports bleeding in aggregate without itemising it in the cached text.

Desmopressin is curated with direct rather than class-level evidence: the cited study found it normalised the bleeding time in a thromboxane synthesis defect and **not** in thrombasthenia or Bernard-Soulier syndrome, which makes it specific to where this lesion sits.

## Deep research

One openscientist run, committed with the entry. Worth flagging for anyone maintaining the research tooling: **the run exited non-zero after the report had already been produced.** Term validation hit a 5-second read timeout resolving `MeSH:C567786` against EBI OLS and aborted the recipe with exit code 3, ~18 minutes into a completed job. The report file survived; both validations were retrofitted in place with `just validate-research-reference` and `just validate-research-terms`, the latter succeeding on an immediate retry, which is what identifies it as a transient timeout rather than a real resolution failure. Reference validation: 10 of 10 resolved, 9 on topic.

`just preflight-dr` returned `SKIP` because MONDO records no causal gene for MONDO:0011588 and so the gene-identity check cannot discriminate. The manual fallback was applied and passes: OMIM 605735 matches, and PTGS1 is the top gene at 31 mentions.

## Validation

`just validate-disorders`, `just validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-environmental-evidence`, `check-empty-snippets`, `check-reference-titles`, `just validate-history` — all pass. 39 of 39 snippets verified against cached references.

Closes #10502

---

### DIFF
--- a/history/disorders/Platelet-type_Bleeding_Disorder_12/2026-09-02T085947Z-claude-code-f2cabc.yaml
+++ b/history/disorders/Platelet-type_Bleeding_Disorder_12/2026-09-02T085947Z-claude-code-f2cabc.yaml
@@ -1,0 +1,46 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Platelet-type_Bleeding_Disorder_12
+  path: kb/disorders/Platelet-type_Bleeding_Disorder_12.yaml
+session:
+  id: 2026-09-02T085947Z-claude-code-f2cabc
+  timestamp: '2026-09-02T08:59:47Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10502
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10510
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - biochemical
+  - treatments
+  - pathophysiology
+  - prevalence
+  - animal_models
+  - notes
+  summary: 'Review round 1: evidence regrading, biochemical section, prevalence normalization'
+  details: |
+    Response to the automated review on PR #10510 (four blocking items, six suggestions). All four blocking items addressed; five of six suggestions taken.
+
+    1. The mouse aggregation readout asserted direction: DECREASED while quoting the methods sentence describing how the mice were built. Verified the reviewer's claim: the results sentence is in the cache. Swapped. The methods sentence was not discarded - it moved to the link-level evidence, where a statement about how the model was constructed is what belongs.
+
+    2. The link evidence quoting the paper's framing statement about low-dose aspirin in people was graded MODEL_ORGANISM. Regraded OTHER with directness: INDIRECT. evidence_source classifies the study behind the quoted sentence, and that sentence is about human pharmacology, not the mouse work.
+
+    3. Added a biochemical: block for arachidonate-stimulated platelet thromboxane B2, with a DIAGNOSTIC readout on the thromboxane node and a reference_ranges entry carrying the 49-81 ng/3x10^8 platelet interval. The interval is explicitly annotated as a single 1996 laboratory's range rather than a consensus interval, because assay methods have changed. Urinary 11-dehydro-thromboxane B2 is named in block notes as the other confirmation route and explicitly not curated, since no cached reference states it for this disorder.
+
+    4. HP:0002584 gastrointestinal haemorrhage and HP:0005261 joint haemorrhage were not added. Checked all ten cached references: neither finding appears in any of them, so both would be uncitable. Argued out in notes with a second, independent reason for haemarthrosis specifically - joint bleeding is the signature of a coagulation-factor deficiency, and curating it in a disorder whose mechanism is failure of primary rather than secondary haemostasis would assert something this entry's own pathograph contradicts.
+
+    Suggestions taken: two PMID:33326144 background sentences regraded IN_VITRO to OTHER; rate_per_100000: 600.0 dropped and prevalence_class set to UNKNOWN, because normalizing a referral-cohort yield would be read downstream as a population prevalence and is wrong by orders of magnitude; the genetic_context block now states explicitly that its zygosity and functional-impact values describe the modal dominant allele class and not the recessive p.Trp322Ser allele cited in the same node; megakaryocyte CL:0000556 added to the COX-1 deficiency node; recombinant activated factor VIIa added as a treatment, flagged as the least directly supported of the four since its trial evidence is from Glanzmann thrombasthenia.
+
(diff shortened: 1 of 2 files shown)


aleixpuigb:
All four blocking items are in, pushed as 9930625, plus five of the six suggestions.

**1. Readout snippet swapped.** You were right — it quoted the methods sentence describing how the mice were built while asserting `direction: DECREASED`. The results sentence was already cached. I didn't discard the methods sentence: it moved up to the link-level evidence, where a statement about model construction is what belongs.

**2. Aspirin sentence regraded** to `OTHER` with `directness: INDIRECT`. It is the paper's framing about low-dose aspirin in people, not a result of its mouse work.

**3. `biochemical:` added** for arachidonate-stimulated platelet thromboxane B2, with a `DIAGNOSTIC` readout on the thromboxane node and the 49–81 ng interval as a `reference_ranges` entry. One caveat is written into the range's `notes`: it is one 1996 laboratory's reference interval, not a consensus one, and platelet thromboxane assays have changed since — so it should not be carried across to a modern result. Urinary 11-dehydro-thromboxane B2 is named in the block notes as the other confirmation route and explicitly **not** curated, since no cached reference states it for this disorder.

**4. HP:0002584 and HP:0005261 argued out rather than added,** which is the option you offered. Two reasons, and the second is the substantive one:

- Neither appears in any of the ten cached references. The HPO annotations are annotation records, not statements in a paper — the same reason the AFD entry declines to curate its HPO frequency table.
- **Haemarthrosis is mechanistically wrong for this disease**, not just uncited. Joint bleeding is the signature of a coagulation-factor deficiency. This disorder impairs amplification of the *primary* platelet plug, which gives mucocutaneous and post-procedural bleeding. Curating it would assert something this entry's own pathograph contradicts. You flagged that it "would be reasonable to argue it out" — I think it's more than reasonable, and if a source does turn up it should be curated with that tension stated rather than added silently.

**Suggestions taken:** background structural sentences regraded `IN_VITRO` → `OTHER`; megakaryocyte `CL:0000556` added; rFVIIa added and flagged as the least directly supported of the four treatments, since its evidence base within the class is Glanzmann thrombasthenia; `genetic_context` now states explicitly that its single zygosity and functional-impact values describe the modal dominant allele class and not the recessive p.Trp322Ser allele cited in the same node.

**On the prevalence rate — I took the stronger version of your suggestion.** `rate_per_100000: 600.0` is dropped entirely and `prevalence_class` is now `UNKNOWN`. Your point is exactly right: the normalized value would be consumed structurally as a population prevalence, and 600/100,000 for an ultra-rare disorder is wrong by orders of magnitude. The 0.6% figure and its referral denominator stay in `notes`, along with a line stating why no rate is set, so the next curator doesn't helpfully "fix" the missing field.

**Not taken:** congenital onset (`HP:0003577`). No cached reference states onset for this disorder — declined for the same reason as the two phenotypes above.

43/43 snippets verified, up from 39. `check-snippet-grading` re-run clean after the regrades.

---

PR #10513, opened 2026-09-02
Add Bgee cross-species model fidelity check and integration proposal

mellybelly:
## Summary

Adds a prototype curation aid (`scripts/bgee_model_fidelity.py`) that checks whether a model organism's ortholog of a disease gene is expressed in the anatomy where the human mechanism operates — a necessary (never sufficient) condition for model fidelity. Accompanies this with a detailed integration proposal (`docs/bgee-integration-proposal.md`) that frames three possible uses of Bgee data in dismech, with careful reasoning about why expression calls cannot be imported as evidence.

## Key Changes

- **`scripts/bgee_model_fidelity.py`** (545 lines): Implements a cross-species expression check pipeline:
  - Maps HGNC gene IDs → Ensembl human genes (HGNC REST)
  - Resolves orthologs in model species (Ensembl Compara REST)
  - Queries Bgee's `expression_comparison` endpoint for homologous-anatomy expression calls
  - Intersects results with the target node's UBERON/CL anatomy bindings
  - Produces verdicts (`DIVERGENT_ABSENT`, `ORTHOLOG_NOT_1TO1`, `CONSERVED`, etc.) with structured output (TSV, findings-only filtering)
  - Includes a resumable JSON cache to avoid re-querying APIs
  - Handles 30+ animal model species with explicit species→Ensembl/taxon mappings

- **`docs/bgee-integration-proposal.md`** (462 lines): Comprehensive design document covering:
  - Why Bgee present calls cannot be imported as evidence (healthy wild-type only; propagation asymmetry; near-vacuous support)
  - Three proposed integrations: (A) cross-species model fidelity [prototype built], (B) absent calls as negative evidence [design sketch], (C) curation-time QC [design sketch]
  - Detailed measurements from live API calls (cell-type coverage 4%, 79% of links not evaluable due to anatomy mismatch)
  - Seven concrete questions for the Bgee team (cell-type resolution, API stability, orthology handling, absent-call retrieval, SPARQL access, citable records, upstream computation)
  - Two questions for Monarch/dismech governance (direct vs. KG consumption, KGX export scope)

- **`mkdocs.yml`**: Added link to the integration proposal in the documentation nav

- **Ortholog lookup failures are never cached** — a transient rate-limit is not evidence of absence. Failed fetches raise rather than being cached as empty, preventing poisoning of the cache across re-runs.

- **`ORTHOLOG_LOOKUP_EMPTY` is deliberately not a finding** — Ensembl REST returned empty ortholog lists for PTEN and VHL against mouse (both directions, HTTP 200), yet both genes plainly exist. The verdict prompts manual verification rather than acting as a fidelity caveat.

- **Cell-type fallback is opt-in** (`--tissue-fallback`) — when a CL-bound node has no Bgee multi-species condition, re-tries against that node's own UBERON locations at explicitly-labelled coarser granularity. This is sound for this check but is reported with `granularity=TISSUE` to flag the loss of resolution.

- **Verdicts are split into findings vs. non-findings** — only `ORTHOLOG_NOT_1TO1` and `DIVERGENT_ABSENT` are actionable; `CONSERVED` is not treated as validating a fidelity grade (shared expression is a precondition, not evidence).

- **Full run statistics from 260 links**: 158 `ANATOMY_UNMATCHED_CL` (79% not evaluable), 31 `CONSERVED`, 17 `ORTHOLOG_NOT_1TO1` (findings), 0 `DIVERGENT_ABSENT`. The zero absent-call verdicts surface a concrete open question for the Bgee team.

## Usage

```bash
uv run python scripts/bgee_model_fidelity.py --limit 20
uv run python scripts/bgee_model_fidelity.py --file kb/disorders/Duchenne_

https://claude.ai/code/session_01EMuHNod5tdBxwwfUYsWfR3

### DIFF
--- a/docs/reports/scfair-dismech-schema-comparison-2026-09-02.md
+++ b/docs/reports/scfair-dismech-schema-comparison-2026-09-02.md
@@ -1,0 +1,249 @@
+# scFAIR and dismech: a schema comparison, and what LinkML would buy both
+
+*Analysis date: 2026-09-02. Author: AI-assisted (Claude Code). All schema
+structure read from source, not from documentation summaries.*
+
+## Question
+
+The [scFAIR consortium](https://sc-fair.org/) (EPFL Deplancke lab + SIB) publishes
+a single-cell metadata schema that extends CZ CELLxGENE's, and — unusually for a
+metadata standard — a **structured description of computational analysis
+workflows**. dismech independently evolved its own analysis-provenance model for
+hypothesis exploration runs. This report compares the two, and asks what LinkML
+modeling would give each.
+
+Companion to [`bgee-integration-proposal`](../bgee-integration-proposal.md), which
+covers the data-layer question (Bgee expression, ASAP's dataset catalogue)
+separately. **This report is about schemas, not data.**
+
+## TL;DR
+
+- The two schemas overlap almost exactly on one thing — **recording how a
+  computational result was produced** — and they are complementary rather than
+  redundant. scFAIR models *the happy path in mechanical detail* (docker digests,
+  conda environments, random seeds, resource usage). dismech models *failure and
+  unverifiability* (execution status, auditability, whether a data source was
+  actually accessed). Neither has the other's half.
+- **Neither side's analysis-provenance model is formally modeled.** scFAIR's
+  schema is prose Markdown plus a 1,798-line imperative validator config.
+  dismech's `MANIFEST.yaml` is validated by hand-written Python and has no schema
+  file at all. dismech's *assessment* sidecar is LinkML; its *execution manifest*
+  is not.
+- scFAIR's `rules.yaml` contains constructs that are, literally, LinkML features
+  written out by hand — a dynamic enum, a conditional rule, a structured pattern,
+  a mixin. The clearest case: `assay_ancestor_terms: ['EFO:0030080', …]` plus
+  ancestor-checking code **is** `reachable_from`, which dismech already uses for
+  its `AssayTerm` enum rooted at `OBI:0000070`.
+- The concrete offer to scFAIR is a LinkML rendering of their schema. The
+  concrete ask is their environment-capture vocabulary. Neither requires the
+  other to adopt the other's model wholesale.
+
+## What each side actually is
+
+### scFAIR
+
+| Artifact | Form | Size |
+|---|---|---|
+| [`scFAIR_schema`](https://github.com/scFAIR/scFAIR_schema) `schema/7.1.0/schema.md` | prose Markdown, HTML tables | 1,906 lines |
+| `schema_analysis_json.md` | prose Markdown | 1,324 lines |
+| `schema_atac.md` / `schema_spatial.md` / `schema_perturb.md` | prose Markdown | 510 / 356 / 301 |
+| [`asap_web` `config/scfair/7.1.0/rules.yaml`](https://github.com/DeplanckeLab/asap_web/blob/main/src/config/scfair/7.1.0/rules.yaml) | validator rule config | 1,798 lines |
+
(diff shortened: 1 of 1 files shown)


mellybelly:
@cmungall @gingin77 @DnlRKorn — review requested. This is a proposal document rather than code, so the useful read is the rendered markdown, not the diff:

- [**docs/bgee-integration-proposal.md**](https://github.com/monarch-initiative/dismech/blob/claude/bgee-gene-expression-ngtxwk/docs/bgee-integration-proposal.md) — the main document
- [docs/reports/scfair-dismech-schema-comparison-2026-09-02.md](https://github.com/monarch-initiative/dismech/blob/claude/bgee-gene-expression-ngtxwk/docs/reports/scfair-dismech-schema-comparison-2026-09-02.md) — companion, schemas rather than data

## What it says

We should **not** import Bgee expression calls as evidence. Bgee curates healthy wild-type samples only, so a call describes the normal baseline a mechanism departs from rather than the mechanism; and Bgee propagates present calls *up* the anatomy hierarchy and absent calls *down*, so a present call is safe to generalize and unsafe to specialize — which is backwards for us, since our mechanism nodes are almost all cell-type-level.

Instead the proposal picks three narrower uses, and there is a working prototype (`scripts/bgee_model_fidelity.py`) for the headline one: checking whether a model species' ortholog is actually expressed where the human mechanism operates. We have 773 `HUMAN_MODEL_MISMATCH` discussions and 138 `FAILS_TO_RECAPITULATE` links that are currently prose judgements with no data behind them.

## Two results that do not flatter the approach

Both are stated prominently in the document rather than buried, and they are the main thing worth your judgement:

- **79% of links are not evaluable** (205 of 260). Almost always because the node's cell type has no Bgee multi-species condition — only about 4% of those conditions carry a CL term at all.
- **`DIVERGENT_ABSENT` never fired**, across all 260 links. That is the verdict that would be the strongest evidence for a curated model failure, and it has never triggered. Proposal B rests on a datum we have not seen the API produce at scale.

## What would help most

**@cmungall** — the boundary question is yours: Monarch KG already ingests Bgee, so should dismech consume this signal through the KG rather than calling Bgee directly? Our instinct is direct, because the KG's gene→anatomy edges do not carry the cross-species homology comparison or the present/absent/no-data breakdown the check turns on. But that is exactly the "don't reimplement Monarch" line, so it should be an explicit call. Related: does proposal B (atlas-derived negative evidence) need a decision-register entry under §6?

**@gingin77 @DnlRKorn** — the substance, from whichever angle is yours. Chiefly: is the model-fidelity check a sound thing to build at all, and are we simply querying Bgee wrong? The cell-type coverage we measured is poor enough that "you are looking at the wrong endpoint" is a live possibility, and we would rather hear it now than after curators start working a worklist.

Not asking for approval of the prototype's code — it is a curation aid that emits a worklist, and by design it never writes an evidence item.

One caveat: three bugs were found and fixed in the prototype mid-way, all in the orthology step (most visibly, it falsely reported no mouse ortholog for PTEN and VHL). The numbers above are from the corrected run.

---

gingin77:
[human]
A note while reading through the issue - in my opinion, understanding the "the normal baseline a mechanism departs from" can be useful for toxicity testing and biomedical research, because a well-characterized baseline serves as a reference for mapping what is known vs what is understudied.

To the question, _should we build a model-fidelity check at all_, I say yes. In the toxicity space, the common phrasing is 'context of use' - as in, a specific NAM should be chosen to match the context of use. Various models and/or screening tools might be good for one context of use, but not for another. A use case I hold is differentiating experimental models that are needed to advance our understanding of complex disease vs a minimal model needed for chemical safety screening based on the best of what we already know about a complex disease.

I need more time to read the markdown files, but wanted to give my first impression.

PR #8880, opened 2026-08-19
curate: Add Visceral Heterotaxy 9 (HTX9, MONDO:0030070, MNS1)

sierra-moxon:
Closes #8762

Adds `kb/disorders/Visceral_Heterotaxy_9.yaml` — heterotaxy, visceral, 9, autosomal, with male infertility (HTX9), MONDO:0030070 / OMIM:618948.

## NEC preflight — this one actually fired

MONDO:0030070 carries **no definition text**, so identity was anchored on OMIM:618948 and the MONDO `RO:0004003` causal-gene relationship:

```
MONDO:0030070  RO:0004003  HGNC:29636  ...  has material basis in germline mutation in  MNS1
```

**Verified causal gene: MNS1 (hgnc:29636).**

A Falcon deep-research report was then generated and **discarded whole**:

```
FAIL  research/Visceral_Heterotaxy_9-deep-research-falcon.md
  intended entity : MONDO:0030070 heterotaxy, visceral, 9, autosomal, with male infertility
  canonical gene  : MNS1
  mentions        : MNS1=0
  top genes       : MMP21=62, AR=3, HP=2, MMP11=2, SI=2
```

The report had resolved "Visceral Heterotaxy 9" to **MMP21-related heterotaxy 7** (MONDO:0014762, OMIM:616749) and opened with *"Visceral heterotaxy 9 (VH9) is a rare, congenital, autosomal-recessive laterality disorder caused by biallelic pathogenic variants in MMP21"*. It is the textbook NEC case from CLAUDE.md §2b: internally coherent, all four citations resolved (`confabulation_rate: 0.0`), and its quotes would have validated as exact substrings. Only the semantic gene check caught it.

Per the SOP the report was discarded rather than cherry-picked, and it is **deliberately not committed** under `research/` so no later curator can pick it up under its misleading filename — likewise the four MMP21-paper DOI cache files its validation pass had fetched. Everything in this entry comes from PubMed-sourced MNS1 primary literature plus one ClinGen structured record. This is recorded in the entry `notes:` and the history record.

## Scope decision: standalone entry

[...]

PR #10528, opened 2026-09-02
Augment Visceral Heterotaxy 9 (MNS1) with a CFAP53 redundancy hypothesis and a gene-environment entry

tannerzhang:
This augments the existing `Visceral_Heterotaxy_9` entry rather than creating one; the disease was already curated in #8880 and sat at 96% compliance. The change is purely additive to the YAML (335 insertions, no deletions), and none of the earlier curator's scope decisions were altered. The substantive additions are a mechanistic hypothesis, the entry's first `environmental:` section, and evidence for a causal edge that had none.

I have replaced the auto-generated description; two lines in it were inaccurate and are corrected below.

## The Named Entity Confusion check is the thing to look at first

This disease produced this repository's textbook wrong-disease report. In #8880 a Falcon report was discarded whole because it had resolved "Visceral Heterotaxy 9" to MMP21-related heterotaxy 7: it never mentioned MNS1 and discussed MMP21 62 times, while being internally coherent with citations that all resolved.

The report in this PR **passes** that check. MNS1 is mentioned 58 times and is the top gene by frequency.

```
PASS  research/Visceral_Heterotaxy_9-deep-research-openscientist.md
  canonical gene  : MNS1
  mentions        : MNS1=58
  top genes       : MNS1=58, ODAD1=7, MIP=5, DNAH5=5, HP=4
```

## Provider fallback, and why the report is named for OpenScientist

Falcon was requested, as `/curate` defaults to it, and returned HTTP 403 on two attempts. Rather than swapping providers by hand, the run was re-issued with `--fallback`, so the report records `fell_back: true`, `requested_provider: falcon`, and the full attempt chain including the auth error. OpenScientist wrote it, and the file is named for OpenScientist. Anyone reading the entry's `notes:` should know this file is **not** the discarded MMP21 report, and no Falcon report for this disease should exist on disk.

## Correcting two lines in the original description

The auto-generated text said term validation "identified 3 mislabeled terms ... flagged for review". That reads as though the entry contains bad bindings. It does not. Two of those identifiers resolve but name entirely different concepts, and both were **rejected and never bound**:

| Report offered it as | What the identifier actually is |
|---|---|
| `HP:0011885` "Abnormal visceral situs" | Hemorrhage of the eye |
| `UBERON:0004341` "left-right organizer / node" | primitive streak |

The third, `MONDO:0030070`, is just the report calling it "MONDO" in prose, which is harmless. Both rejected terms are recorded in the entry's `notes:` so a later curator does not pick them up.

The original text also said evidence items use "directness classifications (INDIRECT ..., IN_VITRO ..., MODEL_ORGANISM ...)". Those are two different slots: `directness` records how directly a quote bears on its claim, while `evidence_source` classifies the cited study.

## What was added, and where it came from

The deep-research report corroborated the existing entry rather than extending it, so the new content comes from PubMed searches made independently of it. All three papers were previously uncited:

- **PMID:39007638** shows MNS1 and CFAP53 are paralogous filamentous microtubule inner proteins that act redundantly in (9+2) cilia. This grounds a new `mechanistic_hypotheses` entry offering a mechanistic account of why the airway is largely spared while laterality fails. The report never mentions CFAP53.
- **PMID:24876927** supplies the MFN2/MNS1 interaction and MNS1's presence in detergent-resistant flagellar structures.
- **PMID:30129265** shows Mns1 dosage interacts with gastrulation-stage alcohol, raising defect rates from 41% in wild-type to 64% in heterozygotes and 92% in nulls. This is the basis of the entry's first `environmental:` section.

The entry's central causal edge, nodal cilia motility failure to randomization of the left-right axis, previously carried no evidence and now carries two quotes that state the causation directly.

- **No `datasets:` block.** `just discover-datasets` found no GEO candidates, which is the expected result for a disease with fewer than twenty reported individuals.
- **No gnomAD or UniProt content.** These are the report's only genuinely novel contributions, and neither has a citable reference whose abstract carries the numbers.
- **The environmental link is `MODULATES`, not `EXACERBATES`.** The evidence is murine, the potentiated phenotype is craniofacial rather than laterality, and the interaction appeared in heterozygotes too, so it describes Mns1 dosage rather than the biallelic state that defines HTX9.

## The hypothesis has a known hole, stated in the entry

Reviewers should weigh this. The redundancy model explains the human airway sparing but does not explain two things, and both are written into the entry rather than smoothed over:

1. The sperm flagellum is itself (9+2) and should be buffered on this model, yet it fails most penetrantly. This is a `KNOWLEDGE_GAP` discussion with a decisive proposed experiment.
2. Mns1 **single** nulls lose tracheal outer dynein arms and develop hydrocephalus (PMID:22396656). Tracheal and ependymal cilia are both (9+2), so on a strict reading neither should fail.

The hypothesis is `EMERGING` for these reasons. A reviewer could reasonably judge it too speculative to carry at all, and that is a fair call to make.

## Review round already applied

A red-team pass before the PR found no fabricated snippets, wrong PMIDs, ontology errors or broken references, and raised ten prose overreach findings, all fixed in the second commit. The two that mattered: a double-mutant result had been explained as evidence for the single-mutant lesion, and the environmental link asserted a nodal-cilia convergence the cited authors explicitly decline, since their own discussion prefers primary (non-motile) cilia and Shh. That link now carries a `REFUTE` item holding the authors' reading so it cannot read more settled than its source.

## Validation

| Check | Result |
|---|---|
| `just validate-disorders` | passed (schema, terms and references, batched, what CI runs) |
| `just count-verified-snippets` | 98/98 verified |
| `just preflight-dr ... MONDO:0030070` | PASS, MNS1=58 |
| Report reference validation | 18/18 resolved, 0 unresolved, 0 off topic |
| Offline gates | enum-values, causal-targets, entity-refs, qualifier-terms, duplicate-keys, folded-hyphens, snippet-length, title-snippets, snippet-grading, environmental-evidence all OK |

Nothing was added to any baseline. Reference cache files were created only by tooling; none was hand-written.


tannerzhang:
Fixed in 7c7c23dd1. The IMPORTANT finding was correct and the error was mine, introduced when I rewrote that description in the pre-PR review pass. I also took all three suggestions, and one of them turned out to be larger than reported.

**The experimental group was wrong.** I checked the cached full text before editing, and the paper attributes the dissociation to the vehicle-treated nulls exactly as the review says:

> loss of Mns1 did not affect the incidence of eye defects or other craniofacial abnormalities in vehicle-treated animals. However 5/8 of these fetuses exhibited situs inversus, demonstrating disruption of motile cilia function.

Writing "alcohol-exposed" there did more than misname a group. The whole point of the sentence was that motile-cilia disruption can occur without craniofacial consequence, and the alcohol-exposed nulls had 92.31% defects, so my version asserted the dissociation using the one group that disproves it. The description now names the vehicle-treated group and carries the 5/8 figure.

**Retired-enum prose: there were four, not one.** The review flagged the ICSI explanation at line 1933. Searching the whole file found four explanations still saying "Recorded as PARTIAL", all pre-existing and none introduced by this PR:

| Location | Claim being qualified |
|---|---|
| `genetic[0].evidence` (ClinGen row) | dispute is against PCD, not against HTX9 |
| `pathophysiology` male-infertility node | source says "likely male infertility" |
| `phenotypes` absent-ODA item | HP:0012256 names complete absence, the defect is partial |
| ICSI `target_mechanisms` | cycle reached cleavage, not pregnancy |

Each was expressing a directness qualification rather than a support direction, which is what `PARTIAL` used to conflate. All four now carry `directness: INDIRECT` and say why. The file no longer mentions `PARTIAL`.

**Edge polarity is now stated.** The hypothesis `notes` say explicitly that the group is declared on the ODA-docking to `Respiratory Ciliary Dysfunction` edge to explain why that edge is *attenuated*, not why it holds, and that the opt-in is not the edge's causal warrant.

**One suggestion I did not act on**, since the review judged both gradings defensible and raised it only against future normalisation: the two PMID:24876927 items keep their different `evidence_source` values. Detergent fractionation of isolated sperm is a biochemical assay, so `IN_VITRO`; immunofluorescence localisation in ex vivo sperm is `MODEL_ORGANISM`. The split is deliberate and each explanation carries its own reasoning, so normalising either way would lose information.

Validation on the new head: `just validate-disorders` passes, 98/98 snippets verified, and all ten offline gates OK with nothing added to any baseline.

---

PR #10532, opened 2026-09-02
Curate Desmoid Tumor (MONDO:0007608)

justaddcoffee:
New entry for **desmoid tumor** / aggressive fibromatosis (`MONDO:0007608`) — a genuine gap: no existing file, no MONDO id collision, no open or closed curation issue.

## The mechanism is convergent, not branching

Two genetically distinct lesions produce one biochemical failure:

| Arm | Lesion | Effect on the destruction complex |
|---|---|---|
| Sporadic (~85%) | somatic `CTNNB1` exon 3 (T41A, S45F, S45P) | removes the residues GSK3β/CK1 phosphorylate |
| FAP-associated | germline `APC` loss | removes the scaffold that presents β-catenin |

Both converge on **Failure of Cytoplasmic β-catenin Degradation** → nuclear β-catenin → **Clonal Myofibroblast Proliferation** → **Infiltrative Growth without Metastatic Capacity** → **Local Tissue and Organ Compromise**. That last step is the whole clinical disease: because the tumour never spreads, all morbidity is a function of what it sits next to.

## Two nodes deliberately outside the chain

**Notch Pathway Dependence** is marked `mechanism_confidence: HYPOTHETICAL`. The evidence for it is the response to γ-secretase inhibition in the DeFi trial — the node is inferred *backwards from therapy*, not from pathway measurement in this tumour. It's included because it's the only target with an approved drug, and graded honestly rather than promoted to the main chain.

**Unpredictable Natural History with Spontaneous Regression** is a property, not a causal step — but it's the single most consequential fact about this disease. It's why surgery lost its first-line role, and why uncontrolled treatment series here systematically overstate benefit. The sorafenib trial makes it quantitative: **36% of placebo patients were progression-free at two years.** Consequently both chemotherapy entries carry their own evidence of having *no randomised control* ("There were no randomized controlled trials", "There was no case-control report (vs. no treatment)") rather than response rates alone.

## Content

- **7 pathophysiology nodes**, 6 causal edges, graded `DIRECT` / `INDIRECT_KNOWN_INTERMEDIATES` / `INDIRECT_UNKNOWN_INTERMEDIATES` by what the source actually establishes
- **2 subtypes** (sporadic, FAP-associated), with desmoid disease recorded as the **second cause of death in FAP** after colorectal cancer
- **5 phenotypes** using HPO's exact terms — `HP:6001034` Desmoid tumor and `HP:0100245` Gastrointestinal desmoid tumor — so no generic soft-tissue parent was needed
- **`CTNNB1` with prognostic granularity**: S45F predicts 5-year RFS 0.45 vs 0.91 for wild type, confirmed by an 8-study meta-analysis
- **6 treatments** with `target_mechanisms` links, including both randomised agents (nirogacestat, sorafenib)
- **Environmental**: antecedent surgical trauma and pregnancy, with `influences_mechanisms` explicitly flagged as association not mechanism
- **Diagnosis**: nuclear β-catenin IHC, recorded *with* its antibody-dependence trap (sensitivity 54–96%, specificity 98–62% across three clones)

## Ontology corrections caught in curation

Three NCIT terms were initially bound wrongly and caught by the term validator: `Gamma-Secretase Inhibitor` and `Tyrosine Kinase Inhibitor` are drug *classes*, not medical actions — they moved to `therapeutic_agent` alongside the specific CHEBI drug. `Surveillance` isn't under `NCIT:C25218` at all, so **Active Surveillance is left with no `treatment_term`** rather than bound to a generic `Observation Activity` that would misrepresent it. Documented in `notes`.

## Provenance

Curated **directly from PubMed**, not from a deep-research report — both configured providers were down at curation time (openscientist.io: CloudFront 403 on POST, 401 on authenticated GET; cyberian: 500 from its local backend). 16 candidate PMIDs found via NCBI esearch, fetched with `just fetch-reference`, and read in full before any YAML was written. Top-level `references` generated programmatically from cache frontmatter rather than typed.

## Validation

```
All validations passed!
Snippets checked: 49/49 verified against cached references
Term validation: ✅ passed
Weighted Compliance: 100.0%
```

Gates clean: `check-reference-titles`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-enum-values`, `check-duplicate-keys`, `check-entity-refs`, `validate-graphs` (no findings for this file). History record added and validates.

Not included, listed as extension points in `notes`: structured `histopathology`, `clinical_trials` (DeFi, Alliance A091105), cryoablation/radiotherapy, and the female predominance — no cleanly quotable ratio was found in the cached abstracts.

### DIFF
--- a/history/disorders/Desmoid_Tumor/2026-09-03T142516Z-claude-code-081a4d.yaml
+++ b/history/disorders/Desmoid_Tumor/2026-09-03T142516Z-claude-code-081a4d.yaml
@@ -1,0 +1,47 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Desmoid_Tumor
+  path: kb/disorders/Desmoid_Tumor.yaml
+session:
+  id: 2026-09-03T142516Z-claude-code-081a4d
+  timestamp: '2026-09-03T14:25:16Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10532
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  - phenotypes
+  - environmental
+  - prevalence
+  - progression
+  - clinical_trials
+  - notes
+  summary: 'Review round 2: restore truncated snippets, reorder the convergence chain,
+    regrade the antiestrogen evidence'
+  details: |
+    Response to review round 2 on PR #10532. Both blocking items, plus all six non-blocking suggestions.
+
+    Both blocking items were regressions introduced by the round-1 push, which is the thing worth recording.
+
+    The first is the sharper one: the fix for round-1 finding 7 reintroduced round-1 finding 2. Asked to replace a background-prose snippet with a result sentence, I quoted the result sentence but cut it mid-clause at "in a very limited subset", which dropped "of tumor types, including desmoid-type fibromatosis (71% of cases)". As committed, the snippet never mentioned desmoid tumour at all and the explanation was carrying the disease and the figure, which is the snippet's job. Fixing a truncation complaint by introducing a fresh truncation is a specific failure mode: the edit was made against the reviewer's stated objection rather than against the source text. The full sentence is now quoted, and a second milder truncation in the misdiagnosis snippet was extended to include "histologic analysis".
+
+    The second is a topology defect created by taking a good suggestion without following it through. Splitting node one into CTNNB1 Exon 3 Activating Mutation and Germline APC Loss of Function was right, but the original Constitutive Activation of the Wnt-beta-catenin Axis node was left in place still pointing at Failure of Cytoplasmic beta-catenin Degradation. That made it a third root running parallel to the two new lesion nodes, and the direction was inverted, since constitutive Wnt output is what failed degradation produces rather than its cause. Its own description said "the shared consequence of the two lesions above" while the graph said otherwise. The chain now reads lesions to degradation failure to constitutive Wnt activation to proliferation, the node was moved in the file to match the causal order, and its description was rewritten. The Drosophila destruction-complex citation was re-homed onto the degradation-failure node itself, where the complex biology belongs, rather than left on an edge that no longer exists.
+
+    Suggestion 3 was accepted on the reviewer's reasoning. The antiestrogen evidence on the pregnancy link was graded REFUTE; the link claims that estrogen exposure exacerbates myofibroblast proliferation, and an uninterpretable therapeutic trial fails to confirm that rather than contradicting it. Regraded NO_EVIDENCE, kept in place because a reader weighing the link should know the therapeutic test was uncontrolled and abandoned by guidelines.
+
+    Also taken: Infiltrative Growth without Metastatic Capacity now parents Gastrointestinal Desmoid Tumor, tying the FAP intra-abdominal form into the chain that produces the obstruction phenotype it shares evidence with (5); an explanation that claimed its snippet gave the soft-tissue-neoplasm share when the quote gave only the incidence was corrected (6); the enum-choice rationale moved out of clinical_trials[1].description into notes, and the unattested ACTIVE_NOT_RECRUITING status was removed since the cached record carries no status field (7); and rate_per_100000 0.4 was added as the midpoint (8).
+
+    An incidental defect was caught while editing and fixed: the RINGSIDE description had a duplicated sentence from an earlier partial replacement.
+
+    Validation: 77/77 snippets verified, term validation passes, weighted compliance 100.0 percent, all content gates clean including check-causal-targets and validate-graphs. Pathograph topology verified by parsing the file and printing every node with its outgoing edges, rather than by inspection.
--- a/kb/disorders/Desmoid_Tumor.yaml
+++ b/kb/disorders/Desmoid_Tumor.yaml
@@ -165,11 +165,81 @@ pathophysiology:
         encodes for a protein regulating β‐catenin levels.'
       explanation: Identifies APC as a regulator of beta-catenin levels, which is how this arm
         reaches the shared node.
+
+- name: Failure of Cytoplasmic beta-catenin Degradation
+  biological_scale: MOLECULAR
+  description: >-
+    Unphosphorylated beta-catenin escapes ubiquitin-mediated destruction, accumulates
+    in the cytoplasm and translocates to the nucleus, where it partners with TCF/LEF
+    transcription factors. Nuclear beta-catenin is the diagnostic hallmark of the
+    tumour on immunohistochemistry, which makes this node directly observable in
+    routine practice rather than inferred.
+  cellular_components:
+  - preferred_term: beta-catenin destruction complex
+    modifier: DECREASED
+    term:
+      id: GO:0030877
+      label: beta-catenin destruction complex
+  evidence:
+  - reference: PMID:31189665
+    reference_title: 'Destruction complex dynamics: Wnt/β-catenin signaling alters Axin-GSK3β interactions
+      in vivo.'
+    supports: SUPPORT
+    evidence_source: MODEL_ORGANISM
+    snippet: The central regulator of the Wnt/β-catenin pathway is the Axin/APC/GSK3β destruction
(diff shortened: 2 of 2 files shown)


justaddcoffee:
All seven blocking items fixed, plus most of the suggestions. Rebased on current `origin/main`.

**Your central diagnosis was right, and it is the part worth naming.** Two claims in `notes` said facts were unavailable when they were sitting in full-text files this PR itself committed:

| `notes` claimed | Reality |
|---|---|
| "Vinblastine has no CHEBI term" | `CHEBI:27375` vincaleukoblastine, already in the committed cache, already bound to this agent in `Classic_Hodgkin_Lymphoma.yaml` |
| "No quotable female-predominance ratio was found" | `PMID:35670122`, cited 5x in this entry: *"DTs occur predominantly in women (approximately 70% of cases)"* |

Same cause for both: `PMID_37436594.md` and `PMID_35670122.md` are **69k and 35k characters of full text**, not abstracts, and I read them selectively for the claims I was writing rather than end to end. I have since read both in full, which is where the codon frequencies, the prospective surveillance figures, the misdiagnosis rate, the postsurgical recurrence ranges and both NCT ids came from. Both claims are **retracted in `notes`** rather than silently deleted.

## Blocking items

1. **Phenotype connectivity** - the big one. `Local Tissue and Organ Compromise` now has `downstream` edges to `Chronic Pain`, `Limitation of Joint Mobility` and `Intestinal Obstruction`, bare names per the convention. The graph is one piece now.
2. **Truncated snippet** replaced with the full sentence carrying the 57-86% figure.
3. **Vinblastine** bound to `CHEBI:27375`.
4. **Female predominance and the estrogen axis** modelled - and the interesting part is that it comes with a `supports: REFUTE` item: antiestrogen ORR of 48-51%, *"the lack of an active surveillance comparator makes this finding difficult to interpret. Therefore, treatment guidelines no longer routinely recommend hormone therapies."* The association is real; the therapeutic inference from it failed.
5. **`Pregnancy`** now declares `influences_mechanisms` with `EXACERBATES` toward `Clonal Myofibroblast Proliferation`.
6. **Structured `prevalence:`** - `ANNUAL_INCIDENCE`, `BAND_1_9_PER_1000000`, 0.3-0.5 per 100k.
7. **`evidence_source`** - `PMID:15375433` `IN_VITRO` to `HUMAN_CLINICAL`, and its background-prose snippet swapped for the tissue-microarray result.

## Suggestions taken

- **8** - regression upgraded to 10-28% plus the prospective **55% under active surveillance** (NCT02547831, treatment-free survival 65.9% at 3 years). The 5-10% figure is retained alongside, so the range across sources is visible rather than only the best number.
- **9** - natural-history node no longer isolated; reached by an edge from `Clonal Myofibroblast Proliferation`. Also added a four-phase `progression:` section.
- **10** - **took this one.** Node 1 split into `CTNNB1 Exon 3 Activating Mutation` and `Germline APC Loss of Function`, both feeding the shared degradation-failure node. The convergence is now in the graph instead of only in prose, and the two subtypes have distinct mechanistic anchors.
- **12, 13, 14, 15, 16** - `clinical_trials` (DeFi and RINGSIDE), codon frequencies evidenced, `biological_scale`, `therapeutic_modality`, human co-citation on the Drosophila edge.

## One declined, with reasons

**11, `conforms_to sustaining_proliferative_signaling`.** I do not think it fits. That module's central node is defined as flux through *"the RAS-RAF-MEK-ERK (MAPK) pathway"* and *"the PI3K-AKT-mTOR pathway"*, bound to `GO:0007265` Ras protein signal transduction and the ERK cascade. Desmoid is canonical Wnt/beta-catenin, a transcriptional pathway sharing none of that machinery. A conforming node here would carry **none** of the module's expected biological processes, and declaring conformance on the strength of "both are proliferative" would make the conformance check mean less.

You were right that `invasion_and_metastasis` does not apply, and for a nice reason: this tumour invades but provably never metastasises, so it satisfies half the module and refutes the other half. Both points are now in `notes`.

## A process note

Three gates initially failed after my edits, in `Epidermolysis_Bullosa_Simplex`, `Appendiceal_Neoplasm`, `Bullous_Pemphigoid`, `Follicular_Lymphoma` and `Meckel_Diverticulum`. **None were in my diff.** The branch had gone stale against a moved `origin/main`, so the gates were evaluating other contributors' files. I rebased on current `origin/main` and reviewed the post-rebase diff for unrelated deletions or protected-path churn before continuing - 30 files, all in scope. All gates then pass.

## Validation

```
All validations passed!
Snippets checked: 76/76 verified against cached references   (was 49/49)
Term validation: passed
Weighted Compliance: 100.0%
```

`check-reference-titles`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-enum-values`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets` all clean.

On GeneReviews: agreed, there is no chapter for desmoid tumour itself; the FAP arm is covered by *APC-Associated Polyposis Conditions*, a predisposition syndrome already curated separately here.

justaddcoffee:
Both blocking items fixed, plus all six suggestions. Rebased on current `origin/main`.

**Both blocking items were regressions from my round-1 push**, and the first one is worth naming precisely.

## 1. The fix for finding 7 reintroduced finding 2

You asked me to swap a background-prose snippet for a result sentence. I quoted the result sentence — and cut it mid-clause at *"in a very limited subset"*, dropping *"of tumor types, including desmoid-type fibromatosis (71% of cases)"*. As committed the snippet **never mentioned desmoid tumour at all**, and the `explanation` was supplying both the disease and the figure. That's exactly the defect you'd flagged one round earlier in a different place.

The mechanism of the error: I edited against your stated objection ("this is background prose, use the result") rather than against the source text. Satisfying the objection and quoting the sentence are different operations, and I did the first. Full sentence now quoted, and the second milder truncation extended to include `histologic analysis`.

## 2. The split left the old node stranded, pointing the wrong way

You were right on both counts — it was a third root, and the direction was inverted. Its own description said *"the shared consequence of the two lesions above"* while the graph said it was their sibling. Before and after:

```
before:  CTNNB1 mutation ─┐
         APC loss ────────┼─→ Failure of Degradation ─→ Proliferation
         Constitutive Wnt ┘        (three roots, Wnt pointing backwards)

after:   CTNNB1 mutation ─┐
                          ├─→ Failure of Degradation ─→ Constitutive Wnt ─→ Proliferation
         APC loss ────────┘
```

Every annotation kept. The node also moved in the file to match the causal order, its description was rewritten, and the *Drosophila* destruction-complex citation was re-homed onto the degradation-failure node itself — where that biology belongs — rather than left on an edge that no longer exists.

I verified the result by **parsing the file and printing every node with its outgoing edges**, rather than by reading it. That's the check I should have run after the round-1 split; it would have caught this immediately.

## All six suggestions taken

- **3** — regraded `REFUTE` → `NO_EVIDENCE`. Your reasoning is right: the link claims *estrogen exposure exacerbates proliferation*, and an uninterpretable therapeutic trial fails to confirm that rather than contradicting it. Kept in place, because a reader weighing this link should know the therapeutic test was uncontrolled and abandoned by guidelines.
- **4** — misdiagnosis snippet extended to `histologic analysis`.
- **5** — `Infiltrative Growth without Metastatic Capacity` → `Gastrointestinal Desmoid Tumor`, using PMID:35670122 on the intra-abdominal predominance in FAP. Good call — it ties that form into the chain producing the obstruction phenotype it already shared evidence with.
- **6** — corrected; the quote gives only the incidence, so the explanation now claims only that.
- **7** — enum rationale moved out of `clinical_trials[1].description` into `notes`. And you're right that `ACTIVE_NOT_RECRUITING` was unattested: the cached record carries no status field, so I've **dropped the slot** rather than assert one.
- **8** — `rate_per_100000: 0.4` added.

Also fixed while in there: the RINGSIDE `description` had a duplicated sentence left by an earlier partial replacement of mine.

## Validation

```
All validations passed!
Snippets checked: 77/77 verified against cached references
Term validation: passed
Weighted Compliance: 100.0%
```

`check-reference-titles`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-enum-values`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `validate-graphs` — all clean. History record added and validates.

Deep research still unavailable; same failure signature as both prior rounds. I'll re-attempt when a provider returns.

PR #10533, opened 2026-09-02
Review 20 skin disease entries against their cited sources

cmungall:
## Summary

I re-read every cited publication for 20 skin disease entries that had no REVIEW event in `history/`, and checked each evidence item three ways: is the quoted snippet actually in the paper, does it support the claim it is attached to, and does `evidence_source` describe the right kind of study. That covered roughly 880 evidence items.

The headline finding is a gap in the tooling rather than in any one entry. `conf/reference_validator_config.yaml` lists `DOI` in `skip_prefixes`, so DOI-referenced snippets are never machine-checked — about a third of what I looked at. Scabies (28 of its 34 references) and Cutaneous SCC (17 of 22) were almost entirely unverified before this pass. Those all turned out to be accurate. What did fail were four *PMID* snippets that the repo's own snippet audit passes, because its relaxed matching ignores differences the strict validator rejects.

## Corrections made (10 entries)

Two are verbatim failures that would fail `linkml-reference-validator`:

- **Staphylococcal Scalded Skin Syndrome** — snippet read `Delta381-Dsg1`; the cache has `Δ381-Dsg1`.
- **Pyoderma Gangrenosum** — two snippets dropped the trailing `(Ortega-Loayza et al. 2022).` the cache carries, and two stopped one word before the interleukins they were cited for (the quote ended at "inhibitors of tumour necrosis factor" while the explanations cited it for IL-1 and IL-17/IL-23 blockade).

Four are claims a cited source does not make, or contradicts:

- **Staphylococcal Scalded Skin Syndrome** — ETD was listed among the SSSS exfoliative toxins and had it converging on the desmoglein-1 cleavage site. That is uncited, and PMID:39411997, which the entry itself cites, says ETD "is not strongly associated with SSSS". Removed from both claims, with the reasoning recorded in `notes`.
- **Oculocutaneous Albinism** — the OCA6, OCA7 and OCA8 gene-causation claims were each graded as supported by one identical sentence: the cited review's scope statement ("In this review, we will focus only in the genetics of skin pigmentation in OCA..."). It names none of the three genes, and the cached record is abstract-only and contains neither the gene symbols nor the subtype labels. All three are now `NO_EVIDENCE`, with explanations saying what the quote does and does not establish, rather than deleting the claims or inventing a substitute quote. A primary gene-delineation citation is still needed.
- **Peeling Skin Syndrome** — removed "A recurrent European founder variant, c.763T>C p.(Trp255Arg), is described." No cached reference anywhere in the repository mentions that allele, and it is not the one the literature reports as the recurrent European founder in acral PSS.
- **Epidermolysis Bullosa Simplex** — a two-word paper title used as a finding, its supporting text and its snippet, against a cache whose body is empty (`content_type: unavailable`). The paper is already cited by PMID with a real verified sentence.

The rest are classification and consistency errors:

- **Kindler Epidermolysis Bullosa** — an author byline (`Youssefian L(1), Vahidnezhad H(1), Uitto J(1).`) was curated as a finding, its supporting text and its evidence snippet. It is an exact cache substring, which is why every mechanical check passed it; it is a list of authors supporting no claim, and at 44 characters it clears the snippet-length floor while not being a title. Also regraded a clinical case report from COMPUTATIONAL.
- **Hailey-Hailey Disease** — `prevalence_class` said `BAND_1_9_PER_1000000` (0.1–0.9 per 100,000) while its own `rate_per_100000: 2.0` and its cited "1/50,000" quote both say 2 per 100,000. Botulinum toxin A, a ~150 kDa bacterial protein, was tagged `SMALL_MOLECULE`. A snippet was cut mid-word and its explanation blamed the cache for a truncation introduced during curation — the cache holds the complete sentence.
- **Sweet Syndrome** — the diagnostic-criteria record was evidenced by a table-of-contents sentence listing what the review covers. The actual von den Driesch criteria rule is reproduced verbatim in the other cached reference, so it now cites that, keeping the original paper as an `INDIRECT` item for the 38-patient series behind the criteria.
- **Netherton Syndrome** — two items graded MODEL_ORGANISM rest on a cell-free recombinant digestion assay; one graded MODEL_ORGANISM is a human-genetics background sentence, whose snippet also carried an "Abstract" heading glued to the first word by the PDF extractor. Also dropped the deprecated `percentage` field and corrected two places calling a case report a review.
- **Bullous Pemphigoid** — a human lesional-skin finding was graded MODEL_ORGANISM while the same quote was HUMAN_CLINICAL elsewhere in the file. A frequency note miscomputed the OCCASIONAL band.

## Two things that happened after the first push

**The `supports` vocabulary changed under me.** Main narrowed `EvidenceItemSupportEnum` in #7439 while this branch was in flight, removing `PARTIAL` in favour of `SUPPORT` plus the new `directness` slot, or `NO_EVIDENCE`. Four items I had added still used `PARTIAL`. I rebased onto main and mapped each per the migration the schema comment describes: the three Oculocutaneous Albinism scope-sentence items to `NO_EVIDENCE` (the "simple irrelevance" case), and the Sweet Syndrome cohort quote to `SUPPORT` with `directness: INDIRECT`, since the criteria follow from the 38-patient series by an inference step.

**One deliberate loose end.** Fixing the Bullous Pemphigoid grading conflict makes its row in `tests/snippet_grading_baseline.txt` stale. I removed the row, which failed the curation-scope check, so I restored it: this PR touches `kb/`, and CI blocks such a PR from also modifying `tests/`. Taking a `scope-override` label to get my own PR past that check seemed like the wrong trade for one line of dead config. The row is inert — `check_snippet_grading` reports only new divergences and does not verify that each baseline row still matches a live one, which I confirmed by running it with the row restored. It should be dropped in a separate infra change. The Bullous Pemphigoid history record says the same.

## Recorded, not fixed

Each of the 20 entries gets a history record carrying its full findings. Two are worth a look beyond the file they sit in:

- **Peeling Skin Syndrome** has a causal edge running opposite to the paper it cites. The edge says defective corneodesmosome cohesion "progresses to complete loss of corneodesmosin"; PMID:20691404 says a homozygous CDSN nonsense mutation *causes* the loss, which then causes the barrier defect. Fixing it means either reversing the edge or renaming the downstream node, so it needs a curator decision rather than a mechanical edit.
- **Epidermolysis Bullosa Simplex** has a prevalence block whose two headline numbers ("approximately 70% of all EB cases", "1 in 30,000 to 50,000") appear in neither cited abstract, and the second conflicts with the registry the entry cites (11.07 per million for all inherited EB is about 1 in 90,000). Both replacements are already sitting in the entry's own bibliography.

Ten entries needed no correction. Merkel Cell Carcinoma came through clean on every axis, including the MCPyV-positive versus UV-driven aetiology split that is the obvious confusion risk there.

## Validation

Schema, term validation and strict snippet re-verification pass on all 10 changed entries. Whole-KB gates are clean: enum values, duplicate keys, snippet grading, title snippets, snippet length, folded hyphens, environmental evidence. All 20 history records validate and the history tests pass.

One caveat: `linkml-reference-validator` returns `Total checks: 0` in my sandbox, which I believe is because it needs fetch access. I relied on `dismech.reference_snippet_audit --strict` against the committed cache instead, which is the deterministic offline equivalent and is what caught the four verbatim failures above.

https://claude.ai/code/session_01M3qRiRWmrH49ZgTdv5nqou

### DIFF
--- a/history/disorders/Netherton_Syndrome/2026-08-29T213448Z-claude-code-9cf215.yaml
+++ b/history/disorders/Netherton_Syndrome/2026-08-29T213448Z-claude-code-9cf215.yaml
@@ -45,6 +45,19 @@ events:
       it cites. Also recorded that the source says "incidence" while the record asserts
       BIRTH_PREVALENCE.
 
+    Grading rule used for background sentences. Several items in this batch quote a
+    paper's introductory background statement rather than its own result, and the
+    question is what evidence_source such a sentence carries. The rule applied: grade
+    by whether the citing publication presents human clinical material at all, since
+    that is what the slot exists to keep honest. The SPINK5 background sentence here
+    stays HUMAN_CLINICAL because that paper studies Netherton syndrome patients
+    directly -- a consumer filtering for human evidence is right to surface it. The two
+    background sentences in Oculocutaneous_Albinism went to OTHER because their paper
+    is a mouse-model study whose only human component is melanocyte culture, graded
+    IN_VITRO elsewhere in that entry; no patients appear in it, so HUMAN_CLINICAL
+    would have been a false signal. The two cases look alike but differ on the point
+    the slot is for.
+
     Recorded, not fixed:
     - pathophysiology[Hair shaft structural fragility from follicular protease imbalance]
       asserts a follicular protease mechanism; its quote reports phenotype co-occurrence only.
--- a/kb/disorders/Pyoderma_Gangrenosum.yaml
+++ b/kb/disorders/Pyoderma_Gangrenosum.yaml
@@ -431,6 +431,7 @@ pathophysiology:
   - reference: PMID:37610614
     reference_title: "Pyoderma Gangrenosum: Treatment Options."
     supports: SUPPORT
+    directness: INDIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       an increasing number of studies on the positive effects of biologic
@@ -438,7 +439,9 @@ pathophysiology:
       interleukin-17, interleukin-23 or complement factor C5a.
     explanation: >-
       The therapeutic responsiveness to cytokine-directed biologics supports
-      these cytokines being mechanistically operative in the disease.
+      these cytokines being mechanistically operative in the disease. This is
+      the schema's worked case for INDIRECT: a therapeutic response cited as
+      validation of the mechanism it targets.
 - name: Neutrophil Recruitment to the Dermis
   biological_scale: CELLULAR
   description: >-
@@ -2284,6 +2287,7 @@ treatments:
   - reference: PMID:37610614
     reference_title: "Pyoderma Gangrenosum: Treatment Options."
     supports: SUPPORT
+    directness: INDIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: >-
       an increasing number of studies on the positive effects of biologic
@@ -2291,7 +2295,7 @@ treatments:
       interleukin-17, interleukin-23 or complement factor C5a.
     explanation: >-
       Supports the biologic class including IL-1 inhibitors as having reported
-      positive effects; graded PARTIAL because this sentence groups IL-1
+      positive effects. Graded INDIRECT because the sentence groups IL-1
       blockade with other biologics rather than reporting a trial of it.
 - name: Adalimumab (anti-TNF biologic therapy)
   description: >-
--- a/kb/disorders/Sweet_Syndrome.yaml
+++ b/kb/disorders/Sweet_Syndrome.yaml
@@ -585,8 +585,9 @@ diagnosis:
     snippet: >-
       This article presents data from 38 patients with Sweet's syndrome
     explanation: >-
-      The foundational 38-patient series behind the modified criteria; it
-      establishes the cohort, not the criteria wording itself.
+      The modified criteria are derived from this 38-patient series, so the
+      criteria follow from the quoted sentence by an inference step rather than
+      being asserted by it.
 - name: Work-up for underlying trigger (malignancy, drug, infection)
   description: >-
     Because Sweet syndrome can be paraneoplastic, evaluation should include a


cmungall:
Thanks — the independent re-verification against the cache is genuinely useful, and the blocking call was right.

**The four `PARTIAL` values were already fixed before this review posted.** CI caught them at 13:15 and I pushed the rebase plus the migration at ~13:25; the review ran against `f59fc94b` and landed at 13:28. The mapping I chose matches yours exactly: the three OCA items to `NO_EVIDENCE`, the Sweet Syndrome item to `SUPPORT` + `directness: INDIRECT`. Current head is `33a5223d`.

**The predicted Pyoderma Gangrenosum conflict did not materialise.** The rebase was clean and that file now carries zero `supports: PARTIAL` — main's migrated values came through intact, and nothing reintroduced the retired value. I verified across every file this PR touches.

**Taken:** the `PMID:37610614` explanation that still read "graded PARTIAL because this sentence groups IL-1 blockade with other biologics". The reasoning was right and is precisely what `directness` records, so that item now carries `directness: INDIRECT` and the prose no longer names a retired value. I did the same for the pathophysiology item quoting the same sentence — a therapeutic response cited as validation of the mechanism it targets is the schema's own worked example of `INDIRECT`. I also reworded the Sweet Syndrome explanation, which still read as a `PARTIAL` justification.

Two further "graded PARTIAL" explanations remain in that file, on evidence items this PR does not otherwise touch. Those are backlog from main's migration and I left them rather than widening the PR.

**On the background-sentence suggestion, I think the two cases genuinely differ, so I stated the rule instead of aligning them.**

The distinction is what `evidence_source` exists to protect — whether a consumer filtering for human evidence would be misled:

- The Netherton paper studies Netherton syndrome **patients** (that is what its comparative design is). Someone querying for human evidence that SPINK5 causes NS is right to surface it, so `HUMAN_CLINICAL` is not a false signal.
- The OCA paper is a mouse-model study. Its MeSH does list `Humans`, but that component is melanocyte **culture**, which this entry already grades `IN_VITRO` elsewhere. No patients appear in it, so `HUMAN_CLINICAL` there was a false signal, and `OTHER` removes it without claiming mouse support for a human genetics fact.

So the rule is: grade a background sentence by whether the citing publication presents human clinical material at all. Aligning Netherton to `OTHER` would have stripped a correct human signal from a claim whose paper really does study patients — and applied consistently it would also have caught the sibling case-report item on that same node, leaving a well-established human genetic claim with no human-graded evidence. That seemed worse than the inconsistency. The rule is now written into the Netherton history record; if you read the boundary differently I am happy to move it.

**On the stale-base gate artefacts** — `Hyperprolinemia_Type_2`, `Follicular_Lymphoma`, `Metastatic_Ovarian_Cancer` and the two chondrodysplasia entries — those have all disappeared post-rebase, which confirms your read that they were base artefacts rather than regressions.

Re-running the full gate set on the current head: schema and strict snippet validation pass on both edited entries (147/147 and 35/35), and enum values, snippet grading, duplicate keys, title snippets and snippet length are clean whole-KB.

---

Issue #8597, opened 2026-08-16
Reference validator strips bracketed abbreviations, breaking otherwise-exact snippets


caufieldjh:
## Summary

`linkml-reference-validator`'s text normalization removes square-bracketed spans before matching. A snippet copied **verbatim** from a cached abstract therefore fails with `Text part not found as substring` whenever the quoted span contains a bracketed abbreviation — which is extremely common in structured biomedical abstracts.

This has now bitten four separate curation PRs in this series.

## Reproduction

Any snippet quoting a span that contains `[...]`. Worked example from `references_cache/PMID_37959386.md`:

```
aHIT patients are at risk for treatment failure with (activated partial
thromboplastin time [APTT]-adjusted) direct thrombin inhibitor (DTI) therapy
```

The validator reports the missing text as:

```
Text part not found as substring: 'aHIT patients are at risk for treatment
failure with (activated partial thromboplastin time -adjusted) direct thrombin
inhibitor (DTI) therapy ...'
```

Note `[APTT]` has become nothing, leaving `time -adjusted`. Round parentheses `(DTI)` survive; only square brackets are stripped.

## Occurrences so far

| PR | Reference | Bracketed span |
|---|---|---|
| #8564 (3q29) | `PMID:33564151` | `[GERD]` |
| #8582 (bone giant cell tumor) | `PMID:22773395` | `[RR]` |
| #8582 (bone giant cell tumor) | `PMID:39236154` | `[CI]` |
| #8595 (HIT) | `PMID:37959386` | `[APTT]` |

The pattern is consistent: abstracts define an abbreviation in square brackets on first use, and that is exactly where the substantive clause tends to sit.

## Why it matters

The failure is silent about its own cause. The error says the text is not a substring, which reads as *"the curator paraphrased"* — the one thing the validator exists to catch. A curator who trusts the message will go looking for a fabrication that isn't there.

Worse, the natural workaround is harmful. Truncating the snippet before the bracket is quick, and it drops the clause the `explanation` depends on. That happened in #8564 and was caught in review as a claim–evidence mismatch. The safe workaround — quote a bracket-free span and state the omitted words verbatim in the `explanation` — is what PRs #8582 and #8595 use, but it produces snippets that begin mid-sentence for no reason a reader can see.

## Suggested fixes, roughly in order of preference

[...]

Issue #8085, opened 2026-08-06
Backfill influences_mechanisms on existing environmental entries


caufieldjh:
Follow-up to #8033 / #8084.

PR #8084 added `Environmental.influences_mechanisms`, the linking slot that puts an environmental exposure into the pathograph, and backfilled it on `Arsenic_Poisoning` as the worked example. Everything else is still disconnected.

**Scope:** ~828 environmental entries across 348 disorder files currently declare no mechanism link, so they stay off the pathograph.

**Why this is a curation pass, not a script:** each link asserts that a specific exposure acts on a specific mechanism node, in a specific direction, and takes its own evidence. There is no mechanical rule that picks the right `target` or the right `environmental_effect` — the useful bit is exactly the judgement a sweep cannot make.

**Suggested approach:** batch by disorder, prioritizing entries where the exposure is the disease's actual initiating step (toxic exposures, infectious/occupational exposures, allergen and dietary triggers). Set `environmental_effect` explicitly; omitting it falls back to the neutral `influences` predicate on purpose. Protective exposures should use `PROTECTS_AGAINST` — that also now drives the KGX predicate, superseding the older free-text `effect:` regex from #2098.

See the "Linking Environmental Factors into the Pathograph" section in CLAUDE.md and `docs/pathographs.md` for the pattern.

caufieldjh:
First tranche is up: #8090 — **33 links across 11 disorders**, in four reviewable commits by exposure class.

**Survey finding worth recording here.** I checked whether any of this could be automated: is there an environmental entry whose `exposure_term` matches a pathophysiology node's `triggers:` ECTO term, or a disorder with exactly one pathophysiology node? Across all 828 entries in 348 disorders:

- ECTO trigger-term matches: **0** (Arsenic, the #8084 worked example, was the only one and is already linked)
- Single-pathophysiology-node disorders: 11
- No pathophysiology nodes at all: 1
- Needs judgement: **814**

So there is no deterministic shortcut. Every remaining link is a judgement about which node the exposure acts on and in which direction, and needs its own evidence. Planning should assume per-entry curation throughout.

**Suggested ordering for the rest**, roughly by decreasing certainty:
1. Drug-toxicity entries — Acetaminophen, Allopurinol-induced SJS/TEN, chemotherapy-induced diarrhea, drug/toxin-induced PAH — plus Cadmium. Same shape as this tranche: the exposure is the initiating step and the entries already have a first node representing it.
[...]

PR #10537, opened 2026-09-02
Evidence audit of 20 eye-disease entries, and fixes for what it found

cmungall:
## Summary

Twenty eye-disease entries had never been reviewed — no `REVIEW` event anywhere in `history/`. This PR reviews all twenty against their cited papers and fixes what the review found. One entry, Central Serous Chorioretinopathy, was citing quotes that do not appear in the papers at all, including a GWAS claim of "seven loci including CDH5" where the paper reports four loci and never mentions CDH5. Those passed CI because the snippet checker skips `DOI:`-prefixed references.

Every evidence item in all twenty files (~900) was re-read against its cached publication, checking four things: that the quote is really in the paper, that it supports the specific claim it is attached to, that `evidence_source` describes the right kind of study, and that the surrounding description does not say more than the paper does.

## What the review found

Eleven entries came back with only minor issues. Nine needed real fixes:

| Entry | Problem |
|---|---|
| Central Serous Chorioretinopathy | Four fabricated or spliced-together quotes; one cited paper whose cache body is empty |
| CRX-Related Retinopathy | ~17 "quotes" were raw PDF line fragments split mid-sentence |
| CAPN5-Related Vitreoretinopathy | Prevalence note contradicted a paper cited in the same file |
| EYS-Related Retinitis Pigmentosa | Zebrafish model and all prevalence numbers unsupported by their citations |
| Cytomegalovirus Retinitis | Antiretroviral therapy claim cited a sentence about an ART complication; five phenotypes had no evidence at all |
| FLVCR1-Related Retinopathy | One mouse measurement cited twice, appearing as independent human support |
| Congenital Glaucoma | A "fails to recapitulate" link attributed another lab's zebrafish result to the cited paper |
| Autoimmune Retinopathy | All 12 `supporting_text` values were curator commentary, not quotes from the source |
| PDE6A-Related Retinopathy | Progression claims rested on content-free sentences like "from 44 families were included" |

## What changed

Fabricated quotes were replaced with real ones from the same papers, and claims corrected to what those papers actually report. Uncited assertions were either sourced (13 new references fetched, including the HAART maintenance-withdrawal literature for CMV retinitis and the Alexandre-Moreno zebrafish paper for congenital glaucoma) or trimmed. Overstated claims were brought back to source strength — FLVCR1's single-patient "determines penetrance" node is now stated as the hypothesis its authors call it, and PDE6A's plasma-cGMP finding no longer "confirms the mechanism in humans" when the paper says "indicating a possible role".

Judgment calls that need a curator rather than a quote repair are recorded per-entry in the history records: a CORDX3 subtype attribution, an incidence-vs-prevalence modelling choice, and CMV retinal-cell tropism bindings.

## Note on the merge

This branch predated #7439, which retired the `PARTIAL` evidence value and added the separate `directness` axis, so every `PARTIAL` the audit wrote had to be migrated. That was done with `scripts/migrate_supports_enum.py`, then each of the 24 migrated items was assessed for directness rather than left unset: 17 indirect, 6 direct, and one corrected to `NO_EVIDENCE` — a review's closing line about future biomarkers, cited in support of rituximab, names no drug and reports no outcome.

Four merge conflicts were resolved on one rule: where main re-graded a sentence this branch left alone, main's value wins; where the audit changed the quoted sentence or replaced the citation, main's grade does not carry over to different text.

## Two tooling gaps worth separate issues

1. `reference_snippet_audit` skips `DOI:`-prefixed references by prefix. All four fabricated quotes in this PR were sitting in that blind spot.
2. `references[].findings[].supporting_text` is never checked against the source. In Autoimmune Retinopathy all twelve had drifted into curator commentary without anything noticing.

## Validation

Schema validation, strict snippet audit (1048/1073 verified; the 25 skipped are DOI-prefixed and were verified by hand), duplicate-key, entity-ref, and snippet-grading checks all pass. Provenance is recorded as 20 `REVIEW` history records for the audit and 20 `EDIT` records for the fixes, each listing findings fixed and findings deliberately skipped with reasons.

https://claude.ai/code/session_01JrKHq2QTD4HWsiUBxdUiNu


cmungall:
All three suggestions are fixed in c6a5828c8. Thanks — the GeneReviews one was a real inconsistency I introduced.

**Suggestion 1, GeneReviews grading.** Correct, and it was my miss. PMID:20301423 is a GeneReviews chapter, and this PR harmonized Congenital_Glaucoma's GeneReviews citations to `OTHER` while leaving CACNA1F's seven items at `HUMAN_CLINICAL`. All seven are now `OTHER`.

**Suggestion 2, the CAPN5 item.** Agreed, and I'd put it more strongly than a labelling nuance: quoting a paper's own limitation as support for the claim it limits is wrong regardless of directness. It is now `NO_EVIDENCE` with `directness` dropped, and the explanation says plainly that the sentence carries no evidence about whether the proteolysis is photoreceptor-specific. It still does useful work bounding the node's `CL:0000210` binding, which is why the item is kept rather than deleted.

For context on how that item got there: the sentence was added by this audit precisely *because* the node was attributing the proteomic signal to photoreceptors when the paper disclaims that. It was written as `PARTIAL`, which the #7439 migration then turned into a blanket `SUPPORT`. Your finding is the third case in this PR where that blanket conversion needed a curator's eye — the other two I caught while assessing directness.

**Suggestion 3, retired-`PARTIAL` prose.** Fixed on the LHON line, keeping the RGC-5 cell-line-authentication caveat intact.

I checked whether to sweep this more widely. There are 21 stale `PARTIAL` prose mentions across six of these files, but only that one sits on a line this PR touched — the other 20 predate the branch and are part of the same follow-up the migration script's worklist describes. I've left them rather than widen the diff; happy to do the full sweep in a separate PR if that's wanted.

**On the PMID-verification caveat:** worth flagging that the four fabricated quotes this audit found were all on `DOI:`-prefixed references, which `reference_snippet_audit` skips by prefix — so a green snippet audit does not mean every quote was checked. I verified those 25 by hand against the cache bodies. Same blind spot would apply to your run. The PR body lists this and the unchecked `references[].findings[].supporting_text` as two gaps worth their own issues.

Validation on the three changed files: schema validation clean, strict snippet audit 218/218, duplicate-key, entity-ref and snippet-grading all pass.

---

cmungall:
Both points addressed in 69eb56240.

**The leftover clause.** You're right, and it wasn't purely cosmetic — "reaches the claim only by inference" is the definition of `INDIRECT`, so the explanation was arguing for the grade the same commit removed. Anyone reading the item would have found the prose and the value disagreeing. Dropped; the rest of the explanation stands.

**History records.** Added for all three entries. Worth doing beyond the convention: the CAPN5 item has now been graded three different ways inside one PR — `PARTIAL` when the audit added it, `SUPPORT` + `directness: INDIRECT` after the #7439 migration and directness pass, and `NO_EVIDENCE` now — and the earlier record described the first of those. The new record traces the full sequence and says why the final value is right, so the reasoning is recoverable without reconstructing it from the diff. The old record stands unedited, since it accurately describes its own session.

On the GeneReviews spread you flagged: `Congenital_Stationary_Night_Blindness.yaml` cites the same PMID:20301423 and still grades it `HUMAN_CLINICAL`. Agreed that it doesn't belong here — it's outside the twenty entries this audit reviewed, and pulling it in would widen the diff past the reviewed set. It's the same shape as the 17 stale-`PARTIAL` prose mentions: a real inconsistency, correctly scoped to a follow-up sweep rather than this PR.

Thanks for verifying the DOI snippets independently rather than trusting the PR body. That check mattering is the point of the finding — a green snippet audit was exactly what made the four fabricated quotes invisible in the first place.

Validation on this commit: schema validation clean, strict snippet audit 48/48, duplicate-key and entity-ref checks pass, and all three history records validate.

---

cmungall:
`claude-review` is red on `69eb562`, but the failure is the review job itself erroring, not a finding against this PR. The job's own comment reads "Claude encountered an error after 8m 2s", and its checklist stops at item 5 of 7 — it never reached "post review and set PR review status". `test (3.13)` passed on the same commit.

Worth noting what that run *had* finished before it died, since both were independent checks of this PR's central claims: 25/25 DOI-prefixed snippets in Central Serous Chorioretinopathy verified exact against their caches, and 20/20 `supporting_text` values in Autoimmune Retinopathy verified exact. The same reviewer approved the two preceding heads.

I can't re-run it. Both `workflow_dispatch` on `claude-code-review.yml` and a failed-job re-run return 403 for my token, so the remedy in CLAUDE.md ("get a new review, not remove the old one") isn't available to me here. Nothing in this diff would change the outcome either — `69eb562` differs from the approved `c6a5828` only by dropping one stale clause from a YAML explanation and adding three history records.

So this needs someone with Actions permissions to re-run the review job, or a maintainer review. The PR is otherwise green and mergeable: `test (3.13)` passing, no conflict, and local gates clean (schema validation, strict snippet audit, duplicate-key, entity-ref, snippet-grading, plus 47,379 pytest tests). I'll keep watching it rather than leaving it here.

---

cmungall:
/review

---

cmungall:
The bare `/review` comment above is mine and it did nothing — worth explaining, because it's a reusable gotcha rather than a one-off.

`claude-code-review.yml` gates the manual escape hatch on `github.event.comment.body == '/review'`, an exact equality check. I posted exactly `/review`, but the Claude Code attribution footer is appended server-side after the tool call, so what GitHub stored was:

```
/review
```

That is not equal to `/review`, so run 33718151548 fired on the comment and skipped the job. **The `/review` escape hatch cannot be used by any Claude Code agent**, since the footer is not suppressible from the tool side. If that path is meant to be agent-usable, `startsWith(...)` or a trimmed-first-line comparison would fix it; if it is meant to be human-only, it already is, just by accident rather than by design.

For this PR specifically, nothing more is needed from me. `pr-shepherd` already covers this exact state — "REVIEW_REQUIRED + cancelled review: ... explicitly re-trigger it" — and it holds the Actions permission I lack. Its agent job runs on `37 */4 * * *`, so the next pass is 08:37 UTC.

State is unchanged and green apart from the review: head `69eb562`, `test (3.13)` passing, no conflict, all local gates clean.

---

cmungall:
@claude please re-review this PR.

The scheduled review is stuck through no fault of the diff: `claude-review` on `69eb562` failed because the reviewer agent itself errored at 8m02s, part-way through its checklist, so it never posted a verdict. The two earlier approvals (on `4f3f408` and `c6a5828`) were dismissed by the pushes that followed them, which is why the PR now shows REVIEW_REQUIRED. `test (3.13)` is green on the current head and there is no merge conflict.

The three commits since the last approval are small and self-contained: seven GeneReviews items regraded `HUMAN_CLINICAL` → `OTHER` in CACNA1F, one CAPN5 evidence item regraded to `NO_EVIDENCE` (its quote is the paper's own limitation statement about cell-type attribution), one stale clause dropped from that item's explanation, and three `EDIT` history records.

Two things worth verifying independently rather than taking from the PR body, since they are what this audit turns on: that the 25 `DOI:`-prefixed snippets in Central Serous Chorioretinopathy are exact substrings of their cached records — `reference_snippet_audit` skips that prefix, which is how four fabricated quotes survived CI in the first place — and that the 12 `references[].findings[].supporting_text` values in Autoimmune Retinopathy are now real excerpts rather than curator commentary.

---

cmungall:
Both points handled in `d95cccdec`. The ellipsis finding was right, and chasing it turned up something more interesting than a typo.

**The ellipses did elide nothing — but deleting them breaks the file.** Removing all seven drops CRX from 32/32 to 25/32. The reason is the validator's bracket stripping: it removes bracketed spans from the snippet before matching, so a curator can gloss a quote (`this cytokine [IL-6] is a key factor`) without transcribing the gloss, and only the query side is stripped, never the cached text. RefSeq variant nomenclature written mid-sentence by the source is not a gloss, so stripping it left `In humans, E168d2 is associated with...`, which the cache does not contain. The ellipsis was splitting the quote so each fragment matched around the bracket.

So the ellipsis was load-bearing, and the honest fix is at the mechanism rather than the symptom. `conf/reference_validator_config.yaml` already exists for exactly this failure (#8597) — its own comment describes the same shape, an abbreviation defined in brackets mid-sentence, and notes it "cost four curation PRs before it was diagnosed." Added one pattern:

```yaml
literal_bracket_patterns:
  - "^[A-Z]{2,}$"
  - "%"
  - "^NM_\d"      # RefSeq transcript nomenclature
```

Anchored and digit-requiring, so a gloss would have to open with a literal RefSeq accession to be caught by it. Validated the way that file documents its existing patterns were chosen — by replaying every snippet in `kb/`: **144044 verified, 5427 skipped by prefix, zero failures**. No other entry's matching changes, and the four editorial glosses in the KB (`[sic, correct designation is R501X]`, `[IL-6]`, `[of dimethyltryptamine]`) stay stripped as their curators intended. Only this one file uses `[NM_` in a snippet, so the blast radius is seven snippets.

With that in place the ellipses are gone and all seven are plain exact substrings: CRX back to 32/32.

**On the staleness note — you were right, and I've refreshed.** The branch was 61 commits behind; merged current `main` (`48bd0eb`), no conflicts this time. `check-snippet-grading --against-ref origin/main` now reports clean, so both Appendiceal_Neoplasm and Bullous_Pemphigoid were exactly the #10533 staleness you diagnosed. `behind_by` is now 0, which also unblocks the auto-merge sweep's ancestry requirement.

Gates on the merged tree: snippet audit 1048/1073 (25 DOI-prefixed, hand-verified), duplicate-key clean across 4273 files, entity-refs resolve across 2829, enum values clean across 2910, causal targets no new breakage, and the new history record validates.

Thanks for the independent re-verification with your own normalizer rather than the repo's matcher — that's what makes the 25/25 and 20/20 numbers worth anything, given a green audit is precisely what hid the original four fabricated quotes.

---

cmungall:
CI was right to reject my last push, and I've reverted it in `d6f6f17b3`. The head is back to the state you reviewed, with the ellipses intact and CRX at 32/32.

The scope guard caught it:

```
conf/reference_validator_config.yaml
```

That guard is correct here and `scope-override` would be the wrong answer. `literal_bracket_patterns` governs snippet matching for the whole KB — every one of the ~150,000 snippets — so a change to it belongs in a PR where that is the change under review, not riding along in a 77-file curation diff where a reviewer is looking at ophthalmic evidence.

**So the ellipsis fix needs to land in two steps, not one.** Your diagnosis was right and my first fix was in the wrong place:

1. An infra PR adding `"^NM_\d"` to `literal_bracket_patterns`, so RefSeq variant nomenclature is treated as source text rather than as a curator gloss. Validated against current `main` the way that file documents its existing patterns were chosen — replaying every snippet in `kb/`: 145543 verified, 5439 skipped by prefix, zero failures, with the KB's editorial glosses still stripped.
2. Then a small curation change here (or after merge) deleting the seven ellipses, which only becomes correct once step 1 exists — deleting them today drops CRX from 32/32 to 25/32, because the ellipsis is what lets each fragment match around the stripped bracket.

The branch for step 1 is prepared locally; I've held off pushing it because my instructions scope pushes to this PR's branch, so that is the user's call.

**The staleness refresh survives the revert.** `d95cccdec` merged current `main` (`48bd0eb`) with no conflicts, `behind_by` is 0, and `check-snippet-grading --against-ref origin/main` is clean — both Appendiceal_Neoplasm and Bullous_Pemphigoid were the #10533 staleness you identified.

Net effect on this PR: unchanged content from the head you reviewed, plus the merge from main. Gates on the current head — snippet audit 1048/1073 (25 DOI-prefixed, hand-verified), duplicate-key, entity-ref, enum-value, causal-target, and grading checks all clean.

---

cmungall:
Thanks — both suggestions are handled, and your independent re-derivation of the bracket-stripping interaction is worth more than the approval.

**Suggestion 1, explaining the revert.** It was explained, in [this comment](https://github.com/monarch-initiative/dismech/pull/10537#issuecomment-5523990565) posted right after the push — your read of the intent is exactly right: the scope guard rejected `conf/reference_validator_config.yaml`, and a change governing snippet matching for all ~150,000 snippets shouldn't ride along in a content diff where nobody is looking at it. Your timing was just unlucky; the run started a minute or two after. The substantive half of your point stands regardless: the *history record* went with the revert, so the reasoning was left in commit messages and comments rather than anywhere durable. That is now fixed by the issues below.

**Suggestion 2, filing the pattern.** Filed as #10684, with the full diagnosis, the proposed one-line change, and both KB-wide replays (144043 and 145543 snippets, zero failures) so the validation does not have to be redone.

**Your two `supporting_text` mismatches are now filed too**, in #10685 alongside the DOI blind spot. I deliberately did not fix them here: both are pre-existing on `main`, neither line is touched by this diff, and a two-line push would dismiss this approval and start another review cycle for defects that are not this PR's. They make the case in that issue better than an abstract argument could — the checked slot (`evidence[].snippet`) quotes the same sentences correctly in both files, and only the unchecked slot is wrong.

That is the pattern across all three findings, and the reason the issues matter more than the fixes: every defect this audit found was in a place nothing checks. Four fabricated quotes behind a `DOI:` prefix the auditor skips, twelve `supporting_text` values in a slot nothing validates, and seven ellipses that exist only because the validator strips a bracket the source actually wrote. The entries that were fully covered by gates were clean.

---

cmungall:
The merge queue rejected this PR on one failing test, and I've fixed it in `e318f9622`.

The failure was real and it was mine:

```
tests/test_node_classes.py::test_committed_tree_examples_resolve_in_kb
  line 1631: FLVCR1-Related_Retinopathy has no pathophysiology node named
  'Genotype-Dependent Residual FLVCR1 Function Determines Ataxia Penetrance'
  (DISPOSITION > penetrance and expressivity modifier)
```

This PR renamed that node, because "Determines Ataxia Penetrance" asserted as established fact what is a hypothesis drawn from a single patient — the authors call it a hypothesis, so the entry now does too:

```
Genotype-Dependent Residual FLVCR1 Function Determines Ataxia Penetrance
  -> Proposed Residual FLVCR1 Function in Ataxia-Sparing Genotypes
```

`docs/superpowers/pathograph_node_classes.txt` cites that node by name as an example of the "penetrance and expressivity modifier" class, so the rename left the citation pointing at a node that no longer exists. The fix updates the citation. The renamed node still belongs in that class — it is exactly the "residual activity level standing between" disposition and effect that the class gloss describes — so this moves the name, not the entry.

### Why this PR's own CI never caught it

pytest is selected by changed path, and until now this PR touched only `kb/`, which matches neither the `python` nor the `schema` filter. So `tests/test_node_classes.py` never ran here. The merge queue runs the full suite, which is where it surfaced — 22 minutes in, after everything else passed.

Worth noting this is the same class of gap that `check-entity-refs` and `check-duplicate-keys` were made ungated whole-KB CI steps to close: a check that protects KB content, skipped by the content-only PRs it exists to protect. A renamed pathophysiology node breaking a `docs/` citation is not covered by either of those, and nothing ungated covers it today.

### On the scope label

I've added `scope-override`, and I want to be explicit about why, because I argued against using it earlier on this same PR.

That earlier case was a change to `conf/reference_validator_config.yaml` governing snippet matching for the whole KB. It was separable, so I reverted it and filed it as #10684. This one is not separable. The tree citation and the KB node name have to change in the same commit:

| What lands | Tree leaf | KB node | Result |
|---|---|---|---|
| `main` today | old name | old name | passes |
| docs fix alone | new name | old name | **`main` red** |
| this PR without the docs fix | old name | new name | **`main` red** |
| this PR as it now stands | new name | new name | passes |

Splitting it leaves `main` broken in the gap, whichever half goes first. That is what the label is for.

### Validation on the new head

Also merged current `main` (the branch was 124 commits behind) — no conflicts.

| Check | Result |
|---|---|
| `tests/test_node_classes.py` | 23 passed |
| schema validation, 20 entries | no issues |
| snippet audit | 1048/1073 (25 DOI-prefixed, hand-verified) |
| `check-enum-values` | clean |
| `check-duplicate-keys` | clean |
| `check-entity-refs` | clean |
| `check-causal-targets` | no new breakage |
| `check-snippet-grading --against-ref origin/main` | clean |

The diff is now 20 KB entries, 43 history records, 14 reference caches, and the one `docs/` line.

This push dismisses the approval, so it needs a fresh review.

---

cmungall:
@claude please re-review this PR — the previous review job crashed rather than reaching a verdict.

The `claude-review` run on `e318f96` ended with "Claude encountered an error after 4m 41s" and no findings. The job log tail is only post-job cleanup, so it is an agent-runtime failure, not a review that found something. This is the second such crash on this PR, and I cannot re-run it myself: the Actions re-run endpoints return `403 Resource not accessible by integration` for this token.

Everything else on `e318f96` is green:

- `test (3.13)` — success
- the curation-scope guard passes now, via the `scope-override` label

For context on what changed since the last approved head `d6f6f17`: the merge queue rejected the PR on one real test failure, `test_committed_tree_examples_resolve_in_kb`. This PR renamed a FLVCR1 pathophysiology node from "Genotype-Dependent Residual FLVCR1 Function Determines Ataxia Penetrance" to "Proposed Residual FLVCR1 Function in Ataxia-Sparing Genotypes" (the original asserted a single-patient hypothesis as established fact), and `docs/superpowers/pathograph_node_classes.txt` line 1631 still cited the old name. The fix updates that citation. It never appeared in this PR's own CI because pytest is selected by changed path and the PR touched only `kb/`; the merge queue runs the full suite.

So the diff since `d6f6f17` is one line in `docs/`, plus a merge of current `main` with no conflicts. The `scope-override` label is there because that one line cannot be split from the rename — landing either half alone leaves `main` red.

Gates I re-ran locally on the merged head: node-class tests 23 passed, schema validation clean, snippet audit 1048/1073 (25 DOI-prefixed, hand-verified), `check-enum-values`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, and `check-snippet-grading --against-ref origin/main` all clean.

---

cmungall:
Took the suggestion; pushed as `f131d4456`.

The CMV finding was right and worth fixing rather than waving through. The link's description claimed anti-CMV therapy can be withdrawn without progression, while the link's one evidence item quoted only the immune-recovery half. A description saying more than the quote attached to it is the exact defect this audit corrects elsewhere, so leaving it would have been inconsistent with the PR's own standard.

Added a second evidence item on the link, from the same paper:

> In some patients, retinitis has not progressed when specific anti-CMV therapy was discontinued

Verified as an exact substring of `references_cache/PMID_10665706.md`, and graded `SUPPORT` / `OTHER` to match the identical sentence already in the treatment-level block — so `check-snippet-grading` sees no new divergence. A history record for the round is in the same commit.

**The other findings I deliberately did not act on**, and why: the two `supporting_text` mismatches (EYS, MERTK) and the 22 stale-`PARTIAL` prose mentions are all pre-existing on `main` and untouched by this diff. They belong to #10685 and the #7439 sweep. Fixing them here would cost a review cycle for defects that are not this PR's.

Gates on the changed file: schema validation clean, snippet audit 37/37, duplicate-key, entity-ref, causal-target and enum-value checks clean, `check-snippet-grading --against-ref origin/main` no new divergences.

**On the approval, since the last round hit this:** the `@claude` action correctly noted it cannot submit a formal review. This push resolves that on its own — `claude-code-review.yml` triggers on `pull_request: [opened, synchronize]`, so it is now queued on `f131d4456` and can approve. That workflow had crashed twice before reaching a verdict (runs 33926839930 and 33936025738, the latter after posting its findings), and the Actions re-run endpoints return `403 Resource not accessible by integration` for my token, so a push was the only re-trigger available to me.

---

cmungall:
Resolved a merge conflict with `main` and pushed as `bd98d0c87`. Both sides are kept.

`main` moved 104 commits, and #10980 added a `review_notes:` no-link disposition (issue #8085) to the "Male Sex and Hormonal Factors" environmental entry in `Central_Serous_Chorioretinopathy.yaml` — landing on the exact lines this audit had rewritten. One conflict, and it is a real one rather than a formatting collision, because the two sides are editing the same evidence block for different reasons.

**This branch's side is the audit's fix, and that is checkable rather than a matter of preference.** `main` still carries the spliced snippet:

> Expression/splicing quantitative trait loci analyses showed association of identified GWAS hits with expression and/or splicing of genes in genital organs, potentially explaining the sex differences in CSCR. Men represent 72–87.5% of patients, but older women show similar prevalence rates to men.

That is two sentences stitched together, and it is not in the cited paper. I re-verified during the merge with an independent matcher (NFKC, dash and quote folding, whitespace collapse, plain substring):

| Snippet | Cache | Result |
|---|---|---|
| the real QTL sentence (kept) | `DOI_10.1038_s41598-025-92210-6.md` | exact match |
| the 72–87.5% figure (kept) | `DOI_10.1186_s40662-023-00349-y.md` | exact match |
| `main`'s spliced snippet (dropped) | `DOI_10.1038_s41598-025-92210-6.md` | **not found** |

So the audit's two evidence items are kept and `main`'s single item is dropped. Note where the second half of that spliced quote actually comes from — a different paper, which is why the fix needed two items rather than one.

**`main`'s `review_notes:` block is kept verbatim**, appended after them. Its prose describes the cited evidence as "a sex ratio, 72 to 87.5% male, plus an expression-and-splicing association" — both of which are still cited after the merge, now as two correctly attributed items instead of one spliced quote. So nothing in it argues for the version being dropped, and no trimming was needed. Dropping that block instead would have thrown away somebody's #8085 curation decision.

### Gates on the merged tree

| Gate | Result |
|---|---|
| `linkml-validate`, 20 entries | No issues found |
| snippet audit | 1049/1074 (25 DOI-prefixed, hand-verified above) |
| `tests/test_node_classes.py` | 23 passed |
| `check-environmental-evidence` | clean (4 waivers, including the one merged here) |
| `check-enum-values` | clean |
| `check-duplicate-keys` | clean |
| `check-entity-refs` | resolve |
| `check-causal-targets` | no new breakage |
| `check-snippet-grading --against-ref origin/main` | no new divergences |

Diff scope vs `main` is unchanged — 1 `docs`, 44 `history`, 20 `kb`, 14 `references_cache`, no deletions.

### One thing that needs a human

**`claude-review` has now failed three times without producing a verdict** on this PR — `33926839930` (4m41s), `33936025738` (posted findings, then errored at 43m59s), and `33949989961` (7m12s). Each ends with "Claude encountered an error" and a job log containing only post-job cleanup, so these look like agent-runtime failures rather than anything about the diff.

The substance has been reviewed: the run that did reach a verdict said **"looks good to merge"** with no critical or important issues, and its one suggestion is implemented. What is missing is only the formal approving review, which that run explicitly could not submit because it was the `@claude` action rather than the `ai4c-reviewer` app.

I have re-triggered what I can — a push re-fires `claude-code-review.yml`, and this merge does that again — but the Actions re-run endpoints return `403 Resource not accessible by integration` for my token, so if this run crashes too I have no further lever. @cmungall, if the fourth attempt also dies, this needs either a manual re-run of `claude-code-review.yml` or a maintainer review. I have not dismissed anything and will not.

---

PR #10539, opened 2026-09-02
Curate ibogaine mechanism hypotheses, knowledge gaps, and treatment on Opioid Use Disorder

cmungall:
Ibogaine is reported to interrupt opioid use disorder after a single dose, it is used for that in unregulated clinics, and nobody has established how it works. This PR curates the three competing accounts of its mechanism onto the Opioid Use Disorder entry as explicit rival hypotheses, records what each one is missing along with the experiment that would settle it, and adds ibogaine itself as a treatment wired into the disease's causal graph.

The finding worth leading with: **all three hypotheses are missing the same kind of experiment.** Nobody has ever microinjected ibogaine itself at a candidate site, or blocked a candidate receptor, while measuring opioid self-administration. Each mechanism has real evidence behind it, but for the wrong molecule (a congener), the wrong reinforcer (alcohol), or for sufficiency rather than necessity.

## Three competing hypotheses

Added to `kb/disorders/Opioid_Use_Disorder.yaml`, all `EMERGING`. Each carries PMID evidence with snippets verified against the reference cache, plus a `notes` field naming the experiment that would separate it from the other two.

| Hypothesis | Claim | Stated weakness |
|---|---|---|
| `ibogaine_gdnf_vta_neuroplasticity_model` | GDNF and Ret/ERK-MAPK in the ventral tegmental area carry the durable anti-relapse effect | The sufficiency and necessity experiments were done for **ethanol**, not opioids |
| `ibogaine_kappa_biased_agonism_model` | Noribogaine's G-protein-biased kappa agonism is the principal driver | Engineered analogues show kappa agonism is *sufficient*, not that ibogaine acts *through* kappa |
| `ibogaine_alpha3beta4_habenular_model` | Alpha3beta4 blockade in the medial habenula and interpeduncular nucleus damps the sensitized dopamine response | Nearly all in vivo work uses 18-MC, a congener that does not share ibogaine's neurotrophic route |

## Deep-research reports and assessments

One OpenScientist hypothesis-search run per hypothesis, under `kb/hypotheses/Opioid_Use_Disorder/<id>/`, each with an assessment sidecar recorded via the `review-hypothesis-exploration` skill.

Of roughly 37 PMIDs the reports introduced, 31 were checked against PubMed directly. All 31 resolve to real papers saying what was claimed — no fabricated citations. Six substantive problems were recorded rather than accepted:

- **A search failure presented as a knowledge gap.** The GDNF report flagged the self-amplifying VTA GDNF loop as unretrievable. It was demonstrated in vivo in 2011, with shRNA knockdown abolishing the durable effect (PMID:22238721). The report named the right lab and the wrong journal.
- **Two citations mischaracterized in the hypothesis's favour** in the alpha3beta4 report: a paper cited as supporting evidence whose own conclusion is the opposite, and a human genetics paper credited with an opioid association its abstract does not report.
- **Sample-size inflation on a load-bearing result** in the kappa report: an antidepressant finding attributed to n=14 is actually n=8 completers, which matters because the report's central argument rests on it.
- **An unqualified selectivity claim**, and separately **one mis-graded evidence type** and **proposed snippets containing ellipses**, which are not exact substrings and cannot be promoted as written.

Each assessment reaches `PARTIALLY_SUPPORTED`. None of the three hypotheses was promoted on the strength of a provider report.

## Six curated knowledge gaps

`discussions` entries with structured `proposed_experiments` — perturbations, controls, readouts, `decision_criterion`, and `would_support` / `would_refute` pointing at the hypotheses each result would settle.

Five are `KNOWLEDGE_GAP`: the shared necessity gap above (two experiments, a site dissociation and a factorial blockade); the fact that a selective alpha3beta4 antagonist exists and has never been run against opioid self-administration; the untested pharmacokinetic alternative to durability; the unmeasured human neurotrophin response; and the untested state-dependence that would resolve the kappa polarity question.

One is `HUMAN_MODEL_MISMATCH`: rodent and uncontrolled human data report large effects, while the one randomized trial in opioid-dependent patients found non-significant withdrawal reductions, with seven of forty preclinical studies null and the strongest recent result obtained in male rats only. Evidence is not absent here — its translational validity is the open question, which is why it is not filed as a plain knowledge gap.

## Ibogaine as a treatment

Four `target_mechanisms` links into the pathograph: opioid withdrawal syndrome and compulsive opioid seeking (`INHIBITS`), hyperkatifeia (`MODULATES`), and VTA dopamine release (`INHIBITS`).

Hyperkatifeia is `MODULATES` rather than `INHIBITS` on purpose. The validated pharmacology for treating that state is kappa *antagonism*, so asserting `INHIBITS` would encode one side of a live dispute as fact.

The entry says first that ibogaine is not approved for this indication anywhere. Treatment-level evidence carries the safety record alongside the efficacy claims: 50% of monitored subjects exceeding QTc 500 ms, the iPSC-cardiomyocyte repolarization mechanism that explains why arrhythmic events appear days after dosing, 33 documented deaths, and the CYP2D6 clearance dependence. It also records that ibogaine is a mu-opioid *antagonist*, so withdrawal relief cannot be the agonist substitution that explains methadone and buprenorphine two entries above it.

Two things were deliberately left out, both recorded in `notes` so they don't read as oversights: a link to opioid-induced hyperalgesia (that result belongs to the oxa-noribogaine analogue, not to ibogaine), and treatment entries for the safer analogues (18-MC, tabernanthalog, oxa-iboga), none of which has been tested in humans for this indication.

## Validation

`just validate` passes at 152/152 snippets verified with term validation clean. Also clean: `check-entity-refs`, `check-duplicate-keys`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-reference-titles`, `check-term-cache-integrity`, and the treatment/entity-ref/evidence pytest lane (12305 passed).

Three history records cover the three KB-entry commits. The term caches gained one row each (`CHEBI:5852`, ibogaine) through the validator rather than by hand.

`check-environmental-evidence` reports 33 findings. All are in files this branch does not touch — that gate cannot grandfather against a ref, so they are pre-existing on `main` rather than introduced here.

https://claude.ai/code/session_014bevqoUjdynxNgqiqxdKXd

### DIFF
--- a/kb/hypotheses/Opioid_Use_Disorder/ibogaine_alpha3beta4_habenular_model/assessments/openscientist-assessment-by-claude-code.yaml
+++ b/kb/hypotheses/Opioid_Use_Disorder/ibogaine_alpha3beta4_habenular_model/assessments/openscientist-assessment-by-claude-code.yaml
@@ -22,10 +22,10 @@ summary: >-
   opioid-dependence association its abstract does not report. A third problem
   runs the other way and is more interesting: the report declares ibogaine
   non-reinforcer-selective without the temporal qualification that makes the
-  claim false for the effect actually under discussion. The provider returned an
-  HTML and PDF report bundle under `../openscientist_artifacts/`, which
-  `.gitignore` excludes from the repository, so no `artifacts` path is declared
-  here; the assessed content is the committed `openscientist.md`.
+  claim false for the effect actually under discussion.
+artifacts:
+  - ../openscientist_artifacts/final_report.html
+  - ../openscientist_artifacts/final_report.pdf
 claims:
   - claim_id: congener-rescoping
     statement: >-
--- a/kb/hypotheses/Opioid_Use_Disorder/ibogaine_gdnf_vta_neuroplasticity_model/assessments/openscientist-assessment-by-claude-code.yaml
+++ b/kb/hypotheses/Opioid_Use_Disorder/ibogaine_gdnf_vta_neuroplasticity_model/assessments/openscientist-assessment-by-claude-code.yaml
@@ -24,10 +24,12 @@ summary: >-
   release and conversion to noribogaine) that directly competes with the
   induced-plasticity premise, and the report does not engage it. One evidence
   row is mis-graded and the proposed curation snippets are elided and cannot
-  be used as written. The provider returned an
-  HTML and PDF report bundle under `../openscientist_artifacts/`, which
-  `.gitignore` excludes from the repository, so no `artifacts` path is declared
-  here; the assessed content is the committed `openscientist.md`. This run also emitted no `openscientist.md.citations.md` sidecar, unlike the two sibling runs, so its citations are checkable only from the report body.
+  be used as written. This run emitted no `openscientist.md.citations.md`
+  sidecar, unlike the two sibling runs, so its citations are checkable only
+  from the report body.
+artifacts:
+  - ../openscientist_artifacts/final_report.html
+  - ../openscientist_artifacts/final_report.pdf
 claims:
   - claim_id: ethanol-only-scope
     statement: >-
--- a/kb/hypotheses/Opioid_Use_Disorder/ibogaine_kappa_biased_agonism_model/assessments/openscientist-assessment-by-claude-code.yaml
+++ b/kb/hypotheses/Opioid_Use_Disorder/ibogaine_kappa_biased_agonism_model/assessments/openscientist-assessment-by-claude-code.yaml
@@ -24,10 +24,10 @@ summary: >-
   between two marginal findings is worth recording; it is not the decisive
   conceptual argument the report presents. Two smaller issues: a single-author
   hypothesis paper is used as corroboration, and the stated corpus size exceeds
-  the deposited citation manifest. The provider returned an
-  HTML and PDF report bundle under `../openscientist_artifacts/`, which
-  `.gitignore` excludes from the repository, so no `artifacts` path is declared
-  here; the assessed content is the committed `openscientist.md`.
+  the deposited citation manifest.
+artifacts:
+  - ../openscientist_artifacts/final_report.html
+  - ../openscientist_artifacts/final_report.pdf
 claims:
   - claim_id: kappa-necessity-untested
     statement: >-


cmungall:
All four items addressed in 9791fb4e2. The CI failure was the same root cause as the first critical item, so both are fixed by the rebase.

**One correction to the review.** For the artifacts item you offered two options — commit the three HTML files, or drop the block. Committing them is not available: `.gitignore` line 243 excludes `kb/hypotheses/**/openscientist_artifacts/`, so the provider bundle is deliberately kept out of the repository. I took the other option. Each assessment's `summary` now records that the provider returned an HTML and PDF bundle the repository excludes, and that the assessed content is the committed `openscientist.md`. All three now pass `just validate-hypothesis-assessment`.

### `supports: PARTIAL` — 18 converted, but two are `REFUTE`

You expected most to be the first row of your table, and 16 were: `SUPPORT`, with `directness: INDIRECT` on the 11 where the quote reaches the claim through a real inference step (a different reinforcer, an analogue compound, an in vitro measurement standing in for an in vivo state).

Two are not, and converting them to `SUPPORT` would have been wrong:

| Item | Why `REFUTE` |
|---|---|
| PMID:30890941 on the GDNF hypothesis | Broad BDNF induction across NAcc, SN and PFC contradicts attributing the durable effect *specifically* to VTA GDNF |
| PMID:21040239 on the alpha3beta4 hypothesis | The noribogaine-versus-18-MC dissociation contradicts transferring the model from the congener to ibogaine |

Both were curated as `PARTIAL` to mean "cuts against part of this", which is direction, not degree. Both `explanation` fields are reworded to say what is contradicted.

I also repaired a pre-existing line in the entry's `notes` that named `supports: PARTIAL` as a curation todo. The underlying gap — syndrome-level rather than sign-specific citations on some withdrawal signs — is unchanged; only the retired enum name is gone.

### `hypothesis_groups` — your counter-argument is the right one, now recorded

I checked, and `hypothesis_groups` exists only on `CausalEdge`. `TreatmentMechanismTarget` has `target`, `treatment_effect`, `description`, `evidence` and nothing else, so a treatment mechanism link has no slot to opt into a group with.

Opting the pathogenesis edges in instead would be worse than leaving it empty: those edges describe how untreated opioid use disorder progresses, and none of them is disputed by the three competing accounts of ibogaine's mechanism. The competition is over drug action, and it is recorded where it lives — in the hypothesis descriptions, in the `discussions` gaps that name the discriminating experiments, and in each proposed experiment's `would_support` / `would_refute`. That reasoning is now in the entry's `notes`.

### Citations sidecar — recorded rather than manufactured

There is no way to regenerate one. The sidecar is written at generation time by `--separate-citations`, and no `deep-research-client` subcommand extracts citations from an existing report. Re-running the job would produce a *different* report, which the committed assessment no longer describes — so the assessment would have to be redone to match, for a consistency gain. Hand-writing the file would fabricate a provider artifact.

The GDNF assessment's `summary` now records that this run emitted no sidecar, unlike its two siblings, so its citations are checkable only from the report body.

### Validation

`check-enum-values` clean across 2922 files — that is the gate I should have run before opening this, and did not. `just validate` at 152/152 snippets verified with term validation clean; `validate-hypothesis-assessment` on all three sidecars; entity-refs, duplicate-keys, folded-hyphens, snippet-length, title-snippets, snippet-grading, reference-titles, term-cache-integrity, `validate-history-all`; and the entity-ref/evidence/treatment/hypothesis pytest lane at 15986 passed.

The diff is still 46 files and 9044 insertions with no deletions, so the rebase dropped nothing.

---

cmungall:
Correcting myself: the claim in my last comment that committing the artifact files "is not available" was wrong, and 8dec0be29 fixes it.

`.gitignore` on `main` changed between the base I first checked and the base I rebased onto. The blanket `kb/hypotheses/**/openscientist_artifacts/` exclusion is gone, replaced by rules that only exclude `raw/`, `local/` and `controlled/` subdirectories. So two things happened that I did not notice: the previous commit's `git add` silently committed all six report files, and the three assessment summaries went on asserting that the repository excludes them — a false statement in curated content.

Committing them is the convention here, not an exception: 116 artifact files are already tracked under `kb/hypotheses/`. So this takes your first option after all. The six files stay, each assessment declares both the HTML and the PDF, and the false sentence is gone from all three summaries.

The GDNF assessment keeps its separate note that this run emitted no `openscientist.md.citations.md` sidecar. That one is still true and has nothing to do with the artifact bundle — it remains the one 🟡 item I am recording rather than fixing, for the reason in my previous comment.

`validate-hypothesis-assessment` passes on all three sidecars with the declarations restored, and `check-enum-values`, `check-folded-hyphens` and `just validate` are clean.

---

PR #10555, opened 2026-09-02
curate(Spinocerebellar_Ataxia_Type_31): add SCA31 BEAN1 repeat-RNA entry (MONDO:0007296)

cmungall:
New disease entry for spinocerebellar ataxia type 31, an adult-onset autosomal dominant cerebellar ataxia with a strong Japanese founder effect. It is the first dismech entry for the disease; `Autosomal_Dominant_Cerebellar_Ataxia_Type_III` mentions SCA31 as a member and is left untouched, following the pattern already used for SCA2, SCA17, SCA23 and SCA43.

## What the mechanism is

The causal lesion is a 2.5-3.8 kb insertion of complex pentanucleotide repeats at 16q22.1, sitting in an intron shared by BEAN1 and TK2, which are transcribed in opposite directions. Only the `(TGGAA)n` motif segregates with disease; the other motifs in the same insertion turn up on control chromosomes, and rare control insertions of comparable size lack `(TGGAA)n` entirely. So insertion size alone does not make an allele pathogenic, and a fragment-length assay that reports only size cannot make the call.

The pathophysiology is curated as a seven-node chain with a branch point. The insertion is transcribed bidirectionally; the BEAN1-direction `(UGGAA)n` transcript is brain-restricted and is the toxic species. From there it branches: nuclear RNA foci in Purkinje cells on one arm, repeat-associated translation of a poly-WNGME pentapeptide protein on the other. Both arms converge on Purkinje cell degeneration, cerebellar cortical atrophy and progressive ataxia. Three nodes declare `conforms_to` against `cerebellar_purkinje_degeneration`.

Three things the entry is deliberately careful about, because the shorthand summaries get them wrong:

- **It is not simple splicing-factor sequestration.** SRSF1/SRSF9 do bind the repeat, and SRSF9 co-precipitates with it independently. But TDP-43, FUS and hnRNPA2B1 act as RNA *chaperones* — raising their level suppresses toxicity rather than worsening it. The node is curated as an altered RNA-protein equilibrium, and no splicing target has been shown mis-spliced in human SCA31 cerebellum.
- **The translation route is unresolved.** SCA31 is often described as a RAN-translation disorder. The authors who found the PPR protein explicitly decline that attribution, because the TGGAA repeat contains its own AUG and the flanking motifs supply stop codons.
[...]

PR #10551, opened 2026-09-02
curate(SPOP-Related_Neurodevelopmental_Disorder): add SPOP germline NDD entry (MONDO:0032942)

cmungall:
Adds a new disorder entry for SPOP-related neurodevelopmental disorder (Nabais Sá-de Vries syndrome, NSDVS), a dominant disorder caused by de novo missense variants in SPOP, the substrate-recognition adaptor of the CUL3 ubiquitin ligase. The defining 2020 cohort study (PMID:32109420) split the variants into two functional classes with opposite effects: gain-of-function variants (p.Arg121Gln, p.Asp144Asn) increase substrate degradation and track with congenital microcephaly (type 1, OMIM:618828, this entry's MONDO term), while dominant-negative variants (p.Thr25Ala, p.Tyr83Cys, p.Gly132Val, p.Arg138Cys) impair degradation and track with relative macrocephaly and hypertelorism (type 2, OMIM:618829, MONDO:0032943, carried here as a subtype).

Closes #10547

## What the entry contains

- **Pathophysiology as a branching chain**: de novo SPOP missense variation splits into two mechanism nodes — increased vs impaired CUL3-SPOP substrate degradation, each carrying `genetic_context.functional_impact_category` (`GAIN_OF_FUNCTION` / `DOMINANT_NEGATIVE`) — leading to BET protein (BRD2/3/4) depletion vs accumulation (PMID:28805821), then to aberrant neurodevelopment. The substrate-to-brain edges are marked `INDIRECT_UNKNOWN_INTERMEDIATES` and grouped under an `EMERGING` mechanistic hypothesis (`bet_bidirectional_dysregulation`), because the functional assays were run in endometrial cancer cell lines, not neural cells.
- **Two types as `has_subtypes`** (Type 1 = MONDO:0032942, Type 2 = MONDO:0032943), with phenotypes keyed to subtype where the literature splits them (microcephaly/hearing loss/blepharophimosis → Type 1; relative macrocephaly/hypertelorism → Type 2).
- **Full published variant spectrum** (10 variants) under `genetic.variants`, including the later case-report variants p.Arg121Trp, p.Met117Ile, p.Cys23Arg, and the BACK-domain truncation p.Tyr353Ter that produced mixed type 1/2 features.
[...]

PR #10549, opened 2026-09-02
curate(Hereditary_Sensory_Neuropathy_Type_1D): add ATL1 HSN1D entry (MONDO:0013381)

cmungall:
## Summary

New disease entry for **hereditary sensory neuropathy type 1D** (HSN1D / HSAN1D, MONDO:0013381; OMIM 613708), the ATL1 form of autosomal dominant hereditary sensory neuropathy.

HSN1D is an adult-onset, autosomal dominant, ulcero-mutilating axonal sensory neuropathy caused by heterozygous variants in ATL1, encoding atlastin-1, a dynamin-family GTPase that mediates homotypic fusion of endoplasmic reticulum tubules. ATL1 is also the SPG3A hereditary spastic paraplegia gene, so HSN1D and SPG3A are allelic disorders.

I curated this as a dedicated per-type file rather than folding it into the existing `Hereditary_Sensory_and_Autonomic_Neuropathy` umbrella entry, which does not cover the ATL1 form, and following the precedent of the standalone `Hereditary_Sensory_and_Autonomic_Neuropathy_Type_2` file.

### Pathophysiology chain (ontology-grounded, causally linked)

1. Heterozygous ATL1 variant with reduced atlastin-1 GTPase activity (MOLECULAR)
2. Impaired homotypic ER membrane fusion (MOLECULAR)
3. Disrupted tubular ER network (CELLULAR)
4. Length-dependent distal degeneration of long DRG sensory axons (TISSUE)
5. Distal sensory loss

Phenotypes: distal sensory loss (diagnostic), painless skin ulcers, osteomyelitis / distal amputations, and pyramidal-tract features (documented in the N355K family that overlaps the SPG3A phenotype).

### Allelic disorder knowledge gap

A `KNOWLEDGE_GAP` discussion records the open question of why the same atlastin-1 GTPase defect targets long peripheral sensory axons in HSN1D but corticospinal motor axons in SPG3A — sometimes overlapping within one family for the recurrent N355K allele. Recorded honestly rather than asserting a resolved mechanism.

### Key references

- **PMID:21194679** — Guelly et al. 2011, Am J Hum Genet: identifies ATL1 mutations as a cause of HSN I; reduced GTPase activity and disrupted ER morphology; HSN I and SPG3A allelic.
- **PMID:22340599** — Leonardis et al. 2012: N355K family clinical/electrophysiologic description (painless ulcers, amputations, pyramidal features, axonal NCS).
[...]

PR #10557, opened 2026-09-02
curate(Combined_Oxidative_Phosphorylation_Defect_Type_13): add PNPT1 COXPD13 entry (MONDO:0013977)

### LINKED ISSUE #10544: Curate combined oxidative phosphorylation defect type 13 (MONDO:0013977)
Curate a dismech entry for **combined oxidative phosphorylation defect type 13** ([MONDO:0013977](https://monarchinitiative.org/MONDO:0013977)) — an autosomal recessive mitochondrial disease caused by biallelic PNPT1 variants; loss of polyribonucleotide nucleotidyltransferase impairs mitochondrial RNA import/degradation, producing combined respiratory-chain deficiency with severe encephalomyopathy, choreoathetosis, and deafness.

Note: PNPT1 is also the DFNB70 isolated-deafness gene — a distinct allelic entity; keep the entries separate if both are curated.

Stub: `stubs/Combined_Oxidative_Phosphorylation_Defect_Type_13.yaml`

cmungall:
New entry for combined oxidative phosphorylation defect type 13 (MONDO:0013977, OMIM 614932) — the PNPT1/PNPase mitochondrial disease. Deletes the stub it curates.

## The mechanism has two branches, and that is the main curation decision here

PNPase does two jobs from one enzyme, so breaking it breaks two things at once. The entry draws both branches from a single molecular lesion rather than picking one:

- **Respiratory-chain branch.** Loss of PNPase trimer assembly and catalysis → impaired import of small nuclear-encoded RNAs (5S rRNA, RNase MRP and RNase P RNA) *and* defective mtRNA processing and turnover by the SUV3–PNPase degradosome → impaired mitochondrial translation → combined complex I/III/IV deficiency → tissue energy failure in basal ganglia, cochlea, white matter and optic pathway.
- **Innate-immune branch.** The same degradosome loss lets mitochondrial double-stranded RNA accumulate and escape to the cytosol, where MDA5 drives a type I interferon response. This is documented in PNPT1 patients, not just in knockdown cells — one patient met criteria for Aicardi-Goutières syndrome, and two had interferonopathy-pattern cystic leukoencephalopathy.

Nobody has measured both arms in the same patients and related them to which features those patients have, so the interferon node is typed as a `modifier` with `PROVISIONAL` confidence and the question is recorded as a discussion rather than answered. That is the weakest claim the evidence supports.

## PNPT1 carries three inherited diseases; they are kept separate

Only the recessive combined-OXPHOS one is curated here. The other two are cross-referenced under `differential_diagnoses` with the features that tell them apart:

- **DFNB70** (OMIM 614934), recessive nonsyndromic deafness. This boundary is genuinely soft — one family's "isolated" congenital deafness became ataxia, dystonia and cognitive decline in their forties. The entry records the softness rather than merging the entities, since merging would discard the OMIM distinction on one family's evidence.
- **SCA25**, dominant spinocerebellar ataxia from heterozygous truncating and splice alleles in the S1 domain. A firmer separation: different zygosity, allele class, domain and inheritance. Worth flagging because SCA25 also has a raised interferon signature, so "PNPT1 variant + interferon score" does not by itself mean COXPD13.

## Two things a reader might otherwise get wrong

- **A normal respiratory-chain assay does not exclude this diagnosis.** In one genetically confirmed family, blood lactate, CSF lactate and muscle enzymology were all normal. PNPase's RNA-processing function is the more sensitive readout, which inverts the usual mitochondrial workup order — that is curated in `diagnosis` and cited to the paper that states it.
- **Complex V.** The standard summary of this disease class is "complexes I, III, IV and V down, II spared". The only patient study that measured all five found complexes II *and V* unchanged. One family, steady-state protein rather than enzyme activity, so possibly a method artefact — but the entry does not propagate the summary as if it had been observed, and a `KNOWLEDGE_GAP` discussion says so.

## Contents

21 phenotypes, 12 pathophysiology nodes, 5 variants, 3 biochemical markers, 1 animal model with 2 mechanism links and 2 readouts, 3 diagnostic entries, 4 differential diagnoses, 3 treatments, 4 discussions. 17 references, all PMIDs.

Module conformance is declared only at `mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress`, and only for its bioenergetic-decline arm — the module's age-related-damage upstream node does not apply to a primary nuclear-gene RNA-metabolism defect. `innate_antiviral_interferon_response` was considered for the interferon node and rejected: its nodes are about pathogen sensing and viral restriction, and this is sterile self-RNA sensing with no virus. A mitochondrial-translation module and a sterile-interferonopathy module would both fit; neither exists, and the sibling COXPD21 and COXPD26 entries noted the same gap.

No `datasets` block. With roughly 15–24 reported patients worldwide there is no disease-specific dataset, and a gene-only search on PNPT1 surfaces the cancer literature — Named Entity Confusion reached through dataset search.

## Deep research

A falcon (Edison) run is committed under `research/`. It passed `just preflight-dr` against MONDO:0013977 (PNPT1 mentioned 50 times, OMIM 614932 agreeing), with 6/6 citations resolved and no unresolved terms. The four "mislabelled terms" in its frontmatter are the validator reading a table's commentary column as a term label, not real mis-bindings.

The report cites by author-year key rather than by PMID, so its leads were resolved through DOI and the PMC ID converter before use — that is where PMID:30024931 (the Shimada mouse Pnpt1 knockout) came from. Every other reference was found independently through PubMed. Every snippet in the entry was copied from the cached source, not from the report.

GeneReviews: searched for PNPT1 and for COXPD13, none exists. Recorded as a checked negative in the entry notes.

## Validation

| Check | Result |
|---|---|
| `just validate-disorders` (schema + terms + references, batched) | passed |
| `just count-verified-snippets` | 87/87 verified |
| `just check-snippet-length` | passed |
| `just check-title-snippets` | passed |
| `just check-snippet-grading` | passed |
| `just check-reference-titles` | passed |
| `just check-term-cache-integrity` | passed |
| `just compliance` | 89.9% global, 90.2% weighted |

`check-snippet-grading` caught one real defect during curation: the same Rius 2019 sentence reports a fibroblast finding and a blood finding, and I had graded it `IN_VITRO` in one place and `HUMAN_CLINICAL` in another. Fixed by splitting the quote so each evidence item cites its own half.

Closes #10544

https://claude.ai/code/session_01LJQrSkzPDMk5d7S8dw1c2D

---
_Generated by [Claude Code](https://claude.ai/code/session_01LJQrSkzPDMk5d7S8dw1c2D)_


cmungall:
The `claude-review` run on this PR failed for infrastructure reasons, not content: it died with "error after 16s" at ~14:05 UTC in a provider rate-limit window that killed the reviewer on #10549, #10551 and #10555 within the same minutes. No review was produced; the PR is unchanged and all local gates pass (87/87 snippets verified, `validate-disorders` green).

This session cannot re-run the workflow (`workflow_dispatch` and job re-run both return 403 for its token). A maintainer can re-trigger with:

```
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10557
```

I'll keep the PR watched and act on the review when it lands.

---

cmungall:
All four review items are addressed in one push (7ebdaa2f1). One of them turned up a wrong CURIE in the suggestion itself, so please read that part before re-reviewing.

**One correction to flag first.** The review suggested `CHEBI:9639` for topiramate. That CURIE resolves to **Tos-Arg-CH2Cl**, an unrelated chloromethyl ketone — not a drug. I verified against CHEBI and used `CHEBI:63631` (topiramate) instead. `CHEBI:6437` (levetiracetam) was correct as suggested. Worth noting because a plausible-looking wrong CHEBI code in a `therapeutic_agent` slot is exactly the class of defect `just check-qualifier-terms` was built for, and `therapeutic_agent` *is* covered by `validate-terms` — but only for label agreement, which a wrong-but-consistent binding would still pass if the label were copied from the wrong term.

**Feeding difficulty and dysphagia (the blocking item).** Added both, not one. The cited review names them as separate items in the same sentence, and gives dysphagia weight the feature list does not convey: the authors single out dysphagia together with dystonia in the neonatal or infantile period as the combination that should prompt genetic testing. So dysphagia is a diagnostic trigger, not only a burden, and it carries that prognostic sentence as its own evidence item. Feeding difficulties carries the index patient's own recorded poor feeding at six months rather than the literature summary.

They are chained into the pathograph rather than left as loose leaves:

`Basal Ganglia and Brainstem Neuronal Injury` → `Dysphagia` → `Feeding difficulties` → `Growth delay`

That gives Growth delay the upstream it was missing. The first edge is `INDIRECT_KNOWN_INTERMEDIATES` with bulbar dysfunction named as the intermediate — brainstem signal change and bulbar dysfunction are both reported, and dysphagia is reported, but no source measures a swallow against a brainstem lesion in one patient, so I did not draw it as direct.

**Infantile spasms.** Added `HP:0012469` alongside the generic Seizure, with both the formal diagnosis and the semiology (nodding, flexion of both upper limbs, clenched fists, in clusters). Linked to the antiseizure treatment via `target_mechanisms` with `treatment_effect: INHIBITS`. The schema explicitly allows a phenotype target for a symptomatic treatment, which this is — and that placement is the honest one, since nothing in this disease's causal chain is treated by an antiseizure drug.

**Therapeutic agents.** Levetiracetam and topiramate only. ACTH is left out of the structured field as you asked and named in the description prose instead, so the slot cannot be read as a therapy this disease responds to.

**Muscle weakness.** Added, with the graded strength examination (3–4 in the upper limbs, 0 in the lower limbs) as a second evidence item alongside the narrative one, and drawn downstream of cellular energy failure — skeletal muscle is already one of the oxidative-demand tissues that node names.

**What I deliberately did not add.** Nystagmus and muscle atrophy. Both appear only inside the same composite feature-list sentence, with no per-patient observation anywhere in the cached text. Curating them would mean three or four phenotypes resting on one quote that names none of them individually, and neither changes a mechanism or a management step the entry does not already state. If a later case series reports either with its own observation, they are cheap to add then.

**Validation**

| Check | Result |
|---|---|
| `just validate-disorders` (schema + terms + references, batched) | passed |
| `just count-verified-snippets` | 97/97 verified (was 87/87) |
| `just check-causal-targets` | passed |
| `just check-enum-values` | passed |
| `just check-qualifier-terms` | passed |
| `just check-snippet-grading` | passed |
| `just check-title-snippets` | passed |
| `just check-snippet-length` | passed |
| `just check-reference-titles` | passed |
| `just normalize-cache` + `just check-term-cache-integrity` | no churn, passed |
| `just compliance` | 89.7% global, 90.0% weighted |

All six new CURIEs were already cache-backed with matching canonical labels, including the two CHEBI ones in the `chemicalentityterm` enum cache, so `validate-terms` genuinely checked them rather than skipping — that is why there is no cache diff in this push.

A round-2 history record is added rather than the existing one edited, since history records are append-only.

---

PR #966, opened 2026-04-06
[codex] Add pathograph CX2 export for NDEx

cmungall:
## Summary

This PR adds a pathograph-to-CX2 export flow for publishing dismech networks to NDEx.

It includes:

- a new `dismech-cx2` exporter for converting disorder pathographs into CX2
- NDEx upload support and `just` targets for test-server publishing
- NDEx/iQuery-friendly handling for gene-referencing nodes while preserving all pathograph node types
- richer ontology-linked metadata on event nodes, including CL, GO, molecular function, and UBERON/location links
- bulk export/upload behavior that skips disorders with no pathograph edges instead of aborting
- internal operator docs for the CX2/NDEx workflow in `docs/`

## Why

We need a reproducible path from dismech pathographs to NDEx-publishable CX2 so disorder mechanism networks can be uploaded, browsed, and queried directly in NDEx.

## Notes

- NDEx upload visibility now defaults to `PUBLIC`, with explicit override still available.
- Generated `elements/` site output from local docs builds was intentionally not included in this PR.
- The uploader treats post-upload `index_level=META` failures as non-fatal because the NDEx test server can return a server-side error after the network itself has already been created.

## Validation

- `uv run pytest tests/test_cx2_export.py -q`
- `uv run --group dev ruff check src/dismech/export/cx2_export.py tests/test_cx2_export.py`
- `uv run mkdocs build`
- `just export-cx2-all -o /tmp/cx2`
- live upload smoke test against `https://test.ndexbio.org`

## Impact

This gives us a documented, repeatable NDEx publishing workflow for dismech pathographs and makes uploaded networks substantially more useful for NDEx-side search and inspection.

PR #10558, opened 2026-09-02
Add guarded NDEx production publishing

cmungall:
Adds the production release path proposed in #966: a manually triggered, protected GitHub Actions workflow that audits the complete CX2 corpus, uploads privately, verifies every network, and only then allows batch publication.

The release manifest preserves stable NDEx UUIDs across versions, records source/output hashes and counts, and supports restart after an interrupted upload. Production metadata is explicit, source links are commit-pinned, and ordinary test uploads now default to private.

The current full-corpus audit exports 2,452 of 2,515 disorders and intentionally blocks publication on 106 findings: 72 networks with orphan nodes, one unknown node, and 33 missing disease metadata records. A failed audit still retains the manifest artifact as the cleanup worklist.

Validation:
- `uv run pytest tests/test_cx2_export.py tests/test_ndex_publish.py -q` (15 passed)
- pre-commit hooks on all changed files (passed)
- `just test-code` (11,609 passed, 41 skipped; one unrelated existing macOS case-insensitive-filesystem failure in `test_prune_keeps_a_differently_named_page_that_is_the_same_file`)

Relates to #966.

### DIFF
--- a/.github/workflows/publish-ndex.yml
+++ b/.github/workflows/publish-ndex.yml
@@ -82,6 +82,7 @@ jobs:
           : "${NDEX_AUTHOR:?Configure the NDEX_AUTHOR repository variable}"
           : "${NDEX_RIGHTS:?Configure the NDEX_RIGHTS repository variable}"
           : "${NDEX_RIGHTS_HOLDER:?Configure the NDEX_RIGHTS_HOLDER repository variable}"
+          : "${PREVIOUS_MANIFEST:?previous_manifest must name the reserved registry path}"
           if [[ "$FIRST_RELEASE" == "true" ]]; then
             if [[ -e "$PREVIOUS_MANIFEST" ]]; then
               echo "Refusing first release because $PREVIOUS_MANIFEST already exists" >&2
@@ -124,6 +125,7 @@ jobs:
     if: inputs.upload
     needs: export
     runs-on: ubuntu-latest
+    timeout-minutes: 300
     environment: ndex-production
     steps:
       - name: Check out requested revision
--- a/docs/cx2-ndex-publishing.md
+++ b/docs/cx2-ndex-publishing.md
@@ -196,9 +196,10 @@ The workflow has two jobs:
    `ndex-production` GitHub environment. It uploads to
    `https://www.ndexbio.org`, verifies the NDEx network summaries, and changes
    newly created or previously private networks to public only after the
-   complete staged upload has succeeded. Existing public networks remain public
-   during a successful update; if one fails verification, that network is made
-   private before the workflow stops.
+   complete staged upload has succeeded. For a `PUBLIC` release, existing public
+   networks remain public during a successful update; if one fails verification,
+   that network is made private before the workflow stops. A `PRIVATE` release
+   intentionally makes every network it touches private.
 
 Configure these repository variables before running the workflow:
 


cmungall:
All five required publication-safety findings are addressed in c34b76b00d.

- The publisher verifies every pending CX2 file against its audited SHA-256 before making any NDEx change, and independently refuses manifests with export defects unless its explicit override is used.
- Existing public networks now remain public during successful in-place updates. Newly created or previously private networks stay private until the full staged release verifies. A verification failure makes only the affected network private before stopping.
- The manifest reports UUID-bearing slugs that disappeared from the export as retired; retirement remains an explicit operator decision.
- Tests now cover the persisted defect worklist, hash rejection, retirement detection, first-release UUID creation/checkpointing, resume, and preservation of public visibility.
- The workflow uses the exact audited commit in both jobs, passes choice inputs through environment variables, and emits a compact verified UUID registry instead of asking maintainers to commit the full run manifest.

I also removed the unused generic-uploader parameters, removed the redundant privacy change, fail fast on permanent completed-summary mismatches, use an explicit repository blob base for source links, and converted the exception test to pytest.raises.

I intentionally kept NDEx warnings as a release failure: this is a production gate, and silently accepting an unreviewed server warning would weaken it. I also kept networkType as a scalar. With the installed ndex2 serializer, a Python list is declared as string and serialized as the literal bracketed text; NDEx documents networkType as free text, so the scalar produces the correct CX2 value.

Validation: 19 focused CX2/publisher tests pass, 19 workflow/config tests pass, and all changed-file pre-commit hooks pass.

cmungall:
The three non-blocking re-review notes are now resolved in 2afd4b209c. CX2 and manifest writes explicitly use UTF-8; permanent completed-summary mismatches must repeat on two consecutive polls before failing; and the build test now proves networkType is emitted as a scalar with a CX2 string declaration.

Validation: 20 focused tests pass and all changed-file pre-commit hooks pass.

cmungall:
The three operator-path blockers are addressed in 3ab52c1af4. A requested registry path must now exist; the workflow has an explicit first_release switch instead of silently falling back to minting; and the job summary reports carried UUID and retirement counts. The operator guide now requires environment reviewers and gives the exact failed-manifest recovery sequence.

I also removed the obsolete UPLOADED_PRIVATE state, made requested PRIVATE visibility override stale resume metadata, and clear an optimistic verified status after any failed re-verification. The final public promotion remains sequential because NDEx has no atomic corpus-wide visibility operation; each network is verified and checkpointed, and the interrupted-release procedure covers partial runs.

Validation: 40 focused and workflow/config tests pass, including the new missing-registry case, and all changed-file pre-commit hooks pass.

cmungall:
The final three review notes are closed in 0f94bf37d5: first_release now refuses when the configured registry already exists, unused previous_visibility fields are gone, and final-promotion recovery cannot mask the original verification error with a missing verification block. The recovery guide also states how an interrupted public promotion resumes.

Validation remains 40 focused/workflow tests passing with all changed-file hooks green.

cmungall:
The approved review’s last operator suggestions are addressed in 3a037534ff. The publish job now has an explicit 300-minute bound, the docs distinguish PUBLIC in-place availability from the default PRIVATE behavior, and an empty previous_manifest input is rejected even when first_release is selected.

Validation: the same 40 focused/workflow tests pass and both changed files pass all applicable hooks. I am leaving the implementation stable now while final review and CI run.

cmungall:
The red CI result was caused by an Appendiceal_Neoplasm snippet-grading baseline change that had landed on main after this branch was based; this PR does not modify that file. I rebased cleanly onto current main at 7af2b8686e. The post-rebase diff remains the same nine NDEx files.

Post-rebase validation: 26/26 snippet-grading tests pass against origin/main, 40/40 focused/workflow tests pass, and all changed-file hooks pass.

Issue #10556, opened 2026-09-02
CLAUDE.md documents `supports: PARTIAL` / `WRONG_STATEMENT`, but the schema only accepts SUPPORT / REFUTE / NO_EVIDENCE


cmungall:
Withdrawn — this was filed on a stale copy of `CLAUDE.md`. Current `main` already documents the retired values correctly (the Evidence Items section now reads `supports: SUPPORT  # or REFUTE, NO_EVIDENCE` and carries the full "`PARTIAL` and `WRONG_STATEMENT` were removed (issue #7439)" guidance, plus the `check-enum-values` gate from #10061). The two curation sessions that hit this were reading an outdated snapshot, not the current file. Nothing to fix.

cmungall:
@ai4c-agent please fix it's trivial

PR #6979, opened 2026-07-25
fix(dragon-ai): run as ai4c-agent App; drop assignment dispatch

cmungall:
Migrates the dragon-ai mention responder off the retired machine account + `PAT_FOR_PR` onto the **ai4c-agent GitHub App**.

## Changes
- **Execution identity → ai4c-agent App token** (SHA-pinned `create-github-app-token`): checkout (`persist-credentials: false` + `gh` credential helper), git identity (`ai4c-agent[bot]`), and the agent `GH_TOKEN` in both the prompt-build and `claude` run steps.
- **check-mention → read-only `GITHUB_TOKEN`** (no API calls; inspects the event payload + `ai-controllers.json`).
- **Dropped the `assigned` trigger + assignment logic.** GitHub Apps cannot be assignees, and the programmatic assigner (`stale-pr-reassign`) is retired (#6958). The mention keyword is unchanged.
- **Injection fix** in `Set up environment`: request text now via env + `printf` instead of `echo "${{ ... }}"` (a pre-existing shell-break / command-injection bug).

## Milestone
With this + #6958 merged, `PAT_FOR_PR` is fully retired from the repo.

_(This PR intentionally avoids the literal mention keyword in its body to prevent self-triggering the workflow.)_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10560, opened 2026-09-02
ci(dragon-ai): summon the agent as @ai4c-agent, keeping @dragon-ai-agent as a legacy alias

cmungall:
Typing `@ai4c-agent please ...` did nothing. The workflow that dispatches the mention-driven agent only ever matched the old `@dragon-ai-agent` handle, so a request using the new one ran the check job, decided there was no mention, and quietly stopped. Both handles work after this change.

This is visible in the run from issue #10556 earlier today:

```
User: cmungall, Allowed: true, Has mention: false
check-mention:      success
respond-to-mention: skipped
```

## Why it was missed

PR #6979 moved the agent onto the ai4c-agent GitHub App. That changed **who the agent runs as**, not **what you type to summon it**, and the trigger text was left behind on the retired account name. Neither handle is a real account you can notify: the App signs as `ai4c-agent[bot]`, and GitHub Apps cannot be @-mentioned, so both are plain text keywords.

The same stale handle had also been copied into the comment trust gate, which decides whether a comment looks like an attempt to drive an agent. It was not recognising the new handle either.

## What changed

The handles and the matching logic now live in one file, `.github/scripts/agent-mention.js`, which both the workflow and the trust gate read. A handle spelled out anywhere else is what caused this bug, so a test fails if a copy reappears.

This follows what go-ontology already does in its `ai-agent.yml`. Behaviour picked up from there:

- Either handle is accepted, and the reply nudges once toward `@ai4c-agent` when the old one was used.
- `@ai4c-agent[bot] please ...` is accepted, since people paste back the name they see the bot signing as.
- A multi-line request is captured whole. The old pattern stopped at the first newline, which silently truncated anything written as a list.
- Comments authored by a bot are ignored. Unlike the default token, App-authored events do retrigger workflows.
- The agent introduces itself by the configured handle instead of a hardcoded one.

Two dismech-specific protections are deliberately kept: the mention is ignored inside code spans and fenced blocks, so documenting the keyword (as this PR body does) does not fire the agent, and the controller allowlist is still read from the default branch rather than the PR ref.

The workflow file keeps its `dragon-ai.yml` name so no other references break. Only its display name is updated.

## Validation

```
just test-search                          68 tests, 68 pass
uv run pytest tests/test_dragon_ai_workflow.py tests/test_pr_shepherd_workflow.py \
  tests/test_auto_merge_workflow_tokens.py tests/test_agent_config.py tests/test_apply_cron_profile.py
                                          30 passed
```

Each test was written first and watched fail for the right reason. I also replayed the workflow's check script against synthetic event payloads: both handles and a multi-line body dispatch, while a code-span mention, a non-controller author, and a bot author do not.

The real check is whether `@ai4c-agent please ...` works on a live comment, which can only be confirmed once this is on `main`.

### DIFF
--- a/tests/test_dragon_ai_workflow.py
+++ b/tests/test_dragon_ai_workflow.py
@@ -193,8 +193,12 @@ def _handles() -> tuple[str, list[str]]:
     return canonical, legacy
 
 
-def _retired_handle_uses(legacy: list[str]) -> list[str]:
-    """Every place a retired handle is used to summon the agent."""
+def _retired_handle_uses(legacy: list[str], base: Path = ROOT) -> list[str]:
+    """Every place under *base* a retired handle summons the agent.
+
+    `base` exists so the meta-test below can scan a throwaway tree instead of
+    planting a probe in the working copy.
+    """
     searched = [
         path
         for directory, pattern in (
@@ -202,12 +206,12 @@ def _retired_handle_uses(legacy: list[str]) -> list[str]:
             ("actions", "*.y*ml"),
             ("scripts", "*.js"),
         )
-        for path in (ROOT / ".github" / directory).rglob(pattern)
+        for path in (base / ".github" / directory).rglob(pattern)
     ]
 
     offenders = []
     for path in searched:
-        rel = path.relative_to(ROOT).as_posix()
+        rel = path.relative_to(base).as_posix()
         if rel in LEGACY_HANDLE_ALLOWED:
             continue
         for number, line in enumerate(
@@ -244,15 +248,25 @@ def test_no_workflow_emits_a_retired_handle():
 
 
 def test_the_retired_handle_guard_actually_bites(tmp_path):
-    """The guard is only worth having if it fails on the thing it looks for."""
-    planted = ROOT / ".github" / "scripts" / "_guard_probe.js"
-    planted.write_text('// "@dragon-ai-agent please do X"\n', encoding="utf-8")
-    try:
-        offenders = _retired_handle_uses(_handles()[1])
-    finally:
-        planted.unlink()
+    """The guard is only worth having if it fails on the thing it looks for.
 
-    assert any("_guard_probe.js" in item for item in offenders)
+    The probe goes in a throwaway tree, not the working copy: a probe that
+    outlived its cleanup would fail the real guard repo-wide, pointing at a file
+    nobody wrote. It is planted as .js so the scan's script root is exercised
+    too -- a .yml probe would pass whether or not that root is reached.
+
+    The handle comes from the module rather than being typed here, so this file
+    keeps holding no handle literal of its own, and dropping a name from
+    LEGACY_AGENT_MENTIONS cannot leave the probe silently naming nothing.
+    """
+    legacy = _handles()[1]
+    probe = tmp_path / ".github" / "scripts" / "_guard_probe.js"
+    probe.parent.mkdir(parents=True)
+    probe.write_text(f'// "@{legacy[0]} please do X"\n', encoding="utf-8")
+
+    assert _retired_handle_uses(legacy, base=tmp_path) == [
+        ".github/scripts/_guard_probe.js:1"
+    ]
 
 
 def test_check_script_warns_when_an_authorized_mention_did_not_parse(tmp_path):


cmungall:
The `check-mention` check failed on the first push, and the failure was worth keeping rather than papering over. Fixed in 6c76974.

The job loads the shared parser from a checkout of the **default branch**, not the PR branch. On this PR that branch does not have `agent-mention.js` yet, so the load failed:

```
Error: Cannot find module ".../.github/scripts/agent-mention.js"
```

Reading it from the default branch is deliberate and stays that way. The alternative is a real security hole, not a style preference: if the parser came from the PR branch, someone could open a PR that rewrites the parser so that **any** comment by an authorized reviewer is reported as a valid request with attacker-chosen text. The authorization check would pass, because the reviewer really is authorized. The allowlist is read from the default branch for exactly this reason, and the parser sits next to it.

So the job now checks whether the parser is there and declines the event with a warning if it is not, instead of crashing. Without that, adding or renaming this module would turn every comment in the repo into a red check for as long as the change sat in review.

Also widened the path filters in `main.yaml`. The new guard tests currently run only because they are Python files under `tests/`, which means a future edit to just the workflow or just the script would skip them.

Verification:

```
uv run pytest tests/test_dragon_ai_workflow.py tests/test_pr_shepherd_workflow.py \
  tests/test_auto_merge_workflow_tokens.py tests/test_agent_config.py tests/test_apply_cron_profile.py
                                          32 passed
just test-search                          68 tests, 68 pass
```

Two of those tests run the workflow's own check script under Node: one confirms `@ai4c-agent please fix it` dispatches, the other reproduces the exact crash above and confirms it now declines cleanly.

cmungall:
Picked up items 1 through 5 in 8ddea2f. Item 6 I think should stay as it is; reasoning below.

**Item 1, the emitted handle.** Fixed, and the guard is now repo-wide as suggested. Rather than banning handle literals outright, it bans *retired* ones: a prompt that tells an agent how to summon another agent has to spell the name out, and there is no way to inject it from the module into a YAML prompt string. So spelling out the current handle is allowed anywhere, and spelling out a retired one fails. The test reads both from `agent-mention.js`, so it holds no handle literal of its own. I checked it actually bites by putting the old handle back and watching it fail.

**Item 3, dropped code blocks.** Fixed rather than documented. Masking now replaces each character one for one and keeps newlines, so the masked copy is the same length as the original and every offset still lines up. The mention is detected in the masked text, then the request is read from the original. `@ai4c-agent please apply this patch:` followed by a diff now keeps the diff. A mention that is itself inside a code block is still ignored. Two tests cover it, and CONTRIBUTING now says pasting a patch works.

**Item 4, the fence.** Worth doing precisely because item 3 makes it reachable: fences now survive into the prompt, so a fixed wrapper could be closed from inside the request. The request is delimited by a marker generated per run instead, which the request cannot contain because it did not exist when the request was written. I picked this over counting backtick runs because it removes the class rather than raising the bar.

**Items 2 and 5.** Both applied. The scanner prompts now name the current phrase and its alias, and discussion-scanner introduces itself by the current handle.

**Item 6, the missing left boundary.** Leaving this one. Tightening the parser to require `(?:^|\s)` would also stop matching `(@ai4c-agent please do X)`, since an opening parenthesis is not whitespace. The two regexes fail in opposite directions on purpose: the trust gate is over-inclusive because hiding a comment is cheap and reversible, whereas a parser that is too strict silently ignores a curator, which is exactly the bug this PR exists to fix and exactly how it went unnoticed for weeks. Only allowlisted controllers can reach the parser at all, so over-inclusiveness costs nothing here. Happy to change it if you read the trade-off the other way.

Verification:

```
uv run pytest tests/test_dragon_ai_workflow.py tests/test_pr_shepherd_workflow.py \
  tests/test_auto_merge_workflow_tokens.py tests/test_agent_config.py tests/test_apply_cron_profile.py
                                          34 passed
just test-search                          70 tests, 70 pass
```

On the `search_ranking.test.mjs` failure you hit: that is a missing `node_modules`, not the code. `npm ci` fixes it and the suite goes green.

cmungall:
All three picked up in a1aea7d.

**Item 1, the filter gap.** Fixed with globs rather than one more named file, since the guard scans every workflow and composite action, and this is the third time in this PR that a check has not run on the thing it checks. I first removed the individually named entries as redundant, then put them back: two existing guard tests assert those exact lines, and rewriting other features' tests is not this PR's business. Redundant against the globs, kept deliberately, with a comment saying why.

**Item 2, the overstated docs.** Added tilde fences to the mask, so both fence styles behave as documented. Indented lines I left as prose on purpose, and said so in CONTRIBUTING and in a test. A continuation line inside a list item is routinely indented, so treating indentation as code would drop real requests. That is the wrong direction to fail in, and it is the direction this whole PR is about.

**Item 3, the swallowed summon.** Rather than parse backtick runs properly, I made the failure visible. When someone authorized names the agent and no request comes out, the check job now annotates the run with what to do about it. That covers your case, a mention with no `please`, and any future variant, without a markdown parser. CONTRIBUTING tells curators where to look if they ask and nothing happens.

That last one is the part I would have wanted on day one. The original bug was not that the parser was wrong, it was that being ignored and being unheard looked identical from the outside.

Verification:

```
uv run pytest tests/test_dragon_ai_workflow.py tests/test_pr_shepherd_workflow.py \
  tests/test_auto_merge_workflow_tokens.py tests/test_agent_config.py tests/test_apply_cron_profile.py
                                          36 passed
just test-search                          73 tests, 73 pass
```

Nine of those run the workflow's own check script under Node, now including the two new diagnostic cases.

cmungall:
Picked up observations 1 and 3 in 71b270a. Observation 2 is filed as issue #10585.

**Observation 3, the guard scanning YAML only.** Widened to scripts. You are right that nothing under `.github/scripts/` emits a summon today, but this PR has now closed the same "the check does not cover the thing it checks" gap three times, and each time it was a surface nobody expected to matter. Closing the fourth one while it is still hypothetical is cheaper than the previous three were. I also added a test that plants a retired handle in a script and asserts the guard reports it, so the guard cannot quietly stop working the way the trigger regex did.

**Observation 1, the self-warning.** Left as is, and I wrote your analysis into the workflow so the first person to ask has the answer next to the code. Your reasoning is what convinced me: documenting the phrase in a code span and having a stray backtick swallow a real request produce the identical state, so there is no predicate to separate them. Tightening it would only drop the `cc @ai4c-agent for visibility` case, which is worth flagging anyway.

**Observation 2, the retired account grant.** Filed as issue #10585 rather than fixed here. The recipe cannot be repointed, because the App is installed rather than added as a collaborator, so the real question is whether to delete it and whether the account still holds write today. That second part is a live permission change, which is a maintainer decision and not something I should make on my own. I confirmed your read on the other two hits: `pr-shepherd.yml` lists both logins, and `pr_shepherd_policy.py` is correct as written.

Verification:

```
uv run pytest tests/test_dragon_ai_workflow.py tests/test_pr_shepherd_workflow.py \
  tests/test_auto_merge_workflow_tokens.py tests/test_agent_config.py tests/test_apply_cron_profile.py
                                          37 passed
just test-search                          73 tests, 73 pass
```

cmungall:
Both suggestions taken in f7b6735, and I am treating the review as settled there.

**Planting in the working copy was a real defect in that test.** `_retired_handle_uses` now takes the tree to scan, so the probe goes in `tmp_path`, which also makes the fixture parameter honest instead of accepted-and-ignored. Your point about the failure mode is the convincing part: a surviving probe would have failed the actual guard repo-wide against a file nobody wrote, which is a bad way for a test about legibility to fail.

**The hardcoded handle is fixed the same way the rest of the file already worked.** It now derives the name from the module, so this file still holds no handle literal of its own and dropping a name from the legacy list cannot leave the probe naming nothing while the assertion fails for an unrelated reason.

I also re-confirmed your check that the scripts scan root is load-bearing: removing that tuple fails the meta-test, so the coverage claim is still tested rather than assumed.

On suggestion 3, agreed on all three points and no action here. Other extensions under the scripts directory are outside the scan and nothing there emits a summon today; the retired account grant is issue #10585.

Verification:

```
uv run pytest tests/test_dragon_ai_workflow.py tests/test_pr_shepherd_workflow.py \
  tests/test_auto_merge_workflow_tokens.py tests/test_agent_config.py tests/test_apply_cron_profile.py
                                          37 passed
just test-search                          73 tests, 73 pass
```

Unless something blocking turns up, I would rather stop polishing here and let the required build decide. Four rounds have each found something real, but the last two have been about my tests rather than the behaviour, and the fix curators are waiting on is in the first commit.

Issue #5074, opened 2026-06-30
main is red: test_disorder_to_cx2_uses_top_level_references_for_network_header fails, blocking the whole merge queue


dragon-ai-agent:
## `main` is red: `test (3.13)` fails on every PR

`tests/test_cx2_export.py::test_disorder_to_cx2_uses_top_level_references_for_network_header`
**fails on current `main`** (verified at 47d6c4789). Because `test (3.13)` is a
required check, this blocks merging of **every** open PR — including approved,
otherwise-mergeable bot PRs (e.g. #4984, #4991, #4697 and the auto page-regen PRs).

### Symptom
```
assert 'DOI:10.1007/s00415-024-12724-3' in reference_html
E   assert 'DOI:10.1007/s00415-024-12724-3' in '<ul><li>...PMID:29290580...PMID:20301526...</li></ul>'
```

### Root cause
`_collect_reference_entries` in `src/dismech/export/cx2_export.py` keeps only the
first **`limit=3`** top-level references for the NDEx network header. The test asserts
that `DOI:10.1007/s00415-024-12724-3` ("Cytoskeleton dysfunction of motor neuron in
spinal muscular atrophy") appears in that header.

That DOI used to sit within the top-3 of `kb/disorders/Spinal_Muscular_Atrophy.yaml`'s
`references:` list, but **cumulative curation edits prepended new references**, pushing it
out of the window:

| commit | change | DOI index |
|--------|--------|-----------|
| before #4305 | — | 1 (in top-3 ✅) |
| #4305 (`4dac26639`) | inserted `PMID:42260294` (Flunarizine microRNA) | 2 |
| #4942 (`7a070eeb7`) | inserted `PMID:29290580` (2018 care guideline) | **3 (out of top-3 ❌)** |

Current loaded order: `0 PMID:29290580`, `1 PMID:42260294`, `2 PMID:20301526`,
`3 DOI:10.1007/s00415-024-12724-3`.

The test is **fragile**: it couples a hard-coded reference to both the SMA file's
reference ordering and the arbitrary `limit=3`, so any future reference addition can
re-break it.

### Recommended fix (test/src scope — out of curation-PR scope)
Pick one:
1. **Make the test robust** — assert against a reference that is intentionally first
   (e.g. `PMID:29290580`), or assert the header is non-empty / well-formed rather than
   pinning a specific DOI that curation can reorder.
2. **Decouple from ordering** — have the test build a tiny synthetic disorder with a
[...]

PR #10563, opened 2026-09-02
Verify merge integrity on every push to main

kevinschaper:
The safety half of the #10168 rollout, required before a merge queue with build concurrency above 1 is enabled on `main`: a post-push job that verifies each squash merge actually contains what its PR changed, plus the documented break-glass procedure for pausing a required queue.

## Why

Passing tests are not evidence of correct history. In the April 2026 GitHub merge-queue incident (#2034), squash merges of multi-PR merge groups silently reverted other PRs' changes — and CI stayed green, because the reverted code still passed its own tests. GitHub learned of the bug from customer reports; it had no monitoring that could see a content-corrupting merge. The queue configuration proposed in #10168's discussion uses speculative merge groups containing up to 5 PRs — the structural shape of that incident — so local detection needs to exist before the queue does. At ~50 merges/day it is also cheap standing insurance against non-queue merge anomalies.

## What it does

`verify-merge-integrity.yaml` runs on every push to `main`. For each first-parent commit in the push range, `scripts/verify_merge_integrity.py` checks the identity that holds for every clean squash merge:

```
git merge-tree --write-tree <parent-of-C> <pr-head>  ==  C^{tree}
```

A mismatch means the pushed commit does not contain what merging that PR into the previous tip produces — the incident signature. The run goes red and opens (or comments on) a `merge-integrity`-labelled issue, so a bad merge at 3am doesn't rely on someone noticing a red check among fifty daily pushes.

Deliberately skipped rather than judged: merge commits (two parents), rebase merges and direct pushes (no usable PR mapping), and commits whose PR's recorded `merge_commit_sha` doesn't match — a wrong guess in either direction is worse than an honest "could not verify". A conflicted merge-tree reconstruction reports as unverifiable, not as a mismatch. Each skip category is visible in the run summary, so coverage is auditable rather than assumed.

Tests exercise the tree identity against a real scratch git repository, not a mock — including the corrupted-squash case where another PR's change silently vanishes, which must report `MISMATCH`. Verification runs with a full-history blobless checkout (`fetch-depth: 0`, `filter: blob:none`); no concurrency group on purpose, since each run verifies exactly its own disjoint push range and a cancelled run would be an unverified push.

## Break-glass documentation

`docs/explanation/automation-and-agents.md` gains the queue pause/resume procedure — one ruleset API call each way, rehearsed live during the #10168 dry run — and when to use it: a red `main` no queued PR can fix (the #5074 shape), queue misbehavior, or a `merge-integrity` alarm. This was an explicit gap named in the #10168 discussion: a required queue removes the manual-merge escape hatch, so the escape hatch has to be written down before the queue exists.

The new workflow and script are added to the `python` paths filter so edits to them run the wiring tests.

Refs #10168, #2034.

### DIFF
--- a/.github/workflows/verify-merge-integrity.yaml
+++ b/.github/workflows/verify-merge-integrity.yaml
@@ -21,6 +21,7 @@ on:  # yamllint disable-line rule:truthy
 
 permissions:
   contents: read
+  pull-requests: read
   issues: write
 
 jobs:
--- a/docs/explanation/automation-and-agents.md
+++ b/docs/explanation/automation-and-agents.md
@@ -379,9 +379,10 @@ managed as a repository ruleset, and pausing it is one admin API call each
 way:
 
 ```bash
-# find the ruleset id
-gh api repos/monarch-initiative/dismech/rulesets \
-  --jq '.[] | select(.name == "merge-queue-main") | .id'
+# find the ruleset carrying the merge_queue rule (don't rely on its name)
+gh api repos/monarch-initiative/dismech/rulesets --jq '.[] | "\(.id) \(.name)"'
+gh api repos/monarch-initiative/dismech/rulesets/<id> \
+  --jq '[.rules[].type]'   # confirm it includes "merge_queue"
 # pause (PRs merge normally again; queue state is abandoned)
 gh api -X PUT repos/monarch-initiative/dismech/rulesets/<id> -f enforcement=disabled
 # resume
--- a/scripts/verify_merge_integrity.py
+++ b/scripts/verify_merge_integrity.py
@@ -30,6 +30,9 @@ What is deliberately NOT verified:
 - **Rebase merges and direct pushes**: a rebase merge maps one PR onto
   several commits none of which is the PR's ``merge_commit_sha`` recorded
   here, and a direct push has no PR at all. Both are reported as skipped.
+- **History rewrites**: the check walks ``before..after``, so a force-push
+  that *drops* commits contributes nothing to the range and passes. ``main``
+  forbids force pushes; if that ever changes, this needs an ancestry check.
 - **PRs whose recorded ``merge_commit_sha`` does not match the commit**:
   skipped with a note rather than guessed at -- a wrong guess either way
   (false alarm, or false clearance) is worse than an honest "could not
@@ -235,7 +238,6 @@ def main() -> int:
     parser.add_argument("--after", required=True,
                         help="main tip after the push (github.event.after)")
     args = parser.parse_args()
-    emit_output("mismatch", "false")
 
     if set(args.before) == {"0"}:
         # A branch-creation push reports the null SHA; nothing to compare.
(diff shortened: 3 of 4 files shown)


kevinschaper:
Both blocking findings are fixed, and six of the eight suggestions adopted; one suggestion led to a correction of the review's own premise, noted below.

**False-incident gating:** the script now writes an explicit `mismatch=true/false` step output, and the issue step fires on `steps.verify.outputs.mismatch == 'true'` — a crash, a rate-limited lookup, or a usage error still goes red but can no longer post "treat main as suspect". Per-commit verification is additionally wrapped so one commit's git failure becomes an `UNVERIFIABLE` result instead of aborting the run.

**Fail closed:** lookup failures now get their own status (`SKIPPED_LOOKUP_ERROR`, covering `CalledProcessError` and `OSError`, so a missing `gh` binary lands here too). When a run's squash-shaped commits produce only lookup/reconstruction failures and zero `VERIFIED`, `main()` exits 2 with an error annotation — red, no incident issue. Partial failures emit a warning annotation. Policy skips (merge commits, direct pushes, mapping mismatches) still pass, since those are out-of-scope by design rather than blindness.

**On suggestion 1 (conflict vs error return codes):** adopted, but the proposed mechanism doesn't hold — on git 2.43, `merge-tree --write-tree` exits **1 for a failed lookup too** (`merge-tree: <sha> - not something we can merge`), not >1. Verified directly, and there's a test pinning it. The discriminator is whether stdout's first line is a tree OID: a conflicted `--write-tree` run still writes one, a failed invocation writes none. The `UNVERIFIABLE` detail now carries stderr for the error case.

**Also adopted:** the `UNVERIFIABLE` conflict-path test (both branches editing the same line, exactly as sketched); the head-at-merge-time and rename-detection caveats, in the module docstring's not-verified list and in the docs section (with "compare trees by hand before declaring an incident, especially in the first weeks" as the operational guidance in place of a warn-only mode); the null-SHA short circuit; the `refs/pull/<n>/head` fetch fallback with the fetch-by-SHA caveat in its docstring; and the root-commit wording.

**Not changed:** the yamllint line-length warnings — the remaining long lines are the issue-body prose and the gating expression, and the file matches the repo's existing workflows on this. The `merge-integrity` label color/description stay as-is pending anyone caring.

kevinschaper:
The blocking permissions item is fixed — `pull-requests: read` added — and all five suggestions are taken:

1. The `mismatch=false` pre-write is gone; the output is written once, on an actual mismatch.
2. `main()` now has its own tests, driving it with a stubbed `verify_commit`, a scratch `GITHUB_OUTPUT`, and a real repo range: the null-SHA no-op (asserting `verify_commit` is never reached), the mismatch path (exit 1 and `mismatch=true` emitted), and the fail-closed path (exit 2, no output gate). 12 tests in the module now.
3. Zero-`VERIFIED` runs whose PR-numbered commits were all mapping skips now emit a warning annotation naming the possibly-broken squash-mapping assumption; still green, per the mixed-push reasoning you laid out.
4. The force-push range limitation is documented in the docstring's not-verified list, with the note that `main` forbids force pushes and an ancestry check becomes necessary if that changes.
5. The break-glass lookup no longer assumes a ruleset name: it lists rulesets and confirms the target carries a `merge_queue` rule.

(The fix commit is `dd93323`, replayed on top of the branch-update merge that landed while this round was in flight.)

PR #10565, opened 2026-09-02
Wire phenotypes into the causal graph across 38 cardiac entries, and review why they were not

cmungall:
Across the non-Mendelian cardiac entries, the clinical manifestations were curated but never connected to the mechanisms that produce them. 68 of 414 phenotypes were reached by a causal edge; 21 of 40 entries had none at all. This PR connects them, binds six inheritance blocks that were free text, and adds a written review of what the gap was and what caused it.

**No phenotypes, pathophysiology nodes, references or snippets were invented anywhere in this PR**, and nothing was renamed, deleted or reordered. 104 evidence blocks were added, every one of them a copy attached to a new edge. They carry 51 distinct references, and **not one file cites a reference it was not already citing** — verified by diffing the reference set of each file against its base version. Every new edge carries a `description`; an `evidence` block appears on a new edge only where an item already present in that same file supports that specific edge.

## Result

| Measure | Before | After |
|---|---|---|
| Phenotypes reached by a causal edge | 68/414 (16.4%) | 353/414 (85.3%) |
| Phenotypes touching the graph at all | — | 408/414 (98.6%) |
| Entries with no wired phenotype | 21 | 0 |
| Edges across the 40 entries | 568 | 1,126 |
| Grandfathered dangling targets, KB-wide | 126 | 123 |

For comparison, the knowledge base as a whole sits at 48% on the same measure (`PhenotypeConnectivityPlugin`, `src/dismech/qc_plugins.py`).

The mechanism layer was never the problem: only 6% of pathophysiology nodes were unconnected, against 1% in the Mendelian cardiac entries. The deficit was specifically the mechanism-to-phenotype edges. Largest recoveries: Kawasaki disease 1/40 to 40/40, congenital heart disease 6/20 to 20/20, coronary artery congenital malformation 0/13 to 13/13, histiocytoid cardiomyopathy 0/15 to 14/15, Eisenmenger syndrome 1/17 to 13/17.

Three dangling bare-name targets were repaired rather than baselined. In `Idiopathic_Spontaneous_Coronary_Artery_Dissection` both were the same defect — one edge bundling four phenotypes — now split into separate edges naming the real phenotypes, with the flow-limitation and regional-ischemia steps carried as `intermediate_mechanisms` instead of as phantom nodes.

## Inheritance was not machine-readable

Six cardiac entries stated their mode of inheritance in a free-text name and description with no `inheritance_term` at all, which is why "which cardiac entries are non-Mendelian" had to be answered by reading prose. Now bound: `CADASIL_Type_1` and `Cardiofaciocutaneous_Syndrome` to `HP:0000006`, `Heart_Defect_Tongue_Hamartoma_Polysyndactyly_Syndrome` to `HP:0000007`, `Cardiac_Sarcoidosis` and `Double_Outlet_Right_Ventricle` to `HP:0001426`, `Dilated_Cardiomyopathy_1Z` to `HP:0025352`.

Two things worth knowing here. **HPO has no "Multifactorial inheritance" term** — `HP:0001426`, which carried that label historically, is now "Non-Mendelian inheritance", and the 23-member `InheritanceTerm` enum has no multifactorial concept at all. A curator reaching for that word finds nothing, which is a plausible reason both multifactorial blocks were left as free text. And `Dilated_Cardiomyopathy_1Z` is bound to `HP:0025352` "Typically de novo" rather than `HP:0001442` "Typified by **somatic** mosaicism": the block describes **germline** mosaicism, which is what explains sibling recurrence, and no enum term carries that.

A further 43 cardiac-adjacent entries have no `inheritance:` block at all, including several unambiguously Mendelian ones (hypertrophic cardiomyopathy, dilated cardiomyopathy, Brugada syndrome). Those are left alone here.

## Three findings that are not about wiring

**`Pathophysiology.consequences` is a dead slot.** It validates, but `dismech.graph`, the renderer, all thirteen exporters and the claim extractor all ignore it, so anything written there is invisible. Unlike a dangling target it produces no phantom node, so no gate catches it. 27 nodes across the knowledge base use it — 22 of them in `Alzheimer_Disease`, which states most of its causal architecture that way. Removed here from `Pulmonary_hypertension`, where it duplicated an existing edge. No open issue covers this.

**Genetic susceptibility loci cannot reach the pathograph.** `graph.py` links a genetic item to a mechanism only when a pathophysiology node carries a matching gene descriptor. Entries typing a locus as `SUSCEPTIBILITY` or `UNKNOWN` with no such descriptor cannot be connected from any target slot. This accounts for most of the remaining isolated nodes — 27 in the congenital group alone, where not one pathophysiology node carries a `genes:` block.

**`Pulmonary_hypertension` does not conform to `pulmonary_vascular_remodeling`,** though nine entries do, three of which are its own downstream consequences. 24 of the 40 entries declare no module conformance at all.

## Deliberate non-fixes

Recorded rather than forced, so reviewers do not re-litigate them:

- **Node-name collisions** where one name is both a pathophysiology node and a phenotype, which the flat namespace merges (issue #9896). Found in `Pulmonary_hypertension`, `Tetralogy_of_Fallot`, `Pericarditis`, `Rheumatic_Heart_Disease`, `Endomyocardial_Fibrosis` and `Myocardial_Infarction`. Renaming curated nodes to satisfy the graph model is the wrong direction of fix.
- **Curator decisions recorded in `notes`.** `Ventricular_Septal_Defect` leaves two phenotypes unwired with a stated reason. `Eisenmenger_Syndrome` prostanoid therapy has an obvious mechanism target contradicted by the entry's own `REFUTE` evidence that pulmonary vascular resistance did not improve. `Myocarditis` records that no `influences_mechanisms` link is asserted for exercise.
- **Monitoring, screening and counselling treatments** left unlinked, matching the convention that 0 of 222 such actions carry `target_mechanisms` anywhere in the knowledge base.
- **`Pulmonary_hypertension`'s one remaining dangling target.** `Right Heart Failure` names no node, and retargeting it to `Edema` or `Fatigue` would downgrade a claim whose evidence is explicitly about right-sided failure. The correct fix is a new pathophysiology node.
- **A two-cycle in `Coronary_Vasospasm`**, where an edge's description states the reverse direction and duplicates its own reciprocal. Either the edge is redundant or the description is wrong.

## Two things worth knowing when reading the numbers

**The QC metric understates correct curation.** It counts only `causes`, `leads_to`, `triggers` and `exacerbates`. An ECG sign, echo measurement or lab value should be modelled as a `reports_on` readout, which is not counted — so 55 phenotypes here score as unconnected while being correctly wired. `Posterior_Myocardial_Infarction` reads 1/5 while fully connected, because four of its five phenotypes are ECG signs and a troponin level.

**Unlinked treatments were not a defect.** Restricted to therapeutic actions, 105 of 206 treatments in scope already carried `target_mechanisms` at baseline, against 48% knowledge-base-wide.

## Entries where wiring is not the fix

`Kawasaki_Disease` is now fully connected but structurally a two-hub star: 40 phenotypes hanging off 2 mechanism nodes, with no node for the unknown trigger, no inflammasome node despite anakinra being curated and NLRP3 listed among susceptibility genes, and none of the standard three-stage arteritis histopathology. It also carries two separate phenotypes bound to the same term, `HP:0002617`. `Coronary_Arterial_Fistulas` compresses steal physiology, volume load, ischemia, arrhythmia and heart failure into one node. `Patent_Ductus_Arteriosus` has no node for the diastolic-steal physiology its own evidence turns on. `Heritable_Pulmonary_Arterial_Hypertension`, Mendelian and out of scope here, was found in the pre-repair state and needs the same pass.

## Scope

43 `kb/disorders/` entries, one history record each, and the review report with its `mkdocs.yml` nav entry. The report and `mkdocs.yml` fall outside the curation-only path allowlist, which is what the `scope-override` label is for — `CLAUDE.md` asks for analysis of the knowledge base's content to live in `docs/reports/` and be surfaced in the nav, so a PR that both reviews and curates will always trip that guard.

## Validation

`just validate-disorders` over the changed entries: schema, ontology terms and references pass, with 2,186 of 2,350 snippets verified against the reference cache. `just validate-terms` passes with labels on the six inheritance bindings. The four ungated whole-knowledge-base gates and `check_snippet_grading` are clean, `just validate-history-all` passes over all 8,016 records, and the full `tests/test_data.py` suite passes at 47,286 tests.

https://claude.ai/code/session_018ovCmdvhv5coGCnEZodEYq


cmungall:
`test (3.13)` failed on the curation-scope guard, not on a test. The guard blocks a PR that touches `kb/` from also touching paths outside the curation allowlist, and this PR touches two:

```
docs/reports/non-mendelian-cardiac-pathograph-review-2026-09-02.md
mkdocs.yml
```

Those are the written review that this curation came out of, plus its nav entry. `CLAUDE.md` asks for analysis of the knowledge base's content to live in `docs/reports/` and to be surfaced in the mkdocs nav, so a PR that both reviews and curates will always trip this guard. I have applied the `scope-override` label, which is the remedy the guard's own error message names.

**The check needs a re-run to see the label, and I cannot trigger one** — the API returns 403 for re-running workflows with the token I have. The workflow is `on: pull_request` with no `types:`, so it fires on open, synchronize and reopen but not on `labeled`, which means adding the label did not re-trigger it by itself. Either a maintainer re-runs the failed job, or the next push to this branch picks it up.

Nothing else is red. I will keep watching and will report if anything changes.

For reference, everything the guard would have run passes locally on this head: `just validate-disorders` over all 38 changed entries (schema, ontology terms and references, with 2,186 of 2,350 snippets verified against the reference cache), the four ungated whole-knowledge-base gates, `check_snippet_grading`, and `just validate-history-all`.

---

cmungall:
Thanks — both actionable suggestions are fixed in `b5bfa4b76`.

**Suggestion 1, the retired `PARTIAL` prose.** Good catch, and it was a miss on my side rather than an oversight: I found this same defect earlier in the session in `MT-ATP6_MT-ATP8-Related_Infantile_Hypertrophic_Cardiomyopathy` and fixed four explanations there, but I grepped only that one file instead of the whole branch diff. These nine survived because of that.

You were right that `directness: INDIRECT` is what every one of them was reaching for, so that slot is now set on all nine and the sentences say the support is indirect and why. Eight in `Takotsubo_Cardiomyopathy`, one in `Endomyocardial_Fibrosis`. Only explanations this branch added are touched — the pre-existing backlog of the same pattern in those files is left alone.

One thing worth flagging for anyone else doing this rewrite in bulk: my first attempt phrased it "The support is indirect: …", which puts a colon-space inside an unquoted plain scalar and breaks the YAML parse. `linkml-validate` caught it, but `check-causal-targets` still passed on the broken file, so the offline gates alone would not have stopped it. The wording is now "The support is indirect, since …".

**Suggestion 2, the stale baseline.** `tests/causal_target_baseline.txt` goes 132 to 129 lines. The three removed rows are exactly the dangling targets this branch repaired — two in `Idiopathic_Spontaneous_Coronary_Artery_Dissection`, one in `Posterior_Myocardial_Infarction`. Removals only.

**Suggestion 3, the identity edges.** Left as they are, per your reasoning. `Kawasaki_Disease`'s `Vasculitis` and `Coronary_Vasospasm`'s `Coronary artery vasospasm` really are "is a manifestation of" rather than "causes", but `downstream` is the only slot available and inventing a predicate is not this PR's job. It belongs with issue #9896.

Also confirming your note on `check-snippet-grading`: the single finding is in `Follicular_Lymphoma.yaml` from #10418, on the base branch and untouched here.

Validation on the three affected entries: `just validate-disorders` passes with 345 of 345 snippets verified, plus `linkml-validate` and the four offline gates, the last three whole-knowledge-base. History records added and validated across all 8,018.

---

PR #10567, opened 2026-09-02
Add dbGaP and ImmPort dataset discovery via native APIs

cmungall:
## Summary

Implements dataset discovery for dbGaP and ImmPort studies using their native APIs rather than the NLM Dataset Catalog. These are the only two repositories in the catalog with coded disease indexing (MeSH descriptors for dbGaP, disease fields for ImmPort), and direct API access is strictly superior to catalog queries on precision, coverage, and available fields.

## Key Changes

- **New discovery script** (`scripts/discover_dbgap_immport.py`): Queries dbGaP FHIR `ResearchStudy` endpoint and ImmPort search API, keyed on MONDO→MeSH descriptor cross-references. Implements relevance tiering (TITLE_MATCH / VARIABLE_MATCH / SUBJECT_ONLY / CONFLICT) with data-dictionary inspection to distinguish outcome variables from incidental mentions.

- **Enhanced title matching** (`scripts/disease_title_match.py`): Added diacritic folding (Sjögren → Sjogren) and CamelCase compound boundary relaxation (AsthmaNet → matches Asthma) to recover on-target studies that were previously scored as SUBJECT_ONLY.

- **Accession verification** (`scripts/verify_dataset_accessions.py`): Added dbGaP FHIR resolver (replacing withdrawn E-utilities `gap` database) and ImmPort resolver. Both validate canonical accession format and study existence.

- **Test coverage** (`tests/test_dbgap_immport_discovery.py`): Regression tests for diacritic folding, relevance tiering, record construction, and accession verification (all offline, no API calls).

- **Documentation**: Added evaluation report (`docs/reports/nlm-dataset-catalog-evaluation-2026-08-07.md`) explaining why native APIs are superior to the catalog, and updated dataset-curation guide with dbGaP/ImmPort discovery workflow.

- **Schema and config**: Added `immport:` prefix to dismech schema; updated justfile and mkdocs navigation.

## Implementation Details

- **Relevance tiering** uses study titles and phenotype data dictionaries (fetched from dbGaP FTP) to distinguish outcome variables (case/control, affection status) from incidental mentions (medical history, comorbidity). This prevents false positives like GTEx (coded for asthma, not an asthma study).

- **Diacritic handling** uses Unicode normalization to fold accented characters before phrase matching, preserving the original entry spelling in output for curator visibility.

- **Data type inference** maps ImmPort `assay_method` to dismech enums (RNA sequencing → BULK_RNA_SEQ, etc.) but deliberately does not infer from dbGaP study design (Case-Control ≠ GWAS).

- **Rare disease handling**: Entries with only MeSH Supplementary Concept Records (C######) receive no coded query, as spot-checking confirms dbGaP holds no studies for them.

- **No dataset cache write.** An earlier revision of this branch populated `cache/dataset_accessions.json` with 30 study records. That file is frozen — nothing reads or writes it, and `geo:` accessions are now verified through `references_cache/GEO_*.md` instead — so the change was dropped and is no longer in the diff.

https://claude.ai/code/session_01K48gbRmWaM94DMH4NGefqn


caufieldjh:
This branch is mergeable again, and all five blocking findings are fixed. Two commits: one merging `main` and one applying the review. Every fix has a regression test, and each defect was reproduced before it was changed.

The merge also removes this PR's edits to `cache/dataset_accessions.json` entirely. That file was frozen after this PR was opened — nothing reads or writes it now, `CLAUDE.md` says never to stage it in any change, and `test_no_automation_touches_the_frozen_dataset_cache` enforces that. The branch's verifier still had `CACHE_PATH` pointing at it while `main` had already replaced that whole model with per-record `references_cache/` files and `REFERENCE_CACHED_PREFIXES`, so that was a real structural merge rather than a context conflict. Main's model wins and the dbGaP/ImmPort resolvers now sit on top of it.

## The five blocking findings

**Finding 1 — `strip("'s")`.** `str.strip` takes a character set, so it ate the head noun's own trailing `s`: `diabetes` → `diabete`, `stenosis` → `tenosi` at both ends. Seven of the fourteen `ADJECTIVAL_FORMS` keys were unreachable. Replaced with a real suffix strip, plus a table-driven test that every key round-trips and a test that a possessive head still reaches the table.

**Finding 2 — the CamelCase boundary.** Accepting any following uppercase letter admitted exactly what the docstring says is blocked, because repository titles are full of ALL-CAPS fragments:

```
'Adenoma'  in 'FAMILIAL ADENOMATOUS POLYPOSIS' -> matched
'Lymphoma' in 'LYMPHOMATOID PAPULOSIS COHORT'  -> matched
```

Now requires a true CamelCase seam, `(?<=[a-z])(?=[A-Z])`. `AsthmaNet` still matches; both cases above are rejected. This is shared code behind five discovery scripts and `TITLE_MATCH` is auto-approved, so it was the one worth getting exactly right.

**Finding 3 — outcome cues beating incidental cues within one variable.** An incidental cue now vetoes the outcome reading inside a single variable, so `("MHASTHMA", "Self-reported physician diagnosis of asthma (medical history)")` stays in `SUBJECT_ONLY` for manual triage. Across *different* variables the outcome reading still wins, which is what `test_outcome_beats_incidental_when_a_study_has_both` was checking and that behaviour is unchanged. The weak standalone `\bdiagnosis of\b` cue is dropped, which is what was promoting `("AST_DX", "Age at diagnosis of asthma")`.

**Finding 4 — the "3 dbGaP accessions" claim.** It is 71 unique accessions across 65 files. Corrected in the script docstring and in the evaluation report, with the command that measures it. Worth noting that `docs/dataset-curation.md` on `main` independently reports the same 71 in its "worth migrating next by volume" list.

**Finding 5 — single-page queries.** `_fhir_studies` now follows `Bundle.link[rel=next]` through every page, under a cap that is reported when it bites — the convention `MAX_DICT_TABLES` already set — and stops if a server hands back a self-referential next link. `query_immport` reports when `hits.total` exceeds what the page returned, rather than guessing an undocumented paging parameter.

## Optional items

All taken. The matcher docstring no longer contradicts the adjectival table shipped beside it. Inflected variants are filtered out of query strings by a new `query_phrases`, since neither repository's disease field contains "Asthmatic". Description truncation no longer cuts mid-word and then punctuates it as a finished sentence. `resolve_dbgap` reports a response carrying no `phs` identifier instead of falling through to `OK`. `resolve_immport` reports organism as a scalar like every other resolver, and its docstring now records that no KB file cites an `immport:` accession yet, so it has been exercised against the response shape and not against a corpus.

## What I did not do

**The "9 hits to 1" figure in `docs/dataset-curation.md` is withdrawn, not restated.** It was measured before the veto in finding 3 and before the weak cue was dropped, both of which promote strictly fewer studies, so it is stale by construction. Re-measuring needs a live dbGaP run. The paragraph now says that rather than quoting a number I cannot stand behind.

**The PR description is now out of date and I have not edited it.** The "Dataset cache update" bullet describes a change that is no longer in the diff, and the reviewer had separately flagged its numbers as wrong. Happy to rewrite that bullet if you would like — it is your description, so I left it alone.

## Validation

`test_dbgap_immport_discovery.py` and `test_dataset_discovery.py`: 72 passed. `test_no_automation_touches_the_frozen_dataset_cache`: passed. `ruff check .`: clean repo-wide. The net diff against `main` is 10 files, +2124/-9, with no `kb/` files touched.

---

caufieldjh:
The `test (3.13)` failure here was the lint step, not the tests. `uvx ruff check .` resolves the newest ruff rather than a pinned one, and 0.16.x reports nine findings — all nine inside this PR's own two new files, so they're this PR's to fix rather than base-branch drift. Pushed `020dea2f` with all nine fixed:

- **EXE001** — `scripts/discover_dbgap_immport.py` had a shebang but no exec bit (0644). Every sibling `discover_*.py` is 0755, so it's now 0755 too.
- **RUF100 ×2** — the two `# noqa: BLE001` directives can never fire, because `BLE001` is already in `pyproject.toml`'s global `ignore` list. Removed the dead directives and kept the explanatory prose as plain comments.
- **UP012 ×2** — the latin-1 decode test built its fixture from `"...".encode()` calls, which are UTF-8 no-ops. They're bytes literals now, which says the same thing more directly.
- **RUF059 ×4** — four unpacked tuple elements are never read by the surrounding assertions. Underscore-prefixed rather than replaced with `_`, so the arity of each resolver's return value stays readable.

Verified with ruff 0.16.8 (clean across the whole tree, which also clears the pre-existing `discover_ega.py:165` finding the review scoped out) and the PR's own 50 tests, all passing.

Also updated the PR body, which still advertised a `cache/dataset_accessions.json` write. That file is frozen and no longer in the diff; the body now records that it was dropped and why, so the history isn't just silently missing.

What I did not change: nothing else in the diff. This is the lint fix only.

One cost worth naming — this push drops the round-2 approval, since `main` has `dismiss_stale_reviews` on. That's the accepted price of getting to green, and the alternative (leaving the PR red) isn't one.

---

caufieldjh:
Thanks for the round-3 approval. One correction, on the description item.

**The stale bullet is already gone, and was before this review ran.** The body was edited at 13:13Z today; the round-3 review was submitted at 16:16Z. The current body carries:

> **No dataset cache write.** An earlier revision of this branch populated `cache/dataset_accessions.json` with 30 study records. That file is frozen — nothing reads or writes it, and `geo:` accessions are now verified through `references_cache/GEO_*.md` instead — so the change was dropped and is no longer in the diff.

There is no "Dataset cache update … Populated 30 dbGaP and ImmPort study records" line anywhere in it. I replaced the bullet rather than deleting it, on the same reasoning you gave — the squash body becomes permanent history, so it seemed better for that history to record that the write was dropped and why than to fall silent about it.

Not pushing anything for this. The PR is `MERGEABLE`, green on `020dea2f`, and 15 days old, so it is eligible for the deterministic sweep now; a push would only dismiss your approval and delay it.

**On the other two, left alone deliberately:**

- `immport:` missing from the generated `src/dismech/datamodel/` copies — agreed it's a note, not a request. Those are regenerated out of band and nothing gates the sync, so regenerating here would mean a push that costs the approval for a file this PR has no reason to own.
- The self-referential-`next` notice and the unpropagated `_truncate` fix in `discover_ega.py:165` stay out of scope, as in round 2. The `discover_ega.py` one is the same shared shape and worth its own small PR rather than widening this one.

For the record on the lint round that dismissed round 2: all nine findings were inside this PR's own two new files, and `uvx ruff check .` resolves the newest ruff rather than a pinned one — the same unpinned-default drift `pyproject.toml` already documents for E402. Verified clean under 0.16.8 across the whole tree.

---

PR #10575, opened 2026-09-02
curate: add Acute Ackee Fruit Intoxication (MONDO:0019140)

caufieldjh:
## Acute Ackee Fruit Intoxication (MONDO:0019140)

Jamaican vomiting sickness. Selected from a shared *Toxicologic Phenocopy Atlas* of 48 exposure/gene pairs as the one case that could actually be curated here: specific MONDO term, absent from the KB, and both Mendelian mimics already curated so the differentials resolve rather than dangle.

Hypoglycin A is a protoxin. The body converts it to methylenecyclopropylacetyl-CoA, a suicide substrate that covalently modifies the flavin of acyl-CoA dehydrogenases and takes activity below 2 percent. Beta oxidation stops. Ketogenesis and gluconeogenesis fail together, which is why the hypoglycaemia is profound and why it comes back after correction.

8 nodes, fully connected, all five phenotypes wired. 4 `conforms_to` against `metabolic_intoxication_decompensation`.

### The atlas claim is not curated as given

The atlas calls this a Tier 1 phenocopy of **MCAD deficiency**, with the shared lesion stated as MCPA-CoA inhibiting the **medium chain** acyl-CoA dehydrogenase.

The two primary enzyme papers say the opposite about that substrate.

- `PMID:1247597`: in hypoglycin dosed rat liver mitochondria, butyryl-CoA activity was "drastically reduced" while "Activity with octanoyl- and palmitoyl-CoA was unaffected."
- `PMID:7275979`: "Octanoyl-CoA affords strong protection against this inhibition."

Octanoyl-CoA is MCAD's substrate.

What the clinical and veterinary literature does support has a name. `PMID:23167695` calls hypoglycin A a cause of **acquired multiple acyl-CoA dehydrogenase deficiency**, with acylcarnitine and organic acid profiles "typical for MADD."

So the entry curates a multi enzyme block, names MADD as the closer phenocopy, and keeps MCAD deficiency as a differential with the octanoyl-CoA evidence attached to say why the resemblance is clinical and not enzyme level. The node `notes` state the disagreement, give the 1976 and 1981 dates and the species, and say plainly that those results are enough to decline naming the medium chain enzyme and not enough to settle it. A modern isoform resolved study in human enzyme is named as what would settle it.

The phenocopy is also more precise than the atlas allows. Genetic MADD is ETFDH, ETFA, ETFB, which is the electron acceptor. Hypoglycin destroys the dehydrogenases that feed it. One rung apart on the same electron path, and downstream nothing separates them.

### Differentials

| Entity | Why |
|---|---|
| `MONDO:0009282` MADD | The match. Literature calls this an acquired MADD outright. |
| `MONDO:0008721` MCAD deficiency | Clinical resemblance only, with the enzyme evidence against it attached. |
| Reye syndrome | The historical confusion, with the `PMID:940578` sentence that separated them. |

### A knowledge gap, not a mechanism

A falcon run offered a dual toxin picture: hypoglycin A to MCPA-CoA blocking the dehydrogenases at step one, and MCPG to MCPF-CoA blocking the enoyl-CoA hydratases at step two. The framework is real and `PMID:34765670` states it for the maple toxins in a foal.

For ackee it is not established. `PMID:30056267`:

> To date, MCPG exposure has not been evaluated in cases of JVS

So it is curated as a `KNOWLEDGE_GAP` discussion attached to the central node, with two proposed experiments, rather than put into the pathograph. The discussion `notes` record that the report presented it as settled.

### Deep research

Three falcon reports via Edison. Two were the same trajectory returned twice; the duplicate was deleted rather than committed. The third differed substantively.

Run one named MCAD and IVD among the inhibited enzymes. Run three said "multiple acyl-CoA dehydrogenases" and named neither. Two runs of one provider on one question disagreeing on exactly the disputed point is better support for withholding the enzyme name than either run alone.

`PMID:30056267` also closed two gaps this entry had recorded against itself: the unripe fruit claim, and the US import regulation behind the canned route in the Toledo case.

NEC preflight `SKIP` (no causal gene for a poisoning), cleared manually: top gene mentions were MADD, ACADM, IVD.

### Ontology

`FOODON:03301360` akee (raw) bound on the environmental `food_source` slot, following the pattern in `Gout.yaml` and `Celiac_Disease.yaml`. Raw rather than canned because the toxin is in unripe fruit. The canned route is documented and noted; the slot takes one term.

### Validation

`just validate-disorders` (the batched sweep CI runs) passed. 29/29 snippets verified over 9 references. `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets`, `check-environmental-evidence`, `check-entity-refs` all clean. Graph checked for dangling targets, nodes without inbound edges, and unwired phenotypes: none. Discussion `attaches_to` resolves. All `reference_title` values generated programmatically from the cache.

### DIFF
--- a/history/disorders/Acute_Ackee_Fruit_Intoxication/2026-09-02T161121Z-claude-code-9fb175.yaml
+++ b/history/disorders/Acute_Ackee_Fruit_Intoxication/2026-09-02T161121Z-claude-code-9fb175.yaml
@@ -70,7 +70,7 @@ events:
     Reye syndrome is curated as a third differential, with the sentence from
     PMID:940578 that historically separated them.
 
-    Evidence: 29 items over 9 references, every snippet pre-verified against the local
+    Evidence: 43 items over 9 references, every snippet pre-verified against the local
     cache before writing. Human anchors are PMID:28613753 (clinical syndrome),
     PMID:940578 (MCPA in the urine of two patients, dicarboxylic acids 70 to 1000
     times normal, and the Reye distinction), PMID:8185109 (canned-product case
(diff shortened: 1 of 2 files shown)


caufieldjh:
All four IMPORTANT items are in. Item 3 was not a coverage gap and I want to be plain about it.

### 3. The steatosis claim

You are right, and worse: this is the second time. On the isoniazid entry you caught prose asserting what the evidence did not carry, and I replied there that I had started reading `description` against `notes` before pushing. I made that check too narrow. It should have covered every free text field that renders as disease content, read against what its own block cites. `differential_diagnoses` description was outside the check I actually adopted, so the same defect recurred in a different slot.

The claim is removed. The resemblance is now stated at the level `PMID:940578` states it, and the entry says plainly that it does not curate the histology.

### 1. `biochemical:`

Added, and you were right about why it was the lead item: the entry's own description leads with the dicarboxylic aciduria and the MADD-like acylcarnitine profile, and both were prose. Four entries with `readouts` joining them to the mechanism nodes, following `Multiple_Acyl-CoA_Dehydrogenase_Deficiency.yaml`.

- Urinary dicarboxylic acids, with the named analytes and the 70 to 1000 times figure
- Urinary short-chain fatty acids at up to 300 times normal, which sits well with the enzyme evidence where the short-chain substrate was the sensitive one
- Acylcarnitine profile read as MADD
- MCPA-carnitine, which reports activation rather than the block

Two of the four carry `notes` recording that the acylcarnitine profile and the carnitine conjugate are demonstrated in horses while the human evidence is the urinary organic acids. That distinction is the reason those two blocks exist at all, so it seemed wrong to leave it implicit.

### 2. Coma

Added, with a `downstream` edge from the encephalopathy node. It was the third item in a sentence I had already quoted twice and evidently stopped reading two thirds of the way through.

### 4. `animal_models:`

Equine atypical myopathy added. `RECAPITULATES` with `HIGH` fidelity on the activation node, since the chemistry is identical and the horse literature is where the activated metabolite has been quantified. `PARTIALLY_RECAPITULATES` with `MODERATE` on the beta-oxidation node, with your muscle-versus-CNS divergence as the `limitations` line, stated so a reader cannot take the horse phenotype as evidence for the human one. Readouts on both.

I deliberately did not move `PMID:34765670` here. The foal hydratase finding is one animal with a suspected additional sensitivity, so it is a property of that case rather than of the model, and it stays in the knowledge-gap discussion where the uncertainty is the point.

### Suggestions

Taken: dextrose bound to `CHEBI:17234`; the `PMID:1247597` whole-study conclusion regraded `IN_VITRO` to `OTHER`, since it spans both the rat arm and the enzyme-preparation arm.

Declined with reason: `HP:0001985` Hypoketotic hypoglycemia would strengthen the phenocopy claim and you are right that MADD carries it. The cited source says only "profound hypoglycemia". The binding stays at `HP:0001943` until something supports the qualifier, which is the same rule that kept the ripening claim out of the evidence.

Not taken this round: `diagnosis:` section. Lower priority than the four, and I would rather land these than half-do five.

43/43 snippets over 9 references. `just validate-disorders`, `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets`, `check-environmental-evidence`, `check-entity-refs` all clean. 47435 tests pass. All model and biochemical readout targets resolve against pathophysiology nodes. Branch rebased onto current `main`.

PR #10583, opened 2026-09-02
Audit knowledge-gap discussion completeness and repair bare targets

cmungall:
## Summary

Audits every `kind: KNOWLEDGE_GAP` discussion in `kb/` for structural completeness, repairs the one class of defect no existing check could see, gates against its return, and records the rest as reported backlog.

The prose half of these entries is good: prompts state real unanswered questions, rationales explain what hangs on them, and most gaps cite evidence. The structural half was unchecked. A gap that anchors to nothing, or that proposes an experiment with no way to tell a supporting result from a refuting one, validated and rendered exactly like a complete one.

## What was broken, and why nothing caught it

`discussions[].proposed_experiments[].{perturbations,readouts}[].target` uses the `<kind>#<name>` entity-reference grammar, but a bare name written there resolved to nothing in silence. Both gates that look at targets decline this slot:

- `check-entity-refs` skips a `target` with no `#`, because the slot legitimately carries bare names in its `downstream` and `target_mechanisms` homes.
- `check-causal-targets` excludes experiment readouts outright — `test_experiment_readout_targets_are_not_checked_here` records that deliberately.

Thirteen sites across six entries were in that state.

## Repairs

**Eleven** named a real pathophysiology node in the same file and were fixed by prefixing `pathophysiology#`.

**Two, both in `Alcoholic_Liver_Disease`**, held a Gene Ontology process label rather than a node name and needed a curator judgment. After review, the AhR and lactate arms of `exp_baijiu_fraction_gnotobiotic_pathway_mediation` read as follows:

- Both **AhR** items target `pathophysiology#Kupffer-Cell and Hepatocyte Inflammatory Injury`. Anchoring them to the endotoxin node contradicted the readout's own description, which says it tests "a causal microbial-metabolite route independently of lactate and endotoxin". The new node is one of the discussion's own anchors.
- `Species-resolved microbial lactate flux` targets `pathophysiology#Lipotoxic and Oxidative Hepatocyte Stress`, matching the perturbation it is paired with instead of pointing somewhere else.

**On the two GO concepts, the outcome is not symmetrical.** Lactate keeps its grounding — `GO:0019249` is on both lactate items. AhR cannot: GO has no aryl-hydrocarbon-receptor signalling-pathway term (only `GO:0017162`, a molecular function, plus three receptor-complex terms), neither experiment class has a `molecular_functions` slot, and CHEBI has no FICZ entry. So the concept is carried where the schema does allow it — the perturbation has a `genes` descriptor for `AHR` (`hgnc:348`), which is what that arm perturbs, since it crosses arms with Ahr-deficient recipients, and the readout keeps its `AhR reporter assay` descriptor.

**Both `RESOLVED` gaps lacking a `resolution_note` now have one**, each summarizing the scoping decision its own `rationale` already states:

- `IREB2-Related_Neurodegeneration` — the 15q25.1 COPD/lung-cancer association literature is out of scope.
- `X-linked_Chondrodysplasia_Punctata_2` — the male hypomorphic phenotype is curated separately as MEND syndrome (`MONDO:0010498`).

The second is worth noting: it was written after this review's census and was caught by the new gate when `main` was merged in — the first thing the gate found that was not already known.

## The gate

`scripts/knowledge_gap_discussion_audit.py` reports seven states; two of them are breakage rather than backlog and now gate.

- `just knowledge-gap-audit` — the advisory census (`--format list --state <STATE>` for per-gap detail, `--format tsv` for a table).
- `just check-knowledge-gap-targets` — the gating half, in the `qc` dependency list and in CI as an ungated, whole-KB step beside `check-entity-refs` and `check-causal-targets`. A bare experiment target is written by a curation PR, and a curation PR matches neither pytest path filter — the same argument those two steps rest on. Both strict states are at zero, which is the condition `CLAUDE.md` sets for promoting a reported state to a hard gate.
- `tests/test_knowledge_gap_discussion_audit.py` — 19 tests, four of them on the bare-target rule itself.

## Reported, not changed

Counts move daily as curation lands, so the report states its table as a **2026-09-02 snapshot** and carries a re-check section rather than restating it. Both readings, for the states that are backlog:

| State | 2026-09-02 | 2026-09-12 |
|---|---:|---:|
| Knowledge gaps in scope | 1,781 | 2,458 |
| Proposed experiment with no decision logic | 626 | 678 |
| No `status` | 302 | **625** |
| No `attaches_to` | 47 | 50 |
| Evidence prose still arguing the retired `PARTIAL` grade | 66 | 64 |

The largest finding is unchanged in substance: **969 of 1,719 proposed experiments (56%) carry no decision logic at all** — no `decision_criterion`, no `would_support`/`would_refute`, no `supporting_outcome`/`refuting_outcome`, no `readouts`. It was 59% ten days earlier, so the share is not falling as the corpus grows: this describes how experiments are being proposed now, not a historical backlog. `would_refute` appears on 113 against `would_support`'s 374, so proposals lean toward naming only the confirming direction. And `NO_STATUS` doubled — the only state growing faster than the corpus.

Each of these is either a curatorial judgment or a multi-hundred-entry sweep that should not ride along with a review. Finding 2 deserves its own follow-up issue, filed against the merged report.

Full write-up in [`docs/reports/knowledge-gap-discussion-review-2026-09-02.md`](docs/reports/knowledge-gap-discussion-review-2026-09-02.md), added to the mkdocs nav.

## Validation

`linkml-validate` and `linkml-term-validator` on the changed entries; `check_entity_refs`, `check_causal_targets`, `check_enum_values` and `check_duplicate_yaml_keys` whole-KB; history-record validation on all nine records; `uvx ruff check .`; and pytest across the audit, entity-ref, history, causal-target, render and discussions-export lanes.

This PR touches `kb/` and infra together, so it carries the `scope-override` label.

https://claude.ai/code/session_01MP5wSUCasm5ybnhcboEfSR


cmungall:
All three blocking items are fixed, plus six of the seven suggestions. Item 1 was right and it was worse than reported: the AhR retarget also contradicted the readout's own text, not just the claim about it.

## Blocking items

**Item 1 — the AhR concept, and the retarget itself.** Two separate problems.

The claim was wrong. "The GO identity is retained in each item's own `biological_processes` descriptor" is true for lactate (`GO:0019249` is on both lactate items) and false for AhR. GO has no aryl-hydrocarbon-receptor signalling-pathway term at all — a search returns `GO:0017162` (a molecular function, and neither `ExperimentalPerturbation` nor `ExperimentalReadout` has a `molecular_functions` slot) plus three receptor-complex terms — and CHEBI has no FICZ entry. So there was no term to bind in the slot the claim named.

The target was also wrong, which suggestion 4 spotted from the other direction. The readout says it tests "a causal microbial-metabolite route independently of lactate and endotoxin", and I had anchored it to the "…and Endotoxin Translocation" node. Both AhR items now target `pathophysiology#Kupffer-Cell and Hepatocyte Inflammatory Injury` — the node that route bears on, and one of the discussion's own anchors.

The concept is now carried where the schema allows: the perturbation has a `genes` descriptor for `AHR` (`hgnc:348`), which is what that arm actually perturbs, since it crosses arms with Ahr-deficient recipients. The readout keeps its `AhR reporter assay` descriptor. The report says all of this plainly instead of overclaiming, and a new history record carries the correction with the earlier record left as written.

**Item 2 — the check now gates.** `just check-knowledge-gap-targets` runs the strict half, and it is in `qc` and in CI as an ungated, whole-KB step placed beside `check-entity-refs` and `check-causal-targets`, for the argument you made: these targets are written by curation PRs, and a curation PR matches neither pytest path filter.

**Item 3 — tests added.** `tests/test_knowledge_gap_discussion_audit.py`, 19 tests. The bare-target rule gets four of its own: a prefixed target is not flagged while a bare name is, the state is recorded once per gap while every site is detailed, and the detail line falls back to the experiment name when there is no id (suggestion 10).

## Suggestions

**Item 4 — taken, and it found a second problem.** Besides the AhR mismatch above, `Species-resolved microbial lactate flux` was pointing at a different node from the perturbation it is paired with. It now targets `pathophysiology#Lipotoxic and Oxidative Hepatocyte Stress`, matching `Stable-isotope lactate tracing and lactate add-back`. The donor-community perturbation stays on the barrier node, which is the arm it belongs to.

**Item 5 — taken.** All eight records now carry `model:` and the PR link.

**Item 6 — taken, and the count did not move.** `PARTIAL` is now matched as a whole word, and the walk covers evidence nested in `proposed_experiments[].readouts[].evidence` as well as discussion-level evidence. Finding 5 is now a count rather than a floor, and it is still 75 items in 66 gaps — no gap carried the retired grade only in nested evidence, and nothing was matching on "PARTIALLY".

**Item 7 — taken, and it was hiding a bug.** `--out` was not merely limited to `--format tsv`: it was silently ignored for `list` too, because `list_text` still printed instead of returning, so `report` was `None` and the write never happened. All three formats now build text through one path, `--out` works for each, and `--state` narrows `list` and `tsv` alike (`summary` always counts the whole corpus, and the help says so).

**Item 8 — taken as a reported state, not a gated one.** `RESOLVED_NO_DATE` fires on the `IREB2-Related_Neurodegeneration` gap. It stays out of `--strict` because unlike a resolution note, a date cannot be reconstructed by whoever notices it is missing — I have no source for when that scoping decision was made, so I did not invent one.

**Item 9 — taken.** The summary now emits the experiment count, the undecidable share and the decision-slot table, so Finding 2 is reproducible from the command. The regenerate line now says explicitly that the kind-drift candidates in Finding 6 and the Minor observations were one-off analyses the script does not emit.

## One thing I did not do

I did not open the follow-up issue for Finding 2 (59% of proposed experiments carry no decision logic). It is worth one, and I would rather it be filed against a merged report it can link to than opened from a branch.

Verification on the current head: `linkml-validate` and `linkml-term-validator` on the changed entry, `check_entity_refs`, `check_causal_targets`, `check_enum_values` and `check_duplicate_yaml_keys` whole-KB, history-record validation on all eight records, and pytest — 5,445 in the entity-ref/history sweep plus 33 in the audit, causal-target and discussions-export lanes.

The scope-guard failure on `test (3.13)` was unrelated to the review: this PR changes `kb/` and infra together, so it needed the `scope-override` label, which I have added.

---

cmungall:
This sat approved and green for nine days without merging because the branch was **700 commits behind main**. The auto-merge sweep requires the PR head to contain the current `main` (`behind_by == 0`), so it skipped this PR on every run. Merging `main` in fixes that; the approval is dismissed by the push, so this needs a fresh review.

Three conflicts, all additive — both sides had appended to the same lists, and everything from both sides is kept:

- `mkdocs.yml` — four new report entries in the nav.
- `project.justfile` — the `qc` list gained `check-cancer-origin` from main and `check-knowledge-gap-targets` from here.
- `project.justfile` — main's `model-scale-audit`, `diet-audit` and `immune-antigen-audit` recipes alongside this PR's two.

## The gate found a real case on its first contact with main

Merging brought in one new violation: `X-linked_Chondrodysplasia_Punctata_2` has a `RESOLVED` knowledge gap with no `resolution_note`. It was written after this review's census, so it is the first thing the gate caught that was not already known — which is the argument for the gate, made by the gate.

Repaired the same way as the other one: its `rationale` already states the resolution (the male hypomorphic phenotype is curated separately as MEND syndrome, `MONDO:0010498`, while a mosaic male or one with a null allele plus a supernumerary X stays in this entry), so the note summarizes that rather than inventing anything. `resolved_date` is still absent — reported, not gated, because there is no source for when the decision was made.

## The report is now explicitly a snapshot

The corpus grew 38% in ten days — 1,781 gaps to 2,458 — so refreshing the figures on every merge is a treadmill that ends with them stale again. The report now says its counts are a 2026-09-02 snapshot and carries a short re-check section giving both sets. Two things in that comparison are worth more than the numbers:

| | 2026-09-02 | 2026-09-12 |
|---|---:|---:|
| Knowledge gaps | 1,781 | 2,458 |
| Proposed experiments with no decision logic | 899 / 1,525 (59%) | 969 / 1,719 (56%) |
| No `status` | 302 | **625** |

The undecidable share is not falling as the corpus grows, so Finding 2 describes how experiments are being proposed now, not a historical backlog. And `NO_STATUS` doubled — it is the only state growing faster than the corpus, which makes it the one worth a convention if only one gets one.

## Validation on the merged tree

`check-knowledge-gap-targets` (0 in both strict states across 2,458 gaps), `check-entity-refs` (3,187 files), `check-causal-targets`, `check-enum-values` (3,283 files), `check-duplicate-keys` (4,303 files), `linkml-validate` and history-record validation on the changed entry and its new record, `uvx ruff check .`, and the 19 audit tests.

---

cmungall:
`claude-review` failed on `4e039b641`, and the cause is not this PR. The review agent never ran:

```
detail="You've hit your limit · resets 8pm (UTC)"  num_turns=1  cost=$0
```

That is the Anthropic credential the review workflow uses being out of quota, which the workflow's own error handler calls out as "an account-quota issue on the credential in use, NOT a problem with this PR". It also names the likely cause: the workflow prefers `ANTHROPIC_API_KEY` (pay-per-use, no weekly cap) and falls back to `CLAUDE_CODE_OAUTH_TOKEN` only when no API key is set, so a weekly-limit error points to `ANTHROPIC_API_KEY` being unset and the run having used the subscription token. **This is repo-wide, not specific to this PR** — any PR whose review fires before the reset hits the same wall.

**I am not re-running it now**, because it would hit the same limit. The re-run is scheduled for after the 20:00 UTC reset, via the `claude-code-review.yml` dispatch path `CLAUDE.md` documents. Nothing here needs a code change.

Where that leaves this PR: merging `main` in dismissed the previous approval, and the reviewer cannot issue a new one until the quota resets, so the PR is parked on a credential rather than on its content. The failed check also keeps the auto-merge sweep away, which requires nothing failing in the rollup.

`test (3.13)` is still running on this head; if it fails I will fix and push, separately from this.

The durable fix is setting the `ANTHROPIC_API_KEY` repository secret — a maintainer call, not something I should change on my own.

---

PR #10588, opened 2026-09-02
Add 10q22.3q23.3 microduplication syndrome entry

### LINKED ISSUE #10578: Curate 10q22.3q23.3 microduplication syndrome (MONDO:0017180)
Curate a dismech entry for **10q22.3q23.3 microduplication syndrome** ([MONDO:0017180](https://monarchinitiative.org/MONDO:0017180)) — a recurrent genomic duplication between low-copy repeats LCR3 and LCR4 on chromosome 10q, presenting with developmental delay, hypotonia and variable dysmorphism.

This is the reciprocal duplication of the better-known 10q22.3q23.3 microdeletion syndrome (`MONDO:0012830`), and the existing `WAPL-Related_Developmental_Disorder` entry already says so explicitly — it curates the monogenic *WAPL* disorder and records MONDO:0017180 as the separate reciprocal duplication entity. So the duplication is genuinely uncurated, and that entry is the place to look for the cytoband conventions this locus uses (`10q22.3q23.2` vs `10q22.3q23.3` are both in circulation for overlapping intervals).

The mechanistic question worth deciding early: whether the duplication phenotype is best modelled as dosage sensitivity of a specific gene in the interval (*BMPR1A* and *WAPL* are the usual candidates) or as a whole-interval copy-number effect. That choice determines whether the pathograph hangs off a named gene node or a genomic-interval node.

Stub: `stubs/10q22.3q23.3_Microduplication_Syndrome.yaml`

bpow:
Curates 10q22.3q23.3 microduplication syndrome (MONDO:0017180) — the recurrent copy-number gain between the low-copy repeats LCR3 and LCR4 on chromosome 10q. It is the reciprocal of the much better characterised 10q22.3q23.3 microdeletion syndrome, and until now only the deletion side was represented in the knowledge base.

```
NAHR between LCR3 and LCR4
        |
        v
Increased dosage of the 10q22.3q23.3 interval
        |
        +--> speech delay, motor delay, facial gestalt        (DIRECT)
        +--> microcephaly, cardiac defect, mild ID            (INDIRECT_UNKNOWN_INTERMEDIATES)
```

The split in edge types is the main curation judgement. Speech and motor delay and the facial features come from a cohort of six duplication probands; microcephaly, the cardiac defect and mild intellectual disability come from a single case report. No gene in the interval has been shown to drive the duplication phenotype, so those edges have unknown intermediates rather than being asserted as direct.

**Penetrance is variable and the entry says so.** The same duplication has been found in healthy family members of probands, which is why this is described as a susceptibility-conferring dosage change and not a fully penetrant syndrome.

### After review

Three blocking findings from `ai4c-reviewer` were addressed, and the PR was approved on the second pass:

- **`genetic:` section added** for BMPR1A, GRID1 and NRG3 as `relationship_type: SUSCEPTIBILITY`. The notes record which arm each candidacy came from: van Bon named BMPR1A and GRID1 from the *deletion* cohort, Tang proposed BMPR1A independently on the *duplication* side, and the NRG3 proposal is for a deletion-side phenotype (breast aplasia) that duplication carriers do not have.
- **Facial gestalt rebound** from `HP:0000337` (Broad forehead — one of its five components) to `HP:0001999` Abnormal facial shape, with Broad forehead promoted to a sibling phenotype.
- **Graph connectivity closed** — the gestalt node now has `sequelae` edges to all five component features; no orphan phenotypes remain.

Also taken: the PTEN counselling caveat (surveillance designed for PTEN and BMPR1A *loss*-of-function syndromes must not be applied to duplication carriers — the mechanisms run in opposite directions), and a note recording why the Siggberg reference is cached but uncited.

### Sources

| PMID | Role |
|---|---|
| [21248748](https://pubmed.ncbi.nlm.nih.gov/21248748/) | van Bon et al. 2011 — cohort of eight deletions and six duplications; source for the LCR3/LCR4 architecture, the duplication facial gestalt, speech and motor delay, and the reduced-penetrance observation |
| [26383923](https://pubmed.ncbi.nlm.nih.gov/26383923/) | **Tang et al. 2015** — 8.2 Mb duplication case report adding microcephaly, cardiac defect and mild intellectual disability |

The falcon deep-research run is committed but was thin — two references, no PMIDs — so the evidence base was built by direct PubMed search instead.

### Two things deliberately not done

- **Dropped hypotonia and behavioural difficulty** from my initial draft. I had written them from background knowledge and neither cited source supports them.
- **Left the cytoband naming conflict as a note rather than "fixing" it.** The recurrent LCR3-LCR4 rearrangement is called 10q22.3q23.2 in van Bon et al. and 10q22.3q23.3 in the MONDO label. Same entity, two conventions; the description records this rather than picking one silently. The existing `WAPL-Related_Developmental_Disorder` entry documents the same conflict and identifies MONDO:0017180 as this separate reciprocal duplication, so the two entries agree.

`treatments:` is absent and stays absent — neither cached abstract discusses management, so there is nothing quotable.

### Validation

- `just validate` — **22/22 snippets verified**, term validation passed
- `just check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets` — all clean

Deletes `stubs/10q22.3q23.3_Microduplication_Syndrome.yaml` and adds a history record.

Closes #10578

---


bpow:
All three blocking findings are fixed, plus two of the suggestions. Snippets 17/17 → **22/22 verified**.

**Finding 1 — `genetic:` section.** Added BMPR1A, GRID1 and NRG3 as `relationship_type: SUSCEPTIBILITY`, following the `22q11.2_Duplication_Syndrome` pattern you pointed at. Your point about which arm each proposal came from was the useful part and I've made it explicit in the notes rather than leaving it implied:

- **BMPR1A** — van Bon proposed it from the deletion cohort, but Tang proposed it independently on the duplication side. The notes say that second proposal is what makes the attribution credible here rather than borrowed from the reciprocal deletion.
- **GRID1** — deletion cohort only; no duplication-side evidence names it.
- **NRG3** — proposed for congenital breast aplasia, which is a *deletion* feature. Recorded because the gene is in the interval and is the one with a named developmental role, but the notes say plainly that the phenotype it was proposed for is on the other side of the rearrangement.

One mechanical note on the NRG3 snippet: I had to quote from mid-sentence (`"known to be involved in early mammary gland development in mice, is a putative candidate gene"`) because the cached full text runs the gene symbol into the surrounding words as `theNRG3gene`. The `explanation` says so.

**Findings 2 and 3 — fixed together, as you said they would be.** The gestalt is rebound to `HP:0001999` (the Wolf-Hirschhorn precedent), `Broad forehead` is promoted to its own phenotype alongside its four siblings, and the gestalt node now carries `sequelae` edges to all five components. That closes the connectivity gap — no orphans left. You were right that the asymmetry looked accidental; it was.

**Suggestion 6 — thank you, that was a real error.** PMID:26383923 is Tang et al., not Li et al. I had it wrong in both the history record and the PR body. Fixed in the record before merge, since it is append-only afterwards.

**Suggestion 7 — PTEN.** Taken, as `notes:` prose with an explicit statement that no cached source addresses it: surveillance designed for PTEN and BMPR1A *loss*-of-function syndromes must not be applied to duplication carriers, because the mechanisms run in opposite directions. That is worth having in the entry even unevidenced.

**Suggestion 5 — Siggberg.** Left uncited, but the reason is now recorded in `notes:` so the next curator does not re-derive it: the cached record is the generic 150-patient cohort abstract and the 10q case detail is full-text only.

**Suggestion 4 — `treatments:`.** Not added. You checked this yourself and reached the same place I did: neither cached abstract says anything about management, so there is no quotable snippet. I would rather leave the gap than bind `NCIT:C15747` to an uncited entry.

Two things in the review I want to acknowledge specifically, because they are checks I would not have expected a reviewer to run: verifying that the `NGR3` typo is the publication's own and matches the cache byte-for-byte, and actually building the connectivity graph rather than trusting `check-causal-targets` — which passes, because dangling and disconnected are different properties. The second one found a real defect that no gate in the repo catches.

---

PR #10591, opened 2026-09-02
Add CHEK2-related cancer predisposition entry

### LINKED ISSUE #10577: Curate CHEK2-related cancer predisposition (MONDO:0700271)
Curate a dismech entry for **CHEK2-related cancer predisposition** ([MONDO:0700271](https://monarchinitiative.org/MONDO:0700271)) — a moderate-penetrance germline cancer predisposition caused by heterozygous loss-of-function variants in *CHEK2* (`hgnc:16627`), raising the risk of breast, colorectal, prostate and thyroid cancer.

Two questions to settle first:

- *Is this in scope at all?* MONDO places it under `MONDO:0015356` (hereditary neoplastic syndrome), not under `inherited disease susceptibility`, and it has NCCN management guidance, so I think it is a disease entry rather than a susceptibility term. But it sits close to the line and is worth a second opinion.
- *Where does it stop?* CHEK2 acts through one mechanism — failed ATM-CHK2-p53/BRCA1 DNA damage checkpoint signalling — across several tumour types, which argues for a single entry. Design decisions §3a says a germline predisposition syndrome stays separate from the somatic cancers it predisposes to, so this should not absorb any breast or colorectal cancer entry.

The KB already mentions *CHEK2* as a gene inside `Hereditary_Breast_and_Ovarian_Cancer_Syndrome`, `Familial_Nonmedullary_Thyroid_Carcinoma`, `Glioma` and `Nijmegen_Breakage_Syndrome-like_Disorder`, but there is no dedicated *CHEK2* entry. The `c.1100delC` founder allele and the lower-penetrance `p.Ile157Thr` missense allele behave differently and may warrant `has_subtypes` entries.

Stub: `stubs/CHEK2-related_Cancer_Predisposition.yaml`

bpow:
Curates germline CHEK2 cancer predisposition (MONDO:0700271). CHEK2 already appeared in the knowledge base as a gene named inside other entries — hereditary breast and ovarian cancer syndrome, familial nonmedullary thyroid carcinoma, glioma, NBS-like disorder — but had no entry of its own.

Three curation calls are worth flagging, because in each case the confident version of the story is the wrong one.

**The pathograph does not assert a failed p53 checkpoint.** The obvious mechanism to write is "ATM activates CHK2, CHK2 activates p53, no CHK2 means no checkpoint, cancer follows." Two published observations contradict it: CHK2's role in activating p53-dependent arrest is described as dispensable, and CHEK2-associated breast cancers lack the homologous recombination deficiency signature that BRCA1/2 tumours show. So the node is curated as *attenuated* ATM-CHK2 signalling, the edges to tumour phenotypes are `INDIRECT_UNKNOWN_INTERMEDIATES`, and a `KNOWLEDGE_GAP` discussion carries the primary series' own statement that the relationship remains uncertain.

```
Germline CHEK2 loss of function
        |
        v
Attenuated ATM-CHK2 damage response signaling
        |
        v
Biallelic inactivation and genomic instability      (LOH in 57% of sequenced tumours)
        |
        +--> breast, prostate, testicular            (INDIRECT_UNKNOWN_INTERMEDIATES)
```

**CHEK2 is not a Li-Fraumeni gene, and the entry says so out loud.** An early case series proposed the association in TP53-negative families meeting Li-Fraumeni criteria; it has since been refuted, including by restudy of the original families. That refutation is curated as a `CONTROVERSY` discussion holding two `REFUTE` evidence items rather than left as an absence, because the claim is still in circulation and getting it wrong would imply Li-Fraumeni surveillance for CHEK2 heterozygotes.

**Colorectal, thyroid and kidney cancer are not curated as established phenotypes.** My first draft asserted them. GeneReviews grades associations beyond breast and prostate as "less well established or conflicting", and the ACMG resource manages colorectal surveillance on family history rather than carrier status, so they are recorded as one unresolved association with that grading quoted.

Risk is likewise curated as a distribution, not a number — a second discussion records that cancer risk for a heterozygote runs from population level to high risk depending on variant, family history and modifiers, which is why both management sources direct clinicians to a personalised estimate. The two commonest alleles are modelled as behaving differently: `c.1100del` truncating versus `p.Ile157Thr`, which damages protein function but has penetrance too low to be actionable alone.

### Sources

GeneReviews **CHEK2-Related Cancer Predisposition** ([40440438](https://pubmed.ncbi.nlm.nih.gov/40440438/)) is tagged in the top-level `references:` block and was used as the phenotype baseline, alongside the ACMG clinical practice resource ([37490054](https://pubmed.ncbi.nlm.nih.gov/37490054/)), the mechanism review ([33322746](https://pubmed.ncbi.nlm.nih.gov/33322746/)), the histopathologic series ([38091153](https://pubmed.ncbi.nlm.nih.gov/38091153/)), CARRIERS contralateral risk ([36623243](https://pubmed.ncbi.nlm.nih.gov/36623243/)) and the testicular germ cell tumour case-control study ([30676620](https://pubmed.ncbi.nlm.nih.gov/30676620/)).

### On the deep-research run

The falcon report resolved 10/10 of its references but cited only DOIs, so each was converted through the PMC ID converter and cited as a PMID instead — a DOI-keyed evidence item is not snippet-checked by the gating validator. The report's own term validation earned its keep: it flagged `HP:0030766` bound to "multifocal breast carcinoma" when HPO calls that term **Ear pain**, and `GO:0090305` as obsolete. Neither was used.

### Validation

- `just validate` — **33/33 snippets verified**, term validation passed
- `just check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-grading` — all clean

Deletes `stubs/CHEK2-related_Cancer_Predisposition.yaml` and adds a history record.

Closes #10577

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01VDikyAMR8621h2D5Hh3S6g)_


bpow:
All four blocking findings are fixed, plus most of the suggestions. Snippets went 33/33 → **43/43 verified**.

| Finding | What changed |
|---|---|
| 1 — "Other cancer associations" asserts colon cancer | Removed from `phenotypes:`, re-curated as a `KNOWLEDGE_GAP` discussion (`attaches_to: disease#`) keeping both evidence items. You are right and the reasoning matters: anything in `phenotypes:` exports as an asserted feature, so an `HP:0003003` binding published the opposite of what the description said. |
| 2 — bare `genetic:` section | Split into three entries: the gene, the **c.1100del** truncating founder allele, and the **p.Ile157Thr** missense allele. Frequencies, effect sizes and the lifetime-risk range are all in, from the cache lines you named. |
| 3 — the ~1% claim | Now carries its source and the allele-specific context. |
| 4 — `PMID:33322746` under-consumed | Now supplies the functional kinase-abrogation evidence for the truncating allele. |

**On the allele split, one choice is worth flagging.** The two alleles are typed differently, not just described differently: `c.1100del` is `relationship_type: CAUSATIVE`, `p.Ile157Thr` is `RISK_FACTOR`. ACMG says I157T penetrance is "in isolation not at a level of clinical actionability", so typing it as causative alongside the truncating allele would flatten exactly the distinction that drives management. If you would rather both were `CAUSATIVE` with the difference carried only in prose, say so and I will change it.

**One correction to finding 3.** You described the ~1% figure as unsourced. It is a direct quote from `PMID:38091153` — *"Germline CHEK2 variants have now been identified in ~ 1% of the population"* — so it was sourced, just not cited. Your substantive point stands, though: quoting a whole-gene figure bare, next to founder alleles at 0.87% and 2.5% with different risks, is misleading. The note now says so explicitly and the evidence item is attached.

Suggestions taken: male breast carcinoma (OR 3.13) with the conflicting-findings caveat; the GeneReviews 50%-transmission sentence on the `inheritance` entry; a `notes` on the contralateral entry explaining that HPO has no laterality-qualified breast carcinoma term and the descriptor has no laterality slot, so the two entries are distinguishable by name but not by binding; and the three cached-but-uncited DOI files are dropped from the PR.

Not done, and why: `clinical_trials:`, `animal_models:` and `histopathology:` are all real gaps, but each needs its own evidence pass rather than a quick fill, and adding them would widen a PR that is already the largest of six from this run. `cell_types:` on the pathophysiology nodes is the one I would most want next — the breast-versus-lung CHK2 activity difference would ground the tissue specificity of the tumour spectrum. Happy to add any of these here if you would rather they land now than as follow-ups.

Also pushed separately: a folded-scalar line ending in a hyphen that failed `check_folded_hyphens` on three of the six PRs in this run.

---

bpow:
Both items done. Snippets 43/43 → **45/45**.

**Finding 1 — four fragments lengthened**, using your replacement strings verbatim. The odds-ratio one is the case worth naming: five different odds ratios live in that paragraph, so `an odds ratio (OR) of 2.66 (95% CI 2.27–3.11)` verified as an exact substring while establishing nothing about *which* allele it belonged to. That is a failure mode the snippet validator cannot see — exactness and support are different properties, and I had been treating a passing validator as evidence of both.

**Finding 2 — kept as a phenotype, with both sides cited on it.** The evidence block now carries a `SUPPORT` item for the allele-specific ten-fold male risk with `c.1100delC`, and a `REFUTE` item quoting the ACMG insufficient-evidence sentence that names MBC alongside renal, thyroid and pancreatic. The description states the asymmetry rather than hiding it, gives the reason for it — the effect is allele-specific and quantified, and the reported inconsistency has a stated cause (small studies in populations lacking the founder allele) rather than being unexplained conflict — and says explicitly that a curator weighing the ACMG grading more heavily should move it into `chek2_wider_tumour_spectrum`, where the evidence for doing so is now already present.

I went with keeping it rather than moving it, but I want to be clear that is a judgement call and the entry now contains everything needed to reverse it in one edit.

Suggestions taken: `variant_origin: GERMLINE` on all three genetic entries; `Variant.functional_effects` distinguishing complete kinase abrogation from reduced-but-not-abolished function (thank you for the correction — I had the slot in the wrong class too); a `review_notes` block warning a future curator not to normalise the two `relationship_type` values and noting the `frequency` values are gnomAD allele frequencies rather than case shares; the five remaining DOI caches dropped; and a **second history record** for this round, leaving the first unedited since records are append-only.

Left alone: `clinical_trials:`, `animal_models:`, `histopathology:`, `cell_types:`. The `cell_types:` one is still the one I would most want — the breast-versus-lung CHK2 activity difference would ground the tissue specificity of the tumour spectrum — but it deserves its own evidence pass rather than being appended to a round of snippet fixes.

---

PR #10594, opened 2026-09-02
Anchor the IFT140 Syndromic Conorenal subtype to MONDO:0009964 (was: new SRTD9 entry)

### LINKED ISSUE #10579: Curate short-rib thoracic dysplasia 9 with or without polydactyly (MONDO:0009964)
Curate a dismech entry for **short-rib thoracic dysplasia 9 with or without polydactyly** ([MONDO:0009964](https://monarchinitiative.org/MONDO:0009964)) — the *IFT140* skeletal ciliopathy, historically called conorenal or Mainzer-Saldino syndrome, combining cone-shaped phalangeal epiphyses, a narrow thorax, nephronophthisis and retinal dystrophy.

I think this is a disease entry: one gene (*IFT140*, `hgnc:29077`), one mechanism (failure of IntraFlagellar Transport-A retrograde traffic in the primary cilium, so signalling cargo accumulates in the ciliary tip and Hedgehog output is disturbed). The KB precedent supports it too — `Short-Rib_Thoracic_Dysplasia_6_With_Or_Without_Polydactyly` is already its own entry rather than a subtype.

Two things to check while curating:

- MONDO gives this four parents, including `MONDO:0018770` (Jeune syndrome) and `MONDO:0100509` (IFT140-related recessive ciliopathy). The KB has a `Jeune_Asphyxiating_Thoracic_Dystrophy` entry, so the relationship between the two needs stating rather than duplicating.
- Biallelic *IFT140* also causes isolated retinitis pigmentosa and isolated cystic kidney disease. Whether those belong as `has_subtypes` here, or stay outside this entry, is the main lump/split call.
- `kb/groupings/Ciliopathies.yaml` exists and should probably gain this as a member.

Stub: `stubs/Short-rib_Thoracic_Dysplasia_9_With_Or_Without_Polydactyly.yaml`

bpow:
**This PR changed shape after review.** It originally added a new `Short-Rib_Thoracic_Dysplasia_9_With_Or_Without_Polydactyly.yaml` entry. The reviewer established that the disease was already curated, and they are right, so the new entry is deleted and its content is folded into the existing `IFT140-related_Recessive_Ciliopathy` entry instead.

MONDO:0009964 is the **Syndromic Conorenal subtype** of that entry, not a separate disease.

### Why my preflight missed it

The claim-disease duplicate check searches `origin/main` for the MONDO ID and the disease label. I ran both. Neither hits `IFT140-related_Recessive_Ciliopathy.yaml`, because that entry is named for the gene and does not contain the string "short-rib thoracic dysplasia" or MONDO:0009964. **I never searched the causal gene against `kb/disorders`**, and that is the search that would have found it in one line:

```bash
git grep -l -i -e "IFT140" origin/main -- kb/disorders
```

`kb/groupings/Ciliopathies.yaml` *was* in my grep results — it lists the existing entry as the DISEASE member for `hgnc:29077` — and I did not open it. The stub also listed `MONDO:0100509` among its `mondo_parents`, so the answer was in the file I read first.

### What this PR does

| Change | Note |
|---|---|
| `Syndromic Conorenal` gains `subtype_term: MONDO:0009964` | Fills a gap the entry documented about itself — its description said *"No non-obsolete MONDO leaf term currently anchors this conorenal subtype"* |
| Polydactyly (`HP:0010442`) added | With the caveat that it is variable, which is what the "with or without polydactyly" qualifier records |
| Cauli Ift140 ENU mouse added to `animal_models` | `modeled_mechanisms` → the IFT-A transport node, `RECAPITULATES` at `MODERATE` fidelity; `limitations` record that mid-gestation lethality with neural tube and cardiac defects means it cannot speak to the postnatal renal and retinal course |
| `INTERPRETATION` discussion added | The boundary against monoallelic IFT140 ADPKD — same gene, different disease — plus the IFT140/PKD1 0.5 Mb proximity that causes diagnostic confusion |
| **A single pathogenic variant no longer implies "not this disease"** | See below |

The mild-chest-narrowing `PMID:23418020` evidence on the reviewer's list was **already present** on the Thoracic Hypoplasia phenotype, so it was not duplicated.

The stub is still deleted. Recording `SUBTYPE` is a resolved curation outcome, not an avoided one.

### The clinically pointed addition

Sharova 2023 reports a recurrent IFT140 **exons 27–30 tandem duplication that panel and exome sequencing miss**. In both of its patients only one pathogenic variant was found by first-line NGS; the second allele came back on genome sequencing in one, and on targeted screening for that same duplication in the other. The `diagnosis:` block previously said only "molecular genetic testing" with no CNV caveat.

It now states that a single pathogenic variant on panel or exome sequencing does not exclude the diagnosis, and names the assay that recovers the missed allele — which is the part that makes it actionable rather than merely cautionary. This was a pre-existing gap in the target entry, not one the fold introduced; I did it here because it is two additive evidence items in a block this PR already amends, and because it converts a committed-but-uncited reference cache into a cited one.

### On the nosology call

The reviewer flagged @cmungall in case SRTD9 warrants a distinct entry despite the overlap. I took the fold because the evidence for it is unambiguous — the existing subtype's own `display_name` is "Syndromic conorenal ciliopathy (Mainzer-Saldino / Jeune-overlap)", which is this concept by name — and because folding is additive and reversible while a duplicate entry is not. The re-review confirmed it independently: MONDO:0009964 is a child of MONDO:0100509 and its synonyms include "Mainzer-Saldino syndrome" and "Conorenal syndrome". If a maintainer prefers a separate entry, the right follow-up is to narrow that subtype rather than restore this file.

### Declined, with reasons

- **Short ribs (`HP:0000894`) is not added.** No cached source states short ribs as an IFT140 finding. PMID:23418020 uses "short rib" only in background about JATD as a class; the cohort's own thoracic finding is the mild chest narrowing already carried as Thoracic Hypoplasia (`HP:0005257`). Binding a phenotype from the disease name would be the wrong move, and the gap is better left visible.
- **The ten remaining DOI caches stay.** Each is cited between two and nine times by the deep-research report committed here, so dropping them would leave that report's own citations unbacked.
- **`treatments:` and `parents:`** were IMPORTANT items against the standalone entry and are moot — the target entry has three treatments and a `parents:` block already.

### Validation

- `just validate` — **37/39 snippets verified**; the 2 skipped by prefix are the new `DOI:10.3390/genes14081553` items, both confirmed by hand as exact substrings of their cache
- `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens` — all clean
- Ratcheted against `origin/main`: `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-reference-titles` — no new entries

Two reference caches picked up incidental metadata rewrites from running the validator (`full_text_attempted`, a quoting change) and were reverted rather than committed, so `references_cache/` carries only additions.

Both history records target `IFT140-related_Recessive_Ciliopathy` with `event: EDIT`. The first records the preflight miss so the next curator does not repeat it.

Closes #10579

---

### DIFF
--- a/history/disorders/IFT140-related_Recessive_Ciliopathy/2026-09-02T201127Z-claude-code-682494.yaml
+++ b/history/disorders/IFT140-related_Recessive_Ciliopathy/2026-09-02T201127Z-claude-code-682494.yaml
@@ -24,4 +24,51 @@ events:
   - evidence
   summary: Add the exons 27-30 tandem duplication caveat to molecular diagnosis
   details: |
-    PLACEHOLDER
+    Follow-up on PR #10594, addressing the one reviewer finding that survived the
+    fold as a real gap in the target entry.
+
+    The reviewer flagged Sharova 2023 (DOI:10.3390/genes14081553) as the most
+    consequential of the uncited reference caches this PR commits: it reports a
+    recurrent IFT140 exons 27-30 tandem duplication that panel and exome sequencing
+    miss, and the entry's molecular diagnosis block said only "molecular genetic
+    testing" with no CNV caveat. In both of Sharova's patients only one pathogenic
+    variant was found by first-line NGS; the second allele was recovered by genome
+    sequencing in one and by targeted screening for the same duplication in the
+    other. The diagnosis description now says that a single pathogenic variant on
+    panel or exome sequencing does not exclude the diagnosis, and two evidence items
+    from that paper carry the claim.
+
+    In my PR body I had offered to do this as a separate PR since it is a
+    pre-existing gap rather than something the fold introduced. Doing it here
+    instead, because it is a two-item additive edit to a block this PR already
+    amends and it converts a committed but uncited cache into a cited one, which was
+    the reviewer's other complaint about the same file.
+
+    Not done: HP:0000894 short ribs. The subtype this PR anchors is named short-rib
+    thoracic dysplasia 9, so the phenotype is an obvious candidate, but no cached
+    source states short ribs as an IFT140 finding. PMID:23418020 mentions the short
+    rib-polydactyly spectrum only in discussing allelism between ciliopathy genes,
+    and describes these patients' own thoracic finding as mild chest narrowing,
+    which the entry already carries as Thoracic Hypoplasia (HP:0005257). Adding a
+    short-rib phenotype would need a source that states it.
+
+    The remaining ten DOI caches stay. Each is cited between two and nine times by
+    the deep-research report committed in this PR, so dropping them would leave the
+    report's own citations unbacked.
+
+    Two reference caches, PMID_22503633.md and PMID_26359340.md, picked up
+    incidental metadata rewrites from running the validator (a full_text_attempted
+    flag and a quoting change). They were reverted rather than committed, so
+    references_cache/ still carries only additions.
+
+    Validation: just validate passes with 37/39 snippets verified; the two skipped
+    are the new DOI items, both confirmed by hand as exact substrings of
+    references_cache/DOI_10.3390_genes14081553.md. check-entity-refs,
+    check-causal-targets, check-duplicate-keys, check-enum-values,
+    check-qualifier-terms, check-folded-hyphens and the four ratcheted checks against
+    origin/main all pass.
+
+    Note on this record: it was first committed with the scaffold's unfilled placeholder token as its details
+    body. The scaffolding script nests details under events:, and the script used to
+    fill it matched only a top-level key, so it silently wrote nothing. Filled in
+    place before merge.


bpow:
I said I would open the CNV caveat as a separate PR if you wanted it. On reflection that was the wrong call, so I have done it here instead — e1cba22cf.

**The exons 27–30 tandem duplication caveat is added.** You were right that this is the pointed one. `diagnosis:` now says a single pathogenic variant on panel or exome sequencing does not exclude the diagnosis, and names the assay that recovers the missed allele. Two evidence items from Sharova 2023:

- *"In both cases, at first only one pathogenic variant was detected following panel or exome NGS sequencing."*
- *"Further WGS was performed for one of them where tandem duplication was found."*

Doing it here rather than separately, because it is a two-item additive edit to a block this PR already amends, and it converts a committed-but-uncited cache into a cited one — which was your other complaint about the same file.

**Short ribs (HP:0000894) — not added, and I think that is right.** The subtype this PR anchors is literally named "short-rib thoracic dysplasia 9", so the phenotype is the obvious candidate, but no cached source states short ribs as an IFT140 finding. PMID:23418020 uses "short rib-polydactyly" only when discussing allelism between ciliopathy genes, and describes these patients' own thoracic finding as *"mild chest narrowing"* — which the entry already carries as Thoracic Hypoplasia (HP:0005257). Adding the phenotype would need a source that says it, and I would rather leave the gap visible than bind it from the disease name.

**The remaining ten DOI caches stay.** I checked each against the deep-research report committed in this PR: every one is cited there between two and nine times. Dropping them would leave the report's own citations unbacked, so the right home for them is here with the report.

**Also now moot from your list:** `treatments:` (the target entry has three), `parents:` (present), short stature (present), prevalence (present). The pathograph `conforms_to` and `cell_types` suggestions are against the pre-existing entry rather than this PR's changes, and I have not widened into them.

`just validate` passes with 37/39 snippets verified; the two skipped are the new DOI items, both confirmed by hand as exact substrings of the cache. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens` and the four ratcheted checks against `origin/main` all pass. Two reference caches picked up incidental metadata rewrites from the validator (`full_text_attempted`, a quoting change) and were reverted rather than committed, so `references_cache/` still carries only additions.

The blocking review predates the fold, so this PR needs a fresh look whenever the reviewer next runs.

---

PR #10598, opened 2026-09-02
Add immunodeficiency 89 and autoimmunity (CARD10) entry

### LINKED ISSUE #10580: Curate immunodeficiency 89 and autoimmunity (MONDO:0030484)
Curate a dismech entry for **immunodeficiency 89 and autoimmunity** ([MONDO:0030484](https://monarchinitiative.org/MONDO:0030484), IMD89) — an inborn error of immunity presenting in childhood with both susceptibility to infection and autoimmune disease, the combination that marks an immune dysregulation defect rather than a pure immunodeficiency.

The stub carries no `genes:` block, so MONDO records no causal gene for this term. Establishing the gene from OMIM is therefore the first curation step, and it decides everything downstream — if the gene turns out to be one the KB already covers under a different disease name, this becomes a duplicate or a subtype rather than a new entry. The numbered `immunodeficiency NN` series is a naming convention, not a mechanistic family, so the number tells you nothing about which pathograph applies.

Once the gene is settled, the entry should carry an IUIS immunodeficiency classification (see the `disease-classification` skill) and probably belongs in `kb/groupings/Inborn_Errors_of_Immunity.yaml`.

Stub: `stubs/Immunodeficiency_89_And_Autoimmunity.yaml`

bpow:
Curates IMD89 (MONDO:0030484) — an autosomal recessive inborn error of immunity attributed to biallelic CARD10 variants.

## I got the gene wrong first, and it is worth saying how

The stub carried **no `genes:` block**, because MONDO records no causal gene for this term. So the gene had to be established before anything else could be curated.

My initial working draft said **IKZF2 (Helios)**. I had reasoned from a plausible 2021 paper describing a loss-of-function IKZF2 variant causing immunodeficiency with immune overactivation — right decade, right phenotype shape, wrong disease. Resolving the MONDO term's OMIM cross-reference (OMIM:619632) through MedGen gave the actual description — *"autosomal recessive immune disorder characterized by adult onset of recurrent infections, allergies, microcytic anemia, and Crohn disease (Yang et al., 2020)"* — and named **CARD10**.

**Nothing in the validation stack would have caught this.** `hgnc:13177` is a real CURIE, correctly labelled IKZF2. The IKZF2 paper is a real paper with a real PMID and quotable abstract. Every identifier would have validated; the entry would simply have been about a different disease. This is Named Entity Confusion arriving through entirely valid identifiers.

The falcon run had already been launched against the IKZF2 draft, since the research recipe seeds its prompt from the disorder file. It was killed before it wrote a report, the entry was rewritten around CARD10, and research was relaunched. The committed report mentions CARD10 59 times and IKZF2 zero times.

## The evidence base is thin, and the entry says so

This entity rests on **one consanguineous family with two affected siblings**, homozygous for a single missense variant in the CARD10 coiled-coil domain, reported as a short communication. The authors call it *"the possible cause"*. No independent family has been reported.

That is curated, not smoothed over:

- `relationship_type: DISPUTED` on the genetic entry rather than `CAUSATIVE`
- a `KNOWLEDGE_GAP` discussion recording the single-family basis and what would settle it (independent families, or a ClinGen validity assertion)
- the authors' own hedge quoted as the evidence for that discussion

The gene is biologically plausible — CARD9, CARD11 and CARD14 all cause recognised immune disease, and CARD10 is highly expressed in gut epithelium — but plausibility is not replication.

## Mechanism

```
CARD10 loss of function (p.Arg420Cys, coiled-coil domain; reduced mRNA and protein)
        |
        v
Impaired CBM complex NF-kappaB signaling      (CARD10-BCL10-MALT1, downstream of GPCR/EGFR)
        |
        +--> Innate effector cell depletion with chemokine dysregulation --> recurrent infection
        |
        +--> Mucosal barrier inflammation --> Crohn disease
```

Two details make this a dysregulation defect rather than a plain immunodeficiency. Intermediate and non-classical monocytes and monocyte-derived dendritic cells were **reduced** while inflammatory chemokines were **raised** — depletion plus inflammatory drive, not global failure. And the siblings had asthma with **low** blood eosinophils, matching the Card10-knockout mouse asthma model where airway eosinophils fall but hyperresponsiveness does not.

Two evidence items are marked `directness: INDIRECT` because they support their claims through an inference step: the Crohn disease expression association (measured in unrelated IBD patients, not this family) and the mouse eosinophil dissociation.

## Intrafamilial variability

The two siblings differed markedly in severity and age at presentation despite an identical genotype. The authors attribute this to environmental factors — the elder brother had occupational metal dust exposure. That is recorded both as an `OPEN_QUESTION` discussion and as an `environmental:` entry with its own evidence, with `review_notes` explaining why a single-family modifier hypothesis is worth keeping: it is the only account the literature offers, and it bears on how a future carrier would be counselled.

## Sources

[32238915](https://pubmed.ncbi.nlm.nih.gov/32238915/) — Yang et al. 2020, the defining report. No GeneReviews chapter exists for this entity.

## Validation

- `just validate` — **20/20 snippets verified**, term validation passed
- `just check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-environmental-evidence` — all clean

Deletes `stubs/Immunodeficiency_89_And_Autoimmunity.yaml` and adds a history record.

Closes #10580

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01VDikyAMR8621h2D5Hh3S6g)_


bpow:
All eight blocking findings are fixed, plus two of the suggestions. Snippets 20/20 → **29/29 verified**.

The two I want to comment on rather than just list:

**Finding 4 — the anemia entry.** You are right that it was three different things wearing one name, and the failure mode you describe is the real cost: a query on `HP:0001935` got the right term for the wrong claim, and a query for autoimmune anemia got nothing. Now split into two phenotypes — autoimmune anemia in both siblings (bound to `HP:0002960`, since the source does not characterise it further and a haemolytic term would assert more than the paper says), and microcytic anemia in the sister, with the description saying explicitly that the source does not call it autoimmune.

**Finding 2 — `LOSS_OF_FUNCTION`.** Downgraded to `PARTIAL_LOSS_OF_FUNCTION`. This was the sharpest finding in the review: it was the one place the entry was more confident than the literature, in an entry whose whole argument is about not being. Reduced mRNA and protein establishes reduced function; it does not settle whether the allele is hypomorphic, null or activating, and the sources say so.

| Finding | Fix |
|---|---|
| 1 — six caches fetched, none consumed | Shyamsunder 2018 now supports the myeloid node (Card10 silencing impairs granulopoiesis); Moud 2024 supports the CBM node. Both `INDIRECT`. The other four are dropped. |
| 3 — genetic evidence didn't support its claim | Variant-identifying sentence added as the primary item, "elusive" quote kept as secondary for the `DISPUTED` call, plus a structured `Genetic.variants` entry for c.1258C>T p.(Arg420Cys). |
| 5 — exposure unbound and outside the pathograph | Bound to `ECTO:7000001` and linked with `environmental_effect: EXACERBATES`. |
| 6 — orphan phenotypes, two branches unmodelled | Added `Airway Epithelial CBM Signaling Failure` and `Loss of Peripheral Tolerance`. Bronchiectasis is a `sequelae` of recurrent infections. No orphans left. |
| 7 — mouse quoted but not modelled | Added as `animal_models` with `PARTIALLY_RECAPITULATES`, a `DECREASED` airway-eosinophil readout, and limitations. |
| 8 — background prose graded `HUMAN_CLINICAL` | Both regraded `OTHER`. |
| Cell-type bindings too broad (🔵) | Narrowed to intermediate monocyte, CD14+CD16-high monocyte and human myeloid DC. |

**On the ECTO choice, since it is a judgement rather than a lookup:** I used `ECTO:7000001` (exposure to dust) rather than `ECTO:9002163` (exposure to heavy metal), and recorded why in `review_notes`. The source says "metal dust" without naming the metal, so the heavy-metal term asserts a composition the evidence does not establish. If you would rather see both bound, that is a reasonable alternative.

**Not done:** seasonal urticaria, lung abscess, pulmonary bulla and autoantibody positivity as separate phenotypes; the `OMIM:619632` mapping; and `progression:`. All fair, all additive. The OMIM mapping is the one I would prioritise, since resolving that identifier is literally what produced the correct gene — happy to add these here if you would rather they land now.

---

PR #10600, opened 2026-09-02
curate: Add Mitochondrial Complex V (ATP Synthase) Deficiency, Nuclear Type 3 (MONDO:0013547, ATP5F1E)

### LINKED ISSUE #10589: Curate mitochondrial complex V (ATP synthase) deficiency, nuclear type 3 (MONDO:0013547)
Curate a dismech entry for **mitochondrial complex V (ATP synthase) deficiency, nuclear type 3** ([MONDO:0013547](https://monarchinitiative.org/MONDO:0013547)) — biallelic *ATP5F1E* variants; loss of the epsilon subunit of the F1 catalytic head, and one of the very few nuclear-gene causes of isolated ATP synthase deficiency.

Nothing anywhere in `kb/` mentions ATP5F1E and no file carries this MONDO term — a clean start. The KB now holds several `mitochondrial complex I deficiency, nuclear type N` entries and one complex II entry, so per-subunit mitochondrial entries are established convention; read one of the recent ones (`Mitochondrial_Complex_I_Deficiency_Nuclear_Type_1` or `_34`) for the shape before starting.

Two things to watch. The **maternally inherited MT-ATP6 diseases (NARP, Leigh) are a different entity** with the same enzyme — do not let their much larger literature supply evidence here. And complex V deficiency is often reported as part of a **combined OXPHOS defect**; only the isolated ATP5F1E presentation belongs in this entry.

Stub: `stubs/Mitochondrial_Complex_V_ATP_Synthase_Deficiency_Nuclear_Type_3.yaml`

kevinschaper:
Adds `kb/disorders/Mitochondrial_Complex_V_ATP_Synthase_Deficiency_Nuclear_Type_3.yaml` — MC5DN3, the isolated ATP synthase deficiency caused by biallelic *ATP5F1E* variants, and the first mitochondrial disease attributed to a nuclear-encoded **structural** subunit of complex V.

Closes #10589

## The entity, and how small it is

The whole disease is three patients from three unrelated families, all homozygous for the same allele, c.35A>G (p.Tyr12Cys), reported in two papers eleven years apart: the 2010 index case (PMID:20566710) and a 2022 nuclear-ATPase cohort that found two more and re-examined the first (PMID:34954817). The shared picture is neonatal or infantile metabolic crisis with lactic acidosis and respiratory compromise, then a stabilising course with intellectual disability, ataxia, dystonia and peripheral neuropathy.

## `entry_type`: DISEASE (standalone)

MONDO places `MONDO:0013547` under *mitochondrial respiratory chain complex deficiency* and *mitochondrial proton-transporting ATP synthase complex deficiency* — not under Leigh syndrome, and not under any grouping dismech currently curates. The entity also has one conserved pathograph shared by all three patients. So it goes in `kb/disorders/`, following the per-subunit convention already set by the `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_N` entries. The stub is deleted in this PR.

## Keeping the two rival entities out

**MT-ATP6 / MT-ATP8 disease supplied no evidence.** The deep-research report committed here mentions MT-ATP6 eleven times, and `just preflight-dr` flagged it as a rival gene at 28% of *ATP5F1E*'s mention count. Rather than cherry-pick around that, every reference in the entry was found and read directly from PubMed; each is either specifically about ATP5F1E/ATP5E or a review section explicitly about the ATP5E patient. `NARP_syndrome` and `MT-ATP6_MT-ATP8-Related_Infantile_Hypertrophic_Cardiomyopathy` appear only as differential diagnoses, and nothing was imported from them. The distinction is substantive, not pedantic: those are maternally transmitted with heteroplasmy-dependent severity, this is nuclear recessive.

**Combined-OXPHOS material is excluded the same way.** Complex V deficiency is often reported as one arm of a combined defect and a search on the enzyme returns those cohorts; none is cited. There is a dedicated `Isolated Complex V Deficiency with Preserved Respiratory Chain` node stating the boundary — in the index patient the electron-transport complexes were *increased*, which is the opposite of a combined defect.

## What the mechanism turns out to be

Epsilon matters for building the enzyme, not for running it. Activity and assembled-complex content both fall 60–70%, but the complex that does assemble is normal-sized, contains the mutant subunit, and hydrolyses and synthesises normally. Every F1 and Fo subunit falls in proportion except subunit c, which instead accumulates in a detergent-insoluble form — curated as its own node, because that asymmetry is what implicates epsilon in loading the c-ring rather than only in stabilising F1. HEK293 knockdown reproduces both halves without any patient allele.

**No phenotype frequency bands.** Three patients, two papers of unequal clinical detail. "2 of 3" is not the same claim as "30–79% of patients", so the per-patient distribution is written into each phenotype description instead. The 2022 paper's figure of dystonia in 7/10 is explicitly not used — that denominator spans four different ATPase genes and the authors flag its ascertainment bias themselves.

**Absence of cardiomyopathy is not a phenotype record.** It is the sharpest clinical discriminator from TMEM70-related complex V deficiency, but `Phenotype` has no `presence` slot, so an entry bound to HP:0001638 would assert the cardiomyopathy rather than deny it. It is recorded in the TMEM70 differential instead.

## Models

- HEK293 ATP5F1E knockdown (`CELL_LINE`), two `RECAPITULATES` links with readouts.
- *S. cerevisiae* Y11C strain, filed under `experimental_models` because yeast is not an animal, with a `FAILS_TO_RECAPITULATE` link: the mutation that destroys assembly in humans does essentially nothing in yeast. The `limitations` say plainly that this is evidence about species divergence in ATP synthase biogenesis, not evidence against pathogenicity.

## Known limitation in the sources

`PMID:21815885` (the one published attempt to find more ATP5E patients, by screening 3-methylglutaconic aciduria cohorts) is listed in `references` but carries no evidence item and no claim: its reference cache is `content_type: unavailable`, so nothing it reports can be quoted or verified. It is listed so a reader knows it exists.

## Validation

All run, all exit 0:

- `just validate` — schema, terms, 45/45 snippets verified
- `just validate-disorders` on this file only
- `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`
- `just validate-history` on the history record

### DIFF
--- a/kb/disorders/Mitochondrial_Complex_V_ATP_Synthase_Deficiency_Nuclear_Type_3.yaml
+++ b/kb/disorders/Mitochondrial_Complex_V_ATP_Synthase_Deficiency_Nuclear_Type_3.yaml
@@ -322,6 +322,16 @@ pathophysiology:
     description: >-
       Reduced oxidative ATP capacity limits sustained skeletal-muscle work.
     causal_link_type: INDIRECT_KNOWN_INTERMEDIATES
+  - target: Vomiting
+    description: >-
+      Part of the neonatal metabolic decompensation; the intermediates are
+      unknown.
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+  - target: Reduced consciousness
+    description: >-
+      The encephalopathic component of the neonatal metabolic decompensation;
+      the intermediates are unknown.
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   - target: Neonatal respiratory distress
     description: >-
       Part of the neonatal and infantile metabolic decompensation rather than a
@@ -543,6 +553,16 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: "Patient 3, a 13-year-old girl, was evaluated for intellectual disability, gait ataxia, and peripheral neuropathy. During infancy, she manifested lactic acidosis with transient respiratory failure, developmental delay, and episodes of generalized tonic-clonic seizures."
     explanation: Gait ataxia in patient 3.
+  - reference: PMID:34954817
+    reference_title: "Variants in Mitochondrial ATP Synthase Cause Variable Neurologic Phenotypes."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "we observed a non-progressive disease course with intellectual disability, ataxia, and axonal-demyelinating polyneuropathy; she had developed gait impairment with limb dystonia."
+    explanation: >-
+      The index patient's ataxia at 11-year follow-up, recorded as ataxia without
+      a gait qualifier and separately from her gait impairment. The two evidence
+      items together are why this record is bound to the parent HP:0001251 rather
+      than to Gait ataxia.
 
 - name: Dystonia
   category: Nervous System
@@ -564,6 +584,15 @@ phenotypes:
     evidence_source: HUMAN_CLINICAL
     snippet: "Metabolic abnormalities improved spontaneously, but she experienced development delay and progressively developed generalized dystonia with poor response to trialed medications; she also had visual and hearing deficits."
     explanation: Generalised refractory dystonia in patient 2.
+  - reference: PMID:34954817
+    reference_title: "Variants in Mitochondrial ATP Synthase Cause Variable Neurologic Phenotypes."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "we observed a non-progressive disease course with intellectual disability, ataxia, and axonal-demyelinating polyneuropathy; she had developed gait impairment with limb dystonia."
+    explanation: >-
+      Limb dystonia in the index patient. Generalized in one patient and limb in
+      another is why this record is bound to the parent HP:0001332 rather than to
+      Generalized dystonia.
   notes: >-
     The 2022 paper reports dystonia in 7 of its 10 patients, but that denominator
     spans four different ATPase genes and includes an ascertainment bias the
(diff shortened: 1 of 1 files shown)


kevinschaper:
All four IMPORTANT findings are addressed, plus the suggestions. One is declined with a reason.

**1. `PMID:33917760` cited by nothing** — fixed. The paragraph on the Tyr12Cys fibroblasts and the detergent-insoluble subunit c now evidences the `Subunit c Accumulation in Insoluble Form` node, and the same sentence is attached as `evidence` on the `mc5dn3_why_the_epsilon_lesion_spares_catalysis` discussion. You were right that the discussion asserted the Batten-disease connection with nothing behind it; the review's own sentence names Batten disease and fragile X, so it is now the source for exactly that claim. Graded `OTHER` as a review synthesis.

**2. Fragmented pathograph** — fixed. `Reduced Mitochondrial ATP Synthesis` now has downstream edges to all twelve phenotypes instead of to lactic acidosis alone. Nine are `INDIRECT_UNKNOWN_INTERMEDIATES` and say so in the edge description: nothing in a three-patient literature traces the route from the enzyme deficit to intellectual disability, dystonia or neuropathy, and I did not want the edges to read as if it did.

**3. `supports: NO_EVIDENCE` on the dystonia treatment** — agreed, changed to `SUPPORT`. The quoted sentence plainly bears on the claim: it documents that symptomatic pharmacotherapy was attempted, which is what the treatment record asserts, and that the response was poor. The explanation now says both.

**4. Infection-triggered decompensation not modelled** — added as an `environmental:` entry with an `influences_mechanisms` link (`environmental_effect: EXACERBATES`) onto `Reduced Mitochondrial ATP Synthesis`, carrying its own evidence.

`exposure_term` is left unbound on purpose, with the search recorded in `notes`. ECTO through the `ols:ecto` adapter has no general "exposure to infectious agent" concept — only organism-specific terms such as *exposure to Infectious bronchitis virus* and *exposure to Human Papillomavirus Infection* — and neither report names an organism. Per the "no term beats a bad one" rule, I recorded the failed search instead of forcing a match.

**Suggestions taken.** Added a `Global developmental delay` phenotype. Added a `notes` paragraph explaining the three HPO granularity choices that could read as under-specific: `Ataxia` rather than `Gait ataxia` (the index patient's follow-up says "ataxia" and separately "gait impairment"; only patient 3's is specifically gait ataxia), `Dystonia` rather than `Generalized dystonia` (limb in one patient, generalized in another), and `Peripheral neuropathy` rather than an axonal or demyelinating child term (the one characterised neuropathy is axonal-demyelinating, which no single HPO child term states).

**One suggestion declined.** There is no `omim_mappings` slot in this schema — `DiseaseMappings` carries ICD10CM, ICD11F, MONDO and NCIT only. MIM 614053 appears in a quoted evidence snippet and is now also recorded in `notes`, with that schema fact stated so the next curator does not go looking.

**GeneReviews baseline** — I could run the search, so here it is. `ATP5F1E AND GeneReviews[Book]`, `ATP5E AND GeneReviews[Book]` and `"ATP synthase deficiency" AND GeneReviews[Book]` each return 0 with no `phrasesnotfound` warning, so these are real negatives rather than the unresolvable-field failure mode that makes a bad query look like a clean one. Your judgement was right: no chapter exists. Recorded in `notes` with the queries.

**Validation after the changes**, all exit 0: `just validate` (schema, terms, 49/49 snippets), `just validate-disorders` on this file, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-environmental-evidence`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`.

kevinschaper:
Correction to the snippet count in my previous comment: it is 51/51 verified, not 49/49. I wrote the number from memory instead of from the run. The run itself was clean — `Snippets checked: 51/51 verified against cached references` — and every other check listed there did execute and exit 0.

PR #10601, opened 2026-09-02
curate: Add Spinocerebellar Ataxia, Autosomal Recessive 32 (MONDO:0859245, PRDX3)

### LINKED ISSUE #10590: Curate spinocerebellar ataxia, autosomal recessive 32 (MONDO:0859245)
Curate a dismech entry for **spinocerebellar ataxia, autosomal recessive 32 (SCAR32)** ([MONDO:0859245](https://monarchinitiative.org/MONDO:0859245)) — biallelic *PRDX3* variants; peroxiredoxin 3 is the mitochondrial thioredoxin-dependent peroxidase that clears matrix hydrogen peroxide, so this is an ataxia of failed mitochondrial redox buffering rather than of a structural or channel defect.

`MONDO:0859245` is bound nowhere in `kb/`, and PRDX3 appears only in `Pre-descemet_Corneal_Dystrophy.yaml`, which is a different disease — not prior coverage. But the ataxia space here is well populated: `kb/groupings/Cerebellar_Ataxias.yaml` exists, as does `Autosomal_Recessive_Cerebellar_Ataxia_With_Late_Onset_Spasticity.yaml`. Check whether SCAR32 should join the grouping, and check the stub's `mondo_parents` (autosomal recessive cerebellar ataxia) against what that grouping actually binds before deciding.

Stub: `stubs/Spinocerebellar_Ataxia_Autosomal_Recessive_32.yaml`

kevinschaper:
Adds `kb/disorders/Spinocerebellar_Ataxia_Autosomal_Recessive_32.yaml` — SCAR32, also called PRDX3-associated neurodegeneration (PRAN).

Closes #10590

## The entity

Peroxiredoxin 3 is the one member of the six-isoform human peroxiredoxin family confined to mitochondria, where it clears the hydrogen peroxide the respiratory chain generates. SCAR32 is therefore an ataxia of failed mitochondrial redox buffering rather than of a structural, channel, repeat-expansion or DNA-repair lesion, and it sits alongside TXN2 deficiency as one of two ultra-rare diseases of the mitochondrial thioredoxin system.

Roughly twenty patients across eight reports. The core picture is a slowly progressive cerebellar syndrome with cerebellar atrophy present in every patient, often disproportionate to the clinical severity, with dentate and cerebellar-cortical T2 hyperintensity proposed as a radiological signature. What is distinctive and unexplained is the onset range: 19 months to 35 years, across both nonsense and missense alleles, with no genotype-phenotype rule.

## `entry_type`: DISEASE (standalone)

MONDO places `MONDO:0859245` under *autosomal recessive cerebellar ataxia*; the concept is gene-defined with one conserved pathograph; and dismech already curates sibling SCAR entries (SCAR15, SCAR20) as standalone diseases. The stub is deleted in this PR.

## Should SCAR32 join `Cerebellar_Ataxias`? Yes — but not here

`kb/groupings/Cerebellar_Ataxias.yaml` binds 32 explicit `DISEASE` members, including several SCAR entries and `Autosomal Recessive Cerebellar Ataxia With Late-Onset Spasticity`, so the grouping is a curated union of entries rather than a mention list. SCAR32 satisfies both `NECESSARY` criteria: ataxia through the `HP:0001251` closure, and cerebellar localization through the `HP:0001317` closure via the curated `Cerebellar atrophy` phenotype. It belongs in the grouping.

Adding a member edits a shared file, so I have left it for a separate PR rather than bundling it here.

## The call I most want a second opinion on: no module conformance

The entry does **not** declare `conforms_to: cerebellar_purkinje_degeneration#Purkinje Neuron Degeneration`, and this was deliberate.

No autopsy or biopsy neuropathology of a SCAR32 patient has been published, no published model measures cerebellar cell populations, and the word "Purkinje" does not appear anywhere in the PRDX3 ataxia literature — I checked all eight cached papers. The available human evidence is imaging: cerebellar cortical and vermian atrophy with dentate, cortical, peduncular and olivary T2 change. That localises the lesion to the cerebellar cortex and its connections, not to a cell type. So the tissue node is `Cerebellar Cortical Degeneration`, bound to cerebellar cortex / dentate nucleus / inferior olivary complex, and there is no Purkinje cell type asserted anywhere.

This matters beyond bookkeeping: the disease involves the inferior olives and brainstem too, and if the primary lesion were olivary or afferent rather than Purkinje-intrinsic, the shared module would be the wrong causal story. A `KNOWLEDGE_GAP` discussion records the gap with two experiments that would settle it.

## Frequency discipline

Bands only where a source gives the denominator:

- `Progressive cerebellar ataxia` and `Cerebellar atrophy` are `OBLIGATE` — the latter on an explicit published statement that atrophy is ubiquitous in the reported cohort.
- Everything else is unbanded. The patients are scattered across single-case reports and small series with no pooled denominator, and two features are recorded as explicitly *normal* in other patients — hearing impairment and cognitive impairment each carry a `NO_EVIDENCE` item to that effect.
- The Japanese figure of 0.97% is a diagnostic yield in a selected repeat-expansion-negative ataxia cohort, not a population rate, so it is kept out of `prevalence` and recorded against the p.Arg207* allele instead.

One allele, p.Asp163Glu, has `functional_impact_category` left absent on purpose: the published data show enzyme loss *plus* a gain of aggregation-driven proteostatic stress the null alleles do not produce, and `FunctionalImpactEnum` has no value that says both.

## Models

- Zebrafish `prdx3` F0 crispant and *Drosophila* pan-neuronal/pan-glial knockdown in `animal_models` with `modeled_mechanisms`. The fly link is `PARTIALLY_RECAPITULATES` at `LOW` fidelity — *Drosophila* has no cerebellum, so it cannot address the regional selectivity that is the central unexplained feature.
- Patient dermal fibroblasts in `experimental_models` with a `FAILS_TO_RECAPITULATE` link: the 2024 line's respirometry was indistinguishable from controls and its oxidative-stress response was not increased. That is a real negative, curated with `direction: UNCHANGED`, and its `limitations` say what it does and does not overturn — a normal fibroblast assay cannot exclude SCAR32.

## Known limitation in the sources

`PMID:37553955` (the noncanonical-splice/exon-skipping report) is listed in `references` and mentioned in `diagnosis` but carries no evidence item: its reference cache is `content_type: unavailable`, so no snippet can be verified against it. It is listed so the splice-allele class is not invisible.

## Validation

All run, all exit 0:

- `just validate` — schema, terms, 58/58 snippets verified
- `just validate-disorders` on this file only
- `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`
- `just validate-history` on the history record
- `just preflight-dr` PASS (PRDX3 = 59 mentions, nearest rival TXN2 = 3)

### DIFF
--- a/kb/disorders/Spinocerebellar_Ataxia_Autosomal_Recessive_32.yaml
+++ b/kb/disorders/Spinocerebellar_Ataxia_Autosomal_Recessive_32.yaml
@@ -1691,7 +1691,12 @@ notes: >-
   cite for it.
 
   Not curated: `biochemical`, because no laboratory marker of SCAR32 has been
-  described beyond the research assays already captured as experimental models;
+  described. The two laboratory abnormalities in the literature - the thyroid
+  autoantibody titres and the low selenium in one Chinese patient - are curated as
+  phenotypes rather than as biomarkers, because a single patient does not
+  establish either as a marker OF the disease, and the research assays (glutathione
+  peroxidase activity, mitochondrial superoxide, respirometry) belong to the
+  experimental models that measured them;
   `progression`, because no staged natural-history study exists and the
   onset-to-severity relationship is exactly what the second open discussion says
   is unexplained; `clinical_trials` and `datasets`, because there are none.


kevinschaper:
Both blocking items are addressed. One suggested snippet is declined, and it is worth flagging because acting on it would have inverted a finding.

**Blocking item 1 — missing phenotypes.** Added `Abnormal saccadic eye movements` (HP:0000570), `Ophthalmoplegia` (HP:0000602) and `Hypokinesia` (HP:0002375), and gave `Tremor` a second evidence item for the postural tremor documented on examination. Dysphagia, Tremor and Saccadic smooth pursuit interruptions had already gone in with the commit that landed a few minutes after your review was submitted, so those three are covered too.

**The declined snippet.** The review offers `PMID:37553803` — "dysphagia was reported" as support for a Dysphagia phenotype. The sentence in the cache reads:

> He is still able to walk without support and **no** dysphagia was reported.

That patient is the 46-year-old with a pure cerebellar phenotype, and the sentence is one of the entry's negatives. Quoting the fragment would have asserted the opposite of what the paper says. Dysphagia is curated from `PMID:37731903` instead, whose review of the nine then-published patients says most eventually developed it — which is the claim the phenotype actually needs.

**Blocking item 2 — bundled treatment.** Split into four atomic entries: physiotherapy and gait/balance training (NCIT:C15302), speech and language therapy (NCIT:C159273), genetic counselling (NCIT:C15240, citing the compound-heterozygote counselling point from the Korean report), and human chorionic gonadotropin replacement (NCIT:C15986 Pharmacotherapy + CHEBI:81570 as `therapeutic_agent`, `therapeutic_modality: PEPTIDE`) with an explicit note that it treated a single-patient comorbidity and is not a SCAR32 therapy.

Occupational therapy is not added: nothing in the committed caches records it for a SCAR32 patient, and the term came from the research artifact's generic supportive-care list rather than from a report.

`target_mechanisms` links are added where they are honest. On the two rehabilitative entries I left `treatment_effect` absent rather than picking a value: the enum is INHIBITS / ACTIVATES / MODULATES / BYPASSES / RESTORES, and none of those is true of a symptomatic intervention that does not touch the mechanism. The reason is written into the link description. The gonadotropin link uses `RESTORES`, which is accurate.

**Suggestions taken.**

- Thyroid involvement now has two phenotype records — `Anti-thyroid peroxidase antibody positivity` (HP:0025379) and `Goiter` (HP:0000853) — with the measured titres and the low selenium quoted, plus the Japanese patient with autoimmune comorbidities as a second case, and a note that two patients is a reported association awaiting replication rather than a feature of the disease.
- gnomAD counts (2/251,446 overall; 1/18,394 East Asian) added to the p.Arg207* variant.
- p.Lys166* and p.Ala142Gly promoted to structured `variants`.
- The OMIM 619862-vs-619648 discrepancy is recorded in `notes`, together with the fact that this schema has no `omim_mappings` slot so it cannot be a structured mapping.

**Four alleles deliberately left in prose**, with the reason now in `notes`: c.340dupG, p.Arg15*, p.Gln220* and the two splice alleles appear only inside a whitespace-collapsed summary table in the cached full text of `PMID:41351775`, from which no quotable propositional snippet can be extracted. Promoting them would mean asserting genotypes with no verifiable evidence item.

**GeneReviews baseline** — I could run the esearch, so here it is: `PRDX3 AND GeneReviews[Book]` and `SCAR32 AND GeneReviews[Book]` both return 0, while `"spinocerebellar ataxia" AND GeneReviews[Book] AND recessive` returns five chapters. The field resolves, so the negatives are real. Your judgement was right.

**Validation after the changes**, all exit 0: `just validate` (schema, terms, 73/73 snippets), `just validate-disorders` on this file, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence`.

kevinschaper:
Correction to the snippet count in my previous comment: it is 77/77 verified, not 73/73. I wrote the number from memory instead of from the run. The run itself was clean — `Snippets checked: 77/77 verified against cached references` — and every other check listed there did execute and exit 0.

kevinschaper:
Thanks — and thanks for checking the dysphagia sentence yourself rather than leaving it as a disputed suggestion.

I had a commit ready for the four non-blocking points, but the branch is now in the merge queue and the push was rejected, which is the right outcome — I am not dequeuing an approved PR to land optional polish. It will follow as a small separate PR once this merges. What is in it:

- `Muscle weakness` (HP:0001324), from the sentence about minority features in the nine published patients. No frequency band: "a few" of nine would land in OCCASIONAL, but the review gives no count and inferring one from the word would be manufacturing a denominator.
- `Cerebellar cortical T2 hyperintensity` as its own phenotype, bound to HP:0031422 with the specificity in `preferred_term` and a note, the same way the dentate phenotype handles it. Two sources: three of four Japanese patients, and the infantile-onset patient early in her course.
- `directness: INDIRECT` on the "one patient had autoimmune comorbidities" item, for the reason you gave.
- The dentate `preferred_term` no longer echoing the HP label.

On the four DOI caches: I am keeping them, with the reason recorded in `notes`. They are the deep-research report's own resolved citations, and `scripts/qc_deep_research.py` looks up a cache file for every reference a committed report cites — so deleting them would make this report's citations unresolved under `just qc-deep-research-strict`. They duplicate cited PMIDs only because I fetched the same papers by PMID for the entry itself.

Two suggestions from the earlier review are declined, and the follow-up records why in `notes`: bradykinesia and COX-negative muscle fibres appear in no committed cache at all, and myoclonus appears only as a plus-or-minus column inside a whitespace-collapsed summary table in one cached full text, with no surrounding prose to quote.

PR #10603, opened 2026-09-02
curate: add Drug-Induced Methemoglobinemia (MONDO:0018740)

caufieldjh:
## Drug-Induced Methemoglobinemia (MONDO:0018740)

Second entry curated from the shared *Toxicologic Phenocopy Atlas*, and the cleanest pair in it. The atlas states the shared lesion in one clause: haemoglobin iron oxidised faster than the reductase recovers it. That sentence describes both diseases.

Haemoglobin iron is oxidised all the time. Ferrous haem sits in high oxygen and gives up an electron, and the red cell reduces it back so efficiently that almost none of the oxidised form is ever present. This disease is that balance losing on the production side. `Hereditary_Methemoglobinemia` is the same balance losing on the recovery side. They share their endpoint node outright.

### Why this pair and not the others

Four Tier 1 pairs in the atlas are equally clean mechanistically. Vigabatrin against `ABAT`, leflunomide against `DHODH`, and loop and thiazide diuretics against Bartter and Gitelman have **no MONDO disease term**, so none can be an entry. Warfarin embryopathy has only an obsolete one. This pair has a specific term and its genetic sibling already curated here, so the differential resolves.

### Pathophysiology: 8 atomic nodes

| Node | Claim |
|---|---|
| Baseline Haemoglobin Autoxidation | Ferrous haem oxidises continuously |
| NADH-Cytochrome b5 Reductase Recovery Capacity | The reductase drains the pool |
| Systemic Oxidant Exposure | The oxidant is present in the circulation |
| Biotransformation to an Oxidising Intermediate | Some compounds must be converted first |
| Direct Oxidation of Ferrous Haem | Electrons pass straight to the compound |
| Co-Oxidation of Ferrous Haem | Oxidation requiring haemoglobin-bound oxygen |
| Methemoglobin Accumulation | Ferric haem at a level that matters |
| Impaired Oxygen Delivery | Hypoxia with a normal arterial oxygen tension |

The balance is two nodes, both entry points. Without them the disease reads as an abnormality appearing rather than an existing balance losing, and the phenocopy stops being structural.

Biotransformation is separate from oxidation because the parent compound is inert toward haem and the metabolism is what makes it dangerous. Direct and co-oxidation are separate because their oxygen requirements are opposite: direct proceeds most readily *without* oxygen, co-oxidation *requires* it bound. Both statements are quotable, so the distinction is carried as evidence rather than prose.

### Haemolysis is not downstream of methemoglobin

`PMID:14579544` states that not all features observed in these patients are due to methemoglobin formation, and names chlorates as a case where the haemolysis is the greater threat to life. The `Hemolytic anemia` edge is therefore drawn from the oxidant exposure node, not from the accumulation. Drawing it from the accumulation would assert a causal claim the source contradicts, and the edge description says so.

### The antidote has its own failure entry

Methylene blue depends on NADP+, so in G6PD deficiency it is less effective or ineffective, and at high dose with haemolysis it can generate methemoglobin itself. That is curated as a treatment in its own right with a `REFUTE` item, because a patient who does not respond to the antidote is a recognisable clinical situation with its own management, not an absence of treatment. Exchange transfusion is the alternative the source offers.

The six-fold conversion figure carries the source's own hedge: it is clinical experience, and the same review states no controlled trial of methylene blue efficacy has been performed. Ascorbic acid, N-acetylcysteine and tocopherol are curated as a `REFUTE` entry on the review's "no confirmed benefit".

### A claim removed, then recorded as a gap

The first draft's description asserted that ferric haem "distorts the remaining ferrous subunits so they release what they carry less readily". No node carried it, no evidence supported it, and none of the four references states it. Removed.

The falcon report then returned that claim with a name, the Darling-Roughton effect, as an ordered step in its chain, citing sources this entry does not hold. Targeted searching found no citable primary source. It is curated as a `KNOWLEDGE_GAP` attached to `Impaired Oxygen Delivery`, with an experiment that would settle it, and a note saying the claim may well be correct and is simply not verified here.

### Deep research

Falcon via Edison. 104 KB, zero PMIDs (normal for falcon, which cites by DOI), 14/14 references resolved, 0 unresolved, 0 off topic. Exit code 3 was a term-validation OLS timeout on `HP:0012378`, not a content failure. NEC preflight `SKIP`, cleared manually: top gene mentions were G6PD, CYB5R3 and CYB5A, with G6PD leading, which is the right profile and confirms the antidote failure belongs in the entry.

Not taken from the report: a 2024 cohort with per-agent case counts, and rebound methemoglobinemia after methylene blue with dapsone requiring repeat dosing. Both worth having, neither curated on the report's authority.

**Correction.** An earlier version of this section said that material was left out "for want of a source I hold". That was false. `PMID:22024786` has been committed in `references_cache/` since the hereditary entry was curated, and the sibling entry cites it. Three claims it supports are now curated in this PR (see the review thread): the saturation gap and chocolate-brown blood, benzocaine as a culprit alongside dapsone, and a `Reduced Oxygen Transport Reserve` modifier node. The cohort and rebound material remain uncurated, correctly, because no source held here supports them.

### Differentials

`MONDO:0018963` hereditary methemoglobinemia (the phenocopy), haemoglobin M disease (the third route to the same node, named in the same source sentence), and cyanotic cardiac or pulmonary disease (what the patient is usually assumed to have, with the two bedside observations that point away).

### Validation

`just validate-disorders` (the batched sweep CI runs) passed. 31/31 snippets verified over 4 references. `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets`, `check-environmental-evidence`, `check-entity-refs` all clean. 36 tests pass. Graph checked for dangling targets and unwired phenotypes: none. All `reference_title` values generated programmatically from the cache. `MONDO:0018740` was seeded from the local sqlite adapter after an OLS timeout, and `conf/oak_config.yaml` restored and verified clean.

### DIFF
--- a/history/disorders/Drug-Induced_Methemoglobinemia/2026-09-02T181242Z-claude-code-af813d.yaml
+++ b/history/disorders/Drug-Induced_Methemoglobinemia/2026-09-02T181242Z-claude-code-af813d.yaml
@@ -190,3 +190,28 @@ events:
     references_cache/ for the disease and its siblings. Two of this round's items
     were closed with a reference the repository had all along, and one of them I
     had already verified a quote from.
+
+    Review round 2 (ai4c-reviewer, CHANGES_REQUESTED) confirmed both round-1 items
+    closed and raised one new item, which was a good catch about this entry's own
+    reasoning rather than about coverage.
+
+    All four diagnosis_term bindings were NCIT:C124351 Clinical Evaluation with
+    preferred_term "clinical assessment". Two of the four name procedures NCIT
+    codes specifically, and both codes were already committed here, enum-proven,
+    and already used as diagnosis_term in other entries. The point that matters:
+    this entry splits pulse oximetry from arterial blood gas into separate
+    diagnosis entries BECAUSE the diagnostic identity of the disease is that those
+    two measurements disagree, and then bound both to the same generic term, so
+    the distinction survived only in the entry names and nowhere a query could
+    reach it. Pulse oximetry is now NCIT:C38085 and the blood gas is NCIT:C217391,
+    both verified cached and enum-valid before use.
+
+    The two entries that remain on NCIT:C124351 are recognition from cyanosis and
+    exposure history. Taking a history is not a coded procedure and no more
+    specific class fits, and a note on the exposure-history entry records that
+    choice alongside the reason the other two were made specific, so the generic
+    bindings read as decisions rather than as the same oversight repeated.
+
+    The four preferred_term values were also paraphrases of the broad label, less
+    specific than the entries they described, so they recovered nothing the
+    binding dropped. All four replaced.
--- a/kb/disorders/Drug-Induced_Methemoglobinemia.yaml
+++ b/kb/disorders/Drug-Induced_Methemoglobinemia.yaml
@@ -465,7 +465,7 @@ differential_diagnoses:
 diagnosis:
 - name: Recognition from cyanosis with a saturation gap
   diagnosis_term:
-    preferred_term: clinical assessment
+    preferred_term: bedside recognition of cyanosis with a saturation gap
     term:
       id: NCIT:C124351
       label: Clinical Evaluation
@@ -501,10 +501,10 @@ diagnosis:
       States when to think of it at all, which is the step before any measurement.
 - name: Pulse oximetry is unreliable
   diagnosis_term:
-    preferred_term: clinical assessment
+    preferred_term: pulse oximetry
     term:
-      id: NCIT:C124351
-      label: Clinical Evaluation
+      id: NCIT:C38085
+      label: Pulse Oximetry
   description: >-
     Curated as its own diagnostic entry because the instrument does not simply
     read low, it reads wrongly, and a clinician who trusts it is misled rather
@@ -526,10 +526,10 @@ diagnosis:
       carried.
 - name: Arterial blood gas with co-oximetry
   diagnosis_term:
-    preferred_term: clinical assessment
+    preferred_term: arterial blood gas with co-oximetry
     term:
-      id: NCIT:C124351
-      label: Clinical Evaluation
+      id: NCIT:C217391
+      label: Diagnostic Blood Gas Testing
   description: >-
     The confirmatory test, and the one that makes the contradiction explicit. The
     dissolved oxygen tension is normal, the calculated saturation is normal, and
@@ -553,7 +553,7 @@ diagnosis:
       raised methemoglobin interpretable.
 - name: Exposure history
   diagnosis_term:
-    preferred_term: clinical assessment
+    preferred_term: oxidant exposure history
     term:
       id: NCIT:C124351
       label: Clinical Evaluation
@@ -585,6 +585,15 @@ diagnosis:
     explanation: >-
       Names the two commonest medication culprits, which is what the history is
       looking for.
+  notes: >-
+    Bound to NCIT:C124351 Clinical Evaluation rather than a procedure term,
+    because taking a history is not a coded procedure and no more specific NCIT
+    class fits. The same applies to the cyanosis recognition entry. The two
+    instrument-based entries in this section are bound to their own procedure
+    terms, NCIT:C38085 and NCIT:C217391, because the diagnostic identity of this
+    disease is that those two measurements disagree, and a shared generic binding
+    would leave that distinction in the entry names and nowhere a query could
+    reach it.
 phenotypes:
 - name: Cyanosis
   description: >-


caufieldjh:
Both items are in. The second one caught a false statement in my PR body, and I want that on the record before anything else.

### The PR body was wrong

It said the cohort and rebound material were "not taken from the report for want of a source I hold." `PMID:22024786` has been committed in `references_cache/` since the hereditary entry was curated, and `Hereditary_Methemoglobinemia.yaml` cites it. I wrote this disease without checking what its twin already used. The rationale was false, not merely thin.

Three claims from it are now curated:

- **Saturation gap and chocolate-brown blood**, in the new diagnosis section. Your point that the sibling is the template was right, and it cites the same sentence, which is correct: the diagnostic picture is shared and the exposure history is what separates the two diseases.
- **Benzocaine alongside dapsone.** This adds the local anaesthetic culprit class the entry lacked, and it matters because the exposure is topical and often incidental to a procedure.
- **A `Reduced Oxygen Transport Reserve` modifier node.** The same methemoglobin fraction is tolerated or not depending on whether oxygen transport was already impaired. It changes the treatment threshold without changing the biochemistry, so it is a modifier on delivery rather than folded into it.

### 1. Diagnosis section

Added, and this is the sharpest instance of a defect I have hit four times in this session. The description asserts "a pulse oximeter that cannot be trusted" and the cyanotic-cardiac differential turns on oximetry being unreliable rather than low.

**I pre-verified the supporting sentence earlier in this same session and then never used it.** `"Pulse oximetry is unreliable in the presence of methaemoglobinaemia."` passed my checker, and I wrote two things that depend on it, and never connected them. Verifying a quote and writing the claim it supports are two separate acts, and doing them separately is how the gap opens.

Four entries: recognition from cyanosis with a saturation gap; the oximetry unreliability as its own entry, because the instrument reads *wrongly* rather than low and a clinician who trusts it is misled rather than uninformed; arterial blood gas with co-oximetry, where normal numbers beside a blue patient are the finding; and exposure history, which is what separates this from the phenocopy and which the source says makes the diagnosis straightforward if taken.

### Suggestion taken

`progression:` record for delayed onset over hours to days after low-level inhalational or repeated dermal exposure, from `PMID:14579544`. Clinically load-bearing: the exposure may have ended before the patient presents, and one reassuring early measurement does not settle it.

### Not taken this round

`clinical_trials:` for the two NCTs, the Iolascon 2021 consensus by DOI, infant susceptibility, `animal_models:`, and the three `OTHER` to `HUMAN_CLINICAL` regrades. All reasonable. I would rather land the two blocking items and the one suggestion that changes the clinical picture than half-do eight.

One note on the regrade suggestion: I will look at those three individually rather than as a batch, since `PMID:14579544` is a review whose clinical observations are the author's synthesis rather than reported cases, and `OTHER` may be right for some of them.

9 nodes, 41/41 snippets over 5 references. `just validate-disorders`, `check-folded-hyphens`, `check-duplicate-keys`, `check-title-snippets`, `check-environmental-evidence`, `check-entity-refs` all clean. 36 tests pass. Branch rebased onto current `main`.

A `normalize-cache` pass rewrote frontmatter quoting in `PMID_22024786.md`; I reverted that rather than commit format churn to a file this PR does not own.

caufieldjh:
Fixed. You were right that this one matters more here than it usually would, and the reason is the part I had got wrong myself.

I split pulse oximetry from arterial blood gas into separate diagnosis entries *because* the diagnostic identity of this disease is that those two measurements disagree. Then I bound both to the same generic term. The distinction I had gone out of my way to preserve survived only in the entry names.

| Entry | Now |
|---|---|
| `Pulse oximetry is unreliable` | `NCIT:C38085` Pulse Oximetry |
| `Arterial blood gas with co-oximetry` | `NCIT:C217391` Diagnostic Blood Gas Testing |

Both verified cached and enum-valid before use, and both already carried as `diagnosis_term` in `Eisenmenger_Syndrome` and `Formaldehyde_Poisoning` as you said.

The two that stay on `NCIT:C124351` are `Recognition from cyanosis with a saturation gap` and `Exposure history`. Taking a history is not a coded procedure and nothing more specific fits. I added a `notes` line on the exposure-history entry recording that, and recording why the other two were made specific, so the generic bindings read as a decision rather than as the same oversight left in place twice.

You were also right about `preferred_term`. All four said "clinical assessment", which is a paraphrase of the broad label and *less* specific than the entry it describes, so nothing recovered what the binding dropped. All four replaced with terms that name the actual assessment.

41/41 snippets over 5 references. `just validate-disorders`, `check-duplicate-keys`, `check-folded-hyphens` and `check-entity-refs` clean. 36 tests pass. Branch is on current `main`.

PR #10596, opened 2026-09-02
feat: Add cyanosis, transient neonatal (HBG2)

cmungall:
## Summary

New `Disease` entry for **cyanosis, transient neonatal** (MONDO:0013511), an ultra-rare autosomal dominant fetal hemoglobinopathy caused by missense variants in **HBG2** (`hgnc:4832`, the Ggamma fetal-globin gene). HBG2 was not previously curated in dismech. Closes #10523; deletes the curated stub.

## Mechanism modeled

The variant Ggamma-globin is incorporated into fetal hemoglobin (HbF, alpha2gamma2). Because HbF dominates at birth, a large fraction of circulating hemoglobin is abnormal, and the chain confers a novel deleterious property (`functional_impact_category: NEOMORPHIC`) that branches three ways:

- **M-hemoglobin / methemoglobin formation** — His→Tyr near the heme pocket stabilizes ferric iron (Hb F-M-Fort Ripley).
- **Low oxygen affinity** — ligand-pocket substitutions lower O2 affinity and binding rate (Hb F-Toms River Val→Met; Hb F-SickKids Asn→Ser).
- **Hemoglobin instability and hemolysis** — heme-driven post-translational Met→Asp conversion enhances autoxidation and denaturation (Hb F-Toms River), giving transient hemolytic anemia.

The first two converge on **impaired fetal oxygen transport → cyanosis / hypoxemia**. A final node captures the **developmental gamma-to-beta (fetal-to-adult) hemoglobin switch**, which lowers the variant HbF fraction over the first months of life and drives **spontaneous resolution** — the "transient" in the disease name.

Phenotypes (HP): cyanosis (transient, neonatal onset), methemoglobinemia, hypoxemia, hemolytic anemia. `has_subtypes` records the Toms River variant (MONDO:0017238). Treatments (NCIT): supportive care/reassurance and genetic counseling.

## Evidence

Grounded directly in primary literature (no deep-research run):

- PMID:21561349 — NEJM, Hb F-Toms River (HBG2 missense, father+daughter, low affinity + denaturation)
- PMID:26969357 — BMJ Case Rep, Hb F-M-Fort Ripley congenital methemoglobinemia, resolution by corrected 42 weeks
- PMID:7517266 — Hum Mutat, autosomal dominant inheritance with incomplete penetrance, cyanosis lasting 1–3 months
- PMID:24939847 — JBC, heme-driven Met→Asp mechanism (IN_VITRO)
[...]

PR #10576, opened 2026-09-02
Add MSMD due to complete IL12B deficiency (IL-12p40 deficiency)

cmungall:
## What this adds

A new disorder entry for **Mendelian susceptibility to mycobacterial disease due to complete IL12B deficiency** (MONDO:0013954), the IL-12p40 etiology of MSMD. This closes a real gap: IL12B previously appeared in the KB only as a susceptibility locus (Psoriasis, Takayasu arteritis) and in IEI gene lists, never as its own Mendelian disease.

## Mechanism

IL12B encodes the **p40 subunit shared by IL-12 (p35/p40) and IL-23 (p19/p40)**. Autosomal recessive biallelic loss of function abolishes p40, so a single gene defect removes both cytokines at once. The entry models this as a causal chain:

- IL12B loss of function → absent IL-12 and IL-23 production → impaired IL-12-dependent IFN-gamma production → failed IFN-gamma-dependent macrophage activation → uncontrolled intramacrophage replication of weakly virulent mycobacteria (BCG, environmental NTM) and Salmonella
- a separate, second arm: absent IL-23 → impaired IL-23-dependent IL-17 immunity → chronic mucocutaneous candidiasis

Phenotypes cover disseminated BCG disease, non-tuberculous mycobacterial infection, non-typhoidal salmonellosis, mucocutaneous candidiasis, and reduced IFN-gamma output. Treatments cover multidrug antimycobacterial chemotherapy, recombinant IFN-gamma (mechanistically indicated because this is a production defect with an intact IFN-gamma receptor — the therapeutic dividing line within MSMD), and BCG avoidance.

## Evidence

Every evidence item is a real PMID with an exact-quote snippet. Foundational sources were verified by PubMed fetch, and I deliberately avoided the IL-12 *receptor* (IL12RB1) papers that surface alongside them:

- **PMID:9854038** — Altare 1998 (J Clin Invest), the founding report; first human disease from a cytokine gene defect
- **PMID:23429356** — Prando 2013 (Medicine), the definitive 49-patient / 30-kindred cohort
- **PMID:29589181**, **PMID:34134458** — worked cases with distinct IL12B null alleles
- **PMID:38025345** — MSMD review for the syndrome-level and therapeutic framing

## Validation

- `just validate` — passes
- `just validate-terms` — passes
- `just count-verified-snippets` — 32/32 verified
- `just validate-disorders` — passes (schema + terms + references)
- duplicate-keys, entity-refs, causal-targets, enum-values — all pass
[...]

PR #10604, opened 2026-09-02
Skip already-queued PRs so the sweep stops starving the queue

kevinschaper:
**Corrected premise.** The first version of this PR claimed `gh pr merge --squash` fails on a queue-required branch. It does not — it warns and enqueues anyway with exit status 0. The reviewer was right, and I verified it two ways: the `// only warn for now` branch in `cli/cli`'s `mergeRun` (unchanged since v2.40.0), and this session's own queue dry run, where the `--squash` call armed auto-merge at 21:43:55 rather than failing. **The sweep was never broken by `--squash`.**

What *is* broken is the thing that premise was hiding.

## The real bug: reselection starves the queue

`gh` also exits 0 when a PR is **already** in the queue. So when the sweep reselects a queued PR — which it will, because a queued PR stays open, approved, green and mergeable — the call *succeeds*: it re-posts the enqueue comment, sets `merge_completed`, and hits the one-action `break`. Every subsequent run does the same thing to the same PR, and the sweep never reaches the PRs behind it.

`read_queue_state()` now answers both questions a queue raises — is a queue in force, and which PRs are in it — in **one** GraphQL call, taken once per run *before* the candidate loop rather than per-merge inside the window between the final base-SHA check and the write. Queued PRs are skipped with a reason. The three `already queued` strings added to `BENIGN_MERGE_FAILURES` in the previous revision were unreachable (gh exits 0), so they and their test are gone.

## The diagnostic bug that made the premise look confirmed

`_gh_error` returns the first non-empty stderr line. On a queue-required branch `gh` prints its warning *before* doing its work, so **every** merge failure — whatever its actual cause — was being reported as "The merge strategy for main is set by the merge queue". It now deprioritizes lines `gh` marked as warnings, while still returning one when stderr holds nothing else (an existing test pins that case).

## What remains of the original change

Dropping `--squash` on the queue path is kept, but for accuracy rather than necessity, and the docstring says so: passing it is harmless to the merge but poisons diagnosis via the bug above. `--match-head-commit` survives on both paths — `gh` assigns it to `payload.expectedHeadOid` before the queue branch, so "act only on the verified commit" genuinely holds for an enqueue.

Also addressed: `ENQUEUE_COMMENT` no longer says "no further action needed" when the queue may still reject the PR; `merge_completed = True` carries a comment explaining it is load-bearing on the enqueue path (it suppresses the draft rollback that would eject the PR from the queue); and the command list is built conditionally instead of via an index insert.

## Not in this PR

The operator docs describing this controller as merging are being fixed in #10606, which was opened before this review and now also needs the auto-merge carve-out the review identified — on a queue-required branch `gh` implements the enqueue *by* setting `payload.auto = true`, so the standing "ordinary agents must not arm auto-merge" rule needs to say that the controller's own enqueue path is not a violation. I will fold that into #10606 rather than growing this PR.

Worth noting for the rollout: the controller has merged nothing in the last 24 hours, and `--squash` was never the reason. Its `main`-health circuit requires `test (3.13)` to be *successfully completed* on the exact current `main` SHA, and with the queue landing a PR every 10–20 minutes that build is almost always still in flight. That, the one-action-per-run cap, and the exact-ancestry requirement are the real limiters, and they belong to rollout steps 4 and 5 in #10168.

132 tests pass.

### DIFF
--- a/scripts/auto_merge_ready_prs.py
+++ b/scripts/auto_merge_ready_prs.py
@@ -495,6 +495,30 @@ class QueueState:
 
     active: bool
     queued_pr_numbers: frozenset[int]
+    # False only when the read itself failed. Without this, "no queue in
+    # force" and "could not tell" are the same value, and an inert fix is
+    # indistinguishable from the starvation bug it was meant to remove.
+    readable: bool = True
+    truncated: int = 0
+
+    def summary_line(self) -> str:
+        """One line for the run report, so inertness is visible immediately."""
+        if not self.readable:
+            return (
+                "**Merge queue:** state unavailable — falling back to direct "
+                "merge. If a queue is in force on the base branch, queued PRs "
+                "will be reselected and re-enqueued."
+            )
+        if not self.active:
+            return "**Merge queue:** not in force on the base branch."
+        line = f"**Merge queue:** active, {len(self.queued_pr_numbers)} queued."
+        if self.truncated:
+            line += (
+                f" Only the first {len(self.queued_pr_numbers)} of "
+                f"{self.truncated} entries were read, so a queued PR beyond "
+                "that may be reselected."
+            )
+        return line
 
 
 def read_queue_state(repo: str, branch: str) -> QueueState:
@@ -514,9 +538,11 @@ def read_queue_state(repo: str, branch: str) -> QueueState:
         data = json.loads(payload)
     except (subprocess.CalledProcessError, OSError, ValueError) as exc:
         print(f"WARN  could not read merge-queue state: {exc}", file=sys.stderr)
-        return QueueState(False, frozenset())
+        return QueueState(False, frozenset(), readable=False)
     if not isinstance(data, dict):
-        return QueueState(False, frozenset())
+        print("WARN  merge-queue read returned an unexpected shape",
+              file=sys.stderr)
+        return QueueState(False, frozenset(), readable=False)
     repository = (data.get("data") or {}).get("repository") or {}
     queue = repository.get("mergeQueue")
     if not queue:
@@ -529,7 +555,8 @@ def read_queue_state(repo: str, branch: str) -> QueueState:
         if isinstance(entry, dict) and (entry.get("pullRequest") or {}).get("number")
     }
     total = entries.get("totalCount")
-    if isinstance(total, int) and total > len(nodes):
+    truncated = total if isinstance(total, int) and total > len(nodes) else 0
+    if truncated:
         # Truncation is only a missed skip, never a bad merge -- but say so
         # rather than letting the page size silently bound correctness.
         print(
@@ -538,7 +565,7 @@ def read_queue_state(repo: str, branch: str) -> QueueState:
             "reselected",
             file=sys.stderr,
         )
-    return QueueState(True, frozenset(numbers))
+    return QueueState(True, frozenset(numbers), truncated=truncated)
 
 
 def merge_pr(
@@ -637,16 +664,25 @@ def render_summary(
     failed: list[dict],
     *,
     dry_run: bool = False,
+    queue_state: QueueState | None = None,
 ) -> str:
     """Render the run report.
 
     ``dry_run`` retitles the merged section: a dry run that logs "Merged 3"
     into the step summary leaves a permanent, false audit trail.
+
+    ``queue_state`` is reported unconditionally, not just when it changed the
+    outcome. A failed queue read makes this controller behave exactly as it
+    did before queue awareness -- reselecting and re-enqueueing a queued PR --
+    so "the fix is inert" must be visible in the artifact operators read,
+    rather than inferred from the absence of a skip line.
     """
     title = "## 🐑 Deterministic auto-merge sweep"
     if dry_run:
         title += " (dry run — nothing was merged)"
     lines = [title, ""]
+    if queue_state is not None:
+        lines.extend([queue_state.summary_line(), ""])
     if dry_run:
         verb = "Would merge"
     elif merged and all(row.get("queued") for row in merged):
@@ -655,6 +691,9 @@ def render_summary(
         # same permanent false audit trail this function avoids for dry runs.
         verb = "Added to the merge queue"
     elif merged and any(row.get("queued") for row in merged):
+        # Defensive: the loop breaks after one action, so `merged` holds at
+        # most one row today and this cannot be reached. Kept so the function
+        # stays correct if the one-action cap is ever lifted.
         verb = "Merged or queued"
     else:
         verb = "Merged"
@@ -838,6 +877,9 @@ def main(argv: list[str] | None = None) -> int:
 
         ready_transition_succeeded = False
         merge_completed = False
+        # Bound here, not only inside the try below: line-of-sight beats a
+        # non-local invariant for a variable read after a write has happened.
+        enqueued = False
         pr_gone = False
         try:
             if was_draft and not args.dry_run:
@@ -960,6 +1002,7 @@ def main(argv: list[str] | None = None) -> int:
         skipped,
         failed,
         dry_run=args.dry_run,
+        queue_state=queue_state,
     )
     print()
     print(report)
--- a/tests/test_auto_merge_ready_prs.py
+++ b/tests/test_auto_merge_ready_prs.py
@@ -1209,3 +1209,29 @@ def test_a_queued_pr_is_skipped_and_the_sweep_reaches_the_next_one(
     assert acted_on == [next_up], (
         "the queued PR must be skipped and the next candidate acted on"
     )
+
+
+def test_summary_reports_queue_state_so_an_inert_fix_is_visible():
+    """A failed queue read makes the controller behave as it did before queue
+    awareness, so it must be legible in the report rather than inferred from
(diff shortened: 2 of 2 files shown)


kevinschaper:
All three items fixed, and item 3 has an empirical answer.

**Item 3 — I tested it rather than reasoning about it.** On a scratch branch with its own merge-queue ruleset:

| | `mergeQueue(branch:)` | `rules/branches/<branch>` |
|---|---|---|
| ruleset active | `{"id":"MQ_kwDOQio5184AA_Be"}` | `["merge_queue"]` |
| ruleset disabled | `null` | `[]` |

So the GraphQL node *does* null out on a pause and the failure you described doesn't occur. I switched to the effective-rules endpoint anyway, because it answers exactly the question the break-glass procedure changes and so cannot drift from the pause mechanism — plus it's a cheaper REST call. The live queue on `main` was untouched by this test (a PR was in it at the time).

**Item 1 — "Merged" for a queued PR:** `merge_pr` now returns whether it enqueued. The run line prints `QUEUED #N`, and `render_summary` says "Added to the merge queue" (or "Merged or queued" for a mixed set). Same argument as the dry-run case in its docstring: the step summary is a permanent audit trail, and the queue may still reject the PR.

**Item 2 — reselection:** `"already queued"` and its variants join `BENIGN_MERGE_FAILURES`. I took the classify-as-benign option rather than the pre-filter one deliberately: the pre-filter still races (a PR can be enqueued between the read and the merge call), so the rejection has to be handled benignly regardless — and with that in place the pre-filter only saves an API call. If you'd rather have both, the GraphQL `isInMergeQueue` field is the hook and I'll add it.

**Non-blocking, adopted:** the magic `insert(5, ...)` now inserts relative to `--match-head-commit`; a non-list payload fails closed; and detection retries once so a single blip can't disable the strategy flag for a whole run. The `WARN` line after a failed courtesy comment also says "enqueued" where appropriate.

**Not adopted:** hoisting detection out of the residual base-race window — the lookup is branch-level state that a base push doesn't change, so moving it earlier trades nothing for a slightly staler read. CLAUDE.md's auto-merge section does now describe pre-queue behavior; that's real drift, but it belongs with the step-4/step-5 shepherd changes rather than this fix, and I'd rather update it once when the enqueue policy settles than twice.

131 tests pass.

cmungall:
🎉

kevinschaper:
You were right on every point, and I verified the load-bearing one independently rather than taking it on faith: the `// only warn for now` branch in `cli/cli`'s `mergeRun`, plus this session's own dry run, where the `--squash` call armed auto-merge at 21:43:55 instead of failing. The premise was wrong and the PR is rewritten around what is actually broken — title and body updated, since the body becomes the commit message.

**Reselection starvation** is now the headline fix. `read_queue_state()` reads queue-in-force and queue membership in one GraphQL call, once per run, outside the base-race window (your item 1 consequence 2), and queued PRs are skipped. The unreachable `already queued` strings and their test are gone. The new test drives `main()` through the real path rather than asserting the filter inline.

**`_gh_error`** now deprioritizes warning lines rather than dropping them — an existing test pins the case where a warning is all stderr contains, so "skip warnings" outright would have regressed it. This is the bug that made the false premise look confirmed, so thank you for spotting the mechanism and not just the symptom.

**Kept:** dropping `--squash` on the queue path, reframed as accuracy rather than necessity, with the reason in the docstring. `ENQUEUE_COMMENT` no longer promises "no further action needed". `merge_completed` carries the comment about suppressing draft rollback.

**Deferred to #10606** (docs PR, opened before your review): the operator-doc drift *and* the auto-merge carve-out you identified — the controller does now arm auto-merge, because that is how `gh` implements an enqueue, and the written rule needs to say so.

One thing your review surfaced indirectly that is worth recording: the controller has merged nothing in 24h, and `--squash` was never why. Its main-health circuit needs `test (3.13)` *completed successfully* on the exact current `main`, which with the queue landing a PR every 10–20 minutes is almost never true. That plus the one-action cap and the ancestry requirement are the real limiters — #10168 steps 4 and 5.

kevinschaper:
You asked for a counter-observation rather than a softened comment, and I have one — from this session's own half-hourly queue monitoring, not a fresh probe.

**PR #10576, 2026-09-02T23:37Z**, while it was in the queue:

```
queue entry:  isInMergeQueue=true, state=UNMERGEABLE, position=3
PR itself:    mergeable=MERGEABLE, mergeStateStatus=CLEAN, review=APPROVED, draft=false
```

It held that state for **~48 minutes** (enqueued 23:30:22, ejected 00:18:16 with reason `merge_conflict`) — a wide window in which the PR passes every predicate `evaluate` applies. Re-enqueueing succeeds (gh exits 0), so a sweep landing in that window spends its one action re-announcing a PR that is already queued. That is the budget-consuming case.

So both of us measured correctly and were describing different entry states:

| entry state | PR reports | without the skip |
|---|---|---|
| `AWAITING_CHECKS` | `UNKNOWN` / `UNKNOWN` (your measurement, #10619) | rejected by `evaluate`, loop `continue`s — sweep already moves on |
| `UNMERGEABLE` | `MERGEABLE` / `CLEAN` (#10576, 48 min) | passes every predicate; one-action budget consumed |

The comment, the test docstring and the PR body now say exactly that, with both observations and their dates, instead of generalizing from either one. Your framing of what the skip buys in the `AWAITING_CHECKS` case is adopted verbatim — cheaper path, honest SKIP reason, and not resting on undocumented UNKNOWN reporting.

Worth adding: an `UNMERGEABLE` entry is not hypothetical or rare. It is what a co-queued cache-file conflict produces, which for this repo means two curation PRs both appending to `cache/enums/*.csv` — and #10576 hit exactly that, against #10596/#10597.

**Non-blocking items, all fixed:** `entries(first:100)` now also reads `totalCount` and prints a WARN naming how many entries went unread, so truncation cannot silently bound the skip; `is_warning` uses the shared `GH_WARNING_MARKER` constant that the stripping loop uses; and I will watch the first live sweep for the "could not read merge-queue state" WARN under the job's narrower read scopes, as you suggest.

132 tests pass.

kevinschaper:
Conflicts resolved against #10653, which landed while this was open. That PR removed the main-health circuit and the exact-ancestry requirement (`base_alignment_decision`, `main_health_decision`, `get_base_sha`, `get_base_health`, `get_base_alignment`) — effectively rollout steps 4 and 5 of #10168 — so the resolution was not mechanical and is worth stating.

**How each conflict was resolved:**

- **`ENQUEUE_COMMENT` vs `NON_HUMAN_ASSIGNEE_LOGINS`** — purely additive on both sides; kept both. Reworded the enqueue comment from "unassigned" to "no human assignee" to match #10653's new `MERGE_COMMENT` phrasing, since that PR changed what the criterion means.
- **`render_summary`** — took **main's** structure, not mine. My side still carried the `circuit_open` parameter and its summary line, which #10653 deleted; keeping it would have resurrected a removed feature. Only the queued-verb logic ("Added to the merge queue" / "Merged or queued") is layered on top.
- **`_run_main` test harness** — took main's signature and added only `queue_payload`. Dropped my `comparison` parameter and the `/branches/main`, `/check-runs`, `/compare/` stubs: those APIs are no longer called, so the stubs were dead code that would have looked like coverage.
- `test_summary_does_not_claim_a_queued_pr_was_merged` no longer passes `circuit_open`.

**What this PR still contributes now that #10653 has landed**, since the overlap is real and worth being explicit about: the queue-membership skip (`read_queue_state`), the enqueue-vs-merge reporting, and the `_gh_error` warning deprioritization. The limiters I described in the body — the health circuit and the ancestry requirement — are gone, so that framing is now historical; the body's closing paragraph should be read as "what was true when this was opened".

127 tests pass, ruff clean, and the branch is `MERGEABLE` again.

Note this PR touches only `scripts/` and `tests/`, so it does not match the `schema_terms` path filter and its own run is unaffected by the EBI OLS outage that has been ejecting queue builds today (#10731 fixes that separately).

kevinschaper:
All three items addressed.

**Queue state is now in the run summary**, unconditionally — your point that an inert fix and the bug are indistinguishable in the artifact operators actually read is exactly right, and it is the same failure shape that cost the earlier misdiagnosis here. `QueueState` now separates *no queue in force* from *could not tell* (`readable`), and `render_summary` emits one of:

- `**Merge queue:** active, 3 queued.`
- `**Merge queue:** not in force on the base branch.`
- `**Merge queue:** state unavailable — falling back to direct merge. If a queue is in force on the base branch, queued PRs will be reselected and re-enqueued.`

Truncation moved there too, for the same stderr-only-visibility reason. A test pins all four renderings.

**Minor 1:** `enqueued = False` now sits beside `merge_completed = False`. Agreed on the reasoning — a non-local invariant guarding a line that runs after a write is the wrong thing to rely on.

**Minor 2:** kept, with a comment saying it is unreachable under the one-action cap and why it is retained.

On the scoped-token question: I cannot verify `repository.mergeQueue` under the `merge-ready` job's read-scoped token from here either, which is precisely why the summary line matters — the first live run will now say `state unavailable` in the step summary if the token cannot read it, instead of looking like a no-op. A `dry_run` dispatch as the first exercise is a good call.

128 tests pass.

PR #10610, opened 2026-09-02
curate(Oculocutaneous_Albinism): bind subtype_term on all nine OCA subtypes; OCA1A is SUBTYPE

### LINKED ISSUE #10608: Curate oculocutaneous albinism type 1A (MONDO:0008745)
Curate a dismech entry for **oculocutaneous albinism type 1A** ([MONDO:0008745](https://monarchinitiative.org/MONDO:0008745)) — biallelic *TYR* null variants abolishing tyrosinase activity entirely, giving the complete-absence-of-melanin end of the OCA1 spectrum.

**This is a subtype-binding question, and it has a direct precedent.** `kb/disorders/Oculocutaneous_Albinism.yaml` is curated at the umbrella term `MONDO:0018910` and its `has_subtypes` block already carries **OCA1A** explicitly (`display_name: OCA1A (TYR, tyrosinase-negative)`) alongside OCA1B through OCA8. But **none of those eight subtypes carries a `subtype_term`**, so `MONDO:0008745` is bound nowhere and `just tidy-stubs` cannot see the coverage — which is exactly why this stub survived long enough to be claimed.

That is the same defect fixed for the arterial-calcification and FCAS subtypes in #9911, where binding every subtype in the section retired four stubs rather than the two it set out to fix. Consider whether the right output here is the same: bind `subtype_term` on the umbrella's OCA1A entry — and, if the terms exist and are verifiable, on its siblings too — rather than adding a ninth standalone entry.

Decide it explicitly and defend it in the PR body. Note the stub records **one MONDO descendant** for this term, so check what that is before concluding it is a leaf. If the answer is `SUBTYPE`, record it in the stub and delete it; that is a completed curation. Binding the umbrella's subtypes edits a shared file, so if you conclude that is the fix, it may belong in its own PR — say which you chose and why.

Stub: `stubs/Oculocutaneous_Albinism_Type_1A.yaml`

kevinschaper:
Oculocutaneous albinism type 1A was claimed as a disease to curate. It is already curated — it just was not labelled with an ontology term, so no tool could tell.

`kb/disorders/Oculocutaneous_Albinism.yaml` is the umbrella entry (`MONDO:0018910`), and its `has_subtypes` block already carries **OCA1A** with a description of the tyrosinase-negative phenotype, a `TYR` genetic entry keyed to `subtype: OCA1A`, and a cited quote from PMID:21968110 that draws the exact distinction this entity is about: *"Individuals with the former lack functional tyrosinase and therefore lack melanin, while individuals with the latter produce some melanin."* What was missing was `subtype_term`. None of the nine subtypes had one, so `just tidy-stubs` matched no coverage and the stub stayed in the queue until somebody claimed it.

So: **`entry_type: SUBTYPE`**, no ninth standalone entry.

## What changed

Nine `subtype_term` blocks on the umbrella, plus the derived cache rows:

| Subtype | Term | | Subtype | Term |
|---|---|---|---|---|
| OCA1A | `MONDO:0008745` | | OCA5 | `MONDO:0014127` |
| OCA1B | `MONDO:0011749` | | OCA6 | `MONDO:0018264` |
| OCA2 | `MONDO:0008746` | | OCA7 | `MONDO:0014070` |
| OCA3 | `MONDO:0008747` | | OCA8 | `MONDO:0030899` |
| OCA4 | `MONDO:0011683` | | | |

I bound all nine rather than just OCA1A, following #9911: binding half a section is how the next reader ends up believing the unbound half is deliberate. As there, it paid off — `stubs/Oculocutaneous_Albinism_Type_2.yaml` (`MONDO:0008746`, unclaimed) is retired too, so this closes one claim and clears two stubs.

## Term verification

Every CURIE was resolved through the EBI OLS4 MONDO API before binding: not obsolete, canonical label copied from the API rather than typed, hierarchical parents inspected, and cross-checked against its OMIM xref (OCA1A 203100, OCA1B 606952, OCA2 203200, OCA3 203290, OCA4 606574, OCA5 615312, OCA6 113750, OCA7 615179, OCA8 619165).

The OCA5/OCA7 pair is the one that needed that check. MONDO's own definition of OCA7 says it was *"formerly called OCA5"*, which is an easy way to swap the two. The umbrella describes OCA5 as the 4q24 locus and OCA7 as LRMDA, and the OMIM xrefs confirm that assignment, so the bindings follow the umbrella.

## The MONDO descendant, since the issue asked

`MONDO:0008745` is **not** a leaf. Its single descendant is `MONDO:0011749`, oculocutaneous albinism type 1B — MONDO nests OCA1B *under* OCA1A rather than beside it.

Clinically they are siblings on the OCA1 allelic spectrum, separated by whether the *TYR* variants abolish tyrosinase activity or leave residual activity, and the umbrella already models them that way as two peer `has_subtypes` entries. Binding both explicitly keeps it that way: the dismech representation does not inherit MONDO's nesting, and neither subtype's coverage depends on the other's.

## Why this is one PR and not two

The issue flagged that binding the umbrella's subtypes edits a shared file and might belong on its own. I kept it together, because here the shared-file edit *is* the curation — split out, this claim's PR would contain nothing but a stub deletion, and the reasoning for the deletion would live in a different PR from the change that justifies it. The risk the split guards against is a subtype-binding sweep getting entangled with unrelated content edits, and that is not what this is: the `kb/` diff is nine `subtype_term` blocks and nothing else. No mechanism, phenotype, evidence, treatment, or prose in the umbrella was touched.

- No standalone `Oculocutaneous_Albinism_Type_1A.yaml`. The umbrella carries one shared melanin-biosynthesis pathograph that OCA1A does not diverge from; what is specific to OCA1A is the null-vs-hypomorphic allele distinction, which is already curated in the `genetic:` section.
- No deep-research run. There was no new content to source.
- `just tidy-stubs` also lists three unrelated stale stubs (colobomatous microphthalmia, KINSSHIP, NEK9). Not mine, left alone.
- A reference-cache upgrade that the validation run produced as a side effect (`PMID_21968110` and `PMID_33960688` flipping to full text) was reverted — probably a good change, but not this change.

## Validation

`just validate`, `just validate-terms`, `just validate-disorders`, `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-snippet-length`, `just check-title-snippets`, `just check-snippet-grading`, `just check-environmental-evidence`, `just check-folded-hyphens`, `just check-stubs`, `just validate-stubs`, `just check-term-cache-integrity`, `just validate-history-all` — all exit 0.

Closes #10608


kevinschaper:
Thanks — and thank you for reproducing the reference-cache flip rather than taking my word for it.

On the three non-blocking findings, all of which I agree are real and none of which I am fixing here:

**PMID:21968110 graded `HUMAN_CLINICAL` on pre-existing lines.** You are right, and it is not a close call — "Nitisinone improves eye and skin pigmentation defects in a **mouse model** of oculocutaneous albinism" is `MODEL_ORGANISM` by the classification rules in CLAUDE.md, which put any in vivo animal data there even when observational. It appears on three evidence items in the umbrella, none of them on a line this PR touches.

I am leaving it because this PR is approved and its whole argument is that the `kb/` diff is nine `subtype_term` blocks and nothing else. Re-grading three evidence items would push a new commit, drop the approval under `dismiss_stale_reviews`, and make a clean single-purpose diff into a mixed one — for a change that has nothing to do with subtype binding. I will send it as a follow-up once this merges. It is also worth checking whether the same paper is graded that way in other entries, since a grading error tends not to be alone; `check-snippet-grading` would not have caught it, because it keys on divergence *between* items rather than on any single item being wrong.

**No GeneReviews reference on the umbrella.** Agreed, and worth adding — but adding a reference means adding evidence that cites it, which is curation, not binding. Same follow-up.

**MONDO:0018135 (OCA1) unrepresented — read as deliberate.** Correct. OCA1 is the parent of OCA1A and OCA1B rather than a peer of them, so it has no `has_subtypes` entry to bind to. Binding it as a tenth sibling would assert that OCA1 is a subtype coordinate with its own children, which is exactly the nesting the PR body argues dismech should not inherit from MONDO. If OCA1 should be represented, the right shape is a `mappings.mondo_mappings` entry with a `skos:broadMatch` predicate — and per CLAUDE.md a `broadMatch` deliberately does not retire the concept from the curation queue, so it would be a modelling improvement rather than coverage. Not in this PR either.

kevinschaper:
Correcting myself on the PMID:21968110 point — I agreed with it too fast, and having now looked at the five evidence items rather than the paper's title, the picture is more defensible than either of us said.

The paper is graded **three** ways in this file, not one, and the split tracks what each quoted sentence actually reports:

| Grade | Where | What the sentence is |
|---|---|---|
| `MODEL_ORGANISM` ×2 | `animal_models[0]`, `animal_models[1]` | mouse results — "Only nitisinone-treated Tyr^c-h/c-h mice manifested increased pigmentation…" |
| `IN_VITRO` ×1 | `treatments[4]` | the human-melanocyte arm — "…increased overall melanin levels in melanocytes from a human with OCA…" |
| `HUMAN_CLINICAL` ×2 | `genetic[0]`, `genetic[1]` | background statements about the human disease — "Mutation of the tyrosinase gene (TYR) causes oculocutaneous albinism, type 1 (OCA1)…" and "Individuals with the former lack functional tyrosinase and therefore lack melanin, while individuals with the latter produce some melanin." |

That is CLAUDE.md's own instruction being followed: *"If a paper mixes sources, split evidence items so each item gets a single `evidence_source`."* The mouse results are graded as mouse results. So "it's a mouse study, therefore every item is MODEL_ORGANISM" would be the wrong fix — it would re-grade the human-melanocyte experiment and the human-disease statements as animal data.

Where I think there is still a real question, narrower than the original finding: the two `HUMAN_CLINICAL` items quote *introduction* sentences that restate established facts about human OCA1 rather than reporting an observation the paper made. `evidence_source` classifies the evidence the publication presents, and a background sentence presents none — so `OTHER` is arguably better than `HUMAN_CLINICAL` for those two. I would not call the current grading a defect; I would call it the least-bad of two imperfect options, and I am not confident enough that `OTHER` is an improvement to change it unprompted.

The second `genetic[1]` quote is also load-bearing for this PR's argument — it is the sentence that distinguishes OCA1A from OCA1B, and it is why OCA1A did not need a standalone entry.

So I am withdrawing my offer to re-grade this in a follow-up, and will open an issue instead if you still think `OTHER` is right for background sentences quoted out of a primary-research paper — that is a general convention question affecting far more than this file, not an OCA fix.

Worth noting for the tooling: `check-snippet-grading` passes this file cleanly, because it keys on the quoted *sentence* and these are five different sentences. That is by design and correct here — but it does mean one paper carrying several `evidence_source` values is invisible to it, which is the case both when the split is right (as here) and when it is wrong.

My other two replies stand unchanged.

PR #10613, opened 2026-09-02
curate: Add Mitochondrial Complex I Deficiency, Nuclear Type 14 (MONDO:0032619, NDUFA11)

### LINKED ISSUE #10607: Curate mitochondrial complex I deficiency, nuclear type 14 (MONDO:0032619)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 14** ([MONDO:0032619](https://monarchinitiative.org/MONDO:0032619)) — biallelic *NDUFA11* variants; an accessory subunit at the membrane arm, loss of which destabilises the whole complex.

`MONDO:0032619` is bound nowhere in `kb/`, and NDUFA11 appears only in `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24.yaml`, as a differential rather than as coverage. The KB now holds MC1DN types 1, 6, 23, 24 and 34 as standalone entries, so per-subunit entries are settled convention here — read a recent one for the shape, and check where MONDO places this term before assuming the Leigh relationship resolved for type 1 applies again.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14.yaml`

kevinschaper:
Mitochondrial complex I deficiency nuclear type 14 is the isolated complex I deficiency caused by biallelic *NDUFA11* variants. `MONDO:0032619` was bound nowhere in `kb/`, and NDUFA11 appeared only as a differential in the type 24 entry.

**`entry_type: DISEASE`.** `MONDO:0032619` is a leaf term whose only parent is `MONDO:0100223` (mitochondrial complex I deficiency, nuclear type). It sits beside the MC1DN types already curated as standalone entries (1, 6, 23, 24, 34), not under any of them, so the settled per-subunit convention applies without argument.

## The two things worth knowing about this entry

**It does not curate Leigh syndrome.** The issue asked me to check rather than assume the type 1 relationship transfers, and it does not. `MONDO:0032619` has no Leigh ancestor anywhere in its closure. The defining paper *does* carry a `Leigh Disease` MeSH keyword, which is the trap — but the only sentence in its abstract that mentions Leigh's disease is describing complex I deficiency *as a class*, not these patients. That is an indexing decision, not a finding, and the entry says so in `notes`.

**The clinical literature is two papers, and one of them cannot be read.** Berger 2008 (PMID:18306244) defines the entity: six patients, three unrelated families, homozygosity mapping, a splice-site allele predicted to abolish the first transmembrane domain, presenting as encephalocardiomyopathy *or* fatal infantile lactic acidemia. The 2019 letter (PMID:31074871) is titled as reporting missense variants in late-onset myopathy — `just fetch-reference` returns `content_type: unavailable` for it, no abstract and no full text. So **nothing is curated from it.** It is listed in `references`, and its unreadability is recorded as a `KNOWLEDGE_GAP` discussion, because a title promising a second, milder clinical pole is exactly the thing a future curator with journal access should reopen.

## How the mechanism is evidenced, given that

Most of the pathograph is carried by evidence that is not about NDUFA11 patients, and every such item carries `directness: INDIRECT` with the node confidence graded to match:

| Source | What it carries | Node confidence |
|---|---|---|
| PMID:27626371 — gene-edited human KO lines for every CI accessory subunit | losing one destabilises its own structural module | ESTABLISHED (agrees with three other lines) |
| PMID:27654917, PMID:27912063 — two porcine respirasome cryo-EM structures | NDUFA11 at the CI/CIII interface | **HYPOTHETICAL** |
| PMID:34106255 — *C. elegans* nduf-11 KO and RNAi | destabilised CI and supercomplexes; widened cristae junctions | HYPOTHETICAL for cristae |

The supercomplex node is the specific one and has the least patient evidence behind it, which is a tractable combination: nobody has run blue native PAGE on an NDUFA11 patient sample and asked whether the respirasome is there. That is written up as a `KNOWLEDGE_GAP` with a proposed experiment rather than asserted.

The worm work goes in `animal_models:` with two `ModelMechanismLink`s (`RECAPITULATES` at MODERATE fidelity against the destabilisation node, LOW against cristae), each with `limitations` and readouts whose `target` repeats the link's. A `HUMAN_MODEL_MISMATCH` discussion records the specific mismatch: the worm null is L2-lethal, while human patients are born and present in infancy homozygous for their allele — so either that allele is not a true null, or human complex I tolerates NDUFA11 loss better than the worm enzyme does. Those have different implications and neither has been tested.

## Two ontology findings

- **`HP:0001522` (Death in infancy) is not in the schema's `PhenotypeTerm` enum.** HPO places it under clinical course, not phenotypic abnormality. The fatal infantile outcome is therefore curated as a `progression:` phase, with the reason recorded in its notes rather than left as a mystery for the next person who tries it.
- **`GO:0005747` is obsolete**, replaced by `GO:0045271`. The deep-research report proposed it and the term validator flagged it independently. Not used.

## Deep research

A Falcon run was made for this entry and is committed under `research/`. `just preflight-dr … MONDO:0032619` **PASSES**: NDUFA11 mentioned 41 times and dominant, report OMIM 618236 matching MONDO's. Its own reference validation resolved 9/9 with 0 possible confabulations; term validation flagged 1 obsolete term and 0 unresolved.

The report holds variant-level detail from Berger 2008 full-text pages that this repo's abstract-only cache does not have — a legacy `IVS1+5G>A` nomenclature, an activated cryptic splice site, a transcript missing the final 78 bp of exon 1. **None of it is asserted as evidence**; it is recorded in `notes` as an uncited lead, since it cannot be snippet-verified against the cache.

- **No hypertrophic qualifier on the cardiomyopathy.** Hypertrophic is the pattern usual in complex I deficiency generally and would have been easy to write; the source word is *encephalocardiomyopathy*, which names the organ and not the morphology.
- **No phenotype frequencies.** The one sentence describing the cohort's presentation is disjunctive — encephalocardiomyopathy *or* fatal infantile lactic acidemia, with no split given — so there is no denominator for this disease from which a band could be derived.
- **No enzyme activity values or reference intervals**, and **no treatment section**. The retrievable record gives the biochemical category without measurements and reports no management at all. Cofactor cocktails are prescribed across primary mitochondrial disease, but a class-level prescribing practice is not this disease's treatment evidence.

## Validation

`just validate`, `validate-terms`, `validate-disorders`, `count-verified-snippets` (**31/31 verified**), `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history` — all exit 0. `pytest tests/test_data.py -k "entity_ref or subtype_foreign or unique or model_readout or failure_to_recapitulate or history or causal or prefix"` — 18544 passed.

Closes #10607

### DIFF
--- a/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14/2026-09-02T220104Z-claude-code-c29cef.yaml
+++ b/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14/2026-09-02T220104Z-claude-code-c29cef.yaml
@@ -1,0 +1,97 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14
+  path: kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14.yaml
+session:
+  id: 2026-09-02T220104Z-claude-code-c29cef
+  timestamp: '2026-09-02T22:01:04Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10607
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10613
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - experimental_models
+  - description
+  summary: Re-grade H9c2 evidence INDIRECT; add the membrane-potential specificity
+    control
+  details: |
+    Third and fourth revision sessions on PR #10613, both responding to reviewer
+    notes. Recorded in one record because they were one continuous piece of work.
+
+    SESSION 3 - directness re-grading (commit c35b2428, plus a rebase).
+
+    The reviewer left a non-blocking note that the three PMID:30531981 items under
+    `pathophysiology` were graded `directness: DIRECT` while the porcine cryo-EM and
+    C. elegans items in the same file were INDIRECT. Checked before acting, and the
+    note was right in the strong sense: CLAUDE.md's table maps "right mechanism,
+    inverted or non-human model system" to `SUPPORT` + `directness: INDIRECT`, and
+    the schema's own DirectnessEnum.INDIRECT description names "a result from an
+    inverted or non-human model system". H9c2 is a rat cardioblast line. The
+    "tests-vs-infers" reading I had used was a redefinition of the axis rather than
+    an application of it, and it had also been argued for in the PR body.
+
+    All three re-graded to INDIRECT; no DIRECT items remain in the entry. Chose
+    re-grading over the alternative the reviewer offered (a `notes:` line explaining
+    the local convention), because such a line would have documented a convention
+    that contradicts the global one and left the next reader to decide which to
+    believe. The per-item explanations and the entry description were rewritten to
+    match, stating the point the mis-grading was reaching for: strength of evidence
+    and directness diverge here, `mechanism_confidence` carries the strength, and
+    the only DIRECT-eligible evidence would come from a patient.
+
+    Also rebased onto origin/main (9a04ca7). `check-snippet-grading` had begun
+    failing on kb/disorders/Bullous_Pemphigoid.yaml, a file this PR does not touch.
+    The reviewer attributed this to a shallow clone on their runner; it reproduced
+    on a full clone, so the cause was staleness rather than environment - these
+    gates measure against current origin/main with grandfathering off, so a branch
+    can go red on a file it never touched simply by sitting still while main moves.
+    Rebase was clean, no conflicts, no cache-CSV collisions, and all three snippet
+    gates pass afterwards.
+
+    SESSION 4 - the three remaining non-blocking notes from the re-approval.
+
+    1. The ROS limitation said the SDHC effect was "of similar size". Verified
+    against the full text: NDUFA11 raised mitochondrial ROS by 62 and 30 percent,
+    SDHC by 114 and 82 percent - roughly double. The paper's own prose does describe
+    them as "similar effects", so the wording was not invented, but it was imprecise
+    against the figures. Rewritten to give both numbers and to attribute the
+    "similar" characterisation to the paper. The conclusion is unchanged: the ROS
+    rise is not specific to NDUFA11.
+
+    2. The same check surfaced something better, which the reviewer had spotted and
+    this entry had missed - a second specificity control. "In contrast to NDUFA11,
+    SDHC silencing by both siRNAs did not affect the ΔΨm." So while the ROS readout
+    is non-specific, the membrane-potential readout IS specific to NDUFA11, which is
+    exactly what one would predict if the depolarisation follows from losing a
+    proton-pumping complex rather than from reduced respiration in general. Added as
+    an evidence item on the membrane-potential readout, with the interpretation and
+    the link limitations updated to draw the ROS/ΔΨm contrast explicitly.
+
+    3. Fixed a long line in the folded description block - and while doing so found
+    that an earlier edit in session 2 had destroyed a paragraph break, joining "That
+    disagreement is curated, not smoothed over." to "Second, this entry does not
+    curate Leigh syndrome." on one line. Both fixed.
+
+    Note on process: the script applying these edits aborted partway on a failed
+    assertion and wrote nothing, while reporting a traceback that could have been
+    mistaken for a partial success. The file was re-checked with grep before any of
+    this was described, per the repo's own guidance.
+
+    Validation after both sessions, all exit 0: validate (44/44 snippets verified,
+    none skipped), validate-terms, validate-disorders, check-entity-refs,
+    check-duplicate-keys, check-causal-targets, check-snippet-grading,
+    check-snippet-length, check-title-snippets, check-folded-hyphens,
+    check-environmental-evidence, check-stubs, check-term-cache-integrity,
+    validate-history-all.
--- a/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14.yaml
+++ b/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_14.yaml
@@ -46,9 +46,11 @@ description: >-
   DIRECT-eligible evidence would come from a patient, and none exists. The
   cardioblast work also disagrees with the patients on one point, taking complexes
   III and IV down with complex I where the patients are described as having an
-  isolated complex I deficiency. That disagreement is curated, not smoothed over. Second, this entry does not curate
-  Leigh syndrome. MONDO places MONDO:0032619 under mitochondrial complex I
-  deficiency, nuclear type and under nothing else - there is no Leigh ancestor -
+  isolated complex I deficiency. That disagreement is curated, not smoothed over.
+
+  Second, this entry does not curate Leigh syndrome. MONDO places MONDO:0032619
+  under mitochondrial complex I deficiency, nuclear type and under nothing else -
+  there is no Leigh ancestor -
   and although the defining paper carries a Leigh Disease MeSH keyword, the only
   sentence in its abstract that mentions Leigh's disease is describing complex I
   deficiency as a class, not these patients. The Leigh relationship curated for
@@ -899,9 +901,17 @@ experimental_models:
       An acutely silenced rat cell line, so it does not speak to the chronic,
       whole-organism energy deficit that reaches a patient's heart and brain, and
       cardioblasts are proliferating cells whose energetic demands differ from
-      those of post-mitotic myocardium. The ROS effect was of similar size for
-      SDHC knockdown, so raised mitochondrial ROS is a general consequence of
-      losing respiratory capacity here rather than something specific to NDUFA11.
+      those of post-mitotic myocardium.
+
+      The ROS half of this link is not specific to NDUFA11: knocking down the
+      complex II subunit SDHC raised mitochondrial ROS as well, and by a larger
+      margin (114 and 82 percent, against 62 and 30 percent for NDUFA11) even
+      though the paper's own prose characterises the two as similar effects. So
+      raised mitochondrial ROS reads here as a general consequence of losing
+      respiratory capacity rather than as a signature of losing this subunit. The
+      membrane-potential half is a different matter - SDHC knockdown did not
+      affect membrane potential at all - and that specificity is recorded on the
+      readout below.
     readouts:
     - name: Cellular ATP level
       target: Impaired Mitochondrial NADH Reoxidation and ATP Synthesis
@@ -926,7 +936,11 @@ experimental_models:
       interpretation: >-
         The proton-motive force behind ATP synthesis is itself reduced, which is
         the mechanistic link between losing a proton-pumping complex and losing
-        ATP rather than merely a second symptom of it.
+        ATP rather than merely a second symptom of it. Unlike the ROS readout,
+        this one is specific to NDUFA11: silencing the complex II subunit SDHC
+        left membrane potential untouched, which is what one would predict if the
+        depolarisation follows from losing a proton pump rather than from a
+        general loss of respiratory capacity.
       evidence:
       - reference: PMID:30531981
         reference_title: "Elucidating the contribution of ETC complexes I and II to the respirasome formation in cardiac mitochondria."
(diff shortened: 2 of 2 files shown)


kevinschaper:
Note on the entry-type second opinion posted to #10607, which reached the same `DISEASE` call I did. It raised one thing I want to answer explicitly, because I considered it and decided against it.

**Why I did not split this into `has_subtypes`.** The suggestion was that the two clinical pictures — infantile encephalocardiomyopathy versus late-onset myopathy — might warrant a subtype split inside the entry rather than one set of pathophysiology nodes. That would be the right instinct if both poles were readable. They are not. The late-onset pole exists only as the *title* of PMID:31074871; `just fetch-reference` returns `content_type: unavailable` for it, so there is no abstract, no full text, and nothing to quote. Creating a `has_subtypes` entry for a clinical picture I cannot cite a single sentence about would put an unevidenced claim into a machine-readable slot.

So it is a `KNOWLEDGE_GAP` discussion instead (`ndufa11_phenotypic_spectrum_unreadable`), which says what the title promises, says the content cannot be read, and asks whoever has journal access to reopen it. If someone can read that letter, the subtype split becomes a two-line follow-up.

**On the Monarch phenotype list.** The same comment lists 15 phenotypes associated with `MONDO:0032619` in Monarch — including hypertrophic cardiomyopathy, seizures, nystagmus, brain atrophy, and optic atrophy. Those are almost certainly propagated from the OMIM 618236 clinical synopsis. I did not curate them: dismech evidence needs an exact quote verified against a cached reference, and there is no OMIM structured source in `references_cache/` to quote from. This is also why the entry binds the generic `Cardiomyopathy` (HP:0001638) rather than `Hypertrophic cardiomyopathy` (HP:0001639) — the word in the citable source is *encephalocardiomyopathy*, which names the organ and not the morphology.

That is a real gap and it is a tooling one, not a curation one: an OMIM structured source in the mould of the existing `ORPHA:`/`CGGV:` sources would make that clinical synopsis quotable, and would immediately improve every ultra-rare entry whose primary paper predates structured phenotyping. I have not opened an issue for it; happy to if that would be useful.

**Reusable evidence.** The comment was right that PMID:27654917 and PMID:27912063 name NDUFA11 in the same sentences already cited for NDUFB9 in `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_24.yaml`. Both are reused here, with the same `IN_VITRO` grading they carry in that file, and marked `directness: INDIRECT` since they are porcine structural work with no disease claim.

kevinschaper:
Self-correction pushed (ad4493d) before review, one evidence item.

The Cardiomyopathy phenotype carried a second PMID:22972949 item that quoted the paper's genetic-cause breakdown — "Approximately 25% of cases have mitochondrial DNA mutations…" — while the explanation beside it argued that cardiac involvement is a recognised part of the complex I deficiency clinical repertoire. The quote does not say that. Machine-readable content asserting one thing while the prose beside it asserts another is the defect class this repo has been bitten by most, and I would rather catch my own than have a reviewer catch it.

Replaced with the sentence that actually carries the claim — the paper's list of prevalent clinical presentations, which includes hypertrophic cardiomyopathy — and the item now also justifies its own binding choice: hypertrophic is the *class-level* pattern, so importing it would attribute a morphology to NDUFA11 patients that their own source never states, which is why the phenotype stays on the generic `HP:0001638`. The explanation flags the Leigh syndrome in that same list the same way, since it belongs to the class and not to this entity.

The 25%/25% quote stays where it is the right evidence: on the mitochondrial-genome differential diagnosis, where it quantifies how live that alternative is.

Snippets re-verified 31/31; `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-entity-refs` all re-run and exit 0.

kevinschaper:
Both findings were right, and the first one was a real miss on my part. Fixed in 31056a1.

## 1 — Jang & Javadov cached but uncited

You were right that I cached this and never used it, and right that it is the strongest mammalian evidence available for `Impaired Supercomplex Assembly`. Chasing it turned up a second problem I had not noticed: the DOI-keyed cache was abstract-only, and its snippets were being **silently skipped** by the verifier — the run reported `31/40 verified, 9 skipped by prefix`, which is a pass that verifies nothing. Re-fetched as **PMID:30531981**, which returns full text (41,916 chars). All references now use the PMID form and the entry verifies **43/43 with nothing skipped**. The uncited DOI duplicate is deleted.

The full text also settled something I had deliberately left open. Against the abstract I had recorded that the cardioblast line's species was unstated and declined to name it — the abstract's only "human" is in "associated with human diseases". The full text names **H9c2 embryonic rat cardioblasts (ATCC)** and **siRNA** explicitly, so your description was accurate; it just wasn't verifiable from what I had. The model is now curated with its real identity, `NCBITaxon:10116`, `cell_source` and `culture_system`, and the note about deliberate omission is gone as obsolete.

New `experimental_models:` entry (`CELL_LINE`), two mechanism links, seven readouts, all quantified from the full text:

| Readout | Direction | Value |
|---|---|---|
| Respirasome level | DECREASED | −13% / −30% (two siRNAs) |
| Complex III activity | DECREASED | −72% / −63% |
| Complex IV activity | DECREASED | −71% / −72% |
| Respirasome after **SDHC** knockdown | **UNCHANGED** | specificity control |
| Cellular ATP | DECREASED | to 48% |
| Mitochondrial membrane potential | DECREASED | −47% / −14% |
| Mitochondrial ROS | INCREASED | +62% / +30% |

The SDHC arm is the one I would most want kept: losing a complex II subunit reduces complex II activity *without* disintegrating the respirasome, so the NDUFA11 effect is attributable to NDUFA11 rather than to perturbing the inner membrane. That is the control the porcine structures cannot supply.

`Impaired Supercomplex Assembly` moves **HYPOTHETICAL → PROVISIONAL** — demonstrated in mammalian cells, still never measured in a patient. Your point about the discussion rationale was well taken and I rewrote it: the gap is specifically a *patient* gap, and saying "nobody has tested this" was wrong once cardioblasts had.

Your 🔵 on the `conforms_to` target was also a real inconsistency, not a nicety: the node declares conformance to `mitochondrial_dysfunction#Bioenergetic Decline and Oxidative Stress`, whose module node carries `GO:0072593` ROS INCREASED, and the conforming node had never evidenced it. The cardioblast ROS measurement supplies it and the annotation is added.

**One thing I did not smooth over.** The knockdown drops complex III and IV activity *alongside* complex I, while the patients' biochemistry is on record as an **isolated** complex I deficiency. Either the splice-site allele is milder than an siRNA knockdown, or a supercomplex phenotype is present in patients and nobody has looked. That disagreement is now written into the link `limitations`, the node description, and the discussion, as two testable alternatives rather than as agreement.

## 2 — orphaned phenotypes

Confirmed, and confirmed that patho→phenotype edges are settled convention (3,668 of them across `kb/disorders`). `check-causal-targets` passed only because every *written* target resolved — it cannot see an edge that was never written, which is worth knowing about that gate.

Added `Energy Deficit in Heart and Brain` → `Cardiomyopathy` and → `Encephalopathy`.

For the third I took neither of your two options but a third: rather than link or collapse the `Lactic Acidemia` / `Lactic Acidosis` near-duplicate, I **renamed the pathophysiology node to `Cytosolic Redox Shift and Lactate Accumulation`** and linked it to the phenotype. The node now names the mechanism — raised cytosolic NADH/NAD⁺ driving pyruvate to lactate — and the phenotype names the clinical finding, so the duplication disappears without losing the mechanistic step. If you would rather see the collapse, say so and I will do it. No phenotype is orphaned now.

## Remaining 🔵 notes

Both taken. GeneReviews decision recorded in `notes:` following the MC1DN24 precedent (no chapter covers NDUFA11; PMID:26425749 does not mention it, and tagging a Leigh overview would import the class membership this entry declines). And the "not reported" claims about the clinical split, enzyme values and treatment are reworded to say they are not quotable from the cached abstract-only record — which is the accurate statement.

## Validation

`validate` (43/43 snippets), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `check-term-cache-integrity`, `validate-history` — all exit 0. `pytest -k "model_readout or failure_to_recapitulate or entity_ref or causal"` — 10809 passed.

kevinschaper:
Thanks for the approval. I took the 🔵 rather than leaving it, because on checking it is not a stylistic disagreement — I was simply wrong.

## The `directness` grading was incorrect

You cited CLAUDE.md; I verified it and the schema says the same thing, twice:

> **CLAUDE.md**: right mechanism, inverted or non-human model system → `SUPPORT` + `directness: INDIRECT`
>
> **`DirectnessEnum.INDIRECT`**: "The quoted text asserts something from which the claim follows by an inference step — for example a therapeutic response cited as validation of the mechanism it targets, **or a result from an inverted or non-human model system**."

H9c2 is a rat line. My "tests-vs-infers" reading of the axis was a redefinition, not an application, and I argued for it in the PR body — which was wrong too. It also left the file self-inconsistent, exactly as you said: porcine cryo-EM and *C. elegans* INDIRECT, a rat cell line DIRECT.

All three items are now `INDIRECT`. There are no `DIRECT` items left in the entry.

I chose re-grading over the `notes:` line because the note would have documented a local convention that contradicts the global one, and the next reader would have had to decide which to believe.

The explanations and the entry description now carry the point the mis-grading was reaching for, which I think is worth making explicit since it is genuinely counter-intuitive: **strength of evidence and directness diverge here.** The cardioblast knockdown is by far the strongest item in the entry — the actual gene, removed, in a mammalian cell, against a negative control — and it is `INDIRECT` all the same. `mechanism_confidence` on the node is what carries strength. The only `DIRECT`-eligible evidence would come from a patient, and none exists; that is the same fact the `KNOWLEDGE_GAP` is about.

## Also rebased — your CI observation was half right

You put the `check-snippet-grading` / `check-title-snippets` / `check-snippet-length` findings down to a shallow clone with no merge base. On a full clone they reproduce, so it was not only an environment artifact:

```
kb/disorders/Bullous_Pemphigoid.yaml: PMID:29545809
  evidence_source=MODEL_ORGANISM at pathophysiology[1].evidence[0].snippet
  but evidence_source=HUMAN_CLINICAL at treatments[3].target_mechanisms[0].evidence[0].snippet
```

Your conclusion was still right — it is not a finding against this PR. The cause is that these gates measure against *current* `origin/main` with grandfathering off, and `origin/main` had moved under me (#10533, the skin-disease review). A stale branch fails on innocent files, and the documented fix is to rebase rather than to "fix" them. Rebased onto `9a04ca7` — clean, no conflicts, no cache-CSV collisions — and all three now pass.

Worth flagging for whoever owns those gates: a branch can go red on a file it never touched, purely by sitting still while `main` moves.

## Validation after re-grade and rebase

`validate` (43/43 snippets), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence`, `check-stubs`, `check-term-cache-integrity`, `validate-history-all` — all exit 0.

Head is now `c35b242`, so this needs a fresh look. Sorry for the extra cycle; shipping metadata I had verified to be wrong seemed worse than the re-review.

kevinschaper:
All three notes taken in `0725693`. The first was substantive in both directions, and you found something the entry had missed.

## The ROS comparison — you were right, and the reason is interesting

Checked against the full text:

| | siRNA-1 | siRNA-2 |
|---|---|---|
| NDUFA11 ROS | +62% | +30% |
| **SDHC ROS** | **+114%** | **+82%** |

Roughly double, as you said. What makes it worth recording rather than just correcting: **the paper's own prose calls them "similar effects"** — "Analysis of mitochondrial ROS by mitoSOX revealed similar effects of silencing induced by NDUFA11 and SDHC knockdown". So my wording tracked the source's characterisation while diverging from the source's own numbers. The limitation now gives both figures and attributes "similar" to the paper rather than asserting it. The conclusion is unchanged — the ROS rise is not specific to NDUFA11.

## The membrane-potential arm — a second specificity control I had missed

Your parenthetical was the more valuable half of the note, and it checks out:

> In contrast to NDUFA11, SDHC silencing by both siRNAs did not affect the ΔΨm

So the two bioenergetic readouts on that link behave *oppositely* with respect to specificity, and the entry was treating them as one undifferentiated caveat. ROS is a general consequence of losing respiratory capacity; **membrane potential is specific to NDUFA11** — which is exactly what one would predict if the depolarisation follows from losing a proton-*pumping* complex rather than from reduced respiration in general. Complex II does not pump protons, so SDHC knockdown leaving ΔΨm untouched is the mechanistically expected result, and it makes the ΔΨm readout attributable in a way the ROS readout is not.

Added as an evidence item on the membrane-potential readout, with the interpretation and the link `limitations` rewritten to draw the contrast explicitly instead of lumping both under "non-specific". 44/44 snippets verified.

## Long line — and a paragraph break I had destroyed

Fixed the long line. While there I found that my session-2 edit had run `That disagreement is curated, not smoothed over.` straight into `Second, this entry does not curate Leigh syndrome.` on a single line, collapsing a paragraph break in the description. Also fixed.

## History record

Added, covering both the directness re-grade and this session. You were right that it was missing.

## One process note worth passing on

The script applying these edits **aborted partway on a failed assertion and wrote nothing**, while emitting a traceback interleaved with the *previous* run's successful validation output — which read at a glance like a partial success. I only caught it because I re-grepped the file before describing the change (line still 116 chars). That is the failure mode CLAUDE.md warns about under "before describing an edit, grep the file to confirm it landed", and it is a genuinely convincing near-miss.

## Validation

`validate` (**44/44** snippets, none skipped), `validate-terms`, `validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence`, `validate-history` — all exit 0.

Head is `0725693`. This is my last push; anything further I'd take to a follow-up PR rather than churn an approved one again.

PR #10618, opened 2026-09-02
curate(SCID): bind MONDO subtype_term to all six has_subtypes entries (#10615)

### LINKED ISSUE #10615: Curate severe combined immunodeficiency due to adenosine deaminase deficiency (MONDO:0007064)
Curate a dismech entry for **severe combined immunodeficiency, autosomal recessive, T-B-NK-, due to adenosine deaminase deficiency** ([MONDO:0007064](https://monarchinitiative.org/MONDO:0007064)) — biallelic *ADA* variants; accumulating deoxyadenosine and dATP are directly lymphotoxic, which is why ADA-SCID is a systemic metabolic disease with a SCID phenotype rather than a purely immunological one, and why it was the first human disease treated by gene therapy.

**This is almost certainly a subtype-binding job, not a new entry, and it has a precedent decided hours ago.** `kb/disorders/Severe_Combined_Immunodeficiency.yaml` already carries **ADA deficiency** in `has_subtypes` (`display_name: Adenosine deaminase deficiency (ADA-SCID, T-B-NK-)`), binds `hgnc:186` inside that subtype, and carries a second ADA record in `genetic:` explaining the purine mechanism. What it does **not** have is a `subtype_term` — on that subtype or on any of its five siblings (X-linked SCID, RAG1/RAG2, IL7R, JAK3, Artemis). So `MONDO:0007064` is bound nowhere and `tidy-stubs` sees no coverage for any of them.

That is the same defect fixed for the arterial-calcification and FCAS subtypes in #9911 and, just now, for all nine oculocutaneous albinism subtypes in **#10610** — where binding the whole section retired an extra unclaimed stub as a side effect. At least one sibling stub is still in the queue (`Severe_Combined_Immunodeficiency_Autosomal_Recessive_T_Cell-negative_B_Cell-negative_NK_Cell-positive.yaml`), so the same is likely here.

Bind every subtype whose MONDO term you can resolve and verify through OLS — not just this one; #9911's rule is that binding half a section teaches the next reader the unbound half is deliberate. Do not bind a term you have not resolved, and do not otherwise modify the umbrella.

If you conclude ADA-SCID instead warrants a standalone entry, that is allowed, but argue it against #10610 and the cblA (#10260, standalone) / VWM5 (#10424, subtype) precedents rather than by default. The purine-metabolic axis is the strongest argument for standalone, since it is the one SCID whose pathology is not confined to the immune system — weigh that honestly.

Stub: `stubs/Severe_Combined_Immunodeficiency_Autosomal_Recessive_T_Cell-negative_B_Cell-negative_NK_Cell-negative_Due_To_Adenosine_Deaminase_Deficiency.yaml`

kevinschaper:
ADA-SCID was claimed as a disease to curate. It turns out to be already curated — it just had no ontology term on it, so nothing could tell.

Closes #10615

## The decision: SUBTYPE, not a standalone entry

`kb/disorders/Severe_Combined_Immunodeficiency.yaml` (`MONDO:0015974`) already carries ADA-SCID as real content, not a placeholder:

| Where | What is there for ADA |
|---|---|
| `has_subtypes` | "ADA deficiency" / "Adenosine deaminase deficiency (ADA-SCID, T-B-NK-)", with `hgnc:186` and the toxic-metabolite description |
| `genetic:` | An ADA record keyed `subtype: ADA deficiency`, with the GeneReviews quote "Adenosine deaminase (ADA) deficiency is a systemic purine metabolic disorder that primarily affects lymphocyte development, viability, and function." |
| `pathophysiology:` | A dedicated node, **Toxic purine metabolite accumulation** — `ADA` DECREASED, `GO:0006154` adenosine catabolic process DECREASED, `CHEBI:17256` 2'-deoxyadenosine INCREASED, `CHEBI:16335` adenosine INCREASED, `CL:0000893` thymocyte — with an `INDIRECT_KNOWN_INTERMEDIATES` edge into *Blocked T-lymphocyte development* naming dATP accumulation and lymphocyte apoptosis as the intermediates |
| `phenotypes:` | Two entries tagged `subtype: ADA deficiency` (absent B cells, absent NK cells) |
| `treatments:` | Pegademase (PEG-ADA) enzyme replacement and retroviral HSC gene therapy, both ADA-specific |

What was missing was `subtype_term`. None of the six subtypes had one, so `MONDO:0007064` was bound nowhere, `just tidy-stubs` matched no coverage, and the stub stayed in the queue until somebody claimed it. This is the same defect fixed in #9911 and #10610.

**The standalone argument, weighed honestly.** The issue is right that ADA-SCID is the one SCID whose pathology is not confined to the immune system — dATP toxicity gives it sensorineural, skeletal, hepatic and pulmonary features the other five do not have. That is a real argument, and it is the shape of argument that carried cblA (#10260) to a standalone entry.

It does not carry here, for the reason cblA's own PR gave. cblA went standalone because its pathograph *diverges upstream* — MMAA never touches methylmalonyl-CoA — and because the umbrella's subtype axis was response-class, the wrong shape to hold a complementation group. Neither is true here. The umbrella's subtype axis **is** genetic form, exactly the right shape; and the umbrella already carries the purine axis as its own node with the metabolite chemistry bound, converging on the same *Blocked T-lymphocyte development* node the other five reach by different routes. A standalone entry would duplicate that node and its convergent chain to get at the extra-immune features, and those features are better added to the umbrella's phenotype list tagged `subtype: ADA deficiency` — which is a content follow-up, not a reason for a second file.

So: `entry_type: SUBTYPE`, no seventh SCID file.

## What changed

Six `subtype_term` blocks, plus the derived cache rows. Nothing else.

| Subtype | Term | Cross-reference checked |
|---|---|---|
| X-linked SCID | `MONDO:0010315` | OMIM:300400 |
| ADA deficiency | `MONDO:0007064` | OMIM:102700 |
| RAG1/RAG2 deficiency | `MONDO:0011086` | OMIM:601457 |
| IL7R deficiency | `MONDO:0015701` | Orphanet:169154 |
| JAK3 deficiency | `MONDO:0010938` | OMIM:600802 |
| Artemis deficiency | `MONDO:0011225` | OMIM:602450 |

I bound all six rather than just ADA, following #9911: binding half a section is how the next reader ends up believing the unbound half is deliberate.

## Term verification

Every CURIE was resolved through the EBI OLS4 MONDO API before binding — confirmed not obsolete, canonical label copied from the API response rather than typed, and the OMIM/Orphanet xref checked against the gene the umbrella assigns to that subtype.

The RAG1/RAG2 and Artemis pair is the one that needed that check, because both are T-B-NK+ and it would be easy to bind either term to either subtype. `MONDO:0011086`'s own MONDO definition says "due to null mutations in recombination activating gene (RAG) 1 and/or RAG2", and `MONDO:0011225` is the DCLRE1C term (OMIM:602450), so the assignment follows the umbrella's own gene lists.

MONDO nests none of these six under another, so no subtype's coverage depends on another's.

## Stubs retired

Three, confirmed by running `just tidy-stubs` after the edit — before it, none of these registered:

- `Severe_Combined_Immunodeficiency_Autosomal_Recessive_T_Cell-negative_B_Cell-negative_NK_Cell-negative_Due_To_Adenosine_Deaminase_Deficiency.yaml` — the claim
- `Severe_Combined_Immunodeficiency_Autosomal_Recessive_T_Cell-negative_B_Cell-negative_NK_Cell-positive.yaml` (`MONDO:0011086`)
- `T-B_Severe_Combined_Immunodeficiency_Due_To_Gamma_Chain_Deficiency.yaml` (`MONDO:0010315`)

Neither sibling has an open `claim` issue — I checked all 154 open claims for these MONDO IDs before deleting them.

The remaining SCID stubs (CARD11, CARMIL2, CORO1A, DNA-PKcs, IKK2, LAT, LCK) are untouched: those genes are not among the umbrella's six subtypes, so nothing here covers them.

## Why this is one PR and not two

The issue flagged that editing a shared umbrella might belong in its own PR. I kept it together, for the reason #10610 gave: here the shared-file edit *is* the curation, and split out, this claim's PR would contain nothing but a stub deletion with its justification in a different PR. The risk a split guards against is a binding sweep getting entangled with unrelated content edits, and the `kb/` diff here is six `subtype_term` blocks and nothing else.

- **No new phenotypes for the extra-immune ADA features.** Sensorineural deafness, costochondral abnormalities, hepatic dysfunction and pulmonary alveolar proteinosis are genuinely under-curated in the umbrella, but adding them is a content change to a shared file, and this PR is scoped to bindings. Worth a follow-up issue.
- **No deep-research run.** There was no new content to source.
- **No `updated_date`.** Deprecated.
- **Did not commit unrelated `references_cache/` churn.** `just validate` re-fetched full text for four PMIDs it happened to touch; I reverted that, since it is 6,000 lines of diff that has nothing to do with this change.

## Validation

All exit 0: `just validate`, `just validate-terms`, `just validate-disorders` (38/38 snippets verified), `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-snippet-length`, `just check-title-snippets`, `just check-snippet-grading`, `just check-environmental-evidence`, `just check-folded-hyphens`, `just check-stubs`, `just validate-history`.


kevinschaper:
The automated second opinion posted on #10615 after this PR was opened reached the same conclusion independently, which is worth recording since the standalone-vs-subtype call was the substantive question.

It resolved the same six MONDO terms I did (`MONDO:0010315`, `MONDO:0007064`, `MONDO:0011086`, `MONDO:0015701`, `MONDO:0010938`, `MONDO:0011225`), found the same three retirable stubs, and recommended the section-wide binding pass as the lower-risk default.

It also found a paper I had not: PMID:27579027, Whitmore & Gaspar, "Adenosine Deaminase Deficiency - More Than Just an Immunodeficiency" (*Front Immunol* 2016), which argues explicitly that ADA deficiency should be treated as a systemic metabolic disorder rather than a purely immunological one. That is the strongest citable version of the standalone argument, and it is fair to say I decided against standalone without having read it.

Having now checked it, it does not change the call, and the second opinion landed the same way for the same reason: the multi-organ toxicity is real and citable, but it flows from the *same* upstream root — dAdo/dATP accumulation — that drives the SCID phenotype. That is the distinction from cblA (#10260), where the upstream lesion genuinely branched away from the umbrella's chain. What the paper does establish is that the umbrella's ADA phenotype list is incomplete, which is a content gap rather than a nosology problem.

Opened #10621 for that, with the paper as the anchor and the specific phenotypes to add. Deliberately not folded into this PR, which is scoped to bindings and touches no mechanism, phenotype, evidence or prose.

Branch rebased onto `main` since opening, after #10538 landed.

kevinschaper:
Thanks — and the non-blocking point is already filed: #10621 covers the extra-immune ADA features (sensorineural deafness, costochondral abnormalities, hepatic dysfunction, pulmonary alveolar proteinosis), anchored on PMID:27579027, which argues that ADA deficiency should be read as a systemic metabolic disorder rather than a purely immunological one.

Agreed on why it matters rather than just being backlog: the SUBTYPE decision here rests on the umbrella being able to carry those features. If curating them turns out to need its own pathograph branch off the purine node rather than a set of subtype-tagged phenotypes, that is the point at which the standalone question is worth reopening — the issue says so explicitly so the next person does not have to reconstruct the reasoning.

kevinschaper:

PR #10620, opened 2026-09-02
curate: Add Developmental and Epileptic Encephalopathy 81 (MONDO:0032858, DMXL2)

### LINKED ISSUE #10616: Curate developmental and epileptic encephalopathy, 81 (MONDO:0032858)
Curate a dismech entry for **developmental and epileptic encephalopathy, 81 (DEE81)** ([MONDO:0032858](https://monarchinitiative.org/MONDO:0032858)) — biallelic *DMXL2* variants; DMXL2 (rabconnectin-3alpha) is a subunit of the complex that assembles and regulates the vacuolar V-ATPase, so this is an encephalopathy of failed organelle acidification and synaptic vesicle handling.

`MONDO:0032858` is bound nowhere in `kb/`, and DMXL2 appears only in `Aromatase_Excess_Syndrome.yaml` — that is a neighbouring-locus artefact (DMXL2 sits beside CYP19A1), not coverage of this disease. Treat it as a clean start, but check that any DMXL2 citation you use is about the neurological phenotype and not the endocrine one.

Note DMXL2 also causes a dominant deafness (DFNA71) and a polyendocrine-polyneuropathy syndrome. Those are separate entities; gene-frequency preflight will not separate them.

Stub: `stubs/Developmental_And_Epileptic_Encephalopathy_81.yaml`

kevinschaper:
Adds a dismech entry for **developmental and epileptic encephalopathy 81** (MONDO:0032858, biallelic *DMXL2*). The short version: DEE81 is not a channelopathy. DMXL2 is rabconnectin-3alpha, a core subunit of the complex that assembles the vacuolar H+-ATPase, so this is an encephalopathy of failed organelle acidification — and none of that was curated anywhere in the knowledge base.

Closes #10616

## The `entry_type` decision: DISEASE

`MONDO:0032858` was bound nowhere in `kb/`. MONDO puts DEE81 directly under `genetic developmental and epileptic encephalopathy` with no closer curated parent, and the KB already curates numbered DEEs as standalone entries (DEE 50, 83, 84, 89, 116, plus the gene-named ones). It has one reasonably conserved pathograph of its own — a V-ATPase assembly failure, which no other entry models — so it is a Disease, not a grouping and not a subtype of anything curated.

The only entry it could plausibly have been folded into is `Genetic_Developmental_and_Epileptic_Encephalopathy.yaml`, which is the umbrella for the whole class; adding a V-ATPase mechanism there would not have been a subtype so much as a mechanism that belongs in its own file.

## Separating DMXL2's three diseases

This was the real hazard the issue flagged, so here is exactly how it was handled.

DMXL2 causes **three** distinct human phenotypes — recessive DEE81, dominant non-syndromic deafness DFNA71, and a polyendocrine-polyneuropathy syndrome — and no gene-keyed preflight distinguishes them. So the separation was done by reading every candidate rather than by a filter.

The DEE81 clinical literature turns out to be **exactly two papers**, and I checked that by running the search several ways rather than assuming:

| PMID | What it is | Used for |
|---|---|---|
| 31688942 | Esposito et al., *Brain* 2019 — the founding report; three sibling pairs from three unrelated families | genotypes, clinical course, MRI, fibroblast and cultured-neuron experiments |
| 40180340 | *BMJ Case Rep* 2025 — one consanguineous neonate, homozygous nonsense | independent replication, refractory seizures, focal onset |

The two other references cited are gene- and complex-function sources carrying **no disease claim**: PMID:40646309 (the paper that identifies the metazoan RAVE complex as DMXL1/DMXL2 + WDR7 + ROGDI) and PMID:34249946 (the RAVE/rabconnectin-3 review).

**What I found and deliberately did not use:** the mouse-cochlea and hair-cell *Dmxl2* literature (PMID:39147590, PMID:40118164). It is real and it is about DMXL2, but it is written against DFNA71. DEE81 patients *are* deaf, so it would have been easy to reach for — the deafness phenotype is cited from Esposito instead, and the hair-cell papers are not in the entry.

**`Aromatase_Excess_Syndrome` checked, and it is the artefact the issue said it was.** The DMXL2 mentions there are heterozygous 15q21.2 rearrangements that delete DMXL2 exons as a by-product of creating a chimeric *CYP19A1* transcript. Heterozygous, and the phenotype is aromatase excess. Not coverage. That is recorded in the entry's `notes` so nobody has to re-derive it.

## No deep-research run

The entire DEE81 clinical literature is two papers, both read in full abstract, plus two gene-function sources. A DR pass would have added retrieval risk across three same-gene diseases without adding sources. Recording that as a deliberate choice rather than an omission.

## The mechanism

Eleven pathophysiology nodes:

```
Biallelic DMXL2 Loss of Function
  └─ Failed mRAVE-Dependent V-ATPase Assembly
       ├─ Defective Endolysosomal Acidification
       │    └─ Impaired Autophagic Degradation
       │         ├─ Impaired Neurite Elongation and Synapse Formation
       │         │    ├─ Impaired Prenatal Brain Development
       │         │    └─ Neuronal Hyperexcitability and Hypersynchrony   [hypothesis B]
       │         └─ Progressive Brain Atrophy and Leukoencephalopathy
       └─ Impaired Synaptic Vesicle Loading
            └─ Neuronal Hyperexcitability and Hypersynchrony             [hypothesis A]
```

converging on `Neonatal-Onset Refractory Seizures`, `Developmental Encephalopathy` and `Progressive Brain Atrophy and Leukoencephalopathy`. Two nodes conform to `epilepsy_excitation_inhibition_imbalance`.

## The thing I would not assert

**Which arm actually causes the seizures is not known**, and the entry says so structurally rather than in prose. The synaptic-vesicle route and the lysosomal-autophagy route are curated as two competing `mechanistic_hypotheses` groups, both `EMERGING`; the edges from each into hyperexcitability are `INDIRECT_UNKNOWN_INTERMEDIATES` and opt into their own group. A `KNOWLEDGE_GAP` discussion attaches to all three nodes with a proposed experiment that would separate them.

The reason for the caution: the lysosomal arm has direct patient-cell evidence but has never been connected to network excitability, and the synaptic-vesicle arm is stated as a *suggestion* in its source ("which could suggest that..."). No electrophysiological recording has been reported from a DEE81 neuron. Asserting either chain would have overstated both.

A second `KNOWLEDGE_GAP` records the three-disease allelic heterogeneity, so the next curator does not have to rediscover it.

## Evidence grading — one paper, three values, on purpose

PMID:31688942 carries `HUMAN_CLINICAL`, `IN_VITRO` **and** `OTHER` in this file, per sentence, because it reports three kinds of thing:

- family genotypes, clinical course and MRI → `HUMAN_CLINICAL`
- the fibroblast and cultured-neuron experiments → `IN_VITRO`
- the sentence summarising DMXL2's function, and the sentence defining Ohtahara syndrome → `OTHER`, because they are the paper's background, not its results

`check-snippet-grading` keys on the quoted sentence, so this passes — but flagging it explicitly, since a blanket re-grade of this reference would be wrong.

**The Dmxl2-silenced hippocampal neurons are `IN_VITRO`, not `MODEL_ORGANISM`.** They are a dissociated primary culture, not a whole animal. That is also why they sit in `experimental_models` as `PRIMARY_CELL_CULTURE` and why the entry has **no** `animal_models` section: no whole-organism DEE81 model has been reported.

## Small calls worth naming

- **Fatal outcome is in `progression:`, not `phenotypes:`.** Five of six patients died within nine years. HPO files death in infancy/childhood under clinical course, not under phenotypic abnormality.
- **Frequency is asserted only where the source gives the whole cohort.** `OBLIGATE` on seizures, profound developmental delay, thin corpus callosum and hypomyelination ("in all"). Cerebral atrophy and leukoencephalopathy carry **no** frequency, because their denominator is the three patients who had follow-up scans — not the cohort. Focal-onset seizure carries none because it rests on a single case. Each says so in `notes`.
- **Sensorineural hearing impairment carries a caveat.** The source says "deafness" without an audiological subtype per patient. The `notes` say what the underlying claim actually is.
- **`prevalence` is `CASES_IN_LITERATURE` / `NOT_YET_DOCUMENTED`** with no numeric rate. Seven patients total; nobody has estimated a population rate.
- **No `therapeutic_agent` on the antiseizure treatment.** The only evidence is that multiple antiepileptics were tried and failed. Naming an agent would imply a recommendation the sources do not make.

## Term verification

Every CURIE resolved through OLS4 with the label copied from the API response, not from memory. **One was rejected during that check:** `HP:0007260`, which I had reached for as a hypomyelination term, is actually **Type II lissencephaly**. The correct term is `HP:0006808` Cerebral hypomyelination. Recording the near-miss because it is the exact failure mode the issue warned about.

## Validation

All exit 0 on the rebased branch: `just validate` (**48/48 snippets verified**), `just validate-terms`, `just validate-disorders`, `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-snippet-length`, `just check-title-snippets`, `just check-snippet-grading`, `just check-environmental-evidence`, `just check-folded-hyphens`, `just check-stubs`, `just validate-history`.

One note for whoever reads a red CI here: `check-snippet-grading` failed once before the rebase, on a finding in `Appendiceal_Neoplasm.yaml` — a file this diff never touches. It was baseline drift from #10538 landing on `main` mid-run. Rebasing cleared it; I did not "fix" the innocent file.

Only the four references actually cited are committed; five more were fetched during screening and discarded rather than committed as unused cache.


kevinschaper:
The automated second opinion posted on #10616 after this PR was opened agrees on `entry_type: DISEASE` and on the same two clinical papers (PMID:31688942, PMID:40180340), so the core calls converge. Two of its suggestions I did **not** take, and the reasons are worth stating rather than leaving as silent omissions.

**PMID:39147590** (*J Neurosci* 2024, hair-cell Dmxl2 knockout causing synaptopathy) was offered as mechanistic support for the synaptic-vesicle arm. I found it during screening and left it out on purpose: the paper is written against DFNA71, the *dominant* non-syndromic deafness, and its own framing says so. DEE81 patients are deaf, which makes it tempting, but citing a DFNA71 mechanism paper for a DEE81 mechanism claim is precisely the same-gene confusion the issue asked to guard against. The deafness phenotype here is cited from Esposito instead. If someone wants the vesicle-recycling mechanism grounded, the honest route is a mechanism node framed as DMXL2 gene function with the model system named, not a disease claim.

**PMID:37371135** (photoreceptor ribbon synapse localization) is real and interesting but there is no retinal phenotype in either DEE81 report, so a node built on it would have nothing in this disease to attach to.

**PMID:40646309** (the mRAVE complex paper) I did take, and it is doing exactly the job suggested — it is the source for the `Failed mRAVE-Dependent V-ATPase Assembly` node and for the three functions that branch out of it.

One thing the second opinion states more confidently than the sources do: it gives the mechanism as a single chain from lysosomal acidification through to the encephalopathy. The sources do not support that. The lysosomal/autophagic arm has direct patient-cell evidence but has never been connected to network excitability, and the synaptic-vesicle arm appears in its source as a suggestion ("which could suggest that..."). No electrophysiology has been reported from a DEE81 neuron. That is why the two routes are curated here as competing `EMERGING` hypothesis groups with `INDIRECT_UNKNOWN_INTERMEDIATES` edges, rather than as one asserted chain.

kevinschaper:
Pushed one self-review fix before the reviewer got to it.

Four phenotypes carry `frequency: OBLIGATE`, and I had left the denominator implicit. `Seizure` was the weak one: it was marked OBLIGATE but its only evidence was the single BMJ case, which cannot establish universality. It now also cites the Esposito sentence covering all six patients of the founding cohort, and each of the four says in `notes` that "in all" means those six patients — not a larger series that does not exist.

`Profound global developmental delay` additionally records why the seventh patient is not in that denominator: he was a neonate at publication, so no developmental outcome was reported.

`just validate-disorders` re-run, 49/49 snippets verified, all gates exit 0.

kevinschaper:
All four blocking items addressed. One sub-item of the first declined, with reasoning below. Branch also rebased onto current `main`.

## 1. Pathograph joined to the clinical picture

You were right that the mechanism chain stopped before it reached anything clinical. Added the joins you listed, using bare node names:

- `Neuronal Hyperexcitability and Hypersynchrony` → EEG with Burst Suppression
- `Neonatal-Onset Refractory Seizures` → Seizure, Focal-Onset Seizure, Epileptic Encephalopathy
- `Developmental Encephalopathy` → Profound Global Developmental Delay
- `Impaired Prenatal Brain Development` → Thin Corpus Callosum, Cerebral Hypomyelination
- `Progressive Brain Atrophy and Leukoencephalopathy` → Cerebral Atrophy, Leukoencephalopathy

Nine of twelve phenotypes now have an incoming edge. **Three are deliberately left unattached**: hearing impairment, peripheral neuropathy and dysmorphic features. They are well attested clinically, but no source explains how DMXL2 loss produces them, and DEE81 has no extraneural mechanism work at all. Drawing an edge from any existing node would assert a mechanism nobody has shown. That is recorded in `notes` so it reads as a decision rather than an oversight.

**Declined: `target_mechanisms` on Antiseizure Medication.** `TreatmentEffectEnum` offers INHIBITS, ACTIVATES, MODULATES, BYPASSES and RESTORES. Every one of them claims an effect on `Neuronal Hyperexcitability and Hypersynchrony` that the only cited evidence contradicts — the quote is "Despite initial treatment with multiple antiepileptics, seizures persisted." Antiseizure drugs do act on excitability as a class, but here that is the intent, not the outcome, and the enum has no value for an intended target that failed. Per CLAUDE.md, when a controlled value would misstate the claim it is left absent with the reason in `notes`, which is what I did.

## 2. `diagnosis:` section — you were right and I had misread the sibling entry

I had looked at `Developmental_And_Epileptic_Encephalopathy_89.yaml` with a grep that did not distinguish sections and concluded it kept Molecular Genetic Testing under `treatments:`. It does not; it is in `diagnosis:`. Moved, and added Brain MRI and Electroencephalography while there — both snippet-supported, and the MRI one carries the point that early imaging in DEE81 is *not* normal, which is diagnostically useful in the other direction.

## 3. Deep research — the requirement is real, and the run changed the entry

I checked before complying: `.claude/skills/dismech-pr-review/SKILL.md` line 91 does say a new entry must carry at least one deep-research artifact. My rationale for skipping it was, as you said, the claim the run would test. So I ran it.

`falcon`, 601s, 19 citations. `just preflight-dr` **PASS** against MONDO:0032858 — DMXL2 at 49 mentions, no competing gene. More usefully, the report is DEE81-specific and separates the three DMXL2 diseases explicitly rather than blending them, which was the risk.

It was worth running. Four things came out of the cross-check, all now in the entry:

| Finding | What it added |
|---|---|
| **Falace et al. 2024** (PMID:39273013), a V-ATPase-in-brain review co-authored by the founding paper's second author | The sentence tying DMXL2 specifically to *incorporating* the V-ATPase into synaptic vesicles via its Rav1p domain — which is what makes `Impaired Synaptic Vesicle Loading` a DMXL2 consequence rather than a general statement about the pump |
| **Mouse genetics** | An `animal_models:` entry. Complete Dmxl2 loss is embryonic lethal in mice, so the genotype matching the patients cannot exist in an animal; the viable heterozygote shows corpus callosum dysplasia and macrocephaly. Curated as `PARTIALLY_RECAPITULATES` / `LOW` fidelity with the zygosity mismatch stated |
| **The lethal null** | A `HUMAN_MODEL_MISMATCH` discussion: every mechanistic result in this disease comes from a fibroblast or a dissociated culture, and neither can produce a seizure, an EEG, myelin or an atrophy. That is *why* the edges into hyperexcitability are `INDIRECT_UNKNOWN_INTERMEDIATES` — not that nobody looked, but that the available systems cannot look |
| **An eighth patient** (PMID:30237576, Maddirevula et al.) | A `KNOWLEDGE_GAP`. A homozygous DMXL2 truncating variant in a child with ID, epilepsy, macrocephaly and dysmorphism. The primary report classifies it as "phenotypic expansion or potentially distinct allelic disorders"; Falace 2024 counts it as a DMXL2 DEE case. Both readings are defensible and they give different patient counts, so `prevalence` still counts seven and points at the discussion |

**What the report surfaced that I did not curate, and why.** The report reads the founding paper's *full text*; `references_cache/PMID_31688942.md` is abstract-only, and Brain 2019 is not deposited in PMC — I checked the PMC ID converter, which returns "Identifier not found in PMC". So there is no route to quotable full text. The uncurated findings are: per-patient seizure semiology (tonic predominant, some myoclonic), hypotonia and quadriparesis in 6/6, simplified gyral pattern in 2/6, the deafness specified as sensorineural in 6/6, and the rescue of the fibroblast lysosomal phenotype by wild-type DMXL2. Each is real and each is listed in `notes` for whoever can quote it later. None is asserted here on the strength of a secondary summary.

## 4. HP:0000407 → HP:0000365

Agreed, and this was the sharpest finding. The abstract says only "deafness", and my justification for the narrower term appealed to the DFNA71 hair-cell literature that the rest of the entry deliberately refuses to cite for DEE81 claims — taking a binding from a source about a different disease is exactly the error the narrow term would encode. Rewritten to say the term can be specialised when a DEE81 report gives audiological detail.

Worth noting: the DR report *does* say sensorineural, 6/6, from the full text. So the specialisation is probably coming — but it is not quotable, so it is not asserted.

## Non-blocking suggestions, all taken

- `classifications:` — NEUROLOGIC and GENETICS_ENVIRONMENT_DISEASE.
- `differential_diagnoses:` — an "other genetic causes of Ohtahara syndrome" entry citing the aetiological-heterogeneity sentence, plus STXBP1 encephalopathy. I discriminated STXBP1 on **zygosity** (heterozygous haploinsufficiency vs recessive, with an added quote) rather than on imaging, because I could not source a claim about STXBP1's early MRI and did not want to assert one. The specific gene list you suggested is named in `notes` as KB cross-references rather than as evidence-backed differentials, since the DEE81 sources do not enumerate them.
- `CL:0000540` added to `Neonatal-Onset Refractory Seizures`, matching the module node it conforms to.

## Validation

All exit 0 on the rebased head: `just validate` (**68/68 snippets verified**, up from 49), `validate-terms`, `validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence`, `check-stubs`, `validate-history`. 13,484 targeted pytest tests pass, including `test_model_readout_targets_match_link` for the new animal-model readout.

One note on the DR run's side effects: its citation-validation pass created five `references_cache/DOI_*.md` files for report citations the entry does not cite. I removed those rather than committing them; only PMID:39273013 and PMID:30237576 are added.

A second history record covers this review response.

Issue #1571, opened 2026-04-20
CI: simple scope-violation check for curation PRs touching infrastructure files


cmungall:
## Replaces the approach in PR #1435

PR #1435 proposed a 596-line Python audit module. This issue proposes a simpler alternative: ~20 lines of bash in CI that catches the same class of bug.

## The problem (two confirmed incidents)

Curation PRs accidentally revert infrastructure files during conflict resolution:
- PR #1166 reverted boss skill + CLAUDE.md changes from #1150 (see #1430)
- PR #1425 reverted OpenScientist Worker config from #1442 (see #1430 comment)

Both times: agent rebased a stale curation branch, took its own version of infrastructure files instead of main's, and the silent reversion slipped past both CI and claude-review.

## Proposed: scope-violation check

Add a CI step that checks: **if a PR touches `kb/` (curation), does it ALSO touch files outside the expected curation scope?**

Expected curation scope (files a curation PR may legitimately modify):
- `kb/` — the disease YAML
- `cache/` — validation caches (regenerated)
- `references_cache/` — PMID caches (fetched)
- `research/` — deep research artifacts
- `pages/` — rendered HTML (derived)

If a curation PR also modifies files in `src/`, `proxy/`, `.github/`, `conf/`, `tests/`, `scripts/`, or repo-root config files — that's a scope violation and should block the PR unless explicitly overridden (e.g., via a `scope-override` label).

## Implementation sketch

```bash
# In .github/workflows/main.yaml, as a new step:
- name: Check curation PR scope
  if: github.event_name == 'pull_request'
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    CHANGED=$(gh pr diff ${{ github.event.pull_request.number }} --name-only)
    if echo "$CHANGED" | grep -q '^kb/'; then
      INFRA=$(echo "$CHANGED" | grep -vE '^(kb/|cache/|references_cache/|research/|pages/|dashboard/)' || true)
      if [ -n "$INFRA" ]; then
        echo "::error::Curation PR modifies files outside curation scope:"
        echo "$INFRA"
        echo "If intentional, add the 'scope-override' label to this PR."
        LABELS=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '[.labels[].name] | join(",")')
        if echo "$LABELS" | grep -q 'scope-override'; then
          echo "scope-override label found — proceeding."
        else
          exit 1
        fi
      fi
    fi
```

[...]

PR #10629, opened 2026-09-02
Derive cancer cell of origin from the pathograph; no new slot

cmungall:
A cancer entry now says which cell it arose from, and it says it without a new field. The cell of origin is derived from the entry's own pathograph: find the node where the transforming event happened, read that node's `cell_types`. 124 of 252 somatic cancer entries now answer the question, up from 30.

Deriving **more than one** cell of origin is the useful signal. It means the entry is probably several diseases wearing one entry's clothes, and eight entries currently report it.

## How an entry says where it starts

Two markers, both structured claims that already existed in the schema:

| Marker | Means |
|---|---|
| `genetic_context.variant_origin: SOMATIC` on a pathophysiology node | the transforming lesion happened here |
| an `environmental[].influences_mechanisms` link with `environmental_effect: TRIGGERS` | non-mutational initiation — HPV, H. pylori, asbestos, UV |

The exposure marker only speaks when no lesion is recorded. That is about meaning, not confidence: the cell of origin is the cell the transforming event occurred in, so once the entry names that event, the exposure is upstream context. Pancreatic cancer is where it shows — chronic pancreatitis really does trigger its inflammation node, but that node binds macrophage and stellate cell, while the disease arises in the acinar cell named on the KRAS lesion.

```yaml
pathophysiology:
- name: KRAS Oncogene Activation
  genetic_context:
    variant_origin: SOMATIC          # <- this makes it the origin node
  cell_types:
  - preferred_term: pancreatic acinar cell
    term: {id: CL:0002064, label: pancreatic acinar cell}
```

A virally driven mechanism is **not** a lesion: HPV E6/E7 leaves no host variant for `variant_origin` to describe, which is the rule `CLAUDE.md` already applies to HTLV-1 Tax. Those entries record the initiating exposure instead.

## Why no `cell_of_origin:` slot

It would restate what the pathograph already says, and the two would drift. It would also sit at the disease level, unable to tell a subtype's origin from its parent's, and it would invite filling in from the disease name rather than from the mechanism. Deriving it keeps the claim attached to the node that holds its evidence, and makes an unmarked entry visible as unmarked rather than as an empty field.

## What was tried and dropped

The first version had a third rule that read an initiating-sounding `role` string, plus a fallback chain between rules. Both existed to guess at entries that had not recorded their origin, and both guessed wrong — the role rule is what derived macrophage and pancreatic stellate cell as the cell of origin of pancreatic cancer. The records were marked instead and the heuristics deleted.

## The backfill

`scripts/backfill_cancer_origin.py` marked 111 entries and invents nothing: it marks a node only when the node's own **name** already states a lesion, so the edit transcribes what the curator wrote. Pathway states, germline variants, viral oncoprotein mechanisms, microenvironment nodes and acquired-resistance nodes are all excluded. `--bind-single-cell` copies the entry's cell type onto the lesion node when the entry names exactly one; entries naming several are left for a curator.

## Numbers

| | Before | After |
|---|---|---|
| Cell of origin derived | 30 | 124 |
| Origin marked but no cell bound | 13 | 0 |
| Multi-origin findings | 3, one spurious | 8, all real |
| No origin recorded | 220 | 128 |

`ORIGIN_WITHOUT_CELL` at zero makes that class ready to become a hard gate. The entries with no origin recorded name no genetic lesion, or name one with no cell type anywhere in the entry — both need a curator, not a script, so they are reported as unmarked rather than guessed at. (Counts are a snapshot against the currently merged `main` and drift upward as `main` adds cancer entries; they were last refreshed at 252 assessed / 128 no-origin.)

## NCIT

NCIT asserts its own cell of origin per disease inside its logical definitions — DLBCL is *Mature B-Lymphocyte* becoming *Neoplastic Large B-Lymphocyte*. `NCIT:R104`, `R112` and `R105` (the Abnormal Cell branch) are added to the ontology-edge manifest as a cross-check and evidence source. They live in semsql's `edge` table rather than `statements`, because NCIT states them through equivalent-class restrictions, so `OntologyEdgeSource` now reads both. `cell_types` stays CL-only.

## What review turned up

Four defects, all found by review, all of a kind worth recording.

**Two tooling bugs, the same mistake twice.** `_terms()` was changed to report a binding's `preferred_term` rather than its ontology label, so the census would stop asserting more than a hedged record does. But `backfill_cancer_origin.py` imports that function and **writes YAML from it**, which made the change unsafe there twice over: a borrowed binding would have put the `preferred_term` into `term.label`, which must match the ontology exactly; and the value is spliced into YAML as text, where curator free text like `EMG: myopathic abnormalities` (seven such values are already in the KB) makes a nested mapping unquoted. `_terms()` now takes `display=`, and PyYAML decides the quoting. Neither reached a committed entry — the backfill ran before either change — and both have regression tests asserting the round trip.

**PDAC was deriving the wrong cell of origin.** The KRAS node carries this entry's somatic lesion, so it is the node the origin derives from, and it bound `CL:0002079 pancreatic ductal cell`. The entry says the opposite in as many words: its Acinar-to-Ductal Metaplasia node places the origin in the acinar compartment *"rather than in pre-existing ducts despite the tumour's ductal histology"*, citing lineage tracing where K-Ras(G12V) driven in the acinar/centroacinar lineage yields PanINs and invasive PDA (PMID:17349585). The derivation was reproducing precisely the naive error the entry was curated to warn against — reading the tumour's histology as its origin — on the flagship worked example. Rebound to `CL:0002064 pancreatic acinar cell`. The ductal binding predates this branch; marking the node as the origin is what turned a latent inconsistency into a published derived claim, which is the check earning its keep on its own exemplar.

The human cell of origin of PDAC remains debated, and binding both cells would make it a `MULTI_ORIGIN_CELL` finding under the "genuinely unsettled origin" case the docs already describe. That needs the ductal-origin position curated with its own evidence, which is curation work rather than a correction, so it is left as a follow-up. This change makes the record self-consistent; it does not settle the field.

**Two schema cross-references still pointed at 3b.** When this branch's decision was renumbered 3b → 3d during the earlier merge, the sweep updated `docs/cancer-curation-sop.md` but missed two slot `comments` in `dismech.yaml`. They resolved to main's unrelated *"3b. Model-to-target scale gap"* — which is also a "derived, not asserted" decision, so the pointer landed somewhere that read plausibly on-topic. One of the two is the curator-facing answer to "why is there no `cell_of_origin:` slot", which makes it the pointer that most needs to be right. History records under `history/` still say 3b and are left alone: they are append-only records of sessions that ran while the section really was 3b.

## The merge from main

`main` added `### 3b. Model-to-target scale gap` and `### 3c. Model divergence is typed` at the same anchor in `docs/explanation/design-decisions.md` where this branch added its own `3b`. Both were kept and this branch's section renumbered to **3d**, since main's is already merged and already cited by `docs/explanation/model-credibility.md`. Resolving on the conflict text alone would have left two sections numbered 3b and silently redirected every cross-reference.

A second merge from main followed, and it surfaced a regression of exactly the kind this check exists to catch. `main` had rewritten `Ewing_Sarcoma`, adding a `Permissive Progenitor Cell State` node that carries both candidate lineages and says the human cell of origin is unresolved. This branch's `SOMATIC` marker sits on the `EWS-FLI1 Fusion Oncogene` node, which bound no cell — so `ORIGIN_WITHOUT_CELL` went from 0 to 1, contradicting the claim that the class is ready to become a hard gate.

Both candidate cells are now bound on the lesion node, with a note saying why the pair is deliberate. This asserts nothing new: both CL terms and the unresolved-origin position are already in the entry, on a node whose own description says so. Removing the `SOMATIC` marker would have been wrong, since the fusion really is the transforming lesion; binding a single cell would have asserted a resolution the entry explicitly declines to make. Ewing now reads as `MULTI_ORIGIN_CELL`, which is the class meant for a genuinely unsettled origin — the same shape as `Melanoma_in_Congenital_Melanocytic_Nevus`.

Verified after both merges that no main content was reverted: against `main` the KB footprint is +551/−3 with the three deletions being the PDAC rebinding, the infra footprint is +2435/−1 with the one deletion being the `qc:` recipe line, and zero `variant_origin` marks were dropped.

## Checks

All changed entries pass schema, term and reference validation. The whole-KB duplicate-key, enum-value, entity-ref and causal-target gates are clean, `validate-history-all` passes, and the three reference-snippet baseline gates (`check-title-snippets`, `check-snippet-grading`, `check-snippet-length`) pass against the refreshed `main` — they had been reporting entries this branch never touched, purely because the branch sat behind main's fixes to them. 35 tests across three modules. Every changed entry has a history record.

Recorded as design decision 3d, with `docs/cancer-cell-of-origin.md`.

https://claude.ai/code/session_01HFumBTrcxA5P7xnUb4hiuP

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -925,7 +925,7 @@ never gated. It means a grouping wearing a Disease entry's clothes
 cell-of-origin subtypes (DLBCL's GCB/ABC — remedy `has_subtypes`), or an
 unsettled origin (melanoma in congenital melanocytic nevus — remedy a note).
 
-The check is **advisory** and runs inside `just qc`, exiting 0 because 122 of 245
+The check is **advisory** and runs inside `just qc`, exiting 0 because 128 of 252
 assessed neoplasm entries are still unmarked; `--fail-on <CLASS>` or `--strict`
 gates when you want one. `ORIGIN_WITHOUT_CELL` is currently at zero — every entry
 that marks an origin binds a cell there — so that class is ready to become a real
--- a/docs/cancer-cell-of-origin.md
+++ b/docs/cancer-cell-of-origin.md
@@ -144,10 +144,11 @@ Always re-validate afterwards — `just validate-disorders` on the changed files
        remedy is `has_subtypes`;
     3. an origin the literature has not settled —
        `Melanoma_in_Congenital_Melanocytic_Nevus` names melanocyte and neural
-       crest cell on one node. Naming both is the honest answer, and the remedy
-       is a note saying so.
+       crest cell on one node, and `Ewing_Sarcoma` names mesenchymal stem cell
+       and migratory neural crest cell. Naming both is the honest answer, and
+       the remedy is a note saying so.
 
-  The seven current findings are all of these kinds, which is the point: the
+  The eight current findings are all of these kinds, which is the point: the
   list is short enough to work through, and every row is a real modeling
   question. `Gastrointestinal_Lymphoma` is the clearest — B cell *and*
   intraepithelial lymphocyte, because the entry covers both MALT lymphoma and
@@ -185,7 +186,7 @@ colon-space would make a nested mapping instead of a string.
   since the marking is already there.
 
 `NO_ORIGIN`
-: Neither marker is present. This is the remaining backlog: 122 of 245 assessed
+: Neither marker is present. This is the remaining backlog: 128 of 252 assessed
   neoplasm entries, down from 220 before the backfill. Most are entries whose
   pathograph names no genetic lesion at all, or names one with no cell type
   anywhere in the entry — both need a curator, not a script.
@@ -198,7 +199,7 @@ just check-cancer-origin --fail-on ORIGIN_WITHOUT_CELL # gate one class
 just list-cancer-origin
 ```
 
-It runs inside `just qc` as a report and exits 0 by default, because 122 entries
+It runs inside `just qc` as a report and exits 0 by default, because 128 entries
 are still unmarked. `ORIGIN_WITHOUT_CELL` is currently at **zero**, though, so
 that class is a candidate for a real gate: every entry that marks an origin
 binds a cell there, and it would be a regression for one to stop doing so.
--- a/docs/explanation/design-decisions.md
+++ b/docs/explanation/design-decisions.md
@@ -368,9 +368,9 @@ class is a curator's judgement, not a computed match.
 
 **Enacted.** `scripts/check_cancer_origin.py` (`just check-cancer-origin`, advisory,
 inside `just qc`), `scripts/backfill_cancer_origin.py`, `tests/test_cancer_origin.py`, and
-105 entries marked. 123 of 245 assessed neoplasm entries now derive a cell of origin, and
+105 entries marked. 124 of 252 assessed neoplasm entries now derive a cell of origin, and
 `ORIGIN_WITHOUT_CELL` is at zero — every entry that marks an origin binds a cell there,
-which makes that class ready to become a hard gate. **Still open:** the remaining 122
+which makes that class ready to become a hard gate. **Still open:** the remaining 128
 unmarked entries, whose pathographs name no genetic lesion, or name one with no cell type
 anywhere in the entry — both need a curator rather than a script; and whether CL should
 gain lineage-specific transformed-cell terms, or `CellTypeTerm` gain `NCIT:C12913` as a
--- a/history/disorders/Ewing_Sarcoma/2026-09-07T182720Z-claude-code-e14789.yaml
+++ b/history/disorders/Ewing_Sarcoma/2026-09-07T182720Z-claude-code-e14789.yaml
@@ -1,0 +1,31 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Ewing_Sarcoma
+  path: kb/disorders/Ewing_Sarcoma.yaml
+session:
+  id: 2026-09-07T182720Z-claude-code-e14789
+  timestamp: '2026-09-07T18:27:20Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10629
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: Bind both candidate cells of origin to the EWS-FLI1 lesion node
+  details: |
+    Merging main into this branch surfaced a regression the cell-of-origin derivation is designed to catch. Main had rewritten this entry and added a dedicated 'Permissive Progenitor Cell State' node carrying both candidate lineages (early mesenchymal stem cell CL:0000134 and migratory neural crest cell CL:0000333, mechanism_confidence PROVISIONAL). The branch's own marking sits on 'EWS-FLI1 Fusion Oncogene' -- genetic_context.variant_origin: SOMATIC -- and that node bound no cell, so check-cancer-origin reported ORIGIN_WITHOUT_CELL and the repo-wide count for that class went from 0 to 1.
+
+    Bound both candidate cells onto the lesion node and added a notes: line explaining why the pair is deliberate. This makes no new biological claim: both CL terms, both preferred_term wordings, and the unresolved-origin position are already asserted by this entry, on the node whose description says the human cell of origin remains unresolved. The fusion is genuinely the transforming lesion, so removing the SOMATIC marker would have been the wrong repair; binding a single cell would have asserted a resolution the entry explicitly declines to make.
+
+    The entry now derives two cells of origin and is reported as MULTI_ORIGIN_CELL, which is the class intended for a genuinely unsettled origin -- the same shape as Melanoma_in_Congenital_Melanocytic_Nevus. That class is advisory and never gates. ORIGIN_WITHOUT_CELL returns to 0, so it remains a candidate for a hard gate.
+
+    Validated with linkml-validate (Disease), linkml-term-validator --labels, check_duplicate_yaml_keys, check_enum_values, check_entity_refs, check_causal_targets, and 35/35 tests in test_cancer_origin, test_backfill_cancer_origin and test_ontology_edge_relations.
--- a/kb/disorders/Ewing_Sarcoma.yaml
+++ b/kb/disorders/Ewing_Sarcoma.yaml
@@ -59,6 +59,22 @@ pathophysiology:
     term:
       id: hgnc:3749
       label: FLI1
+  cell_types:
+  - preferred_term: early mesenchymal stem cell
+    term:
+      id: CL:0000134
+      label: mesenchymal stem cell
+  - preferred_term: migratory neural crest cell
+    term:
+      id: CL:0000333
+      label: migratory neural crest cell
+  notes: >-
+    Both candidate cells are bound here because the fusion is the transforming
+    lesion but the lineage it arises in is unresolved; see the Permissive
+    Progenitor Cell State node, which carries the evidence for each candidate.
+    The pair is deliberate rather than a hedge to be tidied away: it makes the
+    derived cell of origin read as two candidates, which is what this entry
+    claims.
   biological_processes:
   - preferred_term: positive regulation of transcription by RNA polymerase II
     modifier: ABNORMAL


cmungall:
The HPV finding is correct and was the important one. Four entries had `variant_origin: SOMATIC` on E6/E7 oncoprotein nodes where no host variant exists, and because a recorded lesion suppresses the exposure rule, the four entries that should have demonstrated that rule were the four where it was dead. Fixed in 44b9ef006.

**The four HPV entries.** Removed the marks, then recorded the origin where it belongs: each entry gets an environmental exposure bound to `ECTO:2000053` with an `influences_mechanisms` link carrying `environmental_effect: TRIGGERS` onto its initiating node. `Anal_Canal_Carcinoma` already had the link, so it just needed the spurious mark gone. All four now derive through `ENVIRONMENTAL_TRIGGER`:

| Entry | Origin node | Cell |
|---|---|---|
| Anal_Canal_Carcinoma | HPV Infection of Anal Squamous Epithelium | stratified squamous epithelial cell of anal canal |
| Cervical_Cancer | E6 Oncoprotein-Mediated p53 Degradation | epithelial cell of cervix |
| Cervical_Squamous_Cell_Carcinoma | Persistent High-Risk HPV Infection | squamous epithelial cell |
| HPV_Positive_Head_and_Neck_Cancer | E6-Mediated p53 Degradation | squamous epithelial cell |

Evidence on each exposure and each link reuses a snippet already verified in that entry. `check-environmental-evidence` passes. The backfill script now excludes viral mechanisms outright, and the rule is written into `CLAUDE.md` and the docs next to the existing HTLV-1 Tax precedent.

**Epithelioid_Sarcoma.** Marking moved to `SMARCB1 (INI1) Loss`. Two regex fixes behind it: gene loss written with a parenthetical alias now matches, and bare "hypermethylation" no longer marks a node — only promoter-level silencing does.

**Gastrointestinal_Lymphoma.** Both lineages are now bound, so the entry reports `MULTI_ORIGIN_CELL` rather than hiding the split question behind a single wrong binding. You were right that this is the entry in the tranche that most deserved the signal.

**`apply_proposal()` silent drop.** Real bug, fixed, with a regression test covering a node that already carries a `genetic_context` plus a borrowed cell.

**The rest.** Stale `role` comment in the schema rewritten to describe the two-rule design. `--format list` is now wired to `render_list()` and tested. `find_origins()`'s docstring rewritten to say precedence, and to say explicitly that a lesion mark suppresses the trigger rule. `_edge_table_rows()` now returns what it read and warns instead of discarding chunks, and skips blank nodes. The NCIT evidence example in the docs now says to run `just ncit-edges-rebuild` first, since no `NCIT_*.md` carries those predicates yet.

**Two suggestions I did not take.**

The borrowed-cell spot-check: I looked at `Acute_Erythroid_Leukemia` and left `proerythroblast` on the TP53 node. WHO 2022 treats pure erythroid leukemia as erythroid-lineage, and it is the only cell the entry names, so the binding is the entry's own claim rather than the script's. It is arguable, as you say, and a curator who disagrees should move it — but I would rather leave a defensible binding than strip it and reopen `ORIGIN_WITHOUT_CELL`.

The `model` field on history records: I added `prs: [10629]` to all 113, but left the model off. This session operates under a rule against embedding a model identifier in repository artifacts, which conflicts with the `CLAUDE.md` guidance here. Worth someone reconciling those two.

Counts after the corrections: 123 of 245 assessed entries derive a cell of origin, `ORIGIN_WITHOUT_CELL` stays at zero, `MULTI_ORIGIN_CELL` is six. The PR body has been rewritten — it described the superseded three-rule design.

On the execution caveat: `just validate-disorders`, `just qc`'s offline gates, and all 29 tests were run here and pass.

---

cmungall:
`test (3.13)` failed on the curation-scope guard, not on a test — the guard exits before the suite runs. This PR touches `kb/` and also `scripts/`, `tests/`, `src/`, `docs/` and `CLAUDE.md`, which the guard blocks unless the PR carries `scope-override`.

I added the label, after checking the specific thing the guard is for.

That guard was added by issue #1571 to catch curation PRs that **silently revert infrastructure files during conflict resolution** — two real incidents where an agent rebased a stale curation branch, kept its own copy of an infra file, and reverted someone's work without anyone noticing. That is worth being paranoid about here, because I did just resolve a conflict with main in the previous commit. So I checked every infra file on the branch against main:

| File | Also changed on main since the merge base? |
|---|---|
| `project.justfile` | yes — see below |
| the other 14 | no |

For `project.justfile`, main's change (NDEx comments, public → private, in #10558) landed **after** the merge, so the branch is simply behind on it rather than reverting it. Nothing on this branch reverses anything from main. The one conflict I did resolve, in `Cervical_Cancer.yaml`, kept both sides — main's corrected evidence snippet and this branch's new environmental block.

The mixed scope is inherent to the change rather than accidental: the PR adds a derivation (`scripts/check_cancer_origin.py`), the schema comment and docs describing it, tests, and the 111 KB entries that the derivation reads. Splitting it would put the tooling in one PR and the records it is derived from in another, with neither half checkable on its own.

If you would rather see it split, say so and I will — but the label is the mechanism #1571 designed for exactly this case, and it is visible in the PR UI, which was the design point.

Re-running the check now.

---

cmungall:
Two pushes since the approval, so it will need re-running.

**A real CI failure first.** `test (3.13)` went red on `validate-history-all`: `10629 is not of type 'string' in /links/prs/0`, across all 113 records. My mistake — `links.prs` has `range: uri`, and when I added the PR link in response to the review I wrote the bare number instead of the URL that `just new-history --pr` emits. The six correction records already had it right, which is what makes the slip obvious in hindsight. Fixed in 29e673cb0; `validate-history-all` now passes over all 8286 records.

Worth noting the sequencing: this step runs after lint in the same job, so it only became reachable once the ruff fixes landed. It was never a new regression, just the next thing in line.

**Then the suggestions, three of four taken** (4b7284122):

*Borrowed cell bindings.* Right on all three, and right about why they matter — each existed because a heuristic put it there, not because anyone judged that lineage for that node. Removed from the two E7 nodes and the EZH2/PRC2 node. The derived cell of origin is unchanged in all three, since it now reads from a different node. The Epithelioid_Sarcoma catch was the sharpest: its lesion node hedges the binding as `preferred_term: mesenchymal cell of uncertain differentiation` against `label: mesenchymal stem cell`, while the borrowed copy asserted the same CURIE flatly.

*Lung_Carcinoma.* Also right, and backwards in exactly the way you said — the broad pool is where the split question presses hardest, and it was the one reporting a settled single origin while its own subtype entry reported two. `CL:0002063` alveolar type 2 cell is now bound alongside bronchial epithelial on the driver node, matching how `Non-Small_Cell_Lung_Cancer` binds its equivalent. The two entries agree now, and `MULTI_ORIGIN_CELL` is 7.

*Duplicated rebuild block.* Dropped the second copy; the first keeps the "rebuild before citing" warning.

**Not taken: the `Cervical_Cancer` HPV snippet.** You are right that `PMID:41503346` is a prevalence study restating an established fact in its introduction rather than reporting it. The snippet is exact and does support the claim, so it passes, but swapping in a primary citation means fetching and verifying a new reference — a curation task rather than a correction to this PR. Leaving it visible here rather than doing it quietly.

**On the `model` field**, agreed it is a standing contradiction rather than something to settle in a PR, and thank you for flagging it rather than pushing it either way.

Counts after all of this: 245 assessed, 123 derive a cell of origin, `ORIGIN_WITHOUT_CELL` still zero, `MULTI_ORIGIN_CELL` 7. 29 tests pass, `uvx ruff check .` clean, the whole-KB gates clean, `validate-disorders` green on the four entries touched.

---

cmungall:
Good catch on `check-title-snippets`, and you diagnosed it exactly right — it was staleness, not breakage. Merged main in (`a555f53bf`), which resolves it.

The merge shows it cleanly: the grandfathered baseline went from **339 occurrences of 277 snippets** to **337 of 275**. Two fewer, which is precisely the two entries main fixed. The branch had both the unfixed entries and the baseline still covering them, so it was internally consistent and green in isolation — which is why I could not reproduce a failure before merging. Against main's newer baseline it would go red, and that gap is what you saw.

The branch was 19 commits behind. No conflicts this time.

On the merged tree: `check-title-snippets`, `check-duplicate-keys` (4273 files), `check-enum-values` (2910), `check-entity-refs` (2829), `check-causal-targets` all clean; `validate-history-all` passes over 8324 records; 29 tests pass; `uvx ruff check .` clean. The derivation is unchanged at 245 assessed, 123 derived, 0 `ORIGIN_WITHOUT_CELL`, 7 `MULTI_ORIGIN_CELL`.

---

cmungall:
Took the `preferred_term` suggestion (`e47eb8c69`). It was worth another round, because a census that overstates what a record claims is the same defect this whole PR is about — and the hedge you found is a curator being careful, which the tooling was quietly flattening.

`_terms()` now prefers `preferred_term` and falls back to the label. Epithelioid sarcoma reads as its curator wrote it:

```
before  CL:0000134  mesenchymal stem cell
after   CL:0000134  mesenchymal cell of uncertain differentiation
```

Only the display name changes — the CURIE is what identifies a cell, so de-duplication, multi-origin detection and the generic-term check are untouched. Two tests pin that: one gives a single CURIE two different `preferred_term`s and asserts it is still one cell, another asserts a binding with no `preferred_term` still falls back to the label. 32 tests now, and the counts are unchanged at 245 / 123 / 0 / 7.

On the two siblings — `Anal_Canal_Adenocarcinoma` and `Cervical_Adenocarcinoma` — agreed, and agreed on where this PR stops. Both need a curator reading the mechanism to say which of their bound cells the disease arises from, and that is exactly the judgement the `NO_ORIGIN` backlog is meant to surface rather than have a script guess at.

Sorry for the approval churn. Every push since your first approval has been either a real CI failure or one of your findings, but I am conscious each one costs you a re-review.

---

cmungall:
Both suggestions taken in `88f12fbdc`. Acting on the first one turned up a real bug, so thank you for it.

## The bug

You traced every consumer of `_terms()` and concluded only the display name changes. That holds for everything that *reads* it — but `backfill_cancer_origin.py` imports it and **writes YAML from it**. `_cell_block()` used the same string for both `preferred_term:` and `term.label:`, so once `_terms()` started returning `preferred_term`, a borrowed binding would have produced:

```yaml
- preferred_term: mesenchymal cell of uncertain differentiation
  term:
    id: CL:0000134
    label: mesenchymal cell of uncertain differentiation   # not a CL label
```

`term.label` has to match the ontology exactly, so `just validate-terms` rejects that. Latent only — the backfill ran before `e47eb8c69`, and I scanned every committed entry to confirm none carries it.

`_terms()` now takes a `display=` flag. The census asks for the curator's wording; the backfill asks for the canonical label and reproduces both slots as the source entry had them, so a borrowed binding keeps the hedge in `preferred_term` without asserting it as the ontology's name. There is a regression test on the generated YAML.

## The two suggestions

**Key naming.** `origin_cell_labels` → `origin_cell_names` in the TSV, `"label"` → `"name"` in both JSON sites. Your reasoning was the deciding one: a consumer joining that string against CL labels would miss on exactly the hedged bindings the previous commit existed to preserve. Nothing consumes these yet, so renaming now costs nothing and later would be a break.

**Collapsed hedges.** `origin_cells` now keeps every distinct wording for a CURIE and joins them with ` / ` instead of `setdefault` dropping all but the first. Still one cell — the CURIE is the identity and `MULTI_ORIGIN_CELL` is unmoved — but a disagreement about how firmly to name it is the sort of thing this census is for. Test asserts one cell, no `MULTI_ORIGIN_CELL`, both wordings present.

## Checks

Census unchanged: 245 assessed, 123 derived (107 `SOMATIC_LESION` + 16 `ENVIRONMENTAL_TRIGGER`), `ORIGIN_WITHOUT_CELL` zero, `MULTI_ORIGIN_CELL` seven. 34 tests pass across the three modules; `uvx ruff@latest check` clean on all four changed Python files. No KB, schema or cache files in this commit, so the whole-KB gates are untouched.

The `model` field on history records is still yours to rule on — I've deliberately not resolved a contradiction between two standing instructions inside a PR.

---

cmungall:
All three taken in `d24507da6`. The quoting one is the same class of bug as the last commit, from the same cause — you caught this one, I caught the other.

**Unquoted scalars.** `_cell_block()` splices values into YAML text, which was safe while the value was an ontology label and stopped being safe when `preferred_term` started flowing through it. PyYAML now decides the quoting rather than a hand-written character test, because the distinction is subtler than it first looks: `ATP:ADP antiporter activity` is a perfectly good plain scalar, since only a colon *followed by a space* separates a mapping — a naive "contains a colon" test would over-quote. A value containing a newline falls back to JSON's double-quoted form, which is a valid YAML scalar and stays on one line, so it can't break the fixed indentation it's spliced into.

The test asserts the round trip rather than the output text: a borrowed `preferred_term` of `EMG: cell of uncertain differentiation` survives `safe_load` with the string intact.

**Locals.** `_fmt_cells` and the multi-origin renderer no longer bind to `label`/`lab`. I left `cell_terms` as it is — it holds `(id, name)` pairs, i.e. terms, so the field name isn't claiming anything about labels; renaming it to `cell_names` would be less accurate, not more. Say the word if you'd rather it moved.

**`propose()` nits.** The inner comprehension variables are `cell_name` now, so `name` means only the node name. `labels.get(c, name)` is a direct index with the invariant stated in a comment — you're right that the default was unreachable, since `labels` is keyed by the same `term.id` guard and is unfiltered, so it's a superset by construction.

Census unchanged: 245 assessed, 123 derived (107 + 16), `ORIGIN_WITHOUT_CELL` zero, `MULTI_ORIGIN_CELL` seven. 35 tests pass. Ruff clean on both the pinned 0.15.8 and latest — the earlier import-order slip is why I check both.

On the churn: this is the sixth approval a push will have dismissed. Both of the last two rounds found real latent bugs, so the rounds have earned their cost, but I'll treat further suggestions of the cosmetic kind as follow-up issues rather than pushing again.

---

cmungall:
Noting the `ensure_ascii=False` suggestion rather than pushing for it, per what I said last round.

You're right that the two paths disagree about the same value, and `json.dumps(value, ensure_ascii=False)` is the fix. But it's reachable only for a `preferred_term` containing a literal newline, no such value exists in the KB, and both forms round-trip correctly — so it changes the readability of output nothing currently produces. Against that, a push dismisses the approval and costs another eight-minute review round. Happy to take it if you'd rather it went in now, or as part of whatever next touches this file.

Thank you for checking the `labels[c]` invariant in both branches instead of taking the comment's word for it — converting an unreachable `.get` default into a `KeyError` is exactly the kind of change that deserves that, and stating the borrowed branch separately is the part I'd left implicit. Same for finding the seven committed colon-space `preferred_term` values: I'd argued the quoting bug from the shape of the code, and you confirmed the premise from the corpus.

The `model`-field question is the only thing left, and it's @cmungall's to settle in `CLAUDE.md` — agreed it shouldn't be raised in this thread again.

---

cmungall:
Fixed in `09663692e` and `5aa0765ef` — but I went further than you asked on PDAC, so I want to be explicit about that rather than have it look like I missed the boundary you drew.

## PDAC: I changed the binding, which you said was not for this PR

You asked me to drop the trailing clause and leave `CL:0002079` alone. I dropped the clause and **also rebound the node to `CL:0002064 pancreatic acinar cell`**.

The reason is in your own 🔵 section: with the ductal binding in place, the flagship worked example derives a cell its entry's cited lineage tracing rules out, and that derived answer is published in three docs and the PR body. Leaving a known-wrong claim on a rendered disorder page because correcting it is out of scope seemed worse than widening the commit. It also isn't me adjudicating the literature — the entry already adjudicated it, in prose, with a citation, and the binding contradicted the entry. The tumour is *called* ductal because of its histology, which is the exact confusion the ADM node exists to warn against.

If you disagree, revert is one line plus the four call sites, and I'll take the outcome without further argument.

**On the `MULTI_ORIGIN_CELL` option** — I think you're right that it is the better long-term answer, and I deliberately did not do it. Binding `CL:0002079` alongside acinar would assert a ductal-origin position this entry does not currently curate: there is no evidence item for it anywhere in the file. Doing it properly means finding the ductal-origin literature, verifying a snippet, and writing the node up as genuinely contested — curation work, not a correction, and it moves `MULTI_ORIGIN_CELL` from 7 to 8 with another docs cascade. Worth a follow-up issue, as you say. I've noted it to the maintainer as their call.

Four call sites moved with the binding: the YAML block and the chronic-pancreatitis contrast in `docs/cancer-cell-of-origin.md`, the curation SOP, `CLAUDE.md`, and `test_worked_exemplars_still_derive_their_documented_cell` — which now also asserts `CL:0002079` is *absent*, so a regression can't pass quietly. PR body updated too; it carried the same ductal claim.

## The two minor notes, both taken

**CML redundancy** — you were right, and it was worse than redundant: my replacement restated the parent node almost clause for clause. It now carries variant-level content instead, the t(9;22) and the p210 BCR-ABL1 product, the latter already asserted in this entry's own `genetic` block so nothing new is claimed.

**History records** — added for both entries.

## On your self-correction

Noting it because it is the useful part of this round: you flagged that you had approved the `docs/cancer-cell-of-origin.md` ductal sentence in an earlier pass without checking it against the entry. That sentence is what I copied when I wrote the bad one, and the binding underneath it is what made the exemplar wrong in the first place. Three rounds of us both reading it as a marker rather than as a claim is exactly how a wrong fact survives review, and you broke that by reading it as prose this time.

Checks on the current head: 35 tests pass; schema, terms, duplicate-key, entity-ref and causal-target gates clean on both entries; no cache churn; census unchanged at 245 assessed / 123 derived (107 + 16) / 0 `ORIGIN_WITHOUT_CELL` / 7 `MULTI_ORIGIN_CELL`, with CML deriving `CL:0000037` and PDAC now `CL:0002064`.

---

cmungall:
Your last note — the one you flagged as unverifiable rather than asserting — was the most valuable thing in the round. It was right, and I was wrong.

## The provenance claim was false

The history record said the ductal binding "came in with PR #10334". You couldn't check it on a depth-1 clone. Checking properly: **my clone is shallow too**, `faf1ed8df` is its graft boundary, and `git log --diff-filter=A` on a boundary commit reports it as *adding* a file it merely inherited. I read that output as an answer. That the commit was a "Regenerate pages" bot run should have been the tell — KB entries are not authored in page builds.

The record now says only what is verified: the binding is present on `origin/main` and this branch's diff does not add it. The file was actually created in **#528** (2026-03-09), from the API rather than the truncated local history. Which commit introduced this particular binding is not established, and the record now says so instead of guessing. I left the erroneous claim recorded as an error rather than quietly deleting it, since these are provenance artifacts.

Flagging a claim you can't verify, instead of nodding it through, is what caught this. Worth keeping.

## Two suggestions taken

- **p210 read as exclusive.** Fair — p190 and p230 occur in a minority of CML. Now "typically", matching how the entry's own `genetic:` block hedges it as "(p210 in CML)".
- **The abridged docs example.** You're right that eliding the sentence dropped exactly the part the example exists to teach. A curator copying the block now gets the ductal-phenotype explanation with it.

## Two not taken, with reasons

- **PR body.** Already updated — I rewrote it at 00:24, before your run started, so you were reading the pre-update snapshot. It carries the acinar block and a section on what review turned up.
- **The KRAS → Acinar-to-Ductal Metaplasia edge already exists.** It's on the KRAS node with `causal_link_type: DIRECT` and its own PMID:17349585 evidence. ADM has two parents — `Chronic Pancreatic Inflammation` *and* `KRAS Oncogene Activation` — so the graph does draw what the prose says. I think the layout put KRAS at the left margin as a root and made the two branches look parallel:

```
edge into ADM from: Chronic Pancreatic Inflammation
edge into ADM from: KRAS Oncogene Activation
```

Left open deliberately, as you framed it: `MULTI_ORIGIN_CELL` for PDAC needs the ductal-origin position curated with its own evidence, which is curation rather than correction, and is the maintainer's call.

35 tests pass; both entries schema-validate; census unchanged at 245 assessed / 123 derived / 0 `ORIGIN_WITHOUT_CELL` / 7 `MULTI_ORIGIN_CELL`.

---

cmungall:
Merged `origin/main` (124 commits behind) in `7705f4ff6`. One conflicted file, and the textual conflict was the smaller half of the problem.

## The conflict was a section-number collision

Both sides appended a new section at the same anchor after §3a in `docs/explanation/design-decisions.md`:

| Side | Added |
|---|---|
| main | `3b. Model-to-target scale gap`, `3c. Model divergence is typed` |
| this branch | `3b. Cell of origin is derived from the pathograph` |

Resolving on the text alone — keep both blocks, drop the markers — produces a clean-looking file with **two sections numbered 3b**. Every cross-reference this branch wrote to "design decisions §3b" would then resolve to main's model-scale decision instead of the cell-of-origin one. Silent, and exactly the sort of thing that reads as resolved.

Kept both sections and renumbered this branch's to **3d**. Main's 3b/3c are already merged and already cited elsewhere — `docs/explanation/model-credibility.md` says "design decisions 3b and 3c" — so the unmerged side is the one that yields the number. Updated the two references that meant cell-of-origin (`CLAUDE.md`, `docs/cancer-curation-sop.md`); left `§3a` references and model-credibility's `3b and 3c` untouched, since those still point where they should.

## Verified rather than assumed

- **No main content reverted.** 276 files differ from main; 276 files were touched by this branch; the set difference is empty. So nothing differs from main that this branch did not deliberately change. The `references_cache/PMID_*.md` diffs looked alarming until I traced them — they are this PR's own `f0f685bc9`, and main never touched those files.
- **No duplicate keys** across 4,286 YAML files — the #8623 hazard, which is precisely a merge-shaped defect.
- **The six cancer entries both sides edited** kept their `SOMATIC` marks.
- **Derivation intact**: 123 origins derived (107 `SOMATIC_LESION` + 16 `ENVIRONMENTAL_TRIGGER`), `ORIGIN_WITHOUT_CELL` zero, `MULTI_ORIGIN_CELL` seven, CML → `CL:0000037`, PDAC → `CL:0002064`. Assessed rises 245 → 247 and no-origin 93 → 95, purely from cancer entries main added.
- 35 tests pass. `enum-values` (2,990), `entity-refs` (2,909), `causal-targets`, `qualifier-terms` (238), `title-snippets`, `snippet-grading`, `snippet-length`, `folded-hyphens`, `environmental-evidence` all clean. `behind_by` now 0, no conflicts.

## One thing worth flagging for anyone auditing this branch

My earlier conflict checks used `git merge-tree <base> <a> <b> | grep -c '^<<<<<<<'` and reported **zero conflicts** on the run immediately before this one, when the conflict already existed. That legacy form does not emit conflict markers the way the pipeline assumed. `git merge-tree --write-tree` reports it correctly via exit status. Anything else relying on the legacy form as a conflict test is reporting false negatives.

---

cmungall:
`claude-review` failed on `7705f4ff6`, and it is not this PR's failure. Recording what it is, since the check is red and will stay red until the underlying condition clears.

**The agent hit an account usage limit.** From the job log:

```
detail="You've hit your limit · resets 1:20am (UTC)"  num_turns=1  cost=$0  subtype=success
```

No review ran — one turn, zero cost. The workflow's own diagnostic explains the shape: it prefers `ANTHROPIC_API_KEY` (pay-per-use, no weekly cap) and falls back to `CLAUDE_CODE_OAUTH_TOKEN` only when no API key is set, so a weekly-limit error implies the key is unset. The job's env dump confirms it — `ANTHROPIC_API_KEY:` is empty, so the run used the subscription token.

**Nothing in this PR caused it and there is no fix to port.** It is a credential/quota condition on the runner, independent of the diff; setting a repository secret is a maintainer action, not something I should or can do from here.

**What I am doing:** the limit resets at 01:20 UTC, so re-running now would fail identically. I'll re-run the check once after the reset rather than spending the retry on a run that cannot succeed. If it fails again for the same reason, that is a signal `ANTHROPIC_API_KEY` needs setting on the repo — @cmungall.

`test (3.13)`, the substantive check, is still running on this head; I'll report it when it lands.

For completeness, the merge this ran against was verified locally before pushing: 35 tests, nine whole-KB gates, no duplicate keys across 4,286 YAML files, no main content reverted (276 files differ from main, 276 touched by the branch, set difference empty), derivation intact at 123 origins / 0 `ORIGIN_WITHOUT_CELL` / 7 `MULTI_ORIGIN_CELL`.

---

cmungall:
Correcting my last comment: I said I'd re-run `claude-review` after the usage limit reset. **I can't** — I don't have the permission, and I should have checked that before promising it.

The limit has reset and `test (3.13)` finished green on `7705f4ff6`, so the merge itself is validated. But both re-run routes are refused:

```
POST .../actions/runs/33928605845/rerun-failed-jobs   → 403 Resource not accessible by integration
POST .../actions/workflows/claude-code-review.yml/dispatches → 403 Resource not accessible by integration
```

Identical on both endpoints, so it's a permissions boundary rather than a transient error.

**Where that leaves the PR.** `claude-review` is red for a reason unrelated to the diff, and the deterministic auto-merge sweep requires a check rollup with nothing failing — so this red check will hold the PR even once it becomes age-eligible today. It needs one of:

- someone with Actions write re-running the failed job on run `33928605845`, or dispatching `claude-code-review.yml` with `pr_number=10629`; or
- setting the `ANTHROPIC_API_KEY` repo secret, which is the underlying fix — the run fell back to the subscription token because that key is unset, which is what hit the weekly cap.

@cmungall — flagging because it needs Actions write or a repo secret, neither of which is mine to do.

I'm not pushing anything to force a re-trigger: an empty commit or a no-op change to kick CI is exactly what the contributing rules forbid, and the next genuine push will re-run the check anyway. Keeping the PR watched until it's green or merged.

State on `7705f4ff6`: `test (3.13)` SUCCESS, `dispatch-guard` skipped, `claude-review` failed on the quota condition above. `behind_by` 0, no conflicts.

---

cmungall:
Both findings fixed, plus one the merge turned up. All in this push (`6e00aa4d5`).

**🟡 The stale schema pointers.** Fixed — `dismech.yaml:3186` and `:4914` now say `3d`. You were right that this was worth a round rather than a nit, and for the reason you gave: main's §3b is also a "derived, not asserted" decision, so the pointer landed somewhere that reads on-topic.

I checked whether the sweep missed anything else. Repo-wide there were 40 occurrences of a `3b` cross-reference; 2 were these, and the other 38 are history records under `history/`. I left those alone — they are append-only records of sessions that ran while the section really was `3b`, so they describe their own moment accurately. Say the word if you'd rather they were swept too.

**🔵 The counts.** Refreshed, but derived against the merged tree rather than carried across, and they moved again because main kept adding entries: **252 assessed / 124 derived / 128 unmarked / 8 multi-origin**. Updated in `docs/cancer-cell-of-origin.md` (both sites plus the "seven current findings" line, now eight), `docs/explanation/design-decisions.md`, `CLAUDE.md`, and the PR body.

Since this is the second review round these counts have cost, I reworded the PR body's parenthetical to say they are a snapshot that drifts with main, rather than quoting a figure that reads as current.

**The merge surfaced a real regression — `ORIGIN_WITHOUT_CELL` went 0 → 1.**

This is the one thing beyond what you asked for, and it is the check catching something rather than a cleanup. Main rewrote `Ewing_Sarcoma` and added a `Permissive Progenitor Cell State` node holding both candidate lineages with `mechanism_confidence: PROVISIONAL`. This branch's `SOMATIC` marker sits on `EWS-FLI1 Fusion Oncogene`, which binds no cell — so the entry marked an origin without one, and the docs' claim that the class is ready to be a hard gate became false in the same push that restates it.

Three ways to fix it, and two are wrong:

- removing the `SOMATIC` marker — wrong, the fusion genuinely is the transforming lesion
- binding one cell — wrong, it asserts a resolution the entry explicitly declines to make
- binding both — which is what the `MULTI_ORIGIN_CELL` class is *for*

Bound both, with a `notes:` line saying the pair is deliberate so nobody tidies it into one later. This asserts nothing new: both CL terms and the unresolved-origin position are already in the entry, on a node whose description says the human cell of origin remains unresolved. Ewing now reads as `MULTI_ORIGIN_CELL` — same shape as `Melanoma_in_Congenital_Melanocytic_Nevus`, and I added it to that worked example in the docs. `ORIGIN_WITHOUT_CELL` is back to **0**. History record added.

**ℹ️ The three ref-baseline gates.** Your diagnosis was right, and I verified it rather than assuming: after the fresh main merge, `check-title-snippets`, `check-snippet-grading` and `check-snippet-length` all exit 0. No baseline was touched.

**Verification on the merged tree**

| Check | Result |
|---|---|
| `check_duplicate_yaml_keys`, `check_enum_values`, `check_entity_refs`, `check_causal_targets` | OK |
| `check-title-snippets` / `check-snippet-grading` / `check-snippet-length` | exit 0 (were red pre-merge) |
| `linkml-validate` + `linkml-term-validator --labels` on `Ewing_Sarcoma` | passed |
| `validate-history` on the new record | passed |
| `test_cancer_origin` + `test_backfill_cancer_origin` + `test_ontology_edge_relations` | **35 passed** |
| `check-cancer-origin` | 252 assessed / 124 derived (107 `SOMATIC_LESION`, 17 `ENVIRONMENTAL_TRIGGER`) / **0 `ORIGIN_WITHOUT_CELL`** / 8 `MULTI_ORIGIN_CELL` |

Scope check on the merged result, since that is what the merge could have quietly broken: KB footprint vs main is +551/−3 with the three deletions being the previously-approved PDAC rebinding, infra is +2435/−1 with the one deletion being the `qc:` recipe line, and zero `variant_origin` marks were dropped.

---

Issue #9865, opened 2026-08-28
Stub queue: 31 numbered cardiomyopathy stubs sit under existing Familial_Dilated/Hypertrophic_Cardiomyopathy groupings


cmungall:
The queue holds 23 numbered dilated cardiomyopathy stubs (`Dilated_Cardiomyopathy_1A`-style plus the `Cardiomyopathy_Dilated_*` spelling) and 8 hypertrophic cardiomyopathy stubs (`Hypertrophic_Cardiomyopathy_*`, `Cardiomyopathy_Familial_Hypertrophic_*`). The KB already has `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml` and `kb/groupings/Familial_Hypertrophic_Cardiomyopathy.yaml` over per-gene disease entries, so these numbered stubs are in limbo: they are neither stale (their own MONDO IDs are not covered) nor obviously curation targets as numbered entities.

The lump/split review left them alone because the right disposition depends on how the cardiomyopathy corner of the KB intends to grow, which is a curator call:

- If per-gene entries are the pattern (as the existing groupings' member lists suggest), the useful move is to rename/replace each numbered stub with a gene-shaped one for genes not yet covered (e.g. "CMD1A" is really LMNA-related DCM, which likely belongs with the existing laminopathy curation rather than as a fresh entry), and delete numbered stubs whose gene is already covered by a member entry.
- If the numbered OMIM concepts are not curation targets at all, they can be deleted as `SUBTYPE`/covered-by-grouping decisions in one PR, leaving gene gaps to be nominated properly.

Either way, someone who knows the cardiomyopathy curation plan should make one pass over the 31; a mechanical rule was deliberately not applied in [PR #9855](https://github.com/monarch-initiative/dismech/pull/9855).

Issue #7837, opened 2026-08-02
Coverage gap: obstetrics is nearly absent from the KB (25 neoplastic vs 7 non-neoplastic women's-health entries)


cmungall:
A coverage survey of women's-health entries found the gap is systematic rather than scattered.

## The finding

Of ~32 women's-health entries in `kb/disorders/`, **25 are neoplastic and 7 are not**:

> Breast_Fibroadenoma · Endometriosis · Peripartum_Cardiomyopathy · Polycystic_Ovary_Syndrome · Preeclampsia · Premenstrual_Dysphoric_Disorder · Uterine_Leiomyoma

Gynecologic oncology is curated to a fine grain — ovarian carcinoma alone has six histotype-specific entries. Before this work, the entire obstetric domain was three entries (Preeclampsia, Peripartum Cardiomyopathy, GTN) out of 1,623 disorders.

## Now added

Three entries landed on `claude/womens-disorders-kb-review-9cge2f`: `Intrahepatic_Cholestasis_of_Pregnancy`, `Placenta_Accreta_Spectrum`, `Gestational_Diabetes_Mellitus`.

## Still absent — obstetrics, roughly by global burden

- **Postpartum haemorrhage** — leading cause of maternal death worldwide. *Blocked on a MONDO gap, see below.*
- **Preterm birth** — leading cause of neonatal death worldwide
- **Placental abruption**, **placenta previa**
- **Postpartum depression** (MONDO:0005929 exists) — neurosteroid/allopregnanolone withdrawal mechanism with two mechanism-matched drugs (brexanolone, zuranolone); unusually tractable
- **Ectopic pregnancy**, **recurrent pregnancy loss**
- **Chorioamnionitis**, **maternal sepsis**
- **Rh alloimmunisation / haemolytic disease of the fetus and newborn** — classic mechanism, prevented by anti-D
- **Amniotic fluid embolism**, **fetal growth restriction**, **obstetric fistula**, **hyperemesis gravidarum**

## Still absent — benign gynecology

[...]

PR #10647, opened 2026-09-02
curate(Dilated_Cardiomyopathy_1CC): add NEXN Z-disc/junctional-membrane DCM entry (MONDO:0013147)

### LINKED ISSUE #10633: Curate dilated cardiomyopathy 1CC (MONDO:0013147)
Curate a dismech entry for **dilated cardiomyopathy 1CC** ([MONDO:0013147](https://monarchinitiative.org/MONDO:0013147)) — NEXN-related familial dilated cardiomyopathy, where a defective F-actin-binding Z-disc protein destabilises sarcomeric force transmission.

**Why this one.** Selected from the priority list in #1079 by filtering its entries through `dashboard/priority.json` and keeping only rows recommended `CURATE_ROOT` (dashboard score **66.5**) that are absent from `kb/` and carry no prior open or closed issue or PR. Unchecked boxes in #1079 are not a curation queue on their own — roughly three quarters of them are `LUMP_INTO_PARENT` rows, and filing against one of those has previously been closed `wontfix`.

**Coverage check** (`origin/main`, `kb/`). `NEXN` appears in `kb/` only as a gene named inside `Dilated_Cardiomyopathy`, `Hypertrophic_Cardiomyopathy` and `Hypertrophic_Cardiomyopathy_20`. It is **not** among the genes the `Familial_Dilated_Cardiomyopathy` grouping lists as represented by a member entry (TTN, LMNA, MYH7, TNNT2, TNNC1, DES, SCN5A, BAG3, RBM20, LMOD2, ACTN2, TNNI3), so the NEXN disease concept has no anchor anywhere in the KB.

**What a separate entry has to justify.** The KB's cardiomyopathy corner already curates per-gene entries under the two familial cardiomyopathy groupings, so the pattern is established. A standalone entry should state what NEXN adds beyond the `Dilated_Cardiomyopathy` umbrella — the expected line is the Z-disc/actin-anchoring mechanism, which is distinct from both the thick-filament and the nuclear-envelope arms already curated.
Note #9865, which flags the 31 numbered cardiomyopathy stubs as awaiting a curator's lump/split pass. This issue is one such call, made explicitly rather than mechanically.

**NEC check:** the dashboard flags `ACRONYM_AMBIGUITY` on `CMD1CC`. Anchor identity on the OMIM xref and the stub's `genes:` block (`hgnc:29557` NEXN) before curating from any deep-research output, and run `just preflight-dr <report.md> MONDO:0013147 --strict` — `just research-disorder` hardcodes an empty `mondo_id` (#10335), so reports carry no ontology anchor of their own.

Stub: `stubs/Dilated_Cardiomyopathy_1CC.yaml`
Causal gene per MONDO: `hgnc:29557` (NEXN)

AO33:
## The decision: standalone entry, not a subtype

NEXN-related dilated cardiomyopathy now has its own entry, and I added it as the
fourteenth member of the `Familial Dilated Cardiomyopathy` grouping. I made this
call explicitly rather than by default, since #9865 flags the numbered
cardiomyopathy stubs as awaiting exactly that judgement.

Three things pushed it over the line:

1. **Gene-disease validity.** ClinGen's Dilated Cardiomyopathy Gene Curation
   Expert Panel rates NEXN **Strong** under SOP10 as of 2024-11-15. That is an
   upgrade from the moderate tier recorded in the 2021 reappraisal — and ACTN2,
   still sitting in that moderate tier, already has its own entry as
   `Dilated Cardiomyopathy 1AA`. Holding NEXN to a stricter bar than the
   precedent would have been inconsistent.
2. **Mechanism.** NEXN is the F-actin *anchoring* lesion — at the Z-disc and at
   the junctional membrane complex. It is the only member that reaches
   contractile failure through a failure of transverse-tubule biogenesis. The
   grouping's own notes list the genes already represented (TTN, LMNA, MYH7,
   TNNT2, TNNC1, DES, SCN5A, BAG3, RBM20, LMOD2, ACTN2, TNNI3 via CMD2A); NEXN
   was absent.
3. **Clinical distinctness.** The zygosity split has no sibling analogue:
   heterozygous disease is incompletely penetrant and typically presents near
   age 50, while biallelic loss of function causes fetal-onset cardiomyopathy
   with cardiomegaly and endocardial fibroelastosis. That changes both the
   differential and the counselling.

MONDO:0013147 is a direct child of MONDO:0700335, so the entry satisfies the
parent-child relationship the grouping's MONDO mapping asserts of every member,
and it carries HP:0001644 plus four `conforms_to` edges into
`cardiomyopathy_maladaptive_remodeling` — that grouping's two NECESSARY criteria.
`check-groupings` confirms all fourteen members still satisfy them.

## The interesting part: an unresolved mechanism, left unresolved

The literature genuinely disagrees about what nexilin does. Hassel 2009 isolated
it as a Z-disc protein and found matching Z-disc pathology in human carrier
myocardium. Liu 2019 then stated flatly that NEXN is *"rather than being a Z-disk
protein as previously thought"* a junctional membrane complex component required
for T-tubule formation, and reported no sarcomeric alteration in their mouse.

I curated this as two competing `mechanistic_hypotheses` (CANONICAL and
ALTERNATIVE) with a `KNOWLEDGE_GAP` discussion, rather than picking a winner.
The Liu quote is attached to the Z-disc node as a **REFUTE** evidence item, and
the mouse model carries a `FAILS_TO_RECAPITULATE` link against that node with its
limitations spelled out. A reader should be able to see the dispute from the
entry.

A second `HUMAN_MODEL_MISMATCH` discussion records what I think is the weakest
link in the entry: every published NEXN model is a biallelic null or homozygous
knock-in, while the disease most patients have is heterozygous and adult-onset.
The striking AAV-Nexn rescue result was obtained in an arm of the disease that is
not the common one.

## The deep-research report was a lead, and four of its bindings were wrong

The Falcon/Edison report passed `just preflight-dr ... MONDO:0013147 --strict`
(NEXN mentioned 48 times, correct gene, correct entity), and its primary-literature
pointers were good. Its **ontology suggestions were not**, and I want this on the
record since these all look plausible:

| Report offered it as | The code actually means |
|---|---|
| `HP:0004308` Ventricular dilatation | **Ventricular arrhythmia** |
| `HP:0001622` Premature death | **Premature birth** |
| `NCIT:C16502` Echocardiography | **Diagnostic Imaging Testing** |
| `NCIT:C16810` Magnetic Resonance Imaging | **Magnetic Resonance Spectroscopy** |

All four were rejected and re-derived against the ontology. Separately, a PMID I
drafted from memory for the Han Chinese cohort was wrong (32054922); the real one
is 32041989, caught by resolving the DOI before citing.

Stillbirth and neonatal death are curated as phenotypes **without** a bound
`phenotype_term`, following the existing convention in
`Hemolytic_Disease_of_the_Fetus_and_Newborn` — HP:0003826 and HP:0003811 sit
under Mortality/Aging, outside the `PhenotypeTerm` enum root (#7837).

- **No dominant/recessive subtype split.** MONDO does not split them, it is one
  gene and one convergent mechanism, and the difference is already carried by two
  `inheritance` blocks plus zygosity-specific phenotypes.
- **No `datasets` block.** No NEXN-specific dataset was identified, and a
  gene-symbol search would have surfaced whatever NEXN is best known for rather
  than this disease.
- **No `clinical_trials` block.** No interventional trial has enrolled on NEXN
  genotype. The AAV-Nexn work is preclinical and is curated as an animal-model
  readout, not as a therapy.
- **No prevalence rate.** The only numbers available are gene shares of
  clinically ascertained sequencing cohorts. That is a diagnostic yield, so it
  went into `Genetic.case_fractions` (4.8%, n=118 Han Chinese cohort), not
  `prevalence`.

While editing the grouping I also corrected two counts in its notes that had gone
stale before this PR — "all eight carry HP:0001644" had been wrong since the
membership grew past eight, and the mechanistic-class enumeration said "seven"
while listing ten. Both are now stated for the current fourteen members.

## Validation

All run, all output real:

```
just validate                     → ✅ Validation passed (schema + terms)
just count-verified-snippets      → Snippets checked: 67/67 verified
just check-duplicate-keys         → OK: no duplicate mapping keys
just check-entity-refs            → OK: entity references resolve
just check-causal-targets         → OK: no new broken pathograph targets
just check-qualifier-terms        → OK
just check-enum-values            → OK: no out-of-enum values in 2 KB file(s)
just check-folded-hyphens         → OK: no new folded-scalar hyphen splits
just check-snippet-length         → OK: no new snippets under 5 words
just check-title-snippets         → OK: no new title-quoting snippets
just check-snippet-grading        → OK: no new snippet-grading divergences
just validate-grouping            → ✓ All validations passed
just check-groupings              → all 14 members SATISFIED
just validate-history             → No issues found
just validate-disorders           → ✓ All 1 disorder file(s) passed validation
```

Nothing failed. Cache churn is 8 additive lines from my own terms; no `pages/`,
`dashboard/`, or `app/models/data.js` staged.

Closes #10633
Refs #9865


AO33:
Heads-up for whoever merges this: **#10647 and #10649 both edit `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml`**, and each was branched from `origin/main` without the other. Each adds its own entry as "the 14th member" and each independently fixes the same stale counts in that file's notes ("the thirteen members", the "all eight carry HP:0001644" line).

Two consequences:

1. Whichever merges second will conflict, or — worse — may *not* conflict. This is the #8623 pattern: two PRs adding a block at different points in one file can merge cleanly and leave a duplicate key, which every test and renderer here silently tolerates while `linkml-reference-validator` aborts on it. Please resolve by **merging both members in and reconciling the counts to 15**, not by taking one side.
2. The member count in whichever merges first will be wrong the moment the other lands.

A third entry, CMD1NN (#10650), belongs in this grouping by MONDO but was deliberately left out precisely because of this contention — its curator flagged that the grouping's rationale enumerates mechanistic classes and would need a prose edit to admit a signalling-kinase lesion. That is worth doing as a single follow-up PR once these two are in, rather than a third concurrent edit to the same file.

I am the author of all three and raised the conflict rather than resolving it unilaterally, since the counts and the rationale prose are a curator call.

AO33:
All three blocking findings are fixed in 4d54048. I also took three of the four non-blocking suggestions, and declined two things with reasons below.

### The three blocking items

**Item 1 — the hypothesis description contradicted its own status.** The sentence said "Marked ESTABLISHED rather than CANONICAL", and `ESTABLISHED` is not a value the schema has. I kept `status: CANONICAL` and rewrote the sentence: the Z-disc model is retained as canonical because it is still the default reading of the disease, but the text now says plainly that it is contested rather than settled, and points at the `ALTERNATIVE` junctional-membrane hypothesis below it. That was the real content of the original sentence, minus the invented enum value.

**Item 2 — background sentence used as causal evidence.** Swapped in the result sentence you identified, "essential for optimal calcium transients", and rewrote the explanation to say what the measurement establishes instead of conceding it is background.

**Item 3 — GeneReviews baseline.** Added PMID:20301486 with `tags: [GeneReviews]`.

On which of the two sibling patterns to copy: I followed 1E, not 1JJ, and said why in `notes`. The cached record is `content_type: abstract_only` and its entire body is the chapter's four-point purpose statement — "Define DCM", "Identify the categories of DCM", and so on. 1JJ quotes two of those clauses as evidence items. They are exact substrings, so they validate, but a statement of what a chapter sets out to cover is not a finding about NEXN disease, and quoting it would be the same defect as item 2 in a different place. So the reference is tagged and the reasoning is recorded, with no snippet mined from it.

### Non-blocking suggestions taken

All three use references already in the cache, so no new fetches:

- **Heart transplantation** — added as a treatment (`NCIT:C15246`) with a `target_mechanisms` link to `Heart Failure`. Evidence is the NEXN heterozygote assessed for transplant at 15% ejection fraction (PMID:35166435) on the link, and the EFE-to-transplant statistic from the NEXN case report (PMID:33949776) at treatment level, which covers the biallelic arm as well as the adult one.
- **NT-proBNP** — added a `biochemical` block (`NCIT:C88524`) with a `THRESHOLD_DEPENDENT` readout on `Heart Failure`. The reason it earns a place in a gene-specific entry is that it is what discriminated severity among carriers in the index family: high in the one who became a transplant candidate, normal in the one who improved.
- **Arrhythmia** — added `HP:0011675` at `OCCASIONAL`. I sourced it specifically to the biallelic patient in whom the authors explicitly excluded a second conduction-gene allele, because several other published arrhythmic NEXN cases carry a co-inherited SCN5A variant and are confounded.

### Not done, and why

- **A standalone ventricular assist device treatment.** The only NEXN-specific record of VAD use in the cached sources is a summary endpoint table that the cache has flattened into an unreadable run-together string ("Severe DCMAnti-congestivemedicationVAD(22 years)"). There is no quotable finding in it. VAD is described inside the transplantation entry as the bridge therapy instead.
- **Pregnancy as a second hit.** The homozygous p.Glu528del sisters come from a 2013 report that is not in the reference cache, so the claim would have had to be sourced from the deep-research narrative rather than a verifiable quote.

Both are recorded in `notes` so they are not silently dropped.

### The grouping file

Deliberately untouched. Since this PR and #10649 both edit `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml` and the member count needs reconciling to 15 across both, adding anything here would only make that merge harder. The grouping still validates as it stands.

### Validation

`just validate-disorders` passes, 72/72 snippets verified against the cache (up from 67 — the five new ones are the additions above). Also green: `validate`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-grading`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`. `validate-grouping` and `check-groupings` pass on the unchanged grouping file.

One note for anyone adding evidence near a citation marker: I first tried to quote the PMID:40161564 sentence "...requiring heart transplantation[12]." Neither writing the brackets nor omitting them matches, because the validator strips the bracketed span and leaves a stray space before the period. I used a bracket-free sentence instead rather than touching `literal_bracket_patterns`.

Ontology terms were checked against the ontology rather than recalled. Worth noting that my own first guess for a heart transplantation code, `NCIT:C158032`, turned out to be "Kidney Epithelioid Angiomyolipoma" — the same class of plausible-looking wrong binding the original curator caught four of.

PR #10649, opened 2026-09-03
curate(Dilated_Cardiomyopathy_1FF): dominant TNNI3 dilated cardiomyopathy (MONDO:0013211)

### LINKED ISSUE #10637: Curate dilated cardiomyopathy 1FF (MONDO:0013211)
Curate a dismech entry for **dilated cardiomyopathy 1FF** ([MONDO:0013211](https://monarchinitiative.org/MONDO:0013211)) — dominant TNNI3-related dilated cardiomyopathy, where cardiac troponin I variants reduce myofilament calcium sensitivity.

**Why this one.** Selected from the priority list in #1079 by filtering its entries through `dashboard/priority.json` and keeping only rows recommended `CURATE_ROOT` (dashboard score **63.5**) that are absent from `kb/` and carry no prior open or closed issue or PR. Unchecked boxes in #1079 are not a curation queue on their own — roughly three quarters of them are `LUMP_INTO_PARENT` rows, and filing against one of those has previously been closed `wontfix`.

**Coverage check** (`origin/main`, `kb/`). `TNNI3` is well covered in `kb/` — `Hypertrophic_Cardiomyopathy`, `Hypertrophic_Cardiomyopathy_7`, `Dilated_Cardiomyopathy`, `Dilated_Cardiomyopathy_2A` and `Restrictive_Cardiomyopathy` — but MONDO:0013211 itself is bound nowhere.

**What a separate entry has to justify.** `Familial_Dilated_Cardiomyopathy` states in its own notes that **"TNNI3 is represented by the recessive CMD2A entity"**. CMD1FF is the *dominant* TNNI3 dilated cardiomyopathy, so the gap is an inheritance-mode split of the kind the KB already makes elsewhere. Confirm the dominant/recessive distinction is real in the literature before creating a separate entry; if it is, say so against `Dilated_Cardiomyopathy_2A` explicitly, and if it is not, this is a `has_subtypes` entry there.
Note #9865, which flags the 31 numbered cardiomyopathy stubs as awaiting a curator's lump/split pass. This issue is one such call, made explicitly rather than mechanically.

**NEC check:** the dashboard flags `ACRONYM_AMBIGUITY` on `CMD1FF`. Anchor identity on the OMIM xref and the stub's `genes:` block (`hgnc:11947` TNNI3) before curating from any deep-research output, and run `just preflight-dr <report.md> MONDO:0013211 --strict` — `just research-disorder` hardcodes an empty `mondo_id` (#10335), so reports carry no ontology anchor of their own.

Stub: `stubs/Dilated_Cardiomyopathy_1FF.yaml`
Causal gene per MONDO: `hgnc:11947` (TNNI3)

AO33:
Curates **dilated cardiomyopathy 1FF** (MONDO:0013211, OMIM 613286) — the autosomal dominant TNNI3 dilated cardiomyopathy — as a new entry, and replaces the seed stub.

## The lump/split decision: separate entry

TNNI3 already appears across five entries, and `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml` said in its own notes that "TNNI3 is represented by the recessive CMD2A entity". So the question was whether CMD1FF is a real entity or just a second OMIM number on `Dilated_Cardiomyopathy_2A`.

It holds up, on four axes rather than on nomenclature:

- **Molecular lesion.** CMD2A is a biallelic null — no cardiac troponin I is made at all. CMD1FF always retains a wild-type allele, and the founding alleles are missense, so an altered cTnI is present in the thin filament. There is no null state.
- **What breaks, and in which direction.** Reconstituted troponin carrying either founding CMD1FF allele (Lys36Gln, Asn185Lys) gives lower maximum actin-tropomyosin-activated myosin ATPase and lower calcium sensitivity, with reduced thin-filament calcium affinity (Carballo 2009, PMID:19590045). That is the sarcomeric-DCM signature, and it is directionally opposite to the calcium-sensitizing TNNI3 alleles of hypertrophic and restrictive disease (Tadros 2020, PMID:32278834). CMD2A's mechanism is instead global regulatory failure gated by the perinatal troponin I isoform switch — a clock that has no meaning when one allele is intact.
- **Natural history.** Variable, incompletely penetrant, childhood-to-adult onset here; obligate neonatal-to-infantile and near-uniformly severe there.
- **What a clinician does next.** 50% per-offspring recurrence with longitudinal surveillance of first-degree relatives, versus 25% per sibship with carrier testing and reproductive counselling.

ClinGen backs the split externally: its Dilated Cardiomyopathy GCEP issued two TNNI3-DCM assertions on the same day differing only in mode of inheritance, both rated Strong.

So `entry_type` is `DISEASE`, and the contrast with CMD2A is stated explicitly in the entry description, in a `RESOLVED` discussion, and in reciprocal differential-diagnosis blocks on both entries.

## What is in the entry

Seven pathophysiology nodes, four of them conforming to `cardiomyopathy_maladaptive_remodeling`, running from the heterozygous TNNI3 variant through altered thin-filament calcium regulation to a dilated, failing ventricle. Three mechanistic hypotheses are curated rather than one:

- **CANONICAL** — altered thin-filament calcium regulation by mutant cTnI.
- **ALTERNATIVE** — the primary defect is loss of *regulation* of calcium sensitivity, not its resting value: PKA phosphorylation of troponin I normally desensitizes the myofilament two- to three-fold, and in thin filaments carrying the CMD1FF allele K36Q that coupling is abolished entirely (Memo 2013, PMID:23539503).
- **EMERGING** — haploinsufficiency from a heterozygous truncating allele, characterized in explanted human myocardium (Bollen 2017, DOI:10.1113/jp274145).

Three discussions, of which one is worth flagging to a reviewer. The organising generalization for dominant thin-filament DCM is that its alleles *desensitize* the myofilament to calcium. The one heterozygous TNNI3 truncating allele studied in human myocardium did the opposite — increased calcium sensitivity with impaired length-dependent activation. Both results look sound and they differ in allele class, preparation, and whether end-stage remodelling could contribute. Rather than pick a side, the entry records this as an open `KNOWLEDGE_GAP` with the Bollen result graded `REFUTE` against the generalization, and proposes the two experiments that would settle it. It matters clinically: a calcium sensitizer would be predicted to help one allele class and harm the other.

## Grouping note fixed

CMD1FF was added to `Familial_Dilated_Cardiomyopathy` as a fourteenth member, and the stale claim that TNNI3 is represented only by CMD2A was rewritten — TNNI3 now contributes both members and is the group's one intragenic dominant/recessive pair. Two other stale counts in the same file were corrected while there ("the thirteen members", and an "all eight carry HP:0001644" line that predated several admissions). Both members satisfy both NECESSARY criteria; `just check-groupings` confirms.

This entry is one of the numbered cardiomyopathy stubs tracked in #9865.

`just preflight-dr … MONDO:0013211 --strict` passed (TNNI3 mentioned 51 times; the report's OMIM set covers 613286), and the report does correctly separate CMD1FF from CMD2A and from the hypertrophic and restrictive TNNI3 diseases. It was still used as leads only, and three of its outputs did not survive checking:

- It reports being unable to establish a MONDO identifier and suggests mapping to parent concepts. MONDO:0013211 exists; resolved directly with OAK.
- It suggests `HP:0001712` for "Left ventricular dilatation". HP:0001712 is **Left ventricular hypertrophy**. Used HP:4000141 instead.
- Its citation set does not include Carballo 2009, which defines this entity and is the OMIM 613286 anchor. That, plus the Memo 2013 uncoupling study, the Robinson 2020 cardiomyocyte study of the K36Q allele, and the Tadros 2020 meta-analysis, were all found independently.

- **No prevalence figure.** The report's Polish-cohort claim (TNNI3 ≈ 1% of gene findings) is plausible, but that reference's cache file has no abstract body, so no exact snippet could be verified. Left out rather than paraphrased.
- **No `environmental:` section.** The report's second-hit list (alcohol, anthracyclines, peripartum stress, myocarditis) is general genetic-DCM biology with no TNNI3-specific evidence; attaching it to a mechanism node would assert more than the sources do.
- **No animal model.** None carrying a CMD1FF allele has been reported — recorded as a knowledge gap, with the existing thin-filament knock-in models named to show what the gap looks like.
- **No clinical trials or datasets block.** Nothing registered or deposited for this entity was found.
- **No `qualifiers:` anywhere**, so the `check-qualifier-terms` blind spot does not apply here (`check-qualifier-terms` reports 0 qualifier terms in the file).

## Validation

All run, all output real:

```
just validate kb/disorders/Dilated_Cardiomyopathy_1FF.yaml
  Schema: No issues found | Terms: ✅ | References: All validations passed

just validate-disorders kb/disorders/Dilated_Cardiomyopathy_1FF.yaml kb/disorders/Dilated_Cardiomyopathy_2A.yaml
  Snippets checked: 114/127 verified against cached references (13 skipped by prefix)
  ✓ All 2 disorder file(s) passed validation.

just check-duplicate-keys / check-entity-refs / check-enum-values      OK
just check-causal-targets    OK: no new broken pathograph targets
just check-qualifier-terms   OK: 0 qualifier term label(s)
just check-folded-hyphens / check-snippet-length / check-title-snippets
  / check-snippet-grading                                             OK, no new entries
just validate-grouping + check-groupings kb/groupings/Familial_Dilated_Cardiomyopathy.yaml
  Dilated Cardiomyopathy 1FF: SATISFIED (both NECESSARY criteria)
just validate-history history/disorders/Dilated_Cardiomyopathy_1FF/…  No issues found
```

The 13 snippets `validate-disorders` skips by prefix are the `DOI:` ones, so I checked all 63 evidence snippets in the new entry against their cached reference files with an independent exact-substring script: 63 checked, 0 problems. That caught one real defect — a `wild-type` I had typed with an ASCII hyphen where the cached source uses U+2010 — which is now fixed.

Closes #10637


AO33:
Heads-up for whoever merges this: **#10647 and #10649 both edit `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml`**, and each was branched from `origin/main` without the other. Each adds its own entry as "the 14th member" and each independently fixes the same stale counts in that file's notes ("the thirteen members", the "all eight carry HP:0001644" line).

Two consequences:

1. Whichever merges second will conflict, or — worse — may *not* conflict. This is the #8623 pattern: two PRs adding a block at different points in one file can merge cleanly and leave a duplicate key, which every test and renderer here silently tolerates while `linkml-reference-validator` aborts on it. Please resolve by **merging both members in and reconciling the counts to 15**, not by taking one side.
2. The member count in whichever merges first will be wrong the moment the other lands.

A third entry, CMD1NN (#10650), belongs in this grouping by MONDO but was deliberately left out precisely because of this contention — its curator flagged that the grouping's rationale enumerates mechanistic classes and would need a prose edit to admit a signalling-kinase lesion. That is worth doing as a single follow-up PR once these two are in, rather than a third concurrent edit to the same file.

I am the author of all three and raised the conflict rather than resolving it unilaterally, since the counts and the rationale prose are a curator call.

PR #10653, opened 2026-09-03
Fix deterministic PR sweep eligibility

cmungall:
The deterministic sweep was enforcing a different policy from the repository owner intent. It excluded `auto/*` branches, treated every assignee as a hold, required the PR head to contain the exact current `main`, and stopped whenever the current `main` check was not already green. Those filters can leave an approved, green PR untouched indefinitely.

This change makes the candidate rule author- and lane-neutral:

- any PR targeting `main` is eligible after the three-day objection window;
- only a human assignee is a hold; known bot/agent assignees remain eligible;
- draft status remains metadata and is handled by the existing just-in-time ready transition;
- approval, conflict state, PR checks, mergeability, age, and the final head-SHA pin remain enforced;
- exact ancestry with the latest `main` and the current-main health circuit are removed;
- the default scan limit rises from 300 to 1,000 because the live backlog is now 365 PRs and the old cap omitted the oldest candidates.

A live dry-run with this branch selects #10100, which is human-authored, unassigned, approved, green, and four days old. No GitHub state was changed by that preview.

This defines candidate policy; #10604 improves how an accepted candidate is submitted to the active merge queue and prevents queue reselection from starving later candidates. Recommended merge order: merge this PR first, then update #10604 onto the resulting controller so queue mechanics implement this candidate universe.

Validation:

- `uv run pytest tests/test_auto_merge_ready_prs.py tests/test_pr_shepherd_workflow.py tests/test_auto_merge_workflow_tokens.py tests/test_pr_shepherd_policy.py -q` — 145 passed
- `uv run ruff check scripts/auto_merge_ready_prs.py tests/test_auto_merge_ready_prs.py tests/test_pr_shepherd_workflow.py` — passed
- pre-commit hooks — passed

Related: #10168.

### DIFF
--- a/docs/explanation/automation-and-agents.md
+++ b/docs/explanation/automation-and-agents.md
@@ -341,9 +341,10 @@ file lists — so those validate the KB files actually entering `main`, in the
 merged context where a co-queued PR may have changed `cache/**` or
 `references_cache/**`. `tests/test_merge_group_ci.py` pins the trigger, the
 filter's event coverage and its v4.0.1 version floor, the no-suppression rule,
-and the never-forced rule for the two file-scoped steps. Enabling an actual
-queue on `main` (branch ruleset, shepherd enqueue behavior) is the remaining,
-separately gated half.
+and the never-forced rule for the two file-scoped steps. The queue ruleset is
+now active on `main` with `ALLGREEN`, five speculative build slots, and a
+one-PR merge limit. Accurate enqueue reporting and queued-PR reselection remain
+controller follow-ups in #10604.
 
 **Merge-integrity verification.** Passing tests are not evidence of correct
 history: in the April 2026 GitHub merge-queue incident (#2034), squash merges


cmungall:
The three documentation blockers are fixed.

- The green-check bullet no longer claims an exact-base comparison exists.
- The `headRefOid` comment now describes only the retained head-SHA pin.
- The operating policy is explicit: the queue is active and supplies latest-main integration testing; if it is disabled, direct loose merging of a verified green PR remains intentional repository-owner policy rather than silently restoring an ancestry gate.
- The docs now say explicitly that the LLM tending lane is narrower because mutation authority and deterministic merge eligibility are separate policies.
- The stale `just auto-merge-preview` description now says “not human-assigned.”

I did not fold in the queue reporting/reselection changes because #10604 owns those mechanics and this PR body already specifies the merge order. I also left the machine-login helper consolidation as a non-blocking follow-up; unknown identities fail closed as human, so drift cannot silently remove a hold.

Validation after the changes: 145 targeted tests pass; Ruff check and format check pass; pre-commit hooks pass.

cmungall:
Adopted the remaining documentation suggestion. The merge-group section now records the active queue configuration (`ALLGREEN`, five speculative build slots, one-PR merge limit) and leaves accurate enqueue reporting/reselection explicitly with #10604. Pre-commit checks pass.

Issue #7780, opened 2026-08-01
Epic #1079 checkbox audit: 23 ticked entries aren't machine-verifiable, 24 duplicated rows (unticked boxes are reliable)


jmcmurry:
Full audit of the #1079 checkbox state against `kb/` content, prompted by a reviewer asking whether an unticked box truly means "not yet in the repo."

## Headline: unticked boxes are reliable; ticked boxes are not machine-verifiable

| Measure | Count |
|---|---|
| Distinct MONDO IDs in the epic | 586 |
| Checkbox rows carrying a MONDO ID | 610 |
| **Duplicated MONDO IDs** (same disease listed twice) | **24** |
| Ticked | 366 |
| Unticked | 220 |
| Unticked whose MONDO ID appears anywhere in `kb/` | 9 — **all false alarms** |
| Ticked with no matching MONDO ID anywhere in `kb/` | 23 |

## 1. Unticked boxes are trustworthy ✅

All 9 unticked-but-mentioned IDs turned out to be **mention-only** — differentials, `notes:` prose, or module cross-references — never a primary `disease_term`. Examples: `MONDO:0859232` appears only as a differential inside `2q37_Microdeletion_Syndrome.yaml`; `MONDO:0000437` only inside `kb/modules/cerebellar_purkinje_degeneration.yaml`.

So the sync correctly ignores incidental mentions, and **an unticked box does mean uncurated.** Good news for anyone picking work off this list.

## 2. The 23 "ticked but absent" are mostly a *mapping* gap, not a curation gap

My first read was that 23 diseases had been silently skipped. That's wrong, and worth correcting before anyone acts on it. Spot-checking four:

| Epic entry | Actually curated as | Entry's `disease_term` |
|---|---|---|
| `MONDO:0019438` AL amyloidosis | `AL_Amyloidosis.yaml` | `MONDO:0017816` primary systemic amyloidosis |
| `MONDO:0018324` adult-onset myasthenia gravis | `Myasthenia_Gravis.yaml` | `MONDO:0009688` myasthenia gravis |
| `MONDO:0011101` peroxisome biogenesis disorder 1B | `Peroxisome_Biogenesis_Disorder.yaml` | `MONDO:0019234` |
| `MONDO:0011445` hereditary spastic paraplegia 11 | `Hereditary_Spastic_Paraplegia.yaml` | `MONDO:0019064` |

[...]

PR #10658, opened 2026-09-03
feat(NEDAMSS): curate IRF2BPL-related neurodevelopmental disorder with regression

### LINKED ISSUE #10639: Curate neurodevelopmental disorder with regression, abnormal movements, loss of speech, and seizures (MONDO:0060759)
Curate a dismech entry for **neurodevelopmental disorder with regression, abnormal movements, loss of speech, and seizures** ([MONDO:0060759](https://monarchinitiative.org/MONDO:0060759)) — NEDAMSS — IRF2BPL-related progressive neurological regression, one of the few neurodevelopmental disorders that is genuinely neurodegenerative rather than static.

**Why this one.** Selected from the priority list in #1079 by filtering its entries through `dashboard/priority.json` and keeping only rows recommended `REVIEW_AGAINST_PARENT` (dashboard score **50.0**) that are absent from `kb/` and carry no prior open or closed issue or PR. Unchecked boxes in #1079 are not a curation queue on their own — roughly three quarters of them are `LUMP_INTO_PARENT` rows, and filing against one of those has previously been closed `wontfix`.

**Coverage check** (`origin/main`, `kb/`). Neither MONDO:0060759 nor `IRF2BPL` appears anywhere in `kb/`. This is a completely uncovered gene and concept.

**What a separate entry has to justify.** **Note this entry is ticked in #1079 but is not actually curated** — the MONDO ID is absent from `kb/` entirely. It is one of the false ticks #7780 flags, which is why it had no claim issue. The curation angle worth getting right is that NEDAMSS is *regressive*: children reach milestones and then lose them, which separates it from the static encephalopathies its MONDO parent groups it with, and is the thing a `pathophysiology` chain should actually model.

**Anchor check:** `just research-disorder` hardcodes an empty `mondo_id` (#10335), so deep-research reports carry no ontology anchor. Run `just preflight-dr <report.md> MONDO:0060759 --strict` and confirm by hand that the report is about IRF2BPL and the right zygosity before curating from it.

Stub: `stubs/Neurodevelopmental_Disorder_With_Regression_Abnormal_Movements_Loss_Of_Speech_And_Seizures.yaml`
Causal gene per MONDO: `hgnc:14282` (IRF2BPL)

AO33:
Curates NEDAMSS — neurodevelopmental disorder with regression, abnormal movements, loss of speech, and seizures ([MONDO:0060759](https://monarchinitiative.org/MONDO:0060759)), caused by IRF2BPL — as a new disease entry, and deletes the stub it consumes.

Closes #10639

## Decision: DISEASE

Both MONDO parents (`MONDO:0002254` syndromic disease, `MONDO:0100500` Mendelian neurodevelopmental disorder) are broad groupings with no curated dismech counterpart, so there is no parent disease for a `has_subtypes` entry. NEDAMSS has one causal gene, one coherent pathograph, and a mechanism that is specific to it rather than shared with the encephalopathies it is grouped with.

## No deep-research report exists for this disease

Worth stating plainly, since it is unusual. Edison/falcon returned HTTP 402 — the provider account is out of credits. The documented `claude_code` fallback was tried twice, once plain and once with `dr_fallback='--fallback'`, and failed identically both times:

```
WARNING - agentapi not found in PATH
ERROR - Error: Claude Code exited with code 1: <no stderr>
```

So this entry was curated from literature retrieved directly through PubMed E-utilities, with all nine references fetched via `just fetch-reference`. Nothing was written to `research/` and no report is described anywhere as existing. That is recorded in the entry `notes` and the history record too, so it stays visible after this PR scrolls away.

## I checked the issue's claims rather than assuming them

- `MONDO:0060759` and `hgnc:14282` both resolve with the exact labels the issue gives (verified with OAK).
- The "absent from `kb/`" claim holds. The only `IRF2BPL` hits on `origin/main` are microarray probe rows inside `kb/hypotheses/Wilsons_Disease/cuproptosis_model/biomni_artifacts/*.tsv` — a gene symbol in a results table, not a causal `genetic:` binding. So this is genuinely one of the false #1079 ticks that #7780 flags.
- The regressive character is real and is the thing worth modelling. Two independent cohorts describe "mostly normal initial psychomotor development followed by severe global neurological regression", the systematic review of 28 patients reports that all of them shared developmental delay/regression, and the one published autopsy describes progressive neurodegeneration with polyglutamine inclusions.

## What the pathograph says

The chain models progression rather than listing symptoms:

de novo truncating variant → NMD escape → cytoplasmic sequestration of wild-type IRF2BPL → loss of transcriptional repression → excess Wnt signaling → progressive neuronal and axonal loss → loss of previously acquired milestones

with two parallel arms feeding the degeneration node: astrocyte mitochondrial failure, and polyQ inclusion formation.

The NMD-escape step is the interesting one and is why `functional_impact_category` is `DOMINANT_NEGATIVE` rather than `LOSS_OF_FUNCTION`. IRF2BPL is intronless, so a premature-stop transcript is not degraded — it is expressed at roughly 1:1 with the wild-type allele, and the truncated protein it makes drags full-length IRF2BPL into cytoplasmic aggregates. That is interference with the remaining wild-type product, not a halved dose.

- **Keratoconus is not curated.** It is reported for this gene, but `PMID:32291157`, the paper devoted to it, returned no retrievable content through the fetcher, so there is no exact quote to attach. The systematic review's more general "ocular disturbances" is curated instead, bound to a correspondingly general HP term.
- **CuATSM is not a treatment.** It rescues neuronal survival and mitochondrial function in patient-derived cells and ameliorates motor defects in flies, but there is no human trial. It is recorded as `RESCUES` links on the models that actually demonstrated it, which is where the evidence sits.
- **Most phenotypes carry no `frequency`.** Only regression has an explicit denominator in a source ("all patients"), so only it is `OBLIGATE`. The rest are described qualitatively and I would have been inventing a band.

## Two errors caught during curation

Both are worth flagging because they are the kind that pass silently:

- `GO:0070997` "neuron death" is **obsolete** — as is the entire neuron-death branch. Replaced with `GO:1990535` neuron projection maintenance.
- I initially bound the developmental-support treatment to `NCIT:C159273` Speech Language Therapy. GeneReviews never mentions speech therapy anywhere in its management section. Rebound to `NCIT:C17874` Educational Intervention, which is what the source actually recommends. `check-snippet-length` is what surfaced this, by rejecting the four-word snippet I had hung on it.

## Validation

All run, all passing:

```
just validate                  -> All validations passed; 62/62 snippets verified
just validate-terms            -> Validation passed
just count-verified-snippets   -> 62/62 verified against cached references
just check-duplicate-keys      -> OK: no duplicate mapping keys
just check-entity-refs         -> OK: entity references resolve
just check-causal-targets      -> OK: no new broken pathograph targets
just check-qualifier-terms     -> OK (0 qualifier terms used in this entry)
just check-enum-values         -> OK: no out-of-enum values
just check-snippet-length      -> OK: no new snippets under 5 words
just check-title-snippets      -> OK: no new title-quoting snippets
just check-snippet-grading     -> OK: no new snippet-grading divergences
just check-folded-hyphens      -> OK: no new folded-scalar hyphen splits
just validate-disorders        -> All 1 disorder file(s) passed validation
just validate-history          -> No issues found
```

Cache churn is limited to the three CURIEs this entry introduces (`MONDO:0060759`, `hgnc:14282`, and the matching `diseaseterm` enum row). No `pages/`, `dashboard/`, or `app/` files are touched.

### DIFF
--- a/history/disorders/IRF2BPL-Related_Neurodevelopmental_Disorder_with_Regression/2026-09-03T041133Z-claude-code-36cb94.yaml
+++ b/history/disorders/IRF2BPL-Related_Neurodevelopmental_Disorder_with_Regression/2026-09-03T041133Z-claude-code-36cb94.yaml
@@ -1,0 +1,34 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: IRF2BPL-Related_Neurodevelopmental_Disorder_with_Regression
+  path: kb/disorders/IRF2BPL-Related_Neurodevelopmental_Disorder_with_Regression.yaml
+session:
+  id: 2026-09-03T041133Z-claude-code-36cb94
+  timestamp: '2026-09-03T04:11:33Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10658
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - references
+  - phenotypes
+  - pathophysiology
+  - treatments
+  - prevalence
+  - progression
+  - mappings
+  - evidence
+  summary: 'Review response: references block, published frequency denominators, node
+    split'
+  details: |
+    Addressed the CHANGES_REQUESTED review on PR 10658. Added a top-level references block with cache-verbatim titles and tags: [GeneReviews] on PMID:39571061. Mined the committed full texts for published denominators and rewrote the notes paragraph that wrongly said the sources were only qualitative: FREQUENT bands for epilepsy (21/28), spasticity (14/28), dystonia (12/28), ocular disturbance (12/28) and ataxia (10/28), OCCASIONAL for dysmorphic features (5/28), dysphagia (3/11), dysarthria (2/11) and nystagmus (2/11). Added dysphagia, dysarthria, nystagmus, mild dysmorphic features and delayed puberty as phenotypes. Split the bundled astrocyte node into Astrocyte Mitochondrial Respiratory Dysfunction and Loss of Astrocytic Trophic Support for Neurons, re-pointing the CuATSM RESTORED readout at the mitochondrial step and adding the patient muscle ETC deficiency evidence there. Regraded the intronless-gene NMD-escape evidence from HUMAN_CLINICAL to COMPUTATIONAL. Added prevalence (CASES_IN_LITERATURE), a two-phase progression block, the OMIM:618088 mapping, glycogen inclusions on the aggregate node, the coiled-coil variant clustering in genetic notes, and the unmined GeneReviews management items (ophthalmology, endocrine, gastric dysmotility, family and social work support). Replaced the Loss of speech evidence item that quoted the syndrome acronym expansion. Validated: just validate, validate-terms, count-verified-snippets (92/92), check-duplicate-keys, check-entity-refs, check-causal-targets, check-qualifier-terms, check-enum-values, check-snippet-grading, check-snippet-length, check-title-snippets, check-folded-hyphens, check-reference-titles, and validate-disorders.
--- a/kb/disorders/IRF2BPL-Related_Neurodevelopmental_Disorder_with_Regression.yaml
+++ b/kb/disorders/IRF2BPL-Related_Neurodevelopmental_Disorder_with_Regression.yaml
@@ -43,6 +43,36 @@ disease_term:
   term:
     id: MONDO:0060759
     label: neurodevelopmental disorder with regression, abnormal movements, loss of speech, and seizures
+mappings:
+  mondo_mappings:
+    - term:
+        id: MONDO:0060759
+        label: neurodevelopmental disorder with regression, abnormal movements, loss of speech, and seizures
+      mapping_predicate: skos:exactMatch
+      mapping_source: OMIM:618088
+      mapping_justification: >-
+        NEDAMSS is catalogued in OMIM as 618088, the number quoted in the
+        neuropathology report that named the syndrome. MONDO:0060759 is the
+        exact anchor for that OMIM concept and is the term this entry curates.
+references:
+  - reference: PMID:39571061
+    title: IRF2BPL-Related Disorder.
+    tags:
+      - GeneReviews
+  - reference: PMID:30057031
+    title: IRF2BPL Is Associated with Neurological Phenotypes.
+  - reference: PMID:30166628
+    title: De novo truncating variants in the intronless IRF2BPL are responsible for developmental epileptic encephalopathy.
+  - reference: PMID:35044823
+    title: Loss of IRF2BPL impairs neuronal maintenance through excess Wnt signaling.
+  - reference: PMID:35525099
+    title: "Neurological phenomenology of the IRF2BPL mutation syndrome: Analysis of a new case and systematic review of the literature."
+  - reference: PMID:36476864
+    title: Mechanisms of IRF2BPL-related disorders and identification of a potential therapeutic strategy.
+  - reference: PMID:37114479
+    title: IRF2BPL as a novel causative gene for progressive myoclonus epilepsy.
+  - reference: PMID:39224955
+    title: "IRF2BPL-Related Disorder, Causing Neurodevelopmental Disorder with Regression, Abnormal Movements, Loss of Speech and Seizures (NEDAMSS) Is Characterized by Pathology Consistent with DRPLA."
 notes: >-
   Lump/split. Curated as a standalone DISEASE. Its two MONDO parents
   (MONDO:0002254 syndromic disease, MONDO:0100500 Mendelian neurodevelopmental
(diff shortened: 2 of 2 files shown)


AO33:
All five blocking items are fixed, plus the non-blocking suggestions. Pushed as `58f6c99`.

You were right that the interesting content was sitting unread in the two committed full texts — the entry now carries frequency bands on nine phenotypes where it previously had one, and four new phenotypes that came out of those same two papers with no new fetching.

**Item 1 — GeneReviews tag.** Added a top-level `references:` block for all eight PMIDs, with `tags: [GeneReviews]` on `PMID:39571061`. Titles are copied verbatim from each `references_cache/` frontmatter. Worth noting for anyone automating this later: `scripts/tag_references.py` would not have found this one, because it detects GeneReviews by looking for the literal string "GeneReviews" in the cached body, and this cache file does not contain it. The tell is the editor list (Adam MP / Mirzaa GM / Wallace SE / Amemiya A) and `year: '1993'`.

**Item 2 — the frequency claim.** The `notes` paragraph is rewritten. Bands added: epilepsy 21/28, spasticity 14/28, dystonia 12/28, ocular disturbance 12/28, ataxia 10/28 — all `FREQUENT` — and dysmorphic features 5/28 as `OCCASIONAL`. Regression keeps `OBLIGATE` and now also cites the stronger sentence you pointed at, "present in all 28 patients (100%)".

**Item 3 — missing phenotypes.** Dysphagia 3/11 (`HP:0002015`, with the second independent mention from `PMID:37114479` as a separate evidence item), dysarthria 2/11 (`HP:0001260`), nystagmus 2/11 (`HP:0000639`), mild dysmorphic features 5/28, and delayed puberty (`HP:0000823`). Dysarthria and nystagmus are wired downstream of the cerebellar atrophy node, which is how the source frames them; dysphagia hangs off the degeneration node.

Two judgement calls there. Mild dysmorphic features is left **unbound** — the source counts facial weakness, a wider inter-nipple distance and joint laxity together, and no single HP term covers that set, so a facial-dysmorphism term would assert something narrower than what was counted. Delayed puberty carries **no frequency band**: GeneReviews drives management for it but publishes no denominator, and both its evidence items are marked `directness: INDIRECT`, since a surveillance recommendation is one inference step away from an observed frequency. It is curated because you are right that it matters — IRF2BPL is EAP1, a hypothalamic regulator of pubertal onset — and the endocrine management is now a treatment too.

**Item 4 — the bundled node.** Split into:

```
Astrocyte Mitochondrial Respiratory Dysfunction
    -> Loss of Astrocytic Trophic Support for Neurons
        -> Progressive Neuronal and Axonal Loss
```

The experimental model now declares two links instead of one. The CuATSM `RESTORED` readout moved to the mitochondrial node, which is the step it acts on, and the neuronal-survival `DECREASED` readout stays with the trophic-support node. The mitochondrial node also picked up the patient muscle evidence you flagged — the electron transport chain deficiencies in patient 1 — so that arm no longer rests entirely on reprogrammed astrocytes. `just check-causal-targets` is clean; nothing referenced the old node name after the split.

**Item 5 — the misgrade.** `HUMAN_CLINICAL` -> `COMPUTATIONAL` on the intronless-genes sentence, with the reason recorded in the `explanation`.

**Also taken:** `prevalence:` (two `CASES_IN_LITERATURE` records, 28 patients to 2022 and 31 to 2023, `prevalence_class: NOT_YET_DOCUMENTED` because nobody has published a rate); a two-phase `progression:` block carrying the full onset distribution; the `OMIM:618088` mapping; the skin-biopsy glycogen inclusions on the aggregate node; coiled-coil variant clustering in `genetic[].notes` with its own evidence item; the four unmined GeneReviews management items (ophthalmology, endocrine, gastric dysmotility, family and social work support); and the `Loss of speech` evidence item that quoted the acronym expansion is replaced with the speech-regression count from the 28-patient series.

## One place I did not do what you suggested, and why

Four of the denominators are stated in the phenotype `description` rather than quoted in a snippet. This is a snippet-matching limitation, not a decision about the evidence.

The sentence that carries them reads:

> including epilepsy (21/28 patients) [ 2 , 6 , 7 , 8 , 10 , 11 , 13 ], dystonia (12/28 patients) [ 2 , 6 , 7 , 9 , 11 , 12 ], ataxia (10/28 patients) [ 6 , 7 , 10 , 12 ], spasticity (14/28 patients) [ 2 , 6 – 9 , 12 ] and a variety of ocular distur-\nbances (12/28)

Each fraction is walled off by an inline numeric citation marker. The validator strips bracketed spans out of a quote before matching but leaves them in the cached text, so a span crossing one can never match — I tried it, and `count-verified-snippets` says so explicitly: "the snippet matches the cached text exactly once [ 2 , 6 , 7 , 8 , 10 , 11 , 13 ] ... are kept". That leaves each fraction as a 3-to-4-word island, which `check-snippet-length` rejects as non-propositional. And the ocular fraction has the hyphen wrap on top of that, exactly as you warned.

So where the paper states the same numbers a second time in an uninterrupted sentence, I cite that instead:

| Phenotype | What is quoted |
|---|---|
| Epilepsy | the Table 1 caption, "patients who did (n. 21) or did not (n.7) had epilepsy" |
| Spasticity | the sex breakdown, "present in 10/14 men and 4/14 women" — sums to 14 |
| Dystonia | the distribution sub-counts, 3/12 and 4/12 — establishes the denominator of 12 |
| Ataxia | the qualitative five-domain sentence only; no second statement of 10/28 exists |
| Ocular | the authors' own "relatively high frequency of impairment in at least five main neurological domains" |

Each `explanation` says which fraction it is standing in for and why it is not quoted directly, and the `notes` block records the general point once. I would rather have the number in prose with the reason attached than put `ocular distur - bances (12/28)` on the rendered page as if that were how the source reads.

## Validation

`validate`, `validate-terms`, `count-verified-snippets` (92/92, up from 62/62), `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-reference-titles`, `validate-history`, and `just validate-disorders` all pass. Every new HP CURIE was resolved against the ontology rather than written from memory. No baseline file was touched, and the diff is the one entry plus a history record — no cache churn.

PR #10659, opened 2026-09-03
Add model-scale auditing and typed divergence taxonomy for computational models

cmungall:
## Summary

Records what a computational model can and cannot support, in two linked pieces. `ModelMechanismLink.model_scale` says which biological scale a model actually *observes*, so the gap against its target node's `biological_scale` becomes derivable rather than buried in prose. `ModelMechanismLink.divergences` types the rest of the caveat, which no scale comparison can reach.

The problem both address: curators reliably *write* the caveat — 80% of model→mechanism links in the KB carry a `limitations` string — but it was unstructured, and `fidelity` compressed every kind of concern into one tier, so `LOW` never said *which* problem it was.

## Schema

**`model_scale`** on `ModelMechanismLink`, ranged on the existing `BiologicalScaleEnum`. The comparison against the target's own scale is **derived, never stored** (`just model-scale-audit`), and is directional: a model *below* its target's scale is extrapolating upward and cannot observe the outcome it is cited for; a model *above* it contains that scale and is unremarkable.

**`divergences`** on `ModelMechanismLink` — multivalued `ModelDivergence`, each with a required type, a required curator explanation, an optional `materiality`, and optional evidence.

**`ModelDivergenceTypeEnum`** — 11 values:

```
BOUNDARY_OMISSION       PROXY_QUANTITY          CALIBRATION_PROVENANCE
CAUSE_UNREPRESENTED     STRUCTURAL_IDEALIZATION TEMPORAL_SCOPE
CONTESTED_ASSUMPTION    SCALE_EXTRAPOLATION     SPECIES_MISMATCH
POPULATION_MISMATCH     OTHER
```

**`ModelDivergenceMaterialityEnum`** — `INVALIDATING` / `QUALIFYING` / `IMMATERIAL` / `UNKNOWN`. Per-divergence, where `fidelity` is per-link.

## The taxonomy is evidenced, not invented

All 50 computational-model `limitations` strings in the KB were read and clustered, with the animal and NAM sets keyword-probed to establish which kinds are shared. The modality profiles differ sharply — the animal set is dominated by species divergence and allele mismatch, the computational set by calibration provenance, absent dynamics and proxy quantities, with species divergence nearly absent. Every enum value has quoted KB evidence in `docs/superpowers/specs/2026-09-02-model-divergence-taxonomy.md`, whose coverage table is dated as the derivation snapshot rather than a live count.

## Why `PROXY_QUANTITY` is the value that justifies the second piece

A model can report a **different quantity at the same scale**, which no scale comparison can see. The Fanconi anemia FA/BRCA link is the demonstration: model and target are both `MOLECULAR`, so the scale comparison reports no gap, yet the link is `PARTIALLY_RECAPITULATES` because homologous recombination appears as network components inside a crosslink-repair model — what is reported is their activation state, where the node describes recombination fidelity.

## Prior art, and the limits of borrowing it

Design decisions 3b and 3c record the choices; `docs/explanation/model-credibility.md` is the reader-facing explanation. Standards are cited from the schema itself via `see_also`, not only in prose:

| Element | References |
|---|---|
| `ModelMechanismLink`, `ModelDivergence` | ten rules (PMID:32993675), ASME V&V 40, FDA credibility guidance |
| `divergences` slot, `ModelDivergenceTypeEnum` | ten rules |
| `ModelDivergenceMaterialityEnum` | ASME V&V 40, FDA guidance |

ASME V&V 40 and the FDA guidance were written for regulatory submissions of physics-based device models, so the doc separates what transfers — assess a model for a stated use, argue applicability separately from validity, grade evidence by how much the answer matters — from what does not: V&V plans, uncertainty quantification, numeric credibility goals, and any claim about numerical accuracy, since dismech annotates published models from their papers rather than building them. The domain-general anchor is the ten rules of credible practice in healthcare modeling, which reach the same conclusions with no device framing; its rule 1 is the link, rule 3 is `materiality`, rule 4 is `divergences`.

## Knowledge base

This PR adds `biological_scale` to 9 pathophysiology nodes and contributes 10 `modeled_mechanisms` links carrying 26 typed divergences, across **4 new Boolean models plus a repair of one that already existed**:

| Entry | Model | PMID | |
|---|---|---|---|
| Rheumatoid_Arthritis | RA-FLS large-scale Boolean model | 37454172 | new |
| Parkinsons_Disease | PD map cohort-specific probabilistic Boolean models | 39429779 | new |
| COVID-19 | C19DMap SBML-qual collection (FAIRDOMHub 714) | 34664389 | new |
| COVID-19 | Type 1 interferon signalling Boolean model | 38414974 | new |
| Fanconi_Anemia | FA/BRCA pathway Boolean model | 22267503 | already present; was an unlinked citation, now carries `modeled_mechanisms` |

Five reference cache entries added. `main` separately carries two other Boolean models (`cellular_senescence`, `mitochondrial_dysfunction`) that this PR does not touch, so the post-merge KB total is 7.

## Tooling and tests

`scripts/model_scale_audit.py` / `just model-scale-audit`, advisory by default. `--strict` fails on an upward-extrapolating link carrying neither `limitations` nor a typed divergence, and on a `SCALE_EXTRAPOLATION` divergence contradicted by the scale slots.

Five tests: `test_model_scale_values_are_valid`, `test_upward_extrapolating_links_are_caveated`, `test_model_divergences_are_typed_and_explained`, `test_scale_extrapolation_divergence_agrees_with_scales`, `test_audit_scale_order_matches_schema`. The existing `FAILS_TO_RECAPITULATE` test now accepts a typed divergence wherever it accepted `limitations`.

## Scope

`divergences` is populated on computational models only. The bulk of animal and NAM links carry prose `limitations` ready to be structured, and the taxonomy extends to them unchanged; extending would likely add `SUPRAPHYSIOLOGICAL_EXPRESSION` and `INCOMPLETE_PHENOTYPE`, each already visible in the animal set.

`docs/schema/` was not regenerated: the current `gen-doc` emits 735 files against the 693 committed, so regenerating would bury this change in unrelated generator drift.

https://claude.ai/code/session_013cddYUVAPcsBfdKYL6qEUs


cmungall:
**The enum values in the PR description are wrong, and a curator following it would hit validation failures.** It lists seven divergence types including `DYNAMICS_ABSENT` and `ALLELE_MISMATCH` — neither of those exists. There are eleven values, and the nearest real ones are `TEMPORAL_SCOPE` and `SPECIES_MISMATCH` / `POPULATION_MISMATCH`.

The actual `ModelDivergenceTypeEnum`:

```
BOUNDARY_OMISSION       PROXY_QUANTITY          CALIBRATION_PROVENANCE
CAUSE_UNREPRESENTED     STRUCTURAL_IDEALIZATION TEMPORAL_SCOPE
CONTESTED_ASSUMPTION    SCALE_EXTRAPOLATION     SPECIES_MISMATCH
POPULATION_MISMATCH     OTHER
```

Other corrections to the description:

| Description says | Actual |
|---|---|
| 3 new tests, incl. `test_model_scale_gap_audit`, `test_model_scale_gap_is_documented` | 4: `test_model_scale_values_are_valid`, `test_upward_extrapolating_links_are_caveated`, `test_model_divergences_are_typed_and_explained`, `test_scale_extrapolation_divergence_agrees_with_scales` |
| `biological_scale` on 11 nodes | 9 |
| 3 flagship model examples | 5 Boolean models, carrying 26 divergences |
| "four new reference cache entries" (then lists five) | 5 |

Two things the description omits entirely: a second enum, `ModelDivergenceMaterialityEnum` (`INVALIDATING` / `QUALIFYING` / `IMMATERIAL` / `UNKNOWN`), and the standards citations described below.

## What the description predates

The last two commits landed after it was written.

Design decisions 3a and 3b are now **enacted** rather than proposed, following maintainer approval.

The credibility standards are now cited **from the schema itself**, via `see_also` on the elements each one informed, rather than only in prose:

- `ModelMechanismLink`, `ModelDivergence` — the ten rules of credible practice in healthcare modeling (PMID:32993675), ASME V&V 40, and the FDA computational-model credibility guidance
- `divergences` slot — the ten rules, whose rule 4 is "list limitations explicitly"
- `ModelDivergenceTypeEnum` — the ten rules
- `ModelDivergenceMaterialityEnum` — the two device standards, whose risk-informed grading it mirrors

## On borrowing a medical-device framework

ASME V&V 40 and the FDA guidance were written for regulatory submissions of physics-based device models, so `docs/explanation/model-credibility.md` separates what carries over from what does not.

Structural ideas carry over, because they are claims about the logic of model-based inference rather than about devices: assess a model for a stated use rather than in the abstract; argue applicability separately from validity; grade how much evidence you need by how much the answer matters.

The procedural apparatus does not, and is not imitated — no V&V plan, no uncertainty quantification, no numeric credibility goals. dismech annotates published models it did not build, usually from the paper alone, so it makes no claim about numerical accuracy.

The domain-general anchor is the ten rules paper, which reaches the same conclusions with no device framing at all and explicitly covers "research, diagnosis, risk assessment, prevention, therapy, rehabilitation, surgery, intervention design, and regulation". Its rule 1 ("define context clearly") is the link, rule 3 ("evaluate within context") is `materiality`, and rule 4 is `divergences`.

## Why `PROXY_QUANTITY` is the value that matters

`model_scale` gives a derivable gap — is the model observing below its target's scale. It cannot see a model reporting a *different quantity at the same scale*, and that turns out to be the commonest reason a link is `PARTIALLY_RECAPITULATES`.

The Fanconi anemia FA/BRCA link is the demonstration: model and target are both `MOLECULAR`, so the scale comparison reports no gap, yet homologous recombination appears as network components inside a crosslink-repair model — what the model reports is their activation state, where the node describes recombination fidelity.

## Validation

`linkml-lint` reports 592 problems before and after the schema changes, all 15 new ones being the same `standard_naming` warning every enum in this schema already carries. `linkml-validate` clean on all four affected entries; 4,273 schema tests and ~10,000 model-link tests pass; `just model-scale-audit --strict`, folded-hyphen, duplicate-key and entity-ref gates clean; 785/792 evidence snippets verified against `references_cache` with no failures.

Both new tests were checked for vacuity by injecting a defect and confirming they fail, then reverting.

## Not done

`divergences` is populated on computational models only. Roughly 780 animal and NAM links already carry prose `limitations` ready to be structured, and the taxonomy extends to them unchanged — `BOUNDARY_OMISSION`, `PROXY_QUANTITY`, `CALIBRATION_PROVENANCE`, `POPULATION_MISMATCH` and `SPECIES_MISMATCH` all apply as written. Extending would likely add `SUPRAPHYSIOLOGICAL_EXPRESSION` and `INCOMPLETE_PHENOTYPE`, each already visible in the animal set at 20–53 keyword hits. That is a much larger curation pass and is left as documented scope.

Schema docs under `docs/schema/` were also not regenerated: the current `gen-doc` emits 735 files against the 693 committed, so regenerating would bury this change in unrelated generator drift. That looks like a pre-existing issue worth its own ticket.

---

cmungall:
Both blockers fixed, and the three suggestions too. Rebased onto `origin/main` first, which is what made both problems visible — the diff against the new base is clean, no unrelated churn.

## Retired `supports: PARTIAL`

You were right that all three fit `SUPPORT` + `directness: INDIRECT`, and it is worth saying *why* they fit: each explanation was already describing an inference step rather than partial support, so the old value was carrying the wrong meaning even before it was retired. I rewrote the three explanations to name the inference rather than the retired value:

- **COVID-19** — the quote asserts molecular-level blockade of interferon signalling; tissue pathology follows by inference. That is also why fidelity there is `LOW`.
- **Fanconi anemia** — the quote asserts which repair routes become active; impaired homologous recombination follows by inference. The model reports pathway activation state, not recombination fidelity.
- **Parkinson's** — the quote asserts subtype-specific behaviour of dopamine-transcription pathways; striatal dopamine deficiency follows across transcription, synthesis and terminal survival.

For the record, `hence PARTIAL` still appears in 117 other files as leftover explanation prose from the original migration, with `supports` already corrected. Not touched here.

## Duplicate `### 3a`

Cancer keeps `3a` since CLAUDE.md cites it by number; the two decisions here became `3b` and `3c`.

One thing beyond what you asked: renumbering alone would have left the register reading 3b, 3c, 3a, because the cancer section sits after mine in the file. I moved it above them so the numbering runs in order. That is a pure relocation — I verified the moved block is byte-identical to `main`'s copy rather than trusting the eyeball. Cross-references updated in `design-decisions.md`, `model-credibility.md` and the plan doc, which also still called both decisions "proposed".

## Suggestions

**Hand-duplicated lists** — both `BiologicalScaleEnum` and `ModelDivergenceTypeEnum` copies in `tests/test_data.py` now derive from the schema through one cached helper, so a value added to the schema cannot leave a test's list stale.

I kept the audit script's own `SCALE_ORDER`, so `just model-scale-audit` does not load a SchemaView on every run, and added `test_audit_scale_order_matches_schema` to pin that copy — membership *and* order, since the gap arithmetic is ordinal and a silent reorder would invert results without failing anything.

That refactor introduced a bug I caught on the first run: my derived local shadowed the existing `scales` dict inside two tests, failing 419 cases with `'tuple' object has no attribute 'get'`. Renamed to `scale_ladder`.

**Leaked handle** in `collect()` — now a context manager.

**Stale PR body** — rewritten. It had listed seven divergence types including `DYNAMICS_ABSENT` and `ALLELE_MISMATCH`, neither of which exists, so anyone curating from it would have hit validation failures.

## Re-verification

Both new tests were re-checked for vacuity *after* the refactor, since deriving the lists could have made them silently pass. Injecting a contradictory `SCALE_EXTRAPOLATION` divergence fails `test_scale_extrapolation_divergence_agrees_with_scales`; reordering `SCALE_ORDER` fails the new guard. Both reverted.

On the rebased schema: `linkml-validate` clean on all four entries, 785/792 snippets verified with no failures, duplicate-key / entity-ref / folded-hyphen / snippet-grading gates clean, 4116 schema tests and 10799 model-link tests pass, `just model-scale-audit --strict` passes.

---

cmungall:
**`test (3.13)` failed on the curation-scope guard, not on a test.** The guard fires because this PR touches `kb/` *and* infra (`src/dismech/schema/`, `tests/`, `scripts/`, `docs/`, `project.justfile`, `mkdocs.yml`), and its message prescribes the remedy: add `scope-override` if that is intentional. I have added the label.

The reason it is intentional is that **the two halves cannot be separated.** The KB entries use `model_scale` and `divergences`, which do not exist in `main`'s schema. Validating this PR's `Fanconi_Anemia.yaml` against `main`'s schema:

```
[ERROR] Additional properties are not allowed ('divergences', 'model_scale' were
        unexpected) in /computational_models/0/modeled_mechanisms/0
[ERROR] Additional properties are not allowed ('divergences', 'model_scale' were
        unexpected) in /computational_models/0/modeled_mechanisms/1
```

So a curation-only PR would be invalid, and a schema-only PR would add two slots with no worked example and no test coverage of real data. Splitting them means merging a knowingly-red PR first.

Flagging it plainly since a label that turns a check green deserves the scrutiny: this is the guard's own declared affordance for a mixed PR, not a way around a failing test, and it is trivially reversible — remove the label and the check goes red again. If you would rather have this as two PRs, say so and I will split it, landing the schema first and the curation second.

The scope guard is the only thing that was red. `claude-review` is running on the same head.

---

cmungall:
Took up the review note on the parenthetical count, and that push doubles as the fix for the stale `test (3.13)`.

**The count.** You were right — I re-measured and got exactly your numbers: 1,662 links on `main` at 3b1ba68c3, 1,672 on this branch, 80% carrying `limitations` in both. My "1,646" was taken against an earlier `main` that had since moved. Corrected, and I pinned the commit next to the date so the next reader can reproduce the figure instead of guessing which `main` was meant. The dated table in section 1 is untouched — that one is the derivation input for the clustering, not a live count.

**The stale check.** `test (3.13)` was red for one reason only: the curation-scope guard, which refuses a PR that touches `kb/` alongside infra files unless the PR carries `scope-override`. That label is on the PR now, but the failing run finished at 17:39:51 and I added the label around 17:44, so the red result predates its own fix. Labelling does not re-trigger the workflow — `main.yaml` uses a bare `pull_request:` trigger, so it fires on opened/synchronize/reopened only — and my attempts to re-run it directly got `403 Resource not accessible by integration` on both `rerun` and `rerun-failed-jobs`.

The guard reads labels live at run time (`gh pr view --json labels`), not from the event payload, so any fresh run picks up the override. The push above is that fresh run.

I held off pushing purely to kick CI, since an empty commit for that purpose is off the table and it would cost the approval for nothing. Having a real correction to make changed the arithmetic — the fix and the re-run come in the same push.

The approval will have dropped with the push (`dismiss_stale_reviews` is on). The diff since the approved commit is three lines of prose in one markdown file; nothing in `kb/`, the schema, the tooling or the tests moved. Re-requesting review.

---

cmungall:
Correcting my last comment: the scope guard was the **first** failure, not the only one. With `scope-override` in place CI got past it and straight into `ruff check .`, which fails on two findings in this PR's own code. Fixed in `b554216d9`.

Neither earlier run had ever reached the lint step — the first stopped at the scope guard, and the second was the first to get past it — so these were masked from the start rather than introduced late. Worth noting for anyone reading the run history: a green-looking early CI stage here means "did not get there yet", not "passed".

**`RUF100` in `scripts/model_scale_audit.py`** — the inline `# noqa: E402` on the post-`sys.path` import is redundant. `pyproject.toml` already carries a `"scripts/*" = ["E402"]` per-file ignore for exactly this idiom, added when ruff 0.15 started reporting E402 by default. Mine was the only script in the directory carrying an inline directive; the other nine rely on the per-file ignore. Dropped it to match them.

**`UP033` in `tests/test_data.py`** — `@lru_cache(maxsize=None)` on `_enum_values` is spelled `@cache`. The three pre-existing `@lru_cache(maxsize=1)` decorators aren't flagged and I left them alone, so the `lru_cache` import stays alongside the new `cache` one.

Both are behaviour-preserving.

**Validation.** Since two failures had now been masked in sequence, I ran the offline gates that neither run reached, not just the two I broke:

| Check | Result |
|---|---|
| `uvx ruff check .` (exact CI command) | ✅ All checks passed |
| `model_scale_audit.py --strict` | ✅ 5 `MODEL_BELOW_TARGET`, 0 uncaveated, 10 links with typed divergences |
| `pytest -k "model_scale or divergence or upward_extrapolating or scale_order or recapitulate or enum_values"` | ✅ 13,524 passed |
| `check_folded_hyphens` / `check_environmental_evidence` | ✅ |
| `check_duplicate_yaml_keys` (4,274 files) / `check_enum_values` (2,912) | ✅ |
| `check_entity_refs` (2,831) / `check_causal_targets` | ✅ no new dangling targets |
| `check_qualifier_terms` (228 labels) | ✅ |
| `term_cache_integrity` (18 term + 24 enum caches) | ✅ |
| `reference_cache_frontmatter` | ✅ |
| `check_snippet_length` / `check_title_snippets` / `check_reference_titles`, all `--against-ref origin/main` | ✅ no new violations |
| `not4curation_audit` | ✅ |
| `dismech-stubs check` | ✅ 1,337 stubs, 0 errors |

The remaining CI steps I could not run locally are the ones needing `just` and Node (`test-extension`, `test-schema`, `test-examples`, `test-search`, `test-kb`); the schema is untouched by these two commits, and the reviewer's own run covered the KB sweep.

---

cmungall:
`test (3.13)` is green on `b554216d9` — first time it has run to completion on this branch. Two failures had been masked in sequence (the scope guard, then ruff), so I took the whole set of CI lanes locally rather than trusting that the rest would pass:

| Lane | Result |
|---|---|
| `pytest -m "not kb_data"` | 11,676 passed |
| `pytest -m "kb_data" -n auto` | 50,604 passed |
| `validate-history-all` | 8,225 records, no issues |
| `test-extension` / `test-search` | 17 / 73 |
| `gen-project` schema smoke test | pass |

Two of those initially failed with `FileNotFoundError: 'just'` in `tests/test_validate_disorder_hook.py`. That was my container missing the binary, not a real failure — I installed `just` and re-ran the file: 13/13. Similarly `test-search` first failed on `ERR_MODULE_NOT_FOUND: minisearch` until `npm ci`. Worth noting the Node lanes do run for this PR: it touches `project.justfile`, which matches both the `app` and `extension` path filters.

**On the `hence PARTIAL` note — the premise is off, but it found something real.** That string is not the retired `EvidenceItemSupportEnum` value. The full parenthetical is:

> Homologous Recombination Impairment (PARTIALLY_RECAPITULATES — the quoted claim concerns compensatory pathway activation, not recombination fidelity, hence PARTIAL)

It opens with `PARTIALLY_RECAPITULATES`, a live `ModelRelationshipEnum` value, and "hence PARTIAL" abbreviates *that* — partial recapitulation. Different enum, different class, and the live one.

I still changed it, in `bb4b3b88a`, because the misreading is the finding. CLAUDE.md records a cleanup backlog of ~3,600 evidence items whose prose names the retired grade, and notes that prose is the one thing no gate catches. A sweep grepping for it would hit this record as a false positive exactly the way this review did. Spelling the phrase out removes that. Nothing the record claims changes.

The append-only argument doesn't apply here either: append-only protects records already in `main` from being rewritten, and this one is added by this PR and has never been in `main`.

Verified before pushing: `validate-history` on the record, `check-folded-hyphens`, `check-duplicate-keys` (4,274 files), `ruff` — all clean. This dismisses the approval again, for which I'm sorry; the diff is one phrase in one history record's `details`.

---

PR #10678, opened 2026-09-03
Add immune antigen representation audit script and gap analysis

mellybelly:
## Summary

dismech has no way to say what an antigen is. The word "antigen" does not appear once in any of the 21 schema files under `src/dismech/schema/` — there is no antigen class, no antigen slot, no enum value. Every antigen in the knowledge base is a sentence.

The cost, measured over the objects that could actually record it: only three classes own a cell-type slot (`Pathophysiology`, `Biochemical`, `ExperimentalModel`). Of the 3,122 objects that name an antigen across a 573-entry immune cohort, **428 are instances of one of those classes, and 274 of them (64.0%) name no B, T or antigen-presenting lineage**. So for about two thirds of the antigen mentions that *could* say which lymphocyte sees the antigen, the answer exists only in prose.

This PR is a census, not a fix. No schema file and no KB entry is changed.

## What the audit found

**Autoantigen identity.** 116 autoantibody rows in `biochemical` across 66 entries; 11 are bound to an ontology term. Where a binding exists it points at an HP *seropositivity* phenotype, which says the patient tested positive but does not identify the antigen as a molecule — so it cannot be joined to the gene that encodes it or to a T-cell response against the same protein.

**The `genetic` block is being used as an antigen slot, with a written-in disclaimer.** `GeneDiseaseRelationshipEnum` has ten values and none means "the protein this gene encodes is the target of the autoimmune response". Two entries record the antigen gene anyway and work around it in free text — `Anti-GBM_Disease` carries `association: Autoantigen target (not a causal germline mutation)` with `relationship_type` left empty. `Membranous_Nephropathy` shows what this invites: PLA2R1 is filed as a GWAS susceptibility locus, its role as *the* autoantigen is dropped, and the same protein reappears in `biochemical` as an unbound `Anti-PLA2R autoantibodies` string with no link between the two.

**The free-text form cannot tell an autoantigen from a drug target.** Ranking captured `anti-X` strings turns up genuine autoantigens (NMDAR, AChR, AQP4, Ro, La), therapeutic targets that leaked in through evidence prose (TNF, CD20), an anti-streptococcal serology (`anti-DNase B`), and anti-Müllerian hormone, which is not an antibody target at all. Nothing in the YAML separates them.

**HLA — and why almost none of it is backfillable.** 100 `genetic` rows name HLA; 59 are bound, 41 are not. The 41 look like a backlog and are not one: **none of them names an HGNC gene symbol.** All 41 are serotypes (`HLA-DQ2`, `HLA-B27`), alleles (`HLA-DRB1*03:01`), haplotypes (`HLA-DR3-DQ2`) or regions — things no single gene identifies. Six say so in their own `notes`, including both entries where HLA restriction *is* the mechanism. The real gap is that the KB has nowhere to put a serotype or an allele at all, and no peptide is anywhere linked to the allele that presents it.

**Surface markers.** CD4 and CD8 are the best represented, but only as a side effect of riding inside CL labels like `CD8-positive, alpha-beta T cell` (~24% in an identity field). CD3 (2.1%), CD20 (2.9%) and CD19 (3.8%) have no CL term to ride on and are almost entirely prose — across 79 entries that mention rituximab.

## Worked exemplars

`Celiac_Disease` is the textbook case of B and T cells seeing different antigens: B cells target tissue transglutaminase, T cells target deamidated gliadin peptides on HLA-DQ2/DQ8. The entry splits the two lineages into separate pathophysiology nodes correctly — and names the antigen on neither. TGM2 is absent from `genetic`, the three antibody rows are unbound, and `Gluten Exposure` has no `exposure_term` and no `influences_mechanisms`, so the T-cell antigen is not in the pathograph at all.

`Type_I_Diabetes` has no autoantibody rows at all — GAD65, IA-2 and ZnT8 appear zero times — while eight evidence `explanation` fields say some version of "does not discuss autoantibody production".

The one antigen-centric module, `molecular_mimicry_autoimmunity`, collapses the distinction into a single node `Cross-Reactive Autoreactive Lymphocyte Activation` carrying both `T cell` and `B cell`. Its description says the antigen activates "autoreactive T and/or B cells" — the *and/or* is doing work the data model cannot express.

## What this deliberately does not decide

Whether to change the schema. The numbers say the information is missing from structure; they do not say a new slot is worth the cost. 105 of the unbound autoantibody strings could be bound to HP terms already sitting in `cache/hp/terms.csv` with no schema change at all. The report lists the existing descriptors and the link-object pattern already used by `target_mechanisms` and `influences_mechanisms`, which is the shape an antigen link would take.

It also does not claim every unbound antigen is a defect. `Anti-Endomysial Antibodies` names a tissue staining pattern, not a molecule, and the HLA serotype rows above are the case where "no term beats a bad one" was already applied and written down.

## Files

- `scripts/immune_antigen_audit.py` — offline; ~70 s on CI. `just immune-antigen-audit`, `--format summary|tsv|list`, `--entry <stem>`. Committed so the numbers regenerate rather than being trusted.
- `docs/reports/immune-antigen-representation-gap-analysis-2026-09-03.md` — the report, with a method-and-caveats section.
- `project.justfile`, `mkdocs.yml` — recipe and nav entry.

## Caveats a reviewer should weigh

The cohort is a keyword union, not a MONDO closure: 263 of 573 entries qualify only because their prose says "autoimmune", so false positives are certain. The antigen regexes are deliberately generous, since the whole point is that none of the places an antigen is named is a slot — the anti-Müllerian and anti-DNase hits above are that noise. Raw pattern counts are upper bounds; the structural counts are exact. Section 9 records the rest, including how schema-eligibility is decided and the limits of the HLA row test.

https://claude.ai/code/session_01Uhza23Nsv2k8YEyjv4bymw

### DIFF
--- a/docs/reports/immune-antigen-representation-gap-analysis-2026-09-03.md
+++ b/docs/reports/immune-antigen-representation-gap-analysis-2026-09-03.md
@@ -21,8 +21,8 @@ just immune-antigen-audit --entry Celiac_Disease
 
 **The schema has no antigen concept.** The string `antigen` does not appear once
 in any of the 21 files under `src/dismech/schema/` (7 top-level YAML, 13 more
-under `classifications/`, and a README) — not as a class, not as a slot, not as an enum value.
-Every antigen in the knowledge base is free text.
+under `classifications/`, and a README) — not as a class, not as a slot, not
+as an enum value. Every antigen in the knowledge base is free text.
 
 The measurable consequence, stated over the denominator that can actually carry
 the answer: **only three classes own a cell-type slot** — `Pathophysiology`,
@@ -140,8 +140,9 @@ biochemical:
 This is a real improvement over a bare string, and HPO supports it further than
 the KB uses it — **33 HP terms with `antibody` in the label are already in
 `cache/hp/terms.csv`** — 26 of them of the form `... antibody positivity`,
-headed by `HP:0030057 Autoimmune antibody positivity` — covering ANA, anti-dsDNA-adjacent, ANCA/MPO/PR3, anti-Ro/SS-A,
-anti-cardiolipin, anti-β2GPI, anti-MuSK and more. Three entries use them.
+headed by `HP:0030057 Autoimmune antibody positivity` — covering ANA,
+anti-dsDNA-adjacent, ANCA/MPO/PR3, anti-Ro/SS-A, anti-cardiolipin, anti-β2GPI,
+anti-MuSK and more. Three entries use them.
 
 But note what the HP binding says: it asserts *the patient is seropositive*. It
 does not identify the antigen as a molecular entity, so it cannot be joined to
@@ -272,18 +273,34 @@ a curated identity field (`name` / `preferred_term` / `label`) or in prose:
 |---|---:|---:|---:|
 | CD4 | 952 | 226 (23.7%) | 726 |
 | CD8 | 737 | 183 (24.8%) | 554 |
-| CD27 | 245 | 12 (4.9%) | 233 |
+| CD27 † | 245 | 12 (4.9%) | 233 |
 | CD3 | 141 | 3 (2.1%) | 138 |
 | CD20 | 140 | 4 (2.9%) | 136 |
 | CD19 | 104 | 4 (3.8%) | 100 |
 
+The table stops at 100 mentions; below it the script also reports CD21 (25),
+CD38 (10), CD79a (3), CD138 (1) and CD22 (1).
+
 CD4 and CD8 are the best-represented, but only as a side effect: they ride along
-inside CL labels such as `CD8-positive, alpha-beta T cell`. CD27 is the third
-most-mentioned and sits at 4.9%, which is the same story — it appears in CL
-labels only where a term happens to be named for it. CD19, CD20 and CD3 —
+inside CL labels such as `CD8-positive, alpha-beta T cell`. CD19, CD20 and CD3 —
 the B-lineage and pan-T markers, and the ones that matter for therapy — have no
 CL term to ride on and are therefore almost entirely prose. **79 entries mention
 rituximab**; the CD20 it depletes is a sentence in a `description`, not a target.
+
+† **CD27 is in the table because it outranks the three rows below it, but it is
+not the same signal, and it should not be read as one.** 217 of its 245 mentions
+(89%) come from a single entry, `CD27-related_lymphoproliferative_and_immune_disorder`,
+which is named after the gene; the other 28 are scattered across 12 entries. And
+none of its 12 identity-field hits is a CL label — they are the disease `name`,
+its MONDO `disease_term`, a pathophysiology node name, two `diagnosis` names,
+and HGNC **gene** descriptors under `pathophysiology`, `genetic` and `diagnosis`.
+
+That makes CD27 a *different* failure mode from CD4/CD8, and a sharper one for
+this report's argument. Where CD27 is bound at all, it is bound as a gene. The
+KB can say "the CD27 gene" and it can say "a memory B cell" — it has no way to
+say that the cell displays the molecule. A lineage marker is exactly a molecule
+displayed by a cell, so the one relation that would make it queryable is the one
+missing, and binding the gene does not supply it.
 
 ## 6. Worked exemplars
 


mellybelly:
All three blocking items were right, and I reproduced each one before changing anything. Pushed in b41e9143c.

**Item 1 — the headline denominator.** Confirmed exactly: 2,892 objects have an empty or absent cell-type slot, 75 have one naming a non-lymphoid cell, and 2,694 of 3,122 (86.3%) belong to classes that cannot own the slot at all. The report now leads with **64.0% over the 428 eligible objects** and reports 95% beside it, labelled as the artifact it is. The counter was genuinely mislabelled — it said "no cell_types at all" and meant "no B/T/APC lineage" — so it is now split into `antigen_obj_no_cell_types` and `antigen_obj_cell_types_non_lymphoid`, and the script prints the eligible-subset rate as its own block. Eligibility is decided by path, since the walker sees raw mappings with no class information; section 9 records that, so `pathophysiology[0].downstream[1]` is correctly ineligible.

**Item 2 — CD-table provenance.** You were right that it came from an ad-hoc pass. Rather than annotate the provenance I moved the computation into the script, so the report claim holds literally. The table is now cohort-scoped like everything else and regenerates: CD4 952 (226 in an identity field, 23.7%), CD8 737 (183, 24.8%), CD3 141 (3, 2.1%), CD20 140 (4, 2.9%), CD19 104 (4, 3.8%).

**Item 3 — the HLA framing.** This one went further than you flagged. Both `Celiac_Disease` *and* `Type_I_Diabetes` document the decision — T1D's rows say "HLA-DQ2 is a haplotype-level risk label, not a single HGNC-resolvable gene". So I checked the whole set, and of the 41 unbound HLA rows, **zero name an HGNC gene symbol**; all 41 are serotypes, alleles, haplotypes or region descriptions. Section 8's "mechanical backfill for 57 entries" therefore described work that does not exist, and is gone. The section now says the opposite: the HLA problem is the genuinely new modelling one, because the KB has nowhere to put a serotype or an allele, and the six explained rows are cited as prior art. The script classifies each HLA row as bound, unbound-but-bindable, or unbound-and-not-a-gene, and records whether the entry wrote down its reasoning.

Suggestions taken: added the `just immune-antigen-audit` recipe; fixed `--cohort` being ignored in summary mode (it now aggregates over the selected cohort and says so); `lineage_of` reads `FunctionalEffect.affected_cell_types`; `THERAPEUTIC_SLOTS` moved to module level; corrected the HP term count to 33, the 531 figure's wording (B/T-lineage nodes, against 2,158 carrying any `cell_types` — I confirmed 2,158), the schema file count to 21, the Celiac `biomarker_term` wording, and the runtime to ~70 s on CI / ~3 min 40 s here.

Two notes on things I did *not* simply accept:

- The PR description was already updated before your review posted, so I think you read the original. It is updated again now, since fixing item 1 changed the headline.
- Your HLA row test surfaced one false positive of my own matcher worth recording: `Juvenile_Idiopathic_Arthritis`'s row is named "Non-HLA immune-susceptibility genes…", which matches the `HLA` substring while being about the opposite. It is counted in the 41 and section 9 says so rather than quietly special-casing it.

Separately, while fixing the earlier ruff failure I found the summary output was not reproducible at all: `named_antigens` is a set, so `Counter.most_common` broke ties in `PYTHONHASHSEED` order and two runs of the same tree disagreed. That is now sorted deterministically and verified identical across seeds — which matters given you re-ran the script to check the report.

Every figure in the report is cross-checked against a fresh run, and `uvx ruff check .` passes on latest.

---

mellybelly:
Thanks — took the suggestions in `6bbe96bd8`. Four were right, one was right about the discrepancy but not its cause, and one no longer applies. Flagging that this push drops the approval; four of the six were wrong numbers in a document meant to inform a schema decision, so a re-review seemed the cheaper error.

**Fixed:**

- **Schema file split.** 21 is 7 top-level YAML + 13 under `classifications/` + a README, not "7 plus 14". This also prompted me to re-check the framing claim across all 21 files rather than the 7 YAML I originally grepped — still zero occurrences of `antigen`, so the claim holds on the wider count too.
- **CD27.** You were right that it makes §5 better, not worse. It is now in the table (245 mentions, 12 in identity fields, 4.9%) and in the prose, since it is the same story as CD4/CD8: it appears in a CL label only where a term happens to be named for it.
- **`biomarker_term` prefix counts.** Your 452/249 versus my 451/248 was not a tie-breaking difference — my figures came from a `grep -A3` over `kb/disorders/` only, which both missed blocks whose `term.id` sits more than three lines down and excluded `kb/modules/` entirely. Recounted structurally over the report's stated scope: CHEBI 452, NCIT **269**, HP 71, `hgnc` 4, GO 2, plus 22 blocks with no `term.id` at all. So we were both undercounting NCIT.
- **`stages[].pathophysiology`.** Confirmed it is a `Pathophysiology` list, so the `$`-anchored test was wrong in principle even though it happened to be right in effect. Rather than enumerate the nesting I changed the test to match on the path's last segment, which follows these slots wherever the schema puts them; each of the three slot names has exactly one range and no class overrides it, so the suffix identifies the class on its own. Zero antigen-naming objects currently sit in a stage, so 428 / 154 / 274 / 2,694 are unchanged — verified — and §9 now records both the nesting rule and the 907-object trap it exists to avoid.

**Right about the discrepancy, wrong about the cause.** The 33-vs-32 is not 33-including-the-parent and 32-below-it. Of the 33 cached terms with `antibody` in the label, **7 are not positivity terms at all** — `Impaired specific antibody response`, `Decreased specific antibody response to vaccination`, `Increased anti-meat allergen IgE antibody level` and similar. The real split is 26 `... antibody positivity` terms among the 33. Both statements now say what they actually counted, so they no longer read as a contradiction.

**No longer applies.** The PR description already leads with 64.0% — I updated it at 09:39, a few minutes after your re-review job started, so you would have read the previous version.

Every figure re-verified against a fresh run; `uvx ruff check .` clean on latest.

---

mellybelly:
You were right and the sentence was wrong. Fixed in `4706e4e31`.

I pulled every CD27 occurrence in the cohort and it confirms both of your points:

```
total mentions: 245
CD27-related_lymphoproliferative_and_immune_disorder: 217  (89%)
next: Immunodeficiency_64 (5), Waldenstrom_Macroglobulinemia (5), Immunodeficiency_92 (4)

identity-field hits (12) — none is a CL label:
  <root>.name                                   = CD27-related lymphoproliferative and immune disorder
  disease_term.preferred_term                   = (same)
  pathophysiology[].name                        = Loss of CD27-CD70 costimulation
  pathophysiology[].genes[].preferred_term      = CD27
  pathophysiology[].genes[].term.label          = CD27
  genetic[].name / gene_term.preferred_term / gene_term.term.label
  diagnosis[].name                              = CD27 expression by flow cytometry
  diagnosis[].name                              = TNFRSF7/CD27 molecular genetic testing
  diagnosis[].diagnosis_term.qualifiers[].value.{preferred_term,term.label}
```

So it is neither a CL-label signal nor a cohort-wide one, and the other rows in that table are both. I kept the row — it does genuinely outrank CD3/CD20/CD19, and dropping it would leave a reader wondering — but marked it `†` and footnoted it, rather than explaining it with a mechanism that is not there.

I took your reframing, because it is the stronger claim and it is the one the data supports: where CD27 is bound at all, it is bound as a **gene**. The KB can say "the CD27 gene" and it can say "a memory B cell", and has no way to say the cell *displays* the molecule. A lineage marker is exactly a molecule displayed by a cell, so the one relation that would make it queryable is the one missing — and binding the gene does not supply it. That is a different failure mode from CD4/CD8 riding a CL label, and §5 now says so.

Both suggestions taken as well: a clause recording where the table cuts (100 mentions — CD21 25, CD38 10, CD79a 3, CD138 1, CD22 1 fall below), and the two over-long prose lines rewrapped. I checked the rest of the document for the same problem and those two were the only ragged prose lines, as you said.

No numbers changed; re-verified against a fresh run, and `ruff` is clean.

---

mellybelly:
Heads-up on a red check that isn't this PR's, and which I can't clear myself.

**`claude-review` failed on `d9a4873e5`** ([job](https://github.com/monarch-initiative/dismech/actions/runs/33835497079/job/100907051085)) with "Claude encountered an error after 0s". The job log contains only post-job cleanup — no checkout of the diff, no review work, nothing that touched this PR's contents. It died at startup, so it is a failure of the review workflow itself rather than a finding against the change.

I tried to re-run it and got `403 Resource not accessible by integration`, so I don't have the permission. There's also nothing in this PR's scope that could make it robust — the failure is in the workflow's own initialization, not in code this PR touches.

`test (3.13)` is still running on the same head and is the substantive gate; I'll report if it goes red.

For context on why this head is new: `main` moved 36 commits overnight and added a `model-scale-audit` recipe to `project.justfile` at the same insertion point as this PR's `immune-antigen-audit`, which put the PR into conflict. I rebased onto `origin/main` keeping both recipes — that's the only change, and the diff is unchanged at 4 files / 1,048 insertions. `mkdocs.yml` also moved on main but auto-merged, since its edits are in the Explanation and superpowers sections while this PR's line is in Reports. `uvx ruff@latest check .` passes repo-wide and the script still runs post-rebase.

The rebase also dismissed the previous approval, so this needs a fresh review pass once `claude-review` can start. I'm keeping the PR watched until it's green.

---

mellybelly:
Correcting my earlier comment: the `claude-review` failure is not a workflow bug. The reviewer has run out of usage quota, and it resets at 07:30 UTC. Nobody needs to debug anything — the job just needs to be re-run after that.

The earlier failure logged only "Claude encountered an error after 0s" with an empty job log, which is why I read it as a startup fault. The run on the current head (`ec032ddc1`) got far enough to log the real reason:

```
error='' detail="You've hit your limit · resets 7:30am (UTC)" num_turns=1 cost=$0
```

I still can't re-run it myself — `403 Resource not accessible by integration` on both the job re-run and a `claude-code-review.yml` dispatch — so someone with workflow permission clicking "Re-run failed jobs" after 07:30 UTC is all this needs. I'll retry from my side then too, in case the permission behaves differently once the run is no longer rate-limited.

One thing worth watching beyond this PR: a quota limit is account-wide, so any other PR whose review ran in the same window will have failed the same way, and the reset time is shared. If this recurs it's worth looking at rather than re-running each PR by hand.

Otherwise unchanged. The rebase onto `2b2132c5f` cleared the `project.justfile` conflict (`main` had added `diet-audit` at the same insertion point; all three recipes now sit side by side), the diff is still 4 files / 1,048 insertions, and `uvx ruff@latest check .` passes.

---

mellybelly:
@cmungall — this PR needs one click from someone with workflow permission: **Re-run failed jobs** on [run 34013807668](https://github.com/monarch-initiative/dismech/actions/runs/34013807668). No debugging needed.

`claude-review` failed at 05:24 UTC because the reviewer account was out of quota, not because of anything in this PR:

```
error='' detail="You've hit your limit · resets 7:30am (UTC)" num_turns=1 cost=$0
```

The quota reset at 07:30 UTC, so a re-run should now go through. I've retried `rerun-failed-jobs` and a `claude-code-review.yml` dispatch five times since then and get `403 Resource not accessible by integration` on both — I don't have the permission.

Everything else on the PR is green and ready: `test (3.13)` passes, there are no merge conflicts against current `main`, and the diff is unchanged at 4 files / 1,048 insertions. The only thing standing between this and a review is the re-run.

One thing worth a look beyond this PR: a quota exhaustion is account-wide, so any other PR whose review ran in that window failed the same way and will also be sitting on a red `claude-review` that nobody can clear.

---

mellybelly:
Thanks for the re-run — it failed again on quota, with a new reset time of **17:40 UTC**. But the re-run surfaced two things worth more than another retry.

**1. The workflow's own error handler names the likely root fix: `ANTHROPIC_API_KEY` is probably unset.** From `.github/workflows/claude-code-review.yml` (line 313), the OAuth token is only used when no API key is set:

```yaml
anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
claude_code_oauth_token: ${{ secrets.ANTHROPIC_API_KEY == '' && secrets.CLAUDE_CODE_OAUTH_TOKEN || '' }}
```

So hitting a subscription limit at all implies the run fell back to `CLAUDE_CODE_OAUTH_TOKEN`. Setting the `ANTHROPIC_API_KEY` repo secret would stop this recurring — pay-per-use has no weekly cap. Re-running before then will likely just exhaust again; the account went from a 07:30 reset to a 17:40 one in a single day.

**2. There's a small bug in the limit detector, and it's why this looked like a workflow fault for two days.** Line 444:

```python
```

The actual detail string is `"You've hit your limit · resets 5:40pm (UTC)"` and `reason` is empty, so none of the three tests match. The run falls through to the generic branch and reports "Common causes: an invalid/retired `--model`, a bad `CLAUDE_CODE_OAUTH_TOKEN`, or an aborted run" — none of which is true — instead of the accurate "usage limit reached, here's the fix" message sitting right above it. Matching `"hit your limit"` as well would fix it.

I haven't touched either file: this PR is a docs and audit change, and widening it to patch a workflow would be a different review. Happy to open a separate PR for the detector string if you want it.

Nothing else is blocking. `test (3.13)` passes, no merge conflicts against current `main`, diff unchanged at 4 files / 1,048 insertions.

---

Issue #8549, opened 2026-08-15
Curate vertebral, cardiac, renal, and limb defects syndrome 1 (MONDO:0060554)


sierra-moxon:
Curate a dismech entry for **vertebral, cardiac, renal, and limb defects syndrome 1** ([MONDO:0060554](https://monarchinitiative.org/MONDO:0060554); VCRL1, OMIM:617660) — autosomal recessive multiple-malformation syndrome caused by biallelic mutation in **HAAO** ([hgnc:4796](https://www.genenames.org/tools/search/#!/?query=HAAO)), 3-hydroxyanthranilate 3,4-dioxygenase.

Mechanistically notable as a **congenital NAD deficiency disorder**: HAAO loss blocks the de novo kynurenine → NAD+ biosynthetic pathway, so embryonic NAD+ falls below the threshold needed for somitogenesis and organogenesis, producing vertebral segmentation defects, congenital heart disease, renal anomalies and limb defects. Niacin/nicotinamide supplementation rescues the phenotype in the mouse model — a rare gene-to-preventive-treatment mechanism worth modeling explicitly (`evidence_source: MODEL_ORGANISM` for the rescue data). The vertebral arm may cross-reference `axial_segmentation_serial_homology`.

Priority dashboard action: `LUMP_INTO_PARENT` (score 35.5, subtype series, family size 3, parent *congenital vertebral-cardiac-renal anomalies syndrome*). VCRL2 (KYNU) and VCRL3 (NADSYN1) are the sibling NAD-pathway entities — consider a root entry with `has_subtypes` spanning the pathway.

Source: [priority dashboard](https://dismech.monarchinitiative.org/dashboard/priority.html).

Tracker: part of #1079.

sierra-moxon:
Closing as completed: priority dashboard action is `LUMP_INTO_PARENT` (score 35.5, subtype series, family size 3, parent *congenital vertebral-cardiac-renal anomalies syndrome*). VCRL1 (HAAO) belongs as a `has_subtypes` entry under a root spanning the congenital NAD-deficiency pathway alongside VCRL2 (KYNU) and VCRL3 (NADSYN1).

PR #10680, opened 2026-09-03
Add Congenital Vertebral-Cardiac-Renal Anomalies Syndrome disorder entry

tannerzhang:
## Summary

This curates **congenital NAD deficiency disorder** (CNDD), an autosomal recessive multiple-malformation syndrome that looks like VACTERL association but has a testable metabolic cause: biallelic loss of one of three enzymes that make NAD from tryptophan. Embryonic NAD falls below what organogenesis needs, and the result is vertebral, cardiac, renal and limb malformation.

**The entry is at the parent term, not at VCRL1.** The request was to curate VCRL1 (MONDO:0060554, HAAO). Issue #8549 was closed with the decision `LUMP_INTO_PARENT`, and GeneReviews independently treats HAAO, KYNU and NADSYN1 as one disorder. So this is one entry on MONDO:0020831 with VCRL1/2/3 as `has_subtypes`. All three MONDO terms are covered; the VCRL1 stub is deleted. If reviewers disagree with that shape, that is the thing to argue about first — everything else follows from it.

## What is in the entry

- 30 phenotypes bound to HPO, frequencies mapped from the GeneReviews prose bands
- A causal chain, not a list of mechanisms: enzymatic block → embryonic NAD deficiency → per-organ morphogenesis nodes → phenotypes, with maternal NAD precursor supply entering as a **second, independent input** to the same node. That second arm is the point of the shape: mice with no genetic lesion develop the same malformations on a B3-restricted maternal diet (PMID:32015132)
- 117 evidence snippets, all verified against the reference cache
- Four animal models (Haao-null, Kynu-null, Nadsyn1-null mice, ATDA zebrafish) with per-link fidelity and limitations
- Gestational niacin as a protective exposure, and antenatal supplementation as a treatment

## Where the human evidence stops

Worth stating plainly, because the mechanism is unusually clean and it would be easy to overread it. Every prevention result is murine. No trial or cohort has tested supplementation in an at-risk human pregnancy, and the paper that did the mouse work says so itself: *"Additional clinical trials are needed before accurate dosage recommendations required to prevent CNDD-dependent malformation can be made."* The only human data are population-level (PMID:34748060, an association between low maternal niacin intake and congenital anomalies in general).

The entry keeps that line everywhere: the treatment is described as mechanistically motivated rather than established, the mouse-only vitamer-specificity result is labelled as such, and two `discussions` record the gap and the model mismatch.

A third `discussion` records something the pathograph cannot currently do: the effector layer between "NAD falls" and "this organ fails" is unknown, and it has to explain a selectivity problem, since NAD is needed in every cell but the malformations land on a specific set of organs. The zebrafish paper states this directly (PMID:39829932).

## Commit history

Five commits. The entry was written once and then corrected four times, and the corrections are the interesting part.

**Commit 2** fixed findings from an adversarial self-review run in fresh context before the PR opened. Three were claim-evidence mismatches:

- PMID:34681008 was cited as causative evidence for NADSYN1. Its variants are **heterozygous** and classified as VUS, which does not evidence a biallelic causative relationship and contradicted the entry's own framing.
- The 3-hydroxyanthranilic acid biomarker asserted `presence: PRESENT` on a quote saying only that metabolites were measured. Nothing cited reports that elevation in a patient.
- The postnatal *amidated*-precursor treatment was evidenced by nicotinic acid (not amidated) in a HAAO patient, the subtype where that route works anyway.

That review also caught that there was **no limb phenotype at all** — the "L" of VCRL, 8 of 12 in the largest genotyped series — while a pathophysiology node was already asserting it downstream.

**Commit 3** discloses that a batch of commit 2's fixes never reached disk: an edit script asserted on strings a `reference_title` backfill had reformatted, exited before its single write, and the follow-up validation proved the file well-formed rather than the edits present. Commit 3 reapplies them, verified on disk, and addresses the reviewer's two blocking items.

**Commits 4 and 5** close the reviewer's deferred gaps (a CHEBI binding for the NADSYN1 substrate, and hydroxykynureninuria as a differential) and drop one more unsupported zygosity claim the reviewer spotted.

## Deliberately not done

- **No `conforms_to` on `axial_segmentation_serial_homology`.** That module is built on the Notch segmentation clock, and nothing shows NAD deficiency acts through that oscillator. The phenotypes converge; the mechanisms do not. Recorded in `notes` instead.
- **No `datasets`.** The only GEO candidate was a bladder-cancer study matched on the NADSYN1 gene symbol alone.
- **No disease-level prevalence.** The GeneReviews cohort count is in the full chapter but not the cached abstract, and no other cached source gives one. The prevalence record is VCRL3-scoped and says so.
- **One phenotype left unbound.** HPO codes pregnancy loss outside the branch the `PhenotypeTerm` enum expands from, so there is no valid binding.
- **No monoallelic-NADSYN1 `discussions` entry**, though the deep research proposed one. It traces back to the same heterozygous-VUS cohort removed as evidence above, so filing it as a hypothesis would smuggle the claim back in without new support.

## Provenance

falcon was requested and returned 403 on both attempts. Rather than swapping providers by hand, the run was reissued with `--fallback`, which fell back to **openscientist** and recorded the failed attempt in the report frontmatter. A **claude_code** run went in parallel. Both reports are committed and both validated clean on references.

The term validation on those reports earned its keep. The claude_code report offered `HP:0004392` for hypoplastic left heart — that code is *Prune belly*. It also mislabelled `GO:0034354` as the salvage pathway (it is the de novo route) and `UBERON:0002298` as vertebral column (it is brainstem). None were copied across; every binding was resolved independently with OAK.

## Validation

`just validate-disorders` passes: schema, terms, and 117/117 snippets. Also passing: `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-term-cache-integrity`, `check-cache-order`, `check-stubs`, `validate-history`. Compliance 94.9%.

Related: #8549 (the decision this entry's shape follows)

https://claude.ai/code/session_015rdBBTsSyLPp2TjvjSCQ5V


tannerzhang:
Pushed `c182576dd`. Thanks — the review caught something bigger than the two findings.

## First, a correction to the previous commit

**A batch of fixes I claimed in commit 2 never reached disk.** The edits ran through a script that asserts on each expected match before its single write. Two assertions failed on strings a `reference_title` backfill had reformatted, so the script exited before writing, and the rest of the batch went with it. I re-ran validation afterwards, it passed, and I read that as confirmation — but it only proved the file was still well-formed, not that the edits were there. The commit message overstated what changed.

That is why findings B and C looked unaddressed: the reviewer was reading the state those lost edits should have changed. They are now applied and verified present on disk, not just validated:

- the monoallelic-VUS PMID:34681008 item removed from `genetic[NADSYN1]`
- the 3-HAA biomarker reframed as predicted (finding B): `presence` unset, `notes` rewritten to say no cited source reports the elevation, evidence marked `directness: INDIRECT`
- the `GO:0019805` `DECREASED` descriptor removed (finding C) — not just for readability but because it is *false for VCRL3*, where the NADSYN1 block sits downstream of quinolinate, so a node-scoped tag was wrong for a third of the disorder
- a wrong step count, two murine results stated as general fact, and a mis-cited spectrum claim

## Finding A — correct, and my doing

You are right, and this one I introduced. PMID:32015132 is a mouse publication; the sentence naming humans is its framing of prior work, not its own human data. An earlier self-review flagged that a human phenotype rested only on `MODEL_ORGANISM` evidence, and I fixed that rule by breaking this one.

Took your remedy rather than the re-grade: back to `MODEL_ORGANISM` with the reason in its `explanation`, plus a genuinely human second item — PMID:41170824, altered NAD-pathway metabolites in women with recurrent miscarriage, `directness: INDIRECT` since that cohort is recurrent miscarriage generally, not CNDD.

One correction to the finding's reasoning, which does not change the outcome: the other PMID:32015132 items quote *different sentences*, so `check-snippet-grading` was never going to fire here. The finding stands on the publication-type rule by itself.

## E — taken

`directness: INDIRECT` added to the GeneReviews management-list snippet on Ptosis, matching its four siblings.

## D — no term to move to

`HP:0004211` does not resolve in HPO; `runoak info` returns no label. So `HP:0004383` (Hypoplastic left ventricle) stays, with `preferred_term: hypoplastic left heart syndrome` carrying the narrower concept — the documented pattern for a `preferred_term` more specific than its binding.

## Also fixed while in there

The differential-diagnosis text said NADSYN1 variants "accounted for a minority of cases" in the vertebral malformation cohort. Those variants were heterozygous and of uncertain significance, which is the same overstatement that made the evidence item wrong. Now says so.

## Not taken, with reasons

Hydroxykynureninuria as an allelic KYNU condition, and the monoallelic-NADSYN1 hypothesis as a `discussions` entry. Both need sources beyond the current cache. The second also cuts against removing the VUS evidence item — adding it now would reintroduce as a hypothesis the claim just removed as evidence, with no new support behind it.

## Validation

`just validate-disorders` passes, 115/115 snippets verified. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` all pass. Compliance 94.9%.

A history record documenting the lost-edit failure and this round is in `history/disorders/Congenital_Vertebral-Cardiac-Renal_Anomalies_Syndrome/`. The earlier record is left as written, since history is append-only and it accurately records what that session believed.

---

tannerzhang:
Thanks for the approval. Pushed `74b3dc8c0` closing both non-blocking items — knowing it resets the approval, since they're small and the re-review is quick.

**NaAD binding.** The accumulation node claimed a gene-specific pattern across all three subtypes but bound CHEBI entities for only two, so the VCRL3 arm was prose the machine couldn't see. `CHEBI:18304` (deamido-NAD(+)) is now bound and named in the description instead of "the terminal precursor".

**Hydroxykynureninuria.** Added as a differential, and it turned out to matter more than a taxonomic gap. PMID:34200361 (already cached) describes brothers with a KYNU missense variant "presenting with hydroxykynureninuria and xanthurenic aciduria but without congenital malformations", with residual kynureninase activity keeping plasma NAD high enough as the proposed basis for the split. That is the same gene and the same block as VCRL2, separated only by how much enzyme survives — so **a raised xanthurenic acid on its own does not establish CNDD**. The entry presents that metabolite as a diagnostic pointer, so leaving the allelic condition out was a real hole in the diagnostic story rather than a missing cross-reference.

**Still not taking the monoallelic-NADSYN1 hypothesis as a `discussions` entry.** It would reintroduce as a hypothesis the same claim removed as evidence when the heterozygous-VUS item was deleted, with no new support behind it. If a source establishes monoallelic contribution independently, it belongs — but the DR report's suggestion traces back to the same VUS cohort.

**On your grading-coherence note** (somitogenesis sentence at `OTHER` vs finding A resolved on publication type): I've left it. Both are consistent with the rule as CLAUDE.md states it — `evidence_source` classifies *the type of evidence presented in the cited publication*, and a background sentence from an introduction presents no evidence of any kind, which is what `OTHER` is for. Finding A was different: that sentence *is* the mouse paper's own framing of its subject. The `explanation` on the somitogenesis item says which case it is, so the file's rationale is self-documenting if anyone revisits it.

Validation: 117/117 snippets, `just validate-disorders` passes, all offline gates pass.

---

tannerzhang:
Took the nit anyway — `2c1f789e6`.

You were right that the source doesn't carry it. PMID:34200361 says only "two brothers with **a missense variant** in KYNU" and reports no zygosity, so "Biallelic KYNU variants" was an assertion the citation behind it didn't make.

I'm fixing it despite your "not worth another round" call for the reason you identified: it is the same class of defect as the PMID:34681008 finding earlier in this PR, where a heterozygous-VUS cohort was cited as evidence of a biallelic causative relationship. A knowledge base whose value is claim precision shouldn't keep a known zygosity overstatement because correcting it costs a CI cycle — and the PR can't auto-merge inside three days regardless, so the cycle is free.

The description now makes the claim the source does support (residual enzyme activity, not allele configuration) and records that the cited review doesn't report zygosity. Hydroxykynureninuria is autosomal recessive, so biallelic involvement is very likely right — it just isn't what this citation establishes, and the primary Christensen report isn't in the cache to check.

`just validate-disorders` passes, 117/117 snippets.

---

tannerzhang:
`claude-review` is red on the current head, and it is not this PR's failure.

The job log gives the cause directly:

```
```

An account usage limit, one turn in, zero cost — the agent never got far enough to read the diff. Nothing in this PR causes it, and re-running before **11:20 UTC** would fail the same way, so I have not spent a re-run on it yet. I will re-trigger the review once the limit resets. `test (3.13)`, the actual test suite, is unaffected and running normally.

Two pushes since the last approval, both forced by `main` moving:

- `7d5e45ce3` merges `origin/main` after the branch hit a genuine conflict in two derived term caches (`cache/go/terms.csv`, `cache/enums/molecularfunctionterm_305077eac108.csv`). Resolved by taking main's side and re-deriving this entry's rows with `just validate` + `just normalize-cache`, rather than hand-merging generated files.
- `1ba9f887e` fixes my own mistake in that resolution: I regenerated the dropped rows but never re-staged them, so the merge commit shipped without the four GO terms the YAML binds (`GO:0000334`, `GO:0030429`, `GO:0008795`, `GO:0034354`). Caught by a working-tree check, verified this time by diffing the pushed commit per term rather than trusting that validation passed.

Cache diff against `main` is additive only — 36 insertions, 0 deletions. The disorder YAML is unchanged by either commit. Local gates all pass on the current head: `just validate-disorders`, 117/117 snippets, plus `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-grading`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-environmental-evidence`, `check-term-cache-integrity`, `check-cache-order`, `check-stubs`.

---

tannerzhang:
Correcting my last comment: I said I would re-trigger the review once the usage limit reset. The limit has reset, but **I do not have permission to re-run it.**

Both routes return 403 `Resource not accessible by integration`:

- `POST /actions/runs/33957303959/rerun-failed-jobs`
- `POST /actions/workflows/claude-code-review.yml/dispatches`

And the `issue_comment` trigger on that workflow is gated to `OWNER`/`MEMBER`/`COLLABORATOR`, so a comment from me does not fire it either — all six review runs on this branch so far were triggered by `pull_request` events. So the deferred re-run is not something I can spend.

**What will fix it without anyone doing anything:** the workflow runs on `synchronize`, and `main` moves several times a day while this branch's derived term caches (`cache/go/terms.csv`, `cache/enums/molecularfunctionterm_305077eac108.csv`) conflict on most of those moves. The next merge-resolution push will re-trigger the review naturally. I'm keeping the PR watched and will handle that conflict when it appears.

**If someone wants it sooner**, a maintainer can dispatch it directly:

```
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10680
```

State of the PR meanwhile, on head `1ba9f887e`:

| | |
|---|---|
| `test (3.13)` | success |
| `claude-review` | failure — quota, never read the diff |
| merge vs `main` | clean (`git merge-tree --write-tree`, exit 0) |
| local gates | all pass, 117/117 snippets |

The only red thing is a review that could not start. Nothing in the diff is implicated.

---

tannerzhang:
Correcting myself again, and this time in the useful direction: **I can re-trigger the review.**

My last comment said the `issue_comment` path would not fire for me. Wrong. The gate is not the author association — my comments post as `MEMBER`, which passes. It is an exact-body match:

```yaml
github.event_name == 'issue_comment' &&
github.event.issue.pull_request != null &&
github.event.comment.body == '/review' &&
contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
```

My earlier comments were prose, so they never matched `'/review'`. I read the fragmentary condition, saw the author list, and concluded the wrong thing instead of reading the whole block. The workflow's own comments describe this path as "the one manual escape hatch", which is exactly the situation here.

Posting `/review` next. The 403s on `rerun-failed-jobs` and `workflow_dispatch` were real — that part stands — but they were not the only route.

---

tannerzhang:
/review

---

tannerzhang:
The `/review` comment didn't work either, and the reason is worth recording since it makes that escape hatch unusable for any agent posting through this tooling.

The condition is `github.event.comment.body == '/review'` — an exact match. An attribution footer is appended to my comments **server-side**, so what actually landed was:

```
/review
```

Both my comments did fire `issue_comment` runs (`34013815009`, `34013817639`); both concluded `skipped` on that condition. So the escape hatch is open to humans and closed to me, which is a reasonable design — I just misread it twice before checking the posted body.

Full tally of what I cannot do: `rerun-failed-jobs` 403, `workflow_dispatch` 403, `/review` defeated by the appended footer.

**Pushed `5ef2eb1de`**, which re-triggers the review as a `synchronize` event. To be straight about it: that push helps me here, but it is not a pretext commit. This PR's merge round was the one round without a history record, and the thing it documents is worth having — the cache conflict will recur for any branch that binds new GO terms, since every curation PR appends rows to the same two files, and the record spells out the resolution (take main's side, re-derive with `just validate` + `just normalize-cache`, confirm the diff against main is additive only).

It also records the staging failure, which is the second instance of one pattern in this PR: a check ran green, I read that as confirmation the change was committed, and it was not. First time an edit script exited before its write; this time regenerated cache rows were never re-staged. A green check proves the state it ran against was sound, not that that state is what got committed.

No KB content changed in that commit.

---

tannerzhang:
The push did re-trigger the review, and it hit the same quota wall — new reset time, `7:30am (UTC)`:

```
detail="You've hit your limit · resets 7:30am (UTC)" num_turns=1 cost=$0
```

Not repeating the earlier analysis. One new thing worth flagging to whoever maintains the workflow, though, because it is costing you the good error message:

**`claude-code-review.yml`'s usage-limit detection does not match this wording.** The check is:

```python
```

The actual detail string is `"You've hit your limit · resets 7:30am (UTC)"` — no `"usage limit"`, no `"weekly limit"`, and `reason` is empty. So it falls through to the generic branch and reports *"Common causes: an invalid/retired --model, a bad CLAUDE_CODE_OAUTH_TOKEN, or an aborted run"* — three wrong guesses — instead of the quota branch's genuinely useful message, which is:

> This workflow prefers ANTHROPIC_API_KEY (pay-per-use, no weekly cap) and only falls back to CLAUDE_CODE_OAUTH_TOKEN when no API key is set — so a weekly-limit error here most likely means ANTHROPIC_API_KEY is unset and the run used the subscription token.

Adding `"hit your limit"` to that substring list would classify it correctly. That points at the real fix for this PR and any other blocked the same way: **set the `ANTHROPIC_API_KEY` repo secret**, so review runs stop drawing on a capped subscription token.

I am not fixing that here — it is a workflow change and this is a curation PR. Filing it as an observation for a maintainer.

Not pushing again to retry. The entry itself has been unchanged since `2c1f789e6`, which the reviewer already approved; everything after that is merge upkeep and this history record.

---

tannerzhang:
@cmungall — escalating this one, since it needs a permission I don't have.

This PR is green and was approved on its content (`2c1f789e6`). It is stuck only because `claude-review` has failed twice on an **account usage limit** — the agent never read the diff (`num_turns=1`, `cost=$0`). I can't re-run it: `rerun-failed-jobs` and `workflow_dispatch` both 403 for me, and the `/review` comment hatch needs an exact body match that a server-appended attribution footer breaks.

Two things would help, in order of usefulness:

1. **Set the `ANTHROPIC_API_KEY` repo secret.** The workflow's own error text says it prefers that key and falls back to `CLAUDE_CODE_OAUTH_TOKEN` only when it's unset — so review runs are currently drawing on a capped subscription token. This would unblock every PR hitting the same wall, not just this one.
2. Failing that, a one-off dispatch: `gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10680`

Separately, a small bug worth a fix whenever someone's in there: the workflow's quota detector checks for `"usage limit"` / `"weekly limit"`, but the actual string is `"You've hit your limit"`, so these failures fall through to a generic message blaming the model or the token instead of the quota.

Not assigning this to anyone — that would take it out of the auto-merge sweep, and it should merge normally once a review can run.

---

PR #8772, opened 2026-08-17
feat: Add Secondary Erythromelalgia (MONDO:0035149)

sierra-moxon:
Curates **secondary (acquired) erythromelalgia** (`MONDO:0035149`, Orphanet:529864) as a standalone entry — the sibling of `Primary_Erythermalgia` (SCN9A/Nav1.7) beneath the `erythromelalgia` umbrella.

Closes #8545

## What this models

The **mechanistically distinct secondary route**, deliberately *not* the nociceptor-hyperexcitability mechanism of the primary form:

```
Clonal thrombocythemia in MPN
  -> Thromboxane/COX-1-dependent platelet activation and aggregation
     -> Arteriolar inflammation + fibromuscular intimal proliferation
     -> Acral arteriolar microvascular occlusion (VWF-rich, fibrin-poor)
        -> Episodic acral burning pain, erythema, warmth
        -> Progression to acrocyanosis and digital ischemic necrosis
```

Plus a separate, deliberately low-resolution (`PROVISIONAL`) node for the non-MPN acquired routes (neuropathic, autoimmune, infectious, neoplastic, drug/toxin).

**Aspirin responsiveness** — the key differentiator called out in the issue — is curated across `treatments`, `diagnosis`, and `differential_diagnoses`, with the pharmacological dissection that grounds it: COX inhibitors (aspirin, indomethacin) abolish the syndrome while sodium salicylate, dipyridamole, sulfinpyrazone, ticlopidine and dazoxiben do not. Aspirin and indomethacin carry `INHIBITS` `target_mechanisms` edges onto the platelet-activation node; cytoreduction targets the upstream thrombocythemia node.

The underlying MPN pathophysiology is **cross-referenced, not re-derived** — it stays in `Polycythemia_Vera` / `Essential_Thrombocythemia`.

## Two decisions recorded rather than left silent

**No `conforms_to` on the `thrombogenesis` module.** This looks like an obvious conformer, and isn't. The acral thrombi are von Willebrand factor-rich with only weak fibrin staining, form despite therapeutic anticoagulation, and show no rise in prothrombin fragment 1+2 (PMID:8883266) — thrombin generation is not required. The module explicitly scopes out "VWF-platelet microangiopathy … unless a distinct conventional thrombin/fibrin thrombosis branch is explicitly evidenced". The non-conformance and its reason are in the entry `notes`.

[...]

PR #10681, opened 2026-09-03
Augment Secondary Erythromelalgia with immune/dysautonomic mechanism and classifications

tannerzhang:
## Summary

Deep-research augmentation of the Secondary Erythromelalgia disorder entry with new mechanistic insights, clinical burden data, disease classifications, and evidence curation. This is a targeted enhancement of an existing merged entry (from PR #8772) using OpenScientist autonomous research rather than the default Falcon provider.

## Key Changes

- **New mechanistic pathway**: Added immune/complement-mediated microvascular injury and dysautonomia as a second, complementary mechanism alongside the well-characterized platelet-mediated MPN route. Documented as an emerging hypothesis with C5b-9 membrane-attack-complex deposition, type-I interferon upregulation, and diminished autonomic innervation of eccrine coils and arteries.

- **Expanded acquired associations**: Extended the documented myeloid associations beyond the classical polycythemia vera and essential thrombocythemia pair to include thrombotic thrombocytopenic purpura and chronic myelogenous leukemia (each on single-case evidence, recorded as associations rather than modeled mechanisms).

- **Disease classifications**: Added Harrison's Manual chapter assignments:
  - `CARDIOVASCULAR` (primary): end-arterial microvasculature occlusion across peripheral, cerebral, and coronary beds
  - `ONCOLOGY_HEMATOLOGY` (secondary): underlying myeloproliferative neoplasm diagnosis and cytoreduction as mechanism-directed treatment

- **Synonyms**: Added "Secondary erythermalgia" and "Acquired erythromelalgia" to the preferred-term entry.

- **Progression and burden**: Curated temporal development (onset, progression rate, disease course) and quality-of-life impact from literature.

- **Evidence curation**: Resolved 26/26 references from the OpenScientist report; added 18 new reference cache entries (PMIDs spanning 1989–2026) covering MPN diagnosis/management, erythromelalgia epidemiology, and pathophysiology.

- **Mechanistic clarity**: Clarified the Drenth-Michiels three-way classification (aspirin-responsive thrombocythemic vs. aspirin-resistant secondary vs. primary inherited) and removed retired PARTIAL-support prose in favor of explicit hypothesis tagging.

## Implementation Details

- **Research provider**: Deliberately ran OpenScientist (autonomous mode) rather than Falcon to avoid overwriting the existing Falcon report (2026-08-17) at the same path. OpenScientist writes to a distinct artifact path, so both reports now coexist.
- **Validation**: Ran both `validate-research-reference` and `validate-research-terms` retrofits; all 26 references resolved successfully.
- **Ontology**: Updated term caches for Cell Ontology (CL:0011030) and Uberon (UBERON:0000423) to support new mechanistic annotations.
- **Evidence policy**: Maintained strict PMID citation; single-case associations (TTP, CML) recorded as such rather than speculated into mechanism chains.

## Files Modified

- `kb/disorders/Secondary_Erythromelalgia.yaml` — pathophysiology, mechanistic hypotheses, progression, clinical burden, classifications, synonyms, phenotypes, treatments, diagnosis, discussions, evidence
- `research/Secondary_Erythromelalgia-deep-research-openscientist.md` — full 958-line research report (new)
- `research/Secondary_Erythromelalgia-deep-research-openscientist_artifacts/` — HTML and PDF final reports (new)
- `references_cache/` — 18 new PMID cache entries
- `history/disorders/Secondary_Erythromelalgia/` — curation session record (new)
- `cache/` — ontology term cache updates (CL, UBERON)

https://claude.ai/code/session_01QttvDDjqeXyxVZJi2xPLP5


tannerzhang:
The PR description covers the first commit. There is a second one (`2ade296`) it does not mention, and it is the more interesting half: I ran an adversarial review of my own diff before this PR existed, and it found real problems in the first commit. Sixteen findings, all content objections rather than CI failures, all acted on.

The ones a reviewer should check:

**A causal edge claimed more than its citation.** The edge from the non-platelet acquired routes into the new immune node was graded `causal_link_type: DIRECT`, but its snippet only establishes that 3 of 9 patients in the source series carried an autoimmune diagnosis. The edge description additionally asserted three things PMID:39846717 does not say: that the lesion belongs to the autoimmune arm, that it occurs instead of a platelet-rich thrombus, and that the acquired route causes it. Regraded `directness: INDIRECT`, description and explanation rewritten to claim only what the quote supports.

**The hypothesis label overstated the source.** It read "...in Non-MPN Secondary Disease". That series was assembled by clinical concern for erythromelalgia and is not stratified primary/secondary, let alone MPN/non-MPN, and its own conclusion is about erythromelalgia in general. The label is now aetiology-neutral, and the group description says explicitly that restricting this route to non-MPN disease is part of the hypothesis rather than something carried over from the paper.

**A synonym contradicted this entry's own open question.** The first commit added "Platelet-mediated erythromelalgia". That names the MPN arm alone, has no MONDO backing, and cuts against the entry's existing `secondary_em_nomenclature_scope` discussion. Removed. "Secondary erythermalgia" stays because MONDO carries it as an exact synonym, and the nomenclature discussion now says so, since under Drenth-Michiels that term historically meant the narrower aspirin-resistant category.

**My own cleanup introduced one regression.** Removing the retired `PARTIAL` prose flipped the direction of a claim in one place: the acrocyanosis band explanation went from "this does not support the band" to "the band is read off that figure". The hedge is restored, with `directness: INDIRECT`, and it now records that 14/31 is a curator computation joining two sentences of an abstract that says "in 14 cases" with no denominator attached.

**An over-reading of a case definition.** The Olmsted County paper says a myeloproliferative neoplasm diagnosed "within 2 years of the diagnosis of erythromelalgia", which is symmetric. I had written "up to two years after". Corrected; the directional claim is now carried by PMID:2643412 (median 2.5-year lead) and PMID:11208311, which do state it.

Also in that commit: `CL:0011030` is labelled exactly "dermal microvascular endothelial cell" and replaces the generic `CL:0000115`; two `evidence_source` values regraded `OTHER` (a recommendation sentence and a narrative-review synthesis); the 5.3% polycythemia vera frequency regraded `NO_EVIDENCE`, since it is a different entity's denominator and does not bear on this phenotype's `OBLIGATE` band either way; and PMID:22844295 is no longer described as the syndrome "recurring" on aspirin withdrawal, because that patient was taking aspirin for stroke prophylaxis and the erythromelalgia emerged de novo.

Two suggestions from the review I did not take, both recorded in the entry: rebinding the pre-existing arteriolar endothelial cell term, which is out of scope for this change, and adding `DERMATOLOGY` as a third Harrison's Part, which is now explicitly considered and declined in the classification notes rather than silently omitted.

**Two corrections to the PR description above.** It says 18 new reference cache entries; there are 22. And the classifications are Harrison's *Parts*, which is what the `harrisons_chapter` slot actually holds despite its name, not chapters from Harrison's Manual.

**Things worth a reviewer's attention that are unchanged from the first commit.** The new mechanism rests on a single retrospective nine-patient dermatopathology series, which is why it is `PROVISIONAL` inside an `EMERGING` hypothesis group rather than an asserted node, and why an independent histology report that names neither complement nor denervation is cited beside it as a `NO_EVIDENCE` counterweight. `datasets:` and `clinical_trials:` stay empty; the existing notes give the reasons and I did not disturb them. The umbrella `erythromelalgia.yaml` entry still carries one stale `PARTIAL` line of the same class I cleaned up here, which I left out of scope rather than widen this diff.

Validation on the final bytes: `just validate-disorders` passes with 110/110 snippets verified, `just qc`'s offline gates are green, and the causal graph builds with 25 nodes, 16 edges and no orphan targets. `just preflight-dr` skips, because MONDO records no causal gene for MONDO:0035149; the report's top genes are the MPN drivers JAK2/MPL/CALR, so there is no entity-confusion signal.

---

tannerzhang:
Fixed in `0bf9876`. The blocking finding was correct, and the detail that made it land was that PMID:35835622 was pulled into `references_cache/` by this PR and then never used. I checked the sentence against the cache before citing it.

**The diagnostics gap.** Three changes, all from that one abstract plus PMID:39846717:

- JAK2 mutation testing added to the MPN-evaluation entry. The `results` prose now says why the confirmatory test earns its place here rather than being boilerplate: the flare precedes the haematological diagnosis by a median of 2.5 years, so a normal count at presentation does not close the question.
- New diagnosis entry for the thyroid, antinuclear antibody and small-fiber neuropathy workup. Those are the three non-MPN routes the entry models and the ones `immune_dysautonomic_route` is about, and they are the aspirin-resistant forms, so identifying the underlying condition is what management turns on.
- The skin punch biopsy entry covered only platelet-arm histology. It now records that which stains are ordered decides which arm the biopsy can speak to. C5b-9, type-I interferon and CD56 are invisible to routine histology, so a routine biopsy reported as unremarkable does not exclude the immune route — it was never tested for. That seemed the more useful thing for a curator to know than simply listing the stains.

The two misspellings in that snippet (`thryoid`, `hormon`) are in the source and are kept verbatim, as you noted they must be.

**Suggestion 1, the co-bundled node — taken.** You left it optional on house-style precedent, but CLAUDE.md's single-value discipline is explicit that a node which would take two `biological_scale` values is bundling two claims, so I split it: `Complement-Mediated Microvascular Endothelial Injury` at `CELLULAR` and `Autonomic Denervation of Acral Arteries and Eccrine Glands` at `TISSUE`. They are parallel observations from one biopsy rather than a causal chain, so no edge is drawn between them; each takes its own edge to the flare and its own upstream edge from the acquired-routes node, all still inside the hypothesis group. The split also let the denervation node carry the shunting connection properly, since arteriovenous shunts are autonomically controlled, and the `av_shunting_vs_occlusion_paradox` discussion now attaches to that half specifically instead of to a bundle.

**Suggestion 2, I73.81 — taken, and you were right that it exists.** Worth recording how: OLS returns nothing at all for the `icd10cm` ontology, so `runoak -i sqlite:obo:icd10cm info ICD10CM:I73.81` was the check that settled it, returning `Erythromelalgia`.

I added it as a second row rather than replacing I73.8, because the existing row records the cross-reference MONDO itself asserts and replacing it would lose that provenance. It is `broadMatch` rather than `exactMatch`: I73.81 is "Erythromelalgia" unqualified, so it also covers the primary SCN9A form curated in `Primary_Erythermalgia`, and an I73.81-coded cohort still cannot serve as a case definition for this entry. Both justifications now point at each other.

**Suggestion 3 — not taken, agreed with your read.** PMID:27639908 is redundant with evidence already on the acquired-routes node, and the rest are PV/ET prognosis papers that belong to `Polycythemia_Vera` and `Essential_Thrombocythemia`. That is the same deferral the entry already makes for the MPN pathophysiology itself.

Validation on the new head: `just validate-disorders` passes with 116/116 snippets verified, up from 110. All ten offline gates green. The causal graph builds with 26 nodes and 18 edges and an empty orphan-target set, so the split introduced no phantom nodes. The 18 pytest cases parametrized on this entry pass, as do the 33 history tests.

One thing I did not run to completion, so it should not be read as checked: a full `pytest tests/test_data.py` over the whole corpus exceeds this environment's command timeout. The per-entry and per-gate checks above are what actually ran.

---

Issue #8548, opened 2026-08-15
Curate sick sinus syndrome 2, autosomal dominant (MONDO:0008102)


sierra-moxon:
Curate a dismech entry for **sick sinus syndrome 2, autosomal dominant** ([MONDO:0008102](https://monarchinitiative.org/MONDO:0008102)) — familial sinus node dysfunction caused by mutation in **HCN4** ([hgnc:16882](https://www.genenames.org/tools/search/#!/?query=HCN4)), the pacemaker channel carrying the funny current (I_f) that sets diastolic depolarization rate in the sinoatrial node. Loss-of-function/dominant-negative HCN4 variants slow SA-node automaticity, producing bradycardia, chronotropic incompetence, sinus pauses and syncope; some HCN4 variants co-segregate with left ventricular noncompaction.

Strong module fit: this is the sinoatrial-node automaticity-failure branch of `cardiac_ion_channel_repolarization` (a bradyarrhythmia rather than tachyarrhythmia channelopathy) — declare `conforms_to` against that branch. Also check membership in the `Inherited_Arrhythmia_Syndromes` grouping, which already carries a familial sick sinus syndrome criterion.

Priority dashboard action: `CURATE_ROOT` (score 58.0, root candidate; parent *familial sick sinus syndrome*). Note `kb/disorders/Sinoatrial_Block.yaml` exists and should be cross-referenced, not duplicated.

Source: [priority dashboard](https://dismech.monarchinitiative.org/dashboard/priority.html).

Tracker: part of #1079.

PR #10683, opened 2026-09-03
Augment Sick Sinus Syndrome 2 (HCN4) with epidemiology, models, datasets and knowledge gaps

tannerzhang:
## Summary

`Sick_Sinus_Syndrome_2_Autosomal_Dominant` already existed on `main` (created 2026-08-17 under #8548, 89.5% compliance). This is an augmentation pass on it, not a new entry. Compliance 89.5% → 99.7%.

## Deep research

A falcon report for this disorder was already committed, and re-running falcon would have overwritten the report the original curation was built from, so this pass ran **openscientist** and read the existing falcon report as the second provider.

The openscientist run validated clean: 16/16 references resolved, `confabulation_rate` 0.0, 24/24 CURIEs resolved. `needs_review` is true for one quote (PMID:18375593) that could not be matched beyond the abstract — that reference is not cited by the entry. `just preflight-dr` against MONDO:0008102 returned PASS (HCN4 56 mentions vs SCN5A 4, OMIM 163800 matching on both sides). The report's suggested names for `HP:0011702` and `HP:0011664` were paraphrases of the ontology labels and were not copied.

## What changed in the entry

- **`prevalence`** — one record only, the Danish nationwide familial risk of pacemaker-treated sinus node dysfunction. This is heritable SND at population scale and is not genotyped, which the record's `notes` and `directness: INDIRECT` both say. There is still no general-population prevalence for SSS2 and none is asserted.
- **`genetic.case_fractions`** — the HCN4 shares of ascertained cohorts (mutation-positive familial SSS probands; paediatric and adult isolated LVNC cohorts) go here rather than under `prevalence`, since they are genetic-spectrum shares, not population rates. The LVNC ones are ascertained through the structural phenotype rather than through bradycardia, so they carry `directness: INDIRECT`.
- **`progression`** — four phases: prenatal/neonatal bradycardia, asymptomatic childhood, symptomatic SND and device therapy, and aortic dilatation. The aortic phase is recorded as *not* reliably age-ordered, because the cited family has dilatation in two adolescents and none in their 50-year-old mother.
- **`phenotypes`** — biventricular noncompaction with preserved systolic function, mitral valve prolapse, prolonged QT interval. Quantitative exercise-test evidence added to chronotropic incompetence; a haploinsufficiency route added to atrial fibrillation.
- **`genetic.variants`** — four new alleles (D553N, 695X, the c.1737+1 G&gt;T splice allele, P883R), with `functional_effects` on all nine and `clinical_significance: UNCERTAIN_SIGNIFICANCE` on P883R only.
- **`experimental_models`** (new section) — the patient-derived hiPSC gene-dose pair and the wild-type hiPSC pacemaker-cardiomyocyte I_f platform.
- **`datasets`** (new section) — three GEO accessions, all verified with `just verify-datasets`. None is an SSS2 patient cohort and each record says so; the mouse accession is a gain-of-function allele and is labelled as the opposite-direction control.
- **`discussions`** (new section) — five: the unexplained aortic-wall mechanism, developmental vs rate-driven noncompaction, the HCN4 variant-causality threshold, a `HUMAN_MODEL_MISMATCH` on chronotropic incompetence, and anticoagulation without documented atrial fibrillation.
- **`treatments`** — stroke prevention in atrial fibrillation, scoped to *documented* AF because the cited meta-analysis states in terms that anticoagulation in SND is untested.
- **Causal-edge evidence** on 11 of the 12 `downstream` edges. The compaction→aortic-dilatation edge is left uncited on purpose: that mechanism is genuinely unknown, and the knowledge-gap discussion carries it instead.
- **Cleanup** — nine `explanation` strings still argued for the retired `supports: PARTIAL` value (the #10003 prose backlog). Rewritten, with the inference step encoded as `directness: INDIRECT` where that is what they meant.

## Pre-PR review

A red-team review against `.claude/skills/dismech-pr-review` ran before the first commit and raised 20 findings; all were acted on. The two that mattered most: the prevalence records initially asserted `prevalence_class: ABOVE_1_IN_1000` and `rate_per_100000` on ascertainment-conditioned cohorts, which would have exported as SSS2 population prevalence (hence the move to `case_fractions`); and biventricular noncompaction was banded `FREQUENT` off a denominator conditioned on already having noncompaction, now `OCCASIONAL`.

## Review rounds

Eight commits, six review rounds after the first approval. Every suggestion was optional; all were acted on.

1. **`PMID:31481236` cached but uncited.** Reading it, HCN4 p.Pro883Arg is the one allele here that does not fit the loss-of-function pattern — it shifts activation to more *depolarised* potentials and *raises* current density in the heterozygous configuration, and at ~0.9% minor allele frequency is too common to be a fully penetrant cause. One paper has it segregating with the combined SND and noncompaction phenotype; the functional study reads it as an atrial fibrillation modifier. Curated as a variant carrying both readings, with a matching `REFUTE` item against the "all mutations described so far are loss-of-function" statement in the `genetic` block. Also added `PMID:19166829` (Shox2 → Tbx3 → Hcn4) to the developmental-vs-rate discussion with both its limits stated, and a note on why `PMID:27173043` and `PMID:33185997` stay uncited (empty cache bodies).
2. **P883R's discordance was prose-only.** `functional_effects` populated on all variants, `clinical_significance` on P883R alone — grading the others `PATHOGENIC` would assert what no cited source here says.
3. **Self-caught.** Adding variants had left the `features` paragraph listing six alleles and calling the set uniformly loss-of-function, next to one now tagged gain-of-function. Refreshed, naming P883R as the exception.
4. **Type spelling and direction-vs-mechanism.** Verified the census independently (18 hyphenated vs 4 spaced) and moved to the majority form; the three dominant-negative alleles now also carry the loss-of-function direction, since dominant-negative is a mode of dominance rather than a direction.
5. **Prose naming the schema.** The `features` paragraph named a slot and an enum value, and P883R's effect description referred to a `REFUTE` item — the same #10003 failure mode this PR otherwise cleaned up. Rewritten to state only biology. 695X curated, its consequence recorded as inferred rather than measured.
6. **Curation rationale out of rendered prose**, and the two inferred alleles' `function` fields switched from naming the lesion to naming the consequence, with the inference hedge left intact.

One sentence of similar shape was deliberately kept and the reasoning stated: the `Impaired Ventricular Trabecular Compaction` node's argument that noncompaction is a parallel branch of the genetic lesion rather than remodeling secondary to the arrhythmia. That is mechanistic model reasoning, not process bookkeeping. The reviewer independently checked it against `origin/main` and agreed.

Two schema gaps surfaced along the way and were filed as [#10686](https://github.com/monarch-initiative/dismech/issues/10686) rather than settled here: `FunctionalEffect.type` is free text carrying 13 spellings across `kb/`, and `FunctionalEffect` has no `evidence` slot or measured-versus-inferred marker, which is why the two inferred alleles can only record that distinction in free text.

## Validation actually run

`just validate`; `just validate-terms` (passed); `count-verified-snippets` **139/139**; `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms` plus `check-qualifier-terms-online`, `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` (one new divergence found and fixed), `check-environmental-evidence`, `check-source-defect-claims` (nothing reported against this file); `just verify-datasets` 3/3 OK; `check-groupings` on `Inherited_Arrhythmia_Syndromes` (SATISFIED); `just validate-disorders` as the final batched sweep; `tests/test_data.py` 47757 passed, re-run after each round of edits.

Commit 2 adds reference caches that the entry and the research report cite but that were fetched after commit 1's file list was assembled — without them reference validation would fail on a clean checkout while passing locally.

https://claude.ai/code/session_019unfsvh7hp9w9gTMjvE5or

### DIFF
--- a/history/disorders/Sick_Sinus_Syndrome_2_Autosomal_Dominant/2026-09-03T115948Z-claude-code-208882.yaml
+++ b/history/disorders/Sick_Sinus_Syndrome_2_Autosomal_Dominant/2026-09-03T115948Z-claude-code-208882.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Sick_Sinus_Syndrome_2_Autosomal_Dominant
+  path: kb/disorders/Sick_Sinus_Syndrome_2_Autosomal_Dominant.yaml
+session:
+  id: 2026-09-03T115948Z-claude-code-208882
+  timestamp: '2026-09-03T11:59:48Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10683
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - genetic
+  summary: Name the consequence rather than the lesion on the splice allele's functional
+    effect
+  details: |
+    Acts on the single optional suggestion in the seventh approving review of PR #10683, which caught an incompleteness in the previous commit rather than a new issue. That commit changed 695X's functional_effects function field from naming the lesion, C-terminal truncation, to naming the consequence, reduced pacemaker current, and did not apply the same change to the other inferred-consequence allele in the block. The c.1737+1 G>T splice variant carried function canonical splice-donor disruption, which is the variant class its description already states, so it had the identical mismatch. Changed to reduced pacemaker current. The description keeps the epistemic hedge, that the consequence is inferred from the canonical +1 splice-donor class and from the family's bradycardia phenotype rather than measured, since that family was characterised clinically and not by patch clamp; the hedge is the part that matters and it survived the rewrite in both alleles. A sweep of all twelve function values in the block now finds every one naming what the channel does differently rather than what the variant is. NOT ACTED ON: the spelling spread in FunctionalEffect.type across kb/ remains a schema question about whether the slot should carry a controlled vocabulary, which both the reviewer and this session agree is not a curation PR's to settle. No evidence item, snippet, reference cache, ontology binding, causal edge or variant record was added or removed; the change is one free-text string and its description. VALIDATION ACTUALLY RUN: just validate; count-verified-snippets 139/139 unchanged; check-duplicate-keys, check-entity-refs, check-causal-targets, check-enum-values, check-snippet-grading, check-title-snippets, check-snippet-length, check-folded-hyphens; validate-disorders as the final batched sweep. Compliance holds at 99.7 percent.
--- a/kb/disorders/Sick_Sinus_Syndrome_2_Autosomal_Dominant.yaml
+++ b/kb/disorders/Sick_Sinus_Syndrome_2_Autosomal_Dominant.yaml
@@ -1856,11 +1856,11 @@ genetic:
       increased short-term heart rate variability even after normalisation for
       rate.
     functional_effects:
-    - function: canonical splice-donor disruption
+    - function: reduced pacemaker current
       description: >-
-        A canonical +1 splice-donor change, so the functional consequence is inferred
-        from the variant class and the family's bradycardia phenotype rather than
-        measured: this allele was characterised clinically, not by patch clamp.
+        Inferred from the canonical +1 splice-donor class and from the family's
+        bradycardia phenotype rather than measured: this allele was
+        characterised clinically, not by patch clamp.
       type: loss-of-function
     gene:
       preferred_term: HCN4


tannerzhang:
Acted on all three suggestions. Pushed as f8b827d, which resets the approval.

**Suggestion 1 turned out to be more than a bookkeeping gap.** `PMID:31481236` is now curated as a variant, `HCN4 p.Pro883Arg`, rather than explained away — because it is the one allele in this entry that does not behave like the others. `PMID:25145518` reports it segregating with the combined sinus node dysfunction and noncompaction phenotype in an unrelated proband. But when it was characterised functionally, it shifted activation to *more depolarised* potentials and, co-expressed with wild-type subunits as in a heterozygote, produced *higher* current density — the opposite direction from every other allele here. Its minor allele frequency is around 0.9%, which is too common for a fully penetrant cause on its own, and the functional study reads it as a modifier acting with a common KCNE1 variant to favour atrial fibrillation.

Both readings are curated: the segregation observation as `SUPPORT`, the electrophysiology as `REFUTE`, the modifier conclusion as `SUPPORT`. A matching `REFUTE` item now sits in the `genetic` block against the `PMID:26835093` statement that "all mutations described so far are loss-of-function", and that item's explanation says the statement holds for every allele curated here except this one. That statement was previously unqualified in three places.

**Suggestion 2, with its limit stated.** `PMID:19166829` is now cited in the developmental-versus-rate-driven discussion. It shows Hcn4 sitting downstream of a transcriptional cascade — losing Shox2 abolishes Tbx3 and Hcn4 expression and leaves a hypoplastic sinoatrial node — which is what makes a developmental reading of the structural phenotype coherent at all. It is not written as evidence that the compaction phenotype is developmental, because the paper does not show that: the cascade it describes builds the sinoatrial node, not the compacted ventricular wall, and it is mouse. Both limits are named in the rationale and in the evidence explanation, and the item carries `directness: INDIRECT`.

**Suggestion 3 was already fixed, and the timing explains the mismatch.** The review job started at 08:54:53 and the body rewrite landed shortly after, so the review read the original auto-generated description. The current body names the fourth phase as "Aortic dilatation, onset not reliably age-ordered", matching the YAML.

**One more thing, prompted by the "3 unused" count.** A new `UNCITED REFERENCE CACHES` paragraph in `notes` records why `PMID:27173043` and `PMID:33185997` are cached but not cited. Both cache bodies are empty — zero bytes after the `## Content` heading — so there is no text to attach a snippet to. Worth flagging that `PMID:27173043` is Vermeer et al 2016, the primary report of ascending aortic dilatation in HCN4 families. It would be the better citation for that phenotype than the case report currently carrying it; the case report is used with `directness: INDIRECT` precisely because the primary is a JACC letter with no retrievable abstract.

Validation after the change: 138/138 snippets verified, `validate-terms` passes, `validate-disorders` passes, the offline `check-*` gates pass, `check-source-defect-claims` reports nothing against this file, and `tests/test_data.py` is 47757 passed. Compliance holds at 99.7%.

---

tannerzhang:
Done in ec3465e. The suggestion was right — the reason P883R is curated at all is that it doesn't behave like the others, and that fact was only readable by a human.

`functional_effects` is now populated on all eight variants, and `clinical_significance: UNCERTAIN_SIGNIFICANCE` on P883R.

| Variant | `functional_effects[0].type` | `clinical_significance` |
|---|---|---|
| p.Ser672Arg | loss of function | — |
| 573X (c.1631delC) | dominant negative | — |
| p.Gly480Arg | loss of function | — |
| p.Gly482Arg | dominant negative | — |
| p.Lys530Asn | loss of function | — |
| p.Asp553Asn | dominant negative | — |
| **p.Pro883Arg** | **gain of function** | **UNCERTAIN_SIGNIFICANCE** |
| c.1737+1 G&gt;T | loss of function | — |

**Two scope decisions worth stating, since both go beyond or below what was asked.**

The suggestion named P883R alone. I populated all eight, because filling the slot for one variant would make the layer answer "which HCN4 alleles are not loss-of-function" while leaving "which are loss-of-function" empty — a worse asymmetry than either populated or unpopulated. Each entry is a transcription of what that variant's own `description` and evidence already say, not new curation. The splice allele's entry says explicitly that its consequence is inferred from the variant class and the family's bradycardia phenotype rather than measured, since that family was characterised by Holter and exercise testing rather than patch clamp — recording an inferred effect as if it were a measured one is the failure mode this slot invites.

`clinical_significance` is set on P883R only. Grading the other seven `PATHOGENIC` would assert something no cited source in this entry actually states, and the suggestion was about making existing curation queryable rather than adding pathogenicity calls. P883R is different because the entry already carries evidence cutting both ways on it — that is what `UNCERTAIN_SIGNIFICANCE` is for.

Snippets unchanged at 138/138, since no evidence item was touched. `validate`, `validate-terms`, `validate-disorders`, the offline `check-*` gates and `tests/test_data.py` (47757 passed) are clean. Compliance holds at 99.7%.

---

tannerzhang:
One more, 368afb0 — and this one is a defect this PR introduced rather than an optional improvement.

The `genetic` block's `features` paragraph summarises the allele set. Three variants were added across the previous two commits without updating it, so it listed six alleles while the block held eight, and described the set as uniformly loss-of-function while one member now carries `functional_effects` type "gain of function" and `clinical_significance: UNCERTAIN_SIGNIFICANCE`. It now lists D553N and the splice allele among the loss-of-function alleles and names P883R as the single exception, with its disputed causal status stated.

The re-review judged the old wording defensible on the grounds that "sole established cause" does not strictly cover a disputed allele. That was right when it was written. It predates the machine-readable layer added in ec3465e, and a summary asserting uniform loss of function sitting next to a variant record tagged "gain of function" reads as a mismatch to anyone scanning the block, whether or not it parses as consistent.

No evidence item, snippet, ontology binding or causal edge was touched — snippets stay at 138/138. `validate-disorders` and `tests/test_data.py` (47757 passed) are clean, compliance holds at 99.7%.

That closes both halves of the last suggestion. Nothing outstanding from my side.

---

tannerzhang:
Pushed 39c9491, covering the suggestions from the last two reviews.

**Prose naming the schema — agreed, and it was mine.** The `features` paragraph named the slot and the enum value, and P883R's effect description referred to "the machine-readable form" and to a `REFUTE` item. Both sentences were written in this session, and both are the same failure mode as the #10003 backlog this PR spent nine explanations cleaning up. Rewritten to state only the biology: the allele's causal status is disputed, and the claim that every reported HCN4 disease allele is loss-of-function does not hold for it. Grepping the file, the only surviving `REFUTE` strings are actual `supports:` values.

**695X curated.** Its consequence is recorded as *inferred* from the truncating class and the segregating family rather than measured — PMID:25145518 presents no patch-clamp characterisation of this allele, unlike 573X in the same series. Evidence quotes that paper, already cached and already cited here.

**Spelling**, verified rather than assumed: a census across `kb/` gives 18 hyphenated `loss-of-function` against 4 spaced, and 6 hyphenated `gain-of-function` against 0 spaced. Your count was exact. This entry now uses the majority form.

**Direction vs mechanism** was the one that changed behaviour. Dominant-negative is a mode of dominance, not a direction, so 573X, G482R and D553N are now typed both ways:

| Variant | `functional_effects[].type` |
|---|---|
| p.Ser672Arg | loss-of-function |
| 573X | dominant-negative, loss-of-function |
| 695X | loss-of-function *(inferred)* |
| p.Gly480Arg | loss-of-function |
| p.Gly482Arg | dominant-negative, loss-of-function |
| p.Lys530Asn | loss-of-function |
| p.Asp553Asn | dominant-negative, loss-of-function |
| **p.Pro883Arg** | **gain-of-function** |
| c.1737+1 G&gt;T | loss-of-function *(inferred)* |

A loss-of-function query now returns eight of nine. The miss is P883R, which is the point.

**One thing I'm not doing.** The sixteen distinct spellings across `kb/` in that slot are a schema question about whether `FunctionalEffect.type` should have a controlled vocabulary. That is not a curation PR's to settle, and this entry now uses the majority form rather than widening the spread. Happy to file it as an issue if that is useful.

Snippets 139/139. `validate`, `validate-terms`, `validate-disorders`, all eight offline `check-*` gates, and `tests/test_data.py` (47757 passed) clean on this exact commit. Compliance 99.7%.

---

tannerzhang:
Pushed 78e9088 — both suggestions taken.

**Curation rationale out of rendered prose.** You're right that this was the wrong place for it. The 695X and P883R descriptions each ended by explaining why the record exists rather than what the allele does, and the Prolonged QT Interval phenotype carried two such sentences. That reasoning already lives in the history records. What remains in the descriptions is biology only: for P883R, that both readings stand in the literature and neither has been withdrawn; for the QT phenotype, that both reports are in genotyped carriers, that the distinction matters for drug avoidance, and that no pathophysiology node produces it so it sits outside the causal graph.

**One sentence of the same shape left in place, deliberately.** The `Impaired Ventricular Trabecular Compaction` node opens "It is curated here as a parallel branch of the same genetic lesion rather than as maladaptive remodeling secondary to the arrhythmia: the noncompaction is present from childhood and does not follow a preceding phase of cardiomyocyte injury and neurohormonal activation." That is pre-existing content from the original August curation, and it is a different kind of claim — a mechanistic argument for how the node is modelled, grounded in an observation about onset. A curator reading the pathograph needs it. Happy to be overruled if you read it as the same defect.

**695X effect field** now names the consequence ("reduced pacemaker current") like the other eight rather than the lesion, with the description still stating that this is inferred from the truncating class and the segregating phenotype rather than from a functional assay.

Snippets unchanged at 139/139 — no evidence item, snippet, ontology binding, causal edge or variant record was added or removed. `validate`, `validate-disorders` and the eight offline `check-*` gates are clean on this commit.

**Where I think this stands.** Six rounds in, the remaining open item is the sixteen distinct spellings of `FunctionalEffect.type` across `kb/`, which we both read as a schema question about a controlled vocabulary rather than a curation one. I don't plan further changes to this entry unless something blocking turns up. Say the word if you'd like the vocabulary question filed as its own issue.

---

tannerzhang:
Fixed in 526ccdf. You caught an incompleteness in my own previous commit rather than a new issue — I changed 695X from naming the lesion to naming the consequence and didn't apply the same change to the other inferred-consequence allele one entry over. `c.1737+1 G>T` is now "reduced pacemaker current", with the hedge intact: the consequence is inferred from the canonical +1 splice-donor class and the family's bradycardia phenotype rather than measured, since that family was characterised clinically and not by patch clamp.

I swept all twelve `function` values in the block afterwards. Every one now names what the channel does differently rather than what the variant is:

| Variant | `type` | `function` |
|---|---|---|
| p.Ser672Arg | loss-of-function | hyperpolarizing shift of activation voltage |
| 573X | dominant-negative | loss of cAMP-dependent gating |
| 573X | loss-of-function | loss of cAMP-dependent acceleration of the pacemaker current |
| 695X | loss-of-function | reduced pacemaker current |
| p.Gly480Arg | loss-of-function | reduced channel synthesis and membrane trafficking |
| p.Gly482Arg | dominant-negative | nonfunctional pore with suppression of wild-type current |
| p.Gly482Arg | loss-of-function | reduced pacemaker current |
| p.Lys530Asn | loss-of-function | heteromer-restricted hyperpolarizing shift of activation |
| p.Asp553Asn | dominant-negative | trafficking defect reducing surface channel |
| p.Asp553Asn | loss-of-function | reduced surface pacemaker current |
| p.Pro883Arg | gain-of-function | depolarizing shift of activation with increased current density |
| c.1737+1 G&gt;T | loss-of-function | reduced pacemaker current |

Snippets unchanged at 139/139 — no evidence item, cache, ontology binding, causal edge or variant record touched. `validate`, `validate-disorders` and the eight offline `check-*` gates are clean on this commit.

Thanks for confirming the one sentence I declined to remove. Your read matches mine: it is a mechanistic argument about how the node is modelled, not process bookkeeping, and deleting it would lose reasoning a reader needs.

That leaves nothing open from my side except the `FunctionalEffect.type` vocabulary question, which we both put outside this PR's scope.

---

tannerzhang:
Filed as [#10686](https://github.com/monarch-initiative/dismech/issues/10686), with both parts.

Verified the numbers rather than carrying them over: `FunctionalEffect.type` has 57 uses across 21 files under 13 distinct spellings, and I listed which files still sit on a minority form so the normalisation is actionable once the vocabulary question is settled. Also confirmed your second point directly against the schema — `FunctionalEffect` has no `evidence` slot and no measured-versus-inferred marker, and `type` has range `string`.

One thing I added that seemed worth recording: `FunctionalImpactEnum` already exists for `GeneticContext.functional_impact_category` and covers most of the same ground. That makes "bind `type` to the existing enum" the obvious first proposal — but not an automatic one, since values like `toxic RNA gain-of-function` and `gain of ectopic expression` carry mechanism detail an enum would flatten. The issue says that rather than assuming it.

Nothing further from me on this PR.

---

PR #10696, opened 2026-09-03
fix: pass the disease's MONDO ID to deep-research runs

gingin77:
Every deep-research run this repository dispatches goes out with a disease **name** and no identifier. Two of the three research recipes hardcoded it empty:

```
--var "disease_name=$disease_name" \
--var "mondo_id=" \                     # <- project.justfile
--var "category=$category" \
```

This makes all three recipes read the entry's own `disease_term.term.id` and pass it.

## Why it matters

It caused a provider to research the wrong disease.

An openscientist run for **"Cardiomyopathy Dilated 2H"** returned a well-organised report naming **RPL3L** as the causal gene — 44 mentions of RPL3L, zero of ASNA1/GET3. RPL3L is cardiomyopathy dilated 2**D** (OMIM:619371), not 2**H** (OMIM:620203, GET3). Both are recessive, neonatal-onset, rapidly progressive and fatal without transplant; the names differ by one letter and CMD2D is better documented. Given only the name, the provider picked the better-documented one.

**Verified by re-running with the identifier as the only variable:**

| Run | `mondo_id` | RPL3L | ASNA1/GET3 |
|---|---|---|---|
| Original | `''` | 44 | 0 |
| Retry via trailing `--var` | `''` | 44 | 0 |
| With this fix | `MONDO:0859358` | **0** | **49** |

The corrected report confirms the mechanism unprompted — it records that early iterations followed the *GATAD1* literature (CMD2B, OMIM:614672) and that cross-referencing MONDO→OMIM→gene corrected it. So this disease name sits between **two** decoys, and the identifier is what separates them.

**Nothing downstream catches this.** The wrong-disease report's own reference validation returned 19/19 resolved, 0 off topic — relevance is scored against the report's *own* vocabulary, so a report built around the wrong disease scores its wrong-disease citations as on topic. Term validation was equally quiet: the RPL3L identifiers are all correct *for RPL3L*.

Detail in #10495; the affected curation is #10494.

## What changed

**Extraction is a YAML parse, not a line window.** Two window-based attempts were tried and both were wrong:

| Extraction | Empty | Wrong |
|---|---|---|
| `grep -A3` | 20 | 0 |
| `grep -A6` | 6 | 0 |
| `grep -A8` | 0 | **2** |
| `yaml.safe_load` | 0 | **0** |

A window has no idea where the `disease_term` block ends. `-A3` missed 20 entries whose block carries a `description:` (Marfan, seven Noonan-family entries, Kawasaki, Hirschsprung). `-A8` fixed those but read *past* the block on two entries with a deliberately unbound `disease_term`, returning an ID the entry itself rejects — a "do not use this term" comment in `Acute_Post-Surgical_Pain`, and a `skos:closeMatch` cross-reference in `CKD-Mineral_Bone_Disorder`. CLAUDE.md is explicit that `closeMatch` is a cross-reference and not disease identity. Handing a provider a *wrong* ID is worse than handing it none, which is the failure this PR exists to prevent.

**A guard against the ontology root.** `Dorsalgia` binds `MONDO:0000001` — the root of MONDO, labelled "disease". Sending that would assert the disease's identifier is the root of the ontology. The guard is adopted verbatim from the repo's existing one at `src/dismech/export/mondo_emc_export.py:80`. It is the only entry KB-wide that does this; the entry's own binding is arguably a curation defect that should also be fixed at source.

**`research-datasets` was rewritten too**, so all three recipes share one extraction. It already worked, so the behaviour change was measured: **21 entries change and every one is an improvement** — 20 previously returned empty and now resolve correctly, and `Dorsalgia` goes from the ontology root to empty. No regressions.

**The identifier is now visible at dispatch**, because `2>/dev/null || echo ""` otherwise collapses "no bound term" and "the extractor malfunctioned" into the same silent empty string:

```
Researching: Marfan_Syndrome [MONDO:0007947] (openscientist) -> ...
Researching: Dorsalgia [no MONDO ID] (openscientist) -> ...
```

**A test file**, `tests/test_research_recipe_mondo_id.py`. Five tests: no recipe hardcodes an empty `mondo_id`; every recipe forwards the extracted value; **no recipe uses a line window at all**; extraction matches the parsed `disease_term.term.id` for every entry in the corpus; and the named edge cases resolve correctly.

The behavioural tests read the one-liner out of `project.justfile` rather than restating it — an earlier version hardcoded a copy, so it was verifying my transcription rather than the shipped code. The whole-corpus sweep carries the `kb_data` marker so it runs parallel in `just test-kb` and the nightly sweep rather than costing seven serial minutes in the fast lane, which is now 0.22s. Deleting the root guard still fails the fast lane, so a code-only PR keeps the coverage that matters.

## What this does not cover

- **`asta` still dispatches name-only.** Its template is a single-line free-text retrieval query using `{disease_name}` alone. That is probably correct — a bare CURIE would more likely hurt recall than help — but it means coverage is not blanket.
- **`research-comorbidity` has the same bug**, name-only for two diseases at once. It needs a template change, so it is left for its own issue.
- **An identifier is not always sufficient.** 39 entries share 17 identifiers — five NSCLC strata all dispatch as `MONDO:0005061` — so for near-identical siblings the ID alone will not disambiguate. This removes a whole class of confusion, not every instance of it.

https://claude.ai/code/session_01NgzW11dBNCm5ZjtgqMzf7T

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -1880,7 +1880,7 @@ research-datasets provider disorder *args="":
     mondo_id=$(uv run python -c "import sys,yaml;d=yaml.safe_load(open(sys.argv[1])) or {};t=(d.get('disease_term') or {}).get('term') or {};i=(t.get('id') or '').strip() if isinstance(t,dict) else '';print(i if i.startswith('MONDO:') and i != 'MONDO:0000001' else '')" "$yaml_file" 2>/dev/null || echo "")
     output_file="{{research_dir}}/datasets/{{disorder}}-datasets-{{provider}}.md"
     requested_provider="{{provider}}"
-    echo "Dataset discovery: $disease_name ({{provider}}) -> $output_file"
+    echo "Dataset discovery: $disease_name [${mondo_id:-no MONDO ID}] ({{provider}}) -> $output_file"
     provider_arg=$([[ "{{provider}}" == "cborg" ]] && echo "--use-cborg" || echo "--provider {{provider}}")
     {{dr_client}} research \
         --template {{templates_dir}}/disease_datasets_research.md \
@@ -1995,7 +1995,7 @@ research-disorder provider disorder *args="":
     output_file="{{research_dir}}/{{disorder}}-deep-research-{{provider}}.md"
     requested_provider="{{provider}}"
     template_file=$([[ "{{provider}}" == "asta" ]] && echo "{{templates_dir}}/disease_pathophysiology_research_asta.md" || echo "{{templates_dir}}/disease_pathophysiology_research.md")
-    echo "Researching: $disease_name ({{provider}}) -> $output_file"
+    echo "Researching: $disease_name [${mondo_id:-no MONDO ID}] ({{provider}}) -> $output_file"
     provider_arg=$([[ "{{provider}}" == "cborg" ]] && echo "--use-cborg" || echo "--provider {{provider}}")
     {{dr_client}} research \
         --template "$template_file" \
@@ -2229,7 +2229,7 @@ research-disorder-cyberian-codex disorder *args="":
     category=$(grep "^category:" "$yaml_file" | head -1 | sed 's/category: *//' || echo "")
     output_file="{{research_dir}}/{{disorder}}-deep-research-cyberian-codex.md"
     requested_provider="cyberian-codex"
-    echo "Researching: $disease_name (cyberian-codex) -> $output_file"
+    echo "Researching: $disease_name [${mondo_id:-no MONDO ID}] (cyberian-codex) -> $output_file"
     {{dr_client}} research \
         --template {{templates_dir}}/disease_pathophysiology_research.md \
         --var "disease_name=$disease_name" \


gingin77:
Correction to the last section of this PR body: **the re-run I described as "in flight" did not test this fix, because it never received the MONDO ID.**

I launched it as `just research-disorder openscientist Cardiomyopathy_Dilated_2H --var mondo_id=MONDO:0859358`, on the assumption that the trailing `*args` would override the hardcoded empty value. It does not — the report came back with `mondo_id: ''` in its frontmatter and, again, 44 mentions of RPL3L and zero of ASNA1/GET3.

So that run reproduced the original failure rather than testing the remedy. Two things follow:

1. **This PR is still the right fix and is now the *only* way to supply the ID.** The `*args` escape hatch I assumed existed doesn't work, which makes the hardcoded empty value harder to work around than I thought, not easier.
2. **Whether supplying the ID is sufficient is still unverified.** I am re-running now with the patched recipe applied locally, and will report the result here either way — including if it comes back RPL3L again, which would mean the identifier alone is not enough and the real answer is a `just preflight-dr` gate in the curation path.

I would rather flag this than leave a claim standing that this PR had been validated end-to-end when it hasn't.

gingin77:
**Tested, and the result is clean.** Same provider, same disease name, same prompt template — the identifier was the only variable:

| Run | `mondo_id` | RPL3L | ASNA1/GET3 |
|---|---|---|---|
| Original | `''` | 44 | 0 |
| Retry with trailing `--var` | `''` | 44 | 0 |
| **This fix applied** | `MONDO:0859358` | **0** | **49** |

The corrected report names GET3/ASNA1 as the causal gene and cites the right primary literature. Report and validation are on #10494.

**The corrected report also confirms the mechanism from the other side, unprompted.** It contains this:

> Early iterations of this investigation initially followed the GATAD1 literature; identifier cross-referencing through Monarch/OLS4 established that MONDO:0859358 → OMIM:620203 → GET3/ASNA1, and the report was corrected accordingly.

So the provider *did* go down a wrong path, and used the MONDO ID to get back. GATAD1 is CMD2B (OMIM:614672) — meaning "Cardiomyopathy Dilated 2H" sits between at least **two** decoys, CMD2D/RPL3L and CMD2B/GATAD1, and the identifier is what separates them. Without it, which decoy you land on is luck.

That is stronger evidence than I expected to be able to offer for a two-line change. The failure is reproducible, the fix is verifiable, and the provider's own reasoning trace names the identifier as what resolved it.

One correction to my earlier comment: I said the run I'd launched with a trailing `--var mondo_id=...` would test this. It didn't — `*args` does not override the hardcoded value, and that run came back with `mondo_id: ''` and reproduced the original failure. The result above came from applying this PR's patch locally in the CMD2H worktree, running, and reverting the patch immediately (`project.justfile` is unmodified on that branch — verified before commit).

The practical upshot is that the hardcoded empty value **cannot be worked around from the command line at all**. This PR is the only route.

gingin77:
Good catch, and it goes one step further than proposed: **`-A6` is not enough either.** I measured every window against the YAML-parsed `disease_term.term.id` for all 2,499 entries with a bound term:

| Window | Misses | Mismatches |
|---|---|---|
| `-A3` | 20 | 0 |
| `-A6` | **6** | 0 |
| `-A8` | **0** | 0 |

`-A6` still drops `CBL-related_Disorder`, `DLG4-Related_Synaptopathy`, `KDM6B-Related_Neurodevelopmental_Disorder`, `Legius_Syndrome`, `Noonan_Syndrome_11` and `Noonan_Syndrome_with_Multiple_Lentigines` — four of which are in the set your own comment named. Those blocks carry a longer `description:`. Using `-A8`, applied to `research-datasets` as well so the three stay verbatim-identical.

Spot-checked the previously failing entries: Marfan `MONDO:0007947`, Noonan `MONDO:0018997`, Legius `MONDO:0012669`, CBL-related `MONDO:0013308`, Noonan 11 `MONDO:0032786` — all resolve now.

**On why my verification table missed this** — worth stating plainly, because the check was wrong in kind, not just in coverage. I sampled four entries and all four happened to have no `description:` in the `disease_term` block, so the sample was blind to the only failure mode the change has. That includes Marfan_Syndrome, which `CLAUDE.md` uses as its own worked example for `research-datasets` and which is in the failing set. For a change whose failure mode is structural, a whole-corpus comparison against the parsed value is the right check and four hand-picked entries never was. That's the comparison in the table above, and it's how I'd verify this kind of change from the start.

**One thing worth deciding rather than leaving implicit.** `-A8` is verified at 0/0 across the corpus today, but it is still a line-window heuristic over YAML — a longer `description:` in some future entry would silently reopen the same hole, and silently is the operative word. A parser-based extraction would be exact:

```
mondo_id=$(uv run python -c "import sys,yaml;d=yaml.safe_load(open(sys.argv[1]));print((((d or {}).get('disease_term') or {}).get('term') or {}).get('id') or '')" "$yaml_file")
```

I did not do that here because it changes the shape of an existing working recipe and adds a Python startup to every research invocation, and because your instruction was to keep the three identical — which `-A8` does. Happy to switch if you would rather have exactness than minimality; it is a one-line swap in three places.

gingin77:
You're right, and I've taken the option you offered rather than a third heuristic: **the recipes now parse the YAML.**

`-A8` was still wrong, and wrong in the way that matters most for this PR — handing a provider an ID the entry itself rejects is worse than handing it none. Both cases you found are exactly that: a comment whose purpose is to say *don't use this term*, and a `skos:closeMatch` whose own justification says it covers only the bone component. CLAUDE.md is explicit that closeMatch is a cross-reference and not disease identity, so the second one is a rule violation and not just a parsing miss.

Verified against the parsed `disease_term.term.id` for all 2,535 files:

| Extraction | Empty | **Wrong** |
|---|---|---|
| `-A3` | 20 | 0 |
| `-A8` | 0 | **2** |
| Parsing | 0 | **0** |

Marfan, Noonan and the rest of the recovered 20 still resolve; `Acute_Post-Surgical_Pain`, `CKD-Mineral_Bone_Disorder` and the two pituitary-adenoma siblings all correctly return empty. Block-scoping via `sed` would also have worked — I went with parsing because it retires the whole class of question rather than moving the boundary, and it cannot regress when a future entry grows a longer `description:` or a new sibling key. The Python startup is irrelevant against a research run measured in minutes, as you say.

**Added the test you asked for**, `tests/test_research_recipe_mondo_id.py`, five cases: no recipe hardcodes an empty `mondo_id`; every recipe forwards the extracted value; **no recipe uses a line window at all**; extraction matches parsed truth for every entry in the corpus; and an unbound `disease_term` yields nothing. The third is the one that would have stopped both of my attempts, and the corpus test is deliberately whole-corpus — sampling four entries is precisely what let `-A3` through.

**On `asta`** — good catch, and I've deliberately not fixed it here. `disease_pathophysiology_research_asta.md` carries only `{disease_name}`, so adding the var to the recipe does nothing for that provider; it needs a template change, which is a different kind of edit with its own review surface. Happy to open it as a follow-up issue so it doesn't get lost.

Thanks for pushing on this twice. The first version of my verification was four sampled entries, which was the wrong shape of check for a structural failure mode, and the second was a wider window that fixed the visible symptom while introducing a worse one.

gingin77:
**The red `test (3.13)` here is a repo-wide outage, not this diff.** I have stopped retrying it.

The failing step is `just validate-terms-schema`, timing out against EBI:

```
could not reach ontology service to resolve NCIT:C2916:
```

The CURIE differs between attempts (`NCIT:C2929`, then `NCIT:C2916`), which is the signature of a slow service rather than a broken term. And it is not confined to this branch — every completed test run in the last hour fails at the same step, including three merge-queue runs:

| Time (UTC) | Branch |
|---|---|
| 17:16 | `claude/disease-claim-curation-open-scientist` |
| 17:07 | `gh-readonly-queue/main/pr-10727-ddfbfbed` |
| 17:07 | `gh-readonly-queue/main/pr-10161-3b1ba68c` |
| 16:59 | `gh-readonly-queue/main/pr-10727-98a708d8` |
| 16:58 | `gh-readonly-queue/main/pr-10204-3b1ba68c` |

This diff touches `project.justfile` and one new test file. `validate-terms-schema` validates the *schema's* dynamic enums and reads neither.

I retried four times across about 90 minutes and stopped, since it is an external service problem and retries do not help. Reported on #10396, which already describes the same 5-second OLS timeout in the deep-research path.

The recipe supports `--offline` against the local cache, which the error message itself suggests — but switching CI to it would validate the schema's ontology enums against a cache that could be stale, which is close to the thing you would least want to skip. That is a maintainer trade-off and not something to decide inside a PR about a different subject, so I have left it alone.

Nothing to do here from my side. The PR is approved and the diff is unchanged since that approval.

gingin77:
The EBI outage cleared at ~18:34 UTC, and with it gone the job got far enough to reach a real failure of mine: `ruff` `PLW1510` — `subprocess.run` without an explicit `check` argument — in the test I added last commit. Fixed in `8431124ba` with `check=False`, which was the intent throughout: the test reads stdout and compares it, and a non-zero exit is a result worth asserting on rather than raising.

Worth noting how it stayed hidden. The ontology timeout was failing the same job at an earlier step for about two hours, so `Lint with ruff` never ran. The outage masked my lint error, and clearing the outage revealed it. I read four consecutive failures as "still the outage" without re-reading the log each time — the fourth one had already changed cause.

`uvx ruff check .` and `pytest tests/test_research_recipe_mondo_id.py` (5 passed) are both green locally now.

Pushing this drops the approval, so this will need a re-review. Sorry for the extra round on a one-line fix.

gingin77:
Both fixed. `c252eb2b5`.

**The `Dorsalgia` case is a fair thing to hold this PR on.** You're right that it isn't a regression, and right that the underlying defect is the entry's own binding — but sending `MONDO:0000001` is a *wrong* identity rather than a missing one, and this is the PR whose whole argument is that a wrong identifier is worse than none. Adopted the repo's existing guard verbatim from `src/dismech/export/mondo_emc_export.py:80` rather than inventing one, at all three sites. `Dorsalgia` now returns empty; Marfan and the rest are unaffected.

**The test finding is the sharper of the two and it was mine.** `_extract` hardcoded the one-liner as a string literal while its docstring claimed to "run the same extraction the recipes use". So the corpus test — the one I cited in my last two comments as evidence the change was exact — was verifying my transcription of the code, not the code. A test that restates the thing it tests can only ever confirm the restatement, and I'd have caught a drift between the two never.

It now reads the one-liner out of `project.justfile` with a regex, asserts the three recipes still share a single extraction, and execs that string. Running it in-process rather than spawning a subprocess per file also takes the corpus test from ~5 minutes to ~3.

Added a sixth test asserting no entry sends `MONDO:0000001`, so the Dorsalgia case cannot return silently.

`uvx ruff check .` clean; 6 tests pass.

Worth noting the sequence on this PR, since it is not flattering and is the useful part: the corpus verification I offered as proof in rounds 2 and 3 was a sampled check the first time, a wider window the second, and a self-referential test the third. Each was wrong in a different way and each was presented as settled. The current version reads the shipped string and compares against `yaml.safe_load` across all 2,535 entries, which is the check I should have written at the start.

gingin77:
Fixed in `298aa3230`. You were right that the runtime half got *worse* — going in-process removed the subprocess spawns but added a second parse of every file.

```
fast lane   0.22s   (was 7m08s)
kb_data     2m01s   parallel
```

Marked the corpus exactness test `kb_data` and kept the three string assertions plus a named-entry test in the fast lane, covering the five entries that have actually broken: `Dorsalgia` (the root), `Acute_Post-Surgical_Pain` and `CKD-Mineral_Bone_Disorder` (unbound, read past the block), `Marfan_Syndrome` and `Noonan_Syndrome` (a `description:` inside the block).

**On your concern that a `project.justfile`-only change no longer triggers the full sweep** — I checked rather than assuming. Deleting the `MONDO:0000001` half of the guard from all three sites still fails the *fast-lane* test, naming `Dorsalgia`. So the coverage that matters for a code-only PR survives the split; the corpus sweep adds breadth over the other 2,530 entries, which is what the nightly backstop is for.

I did not take the `CSafeLoader` option — you're right that it's the wrong lane regardless, and halving a number that shouldn't be there isn't a fix.

**Two of the non-blocking items taken, both one-liners:**

- `_recipe_body` now anchors on `(?=[ :])` rather than `\b`. You're right it resolved correctly only because `research-disorder` happens to be defined before `research-disorder-cyberian-codex` — that's a coupling to file order, not a property of the regex.
- The empty-`mondo_id` assertion is now a whitespace- and quoting-tolerant regex, so a single-quoted or line-continued form can't reintroduce the bug and pass.

**Not taken here, with reasons:**

- **`asta` and `research-surrogacy`.** Real gaps, but both need template changes (`disease_pathophysiology_research_asta.md` carries only the name placeholder), which is a different review surface. I'd rather file it than widen this PR — say the word and I'll open the issue.
- **The `2>/dev/null` fallback.** Agreed it degrades silently into the condition being fixed. I left it because the current behaviour matches `research-datasets`, which has always had it, and adding an `echo` to all three is a behaviour change I'd rather make deliberately than as a rider. Happy to add it if you want it in this PR.
- **The 39-entries-sharing-17-identifiers point** is a good one and I've added nothing to the PR body for it yet. It's the honest limit of this change: for five NSCLC strata all dispatching as `MONDO:0005061`, the identifier alone won't disambiguate. Worth stating plainly rather than leaving the PR implying the ID is sufficient in all cases.

gingin77:
Both done. `a57fecc14` plus a full rewrite of the PR body.

**On the description — you're right that this one matters more than usual.** The body was describing `grep -A3` and claiming the extraction was "copied verbatim from `research-datasets`", when the shipped code is a YAML parse and `research-datasets` was itself rewritten. Since the repo squash-merges, that would have become the permanent commit message for a provenance fix, describing an approach abandoned mid-review. Rewritten to cover the parse and why it replaced the window (including the `-A8` step that was *wrong* rather than merely incomplete), the root guard, the 21-entry improvement in `research-datasets`, the test file and its lane split.

I also added a "what this does not cover" section with your three points: `asta` name-only, `research-comorbidity` name-only for two diseases, and — the one I think is most worth a reader seeing — that **39 entries share 17 identifiers**, five NSCLC strata all dispatching as `MONDO:0005061`. This removes a class of confusion, not every instance, and the body shouldn't imply otherwise.

**On the silent empty — agreed, and the argument is the PR's own.** The case for this change is that nothing downstream can detect a wrong-disease report; that applies to the fix's own failure mode too. Failing open stays right for the three entries that legitimately have no term, but the operator can now tell the two apart:

```
Researching: Marfan_Syndrome [MONDO:0007947] (openscientist) -> ...
Researching: Dorsalgia [no MONDO ID] (openscientist) -> ...
```

Applied to all three recipes.

**On the suggestions:**

The triplicated one-liner is a fair hit, and you're right that `assert len(codes) == 1` is the test admitting it. I've left it: collapsing it into a `scripts/` helper would break three of the five tests for no behaviour change, and doing that refactor inside a PR that has already changed implementation twice under review seems like the wrong moment. Worth its own PR where the test loosening can be reviewed on its merits.

`research-comorbidity` and the `asta` template I'll file as issues rather than widen this further. And I agree `Dorsalgia` binding `MONDO:0000001` is a curation defect in the entry, not a deliberate choice — this PR only defends against it. That's a third issue.

Thanks for measuring the `research-datasets` blast radius independently. I'd asserted the change was safe without quantifying it, and 21-entries-all-improvements is a better answer than the one I gave.

PR #10697, opened 2026-09-03
Add Coal Workers Pneumoconiosis (MONDO:0006654)

caufieldjh:
Adds coal workers' pneumoconiosis — black lung — as a new disorder entry. It was the one classic occupational mineral-dust disease the knowledge base was missing, alongside the existing Silicosis, Asbestosis and Chronic Beryllium Disease entries. Deep research was run with the `claude_code` provider. Everything validates: 103/103 evidence snippets verified, all gates green, compliance 85.4%.

Preflight found no duplicate — no existing entry, stub, PR or issue for MONDO:0006654 or its synonyms.

## What the entry contains

A 14-node causal chain running from dust deposition to respiratory failure:

```mermaid
graph LR
  A[Respirable coal mine<br/>dust deposition] --> B[Alveolar macrophage<br/>phagocytosis]
  A --> S[Silica-enriched<br/>dust burden]
  S --> B
  B --> C[Reactive oxygen<br/>species]
  B --> D[Dust-laden<br/>macrophage death]
  D --> E[Alveolar and small-airway<br/>tissue injury]
  C --> F[NF-kB / NLRP3<br/>inflammasome]
  C --> G[Epithelial injury<br/>and EMT]
  E --> H[Sustained inflammatory<br/>recruitment]
  F --> H
  H --> I[Fibroblast activation /<br/>myofibroblasts]
  G --> I
  I --> J[Coal macule<br/>formation]
  J --> K[Focal centrilobular<br/>emphysema]
  J --> L[Progressive massive<br/>fibrosis]
  S --> L
  K --> M[Impaired gas exchange /<br/>mixed ventilatory defect]
  L --> M
```

Plus 3 subtypes, 10 phenotypes (all connected to the graph), 2 environmental exposures, 3 susceptibility genes, 3 prevalence records, 2 biomarkers, 6 treatments, 1 clinical trial, and diagnosis, imaging, histopathology and differential-diagnosis sections. Five nodes conform to the `fibrotic_response` module, following the Asbestosis pattern.

The separate silica node is the mechanistically interesting part. CWP resurged in central Appalachia after decades of decline, and the reason is that thin-seam mining requires cutting through silica-bearing rock. Contemporary miners' lungs carry more silica than their historical counterparts', and silica-type PMF has gone from 18% to 57% of severe cases. The entry models that as its own node feeding both the macrophage response and PMF directly, rather than folding it into the coal dust exposure.

## The deep-research report's ontology suggestions were mostly wrong

Worth flagging for anyone else curating from a `claude_code` report. Every term identifier the report offered resolved to a real term, and several named something entirely different from what the report called it:

| Report called it | The ontology calls it |
|---|---|
| `HP:0002105` "Restrictive ventilatory defect" | Hemoptysis |
| `HP:0009777` "Rheumatoid nodules" | Absent thumb |
| `HP:0100750` "Pulmonary nodule" | Atelectasis |
| `HP:0031246` "Chronic cough" | Nonproductive cough |
| `NCIT:C64582` "Oxygen" | Hypopharyngeal Cancer pNX TNM Finding v6 and v7 |

The report had no `reference_validation` or `term_validation` block, so both were added retrospectively with `just validate-research-reference` and `just validate-research-terms`. Term validation caught two of these and flagged four more as worth a second look — but it only checked 14 of the report's 36 names, and the hemoptysis one was not among them. It was found by looking every term up in OAK directly. **The lesson is that a clean term-validation section is not permission to copy identifiers across.** Every binding here was re-derived from OAK.

Reference validation was clean: 31/31 resolved, none off topic. The report cites by URL and PMC identifier rather than PMID, and its citations sidecar contains only the prompt, so 21 PMC identifiers were converted through the PMC ID Converter and cited as PMIDs.

## What a pre-PR review caught

A red-team review was run against the entry before opening this PR. It returned changes-requested with six blocking findings, all of which turned out to be correct. Two are worth describing because no automated gate can catch either — both snippets pass the exact-substring check perfectly.

**A pronoun hid a wrong-subject quote.** The deposition node cited PMID:35831008 for `"Their aerodynamic size enables the particles to escape the upper respiratory tract's primary defences..."`. The sentence before it in that paper reads *"Fine and ultrafine particles that are ≤2.5 µm diameter (i.e. PM2.5 and PM0.1) are termed respirable PM."* — so "Their" means ambient urban particulate matter, not coal mine dust. The referent sat one sentence outside the quote, which is exactly why substring checking cannot see it.

**A truncated quote inverted its source.** The macrophage-death node cited PMID:16966090 as `"CD exposure increased alveolar expression of the proapoptotic mediator Bax"`. The full sentence begins *"In rats exposed to BNF,"* — the result holds only under a co-exposure that was silently dropped — and that paper concludes *"Bax and apoptosis were not the major determinants of early lung injury in this model"*. A refutation had been cited as support. It is now quoted in full with `directness: INDIRECT`, and the paper's contrary conclusion is curated beside it as a `REFUTE` item.

The other four:

- Nine of thirteen quotes from PMID:36581638 came from that paper's introduction, where it cites other people's work. The consequence was structural rather than cosmetic: the NLRP3 node and the declared central-effector fibroblast node had no primary and no coal-specific evidence at all. Both now cite the study's own in vivo and in vitro results.
- "PMF can progress after exposure ceases" was asserted in four places with nothing supporting it, while the nearest evidence said the opposite. Two verbatim sentences were already sitting in references the entry cites.
- Caplan syndrome was modelled as a CWP subtype. I think that was wrong: `MONDO:0005690` is a direct subclass of pneumoconiosis, which makes it a sibling of anthracosis rather than a child, and it arises on silicosis and asbestosis too. Because subtype terms count as curated, leaving it there would also have retired the term from the curation queue. Moved to `differential_diagnoses`.
- No diagnostic modelling, despite the entry defining its own subtypes in ILO radiographic terms. Added `diagnosis`, `imaging_findings` and `histopathology`.

Snippet count went from 75 to 103 across that pass.

## Deliberate omissions

**No `datasets:` block.** `just discover-datasets` returned 13 candidates, all gene-only matches through TNF, TGFB1 and H19, and all about unrelated diseases — Crohn's, macular degeneration, ovarian cancer. A direct GEO search found one plausible series, GSE311671, whose title says "Coal Mine Dust" but whose summary and stated design are a silica suspension murine silicosis model. Adding it would have been the named-entity-confusion failure `docs/dataset-curation.md` warns about.

**No IARC carcinogen group on the coal dust exposure.** IARC evaluated coal dust in Monographs volume 68 as Group 3, and the schema enum does admit `GROUP_3`. The omission is a judgement, not a schema limit: that listing is for coal dust as an agent, while the exposure curated here is coal *mine* dust, a mixture whose silica fraction carries its own classification. The exposure's `notes` say so. The coal-dust/lung-cancer association is curated as evidence instead.

**ILO item 2.1.9 not claimed.** That item covers COPD from coal dust, which is a different disease outcome of the same exposure and belongs to the COPD entry. Item 2.1.1 is claimed, since its text names anthraco-silicosis.

**No GeneReviews baseline.** A PubMed search returned zero hits, as expected for a purely acquired occupational disease.

**Most causal edges carry no edge-level evidence.** Edge evidence was added only for the edges where a source makes the causal claim itself. The rest were left uncited rather than reusing a node's citation to cover an edge that has no source.

## Known gaps

The 0.33%-to-3.2% PMF prevalence figure is quoted from the introduction of PMID:35353671, where that paper is citing six others. A caveat in the entry says so and points at the primary surveillance reports, but the figure was not re-sourced — PMID:30114941 is fetched and would do it.

Several fetched references remain uncited: the GBD 2019 global burden analysis (the entry's prevalence records are US-only while its description claims worldwide scope), a coal-versus-silica particle composition study, and two rodent model papers that would support an `animal_models` section this entry does not have.

Gene wiring to mechanism is 0%, because the `Genetic` class has no mechanism-target slot. Every sibling dust-disease entry reports the same.

The EMT branch carries a `HUMAN_MODEL_MISMATCH` discussion. It rests entirely on one study whose in vitro arm used A549 — a lung adenocarcinoma line standing in for alveolar type II cells — and whose in vivo arm used coal dust nanoparticles rather than respirable mine dust, while the node asserts type-II EMT in human miners. Two resolving experiments are proposed.

## Validation

```
just validate-disorders kb/disorders/Coal_Workers_Pneumoconiosis.yaml   PASS, 103/103 snippets
check-entity-refs, check-causal-targets, check-duplicate-keys           OK
check-enum-values, check-qualifier-terms, check-qualifier-terms-online  OK
check-snippet-length, check-title-snippets, check-folded-hyphens        OK
check-snippet-grading, check-environmental-evidence                     OK
compliance                                                              85.4%
```

For comparison: Silicosis 79.4%, Chronic Beryllium Disease 79.9%, Asbestosis 88.6%.

A history record at `history/disorders/Coal_Workers_Pneumoconiosis/` carries three events (create, edit, review) with the full reasoning, including the findings acknowledged but not acted on.

https://claude.ai/code/session_014fxxHB3foDi2wPT3ZW91Ez

---
_Generated by [Claude Code](https://claude.ai/code/session_014fxxHB3foDi2wPT3ZW91Ez)_


caufieldjh:
Fixed the blocking finding and acted on three of the five suggestions. Pushed as 6fef93c. Snippets are now 114/114 verified, up from 103, and all gates stay green.

**The truncated snippet was my error, and the diagnosis is worth recording.** I had read that abstract through a `cut -c1-700` terminal pipe and pasted the snippet out of the truncated display rather than out of `references_cache/PMID_11952275.md`. The cut happened to land mid-word at "than d". Every validator passed it because a truncation of an exact quote is still an exact substring. You are right that the comparator is the claim — a `MODULATES` edge asserting that smoking adds burden on top of dust-driven impairment needs the "than did their non-smoking counterparts" half to say anything.

Since the cause was mechanical rather than a one-off, I swept the whole entry for it: for each snippet, check whether the character immediately following it in the cached text is alphanumeric. One real hit — this one. Three apparent hits on PMID:36972614 are false positives, complete sentences followed by that paper's superscript inline citation numbers.

| Finding | Action |
|---|---|
| 1 — truncated PMID:11952275 snippet | Fixed. Full sentence quoted, plus the sweep above |
| 2 — no `animal_models:` section | Added, 2 models with `modeled_mechanisms` |
| 3 — mitochondrial branch absent | Added as a node, wired between ROS and macrophage death |
| 4 — Gremlin-1 and syndecan-4 unmodeled | Added as `Biochemical` entries with readouts |
| 5 — melanoptysis | Not added, reasoning below |
| 6 — 35 items graded `OTHER` | No change; it is a deliberate convention |

**Suggestions 2 and 3 are the same gap, so they are fixed together.** A new `Macrophage Mitochondrial Dysfunction and Defective Mitophagy` node (`GO:0000423`, `DECREASED`) sits between reactive oxygen species and both macrophage death and inflammatory recruitment. Two models hang off it:

- **PMID:41894156** — coal-silica mixed dust in mice with kinetin mitophagy rescue and PINK1 knockdown. `RECAPITULATES` the mitochondrial node with two readouts: `DECREASED` membrane potential and oxidative phosphorylation, and `RESTORED` mitophagy on the rescue arm. The PINK1 knockdown abolishing the rescue is what makes this a mechanism rather than an association, which is why the node is curated at all. Also `PARTIALLY_RECAPITULATES` PMF at `LOW` fidelity — murine fibrosis after months of instilled dust is not the conglomerate mass lesion that defines human PMF after decades of inhalation.
- **PMID:35057792** — nine-month coal dust particle instillation with a vitamin D arm. Its dust is characterized as under 5 µm with under 10% silica, so unlike the model above it isolates a coal-predominant exposure, which is useful given how much of this entry's mechanism evidence comes from silica-containing or nanoparticle preparations. `RECAPITULATES` the tissue injury node; `PARTIALLY_RECAPITULATES` fibroblast activation at `LOW` fidelity, because its fibrogenic readout is transcriptional and macrophage-phenotypic rather than the α-SMA-positive myofibroblast that node names.

You noted that the `HUMAN_MODEL_MISMATCH` discussion was a legitimate alternative to model links. Both are now present, and they are doing different jobs: the discussion is about the EMT branch, whose only support is a carcinoma cell line, while these links are about the mitochondrial and tissue-injury branches, where real mouse work exists with a rescue arm. Neither replaces the other.

**Suggestion 4:** syndecan-4 reads against the matrix deposition node and gremlin-1 against mesenchymal activation, matching what each protein does rather than lumping both onto the same target. Neither binds an ontology term — NCIT has no measurement concept for either, and a broader proteoglycan or BMP-antagonist term would name a different analyte. Recorded in `notes` per *no term beats a bad one*.

**Suggestion 5, not taken.** Melanoptysis is characteristic, and you are right that there is no HPO term. But a phenotype whose only content is a free-text `preferred_term` adds a row to the entry without adding anything queryable, and the mechanism that produces it — dust-laden macrophages in the airway — is already modelled. I would rather see an NTR filed than a placeholder row committed. Happy to add it if you disagree.

**Suggestion 6, no change.** Confirming it is conscious. Every `OTHER` item is a synthesis sentence from a narrative review, and grading those `HUMAN_CLINICAL` would claim the review made an observation it did not. The convention also came out of this PR's earlier pass, where nine quotes from PMID:36581638 turned out to be introduction sentences citing other people's work and were regraded and then largely replaced with that paper's own results.

**One number moved the wrong way and I want to flag it rather than let it look like a regression.** Compliance went 85.4% → 84.0%. The two new animal models introduce `modeled_mechanisms` links whose optional recommended slots are not all populated, so the denominator grew faster than the numerator. Nothing that was scored before is scored worse now.

Validation on the new head: `just validate-disorders` passes with 114/114 snippets; entity-ref, causal-target, duplicate-key, enum-value and qualifier-term gates all OK; whole-KB snippet-length, title-snippet, folded-hyphen, snippet-grading and environmental-evidence gates report no new violations; the model-link, entity-ref and animal-model test selection passes (16,242 tests); graph is 42 nodes and 47 edges with no orphan targets.

The known gaps in the PR description still stand — the PMF prevalence figure quoted from another paper's introduction, and the uncited GBD 2019 burden analysis against a description that claims worldwide scope.

---

Issue #10679, opened 2026-09-03
Curation-queue preflight is lexical, so it cannot see concepts already covered under a parent term


mellybelly:
The duplicate check that runs before a disease is claimed compares MONDO IDs, entry labels and synonyms. It has no way to notice that a concept is already curated under a *different* MONDO term — usually the target's parent. Issue #10069 was filed because of this: pituitary gland adenoma (MONDO:0006373) had been covered for ten days by `Pituitary_Tumor.yaml`, which anchors the parent term MONDO:0017611, and neither a filename search for `Pituitary_Adenoma.yaml` nor an ID search for MONDO:0006373 could match it.

The fix is to walk the MONDO `subClassOf` closure when checking coverage. `scripts/audit_curation_queue_coverage.py` does that, and running it over the queue turned up two things worth acting on separately.

## What the audit found in the stub queue

Against 1,338 open stubs and 3,857 curated MONDO IDs:

| Finding | Count |
|---|---|
| Same MONDO term claimed by more than one open stub | 0 |
| Term already curated exactly — the stub can be closed | 3 |
| Term has a curated *ancestor* | 641 |

The three that can be closed now:

- MONDO:0014380 colobomatous microphthalmia-rhizomelic dysplasia syndrome → `Colobomatous_Microphthalmia-Rhizomelic_Dysplasia_Syndrome.yaml`
- MONDO:0014870 NEK9-related lethal skeletal dysplasia → `Lethal_Congenital_Contracture_Syndrome.yaml`
- MONDO:0851095 KINSSHIP syndrome → `KINSSHIP_Syndrome.yaml`

**The 641 are not 641 duplicates.** Almost none of them are the pituitary case, where an entry genuinely covered the concept and the checker could not see it. They are lump/split decisions that the queue is presenting as "curate a new entry", and they need a ruling before anyone spends effort:

- MONDO:0012481 mevalonic aciduria and MONDO:0009849 hyperimmunoglobulinemia D with periodic fever both sit under the curated `Mevalonate_Kinase_Deficiency` — they are the two ends of one spectrum
- MONDO:0013679 sclerosteosis 2 under `Sclerosteosis`
- MONDO:0016855 Mowat-Wilson syndrome due to monosomy 2q22 under `Mowat-Wilson_syndrome`
- MONDO:0012287 Stickler syndrome type I nonsyndromic ocular under `Stickler_Syndrome_Type_1`

[...]

Issue #6951, opened 2026-07-25
Policy needed: grouping/spectrum `frequency:` bands leaking into subtype entries (OI type I, Stickler type 1, DEB)


cmungall:
## Problem

Reviewing ten ECM disorder entries for claim–evidence match surfaced a recurring
structural problem: **a subtype entry inherits a `frequency:` band from a
parent/grouping-level source, or asserts a severe-subtype rate at disease level.**
The band then reads as a claim about the curated entity when the evidence only
supports a claim about its superclass or one of its subtypes.

Three instances, all needing curator judgment rather than a mechanical fix.

### 1. `Osteogenesis_Imperfecta_Type_I` — ORPHA:666 is the whole OI spectrum

Five phenotypes take the ORPHA:666 band as the type I `frequency:` and mark it
`SUPPORT`: `Osteoporosis` (FREQUENT), `Bone Pain` (FREQUENT),
`Progressive Hearing Impairment` (FREQUENT), `Wormian Bones` (OCCASIONAL),
`Bruising Susceptibility` (OCCASIONAL).

`Hyperhidrosis` in the same file explicitly declines to do this:

> "No type I-specific frequency is asserted here because the Orphanet band reflects
> the whole OI spectrum rather than type I alone."

That reasoning is right, and type I is the *mildest* OI form — importing
spectrum-level bands likely overstates severity-linked phenotypes.

### 2. `Stickler_Syndrome_Type_1` — ORPHA:828 is the Stickler grouping

Roughly 18 phenotypes take ORPHA:828 (all Stickler types) verbatim as the STL1
frequency: `Arachnodactyly`, `Kyphosis`, `Pectus Carinatum`, `Spondylolisthesis`,
`Hypotonia`, `Chronic Otitis Media`, and others. The file already demonstrates the
right pattern twice — `Cataract` is downgraded and `Membranous Vitreous` upgraded,
both with stated reasoning — and the entry itself notes STL1 skeletal features are
milder than STL2/STL3.

### 3. `Dystrophic_Epidermolysis_Bullosa` — severe-RDEB rates at disease level

Seven phenotypes carry a disease-level `frequency:` restricted only by free text:

| Phenotype | frequency | note |
|---|---|---|
| Pseudosyndactyly | VERY_FREQUENT | "Characteristic of severe RDEB" |
| Esophageal Stricture | VERY_FREQUENT | "Primarily in RDEB" |
[...]

PR #10698, opened 2026-09-03
stubs: retire McCune-Albright syndrome as a SUBTYPE of Fibrous Dysplasia

### LINKED ISSUE #10689: Curate McCune-Albright syndrome (MONDO:0018919)
Curate a dismech entry for **McCune-Albright syndrome** ([MONDO:0018919](https://monarchinitiative.org/MONDO:0018919)) — a postzygotic mosaic *GNAS* disorder causing constitutive Gsα/cAMP signalling, with fibrous dysplasia, café-au-lait macules, and gonadotropin-independent precocious puberty alongside other endocrine hyperfunction.

Inheritance modelling needs care: MAS is **not inherited** — it arises from a postzygotic somatic *GNAS* activating variant (typically p.Arg201His/Cys), and the mosaic distribution of the mutant clone, not allelic differences, explains the variable tissue involvement. Note the relationship to isolated fibrous dysplasia and to *GNAS*-related endocrine disease without conflating them; `GNAS-related_pituitary_adenoma_3.yaml` already exists and should not be duplicated.

Stub: `stubs/McCune-Albright_Syndrome.yaml`

yaseminbridges:
McCune-Albright syndrome (MONDO:0018919) is already curated. `kb/disorders/Fibrous_Dysplasia.yaml` covers the combined FD/MAS spectrum rather than skeletal-limited disease, so a separate entry would duplicate it. This PR records that decision on the stub and retires it.

The existing entry is already MAS-scoped:

- synonyms include `Fibrous dysplasia/McCune-Albright syndrome` and `FD/MAS`
- `category: Somatic mosaic`
- the causal lesion is cited to FD/MAS spectrum reviews (PMID:31037426, PMID:27492469)
- the pathophysiology chain branches into the extraskeletal arm explicitly:

```
Somatic GNAS activation in mosaic skeletal progenitors
  |-> Impaired osteogenic differentiation and marrow replacement   (skeletal arm)
  `-> Extraskeletal mosaic GNAS activation
        |-> Large cafe-au-lait macules with irregular margins
        |-> Precocious puberty in females
        |-> Hyperthyroidism
        |-> Elevated circulating growth hormone concentration
        `-> Ovarian cyst
```

That second branch is the MAS pathograph. Per `docs/curation-stubs.md`, a `SUBTYPE` decision retires the stub without a `kb/disorders/` entry: MAS becomes a `has_subtypes` entry on Fibrous Dysplasia, bound to MONDO:0018919 so the concept stays covered. The reasoning is committed to the stub's `notes` before deletion so it survives in git history.

**The frequency leak is now fixed here.** Those MAS phenotypes carried ORPHA:249 `VERY_RARE` bands. ORPHA:249 is the *fibrous dysplasia of bone* record, so they were frequencies within FD, not within MAS, where cafe-au-lait macules and precocious puberty are core features — the scope-boundary leak in #6951, with a QC check proposed in #8769. The five phenotypes are now assigned to the MAS subtype with the bands dropped, and the Orphanet evidence retained as evidence that the feature occurs.

**One thing this PR does not fix.** `stubs/Polyostotic_Fibrous_Dysplasia.yaml` (MONDO:0008274) is the same situation and still `UNDECIDED`. Its MONDO parent is MONDO:0000845, the term this entry curates, and `just check-stubs` already flags it as a possible duplicate.

Separately: `check-stubs` did not flag the McCune-Albright stub, because that match is lexical and the entry is named "Fibrous Dysplasia". That is the blind spot described in #10679.

Closes #10689


yaseminbridges:
Took option 1 — the `has_subtypes` entry, bound to MONDO:0018919.

You were right about the actual failure mode, and it is worth stating plainly: the commit message claimed to preserve the reasoning, but with no coverage the next `just seed-stubs` would have re-created the stub as UNDECIDED and the reasoning would have survived only in the git history of a deleted path. The PR defeated its own purpose.

Verified directly rather than by inspection:

```
build_coverage_index().covered_by("MONDO:0018919")
  -> disorders/Fibrous_Dysplasia.yaml
```

So the `ids` path in `build_coverage_index` now catches it, and `seed_stubs` will skip.

## The frequency leak went with it

Since writing the block was the natural place, I did not leave this to #6951. The five extraskeletal phenotypes — cafe-au-lait macules, precocious puberty in females, hyperthyroidism, elevated growth hormone, ovarian cyst — are now `subtype: McCune-Albright syndrome` with their `VERY_RARE` bands removed.

I kept the ORPHA:249 evidence rather than deleting it, and rewrote each explanation to say what it actually supports: that the feature occurs, not how often it occurs in this subtype. Asserting no frequency seemed better than asserting one measured in the wrong population, and the subtype description records why, so the next curator does not "helpfully" restore the bands.

## On your two blue notes

**The precedent inversion.** Your reading is right — this is the one SUBTYPE deletion where the parent did not hold the coverage, which is what made it the wrong one to delete. With the `has_subtypes` entry written, it now matches the other 11 rather than inverting them.

**The relation direction.** Agreed, and this is why I did not take option 3. MAS is FD *plus* extraskeletal disease, so it is clinically broader, and `skos:narrowMatch` would assert the opposite. `has_subtypes` lets the relation be stated as what it is — a scope split within one mosaic spectrum — without forcing it through a mapping predicate that does not fit. The underlying tension you identify is real: the entry's anchors (`disease_term` MONDO:0000845, exactMatch ORPHA:249) are skeletal-FD-scoped while the prose claims spectrum scope. The subtype block does not resolve that, it just stops the frequency bands leaking across it.

## Not done

Monostotic and polyostotic subtypes. The PR body mentioned them and I have corrected that wording — `Polyostotic_Fibrous_Dysplasia.yaml` (MONDO:0008274) is a separate UNDECIDED stub, and deciding it inside this PR would be exactly the scope creep you noted this PR had so far avoided.

Validated: `just validate` clean, 75/75 snippets; `check-stubs` (1337 stubs, no errors), `check-entity-refs`, `check-snippet-length`, `check-title-snippets`, `check-empty-snippets` all pass.

yaseminbridges:
Cache rows committed — you were right, and the reasoning about *why* matters more than the fix.

```
cache/mondo/terms.csv                            +MONDO:0018919,McCune-Albright syndrome,...
cache/enums/diseaseorsubtypeterm_6d5891ec946f.csv +MONDO:0018919
```

Generated with `just validate-terms kb/disorders/Fibrous_Dysplasia.yaml` rather than hand-written, and `just check-term-cache-integrity` passes (18 term cache files, 24 enum cache files match the structural contract).

The failure mode you describe is the one that would actually have bitten: CI would have gone green over OLS, so nothing would have surfaced it, and the rows would have appeared as unexplained churn in the next person's working tree — precisely the situation CLAUDE.md tells them to stop and surface rather than commit. I had run `just validate` and seen it pass without checking whether it had *written* anything, which is the gap.

One unrelated thing I am deliberately leaving out: `references_cache/PMID_25033066.md` is modified in my working tree, upgraded from `abstract_only` to `full_text_html` by a validate run. It touches neither this PR nor the Frasier one, so I have not committed it here rather than smuggle unrelated cache churn into a stub-retirement PR. Flagging it in case it is wanted as its own change.

PR #10700, opened 2026-09-03
Quarantine known defects in NDEx production releases

cmungall:
The first production audit found the 106 export defects already recorded on #966, which currently blocks all 2,367 clean networks from release. This change explicitly quarantines the 105 affected networks while preserving a fatal gate for every new or changed defect.

The allowlist records the exact affected node names or missing-metadata finding, not just per-network counts. A matching network is written to the manifest as `SKIPPED_EXPORT_DEFECT` and is never uploaded. If its defect is fixed it becomes publishable automatically; if its defect changes, the audit fails. Existing UUIDs for a temporarily quarantined network remain in the compact registry, so quarantine neither updates nor falsely retires an existing NDEx network.

Current full-corpus result on the rebased branch:

- 2,535 disorder records
- 2,367 clean networks ready to publish
- 105 quarantined networks covering 106 reviewed findings
- 63 records skipped because they have no pathograph edges
- 0 unapproved export defects

The publisher also independently rejects a manifest that marks a defective network as publishable, even if its defect appears in the allowlist.

Validation:

- `uv run pytest tests/test_ndex_publish.py tests/test_cx2_export.py -q` — 24 passed
- changed-file pre-commit hooks — passed
- full production export audit at current `main` — passed with zero unapproved defects

Follow-up to #966 and #10558.

### DIFF
--- a/.codespell-ignore-words.txt
+++ b/.codespell-ignore-words.txt
@@ -20,6 +20,7 @@ disjointness
 displa
 diviation
 ETO
+Fo
 Fulfilment
 Hart
 hsi
--- a/.github/workflows/publish-ndex.yml
+++ b/.github/workflows/publish-ndex.yml
@@ -111,7 +111,8 @@ jobs:
             carried="$(jq '[.networks[] | select(.ndex_uuid != null)] | length' output/ndex/manifest.json)"
             retired="$(jq '.summary.retired_network_count' output/ndex/manifest.json)"
             quarantined="$(jq '.summary.quarantined_network_count' output/ndex/manifest.json)"
-            echo "Carried forward ${carried} NDEx UUIDs; quarantined ${quarantined} known-defect networks; reported ${retired} retired networks." >> "$GITHUB_STEP_SUMMARY"
+            stale="$(jq '.summary.unmatched_allowed_export_defect_count' output/ndex/manifest.json)"
+            echo "Carried forward ${carried} NDEx UUIDs; quarantined ${quarantined} known-defect networks; found ${stale} stale quarantine entries; reported ${retired} retired networks." >> "$GITHUB_STEP_SUMMARY"
           fi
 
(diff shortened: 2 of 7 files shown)


cmungall:
All three requested changes are addressed in `84f87be2f1`:

- `BLOCKED_EXPORT_DEFECT` records with prior UUIDs are preserved by `write_uuid_registry`, so `--allow-export-defects` cannot finish uploads and then fail while writing continuity state. The docs now state that this flag suppresses the audit failure but still does not upload defective networks; a regression test covers the interrupted-release risk.
- The summary now includes `blocked_network_count`, so exported, skipped, and blocked records account for the entire corpus.
- Stale entries are reported in `unmatched_allowed_export_defects` and `unmatched_allowed_export_defect_count`, and the workflow summary surfaces the count. Tests cover both a changed same-count defect and a stale entry.

I also stored defects structurally on each network record, retained the legacy-manifest fallback in the publisher, added deterministic labels for nameless defective nodes, documented `BLOCKED_EXPORT_DEFECT` and the allowlist string contract, and fixed the prefixed biochemical readout in `Mitochondrial_Complex_V_ATP_Synthase_Deficiency_Nuclear_Type_3`.

I did not apply the suggested one-line change to the CMT4K imaging target. `collect_graph_nodes()` does not include `imaging_findings`, and `scripts/check_causal_targets.py --report` confirms that the bare target still dangles after removing the prefix. It therefore remains explicitly quarantined pending a real graph-model or curation fix.

Refreshed full-corpus audit:

- 2,535 records
- 2,368 publishable networks
- 104 quarantined networks / 105 reviewed findings
- 63 edge-empty records
- 0 blocked networks
- 0 unapproved findings
- 0 stale allowlist entries

Validation: 26 focused tests passed, the changed disorder passed schema/term/reference validation and both target/reference checks, and all changed-file hooks passed.

PR #10701, opened 2026-09-03
Add Patent Ductus Arteriosus 3 (PRDM6-related) disorder entry

tannerzhang:
## Summary

This PR adds a comprehensive knowledge base entry for Patent Ductus Arteriosus 3 (PDA3), a Mendelian disorder caused by loss-of-function variants in PRDM6. The entry documents the genetic basis, pathophysiology, clinical presentation, and three competing mechanistic hypotheses explaining how PRDM6 loss leads to ductal patency.

## Key Changes

- **New disorder entry** (`kb/disorders/Patent_Ductus_Arteriosus_3.yaml`): 1,830-line YAML file covering:
  - Disease classification (autosomal dominant, incomplete penetrance, variable expressivity)
  - Inheritance evidence from linkage analysis and whole-exome sequencing (PMID:27181681)
  - Ultra-rare prevalence (case-count based, not population rate)
  - Three mechanistic hypotheses:
    - **Premature differentiation model**: PRDM6 loss releases ductal smooth muscle from undifferentiated proliferative state
    - **SMC identity model**: PRDM6 required to build contractile gene program (contradicts first model)
    - **Neural crest specification model**: PRDM6 acts upstream on cardiac neural crest cell specification
  - Detailed pathophysiology chain from PRDM6 loss-of-function variants → deregulated transcriptional program → cellular/hemodynamic consequences
  - Three-phase disease progression (neonatal presentation → hemodynamically significant shunt → pulmonary vascular disease)
  - Evidence from human genetics (PMID:27181681, PMID:38071433) and three complementary mouse models (PMID:36749647, PMID:35108221)

- **Deep research report** (`research/Patent_Ductus_Arteriosus_3-deep-research-claude_code.md`): 1,003-line research synthesis with 22 citations covering disease characteristics, etiology, phenotypes, genetic/molecular basis, and mechanistic pathways

- **Reference cache entries**: 15 new references added (PMIDs and DOIs) documenting:
  - Original PRDM6 discovery (PMID:27181681)
  - Recent PDA3 + coarctation cases (PMID:38071433)
  - Mouse model studies on SMC identity (PMID:36749647) and neural crest specification (PMID:35108221)
  - General PDA pathophysiology and clinical management literature

- **Ontology cache updates**: Added MONDO:0024266 (patent ductus arteriosus 3), ICD10CM:Q25.0, GO:0097070, and treatment action terms to dynamic enums

- **Curation history**: Two session records documenting creation and review workflow

- **Mechanistic divergence**: The entry explicitly documents that two competing cellular models (premature differentiation vs. SMC identity) both explain the same molecular observation (altered transcriptional program) but propose opposite roles for PRDM6 (repressor vs. promoter of contractile genes). This is recorded as an open discussion rather than resolved.

- **Evidence grading**: Molecular mechanisms are graded ESTABLISHED (human genetics + in vitro characterization); cellular mechanisms are PROVISIONAL (mouse models); neural crest route is EMERGING (single mouse study).

- **Incomplete penetrance**: Obligate carriers in the index kindred are reported unaffected, documented as enum values rather than cited claims per the primary article's abstract-only availability in the reference cache.

- **Coarctation association**: The entry notes that a subset of PRDM6 carriers also present with coarctation of the aorta, expanding the phenotypic spectrum beyond isolated PDA.

## Validation Status

- Schema validation: passes
- Term validation: 14/15 references verified; 1 off-topic reference (DOI:10.1161/01.cir.0000145159.16637.5d) flagged for review
- Ontology terms: 16/20 verified; 1 obsolete term (GO:0034770); 3 mislabeled terms noted for review

https://claude.ai/code/session_01Aqy2M7tNZDPd2j6UYL5Uhb


tannerzhang:
Fixed the corrupted description and pushed as c6d373ea3. The reviewer is right and this was a real defect: a copy of a `biological_processes` block landed inside the description's folded scalar during an earlier edit, with the intended prose fused onto the end of a `label:` line. Because it all sat inside a `>-` scalar it was one valid string, so every validator passed while the disease page would have rendered a YAML fragment as the entry's description.

I also swept every free-text scalar in the file for the same failure (parsed the YAML and checked each string value for leaked keys like `preferred_term:`, `snippet:`, `modifier:`). No others. And I rendered the page to confirm the description now comes out as prose.

The three suggestions I acted on:

- **`HP:0031670 Continuous heart murmur` reverted to `HP:0030148 Heart murmur`.** The continuous machinery murmur is the classic sign, but the word "continuous" appears in none of this entry's cached references, so the binding was narrower than its snippet could reach. The phenotype name and description now agree with the term, and the notes record why the narrower term was not used.
- **Dropped `therapeutic_modality: SMALL_MOLECULE` from the supportive-care entry.** It bundles fluid restriction with diuretics, and neither `SMALL_MOLECULE` nor `BEHAVIORAL` is true of both, so it now carries no modality tag and says why.
- **`Impaired Ductal Wall Remodeling`** now carries `CL:0000359` and `GO:0048771` (tissue remodeling), like its siblings.

On the full-text suggestion: I re-ran `just fetch-reference PMID:27181681` against the PMC4908195 record and it still returns `abstract_only`. So the in vitro histone-mark reversal, the MYH11 suppression failure, and the no-dominant-negative finding stay uncurated, and `review_notes` now records the retry rather than leaving it as an open lead.

I left the optional research-surfaced additions out (infective endarteritis, Eisenmenger physiology, failure to thrive, paracetamol). None is specific to the PRDM6 mechanism, and the progression block already carries the pulmonary-vascular-disease phase.

**One correction to the PR description, which I did not write.** Its "Validation Status" section reports "14/15 references verified", "16/20 ontology terms verified", an obsolete `GO:0034770`, and an off-topic DOI. Those are the *deep-research report's* validation numbers, not the entry's, and reading them as the entry's would be misleading. The entry itself has all 77 snippets verified against the reference cache and all ontology terms passing. `GO:0034770` is obsolete and was never bound anywhere in the YAML; the off-topic DOI is only present in the reference cache and is never cited. Both were flagged by the report's own validator and deliberately not used, which is the check working as intended.

Current state: `just validate-disorders` passes, 77/77 snippets verified, and the entity-ref, causal-target, snippet-length, title-snippet, snippet-grading, folded-hyphen, enum-value, duplicate-key and qualifier-term gates are all clean.

---

tannerzhang:
Took suggestions 1-3 and pushed as 62a612fd9. This resets the approval, which is worth it for getting the prose in the right fields.

Moved the curation rationale out of `description` and into `notes` in four places: why the murmur phenotype binds the general HP term, why the supportive-care treatment carries no modality, that the entry carries both readings of PRDM6's direction of effect (now in the entry `notes`, pointing at the hypothesis groups and the discussion), and the inheritance sentence about enum values, which `review_notes` already covered so it is simply dropped. Description fields now carry clinical prose only.

On the history record: good catch that the 10:07 record documents the `HP:0031670` change this branch later reversed. Added a third `REVIEW` record that states the reversal and why, and leaves the earlier line alone since history is append-only and that record accurately describes what that session did.

Re-ran the gates plus the full-file scalar scan (984 strings, zero suspect) and `just validate-disorders`. All clean.

**Suggestion 4, the stale validation section in the PR description — leaving that to @tannerzhang.** It reports the deep-research report's numbers rather than the entry's, including an obsolete `GO:0034770` that appears nowhere in the YAML. I corrected it in the comment above rather than editing the body, since the UI generated that text and it is not mine to rewrite. Happy to edit it if you would rather I did.

---

PR #7872, opened 2026-08-02
Add Distal Myopathy 6, Adult-Onset, Autosomal Dominant (ACTN2, MONDO:0032853)

nlharris:
## Summary

New dismech entry for **Distal Myopathy 6, Adult-Onset, Autosomal Dominant** ([MONDO:0032853](https://monarchinitiative.org/MONDO:0032853); MPD6 / OMIM:618655) — the **ACTN2**-related autosomal dominant distal myopathy ("actininopathy"). Closes #7854 (top of the priority dashboard, `CURATE_ROOT`).

α-Actinin-2 is the sarcomeric Z-disk protein that cross-links actin and titin; heterozygous ACTN2 variants disrupt Z-disk architecture and drive a slowly progressive distal myopathy that begins in the anterior lower leg (ankle dorsiflexion), often with facial weakness, later spreading proximally.

## What was curated

- **Pathophysiology** (4 atomic nodes, molecular → molecular → cellular → tissue): ACTN2 pathogenic-variant classes (missense + last-exon protein-extending frameshift) → α-actinin-2 protein aggregation / Z-disk & myofibrillar disorganization → distal skeletal-muscle degeneration.
- **Phenotypes:** distal lower-limb weakness, foot-dorsiflexor weakness (foot drop), facial weakness, later proximal spread, anterolateral lower-leg fatty replacement on muscle MRI (HP:0012548), rimmed vacuoles, centrally nucleated fibers, multi-minicores.
- **Genetics:** ACTN2 (hgnc:164), autosomal dominant inheritance.
- **Treatment:** supportive/rehabilitative care (no disease-modifying therapy exists).

## Evidence (verified primary literature)

- PMID:30900782 — Savarese 2019, original actininopathy cohort (AD, distal→proximal, missense)
- PMID:34386585 — Savarese 2021, last-exon out-of-frame variants with facial weakness (MRI pattern, biopsy)
- PMID:34170073 — Chen 2021, frameshift multi-minicore family (increased α-actinin-2, abnormal Z-line)
- PMID:39095936 — Ranta-Aho 2024, protein-aggregation mechanism (cell model, IN_VITRO)
- PMID:36116040 — Ranta-Aho 2022, ACTN2 mutation update (gene identity/expression)

## Validation

- `just validate`: schema ✓, term validation ✓, **19/19 evidence snippets verified** against cached abstracts
- `just compliance`: 88.2%
[...]

PR #10702, opened 2026-09-03
Augment MPD6 with epidemiology, progression, burden, and mechanism

tannerzhang:
## Summary

Augments the existing Distal Myopathy 6 (MPD6, ACTN2) entry, which was created in #7872. Compliance goes from 81.7% to 98.5% (weighted 82.4% → 99.0%) and evidence snippets from 30 to 95, all verified against the reference cache.

The curation problem specific to this disease is that ACTN2 is far better known as a **cardiomyopathy** gene. Most of the mechanistically useful literature is therefore about a different phenotype, and both deep-research reports pulled in wrong-gene and wrong-disease citations. Handling that carefully is most of what this PR does.

## What was added

**Clinical and epidemiological.** Prevalence (recorded as `CASES_IN_LITERATURE` with no rate asserted — the reported cases concentrate in two founder populations and index ascertainment, not population frequency); a three-phase progression; `clinical_burden: MODERATE`; a `diagnosis:` section covering panel sequencing, muscle MRI and EMG/nerve conduction studies; and baseline cardiac evaluation.

**Findings.** Histopathology (Z-disk undulation on electron microscopy, plus the largest cohort's non-specific biopsy result, which qualifies how much diagnostic weight the biopsy features elsewhere in the entry can carry); imaging findings (the 37-patient MRI distribution and a machine-learning-derived discriminating signature, including which muscles are *preserved*); serum creatine kinase as a biochemical marker, with the point that mild elevation makes CK a poor discriminator here; two phenotypes.

**Mechanism.** A new upstream pathophysiology node for conformational destabilization of alpha-actinin-2, evidence on four of five causal edges, and a mechanistic hypothesis for why a uniform Z-disk defect produces weakness that starts in the anterior lower leg.

**Models.** Two animal models and one cell model, each with `modeled_mechanisms` links and readouts.

**Genetics and differentials.** Four variant records; `penetrance: INCOMPLETE` and `expressivity: VARIABLE` on inheritance; two new differential diagnoses (biallelic recessive ACTN2 myopathy, and the allelic congenital core myopathy MsCD) plus distinguishing features on the existing one; three treatments added and evidence on all six.

## Judgment calls worth reviewing

**Cross-phenotype evidence is borrowed deliberately and labelled.** PMID:42481460 and PMID:27287556 profile ACTN2 hypertrophic-cardiomyopathy variant panels. PMID:30701273's zebrafish and AAV-mouse models express p.Leu727Arg, an allele of the allelic *congenital* myopathy, not of MPD6. These are the only handle on the missense-class mechanism the entry previously recorded as undetermined, so they are used — but each carries `directness: INDIRECT` with the allele mismatch written out, the animal models are `PARTIALLY_RECAPITULATES` at `fidelity: LOW`, and a `HUMAN_MODEL_MISMATCH` discussion states that no existing model addresses the feature that defines MPD6: adult onset selecting the anterior compartment.

**Two deep-research citations were declined**, recorded in `review_notes`. The falcon report attributes the founding Savarese 2019 MPD6 abstract to two neighbouring abstracts printed on the same supplement page (UNC-45B and PAX7). The OpenScientist report cites an **alpha-actinin-3** paper for an ACTN2 mechanism — a different gene.

**`datasets:` is deliberately empty.** `just discover-datasets` returns GNE-myopathy mouse series, matched on the words "myopathy distal adult onset", and ACTN2 cardiac series matched on the gene. Both resolve perfectly and neither is about this disease. The triage is recorded rather than left implicit.

## Deep research

Falcon was requested and its endpoint returned 403, so the run was re-issued with `--fallback` rather than substituting a provider by hand. OpenScientist produced the report, and its frontmatter records `fell_back: true` with `requested_provider: falcon` and the full attempt list. The pre-existing falcon report from the August session is retained unchanged.

Its reference validation came back 17/17 resolved with a 0.0 confabulation rate. Its **term** validation came back `needs_review: true` with three mislabelled terms, including `MONDO:0014138` offered as this disease's MONDO identifier when the term is *nemaline myopathy 8*. None of the three was bound; every ontology term in this PR was resolved with OAK first.

## Validation

`just validate-disorders` (the batched schema, term and reference gate CI runs) passes, 95/95 snippets verified. Offline gates all green: `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms` plus the online variant (needed because the `NCIT:C86054` device qualifier is a qualifier-only CURIE the offline gate cannot see), `check-enum-values`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-environmental-evidence`, `check-snippet-grading`.

Three history records under `history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/` cover the augmentation, a pre-PR self-review, and this review response.

---

*This description replaces an auto-generated one that contained errors: it labelled `UBERON:0001385` "soleus muscle" (the term is tibialis anterior; soleus is `UBERON:0001389`), miscounted the reference-cache entries and progression phases, and omitted most of the sections above.*

https://claude.ai/code/session_01WPvTTgYEKX1xKes6bWTJw5

### DIFF
--- a/history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/2026-09-03T092637Z-claude-code-cf7d7c.yaml
+++ b/history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/2026-09-03T092637Z-claude-code-cf7d7c.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10702
   urls: []
 events:
 - type: EDIT
--- a/history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/2026-09-03T094646Z-claude-code-a74ce2.yaml
+++ b/history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/2026-09-03T094646Z-claude-code-a74ce2.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10702
   urls: []
 events:
 - type: EDIT
--- a/history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/2026-09-03T145043Z-claude-code-8324c0.yaml
+++ b/history/disorders/Distal_Myopathy_6_Adult-Onset_Autosomal_Dominant/2026-09-03T145043Z-claude-code-8324c0.yaml
@@ -13,7 +13,8 @@ session:
     agent_tool: claude-code
 links:
   issues: []
-  prs: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10702
   urls: []
 events:
 - type: EDIT
@@ -64,6 +65,13 @@ events:
     rather than by this session; the same body also miscounted the reference
     cache entries and the progression phases and omitted several added sections,
     so it was rewritten to match the diff.
+
+    Follow-up from the approving re-review: all three of this session's history
+    records had an empty `links.prs`, while the 2026-08-02 record in this same
+    directory populates it. Structured links are what make provenance queryable,
+    so the PR URL was added to all three rather than left named in prose here.
+    The reviewer marked this non-blocking and not worth its own commit; it is
+    taken anyway because the re-review that a push costs is automated.
 
     Validated: `just validate`, `just validate-disorders` (batched), 95/95
     snippets, check-entity-refs, check-causal-targets, check-duplicate-keys,


tannerzhang:
Both review points were right and are fixed in 65dec38bf.

**The requested change.** Added `penetrance: INCOMPLETE` and `expressivity: VARIABLE` to `inheritance:`, with the PMID:39812845 genotype-dependence sentence as its own evidence item. The argument for structuring them is the convincing part: the entry already asserted both facts in prose, quoting that same sentence in the presymptomatic progression phase, so leaving them unstructured meant a query for autosomal dominant diseases with incomplete penetrance would miss this one. `penetrance_percentage` stays off, since no cohort reports a figure.

**Suggestion 1, taken.** Removed the `Asymmetric muscle weakness` phenotype. You're right that it re-bound `HP:0009053`, and it was worse than redundant — the existing distal-weakness node already carried the *identical* PMID:30900782 asymmetry snippet as evidence and already described the weakness as asymmetric. Adding it was a mistake on my part rather than a judgment call, so there's nothing to weigh.

While there I fixed something adjacent that your note surfaced. The surviving node justified leaving asymmetry unbound with "HP:0031500 is Abdominal mass", which is not a plausible candidate for a laterality qualifier and reads like an unverified check inherited from the original entry. I replaced it with the terms actually checked, verified with OAK rather than taken from your review: the `HP:0012831` laterality modifiers are Bilateral, Unilateral, Right, Left and Alternating laterality, none of which expresses side-to-side asymmetry of severity, and the nearest asymmetry term, `HP:0007156`, is asymmetric limb muscle *stiffness*. Same conclusion you reached, now with the right reasoning recorded in the file.

**Suggestion 2** was about the PR description, not the entry, and it was worse than the one label you caught. That description was generated by the Claude Code UI rather than by this session; besides calling `UBERON:0001385` "soleus muscle", it listed eight reference-cache entries under a heading saying seven, described four progression phases and listed three, and omitted the diagnosis section, models, variants, differential diagnoses, treatments, histopathology, imaging findings and discussions entirely. I've rewritten it to match the diff and noted the correction at the bottom.

Compliance is 98.5% (weighted 99.0%), down a fraction from 98.6% purely because the removed duplicate phenotype had been contributing a fully-scored node. 95/95 snippets verified; `just validate-disorders` and every offline gate pass.

---

tannerzhang:
One correction to my previous comment, now pushed in cd5f51970.

Your full audit notes that the in-prose claim about `HP:0031500` being "Abdominal mass" is correct. I checked with OAK and you're right — it is. So my calling that note "an unverified check" was wrong and unfair to whoever wrote it. The note was accurate; it was just an irrelevant term to cite, since nothing would lead a curator to consider an abdominal-mass term for a laterality qualifier. I've fixed that characterization in this session's history record, which is the durable artifact where it would otherwise have stuck.

I also took the substance of your check rather than only the conclusion: the node's justification now names `HP:0007156` (asymmetric limb muscle *stiffness*, not weakness) alongside the `HP:0012831` laterality modifiers, and points at the term contract clause this case falls under — `preferred_term` may be more specific than the best available term.

On finding 2, I went with folding rather than keeping both nodes, so there's no shared-binding note to write. `HP:0009053` is bound once again.

95/95 snippets still verified; `just validate-disorders` and the offline gates pass on the new head.

One process note, since it affects who has to do what next: your `CHANGES_REQUESTED` review is sticky and I have not dismissed it. I tried to re-request a review through the `claude-code-review.yml` workflow and got `403 Resource not accessible by integration` — but the pushes are triggering `claude-review` runs on their own, so a fresh review appears to be arriving without anyone needing to intervene.

---

tannerzhang:
Took the nit as well, in c8e1ea81f. All three history records now populate `links.prs`.

I checked the convention claim before acting on it, and you're right: the 2026-08-02 record in this same directory carries both its issue and PR, so mine were the outlier rather than the norm. The 2026-08-11 record populates `issues` but not `prs`, so it's partway there.

You judged this not worth a commit on its own, which is a fair call against a human's time — it isn't against mine, since the re-review a push triggers here is automated. So the cost you were weighing doesn't apply, and the durable artifact gets the structured link instead of a prose mention. `validate-history` passes on all three, and `validate-history-all` passes across all 8204 records.

Heads up that this push resets the approval. That's the expected consequence and not something I'd ask you to work around.

Two things from your re-review worth acknowledging, since they're both cases where you checked the thing that actually mattered rather than the thing I claimed:

- Verifying that the surviving `Distal lower limb muscle weakness` node still carries the PMID:30900782 asymmetry snippet before agreeing the node removal was safe. That's the right check — the failure mode of deleting a "duplicate" is deleting the only copy of its evidence, and confirming it was the second copy is what makes the removal safe rather than plausible.
- Flagging that you couldn't re-resolve `HP:0012831` and `HP:0007156` because EBI timed out, and reasoning about what follows: neither is bound in the YAML, so they're prose and outside the term validator's scope either way. Saying which check didn't run and why it doesn't change the conclusion is more useful than a silent tick.

---

PR #10703, opened 2026-09-03
docs: record MONDO:0016343 "unclassified cardiomyopathy" as OUT_OF_SCOPE

tannerzhang:
A curation request came in for "obsolete unclassified cardiomyopathy". It is a real MONDO term, but it cannot become a dismech entry, so this PR records that decision instead of adding one. No `kb/` files change — it is one note under `docs/curation-notes/` plus its mkdocs nav line.

`MONDO:0016343` is deprecated. MONDO's own recorded reason for retiring it is `OMO:0001000`, whose label is literally **"out of scope"**, and the ticket behind it (monarch-initiative/mondo#2824) obsoleted it as a "too general/non-specific grouping class". It was never a disease: both its subsets (`ordo_group_of_disorders`, `disease_grouping`) mark it as the residual bucket in the Orphanet classification for cardiomyopathies fitting no other category. MONDO points to `MONDO:0004994` (cardiomyopathy) instead.

It is also structurally uncurateable here. The term has no ancestors, so it is not reachable from `MONDO:0000001` and cannot satisfy the `DiseaseTerm` enum — binding it would fail `just validate-terms`.

### Why a doc note and not a stub

Three existing mechanisms already reject an obsolete term, so this is not new policy:

- `stubs/seed.py` skips any nomination whose label starts with `obsolete` ("Nothing downstream should ever ask a curator to model one")
- `stubs/model.py` reports such a stub as `obsolete_term`, and `just tidy-stubs --apply` deletes it
- `docs/mondo-prioritizer.md` uses this exact term as its worked example of a candidate dropped as obsolete

A tombstone stub is precisely what `tidy-stubs` sweeps, so it would not survive as a record. `docs/curation-notes/` is the durable home.

### Nothing is lost

All five ESC morphofunctional categories are curated, and so are both members of the historical "unclassified" bucket — `Left_Ventricular_Noncompaction` (`MONDO:0018901`) and `Takotsubo_Cardiomyopathy` (`MONDO:0019018`). The retired bucket's clinical content is covered by name.

### Two gaps flagged, neither decided here

- **NDLVC.** The 2023 ESC guideline (PMID:37622657) retired the "unclassified" category and replaced it with non-dilated left ventricular cardiomyopathy, which has no entry and no stub. The blocker is upstream: MONDO has no NDLVC term (`l~non-dilated` returns nothing; `l~left ventricular cardiomyopathy` returns only `MONDO:7770163`, a canine term), so it cannot be stubbed or curated until one exists.
- **`MONDO:0004994`.** The root MONDO points to is uncovered, appearing only in the prose description of `kb/modules/cardiomyopathy_maladaptive_remodeling.yaml`. Whether it deserves a `Grouping` should be argued on its own merits, not carried in on an obsoleted term.

### Validation

Docs-only, so the KB gates have nothing to act on. `just check-stubs` is clean (1338 stubs, no errors) and the 59 tests covering mkdocs nav and history layout pass.

No history record: `test_committed_history_records_follow_layout` requires `target.path` to resolve to a real `kb/` file, and there is none. No `Closes #N`: no open `claim` issue exists for this concept.

### DIFF
--- a/docs/curation-notes/Unclassified_Cardiomyopathy-scope-decision-2026-09-03.md
+++ b/docs/curation-notes/Unclassified_Cardiomyopathy-scope-decision-2026-09-03.md
@@ -103,12 +103,15 @@ in MONDO — it is obsolete in the clinical nosology too, which strengthens the
 decision rather than complicating it.
 
 NDLVC itself has no `kb/` entry and no stub. It appears in `kb/` only inside two
-other entries' evidence, `Dilated_Cardiomyopathy_1FF.yaml` and
-`Hypertrophic_Cardiomyopathy_20.yaml` — abbreviated as the cohort label
-"DCM/NDLVC" in snippets, explanations and notes, and spelled out in full in the
-reference title *Prediction and Prognostic Role of Left Ventricular Systolic
-Dysfunction in Family Screening for Dilated Cardiomyopathy and Non-Dilated Left
-Ventricular Cardiomyopathy*. Never as an entity in its own right.
+other entries' evidence, never as an entity in its own right:
+
+- `Hypertrophic_Cardiomyopathy_20.yaml` — the abbreviated cohort label
+  "DCM/NDLVC" in snippets, explanations and notes, and nothing else.
+- `Dilated_Cardiomyopathy_1FF.yaml` — the same abbreviated label, plus the only
+  place the name is spelled out in full: five `reference_title` occurrences of
+  *Prediction and Prognostic Role of Left Ventricular Systolic Dysfunction in
+  Family Screening for Dilated Cardiomyopathy and Non-Dilated Left Ventricular
+  Cardiomyopathy*.
 
 The blocker is upstream, though, and it needs saying before anyone files a stub:
 **MONDO has no NDLVC term.** Searching `l~non-dilated` returns nothing, and


tannerzhang:
Thanks — all four suggestions acted on, in 092b975f0 plus a rewritten PR body. Two of them changed the finding rather than just the wording, so worth spelling out.

| Suggestion | What I did |
|---|---|
| NDLVC belongs in the gap section | Added, with two corrections (below) |
| "Deprecated MONDO terms are detached" is a generalization | Narrowed to `MONDO:0016343`, which is what was actually checked |
| Review could not verify the MONDO metadata table | No change; noting it was verified against a live MONDO build here |
| PR body is just a session link | Rewritten. No `Closes #N` — see below |

**On NDLVC.** The substance is right and it is a better gap than the one I had, so it is in. Two things came out differently when I checked:

NDLVC is not confined to two `reference_title` strings. It also appears in evidence snippets and explanations in `Dilated_Cardiomyopathy_1FF.yaml` and `Hypertrophic_Cardiomyopathy_20.yaml`, always as part of the cohort label "DCM/NDLVC", never as an entity. Same conclusion, wider footprint.

More consequential: **MONDO has no NDLVC term.** `l~non-dilated` returns nothing, and `l~left ventricular cardiomyopathy` returns only `MONDO:7770163` (*arrhythmogenic left ventricular cardiomyopathy, dog*). A stub requires a `mondo_id` matching `^MONDO:[0-9]{7}$` and a `disease_term` requires a term reachable from `MONDO:0000001`, so NDLVC can be neither stubbed nor curated today — it needs a MONDO term request first. I put that in the note explicitly, because otherwise the next reader files a stub that cannot validate. This also strengthens the PR's main argument: the bucket is obsolete in the 2023 clinical nosology as well as in MONDO.

**On the entry count.** Your 53 and my 50 are the same underlying set counted differently — 50 disorder files plus 2 groupings and 1 module. I replaced the vague "around fifty" with the reproducible figure: 50 files in `kb/disorders/` whose name contains "cardiomyopathy", 45 beyond the **five** listed that carry the word. `Left_Ventricular_Noncompaction` is in the coverage table but has no "cardiomyopathy" in its filename, which is where the off-by-one sits.

**On the metadata table.** Fair to flag what the sandbox could not reach. For the record, it was read from a live `sqlite:obo:mondo` build via OAK `term-metadata`, and `mondo#2824` was fetched directly. The `ORDO:217678` exactMatch and the `OMO:0001000` obsolescence reason both come from that read; `OMO:0001000` resolves to "out of scope" via OLS.

**No `Closes #N`.** I searched `label:claim` for this concept and there is no open claim issue for `MONDO:0016343` — the cardiomyopathy claims are all for specific CMD/HCM subtypes. Nothing to release on merge.

---

tannerzhang:
Both prose points fixed in 47e5801ec.

The NDLVC one was a real inaccuracy, not a nit — "only as part of the cohort label DCM/NDLVC" is wrong, since the name is spelled out in full in the reference title *Prediction and Prognostic Role of Left Ventricular Systolic Dysfunction in Family Screening for Dilated Cardiomyopathy and Non-Dilated Left Ventricular Cardiomyopathy*, carried by both entries. The note now says which form appears where. The conclusion is unchanged: never an entity in `kb/`.

The hedge at lines 37-39 is gone. It now just states that `MONDO:0016343` is detached from the hierarchy, which is the fact a reader of a decision note needs; what an earlier draft did or did not check is not.

On the unverifiable items: the MONDO metadata table and the two NDLVC `runoak` searches were run here against a live `sqlite:obo:mondo` build, and `OMO:0001000` was resolved through OLS. Nothing to change — recording it so the gap in review coverage is closed in the thread rather than left open.

---

tannerzhang:
You were right about the one-file overstatement, and I fixed it in the note rather than leaving it, in 88c7ce760. Counts:

```
Dilated_Cardiomyopathy_1FF.yaml       5   (all reference_title)
Hypertrophic_Cardiomyopathy_20.yaml   0
```

You called the committed prose defensible, and it was — but it listed both forms across the pair without saying which file had which, which a careful reader can take as both carrying both. For a note whose whole value is that someone can trust it without re-deriving it, that ambiguity is worth one more push. It is now a per-file list. The conclusion is unchanged, and that is now the third round it has survived: NDLVC is never an entity in `kb/`.

The commit message you flagged is wrong and stays wrong, since history is not rewritable here. 88c7ce760 states the correction and the counts, so the record self-corrects going forward.

Nothing outstanding from your other two points — the sandbox-coverage boundary needs no change, and the PR body is resolved.

---

tannerzhang:
This PR left the merge queue without merging, and nothing about it failed. Flagging it so it gets re-added rather than sitting here unnoticed.

What I can verify:

- It was enqueued at 16:29Z. It is not in the queue now — eight other `gh-readonly-queue/main/pr-*` branches exist, none for 10703.
- **It never got a merge-group run.** Zero `merge_group` workflow runs reference `pr-10703` anywhere in the listing, across all of 2026-09-03 and 2026-09-04. So it was dropped before GitHub built a merge group for it, not after failing one.
- No dequeue notice was posted here. The enqueue notice says that follow-up is best-effort, and in this case it did not arrive.

The queue had a rough day: 28 failed `merge_group` runs on 2026-09-04, several PRs failing repeatedly (`pr-9852` four times, `pr-10325` and `pr-10351` three each), with clusters of simultaneous failures at 10:20, 13:36 and 14:25 that look like whole batches being evicted at once. `pr-9852`'s group failed at 16:54Z, 25 minutes after this PR was enqueued. I can't prove this PR was evicted as collateral of that batch, but the timing fits and it is the only failure in the window.

This PR's own state is unchanged and still eligible:

| | |
|---|---|
| Approval on `88c7ce760` | still standing, not dismissed |
| `test (3.13)`, `claude-review` | success |
| Merge into current `main` (`bfc698620`) | clean, verified by trial merge |
| Human assignee | none |
| Age | over 1 day |

So the next deterministic sweep should re-enqueue it on the same predicate that picked it up the first time. I have not re-enqueued it myself: adding to the queue outside the controller bypasses its age and assignment guards, which `CLAUDE.md` rules out.

The branch is behind `main` and I have deliberately left it that way. The sweep does not require the head to contain the latest `main`, and refreshing would dismiss the standing approval and cost another review cycle for no gain.

Nothing needed from a reviewer here. If the sweep does not pick it up again, that is worth a look at the sweep rather than at this PR.

---

tannerzhang:
Correcting my previous comment: nothing here needs looking at, and the sweep is not failing to re-pick this PR. It is not eligible yet.

I said the next deterministic sweep should re-enqueue this on the same predicate that picked it up at 16:29Z. That was wrong. The 16:29Z run was a `workflow_dispatch` with `min_age_days: 0` — its own comment says "open longer than **0** days". Scheduled runs use a different threshold:

```yaml
# .github/workflows/pr-shepherd.yml
MIN_AGE_DAYS: ${{ inputs.min_age_days || '3' }}
```

and `scripts/auto_merge_ready_prs.py:312` rejects anything younger:

```python
if age < timedelta(days=min_age_days):
    return Decision(False, f"only {hours:.0f}h old (<{min_age_days}d)", TOO_YOUNG)
```

This PR was created 2026-09-03T14:00:07Z, so it reaches 3 days at **2026-09-06T14:00Z**. The merge controller runs hourly at :37, so the first sweep that can pick it up is **~2026-09-06T14:37Z**. Every PR currently in the queue (10539, 10550, 10554, 10572, 10584, 10588, 10591, 10594) is older than this one and clears that gate.

So the sequence was: a one-off 0-day dispatch enqueued it early, it was evicted from the queue without ever being tested, and the normal 3-day gate now applies. Working as designed — disregard the "worth a look at the sweep" line in my previous comment.

State is unchanged and still green: approval standing on `88c7ce760`, both checks passing, clean merge into current `main` (`4569c848d`, re-verified by trial merge).

---

tannerzhang:
Merged as `263a125d4`. My two earlier comments diagnosing why it hadn't merged were both wrong — correcting them so nobody reuses the reasoning.

**"It was dequeued without ever being tested" was an unsound inference.** I based it on two things: no `gh-readonly-queue/main/pr-10703-*` branch on the remote, and no `merge_group` workflow run referencing `pr-10703`. Neither shows queue membership. GitHub only creates a `gh-readonly-queue` branch and a merge group for entries it is *actively testing*, so both are absent for an entry that is simply waiting further back. The PR was in the queue the whole time. The squash commit's author date is 2026-09-04T16:29:51Z, the exact moment it was enqueued, which is when GitHub created it.

**The 3-day timeline that followed was therefore also wrong.** I predicted the first possible merge at ~2026-09-06T14:37Z. It merged at 2026-09-05T05:25Z, about 33 hours earlier, because it had never left the queue and the `MIN_AGE_DAYS` gate was never re-applied to it.

What holds up: the `MIN_AGE_DAYS` reading itself is right (scheduled sweeps use 3, the 16:29Z run was a `workflow_dispatch` with 0), and the queue really did have 28 failed `merge_group` runs on 2026-09-04, which is a fair explanation for why a queue that size took ~13 hours to reach this entry. That backlog was the delay, not an eviction.

Also worth flagging for anyone scripting against this repo: `git merge-base --is-ancestor <pr-head> origin/main` returns false for a merged PR here, because merges are squashed. Check for the squash commit instead.

Content landed as intended — the note plus its one `mkdocs.yml` nav line, 143 insertions, no `kb/` changes.

---

Issue #8754, opened 2026-08-17
Curate necrotizing vasculitis (MONDO:0800113)


sierra-moxon:
Curate a dismech entry for **necrotizing vasculitis** ([MONDO:0800113](https://monarchinitiative.org/MONDO:0800113)).

> A type of vasculitis that is comprised of vasculitides that present with necrosis.

**Coverage check** (`origin/main`, `kb/`): MONDO:0800113 is not bound as a `disease_term`, subtype term, or MONDO mapping by any existing entry, and no open or closed curation issue covers it.

Six vasculitis entries already exist (`Allergic_Cutaneous_Vasculitis`, `CNS_Vasculitis`, `Cryoglobulinemic_Vasculitis`, `IgA_Vasculitis`, `Postinfectious_Vasculitis`, `Rheumatoid_Vasculitis`). MONDO defines this as a *type of vasculitis comprised of vasculitides that present with necrosis* — i.e. a class over other diseases, so the first decision is **Grouping vs Disease**: a `kb/groupings/` union with an explicit `grouping_basis` and membership criteria may model this better than a `Disease` entry. See the Disease Groupings section of CLAUDE.md.

Tracker: part of #1079.

PR #10705, opened 2026-09-03
Add necrotizing_vasculitis mechanism module with conformance bindings

tannerzhang:
## Summary

"Necrotizing vasculitis" is already curated in dismech, as a Grouping over eight vasculitides (issue #8754, PR #8860). What was missing is the mechanism those eight share. This PR adds it as a mechanism module and wires all eight members to it.

The Grouping asked for this itself. Its second membership criterion says vessel-wall necrosis in every member is produced by leukocytes degranulating inside the wall, and then admits the claim is prose-only "because no single shared GO process was annotated across the members". This PR curates that mechanism as `kb/modules/necrotizing_vasculitis.yaml` and gives the members a `conforms_to` target, so the claim becomes checkable.

## The chain

```
Vessel wall immune stimulus deposition        trigger           ANCA on primed neutrophils
  + endothelial activation                                      OR deposited immune complexes
              |
              v
Complement C5a amplification                  amplifier         <- avacopan (C5aR1) acts here
  + neutrophil recruitment  <------+                            alternative pathway (pauci-immune)
              |                    |                            classical pathway (immune-complex)
              v                    |
Intramural leukocyte degranulation |          central effector   KEY CONFORMANCE TARGET
  + oxidative burst ---------------+          (NETs scaffold complement: feed-forward)
              |
              v
Fibrinoid necrosis of the vessel wall         effector          HP:6000253, the biopsy lesion
              |
              v
Luminal compromise + tissue infarction        consequence       organ pattern differs by member
```

The key conformance target is the **central effector**, not the amplifier: it is the step that is most disorder-agnostic. The amplifier is the one node where the members genuinely diverge, because the two arms enter complement by different routes. The deep-research report recommended the amplifier instead, on the strength of its trial and knockout evidence; the module records why that recommendation was not followed.

## Conformance bindings

21 `conforms_to` additions across all eight Grouping members. No pathophysiology content was rewritten — each is a one-line addition.

| Member | Bindings |
|---|---|
| Allergic_Cutaneous_Vasculitis | 4 |
| Eosinophilic_Granulomatosis_with_Polyangiitis | 3 |
| Microscopic_Polyangiitis | 3 |
| Rheumatoid_Vasculitis | 3 |
| Cryoglobulinemic_Vasculitis | 2 |
| Granulomatosis_with_Polyangiitis | 2 |
| IgA_Vasculitis | 2 |
| Primary_Polyarteritis_Nodosa | 2 |

By module node: effector 7, trigger 5, central effector 4, amplifier 3, consequence 2.

**The central effector carries four of eight, not eight.** That gap is in the disorder entries, not the module, and the module says so rather than implying otherwise: IgA vasculitis and cryoglobulinemic vasculitis each describe neutrophil degranulation *and* the resulting fibrinoid necrosis in a single node, so they are wired to the effector, the more specific of the two claims they make. Splitting those bundled nodes is the named follow-up. It was deliberately left out — it would restructure other people's curations inside a module PR.

**Primary polyarteritis nodosa is wired at the terminal nodes only.** Its entry has no neutrophil or complement node, and its own evidence says the pathogenesis of idiopathic PAN remains enigmatic. The module's exclusion rule says the boundary is the C5a/neutrophil hub rather than the lesion, so PAN needed an explicit distinction: its mechanism is *not yet curated* (and Chapel Hill defines PAN as necrotizing arteritis with mural leukocyte infiltration, which puts it on this chain), whereas the excluded diseases have a *known and different* effector. DADA2 is the contrast — it reaches the same lesion through M1 macrophage skewing and responds to TNF inhibition, so it is deliberately not wired.

**The evidence behind the amplifier is asymmetric between the two arms.** The ANCA arm has genetic dissection (factor B and C5 knockouts protected, C4 knockouts not), a compartment-specific C5aR requirement, and an interventional RCT. The immune-complex arm rests on histopathology and review-level statements, with no branch-specific complement-blockade trial, and it enters complement by a different pathway. Convergence on an identical C5a-dependent step is therefore an inference from the shared lesion, not a demonstrated fact. This is filed as a `discussions` KNOWLEDGE_GAP on the amplifier node with a proposed trial, because it determines whether avacopan should be expected to work in cryoglobulinemic or IgA vasculitis at all.

## Other files

- `kb/groupings/Necrotizing_Vasculitides.yaml` — its criterion-2 prose said no shared GO process was annotated across members. That is now half-false, so the criterion records where the machine-checkable form lives, that four of eight members carry it today, and why the criterion still stays prose there.
- `research/modules/necrotizing_vasculitis-deep-research-openscientist.md` — falcon was the requested provider and returned 403 (auth). The run used the recorded-fallback path rather than a hand substitution, so the report carries `fell_back: true`, `requested_provider: falcon`, and the failed attempt with its status code.
- Cache: `GO:0002536`, `GO:0030194` and `UBERON:0035965` label rows, derived by term validation.

## Ontology notes

The module binds GO, CL and UBERON only, per the module contract.

Locations use `UBERON:0035965` *wall of blood vessel* rather than `UBERON:0001981` *blood vessel*, because the central effector is defined by location — the same degranulation in the lumen or the interstitium does not produce vasculitis, and the broader term cannot express that. `UBERON:0002522` *tunica media* is deliberately not used: it is the middle layer of an artery or vein, which excludes the dermal postcapillary venules that four of the immune-complex members involve.

The amplifier binds both `GO:0006957` (alternative) and `GO:0006958` (classical) under the parent `GO:0006956`, matching the two entry routes.

## Xogenesis anchor, with an OBO gap

The terminal output is a canonical structure made pathological rather than a new one, so the process genus is `OGMS:0000061` sub-typed as `OGMS:0000080` pathological transformation, at site `UBERON:0035965`. `MPATH:9` *fibrinoid necrosis* sits in MPATH's **process** branch, not the `MPATH:603` pathological-anatomical-entity subtree the Xogenesis convention normally draws from, and there is no MPATH continuant for a fibrinoid-necrotic vessel wall. That is recorded in the module notes as a gap rather than papered over by binding a continuant term that does not mean this.

## Validation

`just validate-module` passes on the module (schema, terms, and all 27 snippets verified as exact substrings of their cached references). `just validate-grouping` passes on the edited grouping. All eight edited conformers pass `linkml-validate` and `check-entity-refs`. `check-causal-targets`, `check-enum-values`, `check-duplicate-keys`, `check-term-cache-integrity`, `check-snippet-grading`, `check-title-snippets` and `check-snippet-length` pass. `pytest -k conforms_to`: 2728 passed.

The module cites 15 PMIDs. Note the deep-research report's own validation block reports `needs_review: true` — one of its quotes (`PMID:28029757`) did not match the cited paper. That PMID is not used anywhere in this PR.

Three curation history records are included: the module and its conformers, a pre-PR self-review round, and the round addressing this PR's review.

https://claude.ai/code/session_018oLgPt3Qgyx72SuaH9iwBZ

### DIFF
--- a/history/modules/necrotizing_vasculitis/2026-09-03T152936Z-claude-code-48a8b5.yaml
+++ b/history/modules/necrotizing_vasculitis/2026-09-03T152936Z-claude-code-48a8b5.yaml
@@ -1,0 +1,41 @@
+history_version: 1
+target:
+  kind: module
+  slug: necrotizing_vasculitis
+  path: kb/modules/necrotizing_vasculitis.yaml
+session:
+  id: 2026-09-03T152936Z-claude-code-48a8b5
+  timestamp: '2026-09-03T15:29:36Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10705
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - evidence
+  summary: 'Address automated review findings on PR #10705'
+  details: |
+    Third pass on kb/modules/necrotizing_vasculitis.yaml, addressing the two blocking findings from the ai4c-reviewer review on PR #10705. Both findings were verified against the repository before acting. Both were the same class of defect: a claim the module asserted without evidence, where the evidence was already sitting in a cache file this branch commits.
+
+    FINDING 1, classical-pathway claim unsupported. The amplifier node bound GO:0006958 complement activation, classical pathway and its description asserted that the immune-complex members enter the amplification loop by that route, while all four of its evidence items were ANCA/alternative-pathway (PMID:17200182 C5-deficient mice, PMID:19073822 twice for alternative pathway and C5aR, PMID:32602136 human C5a in AAV). So a GO binding and a mechanistic claim covering half the module's scope rested on nothing. The previous round's notes acknowledged the ANCA bias honestly, but acknowledging a gap is not the same as closing one when the source is available. Added PMID:16225258, which names deposited immune complexes activating the CLASSIC complement pathway and recruiting polymorphonuclear neutrophils in hepatitis-B-associated polyarteritis nodosa, cryoglobulinemic vasculitis and Henoch-Schonlein purpura - three of the module's own immune-complex members - and PMID:22844322 for the cryoglobulinemic member specifically. The reviewer also offered PMID:35396080; it was not used, because urticarial vasculitis is not among the eight members this module scopes, and citing it would have stretched the claim past the conformer set.
+
+    FINDING 2, consequence node claims outran its evidence. The node claimed thrombogenic occlusion, microaneurysm, rupture, fibrotic luminal narrowing and downstream infarction, supported by an organ-distribution statement (PMID:32855422) and a background-framing sentence (PMID:29078325), neither of which asserts any of those. Added two PMID:32855422 sentences that do: the thrombosis hallmark sentence and the chronic transmural-scarring sentence. The second reproduces a word break present in the cached full text ('characteri zed') exactly, because the snippet contract requires an exact substring; the explanation records why the odd spelling is there so a later reader does not 'fix' it and break the match.
+
+    With the prothrombotic claim now sourced, the node regained ontology grounding through GO:0030194 positive regulation of blood coagulation. The previous round had removed GO:0007596 blood coagulation from it - correctly, since that is the normal haemostatic cascade and duplicated the thrombogenesis module - but that left this as the only node in the module with no locations, cell types or biological processes, which the reviewer flagged. GO:0030194 is the right term because it is the positive-regulation child and it is now cited rather than assumed.
+
+    ALSO ADDRESSED FROM THE SUGGESTIONS. The DADA2 exclusion argument in notes was paraphrasing two papers without naming them; PMID:30565235 (M1 macrophage skewing) and PMID:32845415 (response to TNF-alpha inhibition) are now cited inline, so the exclusion is attributable. The stale UBERON:0002049 vasculature cache rows were dropped in a separate commit - the term was bound at the consequence node, removed in the previous round, and nothing in kb/ binds it now, making it the same category of churn as the NCIT FGFR2 row already cleaned. The PR description, which was auto-generated and contained wrong ontology labels (it described GO:0002536 as 'leukocyte migration involved in immune response' and UBERON:0002049 as 'blood vessel wall'), an undercount of the conformers (six rather than eight), and a statement that the module is anchored on the amplifier node when the key conformance target is the central effector, was rewritten from the actual diff.
+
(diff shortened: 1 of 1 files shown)


tannerzhang:
Both blocking findings are fixed, and both were right — each was a claim the module asserted without evidence, with the evidence already sitting in a cache file this branch commits.

### Finding 1 — classical-pathway claim

Added two evidence items to the amplifier node. `PMID:16225258` names deposited immune complexes activating the *classic* complement pathway and recruiting polymorphonuclear neutrophils in hepatitis-B polyarteritis nodosa, cryoglobulinemic vasculitis and Henoch-Schönlein purpura — three of this module's own immune-complex members, so it carries the claim for the arm rather than adjacent to it. `PMID:22844322` covers the cryoglobulinemic member specifically. The `GO:0006958` binding now has a source.

I did not use `PMID:35396080`. Urticarial vasculitis is not one of the eight members this module scopes, so citing it would stretch the claim past the conformer set.

### Finding 2 — consequence node

Added the two `PMID:32855422` sentences you pointed at: the thrombosis-hallmark sentence and the chronic transmural-scarring sentence. One caveat worth flagging for anyone editing that snippet later — the cached full text reads `characteri zed`, a PDF extraction artifact, and the snippet reproduces it exactly so the substring match holds. The `explanation` says so, to stop someone "correcting" the spelling and silently breaking the quote.

With the prothrombotic claim now sourced, the node regains ontology grounding via `GO:0030194` *positive regulation of blood coagulation*, which also closes your observation that it was the only node in the module with none. It got there because the previous round removed `GO:0007596` *blood coagulation* as the normal haemostatic cascade duplicating `thrombogenesis` — right call, but it left a hole, and `GO:0030194` is the term that is actually cited rather than assumed.

### Suggestions taken

The stale `UBERON:0002049` cache rows are dropped — same cleanup as the `NCIT:C124074` row, and thank you for catching that the earlier pass missed it. The DADA2 exclusion prose now cites `PMID:30565235` and `PMID:32845415` inline instead of paraphrasing them unattributed. The PR description is rewritten from the actual diff; it was auto-generated and had the wrong label for `GO:0002536`, the wrong label for `UBERON:0002049`, six conformers instead of eight, and it named the amplifier as the key conformance target when it is the central effector.

### Suggestions not taken

**GPA's B-cell/ANCA-production node should not conform to the trigger.** The trigger is the localization of an immune stimulus to the vessel wall with endothelial activation; ANCA production is the generation of the autoantibody, a step further upstream. The module already encodes exactly that relationship — rituximab's `target_mechanisms` points at the trigger and describes B-cell depletion as starving it of ANCA. Wiring the production node to the trigger would collapse the distinction the treatment block depends on.

**`HP:6000253` left unbound.** 7 of 168 modules carry a `phenotypes:` block, and that term is already the Grouping's machine-checkable criterion leaf. Binding it here would give one assertion two homes.

**Animal models deferred.** `PMID:27446086` and `PMID:31883830` would populate a real gap, but it is additive rather than corrective, so it belongs in its own change rather than widening a review-response commit.

On the two review articles graded differently: agreed it is a soft call. `PMID:35254509` is graded `OTHER` where it is quoted for a synthesis statement, `PMID:33713282` `HUMAN_CLINICAL` where it is quoted for the diagnostic histopathology. Each is internally consistent, and `check-snippet-grading` is keyed on the quoted sentence rather than the PMID, so it reports no divergence.

### Validation on the new head

```
just validate-module      → schema, terms, references all pass; 27/27 snippets verified
just validate-grouping    → passes
check-snippet-grading     → no new divergences vs origin/main
check-title-snippets      → no new title-quoting snippets
check-term-cache-integrity→ OK
pytest -k conforms_to     → 2728 passed
```

A history record for this round is included, and it names a mistake I repeated: two `reference_title` values were written from memory rather than read from the cache, the same slip as the previous round. The reference validator caught both. Every `reference_title` in the file has since been checked programmatically against its cached record.

---

PR #10706, opened 2026-09-03
curate: Frasier Syndrome (MONDO:0007635) — WT1 +KTS isoform loss driving gonadal and renal disease

### LINKED ISSUE #10699: Curate Frasier syndrome (MONDO:0007635)
Curate a dismech entry for **Frasier syndrome** ([MONDO:0007635](https://monarchinitiative.org/MONDO:0007635)) — a WT1 disorder in which intron 9 splice-donor variants shift the +KTS/-KTS isoform ratio, producing 46,XY gonadal dysgenesis with late-onset focal segmental glomerulosclerosis and gonadoblastoma risk.

The `WT1_Disorders` grouping added in #10448 names this entry as its outstanding member: "Frasier syndrome (MONDO:0007635) is a member of this concept and is deliberately absent from the member list because it has no Disease entry yet; that is a curation gap, not an exclusion, and it is the most important addition this grouping is waiting on." Curating it should make the +KTS isoform imbalance readable against the dominant-negative lesion curated in Denys-Drash.

Two boundaries to respect. `46_XY_Partial_Gonadal_Dysgenesis.yaml` already carries a thin `WT1-related PGD` subtype describing a Denys-Drash/Frasier continuum from a single case report — cross-link rather than duplicate, and note that Frasier classically presents with complete rather than partial dysgenesis. GeneReviews has retired Denys-Drash and Frasier as clinical designations in favour of one continuum, so the entry should state that framing rather than assert a clean split.

Stub: `stubs/Frasier_Syndrome.yaml`

yaseminbridges:
New entry for Frasier syndrome, a rare condition where a single splicing mutation in the *WT1* gene causes two very different problems at once: the testes fail to develop in children who are chromosomally male, and the kidneys' filtering cells are progressively damaged, leading to kidney failure. Both stem from the same underlying defect. This PR curates that shared origin as an explicit causal chain rather than listing the gonadal and renal features as separate consequences of "a WT1 mutation".

Closes #10699.

## The mechanism

*WT1* is normally made in two versions that differ by three amino acids (+KTS and -KTS), with +KTS usually the more abundant. The Frasier mutation sits in a splice site and stops the mutant copy from making +KTS at all. No abnormal protein is produced — the lesion is purely a shift in the ratio between two normal isoforms, which is what distinguishes it from Denys-Drash syndrome.

```
WT1 intron 9 donor splice-site variant
            │
            ▼
Loss of +KTS isoform from the mutant allele
            │
            ▼
   +KTS / -KTS isoform imbalance
            │
     ┌──────┴──────┐
     ▼             ▼
Failure of      Podocyte dysfunction and
testis          segmental glomerular
determination   scarring
```

## Evidence

All three mechanism-to-mechanism edges carry evidence on the edge itself, not just on the nodes at either end.

The branch point is the interesting one. A *WT1* +KTS mouse dosage series does the work that patient data cannot: heterozygous reduction of +KTS produces glomerulosclerosis, while homozygous loss produces complete XY sex reversal. Because the only variable is +KTS dose, this supports the causal links rather than merely establishing that both features co-occur in patients. Human transcript data then confirms the ratio shift is real in patients and not just predicted from the splice site.

`just validate` passes: schema, terms, and 32/32 evidence snippets verified against cached references.

## What I did not do

**The eight mechanism-to-phenotype edges are uncited** (for example, podocyte dysfunction → proteinuria). This is deliberate. I sampled 60 disorder files and 67% of such edges carry evidence — the majority convention, but not universal, and validation does not flag them. The phenotypes themselves are cited. Happy to fill these in if reviewers would rather the file sit at the stricter end of current practice.

**No schema changes.**


yaseminbridges:
All nine findings addressed. Summary of what changed and the two places I diverged.

## Fixed

**Item 1 (critical).** Snippet replaced with the full sentence. `check-snippet-length` now exits 0 against `origin/main`.

**Item 2.** You were right, and it was my error — the description pointed at a grouping file and a Denys-Drash entry that do not exist. Reworded to describe the contrasting mechanism without claiming it is curated here, and the absent counterpart is now recorded as a `CURATION_TODO` discussion rather than left implicit.

**Item 3.** Added a `diagnosis` section: WT1 molecular testing, karyotype (which is what stratifies tumor risk), and both surveillance schedules — proteinuria six-monthly to age ten and gonadal insufficiency yearly after puberty.

**Item 4.** Added Primary amenorrhea, Delayed puberty, and Hypergonadotropic hypogonadism downstream of `Failure of Testis Determination`.

**Item 5.** van Peer 2024 now carries the differential: the intron-9 Wilms figure (none in any intron-9 patient, against 18 of 19 truncating) and the intron-9 CKD course (all affected, median 15 years, versus 6 months for other splice sites). Shao 2023 supports the delayed-puberty presentation. PMID:10762296 anchors the continuum argument, which previously rested on GeneReviews alone. Drayer 2022 supports transplantation.

**Item 6.** `has_subtypes` for types 1/2/3, with the per-subtype tumor rates attached. "Unnecessary in type 3" is now resolvable without reading an evidence explanation.

**Item 7.** `animal_models` entry for the +KTS dosage series, `PARTIALLY_RECAPITULATES` against the podocyte node and `RECAPITULATES` against testis determination. The `limitations` note that the gonadal branch needs homozygous loss whereas patients are heterozygous.

**Item 8.** `GO:0000398` on both splicing nodes, `CL:0000653` on the podocyte node.

**Item 9.** Kidney transplantation (with the graft-recurrence rationale), dialysis, and post-gonadectomy hormone replacement.

## Two divergences

**`HP:0000132` is Menorrhagia, not Primary amenorrhea.** I used `HP:0000786`. Worth checking wherever that ID came from.

**OMIM and ORPHA mappings are not representable.** `DiseaseMappings` offers only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings`. Adding OMIM:136680 and ORPHA:347 would need a schema change, which felt out of scope for a curation PR — happy to open an issue if the slot is wanted.

## Still not done

The mechanism→phenotype edges remain uncited. You are right that van Peer and GeneReviews make several of the renal ones cheap now, but I have left them rather than expand this PR further.

Also skipped: `NCT01252901`, per your own reasoning; and the duplicate Shao cache entry is harmless, though I cite the PMID form. Drayer stays DOI-cited — it has no PMID in the cache, and I am not going to guess one.

Validated: `just validate` clean, plus `check-snippet-length`, `check-title-snippets`, `check-empty-snippets`, `check-snippet-boundaries`, `check-entity-refs`, `check-qualifier-terms`. 53/55 snippets verified against cache; the 2 skipped are DOI-prefixed and were checked by hand.

yaseminbridges:
All three Shao sentences swapped in. You were right that they cite the claim rather than sitting next to it — the hormone replacement case especially, where the only evidence was a sentence about the problem and not the intervention.

I verified all three as exact substrings of the cached text before using them.

## Two corrections in the same edit

**The Drayer scope caveat.** You are right that it did not make it across from the research report. The median transplant age of 5 years now carries an explicit guard that the cohort is spectrum-wide with a median presentation of 0.9 years, so it is not read as a Frasier figure. That is the same discipline the `notes:` block commits to, and I had broken it in the treatment I added.

**The recurrence claim.** Also right, and this one was mine rather than a sourcing slip — I wrote that recurrent disease "is not expected" and neither snippet says so. "No post-transplant malignancy" is not "no recurrence", and the two allograft losses are unattributed in the source. It now reads as mechanistic rationale, explicitly labelled as not evidenced by the cited sources, rather than as an outcome claim.

## The two suggestions I did not take

**`ORPHA:347` cannot be fetched.** `just fetch-reference ORPHA:347` fails with "No source found for reference type", and `data/orphadata/` holds only `MANIFEST.yaml` locally — the bulk XML is not downloaded. So the Frasier-scoped frequency source my `notes:` block asks for is not reachable through the reference-cache path that every other citation in this entry uses. Worth knowing that the ORPHA suggestion is not actionable as written; if ORPHA frequencies are wanted on disorder entries, the bulk-refresh route is a different workflow and probably its own PR.

**The animal model `publication:`.** It points at PMID:35894442 because that review is where the quoted sentence actually comes from. Repointing it at the primary Hammes paper would mean citing a paper I have not fetched or verified, and I would rather not put an identifier in the file that I have not checked — I already guessed a PMID once this session and landed on an unrelated orchiopexy meta-analysis. Happy to do it properly if you can give me the PMID.

Lipska 2014 and the type-2 gonadoblastoma DOI remain uncited, as you said was reasonable.

`just validate` clean at 56/58 (the 2 skipped are DOI-prefixed, hand-checked), plus `check-snippet-length`, `check-title-snippets`, `check-snippet-boundaries` and `check-entity-refs`.

yaseminbridges:
Filled the mechanism-to-phenotype edges as well. You flagged twice that they were cheap once van Peer and GeneReviews were in hand, and that was right — **the entry now has no uncited causal edge: 15 of 15, up from 3 of 12 when this PR opened.**

Each edge cites the claim that the mechanism produces that specific finding, rather than reusing the node-level evidence. Three worth calling out:

- **Gonadoblastoma** is cited to a sentence tying tumor risk to the dysgenetic gonad specifically, so the edge asserts a consequence of failed testis determination rather than an independent tumor predisposition sitting next to it.
- **Stage 5 CKD** uses the intron-9 cohort figure — every patient followed developed CKD, median 15 years — rather than a generic statement that the disease progresses.
- **Steroid-resistant nephrotic syndrome** uses the proteinuria-to-SRNS progression, which carries the ordering the node asserts.

Snippets are at 69, from 32 at open. The history record is updated to match.

All gates: `just validate` clean, 67/69 verified (2 DOI-prefixed skipped, hand-checked), plus `check-snippet-length`, `check-title-snippets`, `check-snippet-boundaries`, `check-snippet-grading`, `check-entity-refs`, `check-empty-snippets`.

Ready for re-review. Still open from your last round, both non-blocking and both explained above: `ORPHA:347` is not fetchable through the reference-cache path, and the animal model's `publication:` stays on the review that carries the quoted sentence until I have a verified PMID for the primary Hammes paper.

Issue #10193, opened 2026-08-30
prevalence_class is never cross-checked against rate_per_100000, so a band off by a decade passes every gate


sierra-moxon:
`prevalence_class` and `rate_per_100000` record the same fact twice, and nothing checks that the two agree. A curator can write a rate of 4.3 per 100,000 next to `BAND_1_9_PER_1000000` — off by a factor of ten — and every validator, gate, and test in the repo passes.

I hit this in #10022. Three of that entry's four prevalence records had bands off by exactly one decade. They were caught by reading the file, not by CI.

The bands are fixed numeric intervals, so this is mechanically checkable, and the mapping already exists: `scripts/migrate_prevalence.py::_band_from_rate()`. It runs at migration time and nothing re-checks the result afterwards.

## Scope

Applying `_band_from_rate()` to every `prevalence` record in `kb/` that carries both a numeric `prevalence_class` and a rate:

| | Count |
|---|---|
| Records with both a numeric class and a rate | 950 |
| Class disagrees with the rate | 39 |
| — rate sits exactly on a band boundary | 18 |
| — genuine disagreement | 21 |

All 21 are off by exactly one band. A sample:

```
   0.93 /100k   BAND_1_9_PER_100000  -> BAND_1_9_PER_1000000   Appendiceal_Neoplasm.yaml
   74.6 /100k   ABOVE_1_IN_1000      -> BAND_1_5_PER_10000     Cartilage-Hair_Hypoplasia.yaml
  20.57 /100k   BAND_1_9_PER_100000  -> BAND_1_5_PER_10000     Complex_Regional_Pain_Syndrome_Type_1.yaml
  125.0 /100k   BAND_1_5_PER_10000   -> ABOVE_1_IN_1000        Congenital_Sucrase-Isomaltase_Deficiency.yaml
    2.0 /100k   BAND_1_9_PER_1000000 -> BAND_1_9_PER_100000    Hailey-Hailey_Disease.yaml
   10.4 /100k   BAND_1_9_PER_100000  -> BAND_1_5_PER_10000     Thoracic_Aortic_Aneurysm.yaml
```

About 2% of records. Small, but each one is a wrong number sitting beside a right one, and the class is what filtering and rollups read.

[...]

Issue #7005, opened 2026-07-25
prevalence_class contradicts rate_per_100000 in 9 KB records — add a deterministic QC check


cmungall:
## The problem

`Prevalence` records carry both a numeric `rate_per_100000` and a categorical `prevalence_class`. Nothing checks that they agree, and **9 records currently disagree**.

I found this the hard way: I introduced this exact error **three times** in one review sweep (#6991, #6966, #7002), each caught by review rather than by tooling. That prompted me to audit the whole KB.

## Audit results

312 records carry both fields. Running each `rate_per_100000` through the repo's own canonical `_band_from_rate()` (`scripts/migrate_prevalence.py:186`, cited as authoritative at `src/dismech/schema/dismech.yaml:276`):

### Genuine numeric-band errors (3)

| file | rate | declared | expected |
|---|---|---|---|
| `Fibrochondrogenesis.yaml` | 0.1 | `BELOW_1_IN_1000000` | `BAND_1_9_PER_1000000` |
| `Glutathione_Synthetase_Deficiency.yaml` | 0.1 | `BELOW_1_IN_1000000` | `BAND_1_9_PER_1000000` |
| `SCN2A-Related_Developmental_and_Epileptic_Encephalopathy.yaml` | 7.5 | `BAND_1_5_PER_10000` | `BAND_1_9_PER_100000` |

The two `0.1` cases are boundary errors — 0.1 per 100,000 *is* 1 per million, which sits in the 1–9 band, not below it.

### Qualitative class paired with a numeric rate (6)

| file | rate | declared |
|---|---|---|
| `Allergic_Cutaneous_Vasculitis.yaml` | 1.54 | `RARE` |
| `Carbon_Monoxide_Poisoning.yaml` | 15.1 | `COMMON` |
| `Chronic_Beryllium_Disease.yaml` | 5500.0 | `UNKNOWN` |
| `DEPDC5-Related_Epilepsy.yaml` | 30.0 | `RARE` |
| `Lupus_Nephritis.yaml` | 50000.0 | `COMMON` |
| `Punctate_Palmoplantar_Keratoderma.yaml` | 1.17 | `RARE` |

[...]

PR #10723, opened 2026-09-03
curate(CDG-IIw): add SLC37A4-CDG with competing localization hypotheses

### LINKED ISSUE #10711: Curate congenital disorder of glycosylation, type IIw (MONDO:0030437)
Curate a dismech entry for **congenital disorder of glycosylation, type IIw** ([MONDO:0030437](https://monarchinitiative.org/MONDO:0030437)) — an autosomal *dominant* CDG caused by a recurrent de novo SLC37A4 variant that abolishes the ER retention signal of the glucose-6-phosphate transporter, mislocalising it to the Golgi and disrupting glycosylation, presenting with liver disease and a coagulopathy.

Note the lump/split question here is not the usual one. `SLC37A4` is already in the KB as the *recessive* glycogen storage disease type Ib gene (`kb/disorders/Glycogen_Storage_Disease_Type_I.yaml`, `kb/groupings/Glycogen_Storage_Diseases.yaml`). CDG-IIw is a mechanistically distinct dominant disease from the same gene, so the decision is whether that separation holds, not whether it is a subtype of GSD I.

Stub: `stubs/Congenital_Disorder_Of_Glycosylation_Type_IIw.yaml`

aleixpuigb:
Adds `kb/disorders/Congenital_Disorder_of_Glycosylation_Type_IIw.yaml` for congenital disorder of glycosylation type IIw ([MONDO:0030437](https://monarchinitiative.org/MONDO:0030437)), also called SLC37A4-CDG.

Closes #10711

## The lump/split call

`entry_type: DISEASE`, and the interesting part is what it is being split *from*. `SLC37A4` is already in the KB as the recessive glycogen storage disease Ib gene (`kb/disorders/Glycogen_Storage_Disease_Type_I.yaml`, `kb/groupings/Glycogen_Storage_Diseases.yaml`). CDG-IIw is a dominant disease of the same gene, caused by a recurrent de novo `c.1267C>T (p.Arg423*)` allele that truncates the C-terminal ER retrieval motif.

The reason it is a separate disease rather than a dominant-negative form of GSD Ib is that the lesion is the escape of a truncated transporter from the ER, not the loss of transport activity. The intact allele supplies enough transporter for normal glucose-6-phosphate handling, so there is no fasting hypoglycaemia, no lactic acidosis and no neutropenia. There is a pathophysiology node recording exactly that negative, because it is the claim the split rests on.

## What is modelled

The chain runs: loss of the ER retrieval motif → mislocalization out of the ER → reduced intraluminal Golgi pH and altered Golgi morphology → disturbed N-glycan processing of secreted hepatic glycoproteins → hepatic coagulation factor deficiency and hepatocellular injury.

Two things worth a reviewer's attention:

**The localization is disputed, and the entry does not resolve it.** PMID:32884905 proposes that removing the ER retrieval signal exposes a weak Golgi retention signal that captures the protein in the Golgi. PMID:33964207, using a CRISPR base-edited hepatoma line carrying the patient allele, places a portion of the mutant protein in a distinct non-Golgi compartment, possibly ER exit sites. Both agree on the downstream Golgi pH and morphology change, and nothing in the clinical phenotype discriminates. So it is carried as two `mechanistic_hypotheses` groups — `golgi_retention_model` (CANONICAL) and `er_exit_site_model` (EMERGING) — with the shared downstream edge annotated to both, plus a `KNOWLEDGE_GAP` discussion.

**Two evidence items are deliberately `REFUTE`.** PMID:34245688 reports that procoagulant factors II and XI fall alongside antithrombin, suggesting a partly preserved haemostatic balance, and patients have come through cardiac and cleft surgery. PMID:37193126, a CDG haemostasis review covering the same coagulation profile, calls the balance precarious under metabolic stress. Both are curated on the same node so the tension is visible rather than silently resolved toward the reassuring reading.

## Deep research

One openscientist run, committed under `research/`. It ran 766s over 5 iterations. Its own validation reported 17/17 references resolved with `confabulation_rate: 0.0`, but only 6 of 17 weighed as on topic, and `term_validation.needs_review` was true (three obsolete GO terms, one mislabelled MONDO name). `just preflight-dr` passed with SLC37A4 mentioned 16 times, while noting the report carries OMIM 232220 (GSD Ib) alongside 619525 — which is the gene-sharing conflation the low on-topic ratio was pointing at.

Taken from the report:

- **PMID:10940311** — a 2000 structure-function study showing the last 12 residues of the G6PT cytoplasmic tail are dispensable for transport activity. That is independent support for the "transport is preserved" node, from twenty years before the disease was described. Graded `directness: INDIRECT` because the experiment was done on GSD Ib alleles and does not mention this disorder.
- **PMID:37193126** — CDG haemostatic defects, giving the antithrombin-plus-factor-XI signature and, usefully, that this profile is diagnostically discriminating against liver failure. That matters here because these patients also have liver disease.

Not taken, and why: PMID:28126686 and PMID:35506446 are about GSD Ib and G6PC3 neutropenia, not this disorder. PMID:41136424 is a 2025 G6PT cryo-EM structure paper whose text is framed entirely around GSD Ib; it would be a fine transporter-biology citation but importing it risks pulling GSD Ib framing into a CDG entry.

## Validation

```
just validate  -> schema OK, terms OK, 54/54 snippets verified
     / check-qualifier-terms / check-enum-values  -> all OK
just validate-history  -> OK
```

`reference_title` values were filled with `just normalize-reference-titles` after the ones I typed by hand failed the title check.

## Not done

No `datasets:` block. `just discover-datasets` searches by disease name and gene, and for a nine-patient disorder whose gene is famous for a different disease, every candidate it could surface would be a GSD Ib dataset — the exact relevance trap `docs/dataset-curation.md` describes. Better left absent than filled with resolvable but wrong accessions.

---


aleixpuigb:
All six items fixed and pushed in 34ac1e504. Taking them in order, then two things I found while doing it.

| Item | What I did |
|---|---|
| 1. Wrong NCIT binding | Fixed. `NCIT:C116475` "Fresh Frozen Plasma Transfusion", verified reachable from `NCIT:C25218`. |
| 2. Cardiac unmodeled | Added `HP:0001626` at `VERY_FREQUENT`, plus a second `INDIRECT` item from PMID:34245688 (patients underwent heavy cardiac surgery). |
| 3. Five missing `reference_title` | Backfilled. Exactly five, matching your count. |
| 4. Factor XI unmodeled | Added `HP:0001929` at `VERY_FREQUENT`, with the PMID:37193126 item establishing the antithrombin-plus-F11 pairing as the characteristic pattern. |
| 5. Generic glycosylation biomarker | Transferrin biomarker rebound to `HP:0012301`. |
| 6. Pathograph reaches 2 of 16 | Nine bare-name edges added from the coagulation and hepatic-injury nodes. Now 10 of 18. |

**Item 1 was a real error, not a near-miss.** I picked `NCIT:C15431` out of the term cache because it was there and validated, without checking what it meant. The label matched the ID so `just validate-terms` passed on a term naming a completely different intervention. Worth noting the class: this is the same failure the `qualifiers` blind spot in `CLAUDE.md` describes — a plausible code naming the wrong concept — except in a slot that *is* validated. Validation confirms the ID and label agree; it cannot confirm they are the right concept for the claim.

**Item 3 is a tooling nuance worth recording.** I had run `just normalize-reference-titles`, which only *re-syncs titles that are already present* — it reported "0 re-synced" and I read that as clean. `just backfill-reference-titles` is the one that *adds* missing ones, and it added exactly five. The two recipes read as near-synonyms from their names and do different halves of the job.

## Two things I found that were not in the review

**Choanal atresia is probably not a feature of this disorder.** PMID:32884905 also says:

> The choanal atresia was surgically treated as it was in her mother who also had bilateral choanal atresia at birth.

The mother had the same malformation, and the SLC37A4 allele is *de novo* in the proband — so the mother cannot carry it. The atresia therefore segregates independently of the causal variant in that family, and the parsimonious reading is an unrelated familial trait. I kept the phenotype, since it was reported in an affected patient, but added a `REFUTE` evidence item carrying that sentence and said so in the phenotype `description`.

**On item 6, the remaining eight phenotypes are deliberately unconnected, and I would push back on connecting them.** Your two suggested source nodes were right and I used both. What is left is thrombocytopenia, cardiac involvement, choanal atresia, micrognathia, ankyloglossia, scoliosis, type 1 diabetes and membranoproliferative glomerulonephritis — and no source curated here proposes a route from the mechanism to any of them.

That is not an oversight; it is the entry's most interesting unresolved tension, so I added a `KNOWLEDGE_GAP` for it rather than papering over it with edges. The glycosylation defect is *tissue-restricted*: serum glycoproteins are grossly abnormal while N-glycans in fibroblasts and undifferentiated iPSC are normal (PMID:33964207). That restriction is load-bearing evidence in this entry — it is why the liver is treated as the site of disease. But micrognathia, ankyloglossia and scoliosis are developmental, arise in tissues where the glycan profile is reportedly normal, and predate any systemic effect of a liver-derived secretory defect. Either the tissue restriction is less complete than fibroblast and iPSC profiling suggests, or these malformations arise by a route the glycan phenotype does not capture. Drawing an edge would assert an answer the literature does not have, and `CLAUDE.md` is explicit that an unconnected node is honest where a fabricated edge is not.

Type 1 diabetes and MPGN stay unconnected for the separate reason already recorded in the entry: both are single-patient reports in a cohort of nine, and there is a `KNOWLEDGE_GAP` asking whether they belong to the disorder at all.

Validation after the fixes: `just validate` passes with 59/59 snippets verified; `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets` and `check-reference-titles` all clean.

---

aleixpuigb:
Thanks for the approval. Pushed 000b9641c with the two follow-ups that seemed worth taking, which resets the approval — I judged that acceptable since one of them corrects a factual error in the entry.

**Cardiac frequency band.** You were right, and I verified both sentences in `references_cache/PMID_33728255.md` before changing anything rather than taking the arithmetic on trust:

> "Ophthalmological, hearing, and cardiac assessments were normal"

> "as were ophthalmological, cardiological, and electroencephalographical examinations"

7 of 9, not 7 of 7. Rebanded to `FREQUENT`, and I added a `REFUTE` evidence item carrying the first sentence so the denominator correction lives in the record rather than only in the description prose — a reader checking why the band is 78% can now see the patient who supplies the denominator.

Worth noting this is the third instance across these five PRs of the same shape: a legal enum value in the wrong bucket. `Renal cyst`'s `FREQUENT` band on #10730, both `prevalence_class` values on the same PR, and this. All three are *derivable* — a cohort count determines `FrequencyEnum`, `rate_per_100000` determines `prevalence_class` — and none is checked. `scripts/migrate_prevalence.py` already contains `_band_from_rate`, so half the check exists as a function and is not wired to a gate. I have filed that as #10748.

**Biomarker `preferred_term`.** Taken. The bindings stay on the generic `HP:0012345` since HPO has no term for either concept, but `preferred_term` now reads `Serum high-mannose and hybrid N-glycan accumulation` and `Serum bikunin proteoglycan heavy-form deficiency` rather than echoing the ontology label, so the precision the binding lost is at least in the record.

**History record.** Added, covering both review rounds. I left the CREATE record untouched — you are right that its "54/54" was true when written, and rewriting an append-only record to match a later state would be the wrong fix.

**Not taken**, and I would push back gently on one: `modifier: ABNORMAL` on Golgi organization is arguably better English, but the source reports a *gene dosage-dependent alteration* in Golgi morphology in the direction of disruption, and `DECREASED` is PATO-bound where `ABNORMAL` is not, so I would rather keep the queryable value. `therapeutic_modality: PROTEIN_REPLACEMENT` on FFP I also left as `OTHER`: fresh frozen plasma is whole plasma carrying many factors plus everything else in it, not a replacement of a specific protein, and the treatment record covers antithrombin III, prothrombin complex and vitamin K alongside it. The no-animal-model `KNOWLEDGE_GAP` is a fair suggestion and I did not take it only because the entry already carries an `experimental_models` entry (the base-edited hepatoma line) whose `limitations` says what it does not model.

`just validate` clean, 60/60 snippets verified, all offline gates clean.

---

aleixpuigb:
Correction to my previous comment: I wrote that I had filed the derivable-band consistency problem as "#10748". **That issue number does not exist** — I wrote it before searching, and when I did search, the problem was already filed twice over:

- **#10193** — `prevalence_class` is never cross-checked against `rate_per_100000`
- **#7005** — the same, with a census of 9 affected KB records and a request for a deterministic QC check
- **#9390** — the `FrequencyEnum` half: a phenotype `frequency:` band has no machine-readable link to the count that grounds it

I have [added the two 17q12 instances to #10193](https://github.com/monarch-initiative/dismech/issues/10193#issuecomment-5529850493), including the detail that `scripts/migrate_prevalence.py` already contains `_band_from_rate` — so the oracle this needs exists as a function and is simply not wired to a gate.

Nothing in the entry changed; only the issue reference in my comment was wrong.

---

Issue #7461, opened 2026-07-31
claude-code-review: rate-limit guard misses "You've hit your limit", reports quota exhaustion as a bad-token/bad-model error


jmcmurry:
## Summary

When the review agent dies from **quota exhaustion**, `.github/workflows/claude-code-review.yml` reports it as an *"invalid/retired `--model`, a bad `CLAUDE_CODE_OAUTH_TOKEN`, or an aborted run"*. The workflow has a correct rate-limit branch — the string match just doesn't cover the message the API actually returns.

## The bug

`.github/workflows/claude-code-review.yml:215`

```python
if reason == "rate_limit" or "usage limit" in haystack or "weekly limit" in haystack:
```

The message emitted on quota exhaustion is:

```
Review agent errored (is_error=true); NO review was posted.
error='' detail="You've hit your limit · resets 8pm (UTC)"
num_turns=2 cost=$0.291 subtype=success
```

`"You've hit your limit"` contains neither `"usage limit"` nor `"weekly limit"`, and `reason` is not `"rate_limit"` — so it falls through to the generic branch at line 227.

Note also `subtype=success` and a non-error `error=''`: the *only* signal that this was a quota failure is that `detail` string. And `num_turns=2` means the agent died before reading the diff — it never reviewed anything.

## Suggested fix

```diff
-if reason == "rate_limit" or "usage limit" in haystack or "weekly limit" in haystack:
+if (
+    reason == "rate_limit"
+    or "usage limit" in haystack
+    or "weekly limit" in haystack
+    or "hit your limit" in haystack
+):
```

`"hit your limit"` (rather than the full sentence) also catches the `"You've hit your usage limit"` and `"You've hit your session limit"` variants, which use the same stem with different middles.

## Why it's worth fixing

This cost roughly 40 minutes of misdiagnosis tonight. Five PRs from the #1079 curation batch — #7380, #7451, #7452, #7454, #7455 — showed `claude-review: failure` with `test (3.13)` green, and the surfaced error pointed at `agent-config.yaml`'s `--model` and at the OAuth token. Both are load-bearing shared config, so the misleading message points an investigator at exactly the things they should not touch. One failure was also a **0-second crash**, which reads as a workflow defect rather than a quota wall.

With the fix, the existing branch already does the right thing — it prints `⛔ Claude review skipped — usage limit reached` and writes a job summary, which is immediately actionable ("wait for reset") instead of a wild goose chase.

## Related

[...]

PR #10730, opened 2026-09-03
curate(17q12): add chromosome 17q12 deletion syndrome as a contiguous gene syndrome

### LINKED ISSUE #10712: Curate chromosome 17q12 deletion syndrome (MONDO:0013797)
Curate a dismech entry for **chromosome 17q12 deletion syndrome** ([MONDO:0013797](https://monarchinitiative.org/MONDO:0013797)) — a recurrent ~1.4 Mb NAHR-mediated microdeletion whose HNF1B haploinsufficiency gives renal cysts and early-onset diabetes (RCAD/MODY5), with a neurodevelopmental and psychiatric burden that the point-mutation form of HNF1B disease does not carry.

The contiguous-gene aspect is the interesting part: HNF1B haploinsufficiency accounts for the renal and pancreatic phenotype, but the deletion removes ~15 additional genes, and the neurodevelopmental phenotype is the part that separates the deletion from an HNF1B point mutation. HNF1B currently appears in the KB only inside other entries (`Diabetes_Mellitus`, `Familial_Vesicoureteral_Reflux`, `Mitchell-Riley_Syndrome`, `Prune_Belly_Syndrome`); there is no dedicated HNF1B or 17q12 entry.

Stub: `stubs/Chromosome_17q12_Deletion_Syndrome.yaml`

aleixpuigb:
Adds `kb/disorders/Chromosome_17q12_Deletion_Syndrome.yaml` for chromosome 17q12 deletion syndrome ([MONDO:0013797](https://monarchinitiative.org/MONDO:0013797)).

Closes #10712

## The lump/split call

`entry_type: DISEASE`, and the reason is an asymmetry rather than a novelty. The deletion removes about fifteen genes, one of which is HNF1B. Its renal and pancreatic phenotype is HNF1B haploinsufficiency and is the same phenotype an HNF1B point mutation produces. Its neurodevelopmental and psychiatric phenotype is not: PMID:21055719 found the deletion in 18 of 15,749 patients referred for neurodevelopmental testing and 0 of 4,519 controls, then in 4 of 6,340 schizophrenia patients and 0 of 47,929 controls, and concludes the syndrome is a contiguous gene syndrome extending beyond renal cysts and diabetes, which HNF1B mutations alone cause.

So it is not a subtype of HNF1B disease. The pathograph branches at the deletion into an HNF1B arm and a second arm that names the class of lesion — "haploinsufficiency of additional dosage-sensitive interval genes" — because which gene it is has not been established. I would rather have a node that says that than one that guesses.

## GeneReviews baseline

PMID:27929632 is fetched, cached, and tagged `GeneReviews` in the top-level `references:` block. Every Clinical Characteristics feature is carried into `phenotypes:`, frequencies come from its ascertainment-corrected figures, and the *Agents/Circumstances to Avoid* section is curated as two treatment records rather than left in prose:

- transplant immunosuppression avoiding tacrolimus and mTOR inhibitors, because transplantation raises post-transplant diabetes risk in a population already predisposed
- lithium and weight-gaining antipsychotic cautions, which is where this syndrome's psychiatric and renal/diabetic arms collide

## Two things a reviewer might push back on

**The deletion is renally *milder* than a point mutation.** PMID:28420700 compared 101 HNF1B point-mutation carriers with 100 deletion carriers and found CKD stage 3-4 and end-stage renal disease significantly *less* frequent in the deletion group, both at diagnosis and long term (40 of 78 vs 71 of 91, P = 0.0003). A pure haploinsufficiency model predicts equivalence. I have not resolved it — it is a `KNOWLEDGE_GAP` discussion that names both candidate explanations (a dominant-negative effect of some missense alleles, or ascertainment by referral route) and records that PMID:31498910, which asked the differentiation question directly across 82 studies, concluded only that both groups are heterogeneous.

**The HNF1B score is curated but is not a test for this syndrome.** PMID:24897035's 17-item score is a genuinely validated instrument (AUC 0.78 in 433 individuals, NPV >99% at cutoff 8), so it is a `PHENOTYPE_ALGORITHM` definition with `VALIDATED_AGAINST_GOLD_STANDARD`. The description says explicitly that it is validated for HNF1B disease as a whole, that its specificity is 41.1% so a positive score is not diagnostic, and that the sequencing it triggers cannot see a 1.4 Mb deletion. There is a second `DIAGNOSTIC_CRITERIA` definition for the microarray detection that can.

## Deep research

One openscientist run, committed under `research/`.

**It shipped with no validation blocks at all.** Term validation aborted on a 5-second EBI OLS read timeout, which discarded the reference-validation results as well. That is dismech#10396 and I [commented there](https://github.com/monarch-initiative/dismech/issues/10396#issuecomment-5527972608) with the reproduction, plus [a correction](https://github.com/monarch-initiative/dismech/issues/10396#issuecomment-5528027472) once I established the workaround only half works.

- `just validate-research-reference` retro-fitted the reference section: 27/27 resolved, 22 of 27 on topic, 3 of 5 quoted claims found in source. Worth having — the two failures are near-misses rather than confabulations (a Unicode `Mg²⁺` against the abstract's `Mg(2+)`, and an ellipsis-joined quote), which is exactly the distinction that section exists to let a curator draw.
- `just validate-research-terms` failed on four consecutive attempts, on a different CURIE each time (`CHEBI:17234`, `HP:0000107`, `ORPHA:261265`, `ORPHA:93111`). **So this report carries no term-validation section**, and I am saying so here rather than leaving it as an absence a reviewer has to notice. `HP:0000107` is in `cache/hp/terms.csv`, so a cache-first path would have avoided that one entirely.

Taken from the report, and it materially improved the entry:

| PMID | What it added |
|---|---|
| 32778765 | Population birth prevalence, ~1 in 4,000 from 12,252 unselected trios. Replaced a referral-cohort rate as the primary prevalence record; the referral rate is kept below it, labelled as such. |
| 26340261 | Hypomagnesaemia in 50-60% with high fractional magnesium excretion. Replaced an `INDIRECT` monitoring-guidance quote with direct evidence and a frequency band. |
| 24897035 | The validated HNF1B score above. |
| 31498910 | Deletion vs intragenic variant comparison, cited on the open genotype-phenotype gap. |
| 24204001 | The PCBD1-HNF1B-FXYD2 route to distal tubule magnesium reabsorption, graded `INDIRECT` because the experiment interrogates PCBD1 loss in an overexpression cell line. |

## Two corrections made during curation

Both worth stating because a reviewer would otherwise have to find them.

1. Two snippets were initially written from memory and failed reference validation. That is the failure mode the evidence SOP warns about, and it was caught by the gate rather than by me.
2. Fixing the second one changed a claim. PMID:22130109 does not independently replicate the 17q12 schizophrenia association — it reports the deletion as "more common in schizophrenia cases but not significantly so". The entry now cites the accurate quote, grades it `directness: INDIRECT`, and says in prose that the replication is partial rather than implying two concordant significant datasets.

## Cache defect noted

`references_cache/PMID_24204001.md` holds the same sentence twice, once with the italicised gene name concatenated to its neighbours (`theFXYD2promoter`) and once clean. The snippet is taken from the clean occurrence. Not filed separately since it is a single-file artefact of markup stripping, but flagging it in case it is systematic.

## Validation

```
just validate  -> schema OK, terms OK, 50/50 snippets verified
     / check-qualifier-terms / check-enum-values             -> all OK
     / check-title-snippets / check-folded-hyphens           -> no new findings
just validate-history  -> OK
```

---


aleixpuigb:
All five IMPORTANT items and most of the suggestions are in 9247b7c15. Two I did not take, with reasons.

| Item | What I did |
|---|---|
| 1. No `reference_title` | Backfilled on all 69 evidence items. |
| 2. `Renal cyst` band | `VERY_FREQUENT` → `FREQUENT`. 122/166 is 73.5%. Added your secondary point too: the evidence item is now `directness: INDIRECT` because that cohort is HNF1B disease overall, roughly half point mutations. |
| 3. GeneReviews cherry-picked | Added liver abnormalities, eye abnormalities and congenital heart defects from the same sentence, plus hypokalaemia and hyperuricaemia from the monitoring list. |
| 4. No `animal_models:` | Added the zebrafish model. See the caveat below about the second paper. |
| 5. HNF1B target axis missing | PKHD1 and UMOD from PMID:15649945, plus pdx1/shh/pax2/wt1 from the zebrafish work, on the haploinsufficiency node. |

Suggestions taken: `directness: INDIRECT` on the three surveillance-only phenotypes; the 50% transmission risk and prenatal/PGT statement in `inheritance`; hypokalaemia and hyperuricaemia as phenotypes; and the UK Biobank microduplication contrast, which was the best of them — deletions associate with diabetes and duplications do not, in 450,993 individuals, which is a dosage-specificity argument the entry did not have.

**On item 1, a tooling note.** I *had* run `just normalize-reference-titles`, which reported "0 re-synced" and I read that as clean. It only re-syncs titles that are already present. `just backfill-reference-titles` is the one that adds missing ones. The two recipe names read as near-synonyms and do different halves of the job; the same thing bit me on #10723.

## Item 4: I used the zebrafish model but not PMID:20378824

The zebrafish `vhnf1` mutant is exactly right and I have used it for three links, including the `FAILS_TO_RECAPITULATE` you suggested — which is a good structural signal, since the model cannot have a neurodevelopmental phenotype: it perturbs only the HNF1B orthologue, and the entry's whole argument is that the brain limb comes from other interval genes. Its `NO_EVIDENCE` evidence item records that the authors relate their model to MODY5 and glomerulocystic kidney disease and make no neurodevelopmental claim.

**I did not use PMID:20378824 as an HNF1B model, and I would push back on it.** It is a conditional *Notch1/Notch2/RBP-J* inactivation model, and it explicitly dissociates its phenotype from HNF1B:

> Epithelial stratification correlated neither with reduced expression of the transcriptional regulator of ciliary proteins TCF2/HNF1beta nor with loss of apical-basal polarity.

TCF2 is HNF1B. So the paper reports that its cystic phenotype arises *without* reduced HNF1B, via mitotic spindle orientation in the proximal tubule. Curating it as an animal model of HNF1B haploinsufficiency would attribute to HNF1B a mechanism the paper specifically excludes it from. Happy to add it if you read it differently, but as written it looks like a Named Entity Confusion by mechanism rather than by name.

## Item 6 of the suggestions: OMIM mapping is not representable

`mappings:` cannot carry OMIM. `DiseaseMappings` has exactly four slots — `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings` — and `omim_mappings` fails schema validation:

```
[ERROR] Additional properties are not allowed ('omim_mappings' was unexpected) in /mappings
```

That looks like a real schema gap rather than an oversight on this entry: OMIM is the primary identifier for Mendelian disease, and the class carries slots for two ICD variants and NCIT but not for it. I have removed the block rather than wedging the ID somewhere it does not belong, and will file it separately.

## The NAHR suggestion

Not added, because I could not source it from anything in this PR. The recurrence and fixed breakpoints are stated and used (they are why microarray detects it and why sequencing panels miss it), but none of the ten cited papers states the segmental-duplication mechanism generating the interval. I would rather leave the mechanism unstated than assert it from background knowledge — it is exactly the kind of claim that reads as sourced because everything around it is.

## Connectivity

Phenotype connectivity went from 9/21 to 16/21, including a new node — **Disrupted Hepatic and Otic Development** — that carries the liver and hearing limbs on the zebrafish route (liver underdevelopment and reduced otic vesicle size from the same orthologue that patterns kidney and pancreas).

Five remain unconnected: macrocephaly, hyperparathyroidism, eye abnormality, congenital heart defect and seizure. None has a sourced route from HNF1B or from any named interval gene, and given that the entry's central claim is that part of this phenotype is *not* HNF1B, inventing edges into the HNF1B branch would undercut the argument it is making.

## Artifacts dead link

Good catch — the `_artifacts/` directory is now committed, so the report's frontmatter links resolve.

Validation after the fixes: `just validate` clean with 69/69 snippets verified; `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms`, `check-enum-values`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets` and `check-reference-titles` all clean.

---

aleixpuigb:
Both fixed in b823cdc32.

**1. Prevalence bands.** Both were wrong; I took your approach and checked against the repo's own bucketing function rather than by eye:

```
25.0  -> BAND_1_5_PER_10000   (was BAND_1_9_PER_100000)
114.3 -> ABOVE_1_IN_1000      (was BAND_1_9_PER_100000)
```

Worth naming the pattern, since this is the second time in two rounds: a legal enum value in the wrong bucket. The frequency band on `Renal cyst` was the same shape. Both are *derivable* — `rate_per_100000` determines `prevalence_class`, and a cohort count determines `FrequencyEnum` — and neither is checked, so a curator can put a number and a band next to each other that contradict, and everything passes. `scripts/migrate_prevalence.py` already contains `_band_from_rate`, so the consistency check for the prevalence half exists as a function and is simply not wired to a gate. I will raise that separately rather than in this PR.

**2. PMID:15649945 snippet.** Correct, and it is the mis-picked adjacent sentence you describe. Re-pointed to the sentence that actually names the genes:

> "In these cases, the expression of PKHD1 (polycystic kidney and hepatic disease 1) and UMOD (Uromodulin), two genes regulated by HNF1beta, was turned off."

The `INDIRECT` grading and the chromophobe-vs-haploinsufficiency caveat are unchanged, since neither was affected.

## On the three suggestions

**The conjoined hepatic/otic node** — your objection is right and I want to be honest that "grouped because both come from the same source" is a bad reason, which is why I wrote it down rather than dressing it up. I have left it as one node this round because splitting it produces two nodes whose only content is a single shared zebrafish sentence each, and the entry already has five phenotypes with no route at all. But it is a conjunction, not a chain, and I would not defend it if a reviewer wanted it split.

**`NO_EVIDENCE` on the `FAILS_TO_RECAPITULATE` link** — I think `NO_EVIDENCE` is right here and would keep it. Your reading ("this paper's scope is the reason we know it fails") makes the paper's silence into positive evidence about the model's limits, and the paper is silent because zebrafish have no autism or schizophrenia phenotype to look for, not because anyone looked and found the model adequate or inadequate. `NO_EVIDENCE` is the schema's value for "the cited reference does not bear on the claim", which is exactly true. The substantive claim — that the model structurally cannot carry the contiguous-gene arm — is in `limitations` and `description`, where it is argued rather than cited.

**`mappings:` with OMIM 614527** — still not possible. `DiseaseMappings` has no OMIM slot and `omim_mappings` fails schema validation. Commented on #7906 with this instance.

`just validate` clean, 69/69 snippets verified, all offline gates clean.

---

aleixpuigb:
This PR was ejected from the merge queue with `MERGE_CONFLICT`. I have rebased it onto current `main` and pushed. **No curated content changed.**

## What the conflict actually was

Two rows in the derived term caches, not entry content. Another PR added `MONDO:0013791` (thrombophilia due to protein S deficiency) at the same sorted position where this PR adds `MONDO:0013797`:

```
cache/mondo/terms.csv
cache/enums/diseaseterm_f618932bb23f.csv
```

Both rows belong. I resolved it by keeping both in canonical sorted order rather than picking a side, which is the same "merge the blocks, do not delete a block" rule `CLAUDE.md` gives for duplicate keys. The resulting diff against `main` is one added line per file:

```
+MONDO:0013797
+MONDO:0013797,chromosome 17q12 deletion syndrome,2026-09-03T15:00:11.626923
```

I did not hand-edit any other cache row. `just check-cache-order` and `just check-term-cache-integrity` both pass, which is the check that matters after resolving a cache conflict by hand.

## The entry is unchanged

Verified rather than assumed. Both files are byte-identical to the approved commit `b823cdc32`:

```
git diff --quiet b823cdc32:kb/disorders/Chromosome_17q12_Deletion_Syndrome.yaml \
                 HEAD:kb/disorders/Chromosome_17q12_Deletion_Syndrome.yaml   -> identical
git diff --quiet b823cdc32:history/.../2026-09-03T153605Z-claude-code-52824b.yaml \
                 HEAD:history/.../2026-09-03T153605Z-claude-code-52824b.yaml -> identical
```

Scope is still 23 files: the entry, its history record, its cited references, the research report and artifacts, the stub deletion, and the two cache rows above.

## Re-validated against the new base

```
just validate                                     -> schema OK, terms OK, 69/69 snippets
check-duplicate-keys / check-entity-refs / check-causal-targets
     / check-qualifier-terms / check-enum-values  -> all OK
check-snippet-length / check-snippet-grading / check-title-snippets
     / check-reference-titles / check-folded-hyphens
     / check-environmental-evidence               -> no new findings
check-cache-order / check-term-cache-integrity    -> OK
```

Working tree clean afterwards, so there is no uncommitted cache churn.

**No new history record.** History records are curation provenance, and no curation happened here: this is a base refresh whose only content change is one row in each of two derived cache files. A record saying "changed nothing" would be noise in the entry's history.

**I did not touch anyone else's PR.** The queue congestion that preceded this ejection involves other authors' PRs, which are theirs to resolve.

The force-push was a rebase on a branch I created, per the "prefer rebase for branches you own" convention. It dismissed the approval, so this needs a re-review before it can re-enter the queue.

---

aleixpuigb:
The re-review triggered by my rebase push [failed to start](https://github.com/monarch-initiative/dismech/actions/runs/33957569494). Flagging it because this PR now needs a review it cannot get on its own, and I cannot re-trigger one.

**It is not a problem with this PR's diff.** Every setup step passed, including `Check the PR diff is enumerable by the review action`. The `Run Claude Code Review` step then failed 11 seconds in (09:20:04 to 09:20:15), and `Verify the review agent actually ran` failed after it. That guard is doing its job: the agent did not run, so no review was produced.

I tried both documented re-run paths and both are refused for this token:

```
POST /actions/runs/33957569494/rerun-failed-jobs                  -> 403 Resource not accessible by integration
POST /actions/workflows/claude-code-review.yml/dispatches         -> 403 Resource not accessible by integration
```

So this needs someone with workflow-dispatch access to re-run `claude-code-review.yml` for PR 10730, or any maintainer review.

**What I am not doing about it.** The review workflow fires on `synchronize`, so a push would start a new one. I am not pushing anything to force that. An empty commit to kick CI is off the table, and manufacturing a content change to trigger a re-review would mean editing curated content for a reason that has nothing to do with the content. The three non-blocking suggestions from the last round are still open on merit, and I would rather leave them open than take one now as a pretext.

State of the PR is otherwise unchanged from my previous comment: rebased onto current `main`, entry and history record byte-identical to the approved commit, 69/69 snippets verified, all offline gates green including `check-cache-order` and `check-term-cache-integrity`.

I will keep checking this PR until it merges or closes.

---

aleixpuigb:
Correcting my previous comment: I said the review failure had no established cause. It does, and I should have pulled the log before saying otherwise.

The reviewer hit a **usage limit**, not an infrastructure fault:

```
```

So the agent died before reading anything, and the fix is simply to wait for the 11:20 UTC reset and re-run. That is a considerably more useful answer than "unknown", and it was one API call away the whole time.

The reason the workflow told me otherwise is a known bug, [#7461](https://github.com/monarch-initiative/dismech/issues/7461): the rate-limit guard matches `usage limit` and `weekly limit`, but the message says `You've hit your limit`, so it falls through to the generic branch that blames a bad `--model`, a bad OAuth token, or an aborted run. I have [added today's instance](https://github.com/monarch-initiative/dismech/issues/7461#issuecomment-5550896751) to that issue rather than opening a new one.

Nothing about this PR changes. It stays as described: rebased onto current `main`, entry and history byte-identical to the approved commit, 69/69 snippets verified, all offline gates green. It needs a review after the quota resets, and I will re-request one then.

The 403s on both re-run paths are unrelated and still stand, so if a maintainer gets there first, re-running `claude-code-review.yml` for this PR after 11:20 UTC is all it needs.

---

Issue #10738, opened 2026-09-03
A vaccination exposure cannot be bound: ECTO:2000129 is a label-less dangling node in the sqlite build, and `environmental-term-audit` recommends it anyway


aleixpuigb:
Curating BCG vaccination as an `environmental:` exposure, I could not bind an `exposure_term`, and the way that failure presents is worth fixing separately from the missing-term problem itself.

## What happens

`ECTO:2000129` "exposure to vaccination" is real. OLS serves it with a definition:

```
$ curl -s "https://www.ebi.ac.uk/ols4/api/search?q=exposure%20to%20vaccination&ontology=ecto&rows=10"
ECTO:2000129 | exposure to vaccination     ("An exposure to vaccination.", synonym "vaccination exposure")
ECTO:3000274 | exposure to Orthopoxvirus vaccinia
```

It is exactly the right term for a vaccination exposure. Binding it fails:

```
❌ ERROR: Value 'ECTO:2000129' not in dynamic enum 'ExposureTerm' (expanded from ontology)
❌ ERROR: Term 'ECTO:2000129' not found in ontology
```

Because in the build the term validator actually reads, the node exists but has no label:

```
$ uv run runoak -i sqlite:obo:ecto info ECTO:2000129
ECTO:2000129 ! None
```

That is a dangling reference — declared, not defined. And `ExposureTerm` is `reachable_from` `ExO:0000002` / `XCO:0000000`, which a label-less orphan is not under, so both errors are the same underlying fact reported twice.

## The part that is a dismech bug rather than an upstream one

`just environmental-term-audit` **recommends this term**. Its reuse-candidate column prints:

```
[BOUND  ] X-linked_Mendelian_Susceptibility_..._CYBB_Deficiency | BCG vaccination  -> reuse ECTO:2000129 (exposure to vaccination)
```

So the audit tells a curator to bind a CURIE the validator then rejects. `CLAUDE.md` is already careful that a reuse suggestion is advisory on *meaning* — "it matches curator-written names, not meanings" — but this is a different failure: the suggestion is wrong on *validity*, which the audit could check cheaply since it is reading the caches anyway. A suggestion that cannot pass `just validate-terms` should not be offered, or should be labelled as unvalidatable.

That matters because the audit exists to drive a binding tranche (#8430). If it surfaces dangling CURIEs, a tranche worked from its output goes red at validation time, one entry at a time, for a reason that has nothing to do with the curator's judgement.

## What is genuinely missing upstream

[...]

PR #10740, opened 2026-09-03
curate(MSMD-CYBB): add X-linked MSMD due to CYBB deficiency

### LINKED ISSUE #10713: Curate X-linked Mendelian susceptibility to mycobacterial diseases due to CYBB deficiency (MONDO:0010389)
Curate a dismech entry for **X-linked Mendelian susceptibility to mycobacterial diseases due to CYBB deficiency** ([MONDO:0010389](https://monarchinitiative.org/MONDO:0010389)) — hypomorphic CYBB missense variants that leave the neutrophil and monocyte respiratory burst intact but abolish it selectively in macrophages and EBV-transformed B cells, so patients get BCG and environmental mycobacterial disease without the broader infection spectrum of X-linked chronic granulomatous disease.

There is precedent for per-genotype MSMD entries at this granularity (`kb/disorders/Mendelian_Susceptibility_To_Mycobacterial_Diseases_Due_To_Complete_IL12B_Deficiency.yaml`). The lump/split question is against `kb/disorders/Chronic_Granulomatous_Disease.yaml`, which already covers X-linked CYBB CGD as a subtype: the case for a separate entry is that the cell-type-restricted burst defect is a different pathograph from the global one, not a milder degree of it.

Stub: `stubs/X-linked_Mendelian_Susceptibility_To_Mycobacterial_Diseases_Due_To_CYBB_Deficiency.yaml`

aleixpuigb:
Adds `kb/disorders/X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency.yaml` for [MONDO:0010389](https://monarchinitiative.org/MONDO:0010389).

Closes #10713

## The lump/split call

`entry_type: DISEASE`. `kb/disorders/Chronic_Granulomatous_Disease.yaml` already carries `X-linked CYBB` as a `has_subtypes` entry, so a reviewer should expect me to justify a second CYBB entry.

The justification is that this is not a milder CGD, it is a narrower one, and the difference is cell type rather than degree. The Q231P and T178P alleles abolish the respiratory burst in monocyte-derived macrophages and leave it intact in blood neutrophils and monocytes — verified across cytochrome c reduction, dihydrorhodamine flow cytometry, nitroblue tetrazolium and chemiluminescence, plus a functional *S. aureus* killing assay. Affected males get BCG and tuberculous disease and nothing else; no staphylococcal disease, even in affected adults of advanced age. PMID:21278736 is explicit that these two alleles had never been associated with CGD.

There is precedent for per-genotype MSMD entries at this granularity (`Mendelian_Susceptibility_To_Mycobacterial_Diseases_Due_To_Complete_IL12B_Deficiency`).

Two pathophysiology nodes exist to carry that argument rather than leave it in prose: **Preserved Neutrophil and Monocyte Respiratory Burst** and **Absence of the Chronic Granulomatous Disease Infection Spectrum**. Both are mechanistic negatives with their own evidence.

## One nosological point

MSMD is defined as a set of inborn errors of interferon-gamma immunity, and 61% of 830 reported MSMD patients have IL12RB1 or IFNGR1 defects. This disorder is not one of those — the lesion is at the effector end, in the oxidase the interferon-gamma-primed macrophage uses. That is why it sits inside MSMD as a mechanistic exception, and why interferon-gamma is a plausible therapy here despite the signalling pathway being intact. Curated in the description and on the assembly node.

## Three deliberate non-SUPPORT gradings

Worth calling out so they do not read as mistakes:

- **`NO_EVIDENCE`** on the EBV-transformed B cell node. The authors state the B cell phenotype is irrelevant to pathogenesis, since B-cell-deficient patients are not prone to mycobacterial disease, and that they used it as an accessible surrogate. The node carries no downstream edges for the same reason. It is curated because it is the system in which the alleles were shown to be hypomorphic, not because it causes anything.
- **`REFUTE`** on interferon-gamma therapy. Interferon-gamma abolished the BCG growth-control deficit, so it is on the `target_mechanisms` link as `SUPPORT`. But Q231P macrophages still failed to mount the burst under interferon-gamma, so the entry-level evidence is graded `REFUTE` against the stronger claim that interferon-gamma corrects the oxidase defect. The benefit cannot be attributed to burst restoration.
- **`directness: INDIRECT`** on all three additions from the 830-patient MSMD review. Lymphadenopathy at 45.5%, fever at 30.2%, and the genotype distribution are MSMD-wide figures across 21 genotypes. With seven reported CYBB patients there is no genotype-specific frequency to give, and each `explanation` says so rather than letting a cohort number look like a CYBB number.

## The load-bearing limitation

A `HUMAN_MODEL_MISMATCH` discussion, not a `KNOWLEDGE_GAP`, because the evidence exists and it is the translational validity that is open. Every measurement in the defining paper is in macrophages differentiated from blood monocytes over 14-15 days; the authors state plainly they could not test the patients' macrophages in vivo or ex vivo. Since the defining feature of the disorder is that one allele behaves differently in different cell states, a model whose fidelity turns on getting the cell state right is the wrong place for an untested assumption. A proposed experiment specifies the ex vivo tissue-macrophage assay with an internal neutrophil control, and both supporting and refuting outcomes.

## Deep research

One openscientist run, committed under `research/`. Like the 17q12 run in the same batch, **it shipped with no reference-validation or term-validation block**, because term validation aborted on a 5-second EBI OLS timeout on `HP:0002716` — a CURIE that is in `cache/hp/terms.csv`. That is dismech#10396, [commented on with the reproduction](https://github.com/monarch-initiative/dismech/issues/10396#issuecomment-5527972608). I retro-fitted the reference section; it flagged PMID:1531037 as resolving but off topic (a 1992 paper on signal transduction regulating NADPH oxidase component gene expression in polymorphonuclear leukocytes), and it was not used.

Taken from the report: PMID:38341181 (the 830-patient systematic review) and PMID:30264912 (MSMD 2014-2018 update, for the interferon-gamma-immunity definition this disorder is the exception to).

## A tooling issue found and filed

The BCG exposure ships with **no `exposure_term`**, and the reason turned out to be a bug rather than a missing term — filed as #10738.

I first wrote "no ECTO term exists", then checked. `ECTO:2000129` "exposure to vaccination" does exist and is exactly right, but it is a label-less dangling node in the `sqlite:obo:ecto` build the validator reads (`runoak` returns `ECTO:2000129 ! None`) and is not reachable from the `ExposureTerm` enum roots, so binding it fails. Separately, `just environmental-term-audit` **recommends** `ECTO:2000129` as a reuse candidate, so the audit points curators at a CURIE the validator rejects — which matters for anyone working a binding tranche from its output.

The search is recorded in the exposure's `review_notes` per the #8296 convention. The entry has its own evidence, so that note is documentation rather than a waiver, and `just check-environmental-evidence` passes.

## Validation

```
just validate  -> schema OK, terms OK, 36/36 snippets verified
     / check-qualifier-terms / check-enum-values                -> all OK
     / check-title-snippets / check-environmental-evidence      -> no new findings
just validate-history  -> OK
```

## Not done

No `datasets:` block, and no `animal_models:`. The report cites PMID:36936954 (DUOX1 dispensable during *M. tuberculosis* infection in mice), but that is a different oxidase in a model that does not carry the human alleles, so it would be a model of the wrong lesion.

---

### DIFF
--- a/history/disorders/X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency/2026-09-03T181055Z-claude-code-1104da.yaml
+++ b/history/disorders/X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency/2026-09-03T181055Z-claude-code-1104da.yaml
@@ -1,0 +1,40 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency
+  path: kb/disorders/X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency.yaml
+session:
+  id: 2026-09-03T181055Z-claude-code-1104da
+  timestamp: '2026-09-03T18:10:55Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10740
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  - phenotypes
+  - evidence
+  summary: 'Review round 1: treatment coverage, tuberculosis term, evidence alignment'
+  details: |
+    Responded to the first automated review on PR #10740 (REQUEST_CHANGES, four IMPORTANT items, no critical defects).
+
+    1. Tuberculosis phenotype rebound from HP:0011274 (Recurrent mycobacterial infections) to HP:5210111 (Tuberculosis infection). HP:0011274 was carried by two distinct phenotypes, so the machine-readable layer could not tell recurrent mycobacterial disease apart from tuberculosis in an unvaccinated patient, which is the whole point of that phenotype. HP:0011274 now appears exactly once.
+
+    2. Added an Antimycobacterial Therapy treatment (NCIT:C15986 Pharmacotherapy; therapeutic_agent NCIT:C280 Antitubercular Agent, CHEBI:28077 rifampicin, CHEBI:6030 isoniazide; therapeutic_modality SMALL_MOLECULE), targeting Failure to Restrict Intramacrophage Mycobacterial Growth with treatment_effect INHIBITS. The entry previously curated only IFN-gamma and BCG avoidance, omitting the mainstay of MSMD management. No regimen or duration is asserted. Two PMID:38341181 quotes, both graded directness INDIRECT because the review is MSMD-wide across 21 genotypes and reports no CYBB-specific outcome.
+
+    3. Added Hematopoietic Stem Cell Transplantation (NCIT:C15431, CELL_THERAPY, RESTORES on Cell-Type-Restricted Failure of NADPH Oxidase Assembly). Curated as an option with an explicit caveat rather than a recommendation: no CYBB-MSMD transplant outcome is reported in any source curated here, and the risk-benefit calculation is not obviously that of a global phagocyte defect, since these patients have normal neutrophil function and a mycobacteria-restricted phenotype.
+
+    4. IFN-gamma therapy gained therapeutic_agent NCIT:C100089 (Interferon Gamma-1b) and treatment_effect RESTORES on its existing mechanism link.
+
+    5. BCG avoidance gained the PMID:38341181 contraindication sentence. The two pre-existing quotes established the premise (BCG disease as presenting illness) but never stated the recommendation.
+
+    Validation: schema OK, term validation passed, 42/42 snippets verified (was 38/38). Offline gates all green: duplicate-keys, entity-refs, causal-targets, qualifier-terms, enum-values, snippet-length, snippet-grading, title-snippets, reference-titles, folded-hyphens, environmental-evidence. All four reviewer-supplied quotes were verified verbatim against references_cache/PMID_38341181.md before use, and every claim in this record was grepped out of the committed file rather than taken from the edit script's own success output.
--- a/kb/disorders/X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency.yaml
+++ b/kb/disorders/X-linked_Mendelian_Susceptibility_to_Mycobacterial_Diseases_due_to_CYBB_Deficiency.yaml
@@ -473,10 +473,10 @@ phenotypes:
     patient. This extends the phenotype beyond MSMD sensu stricto and is what makes the
     disorder relevant to tuberculosis susceptibility genetics generally.
   phenotype_term:
-    preferred_term: Recurrent mycobacterial infections
+    preferred_term: Tuberculosis
     term:
-      id: HP:0011274
-      label: Recurrent mycobacterial infections
+      id: HP:5210111
+      label: Tuberculosis infection
   evidence:
   - reference: PMID:21278736
     reference_title: "Germline CYBB mutations that selectively affect macrophages in kindreds with X-linked predisposition to tuberculous mycobacterial disease."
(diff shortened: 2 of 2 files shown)


aleixpuigb:
Flagging a CI oddity and what I did about it.

**The `claude-review` run on this PR was cancelled, not completed** ([run 33775134894](https://github.com/monarch-initiative/dismech/actions/runs/33775134894), started 15:51, cancelled 16:16), so no review arrived. `test (3.13)` passed. I could not re-trigger it — both `workflow_dispatch` on `claude-code-review.yml` and a rerun of the cancelled run return `403 Resource not accessible by integration` for this token.

The four sibling PRs from the same batch (#10723, #10730, #10743, #10744) all got reviews within a few minutes of each other, and this one was cancelled in the middle of that window. It looks like cross-PR contention rather than anything about this diff, though I have not confirmed the cause.

Rather than sit on it, I ran a self-review against what the other four reviews found and pushed 4633a6655, which also re-triggers the workflow on `synchronize`:

1. **`reference_title` on all 38 evidence items.** All four reviews flagged this as blocking, and this entry had zero. Worth noting for anyone else hitting it: `just normalize-reference-titles` only re-syncs titles that are already present and reports "0 re-synced" when there are none to sync; `just backfill-reference-titles` is the one that adds them.

2. **Pathograph connectivity.** `Lymphadenopathy` and `Fever` were orphans; both now hang off the mycobacterial growth-control node with `INDIRECT_UNKNOWN_INTERMEDIATES`. The EBV-transformed B cell node had no incoming edge and now has one from the oxidase assembly failure, which is its actual cause. It still has no *downstream* edges, and that is deliberate — the authors state the B cell phenotype is irrelevant to pathogenesis, and its evidence item is graded `NO_EVIDENCE` for that reason.

3. **`diagnosis:` section added.** CYBB sequencing after a normal CGD screen, which is a distinct modality from the macrophage burst assay. The burst assay stays as a `DIAGNOSTIC_CRITERIA` definition rather than moving, because the `Definition` class can carry `validation_status` (recorded here as `UNVALIDATED`) and the `Diagnosis` class cannot — that seemed the more informative home for a test derived from seven patients with no accuracy study behind it.

`just validate` clean with 38/38 snippets verified; all eight offline gates clean.

Happy to take a full review whenever the workflow runs.

---

aleixpuigb:
Thanks, all four addressed in dcbc5177b. Details below, including one place where I curated something weaker than the review suggested.

**Treatment coverage was the real gap.** The entry had IFN-gamma and BCG avoidance but not the drugs that actually treat the infection, which is not a defensible set of treatments for a mycobacterial susceptibility disorder. Added `Antimycobacterial Therapy` (NCIT:C15986, agents NCIT:C280 Antitubercular Agent, CHEBI:28077 rifampicin, CHEBI:6030 isoniazide, `SMALL_MOLECULE`), `INHIBITS` on "Failure to Restrict Intramacrophage Mycobacterial Growth". I deliberately assert no regimen or duration; composition follows species identification and susceptibility testing, and nothing I have cited fixes one. I did keep the "start three or four drugs before the genetic defect is identified" sentence, because for a disorder reached only by sequencing that ordering is the clinically load-bearing part.

**HSCT is in, but curated as an option rather than a recommendation.** Added `Hematopoietic Stem Cell Transplantation` (NCIT:C15431, `CELL_THERAPY`, `RESTORES` on "Cell-Type-Restricted Failure of NADPH Oxidase Assembly"). The description carries an explicit caveat: no CYBB-MSMD patient's transplant outcome is reported in any source curated here, and the risk-benefit calculation is not obviously the one for a global phagocyte defect, since these patients have normal neutrophil function and a phenotype restricted to mycobacteria. "The only curative treatment in MSMD" is true of the syndrome; extending it to this genotype is an inference, and I would rather the entry say so than imply a recommendation the literature has not made here.

**Tuberculosis term.** Rebound from `HP:0011274` (Recurrent mycobacterial infections) to `HP:5210111` (Tuberculosis infection). You were right that this mattered more than it looks: `HP:0011274` was on two different phenotypes, so nothing downstream of the YAML could distinguish recurrent mycobacterial disease from tuberculosis in an unvaccinated patient, and that unvaccinated patient is the entire reason the phenotype is there. `HP:0011274` now appears exactly once in the file.

**IFN-gamma and BCG.** IFN-gamma gained `therapeutic_agent` NCIT:C100089 (Interferon Gamma-1b) and `treatment_effect: RESTORES`. BCG avoidance gained the contraindication sentence itself; the two quotes already there established the premise (BCG disease as the presenting illness) but never stated the recommendation, so the treatment was resting on an inference the reader had to make.

**All five new evidence items are `directness: INDIRECT`.** PMID:38341181 is an MSMD-wide systematic review across 21 genotypes and reports no CYBB-specific treatment outcome. Every treatment claim in this entry is therefore syndrome-level rather than genotype-level, and I would rather that be visible in the grading than smoothed over.

**Process note.** I verified all four quotes you supplied verbatim against `references_cache/PMID_38341181.md` before using them (each `grep -c` returned 1), and I grepped every claim above out of the committed file rather than trusting my edit script's own success output. That second habit exists because on an earlier PR in this run three of seven edits silently no-op'd and I wrote a commit message claiming all seven.

**Validation.** Schema OK, term validation passed, 42/42 snippets verified (was 38/38). Offline gates green: duplicate-keys, entity-refs, causal-targets, qualifier-terms, enum-values, snippet-length, snippet-grading, title-snippets, reference-titles, folded-hyphens, environmental-evidence.

---

PR #10748, opened 2026-09-03
Curate HOD syndrome (CLCN7 gain-of-function, MONDO:0032805)

### LINKED ISSUE #10745: Curate hypopigmentation, organomegaly, and delayed myelination and development (MONDO:0032805)
Curate a dismech entry for **hypopigmentation, organomegaly, and delayed myelination and development** ([MONDO:0032805](https://monarchinitiative.org/MONDO:0032805), "HOD syndrome") — a multisystem disorder attributed to a *de novo* variant in `CLCN7`, the lysosomal chloride/proton antiporter whose **loss** of function causes osteopetrosis.

**Why this one is worth having.** The knowledge base already carries three CLCN7 entries — `Osteopetrosis`, `Autosomal_Dominant_Osteopetrosis_Type_II`, `Autosomal_Recessive_Osteopetrosis` — and all of them are built on losing CLCN7 activity. HOD is reported as the opposite direction on the same gene, and the phenotype is not a milder or partial osteopetrosis; it is a different disease affecting pigment, viscera and myelination. If that holds up in the sources, it is a clean worked case for the distinction `CLAUDE.md` draws between `GeneticContext.functional_impact_category` (what a variant does) and `Descriptor.modifier` (what a pathway is doing), which currently has no entry demonstrating a gain-of-function variant alongside the existing loss-of-function ones.

I will treat the direction claim as something to verify from primary sources rather than assume, and record it as prose if the evidence does not support binding it.

The stub records no MONDO descendants and one causal gene (`hgnc:2025` CLCN7), so this looks like a leaf concept rather than a grouping. The open question is instead whether it is a distinct disease or belongs as a `has_subtypes` entry on an existing CLCN7 entry — I think distinct, because the organs involved do not overlap, but that is the call to make first.

Stub: `stubs/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development.yaml`

---

### Two things found while picking, for whoever sweeps the queue

**A near-miss on my own duplicate check.** `spondyloepiphyseal dysplasia, Nishimura type` (`MONDO:0032835`, MIR140) matched three files on `origin/main`. All three matches are the author name inside a citation for **Gillessen-Kaesbach-Nishimura skeletal dysplasia**, which is ALG9-CDG and a different disease. Searching the label alone would have read that as "already curated" and skipped a real gap. The MONDO ID search is what separated them.

**Two diseases are each claimed twice.** `MONDO:0011244` Marshall-Smith syndrome is claimed by both #10514 and #10516; `MONDO:0008787` microcytic anemia with liver iron overload by both #10515 and #10517. Each pair was filed within about a minute, so this looks like one run filing twice rather than two people colliding. Worth closing one of each.

Phillip-a-richmond:
Adds `kb/disorders/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development.yaml` — HOD syndrome, an ultra-rare multisystem lysosomal disorder caused by de novo gain-of-function variants in `CLCN7`.

Closes #10745

> **Current at head `9cbb2e404`, approved after three review rounds.** Counts below describe this head. What the rounds changed is summarised at the bottom.

## The same gene, in the other direction

The knowledge base already has three CLCN7 entries — `Osteopetrosis`, `Autosomal_Dominant_Osteopetrosis_Type_II`, `Autosomal_Recessive_Osteopetrosis` — and all three are built on **losing** ClC-7 activity. HOD is allelic with all of them and shares none of their bone disease. Every reported patient has been explicitly free of osteopetrosis, and one had a structurally normal skeleton on X-ray at six years despite delayed bone age.

Bone here is not a mildly affected organ. It is an unaffected one, and the primary source says why: *"Osteopetrosis results from a loss of ClC-7, but osteoclasts remain resilient to increased ClC-7 activity."*

The pathophysiology chain — 13 nodes, one root, no orphans:

```
CLCN7 missense variant → loss of PI(3,5)P2 inhibition → increased Cl-/H+ exchange
  ├→ lysosomal hyperacidification ─┬→ impaired autophagic degradation ─┬→ foam cells ─┬→ organomegaly
  │                                │                                   │              └→ protein-losing enteropathy → failure to thrive
  │                                │                                   ├→ delayed myelination
  │                                │                                   └→ global developmental delay
  │                                └→ reduced melanin synthesis → hypopigmentation
  └→ enlarged lysosome-related vacuoles ──┘
```

## Four allelic outcomes, and they are not a severity gradient

Curating this forced a correction to a claim I had made in the claim issue. I wrote that heterozygous CLCN7 loss causes Albers-Schönberg ADO II. It does not. ADO II is **dominant-negative**, and plain heterozygous loss of function causes no abnormality at all:

> it seems that ADO II reflects a dominant negative effect, whereas loss-of-function mutations in ClCN7 do not cause abnormalities in heterozygous individuals

So CLCN7 haploinsufficiency is silent. ADO II is not "half of" recessive osteopetrosis, and HOD is not the far end of a dose axis — it is a fourth, separate outcome.

The tidy version is not sufficient either, and the entry says so rather than flattening it. Some osteopetrosis-causing CLCN7 variants *also* accelerate ClC-7 activation, so direction of functional change alone does not predict which disease you get. That is curated as a knowledge gap with the contradicting quote cited as `REFUTE`.

## The weakest link is labelled as the weakest link

Hypopigmentation is one of the four features the syndrome is named for, and the route to it is the least established thing here. Melanosomes are lysosome-related organelles and tyrosinase is inhibited at acidic pH, so a transporter that over-acidifies lysosomes is *expected* to over-acidify melanosomes and suppress pigment.

None of that has been measured in this disease — no melanosomal pH, no tyrosinase activity, no melanosome ultrastructure, in any patient or in the knock-in mouse. That node's evidence is general melanocyte biology, carries `directness: INDIRECT`, and has a knowledge gap with a proposed experiment attached.

The gap also carries a `REFUTE` item that makes it sharper. Mice expressing a **transport-dead** ClC-7 have *normal coat colour*: pigmentation survives the loss of ion transport while still depending on the protein being present. So whatever excess transport does to the melanosome in HOD, it is not simply the mirror of losing transport.

## Four wrong identifiers in the research report, none of which term validation would catch

The OpenScientist report is good — 19/19 references resolved, zero confabulated, and it reproduced this entry's causal chain without having seen it. Its identifiers are another matter. The first three are real CURIEs that resolve cleanly:

| Report offered | Actually is |
|---|---|
| `MONDO:0032781` as **this disease's own Mondo ID** | congenital hypotonia, epilepsy, developmental delay, and digital anomalies — a different syndrome |
| `HGNC:2023` for CLCN7 | **CLCN5** — same CLC family, different gene, Dent disease |
| `HP:0007730` labelled "Ocular albinism" | label is *Iris hypopigmentation* |
| a second Nicoli variant `p.Tyr99Cys` | absent from `PMID:31155284` entirely — found by the reviewer, not by me |

The first is the entry's identity. Copying it would have bound the whole record to the wrong disease, and `just validate-terms` would have passed. The report hedged that one — *"to be confirmed against current Mondo release"* — which is the honest flag and the part worth reading.

Every identifier used here was derived from the curation stub and resolved against local OAK builds instead.

## Validation at `9cbb2e404`

| Check | Result |
|---|---|
| `just validate-disorders` | passed, **67/67** snippets verified |
| `check-snippet-length` / `check-title-snippets` / `check-folded-hyphens` | clean |
| `check-snippet-grading` / `check-enum-values` / `check-entity-refs` | clean |
| `check-causal-targets` / `check-duplicate-keys` / `check-qualifier-terms` | clean |
| reference titles vs cache | **67/67** exact, audited by script |
| pathophysiology subgraph | 13 nodes, 1 root, 0 orphans, 0 unresolved targets |

`cache/**.csv` is six additive rows, zero deletions, no reordering.

## What the three review rounds changed

**Round 1 requested changes on four items, all correct.** A snippet that began mid-word at `"ented with"`; a missing `HP:0001107 Ocular albinism` phenotype; a missing IVIG treatment; and `evidence_source: HUMAN_CLINICAL` on a Xenopus oocyte electrophysiology result. Four non-blocking suggestions came with them. All eight were taken.

Two are worth more than a line:

- **The ocular albinism finding was two errors.** Missing the binding was the smaller half. The larger was that my `notes` generalized from one patient's spared irides to "an iris term should not be bound here at all" — a claim about the disease drawn from a single case, when a different patient in a different paper has ocular albinism documented from birth. The note now rejects only `HP:0007730` specifically, and the phenotype carries the normal-iris quote as `REFUTE`.
- **`functional_impact_category` was genuinely absent, and the reason matters.** An earlier edit of mine used a string replace whose anchor did not match, with no assertion guarding it. It silently did nothing, `just validate` passed on the unchanged file, and I reported having made the change. Every edit since asserts its anchor first.

**Round 2 approved, and left four more suggestions.** I took two and declined two:

- **Suggestion A was a regression I had introduced**, not a suggestion. Splitting the bundled myelination/development node in round 1 removed the only incoming edge to `Global Developmental Delay`, so an outcome the node description still claimed followed the lysosomal lesion had no representation in the graph. Restored with the same `INDIRECT_UNKNOWN_INTERMEDIATES` type as the myelination edge — and deliberately *not* routed through `Delayed Myelination`, since no source establishes that link.
- **Declined B**, connecting the twelve unconnected phenotype nodes: the only available fix is renaming terminal pathophysiology nodes to collide with phenotype names, which is the flat-namespace collapse recorded as a graph-model bug in #9896. It belongs upstream, not in one entry. The reviewer agreed.

**Round 3 approved with no new findings.**

## Two environment notes

**Report term validation could not be run.** `just validate-research-terms` failed on four attempts with OLS read timeouts at the validator's five-second limit, so the committed report has a `## Reference Validation` section but no `## Term Validation` section. Rather than skip the check, I extracted all 45 CURIEs from the report and resolved them against local OAK builds — which is how three of the four errors above were found.

**The report cites by bare PubMed number**, in a `## Key References (PMID)` block, not as `PMID:` identifiers, and no `.citations.md` sidecar was produced. The identifier sweep the curation skill suggests (`grep -o "PMID:[0-9]*"`) returns **zero** on this report. It is not an uncited report; it just cannot be swept that way.

## Frequencies

Every band was derived by division over the five published patients rather than assigned by feel. Hypopigmentation, organomegaly and developmental delay are 5/5 → `OBLIGATE`; delayed myelination 4/5 → `VERY_FREQUENT`; hypotonia, failure to thrive, vacuolated lymphocytes and nephromegaly 2/5 → `FREQUENT`; the rest 1/5 → `OCCASIONAL`.

The `OBLIGATE` bands carry a caveat in their descriptions: the syndrome is *named* for those features, so 100% is partly definitional rather than a measured penetrance. That is recorded rather than presented as a finding.

### DIFF
--- a/history/disorders/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development/2026-09-03T175324Z-claude-code-7ceba8.yaml
+++ b/history/disorders/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development/2026-09-03T175324Z-claude-code-7ceba8.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development
+  path: kb/disorders/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development.yaml
+session:
+  id: 2026-09-03T175324Z-claude-code-7ceba8
+  timestamp: '2026-09-03T17:53:24Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10748
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - evidence
+  summary: 'Address review round 2 suggestions A and C on PR #10748'
+  details: |
+    Round 2 approved the entry and left four non-blocking suggestions. Took two. Suggestion A was a regression introduced by my own round-1 node split: separating delayed myelination from the developmental phenotype removed the only incoming edge to Global Developmental Delay, so an outcome the node description still claimed followed the lysosomal lesion had no pathograph representation. Restored as a second downstream target of Impaired Lysosomal and Autophagic Degradation with INDIRECT_UNKNOWN_INTERMEDIATES, matching the myelination edge beside it, and deliberately not routed through Delayed Myelination since no source establishes that link. Suggestion C replaced a trailing noun-phrase snippet on the Failure to Thrive node with the full sentence already cited on the sibling enteropathy node; the fragment cleared check-snippet-length at eight words but asserted no proposition and left the causal reading resting on word order. Declined suggestion B, the twelve phenotype nodes unconnected in the whole-entry graph: connecting them would require renaming terminal pathophysiology nodes to match phenotype names, which is exactly the flat-namespace collapse recorded as a graph-model bug in issue 9896, so the fix belongs upstream rather than in this entry. Declined suggestion D as the reviewer noted it does not bite against a specific action term. Pathograph now 13 nodes and 14 edges. 67/67 snippets verified, 67/67 reference titles machine-audited, all gates green.
--- a/kb/disorders/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development.yaml
+++ b/kb/disorders/Hypopigmentation_Organomegaly_And_Delayed_Myelination_And_Development.yaml
@@ -277,6 +277,14 @@ pathophysiology:
     causal_link_type: DIRECT
   - target: Delayed Myelination
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+  - target: Global Developmental Delay
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+    description: >-
+      Restores the link that the myelination/development split dropped. Held to the same
+      standard as the myelination edge beside it - both outcomes follow the lysosomal lesion,
+      and the intermediates are unknown for each. This is deliberately not routed through
+      Delayed Myelination, because no source establishes that the myelination delay is what
+      causes the developmental delay.
   evidence:
   - reference: PMID:38136669
     reference_title: "Impaired Autophagic Clearance with a Gain-of-Function Variant of the Lysosomal Cl(-)/H(+) Exchanger ClC-7."
@@ -419,8 +427,12 @@ pathophysiology:
     reference_title: "Gain-of-function variants in CLCN7 cause hypopigmentation and lysosomal storage disease."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
-    snippet: "malabsorption, loss of protein and electrolytes, and failure to thrive"
-    explanation: Names the losses and the growth failure together, in that order.
+    snippet: "The clinical picture is dominated by progressive thickening of bowel walls, resulting in chronic impairment of intestinal transit, an extremely distended abdomen, malabsorption, loss of protein and electrolytes, and failure to thrive"
+    explanation: >-
+      The same sentence cited on the enteropathy node, quoted whole here too. The earlier form
+      was the trailing noun phrase alone, which asserted nothing and left the causal reading
+      resting on word order; this version carries the claim that the growth failure follows
+      the enteric losses.
 - name: Delayed Myelination
   biological_scale: TISSUE
   description: >-


Phillip-a-richmond:
All eight findings addressed in `e2a6080f2` — the four blocking ones and the four marked non-blocking. Every quote came from the committed cache, as you said it would.

| Finding | What changed |
|---|---|
| 1 — mid-word snippet | Now quotes the whole sentence, `"Nicoli et al. (2019) previously reported two similar cases…"` |
| 2 — ocular albinism | `HP:0001107` added, 1/5 → `OCCASIONAL`, plus a `REFUTE` item for the normal-iris patient |
| 3 — IVIG | Added, `NCIT:C62710`, `therapeutic_modality: PROTEIN_REPLACEMENT`, linked to the hypogammaglobulinemia |
| 4 — mis-graded source | Split into two items: genetics stays `HUMAN_CLINICAL`, Xenopus electrophysiology becomes `IN_VITRO` |
| 5 — nutritional support | Added, `NCIT:C15433`, linked to the enteropathy node |
| 6 — nephromegaly | Re-banded `OCCASIONAL` → `FREQUENT` on the second case, now cited |
| 7 — `functional_impact_category` | Added |
| 8 — node bundling | Both nodes split |

Four of these are worth more than a table row.

**Finding 7 was worse than you diagnosed, and the diagnosis is the useful part.** You reported the slot as unset. I initially went to reply that it was already set on `GeneticContext` and you had looked in the wrong place — then checked, and it was absent from every commit on this branch. An earlier edit of mine used a string replace whose anchor did not match, with no assertion guarding it, so it silently did nothing and `just validate` passed on a file it had not changed. I then reported having added it. So the slot was missing, my memory of adding it was wrong, and the check I ran could not tell the difference. Every subsequent edit in this round asserts its anchor before writing.

**Finding 2 was two errors, not one.** Binding `HP:0001107` is the smaller half. The larger half is that my `notes` generalized from one patient's normal irides to "an iris term should not be bound here at all", which is a claim about the disease drawn from a single case — and a different patient in a different paper has ocular albinism documented from birth. The note now rejects only `HP:0007730` specifically, and the phenotype carries the normal-iris quote as `REFUTE` so the variability is visible rather than resolved by omission.

**Finding 8, on the split.** I did not draw an edge between delayed myelination and the developmental phenotype. It is a reasonable guess that the myelination delay drives the developmental delay, but no source in the cache says so, and an unsourced edge in a pathograph is worse than two honest leaves. Protein-losing enteropathy → failure to thrive *is* drawn, because the source attributes the growth failure to the enteric losses. Graph is now 13 nodes, 13 edges, one root, four leaves, no orphans.

**Finding 4 — I took the split rather than the relabel.** Changing `HUMAN_CLINICAL` to `IN_VITRO` on the existing item would have left one evidence item covering both the patient sequencing and the oocyte recording. The sentence I was quoting only supports the electrophysiology, so the genetics claim needed its own quote.

Validation at `e2a6080f2`: 67/67 snippets verified, 67/67 reference titles machine-audited against cache frontmatter, `validate` / `validate-terms` / `check-enum-values` / `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` / `check-qualifier-terms` / `check-snippet-grading` all clean, and the three whole-KB checks I had skipped before the first CI failure — `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens` — now run clean too. No `cache/**` change in this push.

Thanks for catching the `p.Tyr99Cys` non-existence independently. That is a fourth report error on top of the three in the PR body, and it is the same shape as the other three: plausible, specific, and absent from the source it is attributed to.

---

Phillip-a-richmond:
Thanks for the approval. I took two of the four remaining suggestions in `9cbb2e404` and am declining the other two with reasons.

**A — taken. It was a regression I introduced, not a suggestion.** You were right that `Global Developmental Delay` lost its incoming edge, and the cause is my own round-1 split: the bundled node had one, and after separating it I kept the edge on `Delayed Myelination` and simply dropped the other half. So the entry was left asserting in prose that both outcomes follow the lysosomal lesion while the graph recorded only one of them. Restored as a second downstream target of `Impaired Lysosomal and Autophagic Degradation` with the same `INDIRECT_UNKNOWN_INTERMEDIATES` type as the myelination edge beside it.

I did **not** route it through `Delayed Myelination`, which would have been the tidier-looking graph. That would assert the myelination delay causes the developmental delay, and nothing in the cache says so — the same reason I declined to draw that edge when splitting the node.

**C — taken.** The fragment now quotes the same full sentence already cited on the sibling enteropathy node. You put your finger on why the eight-word form was bad despite clearing the gate: the `explanation` was doing the work the quote should do, via "in that order". A citation that needs its own explanation to supply the proposition is not carrying the claim.

**B — declining, with a reason I'd want on the record.** You are right that twelve phenotype nodes sit unconnected and that this is the largest structural gap here. But the available fix is to rename terminal pathophysiology nodes to match phenotype names — `Hypopigmentation of Skin and Hair` → `Hypopigmentation`, `Organomegaly with Preserved Organ Function` → `Hepatosplenomegaly` — and that is precisely the flat-namespace collapse recorded as a graph-model bug in #9896, where one name spanning both sections becomes a single self-referential node. Doing it would make this entry's graph look better while adding twelve instances of a known defect. Your own observation that the three currently-connected phenotypes are connected *only* by name-sharing is the same point: that is the bug operating, not curation. It belongs upstream.

**D — declining.** Agreed it does not bite: `NCIT:C62710 Immunoglobulin Therapy` already names the agent class, so a `therapeutic_agent` would restate the treatment term rather than add an identifier.

Validation at `9cbb2e404`: 67/67 snippets verified, 67/67 reference titles machine-audited, pathograph 13 nodes / 14 edges / 1 root / 0 orphans / 0 unresolved targets, and `validate-disorders`, `validate-terms`, `check-snippet-length`, `check-title-snippets`, `check-folded-hyphens`, `check-snippet-grading`, `check-enum-values`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `validate-history` all clean. No `cache/**` change in this push.

I recognise this push drops the approval you just gave. Suggestion A was a live regression rather than polish, and shipping a known one to preserve an approval seemed the wrong trade.

---

Phillip-a-richmond:
Rebased onto `main` to clear a merge conflict, and re-ran everything at the new head.

The conflict was two cache CSVs — `cache/go/terms.csv` and `cache/enums/molecularfunctionterm_305077eac108.csv` — where another PR added `GO:0015279` (store-operated calcium channel activity) at the same sorted position this branch adds `GO:0015297` (antiporter activity). Resolved as a union keeping both sides' rows byte-for-byte, in CURIE order. No row was retyped and nothing was dropped: the cache diff against `main` is still six added lines across six files, zero deletions, zero reordering.

Every non-cache file is byte-identical to the approved head `9cbb2e404` — verified by comparing blob hashes file by file, not by reading the diff. So the curated content that was reviewed is exactly the curated content here; only the cache merge and the base moved.

Re-validated at `c79f5166c`:

| Check | Result |
|---|---|
| `just validate` | passed |
| `just count-verified-snippets` | 67/67 |
| reference titles vs cache | 67/67 exact, re-audited by script |
| `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` / `check-qualifier-terms` | clean |
| `check-snippet-length` / `check-title-snippets` / `check-folded-hyphens` | clean |
| `check-snippet-grading` / `check-enum-values` / `check-environmental-evidence` | clean |
| `just validate-history-all` | no issues |

The push dismissed the round-3 approval, as branch protection is set to do. Nothing in the reviewed content changed, so this needs a re-review rather than a re-read.

---

Phillip-a-richmond:
The merge queue dequeued this PR with `MERGE_CONFLICT`. Fixed and pushed — head is now `b9b5bf39d`.

The conflict is the same shape as the one on 4 September, in the same two files: `cache/go/terms.csv` and `cache/enums/molecularfunctionterm_305077eac108.csv`. Another PR added `GO:0015280` (ligand-gated sodium channel activity) at the sorted position where this branch adds `GO:0015297` (antiporter activity). Resolved as a union keeping both sides' rows byte-for-byte, in CURIE order. No row was retyped: the cache diff against `main` is still six added lines across six files, zero deletions, zero reordering.

Every non-cache file is byte-identical to the approved head `c79f5166c` — verified by comparing blob hashes file by file, not by reading the diff. The curated content that was approved on round 4 is exactly the content here.

**Why the fix took four hours to land.** The conflict appeared while the PR was sitting in the merge queue, and a queued branch cannot be pushed to:

```
GH006: A pull request for this branch has been added to a merge queue.
Branches that are queued for merging cannot be updated.
```

The resolution was ready at 05:20 UTC and refused on six attempts until the queue dequeued the PR at 09:27. The alternative — dequeuing it myself to force the push through — isn't mine to do, so the branch waited.

Re-validated at `b9b5bf39d`:

| Check | Result |
|---|---|
| `just validate` | passed |
| `just count-verified-snippets` | 67/67 |
| reference titles vs cache | 67/67 exact, re-audited by script |
| `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` / `check-qualifier-terms` | clean |
| `check-snippet-length` / `check-title-snippets` / `check-folded-hyphens` | clean |
| `check-snippet-grading` / `check-enum-values` / `check-environmental-evidence` | clean |
| `just validate-history-all` | no issues |

The push drops the round-4 approval, as branch protection is set to do. Nothing in the reviewed content changed, so this needs a re-review rather than a re-read.

---

Phillip-a-richmond:
`claude-review` is red on `b9b5bf39d`, and the cause is not in this PR. The job log says the review agent never ran:

```
num_turns=1 cost=$0
```

That is the reviewer App's own usage limit, hit four seconds into the job. No review was posted, so there are no findings to address, and nothing in the diff would change the outcome.

I'm deliberately not re-running it yet: a re-dispatch before 11:20 UTC fails identically and burns a run. I'll re-dispatch `claude-code-review.yml` for this PR once the limit resets, and keep watching until the review lands.

`test (3.13)` is unaffected and running on its own.

---

Phillip-a-richmond:
Correcting my previous comment: I said I would re-dispatch `claude-code-review.yml` once the reviewer's usage limit reset at 11:20 UTC. I can't — my token isn't allowed to.

Both routes return the same thing:

```
POST /actions/workflows/claude-code-review.yml/dispatches   → 403 Resource not accessible by integration
POST /actions/runs/33958053265/rerun-failed-jobs            → 403 Resource not accessible by integration
```

So the commitment I made in that comment was wrong, and I'm flagging it rather than leaving it to look like something is in progress.

Where that leaves the PR: `test (3.13)` is green on `b9b5bf39d`, there's no merge conflict, and the round-4 approval was dropped by the rebase push. The only thing outstanding is a review, and the check that would produce one is red because the reviewer never ran. Re-running it needs someone with workflow-dispatch permission — or it will re-trigger by itself on the next push to this branch, which is likely, since this branch has already collided with `cache/go/terms.csv` twice in two days.

I'll keep watching and will re-run every gate on any new head.

---

Phillip-a-richmond:
Third cache-CSV conflict on this branch, resolved the same way. Head is now `9e724e3bf`.

This one was `cache/enums/diseaseterm_f618932bb23f.csv`: `main` added `MONDO:0032822` at the sorted position where this branch adds `MONDO:0032805`. Union, both sides' rows kept byte-for-byte, in CURIE order. The cache diff against `main` is still six added lines across six files, zero deletions, zero reordering.

Every non-cache file is byte-identical to the previous head `b9b5bf39d` — blob hashes compared file by file — and that head was in turn identical to the round-4 approved content.

Re-validated at `9e724e3bf`: `just validate` and `just validate-terms` pass, 67/67 snippets verified, 67/67 reference titles re-audited against the cache, and `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` / `check-qualifier-terms` / `check-snippet-length` / `check-title-snippets` / `check-folded-hyphens` / `check-snippet-grading` / `check-enum-values` / `check-environmental-evidence` / `validate-history-all` are all clean.

**This push also re-triggers `claude-code-review.yml`, which is the thing that was stuck.** The previous review run died on the reviewer App's own usage limit and I had no permission to re-dispatch it. A fresh run on this head should produce the review the PR has been waiting for since 09:28.

Worth noting for whoever lands this: the branch has now collided with a `cache/**` CSV three times in about 26 hours — `cache/go/terms.csv` twice and the disease-term enum once — because it adds rows to files that almost every curation PR also touches. Each round-trip costs an approval. It's worth merging promptly once green rather than letting it age.

---

Issue #8431, opened 2026-08-13
Prevalence vs incidence: measure_type is a discriminator no slot or test respects


cmungall:
## The problem

`PrevalenceMeasureEnum` was added as an explicit discriminator, and its own description states the goal precisely:

> so a point prevalence is never silently compared with an incidence rate or a literature case-count

The discriminator is good. But **the slots underneath it are still prevalence-shaped**, so the separation is documented rather than enforced, and the KB has drifted. Nothing downstream of `measure_type` — not the rate slot, not the band slot, not any test — knows the difference between a proportion and a rate.

I hit this while trying to curate predicted incidence estimates from López-Rivera et al. 2020 (`PMID:32168371`, *Brain* 143:1099), whose source column is *"Predicted de novo disease incidence in 100,000 **births**"*. Deciding how to record that surfaced all three defects below.

## Defect 1 — `rate_per_100000` has no dimension

`Prevalence` has no `unit` slot (`EpidemiologyInfo` does; `Prevalence` doesn't). The slot description says "cases per 100,000" and every conversion rule it gives is a **proportion** conversion (`% -> x1000`, `per million -> /10`, `1 in N -> 100000/N`).

- A point prevalence of `5.0` is dimensionless — a proportion of a population.
- An annual incidence of `5.0` is 5 per 100,000 **per year** — dimension time⁻¹.

Both are stored in the same slot under the same name. Two records reading `rate_per_100000: 5.0` are not comparable and nothing in the data says so. Person-time is also inexpressible: there is no way to distinguish "per 100,000 person-years" from "per 100,000 population per year", nor "per 100,000 live births" from either.

## Defect 2 — Orphanet *prevalence* bands are applied to non-prevalence measures

`_band_from_rate()` (`scripts/migrate_prevalence.py:186`, cited as authoritative at `src/dismech/schema/dismech.yaml:276`) is a pure function of the numeric rate. It never inspects `measure_type`. So an incidence or a carrier frequency gets bucketed into bands whose semantics are "proportion of the population that has this disease".

Audit over `kb/disorders/` + `kb/comorbidities/` (1,469 prevalence records):

| `measure_type` | records | carrying a **numeric** `prevalence_class` |
|---|---|---|
| `ANNUAL_INCIDENCE` | 100 | **79** |
| `CARRIER_FREQUENCY` | 18 | **17** |

[...]

Issue #7520, opened 2026-08-01
Render structured epidemiology, prevalence, diagnosis, and progression on disorder pages


aparente:
### Ask

Add disorder-page rendering for four populated reader-facing slots that the current template
does not render:

- `diagnosis`
- `epidemiology`
- `prevalence`
- `progression`

### Evidence of the gap

Across 1,712 disorder YAML files:

| Slot | Files with data |
|---|---:|
| `diagnosis` | 860 |
| `prevalence` | 807 |
| `progression` | 505 |
| `epidemiology` | 116 |

`src/dismech/templates/disorder.html.j2` is 6,416 lines long and contains no case-insensitive
reference to any of those slot names. The template has no renderer block for them, so the only
deterministic presentation of those structures is the collapsed Source YAML block.

Some pages contain similarly named headings inside embedded deep-research reports. Those are
report prose, not renderer-generated views of the structured slots, and they do not provide
consistent coverage or field-level presentation.

### Worked example

[`Cystic_Fibrosis.html`](https://dismech.monarchinitiative.org/pages/disorders/Cystic_Fibrosis.html)
has five structured prevalence records spanning several populations. The records include
measure type, prevalence class or normalized rate, notes, and evidence.

The generated page contains 29 occurrences of “prevalence.” Three are reader-facing statements
about the prevalence of particular features; the other 26 are inside Source YAML. None presents
the five disease-prevalence records as a structured section.

### Why `epidemiology` belongs in the same scope

`prevalence` and `epidemiology` are distinct schema slots, but they overlap semantically in the
current corpus:

| Population of files | Count |
|---|---:|
| Both slots | 47 |
| `prevalence` only | 760 |
| `epidemiology` only | 69 |

Some of the 69 epidemiology-only files contain prevalence data. For example,
`Granulomatosis_with_Polyangiitis.yaml` stores European and US point-prevalence records under
`epidemiology`; `Microscopic_Polyangiitis.yaml` stores both annual incidence and point
prevalence there. Rendering only `prevalence` would leave those records invisible.

### This appears to be known unfinished work

Design Decisions §8 explains that the prevalence remodel carried no display-breakage risk
because prevalence was not yet rendered on disorder pages. That accurately describes the
current template, but I could not find a later issue covering the display step after #5078.

### Suggested presentation

Use one **Epidemiology** container for both population-level collections, plus dedicated
Diagnosis and Natural History sections. Reuse the schema labels and existing evidence partials
[...]

PR #10758, opened 2026-09-03
Define prevalence bands as magnitude, add rate_denominator (#8431 item 1)

cmungall:
Item 1 of #8431, and only item 1: pin what a prevalence band *means*. No KB files touched.

## The problem

`PrevalenceMeasureEnum` exists so that "a point prevalence is never silently compared with an incidence rate" — its own words. Nothing below it honoured that:

- `rate_per_100000` has no dimension. A point prevalence of `5.0` is a dimensionless proportion; an annual incidence of `5.0` is 5 per 100,000 **per year**. Same slot, same name, no `unit` field on the class.
- `_band_from_rate()` (`scripts/migrate_prevalence.py:186`) assigns Orphanet **prevalence** classes purely as a function of the number, never reading `measure_type`.
- Nothing enforces any of it — zero references to `measure_type` in `tests/` or the QC plugins.

Current drift: **116 of 138** `ANNUAL_INCIDENCE` and **41 of 45** `CARRIER_FREQUENCY` records carry numeric prevalence bands.

## What this changes

**`PrevalenceClassEnum` now documents bands as reporting magnitude only**, with `measure_type` supplying what is being measured. `BAND_1_9_PER_100000` means "1–9 per 100,000 of whatever this record measures" — people with the disease, new cases per year, or carriers — and is meaningless read alone. Consumers must read the two together.

**The qualitative tiers are the exception.** `COMMON`/`RARE`/`ULTRA_RARE` are each defined by a *prevalence* threshold and each presuppose the source gave no numeric estimate, so they are invalid on `ANNUAL_INCIDENCE` and `CARRIER_FREQUENCY`. They stay correct on the prevalence measures, on `CASES_IN_LITERATURE`, and on `UNKNOWN`.

That last scoping matters and I got it wrong on the first pass. Of the 90 records pairing a qualitative tier with a non-prevalence measure, **82 are `UNKNOWN`** — a source that says only "rare" without naming its measure, which is exactly the case these tiers exist for. Only **8** are real errors, all `ANNUAL_INCIDENCE`. An unscoped rule would have condemned 82 correct records.

**New optional `rate_denominator` slot** (`RateDenominatorEnum`: `POPULATION`, `LIVE_BIRTHS`, `PERSON_YEARS`, `POPULATION_PER_YEAR`) makes the dimension explicit rather than implied. Absent, consumers fall back to the denominator implied by `measure_type`.

## Why this reading rather than stripping the bands

The alternative — bands as strictly prevalence-typed, stripped from incidence and carrier records — would rewrite **157 KB records** against a queue of 100+ open curation PRs, and would *lose* information: the only landing place for those records is a qualitative tier, which is both less precise and itself prevalence-semantic, so the category error survives in a less detectable form.

Worked counterexample, recorded in the decision register. `2-Methylbutyryl-CoA_Dehydrogenase_Deficiency.yaml` carries a 21.8% Hmong carrier frequency *and* a 1-in-227,571 Zhejiang birth prevalence. Demoting the carrier record's band to its nearest tier (`COMMON`) would leave one file asserting both that the disease is common and that it is 1–9 per million.

The magnitude reading costs one thing: a band is no longer self-describing. That cost is unusually cheap to control **right now**, because these fields have no consumers — grep finds no reference to `prevalence_class`, `rate_per_100000`, or `measure_type` outside the schema and two migration scripts, and #7520 (which would render them) is still open. The rule is a requirement on the first consumer rather than a migration of existing ones.

## Files

| file | what |
|---|---|
| `src/dismech/schema/dismech.yaml` | enum semantics; `RateDenominatorEnum`; `rate_denominator` slot on `Prevalence` |
| `docs/explanation/design-decisions.md` | §8 amendment: the decision, the rejected alternative, accepted debt, deferred items |
| `CLAUDE.md` | curator guidance + an incidence example showing an explicit denominator |
| `references_cache/*.md` (2nd commit) | incidental — see below |

The reference-cache commit is fallout from running `just validate` on two disorder files to check the schema against real data. The validator fetched full text upstream doesn't have (two files gain ~840 lines each). Per CLAUDE.md these enrichments are meaningful and get committed rather than restored away, so they're here in their own commit; the diff is additive and whitespace-normalised by the repo's own pre-commit hooks.

## Verification

`47713 passed` — `tests/test_data.py`, `test_clinical_burden_schema.py`, `test_genetic_schema.py`, `test_enum_cache.py`. Schema loads through `SchemaView` (136 classes, 422 slots, 122 enums) and `just validate` passes on disorder files carrying both prevalence and carrier-frequency records.

Note the one failure I hit mid-work (`Urticaria.yaml`, unprefixed `https://pmc.ncbi.nlm.nih.gov/...` references) was pre-existing on my stale base and is already fixed on `main`.

## Not in this PR

Deliberately deferred, per the sequencing in #8431:

1. The QC check gating qualitative tiers to non-`ANNUAL_INCIDENCE`/`CARRIER_FREQUENCY` measures — 8 records to fix. Should land with the band-vs-rate check proposed in **#7005**, which has had no PR in three weeks, so that scaffold is still unbuilt.
2. Backfilling the **286** records with no `measure_type` (a further 203 are `UNKNOWN`), after which the slot can become required.
3. Whether `CARRIER_FREQUENCY` belongs on `Prevalence` at all — it is a genotype frequency among unaffected people, and `GeneCaseFraction` is structurally closer. Moving it would remove 41 flagged records from this class outright.
4. **#7003** — `SCREEN_POSITIVE_RATE` missing from `PrevalenceMeasureEnum`, the same modelling gap one level up.

One ordering constraint worth carrying forward: **#7520** is the issue that turns these fields from write-only into user-visible, and it is where the "read `measure_type` alongside the band" rule needs to be honoured.

Refs #8431, #7005, #7003, #7520


cmungall:
All six addressed. I reproduced each before acting, and every one held up — including both places where my own numbers were wrong.

## IMPORTANT 1 — you're right, and I can say exactly how I got it wrong

The renderer has consumed these fields since **#9510** ("Render nine curated sections that no page showed"). My grep was run on a base ~1,100 commits stale, and #9510 landed inside that gap:

```
$ git show 6c797e08ef:src/dismech/templates/disorder.html.j2 | grep -c measure_type   # my grep's base
0
$ git show 036da5b6e9:src/dismech/templates/disorder.html.j2 | grep -c measure_type   # my actual base
2
```

The specific failure is worth naming: after rebasing I re-verified the *statistics* against the new base but not the *premise*. The numbers were the part I expected to drift, so that's what I re-ran.

**Fixed the template rather than deferring it**, because the `rate_denominator` half isn't really deferrable — shipping a slot that the only consumer silently drops is a half-delivered feature. Rendered output now:

| record | before | after |
|---|---|---|
| Hmong carriers (`CARRIER_FREQUENCY`) | `21800.0 per 100,000` · `>1 in 1,000` | `21800.0 per 100,000` · `>1 in 1,000 (carriers)` |
| Olmsted County (`ANNUAL_INCIDENCE`) | `1.2 per 100,000 (0.8–1.6)` · `1–9 per 100,000` | `1.2 per 100,000 (0.8–1.6)` · `1–9 per 100,000 per year` |
| Newborns, Zhejiang (`BIRTH_PREVALENCE`) | `0.4394 per 100,000` · `1–9 per 1,000,000` | unchanged |

`rate_denominator` now renders next to the rate (`person-years`, `live births`, `per year`). §8 records the corrected fact, that the consumer did not honour the rule, and the stale-base error itself — the false premise was load-bearing, so burying it would defeat the point of the register.

## IMPORTANT 2 — removed rather than flipped

You're right that it was self-refuting: set `PERSON_YEARS`, then say to override it "which is most common". I took your second option. `ANNUAL_INCIDENCE` now has **no** fallback, documented as deliberate: both denominators are common, they're interchangeable only when the population is stable, and a fallback would assert a dimension for 138 records migrated with no denominator information. Your framing is in the schema and the register nearly verbatim — an unstated default costs a consumer one branch; a usually-wrong default costs it correctness. CLAUDE.md now says to always set the slot on an incidence record.

## IMPORTANT 3 — 21, and your count was right where mine wasn't

My first recount got **15**, not your 18. You were right: I filtered on `rate_per_100000` alone, and three of the violations carry only `rate_low`/`rate_high` — `Bosch-Boonstra-Schaaf` (0.4–1.0), `Vogt-Koyanagi-Harada` (0.6–1.7), `Langer_Mesomelic_Dysplasia` (≤0.1). Including ranges gives 18, overlap 5, union **21**. Deferred item (a) now says 21 with the breakdown, and flags the five `rate_per_100000: 0.0` `CASES_IN_LITERATURE` records as a separate pre-existing bug.

## SUGGESTION 4 — taken

`POPULATION` and `LIVE_BIRTHS` now read "whatever `measure_type` counts, per 100,000 …" instead of "Cases per 100,000". You're right that wording the denominator enum in terms of cases gave back exactly what the split buys, and the `CARRIER_FREQUENCY` → `POPULATION` mapping is where it showed.

## SUGGESTION 5 — taken, and it's larger than either of us said

Both reworded. The KB exposure is bigger than the 1 + 7 in your note — **109** records pair `NOT_YET_DOCUMENTED` or `UNKNOWN` with a non-prevalence measure, mostly `CASES_IN_LITERATURE` (61 + 40):

```
{('NOT_YET_DOCUMENTED','CASES_IN_LITERATURE'): 61, ('UNKNOWN','CASES_IN_LITERATURE'): 40,
 ('UNKNOWN','ANNUAL_INCIDENCE'): 5, ('UNKNOWN','CARRIER_FREQUENCY'): 2,
 ('NOT_YET_DOCUMENTED','CARRIER_FREQUENCY'): 1}
```

## SUGGESTION 6 — taken

`just gen-python` run; `datamodel/dismech.py` and `dismech_pydantic.py` both regenerated. Your inference was right that it's kept in sync — `biological_scale` was already present there.

## Verification

- `4455 passed` across the render, template, schema, and enum suites.
- One failure, `test_render_prune_orphan_pages.py::test_prune_keeps_a_differently_named_page_that_is_the_same_file`, is macOS-only and pre-existing: it hardlinks `Holt-Oram_Syndrome.html` → `Holt-Oram_syndrome.html`, which collide on case-insensitive APFS (`FileExistsError`). dismech#565. It passes on CI's Linux runner.
- `just validate` passes on the worked counterexample entry; pages rendered and inspected for the three record shapes in the table above, then reverted (derived).

Thanks for catching the consumer claim in particular — I'd have shipped a design record whose central argument was false, and the register is exactly the artifact where that would have gone unnoticed longest.

cmungall:
Both non-blocking notes taken, plus a third that the second one surfaced.

**`BIRTH_PREVALENCE` qualifier — added.** One dict entry, as you said. The Zhejiang record now renders `1–9 per 1,000,000 (births)` rather than a bare population-shaped band.

**`Rare per year` — this turned out to be more than cosmetic.** You flagged it as an acceptable way for the 8 bad records to surface, but it's the same error the PR exists to fix, one level down: a qualitative tier is explicitly *not* a magnitude statement (the schema text I added says so), so attaching a denominator qualifier to one repeats the numerator/denominator conflation in the renderer. The qualifier now applies only to the five numeric bands. `Carbon_Monoxide_Poisoning` renders `Common`, not `Common per year`. Those records stay visibly wrong via the adjacent `Annual Incidence` tag, which is the honest signal — deferred item (a) is what fixes them.

**Counts — your recount reproduces exactly.** I get 118/140 `ANNUAL_INCIDENCE`, 41/45 `CARRIER_FREQUENCY`, 218 `UNKNOWN`, 286 unset, and 8 / 20 / overlap 5 → **23** across 22 files (`Specific_Antibody_Deficiency` carrying two, as you noted). §8 updated.

Rather than re-chase them next week, the amendment now states outright that the figures are a snapshot at a named commit and date, that this repo merges fast enough for them to drift within days, and how to re-derive them. Same reasoning as taking the counts out of the schema description in `c549667` — the register can carry a dated snapshot because it *is* dated; a schema description can't, because it isn't.

Rendered output after the change:

```
Hmong (Wisconsin)        Birth Prevalence   1300.0 per 100,000    >1 in 1,000 (births)
Hmong carriers           Carrier Frequency  21800.0 per 100,000   >1 in 1,000 (carriers)
Newborns, Zhejiang       Birth Prevalence   0.4394 per 100,000    1–9 per 1,000,000 (births)
United States (CO)       Annual Incidence   15.1 per 100,000      Common
```

Verification: `pytest -k "render or template or prevalence"` → 253 passed; the single failure is the pre-existing macOS-only `test_prune_keeps_a_differently_named_page_that_is_the_same_file` (hardlinks `Holt-Oram_Syndrome.html` → `Holt-Oram_syndrome.html`, which collide on case-insensitive APFS — dismech#565; passes on CI's Linux runner). Derived pages rendered, inspected, not committed.

This dismissed your approval, sorry — the push was needed to land the fixes. Re-review whenever convenient.

cmungall:
Findings 2 and 5 fixed. Findings 1, 3 and 4 were already fixed in `762bed4d8e` — this review evaluated the previous head.

**Timing:** the review posted at 22:37:58; run `33925974046` (headSha `762bed4d8e`) was still `queued` at that moment and had not executed. So the tree reviewed was `c5496675ec`. Confirmed by rendering your own example on the head you didn't see:

```
Asbestosis  ->  Annual Incidence   Unknown        # not "Unknown per year"
```

- **Finding 1** (qualifier glued to non-magnitude labels) — fixed in `762bed4d8e`, and by exactly the mechanism you prescribed: the qualifier is gated on `value in NUMERIC_PREVALENCE_CLASSES`, not on `measure_type` being truthy. Your diagnosis and mine converged independently — I'd reached it from the `Rare per year` aside in your *previous* review, which pointed at the same defect from the qualitative-tier side while your `UNKNOWN` five come at it from the measure-agnostic side. Both are covered by the one gate; all 13 records you listed render clean.
- **Finding 3** (`BIRTH_PREVALENCE`) — added in the same commit. Worth noting your 271-record count makes it the *largest* of the three qualifier cases, bigger than the 41 carriers that motivated the mechanism.
- **Finding 4** (drifted counts) — §8 already reads 118/140, 41/45, 8 / 20 / overlap 5 → 23, 286 / 218. My recount reproduced yours exactly. It also now states that the figures are a snapshot at a named commit and date and how to re-derive them, so the next drift reads as expected rather than as an error.

**Finding 2 — fixed, and you were right that it's more than a doc nit.** The slot description promised a fallback the enum had just removed. Since slot descriptions are what generated schema docs and editor tooltips surface, it was the one of the three texts most likely to be the *only* one a curator reads — and it said the opposite of the other two. It now leads with the unconditional `ANNUAL_INCIDENCE` obligation and scopes the fallback to the measures that actually have one.

**Finding 5 — fixed.** The lookup now humanises unmapped members instead of returning `''`, matching how `PREVALENCE_CLASS_LABELS` handles the same case three lines up. `POPULATION` still maps to `''` deliberately — it's the implicit reading and needs no label — so the lookup distinguishes "mapped to empty" from "not mapped", which is the distinction that made the bug possible.

Verification: `pytest -k "render or template or prevalence or schema"` → 4380 passed; the one failure is the pre-existing macOS-only `test_prune_keeps_a_differently_named_page_that_is_the_same_file` (dismech#565), which passes on CI's Linux runner. `just gen-python` re-run so both datamodels carry the corrected slot description.

Housekeeping: nothing to discard on my side — my worktree is clean and no `pages/` files are staged or committed on this branch.

PR #10759, opened 2026-09-03
curate(DFNA78): add hearing loss, autosomal dominant 78 (SLC12A2)

### LINKED ISSUE #10756: Curate hearing loss, autosomal dominant 78 (MONDO:0033665)
Curate a dismech entry for **hearing loss, autosomal dominant 78** (DFNA78, [MONDO:0033665](https://monarchinitiative.org/MONDO:0033665)) — nonsyndromic dominant hearing loss caused by heterozygous *SLC12A2* (hgnc:10911) variants; SLC12A2 encodes NKCC1, the basolateral cotransporter that drives potassium into stria vascularis marginal cells and so sustains endolymph secretion and the endocochlear potential.

Preflight: MONDO:0033665 appears nowhere in `kb/` on `main`, and no open PR covers it. Three *SLC12A2* entries already exist — `Delpire-McNeill_Syndrome.yaml` and `Kilquist_Syndrome.yaml` (syndromic), and `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103.yaml` (the recessive hearing-only form). DFNA78 is the dominant hearing-only form, reported as acting through a dominant-negative rather than haploinsufficiency mechanism; the entry should make that contrast with its siblings explicit rather than duplicating them.

Stub: `stubs/Hearing_Loss_Autosomal_Dominant_78.yaml`

aleixpuigb:
Adds `kb/disorders/Hearing_Loss_Autosomal_Dominant_78.yaml` for hearing loss, autosomal dominant 78 ([MONDO:0033665](https://monarchinitiative.org/MONDO:0033665)) — dominant nonsyndromic sensorineural hearing loss from heterozygous *SLC12A2* variants.

Closes #10756

## What organises the entry

Every published pathogenic DFNA78 allele sits in exon 21 — either a missense change inside it or a splice variant that removes it. That is not a coincidence of small numbers: *SLC12A2* has several isoforms, only one contains exon 21, that isoform is almost exclusively expressed in the inner ear, and it is the one needed for endolymph homeostasis. So a lesion confined to one exon is, in practice, confined to the cochlea, which is how a cotransporter expressed in neurons, glia, trachea, salivary glands, intestine and sweat glands yields a hearing-only disease.

```
SLC12A2 exon 21 variant
        |
        v
Loss of the inner-ear-restricted NKCC1 exon 21 isoform
        |
        v
Failed basolateral K+ uptake by strial marginal cells
        |
        v
Reduced endolymph production and strial atrophy
        |
        +--> sensorineural hearing impairment       (DIRECT)
        +--> abnormal vestibular function           (INDIRECT_UNKNOWN_INTERMEDIATES)
```

## The judgement call a reviewer should check

**I did not tag the variants `DOMINANT_NEGATIVE`, and I think that is right even though it looks like the obvious answer.** The disease is dominant, NKCC1 is a dimer, and heterozygous whole-gene loss of function is compatible with normal hearing — so haploinsufficiency of the *gene* is ruled out and a poisoned dimer is the natural story.

But the splice allele argues the other way. PMID:32294086 showed its exon-21-skipped transcript is present at much lower levels than the exon-21-included transcript in cochlea. What that patient loses is isoform *quantity*, not a subunit that spoils its partner. A missense subunit could go either way and nobody has tested it.

So `functional_impact_category` records what was measured — `LOSS_OF_FUNCTION`, from chloride influx significantly decreased for every tested allele in two independent cohorts — and the route to dominance is an `OPEN_QUESTION` discussion that names the experiment which would settle it. I would rather the entry say the mechanism is unresolved than assert a dominance model the data do not pick out.

## Model fidelity is recorded as a mismatch, not a caveat

The `Slc12a2Em2/Em2` exon-21-skip mouse reproduces the chain convincingly: reduced endolymph at P1, smaller stria vascularis, no ABR at four weeks. It is linked with `RECAPITULATES` / `fidelity: MODERATE` and two readouts.

The dominance does not carry across. The heterozygous mouse shows only mild ABR threshold elevation, in females, at some frequencies, whereas the heterozygous human is severely to profoundly deaf. That is a `HUMAN_MODEL_MISMATCH` discussion rather than a line in `limitations`, and the paper supplies a concrete candidate explanation: a single nucleotide differs between human and mouse at the 5' end of exon 21 and affects how the exon is spliced.

## Deep research

One openscientist run, committed under `research/`. It materially improved the entry and it also produced a checkable error.

**What it added.** It surfaced the two founding 2020 papers I had not found by direct search — PMID:32294086 (Mutai, *PLoS Genet*) and PMID:32658972 (McNeill, *Brain*). Those gave the entry its direct functional evidence (decreased chloride influx in all tested variants, replicated in a second cohort), the immunohistochemical localisation of SLC12A2 to strial marginal cells, and the transcript-level result above. **They changed a call**: I had drafted `DOMINANT_NEGATIVE` before reading them.

**What it got wrong, and what caught it.** The report offered `CL:0002516` for strial marginal cells. That CURIE resolves — to *interrenal chromaffin cell*, an adrenal cell type. The correct term is `CL:0002492`. The report's own `## Term Validation` section did **not** flag it, and the reason is worth stating: of 43 CURIEs it checked, only 4 had their names checked. The other 39 were confirmed to exist and nothing more. It did catch `UBERON:0002429` (cervical lymph node offered for stria vascularis), because there the label happened to sit where the extractor could find it.

That asymmetry is the useful lesson: a report's term-validation summary saying "43 checked, 0 unresolved" reads like assurance and is not one. I have filed the gap rather than only working around it.

## Cache hygiene

Re-fetching references that already exist on `main` rewrote their frontmatter (quote style, a `full_text_attempted` flag) and in two cases appended full text. I reverted all four to the committed versions and re-verified — 34/34 snippets still pass against them — so this PR adds exactly one new cache file, `PMID_34374074.md`, and touches no existing one.

## Validation

```
just validate         -> schema OK, terms OK, 34/34 snippets verified
just validate-history -> OK
```

Two ontology bindings are deliberately broader than the claim, with the specificity carried in `preferred_term` and the reason in `notes`: `HP:0001751` (Abnormal vestibular function) for vestibular areflexia, and the cell-level potassium node. `PMID:36035115` is cached but **not** cited, and `notes` says why — it reports functional work on an *SLC12A2* variant from a family whose inheritance mode that paper does not state, so attributing it to DFNA78 would be an assumption.

---

### DIFF
--- a/history/disorders/Hearing_Loss_Autosomal_Dominant_78/2026-09-03T200314Z-claude-code-0b1c2e.yaml
+++ b/history/disorders/Hearing_Loss_Autosomal_Dominant_78/2026-09-03T200314Z-claude-code-0b1c2e.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Hearing_Loss_Autosomal_Dominant_78
+  path: kb/disorders/Hearing_Loss_Autosomal_Dominant_78.yaml
+session:
+  id: 2026-09-03T200314Z-claude-code-0b1c2e
+  timestamp: '2026-09-03T20:03:14Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10759
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  summary: 'Review round 2: complete the finding-6 regrade on #10759'
+  details: |
+    One line. The round-1 regrade of finding 6 was half-applied: the PMID:40295800 allelic-spectrum snippet was regraded to OTHER at genetic[0].evidence[0] but left HUMAN_CLINICAL at pathophysiology[0].evidence[0], while its explanation had already been rewritten to say 'Graded OTHER'. The file therefore graded one quoted sentence two ways and check-snippet-grading exited 1. Corrected to OTHER. Correcting the record of the round-1 session rather than that record itself, per the append-only rule: the round-1 history record and commit message both state that two statements were regraded, and what actually happened is that one was regraded in place and a different instance of the same sentence was regraded instead of the named one.
--- a/kb/disorders/Hearing_Loss_Autosomal_Dominant_78.yaml
+++ b/kb/disorders/Hearing_Loss_Autosomal_Dominant_78.yaml
@@ -122,7 +122,7 @@ pathophysiology:
   - reference: PMID:40295800
     reference_title: "Complete omission of exon 21 from Slc12a2 transcripts in mice results in hearing loss."
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "SLC12A2 is linked to autosomal dominant nonsyndromic hearing loss, DFNA78, with all the pathogenic variants affecting the exon 21."
     explanation: >-
       The confinement of the human allelic spectrum to one exon, which is the fact this whole


aleixpuigb:
All six findings taken, plus both structural suggestions. Pushed as `05faa1d`.

**Finding 1 is the one that stings, and you diagnosed it exactly right.** That paragraph was a survivor of the discarded draft. I *did* write a replacement for it during curation — the edit silently failed to match on line wrapping, and I did not re-read the description afterwards because I had verified the change I cared about (`genetic_context`) and moved on. So the PR body confidently described a call the entry's own opening paragraph contradicted. Rewritten, and line 26 narrowed from "argue against haploinsufficiency" to "rule out haploinsufficiency of the *gene*", which is what the discussion actually supports.

**Finding 2 is the most valuable thing in this review** and I want to be explicit about why. The abstract does not name a species. Nothing in this repo's stack could have caught it: `just validate` checks the snippet is an exact substring of the cached text and passes; `check-snippet-grading` only fires when the same sentence carries two different `evidence_source` values in one file, and this one is consistent. You had to open the PDF methods section to find *Macaca fascicularis*. Regraded `MODEL_ORGANISM` with `directness: INDIRECT`, and the explanation now says the species out loud, because the strial marginal cell is the anatomical anchor of this whole mechanism and it turns out to rest on non-human primate tissue.

That is a general gap rather than a one-off, and I have written it up as such in the retrospective for this batch: **`evidence_source` is often unverifiable from the abstract the validator checks against, and nothing flags that it was unverifiable.**

**Finding 3 — you are right and my note was wrong.** `HP:0008568` is in `cache/hp/terms.csv` from May, and bound in six entries including `Delpire-McNeill_Syndrome`. I searched for a vestibular-areflexia term, did not find one under the IDs I guessed, and then wrote a `notes` asserting none existed rather than treating my own failed search as a failed search. Asserting an ontology gap should require the same standard as asserting a finding, and it did not get it here. Rebound, node renamed to `Vestibular Areflexia`, downstream target and the new treatment's `target_phenotypes` updated, note deleted.

**Finding 4** — split into `Reduced Endolymph Production` → `Stria Vascularis Atrophy`. You are right that the bundling cost an edge, and it is the more interesting of the two claims: the paper's position is that strial shrinkage is an *osmotic rebalancing* downstream of the secretory defect, not primary degeneration. That reading now has a node to live on. The animal-model link, both readout targets and the mismatch discussion's `attaches_to` are repointed; `check-causal-targets` and `check-entity-refs` clean.

**Finding 5** — treatments added on the `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_104` pattern: cochlear implantation with the `NCIT:C15329` + device-qualifier binding and a `BYPASSES` edge into the hearing-impairment node, amplification, vestibular rehabilitation, and the loop-diuretic avoidance record. Each carries no evidence item with `notes` saying why. Thank you for pointing at the sibling entry — "no quotable snippet" had felt like a sufficient reason and it was not, given the repo already had a sanctioned shape for exactly this.

The loop-diuretic entry is the one I would most like a second opinion on. Nothing in the cached references mentions bumetanide or furosemide; the pharmacology and the ototoxicity are both established general knowledge and the inference joining them to DFNA78 is mine. I recorded it with that labelled explicitly rather than dropping it, because a drug class whose target is the disease protein seems like precisely what this KB should make queryable — but if the house view is that an uncited mechanistic inference should not sit in `treatments` at all, I will pull it.

**Finding 6** — both regraded to `OTHER`. You are right that a mouse paper's introduction presents neither human clinical data nor mouse data, and my own explanation ("graded HUMAN_CLINICAL as background") was visibly uncomfortable with the value it was defending. Four instances in total, since each snippet is used twice.

**Suggestions:** took `GO:0071805` for the basolateral node. Took the severity one too — collapsed to a single `Sensorineural Hearing Impairment` node with `severity: SEVERE`, and the description now says why, because you are right that a `sequelae` edge from a term to its own HPO descendant encodes an is-a relation as causation. On the `references:` block I have left it out: without a DFNA78- or *SLC12A2*-specific chapter, adding the general hearing-loss overview would be a reference the entry does not draw on, and I would rather the block be absent than decorative.

Validation after the changes: `just validate` → schema OK, terms OK, **33/33 snippets verified**; `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms` all clean, plus `check-qualifier-terms-online` for the two new NCIT device terms. A `history/` EDIT record for this round is included.

---

aleixpuigb:
Fixed in `11d91f9`. Line 125 is now `OTHER`; `check-snippet-grading` is clean, and `just validate` still gives 33/33.

You are right that the regrade was half-applied, and the way it failed is worth a sentence. I wrote two `replace` operations for finding 6 — one keyed on the explanation text, one on the bare `evidence_source` + snippet pair. The first matched the instance at `pathophysiology[0]` and rewrote its explanation to say "Graded OTHER" without touching the value above it; the second matched the *other* instance of the same sentence, at `genetic[0]`, and regraded that one. Net effect: one explanation claiming a grade it did not have, one grade changed that I had not set out to change, and a file that graded one quote two ways.

**And I did not run `check-snippet-grading` before pushing.** It is in CLAUDE.md's list, I ran the five gates I had been running all session, and this is precisely the divergence it exists to catch. That is the reason the defect reached you rather than the reason it existed.

Taking your instruction on the history record: the round-1 record and its commit message both say two statements were regraded, which is not what happened. Records are append-only, so that one stands and the new one states what actually occurred.

Noted on `check-source-defect-claims` flagging this entry as a false positive — the claim is specifically that the *abstract* names no species, which is correct, and the matcher is seeing "species" in the full text. Leaving it.

Ready for re-review.

---

Issue #10459, opened 2026-09-02
A deep-research report's Term Validation section can pass over an NCIT term without checking it or saying so


nlharris:
> This issue was generated by Claude.

A report's `## Term Validation` section is the thing CLAUDE.md tells curators to read before binding any CURIE a report suggests. It has two honest outcomes for a term it cannot check — report it `UNVERIFIABLE` and name its prefix under **Prefixes with no resolver** — and in at least one report an NCIT term got neither.

## What happened

`research/3-methylglutaconic_Aciduria_With_Deafness_Encephalopathy_And_Leigh-like_Syndrome-deep-research-openscientist.md` (on `claude/curate-megdel-serac1`) writes:

> **No disease-modifying or curative therapy exists; management is supportive, symptomatic, and multidisciplinary** (NCIT:C15277 Supportive Care).

`NCIT:C15277` is **Mastectomy**. That is the textbook mismatch this validator exists to catch — a real identifier named as something unrelated — and it is exactly the shape the schema's own docstring uses as its worked example (`NCIT:C16814` is Malaysia).

The report's Term Validation section does not flag it. It reports 55 terms checked, 53 resolved, 2 unverifiable, and 13 named as a different term — none of the 13 is NCIT. The **Prefixes with no resolver** line names only `ORPHA`. A prefix census of the report body shows exactly one NCIT term among 74 CURIE mentions, so a reader has no way to tell from the section that NCIT went unchecked.

Whatever the underlying cause — the term counted as one of the 2 unverifiable without its prefix reaching `unresolvable_prefixes`, or the default `ols:` adapter not serving NCIT — the visible outcome is the same: a wrong NCIT code passed through the section silently.

## Why it matters here specifically

NCIT is the ontology dismech binds every `treatment_term` to, and a wrong NCIT code is unusually survivable downstream. `linkml-term-validator` will confirm the CURIE resolves; `just check-not4curation` will pass it; nothing in the stack asks whether *Mastectomy* is a plausible treatment for a mitochondrial encephalopathy. The Term Validation section is the only place that comparison is made, so a silent skip there is the whole check.

## Suggested fix

1. Make a skipped prefix always reachable from the section: if a term is `UNVERIFIABLE`, its prefix belongs in `unresolvable_prefixes`, and the rendered section should say so even when the count is small.
2. Consider stating the checked-prefix set positively — "names checked for HP, GO, CL, UBERON, MONDO" — so absence is legible rather than inferred.
[...]

PR #10761, opened 2026-09-03
curate(3q29dup): add chromosome 3q29 microduplication syndrome

### LINKED ISSUE #10755: Curate chromosome 3q29 microduplication syndrome (MONDO:0012761)
Curate a dismech entry for **chromosome 3q29 microduplication syndrome** ([MONDO:0012761](https://monarchinitiative.org/MONDO:0012761)) — the reciprocal copy-number gain of the recurrent ~1.6 Mb 3q29 interval flanked by low-copy repeats, associated with variable intellectual disability, microcephaly/macrocephaly and dysmorphism, and with markedly reduced penetrance compared with the deletion.

Preflight: MONDO:0012761 appears nowhere in `kb/` on `main`, and no open PR covers it. `kb/disorders/Chromosome_3q29_Microdeletion_Syndrome.yaml` already exists and covers the reciprocal *deletion*, which is a separate entity with a much stronger phenotype; the duplication entry should say how the two relate rather than restate the deletion.

Stub: `stubs/Chromosome_3q29_Microduplication_Syndrome.yaml`

aleixpuigb:
Adds `kb/disorders/Chromosome_3q29_Microduplication_Syndrome.yaml` for [MONDO:0012761](https://monarchinitiative.org/MONDO:0012761) — the reciprocal copy-number *gain* of the recurrent ~1.6 Mb 3q29 interval. The knowledge base already has the reciprocal deletion; this is a separate disease rather than a milder version of it, and the entry is written to keep that distinction visible.

Closes #10755

## What is settled, and what is not

The formation mechanism is settled. The interval is flanked by near-identical low-copy repeats and NAHR between them yields the reciprocal pair; optical genome mapping localised the crossovers to paralogous copies of a single 20 kb segment in 89% (16/18) of probands. I also curated a negative result: no parent-of-origin inversion was found, which matters because an inversion polymorphism is what explains other recurrent NAHR rearrangements.

**Everything downstream of the dosage change is not settled, and the entry says so structurally rather than in prose.**

### The minimal critical region is a `CONTROVERSY`, carrying both sides

| Position | Source |
|---|---|
| DLG1 + BDH1 (448 kb), then the same two genes at 433 kb in an independent family | PMID:29501613, PMID:36691815 |
| **BDH1 alone**, from five patients with 232–486 kb duplications | PMID:39739615 |
| **No critical region is supported**, from the 46-patient cohort that examined its own nine sub-1 Mb duplications | PMID:38421086 |

I did not pick a winner. The disagreement is not about a boundary — a single-gene region would make this a BDH1 dosage disorder with a tractable mechanism, and no region would make it a contiguous-gene effect. So the candidate genes are curated as candidates, the `REFUTE` item sits in the same node as the claim it contradicts, and the causal link from that node to the clinical layer is `INDIRECT_UNKNOWN_INTERMEDIATES` rather than `DIRECT`.

The discussion also records the fact that complicates *both* positions: in the very family used to refine the region, the same 432.8 kb duplication is carried by an unaffected sibling and an unaffected transmitting mother.

### What the duplication means for a carrier

Penetrance is reduced, most duplications are inherited from a similarly mildly affected parent, and additional contributory genetic findings were identified in 11 of 46 patients. The cohort's own recommendation is curated as a `diagnosis` record rather than left in prose: a severe or syndromic presentation warrants further genetic analysis rather than attribution to the duplication.

### The deletion's findings do not all transfer

An `INTERPRETATION` discussion records three distinct patterns from the direct comparison, because assuming one is how counselling goes wrong:

- most phenotypes occur on both sides but **significantly less often** in duplication carriers
- overweight and weight deficit are **mirrored** — opposite directions, not graded severity
- schizophrenia, generalised anxiety and recurrent ear infections are **deletion-only**

The schizophrenia association is the most cited finding in the 3q29 literature and it is not this entry's.

## Deep research

One openscientist run, committed under `research/`. It surfaced PMID:37165454, PMID:38421086 and PMID:39739615, which I had not reached by direct search and which supply the NAHR breakpoint data, the largest cohort and the BDH1 position respectively. Without it the critical-region controversy would have been curated as a settled single-gene answer.

**The report shipped with no `## Reference Validation` and no `## Term Validation` section at all** — the failure described in #10396. I retro-fitted both with `just validate-research-reference` and `just validate-research-terms`; references came back 17/17 resolved, 16/17 on topic.

The term validation is worth a caveat rather than a clean bill. It flagged 16 terms as "named as a different term", and **all 16 are false positives** — 15 are the phenotype table's *Onset* column being read as the label (`HP:0001250` "the report calls it Childhood; HP calls it Seizure`), and the 16th is markdown bleeding into an extracted label. I [commented on #10459](https://github.com/monarch-initiative/dismech/issues/10459#issuecomment-5530863547) with this and the complementary miss from the DFNA78 run in #10759, because the two together locate the defect in label extraction rather than in prefix coverage.

## A correction made during curation

Three `reference_title` values were written from memory and were wrong — PMID:38421086 is "3q29 duplications: A cohort of 46 patients and a literature review", not the phrasing I gave it. `just validate` caught all three, and they were corrected against the cache. Worth stating because `reference_title` is exactly the kind of field that looks too minor to verify.

One reported case is deliberately **not** curated: a late-onset presentation at age 34 with progressive cortical atrophy (PMID:32874693). One case cannot establish whether it belongs to the syndrome, and `notes` records that it was seen and set aside rather than missed.

## Validation

```
just validate         -> schema OK, terms OK, 33/33 snippets verified
just validate-history -> OK
```

All six reference-cache files are new; no existing cache file is touched.

---


aleixpuigb:
All three blocking findings fixed, plus the treatments point. Pushed as `a9d1dea`.

**Finding 1 — you are right that this was not a research gap, and that is the part worth owning.** The evidence for all four features was already in the file, inside snippets the entry reuses several times. I wrote the description from the literature, curated the phenotypes I had built pathograph edges for, and never re-read the description against the phenotype list. Added `HP:0000256`, `HP:0000478`, `HP:0001999` and `HP:0001627`, reusing the already-verified snippets, with downstream edges from the dosage node. As you note, this also makes the duplication-versus-deletion asymmetry inspectable from this side, which is what `dup3q29_not_the_deletion` is arguing about — it was making a claim the entry gave a reader no way to check.

Two of the four are bound at a category level (`Abnormality of the eye`, `Abnormal facial shape`) with `notes` saying why: the sources name "ocular abnormalities" and "distinctive facial features" without specifying structures or components, so a narrower term would assert a finding the literature does not report.

**Finding 2 — plain error, and the interesting part is that nothing could catch it.** `BAND_1_5_PER_10000` covers 10–99 per 100,000 and so contains 20.0. I picked the band by reading "1 in 5000" as a rare-disease-sounding number rather than by converting it. `just validate` passes because the schema never cross-checks `prevalence_class` against `rate_per_100000`, and both fields are individually well-formed. The record now carries a `notes` line stating the correct band and that the pairing is unchecked, so the next curator has the boundary in front of them.

**Finding 3 — and thank you for the character correction.** You are right on both counts: the snippet stopped before the clause that names the mechanism, so a NAHR node's lead evidence did not mention NAHR; and the character is U+2010 HYPHEN, not U+00AD SOFT HYPHEN. I wrote "soft hyphen" from a guess about why the match failed rather than from looking at the byte. Replaced with the clause immediately after it, which states the mechanism and verifies unchanged.

**Treatments** — added a `Genetic Counselling` record (`NCIT:C15240`). You are right that the entry's whole argument is a counselling argument and it had nowhere to attach. It carries two evidence items substantiating the *facts to be conveyed* — that most duplications are inherited from a mildly affected parent, and that schizophrenia is a deletion-only phenotype — with `notes` stating plainly that no study measures counselling outcomes here, so the evidence supports the content and not the intervention's effectiveness.

On the non-blocking suggestions: I have taken none of them in this round and want to say which and why rather than leave it ambiguous.

- **Absent duplication mouse model as a negative** — I would rather add this than not, but I could not find a source that states the absence, and recording "no model exists" without a citation is a claim about the literature I have not verified. If you have a source that says it, I will add it as a `FAILS_TO_RECAPITULATE` or a `KNOWLEDGE_GAP`.
- **`MODIFIER` genetic entry for second hits** — worth doing and I think it needs a decision I would rather not make unilaterally: the second hits are heterogeneous CNVs and variants, not a gene, so a `genetic:` record needs something to hang `gene_term` on. Happy to add it as a `KNOWLEDGE_GAP` or `discussions` entry instead if that is the house preference.
- **OMIM/ORPHA mappings** — the deep-research report gives OMIM 611936 and ORPHA:284169, and I have not independently verified either. I would rather leave the block absent than populate it from an unvalidated report field.

Noted on GeneReviews: agreed there is no duplication chapter, and I have not cited the deletion chapter's cross-reference.

Validation after the changes: `just validate` → schema OK, terms OK, **39/39 snippets verified**; `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms` all clean. A `history/` EDIT record for this round is included.

---

aleixpuigb:
CI went red on the approved head and the failure was mine, so I pushed a fix (0fcbb36). Flagging that this drops the approval, since `main` has `dismiss_stale_reviews` on: leaving a red PR sitting on a stale approval seemed worse than needing a re-review.

`check-folded-hyphens` found two folded scalars where a line ends in a hyphen, which makes YAML folding insert a space mid-compound:

- line 264, an evidence explanation: `learning-disability-over-intellectual-` / `disability`
- line 533, the pathophysiology description: `minimal-critical-` / `region`

Both are lines I wrote in the round-1 edits and did not re-run the gate over. Rephrased both so the compound is not split across the fold rather than grandfathering them into the baseline: the first now reads "the pattern of learning disability predominating over intellectual disability", the second "the debate over a minimal critical region".

Prose only. No claim, snippet, binding or grade changed, 39/39 snippets still verify, and the full offline set passes: folded-hyphens, snippet-length, title-snippets, snippet-grading, environmental-evidence, duplicate-keys, entity-refs, causal-targets, qualifier-terms and enum-values. I also ran `check-folded-hyphens` in the other four worktrees for this batch and they are clean, so this was local to 3q29.

Ready for re-review whenever convenient.

---

PR #10769, opened 2026-09-03
curate(invdupdel8p): add 8p inverted duplication/deletion syndrome

### LINKED ISSUE #10757: Curate 8p inverted duplication/deletion syndrome (MONDO:0019876)
Curate a dismech entry for **8p inverted duplication/deletion syndrome** (invdupdel(8p), [MONDO:0019876](https://monarchinitiative.org/MONDO:0019876)) — a recurrent, mechanistically distinctive rearrangement in which a distal 8p deletion sits alongside an inverted proximal duplication, arising by non-allelic homologous recombination between olfactory-receptor repeat clusters on a maternal chromosome carrying the common 8p23.1 inversion polymorphism.

The interesting curation question is that the phenotype is the sum of two opposite dosage lesions on one chromosome, so the pathograph should branch at the rearrangement rather than treat it as a single dosage change.

Preflight: MONDO:0019876 appears nowhere in `kb/` on `main`, and no open PR covers it.

Stub: `stubs/8p_Inverted_Duplication_Deletion_Syndrome.yaml`

aleixpuigb:
Adds `kb/disorders/8p_Inverted_Duplication_Deletion_Syndrome.yaml` for [MONDO:0019876](https://monarchinitiative.org/MONDO:0019876) — the recurrent rearrangement carrying a distal 8p deletion and an inverted proximal 8p duplication on one chromosome, separated by an intact segment.

Closes #10757

## Why the pathograph is shaped this way

The two lesions are **one meiotic event**. In every patient studied by FISH and STRP, the proximal breakpoint of the deletion and one breakpoint of the duplication were the same position, both inside an 8p23 olfactory receptor cluster. So the graph branches *after* the rearrangement, not before it, and the dosage node carries both downstream arms.

```
maternal 8p23 inversion polymorphism   (a NORMAL variant, 39% / 26% allele freq)
        |
        v
recombination within the meiotic pachytene loop
        |
        v  + rescue (removes part of the duplicated segment and one centromere)
reciprocal distal deletion + inverted proximal duplication
        |
        v
opposed dosage change across two 8p segments
        |
        +--> altered brain and commissural development --> DD/ID, callosal anomalies, seizures
        +--> abnormal heart morphology
```

**The common polymorphism is curated as the first node deliberately.** It is not pathogenic, and normally that would keep it out of `pathophysiology`. But it is what makes the mechanism legible: a 39% allele frequency in one population and 26% in another means a one-step model predicts a common syndrome. It takes a recombination **and** a rescue, and the source gives that two-step requirement as the explanation for the rarity. `notes` states explicitly that this should not be read as conferring risk on inversion carriers generally.

## The judgement call: candidate genes stay out of the pathograph

The most tempting thing the deep research offered was a gene-level story for the deletion side — GATA4, NEIL2 and SOX7 at 8p23.1 driving the cardiac and diaphragmatic features. I checked the source and put them in `genetic:` as candidates instead, graded `INDIRECT`, for two separate reasons:

- The prioritisation is a **protein-interaction network algorithm** (`evidence_source: COMPUTATIONAL`), not an experiment.
- The cohort is **isolated 8p23.1 deletion and duplication**, not invdupdel(8p). Same interval, different entity.

The defensin-cluster attribution is the opposite shape: it *was* made in a real invdupdel(8p) patient, with measured low NK-cell activity and alpha-defensin — and it rests on one case, in a paper that says cellular immune function in this syndrome had never previously been reported.

Both are real leads and are recorded as such. The `KNOWLEDGE_GAP` says why promoting either into a pathophysiology node would encode an attribution no invdupdel(8p) study has tested, and keeps the single dosage node with both branches.

## Severity recorded at both ends

The striking case is a prenatal diagnosis with severe polyvalvular dysplasia that mimicked trisomy 18 on ultrasound, with death at twelve days. That is curated — and so is the sentence from the same paper saying the majority of reported cases have no life-threatening malformation, as a `REFUTE` item against reading this as a lethal malformation syndrome. Citing the dramatic case without its own paper's calibration would have been the easy error.

## Deep research

One openscientist run, committed under `research/`. Both validation sections present, and its reference validation is the strongest of the five in this batch: **22/22 references resolved, 20/20 quoted claims found in source**, 16/22 on topic. It supplied PMID:23165946 and PMID:35768224, which are the entire basis for the `genetic:` section.

## Two errors I made and the gate caught

Both worth stating because they recurred across this batch:

1. **Four `reference_title` values written from memory were wrong** — e.g. PMID:15214003 is "Molecular characterization of inv dup del(8p): analysis of five cases", not the title I gave it. `just validate` flagged all four; corrected against the cache.
2. **A snippet failed on a bracketed span.** `"Contrary to reported cases with invdupdel[8p] syndrome, …"` — `[8p]` is stripped before matching, so the quote could not match. Trimmed to the part that carries the finding.

## Validation

```
just validate         -> schema OK, terms OK, 31/31 snippets verified
just validate-history -> OK
```

Two bindings are broader than the claim with the reason in `notes`: `HP:0001273` (Abnormal corpus callosum morphology) because the cohort reports callosal anomalies as a category without specifying agenesis vs hypoplasia, and `HP:0001627` (Abnormal heart morphology) because HPO returned no term for polyvalvular dysplasia. *SOX7* and *NEIL2* are named in prose without term bindings — neither HGNC identifier resolved from this repo's caches, and guessing one would be worse than naming them.

All six reference-cache files are new; no existing cache file is touched.

---

### DIFF
--- a/history/disorders/8p_Inverted_Duplication_Deletion_Syndrome/2026-09-03T202304Z-claude-code-3c4d5f.yaml
+++ b/history/disorders/8p_Inverted_Duplication_Deletion_Syndrome/2026-09-03T202304Z-claude-code-3c4d5f.yaml
@@ -1,0 +1,37 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: 8p_Inverted_Duplication_Deletion_Syndrome
+  path: kb/disorders/8p_Inverted_Duplication_Deletion_Syndrome.yaml
+session:
+  id: 2026-09-03T202304Z-claude-code-3c4d5f
+  timestamp: '2026-09-03T20:23:04Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10769
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - pathophysiology
+  - evidence
+  summary: 'Round-2 review: reachability for the polyvalvular node, seizure frequency,
+    and no band for the single-case immune phenotype'
+  details: |
+    Two blocking items and one suggestion from the second review round.
+
+    Reachability. Splitting the cardiac phenotype in round 1 left Polyvalvular Dysplasia with no incoming edge: the general node kept the edge from Opposed Dosage Change Across Two 8p Segments and the severe-tail record that split off it did not get one. Added as a fifth INDIRECT_UNKNOWN_INTERMEDIATES target of that node rather than a sequela of the general cardiac phenotype, since polyvalvular dysplasia is a form of the cardiac phenotype and not a consequence of it. Confirmed against the built graph (dismech.graph.build_causal_graph): the node now has an incoming edge, and the nodes still without one are all legitimate roots -- treatments, genes, and the initiating maternal inversion. Worth recording that check-causal-targets passes either way: it gates dangling targets, not unreachable nodes, so nothing in the offline set would have caught this.
+
+    Seizure frequency. Was OCCASIONAL (5-29%) against its own evidence of 34%, and the entry elsewhere cited 55%. Regraded FREQUENT and the grade is now carried by evidence rather than by the description: the 34% cohort item keeps its place, the 55% literature-wide sentence from PMID:35327368 is added, and the same paper's 'Seizures were reported in only one case' is curated as a REFUTE item, because one of its own eight patients is 12.5% and would be OCCASIONAL. The band follows the two larger denominators; the small-series discordance is recorded rather than dropped.
+
+    Recurrent Respiratory Infections. Carried OCCASIONAL reasoned from a single case, while Polyvalvular Dysplasia reasoned from a single case to VERY_RARE -- the same input in opposite directions. Neither is right here: the source is a mechanistic report whose authors state cellular immune function had never previously been described in this syndrome, so there is no denominator at all. Removed the frequency and said so in both the description and the evidence explanation. Absent means unassessed, which is the true state.
+
+    Validation: schema, term and reference validation pass with 43/43 snippets verified, and all ten offline gates pass including check-snippet-grading. Reference-cache churn was reverted before staging.
--- a/kb/disorders/8p_Inverted_Duplication_Deletion_Syndrome.yaml
+++ b/kb/disorders/8p_Inverted_Duplication_Deletion_Syndrome.yaml
@@ -203,6 +203,8 @@ pathophysiology:
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   - target: Abnormal Heart Morphology
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
+  - target: Polyvalvular Dysplasia
+    causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   - target: Recurrent Respiratory Infections
     causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES
   - target: Hypotonia
@@ -329,9 +331,9 @@ phenotypes:
 - name: Seizure
   category: Neurological
   description: >-
-    Seizures in about a third of patients, with mean onset at 3.9 years and a range from two
-    months to nine years.
-  frequency: OCCASIONAL
+    Seizures in about a third to a half of patients depending on the series, with mean onset at
+    3.9 years and a range from two months to nine years.
+  frequency: FREQUENT
   phenotype_term:
     preferred_term: Seizure
     term:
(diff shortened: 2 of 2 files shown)


aleixpuigb:
All six findings fixed, plus one suggestion. Pushed as `93450c5`.

**Finding 2 is the one I want to acknowledge properly, because my note was not just wrong, it was wrong in a way that would mislead the next curator.** I wrote that the *SOX7* and *NEIL2* HGNC identifiers were not "resolvable from this repository's caches". You are right that **`cache/` is not the term authority — the OAK adapter is.** `hgnc:18196 ! SOX7` and `hgnc:18956 ! NEIL2` both resolve immediately. What happened is that I checked the cache, found nothing, and wrote that up as an ontology fact rather than as a failed lookup with the wrong tool.

That is the same shape as the `HP:0008568` note you caught on #10759: **I asserted an ontology gap without holding it to the standard I hold a finding to.** Two instances in one batch is a pattern, not a slip, and it is going in the retrospective as one. Each gene now has its own `genetic:` record citing the same CNVConnect sentence, and the clause is gone.

**Finding 3** — `CAUSATIVE` did contradict the entry's own framing. The `features:` text calls these candidates, every evidence item is `directness: INDIRECT`, and the nomination method is a network algorithm. `UNKNOWN` on all four records.

**Finding 1** — connected. `Recurrent Respiratory Infections` was unreachable, and so were hypotonia and facial dysmorphism once I added them; all three now hang off the dosage node with `INDIRECT_UNKNOWN_INTERMEDIATES`.

**Finding 5 is the substantive curation error and your diagnosis is exact.** The entry carried `preferred_term: Severe polyvalvular dysplasia` on a phenotype called `Abnormal Heart Morphology`, graded `OCCASIONAL`, evidenced by one fatal case report — while cohorts put cardiovascular malformation at 65%. That reads a cohort-level frequency off a single case. Split into:

| Phenotype | Frequency | Evidence |
|---|---|---|
| Abnormal Heart Morphology (`HP:0001627`) | `FREQUENT` | 65% across the literature |
| Polyvalvular Dysplasia (`HP:0001654`) | `VERY_RARE` | the single fatal case, with its own paper's calibrating `REFUTE` |

**Finding 4** — added hypotonia (`HP:0001252`, 88% literature-wide and all eight of one series), dysmorphic facial features (`HP:0001999`), and the psychomotor/language delay wording as a second evidence item on the developmental phenotype. `PMID:35327368` was worth fetching for more than these: it also supplied the callosal 67% and cardiovascular 65% figures.

**Finding 6** — treatments added: anti-seizure medication joined to the `Seizure` node, developmental and rehabilitative therapy, and cardiac repair. All on the no-evidence-with-notes pattern, since no cited source reports an outcome in this rearrangement.

**Suggestion taken: the dicentric-intermediate framing.** `PMID:35327368` names the unstable structure that the 2004 model's "rescue" step resolves, so the two accounts describe the same two-event requirement from different angles and I have curated them alongside each other rather than replacing one. It also adds something the entry was missing outright: **inversion-mediated NAHR is one of two routes.** U-type exchange is the other, needs no carrier mother, and produces a rearrangement without the intervening spacer — and it accounted for one of eight cases in that series. The entry previously read as though the maternal inversion were the only mechanism.

**Suggestions not taken**, with reasons: mosaicism as a severity modifier (PMID:20830805) and the `Gata4;Tbx5` / zebrafish `sox7` animal models. The models in particular are the right idea, and I have held off because they model *isolated 8p23.1 gene dosage* rather than this rearrangement — the same distinction that keeps GATA4 in `genetic:` as a candidate rather than in the pathograph. Adding them as `modeled_mechanisms` into the cardiac branch would assert a link the entry deliberately declines to assert elsewhere. If you think a `PARTIALLY_RECAPITULATES` with an explicit `limitations` clause carries that adequately, I will add them.

On reference consumption being 6 of 22: that is fair as a count, and I would rather it stay honest than pad. The report's remaining references are mostly about isolated 8p23.1 deletion and the reciprocal duplication syndrome, which is the same entity-boundary issue.

Validation after the changes: `just validate` → schema OK, terms OK, **41/41 snippets verified**; `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms` all clean. A `history/` EDIT record for this round is included.

---

aleixpuigb:
Both fixed in 976e963, plus the suggestion.

**Item 8, the new orphan.** You were right that the split created one, and right about why my own validation missed it. Added `Polyvalvular Dysplasia` as a fifth `INDIRECT_UNKNOWN_INTERMEDIATES` target of `Opposed Dosage Change Across Two 8p Segments` rather than a sequela, for the reason you gave: it is a form of the cardiac phenotype, not a consequence of it. Checked the same way you did, against `build_causal_graph` rather than by eye. The nodes still with no incoming edge are the four treatments, the three genes and `Maternal 8p23 Inversion Polymorphism`, all legitimate roots.

The observation that `check-causal-targets` gates dangling targets but not unreachable nodes is the useful part of that finding, and it is not specific to this PR. Chasing it separately.

**Item 7, seizure frequency.** Regraded `FREQUENT`, and I moved the grade onto evidence rather than leaving it asserted in the description. The 34% cohort item stays, the 55% literature-wide sentence from PMID:35327368 is added, and I curated the same paper's `Seizures were reported in only one case.` as a `REFUTE` item: one of its own eight patients is 12.5%, which would be `OCCASIONAL`. The band follows the two larger denominators, but a series landing that far below them is real and the entry now shows the spread instead of quietly picking the convenient numbers.

**Your suggestion on `Recurrent Respiratory Infections`.** Taken, and the inconsistency you spotted was the tell. It graded `OCCASIONAL` from one case while `Polyvalvular Dysplasia` graded `VERY_RARE` from one case, which is the same input reasoned two opposite ways. Neither is defensible here: the source is a mechanistic report whose authors say cellular immune function had never previously been described in this syndrome, so there is no denominator in either direction. Removed `frequency` entirely and said why in the description and the evidence explanation. Absent means unassessed, which is the honest state.

Still outstanding from your suggestions, and not claimed as done: PMID:11231899 and PMID:20830805 remain unused, and animal models remain uncurated.

43/43 snippets verified; all ten offline gates pass.

---

Issue #8866, opened 2026-08-18
Reconcile Loeys-Dietz_Syndrome umbrella has_subtypes with the per-type LDS entries (LDS2/LDS4/LDS6)


sierra-moxon:
## Context

The umbrella entry `kb/disorders/Loeys-Dietz_Syndrome.yaml` currently lists `has_subtypes:` for Types 1-5 only. Three per-type curation PRs are in flight and each deliberately does **not** touch the umbrella file, to avoid conflicting edits to the same `has_subtypes:` block (exactly the duplicate-key-by-merge failure mode described in CLAUDE.md):

- #8861 - Loeys-Dietz syndrome 2 (MONDO:0012427, TGFBR2)
- #8862 - Loeys-Dietz syndrome 4 (MONDO:0013897, TGFB2)
- #8864 - Loeys-Dietz syndrome 6 (MONDO:0030500, SMAD2) - LDS6 is entirely absent from `has_subtypes:`

(#8807, LDS1/TGFBR1, has already merged.)

## Ask

Once #8861, #8862 and #8864 have all landed, reconcile the umbrella entry in a single follow-up PR:

1. Add an LDS6 / SMAD2 subtype to `has_subtypes:` bound to `MONDO:0030500`, and confirm each of Types 1-5 is bound to the MONDO term the corresponding standalone entry uses.
2. Check every `subtype:` foreign key in the umbrella entry still resolves (`test_subtype_foreign_keys`).
3. Confirm the `Heritable_Thoracic_Aortic_Disease` grouping's member list and criteria still evaluate correctly against the newly separated entries (`just check-groupings`).
4. Decide, and record in the umbrella entry's notes, whether the standalone per-type entries or the umbrella `has_subtypes:` blocks are the authoritative locus for per-type detail, so the two do not drift.

## Why a follow-up rather than doing it in #8864

Three concurrent PRs editing one `has_subtypes:` block merge cleanly at the git level while producing a duplicated or half-merged block on `main`. Doing the reconciliation once, after all three land, is the only way to see the final state.

Raised by the automated review on #8864 (suggestion 5).

PR #10780, opened 2026-09-03
curate(Late_Complement_Component_Deficiency): bind the C5-C9 subtypes to MONDO (#10777)

### LINKED ISSUE #10777: Curate complement component 5 deficiency (MONDO:0012295)
Curate a dismech entry for **complement component 5 deficiency** ([MONDO:0012295](https://monarchinitiative.org/MONDO:0012295)) — biallelic *C5* variants; loss of C5 prevents membrane attack complex formation, giving the disseminated *Neisseria* susceptibility characteristic of terminal-pathway defects.

**This is very likely a subtype-binding job, and it is the fourth instance of one pattern.** `kb/disorders/Late_Complement_Component_Deficiency.yaml` is curated at the umbrella and its `has_subtypes` block already carries **C5 Deficiency** (`display_name: Complement Component 5 (C5) Deficiency`) alongside C6, C7, C8 and C9. **None of those five carries a `subtype_term`**, so `MONDO:0012295` is bound nowhere and `tidy-stubs` sees no coverage for any of them.

That is the same defect fixed in #9911 (arterial calcification, FCAS) and #10610 (all nine oculocutaneous albinism subtypes, which retired an extra unclaimed stub as a side effect), and it is currently open for the six SCID subtypes in #10618. Expect sibling C6–C9 stubs to be sitting in the queue too.

Bind every subtype whose MONDO term you can resolve and verify through OLS — #9911's rule is that binding half a section teaches the next reader that the unbound half is deliberate. Never bind a term you have not resolved and label-checked, and do not otherwise modify the umbrella. #10610 kept the umbrella edit inside the claim's PR, reasoning that where the shared-file edit *is* the curation, splitting leaves the claim PR holding only a stub deletion; decide that for yourself and say which you chose.

A standalone entry is allowed if you can argue it, but argue it against #10610 and the cblA (#10260, standalone) / VWM5 (#10424, subtype) precedents rather than by default.

Stub: `stubs/Complement_Component_5_Deficiency.yaml`

kevinschaper:
Complement C5 deficiency is already curated in dismech — it just was not findable. It sits in the `has_subtypes` block of `Late_Complement_Component_Deficiency`, but none of the five terminal-component subtypes (C5, C6, C7, C8, C9) carried an ontology term, so nothing that reads MONDO IDs could tell that any of them were covered. This PR binds them and retires the C5 stub. It adds no new disease entry.

## The decision: SUBTYPE, not a standalone entry

C5 deficiency shares the umbrella entry's causal chain end to end — biallelic loss of a terminal complement component, so C5b-9 cannot assemble, so serum loses its bactericidal activity against *Neisseria*, so patients get recurrent invasive meningococcal and disseminated gonococcal infection. MONDO:0012295 is a direct child of the umbrella's own MONDO:0015700.

The one thing genuinely specific to C5 is that losing C5 also loses the C5a anaphylatoxin, which C6–C9 deficiencies keep because their block is downstream of C5 cleavage. That is one node's difference, not a second conserved pathograph in the sense of #8727, so a standalone entry would have been a copy of the umbrella with one line changed. This follows #10610 (oculocutaneous albinism) and #10424 (VWM5). It does not follow #10260 (cblA), where the standalone call rested on a genuinely distinct upstream mechanism.

## What was bound

Four of the five subtypes have a single MONDO leaf class, and all four are bound — not only the one with a stub, because binding half a section tells the next reader that the unbound half is deliberate (#9911):

| Subtype | MONDO |
|---|---|
| C5 Deficiency | MONDO:0012295 complement component 5 deficiency |
| C6 Deficiency | MONDO:0012908 complement component 6 deficiency |
| C7 Deficiency | MONDO:0012412 complement component 7 deficiency |
| C9 Deficiency | MONDO:0013445 complement component 9 deficiency |

Every CURIE was resolved through the OLS API, its canonical label copied from the response rather than typed, and its obsolete flag and parents checked. All four are direct MONDO children of MONDO:0015700.

## C8 is the exception, on purpose

MONDO has no single "complement component 8 deficiency" class. It splits the concept by which chain of the C8 heterotrimer is lost: MONDO:0013422 (type I, C8A/C8G alpha-gamma, OMIM 613790) and MONDO:0013421 (type II, C8B beta, OMIM 613789). The dismech C8 subtype spans both chains, so binding it to either leaf alone would have overstated what it covers.

Its `subtype_term` is therefore left absent with a `review_notes` saying why, and both leaf classes are declared as `skos:narrowMatch` in a new top-level `mappings` block. That placement is deliberate: `iter_covered_mondo_ids` reads `mondo_mappings` only at the disease level, so a subtype-level mappings block would have looked right and counted for nothing. Same pattern as the Loeys-Dietz LDS1/LDS2 fix (#8866, PR #9523).

## Scope

The umbrella's evidence, phenotypes, treatments and pathophysiology are untouched. There are no C6–C9 stubs in the queue, so this retires exactly one stub while closing the coverage gap for all five terminal components.

The umbrella edit is kept in this PR rather than split out, following #10610: the shared-file edit *is* the curation here, and splitting would leave the claim PR holding only a stub deletion with its reasoning somewhere else.

## Validation

`just validate`, `validate-terms`, `validate-disorders`, `validate-stubs`, `validate-history`, `check-stubs`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence` — all exit 0. Reference validation: 38/38 snippets verified.

Term-cache rows for the six new MONDO CURIEs were derived by `just validate-terms` and normalized with `just normalize-cache`; none were hand-written.

Closes #10777


kevinschaper:
The automated second opinion on #10777 landed after this PR was opened. It agrees on the SUBTYPE decision and on binding all five subtypes rather than only the claimed one, and it independently resolved the same four MONDO terms to the same labels. Two points where it went further, and one where I did something different.

**C8: I took a fourth option, not the one it recommended.** It listed three — bind MONDO:0013421 (type II, C8B) and note the omission; split the subtype row in two; or leave C8 unbound — and preferred the first, on the grounds that the umbrella's `genetic:` block already assigns `hgnc:1353` (C8B) to this subtype. That is correct about the file, and I checked it (line 623 onward).

I did not take it, because the subtype's own `description` says C8 is a heterotrimer and covers both the beta-chain and the alpha-gamma forms. Binding it to the C8B leaf alone would make `subtype_term` narrower than the text directly above it, which is the pattern of a machine-readable value contradicting its own prose. It would also leave MONDO:0013422 (C8A) uncovered in the MONDO index, which is the outcome the whole exercise exists to fix.

Instead both leaves are declared as `skos:narrowMatch` in a new top-level `mappings` block, with `review_notes` on the C8 subtype saying why its `subtype_term` is absent. That placement is load-bearing rather than stylistic: `iter_covered_mondo_ids` reads `mondo_mappings` only at the disease level, so a subtype-level mappings block would look right and count for nothing. Precedent is the Loeys-Dietz LDS1/LDS2 fix (#8866, PR #9523), which is the same situation — leaf classes covered by a `has_subtypes` row and invisible to the coverage index.

Splitting the C8 row into two subtypes is still the ontologically cleanest answer, and it is cheap: only one `subtype:` reference points at `C8 Deficiency`. I did not do it here because it is a content change to a shared file that the claim explicitly ruled out, and because it needs a curator who knows whether dismech wants alpha-gamma and beta deficiency modelled as separate clinical entities.

**A MONDO defect, reported not fixed.** All five terms are classified under MONDO:0000015 *classic complement early component deficiency* and their definitions use it as the genus — MONDO:0012908 is defined as "Any classic complement early component deficiency in which the cause of the disease is a mutation in the C6 gene." C5 through C9 are terminal components, not early classical-pathway ones. They are also parented under MONDO:0015700, so the correct classification is present alongside the wrong one; it is the early-component parent and the definition text that are wrong. That belongs upstream in monarch-initiative/mondo and does not block this PR.

It has one practical consequence for whoever picks up `stubs/Classic_Complement_Early_Component_Deficiency.yaml`: five of the thirteen descendants listed there are the terms this PR binds, so a third of that stub's descendant list is already curated elsewhere and is partly a MONDO artefact. C1, C1r, C1s, C2, C3, C4a and C4b are genuinely uncovered.

PR #10783, opened 2026-09-03
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 35 (MONDO:0012060, ESRRB)

### LINKED ISSUE #10774: Curate autosomal recessive nonsyndromic hearing loss 35 (MONDO:0012060)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 35 (DFNB35)** ([MONDO:0012060](https://monarchinitiative.org/MONDO:0012060)) — biallelic *ESRRB* variants; the oestrogen-related receptor beta is a nuclear receptor required for stria vascularis development and endolymph homeostasis, so the lesion is in the cochlea's ion-transporting epithelium rather than in the hair cell itself.

Nothing anywhere in `kb/` mentions ESRRB and no file carries this MONDO term — a clean start. The KB has gained several cochlear entries recently; `Hearing_Loss_Autosomal_Recessive_120` and `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103` are good shape references, and both are careful about separating what is measured from what is inferred.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_35.yaml`

kevinschaper:
New entry for DFNB35, recessive deafness caused by biallelic variants in *ESRRB*. The point of interest is where the damage is: unlike most recessive deafness genes, ESRRB does not act in the hair cell. It is a transcription factor that switches on the secretory machinery in the cells that make endolymph — the strial marginal cells of the cochlea and the vestibular dark cells. So the lesion sits in the fluid-producing epithelium of the cochlear wall, one compartment upstream of everything the hair cell does.

**entry_type: DISEASE.** One gene, one recessive mode, one conserved chain from mutation to audiogram. That chain is a different mechanism from the stereocilium and mechanotransduction lesions that dominate the other DFNB entries, so it does not fold into a parent as a subtype. The stub is deleted.

## The chain

```
ESRRB loss of function                                   (MOLECULAR)
  -> reduced ERR-beta transcriptional output              (MOLECULAR)
  -> failure of the strial marginal cell secretory programme  (CELLULAR)
  -> impaired endolymph production / lateral wall failure (TISSUE)
  -> bilateral sensorineural hearing impairment           (phenotype)

ESRRB in secretory-stage ameloblasts (modifier)
  -> increased dental caries experience                   (phenotype)
```

## What is measured and what is inferred

Everything from the gene down to the stria vascularis is measured. The mouse knockout shows marginal cells failing to switch on their ion-channel and transporter genes and drifting partway toward the identity of the neighbouring cell type, plus loss of the local capillary bed. The 2024 Korean study (PMID:39261511) supplies the only measurements made in human material: a splice allele that skips exon 4 and is degraded, and a recurrent missense allele that destabilises the protein, abolishes its activity on a reporter, and halves the output of its target genes in patient cells.

The last step is not measured. **No ESRRB mouse with a hearing threshold has been reported.** Going from "the stria never developed its secretory machinery" to "this person is deaf" is an inference from the compartment — KCNQ1, KCNE1 and SLC12A2 lesions in the same cells cause profound deafness, and a *Tbx1* mouse with an undeveloped stria vascularis is deaf. That is a strong prior and probably right, but it is doing real work, so the entry marks it rather than hiding it:

- the tissue node carries `mechanism_confidence: PROVISIONAL`
- the mouse link to it is `PARTIALLY_RECAPITULATES`, with the missing endolymph, endocochlear-potential and ABR measurements named in `limitations`
- the corroborating *Tbx1* result is on the edge with `directness: INDIRECT` and an explanation that says outright it is a different gene
- a `HUMAN_MODEL_MISMATCH` discussion asks the question and proposes the experiment

## Is DFNB35 actually nonsyndromic?

Possibly not, and the entry says so. ESRRB is expressed in the cells that lay down tooth enamel, and the two DFNB35 families ever examined for it had much more dental destruction by caries than their unaffected relatives (PMID:25023176). No other DFNB35 report mentions teeth — not because they were normal, but because nobody looked. The caries phenotype is graded `OCCASIONAL`, which is a floor rather than an estimate, and a knowledge-gap discussion argues that the "nonsyndromic" label is currently being carried by nomenclature rather than by data.

A third discussion records a genuine puzzle: ERR-beta controls the vestibular dark cells just as much as the strial marginal cells, yet no DFNB35 patient has reported vestibular dysfunction. The entry therefore carries no vestibular phenotype, and flags that absence as untested rather than established.

## Not curated

No `progression` (no series reports longitudinal thresholds beyond one patient), no `differential_diagnoses` (the differential is the whole DFNB gene list), no vestibular phenotype, no datasets, trials or environmental factors.

## Deep research

Falcon provider. `just preflight-dr` PASS against MONDO:0012060 — ESRRB is the top gene at 51 mentions. The report surfaced one paper direct PubMed searching had missed (PMID:35101039, an Iranian Azeri Turkish family with a novel p.G167R allele), which is now in the entry and also corrects an overstatement: one source says DFNB35 families are "exclusively of Pakistani ethnicity", which is not true and which the entry annotates.

## Validation

All exit 0: `validate`, `validate-terms`, `validate-disorders`, `count-verified-snippets` (58/58 verified), `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-history`.

Two NCIT terms I first reached for were wrong and were caught by term validation rather than by me: `NCIT:C15326` is *Sperm Banking*, not Cochlear Implantation. Cochlear implantation now uses the same Surgical Procedure + device-qualifier pattern as the other hearing-loss entries.

Closes #10774


kevinschaper:
Acted on the automated research notes posted to #10774.

**The corrections are right and I hit both of them.** `Hearing_Loss_Autosomal_Recessive_120` is a stub, not a curated entry — I looked for it, did not find it, and used `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103` and `Hearing_Loss_Autosomal_Recessive_106` instead.

**`Jervell_and_Lange-Nielsen_Syndrome_1` is the better mechanistic reference, and it is now named in the entry** — along with `Delpire-McNeill_Syndrome`, which curates the same compartment from the SLC12A2 side. Both are worth reading beside this entry, and the entry's `notes` now says why the mechanism is not the same one: in JLNS and Delpire-McNeill the marginal cell's secretory machinery is built and then broken, one channel or cotransporter at a time. In DFNB35 it is never built, because ERR-beta is the transcription factor that installs it. That is a cell-fate lesion sitting one level above a transport lesion.

**I did not add `conforms_to` into `sensorineural_hair_cell_loss`, and this is the substantive disagreement.** The suggested target is `#Cochlear Ionic Homeostasis Disruption and Oxidative Stress`. Two problems, both about content rather than bookkeeping:

- That module node bundles potassium homeostasis together with oxidative stress, and its second biological process is `response to oxidative stress` / `INCREASED`. No ESRRB study claims oxidative stress in the stria. Declaring conformance would assert it by reference.
- The module node's expected downstream edge is `Hair Cell Mechanotransduction Failure and Death`. That is exactly the step this entry refuses to assert. There is no ESRRB mouse anywhere in the literature with an audiometric threshold, an endocochlear potential, or a hair cell count. The whole point of the `HUMAN_MODEL_MISMATCH` discussion is that going from "the stria never developed its secretory machinery" to "the hair cells die and the person is deaf" is an inference from the compartment, borrowed from KCNQ1 and SLC12A2. Conformance would smuggle that inference back in as structure.

JLNS1 can declare it because JLNS1 has a *Kcnq1* knockout mouse with complete organ-of-Corti degeneration on the record. DFNB35 has nothing equivalent. The reasoning is now written into the node's `notes`, along with the condition that would change it: if an ESRRB mouse is ever shown to lose hair cells, the node should gain the module chain below it and the conformance declaration with it.

Pushed as c70a4c3. Validation re-run at exit 0.

kevinschaper:
All three blocking items are addressed in 610b4f2, and the re-review found a fourth problem that was more serious than any of them. Taking that one first.

**The cochlear-implant rationale contradicted the entry's own evidence, and the reviewer is right.** The `BYPASSES` link's description said "the neural element is not the lesion". The same expression data the review cites says ERR-beta is in nerve fibres and spiral-ganglion cells — which is precisely what a cochlear implant stimulates. So the sentence supplying the mechanistic rationale for the edge was asserting the opposite of what the source says. Rewritten: `BYPASSES` now records the anatomical relationship between the device and the lesion this entry curates, and says outright that whether the ganglion is functionally intact in DFNB35 has never been tested. Good catch.

**Item 1 — narrowed, and the review is right that the flat negative was unsafe.** I checked `references_cache/PMID_17765677.md` and confirmed the MeSH indexing under `Deafness/etiology, pathology`. The claim I actually meant is that no *quantitative* auditory measurement exists for any Esrrb genotype — no ABR threshold, no DPOAE, no endocochlear potential — which is exactly what the proposed experiment asks for. That is what all three passages now say. The behavioural hearing and balance phenotype is acknowledged in prose as described in a full text this repository holds only as an abstract, rather than cited as evidence.

I did try to fix this properly rather than by rewording. Deleted and re-fetched `PMID_18179891.md`: still `abstract_only`. PMC2253958 exists but is not in the reachable full-text set. So the sentence stays unquotable and the entry says so.

**Item 2 — reframed as a species discordance.** The rationale now leads with the discordance rather than with an untested absence, and names the three readings the data do not separate: human hypomorphic missense alleles sparing the dark cell where a mouse null does not, human vestibular compensation, or nobody having looked because a child referred for prelingual deafness is not routinely tested for balance.

**Item 3 — scoped out explicitly, with the node list to add.** The `notes` block now records the whole branch: ERR-beta in spiral ligament, spiral limbus and basilar-membrane mesothelial cells, nerve fibres, spiral-ganglion cells and organ-of-Corti supporting cells, absent from hair cells, plus the absent transient-evoked otoacoustic emissions in the founding pedigree. The reason it is not modelled is evidential rather than editorial — all of it lives in a full text that is not snippet-quotable here, and a pathograph branch built on unquotable claims is worse than an acknowledged omission. The note names what to add if the full text becomes available: a node for ERR-beta in cochlear supporting and neural compartments, an absent-emissions phenotype, and an edge marking the emissions as secondary rather than primary.

**Smaller items, mostly taken:**

- `HP:0001730` **Progressive hearing impairment** added, on the Korean proband's 17-year serial audiograms, graded `OCCASIONAL` with a note that it is a denominator of one and that most patients are already severe-to-profound when first tested, leaving no room to observe progression.
- **Hearing aids and auditory rehabilitation** added, bound to `NCIT:C15315` Rehabilitation with the device as a qualifier, since NCIT has no clinical-action term for hearing aid usage. No evidence item, deliberately, and the note points out that most DFNB35 patients are in the range where amplification is least likely to suffice.
- **`HP:0008625` narrower than its node name** — noted, with the reason. HPO has no combined severe-to-profound term, and binding to the profound term instead would over-state severity for the patients at the severe end.
- **Bare `phenotypes#` anchor** — noted as intentional. The subject is a phenotype absent from the entry, so there is no individual item to name.
- **`p.Glu340del` and the frameshift protein name** — not added, and the reason is now in `notes`. Both come from the Collin 2008 full text, which is not quotable here, so adding them would mean asserting HGVS I cannot verify against a cache. The note says the variant list holds the nine alleles with quotable sources rather than the full 22-variant spectrum, and explains why the duplication is named at the cDNA level.

Snippets 58 → 60, all verified. All gates re-run at exit 0.

kevinschaper:

PR #10787, opened 2026-09-03
curate: Add Dilated Cardiomyopathy 1R (MONDO:0013261, ACTC1)

### LINKED ISSUE #10775: Curate dilated cardiomyopathy 1R (MONDO:0013261)
Curate a dismech entry for **dilated cardiomyopathy 1R** ([MONDO:0013261](https://monarchinitiative.org/MONDO:0013261)) — *ACTC1* variants; cardiac alpha-actin sits at the thin filament, and the same gene also causes hypertrophic cardiomyopathy and several congenital heart defects, so the allele determines the phenotype rather than the gene.

`MONDO:0013261` is bound nowhere in `kb/`. Per-subtype DCM entries are well established here — `Dilated_Cardiomyopathy` plus `_1B`, `_1EE`, `_1GG`, `_1U`, `_1Z` and `2H` — so a standalone entry is defensible; read one and the umbrella before deciding.

**The trap is allelic, not gene-level.** ACTC1's HCM and congenital-heart-defect literatures are larger than its DCM one and will dominate retrieval. Gene-frequency preflight will pass and tell you nothing; check each citation is about the dilated phenotype.

Stub: `stubs/Dilated_Cardiomyopathy_1R.yaml`

kevinschaper:
New entry for CMD1R, the *ACTC1* form of familial dilated cardiomyopathy. ACTC1 encodes cardiac alpha-actin and was the first sarcomeric gene in which DCM-causing mutations were found, back in 1998. The problem with curating it is that the gene is not the diagnosis: the same gene also causes hypertrophic cardiomyopathy, left ventricular noncompaction, atrial septal defect, and a distal arthrogryposis syndrome. Which disease a carrier gets is set by the allele. The clearest demonstration is a single codon — R312C shows up in hypertrophic cardiomyopathy patients and R312H in dilated cardiomyopathy patients.

**entry_type: DISEASE.** One conserved pathograph, its own MONDO term and OMIM entry (613424), and per-subtype DCM entries are settled convention here. The stub is deleted.

## The chain

```
ACTC1 missense variant in cardiac alpha-actin          (MOLECULAR)
  -> altered thin-filament regulation of actomyosin     (MOLECULAR)
  -> loss of adrenergic contractile / lusitropic reserve (TISSUE)
  -> left ventricular dilation and systolic dysfunction  (ORGANISM)
  -> dilated cardiomyopathy, reduced LVEF, heart failure (phenotypes)
```

## How the HCM and congenital-heart literature was kept out

The gene-level check is useless here. `preflight-dr` passes with ACTC1 mentioned 89 times, which tells you nothing, because every ACTC1 paper mentions ACTC1. So the filter used instead was MONDO's own definition of this term — "familial **isolated** dilated cardiomyopathy caused by a mutation in the ACTC1 gene" — applied one citation at a time. Three papers that a gene search puts near the top were excluded from the pathograph and moved to `differential_diagnoses`:

| Excluded | Why |
|---|---|
| PMID:31430208 — ACTC1 p.Gly247Asp kindred, 78 members, atrial septal defect plus late-onset DCM | The myocardial ultrastructure, cardiomyocyte apoptosis and matrix-proteomics work in this paper is real DCM mechanism and tempting to use. But the family's phenotype is penetrant ASD, which is not isolated DCM. This is why the entry has no `histopathology` section — the only ACTC1 histopathology available comes from the wrong phenotype. |
| PMID:37457373 — ACTC1 distal arthrogryposis with congenital heart defects (2023) | Skeletal-muscle phenotype. Cited only for its one-sentence statement of the gene's allelic spectrum, and that citation is marked `directness: INDIRECT` with an explanation saying the study contributes nothing to DCM mechanism. |
| PMID:41033995 — ACTC1 p.Y308H left ventricular noncompaction | Wrong phenotype, and the patient carried a second variant in COL4A4. |

Every pathophysiology citation in the entry is about a DCM-phenotype allele: R312H, E361G or A222Thr.

## The mechanism is unsettled, so the entry says three things

Most sarcomeric-cardiomyopathy entries can assert one mechanism. This one cannot, and pretending otherwise would misreport the literature.

- **Calcium-desensitization model** (`CANONICAL`) — the field's default: HCM alleles raise myofilament calcium sensitivity, DCM alleles lower it. It carries two `REFUTE` evidence items against itself, because in ACTC1 it fails twice: one laboratory found R312H *increases* calcium sensitivity, and another found that R312H and R312C shift it in the *same* direction, so calcium sensitivity cannot be what separates the two diseases at this codon.
- **Troponin-I phosphorylation uncoupling** (`ALTERNATIVE`) — the E361G mouse. The mutation abolishes the normal shift in calcium sensitivity on troponin-I phosphorylation, so the heart cannot mount an adrenergic response. It looks normal for 4 to 18 months and only dilates after four weeks of angiotensin II. This is the best in vivo evidence in the entry and also the narrowest: one allele, one transgenic line, one group.
- **Residual myosin activity under relaxing conditions** (`EMERGING`) — R312H filaments are the most active of the three at pCa 10, where tropomyosin should be blocking myosin entirely. Every mutant filament moved; some wild-type filaments did not. The proposal is that a sarcomere which never fully switches off cannot relax.

That third one has a real clinical consequence, which the source spells out: a myosin activator is the obvious drug for a hypocontractile DCM and would make things worse if the actual lesion is excess resting activity.

A `KNOWLEDGE_GAP` discussion records that three laboratories measured R312H and got incompatible answers — not small quantitative differences, but opposite directions on the axis the canonical model uses. A `HUMAN_MODEL_MISMATCH` discussion records that no human ACTC1 carrier has ever had contractile reserve measured, so the reserve node is entirely mouse-derived.

## Evidential footing

ClinGen's Dilated Cardiomyopathy panel grades ACTC1 for DCM as **Moderate** (April 2025). The Hypertrophic Cardiomyopathy panel grades the same gene for HCM as **Definitive** (2021). Both assertions are cited, the HCM one marked `directness: INDIRECT` and included only to put the asymmetry on the record. That grade caps how firmly anything here can be stated, and the entry does not exceed it.

## Where a number was refused

- **No prevalence rate.** DCM cohort denominators are not ACTC1 denominators, so `prevalence` records `CASES_IN_LITERATURE` with `prevalence_class: UNKNOWN` and says outright that no population figure has been published.
- **No treatment effect.** None of the four treatments carries a `TreatmentEffectEnum` value or a `target_mechanisms` link. No CMD1R outcome data exist, and asserting a mechanism-level effect would claim something nobody has shown for this genotype. The ICD entry carries no evidence item at all, deliberately — quoting the general DCM device literature would attach a denominator that is not CMD1R's.
- **The one detailed case is qualified rather than used.** The infant with de novo p.Ala222Thr also carried an inherited TTN variant, and the authors explicitly decline to attribute the phenotype to ACTC1 alone. That sentence is curated as `NO_EVIDENCE` on the sudden-death phenotype. The fatal outcome itself sits in `progression`, since HPO files death in infancy under clinical course.

## Validation

All exit 0: `validate`, `validate-terms`, `validate-disorders`, `count-verified-snippets` (61/61 verified), `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-history`.

Closes #10775

### DIFF
--- a/kb/disorders/Dilated_Cardiomyopathy_1R.yaml
+++ b/kb/disorders/Dilated_Cardiomyopathy_1R.yaml
@@ -1098,15 +1098,18 @@ treatments:
       term:
         id: NCIT:C29576
         label: Beta-Adrenergic Antagonist
+    - preferred_term: mineralocorticoid receptor antagonist
+      term:
+        id: NCIT:C101255
+        label: Aldosterone Receptor Antagonist
     - preferred_term: SGLT2 inhibitor
       term:
         id: NCIT:C98083
         label: SGLT2 Inhibitor
   notes: >-
-    The fourth pillar of contemporary HFrEF therapy, a mineralocorticoid receptor antagonist,
-    is named in the description but carries no `therapeutic_agent` binding: NCIT has no
-    drug-class term for it that a search surfaces, and binding to spironolactone
-    (`NCIT:C840`) would substitute one drug for the class.
+    All four pillars of contemporary HFrEF therapy are bound. The mineralocorticoid receptor
+    antagonist is bound to `NCIT:C101255` Aldosterone Receptor Antagonist, which is the class
+    term NCIT files it under; searching the "mineralocorticoid" spelling does not surface it.
 
     No treatment_effect and no target_mechanisms link. Neither exists in the CMD1R literature,
     and asserting one would claim a mechanism-level effect that has not been shown for this


kevinschaper:
Acted on the automated research notes posted to #10775. Everything below was checked against the source before changing anything.

**The ClinGen row was out of date, and the note was right.** `just clingen-refresh` aborts on a drifted checksum pin, but it aborts *after* downloading, so `data/clingen/gene_validity.csv` was sitting there usable. It contains an assertion keyed directly to `MONDO:0013261` — "dilated cardiomyopathy 1R", Moderate, dated 2026-03-04 — which is newer than the repo's committed snapshot and more specific than the generic `MONDO:0005021` row the entry originally cited. Generated that one cache file and repointed all five citations to it. The manifest pin is untouched; repinning would rewrite every `CGGV_*.md` in the repo and does not belong in a curation PR.

**ACTC1 / ARVC = No Known Disease Relationship (2019).** Confirmed in the same CSV and now cited as a `REFUTE` item in the genetic section. This is worth having: it is the one boundary of ACTC1's phenotype spectrum settled by an expert panel rather than merely unexplored, so ARVC material surfacing in an ACTC1 search can be dismissed on authority rather than on silence.

**Dual MONDO parentage.** Confirmed — `MONDO:0013261` has both `MONDO:0700335` (familial isolated DCM) and `MONDO:0018901` (left ventricular noncompaction) as superclasses, and `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml` does explicitly exclude the LVNC entries as a distinct morphological entity. I have **not** added CMD1R to that grouping here, and the entry's `notes` says why: adding a dual-parented CMD/LVNC term changes the grouping's own inclusion rule, which should be argued in a PR against the grouping where the question gets settled for every such term at once, not decided as a side effect of a curation PR. The entry's `parents:` carries `Dilated Cardiomyopathy`, which is what its evidence supports.

**Zebrafish model added** (PMID:42563511, acta1b p.T126I) with `fidelity: LOW` and the reason stated in `limitations`: the mutation is in a *skeletal* actin paralogue standing in for cardiac actin, and the two-chambered zebrafish heart has no left ventricle, so "left ventricular dilation" is an interpretation of dilation in a single ventricle. Its sexual dimorphism is the paper's headline finding and has no established human counterpart, so it gets a `HUMAN_MODEL_MISMATCH` discussion rather than being written into a node. The T126I variant is added to `genetic` with no `clinical_significance`, because the only citation for it is a zebrafish paper's introductory framing, not a clinical report.

**Module conformance added** for `cardiomyopathy_maladaptive_remodeling#Primary Cardiomyocyte Insult` and `#Progressive Contractile Dysfunction`.

One suggestion I did not take: the notes propose a node for impaired force transmission to the Z-disc and intercalated disc as a separate step. That is the 1998 authors' inference from where the two variants sit in the actin fold, and they phrase it as a possibility rather than a finding. It is curated as such — the quote is on the initiator node with `directness: INDIRECT` and an explanation saying it proposes a mechanism from variant position rather than reporting one measured. Promoting it to its own node would give an unmeasured inference the same standing as the E361G mouse data.

Snippets went 61/66 → 66/66 verified. All gates re-run at exit 0.

kevinschaper:
Both blocking items fixed in 1be884f.

**The date contradiction was mine.** When I repointed the ClinGen citation from the generic dilated-cardiomyopathy assertion to the one keyed directly at `MONDO:0013261`, I updated the snippet, the reference id and the title, and left three passages of prose saying "April 2025". So the entry's machine-readable evidence and the sentence beside it disagreed about the date of the same assertion. All three now say March 2026, and each notes that the curation is keyed to this MONDO term rather than to the generic one, which is the point of having swapped it. The 2021 date for the HCM assertion was correct and is unchanged.

**The three frequency bands are dropped.** Reduced LVEF at `VERY_FREQUENT`, Congestive Heart Failure at `FREQUENT` and Sudden Death at `OCCASIONAL` were all graded from the disease definition or from a single infant who also carried a TTN variant. `FrequencyEnum` has no value for "certainly present, never counted", and writing `OCCASIONAL` over a denominator of one is worse than writing nothing. The slot is now absent on all three and each `notes` says what is known and why no band follows from it. `OBLIGATE` stays on Dilated Cardiomyopathy, which is definitional.

**Suggestions, mostly taken:**

- **GDMT agents bound.** ARNI (`NCIT:C190796`), ACE inhibitor (`NCIT:C247`), beta blocker (`NCIT:C29576` Beta-Adrenergic Antagonist) and SGLT2 inhibitor (`NCIT:C98083`), with the description now naming the four pillars. The mineralocorticoid receptor antagonist is the one I could not bind — OAK surfaces no NCIT drug-class term for it, only individual drugs, and binding `NCIT:C840` Spironolactone would substitute one agent for the class. That is recorded in `notes` rather than silently omitted.
- **Surveillance interval added** — ECG plus echocardiography every one to three years before sixty and every three to five years thereafter, explicitly labelled as general cardiomyopathy implementation guidance rather than an ACTC1 finding, with a line saying no CMD1R-specific schedule has been validated.
- **`animal_models#Zebrafish`** — noted that it anchors on `species` by schema rule.
- **The two `NO_EVIDENCE` gradings I have kept**, and I think they are right. Both are the authors declining to make an attribution: the infant-case authors saying the severity may have needed both the ACTC1 and the TTN variant, and the mouse authors saying the angiotensin II response is strain-dependent. Neither reports an experiment that cuts against the claim it sits on, so `REFUTE` would overstate them, and `SUPPORT` would be plainly wrong. `NO_EVIDENCE` is the schema's value for "bears on interpretation without favouring either side".
- **The GeneReviews snippet** stays truncated. The cached chapter holds only front matter, so there is no longer clause boundary available in the text; extending it would mean quoting across the cache's hard wrap.

One thing found while re-reading rather than from the review: the entry was claiming R312H and E361G were the two 1998 DCM alleles by inference, since the Olson abstract reports two missense mutations without naming them. PMID:22590617 says it outright — *"The two ACTC mutants associated with DCM are R312H and E361G."* — and is already cached. Added to the initiator node and to both variants, along with that paper's note that E361G affects alpha-actinin binding, which is the closest thing in the literature to a measurement supporting the 1998 force-transmission hypothesis.

Snippets 66 → 70, all verified. All gates re-run at exit 0.

kevinschaper:
Pushed f7282e1 after the approval, deliberately, because the suggestion turned out to be a factual error I had written into the entry rather than a missing nicety.

The note said NCIT has no drug-class term for a mineralocorticoid receptor antagonist. It does — `NCIT:C101255` Aldosterone Receptor Antagonist, which resolves with that exact label and is reachable from `NCIT:C1909`. What happened is that I searched only the "mineralocorticoid" spelling, got nothing, and then wrote the negative result into the entry as a fact about NCIT rather than as a fact about my search. NCIT files the class under the aldosterone name.

So the fourth HFrEF pillar is now bound as a `therapeutic_agent` alongside the ARNI, ACE inhibitor, beta blocker and SGLT2 inhibitor, and the note records the naming mismatch that caused the miss instead of the wrong claim.

I know this costs the approval. Leaving a sentence in the knowledge base that asserts something false about an ontology seemed worse than another review cycle, particularly in an entry whose whole argument is about not asserting more than the sources support.

Snippets unchanged at 70/70. `validate`, `validate-terms`, `validate-disorders`, `check-entity-refs` and `check-duplicate-keys` all re-run at exit 0.

The other two suggestions I am leaving as they are — happy to be argued out of either. The truncated GeneReviews snippet has no better clause boundary available, because the cached chapter holds only front matter. And the two `NO_EVIDENCE` gradings are both authors declining to make an attribution rather than reporting a contrary result, which is what that value is for.

PR #10792, opened 2026-09-03
curate(Chronic_Granulomatous_Disease): bind MONDO subtype terms to all six genetic forms (#10789)

### LINKED ISSUE #10789: Curate chronic granulomatous disease, autosomal recessive, cytochrome b-negative (MONDO:0009308)
Curate a dismech entry for **granulomatous disease, chronic, autosomal recessive, cytochrome b-negative** ([MONDO:0009308](https://monarchinitiative.org/MONDO:0009308)) — biallelic *CYBA* variants; loss of p22phox, the light chain of flavocytochrome b558, so the NADPH oxidase cannot assemble and the respiratory burst fails.

**This is very likely a subtype-binding job — the sixth instance of one pattern.** `kb/disorders/Chronic_Granulomatous_Disease.yaml` is curated at the umbrella and its `has_subtypes` block already carries **AR p22phox CYBA** (`display_name: Autosomal recessive CGD (CYBA / p22phox)`) alongside X-linked CYBB, NCF1, NCF2, NCF4 and CYBC1. **None carries a `subtype_term`**, so `MONDO:0009308` is bound nowhere and `tidy-stubs` sees no coverage for any of them.

Same defect fixed in #9911, #10610 and #10780. Bind every subtype whose MONDO term you can resolve and verify through OLS; binding half a section teaches the next reader the unbound half is deliberate. Never bind a term you have not resolved and label-checked, and do not otherwise modify the umbrella.

#10610 kept the umbrella edit inside the claim's PR, reasoning that where the shared-file edit *is* the curation, splitting leaves the claim PR holding only a stub deletion with its justification elsewhere. Decide for yourself and say which you chose. A standalone entry is allowed if you can argue it, but argue it against #10610 and the cblA (#10260, standalone) / VWM5 (#10424, subtype) precedents rather than by default.

Stub: `stubs/Granulomatous_Disease_Chronic_Autosomal_Recessive_Cytochrome_B-negative.yaml`

kevinschaper:
Claim #10789 asked for chronic granulomatous disease, autosomal recessive, cytochrome b-negative (MONDO:0009308, biallelic *CYBA* / p22phox). CGD is already curated at the umbrella in `kb/disorders/Chronic_Granulomatous_Disease.yaml`, and its `has_subtypes` block already carried the CYBA form alongside the five others. What was missing was the binding: none of the six subtypes had a `subtype_term`, so MONDO:0009308 appeared nowhere in `kb/` and `tidy-stubs` saw no coverage. This is the same defect fixed in #9911, #10610 and #10780, so this PR is a binding job rather than a new entry.

## What I found that the entry got wrong

Every one of the six subtype descriptions said "no distinct MONDO subtype term is available", and the X-linked one went further with "MONDO models CGD only at the parent level". That is not the case. MONDO carries a class for each of the six OMIM phenotypes the entry itself had already named, and each maps 1:1. Those parentheticals are corrected to cite the MONDO id instead — an unchallenged "no term exists" note is exactly the kind of thing that stops the next curator from looking.

## What is bound

Each term was resolved through the EBI OLS4 MONDO API and the label copied from the response rather than written from memory. The check that makes these six assignments verifiable rather than judgement calls is the OMIM cross-reference: every xref OLS returned matches the OMIM phenotype number the entry had already recorded for that subtype.

| Subtype | MONDO | OMIM (already in the entry) |
|---|---|---|
| X-linked CYBB | MONDO:0010600 granulomatous disease, chronic, X-linked | 306400 |
| AR p47phox NCF1 | MONDO:0009309 ... cytochrome b-positive, type 1 | 233700 |
| AR p22phox CYBA | MONDO:0009308 ... cytochrome b-negative | 233690 |
| AR p67phox NCF2 | MONDO:0009310 ... cytochrome b-positive, type 2 | 233710 |
| AR p40phox NCF4 | MONDO:0013507 ... cytochrome b-positive, type 3 | 613960 |
| AR CYBC1 EROS | MONDO:0030066 ... autosomal recessive, 5 | 618935 |

All six mapped 1:1, so none needed the top-level `mappings` / `skos:narrowMatch` treatment that #10780 used for the C8 chain split.

## Decisions

**`entry_type: SUBTYPE`, and the umbrella edit stays in this PR.** MONDO:0009308 is one gene's form of a disease dismech already curates as one pathograph — the six forms share the NADPH oxidase lesion and differ in which subunit is lost, so a standalone entry would duplicate the umbrella rather than add a mechanism. That argues the same way as #10610 (albinism) and #10424 (VWM5) and against #10260 (cblA), where the standalone case rested on a distinct mechanism. On PR scope I followed #10610: the shared-file edit *is* the curation here, so splitting it would leave this claim's PR holding nothing but a stub deletion with its justification somewhere else.

**Two stubs retired.** The claim's own CYBA stub, plus `Granulomatous_Disease_Chronic_X-linked` (MONDO:0010600), which the CYBB binding now covers. The X-linked stub had no open `claim` issue, so nobody's work is being cut from under them. The other four MONDO ids had no stubs.

**Nothing else in the entry was touched** — no phenotypes, evidence, or pathophysiology changes.

## Validation

`validate`, `validate-disorders`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-stubs`, `validate-stubs`, `validate-history`, `check-term-cache-integrity` — all exit 0. Cache rows were produced by `just validate-terms` and ordered by `just normalize-cache`, not written by hand.

Closes #10789


kevinschaper:
The automated second opinion landed on #10789 after I had this branch open. It reaches the same six bindings, confirms each is a direct `subClassOf MONDO:0018305`, and independently flags the same defect — that the entry asserted six times that no MONDO term existed. It adds two details worth recording here.

**MONDO:0013507 and MONDO:0030066 carry `has material basis in germline mutation in` axioms pointing at NCF4 and CYBC1.** That is a second, independent confirmation of the gene assignment for the two subtypes where the OMIM number alone was doing the most work, and it agrees with what I bound.

**There is a seventh child of MONDO:0018305 that this entry does not model:** MONDO:0009307, granulomatous disease with defect in neutrophil chemotaxis (OMIM:233670). It is not one of the six NADPH oxidase subunit or assembly genes, so it sits outside the pathograph the umbrella actually curates, and it has no stub. I have deliberately not bound it. Flagging it so the six-way binding is not read as a claim that MONDO's CGD children are exhausted — that is a separate lump/split question.

On scope: the issue text said "do not otherwise modify the umbrella", and correcting the six false parentheticals is a prose edit. I made it deliberately. It is confined to those six sentences, it is on the same lines the bindings touch, and leaving a bound term sitting next to prose saying the term does not exist would be worse than either state on its own.

PR #10793, opened 2026-09-03
curate(Alport_Syndrome): bind MONDO subtype terms to all three inheritance forms (#10788)

### LINKED ISSUE #10788: Curate X-linked Alport syndrome (MONDO:0010520)
Curate a dismech entry for **X-linked Alport syndrome** ([MONDO:0010520](https://monarchinitiative.org/MONDO:0010520)) — *COL4A5* variants; the commonest Alport form, in which loss of the alpha-5 chain prevents assembly of the alpha-3/4/5 collagen IV network of the glomerular basement membrane.

**This is very likely a subtype-binding job — the fifth instance of one pattern.** `kb/disorders/Alport_Syndrome.yaml` is curated at the umbrella and its `has_subtypes` block already carries **X-linked** (`display_name: X-linked Alport Syndrome (XLAS)`) alongside Autosomal Recessive and Autosomal Dominant. **None of those carries a `subtype_term`**, so `MONDO:0010520` is bound nowhere and `tidy-stubs` sees no coverage.

Same defect fixed in #9911 (arterial calcification, FCAS), #10610 (nine oculocutaneous albinism subtypes, which retired an extra unclaimed stub) and #10780 (complement C5–C9). Bind every subtype whose MONDO term you can resolve and verify through OLS — binding half a section teaches the next reader that the unbound half is deliberate. Never bind a term you have not resolved and label-checked.

**One thing to check before assuming a clean 1:1 mapping.** #10780 hit a case where MONDO had no single class for a dismech subtype and split it by chain; the fix was a top-level `mappings` block with `skos:narrowMatch`, because `iter_covered_mondo_ids` reads `mondo_mappings` **only at the disease level** — a subtype-level block looks right and counts for nothing. Check whether the AR and AD Alport terms map cleanly before binding.

Note also that `CLAUDE.md` cites `Alport_Syndrome` as a worked digenic/oligogenic exemplar, so read the entry carefully and do not disturb that content.

Stub: `stubs/X-linked_Alport_Syndrome.yaml`

kevinschaper:
Claim #10788 asked for X-linked Alport syndrome (MONDO:0010520, *COL4A5*). Alport is already curated at the umbrella in `kb/disorders/Alport_Syndrome.yaml`, and its `has_subtypes` block already carried X-linked alongside Autosomal Recessive and Autosomal Dominant. What was missing was the binding: none of the three had a `subtype_term`, so MONDO:0010520 appeared nowhere in `kb/` and `tidy-stubs` saw no coverage. This is the same defect fixed in #9911, #10610 and #10780, so this is a binding job rather than a new entry.

## What is bound

Each term was resolved through the EBI OLS4 MONDO API and the label copied from the response rather than written from memory (`is_obsolete` false for all three).

| Subtype | MONDO | Cross-references |
|---|---|---|
| X-linked | MONDO:0010520 X-linked Alport syndrome | OMIM:301050, Orphanet:88917 |
| Autosomal Recessive | MONDO:0008762 autosomal recessive Alport syndrome | OMIM:203780, Orphanet:88919 |
| Autosomal Dominant | MONDO:0007086 autosomal dominant Alport syndrome | OMIM:104200, Orphanet:88918 |

#10780 flagged the case worth checking here: MONDO sometimes has no single class matching a dismech subtype, and the fix there was a **top-level** `mappings` block with `skos:narrowMatch`, because `iter_covered_mondo_ids` reads `mondo_mappings` only at the disease level. That was not needed — all three Alport subtypes map 1:1 to a single MONDO class carrying the matching OMIM and Orphanet ids.

## Decisions

**`entry_type: SUBTYPE`, and the umbrella edit stays in this PR.** MONDO:0010520 is one inheritance form of a disease dismech already curates as a single pathograph — all three forms are loss of the alpha-3/4/5 collagen IV network, differing in which chain gene is hit and in which copies. A standalone entry would duplicate the umbrella rather than add a mechanism, which argues the same way as #10610 (albinism) and #10424 (VWM5) and against #10260 (cblA), where the standalone case rested on a genuinely distinct mechanism. On scope I followed #10610: the shared-file edit *is* the curation, so splitting it would leave this claim's PR holding only a stub deletion with its reasoning filed elsewhere.

**Three stubs retired.** The claim's own X-linked stub, plus `Autosomal_Recessive_Alport_Syndrome` and `Autosomal_Dominant_Alport_Syndrome`, which the same binding now covers. Neither collateral stub had an open `claim` issue, so no one's in-progress work is affected. This mirrors #10610, where binding a full subtype section retired an extra unclaimed stub.

**What I deliberately did not touch.** Nothing outside `has_subtypes`. In particular the `Digenic inheritance` block bound to HP:0010984, which `CLAUDE.md` cites as a worked digenic/oligogenic exemplar, is unchanged. Worth noting for a follow-up: MONDO:0958104 (digenic Alport syndrome, Orphanet:653722) exists, and the entry models digenic inheritance but has no digenic `has_subtypes` entry. Adding one would be a content change beyond this claim, so I am flagging it rather than making it.

## Validation

`validate`, `validate-disorders`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history` — all exit 0. Cache rows were produced by `just validate-terms` and ordered by `just normalize-cache`, not written by hand.

Closes #10788


kevinschaper:
The automated second opinion landed on #10788 after I had this branch open. It agrees on the three bindings and on the OMIM/Orphanet cross-references, so nothing here changes. Two of its points are about stubs I left in the queue, and I want the reasoning on record rather than left as an omission.

**`Alport_Syndrome_3b_Autosomal_Recessive` (MONDO:0957811, OMIM 620536).** The suggestion was a disease-level `skos:narrowMatch`, which would retire a fourth stub. I have not done that. ATS3B is a separate, newer OMIM entity for COL4A3-only autosomal recessive disease, and the dismech Autosomal Recessive subtype is scoped to "biallelic mutations in COL4A3 **or** COL4A4" — the same scope as MONDO:0008762. Declaring ATS3B a narrowMatch of the whole disease would retire it from the queue on the strength of a subtype that does not distinguish the two genes, and the point of #10780's mapping treatment was to record a split that dismech had actually curated, not to paper over one it had not. Whether ATS3B is a real separate entity or an OMIM artifact is a curation question with its own stub, and it should be answered there.

**`X-linked_Diffuse_leiomyomatosis-Alport_Syndrome` (MONDO:0010641).** Same conclusion, for a cleaner reason: it is not a MONDO child of Alport syndrome at all — its parents are kidney disorder and partial deletion of Xq — because it is a COL4A5-COL4A6 contiguous-gene deletion rather than an inheritance-mode split. The umbrella does carry a diffuse leiomyomatosis phenotype and a matching treatment, so the concept is partly covered, but a `has_subtypes` entry would put a contiguous-gene syndrome on an axis that is otherwise strictly inheritance mode. Left for its own stub.

So this PR retires three stubs and leaves two, and neither of the two is left by accident.

One thing the second opinion raises that I agree with and did not act on: the three `genetic:` entries (COL4A5, COL4A3, COL4A4) carry no `subtype:` key, so the subtype blocks are descriptive rather than load-bearing. `test_subtype_foreign_keys` only enforces the reverse direction so this is not a defect, but keying them would make the bindings do real work. That is a content change on top of a term-binding sweep, so I left it out; happy to do it here if a reviewer would rather see it in one PR.

PR #10795, opened 2026-09-03
curate(Kallmann_Syndrome): add SPRY4/HH17 subtype and bind the whole has_subtypes section (#10790)

### LINKED ISSUE #10790: Curate hypogonadotropic hypogonadism 17 with or without anosmia (MONDO:0014102)
Curate a dismech entry for **hypogonadotropic hypogonadism 17 with or without anosmia** ([MONDO:0014102](https://monarchinitiative.org/MONDO:0014102)) — *SPRY4* variants; Sprouty4 is an intracellular antagonist of FGF/RAS-ERK signalling, so this sits directly downstream of the FGFR1 pathway that already has an entry here.

`MONDO:0014102` is bound nowhere in `kb/`. SPRY4 appears only in `kb/disorders/FGFR1_Hypogonadotropic_Hypogonadism.yaml` — read that first, since Sprouty4 is a modulator of the very pathway that entry curates, and the two should agree about the signalling axis.

**The hard part is inheritance, not identity.** Congenital hypogonadotropic hypogonadism is substantially oligogenic, and SPRY4 variants are frequently reported alongside a second locus rather than as a self-sufficient cause. `CLAUDE.md` has a dedicated section on digenic/oligogenic inheritance — bind `inheritance_term` to `HP:0010983` (oligogenic) or `HP:0010984` (digenic) if the evidence supports it, with its own PMID and exact-quote snippet, rather than leaving the multi-locus claim as prose. If the monogenic sufficiency of SPRY4 is contested, curate it as contested.

Stub: `stubs/Hypogonadotropic_Hypogonadism_17_With_Or_Without_Anosmia.yaml`

kevinschaper:
Claim #10790 asked for hypogonadotropic hypogonadism 17 with or without anosmia (MONDO:0014102, *SPRY4*).

## The lump/split call: `entry_type: SUBTYPE`

I curated this as a `has_subtypes` entry on `kb/disorders/Kallmann_Syndrome.yaml` rather than a standalone disease. Three things decided it:

- **MONDO already places it there.** MONDO:0014102 is a direct child of MONDO:0018800 (Kallmann syndrome), which is this entry's `disease_term`. All eighteen numbered "HHn with or without anosmia" classes sit under it.
- **There is no distinct pathograph.** The existing node `Defective GnRH neuron and olfactory axon migration` already carries `GO:0008543` (fibroblast growth factor receptor signaling pathway) with `modifier: DECREASED` — which is exactly where the SPRY4 lesion lands. A standalone entry would restate the Kallmann chain for one non-distinctive molecular node.
- **The evidence does not support a disease-level identity.** There is no ClinGen gene-disease validity assertion for SPRY4; GenCC submitters classify it Limited to Moderate, and Genomics England PanelApp lists it amber on both hypogonadotropic hypogonadism panels — on the panel, not green-listed for routine diagnostic reporting.

Coverage is not lost by going this route: `has_subtypes[].subtype_term` is what `iter_covered_mondo_ids` reads, and `just tidy-stubs` now reports MONDO:0014102 as covered by this file.

## The inheritance call: `HP:0010983`, not `HP:0010984`

Oligogenic, not digenic. The published combinations are three-locus as well as two-locus (PROKR2/SPRY4/NSMF, PMID:35316923), and the founding cohort frames the FGF-pathway genes as contributors to an architecture rather than a defined two-locus requirement (PMID:23643382). The block sits on the subtype, with its own PMIDs and verified quotes, rather than as prose.

**Curated as contested, in both directions.** An independent Chinese cohort separates SPRY4 from DUSP6 on exactly this point — DUSP6 "alone seem sufficient", SPRY4 "in synergy with variants in other IHH-associated genes" (PMID:32389901) — and a 2025 family carried the SPRY4 variant in two unaffected relatives (PMID:40034250). Against that, PMID:31781046 reports a SPRY4 variant as the only finding across a 28-gene panel, and that counter-case is curated alongside so the oligogenic framing is not overstated.

## The mechanism, and the sign problem

The issue text says Sprouty4 is an FGF/RAS-ERK antagonist sitting downstream of FGFR1. True, but the naive reading points the wrong way: losing an antagonist should *raise* FGF signalling, while Kallmann syndrome is a decreased-FGFR1-signalling disease. Writing a "SPRY4 loss of function" node would have contradicted the entry it is meant to agree with.

The functional literature resolves it. Two Kallmann-reported substitutions — S241Y (PMID:33670044) and K177R (PMID:40806483) — are each characterised as *hyperactive* inhibitors, suppressing migration, proliferation and FGF- or serum-induced ERK activation where wild-type Sprouty4 does not, and S241Y does so selectively for FGF over EGF. Gain of inhibitory function in the antagonist converges on the same decreased-FGFR1 node as FGFR1 and FGF8 loss of function. That is recorded on the `genetic` entry with `relationship_type: SUSCEPTIBILITY`, both evidence items graded `IN_VITRO` with `directness: INDIRECT`.

A `HUMAN_MODEL_MISMATCH` discussion records what that evidence does not cover: the assays are in lung fibroblasts, osteosarcoma and lung-cancer lines, never in a GnRH neuron, and the same paper reports that in the lung-cancer lines both wild-type and variant Sprouty4 inhibited proliferation. That is a model-fidelity question, not an absence of evidence, which is why it is `HUMAN_MODEL_MISMATCH` and not `KNOWLEDGE_GAP`.

## Binding the rest of the section

The other five subtypes were also unbound, so rather than adding one bound subtype into an unbound section — which teaches the next reader the unbound half is deliberate — all were bound, following #9911, #10610 and #10780. Every CURIE was resolved through OLS4, the label copied from the response, and the OMIM cross-reference checked against the standard HH-series number.

| Subtype | MONDO | OMIM |
|---|---|---|
| ANOS1 | MONDO:0010635 | 308700 |
| FGFR1 | MONDO:0007844 | 147950 |
| CHD7 | MONDO:0012880 | 612370 |
| FGF8 | MONDO:0012988 | 612702 |
| SPRY4 (new) | MONDO:0014102 | 615266 |

**`PROKR2_PROK2` is the #10780 case.** One dismech subtype spans two MONDO leaves — HH3 (PROKR2) and HH4 (PROK2) — and `subtype_term` is single-valued, so binding either alone would silently drop the other. Both are declared as **disease-level** `skos:narrowMatch` mappings, since `iter_covered_mondo_ids` reads `mondo_mappings` only at the disease level, and the subtype carries a `review_notes` recording that the omission is deliberate.

MONDO:0007844 is also the `disease_term` of `FGFR1_Hypogonadotropic_Hypogonadism.yaml`. That is not a conflict: dismech curates HH2 both as its own disease and as a Kallmann subtype, and both statements are true. Coverage is unaffected either way.

I also keyed the five existing `genetic` entries to their subtypes with `subtype:`, which makes the subtype names load-bearing rather than descriptive — the improvement flagged on #10788.

## Stubs, research, validation

**Three stubs retired**: HH17 (the claim), plus the previously unclaimed HH3 (MONDO:0009482) and HH5 (MONDO:0012880), which the mapping and the CHD7 binding now cover. Confirmed with `just tidy-stubs`, which reports all three as `already_curated` by this file. Neither collateral stub had an open `claim` issue.

**Deep research** (`research/SPRY4_Hypogonadotropic_Hypogonadism-deep-research-falcon.md`, falcon) is included and passes `just preflight-dr … MONDO:0014102`: SPRY4 mentioned 83 times against 5 for the next gene, and the report's OMIM matches MONDO's. Its conclusions agree with what is curated here — gain of inhibitory function, oligogenic architecture, no SPRY4-specific epidemiology. None of its text was used as a snippet; every evidence item was taken from a cache-verified abstract.

**Validation**: `validate` (43/43 snippets verified), `validate-disorders`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history` — all exit 0.

No new pathophysiology node. The SPRY4 lesion lands on an existing node that already has the right term and the right modifier, and adding a parallel node would double-model it. No prevalence record: the DR run and the cohort papers agree that reported detection rates in referral cohorts are not population prevalence, so there is nothing honest to put there. No `Digenic` value on the inheritance term, for the reason given above.

Closes #10790

### DIFF
--- a/kb/disorders/Kallmann_Syndrome.yaml
+++ b/kb/disorders/Kallmann_Syndrome.yaml
@@ -197,6 +197,17 @@ has_subtypes:
       evidence_source: HUMAN_CLINICAL
       snippet: "Here we report the identification of new oligogenic variants in SPRY4/SEMA3A, SRA1/SEMA7A, CHD7/SEMA7A, CCDC141/POLR3B/POLR3B, and PROKR2/SPRY4/NSMF."
       explanation: Names the specific two-locus (SPRY4/SEMA3A) and three-locus (PROKR2/SPRY4/NSMF) combinations that make oligogenic, not digenic, the correct term here.
+    - reference: PMID:31781046
+      reference_title: A Rare SPRY4 Gene Mutation Is Associated With Anosmia and Adult-Onset Isolated Hypogonadotropic Hypogonadism.
+      supports: REFUTE
+      evidence_source: HUMAN_CLINICAL
+      snippet: "No other variants in the 27 remaining loci were identified by NGS."
+      explanation: >-
+        Counter-evidence to obligate oligogenicity, kept on the block it argues against
+        rather than only in prose. In this case the SPRY4 variant was the sole finding, so
+        a second locus is not required for the phenotype to appear. It weakens rather than
+        overturns the oligogenic reading, because the panel covered 28 loci and a second
+        hit outside them would not have been seen.
   evidence:
   - reference: PMID:23643382
     reference_title: "Mutations in FGF17, IL17RD, DUSP6, SPRY4, and FLRT3 are identified in individuals with congenital hypogonadotropic hypogonadism."
@@ -628,7 +639,7 @@ genetic:
     snippet: "Seven heterozygous mutations, two splice and \nfive missense, which were absent in > or = 180 controls, were identified in \nthree sporadic KS and four sporadic normosmic IHH patients."
     explanation: CHD7 mutations were identified in sporadic Kallmann syndrome patients.
 - name: SPRY4 variants
-  association: Contributory / oligogenic
+  association: Susceptibility
   relationship_type: SUSCEPTIBILITY
   subtype: SPRY4
   gene_term:
@@ -684,6 +695,17 @@ genetic:
     evidence_source: IN_VITRO
     snippet: "This mutation was found in two independent studies, and in both cases associated with the Kallmann syndrome"
     explanation: Establishes that the variant tested in that functional work is one reported in Kallmann syndrome patients, not an arbitrary substitution.
+  - reference: PMID:40806483
+    reference_title: The Lysine at Position 177 Is Essential to Limit the Inhibitory Capacities of Sprouty4 Protein in Normal and Cancer-Derived Cells.
+    supports: SUPPORT
+    directness: INDIRECT
+    evidence_source: MODEL_ORGANISM
+    snippet: "Knockout mice of Spry1, Spry2 and Spry4 show characteristic phenotypes often similar to the one observed with FGF overdose"
+    explanation: >-
+      The independent argument that simple SPRY4 loss of function is the wrong reading for
+      this disease. Deleting the antagonist in mouse produces an FGF-EXCESS phenotype, the
+      opposite direction from the decreased FGFR1 signalling Kallmann syndrome requires, so
+      the human variants have to be doing something other than abolishing Sprouty4.
   - reference: PMID:32389901
     reference_title: "Prevalence and associated phenotypes of DUSP6, IL17RD and SPRY4 variants in a large Chinese cohort with isolated hypogonadotropic hypogonadism."
     supports: SUPPORT


kevinschaper:
Thanks — I took three of the four suggestions and left one. Pushing dismisses the approval, so I batched them into a single commit rather than trickling.

**Suggestion 1 (Spry4-null mouse) — taken, but as an evidence item rather than an `animal_models` section.** You are right that this is the cleanest single piece of evidence that the loss-of-function reading is wrong, and right that it was missing. I disagree about the shape: a `FAILS_TO_RECAPITULATE` link is a substantive negative claim about a model's fidelity to a human mechanism, and the quotable sentence does not make that claim — it says the knockouts phenocopy FGF *overdose*, which is a statement about the direction of the pathway effect, not about whether a mouse reproduces HH17. Curating it as `FAILS_TO_RECAPITULATE` would also need `limitations` plus its own substantiating evidence, neither of which that sentence supplies.

So it now sits on the `genetic` entry, `MODEL_ORGANISM` + `directness: INDIRECT`, where it does the work it can actually do: deleting the antagonist in mouse produces an FGF-excess phenotype, the opposite direction from the decreased FGFR1 signalling this disease requires, so the human variants must be doing something other than abolishing Sprouty4. That is the same argument the in-vitro hypermorph data makes, now with an independent second line. No `animal_models:` section created, so no scope expansion.

**Suggestion 2 (PMID:31781046) — taken, as an addition rather than a move.** It is now also on the `Oligogenic inheritance (SPRY4)` block as `supports: REFUTE`, which is where it argues. I kept the copy on the subtype as `SUPPORT`, because there it supports the description's caveat sentence rather than contradicting anything — the same sentence reading `SUPPORT` for one claim and `REFUTE` for another is exactly the claim-relative pattern the evidence rules describe, and `check-snippet-grading` is unaffected since `evidence_source` is `HUMAN_CLINICAL` on both.

The `explanation` on the REFUTE item states the limit rather than letting the grade overstate: the panel covered 28 loci, so a second hit outside them would not have been seen. It weakens obligate oligogenicity; it does not overturn it.

**Suggestion 3 (`association`) — taken.** Now `Susceptibility`, matching the 49 existing uses and the enum-backed `relationship_type` on the same entry.

**Suggestion 4 (duplicate case-variant DOI cache) — declined.** Both files were written by the deep-research client's own reference resolution, three such collisions already exist on `main`, and neither is referenced from `kb/`. Deleting one would make this PR's committed artifacts differ from what the tool produces on a rerun, for no gain. If the case-collision is worth fixing it belongs in the fetcher, not in a curation PR — happy to open an issue for that if you think it is worth one.

Re-validated after the changes: `just validate` 45/45 snippets verified, and `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens`, `check-causal-targets` all exit 0.

PR #10797, opened 2026-09-03
curate: Add Platelet-type Bleeding Disorder 19 (MONDO:0014518, PRKACG) (#10791)

### LINKED ISSUE #10791: Curate platelet-type bleeding disorder 19 (MONDO:0014518)
Curate a dismech entry for **platelet-type bleeding disorder 19** ([MONDO:0014518](https://monarchinitiative.org/MONDO:0014518)) — *PRKACG* variants; the gamma catalytic subunit of protein kinase A, whose loss impairs filamin A phosphorylation and platelet cytoskeletal assembly, giving macrothrombocytopenia with impaired aggregation.

`MONDO:0014518` is bound nowhere. PRKACG appears in `kb/disorders/Dilated_Cardiomyopathy_1B.yaml` (a different disease) and in `kb/groupings/Inherited_Platelet_Function_Disorders.yaml`. **Read the grouping first** and check whether this disease should be a member — the `curate-grouping` skill covers membership logic. A grouping-membership addition edits a shared file, so if you conclude it belongs, say so in the PR body and consider whether it belongs in this PR or its own.

The evidence base here is small — this is essentially a single-family disease — so be careful with frequency bands and prevalence, and record what is unknown rather than inferring it.

Stub: `stubs/Platelet-type_Bleeding_Disorder_19.yaml`

kevinschaper:
Claim #10791 asked for platelet-type bleeding disorder 19 (MONDO:0014518, *PRKACG*). New entry: `kb/disorders/Platelet-type_Bleeding_Disorder_19.yaml`.

## The thing that shaped the entry, which was not visible from the stub

**ClinGen classifies the PRKACG–BDPLT19 relationship as DISPUTED** (Hemostasis/Thrombosis Gene Curation Expert Panel, SOP10, 2024-05-03). That record is not in `references_cache/` on main; I built it with `just clingen-rebuild --id CGGV:assertion_5d74cf9f-…`, and it is included here.

Reading ClinGen's narrative, the dispute is not about the experimental work. It says the functional evidence is real — reduced filamin A in patient megakaryocytes and platelets, platelet cAMP three to five times control, and a rescue in which wild-type PRKACG restored proplatelet formation in patient cells while the mutant construct did nothing. What it disputes is the genetic evidence, on two counts: the case-level evidence is two siblings in one consanguineous family, and the proband also carried a missense variant in *GNE* whose contribution was never evaluated, because GNE was not a thrombocytopenia gene in 2014.

It is one now. GNE-related severe congenital thrombocytopenia was established in 2018 (PMID:29941673) and has a substantial 2024–2026 literature. So the alternative ClinGen declined to score is harder to dismiss than it was when the family was reported — though nobody has gone back to the family to settle it, which is the actual state of the question and what the entry says.

That fact is curated in front rather than in a footnote: it is in the `description`, in an `external_assertions` record quoting the classification row and ClinGen's summary sentence, in the gene record as `relationship_type: DISPUTED`, in a `differential_diagnoses` entry for the GNE alternative, and in a `KNOWLEDGE_GAP` discussion that says what would resolve it (a second unrelated family, or a re-analysis of the first) and notes that neither has appeared in a decade.

## `entry_type: DISEASE`

dismech already curates the neighbouring BDPLT series as standalone entries — BDPLT8, BDPLT18, BDPLT20, BDPLT21 — alongside `TUBB1-related_Macrothrombocytopenia` and `CYCS-Related_Thrombocytopenia`. BDPLT19 has the same shape: its own OMIM/MONDO identity and a mechanism chain that is not a specialisation of anything curated. There is no candidate parent disease to hang it on as a subtype, and a `GROUPING` makes no sense for a single-gene entity. A `DISPUTED` gene-disease relationship is a reason to curate the dispute, not a reason to leave the concept uncurated — the stub queue has no state for "real concept, contested gene".

## Should it join the Inherited Platelet Function Disorders grouping? Not yet, and no `conforms_to`

The issue asked. The answer is no in this PR, and the reason is evidence, not caution.

That grouping's defining criterion is conformance to `primary_hemostatic_plug_failure`, and its scope note admits an inherited thrombocytopenia only when a qualitative functional defect is also curated. A functional defect *is* reported here — "Platelet activation and cytoskeleton reorganization were impaired" — but the primary report reaches `references_cache/` as an abstract only, and the abstract states it at exactly that level. The module's arm node is built from granule-secretion, integrin inside-out and agonist-specific aggregation claims, and asserting conformance would mean asserting those from text I cannot quote. The primary lesion is also platelet biogenesis, which is not one of the module's four arms.

This is the same treatment the already-curated `Platelet-type_Bleeding_Disorder_20` (SLFN14) gets, and the grouping's own notes list both SLFN14 and PRKACG as known gaps rather than members. `just check-groupings` on the grouping exits 0 and does not surface this entry as a candidate, so nothing is left in an inconsistent state.

**It is worth revisiting rather than filing away**, and `notes` says so: the deep-research report included here reports from the primary full text that patient platelets showed absent P-selectin externalization, no rise in surface αIIbβ3 on stimulation, and reduced GPIb internalization. That is arm-level data of exactly the kind the module node wants. None of it is quotable from the cache, so none of it is curated. If the full text becomes retrievable, the conformance and the membership should be added together — and the DISPUTED classification stated at the point of membership, so the grouping does not silently absorb a contested entity.

Membership would edit a shared file, so it belongs in its own PR either way.

## The pathograph

Five mechanism nodes, fully connected, ending on all four phenotypes:

```
Homozygous PRKACG p.Ile74Met Variant
  → Impaired PKA Catalytic Activity
    → Reduced Filamin A in Megakaryocytes and Platelets
      → Defective Proplatelet Formation → Macrothrombocytopenia, Giant platelets
      → Impaired Platelet Activation and Cytoskeletal Reorganization
            → Abnormal platelet function, Abnormal bleeding
Macrothrombocytopenia → Abnormal bleeding
```

The rescue experiment is curated as an `experimental_models` entry (`PRIMARY_CELL_CULTURE`, patient-derived megakaryocytes) with `modeled_mechanisms` `RESCUES` against the proplatelet node and a `RESTORED` readout, so it is a node in the graph rather than prose. Its `limitations` field says what it does and does not buy: a within-patient-cell rescue controls for genetic background in a way the pedigree cannot, but it does not supply the independent families the dispute turns on.

Two edges are deliberately `INDIRECT_KNOWN_INTERMEDIATES` rather than `DIRECT`: the PKA-to-filamin-A step goes through a phosphorylation event that was inferred rather than measured in patients, and the entry says so at the edge.

## What is deliberately absent, and why

- **No phenotype frequencies, no penetrance, no prevalence number.** The denominator is two siblings. `prevalence` carries `CASES_IN_LITERATURE` / `NOT_YET_DOCUMENTED` with the ClinGen sentence that bounds the evidence base, and nothing else.
- **No zygosity or identity for the GNE variant.** ClinGen says only "a missense variant in the GNE gene"; the Manchev abstract does not mention GNE at all. The full text would settle it and is not retrievable, so the entry states the uncertainty rather than filling it in.
- **No treatments.** Neither cached source says anything about management, and transfusion or antifibrinolytic guidance copied from the inherited-platelet-disorder literature generally would not be about this disease.
- **No animal model, no per-symptom bleeding phenotypes.** None reportable from a quotable source.

## Deep research

`research/Platelet-type_Bleeding_Disorder_19-deep-research-falcon.md` (falcon). `just preflight-dr … MONDO:0014518` returns **WARN, not FAIL**: PRKACG 50 mentions, and the rival-gene warning fires on FLNA at 17. FLNA is the PKA substrate at the centre of the mechanism, not a second disease entity, and the report's OMIM (616176) matches MONDO's. The other warning-adjacent name, GNE at 4, is the confounder discussed above. No DR text was used as a snippet; every evidence item quotes a cached abstract or the ClinGen record.

## Tooling defect found

`just refresh-orphadata` downloads `en_product1.xml` successfully and then aborts on a drifted checksum pin — upstream now serves sha256 `df8d562…` where the manifest pins `fb2fbe8…`. The file is left in place and usable, which is the documented behaviour. I used it to build ORPHA:438207 ("Severe autosomal recessive macrothrombocytopenia", exact match to MONDO:0014518, with a definition that enumerates the bleeding pattern), then **discarded it rather than committing it**: the serializer takes `snapshot_date` from the manifest, so the cache file's `## Source` line would have named the 2025-12-09 release it was not built from. Repinning is a separate change and should come with a full rebuild and a reviewed diff. The XML was deleted rather than left on disk.

## Validation

`validate` (25/25 snippets verified), `validate-disorders`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history`, `check-groupings` — all exit 0. Cache rows came from `just validate-terms`, not from hand.

Closes #10791

### DIFF
--- a/kb/disorders/Platelet-type_Bleeding_Disorder_19.yaml
+++ b/kb/disorders/Platelet-type_Bleeding_Disorder_19.yaml
@@ -409,6 +409,14 @@ genetic:
     Manchev abstract does not mention GNE at all, so the allele's zygosity is deliberately
     not asserted in this entry. Establishing it needs the primary full text, which is not
     retrievable into the reference cache.
+
+    As an unverified lead for whoever revisits this: the deep-research report in this
+    entry's PR reads the primary full text as reporting the allele as GNE c.1675G>A,
+    p.Gly559Arg, homozygous in both affected siblings, deprioritised at the time because
+    neither patient had myopathy or sialuria. Recorded here as a pointer to check, not as
+    a curated fact - if it is right, the alternative explanation is a good deal stronger
+    than a single heterozygous variant would make it, since GNE-related thrombocytopenia
+    is recessive.
   evidence:
   - reference: CGGV:assertion_5d74cf9f-a082-4cca-9a70-d634c3d0078b-2024-05-03T170000.000Z
     reference_title: PRKACG / platelet-type bleeding disorder 19 (Disputed)


kevinschaper:
Thanks — took suggestion 3, opened an issue for suggestion 1, and left the other three with reasons.

**Suggestion 3 (GNE variant identity) — taken.** The `notes` on the PRKACG genetic record now carries `GNE c.1675G>A, p.Gly559Arg`, homozygous in both siblings, explicitly labelled as a deep-research-derived lead to check rather than a curated fact. It earns its place because of the zygosity: if the allele really is homozygous, the alternative explanation is considerably stronger than a single heterozygous variant would make it, since GNE-related thrombocytopenia is recessive. That is exactly the thing someone revisiting the dispute needs to know, and there is no way to carry it without a quotable source other than labelling it as unverified.

**Suggestion 1 (bleeding phenotypes behind the Orphadata repin) — issue opened: #10798.** I filed it against the underlying tooling problem rather than just the missing phenotypes, because the failure mode is worse than "cannot refresh". The rebuilder takes `snapshot_date` from the manifest rather than from the file it parsed, so a cache file built from the drifted download would have carried `Orphadata snapshot 2025-12-09` — a release it was not built from — into a corpus where every other `ORPHA_*.md` genuinely is from that release. The issue asks for provenance to follow the parsed file, and notes the BDPLT19 phenotypes (HP:0000421, HP:0000132, a hematoma term) as the concrete thing waiting on the repin.

**Suggestion 2 (`evidence_source` on cultured-MK findings) — left, and not only because convention is mixed.** The sentence in question reports filamin A reduced "in mature megakaryocytes (MKs) **and platelets** of patients". The megakaryocytes are cultured; the platelets are direct patient samples. It is one sentence spanning both, so neither grade is fully right and `HUMAN_CLINICAL` is the more defensible half. The measurement that is unambiguously a culture experiment — the lentiviral rescue readout — is already `IN_VITRO`, so the distinction that matters is drawn where the evidence actually supports it.

**Suggestion 4 (node bundling) — left, for the reason you give.** "Impaired Platelet Activation and Cytoskeletal Reorganization" bundles two processes because the only quotable sentence bundles them. Splitting would mean asserting the two independently from a source that does not, which is the failure the bundling rule exists to prevent, pointed the other way. The node description says this and says what would let it be split.

**Suggestion 5 (`Macrothrombocytopenia` / `Giant platelets` overlap) — left deliberately.** They are separate because the proplatelet node reaches each by a different argument: count falls because fragmentation fails, size rises because platelet size is set at release. Collapsing them would lose the second edge. There was a real defect adjacent to this, though — an earlier draft had a *pathophysiology* node also called `Macrothrombocytopenia`, which the flat pathograph namespace would have collapsed into the phenotype and turned a legitimate mechanism→phenotype edge into a self-loop, the exact case `check_causal_targets.py` documents as issue #9896. That node was removed before this PR was opened and the chain rewired onto the phenotypes directly.

kevinschaper:
Thanks for the re-review. The two carried-over suggestions I answered in my previous comment. On the new one:

**`biological_scale` on the filamin A node — leaving it `CELLULAR`, and the reason is the same one that stopped me splitting the node.** You are right that a reduced protein level reads as `MOLECULAR` on its own; CLAUDE.md puts "a fusion protein's existence" at that scale. But the node is not only the protein level — it carries `GO:0030036` actin cytoskeleton organization `DECREASED` and both cell types, because the claim that does the causal work downstream is that the megakaryocyte's actin crosslinking is compromised, not that a band is fainter on a blot.

That the node would plausibly take two values is exactly the bundling signal CLAUDE.md describes, and normally the answer is to split. I did not, because the split would produce a "reduced filamin A protein" node with no independent evidence: the only quotable sentence bundles the filamin A deficit and the proplatelet defect in one clause, and the ClinGen restatement bundles the megakaryocyte and platelet compartments in another. A node whose sole evidence is a sentence that also evidences its successor is not an atomic claim, it is a restatement.

So the single value follows the load-bearing half. If the primary full text becomes retrievable — the same constraint behind the missing bleeding phenotypes and the arm-level platelet data, tracked in #10798 — the protein-level measurement would have its own evidence and the split would become the right call.

PR #10813, opened 2026-09-03
Add Immunodeficiency 60 (BACH2 haploinsufficiency) disorder entry

### LINKED ISSUE #10765: Curate immunodeficiency 60 (MONDO:0032723)
Curate a dismech entry for **immunodeficiency 60** ([MONDO:0032723](https://monarchinitiative.org/MONDO:0032723)) — `BACH2`-related immunodeficiency and autoimmunity, where haploinsufficiency for a transcriptional repressor that governs B-cell class switching and regulatory T-cell differentiation produces the unusual combination of humoral immunodeficiency and lymphocytic organ infiltration.

The stub is a MONDO leaf with one causal gene (`hgnc:14078` BACH2) under the broad `immunodeficiency disease` parent, so `DISEASE` is the expected outcome. `BACH2` appears in `kb/` only as a susceptibility locus in polygenic autoimmune entries (Type I Diabetes, Crohn, celiac, vitiligo and similar) — a different claim from this Mendelian entity, and worth keeping distinct. Membership in `Inborn_Errors_of_Immunity` / the relevant IEI grouping should be checked while curating.

Stub: `stubs/Immunodeficiency_60.yaml`

gingin77:
## Summary

Curates a new disorder entry for immunodeficiency 60 (MONDO:0032723), an autosomal dominant inborn error of immunity caused by BACH2 haploinsufficiency — the syndrome named BRIDA. Created de novo from the stub, which this PR deletes. `entry_type` decided `DISEASE`.

The blocking review item is resolved: the entry now rests on `PMID:28530713` (Afzali et al., *Nat Immunol* 2017) and the follow-up literature.

## What the entry says

**Mechanism.** One dosage lesion in a transcription factor that both lymphoid lineages need, producing a T-cell and a B-cell defect at once:

```
BACH2 Protein Destabilization  →  BACH2 Haploinsufficiency
   (homodimerization blocked,          │
    or aggregation)                    ├─→ Impaired Treg Differentiation → Lymphocytic colitis
                                       ├─→ Impaired Memory B Cell Dev.   → Immunoglobulin deficiency
                                       │                                    → Recurrent sinopulmonary infections
                                       └─→ Progressive T-cell lymphopenia
```

Two things here are easy to get backwards, and the entry is built to state them explicitly:

- **The defect is a loss, not an interference.** The mutations destabilize BACH2 by blocking homodimerization or causing aggregation, so a carrier has half the protein rather than a poisoned pathway. That is why the entry records `LOSS_OF_FUNCTION` and not `DOMINANT_NEGATIVE`.
- **BACH2 represses immunoglobulin production, yet losing it reduces useful antibody.** The resolution is that plasmacytic differentiation does not require BACH2 — Bach2-null B cells still make IgM and express Blimp-1 and XBP-1 *abundantly*, which is de-repression rather than mere preservation — while class switch recombination does require it. So the null cell terminally differentiates without ever switching isotype. Carried as an open question in earlier commits, now `RESOLVED` with a `resolution_note`.

**Classification.** IUIS Table 4 (immune dysregulation), not predominantly antibody deficiency, and added to `kb/groupings/Immune_Dysregulation_IEIs.yaml` where the `NECESSARY` criterion evaluates SATISFIED. The clinical picture is antibody-deficiency-shaped and ClinGen curated the gene through its Antibody Deficiencies panel, so that disagreement is recorded as an `INTERPRETATION` discussion rather than settled silently.

**Models.** Four entries, each one genotype and one publication, all with `modeled_mechanisms` and grounded readouts:

| Entry | Genotype | Target | Fidelity |
|---|---|---|---|
| BRIDA report | `Bach2+/-` | BACH2 Haploinsufficiency | MODERATE |
| Igarashi line, Treg | `Bach2+/-` | Impaired Treg Differentiation | **HIGH** |
| Igarashi line, Treg | null | Impaired Treg Differentiation | MODERATE |
| Igarashi line, class-switching | null | Impaired Memory B Cell Development | MODERATE |

The heterozygote Treg link is the best-evidenced one — Treg formation is *Bach2 gene-dose dependent*, matching both the dosage and the compartment of the human disease. The three Igarashi entries are **one mouse line characterized in two studies**, not independent derivations; each carries a `notes` line saying so, because a reader counting independent replication would otherwise over-count them. They are split only so each `publication:` matches the evidence on its own links.

**What is deliberately not asserted.** Only the abstract of the founding report is retrievable, so the immunoglobulin deficiency is asserted but not quantified. Single-case reports of SLE, juvenile dermatomyositis, and autoimmune enteropathy with T-LGL are kept out of the phenotype list and confined to a phenotype-expansion discussion — the T-LGL case carries an HLA-DQ8 allele and a concurrent lymphoproliferative disorder that is itself associated with autoimmunity. Immunoglobulin replacement is absent because no source consulted states it for BRIDA specifically, and the `treatments` note says so rather than leaving it to look like an oversight.

**Named Entity Confusion guard.** BACH2 is also a common-variant susceptibility locus in a dozen-plus dismech entries. That is a different claim from this Mendelian coding haploinsufficiency; the `notes` say so and name the entries. Whether the two act on one dosage axis is a `mechanistic_hypotheses` entry with `status: EMERGING`, carrying the founding report's own super-enhancer argument — not drawn as a causal edge.

**Schema gap surfaced.** `GeneDiseaseRelationshipEnum` has no tier for ClinGen *Moderate* or *Limited*. Recorded `CAUSATIVE` here with the Moderate rating carried in `external_assertions` and explained in `notes`. Filed separately as #10865, since it affects roughly a quarter of ClinGen's assertions rather than just this entry.

## Files

`kb/disorders/Immunodeficiency_60.yaml` (new), `kb/groupings/Immune_Dysregulation_IEIs.yaml` (member added), `stubs/Immunodeficiency_60.yaml` (deleted), six literature cache files plus the ClinGen record (all tool-generated, never hand-written), four term-cache rows for `GO:0002319` and `MONDO:0032723` (generated by `just validate-terms`), and two history records.

## Validation

`just validate-disorders` passes end to end. 50/50 evidence snippets verified; term validation clean; `check-entity-refs`, `check-causal-targets`, `check-enum-values`, `check-duplicate-keys`, `check-qualifier-terms`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `check-environmental-evidence`, `check-term-cache-integrity`, `check-cache-order` and `check-groupings` all pass; targeted test suites pass. `Closes #10765` is in the first commit message.

https://claude.ai/code/session_01S2SqJ5jZGm4Ppq4QnLviSk

### DIFF
--- a/kb/disorders/Immunodeficiency_60.yaml
+++ b/kb/disorders/Immunodeficiency_60.yaml
@@ -634,9 +634,11 @@ animal_models:
       defects" without resolving which compartment contributes what, so the model
       supports the dosage claim more strongly than any individual downstream node.
       Fidelity is recorded MODERATE rather than HIGH for that reason: the gene dosage
-      matches the human disease exactly, which is the strongest thing about this model,
-      but only the abstract of the founding report is available and it does not
-      characterise the mouse phenotype in enough detail to claim high fidelity.
+      matches the human disease exactly by construction, which is the strongest thing
+      about this model, but a matched dose is not a demonstrated dose dependence: only
+      the abstract of the founding report is available, it compares no genotypes, and
+      it does not characterise the mouse phenotype in enough detail to claim high
+      fidelity.
     readouts:
     - name: Lymphocyte maturation defects in heterozygous animals
       target: BACH2 Haploinsufficiency
@@ -681,8 +683,12 @@ animal_models:
       functional suppression assays that would show whether the remaining Tregs work.
       That caveat is about Treg function rather than Treg differentiation, so it does
       not weaken the model's fidelity for this node, which is a differentiation node -
-      hence HIGH here against MODERATE on the BRIDA heterozygote, whose supporting
-      sentence resolves neither dosage effect nor compartment. Same line as the null
+      hence HIGH here against MODERATE on the BRIDA heterozygote. The difference
+      between the two is dose dependence, not dose match: both are heterozygotes, so
+      both match the human gene dosage by construction, but only this one compares
+      genotypes and so demonstrates that Treg formation responds to dosage in a graded
+      way. The BRIDA heterozygote's supporting sentence demonstrates no such graded
+      response and names no compartment. Same line as the null
       entries below, so not an independent replication of them.
     readouts:
     - name: Regulatory T cell frequency in heterozygous animals


gingin77:
Pushed 829ee3be8, which fixes all five IMPORTANT items and two of the suggestions. The CRITICAL item is real but I cannot do it from this session, and the reason is not the one the review assumes.

## The network claim is environment-specific

The review says "Network access works in this environment" and treats the entry's stated constraint as no longer true. That inference is wrong for the session that wrote the entry. `claude-code-review` runs on a GitHub Actions runner with open egress; this curation session runs in a restricted container whose egress proxy denies PubMed. I re-ran the exact command after reading the review:

```
just fetch-reference PMID:28530713
  Fetching PMID:28530713...
  Failed to fetch PMID:28530713 from NCBI: <urlopen error Tunnel connection failed: 403 Forbidden>
```

Same for Europe PMC, Crossref, ClinicalTrials.gov, OLS and the OBO SQLite mirror. So both observations are correct — the paper is reachable from CI and not from here.

That does not weaken the finding underneath it, which I agree with: **this entry should rest on Afzali et al. 2017, and it does not.** The abstract quoted in the review supplies the immunoglobulin deficiency, the BRIDA synonym, the protein-destabilization step upstream of the haploinsufficiency node, and the `Bach2`-heterozygous mouse. I have not added any of it, because evidence snippets must validate against a file in `references_cache/`, that file can only be produced by `just fetch-reference`, and hand-writing one is prohibited. Quoting the abstract out of the review body would produce an evidence item that no validator can check.

**What is needed:** one curation pass from a session that can reach PubMed. That session should fetch PMID:28530713, add the immunoglobulin-deficiency node and phenotype, the protein-destabilization mechanism node, the mouse model with `modeled_mechanisms`, BRIDA in `synonyms`, and immunoglobulin replacement under `treatments`. The three knowledge-gap discussions in the entry name exactly what to fill.

## Fixed in this push

- **Item 1 is a genuine evidence defect and you are right about it.** The `PMID:22561518` snippet quoted a sentence about vitiligo loci in general while its `explanation` claimed the sentence named BACH2. It did not. Swapped for the sentence that does, with the association statistic.
- **Item 2**: `HP:0002205` → `HP:0005425 Recurrent sinopulmonary infections`.
- **Item 3**: dropped `GO:0045190 isotype switching / DECREASED`. Good catch — the entry asserted a decreased class switch in a binding while its own discussion said that account was unconfirmed. Now `GO:0002319 memory B cell differentiation`, which is what the IUIS row states.
- **Item 4**: took the direction the evidence supports. Kept `CL:0000787`, dropped "class-switched" from the `preferred_term`, and reworded the node description so it no longer claims a class-switched compartment. `CL:0000972` would have asserted the same specificity item 3 objects to.
- **Suggestions**: `PMID:29670635` regraded `IN_VITRO`; the susceptibility-entry list in `notes` is now marked a dated snapshot with an instruction to re-derive it.

Not done: **item 5, treatments.** Immunoglobulin replacement is certainly right, but nothing in the local reference cache states it for this disorder, and the cached BACH2 files are autoimmune-GWAS and lymphoma sources. Citing one for a treatment claim is the entity confusion the entry's own notes warn about. This needs the same networked pass as the CRITICAL item.

Also skipped `HP:0030388` for the same reason as items 3 and 4 — it asserts a class-switched defect that is not yet evidenced here. On the OMIM suggestion: agreed that `605394` in the IUIS row is the gene MIM, which is why the entry does not assert a disease MIM.

## One binding CI has to check

`GO:0002319` is not in `cache/go/terms.csv`, so I could not validate it offline. It is your network-verified suggestion and I took it; CI is the check. `MONDO:0032723` was in the same position last round and `test (3.13)` passed it.

---

gingin77:
Thanks for the retraction — and the deadlock is now confirmed from both ends, which is worth recording since it is a repo-level finding rather than anything about this entry.

You could fetch but not push (`github-actions[bot]` denied, 403). I can push but not fetch (egress proxy denies PubMed and OLS). Neither of us can complete a curation that needs both. I have assigned @cmungall, since you noted your token could not add an assignee.

CI is green on `f50ed414e` and every review item that does not need the network is done. What remains, all of it needing one networked pass:

- `just fetch-reference PMID:28530713`, then add the immunoglobulin-deficiency node between `Impaired Memory B Cell Development` and the sinopulmonary infection phenotype, the protein-destabilization node upstream of `BACH2 Haploinsufficiency`, the `Bach2`-heterozygous mouse under `animal_models` with `modeled_mechanisms`, and a `treatments` section.
- Narrow `discussions[bach2_imd60_humoral_phenotype_uncurated]` to the quantitative question rather than closing it.
- `just validate-terms kb/disorders/Immunodeficiency_60.yaml` and commit the four cache rows for `GO:0002319` and `MONDO:0032723`.

I am leaving the entry as it stands rather than quoting the abstract out of your review body. A snippet has to be an exact substring of a file in `references_cache/`, that file may only be produced by `just fetch-reference`, and hand-writing one is prohibited — so evidence transcribed from a review comment would be unverifiable by construction, which is the failure mode the evidence SOP exists to prevent. The entry names `PMID:28530713` in its `description`, `notes` and knowledge-gap discussion so the next curator does not have to rediscover any of this.

One thing worth separating out: the entry is thin, but it is not wrong. Everything in it is quoted from the IUIS 2022 Table 4 row or the committed ClinGen record, the pathograph is a connected component, and the gaps are recorded rather than papered over. Whether that clears the bar for merge with a follow-up issue, or should wait for the networked pass, is a call for @cmungall.

---

gingin77:
The deadlock is broken — egress was opened for this session, so I fetched the paper and rebuilt the entry on it. `4ab0c928` should close out the blocking item.

`just validate-disorders` now passes end to end for the first time, 48/48 snippets verified (up from 19), and the four term-cache rows are committed. `GO:0002319` resolves to `memory B cell differentiation` and `MONDO:0032723` to `immunodeficiency 60`, so both of the bindings that could only be taken on trust are now confirmed.

**Added from `PMID:28530713` and the follow-up literature:**

- **`BACH2 Protein Destabilization`** node upstream of the haploinsufficiency node — the mutations block homodimerization or cause aggregation. The `genetic_context` moved onto it, which is where the variant-level claim belongs.
- **`Immunoglobulin deficiency`** phenotype (`HP:0004313`) between the memory B-cell node and the sinopulmonary infection phenotype, replacing the single indirect edge.
- **Two animal models** with `modeled_mechanisms` and readouts: the dose-matched `Bach2`-heterozygous mouse (`RECAPITULATES`, `HIGH`) and the `Bach2` knockout (`PARTIALLY_RECAPITULATES`, `MODERATE`) with the homozygous-versus-heterozygous mismatch in `limitations`.
- **`treatments`**, **`prevalence`**, the reported R576L variant, and an incomplete-penetrance observation.

**The repressor paradox is answered, and it was worth chasing.** `PMID:15152264` shows Bach2-null B cells still make IgM and still induce Blimp-1 and XBP-1 — so plasmacytic differentiation does not require BACH2, and what fails is class switch recombination specifically. That is why losing a repressor of immunoglobulin production reduces *useful* antibody. The discussion is now `RESOLVED` with a `resolution_note` rather than deleted, since the reasoning is the interesting part.

**Two things I did not do, deliberately.**

Only the abstract of `PMID:28530713` is retrievable — Unpaywall and OpenAlex are still blocked here — so the immunoglobulin deficiency is asserted but not quantified, and the humoral discussion narrows from *absent* to *unquantified* rather than closing. Isotype-level values need the full text.

The single-case reports stay out of the phenotype list and sit in a phenotype-expansion discussion instead. The T-LGL case in particular carries an HLA-DQ8 allele and a concurrent lymphoproliferative disorder that is independently associated with autoimmunity, so attributing autoimmune enteropathy to BACH2 from it would be over-reading one patient. I also left immunoglobulin replacement out of `treatments` — clinically it is surely right, but no source consulted states it for BRIDA, and the section notes that absence rather than leaving it looking like an oversight.

On the MeSH leads you flagged (`Splenomegaly`, `Pancytopenia`, `Fever`, `Adrenal Cortex Hormones/therapeutic use`): those are indexing terms rather than quotable findings, and none appears in the retrievable abstract, so I did not curate phenotypes from them. The corticosteroid lead is covered from the second family's actual sentence instead.

Thanks for the earlier retraction, and for committing the fetched files locally even though the push failed — knowing the fetch worked from your side is what made it worth re-testing here.

---

gingin77:
Thanks for the approval. Pushed `67947c14` acting on all four actionable suggestions — this resets the approval, which seemed the right trade for three accuracy fixes and one structural one.

- **"Normally" understated the Blimp-1 result**, and you were right that it matters. The source says *abundantly*, and since BACH2 represses BLIMP1 that abundance is de-repression rather than preservation — the null cell proceeds to terminal differentiation without ever switching isotype. Both the description and the `resolution_note` now say that.
- **Split the knockout into two models.** `publication: PMID:23728300` with the B-cell link evidenced from `PMID:15152264` was a real inconsistency, not just tidiness. Each entry now names its own line and its own paper.
- **`fidelity` HIGH → MODERATE** on the heterozygous mouse. Taking your prompt to make the choice deliberate: MODERATE is right. The dose match is the strongest thing about the model, but claiming high fidelity from one abstract sentence overstates it, and HIGH sat awkwardly beside the link's own `limitations`.
- **`OTHER` → `COMPUTATIONAL`** for the super-enhancer enrichment sentence. It is a genome-wide analysis and the finer grade is more useful.

On suggestion 5, the `full_text_attempted: true` marker: I could not reproduce it. Neither `just validate` nor `just validate-kb-references` dirties `references_cache/` here, so there is nothing for me to pre-commit — it must be added by a code path that behaves differently on the runner. Flagging that rather than leaving it looking handled.

Two notes for whoever picks this up next. The remaining substantive gap is the **full text of `PMID:28530713`** — Unpaywall and OpenAlex are still blocked from this session, so isotype-level immunoglobulin values, patient class-switch assays and per-compartment counts stay in the humoral discussion rather than the entry. And the **ClinGen *Moderate* tier has no home in `GeneDiseaseRelationshipEnum`**, which maps Definitive/Strong to `CAUSATIVE` and Disputed/Refuted to `DISPUTED` but skips Moderate and Limited entirely. It is carried in `external_assertions` and explained in `notes` here, but a repo-wide fix would be a schema issue rather than a per-entry workaround. Happy to file it if that seems worth doing.

---

gingin77:
You're right, and I verified it against the cache rather than taking it on trust — `PMID_23728300.md` does say Roychoudhuri characterized KO mice "in which the Bach2 gene had been disrupted⁹", reference 9 being the same citation used two sentences earlier for the class-switching work, and Muto A and Igarashi K co-author both papers. One Igarashi line, two studies. Fixed in `b5b0ca67`.

I should have checked that before writing "independently derived" into the file. The premise came from the earlier review, but repeating a claim in a KB entry makes it mine, and this one would have caused a reader to over-count independent replication — which is exactly the kind of error the entry's other caveats exist to prevent.

**Kept the split, fixed the justification.** Each `publication:` genuinely should match the evidence on its own links, which is what the original suggestion was reaching for. Both entries are renamed to make the shared line visible (`Bach2-null mouse (Igarashi line, …)`) and both carry a `notes` line stating the shared provenance and warning against reading them as independent replication.

**Your other suggestion turned out to be the more valuable one.** The gene-dose sentence is now curated as its own model entry:

> "Treg cell formation was Bach2 gene-dose dependent since mice heterozygous for the KO allele had reduced frequencies of Treg cells"

That is better evidence for the regulatory T-cell arm than anything the entry had — it matches both the gene dosage *and* the compartment of the human disease, where the null link's `limitations` had to concede the dose mismatch and the BRIDA heterozygote only supports the haploinsufficiency node in general. It is now the entry's best-evidenced model link, at `fidelity: HIGH`, and it makes the Treg arm the dose-matched one rather than the inferred one.

Four `animal_models` entries now, each one genotype and one publication:

| Entry | Genotype | Target | Fidelity |
|---|---|---|---|
| BRIDA report | `Bach2+/-` | BACH2 Haploinsufficiency | MODERATE |
| Igarashi line, Treg | `Bach2+/-` | Impaired Treg Differentiation | **HIGH** |
| Igarashi line, Treg | null | Impaired Treg Differentiation | MODERATE |
| Igarashi line, class-switching | null | Impaired Memory B Cell Development | MODERATE |

50/50 snippets verified (up from 48), schema and term validation clean, all offline gates pass, 13,598 targeted tests pass.

On the `full_text_attempted` marker: still not reproducible here — neither `just validate` nor `just validate-kb-references` dirties `references_cache/` in this environment, so there is nothing for me to pre-commit.

---

gingin77:
Thanks. Pushed `1f7c2e7f` acting on all three actionable items, and filed both follow-ups.

**Your first suggestion was the same defect one entry over, and I had missed it.** The BRIDA heterozygote and the Igarashi heterozygote sit adjacent with the same `genotype:`, one declaring a shared line and the other silent — which reproduces the over-counting hazard the previous commit set out to close. The BRIDA entry now carries a `notes` line stating that its line provenance is not in the retrievable abstract and instructing readers not to count the two as independent unless full text establishes otherwise.

**On the fidelity asymmetry, I kept `HIGH` and made the reasoning explicit** in `limitations`. The missing functional suppression assays concern Treg *function*, whereas the target is a *differentiation* node and the measurement is Treg frequency at matched dosage — so that caveat does not bear on fidelity for this link. The BRIDA link stays `MODERATE` because its single abstract sentence resolves neither dosage effect nor compartment. Deliberate now rather than incidental.

**History record added** covering `67947c14` and `b5b0ca67`, including why the provenance claim was wrong and how it was verified.

Follow-ups, both filed:

- **#10865** — `GeneDiseaseRelationshipEnum` has no value for ClinGen *Moderate* or *Limited*. The numbers make a stronger case than I expected: those two tiers are **905 of the 3,414 real assertions** in the committed snapshot, about 26%. All three fallbacks lose information, and a query for `CAUSATIVE` today mixes Definitive- and Moderate-tier relationships across 1,491 entries. I laid out three options and leaned toward a separate `evidence_strength` slot, since ClinGen's tiers grade confidence in the assertion rather than the kind of relationship.
- **#10867** — mining the full text of `PMID:28530713`. Enumerates what the abstract withholds: isotype-level values, patient class-switch assays, per-compartment counts, the mouse line's provenance (which would resolve the note above), and whether the MeSH-indexed splenomegaly, pancytopenia, fever and lymphopenia belong as phenotypes.

Also removed the human assignee, since the deadlock that prompted it is resolved and it was the only thing holding this out of the auto-merge sweep.

---

gingin77:
Took the cosmetic one too — `421021ed`. It cost a re-review but no merge time, since the 3-day age gate dominates either way.

You identified the distinction exactly: **dose match is a property of the construct, dose dependence is a demonstrated graded response.** Both models are heterozygotes, so both match the human gene dosage by construction; only Roychoudhuri compares genotypes and so shows Treg formation responding to dosage in a graded way. Both `limitations` fields now name that difference, so the two entries visibly agree instead of requiring a reader to reconstruct why they aren't contradicting each other.

Prose only — no evidence item, binding or fidelity value changed. 50/50 snippets still verify, gates pass.

That closes everything raised across six rounds. The two follow-ups are filed (#10865, #10867), and cmungall is unassigned, so this is now eligible for the deterministic auto-merge sweep once it clears the age gate.

---

Issue #10764, opened 2026-09-03
Curate autosomal dominant aplasia and myelodysplasia (MONDO:0013851)


gingin77:
Curate a dismech entry for **autosomal dominant aplasia and myelodysplasia** ([MONDO:0013851](https://monarchinitiative.org/MONDO:0013851)) — an inherited bone marrow failure / MDS-predisposition syndrome caused by heterozygous `SRP72` variants, placing a signal recognition particle (co-translational protein targeting) defect upstream of marrow failure.

**First decide what this should be**: a `kb/disorders/` Disease, a `kb/groupings/` Grouping, a `has_subtypes` entry on an existing disease, or out of scope. See the Disease Groupings section of `CLAUDE.md` and #8727 for the rule — a disorder entry needs one reasonably conserved pathograph. Record the decision in the stub's `entry_type` before curating.

The stub is a MONDO leaf with one causal gene (`hgnc:11303` SRP72); its parents are the `bone marrow failure syndrome` / `inherited aplastic anemia` grouping terms, so `DISEASE` is the expected outcome. Worth checking during curation whether an existing inherited-bone-marrow-failure grouping should gain it as a member.

Stub: `stubs/Autosomal_Dominant_Aplasia_And_Myelodysplasia.yaml`
Nominated by: rare-disease-identification

gingin77:
Deferring this one, and recording why so the next person doesn't repeat the search.

The session that filed this claim had no outbound network access except GitHub — PubMed, Crossref, ClinicalTrials.gov and OLS all return 403 at the agent proxy — so `just fetch-reference` could not fetch anything and evidence had to come from what was already in `references_cache/`.

Searching that cache for SRP72 returns six files and **none of them supports the SRP72 to bone-marrow-failure claim**. Worth knowing what they actually are, because the list looks more promising than it is:

- PMID:28972538 is *SRP54*, not SRP72 — a different gene causing a different disorder (syndromic neutropenia with Shwachman-Diamond-like features). Easy to misread as on-target.
- PMID:37226705 is Shwachman-Diamond syndromes, again SBDS/SRP54 territory.
- The remaining hits are AML mutation panels, pure erythroid leukemia, the IUIS classification, and a monosomy-7 GeneReviews overview.

[...]

PR #10814, opened 2026-09-03
Curate Autosomal Dominant Aplasia and Myelodysplasia (BMFS1/SRP72)

gingin77:
## Summary

De novo curation of MONDO:0013851 (bone marrow failure syndrome 1, BMFS1), the autosomal dominant SRP72-deficiency marrow failure and myelodysplasia syndrome. Converts the stub to a full disorder entry.

*(Body current as of head `3f60b1e`, after six review rounds.)*

## Key Changes

- **New disorder entry**: `kb/disorders/Autosomal_Dominant_Aplasia_And_Myelodysplasia.yaml`
  - SRP72 (`hgnc:11303`) as the causative gene, autosomal dominant inheritance
  - A single connected chain terminating on the HPO-bound phenotype:
    `SRP72 Deficiency` → `Impaired SRP-Dependent Cotranslational Protein Targeting` → `Impaired Haematopoietic Output` → `Bone marrow hypocellularity` (HP:0005528) → `Pancytopenia` / `Myelodysplasia`
  - Four phenotypes, including congenital sensorineural hearing impairment (`HP:0008527`), left deliberately unattached to the causal chain because no mechanism links an SRP defect to cochlear pathology
  - `animal_models`: the *Srp72* heterozygous null mouse — `FAILS_TO_RECAPITULATE` against marrow hypocellularity, `PARTIALLY_RECAPITULATES` against haematopoietic output, with readouts
  - Two treatments: HSCT, and supportive transfusion / infection prophylaxis
  - A `prevalence` record: `CASES_IN_LITERATURE` / `ULTRA_RARE`, no rate
  - Five discussions: the molecular lesion, the mouse/human mismatch (`HUMAN_MODEL_MISMATCH`), tissue restriction of a housekeeping defect, the unexplained deafness, and penetrance
  - Tagged top-level `references:` block

- **Updated parent entry**: `kb/disorders/Inherited_Aplastic_Anemia.yaml` — BMFS1 added to `has_subtypes` with `subtype_term` MONDO:0013851, as a sixth mechanistic axis (co-translational protein targeting) alongside DNA repair, telomere, ribosome biogenesis, thrombopoietin signalling and inflammatory suppression

- **Term-cache rows** for the three new CURIEs, generated by `just validate-terms`

- **Removed stub**; five history records

## Implementation Details

**Lump/split decision**: DISEASE. The entry carries its own pathograph rather than being a `has_subtypes` row alone. The founding report makes the case itself, calling an SRP-component defect "a third pathway for developing these disorders alongside transcription factor and telomerase mutations."

**Evidence.** Curation began in a session whose egress policy blocked every literature host, so the first commits cited only what was already cached. Egress opened later in the same session and the primary literature was fetched and curated: `PMID:22541560` (Kirwan 2012, the gene-discovery report), `PMID:31254415` (the *Srp72* null mouse), `PMID:40922878` (incomplete penetrance, an unaffected carrier), and `PMID:41472573` (transplant outcomes, supportive-care thresholds, the case count, and a second truncating allele).

**Snippets: 33/33 verified on this entry, 72/72 across both changed files.**

**Two claims were weakened during review, and both weakenings were right.**

The first pathophysiology node was originally `SRP72 Haploinsufficiency`, inferred from the IUIS table's word "deficiency". That is a different claim, and the primary data argue against it — the mutant protein mislocalizes and one variant fails to associate with 7SL RNA, while a heterozygous null mouse gets no haematological disorder at all. The node is now `SRP72 Deficiency`. `srp72-molecular-lesion` frames the remaining question as whether the mechanism is *allele-dependent*: missense variants through a misassembling subunit, frameshift and splice variants through simple loss of function. `functional_impact_category` stays unset because the slot takes one value per variant context and this entry curates the gene, so any value would erase that split.

The second: an evidence item was graded `supports: REFUTE` on the reasoning that it refuted the phenotype's *constancy* rather than its association. `supports` is claim-relative in the schema, but it is **consumed** — `hpoa_export.py` maps REFUTE to the HPOA `NOT` qualifier, which asserts phenotype exclusion, and the export emitted a row denying that BMFS1 has congenital sensorineural hearing impairment. Now `SUPPORT`, with the reason recorded in the explanation so the next curator does not re-derive the same plausible mistake.

**Deliberate omissions**, each with its reason in `notes`: no `frequency` bands (few families, heterogeneous, deafness absent in some); no `rate_per_100000` on the prevalence record, because a case count is not a population rate; no module conformance — `stem_cell_exhaustion` scopes itself to age-associated decline and `myelosuppression` to treatment toxicity, so neither fits a germline targeting defect.

**Deliberately not cited**: `PMID:39501104` is cached and looks like a penetrance figure, but reports a pooled odds ratio across eight predisposition genes with no SRP72 breakout — citing it here would be a misattribution. Named in the `CURATION_TODO` alongside `PMID:27899666` (structural work, which would let the molecular node split into SRP68-binding and 7SL-binding sub-lesions) and the per-variant allele spectrum from `PMID:41472573`.

**No OMIM mapping** is possible: `DiseaseMappings` has only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings` and `ncit_mappings`.

## CI

`test (3.13)` is green on this head, including term validation, with zero cache churn. The HPOA exporter emits no NOT-qualified row.

`check-snippet-grading`, `check-title-snippets` and `check-snippet-length` exit non-zero repo-wide. Grepping each report for this PR's two files returns zero hits, and the flagged entries are ten files this branch does not touch. These gates passed earlier in this PR and `origin/main` was re-fetched before checking, so the drift arrived on `main` between Sep 4 and Sep 7 — not from this branch, and worth a separate issue.

Resolves issue #10764.

https://claude.ai/code/session_01Mi29QiWn65qhcUPM2JhjeP

### DIFF
--- a/history/disorders/Autosomal_Dominant_Aplasia_And_Myelodysplasia/2026-09-07T193630Z-claude-code-c53895.yaml
+++ b/history/disorders/Autosomal_Dominant_Aplasia_And_Myelodysplasia/2026-09-07T193630Z-claude-code-c53895.yaml
@@ -1,0 +1,80 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Dominant_Aplasia_And_Myelodysplasia
+  path: kb/disorders/Autosomal_Dominant_Aplasia_And_Myelodysplasia.yaml
+session:
+  id: 2026-09-07T193630Z-claude-code-c53895
+  timestamp: '2026-09-07T19:36:30Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10764
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10814
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - phenotypes
+  - discussions
+  summary: 'Round 5 review: fix the HPOA NOT row and the minigene evidence grade'
+  details: |
+    Addressed the round-5 review on PR #10814. Two findings, both correct, both
+    fixed in one push.
+
+    Finding 1 - `supports: REFUTE` on the hearing phenotype emitted a factually
+    wrong HPOA row. This is the more instructive of the two. The round-4 grade
+    was defended on the ground that `supports` is claim-relative, which CLAUDE.md
+    does say: the item refuted the constancy of the phenotype, not its
+    association with the disease. That reasoning is sound in the abstract and
+    still wrong here, because the value is consumed. `hpoa_export.py` maps REFUTE
+    to the HPOA `NOT` qualifier via SUPPORTS_TO_QUALIFIER, and its own comment
+    says REFUTE is deliberately not dropped. Reproduced before changing anything:
+
+        BEFORE  qualifier=''     HP:0008527  PMID:35748970  IEA
+                qualifier='NOT'  HP:0008527  PMID:40922878  PCS
+
+    The NOT row asserts BMFS1 does not have congenital sensorineural hearing
+    impairment - contradicting the positive row in the same file and the IUIS
+    row this entry cites for "Bone marrow failure and congenital nerve deafness".
+    HPOA reserves NOT for exclusion; incomplete penetrance belongs in the
+    frequency column, and genuine absence has its own path in the exporter
+    (`frequency: EXCLUDED`).
+
+    Changed to SUPPORT. The phenotype description already asserts non-constancy,
+    so the snippet supports that claim rather than refuting anything. The
+    explanation now records why the grade is SUPPORT and not REFUTE, so the next
+    curator does not re-derive the same plausible mistake. Verified after:
+
+        AFTER   qualifier=''  HP:0008527  PMID:35748970  IEA
+                qualifier=''  HP:0008527  PMID:40922878  PCS
+                NOT-qualified rows anywhere: none
+
+    The general lesson, worth carrying: a grade that is claim-relative in the
+    schema is not free-text when an exporter consumes it. Check what reads a
+    value before reasoning about what it means.
+
+    Finding 2 - a minigene splicing-assay readout was graded HUMAN_CLINICAL.
+    Confirmed against references_cache/PMID_41472573.md: the sentence sits inside
+    the minigene analysis paragraph, so IN_VITRO is right, and it now matches the
+    sibling PMID:40922878 in-vitro item in the same discussion rather than
+    disagreeing with it. Took both sub-points as well: the snippet was widened
+    from a mid-sentence fragment beginning after a close-paren to the full
+    self-contained sentence, and the explanation now carries the source's hedge
+    that the frameshift is *predicted* from the observed 2-base retention rather
+    than demonstrated at protein level.
+
+    Validation: `just validate-disorders` passes on both changed files including
+    term validation; snippets 33/33; entity refs, causal targets, duplicate keys,
+    enum values and qualifier terms all pass; `git status cache/` empty; the HPOA
+    exporter emits no NOT-qualified row for this entry. check-snippet-grading,
+    check-title-snippets and check-snippet-length still exit non-zero repo-wide
+    with zero hits in this entry - unchanged from the round-4 record, and still
+    drift that arrived on main between 2026-09-04 and 2026-09-07 rather than
+    anything this branch introduced.
--- a/kb/disorders/Autosomal_Dominant_Aplasia_And_Myelodysplasia.yaml
+++ b/kb/disorders/Autosomal_Dominant_Aplasia_And_Myelodysplasia.yaml
@@ -417,15 +417,19 @@ phenotypes:
       associated with SRP72 deficiency.
   - reference: PMID:40922878
     reference_title: "A 14-Year-Old Male Patient With Bone Marrow Failure Syndrome, Without Deafness, Caused by a Novel SRP72 Mutation Inherited From His Father: A Case Report."
-    supports: REFUTE
+    supports: SUPPORT
     directness: DIRECT
     evidence_source: HUMAN_CLINICAL
     snippet: "Some patients with SRP72 gene mutations have hearing loss, but our patient had normal results in the hearing test."
     explanation: >-
-      Graded REFUTE against the proposition that hearing impairment is a
-      constant feature of the syndrome: a genetically confirmed patient tested
-      with normal hearing. It does not refute the phenotype's association with
-      the disease, only its constancy.
+      Evidences the non-constancy asserted in this phenotype's description: a
+      genetically confirmed patient tested with normal hearing. Graded SUPPORT,
+      not REFUTE. `supports` is claim-relative in principle, but it is consumed
+      - `hpoa_export` maps REFUTE to the HPOA `NOT` qualifier, which asserts the
+      phenotype is excluded from the disease, and that is false here and
+      contradicts the positive row from the IUIS classification. Incomplete
+      penetrance belongs in the frequency column, and genuine absence would be
+      `frequency: EXCLUDED` instead.
   notes: >-
     The mechanism connecting an SRP targeting defect to cochlear pathology is
     unknown, so this phenotype is deliberately left unattached to the causal
@@ -699,12 +703,17 @@ discussions:
     reference_title: "A De Novo Splicing Mutation of SRP72 in Bone Marrow Failure Syndrome Type 1: Case Report and Review of the Literature."
     supports: SUPPORT
     directness: DIRECT
-    evidence_source: HUMAN_CLINICAL
-    snippet: "Leu502ValfsTer14), which alters the amino acids and derives a premature termination codon."
+    evidence_source: IN_VITRO
+    snippet: "This is predicted to result in a frameshift variant (p.Leu502ValfsTer14), which alters the amino acids and derives a premature termination codon."
     explanation: >-
       A second truncating allele, from a splice variant. A premature
       termination codon is the canonical loss-of-function lesion, reinforcing
       that the truncating and missense alleles may not share a mechanism.
+      Graded IN_VITRO because the sentence reports the readout of a minigene
+      splicing assay rather than a clinical observation, matching the sibling
+      in-vitro item above. Note the source's own hedge - the frameshift is
+      predicted from the observed 2-base retention, not directly demonstrated
+      at the protein level.
 
 - discussion_id: srp72-mouse-model-mismatch
   kind: HUMAN_MODEL_MISMATCH


gingin77:
Thanks — this caught two real defects I would not have found. Seven of the ten findings are fixed and pushed in `ab08201`. Four are not, and three of those are because the fix is either impossible or would break a different gate; the reasons are below so they can be argued with.

## Fixed

**The duplicated marrow node (finding 5) was the best catch here.** You were right, and I confirmed it by building the graph: `Bone Marrow Hypocellularity` (pathophysiology) and `Bone marrow hypocellularity` (phenotype, `HP:0005528`) were two separate nodes differing only in case, and the one carrying the HPO binding had no incoming edge. I took your first option — dropped the pathophysiology node, re-pointed the upstream edge at the phenotype's exact name, and moved its two downstream edges onto the phenotype as `sequelae`. The chain is now:

```
SRP72 Haploinsufficiency
  -> Impaired SRP-Dependent Cotranslational Protein Targeting
    -> Impaired Haematopoietic Output
      -> Bone marrow hypocellularity   (HP:0005528, one node, in the chain)
        -> Pancytopenia
        -> Myelodysplasia
```

Worth noting for anyone hitting this later: `check-causal-targets` stayed green the whole time, because nothing dangled. A duplicated node is invisible to it.

**GeneReviews was under-mined (finding 3), and my note about it was wrong.** The entry said "nothing in the cached references states it for this disease" about treatment. That was an overstatement — the same cached page carries the HSCT referral guidance you quoted. There is now an HSCT `treatments` entry citing it with `directness: INDIRECT`, since the guidance is stated for the monosomy 7 predisposition class whose Table 2 includes SRP72/BMFS1 rather than for BMFS1 specifically. The entry note is corrected and says it was corrected. A tagged top-level `references:` block is added too.

Also fixed: `HP:0008527` bound in place of the broader term (finding 4); the two GeneReviews clinical items regraded to `HUMAN_CLINICAL` (finding 6); `subtype_term` added to the has_subtypes row — the slot is `subtype_term`, not `disease_term` (finding 8); the second Harrison's justification strengthened (finding 10).

The `reference_title` mismatch (finding 1) was already fixed in `647ad36`, pushed before your review posted. Root cause worth recording: my pre-PR sweep ran `just check-reference-titles`, which compares by **similarity against a 0.85 threshold**, while `linkml-reference-validator` requires an **exact** match. The truncated title scored well above 0.85, so the local gate was green while CI was always going to fail. Those two checks disagree by design.

## Not fixed

**Primary SRP72 literature (finding 2).** I re-ran `just fetch-reference PMID:22541560` as you asked. It still fails:

```
Failed to fetch PMID:22541560 from NCBI: <urlopen error Tunnel connection failed: 403 Forbidden>
```

The inference that egress is open because term validation reached OLS does not hold — those are different machines. OLS is reachable from the GitHub runner; PubMed is not reachable from the curation session, where the organization's egress policy returns 403 for every literature host. So this genuinely needs a session with network, and I agree the mouse study is the highest-value thing missing.

**Cache rows (finding 7).** Same cause, plus a rule: `CLAUDE.md` says cache rows must come from `just validate-terms` and must never be hand-written. You supplied the row contents, which is useful confirmation that the three bindings are right, but transcribing them by hand is the thing the cache contract prohibits.

**OMIM mapping (finding 9) is not possible.** The `DiseaseMappings` class has exactly four slots — `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings`. There is nowhere to put 614675. It stays in the description prose.

**The IUIS row stays `OTHER` (rest of finding 6).** That single snippet backs six evidence items across `classifications`, `inheritance`, `genetic` and `phenotypes`. `check-snippet-grading` is keyed on the quoted sentence, so grading it `HUMAN_CLINICAL` in the phenotype items while leaving it `OTHER` elsewhere would create exactly the same-snippet divergence that gate exists to catch. Regrading all six would classify an expert classification table as human clinical evidence, which your review agrees it is not.

## Validation

Reference validation — the gate that failed — now reports "All validations passed". Schema passes on both files, snippet verification is 53/53, the eleven other offline gates pass, and the page renders. Term validation still cannot run here, but it passed in CI on the previous commit for both files, so the three new CURIEs resolve.

---

gingin77:
Two things: the re-review raced my push by 25 seconds, so most of what it lists as unaddressed was already fixed. But its one new finding is correct and I've acted on it in `0abfcce`.

## The node name was wrong, and that is a real finding

`pathophysiology[0]` was named **SRP72 Haploinsufficiency**. I inferred that from the IUIS table calling the entity "SRP72-deficiency", and those are not the same claim — a classification table saying a functional deficiency exists says nothing about whether the deficit is a halved dose of a normal subunit or a mutant subunit that misassembles. Reduced 7SL RNA association and mislocalized protein argue against simple dosage. Naming the node for a mechanism I could not cite is exactly the inference this entry is otherwise careful to avoid.

Fixed by renaming the node to **SRP72 Deficiency**, which is what the cited source supports, stating the unresolved choice in the node description, and adding a fourth discussion `srp72-molecular-lesion` (`KNOWLEDGE_GAP`) attached to the node and to `genetic#SRP72`. `functional_impact_category` is left unset deliberately — every value it could take would assert the distinction the entry cannot yet make.

**I did not use the Kirwan abstract you quoted as a snippet.** It reached this session through a review comment rather than through `references_cache/`, and quoting unverified relayed text as an exact source substring is the thing the evidence SOP exists to prevent. So the finding is acted on by weakening the claim to what is citable, not by citing text I cannot verify.

## On the timing

The re-review was submitted at 23:37:44 against `647ad36`. My commit `ab08201` landed at 23:38:09. So findings 2 through 7 were already fixed when the review said none had been addressed. On the current head:

| Re-review finding | State |
|---|---|
| 2 — GeneReviews untagged, Management unmined | Fixed in `ab08201` — `references:` block with `tags: [GeneReviews]`, plus an HSCT `treatments` entry quoting the Management section at `directness: INDIRECT` |
| 3 — `HP:0008527` unused | Fixed in `ab08201` |
| 4 — duplicated marrow node | Fixed in `ab08201` — node dropped, edge re-pointed, downstream moved to `sequelae` |
| 5 — blanket `OTHER` grading | Partly fixed in `ab08201`; the IUIS row stays `OTHER` for the reason below |
| 6 — cache CSV rows | Still open, and not fixable here |
| 7, 8, 9 — subtype term, OMIM mapping, Harrison's justification | `subtype_term` and the justification fixed in `ab08201`; the OMIM mapping is impossible — `DiseaseMappings` has only `icd10cm_mappings`, `icd11f_mappings`, `mondo_mappings`, `ncit_mappings` |

On the IUIS grading specifically: that one snippet backs six evidence items across four sections, and `check-snippet-grading` keys on the quoted sentence. Grading it `HUMAN_CLINICAL` in the phenotype items alone would create the same-snippet divergence that gate exists to catch, and grading all six that way would classify an expert classification table as human clinical evidence.

## The fetch

I have now attempted `just fetch-reference PMID:22541560` three times from this session. It fails every time:

```
Failed to fetch PMID:22541560 from NCBI: <urlopen error Tunnel connection failed: 403 Forbidden>
```

Your runner's fetch succeeded but committed nothing — there is no `references_cache/PMID_22541560.md` on this branch. The egress difference is between machines, not between this repo and CI, so "one command away" is true on the runner and not true here. That is why the primary literature remains uncited and why the `srp72-evidence-completeness` `CURATION_TODO` is still the right home for it. If a session with PubMed access picks this up, resolving `srp72-molecular-lesion` is the highest-value thing to do first.

---

gingin77:
CI is red on `0abfcce`, and it is worth being precise about why, because it turns the reviewer's cache-rows finding from a policy point into the cause of a broken build.

## What failed

`test (3.13)`, at the term-validation step:

```
🌐 Unable to validate at this time: ontology service unavailable.
   could not reach ontology service to resolve GO:0006614:
```

Everything else in that run passed — schema, reference validation, snippets 53/53, stub check.

This is a five-second read timeout against EBI, not a bad binding. The same three CURIEs resolved on this same branch 38 minutes earlier, in the run on `889ba56`:

```
Term validation (batch)...
✅ Autosomal_Dominant_Aplasia_And_Myelodysplasia.yaml
✅ Inherited_Aplastic_Anemia.yaml
✅ All 2 files passed validation
```

Nothing about the terms changed between those two runs. `GO:0006614`, `MONDO:0013851` and `hgnc:11303` are the same bindings, and the validator itself says the condition is not a data error.

## Why this happened, and the real fix

The durable fix is the cache rows — finding 6 in the last review. Term validation is cache-first: a CURIE already in `cache/` is never looked up over the network. Because these three are absent, **every CI run on this PR has to reach OLS**, so an EBI hiccup fails the build. That is the concrete cost of the missing rows, beyond determinism in the abstract.

I cannot generate them. Producing those rows requires resolving the CURIEs against OLS, which is exactly what this session cannot do, and `CLAUDE.md` prohibits hand-writing cache rows — they must come from `just validate-terms`. So the one change that would make this build robust is the one change blocked by the same egress restriction that caused the rest of this PR's limitations.

## What I did about it

I tried to re-run the failed job and do not have permission:

```
POST /repos/monarch-initiative/dismech/actions/runs/33818999041/rerun-failed-jobs
403 Resource not accessible by integration
```

I am not pushing an empty commit to kick CI, and there is no test to make more robust within this PR's scope — the fragile step is a network call to a third-party service, not a test. So this needs one of:

- a re-run by anyone who can trigger one, which I expect to pass; or
- a session with OLS access running `just validate-terms` on this branch and committing the five resulting cache files, which fixes it permanently.

I am keeping the PR watched and will re-check rather than leaving it red.

---

gingin77:
`claude-review` is failing on this PR for a reason that has nothing to do with the diff: **the reviewer bot has hit its API usage limit.** From the job log on `141c0aa`:

```
detail="You've hit your limit · resets Sep 5, 8pm (UTC)"
num_turns=1  cost=$0
```

`num_turns=1` and `cost=$0` — it stopped before doing any work. That is why it errored twice, "after 44s" on `4b730d1` and "after 0s" here. Re-running will not help; the limit holds until **Sep 5, 8pm UTC**.

This has a knock-on effect worth naming. `claude-review` is the approving reviewer here, and my last two pushes dismissed the earlier approval, as branch protection is meant to. So this PR cannot regain an approval until that limit resets, however sound the content is. Nothing to do but wait; I am not re-running it and not touching the review state.

## Meanwhile, the two blockers from earlier are resolved

Egress from my session opened about 14 hours after the entry was written, so the constraint behind both of them is gone.

**The OLS timeout that reddened `test (3.13)` is fixed at the root** (`4b730d1`). `just validate-terms` now runs and produced exactly the six cache rows predicted in review, labels matching what the entry asserts. Term validation is cache-first, so those rows remove the EBI lookup that timed out, rather than just getting one run lucky.

**The primary SRP72 literature is now curated** (`141c0aa`):

- Kirwan 2012 (`PMID:22541560`) supports `genetic#SRP72` and gives two pathophysiology nodes their first direct, disease-specific evidence. The transfection and coimmunoprecipitation results are graded `IN_VITRO`.
- The `Srp72` null mouse (`PMID:31254415`) is in a new `animal_models` block with `FAILS_TO_RECAPITULATE` against marrow hypocellularity and `PARTIALLY_RECAPITULATES` against haematopoietic output, plus a fifth discussion `srp72-mouse-model-mismatch` (`HUMAN_MODEL_MISMATCH`).
- Incomplete penetrance is cited (`PMID:40922878`, an unaffected carrier of the same variant) instead of asserted.

Snippets went from 14 to 26 on this entry; 53/53 across both changed files.

The `SRP72 Deficiency` rename turned out to be right for a second reason as well as the first. The in-vitro assembly defect was one argument against haploinsufficiency; the dosage-only mouse getting no haematological disorder is an independent one. `functional_impact_category` still stays unset, because neither shows the mutant subunit impairing wild-type-containing particles, which is what a dominant-negative claim would need.

One thing I deliberately did not do: `PMID:39501104` is cached and looks like a penetrance figure, but it reports a pooled odds ratio across eight predisposition genes with no SRP72 breakout. Citing it for this disease would be a misattribution, so it is named in the `CURATION_TODO` instead, alongside the structural paper.

`test (3.13)` is still running on `141c0aa` — the first CI run to validate the cache rows and the new evidence together. Locally that combination passes `just validate-disorders` end to end. I will report if it comes back red.

---

gingin77:
Both IMPORTANT items and all four suggestions are fixed in `4e4f9f7`, one push.

## The first finding is a fair hit, and I'd made it twice

`PMID:41472573` was fetched in the round-3 batch, committed, and then cited nowhere — while the entry went on asserting that no BMFS1-specific transplant outcome, no supportive-care statement and no prevalence figure was quotable. All three are in that file. That is prose describing repository state, written about a cache the same change set had already altered, which is exactly the defect this PR corrected once before. Once is a slip; twice is a pattern worth naming.

I curated the three claims rather than deleting the sentences:

- **`treatments[0]` gains two `DIRECT` items.** The prior transplant experience — two patients, one good outcome, one death from acute immune rejection — and the index case, normal parameters at day 132. The GeneReviews referral pathway stays `INDIRECT` alongside them.
- **A second treatment** for supportive transfusion and infection prophylaxis, carrying the source's own thresholds. Deliberately **no `target_mechanisms` link**: transfusion replaces the marrow's product rather than acting on any node in the chain, so a mechanism edge would overstate what it does.
- **A `prevalence` record**, `measure_type: CASES_IN_LITERATURE`, `prevalence_class: ULTRA_RARE`, and no `rate_per_100000`. Six families is a case count, not a population rate, and the source says in the same paragraph that the undiagnosed fraction cannot be determined.

One thing I added beyond the ask: the treatment description now names the transplant denominator and warns against reading day 132 as a success rate. Three transplanted patients exist in the whole literature and one of them died of rejection, so an entry that cited only the good outcome would mislead.

## The second finding is the better one

You're right that `srp72-molecular-lesion` presented one side while two sources on this branch report truncating alleles with an explicit loss-of-function mechanism. The gap now asks whether the mechanism is **allele-dependent** — the founding missense variants acting through a misassembling subunit, the frameshift and splice variants through simple LOF — and cites both directions: the mouse and the 7SL co-immunoprecipitation against dosage, `PMID:40922878`'s in-vitro LOF finding and `PMID:41472573`'s premature termination codon for it.

That also sharpens why `functional_impact_category` stays unset. The slot takes one value per variant context; this entry curates the gene rather than individual alleles, so any single value would erase the split the evidence suggests. The discussion now says that.

## Suggestions

All four taken. The deafness-absence claim is cited — graded `REFUTE` against the proposition that hearing impairment is *constant*, not against its association with the disease, and quoting the body sentence rather than the title so it doesn't trip `check-title-snippets`. `PMID:41472573` is named in the `CURATION_TODO` for the one thing still uncurated from it, the per-variant allele spectrum. The Kirwan conclusion snippet drops to `INDIRECT` — you're right that it asserts a role in AA/MDS pathophysiology, not measured targeting impairment. And the model-mismatch discussion now attaches by the mouse's unique `name` rather than by `species`.

## Validation

Snippets **33/33**, up from 26. `just validate-disorders` passes on both files including term validation, with `git status cache/` empty — the committed rows still cover this entry offline.

On the three gates that exit non-zero repo-wide: I checked rather than assuming your stale-baseline explanation. Grepping each report for this entry returns nothing, and the flagged files are `CRX_Related_Retinopathy`, `Ewing_Sarcoma`, `Glutaryl-CoA_Dehydrogenase_Deficiency`, `Tangier_Disease`, `IgG4-Related_Disease`, `Primary_Ciliary_Dyskinesia`, `AIPL1-Related_Retinopathy`, `Congenital_Glaucoma`, `EYS_Related_Retinitis_Pigmentosa` and `Tay-Sachs_Disease` — none touched by this branch, which changes exactly two KB files. Worth flagging separately though: these gates **passed** earlier in this PR, and I re-fetched `origin/main` before checking, so this is not a stale local baseline. The drift arrived on `main` between Sep 4 and Sep 7 and is somebody's to look at.

---

gingin77:
Both fixed in `3f60b1e`. The first finding is the best catch of this review cycle, because I'd defended the wrong answer with correct reasoning.

## The `REFUTE` grade

I graded that item `REFUTE` and argued it refuted the phenotype's *constancy* rather than its association — leaning on `supports` being claim-relative, which `CLAUDE.md` does say. That is true of the schema and still wrong here, because **the value is consumed**. `hpoa_export.py` maps it through `SUPPORTS_TO_QUALIFIER`, and the comment right above that mapping says REFUTE is deliberately *not* dropped.

I reproduced it before changing anything:

```
BEFORE  qualifier=''     HP:0008527  PMID:35748970  IEA
        qualifier='NOT'  HP:0008527  PMID:40922878  PCS
```

And after:

```
AFTER   qualifier=''  HP:0008527  PMID:35748970  IEA
        qualifier=''  HP:0008527  PMID:40922878  PCS
        NOT-qualified rows anywhere: none
```

Changed to `SUPPORT`. The phenotype description already asserts non-constancy, so the snippet supports that rather than refuting anything. I've put the reasoning *into the explanation* — that the grade is SUPPORT and not REFUTE specifically because the exporter reads REFUTE as exclusion — so the next curator doesn't re-derive the same plausible mistake from the same correct premise.

The transferable lesson, which I've also recorded in the history record: a grade that is claim-relative in the schema stops being free text the moment an exporter consumes it. Check what reads a value before reasoning about what it means.

## The minigene grade

Confirmed against the cache — the sentence sits inside the minigene analysis paragraph, so `IN_VITRO` is right, and it now agrees with the sibling `PMID:40922878` in-vitro item in the same discussion instead of contradicting it. Both sub-points taken as well: the snippet was a mid-sentence fragment starting after a close-paren and is now the full self-contained sentence, and the explanation carries the source's hedge that the frameshift is *predicted* from an observed 2-base retention rather than demonstrated at protein level. You were right that the entry's own standard elsewhere is to carry that hedge.

## Validation

Snippets 33/33. `just validate-disorders` green on both files including term validation, `git status cache/` empty. Entity refs, causal targets, duplicate keys, enum values and qualifier terms all pass. The HPOA exporter emits no NOT-qualified row for this entry.

The three repo-wide gates still exit non-zero with zero hits in this entry — unchanged from round 4, and still drift that arrived on `main` between Sep 4 and Sep 7 rather than anything this branch introduced. Worth someone opening an issue for separately.

---

PR #10817, opened 2026-09-03
curate(DeSanto-Shinawi_Syndrome): bind both MONDO subtypes (#10811)

### LINKED ISSUE #10811: Curate DeSanto-Shinawi syndrome due to WAC point mutation (MONDO:0014741)
Curate a dismech entry for **DeSanto-Shinawi syndrome due to WAC point mutation** ([MONDO:0014741](https://monarchinitiative.org/MONDO:0014741)) — the point-mutation form of WAC-related DeSanto-Shinawi syndrome, as distinct from the 10p12 deletion form.

**This is a lump/split question and it should be settled before any curation.** `kb/disorders/DeSanto-Shinawi_Syndrome.yaml` already exists, curated at `MONDO:0018760` — the parent concept. This stub's term is the narrower *point-mutation* child. So the real question is whether dismech needs both.

Check what the existing entry actually **binds** versus merely mentions, and whether it has a `has_subtypes` block distinguishing point mutations from the deletion. Three outcomes are all legitimate:

- **`SUBTYPE`** — bind `MONDO:0014741` as a `subtype_term` on the existing entry, adding the subtype if it is absent. This is the pattern applied six times this week (#9911, #10610, #10780, #10792, #10793, #10795); note the rule that binding half a section teaches the next reader the unbound half is deliberate.
- **Already covered** — if the existing entry's `disease_term` or a `skos:exactMatch`/`narrowMatch` mapping already accounts for it, record that and delete the stub.
- **`DISEASE`** — only if the point-mutation and deletion forms have genuinely divergent pathographs, which would need arguing.

Whichever you choose, argue it against the cblA (#10260, standalone) and VWM5 (#10424, subtype) precedents rather than by default, and do not modify the existing entry beyond what the decision requires.

Stub: `stubs/DeSanto-Shinawi_Syndrome_Due_To_WAC_Point_Mutation.yaml`

kevinschaper:
## What this settles

`MONDO:0014741` (DeSanto-Shinawi syndrome due to WAC point mutation) is curated as a **`SUBTYPE`**, not a new disease entry.

`kb/disorders/DeSanto-Shinawi_Syndrome.yaml` already existed. What it **bound** was only the parent concept: `disease_term` = `MONDO:0018760`, plus one `mappings.mondo_mappings` entry that is a `skos:exactMatch` to that same parent. There was **no `has_subtypes` block at all**. An `exactMatch` on the parent does not account for a child, so `MONDO:0014741` was genuinely uncovered.

What the entry **mentioned** but did not bind: its description and `genetic` section talk about "heterozygous de novo truncating, splice, or other loss-of-function variants" — i.e. the point-mutation form — without ever naming it as a distinct concept.

## What changed

`MONDO:0018760` has exactly two children. Both are now bound as `has_subtypes` entries:

| Subtype | MONDO |
|---|---|
| WAC point mutation | `MONDO:0014741` |
| 10p11.21p12.31 microdeletion | `MONDO:0017283` |

Binding both rather than only the claimed one follows the rule from #9911: binding half a section teaches the next reader that the unbound half is deliberate. There is no separate stub for the deletion form, so it would otherwise have stayed invisible.

## Why not a standalone disease entry

The existing pathograph — WAC haploinsufficiency → impaired ULK activation and autophagy → GABAergic-neuron perturbation → seizures and neurobehavioral abnormalities — comes entirely from intragenic-variant cohorts. A separate point-mutation entry would duplicate it wholesale rather than add a divergent mechanism.

The deletion form is the one with a plausible claim to a divergent pathograph, since it is a contiguous-gene deletion. DeSanto 2015 (PMID:26264232) argues the other way: "Our data highly suggest that WAC haploinsufficiency is responsible for most of the phenotypic features associated with deletions encompassing 10p11.23." That is the basis for keeping both under one disease. Note it says *most*, not all, and the subtype description says so.

The deletion subtype description records what does **not** carry over:

- Reported deletions run from about 1 Mb to over 10 Mb, so genes other than WAC are lost in most patients and the phenotype is not necessarily reducible to WAC dosage.
- Wentzel 2011 (PMID:21522184), the defining deletion series, reports cardiac malformation in six of seven patients. That is not an established frequency for the intragenic form, so it was **not** added to the entry's `phenotypes` section. If a denominator is not the whole disease, it is not the disease's frequency.

Nothing else in the entry was touched.

## Terms

Both MONDO CURIEs resolved through the configured adapter and their labels are copied verbatim, including the inconsistent capitalisation MONDO itself uses on `MONDO:0017283` ("DeSanto-Shinawi **S**yndrome due to…").

## Validation

All run against this branch; every one exit 0.

| Check | Exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just count-verified-snippets` | 0 (18/18 verified) |
| `just validate-disorders kb/disorders/DeSanto-Shinawi_Syndrome.yaml` | 0 |
| `check-duplicate-keys` / `check-entity-refs` | 0 / 0 |

No deep-research artifact: this PR adds no new `kb/disorders/` entry.

Closes #10811

### DIFF
--- a/kb/disorders/DeSanto-Shinawi_Syndrome.yaml
+++ b/kb/disorders/DeSanto-Shinawi_Syndrome.yaml
@@ -77,7 +77,8 @@ has_subtypes:
     the intragenic-variant cohorts recapitulate the core neurodevelopmental and
     craniofacial features. Cardiac malformation was reported in six of seven
     patients in the defining deletion series, a frequency not established for
-    the intragenic form and not carried over to this entry's phenotype section.
+    the intragenic form; it is recorded in this entry's phenotype section scoped
+    to this subtype only.
   subtype_term:
     preferred_term: DeSanto-Shinawi syndrome due to 10p11.21p12.31 microdeletion
     term:
@@ -357,7 +358,7 @@ phenotypes:
     abnormalities in five of six patients, but under different wording and
     ascertainment, so the two are not pooled here.
   phenotype_term:
-    preferred_term: Visual impairment
+    preferred_term: Various visual impairments
     term:
       id: HP:0000504
       label: Abnormality of vision


kevinschaper:
Done — both blocking and non-blocking points addressed.

**The subtype-scoped phenotypes.** You are right that I treated this as all-or-nothing. `Phenotype.subtype` scopes the claim to one `has_subtypes[]` entry, so a phenotype tagged `subtype: 10p11.21p12.31 microdeletion` says nothing about the intragenic form, and the reason I left it out does not apply to it. Two phenotypes added from the sentence you quoted:

- **Cardiac malformation (6/7)** → `HP:0001627` Abnormal heart morphology. I checked the term myself rather than taking the suggestion: the abstract names no specific lesion, so the general heart-morphology term is the accurate binding and anything narrower would be manufactured.
- **Abnormality of vision (7/7)** → `HP:0000504`. I did **not** use `HP:0000505` (Visual impairment), whose HPO definition is vision loss "that cannot be corrected by conventional means, such as refractive correction". Wentzel says "various visual impairments" without identifying a deficit, so `HP:0000505` would assert more than the source does. `HP:0000504` is its parent and is what the sentence supports.

Both carry the caveat in their own `description`, and the subtype `description` caveat is unchanged, as you asked.

I noted in the vision phenotype that DeSanto 2015 separately reports eye abnormalities in 5/6 intragenic patients. I did not pool the two — different wording, different ascertainment — so that stays out of the deletion-scoped record.

**Non-blocking:** `links.issues` and `links.prs` populated on the history record, and its `details` now records the review follow-up and the reasoning for both term choices.

**Re-validated after the change:** `just validate` 0, `just validate-terms` 0, `count-verified-snippets` 20/20, `validate-history` 0, and `pytest -k subtype_foreign_keys` passes (2544 passed). Branch rebased on current `main`. Incidental `references_cache` refetch churn from the validator was reverted so the diff stays scoped.

On GeneReviews: there is no DESSH chapter that I could find. Worth noting for the next person that `GeneReviews[TA]` returns zero results with `phrasesnotfound` — `GeneReviews[Book]` is the filter that works.

kevinschaper:
Both fixed.

**The false clause.** You are right, and right about why it matters: the sentence had become an instruction to the next reader to re-add what is already there. It now reads "...a frequency not established for the intragenic form; it is recorded in this entry's phenotype section scoped to this subtype only." The caveat itself is kept; only the account of what the entry omits is gone.

**The `preferred_term`.** Changed to "Various visual impairments". Not cosmetic in the way I first read it — the display text was quietly asserting the thing the binding rationale declines to assert, which is exactly the "machine-readable value the prose contradicts" problem in the other direction. The `HP:0000504` binding is unchanged.

The history record was left alone, as you said — append-only.

Re-validated after the change: `just validate` 0, `just validate-terms` 0, `count-verified-snippets` 20/20. Branch rebased on current `main`; incidental reference-cache refetch churn reverted again.

PR #10818, opened 2026-09-03
curate: Add Spondyloepiphyseal Dysplasia, Nishimura Type (MONDO:0032835, MIR140)

### LINKED ISSUE #10804: Curate spondyloepiphyseal dysplasia, Nishimura type (MONDO:0032835)
Curate a dismech entry for **spondyloepiphyseal dysplasia, Nishimura type** ([MONDO:0032835](https://monarchinitiative.org/MONDO:0032835)) — *MIR140* variants; a seed-region change in a cartilage-specific microRNA, giving a skeletal dysplasia rather than the deafness that the other curated miRNA disease causes.

**First decide what this should be**: a `kb/disorders/` Disease, a `kb/groupings/` Grouping, a `has_subtypes` entry on an existing disease, or out of scope. See the Disease Groupings section of `CLAUDE.md` and #8727 for the rule. Record the decision in the stub's `entry_type` before curating.

Nothing in `kb/` mentions MIR140 and the MONDO term is bound nowhere — a clean start.

**Read `kb/disorders/Autosomal_Dominant_Nonsyndromic_Hearing_Loss_50.yaml` first.** It is the KB's other miRNA-caused disease (MIR96, DFNA50, PR #10465) and it worked out the modelling questions this entry will hit: a seed mutation is simultaneously hypomorphic for the wild-type target set and neomorphic for an acquired one, so `functional_impact_category` was left **absent** with a `notes:` explanation rather than forced; `modifier` was likewise left absent on the novel-target node because what changed is silencing *specificity*, not level. It also established that a miRNA `GeneDescriptor` needs no special shape — `GeneTerm` has no `reachable_from` constraint. Follow that pattern where the biology matches, and say so where it does not.

Stub: `stubs/Spondyloepiphyseal_Dysplasia_Nishimura_Type.yaml`

kevinschaper:
Adds `kb/disorders/Spondyloepiphyseal_Dysplasia_Nishimura_Type.yaml` and deletes the stub.

This is the knowledge base's second microRNA-caused disease, after DFNA50/MIR96 (#10465). It is also the case that made "neomorphic microRNA mutation" a real category: one nucleotide at the start of the miR-140-5p seed, the same allele in all three reported patients from two unrelated families, and a knock-in mouse whose skeleton differs from the miR-140-null mouse in exactly the features the patients have.

**Scope decision: DISEASE.** One gene, one recurrent allele, one causal chain, no member diseases to unite. The MONDO parent (spondyloepiphyseal dysplasia) is an ontology grouping of dozens of unrelated loci and is not a dismech entry.

## The two miRNA modelling questions, and where this entry disagrees with the MIR96 one

The MIR96 entry settled two things. This entry follows one of them and departs from the other, deliberately.

**Followed:** `modifier` is left absent on the acquired-target node. What changed is the specificity of the silencing, not its level, and `ModifierEnum` has no value for that.

**Not followed:** `functional_impact_category` is set to `NEOMORPHIC` here, where the MIR96 entry leaves it absent. The biology is the same shape - a seed mutation is hypomorphic for the old target set and neomorphic for the new one - but the mouse genetics is not. The miR-140-null mouse is not phenotypically silent, so this is not the MIR96 argument; it is that the null mouse lacks precisely the delayed secondary ossification, widened basal skull growth plate and expanded resting zone that both the knock-in and the patients have, while sharing the short stature and short nose. The founding authors write that the substitution is "neomorphic, and not just loss-of-function". `NEOMORPHIC` in Muller's sense names an allele with a new activity without denying the old one is impaired, and the impaired arm is curated as its own node with its own evidence. The full reasoning, including what the single value cannot record, is in the `genetic_context` notes.

A third slot is left empty for a related reason: the node recording that the mutant miRNA is produced abundantly **without** a processing defect carries no `biological_processes` descriptor, because the claim is that processing is unchanged and there is no `UNCHANGED` in `ModifierEnum`.

## What is in the entry

Fifteen pathophysiology nodes running from the seed substitution to the clinical picture, branching at the acquired-target node into the YBX1 competition (which explains why an evolutionarily naive seed represses hard), the HIF1A arm (lysyl oxidases, collagen cross-linking), and the metabolic arm from a 2023 preprint (glycolysis down, acetyl-CoA and histone acetylation down, resting zone expanded). Fifteen phenotypes, two mouse models with `modeled_mechanisms` - the knock-in as `RECAPITULATES` and the null as `FAILS_TO_RECAPITULATE` for the skeletal node, which is the load-bearing negative - two verified GEO datasets, an acrodysostosis differential, and three open discussions.

The metabolic arm is marked `PROVISIONAL` throughout. Its middle steps were established in Ldha and Acly conditional knockouts rather than in the miR-140 mutant, and the onward step to FGFR3 is deliberately not curated as a causal edge because the authors say the mechanism is not clear. Its source is also a bioRxiv preprint that has not appeared in a journal.

## Deliberately absent

- No `treatments:` block. Nothing disease-modifying exists or has been proposed, and no intervention has been tested even in the mice. Reasoning is in `sedn_no_treatment`.
- No `isds_skeletal_category`. Every other skeletal dysplasia entry here carries one with a note naming the nosology table row. The cached records of both the 2019 and 2023 nosologies are abstract-only and contain no occurrence of MIR140, so classifying this entry would be a guess dressed as a citation.
- No numeric growth data. The auxology is in a supplementary table the reference cache does not carry.
- No human tissue claims. Every mechanistic measurement in the entry is from mouse cells or mouse bone; the gap is recorded in `sedn_no_human_tissue` rather than papered over.

## Two identifier errors caught, not propagated

- The deep-research report gives ORPHA:163649 for this disease. That code is Spondyloepiphyseal dysplasia-craniosynostosis-cleft palate-cataracts-intellectual disability syndrome, a different disorder. MONDO:0032835 has no Orphanet cross-reference and the Orphanet API returns no record.
- PMID:36711926 gives the OMIM number as 611894 in both its abstract and its introduction; MONDO and the founding paper give 618618.

Both are recorded in the entry notes so the next curator does not rediscover them.

## Deep research

`falcon` (Edison), run through `scripts/run_deep_research_client.sh` directly because `just research-disorder` reads `kb/disorders/<slug>.yaml` first and cannot bootstrap a new disease. NEC preflight passes: MIR140 mentioned 54 times against 10 for the next gene, and the report's OMIM number matches MONDO's cross-reference. Its main contribution was surfacing PMID:36711926, which a search on the disease name alone does not return. Its obsolete GO:0016573 suggestion is not used; every ontology term here was resolved through OLS and checked against the label the service returned.

## Validation

All exit 0: `just validate`, `just validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history`. 80/80 snippets verified against the cache. Both GEO accessions verified with `just verify-datasets`.

`check-snippet-grading` did catch one real defect during drafting: the same clause was graded HUMAN_CLINICAL under the phenotypes and MODEL_ORGANISM under the animal model. The sentence is a mouse-to-human comparison, so it was split - the mouse half is quoted for the mouse claim and the human half for the patient features.

Closes #10804

### DIFF
--- a/cache/dataset_accessions.json
+++ b/cache/dataset_accessions.json
@@ -8179,22 +8179,6 @@
     "status": "OK",
     "title": "Specific methylation marks in promoter regions are associated to the pathogenic process of Chronic Chagas disease Cardiomyopathy by modifying transcription factor binding patterns [methylation]"
   },
-  "geo:GSE192971": {
-    "checked": "2026-09-03",
-    "detail": "",
-    "extra": {
-      "gds_type": "Expression profiling by high throughput sequencing",
-      "organism": "Mus musculus",
-      "platform": "28457",
-      "pubmed_ids": [
-        "36711926"
-      ],
-      "sample_count": 12,
-      "summary": "To test the consequence of reduced glycolysis and reduced cytoplasimic acetyl-CoA in mouse skeletal devleopment.  To reduce glycolysis, Ldha was homozygously deleted in collagen 2 expressing chondrocytes using the Cre-lox sytem [Col2-Cre:Ldha(fl/fl)] on top of one allele germline deletion of Ldhb [Ldhb(+/-)].  To reduce cytoplasmic/nuclear acetyl-CoA, Acly that catalizes conversion of citreate into acetyl-CoA was deleted in chondrocytes using Col2-Cre.  Both mutant mice develop skeletal dysplasia, but former model survive postnatally, whereas Acly mutants die at birth."
-    },
-    "status": "OK",
-    "title": "Regulatory role of energy metabolism in skeletal development"
-  },
   "geo:GSE193398": {
     "checked": "2026-08-01",
     "detail": "",
@@ -20444,22 +20428,6 @@
     },
     "status": "OK",
     "title": "RNA expression in anterior nasal septum cartilage of wildtype and Fgfr2+/S252W Apert syndrome model mice at embryonic day (E)14.5."
-  },
-  "geo:GSE98309": {
-    "checked": "2026-09-03",
-    "detail": "",
-    "extra": {
-      "gds_type": "Expression profiling by high throughput sequencing; Non-coding RNA profiling by high throughput sequencing; Other",
-      "organism": "Mus musculus",
-      "platform": "13112;19057",
-      "pubmed_ids": [
-        "30804514"
-      ],
-      "sample_count": 32,
-      "summary": "This SuperSeries is composed of the SubSeries listed below."
-    },
-    "status": "OK",
-    "title": "miR-140 mutation and skeletal dysplasia"
   },
   "geo:GSE98454": {
     "detail": "",
--- a/kb/disorders/Spondyloepiphyseal_Dysplasia_Nishimura_Type.yaml
+++ b/kb/disorders/Spondyloepiphyseal_Dysplasia_Nishimura_Type.yaml
@@ -900,7 +900,6 @@ pathophysiology:
     reference_title: "Transcriptomic profiling confirms microRNA-140 is more functional in joint development than in disease."
     supports: NO_EVIDENCE
     evidence_source: MODEL_ORGANISM
-    directness: INDIRECT
     snippet: "In contrast, its influence on post-traumatic OA is modest, with limited impact on cartilage degeneration but a role in modulating osteophyte formation."
     explanation: >-
       The counterweight, and the reason this node is not built on the osteoarthritis
--- a/references_cache/GEO_GSE192971.md
+++ b/references_cache/GEO_GSE192971.md
@@ -2,6 +2,7 @@
 reference_id: GEO:GSE192971
 title: Regulatory role of energy metabolism in skeletal development
 content_type: summary
+full_text_attempted: true
 ---
 
 # Regulatory role of energy metabolism in skeletal development
--- a/references_cache/GEO_GSE98309.md
+++ b/references_cache/GEO_GSE98309.md
@@ -1,0 +1,12 @@
+---
+reference_id: GEO:GSE98309
+title: miR-140 mutation and skeletal dysplasia
+content_type: summary
+full_text_attempted: true
+---
+
+# miR-140 mutation and skeletal dysplasia
+
+## Content
+
+This SuperSeries is composed of the SubSeries listed below.


kevinschaper:
Second commit acts on the automated research notes posted to #10804. Three things.

**The TRPS1 bridge is now in, as a hypothesis rather than an edge.** TRPS1 is one of the targets the mutant seed acquires, and heterozygous TRPS1 loss causes trichorhinophalangeal syndrome, whose diagnostic radiographic finding is cone-shaped epiphyses - the same feature that is a radiological hallmark here. That is a tidy single-target explanation for a phenotype the entry otherwise attributes to a diffuse transcriptional shift. It is recorded as a `mechanistic_hypotheses` group with `status: EMERGING`, plus one causal edge tagged into that group, because the founding paper names Trps1 in a list of acquired targets and attributes no patient feature to it. Nobody has measured TRPS1 in a mutant chondrocyte, in a patient, or in a reporter assay. The hypothesis text also records what argues against it: the two disorders share almost nothing else, so if the bridge holds it is partial.

**The osteoarthritis literature is now explicitly bounded.** miR-140 has a large osteoarthritis literature that dominates a naive search, and the joint node in this entry cited the 2010 null-mouse osteoarthritis result as indirect support. PMID:41242538 (2026) pushes back directly - a newer null mouse subjected to destabilisation of the medial meniscus shows only modest cartilage damage with no enrichment of predicted miR-140-5p targets. That is cited on the joint node as `NO_EVIDENCE`, which is the honest grading: it neither supports nor refutes a claim about the seed allele, it limits how far the null-mouse result can be pushed. The same paper independently corroborates the resting-zone finding by spatial transcriptomics, from a different laboratory and a different genotype, and is cited for that on the resting-zone node.

**Module conformance is now argued rather than silently omitted.** `fgfr_gain_of_function_skeletal_dysplasia` looks like a fit and the metabolic branch here even ends in Fgfr3 upregulation, but that module's own notes scope it to germline FGFR gain-of-function alleles and this disease has no FGFR variant. `osteoarthritis_cartilage_degradation` is wrong for the reason above. The real observation is that there is no generic growth-plate or endochondral-ossification module in `kb/modules/` despite the number of dysplasia entries; that is a proposal in its own right, not something to build inside a curation PR.

Not acted on, and why: the HPOA annotation set for OMIM:618618 gives 3/3 denominators for the core skeletal features, which would justify `OBLIGATE` where this entry uses `VERY_FREQUENT`. I have left the more conservative band because the entry's evidence comes from the paper's own prose, which states these as the disorder's features rather than as a per-patient count, and the notes say so. Happy to change it if a reviewer prefers the HPOA reading. The additional HPOA-only phenotypes (short femoral neck, broad thumb, intervertebral disk degeneration, laryngeal stenosis) are likewise not added, because they come from tables the reference cache does not carry and would go in without quotable evidence.

84/84 snippets verified. `just validate`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-folded-hyphens` all exit 0.

kevinschaper:
Fixed the blocking finding. You were right, and the argument you gave for it is the one that settles it: the sentence follows directly on from "Mice with the corresponding single nucleotide substitution show skeletal abnormalities", so the chondrocytes in it are knock-in mouse chondrocytes. Both items are now `evidence_source: IN_VITRO` with `directness: INDIRECT`, matching the five siblings that quote the same experiment.

I also audited the other twenty distinct `HUMAN_CLINICAL` snippets in the file. All are about the three reported patients - the variant calls, the clinical and radiographic descriptions, the fiberscopy, the bone density, the gene-panel exclusions and the acrodysostosis differential. Nothing else needs regrading.

`just validate-disorders` and `check-snippet-grading` both exit 0 on the new commit.

On the four non-blocking suggestions, taking them in order.

**HP:0002983 and HP:0002656.** Leaving both out. `HP:0002656` (Epiphyseal dysplasia) is the parent of the two epiphyseal features already bound - `HP:0002663` Delayed epiphyseal ossification and `HP:0010585` Small epiphyses - and the entry's rule is to bind the most specific term that the source supports. Adding the parent alongside its own descendants would add a row to the phenotype table without adding a claim, and would be the only place in the entry where a phenotype is asserted at two levels of the hierarchy. `HP:0002983` (Micromelia) is a stronger word than the source uses: the paper says "disproportionate short stature with short limbs", which is what `HP:0008873` Disproportionate short-limb short stature says, and micromelia would upgrade that.

**Preserved findings as explicit phenotypes.** I agree these carry diagnostic weight for a skeletal dysplasia, and they are cited - "All three affected individuals had normal intelligence, dentition, hearing, visual acuity, and basic blood tests" is an evidence snippet, and the normal bone density is another. What I cannot do is mark them as *excluded*. The `Phenotype` class has no `presence` slot and `FrequencyEnum` has no excluded tier, so a phenotype row saying "Normal intelligence" would be read by every consumer as a phenotype this disease has. That is worse than prose. If the schema grows an exclusion mechanism this is a good candidate entry to backfill.

**Radiographic differentials.** Not added, and I would rather say why than add them thinly. The founding paper discusses exactly one differential, acrodysostosis, and gives a discriminator that runs the right way (delayed versus advanced epiphyseal maturation). COL2A1-, TRPV4- and ACAN-related SED and multiple epiphyseal dysplasia are real radiographic neighbours, but nothing in this entry's reference cache says how to tell them from SED Nishimura type, and a `differential_diagnoses` entry whose `distinguishing_features` I wrote from general knowledge would be exactly the kind of unsourced clinical claim the evidence policy is there to prevent. It is a good follow-up for whoever curates the other SED entries, where the sources for the comparison will be at hand.

**Two uncited cached references.** Correct reading - both were pulled in by the deep-research run's own reference validation rather than by me, and its validator flagged `DOI:10.1002/jbmr.2438` as off topic. They are committed so the report's reference-validation section resolves, not because I intended to cite them.

On the GeneReviews tooling limitation you flagged: I did run the PubMed query, and there is no GeneReviews chapter for this disorder. The `[Book]` search returns the trichorhinophalangeal chapter (`PMID:28426188`, which the entry cites for the TRPS1 hypothesis) and nothing for MIR140 or SED Nishimura type. Your reasoning was right; noting it here so the conclusion is on the record rather than inferred.

kevinschaper:

kevinschaper:
Both findings fixed in one push, and you were right that they are one problem rather than two.

**The frozen cache file.** Restored from `main`. I confirmed the staged version now matches `origin/main` byte-for-byte and the working tree is clean.

**The missing GEO reference — and why the verifier was lying.** This is the part worth recording. `just verify-datasets` reported `[2/2] OK` while `references_cache/GEO_GSE98309.md` existed neither on the branch nor on `main`. It passed because it was reading the answer out of the very `cache/dataset_accessions.json` row this PR was adding. Removing the frozen-cache row is what exposed the gap — the two findings were the same defect seen from two sides, exactly as you said. I then re-ran the verifier under main's policy, it fetched GSE98309, and the file is committed, so verification after merge is deterministic rather than network-dependent.

`GEO_GSE192971.md` picked up `full_text_attempted: true` in that run. I checked before staging it: the file is added by this PR and is not on `main`, so it collides with nothing, and the newly fetched GSE98309 carries the same field — leaving them inconsistent would have been the odd outcome.

**One check I want to flag, because it acquits the tooling.** After restoring the file I re-ran the verifier and `cache/dataset_accessions.json` came back *unmodified*. So the post-#10162 verifier does not write it, and there is no residual tooling drift to file an issue about here. The 32 lines were carried in from the pre-#10162 curation commit and survived the rebase.

**The `directness` nit — taken, and it is more than a slot-semantics point.** Removed `directness: INDIRECT` from the PMID:41242538 `NO_EVIDENCE` item. `directness` grades how directly a quote bears on a claim; `NO_EVIDENCE` says it does not bear on it at all. Both cannot be true, so the pair was not merely odd-reading but contradictory. The file's other two `INDIRECT` values sit on `SUPPORT` items and are correct, so I left them alone.

A near-miss worth owning: my first attempt at that edit matched the wrong evidence item — `PMID:41242538` appears three times in the file — and stripped `directness` from a `SUPPORT`/`MODEL_ORGANISM` item where `INDIRECT` is right. I caught it in the diff, reverted, and re-cut the edit against the exact line. The committed diff touches only the `NO_EVIDENCE` block.

Validation: schema, terms, 84/84 snippets, entity refs, enum values — all pass. `just verify-datasets` is 2/2 OK with the frozen file untouched.

PR #10819, opened 2026-09-03
curate: Add Distal Myopathy 7, Adult-Onset, X-Linked (MONDO:0024771, SMPX) (#10810)

### LINKED ISSUE #10810: Curate myopathy, distal, 7, adult-onset, X-linked (MONDO:0024771)
Curate a dismech entry for **myopathy, distal, 7, adult-onset, X-linked** ([MONDO:0024771](https://monarchinitiative.org/MONDO:0024771)) — *SMPX* variants; the small muscle protein X-linked is a mechanosensitive regulator of the muscle cytoskeleton.

`MONDO:0024771` is bound nowhere. **SMPX's better-known disease is deafness, and it is already in the KB**: `kb/disorders/X-linked_Nonsyndromic_Hearing_Loss.yaml` covers DFNX4/SMPX. Read that entry first. The two phenotypes are reported from different families and it is not settled whether they are one pleiotropic disease or two — so decide explicitly whether this entry claims any auditory phenotype, and if the literature does not settle it, record a `KNOWLEDGE_GAP` rather than choosing silently. That is the treatment the SPEF2 spermatogenic-failure/PCD boundary got in #10786.

Stub: `stubs/Myopathy_Distal_7_Adult-onset_X-linked.yaml`

kevinschaper:
## The decision

`entry_type: DISEASE`. `MONDO:0024771` gets its own `kb/disorders/` entry rather than becoming a subtype of `X-linked_Nonsyndromic_Hearing_Loss`, which already carries DFNX4/SMPX.

## How I set the SMPX myopathy/deafness boundary

The issue asks whether these are one pleiotropic disease or two, and says to record a `KNOWLEDGE_GAP` rather than choose silently if the literature does not settle it. My reading is that the literature settles the **clinical** question and leaves a narrower one open, so I did both — split the entries, and curated the part that is genuinely unresolved as a gap.

What settles it: the split tracks the **variant class**, not the family.

| | DFNX4 hearing loss | This entry (MPD7) |
|---|---|---|
| Alleles | truncating / splice-disrupting | missense |
| Mechanism | loss of function | toxic gain through aggregation |
| Other organ | no reported skeletal myopathy | no hearing loss on examination into the 8th decade |

Johari 2021 (PMID:33974137) states the contrast in as many words: "Our findings suggest that the missense variants described here cause muscle disease through a gain-of-function mechanism of mutant SMPX, contrasting with the loss-of-function mutations associated with hearing loss." Lu 2024 (PMID:39272213) states the same split independently, from the other side, in a 3,646-patient X-linked hearing-loss cohort. MONDO agrees structurally: `MONDO:0024771` sits under `distal myopathy`, with no path to the hearing-loss branch.

What does **not** settle it, and is curated as `KNOWLEDGE_GAP`: the no-hearing-loss claim rests on a sentence saying no hearing loss was *recorded*, not on a screened cohort with audiograms. Aggregated third-party phenotype annotations for MPD7 do list hearing impairment, so the annotation layer and the primary literature disagree. The gap's `prompt` is therefore "has anyone systematically audiogrammed a missense-variant cohort", with a proposed audiometry experiment stratified by N-terminal vs C-terminal allele, rather than a generic boundary question.

A second discussion is `HUMAN_MODEL_MISMATCH`: the *Smpx*-deficient zebrafish shows **both** inner-ear and muscle defects, which is the strongest single argument for one pleiotropic gene — but it is a knockdown, so it models the allele class that in humans causes deafness without myopathy. The `animal_models` link is `PARTIALLY_RECAPITULATES` with `fidelity: LOW` and a `limitations` block saying exactly that. No model expresses a human missense allele at endogenous levels, so the mechanism this entry models has no in vivo model at all.

I did not modify `X-linked_Nonsyndromic_Hearing_Loss.yaml`. Its DFNX4 subtype description already says "nontruncating variants can be associated with non-auditory myopathy phenotypes", so it is not silent about the other allele class, and the reciprocal statement here is `differential_diagnoses` bound to `MONDO:0010238`.

## Where I was careful not to overstate

**The stress-granule mechanism.** The abstract says "overexpressed SMPX localized to stress granules and slowed down their clearance", which reads as a disease mechanism. The discussion section says: "as our experiments did not reveal major differential effects on SG dynamics between wild-type and mutant SMPX, the potential role of SGs in the pathomechanism remains unclear". The abstract sentence is about SMPX, not about the mutant alleles. So the node exists, its incoming edge is `causal_link_type: UNKNOWN`, and it carries a `NO_EVIDENCE` item quoting the authors' own limitation. It makes no downstream claim.

**Amyloid.** Congo red positivity with apple-green birefringence is reported, but for one patient's biopsy, and the authors say "amyloid-like". Recorded in the histopathology description and evidence; **no** `conforms_to: amyloidogenesis#...` declared.

**Autophagy.** The chaperone-assisted selective autophagy explanation for the myofibrillar lesions is offered by the source as "one possibility". Not modelled as a node, and no `disabled_macroautophagy` conformance.

**Inheritance.** Bound to the general `HP:0001417` (X-linked), not to X-linked recessive or dominant. No female heterozygote has been phenotyped, so there is no basis to choose, and published tabulations of this disease disagree with each other.

**Aggregation propensity.** The `MOLECULAR` node carries no GO term. GO has no term for a protein's aggregation propensity, and a generic folding or oligomerization term would assert an activity the source does not report. The consequence is bound at the next node with `GO:0070841`.

**Frequency.** `CASES_IN_LITERATURE` / `ULTRA_RARE`, no rate. Ten patients, nine families. The two founder haplotypes mean occurrence is regionally clustered, so a single worldwide rate would misdescribe it; that is in `notes`.

- PMID:38615630 (axial involvement) and PMID:39230604 (scapuloperoneal/myofibrillar) have **no abstract indexed**, so no snippet could be verified against the cache. Neither is cited as evidence. The axial finding is instead carried through PMID:39017652, a 2024 review that summarizes it in quotable form.
- The 2024 review and the primary report **disagree** about which thigh muscles go first at the late stage — hamstrings (primary) vs adductors and quadriceps (review). The imaging finding follows the primary report and records the discrepancy in its `notes` and in the second evidence item rather than picking one silently.
- Module conformance is declared only to `loss_of_proteostasis#Misfolded-Protein Aggregation`.

## Validation

Every check run on this branch; all exit 0.

| Check | Exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just count-verified-snippets` | 0 (32/32 verified) |
| `check-entity-refs` / `check-duplicate-keys` | 0 / 0 |

`just preflight-dr … MONDO:0024771` → **PASS**, SMPX mentioned 56 times, top genes SMPX=56. Report reference validation: 8/8 verified, confabulation rate 0.

Every `reference_title` was checked programmatically against its cache frontmatter — one was wrong on first pass and is fixed.

Closes #10810


kevinschaper:
All six blocking items addressed. Two of them with a different term than suggested — details below, because in both cases I checked the term and it did not say what the review said it said.

## Where I substituted a term

**`HP:0009099` is "Median cleft palate", not finger-extensor weakness.** I ran `runoak -i sqlite:obo:hp info HP:0009099` before using it. The right term is **`HP:0009077` Weakness of long finger extensor muscles**, which is what I bound, on the sentence the entry already cited: "In the probands of families F1–F5, onset of finger extensor/ankle dorsiflexor weakness was observed in the 3rd–4th decade." For the ankle side I used **`HP:0009027` Foot dorsiflexor weakness** rather than `HP:0009053`, which is "Distal lower limb muscle weakness" — accurate but no more specific than the `HP:0002460` node the finding was already living in, so it would not have answered the point.

**`HP:0003691` Scapular winging is not what the source says.** The quoted sentence is "progressed to scapular arm abduction muscles" — that is shoulder-girdle weakness. Winging is a distinct clinical sign (scapular stabiliser failure visible on protraction) and no cited source reports it in these patients. I bound **`HP:0003547` Shoulder girdle muscle weakness** and put the reason in the phenotype's own `description`, so the next reader does not re-propose winging.

## The rest

| Item | What I did |
|---|---|
| 1, 2 — cardinal phenotypes | Added finger extensor, foot dorsiflexor, shoulder girdle. Three new phenotypes, four new evidence items. |
| 3 — missing alleles | Added p.A13V and p.P27A. You are right that this was internally inconsistent: the proposed experiment stratified by N-terminal vs C-terminal allele while only one N-terminal allele was curated. |
| 4 — `HYPOTHESIZED` in prose | Fixed. The edge itself was already `UNKNOWN`; only the `explanation` named a value that does not exist. |
| 5 — duplicate cache files | Both `DOI_10.1097_WCO…` and `DOI_10.1097_wco…` removed. The entry cites the PMID. |
| 6 — no `diagnosis:` | Added four entries: SMPX sequencing (with the point that a prior negative neuromuscular panel does not exclude it, since the defining cohort was found only after those genes were excluded), lower-limb MRI as supportive, muscle biopsy where the SMPX immunostain rather than the rimmed vacuoles is the discriminator, and cardiac/respiratory assessment. |

Suggestions taken: cardiac and respiratory sparing is now an evidenced item inside `diagnosis`, since as you say it is clinically load-bearing; the treatment is split into physical therapy (`NCIT:C15302`, `BEHAVIORAL`) and ankle-foot orthosis / mobility aids (`DEVICE`); and a `classifications` block matching the sibling MPD6 entry.

The orthosis entry carries **no** `treatment_term`. NCIT has no clinical-action term for orthosis use reachable from Clinical Intervention or Procedure — which is the same gap `CLAUDE.md` records for device usage generally — so it stays unbound with a note rather than taking a wrong term.

Suggestion not taken: elevated CK and myopathic EMG, for exactly the reason you gave. The cached text says only that the measurements "were obtained"; the values are in Table 1, which is not in the cache. I would rather leave the gap than stretch a quote.

## Re-validated

`just validate` 0, `just validate-terms` 0, `count-verified-snippets` **43/43**, `check-entity-refs` 0, `check-causal-targets` 0, `check-duplicate-keys` 0, `check-snippet-grading` 0, `check-snippet-length` 0, `check-title-snippets` 0, `check-folded-hyphens` 0, `validate-history` 0. Branch rebased on current `main`; incidental reference-cache refetch churn reverted.

On GeneReviews: I checked and there is no SMPX myopathy chapter. Worth recording for the next run that `GeneReviews[TA]` returns zero with `phrasesnotfound` — `GeneReviews[Book]` is the filter that works.

kevinschaper:
Took the main non-blocking suggestion: the new phenotypes were rendering disconnected.

Four phenotypes had no incoming edge — the three weakness nodes from the last commit plus the axial one added earlier. All four now hang off `Myofibrillar lesion formation and myofibre degeneration` with `INDIRECT_UNKNOWN_INTERMEDIATES`, matching the link type its three existing phenotype edges already use. No evidence, terms, or claims changed; this is graph wiring only.

Worth noting why it slipped through: `check-causal-targets` validates the targets that *were* written, so it cannot see a phenotype that nothing points at. It reported clean both before and after this commit.

Re-validated: `just validate` 0, `check-entity-refs` 0, `check-causal-targets` 0, `count-verified-snippets` 43/43 unchanged.

Sorry to drop the approval for this — it seemed worth doing while the PR is still outside the merge window rather than leaving a disconnected subgraph behind. The other two suggestions I have left for a follow-up pass.

kevinschaper:

Issue #8923, opened 2026-08-19
Review cardiac, facial, and digital anomalies with developmental delay (MONDO:0032572) against its MONDO parent


sierra-moxon:
**This is a review/scoping task, not a straightforward new-entry curation.**

The [priority dashboard](https://dismech.monarchinitiative.org/dashboard/priority.html) recommends `REVIEW_AGAINST_PARENT` for **cardiac, facial, and digital anomalies with developmental delay** ([MONDO:0032572](https://monarchinitiative.org/MONDO:0032572)), score 46.0. The `claim-disease` skill treats only `CURATE_ROOT` and `CURATE_ROOT_WITH_SUBTYPES` as actionable curation targets, so this was retitled and relabelled from a `Curate …` issue rather than being dispatched as one.

**Context:** MONDO's only parents for this term are the very broad *hereditary disease* / *syndromic disease*, so there is no close curated parent to compare against. The review question is whether this term is specific enough to stand as its own entry, or is a thin OMIM-derived label better represented elsewhere.

## What to decide

1. Compare `MONDO:0032572` against its MONDO parent(s) and any curated near-neighbour named above.
2. Decide one of:
   - **Distinct entry** — curate it, and state explicitly in the PR what it adds beyond the parent; or
   - **Bind onto an existing entry** — add `MONDO:0032572` as a `has_subtypes` term or a `mappings.mondo_mappings` entry (`skos:exactMatch`/`narrowMatch` counts as curated coverage per `CLAUDE.md`); or
   - **Neither** — close with the rationale recorded.
3. Record the decision in the closing PR or comment either way.

**NEC check** before curating from any deep-research output: anchor identity on the OMIM xref and verify the causal gene with `uv run runoak -i sqlite:obo:mondo info MONDO:0032572 -O obo`, then run `just preflight-dr <report.md> MONDO:0032572`.

Tracker: part of #1079.

Issue #10798, opened 2026-09-03
refresh-orphadata: pinned snapshot no longer downloadable, so new ORPHA cache files carry a false provenance line


kevinschaper:
`just refresh-orphadata` cannot reproduce the snapshot pinned in `data/orphadata/MANIFEST.yaml`. Upstream serves a newer file behind the same unversioned URL:

```
downloading https://www.orphadata.com/data/xml/en_product1.xml ...
checksum mismatch for en_product1.xml after download:
  expected (manifest pin): fb2fbe8cc3f04c2ffb3fac3c498bce75e56cfc46b5475e3da30e97f8349140db
  actual   (upstream now): df8d562a0c6011af36a74eb4000ce81ca7d723e8031010819fb71727c0962bbb
```

The manifest anticipates this and calls it expected drift. The part that bites is what happens next.

**The file is left in place and usable, and the rebuilder will happily use it.** `just structured-rebuild-orphanet --id N` then produces a cache file whose `## Source` line reads `Orphadata snapshot **2025-12-09**` — because the serializer takes `snapshot_date` from the manifest, not from the file it actually parsed. So a cache file built today names a release it was not built from, and drops into a corpus where every other `ORPHA_*.md` genuinely is from the pinned release. Anyone who later runs a full rebuild gets a diff they cannot attribute.

The abort also stops after the first file, so only `en_product1.xml` downloads. A rebuild then silently emits a record with no `## Phenotypes`, `## Genes` or `## Epidemiology` sections — indistinguishable from a disorder that genuinely has none.

Hit while curating BDPLT19 (#10797). ORPHA:438207 ("Severe autosomal recessive macrothrombocytopenia", exact match to MONDO:0014518) has a definition enumerating the bleeding pattern — epistaxis, spontaneous hematomas, menorrhagia — which is exactly the per-symptom phenotype detail that entry lacks. I discarded the generated file rather than committing it, and the entry records why, so those phenotypes are uncurated purely because of this.

Two things that would help, roughly in order of value:

**Take `snapshot_date` from what was parsed, not from the manifest.** A cache file should never be able to claim a provenance it does not have. Even just refusing to serialize when the on-disk file's sha256 does not match the pin would be an improvement over emitting a mislabelled record.

[...]

PR #10820, opened 2026-09-04
curate: Add Cardiac, Facial, and Digital Anomalies with Developmental Delay (MONDO:0032572, TRAF7) (#10808)

### LINKED ISSUE #10808: Curate cardiac, facial, and digital anomalies with developmental delay (MONDO:0032572)
Curate a dismech entry for **cardiac, facial, and digital anomalies with developmental delay (CAFDADD)** ([MONDO:0032572](https://monarchinitiative.org/MONDO:0032572)) — de novo *TRAF7* missense variants clustering in the WD40 repeats, disrupting a scaffold that sits on MAP kinase and NF-kB signalling.

`MONDO:0032572` is bound nowhere. TRAF7 appears only in `kb/disorders/Meningioma.yaml` — and that is the trap: those are **somatic** TRAF7 mutations in tumour tissue, a different disease and a different mutational mechanism from the germline de novo variants here. Some residues recur between the two, which makes citation-level checking necessary rather than optional; gene-frequency preflight will not separate them.

Stub: `stubs/Cardiac_Facial_And_Digital_Anomalies_With_Developmental_Delay.yaml`

kevinschaper:
## The decision

`entry_type: DISEASE`. `MONDO:0032572`'s only parents are `syndromic disease` and `hereditary disease`, both top-level groupings, so there is nothing this could be a subtype of and nothing in `kb/` it duplicates.

## Heads-up: this MONDO term is open on two issues

**#8923** ("Review cardiac, facial, and digital anomalies with developmental delay (MONDO:0032572) against its MONDO parent") is the same term, assigned to @sierra-moxon. It carries no PR. It was retitled from `Curate …` to `Review …`, which is why `just check-claims` could not see it holding this MONDO ID — the claim key is the `Curate <label> (MONDO:NNNNNNN)` title pattern. Someone should decide which issue is live and close the other; I have not touched #8923. Its research summary was read and used.

## How I kept somatic TRAF7 out

TRAF7 previously appeared in `kb/` only in `Meningioma.yaml`, as a recurrent **somatic** driver in tumour tissue. Gene-frequency preflight cannot separate that from this, so the check was done at the **sentence** level rather than at the paper level.

- Every citation in this entry was read to confirm it is about germline syndrome patients.
- One paper spans both diseases — PMID:37043537, "Pleiotropic role of TRAF7 in skull-base meningiomas and congenital heart disease". It is cited only for sentences that explicitly name the germline or inherited arm, or the shared molecular mechanism (dominant-negative heterodimerization, TRAF7–IFT57 disruption, the neural-crest argument). No sentence about meningioma is used to support a claim here.
- No tumour predisposition is asserted anywhere in the entry.

The residue-level overlap the issue warned about is real, and it is curated rather than glossed. Two independent sources state that the germline and somatic **mutation** sets do not overlap:

> "most reported germline and somatic mutations are recurrent across patients, either as germline or somatic, without overlap between mutations that occur in the germline versus those that are somatic. The reason somatic and germline mutations are mutually exclusive is unclear." (PMID:41372821)

> "Although almost all identified mutations are recurrent (the same in different tumors or patients), somatic and germline mutations do not overlap." (PMID:37583551)

And PMID:38569228 reports a germline p.Arg653Leu adjacent to somatic hotspots, which "underscores the complex relationship between germline TRAF7 variants and somatic changes linked to meningiomas". So **residue-level proximity and mutation-level exclusivity are separate facts**, and the entry states them separately. Why they are exclusive is a `KNOWLEDGE_GAP` with two proposed experiments — an allele-matched interactome comparison, and a tumour-incidence cohort.

## The one thing I curated as an active dispute

There is a published PNAS exchange over whether **inherited** (as opposed to de novo) TRAF7 variants cause isolated congenital heart disease:

| | |
|---|---|
| PMID:37043537 | reports three inherited variants in a CHD cohort as pathogenic |
| PMID:38466850 | letter disputing all three, on gnomAD frequency, inheritance from healthy parents, gnomAD pLI 0.02, and absence of the syndrome's typical features |
| PMID:38466853 | the authors' reply, defending them on functional and in silico grounds |

Both sides are in print and neither has withdrawn, so this is a `CONTROVERSY` discussion carrying both, and the entry does **not** assert incomplete penetrance. Note the disputed variants are also the only argument for haploinsufficiency in TRAF7, so the outcome affects the mechanism claim, not just the inheritance claim — that is said in the discussion's `rationale`.

One pleasing detail: the two sides **agree** that the variant pattern argues against haploinsufficiency, and the sentence supporting that on the mechanism node is quoted from the disputing letter.

**Per-feature frequencies.** The widely quoted percentages over 68 aggregated cases (developmental delay 90.7%, PDA 58.2%, and so on) live in full-text tables that are not in this repository's reference cache, so no snippet could verify them. Rather than copy unverifiable numbers, phenotypes carry no `FrequencyEnum` and the descriptions record only the qualitative ordering the abstracts do support: PDA as the most frequent cardiac lesion, blepharophimosis as the leading facial feature.

**Developmental delay as obligate.** The disease name says "with developmental delay", but the 2024 series reports individuals whose early psychomotor delay resolved into normal development, so the phenotype description says frequent, not defining. PMID:41201137 (a patient with no delay at all) has no abstract in the cache and is therefore mentioned in `notes` but not cited as evidence.

**Module conformance.** `ciliopathy_dysfunction` is a candidate on the cilium result, but IFT57 is an IFT-B component while the module's nodes are built around basal body and transition zone dysfunction — conforming would assert a compartment the source does not name. `cranial_suture_premature_fusion` was rejected because craniosynostosis is a minority finding here, not the entry's mechanism. Both recorded in `notes`.

**An unresolved direction in the mechanism.** Tokita found *reduced* ERK1/2 phosphorylation from germline mutants, but wild-type TRAF7 *inhibits* MEK1/2, so naive loss of function should raise ERK. The node records the measured direction and states plainly that the reconciliation with pathway topology is not asserted, with both the measurement and the topology cited.

**Mouse model.** The `Traf7` null link is `FAILS_TO_RECAPITULATE` with `limitations` and `evidence`, because complete loss is embryonic lethal in mouse while human heterozygotes survive with a congenital-anomaly syndrome — so the model tests the mechanism the human variant pattern argues against. No knock-in of a recurrent human allele exists.

## Validation

Every check run on this branch; all exit 0.

| Check | Exit |
|---|---|
| `just validate` | 0 |
| `just validate-terms` | 0 |
| `just count-verified-snippets` | 0 (44/44 verified) |
| `just validate-disorders <this file>` | 0 |
| `check-entity-refs` / `check-duplicate-keys` | 0 / 0 |

`just preflight-dr … MONDO:0032572` → **PASS**, TRAF7 mentioned 64 times, top genes TRAF7=64, IFT57=7, SCRIB=6. Report reference validation: 8/8 verified, confabulation rate 0.

Every `reference_title` was checked programmatically against its cache frontmatter. One was wrong on the first pass — I had attached the title of PMID:34513876 to PMID:34247275 — and is fixed. Worth noting because `reference_title` **is** validated against the cache, so this would have failed CI rather than shipping quietly.

Closes #10808


kevinschaper:
All four blocking items addressed, and the suggestion about Kim 2024 turned out to be the key that unlocked most of the rest.

## Phenotypes (findings 1 and 2)

Five added, each on a sentence I could verify against the cache:

| Phenotype | HPO | Source |
|---|---|---|
| Hypotonia | `HP:0001252` | PMID:38612512 — "In the early infant period, he showed generalized hypotonia, and growth motor slowly improved in childhood." |
| Seizure | `HP:0001250` | PMID:38612512 — "Some individuals with this condition may also experience seizures and exhibit ocular or aural abnormalities" |
| Abnormality of vision | `HP:0000504` | same sentence |
| Brain imaging abnormality | `HP:0410263` | PMID:32459067 (dysgyria, vermis hypoplasia) **and** PMID:38612512 (diffuse cerebral atrophy) |
| Obstructive sleep apnea | `HP:0002870` | PMID:38612512 |

You were right that the `notes` rationale covered the frequency qualifier and not the phenotypes themselves. Fixed, and the note now says so.

The sleep apnea is its own phenotype rather than a refinement folded into sleep disturbance, because Kim attributes it to the syndrome's **own** vertebral anomaly and short neck — "likely exacerbated by the vertebral anomaly and short neck associated with CAFDADD syndrome" — so it is mechanistically downstream of the skeletal phenotype rather than a coincidental comorbidity. It also drove dyspnoea that persisted after aortic root replacement until CPAP.

## Kim 2024 (finding 3)

Now cited nine times. It is the source for four of the five new phenotypes and for most of the `diagnosis:` section. Good catch — I had fetched it during triage and then built the entry from the cohort papers.

## `diagnosis:` (finding 4)

Two entries. **Trio exome or genome sequencing**, carrying three things worth a curator's attention: karyotype and chromosomal microarray were normal in a confirmed case, so a prior negative array does not exclude the diagnosis; parental segregation established de novo status; and there is a worked ACMG classification for the recurrent p.Arg655Gln allele (PS2, PM1, PM2, PP1, PP2, PP3). **Baseline cardiac assessment**, which I framed as continued rather than one-off, because the only adult patient reported developed a bicuspid aortic valve with aortic root aneurysm needing a Bentall procedure at 36.

## Suggestions

`external_assertions` added for OMIM:618164 and ORPHA:592570. I did **not** build an ORPHA cache record: the Orphadata rebuilder takes `snapshot_date` from the manifest, so a file built today would claim a release it was not built from (#10798). Recorded as a registry cross-reference with that reason in its `notes`.

**The endothelial branch I did not add as a pathophysiology node**, and I want to say why rather than just decline. The MEKK3–MEK5–ERK5–KLF2 result is solid in the mouse, but nothing connects it to this syndrome: the human cardiovascular lesions are structural malformations (PDA, septal, valvular), not vascular-integrity failure; no patient has been reported with the cerebral haemorrhage that endothelial deletion causes; and the mouse result comes from complete loss, which is the mechanism the human variant pattern argues against. A node would assert a link nobody has shown. It is now an `EMERGING_HYPOTHESIS` discussion instead, which also carries the `Cerebral_Cavernous_Malformation` cross-link you and #8923 both flagged, since CCM is built on the same axis.

Sleep disturbance sharpened with the OSA finding, as suggested.

## What is still out, and why

Four features from the 68-case table remain uncurated because no cached source states them: motor delay distinct from global developmental delay, palate and mouth anomalies, pes planus or valgus, and scoliosis. **Scoliosis is worth flagging specifically**: PMID:38612512 mentions it only in its comparison with Loeys-Dietz syndrome ("The common features of LDS include aortic aneurysm, arterial tortuosity, arachnodactyly, pectus deformity, scoliosis…"), not as a finding in the TRAF7 patient. Quoting that sentence would attribute an LDS feature to this syndrome. These are gaps in what can be verified, not claims the features are absent, and `notes` now says that.

## Re-validated

`just validate` 0, `just validate-terms` 0, `count-verified-snippets` **56/56**, `check-entity-refs` 0, `check-causal-targets` 0, `check-snippet-grading` 0, `check-snippet-length` 0, `check-title-snippets` 0, `check-folded-hyphens` 0, `validate-history` 0. All 20 `reference_title` values re-checked programmatically against cache frontmatter: 0 mismatches. Branch rebased on current `main`.

PR #10824, opened 2026-09-04
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 124 (MONDO:0968981, PKHD1L1)

### LINKED ISSUE #10805: Curate autosomal recessive nonsyndromic hearing loss 124 (MONDO:0968981)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 124 (DFNB124)** ([MONDO:0968981](https://monarchinitiative.org/MONDO:0968981)) — biallelic *PKHD1L1* variants; a large stereocilia-surface protein required for the hair-bundle coat, so the lesion is in the bundle's extracellular surface rather than in transduction or the synapse.

**First decide what this should be**: a `kb/disorders/` Disease, a `kb/groupings/` Grouping, a `has_subtypes` entry on an existing disease, or out of scope. See the Disease Groupings section of `CLAUDE.md` and #8727 for the rule. Record the decision in the stub's `entry_type` before curating.

Nothing in `kb/` mentions PKHD1L1 and the term is bound nowhere. Several cochlear entries were curated recently — `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_103` (CLIC5, taper region) and `_79` (TPRN) are the nearest compartment neighbours; read one for shape and make the entries agree about what is established for the bundle surface.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_124.yaml`

kevinschaper:
Adds `kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_124.yaml` and deletes the stub.

**Scope decision: DISEASE.** One gene, one proximal lesion, one clinical picture. The MONDO parent is an ontology grouping of dozens of unrelated DFNB loci and is not a dismech entry.

## What is different about this lesion

The cochlear entries curated alongside this one fail inside the hair bundle: DFNB103 (CLIC5) at the stereocilia taper, tethering membrane to actin core; others in transduction or at the synapse. PKHD1L1 is none of those. It is a 4243-residue protein that is almost entirely extracellular, and it is the first identified molecular component of the electron-dense surface coat that had been described morphologically for decades without anyone knowing what it was made of.

The lesion is compartmentalised twice over. It is on the outside of the membrane rather than inside the cell, and it is on the *upper* stereocilia only - mice lacking PKHD1L1 lose the coat at the tips while the coat at the bases is unaffected. That second point is the one claim in this entry that a neighbouring entry might want to carry across: "the stereocilia coat" is not a single structure.

## The ordering is the substance of the entry

Development is normal, and a stack of negative results establishes that: bundle cohesion is undisturbed, tectorial-membrane attachment crowns form, stereocilin localises correctly, the bundle imprints in the tectorial membrane are still there at six months, planar cell polarity is normal, and FM1-43 loads normally so transduction channels are open and working. Only from about six weeks do stereocilia start to disappear - shortest row first, high-frequency base worst - and bundles lose coherence.

So this is a failure to *maintain* a correctly built bundle. That separates PKHD1L1 from the link and transduction genes, whose mutants have severely damaged bundles from the start, and it is why the FM1-43 result is curated as evidence rather than mentioned in passing.

## Three decisions I would flag for a reviewer

**Module conformance is declared only at the terminal node.** The `sensorineural_hair_cell_loss` module's chain runs insult → ionic homeostasis and oxidative stress → mechanotransduction failure and death. Conforming to the mechanotransduction node would assert something the data contradict, since transduction is demonstrably intact here; and nobody has looked for ionic or oxidative disturbance in PKHD1L1 deficiency at all. The reasoning is on the node. For context, `_103` and `_104` declare no conformance at all while `Noise_Induced_Hearing_Loss` conforms at five nodes, so this entry sits between them on purpose.

**The noise result is an environmental entry, not a treatment.** A noise dose that gives wild-type mice a temporary threshold shift gives PKHD1L1-deficient mice a permanent one. That is a real gene-environment interaction and it is curated as one: an `environmental:` entry with `influences_mechanisms` → `EXACERBATES` pointing at a `PROVISIONAL` node. It is deliberately *not* curated as "avoid noise exposure" under `treatments:`, because a treatments entry reads as a recommendation for these patients, and the recommendation has not been made by anyone who has seen one. The human paper names age and noise trauma as untested. That sentence is cited as `NO_EVIDENCE` on the model link, which is what fixes the `fidelity: UNKNOWN` call - the experiment is clean, but there is no human comparator to be faithful to.

**The Family 3 confound is on the phenotype, not in a footnote.** The proband with the most severe loss is homozygous for PKHD1L1 p.(His2479Gln) *and* homozygous for MYO7A p.(Leu375Val), both at more than 150 reads. MYO7A was deprioritised on neutral in-silico predictions, which is reasonable triage and is not functional evidence; the authors leave all three readings open. So the severe pole of this disease's range rests on the least clean family. `Severe Hearing Impairment` is therefore `OCCASIONAL` with the caveat recorded on it, and there is an open discussion with a proposed experiment.

## HPOA annotations not used, and why

The Monarch/HPOA set for MONDO:0968981 carries five phenotypes, all sourced to PMID:38459354: sensorineural hearing impairment, abnormal vestibular function, abnormality of the eye, intellectual disability, and abnormality of the cardiovascular system. The bottom three contradict the paper they cite, which reports the ophthalmology exam, the electrocardiogram and development as normal and states that syndromic involvement was excluded in all four probands. The most likely explanation is an import that lost the negation from an OMIM clinical synopsis. The vestibular annotation traces to one resolved episode of benign paroxysmal positional vertigo. None is curated; the reasoning is in the entry notes. This is worth reporting upstream and may not be specific to this term.

## Also in the entry

A `HUMAN_MODEL_MISMATCH` discussion on the onset divergence - mouse hears normally then deteriorates, humans and zebrafish are impaired from the start - with the candidate explanations and what would test them. Two animal models with `modeled_mechanisms` and readouts including two real negatives (`UNCHANGED` stereocilin localisation, and the model-link `NO_EVIDENCE` above). Hearing amplification as the one reported treatment, with cochlear implantation argued out rather than imported.

## Deep research

`falcon` (Edison), run through `scripts/run_deep_research_client.sh` directly because `just research-disorder` cannot bootstrap a new disease. NEC preflight passes: PKHD1L1 mentioned 50 times, and it correctly identifies MYO7A as the competing gene in one family rather than as a second cause. All 11 references resolved. It added nothing the primary literature does not carry, which is expected for a disease with one clinical paper; its useful contribution was independent confirmation that no trial and no Orphanet record exist. Its one obsolete term suggestion is not used.

## Validation

All exit 0: `just validate`, `just validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history`. 62/62 snippets verified.

One snippet had to be re-sourced: the abstract sentence listing all six alleles uses a bracketed phase notation (`p.[(Gly129Ser)];p.[(Gly1314Val)]`) that the reference validator strips before matching, so the two alleles are quoted from the case descriptions instead. That is recorded in the evidence explanation so the next curator does not repeat the attempt.

Closes #10805


kevinschaper:
Fixed the blocking defect. `NCIT:C50076` does name Motherboard Device; I checked it against OLS before changing it rather than taking the reviewer's word, and the correction to `NCIT:C183182` is confirmed both ways. `just check-qualifier-terms-online` exits 0 on the file now.

The mechanism by which that got in is worth recording for anyone else adding a device qualifier: linkml-term-validator does not descend into `qualifiers` (#10197), so `just validate-terms` and `just validate-disorders` are both blind to a qualifier-only CURIE, and a CURIE used nowhere else never enters the term cache to be caught offline either. `check-qualifier-terms-online` is the only gate that sees it, and it is not in the standard pre-PR list in CLAUDE.md. I have added it to my own loop.

**Three of the four suggestions are taken, and the first one changed the entry's argument rather than adding to it.**

The expression time course is the substantive one. PKHD1L1 protein is on the bundle only for the first ten postnatal days, is largely gone by P10, and is completely cleared from mature hair cells by P21 - three weeks before the bundle starts to fail at six weeks. So the protein is not present to perform ongoing maintenance, and "maintenance" in the entry's own framing had to be sharpened: what its presence during development sets is how durable the finished bundle is, which is also how the authors put it. Two quotes are now on the bundle node, and the same observation is now the *leading* explanation in `dfnb124_onset_species_mismatch` rather than one of three untested candidates - because if the lesion is set during a developmental window, and the human equivalent of that window closes before birth, a congenital human phenotype is what the same mechanism predicts. That is a much better-supported reading than the one the discussion had.

Genetic counselling and cascade testing added to `treatments:`, cited on the confirmed parental carrier status and on the consanguineous family structure, which is where recurrence counselling actually has purchase and where family history is least informative.

The Family 4 reanalysis added to `diagnosis:`, with the point that matters clinically: that exome had already been reported negative and was solved only after the gene was named, so a negative exome predating 2024 does not exclude this diagnosis. Its cohort of 1450 hearing-loss probands is now in `prevalence:` as the one real denominator anywhere in the series - a ceiling on how common this is among people already referred for deafness genetics, not a population figure.

**The AAV suggestion is the one I am not taking, and I want to be clear it is for lack of a source rather than disagreement.** The point is a good one - a ~13-kb CDS against a ~4.7-kb single-AAV capacity is exactly the negative a reader would want for a hearing-loss gene in 2026. But `DOI:10.1016/j.medj.2025.100886` is cached abstract-only and contains no statement about packaging capacity, oversized genes, dual-AAV strategies or the PKHD1L1 CDS length; I checked for all of those. Writing it from general knowledge would be an unsourced claim in a section where the entry is otherwise strict, and stating the CDS size myself would be arithmetic on the protein length rather than a citation. If a review with the relevant sentence gets cached, it is a one-line addition.

Also agreed on GeneReviews: I ran the `[Book]` query and there is no chapter for DFNB124, which is expected for an entity designated in 2024 with one clinical paper.

69/69 snippets verified. `just validate-disorders`, `check-qualifier-terms-online`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-environmental-evidence`, `check-folded-hyphens` all exit 0.

caufieldjh:
@copilot please fix the merge conflicts in this pull request.

caufieldjh (approved):

PR #10825, opened 2026-09-04
curate: Add Mitochondrial Complex I Deficiency, Nuclear Type 13 (MONDO:0032618, NDUFA2)

### LINKED ISSUE #10806: Curate mitochondrial complex I deficiency, nuclear type 13 (MONDO:0032618)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 13** ([MONDO:0032618](https://monarchinitiative.org/MONDO:0032618)) — biallelic *NDUFA2* variants; a small accessory subunit of the complex I matrix arm.

`MONDO:0032618` is bound nowhere and nothing in `kb/` mentions NDUFA2. The KB now holds MC1DN types 1, 6, 14, 23, 24 and 34 as standalone entries, so per-subunit entries are settled convention — read a recent one for shape.

**Check the Leigh relationship for this term specifically.** MC1DN1 went standalone partly because MONDO does *not* place it under Leigh syndrome, which was the structural difference from the VWM5 case; do not assume that answer transfers. `kb/disorders/Leigh_Syndrome.yaml` binds several complex I genes in its `genetic:` section, so check whether it binds this one.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_13.yaml`

kevinschaper:
Adds `kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_13.yaml` and deletes the stub.

**Scope decision: DISEASE, standalone**, consistent with MC1DN types 2, 6, 12, 23, 24, 31 and 34 already in `kb/disorders/`.

## The Leigh check, done for this term rather than assumed

The issue asked not to carry the MC1DN1 answer over without checking, so both halves were checked:

- `MONDO:0032618` has exactly one superclass, `MONDO:0100223` (mitochondrial complex I deficiency, nuclear type). MONDO does **not** place it under Leigh syndrome, so the MC1DN1 precedent transfers rather than the VWM5 subtype one.
- `kb/disorders/Leigh_Syndrome.yaml` does **not** bind NDUFA2. Its `genetic:` block binds NDUFS4, NDUFV1, SURF1, LRPPRC, MT-ATP6 and PDHA1, so there is no gene-level overlap to resolve either.

There is a real tension underneath that clean structural answer, and the entry curates it rather than smoothing it. **Only one of the four reported patients had Leigh disease.** The other three had cystic, cavitating or tigroid-like leukoencephalopathy and survived into childhood. So MONDO's placement, the OMIM label and the founding paper's title all point one way while the accumulated case series points another. Leigh syndrome is curated as an `OCCASIONAL` phenotype and leukoencephalopathy as the `VERY_FREQUENT` one, and the allele-class hypothesis that might reconcile them is an open question rather than a mechanism.

## The mechanism is assembly, not catalysis

NDUFA2 is a 99-residue accessory subunit with a thioredoxin-like fold that sits on the matrix arm and has **exclusive contacts with one other subunit**, the core subunit NDUFS1 - at least 40 angstroms from any iron-sulfur cluster, so it cannot be doing chemistry. Knocking it out of a cell line leaves the N-module only partly assembled: NDUFS1, NDUFV1 and NDUFV2 go with it, and with them the NADH binding site. That is why activity falls further than the residual complex would predict, and it is the distinction the pathophysiology nodes carry.

Both patient missense residues, Lys45 and Glu57, map to that NDUFS1 interface. That node is marked `PROVISIONAL`: the interface is established from cryo-EM structures, but the step from a patient residue to a destabilised contact is a structural inference and the source says so ("It is likely that the binding of NDUFA2 to NDUFS1 is destabilized"). Nobody has measured NDUFA2 abundance or complex I assembly for either missense allele in a patient cell.

## Three things I would flag

**The genotype-phenotype pattern is a discussion, not a mechanism.** The one fatal infantile case carries a demonstrated null - no protein, no transcript. The three leukodystrophy cases carry interface missense alleles. Residual function explaining severity is the obvious reading and it is consistent with everything published, but nothing has been measured, one of the milder patients carries a frameshift on her second allele, and four patients across three laboratories and twelve years are not a series. `functional_impact_category` stays `LOSS_OF_FUNCTION` for the block rather than being split, with the reasoning in the `genetic_context` notes.

**Decreased circulating carnitine is not curated, and the reason matters.** It appears in the HPOA set for OMIM:618235, sourced to the first 2018 patient - who also carried a homozygous SLC22A5 variant and a separate diagnosis of primary systemic carnitine deficiency, made *before* the complex I deficiency was suspected. Curating it from the HPOA row would import a dual-diagnosis confound as an NDUFA2 phenotype, and in any export it would be indistinguishable from an observed feature. In a four-patient entity, one patient's second diagnosis is a quarter of the apparent phenotype. There is an `INTERPRETATION` discussion recording this.

**A structural review carries much of this entry's detail, and it is graded as second-hand.** The three primary reports are cached as abstracts only. PMID:33233646 supplies the null status of the founding allele, the cardiomyopathy and the age at death, the p.Lys45Thr patient's course, the placement of both missense residues, and the knockout cell-line assembly data. Evidence taken from it is graded `OTHER` where it restates another paper's patients or experiments and `COMPUTATIONAL` where it is the review's own analysis of published structures - never as though the review had made the observation.

## One identification I want on the record as an inference

The review describes "In 2020, another individual" with p.Glu57Ala, white matter abnormality, microcephaly, seizures and movement disorders; PMID:32154054 reports "a further female child" with a cavitating leukodystrophy in the same year. The entry treats them as the same patient. Neither source names the other and the review's numbered reference list is not in the cached record, so this is inference from year, count and phenotype. If they are two people the series is five, not four - which changes no mechanism but changes every frequency.

## Deep research

`falcon` (Edison), run through `scripts/run_deep_research_client.sh` directly. NEC preflight passes: NDUFA2 mentioned 62 times against 6 for SLC22A5, which it correctly identifies as a dual diagnosis rather than a competing cause; its OMIM number matches MONDO's cross-reference. It counts three patients rather than four - it did not retrieve PMID:32154054 - which is the one place we disagree.

It had full-text access to the two papers the cache holds only as abstracts. That detail is recorded as **uncited leads in prose only**, never in a snippet or a structured slot: the founding patient's 20 percent muscle and 36 percent fibroblast complex I activity, the reduced NDUFA2 and NDUFA9 with accumulating subcomplexes, the 750 kDa subcomplex in the p.Lys45Thr fibroblasts, the imaging in all three, and the shared haplotype between the two p.Lys45Thr families. The catabolic-stress observation - varicella, vomiting and fever preceding decompensation - is deliberately *not* curated as an environmental entry for the same reason.

## Validation

All exit 0: `just validate`, `just validate-disorders`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`, `check-folded-hyphens`, `check-stubs`, `validate-stubs`, `validate-history`. 42/42 snippets verified.

`check-snippet-grading` earned its place during drafting: the sentence "The mutation in this accessory subunit causes reduced activity and disturbed assembly of complex I" was graded `HUMAN_CLINICAL` in two blocks and `IN_VITRO` in a third. Resolved to `HUMAN_CLINICAL` throughout, following how MC1DN24 grades the equivalent summary sentence - the deficiency was demonstrated in muscle as well as fibroblasts, and the in-vitro half is the rescue experiment, which is quoted separately on the readout.

Closes #10806


kevinschaper:
All five findings applied in the push above.

**Leukoencephalopathy frequency.** You were right that the band contradicted the prose next to it — the description says three of the four reported patients, which is 75%, and `VERY_FREQUENT` starts at 80%. Now `FREQUENT`.

**Seizure binding.** Split in two. The general `Seizures` entry (`HP:0001250`, `FREQUENT`) carries both patients; the narrower `Focal Seizures` entry (`HP:0007359`, `OCCASIONAL`) now carries only the 2018 patient whose epilepsy is actually characterised as focal. The 2020 patient's seizures are reported with no type, so reading them into the focal term would have been an inference the source does not make.

**Movement disorder.** Added as `HP:0100022`, `FREQUENT`. I checked the count rather than taking it on trust, because the band depends on it: the second 2018 patient "also had movement difficulties", and the 2020 patient had "movement disorders" — two of four, so 50%. (The review's other "loss of movement" sentence is a different subunit — Glu111 cannot be in the 99-residue NDUFA2 protein — so it is not counted here.) Bound to the general movement-abnormality term deliberately: neither source names a movement phenotype, and there is no dystonia, ataxia, chorea or tremor to bind to.

**Model type.** `PRIMARY_CELL_CULTURE`. The system is patient skin fibroblasts, which is exactly what that value is for; `OTHER` was wrong.

**The null allele's route through the graph.** This was the best of the five. The founding splice allele has no protein and no transcript, so it cannot destabilise a contact — there is nothing left to do the contacting — yet every path from the variant node ran through "Destabilisation of the NDUFA2-NDUFS1 Contact", whose own description says the alleles "weaken the contact rather than remove the protein". So the graph asserted something the node text denied. There is now a parallel edge from the variant node straight to "Failure of Complex I N-Module Assembly", carrying the null-allele evidence, and the original edge is labelled as the missense route.

Validation: schema, terms, 46/46 snippets, entity refs, duplicate keys, snippet grading, title snippets, snippet length — all pass.

caufieldjh:
@copilot please fix the merge conflicts in this pull request.

caufieldjh (approved):

PR #10827, opened 2026-09-04
Curate Stickler syndrome type 2 (COL11A1, MONDO:0011493)

caufieldjh:
Adds a new disorder entry for Stickler syndrome type 2, the COL11A1-related form of the syndrome. Stickler type 1 (COL2A1) was already curated; type 2 is a distinct genetic entity with its own gene, its own vitreous appearance, and considerably more hearing loss. Every claim in the entry is backed by a quote checked against a cached reference: 89 of 89 snippets verify.

The entry was built from a `claude_code` deep-research run (also committed, under `research/`) with the GeneReviews Stickler chapter (PMID:20301479) as the required phenotype baseline. The stub is deleted and a history record is included.

## What the disease is, and how the entry models it

Type XI collagen sits alongside the much more abundant type II collagen and controls how thick the resulting fibrils get. COL11A1 encodes one of its three chains. In type 2 Stickler syndrome a single altered copy still produces a chain that gets built into the collagen molecule — so the defect is dominant-negative, not a simple loss of half the protein. That distinction is the organising idea of the entry, and it is the reason the mouse model behaves the way it does (see below).

The pathophysiology is curated as a connected causal chain rather than a list:

```
COL11A1 triple-helical domain variant
  └─> impaired type XI regulation of collagen fibrillogenesis
        ├─> beaded vitreous anomaly ─> vitreoretinal traction and tear ─> retinal detachment
        ├─> disorganized cartilage matrix ─┬─> craniofacial malformation ─> cleft palate, midface retrusion
        │                                  └─> articular cartilage degeneration ─> premature osteoarthritis
        ├─> cochlear extracellular matrix abnormality ─> sensorineural hearing impairment
        └─> middle ear collagen abnormality ─> conductive hearing impairment
```

Also included: 12 phenotypes, 8 treatments, 2 animal models linked into the pathograph, one environmental exposure (trauma) linked into it, a diagnosis record, three differential diagnoses, and the ISDS type 11 collagen classification.

## The mouse model does not reproduce the hearing loss, and that is informative

The `cho/+` mouse is heterozygous for a Col11a1 null allele. It develops the joint disease on schedule, but its auditory brainstem response thresholds are indistinguishable from wild-type littermates at every age tested through 10 months (PMID:12527136), while human COL11A1 heterozygotes commonly have sensorineural hearing loss.

This is not a failed model — it is the argument that the human cochlear lesion needs the mutant chain physically present in the collagen, rather than just less collagen. The entry records it as a `FAILS_TO_RECAPITULATE` model link with an `UNCHANGED` readout, plus a `HUMAN_MODEL_MISMATCH` discussion. The `cho/cho` homozygote does have hearing loss and a disorganised organ of Corti (PMID:1952599), but that genotype is perinatally lethal and models fibrochondrogenesis rather than type 2 Stickler syndrome, so it is noted as the exception it looks like but is not.

## Two curation decisions reviewers should look at

**No prevalence band or rate is asserted.** Orphanet has no separate record for type 2, and its rows for the Stickler group as a whole disagree with each other by a factor of 500 — from 1-9 per 1,000,000 in France to 1-5 per 10,000 worldwide. Publishing any single number from those rows as a type 2 rate would put a quantitative claim into the machine-readable fields that the sources do not support. The record carries `prevalence_class: UNKNOWN` and states why the number was declined. The share of Stickler patients carrying a COL11A1 variant (10-20%, PMID:32558342) is recorded under `genetic.case_fractions`, which is the right slot for it.

The same reasoning applies to the pooled myopia (83%) and retinal detachment (45%) figures from PMID:32316871, which cover types 1 and 2 together and are not adopted as bands. The two frequency bands that are set — cataract `FREQUENT` and sensorineural hearing impairment `FREQUENT` — rest on type-2-scoped numbers (59% and 69% respectively).

**All retinal prophylaxis evidence is marked `directness: INDIRECT`.** Prophylactic cryotherapy and laser retinopexy are the standard of care, but the controlled cohorts are in type 1 patients. The Cambridge protocol study enrolled 487 type 1 patients, and the HTA systematic review says outright that only the type 1 population has been evaluated. Support for using these in type 2 comes by extension from the shared vitreoretinal mechanism, and the entry says so rather than presenting the evidence as direct.

## What was deliberately left out

Each of these is recorded in the entry's `notes:` with its reason, so a later reader can tell a decision from an oversight.

- **Datasets.** `just discover-datasets` returns only GENE_ONLY matches on COL11A1: GEO series about collagen XI in ovarian cancer and in cutaneous squamous cell carcinoma invasion. Those accessions resolve perfectly and are about a different disease. None is admitted.
- **Lattice retinal degeneration.** The research report puts paravascular lattice retinopathy at 38% and places it inside the vitreous-traction chain, so it belongs here on the merits — but no quotable source is reachable. The figure traces to Poulson et al. (PMID:15286167), which has no PubMed abstract, whose BMJ full text returns HTTP 403, and whose PMC copy no `just fetch-reference` prefix reaches. Every other cached source mentioning lattice is scoped to type 1 or to the general detachment population. `HP:0007992` is the right term if a citable source turns up; the report's proposed `HP:0000875` is *Episodic hypertension* and was rejected.
- **Repair of an established detachment** (scleral buckle, vitrectomy). Part of management, but no quotable sentence about it was found in any cached source and the GeneReviews abstract does not mention it. Prophylaxis, surveillance and education are curated instead.
- **The rare biallelic COL11A1 form** reported with unusually severe hearing loss. Different inheritance mode; it belongs in its own record.
- **An `exposure_term` for trauma.** ECTO was searched for trauma, injury, mechanical and physical exposure terms and for sport. It has `ECTO:6000002` exposure to exercise and `ECTO:6000031` exposure to strenuous exercise, neither of which means blunt ocular trauma. No term beats a wrong one.

## Handling of the deep-research report

The report generated no frontmatter validation, so both retro-fits were run. Results worth knowing:

| Check | Result |
|---|---|
| References resolved | 17/17, 0 unresolved, 0 off topic |
| Quoted claims found in source | **1 of 3** |
| Ontology terms resolved | 39/41 |
| Terms named as a **different** term | 4 |

Because two of its three quotes were paraphrases rather than quotes, no snippet in this entry was copied from the report — all were taken directly from the cached abstracts and full texts. All four mislabelled CURIEs were rejected: the report offered `UBERON:0000955` for "cochlea" (UBERON calls it *brain*), `UBERON:0002481` for "mandible" (*bone tissue*), `HP:0000875` for lattice degeneration (*Episodic hypertension*), and `HP:0000272` *Malar flattening* for midface hypoplasia. The entry uses `UBERON:0001844` and `HP:0011800` instead.

Separately, three PMIDs that surfaced during reference gathering turned out to be unrelated papers (about synapses, dystrophin, and HIV-1). They were discarded and their cache files are not committed.

## Review history on this branch

An adversarial review pass was run before opening the PR. It returned three blocking findings and eleven important ones, all addressed in the first commit: a group-scoped Orphanet prevalence band published in machine-readable numeric fields, a phenotype description that attributed a source's middle-ear-collagen claim to the wrong referent, and a treatment record whose only evidence described a different intervention.

Three follow-up commits address findings raised after the PR opened:

- `53bd820` — `PMID:23110709` was tagged `GeneReviews` in the `references:` block but is a systematic review in *Orphanet J Rare Dis*. `just tag-references` had matched the word "GeneReviews" inside that paper's bibliography, which the cache only contains because the reference validator upgraded the record to full text during this branch. Worth knowing generally: as more caches gain full text, any paper that cites a GeneReviews chapter becomes a candidate for the same false positive.
- `ebd1310` — three pieces of stale prose left behind by the pre-PR revision: a `notes:` paragraph describing prevalence handling the entry no longer does, a treatment cross-reference pointing at a paragraph that was never written, and a definition `description` claiming more than its snippet carried.
- `6c60bed` — the lattice omission recorded in `notes:`, `causal_link_type: INDIRECT_KNOWN_INTERMEDIATES` plus evidence on the conductive-hearing edge (the only previously uncited edge in the pathograph), and an uncited cache file dropped.

## Validation

All pass:

```
just validate-disorders kb/disorders/Stickler_Syndrome_Type_2.yaml   # the batched CI gate
just validate            # schema + terms + 89/89 snippets
just check-entity-refs   just check-causal-targets   just check-duplicate-keys
just check-enum-values   just check-qualifier-terms-online
just check-folded-hyphens   just check-snippet-length   just check-title-snippets
just check-snippet-grading  just check-environmental-evidence
just check-reference-titles
```

No baseline file was widened.

## Files

- `kb/disorders/Stickler_Syndrome_Type_2.yaml` — the entry
- `history/disorders/Stickler_Syndrome_Type_2/` — curation history record
- `research/Stickler_Syndrome_Type_2-deep-research-claude_code.md` (+ citations sidecar) — the deep-research output, with both validation sections
- `references_cache/` — 17 new cached references, plus full-text upgrades to two existing ones
- `cache/` — four rows added by term validation
- `stubs/Stickler_Syndrome_Type_2.yaml` — deleted

https://claude.ai/code/session_01PNRkJeTACVpLfM4XWjGor6


caufieldjh:
Both reviewer findings are fixed and pushed. Chasing the second one turned up two more of the same kind, so all three are in `ebd131060`.

**The GeneReviews tag (fixed in `53bd820`).** `PMID:23110709` is Acke et al., a systematic review in *Orphanet J Rare Dis*, not a GeneReviews chapter. The reviewer is right that it should not have carried that tag.

The cause is worth recording because it will recur. `just tag-references` decides GeneReviews membership by grepping the cached body for the word "GeneReviews", and that paper's bibliography cites a GeneReviews chapter. Its cache only contains that bibliography because `linkml-reference-validator` upgraded the record from abstract-only to full text during this branch. So the tagger read a citation in someone else's reference list and concluded the paper itself was a chapter. As more caches gain full text, more papers will be mis-tagged this way. I have not measured how many existing entries are already affected, so I am not claiming a number — but the detector matching on a bibliography rather than on the article's own citation line looks like a bug worth a separate issue.

**The stale prevalence prose (fixed in `ebd131060`).** Correct, and it was two sentences, not one. The closing `notes:` still said "only the group-scoped Orphanet birth-prevalence band is recorded" after that band had been dropped, and separately said the "10-20% of Stickler cases" figure could not be traced to a citable source, which stopped being true once it was cited to PMID:32558342 under `genetic.case_fractions`. Both rewritten to match what the entry does.

**Two more found while confirming it**, same root cause — an edit script aborted partway through the pre-PR revision and three of its patches were silently lost:

- The `Ophthalmic Surveillance` treatment says repair of an established detachment is "not curated as a separate treatment here — see the entry `notes`", but the paragraph it pointed at was never written. The cross-reference dangled. Restored.
- The Orphanet definition's `description` claimed the type-2 vitreous phenotype while its only snippet quoted the group definition. Orphanet's own type-2 clause is not quotable as a clean substring — the cached record renders the "fi" in "fibers" as a ligature character — so the description is narrowed to what the quote carries, and the type-2 half is now evidenced from PMID:20513134.

88/88 snippets verify. `just validate-disorders` plus entity-refs, causal-targets, duplicate-keys, enum-values, snippet-length, title-snippets, snippet-grading, reference-titles and environmental-evidence all pass. No baseline was widened.

One process note: each of these pushes cancels your in-flight run, since the workflow is `on: pull_request [synchronize]` with `cancel-in-progress: true`. That is why your checklist has restarted twice. Not deliberate — the alternative was sitting on validated fixes.

---

caufieldjh:
All four items addressed. Pushed as `6c60bed36`; 89/89 snippets verify.

**IMPORTANT 1 — stale `notes:`.** Already fixed in `ebd131060`, which landed before your run finished. Your review read `53bd820`, so it saw the old text. Both sentences you quoted are gone: the prevalence paragraph now says no band is asserted and why, and the 10-20% figure is described as the case fraction it is, sourced to PMID:32558342. The case-fraction-vs-prevalence distinction you flagged is now stated explicitly in that paragraph.

**IMPORTANT 2 — lattice retinal degeneration.** You're right on the substance and right about how it happened. I rejected `HP:0000875` because it is *Episodic hypertension*, and the concept went out with the bad code. `HP:0007992` is the correct term.

I could not add it, and the reason is the sourcing rather than the term. The 38% traces to Poulson et al., "Clinical features of type 2 Stickler syndrome" (PMID:15286167, DOI:10.1136/jmg.2004.018382). That record has no abstract in PubMed — the cache comes back `content_type: unavailable` — the BMJ full text returns HTTP 403, and the PMC copy (PMC1735871) is not reachable through any `just fetch-reference` prefix. So there is no cached text for a snippet to verify against.

Every other cached source that mentions lattice is scoped somewhere else:

| Source | Scope |
|---|---|
| PMID:21466760 | lattice in **type 1** Stickler cohorts |
| PMID:36583093 | lattice in the general rhegmatogenous-detachment population |
| PMID:41997526 | a lattice-plus-Stickler case report, but **COL2A1**, so type 1 |

Citing any of those would import a claim across exactly the scope boundary this entry declines to cross for prevalence and for prophylaxis. So I took your second option: the `notes:` now records the concept, the correct term, the source chain, and each search that failed, with a line to revisit if the Poulson text becomes fetchable. The gap is signalled rather than silent, which was your actual objection.

**Suggestion — uncited conductive edge.** Fixed, and your reading of the source was the useful part. It counts a conductive loss in 3% of ears and separately attributes middle-ear collagen to *tympanic hypermobility*; it never joins the two. The edge now carries `causal_link_type: INDIRECT_KNOWN_INTERMEDIATES`, its own evidence with `directness: INDIRECT`, and a description naming both the inferential step and the competing route (palatal defect with middle ear effusion), which is probably the commoner cause of conductive loss in this syndrome.

**Suggestion — orphan cache.** Dropped `references_cache/PMID_36278547.md`.

Two notes on your process, both useful to you rather than complaints. Your report-line citations were accurate and the `HP:0007992` catch was the substantive find of this review — I'd missed a real gap. And the near-miss you flagged yourself, where a line-wrapped cache made a present snippet look absent, is worth keeping: `grep` and the validator disagree on whitespace, so the validator is the authority there.

One thing outside my control: each of these fixes cancels your in-flight run, since the workflow is `on: pull_request [synchronize]` with `cancel-in-progress: true`. That is the third restart. Not deliberate.

---

PR #10849, opened 2026-09-04
Curate congenital leptin deficiency (MONDO:0013991, LEP)

### LINKED ISSUE #10832: Curate obesity due to congenital leptin deficiency (MONDO:0013991)
Curate a dismech entry for **obesity due to congenital leptin deficiency** ([MONDO:0013991](https://monarchinitiative.org/MONDO:0013991)) — biallelic LEP loss-of-function abolishes circulating leptin, so the hypothalamic arcuate nucleus never receives the adipostatic signal, producing severe early-onset hyperphagic obesity, hypogonadotropic hypogonadism, and T-cell dysfunction that recombinant leptin (metreleptin) reverses.

Leaning DISEASE. The stub records no MONDO descendants and a single causal gene (`hgnc:6553` LEP), and the mechanism is a single conserved chain (absent ligand → no LEPR/JAK2-STAT3 signalling in ARC → loss of POMC/CART drive and failure to suppress AgRP/NPY → hyperphagia + reduced energy expenditure), with a second arm to hypogonadotropic hypogonadism and immune dysfunction.

Disambiguation to respect: this is upstream of, and distinct from, `Obesity_Due_to_MC4R_Pathway_Disruption` (already curated) and from leptin *receptor* deficiency. The distinguishing feature is that leptin deficiency is the one monogenic obesity that is pharmacologically correctable.

Stub: `stubs/Obesity_Due_To_Congenital_Leptin_Deficiency.yaml`

gingin77:
Adds a Disease entry for congenital leptin deficiency and retires its stub.

## The lump/split call

I curated this as its own entry rather than folding it into `Obesity_Due_to_MC4R_Pathway_Disruption`, which already has a `LEP` subtype. That subtype is a one-line description with a defining citation — it records that the entity exists. It binds no MONDO term, so it does not retire MONDO:0013991 from the curation queue, and it carries no pathograph.

The umbrella entry is explicitly mechanism-based: its `disease_term` is the broad `MONDO:0011122` (obesity disorder) and its three pathophysiology nodes are the generic ones shared by every leptin-melanocortin lesion. The things that make leptin deficiency a distinct disease — the specific molecular lesion, the reproductive and immune arms, and a treatment that corrects all of them — are not properties of that umbrella and could not be added to it without making it something other than an umbrella.

I did not touch the umbrella entry, to keep this diff to one disease. Binding `MONDO:0013991` onto its `LEP` subtype would be a reasonable small follow-up.

## What's in the entry

Five pathophysiology nodes run absent-or-bioinactive leptin → failed hypothalamic leptin signaling, then branch three ways: unopposed orexigenic drive, starvation-pattern neuroendocrine suppression, and impaired T cell number and function.

Two points worth a reviewer's eye:

- **The molecular node covers two allele classes on purpose.** The classical lesion produces no hormone, so serum leptin is undetectable. A later-recognised missense class produces leptin that is secreted and immunoreactive but biologically inactive, so measured leptin is normal or high. Both abolish signalling; only the assay differs. I wrote the node around signal failure rather than hormone concentration, because a normal leptin level does not exclude the diagnosis.
- **Metreleptin carries four `target_mechanisms` links, one per arm.** That is the point of the disease: the receptor and downstream circuitry are intact, so replacing the ligand corrects appetite, the thyroid and gonadotropin axes, and the T cell defect. Treating it as a single "treats obesity" link would lose what makes this genotype different from every other monogenic obesity.

Deliberately kept separate from leptin *receptor* deficiency, which looks clinically similar but has extremely elevated circulating leptin and does not respond to replacement. That disambiguation is recorded in `notes` so it does not have to be rederived.

## Validation

- Evidence snippets: **28/28 verified** against the local reference cache

One process note: I did not trust any recalled PMID. An identifier I initially recalled for a key leptin paper turned out to name an unrelated pancreatic-islet study, so every reference here was resolved through a PubMed lookup and its cached title checked before use. `PMID:10486419` is cited only under `references:` — its cache entry has no retrievable abstract, so it carries no snippet.

Closes #10832

---

### DIFF
--- a/history/disorders/Congenital_Leptin_Deficiency/2026-09-04T052647Z-claude-code-56d398.yaml
+++ b/history/disorders/Congenital_Leptin_Deficiency/2026-09-04T052647Z-claude-code-56d398.yaml
@@ -1,0 +1,27 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Congenital_Leptin_Deficiency
+  path: kb/disorders/Congenital_Leptin_Deficiency.yaml
+session:
+  id: 2026-09-04T052647Z-claude-code-56d398
+  timestamp: '2026-09-04T05:26:47Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10849
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - has_subtypes
+  summary: Delayed Puberty frequency band; subtype dosing and infection rate
+  details: |
+    Acted on the round-three review of PR 10849. The blocking item was a self-inconsistency: Delayed Puberty was left at FREQUENT while the description directly beneath it stated that the systematic series counts delayed puberty inside the same 16 of 103 hypogonadism figure. I had corrected the sibling Hypogonadotropic Hypogonadism band in the previous round and missed its dependent, which is the failure mode of correcting one member of a pair. Now OCCASIONAL, carrying the same age caveat: unobservable before pubertal age since the cohort is mostly small children in whom puberty cannot yet be late, and effectively obligate after, since all 12 adult patients showed hypogonadism. Also took two non-blocking suggestions, both verbatim from the already-cached full text. The Classical subtype now records the standard metreleptin starting dose of 0.03 mg per kg lean body weight, which makes concrete the contrast with the antagonistic subtype that requires higher initial doses. The Bioinactive subtype now records that recurrent or severe infection is significantly more common there than in classical deficiency at 45 percent versus 16 percent, the one comorbidity reported to differ significantly between subtypes, with a note that the entry-level infection frequency therefore understates risk for those patients. Both snippets were extracted programmatically from the cache rather than retyped, because one contains a four-per-em space inside the dose figure and the other contains the source own signicantly typo, and an exact-substring match required reproducing both. Snippets went from 48 to 50, all verified; schema, term, grading, length, folded-hyphen, duplicate-key, enum, entity-ref and causal-target checks all pass.
--- a/kb/disorders/Congenital_Leptin_Deficiency.yaml
+++ b/kb/disorders/Congenital_Leptin_Deficiency.yaml
@@ -87,7 +87,9 @@ has_subtypes:
   description: >-
     No secreted hormone, so circulating leptin is low or undetectable. The
     largest class by far, and the one the disease was originally defined by.
-    Responds to a standard metreleptin starting dose.
+    Leptin is low or undetectable on both the regular and bioLEP assays, and
+    treatment starts at the standard metreleptin dose of 0.03 mg/kg lean body
+    weight -- the reference point the other two subtypes are measured against.
   evidence:
   - reference: PMID:38470203
     reference_title: "Classification of Congenital Leptin Deficiency."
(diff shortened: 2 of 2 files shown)


gingin77:
You were right that the review crossed with the push — I hadn't seen these findings. All nine are now addressed in `0fd1020b4b`.

## The enabling step: the cache file was abstract-only

Your line references (170, 174, 192, 214, 220, 240) didn't exist in my copy — `references_cache/PMID_38470203.md` was **85 lines**, abstract only. Re-fetching returned the PMC full text at **255 lines**, and that is what made the denominators, the assay algorithm and the dosing sentence quotable at all.

Worth flagging as a general hazard: the entry cited a paper whose committed cache held a fraction of it, and nothing in the validation stack notices, because every snippet still verifies against the smaller file. A reference can be "in cache" and still be substantially unavailable. The regenerated full text is committed here.

## Truncated snippets — your diagnosis was exactly right

All four fixed. The autosomal-recessive one is the case that matters: it stopped immediately before **consanguineous**, so the sole evidence for a recessive claim asserted nothing about inheritance. That's the failure mode worth naming — the snippet was an exact substring, so `count-verified-snippets` passed at 31/31 while four quotes didn't support their own explanations. Exactness and sufficiency are different properties and only one of them is checked.

## Frequencies — corrected, and the hypogonadism case argued rather than downgraded

Recurrent infections → OCCASIONAL (19%), hyperinsulinemia → OCCASIONAL (29%), hypogonadotropic hypogonadism → OCCASIONAL (16%).

On hypogonadism I took your suggestion to make the case explicitly. It is 16/103 overall but **100% of the 12 adult patients**, because it cannot be observed in a cohort that is mostly small children. The description now says it is effectively obligate after pubertal age and unobservable before, with the quote carrying both denominators. `Delayed Puberty` records that it sits *inside* that same 16% rather than being an independent cohort — thanks for catching that, it would have double-counted.

## Subtypes, and why they were load-bearing

Added all three (`Classical` 21 variants/128 patients, `Bioinactive` 3/12, `Antagonistic` 3/7). You were right that this got more pressing, not less: the treatment description asserted a uniform response while the source says antagonistic-variant patients **antagonise metreleptin itself** and need higher initial doses with tailored tapering. The treatment block now says so.

The `diagnosis` block now distinguishes the two ways leptin measurement fails — the bioLEP assay resolves the bioinactive class, but for the antagonistic class both assays read inconspicuous, so leptin measurement doesn't identify the disease at all and only sequencing will. As written before, it implied the assay was universally discriminating.

## The rest

- **Prevalence** now carries the 2024 census (148 patients / 28 variants). I kept the 2012 "fewer than 20" figure beside it rather than replacing it, because the growth between them is mostly ascertainment of the classes that present with normal leptin — which is the entry's own diagnostic point.
- **Missing phenotypes** added: dyslipidemia 21%, hypertension 11%, developmental delay 21%. Developmental delay carries the source's own caveat (may reflect consanguinity or immobility) and is deliberately **not** linked into the pathograph for that reason.
- **`CL:4042033`** pro-opiomelanocortin neuron, with generic `neuron` kept for the AgRP/NPY half as you suggested.
- **`therapeutic_modality`** → `PROTEIN_REPLACEMENT`.
- **"Both classes abolish leptin signalling"** softened — the source says "no or only little downstream signaling", with p.Gly59Ser retaining partial agonist activity.
- **ob/ob mouse** added, with `limitations` noting it is a null and therefore models the classical subtype only.

**Deep-research artifact:** still none, and the scope was deliberate. Your inference held here more than anywhere — this entry was the most under-consumed of the five, and the fix was almost entirely transcription from a paper it already cited.

## Two defects I introduced and caught before pushing

Worth recording rather than hiding: a text insert concatenated onto the previous line (`band.- category:`), producing five duplicate YAML keys and a dangling sequela target — the exact `check-duplicate-keys` failure mode CLAUDE.md warns about. And when I lengthened the four-word hypertension snippet, the longer version failed matching because the validator strips bracketed spans; resolved by quoting a bracket-free portion.

**Validation:** 48/48 snippets verified (up from 31) · `linkml-validate` clean · term validation passed · grading, length, title, folded-hyphen, duplicate-key, enum, entity-ref and causal-target checks all pass.

---

gingin77:
Fixed in `9a6c4309cc`. You're right, and the way it went wrong is worth naming.

## The band

`Delayed Puberty` → `OCCASIONAL`, with the same age caveat as its sibling.

This was a self-inconsistency, not a judgement call: the description I wrote *directly beneath it* says the series counts delayed puberty inside the 16/103 hypogonadism figure, while the machine-readable slot above said 30–79%. I corrected `Hypogonadotropic Hypogonadism` last round and missed its dependent — the failure mode of fixing one member of a pair and not asking what else was drawn from the same count. The prose and the slot now agree, and the description spells out why the band understates it (unobservable before pubertal age, effectively obligate after).

## Two suggestions taken

Both verbatim from the full text that's now cached:

- **Classical** records the standard metreleptin starting dose (0.03 mg/kg lean body weight). You were right that this makes the subtype contrast concrete — the antagonistic subtype's "higher initial doses" had nothing to be higher *than*.
- **Bioinactive** records the 45% vs 16% infection rate. This one is more than a detail: it's the only comorbidity reported to differ significantly between subtypes, so the entry-level `OCCASIONAL` infection frequency understates risk for that subtype specifically. The subtype description now says so.

Both snippets were extracted programmatically rather than retyped — one contains a four-per-em space inside the dose figure, the other the source's own "signicantly" typo, and exact-substring matching required reproducing both. The `explanation` on the second notes the typo is deliberate so a later reader doesn't "correct" it and break the match.

## Not taken

- **`biochemical:` block** — thanks for dropping it; agreed the `diagnosis` block now carries the assay logic with citations, which was the substance.
- **`parents: Inherited Obesity`** — leaving as-is. It's free text and descriptive, and inventing a `Grouping` to make it resolve would be a bigger change than the problem.
- **Binding MONDO:0013991 onto the umbrella's LEP subtype** — agreed, still the right follow-up, still a separate PR.

## On the deep-research artifact

Worth acknowledging what you did there rather than just taking the pass. You raised it twice, then tested the premise behind the rule instead of applying it — walked the completeness checklist against the primary source directly, found the gaps it stood proxy for were closed, and declined to block on the proxy. That's the right way to treat a rule whose purpose is inspectable, and the reasoning being on the record is more useful than the pass itself.

**Validation:** 50/50 snippets verified · `linkml-validate` clean · term validation passed · grading, length, title, folded-hyphen, duplicate-key, enum, entity-ref and causal-target checks all pass.

---

PR #10850, opened 2026-09-04
Curate adult-type hypolactasia (MONDO:0006065, MCM6/LCT)

### LINKED ISSUE #10833: Curate lactose intolerance adult type (MONDO:0006065)
Curate a dismech entry for **lactose intolerance adult type** ([MONDO:0006065](https://monarchinitiative.org/MONDO:0006065)) — adult-type hypolactasia, the ancestral human state, in which regulatory variants in an MCM6 intronic enhancer fail to sustain LCT transcription after weaning, so undigested lactose reaches the colon as an osmotic and fermentative load.

Leaning DISEASE. The stub records no MONDO descendants and one gene (`hgnc:6949` MCM6), and the chain is conserved: enhancer variant → declining LCT expression post-weaning → intestinal lactase deficiency → osmotic load plus colonic bacterial fermentation → bloating, flatulence, osmotic diarrhoea.

**Named Entity Confusion guardrail.** `Congenital_Lactase_Deficiency` is already curated and its description explicitly contrasts itself with this entity: CLD is LCT coding-region loss-of-function presenting neonatally, whereas this is a distal regulatory variant that lowers rather than abolishes lactase, only after weaning. Evidence for one must not be reused for the other. There is also an existing `diet_induced_osmotic_diarrhea` module that this entry should probably declare `conforms_to` against rather than re-deriving.

One thing to settle during curation: whether a highly prevalent ancestral trait belongs in `kb/disorders/` at all, or is better recorded as OUT_OF_SCOPE. The symptomatic intolerance is a genuine clinical entity, so I expect DISEASE, but it is worth arguing explicitly in the entry rather than assuming.

Stub: `stubs/Lactose_Intolerance_Adult_Type.yaml`

gingin77:
Adds a Disease entry for adult-type hypolactasia (lactase non-persistence) and retires its stub.

## Three calls a reviewer should check

**The gene is MCM6, not LCT.** This looks wrong at a glance and isn't. Sequencing of the LCT coding and promoter regions found no variation tracking with the phenotype; the causal variants are cis-regulatory and sit in an intron of the neighbouring MCM6 gene, about 14 kb upstream. I annotated `genetic` and the molecular node with MCM6 as the gene that harbours the element, and recorded in `notes` that the gene whose expression is controlled is LCT. LCT itself is listed as a separate `genetic` entry with no `relationship_type`, carrying the `REFUTE` item that excludes it as the mutated gene.

**It is deliberately not `Congenital_Lactase_Deficiency`.** That entry already exists, and its own description explicitly contrasts itself with this one: CLD is a rare LCT coding-region loss-of-function abolishing enzyme activity from birth and presenting neonatally, whereas this is a common regulatory variant that lowers but does not abolish lactase, only after weaning. Much of the literature says only "lactase deficiency", so this is a live Named Entity Confusion risk; the guardrail is recorded in `notes` and no evidence is shared between the two entries.

**The inheritance reads backwards and is correct.** Hypolactasia, the ancestral state, is recessive; lactase *persistence* is the derived, dominant trait. The two cited sources describe the same locus from opposite directions, which is why the `inheritance` block carries one snippet for each. The same inversion governs `relationship_type` in the `genetic` block: the non-persistence allele is `CAUSATIVE`, the African and Middle Eastern persistence alleles are `PROTECTIVE`.

## Conformance, and one node that deliberately does not conform

Three nodes declare `conforms_to` against `diet_induced_osmotic_diarrhea` — luminal solute retention, osmotic water influx, and colonic fermentation — substituting lactose for the generic malabsorbed substrate.

The enzyme-activity node deliberately carries **no** anchor. That module scopes itself to the congenital diarrheas: its trigger node describes a *biallelic loss-of-function* lesion, annotates `GO:0004553` with `modifier: LOSS_OF_FUNCTION` where this entry has `DECREASED`, and its own trigger-node evidence explanation names adult-type hypolactasia as not a conformer. That is a difference in kind, not degree, so the trigger-level claim is withheld and the reasoning is recorded on the node. The downstream pathway, which the module itself calls disorder-agnostic, genuinely is shared — hence conformance there.

## Prevalence, and a retracted source

The figure most often quoted for this condition — global lactose malabsorption of 68% — comes from a meta-analysis PubMed flags as a **Retracted Publication** (PMID:28690131). It is not cited here. Prevalence rests instead on an unretracted 57-65% worldwide range plus a deliberately rate-free qualitative record describing where lactase persistence is rare or absent; inverting a prose statement about the reciprocal trait into a number would manufacture precision the source does not carry. The retraction is recorded in `notes` so the figure does not get re-added, and the retracted paper's cache file is retained uncited so that claim stays checkable in-repo.

Note also that the 57-65% range is labelled "lactose intolerance" by its source when at that magnitude it is the malabsorption trait. Each prevalence record says which of the two it means.

## Scope note

The entry is careful that malabsorption is necessary but not sufficient for symptoms, and that the enzyme deficit is partial, so tolerance is dose-dependent and complete dairy exclusion is usually unnecessary. A `KNOWLEDGE_GAP` discussion states this explicitly: in a randomised double-blind trial, adults who described themselves as severely intolerant had negligible symptoms on 240 mL of milk daily. The causal edges into the phenotype nodes should be read as dose-conditional rather than obligate.

Worth flagging: at roughly two-thirds of the world's adults, this is the global norm rather than a rare disease, so there is a fair argument that it is a trait rather than a disorder. I curated it as a Disease because the symptomatic syndrome is a genuine clinical entity with a diagnostic test and specific management, but I'd rather that be argued explicitly than assumed.

## Deep research

No deep-research run was performed for this disease. The entry was deepened directly from primary literature instead — 9 references and 55 evidence items across `biochemical`, `environmental`, `differential_diagnoses`, `discussions`, four diagnostic methods and three `genetic` entries. Recording this as a documented deviation rather than an omission.

## Validation

- Evidence snippets: **46/46 verified** against the local reference cache, including under `--no-full-text`
- `check_duplicate_yaml_keys`, `check_enum_values`, `check_entity_refs`, `check_causal_targets`, `check_qualifier_terms`, `check_snippet_grading`, `check_snippet_length`, `check_title_snippets`, `check_folded_hyphens`, `check_environmental_evidence`: all OK

Closes #10833

---

### DIFF
--- a/history/disorders/Adult-Type_Hypolactasia/2026-09-04T062740Z-claude-code-0aa899.yaml
+++ b/history/disorders/Adult-Type_Hypolactasia/2026-09-04T062740Z-claude-code-0aa899.yaml
@@ -1,0 +1,45 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Adult-Type_Hypolactasia
+  path: kb/disorders/Adult-Type_Hypolactasia.yaml
+session:
+  id: 2026-09-04T062740Z-claude-code-0aa899
+  timestamp: '2026-09-04T06:27:40Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10833
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10850
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - genetic
+  - diagnosis
+  - treatments
+  summary: 'Review round 2: narrow module conformance, correct two relationship_type
+    values'
+  details: |
+    Response to the round-2 PR review on #10850. All six round-1 items were accepted; these are the three new important items plus two of the three suggestions.
+
+    Module conformance narrowed. The reviewer found - and corrected their own round-1 advice - that kb/modules/diet_induced_osmotic_diarrhea.yaml scopes itself to the congenital diarrheas and its trigger-node evidence explanation names adult-type hypolactasia as explicitly not a conformer. Verified directly: the trigger node's description requires a biallelic loss-of-function lesion and annotates GO:0004553 with modifier LOSS_OF_FUNCTION, where this entry's corresponding node is a cis-regulatory downregulation annotated DECREASED. That is a difference in kind. The trigger-node conforms_to is therefore dropped and the reasoning recorded on the node itself; conformance is retained on the three downstream nodes, which the module describes as disorder-agnostic and which this disease genuinely shares. Chose to narrow rather than amend the module's exclusion sentence: the exclusion is explicit, is backed by PMID:16400612 which makes exactly this regulatory-versus-coding distinction, and reversing another curator's sourced scope decision unilaterally in a disease PR is the wrong venue.
+
+    relationship_type on the African and Middle Eastern lactase-persistence alleles changed CAUSATIVE -> PROTECTIVE. Those derived variants confer lactase persistence, which is the absence of this disease; the non-persistence allele in the preceding genetic entry is the causative one. This was the entry's own documented inheritance inversion reappearing one level down.
+
+    relationship_type dropped entirely from the LCT genetic entry. CAUSATIVE contradicted that entry's own REFUTE evidence and its notes, both of which state that LCT carries no phenotype-tracking coding or promoter variation. No enum value means 'the regulated effector gene, harbouring no causal variation', and the slot is optional.
+
+    The Tishkoff haplotype-background item on the genetic-testing diagnosis regraded REFUTE -> SUPPORT with directness INDIRECT. It bears on the adequacy of one single-variant implementation, not on genotyping as a method; the population-scope limitation already lives in that entry's description.
+
+    Suggestion 4 (therapeutic_agent on the lactase treatment) attempted and reverted, with the reason recorded in the treatment notes. NCIT:C29147 Lactase resolves but is not a member of the ChemicalEntityTerm dynamic enum, which is reachable only from CHEBI:24431 and NCIT:C1909 Pharmacologic Substance; NCIT:C29147 sits under Dietary Supplement and NCIT:C107438 Beta-Galactosidase under Protein, and CHEBI has no term for the enzyme. This is an enum-membership gap, not a missing term - the distinction CLAUDE.md draws between the label cache and the enum cache.
+
+    Suggestion 5 (readout direction) not taken; kept NEGATIVE, since the node's quantity is lactase activity (GO:0000016, DECREASED) and NEGATIVE describes that quantity rather than the breath-hydrogen measurement.
+
+    Validation at this head: linkml-validate clean; snippets 46/46 verified; term validation passed; all offline gates OK; cache diff empty for this commit.
--- a/kb/disorders/Adult-Type_Hypolactasia.yaml
+++ b/kb/disorders/Adult-Type_Hypolactasia.yaml
@@ -175,7 +175,6 @@ pathophysiology:
     causal_link_type: DIRECT
 - name: Declining Intestinal Lactase Activity
   biological_scale: MOLECULAR
-  conforms_to: "diet_induced_osmotic_diarrhea#Loss of a Substrate-Specific Brush-Border Digestive or Absorptive Step"
   description: >-
     Lactase-phlorizin hydrolase at the enterocyte brush border, which hydrolyses
     lactose into glucose and galactose, falls in activity through childhood in
@@ -183,6 +182,15 @@ pathophysiology:
     decline rather than a disease process in the usual sense, and it is the
     default outcome for most of humanity. Reduced, not absent, activity is the
     key quantitative feature separating this from congenital lactase deficiency.
+
+    Deliberately carries no conforms_to. The diet_induced_osmotic_diarrhea
+    trigger node describes a biallelic loss-of-function lesion and is annotated
+    modifier: LOSS_OF_FUNCTION, where this node is a cis-regulatory
+    downregulation annotated DECREASED. That is a difference in kind, not in
+    degree, and the module's own trigger-node evidence names adult-type
+    hypolactasia as not a conformer of it. Conformance is claimed only on the
+    three downstream nodes, which the module describes as disorder-agnostic and
+    which this disease genuinely shares.
   cell_types:
   - preferred_term: enterocyte
     term:
@@ -477,7 +485,7 @@ genetic:
     term:
       id: hgnc:6949
       label: MCM6
-  relationship_type: CAUSATIVE
+  relationship_type: PROTECTIVE
   notes: >-
     Lactase persistence arose independently more than once. Three further
     regulatory variants in the same MCM6 enhancer region — G/C-14010, T/G-13915
@@ -487,6 +495,11 @@ genetic:
     C/T-13910 misclassifies non-European individuals as non-persistent: it
     genotypes the wrong site. The clinical consequence belongs to the diagnostic
     section, but the cause is here.
+
+    relationship_type is PROTECTIVE, not CAUSATIVE: these derived alleles confer
+    lactase persistence, which is the absence of this disease. The
+    non-persistence allele in the entry above is the causative one. This is the
+    same inversion the entry-level notes warn about, one level down.
   evidence:
   - reference: PMID:17159977
     reference_title: "Convergent adaptation of human lactase persistence in Africa and Europe."
@@ -518,7 +531,6 @@ genetic:
     term:
       id: hgnc:6530
       label: LCT
-  relationship_type: CAUSATIVE
   notes: >-
     Listed explicitly because the entry's most surprising claim is that the
     causal variants are not in this gene. LCT encodes lactase-phlorizin
@@ -527,6 +539,11 @@ genetic:
     Coding-region loss of function in LCT causes a different disease
     (Congenital_Lactase_Deficiency), which is the Named Entity Confusion this
     entry guards against.
+
+    relationship_type is deliberately absent. CAUSATIVE would contradict this
+    entry's own REFUTE evidence below, and no available value means "the gene
+    whose expression is regulated, harbouring no causal variation of its own".
+    The slot is optional, so omission is the honest encoding.
   evidence:
   - reference: PMID:11788828
     reference_title: "Identification of a variant associated with adult-type hypolactasia."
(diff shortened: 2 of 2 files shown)


gingin77:
Thanks — the retracted-source catch was the important one, and it was worse than a bad citation: it was holding up the entry's central framing argument. Pushed as `28ca506cf3`.

## The retracted source (item 1)

Removed entirely. `PMID:28690131` no longer appears in `references:` or in any evidence item. Prevalence is rebuilt on two records:

- **Worldwide, 57-65%** from `PMID:33887513`, with `rate_low`/`rate_high` carrying the range.
- **A qualitative geographic record** from `PMID:20144208` (Itan 2010) — "Lactase persistence is common in people of European ancestry as well as some African, Middle Eastern and Southern Asian groups, but is rare or absent elsewhere in the world."

I gave the second record **no rate at all**, deliberately. It states the distribution of the *reciprocal* trait, and turning a prose statement about persistence into a number for non-persistence would manufacture precision the source does not carry. It is `directness: INDIRECT` for the same reason.

You were right that the 57-65% figure needs re-scoping rather than a straight swap, and I went further than relabelling it. The source calls it "lactose intolerance", but a figure of that magnitude is the malabsorption trait — this literature conflates the two constantly, and that conflation is the thing this entry exists to keep straight. So the caveat is recorded in three places: the entry `notes`, the record's own `notes`, and the record says which of the two it means. The `notes` also state plainly that the widely quoted 68% figure traces to a retracted analysis, so nobody re-adds it.

I kept `references_cache/PMID_28690131.md` even though nothing now cites it, so the retraction claim in `notes` stays checkable in-repo rather than resting on my say-so. Happy to drop the file if you'd rather the PR add no uncited cache entries.

On `prevalence_class`: I left it `COMMON` rather than switching to a numeric band, and recorded why. The numeric bands stop at `ABOVE_1_IN_1000`, which is true of a 61% trait but conveys nothing — it is the band designed for the common end of *rare* disease. The rate slots carry the number.

## The missing deep-research artifact (item 2)

I did not run one, and I want to be straight about that rather than claim otherwise. What I did instead was work the completeness gap directly: seven new references and five new sections.

| Section | Before | Now |
|---|---|---|
| `biochemical` | absent | breath H2 readouts against the genotype reference standard |
| `environmental` | absent | dietary lactose as the obligate, dose-graded exposure |
| `differential_diagnoses` | absent | IBS |
| `discussions` | absent | `KNOWLEDGE_GAP` on malabsorption vs. symptomatic intolerance |
| `diagnosis` | 1 method | 4 methods |
| `genetic` | 1 entry | 3 entries |

Snippets went 22 → **46**. Your point that the entry cited nothing on the African persistence alleles it discussed in `notes`, and nothing on the dose-threshold studies its own argument rested on, was fair and both gaps are now closed with primary literature (Tishkoff 2007 for the former, Suarez 1995 NEJM for the latter).

`animal_models` and `datasets` are still absent. I'd rather leave them empty than pad them: this is a common human regulatory-variant trait, and the obvious mouse work is about lactase developmental regulation generally rather than modelling this condition.

## Conformance anchor (item 3)

You were right, and the diagnosis was exact. Moved to `Declining Intestinal Lactase Activity`, which carries `GO:0000016 lactase activity` and is the genuine analogue of `Congenital_Lactase_Deficiency`'s anchored node. I also changed its `biological_scale` from `TISSUE` to `MOLECULAR` to match the module node (your suggestion 8). The transcription node now carries no anchor, which is correct — it is genuinely upstream of anything the module models.

## Splitting the bundled node (item 4)

Split into `Osmotic Water Influx and Luminal Distension` and `Colonic Microbial Fermentation of Lactose`, each conforming to its own module node. Module coverage 2/5 → 4/5.

The rewiring is the part worth checking: diarrhoea and distension now hang off the osmotic arm, and flatulence, pain and distension off the fermentative arm. Distension is on both because both mechanisms distend the bowel. That is exactly the distinction the bundle erased — the pathograph can now express that removing the fermentation substrate relieves gas specifically.

## Evidence source (item 5)

Fixed to `OTHER`, and the `explanation` now says why: it is the paper's background assertion about a heritable trait, not a finding of its Caco-2 reporter and EMSA experiments. Your note that `check-snippet-grading` structurally *cannot* see this — it keys on the quoted sentence, and this sentence appears once — is worth recording; that is a real blind spot in the gate, not an oversight in its baseline.

## Diagnostics (item 6)

All four cited methods are now present. Three carry a `diagnosis_term`: `NCIT:C116515` Breath Test, `NCIT:C15709` Genetic Testing, `NCIT:C15189` Biopsy Procedure.

**The lactose tolerance test is deliberately left unbound.** NCIT has no C25218-reachable term for it, and every near match is wrong in kind — it is a blood glucose measurement after an oral load, so binding it to `Breath Test` would assert the wrong specimen and the wrong analyte. I tried carrying the analyte as a `qualifiers` pair and rejected that too: it dresses up a wrong binding rather than fixing it. Free-text `preferred_term` per the ontology term contract, with the reasoning in `notes`. No term beats a bad one.

The genetic-test entry is where the misclassification caveat now lives, as you suggested, and it carries a `REFUTE` item against a C/T-13910-only test on the strength of Tishkoff's independent-origin finding.

## Suggestions

Taken: 7 (environmental block, `TRIGGERS` → the luminal-retention node), 8 (scales reconciled), 10 (LCT added to `genetic`, with a `REFUTE` item recording that it carries no phenotype-tracking variation — the surprising claim now queryable rather than prose), 11 (rationale recorded).

**Not taken: 9.** "Various extra-intestinal diseases, including cancers" is too vague to seed a `comorbidities` entry — it names no disease, so there is nothing to bind. You raised that possibility yourself and I think it settles it.

One thing I added beyond the review: a `KNOWLEDGE_GAP` discussion on why only a minority of malabsorbers are symptomatic, attached to the fermentation node and the pain phenotype. The Suarez NEJM trial found that adults who reported severe intolerance below 240 mL of milk had negligible symptoms at that dose and no significant difference from lactose-hydrolysed milk. That is the entry's weakest causal step, and it now says so rather than letting a clean-looking pathograph imply determinism.

## Validation at this head

`linkml-validate` clean · snippets **46/46** verified, and still 46/46 under `--no-full-text` · term validation passed · duplicate-key, enum, entity-ref, causal-target, qualifier-term, snippet-grading, snippet-length, title-snippet, folded-hyphen, environmental-evidence and source-defect gates all OK.

Cache diff is two rows — `ECTO:9000240` and `GO:0005990` — plus their enum memberships. I dropped one cache file I fetched but did not end up citing, so the PR adds no unused references beyond the retracted one discussed above.

Two snippets from `PMID:32600320` contain U+2009 thin spaces inside the ppm figures and a trademark sign in the product name, so I extracted them programmatically from the cache rather than retyping them.

---

gingin77:
All three items in one push, `89da7a7ca7`, plus two of the three suggestions.

## The module exclusion (item 1)

Thanks for going back and correcting your own round-1 advice — that's the most useful thing in this review, and I checked it rather than taking it on trust. It holds, and it's stronger than a scope sentence: the module's trigger node annotates `GO:0004553` with `modifier: LOSS_OF_FUNCTION` where this entry's node correctly carries `DECREASED`, and its description requires a biallelic loss-of-function lesion. That's a difference in kind, not degree. My node was already right; the anchor was the thing that was wrong.

**I narrowed rather than argued**, for two reasons beyond the exclusion sentence itself. First, the exclusion isn't an offhand remark — it's backed by `PMID:16400612`, which exists precisely to draw the regulatory-vs-coding distinction, so amending it means arguing against a curator's sourced reasoning. Second, a disease PR is the wrong venue: a reader watching `kb/modules/` for changes has no reason to be watching this branch. If someone wants to make the case that the module should admit regulatory-lesion conformers at the trigger level, that's a module PR where module reviewers will see it.

So the trigger anchor is gone and the reasoning sits on the node, naming the specific `LOSS_OF_FUNCTION`-vs-`DECREASED` mismatch so the next curator doesn't rediscover it. Conformance stays on the three downstream nodes.

## The two `relationship_type` values (items 2 and 3)

Both were straightforwardly wrong.

`PROTECTIVE` on the persistence alleles — you're right that this is my own documented inheritance inversion reappearing one level down, which is a slightly embarrassing place for it to turn up given the entry has a `notes` paragraph warning about exactly that. I've added a line to that `genetic` entry saying why it's `PROTECTIVE`, so the value doesn't look like a typo to someone skimming a block where the neighbouring entry is `CAUSATIVE`.

`relationship_type` dropped from LCT. Agreed that omission beats any available value: the entry asserted and refuted the same proposition on one object, and no enum member means "regulated effector gene harbouring no causal variation of its own". Recorded that in `notes` too, since an absent optional slot otherwise reads as an oversight.

## Suggestion 6, taken

Regraded to `SUPPORT` + `directness: INDIRECT`. Your framing is the right one — the snippet refutes a C/T-13910-only panel, not genotyping as a method, and the `diagnosis` entry is about the method. The `explanation` now says both what it supports and why it isn't `REFUTE`.

## Suggestion 4, tried and reverted — with a finding

This one doesn't work, and the reason is worth recording. `NCIT:C29147` resolves, but it is **not a member of the `ChemicalEntityTerm` enum**, which reaches only from `CHEBI:24431` and `NCIT:C1909` (Pharmacologic Substance):

```
runoak -i sqlite:obo:ncit ancestors -p i NCIT:C29147
  NCIT:C1505   Dietary Supplement
  NCIT:C1908   Drug, Food, Chemical or Biomedical Material
```

`NCIT:C107438` Beta-Galactosidase is no better — it sits under Protein / Gene Product — and CHEBI has no term for the enzyme at all. Term validation rejects `C29147` outright.

That's the label-cache-versus-enum-cache distinction CLAUDE.md draws, and it caught me as well as you: I added the binding, and only the validator's `not in dynamic enum 'ChemicalEntityTerm'` told me otherwise. Arguably NCIT has it right — lactase is sold as a supplement, not a pharmacologic substance — which would make the real gap the `treatment_term`, not the agent. I've recorded the full reasoning and both rejected CURIEs in the treatment `notes` so nobody re-adds it, and discarded the cache row the attempt generated. This commit therefore has **no cache diff at all**.

## Suggestion 5, not taken

Kept `direction: NEGATIVE`. You called it a coin-flip and offered to leave it, and I think the tiebreak is that `direction` describes the target node's quantity rather than the measurement: that quantity is `lactase activity` (`GO:0000016`, `DECREASED`), so `NEGATIVE` is consistent. The node's *name* being the deficit is the confusing part, and the `interpretation` prose carries that.

## Validation at this head

`linkml-validate` clean · **46/46** snippets verified · term validation passed · duplicate-key, enum, entity-ref, causal-target, qualifier-term, snippet-grading, snippet-length, title-snippet, folded-hyphen and environmental-evidence gates all OK.

Diff is two files — the entry and its history record. No cache change, no derived artifacts.

I also updated the PR body, which still said "two nodes declare `conforms_to`" and made no mention of the retracted source or the deliberate non-conformance. It now describes what the branch actually contains.

---

PR #10851, opened 2026-09-04
Curate polymyalgia rheumatica (MONDO:0019735)

### LINKED ISSUE #10834: Curate polymyalgia rheumatica (MONDO:0019735)
Curate a dismech entry for **polymyalgia rheumatica** ([MONDO:0019735](https://monarchinitiative.org/MONDO:0019735)) — an IL-6-driven inflammatory rheumatic disease of older adults causing proximal shoulder and hip girdle pain and morning stiffness, with bursitis and peri-articular synovitis rather than true myositis, and a characteristic rapid response to low-dose glucocorticoids.

Leaning DISEASE. The stub records no MONDO descendants, and the pathophysiology is one chain: age-associated immune dysregulation → IL-6 and IL-1β from monocytes/macrophages → subacromial/subdeltoid bursitis and hip peri-synovitis → girdle pain, stiffness, and a systemic acute-phase response.

**Relationship to Giant Cell Arteritis, which is already curated.** These are not the same entry and this one should not be folded into GCA. They overlap substantially — roughly half of GCA patients have PMR symptoms, a minority of PMR patients have subclinical arteritis, both are IL-6 driven and both respond to tocilizumab — but PMR occurs far more often on its own, lacks the large-vessel vasculitis that defines GCA, and carries no vision-loss risk. The existing `Giant_Cell_Arteritis` entry already mentions polymyalgia rheumatica as an associated feature in three places; those mentions should stay, and this entry should relate to GCA rather than duplicate it. A comorbidity record may be the right way to express the overlap.

Stub: `stubs/Polymyalgia_Rheumatica.yaml`

gingin77:
Adds a Disease entry for polymyalgia rheumatica and retires its stub.

## Why this isn't part of the giant cell arteritis entry

`Giant_Cell_Arteritis` is already curated and mentions polymyalgia rheumatica in three places. Those mentions are correct and I left them alone.

The two diseases are genuinely interlinked — some authors treat them as one GCA-PMR spectrum, they share immunopathophysiology, and both respond to IL-6 receptor blockade. But PMR occurs far more often on its own, its lesion is peri-articular bursitis and synovitis rather than large-vessel vasculitis, and it carries none of the vision-loss risk that makes GCA an emergency. Folding it in would either lose the bursal mechanism or overstate the vasculitis risk of isolated PMR.

The overlap deserves to be expressed quantitatively, and I think a comorbidity record is the right home for it. I did not attempt that here — it belongs in its own change rather than bundled into a new disease entry.

## The name is a misnomer, and the entry says so

"Polymyalgia" implies muscle disease. Muscle histology and enzymes are normal; ultrasound, MRI and PET consistently show bilateral subacromial bursitis, biceps long-head tenosynovitis and trochanteric bursitis. The anatomy is therefore bound to UBERON synovial bursa and shoulder joint on imaging and bursal-biopsy evidence, and `notes` records the trap explicitly so a future curator doesn't bind muscle-pathology terms on the strength of the disease name.

## Structure worth reviewing

Six nodes: age-associated inflammatory susceptibility → bursal macrophage and synoviocyte activation → a feed-forward IL-6 amplification loop, which then branches three ways.

The branch worth checking is **IL-6-mediated constitutional symptoms**, modelled as its own node rather than as a downstream effect of bursitis. The justification is that IL-6 acts on the HPA axis and on peripheral and central pain pathways to produce fatigue, sleep and mood disturbance — symptoms that pain from bursitis would not account for. Modelling it separately makes the entry predict that IL-6 blockade should improve fatigue and mood, not just joint symptoms, which seemed worth stating rather than collapsing.

Treatments deliberately contrast the two mechanisms: glucocorticoids suppress the bursal infiltrate broadly (fast, but 40-60% relapse on taper and >50% never come off), while tocilizumab targets the amplification loop directly. PMR-SPARE is recorded under `clinical_trials`.

`clinical_burden` is `MODERATE` on an unusual profile: no clear effect on survival or organ function, but real quality-of-life impact, with much of the long-term harm coming from cumulative steroid exposure rather than from the disease.

## Validation

- Evidence snippets: **27/27 verified** against the local reference cache

Closes #10834

---

### DIFF
--- a/history/disorders/Polymyalgia_Rheumatica/2026-09-04T025048Z-claude-code-e1153f.yaml
+++ b/history/disorders/Polymyalgia_Rheumatica/2026-09-04T025048Z-claude-code-e1153f.yaml
@@ -1,0 +1,30 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Polymyalgia_Rheumatica
+  path: kb/disorders/Polymyalgia_Rheumatica.yaml
+session:
+  id: 2026-09-04T025048Z-claude-code-e1153f
+  timestamp: '2026-09-04T02:50:48Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10851
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - phenotypes
+  - genetic
+  - biochemical
+  - treatments
+  summary: 'Address review: ontology precision, genetic, biochemical, methotrexate'
+  details: |
+    Acted on the REQUEST_CHANGES review on PR 10851. Five mechanical ontology fixes: fibroblast-like synoviocyte rebound from CL:0000214 synovial cell to CL:0002301 type B synovial cell at both sites, which matters because the cited abstract distinguishes macrophages from FLS and type A synovial cells are the macrophage-like population; morning stiffness now uses the pre-composed HP:0005197 rather than HP:0001387 plus temporality DIURNAL; the Systemic Acute-Phase Response node now carries GO:0006953 acute-phase response rather than interleukin-6 production; the IL-6-Mediated Constitutional Symptoms node now carries GO:0070102 interleukin-6-mediated signaling pathway, which also removes the identical term-plus-modifier pair that made the two nodes indistinguishable in the process layer; and onset_category moved from MIDDLE_AGE to LATE, which the cited evidence actually supports. The genetic section is the substantive part and does not do what the review suggested. The review proposed carrying HLA-DRB1 across from Giant_Cell_Arteritis, but the PMR literature does not support that: the Spanish cohort found the DRB1 allele distribution in PMR very similar to controls, with DRB1*0401 and *0404 overrepresented in GCA only, and the authors conclude the susceptibility alleles for the two diseases are different. HLA-DRB1 is therefore recorded as MODIFIER with a REFUTE evidence item against susceptibility and a SUPPORT item for the DRB1*09 relapse association. That turns the reviewer suggestion into a second independent line of evidence for the lump/split call rather than an unexamined borrowing from GCA. Added a biochemical section for ESR and CRP with the treated-flare caveat, and a methotrexate treatment carrying the randomised placebo-controlled trial, since presenting tocilizumab as the only steroid-sparing route misstated the landscape; methotrexate deliberately carries no target_mechanisms link because the trial endpoints are steroid sparing and relapse rather than a measured mechanistic effect. Replaced the EULAR scope-fragment snippet on glucocorticoid therapy with the dose and relapse sentences that actually carry the claim, and added bursitis-in-more-than-half and mortality-not-increased evidence. Prevalence stays qualitative: no Orphanet record for PMR is cached and no source gives a rate. All 41 snippets verified; schema, term, grading, length and offline gates pass.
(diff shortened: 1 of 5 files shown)


gingin77:
Pushed `07cc655c6c`. All five ontology findings applied, plus methotrexate, `genetic` and `biochemical`. One finding I handled differently, and it turned out to be the most interesting part of the review.

## Findings 1–5, applied

- `fibroblast-like synoviocyte` → `CL:0002301` (type B synovial cell) at both sites. Your reason for it being blocking rather than cosmetic is the right one: the cited abstract names macrophages *and* FLS separately, and type A synovial cells are the macrophage-like population, so the parent term was collapsing a distinction the evidence draws.
- Morning stiffness → pre-composed `HP:0005197`.
- `Systemic Acute-Phase Response` → `GO:0006953`; `IL-6-Mediated Constitutional Symptoms` → `GO:0070102`. The graph consequence you point out is the real argument — the two nodes carried an identical term + modifier, so the separate-route claim the prose works hard to make wasn't represented in the process layer at all.
- `onset_category` → `LATE`. My own two snippets said "≥50" and "older than 50"; `MIDDLE_AGE` (40–60) was the one value they argue against.

## Finding 7's `genetic` item — the literature says the opposite

You suggested carrying HLA-DRB1 across since `Giant_Cell_Arteritis` already has it. I fetched PMID:15305244 to do that, and it says:

> The distribution of DRB1 alleles was very similar between PMR patients and controls.

with DRB1\*0401 and \*0404 overrepresented in **GCA** only, and the conclusion:

> Our data suggest that the HLA-DRB1 alleles associated with susceptibility for developing PMR and GCA are different.

So carrying it across would have been exactly the GCA-borrowing this entry is otherwise careful to avoid. HLA-DRB1 is now recorded as `relationship_type: MODIFIER`, with a `REFUTE` evidence item against susceptibility and a `SUPPORT` item for the DRB1\*09 relapse association.

That inverts the finding usefully: instead of a gap filled by analogy, the entry now has a **second independent line of evidence for the lump/split call** — PMR and GCA don't share their HLA associations either. Thanks for pushing on it; I wouldn't have looked otherwise.

## Methotrexate, biochemical, and the rest

- **Methotrexate** added with the randomised placebo-controlled trial (28/32 vs 16/30 off prednisone at 76 weeks). You were right that presenting tocilizumab as the only steroid-sparing route misstated the landscape. It carries **no** `target_mechanisms` link, deliberately: the trial endpoints are steroid sparing and relapse, not a measured effect on any node here.
- **`biochemical`** added for ESR and CRP, with the treated-flare caveat the entry already discussed in prose.
- **The L496 EULAR scope fragment** — replaced, not defended. You were right it was a scope list rather than a finding. The dose and relapse sentences from PMID:28774422 carry the claim properly.
- Also added, from the same review: bursitis "in more than half of patients with active disease", and "mortality is not increased in patients with isolated disease" supporting the `clinical_burden` rationale.

## Prevalence — kept qualitative, and here's why

I'd already added a prevalence block before your review landed, but not a numeric one, and I've left it that way. There is **no Orphanet record for PMR** in the reference cache — I checked; `ORPHA_397` is giant cell arteritis, which mentions polymyalgia but isn't a PMR record. None of the cached sources gives a rate. So the block is `prevalence_class: COMMON` with `measure_type: UNKNOWN`, no `rate_per_100000`, population recorded as the at-risk over-50 group, and `notes` stating plainly that no numeric rate is recorded and why. Guessing between point prevalence and incidence is what the structured slots exist to prevent.

## Not done

- **`HP:0030834` / `HP:0030838` for shoulder and hip pain** (suggestion 1). Your tension argument is real — `Arthralgia` is joint pain and the entry argues the lesion is peri-articular — but splitting one phenotype into two changes the pathograph shape and the `target_phenotypes` on two treatments. I'd rather that be a deliberate change than bundled into a review-response push. Happy to do it if you want it here.
- **`Elevated Acute-Phase Reactants` frequency** — left at `FREQUENT`. You're probably right that 80–90% makes it `VERY_FREQUENT`, but as you noted that needs a source with a figure, and none of the cached sources gives one.
- **`classifications`** and **`parents` alignment** — separate passes.
- **Deep-research artifact** (finding 6): confirmed absent, and the narrow scope was deliberate. Noted in the PR description.

Thanks also for checking `UBERON:0003668` and confirming it wasn't a compromise — useful to have that recorded.

**Validation:** 41/41 snippets verified · `linkml-validate` clean · term validation passed · grading, length, duplicate-key, enum, entity-ref and causal-target checks all pass.

---

PR #10852, opened 2026-09-04
Curate Frank-Ter Haar syndrome (MONDO:0009579, SH3PXD2B)

### LINKED ISSUE #10835: Curate Frank-Ter Haar syndrome (MONDO:0009579)
Curate a dismech entry for **Frank-Ter Haar syndrome** ([MONDO:0009579](https://monarchinitiative.org/MONDO:0009579)) — an autosomal recessive skeletal dysplasia from loss of SH3PXD2B (Tks4), a podosome adaptor, producing megalocornea, brachydactyly, bowed long bones, wide anterior fontanelle, and cardiac defects.

Leaning DISEASE. The stub records no MONDO descendants and one causal gene (`hgnc:29242` SH3PXD2B), and the mechanism is a single chain: Tks4 loss → failed podosome/invadopodium assembly → impaired MT1-MMP delivery and pericellular matrix degradation → defective osteoclast and mesenchymal cell function → the skeletal and ocular phenotype.

**Naming caution.** The stub carries "Melnick-Needles syndrome, autosomal recessive" and "Borrone dermatocardioskeletal syndrome" as synonyms. The first is a historical misattribution — Melnick-Needles is X-linked FLNA and is a genuinely different disease; do not let its literature or its evidence leak into this entry. Borrone is now generally accepted as allelic to Frank-Ter Haar, so that one is a real synonym and its case reports are usable. Worth stating this split explicitly in the entry so the next curator does not have to re-derive it.

Stub: `stubs/Frank-Ter_Haar_Syndrome.yaml`

gingin77:
Adds a Disease entry for Frank-Ter Haar syndrome and retires its stub.

> **Revised after review** (`a12f4ee7c3`) — pathograph connected, Brachydactyly replaced with Camptodactyly, `prevalence`/`diagnosis`/`treatments` added. See [the reply](https://github.com/monarch-initiative/dismech/pull/10852#issuecomment-5534864164) for the full response, including two suggestions I declined with reasons.

## Curation scope

This was curated from a **targeted literature search**, not a deep-research provider run, so there is no `research/` artifact. That was deliberate, but the review showed what the narrower scope cost: the epidemiology, diagnosis and management content was missing, and one clinical review (PMID:21453629) supplied all three plus independent phenotype support. A broader sweep at the outset would have caught that.

## The synonym trap

The stub carried two synonyms needing **opposite** treatment:

- **"Borrone dermato-cardio-skeletal syndrome"** is a genuine synonym for the SH3PXD2B-mutant cases — two BDCS kindreds carry homozygous SH3PXD2B mutations with complete loss of protein, and the authors conclude a proportion of BDCS and FTHS are allelic. Its case reports are usable. Note "a proportion": a third BDCS family was excluded from the locus, so this is not a clean identity, and `notes` says so.
- **"Melnick-Needles syndrome, autosomal recessive"** is a historical misattribution and is **not** usable. Melnick-Needles is X-linked *FLNA* — different gene, different inheritance, different disease. Importing its literature would silently corrupt this entry.

Both decisions are recorded in `notes`.

## Genetic heterogeneity is stated, not smoothed over

Only 7 of 13 FTHS families in the founding study carried SH3PXD2B mutations. That stays in the `genetic` block because it limits what the entry asserts.

It also drove a structural choice: the pathograph is keyed on **TKS4 protein loss**, not genotype — a mutation-negative individual still had reduced TKS4 in dermal fibroblasts, implying a shared final mechanism reachable by other genetic routes. Keying on the protein means the mechanism still describes those patients.

## Where the mechanism is sharpest

The `Incomplete Podosome Formation and Failed MT1-MMP Recruitment` node rests on a differential rescue: reintroducing Tks4 restores both podosome formation and matrix degradation, while overexpressing the related Tks5 restores podosome formation *only*. That non-redundancy in the proteolytic arm is why losing TKS4 alone causes disease, and it connects this syndrome to the MMP14-related collagen-remodelling disorders — MT1-MMP is the protease TKS4 fails to deliver.

The matrix-accumulation explanation for facial coarsening and refractory glaucoma is `directness: INDIRECT`, because the source frames it as a hypothesis.

## Phenotype provenance

`Camptodactyly` replaced an earlier `Brachydactyly` entry. No reference in this set reports short digits; the digital anomaly the literature describes is a flexion deformity. The substitution is recorded in the phenotype description.

`Developmental Delay` is intentionally **not** connected to the pathograph and carries **no** `frequency`. No source proposes a route from matrix-remodelling failure to impaired neurodevelopment, and the sources say "variable", which is severity rather than a reported proportion.

## Animal models

Both carry `modeled_mechanisms`. The Tks4-null mouse `RECAPITULATES` the malformation node (phenocopying the majority, not all, of the defects). The pretzel zebrafish `PARTIALLY_RECAPITULATES` the fibrotic node, with two typed divergences from the authors' own caveats: `SPECIES_MISMATCH` and `BOUNDARY_OMISSION` (skeletal involvement is mild, so bone phenotypes can't be read off it).

`INCOMPLETE_PHENOTYPE` would have fit better — CLAUDE.md flags it as a likely future addition — but it isn't in the enum yet.

## Validation

- Evidence snippets: **42/42 verified** against the local reference cache
- `check_snippet_grading`, `check_snippet_length`, `check_title_snippets`, `check_folded_hyphens`, `check_duplicate_yaml_keys`, `check_enum_values`, `check_entity_refs`, `check_causal_targets`: all OK

Closes #10835

### DIFF
--- a/kb/disorders/Frank-Ter_Haar_Syndrome.yaml
+++ b/kb/disorders/Frank-Ter_Haar_Syndrome.yaml
@@ -12,7 +12,7 @@ description: >-
   surface, which places the disease in the defective collagen remodelling
   spectrum alongside the MMP14-related multicentric osteolytic syndromes.
   Clinically it presents with craniofacial dysmorphism, megalocornea with
-  frequently severe glaucoma, brachydactyly and other skeletal malformations,
+  frequently severe glaucoma, camptodactyly and other skeletal malformations,
   congenital heart defects, and fibrotic soft-tissue change with joint
   contractures. Severity varies widely, from survival into adulthood to lethal
   respiratory failure in infancy, and the same variant has produced both.
@@ -238,7 +238,7 @@ pathophysiology:
   description: >-
     Failure of matrix remodelling during development produces the skeletal
     dysplasia and craniofacial phenotype: coarse facies, hypertelorism, wide
-    anterior fontanelle, brachydactyly and vertebral abnormalities including
+    anterior fontanelle, camptodactyly and vertebral abnormalities including
     scoliosis. Cardiac malformation arises in the same developmental window and
     is reproduced in the Tks4-null mouse.
   evidence:


gingin77:
Thanks — this was a genuinely useful review. Pushed `a12f4ee7c3` acting on all six IMPORTANT findings and two of the suggestions. Two things I did differently, both explained below.

## Finding 2 was worse than you diagnosed, and you were right to pull the thread

You flagged that the Brachydactyly quote said only "skeletal malformations". I went to check whether a better quote existed and found that **no reference in this entry's set mentions brachydactyly at all** — and you'd already spotted the tell, that the case report describes *camptodactyly*. I fetched PMID:21453629, whose clinical-features list says "flexion deformity of the fingers". So brachydactyly was not an under-evidenced phenotype; it was the wrong phenotype, and I'd imported it from what a skeletal dysplasia is *expected* to look like rather than from what these papers say.

Removed it. The entry now carries **Camptodactyly** (`HP:0012385`), cited to that feature list and to the case report's own "camptodactyly", with the substitution recorded in the description so it isn't silently re-added.

That new reference then solved much of finding 3 for free: its main-clinical-features sentence independently supports hypertelorism, wide fontanels, kyphoscoliosis, congenital heart defects, macrocornea and developmental delay. Coarse Facial Features, Scoliosis and VSD each now carry a feature-list citation alongside the single case, so the frequencies no longer rest on one patient. I also added **Wide Anterior Fontanelle**, which my own description mentioned but the phenotype list omitted.

## Finding 1 — connected, with one deliberate exception

Hypertelorism, Scoliosis and the new Wide Anterior Fontanelle now hang off `Skeletal and Craniofacial Malformation`; Megalocornea off `Defective Pericellular Matrix Remodelling`. Bare names, as you noted. `check_causal_targets` clean.

**Developmental Delay I left unconnected on purpose**, and wrote the reason into its description rather than leaving it silent — your suggested third option. No source here proposes a route from matrix-remodelling failure to impaired neurodevelopment, and the one patient with detailed imaging had a thin corpus callosum with no mechanism offered.

**I did not add the `Megalocornea → Glaucoma` sequelae edge**, because I think the arrow may run the other way. Corneal enlargement in congenital glaucoma is buphthalmos — a *consequence* of raised intraocular pressure — and the case you're drawing on describes exactly that ("buphthalmos worsening"). Committing to megalocornea → glaucoma would encode a direction the sources don't establish and that may be backwards. Both now hang off the matrix node independently.

## Findings 4, 5 and 6 — added

`prevalence` (CASES_IN_LITERATURE, ULTRA_RARE, quoting the 40-patients/20-mutations sentence), `diagnosis` (molecular confirmation, plus PMID:28694206 on exome sequencing given the heterogeneity), and `treatments` (supportive care, and glaucoma management recorded as refractory rather than effective). The treatments carry **no** `target_mechanisms` link — nothing reported acts on the podosome defect, and a link would overstate it.

## Two things I didn't take

**The `fibrotic_response` conformance.** I looked and decided against it. That module's `Excessive ECM Deposition` node is reached through `Mesenchymal Cell Activation` — TGF-beta-driven myofibroblast transdifferentiation, described there as "the central effector step". This disease has no such step: matrix accumulates because cells cannot degrade it, not because activated myofibroblasts overproduce it. Conformance is supposed to mean the entry includes the module's expected processes and causal edges, so declaring it here would be a false claim about the mechanism. You said "worth recording either way" — this comment is that record.

**Deep-research artifact (finding 7).** Confirmed absent, and the narrow scope was deliberate: this was curated from a targeted literature search rather than a DR provider run. Your inference was right, though — items 4-6 are exactly what that scope cost, and the fix came from one clinical review I should have pulled at the outset. I've noted the scope in the PR description.

## Also changed

Regraded the PMID:33234702 background sentences to `OTHER` per your suggestion — it's a zebrafish paper and those sentences are review-style background about the human syndrome. That surfaced a real defect: a fifth instance of the same sentence was still `HUMAN_CLINICAL`, which `check_snippet_grading` caught. Now consistent.

Dropped the `OCCASIONAL` frequency on Developmental Delay. You were right that "variable" describes severity, not occurrence, and neither source gives a denominator.

Grouping membership: agreed `Fibrotic_Disorders` may fit and `Cartilage_Matrix_Degradation_Disorders` does not. Separate PR.

**Validation:** 42/42 snippets verified · `linkml-validate` clean · term validation passed · snippet-grading, snippet-length, title-snippet, folded-hyphen, duplicate-key, enum, entity-ref and causal-target checks all pass.

---

gingin77:
Fixed in `0d17b6650b`. You're right, and it's the kind of inconsistency that's worse than the original error — the summary would have asserted brachydactyly while the phenotype list sat below it explaining that brachydactyly was wrong.

Both prose mentions now say camptodactyly:
- `:15` top-level description
- `:241` `Skeletal and Craniofacial Malformation` node

The third mention at `:414` is deliberate — it's the provenance note in the Camptodactyly phenotype recording that the substitution happened and why, so the wrong phenotype isn't re-added later. I've left that one.

Nothing else changed. `linkml-validate` clean, 42/42 snippets still verified.

## On the two suggestions

**Orphanet (ORPHA:137834)** — agreed this is the right source, and for the reason you give: it would settle the frequencies from per-feature frequency bands rather than by inference from a feature list, and would answer the brachydactyly question independently rather than on my absence-of-evidence argument. I'm leaving it as follow-up, and not just for scope reasons: `data/orphadata/` here holds only `MANIFEST.yaml`, so generating that record means `just refresh-orphadata` first — a large bulk download against a checksum pinned to an unversioned URL, which per CLAUDE.md routinely drifts and needs a deliberate `--repin` decision plus a review of the resulting cache diff. That's a change with its own reviewable surface, not something to fold into a review response. Worth doing as its own PR, and it would improve several skeletal-dysplasia entries at once rather than just this one.

**A more specific NCIT term for the glaucoma treatment** — I checked, since I do have ontology access here. NCIT's glaucoma-directed terms are procedures like trabeculectomy that this disease's literature doesn't report being used, so binding one would assert a specific intervention the sources don't support. `NCIT:C15747` Supportive Care with the specificity carried in `name` and `description` is the honest binding given the entry only establishes that the glaucoma is refractory. Happy to revisit if the Orphanet record or an ophthalmic source names actual interventions.

Thanks for withdrawing the `Megalocornea → Glaucoma` suggestion explicitly — useful to have the buphthalmos reasoning recorded in the thread rather than only in my reply.

---

PR #10861, opened 2026-09-04
Curate hypokalemic tubulopathy and deafness (MONDO:0859167, KCNJ16)

### LINKED ISSUE #10836: Curate hypokalemic tubulopathy and deafness (MONDO:0859167)
Curate a dismech entry for **hypokalemic tubulopathy and deafness** ([MONDO:0859167](https://monarchinitiative.org/MONDO:0859167)) — autosomal recessive loss of KCNJ16 (Kir5.1), the partner subunit of Kir4.1, causing a renal tubulopathy with hypokalemia and a disturbed acid-base phenotype together with sensorineural hearing loss.

Leaning DISEASE. The stub records no MONDO descendants and one causal gene (`hgnc:6262` KCNJ16), and the chain is conserved: loss of Kir5.1 → non-functional Kir4.1/Kir5.1 heteromer at the basolateral membrane → failed K+ recycling supporting the Na/K-ATPase in distal nephron and in the inner ear stria → salt-wasting tubulopathy with hypokalemia plus sensorineural deafness.

**Named Entity Confusion guardrail — this is the main risk on this entry.** `EAST_Syndrome` is already curated and is KCNJ10 (Kir4.1), the *partner* subunit; the two form the same heteromeric channel and their entries will look alike. They are different diseases: EAST/SeSAME adds epilepsy and ataxia from the CNS role of Kir4.1, whereas KCNJ16 loss spares the CNS and reportedly produces a more variable acid-base picture than EAST's classic hypokalemic metabolic alkalosis. The existing `EAST_Syndrome` entry already mentions KCNJ16 contextually, and that mention is correct — it is not an entry for this disease. Evidence must be drawn from KCNJ16-specific reports (the entity was delineated relatively recently), never borrowed from the KCNJ10 literature.

Stub: `stubs/Hypokalemic_Tubulopathy_And_Deafness.yaml`

gingin77:
Adds a Disease entry for KCNJ16-related tubulopathy and retires its stub.

## This entry sits right next to EAST syndrome, on purpose

`EAST_Syndrome` is already curated. It is KCNJ10/Kir4.1 — the **obligate partner subunit of the very channel disrupted here**. The two entries will inevitably look alike, and the literature routinely discusses them together, so this is the highest-confusion pairing I could have curated. The guardrail is recorded in `notes`, and the review cited states the differences directly:

| | EAST/SeSAME (KCNJ10) | This entry (KCNJ16) |
|---|---|---|
| Hearing loss | yes | yes |
| Hypokalemia | yes | yes |
| Acid-base | metabolic **alkalosis** | **acidosis**, variable severity |
| CNS | epilepsy, ataxia, intellectual disability | unaffected |

No evidence is borrowed from the KCNJ10 literature. The existing `EAST_Syndrome` entry mentions KCNJ16 contextually; that mention is correct and is not an entry for this disease.

## Why the pathograph branches the way it does

Kir5.1 forms no homomeric channel — it is an obligate partner. So its loss isn't expressed through "the Kir5.1 channel" but through the *different* heteromers it completes in each nephron segment: Kir4.2 proximally, Kir4.1 distally. That is why a single gene produces a tubulopathy spanning two segments with two different consequences (bicarbonate handling and salt reabsorption), and the node structure follows that rather than flattening it into a single "tubulopathy" node.

The `genetic` block notes something I found neat: the *absence* of CNS involvement in patients is itself evidence, arguing against Kir5.1 forming functionally important channels with other partners in the brain.

## A dating trap in the mouse model, recorded rather than smoothed over

The Kcnj16-null mouse paper is titled "…contrasts with that observed in SeSAME/EAST syndrome" and describes its acidotic mice as showing "the opposite of the phenotype seen in SeSAME/EAST".

Read carelessly, that sounds like the model contradicts the human disease. It doesn't. The mouse paper is from **2011**; the human KCNJ16 entity was delineated in **2021**. The authors could only compare against EAST because this disease did not yet exist in the literature — and their acidotic, hypokalemic mice actually *anticipate* the human KCNJ16 phenotype.

I've recorded the model as `PARTIALLY_RECAPITULATES` with `limitations` stating explicitly that the correspondence is one drawn here, not claimed by the authors, and that the murine hyperchloremic-acidosis-with-hypercalciuria picture is more precisely specified than the "variable severity" acidosis reported in patients, so the match shouldn't be over-read. Two typed divergences: `SPECIES_MISMATCH` (residual Kir4.1 channels lose pH sensitivity and conductance actually *rises* in the knockout — unestablished in humans) and `TEMPORAL_SCOPE`.

## One deliberate omission

The treatment carries **no** `target_mechanisms` link. Management is symptomatic potassium and alkali replacement; nothing restores the channel. The mouse data showing chronic hydrochlorothiazide abolished the acidosis is tempting but is model evidence, not human evidence, so it stays under `animal_models` rather than becoming a treatment claim.

## Validation

- Evidence snippets: **19/19 verified** against the local reference cache

Closes #10836

---


gingin77:
You caught a real error, and it's the worst kind: the entry committed the exact substitution its own `notes` guardrail was written to prevent. Pushed `3649a0b6a0`.

## The cochlear node was wrong

I verified against the cached review and you're right on both halves:

> Kir4.1 is expressed primarily in intermediate cells of the stria vascularis and is responsible for endolymph composition and endocochlear potential (Hibino et al., 1997). Kir5.1 is expressed mainly in the root cells and spindle cells of the spiral ligament (Hibino et al., 2004b)

I had assigned Kir5.1 the one compartment the source assigns to Kir4.1 — inside the entry whose whole point is keeping those two subunits apart. Fixed:

- Rebound to `UBERON:0006725` (spiral ligament).
- Reworded the description to state the Kir5.1 cochlear role as **proposed**, quoting the review's own hedge ("proposed that Kir5.1 is involved in K+ cycling and *may* play a role").
- Added three evidence items: the spiral-ligament localisation, the Kir4.1/stria-vascularis assignment (so the entry records what Kir5.1 is *not* responsible for), and the proposed-role hedge.
- Recorded the error in the node description, so a future editor doesn't quietly restore the stria vascularis.

That last point is the part I'd flag for the record: the guardrail existed, was written in the right words, and I still made the substitution. Prose in `notes` didn't stop it — checking the source did.

## `evidence_source`

Both regraded: the homomeric-channel statement to `IN_VITRO` (the review derives it from oocyte coexpression and says so), and the inner-ear expression statement to `MODEL_ORGANISM` (rodent immunohistochemistry). I didn't touch your two borderline cases.

## The under-consumed review

You were right that it was sitting there. Added, all from the existing cache:

- **The acid-load finding** — preserved urinary acidification with absent ammonium rise — now supports both the `diagnosis` block and the proximal node. You were right that this is the most valuable one: it locates the defect in ammoniagenesis rather than distal proton secretion, which is what separates this from a distal RTA at the bedside.
- Hearing loss now carries `onset_category: CHILDHOOD` and the high-frequency pattern, with the note that the audiogram does *not* discriminate from EAST/SeSAME even though the acid-base picture does.
- The one patient with alkalosis, supporting `FREQUENT` rather than obligate acidosis.
- Greater severity with Kir4.2 coexpression, in `genetic` notes, as the likely reason the proximal arm is prominent.

## The mouse that hears normally

Added as a second `animal_models` entry: `FAILS_TO_RECAPITULATE` against Sensorineural Hearing Loss, `fidelity: LOW`, with an **INVALIDATING** `SPECIES_MISMATCH` divergence quoting the review's compensation hypothesis. Plus a `HUMAN_MODEL_MISMATCH` discussion attached to both the cochlear node and the phenotype.

You were right that leaving it out was the odd gap given the effort on the other mouse — and it's worse than an omission, because the renal knockout's `limitations` said "deafness is not addressed by this study", which left a reader concluding the question was unstudied when in fact it had been studied with a negative result. The discussion records the consequence: the cochlear arm of this disease currently has no validated animal model.

## Suggestions

- **Treatment evidence** — agreed the uncited treatment read as unsourced. Added a `notes` line saying explicitly that no cached source states a management recommendation, so there is nothing quotable to attach and manufacturing one would be worse.
- **`therapeutic_agent`** — added potassium chloride (`CHEBI:32588`) and sodium bicarbonate (`CHEBI:32139`).
- **`classifications.channelopathy_category`** — not done here. It's a separate concern from this review's findings and the `disease-classification` skill deserves a proper pass rather than a guessed value; happy to do it if you'd rather it land in this PR.
- **SIDS association** — agreed, left out. Recording the decision as you suggested.

**Validation:** 33/33 snippets verified · `linkml-validate` clean · term validation passed · grading, length, duplicate-key, enum, entity-ref and causal-target checks all pass.

---

PR #10864, opened 2026-09-04
Resolve CMTX1 (MONDO:0010549) as a subtype of CMT type X

gingin77:
## CMTX1 turned out to be already curated

I claimed MONDO:0010549 to curate as a new `kb/disorders/` entry. It didn't need one — CMTX1 is already the first `has_subtypes` entry of `kb/disorders/Charcot-Marie-Tooth_Disease_Type_X.yaml`, and that file's entire pathophysiology chain *is* the CMTX1 mechanism:

```
Connexin-32 Reflexive Gap Junction Loss in Schwann Cells
  -> Impaired Radial Metabolic Support of the Axon
    -> Demyelination with Secondary Axonal Loss
      -> Length-Dependent Distal Neuropathy

Oligodendrocyte Connexin-32 Loss and Transient CNS Dysfunction
```

So the honest `entry_type` for the stub is `SUBTYPE`, and this PR records that rather than creating a near-duplicate entry.

## What changed

- **`subtype_term` bound to MONDO:0010549** on the CMTX1 subtype. None of the six CMTX subtypes carried one. Under the coverage rule in `CLAUDE.md`, a `has_subtypes` term counts as curated, so this retires the concept from the curation queue.
- **`review_notes` on that subtype** explaining why CMTX1 stays a subtype, so it isn't nominated a third time.
- **Stub deleted.**
- **History record** under `history/disorders/Charcot-Marie-Tooth_Disease_Type_X/`.
- One term-cache row (`MONDO:0010549`), in canonical sorted position.

No evidence, phenotype, or pathophysiology content was touched, and no snippet was altered.

## Why not promote CMTX1 to its own entry

It's a judgment call and worth stating either way. The parent's pathograph is CMTX1's pathograph. Splitting CMTX1 out would leave `Charcot-Marie-Tooth_Disease_Type_X.yaml` as a shell of four rare, mechanistically unrelated loci — AIFM1, PRPS1, PDK3, and the unresolved Xq27.1 insertion — two of which already cross-reference their own entries. That trade looks bad, so the reasoning now lives in `review_notes` rather than only in this PR description.

## How my duplicate preflight missed it

[...]

PR #10869, opened 2026-09-04
Curate Dilated Cardiomyopathy 1X (MONDO:0012704, FKTN)

### LINKED ISSUE #10844: Curate dilated cardiomyopathy 1X (MONDO:0012704)
Curate a dismech entry for **dilated cardiomyopathy 1X** ([MONDO:0012704](https://monarchinitiative.org/MONDO:0012704)) — CMD1X, caused by FKTN (fukutin) variants, is a dilated cardiomyopathy arising from alpha-dystroglycan hypoglycosylation without the overt skeletal muscular dystrophy seen in the other fukutinopathies; it is the cardiac-restricted end of the dystroglycanopathy spectrum and is notable for responding to standard heart-failure therapy.

The lump/split question worth resolving during curation is its relationship to the existing `kb/disorders/Dystroglycanopathy.yaml` entry — CMD1X is mechanistically a dystroglycanopathy but is clinically a separate presentation, so the likely answer is a `DISEASE` entry that `conforms_to` shared mechanism rather than a `has_subtypes` row.

Stub: `stubs/Dilated_Cardiomyopathy_1X.yaml`

gingin77:
## CMD1X is not X-linked, and the entry says so

The most useful thing this curation produced is a correction. **CMD1X is autosomal recessive.** The "X" is simply the position reached in OMIM's sequential lettering of the CMD1 series; FKTN is autosomal at 9q31.2, and every reported patient is a compound heterozygote. X-linked dilated cardiomyopathy is a genetically separate DMD-related entity at Xp21, and the two are routinely conflated — including by my own claim issue, which described this as an X-linked disease. The entry carries a top-level `notes` block stating the distinction, because a curator arriving from the disease name will otherwise make the same mistake.

## What the disease is

Fukutin is a Golgi ribitol-phosphate transferase. With fukutin-related protein it builds the tandem ribitol-phosphate scaffold on the core M3 O-mannosyl glycan of alpha-dystroglycan, and matriglycan is polymerised onto that scaffold. Matriglycan is what lets alpha-dystroglycan bind laminin. Null FKTN genotypes give Fukuyama congenital muscular dystrophy — severe myopathy, brain malformation, intellectual disability. CMD1X is the opposite end of the same allelic series: enough residual activity to spare brain and skeletal muscle, not enough to spare myocardium.

So these patients present to cardiology with what looks like isolated dilated cardiomyopathy and normal intelligence, while their skeletal muscle biopsy is biochemically abnormal and histologically near-normal. Why the heart is the tissue that cannot tolerate the shortfall is the entity's open question, and the entry records it as one rather than resolving it.

## Pathograph

Seven nodes, from the enzyme defect to systolic failure:

```
Reduced Fukutin Ribitol-Phosphate Transferase Activity        MOLECULAR
  -> Alpha-Dystroglycan Hypoglycosylation and Loss of Laminin Binding   MOLECULAR
    -> Weakened Sarcolemma-to-Basement-Membrane Linkage in Cardiomyocytes   CELLULAR
      -> Impaired Compensatory Hypertrophic Signaling         CELLULAR
      -> Ventricular Remodeling and Myocardial Fibrosis       TISSUE
        -> Dilated Cardiomyopathy with Systolic Failure       ORGANISM

Relative Sparing of Skeletal Muscle and Brain                 TISSUE
```

`conforms_to` targets: the alpha-dystroglycan glycosylation node of `pial_basement_membrane_radial_glial_endfoot_failure` (checked first that this node is written generically and names FKTN explicitly, rather than being brain-scoped — otherwise the honest answer would have been no conformance), plus three nodes of `cardiomyopathy_maladaptive_remodeling`.

The two-branch structure past the sarcolemma is deliberate and follows the mouse data, which show that membrane fragility alone is *not* sufficient: cardiac Fktn elimination strips the sarcolemmal glycan at all developmental stages, yet dysfunction appears only in later adulthood. The signalling branch (Akt, MEF2–histone deacetylase) is what makes the phenotype stress-sensitive, which fits a human disease with such variable onset.

## Evidence discipline

All mouse evidence is graded `MODEL_ORGANISM` with `directness: INDIRECT`, because the models are complete nulls and the human genotype is hypomorphic. A `HUMAN_MODEL_MISMATCH` discussion records why no current model can answer the tissue-selectivity question: the available knockouts are made cardiac-restricted by a Cre driver rather than by the biology, so they cannot test the hypothesis that a partial, body-wide reduction crosses a threshold in cardiomyocytes that it does not cross in skeletal myofibres.

19/19 snippets verified against the reference cache.

## Grouping

Added as a member of `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml`, and its rationale updated from five mechanistic classes to six. FKTN is the only member whose failing component lies *outside* the myocyte — every other member perturbs something the cell builds and uses internally (the contractile stroke, the Z-disc, the nuclear envelope, protein quality control, splicing).

That grouping is carefully curated and holds members out on purpose, so I checked the boundary before adding rather than after: MONDO places MONDO:0012704 directly under MONDO:0700335, satisfying the grouping's stated requirement that every member be a direct child of that class; and the "isolated" clause is not breached, because CMD1A/LMNA is already a member and carries a frank skeletal myopathy with contractures, which is more extracardiac involvement than CMD1X has. `just check-groupings` reports CMD1X SATISFIED on both necessary criteria.

## What deep research changed

One `claude_code` run. Reference validation was clean — 8/8 resolved, 0 unresolved, 2/2 quoted claims found in source. Its **term** validation was the part that earned its keep and changed two bindings before they reached the entry:

- `HP:0001635` is **Congestive heart failure**, not Dilated cardiomyopathy. The dilated cardiomyopathy phenotype is bound to `HP:0001644`; `HP:0001635` is used for the heart failure phenotype instead.
- `HP:0003749` is canonically **Pelvic girdle muscle weakness**, which is the label recorded.
- `GO:0006486` was flagged obsolete and not used.

## Deliberately not included

- **No environmental entry for pregnancy** as a decompensation trigger. The research reports a case, but the source was reachable only as a PMC identifier and could not be fetched into the reference cache as a citable PMID. An uncited environmental entry would fail `check-environmental-evidence` and would not deserve to pass it.
- **No prevalence record.** The entity is known from case reports and one series of six patients in four families. There is no honest population rate to state.
- **No `case_fractions`, no clinical trials** — same reason.

## Validation

`just validate`, `just validate-terms`, `just count-verified-snippets` (19/19), `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-qualifier-terms`, `just check-enum-values`, `just validate-grouping`, `just check-groupings`, `just validate-history` all pass.

Closes #10844

---
_Generated by [Claude Code](https://claude.ai/code/session_01PHYDoqVQvHMJimXjxsprAT)_


gingin77:
All four blocking items are closed, plus three of the non-blocking suggestions. Pushed as 4ec07fd.

**1. GeneReviews baseline.** PMID:20301486 tagged, and a "GeneReviews scope" paragraph added to `notes` following the `Dilated_Cardiomyopathy_1S` pattern. You were right that the cached record is `abstract_only` — it carries only the chapter front matter, so no snippet is quoted from it and the notes say why.

**2. `diagnosis:` block.** Four steps: echocardiography, serum CK measurement, muscle biopsy with alpha-dystroglycan glycoepitope staining and laminin overlay, and FKTN sequencing. Two things worth flagging in how they're written, since they're the parts a clinician would actually get wrong:

- The CK step is framed as the *trigger*, because your framing of the problem is right — the disease looks idiopathic until someone measures it.
- The biopsy step records that routine histology on the same specimen is near-normal, so a report of "minimal dystrophic change" does not exclude the diagnosis and the glycoepitope study has to be requested specifically.

**3. Elevated CK phenotype.** Added, bound to `HP:0003236`. Your caveat was correct — PMID:17036286's abstract does not mention CK, so this needed a new source. Fetched PMID:35743126, which reports elevated CK in a genetically confirmed compound heterozygote.

**4. `genetic:` section.** Expanded to the full allelic architecture: the ~3 kb founder retrotransposal insertion in the FKTN 3'UTR (87% of FCMD chromosomes, abolishes expression), p.Ser299Arg and p.Asn442Ser, deep-intronic splicing alleles, and the one-severe-in-trans-with-one-hypomorphic model. Three new evidence items from PMID:9690476 and PMID:35743126. Agreed this is the central explanation rather than colour — the entry was asserting it implicitly through `PARTIAL_LOSS_OF_FUNCTION` and not stating it.

**Non-blocking, taken:**

- `therapeutic_agent` on the GDMT treatment (`NCIT:C247` ACE Inhibitor, `NCIT:C29576` Beta-Adrenergic Antagonist), plus a `target_mechanisms` link so it joins the pathograph.
- **The pregnancy exposure is citable after all** — good call. PMC4630494 resolves to PMID:26566449 through the NCBI ID converter, and it is now an `environmental:` entry with `environmental_effect: EXACERBATES`. It carries a detail worth having: the cardiac parameters returned to pre-pregnancy baseline about a month postpartum while the skeletal weakness persisted, so the two compartments did not recover in step. No ECTO term is bound and `review_notes` says why — the haemodynamic load of pregnancy is not an exposure in ECTO's sense.
- The GO:0035269 conformance divergence is now stated in `notes` as deliberate.
- The PMID:27194101 title now matches the cache frontmatter's Unicode alpha.

**Not taken, and why:**

- **ICD/device therapy, arrhythmia and cardiomegaly phenotypes.** No source currently in this PR carries a usable quote for them. Adding them would mean asserting claims I can only support from the research narrative, which is a lead and not a citation.
- **`model:` on the history record's ai_agent actor.** This one is a constraint on me rather than a disagreement with the convention — the operating instructions for this session forbid writing any model identifier into an artifact pushed to a repository. The commit trailer carries it because the harness adds that itself. If the convention matters more than the constraint, a maintainer should add the line; I have deliberately left the field empty rather than filling it with something inaccurate.

**One thing worth recording for other curators.** Four NCIT codes I wrote from memory were wrong, and `just validate` caught all four: `NCIT:C16463` is *Computer Systems Development*, not Echocardiography; `NCIT:C1614` is *KRN5500*, not an ACE-inhibitor class; `NCIT:C265` is *Antidepressant Agent*, not a beta-blocker class; and `NCIT:C15189` is labelled *Biopsy Procedure*. All were replaced with codes looked up through `runoak`. The two wrong ones had already been written into `cache/ncit/terms.csv` and the `chemicalentityterm` enum cache by the failed runs, so `cache/` was reverted with `git checkout` and re-derived from the corrected entry — the resulting cache diff is empty, which confirms every term now used was already cached.

**Validation:** `just validate`, `just validate-terms`, `just count-verified-snippets` (**29/29**, up from 19/19), `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-qualifier-terms`, `just check-enum-values`, `just check-environmental-evidence`, `just check-snippet-length`, `just validate-history` all pass.

---

gingin77:
Merge conflict resolved. Pushed as `3abdf1e0`.

**One conflict, in `cache/enums/molecularfunctionterm_305077eac108.csv`** — both sides appended a row at the same point: `GO:0043236` here, `GO:0043560` on `main`. The two terms are unrelated and neither replaces the other, so this is the adjacent-insertion case: kept the union. `just normalize-cache` then left the file unchanged, because the union was already in canonical order.

Merged `main` in rather than rebasing, so the branch history stays valid for anyone who has it checked out.

**No curated content changed.** Verified against the post-merge tree rather than assuming:

| Check | Result |
|---|---|
| `just validate` | passes, **29/29** snippets verified |
| `check-duplicate-keys` / `check-entity-refs` / `check-causal-targets` / `check-enum-values` | clean |
| `check-groupings` on `Familial_Dilated_Cardiomyopathy` | CMD1X **SATISFIED** on both necessary criteria |

Diff against `main` is 19 files — the entry, its two history records, six reference caches, the research artifact and its citations sidecar, the grouping edit, the stub deletion, and six cache files. No stray deletions, no protected-path churn, no derived artifacts.

I also checked the other four branches in this batch against current `main`: #10864, #10921, #10922 and #10923 all still merge cleanly, so this was the only one that drifted.

---

gingin77:
Thanks — and one correction that should save you that by-hand check in future reviews.

**The `conforms_to` targets are gated, just not in the lane you looked at.** `test_conforms_to_module_node_references` lives in `tests/test_data.py` and runs under `just test-kb`, whose condition in `main.yaml` is:

```yaml
- name: Run whole-KB schema-conformance sweep
  if: github.event_name == 'merge_group' || steps.changes.outputs.schema == 'true'
  run: just test-kb
```

You're right that a `kb/`-only PR never fires the `schema` filter, so the sweep doesn't run on the PR check. But `merge_group` does fire it — and the merge queue builds the PR against current `main`, which is exactly the concurrent-module-edit scenario. So the case you were worried about is caught before the merge lands, on the combination that actually matters.

I went looking to file this as a gate gap, on the model of #9473 (entity refs made an ungated CI step precisely because curation PRs skip the pytest lane). It isn't one. The difference is that `check-entity-refs` protects against something reaching `main`, whereas here the merge-group run is the last gate before `main` and it does fire.

Your manual check was still the right call for the *review*, since the PR check hadn't run it — I'm noting where the real coverage is, not that checking was unnecessary.

**On the `full_text_attempted` stamping** — good catch, and worth stating for anyone reproducing this locally. My worktree is clean; nothing was staged. The frontmatter re-stamp happens on a networked `just validate` run and is validator bookkeeping, so it should be discarded rather than committed.

**The three carried-forward items stay carried forward.** The `NCIT:C64489` reachability question is a genuine one-command check and worth doing, but not on a push that would cost another review cycle on a PR whose content is settled — this branch has already had one approval dismissed by the merge push. Better as a follow-up than as a fourth head on this PR.

---

PR #10872, opened 2026-09-04
feat(Dilated_Cardiomyopathy_1P): curate PLN-related cardiomyopathy (calcium handling, allele-resolved)

### LINKED ISSUE #10854: Curate dilated cardiomyopathy 1P (MONDO:0012362)
Curating **dilated cardiomyopathy 1P** as a standalone entry in `kb/disorders/`.

| | |
|---|---|
| MONDO | `MONDO:0012362` |
| Causal gene | **PLN** (`hgnc:9080`) |
| Stub | `stubs/Dilated_Cardiomyopathy_1P.yaml` |

Phospholamban regulates SERCA2a, so the lesion is in calcium reuptake rather than in force generation. The R14del founder variant also has a distinct clinical course, with a high arrhythmic burden that drives ICD decisions, which is management the parent grouping does not describe.

- Nothing in `kb/disorders/`, `kb/groupings/` or `kb/modules/` on `origin/main` binds `MONDO:0012362`, and a search on the disease label and the gene name found no entry covering it either.
- No other issue, open or closed, references `MONDO:0012362` or this disease, and no open pull request does.
- PLN appears in several existing cardiomyopathy entries only as a gene-list member, never as the primary gene of an entry.

AO33:
Adds a knowledge base entry for **dilated cardiomyopathy 1P** (MONDO:0012362), the cardiomyopathy caused by variants in *PLN*, the gene for phospholamban. Phospholamban is the brake on SERCA2a, the pump that pulls calcium back into the sarcoplasmic reticulum so the heart can relax — so this disease is a calcium-handling disorder, not a contraction-machinery disorder like most of its neighbours in the numbered dilated cardiomyopathy series.

The other thing this entry does deliberately is keep the *PLN* variants apart instead of averaging them into one mechanism. The three well-studied alleles do genuinely different things, and the deep-research report and much of the review literature blur them.

Closes #10854

## The three alleles, and why they are curated separately

| Allele | What it does to the pump | Inheritance |
|---|---|---|
| **p.Arg14del** (R14del) | Superinhibits SERCA2a, and PKA phosphorylation cannot switch the inhibition off. Separately mislocalizes phospholamban into malformed sarco/endoplasmic reticulum clusters that block autophagosome–lysosome fusion. | Dominant |
| **p.Arg9Cys** (R9C) | Does **not** inhibit SERCA2a at all. It traps protein kinase A, which stops the *wild-type* phospholamban from being phosphorylated. The brake is applied in trans, by the normal protein. | Dominant |
| **p.Leu39Ter** (L39stop) | A true null — essentially no protein, and SERCA2a is left uninhibited. The opposite perturbation from R14del. | Heterozygotes have hypertrophy with normal contractility; only **homozygotes** get dilated cardiomyopathy |

R14del is also the only one with a validated variant-specific arrhythmia risk model, and that model is what drives defibrillator decisions rather than ejection fraction alone — the entry records that explicitly on the ICD treatment.

The pathograph branches accordingly:

```
                              ┌─ Non-Reversible SERCA2a Superinhibition (R14del) ─┐
                              │                                                   ├─→ Impaired SR Calcium Reuptake ─┐
PLN Regulatory Domain Variant ─┼─ PKA Sequestration (R9C) ─────────────────────────┘                                 ├─→ Dysregulated Calcium Homeostasis
                              │                                                                                     │
                              ├─ Loss of Phospholamban Protein (null alleles) ─────────────────────────────────────┘
                              │
                              └─ PLN Mislocalization / S-ER Disorganization ─→ Impaired Autophagic Flux
                                                                                    → Perinuclear PLN Deposits → ER Stress / UPR
```

Both arms then converge on contractile dysfunction, myocyte loss, replacement fibrosis, and from there split into the arrhythmic and the heart-failure endpoints.

## On the LUMP_INTO_PARENT recommendation

`dashboard/priority.json` scores this term `recommended_action: LUMP_INTO_PARENT`, `specificity_bucket: subtype_series`. I think that call is wrong here, and issue #9865 asks for exactly this lump/split pass over the numbered cardiomyopathy stubs, so here is the reasoning.

The recommendation is driven by `nec_risk: NUMBERED_SERIES` — that is, by the shape of the label "dilated cardiomyopathy 1P", not by anything about the biology. Two reasons that heuristic misfires on this term:

**The KB has already made the opposite call 17 times.** Dilated cardiomyopathy 1A, 1AA, 1B, 1D, 1DD, 1E, 1EE, 1FF, 1G, 1HH, 1I, 1JJ, 1S, 1Y, 1Z, 2G and Dilated Cardiomyopathy 100 are all standalone entries. Lumping 1P alone would be inconsistent, not conservative.

**The mechanism is not a variation on the parent's.** Look at what those sibling entries are about: TNNT2, TNNI3, MYH6, MYH7, TPM1, TNNC1, LMOD2 and ACTN2 are sarcomeric; DES and TTN are cytoskeletal/Z-disk; LMNA is nuclear envelope; RBM20 is splicing; BAG3 is a chaperone. Every one of them is a lesion in building or transmitting force. *PLN* is the only calcium-handling gene in the set, and its lesion is in the relaxation arm of excitation–contraction coupling. Folding it into the parent would put the one mechanistically distinct member inside a group defined by a mechanism it does not share.

There is also a management difference that a parent entry cannot carry: R14del carriers are risk-stratified with a variant-specific model (ejection fraction, 24-hour ectopic count, negative T-wave count, low QRS voltage; C-statistic 0.83 against 0.68 for the generic PLN model), because malignant arrhythmia happens at preserved ejection fraction. That is not a detail of the parent disease — it is a different clinical pathway.

I'd suggest the numbered-series risk class is a useful flag for *review* but should not by itself recommend lumping, since it keys on nomenclature rather than on whether the mechanism is shared.

## Two wrong ontology terms in the deep-research report

Worth flagging because the report's own term validation reported 28/28 terms resolved and clean. Both of these CURIEs do exist — they just name different concepts:

| Report claimed | The CURIE actually is | Correct term |
|---|---|---|
| `HP:0001738` "Left ventricular systolic dysfunction" | *Exocrine pancreatic insufficiency* | `HP:0025169` |
| `HP:0031540` "Low-voltage electrocardiogram" | *Linear IgG deposits along the epidermal basement membrane zone* | `HP:0025077` Decreased QRS voltage |

Neither was used. This is the failure mode where identifier resolution passes and the binding is still wrong, so the report's clean validation section is not evidence the terms are right.

## Unresolved mechanism, recorded rather than decided

Whether the primary lesion in R14del is altered SERCA2a kinetics or sarco/endoplasmic reticulum malformation is genuinely contested — recent work disputes the direction of the SERCA effect, and the hallmark "aggregates" turn out to be clustered membrane rather than protein aggregate. Both are curated as `mechanistic_hypotheses` (`pln_serca_superinhibition`, CANONICAL; `pln_ser_malformation`, EMERGING) with the competing causal edges tagged to their group, instead of picking a winner.

Also recorded: a `HUMAN_MODEL_MISMATCH` for the PLN-null mouse. Knocking out phospholamban is benign, even beneficial, in mice, but the human null is lethal. That inversion matters because the leading experimental therapies work by lowering phospholamban, and the mouse cannot tell you the safety margin.

## What is deliberately not in the entry

- Experimental *PLN*-directed therapies (antisense silencing, AAV9-CRISPR, autophagy/UPR modulation) — all preclinical, so they are described in `notes` rather than listed as treatments.
- A phase 1 AZD4063 trial the report mentioned — registry status could not be confirmed, so it is not curated.
- A canine PLN p.Arg9His model the report mentioned — different allele, and no citable record verified.

## Validation

All run to completion on the entry:

```
just validate                  ✅ All validations passed
just validate-terms            ✅ Validation passed
just check-entity-refs         OK: entity references resolve
just check-causal-targets      OK: no new broken pathograph targets
just check-duplicate-keys      OK: no duplicate mapping keys
just check-qualifier-terms     OK
just check-enum-values         OK: no out-of-enum values
just validate-disorders        ✅ All 1 disorder file(s) passed validation
                               Snippets checked: 73/73 verified
```

Whole-KB baseline checks (`check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-environmental-evidence`) all report no new defects; no baseline was modified.

Every one of the 73 evidence snippets is an exact substring of a cached reference. The falcon report carried DOIs only and no PMIDs, so all 12 literature references were resolved to PMIDs and re-fetched before use.

The stub `stubs/Dilated_Cardiomyopathy_1P.yaml` is deleted in this PR.

### DIFF
--- a/history/disorders/Dilated_Cardiomyopathy_1P/2026-09-04T163647Z-claude-code-5e34d5.yaml
+++ b/history/disorders/Dilated_Cardiomyopathy_1P/2026-09-04T163647Z-claude-code-5e34d5.yaml
@@ -1,0 +1,41 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Dilated_Cardiomyopathy_1P
+  path: kb/disorders/Dilated_Cardiomyopathy_1P.yaml
+session:
+  id: 2026-09-04T163647Z-claude-code-5e34d5
+  timestamp: '2026-09-04T16:36:47Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-fable-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10872
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - experimental_models
+  - clinical_trials
+  - evidence
+  summary: 'Address re-review on PR #10872'
+  details: |
+    Closed the three IMPORTANT findings from the 2026-09-04 re-review.
+
+    Cell type: the hiPSC-derived cardiomyocyte in experimental_models was bound to CL:0002322 (embryonic stem cell), wrong on both differentiation state and origin. Rebound to CL:0000746 (cardiac muscle cell), which this file already uses for four other nodes.
+
+    Reference titles: added the 16 missing reference_title values on evidence items in inheritance, progression, diagnosis, clinical_trials, experimental_models and discussions. Thirteen were copied from the same PMID's existing title elsewhere in the file; the three clinicaltrials items took their registry titles from the cache frontmatter. No title was retyped.
+
+    Non-supporting quotes: the Syncope and Exercise intolerance phenotypes cited snippets that did not bear on their claims. Confirmed the reviewer's finding independently - syncope appears in none of the 13 references this entry cites, and a PubMed query for phospholamban AND syncope returns zero records, so no abstract-level source exists. Found the phenotype in PMC full text instead and fetched PMID:30806910 (Cheung et al., Canadian PLN R14del cohort), which records syncope during physical activity followed by fatal ventricular tachycardia storm in a carrier - the arrhythmic mechanism the phenotype attributes it to. Exercise intolerance now cites PMID:33020536, already used five times here, for a decade of progressive activity-limited symptoms in the proband, replacing a quote about ACM prevalence. Both kept directness: INDIRECT with explanations naming the specific inference step.
+
+    Also fixed while in the file: one PMID:33020536 evidence item was graded evidence_source: OTHER while five others in the same file graded the same publication HUMAN_CLINICAL. evidence_source classifies the publication, not the quote, so this is now HUMAN_CLINICAL throughout. check-snippet-grading cannot catch this because it keys on the quoted sentence and these items quote different sentences.
+
+    Suggestions taken: moved the entity-reference bookkeeping sentence out of the head of the pln_null_mouse_human_divergence rationale to the end as a scope note, so the rationale opens with the scientific argument; and curated NCT07241104 (AZD4063 first-in-human) after confirming against the ClinicalTrials.gov API that it is recruiting, phase 1, estimated enrolment 23 - not the 31 the deep-research report gave. It stays out of treatments because the registration does not disclose the agent's mechanism, so there is no node for it to target. The notes paragraph that deferred it was rewritten to record the confirmation.
+
+    Validation: just validate, validate-terms, check-entity-refs, check-causal-targets, check-duplicate-keys, check-enum-values, check-qualifier-terms and validate-disorders all pass. Snippets 86 to 87, all verified. The five whole-KB baseline checks report no new defects and no baseline was modified.
--- a/kb/disorders/Dilated_Cardiomyopathy_1P.yaml
+++ b/kb/disorders/Dilated_Cardiomyopathy_1P.yaml
@@ -85,6 +85,7 @@ inheritance:
       Documents the exception to dominant inheritance: the truncating L39stop allele
       produces the dilated phenotype only in homozygotes.
   - reference: PMID:37144056
+    reference_title: "Phospholamban R14del disease: The past, the present and the future."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: "Disease penetrance is incomplete and age-dependent, with symptoms developing more often within the fifth decade"
@@ -97,6 +98,7 @@ progression:
   age_range: Second to fourth decade
   evidence:
   - reference: PMID:33020536
+    reference_title: "The phenotypic characteristic observed by cardiac magnetic resonance in a PLN-R14del family."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: "The former one can also present in younger PLN-R14del carriers despite completely normal LV structure and function."
@@ -111,6 +113,7 @@ progression:
   age_range: Fifth decade onward
   evidence:
   - reference: PMID:37144056
+    reference_title: "Phospholamban R14del disease: The past, the present and the future."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: "Disease penetrance is incomplete and age-dependent, with symptoms developing more often within the fifth decade"
@@ -118,6 +121,7 @@ progression:
 - phase: Arrhythmic risk period
   evidence:
   - reference: PMID:34113975
+    reference_title: "Prediction of ventricular arrhythmia in phospholamban p.Arg14del mutation carriers-reaching the frontiers of individual risk prediction."
     supports: SUPPORT
     evidence_source: HUMAN_CLINICAL
     snippet: "While the event rate of incident malignant VA is significant in individuals with the PLN p.Arg14del variant, over 90% remain event-free at 5 years."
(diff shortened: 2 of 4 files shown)


AO33:
Thanks — all seven points are addressed. The blocker and both important findings are closed with references that were already in `references_cache/`, so nothing new had to be fetched.

**GeneReviews (blocking).** `PMID:20301486` is now tagged in `references:`, copying the `Dilated_Cardiomyopathy_1E` form exactly. I confirmed the cache record is `content_type: abstract_only` and added the matching scope paragraph to `notes` — no snippet is quoted from it.

**Diagnosis section.** Added cardiac magnetic resonance (`NCIT:C137915`, Magnetic Resonance Imaging of the Heart) and PLN genetic testing with cascade screening (`NCIT:C15709`, Genetic Testing). The CMR entry quotes the sub-epicardial lateral-wall enhancement pattern and, separately, its presence in younger carriers with normal ventricular structure and function — that second quote is what supports the early-detection claim the entry was already making.

**hiPSC-CM model.** Added an `experimental_models` record for the isogenic pair with three `modeled_mechanisms` links rather than one: `RECAPITULATES` the contractile-dysfunction node, `MEASURES` the UPR node, and `RESCUES` on the BiP inducer X arm. Each carries its own readout with direction and its own evidence, so the rescue is not conflated with the phenotype. The UPR link cites the human myocardium confirmation as its link-level evidence, since that is what makes the model informative rather than an in-vitro artifact.

**Progression** is now a three-phase block. The arrhythmic-risk phase carries the "over 90% remain event-free at 5 years" figure and a note that it runs in parallel with the others rather than after them — sudden death can precede pump failure, which is the reason the risk model does not wait on ejection fraction.

**One suggestion I did not take.** The ~70% of carriers with a major cardiac event by age 70 is not in any reference cached for this entry. I searched all of them; the closest verifiable statement is that penetrance is incomplete and age-dependent with symptoms usually in the fifth decade, which is what I curated instead. The report's figure and the reason for leaving it out are recorded in `notes`. I would rather leave the quantitative claim uncurated than cite it to a paper that does not contain it.

The mitochondrial arm omission is now signalled in `notes` as you suggested, and the `animal_models#Mouse` ambiguity is called out in the discussion prose — the reference resolves on `species`, so it genuinely matches both models, and the text now says the null mouse is the one meant.

Snippets went from 73 to 86, all verified. `just validate-disorders`, `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms` and `check-enum-values` all pass, and the five whole-KB baseline checks report no new defects with no baseline modified.

AO33:
All three findings are closed, plus both suggestions.

**The cell type was wrong, and you were right about why.** `CL:0002322` is *embryonic stem cell* — undifferentiated and of embryonic origin, so wrong on both counts for a differentiated cardiomyocyte made from induced pluripotent cells. Rebound to `CL:0000746` *cardiac muscle cell*, matching the four other uses in this file. CURIE and label only; `preferred_term` unchanged.

**Sixteen reference titles added.** Thirteen came from the same PMID's existing title elsewhere in this file, three from the trial cache frontmatter. Nothing was retyped by hand — a script copied each one from an existing verified value.

**The two bad quotes are replaced, and I checked your finding independently before acting on it.** Syncope really does appear in none of the thirteen references this entry cites. It is worse than that: a PubMed search for `phospholamban AND syncope` returns **zero** records, so no abstract-level source exists anywhere. The phenotype is real, though, and it is in PMC full text. I fetched PMID:30806910 (Cheung et al., Canadian PLN R14del cohort), which reports syncope during physical activity followed by death from ventricular tachycardia storm in a carrier — the arrhythmic mechanism this phenotype attributes it to.

Exercise intolerance now cites PMID:33020536, already used five times here, for the proband's decade of progressive activity-limited symptoms. That replaces a quote about ACM prevalence in healthy young adults, which had no inference path to the claim at all. Both stay `INDIRECT`, but the explanations now name the actual inference step rather than apologising for the quote.

**One thing you marked as correct that I changed anyway.** Your checklist passed `evidence_source`, but PMID:33020536 was graded `OTHER` on one item and `HUMAN_CLINICAL` on five others in the same file. It is a human family study, so `HUMAN_CLINICAL` is right and the grade cannot depend on which sentence is quoted. Now consistent. Worth noting `check-snippet-grading` cannot catch this class of defect — it keys on the quoted sentence, and these items quote different sentences.

**Both suggestions taken.** The `animal_models#Mouse` bookkeeping sentence moved from the head of the rationale to the end as a scope note, so the argument leads. And NCT07241104 is now curated: I checked it against the ClinicalTrials.gov API rather than the report, and it is recruiting, phase 1, estimated enrolment **23** — not the 31 the deep-research report gave. It stays out of `treatments` because the registration does not disclose AZD4063's mechanism, so there is no node for it to target. The `notes` paragraph that deferred it now records the confirmation instead.

Snippets 86 to 87, all verified. `validate-disorders`, `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values` and `check-qualifier-terms` pass. The five whole-KB baseline checks report no new defects and no baseline was modified.

PR #10874, opened 2026-09-04
feat(Syndromic_X-linked_Intellectual_Disability_94): curate GRIA3 AMPA-receptor disorder with branched GoF/LoF pathograph

### LINKED ISSUE #10857: Curate syndromic X-linked intellectual disability 94 (MONDO:0010402)
Curating **syndromic X-linked intellectual disability 94** as a standalone entry in `kb/disorders/`.

| | |
|---|---|
| MONDO | `MONDO:0010402` |
| Causal gene | **GRIA3** (`hgnc:4573`) |
| MONDO parent | X-linked syndromic intellectual disability (MONDO:0020119) |
| Stub | `stubs/Syndromic_X-linked_Intellectual_Disability_94.yaml` |

GRIA3 encodes the GluA3 subunit of the AMPA receptor, and both gain- and loss-of-function variants have been reported with different phenotypes. That is a specific, testable synaptic mechanism, which is exactly the kind of content a generic X-linked intellectual disability parent term cannot carry.

- Nothing in `kb/disorders/`, `kb/groupings/` or `kb/modules/` on `origin/main` binds `MONDO:0010402`, and a search on the disease label and the gene name found no entry covering it either.
- No other issue, open or closed, references `MONDO:0010402` or this disease, and no open pull request does.
- There is no GRIA3 entry anywhere in the knowledge base.

AO33:
Adds a knowledge base entry for syndromic X-linked intellectual disability 94 (MRXSW, Wu type; `MONDO:0010402`), caused by variants in **GRIA3**, the gene for the GluA3 subunit of the AMPA glutamate receptor. GRIA3 does not appear anywhere else in `kb/`, so this is the knowledge base's first coverage of AMPA-receptor synaptic disease.

The thing worth knowing about this disorder is that it is **not one mechanism**. GRIA3 variants split into loss-of-function and gain-of-function classes that produce close to opposite clinical pictures, and the split has a treatment consequence. The entry models that as a branch rather than averaging it away.

## The branch

```mermaid
graph LR
  V["GRIA3 Pathogenic Variant"] --> LOF["GluA3 Loss of Function"]
  V --> GOF["GluA3 Gain of Function"]
  LOF --> RED["Reduced AMPAR current"]
  GOF --> PRO["Prolonged AMPAR current"]
  RED --> HYPO["Hypotonia"]
  RED --> SLEEP["Sleep disturbance"]
  PRO --> HYPER["Hypertonia"]
  PRO --> STARTLE["Exaggerated startle response"]
  RED --> SYN["Disturbed synaptic transmission"]
  PRO --> SYN
  SYN --> GDD["Global developmental delay"]
  SYN --> EXC["Cortical network hyperexcitability"]
  EXC --> SZ["Seizure"]
```

In the largest functionally characterised cohort (25 patients, 23 families, PMID:38038360), every patient had global developmental impairment — that is the convergent output on the right of the diagram. Everything else separates by allele class. Loss of function: hypotonia, sleep disturbance, seizure onset at a median of 16 months. Gain of function: hypertonia, hyperekplexia, seizure onset at a median of **1 month**.

Both classes affect both sexes, but not equally: affected males usually carry hemizygous loss-of-function alleles inherited from healthy carrier mothers, and affected females mostly carry de novo heterozygous gain-of-function alleles. Two `inheritance` blocks record this — a general `X-linked inheritance` block for the mechanism-dependent picture, and an `X-linked recessive inheritance` block for the classic carrier-mother pedigrees, which is accurate for the historical Wu-type families but does not describe de novo female disease.

## The treatment observation, including the case where it failed

One boy with the functionally confirmed gain-of-function allele p.Ala615Val had both seizures and hypertonia improve on carbamazepine, which reduces presynaptic glutamate release (PMID:35031858). That is curated as a treatment with a mechanism link to the prolonged-current node.

It is also curated with a `REFUTE` evidence item on the same treatment, quoting a case where carbamazepine did not work (PMID:36726007). That patient's variant was never functionally tested, so the failure is consistent with the mechanism-matched reading rather than against it — but it is one clinical observation each way, and the entry says so rather than presenting the positive case alone.

## Why this should be its own entry, not folded into a parent

`dashboard/priority.json` scores this term `LUMP_INTO_PARENT`. I think that is wrong here, and the reason it fires is visible in the row itself:

```json
"nec_risk": [{"risk_class": "NUMBERED_SERIES", "detail": ["type 94"]}],
"specificity_bucket": "subtype_series",
"features": {"subtype_series": true, "broad_parent": false,
             "grouping_term": false, "child_count": 0}
```

The recommendation comes from the **label** — the words "type 94" and a family stem shared with six sibling terms. It is not derived from anything biological. The same row records `child_count: 0`, `broad_parent: false` and `grouping_term: false`, so the scorer's own features agree this is a leaf, not a grouping.

Two concrete reasons the parent cannot hold this content:

1. The parent is `X-linked syndromic intellectual disability`, a category with hundreds of unrelated causal genes. It cannot carry a receptor-level gain-of-function versus loss-of-function distinction, because that distinction is a property of GRIA3's gating kinetics and means nothing for the other genes in the category.
2. It cannot carry the genotype-guided drug response either. "Carbamazepine helped this patient because his AMPA receptors desensitise too slowly" is not a statement about X-linked intellectual disability. Lumping deletes it.

The numbered-series heuristic is a reasonable prior for the MRX series in general. It is a bad fit for a term where the number happens to sit on a single well-characterised gene with a mechanistically branched allelic spectrum.

## Notes on the source material

Curated from the Edison/falcon report already in `research/`. Three things about it that are recorded in the entry's `notes` so the same report is not re-mined into the same errors:

- **All 13 citations were DOI-only.** The report says outright that it withheld PMIDs rather than guess them. Every DOI was resolved to a PMID before use — 11 through the PMC ID converter, and one (the carbamazepine paper, PMID:35031858) by PubMed title search after the converter came back empty. No evidence item in this entry sits on a `DOI:` prefix, which the gating reference validator skips.
- **Five proposed ontology terms named the wrong concept or carried a stale label**, and the report's own Term Validation section called all 47 of them clean. The one that matters: it offered `HP:0002169` for "Hyperekplexia". `HP:0002169` is **Clonus**. Hyperekplexia is `HP:0002267` Exaggerated startle response, which is what the entry binds. The other four were label drift — `HP:0001252` is now `Hypotonia`, not "Muscular hypotonia"; `HP:0002123` is `Generalized myoclonic seizure`; `HP:0100022` is `Abnormality of movement`; `HP:0002360` is `Sleep disturbance`.
- **Two of its sources were deliberately not used.** The exon 5-12 deletion prenatal diagnosis case is in a journal with no PubMed or PMC record, so it could not be given a PMID; its content is described in `notes` instead of cited. The brain-region transcriptomics work is an unreviewed preprint modelling a schizophrenia-associated allele rather than this disorder, so it is not curated as a pathophysiology node.

ClinicalTrials.gov was queried directly rather than taken from the report. A GRIA3 term search returns one unrelated study (`NCT03676569`, autologous cell therapy in autoimmune refractory epilepsy); a condition search returns nothing. No `clinical_trials` block is curated.

## Also in the entry

- 7 pathophysiology nodes, 23 phenotypes, 7 curated variants with the two that carry ACMG classifications marked `LIKELY_PATHOGENIC`
- 3 animal models, including a `FAILS_TO_RECAPITULATE` link: the *Gria3* null mouse has **intact** spatial memory on both the Morris water maze and the Y-maze, so it does not reproduce the cognitive phenotype that defines the human disorder. That is paired with a `HUMAN_MODEL_MISMATCH` discussion laying out the three readings the data do not separate.
- 2 in vitro assay models (heterologous patch clamp, minigene splicing), which are the assays the whole loss/gain taxonomy rests on
- 3 discussions: the mouse mismatch above, what determines severity in heterozygous females, and whether AMPAR pharmacology should be selected by the patient's own measured variant direction

## Validation

Every check run to completion, output read:

```
just validate                      All validations passed
just count-verified-snippets       103/103 verified against cached references
just validate-terms                Validation passed
just check-entity-refs             OK: entity references resolve
just check-causal-targets          OK: no new broken pathograph targets
just check-duplicate-keys          OK: no duplicate mapping keys
just check-qualifier-terms         OK
just check-enum-values             OK: no out-of-enum values
just validate-disorders            All 1 disorder file(s) passed validation
```

Whole-KB baselines, no new defects added to any:

```
just check-folded-hyphens          OK (330 grandfathered)
just check-snippet-length          OK (1330 distinct grandfathered)
just check-title-snippets          OK (275 distinct grandfathered)
just check-snippet-grading         OK (329 distinct grandfathered)
just check-environmental-evidence  OK
just check-term-cache-integrity    OK
just check-cache-order             OK
```

Deletes `stubs/Syndromic_X-linked_Intellectual_Disability_94.yaml`.

Closes #10857

### DIFF
--- a/kb/disorders/Syndromic_X-linked_Intellectual_Disability_94.yaml
+++ b/kb/disorders/Syndromic_X-linked_Intellectual_Disability_94.yaml
@@ -536,8 +536,10 @@ pathophysiology:
   description: >-
     Both branches converge here. Whether GluA3-mediated current is reduced or prolonged, the
     postsynaptic response to glutamate is mistimed relative to the activity-dependent rules
-    that govern synaptic strengthening, so experience-dependent plasticity, learning and
-    memory circuits develop abnormally. GluA2/3 receptors cycle constitutively under basal
+    that govern synaptic strengthening, so experience-dependent plasticity is degraded and the
+    circuits that support learning and memory develop abnormally. The cognitive outcome itself
+    is carried by the downstream HP phenotypes rather than by a GO behaviour term on this
+    node. GluA2/3 receptors cycle constitutively under basal
     conditions and their auxiliary partners set trafficking and gating, so the lesion is felt
     at the level of synaptic maintenance as well as of acute transmission.
   biological_processes:
@@ -551,11 +553,6 @@ pathophysiology:
     term:
       id: GO:0048167
       label: regulation of synaptic plasticity
-  - preferred_term: learning or memory
-    modifier: ABNORMAL
-    term:
-      id: GO:0007611
-      label: learning or memory
   cell_types:
   - preferred_term: glutamatergic neuron
     term:
@@ -853,7 +850,7 @@ phenotypes:
   sequelae:
   - target: Focal motor seizure
   - target: Generalized myoclonic seizure
-  - target: Status epilepticus
+  - target: Non-convulsive status epilepticus
   evidence:
   - reference: PMID:38038360
     reference_title: "Gain-of-function and loss-of-function variants in GRIA3 lead to distinct neurodevelopmental phenotypes."
@@ -923,7 +920,7 @@ phenotypes:
       Individual semiology and onset age for the myoclonic seizures in that proband, whose
       familial allele was inherited rather than de novo.
 
-- name: Status epilepticus
+- name: Non-convulsive status epilepticus
   category: Neurologic
   phenotype_term:
     preferred_term: Non-convulsive status epilepticus
@@ -932,11 +929,12 @@ phenotypes:
       label: Status epilepticus without prominent motor symptoms
   frequency: VERY_RARE
   description: >-
-    Both convulsive (myoclonic) and non-convulsive status epilepticus are reported. Status
-    epilepticus of either kind has been documented in only a handful of patients; the
-    non-convulsive form was described first in a male and, in 2025, for the first time in a
-    female — a seven-year-old with a de novo p.Thr657Ala allele who resolved back to her
-    neurological baseline after treatment.
+    Status epilepticus of either kind has been documented in only a handful of patients. This
+    phenotype is scoped to the non-convulsive form, which was described first in a male and, in
+    2025, for the first time in a female — a seven-year-old with a de novo p.Thr657Ala allele
+    who resolved back to her neurological baseline after treatment. The convulsive myoclonic
+    form, which escalated to myoclonic status epilepticus in the Italian pedigree, is carried by
+    the separate Generalized myoclonic seizure entry.
   evidence:
   - reference: PMID:41462794
     reference_title: "Non-Convulsive Status Epilepticus and Mild Neurodevelopmental Phenotype in a Female with a Novel p.Thr657Ala Variant in the GRIA3 Gene."
@@ -1161,6 +1159,27 @@ phenotypes:
     explanation: >-
       Names autistic features explicitly, in the GRIA3 pedigree where they were reported,
       together with the co-occurring behavioural problems.
+
+- name: Self-injurious behavior
+  category: Behavioral
+  phenotype_term:
+    preferred_term: Self injury
+    term:
+      id: HP:0100716
+      label: Self-injurious behavior
+  frequency: OCCASIONAL
+  description: >-
+    Self-injury is reported alongside aggressive outbursts in the affected adult males of the
+    Finnish family D174, and is one of the behavioural features that make the adult
+    presentation of this disorder difficult to manage.
+  evidence:
+  - reference: PMID:24721225
+    reference_title: "X-exome sequencing in Finnish families with intellectual disability--four novel mutations and two novel syndromic phenotypes."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "are characterized by severe ID with autistic features, epilepsy, short stature and behavioral problems such as self injury and aggressive outbursts"
+    explanation: >-
+      Names self injury explicitly among the behavioural problems of the GRIA3 pedigree.
 
 - name: Aggressive behavior
   category: Behavioral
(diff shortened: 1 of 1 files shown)


AO33:
Heads-up on the red check: `claude-review` failed before it read anything. The job log says \`You's hit your limit · resets 4:40am (UTC)\`, `num_turns=1`, `cost=$0` — the review agent hit a usage limit, errored on its first turn, and posted no review. Nothing was reviewed and nothing was found; this is not a finding about the PR.

The required `test (3.13)` check passed.

I will re-run the review workflow once the limit resets. If it errors the same way again, the underlying cause is worth looking at separately: the workflow prefers `ANTHROPIC_API_KEY` and only falls back to `CLAUDE_CODE_OAUTH_TOKEN`, so a weekly-limit error here suggests the API key secret is unset and the run used the subscription token.

AO33:
Pushed a self-review pass (`621b8b8`) before the automated reviewer runs. I re-read the entry against the cached full texts rather than against the deep-research report, and found real problems. Summarising them here so a reviewer does not have to rediscover them.

**Two evidence defects, both caught by reading the source rather than the snippet:**

- The `Autistic behavior` phenotype cited a snippet that never mentions autism. It quoted PMID:42666715 on "significant psychiatric symptoms and spatial memory deficits" — true, cited, exact, and not about autism. The claim is real but the source was wrong: PMID:24721225 describes the Finnish family D174, three affected males aged 35–57 carrying GRIA3 p.Gly630Arg, "characterized by severe ID with autistic features, epilepsy, short stature and behavioral problems such as self injury and aggressive outbursts". Swapped.
- The `Status epilepticus` description said the non-convulsive form "has so far been described once". PMID:41462794 says the opposite two paragraphs above the sentence I quoted: "Status epilepticus (SE) has been documented in two cases." Its own claim is scoped — first NCSE case **in a female** — and I had dropped the qualifier. Both facts are now recorded.

**Three ontology bindings were one level too general.** `HP:0006979` is literally *Sleep-wake cycle disturbance*, which is what I had written as `preferred_term` while binding the parent. `HP:0031475` *Status epilepticus without prominent motor symptoms* carries "Nonconvulsive status epilepticus" as an exact synonym — and I had written a justification sentence claiming HPO had no such term, which was simply wrong and is now deleted. `HP:0012389` *Appendicular hypotonia* is defined as hypotonia of one or more limbs, and the cohort figure I cite is specifically "**Limb** muscular hypotonia was reported in 13/25".

**Two frequency bands used the wrong denominator.** `FrequencyEnum` is a fraction of patients. Focal motor seizure is 6 of 25 (24%) and generalized myoclonic is 5 of 25 (20%) — both `OCCASIONAL`. I had graded them `FREQUENT` by computing over the 12 patients with epilepsy. Global developmental delay also dropped from `OBLIGATE` to `VERY_FREQUENT`: the cohort was selected for having a functionally characterised variant, so 25/25 is not a penetrance estimate.

**Prose that ran ahead of its source.** The disease description attributed hyperreflexia to the 25-patient cohort — the string does not occur in PMID:38038360 at all, and hyperreflexia comes only from the single p.Ala615Val male. Both `genetic_context` blocks asserted one `zygosity` value that excluded real cases (the hemizygous gain-of-function male, affected heterozygous females); zygosity is now unset with both stated in the description, because the slot is single-valued and cannot represent the branch. The gain-of-function node listed p.Met661Thr and p.Thr657Ala as exemplars without noting that neither was tested — both source papers say so outright. And p.Gly630Arg is no longer assigned to the loss-of-function branch, since it has no electrophysiology; assigning it was exactly the inference this entry's own thesis warns against.

**One thing found that improves the argument.** Reading the cached full texts turned up the mirror image of the carbamazepine result, which is not in the deep-research report:

> a targeted trial with a selective non-competitive AMPA receptor antagonist (Perampanel) proved to be ineffective, **as expected in case of loss of function of the receptor**

That is an AMPA receptor antagonist failing in documented loss-of-function disease, with the authors saying the failure was predicted by the mechanism. Paired with carbamazepine helping in confirmed gain-of-function disease, it is the strongest published support for treating allele direction as therapeutically decisive, and it is now cited in the `mechanism_matched_ampar_pharmacology` discussion. PMID:41462794 reports the same failure independently.

**Also added:** `reference_title` on all 103 evidence items (2543 of 2545 disorder entries carry it; this was one of two that did not). The direct E787G no-current measurement on the loss-of-function edge. An `Intellectual disability` phenotype — the adult pedigrees are 35 to 57 years old, past the age where `Global developmental delay` is the right term. `Hyporeflexia` as the loss-of-function mirror of hyperreflexia.

**Pathograph:** unconnected phenotypes went from 11 to 4. The four that remain — cerebellar vermis hypoplasia, cerebral atrophy, ventriculomegaly, short stature — have no sourced mechanistic route from the receptor lesion, so they are left unconnected rather than given an invented edge.

Snippets are now 112/112 verified. Every gate and whole-KB baseline re-run clean.

AO33:
Both blocking items fixed in `379621a`, plus all five suggestions.

**Item 1 — you were right, and the entry was contradicting itself.** Both sources describe one failure. PMID:34731330 attributes it to "the male patient reported by **Trivisano et al.**"; PMID:41462794 points back at "a previous report". One event, two citations. My rationale said two patients while my own explanation on the Rubino item said "the same perampanel failure" — those could not both stand, and the rationale was the wrong one.

The subtler half was the better catch. "Documented loss-of-function disease" put an electrophysiological measurement on the treated patient, but the measurement in PMID:34731330 is of the citing authors' own p.Glu787Gly allele. Their "as expected in case of loss of function" is an inference across a shared residue. The discussion now says one reported failure, names whose patient it was, and states that the loss-of-function characterisation belongs to a different allele. The Rubino explanation says a later group read the same result the same way, which corroborates the interpretation and establishes nothing about generality. Same correction in the notes block.

You are also right that this does not weaken the argument. It sharpens it: a same-residue inference that the drug should fail, followed by the drug failing, is a cleaner statement of the mechanism-matched case than a miscounted "two patients" would have been.

**Item 2 — added a four-entry `diagnosis` section.** Trio exome or genome sequencing first-line, and the entry now says explicitly why it carries the whole diagnosis: there is no biochemical marker, no enzyme assay and no imaging finding, so the diagnosis rests entirely on the variant, and the trio design is what establishes de novo status.

Copy-number-sensitive testing is modelled as its own step rather than a fallback, for the reason you gave — exome SNV analysis misses the multi-exon deletions. As you predicted, the cleanest source is the BJSTR paper with no PubMed record, so that case sits in `notes` and the citable evidence is the weaker form: microarray is performed and comes back normal in the point-variant cases, which is itself the useful clinical point that a normal microarray does not exclude the diagnosis.

The two steps you flagged as tying to the therapeutic argument are both in, from references already in the entry. The minigene step quotes the ACMG reclassification sentence from PMID:42666715 rather than the assay result alone, so it records the classification change the assay produced. The electrophysiology step quotes the 31-of-44 figure from PMID:38038360, because the 13 functionally neutral alleles are exactly why the step is not optional. That entry also states that patch clamp is a research or advanced-diagnostics assay rather than a routine clinical test — which is a real limit on the mechanism-matched argument and belongs next to it.

**Suggestions, all taken:**

- Renamed the phenotype to `Non-convulsive status epilepticus` and updated the `sequelae` target. Name, binding and both snippets now agree, and the description points at `Generalized myoclonic seizure` for the convulsive form.
- Dropped `GO:0007611 learning or memory`. Correct that it sits in GO's behaviour branch and was doing nothing beside `GO:0048167`; the cognitive outcome is carried by the downstream HP phenotypes.
- Added `Self-injurious behavior` (HP:0100716) and `Lorazepam for Sleep Disturbance`.

**The two checks you could not run without network access, now run:**

- **GeneReviews:** no chapter. `GRIA3[TI] GeneReviews[TI]` and `GRIA3 AND GeneReviews[Book]` both return zero. Your recollection was right — GeneReviews covers GRIN-related disorders, not GRIA3. Nothing to tag, and that is now recorded in `notes` so it is not re-checked.
- **The "syndromic X-linked intellectual disability 29" synonym:** genuine, and on this term. MONDO:0010402 carries it along with `MRXS29` and "mental retardation, X-linked, syndromic 29". The Wu-type entity was numbered 29 in the older syndromic X-linked series before being renumbered 94, which is why a 29 and a 94 synonym sit on one concept. Also recorded in `notes`.

Entry is now 7 pathophysiology nodes, 27 phenotypes, 4 diagnosis entries, 6 treatments, 3 animal models, 2 experimental models, 3 discussions. Snippets 120/120 verified; every gate and whole-KB baseline re-run clean.

PR #10876, opened 2026-09-04
Mine the BACH2/BRIDA full text into Immunodeficiency 60, and fix the bug that hid it

### LINKED ISSUE #10867: Mine the full text of PMID:28530713 to quantify the Immunodeficiency 60 humoral phenotype
`kb/disorders/Immunodeficiency_60.yaml` asserts immunoglobulin deficiency but cannot quantify it, because only the **abstract** of the founding report was retrievable when the entry was curated. Unpaywall and OpenAlex were both blocked, so `references_cache/PMID_28530713.md` carries `content_type: abstract_only`.

The entry states this limit rather than hiding it, and `discussions[bach2_imd60_humoral_phenotype_uncurated]` names exactly what to fill. This issue is that discussion made actionable.

## What the abstract gives, and what it withholds

The one sentence the entry rests on for the humoral arm:

> "Affected subjects had lymphocyte-maturation defects that caused immunoglobulin deficiency and intestinal inflammation."

That supports the phenotype and the causal edge from the memory B-cell node. It does not say which isotypes fall, by how much, whether the pattern is uniform across the affected subjects, or whether it meets criteria for common variable immunodeficiency. The two reported families differ on exactly this point — the second family's proband had **selective IgA deficiency** ([PMID:37148421](https://pubmed.ncbi.nlm.nih.gov/37148421/)), not a described pan-hypogammaglobulinemia — so the entry cannot currently say whether BRIDA produces a consistent humoral picture or a variant-dependent one.

## To do, once the full text is reachable

```bash
just fetch-reference PMID:28530713   # should come back full_text_xml, not abstract_only
```

Then mine it for:

- **Isotype-level immunoglobulin values** in the affected subjects, which would let the `Immunoglobulin deficiency` phenotype carry a `biochemical` block with reference ranges instead of a bare HP binding.
- **Patient class-switch recombination and plasmablast assay results.** The entry's memory B-cell node currently binds `GO:0002319 memory B cell differentiation` rather than `GO:0045190 isotype switching`, deliberately, because nothing available evidences decreased class switching *in patients* — only in the mouse null. Full text would settle whether the isotype-switching annotation belongs.
- **Per-compartment lymphocyte counts**, which would let the `Bach2-heterozygous mouse (BRIDA report)` link say which compartment the "analogous lymphocyte defects" were in, and would likely justify raising its `fidelity` from `MODERATE`.
- **The mouse line's provenance**, which the abstract does not state. The entry currently carries a note warning that the BRIDA heterozygote and the Igarashi heterozygote may be the same line and must not be counted as independent observations. Full text would resolve that either way.
- **Individual subject phenotypes.** MeSH indexing lists `Splenomegaly`, `Pancytopenia`, `Fever` and `Lymphopenia`, none of which appear in the abstract. MeSH terms are indexing, not quotable findings, so none of them is curated — the full text would say whether they belong as phenotypes.

## Also worth revisiting

Immunoglobulin replacement is absent from `treatments` because no source consulted states it for BRIDA specifically. Clinically it is near-certain; the entry records the absence as a gap rather than a negative finding. The full text may supply it directly.

Context: [PR #10813](https://github.com/monarch-initiative/dismech/pull/10813), which curated the entry. The related schema gap is #10865.

gingin77:
Issue #10867 asked for the founding BRIDA paper (PMID:28530713) to be mined "once the full text is reachable". It turned out to be reachable already. The reference cache said `abstract_only` because of a bug in our own code, not because the publisher blocks the article — so this PR fixes the bug, refetches the paper, and rebuilds the entry on what it actually says.

The headline biology: **the humoral phenotype of BRIDA is not uniform, and it does not follow the variant.** A father and daughter carrying the same E788K allele differ from low-everything to low IgA against *raised* IgM and IgG. A normal or high IgG does not exclude the diagnosis.

## Why the full text was unreachable

`linkml-reference-validator`'s XML extractor throws a paper away if the word "restricted" appears anywhere in it, and only then looks for the article body. The BRIDA paper says "IgM-restricted plasma cells" in its results and "Searches were restricted to the period from ..." in its methods. Fifty thousand characters of full text were discarded over two sentences of ordinary English.

We already had a patch for a narrower version of this — it recovered papers carrying a `<restricted-by>` metadata tag — but the plain-English case fell straight through it.

To fix it properly I asked PMC what a genuinely unavailable paper looks like. It returns the front matter and **no `<body>` element at all**; the phrase the extractor keys on sits in that front matter. So the right test is structural — does this document have a body with paragraphs in it — and the word match is noise. The patch now uses that test. A record with no body still returns nothing and stays unavailable, and the two existing tests for that still pass.

Two more affected papers turned up while checking this, so it is not a one-off.

## Second fix: tables

The same extractor keeps `<body>` paragraphs only, so every table in every paper was dropped. In a clinical report that is where the per-patient phenotype lives. Table 1 of this paper is the only place it records that two of its three subjects were on immunoglobulin replacement, and the only place the third subject's raised IgM and IgG appear.

Tables are now appended to the cached reference as pipe-delimited rows — the same shape the ORPHA and ICEES structured caches already use, so a curator quotes one row the way they quote an ORPHA row:

```yaml
snippet: "| On IvIg treatment | Yes | Yes | No |"
```

Tables are found across the whole document, since NIHMS-converted papers put them in a trailing `<floats-group>` outside the body. Anything over 200 rows is skipped as a data dump. Both fixes are covered by new tests.

## What the paper said, against the issue's checklist

| The issue asked for | Result |
|---|---|
| Isotype-level pattern | **Answered** — and it varies between patients, including within one family |
| Isotype-level *values* | **Not obtainable.** They are in Supplementary Table 1, which PMC does not serve |
| Patient class-switch and plasmablast assays | **Answered** — `GO:0045190 isotype switching` now annotated, as the issue anticipated |
| Per-compartment lymphocyte counts | **Answered** — mouse link raised to `HIGH` fidelity with five readouts |
| Mouse line provenance | **Answered** — same line as PMID:23728300, so the do-not-double-count warning stands |
| MeSH-listed features (splenomegaly, pancytopenia, fever, lymphopenia) | **They belong** — all now curated with their denominators |
| Immunoglobulin replacement | **Gap closed** — Table 1 records IVIg in the two pan-deficient subjects |

On the one that failed: the issue hoped isotype values would let the phenotype carry a `biochemical` block with reference ranges. They will not. Table 1 gives only Low, Normal or High per isotype, and the concentrations sit in a supplementary file PMC does not serve. The entry now records a direction per isotype and says explicitly that it carries no values and no reference ranges. The humoral knowledge-gap discussion is narrowed to that one remaining question rather than closed.

## Changes to the entry

Ten new phenotypes, each carrying its denominator in the description — decreased IgA (the only isotype low in all three subjects), decreased memory and class-switched memory B cells, failure to respond to vaccination, reduced T-cell proliferation, lymphadenopathy, splenomegaly, pancytopenia, non-infectious fever, bronchiectasis.

Three additions the full text made possible that the issue did not ask for:

- **A new pathophysiology node, "Gut-Homing TH1 Effector Skewing"**, between the regulatory T-cell defect and the colitis. It answers a question the entry could not previously answer — why the tissue infiltration lands in the gut. Patient CD4+ T cells show raised T-bet with raised CCR9 and beta-7 integrin, and the heterozygous mouse shows the same paired shift. Its gut-addressing half is deliberately left unbound: GO has no term for homing to intestinal tissue, and the paper reports receptor expression rather than migration, so binding `T cell migration` would assert more than the source says.
- **A new `experimental_models` section** for the ~50% BACH2 RNAi knockdown in healthy human T and B cells. This is the entry's only dose-matched *human* system and the strongest causal evidence in it — every other model is either a mouse at the right dosage or a patient of the right species.
- **A colonic-biopsy histopathology finding.** The biopsy's Treg population is reduced against healthy controls *and* against ordinary inflammatory bowel disease, which is what makes it mechanistically informative rather than merely confirmatory.

Two model links were also added to the heterozygous mouse, including an `UNCHANGED` readout recording that bulk lymphocyte counts are normal in unchallenged animals. That is a real negative and it bounds the model.

## Validation

Schema clean, 94/94 snippets verified, terms pass, and the entity-ref, causal-target, duplicate-key, qualifier-term, enum-value, folded-hyphen, snippet-length, title-snippet and snippet-grading gates are all clean. `just validate-disorders` passes.

Two unrelated cache files (`PMID_15152264`, `PMID_37148421`) gained a `full_text_attempted: true` line — the validator retried full text for them under the fixed patch, did not find any, and recorded that. Generated, not hand-edited.

## Base

Stacked on #10813, which created this entry and is approved but not yet merged. GitHub will retarget this to `main` when that lands.

Closes #10867

https://claude.ai/code/session_01SxMairBXnhqHxLNZjaqwZ1


gingin77:
How far the extractor bug reaches, since the scope of it is a fair question to ask of this PR.

I sampled 200 `abstract_only` files at random from `references_cache/` and asked PMC about each. Two of them are in PMC at all; both of those have a full article body, both were blocked by the word match, and both carry a table. Two earlier spot checks behaved identically, so it is 4 for 4 across everything I looked at.

The 100% rate follows from how such a file gets created rather than being a coincidence: a paper with a retrievable PMC body should have cached as `full_text_xml`, so if it is `abstract_only` instead, something stopped the fetcher — and the word match was that something.

At the 1% PMC-linked rate in the sample, roughly 280 of the 27,939 `abstract_only` files have a full text sitting behind the old bug. That is an extrapolation from 200 files, not a census.

I have **not** refetched any of them here. A bulk rewrite of `references_cache/` would turn every open curation PR red on merge, which is the failure mode `CLAUDE.md` already documents for duplicate keys and retired enum values. Filed as #10878 so the policy is a decision rather than a side effect of this PR.

gingin77:
`test (3.13)` failed on the curation-scope check, which is correct — this PR touches `kb/` and also `src/dismech/patch_reference_validator.py` and its tests. I have added `scope-override`, which is what the check's own error message asks for, and here is the reasoning so it is reviewable rather than just waved through.

The two halves *could* be split, and I want to be straight that they could. `references_cache/PMID_28530713.md` is committed content, and validation reads the cache rather than refetching, so a curation-only PR would pass on its own.

I did not split them for two reasons.

The cache file cannot be regenerated without the code fix. Anyone running `just fetch-reference PMID:28530713` against a tree with the old extractor gets `abstract_only` back and a 50k-character diff. Shipping the regenerated cache in one PR and the code that produces it in another means that, between the two merges, the repository cannot reproduce its own cache file — which is the property `just structured-rebuild-orphanet` and the manifest pinning exist to protect.

And the split costs a three-deep stack. This PR is already stacked on #10813. A separate infra PR underneath it would make the curation PR third in line, for no reviewer benefit — the bug and what the recovered text says are one causal story and would be read together anyway.

If you would rather see them separated, say so and I will split it: an infra PR carrying the extractor fix, its tests and the regenerated cache, and a curation PR carrying `kb/` and `history/` on top. It is a rebase, not a rewrite.

Separately, `claude-review` errored rather than reviewing — the agent hit a usage limit that resets at 04:40 UTC. I will re-run it after that.

gingin77:
All three 🟡 items and all five 🔵 suggestions are addressed in 357243b. Two of them I did slightly differently from what you proposed, and I want to flag those rather than let them pass as agreement.

**The stale sentence.** Fixed. It now records that the report is cached as full text including Table 1, and that the concentrations are in Supplementary Table 1 and are not retrievable. You are right that nothing would have caught it — `check-source-defect-claims` does not match that phrasing.

**HP:0031382.** Good catch, and the reason it is available is exactly your point: the methods sentence naming anti-CD3 and anti-CD28 only exists in the cache because of this PR. Rebound, with `name` and `preferred_term` following the term, and the methods sentence cited on the phenotype so the choice is auditable rather than asserted.

Worth noting the methods carry *two* anti-CD3/28 proliferation protocols — one for nucleofected cells in the RNAi arm, one for patient T cells. I quoted the patient one.

**Pathograph connectivity.** The graph was in four components; it is now one, 30 nodes and 36 edges, no orphan targets. Beyond the four edges you listed I also connected the four immune-dysregulation phenotypes from the Treg node (typed `INDIRECT_UNKNOWN_INTERMEDIATES`, since the source reports them without a mechanism), and gave both corticosteroid treatments `target_mechanisms` links to the features the source says responded.

Here is where I deviated. You suggested `Immunoglobulin deficiency` → `Decreased circulating IgA concentration` and → `Decreased specific antibody response to vaccination`. I routed both from `Impaired Memory B Cell Development` instead. Low IgA is a *component* of immunoglobulin deficiency rather than a consequence of it, and vaccine-response failure occurs with normal total immunoglobulin — that is the classic CVID presentation. An arrow from the phenotype would assert a causal relation that is really part-whole. Routing both from the maturation defect connects them just as well without the claim. Happy to change it if you read it differently.

**Splitting the bundled node.** Agreed, and it turned out to improve accuracy elsewhere. It is now `TH1 Effector Skewing` (keeps `GO:0045063`) and `Gut-Homing Receptor Upregulation on CD4+ T Cells` (honestly unbound). The heterozygous-mouse link now targets the gut-homing node specifically, since those animals were measured for CCR9 and beta-7 integrin and not for T-bet — which is the limitation I had written into that link's own prose and can now express structurally.

**`CL:0000979`.** Agreed, the evidence says CD27+IgG+ and that is `IgG memory B cell` exactly.

**Bronchiectasis.** Agreed. The description now records that the Family B narrative covers father and daughter while Table 1 lists the daughter as not imaged — one confirmed of three, with a second neither confirmed nor excluded.

**The second copy of the bug.** Added as a scope note on the patch docstring: this covers `XMLExtractor.extract` only, the same guard sits in `PMIDSource._fetch_pmc_xml`, that path is not the one supplying full text today, and if it ever becomes so the bug is live there. Left unpatched rather than patched speculatively.

**Double parse.** Fixed — the soup is now built only when there is a body to recover or a `<table-wrap>` to append, so an ordinary full-text paper with no tables is parsed once.

One thing I found while doing the connectivity work, which I have *not* acted on. The gene and variant nodes were isolated because `dismech.graph` indexes genes from top-level `gene` / `gene_term` / `genes` on a pathophysiology item, and this node carried its gene under `genetic_context.gene`. I connected it by adding a `genes:` entry alongside, which is honest curation — the node genuinely is about BACH2 — but the duplication is only needed because of the indexing gap. Any entry that records its gene solely through `genetic_context` has the same silently disconnected gene node. Say the word and I will file it.

Validation after the changes: schema clean, 99/99 snippets verified, terms pass, `just validate-disorders` passes, all offline gates clean, 21,384 KB tests and 16 patch tests pass.

gingin77:
Rebased onto the parent branch, which had moved. #10813 gained 421021ed7a, distinguishing dose match from dose dependence in the two heterozygote limitations, and it touches the same lines this PR rewrites. `mergeStateStatus` is back to CLEAN.

The conflict was worth reading rather than just resolving, because it hid a second problem that git did **not** flag.

The conflicting hunk was on the BRIDA heterozygote's `limitations`, which this PR replaces wholesale — the old text argued for MODERATE fidelity from the fact that only the abstract was available, and that premise is gone. I kept the distinction the parent commit introduced, because it is a good one and still true, and stated that this model now satisfies both halves of it: dose match is a property of the construct and every heterozygote here has it, while dose dependence needs a genotype comparison — and the full text supplies one, comparing Bach2+/- against wild-type for message, protein, Prdm1 and the immunization readouts. That is why the fidelity is HIGH.

The part git could not catch: the same commit also rewrote the **sibling** Igarashi heterozygote's `limitations` to read "hence HIGH here against MODERATE on the BRIDA heterozygote" and "only this one compares genotypes". Both were true against the entry as it stood. Both are false after this PR, which raises that model to HIGH and gives it a genotype comparison. That hunk applied **cleanly**, so it would have merged as a silent contradiction between two adjacent model entries, with each side green on its own base.

This is the shape `CLAUDE.md` already documents for duplicate keys (#8623) and retired enum values (#10061): the defect exists only in the merge result, so neither branch's CI can see it. The prose case has no gate at all. That limitation now names what actually still separates the two entries, which is which compartment each one measured, not dose dependence.

Post-rebase diff reviewed against the new base — only the intended files, no stale reversions or protected-path churn. Schema clean, 99/99 snippets verified, terms pass, entity-ref, causal-target, enum-value and duplicate-key gates clean. The history record carries the rebase and the reasoning.

gingin77:
This PR was fully reviewed and approved days ago, but it was still targeting `claude/immunodeficiency-60-curation-otpq62` — the branch for #10813, which merged into `main` on 2026-09-05. GitHub only auto-retargets a stacked PR when its old base branch is deleted, and that branch was never deleted, so #10876 was stuck pointing at a dead-end base while `main` moved on without it.

Retargeting the base to `main` alone wasn't enough: `main` had gained real, unrelated changes to shared append-only cache files (`cache/hp/terms.csv`, `cache/cl/terms.csv`, etc.) from other curation PRs merged over the same few days, so the retarget flipped `mergeable_state` to `dirty`.

Fix applied: merged `main` into this branch (ordinary merge commit, no rebase/force-push — per the "never rewrite history on someone else's branch" rule) and resolved the two conflict classes:

- **`kb/disorders/Immunodeficiency_60.yaml` and the three `references_cache/*.md` files** (add/add conflicts): confirmed `main`'s copies were byte-identical to this branch's pre-PR baseline (i.e. exactly what #10813 contributed, untouched by anything else since), so this branch's fully-reviewed content was taken as-is.
- **`cache/hp/terms.csv`**: `HP:0031382` had been independently cached by another PR on `main` (same CURIE, same canonical label, only a different `retrieved_at` timestamp). Kept the one already on `main` rather than carrying a duplicate/newer-timestamp row, per the cache-recovery rule against hand-editing derived rows.

Re-verified against the merged tree: schema validation clean, term validation passes, 99/99 evidence snippets verified against the cache, and `check_duplicate_yaml_keys` / `check_enum_values` / `check_entity_refs` / `check_causal_targets` / `check_qualifier_terms` all clean. The PR diff against `main` is now the intended 9 files / 1550+ / 92− (previously inflated to 24 files / 3265+ when compared against the stale base), and `mergeable_state` is clean of conflicts (currently `blocked` only because the push reset the approval per `dismiss_stale_reviews` — CI and `claude-review` are running now).

---

PR #10877, opened 2026-09-04
feat(Dilated_Cardiomyopathy_1W): curate VCL/metavinculin dilated cardiomyopathy (MONDO:0012667)

### LINKED ISSUE #10855: Curate dilated cardiomyopathy 1W (MONDO:0012667)
Curating **dilated cardiomyopathy 1W** as a standalone entry in `kb/disorders/`.

| | |
|---|---|
| MONDO | `MONDO:0012667` |
| Causal gene | **VCL** (`hgnc:12665`) |
| Stub | `stubs/Dilated_Cardiomyopathy_1W.yaml` |

The disease gene acts through metavinculin, the heart-specific VCL splice isoform, and the lesion is in force transmission at the costamere and intercalated disc rather than in the sarcomere itself. That gives the entry a mechanically distinct pathograph from the parent grouping.

- Nothing in `kb/disorders/`, `kb/groupings/` or `kb/modules/` on `origin/main` binds `MONDO:0012667`, and a search on the disease label and the gene name found no entry covering it either.
- No other issue, open or closed, references `MONDO:0012667` or this disease, and no open pull request does.
- VCL appears in existing cardiomyopathy entries only as a gene-list member.

AO33:
Curates **dilated cardiomyopathy 1W** (`MONDO:0012667`), the VCL form of familial dilated cardiomyopathy, as a new entry. The reason it gets its own entry rather than folding into the parent is that its lesion is in a different part of the machine from every other numbered entry we have curated: vinculin and its heart-specific isoform metavinculin anchor the actin cytoskeleton to the intercalated disc and the costamere, so this is a failure of force *transmission* between heart muscle cells, not a failure of force *generation* inside them. CMD1S (MYH7) and CMD1E (TNNT2) break the motor; CMD1W breaks the joint the motor pushes against.

Closes #10855.

## The mechanism, and why it is modelled with two triggers

```
metavinculin-exon allele ─┐
(p.Arg975Trp, p.Leu954del)│
                          ├─> impaired vinculin-actin coupling
VCL loss-of-function    ──┤        │
allele                    │        v
   │                      │   intercalated disc + costamere destabilization
   │                      │        │                    │
   │                      │        v                    v
   │                      │   defective force      connexin 43 lateralization
   │                      │   transmission         + conduction slowing
   │                      │        │                    │
   v                      │        v                    v
additional genetic or ────┘   ventricular dilation   arrhythmia +
environmental stressor         + systolic dysfunction  conduction disease
                                   │
                                   v
                              heart failure ──> reverse remodeling,
                              (reduced EF)      recovered EF
```

The two triggers are kept apart deliberately. The founding alleles sit in the metavinculin-specific exon, so they hit only the muscle isoform and they change how the protein cross-links actin rather than removing it. The predicted loss-of-function alleles hit both transcripts and reduce dosage. Almost all modern evidence — the case-control enrichment, the pediatric skew, the recovery data — comes from the second class only, so merging them would silently attribute infant outcome data to alleles that have none.

**Gene-disease validity and allele penetrance disagree here, and both are curated.** ClinGen classified VCL–dilated cardiomyopathy as *Strong* in August 2024. The largest aggregate study (18,135 cardiomyopathy referrals) concluded that a heterozygous VCL loss-of-function allele on its own is insufficient to cause cardiomyopathy. Those are answers to different questions, so I recorded both: the ClinGen row supports `relationship_type: CAUSATIVE`, and the same sentence from that study appears as `SUPPORT` on the inheritance block and the reduced-penetrance hypothesis and as `REFUTE` against the strong reading of CAUSATIVE in the genetic block.

**The mouse is much sicker than the patients, and I flagged it rather than smoothing it over.** Cardiomyocyte-specific vinculin knockout kills 49% of mice by sudden arrhythmic death before three months, with contractile function still normal. Among 32 human VCL loss-of-function probands, three had bradycardia or heart block and one had supraventricular tachycardia. That is recorded as a `HUMAN_MODEL_MISMATCH` discussion with a proposed heterozygous-dosage experiment, because it is the difference between an arrhythmia-first genotype and a heart-failure-first one.

## On the LUMP_INTO_PARENT score, and #9865

`dashboard/priority.json` scores `MONDO:0012667` at 48.5 with `recommended_action: LUMP_INTO_PARENT`, on `specificity_bucket: subtype_series`. I curated a full entry anyway, and I want to be explicit about that rather than quietly overriding it.

The score is driven by the term being a numbered member of a 16-strong family, which is a fair prior. But one of its features is wrong in a way worth fixing: `clingen_strong_count: 0`. ClinGen does have a Strong assertion for this gene-disease pair — it is filed against `MONDO:0005021` (dilated cardiomyopathy), not against the numbered term, so the counter keyed on `MONDO:0012667` sees nothing. The "no ClinGen support" signal pushing this term toward LUMP is an artifact of which identifier the assertion was filed under, not a statement about the evidence. Any numbered stub whose gene carries a ClinGen assertion against the parent term will score the same way.

This is exactly the pass #9865 asks for, so the disposition for this one, as evidence for that issue: **`entry_type: DISEASE`, curate as a per-gene entry.** It matches the pattern the existing grouping's member list already follows — the KB curates VCL-related DCM, not "type 1W" as an OMIM label — and the force-transmission mechanism is genuinely distinct from the sarcomeric members, which is the test that should decide these rather than the numbering.

**I did not touch `kb/groupings/Familial_Dilated_Cardiomyopathy.yaml`.** Several sibling curation PRs are open and would all conflict on its `members:` list. Adding CMD1W there is left for one batched maintainer pass.

## Sourcing

The Edison/falcon report was used as a lead list only. It cites by author-year key with DOIs and carries no PMIDs, so I resolved every DOI I intended to use to a PMID (PMC ID converter, or PubMed title search) and cited the PMID, keeping every evidence item on the snippet-checked path rather than the `DOI:` skip-prefix path.

Three things in that report needed correction or independent checking:

- Its frontmatter carries `mondo_id: ''` (dismech#10335), so I ran `just preflight-dr` by hand against `MONDO:0012667`. It passed, with VCL mentioned 71 times.
- Its own term-validation section flags five HP identifiers as "named something else". Those are false alarms — the validator read a frequency column as a label — but I re-verified each identifier against OLS before binding it.
- It declines to name a MONDO identifier for the entity and recommends using the parent term instead. `MONDO:0012667` does exist and is what the entry uses.

I also verified `hgnc:12665` resolves to VCL independently via OAK rather than taking the report's `HGNC:12665` on trust, since HGNC sits in the report validator's unresolvable-prefix list.

## Checked and deliberately not curated

- **No clinical trials.** I searched ClinicalTrials.gov directly on 2026-09-03 rather than relying on the report: "vinculin" returns zero studies, and a dilated cardiomyopathy search with "VCL gene" returns zero. No `clinical_trials` block; the absence is a `KNOWLEDGE_GAP` discussion instead.
- **No environmental block.** Alcohol, cardiotoxic chemotherapy and viral myocarditis are all plausible second hits, but none has been measured in VCL carriers. The only citable statement is the study authors' own conjecture, which is curated on the second-hit pathophysiology node where its status as a conjecture is visible.
- **No survival figures.** Contemporary registry outcomes (PMID:32452075) are all-cause and cannot be assigned to this genotype.
- **No GeneReviews snippets.** The applicable chapter is the disease-level "Dilated Cardiomyopathy Overview" (PMID:20301486); its indexed PubMed record carries only the chapter's purpose statement, so it is tagged in `references` but nothing is quoted from it.

## Validation

Every check below was run to completion on this branch.

| Check | Result |
|---|---|
| `just validate` | No issues found |
| `just validate-terms` | Validation passed |
| `just count-verified-snippets` | 76/76 verified |
| `just check-causal-targets` | OK, no new broken targets |
| `just check-qualifier-terms` | OK |
| `just check-qualifier-terms-online` | OK, 326 labels match (97 resolved online) |
| `just check-enum-values` | OK |
| `just check-folded-hyphens` | OK |
| `just check-snippet-length` | OK |
| `just check-title-snippets` | OK |
| `just check-snippet-grading` | OK |
| `just validate-disorders` | All passed |
| `pytest tests/test_data.py` (entity-ref, subtype-FK, model-readout, causal, unique-name selection) | 21241 passed |

Deletes `stubs/Dilated_Cardiomyopathy_1W.yaml`.


AO33:
All three findings are closed using references this PR had already fetched — no new literature was needed, as you said.

**Four-pillar therapy.** `Heart Failure Pharmacotherapy` now carries ARNI, MRA, SGLT2 inhibitor and loop diuretic as `therapeutic_agent` entries alongside the ACE inhibitor and beta-blocker, with five new quotes from `PMID:37653361`. The node description was rewritten to name the four pillars *and* to keep them separate from what the six-infant series actually received, which was ACE inhibitor and beta-blocker only — so nothing now implies infants were given SGLT2 inhibitors.

**Devices.** Three new treatments: ICD (`NCIT:C80435`), CRT (`NCIT:C80436`) and VAD (`NCIT:C80452`), each linked to the mechanism node it acts on. The previously uncited device-indication prose was kept, not deleted, and now points at the ICD treatment.

One thing worth flagging about that claim. The genotype-specific ICD evidence is a sentence about **LMNA**, cited `INDIRECT`, with the inference step written out explicitly: the VCL claim follows from VCL being *absent* from that short list of genotype-specific recommendations, not from any statement about VCL. I also declined to cite `PMID:38867869` there for the same reason — its only genotype-specific content is a gene list that does not name VCL, so citing it would be an argument from silence dressed up as support.

**Diagnosis.** Four entries added ahead of the genetic panel, all from the previously uncited `PMID:33652931`: echocardiography (`NCIT:C16525`), CMR with LGE (`NCIT:C137915`), ECG (`NCIT:C38053`) and Holter monitoring (`NCIT:C38064`).

**Two corrections to note.** `NCIT:C80435`'s canonical label is *Implantable Cardioverter-Defibrillator Placement* — a procedure, not the device, which is why it is legitimately reachable from the `C25218` root. And NCIT has no "mineralocorticoid receptor antagonist" term; `NCIT:C101255` *Aldosterone Receptor Antagonist* is the right concept, its definition reading "aldosterone (mineralocorticoid) receptors". Every CURIE was checked three ways: label in `cache/ncit/terms.csv`, membership in the relevant enum cache, and the OLS definition. No new cache rows were needed.

**One suggestion declined.** Removing `GO:0060047` here alone would break consistency with the sibling numbered-DCM entries and ten-plus other cardiomyopathy files that use it identically, for no gain. Reason recorded in `notes:`.

Snippets 76 to 93, all verified. `validate-disorders` and every per-file gate pass; the whole-KB baselines report no new defects with no baseline modified. `git status` after each validation run showed only the disorder YAML modified — no cache churn.

AO33:
Both done in one push — the blocking rebind and the suggestion.

**Heart transplantation** now binds `NCIT:C15246` Heart Transplantation with `preferred_term: heart transplantation`, matching `Dilated_Cardiomyopathy_1FF` and `Dilated_Cardiomyopathy_2A`. You read it correctly as an oversight rather than a disagreement, and you were right about the mechanism: the finding was in the 06:52Z review, a second review landed seven minutes later with a different set of findings, and my working notes for this batch captured only the most recent review body per PR. So the earlier item was never carried forward. That is a process bug on my side, now fixed — thank you for re-flagging it rather than letting it drop.

**Loop diuretic — the term exists.** You could not confirm one offline because only already-cited NCIT terms are in the local cache. Searching NCIT through OLS turns up `NCIT:C49184` **Loop Diuretic**, and it validates as a member of the `chemicalentityterm` dynamic enum, so the binding is now at the level the snippet actually supports rather than at the `Diuretic` parent.

Snippets remain 93/93 — no evidence was added or changed, both edits are term-level. `validate-disorders` and every per-file gate pass, and the whole-KB baselines report no new defects with no baseline modified. The only cache movement is two additive rows for `NCIT:C49184`.

caufieldjh:
This PR was approved on 4 September and then sat conflict-blocked for 16 days, 1108 commits behind `main`. I've unblocked it. **No content of the disorder entry changed** — this is a merge plus one derived-file removal.

Sorry about the lost approval; `main` dismisses stale reviews on any push, and there was no way to resolve the conflict without one.

## The merge

Merged `origin/main` into the branch rather than rebasing, since it's @AO33's branch and a rebase would invalidate their checkout. Five conflicts, all in derived files.

**Two reference caches, add/add.** `PMID_17785437.md` (Zemljic-Harpf, the cardiomyocyte-specific vinculin knockout) and `PMID_37548861.md` (Zahavich, HFrecEF). Both PMIDs reached `main` independently while this PR waited. I took main's copy of each: main's `PMID_17785437.md` is `content_type: full_text_html` against `abstract_only` here, and main's `PMID_37548861.md` is the longer of two `full_text_xml` fetches — so in both cases main's is a superset, and it's what every other entry citing those PMIDs validates against. Diverging would have left this branch quoting different text from the rest of the KB.

Re-ran the reference validator afterwards: **93/93 snippets still verify**, so nothing this entry quotes was lost in the swap.

**Three cache CSVs, adjacent insertions.** `cache/go/terms.csv`, `cache/enums/cellularcomponentterm_*.csv`, `cache/ncit/terms.csv`. Resolved as the union of both sides keyed on the CURIE, main's row winning a same-key collision, then sorted.

The only same-key collision was `NCIT:C49184` Loop Diuretic — the term this PR added on 4 September in response to the round-3 suggestion, which reached `main` independently a day earlier. Main's timestamp kept.

All of the branch's own contributions survive: `GO:0005925` focal adhesion, `GO:0014704` intercalated disc, `GO:0043034` costamere, `MONDO:0012667`. No row from main was dropped. `check-cache-order` and `check-term-cache-integrity` pass, and `normalize-cache` found nothing to reorder.

## One newly-applicable gate failed, and is fixed

`just check-case-collisions` ([#11204](https://github.com/monarch-initiative/dismech/issues/11204)) postdates this branch and went red after the merge:

```
references_cache/DOI_10.1128_MCB.00728-07.md  <->  references_cache/DOI_10.1128_mcb.00728-07.md
```

The uppercase file was added by this branch; the lowercase one was already on `main`. Neither is cited anywhere in `kb/` — this entry cites that paper as `PMID:17785437` — so I dropped this branch's addition with `git rm --cached`, leaving main's file untouched.

Worth naming what that trades away: it leaves the repo holding the *non*-publisher capitalization for that DOI (`10.1128/MCB.00728-07` is the registered form). That's main's pre-existing state, not this PR's to change, and removing main's file instead would be churning a file this PR has no business touching. The patched fetcher (`src/dismech/doi_cache_case.py`) now reuses a DOI cache differing only in case, so no new duplicate can be written.

## Everything re-verified after the merge

Nothing else regressed across 1108 commits of drift. Each of these was run to completion and its output read:

| Check | Result |
|---|---|
| `just validate-disorders` | passed, **93/93** snippets |
| `just validate` / `validate-terms` | passed |
| `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` | OK |
| `check-enum-values` / `check-qualifier-terms` | OK |
| `check-case-collisions` | OK (was failing) |
| `check-term-cache-integrity` / `check-cache-order` | OK |
| `check-folded-hyphens` / `check-snippet-length` / `check-title-snippets` | OK |
| `check-snippet-grading` / `check-reference-titles` | OK |
| `check-environmental-evidence` | OK |
| `check-genereviews` | `TAGGED` (PMID:20301486, NBK1309) |
| `check-delivery-system` | OK |
| `list-gene-term-mismatches` | 12 gene descriptors, 0 findings |
| `just validate-history-all` | No issues (10,964 records) |

No baseline file was touched, so none of the whole-KB checks passed by widening its own tolerance.

## One thing observed and deliberately not acted on

`just list-disconnected-phenotypes` — also newer than the approval — reports **4 of 8 phenotypes causally connected**. I have not wired the other four. That recipe is report-only by design, the KB-wide `compliance-connectivity` floor is an aggregate no single entry can trip, and connecting a phenotype is real curation that asserts which mechanism produces which clinical feature. Doing it inside a conflict-resolution round would widen a PR whose content was already signed off. Flagging it for whoever wants to pick it up deliberately.

History record for the round: `history/disorders/Dilated_Cardiomyopathy_1W/2026-09-20T195346Z-claude-code-9bcab3.yaml`, recording each conflict resolution and why, since that reasoning is otherwise only reconstructable from a merge diff.

@ai4c-reviewer ready for re-review.

---

PR #10884, opened 2026-09-04
curate: Add Familial Hypocalciuric Hypercalcemia 1 (MONDO:0007791)

### LINKED ISSUE #10881: Curate familial hypocalciuric hypercalcemia 1 (MONDO:0007791)
Curate a dismech entry for **familial hypocalciuric hypercalcemia 1** ([MONDO:0007791](https://monarchinitiative.org/MONDO:0007791)) — a heterozygous loss-of-function disorder of the calcium-sensing receptor (CASR) that raises the set point at which parathyroid chief cells and the renal thick ascending limb read extracellular calcium, producing lifelong mild hypercalcemia with inappropriately normal PTH and low urinary calcium excretion.

Current read is `DISEASE`. The stub records no MONDO descendants, and FHH2 (GNA11) and FHH3 (AP2S1) are separate MONDO terms rather than children of this one, so this is the CASR-specific entity and not a grouping over the FHH types.

**Why this one is worth an entry.** The KB already curates the mirror-image disease: `Autosomal_Dominant_Hypocalcemia_1` (MONDO:0011013) is CASR *gain*-of-function, with a pathophysiology chain running through a left-shifted calcium set point. FHH1 is the same receptor pushed the other way, so the two entries together make the CASR set-point mechanism legible in both directions and give `functional_impact_category` a worked LOSS_OF_FUNCTION / GAIN_OF_FUNCTION contrast on one gene.

The clinical stakes are the reason the stub flags diagnostic delay: FHH1 is chiefly a differential-diagnosis problem against primary hyperparathyroidism, and getting it wrong leads to parathyroidectomy that cannot help. That distinction is the part of the entry worth curating carefully.

**Subtypes.** Homozygous or compound-heterozygous CASR loss of function presents as neonatal severe hyperparathyroidism, a far more severe entity. I will check whether that is already curated separately before deciding whether it belongs here as `has_subtypes` or stays out of this entry.

Stub: `stubs/Familial_Hypocalciuric_Hypercalcemia_1.yaml`

gingin77:
Adds `kb/disorders/Familial_Hypocalciuric_Hypercalcemia_1.yaml` and deletes the stub it curates.

FHH1 is the loss-of-function half of the CASR allelic series, and the knowledge base already curates its mirror image — `Autosomal_Dominant_Hypocalcemia_1` is the gain-of-function disease on the same receptor. I used that entry as the structural template, so the two pathographs are deliberately parallel: a shared calcium set-point node with a parathyroid limb and a renal limb, with every modifier direction reversed. Reading them side by side is what makes "loss of function" concrete here rather than a label.

## The lump/split call

Recorded as `DISEASE`. MONDO records no descendants under MONDO:0007791, and FHH2 (GNA11) and FHH3 (AP2S1) are sibling terms with different genes rather than children of this one, so the term denotes the CASR-specific entity and not a grouping over the FHH types.

Neonatal severe hyperparathyroidism is carried as a `has_subtypes` entry rather than a separate file, because it is the same receptor at a higher mutational dose — biallelic instead of monoallelic loss. It had no entry of its own in the knowledge base. Management inverts between the two, which the subtype records: parathyroidectomy is the thing to avoid in FHH1 and the definitive treatment in NSHPT.

## Two negative findings curated as pathophysiology nodes

This is the part most worth a reviewer's attention, since it is a departure from the usual shape of an entry. In FHH1 the absences carry the mechanism:

- **Preserved bone mineral density despite chronic PTH elevation.** Carriers hold raised PTH and alkaline phosphatase for decades without losing bone, and density does not track variant severity.
- **Near-absence of stones and nephrocalcinosis despite chronic hypercalcaemia.** The renal complications of hypercalcaemia need calcium to be passing through the kidney; the renal limb of the CASR lesion removes that step.

Both follow from the same limb of the lesion, and both are points of differentiation from primary hyperparathyroidism — which is the differential the entire entry turns on, and the reason a missed diagnosis leads to a parathyroidectomy that cannot work.

The renal one is curated as a **negative phenotype**: no `phenotype_term`, no `frequency`, following `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11` and `Amyotrophic_Lateral_Sclerosis_27_Juvenile`. The schema has no negation slot and `FrequencyEnum` has no `EXCLUDED` value, so binding `HP:0000787` with a frequency would export as the assertion that FHH1 patients *have* nephrolithiasis. The description says why the slots are absent, so a later curator does not helpfully restore them. The claim is stated comparatively rather than absolutely, and carries a `REFUTE` item alongside its support.

## Parathyroid localization imaging

`21c1e39` adds a fourth `diagnosis` entry recording what imaging does *not* do. The entry's whole argument is that FHH1 gets mistaken for primary hyperparathyroidism and the patient has an operation that cannot work — and a positive localization scan is one of the real routes to that operation, which the diagnosis section did not cover.

Grounded in PMID:39810898: scintigraphy showed focal uptake above the left thyroid lobe, that plus a mildly raised PTH settled a diagnosis of primary hyperparathyroidism, and the patient was still hypercalcaemic months after total parathyroidectomy. Two further traps from the same case are recorded — the uptake sat where a thyroid nodule also sits, and the resected glands showed four-gland mild hyperplasia, which is a feature of FHH1 rather than evidence against it, so the operative histology did not correct the error either.

## What I deliberately did not claim

- The **pancreatitis** association is recorded as `VERY_RARE`, scoped to the moderate-to-severe end of the hypercalcaemic range, with its attribution flagged as contested rather than asserted.
- The **cardiovascular and neurological rare-variant burden** signal from the DiscovEHR study is held in a `KNOWLEDGE_GAP` discussion rather than folded into the pathograph. The study design cannot separate a mechanistic consequence of impaired calcium sensing from confounding in a healthcare-ascertained cohort. It is a real signal against the "entirely benign" description, so it is recorded rather than dropped, with a proposed replication experiment.
- The **thyroid–parathyroid relationship** itself. It is anatomical and embryological rather than mechanistic, post-thyroidectomy hypoparathyroidism is a different disease, CaSR expression on thyroid C cells has no established FHH1 phenotype, and reported co-occurrence with Hashimoto thyroiditis is the same over-claiming declined above for pancreatitis. The thyroid appears only as the site of a false-positive scan.
- No comorbidity or environmental block, since nothing in the sources supported one.

## Review rounds

Two rounds of review, both CHANGES_REQUESTED, both addressed.

**Round 1** found a semantic inversion on the negative renal phenotype (bound term plus frequency, exporting as the opposite of the claim), four uncited items, a `frequency: RARE` that is not a `FrequencyEnum` value, and a `GO:0070588` (transcellular) binding on a paracellular mechanism. All fixed; the GO node was rebound to `GO:0006816` following the ADH1 precedent, which is now documented in this entry's own `notes` along with the `GO:0004930` choice.

While fixing round 1 I connected two phenotypes to the pathograph with `sequelae`, which runs backwards — `dismech.graph` builds it as `source=phenotype → target, predicate=leads_to`, so it asserted *Chondrocalcinosis leads_to Hypercalcemia*. Caught and corrected to `downstream` edges on the pathophysiology node in `fff68c44f`. No gate catches a reversed edge: `check_causal_targets` only asks whether the bare name resolves.

**Round 2** found that three items quoted a PMID:35566721 sentence truncated exactly where the source's list begins — so the quote named none of chondrocalcinosis, pancreatitis or nephrolithiasis, and for two phenotypes it was the sole citation. Extended through the parenthetical, reproducing the source's `polyuria- polydipsia` and its duplicated `acute pancreatitis, chondrocalcinosis` verbatim. It also found that the negative renal phenotype's only `SUPPORT` item was a bone-density result — correct evidence for the bone node, silent on the kidney. Replaced with two renal-specific items from PMID:21986511 (Christensen 2011): typically normal renal function in explicit contrast with PHPT, and the low urine calcium excretion that is the mechanism behind the low stone burden.

Nothing in the caches, and nothing across six PubMed searches, reports a renal complication *rate* in FHH1 specifically, so the claim is supported at the level the literature actually states it.

## A schema gap this surfaced

`frequency: RARE` passed every gate. `Phenotype.frequency` is `range: Any` with `any_of: [FrequencyEnum, FrequencyQuantity]`, and `FrequencyQuantity` is a bare `str` type, so any string satisfies the union and the enum branch never has to hold. 36 files carry 146 non-enum values as a result, including 23 `EXCLUDED` — curators reaching for the negation slot this PR's critical finding is about. Filed separately as #11419 rather than folded in here.

## Validation

On head `70f292117`:

- `linkml-validate` against `Disease`: passes
- `linkml-term-validator` with `--labels`: passes
- Reference validation: passes, **43/43 evidence snippets verified** against cached references, under both the fast audit and the batched `--no-full-text` validator CI runs
- `check_duplicate_yaml_keys`, `check_enum_values`, `check_entity_refs`, `check_causal_targets`, `check_qualifier_terms`: pass
- `check_folded_hyphens`, `check_snippet_length`, `check_title_snippets`, `check_snippet_grading`, `check_reference_titles`, `check_empty_snippets`, `check_environmental_evidence`: pass, no new baseline entries
- `validate-history` on all three history records: passes
- Cache order and integrity: pass

Three drafting errors were caught and fixed rather than worked around. One snippet quoting the NSHPT sentence spans a bracketed `[OMIM #239200]`, which the reference validator strips before matching; the quote was trimmed to avoid the bracket rather than widening the shared `literal_bracket_patterns` config, whose patterns are documented as deliberately narrow. One `reference_title` was written from memory and was wrong — the title check caught it, and every `reference_title` was then verified against its cache frontmatter verbatim. And the imaging entry's bound label was drafted as "Diagnostic Imaging" when the canonical NCIT label for `NCIT:C16502` is "Diagnostic Imaging Testing".

The cache diff is exactly the three terms this entry introduced (MONDO:0007791, HP:0002918, HP:0003127); no other entry's rows are swept in. `NCIT:C16502` was already in the treatment-action enum cache. Eight reference cache files are added, all cited by this entry.

Closes #10881

https://claude.ai/code/session_01SHJBkfiFu71vfub1qqY9rk

### DIFF
--- a/history/disorders/Familial_Hypocalciuric_Hypercalcemia_1/2026-09-08T020146Z-claude-code-b1c3b9.yaml
+++ b/history/disorders/Familial_Hypocalciuric_Hypercalcemia_1/2026-09-08T020146Z-claude-code-b1c3b9.yaml
@@ -1,0 +1,83 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Familial_Hypocalciuric_Hypercalcemia_1
+  path: kb/disorders/Familial_Hypocalciuric_Hypercalcemia_1.yaml
+session:
+  id: 2026-09-08T020146Z-claude-code-b1c3b9
+  timestamp: '2026-09-08T02:01:46Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10881
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10884
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - phenotypes
+  - evidence
+  summary: Address review round 2 on FHH1
+  details: |
+    Round 2 returned CHANGES_REQUESTED with two findings, both in the evidence
+    layer. Both were verified against the caches before being acted on, and both
+    were correct.
+
+    Finding 1, truncated quote. Three items carried the same PMID:35566721
+    snippet, cut off exactly where the source's list begins - so as quoted it
+    established that some FHH1 patients are symptomatic and named none of the
+    symptoms. For Chondrocalcinosis and Pancreatitis that was the sole citation,
+    so each phenotype's only evidence did not mention the phenotype. Extended
+    through the parenthetical to '...and nephrolithiasis)', stopping before the
+    bracketed citation. The source's own irregularities are reproduced rather than
+    tidied: 'polyuria- polydipsia' with the hyphen and space, and the duplicated
+    'acute pancreatitis, chondrocalcinosis' after both were already listed. That
+    is the same discipline that preserved this entry's 'Familiar' typo, and it is
+    what makes the quote checkable. check_folded_hyphens re-run afterwards, since
+    'polyuria- polydipsia' is exactly the shape that gate watches; clean. The
+    three explanations were rewritten too - each had been pointing at words the
+    quote did not contain, which is what an explanation must never do.
+
+    Finding 2, off-organ evidence on the renal negative. 'Absence of
+    Nephrolithiasis and Nephrocalcinosis' had one SUPPORT item, PMID:19250271's
+    bone-density result, which says nothing about the kidney. It is the correct
+    evidence for 'Preserved Bone Mineral Density Despite Chronic PTH Elevation',
+    where the identical snippet already sits; reused here it argued by analogy
+    across organ systems.
+
+    Rather than take the reviewer's fallback of a general asymptomatic-ness quote,
+    a renal-specific source was fetched: PMID:21986511 (Christensen et al. 2011),
+    from the same Aarhus group whose CCCR study this entry already cites. Two
+    items now carry the claim - 'typically normal renal function' in explicit
+    contrast with PHPT, and the low urine calcium excretion that is the mechanism
+    behind the low stone burden. The bone item was removed rather than regraded to
+    NO_EVIDENCE: it is already correctly placed on the bone node, and a duplicate
+    placeholder would add noise without adding a claim.
+
+    Searched and not found: no committed cache, and nothing surfaced by six PubMed
+    searches, reports a renal complication rate in FHH1 specifically. The claim is
+    therefore supported at the level the literature actually states it - normal
+    renal function and a low filtered calcium load - rather than as a stone-rate
+    comparison nobody has published.
+
+    Not taken into this PR. PMID:42640938 (UK Biobank genotype-first PheWAS, 162
+    FHH1 carriers) reports that pancreatitis- and diabetes-related outcomes did
+    not reach multiple-testing-corrected significance. That is a strong
+    population-scale negative bearing on this entry's contested pancreatitis
+    attribution and on its KNOWLEDGE_GAP discussion about the DiscovEHR burden
+    signal, but adding it is new curation rather than a review fix, so it is
+    recorded here as a follow-up lead.
+
+    Validation. Schema, terms and references pass. Snippets 43/43 verified, up
+    from 42/42, under both the fast audit and the batched --no-full-text
+    validator. check_enum_values, check_duplicate_yaml_keys, check_entity_refs,
+    check_causal_targets, check_qualifier_terms, check_snippet_length,
+    check_snippet_grading, check_title_snippets, check_reference_titles,
+    check_folded_hyphens and check_empty_snippets all pass with no new baseline
+    entries. One new reference cache file, PMID_21986511.md; no cache CSV changes.
--- a/kb/disorders/Familial_Hypocalciuric_Hypercalcemia_1.yaml
+++ b/kb/disorders/Familial_Hypocalciuric_Hypercalcemia_1.yaml
@@ -535,9 +535,12 @@ phenotypes:
     supports: SUPPORT
     evidence_source: OTHER
     snippet: 'Although typically asymptomatic, some patients with moderate-to-severe
-      hypercalcemia may present with classical hypercalcemic symptoms'
-    explanation: Scopes chondrocalcinosis, which the same sentence lists, to the
-      moderate-to-severe end of the range rather than to FHH1 as a whole.
+      hypercalcemia may present with classical hypercalcemic symptoms (i.e., polyuria-
+      polydipsia, fatigue, acute and chronic pancreatitis, gallstones and chondrocalcinosis,
+      acute pancreatitis, chondrocalcinosis, and nephrolithiasis)'
+    explanation: Names chondrocalcinosis among the classical hypercalcaemic
+      presentations and scopes them to the moderate-to-severe end of the range rather
+      than to FHH1 as a whole.
 - category: Gastrointestinal
   name: Pancreatitis
   frequency: VERY_RARE
@@ -558,9 +561,12 @@ phenotypes:
     supports: SUPPORT
     evidence_source: OTHER
     snippet: 'Although typically asymptomatic, some patients with moderate-to-severe
-      hypercalcemia may present with classical hypercalcemic symptoms'
-    explanation: Scopes the symptomatic presentations, pancreatitis among them, to the
-      moderate-to-severe end rather than to FHH1 as a whole.
+      hypercalcemia may present with classical hypercalcemic symptoms (i.e., polyuria-
+      polydipsia, fatigue, acute and chronic pancreatitis, gallstones and chondrocalcinosis,
+      acute pancreatitis, chondrocalcinosis, and nephrolithiasis)'
+    explanation: Names acute and chronic pancreatitis among the classical
+      hypercalcaemic presentations and scopes them to the moderate-to-severe end of the
+      range rather than to FHH1 as a whole.
 - category: Renal
   name: Absence of Nephrolithiasis and Nephrocalcinosis
   description: >-
@@ -585,27 +591,34 @@ phenotypes:
     Pigmentosa") and Amyotrophic_Lateral_Sclerosis_27_Juvenile ("Absence of
     Sensory Neuropathy").
   evidence:
-  - reference: PMID:19250271
-    reference_title: Skeletal consequences of familial hypocalciuric hypercalcaemia vs.
-      primary hyperparathyroidism.
+  - reference: PMID:21986511
+    reference_title: 'Familial hypocalciuric hypercalcaemia: a review.'
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
-    snippet: Familiar hypocalciuric hypercalcaemia is associated with increased PTH and
-      AP compared to normal controls, but not with bone loss irrespective of the severity
-      of the CASR mutations.
-    explanation: The skeletal half of the same comparative point - chronic hypercalcaemia
-      in FHH1 does not carry the end-organ burden it carries in primary
-      hyperparathyroidism.
+    evidence_source: OTHER
+    snippet: Similar to PHPT, FHH is characterized by hypercalcaemia, unsuppressed or
+      elevated plasma parathyroid hormone, and typically normal renal function.
+    explanation: Renal-specific, and comparative in the right direction - FHH1 shares the
+      biochemistry of primary hyperparathyroidism without the renal consequence.
+  - reference: PMID:21986511
+    reference_title: 'Familial hypocalciuric hypercalcaemia: a review.'
+    supports: SUPPORT
+    evidence_source: OTHER
+    snippet: The hallmark is a relatively low urine calcium excretion in contrast to PHPT,
+      in which urine calcium excretion is increased.
+    explanation: The mechanism behind the low stone burden - the filtered calcium load
+      that drives stone formation in primary hyperparathyroidism is not delivered here.
   - reference: PMID:35566721
     reference_title: 'Disorders of the Calcium Sensing Signaling Pathway: From Familial
       Hypocalciuric Hypercalcemia (FHH) to Life Threatening Conditions in Infancy.'
     supports: REFUTE
     evidence_source: OTHER
     snippet: 'Although typically asymptomatic, some patients with moderate-to-severe
-      hypercalcemia may present with classical hypercalcemic symptoms'
-    explanation: Refutes an absolute reading of this absence - the same source lists
-      nephrolithiasis among the presentations seen with moderate-to-severe
-      hypercalcaemia, so the defensible claim is comparative rather than categorical.
+      hypercalcemia may present with classical hypercalcemic symptoms (i.e., polyuria-
+      polydipsia, fatigue, acute and chronic pancreatitis, gallstones and chondrocalcinosis,
+      acute pancreatitis, chondrocalcinosis, and nephrolithiasis)'
+    explanation: Refutes an absolute reading of this absence - nephrolithiasis is named
+      among the presentations seen with moderate-to-severe hypercalcaemia, so the
+      defensible claim is comparative rather than categorical.
 genetic:
 - name: CASR
   gene_term:
@@ -1056,6 +1069,8 @@ references:
   title: Skeletal consequences of familial hypocalciuric hypercalcaemia vs. primary hyperparathyroidism.
 - reference: PMID:20374733
   title: Calcium-sensing receptor and associated diseases.
+- reference: PMID:21986511
+  title: 'Familial hypocalciuric hypercalcaemia: a review.'
 - reference: PMID:32386559
   title: 'Familial Hypocalciuric Hypercalcemia Type 1 and Autosomal-Dominant Hypocalcemia
     Type 1: Prevalence in a Large Healthcare Population.'
--- a/references_cache/PMID_21986511.md
+++ b/references_cache/PMID_21986511.md
@@ -1,0 +1,37 @@
+---
+reference_id: PMID:21986511
+title: "Familial hypocalciuric hypercalcaemia: a review."
+authors:
+- Christensen SE
+- Nissen PH
+- Vestergaard P
+- Mosekilde L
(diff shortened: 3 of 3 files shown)


gingin77:
The `claude-review` check is red, but not because of anything in this diff — the review bot ran out of API quota before it read the PR. From the [job log](https://github.com/monarch-initiative/dismech/actions/runs/33835094849):

```
detail="You've hit your limit · resets 4:40am (UTC)" num_turns=1 cost=$0
```

One turn, zero cost, zero seconds. The agent never started reviewing, so there is no finding here to address and no fix to port — the quota resets at 04:40 UTC and I will re-run the job once after that.

`Build and test`, the check that actually exercises this change, is still running. I will report what it says and fix anything it flags.

---

gingin77:
Update on the two things from my last comment.

**`Build and test` passed** on `4829b69` — the check that actually exercises this change is green.

**I cannot re-run `claude-review` after all.** I said I would once the quota reset at 04:40 UTC; both routes return `403 Resource not accessible by integration` for me — `rerun-failed-jobs` on the run, and the `workflow_dispatch` on `claude-code-review.yml` that `CLAUDE.md` documents for exactly this. So the re-trigger needs someone with workflow write access:

```
gh workflow run claude-code-review.yml --repo monarch-initiative/dismech --ref main -f pr_number=10884
```

The red check is the reviewer bot's own quota exhaustion, not a finding against this diff, and there is nothing in this PR's scope that would make it pass — so I am not pushing anything to force it. An empty commit to kick CI is off the table.

I am still watching the PR and will pick up any review that lands.

---

gingin77:
The push re-triggered `claude-review` and it failed the same way — but this run's log names the root cause, and it's a repo config problem affecting every PR, not just this one.

**`ANTHROPIC_API_KEY` is unset.** From the [job log](https://github.com/monarch-initiative/dismech/actions/runs/33967339249) env dump:

```
ANTHROPIC_API_KEY:
```

The workflow's own error handler already anticipates exactly this:

> This workflow prefers ANTHROPIC_API_KEY (pay-per-use, no weekly cap) and only falls back to CLAUDE_CODE_OAUTH_TOKEN when no API key is set — so a weekly-limit error here most likely means ANTHROPIC_API_KEY is unset and the run used the subscription token.

So every review run is going through the subscription token and hitting its cap. Today's message is `You've hit your limit · resets 4:20pm (UTC)`; yesterday's was 4:40am. Setting the `ANTHROPIC_API_KEY` repo secret is the fix the workflow itself recommends.

**There is also a bug in the classifier, and it is why this has looked mysterious.** That helpful diagnostic never printed. The check is:

```python
```

The actual text is `You've hit your limit`, which matches none of the three. So the run falls through to the generic branch and reports "Common causes: an invalid/retired --model, a bad CLAUDE_CODE_OAUTH_TOKEN, or an aborted run" — pointing at three things that are not wrong, while the accurate message sits unreachable in the branch above. Adding `"hit your limit"` to that alternation would surface the real cause on the first failure instead of the third.

Nothing here is a finding against this diff. `Build and test` was green on the previous head and the same gates pass locally on `21c1e39` — schema, terms, references with 35/35 snippets verified, and every offline check. I still have no permission to re-run the job (403 on both `rerun-failed-jobs` and `workflow_dispatch`), so I am not going to keep pushing to retrigger it.

---

gingin77:
All four findings were correct. Fixed in one push (`33e93adcb`), along with the optional suggestions, since `main` dismisses stale approvals and a follow-up commit would cost another cycle.

**Finding 1 (critical) — the negative phenotype was asserting its own opposite.** You're right, and this is the one I'd have most regretted. `Absence of Nephrolithiasis and Nephrocalcinosis` carried `phenotype_term: HP:0000787` and `frequency: VERY_FREQUENT`; there is no negation slot and no `EXCLUDED` in `FrequencyEnum`, so that pair exports as "FHH1 patients very frequently have kidney stones." Both slots removed, following `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_11` and `Amyotrophic_Lateral_Sclerosis_27_Juvenile`, and the description now says why they're absent so nobody helpfully restores them. I also reframed the claim as comparative rather than absolute — stones *are* reported at the moderate-to-severe end of the range — so it now carries a `REFUTE` item (PMID:35566721) alongside the `SUPPORT` item (PMID:19250271).

**Finding 2 — four uncited items.** Each now has an exact quote: the two magnesium pathophysiology nodes from PMID:39810898, and the `Hypermagnesemia` and `Chondrocalcinosis` phenotypes from PMID:35566721.

**Finding 3 — `frequency: RARE`.** Changed to `VERY_RARE`, and I corrected the sentence in the entry's own `notes` that still described it as RARE.

Worth flagging what I found checking this: **nothing in the stack catches it.** I put `RARE` back in the working tree and ran both gates — `linkml-validate` says "No issues found", `check_enum_values` says "no out-of-enum values". The cause is the slot definition: `Phenotype.frequency` is `range: Any` with `any_of: [FrequencyEnum, FrequencyQuantity]`, and `FrequencyQuantity` is a bare `str` type, so every string satisfies the union and the enum branch never has to hold. `check_enum_values` is right to skip it — the slot genuinely isn't enum-bound.

The corpus reflects that: 36 files carry 146 non-enum frequency values — 23 `EXCLUDED`, 73 HPO frequency CURIEs written with underscores (`HP_0040282`), 41 title-case (`Frequent`, `Occasional`), 8 `VARIABLE`, 2 `RARE`. Several of those look like they were reaching for exactly the negation slot finding 1 is about. I've left them alone — narrowing the slot is a schema change with corpus-wide blast radius and doesn't belong in a curation PR — but I'm happy to open an issue if that's useful.

**Finding 4 — GO:0070588.** Also right. The TAL mechanism is paracellular and claudin-gated, and `calcium ion transmembrane transport` denotes transcellular movement across a membrane. Rebound to the broader `GO:0006816` (calcium ion transport), which is what `Autosomal_Dominant_Hypocalcemia_1` does for the mirror-image node and for the same recorded reason, with the specificity carried in `preferred_term`. A right general term beats a wrong specific one.

**Suggestions taken.** Pancreatitis and Chondrocalcinosis now reach the pathograph via `sequelae` to `Hypercalcemia with Non-Suppressed PTH`, so both read as consequences of the sustained calcium elevation rather than free-floating features; Chondrocalcinosis is also rescoped in prose to the moderate-to-severe range, matching what the source says. Added a top-level `references:` block for parity with ADH1 (8 PMIDs, titles copied verbatim from their caches), and documented both GO binding compromises in `notes`.

**Validation.** Schema, terms and references pass. Snippets 42/42 verified (up from 35/35), under both the fast audit and the batched `--no-full-text` validator CI runs. `check_enum_values`, `check_duplicate_yaml_keys`, `check_entity_refs`, `check_causal_targets`, `check_qualifier_terms`, `check_snippet_length`, `check_snippet_grading`, `check_title_snippets`, `check_reference_titles`, `check_folded_hyphens`, `check_empty_snippets` — all pass, no new baseline entries. No new reference caches and no cache CSV changes: every added quote comes from a file this PR already commits, and `GO:0006816` was already in the `biologicalprocessterm` enum cache.

One reported-not-gated item left in place: `check_causal_targets` notes a self-referential target, the `Hypermagnesemia` phenotype pointing at the pathophysiology node of the same name. That's the flat-node-namespace collapse of #9896, not a curation error.

A history record for this round is in the same commit.

---

gingin77:
Correcting myself on one point in the comment above, before the review lands on it.

I said Pancreatitis and Chondrocalcinosis now reach the pathograph "via `sequelae`". That connection ran backwards. `dismech.graph` builds a `sequelae` entry as `source=phenotype, target, predicate=leads_to`, so what I actually asserted was *Chondrocalcinosis leads_to Hypercalcemia with Non-Suppressed PTH* — the causation inverted. You asked for `downstream` from the hypercalcaemia node specifically, and you were right to; I deviated and got the direction wrong.

Nothing in the stack catches it, which is the same shape as the `frequency: RARE` gap: `check_causal_targets` only asks whether the bare name resolves against another item, and `Hypercalcemia with Non-Suppressed PTH` resolves fine. A backwards edge is a valid edge as far as every gate is concerned.

Fixed in `fff68c44f` — both are now `downstream` targets of the `Hypercalcemia with Non-Suppressed PTH` pathophysiology node, with `causal_link_type: INDIRECT_KNOWN_INTERMEDIATES`. The three remaining `sequelae` blocks in the entry point a phenotype at the pathophysiology node it reports (`Hypercalcemia` → `Hypercalcemia with Non-Suppressed PTH`, and so on), which is the #9896 name-collision shape rather than a causal claim, so I left those alone.

Sorry for burning your in-flight run on `33e93adcb` — I'd rather that than have you approve a backwards causal edge. Everything else in the previous comment stands: schema, terms and references pass on the new head, snippets still 42/42, and the history record is updated with what went wrong.

---

gingin77:
Both round-2 findings were correct. Fixed in `70f292117`.

**Finding 1 — the truncated quote.** You're right that an explanation cannot supply words the quote does not contain, and that was exactly what all three were doing. Extended through the parenthetical to `…and nephrolithiasis)`, stopping before the bracketed citation. Both transcription hazards are reproduced rather than tidied: `polyuria- polydipsia` with the hyphen and space, and the duplicated `acute pancreatitis, chondrocalcinosis`. `check-folded-hyphens` re-run afterwards — clean, no new baseline entry. I also rewrote the three `explanation` fields, since each had been written to lean on the missing text.

**Finding 2 — off-organ support on the renal negative.** Also right, and worse than it looked: reusing the bone snippet meant the entry's own stated point of differentiation was argued by analogy across organ systems.

I went with your third option rather than the fallback. Fetched **PMID:21986511** (Christensen et al. 2011, *Curr Opin Endocrinol Diabetes Obes*) — the same Aarhus group whose CCCR study the entry already cites — which carries two renal-specific statements:

> Similar to PHPT, FHH is characterized by hypercalcaemia, unsuppressed or elevated plasma parathyroid hormone, and typically normal renal function.

> The hallmark is a relatively low urine calcium excretion in contrast to PHPT, in which urine calcium excretion is increased.

The first is comparative in the right direction; the second is the mechanism behind the low stone burden. Both are now `SUPPORT` items on the phenotype.

I removed the bone item rather than regrading it `NO_EVIDENCE`. The identical snippet already sits correctly on `Preserved Bone Mineral Density Despite Chronic PTH Elevation`, so a duplicate placeholder would add noise without adding a claim — but say the word if you'd rather see it kept as `NO_EVIDENCE` for the audit trail.

**What I looked for and did not find.** No committed cache, and nothing surfaced by six PubMed searches, reports a renal complication *rate* in FHH1 specifically. So the claim is now supported at the level the literature actually states it — normal renal function, low filtered calcium load — rather than as a stone-rate comparison nobody has published. Worth knowing before someone re-derives the same dead end.

**One lead I deliberately left out.** PMID:42640938 (UK Biobank genotype-first PheWAS, 162 FHH1 carriers, Aug 2026) reports that pancreatitis- and diabetes-related outcomes did not reach multiple-testing-corrected significance. That bears directly on this entry's contested pancreatitis attribution *and* on its `KNOWLEDGE_GAP` discussion about the DiscovEHR burden signal — but it's new curation, not a review fix, so it's recorded in the history record as a follow-up rather than folded in here.

**On the `sequelae` suggestion** — agreed, and thanks for not filing it. The correspondence links compile to the same reversed edge, but an identity pairing isn't the false causal claim the chondrocalcinosis edge was, and it's KB-wide. That belongs with #9896 against the graph model.

**Issue filed** as you asked: monarch-initiative/dismech#11419. It separates the 146 values into three problems rather than one — case drift (41), HPO CURIEs in underscore form (72), and `EXCLUDED`/`VARIABLE`/`RARE` (33) — because the third is the negation gap that produced round 1's critical finding, and it needs a decision rather than a migration.

**Validation on `70f292117`.** Schema, terms and references pass; snippets **43/43**, up from 42/42. `check_enum_values`, `check_duplicate_yaml_keys`, `check_entity_refs`, `check_causal_targets`, `check_qualifier_terms`, `check_snippet_length`, `check_snippet_grading`, `check_title_snippets`, `check_reference_titles`, `check_folded_hyphens`, `check_empty_snippets` — all pass, no new baseline entries. One new reference cache (`PMID_21986511.md`), no cache CSV changes. History record for the round is in the same commit.

---

gingin77:
Thanks — and both follow-ups are now filed.

**The frequency issue is #11419**, opened about 25 minutes before your round-3 review landed, so you likely didn't see it. It separates the 146 values into three problems rather than one: case drift (41), HPO CURIEs in underscore form (72), and `EXCLUDED`/`VARIABLE`/`RARE` (33). Only the third needs the decision — the other two are migrations.

**The `sequelae` correspondence issue is #11427**, and filing it turned up something that changes the recommendation, so it's worth flagging here rather than leaving in the issue.

You estimated 20+ entries. I measured it across all 2,924 entries with a `pathophysiology` block: **9 edges across 6 files**, out of 336 phenotype `sequelae` edges total.

More importantly, **most of them are correct as written**. Six of the nine are genuine causation — a bicuspid valve does cause altered wall shear stress, hypoxaemia does cause hypoxic pulmonary vasoconstriction, vitamin D and selenium deficiency do drive the mechanisms named in `Hashimotos_Thyroiditis`. Those are phenotypes causing mechanisms, which is exactly what `sequelae` is for; the target living in `pathophysiology` is a section-placement question, not a direction error.

Only three are correspondence rather than causation, and all three are in this entry — the two lab-value-to-mechanism pairs and the `Hypermagnesemia` self-loop.

So the gap is narrower and more specific than "correspondence `sequelae` compile to reversed edges": **there is no slot for "this phenotype is the observable readout of this mechanism."** `biochemical.readouts` and `investigations.reports_on` cover adjacent cases but not a plain phenotype-to-mechanism correspondence, so a curator reaches for the only phenotype-side link available and gets a causal claim they didn't intend. That reframing matters for the fix: a blanket gate on `sequelae` targets resolving into `pathophysiology` would flag the six legitimate edges as often as the three real ones.

I left all nine in place, including my own three — you asked for an issue rather than a change here, and the PR is approved, so a push now would only cost a cycle.

CI is still reporting pending on `70f292117`; I'll keep watching until it's green and mergeable.

---

PR #10888, opened 2026-09-04
curate: Add Mitochondrial Complex I Deficiency, Nuclear Type 36 (MONDO:0030902, NDUFC2) (#10807)

### LINKED ISSUE #10807: Curate mitochondrial complex I deficiency, nuclear type 36 (MONDO:0030902)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 36** ([MONDO:0030902](https://monarchinitiative.org/MONDO:0030902)) — biallelic *NDUFC2* variants; an accessory subunit of the complex I membrane arm.

`MONDO:0030902` is bound nowhere and nothing in `kb/` mentions NDUFC2. Per-subunit MC1DN entries are settled convention here (types 1, 6, 14, 23, 24, 34 are standalone) — read a recent one for shape, and check where MONDO places *this* term rather than assuming the Leigh relationship resolved for type 1 transfers.

Note NDUFC2 also carries a substantial cardiovascular-GWAS literature unrelated to this disease; check each citation is about the mitochondrial disease.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_36.yaml`

kevinschaper:
Closes #10807

Adds a new disease entry for **mitochondrial complex I deficiency, nuclear type 36** (MONDO:0030902), caused by inheriting two damaged copies of *NDUFC2*. The whole published record of this disease is three children from two related-parent families, described in one 2020 paper; nothing has been published on the gene as a disease cause since. All three had Leigh syndrome and died or were severely disabled in early childhood.

The mechanism is unusually clean for a rare disease entry, which is why it is worth having. NDUFC2 is a small subunit of the membrane arm of respiratory complex I that touches twelve other subunits across three assembly modules. When it is missing, the enzyme starts to build and then stops: the first module forms, the two modules NDUFC2 sits between never appear, and no finished complex I is made. Patient cells were left with 16% and 19% of normal complex I activity. Putting a working copy of the gene back into the patients' own cells partly restored assembly, which is what makes the variants causal rather than coincidental.

## What is in the entry

- A nine-node causal chain from the variant through the stalled assembly to the brain lesion, all from the one primary paper
- Twelve phenotypes, including the biochemical defect, Leigh syndrome, raised lactate, developmental delay and regression, spasticity, and seizures
- Genetics, inheritance, diagnosis (both families were only solved by being connected through GeneMatcher), the complementation and complexome experiments as experimental models, and the PRIDE proteomics dataset
- Three discussions: why one missense change costs as much as losing the protein, the two competing names for the same entity, and what is missing under treatment

## Judgement calls

**It gets its own entry rather than becoming a subtype of Leigh syndrome.** MONDO places MONDO:0030902 under "mitochondrial complex I deficiency, nuclear type" and nowhere near Leigh syndrome, and the existing Leigh syndrome entry does not mention *NDUFC2*. This was checked against this MONDO term directly, not carried over from another MC1DN type. The entry keeps both names visible: it is filed under the MONDO name, curates Leigh syndrome as a phenotype present in all three children, and cites ClinGen's decision to fold the Leigh cases into a single gene-defined entity.

**A treatments block was added, holding only what the paper actually records.** The draft had none, on the argument that nothing disease-modifying exists. That is true, but the paper does report two concrete interventions in one child — a ventriculoperitoneal shunt for acute hydrocephalus, and nasogastric tube feeding for an unsafe swallow — and both are quotable. They are curated as complication-directed supportive care, with notes saying no outcome is reported and that the child died at three years. The knowledge-gap discussion was rewritten to match, and still records why the generic GeneReviews Leigh syndrome management recommendations were deliberately not imported: they describe other patients with other genotypes.

**Six frequency bands were wrong and were corrected.** The draft marked six findings `OCCASIONAL`, which asserts 5-29% of patients. Each is present in one of the three reported children, which is 33%, so the schema's band is `FREQUENT` (30-79%). This is the case `docs/frequency-evidence-guidelines.md` calls anti-pattern 1 — the band must match the number the evidence gives. Every changed phenotype now carries a note stating the numerator and denominator, so `FREQUENT` on, say, the Dandy-Walker malformation is not read as a claim that a third of *NDUFC2* patients have one. The `OBLIGATE` and 2-of-3 phenotypes got the same treatment.

**Two claims in the draft's notes pointed at things that do not exist.** It described MC1DN13 as a sibling entry in the knowledge base — it is open PR #10825, not merged — and appealed to an "MC1DN1 precedent" for the Leigh placement question, but there is no MC1DN1 entry to appeal to. Both were corrected, and a GeneReviews precedent list now names MC1DN23, which does carry the tag.

**Named Entity Confusion was a real risk here and is documented in the entry.** *NDUFC2* has a large cardiovascular and complex-trait association literature (hypertension, cardiac hypertrophy, stroke, hyperuricaemia, melanoma) that has nothing to do with this disease, and there is a separate readthrough gene NDUFC2-KCTD14 (HGNC:42956) that regional association signals are often reported against. None of that literature is cited, the binding is `hgnc:7706`, and the trap is written into the genetic notes.

Two fetched references were checked and deliberately not cited: DOI:10.1002/ana.26716 (the ClinGen Leigh spectrum panel curation paper) and DOI:10.1002/1873-3468.13995 (an OXPHOS review). Both are abstract-only in the cache and neither abstract mentions NDUFC2, so nothing in them could support a claim about this disease.

## Validation

| Check | Result |
|---|---|
| `just validate` (schema + terms + references) | pass, 63/63 snippets verified |
| `just validate-disorders` (batched pre-PR gate) | pass |
| Independent snippet re-check against `references_cache/` | all 63 match; the only non-byte-exact matches are Unicode hyphen, prime-apostrophe and non-breaking-space differences the validator normalizes, matching KB-wide convention |
| `reference_title` vs cache frontmatter `title` | all match |
| `just check-entity-refs` / `check-duplicate-keys` / `check-causal-targets` / `check-qualifier-terms` / `check-enum-values` | pass |
| `just check-title-snippets` / `check-snippet-length` / `check-snippet-grading` / `check-folded-hyphens` (whole KB) | pass, no new entries |
| `just verify-datasets` | pass, `pride:PXD014936` resolves and is the NDUFC2 complexome study |
| `just validate-history-all` | pass, 8231 records |
| `just check-stubs` | pass |
| `pytest tests/test_data.py` (selection covering this file plus the unique-name, foreign-key, entity-ref, readout-target and subtype tests) | 18657 passed |

The full unselected `tests/test_data.py` run was not completed — it exceeded a 20-minute local timeout, so the targeted selection above is what was actually verified locally. CI runs the whole suite.

The deep-research report this entry was built from (`research/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_36-deep-research-falcon.md` and its artifacts) is left uncommitted in the worktree; say the word and it can be added in a follow-up commit.

### DIFF
--- a/cache/dataset_accessions.json
+++ b/cache/dataset_accessions.json
@@ -22294,13 +22294,6 @@
     "status": "OK",
     "title": "TMT spikes -  Using R and Bioconductor for proteomics data analysis"
   },
-  "pride:PXD014936": {
-    "checked": "2026-09-04",
-    "detail": "",
-    "extra": {},
-    "status": "OK",
-    "title": "Complexome profiling of mitochondrial protein complexes from patients with mutation in NDUFC2 gene"
-  },
   "pride:PXD021386": {
     "detail": "",
     "extra": {},
--- a/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_36.yaml
+++ b/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_36.yaml
@@ -86,8 +86,9 @@ inheritance:
     molecularly confirmed in two of the three children; DNA was not available from the
     second Saudi sibling, who is presumed homozygous on the strength of parental
     segregation and a concordant phenotype rather than on his own genotype. Carrier
-    testing confirmed both parents of the Moroccan patient heterozygous. NDUFC2 is nuclear-encoded and autosomal, so this is a
-    nuclear recessive mitochondrial disease with a 25 percent sibling recurrence risk and
+    testing confirmed both parents of the Moroccan patient heterozygous. NDUFC2 is
+    nuclear-encoded and autosomal, so this is a nuclear recessive mitochondrial disease
+    with a 25 percent sibling recurrence risk and
     no maternal transmission, which is what distinguishes it from the mitochondrial-genome
     causes of Leigh syndrome.
   inheritance_term:
@@ -110,6 +111,14 @@ inheritance:
     explanation: >-
       Parental carrier confirmation, which turns homozygosity in a consanguineous family
       into a demonstrated recessive segregation.
+  - reference: PMID:32969598
+    reference_title: "Bi-allelic pathogenic variants in NDUFC2 cause early-onset Leigh syndrome and stalled biogenesis of complex I."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "DNA was not available from Subject 2 (affected sibling) to confirm his genotype"
+    explanation: >-
+      The limit on the homozygosity claim, in the paper's own words. It is why this block
+      records molecular confirmation in two of the three children rather than all three.
 pathophysiology:
 - name: Biallelic NDUFC2 Variant
   biological_scale: MOLECULAR
@@ -804,14 +813,17 @@ phenotypes:
   phenotype_term:
     preferred_term: Truncal hypotonia with limb hypertonia
     term:
-      id: HP:0001252
-      label: Hypotonia
+      id: HP:0003808
+      label: Abnormal muscle tone
   notes: >-
     One of the three children, so 1/3 or 33 percent, which is inside the FREQUENT band.
-    Bound to Hypotonia for the truncal component; the limb hypertonia is carried in the
-    preferred_term and description rather than by a second binding, because a single
-    phenotype entry takes one term and splitting one patient's tone examination into two
-    entries would double-count him. The other two children have spasticity, which is
+    Bound to Abnormal muscle tone rather than to Hypotonia. The claim spans hypotonia and
+    hypertonia in the same patient, so a hypotonia binding would have been narrower than
+    the entry name and broader than the preferred_term at once - the one combination that
+    fits neither. Axial hypotonia (HP:0008936) would be the most specific term for the
+    truncal half alone, but taking it would drop the limb hypertonia from the binding
+    entirely; splitting into two entries would instead double-count one patient's single
+    tone examination. The other two children have spasticity, which is
     curated separately - this entry exists so that the third child's documented tone
     abnormality is retrievable rather than surviving only as prose.
   evidence:


kevinschaper:
All four findings applied, plus both suggestions.

**The false-absence claim (CRITICAL).** You were right and it was the worst kind of error to have in the file — the entry asserted that residual NDUFC2 protein "has not been measured", then proposed an experiment to measure it, while the cited paper reports the measurement. I verified it in the cache before changing anything: the complexome profile resolves NDUFC2 abundance directly, none detectable in Subject 1, a scant amount in Subject 3. That is now curated with the exact quote, the false-absence sentences are gone from both the discussion and `genetic_context.notes`, and the proposed experiment is narrowed to the part that genuinely remains open — whether that scant His58Leu protein is excluded from the ND2 module or is simply too scarce to nucleate it. I also took your point that the result is substantive rather than just corrective: the two alleles are *not* identical at the protein level even though the assembly phenotype converges, so the discussion now says that instead of treating them as biochemically indistinguishable.

**Genotype confirmation.** Softened in both places. Molecular confirmation is two of three; the second Saudi sibling is presumed homozygous on parental segregation and concordant phenotype, because no DNA was available. You read it right that this was an oversight rather than a position — the entry was already careful about the same patient's missing biochemistry two sections away.

**Failure to thrive — one correction to the review.** Adding it was right and it is 3/3, but the sentence you attributed to Subject 1 (*"Vomiting, failure to thrive, psychomotor delay..."*) actually opens the **Subject 3** paragraph, and *"impaired growth (weight and length < 3rd centile)"* is Subject 3's too. So those two quotes are the same patient, not two. I checked each subject separately in the cache and the 3/3 does hold, via a different Subject 1 sentence: her height and weight were below the **fifth** centile at six years, where Subjects 2 and 3 are below the third in infancy. The band is `OBLIGATE` per this entry's existing 3/3 convention, but the note says explicitly that the three are not one common threshold, so nobody later reads a single measurement into it.

**Abnormal muscle tone.** Added at 1/3 (`FREQUENT`) bound to Hypotonia, with the limb hypertonia carried in `preferred_term` and the description. I did not split it into two phenotype entries because that would double-count one patient's single tone examination. The 2/3 spasticity count is untouched.

**"The third child".** Named the second Saudi sibling explicitly — you were right that it collided with Subject 3, who is the patient who *did* have material.

Not changed: the McCormick 2023 direct citation. It needs a PMID lookup the research report does not supply, and you flagged it non-blocking; I would rather leave it as the two CGGV records than guess an identifier.

Validation: schema, terms, **68/68 snippets** (was 63), entity refs, duplicate keys, causal targets, enum values, snippet grading, title snippets, snippet length — all pass.

kevinschaper:
Fixed, plus all three suggestions folded into the same push so this doesn't cost an extra re-review cycle.

**The frozen cache file.** You're right, and I checked `CLAUDE.md` on `main` rather than taking it on trust — `cache/dataset_accessions.json` carries the only **NEVER** in the table. Reverted with `git checkout origin/main -- cache/dataset_accessions.json`. Nothing is lost, for the reason you gave: `pride` is not in `REFERENCE_CACHED_PREFIXES`, so PXD014936 re-resolves against the EBI API anyway.

**One correction on where it came from.** The enforcing test does exist — `test_no_automation_touches_the_frozen_dataset_cache` at `tests/test_data.py:2307` — and it is working as designed. It scans code and automation for references to the path, which is a different question from what a commit stages, and no current script names the file. So this isn't tooling drift. The file was already modified in this worktree *before* any of this PR's curation ran, left over from an earlier session, and then got swept up with the term caches. That means there's no separate issue to file against the verifier, but it does suggest the real gap is that nothing checks what a PR *stages* — which is a different (and probably more useful) guard than the one that exists.

**Muscle tone binding.** Took your framing directly — the entry name was broader than the binding while the binding was broader than the `preferred_term`, which fits neither. Bound `HP:0003808` Abnormal muscle tone. I considered `HP:0008936` Axial hypotonia, which is the most specific term for the truncal half, but it would drop the limb hypertonia out of the binding entirely, and the claim genuinely spans both directions in one patient. The reasoning is now in `notes` so the next reader doesn't re-litigate it.

**Genotype qualification.** Now citable rather than prose-only, quoting the paper's own sentence on the `inheritance[0]` evidence list.

**Line 89 wrap.** Restored.

**Failure to Thrive and the pathograph.** Left disconnected, and I want to be explicit that this is a decision rather than an oversight. You framed it as a judgement call, and the deciding factor for me is that no source states the edge from impaired respiration to growth failure. It is biologically plausible, but asserting it would make it my inference rather than the paper's, and this entry is otherwise careful not to do that. It stays consistent with how Microcephaly is handled.

Validation: schema, terms, **69/69 snippets**, entity refs, enum values — all pass.

Issue #10335, opened 2026-09-01
just research-disorder always sends an empty mondo_id, so deep-research reports have no ontology anchor


AO33:
`just research-disorder` never passes the disease's MONDO ID to the research template. It hardcodes the variable to an empty string, so every deep-research report is generated from the disease *name* alone. This removes the one unambiguous identifier that would keep a report on the right disease, and it is the exact condition `just preflight-dr` exists to catch.

2,074 of the 5,068 committed reports in `research/` carry `mondo_id: ''` in their frontmatter.

The relevant line is in `project.justfile` (`research-disorder`, line 1944):

```bash
--var "disease_name=$disease_name" \
--var "mondo_id=" \
--var "category=$category" \
```

The recipe reads `name:` and `category:` out of the disorder YAML a few lines earlier but never reads the MONDO ID, even though the entry always carries one under `disease_term`. The template does use the value — `templates/disease_pathophysiology_research.md` line 5 is `- **MONDO ID:** {mondo_id} (if available)` — so what the provider receives is a blank field.

`research-datasets` already does this correctly at line 1822:

```bash
mondo_id=$(grep -A3 "^disease_term:" "$yaml_file" | grep -o "MONDO:[0-9]*" | head -1 || echo "")
```

So the fix is to copy that line into `research-disorder` and pass `--var "mondo_id=$mondo_id"`. `research-disorder-cyberian-codex` has the same hardcoded empty value at line 2176 and needs the same change.

## Why this matters beyond tidiness

Without the ID, the provider disambiguates from the name, and disease names are not unique enough to do that safely. Two cases from a curation run today:

- **Spinocerebellar ataxia 48.** *STUB1* causes both the dominant SCA48 and the recessive SCAR16 / Gordon Holmes syndrome, and the latter is already curated as `Cerebellar_Ataxia-Hypogonadism_Syndrome.yaml`. A report that blends the recessive phenotype into the dominant entry would read as entirely plausible.
- **Childhood-onset neurodegeneration with cerebellar atrophy.** The name describes a phenotype that many genes produce; only *AGTPBP1* belongs in that entry.

Both reports came back usable, so this is not a claim that the output is wrong — it is that the guardrail was off and the check fell to the curator.

## Possibly related, but a separate question

[...]

PR #10891, opened 2026-09-04
feat: Add Dilated Cardiomyopathy 1O (ABCC9/SUR2A) — the KATP metabolic-sensing member of the numbered DCM series

### LINKED ISSUE #10856: Curate dilated cardiomyopathy 1O (MONDO:0012062)
Curating **dilated cardiomyopathy 1O** as a standalone entry in `kb/disorders/`.

| | |
|---|---|
| MONDO | `MONDO:0012062` |
| Causal gene | **ABCC9** (`hgnc:60`) |
| Stub | `stubs/Dilated_Cardiomyopathy_1O.yaml` |

ABCC9 encodes the SUR2 regulatory subunit of the cardiac ATP-sensitive potassium channel, so this is a channelopathy route into dilated cardiomyopathy rather than a structural one. It is the metabolic-stress-sensing arm of the parent grouping and has no counterpart among the sarcomeric entries.

- Nothing in `kb/disorders/`, `kb/groupings/` or `kb/modules/` on `origin/main` binds `MONDO:0012062`, and a search on the disease label and the gene name found no entry covering it either.
- No other issue, open or closed, references `MONDO:0012062` or this disease, and no open pull request does.
- dismech already curates ABCC9-Related Intellectual Disability and Myopathy Syndrome, but that is the recessive loss-of-function disorder affecting brain and muscle. It is a different disease and does not cover the cardiac phenotype.

AO33:
Adds `kb/disorders/Dilated_Cardiomyopathy_1O.yaml` — dilated cardiomyopathy 1O (CMD1O, MONDO:0012062, ABCC9), and deletes the stub it curates.

CMD1O is the odd one out in the numbered dilated-cardiomyopathy series. Every other numbered sibling gets to a dilated ventricle through a broken structural protein — a sarcomere component, a cytoskeletal linker, a desmosome. ABCC9 does not encode a structural protein at all. It encodes SUR2A, the regulatory half of the heart muscle cell's ATP-sensitive potassium channel, which is a metabolic sensor: it reads how much energy the cell has left and adjusts the cell's electrical behaviour to match. So the entry curates the failure of that sensing system as the lesion, rather than reusing the structural story with a different gene name.

**The gene-disease relationship is not established, and the entry says so everywhere.** ClinGen's Dilated Cardiomyopathy expert panel classifies ABCC9-DCM as **Limited** (2024-11-15, SOP10). The human evidence is two patients from one 1000-patient-era screen, an affected parent nobody could genotype, and a ClinVar record dominated by variants of unknown meaning. The gene block is `relationship_type: DISPUTED` with a `REFUTE`-typed ClinGen evidence item, there is a dedicated knowledge-gap discussion, and the diagnosis, treatment and genetic-counseling entries all repeat the caveat rather than quietly assuming causation.

## The mechanism, as curated

```
ABCC9 C-terminal catalytic-domain variant   (p.Ala1513Thr / p.Leu1524fs, exon 38)
        ↓
Aberrant SUR2A ATP hydrolytic cycle          GO:0016887, DYSREGULATED
        ↓
Defective catalysis-mediated KATP pore gating GO:0015272, LOSS_OF_FUNCTION
        ↓
Compromised metabolic signal decoding in the cardiomyocyte
        ↓
Failure of metabolic-electrical adaptation under cardiac stress
        ↓
Cytosolic calcium overload
        ↓                            ↓
Calcineurin-dependent            Ventricular electrical
maladaptive remodeling           instability
        ↓                            ↓
LV dilation + systolic           Ventricular tachycardia
dysfunction
        ↓
Structural cardiac impairment → Congestive heart failure
```

Ten pathophysiology nodes, one root, five phenotypes, all connected — no orphan nodes and no dangling targets.

Two things I was deliberate about:

- **The frameshift is not treated as a null.** `p.Leu1524fs` truncates *inside* the nucleotide-binding machinery rather than removing the subunit, and the measured defect in the 2004 paper is a redistribution of conformational states in the ATP hydrolytic cycle — a receptor that mis-reports nucleotide state, not a channel that has gone missing. That is why the hydrolysis node carries `DYSREGULATED` and the gating node carries `LOSS_OF_FUNCTION` (loss of *regulatory control*, per the qualitative test in CLAUDE.md) rather than either being flattened to `DECREASED`.
- **Module conformance.** Three nodes conform to `cardiomyopathy_maladaptive_remodeling`. The entry does **not** conform to `cardiac_ion_channel_repolarization` despite being a channelopathy, because that module scopes itself to inherited arrhythmia syndromes in structurally normal hearts and CMD1O is defined by a structurally abnormal one.

## Not the same disease as the other two ABCC9 disorders

This is the thing most likely to get merged by mistake, since all three are described in the literature as "ABCC9 channelopathy":

| | inheritance | channel defect | phenotype |
|---|---|---|---|
| **CMD1O** (this entry) | dominant, heterozygous | regulatory/catalytic | cardiac only: dilated, arrhythmic ventricle |
| **AIMS** (`ABCC9-Related_Intellectual_Disability_and_Myopathy_Syndrome`) | recessive, biallelic | complete loss of function | intellectual disability, myopathy, white matter disease |
| **Cantu syndrome** | dominant | **gain** of function | hypertrichosis, vasodilation, cardiomegaly |

The decisive discriminator is quoted in the entry: heterozygous carriers of AIMS alleles show no consistent clinical problem, so a dominant ABCC9 cardiac disease is a separate claim about a different allele class. The existing AIMS entry was **not** edited.

## Three problems in the deep-research report

The falcon report (`research/Dilated_Cardiomyopathy_1O-deep-research-falcon.md`) was used as leads only, and three of its claims did not survive checking:

1. **A finding attributed to the wrong paper.** The report says a 2024 review (PMID:39337275) "reports the autosomal-dominant ABCC9–DCM relationship as ClinGen Limited". The word "Limited" does not appear anywhere in that paper's cached full text. What it actually says is "Mutations in ABCC9 are directly associated with DCM" — close to the opposite framing. The Limited classification is real, but it comes from ClinGen itself, and that is what the entry cites.
2. **Zero PMIDs, DOIs only.** All 14 references were resolved to PMIDs through the PMC ID converter or PubMed, so nothing sits on the `DOI:` prefix that the gating validator skips.
3. **Empty `mondo_id` in frontmatter** (#10335). The named-entity preflight was run by hand against MONDO:0012062 and passed — ABCC9 mentioned 37 times, no competing gene.

One term correction, and it was mine rather than the report's: I reached for `NCIT:C38048` for ambulatory ECG, which is actually *Vasovagal*; `just validate-terms` caught it and it is now `NCIT:C38064` *Holter Monitoring*. The report proposed only one CURIE (`UBERON:0000948` for heart), which resolves correctly and is unused here. Every HP, GO, CL and NCIT term was re-derived against OLS rather than copied across.

**What I left out on purpose.** The report relays specific ejection fractions, chamber dimensions, ages at diagnosis and death, and a control-panel size. All of those live in the paywalled body of the 2004 *Nature Genetics* paper; the cache holds the abstract only, so none can be supported by an exact quote. They are not written into the entry in either evidence or prose. One visible consequence: the ejection-fraction phenotype is bound to `HP:0012664` *Reduced left ventricular ejection fraction* rather than the severe grade `HP:0012666`, because the grade would depend entirely on numbers I cannot cite.

There is no `clinical_trials` block. ClinicalTrials.gov was searched directly for five term variants; the only ABCC9-named study is NCT04120766, a nicorandil trial in hippocampal sclerosis and dementia. The absence is recorded as a knowledge gap rather than left implicit. `environmental`, `datasets` and ORPHA blocks are likewise absent with a stated reason in `notes`.

## The prioritizer says LUMP_INTO_PARENT, and its reasons are all bookkeeping

`dashboard/priority.json` scores MONDO:0012062 at 48.5 with `recommended_action: LUMP_INTO_PARENT`. That verdict is worth reading closely, because it lines up with what #9865 describes and it is not a judgement about the biology.

The score decomposes exactly:

| feature | value | weight | points |
|---|---|---|---|
| missing from dismech | true | 35 | **+35** |
| has MONDO definition | true | 6 | **+6** |
| synonym count | 5 (capped) | 1.5 | **+7.5** |
| xref count | 6 | 0.5 | **+3** |
| priority subset count | 3 (capped) | 5 | **+15** |
| ClinGen definitive count | **0** | 12 | 0 |
| ClinGen strong count | **0** | 6 | 0 |
| Orphanet/ORDO xrefs | **0** | 4 | 0 |
| subtype series penalty | `subtype_series: true` | −18 | **−18** |
| | | | **48.5** |

Every point comes from "this concept exists in MONDO and we have not curated it", and the only thing pushing it down is an 18-point penalty for having a numbered name in a family of 16. The three features that could carry biological signal — ClinGen validity and Orphanet presence — all contribute zero. The recommendation itself is assigned off `subtype_series: true`, which is derived from the `NUMBERED_SERIES` name-shape heuristic (`"type 1O"`), not from anything about the disease.

**On the ClinGen feature specifically, a sibling PR's finding holds and is stronger than expected.** ClinGen *does* have an ABCC9 / dilated cardiomyopathy assertion, and it is filed against the parent term `MONDO:0005021`, not against `MONDO:0012062`. Checking the whole cached ClinGen set: **52 DCM gene-validity assertions are keyed to the parent term, and zero to any numbered term.** Several of them are Definitive or Strong (BAG3, DES, FLNC, LMNA, plus Strong calls for JPH2 and LDB3), so the signal exists — it just cannot reach a numbered row through the current keying. That is confirmed at scale: across all 23,601 rows in `priority.json`, `clingen_definitive_count` and `clingen_strong_count` are zero *everywhere*. The feature is not merely under-counting numbered terms; it is contributing nothing to any score in the file.

**One honest caveat, since it cuts against the argument I am making.** Fixing that keying would not rescue *this* row. ABCC9's ClinGen assertion is Limited, not Definitive or Strong, so it would still score zero on both features. CMD1O's case for being a separate entry does not rest on gene-validity strength — it rests on the mechanism.

**Why I curated it separately anyway.** Every other numbered member of this series is sarcomeric or cytoskeletal. ABCC9 is the metabolic-sensing member: the channel it builds is not part of the contractile apparatus, and the chain from variant to dilated ventricle runs through nucleotide catalysis, channel gating and stress adaptation before it ever reaches remodeling. Lumping it into the parent would put a channelopathy mechanism inside an entry organised around structural failure, and would lose the arrhythmic phenotype that is written into the concept's own name ("dilated cardiomyopathy with ventricular tachycardia"). The entry carries the Limited classification prominently precisely so that separating it does not read as endorsing the gene-disease claim.

## Left for someone else

`kb/groupings/Familial_Dilated_Cardiomyopathy.yaml` should gain CMD1O as a member. I did not touch it — several sibling curation PRs are in flight against that file's `members:` list and would conflict on it and nothing else. Worth a batched pass once they land.

## Validation

Every check below was run to completion and its output read.

```
just validate kb/disorders/Dilated_Cardiomyopathy_1O.yaml
  Schema validation ....... No issues found
  Term validation ......... ✅ Validation passed
  Reference validation .... All validations passed!
  Snippets checked: 68/68 verified against cached references

just validate-disorders kb/disorders/Dilated_Cardiomyopathy_1O.yaml   (the CI gate)
  ✓ All 1 disorder file(s) passed validation.

just validate-terms ............ ✅ Validation passed
just check-entity-refs ......... OK: entity references resolve in 1 KB file(s).
just check-causal-targets ...... OK: no new broken pathograph targets
just check-duplicate-keys ...... OK: no duplicate mapping keys
just check-enum-values ......... OK: no out-of-enum values
just check-qualifier-terms ..... OK: 1 qualifier term label(s) match the cache
just check-qualifier-terms-online  OK: 2 match (1 resolved online)
```

Whole-KB baseline checks, all reporting no new defects — **no baseline file was modified**:

```
just check-folded-hyphens ........... OK (330 grandfathered)
just check-snippet-length ........... OK (1330 distinct grandfathered)
just check-title-snippets ........... OK (275 distinct grandfathered)
just check-snippet-grading .......... OK (329 distinct grandfathered)
just check-environmental-evidence ... OK
just check-not4curation ............. OK: no bound term carries a 'do not annotate' marker
just check-term-cache-integrity ..... OK: 18 term + 24 enum cache files match the contract
```

`just validate-history` on the history record: `No issues found`.

Cache changes are 8 additive rows across `cache/go`, `cache/hp`, `cache/mondo`, `cache/ncit` and four enum caches, all written by the validators. One shared reference cache file (`PMID_39895490.md`) was upgraded to full text by the refetch; I reverted it, since the quote verifies against the committed version and leaving it out reduces the conflict surface with sibling PRs.

Closes #10856

### DIFF
--- a/kb/disorders/Dilated_Cardiomyopathy_1O.yaml
+++ b/kb/disorders/Dilated_Cardiomyopathy_1O.yaml
@@ -861,8 +861,9 @@ diagnosis:
       label: Magnetic Resonance Imaging of the Heart
   evidence:
   - reference: PMID:39674807
+    reference_title: "Diagnosis and management of dilated cardiomyopathy: a systematic review of clinical practice guidelines and recommendations."
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "cardiac magnetic resonance (CMR) is universally recommended for its ability to assess cardiac function and tissue characterization during the initial evaluation, as well as for identifying specific cardiomyopathy phenotypes and distinguishing them from phenocopies."
     explanation: >-
       Establishes cardiac magnetic resonance in the initial evaluation and, specifically,
@@ -884,8 +885,9 @@ diagnosis:
       label: Laboratory Procedure
   evidence:
   - reference: PMID:39674807
+    reference_title: "Diagnosis and management of dilated cardiomyopathy: a systematic review of clinical practice guidelines and recommendations."
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "All guidelines and recommendations endorse the use of B-type natriuretic peptide (BNP) and troponin assessments as primary diagnostic tools, not only for initial diagnosis but also for evaluating severity, prognosis, and response to treatment."
     explanation: >-
       Supports both analytes as primary diagnostic tools and names the severity,
@@ -1171,7 +1173,7 @@ treatments:
     genotype, aimed at detecting disease early enough to treat. This entry elsewhere warns
     against acting on the genotype; without this record it would leave a reader with no
     positive recommendation for the family at all.
-  therapeutic_modality: BEHAVIORAL
+  therapeutic_modality: OTHER
   treatment_term:
     preferred_term: cardiac surveillance of at-risk relatives
     term:
@@ -1179,16 +1181,18 @@ treatments:
       label: Disease Screening
   evidence:
   - reference: PMID:20301486
+    reference_title: Dilated Cardiomyopathy Overview.
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "Provide a basic view of genetic risk assessment of at-risk asymptomatic relatives of a proband with DCM to inform cardiac surveillance and allow early detection and treatment of DCM to improve long-term outcome."
     explanation: >-
       GeneReviews states the purpose of assessing at-risk asymptomatic relatives as
       informing cardiac surveillance for early detection and treatment, which is the
       recommendation curated here.
   - reference: PMID:37653361
+    reference_title: "Summary and Comparison of the 2022 ACC/AHA/HFSA and 2021 ESC Heart Failure Guidelines."
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "In first-degree relatives of selected patients with genetic or inherited cardiomyopathies for early detection and prompt management"
     explanation: >-
       Records the class-1 recommendation covering first-degree relatives of patients with
@@ -1209,8 +1213,8 @@ treatments:
   treatment_term:
     preferred_term: cardiac resynchronization therapy
     term:
-      id: NCIT:C15329
-      label: Surgical Procedure
+      id: NCIT:C80436
+      label: Cardiac Resynchronization Therapy
   target_mechanisms:
   - target: Structural Cardiac Impairment and Heart Failure
     description: >-
@@ -1218,8 +1222,9 @@ treatments:
       failing ventricle; it does not touch the upstream channel defect.
   evidence:
   - reference: PMID:37653361
+    reference_title: "Summary and Comparison of the 2022 ACC/AHA/HFSA and 2021 ESC Heart Failure Guidelines."
     supports: SUPPORT
-    evidence_source: HUMAN_CLINICAL
+    evidence_source: OTHER
     snippet: "Cardiac resynchronization therapy (CRT) is strongly recommended for symptomatic patients (NYHA class II-IV on best medical management) who have LVEF ≤ 35%, sinus rhythm, left bundle branch block (LBBB) with a QRS of ≥ 150 ms to reduce total mortality and hospitalizations, and improve symptoms and quality of life (CoR: 1)."
     explanation: >-
       States the indication, the eligibility thresholds and the outcomes curated here.


AO33:
All four items are addressed, each from a reference already cached in this PR. Pushed as `be9e109549`.

**GeneReviews is now mined, not just tagged.** You were right that a tag with no citation is worse than useless — it implies the source was consulted. The `abstract_only` record carries exactly one substantive sentence, on assessing at-risk asymptomatic relatives to inform cardiac surveillance, and that is now quoted on the new surveillance treatment. The scope limit is recorded in `notes` so the next reader knows the four-section mining is impossible rather than skipped.

**Relative surveillance is curated as surveillance, deliberately not as cascade testing.** This is the finding I think you identified correctly and it is worth being explicit about why the two are separated here. A Limited-validity gene cannot support predictive genetic testing of relatives — that caveat already lived under genetic counselling. But clinical and imaging follow-up of first-degree relatives is indicated by the familial phenotype regardless of what the genotype is worth, so it survives the gene-validity problem intact. Splitting them means the entry no longer tells a reader to discount the genotype while leaving them nothing to do for the family. Evidenced from both GeneReviews and the class-1 recommendation in `PMID:37653361`.

**CMR and biomarkers** added to `diagnosis` from `PMID:39674807`, with CMR framed around phenocopy discrimination, which is doing most of the diagnostic work given the Limited validity. **CRT** added to `treatments` from `PMID:37653361` with a `target_mechanisms` link to the heart-failure node.

**One thing I did not add.** The LGE ≥7.5% / HR 4.11 figure from deep research §11 is not in any reference cached for this entry, so it is not curated. Noted rather than silently dropped.

**A correction of my own worth recording.** My first pick for the surveillance treatment term was `NCIT:C15320`, which I believed was "Preventive Intervention". It is **Study Design**, and it is not reachable from the `TreatmentActionTerm` root, so `just validate-terms` rejected it. Replaced with `NCIT:C15419` Disease Screening, matching existing KB precedent. Mentioning it because it is the same failure mode this PR already documents in the research report — a resolvable CURIE attached to a confidently wrong label — and here the curator produced it, not the report.

Snippets 68 → 73, all verified. `validate-disorders`, `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-enum-values` all pass; the whole-KB baseline checks report no new defects with no baseline modified. I also restored an incidental `references_cache/PMID_39895490.md` full-text upgrade that validation kept re-applying — all 73 snippets verify against the originally committed abstract, so the PR does not touch that file.

AO33:
All three findings were mine, introduced in `be9e109`, and all three are fixed in `b58d624`.

**`reference_title` on the five new items.** Added, read from each cache file's frontmatter rather than retyped. All 73 evidence items now carry one — I checked structurally by walking the parsed YAML rather than counting string occurrences, since the `references:` block uses `- reference:` too and inflates a naive grep.

**The `evidence_source` regrade is the one worth dwelling on.** You are right, and the file was already self-inconsistent: it graded `PMID:39674807` and `PMID:37653361` as `OTHER` in the items that predate my commit, then graded the same publications `HUMAN_CLINICAL` in the items I added. `evidence_source` classifies the publication, so quoting a different sentence from a guideline systematic review cannot turn it into primary human data. All five are now `OTHER`, matching the earlier items and the GeneReviews chapter.

This also explains why `check-snippet-grading` stayed green through the defect — it keys on the quoted sentence, and my five items quoted sentences no other item used. Worth noting as a gap: the gate cannot see a publication graded two ways when the two gradings quote different sentences.

**CRT binding.** `NCIT:C80436` *Cardiac Resynchronization Therapy* is now bound. I confirmed it is in `cache/ncit/terms.csv` and the `treatmentactionterm` enum cache before using it, so it validates offline with no cache change. You are right that the entry was internally inconsistent as well — the ICD treatment three records down already used its specific `NCIT:C80435`.

**Suggestion taken:** the surveillance treatment's `therapeutic_modality` is now `OTHER` rather than `BEHAVIORAL`. Periodic echo and ECG follow-up is not a behavioral or lifestyle intervention, and `OTHER` is what the two other KB uses of `NCIT:C15419` chose.

**Suggestions not taken, with reasons.** Splitting BNP and troponin into two `diagnosis` records would let each carry a specific term, but the record deliberately describes the paired assessment as the guidelines state it, and the single quoted sentence covers both analytes — splitting would mean citing the same sentence twice for two halves of one recommendation. The duplicate `population: Worldwide` on two `prevalence` records is a real foreign-key ambiguity, but the second record is a sex-ratio observation rather than an occurrence measure, and renaming its population to disambiguate would misdescribe the cohort; I would rather leave it and flag that `prevalence#` keying on `population` is a schema-level sharp edge.

73/73 snippets verified. `validate-disorders`, `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values` and `check-snippet-grading` all pass, with no baseline modified.

PR #10894, opened 2026-09-04
curate: Add Chondrodysplasia Blomstrand Type (MONDO:0008970, PTH1R)

### LINKED ISSUE #10892: Curate chondrodysplasia Blomstrand type (MONDO:0008970)
Curate a dismech entry for **chondrodysplasia Blomstrand type** ([MONDO:0008970](https://monarchinitiative.org/MONDO:0008970)) — a lethal neonatal osteochondrodysplasia caused by biallelic loss-of-function variants in `PTH1R`, in which absent PTH/PTHrP receptor signalling in the growth plate removes the brake on chondrocyte differentiation and produces advanced skeletal maturation with markedly increased bone density.

The stub records a single causal gene (`hgnc:9608` PTH1R) and no MONDO descendants, so this looks like a leaf disease rather than a grouping. It is the loss-of-function mirror of Jansen metaphyseal chondrodysplasia (PTH1R gain-of-function, MONDO:0007982, being curated in #10146); the two are distinct OMIM entities with opposite receptor consequences and should stay separate entries.

Stub: `stubs/Chondrodysplasia_Blomstrand_Type.yaml`

kevinschaper:
Adds a new disorder entry for **Blomstrand osteochondrodysplasia** (BOCD, MIM #215045, [MONDO:0008970](https://monarchinitiative.org/MONDO:0008970)) — a perinatally lethal autosomal recessive osteochondrodysplasia caused by biallelic loss-of-function variants in `PTH1R`.

## Why this disease

It is the clean mechanistic mirror of Jansen metaphyseal chondrodysplasia, which is being curated in #10146. Jansen's constitutively active PTH1R *delays* chondrocyte differentiation; Blomstrand's signalling-dead receptor removes the brake and differentiation runs early. Same gene, opposite direction, opposite skeleton — and the literature says so explicitly (PMID:9649554: "abnormalities that are the mirror image of those observed in Jansen's chondrodysplasia"). Nothing in `kb/` mentioned PTH1R loss-of-function before this.

## The pathograph

Seven nodes, in one chain plus one branch:

1. Biallelic PTH1R loss-of-function (MOLECULAR)
2. Failure of PTH/PTHrP-stimulated signalling in growth plate chondrocytes (CELLULAR)
3. Loss of the Indian hedgehog-PTHrP brake on chondrocyte differentiation (CELLULAR)
4. Accelerated chondrocyte hypertrophic differentiation (CELLULAR)
5. Growth plate hypoplasia and disorganization (TISSUE) → severe micromelia
6. Premature and advanced endochondral ossification (TISSUE) → advanced skeletal maturation, generalized osteosclerosis, dumbbell-shaped long bones
7. Impaired PTHrP-dependent epithelial-mesenchymal signalling in breast and tooth (TISSUE) → absent nipples/breasts, impacted teeth

The extraskeletal branch is worth flagging: absent breasts and impacted teeth were found only when somebody went looking for them (PMID:11297619), in fetuses whose earlier reports had not mentioned them. They are almost certainly under-reported, and the entry says so in `notes`.

## Things I decided deliberately

**Frequency bands are asserted only against the five-fetus series.** PMID:35808914 is the only published series with a denominator (5 fetuses on prenatal ultrasound, 4 on postmortem radiography). Every `frequency:` value traces to that denominator and states it in the phenotype `notes`. Nothing is banded by counting across case reports of different fetuses.

**No numeric prevalence.** There is no population estimate for this condition. The prevalence record is `measure_type: UNKNOWN` / `prevalence_class: ULTRA_RARE` with the reasoning in `notes`, rather than a rate reverse-engineered from a case count.

**Perinatal lethality is left unbound.** `HP:0003826` (Stillbirth) is the obvious label match but is not a member of the `PhenotypeTerm` dynamic enum, and `HP:0003811` / `HP:0001522` are clinical-course modifiers rather than phenotypic abnormalities. Reported outcomes span stillbirth and death after live birth, so no existing HPO phenotypic abnormality covers the claim. The `notes` records what was checked.

**One HUMAN_MODEL_MISMATCH.** The Pth1r-null mouse reproduces the cellular lesion faithfully but is not reported for generalized osteosclerosis — most nulls die in mid-gestation, before there is a skeleton dense enough to radiograph. Whether the human sclerosis is just premature ossification carried to term, or needs something the mouse never lives to show, is open. Two proposed experiments attached.

**ISDS classification is 2023 group 28** (parathyroid hormone signalling cascade), which lists "Blomstrand dysplasia (PTH1R)" by name. The 2019 revision had it in the neonatal osteosclerotic group; provenance is in the assignment `notes`.

## Validation

```
just validate            schema + terms + references — 66/66 snippets verified
just validate-disorders  passed
just check-entity-refs / check-duplicate-keys / check-causal-targets
just check-enum-values / check-qualifier-terms   all OK
just validate-history    OK
```

Whole-KB `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` are being run across both PRs in this batch; I will report the result in a comment.

11 reference cache files, 11 cache rows, one history record, and the stub deletion.

Closes #10892


kevinschaper:
Whole-KB slow gates finished on this branch, all green:

```
just check-title-snippets   OK: no new title-quoting snippets
just check-snippet-length   OK: no new snippets under 5 words
just check-snippet-grading  OK: no new snippet-grading divergences
```

Each was run against `origin/main` as the baseline, so this reports only what the branch adds.

kevinschaper:
Pushed a self-review fix before anyone had to catch it.

Five phenotypes carried `frequency: VERY_FREQUENT` (80-99%) next to notes saying "5 of 5" or "4 of 4". That is an observed fraction of 100%, which the band contradicts — and with a denominator of 4 or 5 you cannot separate OBLIGATE from VERY_FREQUENT anyway. Dropped the band on all five; the numerator/denominator stays in each phenotype's `notes` together with the reason it was withheld.

Perinatal lethality moved the other way, to `OBLIGATE`. That one does not rest on the five-fetus series — no survivor has ever been reported anywhere in the literature, and lethality is definitional for this entity.

Re-ran `just validate` (66/66 snippets) and `just check-enum-values`; both pass. A history record for the edit is included.

kevinschaper:
Rebased onto `main` to clear a conflict.

The Jansen PR (#10146) merged while this was open, and it cites two of the same things — `hgnc:9608` PTH1R and `GO:0004991`, plus PMID:10912527, the PTHrP/Ihh review both entries lean on. The conflict was purely the timestamp column on those cache rows.

Resolved by keeping the rows already on `main` and dropping my duplicates, so this PR no longer touches `cache/hgnc/terms.csv`, `cache/enums/molecularfunctionterm_*.csv`, or `references_cache/PMID_10912527.md` at all. Re-ran `just validate-disorders` (66/66 snippets) and `just check-term-cache-integrity`; both pass.

Worth noting for reviewers that #10146 and this PR are the two ends of the same PTH1R allelic series — activating in Jansen, signalling-dead here — and each entry's `differential_diagnoses` points at the other.

kevinschaper:
All four review items are addressed in one push (b26770b). Here is what changed and one place where I did something narrower than asked.

**The extraskeletal branch is now connected.** `Biallelic PTH1R Loss-of-Function` has a new `downstream` edge to `Impaired PTHrP-Dependent Epithelial-Mesenchymal Signaling in Breast and Tooth`, carrying PMID:11297619's own conclusion: *"Therefore, impairment of the PTHrP/PTHR1 signaling pathway in humans is associated with severe abnormalities in tooth and breast development."*

**Thoracic hypoplasia now has an incoming edge**, from `Growth Plate Hypoplasia and Disorganization` — the node that already produces the micromelia — on the grounds that ribs are endochondral bones growing at the same physes. I graded it `INDIRECT` and said why in the `explanation`, because I could not find a source that actually makes the attribution. I searched for one: all 30 PubMed records indexed for Blomstrand chondrodysplasia (none of the 21 that are not already cited mentions the thorax or ribs at all), the full text of PMID:9649554, and Europe PMC phrase searches for Blomstrand with "short ribs", "narrow thorax" and "thoracic hypoplasia". The two hits that do describe a narrow thorax in a Blomstrand infant (PMID:8445622) and in the PTH1R growth plate (PMID:12368206) are scanned PDFs with no retrievable full text, so nothing quotable exists. The edge therefore rests on two INDIRECT items: PMID:9832466 placing the lesion in "the developing cartilaginous skeleton" as a whole, and PMID:35808914 recording thoracic hypoplasia alongside tetramicromelia in all five fetuses. Both explanations say plainly that this is co-occurrence within one generalized dysplasia rather than a measured cause and effect.

**Retro-micrognathia now has a `sequelae` edge to `Protruding tongue`**, so the causal claim the tongue phenotype's description was already making is carried by the graph. Same treatment: `INDIRECT`, with the explanation noting the source reports the two features together as one prenatal sign and the direction is the standard craniofacial inference.

**The duplicated diagnostic is gone.** You were right that `Molecular confirmation by PTH1R testing` in `treatments` and `PTH1R sequencing` in `diagnosis` were the same claim — same PMID and, in fact, the identical snippet. I dropped the `treatments` copy.

**Deep research artifact added.** `research/Chondrodysplasia_Blomstrand_Type-deep-research-falcon.md` plus its citations sidecar and artifact folder, generated with `just research-disorder falcon` rather than hand-written. Reference validation: 9/9 verified, 0 confabulation. Term validation flags `HP:0002796` as unresolved and `GO:0019933` as obsolete inside the report; neither appears in the KB entry, and I did not import anything from the report into the entry this round.

**One extra fix you did not ask for:** `just check-folded-hyphens` flagged a folded-scalar hyphen split at "splice-site-\ncreating" in the genetic description, which would have rendered as "splice-site- creating". Reflowed.

**What I left alone.** Four phenotypes still have no incoming edge: `Retro-micrognathia`, `Polyhydramnios`, `Increased nuchal translucency` and `Hydrops fetalis`. I did not connect them because nothing in the cited corpus supports a specific upstream node for any of them, and I would rather leave them as roots than invent an edge. The graph is now 24 nodes / 18 edges / 7 components, up from 25 / 15 / 11.

**Validation**

```
just validate                 All validations passed, 69/69 snippets verified
just validate-terms           passed
just check-entity-refs        OK
just check-duplicate-keys     OK
just check-causal-targets     OK, no new broken targets
just check-enum-values        OK
just check-qualifier-terms    OK
just check-title-snippets     OK
just check-snippet-length     OK
just check-snippet-grading    OK
just check-folded-hyphens     OK
just validate-history         No issues found
```

History record: `history/disorders/Chondrodysplasia_Blomstrand_Type/2026-09-04T063627Z-claude-code-2df081.yaml`.

kevinschaper:
Both blocking findings were right, and I checked each against the cached sources before changing anything.

**Finding 1 — type I / type II.** Added a `has_subtypes` block for both types, cited to the Hoogendam 2007 JCEM abstract that this PR had cached and never used. Type I is complete PTHR1 inactivation, anchored on the homozygous R104X nonsense allele; type II is near-complete inactivation with low residual activity, anchored on the intronic splice-creating allele (legacy designation intron M4+27C>T) and the recurrent P132L. The disease-level `description` now names the axis, and the genetic `notes` allelic spectrum picks up both new alleles with their own evidence items.

No subtype frequency is asserted. Hoogendam studied two type I and three type II families — that is a series, not a denominator for the disease, and no other cached source stratifies reported cases by type. A `review_notes` entry also records that no PTH1R mutation was found in the second type I family, so the type I group is defined clinically and is not fully accounted for molecularly.

**Finding 2 — the P132L characterisation.** Read both 1998 primary reports directly. They do not agree with each other:

- PMID:9745456 (Zhang/Silve): cAMP "severely reduced", inositol phosphate undetectable, maximal specific radioligand binding "< 10% of that observed for cells transfected with the wild-type receptor", and normal surface expression by GFP fusion.
- PMID:9832466 (Karaplis): "did not accumulate cyclic adenosine 3',5'-monophosphate ... and did not bind the radiolabeled ligand."
- Hoogendam 2007 resolves it: "the P132L mutant had low residual activity", and assigns P132L to the milder type II.

So the node was wrong twice over — null when the allele is hypomorphic, and implicitly type I when the allele defines type II. The description now separates the null end (R104X, the exon M5 splice allele) from the hypomorphic end (P132L), states the 1998 disagreement plainly rather than picking a side, and cites the 2007 re-analysis. Four evidence items were added to carry those claims.

One thing worth flagging: the `< 10%` sentence could not be quoted whole. The reference validator strips bracketed spans that do not match `literal_bracket_patterns`, and `[Tyr36]` (digits) matches neither pattern. Rather than widen a shared config that the comments say was tuned by replaying every bracketed snippet in `kb/`, I trimmed the quote to start after the radioligand designation and put the full sentence in the `explanation`.

**Non-blocking items taken.**

- **Zygosity.** Removed `zygosity: HOMOZYGOUS` rather than keeping it. PMID:27353973, already quoted in this entry, says "Homozygous and compound heterozygous mutations in PTHR1 gene have been implicated" — so the prose was the accurate half. The slot is single-valued and `ZygosityEnum` has no biallelic value, so asserting HOMOZYGOUS narrows past the source. The reason is recorded in the block description.
- **Absent nipple.** Split into two phenotypes — `Absent nipple` (HP:0002561) and `Breast aplasia` (HP:0100783, already in the term cache) — and split the pathophysiology downstream edge to match, rather than throwing away the breast half by narrowing the name.
- **Coarctation of the aorta.** Added (HP:0001680) from the zebrafish paper's background sentence. Two deliberate restraints: no frequency band, because that sentence says "in some cases" with no numerator or denominator and the ~50% figure comes from the research report rather than a cached source; and no incoming causal edge, because the same sentence says it is unclear whether PTHR1 deficiency causes coarctation directly and the Notch mechanism was shown in zebrafish morphants. Graded `MODEL_ORGANISM`, since `evidence_source` classifies the cited publication's study type; the explanation says the sentence is background rather than a result.
- **GO:0071374.** Kept, with a `notes:` entry on the node saying why. GO has no term for PTHrP-dependent epithelial-mesenchymal signalling in mammary or tooth development, and I checked the GO cache — the alternatives are broader, not closer. The term names the cellular response to the ligand class this receptor transduces, which is what is lost; the tissue-specific claim is carried by `preferred_term` and the node description.
- **Duplicate cache file.** Deleted the lowercase `DOI_10.1161_atvbaha.112.300590.md`, keeping the publisher's uppercase DOI form. For the record this is repo-wide, not specific to this PR: there are 4 such case-duplicate pairs in `references_cache/` on main.

**Declined.**

- **c.1093delG.** The review says the Hoogendam cache supports it. It does not — that abstract lists only R104X, the intronic allele and P132L. c.1093delG comes from a 2025 review (`portalescastillo2025`) that is not in `references_cache/`, so citing it would mean asserting an allele with no cached source. The genetic notes now say so explicitly, so the omission reads as a decision.
- **Pulmonary hypoplasia.** Confirmed your read: grep across `references_cache/` returns no hit for pulmonary/lung hypoplasia in any cited source. The proximate lethal mechanism is already carried by the `Thoracic hypoplasia` phenotype and its edge to perinatal lethality, so nothing is missing mechanistically.

**Validation.** `just validate` green end to end, `validate-terms` passed, `check-entity-refs` / `check-duplicate-keys` / `check-causal-targets` / `check-enum-values` / `check-qualifier-terms` all clean. `count-verified-snippets` reports 73/85 with 12 skipped — `DOI` is in the validator's `skip_prefixes`, so the new Hoogendam and zebrafish items are not machine-checked. I verified all 12 separately with an exact-substring and frontmatter-title check against the cache files: 0 problems. Whole-KB `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` report no new violations. Three files in the commit: the entry, a history record, and the deleted duplicate cache file.

PR #10901, opened 2026-09-04
curate: Add Hearing Loss Autosomal Dominant 84 (MONDO:0030724, ATP11A)

### LINKED ISSUE #10897: Curate hearing loss, autosomal dominant 84 (MONDO:0030724)
Curate a dismech entry for **hearing loss, autosomal dominant 84** ([MONDO:0030724](https://monarchinitiative.org/MONDO:0030724)) — dominantly inherited, postlingual, progressive sensorineural hearing loss caused by variants in *ATP11A*, the P4-ATPase phosphatidylserine flippase, at 13q34.

MONDO records no descendants for this term and a single causal gene (`hgnc:13552` ATP11A), so it looks like a leaf Mendelian entity rather than a grouping. Note that `kb/disorders/Auditory_Neuropathy.yaml` already mentions ATP11A/AUNA2 as one genetic cause among many in the broad ANSD entry; this claim is for the specific DFNA84 entity (OMIM:619810).

Stub: `stubs/Hearing_Loss_Autosomal_Dominant_84.yaml`

kevinschaper:
Adds `kb/disorders/Hearing_Loss_Autosomal_Dominant_84.yaml` — DFNA84 (OMIM:619810), dominantly inherited postlingual progressive bilateral sensorineural hearing loss caused by heterozygous *ATP11A* variants. ATP11A is a class-6 P4-ATPase that flips phosphatidylserine and phosphatidylethanolamine to the inner leaflet of the plasma membrane; this is the first Mendelian phenotype attributed to it.

## Why this entry, and what it says

**The alleles are all in the wrong place, and that is the point.** Every reported disease allele sits in the C-terminal region or the 3' end of the gene rather than the catalytic core: a cryptic-splice substitution retaining 153 bp of intron in the 3' UTR, an 8 bp duplication at the exon 28 boundary, a 5.5 kb deletion of the last coding exon, a further 3' UTR deletion, and a +5 splice-donor variant in intron 12. Several fall outside the coding sequence of the MANE select transcript entirely and were scored benign by in silico prediction. The diagnosis section is therefore built around isoform-aware analysis and minigene confirmation rather than around a gene panel.

**One gene, two auditory phenotypes**, curated as `has_subtypes`: nonsyndromic DFNA84, and AUNA2, where outer hair cell amplification is preserved and the deficit is in neural encoding. The pathograph branches to match — a hair cell / supporting cell arm and a spiral ganglion arm, both downstream of one flippase deficit.

## Entity identity

*ATP11A* sits at 13q34, which is also the historical DFNA33 interval, and the same gene causes AUNA2. This was checked before curating rather than assumed:

- PMID:37671045 states the DFNA84 designation explicitly and attributes it to Pater 2022, and PMID:41007806 independently ties DFNA84 to MIM #619810 and to this gene.
- The same PMID:37671045 reports that the original DFNA33 haplotypes contain two double and one triple recombination — read by its authors as genotyping error. DFNA33 is therefore left unresolved in a discussion rather than merged into this entry, and the shared coordinates are not treated as evidence that the two are one disease.
- `kb/disorders/Auditory_Neuropathy.yaml` already mentions ATP11A/AUNA2 as one cause among many in the broad ANSD entry. This entry is the gene-specific Mendelian entity; AUNA2 is a `has_subtypes` entry on it.

## What is deliberately not claimed

- **No pooled frequency bands.** The only band asserted is `OBLIGATE` on progressive bilateral SNHL, and the `notes:` says it is definitional rather than a counted proportion. The audiometric detail comes from single probands in three separate papers, and pooling them would count different patients toward one denominator.
- **The phosphatidylserine-externalisation-to-phagocytosis step is not an asserted node.** It is the discovery paper's proposal, stated in the conditional, and has not been measured in cochlear tissue in any model. It lives in node descriptions and in a knowledge gap.
- **The dominance mechanism is left open**, as the primary literature leaves it: haploinsufficiency and dominant-negative action are both compatible with the alleles, and they imply different therapeutic routes.

Three knowledge gaps record those open questions plus the DFNA33 disposition.

## Models

Two, both with `modeled_mechanisms`, readouts and explicit limitations: the CRISPR `atp11a` zebrafish (`PARTIALLY_RECAPITULATES`; heterozygotes show an intermediate significant stereocilia deficit, which is what makes it relevant to a dominant disease, and the negative SO1–SO3 neuromast result is recorded alongside the positive O1 one) and the conditional *Atp11a* knockout mouse (`RECAPITULATES` the AUNA2 neural deficit). Both limitations note that the models test loss of the protein rather than the C-terminal alleles patients carry.

## Validation

`just validate`, `validate-terms`, `count-verified-snippets` (57/57), `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `validate-disorders`, plus the whole-KB `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` — all pass.

The `just fetch-reference` run re-wrote the shared `references_cache/PMID_36300302.md` (frontmatter only); that change was reverted so this PR does not touch a cache file another entry owns.

Closes #10897


kevinschaper:
All eight IMPORTANT findings and the three suggestions are addressed in one push. Two I did not do as asked, and I have written the reasons into the entry rather than only here. One thing the review asked for turned out to be worth more than expected: chasing the missing literature turned up a real paper the entry had missed.

## Fixed

**Truncated snippet on PMID:41007806 (finding 7).** You were right on both counts. The snippet now quotes the authors' summary sentence at cache line 102 — "selection of a cryptic splice site that causes partial intron retention was seen in ATP11A and OTOF (c.5533+12G>T) cases" — which contains no typographic prime and names ATP11A. The old `explanation` justified stopping mid-token by claiming the validator normalises the prime glyph; that is not documented in `conf/reference_validator_config.yaml`, and the new explanation says so.

**AUNA2 subtype term (finding 2).** Bound to `MONDO:0957279`, copied from `Auditory_Neuropathy.yaml`.

**AUNA2 in two files (finding 3).** Not resolved silently, and not resolved at all — recorded. The subtype now carries a `review_notes` pointing at the other file, and the entry's top-level `notes` says why both are kept: this entry groups by gene and allelic mechanism, `Auditory_Neuropathy.yaml` groups by audiological presentation, and they cut the same concept along different axes. If a human wants them merged, the note says which half lives where.

**Module conformance (finding 4).** `Cochlear Hair Cell and Supporting Cell Dysfunction` now declares `conforms_to: "sensorineural_hair_cell_loss#Cochlear Sensory Epithelium Insult"`, following the ADNSHL_17 pattern including its "why not the deeper node" sentence — no study shows hair-cell mechanotransduction failure or hair-cell death for any ATP11A allele.

I deliberately did **not** conform `Auditory Nerve and Spiral Ganglion Dysfunction` to the module's `Cochlear Amplification Loss and Spiral Ganglion Neuron Degeneration` node, and wrote that into the node. That module node derives spiral ganglion degeneration from hair-cell loss and lost cochlear amplification. In AUNA2 outer hair cell amplification is explicitly preserved and the neural deficit is primary, so conforming there would assert the opposite of what PMID:36300302 reports.

**Deiters cell (finding 5).** `CL:0002490` → `CL:0000635` "Deiter's cell". Verified against the ontology and it validates.

**OMIM (finding 8).** `external_assertions` entry for `OMIM:619810`, matching the ADNSHL_17 shape.

**Suggestions.** Added a `notes` line on `Hearing Amplification` recording that NCIT was searched and has device terms (`NCIT:C183182` Hearing Aid) but no amplification-fitting action reachable from `NCIT:C25218` — the same negative ADNSHL_11 already records. Added a `notes` line on the intrafamilial-variability phenotype explaining the `HP:0000407` binding and why `HP:0003828` cannot be used. Added `classifications.harrisons_chapter: NEUROLOGIC`.

## Declined, with reasons

**Deep-research artifact (finding 1).** No report was added. `CLAUDE.md` is explicit that `research/` is only for deep-research provider outputs and that ad-hoc markdown must not be hand-written there, and no provider recipe was runnable in this environment. Writing a file that looks like a provider report but is not one would be worse than not having one.

What I did instead was a manual PubMed sweep, recorded in the entry's `notes` and in the history record: ATP11A crossed with hearing loss, deafness, auditory, cochlea, 13q34 and phosphatidylserine flippase, plus "DFNA84" and "AUNA2" as strings. `ATP11A AND hearing loss` returns exactly six records, all six already cited here. "DFNA84" returns nothing indexed.

The sweep did find one genuine gap: **PMID:28601886**, Lang-Roth et al. 2017, the clinical description of the AUNA2 family published six years before the gene was found. It is fetched, committed, and mined into the AUNA2 subtype — natural history (first symptoms in the second decade, moderate loss by the fourth, profound in older members), the linkage result that excluded AUNA1 and placed the family at 12q24 and 13q34, and the explicit statement that AUNA2 is nonsyndromic. So the review's underlying concern was right even though the requested remedy was not available.

**GeneReviews tag on PMID:20301607 (finding 6).** I added the `references:` block you asked for, listing all seven cited papers, but did not tag PMID:20301607 GeneReviews. Its cached record is abstract-only and its entire body is the chapter's six-point statement of scope; `grep -ci 'ATP11A\|DFNA84'` on the cache returns 0. Tagging it would produce a tagged-but-not-mined reference, which is the same defect you flagged as a reason to decline in your review of the sibling DFNB26 PR (#10904). I have followed the `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_2B.yaml` precedent instead and written the negative into the entry's `notes` so it is auditable. If you would still rather have the tag, say so and I will add it.

## Validation

`just validate` (60/60 snippets verified), `just validate-terms`, `just validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` — all pass. `references_cache/PMID_36300302.md` was rewritten again by the validator run and reverted again, so this PR still touches no cache file another entry owns.

A history record for this round is included.

kevinschaper:
Added the deep-research artifact (finding 1). The previous round declined it on a mistaken premise — that producing a report meant hand-writing into `research/`, which CLAUDE.md does forbid. But the recipe **generates** the artifacts, which is exactly what CLAUDE.md permits; hand-writing was never what was being asked for. My apologies, that wrong belief came from me: I put it in the instructions for that round.

Generated with `just research-disorder` (falcon, Edison Scientific Literature backend, 795s). Gates: **6/6 references verified, confabulation_rate 0.0**, 5/6 on topic; term validation 11/12 verified with 0 not found.

**One thing worth flagging for future audiology entries.** `preflight-dr` emits a WARN here about a "rival gene ABR" at 28% of ATP11A's mention count. It is a false positive. Every occurrence of ABR in this report is *auditory brainstem response* — e.g. "hearing can appear normal in the first decade although ABR is already abnormal" — not the *ABR* gene. The gate matches gene symbols against raw text with no disambiguation, so it will fire on essentially any auditory-neuropathy report. The real signal is clean: ATP11A appears 61 times and is the top symbol by a wide margin.

**No YAML change, deliberately.** The entry's content was already reconciled against a manual literature sweep last round — that is what surfaced PMID:28601886, the pre-gene AUNA2 clinical description — and nothing in this report is both new and snippet-supportable from a cached source. Committing the artifact without manufacturing curation from its prose is the honest outcome; the report is a lead, not a source.

The six `DOI_*.md` files are the report's own citations, committed so `validate-research-reference` is deterministic in CI, following the pattern in #10894. I checked each against `origin/main` — none exists there, so no other entry's cache is disturbed.

Since this adds artifacts only and does not touch the YAML, I have **not** dispatched another review — the run from the previous round is still queued and will pick this up.

kevinschaper:
Fixed the IMPORTANT item and took all three suggestions in one push (`40b9d90`).

**The stale sentence.** You are right, and it was my own commit that made it false. The `notes:` literature-base paragraph now records both checks instead of one: the manual PubMed sweep, which is what surfaced PMID:28601886, and the falcon report that commit `c9e341228` added.

I also wrote down what the report contributed, because that turned out to be the useful part. Every indexed source in its citation list resolves to a paper this entry already cites — `DOI:10.21203/rs.3.rs-530835/v1` is the Research Square preprint of PMID:35278131, and the other DOIs are PMID:36300302, PMID:37671045, PMID:40223426 and PMID:41007806. The only two sources left over are unindexed dissertations (von Loh 2021, Pater 2019) with no PMID or DOI, so there is nothing citable in them for a KB entry. Two independent checks agreeing that the literature base is complete is a better claim than either check alone, which is your point and it is now in the entry rather than in a commit message.

The prior round's history record is untouched. New record added for this round.

**Suggestion 1 — `c.1221+5G>C`.** Taken, and you spotted a real internal contradiction. The node opened "All reported disease alleles disturb the C-terminal region or the 3' end" and then listed an intron 12 allele, which is false twice over: intron 12 is mid-gene, and the allele is a VUS rather than an established one. It now reads "every established disease allele", with `c.1221+5G>C` split out as the exception on both counts, and a line saying it is listed because the minigene result is real rather than because it extends the C-terminal pattern. That matches what the prevalence note already said.

**Suggestion 2 — the conforming node's processes.** Taken, in the form you described as defensible. `GO:0007605` is now on the node; `GO:0050910` is deliberately omitted and the conformance paragraph says why.

One correction to how I had to phrase it. I was about to write that no ATP11A study touches mechanotransduction at all, and that would have been an absence I had not checked. PMID:40223426 does mention it: differential Yo-Pro-1 uptake in one zebrafish lateral-line neuromast, which the authors read as "cell loss or loss of mechanotransduction capacity" without deciding between the two. That is not a demonstration of a mechanotransduction defect, and a lateral-line neuromast is not a cochlear hair cell, so the omission stands — but the note says that rather than claiming silence.

**Suggestion 3 — the ABR false positive.** Taken. It is in `notes:` now, with the detail that `hgnc:81` is a real gene called ABR, so a future curator can see why the warning fires and why it is wrong here. I recorded the obsolete `GO:0004012` alongside it for the same reason.

**Not changed**, and flagging so it does not look like drift: DFNA33 stays unresolved (PMID:37671045 reports implausible double and triple recombinations in the original haplotypes), and the spiral-ganglion node still does not conform to the module's amplification-loss node, because AUNA2 has preserved OHC amplification and a primary neural deficit — conforming there would assert the opposite of PMID:36300302.

**Validation.** `validate`, `validate-terms`, 60/60 snippets verified, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` all pass on the file. Whole-KB `check-folded-hyphens`, `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` report no new divergences. `validate-history` clean on the new record. The term validator reserialized `references_cache/PMID_36300302.md` again, exactly as you reproduced; I reverted it, so this PR still touches no cache file another entry owns.

Nothing was regenerated or re-run on account of the review.

PR #10904, opened 2026-09-04
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 26 (MONDO:0011553, GAB1)

### LINKED ISSUE #10898: Curate autosomal recessive nonsyndromic hearing loss 26 (MONDO:0011553)
Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 26** ([MONDO:0011553](https://monarchinitiative.org/MONDO:0011553)) — DFNB26, profound prelingual recessive deafness mapped to 4q31 (OMIM:605428), notable for the dominant modifier locus DFNM1 that suppresses the deafness phenotype in carriers.

MONDO records no descendants and no causal gene for this term, but the literature reports *GAB1* as the DFNB26 gene and a *METTL13* variant as the DFNM1 suppressor — a digenic-modifier architecture worth modeling explicitly.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26.yaml`

kevinschaper:
Adds `kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26.yaml` — DFNB26 (OMIM:605428), prelingual severe-to-profound nonsyndromic recessive deafness caused by homozygosity for *GAB1* p.Gly116Glu, a hypomorphic variant in the pleckstrin homology domain of the scaffold adapter that couples the MET/HGF receptor to its downstream effectors.

## Why this one is worth a full entry

**It has a cloned dominant suppressor.** In the single reported kindred (PK-2), seven members homozygous for the DFNB26 haplotype hear normally, because they additionally carry one copy of *METTL13* p.Arg544Gln at the unlinked DFNM1 locus on 1q24. Suppression is complete, not partial. That is a rare thing in human Mendelian genetics and it has a direct clinical consequence: the *GAB1* genotype on its own does not predict hearing status, so the modifier is curated into the diagnosis section rather than only into the mechanism.

**The mechanism narrows to one transcript.** p.Gly116Glu cuts PtdIns(3,4,5)P3 binding by about 42% without changing binding specificity — the adapter is recruited less efficiently rather than abolished. That hypomorphism is load-bearing: the *GAB1* null is embryonic lethal in mice and *gab1* knockdown is lethal in zebrafish, so only a partial-function allele can present as isolated deafness. Downstream, 37 MET/HGF pathway transcripts were profiled in patient lymphoblastoid lines; fourteen move in deaf and non-penetrant homozygotes alike and cannot be what separates them, and exactly one — *SPRY2* — tracks hearing status. METTL13 coimmunoprecipitates with both GAB1 and SPROUTY2, so the suppressor sits in the same complex as the discriminating transcript.

**The lesion is not in hair cells.** A *Gab1* LacZ reporter labels organ of Corti supporting cells, spiral ganglion neurons, stria vascularis and sulcus cells, with no detectable staining in inner or outer hair cells. That negative result is cited explicitly, and the pathophysiology node is placed accordingly.

## Modelling decision to review

**METTL13 is curated as `relationship_type: MODIFIER`, not as digenic inheritance, and HP:0010984 is deliberately not bound.** Digenic inheritance means variants at two loci are jointly required to produce the disease. Here the *GAB1* genotype alone is sufficient and the second locus subtracts the phenotype, so binding it as digenic would invert the direction of the second allele's effect. The reasoning is written into the second `inheritance:` block so it does not have to be rederived.

## What is hedged, and why

- **The cellular node is expression-based, not demonstrated pathology.** No model of the human *GAB1* allele exists, so the node localises the pathway rather than asserting a cochlear lesion, and says so in its description.
- **The zebrafish entry is `PARTIALLY_RECAPITULATES` with `fidelity: LOW`.** The scored phenotype is early embryonic development — eye formation, epiboly delay — not hearing. It is an allele-function assay used to show that p.Gly116Glu retains partial function and that the METTL13 allele acts allele-specifically (wild-type METTL13 mRNA does not rescue; that negative control is cited). The limitations say all of this.
- **No pooled frequency bands.** DFNB26 is one family. The only band is a definitional `OBLIGATE`, whose `notes:` says it is definitional rather than counted, and the prevalence record states the evidence base explicitly. Affected-individual counts are attributed carefully: the source says eight affected individuals were *included in the linkage calculation*, which is not the same claim as eight being the total, and the entry does not upgrade it.
- Two knowledge gaps: whether normalising SPRY2 *causes* the suppression or merely marks it (METTL13's predicted enzymatic activities have never been demonstrated), and whether a METTL13-based intervention would reach other MET/HGF-pathway deafness genes — recorded at the authors' own "we speculate" strength.

## Snippet handling

The PMID:29408807 cache is PDF-derived and carries mid-word hyphenation artifacts (`developmen-tal`, `senso -rineural`, `homo -zygous`, `METTL13-  dominant`). Rather than quote across them, every snippet was re-cut to a clean span and each affected `explanation:` records what was trimmed and why. `count-verified-snippets` reports 45/45 with zero cache-defect normalisations.

Two references fetched while scoping (PMID:16579977, PMID:39062623) turned out not to be cited; their untracked cache files were deleted rather than committed.

## Validation

`just validate`, `validate-terms`, `count-verified-snippets` (45/45), `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-environmental-evidence`, `validate-disorders`, plus the whole-KB `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` and `check-folded-hyphens` — all pass.

Closes #10898

### DIFF
--- a/history/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26/2026-09-04T171557Z-claude-code-5eaf17.yaml
+++ b/history/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26/2026-09-04T171557Z-claude-code-5eaf17.yaml
@@ -1,0 +1,39 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26
+  path: kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26.yaml
+session:
+  id: 2026-09-04T171557Z-claude-code-5eaf17
+  timestamp: '2026-09-04T17:15:57Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10904
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - notes
+  - genetic
+  summary: 'Review round 2: reconcile notes with the committed deep-research artifact'
+  details: |
+    Addressed the CHANGES_REQUESTED review of 2026-09-04T08:10Z on PR #10904, which found one new blocking item introduced by the previous round's fix.
+
+    IMPORTANT (fixed). The notes: literature-base paragraph still said no deep-research provider report was generated, which commit 5e3db82 in this same PR falsified by adding research/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26-deep-research-falcon.md. Rewrote that sentence to name the report and say what was taken from it, which is nothing, with the reason: the validator scored 1 of 5 verified references as on topic. The paragraph now records the composition behind that number - five of the report's twelve numbered citations are the same 2018 JCI paper already cited as PMID:29408807, and the rest are adjacent-locus deafness papers (Mujtaba 2015 MET/DFNB97; Faridi 2024 MAP3K1/DFNB128, verified via NCBI esummary to be PMID:39062623, the paper the manual sweep had already declined; the Naz and Friedman growth-factor review) plus an Open Targets query. None reports on DFNB26. The manual sweep therefore remains the working literature base and the report serves as a retrieval check. The rest of the paragraph was kept intact. The prior round's history record was left untouched; history is append-only.
+
+    SUGGESTION (taken). Recorded the c.1631G>A / c.1634G>A nomenclature discrepancy in the METTL13 variant description, with the counts checked against the cache: PMID:29408807 prints c.1631G>A eight times (Sanger results, Table 2 against NM_015935, Figure 1E exon 6 legend) and c.1634G>A twice (zebrafish rescue narrative and discussion). The entry uses c.1631G>A, which is the form the sequence-level results carry.
+
+    SUGGESTION (taken). Recorded in notes why references_cache/DOI_10.1172_JCI97350.md and references_cache/DOI_10.1172_jci97350.md both exist: the deep-research citation list emits the lowercase DOI and the fetcher keys on the literal string, so both are needed for deterministic reference validation.
+
+    SUGGESTION (declined, with the reasoning recorded in the item's explanation). The reviewer noted the snippet at the hypomorphic-allele item spans an in-vitro lipid-binding assay and a zebrafish rescue in one sentence. It cannot be split: it is a single sentence and no contiguous substring carries the hypomorphic conclusion without the cross-arm attribution. Both arms are already carried separately with their own grades in this entry - the quantified PIP-array result as IN_VITRO in the same node, the zebrafish morphant rescue as MODEL_ORGANISM under the METTL13 modifier - so nothing is lost. The explanation now states that the sentence is a cross-arm synthesis and that it is graded on the biochemical arm.
+
+    Preserved. The METTL13 suppressor remains relationship_type: MODIFIER rather than digenic, with HP:0010984 left unbound. The MT13-C / MT13-N assignment established in the previous round is unchanged: p.Arg544Gln sits in MT13-C, which targets the eEF1A N terminus, while MT13-N is the Lys55 methyltransferase.
+
+    Validation. just validate, validate-terms, count-verified-snippets (51/51), check-entity-refs, check-duplicate-keys, check-causal-targets, check-enum-values, check-qualifier-terms all pass on the file. Whole-KB check-folded-hyphens, check-title-snippets, check-snippet-length and check-snippet-grading report no new divergences. No cache CSV or references_cache file changed in this round.
--- a/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26.yaml
+++ b/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26.yaml
@@ -72,9 +72,23 @@ notes: >-
   on the modifier side: METTL13's enzymatic activity, described as merely predicted in the
   2018 JCI paper, was characterised later that same year (PMID:30143613) and independently
   in 2019 (PMID:30612740), and both are now cited. PMID:39062623 (MAP3K1/DFNB128) surfaces
-  on a GAB1-deafness query but concerns a different locus and is not cited. No
-  deep-research provider report was generated for this entry; this sweep stands in its
-  place and is recorded in the history record.
+  on a GAB1-deafness query but concerns a different locus and is not cited. A falcon
+  deep-research report was also generated and is committed at
+  research/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_26-deep-research-falcon.md. No
+  curation was drawn from it, and its relevance profile says why: the validator scored only
+  1 of its 5 verified references as on topic. Five of its twelve numbered citations are the
+  same 2018 JCI paper already cited here as PMID:29408807; the remainder are adjacent-locus deafness
+  papers - Mujtaba 2015 on MET/DFNB97, Faridi 2024 on MAP3K1/DFNB128 (which is
+  PMID:39062623, the paper this sweep had already examined and declined), and the Naz and
+  Friedman growth-factor review - plus an Open Targets query. None reports on DFNB26. The
+  manual sweep therefore remains the working literature base for this entry, and the report
+  serves as a retrieval check that surfaced nothing it had missed.
+
+  Reference cache duplication. references_cache/DOI_10.1172_JCI97350.md and
+  references_cache/DOI_10.1172_jci97350.md are the same article in two DOI casings. Both
+  are present deliberately: the deep-research citation list emits the lowercase form and
+  the fetcher keys on the literal string, so removing either one would make reference
+  validation non-deterministic. This is not a stray duplicate.
 
   GeneReviews. PMID:20301607, the Genetic Hearing Loss Overview, is cached and is tagged
   GeneReviews in four sibling hearing-loss entries (ADNSHL_41, Auditory_Neuropathy,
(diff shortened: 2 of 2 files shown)


kevinschaper:
All four IMPORTANT findings and all four suggestions are addressed in one push. One finding I did not do as asked, with the reason written into the entry. And one detail of the review's own argument is wrong in a way that changes the conclusion, so I have written the corrected version instead.

## The METTL13 correction (finding 2) — and where the review has it backwards

You are right that the entry was asserting a stale negative. METTL13's activity is no longer merely predicted, and PMID:30143613 (Jakobsson, *Nat Commun* 2018) and PMID:30612740 (Liu, *Cell* 2019) are both fetched, committed and cited now. The 2018 JCI snippet is untouched, and its `explanation` now says it is the historical baseline that later work superseded rather than a current negative.

But the domain assignment in the review is the wrong way round. The review says p.Arg544Gln sits in "MTase13-C, the eEF1A-Lys55 methyltransferase domain". Jakobsson assigns those two activities to opposite domains:

> METTL13 is capable of methylating eEF1A in vitro and suggest that MT13-C targets the N terminus of eEF1A while MT13-N methylates a different site

and later, "MT13-N is required and sufficient for methylation of Lys55 in eEF1A". Liu confirms it independently from the other direction: "The N-terminal MTase 1 domain is necessary for METTL13 methylation of GST-eEF1AK55, whereas the C-terminal domain is dispensable for this activity", and residues 1-401 alone suffice for Lys55 methylation.

Arg544 is in MT13-C — Jakobsson crystallised its core at residues 470-699, and it is MT13-C that "lacks close paralogs, but is distantly related to spermidine synthase (SpdS)", which is exactly why the JCI paper's annotation called that region a spermine/spermidine synthase domain. So the mapping does connect the old annotation to the new one, as you said. It just lands on the **eEF1A N-terminal** methyltransferase activity, not the Lys55 activity.

That matters for the knowledge gap, so the entry says the specific thing: the suppressor now has a candidate reaction to act through, that reaction is eEF1A N-terminal trimethylation, and what remains open is whether it acts through eEF1A methylation at all (nobody has measured eEF1A methylation in any DFNB26 sample) and by what route that would reach SPROUTY2 levels. I also added a cheaper intermediate experiment than the knock-in mouse: assay eEF1A methylation in the lymphoblastoid lines that already exist from this family.

## Also fixed

**`evidence_source` (finding 3).** All seven regraded, and each `explanation` now says why. Three items quoting the lymphoblastoid RT-PCR panel went `HUMAN_CLINICAL` → `IN_VITRO`; the "normal-hearing humans" one is the case worth naming, because the *subjects* are people but the *measurement* is the same cell-line panel. Four items quoting framing, future work, a therapeutic proposal and explicit speculation went `HUMAN_CLINICAL` → `OTHER`.

**`references:` block and the GeneReviews decision (finding 4).** Block added for all four cited papers. PMID:20301607 is not tagged, and the entry's `notes` now says why in the `Autosomal_Dominant_Nonsyndromic_Hearing_Loss_2B.yaml` style: the cached record is abstract-only, its body is the chapter's six-point statement of scope, and it contains no occurrence of GAB1, METTL13, DFNB26 or DFNM1.

One small correction while I was there. The review says PMID:20301607 is tagged in "roughly fifteen sibling hearing-loss entries". It is four — ADNSHL_41, `Auditory_Neuropathy`, `COL11A2_Hearing_Loss` and `X-linked_Nonsyndromic_Hearing_Loss`. Six entries mention the PMID at all, and the sixth is ADNSHL_2B declining to tag it. The entry's note now names the four rather than guessing at a number.

**Fourteen vs thirteen (suggestion 5).** You were right — thirteen downregulated plus PI3K upregulated is fourteen. The discussion rationale now matches the description and spells out the breakdown.

**Mouse provenance (suggestion 8).** The spiral-ganglion node description now says the LacZ reporter and the colocalisation are mouse.

**Cochlear implantation evidence (suggestion 7).** Declined, and recorded. I searched both cached records for "implant" and neither paper mentions cochlear implantation at all, so there is nothing here to quote. A `notes` field records the search and states that the indication in the description is general standard of care rather than something this disease's literature establishes. Manufacturing a quote for it is exactly what the evidence SOP forbids.

**Prelingual Onset (suggestion 6).** Kept as its own phenotype, with the disconnection now deliberate and documented rather than accidental. Two reasons, both in the `notes`. It is the timing axis of the same clinical event, not a separate consequence of any node, so there is no edge either source supports — and `CLAUDE.md`'s rule for that case is to leave it disconnected and say so. Folding it into an `onset:` block would mean picking an `OnsetEnum` value, and both papers say only "prelingual", which does not distinguish congenital from neonatal or infantile onset; `CONGENITAL` would be more specific than the literature. It is a separate phenotype only so that `HP:0000399` stays queryable alongside the severity term `HP:0011476`, since one phenotype carries one term.

## Declined

**Deep-research artifact (finding 1).** No report was added. `CLAUDE.md` says `research/` is only for deep-research provider outputs and that ad-hoc markdown must not be hand-written there, and no provider recipe was runnable in this environment. A file that looked like a provider report but was not one would be worse than none.

I ran a manual PubMed sweep instead, recorded in the entry's `notes` and in the history record: GAB1 crossed with hearing loss and deafness, DFNB26, DFNM1, METTL13, METTL13 crossed with deafness, METTL13 crossed with eEF1A. Results:

- The disease association is still two papers and one family. No third family, no independent replication.
- The gap it did find is exactly the one behind finding 2 — the post-2018 METTL13 enzymology.
- PMID:39062623 surfaces on a GAB1-deafness query but is MAP3K1/DFNB128, a different locus; screened and not cited.

So your underlying concern was right even though the requested remedy was not available here.

## Unchanged

The METTL13 `relationship_type: MODIFIER` call and the unbound `HP:0010984` are untouched, since the review endorsed both.

## Validation

`just validate` (51/51 snippets verified), `just validate-terms`, `just validate-disorders`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` — all pass. Two new reference caches are included in the PR. No cache CSV row changed, since no new ontology term was bound.

A history record for this round is included.

kevinschaper:
Added the deep-research artifact (the finding the last round declined). That refusal rested on a mistaken premise — that producing a report meant hand-writing into `research/`, which CLAUDE.md forbids. The recipe **generates** the artifacts, which CLAUDE.md permits. The wrong belief came from my own instructions for that round, not from the curation.

Generated with `just research-disorder` (falcon, Edison Scientific Literature backend). Gates: 5/5 references verified, confabulation_rate 0.0; **20/20 terms verified**, 0 not found, 0 obsolete.

Two caveats I would rather state than bury.

**`preflight-dr` returns SKIP, not a pass.** MONDO records no causal gene (`RO:0004003`) for MONDO:0011553, so the automated gene-identity check cannot discriminate at all here. The manual fallback it directs you to does pass: the report's OMIM is 605428, matching MONDO's exactly, and the top gene symbols are GAB1 (61) and METTL13 (28) — precisely this entry's gene and its unlinked suppressor. The entity is right, but that conclusion rests on the manual check, not on the gate.

**Only 1 of 5 references is assessed as on topic.** Every citation is real and verified, but four-fifths being off-topic makes this a low-quality lead rather than a usable source, and it is why I have drawn no curation from it. Mining a report with that relevance profile is exactly the failure mode the "treat reports as leads" rule exists to prevent.

**No YAML change.** The entry was reconciled against a manual sweep last round — which confirmed there is no third family and surfaced the two METTL13 papers behind finding 2 — and nothing here is both new and snippet-supportable. The METTL13 domain correction from that round (p.Arg544Gln in MT13-C, with MT13-N as the Lys55 methyltransferase) stands unchanged.

The five `DOI_*.md` files are the report's own citations, committed for deterministic CI validation per #10894; none exists on `origin/main`.

Artifacts only, no YAML change, so I have **not** dispatched another review — the queued run from last round covers this.

kevinschaper:
Fixed the blocking item and took two of the three suggestions, in one push (`c22a766`).

**The `notes:` sentence.** You are right that my own commit falsified it. It now names the falcon report and says what I took from it, which is nothing, with the reason spelled out rather than just the `on_topic: 1` number.

I went one step further than the sentence you asked for, because the composition is what actually makes the case: five of the report's twelve numbered citations are the same 2018 JCI paper the entry already cites as PMID:29408807. The rest are adjacent-locus deafness papers — Mujtaba 2015 on MET/DFNB97, Faridi 2024 on MAP3K1/DFNB128, and the Naz and Friedman growth-factor review — plus an Open Targets query. None of them reports on DFNB26. The Faridi paper is the interesting one: I resolved its DOI (`10.3390/genes15070845`) through NCBI and it is PMID:39062623, the exact paper the manual sweep had already surfaced and declined as a different locus. So the report re-found what the sweep found and stopped at the same wall. That is now the entry's stated reason for treating the sweep as the working literature base, and it makes the report a retrieval check rather than an unexplained omission. The rest of the paragraph is unchanged.

Prior round's history record untouched; new record added for this round.

**Suggestion — the c.1631/c.1634 fork.** Taken, and I counted rather than asserted. PMID:29408807 prints `c.1631G>A` eight times — the Sanger sequencing results, the Table 2 variant summary against NM_015935, and the Figure 1E legend placing it in exon 6 — and `c.1634G>A` twice, in the zebrafish rescue narrative and once in the discussion. That is in the variant `description` now, so the next curator does not re-derive it.

**Suggestion — the duplicate DOI caches.** Taken; a line in `notes:` says both casings are deliberate and why.

**Suggestion — the mixed-source snippet at line 199. Declined, and here is the argument.** I do not think it splits cleanly, in the specific sense of "cleanly": it is a single sentence, and there is no contiguous substring of it that carries the hypomorphic conclusion without also carrying the cross-arm attribution. The only way to produce two items would be to quote the same sentence twice under two `evidence_source` values, which is exactly the defect `check-snippet-grading` exists to catch — a sentence's study type cannot change because the item moved.

More to the point, nothing is lost by leaving it. Both underlying arms are already carried separately and correctly graded in this entry: the quantified PIP-array result as `IN_VITRO` in the same node, and the zebrafish morphant rescue as `MODEL_ORGANISM` under the METTL13 modifier. What this sentence adds is the authors' synthesis — the inference from both to "hypomorphic, which is why the phenotype is restricted to hearing" — and it is graded on the biochemical arm because that is the one measuring the allele's residual function. I have written that reasoning into the item's `explanation` so it is visible to the next reader rather than sitting in this thread.

**Thanks for re-deriving the METTL13 domains independently.** That was the finding I was least sure I could defend by assertion, and having `PMID_30612740`'s "the N-terminal MTase 1 domain is necessary for METTL13 methylation of GST-eEF1A-K55, whereas the C-terminal domain is dispensable" as an independent confirmation of the split is worth more than my reading of the 2018 paper alone. Nothing in that section changed.

**Validation.** `validate`, `validate-terms`, 51/51 snippets verified, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` all pass on the file. Whole-KB `check-folded-hyphens`, `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` report no new divergences. `validate-history` clean on the new record. No `cache/` CSV and no `references_cache/` file changed in this round — the diff is the disorder YAML and the history record.

PR #10907, opened 2026-09-04
chore(stubs): record entry_type decisions for three nominated concepts

kevinschaper:
Three stub-queue concepts get an `entry_type` decision here, so none of them needs a disorder entry written. One is deleted; two are kept on purpose, and that is the part worth reading — I did not delete the other two, and the reason is the same in both cases.

## The decisions

**narcolepsy without cataplexy (MONDO:0019371) → SUBTYPE. Stub deleted.**

`kb/disorders/Narcolepsy.yaml` (disease_term MONDO:0021107, which is this stub's own MONDO parent) already carries `Narcolepsy Type 2` in `has_subtypes`, described as "Without cataplexy, normal or near-normal orexin levels" and cited to a 2024 review. That is exactly what MONDO:0019371 names, and the stub's own synonym list gives "narcolepsy type 2". Verified by reading both files.

I went one step past the brief and bound MONDO:0019371 as the `subtype_term` on that existing subtype record. Without it, the ID was bound nowhere in `kb/`, and deleting the stub would have let the next `just seed-stubs` run recreate it verbatim as UNDECIDED — the failure mode already written up in `stubs/Dyskinesia_With_Orofacial_Involvement.yaml`. With the binding, `just tidy-stubs` reports the stub as `already_curated: MONDO:0019371 is covered by disorders/Narcolepsy.yaml — stale, tidy up`, so the deletion is what the repo's own tooling asks for rather than a judgement that only survives in a commit message.

Narcolepsy Type 1 was left unbound on purpose: its MONDO mapping is the subject of separate open stubs (`Narcolepsy_1`, `Narcolepsy_7`) that this pass did not decide.

**adult Krabbe disease (MONDO:0016091) → SUBTYPE, with a caveat. Stub kept.**

The brief flagged this one to check carefully, and the check matters. `kb/disorders/Krabbe_Disease.yaml` has **no `has_subtypes` block at all**. It represents the adult form as a `progression` phase (`Adult Onset`, `age_range: ">16 years"`). A progression phase is not a subtype record, so the original evidence does not by itself establish coverage.

SUBTYPE is still the right call on the merits: MONDO:0016091's direct parent is MONDO:0009499, which is that entry's `disease_term`, and adult-onset Krabbe is the same GALC deficiency driving the same psychosine-mediated demyelination — an onset stratum, not a distinct pathograph.

I considered adding a single `has_subtypes` record for the adult form and rejected it. GeneReviews (PMID:20301416, already cited by the entry) classifies Krabbe as infantile-onset versus later-onset — "onset of manifestations after age 12 months and as late as the seventh decade" — and does not treat adult Krabbe disease as a separately named entity. A lone "Adult Onset" subtype would assert a boundary the source does not draw, and would sit asymmetrically beside two unmodelled sibling onset forms.

Worth flagging for someone else: `progression` looks like the wrong axis for all three onset forms. Infantile, late-infantile/juvenile and adult Krabbe are alternative presentations a patient has one of, not phases a patient passes through. Converting them to `has_subtypes` is the structurally correct fix and is recorded in the stub's `notes` as the retire-when condition.

**glutaric aciduria (MONDO:0000129) → GROUPING. Stub kept, priority dropped to LOW.**

Read off the stub's own `mondo_descendants` list, not re-derived: 8 descendants spanning three enzymatically distinct disorders — GA1 (MONDO:0009281, GCDH / hgnc:4189), GA2 (MONDO:0009282, ETFA/ETFB/ETFDH) plus its five severity and complementation-group children (MONDO:0018332, MONDO:0018333, MONDO:0700073, MONDO:0700074, MONDO:0700076), and GA3 (MONDO:0009283). GA1 and GA2 share little beyond the urinary metabolite that named them.

Both curatable arms already exist as separate disorder entries — `Glutaryl-CoA_Dehydrogenase_Deficiency.yaml` and `Multiple_Acyl-CoA_Dehydrogenase_Deficiency.yaml`, gene bindings confirmed in each. That split is itself the evidence that MONDO:0000129 is a union rather than an entity.

## Where I disagreed with the brief

The brief asked for all three stubs to be deleted. I deleted one.

The repository's practice is not what `CLAUDE.md`'s prose suggests. All 24 stubs currently carrying a non-UNDECIDED `entry_type` are still present in `stubs/`, and every stub deletion in `git log` accompanies an actual `kb/` curation. Two of those kept files say why in their own `notes`: `Mucopolysaccharidosis_Type_1.yaml` ("no dismech object currently claims it") and `Dyskinesia_With_Orofacial_Involvement.yaml` ("Because the ID is bound nowhere in kb/, deleting the stub outright would let the next `just seed-stubs` run recreate it verbatim as UNDECIDED and lose this judgement").

The rule that actually holds is: delete when the concept is bound in `kb/`, keep the decision on file when it is not. MONDO:0016091 and MONDO:0000129 are bound nowhere. Making them bound would mean, respectively, restructuring Krabbe's onset axis and writing a `kb/groupings/Glutaric_Acidurias` entry with membership criteria and differentiating mechanisms — both real curation, both outside a decision pass. So both stubs carry the decision, the reasoning, and an explicit retire-when condition instead. Happy to delete them anyway if you would rather the queue shrink.

## Validation

- `just check-stubs` — OK: 1326 stubs, no errors (41 advisory, all pre-existing `possible_kb_duplicate` notices)
- `just validate-stubs` — No issues found
- `just validate kb/disorders/Narcolepsy.yaml` — all validations passed
- `just validate-terms kb/disorders/Narcolepsy.yaml` — passed (added one generated row to `cache/enums/diseaseorsubtypeterm_*.csv`)
- `just check-entity-refs` / `just check-duplicate-keys` on `Narcolepsy.yaml` — OK
- `just validate-history` on the new record — No issues found

`just validate` also rewrote frontmatter in seven unrelated `references_cache/*.md` files as a side effect; I reverted that churn rather than carry it here.

A history record accompanies the `Narcolepsy.yaml` edit.


kevinschaper:

PR #10909, opened 2026-09-04
curate: Add bilateral microtia-deafness-cleft palate syndrome (MONDO:0012854, HOXA2)

### LINKED ISSUE #10906: Curate bilateral microtia-deafness-cleft palate syndrome (MONDO:0012854)
Curate a dismech entry for **bilateral microtia-deafness-cleft palate syndrome** ([MONDO:0012854](https://monarchinitiative.org/MONDO:0012854)) — a *HOXA2*-related craniofacial disorder (OMIM 612290) in which loss of HOXA2 function disturbs second-pharyngeal-arch patterning, giving bilateral microtia with conductive hearing impairment, and, in the recessive form, cleft palate.

Preliminary read: one gene, one developmental mechanism, so a `kb/disorders/` entry. The stub's synonym list distinguishes a recessive form (microtia, hearing impairment and cleft palate) from a dominant form (microtia with or without hearing impairment); those look like `has_subtypes` entries on one entry rather than separate diseases, since both trace to reduced HOXA2 dosage. `kb/modules/pharyngeal_arch_patterning_serial_homology.yaml` is the likely conformance target.

Duplicate preflight run: no `MONDO:0012854` / `HOXA2` / microtia-syndrome match in `origin/main` `kb/disorders` or `kb/groupings`, no open PR, no existing claim.

Stub: `stubs/Bilateral_Microtia-deafness-cleft_Palate_Syndrome.yaml`

kevinschaper:
Adds `kb/disorders/Bilateral_Microtia-Deafness-Cleft_Palate_Syndrome.yaml` — the HOXA2-related external and middle ear malformation spectrum, five published families.

## What the mechanism is

HOXA2 is the selector gene of the second pharyngeal arch. Genetic fate mapping in mouse showed that the auricle derives entirely from HOXA2-expressing second-arch neural crest mesenchyme, correcting the older view that it was a composite of first- and second-arch tissue. So when HOXA2 function is lost, second-arch crest adopts first-arch identity, the second-arch skeletal derivatives are replaced by duplicated first-arch structures — including the middle-ear ossification centres — and the pinna program fails. That gives bilateral microtia, a conductive component to the hearing loss from the ossicular malformation, and, in the recessive form, cleft palate.

The one target tied experimentally to the human alleles is the long-range HMX1 enhancer: both Chinese nonsense alleles fail to activate it in reporter assays.

## The lumping decision

MONDO:0012854 is labelled for the recessive microtia-hearing impairment-cleft palate presentation, but its synonym set also carries "microtia with or without hearing impairment", the dominant form. Both are HOXA2 allelic disorders running through the same second-arch mechanism, so they are curated as one entry with two `has_subtypes` rather than as two diseases:

- **AR-MHIC** — one consanguineous Iranian family, homozygous p.Q186K homeodomain missense, with mixed severe-to-profound hearing loss and partial cleft palate
- **AD-Microtia** — four families with heterozygous nonsense alleles, largely isolated bilateral microtia, no cleft in any of them

The cleft palate is what separates them, and that is recorded on the phenotype rather than left in prose.

## Curation decisions worth a reviewer's attention

**The mouse is both the best evidence here and clearly not the human disease.** Hoxa2-null mice die within 24 hours with cleft palate and a complete second-to-first arch homeotic transformation; no human patient comes close. A `HUMAN_MODEL_MISMATCH` discussion records that gap and keeps the candidate explanations separate — allele dose (human dominant disease is heterozygous, the human recessive allele is a missense with residual protein), timing (the conditional study shows stage of inactivation alone spans absent to merely hypoplastic auricle), and species difference — rather than picking one. Three mouse models are linked through `modeled_mechanisms` with readouts.

**The p.Q186K DNA-binding effect is graded `COMPUTATIONAL` with `directness: INDIRECT`.** It is a homology-modelling prediction from Modeler, not a functional assay, and the entry should not read as though the variant has been characterised.

**A second discussion records what the mechanism does not explain.** The recessive family's hearing loss is explicitly mixed, so there is a sensorineural component that a second-arch skeletal lesion does not account for. None of the cited papers examines the inner ear in HOXA2 loss of function, and the dominant families' hearing loss is not audiologically subtyped in the published abstracts.

**Three nodes conform** to `kb/modules/pharyngeal_arch_patterning_serial_homology.yaml`. HOXA2 is a new trigger class for that module — an arch-identity selector lesion, not a ribosome-biogenesis neurocristopathy and not an EDN1-EDNRA-DLX5/6 dorsoventral-code lesion — which is the substitution the module's own description invites.

**No phenotype carries a frequency band.** Five families is a case series, not a cohort. The all-cause microtia birth prevalence (1 in 8,000-10,000) is recorded as a separate, explicitly labelled context row, since HOXA2 accounts for only a small fraction of it.

## Validation

`just validate` (57/57 snippets verified against the cache), `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` both offline and with `--resolve`, `validate-disorders`, and the whole-KB `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` — all green.

Closes #10906


kevinschaper:
All four blocking items fixed and all three suggestions taken, in one push (`25007c6`).

**1. The prevalence band.** You are right and it was not a rounding call — the whole 10.0–12.5 interval sits above the 9-per-100,000 ceiling, not just the midpoint. Now `BAND_1_5_PER_10000`. Rates and `notes` untouched.

**2. The deep-research pass.** Run and committed, along with its citations sidecar, artifact, and the eleven DOI reference caches the run fetched. `preflight-dr` passes against MONDO:0012854 with HOXA2 mentioned 54 times; 11 of 11 references verified at 0.0 confabulation, 29 of 29 terms resolved.

You were right that "the curator says the literature is exhausted" and "a retrieval pass confirms it" are different claims, and this is the case where the difference mattered. **The report found a sixth published HOXA2 family that my manual sweep missed**: Meddaugh and Zambrano 2020, `PMID:31567444`, a four-generation pedigree with microtia and variable hearing impairment.

I cannot curate it. Both the PubMed record and the DOI record are abstract-free — I fetched each and the caches are title-only — so there is no quotable sentence, and a title is not a finding. So it is recorded in `notes:` rather than added to `references:` as a citation nothing mines, which is the same reasoning the sibling hearing-loss entries use for the GeneReviews overview. The report attributes a truncating allele around Glu229 to that family, distinct from the p.E224* and p.Q235* alleles already curated; I could not verify that and it is not asserted anywhere in the entry. I softened the two family-count statements — the `Bilateral Microtia` phenotype notes and the `CASES_IN_LITERATURE` prevalence notes — so they read as counts of families the entry can *cite* rather than counts of the literature. The `ULTRA_RARE` band does not move.

Two other things the report surfaced, both recorded:

- **Checked and not taken.** Wilderman et al. 2024 describe a ~600-kb noncoding global control region between NPVF and NFE2L3 driving anterior HOXA expression, whose deletion in mouse produces orofacial clefts resembling the Hoxa2 null. That is upstream regulatory architecture for the cluster, not a HOXA2 lesion in any patient with this disease — the report is explicit about that itself. Noted as a candidate for a future regulatory-variant node.
- **Confirmed rather than changed.** Every management citation the report reached for is general microtia and aural-atresia literature (international consensus recommendations, integrated atresia management, a national treatment guideline). None is HOXA2-specific. That independently supports the treatments section's claim that no HOXA2-specific outcome data exists, which until now rested only on my own search.

Incidentally, `preflight-dr` fires the same "rival gene ABR" warning here that it does on the DFNA84 entry, and it is the same false positive: every ABR is the auditory brainstem response.

**3. `CL:0000008`.** Rebound; verified through OAK, passes `validate-terms` and dynamic-enum membership. Both CURIEs were already cached, so no cache CSV moved. Your parenthetical about "migratory" sitting awkwardly against a post-migratory lesion is fair, and `CL:0000008` is still strictly the better of the two.

**4. `NCIT:C25351`.** Rebound; I confirmed independently through OAK that `NCIT:C25218` is among its ancestors. I left `Hearing Amplification` on the generic device term for the reason you gave — the description covers both air and bone conduction.

**Suggestion — treatments not joined to the pathograph.** Taken. `Auricular Reconstruction` → `Bilateral Microtia` and `Hearing Amplification` → `Hearing Impairment`, both `BYPASSES`, both as bare names. I checked with `dismech.graph.build_causal_graph` that both edges resolve and `orphan_targets` is empty rather than trusting the syntax.

Neither link carries evidence, and each treatment's `notes:` now says why. The only outcome literature is general microtia practice; quoting it here would attribute to this disease something no source says about it. `BYPASSES` rather than `RESTORES` for the same reason — the developmental window closed before birth and neither intervention touches the lesion.

**Suggestion — `animal_models#Mouse`.** Taken, but by making it precise rather than by dropping it: it now names `Hoxa2 null mouse (Rijli 1993)` and `Hoxa-2 null mouse (Gendron-Maguire 1993)`, the two models the mismatch is actually between. Both resolve under `check-entity-refs`.

**Suggestion — bilaterality. Taken, but not as a qualifier.** It is on the phenotype's `preferred_term`, which now reads `Bilateral microtia` against the `HP:0008551` binding. I looked for the qualifier route first and decided against it: HP has no live bilateral-microtia term (`HP:0040111` is obsolete), no dismech entry anywhere uses a laterality predicate-value qualifier so there is no precedent to copy, and CLAUDE.md records that qualifier terms are invisible to `validate-terms` and have previously concealed 32 wrong bindings. Carrying the specificity in `preferred_term` is the pattern the Ontology Term Contract endorses for exactly this case. It is less queryable than you wanted; I would rather say so than invent a predicate.

**Validation.** `validate`, `validate-terms`, 57/57 snippets verified, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` all pass on the file. Whole-KB `check-folded-hyphens`, `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` report no new divergences. `check-source-defect-claims` reports nothing for this entry. `validate-history` clean on the new record. `qc-deep-research` shows 11 of the report's citations unconsumed, which is an advisory census and well inside the normal range for the corpus. No `cache/**/*.csv` changed.

One note on the diff: `references_cache/PMID_31567444.md` and `references_cache/DOI_10.1097_mcd.0000000000000297.md` are the same Meddaugh paper. The DOI form came from the DR run's citation list; I fetched the PMID form separately so the "no abstract" claim in `notes:` is checkable against a cache rather than being my word for it.

kevinschaper:
Thanks — the blocking item is fixed, plus two of the suggestions. One push.

**Cleft palate repair (blocking).** Added a `Cleft Palate Repair` treatment. It is bound to `NCIT:C168380` **Palatorrhaphy**, which turned out to need no research: it is already in `cache/ncit/terms.csv` and in the `treatmentactionterm` enum cache, and four entries already use it for cleft repair (`Kniest_Dysplasia`, `Treacher_Collins_Syndrome`, `Apert_Syndrome`, `Cardiofacioneurodevelopmental_Syndrome`). `therapeutic_modality: SURGERY`, with a `target_mechanisms` link to the `Cleft Palate` phenotype using `BYPASSES` — the same shape as auricular reconstruction, because palatoplasty closes the cleft after the fact and leaves the prenatal arch-identity lesion untouched.

It carries no evidence, with the reason stated in `notes`. I grepped all seven cached references for management content and there is none; quoting the general cleft-palate literature would attribute something to this disease that no source says about it. That is the same footing you noted auricular reconstruction stands on.

**One part of that request I could not do as asked.** You asked for the treatment to be subtype-scoped to AR-MHIC. `Treatment` has no `subtype` slot in the schema, and I checked every disorder file — no treatment anywhere in the KB carries one. So the restriction is recorded in the treatment's `notes` instead, saying which subtype it applies to and why it is prose rather than a field. If a structured scope is wanted, that is a schema change and a separate PR.

**Mixed Hearing Impairment orphan (suggestion, taken).** Wired with a `downstream` edge from `Homeotic Transformation of Second-Arch Skeletal Derivatives`, parallel to the existing edge to `Hearing Impairment`. I preferred that over a `sequelae` edge from `Hearing Impairment`, since the mixed characterisation is a subtype-specific refinement rather than a consequence of the general phenotype. The edge description says outright that only the conductive half follows from this node and that the sensorineural half is still unexplained — which is what the `hoxa2_sensorineural_component_unexplained` discussion is for. Its evidence reuses the PMID:23775976 middle-ear-development sentence already on the sibling edge, graded identically.

**Cochlear aplasia and Marx grade (suggestions, taken).** Both added to the entry's "not curated for want of a citable source" list, with the reason: the falcon report sources the unilateral cochlear absence to Alasti's 2009 thesis, which has no PMID and no DOI, so it can be neither cached nor quoted.

**Not changed.** BMP4/BMP5/TWSG1 remain unmodeled, as you allowed.

**Validation.** On the changed file: `validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` and `validate-disorders` all pass. `count-verified-snippets` is 58/58 with nothing skipped by prefix — the entry cites no DOIs, so every snippet is machine-checked. Whole-KB `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` are clean with no new baseline rows.

**Orphan sweep.** I diffed every `downstream`, `sequelae`, `target_mechanisms` and `modeled_mechanisms` target against the entry's node names: zero dangling targets, and zero phenotypes without an incoming causal edge — the orphan list is empty now that Mixed Hearing Impairment is connected.

The `cache/` and `references_cache/` trees are untouched by this push; the diff is the entry plus one history record.

kevinschaper:
## Fixed: the cleft-palate mechanism was invented

The reviewer is right, and I confirmed all three of the points before changing anything.

The treatment block said the cleft was caused by "ectopic second-arch cartilage obstructing elevation of the palatal shelves". Nothing in the entry supports it. I grepped all nine cached references for palate, cartilage, shelf, shelves, tongue, hyoglossus and hyoid: the only near neighbour is PMID:7903600's duplicated Meckel's cartilage beside the otic capsule, which is a middle-ear finding and says nothing about the palate, plus that paper's statement that the mouse null is born with cleft palate, which is the fact and not the mechanism. The falcon report does not contain the claim either; its own wording is the hedged "abnormal palate development provides a route to cleft palate". The clause is gone.

## What replaced it

Two mouse accounts of the Hoxa2 cleft are published, and they are not the same claim, so the entry now carries both rather than picking one.

**Barrow and Capecchi 1999 (PMID:10529419)**, fetched with `just fetch-reference`, is curated as two nodes so that the observation and the interpretation carry separate evidence:

- *Mis-Attachment of the Extrinsic Tongue and Hyoid Muscles* — "We also uncovered multiple defects in the attachments and trajectories of the extrinsic tongue and hyoid muscles in Hoxa2 mutants." and "the abnormal trajectory of two of these muscles, the styloglossus and the stylohyoideus, blocked the attachment of the hyoglossus to the greater horn of the hyoid, which in turn correlated exactly with the presence of cleft palate in Hoxa2 mutants".
- *Abnormal Tongue Posture Obstructing Palatal Shelf Closure* — the "We suggest that the hyoglossus..." sentence, quoted whole so the word "suggest" sits in the evidence rather than only in my prose.
- The edge between them is carried by the rescue: "We show that two compensatory defects, associated with the loss of Hoxa1 function, restore normal attachment of the hyoglossus to the greater horn thereby allowing the palatal shelves to lift and fuse above the flattened tongue." The edge to the phenotype is carried by "Unexpectedly, in Hoxa1/Hoxa2 double mutants, the penetrance of cleft palate is dramatically reduced."
- Added the Hoxa1/Hoxa2 double-mutant mouse as an animal model with two `RESCUES` links and their readouts.

**Smith et al. 2009 (PMID:19653318)**, which I found searching PubMed for `HOXA2 AND palatogenesis`, is the reason I did not simply swap one confident mechanism for another. Those authors set out to test the tongue account and conclude against its sufficiency: "These results demonstrate the cleft palate phenotype of Hoxa2 null embryos is not solely due to abnormal tongue musculature, and indicate a direct role of Hoxa2 in regulating murine palatogenesis." That is curated as *Intrinsic Palatal Shelf Fusion Defect*, hanging off HOXA2 Loss of Function, with Hoxa2 expression in the palate and the tongue-free organ-culture fusion result (graded `IN_VITRO`, since it is explant culture).

A `CONTROVERSY` discussion, `hoxa2_cleft_palate_tongue_versus_intrinsic_palate`, records that both accounts are mouse, that they are compatible, that their relative contribution has not been quantified, and that neither has been examined in a HOXA2-mutant patient. The entry asserts neither as the mechanism of the human cleft.

```
HOXA2 Loss of Function ──> Loss of 2nd Arch NC Identity ──> Mis-Attachment of Tongue/Hyoid Muscles
         │                                                            │
         │                                                            v
         │                                          Abnormal Tongue Posture Obstructing Shelf Closure
         │                                                            │
         └──> Intrinsic Palatal Shelf Fusion Defect ──> Cleft Palate <┘
```

## One edge removed

The `Homeotic Transformation of Second-Arch Skeletal Derivatives -> Cleft Palate` edge is gone. No source links the skeletal transformation to the cleft, and the correct route now runs through the muscle chain. Its evidence — that the mouse null is born with cleft palate — is a genotype-level co-occurrence, so it moved to the Cleft Palate phenotype, where the description already said it. Nothing was deleted without a home.

## Where the treatment link sits

The reviewer suggested pointing the treatment's `target_mechanisms` link at the new node. I kept it on the `Cleft Palate` phenotype and described the two mechanism nodes in the link text instead. Reason: auricular reconstruction in this same entry deliberately links at the phenotype and says so, and the surgery acts on the manifestation, not on either prenatal mechanism. The rationale is recorded in the treatment `notes:`. Happy to move it if you would rather have consistency with the suggestion than with the sibling treatment.

## Other prose I checked

I read every `description`, `notes`, `limitations`, `explanation` and `interpretation` field in the entry and asked which reference each claim comes from. Four more things needed fixing, all small:

- The treatment notes said "none of the seven cited references" — there are nine now.
- A model readout interpretation read "Most double mutants close the palate", which turns "penetrance is dramatically reduced" into a majority claim the abstract does not make. Reworded, noting the abstract gives no figures.
- The HMX1 enhancer node stated the Hoxa2-Hmx1 enhancer binding without the source's "In mouse" qualifier. Restored.
- `CL:0000008` carried `preferred_term: cranial neural crest cell` against the label `migratory cranial neural crest cell` — broader than the term, the inverse of the allowed pattern. Set to the ontology label.

On the two other suggestions: I added a paragraph to `notes:` recording that the falcon report argued against lumping the dominant families and why that was not followed. `prevalence[1]` already carries `population: All causes of microtia, worldwide` with a note saying it is context and not this disease's rate, so I left it.

Searches behind the correction, all recorded in `notes:` — PubMed `HOXA2 AND hyoglossus` returns PMID:10529419 alone; `HOXA2 AND tongue muscle` returns that plus an overexpression study; `HOXA2 AND palatogenesis` returns three, all mouse. Nothing describes tongue, hyoid or palatal-shelf anatomy in a HOXA2-mutant human, which is why both new nodes say so rather than asserting the mechanism of the human cleft.

## Validation

`validate` (schema, terms, references — 77/77 snippets verified), `validate-terms`, `count-verified-snippets`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `validate-disorders`, and the whole-KB `check-folded-hyphens`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading` — all pass. No DOI-cited snippets, so nothing was skipped by prefix.

I also checked the pathograph by hand: no dangling bare-name targets, every phenotype is the target of some edge, and the only node without an incoming edge is the root. The `cache/` diff is 10 rows in sorted order for the five new GO/UBERON terms. A history record is included.

Issue #10785, opened 2026-09-03
PAICS deficiency is claimed as MONDO:0859003 and stubbed as MONDO:0859244, so the claim check offers it as unclaimed


gingin77:
One disease, two MONDO terms, and the claim lock only knows one of them.

| Where | MONDO | Label |
|---|---|---|
| Open claim #10471 | `MONDO:0859003` | PAICS deficiency |
| `stubs/Phosphoribosylaminoimidazole_Carboxylase_Deficiency.yaml` | `MONDO:0859244` | phosphoribosylaminoimidazole carboxylase deficiency |

These are the same enzyme deficiency under its abbreviation and its expansion. PAICS *is* phosphoribosylaminoimidazole carboxylase / phosphoribosylaminoimidazolesuccinocarboxamide synthase.

`just next-unclaimed` matches the MONDO ID out of claim-issue titles, so a claim keyed on `MONDO:0859003` does not suppress a stub keyed on `MONDO:0859244`. The stub was offered to me in a `/claim-disease 5` pool today; I skipped it only because the label looked familiar from the claims list. An agent that took it would have duplicated #10471's work.

**This is the inverse of #10113.** That issue is about several entries sharing one `disease_term`; this is one disease holding two terms, and it bites in the queue rather than in `kb/`. Neither check sees the other's shape.

## What needs deciding first

Which term is right. That is a person's call, not a tooling one:

- if the two MONDO classes are genuine duplicates, this is an upstream MONDO merge request and dismech should follow whichever survives;
- if MONDO considers them distinct for a reason not obvious from the labels, then both the claim and the stub are correct and the queue is right to hold both.

I have not touched either the stub or the claim, since deleting the wrong one loses real work.

## Cheap mitigation regardless of that answer

`check-claims` already parses claim-issue titles. It could additionally compare each claim's **label** against stub labels and synonyms, and report near-matches whose MONDO IDs differ. That is a string comparison over two small lists, needs no ontology query, and would have surfaced this pair — the stub carries no synonyms today, but the claim's title text "PAICS deficiency" appears nowhere in the stub, so matching would need to run the other way, on the expansion. An abbreviation-aware comparison is doing real work here, so treat this as advisory output for a human rather than anything automatic.

Found during the claim run that filed #10762 through #10766.

Issue #9415, opened 2026-08-24
`next-stubs` / `next-unclaimed` hand out stubs that `check-stubs` already knows are curated


matentzn:
## What happened

Following the `claim-disease` skill workflow verbatim on a 5-disease claim run, step 3 (`just next-unclaimed 25`) offered me two diseases that are **already curated on `main`**:

| Stub | MONDO | Already in `kb/disorders/` |
|---|---|---|
| `stubs/Isolated_Growth_Hormone_Deficiency_Type_IA.yaml` | MONDO:0009876 | `Isolated_Growth_Hormone_Deficiency_Type_IA.yaml` |
| `stubs/Immunodeficiency_85_And_Autoimmunity.yaml` | MONDO:0030428 | `Immunodeficiency_85_And_Autoimmunity.yaml` |

Same MONDO ID, same filename, same checkout. `just check-stubs`, run in the same working tree seconds later, reports both correctly:

```
[advisory] Immunodeficiency_85_And_Autoimmunity.yaml: already_curated: MONDO:0030428 is
  covered by disorders/Immunodeficiency_85_And_Autoimmunity.yaml — stale, tidy up
[advisory] Isolated_Growth_Hormone_Deficiency_Type_IA.yaml: already_curated: MONDO:0009876 is
  covered by disorders/Isolated_Growth_Hormone_Deficiency_Type_IA.yaml — stale, tidy up
```

So the information exists and is computed. The command a curator is told to pick work from just doesn't consult it.

## Why

`next_command` in `src/dismech/stubs/cli.py` filters on exactly three things:

```python
stubs = [
    s
    for s in load_stubs(stub_dir)
    if s.status == "OPEN" and s.entry_type in {"UNDECIDED", "DISEASE"}
]

claimed = _load_claim_index(claims_path)
if claimed is not None and not include_claimed:
    stubs = [s for s in stubs if s.mondo_id not in claimed]
```

`status`, `entry_type`, and the claim index. The staleness classification lives a couple of hundred lines further down in the same file — `STALE_KINDS = ("already_curated", "obsolete_term")` — and is wired into `check-stubs` and `tidy-stubs` but never into `next`.

## Why it matters more than it looks

[...]

Issue #712, opened 2026-03-20
Validate against obsolete ontology terms in caches and KB files


cmungall:
## Specific issue: CTCF-related Neurodevelopmental Disorder uses obsolete Mondo term

`kb/disorders/CTCF-related_Neurodevelopmental_Disorder.yaml` currently has:

```yaml
disease_term:
  preferred_term: CTCF-related neurodevelopmental disorder
  term:
    id: MONDO:0700294
    label: obsolete CTCF-related neurodevelopmental disorder
```

`MONDO:0700294` is now obsolete in the current Mondo release. This needs to be replaced with its successor term.

Surfaced during PR #710 cache normalization — the regenerated `cache/mondo/terms.csv` correctly shows the "obsolete" prefix, and the agent dutifully updated the label to match. But the right fix is to find the replacement term, not use an obsolete one.

Similarly, `cache/uberon/terms.csv` contains `UBERON:0000953,obsolete visceral muscle` — though this one is no longer referenced in any KB file (fixed in #710).

## Broader issue: obsolete terms should not pass validation

If dismech uses LinkML dynamic enums backed by ontology term caches, obsolete terms should ideally be **excluded from the enum entirely** or at minimum **flagged during validation**. Currently:

1. Cache regeneration pulls all terms from the ontology, including obsolete ones
2. The dynamic enum accepts any term in the cache
3. A KB file can reference an obsolete term and pass validation as long as the label matches the cache

### Proposed fixes

1. **Cache generation**: Filter out terms with "obsolete" in the label (or use the ontology's deprecation metadata) when building term caches. This way obsolete terms simply aren't valid enum values.

2. **Validation rule**: Add a validate-terms check that flags any `term.label` starting with "obsolete " as an error, regardless of cache membership. Belt and suspenders.

3. **Migration tooling**: When a cache refresh removes a previously-valid term (because it became obsolete), the validation should report which KB files reference the now-invalid term and suggest successors if available from the ontology's `term_replaced_by` metadata.

This would catch cases like the CTCF disorder proactively rather than discovering them during manual review.

## Related

- PR #710 term change audit comment lists all CURIE/label changes
- LTV (linkml-term-validator) controls cache generation — this may be an LTV feature request
- dismech #711 (biolink-model optional dependency)

PR #10918, opened 2026-09-04
Read stub-term obsolescence from MONDO instead of guessing from the label

gingin77:
Refs #10785. This is the exact, ontology-backed half of that issue; the label/synonym near-match comparison it also proposes is deliberately not here (see "Not in this PR"), so the issue stays open.

## What this fixes

The stub queue's only way of noticing that MONDO had retired a term was to look at the stub's `label` and see whether it started with the word "obsolete". That works only if the list a stub was seeded from happened to capture the label *after* MONDO prefixed it. Stubs are normally seeded from lists that predate the retirement, so they keep the clean label forever and the check stays silent.

**It matches zero of the 1,333 committed stubs**, while three of them name terms MONDO has already ruled on. The `obsolete_term` branch of `check-stubs` and `tidy-stubs` was dead code.

This replaces it with the ontology's own answer, and keeps the label pattern as a backstop.

## What is now read from MONDO

`scripts/enrich_curation_stubs.py` already opens the MONDO SQLite and walks every stub ID, so these facts are read in the same pass and written into the stub file. That is what lets `just check-stubs` report them offline — CI never needs the 1.2 GB ontology build.

| Field | From | Means |
|---|---|---|
| `mondo_obsolete` | `owl:deprecated true` | the term is retired |
| `mondo_replaced_by` | `IAO:0100001` | which live term absorbed it |
| `mondo_obsoletion_candidate` | MONDO's `obsoletion_candidate` subset + its own note | MONDO has decided to retire it but has not yet done so |

The third one is not in the issue's scope list, and it is the one that catches things. **The deprecation flag only appears in a MONDO release built after the merge lands, which can be months later.** MONDO puts a term in its `obsoletion_candidate` subset the moment the decision is taken, with a comment naming the survivor. Without it this PR would ship a check that reports nothing and cannot report anything until upstream rebuilds — see the census below.

## Census over the queue

```
$ just stub-obsolescence
mondo: 2026-05-05  stubs: 1333
obsolete mondo_id: 0  scheduled for obsoletion: 3

  Autosomal_Dominant_Optic_Atrophy.yaml
    MONDO:0020250 is scheduled for obsoletion with no replacement proposed
  Epilepsy_Of_Infancy_With_Migrating_Focal_Seizures.yaml
    MONDO:0100025 is scheduled for obsoletion; MONDO will replace it with MONDO:0017385
  Phosphoribosylaminoimidazole_Carboxylase_Deficiency.yaml
    MONDO:0859244 is scheduled for obsoletion; MONDO will replace it with MONDO:0859003
```

So: **0 of 1,333 stubs carry a `mondo_id` MONDO has deprecated, and 3 carry one MONDO has scheduled for retirement** — including `MONDO:0859244`, the stub this issue is about. The other two are new: `MONDO:0100025` will merge into `MONDO:0017385`, and `MONDO:0020250` is being retired as a grouping class with no successor.

## One correction to the issue thread

The earlier comment says `data/mondo/MANIFEST.yaml` is stale and needs repinning. I do not think that is right, and it matters for what this check can do.

I downloaded the current `mondo.db.gz` that `just fetch-ontology-dbs mondo` fetches. Its sha256 is `61b2c87f…89a3fb`, which is byte-for-byte the value the manifest already pins, and its `versionIRI` is `2026-05-05`. The pin is not behind the available build — **the upstream semantic-SQL build itself has not refreshed since 2026-05-05**, so repinning today would change nothing.

That is why `MONDO:0859244` still carries a scheduled-merge note rather than a deprecation flag, even though [mondo#9884](https://github.com/monarch-initiative/mondo/issues/9884) closed on 2026-05-21. Repinning is still worth doing when upstream rebuilds, and re-running the full enrichment still belongs in its own PR — the point is only that it is not the blocker the thread assumed.

## Merged terms are repointed, not deleted

A term that was merged into a live one still names a disease nobody has curated. The identifier moved; the work did not. Deleting the stub would drop real queue content, so the merge case is reported as a distinct kind:

| Finding | Severity | What `tidy-stubs --apply` does |
|---|---|---|
| `obsolete_term` — retired, no successor | advisory | deletes it |
| `obsolete_term_replaced` — merged into a live term | advisory | lists it; repoint `mondo_id` |
| `obsoletion_candidate` — scheduled, still live | advisory | lists it |

All three stay advisory. Staleness never gates, for the reason in `docs/curation-stubs.md`: an unrelated curation PR merging on `main` must not turn every open stub PR red.

## Not in this PR

- **`stubs/Phosphoribosylaminoimidazole_Carboxylase_Deficiency.yaml` is untouched**, and so is claim #10471. @gingin77 holds that claim and has said the curation PR will repoint the stub.
- **No manifest repin and no full `enrich-stubs` run.** A re-enrichment rewrites blocks across the whole queue and belongs on its own; the mechanism here is in place and the fields populate on that pass. As above, there is nothing to repin to yet.
- **No label/synonym near-match comparison.** That is the heuristic half of the issue — catching two *live* MONDO terms that are the same disease needs abbreviation awareness ("PAICS" ↔ "phosphoribosylaminoimidazole carboxylase") and should stay advisory-for-a-human, as filed.

## Related

- **#712** asked for `term_replaced_by`-driven successor suggestions on the `kb/` side. Same mechanism, applied to `stubs/` here.
- **#9415** is why detection alone will not stop the offer: `next-stubs` / `next-unclaimed` do not apply `STALE_KINDS`, so a stub still gets handed out after it is flagged. That is unchanged here and still needs fixing.

## Verification

```
uv run pytest tests/test_stubs.py     # 1401 passed
uv run ruff check / ruff format --check
just check-stubs                      # OK: 1333 stubs, no errors (41 advisory)
```

Eight new tests in `tests/test_stubs.py` cover the SQLite reader against a fixture database rather than the live ontology (the real build's contents move — `MONDO:0859244` is a candidate today and will be plainly obsolete later, so asserting either against the real file schedules a failure), plus the offline check, the repoint-not-delete split, and graceful degradation when the MONDO build is absent.

End to end, on a copy of three stubs rather than the committed queue:

```
$ uv run python scripts/enrich_curation_stubs.py --stub-dir /tmp/queue
mondo: 2026-05-05  stubs: 3  ...  obsolete: 0  obsoletion candidates: 3  changed: 3

$ uv run dismech-stubs check --stub-dir /tmp/queue
[advisory] Phosphoribosylaminoimidazole_Carboxylase_Deficiency.yaml: obsoletion_candidate:
  MONDO:0859244 is scheduled for obsoletion by MONDO: This term is scheduled to be merged
  with MONDO:0859003 PAICS deficiency ... https://github.com/monarch-initiative/mondo/issues/9884
```

Without the MONDO build, `just stub-obsolescence` prints where it looked and exits 0.

---

https://claude.ai/code/session_01XnPnio7pPLDYxX2u9xteQb

---
_Generated by [Claude Code](https://claude.ai/code/session_01XnPnio7pPLDYxX2u9xteQb)_

### DIFF
--- a/docs/curation-stubs.md
+++ b/docs/curation-stubs.md
@@ -105,12 +105,20 @@ Enrichment also asks MONDO whether it has retired the stub's term, or decided
 to, and writes the answer into the file so the check that reads it needs no
 ontology:
 
+A term MONDO has already retired is no longer *scheduled* to be, so a stub
+carries one case or the other, never both. Already retired, and merged into a
+survivor:
+
 ```yaml
 mondo_obsolete: true
 mondo_replaced_by: MONDO:0859003
-mondo_obsoletion_candidate: >-
-  This term is scheduled to be merged with MONDO:0859003 PAICS deficiency ...
-  https://github.com/monarch-initiative/mondo/issues/9884
+```
+
+Scheduled but still live, carrying MONDO's own note verbatim (emitted as one
+long line — the writer does not fold):
+
+```yaml
+mondo_obsoletion_candidate: This term is scheduled to be merged with MONDO:0859003 PAICS deficiency, based on the fact that the concept of these 2 terms are the same. This ID will therefore be obsoleted and replaced with MONDO:0859003 https://github.com/monarch-initiative/mondo/issues/9884
 ```
 
 The signal this replaces was a heuristic on the label. MONDO prefixes a retired
--- a/scripts/enrich_curation_stubs.py
+++ b/scripts/enrich_curation_stubs.py
@@ -45,11 +45,15 @@ from pathlib import Path
 ROOT = Path(__file__).resolve().parents[1]
 sys.path.insert(0, str(ROOT / "src"))
 
-from dismech.stubs.obsolescence import TermStatus, read_statuses
+from dismech.stubs.obsolescence import TermStatus, default_mondo_db, read_statuses
 from dismech.stubs.seed import yaml_scalar
 from dismech.yaml_io import safe_load
 
-DEFAULT_MONDO_DB = Path.home() / ".data" / "oaklib" / "mondo.db"
+#: Resolved the way OAK resolves `sqlite:obo:mondo` — through pystow, which
+#: reads PYSTOW_HOME. Hardcoding `~/.data` meant a contributor who sets it got
+#: `just stub-obsolescence` reading the ontology while `just enrich-stubs`
+#: reported the same database missing.
+DEFAULT_MONDO_DB = default_mondo_db()
 HGNC_CACHE = ROOT / "cache" / "hgnc" / "terms.csv"
 #: Records which MONDO release the committed enrichment came from, so diff churn
 #: in `mondo_descendant_count` across machines is attributable. Mirrors
@@ -152,7 +156,7 @@ def render(parents, descendants, total, genes, status=None) -> list[str]:
         # either way.
         if status.obsolete:
             lines.append("mondo_obsolete: true")
-        if status.replaced_by:
+        if status.obsolete and status.replaced_by:
             lines.append(f"mondo_replaced_by: {status.replaced_by}")
         if status.obsoletion_candidate and status.obsoletion_note:
             lines.append(
--- a/src/dismech/stubs/__init__.py
+++ b/src/dismech/stubs/__init__.py
@@ -16,13 +16,6 @@ from .claims import (
     parse_claims,
     unkeyed_claims,
 )
-from .obsolescence import (
-    ObsolescenceIndex,
-    TermStatus,
-    default_mondo_db,
-    load_obsolescence,
-    read_statuses,
-)
 from .model import (
     CoverageIndex,
     Stub,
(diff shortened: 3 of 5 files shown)


gingin77:
Fixed the lint gate and took all five optional items, in `7d16512`.

**The lint failure was mine, and so was the reason I missed it.** I ran `uv run ruff check`, which resolved the venv-pinned 0.15.8; CI's `uvx ruff check .` got 0.16.6, where RUF022, FURB188 and ISC004 all report. Your note about the unpinned default rule set is exactly the trap I fell into. `uvx ruff@latest check .` is now clean, and I used it for this push.

Three fixed by `--fix`; ISC004 needed the two-part comment string wrapped in `(...)`.

**The five optional items:**

1. **`PYSTOW_HOME`** — `DEFAULT_MONDO_DB` now calls `default_mondo_db()`. Confirmed the failure you described was real: `PYSTOW_HOME=/nonexistent uv run python scripts/enrich_curation_stubs.py` previously looked in `~/.data` regardless, and now reports `/nonexistent/oaklib/mondo.db`.
2. **Non-conforming `term_replaced_by`** — `normalize_term()` now returns `""` for anything that is not `MONDO:NNNNNNN`, so a successor outside MONDO is dropped at read time rather than written into a stub that fails `just validate-stubs` later.
3. **The folded docs example** — you're right, `yaml_scalar` never folds. Split into the two cases, since a stub carries `mondo_obsolete` or `mondo_obsoletion_candidate` and never both.
4. **Triple over prose** — `proposed_replacement` now returns `replaced_by` when `IAO:0100001` states it, and only falls back to parsing the note. `replaced_by` is populated whenever MONDO states the triple, but `mondo_replaced_by` is still only *written* alongside `mondo_obsolete: true`, so the slot's schema description stays accurate.
5. **Several `rdfs:comment` rows** — all rows are now collected and joined instead of `dict()` keeping the last. This one could genuinely have dropped the scheduled-obsoletion note on a term with an unrelated editor comment.

Items 4 and 5 have a test each (`test_the_replacement_triple_beats_the_scheduled_merge_prose`), and item 2 extends `test_term_iri_and_curie_forms_both_normalize`.

**Verification on the new head:** `uvx ruff@latest check .` clean; `ruff format --check` clean on the touched files; `pytest tests/test_stubs.py` 1402 passed; `dismech-stubs check` OK (1333 stubs, 41 advisory); `dismech-stubs obsolescence` unchanged at 0 obsolete / 3 scheduled; `enrich-stubs --dry-run` still reports 35 changed (32 pre-existing drift plus the 3 candidates).

---

Issue #10847, opened 2026-09-04
Curate X-linked spermatogenic failure 1 (MONDO:0056795)


gingin77:
Curate a dismech entry for **X-linked spermatogenic failure 1** ([MONDO:0056795](https://monarchinitiative.org/MONDO:0056795)) — SPGFX1, caused by TEX11 variants, is one of the best-established monogenic causes of non-obstructive azoospermia in men; loss of TEX11 disrupts meiotic crossover formation and chromosome synapsis, arresting spermatogenesis at meiosis I.

**First decide what this should be**: a `kb/disorders/` Disease, a `kb/groupings/` Grouping, a `has_subtypes` entry on an existing disease, or out of scope. See the Disease Groupings section of `CLAUDE.md` and #8727 for the rule — a disorder entry needs one reasonably conserved pathograph. Record the decision in the stub's `entry_type` before curating.

This should be a `DISEASE`, and it slots into an established pattern: the KB already has gene-specific meiotic-arrest entries (`SYCP2-related_spermatogenic_failure`, `SYCP3-related_spermatogenic_failure`, `HFM1-related_gametogenic_failure`, `MCM8`/`MCM9`) collected by the `kb/groupings/Meiotic_Gametogenic_Failure.yaml` grouping, so the new entry should be added to that grouping's members. Worth capturing that the meiotic-arrest histology predicts failed testicular sperm extraction, which is the clinically actionable part.

Stub: `stubs/X-linked_Spermatogenic_Failure_1.yaml`
Nominated by: rare-disease-identification

gingin77:
**This claim names the wrong MONDO term, and I'm redirecting it to [#10920](https://github.com/monarch-initiative/dismech/issues/10920).**

I claimed MONDO:0056795 intending to curate TEX11-related spermatogenic failure. MONDO:0056795 is not that disease:

| Entry | OMIM | MONDO | Gene |
|---|---|---|---|
| SPGFX**1** | 305700 | MONDO:0056795 | **none mapped** |
| SPGFX**2** | 309120 | MONDO:0010647 | **TEX11** (Xq13.1) |

[...]

PR #10921, opened 2026-09-04
Curate TEX11-related spermatogenic failure (MONDO:0010647), and correct a mis-targeted claim

### LINKED ISSUE #10920: Curate spermatogenic failure, X-linked, 2 (MONDO:0010647)
Curate a dismech entry for **spermatogenic failure, X-linked, 2** ([MONDO:0010647](https://monarchinitiative.org/MONDO:0010647), OMIM 309120) — SPGFX2, caused by hemizygous TEX11 variants, is one of the best-established monogenic causes of non-obstructive azoospermia. TEX11 is the mammalian ZIP4 orthologue and a subunit of the ZZS complex (SHOC1–TEX11–SPO16), which couples synaptonemal complex assembly to class I crossover formation; its loss produces asynapsis, unrepaired meiotic double-strand breaks, pachytene-checkpoint elimination of spermatocytes, and meiotic arrest.

**This claim supersedes #10847, which named the wrong MONDO term.** I claimed MONDO:0056795 (X-linked spermatogenic failure 1) intending to curate TEX11 disease. That was wrong, and the mistake is worth recording because the naming actively invites it:

| Entry | OMIM | MONDO | Gene |
|---|---|---|---|
| SPGFX**1** | 305700 | MONDO:0056795 | **none mapped** — Sertoli-cell-only / Del Castillo syndrome, a pedigree-defined phenotype |
| SPGFX**2** | 309120 | MONDO:0010647 | **TEX11** (Xq13.1) |

Verified against OLS and OMIM esummary: MONDO:0056795's definition is *"only Sertoli cells ... line the seminiferous tubules and there are not any sperm cells present"*, with synonyms Del Castillo syndrome and germinal cell aplasia, and xref OMIM:305700. Its stub carries no `genes:` block, which is MONDO's way of saying no causal gene — a signal I should have read before claiming.

**Entry type: `DISEASE`.** This slots into an established pattern — the KB already has gene-specific meiotic-arrest entries (`SYCP2-related_spermatogenic_failure`, `SYCP3-related_spermatogenic_failure`, `HFM1-related_gametogenic_failure`, `MCM8`/`MCM9`) and a `kb/groupings/Meiotic_Gametogenic_Failure.yaml` grouping.

Two things to get right, flagged by the review on #10847:

- The grouping's `description` opens *"A group of **autosomal recessive** reproductive disorders"*. TEX11 is X-linked and hemizygous in affected men, so adding it means editing that sentence.
- Do **not** conform this entry to the grouping's ovarian or somatic-cancer-predisposition nodes. TEX11 is male-restricted and meiosis-specific with no reported somatic phenotype.

Worth capturing: the meiotic-arrest histology predicts a poor yield from testicular sperm extraction, so the actionable intervention here is a negative one — withholding or not repeating a procedure — which is an unusual shape to model.

Stub: none exists for MONDO:0010647; `stubs/X-linked_Spermatogenic_Failure_1.yaml` is the mis-targeted one and is dispositioned separately.

gingin77:
## The claim named the wrong disease

I claimed MONDO:0056795 ("X-linked spermatogenic failure 1") intending to curate TEX11-related azoospermia. That is a different disease, and the naming actively invites the mistake:

| Entry | OMIM | MONDO | Gene |
|---|---|---|---|
| SPGFX**1** | 305700 | MONDO:0056795 | **none mapped** — Sertoli-cell-only / Del Castillo syndrome, pedigree-defined |
| SPGFX**2** | 309120 | MONDO:0010647 | **TEX11** (Xq13.1) |

Verified two ways rather than assumed. OLS returns MONDO:0056795's definition as *"only Sertoli cells ... line the seminiferous tubules and there are not any sperm cells present"*, with synonyms `Del Castillo syndrome` and `Germinal cell aplasia` and xref `OMIM:305700`. The OMIM esummary for both IDs confirms 305700 and 309120 are separate current entries titled SPGFX1 and SPGFX2.

The signal that should have stopped the claim was already in the stub: `stubs/X-linked_Spermatogenic_Failure_1.yaml` carried **no `genes:` block**, and per `CLAUDE.md` an absent block is MONDO saying there is no causal gene — a positive statement, not missing data. A gene-less stub for something about to be curated as a single-gene disease is a contradiction I walked past.

Claim #10847 is closed as superseded; #10920 claims the correct term and this PR closes it.

## The entry

TEX11 is the mammalian Zip4 orthologue and a subunit of the ZZS complex (SHOC1–TEX11–SPO16), which couples synaptonemal complex assembly to class I crossover formation. It also binds SYCP2 in the synaptonemal complex lateral element — the proposed physical link between synapsis and recombination, and a real mechanistic bridge to the existing `SYCP2-related_spermatogenic_failure` entry.

```
TEX11 Loss of Function                                   MOLECULAR
  -> Chromosomal Asynapsis and Reduced Crossover Formation    CELLULAR
       -> Pachytene Checkpoint Elimination of Spermatocytes   CELLULAR
       -> Meiosis I Nondisjunction in Escaping Spermatocytes  CELLULAR
            -> Spermatogenic Arrest at Meiosis I              TISSUE
                 -> Non-obstructive azoospermia
```

**The two loss points are curated as separate nodes on purpose.** The source distinguishes them explicitly — spermatocytes with asynapsed autosomes die at pachytene, while those with only asynapsed sex chromosomes progress and then fail at the first division — and that split is why the histology is mixed rather than a single uniform block. Collapsing it into one "meiotic arrest" node would lose the reason biopsies in this disease show some Sertoli-cell-only tubules alongside arrested ones.

`conforms_to` three nodes of `meiotic_prophase_failure`. **Deliberately not** the ovarian follicle depletion node or the somatic DNA-repair/cancer node — TEX11 is male-restricted and meiosis-specific with no reported somatic phenotype.

## A REFUTE item, and the discussion it grounds

Missense TEX11 variants occur in fertile men too, so gene identity does not establish pathogenicity. That has been tested directly: of three human missense alleles modelled in transgenic mice, one behaved as a candidate infertility allele and **two were non-causative**. That is curated as `supports: REFUTE` against the general proposition, which gives this entry a real negative result rather than an all-SUPPORT evidence set.

It also sets up a `HUMAN_MODEL_MISMATCH` discussion. The mouse is currently the only functional assay applied to these variants and its negative results are being used to reclassify human ones, but nothing establishes that murine spermatogenesis tolerates a given substitution the way human spermatogenesis does. A negative mouse result may mean the allele is benign, or that the mouse is more permissive.

## A treatment that is a negative intervention

The actionable consequence of a TEX11 diagnosis is *withholding* surgery. Meiotic arrest predicts poor yield from testicular sperm extraction, and TEX11 is one of only three genes with enough reported retrieval-negative cases to act on, so a molecular diagnosis can spare a man a procedure unlikely to recover usable sperm — or spare him a repeat of one that already failed. Modelled as a treatment because it is a management decision, unusual shape notwithstanding.

## Grouping, including a pre-existing error

Added to `kb/groupings/Meiotic_Gametogenic_Failure.yaml`. Its `description` opened *"A group of **autosomal recessive** reproductive disorders"*, which TEX11 breaks — it is X-linked and hemizygous in affected men. That claim was **already wrong** before this entry, since SYCP3 is a dominant-negative heterozygous counterexample. The description now says inheritance is autosomal recessive for most members but is not a membership criterion. `just check-groupings` reports TEX11 SATISFIED on the necessary-and-sufficient criterion.

## Stub disposition

`stubs/X-linked_Spermatogenic_Failure_1.yaml` is deleted. SPGFX1 has no mapped gene and no established molecular mechanism, so there is no pathograph for a mechanism knowledge base to carry — it is not curatable until a gene is found. The reasoning lives in the new entry's `notes` so the concept is not re-nominated as a single-gene disease.

## Deep research

One `claude_code` run, clean validation (17/17 references resolved, 1/1 quoted claim found in source). Its most valuable output was a nomenclature note at the head of the report distinguishing SPGFX1 from SPGFX2 and stating that SPGFX1 has no molecular literature — which is what prompted the verification above. The report is committed under its original `X-linked_Spermatogenic_Failure_1` filename because that is the run that actually happened; the entry it informed carries a different name.

It also corrects one point in the automated research comment on #10847, which read the SCO definition as a legacy artifact of 305700 *"before it was renamed SPGFX1 and assigned to TEX11"*. 305700 was never assigned to TEX11. That comment's biology, ZZS framing, grouping-phrasing catch and TESE literature were all correct and used.

## Validation

`just validate`, `just validate-terms`, `just count-verified-snippets` (17/17), `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-qualifier-terms`, `just check-enum-values`, `just check-stubs`, `just validate-grouping`, `just check-groupings`, `just validate-history` all pass.

Closes #10920

---
_Generated by [Claude Code](https://claude.ai/code/session_01PHYDoqVQvHMJimXjxsprAT)_


gingin77:
Six of the seven items are closed by adding content. The seventh is closed by explaining the omission, and I think that was the right call — details below. Pushed as c3f66aa.

**Item 7 (report rename) was the best catch, and I verified it rather than taking it on faith.** `src/dismech/render.py` globs `research/{disorder_slug}-deep-research-*.md`, so the file named for the pre-correction slug would never have surfaced on the disorder page. Renamed with `git mv`, sidecar included. That is a real defect introduced by the mid-curation retarget, and no gate catches it — worth knowing about for anyone else who retargets an entry after running research.

**1. `diagnosis:`** — karyotype plus AZF microdeletion analysis as the two exclusions that precede gene-level testing, testicular biopsy, and TEX11 sequencing on an NOA panel. The sequencing step records that copy-number analysis belongs alongside sequencing, since the recurrent multi-exon deletion was found by array CGH and exon sequencing alone would miss it.

**2. `histopathology:`** — you were right that the entry argued this at length in prose and never modelled it. Added, with the notes recording that focal Sertoli-cell-only tubules can coexist so such a report does not exclude the diagnosis. I also tied the mixed pattern back to the mechanism rather than leaving it as an observation: spermatocytes whose only asynapsed chromosomes are the sex chromosomes escape the pachytene checkpoint and are lost later, so tubules arrest at different points. And I fetched a second family's biopsy report (PMID:29661171) so the histology rests on two independent series rather than one.

**3. Elevated FSH — not added, and the fix was the other half of your finding.** You noted none of the four cached refs mention FSH; a targeted PubMed search for a TEX11 paper stating it also returned nothing quotable. Rather than bind `HP:0008232` to a source that does not support it, I fixed the actual defect: **the entry's own `description` was asserting a hormone profile it could not cite.** That now attributes the profile to primary testicular failure generally rather than to TEX11, and `notes` records why no phenotype is bound and what it would take. Leaving an unsupported assertion in the description while adding an unsupported phenotype would have made the entry worse in two places instead of better in one.

**4. `animal_models:`** — Tex11-null mouse with two `modeled_mechanisms` links against the asynapsis and pachytene-checkpoint nodes, `RECAPITULATES` / `HIGH` / `CELLULAR`, with `limitations`. Your point stands: the `HUMAN_MODEL_MISMATCH` discussion now has a model object to be about. The link-level evidence for treating the model as informative is taken from the human founding series rather than from the model's own authors.

**5. Treatments — genetic counselling added, micro-TESE/ICSI deliberately not.** Counselling has a specific consequence here worth stating: because the allele is X-linked, a daughter conceived from the father's retrieved sperm is an obligate carrier, so transmission is a foreseeable outcome of the treatment rather than an incidental family finding.

On ICSI I'd push back gently. `PMID_40896145` says sperm were recovered in **none of 17** TEX11 patients. Curating micro-TESE/ICSI as the standard of care these patients receive would assert something the same reference contradicts. The existing treatment already models the retrieval decision, which is what the evidence actually supports.

**6.** `NCIT:C15329` → `NCIT:C94427` Sperm Retrieval, plus a `notes` line saying the `SURGERY` modality tags the intervention being *withheld* — thanks for flagging that it misparses.

**Suggestions taken:** the TESE snippet is now the Discussion sentence (*"sperm were not recovered in any patient carrying LP/P variants in TEX11 (n = 17)..."*), with the LIMITATIONS sentence kept as a second item rather than discarded — you were right that one is a finding and the other a caveat about the study. `GO:0007130` added alongside `GO:0007129`.

**Not taken:** oligozoospermia and decreased testicular size. No cached source for this entry states either, and adding them would repeat exactly the mistake item 3 was about. No `prevalence:` block, for the reason you gave yourself — the three `case_fractions` cover it and every available figure is a cohort share, not a population rate.

**Validation:** `just validate`, `just validate-terms`, `just count-verified-snippets` (**28/28**, up from 17/17), `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-qualifier-terms`, `just check-enum-values`, `just check-groupings` (TEX11 SATISFIED), `just validate-history` all pass. Cache diff is empty — every term now used was already cached.

---

PR #10923, opened 2026-09-04
Curate Good Syndrome (MONDO:0015696)

### LINKED ISSUE #10846: Curate Good syndrome (MONDO:0015696)
Curate a dismech entry for **Good syndrome** ([MONDO:0015696](https://monarchinitiative.org/MONDO:0015696)) — thymoma with adult-onset combined immunodeficiency: near-absent peripheral B cells, hypogammaglobulinemia, and variable T-cell defects, presenting with recurrent sinopulmonary and opportunistic infection.

`DISEASE` looks right, and the interesting curation problem is that the mechanism is genuinely unsettled: the immunodeficiency does not reliably remit after thymectomy, so the thymoma is not straightforwardly upstream of the B-cell loss. Candidate mechanisms (a bone-marrow pre-B arrest, autoantibodies against cytokines and marrow precursors, and a shared underlying defect producing both the thymoma and the immunodeficiency) should be modelled as `mechanistic_hypotheses` groups rather than picking one. `kb/disorders/Thymoma.yaml` exists and covers the neoplasm, not this syndrome, so this is not a duplicate — but the two entries should cross-reference.

Stub: `stubs/Good_Syndrome.yaml`

gingin77:
> **Updated after four review rounds.** Earlier versions of this body made three claims that are no longer true — that treatments were omitted for want of a citable source, that PRCA and myasthenia gravis were deliberately not curated as phenotypes, and that there was no prevalence block. All three were corrected during review and the body below describes the current state. The "Corrections" section records what was wrong and why, because the pattern behind all three is worth keeping.

## The main decision was not to assert a mechanism

Good syndrome is adult-onset combined immunodeficiency occurring with thymoma — near-absent B cells and hypogammaglobulinemia alongside CD4 lymphopenia and an inverted CD4:CD8 ratio, with recurrent sinopulmonary and opportunistic infection, and a concurrent autoimmune disorder in about half of patients. The same patient presents with too little immunity and too much of it at once.

The pathophysiology is genuinely unresolved, and one observation forces that: **the immunodeficiency does not remit after thymectomy.** If the thymoma were simply upstream of the B-cell loss, removing it should be curative, and it is not.

So rather than writing one causal chain and calling it the mechanism, three competing accounts are curated as separate `mechanistic_hypotheses` groups, all `status: ALTERNATIVE`:

| Group | Account | What it explains, and what it doesn't |
|---|---|---|
| `gs_thymic_stromal_origin` | One thymic epithelial lesion — plausibly the recurrent GTF2I mutation — produces both the tumour and the immune failure in parallel | Explains the thymectomy failure without needing a self-perpetuating defect. Rests on a mouse carrying the human allele, not on patients |
| `gs_marrow_b_cell_arrest` | The block is a bone marrow pre-B arrest | Explains the thymectomy failure naturally — surgery does not touch the marrow. Does not explain why a marrow defect accompanies a thymic tumour |
| `gs_autoantibody_mediated` | Autoantibodies against precursors and cytokines drive the cytopenias | Parsimonious: one process, loss of tolerance, explains both halves of a syndrome that otherwise needs two. Co-occurrence of autoimmunity does not establish causation |

**The absence of a `CANONICAL` group is deliberate**, and the `KNOWLEDGE_GAP` discussion says so and sets out what would separate the three: marrow B-cell precursor phenotyping before and after thymectomy, paired autoantibody screening, and GTF2I genotyping of the same patients. They are also not mutually exclusive.

## The thymectomy claim is now cited three times, and twice as a dissociation

This is the load-bearing observation, and in the first version of this entry it was asserted in five places and cited in none. It now carries three citations on `Peripheral B Cell Depletion`, two of which are stronger than a plain negative:

- `PMID:34966753` — in one cohort and systematic review, thymoma-associated pure red cell aplasia remits after thymectomy plus immunosuppression in 61% of patients, *"however, Good's syndrome was unaffected."*
- `PMID:38836142` — within a single patient, ciclosporin corrected the red cell aplasia and *"hypoglobulinemia was not recovered."*
- `PMID:34888060` — the plain statement, in a Good syndrome case report.

A bare negative is compatible with thymectomy simply being a weak intervention. A dissociation — the same operation, in the same cohort, resolving the companion condition and not this one — rules that out.

**The same dissociation now arrives from a second class of intervention.** The immunosuppression treatment's `target_mechanisms` link stops at `Pure red cell aplasia` and deliberately does not reach the B-cell or immunoglobulin nodes, because both sources report the hypogammaglobulinemia did not respond. The graph carries the argument structurally rather than in prose.

## Pathograph

Five mechanism nodes into seven phenotypes. 17 nodes, 16 edges, 0 orphan targets, one connected component plus `GTF2I`:

```
Thymic Epithelial Neoplasia
  -> Disrupted Thymic Epithelial Architecture and Medullary Differentiation
       -> Reduced Thymopoietic Output            -> Recurrent infection
       -> Failed Central Tolerance in the Thymic Medulla
                                                 -> Concurrent autoimmune disorder
                                                 -> Pure red cell aplasia
                                                 -> Myasthenia gravis
       -> Peripheral B Cell Depletion            -> Hypogammaglobulinemia
                                                 -> B lymphocytopenia
  -> Thymoma
```

The thymopoiesis and central-tolerance steps are separate nodes because they lead to opposite clinical problems — one node fanning out to both infection and autoimmunity was carrying two claims.

**`GTF2I` renders isolated, and that is correct.** `graph.py:222-236` suppresses the inferred gene-to-mechanism edge when `relationship_type` is `UNKNOWN`. Connecting it would mean asserting that GTF2I causes this syndrome, which has not been tested. The `genetic:` notes record this so nobody later "fixes" the isolation.

## Content

- **`has_subtypes:`** three prognostic clusters from a 162-patient systematic review, derived by unsupervised hierarchical clustering. `review_notes` records that these are statistical strata, not nosological entities — no ontology term, no MONDO concept, and no `subtype_term` should ever be manufactured for them.
- **`genetic:`** GTF2I, `variant_origin: SOMATIC`, `relationship_type: UNKNOWN`.
- **`animal_models:`** the Gtf2i mutant mouse, `PARTIALLY_RECAPITULATES`, with `limitations` recording that it models thymoma biology and is not reported to develop the B-cell depletion defining the human syndrome — so it supports the upstream half of one hypothesis and is silent on the contested half.
- **`treatments:`** immunoglobulin replacement; thymectomy (`NCIT:C29894`), scoped to oncologic control with its link stopping at the tumour node; immunosuppression for the autoimmune arm (rituximab, ciclosporin).
- **`phenotypes:`** seven, including pure red cell aplasia and myasthenia gravis, both curated **without a frequency band** — the findings are quotable and the numbers are not.
- **`prevalence:`** one `CASES_IN_LITERATURE` record. No population rate has been reported, so none is asserted.
- **`diagnosis:`** serial immunoglobulin monitoring in known thymoma as the route to early diagnosis.

**37/37 snippets verified**, across 8 references.

## Corrections

Three claims in earlier versions of this entry were wrong, and all three had the same cause: **I searched the references the entry already cited rather than the repository-wide cache.**

1. Treatments were omitted on the stated grounds that no cached reference carried a quotable statement about immunoglobulin replacement. `PMID:39180607` was already in the repo, cited by `Thymoma.yaml`.
2. The thymectomy claim was said to be uncitable. Four references already in the cache state it.
3. PRCA and myasthenia gravis were omitted because their frequencies could not be quoted. That argues for omitting the *band*, not the phenotype — which is what the entry already did for B lymphocytopenia.

A fourth correction concerns a source named in review rather than in the entry: `PMID:38122866` was put forward as stating that immune compartments decline *even after thymectomy*. It does not — that clause belongs to the deep-research report's paraphrase. The paper is instead used for what it does establish, a multidecade decline across three haematopoietic lineages, which is the first evidence `gs_marrow_b_cell_arrest` has ever carried.

The entry's `notes` record all of this.

## The research report had a failing quote, and it was rejected

The report's validation flagged a **quote not found in its cited source** — the phrase about disrupting the balance between host-defense and self-tolerance, attributed to `PMID:34113351`. It appears nowhere in this entry.

Worth stating plainly because the summary numbers look reassuring in isolation: the report resolved **28/28** references with a 0.0 confabulation rate, and still had **1 of 2** quotes fail. A clean identifier-resolution rate says nothing about whether the quotes attributed to those identifiers are real.

## Relationship to the thymoma entry

`kb/disorders/Thymoma.yaml` curates the neoplasm; this entry curates the immunodeficiency syndrome occurring with it. Different claims — most people with thymoma never develop Good syndrome, and removing the thymoma usually does not resolve it. Neither entry subsumes the other.

## Deliberately not included

**CD4 lymphopenia and inverted CD4:CD8 frequencies.** The only cached sentence that could carry them names neither, so quoting it would be worse than the omission. `notes` names the fix: a full-text-cached source.

**Canine Good syndrome as a comparative model.** The research report cites a bare PMC identifier and a JAVMA abstract URL; `just fetch-reference` has no source for either.

## Validation

`just validate` (37/37 snippets), `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `check-stubs`, `validate-history` all pass. Cache diff empty.

Closes #10846

### DIFF
--- a/history/disorders/Good_Syndrome/2026-09-04T222156Z-claude-code-d477cb.yaml
+++ b/history/disorders/Good_Syndrome/2026-09-04T222156Z-claude-code-d477cb.yaml
@@ -1,0 +1,39 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Good_Syndrome
+  path: kb/disorders/Good_Syndrome.yaml
+session:
+  id: 2026-09-04T222156Z-claude-code-d477cb
+  timestamp: '2026-09-04T22:21:56Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10923
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - treatments
+  summary: Bind the specific thymectomy term; treat the autoimmune arm
+  details: |
+    Both IMPORTANT items from the third review round on PR #10923. Both were right and both drew on material already in the repository.
+
+    1. Thymectomy was bound to NCIT:C15329 (Surgical Procedure) when NCIT:C29894 (Thymectomy) exists, is already in cache/ncit/terms.csv and the treatmentactionterm enum cache, and is the binding five sibling entries already use — Thymoma, Thymic_Carcinoma, Thymus_Neoplasm, Myasthenia_Gravis and Adult-Onset_Myasthenia_Gravis. The deep-research report offered both the generic term and a thymectomy-specific one, and the generic half is what I took. Rebound; it validates offline against the existing cache with no new rows.
+
+    2. The entry gained pure red cell aplasia and myasthenia gravis as phenotypes in the previous round and then offered no treatment for either. An Immunosuppressive therapy treatment is added, carrying both agents documented in Good syndrome specifically: rituximab from PMID:27408396 and ciclosporin from PMID:38836142, the second of which also generalizes beyond its own case.
+
+    Its target_mechanisms link stops at pure red cell aplasia and deliberately does not reach the B-cell or immunoglobulin nodes, because both sources report the hypogammaglobulinemia did not respond. That is the same dissociation the thymectomy evidence carries, from a second class of intervention, and the treatment description says so. It is curated as its own treatment rather than folded into supportive care because giving an immunosuppressant to a patient with an established combined immunodeficiency is a deliberate trade rather than a routine addition.
+
+    The review's diagnosis of the pattern is correct and this is its third instance: PMID:27408396 was cached by this PR, cited for the phenotype from one sentence of its abstract, and the very next sentence — which is also the paper's title — went unread. The entry's notes already name this failure mode.
+
+    One error caught by validation: CHEBI:4031's canonical label is 'cyclosporin A', not 'ciclosporin'. preferred_term keeps the INN spelling; the bound label matches the ontology.
+
+    Validation: just validate passes with 37/37 snippets verified, up from 35/35; validate-terms, check-entity-refs, check-causal-targets, check-duplicate-keys, check-enum-values, check-snippet-grading, check-title-snippets and check-snippet-length all clean; cache diff empty. Pathograph: 17 nodes, 16 edges, 0 orphan targets, one component apart from the deliberately isolated GTF2I node.
+
+    Not taken, as the review agreed: antimicrobial prophylaxis (PMID:29878906 is not cached) and canine Good syndrome (no PMID exists for it).
--- a/kb/disorders/Good_Syndrome.yaml
+++ b/kb/disorders/Good_Syndrome.yaml
@@ -506,8 +506,8 @@ treatments:
   treatment_term:
     preferred_term: thymectomy
     term:
-      id: NCIT:C15329
-      label: Surgical Procedure
+      id: NCIT:C29894
+      label: Thymectomy
   target_mechanisms:
   - target: Thymic Epithelial Neoplasia
     description: >-
@@ -532,6 +532,57 @@ treatments:
     explanation: >-
       States the persistence of the immunodeficiency after the operation, and
       the consequence for management.
+- name: Immunosuppressive therapy for the autoimmune complications
+  description: >-
+    Directed at the autoimmune arm of the syndrome, not at the immunodeficiency.
+    Both agents documented in Good syndrome specifically — rituximab and
+    ciclosporin — corrected the pure red cell aplasia while leaving the
+    hypogammaglobulinemia untouched, which is the same dissociation the
+    thymectomy evidence shows. Curated as its own treatment rather than folded
+    into supportive care because giving an immunosuppressant to a patient who
+    already has a combined immunodeficiency is a deliberate and uncomfortable
+    trade, not a routine addition.
+  therapeutic_modality: MONOCLONAL_ANTIBODY
+  treatment_term:
+    preferred_term: Pharmacotherapy
+    term:
+      id: NCIT:C15986
+      label: Pharmacotherapy
+    therapeutic_agent:
+    - preferred_term: rituximab
+      term:
+        id: NCIT:C1702
+        label: Rituximab
+    - preferred_term: ciclosporin
+      term:
+        id: CHEBI:4031
+        label: cyclosporin A
+  target_mechanisms:
+  - target: Pure red cell aplasia
+    description: >-
+      Both documented agents act on the red cell aplasia, restoring erythroid
+      maturation. No link is drawn to the B-cell or immunoglobulin nodes,
+      because both sources report that the hypogammaglobulinemia did not
+      respond.
+  evidence:
+  - reference: PMID:27408396
+    reference_title: "Thymoma with Concomitant Pure Red Cell Aplasia, Good's Syndrome and Myasthenia Gravis Responding to Rituximab."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "He responded to rituximab with restoration of bone marrow erythroid maturation and stabilization of red blood cell counts."
+    explanation: >-
+      Documents the rituximab response in a patient carrying Good syndrome,
+      pure red cell aplasia and myasthenia gravis together.
+  - reference: PMID:38836142
+    reference_title: "Good's Syndrome With Pure Red Cell Aplasia and Subclinical Myasthenia Gravis: A Case Report and Review of Literature."
+    supports: SUPPORT
+    evidence_source: HUMAN_CLINICAL
+    snippet: "A review of the literature on simultaneous Good's syndrome with PRCA also suggested the efficacy of CsA on PRCA but not hypoglobulinemia, suggesting the distinct underlying mechanisms between these two paraneoplastic symptoms with thymoma."
+    explanation: >-
+      Generalizes the ciclosporin result beyond the single case, and states the
+      inference this entry draws from it: the two paraneoplastic manifestations
+      have distinct mechanisms, which is why the treatment link stops at the red
+      cell aplasia.
 diagnosis:
 - name: Immunoglobulin monitoring in thymoma patients
   description: >-


gingin77:
**You were right about item 1, and the important part is that my `notes` made a false claim about repository state.** Pushed as 5dabd2a.

`references_cache/PMID_39180607.md` is in the repo, `Thymoma.yaml` cites it, and both sentences are there. The defect was in my search, not the cache: I checked only the two references this entry had chosen to cite, then wrote that negative result up as though it were a property of the literature. That's worse than the missing section, so `notes` now carries an explicit correction and so does the history record. Thanks for catching it — I'd have shipped a confidently wrong statement.

**1. Treatments + diagnosis.** IVIG added, bound to `NCIT:C121331` (Intravenous Immunoglobulin Therapy) rather than generic Pharmacotherapy with an agent — `NCIT:C572` Immunoglobulin isn't in the `ChemicalEntityTerm` enum, and the therapy term is the more accurate binding anyway. It carries a `target_mechanisms` link so it joins the pathograph. The second sentence became a `diagnosis:` entry for serial immunoglobulin monitoring in known thymoma.

**2. Pathograph connectivity.** Fixed, and your diagnosis of *why* was exact: the chain restated each phenotype as its own terminal node instead of pointing at it. Duplicate terminal nodes removed, replaced with bare-name `downstream` edges. Verified by building the graph rather than eyeballing it — 11 edges, no unresolved targets, one connected component, single root at `Thymic Epithelial Neoplasia`.

**3. Both bundled nodes split.** The two-edge fan-out to infection *and* autoimmunity was the right tell. `Impaired T Cell Output and Loss of Central Tolerance` → `Reduced Thymopoietic Output` + `Failed Central Tolerance in the Thymic Medulla`, which also lets the autoimmune branch attach where it belongs, at the medullary compartment. `Peripheral B Cell Depletion and Hypogammaglobulinemia` drops the organism-level half its `CELLULAR` tag contradicted.

**5. `genetic:` GTF2I added — with a correction.** Your suggested `hgnc:4661` is **GTF2IRD1**; GTF2I is `hgnc:4659`, which is what's bound. `variant_origin: SOMATIC` and `relationship_type: UNKNOWN`, the latter deliberately: GTF2I is not a germline cause here, and whether GTF2I-mutant thymomas are the ones that give rise to Good syndrome has not been tested. The notes say so rather than letting the section imply causation.

**6. `animal_models:` added** for the Gtf2i mutant mouse, `PARTIALLY_RECAPITULATES` / `MODERATE` / `TISSUE`. The `limitations` record what I think is the salient thing: it was built to study thymoma biology, is not reported to develop B-cell depletion or hypogammaglobulinemia, and so supports the upstream half of one hypothesis while being silent on the half that is actually contested.

**4. Phenotypes — partially.** B lymphocytopenia added (`HP:0010976`), since you're right it's the feature separating this from CVID and the entry turned on it throughout.

The other five I've left out, and recorded why in `notes` rather than silently. Your own caveat is the reason: those percentages are in PMC full text, not the cached `PMID:34113351` abstract. Adding pure red cell aplasia and myasthenia gravis with frequencies I cannot quote would repeat the exact mistake item 1 was about — asserting something about the literature I hadn't verified. They stay named in prose, and the notes say a full-text-cached source is the fix.

**Not taken:** prevalence and the canine comparative model, both needing a fetch for the same reason.

**Validation:** `just validate`, `just validate-terms`, `just count-verified-snippets` (**19/19**, up from 14/14), `just check-duplicate-keys`, `just check-entity-refs`, `just check-causal-targets`, `just check-enum-values`, `just validate-history` all pass. Cache diff empty.

---

gingin77:
## `claude-review` is failing on this PR for a reason unrelated to the diff

The check errored 7 seconds into [run 33878271486](https://github.com/monarch-initiative/dismech/actions/runs/33878271486/job/101040391944). The job log gives the cause outright:

```
num_turns=3 cost=$0.3359632 subtype=success
```

The reviewer agent's account has hit its usage limit. This is account-level quota exhaustion, not a finding about this PR — no review was posted at all, and the run ended before reading the diff.

**Not re-running it, and that is deliberate.** Normally I would spend one re-run to confirm a failure isn't this PR's, but the log already names both the cause and the reset time. A re-run before **Sep 5, 8pm UTC** cannot succeed, so it would burn a cycle to reproduce a message I already have.

**No fix exists that I could port.** Nothing in this repository can raise that quota, and there is no change to this branch that would make the check pass. The same limit will block the pending re-review on [#10922](https://github.com/monarch-initiative/dismech/pull/10922), which is in the same state.

### Where this PR actually stands

The content changes from the review round are pushed and verified locally at head `5dabd2a`:

| Check | Result |
|---|---|
| `just validate` | passes |
| `just count-verified-snippets` | **19/19** (was 14/14) |
| `just check-duplicate-keys` / `check-entity-refs` / `check-causal-targets` / `check-enum-values` | all OK |
| Pathograph connectivity | 11 edges, one component, single root — verified by building the graph |

So the PR is not waiting on work. It is waiting on the reviewer's quota to reset, and it needs either a re-run after the reset or a human review in the meantime.

I will keep this PR watched and re-check after the reset rather than leaving it idle.

---

gingin77:
/review

---

gingin77:
## Remaining suggestions taken — prognostic clusters and prevalence

Pushed as `2838b8e0`. This follows `5dabd2a`, which closed all six blocking items; that push's re-review died on reviewer account quota rather than on the diff. The quota has since reset, so this push re-triggers the check.

Both items here were quotable from `PMID:34113351` — already cached, already cited five times by this entry. Same shape as the IVIG finding you caught: under-consumption of a source in hand.

**`has_subtypes:`** — the three clinical subgroups the systematic review derived by multiple correspondence analysis and unsupervised hierarchical clustering over 162 published cases: cellular immunity defect infections, other immunity defect infections, humoral and phagocytic immunity defect infections. Cluster 1 also carries the independent prognostic-factor odds ratio for its defining infection type, which is what makes the stratum clinically meaningful rather than merely descriptive.

A caveat I put in `review_notes` rather than leaving implicit: **these are statistical clusters, not nosological entities.** No ontology term, no separate MONDO concept, no genetic basis — so no `subtype_term` is bound and none should be manufactured later. They are curated as subtypes because the source treats them as clinically distinct strata with distinct prognosis, and because the entry's prognostic content was otherwise unqueryable: it sat in one `progression` phase's prose.

**`prevalence:`** — one `CASES_IN_LITERATURE` record with no `rate_per_100000`, because no population rate has been reported for this disease. `prevalence_class: RARE` is the source's own qualitative word, not a band inferred from the case count, and `notes` says so — 162 cases assembled from a decade of literature is not a denominator-based estimate.

**Not taken: canine Good syndrome as a comparative model.** No reference cached in this repository carries a quotable veterinary statement, so this needs a fetch and a read rather than a line change. Left as a genuine follow-up rather than half-done.

### Validation

25/25 snippets verified (was 19/19). `just validate`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-enum-values`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `validate-history` all clean. Cache diff empty.

The `CHANGES_REQUESTED` review on `bb5ba8dc` predates both fix pushes. It is sticky and does not clear on its own — a fresh review supersedes it, which is what this push should produce. I have not dismissed it and will not.

---

gingin77:
All four blocking items closed, plus both actionable suggestions. Pushed as `01497ccc`, bundled into one commit as you asked.

## 1. The thymectomy-failure claim — and a correction to the source you named

You were right that this was asserted five times and cited zero times. It is the load-bearing claim of the entry and it had no evidence item.

**But PMID:38122866 does not say it.** I fetched and read it. Its abstract describes a 37-year prospective cohort and a progressive decline in immune compartments; the *"even after thymectomy"* clause belongs to the deep-research report's paraphrase, not to the paper. Citing it for that claim would have been the exact defect the review is trying to fix, one layer down.

So I searched the whole reference cache instead — the same move that found the IVIG reference last round — and three references already in this repository state it:

| Reference | What it gives |
|---|---|
| `PMID:34888060` | *"Despite thymectomy, the immunodeficiencies persist and lifelong immunoglobulin replacement is necessary to prevent infections."* GS-specific, direct. |
| `PMID:34966753` | A **dissociation**: PRCA remits after thymectomy plus IST in 61% of patients, *"however, Good's syndrome was unaffected."* |
| `PMID:38836142` | A second dissociation, inside one patient: ciclosporin corrected the PRCA and *"hypoglobulinemia was not recovered."* |

The dissociations are the better evidence, and worth pointing out as a kind: a bare negative is compatible with thymectomy simply being a weak intervention. A dissociation — where the same operation, in the same cohort, resolves the companion condition and not this one — rules that out and localises the persistence to the syndrome.

**PMID:38122866 earns its place anyway, for something the entry needed more.** Its *"multidecade progressive decline in lymphocytes, platelets, hemoglobin, and red blood cells… suggesting gradual bone marrow failure"* is evidence for `gs_marrow_b_cell_arrest` — the hypothesis that until now carried **no evidence item at all**. A thymus-confined account does not predict falling platelets and red cells. Graded `INDIRECT`, and its `notes` now say plainly that this is *not* the transitional-B-cell finding the hypothesis was originally built on, which is still unquotable from anything cached here.

## 2. Thymectomy as a treatment

Added, `SURGERY`, with the scope in its description: performed for oncologic control, does not correct the immunodeficiency. `target_mechanisms` links to `Thymic Epithelial Neoplasia` and **stops there** — no edge reaches the immunological nodes downstream, because the cited evidence is that removing the tumour does not reverse them. Your framing was right that this carries the argument as structured content instead of prose.

## 3. PRCA and myasthenia gravis

Your diagnosis of my reasoning was correct and the objection answered a question you hadn't asked. Both added with **no frequency band** and a note saying why, exactly as B lymphocytopenia already was. Wired off the central-tolerance node.

`PMID:27408396` was useful, and `PMID:38836142` — found in the same cache sweep — is better for this: one patient with GS, PRCA and serologically confirmed subclinical myasthenia gravis together.

## 4. The truncated snippet

Fixed. You were right that the truncation was working around the cache's line wrap and right that it was unnecessary — the validator normalises the wrap, and the full sentence verifies.

## Suggestions

**`evidence_source` on the genetic GTF2I item → `OTHER`.** Agreed, taken.

**GTF2I as a `genes:` descriptor on the neoplasia node — taken, but it does not do what you expected, and I think the current state is right.** GTF2I still renders isolated. `src/dismech/graph.py` suppresses the inferred gene→mechanism edge when `relationship_type` is `UNKNOWN` (`_NON_CONTRIBUTING_RELATIONSHIP_TYPES`), and that suppression is keyed on the `genetic:` item, not on the descriptor match — so no descriptor placed anywhere can connect it.

The only way to draw the edge is to give the item a contributing `relationship_type`, which is precisely the causal claim this entry declines to make. **The isolated node is the graph correctly showing that we do not assert GTF2I causes Good syndrome.** The descriptor is still worth having — it makes the node queryable by gene — and the `genetic:` notes now record all of this so the next reader does not try to "fix" the isolation.

**Canine Good syndrome — declined again, with the reason.** The report cites a bare PMC identifier and a JAVMA abstract URL; `just fetch-reference` has no source for either. Nothing quotable is reachable, so there is no honest way to add the model.

## Validation

**35/35 snippets verified** (was 25/25). `just validate`, `validate-terms`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-title-snippets`, `check-snippet-length`, `check-snippet-grading`, `validate-history` all clean. Cache diff empty. Pathograph rebuilt: 16 nodes, 15 edges, 0 orphan targets, one component apart from GTF2I.

Five new cache files committed: `PMID_34888060`, `PMID_34966753`, `PMID_38836142`, `PMID_38122866`, `PMID_27408396`.

One thing worth recording beyond this PR, and the entry's `notes` now say it: **all three of this entry's omissions had the same cause** — I searched the references the entry already cited rather than the repository-wide cache. That produced the IVIG error, the uncited thymectomy claim, and the two missing phenotypes. Reference count consumed is now 8 of 28.

---

gingin77:
Both items taken. Pushed as `e220b851`.

## 1. `NCIT:C29894` instead of the generic surgical term

You're right, and the detail that makes it clearly right is that the term is *already in this repo's caches* — `cache/ncit/terms.csv` and the `treatmentactionterm` enum cache both carry it, so the rebind validated offline with no new rows and no network call. Five sibling entries were already using it.

The report offered both halves and I took the generic one. That's the same shape as the `NCIT:C15329` binding elsewhere in this entry being correct — the difference is that a *specific* term existed here and I didn't check.

## 2. Immunosuppressive therapy for the autoimmune arm

Added, with both agents documented in Good syndrome specifically: rituximab from `PMID:27408396` and ciclosporin from `PMID:38836142`.

I used a different sentence from `PMID:38836142` than you might expect. Rather than the case's own ciclosporin result, I quoted the sentence that generalises it:

> A review of the literature on simultaneous Good's syndrome with PRCA also suggested the efficacy of CsA on PRCA but not hypoglobulinemia, suggesting the distinct underlying mechanisms between these two paraneoplastic symptoms with thymoma.

That sentence carries both the efficacy claim and the authors' own inference from it, which is the inference this entry draws: **the treatment link stops at pure red cell aplasia and deliberately does not reach the B-cell or immunoglobulin nodes.** Both sources report the hypogammaglobulinemia did not respond. So this is the same dissociation the thymectomy evidence carries, now arriving from a second, pharmacological class of intervention — which is a stronger position than the entry had an hour ago.

Curated as its own treatment rather than folded into supportive care, because giving an immunosuppressant to a patient with an established combined immunodeficiency is a deliberate and uncomfortable trade, not a routine addition.

## On the pattern

Your read is right and I don't want to soften it: **this is the third instance.** `PMID:27408396` was cached *by this PR*, cited for the PRCA phenotype from one sentence of its abstract, and the very next sentence — which is also the paper's title — went unread. IVIG, then thymectomy, now rituximab. The entry's `notes` name the failure mode and I still reproduced it inside the same PR.

## On the reviewer errors

Noted, and worth saying since you raised it: going to the primary source is cheap and I'd rather do it than not, but two of two isn't a pattern I'd generalise from. The `hgnc:4661` and `PMID:38122866` catches both came from reading a cached file that was already sitting there — the same move that has now caught three of my own omissions. It cuts both ways more often than it cuts yours.

## One error validation caught

`CHEBI:4031`'s canonical label is `cyclosporin A`, not `ciclosporin`. `preferred_term` keeps the INN spelling; the bound label matches the ontology.

## Validation

**37/37 snippets verified** (was 35/35). `just validate`, `validate-terms`, `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-snippet-grading`, `check-title-snippets`, `check-snippet-length`, `validate-history` all clean. Cache diff empty.

Pathograph rebuilt: **17 nodes, 16 edges, 0 orphan targets**, one component apart from the deliberately isolated `GTF2I`.

**Not taken**, as you agreed: antimicrobial prophylaxis (`PMID:29878906` isn't cached) and canine Good syndrome (no PMID exists for it).

---

PR #10929, opened 2026-09-04
Curate immunodeficiency 15A (IKBKB gain-of-function, MONDO:0032599)

caufieldjh:
Adds a new disorder entry for immunodeficiency 15A, the autosomal dominant gain-of-function IKBKB disorder. Every claim is backed by a quote checked against a cached reference: 61 of 61 snippets verify. Built from a `claude_code` deep-research run (committed under `research/`) plus the primary literature.

Independent of any other open PR; branched from `main`.

## Why the direction is the interesting part

IKK2 is the kinase that phosphorylates IκBα and so releases NF-κB into the nucleus. NF-κB drives lymphocyte activation, so a variant that makes IKK2 *more* active ought to produce more immunity. It produces immune deficiency instead.

The entry models why. Constitutive kinase activity uncouples NF-κB from receptor engagement — patient cells show elevated IκBα phosphorylation before any stimulus at all — and lymphocyte development and selection depend on reading the difference between signal and baseline. Naive T cells then hyperproliferate to a stimulus that normally moves few cells, and the cultures end up with *fewer* cells, consistent with substantial death. B cells activate abnormally and fail to make immunoglobulin.

```
IKBKB kinase-domain gain-of-function variant
  └─> constitutive IκBα phosphorylation
        └─> enhanced canonical NF-κB signalling
              ├─> lowered lymphocyte activation threshold
              │     ├─> naive T cell depletion ─> lymphopenia, combined immunodeficiency
              │     └─> impaired B cell differentiation ─> memory B cell loss,
              │           hypogammaglobulinaemia, recurrent respiratory infection
              └─> increased regulatory T cell output
```

Also included: 13 phenotypes, an IVIG treatment, two orthologous knock-in mouse models, inheritance, a progression record, an IUIS classification, a diagnosis record, the IMD15B differential, and one open question.

## How activation produces deficiency is recorded as unresolved, not answered

The defining study states the paradox and does not settle it, so neither does the entry. `discussions[0]` is an `OPEN_QUESTION` carrying both candidate answers — a regulatory one (FOXP3 is under canonical NF-κB control; regulatory T cells and per-cell FOXP3 are increased, and in the mouse the increase is already present in the thymus) and an attritional one (chronic hyperactivation consuming the naive pool). It also carries the observation that cuts against the leading answer: one of the two founding probands had the deficiency with *normal* regulatory T cells.

## The main hazard on this disease, and how it was handled

**IMD15A and IMD15B are different diseases caused by the same gene in opposite directions.** 15A is dominant gain-of-function; 15B is recessive loss-of-function. Several of the best available sources are about 15B, and importing one as a 15A claim would be invisible to every automated check.

Every 15B source is confined to `differential_diagnoses`. Specifically, the research report offered the IKBKB transplant cohort (PMID:30391351) toward treatment — but that cohort is 16 infants homozygous for `c.1292dupG`, which is 15B. It is not used as treatment evidence here. The consequence is that treatment coverage is one documented immunoglobulin course in a single patient, and `notes` says so rather than padding the section.

The mirror-image framing is worth having: 15A lowers the activation threshold and depletes naive cells; 15B leaves lymphocytes numerically normal but frozen in a naive, unresponsive state with absent regulatory and γδ T cells.

## What a pre-PR review caught

An adversarial review pass ran against `dismech-pr-review` before this PR opened. It returned one blocking finding, fifteen important ones and several minor ones; all were addressed in the second commit. Two are worth surfacing.

**A factual error contradicted by a source the entry already cited.** I had called `p.Thr559Met` a kinase-domain variant, in four prose slots plus a pathophysiology node name. PMID:36378426 states that "Both the E518K and T559M are located between the leucine-zipper and NEMO-binding regions". Those variants are now excluded from the kinase-domain allele series with the reason recorded, and the top-level description no longer defines the disease as kinase-domain-only.

**A genotype caveat that was understated in the flattering direction.** That same patient carried the two variants **in trans**, one inherited from each asymptomatic parent — a compound heterozygous genotype, not the single dominant allele that defines IMD15A. The entry had said only "two heterozygous variants", which reads as one allele. Now stated explicitly everywhere the case is cited, together with the fact that only T559M behaved as gain-of-function in vitro.

Also fixed: a differential explanation that extended a 15B-only TREC finding to 15A; three claim–evidence mismatches where a better quote was already in a cached file; two nodes making human claims on model-organism evidence alone; four ex vivo patient-cell assays graded inconsistently between `IN_VITRO` and `HUMAN_CLINICAL` within the same nodes; and two cell-type bindings wrong in opposite directions.

One judgement call I'd defend: `supports: REFUTE` on the open-question discussion became `SUPPORT`. `supports` is scoped to the claim the item attaches to — here, that the reconciliation is unresolved — which the counterexample strengthens rather than contradicts. The nuance lives in the `explanation`.

## Handling of the deep-research report

| Check | Result |
|---|---|
| References resolved | 8/8, 0 unresolved |
| Flagged possibly off topic | 1 |
| Ontology terms resolved | 38/41 |
| Terms named as a **different** term | 4 |
| Obsolete terms offered | 1 |

The single off-topic flag was a false positive worth chasing. `PMID:10229185` is the 1999 IKK-β knockout mouse — no shared vocabulary with a 2018-onward disease report, but directly relevant. Reading it rather than dismissing it established that mouse *Ikbkb* nulls die at E14.5 from liver degeneration while humans homozygous for a null survive with immunodeficiency. That species asymmetry is recorded in `notes`, but *not* modelled as a human/model mismatch, because it concerns the loss-of-function disorder rather than this one.

Mislabelled CURIEs were rejected. The report offered `HP:0001878` labelled "AIHA"; HP calls that *Hemolytic anemia*, and the reported patient had autoimmune haemolytic anaemia with anti-D antibodies, so the entry uses the specific `HP:0001890`. `HP:0010976` and `HP:0002164` were offered with table-cell text as their labels and not used. The obsolete `GO:0002741` was not used.

Separately, `PMID:32554083` — a directly relevant fourth-allele paper (K171R, activation loop) — was surfaced by the report and initially overlooked on my side. Reading it added the allele, a second knock-in mouse, and an ectodermal dysplasia phenotype now supported by two independently ascertained patients rather than one.

## What was deliberately left out

Each is recorded in `notes:` with its reason.

- **No prevalence band or rate.** The published experience is 16 patients from 4 families; that is recorded as `CASES_IN_LITERATURE` with `prevalence_class: UNKNOWN` rather than converted into a rate.
- **No frequency band on lymphopenia.** One series reports it as fixed within a kindred, another as an age-dependent decline across four families. A single band would misdescribe one of them.
- **No datasets.** None specific to this disorder were identified.
- **Single-patient features** — premature cataracts, hidradenitis suppurativa, hepatosplenomegaly, the CMV colitis reported in an adult. Quotable sentences exist for most; the reason is scope, not sourcing.
- **HSCT**, for the reason given above: the transplant evidence is 15B.

## Validation

All pass:

```
just validate-disorders kb/disorders/Immunodeficiency_15A.yaml   # the batched CI gate
just validate            # schema + terms + 61/61 snippets
just check-entity-refs   just check-causal-targets   just check-duplicate-keys
just check-enum-values   just check-qualifier-terms-online
just check-folded-hyphens   just check-snippet-length   just check-title-snippets
just check-snippet-grading  just check-reference-titles
just check-environmental-evidence
```

Re-run after rebasing onto current `main`. No baseline file was widened. Weighted compliance 86.6% — lower than the 90.4% before the review round, because the added causal edges and new sections enlarged the denominator.

No GeneReviews chapter exists for IKBKB, verified against PubMed, so that phenotype baseline is unavailable and no substitute is claimed.

## Files

- `kb/disorders/Immunodeficiency_15A.yaml` — the entry
- `history/disorders/Immunodeficiency_15A/` — two curation history records (initial, and the review round)
- `research/Immunodeficiency_15A-deep-research-claude_code.md` (+ citations sidecar)
- `references_cache/` — 7 new cached references
- `cache/` — rows added by term validation
- `stubs/Immunodeficiency_15a.yaml` — deleted

https://claude.ai/code/session_01PNRkJeTACVpLfM4XWjGor6

---
_Generated by [Claude Code](https://claude.ai/code/session_01PNRkJeTACVpLfM4XWjGor6)_

### DIFF
--- a/history/disorders/Immunodeficiency_15A/2026-09-08T054148Z-claude-code-45847f.yaml
+++ b/history/disorders/Immunodeficiency_15A/2026-09-08T054148Z-claude-code-45847f.yaml
@@ -1,0 +1,35 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Immunodeficiency_15A
+  path: kb/disorders/Immunodeficiency_15A.yaml
+session:
+  id: 2026-09-08T054148Z-claude-code-45847f
+  timestamp: '2026-09-08T05:41:48Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10929
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - prevalence
+  - animal_models
+  - notes
+  summary: 'Review round 2: prevalence tier, model_scale, notes de-duplication'
+  details: |
+    Took all three optional suggestions from the 2026-09-08 re-review (approving, no blocking findings).
+
+    1. prevalence_class UNKNOWN -> ULTRA_RARE. The record is measure_type CASES_IN_LITERATURE with 16 patients from 4 families and no rate_per_100000. PrevalenceClassEnum documents ULTRA_RARE as the tier for 'only a handful of reported cases, with no population rate available. Often paired with measure_type CASES_IN_LITERATURE', and the schema states the qualitative tiers presuppose that no numeric estimate was given. So declining a numeric band never entailed declining the qualitative tier. The entry also contradicted itself: prevalence[0].evidence[0].explanation already called the quote 'the basis for describing this as an ultra-rare disorder' while the slot said UNKNOWN, which is how a later curator ends up reversing a deliberate choice. notes now records why the tier and the declined rate are compatible.
+
+    2. model_scale added to all three modeled_mechanisms links. The review suggested ORGANISM on all three because both knock-in mice are whole animals. That is what the models ARE; the slot records what a link OBSERVES, so blanket ORGANISM would have been wrong on two of the three. Assigned per link from the readout actually cited: MOLECULAR for Constitutive IkappaB-alpha Phosphorylation (splenocyte phospho-IkappaB-alpha to total ratio), CELLULAR for Naive T Cell Depletion (splenic CD3+ proportion and absolute count), ORGANISM for Enhanced Canonical NF-kappaB Signalling (the K171R link cites a whole-animal immunologic phenotype comparison and carries no cellular or molecular readout). Result is two ALIGNED and one MODEL_ABOVE_TARGET; none is upward extrapolation, so no limitations or divergences requirement is triggered and model-scale-audit --strict passes.
+
+    3. Deleted a duplicated sentence in notes. 'each is a single-patient observation rather than an established feature of the disorder' was restated three lines later; the surviving text keeps the one fact the second copy added, that the founding proband's full presentation is preserved under the Eczematoid dermatitis phenotype.
+
+    No evidence item, snippet, reference or ontology binding changed. Validation on the edited tree: validate-disorders clean with 61/61 snippets verified, validate-terms clean, check-enum-values, check-duplicate-keys, check-entity-refs, check-causal-targets, check-qualifier-terms and model-scale-audit --strict all pass.
--- a/kb/disorders/Immunodeficiency_15A.yaml
+++ b/kb/disorders/Immunodeficiency_15A.yaml
@@ -94,11 +94,15 @@ inheritance:
 prevalence:
 - population: Worldwide
   measure_type: CASES_IN_LITERATURE
-  prevalence_class: UNKNOWN
+  prevalence_class: ULTRA_RARE
   notes: >
     No population estimate exists. The published experience is 16 patients
     from 4 unrelated families in the largest series, plus scattered single
-    cases; the count is recorded rather than a rate, and no band is asserted.
+    cases, so the count is recorded rather than a rate and no numeric band
+    is asserted. ULTRA_RARE is the qualitative tier, which the schema
+    defines by threshold rather than by magnitude and pairs explicitly with
+    CASES_IN_LITERATURE; it presupposes no numeric estimate, so recording it
+    is not in tension with declining a rate.
   evidence:
   - reference: PMID:40403933
     reference_title: "IKBKB gain of function: An inborn error with clinical heterogeneity progressing toward combined immunodeficiency."
@@ -825,6 +829,7 @@ animal_models:
   - target: Constitutive IkappaB-alpha Phosphorylation
     relationship: RECAPITULATES
     fidelity: HIGH
+    model_scale: MOLECULAR
     description: >
       Splenocytes show constitutive IkappaB-alpha phosphorylation at baseline
       and prolonged phosphorylation after stimulation, with homozygotes
@@ -859,6 +864,7 @@ animal_models:
   - target: Naive T Cell Depletion
     relationship: RECAPITULATES
     fidelity: MODERATE
+    model_scale: CELLULAR
     description: >
       Mutant mice show a relative and absolute deficiency of splenic T cells
       in both CD4 and CD8 compartments, again dose-graded.
@@ -912,6 +918,7 @@ animal_models:
   - target: Enhanced Canonical NF-kappaB Signalling
     relationship: RECAPITULATES
     fidelity: MODERATE
+    model_scale: ORGANISM
     description: >
       The authors report that the mouse immunologic phenotype parallels the
       human one.
@@ -1132,11 +1139,10 @@ notes: >
   Quotable sentences exist for most of these - several are already used as
   snippets elsewhere in this entry - so the reason is not sourcing but
   scope: each is a single-patient observation rather than an established
-  feature of the disorder. These appear in the founding case
-  descriptions or the research report but are single-patient observations
-  rather than established features of the disorder; the founding proband's
-  full presentation is quoted under the `Eczematoid dermatitis` phenotype,
-  so the record of them is not lost.
+  feature of the disorder. They appear in the founding case descriptions or
+  the research report, and the founding proband's full presentation is
+  quoted under the `Eczematoid dermatitis` phenotype, so the record of them
+  is not lost.
 
   The mouse Ikbkb null allele is embryonic lethal at about E14.5 from liver
   degeneration (PMID:10229185), while humans homozygous for a null IKBKB


caufieldjh:
`claude-review` failed on this PR, and the cause is not this PR's diff — the review agent's Claude account has hit a usage limit. From the job log:

```
```

`num_turns=1` and `cost=$0` — it stopped before reading anything. No review was posted, so there is nothing here to address.

**I am not re-running it.** A re-run is the right move for a job that died on checkout, install, or runner loss, because those clear on retry. A quota does not: the limit resets at 2026-09-05 20:00 UTC, so any re-run before then fails identically and spends the one re-run this PR gets. I would rather keep it for a failure that a retry could actually fix.

Two things worth flagging for maintainers, since neither is visible from this PR alone:

- This is account-level, so it affects **every** PR in the repo until the reset, not just this one. Any PR opened or pushed in the next ~30 hours will show the same red `claude-review`.
- The job's own error message lists likely causes as "an invalid/retired `--model`, a bad `CLAUDE_CODE_OAUTH_TOKEN`, or an aborted run". None of those is what happened here, and the actual reason is only visible in the `detail` field. A quota case looks identical to a broken token at the check-run level, which is a bad failure mode to debug under time pressure — the message may be worth extending.

The rest of CI is unaffected: `test (3.13)` is running normally on this head, and `check-mention` passed. All thirteen local gates passed before push, including the batched `just validate-disorders` that CI runs, with 61/61 snippets verified.

I will re-request the review once the limit resets, by pushing the next real change rather than an empty commit. Until then this PR has no outstanding review feedback because no review ran.

---

caufieldjh:
Thanks — took four of the six suggestions, declined two with the reasoning now written into the entry. All in one push, `eccbf3ef1`.

I know this resets the approval. Three of these are evidence-discipline defects rather than polish, and they would otherwise sit in the KB permanently, so I judged one re-review cycle worth it. Batched into a single push for that reason.

**Taken:**

| # | Change |
|---|---|
| 1 | `FOXP3 expression is known to be regulated by canonical NF-κB signaling` regraded `HUMAN_CLINICAL` → `OTHER`. You're right that it is the authors summarising prior literature, not an observation in their patients. |
| 2 | The thymic-Treg hypothesis snippet regraded → `OTHER`. |
| 3 | The unevidenced developmental claim removed from the `Ectodermal dysplasia` description. |
| 4 | `Enhanced Canonical NF-kappaB Signalling → Eczematoid dermatitis` added as `INDIRECT_UNKNOWN_INTERMEDIATES`. |

On item 2 I went with `OTHER` rather than splitting the item, and I want to be explicit about why since CLAUDE.md's default is to split. The sentence is a single interpretive claim that happens to cite a mouse observation and a human one in support of it. Splitting it would mean quoting half a hypothesis, and neither half asserts anything on its own. The reasoning is in the `explanation` so a later curator doesn't silently undo it.

On item 4 I connected only the skin arm, not all three orphans. Drawing an edge to `Ectodermal dysplasia` would assert exactly the developmental mechanism item 3 just removed as unevidenced, so connecting it while deleting its justification would have been incoherent. Left detached, with the reason now stated in the phenotype description.

**Declined, and both are now in `notes:` rather than implicit:**

- **The biomarker panel.** You reached the right conclusion and I agree: `PMID:36378426` does carry quotable chemokine sentences, but that panel was measured in the patient with the in-trans E518K/T559M genotype the entry quarantines. Importing it reintroduces precisely the contamination the rest of the entry avoids. Recorded so the next curator doesn't rediscover the same quotable text and assume it was missed.
- **`HP:0005415`.** Agreed it is subsumed by the curated lymphopenia and naive-T terms.

One correction to something I said earlier on this PR: I predicted the reviewer's usage limit would keep `claude-review` red repo-wide until 2026-09-05 20:00 UTC. That was wrong — this run went through normally well before then. The earlier failure was real, but my inference about its duration was not, and anyone who read that comment should disregard the timing claim.

Two notes on your tooling report, since you flagged what you couldn't verify:

- The GeneReviews absence is confirmable. I ran `esearch` on `IKBKB AND GeneReviews[All Fields]` and it returns zero; the four `GeneReviews[Book]` hits for adjacent searches are ADA deficiency, CD40L deficiency, cystinosis and Hermansky-Pudlak. Your fallback reasoning reached the right answer.
- Your `runoak` "Unhandled node type: string" is the local-build path. Worth knowing that the local HPO build and OLS disagree on at least one label in this repo right now — `HP:0030374` is *Decreased memory B cell proportion* in OLS and *Decreased proportion of memory B cells* in the local build. `just validate-terms` uses OLS and is authoritative, which you relied on correctly.

61/61 snippets still verify; all gates green; no baseline widened.

---

caufieldjh:
Thanks — and taking the optional note seriously rather than waving it through, because there is a reason it can't be closed rather than just a cost.

**The edge can't carry evidence, because there is nothing to quote.** An evidence item on `Enhanced Canonical NF-kappaB Signalling → Eczematoid dermatitis` would have to assert the step itself: that pathway hyperactivation produces this skin disease in these patients. That is the exact claim I went looking for while addressing suggestion 3 and could not find in anything cached. It is why the `Ectodermal dysplasia` description now says the developmental step "is not evidenced in this entry" out loud. Writing an evidence item for the eczema edge would mean either quoting a sentence that reports the co-occurrence and presenting it as the mechanism, or manufacturing one. Neither is available to me.

So the honest state is what the entry now has: both endpoints separately evidenced, `INDIRECT_UNKNOWN_INTERMEDIATES` marking the step as inferred, and the description saying what it is attributing. If a source turns up that reports the mechanism, the edge is the place to hang it.

Two smaller things worth confirming, since you checked them independently rather than trusting the commit message:

- The `OTHER` resolution on the mixed mouse/human sentence — your framing is better than mine. I reached it by elimination and recorded the reasoning in the `explanation`; you established it is the *only* available resolution, because the obvious fix (split the item) collides with `check-snippet-grading` by construction. That is worth someone lifting into the curation guidance at some point: the "split items that mix sources" rule has an exception whenever the two sources are inside one sentence.
- The third detached node being the gene node is right, and structural.

Not pushing for the optional note. The approval is on a head I believe is correct, and a cycle spent adding an evidence item I would have to invent is worse than the gap it closes.

---

caufieldjh:
Dequeued from the merge queue at 18:56Z with `MERGE_CONFLICT`, now fixed in `ada6c0f39`.

**The conflict was two derived term caches, and the collision was benign in both cases.** Another curation PR had already cached the identical row and landed ahead of this branch:

| File | Row | This branch | main |
|---|---|---|---|
| `cache/go/terms.csv` | `GO:0008384` IkappaB kinase activity | 2026-09-04T13:12:08 | 2026-09-03T14:24:14 |
| `cache/hgnc/terms.csv` | `hgnc:5960` IKBKB | 2026-09-04T13:04:59 | 2026-09-03T14:24:13 |

Same CURIE, same canonical label on both sides — only the derivation timestamp differs. So there is no content question to resolve, just a question of whose timestamp survives.

**Resolution: main's rows taken verbatim.** I checked first that these were this branch's *only* additions to those two files, so taking main's side is the entire resolution rather than a partial one — both files are now byte-identical to main and this PR adds no row to either. Nothing was hand-edited; per CLAUDE.md, cache rows are derived artifacts and I did not retype one.

This is the failure mode the cache CSVs are structurally prone to: every curation PR appends to the same few files, so two PRs touching neighbouring lines conflict on the cache and nothing else. Worth knowing it produces a *timestamp-only* conflict, which looks alarming in the diff but carries no disagreement about the term.

**No curated content changed.** `kb/disorders/Immunodeficiency_15A.yaml` is untouched by the merge, so no history record accompanies it — the entry is exactly the tree you approved.

Re-validated after merging, not assumed:

- `just validate-disorders` — passes, **61/61 snippets verified**
- `check-duplicate-keys`, `check-enum-values`, `check-entity-refs`, `check-causal-targets` — pass (the whole-KB gates, which are the ones a merge group can fail on)
- `check-term-cache-integrity`, `check-cache-order` — pass
- `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `check-environmental-evidence`, `check-reference-titles` — pass

No baseline widened. Final diff is 14 files, all belonging to this curation.

This push resets the approval under `dismiss_stale_reviews`. Unavoidable — a conflict resolution has to be a real commit — but the only delta since the approving review is the merge itself and the two cache rows described above.

---

caufieldjh:
`claude-review` is red on `ada6c0f39`, and it is not this PR's failure. The job log:

```
error='' detail="You've hit your limit · resets 9:30pm (UTC)" num_turns=1 cost=$0 subtype=success
```

The review agent's account hit its usage limit. `num_turns=1` and `cost=$0` mean it exited before reading a single file, so this says nothing about the diff — it is the same repo-wide condition that hit this PR once before, and it will be hitting every other PR reviewed in this window too.

**Not re-running it yet, deliberately.** A quota does not clear on retry, so re-running now would fail identically and spend the one re-run this failure gets. The log names a reset at **21:30 UTC**; I will re-trigger the review after that, which is when a retry can actually succeed. No code change is needed or possible for this — there is no fix to port, because nothing in the repository is broken.

`test (3.13)` is still running on this head; that one is a real check and I will act on it if it goes red.

For context on what this head contains: `ada6c0f39` is the merge that resolved the `MERGE_CONFLICT` dequeue, described in the comment above. The two conflicting rows were timestamp-only collisions in derived term caches, resolved by taking main's rows, and every deterministic gate was re-run and passes — including the whole-KB ones a merge group tests. So the only thing standing between this PR and the queue is a review that has not been able to run.

---

caufieldjh:
Correcting my previous comment: I said I would re-trigger the review after the 21:30 UTC reset. **I can't — I don't have the permission.** Both routes return 403:

- `POST /actions/workflows/claude-code-review.yml/dispatches` → `403 Resource not accessible by integration`
- `POST /actions/runs/33986107559/rerun-failed-jobs` → `403 Resource not accessible by integration`

So the retry I promised is not something I can deliver, and I'd rather say that than leave the commitment sitting here unmet.

**Where that leaves the PR.** The quota window has passed, so a review would very likely succeed now — it just needs something to start it. Two things can:

- a maintainer re-running the failed job or dispatching `claude-code-review.yml` with `pr_number: 10929`;
- any future push to the branch, which triggers the workflow on `pull_request`.

I am not manufacturing that push. There is no change this PR needs, and pushing something contrived to kick CI is exactly the empty-commit pattern the repo prohibits. If a real change becomes necessary for another reason, the review comes along with it for free.

**Nothing about the PR is blocking.** On `ada6c0f39`: `test (3.13)` passes, and locally `validate-disorders` (61/61 snippets), `check-duplicate-keys`, `check-enum-values`, `check-entity-refs`, `check-causal-targets`, `check-term-cache-integrity`, `check-cache-order`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens`, `check-environmental-evidence` and `check-reference-titles` all pass. No baseline widened. The merge conflict that dequeued it is resolved and the branch merges clean against current `main`.

The only red check is a review that has not been able to run, twice, for reasons outside this repository. Worth flagging beyond this PR: the same quota exhaustion will be silently failing `claude-review` on every other PR reviewed in these windows, and a failed review blocks the deterministic auto-merge sweep's "green" predicate — so PRs can accumulate behind an infrastructure limit while looking like they have review problems.

I'll keep watching this PR until it merges or closes.

---

caufieldjh:
Took all three suggestions, in one push (`11beaefd4`). One of them I did not implement the way you proposed, and that is the part worth reading.

**`model_scale` — I did not put `ORGANISM` on all three links.**

You suggested `ORGANISM` throughout on the grounds that both knock-in mice are whole organisms. That is what the models *are*; the slot records what each link *observes*. CLAUDE.md is explicit about the difference, and its worked example is exactly this trap — a Boolean network whose output node is named "bone erosion" still observes only molecular or cellular state.

So I assigned per link, from the readout each one actually cites:

| Link target | `model_scale` | Why |
|---|---|---|
| Constitutive IkappaB-alpha Phosphorylation | `MOLECULAR` | readout is a splenocyte phospho-IκBα to total IκBα ratio |
| Naive T Cell Depletion | `CELLULAR` | readout is splenic CD3+ T cell proportion and absolute count |
| Enhanced Canonical NF-kappaB Signalling | `ORGANISM` | the K171R link cites a whole-animal immunologic phenotype comparison and carries no cellular or molecular readout of its own |

That gives two `ALIGNED` and one `MODEL_ABOVE_TARGET`. None is upward extrapolation, so no `limitations` or `divergences` requirement is triggered, and `just model-scale-audit --strict` passes. Blanket `ORGANISM` would have recorded a scale gap on the first two links that the evidence does not show.

**`prevalence_class` → `ULTRA_RARE`.** Agreed, and your reading of the schema is right: the qualitative tiers presuppose no numeric estimate, so declining a rate never entailed declining the tier, and `PrevalenceClassEnum` names `CASES_IN_LITERATURE` as the usual pairing. The internal contradiction you spotted is the reason I went this way rather than trimming the explanation — the evidence `explanation` was already calling it ultra-rare, and the record carries no `rate_per_100000` to conflict with the tier. `notes` now states why the tier and the declined rate are compatible, so the next curator does not reverse it.

**Duplicated sentence in `notes`.** Gone. The surviving text keeps the one fact the second copy added, that the founding proband's full presentation is preserved under the `Eczematoid dermatitis` phenotype.

**On the observation you raised about `PMID:36378426`** — no change, and I think the asymmetry is right for the reason you gave. The biomarker panel asserts quantitative disease biology from a compound-heterozygous in-trans genotype this entry quarantines; a caveated account of how that patient was managed is a different kind of claim. It is documented in `treatments[0].notes` and the entry `notes`, and now in your review too, which is a better place for it than either.

No evidence item, snippet, reference title or ontology binding changed in this push.

Validated before pushing: `validate-disorders` clean with 61/61 snippets verified, `validate-terms` clean, plus `check-enum-values`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets`, `check-qualifier-terms` and `model-scale-audit --strict`. No baseline widened. History record added at `history/disorders/Immunodeficiency_15A/2026-09-08T054148Z-claude-code-45847f.yaml`.

I know this push drops the approval. Worth it for the prevalence contradiction, which is permanent KB content and is the kind of thing a later curator silently "fixes" the wrong way.

---

Issue #9845, opened 2026-08-28
A wrong gene identifier in a deep-research report passes reference validation with confabulation_rate 0.0


aleixpuigb:
The `openscientist` report for MRD30 (`research/Intellectual_Disability_Autosomal_Dominant_30-deep-research-openscientist.md`) contains this in its key-identifiers table:

```
| Gene (HGNC) | *ZMYND11*, HGNC:29316 |
```

**HGNC:29316 is ZSWIM6.** ZMYND11 is `hgnc:16966`:

```console
$ uv run runoak -i sqlite:obo:hgnc info hgnc:16966 hgnc:29316
hgnc:16966 ! ZMYND11
hgnc:29316 ! ZSWIM6
```

The same report reports:

```yaml
reference_validation:
  total_references: 18
  verified: 18
  not_found: 0
  confabulation_rate: 0.0
```

and `just preflight-dr … MONDO:0014486` returned **PASS**, with `ZMYND11=79` mentions and no rival gene.

### The gap

Both checks are working as designed, and neither is designed to catch this:

- **Reference validation** resolves *cited identifiers* — PMIDs, DOIs, NCTs. An HGNC ID asserted in a prose table is not a citation, so it is never looked up. `confabulation_rate: 0.0` therefore means "every PMID exists", which a curator can easily read as "the identifiers in this report are sound".
- **`preflight-dr`** counts *gene symbol mentions* to detect Named Entity Confusion. ZMYND11 is mentioned 79 times and is the right gene, so it passes. The check never looks at the CURIE next to the symbol.

So a report can name the right gene 79 times, cite 18 real papers, report zero confabulation, pass preflight — and still hand a curator an identifier for a different gene entirely. The identifier is the part a curator is most likely to copy verbatim into `gene_term.term.id`, because it looks like exactly the kind of thing that has already been checked.

In this case the KB's own `just validate-terms` would have caught it downstream (`hgnc:29316` has label `ZSWIM6`, not `ZMYND11`, so the label check fails). But that is a late catch, after the ID is already in the YAML, and it only works because dismech requires the label alongside the ID. A curator copying the ID into an issue, a PR description, or a `notes` field gets no check at all.

### Suggested fix

[...]

Issue #9702, opened 2026-08-27
`--validate-references` silently produces no `reference_validation` block on most deep-research reports


Phillip-a-richmond:
`just research-disorder` passes `--validate-references --validation-cache-dir references_cache` — I confirmed the flags reach the process:

```
uv run python -c ... research --template templates/disease_pathophysiology_research.md
  --var disease_name=Charcot-Marie-Tooth Disease Type X ... --provider openscientist
  --output research/...openscientist.md --separate-citations ...
  --validate-references --validation-cache-dir references_cache
```

The run completed successfully (exit 0, 26 citations, 1013s), but the report has neither a `reference_validation:` frontmatter block nor a `## Reference Validation` body section. No warning was printed.

This is not specific to my run. Across `research/*-deep-research-openscientist.md`, only 21 of 116 reports carry the block:

```bash
grep -l "^reference_validation:" research/*-deep-research-openscientist.md | wc -l   # 21
ls research/*-deep-research-openscientist.md | wc -l                                 # 116
```

## Why this is worth fixing rather than tolerating

The `initiate-new-disorder-creation` skill makes reading these results a required step, and says the answer is already in the report:

> ### Step 3a: Read the report's Reference Validation results (REQUIRED)
> **The answer is already in the report — read it before you cite anything from that report.**

When the block is absent, a curator following that instruction finds nothing and has no signal distinguishing "validation ran and everything passed" from "validation never ran". The anti-confabulation gate fails open and quietly.

That is not hypothetical for this report. It cited PMID:42017539 in support of a serum-NfL claim about CMTX carriers; the paper is about *NEFL*-associated CMT, a different gene. That is exactly the off-topic/relevance case the `relevance_assessed` / `off_topic_references` fields exist to flag, and no flag was raised. I caught it by fetching each PMID and reading the title, which is the manual fallback — but the fallback only happens if the curator already distrusts the report.

## Suggested fix

Two parts, either useful alone:

1. Work out why validation is skipped for most runs (provider-specific? citation-format-specific? a silently swallowed exception?) and fix it.
2. Failing that, make the skip **loud** — emit a `reference_validation: {status: skipped, reason: ...}` block or print a warning, so absence is legible as absence rather than as a pass.

Found while curating Charcot-Marie-Tooth disease type X (https://github.com/monarch-initiative/dismech/pull/9700).

aleixpuigb:
[...]

PR #10936, opened 2026-09-04
curate: Add Hearing Loss Autosomal Dominant 83 (MONDO:0030723, MAP1B)

aleixpuigb:
New entry for DFNA83, dominant nonsyndromic hearing loss caused by heterozygous *MAP1B* missense variants. **The reason it is worth curating carefully is that the lesion is not in the hair cell.** Almost every dominant deafness gene in this knowledge base acts on the cochlear sensory epithelium; this one acts on the spiral ganglion neuron, and the evidence for that is a negative result rather than a positive one.

Distortion product otoacoustic emissions, which report outer hair cell function, were indistinguishable between affected and hearing-normal individuals in the index family, and stayed normal in the mouse at 4, 8, 16 and 32 weeks while auditory brainstem response thresholds were already raised. Cochlear histology in the mouse showed no hair cell loss and no change in spiral ganglion neuron density. Raised ABR thresholds with preserved emissions is the audiological signature of auditory neuropathy, and the entry is built around it: three `UNCHANGED` model readouts carry that negative evidence explicitly rather than leaving it in prose.

### Mechanism

```
MAP1B p.Ser1400Gly (MTA-domain phosphosite)
        │
        ▼  63% of normal protein, 69% of normal type I phosphorylation
Deficient MAP1B phosphorylation
        │
        ▼  acetylated α-tubulin at 168% of control, resistant to vinblastine
Microtubule HYPERstabilisation at the growth cone
        │
        ├──▶ neurites 30% shorter
        └──▶ reduced K+ current, smaller/slower/blunter action potentials
                    │
                    ▼
        Auditory nerve transmission fails — while the cochlea still works
```

The counterintuitive step is the third: losing a microtubule-associated protein makes the cytoskeleton *more* stable, not less, because phosphorylated MAP1B is what maintains the dynamic pool. Growth cone steering needs microtubules that can depolymerise, so a stabilised cytoskeleton is the defect.

### Evidence quality

Unusually good for a two-paper locus, because the founding report carried its own isogenic rescue. Patient-derived iPSCs were differentiated into otic sensory neuron-like cells, the allele was corrected by CRISPR/Cas9 in a sister line, and every measured defect returned toward control. That is a within-experiment causality argument, curated as a `RESCUES` `modeled_mechanisms` link rather than as an association.

The limits are recorded rather than glossed:

[...]

PR #10930, opened 2026-09-04
curate: Add Hearing Loss Autosomal Dominant 75 (MONDO:0032911, TRRAP)

aleixpuigb:
New entry for DFNA75, an autosomal dominant nonsyndromic sensorineural hearing loss attributed to heterozygous *TRRAP* variants. High frequencies go first and the loss extends downward over years. The whole published evidence base is two families and five alleles, so the interesting curation question here was not what the mechanism is but how much of it the literature actually supports — and the answer is: the two ends, not the middle.

**What is observed** is a hair cell and stereocilia deficit in trrap-depleted zebrafish, and a cochlear-pattern audiogram in humans. **What is inferred** is everything between them: that the human missense alleles impair TRRAP-dependent recruitment of the SAGA and NuA4/TIP60 histone acetyltransferase complexes, and that the resulting transcriptional change is what the cochlea cannot tolerate. No functional assay has ever been run on a DFNA75 allele. The second reporting group states that inference as an assumption needing further study, so the entry carries it as a labelled hypothesis rather than a curated causal chain.

Concretely, the three inferential nodes carry `mechanism_confidence: HYPOTHETICAL` (one is `PROVISIONAL`), their edges are typed `causal_link_type: UNKNOWN`, and all of them opt into an `EMERGING` `mechanistic_hypotheses` group. A reader who only looks at the pathograph should still be able to see which arrows are load-bearing and which are proposals.

### What is in the entry

- **Pathophysiology**: five nodes from the TRRAP allele to the audiogram, tagged `MOLECULAR` → `MOLECULAR` → `CELLULAR` → `TISSUE` → `ORGANISM`.
- **Animal model**: the zebrafish knockdown/knockout from the founding report, the only animal work on this gene-disease relationship. It carries two typed `divergences`. `PROXY_QUANTITY` is marked `INVALIDATING` because the experiment manipulates gene dosage to near zero while every human allele but one is a heterozygous missense change of unestablished consequence — complete depletion is not a proxy for a missense allele unless loss of function has been shown, and it has not. `SPECIES_MISMATCH` records that the quantified structure is a lateral line neuromast, an organ humans do not have.
- **Three knowledge gaps**, each of them a real open question rather than a generic call for more research:
[...]

PR #10937, opened 2026-09-04
curate: Add Methylmalonic Aciduria and Homocystinuria cblL Type (MONDO:0975798, THAP11)

### LINKED ISSUE #10928: Curate methylmalonic aciduria and homocystinuria, cblL type (MONDO:0975798)
Curate a dismech entry for **methylmalonic aciduria and homocystinuria, cblL type** ([MONDO:0975798](https://monarchinitiative.org/MONDO:0975798)) — MAHCL, an autosomal recessive disorder of intracellular cobalamin metabolism caused by biallelic *THAP11* variants (OMIM:620940), presenting in infancy with seizures and profoundly impaired neurodevelopment.

Preflight notes:

- MONDO records no descendants. The MONDO label carries a typo (`cb1L`, digit one) that the entry should not reproduce.
- The lump/split call here needs care and is the main thing to settle before writing. `kb/disorders/Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport.yaml` already names *THAP11* inside its `cblX` subtype description as a "cblX-like" form. But cblL is autosomal recessive and cblX is X-linked *HCFC1*, so folding them together would merge two inheritance modes under one subtype. The KB's own precedent points the other way: cblC, cblD, cblE and cblG each have their own `kb/disorders/` entry. Expect to correct the existing cblX description as part of this work rather than inheriting its framing.
- The mechanism is transcriptional rather than enzymatic — *THAP11* (Ronin) is a transcriptional regulator of *MMACHC*, so the metabolic block is downstream of a gene-regulatory lesion. That is worth modelling explicitly instead of collapsing it into "MMACHC deficiency".
- Evidence base is narrow; curate at the strength that exists.
- No existing `kb/` entry carries MONDO:0975798 as its `disease_term`, and no open PR covers it.

Stub: `stubs/Methylmalonic_Aciduria_And_Homocystinuria_cb1L_Type.yaml`

aleixpuigb:
New entry for cblL (MAHCL), a disorder of intracellular cobalamin metabolism caused by biallelic *THAP11* variants. **Its lesion is transcriptional, not enzymatic.** THAP11 and HCFC1 form the complex that drives *MMACHC* expression, so losing THAP11 function starves the cell of the cobalamin chaperone without touching the *MMACHC* gene. The result is a cblC phenocopy in a patient whose *MMACHC* sequence is normal.

```
THAP11 p.Phe80Leu (homozygous, THAP domain)
        │
        ▼   measured: reduced binding to the MMACHC promoter, selectively
Reduced MMACHC transcription      ── MMACHC gene itself is intact ──
        │
        ▼
Loss of the cytosolic cobalamin chaperone
        │
        ▼   both cofactors starved at once, because MMACHC is upstream of the branch
Methylmalonyl-CoA mutase (AdoCbl) + methionine synthase (MeCbl) deficiency
```

The one place the causal claim is directly measured is the second step, and it is measured on the patient's own allele: in HEK-293 cells the F80L protein bound a *subset* of THAP11 targets less well, with the *MMACHC* promoter among the strongly affected ones. That selectivity is what lets a general transcription factor variant produce a pathway-specific disease, so it is curated as edge-level evidence rather than as node evidence.

Everything downstream of MMACHC is inherited from cblC biology and has not been measured in a MAHCL patient. Those two nodes carry `mechanism_confidence: PROVISIONAL` so the inference is visible in the data and not only in the prose.

### The lump/split call

**cblL is curated as its own entry rather than as a subtype of `Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport`.** Two reasons:

1. **Inheritance.** cblL is autosomal recessive (biallelic *THAP11*); cblX is X-linked (hemizygous *HCFC1*). An entry merging them could not carry a coherent inheritance block or support correct recurrence-risk counselling, which is the thing families actually need from it.
2. **Precedent.** cblC, cblD, cblE and cblG each already have their own `kb/disorders/` entry.

The literature has not fully caught up: GeneReviews still calls *THAP11* "cblX-like" rather than giving it a complementation letter, while OMIM has assigned it #620940 and the cblL name. Both halves of that tension are quoted in the entity-boundary discussion.

**One consequence I have deliberately not bundled here.** `kb/disorders/Inborn_Disorder_of_Cobalamin_Metabolism_and_Transport.yaml` describes *THAP11* inside its `cblX` subtype as a "cblX-like form". That was accurate when written and is now superseded by this entry. It is a shared file that other open PRs may touch, so I left it alone and recorded the staleness in the discussion instead. Say the word and I'll do it as a separate small PR.

### A gene-identifier trap worth knowing about

The deep-research report offered **`hgnc:20856` for THAP11**. That is **THAP1**, the DYT6 dystonia gene, a different member of the same family with nothing to do with cobalamin metabolism. The entry uses `hgnc:23194`.

**This section previously said the KB-side term validator skips HGNC as well. That was wrong, and I have corrected both this description and the entry's own `notes`.** What is actually the case:

- **The research lane does skip it.** `just research-disorder` passes `--term-skip-prefix HGNC` (`project.justfile:1826`, case-insensitive so it covers `hgnc:` too), so the report's own Term Validation section reports gene CURIEs as unverifiable rather than checking them. That is why the report offers `HGNC:20856` as THAP11 in four places with nothing flagging it.
- **The KB lane does not.** HGNC is configured in `conf/oak_config.yaml`, and `just validate-terms` compares a bound label against the ontology's canonical label, so copying the report's identifier together with its offered label fails:

  ```
  ⚠️  WARN: Label mismatch for 'hgnc:20856': expected 'THAP1', got 'THAP11'
  ```

- **The gap that does survive is narrower.** A self-consistent binding, `hgnc:20856` carrying its own true label `THAP1` under a `name` and `preferred_term` of THAP11, passes both `linkml-validate` and `validate-terms`, because nothing compares the gene a binding resolves to against the gene the entry names. Tested directly, and filed as #10948. The perverse part is that filling the label correctly from the ontology, which is the responsible-looking move, converts a caught error into a silent one.

I caught the original by resolving the symbol against the HGNC REST API, which is not a step any skill currently asks for. Adjacent to #9845, which covers the report-side half.

The same report's other bad bindings, none used:

| Report offered | Ontology actually says |
|---|---|
| `CHEBI:16009` as "methylmalonic acid" | *9-riburonosyladenine* |
| `NCIT:C61796` as "Betaine" | *Ivermectin* |
| `NCIT:C61837` as "folinic acid" | *Methacholine* |
| `NCIT:C61785` as "Levocarnitine" | *Hydrocortisone Acetate* |

Like the DFNA75 run in the same batch, this report emitted no `reference_validation` or `term_validation` block despite the flags; both were retro-fitted. See my comment on #9702.

The report also gives this allele a 1000 Genomes MAF of `~0.0032`. That figure is **not** the pathogenic allele's: `rs188675529` is multi-allelic (C>G and C>T), the pathogenic allele is C>G, and 0.0032 belongs to C>T (GnomAD exomes: T=0.00055). The entry carries no MAF rather than recording one against the wrong allele.

### Three knowledge gaps, all real

- **One patient, one allele.** No second family, so no penetrance, no expressivity, no allelic spectrum, and no genotype-phenotype statement is possible. The entry looks better evidenced than the disease is, which is why this is stated rather than left to be inferred from the citation count.
- **The disease is named for a pairing its only patient did not show.** OMIM's clinical synopsis for that patient records mild methylmalonic aciduria *without* homocystinuria. That may be mechanistically meaningful — partial, selective loss of *MMACHC* transcription could plausibly starve adenosylcobalamin before methylcobalamin — or it may be one atypical case. It also has a practical edge: a workup finding isolated methylmalonic aciduria will not generate a differential that includes this group.
- **No reported treatment outcome.** The hydroxocobalamin entry is group management applied by inference, and there is a specific reason not to assume it transfers: in enzymatic cblC the chaperone is present but defective, whereas here it is scarce but normal, and saturating a reduced pool of working chaperone is not obviously the same as saturating a normal pool of broken one.

### Validation

Current head `3d1b11c`:

```
just validate-disorders kb/disorders/Methylmalonic_Aciduria_and_Homocystinuria_cblL_Type.yaml   # 41/41 snippets verified
just check-entity-refs / check-causal-targets / check-duplicate-keys / check-qualifier-terms     # clean
just check-snippet-boundaries <file>                                                            # clean
just check-snippet-length / check-title-snippets / check-folded-hyphens                         # no new findings
just check-snippet-grading / check-enum-values                                                   # no new findings
just validate-history <record>                                                                  # no issues
```

Note the MONDO label carries a typo (`cb1L`, digit one). The `disease_term` is bound to the canonical label as-is; every other field in the entry spells it `cblL`.

Closes #10928

https://claude.ai/code/session_01DqyZyY8RRPi72hYMQVGcRT

---
_Generated by [Claude Code](https://claude.ai/code/session_01DqyZyY8RRPi72hYMQVGcRT)_


aleixpuigb:
All six blocking items and all five suggestions are in one push (`4e29f03b7`). Your framing was right and worth repeating: I curated the mechanism half and left the clinical half unbuilt, and the tell was 13 of 18 committed caches going uncited.

**Item 4 is the one that mattered, and you were right that it was a real error.** The entry asserted that the craniofacial phenotype is "a developmental phenotype that no amount of cobalamin deficiency explains", and PMID:25281006 — cached by this PR, uncited — reports that exactly that phenotype is rescued by human *MMACHC*. Its **title** says the same thing outright: *"Hcfc1b, a zebrafish ortholog of HCFC1, regulates craniofacial development by modulating mmachc expression."* I had committed a paper whose title refutes a claim in the same PR.

The node description now says the developmental branch is less independent than it looks, that at least part of it runs back through the same downstream gene as the metabolic lesion, and that the ribosome-biogenesis arm is the clearest remaining candidate for a genuinely MMACHC-independent contribution — since no MMACHC rescue has been reported for the ribosomal phenotypes. The paper is cited there with `supports: REFUTE` against the strong reading, rather than the claim being quietly softened and the evidence left out.

**The rest**

1. **Treatments** — betaine, folinic acid, levocarnitine and dietary management added, citing PMID:23746552 and GeneReviews. Each says no treatment response has been reported for a MAHCL patient. One thing I added that the review did not ask for: the betaine entry notes that its rationale targets the homocysteine arm, which the one reported patient's OMIM synopsis says was not raised — so it is the part of the group regimen whose applicability here is least clear.
2. **GeneReviews** — the 25/50/25 recurrence-risk sentence is now an evidence item on `inheritance`, which was the sharpest omission for the reason you gave. The agents-to-avoid list is quoted in full in the dietary entry, with prose on the two a clinician unaware of the diagnosis would get wrong: nitrous oxide, which irreversibly oxidises cobalamin's cobalt, and methionine restriction, which restricts the amino acid the patient is already short of.
3. **Diagnosis** — the block landed in the push just before your review checked out. This round removes the testing sentence from the phenotype, which now records that its source is the OMIM synopsis and that it carries no evidence item rather than one attesting to something else.
5. **Decyanation/dealkylation** — now cites PMID:19801555 in two items, one for each half of the chemistry.
6. **Structured variant** — `NM_020457.3:c.240C>G (p.Phe80Leu)`, `LIKELY_PATHOGENIC`. Not `PATHOGENIC`: the reporting authors called it *potentially* pathogenic, and there is no second family.

**Suggestions taken, with two corrections to the CURIEs you offered.** I checked both against OLS before using them, which is the habit this entry exists to argue for:

| You suggested | OLS says it is | Used instead |
|---|---|---|
| `HP:0500152` for hypomethioninemia | **Hypocystinemia** | `HP:0003658` Hypomethioninemia |
| `HP:0002123` for myoclonic seizures | **Generalized** myoclonic seizure | `HP:0032794` Myoclonic seizure |

The second is a judgement rather than an error on your part: `HP:0002123` exists and is plausible, but nothing in the sources says the seizures were generalized, so it would have been more specific than the evidence. Onset is captured as `onset_category: INFANTILE`. ZNF143 (`hgnc:12928`) is in `genetic:` as the nearest differential. Prevalence landed in the earlier push.

Worth noting for the record: reference validation also caught three reference titles I had written from memory rather than read from the cache, including PMID:25281006's. Same class of error as the CURIEs above, caught by a different gate.

**Validation**

```
just validate-disorders kb/disorders/Methylmalonic_Aciduria_and_Homocystinuria_cblL_Type.yaml   # 41/41 snippets verified
just check-entity-refs / check-causal-targets / check-duplicate-keys / check-qualifier-terms     # clean
just check-qualifier-terms-online / check-snippet-boundaries                                    # clean
just check-snippet-length / check-title-snippets / check-snippet-grading                        # no new findings
just check-folded-hyphens / check-enum-values                                                   # no new findings
just validate-history <record>                                                                  # no issues
```


---

aleixpuigb:
I pushed after your approval, which drops it. Explaining why, because it was my choice.

**A sentence I wrote in the THAP11 `notes` was wrong.** It said HGNC is on the term validator's skip list and that nothing in the pipeline would have caught the research report's `HGNC:20856` suggestion. Both halves are false, and I only found out because I went to file an issue about it and tested the claim first.

HGNC is configured in `conf/oak_config.yaml` as `sqlite:obo:hgnc`. Injecting the bad identifier under the label I had:

```
❌ Validation failed with 1 issue(s):
  ⚠️  WARN: Label mismatch for 'hgnc:20856': expected 'THAP1', got 'THAP11'
      path: genetic[0].gene_term.term
```

The report offers `HGNC:20856` as *THAP11* in four places, so copying it as offered would have been caught. My note claimed credit for a catch the tooling would have made.

**What does pass is narrower, and I tested that too.** Set `id: hgnc:20856` with its own true label `THAP1`, leave `name: THAP11` and `preferred_term: THAP11` above it, and both `just validate` and `just validate-terms` report clean. Nothing compares the gene a binding resolves to against the gene the entry says it is about. The corrected paragraph says that instead. Filed as #10946.

I judged this worth a cycle rather than leaving for later. It is a confident, specific, wrong statement about repository tooling sitting in curated prose, where the next person to read it has no reason to doubt it, and after merge it is somebody else's problem to discover.

**Folded in your residual since the file was open.** `ClinVar:VCV000393304` and `dbSNP:rs188675529` on the variant, both verified against the primary records rather than taken from the report. ClinVar resolves to `NM_020457.3(THAP11):c.240C>G (p.Phe80Leu)` on THAP11 and classifies it **Likely pathogenic**, which independently corroborates the `LIKELY_PATHOGENIC` call. dbSNP maps to the same SPDI coordinate with the C>G allele.

**One thing I left out on purpose.** The report's `~0.0032` 1000 Genomes MAF is attached to the wrong allele: rs188675529 is multi-allelic (C>G and C>T), and 0.0032 is the **T** frequency, not the pathogenic **G** one. GnomAD exomes gives T=0.00055. The entry never carried the MAF and still doesn't, rather than recording it against the wrong allele.

Validation on `3d1b11c0a`: `just validate` clean, 41/41 snippets verified, `check-entity-refs` / `check-causal-targets` / `check-duplicate-keys` / `check-enum-values` / `check-folded-hyphens` clean, `validate-history` clean on the new record.

---

aleixpuigb:
Correction to my comment above: the issue is **#10948**, not #10946. I wrote the number before filing and guessed it wrong.

---

aleixpuigb:
Description updated, which costs no approval. Thanks for pushing on it, and for pulling `project.justfile:1826` out; that is the piece I was missing and it makes the whole thing clearer than my correction did.

**You were right that I had over-corrected.** My commit message said the old note was wrong "on both halves". It was not. The skip-list half is true of the *research* lane, and I had merged two lanes with different guarantees into one claim, in both directions:

| Lane | HGNC checked? |
|---|---|
| `just research-disorder` | **No.** `--term-skip-prefix HGNC`, case-insensitive, so the report's Term Validation section reports gene CURIEs as unverifiable |
| `just validate-terms` on a `kb/` file | **Yes,** label against ontology label |

So the original note's first sentence was defensible and its second, "nothing in the pipeline would have caught it", was the false part. The description now says exactly that, with the justfile reference.

I also corrected the same over-statement on #10948, where I had written that HGNC "is not skipped anywhere". The nuance strengthens that issue rather than weakening it: the one place a gene identifier is emitted fresh is the place that cannot check it, and the downstream check that exists is the one a careful curator satisfies while still being wrong, because filling the label correctly from the ontology is what converts a caught error into a silent one.

**On the identifiers you could not resolve.** Fair to record as a limit. For the record, both were checked against NCBI eutils before committing, not against the report:

- `VCV000393304` → `NM_020457.3(THAP11):c.240C>G (p.Phe80Leu)`, gene THAP11, SPDI `NC_000016.10:67842793:C:G`, germline classification **Likely pathogenic**. That last is a nice independent corroboration of the entry's `LIKELY_PATHOGENIC`, which I had justified from the authors' hedged wording alone.
- `rs188675529` → THAP11 at the same SPDI coordinate, carrying the C>G allele.

The MAF caveat is now in the description too, since that trap belongs where the next reader will hit it rather than only in a review thread.

---

aleixpuigb:
Merged `origin/main` in to clear a conflict that appeared overnight. This drops the approval again; that is twice from me on this PR, and neither was your doing.

**The conflict was one line in a derived cache.** `cache/enums/diseaseterm_f618932bb23f.csv`: this branch adds `MONDO:0975798` (cblL), main added `MONDO:0975832`. Kept both rows per the `CLAUDE.md` cache rule rather than picking a side; they were already in canonical order, so `just normalize-cache` made no change, and `check-cache-order` and `check-term-cache-integrity` both pass.

**Nothing in the entry moved** — the merge does not touch `kb/disorders/Methylmalonic_Aciduria_and_Homocystinuria_cblL_Type.yaml`. Merge commit, not a rebase, so `3d1b11c0a` and the history record are untouched.

Validation on `b155e5401`, run against the merge result rather than the pre-merge tree: `just validate-disorders` clean with 41/41 snippets verified; `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values` all clean.

The same conflict hit #10930 and #10936 in the same sweep, all three in `diseaseterm_f618932bb23f.csv`. Three curation PRs each adding one MONDO row to a shared sorted cache will conflict with every other one that lands first, which is #9804's shape one file over. Not worth an issue on its own unless it keeps recurring, but noting it in case you are seeing it across other branches this morning.

---

aleixpuigb:
`claude-review` is red on `b155e5401`, same cause as #10930 and #10936, and not this PR. **The reviewer account is out of quota**, so no review ran:

```
error='' detail="You've hit your limit · resets 7:20am (UTC)"
```

Full diagnosis is in [my comment on #10930](https://github.com/monarch-initiative/dismech/pull/10930#issuecomment-5549355777) rather than repeated across three PRs.

**This completes the picture: all three re-reviews triggered at 04:29–04:33Z were refused, in every case at one turn and zero cost.** So this is the reviewer credential being exhausted repo-wide for the window, not anything specific to a branch. Any other PR pushed in that period will have the same red check.

**Not re-running yet** — the reset is 07:20 UTC, currently ~04:35 UTC. I will re-run all three after the reset and report verdicts. If the quota is still exhausted then, that needs a person rather than another re-run and I will escalate rather than keep retrying.

`test (3.13)` is green. The merge commit validates clean: 41/41 snippets verified, plus `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys` and `check-enum-values` across the whole KB.

---

Issue #9496, opened 2026-08-25
HP:0001522 'Death in infancy' and the Mortality/Aging branch are outside PhenotypeTerm, with no guidance on where such facts belong


kevinschaper:
## What happened

While curating `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_2` (#9491) I added a phenotype for the index patient's death in the first months of life, bound to **`HP:0001522 Death in infancy`**. The term exists and `runoak -i sqlite:obo:hp info HP:0001522` returns it normally.

`just validate-terms` rejected it:

```
❌ ERROR: Value 'HP:0001522' not in dynamic enum 'PhenotypeTerm' (expanded from ontology)
    path: phenotypes[5].phenotype_term.term
```

The cause is that `HP:0001522` sits under **`HP:0040006 Mortality/Aging`**, which is a sibling of `HP:0000118 Phenotypic abnormality` rather than a descendant, so it is correctly outside the `PhenotypeTerm` `reachable_from` closure.

**The gate is right.** Death is not a phenotypic abnormality, and I would not want the enum widened to admit the whole Mortality/Aging branch — that would also let in `HP:0011420 Miscarriage`, `HP:0001623 Breech presentation` and similar, which are not phenotypes of the disease either.

## The actual gap: nowhere obvious for the fact to go

Early death is a real, clinically important, frequently-reported fact about severe Mendelian disease, and it appears in abstracts constantly ("died at 11 weeks", "death in the first months of life"). A curator who hits this error has no documented answer to *"where does this belong instead?"*

I resolved it by moving the content to a **`progression:` record** with `phase: Infantile-onset course`, which I think is right — a course of illness ending in death is a progression fact, not a phenotype — and the same entry now carries a matching `phase: Late-onset course` record for contrast. But I arrived at that by trial and error, not by following guidance.

## What would help

1. **A line in the `dismech-terms` skill**, or in the CLAUDE.md phenotype section: *terms under `HP:0040006 Mortality/Aging` are deliberately outside `PhenotypeTerm`; record survival and age-at-death facts under `progression:` (and `clinical_burden:` where relevant), not as phenotypes.*
2. **A better error message**, in the same spirit as the retired-NCIT case in #9495: when a rejected HP term is reachable from `HP:0040006`, say so and name `progression:` as the intended home. Both checks are one extra ancestor query on the failure path.
[...]

PR #10939, opened 2026-09-04
curate: Add Mitochondrial Complex I Deficiency, Nuclear Type 39 (MONDO:0859320, NDUFB7)

### LINKED ISSUE #10927: Curate mitochondrial complex I deficiency, nuclear type 39 (MONDO:0859320)
Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 39** ([MONDO:0859320](https://monarchinitiative.org/MONDO:0859320)) — MC1DN39, an autosomal recessive isolated complex I deficiency caused by biallelic *NDUFB7* variants (OMIM:620135), reported with intrauterine growth restriction and anemia followed by postpartum hypertrophic cardiomyopathy, lactic acidosis and encephalopathy.

Preflight notes:

- MONDO records no descendants; the parent is `MONDO:0100223` mitochondrial complex I deficiency, nuclear type.
- This is in-pattern: the KB already carries nuclear types 2, 6, 12, 23, 24 and 34 as separate `kb/disorders/` entries, so a per-gene entry is the established granularity here.
- Evidence base is narrow (a single reported family, with cell-based complementation), so the honest job is to curate the mechanism at the evidence strength that actually exists rather than inflating it.
- No existing `kb/` entry or open PR carries MONDO:0859320 or *NDUFB7*.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml`

aleixpuigb:
New entry for MC1DN39, isolated mitochondrial complex I deficiency from biallelic *NDUFB7* variants, joining nuclear types 2, 6, 12, 23, 24 and 34 already in the KB.

One patient has been reported, and normally that would make for a thin entry. This one is worth having because **the causal argument is unusually complete for a single case**, and because it is a good worked example of a variant class that routine pipelines throw away.

The variant is intronic, ten bases upstream of an acceptor site — not coding, not a canonical splice dinucleotide, and exactly what a coding-region filter discards. What makes it convincing is that every step was measured rather than predicted:

| Step | How it was established |
|---|---|
| The variant changes splicing | RNA sequencing of patient material |
| Splicing change reduces protein | immunoassay |
| Reduced protein reduces the enzyme | complex I enzymology |
| It is *this gene* doing it | wild-type *NDUFB7* expressed in the patient's own fibroblasts normalised complex I function |

That last row is a rescue inside the patient's own genetic background, so it is not vulnerable to the usual objection that some other variant in that genome caused the defect. It is curated as a `RESCUES` `modeled_mechanisms` link rather than left as prose, because it is the entry's causality argument and not a supporting detail.

### What is inferred, and marked as such

Two nodes carry `mechanism_confidence: PROVISIONAL`.

**The assembly step.** NDUFB7 is an accessory subunit, not one of the fourteen core subunits, so the route from less protein to a broken enzyme runs through assembly. The evidence for that is Stroud et al.'s knockout screen of every accessory subunit, which found that losing any one destabilises the others in its own structural module and that 25 of 31 are strictly required. But the cached record states the rule across subunits without naming NDUFB7's module or its individual result, and no complexome profile exists for a MC1DN39 patient. So the entry applies a strong prior, not a finding — and six accessory subunits are *not* strictly required, which is enough to matter. A knowledge gap notes that the screen's supplementary data almost certainly already contain the answer.

**The tissue-selectivity step.** Heart and brain failing first is standard mitochondrial-medicine reasoning, not a measurement in this patient. No cardiac or brain enzymology was reported.

### Prognosis is deliberately not stated

The index patient died, and a single fatal case reads as a prognosis unless something stops it. The 130-case nuclear complex I deficiency series is what stops it: a quarter died before six months and three quarters before ten years, but some recovered skills and some were alive in their thirties, and **no clinical, biochemical or genetic parameter predicted longer survival**. A range that wide with no predictor means one observation says very little about the next patient. There is also a gene-specific reason to expect variability — the reported allele is a splicing variant producing a quantitative rather than absolute protein reduction.

### A schema constraint worth flagging

I could not curate the fatal outcome as a phenotype. `HP:0003811` **Neonatal death** sits in HPO's mortality branch rather than under *Phenotypic abnormality*, so it is outside the `PhenotypeTerm` dynamic enum and `just validate` rejects it:

```
❌ ERROR: Value 'HP:0003811' not in dynamic enum 'PhenotypeTerm' (expanded from ontology)
```

For this entry that is arguably the right outcome — one patient's death is an observation, not a disease feature — so the outcome lives in the prognosis discussion instead, and I have documented why there rather than working around it. But it generalises past this entry: mortality and clinical-course terms cannot be bound anywhere in a `Disease`, and a lethal Mendelian disorder has no structured way to say so. If that is a gap worth an issue rather than a per-entry note, say so and I'll open one; I did not want to file against a design decision that may well be deliberate.

### Deep research

One OpenScientist run, committed with its artifacts. This one *did* emit both validation blocks and both sections (11/11 references resolved, 8/11 on topic; 37 terms checked, 1 obsolete, 4 labels mismatched).

Three of those four label mismatches are **false positives worth knowing about**: the report's phenotype table has a `Type` column next to the HPO column, and the validator read that column as the term's label, so it reported `HP:0001639` as "Sign/imaging" and `HP:0001903` as "Lab abnormality". The CURIEs are correct and are used. Not a bug I'd file on its own, but a reason not to read a mismatch count as a defect count without opening the section.

Note the report also produced no `.citations.md` sidecar, unlike the other three runs in this batch.

### Validation

```
just validate-disorders kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml   # 25/25 snippets verified
just check-entity-refs / check-causal-targets / check-duplicate-keys / check-qualifier-terms    # clean
just check-snippet-boundaries <file>                                                           # clean
just check-snippet-length / check-title-snippets / check-folded-hyphens                        # no new findings
just check-snippet-grading / check-enum-values                                                  # no new findings
just validate-history <record>                                                                 # no issues
```

Closes #10927

https://claude.ai/code/session_01DqyZyY8RRPi72hYMQVGcRT

---
_Generated by [Claude Code](https://claude.ai/code/session_01DqyZyY8RRPi72hYMQVGcRT)_

### DIFF
--- a/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39/2026-09-04T152924Z-claude-code-b01f76.yaml
+++ b/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39/2026-09-04T152924Z-claude-code-b01f76.yaml
@@ -1,0 +1,26 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39
+  path: kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml
+session:
+  id: 2026-09-04T152924Z-claude-code-b01f76
+  timestamp: '2026-09-04T15:29:24Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10939
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - description
+  summary: 'Correct a history/tree mismatch: MC1DN39 GeneReviews paragraph'
+  details: |
+    The reviewer on PR 10939, while approving, noticed that the history record for the previous round claimed to add a paragraph recording why PMID:26425749 is tagged as GeneReviews but not cited as evidence, and that no such paragraph existed in the entry. That is correct and the cause was mine: the edit was applied with a string replacement whose anchor did not match, and unlike the other replacements in the same script that one carried no assertion, so it failed silently and the round was validated and pushed without it. The GeneReviews tag itself did land; only the explanatory paragraph was missing. This round adds the paragraph to the entry description, so the previous record's details are now true of the tree. That is the reason for pushing to an already-approved pull request and accepting the loss of the approval: a history record is append-only and, once merged, a record describing a change that is not in the tree cannot be corrected. Choosing between amending the earlier record and making the tree match it, making the tree match is the better outcome, because the paragraph is content the reviewer asked for in the first place and the entry is better for having it. The paragraph records that the chapter is tagged following sibling entries MC1DN2, MC1DN6 and MC1DN34, that its cached record carries only the chapter's statement of purpose so any snippet would attest to the chapter's aims rather than to a fact about NDUFB7 disease, and that the index patient's cardiac-predominant congenital presentation is not a Leigh phenotype. The reviewer also corrected their own earlier count of uncited caches in this pull request, from six to zero: four cache files are added by this branch and all four are cited, and PMID:26425749 was already present in main. No action follows from that correction. Validation after this change: just validate passed with 33/33 snippets verified; check-entity-refs, check-duplicate-keys, check-snippet-boundaries and check-folded-hyphens all clean.
--- a/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml
+++ b/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml
@@ -29,6 +29,15 @@ description: >-
   entry marks it as a general principle: the screen reports the rule across subunits and
   does not, in the record cached here, single out NDUFB7.
 
+  On GeneReviews. PMID:26425749 ("Nuclear Gene-Encoded Leigh Syndrome Spectrum Overview") is
+  the applicable chapter for this gene class and is tagged in the top-level references block,
+  following sibling entries MC1DN2, MC1DN6 and MC1DN34. It is deliberately not cited as
+  evidence anywhere: its cached record carries only the chapter's statement of purpose, with
+  no phenotype frequencies and no management detail, so any snippet from it would attest to
+  the chapter's aims rather than to a fact about NDUFB7 disease. Note also that the index
+  patient's presentation was cardiac-predominant and congenital rather than a Leigh
+  phenotype, so the chapter is the right gene family but not a description of this case.
+
   What one patient cannot support is prognosis. The published nuclear complex I deficiency
   series of 130 patients found a wide survival range with no clinical, biochemical or
   genetic parameter predicting it, so this entry does not present the index patient's fatal


aleixpuigb:
All three blocking items and all five suggestions are in one push (`b38c00c5a`). You were right about the shape of the problem: I curated the mechanism half well and left the clinical half unbuilt, and the giveaway was exactly what you spotted — a therapy review sitting in the PR's own caches, uncited.

**Blocking items**

1. **`diagnosis:`** — landed in the push just before your review checked out, so you would not have seen it. It covers the enzymology-then-WGS-then-RNA-seq route and makes the point your comment does: exome-only analysis returns nothing in this patient.
2. **`treatments:`** — two entries. *Supportive and Symptomatic Management* (`NCIT:C15747`) and *Empiric Mitochondrial Cofactor Supplementation* (`NCIT:C15986` with riboflavin, coenzyme Q10 and carnitine as CHEBI agents). Both cite PMID:28943110, graded `OTHER` because it is a narrative review of the disease class rather than a study of this disease, and both state that nothing has been reported for a MC1DN39 patient. I quoted the reviewers' own hedge — *"Based on limited number of clinical trials…"* — rather than paraphrasing it, because the hedge is the content.
3. **GeneReviews** — the tag landed in the same earlier push; this adds the `notes` paragraph you asked for, following Type_34: tagged, deliberately not cited as evidence, and why. I added one thing beyond the template: the index patient's presentation was cardiac-predominant and congenital rather than a Leigh phenotype, so the chapter is the right gene family but not a description of this case.

**Suggestions, all taken**

- **`biochemical:`** — NDUFB7 protein abundance paired with complex I activity. You are right that the protein measurement is the more interesting one; I have written it as the pair, because what makes them diagnostic rather than descriptive here is the complementation arm restoring both, which is the assay that would resolve a future NDUFB7 VUS.
- **Orphaned phenotypes** — left unconnected, and now each says so. IUGR gets the honest version of your point: reading it as systemic bioenergetic failure is plausible and is what the research report does, but nothing measures that route in this patient or this gene. Anemia gets the stronger version: with one patient, no described mechanism and no precedent in the disease group, drawing that edge would invent the thing that makes the finding interesting.
- **`prevalence:`** — added in the earlier push.
- **Uncited caches** — took the "drop" option for five of them (`20157008`, `25085991`, `26363424`, `26824698`, `34069703`) and consumed `28943110`. I did not add an `animal_models` block from the Ndufs4 and Drosophila papers: neither concerns NDUFB7, and a model linked to this entry's nodes would be modelling complex I deficiency in general, which is what the `PROVISIONAL` marking on the assembly node already says the entry is doing at one remove. Happy to reconsider if you think the class-level models earn their place.

**On the HP:0003811 issue — filed as a comment, not a new issue.** [#9496](https://github.com/monarch-initiative/dismech/issues/9496) already covers it: *"HP:0001522 'Death in infancy' and the Mortality/Aging branch are outside PhenotypeTerm, with no guidance on where such facts belong"*. I added this entry's data point there rather than opening a duplicate. The one thing I think is worth adding to that issue's design question, and which this entry made concrete: *"this patient died"* and *"this disease is lethal in infancy"* are different claims. The first is case-level; the second is a disease-level claim with a frequency and an age, and looks structurally like a `phenotypes` entry — which is presumably why curators keep reaching for the HPO term and hitting the gate.

**Validation**

```
just validate-disorders kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml   # 33/33 snippets verified
just check-entity-refs / check-causal-targets / check-duplicate-keys / check-qualifier-terms    # clean
just check-snippet-boundaries <file>                                                           # clean
just check-snippet-length / check-title-snippets / check-snippet-grading / check-enum-values    # no new findings
just validate-history <record>                                                                 # no issues
```


---

aleixpuigb:
You were right about the residual, and I've pushed the fix (`5d2b4dcdf`) despite it costing the approval. Explaining why, since deliberately dropping an approval needs a reason.

**What happened.** The history record claimed to add a paragraph explaining why `PMID:26425749` is tagged but not cited. No such paragraph was in the entry. The cause was mine and is worth naming precisely: that edit was one of several string replacements in a script, and unlike the others it carried no assertion on the anchor. The anchor did not match, the replacement was a no-op, and the round validated and pushed clean because nothing checks that a text edit actually applied. The tag itself landed; only the explanation was missing.

**Why push rather than let it merge.** Your framing settled it: *"that is the one thing that cannot be corrected after merge."* A history record is append-only. Once merged, a record describing a change that is not in the tree stays wrong permanently, in the one artifact whose whole job is to say truthfully what a session did. Weighed against one re-review cycle, that is not close.

Given the choice between amending the earlier record and making the tree match it, I made the tree match — the paragraph is content you asked for in the first place, and the entry is better with it. It records that the chapter is tagged following MC1DN2, MC1DN6 and MC1DN34, that its cached record is only a statement of the chapter's purpose so any snippet would attest to the chapter's aims rather than to a fact about *NDUFB7* disease, and that the index patient's cardiac-predominant congenital presentation is not a Leigh phenotype.

A new history record documents this round, including the silent-replacement failure, so the sequence is legible rather than looking like a paragraph that appeared from nowhere.

**On your correction about the cache count** — thank you for going back and checking. No action follows; nothing was dropped that shouldn't have been. Worth noting the two reviews in this batch reached opposite conclusions about uncited caches: you asked for consume-or-drop here and I dropped five, while the reviewer on [#10930](https://github.com/monarch-initiative/dismech/pull/10930) asked me to keep all sixteen there, on the grounds that the report's `## Reference Validation` section is only reproducible against them. Both are defensible, and they can't both be the convention. That probably wants settling somewhere durable rather than per-review; happy to open an issue proposing the keep-them rule if you think that's the right answer.

**Validation on the current head**

```
just validate kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_39.yaml   # 33/33 snippets verified
just check-entity-refs / check-duplicate-keys / check-snippet-boundaries              # clean
just check-folded-hyphens                                                            # no new findings
just validate-history <record>                                                       # no issues
```

Nothing else changed — the diff is one description paragraph and one history record.


---

PR #10305, opened 2026-08-31
curate: Add Familial Visceral Neuropathy 2 (MONDO:0030399, ERBB2)

aleixpuigb:
New entry for familial visceral neuropathy 2, an autosomal recessive neurocristopathy. `kb/` had nothing matching "visceral neuropathy" and no ERBB2-associated enteric disease.

## The gene is ERBB2, and I initially got it wrong

Worth leading with, because the trap is easy to fall into and cheap to check. The two familial visceral neuropathy concepts run **opposite** to the gene numbering:

| Concept | OMIM | Gene |
|---|---|---|
| VSCN**1** | 243180 | ERBB**3** |
| VSCN**2** (this entry, MONDO:0030399) | 619465 | ERBB**2** |

I drafted this entry as ERBB3. The stub carried no `genes:` block to contradict me, and the defining paper (PMID:33497358) reports both genes together so its abstract reads as if either fits. The deep-research report flagged it; I then verified against MedGen gene-disease links for each OMIM concept rather than taking the report's word. Correction posted on the claim issue.

## Mechanism

ERBB3 binds neuregulin-1 but is catalytically weak; ERBB2 has no ligand of its own but supplies the kinase. They signal only as a heterodimer, so an ERBB2 kinase defect removes NRG1 signalling as effectively as losing the ligand-binding receptor would. The reported homozygous kinase-domain variant p.(Ala710Val) drastically reduces phosphorylation of *both* receptors — so the lesion is a kinase defect propagating across the dimer, not a defect in receptor abundance.

Downstream: enteric progenitors deprived of NRG1/ERBB signalling leave the gut incompletely colonized → aganglionosis, hypoganglionosis and intestinal smooth muscle abnormalities → congenital gastrointestinal dysmotility, plus extra-intestinal neural crest and peripheral nerve involvement.

## The phenotype list is not the paper's phenotype list

The paper reports eight individuals **across both genes**, and copying that list wholesale would attribute the wrong features to this disease. Arthrogryposis with multiple pterygiums and intrauterine growth retardation belong to the ERBB3 fetuses of families F3 and F4. Only what the single ERBB2 family (F5 — two siblings, consanguineous parents) was reported to have is curated: severe constipation that eased after age five, short-segment Hirschsprung disease in the brother, axonal peripheral neuropathy, ptosis, sensorineural hearing loss, clubfeet.

[...]

PR #10940, opened 2026-09-04
history: record the review-response sessions for DEE-55, IMD79 and ORAI1

aleixpuigb:
Three history records, `history/` only. No KB entry is touched.

## Why

`CLAUDE.md` asks every PR that creates or edits a KB entry to carry a matching history record. The four curation PRs from a recent `/claim-disease` run each shipped a CREATE record, and then went through several review-response commits that changed the entries substantially — new phenotypes, retracted claims, corrected evidence — none of which was recorded. The reviewer flagged the gap three times on [#10304](https://github.com/monarch-initiative/dismech/pull/10304) and I deferred it twice, because adding records one at a time would have meant three more pushes and three dismissed approvals for bookkeeping that blocks nothing.

`Familial_Visceral_Neuropathy_2` got its record inside [#10305](https://github.com/monarch-initiative/dismech/pull/10305). These are the other three.

## What each record covers

| Entry | PR | Sessions recorded |
|---|---|---|
| `Developmental_and_Epileptic_Encephalopathy_55` | [#10303](https://github.com/monarch-initiative/dismech/pull/10303) | 3 |
| `Immunodeficiency_79` | [#10304](https://github.com/monarch-initiative/dismech/pull/10304) | 4 |
| `ORAI1_Deficiency` | [#10306](https://github.com/monarch-initiative/dismech/pull/10306) | 4 |

The `details` fields record what actually happened rather than a summary of the diff, including the mistakes and their corrections — those are the part worth being able to find later:

- **DEE-55** — `prevalence_class` contradicted its own `rate_per_100000`; the Ketogenic Diet snippet quoted a sentence that never mentions the ketogenic diet; the deep-research provenance note claimed "none off topic" where the frontmatter records `on_topic: 7` of 12 with no `off_topic` key at all.
- **IMD79** — the entry was first curated from the abstract when the cache held full open-access text; `temporality: CHRONIC` was asserted over a quote saying only "recurrent fever, diarrhea"; and four passages called HPV and *T. whipplei* the only two susceptibilities when the paper says "at least" those two and the node carries six downstream edges.
- **ORAI1** — the `Autoimmunity` description called the phenotype "characteristic of the CRAC channelopathies", the exact sentence its own two `REFUTE` items argue against; the fix for that then described the 1-of-7 and 1-of-6 counts as two independent cohorts when they are the same patient (the 2009 report is a census, the 2015 review's denominator subsumes it).

## Validation

`just validate-history-all` — 8,391 records, no issues. `git status` shows exactly the three new files and nothing else.

---


aleixpuigb:
Took both suggestions in `6951433f`, which drops the approval. Normally I'd leave non-blocking items on an approved PR, but the second one is a factual imprecision in a provenance record — the exact failure mode these records exist to prevent, and the one you're right to say will be trusted later.

**The ORAI1 round attribution was wrong, and I checked the commits rather than my memory of them.** You're right that the Table 2 argument pre-dated `e91bb68`. The full picture:

- `3ddcaaa` introduced the IUIS evidence item entirely — reference, snippet, and the explanation placing the row in Table 2 section 8. That is where the assignment got sourced.
- `e91bb68` separately carried the same-patient correction into the top-level `description` (which still said "two reviews put it at one in six and one in seven" — wrong twice, since PMID:20004786 is a primary case series and is itself the source of the one-in-six), and re-cut the snippet from `"Normal Normal Autoimmunity; EDA; non-progressive myopathy"` to `"ORAI-1 deficiency ORAI1 AR 610277 Normal, defective TCR mediated activation"`.

So the record had two problems, not one: it attributed the sourcing to the wrong change, and it collapsed two sessions into one, losing the top-level-description fix entirely. Now split into rounds 4 and 5 with the commits named, and the snippet swap described as what it was — the identification moving out of the explanation and into the quote.

That also fixes the count in the opening sentence: five sessions, not four.

**`sections` widened** to include `notes` on all three and `description` on IMD79 and ORAI1. You're right that this is where the epistemic work landed — the DR provenance paragraph on DEE-55, the "at least" hedging on IMD79, both on ORAI1 — and `sections` being free-text in the schema means there's no reason not to say so.

I did not touch anything else. The records remain descriptions of sessions that already happened; this corrects what they say about those sessions, before merge, rather than after.

Re-validated on `6951433f`: `just validate-history-all` — 8,391 records, no issues; `pytest -k history` — 33 passed, 1 skipped; `git status` clean apart from the three modified files.

---

aleixpuigb:
Flagging this once for a maintainer, in case the queue needs a look — no action needed from me that I can see.

This PR was added to the merge queue at 07:01Z and is still open 6h20m later, with nothing having touched it since (`updated_at` is unchanged at `07:01:41Z`). No dequeue notice arrived. In that window `main` advanced from `05058dc8` to `685c5337` — several other PRs merged past it.

I re-checked rather than assuming, most recently against `685c5337`:

- **Approved**, and CI green on head `6951433f` — `test (3.13)` and `claude-review` both success.
- **No conflict.** `git merge-tree` against current `main` reports zero conflict markers and zero files changed on both sides. The diff is still exactly the three added `history/` files, so there is nothing here that could collide with a curation PR.
- The REST API reports `mergeable_state: unknown`, which is why I verified with git rather than trusting it.

So every shepherd predicate still holds and I have no way to tell from outside whether it is queued behind others or fell out silently — dequeue notices are best-effort. I'm not going to merge it myself, and I'm not posting this again on later checks.

Worth noting the same thing happened to [#10305](https://github.com/monarch-initiative/dismech/pull/10305) in this batch: eligible, watched several PRs merge past it, then merged about 2h45m after being enqueued without anything changing. So this may simply resolve on its own.

---

PR #10946, opened 2026-09-04
Curate Aneurysmal Bone Cyst (MONDO:0018815)

justaddcoffee:
New entry for **aneurysmal bone cyst** (`MONDO:0018815`). Verified gap against fresh `origin/main`: no existing file, no MONDO collision, no curation issue.

## The mechanism is a promoter swap, not a fusion protein

This is the entry's central claim and it's easy to get wrong. The seminal paper uses the phrase *"fusion gene"*, and the events are named in the familiar `PARTNER-USP6` style — which reads like a fusion oncoprotein. **What is fused is DNA.** The entire USP6 coding sequence is retained and only the upstream promoter changes:

> "a recurring ABC chromosomal translocation t(16;17)(q22;p13) creates a fusion gene in which the osteoblast cadherin 11 gene (CDH11) promoter region on 16q22 is juxtaposed to **the entire** ubiquitin-specific protease USP6 (Tre2) **coding sequence**"

The strongest evidence is the partner list itself — thirteen partners (CDH11, CNBP, COL1A1, CTNNB1, EIF1, FOSL2, OMD, PAFAH1B1, RUNX2, SEC31A, SPARC, STAT3, THRAP3) sharing no functional family, only strong expression in bone and connective tissue. Exactly what you'd expect if the promoter is the operative element and the partner gene is incidental.

Consequence for the model: **`CDH11` and `COL1A1` are recorded as `COOPERATING`, not drivers.** They contribute a promoter and no protein.

## Chain

`promoter swap` → `USP6 overexpression` (deubiquitinase activity required, Arf6 binding **not**) → `NF-κB` via RhoA/ROCK → `MMP-9/MMP-10` → `osteolysis + vascular spaces` → `local destruction` → pain, cord compression, proptosis.

Verified by parsing the file: **one root, zero orphan phenotypes, zero dangling targets**, all four phenotypes connected. That check was added to my workflow after the review on #10532 caught a stranded node visual inspection had missed.

## Three things recorded rather than smoothed over

**1. The mechanism chain rests on one paper.** Everything from USP6 through NF-κB to the MMPs comes from a single 2010 cell-culture and xenograft study, graded `IN_VITRO` throughout. It's a strong study — the xenograft reproduces multiple features of the lesion, and domain mapping cleanly separates the required deubiquitinase activity from the dispensable Arf6-binding function — but it is one paper, and I did not upgrade the chain to human evidence on the strength of clinical papers that *describe* the lesion without testing the pathway.

**2. A diagnostic discrepancy I did not reconcile.** The FISH validation study reports 53% sensitivity while citing a literature rearrangement rate of 70–75%. Both figures are in the entry. The gap is real and its cause isn't established — assay performance, case-mix, or genuine absence in a subset. The practical consequence is what's stated: specificity and PPV are both 100%, so **a positive result is decisive and a negative one is not**.

**3. Denosumab carries a `NO_EVIDENCE` item.** Alongside its supporting evidence, quoting the authors' own conclusion that *"The oncological outcome remains unclear"*. With 18.6% recurrence and **5 of 43 reported patients requiring intensive care for hypercalcaemia**. The drug was borrowed from giant cell tumour on histological similarity rather than shared genetics, and the `target_mechanisms` description notes RANKL blockade doesn't touch the USP6 driver upstream — the likely reason lesions recur when it stops.

## Scope

Secondary ABC is listed as a subtype because it carries the same name and histology, but `notes` states plainly that **nothing in the pathograph applies to it** — it is USP6-negative reactive change inside a different neoplasm and behaves according to the lesion underneath. The diagnosis section also flags the consequential pitfall: telangiectatic osteosarcoma mimics this lesion on imaging, so biopsy precedes treatment.

## Provenance

Curated **directly from PubMed**. Both configured deep-research providers remain unavailable (openscientist.io: CloudFront 403 on POST, 401 on authenticated GET; cyberian: 500 from its local backend) — same signature as the three prior entries. Top-level `references` generated programmatically from cache frontmatter.

## Validation

```
All validations passed!
Snippets checked: 46/46 verified against cached references
Term validation: passed
Weighted Compliance: 100.0%
```

`check-reference-titles`, `check-snippet-length`, `check-snippet-grading`, `check-title-snippets`, `check-folded-hyphens`, `check-enum-values`, `check-duplicate-keys`, `check-entity-refs`, `check-causal-targets` — all clean. History record validates.

Extension points in `notes`: prevalence (no cleanly quotable incidence in the cached sources), structured `histopathology`, `clinical_trials`, bisphosphonates/RFTA/curopsy (named in the management review without quotable outcome data), and the wider USP6-driven family (nodular fasciitis, myositis ossificans) which would suit a mechanism module better than this entry.

### DIFF
--- a/history/disorders/Aneurysmal_Bone_Cyst/2026-09-04T160451Z-claude-code-47e509.yaml
+++ b/history/disorders/Aneurysmal_Bone_Cyst/2026-09-04T160451Z-claude-code-47e509.yaml
@@ -1,0 +1,47 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Aneurysmal_Bone_Cyst
+  path: kb/disorders/Aneurysmal_Bone_Cyst.yaml
+session:
+  id: 2026-09-04T160451Z-claude-code-47e509
+  timestamp: '2026-09-04T16:04:51Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10946
+  urls: []
+events:
+- type: REVIEW
+  outcome: changed
+  sections:
+  - pathophysiology
+  - phenotypes
+  - prevalence
+  - histopathology
+  - treatments
+  - genetic
+  - notes
+  summary: 'Review round 1: RANKL node, prevalence, histopathology, fracture phenotype,
+    evidence regrading, deep-research artifact'
+  details: |
+    Response to review round 1 on PR #10946. All six blocking items, all six suggestions, plus the deep-research artifact.
+
+    The reviewer identified the root cause correctly and it is the third instance of the same failure in this session. Three of the ten cached references are full text, not abstracts: PMID:35941207 is 55 kB, PMID:20418905 is 23 kB and PMID:34754635 is 16 kB. All three were mined only for the abstract at the top. The entry then recorded in notes that no quotable incidence figure existed, when it was present in a file the same commit added. The cache frontmatter carries a content_type field stating full_text_xml or abstract_only outright, and it was never checked. Checking it is now the first step before writing, and doing so here yielded prevalence, histopathology, the pathological fracture rate, the recurrence figures and the sharper USP6 percentages.
+
+    Blocking items. The xenograft evidence was graded IN_VITRO for a mouse experiment and is now MODEL_ORGANISM. It had also been attached to an edge asserting NF-kappa-B to MMP-9 transcription, which a xenograft result does not show; the transcription sentence from the same abstract now carries that edge and the xenograft sits alongside it as separate support. Structured prevalence was added with both incidence 0.14 and prevalence 0.32 per 100,000. The RANKL arm became its own pathophysiology node conforming to osteoporosis_bone_resorption#RANKL-Driven Osteoclastogenesis, and denosumab's target_mechanisms now points at it rather than at the osteolysis node it was reasoning past. Pathologic fracture was added at 15 percent with an edge from the endpoint node.
+
+    Suggestions. The denosumab recurrence item was regraded from NO_EVIDENCE to REFUTE, since the quote does bear on the efficacy claim rather than being silent on it. A histopathology section was added covering the unlined blood-filled spaces, the RANK-positive giant cells against RANKL-positive stromal cells, and the absence of cytologic atypia that separates this lesion from telangiectatic osteosarcoma. Recurrence was quantified at 44.1 percent for curettage, and the sclerotherapy claim was qualified with a REFUTE item recording 90.6 percent persistent disease and 9.4 percent cure against the single-centre 92.3 percent ossification figure, because the two sources measure different endpoints and a curator taking only the first would overstate the treatment.
+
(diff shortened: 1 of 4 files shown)


justaddcoffee:
All six blocking items and all six suggestions, bundled with the deep-research artifact and the history record into a single push as requested.

## You found the root cause, and it's the third instance of it

Three of the ten cached references are **full text, not abstracts**:

| Reference | Size | content_type |
|---|---|---|
| PMID:35941207 | 55 KB | `full_text_xml` |
| PMID:20418905 | 23 KB | `full_text_html` |
| PMID:34754635 | 16 KB | `full_text_xml` |

I mined all three for the abstract at the top and stopped. The entry then asserted in `notes` that no quotable incidence figure existed — in a file the same commit added. The frontmatter carries a `content_type` field that says `full_text_xml` outright, and I never read it. Checking it is now the first step before I write anything, and doing so here produced the prevalence figures, the histopathology section, the fracture rate, the recurrence numbers, and the sharper USP6 percentages.

## Blocking items

**1 + 2** — The xenograft evidence was graded `IN_VITRO` for a mouse experiment (now `MODEL_ORGANISM`) *and* attached to an edge claiming NF-κB drives MMP-9 transcription, which a xenograft doesn't show. The transcription sentence from the same abstract now carries that edge, with the xenograft alongside as separate support.

**3** — `prevalence:` added: incidence 0.14/100,000, prevalence 0.32/100,000, ~1% of bone tumours.

**4** — RANKL is now its own node with `conforms_to: osteoporosis_bone_resorption#RANKL-Driven Osteoclastogenesis`, and denosumab targets it directly instead of reasoning about RANKL while attaching to the osteolysis node. The graph now branches properly:

```
USP6 overexpression ─┬─→ NF-κB → MMP-9/10 ─┐
                     └─→ RANKL → osteoclasts ┴─→ osteolysis → lesion
```

Worth noting what the conformance implies: this axis **isn't specific to this tumour**, which is exactly why denosumab works in giant cell tumour and chondroblastoma, and why blocking it leaves the USP6 driver untouched.

**5** — Pathologic fracture at 15%, with an edge from the endpoint node.

**6** — Deep research ran. See below.

## Suggestions, all taken

`NO_EVIDENCE` → `REFUTE` on the denosumab recurrence item; you're right that the quote bears on the claim. Histopathology section added. Recurrence quantified at 44.1% for curettage — and I added a `REFUTE` item on sclerotherapy recording **90.6% persistent disease and 9.4% cure** against the single-centre 92.3% ossification figure I'd used. The two sources measure different endpoints, and a curator taking only the first would overstate the treatment. USP6 figures sharpened to 65–70% with CDH11-USP6 in 30%.

## Deep research

Ran, 2 iterations, 757 seconds, attached as `research/Aneurysmal_Bone_Cyst-deep-research-openscientist.md`. Its substantive contribution was one finding I'd missed: **USP6/CDH11 rearrangements are restricted to the spindle cells** and absent from the giant cells, inflammatory cells, endothelium and osteoblasts. That identifies the myofibroblastic spindle cell as the neoplastic cell and reframes the giant cells and vascular spaces as recruited rather than transformed — which is now the stated reason the downstream nodes are framed as recruitment. I verified it directly against PMID:15509545 before using it rather than taking the report's word.

**Correction on the provider outage.** My earlier entries said both providers were down. That was wrong, and it was my misconfiguration. `openscientist.io` 301-redirects to `www.openscientist.io`, and `curl -L` strips the `Authorization` header across a host change — producing the 401 I read as a revoked key. The apex CloudFront distribution separately rejects non-cacheable methods, producing the 403 I read as a service outage. Setting `OPENSCIENTIST_URL=https://www.openscientist.io` fixes both. Two error codes, one plausible story, and I stopped instead of testing it — the same pattern you flagged on #10532.

## Validation

```
All validations passed!
Snippets checked: 62/62 verified against cached references
Term validation: passed
Weighted Compliance: 100.0%
```

All gates clean. Topology verified by parsing the file: one root, no orphan phenotypes, no dangling targets.

One housekeeping note: running the repo-wide gates rewrote frontmatter in ~50 unrelated `references_cache` files (quote normalisation plus a new `full_text_attempted` key). I reverted those rather than sweep them in, so this diff stays scoped to four files.

PR #8408, opened 2026-08-12
Backfill environmental pathograph links: the diabetes umbrella (#8085)

caufieldjh:
Tranche 39 of the #8085 backfill. Nine `influences_mechanisms` links, **all into a single node** — because that is what this entry is. Insertions only, 1 entry + 1 history record.

## Why every link shares a target

`Diabetes_Mellitus` is the **retained umbrella** entry (per CLAUDE.md: diabetes is modelled as Grouping + `diabetic_vascular_complications` module + per-type entries, with the umbrella Disease retained to carry `MONDO:0005015`). It has exactly one pathophysiology node — `Chronic Hyperglycemia (shared diabetic complication cascade)` — and the subtype-specific routes live in `Type_I_Diabetes`, `Type_2_Diabetes_Mellitus` and the module.

So all nine exposures target that node, and **the grades rather than the targets are what separate them**. Every link description says this explicitly, so the uniformity does not read as indifference.

## The spread

**`TRIGGERS`** — the two drug exposures, which produce the disease rather than raising its odds:
- **Diabetogenic pharmacotherapy** is the clearest causal exposure in the entry and the only one with a **reversal design**: the drug causes it and stopping the drug often reverses it — bounded by the authors' own note that persistence depends on duration, agent and BMI.
- **Immune checkpoint inhibitors** earn a separate edge because the *mechanism differs*: releasing the brakes on self-reactive T cells produces insulin **deficiency**, sometimes presenting in ketoacidosis, not the insulin resistance most diabetogenic drugs cause. `PARTIAL` — a single case report.

**`PREDISPOSES`** — the risk exposures, graded against each other by what their sentences contain:
- **High-calorie diet** is `INDIRECT_KNOWN_INTERMEDIATES` because its sentence *names* the intervening step (hyperinsulinaemia driving insulin resistance).
- **Sedentary lifestyle** is graded below it precisely because its sentence names no intermediate — it reports co-rising prevalence of metabolic syndrome.
- **Obesogenic environment** is the weakest and the description says so: its only evidence is a **screening recommendation**, which implies risk without stating a mechanism or a magnitude.

⚠️ **Three exposures share one sentence.** The islet-autoimmunity trigger, viral infection and early dietary exposure all rest on a single sentence naming "certain dietary factors and viruses" together — so none can outrank the others on evidence held here, and all three are `PARTIAL`.

**`MODULATES`** — two that do not assert causation:
[...]

PR #8090, opened 2026-08-06
Backfill environmental pathograph links: exposure-caused disorders (#8085)

caufieldjh:
First tranche of #8085, the backfill of `influences_mechanisms` onto existing `environmental:` entries. Structured as **four reviewable commits**, one per exposure class.

**33 links across 11 disorders.** Every environmental entry in these files is now wired into its pathograph instead of rendering in a card and then vanishing from the graph.

## Why these 11 first

I surveyed all 828 environmental entries across 348 disorders looking for a deterministic rule — e.g. an entry whose `exposure_term` matches a pathophysiology node's `triggers:` ECTO term. **There are zero such matches** (Arsenic, the #8084 worked example, was the only one and it is already linked). So there is no mechanical shortcut here; each link is a judgement call about which node the exposure acts on and in which direction.

That makes ordering matter. These 11 are the cases where the exposure is not a risk factor for the disease but *is* its initiating step, and where the entry already contained a deposition / absorption / uptake node sitting disconnected from the exposure that produces it. They also already carried verified evidence, so no new citations were introduced.

| Commit | Disorders | Links |
|---|---|---|
| Inhaled fibers and particles | Asbestosis, Silicosis, Chronic beryllium disease, Malignant mesothelioma | 4 |
| Inhaled gas / ingested alcohol | Carbon monoxide, Methanol | 8 |
| Thallium and organophosphates | Thallium, Organophosphate | 9 |
| Noise, mercury, lead | Noise-induced hearing loss, Mercury, Lead | 12 |

## Judgement calls worth reviewing

These are where the enum earns its keep, and where I'd most want a second opinion:

- **Methylene chloride is not a CO inhalation.** The other four CO sources are `DIRECT` — they differ in where the CO comes from, not how it enters, so they share the mechanistic edge evidence. Methylene chloride is `INDIRECT_KNOWN_INTERMEDIATES`: no CO is inhaled at all, the solvent is metabolized hepatically to CO. That extra step is precisely why the carboxyhemoglobin rise there is delayed and prolonged.
- **Coal combustion is `PREDISPOSES`, not `TRIGGERS`.** The four other thallium routes are documented ways absorbable thallium reaches a person. The cited source frames coal/cement as a *growing environmental-contamination concern*, not a poisoning route — it raises the opportunity for exposure without being one. Calling it TRIGGERS would assert cases the source never reports.
[...]

PR #10960, opened 2026-09-04
Backfill environmental pathograph links: exposures that are their own substrate (#8085)

caufieldjh:
Tranche 43 of the #8085 backfill. **Seven** `influences_mechanisms` links across seven entries, and **one exposure deliberately left unlinked**. Insertions only: **422 / 0** (7 entries + 7 history records).

Picked from the drug-or-toxin and infection buckets of the survey posted on the issue today, on the principle that the plainest edge is the one where **the exposure is the substrate of the node it points at**.

## The links

| Entry | Exposure | Target node | Effect | Link type |
|---|---|---|---|---|
| Aflatoxin_Related_HCC | Aflatoxin B1 Exposure | Aflatoxin B1 Metabolic Activation | `TRIGGERS` | `DIRECT` |
| Arsenic_Related_Cancers | Arsenic Exposure | Arsenic Uptake and Metabolism | `TRIGGERS` | `DIRECT` |
| Cannabis_Hyperemesis_Syndrome | Chronic Cannabis Use | Endocannabinoid System Dysregulation | `TRIGGERS` | `DIRECT` |
| Thromboangiitis_Obliterans | Tobacco smoking exposure | Tobacco-associated endarteritis obliterans | `TRIGGERS` | `DIRECT` |
| Transitional_Cell_Carcinoma | Tobacco Smoke and Aromatic Amine Exposure | Carcinogen-Induced Urothelial Mutagenesis | `TRIGGERS` | **`INDIRECT_KNOWN_INTERMEDIATES`** |
| Stevens-Johnson_Syndrome | High-Risk Drug Exposure | Drug-HLA Direct Interaction | `TRIGGERS` | `DIRECT` |
| Marchiafava_Bignami_Disease | Chronic alcohol use and malnutrition | Alcohol-related thiamine depletion | `TRIGGERS` | `DIRECT` |

**Why six are `DIRECT`.** Not because the route is short, but because the absorption or delivery step is **inside the target node's own description** and therefore not an intermediate the edge steps over. Aflatoxin's node opens "AFB1 is absorbed from the gut and transported to the liver where it undergoes cytochrome P450-mediated oxidation"; the arsenic node opens with absorption from gut or lungs. In each case the ingested compound is what the node acts on. The Thromboangiitis and Stevens-Johnson nodes are `DIRECT` for a different reason: the node is *named for* the exposure, or the drug is a constituent of the step (drug binding HLA), so there is nothing between them to model.

**Why one is not.** `Transitional_Cell_Carcinoma`: the route from smoking to urothelial mutagenesis runs through urinary excretion, contact with the urothelium, and adduct formation. That route is mapped, and the node's description gives it. The quoted sentence gives none of it. Link type records **what the field knows**; `directness` on the item records **what the sentence says**. They disagree here on purpose, per the convention settled on #8408.

## The cleanest edge in this tranche

`Cannabis_Hyperemesis_Syndrome`. The target node's own item states exposure and node in one sentence:

> "prolonged high doses of the main psychotropic compound in cannabis, Δ9-tetrahydrocannabinol (THC), result in changes to the endocannabinoid system by acting on the cannabinoid 1 (CB1) receptor"

That is `directness: DIRECT` on the item as well as `DIRECT` on the link, which is rare in this series.

## 🚫 One exposure deliberately NOT linked, and recorded where a survey can see it

`Cannabis_Hyperemesis_Syndrome` / **High-Potency THC Products**. Its cited sentence reports that CHS "has become more prevalent with increasing cannabis potency and use, as enabled by various states having legalized the recreational use of cannabis". That is a **population trend**, not a statement that potency acts on the endocannabinoid node, and a second edge would assert a dose-response this entry does not evidence.

⚠️ **Convention change, and the reason for it.** Earlier tranches recorded these decisions in the PR's history record only. Today's survey of `main` found **1061 environmental entries, 704 linked, 357 unlinked, and exactly 2 whose entry `notes:` say the no-link was deliberate** — so from a reading of the knowledge base itself, nearly every settled decision looks like unfinished work. This tranche writes the reason into the entry's `notes:` **as well as** the history record. Recommend all remaining tranches do the same.

## Evidence discipline

Every item was **copied verbatim from an item already present in the file** by a generator that parses the YAML and re-emits the block, so `reference`, `reference_title`, `snippet`, `supports` and `evidence_source` cannot drift. Only the `explanation` was rewritten for the narrower claim a link makes, and `directness` added. This is the #8090 lesson (hand-writing citations produced invented titles and wrong PMIDs) applied by construction rather than by audit.

Two links carry a sentence found on the **target node** rather than on the exposure entry, per the #8092 lesson:
- `Marchiafava_Bignami_Disease` — the exposure entry cites PMID:37559750 for the clinical context ("predominantly afflicts individuals with chronic alcohol abuse"); the *node* cites the same paper for the causal step ("alcohol consumption, which leads to thiamine depletion"). The link carries the second.
- `Stevens-Johnson_Syndrome` — the p-i mechanism sentence (PMID:22322005, `IN_VITRO`) sits on the node. Its `evidence_source` was carried across unchanged, so the cell-based origin of the mechanistic half stays visible next to the epidemiologic half (relative risk 172 for sulfonamides, `HUMAN_CLINICAL`, `directness: INDIRECT`).

One item's `evidence_source` is **absent** in its original location (Arsenic, PMID:26411935) and was left absent rather than guessed.

`Transitional_Cell_Carcinoma`'s `exposure_term` is still unbound. That belongs to the environmental-term audit (#8430) and was not widened into here.

## Validation

```
linkml-validate               passed (all 7)
just validate-terms           passed (all 7)
count-verified-snippets       405/430 verified, 25 skipped by prefix, 0 failing
check-causal-targets          OK      check-entity-refs         OK
check-duplicate-keys          OK      check-enum-values         OK
check-snippet-boundaries      OK      check-title-snippets      OK
check-snippet-length          OK      check-folded-hyphens      OK
check-empty-snippets          OK      check-reference-titles    OK
check-environmental-evidence  OK      check-snippet-grading     OK
just validate-history-all     No issues found (8396 records)
pytest tests/test_data.py -k "environmental or conforms_to or foreign or entity or target"
                              29432 passed
git diff --cached --numstat   422 insertions / 0 deletions
```

Reference-cache churn from running `just validate` (17 files upgraded to full text) was reverted, not committed: no snippet here needs full text, and all 405 verify against the committed abstract-only cache.

Branched from current `main` (`717fe882fb`).

https://claude.ai/code/session_01FjG2TgWtZupYyPTFN2T8Z4

### DIFF
--- a/history/disorders/Aflatoxin_Related_HCC/2026-09-04T163209Z-claude-code-f0c893.yaml
+++ b/history/disorders/Aflatoxin_Related_HCC/2026-09-04T163209Z-claude-code-f0c893.yaml
@@ -1,0 +1,30 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Aflatoxin_Related_HCC
+  path: kb/disorders/Aflatoxin_Related_HCC.yaml
+session:
+  id: 2026-09-04T163209Z-claude-code-f0c893
+  timestamp: '2026-09-04T16:32:09Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-fable-5-1
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/8085
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - environmental
+  summary: Link dietary aflatoxin exposure to the AFB1 bioactivation node
+  details: |
+    Backfill for issue #8085, tranche 43. One influences_mechanisms link: Aflatoxin B1 Exposure TRIGGERS 'Aflatoxin B1 Metabolic Activation', causal_link_type DIRECT. Direct because the target node's own description begins with absorption from the gut and transport to the liver, so the edge steps over nothing; the ingested toxin is the substrate of the step.
+
+    Both evidence items were copied verbatim from items already in this file rather than hand-written: PMID:40863326 from the exposure entry (aflatoxin as a major etiological factor, graded directness INDIRECT because it reaches the node through the disease) and PMID:8261428 from the target node (CYP1A2 converting AFB1 to the 8,9-epoxide, INDIRECT because as a microsome experiment it says nothing about dietary exposure). Reference titles, snippets, supports and evidence_source were copied unchanged; only the explanation was rewritten for the narrower claim the link makes, and directness added.
+
+    Validation: linkml-validate passed; just validate-terms passed; count-verified-snippets over the seven files of this tranche reported 405/430 verified, 25 skipped by prefix, none failing. check-causal-targets, check-entity-refs, check-duplicate-keys, check-enum-values, check-snippet-boundaries, check-title-snippets, check-snippet-length, check-folded-hyphens, check-empty-snippets, check-reference-titles, check-environmental-evidence and check-snippet-grading all reported OK. Insertions only, no deletions.
--- a/history/disorders/Arsenic_Related_Cancers/2026-09-04T163209Z-claude-code-27bcfa.yaml
+++ b/history/disorders/Arsenic_Related_Cancers/2026-09-04T163209Z-claude-code-27bcfa.yaml
@@ -1,0 +1,30 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Arsenic_Related_Cancers
+  path: kb/disorders/Arsenic_Related_Cancers.yaml
+session:
+  id: 2026-09-04T163209Z-claude-code-27bcfa
+  timestamp: '2026-09-04T16:32:09Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-fable-5-1
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/8085
+  prs: []
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - environmental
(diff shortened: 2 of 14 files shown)


caufieldjh:
Taking the `review_notes` suggestion. I verified it independently before acting: `review_notes` appears **zero** times in `disorder.html.j2` while `env.notes` renders at line 6209, and `scripts/check_environmental_evidence.py` states the split in its own docstring, *"`notes` is disease content and is not consulted"*. The waiver sentinel is a prefix match on `"left deliberately uncited."`, a different sentence, so a disposition in `review_notes` cannot trip it — I confirmed the waiver count stays at 4.

**I could not apply it here: this PR is already in the merge queue**, and pushing is rejected with `GH006`. I am not dequeuing an approved PR to make a non-blocking change, so the move lands as a follow-up PR once this merges. The change is one line (`notes:` → `review_notes:`) with byte-identical text, verified by parsing both revisions.

The convention is applied at scale in **#10980**, which files 34 no-link dispositions across 27 entries in `review_notes` rather than `notes` — its body credits this suggestion and cites the same evidence. Its "follow-up on that branch" line is now inaccurate for the reason above; corrected in a comment there.

On the wider point you raised: agreed that this is worth settling by slot rather than by accretion. A survey keying on `review_notes` reads a field designed for the purpose, where grepping prose out of `notes` reads disease content and would also collide with genuine curator prose. Today's survey found exactly 2 of 357 unlinked entries recording their decision anywhere in the entry, so the slot choice matters more than the backlog size suggests.

PR #10961, opened 2026-09-04
Curate Christianson Syndrome (MONDO:0010278)

justaddcoffee:
New entry for **Christianson syndrome** (`MONDO:0010278`). Verified gap: no file, no MONDO collision, no curation issue. Angelman syndrome is curated separately and is a distinct disease despite the "X-linked Angelman-like syndrome" synonym.

## The pH direction, stated explicitly because intuition runs backwards

NHE6 **exports** protons from the endosomal lumen, so losing it makes the compartment *more* acidic, not less. The point is sharpened by `p.Gly218Arg`, a probable gain-of-function allele that **alkalinises** the endosome and still causes the core syndrome, while sparing the motor deterioration usually seen after the first decade. Both directions are pathogenic, so the node is framed as loss of pH control rather than as acidification.

## The pathograph branches on purpose

```
SLC9A6 loss → over-acidification → endosomal/lysosomal failure ─┬─→ BDNF-TrkB ↓ → arborisation ↓ → microcephaly, ID,
                                                                │                                  absent speech, seizures
                                                                └─→ Purkinje cell loss → ataxia, regression, atrophy
```

Two arms with different timing, substrates and consequences. Collapsing them would lose the clinically central fact that a child can be developmentally static for a decade and then begin losing skills. Roughly half of patients have documented regression, and a majority of adults lost gross and fine motor skills over a single year of follow-up.

## Deep research earned its place

Ran at **5 iterations** (1059 s, 33 citations, 9 findings sections), attached as `research/Christianson_Syndrome-deep-research-openscientist.md`.

Its concrete contribution: **it supplied a mechanism for a phenotype I had graded `UNKNOWN`.** High pain tolerance is a core feature at 91%, and no source I'd cached explained it — so the edge went into the graph with `causal_link_type: UNKNOWN` rather than a guess. The report pointed at work showing NHE6 expression in nociceptors, reduced cell-surface TRPV1, and blunted nocifensive responses to thermal, mechanical and chemical stimuli in Nhe6-null mice. Verified directly against PMID:32569089 before use; the edge is now `INDIRECT_KNOWN_INTERMEDIATES`.

It also surfaced a **TrkB agonist rescue** (PMID:39341363): 7,8-dihydroxyflavone restores spine density and both functional and structural LTP in Nhe6-null hippocampal neurons. That regrades the BDNF-TrkB node from `HYPOTHETICAL` to `PROVISIONAL` — rescue beats correlation, and it supplies the forebrain evidence the original organ-of-Corti hearing study could not. And a pH-**independent** CDK5/p35 scaffolding role (PMID:42051037), which matters because it means not every consequence of losing NHE6 must run through endosomal pH.

**On iteration count**, since it came up: 2 iterations on the previous entry produced 3 findings sections and no epidemiology at all. 5 produced 9 findings and 33 citations for ~40% more runtime. Sublinear cost, materially better coverage.

## Process

Four of the twelve cached references are **full text, not abstracts** — and `content_type` was checked *before* writing this time, not after a reviewer pointed it out. The natural-history figures throughout come from reading those full texts. One artefact worth flagging for others: PDF-derived full text carries interleaved line numbers that split sentences, so two snippets failed verification and were shortened to contiguous fragments.

Topology verified by parsing the file rather than reading it. **The first pass found eight orphan phenotypes**, all connected before any commit. Final state: one root, no orphans, no dangling targets.

On prevalence: no quotable figure exists in the cached sources, and the deep-research report independently concluded the same, describing precise prevalence and incidence as undefined. That's recorded in `notes` as a property of the literature rather than left as an unexplained gap.

## Validation

```
All validations passed!
Snippets checked: 59/59 verified against cached references
Term validation: passed
Weighted Compliance: 100.0%
```

All gates clean. Three animal models included, each with stated limitations — notably the NHE6-null rat, built specifically because mice don't deposit endogenous tau or amyloid-β, so the species choice *is* the experimental design.


justaddcoffee:
## Review round 1 addressed — all 5 findings and the suggestion

Pushed as 6e45e12b24. `just validate` now reports **76/76 snippets verified** (was 59/59), term validation passes, one root and no orphan phenotypes, weighted compliance 100.0%.

### Finding 1 — bands that contradicted their own numbers

Both corrected, and each `explanation` now records what the band was before rather than presenting the new one as if it had always been there.

| Phenotype | Was | Now | Quoted number |
|---|---|---|---|
| Pain Insensitivity | FREQUENT | VERY_FREQUENT | 91% |
| Developmental Regression | OCCASIONAL | FREQUENT | 50% |

### Finding 2 — one sentence read six ways

This was the sharpest finding in the review and it is right. The core-criteria sentence was read as OBLIGATE for non-verbal status, intellectual disability and epilepsy, and as FREQUENT for microcephaly, ataxia and hyperkinesia. Three items from the identical list, two readings.

All six are now VERY_FREQUENT from the stated >85%. Each carries that sentence as its band evidence with the same explanation: universal presence between ages 6 and 16 is an age-window statement, not a lifespan-wide obligate finding, which is why none of the six is OBLIGATE.

Ataxia was wrong in a second way. Its band was evidenced by the 30-60% rate of cerebellar atrophy on imaging, which is a different phenotype's number. That figure stays on Cerebellar Atrophy where it already sat, and the ataxia evidence item now states explicitly that it supports the cerebellar basis and does not carry the band.

### Findings 3 and 4 — GeneReviews

Tagged. And you found something worth naming: the file is cached `abstract_only`, which is an honest label, but a GeneReviews abstract is *sectioned*. An earlier draft mined CLINICAL CHARACTERISTICS and stopped, leaving GENETIC COUNSELING and MANAGEMENT unread in a file already committed to this branch. Checking `content_type` does not catch this, because nothing is mislabelled. This is the second entry in the series where it has happened.

Both sections are now mined. The 50% transmission risk and its split by sex of offspring sit under `inheritance`. The surveillance recommendation is a treatment, targeting the Developmental Regression node with `treatment_effect: MODULATES` and a description stating that surveillance does not prevent regression, only changes when it is recognised. You were right that it sits squarely on this entry's own thesis about the biphasic course.

### Finding 5 — three phenotypes

All three added from text already on this branch, all wired into the pathograph.

- **Eye movement abnormality**, `HP:0000496`, FREQUENT. One correction to the review: the tabulated counts in the cohort are 69 and 70 percent across the two groups, not ~79%, so FREQUENT is the band the numbers support. Both the GeneReviews naming sentence and the table row are cited.
- **Hypotonia**, `HP:0001252`, FREQUENT, from the GeneReviews sentence the entry was already quoting elsewhere.
- **Upper motor neuron dysfunction**, `HP:0002493`, FREQUENT, from >50% over age 10.

Eye movement abnormality hangs off the arborisation node as INDIRECT_UNKNOWN_INTERMEDIATES, because no cached source localises the eye movements to a circuit and assigning them to the cerebellar arm on plausibility would be a mechanism claim without a source.

While in there I fixed a pre-existing orphan you did not flag: failure to thrive was in the phenotype list but not targeted by any edge. It now links as INDIRECT_KNOWN_INTERMEDIATES with hypotonia and dysphagia as the named intermediates, from the sentence that states all three.

### Suggestion

Corrected: twelve references, five of them full text. The old sentence said four of ten.

### One error of mine, caught by validation

The surveillance treatment initially bound `NCIT:C25243`, which I had assumed was a monitoring term. It resolves to "Vertical". Rebound to `NCIT:C124351` Clinical Evaluation, verified as a descendant of `NCIT:C25218` before use.

### Note on two repository gates

`check-snippet-grading` and `check-title-snippets` fail on this branch, but report zero findings in this file. Their findings are in unrelated entries and are pre-existing on `main`.

PR #10972, opened 2026-09-04
curate: Add Autosomal Recessive Nonsyndromic Hearing Loss 48 (DFNB48, CIB2)

### LINKED ISSUE #10955: Curate autosomal recessive nonsyndromic hearing loss 48 (MONDO:0012273)

Curate a dismech entry for **autosomal recessive nonsyndromic hearing loss 48** ([MONDO:0012273](https://monarchinitiative.org/MONDO:0012273)), DFNB48 — biallelic *CIB2* variants causing prelingual profound sensorineural hearing loss. *CIB2* is a calcium- and integrin-binding protein of the hair-cell stereocilia that associates with the TMC1/TMC2 mechanotransduction machinery, so this is a mechanotransduction-failure entry rather than a synaptic or endolymph one.

Clean leaf: no MONDO descendants, one causal gene (`hgnc:24579` CIB2). The KB already carries many `Autosomal_Recessive_Nonsyndromic_Hearing_Loss_*` entries, so this follows an established pattern.

Worth resolving during curation: *CIB2* was originally also reported for Usher syndrome type 1J (USH1J), and that association has since been disputed. Whether the entry says anything about USH1J, and how it grades that evidence, is a real curation call rather than a detail.

No existing coverage: *CIB2* appears in zero files under `kb/disorders` on `main`, and no open PR curates it.

Stub: `stubs/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48.yaml`

aleixpuigb:
Adds a knowledge base entry for **DFNB48** ([MONDO:0012273](https://monarchinitiative.org/MONDO:0012273)), prelingual, bilateral, generally severe-to-profound sensorineural hearing loss caused by biallelic variants in *CIB2*.

Hair cells detect sound with stereocilia — tiny bundles arranged in a staircase. Force pulls on filaments called tip links, which open a channel at the tip of the shorter rows. CIB2 is not the channel and not the tip link; it is a helper protein that sits on the channel and is needed for it to work.

## The result the entry is built around

Cib2 mutant mice are deaf and show **no mechanotransduction at all — despite the tip links being intact**. That negative is what makes the entry interesting. If the tip links were broken you would suspect a force-delivery problem. They are not, so the fault is in the channel complex itself, and that is what makes CIB2 an auxiliary subunit rather than a structural part of the bundle.

There is a second, separate arm: without CIB2 the shorter rows of stereocilia overgrow, degrading the staircase. The authors state both roles, so the pathograph carries both rather than collapsing them.

## The gene was reassigned, and the entry says so rather than picking a side quietly

*CIB2* was originally reported to cause **both** DFNB48 and Usher syndrome type 1J — deafness plus blindness and balance failure. A later multi-ethnic cohort found non-syndromic deafness for every pathogenic genotype it looked at, loss-of-function alleles included, and set out explicitly to disqualify *CIB2* as an Usher gene.

This is not a bookkeeping detail. Whether a deaf child is told to expect progressive blindness turns on it. So the entry curates the disease as non-syndromic but records **both** claims as a `CONTROVERSY` discussion with the evidence that cuts each way, including the founding paper's own dual assignment.

What tipped it for me is that there is a positive mechanism for the absence, not just a failure to observe:

- Vestibular hair cells co-express CIB2 and its paralogue CIB3; cochlear hair cells express CIB2 alone.
- CIB3 rescues transduction in Cib2-null outer hair cells. CIB1 does not — so it is specific to the closest paralogue, not a side effect of overexpressing any CIB protein.
- Transduction currents are abolished in auditory hair cells while **unchanged** in vestibular hair cells of the same animals.

So the vestibule has a backup and the cochlea does not, and the phenotype is restricted for a reason.

I have flagged the limitation in the discussion rather than leaving it implicit: the reassignment rests on the absence of a finding in a modest cohort, and retinitis pigmentosa in Usher type 1 appears in the first decade, so ascertainment age is a live constraint on the negative claim.

## A node with no downstream edge, on purpose

The CIB3 redundancy node deliberately has no `downstream` target. It explains an *absence* — vestibular dysfunction — rather than causing a phenotype, and the pathograph has no way to represent a protective mechanism inside a disease entry. I recorded that reasoning in the node's `notes` so it does not read as a dangling node somebody should wire up.

## Gene therapy: reported with the authors' own limits

AAV-CIB2 and AAV-CIB3 rescue hearing in the mouse model, and the snippet I quoted keeps the caveats the authors put on it — the recovery is **partial**, restricted to apical-to-middle cochlear turns, and depends on treating between P0 and P4. There is no human trial for DFNB48; the clinical precedent for cochlear gene therapy is OTOF/DFNB9, a different gene, and the entry says so.

## What I got wrong and fixed before pushing

My first draft used publication **titles** as evidence snippets in seven places. That is the thing `check-title-snippets` exists to stop — a title is not a finding — and it also failed validation because the titles were not substrings of the cached bodies. All seven were replaced with quoted results from the cached abstracts. The replacements are better evidence, not just passing evidence: the prevalence claim went from a title to "The five most common HL genes in the Pakistani population are SLC26A4, MYO7A, GJB2, CIB2 and HGF", and the severity claim went from a title to the observation that the same recurrent allele gives "remarkably less severe" deafness in one family than in all others.

## GeneReviews

There is no *CIB2*- or DFNB48-specific chapter. The Genetic Hearing Loss Overview ([PMID:20301607](https://pubmed.ncbi.nlm.nih.gov/20301607/)) is cached and listed under `references` with the `GeneReviews` tag, but its cached record is book front matter with no quotable Clinical Characteristics section, so no evidence item is attached to it.

## Deep research

openscientist, 1102s. 16/16 references resolved, 0 confabulations, 13/16 assessed on topic and 0 off topic. Term validation set `needs_review`, but every flag is a benign abbreviation — the report calls `HP:0008619` "Bilateral SNHL" where HPO calls it "Bilateral sensorineural hearing impairment" — and none of them touch a term bound in this entry.

One thing worth flagging for tooling: the report cites its PMIDs as `PMID 23023331`, with a space rather than a colon, and ships no `.citations.md` sidecar. The fallback documented in the `initiate-new-disorder-creation` skill for sidecar-less reports is `grep -o "PMID:[0-9]*"`, which returns **zero** on this report even though it carries 16 real, resolvable citations. I will raise that separately.

## Validation

`just validate` passes — schema, terms and references — with **27/27 snippets verified**. `just validate-disorders` (the batched CI-equivalent) passes. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys` and `check-qualifier-terms` all pass.

Reference cache is pruned to the 9 PMIDs this entry actually cites; the term-cache diff is exactly 3 rows (`MONDO:0012273` and `hgnc:24579`).

Closes #10955

---

### DIFF
--- a/history/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48/2026-09-05T115739Z-claude-code-560938.yaml
+++ b/history/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48/2026-09-05T115739Z-claude-code-560938.yaml
@@ -1,0 +1,38 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48
+  path: kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48.yaml
+session:
+  id: 2026-09-05T115739Z-claude-code-560938
+  timestamp: '2026-09-05T11:57:39Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10972
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - pathophysiology
+  - evidence
+  summary: 'Correct an overclaim in the previous round: restore the calcium-buffering
+    candidate as a curated retirement'
+  details: |
+    Corrects an error made in the immediately preceding session on this entry, recorded in 2026-09-05T114037Z-claude-code-8fb4b4.yaml. That record and its commit state that the two open candidates removed from the stereocilia-growth node were uncited and were not what the source raises. The first half is true, the second is false, and the reviewer of #10972 caught it.
+
+    PMID:28663585 - already cached, and cited thirteen times in this entry - raises BOTH removed candidates in its discussion at references_cache/PMID_28663585.md:248: that CIB2 may be part of the myosin XVa/whirlin stereocilia elongation complex, and that as a calcium binding protein it may buffer free calcium near the MET channels at the stereocilia tips. So the previous round replaced a real, sourced pair of candidates with a different real, sourced candidate from a different paper and described the removal as correcting an invention. The earlier record is left in place rather than edited, because it accurately records what was believed at the time and the correction is more informative than a silent fix.
+
+    The node text is rewritten to name all three candidates with their sources, and to record two of them as retired on evidence the entry already carried rather than dropped. The myosin XVa/whirlin route is what the WHRN independence results in this same node argue against. The calcium-buffering route is undercut by PMID:26173970, whose ratiometric calcium imaging shows the p.F91S and p.R66W deafness substitutions do not alter CIB2 calcium buffering. The retirement is explicitly on balance rather than a disproof, because PMID:28663585 also reports that p.I123T and p.R186W do impair buffering, and that caveat now carries its own evidence item rather than sitting as unsourced prose - which is the failure mode this whole correction is about.
+
+    Three evidence items added, all from references already cached and cited: PMID:28663585 SUPPORT/DIRECT/MODEL_ORGANISM for the buffering candidate being raised, PMID:26173970 REFUTE/DIRECT/IN_VITRO for the two-allele buffering result, and PMID:28663585 SUPPORT/DIRECT/IN_VITRO for the p.I123T/p.R186W caveat. The last is graded IN_VITRO with the reason stated in its explanation: the sentence restates calcium-imaging results from primary work it cites, not this paper's own mouse experiments.
+
+    This push costs the approval the reviewer gave minutes earlier, and the reviewer had explicitly said the suggestion was not worth a re-review cycle on its own. It is not on its own: the entry itself asserted that GPSM2-GNAI is the candidate this literature raises, which is too narrow, and leaving a known-false characterisation in a merged history record was the worse trade.
+
+    Verification: 47/47 snippets, from 44/44. No new references, no new cache rows. validate, validate-terms, check-entity-refs, check-causal-targets, check-duplicate-keys, check-enum-values, check-qualifier-terms, check-snippet-length, check-title-snippets, check-snippet-grading all clean.
--- a/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48.yaml
+++ b/kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48.yaml
@@ -221,13 +221,27 @@ pathophysiology:
     the first-row stereocilia. Whirlin itself, along with myosin XVa, Eps8 and Eps8L2, still
     reaches the stereocilia tips in Cib2 mutants, so CIB2 is not required to deliver them.
 
-    The positive mechanism is not settled, and the candidate this literature actually raises is
-    a role for CIB2 in the GPSM2-GNAI stereocilia elongation complex. That is not a whirlin
-    route reintroduced under another name, but it is not a claim that WHRN is uninvolved in
-    elongation either: GPSM2-GNAI itself binds WHRN and is carried to the stereocilia tips by
-    MYO15A, so the open question is which complex CIB2 acts in, not whether WHRN participates
-    at all. The authors put it no higher than speculation and this entry does not adopt it as a
-    mechanism arm.
+    The positive mechanism is not settled, and three candidates have been raised across this
+    literature. PMID:28663585 offers two in its discussion: that CIB2 is part of the myosin
+    XVa/whirlin elongation complex, and that it buffers free calcium near the MET channels at
+    the stereocilia tips, where a local calcium change would alter actin remodelling.
+    PMID:40083274 offers a third, that CIB2 acts in the GPSM2-GNAI elongation complex.
+
+    Two of the three are disfavoured on evidence this entry already carries, and are recorded
+    here as retired rather than dropped. The myosin XVa/whirlin route is what the WHRN
+    independence results above argue against. The calcium-buffering route is undercut by
+    PMID:26173970: the p.F91S and p.R66W deafness substitutions do not alter the calcium
+    buffering ability of CIB2, so a buffering deficit cannot be the mechanism by which those
+    alleles cause disease. Neither is refuted outright - the buffering result covers two
+    alleles, and p.I123T and p.R186W are reported to impair buffering - so the retirement is on
+    the balance of the evidence, not a disproof.
+
+    That leaves GPSM2-GNAI as the live candidate, and it is worth being precise about what it
+    would and would not mean. It is not a whirlin route reintroduced under another name, but
+    neither is it a claim that WHRN is uninvolved in elongation: GPSM2-GNAI itself binds WHRN
+    and is carried to the stereocilia tips by MYO15A. The open question is which complex CIB2
+    acts in, not whether WHRN participates at all. The authors put it no higher than
+    speculation and this entry does not adopt it as a mechanism arm.
 
     The same paper separates this arm from the channel arm on the phenotype itself: rows 2 and 3
     elongate in Cib2 mutants, whereas loss of mechanotransduction current makes transducing
@@ -308,6 +322,39 @@ pathophysiology:
       Names the positive candidate this entry records as open. Quoted with the authors' own
       hedge intact, because the hedge is the point - it is why the candidate is described in
       the node text rather than curated as a second mechanism arm.
+  - reference: PMID:28663585
+    reference_title: "CIB2 interacts with TMC1 and TMC2 and is essential for mechanotransduction in auditory hair cells."
+    supports: SUPPORT
+    directness: DIRECT
+    evidence_source: MODEL_ORGANISM
+    snippet: "Alternatively, as a Ca2+ binding protein, CIB2 may buffer free Ca2+ in the vicinity of the MET channels at the tips of transducing stereocilia."
+    explanation: >-
+      Records that the calcium-buffering route is a candidate this literature genuinely raised,
+      not a stray idea. It is quoted so the retirement below is visibly a retirement rather than
+      an omission.
+  - reference: PMID:26173970
+    reference_title: "Novel and recurrent CIB2 variants, associated with nonsyndromic deafness, do not affect calcium buffering and localization in hair cells."
+    supports: REFUTE
+    directness: DIRECT
+    evidence_source: IN_VITRO
+    snippet: "do not affect the calcium-buffering ability of CIB2"
+    explanation: >-
+      Ratiometric calcium imaging of the p.F91S and p.R66W deafness substitutions in transfected
+      COS-7 cells. Disfavours the calcium-buffering candidate for these alleles: if buffering
+      were the route, a deafness-causing substitution should degrade it. Scoped to two alleles -
+      p.I123T and p.R186W are reported elsewhere to impair buffering - so this weighs against
+      the candidate rather than disproving it.
+  - reference: PMID:28663585
+    reference_title: "CIB2 interacts with TMC1 and TMC2 and is essential for mechanotransduction in auditory hair cells."
+    supports: SUPPORT
+    directness: DIRECT
+    evidence_source: IN_VITRO
+    snippet: "the p.I123T and p.R186W mutations only impair the Ca2+ buffering of CIB2"
+    explanation: >-
+      The reason the calcium-buffering candidate is retired on balance rather than refuted: two
+      other deafness alleles do impair buffering. Graded IN_VITRO because this sentence restates
+      calcium-imaging results from the primary work it cites, not this paper's own mouse
+      experiments, which are graded MODEL_ORGANISM elsewhere in this entry.
   - reference: PMID:28663585
     reference_title: "CIB2 interacts with TMC1 and TMC2 and is essential for mechanotransduction in auditory hair cells."
     supports: SUPPORT


aleixpuigb:
All seven IMPORTANT items addressed, plus most suggestions, in one push (`6d66af5a9`). Snippets went from 27/27 to **41/41 verified**.

**Item 2 was worse than flagged, and thank you for it.** You called the whirlin claim uncited. On fetching PMID:40083274 it turned out to be *wrong in direction*, not merely unsupported. The entry said the structural arm "is organised through whirlin rather than through TMC1/2". What that paper actually shows:

- CIB2 does bind WHRN — EF2 domain to HHD2 region, confirmed independently by AlphaFold2 multimer.
- But `Cib2KO/KO;Whrnwi/wi` double homozygotes show the **Whrn single-mutant** bundle phenotype, not an additive one.
- Overexpressing WHRN in `Cib2KO/KO` mice **does not rescue** the stereocilia morphology.
- Double heterozygotes hear normally, so there is no digenic interaction.
- What actually changes in Cib2-null hair cells is MYO15A, which over-accumulates at the tips of first-row stereocilia.

So the conclusion is the opposite of what I wrote: CIB2 has staircase functions *independent* of WHRN. The node is rewritten to say that, with three new evidence items including a `REFUTE` on the rescue experiment. This is the one place a plausible-sounding mechanism sentence had gone in without a citation, and it was wrong — which is a good argument for the rule rather than an exception to it.

**Item 3 and item 4 turned out to have one fix between them.** Rather than regrading PMID:42427029 to `MODEL_ORGANISM` and leaving a human phenotype resting on a mouse paper, I re-sourced the claim. PMID:29112224 contains an audiometry sentence I had missed:

> "Audiometry revealed a bilateral symmetric prelingual severe-to-profound hearing loss across all frequencies in all affected individuals."

That is human, measured, and settles laterality, symmetry, onset, severity and frequency involvement at once. It now supports the core phenotype, the severity phenotype, the new bilaterality phenotype (your 🔵 suggestion), and the onset phenotype — which is `HP:0000399` *Prelingual sensorineural hearing impairment*, exactly matching what the snippet says. PMID:42427029 now appears only on the gene-therapy treatment, where `MODEL_ORGANISM` is correct.

One small correction on that item: you gave `HP:0000399` with the label *Sensorineural hearing impairment*. The ID is right and is what I used; HPO calls it *Prelingual sensorineural hearing impairment*, which is why it fits the snippet so exactly.

**Items 1, 5, 6, 7** — all taken as described. `CL:0000609` on the vestibular node (you were right that binding the auditory term inverted the whole point of that node). `animal_models` for the `Cib2` knockout and the `Cib2 p.F91S` knock-in, each with `modeled_mechanisms`, `limitations` and `readouts`; the knock-in is linked to the molecular node specifically because it carries the prevalent human allele rather than a null. A `diagnosis` section. And `conforms_to: "sensorineural_hair_cell_loss#Hair Cell Mechanotransduction Failure and Death"` with the module's `GO:0006915` binding.

**Suggestions taken:** bilaterality phenotype; `GO:0050910` `preferred_term` corrected to the ontology label and `modifier` changed `LOSS_OF_FUNCTION` → `DECREASED` (you are right that the qualitative form should be reserved); auditory-verbal therapy and genetic counseling treatments.

**Two suggestions I did not take, both for the same reason** — no quotable source in the cached abstracts:

- **`progression:`** — the deep-research report describes stable, non-progressive loss, but no cached abstract states it. The report's own claim is uncited there, and I would rather leave the section absent than assert non-progression on the report's authority.
- **Hearing aids** — clinically obvious, but nothing I have cached says it. Cochlear implantation is cited; amplification is not.

Both are genuine gaps rather than disagreements, and a curator with full-text access should add them.

**One incidental find worth noting:** `NCIT:C159273` needs the label *Speech Language Therapy*; `Speech Therapy`, which is how `CLAUDE.md`'s modality table names it, fails term validation on the label check.

Validation after the round: `just validate` and `just validate-disorders` both pass at 41/41. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-title-snippets` and `check-snippet-length` all pass. A `history/` record for this round is included in the same push.

Note there is also an earlier commit on this branch (`25c7cc103`) that predates your review: a connectivity audit found two audiological phenotypes with no incoming mechanism edge, so they are now downstream of `Hair Cell Dysfunction and Death`.

---

aleixpuigb:
The re-review checked out `25c7cc1`, not the fix commit. Timeline:

| | |
|---|---|
| 17:15:17Z | re-review job starts, checks out head `25c7cc1` |
| ~17:20Z | I push `6d66af5a9` with all seven items fixed |
| 17:18:52Z | re-review reports on `25c7cc1` |

So "all seven IMPORTANT items are still outstanding" was accurate for the commit reviewed and is not true of the branch. On `6d66af5a9`, verified just now:

| # | Item | State on head |
|---|---|---|
| 1 | `CL:0000202` on vestibular node | `CL:0000609` bound |
| 2 | Whirlin asserted without evidence | rewritten, PMID:40083274 cited ×3 |
| 3 | `HUMAN_CLINICAL` on murine PMID:42427029 | claim re-sourced to human audiometry; that PMID now only on the gene-therapy treatment as `MODEL_ORGANISM` |
| 4 | `HP:0008527` / "at or before birth" | `HP:0000399`; zero occurrences of `HP:0008527` remain |
| 5 | No `animal_models:` | present (Cib2 KO + p.F91S knock-in) |
| 6 | No `diagnosis:` | present |
| 7 | No `conforms_to` | `sensorineural_hair_cell_loss#Hair Cell Mechanotransduction Failure and Death` |

That is my fault for pushing into a running review, and it cost you a cycle — noted, and it's exactly the failure mode your process note warns about from the other direction.

**Your item-4 self-correction and mine crossed in flight**, and we reached the same place independently: `HP:0000399` is *Prelingual sensorineural hearing impairment*, which is an exact match for the snippet rather than a downgrade. That is what's bound. The bare-name downstream target was updated in the same push, and `check-causal-targets` is clean.

**Your item-2 analysis was new and I've acted on it** — pushed as `93658758a`.

You checked what PMID:28663585 *did* with whirlin, which I had not. It cuts against the whirlin route rather than merely failing to support it: whirlin, myosin XVa, Eps8 and Eps8L2 all still reach the stereocilia tips in Cib2 mutants, so CIB2 is not required to deliver them. That's now a second evidence item on the node.

More importantly, it made me re-read my own fix and find it overclaiming in the opposite direction. I had replaced "organised through whirlin" with a flat "independent of WHRN", which reads as though the mechanism is settled. It isn't. The description now says what is actually established — the staircase function is *separable* from WHRN — and names the whirlin/myosin XVa complex and local calcium buffering as the two unresolved candidates, without choosing. So the node went from a wrong positive claim, to a right-but-overreaching one, to the hedge the literature supports.

Snippets: 42/42.

**One suggestion declined, with reasoning.** On collapsing the audiological phenotypes into one with `severity:`/`onset:` qualifiers: I'm keeping the separate HPO terms. Sibling entries carry between one and three hearing-loss HP terms (`ARNSHL_104` has three), so multiple granularities is within convention; each term is independently verifiable from the same measured audiometry sentence; and explicit HPO terms are what downstream phenotype matching consumes, where a free-text severity qualifier is not. There are now four rather than three, since bilaterality was your earlier suggestion and I took it. You said reasonable people curate this either way and I agree — flagging the choice rather than making it silently.

Current head is `93658758a`. Both `just validate` and `just validate-disorders` pass at 42/42; `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-qualifier-terms`, `check-title-snippets` and `check-snippet-length` all pass. History records for both rounds are included.

---

aleixpuigb:
Thanks — stopping here rather than taking the two 🔵 suggestions, since `main` dismisses stale approvals and neither is worth a re-review cycle. Recording both so they are not lost:

**The candidate-mechanism tension is real and you are right about the fix.** Naming "a whirlin/myosin XVa elongation complex" as an open candidate immediately after arguing the function is separable from WHRN does read as self-contradictory, even though both halves are sourced — the two candidates come from PMID:28663585's Discussion while the argument against the whirlin route comes from PMID:40083274's experiments. Attributing each candidate to its own paper and adding GPSM2–GNAI, which is what the later paper actually speculates, would resolve it. Worth doing on the next edit to this entry.

**Phenotype granularity**: agreed the base `preferred_term` no longer needs to restate what the sibling phenotypes now say, and the `OBLIGATE` vs `VERY_FREQUENT` split off one "in all affected individuals" snippet is a hedge rather than a distinction. Same disposition — next edit.

I have added your `NCIT:C159273` finding to #10978, which is the issue for `CLAUDE.md` recommending ontology terms that fail validation. It already covered `NCIT:C2322` (offered as a `therapeutic_agent` example but unreachable from `NCIT:C1909`, because NCIT files corticosteroids under Hormone); the speech-therapy label is a second instance of the same class, so it belongs there rather than in a new issue.

Also worth recording, since it is the one durable lesson from this PR: the whirlin clause was the single place I let a plausible-sounding mechanism sentence in without a citation, and it was wrong in direction. Your catch turned it from a wrong positive claim into an overreaching one and then into the hedge the literature supports.

---

aleixpuigb:
Merge-conflict resolution only — no content change. Flagging what this push is, since it costs the approval and I'd rather you not re-derive it.

This PR went un-mergeable against `main` overnight and was skipped by the shepherd's sweep while its three siblings were enqueued. The conflict was a single adjacent row in `cache/hgnc/terms.csv`: someone curated STEAP3 (`hgnc:24592`) on `main` while this branch added CIB2 (`hgnc:24579`), and the two sort next to each other.

Resolved by merging `origin/main` in (not rebasing — this branch has a reviewed history worth preserving), keeping **both** rows, and letting `just normalize-cache` place them rather than hand-positioning. Both are present and `just check-cache-order` and `just check-term-cache-integrity` pass.

What did **not** change: `kb/disorders/Autosomal_Recessive_Nonsyndromic_Hearing_Loss_48.yaml`, its history records, and its reference caches are byte-identical to the tree you approved. 42/42 snippets still verify and `just validate-disorders` passes on the merged tree.

One thing I checked because a merge makes it easy to get wrong: the merge brings in 105 files from `main`, including three paths that must never carry a curation PR's changes — `cache/dataset_accessions.json`, `app/data.js` and `app/models/data.js`. All three are byte-identical to `origin/main`; the branch contributes nothing to them.

The two cosmetic suggestions from your approval are still deliberately not taken — the whirlin candidate-mechanism sentence and the overlapping phenotype `preferred_term`. They were next-edit items when you raised them and this push doesn't change that.

---

aleixpuigb:
**`claude-review` is red on this PR and the failure is not this PR's.** Standing down rather than re-running, with the reason.

The review agent errored at 0s. From the job log ([run 33951720604](https://github.com/monarch-initiative/dismech/actions/runs/33951720604)):

```
detail="You've hit your limit · resets 11:20am (UTC)"
```

An account usage limit on the credential the workflow ran with — no review was attempted against the diff. [#10995](https://github.com/monarch-initiative/dismech/pull/10995) failed identically in the same minute, so this is repo-wide rather than about either branch. Full detail and a note for whoever owns the workflow config is on that thread; I'm not duplicating it here.

**I have not re-run it.** The limit is deterministic until 11:20 UTC, so a re-run now fails the same way. I'll re-request the review after the reset.

`test (3.13)` is still running and unaffected. The push under test is the merge-conflict resolution described in my previous comment — merge-only, with the entry, history records and reference caches byte-identical to the approved tree.

---

aleixpuigb:
/review

---

aleixpuigb:
Pushed `bb8d7e72f`, taking up both non-blocking items from your approval. The first turned out to be worse than "cosmetic" and you were right to name it.

**The candidate-mechanism sentence was not just awkward, it was uncited and wrong.** The node closed with:

> The two candidates raised in this literature are a whirlin/myosin XVa elongation complex and local calcium buffering at the stereocilia tips

Neither had an evidence item, and neither is what PMID:40083274 raises. Its speculation is the **GPSM2-GNAI** elongation complex — *"it could be speculated that CIB2 has a role in the GPSM2-GNAI stereocilia elongation complex."* Calcium buffering at the tips appears nowhere in the cached full text; I cannot reconstruct where it came from, which is the point — it read as sourced because it sat in a node whose every other claim is.

The paragraph now names GPSM2-GNAI, quoted with the authors' hedge intact because the hedge is why it stays prose rather than becoming a second mechanism arm. It also states the thing your note implied but the old text got backwards: GPSM2-GNAI is not a whirlin route smuggled back in, but neither is it a claim that WHRN is uninvolved in elongation — GPSM2-GNAI binds WHRN and is carried to the tips by MYO15A. The open question is which complex CIB2 acts in.

**One thing you didn't ask for, from the same paragraph of the same paper.** Its MET-independence argument was missing entirely: rows 2 and 3 *elongate* in Cib2 mutants, where loss of transduction current makes transducing stereocilia *retract*. That is the observation justifying this node sitting parallel to the transduction node rather than beneath it, so the entry was asserting the parallel structure without the evidence for it. Curated as its own item. Both new items are `MODEL_ORGANISM` / `DIRECT`.

**On the four audiological phenotypes — I kept all four, and explained why instead.** HPO has no single term composing onset, laterality and severity, and the three specific terms carry genuinely different frequencies from different sources (`VERY_FREQUENT`, `VERY_FREQUENT`, `FREQUENT`), so collapsing them loses information. What was wrong is that nothing said so: the `HP:0000407` anchor's `preferred_term` was "Bilateral prelingual sensorineural hearing loss", restating all three axes and making the set look like an accident. It now matches its label, its `description` points at the siblings, and a `notes` block records that these are one annotation decomposed rather than four competing granularities — so a later curator doesn't "tidy" three of them away.

If you'd still rather see them merged, say so and I'll do it in one push.

44/44 snippets verified, up from 42/42. `validate-disorders` passes; entity refs, causal targets, duplicate keys, enum values, qualifier terms, snippet length, title snippets and snippet grading all clean, and `validate-history` on the new record. No cache rows — both new snippets come from a reference already cached and nothing new is bound.

**Why this is a push and not a re-review request:** the approval here was dismissed by the merge-conflict resolution, yesterday's automatic re-review errored at 0s on a usage limit, and after the reset both documented ways to ask for a new one returned 403 for this integration while the `/review` comment above was silently skipped by the workflow's exact-match gate (the harness appends an attribution footer). That is [#9321](https://github.com/monarch-initiative/dismech/issues/9321); today's occurrence is recorded there. `synchronize` is the only trigger left, so the work you'd already asked for is what carries it.

---

aleixpuigb:
You're right, and it's worse than a commit message being imprecise — I've pushed `80e845be6` to fix it, knowing that costs the approval you gave minutes ago. Reasoning below, because spending a cycle against your explicit "not worth a re-review cycle on its own" needs justifying.

**Verified your correction first.** `references_cache/PMID_28663585.md:248` raises *both* candidates I removed:

> The long isoform of myosin XVa is particularly interesting, because CIB2 may bind to whirlin at the tips of stereocilia and be a part of the myosin XVa/whirlin stereocilia elongation complex. Alternatively, as a Ca2+ binding protein, CIB2 may buffer free Ca2+ in the vicinity of the MET channels at the tips of transducing stereocilia.

That paper is cited thirteen times in this entry. So the previous round replaced a real, sourced pair of candidates with a different real, sourced candidate from a different paper, and called the removal a correction of an invention. The uncited half of my claim was true; "not what the source raises" was false.

**Why this wasn't a "suggestion" I could park.** Your framing was that adding the retirement is optional, and on its own I'd have taken your advice. But the sentence I'd left in the entry reads "the candidate this literature actually raises is a role for CIB2 in the GPSM2-GNAI stereocilia elongation complex" — that is KB content asserting a false thing about the literature, not a commit message. And the history record repeating it was about to merge into an append-only tree. A known-false characterisation in curated content outweighs a review cycle; a missing optional candidate would not have.

**What the node says now.** All three candidates named with their sources, and two recorded as *retired on evidence the entry already carried* rather than dropped:

- **myosin XVa/whirlin** — retired by the WHRN-independence results already in this node.
- **Ca²⁺ buffering** — retired by PMID:26173970, whose ratiometric imaging shows the p.F91S and p.R66W deafness substitutions don't alter CIB2 buffering. `REFUTE` / `IN_VITRO`.
- **GPSM2-GNAI** — the live candidate, unchanged.

And I took your "balance of evidence, not a disproof" point literally rather than as a hedge: PMID:28663585 also reports *"the p.I123T and p.R186W mutations only impair the Ca2+ buffering of CIB2"*, so the buffering result covers two alleles and not the class. That caveat now carries its own evidence item instead of sitting as unsourced prose — which is exactly the failure mode this whole correction is about, and I'd have reintroduced it one paragraph later otherwise. Graded `IN_VITRO` with the reason stated in the `explanation`, since that sentence restates calcium-imaging results from work it cites rather than this paper's own mouse experiments.

**On the history record.** I left `…8fb4b4.yaml` in place rather than editing it, and added `…560938.yaml` recording the correction. It accurately records what I believed at the time, and the correction is better provenance than a silent fix — a later reader gets both the error and the catch.

47/47 snippets verified, up from 44/44. No new references, no new cache rows — everything came from files already committed here. `validate-disorders` clean; entity refs, causal targets, duplicate keys, enum values, qualifier terms, snippet length, title snippets, snippet grading and `validate-history` all clean.

Two of the three things you've caught across this PR's rounds have been unsourced or wrongly-sourced assertions sitting in prose fields that no gate reads. That's the most useful thing to come out of this run and I'd rather have it on the record than quietly fixed.

---

PR #10980, opened 2026-09-04
Record 13 deliberate no-link decisions on environmental entries (#8085)

caufieldjh:
Companion to #10960. **No pathograph links.** This records, in each entry's own `review_notes:`, that **13 exposures across 10 entries are deliberately left unlinked**, and why. Insertions only: **581 / 0** (10 entries + 10 history records).

> **History note.** The first revision of this PR dispositioned **34** entries with a single categorical reason. @ai4c-reviewer showed that reason is false as a description of this KB, and a scan confirmed it: **58 host-state or demographic exposures are already linked across 40 files**. Commit `a25f06e` dropped 17 entries and made every reason entry-relative. A second review then found three surviving reasons that the repository itself contradicts; `afb661e` drops those three (Age_Related_Macular_Degeneration/Older Age, Essential_Tremor/Aging, Benign_Prostatic_Hyperplasia/Age) and rewrites two more. A later round found a fourth of the same shape and `16f0953` drops Glaucoma/Age: that entry already cites PMID:26497784 naming aging as accelerating RGC apoptosis, on the node next to the one it would act on. Detail in [this comment](https://github.com/monarch-initiative/dismech/pull/10980#issuecomment-5544167955).

## Why this is a PR and not a silence

Today's survey of `main` found **1061 environmental entries, 704 linked, 357 unlinked, and exactly 2** whose entry text recorded the no-link as deliberate. The 42 earlier tranches did make these decisions and put the reason in the **history record**, where a reader of the knowledge base cannot see it and a survey counts the entry as unfinished. So the ~355 remaining are not 355 pieces of pending work, and this converts one slice into a recorded decision.

## What is *not* claimed

**There is no rule here against linking a host state or a demographic.** The KB does it in 58 places, including `Osteoarthritis` Age → Chondrocyte Senescence, `Endometrial_Carcinoma` Obesity → Unopposed Estrogen Signaling, `Heart_Failure` Hypertension → Ventricular Remodeling. Every note ends by saying so and pointing at Osteoarthritis, so a later curator reads the limit and not a false generalisation.

Each disposition instead names what **this entry's own evidence and nodes** lack.

| Entry | Exposure | The reason, in short |
|---|---|---|
| Arteriosclerotic_Retinopathy | Hypertension | The entry's **own description** declines the causal claim |
| Arteriosclerotic_Retinopathy | Type 2 diabetes | Its description says cross-sectional, no causal direction |
| Cerebral_Amyloid_Angiopathy | Hypertension | Quote is the fragment *"vascular risk factors and concomitant medications"* |
| Dieulafoy_Lesion | Hypertension, Diabetes | A comorbidity prevalence table; the lesion is a congenital artery |
| Benign_Prostatic_Hyperplasia | Obesity, Diabetes | One undifferentiated list is the only support for two entries |
| Glaucoma | Family History | A heritability estimate, already carried in `genetic:` |
| Mixed_Germ_Cell_Tumor | Family History | 78 loci, 44% heritability — a genetic finding |
| Choriocarcinoma | Extremes of Maternal Age | Incidence bands; the chain opens on paternal genome content |
| Central_Serous_Chorioretinopathy | Male Sex | A sex ratio plus an eQTL offered to explain it; sex and hormones bundled |
| Colon_Adenocarcinoma | Obesity | RR 1.19 incidence; no adipose, inflammatory or insulin node |
| Renal_Cell_Carcinoma | Obesity | RR 1.28/1.77 incidence; chain runs VHL → HIF |

**Cerebral_Amyloid_Angiopathy states what would reverse it**: a cited sentence supporting vessel-wall weakening. A disposition that names its own falsifier is worth more than one that does not.

## 17 entries dropped, for a linking tranche

Each has a node the exposure plausibly acts on, several with a sibling that already links it:

`Carotid_Stenosis` · `Chronic_Kidney_Disease` ×2 · `Peripheral_Artery_Disease` ×2 · `Endometrial_Endometrioid_Adenocarcinoma` · `Obstructive_Sleep_Apnea` · `Type_2_Diabetes_Mellitus` · `Psoriasis` · `Gastroesophageal_Reflux_Disease` · `Kummell_Disease` · `Liver_Cirrhosis` · `Pouchitis` · `Sclerosing_Cholangitis` · `Postpartum_Depression` · `Cauda_Equina_Syndrome` · `Breast_Carcinoma`

`Carotid_Stenosis` is the clearest: its own snippet says those comorbidities *"played the most salient role in plaque development"*, it has an `Atherosclerotic carotid plaque formation` node, and `Carotid_Artery_Occlusion` links the identical exposures to a node of the identical name.

## `review_notes`, not `notes`

Adopted from the suggestion on #10960. `notes:` is disease content and renders on the public disorder page (`disorder.html.j2:6209`); `review_notes:` is the curation record and renders nowhere (0 hits in the template). `scripts/check_environmental_evidence.py` states the split in its own docstring and keys its waiver on `review_notes` exclusively. The waiver sentinel is a prefix match on `"left deliberately uncited."`, a different sentence, so **the waiver count is unchanged at 4**.

## Also excluded from the pass at the outset

The name-keyword bucket was mixed, so all 49 candidates were read rather than noted wholesale. Genuine exposures left for a linking tranche include `Placenta_Accreta_Spectrum`/Previous caesarean delivery (its second node is *Uterine Scar*), `Fanconi_Anemia`/DNA-damaging agents, `Merkel_Cell_Carcinoma`/Immunosuppression, and `Acute_Macular_Neuroretinopathy`/Oral Contraceptive Use. `Ebstein_Anomaly` already records its decision, in `description:`, which is the wrong slot by the argument above.

## Two things flagged and not fixed here

**Cerebral_Amyloid_Angiopathy's `description:` over-asserts** relative to its four-word evidence fragment. The disposition records that, but `review_notes:` renders nowhere while `description:` renders on the public page, so a reader still sees the claim. Editing curated disease prose is a different change from recording a disposition and belongs in its own PR with evidence work.

`check-snippet-boundaries` reports `Kummell_Disease.yaml:progression[2].evidence[1]` (PMID:40613656) quoting a mid-word fragment. It **reproduces on unmodified `main`**, sits in `progression:`, and is unrelated.

## Validation

```
linkml-validate               No issues found (all 12)
count-verified-snippets       439/506 verified, 67 skipped by prefix, 0 failing
check-causal-targets          OK      check-entity-refs         OK
check-duplicate-keys          OK      check-enum-values         OK
check-folded-hyphens          OK      check-title-snippets      OK
check-snippet-length          OK      check-environmental-evidence  OK (4 waivers, unchanged)
just validate-history-all     No issues found
git diff <base>..HEAD --numstat  581 insertions / 0 deletions
```

Reference-cache churn from `just validate` was reverted, not committed.

https://claude.ai/code/session_01FjG2TgWtZupYyPTFN2T8Z4


caufieldjh:
**Correction to one line in the body.** It says I will move the #10960 disposition to `review_notes` "in a follow-up on that branch". That is no longer possible: #10960 entered the merge queue before I could push, and GitHub rejects updates to a queued branch (`GH006`). I am not dequeuing an approved PR for a non-blocking change.

The move will land as its own small PR once #10960 merges. It is one line, `notes:` → `review_notes:`, with byte-identical text; the patch is already written and verified by parsing both revisions. Nothing else about this PR changes.

caufieldjh:
You are right, and the check is worse than the finding stated. I scanned every `environmental:` entry whose name matches obesity, hypertension, diabetes, age or aging and that carries `influences_mechanisms`: **58 such links across 40 files**, including `Osteoarthritis` Age → Chondrocyte Senescence, `Endometrial_Carcinoma` Obesity → Unopposed Estrogen Signaling, `Heart_Failure` Hypertension → Ventricular Remodeling, and `Carotid_Artery_Occlusion` linking both Hypertension and Diabetes mellitus to *Atherosclerotic carotid plaque formation*.

So the sentence I wrote 34 times is not the rule this KB follows. Pushed `a25f06e`, which does two things.

## 17 entries dropped from this PR

Not re-worded. **Removed**, because each has a node the exposure plausibly acts on and several have a sibling that already links it. They belong in a linking tranche:

`Carotid_Stenosis` (cardiometabolic comorbidities) · `Chronic_Kidney_Disease` ×2 · `Peripheral_Artery_Disease` ×2 · `Endometrial_Endometrioid_Adenocarcinoma` · `Obstructive_Sleep_Apnea` · `Type_2_Diabetes_Mellitus` · `Psoriasis` · `Gastroesophageal_Reflux_Disease` · `Kummell_Disease` · `Liver_Cirrhosis` · `Pouchitis` · `Sclerosing_Cholangitis` · `Postpartum_Depression` · `Cauda_Equina_Syndrome` · `Breast_Carcinoma`

**`Carotid_Stenosis` is the clearest of the set**, clearer than either you named: its own snippet says those comorbidities *"played the most salient role in plaque development"*, it has an `Atherosclerotic carotid plaque formation` node, and `Carotid_Artery_Occlusion` links the identical exposures to a node of the identical name. I had dispositioned an entry whose own evidence states the mechanism.

Beyond your two: `Sclerosing_Cholangitis` carries an *Aberrant Gut-Primed Lymphocyte Trafficking* node, which **is** the IBD-to-liver route; `Postpartum_Depression` cites a steroid-withdrawal challenge in which 5 of 8 women with prior PPD and 0 of 8 controls developed symptoms, which is a mechanism experiment, not an association.

## 17 kept, each reason now naming this entry's own gap

No shared categorical clause. Examples:

- **`Arteriosclerotic_Retinopathy`** ×2 — cites the entry's *own description* declining the claim: *"support a vascular-risk context, not a claim that every Scheie S sign is caused only by hypertension"*, and for diabetes, *"cross-sectional correlations do not show... causal direction"*. An edge would assert what the entry withholds.
- **`Cerebral_Amyloid_Angiopathy`** — the quoted text is the fragment `"vascular risk factors and concomitant medications"`, which names no exposure, direction or target. The description asserts more than the evidence does. The note **says what would change the decision**: a sentence supporting vessel-wall weakening.
- **`Glaucoma`/Age** — names the Osteoarthritis contrast directly: that entry has a senescence node, this one has four pressure and neurodegeneration nodes.
- **`Benign_Prostatic_Hyperplasia`** Obesity and Diabetes — one sentence listing obesity, diabetes and meat and fat consumption is the *only* support for both entries, so a single undifferentiated list was being asked to carry two edges.
- **`Dieulafoy_Lesion`** ×2 — a comorbidity prevalence table (hypertension 29%, diabetes 21%) reports what patients had, not what it did, and the lesion is a congenital caliber-persistent artery.

Each note still ends by stating that this is **not** a rule against linking host states, and points at Osteoarthritis, so a later curator reads the limit rather than the false generalisation.

On the byte-identical point: the 17 remaining reasons are distinct per entry, and the only shared text is that closing pointer. Insertion-only, **691 / 0**, 12 entries and 12 history records. Gates re-run: linkml-validate clean on all 12, 439/506 snippets verified with 67 prefix skips and none failing, causal-targets, entity-refs, duplicate-keys, enum-values, environmental-evidence (4 waivers, unchanged), folded-hyphens, title-snippets, snippet-length all OK, validate-history-all clean.

caufieldjh:
Both points taken.

**The stale body was a cache.** I rewrote it at 17:32 UTC, before the review ran, and it currently reads "17 exposures across 12 entries … 691 / 0" with the pre-fix state kept only as an explicit history note. Worth flagging that the review read a cached copy, since the squash-message concern is real and would have been right against the old text.

**The frozen count is fixed** in `33ddeb4`. Each note said the KB links such exposures "in 58 places". True today, and it sits in permanent entry text where it reads as a standing claim rather than a dated one. Replaced with "elsewhere".

The exemplar stays, because **the exemplar is the durable evidence and the count was not**. A reader can open `Osteoarthritis.yaml:273` and check that Age points at a Chondrocyte Senescence node; the number only told them somebody had counted once. The measurement keeps its precision where precision belongs, in the 12 history records, which are dated provenance rather than standing content.

One line per note, 12 files, nothing else touched. `linkml-validate` clean, `check-environmental-evidence` still 4 waivers.

This dismisses the approval, which I would not have spent on a non-blocking point except that the defect is in permanent KB content and gets harder to correct once merged.

caufieldjh:
All three verified against the files before acting, and all three were right. `afb661e` drops three and rewrites two.

## Finding 1 — CSCR: the sentence pointed at a route that does not exist

Confirmed: `grep -c influences_mechanisms kb/disorders/Central_Serous_Chorioretinopathy.yaml` → **0**. The file has no links at all, so `Corticosteroid Exposure` is itself unlinked and my sentence was false about the one thing it asserted.

Rewritten to stand alone, and I did not link corticosteroid here: that is a linking decision and this is a disposition PR. The reason is now the sex ratio (72–87.5%) plus an eQTL offered to explain that ratio, neither stating an action on a node, **plus** a point the original missed: the entry bundles male sex with hormonal factors without citing a sentence about hormone action on the choroid, so an edge would have to choose between two claims the entry does not separate.

## Finding 2 — AMD: dropped

Confirmed: `Cigarette Smoking` at :858 targets `RPE Oxidative and Lysosomal Stress`, cites the same **PMID:38193957**, and its own description says the intermediates are "recorded as known on that mapped route rather than on the quoted sentence, which reaches only the disease". So "the evidence does not name the node" is demonstrably not this file's standard. **Dropped to the linking set**, as you suggested — I could not state a difference from smoking that survives inspection.

## Finding 3 — Essential_Tremor: dropped, and Benign_Prostatic_Hyperplasia/Age with it

Confirmed at `Osteoarthritis.yaml:275-283`: `PREDISPOSES` / `INDIRECT_UNKNOWN_INTERMEDIATES`, description "reports a prevalence gradient rather than a count of senescent cells". So the grade exists exactly to carry that.

**I dropped `Benign_Prostatic_Hyperplasia/Age` too**, which you raised only as a SUGGESTION. It rested on the same epidemiology argument, and your suggestion noted the file has `Oxidative Stress and Epithelial Plasticity` and `Inflammation and IL-6 Signaling` nodes my note had not considered. Same defect, so same treatment.

**No surviving reason rests on a citation being epidemiological**, and the closing sentence now says the Osteoarthritis precedent rests on a prevalence-gradient citation. That turns the exemplar into a limit on the notes rather than cover for them. `Glaucoma/Age` was reordered to lead with the absent senescence node, since it half-leaned on the argument the closing sentence now disclaims.

## Suggestions taken

- **Mixed_Germ_Cell_Tumor** no longer claims `genetic:` carries the germline loci. It carries i(12p), KIT, KRAS, TP53, which is now stated, and strengthens the note: inherited susceptibility is represented neither there nor by any node.
- **Partial coverage** is now stated in every note: "Other exposures in this file remain undispositioned."
- **Boilerplate trimmed** from ~50 words to two sentences.
- **CAA description**: not fixed. You are right that the note diagnoses an over-assertion in a slot that renders nowhere while the over-asserting `description:` still renders. Editing curated disease prose is a different change from recording a disposition, and doing it on my own judgement without evidence work is how the first version of this PR went wrong. Flagged in the PR body as a follow-up.

**14 dispositions across 10 entries**, down from 34. Net against the branch base: **597 / 0**. Gates re-run on the pushed tree: linkml-validate clean on all 10, 353/407 snippets verified with 54 prefix skips and none failing, causal-targets, entity-refs, duplicate-keys, enum-values, environmental-evidence (4 waivers, unchanged), folded-hyphens, validate-history-all all OK.

Three rounds, and each one removed entries rather than adding them. The nine you called sound are the nine that survived unchanged.

caufieldjh:
Both remaining points taken.

**The epidemiology wording.** You are right that the catch lands on my own commit message, not just the text: it claimed no surviving reason rests on a citation being epidemiological, while `Colon_Adenocarcinoma` and `Renal_Cell_Carcinoma` opened on the relative risk before reaching the discriminator. `b9b62ab` reorders both to lead with the absent node and mention the risk second, which is the same fix already applied to `Glaucoma/Age`. Two files, nothing removed, only resequenced. The claim and the text now agree.

**The Kummell_Disease boundary defect** is filed as #10991, with the reproduce command, why the other two snippet checks pass on it, and the note that the fix is to re-quote rather than touch the cache. I also recorded there that I have not checked whether other entries carry the same defect, and that a whole-`kb/` run would answer it.

**Cerebral_Amyloid_Angiopathy stays as is**, for the reason given last round: the `description:` over-asserts and the diagnosis sits in a slot that renders nowhere, which is a real asymmetry, but editing curated disease prose is a different change from recording a disposition. It is flagged in the PR body as a follow-up rather than done quietly here.

Net against the branch base is unchanged in shape: **597 / 0**, 14 dispositions across 10 entries, 10 history records.

caufieldjh:
Both verified in the files before acting, and both were right. `71c0653`.

**RCC was an overstatement.** The `Impaired Cuproptosis via PRR15/NF-κB/FDX1 Axis` description ends:

> "...while maintaining enhanced lipid metabolism and altered oxidative stress regulation in the tumor microenvironment"

So "no metabolic or adipose node" was false as written. The note now says **no adipose or insulin node**, names the lipid-metabolism mention rather than pretending it is absent, and gives the reason it does not change the decision: it is a property of the tumor microenvironment, not a step an exposure enters. Correcting this mattered more than the wording suggests, since a note whose stated reason is factually wrong is the exact defect the first two rounds of this PR were about.

**Colon_Adenocarcinoma was true and unhelpful.** You are right that `GUCY2C Tumor-Suppressive Axis Silencing` turns on loss of guanylin and uroguanylin, and that this is the route the obesity-CRC literature actually uses. A disposition that lists which nodes are absent, while the one relevant node sits unnamed in the same file, warns a future tranche off the entry instead of pointing it at the place to look.

The note now names GUCY2C as where a link would attach and says plainly that no sentence cited in this entry makes that connection. **The decision is unchanged; what changes is that it now reads as "not yet, and here is where" rather than "nothing here".** That seems the right shape for every disposition in this series, and I would apply it to the others if a candidate node were as clearly identifiable.

Two files. Insertion-only against the branch base, `just check-folded-hyphens` clean, linkml-validate clean on both.

Six rounds now, and the through-line is consistent: every finding has been a reason that was wrong or overstated rather than a mechanical defect, and none of them would have been caught by a gate.

caufieldjh:
Both taken. `933af01`.

**The history record was the one that mattered.** The RCC record still carried "carries no metabolic or adipose node", the exact clause the previous commit corrected in the entry, so the provenance recorded a claim already conceded as wrong. That is worse than a stale paraphrase: a history record is the durable account of why a decision was made, and this one gave a reason the same branch had already withdrawn. Both the renal and colon records now carry the reason as it stands in the entry, GUCY2C candidate included. Session id, timestamp and all record metadata unchanged, with a round-trip check confirming only the reason line differs.

**The body count is corrected** to 602 / 0, which matters because the squash message derives from it.

Cerebral_Amyloid_Angiopathy still deferred, on the same reasoning as before.

Net against the branch base: **602 / 0** across 20 files, 14 dispositions, 10 entries, 10 history records.

caufieldjh:
Both taken, and the first one was worse than reported. `5762a4b`.

**The closing sentence.** You found `Arteriosclerotic_Retinopathy`. Checking all ten found a second: `Mixed_Germ_Cell_Tumor` has three environmental entries, one dispositioned here and **two already linked**, so nothing remains open there either. Your check was for entries this PR dispositions; the sentence is also false when the remainder are linked rather than noted.

It is now appended only where the file actually leaves an exposure neither linked nor dispositioned, computed rather than asserted:

```
ok  Arteriosclerotic_Retinopathy     Hypertension...        open=0  sentence=False
ok  Arteriosclerotic_Retinopathy     Type 2 diabetes...     open=0  sentence=False
ok  Benign_Prostatic_Hyperplasia     Obesity                open=2  sentence=True
ok  Central_Serous_Chorioretinopathy Male Sex               open=2  sentence=True
ok  Glaucoma                         Age                    open=3  sentence=True
ok  Mixed_Germ_Cell_Tumor            Family History         open=0  sentence=False
...  mismatches: 0
```

You are right that this is the same shape as the earlier findings, and it is worth naming the difference: rounds one and two were **wrong arguments**, this one is **boilerplate applied without checking it held**. A fixed closing clause repeated fourteen times cannot be true fourteen times unless someone verifies it, and I did not. The generator now computes it per file, so it cannot drift as entries are linked later.

The two history records state that the omission is deliberate, so a later reader does not restore the sentence to make the notes look uniform.

**The body counts** are corrected. Both the opening line and the code block now read **603 / 0**; the `691` in the validation block was a leftover from the first revision, which is the one that fed the squash message. `grep -c "691\|597\|602"` on the body now returns 0.

Net against the branch base: 603 insertions, 0 deletions, 20 files. linkml-validate clean on both changed entries, check-folded-hyphens clean, validate-history-all clean.

caufieldjh:
Taken, and you are right that appending was the weaker move. `97801b4`.

The boilerplate line in both records now states the truth for those two files directly, and the appended paragraph is gone, so each record makes one claim instead of two two lines apart. Round-trip checked: only the details text differs, record metadata untouched.

**A note on how that was caught.** My first attempt at this edit was wrong and an assertion stopped it before anything was written. The replacement text begins "All environmental entries in this file are now...", which is the same string I was searching for to locate the appended paragraph, so after substituting the boilerplate line the search found the line I had just written and pointed at itself. `assert j != i` fired and nothing was saved. Reordered to locate both lines first, then delete, then substitute.

Worth stating because the failure mode is generic: a scripted edit whose replacement text matches its own search pattern will silently consume the wrong line if nothing asserts on it. It cost one retry rather than a corrupted record.

**Body count** corrected to **599 / 0**, down from 603, since this commit removes more than it adds.

Remaining, both correctly deferred and both stated in the body: `Cerebral_Amyloid_Angiopathy`'s over-asserting `description:` belongs in its own PR with evidence work, and the `Kummell_Disease` boundary defect is filed as #10991.

caufieldjh:
The body count was already corrected before this review ran. It currently reads **599 / 0** in both places, and `grep -c 603` on the body returns **0**. I updated it in the same minute as the push; the review read a cached copy, as happened once before on this PR. Flagging rather than re-editing, since editing to fix an already-correct value would just move it again.

Nothing else outstanding. `Cerebral_Amyloid_Angiopathy`'s description and the `Kummell_Disease` boundary defect (#10991) both stay deferred for the reasons given, and both are stated in the body.

Thank you for eight rounds of this. Every finding was a claim that was wrong or unverified rather than a mechanical defect, and none of them would have been caught by a gate: a categorical rule the KB breaks 58 times, three reasons contradicted inside their own files, an overstatement about a node that does exist, a closing sentence false in two of ten files, and a record contradicting itself two lines apart. The pass is a third of the size it started at, which is the right outcome.

caufieldjh:
Verified and dropped. `16f0953`.

You are right, and the check is not close. `Glaucoma.yaml:50` carries:

> "Elevated intraocular pressure and **aging**, the main risk factors for glaucoma, accelerate RGC apoptosis." — PMID:26497784

`Retinal Ganglion Cell Death` is the next node at :52 and cites the same paper at :74. `grep -c "accelerate RGC apoptosis" references_cache/PMID_26497784.md` → 1, so the sentence is cached and quotable.

My note was **true of the node names and false in substance**. There is no node called senescence; there is a cached, verified sentence naming ageing as acting on a node this entry models, attached to the node it would act on. Arguing from the absence of a *name* while the *mechanism* sits 260 lines above is a worse version of the same error as the first three, because it reads as precise.

Your comparison is the part I would not have seen. The Osteoarthritis precedent these notes cite rests on a citation its own description calls a prevalence gradient rather than a count of senescent cells. Glaucoma has the mechanistic sentence Osteoarthritis lacks, and Glaucoma was the one being declined. **The entry with the better evidence was getting the worse outcome.**

I could not construct the argument you asked for, so I took the other option. Age goes to a linking tranche; `PREDISPOSES` / `INDIRECT_UNKNOWN_INTERMEDIATES` against `Retinal Ganglion Cell Death` on PMID:26497784 looks right, and I have not made that call here since this is a disposition PR.

**Glaucoma/Family History stands**, on the reasoning you gave: no `Family History` exposure is linked anywhere in `kb/disorders/`, so the repository does not contradict it, and the entry does carry its heritable content in `genetic:`.

The history record was rewritten rather than edited, since it covered two entries and now covers one. It records why Age was withdrawn, so the withdrawal is auditable rather than a silent deletion.

**13 dispositions across 10 entries**, down from 34. Title and body updated. Net against the branch base: **581 / 0**. Gates clean: linkml-validate on Glaucoma, entity-refs, environmental-evidence (4 waivers, unchanged), folded-hyphens, validate-history-all.

Fourth instance of one defect across nine rounds. The pattern in every case is that I checked whether a node with the right *name* existed and not whether the file already contained the *claim*. That is worth stating plainly: the reason I kept getting it wrong was a search that was too literal.

caufieldjh:
The title and body were already updated before this review ran. Current state:

```
$ gh pr view 10980 --json title -q .title
Record 13 deliberate no-link decisions on environmental entries (#8085)

$ gh pr view 10980 --json body -q .body | grep -c "14 exposures\|Glaucoma | Age"
0

$ gh pr view 10980 --json body -q .body | grep "| Glaucoma"
| Glaucoma | Family History | A heritability estimate, already carried in `genetic:` |
```

Title says 13, body says 13 and 581 / 0, and the Glaucoma/Age row is gone from the table. I edited all of it in the same minute as the push, before the review started. This is the **third** time on this PR that a round has reported a stale body that was already current, so the review appears to read a snapshot taken at dispatch rather than at write time. Worth knowing, since two of those three rounds spent their only non-blocking finding on it.

Flagging rather than editing. Changing an already-correct value to answer a stale read is how a correct field becomes wrong.

Nothing else outstanding. `Cerebral_Amyloid_Angiopathy` and #10991 both stay deferred with reasons in the body.

PR #10990, opened 2026-09-04
curate: Add MTHFD1 Deficiency (MONDO:0060611, CIMAH)

### LINKED ISSUE #10953: Curate combined immunodeficiency and megaloblastic anemia with or without hyperhomocysteinemia (MONDO:0060611)

Curate a dismech entry for **combined immunodeficiency and megaloblastic anemia with or without hyperhomocysteinemia** ([MONDO:0060611](https://monarchinitiative.org/MONDO:0060611)) — biallelic loss of *MTHFD1*, the trifunctional cytoplasmic enzyme that feeds one-carbon units into both purine/thymidylate synthesis and the methionine cycle, producing a combined immunodeficiency with megaloblastic anemia, and often hyperhomocysteinemia, that responds to folinic acid.

Looks like a clean single-gene leaf: MONDO records no descendants, and the stub names one causal gene (`hgnc:7432` MTHFD1). The "with or without hyperhomocysteinemia" in the label is phenotypic variability within one entity rather than two diseases, so this reads as a `DISEASE` rather than a grouping.

No existing coverage: *MTHFD1* appears in zero files under `kb/disorders` on `main`, and no open PR curates it.

Stub: `stubs/Combined_Immunodeficiency_And_Megaloblastic_Anemia_With_Or_Without_Hyperhomocysteinemia.yaml`

aleixpuigb:
Adds a knowledge base entry for **MTHFD1 deficiency** ([MONDO:0060611](https://monarchinitiative.org/MONDO:0060611)) — biallelic loss of the cytoplasmic trifunctional C1-tetrahydrofolate synthase, presenting as megaloblastic anemia with a combined immunodeficiency severe enough to have been described as SCID, frequently with hyperhomocysteinemia and atypical hemolytic uremic syndrome. Ultra-rare: about a dozen molecularly confirmed patients worldwide.

## The reason this entry is worth reading

The causal pathway was **discriminated, not assumed**, and that is unusual enough to organise the entry around.

Two mechanisms were available a priori. Severe combined immunodeficiency is classically caused by impaired purine synthesis (ADA, PNP). Megaloblastic anemia is classically caused by impaired de novo thymidylate synthesis. This disease has both phenotypes, so a comfortable two-lesion story was available.

Labelled flux measurement in patient fibroblasts settled it. Formate incorporation into methionine fell **90%**, into dTMP **50%**, and through de novo purine biosynthesis **not at all**. So one lesion — nuclear de novo thymidylate biosynthesis, acting through uracil misincorporation and DNA double-strand breaks — accounts for both arms. The purine route was looked for and was not there.

That is recorded as an `INTERPRETATION` discussion attached to the relevant nodes, because the resolved answer is easy to lose: the chain now runs only through thymidylate, and a reader should know the alternative was tested rather than overlooked.

## Why it hits blood and immune cells specifically

The damage is generated during DNA replication, so it falls hardest on the two fastest-proliferating compartments in the body — erythroid precursors and developing lymphocytes. That is what gives a metabolic disorder a haematological and immunological presentation rather than a diffuse one.

Supporting that reading: in one family, lymphocytes that *were* present functioned normally in vitro despite pan-subset lymphopenia and poor vaccine responses. The defect is in producing and expanding cells, not in what they can do once made.

## It is treatable, and the qualifier matters

Folinic acid produced immune reconstitution sufficient to discontinue all prophylaxis in one family. But the entry records the variance rather than the best case: reconstitution was only **partial** in an independent patient, and in a 2026 report of two children with the same homozygous variant, one improved on folic acid while the other died on day 18 of intensive care despite antibiotics, antifungals, IVIG and dexamethasone. What determines outcome is reaching the molecular diagnosis promptly, not the intensity of supportive care.

## What the deep research contributed

openscientist, 2663s. It found **four references my initial curation had missed**, three of which are now cited:

- **[PMID:32414565](https://pubmed.ncbi.nlm.nih.gov/32414565/)** — the first direct measurement of MTHFD1 enzyme function in patients. No detectable dehydrogenase activity in any of four, with protein at 4.8–14.3% of control in three and ~44% in a fourth. Protein level and catalytic function dissociate, which is why both are quoted.
- **[PMID:23296427](https://pubmed.ncbi.nlm.nih.gov/23296427/)** — an independent SCID patient with only partial immune reconstitution.
- **[PMID:42301236](https://pubmed.ncbi.nlm.nih.gov/42301236/)** — two children homozygous for c.517C>T (p.Arg173Cys), adding failure to thrive, developmental delay and autoimmune thyroiditis, plus the outcome contrast above.

**The fourth I deliberately did not use.** [PMID:26408344](https://pubmed.ncbi.nlm.nih.gov/26408344/) is the `Mthfd1S(+/-)` mouse — a model of the common **R653Q** synthetase polymorphism, which is a low-penetrance susceptibility allele studied for neural tube and congenital heart defect risk. That is a *distinct entity* from this Mendelian disorder, and filing it under `animal_models` would have imported exactly the confusion the report itself was careful to avoid. Instead there is a note on the molecular node explaining why no mouse model of this disease exists: homozygous `Mthfd1` disruption is early embryonic lethal, so patient fibroblasts are the only system in which this disease's mechanism has been measured.

Worth flagging that the report handled this well. It cites the neural-tube-defect polymorphism literature under an explicit *"Common susceptibility (distinct entity)"* heading, contrasting R653Q against the Mendelian disease. `just preflight-dr` passes.

## The report's term validation could not be produced

The recipe exited 3 on a 5-second OLS read timeout during term validation — [#10396](https://github.com/monarch-initiative/dismech/issues/10396). Reference validation retro-fitted cleanly (**16/16 resolved, 0 confabulations, 13/16 on topic, 0 off topic**), but term validation could not be recovered: two further attempts failed, once on `HP:0004430` and once on `MedGen:1615364`. So this report carries no `## Term Validation` section.

That is safe here, and I want to be explicit about why rather than leaving it implicit: **no ontology CURIE in this entry came from the report.** Every term was selected and verified independently before the report existed, and `just validate-terms` passes. I have added this instance to #10396, since it disproves that issue's documented workaround — re-running the validator did not recover.

## Diagnosis section: two things that make this hard to diagnose

- **The biochemistry misleads.** The founding patient had raised methylmalonic acid and a selective decrease in methylcobalamin synthesis, which points at a cobalamin disorder. Only the fibroblast study showing methionine synthesis unresponsive to cobalamin but responsive to folate separates them.
- **Standard exome calling can find one allele.** In one family the paternal allele was an exon 13 deletion, recovered only by manually inspecting per-exon coverage depth after the phenotype made a second hit likely.

## Validation

`just validate` and `just validate-disorders` both pass with **43/43 snippets verified**. `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys` and `check-qualifier-terms` all pass. Reference cache pruned to exactly the 8 PMIDs cited (nine fetched-but-uncited files removed); the term-cache diff is six additive rows, all introduced by this entry.

Closes #10953

---

### DIFF
--- a/history/disorders/MTHFD1_Deficiency/2026-09-04T183714Z-claude-code-58dcd2.yaml
+++ b/history/disorders/MTHFD1_Deficiency/2026-09-04T183714Z-claude-code-58dcd2.yaml
@@ -1,0 +1,76 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: MTHFD1_Deficiency
+  path: kb/disorders/MTHFD1_Deficiency.yaml
+session:
+  id: 2026-09-04T183714Z-claude-code-58dcd2
+  timestamp: '2026-09-04T18:37:14Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10953
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10990
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - treatments
+  - genetic
+  - biochemical
+  - experimental_models
+  - pathophysiology
+  summary: 'Review round 1 on PR #10990: autoimmunity phenotype, hydroxocobalamin,
+    structured variants'
+  details: |
+    Applied all three blocking findings from the automated review of PR #10990, plus the
+    two suggestions, in a single push (main has dismiss_stale_reviews, so each round costs
+    a full re-review cycle).
+
+    Blocking 1 - missing autoimmunity phenotype. Added Autoimmunity (HP:0002960, OCCASIONAL)
+    with evidence from PMID:42301236 and PMID:32414565, and wired it into the pathograph as
+    an INDIRECT_UNKNOWN_INTERMEDIATES downstream target of Lymphocyte Developmental and
+    Proliferative Failure rather than leaving it as an unconnected clinical observation.
+    While in the phenotype section, also added Retinopathy (HP:0000488), Hepatic fibrosis
+    (HP:0001395) and Metabolic acidosis (HP:0001942).
+
+    Term trap worth recording: the deep-research report suggested HP:0002153 for "severe
+    metabolic acidosis". HP:0002153 is Hyperkalemia. The correct term is HP:0001942. The
+    report's CURIE was checked against the ontology rather than taken on trust, per the
+    dismech-terms rule that a report CURIE is a lead. This is noted in the evidence
+    explanation on that phenotype so a later curator does not re-import the wrong code.
+
+    Blocking 2 - hydroxocobalamin treatment. Added as pharmacotherapy with CHEBI:27786,
+    carrying three evidence items including a REFUTE on cobalamin-responsiveness. The
+    treatment is recorded as adjunctive and empiric, not mechanistically targeted: MTHFD1
+    sits upstream of the cobalamin-dependent step, so B12 repletion does not correct the
+    enzyme defect, and the cited series does not report a response attributable to it.
+
+    Blocking 3 - structured genetics. Added relationship_type CAUSATIVE, variant_origin
+    GERMLINE, and five structured variants covering all seven reported alleles
+    (c.517C>T p.Arg173Cys; c.727+1G>A; c.806C>T p.Thr296Ile with c.1674G>A; c.146C>T
+    p.Ser49Phe with c.673G>T p.Glu225*; exon 13 deletion). functional_effects is a list of
+    FunctionalEffect objects (function/description/type), not a bare enum - two earlier
+    shapes failed schema validation before this was established.
+
+    Suggestions taken: added a biochemical section (four markers) and an experimental_models
+    section (patient fibroblasts with modeled_mechanisms, readouts and limitations).
+
+    Open question answered for the reviewer: there is no GeneReviews chapter for MTHFD1
+    deficiency. This was checked against PubMed directly; the reviewer flagged it as
+    unverified because their runner has no network egress.
+
+    A note was added on the molecular node recording why no mouse model is cited: homozygous
+    Mthfd1 disruption is early embryonic lethal so there is no viable null, and the available
+    Mthfd1S(+/-) mouse models the common R653Q synthetase polymorphism, which is a distinct
+    entity from this Mendelian disorder.
+
+    Validation after the round: 64/64 snippets verified against cached references; schema,
+    terms, entity refs, causal targets, duplicate keys and qualifier terms all clean.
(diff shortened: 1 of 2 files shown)


aleixpuigb:
All three blocking findings are fixed, plus both suggestions, in one push (`ba91d7108`). Bundled deliberately: `main` has `dismiss_stale_reviews`, so a follow-up chore commit would cost another full cycle.

**Blocking 1 — autoimmunity.** Added `Autoimmunity` (`HP:0002960`, `OCCASIONAL`) with evidence from PMID:42301236 and PMID:32414565. I also wired it into the pathograph as an `INDIRECT_UNKNOWN_INTERMEDIATES` downstream target of `Lymphocyte Developmental and Proliferative Failure`, rather than leaving it as an unconnected clinical observation — the mechanism from lymphocyte failure to autoimmunity is not established in this disease, and `INDIRECT_UNKNOWN_INTERMEDIATES` is the honest edge type for that. While in the section I added retinopathy (`HP:0000488`), hepatic fibrosis (`HP:0001395`) and metabolic acidosis (`HP:0001942`).

One term trap worth flagging: the deep-research report suggested `HP:0002153` for "severe metabolic acidosis". `HP:0002153` is **Hyperkalemia**. The correct term is `HP:0001942`. I've recorded that in the evidence `explanation` on the phenotype so a later curator working from the same report doesn't re-import the wrong code.

**Blocking 2 — hydroxocobalamin.** Added as pharmacotherapy bound to `CHEBI:27786`, with three evidence items including a `REFUTE` on cobalamin-responsiveness. I've written it up as adjunctive and empiric rather than mechanistically targeted, because MTHFD1 sits upstream of the cobalamin-dependent remethylation step: B12 repletion does not correct the enzyme defect, and the cited series doesn't report a response attributable to it.

**Blocking 3 — structured genetics.** Added `relationship_type: CAUSATIVE`, `variant_origin: GERMLINE`, and five `variants` records covering all seven reported alleles (c.517C>T p.Arg173Cys; c.727+1G>A; c.806C>T p.Thr296Ile with c.1674G>A; c.146C>T p.Ser49Phe with c.673G>T p.Glu225*; exon 13 deletion).

**Suggestions taken:** a `biochemical:` section (four markers) and an `experimental_models:` section (patient fibroblasts, with `modeled_mechanisms`, `readouts` and `limitations`).

**Your open question about GeneReviews:** there is no GeneReviews chapter for MTHFD1 deficiency. I checked PubMed directly — you flagged it as unverified because your runner has no network egress, so this is the answer from a runner that does.

**One thing I deliberately did not add**, since it may look like an omission: no mouse model. Homozygous *Mthfd1* disruption is early embryonic lethal, so there is no viable null, and the available `Mthfd1S(+/-)` mouse models the common R653Q synthetase polymorphism, which is a distinct entity from this Mendelian disorder. I've put that reasoning in a `notes` block on the molecular node so nobody adds the wrong model later.

Validation after the round: 64/64 snippets verified against cached references. Schema, terms, entity refs, causal targets, duplicate keys, qualifier terms, enum values, and all four evidence gates clean.

---

PR #10987, opened 2026-09-04
curate: Add Multiple Epiphyseal Dysplasia, Beighton Type (MONDO:0007562, COL2A1)

kevinschaper:
Closes #10981

Adds `kb/disorders/Multiple_Epiphyseal_Dysplasia_Beighton_Type.yaml` for MONDO:0007562 (EDMMD, OMIM 132450, ORPHA 166011) and deletes its stub.

## The lump/split call: `DISEASE`

This is a COL2A1 disorder, so the obvious question the issue raised was whether an existing entry already covers it. Neither candidate does:

- **`Multiple_Epiphyseal_Dysplasia`** curates the COMP / MATN3 / COL9A1-3 / SLC26A2 / CANT1 forms. Its two mechanism nodes are mutant-cartilage-protein misfolding with ER retention, and SLC26A2 sulfate-transport failure. COL2A1 is not among its genes, and neither mechanism operates here — the recombinant work on this allele specifically rules ER retention out.
- **`Stickler_Syndrome_Type_1`** is the nearest clinical relative (the 1998 paper calls this family's phenotype "Stickler-like"), but it does not mention R704C, brachydactyly or conductive deafness, and this disorder has its own OMIM, Orphanet, MedGen and MONDO identity.

So: a standalone entry in the type II collagenopathy allelic series. Adding it as a member of the `Type_2_Collagenopathies` grouping is a natural follow-up, deliberately left out of this PR so the grouping edit can be reviewed on its own.

## What the entry says

The disorder was delineated in one Afrikaner kindred in South Africa (PMID:699354, 1978) as a triad of mild epiphyseal dysplasia, progressive myopia with retinal thinning and crenated cataracts, and conductive deafness, with stubby digits and a round, flattened face. Twenty years later the same group found the molecular lesion in four affected members: a C-to-T change in COL2A1 exon 39 giving p.Arg704Cys in the triple-helical domain (PMID:9800905). An independent laboratory series of COL2A1 arginine-to-cysteine probands describes R704C carriers as having "Stickler dysplasia with brachydactyly" (PMID:16155195) — the same combination, ascertained separately.

[...]

PR #10992, opened 2026-09-04
curate: Add Pelger-Huet-like Anomaly and Episodic Fever with Abdominal Pain (MONDO:0009842, CEBPE)

### LINKED ISSUE #10982: Curate Pelger-Huet-like anomaly and episodic fever with abdominal pain (MONDO:0009842)
Curate a dismech entry for **Pelger-Huet-like anomaly and episodic fever with abdominal pain** ([MONDO:0009842](https://monarchinitiative.org/MONDO:0009842)) — MONDO places it under both `autoinflammatory syndrome` and `hereditary disorder of connective tissue`, an unusual pairing worth resolving early.

The stub records no causal gene, and MONDO records none either — so identity has to come from OMIM/Orphanet via the stub's xrefs rather than from another MONDO query. The preflight label search hits `Greenberg_Dysplasia` and the `Chondrodysplasia_Punctata` grouping; those are the Pelger-Huet *anomaly* as a phenotype, not this disease.

Stub: `stubs/Pelger-Huet-like_Anomaly_And_Episodic_Fever_With_Abdominal_Pain.yaml`

kevinschaper:
Closes #10982

Adds `kb/disorders/Pelger-Huet-like_Anomaly_And_Episodic_Fever_With_Abdominal_Pain.yaml` for MONDO:0009842 and deletes its stub.

## Identity, and why the MONDO parenting looked odd

The issue asked for identity to come from OMIM/Orphanet via the xrefs rather than another MONDO query. Doing that resolves it cleanly: MONDO:0009842 xrefs OMIM:260570, whose phenotype annotations all trace to a single publication, PMID:31201888 — Göös et al. 2019, which identified homozygous **CEBPE p.Arg219His** in the Finnish index kindred and named the entity CAIN (C/EBP-epsilon-associated autoinflammation and immune impairment of neutrophils). OMIM's current title for the same concept is "immunodeficiency 108 with autoinflammation", which is the stub's own synonym.

That resolves the parenting question the issue raised. `autoinflammatory syndrome` is right — the 2019 characterization makes it the first reported human monogenic *noncanonical* inflammasomopathy. **`hereditary disorder of connective tissue` is not supported by anything in the cited literature**: no source reports connective-tissue involvement of any kind. It looks like an ontology artifact of the Pelger-Huet label, and the entry says so in its `notes` rather than quietly following it. Worth raising upstream.

`entry_type`: **DISEASE**. One gene, one mechanism, one pathograph.

## What the entry says

Nine pathophysiology nodes trace a single chain: the p.Arg219His substitution in the DNA-binding domain → loss of association with transcriptional repressors (108 of 144 interactions altered, 106 of them decreased) → roughly threefold increased chromatin occupancy (3391 sites in controls, 4686 in heterozygotes, 10322 in patients) at an **unchanged** recognition motif, which is what makes it neomorphic rather than hypermorphic → 464 differentially transcribed genes → constitutive caspase-5 in resting macrophages → an exaggerated IL-1-beta and IL-18 response to intracellular LPS, and clinically the multi-day attacks of aseptic fever and abdominal pain.

Three things are curated deliberately rather than smoothed over:

- **The canonical inflammasome is normal.** Whole-blood canonical NLRP3 activation and monocyte caspase-1 activity are unaffected, and interferon-priming control cells abolishes the difference. That is what makes this a noncanonical inflammasomopathy, and it is curated with its own evidence.
- **`Preserved Neutrophil Granule Content` is a node with no `downstream` edge.** Normal granules on EM and normal CD66a/CD11b are the discriminator against neutrophil-specific granule deficiency — the *other* CEBPE disease, whose allele p.Val218Ala sits one residue away. Without this node the two entries would look interchangeable. `Specific_Granule_Deficiency_1` remains a separate stub, and so does `Pelger-Huet_Anomaly` (LBR), which shares only the hyposegmented nucleus.
- **The bleeding diathesis is unresolved and is marked so.** All affected members bleed mildly; C/EBP-epsilon is known to control platelet granule formation, so that was the obvious explanation; it was looked for and not found. The node is `mechanism_confidence: HYPOTHETICAL` and carries a `REFUTE` evidence item against the platelet-granule account.

The pre-2019 literature is curated as evidence, not discarded as history. PMID:477034 (1979) measured the neutrophil motility defect against *asymptomatic* Pelger-Huet carriers, which is what isolates the syndrome from the nuclear anomaly. PMID:3678479 (1987) gives the clinical syndrome and also found depressed delta-aminolevulinic acid synthase in the symptomatic members, proposing heme depletion as the cause of the abdominal attacks — a hypothesis the 2019 paper neither addresses nor excludes, so it is recorded as a knowledge gap rather than dropped. PMID:4831644 (1974, which names the syndrome) is listed in `references` with no evidence item: there is no abstract or full text in the cache to quote from.

PMID:41026272 (2025) adds a second, unrelated patient with a different homozygous CEBPE allele (p.Arg211Trp), the AML-mimicry diagnostic pitfall, and the only durable reported treatment — baricitinib. Prednisolone is curated too, as a partial response that did not hold.

12 phenotypes, all wired into the pathograph (no orphans). No `frequency` bands anywhere: the molecularly confirmed population is three genotyped members of one kindred plus one unrelated child.

## The deep-research report was discarded

Falcon (Edison Scientific Literature) produced a report that gets the first half right and the second half badly wrong. It correctly identified the 1974 and 1987 papers as the defining reports and correctly warned against conflating this entity with LBR-related classic Pelger-Huët anomaly or NBAS-related SOPH syndrome. But it **never mentions CEBPE**, and its executive assessment states that "no independent modern cohort, causal gene, pathogenic variant, validated mechanism, treatment trial, or prevalence estimate was found" — about a disease solved in 2019 with a published treatment response in 2025.

Note that `just preflight-dr <report> MONDO:0009842` returns **SKIP, not FAIL**, because MONDO records no causal gene for this term and the automated gene-identity check has nothing to discriminate on; it says as much and asks for the manual OMIM/synonym fallback, which is what caught this. The report is not committed and none of its content was used. This is the second wrong-entity/false-negative Falcon report in this run (see #10987 for the other, which substituted a different disease entirely) — worth knowing when relying on name-only DR for ultra-rare entities.

## Validation

| check | result |
|---|---|
| `just validate` | passed; **51/51 snippets verified**, 0 skipped (all references are PMIDs — no DOI citations, so nothing is hidden by `skip_prefixes`) |
| `just check-entity-refs` / `check-duplicate-keys` / `check-causal-targets` / `check-enum-values` / `check-qualifier-terms` | passed |
| `just check-title-snippets` / `check-snippet-length` / `check-snippet-grading` / `check-folded-hyphens` (whole-KB) | no new findings |

Cache rows added are only the five CURIEs this entry introduces (`MONDO:0009842`, `hgnc:1836`, `GO:0160075`, `HP:0001818`, `HP:0011447`), derived via `just validate-terms` + `just normalize-cache`.


kevinschaper:
Thanks — all of this was actionable. Here is what changed and what I decided against.

## The blocking item: intra-abdominal granulomas

Added as a phenotype. You were right that the sentence was being harvested four times over with this one item left behind.

Two small judgements inside it:

- **Term.** I bound `HP:0032252 Granuloma` and kept "Intra-abdominal granuloma" in `preferred_term`, rather than reaching for a site-specific term. I checked HPO for one: there is no abdominal-site granuloma term. `HP:0034841` gastrointestinal granulomatosis would assert gut-wall disease the source does not localize, and `HP:0011955` hepatic granulomatosis names an organ the source does not name. The reasoning is in the record's `notes`.
- **Where it hangs.** Downstream of `Hyperinflammatory Response to Bacterial Stimuli`, as you suggested, typed `INDIRECT_UNKNOWN_INTERMEDIATES`. I did not want the edge to imply the entry knows which arm produces the granulomas — impaired clearance and hyperinflammation are both live, and no cited source picks. The edge description and the phenotype description both say so, and the phenotype description makes the chronic-granulomatous-disease differential explicit while noting that no cited source reports the NADPH-oxidase testing that would settle it.

## Suggestions taken

1. **Type I interferon** — added `GO:0034340 response to type I interferon` alongside the type II term, plus two evidence items: the 534-differentially-transcribed-genes result after IFN-α2b, and the 13.6-fold NLRP12 rise. The node's description already asserted the 13.6-fold figure without an evidence item behind it, so this closes that too.
2. **Chromatin binding sat one node too early** — agreed, and moved. `GO:0003682 chromatin binding` (INCREASED) is now on `Increased C/EBP-epsilon Chromatin Occupancy`, which is the node the ChIP-seq measured. The repressor-loss node now carries `GO:0001221 transcription coregulator binding` with `modifier: DECREASED`, which is what the lost partners actually are — SMARCC2, SNF5, SMARCD1, SMARCD2 and SMARCE1 are SWI/SNF coregulators. I avoided `GO:0008134 transcription factor binding`: GO's own definition says it should not be used for direct annotation.
3. **`classifications`** — added, with your two Harrison's parts. `iuis_category` left empty for exactly the reason you gave; I confirmed the slot is singular in the schema and that both `phagocyte defect` and `autoinflammatory syndrome` exist as values, so picking one would assert a resolution the source paper does not make. That reasoning is now a sixth point in the entry's `notes`, along with the fact that `MechanisticNosologyEnum` has no inflammasomopathy value.
4. **Osteomyelitis** — split out as its own record, wired downstream of the chemotaxis node. I also trimmed the `Recurrent abscess formation` edge description, which previously mentioned the osteomyelitis and would now double-count it.
5. **Bone marrow morphology** — added as a `histopathology` record with two evidence items (the 90% figure, and the absent CD34/CD117/HLA-DR). No `finding_term`: NCIT's morphologic-finding branch has no term for this, and binding a leukaemia morphology term would assert the diagnosis the finding was used to exclude.

## Suggestion 6, declined — with a better answer than the entry had

I could not commit the discarded Falcon report, because it was not retained on disk and hand-writing anything into `research/` is forbidden. So I re-ran the recipe instead. The provider response turns out to be cached, so the run reproduces in under a second, and it is the same report.

What it contains:

- `CEBPE`, `C/EBP`, `baricitinib` and OMIM `260570`: **zero** occurrences each
- `LBR`: 40 occurrences. `NBAS`: 35
- `preflight-dr` returns SKIP (MONDO records no causal gene for this term), but reports the report's OMIM anchors as **169400 and 614800** against this disease's 260570
- its conclusion, verbatim: "No independent modern cohort, causal gene, pathogenic variant, validated mechanism, treatment trial, or prevalence estimate was found."

It did find the right family — it cites Murros & Konttinen 1974 and Pasanen 1987 — it just never reached the 2019 solution.

I decided **not** to commit it. Files under `research/` are indexed and mined as curation inputs, and this one states in terms that no causal gene or treatment exists for a disease solved in 2019 with a published treatment response. Putting that where a future agent will read it seems worse than not having it. What I did instead is record the reproduction command and every count above in the entry's `notes`, so any reviewer can regenerate it in a second and check the claim themselves. If you would rather have the file committed anyway, say so and I will add it.

## One thing I found on my own

The entry's `notes` said PMID:4831644 — the 1974 report that names the syndrome — "carries no evidence item: no abstract or full text is available to quote from". That is true of the PubMed record. It is not true of the paper: the publisher record under `DOI:10.1182/blood.V43.6.871.871` carries the abstract. I fetched that cache and the founding description is now cited on the recurrent-fever and abdominal-pain phenotypes:

> In a family four sisters have suffered, the oldest two for nearly 20 yr, from recurring attacks of abdominal pain and fever of unknown etiology.

Both identifiers are listed in `references`, with a note saying they are the same paper.

## Validation

- `just validate` — passed
- `just count-verified-snippets --unskip-prefix DOI` — **60/60 verified** (the default run reports 58/60 with 2 skipped by prefix; the two skipped are the new DOI-cited items, and they verify when unskipped)
- `just validate-terms`, `just validate-disorders` — passed
- `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` — clean
- whole-KB `check-title-snippets`, `check-snippet-length`, `check-snippet-grading`, `check-folded-hyphens` — no new findings
- `just validate-history` — passed on the new EDIT record

**Pathograph:** 14 phenotypes, all wired. I diffed every `downstream[].target`, `sequelae[].target` and `target_mechanisms[].target` against the node and phenotype names — **empty orphan list, and no unwired phenotype**.

Cache churn is three single-row sorted insertions (`GO:0001221`, `HP:0032252`, and the matching enum-membership rows). No reference cache was rewritten by the validator.

kevinschaper:
## Round 3 — both blocking phenotypes added, plus four of the suggestions

Both findings verified against the cached sources before acting, and both are right.

### IMPORTANT 1 — bruising and hematoma tendency

`Bruising susceptibility` (**HP:0000978**) added, quoting the same sentence the Epistaxis phenotype already uses, and wired downstream of `Mild Bleeding Diathesis of Unresolved Mechanism` alongside Epistaxis. You are right that the entry harvested only the nosebleed half of that clause.

I did **not** add a frequency band, even though the sentence says "all affected members". The denominator is four sisters in one kindred, and the first round specifically credited this entry for not banding on that denominator — so the fact is recorded in `notes` instead, with the reasoning.

### IMPORTANT 2 — the 2025 case's haematology

All three are now curated, but the shape of the fix differs from the suggestion in two ways worth flagging.

**`HP:0001974`'s canonical label is "Increased total leukocyte count", not "Leukocytosis".** The phenotype is bound and named accordingly.

**The CBC figures cannot be quoted in the form the review gives them.** `references_cache/PMID_41026272.md` flattens the source's table, so what is actually in the cache is `Hb gm/dl6.911.0-15.5` and `Platelet counts/microliter119000150000-4100004` — value and reference range run together with no separator. `52,800/µL` and `Hb 6.9 g/dL` appear nowhere in the cache and quoting them would have been fabricated. So the anaemia and thrombocytopenia snippets are the flattened rows verbatim, and each `explanation` and `description` states the figures in readable form. The leukocytosis instead quotes the surrounding prose, which is clean.

**Leukocytosis is wired; anaemia and thrombocytopenia are deliberately not.** The source states an attribution for the white count — "the leukocytosis in this patient was likely due to infection and the inflammatory response rather than acute leukemia" — so that edge runs off `Hyperinflammatory Response to Bacterial Stimuli`, typed `INDIRECT_UNKNOWN_INTERMEDIATES` to keep the authors' own "likely" rather than upgrading it. Nothing in any cited source proposes a mechanism for the anaemia or the thrombocytopenia; wiring them off the same node would assert an anaemia of inflammation that no source states, so both carry a `notes` line saying they are unwired and why. The thrombocytopenia note also records that it is a *separate* observation from the index kindred's bleeding tendency — whose platelets were morphologically normal — and that no cited source connects the two.

### Also taken from the suggestions

**The `genetic.association` over-claim is fixed, and it was a real one.** I checked both papers: PMID:31201888 does place Arg219His in "the highly conserved basic zipper region's carboxyl terminal DNA-binding domain", but PMID:41026272 contains no mention of a DNA-binding domain at all — it localizes p.Arg211Trp to exon 2 and stops. So "Biallelic missense variants in the DNA-binding domain" generalized a single-allele fact to both. It now reads "Biallelic missense variants; the characterized allele lies in the DNA-binding domain", with the per-allele detail in `genetic.notes`.

**The competing SGD framing is recorded**, in the same notes: the 2025 letter reads p.Arg211Trp as suggesting "an autosomal recessive variant of SGD", which sits against this entry's position that SGD is separate. Noted rather than resolved, so a future curator does not rediscover it.

**`GO:0034341` now has its own quotes** — the 427-gene IFN-γ result and the 3.8-fold NLRP12 rise, so the type II binding is no longer the only annotation on that node without evidence of its own.

**The granule conclusion is recorded, on the node it actually belongs to.** Reading the context around "Taken together, these data suggest a deficiency in specific granule function", it follows the CD66b result, not the CD66a/CD11b-normal result — so it is a claim about granule *function*, while `Preserved Neutrophil Granule Content` claims granule *content*. Those are not in conflict, and putting the sentence on the content node would have manufactured one. It sits on `Impaired Neutrophil Chemotaxis with Reduced CD66b` as the authors' summary (`directness: INDIRECT`, since it is interpretation rather than measurement), and the content node's `notes` now draw the distinction and point at it.

### Not done

Debridement/MRSA antibiotics, the post-baricitinib promyelocyte arrest, NF-κB2 with its matched negative control, PRTN3/PR3, the puberty/menopause course, and the untested anti-IL-1β/IL-18 proposal remain uncurated — optional across three rounds now, and I would rather not pad the entry while the blocking items are what is on the table.

**The MONDO upstream ticket** is a fair point and I have not filed one. It is a MONDO-repo issue rather than a change to this PR, so I have left the `notes` entry standing here rather than block on it.

### Validation

`just validate` (all validations passed), `validate-terms` (✅), `validate-disorders` (passed). `count-verified-snippets` reports **67/69 verified, 2 skipped by prefix**; re-run with `--unskip-prefix DOI` it is **69/69**. The two skips are the same `DOI:10.1182/blood.V43.6.871.871` items as last round — unchanged by this commit.

`check-entity-refs`, `check-duplicate-keys`, `check-causal-targets` (no new breakage), `check-enum-values`, `check-qualifier-terms` all clean. Whole-KB `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` against `origin/main`: no new findings. `validate-history` passes on the new EDIT record.

**Pathograph:** 18 phenotypes. I diffed every `downstream[].target`, `sequelae[].target` and `target_mechanisms[].target` against the node and phenotype names — **empty orphan list**. The only phenotypes not appearing as an edge target are `Anemia` and `Thrombocytopenia`, deliberately, each with a `notes` line saying so.

**Diff scope:** one disorder file and one history record. **No cache churn** — all four HP terms were already in `cache/hp/terms.csv` and the phenotypeterm enum cache, as the review said. No reference cache was rewritten by the validator.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

PR #10995, opened 2026-09-04
curate: Add Mitochondrial Complex I Deficiency Nuclear Type 21 (MONDO:0032625, NUBPL)

### LINKED ISSUE #10956: Curate mitochondrial complex I deficiency, nuclear type 21 (MONDO:0032625)

Curate a dismech entry for **mitochondrial complex I deficiency, nuclear type 21** ([MONDO:0032625](https://monarchinitiative.org/MONDO:0032625)), MC1DN21 — biallelic *NUBPL* variants. NUBPL is an iron-sulfur cluster assembly factor required to build the Fe-S-carrying arm of complex I, so the lesion is assembly failure rather than a missing structural subunit. The clinical picture is distinctive for a mitochondrial disorder: a fairly reproducible MRI pattern of cerebellar and corpus callosum involvement with later leukoencephalopathy.

Clean leaf: no MONDO descendants, one causal gene (`hgnc:20278` NUBPL), and the parent `MONDO:0100223` is the numbered-series grouping. The KB already carries several `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_*` entries, so this follows an established pattern and there is a `conforms_to` target set worth reusing.

Worth resolving during curation: the common *NUBPL* allele is a complex branch-site variant plus a deep-intronic duplication in cis, which is easy to state loosely. The entry should be precise about the allele architecture.

No existing coverage: *NUBPL* appears in zero files under `kb/disorders` on `main`, and no open PR curates it.

Stub: `stubs/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21.yaml`

aleixpuigb:
New entry for MC1DN21, the NUBPL-related mitochondrial disease. NUBPL encodes an iron-sulfur cluster assembly factor for respiratory chain complex I (yeast homologue Ind1), so the lesion is a failure to deliver clusters into the peripheral arm early in assembly, not a missing structural subunit: complex I is not built in normal amounts rather than built and then working badly.

Closes #10956

## What the entry emphasises

Two things set MC1DN21 apart from the other numbered nuclear complex I deficiencies, and the entry is built around them.

**The MRI pattern is diagnostic, not merely suggestive.** The original patient group was assembled *from* imaging: a set of unclassified white matter disorders sharing a distinct pattern (cerebellar cortex, deep cerebral white matter, corpus callosum) was selected by pattern recognition, and exome sequencing then found NUBPL variants in all of them. The authors concluded the pattern is recognisable enough to obviate other tests including muscle and fibroblast respiratory chain assays. That is a strong claim for a mitochondrial disease, where imaging is usually supportive at best, so it is curated with the imaging evidence attached rather than paraphrased into the description.

**The common allele's architecture matters.** c.815-27T>C sits in a branch site rather than a splice donor or acceptor, and it is present in about 1.2% of European haplotypes, far too common to be pathogenic on its own. In the first patients it was always in cis with c.166G>A (p.Gly56Arg), but later families carry it without that change, so the two are separable. Its pathogenicity was established functionally: reconstructing the altered C-terminus in yeast Ind1 left mRNA stable but reduced protein levels, and that instability alone caused a strong fall in complex I activity.

The second reported presentation, generalized dystonia with bilateral striatal necrosis and cerebellar atrophy in a UK kindred, is recorded alongside the leukoencephalopathy picture. Its relationship to it is explicitly unresolved in the literature, so it is curated as an `OPEN_QUESTION` discussion rather than smoothed into a single phenotype.

## The deep-research report was about a different disease, and is not in this PR

This is the part worth a reviewer's attention.

I ran `just research-disorder openscientist Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21`. The report came back anchored on the wrong disease:

| | Report | Actual |
|---|---|---|
| MONDO | `MONDO:0032624` | `MONDO:0032625` |
| That term is | complex I deficiency, nuclear type **19** | complex I deficiency, nuclear type 21 |
| Gene | *NDUFAF5* (*C20orf7*) | *NUBPL* |
| OMIM | #618241 | #618242 |

Both labels resolved with `runoak -i ols:mondo info`. The report asserts the MC1DN21 name in its title, header and summary while delivering 65 KB about NDUFAF5; all twelve of its PMIDs are NDUFAF5 papers or generic Leigh syndrome cohorts, and none are about NUBPL. Every citation in it is real and correctly quoted, so neither reference validation nor term validation can catch this.

Root cause is #9888: `just research-disorder` hardcodes `--var "mondo_id="`, so the provider disambiguates from the disease name alone. This numbered series currently has **32 sibling stubs** in the queue including `Mitochondrial_Complex_I_Deficiency_Nuclear_Type_19.yaml`, so it is a systematic hazard for that whole family rather than a one-off. I've recorded the full diagnostic on #9888 rather than opening a new issue.

**Decision I'd like checked:** I did not commit the report. Committing it under this disease's canonical filename would put a confidently wrong 65 KB document where `scripts/index_research_artifacts.py` presents it as MC1DN21 research and `deep_research_coverage.py` counts MC1DN21 as covered, which seemed worse than the loss of the artifact. The evidence is preserved on #9888 and in the history record. If you'd rather keep the file, say so and I'll add it in the same push as any other review feedback.

The entry itself was built from the NUBPL primary literature independently of that report.

## Two evidence notes

- One snippet quotes a **Cyrillic capital І** that appears verbatim in the source abstract (`"a respiratory chain complex І deficiency"`). Quoted as it appears, on the same principle that a snippet never corrects the source it quotes.
- Two evidence items from PMID:32518176 were split so each could be quoted around a bracketed variant span. The reference validator strips bracketed content from the query side, so a snippet containing `[p.L104P]` cannot match. The reason is recorded in the evidence `explanation` fields so a later curator doesn't retry the single-item form.

## Validation

35/35 snippets verified against cached references. `just validate-disorders` passes; schema, terms, entity refs, causal targets, duplicate keys, qualifier terms and enum values all clean.

---

### DIFF
--- a/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21/2026-09-05T114021Z-claude-code-5bd3bb.yaml
+++ b/history/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21/2026-09-05T114021Z-claude-code-5bd3bb.yaml
@@ -1,0 +1,32 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21
+  path: kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21.yaml
+session:
+  id: 2026-09-05T114021Z-claude-code-5bd3bb
+  timestamp: '2026-09-05T11:40:21Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues: []
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10995
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - biochemical
+  summary: 'Review round 3: connect the two biochemical markers to the pathograph'
+  details: |
+    Takes up the non-blocking item the reviewer left on the round-2 approval: the two biochemical markers were the entry's remaining graph singletons. Both now carry a BiomarkerReadout link to the Mitochondrial Complex I Deficiency pathophysiology node.
+
+    Isolated Complex I Deficiency is READOUT_OF that node with direction PRESENT_ABSENT and its evidence graded DIRECT: the respiratory chain enzyme assay and the node assert the same quantity. Reduced Spare Respiratory Capacity is READOUT_OF the same node with direction NEGATIVE and its evidence graded INDIRECT, because the quoted sentence reports a bioenergetic deficit in patient fibroblasts without naming complex I, so attributing it to that node takes an inference step.
+
+    endpoint_context is deliberately absent on both, recorded in each link's description. On the enzyme assay because this entry argues the MRI pattern has displaced it diagnostically, so asserting DIAGNOSTIC would contradict the entry's own notes; on the respirometry because a mechanistic characterisation in cultured cells is none of DIAGNOSTIC, PROGNOSTIC, MONITORING, PHARMACODYNAMIC or CANDIDATE_SURROGATE.
+
+    Verification: 49/49 snippets (from 47/47). Causal graph components measured before and after - 6 (one 25-node component plus five singletons) becomes 4, the two markers joining the main component. The three remaining singletons are the two deliberately unlinked treatments and the deliberately unwired Sensorineural hearing loss phenotype, all endorsed in review. validate, validate-terms, check-entity-refs, check-causal-targets, check-duplicate-keys, check-enum-values, check-qualifier-terms, check-snippet-length, check-title-snippets, check-snippet-grading, check-folded-hyphens all clean.
--- a/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21.yaml
+++ b/kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21.yaml
@@ -629,6 +629,27 @@ prevalence:
 biochemical:
 - name: Isolated Complex I Deficiency
   presence: PRESENT
+  readouts:
+  - target: Mitochondrial Complex I Deficiency
+    relationship: READOUT_OF
+    direction: PRESENT_ABSENT
+    interpretation: >-
+      The respiratory chain enzyme assay is the direct measurement of the node it points at:
+      an isolated complex I defect with the other complexes intact is what that node asserts.
+    description: >-
+      No endpoint_context is recorded. The assay is not being claimed as a diagnostic endpoint
+      here, because this entry argues the MRI pattern has displaced it diagnostically; its role
+      is to establish the biochemical lesion, not to find patients.
+    evidence:
+    - reference: PMID:23553477
+      reference_title: "NUBPL mutations in patients with complex I deficiency and a distinct MRI pattern."
+      supports: SUPPORT
+      directness: DIRECT
+      evidence_source: HUMAN_CLINICAL
+      snippet: "The 4 tested patients had a respiratory chain complex І deficiency."
+      explanation: >-
+        The measurement and the node are the same quantity, so the sentence reporting the assay
+        result in NUBPL patients supports the readout link directly.
   notes: >-
     An isolated deficiency of respiratory chain complex I with the other complexes intact. This
     is the biochemical phenotype that places the disorder in the nuclear complex I deficiency
@@ -646,6 +667,28 @@ biochemical:
       Cyrillic capital I the source prints in place of a Latin I in "complex І".
 - name: Reduced Spare Respiratory Capacity
   presence: PRESENT
+  readouts:
+  - target: Mitochondrial Complex I Deficiency
+    relationship: READOUT_OF
+    direction: NEGATIVE
+    interpretation: >-
+      Respirometry on patient fibroblasts reports the functional cost of the complex I defect:
+      the deeper the deficiency, the less reserve the cell can call on above baseline demand.
+    description: >-
+      A functional consequence of the node rather than a measurement of it, which is why the
+      supporting sentence is graded INDIRECT. No endpoint_context is recorded: this is a
+      mechanistic characterisation in cultured patient cells, not a diagnostic, prognostic,
+      monitoring or pharmacodynamic endpoint.
+    evidence:
+    - reference: PMID:32518176
+      reference_title: "NUBPL mitochondrial disease: new patients and review of the genetic and clinical spectrum."
+      supports: SUPPORT
+      directness: INDIRECT
+      evidence_source: IN_VITRO
+      snippet: "Mitochondrial function studies on patient fibroblasts showed significantly reduced spare respiratory capacity."
+      explanation: >-
+        The sentence reports the bioenergetic deficit in patient cells without naming complex I,
+        so attributing it to the complex I deficiency node takes one inference step.
   notes: >-
     Measured in patient fibroblasts. Spare respiratory capacity is the reserve a cell can call
     on above baseline demand, which is why the deficit shows first in tissues whose energy


aleixpuigb:
All four blocking findings fixed plus suggestion 5, in one push (`23c91d524`).

Your diagnosis was the useful part and I want to record it, because it generalises: every blocking finding was under-consumption of a source already sitting in this PR. `references_cache/PMID_32518176.md` is 1,095 lines of full text, and I treated it as an abstract. The abstract of a review is a poor summary of the case series inside it — the cognition finding, the gait description, the treatment content and three phenotypes were all in the body and none in the abstract.

**Blocking 1 — treatments.** Added occupational therapy (`NCIT:C121351`), linked via `target_mechanisms` to the `Tremor` phenotype since the source attributes that patient's improvement to it; a mitochondrial cofactor cocktail under `NCIT:C15986` with CHEBI agents for coenzyme Q10, carnitine and alpha-lipoic acid; and EPI-743 as an experimental agent.

One deliberate choice: **the cocktail carries no `target_mechanisms` link.** The sentence says treatment is "currently limited to" these agents, which is a statement about the state of the field, not evidence that they act on the complex I assembly defect. I did curate that ceiling statement as its own evidence item, for the reason your finding implies — an empty `treatments:` section and a section listing a cocktail read identically to a consumer unless the limit is stated explicitly.

The bracket-stripping problem bit here too. `(e.g., supplements of coenzyme Q10, carnitine, and α-lipoic` and `(e.g., EPI-743)` both sit inside parentheses, so the quotes are taken around them and the drug name is carried in the treatment `name` instead. Noted in the `explanation` so it isn't retried.

**Blocking 2 — cognition.** This was your best finding and the reasoning was exactly right: the entry used preserved cognition in the UK kindred as the basis of an `OPEN_QUESTION` while never curating the impaired cognition it was being contrasted against. Added `HP:0100543` Cognitive impairment (`FREQUENT`), plus hyperreflexia (`HP:0001347`) and clonus (`HP:0002169`) as separate phenotypes since HPO codes them distinctly, dysmetria (`HP:0001310`), and unilateral sensorineural hearing loss (`HP:0000407`, `OCCASIONAL`).

**The Leigh-like phenotype in Patient 3 I did not add**, and I'd rather say why than have it look like an oversight: HPO has no Leigh-like or Leigh syndrome phenotype term — Leigh syndrome is a MONDO disease concept — and binding a basal-ganglia or leukoencephalopathy term in its place would name a different finding than the source does. It's recorded in `notes:` instead, where it bears on the open question and on the relationship to the sibling MC1DN entries.

**Blocking 3 — Gait ataxia.** You're right that this isn't the `preferred_term`-may-be-more-specific carve-out; they're siblings. I took your second option: kept `HP:0002066` and earned it with the full-text gait sentence, setting `preferred_term: Gait ataxia`. The original cerebellar-syndrome quote stays as a second evidence item whose `explanation` now says explicitly that on its own it supports only "ataxia". The gait snippet is quoted with the source's own double space after "as". No rename was needed, so the bare-name `downstream` target at line 253 is untouched.

**Blocking 4 — GeneReviews.** Added a `notes:` block. There's no NUBPL-specific chapter, and I followed MC1DN24 in not tagging PMID:26425749, for a reason specific to this disease rather than by default: NUBPL disease was defined by a white matter MRI pattern (cerebellar cortex, deep cerebral white matter, corpus callosum) rather than by the bilateral symmetric basal ganglia and brainstem lesions that define Leigh syndrome, and the Leigh-like picture appears in one of five patients. Tagging a Leigh spectrum overview would assert a nosological placement the literature doesn't make.

**Suggestion 5 — taken, but not where you suggested.** `DiseaseMappings` has only a `mondo_mappings` slot; there's no OMIM mapping slot. `Autosomal_Dominant_Sensory_Ataxia_1` already records an OMIM identifier under `external_assertions` with a `description` saying exactly that, so I followed that precedent. OMIM 618242 is there now with the `(MIM 613621, 618242)` provenance explained.

I also corrected the create record's `sections:` list, which listed `treatments` when the entry had none — good catch.

**Suggestion 6 — deferred, not declined.** Your reading of `project.justfile` is right and worth testing: `{{args}}` is interpolated after the hardcoded `--var "mondo_id="` on line 2021, so a later `--var` may well win. I'm not running it right now because an openscientist job for another disease in this batch is in flight, and concurrent launches earlier today aged four jobs out against the provider's 3600s timeout. I'll attempt it once the queue drains and report the result on this PR. If it produces a correctly-anchored report I'll add it and accept the re-review cycle — a working DR artifact for a new entry is worth one.

46/46 snippets verified. `just validate-disorders` passes; entity refs, causal targets, duplicate keys, qualifier terms and enum values all clean. No new cache rows — every term added this round was already cached.

---

aleixpuigb:
Both findings fixed plus both suggestions, in one push (`262386bc8`). You were right on both, and both were defects I introduced in the previous round rather than pre-existing ones.

**Finding 6 — and the binding is the smaller half of this.** I wrote a false claim into an evidence `explanation`: "HPO has no unilateral sensorineural hearing impairment term at this level". `HP:0025797` exists, resolves against `ols:hp`, and validates in that slot — I confirmed both before rebinding. The claim came from assumption, not from a lookup.

The reason this one bothers me more than a wrong CURIE would: a wrong binding is caught by `validate-terms`, but a false statement about ontology coverage sitting in an `explanation` is caught by nothing, and it is precisely the kind of prose the next curator trusts instead of re-checking. "No term exists" is a claim like any other and needs a search behind it. Rebound to `HP:0025797`, `preferred_term` keeps the right-sidedness HPO doesn't code, and the explanation now says that accurately.

**Finding 7 — connectivity.** All five round-1 phenotypes were orphans. Your point that `check-causal-targets` cannot see this is the useful part: it verifies that a target *resolves*, never that a phenotype is *reachable*, so an entry can go from one connected component to one plus five orphans with every gate green.

Dysmetria was the clearest case and you framed it correctly — its four sentence-mates from the same source clause were already wired `DIRECT` from that node, so leaving the fifth unwired was an inconsistency inside the entry, not a scoping call. Wired as you suggested: `Dysmetria` `DIRECT`; `Cognitive impairment`, `Hyperreflexia` and `Clonus` `INDIRECT_UNKNOWN_INTERMEDIATES`. The hedged type is doing real work on those three — the source lists them among presenting features without attributing them to the lesion, so the corticospinal-tract route is my inference, and `INDIRECT_UNKNOWN_INTERMEDIATES` states exactly that there are intermediates and they aren't known. Bare names throughout.

**Sensorineural hearing loss stays unconnected**, with the reason now in its `description` as you suggested: nothing in this literature connects that one patient's hearing loss to the cerebellar/white-matter lesion or to any other node here, and an unsupported edge is worse than an unconnected phenotype.

**Suggestion 8 taken, including the better quote.** `NCIT:C152840` Vatiquinone added as `therapeutic_agent` (verified against `ols:ncit`). You were right that the patient-level `"Supplementary data), from treatment with EPI-743."` is the stronger source — it names the drug and sidesteps bracket-stripping rather than working around it. Both quotes are kept since they support different claims: one that the drug was given to a patient, one that the field regards it as experimental.

**Suggestion 9 taken.** `Hyperreflexia with clonus` renamed to `Hyperreflexia` to match its binding — done *before* wiring, so no foreign key broke, which is the sequencing you flagged. Cognitive impairment stays `FREQUENT`, but it's now argued rather than defaulted: 4/5 is nominally 80%, the `VERY_FREQUENT` floor, but at n=5 one reclassified patient moves it to 60%, so the band is set from the interval width rather than the point estimate. That reasoning is in the `explanation` now.

Noted your correction to your own suggestion 5 — checking it rather than restating it is the useful behaviour, and it's the same discipline that would have caught my finding-6 error.

47/47 snippets verified. `just validate-disorders` passes; terms, entity refs, causal targets, duplicate keys, qualifier terms and enum values all clean. Four additive cache rows for the two newly bound terms.

Suggestion 6 from round 1 (the DR retry with an explicit `--var mondo_id`) is still queued behind the in-flight openscientist job for another disease in this batch; I'll report the result here.

---

aleixpuigb:
Merge-conflict resolution only — no content change. Flagging what this push is, since it costs the approval.

This PR went un-mergeable against `main` overnight and was skipped by the shepherd's sweep while three siblings were enqueued. The conflict is a nice illustration of how these arise: someone curated **mitochondrial complex I deficiency, nuclear type 14** (`MONDO:0032619`) on `main` while this branch added **nuclear type 21** (`MONDO:0032625`). Adjacent members of the same numbered series, so their rows sort next to each other in both `cache/mondo/terms.csv` and the `diseaseterm` enum cache, and git cannot tell that both belong.

Resolved by merging `origin/main` in (not rebasing), keeping **both** rows in both files, and letting `just normalize-cache` place them rather than hand-positioning. `just check-cache-order` and `just check-term-cache-integrity` both pass.

What did **not** change: `kb/disorders/Mitochondrial_Complex_I_Deficiency_Nuclear_Type_21.yaml`, its three history records and its five reference caches are byte-identical to the tree you approved. 47/47 snippets still verify and `just validate-disorders` passes on the merged tree. I also confirmed `cache/dataset_accessions.json`, `app/data.js` and `app/models/data.js` are byte-identical to `origin/main` — the merge brings in main's versions and the branch contributes nothing to them.

Your non-blocking note stands unaddressed and deliberately so: the two `biochemical:` markers are still the remaining graph singletons. You declined to make it blocking after marking that dimension adequate twice, and a merge resolution isn't the place to widen the diff.

---

aleixpuigb:
**`claude-review` is red on this PR and the failure is not this PR's.** Standing down rather than re-running, with the reason.

The review agent errored at 0s. From the job log ([run 33951722456](https://github.com/monarch-initiative/dismech/actions/runs/33951722456)):

```
detail="You've hit your limit · resets 11:20am (UTC)"
num_turns=1 cost=$0
```

That is an account usage limit on the credential the workflow ran with — no review was attempted against the diff. [#10972](https://github.com/monarch-initiative/dismech/pull/10972) failed identically in the same minute, so it is repo-wide rather than specific to either branch.

**I have not re-run it.** The limit is deterministic until it resets at 11:20 UTC, so a re-run now would fail the same way and spend the one re-run this situation gets. I'll re-request the review after the reset.

Worth flagging for whoever owns the workflow config, because the workflow diagnoses itself here: it prefers `ANTHROPIC_API_KEY` (pay-per-use, no weekly cap) and only falls back to `CLAUDE_CODE_OAUTH_TOKEN` when no API key is set. Hitting a weekly-style limit therefore suggests the API key secret is unset and scheduled reviews are running on the subscription token. That's a repo-settings question rather than anything a PR can fix, and I'm raising it here rather than filing an issue since it's your call.

`test (3.13)` is still running and unaffected. The push it is testing is the merge-conflict resolution described in my previous comment — merge-only, with the entry, history records and reference caches byte-identical to the approved tree.

---

aleixpuigb:
/review

---

aleixpuigb:
Pushed `db8fd2903`, which takes up the non-blocking item you left on the round-2 approval: the two `biochemical:` markers were the entry's remaining graph singletons.

Both now carry a `BiomarkerReadout` link to `Mitochondrial Complex I Deficiency`.

- **`Isolated Complex I Deficiency`** — `READOUT_OF`, direction `PRESENT_ABSENT`, evidence graded `DIRECT`. The respiratory chain enzyme assay and the node assert the same quantity, so there is no inference step.
- **`Reduced Spare Respiratory Capacity`** — `READOUT_OF`, direction `NEGATIVE`, evidence graded `INDIRECT`. The quoted sentence reports a bioenergetic deficit in patient fibroblasts without naming complex I, so attributing it to that node takes one step.

`endpoint_context` is deliberately absent on both, with the reason in each link's `description` rather than left for you to infer. On the enzyme assay, asserting `DIAGNOSTIC` would contradict this entry's own argument that the MRI pattern has displaced it diagnostically. On the respirometry, a mechanistic characterisation in cultured patient cells is none of `DIAGNOSTIC` / `PROGNOSTIC` / `MONITORING` / `PHARMACODYNAMIC` / `CANDIDATE_SURROGATE`, and picking the nearest would be worse than leaving it off.

**Measured rather than eyeballed**, since you did the same in round 3: building the causal graph at `1e4c474a9` gives 6 components (one 25-node plus five singletons); at `db8fd2903` it gives 4. The three that remain are the two deliberately unlinked treatments and the deliberately unwired `Sensorineural hearing loss` phenotype — all three endorsed in review, and none of them a readout waiting to be drawn.

49/49 snippets verified, up from 47/47. `validate-disorders` passes; `check-entity-refs`, `check-causal-targets`, `check-duplicate-keys`, `check-enum-values`, `check-qualifier-terms`, `check-snippet-length`, `check-title-snippets`, `check-snippet-grading`, `check-folded-hyphens` and `validate-history` all clean. Entry plus one history record; no cache rows, no reference caches.

**On why this arrives as a push rather than a re-review request.** The approval here was dismissed by the merge-conflict resolution, and yesterday's automatic re-review errored at 0s on a usage limit without reading the diff. After the limit reset I tried both documented ways to ask for a new review and both are 403 for this integration (`workflow_dispatch`, and `rerun_failed_jobs` on the failed run), and the `/review` comment above was silently skipped because the harness appends an attribution footer server-side, so the workflow's exact-match `body == '/review'` is false. That is [#9321](https://github.com/monarch-initiative/dismech/issues/9321), already open with the right fix; I've added today's occurrence there rather than filing anything new. `pull_request: synchronize` is the only review trigger an agent can reach, which is why the work you'd already asked for is the thing carrying it.

---

aleixpuigb:
You're right about the number, and I've verified it rather than just conceding: rebuilding the graph at `262386bc8` gives 28 nodes, 22 edges, components `[23, 1, 1, 1, 1, 1]`. The history record's "one 25-node component plus five singletons" is the after-size written into the before-state, and as you say it doesn't even reconcile against a 28-node graph.

Taking your advice and **not** pushing for it alone — a re-review cycle costs more than the defect does, and the commit message and the PR comment both carry the right figures.

I'm not leaving it to chance either. Both of this branch's previous pushes were forced by merge conflicts in `cache/mondo/terms.csv` and the `diseaseterm` enum cache, and the numbered nuclear complex I series is still actively being curated, so another one is more likely than not. If this branch has to be touched again for any reason, the one-line fix goes in that push. If it merges first, it merges with the wrong parenthetical, which is the cheaper of the two outcomes.

Worth naming what the class of error is, since you flagged durability as the reason to care: it is the same shape as the false ontology claim you caught in round 2 — a factual assertion in prose that no gate reads. The difference is that this one is checkable from the repository itself, so a reader who doubts it can rebuild the graph in four lines. The round-2 one required going out to HPO.

Thanks for measuring the components independently rather than taking the commit message's word for it. That is exactly the check that catches this.

---

PR #10996, opened 2026-09-04
curate: Add Auroneurodental Syndrome (MONDO:0970998, NAA80)

### LINKED ISSUE #10984: Curate auroneurodental syndrome (MONDO:0970998)
Curate a dismech entry for **auroneurodental syndrome** ([MONDO:0970998](https://monarchinitiative.org/MONDO:0970998)) — a syndromic disorder combining auditory, neurological and dental features.

The stub records no causal gene and MONDO's only parent is `syndromic disease`, so both the gene and the nosological placement need to come from OMIM/Orphanet via the stub's xrefs. Re-querying MONDO will not help — `enrich-stubs` already read the same release. If the literature turns out to be a single family with nothing mechanistic to build a pathograph from, say so and record it rather than padding an entry.

Stub: `stubs/Auroneurodental_Syndrome.yaml`

kevinschaper:
Closes #10984

Adds `kb/disorders/Auroneurodental_Syndrome.yaml` for MONDO:0970998 (OMIM 620830) and deletes its stub.

## The question the issue asked

> If the literature turns out to be a single family with nothing mechanistic to build a pathograph from, say so and record it rather than padding an entry.

It **is** a single family — two brothers, one consanguineous Portuguese pedigree, PMID:34805998, and a PubMed search for NAA80 turns up no second clinical report. But the second half of the condition does not hold. The defining paper measured the variant's structural consequence, the enzyme's solubility, the residual actin acetylation by mass spectrometry, the downstream cytoskeletal phenotype, and its rescue by wild-type NAA80 — all in the same study. A zebrafish knockout published three years later (PMID:39384430) then tested the hearing arm independently and found it. So there is a real mechanism to curate, and the entry is a full one rather than a stub disposition.

`entry_type`: **DISEASE**.

Identity came from the MONDO xrefs, as the issue directed — OMIM:620830, and the single annotation-source PMID behind every one of MONDO's phenotype annotations. The gene is **NAA80** (`hgnc:30252`), the enzyme that performs the final maturation step on actin, and the allele is homozygous c.389T>C, p.(Leu130Pro).

## The mechanism

Seven nodes. Leu130 sits at the end of beta-strand 1 of the NAA80 catalytic domain with its backbone in the sheet's hydrogen-bond network; proline was predicted to disturb that fold, and recombinant mutant protein is predominantly insoluble where wild type is essentially soluble. Less enzyme means less of the modification it performs: 25 to 65 percent of beta- and gamma-actin is unacetylated in patient cells against about 0.5 percent in healthy donors. Because unacetylated actin depolymerizes more slowly, patient fibroblasts and PBMCs accumulate filamentous actin, grow excess filopodia and migrate faster — and re-expressing wild-type NAA80 puts all three back to control levels.

It is curated as `PARTIAL_LOSS_OF_FUNCTION`, on the authors' evidence rather than by default: driving the mutant cDNA from a weak promoter gives only partial restoration of acetylation while stronger promoters reach near-maximal, so the mutant enzyme retains activity and the defect is one of protein availability.

The four organ-level nodes (inner ear, developing brain, skeletal muscle, craniofacial/dental/digital) are all `mechanism_confidence: HYPOTHETICAL`, and the entry says why: everything measured was measured in fibroblasts, PBMCs, or cell lines, and the step to the affected organs rests on the quantified phenotypic overlap with ACTB and ACTG1 patients rather than on any measurement in ear, brain, muscle or tooth.

27 phenotypes, all wired into the pathograph — no orphans.

## The zebrafish model is curated with its failure

`animal_models` carries three `modeled_mechanisms` links for the naa80-null zebrafish:

| target | relationship | fidelity |
|---|---|---|
| Reduced N-Terminal Acetylation of Cytoplasmic Actin | `RECAPITULATES` | HIGH |
| Disordered Actin Dynamics in the Inner Ear | `RECAPITULATES` | MODERATE |
| Disordered Actin Dynamics in Skeletal Muscle | **`FAILS_TO_RECAPITULATE`** | LOW |

The ear link carries four readouts (otolith size, hair-cell number and stereocilia length, acoustic startle response, Yo-Pro-1 viability). The muscle link carries an `UNCHANGED` readout — a real negative, not an unmeasured one — plus `limitations` and `REFUTE` evidence, and is paired with a `HUMAN_MODEL_MISMATCH` discussion that lays out the three open readings without choosing between them.

The severity mismatch runs in a direction worth noticing: the fish are null and lose actin N-terminal acetylation essentially completely, while the human allele leaves roughly half of it. A null fish with normal muscle is therefore a *stronger* negative than a hypomorphic one would be — which is exactly why the muscle node needs the discussion rather than a shrug.

## Two things recorded rather than smoothed over

- **A second homozygous variant segregates in this family.** Both brothers are also homozygous for PLXNB1 p.Ser911Gly, absent in the healthy siblings; the report states this and does not discuss it further. Nothing cited excludes a PLXNB1 contribution. That is the main threat to the gene-disease assertion in a one-family disorder, so it is a knowledge gap attached to `genetic#NAA80`, not a footnote.
- **No measurement exists in an affected tissue.** Recorded as its own gap, with patient-derived iPSC inner-ear organoids or cortical neurons named as the route that would not require tissue from two individuals.

## The deep-research report was discarded

Falcon (Edison Scientific Literature), given the entry name, opened with an "entity-resolution warning": it could not verify the label "auroneurodental syndrome", proposed **labyrinthine aplasia, microtia and microdontia (LAMM) syndrome** (FGF3, OMIM 610706) as "the likely intended entity", and wrote the whole report about that instead. NAA80 appears **zero** times.

`just preflight-dr <report> MONDO:0970998` returns SKIP rather than FAIL — MONDO records no causal gene for this term, so the automated gene-identity check has nothing to discriminate on — but its diagnostics say it plainly:

```
  top genes       : FGF3=42, AR=2, HP=2, EYA1=2, SIX1=2
  OMIM (MONDO)    : 620830
  OMIM (report)   : 610706
```

The cause is real and worth recording: "auroneurodental syndrome" is an OMIM name that appears in no PubMed record. The defining paper's title never uses it. The entry's `notes` now say to search this disease by gene or OMIM number, never by name.

That makes three of three Falcon runs in this batch unusable — see #10987 (substituted a different disease sharing an eponym) and #10992 (right disease, but concluded no causal gene exists, missing the paper that found it). Name-only deep research on ultra-rare entities looks like a systematic weak spot rather than three unlucky draws.

## Validation

| check | result |
|---|---|
| `just validate` | passed; **68/68 snippets verified**, 0 skipped (all references are PMIDs — no DOI citations, so nothing is hidden by `skip_prefixes`) |
| `just check-entity-refs` / `check-duplicate-keys` / `check-causal-targets` / `check-enum-values` / `check-qualifier-terms` | passed |
| `just check-title-snippets` / `check-snippet-length` / `check-snippet-grading` / `check-folded-hyphens` (whole-KB) | no new findings |

`check-snippet-grading` caught one real defect during curation — the same quoted sentence graded `HUMAN_CLINICAL` in one block and `IN_VITRO` in another — which is fixed rather than baselined. Cache rows added are only the six CURIEs this entry introduces (`MONDO:0970998`, `hgnc:30252`, `GO:0004596`, `GO:0006474`, `HP:0006342`, `HP:0025372`); a stray `GO:0018206` row from a mid-edit run was reverted and re-derived rather than hand-removed.


kevinschaper:
All three IMPORTANT findings are fixed, plus most of the suggestions, in one push. I checked each finding against `references_cache/PMID_34805998.md` before acting — all three were right about the omission. Two of the suggested HPO terms were not.

**Cardiac finding (item 1).** Added as a phenotype for proband 1.2, wired as a second `INDIRECT_UNKNOWN_INTERMEDIATES` edge from the cytoskeletal node in parallel with the constipation edge, exactly as you described. The edge quotes the Discussion sentence that pairs them — "Similarly, the ventricular dysfunction seen in NAA80 individuals could reflect a minor ACTC1 dysfunction" — graded `HUMAN_CLINICAL` / `INDIRECT` to match its constipation sibling.

I did **not** use `HP:0025169`. The label you gave is right — I confirmed it is "Left ventricular systolic dysfunction" — but the term is wrong for this claim. The paper says "mild systolic dysfunction" and later "ventricular dysfunction", and never names a chamber, so binding the left-ventricular term would manufacture specificity the source does not carry. I also rejected `HP:0005185` (Global systolic dysfunction): its definition requires an enlarged left ventricle and marked cardiomegaly, and the source says explicitly that no other abnormality was found. The binding is `HP:0006673` **Reduced systolic function**, the unqualified parent of HP:0005185, with `severity: MILD` and the source wording in `preferred_term`. That was the only new CURIE in the cache diff.

**Two phenotypes from the already-quoted sentence (item 2).** Added `Peripheral edema` (HP:0012398) and `Vomiting` (HP:0002013), reusing the same snippet with the same grading. Both are **deliberately unwired**, and that is the one place I have not done what the shape of the entry would suggest — so it is worth stating plainly. The report gives no mechanism for either: the etiological workup including clinical chemistry and metabolic screening of urine and plasma found no cause, and nothing in the cited literature places episodic infection-associated swelling or vomiting on the actin axis. Drawing an edge from the cytoskeletal node would assert a route no source supports. The reason is in each phenotype's `notes` and in a new entry-level notes point, which also records the intercurrent-illness decompensation pattern you asked about — that is the clinically actionable part, and it is now stated even though nothing explains it.

**Toes (item 3).** Added, wired from the dysmorphogenesis node beside `Tapered finger`. You were right that `HP:0001883` (Talipes) is wrong — it names a foot deformity, not toe placement. I used the generic `HP:0001780` **Abnormal toe morphology** ("An anomaly of a toe") with `preferred_term: Disorganized implant of the toes`. HPO's specific toe terms — overlapping, widely spaced, deviated, short — each assert something "disorganized implant" does not say, so any of them would be a loose match rather than a specific one.

**Suggestions taken.** `Long philtrum` (HP:0000343) and `Bulbous nose` (HP:0000414) from the Fig 1F legend, both wired. The residual-activity window is now an evidence item on the acetylation node, graded `COMPUTATIONAL` / `INDIRECT` because it is an inference from a simplified kinetic model rather than an activity assay, with the rarity argument in the explanation. `external_assertions` now carries OMIM:620830 (confirmed via OAK as the OMIM xref of MONDO:0970998) and the ClinVar record.

**One correction on the ClinVar accession.** You gave `VCV001301869.1`; the live record is at `.6`, so I recorded it without a version suffix and noted the version I checked. More substantively, ClinVar's preferred title names this variant on a different transcript from the paper: `NM_001200016.2(NAA80):c.323T>C (p.Leu108Pro)`, not the `c.389T>C p.(Leu130Pro)` used throughout this entry. It is the same variant — both resolve to GRCh37 chr3:g.50334572A>G, the coordinate the paper gives, and it carries OMIM allelic variant 607073.0001 — and I put that in the record's `notes` so nobody later reads it as a second allele. ClinVar links it to MONDO:0970998 and OMIM:620830 and classifies it Pathogenic/Likely pathogenic.

**One thing I fixed that nobody flagged.** The entry's notes described a "Falcon/Edison run". There is no `edison` provider in this repository, so that now reads "Falcon deep-research run".

**Declined.** The optional split of pathophysiology node 1 — you called it defensible and I agree; the evidence is a single coherent argument. `HP:0012450` Chronic constipation and `HP:0000219` Thin upper lip vermilion were round-1 suggestions dropped in round 2, and I left both: "mild constipation requiring laxatives" listed as an ongoing feature in both brothers is reasonably chronic, and "thin lips" with a separately reported "small upper lip" supports the vermilion term. Happy to change either if you still want them.

**Validation.** `validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms` and `validate-disorders` all pass on the changed file. `count-verified-snippets` is **76/76** with nothing skipped by prefix — the entry cites no DOIs, so every snippet is machine-checked. Whole-KB `check-title-snippets`, `check-snippet-length` and `check-snippet-grading` are clean with no new baseline rows.

**Orphan list.** Sweeping every `downstream`, `sequelae`, `reports_on`, `target_mechanisms` and `modeled_mechanisms` target against the entry's 33 phenotypes and 7 pathophysiology nodes: zero dangling targets. Phenotypes with no incoming causal edge: `Peripheral edema` and `Vomiting` — both deliberate, both documented above. Everything else is connected.

The cache diff is a single tool-generated additions-only row pair for HP:0006673. `references_cache/` is untouched by this push.

PR #10997, opened 2026-09-04
curate: Add PSMC1-related neurodevelopmental disorder (MONDO:0859296)

### LINKED ISSUE #10985: Curate neurodevelopmental disorder with poor growth, spastic tetraplegia, and hearing loss (MONDO:0859296)
Curate a dismech entry for **neurodevelopmental disorder with poor growth, spastic tetraplegia, and hearing loss** ([MONDO:0859296](https://monarchinitiative.org/MONDO:0859296)) — a syndromic neurodevelopmental disorder in MONDO's numbered NDD series.

The stub records no causal gene. Descriptive NDD names like this one are usually a single OMIM entity with a named gene that MONDO has not linked with `RO:0004003`, so go to OMIM via the stub's xrefs for identity. Watch for the gene already being covered by an existing gene-named entry in `kb/` — that has been the most common outcome in this queue lately.

Stub: `stubs/Neurodevelopmental_Disorder_With_Poor_Growth_Spastic_Tetraplegia_And_Hearing_Loss.yaml`

kevinschaper:
New `kb/disorders/` entry for **neurodevelopmental disorder with poor growth, spastic tetraplegia, and hearing loss** (`MONDO:0859296`, OMIM:620071) — an autosomal recessive proteasomopathy caused by biallelic *PSMC1* p.I328T. `entry_type: DISEASE`; the stub is deleted.

## Finding the gene, and checking nobody had it already

The stub recorded no causal gene, so MONDO could not answer this. `MONDO:0859296` xrefs `OMIM:620071`, and NCBI's `mim2gene_medgen` maps MIM 620071 to GeneID 5700 = **PSMC1**; MedGen's curated gene–disease link agrees. The defining publication is Aharoni et al. 2022 (PMID:35861243).

The obvious risk in this queue is that the gene is already covered under a gene-named entry. It is not: `PSMC1` appears in `kb/` only inside review snippets cited by the PSMB1 entry (`Neurodevelopmental_Disorder_with_Microcephaly_Hypotonia_and_Absent_Language`), which is a different gene and a different OMIM entity, and `Proteasome_Associated_Autoinflammatory_Syndrome` covers the immunoproteasome/PRAAS side. The entry is structured to match that PSMB1 sibling so the two read as a pair.

## What the mechanism evidence actually supports, and what it does not

PSMC1 encodes Rpt2, one of six AAA+ ATPases in the base of the 19S regulatory particle. The chain is: variant → destabilised AAA+ ATPase fold → reduced 26S proteolytic capacity → accumulation of ubiquitinated substrates → neuronal proteostasis failure.

The first half is well evidenced and the second half is not, so the grading says so rather than smoothing it over:

- Structural modelling of the hydrophobic-core disruption is graded `COMPUTATIONAL`.
- The fly complementation result — wild-type human PSMC1 nearly fully rescues an Rpt2-knockdown eye phenotype, the patient allele only partly — is graded `MODEL_ORGANISM` and is the **only** evidence that this allele is hypomorphic rather than null. That matters, because the germline null is embryonic lethal in mice.
- **Every one of the eight edges out of the terminal node** is `causal_link_type: INDIRECT_UNKNOWN_INTERMEDIATES` with `directness: INDIRECT`. No source demonstrates the step from reduced neuronal proteostasis to any particular clinical finding; the edges record a clinical association in variant carriers.

## Deliberate omissions

- **No frequency band anywhere.** Every proportion available is a 3/3 or 2/3 count within one sibship. The numerators are in each phenotype's `notes:`.
- **Eight of sixteen phenotypes are left unwired**, each with the reason in `notes:`: failure to thrive, hearing impairment, micropenis, cryptorchidism, macrocytic anemia, neonatal cholestasis, elevated hepatic transaminases, abnormal facial shape. Wired: global developmental delay, severe intellectual disability, spastic tetraplegia, central hypotonia, chorea, absent speech, inability to walk, ventriculomegaly.
- **A duplicated renal collecting system and a ventricular septal defect are not curated.** The source itself says they might not be related to the variant.
- **`HP:0011398` "Central hypotonia" is obsolete**, so `HP:0001252` Hypotonia is bound and the specificity is carried in `preferred_term`.

## Models

Both are mismatched, in different ways, and a `HUMAN_MODEL_MISMATCH` discussion says so:

- **Drosophila Rpt2 eye knockdown with human PSMC1 rescue** — `PERTURBS`, with a `RESTORED` rescue readout. Reports residual protein function, not the disease phenotype.
- **Psmc1 neuron-restricted conditional knockout mouse** (PMID:18701681) — `PARTIALLY_RECAPITULATES`, `fidelity: LOW`. It is a Parkinson-disease model: a complete conditional null in postnatal neurons producing adult neurodegeneration with Lewy-like inclusions, against a germline hypomorphic missense causing a static congenital disorder. It supports the neuronal-dependence step and nothing else.

## Two things worth a reviewer's eye

- **OMIM flags this gene–disease relationship with "?".** `mim2gene_medgen` renders it as the qualifier `question`, defined as "an unconfirmed or possibly spurious mapping". Consistent with one reported family. Recorded in `notes:` as context, not as an evidence item, since it is a property of a data file rather than a citable statement.
- **Named-entity confusion is a live risk for this gene.** A PubMed search on `PSMC1` alone returns almost entirely cancer-prognosis and Parkinson-disease papers. The `notes:` block records that, plus the confusable PSMC3 and PSMB1 entities.

## Deep research

I attempted `just research-disorder edison` first, as preferred; **edison is not available in this environment** (`Provider 'edison' not available. Available: falcon, claude_code`). A `falcon` run was then started and had not produced a report by the time this PR was opened, so this entry is curated entirely from primary literature. If the report lands and passes `just preflight-dr` / `just validate-research-reference`, I will add it in a follow-up commit; otherwise this PR stands without one.

## Validation

`validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `validate-disorders`, `validate-history`, `check-stubs` all pass. `count-verified-snippets`: **67/67 verified, 0 skipped by prefix** (no DOI-cited snippets). Whole-KB `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` clean against `origin/main`.

Every snippet was additionally checked by hand as a **raw** exact substring of its cache file (not merely whitespace-normalised), and every `reference_title` against the cache frontmatter — which caught three titles carrying the frontmatter's literal quote characters.

Cache diff is exactly the two CURIEs this entry introduced (`MONDO:0859296`, `hgnc:9547`).

Closes #10985


kevinschaper:
The falcon deep-research run finished after this PR was opened, so the report is now committed here (`research/…-deep-research-falcon.md`, plus its citations sidecar and one Edison artifact).

**What it contributed: almost nothing, and that is the honest answer.** It reached the same conclusions from the same primary source (PMID:35861243), which is what a report ought to do when the entity has exactly one publication. It did not surface a second family, a functional assay in patient cells, or any source I had not already cited.

**One substantive difference, and I acted on it.** The report reads the paper's per-patient Table 1 as recording chorea in two assessed patients with data unavailable for the eldest sibling, where the narrative names chorea among the features shared by all three. That table is not in the cached full text, so I could not check it. Rather than pick a side, the Chorea phenotype's `notes:` now asserts no count and says why. The cited snippet is unchanged — it is still the narrative sentence.

**Gate results:**

- `just preflight-dr … MONDO:0859296` → **SKIP**, because MONDO records no `RO:0004003` gene for this term and the automated gene-identity check therefore cannot discriminate. That is the same gap that made this claim's identity work manual in the first place. The manual OMIM route stands, and the report's own gene mentions run PSMC1 48 to the next gene's 2.
- `just validate-research-reference` → **4/4 references resolved, 0 possible confabulations**, 3/4 assessed as on topic.
- Term validation → **40/40 resolved, 0 obsolete**. One label was flagged for a second look, `UBERON:0000955`; this entry does not bind it.

Nothing in the report's `## Term Validation` section is unresolved or obsolete, so no binding here rests on a flagged CURIE.

The three `references_cache/DOI_*.md` files in this commit are byproducts of the report's own citation validation, not evidence sources for the KB entry — the entry cites only PMID:35861243 and PMID:18701681 for its own claims, plus the two proteasomopathy reviews already on `main`.

kevinschaper:
Blocking item fixed, and I took the four non-blocking suggestions too. Thanks — the gastrostomy point was a real gap, and an embarrassing one: the treatments list was six standard-of-care inferences with the one thing the paper actually documents left out.

## Blocking: gastrostomy feeding

Added a **Gastrostomy Feeding** treatment bound to `NCIT:C93322` **Enteral Tube Feeding**, which validates as reachable from `NCIT:C25218` (`just validate-terms` passes; the CURIE was already in `cache/ncit/terms.csv` and the treatment-action enum cache, so this adds no cache rows).

- Evidence is the exact quote you named: `was fully nonambulatory, fed through gastrostomy, with severe failure to thrive`, `evidence_source: HUMAN_CLINICAL`.
- A second item quotes the table legend (`Abbreviations: n/a, data not available; PEG, percutaneous endoscopic gastrostomy.`) graded `directness: INDIRECT`, with an explanation saying plainly that it is a legend rather than a finding, and that it establishes gastrostomy status was a tracked per-patient variable rather than a passing mention. No count is claimed from it, since the table itself is not in the cached text.
- `target_mechanisms` → `Failure to Thrive`, `treatment_effect: BYPASSES`, its own evidence graded INDIRECT because the source reports no growth response to feeding. **That link also gives Failure to Thrive its first incoming graph edge**, so it drops off the disconnected list.
- `therapeutic_modality: DEVICE`, not `BEHAVIORAL`. The CLAUDE.md coarse mapping puts dietary intervention under BEHAVIORAL, but this is nutrition delivered through a surgically placed tube, not a change of diet or feeding behaviour. The reason is in the treatment's `notes:` so the next person does not have to re-derive it.

## Non-blocking, all taken

- **Feeding difficulties** (`HP:0011968`) added. Both evidence items are `INDIRECT`, because the source documents the *intervention* and never names the finding — swallowing and bulbar function are not characterised anywhere in the report. Left unwired, with that stated in `notes:`.
- **Cranial vs facial split.** `Abnormal Facial Shape` now covers only the facial features (bushy eyebrows, flat midface, long nasal bridge, micrognathia), and the two cranial-vault findings are their own phenotypes: **Dolichocephaly** (`HP:0000268`) and **Microcephaly** (`HP:0000252`), both recorded as borderline with no count or severity asserted. One correction to the suggestion: `HP:0000256` is **Macrocephaly**, so `HP:0000268` is what dolichocephaly needs. Microcephaly is worth having separately for a second reason — it is the cardinal feature of the sibling PSMB1 proteasomopathy and here it is explicitly only borderline, which is a real difference between the two entities.
- **Widened subarachnoid space** (`HP:0012704`) added for the previously prose-only "broadened extra-axial space", and its evidence snippet extended to the full MRI sentence. Wired from `Neuronal Proteostasis Failure` on the same `INDIRECT_UNKNOWN_INTERMEDIATES` footing as Ventriculomegaly, since it is the same observation on the same scans.
- **Explanatory `notes:` on all five evidence-free treatments**, following the pattern you flagged on `Supportive and Symptomatic Care` — each says what the source does and does not document, rather than leaving the absence of a citation unexplained.

The entry's `notes:` block was updated to match: ten phenotypes now have no upstream mechanism edge (was eight), and Failure to Thrive is described as treatment-linked rather than unwired.

## On your disclosed limitation

The GeneReviews check is moot here: no GeneReviews chapter exists for PSMC1, and nothing in the entry rests on one either way. Worth noting for the general case that `curl` to eutils *does* work from this environment, so a future review may be able to run it.

`validate`, `validate-terms`, `check-entity-refs`, `check-duplicate-keys`, `check-causal-targets`, `check-enum-values`, `check-qualifier-terms`, `check-folded-hyphens`, `validate-disorders`, `validate-history` all pass. `count-verified-snippets`: **75/75 verified, 0 skipped by prefix** (was 67/67). Whole-KB `check-title-snippets`, `check-snippet-length`, `check-snippet-grading` clean against `origin/main`.

Every snippet re-verified by hand as a **raw** exact substring of the cache and every `reference_title` against the frontmatter. **No cache rows added** — all five new CURIEs were already cached, so the cache diff is still exactly the two CURIEs from the original commit.

A history record for this review round is included.

Issue #10978, opened 2026-09-04
CLAUDE.md recommends NCIT:C2322 (Corticosteroid) as a therapeutic_agent, but it fails the ChemicalEntityTerm enum


aleixpuigb:
`CLAUDE.md:1720` offers three worked examples of NCIT drug-class terms for `therapeutic_agent`:

> **NCIT**: use for drug classes, or for biologics/newer drugs that lack a CHEBI term (`NCIT:C20401` Monoclonal Antibody, `NCIT:C2322` Corticosteroid, `NCIT:C65216` Adalimumab)

Two of the three validate. The middle one does not.

## Symptom

Curating `Neuralgic_Amyotrophy` I followed the documented pattern for a treatment the source describes only as "high-dose oral corticosteroids" — a class, not a named drug, so exactly the case the guidance is written for:

```yaml
therapeutic_agent:
- preferred_term: corticosteroid
  term:
    id: NCIT:C2322
    label: Corticosteroid
```

```
❌ ERROR: Value 'NCIT:C2322' not in dynamic enum 'ChemicalEntityTerm' (expanded from ontology)
    path: treatments[0].treatment_term.therapeutic_agent[0].term
```

The CURIE is not wrong. `NCIT:C2322` **does** resolve to "Corticosteroid" — the label in CLAUDE.md is correct. It is a reachability problem, not a naming one.

## Cause

`ChemicalEntityTerm` is defined as:

```yaml
reachable_from:
  is_direct: false
  relationship_types: [rdfs:subClassOf]
  source_nodes: [CHEBI:24431, NCIT:C1909]
```

`NCIT:C2322` has four `subClassOf` ancestors in NCIT, and `NCIT:C1909` is not among them:

| ancestor | label |
|---|---|
| `NCIT:C1899` | Physiology-Regulatory Factor |
| `NCIT:C1908` | Drug, Food, Chemical or Biomedical Material |
| `NCIT:C2264` | Adrenal Hormone |
| `NCIT:C2315` | Hormone |

NCIT classifies corticosteroids as **hormones** rather than under Pharmacologic Substance, so the enum's `NCIT:C1909` root cannot see them. The other two examples in the same sentence are fine — `NCIT:C20401` and `NCIT:C65216` are both in `cache/enums/chemicalentityterm_b3afbb412a9e.csv` and are used 22 and 15 times in `kb/` respectively.

## This has already cost a curator once

[...]

PR #10998, opened 2026-09-04
curate: Add Neuralgic Amyotrophy (MONDO:0017362, Parsonage-Turner syndrome)

### LINKED ISSUE #10954: Curate neuralgic amyotrophy (MONDO:0017362)

Curate a dismech entry for **neuralgic amyotrophy** ([MONDO:0017362](https://monarchinitiative.org/MONDO:0017362)) — Parsonage-Turner syndrome, an immune-mediated brachial plexopathy that begins with severe shoulder-girdle pain and is followed by patchy weakness and wasting, frequently with hourglass-like nerve constrictions on imaging.

Scope worth pinning down early: this MONDO term sits under `MONDO:0015923` acquired peripheral neuropathy, so it is the **idiopathic/acquired** form. Hereditary neuralgic amyotrophy (*SEPTIN9*) is a separate MONDO concept and should not be folded in — at most it belongs as a contrast in `notes`, since its mechanism (a duplicated septin filament gene) differs from the immune-mediated one. MONDO records no descendants for this term.

No existing coverage on `main` for "neuralgic amyotrophy" or *SEPTIN9*, and no open PR curates it.

Stub: `stubs/Neuralgic_Amyotrophy.yaml`

aleixpuigb:
New entry for neuralgic amyotrophy, the painful multifocal brachial plexus neuropathy also known as Parsonage-Turner syndrome.

Closes #10954

## Scope: the acquired form only

Hereditary neuralgic amyotrophy, caused by SEPTIN9 duplication, is a separate MONDO concept with a different initiating lesion, and MONDO agrees with the split — the parent of `MONDO:0017362` is *acquired peripheral neuropathy*. The entry has no `genetic:` or `inheritance:` section for that reason, which would otherwise read as an omission.

I've stated this in the entry description rather than leaving it implicit, because the deep-research report blends the two forms and reports OMIM 162100, which is the hereditary one.

## What the pathograph asserts

The entry is built on a **three-factor susceptibility model** — genetic predisposition, biomechanical vulnerability of the plexus (likely the epineurial blood-nerve barrier), and an immune trigger for individual attacks — rather than a single initiating lesion, because that is what the literature supports rather than because it is tidier. All six phenotypes are wired into it.

Whether the pathognomonic **hourglass nerve constrictions** cause the deficit or follow the inflammatory attack is genuinely unsettled in this literature, so it is curated as a `CONTROVERSY` discussion rather than resolved in either direction.

## Hepatitis E as a trigger

An `environmental:` entry links HEV to the inflammatory-attack node with `environmental_effect: TRIGGERS`. It earns curation because the association is clinically distinctive rather than incidental: HEV-associated attacks are enriched for phrenic nerve involvement and diaphragmatic paralysis rather than the usual shoulder-girdle picture, and a 13-patient series found **normal liver enzymes in every case** — which is precisely why the trigger gets missed, since by the time the neuropathy presents there is no hepatic signal to prompt serology.

Both sources are small and uncontrolled (13 patients and 2 patients), and the entry records that rather than letting the edge imply more.

`exposure_term` is bound to the broader `ECTO:3000001` (exposure to virus) with the specificity in `preferred_term`. **I have not claimed ECTO lacks a hepatitis E term** — I tried to check, and the local ECTO index returned nothing for a control query as well as for the real one, so the search was uninformative rather than negative. That is recorded in `notes:` as a lead for whoever next has a working adapter. (I make the point because on the concurrent NUBPL PR I asserted an ontology absence I had not actually checked, and it was false.)

## Two things deliberately not curated

Both recorded in `notes:` so they are not re-litigated:

**The feline model.** The report offers feline hypertrophic brachial plexus neuritis (PMID:29925717) as a naturally occurring comparative-species analogue, and it is tempting. Reading the source rather than the summary: that cat had a chronic hypertrophic **demyelinating** neuritis with onion-bulb formation, which its own authors compare to human CIDP. NA is an acute, painful, multifocal **axonal** process. Shared brachial plexus inflammation is not enough to make it a model of this disease.

**Module conformance.** `peripheral_axonal_degeneration` was checked and does not fit — it scopes to length-dependent distal degeneration, while NA is multifocal and proximal. No `conforms_to` is asserted.

## The report, and why none of its CURIEs were used

This is the first report in this batch that is actually about the disease it names, so unlike [#10995](https://github.com/monarch-initiative/dismech/pull/10995) it is committed here.

Its identifiers are another matter, and its own `## Term Validation` section is the best argument for that tooling I have seen: **8 of 23 checked terms name a different concept**, including every treatment term it proposed.

| Report calls it | The CURIE actually means |
|---|---|
| `NCIT:C198` Prednisone | Acetaminophen |
| `NCIT:C1409` Corticosteroid | Carbenicillin Indanyl Sodium |
| `NCIT:C579` Immunoglobulin therapy | Inorganic Chemical |
| `NCIT:C157810` Neurolysis | Cervical Cancer Surgery |
| `NCIT:C15296` Physical Therapy | Isolated Chemotherapeutic Perfusion |
| `UBERON:0001824` brachial plexus | mucosa of larynx |
| `UBERON:0001379` deltoid | vastus lateralis |
| `MONDO:0100053` neuralgic amyotrophy | **anaphylaxis** |

`UBERON:0003732` "phrenic nerve" does not exist at all. The entry was written before the report and uses none of these; `MONDO:0017362` comes from the stub and is verified against the ontology.

That last row is the second wrong MONDO ID in two reports in this batch, in a different way from the first — [#10995](https://github.com/monarch-initiative/dismech/pull/10995)'s report drifted to a sibling *disease*, this one invented an ID for the *right* disease. Both are #9888.

`just preflight-dr` returns SKIP here, correctly: MONDO records no causal gene for an acquired disease, so the gene-identity check cannot discriminate and the manual check is what stands.

## Evidence base, stated in the entry

Four of the seven references are narrative reviews carrying MEDLINE publication type "Review"; three report original human observations (a four-patient biopsy series, and the two HEV series). All seven are graded `HUMAN_CLINICAL`, which is what the schema's definition supports and the KB's dominant convention, but a reader cannot tell a review from a biopsy series from a two-patient case report at the evidence item. That is recorded in the entry's `notes:` rather than left invisible — and it is the concrete case behind the measurement I posted on #9421.

## Validation

33/33 snippets verified against cached references. `just validate-disorders` passes; terms, entity refs, causal targets, duplicate keys, qualifier terms, enum values and `check-environmental-evidence` all clean. Six additive cache rows.

---

### DIFF
--- a/history/disorders/Neuralgic_Amyotrophy/2026-09-04T193941Z-claude-code-853c3c.yaml
+++ b/history/disorders/Neuralgic_Amyotrophy/2026-09-04T193941Z-claude-code-853c3c.yaml
@@ -1,0 +1,113 @@
+history_version: 1
+target:
+  kind: disorder
+  slug: Neuralgic_Amyotrophy
+  path: kb/disorders/Neuralgic_Amyotrophy.yaml
+session:
+  id: 2026-09-04T193941Z-claude-code-853c3c
+  timestamp: '2026-09-04T19:39:41Z'
+  actors:
+  - type: ai_agent
+    name: claude-code
+    model: claude-opus-5
+    agent_tool: claude-code
+links:
+  issues:
+  - https://github.com/monarch-initiative/dismech/issues/10954
+  prs:
+  - https://github.com/monarch-initiative/dismech/pull/10998
+  urls: []
+events:
+- type: EDIT
+  outcome: changed
+  sections:
+  - phenotypes
+  - treatments
+  - environmental
+  - diagnosis
+  - progression
+  - notes
+  summary: 'Review round 1 on PR #10998: consume the primary studies the entry''s
+    own notes said were missing'
+  details: |
+    The review's central criticism was correct and is the thing worth recording: this entry's own
+    notes complained that its evidence base was thin because four of seven references were
+    narrative reviews, while the deep-research report committed in the same PR contained the
+    primary studies that fix it. An RCT, a 246-patient series and a prospective matched
+    case-control were sitting in the artifact, unread. Seven of the report's 40 verified
+    references had been used.
+
+    This is the same defect the NUBPL PR was pulled up on an hour earlier, in a different guise:
+    there, the unread source was a 1,095-line full text mined only for its abstract; here it was a
+    committed report mined for a fraction of its citations. The general form is treating a source
+    as consumed once something has been taken from it.
+
+    Blocking 1 - Paresthesia frequency was OCCASIONAL (5-29%) on a snippet that carried no
+    frequency at all. Corrected to FREQUENT on PMID:16371410's 78.4% sensory involvement, with the
+    original snippet's explanation rewritten to say explicitly that it supports occurrence and not
+    frequency. Recorded that the 246-patient denominator mixes 199 idiopathic with 47 hereditary
+    cases while this entry scopes to the acquired form, and that the paper reports no separate rate
+    by subtype.
+
+    Blocking 2 - Fatigue (HP:0012378) added, FREQUENT, and wired DIRECT from the Reduced Nerve
+    Endurance node. It is the clinical readout of this entry's central claim. The second evidence
+    item is the load-bearing one: fatigue does not correlate with residual paresis on the MRC
+    scale, which is the dissociation that separates the endurance model from simple deconditioning
+    and is why the phenotype hangs off that node rather than off axonal injury.
+
+    Blocking 3 - Rehabilitation cited two review sentences for the intervention this entry argues
+    is disease-modifying. Added PMID:36697215, the only RCT in this disease, with both the adjusted
+    group difference and the responder analysis, and noted that the CI nearly touches zero and 47
+    randomised became 37 analysed. Added PMID:39402917 for the explicit negative that strength
+    training may worsen symptoms - curated because the obvious rehabilitation approach to a weak
+    limb is the wrong one here, and an entry recording only what to do leaves that trap in place.
+
+    Blocking 4 - Only hepatitis E was curated as a trigger, and it is the weakest-evidenced one,
+    while the first pathophysiology node depends on a general immune trigger. Added an
+    "Antecedent infection or vaccination" environmental entry from PMID:39364568, the prospective
+    matched case-control with central serology: 26.3% microbiologically confirmed infection, 12.3%
+    COVID-19 vaccination as potential trigger, 38.6% confirmed overall, and the association of
+    acute viral infection with bilateral plexus involvement. The description states that 38.6%
+    means most attacks have no identified trigger, so this is a precipitant and not a necessary
+    cause. HEV remains as the clinically distinctive instance rather than the only one.
+
+    Blocking 5 - The diagnostic action implied by the HEV finding was absent. Added a
+    "Hepatitis E serology in phrenic or diaphragmatic involvement" diagnosis entry, explicitly at a
+    lower confidence than the imaging entry: it rests on two uncontrolled series plus a suggestion
+    in the case-control study, and no study has evaluated it as a strategy, so its yield is unknown.
+
+    Suggestions taken: corticosteroid comparative data (PMID:19321467, with the authors' own
+    "pending a prospective, randomised trial" caveat), imaging yields (PMID:36214185, 90% and 74%,
+    with the timing distribution that makes a repeat scan the right response to a normal early
+    one), the surgical comparative series quoted as counts rather than percentages because the arms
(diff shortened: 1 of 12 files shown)


aleixpuigb:
All five blocking items fixed plus six of the seven suggestions, in one push (`cb4bab8d0`). 33/33 → **54/54 snippets**, 7 → 16 references, ten of them now reporting original human observations.

Your central criticism is the part I want to acknowledge directly, because it's exactly right and it's not a nit: **the entry's own `notes:` complained the evidence base was thin while the report committed in the same PR held the studies that fix it.** Seven of forty verified references used. That is a worse failure than a missing citation, because the entry was articulate about a gap it had the means to close.

It's also the same defect [#10995](https://github.com/monarch-initiative/dismech/pull/10995) was pulled up on an hour earlier in different clothing — there the unread source was a 1,095-line full text mined only for its abstract. The general form is treating a source as consumed once *anything* has been taken from it.

**1. Paresthesia.** Now `FREQUENT` on PMID:16371410's 78.4%. I also rewrote the original snippet's `explanation` to say explicitly that it supports occurrence and carries no frequency — your "phenotype support is not frequency support" is the thing worth having written down. One caveat recorded: that 246-patient denominator mixes 199 idiopathic with 47 hereditary cases while this entry scopes to the acquired form, and the paper reports no separate rate by subtype.

**2. Fatigue.** `HP:0012378`, `FREQUENT`, wired `DIRECT` from the endurance node. The second evidence item is the one doing the work: *"There was no correlation of pain or fatigue with the level of residual paresis on a Medical Research Council scale"* — that dissociation is what separates the endurance model from simple deconditioning, and it's why the phenotype hangs off that node rather than off axonal injury.

**3. Rehabilitation.** Both the adjusted group difference and the responder analysis from PMID:36697215, with the imprecision stated rather than glossed: the CI nearly touches zero and 47 randomised became 37 analysed. Plus PMID:39402917 for the strength-training negative — worth curating because the obvious rehabilitation approach to a weak limb is the wrong one here, and an entry recording only what to do leaves that trap in place.

**4. General trigger.** Added as its own `environmental:` entry from PMID:39364568, with all three figures. Its description records that a confirmed trigger in 38.6% means most attacks have none identified, so this is a precipitant in a susceptible individual and not a necessary cause. HEV stays as the clinically distinctive instance rather than the only one, which was your framing.

**5. HEV serology.** Added to `diagnosis:`, explicitly at lower confidence than the imaging entry: two uncontrolled series plus a suggestion in the case-control study, and no study has evaluated it as a strategy, so its yield is unknown.

**Suggestions 6-10 and 12 taken.** Corticosteroid comparative data with the authors' own "pending a prospective, randomised trial" caveat; imaging yields including the timing distribution, which is the operationally useful part since the earliest constriction appeared within 12 hours and a normal early scan therefore doesn't exclude them; the surgical series quoted as counts rather than percentages because the arms are 11 and 13, with its self-assigned Therapeutic IV grade; a three-phase `progression:` section; phrenic nerve reconstruction linked to the diaphragmatic weakness phenotype, flagged in its own description as a review across all causes of phrenic injury with no NA-specific outcome; and the `mechanism_confidence` vs `causal_link_type` wording fix.

**Suggestion 11 declined,** and I'd rather say why than quietly skip it. The node's claim *is* the conjunction — that no single factor is sufficient — so splitting the biomechanical arm out would assert three independent contributing mechanisms where the literature asserts one interaction. If the blood-nerve-barrier work firms up into a standalone mechanism it should become its own node then, on that evidence.

**Thank you for resolving the ECTO lead.** The `notes:` now records a confirmed absence and names what was searched, rather than an open question. That distinction is exactly what I got wrong on #10995 — asserting an ontology absence I hadn't actually established — so having it settled the right way round here matters more than the binding.

**On `NCIT:C2322`:** already filed as #10978, earlier today, with the root cause — NCIT classifies corticosteroids under Hormone rather than Pharmacologic Substance, so the enum's `NCIT:C1909` root can't reach them, and the same will be true of thyroid hormone, insulin, oestrogens and androgens. The entry's `notes:` now points there rather than carrying the analysis alone.

One thing to flag on scope: `references_cache/PMID_29925717.md` (the feline paper) is committed although no evidence item cites it. The entry names that PMID in `notes:` when explaining why the model was rejected, so a reader checking that reasoning needs the cache. Happy to drop it if you'd rather cache files tracked evidence citations strictly.

All gates clean. I also checked explicitly that all seven phenotypes are the target of a `downstream` edge — `check-causal-targets` doesn't detect orphans, which is how five of them slipped through on #10995.

---

PR #10999, opened 2026-09-04
curate: Add Spondylo-megaepiphyseal-metaphyseal Dysplasia (MONDO:0013228, NKX3-2)

### LINKED ISSUE #10957: Curate spondylo-megaepiphyseal-metaphyseal dysplasia (MONDO:0013228)

Curate a dismech entry for **spondylo-megaepiphyseal-metaphyseal dysplasia** ([MONDO:0013228](https://monarchinitiative.org/MONDO:0013228)), SMMD — biallelic *NKX3-2* (BAPX1) variants. *NKX3-2* is a homeobox transcription factor acting in the chondrocyte differentiation program, so this is a developmental-transcription-factor skeletal dysplasia: disproportionate short stature with a short neck and trunk, abnormally large epiphyses, and delayed ossification.

Clean leaf: no MONDO descendants, one causal gene (`hgnc:951` NKX3-2). Parent is `MONDO:0016761` spondyloepiphyseal dysplasia, which is the grouping — this term is the specific *NKX3-2* entity within it.

Worth watching: the disease is reported in very few families, so the phenotype list should stay inside what the primary reports actually state rather than borrowing the broader SED picture from the parent term.

No existing coverage: *NKX3-2* appears in zero files under `kb/disorders` on `main`, and no open PR curates it.

Stub: `stubs/Spondylo-megaepiphyseal-metaphyseal_Dysplasia.yaml`

aleixpuigb:
New entry for SMMD, the NKX3-2-related skeletal dysplasia.

Closes #10957

## The mechanism

NKX3-2 loss of function derepresses the chondrocyte maturation program, and the resulting endochondral ossification defect is **deranged rather than simply deficient**: axial vertebral ossification is delayed or absent while appendicular epiphyses are enlarged and pseudoepiphyses appear at all metacarpals and phalanges. That combination is what the disease name describes, and it is why the entry models one node producing opposite-signed consequences in two compartments rather than a global ossification failure.

A second, later-acting arm comes from the adult viable zebrafish mutant — loss of post-embryonic restraint on joint chondrocyte proliferation — which accounts for the overgrowth component that the embryonic lethality of the mouse null cannot show.

## The cervical spine is the part that matters clinically

Reduced or absent ossification of the cervical vertebrae leaves the spine mechanically unsupported, producing instability with anterior or posterior kinking (the swan-neck deformity, *kyknodysostosis*), and from there cord injury presenting as limb spasticity. **Five of six patients** in the only series to examine this developed it.

This gets its own pathophysiology node rather than being folded into the general ossification derangement, for two reasons the source supplies. It is the rule within that series rather than a complication. And the authors state the high incidence of cervical spine deformation here is *unique among skeletal dysplasias*, so it is not a generic consequence of poor vertebral ossification inherited from a sibling disorder.

It is also **the only point in this disease where an intervention changes the outcome**: the upstream lesion is an embryonic transcription-factor failure and is inaccessible after birth, while cervical instability is detectable and stabilisable before the cord is injured. So the node carries the two phenotypes it causes (`HP:0008462`, `HP:0001257`), the prophylactic stabilisation it argues for, and the surveillance that makes that possible.

**Both new frequency bands carry their fragility, not just their number.** 5/6 is 83% and sits in `VERY_FREQUENT`, but on six patients from a single series whose authors flag the small denominator themselves — that caveat is in the evidence `explanation`, not just implied. The stabilisation is recorded as a proposal rather than a demonstration: the source says early diagnosis *"might prevent"* the complications, and no series has reported outcomes in prophylactically stabilised patients. With six patients described in total, there is no prospect of one soon.

## I mined the report before opening this PR, not after

Worth flagging because it is a direct response to the two concurrent PRs in this batch. [#10995](https://github.com/monarch-initiative/dismech/pull/10995) was pulled up for mining a 1,095-line cached full text only for its abstract; [#10998](https://github.com/monarch-initiative/dismech/pull/10998) for using 7 of 40 verified references while its own `notes:` complained the evidence base was thin. The general defect is treating a source as consumed once *anything* has been taken from it.

My first draft here covered cervical cord compression only in the perinatally lethal neonate, as a single-case observation. The report surfaced PMID:22791571, a six-patient series devoted to exactly this. That is the largest single content gain in this entry and I would have shipped without it.

## Report identity and what was not taken from it

This is **the only one of three openscientist reports in this batch to state a correct MONDO ID** — `MONDO:0013228`, matching the entry and the stub, with OMIM 613330 agreeing on both sides. (#10995's drifted to a sibling disease and was not committed; #10998's proposed `MONDO:0100053`, which is anaphylaxis.)

`just preflight-dr` returns WARN because RUNX2 is mentioned at 28% of NKX3-2's rate and it suspects a second disease entity. It isn't one: NKX3-2 represses RUNX2, and that repression is this entry's second pathophysiology node. Recording that here so the warning is not re-investigated.

Not used from the report: the obsolete `GO:0051092`; `HP:0006262`, which it labels "Pseudoepiphyses of the hand bones" when it is *Aplasia/Hypoplasia of the 5th finger*; its off-topic `PMID:17310243`; and its one unsupported quote (`PMID:10572046`). All four were flagged by the report's own validation sections.

## Two deliberate omissions

**Duodenal atresia stays unwired into the pathograph.** It is a single-patient observation with no sourced mechanistic link to the skeletal lesion. The reason is in its `description` — an unsupported edge is worse than an unconnected phenotype.

**No `conforms_to`.** `defective_skeletal_mineralization` was checked and does not fit: it scopes to rickets, a mineralization failure of otherwise normal cartilage, whereas the lesion here is a transcriptional failure of the chondrocyte maturation program upstream of mineralization.

## Validation

39/39 snippets verified against cached references. `just validate-disorders` passes; terms, entity refs, causal targets, duplicate keys, qualifier terms and enum values all clean. Three additive cache rows. Every phenotype except the deliberately-excluded duodenal atresia is the target of a `downstream` edge, checked explicitly since `check-causal-targets` does not detect orphans.

---


aleixpuigb:
All six blocking findings fixed plus all three suggestions, in one push (`3b4f42a92`). 39 → **51 verified snippets**.

**Items 5 and 6 are the ones I want to own directly.** You're right, and the failure is worse than "under-consumption" — I treated the report's own advisory flags as verdicts. Its relevance section says, in the paragraph immediately above the flag I cited: *"That is a clue and not a verdict — a paper can be relevant in ways its title and abstract do not spell out — so read them before deciding."* I quoted the flag and did not read the paper. PMID:10572046's "unsupported quote" turned out to differ from its source by a single `&agr;` entity, with the supporting text printed directly beneath the flag.

That is the inverse of the mistake I made on [#10995](https://github.com/monarch-initiative/dismech/pull/10995), where I asserted an ontology absence I hadn't checked. There: too much trust in my own assumption. Here: too much trust in an automated flag, in the direction of discarding good evidence. Both are the same failure to read the primary text, and I've written both into the history record.

What it cost was not a citation but **an entire arm of the mechanism**. NKX3-2 is a dual-function node — a maturation brake *and* a survival signal for the cells being braked, via ligand-independent constitutive RelA activation. That's now `Loss of Chondrocyte Survival Signalling`, wired off the LoF node and marked `PROVISIONAL`, since the mechanism is established in chick and mouse chondrocytes and has not been shown in an SMMD patient. The Bapx1-null paper is now cited directly on the animal model rather than through the human-genetics paper that summarised it, and it supplies the dissociation that justifies two arms rather than one: losing the gene *raises apoptosis while leaving cell cycling rates unaltered* — a lost survival signal, not a lost proliferation brake.

**Items 1–4, all as diagnosed.** `GO:0035988 chondrocyte proliferation` replaces the differentiation term — you're right that it asserted the opposite of the node above it, not merely something vaguer. `DYSREGULATED` replaces `DECREASED`, which contradicted the entry's own thesis. The epiphysis term is split three ways, and `HP:0025369` gives the entry its only metaphyseal finding, which I hadn't noticed was missing entirely. Pubic and ischial non-ossification is now its own phenotype.

Your framing of the pseudoepiphysis case is the lesson I'll carry: I rejected the report's `HP:0006262` correctly, and then **dropped the concept instead of searching for the right term**. Rejecting a wrong binding is only half the job, and stopping there loses a finding both sources single out.

**All three suggestions taken**, with one carve-out. Scoliosis loses its `FREQUENT` band (no source counts it) and drops to `evidence_source: OTHER` — you're right that a zebrafish paper's background sentence isn't human clinical evidence. Supernumerary distal limb joints and coronal cleft vertebrae added. NKX3-1 added as `MODIFIER` with both halves of the result, since the single null having *no* skeletal phenotype is what makes it a modifier rather than a second disease gene.

**Declined: the gnomAD constraint metrics**, with the reason recorded in the genetic `notes:`. They're a database lookup with no citable publication behind them in the report, and I couldn't verify them from this environment. Recorded as a lead rather than asserted — the same standard I applied to the ECTO question on [#10998](https://github.com/monarch-initiative/dismech/pull/10998).

**One thing your review caused that I then had to catch myself:** adding four phenotypes created three new pathograph orphans — the exact defect #10995 was pulled up on. Two are now wired `DIRECT` from the ossification node. Supernumerary distal limb joints is deliberately left unwired: the joint-proliferation node is the tempting target and both concern joints, but that's a shared anatomical locus rather than a stated mechanism, and an edge drawn because the word "joint" appears in both would be the unsupported inference this entry declines elsewhere. Reason recorded in its description.

Also caught during the round: I wrote a `reference_title` from memory instead of copying it from the cache, and `linkml-reference-validator` rejected it. Fixed from the cache — a good demonstration that the title check earns its place.

---

**Separately, an update on the deep-research retry**, since it bears on this PR's sibling. I re-ran the MC1DN21 report with the explicit `--var mondo_id=MONDO:0032625` override the #10995 reviewer suggested. The override *works* — the report received and echoed `MONDO:0032625` — and the report is **still about a different disease**, a third one: it names *NDUFAF6*, mentions NUBPL zero times, and pairs the correct MONDO ID with the wrong OMIM number. `just preflight-dr` catches it in under a second, offline. Full detail on #9888; the upshot is that the recipe fix is necessary but not sufficient, and running `preflight-dr` automatically is the load-bearing half.

---

Issue #11002, opened 2026-09-04
kb_data sweep collects ~60k one-file test items and spends most of its CPU on pytest/xdist overhead, not on checks


caufieldjh:
## Problem

`tests/test_data.py` parametrizes 36 test functions over every KB file (`DISORDER_FILES`, `MODEL_BEARING_FILES`, `CONFORMS_TO_FILES`, ...). Each item opens and parses one YAML file and runs one small check. On origin/main the module alone collects 59,964 items. The number grows with every entry and every new per-file check.

Measured on a slightly older main (34,690 items, 31,148 in the `kb_data` sweep), 12-core WSL2, `-n 6`, after the `_disease_names` cache from #9258 landed:

| Run | Wall | CPU |
|---|---|---|
| Full `kb_data` sweep | 3 min 18 s | 1007 s user |
| One trivial parametrized test over all 1990 disorder files, single process | 22.5 s, of which 12.6 s is collection | |
| Parsing the whole disorder corpus once with `dismech.yaml_io.safe_load` | ~7 to 10 s | |

So the actual work is roughly 16 corpus parses, on the order of 150 CPU seconds. The other ~850 CPU seconds are per-item overhead: each of the 6 workers collects all items itself (about 13 s each), and xdist serialises a report for each of 31k items.

Two smaller things in the same family:

- `test_hypothesis_based_definition_attaches_to_foreign_keys` is parametrized over `DISORDER_FILES` but carries no `kb_data` marker, so `just test-code` runs its ~2000 items too.
- Long parametrize ids (absolute paths) make collection and report traffic heavier than they need to be.

## Proposal

Parse each KB file once and run every per-file check against that parse, so the sweep has one item per file (about 2200) instead of one item per file per check. Two shapes that would work:

- One `test_disorder_file(filepath)` that calls a list of check functions and collects their errors, reporting all failures for the file at once. Keeps the per-file id for `-k` selection. Loses per-check granularity in the report, which can be recovered by naming the failing check in the message.
- Keep the separate tests but load documents through an `lru_cache`d `_load(filepath)` and group items by file so the same worker sees the same file (`--dist loadgroup` with an `xdist_group` mark per file). Less change, smaller win, since collection overhead stays.

Either way the `kb_data` marker split from #5155 stays as is. This is about how fast the sweep is when it does run, not whether it runs.

Related: #9298 (the cached helper, already fixed on main), #8638 (network closure in the same sweep), #11001 (the schema fixture was a no-op).

Issue #8638, opened 2026-08-16
tests/test_data.py is too slow to run locally — grouping/closure fixtures exceed timeouts


cmungall:
## Problem

`tests/test_data.py` is prohibitively slow to run locally — even scoped invocations like `uv run pytest tests/test_data.py -k "grouping"` repeatedly exceeded a 120–200s wall-clock budget and had to be killed (exit 143) before emitting any pass/fail line. This came up while validating a new grouping: the authoritative offline checks (`check-groupings`, schema, term validation, the `conforms_to` node-reference test) all completed quickly, but the corresponding pytest could not be run to completion in a reasonable window.

## Likely cause

The suite appears to compute ontology closures at collection/fixture time. Grouping membership evaluation (`src/dismech/groupings.py`) resolves criteria over the HP/GO **closure**, and closure is computed via the OAK `sqlite:obo:*` / `ols:` adapters. When the test module builds those closures eagerly (and/or over the network) for every grouping and every disorder at import/fixture setup, the fixed cost dominates and scales with KB size — so the whole module is slow even when `-k` selects a handful of tests.

This is the test-suite analog of the curation-loop split already documented in `CLAUDE.md` (fast cache-only `count-verified-snippets` vs. the slow network-touching `validate-references`): the deterministic offline path is seconds, the ontology-touching path is minutes-plus.

## Impact

- Contributors (and agents) cannot practically run the data tests locally before pushing, so they fall back to the faster offline validators and rely on CI for the pytest gate.
- Repeated local timeouts waste time and produce misleading `exit 143` "failures" that are actually terminations, not assertion failures.

## Possible directions (for discussion, not prescriptive)

- **Session-scope and cache the closures.** Compute each criteria term's closure once per session (module/session-scoped fixture) instead of per-test/per-file, and reuse the cached closure the production code already builds.
- **Force cache-only / offline closure in the test path.** The membership evaluator already degrades to exact matching when the ontology is unreachable (per `CLAUDE.md`); a test-mode flag that pins it to the committed `cache/enums/*.csv` + `cache/*/terms.csv` would make the suite deterministic and fast, at the cost of some closure coverage (which the dedicated `check-groupings`/term-validation jobs still exercise).
- **Split the module.** Separate the fast structural/foreign-key assertions (unique names, member FK resolution, `conforms_to` node references — these are pure YAML graph checks) from the slow closure-dependent membership assertions, so contributors can run the fast half in seconds.
[...]

Issue #7652, opened 2026-08-01
Whole-KB conformance sweep (test-kb) is not in the PR gate, so KB-breaking entries merge with green CI


kevinschaper:
Found during a 16-disease curation batch. Two agents disagreed about whether `main` was broken; resolving that disagreement turned up a real gap in the CI gate.

## The gap

`justfile` splits the test suite deliberately:

```
test: test-code test-kb

# Fast code/logic tests: everything except the whole-KB `kb_data` conformance sweep
test-code: _test-schema _test-python-code _test-examples test-search test-extension

# Python code/logic tests, excluding the whole-KB `kb_data` sweep.
test-python-code: _test-python-code

# Whole-KB schema-conformance sweep (parametrized over every KB file), parallelized.
test-kb: ...
```

The PR gate (`.github/workflows/main.yaml`, the `Build and test` check) runs `just test-schema` and `just test-python-code` — **it never runs `just test-kb`.** So the whole-KB conformance sweep, the thing that actually checks every entry against the schema's cross-cutting rules, is not gating anything.

## Demonstration

`test_non_therapeutic_actions_do_not_use_treatment_targets` currently **fails on `main`** against `kb/disorders/SETD5_Haploinsufficiency_Syndrome.yaml`. Reproduced in a clean worktree at `origin/main` (`8a30e01e44`) with zero local changes. Meanwhile `Build and test` on `main` shows five consecutive `success` runs, because the failing test isn't in the gate.

So `main` is simultaneously "green" and carrying a KB conformance failure. That is the worst combination: the signal says fine, the invariant is broken, and nobody is told.

## Why it matters beyond this one file

Curation PRs are almost entirely `kb/` changes. The suite designed to validate exactly those changes is the one excluded. In this batch several agents ran `pytest tests/test_data.py` locally and hit multi-minute or timing-out runs (16k+ parametrized tests), so the practical effect is that whole-KB conformance is checked *only* when a curator happens to run it by hand and waits it out.

It also actively misleads. I told an agent "treat `main` as green" on the strength of the CI badge; it correctly refused to take that on trust, reproduced the failure against a clean `origin/main`, and identified the gate gap. An agent that had believed me would have been primed to dismiss a genuine regression of its own as "pre-existing."

## Suggested fix

Some combination of:
1. **Add `just test-kb` to the PR gate**, ideally scoped to changed KB files for speed (the workflow already computes `kb_disorders_files` for `just validate-disorders`, so the plumbing exists).
[...]

PR #11006, opened 2026-09-04
tests: give the Disease validator fixture a real JSON Schema plugin (fixes #11001)

### LINKED ISSUE #11001: tests/test_data.py `validator` fixture has no validation plugin, so the whole-KB schema-conformance sweep asserts on nothing
## Problem

The module-scoped `validator` fixture in `tests/test_data.py` is built as `Validator(SCHEMA_PATH)`. linkml's `Validator` has no default plugin. With `validation_plugins=None`, `iter_results_from_source` returns `[]` before it looks at the instance. Every report is empty. Every assertion on it passes.

Proof, on the current tree:

```python
from linkml.validator import Validator
val = Validator("src/dismech/schema/dismech.yaml")
val.validate({"name": 123, "bogus_slot": 1, "phenotypes": "notalist"}, target_class="Disease").results
# []
```

This has been the case since the initial commit (linkml 1.9.5 then, 1.11.1 now; neither has a default plugin).

## What it silences

- `test_valid_disorder_files`, `test_valid_comorbidity_files`, `test_valid_grouping_files`. The `kb_data` sweep that `just test-kb` and the CI step "Run whole-KB schema-conformance sweep" exist to run.
- Every positive `*_validates(validator)` test in the module (biological scale, food source, dietary modifications, action category, reference ranges, subtype NCIT mappings, multivalued subtypes, and so on). They construct a synthetic instance and assert no errors. Nothing was ever checked.

The `synthesis_validator`, `hypothesis_assessment_validator`, and `hypothesis_reconciliation_validator` fixtures in the same file already pass `JsonschemaValidationPlugin(closed=True)`, as do `tests/test_genetic_schema.py`, `tests/test_clinical_burden_schema.py`, and `tests/test_history_schema.py`. Only the main Disease fixture was missed. `test_reference_range_band_rejects_invalid_abnormal_flag` even builds its own strict validator with the comment that the fixture "does not enforce enum membership".

## Cost of turning it on

Real closed validation is cheap. Measured on a 12-core WSL2 box:

| What | Time |
|---|---|
| One disorder file, average | ~50 ms |
| Largest file (`Fanconi_Anemia.yaml`, 320 KB) | 235 ms |
| First call (JSON Schema generation, once per process) | 1.2 s |
| All 2199 KB files, serial, one process | 37 s |
| The `-k valid_` sweep on origin/main, 2865 items, 6 xdist workers | 41 s |

Every KB file passes. One synthetic test does not: `test_phenotype_multivalued_subtypes_validates` builds a `disease_term` without the required `preferred_term`. That is a test-data bug the empty validator hid.

## Fix

Pass `validation_plugins=[JsonschemaValidationPlugin(closed=True)]` to the fixture, matching the sibling fixtures, and add the missing `preferred_term` to the synthetic instance. PR to follow.

Found while profiling the sweep for #9298 and #8638.

caufieldjh:
## What

`tests/test_data.py`'s module-scoped `validator` fixture was `Validator(SCHEMA_PATH)` with no plugin. linkml's `Validator` has no default plugin, so every report it produced was empty and every assertion on it passed. That covers `test_valid_disorder_files`, `test_valid_comorbidity_files`, `test_valid_grouping_files` (the `kb_data` schema-conformance sweep) and every positive `*_validates(validator)` test in the module. Details and proof in #11001.

This PR passes `validation_plugins=[JsonschemaValidationPlugin(closed=True)]`, matching the sibling `synthesis_validator`, `hypothesis_assessment_validator`, and `hypothesis_reconciliation_validator` fixtures in the same file and the fixtures in `tests/test_genetic_schema.py`, `tests/test_clinical_burden_schema.py`, and `tests/test_history_schema.py`.

## What turning it on found

One synthetic instance was invalid: `test_phenotype_multivalued_subtypes_validates` built a `disease_term` with no `preferred_term`, which the schema requires. Added the field. Every real KB file already passes closed validation.

## Verification (worktree on origin/main 9f6d0e5d32)

| Run | Result |
|---|---|
| `pytest tests/test_data.py -m "not kb_data"` | 8214 passed, 3 min 59 s |
| `pytest tests/test_data.py -m kb_data -k valid_ -n 6` | 2865 passed, 41 s |
| `uvx ruff check tests/test_data.py` | clean |

Real closed validation costs about 50 ms per disorder file, 235 ms for the largest (`Fanconi_Anemia.yaml`), and 1.2 s once per process to generate the JSON Schema. The whole KB serial is 37 s. So the sweep stays affordable.

## What this does not change

The `kb_data` marker split, the CI gating of the sweep on schema changes (#7652), and the sweep's remaining cost (#11002 item overhead, #8638 network closure) are untouched.

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -1716,18 +1716,13 @@ def test_reference_range_interpretation_bands_validate(validator):
     assert not errors, f"Unexpected validation errors: {[str(e) for e in errors]}"
 
 
-def test_reference_range_band_rejects_invalid_abnormal_flag():
-    """An out-of-enum abnormal_flag on a band must fail strict validation.
+def test_reference_range_band_rejects_invalid_abnormal_flag(validator):
+    """An out-of-enum abnormal_flag on a band must fail validation.
 
-    Uses a closed jsonschema validator because the lenient module-scoped
-    ``validator`` fixture does not enforce enum membership.
+    The negative counterpart of the positive ``*_validates`` tests above: it is
+    the one check in this module that fails if the ``validator`` fixture stops
+    enforcing the schema, which is what happened before it carried a plugin.
     """
-    from linkml.validator import Validator as _Validator
-    from linkml.validator.plugins import JsonschemaValidationPlugin
-
-    strict = _Validator(
-        SCHEMA_PATH, validation_plugins=[JsonschemaValidationPlugin(closed=True)]
-    )
     data = {
         "name": "Test Disease",
         "biochemical": [
@@ -1750,7 +1745,7 @@ def test_reference_range_band_rejects_invalid_abnormal_flag():
             }
         ],
     }
-    report = strict.validate(data, target_class="Disease")
+    report = validator.validate(data, target_class="Disease")
     errors = [r for r in report.results if r.severity.name == "ERROR"]
     assert errors, "Expected a validation error for an invalid abnormal_flag value"
 


caufieldjh:
Pushed fa34f71c32 for the blocking item.

**Changed.** `test_reference_range_band_rejects_invalid_abnormal_flag` now takes the module `validator` fixture, the local `Validator` / `JsonschemaValidationPlugin` imports and the `strict` instance are gone, and the docstring says what the test is for: the one negative check in the module that fails if the fixture ever stops enforcing the schema. I went with the fixture rather than an independent validator, since the point of this PR is that the fixture is now the thing worth trusting, and a negative test on it is what proves that.

**Not changed.** `ruff format --check` reports two pre-existing spots in this file (lines around 1300 and 1340, untouched by this PR). CI runs `ruff check` only, which is clean.

**Follow-up.** Filed #11011 for the other plugin-less validators you found: `tests/test_stubs.py:59` and the four sites in `tests/phenoagent/`. Confirmed each on origin/main before filing.

Verification: `pytest tests/test_data.py -k reference_range` passes (3 items); `uvx ruff check tests/test_data.py` clean.

Issue #11005, opened 2026-09-04
`just validate <file>` pays about 11 s of fixed cost per entry: normalize-cache and linkml-validate startup


caufieldjh:
## Problem

The per-entry curation loop (`just validate kb/disorders/X.yaml`, and the `validate-pre-edit` recipe the write hook runs on every edit) pays costs that have nothing to do with the entry being validated. Measured on `Sweet_Syndrome.yaml` (44 KB), 12-core WSL2:

| Stage | Time | Of which |
|---|---|---|
| `uv run linkml-validate --schema ... --target-class Disease <file>` | 5.0 s | 2.4 s importing `linkml.validator.cli`, 0.4 s schema load, 1.2 s JSON Schema generation, ~50 ms validating the file |
| `just normalize-cache` | 6.4 s | `linkml-term-validator migrate-cache --sort-only` plus a sort of every `cache/enums/*.csv` |
| `just fix-references-cache <file>` | 0.3 s | |
| term validation through the wrapper | 13.4 s | see #11004 |

The validation of the file itself is around 50 ms. Everything else is startup and housekeeping. With the KB growing, this per-entry overhead is what curators and agents pay most often.

## Proposals

- **Skip `normalize-cache` when the caches did not change.** `git diff --quiet -- cache/` before and after the term validator, and normalise only on a change. `validate-pre-edit` already omits it; `validate` could do the same and leave normalisation to `validate-disorders` and `qc`.
- **Reuse the generated JSON Schema.** `JsonschemaValidationPlugin` accepts `json_schema_path`; a schema generated once by `gen-json-schema` into `tmp/` and invalidated on `src/dismech/schema/**` mtime would save the 1.2 s generation per call. Whether `linkml-validate` exposes that via `--config` needs checking.
- **One interpreter for the loop.** Schema and term validation are both Python and both start their own interpreter and import tree. A single `dismech.validate_entry` entry point that runs both plugins in one process, or a batched invocation that takes several files (as `validate-disorders` already does for CI), removes one or two startups per entry.

The 5 s `linkml-validate` cost is also why the `kb_data` pytest sweep should stay the place for whole-KB schema checks: in-process it validates all 2199 KB files in 37 s serial.

Related: #11004, #11001.

caufieldjh:
[...]

Issue #11004, opened 2026-09-04
scripts/run_term_validator.sh runs a `--help` probe before every validate-data call, doubling per-file term validation time


caufieldjh:
## Problem

`scripts/run_term_validator.sh` runs `uv run linkml-term-validator validate-data --help` on every invocation to find out whether the installed version supports `--strict` or `--fail-on-warnings`. That probe is a full interpreter start plus the linkml import tree. Then the real call pays the same startup again.

Measured on one already-cached disorder file (`Sweet_Syndrome.yaml`, 44 KB), 12-core WSL2:

| What | Time |
|---|---|
| `just validate-terms kb/disorders/Sweet_Syndrome.yaml` (through the wrapper) | 13.4 s wall |
| `linkml_term_validator.cli.main(...)` in-process, cProfile | 5.3 s |
| `uv run python -c "import linkml.validator.cli"` | 2.4 s |
| `uv run python -c pass` | 0.14 s |

So roughly 5 s of work, and 8 s of two interpreter startups plus `uv run` overhead. `validate`, `validate-pre-edit` (the Claude Code write hook), and `validate-disorders` all go through this wrapper, so every entry edit pays it.

Inside the 5.3 s of work, for whoever wants the next layer:

- 2.2 s in `binding_plugin.pre_process` walking `induced_slot` for 136 classes (1267 calls).
- 1.5 s building the `Validator` (schema load).
- 0.7 s in `oak_utils.load_cache`, called 33 times for one file. Upstream behaviour.

## Proposal

Drop the probe. `linkml-term-validator` is pinned in `uv.lock`; the wrapper's own comment says the probe exists until linkml/linkml-term-validator#29 ships, and it can check that once at review time rather than at every call. If a runtime check is still wanted, cache the answer keyed on the installed version (`uv run python -c "import importlib.metadata as m; print(m.version('linkml-term-validator'))"` is still a startup, so better to write the result to a file under `tmp/` and reuse it).

Related: #11005 for the other fixed costs in `just validate`.

caufieldjh:
[...]

PR #11007, opened 2026-09-04
just validate: skip normalize-cache when caches are unchanged, reuse a cached JSON Schema (#11005)

caufieldjh:
## What

Addresses the first two proposals in #11005. The single-interpreter idea (one process for schema and term validation) is left for a follow-up; it overlaps with #11004.

**Conditional `normalize-cache`.** `validate`, `validate-disorders`, and `validate-all` take a stamp (`mktemp tmp/dismech_cache_stamp.XXXXXX`) before the validators run, and a new private recipe `_normalize-cache-if-changed <stamp>` runs `normalize-cache` only when some file under `cache/` is newer than the stamp. The term validator writes to `cache/` only when it learns a new term, so a typical single-entry run skips the re-sort. A cache that was already out of order before the run is still caught by `check-cache-order` and `check-term-cache-integrity`. The stamp is removed on exit by a `trap`.

**Cached JSON Schema for `linkml-validate`.** A new private recipe `_linkml-validate-config <class>` writes `tmp/linkml-validate/<class>.closed.json` with `gen-json-schema --closed --top-class <class> --include-range-class-descendants` and a small `linkml-validate --config` YAML pointing at it. It regenerates when anything under `src/dismech/schema/` or `uv.lock` is newer than the cached document. `validate`, `validate-pre-edit` (the write hook), and `validate-schema` use it. The batched recipes (`validate-all`, `validate-disorders`, `validate-comorbidity-batch`) keep the plain command line; they pay generation once per run and `tests/test_enum_cache.py` pins their exact form.

## Equivalence

The cached document is byte-identical to the schema `JsonschemaValidationPlugin(closed=True)` builds in memory (compared via `ValidationContext.json_schema_validator(...).schema` on the current tree). A deliberately broken instance yields the same 3 errors both ways; `Sweet_Syndrome.yaml` passes both ways.

## Measurements (clean venv, 12-core WSL2, `Sweet_Syndrome.yaml`)

| Run | Before | After |
|---|---|---|
| `just validate-pre-edit` | 11.4 s | 8.3 s (warm cache); 10.8 s on the first call that generates it |
| `just validate-schema` | 2.2 s | 1.4 s |
| `just normalize-cache` inside `validate` / `validate-disorders` | 2.8 s | skipped when `cache/` is unchanged |

A note on the numbers in #11005: they were taken in a checkout whose `uv run` reinstalled a broken `dismech` dist on every call, which roughly doubled every interpreter start. I have posted the clean figures on the issue.

## Verification

- `pytest tests/test_enum_cache.py tests/test_ci_batch_validation.py tests/test_validate_disorder_hook.py`: 36 passed. These pin the recipe bodies, including that `validate-pre-edit` calls no mutating step and that `project.justfile` uses only repo-local temp files.
- `just validate-disorders kb/disorders/Dorsalgia.yaml` and `just validate kb/disorders/Dorsalgia.yaml` both take the skip path and leave no stamp files.
- Forced change (`touch` on one enum cache file) takes the normalize path.
- `touch src/dismech/schema/dismech.yaml` triggers regeneration; a second call does not.

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -98,7 +98,8 @@ validate file:
     cache_stamp=$(mktemp tmp/dismech_cache_stamp.XXXXXX)
     trap 'rm -f "$cache_stamp"' EXIT
     echo "Schema validation..."
-    uv run linkml-validate --config "$(just _linkml-validate-config Disease)" {{file}}
+    lv_config=$(just _linkml-validate-config Disease)
+    uv run linkml-validate --config "$lv_config" {{file}}
     echo "Term validation..."
     {{term_validator}} validate-data {{file}} -s {{schema_path}} -t Disease --labels -c {{oak_config}}
     echo "Reference validation..."
@@ -134,7 +135,8 @@ validate-pre-edit file:
     #!/usr/bin/env bash
     set -e
     echo "Schema validation..."
-    uv run linkml-validate --config "$(just _linkml-validate-config Disease)" {{file}}
+    lv_config=$(just _linkml-validate-config Disease)
+    uv run linkml-validate --config "$lv_config" {{file}}
     echo "Term validation..."
     {{term_validator}} validate-data {{file}} -s {{schema_path}} -t Disease --labels -c {{oak_config}}
     echo "Reference validation (advisory, cache-bound)..."
@@ -198,7 +200,10 @@ validate-disorders *files:
 # Schema-only validation (fast, structure check)
 [group('QC')]
 validate-schema file:
-    uv run linkml-validate --config "$(just _linkml-validate-config Disease)" {{file}}
+    #!/usr/bin/env bash
+    set -e
+    lv_config=$(just _linkml-validate-config Disease)
+    uv run linkml-validate --config "$lv_config" {{file}}
 
 # Print the path of a `linkml-validate --config` file that points at a
 # pre-generated JSON Schema for `target_class`.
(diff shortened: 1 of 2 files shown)


caufieldjh:
Pushed 6c4b3e05f6. Both blocking items and the three suggestions, in one push.

**Fixed path temp file.** The document is now written to `mktemp "$dir/<class>.closed.json.XXXXXX"` and moved into place. Two overlapping hook runs each write their own file and the last `mv` wins with a whole document.

**The comment on the skip.** Rewritten. It now says ordering is audited only by `check-cache-order`, which is advisory and in neither `qc` nor CI, that `check-term-cache-integrity` checks row structure and not order, and that disorder introduced by a merge is no longer self-healed by `validate` runs and needs a manual `just normalize-cache`. I also noted the merge case on #11005.

**Suggestion 3, config shape.** `tests/test_enum_cache.py` now pins that the three single-file recipes go through `_linkml-validate-config`, that the template contains `json_schema_path` and no `data_sources`, that the temp name is a private `mktemp`, and that `--include-range-class-descendants` is passed explicitly (your note on the differing defaults is recorded there as the assertion message).

**Suggestion 4, unusable stamp.** A missing stamp now prints a message and normalizes. Verified: `just _normalize-cache-if-changed tmp/does_not_exist` runs the full normalize.

**Suggestion 5.** The single-file recipes assign `lv_config=$(just _linkml-validate-config Disease)` before use, so a generation failure aborts at the real error. The recipe comment now says `closed: true` in the generated config is documentary, and why the config must not carry `data_sources`.

Verification: 39 passed across `test_enum_cache.py`, `test_ci_batch_validation.py`, `test_validate_disorder_hook.py`; cold and warm `validate-schema` both pass with no stray temp files; `validate-pre-edit` on `Dorsalgia.yaml` passes end to end; `ruff check` and `ruff format --check` clean on the test file.

